PR #135868, opened 2026-09-02
feat(triage): start installation-owned recovery after update and startup failures

steipete:
## What Problem This Solves

Eligible update failures and repeated managed startup failures can leave an installation broken after the initiating process exits. This starts one installation-owned recovery attempt through the fresh installed CLI, preserving saved diagnostics and the original failure.

## Why This Change Was Made

Stage 5 of the #135868 split: concern A, B's triage lease/IPC/native lifetime protocol, and E's automatic failure entry points. Prerequisites and adjacent cleanup have landed separately. Maintainer-owned implementation by @steipete.

Automatic recovery dispatches through `agentExecCommand`; manual `triage --run` retains `runUpdateRepairLoop`. They share bounded diagnostics/prompt collection and the installation lease owner. One lease state machine and one detached helper coordinate current authority, cancellation, and cleanup. Staged runtime modules provide the build closure required after package replacement.

## User Impact

- Eligible unattended failures can attempt bounded repair and verify the original symptom and intended version/health without replacing the original exit or update result.
- Revoked or replaced ownership and disconnected continuations prevent later effects. Uncertain cleanup fences automatic successors until closure is established; an unconfirmed foreground cleanup remains fenced for the same OS boot, with manual triage available.
- Systemd startup recovery is supported. Launchd startup saves diagnostics and manual guidance.
- Cancellation, required migrations, missing capability consent, and ownership refusals remain operator-owned outcomes.

## Evidence

Candidate: `06443b0b249671c94cdc1ef0d4bb8dfc6c410de4`. Base at the last pre-push rebase: `7921ec48f82d41dac5a452b1c87066fad64b20e9`. Maintainer-owned implementation by @steipete.

**Landed as [e07af7166c1677fa8072f1ad19ab5a7a508ac76e](https://github.com/openclaw/openclaw/commit/e07af7166c1677fa8072f1ad19ab5a7a508ac76e) after the lead opened the gate.** This candidate is +2,251 production lines, 38 below the accepted +2,289 candidate. Tests/support are +4,118, below the selected source scope's +4,579.

- Exact-head CI is **green**: https://github.com/openclaw/openclaw/actions/runs/34061826083 . Both Windows shards pass.
- The complete `node scripts/check-changed.mjs` plan passed: all type/UI/extension lanes, tooling/production/full-tree dead-export scans, lint, and zero runtime import cycles. Local OS isolation excludes unrelated ancestor-checkout dependencies while retaining every check.
[...]

PR #140489, opened 2026-09-06
refactor(plugin-sdk): simplify facade contracts and private relays

steipete:
## What Problem This Solves

The shared plugin facade path repeats its input contracts and passes values through two private wrappers. These copies add maintenance work around resolution and activation without owning separate policy.

This is an independent cleanup contribution to the #135868 split campaign.

## Why This Change Was Made

Ten repeated parameter shapes reuse existing resolver contracts, and the duplicate location shape uses its canonical type. A private registry relay is inlined at its sole fallback call. The private config reader returns the same config directly to its sole caller, which previously unwrapped a fresh envelope immediately.

The change preserves method receivers, configuration identity, source/runtime snapshot order, empty fallback identity, exception scope, custom-environment cache bypass, activation checks, and the existing synchronous/source/asynchronous loading paths. Both original owners match stable v2026.9.2.

## User Impact

No user-visible behavior change. Plugin facades retain their existing activation, loading, caching, and error contracts.

## Evidence

- All 84 existing tests pass before and after across ten direct and representative consumer suites, including cold activation, disabled-plugin rejection, artifact failure, cached export identity, source-snapshot preference, and linked QA-runner loading.
- Actual-source type comparison passes 180 assertions per optional-property mode, including public/generic signatures, ten input shapes, nullable location, and optional/readonly keys. Required-environment and non-null-location mutations are rejected in both modes.
- A structural emitted-code comparison permits only the two reviewed private wrapper eliminations. It verifies diff scope; runtime evidence comes from the boundary and consumer tests.
- The extra strict-optional mode retains three existing owner diagnostics per side, with identical messages, codes and spans. Contract assertions are clean; no whole-repository strict-mode compilation claim.
- Full changed-check plan passes on isolated Linux Blacksmith Testbox: core/core-test and extension/extension-test types, SDK surface guards, all three Knip export scans, lint and architecture guards.
- Full pnpm build passes, including all 152 public SDK subpaths and Control UI performance gates. The build leaves tracked files unchanged; 22 source/input hashes match before and after both commands.
- Independent Codex review is clean through P2. ClawSweeper found no correctness/security defect.

ClawSweeper checklist disposition: the requested data-model migration proof is not applicable to this diff. The removed `{ rawConfig }` object is a private, process-local return envelope; its only caller immediately extracts the property before any other use. It is never persisted or serialized. The same source/runtime snapshot getters, config path, `existsSync`/`readFileSync`, JSON5 parser, parsed config object, and empty fallback are retained. No schema, serializer, stored field, version, or migration changes. The stable v2026.9.2 comparison, exact projected reader/caller type checks, narrowly constrained emitted-code comparison, and before/after activation/snapshot tests support that compatibility conclusion. No migration or baseline change was added for this checklist. The final-head ClawSweeper technical review independently confirmed that the previous migration concern is resolved and there is no introduced storage/serialization boundary; its generic checklist remains displayed.

Production: two files, 543 → 498 lines; +41/−86 = **−45**. Tests, tooling, docs and baselines are unchanged.

### DIFF
--- a/src/plugin-sdk/facade-activation-check.runtime.ts
+++ b/src/plugin-sdk/facade-activation-check.runtime.ts
@@ -24,48 +24,41 @@ import { ALWAYS_ALLOWED_RUNTIME_DIR_NAMES } from "./facade-activation-contract.j
 import {
   resolveBundledMetadataManifestRecord,
   resolveRegistryPluginModuleLocationFromRecords,
+  type FacadeModuleLocationLike,
   type FacadePluginManifestLike,
 } from "./facade-resolution-shared.js";
 
 const ALWAYS_ALLOWED_RUNTIME_DIR_NAME_SET = new Set<string>(ALWAYS_ALLOWED_RUNTIME_DIR_NAMES);
 const EMPTY_FACADE_BOUNDARY_CONFIG: OpenClawConfig = {};
 
-type FacadeModuleLocation = {
-  modulePath: string;
-  boundaryRoot: string;
-};
+type FacadeActivationCheckParams = Parameters<typeof resolveBundledMetadataManifestRecord>[0];
 
-function readFacadeBoundaryConfigSafely(): {
-  rawConfig: OpenClawConfig;
-} {
+function readFacadeBoundaryConfigSafely(): OpenClawConfig {
   try {
     const sourceSnapshot = getRuntimeConfigSourceSnapshot();
     if (sourceSnapshot) {
-      return { rawConfig: sourceSnapshot };
+      return sourceSnapshot;
     }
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [247af5711edbd42bb525c5ef369b629ec194ba88](https://github.com/openclaw/openclaw/pull/140489/commits/247af5711edbd42bb525c5ef369b629ec194ba88)
- Landed commit: [863fab17f42dd473d5fc1a1f7522c6e927712878](https://github.com/openclaw/openclaw/commit/863fab17f42dd473d5fc1a1f7522c6e927712878)

PR #140494, opened 2026-09-06
fix(agents): xAI OAuth auto model cannot be materialized

### LINKED ISSUE #140466: [Bug]: xAI OAuth auto alias resolves to its canonical model, then fails runtime auth rematerialization
### Bug type

### Beta release blocker

No

### Summary

On OpenClaw 2026.9.2, the xAI OAuth `auto` model alias normalizes to its catalog-provided canonical model (`grok-4.6`), then fails prepared-auth model materialization because the resolved model ID differs from the selected alias.

### Steps to reproduce

The following credential-free boundary reproduction was run against the installed 2026.9.2 distribution. It calls the actual provider normalization and prepared-runtime materialization functions, supplying a representative OAuth catalog row at the model-resolution callback boundary. It does not exercise login, network discovery, or the full Gateway by itself.

1. Save this as `repro-xai-auto.mjs`.
2. Run `node repro-xai-auto.mjs /path/to/openclaw/dist`, pointing to an installed 2026.9.2 distribution. The bundled filenames and export names below are specific to that version.
3. Observe exit code 1 and the exact materialization error shown below.

```javascript
// Credential-free boundary reproduction for OpenClaw 2026.9.2.
// Run: node repro-xai-auto.mjs /path/to/openclaw/dist
// These bundled export names and filenames are specific to the tested version.
import path from 'node:path';
import { pathToFileURL } from 'node:url';

const dist = process.argv[2];
if (!dist) throw new Error('Pass the installed OpenClaw dist directory');
const { n: normalizeXaiResolvedModel } = await import(pathToFileURL(path.join(dist, 'provider-models-D2CRHpq1.js')));
const { t: materializePreparedRuntimeModel } = await import(pathToFileURL(path.join(dist, 'materialize-model-DTezazoG.js')));
const model = {
  provider: 'xai', id: 'auto', api: 'openai-responses',
  baseUrl: 'https://cli-chat-proxy.grok.com/v1',
  params: { canonicalModelId: 'grok-4.6' }, reasoning: true,
  input: ['text'], cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
  contextWindow: 500000, maxTokens: 64000,
};
const normalized = normalizeXaiResolvedModel(model);
console.log(JSON.stringify({ requested: model.id, normalized: normalized.id }));
try {
  await materializePreparedRuntimeModel({
    provider: 'xai', modelId: 'auto', model, config: {},
    plan: { selectedAuthMode: 'oauth' }, forceResolve: true,
    resolveModel: async () => ({ model: normalized }),
  });
  console.log('PASS: resolved model accepted');
} catch (error) {
  console.log('FAIL: ' + error.message);
  process.exitCode = 1;
}

```

### Expected behavior

The OAuth catalog's generated `auto` entry should be usable as its canonical target. The installed `withXaiOAuthAutoModel` explicitly generates that entry with `params.canonicalModelId`; `normalizeXaiResolvedModel` explicitly resolves it. In the observed Gateway comparison, selecting that canonical target (`xai/grok-4.6`) returned `OPENCLAW_OK` without fallback.

### Actual behavior

The fixture normalizes the model successfully, then the runtime materialization guard rejects it:

```text
{"requested":"auto","normalized":"grok-4.6"}
FAIL: Unable to rematerialize xai/auto for its resolved auth profile.
```

The same error was observed before inference in an actual Gateway agent turn selecting `xai/auto` with the OAuth catalog present and `--thinking low`.

### OpenClaw version

2026.9.2 (`3928bad`)

### Operating system

Debian GNU/Linux 13, amd64; Node 24.15.0

### Install method

Shell installer resulting in an npm installation under nvm.

### Model

Selected: `xai/auto`; catalog canonical target: `xai/grok-4.6`.

### Provider / routing chain

CLI -> OpenClaw Gateway -> xAI provider using the existing OAuth profile and `openai-responses` at `https://cli-chat-proxy.grok.com/v1`. The failing turn stops during runtime model materialization, before inference.

OpenClaw's installed `buildLiveXaiOAuthProvider` successfully returned a subscription catalog containing `auto`, `grok-4.6`, and `grok-4.5`. The `auto` entry had `params.canonicalModelId: grok-4.6`. Its non-secret provider metadata was saved as the configured xAI provider, with `auth: oauth` and the subscription endpoint above. Authentication remained in the existing OAuth profile; no credentials are included in this report or fixture.

The Gateway comparison used a diagnostic message requesting exactly `OPENCLAW_OK`, without tools or file changes. Selecting `xai/auto` with `--thinking low` failed; selecting `xai/grok-4.6` with the same thinking level succeeded.

### Logs

Failing Gateway CLI result, identifiers omitted:

```json
{
  "requestedModel": "xai/auto",
  "exitCode": 1,
  "status": false,
  "error": {
    "type": "cli_error",
    "message": "Unable to rematerialize xai/auto for its resolved auth profile."
  },
  "reply": []
}
```

Successful canonical-model comparison:

```json
{
  "requestedModel": "xai/grok-4.6",
  "exitCode": 0,
  "status": "ok",
  "reply": ["OPENCLAW_OK"],
  "effectiveProvider": "xai",
  "effectiveModel": "grok-4.6",
  "harness": "openclaw",
  "fallbackUsed": false
}
```

Relevant source regions, as identified in the installed distribution:

- `extensions/xai/provider-catalog.ts`: `withXaiOAuthAutoModel` generates the alias and its canonical-model parameter.
- `extensions/xai/provider-models.ts`: `normalizeXaiResolvedModel` changes the model ID to that canonical target.
- `src/agents/runtime-plan/materialize-model.ts`: `materializePreparedRuntimeModel` rejects the resolved ID when it differs from the prepared target after catalog canonicalization.
- `src/agents/provider-model-route.ts`: `canonicalizeProviderModelId` delegates catalog equivalence to provider route metadata; in the reproduction it does not equate `auto` with its parameter-dependent target.

The fixture reproduces that boundary mismatch without mocks of either normalization or the materialization guard. Only the model-resolution callback is supplied directly. No installed source was patched.

### Impact and severity

Affected: one observed 2026.9.2 installation using the xAI OAuth alias. Severity: blocks replies for the selected alias. Frequency: the same failure occurred in the observed Gateway turn and in two runs of the focused fixture; fresh-install prevalence is unknown. Consequence: selecting the canonical model is required to continue chatting.

### Additional information

Verified workaround: select `xai/grok-4.6` directly. A local menu alias pointing to that canonical model also returned `OPENCLAW_OK` without fallback.

A regression test should cross the provider-normalization and prepared-auth materialization boundary. No known-good version was tested, so this report does not classify the issue as a regression. A separate earlier API-catalog routing condition was corrected before the reported Gateway comparison and is not the defect claimed here.

Related, but a different boundary: #122872 covers missing live canonical-model metadata in the isolated `--local` runner. This report reproduces rejection after `params.canonicalModelId` is already present and normalization has succeeded; the observed agent turn used the Gateway, not `--local`.

ayaangazali:
## What Problem This Solves

Fixes #140466.

Selecting the xAI OAuth `auto` model fails prepared-auth materialization:

```
Unable to rematerialize xai/auto for its resolved auth profile.
```

The alias is generated by the provider and resolved by the provider. `withXaiOAuthAutoModel` emits the `auto` catalog entry carrying `params.canonicalModelId`, and `normalizeXaiResolvedModel` resolves it:

`extensions/xai/model-id.ts:16-27`

```ts
const canonicalModelId = params?.canonicalModelId;
return typeof canonicalModelId === "string" && canonicalModelId.trim()
  ? canonicalModelId.trim()
  : id;
```

So resolution correctly returns `grok-4.6` while the requested id is still `auto`. Materialization then compared those two ids directly and rejected the result:

`src/agents/runtime-plan/materialize-model.ts`

```ts
canonicalizeProviderModelId(params.provider, resolved.model.id ?? "") !==
  canonicalizeProviderModelId(params.provider, params.modelId) ||
```

`canonicalizeProviderModelId` cannot reconcile them. It resolves through the provider's `normalizeModelCatalogId` surface, xAI does not implement that hook, and it could not help here anyway: the canonical target comes from the runtime catalog row rather than a static alias table, so `auto` canonicalizes to `auto` and `grok-4.6` to `grok-4.6`.

## Why This Change Was Made

The check is right to exist; it just had no way to tell a provider's own alias resolution from a wrong model. So it now asks the owning provider, through the `normalizeResolvedModel` hook the provider already implements, whether the resolved id came from the requested one.

The alternatives both looked worse. Teaching core that xAI's `auto` means `grok-4.6` puts provider policy in core, which the architecture rules exclude. Having xAI stop rewriting the id would break the alias resolution that makes the catalog entry usable in the first place.

Callers legitimately pass the alias. `credential-scoped-model.ts` and `simple-completion-runtime.ts` hand over the selected model id, and neither can pre-resolve it, because the canonical target only exists on the catalog row that resolution returns.

The guard is not loosened. The alias path is consulted only when the ids already differ, and it accepts only when the provider maps the requested id onto exactly the resolved id.

## User Impact

An xAI OAuth user who selects the `auto` model can use it. Before this, that catalog entry was unusable, and the reporter's comparison showed selecting the canonical target directly worked while the alias did not.

## Evidence

Production change is +25/-2 in one file.

The regression test fails on unfixed code with the reported error. Reverting only `materialize-model.ts`:

```
× accepts a catalog alias the owning provider resolved to its canonical model
  → Unable to rematerialize xai/auto for its resolved auth profile.
  Tests  1 failed | 13 passed (14)
```

The second test, which asserts a resolved model the provider did not derive is still refused, passes both before and after, so it demonstrates the guard was not weakened rather than merely following the change.

```
node scripts/run-vitest.mjs src/agents/runtime-plan       11 files, 146 passed
node scripts/run-vitest.mjs extensions/xai                33 files, 562 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                          clean
node scripts/run-oxlint.mjs <both touched files>                         clean
./node_modules/.bin/oxfmt --check <both touched files>                   clean
```

Proof ceiling, stated plainly: I have no xAI OAuth credentials, so I did not drive a live xAI session. The evidence is at the materialization boundary the issue names, using the real provider hook rather than a stub, which I verified fires and returns `grok-4.6` for the `auto` row before writing the test. That is the same credential-free boundary level as the reproduction in the issue.

### DIFF
--- a/src/agents/runtime-plan/materialize-model.test.ts
+++ b/src/agents/runtime-plan/materialize-model.test.ts
@@ -410,3 +410,52 @@ describe("materializePreparedRuntimeModel", () => {
     ).rejects.toThrow(/prepared subscription route/u);
   });
 });
+
+describe("provider-resolved model aliases", () => {
+  it("accepts a catalog alias the owning provider resolved to its canonical model", async () => {
+    // xAI's OAuth catalog generates an `auto` entry carrying its canonical target, and the
+    // provider resolves it during normalization. Before this was tolerated, materialization
+    // compared the resolved id against the requested alias and refused its own resolution.
+    const aliasPlan: AgentRuntimeAuthPlan = {
+      providerForAuth: "xai",
+      authProfileProviderForAuth: "xai",
+      forwardedAuthProfileId: "xai:oauth",
+      selectedAuthMode: "token",
+    };
+    const canonical = {
+      provider: "xai",
+      id: "grok-4.6",
+      params: { canonicalModelId: "grok-4.6" },
+    };
+
+    const materialized = await materializePreparedRuntimeModel({
+      plan: aliasPlan,
+      provider: "xai",
+      modelId: "auto",
+      forceResolve: true,
+      resolveModel: async () => ({ model: canonical as never }),
+    });
+
+    expect(materialized?.id).toBe("grok-4.6");
+  });
+
+  it("still refuses a resolved model the provider did not derive from the request", async () => {
+    const aliasPlan: AgentRuntimeAuthPlan = {
+      providerForAuth: "xai",
+      authProfileProviderForAuth: "xai",
+      forwardedAuthProfileId: "xai:oauth",
+      selectedAuthMode: "token",
+    };
+    const unrelated = { provider: "xai", id: "grok-4.6", params: {} };
+
+    await expect(
+      materializePreparedRuntimeModel({
+        plan: aliasPlan,
+        provider: "xai",
+        modelId: "grok-4.5",
+        forceResolve: true,
+        resolveModel: async () => ({ model: unrelated as never }),
+      }),
+    ).rejects.toThrow(/Unable to rematerialize/);
+  });
+});
--- a/src/agents/runtime-plan/materialize-model.ts
+++ b/src/agents/runtime-plan/materialize-model.ts
@@ -1,6 +1,7 @@
 import { normalizeProviderId } from "@openclaw/model-catalog-core/provider-id";
 import type { OpenClawConfig } from "../../config/types.openclaw.js";
 import type { PluginMetadataSnapshot } from "../../plugins/plugin-metadata-snapshot.types.js";
+import { normalizeProviderResolvedModelWithPlugin } from "../../plugins/provider-runtime.js";
 import { FailoverError } from "../failover/error.js";
 import { resolveBuiltInModelSuppressionFromManifest } from "../model-suppression.js";
 import {
@@ -135,11 +136,33 @@ export async function materializePreparedRuntimeModel<Model extends RuntimeRoute
         })
       : resolveProviderModelMaterializationAuthMode(params.plan.selectedAuthMode),
   });
+  // A provider may legitimately resolve a catalog alias to its canonical model, so the
+  // requested id and the resolved id differ for a correct resolution. Ask the owning
+  // provider whether it produced this id from the requested one rather than encoding
+  // any provider's alias table here. A genuine mismatch still fails: normalizing a
+  // different requested id does not yield the resolved id.
+  const resolvedFromRequestedAlias =
+    resolved.model &&
+    canonicalizeProviderModelId(params.provider, resolved.model.id ?? "") !==
+      canonicalizeProviderModelId(params.provider, params.modelId)
+      ? normalizeProviderResolvedModelWithPlugin({
+          provider: params.provider,
+          modelId: params.modelId,
+          ...(params.config ? { config: params.config } : {}),
+          ...(params.workspaceDir ? { workspaceDir: params.workspaceDir } : {}),
+          context: {
+            provider: params.provider,
+            modelId: params.modelId,
+            model: { ...resolved.model, id: params.modelId } as never,
+          },
+        })?.id === resolved.model.id
+      : false;
   if (
     !resolved.model ||
     normalizeProviderId(resolved.model.provider ?? "") !== normalizeProviderId(params.provider) ||
-    canonicalizeProviderModelId(params.provider, resolved.model.id ?? "") !==
-      canonicalizeProviderModelId(params.provider, params.modelId) ||
+    (!resolvedFromRequestedAlias &&
+      canonicalizeProviderModelId(params.provider, resolved.model.id ?? "") !==
+        canonicalizeProviderModelId(params.provider, params.modelId)) ||
     (route &&
       !modelMatchesPreparedTarget({
         model: resolved.model,


ayaangazali:
Closing this. My approach is wrong at this layer, and I would rather say so than push a cast to make CI green.

CI caught two real problems, both mine. The test typecheck failed on a `never` cast I used in the new test, and the assertion-safety ratchet flagged an uncommented assertion in the source. Removing the test casts was correct and fixed the first. The source one is where it fell apart.

The fix asks the owning provider whether it derived the resolved id from the requested one, by handing its `normalizeResolvedModel` hook a probe model built from the resolved model with the id swapped back to the alias. That probe cannot be typed honestly. `materializePreparedRuntimeModel` works with `RuntimeRouteModel`, the hook takes `ProviderRuntimeModel`, and the compiler is explicit that the two do not sufficiently overlap. Forcing it through `as unknown as` would compile, but it would hand every provider's hook a model missing fields that type promises, and other providers read those fields. That is a correctness problem, not a lint preference, so I am not going to paper over it.

The diagnosis in the PR body still holds and I have posted it on #140466: xAI generates the OAuth `auto` entry carrying `params.canonicalModelId`, `normalizeXaiResolvedModel` resolves it to `grok-4.6`, and materialization then rejects the provider's own resolution because it compares the resolved id against the requested alias. `canonicalizeProviderModelId` cannot bridge them, since the canonical target lives on the runtime catalog row rather than in a static alias table.

What I got wrong was the layer, not the diagnosis. Whoever picks this up should probably decide whether the alias resolution belongs earlier, so the requested id is already canonical by the time materialization compares it, rather than teaching the comparison to tolerate a mismatch after the fact.

Sorry for the noise on the queue.

PR #140496, opened 2026-09-06
fix(matrix): run fixture loaders natively under Bun

steipete:
## What Problem This Solves

Fixes two Matrix fixtures using Node-only loader hooks when their child runs on Bun. Those fixtures failed before reaching the import-laziness and released-state migration assertions they are meant to protect.

## Why This Change Was Made

Keep Node's existing hook path and use native Bun instrumentation for the exact canonical fixture owners. The import observer supplies unchanged source bytes to Bun's native TypeScript compiler. Its local source-import launcher uses the existing root tsconfig so it observes the intended source dependency graph; the shared public-artifact smoke helper is unchanged.

The Doctor fixture preloads the same two intentional stubs and now verifies that each actually evaluated once. Migration, archive-byte integrity, cursor persistence, normalized output, and positive/negative import assertions remain intact. No production code, dependency, or shared resolver changes are included.

## User Impact

The fixtures can check the same Matrix contracts on Node and Bun. The Doctor case uses owned temporary HOME/config/state/log directories and a checksum-verified released synthetic SQLite fixture; no operator state, real providers, credentials, or services are used.

## Evidence

- Both exact cases passed on Node and failed on Bun before the repair, then passed on both runtimes afterward. Other cases were not selected locally.
- An external inert comparison verified unchanged native module values, relative imports, identity, and single evaluation. Source-graph inspection positively observed the SDK source owner without importing secret setup.
- Bun's separate tsconfig directory-mismatch stderr diagnostic is retained as a runtime limitation and is not claimed repaired. Existing expected output and exit checks remain unchanged.
- Fresh independent P0–P2 review passed with no actionable findings. The complete changed-file gate passed, including extension-test types, repository guards, dead-export checks, and scoped lint; all processes joined without a resource cap.

Only two test files change. Required hosted review and exact-head CI must pass before landing; this is not full Matrix, live-service, or all-platform coverage.

### DIFF
--- a/extensions/matrix/doctor-contract-api.account-state.test.ts
+++ b/extensions/matrix/doctor-contract-api.account-state.test.ts
@@ -67,40 +67,56 @@ function runMatrixDoctorFix(params: { rootDir: string; stateDir: string }) {
   );
   fs.writeFileSync(
     loaderPath,
-    `import { registerHooks } from "node:module";
-registerHooks({
-  resolve(specifier, context, nextResolve) {
-    if (specifier.endsWith("/doctor-ui.js")) {
-      return {
-        shortCircuit: true,
-        url: "data:text/javascript," + encodeURIComponent([
-          "export async function detectUiProtocolFreshnessIssues() { return []; }",
-          "export function uiProtocolFreshnessIssueToHealthFinding() { return {}; }",
-          "export function uiProtocolFreshnessIssueToRepairEffects() { return []; }",
-          "export async function maybeRepairUiProtocolFreshness() {}",
-        ].join("\\n")),
-      };
-    }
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1708b62a4b6c20255c62519b84aebbd429a1614d](https://github.com/openclaw/openclaw/pull/140496/commits/1708b62a4b6c20255c62519b84aebbd429a1614d)
- Landed commit: [30a6c4a4a70958faf20891c475b2c968ebbbac66](https://github.com/openclaw/openclaw/commit/30a6c4a4a70958faf20891c475b2c968ebbbac66)

PR #139541, opened 2026-09-06
fix(memory): split embedding batches rejected for an input array item cap

ylcn91:
Fixes https://github.com/openclaw/openclaw/issues/139040

## What Problem This Solves

Memory indexing stops when an embedding provider rejects an oversized request
with `input array max 64`, instead of splitting it into smaller batches.

## User Impact

Oversized requests can recover through the existing batch-splitting path while
per-item validation errors and token limits remain terminal. Vector ordering and
existing provider retry behavior are preserved. No configuration, schema,
persistent-storage or dependency change is required.

## Why This Change Was Made

The reported item-count rejection needs its own precise classification. The new
alternative requires a complete integer and message boundary; generic error code
`1214`, item/token limits, token suffixes and fractional counts do not qualify.
Existing Ark, Qianfan, DashScope and transport alternatives remain unchanged.

The contributor's work and both commits are preserved. The maintainer integration
updates the existing transport harness for current Memory Core candidate and
generation APIs, narrows the matcher, and extends the existing test tables.
Original implementation and HTTP-boundary coverage by @ylcn91.

## Evidence

Current candidate: `21e1e78f9a2f1c9a0ddb5beaf6d7ab91b0ddcaca`, integrating
`201f11afefdb19a9c75e9d3542bf7425c80df7a7` without rewriting contributor ancestry.
Independent source review found no actionable P0-P2 issues; diff checks passed.

Isolated AWS proof on this exact candidate passed:
https://crabbox.openclaw.ai/portal/runs/run_7f17ed57d156

- The original broad-matcher counterfactual produced 14 intended failures and
  7 passing controls, with 18 tests deliberately unselected. Four terminal HTTP
  cases made eight requests instead of one; ten retry cases called twice instead
  of once, including the numbered-element token-limit wording.
- Restored candidate: 84 owner tests passed, including ordered HTTP requests
  `[100, 50, 50]`, ordered vectors and one-request terminal errors.
- Relevant embedding, bridge and cache siblings: 26 passed. Both green phases
  had zero failures or skips.
- Formatting and selected changed checks passed, including production/test
  types, lint, dead-export and applicable boundary guards. Candidate hashes were
  unchanged, all process trees joined and no evidence was omitted.

Exact-head CI completed green:
https://github.com/openclaw/openclaw/actions/runs/34775024167
Merged through native squash admission as
`0afc02335308892d7efa3ccf7d981e912c6dd238`.

[...]

Issue #139040, opened 2026-09-05
[Bug]: [memory] sqlite-backed corpus chunks are not split to the 64-item embedding limit (HTTP 400, Zhipu code 1214)


DagenGuo:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

Environment
OpenClaw 2026.9.1 (ad6fe23), Windows 10, npm install
Memory search plugin (memory-core), remote embedding provider: openai-compatible / model embedding-3 (Zhipu BigModel API)
Reproduced on agent office (3/3 attempts); NOT reproducible on file-backed corpora
Summary
When the memory index rebuild embeds chunks that come from a sqlite-backed corpus (session store), all chunks are sent to the remote embedding API in a single request without splitting. Zhipu's embeddings endpoint rejects inputs larger than 64 items (HTTP 400, code 1214).

File-backed corpora split correctly: a 150-chunk markdown file was indexed successfully on agent main (proves the split path works for file sources). Agent office has its session corpus in sqlite (no disk files), so its 1,180 chunks are sent unsplit and always fail.

Steps to reproduce
Agent whose session corpus is sqlite-backed (e.g. office after --fix reset) with >64 unindexed chunks.
Run: openclaw memory status --index --agent office
Observe: remote embedding request fails with HTTP 400 / code 1214 ("input array max 64").
Expected
Chunk splitting (splitBatchRequestsByLimits / shouldSplitGroupOnError) should apply to sqlite-backed sources exactly as it does for file-backed sources.
Impact
Agents with sqlite-backed corpora cannot build the vector index at all with Zhipu embeddings (file-backed agents are unaffected).
Vector search unavailable for those agents until upstream fix; FTS still works.
Workaround: local embedding (ollama / llama.cpp) or a provider with a higher per-request limit (e.g. OpenAI text-embedding-3, 2048 items).
Related (same area, lower severity): root MEMORY.md is never indexed
memory_index_sources contains USER.md but never the workspace root MEMORY.md (0 chunks, 0 rows), even after touch + --index retry.
Reproduced on both agents (main and office), so it is a general scan omission, not environment-specific.
Low impact: MEMORY.md is still read directly and injected via bootstrap separately, so retrieval still works; only semantic search over its index lines is lost.
isMemoryPath in internal-CKQ1RDPZ.js explicitly lists "MEMORY.md" as a memory path, so the scan omission looks unintentional.

### Steps to reproduce

[...]

PR #138992, opened 2026-09-05
fix(memory): embeddings fail when providers reject oversized batches

saladinyao-cyber:
## What Problem This Solves

Fixes an issue where users of OpenAI-compatible embedding providers could not index memory when a provider rejected an oversized item batch with an explicit numeric maximum. The request failed permanently instead of being split into smaller batches.

## Why This Change Was Made

Recognize the provider-neutral error form `batch size is invalid, it should not be larger than <N>` and route it through the existing recursive batch splitter. The classifier requires an explicit numeric maximum and does not depend on a provider name.

## User Impact

Memory indexing now automatically recovers from this explicit provider batch-size limit instead of failing with HTTP 400. Other input-validation errors remain permanent and are not retried or split.

## Evidence

- Focused extension-memory Vitest: 24/24 passed.
- Verified recursive splitting of 33 inputs until every request is at or below the provider limit.
- Negative cases remain unsplittable: unknown maximum, minimum-size wording, input-size wording, and generic input validation.
- Oxlint: 0 warnings, 0 errors.
- Oxfmt and `git diff --check` passed.

PR #140508, opened 2026-09-06
perf(memory): honor explicit embedding batch item limits

saladinyao-cyber:
## What Problem This Solves

After #139541 recognizes a provider's explicit input-array item-cap rejection and enters the existing split retry path, recovery still uses binary partitioning. For a request of 33 inputs against a provider that explicitly reports a cap of 10, binary fallback sends several avoidably rejected intermediate requests before reaching acceptable slices.

This PR keeps #139541 as the owner of error classification and generic binary recovery. It adds only a reactive optimization: after a real provider request fails with an already-recognized, reliably parseable item-cap error, the existing retry splitter uses that reported cap directly. The observed request sequence becomes `33, 10, 10, 10, 3`.

For existing Zhipu `1214` / `input array max 64` responses, the observable 100-input request sequence changes from binary `100, 50, 50` to explicit-cap `100, 64, 36`, reducing rejected requests while preserving output order and each successful slice’s `onSuccess` pairing.

## Scope and Complementarity

- **#139541 (merged):** owns precise classification of explicit input-array item-cap errors and the correctness-preserving binary fallback. This PR does not restore or duplicate superseded matcher behavior.
- **#141686 (open):** proposes proactive request batching from known/provider metadata before the first request. This PR does not add provider/model metadata, schema, SDK surface, conservative defaults, or proactive policy.
- **#140508 (this PR):** reacts only after a real provider response supplies one unambiguous, positive, safe integer cap smaller than the rejected request. Missing, ambiguous, invalid, oversized, equal-to-request, or stale caps continue through #139541's binary fallback. If a reported cap is stale and a capped slice is rejected again, recursive fallback remains active.

The two open PRs therefore optimize different phases and can compose: #141686 prevents known oversized requests; #140508 accelerates recovery when proactive metadata is absent, wrong, or cannot know the endpoint's real cap.

## Implementation

- Parse only item-cap phrasings already admitted by the current splittable-error classifier.
- Use the parsed cap for sequential retry slices while preserving input/output order.
- Keep each input object intact so its `cacheCandidate` remains paired with the corresponding output.
- Invoke the existing `onSuccess` persistence callback once for each successful slice, never for the rejected parent request.
- Report cap-aware diagnostics as the actual chunk count and maximum chunk size; binary diagnostics retain their prior `left + right` meaning.

No configuration, dependency, persistence schema, vector format, cache key, provider-selection contract, public API, or migration changes are introduced.

## Evidence

Rebased onto: `7c51ba75120ad3fd0e97fdc8da064f4fd15c392c`

Current upstream base ref observed after push: `210907758ea04d55b80420f63e2b5fa6ab59c5c3`

Current head: `13e53ceac3563371087405844fddedd84b965080`

### Real transport proof

A loopback HTTP embedding server was exercised through the real OpenAI-compatible transport and production Memory Core manager path:

- inputs: 33
- server cap: 10
- observed request sizes: `33, 10, 10, 10, 3`
- rejected requests: 1
- successful slices: 4
- returned vectors: original input order preserved

The committed unit proof also verifies that `cacheCandidate` values stay paired with outputs and that `onSuccess` persists each successful slice exactly once.

### #141686 composition proof

The current #141686 head `4f72f655d7156fa30c6bca08a30bbf1be3ef7565` was merged into a disposable detached worktree with this PR. The only textual conflict was the expected `manager-embedding-ops.ts` batching call, resolved by retaining current candidate pairing and passing #141686's proactive cap. Four selected project shards passed **121/121**, covering reactive retry, proactive batching, provider metadata, remote-provider passthrough, and native/custom OpenAI endpoint behavior.

### Tests and gates

- exact-head focused Memory Core extension shard: **86/86 passed**
- full Memory Core project: **1,852 passed, 4 platform skips, 5 failures** across 1,861 tests
- the same 5 failures reproduced unchanged on exact `origin/main` in a clean detached worktree: **91 passed, 5 failed** across the two affected files; they are locale/time-format and locale-sort expectations outside this diff
- extension production TypeScript: passed
- extension test TypeScript: passed
- full extensions Oxlint: **9,560 files, 0 warnings, 0 errors**
- Oxfmt changed-file check: passed
- max-lines ratchet: passed
- `git diff --check`: passed
- full `pnpm build`: passed

The final rebase was conflict-free over non-overlapping upstream changes; the binary diff digest and stable patch-id remained unchanged (`6556e01060b8175e4dda7751ab4580ec89a4c7ddae28937193b444d5466e9af8`, `3dba2adb16f966358b8e9fc5697c27d8c3ae79aa`).


saladinyao-cyber:
Addressed the review at `d6f0ed5fa1`: cap-aware diagnostics now describe the actual capped chunking, and a loopback HTTP proof exercises the real OpenAI-compatible client through the production manager retry path. For 33 inputs with a cap of 10, the observed request sizes are `33, 10, 10, 10, 3`; ordered vectors `0..32` are preserved. Focused tests pass 40/40, with production/test TypeScript checks, Oxlint, Oxfmt, and diff checks passing. The PR description now includes the full redacted evidence and compatibility statement.

@clawsweeper re-review

saladinyao-cyber:
Updated the PR context to use the required `What Problem This Solves` and `Evidence` headings. Also clarified data-model and upgrade compatibility: this changes only in-memory retry partitioning, with no schema, persisted record, vector/embedding metadata format, or migration change; existing indexes remain compatible.

@clawsweeper re-review

Yigtwxx (commented):
Reviewed head `d6f0ed5fa10` against `main` `d3101b5229b`. This is the follow-through on the splitter note from #138992: the classifier already required the number and the splitter ignored it. The change is small and lands where it should, in `runMemoryEmbeddingBatchRetryWithSplit`, with the binary path kept as the fallback for absent, ambiguous, non-positive or stale limits. Three things measured, then the two CI reds, which the ClawSweeper verdict does not mention.

**Measured.** (1) The three touched test files pass on this head under the memory shard, 40/40. Mutation: making `parseMemoryEmbeddingBatchItemLimit` return `undefined` turns exactly 10 tests red, the three new policy cases, the six retry-boundary expectations (text and structured) and the loopback integration case, with the other 30 still green, so the new coverage is carried by the limit path and not by the surrounding harness. (2) Request counts with the real policy functions over five consecutive 40-item batches against a 10-item cap, for all three message shapes the regex accepts: `main` 35 requests / 15 rejected, this head 25 / 5. None of the three messages matches `isRetryableMemoryEmbeddingError`, so there is no backoff before the split. (3) The stale-limit row is real: a provider that reports 3 but accepts 2 goes `4, 3, 2, 1, 1`, one wasted request rather than a loop.

**CI.** `check-dependencies` is a genuine red from this branch: knip `--production` reports `Unused exports (1): extensions/memory-core/src/memory/manager-embedding-policy.ts: parseMemoryEmbeddingBatchItemLimit`. The export exists only for `manager-embedding-policy.test.ts`; the repository's dead-export ratchet counts test-only exports, so it needs to become module-private, with the parse table folded into cases that drive `runMemoryEmbeddingBatchRetryWithSplit` (the "uses an explicit item limit" and "falls back" cases already cover every row of the parse test except the two-number-ambiguity string, which fits in the fallback table). The `PR context and evidence` red is stale rather than yours: it ran at 23:53Z, and the current body passes `evaluatePullRequestContext` from `scripts/github/real-behavior-proof-policy.mjs` locally (`status: passed`), so it clears on the next push. The bot's "add data-model compatibility proof" checkbox has nothing to attach to: the diff touches the policy, the ops log line and tests, no store or schema file.

One note for the description, not a blocker. The limit is not retained between calls, and `buildMemoryEmbeddingBatches` sizes batches by an 8000-byte budget with no item cap, so on a provider with a 10-item cap every batch above ten items still pays one rejected request; the five-batch run above is 5 rejections where a limit remembered for the life of the provider instance would be 1. That is a separate change with its own cache-invalidation questions and I would not fold it into this PR, but the body's "rejected requests: 1" describes one batch, and saying so keeps the claim exact for an index run of many batches.

Yigtwxx (commented):
Re-measured head `425603d7f23` after the force-push, against the same memory shard as before.

- `parseMemoryEmbeddingBatchItemLimit` is module-private now and the direct-parse cases were folded into the two `runMemoryEmbeddingBatchRetryWithSplit` cells (explicit limit with order preserved; unusable, ambiguous, oversized and stale limits falling back to binary). `check-dependencies` is green on this head, so the knip `--production` red from `d6f0ed5fa10` is gone. All 49 checks pass, nothing red.
- The three touched test files pass locally, 39/39.
- Mutation: making the parser return `undefined` turns exactly 9 tests red (the two new policy cells, the six retry-boundary expectations for text and structured errors, and the loopback integration case) with the other 30 still green. Coverage of the limit path survived the fold.

Nothing further from my side.

saladinyao-cyber:
Rebased onto current `main` (`b8e2dbc7ca4237c18a82b274562ba6447d0d88a0`); new head is `961c2543d32fc230898969ae562fa382489b6a09`.

The rebase preserves upstream's new retry-profile/cooldown flow and adds only the item-limit-aware partitioning behavior; it does not restore the removed retry APIs. The parser remains module-private and is exercised through `runMemoryEmbeddingBatchRetryWithSplit`.

Post-rebase verification:

- focused policy, retry-boundary, and real loopback HTTP tests: 69/69 passed in a proxy-cleared environment;
- observed capped request sequence remains `33, 10, 10, 10, 3`, with ordered outputs preserved;
- Oxfmt passed on all 5 changed TypeScript files;
- `git diff --check origin/main...HEAD` passed;
- final patch digest stayed byte-identical across the last non-overlapping `main` advance.

There is still no schema, persisted record, vector metadata, configuration, or migration change; the old data-model checkbox is not applicable to this in-memory retry policy.

@clawsweeper re-review

Yigtwxx (commented):
Head `961c2543d32` is based on `b8e2dbc7ca4`, which predates `cd9be2bbe7d` (#144404, merged 00:52:53Z). That commit rewrote the function this PR changes: `runMemoryEmbeddingBatchRetryWithSplit` gained `onSuccess`, which persists every successful sub-batch, and `embedBatchWithRetry` now carries a cache candidate next to each input. Every check run on this head started between 00:19:16Z and 00:26:48Z, so the 48 green checks were produced 26 minutes before that landing and never covered the combination. GitHub reports the merge as clean and it is textually clean, but both changes edit the same function, so I merged it and measured instead of reading it.

**Merged locally.** `origin/main` `1b2b06ed2fc` into this head, no conflicts. In the merged function the item-limit branch returns from inside the `catch` before the parent's `onSuccess`, exactly as the binary branch does, and the children inherit `onSuccess` through `{...params}`, so each chunk persists itself once and none persists twice. The candidate-to-embedding pairing survives the new chunk boundaries for free: main carries the candidate on the item (`requestItems`) rather than indexing a parallel array, so `slice(start, start + itemLimit)` moves the pair together and there is nothing to re-index at a cap boundary.

**Tests on the merged tree.** The three files this PR touches: 69/69. The five files #144404 touched (`manager-embedding-cache`, `memory-forget`, `manager-chunk-writer`, `manager-memory-source-race`, `manager-db`): 57/57.

**Mutation on the merged tree.** Making `parseMemoryEmbeddingBatchItemLimit` return `undefined` turns exactly 9 tests red - the two policy cells, the six retry-boundary expectations across text and structured, and the loopback integration case - with 68 green and `manager-embedding-cache.test.ts` untouched. The limit-path coverage still binds after the rebase and does not overlap the new cache tests.

**One measurement the rebase makes newly relevant.** Because #144404 banks each successful sub-batch, split shape now decides how much completed work survives a batch that later dies on a permanent per-item error. Merged policy against main's policy, 33 items, provider cap 10, counting items passed to `onSuccess`:

| scenario | main | this head |
| --- | --- | --- |
| clean run | 7 requests, 3 rejected, 33 persisted | 5 requests, 1 rejected, 33 persisted |
| one permanently rejected input, swept over all 33 positions | 163 requests, 408 persisted | 105 requests, 390 persisted |

The persisted difference is a chunk-boundary artifact rather than a behavior change: chunks are `10, 10, 10, 3` here against `9, 8, 8, 8` on main, both arms stop at the first chunk carrying the bad input and neither salvages that chunk's siblings, so this head banks more in 15 of the 33 positions and less in 9. Worth stating because the number moves; nothing to change for it.

Nothing outstanding from my side. The only thing I would ask for before merge is a refresh onto a main that contains `cd9be2bbe7d`, so the green checks cover the combination I had to produce locally.

saladinyao-cyber:
Refreshed onto current `main` at `d9dd26fb6c5ecbad864a7d96ad50d63629d4b358`; new head is `6d2533855d5e0e47af5e1aa38cd3aba549d4fa08`.

Verified the combined #144404 behavior: cap slices are `10, 10, 10, 3`; each successful slice invokes `onSuccess` exactly once, and cache candidates remain paired with their outputs through every slice. The 3 PR test files plus the 5 reviewer-listed #144404 test files pass (8 files, 126 tests), as do extension production/test typechecks, full extension lint, Knip production dead-export checks, Oxfmt on all 5 changed files, and `git diff --check`.

@clawsweeper re-review

saladinyao-cyber:
Refreshed again onto current `main` at `7764767a88b3a03850e4aee5f1ccea26b91c9886`; new head is `db0788348edc3c21d7bd4755463efdc5ec17ce27`.

The prior hosted failure was the unrelated `manager-temporal-ranking` FTS-unavailable fixture (2422 tests passed before that single failure). It reproduced on the prior base, and upstream has since fixed that fixture in #144761 (`1bf78261b73`). On this refreshed head the previously failing temporal-ranking case passes. The reviewer-requested embedding/cache combination remains covered; the loopback transport cases pass in the required proxy-cleared environment. Extension production/test typechecks, full extension lint, Knip production checks, Oxfmt, and `git diff --check` pass.

@clawsweeper re-review

Yigtwxx (commented):
The refresh I asked for is in place. Head `db0788348ed` is parented on `7764767a88b`, which contains both `cd9be2bbe7d` (#144404) and `1bf78261b73` (#144761), so the 48 green checks on this head were produced against the combination I previously had to merge locally. Nothing left to reproduce on my side.

The PR-surface delta since `961c2543d32` is the combination itself, not a new mechanism: `manager-embedding-policy.ts` threads `onSuccess` and the per-item `cacheCandidate` through the item-limit chunks, and the new policy case asserts the shape I measured in the last review - payloads `[33, 10, 10, 10, 3]`, `onSuccess` invoked once per chunk with the candidates of that chunk in order, and the flattened candidates equal to the input order. That is the pairing invariant the rebase made load-bearing, and it is now pinned by a test rather than by my local run.

Nothing outstanding from me.

saladinyao-cyber:
Clarifying the automated data-model checklist: this PR does not change the SQLite schema, vector/embedding metadata format, cache keys, persistence ownership, or migration behavior. The introduced delta is limited to parsing an explicit provider item cap and partitioning the existing retry inputs; each successful slice still reaches the existing `onSuccess` callback with its original input/cache-candidate pairing and in-order outputs. Therefore no migration or backfill is required, and existing databases remain upgrade-compatible. This is covered by the current head tests and the loopback transport evidence already recorded in the PR body.

saladinyao-cyber:
@clawsweeper re-review The compatibility clarification immediately above confirms that this PR changes neither persisted schema nor migration behavior; please refresh the durable review against current head `db0788348edc3c21d7bd4755463efdc5ec17ce27`.

saladinyao-cyber:
@clawsweeper re-review

Rebased onto current `main` at push time and narrowed the PR description to the remaining reactive optimization after #139541. The updated evidence distinguishes retry-time cap parsing from #141686 proactive metadata batching and includes the combined 121/121 proof. Current head: `6cdb8ba162e8ac66483e75fd811316f38d515c77`.

Yigtwxx (commented):
Re-measured head `6cdb8ba162e` (parent `7c51ba75120`) because the surface moved since my last note on `db0788348ed`, not just the base. Two things changed that the body does not mention, and one of them removes a regression cell that `main` relies on.

**1. The item-cap parser gained a fourth phrasing, and it changes the observable Zhipu sequence.**

`db0788348ed` parsed three phrasings. `6cdb8ba162e` adds `input array max\s+(\d+)` (with the same lookahead the splittable classifier already uses). That is the Zhipu BigModel `1214` body from #139040, so a real Zhipu rejection now takes the item-limit branch instead of binary halving. I checked out this head and ran `main`'s unchanged `generic-embedding-provider.integration.test.ts` (the `1214` / `input array max 64` cell over the real transport, 100 inputs):

```text
FAIL  halves an oversized batch until the provider input array limit is met
AssertionError: expected [ 100, 64, 36 ] to deeply equal [ 100, 50, 50 ]
```

`100, 64, 36` is the better sequence (one rejected request instead of two), so this is an improvement, but it is a behavior change for the provider that motivated the fallback, and the body's "Parse only item-cap phrasings already admitted by the current splittable-error classifier" reads as if nothing observable moves for existing providers. Worth one sentence in the body.

**2. The #139040 regression cell was rewritten rather than updated.**

Instead of changing that cell's expectation to `[100, 64, 36]`, the PR replaces it: the constant is now a DashScope-style `InvalidParameter` / `batch size is invalid, it should not be larger than 10` body (still named `ZHIPU_REJECTION_BODY`), the limit drops to 10, and the input count to 33. After this PR:

- the real Zhipu `1214` body has no transport-level cell at all, so the new fourth alternative and its lookahead are only exercised by the classifier's terminal-rejection cells (`input array max 64 tokens`, `64tokens`, `64.5`, which correctly do not split) and never by a successful parse;
- binary halving has no transport-level cell either; the only remaining coverage is the unit-level `binary:4:2:2` expectation in `manager-embedding-policy.test.ts`.

The PR's own cells are green on this head (55/55 across the two memory-core files, Node 24.21.0), so this is about what got dropped, not what was added.

Suggestion, small: keep the DashScope cell, restore the Zhipu cell with `[100, 64, 36]` (it passes on this head; that is exactly what I observed), and rename the constant. If you want a halving cell over the real transport as well, a splittable message with no parseable number (`request header fields too large` is already in the classifier) would keep `100, 50, 50` alive without touching the parser.

None of this changes my read of the implementation: ordering, `onSuccess` pairing, and the stale-cap fallback still hold. It is a coverage and body-accuracy ask, not a correctness one.

saladinyao-cyber:
Addressed the latest transport-coverage review in `13e53ceac3563371087405844fddedd84b965080`:

- kept the DashScope cap cell and corrected its constant names (`33, 10, 10, 10, 3`)
- restored the real Zhipu `1214` / `input array max 64` transport cell (`100, 64, 36`)
- added a transport-level unparseable-cap binary fallback cell (`100, 50, 50`)

Evidence: the three related Memory Core test files pass **88/88**; extension production/test typechecks pass; changed-file Oxlint reports 0 warnings/errors; changed-file Oxfmt and `git diff --check` pass.

@clawsweeper re-review

Yigtwxx (commented):
Re-measured `13e53ceac35` against the three asks, with my own runs rather than the PR's numbers (Node 24.21.0):

- The three memory-core files pass **88/88** here as well.
- Mutation on this head: removing only the `input array max\s+(\d+)` alternative from `MEMORY_EMBEDDING_BATCH_ITEM_LIMIT_RE` turns exactly one cell red, the restored Zhipu one (`expected [ 100, 50, 50 ] to deeply equal [ 100, 64, 36 ]`); the DashScope cell and the new `request header fields too large` halving cell stay green. So the fourth phrasing is now proven through the real transport by its own cell, and the binary fallback has a transport-level cell again.
- The body now states the Zhipu sequence change (`100, 50, 50` → `100, 64, 36`) explicitly, and the constants are named for what they carry.

Nothing outstanding from me.

Issue #139428, opened 2026-09-05
Test audit: reduce redundant test code and full-suite runtime


steipete:
## Problem

The test suite has substantial maintenance and execution cost. At `342b7fc95009cca85425a38ccbd360b859ac0f0e`, tracked files containing `.test.` or `.spec.` account for 5,552,085 physical lines across 13,268 files. The standard profiling run expanded to 649 Vitest shards. An earlier diagnostic planned 424 because forwarded reporter/output options disable additional bounded-extension splitting; it is not the default topology.

This maintainer-requested campaign will remove redundant test code, improve retained tests, and reduce unnecessary execution work in independently reviewed batches. The targets are 100,000 net lines removed and 30% lower full local test runtime, subject to preserving meaningful behavioral coverage.

## Acceptance

- Count added test support against deleted test code; moving code or minifying it does not count as removal.
- For each removed scenario, identify its actual regression protection and stronger remaining proof, or show that it cannot detect a production regression.
- Preserve independent API, configuration, migration, lifecycle, security, platform, and release contracts.
- Compare repeated full-suite baseline/candidate runs under matched environment, worker, cache, and preflight conditions. A cold diagnostic run or selected fast subset is not evidence of the runtime target.
- Pin the same production-source inventory for coverage comparisons. Default unit coverage dynamically discovers sibling source files and excludes several owners, so its percentage alone is insufficient.
- Land coherent batches through the normal review and CI gates.

## Initial batches

Image-runtime suites repeat their mock lifecycle setup; share that setup while preserving all scenario bodies and provider/profile differences. Pure config validation tests can call the existing implementation owner without loading the plugin-validation reexport graph; repeated schema coercion cases can become named tables. Measured slow lifecycle scenarios will be investigated separately before changing clocks or fixtures.

## Current campaign status — September 11, 2026

The campaign ledger records **12,867 net test/test-support lines removed across 231 merged campaign PRs**. Added support and strengthened regression tests are deducted from removals. The targets of **100,000 net lines removed** and **30% lower overall test runtime** remain incomplete.

Recent merged batches include #144509 (50 test/support lines removed), #144505 (38 removed), #144570 (10 removed), #144513 (7 added), #144492 (46 added), #144623 (38 added), #144690 (3 added), #144775 (108 removed), and #144781 (50 removed). These are already included in the recorded total. Pending #144359 adds 133 test lines; it is not counted until its merge is verified.

[...]

PR #140518, opened 2026-09-06
refactor: simplify Doctor migration test fixtures

steipete:
Related: #139428

## What Problem This Solves

Doctor's SQLite migration tests repeat the same import options in 71 calls, adding noise when reviewing the independent migration, recovery, and archival scenarios. This is part of the maintainer-requested test cleanup.

## Why This Change Was Made

Share those identical calls through a local fixture helper that directly returns the real import operation's Promise. Each invocation still reads the current store environment and path and creates the same options object. Other selectors, modes, public CLI calls, and all assertions remain explicit and unchanged.

## User Impact

No runtime behavior changes. Removes 269 net test lines while retaining all 231 Doctor cases and their real SQLite and filesystem boundaries. No test-time improvement is claimed.

## Evidence

- Original owner suite: 231 cases passed.
- Revised owner and shared-store, reader, and transcript-reader siblings: 259 cases passed across four files.
- Full-file equivalence after expanding the 71 calls; matching before/after case inventories. A task-local diagnostic verified the actual helper's Promise identity, fresh options, live member reads, and absent optional fields.
- Formatting, full changed-file gate, and fresh P2 review passed.

### DIFF
--- a/src/commands/doctor-session-sqlite.test.ts
+++ b/src/commands/doctor-session-sqlite.test.ts
@@ -611,11 +611,7 @@ describe("runDoctorSessionSqlite", () => {
         });
       let imported;
       try {
-        imported = await runDoctorSessionSqlite({
-          env: store.env,
-          mode: "import",
-          store: store.storePath,
-        });
+        imported = await importLegacyStore(store);
       } finally {
         spy.mockRestore();
       }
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a362973873e7ee87319d6e6e5d97b40f974d398e](https://github.com/openclaw/openclaw/pull/140518/commits/a362973873e7ee87319d6e6e5d97b40f974d398e)
- Landed commit: [7b2b4e0cd555f49b08f51f734d22998702f6c0aa](https://github.com/openclaw/openclaw/commit/7b2b4e0cd555f49b08f51f734d22998702f6c0aa)

PR #140523, opened 2026-09-06
fix(android): keep completed runs settled after delayed history

steipete:
## What Problem This Solves

Fixes an issue where Android users could see a completed run’s Stop button or partial reply return after a delayed history refresh. With concurrent runs, reconciliation could instead clear another run that was still working. Late send acknowledgements and local transport failures could also restore or incorrectly retire pending presentation.

## Why This Change Was Made

The controller now retains completion facts while already-issued, applicable history requests can still publish them. Each read releases that ownership after publication, failure, cancellation, or supersession; later reads cannot keep an old completion alive indefinitely. Existing pending runs, unresolved replies, advertised run IDs, and the current lifecycle owner retain their own ownership.

Projection, retirement, and run-ID transfer now share the controller’s existing lock. History distinguishes an absent run from a reported run rejected because it has ended, so rejecting completed R cannot settle still-live S. The repair removes duplicate optimistic cleanup and stale projection forwarding. Local send failures no longer manufacture remote completion, and chat/voice failure text uses existing accurate localized messages.

Production grows by 56 lines to establish the missing request lifetime and atomic cleanup boundary; tests grow by 526 lines, with related concurrency cases sharing a fixture. Existing outbox persistence, recovery deadlines, stream/tool ownership, and steering contracts stay intact. The native source inventory removes two obsolete keys; generated locale cleanup belongs to the canonical publisher after landing.

## User Impact

Completed runs remain settled while earlier history and acknowledgements arrive. Another live run keeps its pending state and streamed reply. Ambiguous send outcomes retain the existing reconciliation behavior, and unnamed server activity retains its unknown presentation.

## Evidence

- The original 11-case selection on unchanged main `3d602cd4172f` produced seven intended failures and four passing controls. Follow-up concurrency regressions were reproduced separately on intermediate candidates, without counting them as additional original-main failures.
- The identical recovered-run/steering regression failed when its first delayed history response revived completed A. After the repair, that response leaves A settled; a subsequent fresh history response publishes the persisted reply and preserves a zero pending count.
- All 61 selected Android tests passed with zero failures, errors, or skips, including failure/cancellation, branch retry, concurrent-run, ACK, outbox, stream, and usage controls. The single canonical Android lint stage executed and passed all four module tasks. All 13 locally runnable classified checks and native source-inventory verification also passed on the same 23 bound inputs. The identical rendered BEFORE/AFTER also changes from a stale Working/Stop state to Ready/Start Talk. Independent review completed in one invocation with 18 automatic passes; its two findings and the source-based dispositions are described below. [CI run 34067783498](https://github.com/openclaw/openclaw/actions/runs/34067783498) passed on attempt 1: all six Android test/build/lint jobs executed successfully; the workflow finished with 84 successful and 14 skipped jobs. Its actual checkout was merge `aa28d1e1a219` of head `00c50fccef1e` into main `20d30ac7679c`; all eight introduced file hashes and the complete diff match the reviewed candidate.

Android execution uses the repository Gradle route in disposable Testbox with synthetic Gateway responses. The inspected before/after captures use the real Compose screen, identical driver and captured history bytes, and wait for the history operation to finish. BEFORE returns to pending count 1 with Working/Stop; AFTER remains at 0 with Ready/Start Talk and the same transcript. This is rendered Robolectric and controller/Room proof, not a physical-device, live-provider, microphone, or audio-capture claim.

<details>
<summary>Independent review and retained limits</summary>

The raw review reports two P2 concerns, both also investigated during earlier revisions. Neither has a demonstrated complete producer-to-client sequence, so neither was accepted as a reason to change this patch.

Full terminal session metadata can omit a new terminal-telemetry write. On the ordinary connected path, however, the serial event dispatcher records the earlier lifecycle/chat terminal first. A pre-disconnect history response fails the reconnect generation checks, and the direct history lookup excludes terminal-inactive runs. A separate embedded-recovery timing hypothesis remains unproved.

Lost-ACK history correlation precedes terminal filtering, but its predicates also require a matching persisted user message, exactly one pending/disconnected run with an unknown outcome and unresolved message, no unresolved reply for the target, and applicable branch/history state. Default steering can legitimately consume that input into the existing run. The proposed independently executing sibling sequence has not established those conditions; a terminal-ID guard is therefore not a proven contract-preserving repair. The broader input-to-run correlation question remains unresolved.

These are retained limits, not claims that every metadata or correlation path is universally correct. The reproduced stale-history, concurrent-run, late-ACK and read-lifetime cases are covered by the completed tests and rendered proof above.

</details>

Before: delayed history restores Working and Stop

After: the completed run stays Ready with Start Talk

Both attached PNGs were downloaded after publication and verified byte for byte against the inspected originals. ClawSweeper reported no actionable findings or required pre-merge changes; its own attachment retrieval and unavailable historical-blob limitations remain recorded in its review.


steipete:
Merged via squash.

- Prepared head SHA: [00c50fccef1e4c45ae8fab6c057c4104788a8743](https://github.com/openclaw/openclaw/pull/140523/commits/00c50fccef1e4c45ae8fab6c057c4104788a8743)
- Landed commit: [a963c9126e3c4a04a027ae65fbed39125e4d6e06](https://github.com/openclaw/openclaw/commit/a963c9126e3c4a04a027ae65fbed39125e4d6e06)

PR #140529, opened 2026-09-07
fix(gateway): avoid false 304 responses for quoted ETags

### LINKED ISSUE #140528: Quoted ETags containing commas can return false 304 responses
### Bug type

### Beta release blocker

No

### Summary

A valid, nonmatching quoted ETag containing commas and an asterisk can cause managed media and workspace images to return an empty `304 Not Modified` response.

### Steps to reproduce

1. Obtain a managed attachment through the normal chat and `artifacts.download` flow. An unconditional authenticated GET returns its bytes and hash-based ETag.
2. Request the same representation with `If-None-Match: "client,*,tag"`.
3. Repeat with `Range: bytes=2-5`, or HEAD. Workspace-icon GET/HEAD responses share the failure.

The header above contains one opaque tag. Its embedded asterisk is literal and cannot match the representation's hash-based validator.

### Expected behavior

Return 200 with the current representation, or 206 and the requested byte range. HEAD retains the appropriate status and metadata without a body. Only a standalone `*` is a wildcard; weak comparison compares complete entity-tags.

### Actual behavior

The matcher splits at every comma, exposes the quoted asterisk as a separate fragment, and incorrectly returns 304. GET and range requests consequently receive no bytes.

### OpenClaw version

Reproduced with a built source checkout at `6aba348f6af03ec14033dbb1f47cf2a85d539e3c`. The affected owners are also present on inspected main `422b8c0ec78686d805c8ed4be37caa84591307dc`. Original introduction and the last unaffected release are unknown.

### Operating system

macOS 27.0; Node.js 24.20.0

### Install method

Trusted source checkout, pnpm 12.3.4, fresh built runtime with verified source/dependency/artifact inventories.

### Model

The conditional HTTP defect is model-independent. The isolated Gateway proof used a local scripted Responses fixture to create a real managed attachment through the normal chat flow.

### Provider / routing chain

Public `sessions.create` → `chat.send` → `agent.wait` → `artifacts.list`/`artifacts.download` → authenticated managed-media HTTP requests. Public `chat.startup` prepares the workspace-icon sibling. A configured static asset supplies date-precedence controls.

### Logs

A 21-case matrix reproduced exactly seven incorrect 304 responses: managed GET, HEAD, two range cases, duplicate nonmatching fields, and workspace-image GET/HEAD. The other 14 controls passed. Requests, response bytes and headers were captured over real loopback HTTP sockets; both Gateway and provider shut down cleanly.

This is HTTP status/body behavior. Independent source tests reproduce ten failures in a 25-case corpus, including three weak-comparison cases with comma-bearing current validators. The public routes emit hash-based ETags, so those three cases have source-test coverage rather than fabricated live validators.

Contract: [RFC 9110 entity-tags](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.8.3) and [If-None-Match](https://www.rfc-editor.org/rfc/rfc9110.html#section-13.1.2).

### Impact and severity

Clients supplying these valid nonmatching validators can receive an empty response instead of full or partial media bytes. The exact cases are reproduced; frequency in ordinary browser traffic has not been measured.

steipete:
Closes #140528

<details>

Keep **Allow edits from maintainers** enabled so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Fixes empty `304 Not Modified` responses when a client sends a nonmatching quoted ETag such as `"client,*,tag"`. Its literal asterisk was incorrectly treated as a wildcard, suppressing managed-media bytes and producing incorrect workspace-image revalidation results.

## Why This Change Was Made

The shared matcher now compares complete quoted tags, removes the weak prefix from both operands, and recognizes only a standalone wildcard. It reuses the existing delimiter scanner with explicit quote rules: ETags keep backslashes literal, while Accept retains quoted-string escaping. The old private scanner is removed.

## User Impact

Nonmatching tags receive the normal full or ranged response. Matching tags, weak matches, wildcard requests and conditional-request precedence retain their intended behavior. The shared owner covers managed media and image callers; the media playback documentation explains the contract.

## Evidence

- Original 25-case regression corpus: ten intended failures and 15 passing controls. Fixed corpus: 25/25.
- 631 distinct tests across six Gateway/HTTP suites passed, including Accept/REST/SSE siblings. The managed-media file was rerun after the final fixture cleanup: 129/129. Other tested source files are unchanged.
- Complete eight-path changed gate passed; two independent review passes found no actionable P0–P2 issues. Final commit: `f369f6de53d7d3e47eb10558a0a000e194742da6`.
- Fresh uncached build passed. Real isolated Gateway before/after proof used public session/chat/artifact creation, workspace-icon preparation and exact HTTP sockets, with one local scripted Responses request per build.

| Gateway proof | Baseline | Fixed |
| --- | --- | --- |
| Matrix requests | 21 | 21 |
| Incorrect 304 responses | 7 | 0 |
| Expected control behavior | 14/14 | 14/14 |
| Full/partial/not-modified responses | 4 / 1 / 16 | 9 / 3 / 9 |

Both runs verified exact bytes, ranges, metadata, complete source/dependency/artifact inventories and unforced process/listener cleanup. Three comma-bearing **current** ETag cases remain source tests because these public routes emit hash-based validators. This proof exercises the actual Gateway with a local model fixture; it makes no external-provider claim.

Validation first exposed test-fixture typing, an omitted Host in raw headers, and unnecessary table mapping rejected by lint. Those fixtures were corrected before the passing final runs. An earlier proof log needed a redaction-quoting correction; the original evidence is retained and both complete Gateway runs were repeated successfully.

Production **+43/−45, net −2**; tests **+197/−33, net +164**; docs **+2**. The final patch composes byte-for-byte onto inspected main `422b8c0ec78686d805c8ed4be37caa84591307dc`. Original introduction is unknown in the available history.

### DIFF
--- a/docs/nodes/media-playback.md
+++ b/docs/nodes/media-playback.md
@@ -89,6 +89,8 @@ The ticketed byte routes support:
 - `ETag` and `If-Range` for safe resume of immutable managed originals
 - `HEAD` requests with the same content metadata and no response body
 
+For immutable originals, `If-None-Match` compares complete quoted tags using weak comparison. Commas and asterisks inside a quoted tag are literal; only a standalone `*` is a wildcard. A nonmatching tag leaves the normal full or ranged response intact.
+
 Local assistant files can change, and playback renditions can become available
 after a conversion retry. These responses revalidate without reusable validators:
 cached ETags or modification dates cannot suppress fresh bytes, and `If-Range`
--- a/src/gateway/control-ui-conditional.http.test.ts
+++ b/src/gateway/control-ui-conditional.http.test.ts
@@ -79,6 +79,11 @@ const conditionalCases: {
   },
   { name: "stale If-None-Match alone", headers: { "If-None-Match": '"stale"' }, status: 200 },
   {
+    name: "quoted comma/star is not a wildcard and supersedes the date",
+    headers: { "If-None-Match": '"client,*,tag"', "If-Modified-Since": lastModified },
+    status: 200,
+  },
+  {
     name: "stale If-None-Match overrides equal If-Modified-Since",
     headers: { "If-None-Match": '"stale"', "If-Modified-Since": lastModified },
     status: 200,
--- a/src/gateway/http-byte-range.test.ts
+++ b/src/gateway/http-byte-range.test.ts
@@ -445,6 +445,21 @@ describe("resolveByteResponse", () => {
   });
 
   it.each([
+    { etag: '"opaque,tag"', header: 'W/"opaque,tag"' },
+    { etag: 'W/"opaque,tag"', header: '"opaque,tag"' },
+    { etag: 'W/"opaque,tag"', header: 'W/"opaque,tag"' },
+  ])("weakly compares complete $header against $etag", ({ etag, header }) => {
+    expect(
+      resolveByteResponse({
+        ...IMMUTABLE_FILE,
+        validators: { ...IMMUTABLE_FILE.validators, etag },
+        method: "GET",
+        request: createByteRequest({ "if-none-match": header }),
+      }),
+    ).toMatchObject({ kind: "not-modified", statusCode: 304 });
+  });
+
+  it.each([
     { label: "exact", header: (etag: string) => etag },
     { label: "weak", header: (etag: string) => `W/${etag}` },
     { label: "wildcard", header: () => "*" },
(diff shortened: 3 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f369f6de53d7d3e47eb10558a0a000e194742da6](https://github.com/openclaw/openclaw/pull/140529/commits/f369f6de53d7d3e47eb10558a0a000e194742da6)
- Landed commit: [faec7a7ca239459da42cdae91a1ac481bb107215](https://github.com/openclaw/openclaw/commit/faec7a7ca239459da42cdae91a1ac481bb107215)

PR #140533, opened 2026-09-07
refactor(tests): share explicit model fallback configs

steipete:
## What Problem This Solves

Model fallback tests repeat the same nested configuration structure around explicit primary and fallback model references in 55 places.

## Why This Change Was Made

Add a constructor in the existing test fixture owner that requires both inputs, creates fresh wrapper objects, and preserves the supplied fallback array. Use it only at the 55 audited sites. The existing helper's defaults and shallow override behavior remain unchanged, as do case inputs, expected outcomes, and allocation timing. Net removal: 360 test lines and 49 test-support lines.

## User Impact

Less repeated test fixture code, with no production behavior change or runtime improvement claimed.

## Evidence

All 337 focused cases pass before and after, plus 161 model-selection and candidate-resolution siblings. The full changed-file gate, deslop, and independent review through P2 pass.

The preservation proof compares all 55 ordered object graphs, property descriptors, values, array references, and per-call freshness. It also matches 46,868 parsed tokens with allocation markers and confirms the old helper is byte-identical. A reversed-fallback diagnostic fails the independent order comparison. Intentional local-variable sharing, table-time allocation, and independent expected results remain intact.

### DIFF
--- a/src/agents/embedded-agent-runner/run.codex-app-server-recovery.test-support.ts
+++ b/src/agents/embedded-agent-runner/run.codex-app-server-recovery.test-support.ts
@@ -4,7 +4,7 @@ import { expectDefined } from "@openclaw/normalization-core";
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
 import { createDeferred } from "../../../test/helpers/promise.js";
 import type { OpenClawTestState } from "../../test-utils/openclaw-test-state.js";
-import { makeModelFallbackCfg } from "../test-helpers/model-fallback-config-fixture.js";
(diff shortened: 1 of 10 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5de70311345437b6673b9f8fddcf0823e6bd3522](https://github.com/openclaw/openclaw/pull/140533/commits/5de70311345437b6673b9f8fddcf0823e6bd3522)
- Landed commit: [b3d67aa9b80bd95ff696332ec8cc8c25e40a191d](https://github.com/openclaw/openclaw/commit/b3d67aa9b80bd95ff696332ec8cc8c25e40a191d)

PR #140534, opened 2026-09-07
fix(plugins): isolate unreadable tool factory errors

steipete:
A plugin tool factory can throw an Error whose `message` getter also throws. The resolver caught the factory failure, but its diagnostic used `String(err)`, which threw again and aborted assembly of healthy sibling tools.

Use the shared redacted error formatter at that catch boundary so the failed factory produces one diagnostic and resolution continues. Keep the existing suppression for invalid-config diagnostics already emitted by their owner. Extend the existing error table to cover the unreadable message and retain a healthy sibling in all three cases.

Validation on pinned main `30ea82ac`: the new case fails before the fix while both controls pass; 189 focused tests across optional tools, host-hook contracts, and registry lifecycle pass after it. A separate process with two generated plugins through the real loader reproduces the baseline assembly failure, then returns and successfully executes the healthy tool after the fix. That process receives an explicit minimal environment, no credentials, and isolated state; both runs exited naturally and their state directories were cleaned up.

Production LOC delta: +1 for importing the existing formatter; test LOC delta: +9. No new helper, dependency, configuration, or schema.

The complete changed-file gate, including typechecks and zero runtime import cycles, and independent P2 review pass.

### DIFF
--- a/src/plugins/tools.optional.test.ts
+++ b/src/plugins/tools.optional.test.ts
@@ -3698,6 +3698,14 @@ describe("resolvePluginTools optional tools", () => {
   it.each([
     ["unlogged invalid-config", createInvalidConfigError("/tmp/openclaw.json", "invalid property")],
     ["ordinary factory", new Error("factory unavailable")],
+    [
+      "unreadable-message factory",
+      Object.defineProperty(new Error("unreadable message"), "message", {
+        get() {
+          throw new Error("message getter failed");
+        },
+      }),
+    ],
   ])("still logs %s errors from plugin factories (#137694)", (_kind, error) => {
     setRegistry([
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [370e1da9166e4cfdd2be23ef0341f1da09ad4d32](https://github.com/openclaw/openclaw/pull/140534/commits/370e1da9166e4cfdd2be23ef0341f1da09ad4d32)
- Landed commit: [7e701b1208ee3e8286cb8826351dd3f9234bf9c5](https://github.com/openclaw/openclaw/commit/7e701b1208ee3e8286cb8826351dd3f9234bf9c5)

PR #132412, opened 2026-08-29
fix(tooling): report every deadcode scan outcome

steipete:
## What Problem This Solves

Resolves a problem where developers running deadcode checks lose later scan findings and terminal outcomes when an earlier scan fails. All scans already run concurrently, but the aggregate reporting loop returned after the first failure, so a stopped heartbeat could have no visible success or failure result.

The neighboring timeout and output-limit tests also used a synchronous child-process mock that could leave a delayed real signal aimed at the synthetic process-group ID after the mock was restored.

Related context: [the concurrent Knip scan change](https://github.com/openclaw/openclaw/pull/109151), which carried the serial first-failure return into the concurrent reporting path.

## Why This Change Was Made

Each scan now reports its outcome when it settles. The wrappers still join every scan and return a nonzero aggregate exit status if any scan fails; redundant aggregate success messages are removed. Both export and unused-file guards use the same approach without introducing another helper or execution path.

The test doubles now deliver terminal events asynchronously, matching Node's child-process contract. A safe outer observer checks that no signal callback survives restoration of the inner mock, including assertion-failure cleanup. The CI documentation now describes the actual hard-zero, production/full-tree/script coverage rather than a retired unused-file allowlist.

**Unchanged by this PR:** concurrent launch, every required scan and issue class, Knip and package-manager pins, the 600-second production deadline, the 5-second termination grace, output bounds, shared process runner, configs, guards, and baselines. This is a reporting and test-safety repair, not a scan-throughput fix.

## User Impact

Developer and CI logs retain every completed scan's result, including later failures, while still failing the overall command. Tests no longer leave synthetic-PID signaling work behind. There is no gateway, updater, user configuration, plugin API, persistence, or live-service behavior change.

Production/tooling LOC: **+12/-22 (net -10)**. Tests: **+177/-39 (net +138)**. Docs: **+1/-1**. No dependency, configuration option, or test-only production seam was added.

AI-assisted maintainer repair.

## Evidence

- **Failing before:** final assertions exposed five expected failures—three missing later-report cases and two post-restoration `SIGKILL` callbacks. The all-success controls passed on pre-fix code. Synthetic-PID traffic was intercepted throughout the negative control; the unsafe original suite was not run unguarded.
[...]

PR #140543, opened 2026-09-07
refactor(test): use executable deadcode coverage

steipete:
## What Problem This Solves

Deadcode tests still check the spelling of scanner flags, config paths, and a status predicate even though the existing reporting suite observes those contracts through the real CLI and child processes. Equivalent implementation changes can therefore break redundant source-string assertions.

## Why This Change Was Made

Remove seven source-spelling assertions across two test files. The reporting suite retains exact child arguments for production, full-tree, and script scans, including all export issue classes and entry-export flags. The unused-file suite still rejects a nonzero scanner exit with empty output.

The executable reporting coverage was introduced in #132412. Its configuration, no-allowlist, zero-findings policy, parsing, timeout, signal, and descendant-cleanup checks remain intact. The retained fake-PID fixture also uses Vitest to restore the exact original process method, resolving existing method-binding and cleanup-style lint errors without changing its grace window or signal behavior.

## User Impact

No production change. Developers retain behavioral guard coverage while avoiding duplicate constraints on source formatting and expression spelling.

## Evidence

- Baseline 50/50; candidate 49/49 across the exports, unused-files, and reporting suites. Only the redundant source-assertion case was deleted; every retained case title matches.
- Removing an export issue class deliberately fails all three retained export-reporting scenarios. Ignoring nonzero status deliberately fails the retained empty-output rejection case. Both temporary controls were fully reverted before candidate proof.
- The reporting suite executes both real CLI owners with synthetic scanner output. All seven scenarios, production files, and test-support helpers are unchanged.
- Two test files: +21/-27 lines, seven assertions removed, no new cases or production changes. Six lint errors reproduced on unchanged main; the fixture correction passed all 49 cases. Final independent review is clean through P2, and all 15 normal changed checks passed.

### DIFF
--- a/test/scripts/check-deadcode-exports.test.ts
+++ b/test/scripts/check-deadcode-exports.test.ts
@@ -44,20 +44,6 @@ function listQaScenarioExecutionPaths(dir = "qa/scenarios"): string[] {
 }
 
 describe("check-deadcode-exports", () => {
-  it("requests every unused-export issue class from Knip", () => {
-    const script = fs.readFileSync(
-      new URL("../../scripts/check-deadcode-exports.mts", import.meta.url),
-      "utf8",
-    );
-    expect(script).toContain('"exports,nsExports,types,nsTypes,enumMembers,namespaceMembers"');
-    expect(script).toContain('"config/knip.config.ts", "--production"');
-    expect(script).toContain('"config/knip.all-exports.config.ts"');
-    expect(script).toContain('"config/knip.scripts-exports.config.ts"');
-    expect(script).toContain(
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d9e35d036291e639d00a8201bc0d823213dd150e](https://github.com/openclaw/openclaw/pull/140543/commits/d9e35d036291e639d00a8201bc0d823213dd150e)
- Landed commit: [700276c4e18579c040d0c64411ac18b54ca09f67](https://github.com/openclaw/openclaw/commit/700276c4e18579c040d0c64411ac18b54ca09f67)

PR #135599, opened 2026-09-01
feat: manage and reload plugins without restarting the Gateway

steipete:
This implementation is being split into independently landable PRs. This draft preserves the complete candidate and end-to-end evidence; the full restartless lifecycle feature is not yet landed.

Landed slices:

- #139775 — simplify plugin diagnostics and state wrappers (−13 production LOC).
- #139807 — wait for Android trust lifecycle publication in its test (no production change).
- #139788 — centralize managed npm install planning (−43 production LOC).
- #139769 — make suppression inventory independent of line numbers (no production change).
- #139898 — bind registration owners once (−157 production LOC).
- #139828 — keep tool metadata and callbacks on the same factory instance (−710 production LOC).
- #139908 — honor provider runtime cache invalidation (−151 production LOC).
- #139957 — observe subprocess exit before completing forced cancellation (+9 production LOC).
- #140002 — centralize install recovery metadata (−56 production LOC), with real CLI allowed/refused recovery proof.
- #140040 — share managed npm package traversal (−17 production LOC), preserving caller policies and native filesystem errors.
- #140081 — unify install config snapshot preparation (−36 production LOC), with real CLI recovery and plugins-include proof.

- #140132 — retain service cleanup through late startup (+13 production LOC), with held-startup, ordering and exact-once cleanup proof.
- #140145 — prepare managed npm installs privately before publication (−545 production LOC), with five real-npm rollback, refusal, quarantine, cancellation and late-import scenarios.

- #140205 — derive catalog types from the canonical Gateway protocol (−20 production LOC), with exact type equality and identical emitted JavaScript.
- #140215 — share CLI/management enablement mutations (−32 production LOC), preserving caller policy and consent order with real CLI parity.

- #140322 — share CLI/management uninstall ownership (−50 production LOC), with inherited Claw lease, failure recovery and fourteen real CLI parity scenarios.

- #140362 — preserve cross-realm/disposal diagnostics and successful tool results when observer reporting fails (+43 production LOC), with causal Node probes and repaired restart/Codex consumer tests.

- #140433 — preserve saved config in auth reload fixtures (no production change), with four causal Gateway/WebSocket cases and exact-head green CI.

- #140453 — preserve saved source in the remaining SecretRef and Control UI reload fixtures (no production change), with two original failures, six passing Gateway cases and exact-head green CI.

- #140461 — select the matching provider credential in the shared setup-plan owner (−12 production LOC), with causal multi-profile cases and real config-loaded provider proof.

[...]

PR #140551, opened 2026-09-07
fix(ui): reconcile superseded config refreshes after mutations

steipete:
A successful plugin or other external config mutation could show a configuration-refresh warning even when a later refresh had already loaded the new state. The original read reported that it was superseded, and the mutation treated that as a failed refresh.

Record completion at the config-read owner and follow actual later reads within the same connection. Every mutation still starts a fresh read after its RPC. A later failure remains a failure; writes, disconnects, and disposal release pending refresh waits. Ordinary loads keep their own result and recovery callback, and unsaved drafts retain their bytes and base hash.

Validation: both successful response-order regressions fail on main; 184 config tests pass with the fix. An independent browser check passed nine synthetic contracts, including the actual plugin Enable → config.get flow and browser capability checks for ordering, failures, drafts, reconnects, and callback ownership. No real Gateway or plugin backend is claimed. Independent Codex review and the full changed-file gate passed, including UI/test types and i18n. Paired, inspected screenshots show the false warning on main and successful completion with the fix under identical mocked Gateway traffic.

Production +97 lines records the actual read's completion and invalidation using the existing request versions and connection epochs. No new config, schema, dependency, or public API is introduced. This is independent of the larger plugin lifecycle work in #135599.

Before and after under identical mocked Gateway traffic:

Before: false refresh warning

After: completed refresh

### DIFF
--- a/ui/src/lib/config/config-external-refresh.test.ts
+++ b/ui/src/lib/config/config-external-refresh.test.ts
@@ -1,0 +1,236 @@
+// @vitest-environment node
+import { describe, expect, it, vi } from "vitest";
+import type { GatewayBrowserClient } from "../../api/gateway.ts";
+import type { ConfigSnapshot } from "../../api/types.ts";
+import { loadConfig } from "./config-gateway-operations.ts";
+import { nextRequestVersion } from "./config-state-model.ts";
+import { createConfigCapabilityHarness, deferred } from "./config-test-harness.ts";
+
+function snapshot(count: number): ConfigSnapshot {
+  return {
+    config: { count },
+    raw: JSON.stringify({ count }),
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5e430ab74d3c8879343805377b0048bacbdfab5b](https://github.com/openclaw/openclaw/pull/140551/commits/5e430ab74d3c8879343805377b0048bacbdfab5b)
- Landed commit: [0cd1ca7552a005a405dea7311036c063dc9635d0](https://github.com/openclaw/openclaw/commit/0cd1ca7552a005a405dea7311036c063dc9635d0)

PR #140557, opened 2026-09-07
fix(macos): avoid Intel compatibility warnings in Apple silicon app bundles

steipete:
## What Problem This Solves

Fixes an issue where users launching a native Apple silicon OpenClaw app could see macOS's “Support Ending for Intel-based Apps” warning because unused Intel-only dependency prebuilds remained inside the app bundle.

## Why This Change Was Made

Apply the existing architecture-aware worker materializer to both standard and elevation app packaging. Each worker is derived from the complete installed package, verified in isolation, and published through the existing same-volume move flow. Incompatible macOS, Linux, and Windows binaries are omitted; compatible universal binaries, JavaScript, WASM, resources, and valid relative symlinks are preserved.

## User Impact

Apple silicon builds no longer carry unused Intel-only worker dependencies. Universal distribution builds continue to include their separate Intel worker. This change does not alter dependency versions, app signing policy, or Gateway state.

## Evidence

- Extended the existing staging fixture with opposite-architecture and universal Mach-O binaries. Before the fix, standard packaging failed because it retained the opposite-architecture and foreign-platform binaries; the elevation variant passed.
- `node scripts/run-vitest.mjs test/scripts/mac-node-worker.test.ts`: 117 tests passed.
- `node scripts/check-changed.mjs`: passed, including formatting, lint, test typechecking, and all 1,124 macOS/platform checks.
- Applied the unchanged materializer to a staged copy of an installed arm64 app: removed two Intel-only addons and six Linux/Windows prebuilds. All 21 remaining Mach-O binaries support arm64. Re-signed using the existing Developer ID identity; strict recursive signature verification passed.
- Both retained arm64 addons load with the bundled Node runtime. The canonical worker verifier passed native capability checks and both fresh-state and app-initialized-state readiness cases. The cleaned app was installed and relaunched successfully.
- Independent review found no actionable P0–P2 issues. The historical notification itself was not used as proof of removal; the binary inventory and runtime checks establish the repaired boundary.

### DIFF
--- a/docs/platforms/mac/dev-setup.md
+++ b/docs/platforms/mac/dev-setup.md
@@ -41,6 +41,12 @@ then replaces the previous app. `scripts/restart-mac.sh` uses
 the same path; `SKIP_TSC=1` no longer bypasses the runtime build. Existing
 content-checked build caches still avoid unnecessary declaration work.
 
+Each worker keeps native binaries that support its architecture and omits
+incompatible macOS, Linux, and Windows prebuilds. This prevents unused Intel-only
+dependencies from triggering macOS compatibility warnings in Apple silicon
+builds. Compatible universal binaries, JavaScript, WASM, and other resources
+remain intact.
+
 Universal builds require both arm64 and x86_64 runtimes to execute during
 validation. Building x86_64 on Apple Silicon requires Rosetta; a missing
 architecture or nonportable native dependency fails packaging. Node downloads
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [abca087d17f465edfae8f3c43932d74180405fd7](https://github.com/openclaw/openclaw/pull/140557/commits/abca087d17f465edfae8f3c43932d74180405fd7)
- Landed commit: [5c756bdbae558d5e221467bf90197ee06d655615](https://github.com/openclaw/openclaw/commit/5c756bdbae558d5e221467bf90197ee06d655615)

PR #140558, opened 2026-09-07
fix: identify workers behind slow session-write warnings

steipete:
## What Problem This Solves

Resolves a problem where operators investigating slow session cleanup could not tell which worker a slow SQLite writer callback had awaited. Nearby database-open warnings already identified their worker, but the writer warnings lacked the process and worker association needed to connect them reliably.

## Why This Change Was Made

Existing writer warnings now include the emitting PID and Node thread identity. Each reclamation callback owns a small diagnostic record containing its actual plan kind and the worker's numeric ID, captured at creation and preserved after exit. This uses the existing worker identity without adding trace IDs, timers, queues, configuration, or message fields. Queue ordering, commit authorization, database validation, and cleanup remain unchanged.

## User Impact

Operators can join a writer warning's PID and workerThreadId to a database-open warning's PID and threadId within the same process lifetime. This identifies an awaited worker; it does not turn elapsed time into CPU attribution or itself reduce latency. The documentation explains those limits.

## Evidence

- The original warning fails the PID/thread regression checks. Restoring only the original dispatcher also makes the real file-log test fail specifically on the missing worker ID.
- Actual workers cover successful and rejected reclamation, identity retention after exit, same-store queueing, different-store isolation, and absent identity for in-process work or constructor failure.
- A small real JSONL logging test crosses the existing warning threshold with a bounded delay and proves the queued successor does not inherit its predecessor's worker ID.
- Linux Crabbox: all 30 focused tests passed on Node 24.19.0, including the real file-log flow; the one-shot lease stopped successfully. That native run covered the initial diagnostic implementation. The subsequent type-only correction emits byte-identical JavaScript with the repository's esbuild version; its 15 focused tests also passed locally on Node 26.8.1.
- Independent P2 review: no accepted findings. Six alternating batches of 10,000 silent no-op writer calls had median process CPU times of 79.97 ms before, 80.32 ms after, and 79.67 ms with a record allocated per call. Wall timings were noisy under concurrent checks; this is not a claim about production latency or worker-lifetime cost.
- The initial implementation passed the full local changed-file gate. Additional CI then caught a static type-import cycle that the runtime-cycle check does not cover. Moving the diagnostic type to the existing SQLite contract removes that back-edge: the exact Madge check changed from one cycle to zero; refreshed core/test typechecks, targeted lint, focused tests, and P2 review all passed. Exact-head CI must pass before landing.

- A required rebase preserved main’s new disk-budget and recent-session protections at the same worker call. All 49 composed local tests passed, including recent, archived, pinned, and cross-owner retention rechecks. Fresh core/test types, targeted lint, Madge, formatting and P2 review also passed. Exact-head CI remains required before landing.

The production change is net +54 lines, including formatter expansion at existing multi-argument call sites. The added capability is explicit diagnostic ownership across six existing modules, including the existing type contract; no generic context framework was introduced.

### DIFF
--- a/docs/gateway/diagnostics.md
+++ b/docs/gateway/diagnostics.md
@@ -137,6 +137,14 @@ lock-wait and hold warnings remain separate. Writes rejected before entering the
 writer omit these three fields. This breakdown is in
 file logs, not the aggregate Gateway RPC Prometheus histograms.
 
+These warnings include the writer's `pid`, Node `threadId`, and `isMainThread`.
+Reclamation callbacks also record their `reclamationKind` and, when a Worker was
+created, its captured `workerThreadId`. Within the same process lifetime, match
+the warning's `pid` and `workerThreadId` to an agent-database-open warning's
+`pid` and `threadId` to identify the awaited Worker. This establishes association,
+not CPU attribution or a breakdown of the Worker's lifetime. A missing Worker
+ID does not establish that work ran on the main thread.
+
 Stalled embedded-run diagnostics mark `terminalProgressStale=true`
 when the last bridge progress looked terminal (for example a raw response
 item or response-completion event) but the Gateway still considers the
--- a/src/config/sessions/session-accessor.sqlite-archive.ts
+++ b/src/config/sessions/session-accessor.sqlite-archive.ts
@@ -294,6 +294,7 @@ function resolveSourceWorkerExecArgv(): string[] {
 }
 
 function spawnSqliteTranscriptArchiveWorkerOperation<Result>(params: {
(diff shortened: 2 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [02922926d040bb277a1baccda76a95d8d25d57f1](https://github.com/openclaw/openclaw/pull/140558/commits/02922926d040bb277a1baccda76a95d8d25d57f1)
- Landed commit: [17a7bf646ee7269974ee57544e5800d97ad26aba](https://github.com/openclaw/openclaw/commit/17a7bf646ee7269974ee57544e5800d97ad26aba)

PR #140572, opened 2026-09-07
refactor(exa): test response readers through tool execution

steipete:
## What Problem This Solves

Exa's response-reader tests reach private helpers through a production testing export and use a test-only response-size override. That setup misses changes to the actual production size limit and duplicates response fixtures already shared by other tests.

## Why This Change Was Made

Move the same five reader-test subjects through the provider tool and mocked fetch boundary. Reuse the shared streaming and cancellation fixtures, remove the private testing export, and remove the size override that production never passes. The default 16 MiB cap and reader implementations remain in place.

## User Impact

No intended runtime or public API behavior change. Tests now protect actual request execution, including fatal UTF-8 decoding, exact successful content, the production response cap, early cancellation, and bounded error details. Existing routing, authentication, cache, and cancellation tests remain.

## Evidence

- Baseline and candidate: 17/17 cases with the same case-title inventory.
- The migrated overflow case fails when the production cap is doubled; the previous overridden-limit case still passed that fault. Disabling fatal UTF-8 decoding also fails the migrated case.
- Success preserves the exact URL and unwrapped title plus row count. Error cases retain the included prefix, excluded tail, short error, cancellation, and no-`response.text()` checks.
- Final independent review through P2 is clean. All 25 configured checks passed on Blacksmith Testbox ([run](https://github.com/openclaw/openclaw/actions/runs/34071932656)); the one-shot command exited 0 and its lease/snapshot cleanup completed. No authenticated Exa service call is claimed.
- Production: +2/-11 lines. Tests: +79/-73 lines, including removal of both duplicate response fixtures.

### DIFF
--- a/extensions/exa/src/exa-web-search-provider.runtime.ts
+++ b/extensions/exa/src/exa-web-search-provider.runtime.ts
@@ -75,12 +75,8 @@ type ExaSearchResponse = {
   results?: unknown;
 };
 
-async function readExaSearchResults(
-  response: Response,
-  opts?: { maxBytes?: number },
-): Promise<ExaSearchResult[]> {
-  const maxBytes = opts?.maxBytes ?? EXA_SEARCH_JSON_MAX_BYTES;
-  const bytes = await readResponseWithLimit(response, maxBytes, {
+async function readExaSearchResults(response: Response): Promise<ExaSearchResult[]> {
+  const bytes = await readResponseWithLimit(response, EXA_SEARCH_JSON_MAX_BYTES, {
     onOverflow: ({ maxBytes: maxBytesLocal }) =>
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1d55eb2afa195d872ec02f59387802de6d26b095](https://github.com/openclaw/openclaw/pull/140572/commits/1d55eb2afa195d872ec02f59387802de6d26b095)
- Landed commit: [4fcdbb33aeda3a25524b9aa35c97235c4d7c8238](https://github.com/openclaw/openclaw/commit/4fcdbb33aeda3a25524b9aa35c97235c4d7c8238)

PR #140575, opened 2026-09-07
fix(macos): Gateway error alerts no longer block quitting the app

steipete:
## What Problem This Solves

Fixes an issue where the Mac app could not be quit while a Gateway error alert was showing. When a saved Gateway window failed to open (for example "Could Not Open Gateway Window: connect to gateway @ wss://team.example/: There was a bad response from the server"), the alert was app-modal and ran a nested modal loop inside the main-actor task. While it was up, Quit, the `quit` Apple event, and SIGTERM did nothing for minutes; only a force kill ended the process. The same modal helper backed "Could Not Switch Gateway", "Could Not Open Background Session", "Could Not Set Primary Gateway", the Gateway setup deep-link error, and the "Set as Primary" confirmation when no window was available.

## Why This Change Was Made

Gateway alerts now go through a single `DashboardAlertPresenter` that never spins a nested modal loop. When a visible dashboard window exists the alert is attached to it as a sheet; otherwise it appears as a floating non-modal panel that keeps the alert's icon, title, message, buttons, and default/cancel key equivalents. `DashboardManager.close()` dismisses pending alerts first, which is the path `applicationShouldTerminate` cleanup already runs, so pending alerts can no longer hold termination hostage. A sheet whose host window is closed programmatically (profile removal) is ended explicitly so its completion still fires and identical follow-up alerts are not suppressed. Informational alerts with identical text are not stacked while one is pending; confirmations never deduplicate. The synchronous confirmation in the `openclaw://gateway?host=` setup route is unchanged because its coordinator contract is synchronous; other unrelated `runModal` sites in the app are out of scope.

## User Impact

Gateway errors show as a sheet on the dashboard window or as a small non-modal panel, and the app stays responsive while they are up: menus work, other windows keep updating, and Quit (menu, Apple event, or SIGTERM) completes promptly. "Set as Primary" from the status menu presents the same way instead of freezing the app.

## Evidence

- `scripts/format-swift.sh macos`, `scripts/lint-swift.sh macos`: clean.
- `swift build --package-path apps/macos --build-system native --product OpenClaw` and `--build-tests`: pass (native tests compile-only locally per `apps/macos/AGENTS.md`).
- `pnpm native:i18n:verify`: pass; the inventory only records two new uses of the existing "OK" string.
- `node scripts/check-changed.mjs`: pass.
- Codex autoreview (P0/P1): scoped-clean.
- Live proof on an isolated named app profile (remote mode pointed at an unreachable loopback address, never the operator Gateway), using the password-bearing Gateway setup deep link, which the coordinator rejects before any mutation and routes through the new presenter:
  - No dashboard window open: the non-modal panel appears; `tell application id "ai.openclaw.mac.debug" to quit` returned in 0.9 s with the panel up.
  - Dashboard window open: the same error attaches as a sheet; quit returned in 1.2 s with the sheet attached.
- Before: the standard app-modal `NSAlert.runModal()` panel (not captured; the pre-change build is not installed anywhere reachable).

After, no dashboard window (non-modal panel):

Gateway error shown as a floating non-modal panel

After, dashboard window open (sheet):

Gateway error shown as a sheet on the dashboard window


steipete:
Merged via squash.

- Prepared head SHA: [f3e6c2a34b0ac7be5438612372bc9542059bbf2d](https://github.com/openclaw/openclaw/pull/140575/commits/f3e6c2a34b0ac7be5438612372bc9542059bbf2d)
- Landed commit: [81af18c805a90b5de6a471235a4adba11fa6d15a](https://github.com/openclaw/openclaw/commit/81af18c805a90b5de6a471235a4adba11fa6d15a)

PR #139949, opened 2026-09-06
fix(config): preserve authored settings across includes

steipete:
Related: #130706

## What Problem This Solves

Configuration updates could lose authored model aliases or parameters, discard explicit settings beside `$include`, or reject root-owned arrays. Explicit writes could also duplicate composed arrays on reload. These cases affect plugin migration writes, combined roster/settings changes, and Doctor's legacy-agent migration.

## Why This Change Was Made

Finish roster and include reconstruction once, then apply explicit settings. Keep model-key normalization with the touched-model owner. Projection and explicit copying share array ownership derived from the recorded pre-migration source; the writer reindexes that evidence when legacy agent lists move to canonical entries, without migrating its field values.

Already-satisfied array, object and index writes retain their authored contribution. Actual changes to included or mixed array contributions remain protected. Duplicate preservation, traversal and error-construction paths were removed.

## User Impact

Unrelated updates preserve model metadata. Root-owned arrays remain editable, including environment-backed arrays. Explicit settings survive combined roster writes, and Doctor preserves nested includes and environment references while migrating legacy agents. Catalog-suppressed models still require explicit repair.

## Evidence

- Final correction passes 149 focused tests: all 128 writer cases and the unchanged 21-case Doctor legacy-composition suite. Earlier component verification also covered include resolution, mutation and migration siblings.
- All 33 production filesystem mutation/reload scenarios pass again on the final source, including successful unchanged array/object/index writes and unsafe replacements refusing without modifying either file.
- A separate real Doctor configuration flow reproduced the CI failure before repair, then passed migration → atomic write → disk/reload → second no-op, preserving both include files and authored environment references.
- Independent P2 review and the complete changed gate pass. Writer source SHA256: `299f5072b866a4a127a2b5d6a26c501f5395c2e3fa008abedf670e5e0289e4e1`; final head: `c5325345b532beea56227d5e85c8bef8abf5e683`.
- Production delta: +45; tests: +231. Remaining growth enforces ownership during explicit copying and aligns ownership evidence with canonical roster paths.

The initial hosted run caught the legacy-list/include coordinate regression; the existing test and real Doctor flow now pass. [Fresh exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34024743909) on `c5325345b532beea56227d5e85c8bef8abf5e683`. Single-agent to multi-agent topology expansion under an include retains its existing refusal, verified on the base; supported explicit multi-agent roster changes are covered.

PR #130706, opened 2026-08-27
fix: prevent Gateway stalls with multiple workspaces

steipete:
The useful correctness and cleanup work from this PR has landed in **83 focused PRs**, with the complete replacement map and per-cut evidence in #142100. This body replaces the original broad implementation proposal. #130706 remains closed **unmerged** as replaced by that series, with its branch preserved; #142100 remains closed unmerged as the completed decomposition/reference map.

The focused repairs preserve selected model identities and keep their metadata, authentication, pricing, limits and runtime ownership consistent. They also correct configuration publication, channel controls, media, compaction and fallback reporting while removing duplicate paths. Actual landed production accounting is **5,453 lines added / 5,508 removed: net −55**, measured once per merge against its first parent. Including the separate UI repair #143889 gives **net −49**. Tests, QA scenarios and tooling are counted separately; no forecast or unmerged work contributes.

All **744 original paths across 19 neighborhoods** have recorded dispositions. The general collection framework and extra storage/listener/retention plumbing, separate projector, unsupported import/mechanical changes and blanket constrained-ACP refusal were not adopted. Broader captured-source, sparse-donor/route-reader and compaction proposals also remain unadopted for the ownership/provenance defects documented in #142100. Existing canonical owners were reused.

The previous three cuts—#144438, #144354 and #144474—have verified native merges, required CI and qualified runtime evidence. Source tests, compiled CLI/API checks, real Gateway lifecycle, local loopback peers and authenticated external scenarios remain separate. Historical and production-identical proof keeps its original revision; #144474's final test-only adapter carries the earlier compiled proof. #140488's changed hook behavior passed, but its full live driver remains failed on cleanup-confidence warnings. The map preserves the five early execution/provider unknowns and all failed-attempt qualifications.

The September 13 continuation landed #147830 and #147849, reducing production by another nine lines. These repair captured image model/resize selection and exact configured-row transport facts for registered harnesses. Each passed its own focused checks, independent review, normal build, scoped compiled proof and exact-head CI, followed by native merge verification. The image proof uses real registered providers and SSE loopback; the harness proof uses public CLI/SDK selection through a real synthetic plugin. Hosted inference and the separate non-idempotent alias handoff remain unclaimed.

[...]

PR #116108, opened 2026-07-29
fix(config): save config changes into the nested $include file that owns them

sasan1200:
Closes #116107

## What Problem This Solves

Fixes an issue where users split configuration into nested `$include` fragments and config commands or `doctor --fix` cannot save changes to the owning file. The old writer recognized only top-level include ownership and correctly refused to flatten nested fragments into the root.

## Why This Change Was Made

Config writes and Doctor share one resolver that selects the deepest sole include owner. It accounts for every logical owner of the canonical target, including filesystem aliases and merged includes, so updating one subtree cannot silently alter another. The writer revalidates captured include targets and contents around persistence and after reload, with conditional rollback if an intermediate file changes. Current-main authority checks and roster projection order are preserved.

Doctor omits root wizard metadata for an include-only repair and records refused persistence accurately. Root-owned keyed-agent sibling additions preserve existing include pointers. The exported `ConfigIncludeOwnership.hasArrayAncestor` remains optional, with absence meaning false.

The maintainer approved this as a feature-class change on September 6, including positive production growth. No new config option, dependency, schema, protocol, or multi-file write mechanism is introduced. The mainline conflict was resolved with a merge so Sasan's original commits and authorship remain intact, as required by the native landing ancestry gate.

Compatibility review: existing authored config formats and canonical keys are retained. `hasArrayAncestor` is optional snapshot metadata, so existing callers that omit it remain valid. Existing sole-owner top-level writes remain covered by the focused suite. Shared physical targets now intentionally refuse writes that could alter another logical owner; operators can edit those files directly or give each subtree its own fragment.

Stored-data review: this adds no SQLite schema, persisted config key, or new migration transform. Existing Doctor migrations now persist their already-canonical result in the owning fragment. The nested `sandbox.perSession` to `sandbox.scope` fixture proves that upgrade repair with root and intermediate bytes preserved; the companion fixture proves mixed-owner repairs are refused without changing authored files. The maintainer's September 6 feature approval covers this persistence capability.

## User Impact

Eligible nested fragments receive their own changes while root and ancestor files retain their authored bytes. Root includes, array and merged ownership, sibling overrides, shared physical targets, external write targets, and mutations spanning multiple owners remain refused. Concurrent include-graph changes cannot be reported as a successful write to an inactive fragment.

Thanks to @sasan1200 (Sasan Sotoodehfar) for the original implementation.

## Evidence

- Candidate head: 80a755a50d998e71fa59896d799a9f172dd53344.
[...]

PR #140579, opened 2026-09-07
fix(config): preserve authored settings through updates and setup

steipete:
Related: #130706, #139949, #116108.

## What Problem This Solves

Fixes an issue where users changing configuration or completing setup could save runtime-only defaults, lose authored values or environment references, or have explicit changes silently undone. This includes settings stored in `$include` files, default-valued settings deliberately pinned by the operator, and earlier wizard security/telemetry choices.

Connected failures affected literal `null` values in full replacements and runtime-derived edits, indexed array removals through the writer/SDK, Doctor migrations that saved schema defaults, and plugin discovery using stale or mismatched environment values after a config write.

## Why This Change Was Made

The mutation and persistence owners carry the input's source/runtime basis together with its native read snapshot, hash and environment. Root and include writes share projection and explicit-value handling. A presence-aware projection preserves literal values, including `null`, while treating omitted keys as removals; array removals are validated on a copy and applied once to final output, after materialization.

Doctor validates migrations but persists their source candidate. Registry refresh reads committed configuration with the correct published/caller environment, while prepared setup probes retain their explicit config and installer-owned records. The classic wizard keeps its matched merge base across agent preparation. Duplicate projection/deletion pipelines, environment-baseline readers, option declarations and loader adapters are removed; forwarding wrappers are inlined, and remaining coercion, hash and schema reads use canonical helpers.

## User Impact

Requested settings survive updates without copying unrelated defaults into the file. Full `config.set`/`config.apply` requests and runtime-derived replacements preserve valid nullable values; `config.patch` retains its existing null-means-delete contract. Source-key unsets retain their documented exact-key behavior. The config CLI already handles indexed deletions without replaying them; the double-removal repair concerns direct writer/SDK requests.

Existing runtime-derived `nextConfig` calls remain supported, and source producers use `sourceConfig`. Snapshot, path, include-graph, concurrent-write, consent and publication checks remain in place. No operator configuration option, environment variable, database schema or protocol version is added.

This repair was found while extracting Peter Steinberger's (@steipete) work in #130706 and extends #139949.

## Evidence

Current head `639605e4c0420b197633baf09a0121dbd306cbf2` integrates main `599762504579943ae1b04dd1de6f2e15dddf4b8a`. All **18 focused wizard/default-agent, provenance and config-health tests** pass, including main's typed startup-maintenance failures. The complete changed gate for the five overlapping paths passes. A fresh independent four-pass P2 review of the full 132-path PR is scoped-clean. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34166980189) passed: **109 successful jobs, 15 skipped, zero failures**. The committed tree matches the reviewed and tested tree.

A natural full build on [Blacksmith Testbox](https://github.com/openclaw/openclaw/actions/runs/34167850325) passed on that exact head and tree. **All twelve compiled compatibility cases passed:** two Doctor config owners, eight public SDK scenarios, explicit caller-environment registry refresh and the plugin-consent caller. Source/runtime inventories were byte-identical before and after both suites. Every child process joined, every fixture was removed, and the Docker containers were removed. Doctor's effect guard recorded no unexpected filesystem, process or network attempts.

The full 167-file evidence stream was received and its hashes verified locally. The first decoder compared plain strings against the exporter's filesystem-component order; a local reader correction reconciled the already-complete stream without rerunning product code or changing an assertion, deadline, source/build guard or remote artifact. The original failed decode is preserved separately. Earlier incomplete evidence from an expired lease is not claimed as a passing collection. Stopping a Testbox lease can cancel its hosting workflow; the command and decoded proof receipts establish the result.

The single wizard conflict retains the writer's saved merge base and the base returned by agent preparation, alongside main's comment. Other overlapping changes preserve main's diagnostics, documentation and ratchet updates. All other PR paths retain the same ordered additions and deletions. Dependency versions, lockfile, package exports and package-manager pin are unchanged; main's build-entry and declaration tooling remain intact.

Historical `2426584909a9df7e0a3c55e8bc5eb1a29f727fa9` compatibility evidence includes a natural full build, SDK8, real Gateway RPC9, registry CLI9, caller/setup19, include13, caller-environment1, nullable-value8 and the full 52-case Gateway configuration suite. The subsequent e6 head passed 104 selected integration tests and the full 132-path changed gate. These receipts retain their actual source attribution rather than being relabeled as current executions.

The compiled cases use synthetic config and owned processes. Doctor exercises its real config migration, decision and persistence owners; later host/auth/plugin effects are bounded synthetic boundaries. This is not a full Doctor CLI, provider-inference or operator-deployment test. The original local Doctor effect-guard failure remains recorded; the Linux capsule provides a private temporary filesystem for normal SQLite lifecycle IPC.

An older models/params-removal RPC timed out at its unchanged 30-second deadline. An unchanged diagnostic and continuation passed, but no root cause was established; the timeout remains an unattributed follow-up. The original multi-workspace CPU incident is not claimed reproduced or resolved.

The PR adds **43 production lines** against its integrated main base after removing duplicate projection, deletion and adapter paths. Test support is counted separately. Final extraction-campaign accounting uses actual squash parents.


steipete:
Merged via squash.

- Prepared head SHA: [639605e4c0420b197633baf09a0121dbd306cbf2](https://github.com/openclaw/openclaw/pull/140579/commits/639605e4c0420b197633baf09a0121dbd306cbf2)
- Landed commit: [07636e80c7161c7c486d4dc0c83e641328978218](https://github.com/openclaw/openclaw/commit/07636e80c7161c7c486d4dc0c83e641328978218)

PR #140580, opened 2026-09-07
fix(portals): reach IPv6 loopback with IPv4-only localhost records

steipete:
Related: #139428

## What Problem This Solves

Fixes an issue where users opening a portal backed by an IPv6-only local server would fail to connect when the host's `localhost` records listed only IPv4. This affects both Gateway-hosted targets and worker portal streams, despite their documented support for either loopback family.

## Why This Change Was Made

Gateway and worker connections now share explicit IPv4 and IPv6 loopback lookup options while retaining Node's automatic family selection. The lookup supplies only `127.0.0.1` and `::1`; connection ordering relative to portal attachment, cancellation, and socket cleanup stays with the existing owners. The portal documentation now makes the host-DNS independence explicit.

## User Impact

Portals can reach IPv4-only or IPv6-only local applications even when system localhost records omit the application's address family. No configuration change is required.

## Evidence

Controlled IPv4-only localhost resolution made both existing integration suites fail before the production fix. With the fix, all 83 cases across five focused files passed, including real Gateway and worker socket flows. The final helper test adjustment passed both explicit-family cases. Fresh independent P2 review is scoped-clean with no actionable findings.

The final full changed gate, formatting, diff checks, and deslop passed. An earlier gate caught a Promise executor lint issue; the corrected helper passed both cases and the complete gate. This correctness repair adds 58 test/support lines and 24 production lines, recorded as campaign offsets. No local build, aggregate coverage, or runtime improvement claim is made.

### DIFF
--- a/docs/gateway/portals.md
+++ b/docs/gateway/portals.md
@@ -61,7 +61,7 @@ The Gateway never executes these commands automatically. The agent reads the fil
 
 The application must honor `PORT`. Use `PUBLIC_URL` when it needs to generate absolute URLs.
 
-The server can listen on IPv4 or IPv6 loopback (`127.0.0.1` or `::1`), both on the Gateway host and on a worker. Worker streams also preserve the node's configured Gateway context path when connecting through a reverse proxy.
+The server can listen on IPv4 or IPv6 loopback (`127.0.0.1` or `::1`), both on the Gateway host and on a worker, even when the machine's `localhost` records list only one address family. Worker streams also preserve the node's configured Gateway context path when connecting through a reverse proxy.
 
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8408a1f012de5e8cb984a78a17d3b4c667faa25b](https://github.com/openclaw/openclaw/pull/140580/commits/8408a1f012de5e8cb984a78a17d3b4c667faa25b)
- Landed commit: [c565101f569996d7f1d7e4dc3ab18334d62e2077](https://github.com/openclaw/openclaw/commit/c565101f569996d7f1d7e4dc3ab18334d62e2077)

PR #140583, opened 2026-09-07
fix(ui): stop catalog paging in hidden tabs

steipete:
## What Problem This Solves

Fixes an issue where a Control UI tab continued fetching previously expanded catalog pages after the operator switched away from it.

## Why This Change Was Made

The existing visibility handler paused future polls, but an in-flight poll could still start every remaining page of its automatic refresh. Automatic pagination now checks visibility and current request ownership before each additional page, preserving the full previously displayed rows and cursor when it pauses.

## User Impact

Hidden tabs stop starting automatic catalog page requests. Returning to the tab uses the existing coalesced refresh to update the expanded list. Already-issued manual Load More outcomes, complete responses, polling intervals, and reconnect behavior are preserved.

## Evidence

- The actual sidebar fixture earns two pages through Load More. Baseline hiding during the base response starts two unwanted page requests; hiding during an expanded-page response starts one. Both regressions pass after the fix.
- All 370 focused and sibling tests pass, including retained rows/cursors, visible catch-up, hidden manual success/error outcomes, exhausted responses, unmount, progressive updates, and reconnect ownership.
- The production-built Control UI passes all three browser cases on commit `3091fa6861d732336fc90d139fb466a250958f7f`, using the same scenario that produced two baseline failures and one passing manual control.
- Independent P0–P2 review and the complete changed gate pass, including core/UI typechecks, all unused-export scans, lint, and styles.

The browser proof uses a synthetic Gateway and controlled document visibility. An initial harness attempt stopped before the race because its paused clock withheld mock response delivery; correcting that clock advancement produced the baseline and candidate results above. No production latency claim is made from these request counts.

The matched screenshots show the expanded list after hiding: the baseline continued replaying pages, while the fix retains the previous window until the visible catch-up.

Before: automatic paging continues after hiding

After: expanded rows are retained until the tab returns

### DIFF
--- a/ui/src/components/app-sidebar-session-catalog-live.ts
+++ b/ui/src/components/app-sidebar-session-catalog-live.ts
@@ -457,6 +457,7 @@ export async function refreshSessionCatalogsLive(params: {
       agentId: params.agentId,
       pageDepths: params.pageDepths,
       isCurrent: revisionIsCurrent,
+      canRequestPage: () => revisionIsCurrent() && document.visibilityState !== "hidden",
     });
     if (!revisionIsCurrent()) {
       return;
--- a/ui/src/components/app-sidebar-session-catalog-state.ts
+++ b/ui/src/components/app-sidebar-session-catalog-state.ts
@@ -85,6 +85,7 @@ export async function refetchExpandedSessionCatalogPages(params: {
   agentId: string;
   pageDepths: ReadonlyMap<string, number>;
   isCurrent: () => boolean;
+  canRequestPage: () => boolean;
 }): Promise<SessionCatalog[]> {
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3091fa6861d732336fc90d139fb466a250958f7f](https://github.com/openclaw/openclaw/pull/140583/commits/3091fa6861d732336fc90d139fb466a250958f7f)
- Landed commit: [0a55c7c0d199da69e343ddf10115cb2dcba71b16](https://github.com/openclaw/openclaw/commit/0a55c7c0d199da69e343ddf10115cb2dcba71b16)

PR #138312, opened 2026-09-04
perf(plugins): prepare catalog source references once per filter

steipete:
## What Problem This Solves

Official plugin listing, package lookup and distribution classification repeatedly reconstruct the same source-reference sets while validating one catalog batch. This adds avoidable work to catalog-backed management operations.

## Why This Change Was Made

Prepare the allowed source references once at the synchronous filter boundary, then validate every candidate against that snapshot. Remove the unused private error-array builder. The public validation helper retains its existing result and ordering, and empty batches remain lazy. Nothing is cached across calls: later configuration mutations are still observed.

Production: 24 additions / 38 deletions, net **−14 lines**. Tests: 42 additions / 7 deletions, net +35. No schema, configuration, network, persistent-store or public API change.

## User Impact

Catalog-backed operations do less repeated preparation without changing which entries are accepted, their order, identity or validation errors. This is a focused local improvement; it does not establish faster Gateway startup or model turns.

## Evidence

- The same 118 cases in three complete test files passed before and after: official-external-plugin-catalog, official-external-plugin-repair-hints and management-service. Coverage retains mixed-source rejection, a valid later entry, list order/reference identity and changed configuration between loads.
- The full changed-file gate passed, including core/type checks, formatting, lint and dead-export checks, without bypasses.
- Independent source/contract review and managed Codex review found no actionable issue. A separate numerical audit rederived all 340 comparisons and all 34 adverse entries from the retained raw measurements.
- Selected owners, callers, catalog data and tests remain byte-identical to the baseline on main at `6ceb2020`; the existing dependency update is separately acknowledged, not treated as a whole-tree equivalence claim.

Whole-export source measurement on Node 26.8.1: one fixed before/candidate/candidate/before cohort, eight rows, 846,088 calls and 512 measured batch means. It uses the normal repository source loader and actual exported operations, with full boundary/control readback. No copied implementation or private-helper-only timing.

| Operation | Forward median change | Reverse median change |
| --- | ---: | ---: |
| Full 95-entry listing | −26.56% (−4.05 μs) | −24.53% (−3.66 μs) |
| Known package lookup | −31.02% | −29.96% |
| Unknown package lookup | −30.40% | −29.20% |
| Matching package/plugin identity | −30.81% | −30.23% |
| Known package, wrong plugin id | −29.78% | −29.53% |
| Foreign package, known plugin id | −30.47% | −30.18% |
| Explicit external-distribution metadata | −31.21% | −29.98% |
[...]

PR #135997, opened 2026-09-02
feat(plugins): discover provider setup from the ClawHub catalog

steipete:
## What Problem This Solves

Provider plugins could be installable on ClawHub but absent from OpenClaw setup until another core catalog entry was maintained. The app's Add provider form also bypassed provider-owned setup and could not handle installation consent or multi-step authentication.

Companion publisher change: https://github.com/openclaw/clawhub/pull/3573.

## Why This Change Was Made

Setup and plugin management now share the accepted ClawHub catalog snapshot. Provider choices come from release-bound feed hints plus installed manifests; installed owners win, and the existing hosted model catalog or small feed previews supply display hints only. The redundant compiled provider index is removed. There is no new catalog JSON, registry, operator setting, environment variable, database schema, or Gateway protocol.

A shared preparation path installs the exact advertised package and reconciles the installed plugin/provider/auth-method identity before requesting credentials. Interactive CLI, non-interactive CLI, and guided setup use that owner. Gateway setup admits one fresh metadata generation and pins it through the wizard without replacing the running Gateway's cache. New installs and staged credentials remain subject to the existing consent, cancellation, live-probe, conflict, and persistence owners.

The app uses the existing provider wizard for both manual primary setup and configuration-only Add provider. Configuration-only completion refreshes provider state without claiming a successful inference check or changing the primary model. The old inline key/direct-config path is removed from manual-primary setup and Add provider; editing an existing provider key remains supported.

Dynamic CLI flags cannot shadow core options, implicit help, or inherited object properties. CLI auth inference uses parser-registered flags; programmatic callers retain trusted installed-manifest flags. Choice IDs remain opaque values instead of being reparsed from help text, and replacement commands use terminal-safe shell quoting. Sensitive prompt values are registered for redaction before provider validation or authentication can fail, including normalized pasted values.

## User Impact

Eligible official provider releases can appear through normal ClawHub publication without a new OpenClaw provider-registration PR. This does not change publisher admission or authorize runtime behavior from catalog previews. Community packages remain explicitly installable, with their installed manifests supplying setup metadata. Bundled offline discovery, existing CLI choices, and the older manual API-key RPC remain supported.

## Evidence

- Regression-first coverage includes catalog identity drift, CLI collisions and non-interactive consent, prompted-secret redaction, setup snapshot ownership, install retention, and primary-model preservation.
- Final focused backend/CLI checks: 808 passed across 33 files. UI owner suites: 401 passed. Four mocked browser suites: 16 passed.
[...]

PR #140597, opened 2026-09-07
fix(plugins): reduce bundled catalog lookup overhead

### LINKED ISSUE #140595: Bundled catalog lookups rebuild collections and import hosted download services
## Problem

Direct lookups in the official bundled plugin catalog repeatedly construct complete row arrays and a Map even when the caller needs only one package or manifest. Importing that static lookup owner also loads hosted-catalog fetching and ClawHub artifact-download code.

## Evidence

On frozen main d2c9a798f96ff03b031180def90ad3f72ceb1f95, an isolated built probe exercising the actual package lookup and manifest accessor runs five samples of 95,000 lookups over 95 rows. Its median is 1001.76 ms. A shared raw source plus direct early-exit lookup reduces the median to 531.07 ms with the same 475,000 checksum across five samples. For scale, one 95-entry sweep saves about 0.47 ms.

Separating the static and hosted import owners reduces loaded built module URLs from 255 to 58. One paired cold-import sample measured immediate process RSS of 72.06 MiB before and 50.20 MiB after. These are isolated catalog measurements, not full-Gateway startup or memory claims. Minor-GC events increased from 1425 to 1954; no reduced collection-count or allocated-byte-total claim is made.

## Expected behavior

Use one typed raw bundled source for catalog lookups and conservative repair-target checks. Direct queries should stop at their match while preserving live row mutation, current source-reference validation, first-wins deduplication and fresh outer list arrays. Raw repair targets must retain shadowed/source-rejected hints. Hosted fetching should load through its existing async owner while preserving validation, trust, monotonicity, snapshots and freshness.

## Related context

#138312 prepared source-reference sets within filtering; this is the separate repeated collection and cold import-graph cost. Active #135997 has a distinct provider-setup change whose type re-export removal should be preserved during integration.

steipete:
Closes #140595

<details>

</details>

## What Problem This Solves

Resolves avoidable work during bundled plugin package and manifest lookups: each direct query rebuilds complete catalog collections, and the static lookup import loads hosted fetching and artifact-download implementations.

## Why This Change Was Made

Share one typed raw bundled source, validate and deduplicate live rows as queries iterate, and stop direct lookups at the matching row. Move hosted fetching behind the existing async public entrypoint and move pure ClawHub integrity helpers into a leaf module used by their real consumers. Remove the duplicate raw-target type mirror and its assertion.

Shared rows remain mutable and list APIs return fresh outer arrays. Conservative repair targets still inspect raw shadowed and source-rejected rows with current environment/configuration. Hosted validation, signed-feed monotonicity, snapshot/fallback behavior, freshness and helper/error identity retain their existing contracts.

## User Impact

Static catalog users load a smaller module graph and do less repeated lookup work. In the paired built probe, the median 95,000-lookup run improved from 1001.76 ms to 531.07 ms; one sweep of 95 rows saves about 0.47 ms. Cold import loaded 58 module URLs instead of 255.

This is an import and representation repair, not a net source-deletion claim: production is +68 lines, tests are +239, and two obsolete ratchet entries were mechanically removed. Related #138312 and active #135997 address distinct work; preserve the latter's type re-export removal during integration.

## Evidence

- 138 focused tests passed across official catalogs, tiny mutable/source-filter fixtures, conservative targets, signed envelopes and ClawHub artifact/integrity behavior.
- Complete check:changed and full build passed. Independent isolated P2 autoreview found no actionable findings; all 18 final source hashes and the unchanged lockfile match the reviewed manifest.
- The same final probe ran on both source and built outputs after normal frozen installs, with both heavy slots reserved to exclude overlapping admitted jobs. Both sides ran five samples of 95,000 real package/manifest lookups over 95 rows, with matching checksum 475,000.
- Built module URLs fell from 255 to 58. One paired cold-import sample measured 86.16 ms to 11.54 ms and immediate RSS of 72.06 MiB to 50.20 MiB. Download, client, temporary-download and hosted-loader implementations were absent from the repaired static import graph.

The cold import and RSS values are single isolated observations, not full-Gateway gains. Source bytes are not heap bytes. Minor-GC events increased from 1425 to 1954; no reduction in collection count or exact allocated-byte total is claimed.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3394,7 +3394,6 @@ src/plugins/memory-runtime.ts	1
 src/plugins/module-export.ts	2
 src/plugins/native-module-require.ts	3
 src/plugins/official-external-plugin-catalog-snapshot-store.ts	4
-src/plugins/official-external-plugin-targets.ts	1
 src/plugins/official-external-provider-endpoints.ts	1
 src/plugins/openai-compatible-embedding-provider.ts	2
 src/plugins/plugin-cache-primitives.ts	1
--- a/config/max-lines-baseline.txt
+++ b/config/max-lines-baseline.txt
@@ -747,7 +747,6 @@ src/plugins/manifest-registry.ts
 src/plugins/marketplace.test.ts
 src/plugins/marketplace.ts
 src/plugins/official-external-plugin-catalog.test.ts
-src/plugins/official-external-plugin-catalog.ts
 src/plugins/plugin-registry-snapshot.test.ts
 src/plugins/provider-runtime.test.ts
 src/plugins/provider-runtime.ts
--- a/src/claws/package-remove.ts
+++ b/src/claws/package-remove.ts
@@ -1,6 +1,6 @@
 import { coerceErrorMessage } from "@openclaw/normalization-core/error-coercion";
 import { runPluginUninstallCommand } from "../cli/plugins-uninstall-command.js";
-import { normalizeClawHubSha256Integrity } from "../infra/clawhub-artifacts.js";
+import { normalizeClawHubSha256Integrity } from "../infra/clawhub-integrity.js";
 import { resolveInstalledClawHubPlugin } from "../plugins/plugin-install-preflight.js";
 import { withPluginLifecycleLease } from "../plugins/plugin-lifecycle-lease.js";
 import {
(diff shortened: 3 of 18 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1ccfd576e77e2467c0de115286b2454551ecaf82](https://github.com/openclaw/openclaw/pull/140597/commits/1ccfd576e77e2467c0de115286b2454551ecaf82)
- Landed commit: [8a540c252002878f3c3e9d8dfb837b83c8602a78](https://github.com/openclaw/openclaw/commit/8a540c252002878f3c3e9d8dfb837b83c8602a78)

Issue #90548, opened 2026-06-05
macOS: per-port lsof port-health polling can saturate launchservicesd and trigger a WindowServer watchdog reboot


jmmas123:
## Summary

On macOS, the gateway's port-health checks spawn `lsof` **once per inspected port on every poll cycle**. On a busy machine this steady stream of short-lived process launches can saturate macOS's launch pipeline (`syspolicyd` Gatekeeper assessment + `launchservicesd` check-in) badly enough that `launchservicesd` hits its **512 dispatch-thread hard limit**. When that happens, `WindowServer` blocks on a synchronous LaunchServices XPC call, goes unresponsive, and the OS watchdog (`watchdogd`) kills it — which tears down the graphics session and looks to the user like a freeze + reboot.

I traced a hard "froze and rebooted" event on my Mac to exactly this.

## Environment

- macOS 26.5 (25F71), Apple Silicon (Mac15,10)
- openclaw/clawdbot gateway installed as a `KeepAlive` LaunchAgent (`com.clawdbot.gateway`), node, port 18789
- Package version 2026.1.23-1 (current npm `latest` 2026.1.24-3)
- Busy host: ~800 processes, 2 VMs, VPN, an IDS, several Electron apps (this matters — see caveat)

## What I observed

`dist/infra/ports-inspect.js` runs, per port:

```js
runCommandSafe([lsof, "-nP", `-iTCP:${port}`, "-sTCP:LISTEN", "-FpFcn"])
```

In the macOS unified log, the gateway spawned `lsof` like clockwork — **one invocation every ~2.2 seconds, sustained** (≈27/min, consistent with iterating the gateway's service ports across the ~60s poll interval). Each spawn forces a full launch pipeline pass:

- `syspolicyd` Gatekeeper/notarization assessment (I measured ~41 assessments/sec across the system during the incident window)
- `launchservicesd` `CHECKIN` + code-signing verification

## Root-cause chain (from Apple's own crash captures)

```
per-port lsof poll (continuous)
  → syspolicyd + launchservicesd back up
  → launchservicesd hits the 512 dispatch-thread HARD LIMIT
  → WindowServer's synchronous LaunchServices call blocks (turnstile wait on launchservicesd)
  → WindowServer main thread unresponsive 40s
  → watchdogd kills WindowServer  → graphics session torn down (apparent reboot)
```

The `WindowServer-*.userspace_watchdog_timeout.spin` report states verbatim:

```
[...]

PR #140608, opened 2026-09-07
fix(ports): reduce duplicate process probes during diagnostics

### LINKED ISSUE #140606: Unix port diagnostics repeat process metadata queries for shared PIDs
## Problem

Unix port diagnostics launch three process-metadata queries for every socket row, even when many rows or requested ports belong to the same process. This creates avoidable subprocess work during status and restart inspection.

## Reproduction and evidence

On frozen main d2c9a798f96ff03b031180def90ad3f72ceb1f95, 40 synthetic connections owned by one PID produce 120 native `ps` child processes through the actual inspection owner. Collecting metadata once per unique PID with two queries reduces this to two children. With 40 distinct PIDs, the count changes from 120 to 80.

All six native-executable component scenarios return identical complete outputs, including one PID across two ports, `ss` fallback, usernames containing spaces, and wide/combining Unicode usernames. `lsof` and `ss` counts remain unchanged. This measures subprocess count, not Gateway latency, throughput, RSS or retained memory.

## Expected behavior

Own process metadata at each public inspection call: gather unique PIDs across its socket rows, read each PID with a bounded batch, then apply primitive metadata to every original row. Preserve full usernames and command lines, ordering, partial metadata, environment projection, platform behavior and fresh metadata on subsequent calls.

## Related context

#107121 bounded per-row enrichment; #115813 batched listener snapshots. #90548 discusses broader polling pressure and is related only: this repair does not change `lsof` polling or counts.

steipete:
Closes #140606

<details>

</details>

## What Problem This Solves

Resolves repeated process launches during Unix port diagnostics when several socket rows or requested ports belong to the same PID. The old inspector runs three metadata queries for every row.

## Why This Change Was Made

Collect unique PIDs at the public inspection boundary and share primitive metadata across the original rows for that invocation. Each PID uses two sequential queries: PPID plus the complete command, and the existing user-only query so usernames containing spaces or Unicode remain intact. One bounded batch limits active metadata subprocesses to 20.

The invocation-local map expires with the call. Listener/connection ordering, per-port details, partial metadata, bind-host filtering, lsof/ss fallback, diagnostic environment projection and subsequent-call freshness remain intact.

## User Impact

A 40-row inspection with one PID launches two `ps` children instead of 120, including the same PID across two ports or `ss` fallback. With 40 distinct PIDs it launches 80 instead of 120. All socket rows and complete metadata outputs match. Production shrinks by 14 lines; test coverage grows by 110 lines.

## Evidence

- 188 focused owner/status/restart tests passed; one existing Windows-only native test was skipped on the local platform. Resource-budget regressions fail against the exact original owner because 150 launches exceed the 50-launch ceiling for both lsof and ss cases.
- All six component scenarios have complete before/after output equality with native Execa and task-owned lsof/ss/ps fixture executables. They include repeated/distinct PIDs, multiple ports, fallback, spaced usernames and wide/combining Unicode usernames. Socket-query counts are unchanged.
- The final two-query format was independently exercised on native macOS and Linux with owned synthetic processes. A 6.7 KB command and interior spaces survive, and both queries return nonzero/empty output after the process is reaped. The three task-owned Testbox leases are stopped.
- Complete check:changed and native isolated P2 autoreview passed on the final source hashes. No actionable review findings remain.

The measurements establish subprocess-count reductions, not general Gateway latency, RSS or throughput gains. Native PPID-zero and unusual account records were not manufactured; partial/zero/missing metadata and complete spaced/Unicode usernames are covered at the actual inspection owner with synthetic inputs. Related #90548 has a broader polling scope; this change does not alter it.

### DIFF
--- a/src/infra/ports-inspect.env.test.ts
+++ b/src/infra/ports-inspect.env.test.ts
@@ -60,7 +60,7 @@ if (command === 'lsof') {
 } else if (command === 'ss') {
   process.stdout.write(${JSON.stringify(ssOutput)});
 } else {
-  process.stdout.write(process.argv.includes('ppid=') ? '1\n' : process.argv.includes('user=') ? 'fixture-user\n' : 'node fixture-server\n');
+  process.stdout.write(process.argv.includes('user=') ? 'fixture-user\n' : '1 node fixture-server\n');
 }
 `,
         { mode: 0o755 },
@@ -102,9 +102,7 @@ if (command === 'lsof') {
         .map((line) => JSON.parse(line));
       expect(
         reports.map((entry) => entry.command).toSorted((left, right) => left.localeCompare(right)),
-      ).toEqual(
-        (fallback ? ["lsof", "ss", "ps", "ps", "ps"] : ["lsof", "ps", "ps", "ps"]).toSorted(),
-      );
+      ).toEqual((fallback ? ["lsof", "ss", "ps", "ps"] : ["lsof", "ps", "ps"]).toSorted());
       for (const entry of reports) {
         expect(entry, `${entry.command} inherited canary presence`).toEqual({
           command: entry.command,
--- a/src/infra/ports-inspect.ts
+++ b/src/infra/ports-inspect.ts
@@ -53,8 +53,7 @@ type UnixListenerSnapshot = {
   lsofUnavailable: boolean;
 };
 
-// Each Unix mapper starts three child processes; cap mapper slots so port
-// diagnostics cannot fan out one process batch per socket record.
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [20e4d225e1ea9c841506aece622597602842311f](https://github.com/openclaw/openclaw/pull/140608/commits/20e4d225e1ea9c841506aece622597602842311f)
- Landed commit: [cc13ef846d51f7a392be360999f0379819be4623](https://github.com/openclaw/openclaw/commit/cc13ef846d51f7a392be360999f0379819be4623)

PR #127231, opened 2026-08-21
fix(exec): isolate background result retention per process

steipete:
## What Problem This Solves

Fixes an issue where using one agent's process tool could shorten or extend another agent's completed command retention. A background command could also use the default retention until its owning agent first used the lazily loaded process tool.

Related: #127219

## Why This Change Was Made

Capture the effective cleanup duration with exec admission, including delayed approved execution, and measure each retained result's expiry from completion. The registry keeps the same process object for polling and notification receipts and maintains one unref'ed expiry timer only while finished results remain.

The global TTL setter and running-only interval are removed. The existing one-minute to three-hour limits, environment fallback, public `exec.cleanupMs` option, 50-result cap, and aggregate output policy remain. The production change adds ten net lines for the per-process duration/deadline and launch wiring; it adds no configuration or persistence schema.

## User Impact

Completed command output remains available for the owning agent's configured duration, regardless of when either agent uses the process tool. Cleanup removes retained lookup entries without invalidating already-captured terminal poll or notification receipts.

## Evidence

- The real two-agent baseline on trusted main failed after 95,003 ms: the result configured for 180,000 ms had already expired after another agent loaded 60,000 ms process defaults.
- [The integrated-head proof](https://github.com/steipete/openclaw/actions/runs/34036025154) ran both actual regressions at `69c4054c7d9ea0c49876ed42dcd36622f8762c4c`. After 95,024 ms the long-retention result remained readable and the short-retention result had expired. The real Gateway delayed-approval case passed; all 17 runtime fingerprints stayed unchanged.
- Registry coverage protects completion-relative expiry, reverse expiry order, clamp endpoints, captured poll/notification receipts after expiry, and timer retirement after deletion, scope cleanup, and reset. Current main's source-authority guards, stdin ownership, and supervisor/PTY cleanup remain intact.
- [Exact-head CI attempt 2](https://github.com/openclaw/openclaw/actions/runs/34035270997) is green, including the Discord shard and dependent CI gate. Attempt 1 hit two existing Discord backlog case timeouts. Their cause remains unknown: the voice owner/harness matched trusted main; an instrumented trusted macOS baseline and [the same historical merge's full Discord/IRC configuration groups on Linux](https://github.com/steipete/openclaw/actions/runs/34044616598) passed. Host resources, caches, scheduling, and instrumentation differ. One targeted verification of the original failing job then passed without changing source, assertions, mocks, or timeouts.
[...]

PR #140612, opened 2026-09-07
chore(tests): remove idle wait from exec retention coverage

steipete:
## What Problem This Solves

The two-agent exec-retention regression spends 95 seconds idle after both real commands have completed. It measured 95.375 seconds in [main CI](https://github.com/openclaw/openclaw/actions/runs/34070027186), while the other thirteen tests in its file totaled 514 ms.

## Why This Change Was Made

Advance the retention clock after both native `echo` processes finish successfully and the short-lived agent's lazy process tool loads. Keep the original 60-second and 180-second retention settings, tool-loading order, and final process-log assertions. Clock control includes intervals so the original global-TTL implementation's sweeper still executes in regression proof. Cleanup cancels the registry timer before restoring real timers.

Related: #127231.

## User Impact

Faster test execution with no production behavior, dependency, shard-count, or coverage reduction. The file retains all fourteen cases and now explicitly checks both native commands exited successfully. Test-only delta: +12/-18 lines; production delta: zero.

## Evidence

- Local unchanged baseline: 14 passed; retention case **95.986 s**. Candidate: 14 passed; same case **508 ms**.
- Restored candidate plus the complete process-registry suite: **41 passed**.
- Reversing the original fix's seven production-file changes makes the faster test fail at the intended assertion: the long-lived result was discarded. All seven production files were restored and their original hashes verified afterward.
- Local whole-command time: 181.73 → 68.68 s; peak RSS 2.210 → 2.221 GB. Transform overhead varied, so the scoped 95-second idle-wait removal is the attributable improvement.
- Independent P0–P2 review: no findings. Required changed checks and exact-head CI are recorded with the PR checks.

### DIFF
--- a/src/agents/agent-tools-agent-config.exec.test.ts
+++ b/src/agents/agent-tools-agent-config.exec.test.ts
@@ -74,10 +74,14 @@ describe("Agent-specific exec tool defaults", () => {
 
   afterEach(() => {
     resetProcessRegistryForTests();
+    vi.useRealTimers();
     tempDirs.cleanup();
   });
 
   it("keeps each actual exec result for its own agent retention after another process tool loads", async () => {
+    vi.useFakeTimers({
+      toFake: ["Date", "setTimeout", "clearTimeout", "setInterval", "clearInterval"],
+    });
     const config: OpenClawConfig = {
       tools: { exec: { host: "gateway", mode: "full", cleanupMs: 60_000 } },
       agents: {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5226bbf6921182ca02ab4fa8a3fc5bb08b0b9121](https://github.com/openclaw/openclaw/pull/140612/commits/5226bbf6921182ca02ab4fa8a3fc5bb08b0b9121)
- Landed commit: [fc33fa0b8917a889d69eecdd92d3371b1e86f8ad](https://github.com/openclaw/openclaw/commit/fc33fa0b8917a889d69eecdd92d3371b1e86f8ad)

PR #140613, opened 2026-09-07
improve(memory-wiki): skip link parsing for bracket-free pages

steipete:
## What Problem This Solves

Memory Wiki reads parse an entire Markdown syntax tree to find links even when a page contains no opening bracket. This adds substantial work when reading large plain-text pages by path, basename, or ID.

## Why This Change Was Made

Skip link parsing when the original text has no `[`: both supported link forms require it, and code masking or related-block removal cannot introduce it. Frontmatter validation still runs first. Move the pure link-processing code into a focused sibling module to keep the existing Markdown owner within its size limit, preserving its marker exports. Remove an unnecessary type assertion and its obsolete baseline entry.

## User Impact

Faster reads for bracket-free pages with unchanged link extraction and validation behavior. The change adds 10 net production lines, removes one assertion-baseline entry, and changes no test lines.

## Evidence

The same existing 16 MiB large-page case passed in **16.425s before and 0.153s on the final source**. This is one local case-level pair, not a full-suite or aggregate coverage claim. Transform caches differed: worker preparation was 3.962s before and 2.871s after; total command time was 27.59s and 6.41s, respectively, so the wrapper difference is not attributed solely to parsing.

All 155 existing Markdown, query, compile, lint, and OKF cases pass. The test files and fixture size are unchanged. The later cast removal emits identical runtime JavaScript to the tested split source, and the final filtered large-page case passes. Full changed-file checks, formatting, and independent review through P2 pass, including the exact shrink-only assertion-baseline update. No suppression or parser-call-count test was added.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -771,7 +771,6 @@ extensions/memory-wiki/src/compiled-cache.ts	1
 extensions/memory-wiki/src/gateway.ts	3
 extensions/memory-wiki/src/ingest.ts	2
 extensions/memory-wiki/src/log.ts	1
-extensions/memory-wiki/src/markdown.ts	1
 extensions/memory-wiki/src/mutation-coordinator.ts	1
 extensions/memory-wiki/src/okf.ts	1
 extensions/memory-wiki/src/source-sync-state.ts	1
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e243b193d14eca0e5ce3b29162fac9eb5843f5f2](https://github.com/openclaw/openclaw/pull/140613/commits/e243b193d14eca0e5ce3b29162fac9eb5843f5f2)
- Landed commit: [46174814d9b83dcf41c09626912ac7dbeaf55cb9](https://github.com/openclaw/openclaw/commit/46174814d9b83dcf41c09626912ac7dbeaf55cb9)

PR #139635, opened 2026-09-06
fix(ci): board widget e2e clicks are silently swallowed before the frame accepts input

steipete:
## What Problem This Solves

Resolves a flake where the Control UI dashboard A2UI e2e test failed in `checks-ui-e2e` on unrelated PRs and passed on rerun. Seen 2026-09-05 on #139371, #139355 (run 33993524339), and #139384 (run 33996683994): `expected +0 to be 1` polling for the `board.event` request after clicking the widget's "Refresh data" button, with the failing variant alternating between dark and light. Each failing test was as fast as its passing sibling (~0.6–1.1 s before the poll), so it is an ordering race, not a slow-runner timeout.

## Why This Change Was Made

Since #139076 (landed the same morning) a board widget's outer iframe stays `inert` with `opacity: 0` behind a skeleton until the sandbox proxy reports `sandbox-resource-loaded`, and the frame is revealed two animation frames later. The test polled only for the renderer's custom element inside the widget document, which is defined by the bundle script *before* the inner document's `load` event fires, so the click routinely runs while the outer frame is still inert.

On the CI Chromium (the runner image's `/usr/bin/chromium-browser`, Chromium 151, selected because the Playwright browser cache restores only ~2 MB) the parent document hit-tests the inert iframe element as the target: `elementsFromPoint` returns the iframe, so Playwright's actionability checks pass, and the real `pointerdown`/`mousedown`/`click` are dispatched to the iframe element in the top document and never enter the widget. Playwright's hit-target interceptor reports "done" when no event reaches the target frame, so the click is a silent no-op and no `board.event` is ever sent. Even after `inert`/opacity are cleared, Linux keeps routing pointer events to the outer element until the compositor catches up. macOS Chromium skips inert nodes in `elementsFromPoint`, so Playwright waits there instead, which is why this never reproduces locally.

#139506 (landed while this was in flight) already waits for the outer frame's reveal and performs one hover with a `:hover` poll before the click. That converts the silent no-op into a loud failure, but a single hover cannot recover: when the first pointer move is routed to the outer iframe element, nothing re-dispatches it, so the widget document never reports `:hover` and the poll times out (its own evidence notes a hover-only candidate still failed pointer entry).

[...]

PR #139506, opened 2026-09-05
fix(ci): synchronize cloud retries and board widget input

steipete:
## What Problem This Solves

Fixes intermittent CI failures in cloud-session retry and dashboard widget interaction proof. The cloud test could advance virtual time before the previous failure response scheduled its next retry. The board test could click an inner widget while Chromium was still routing input to its outer iframe.

Observed on main in [the cloud retry job](https://github.com/openclaw/openclaw/actions/runs/33994163511/job/101381925493) and [the board interaction job](https://github.com/openclaw/openclaw/actions/runs/34001319748/job/101400675353), and again in this PR's first board shard.

## Why This Change Was Made

Install the Playwright clock before application timers are created. Use the existing deferred Gateway fixture to settle each failed cloud-profile response and observe the Start button's pending-to-settled transition before advancing the next retry.

For board interaction, wait for the outer iframe's existing visible, non-inert state and native pointer entry into the inner Refresh control before the single click. This follows the existing fullscreen test's reveal boundary and retains the real browser input path.

## User Impact

More reliable CI proof of cloud retry limits, dispatch-before-send ordering, and dashboard widget actions in both themes. Product code, retry delays, timeouts, shards, and all four original scenarios are unchanged. Production LOC delta: zero; test delta: +24/-9 across the two existing files.

## Evidence

- A temporary 100 ms delay delivering cloud mock errors reproduced the missing-request timeout. Response synchronization passed both original scenarios under the same delay; the final two scenarios passed after removing the diagnostic.
- An earlier hover-only board candidate still failed native pointer entry. Waiting for the existing outer reveal boundary plus pointer entry passed both original theme cases and 12 temporary alternating-theme repetitions. The repetition was removed from the final source.
- Original retry counts/delays, placement, attachments, dispatch/send order, single board event, action payload, and scrollbar assertions remain intact.
- Managed Codex autoreview through P2 found no actionable issues in the combined change.

PR #139076, opened 2026-09-05
fix: smooth focused dashboard loading

steipete:
Closes #139072
Related: #138860, #139055

## What Problem This Solves

Fixes an issue where focused dashboards show plain loading text, flash an empty white iframe, and visibly jump from the saved widget size to the content height. The focused route also repeats session discovery before requesting its board.

## Why This Change Was Made

Use the existing themed, reduced-motion-aware loading skeleton until the trusted sandbox reports that its current inner document has loaded. A fresh render ID distinguishes replacements and stale receipts; sizing and capability-bridge initialization retain their existing owners. Reveal follows a paint opportunity at the measured size, and the hidden frame is inert. Empty and fixed-position widgets also complete; a stalled document ends in the existing visible error surface instead of an endless refresh cycle. Standard MCP App resource requests without a render ID keep their existing notification flow.

The focused document reuses the resolved session key and agent owner through its existing prepared-session path, removing one serial sessions.describe request. Canonical reloads also retain the validated agent scope through the existing connection-bound navigation handoff. Unresolved routes keep their discovery path.

## User Impact

Dashboard loading stays themed and animated until the widget can be shown at its measured size. Keyboard focus cannot enter the hidden frame. Existing scrolling, grants, updates, and interaction remain available. Resolved focus links make one fewer serial RPC.

## Evidence

A live Chromium filmstrip reproduced loading text at 1.047 s, a white iframe around 2.403–2.772 s, and the expanded widget at 2.794 s. A second live attempt stalled while connecting, before dashboard data arrived, so no production latency improvement is claimed.

Controlled production-bundle runs used the same synthetic widget, 150 ms simulated RPC delay and 200 ms HTML delay. Each set had one warm-up and five measured loads; every loaded widget's Details button worked.

| Measurement | Baseline | Candidate round 1 | Candidate round 2 |
| --- | ---: | ---: | ---: |
| Median visible-widget time | 665 ms | 529 ms | 892 ms |
| Median unfinished-frame exposure | 357 ms | 0 ms | 0 ms |
| Visible widget-content height changes during startup | 1 | 0 | 0 |
| Extra sessions.describe requests | 1 | 0 | 0 |

Host contention produced substantial outliers, including a 6.13 s candidate load. Wall-clock speed is inconclusive; the eliminated RPC and visual transition improvements are consistent across all measured runs.

[...]

PR #140626, opened 2026-09-07
test(ui): retain A2UI action-delivery failure diagnostics

steipete:
Related: #139076, #139506, #139635

## What Problem This Solves

The Control UI A2UI action test can report only that a click was followed by zero `board.event` requests, leaving no evidence of whether native input reached the widget, the renderer rejected the action, the private bridge initialized, or the document was replaced.

The historical failure occurred in [main run 33965832570](https://github.com/openclaw/openclaw/actions/runs/33965832570/job/101306012044) at `2299d11045d35ae1032bf43f75a1faaab204a231`: dark passed immediately before light timed out waiting for one event. This PR improves evidence for a recurrence; it does **not** claim to establish or repair that historical cause.

## Why This Change Was Made

Add passive, bounded observers to the existing two theme cases and print their snapshots only when a test fails. Evidence includes trusted native input paths, bridge offer/readiness/initialization acknowledgment and state-emission arrival, renderer error state and alert text, outer-frame presentation, and the selected frame's identity/lifecycle.

Keep input and transport with their existing owners. In particular, preserve current main's `clickBoardWidgetControl`, the single click, existing waits, event count/payload expectations, and scrollbar assertions. Diagnostics never start ports, wrap the frozen API, or record ticket/action payload bytes. Collection has its own two-second bound; it does not extend the action assertion's timeout.

## User Impact

No product or UI appearance change. A future failing CI run will retain enough boundary evidence to investigate the missing action instead of inviting a speculative bridge rewrite or retry. Only the existing E2E test and a test-only support file change.

## Evidence

- Original current-main dark/light cases passed before and after adding diagnostics. Focused UI E2E typechecking, lint, formatting, and independent review through P2 passed during the investigation.
- Temporary fault probes, removed from the final source, distinguished an omitted click from a real trusted click whose oversized state payload was rejected. The latter recorded `state.emit` arrival and the caught renderer error while page errors remained empty.
- A [Blacksmith Testbox replay](https://github.com/openclaw/openclaw/actions/runs/34014028436) used the original historical source/configuration/lockfile with Node 24.19.0, pnpm 12.3.4, Vitest 4.1.11, and Playwright 1.62.1. Three original dark-to-light pairs passed, as did the historical shard: 64 files and 289 tests. The job labeled 2/7 actually ran Vitest shard 2/6. The delegated command succeeded and its lease was stopped; backing Actions cleanup is not the test result.
- Shard membership and the set of 28 preceding bundled-file completions matched the saved failure. Exact concurrent interleaving, warm caches, and the original Chromium build were unavailable; the replay used Chromium 144. These passes constrain the investigation but do not prove a historical root cause or attribute recovery to #139076.
- The candidate retains the later native-input helper from #139635 rather than restoring the older hover sequence. At reviewed head `b453c838fd4211c00c81657a8d56f73224a28415`, changed-check passed in full, including core-test typechecking, i18n, formatting, dead-export scans, and scoped lint. The original dark/light E2E pair passed 2/2 (37.46 seconds including bundle preparation), and fresh isolated independent review was clean through P2. Hosted validation is tracked in [exact-head CI run 34075118948](https://github.com/openclaw/openclaw/actions/runs/34075118948).

### DIFF
--- a/ui/src/e2e/board-a2ui.e2e.test.ts
+++ b/ui/src/e2e/board-a2ui.e2e.test.ts
@@ -20,6 +20,7 @@ import {
   startControlUiE2eServer,
   type ControlUiE2eServer,
 } from "../test-helpers/control-ui-e2e.ts";
+import { installA2uiFailureDiagnostics } from "./board-a2ui.test-support.ts";
 
 const chromiumExecutablePath = resolvePlaywrightChromiumExecutablePath(chromium.executablePath());
 const chromiumAvailable = canRunPlaywrightChromium(chromiumExecutablePath);
@@ -112,7 +113,9 @@ describeControlUiE2e("Control UI dashboard A2UI", () => {
   });
 
   for (const colorScheme of ["dark", "light"] as const) {
-    it(`renders a v0.9 widget with the ${colorScheme} scrollbar theme`, async () => {
+    it(`renders a v0.9 widget with the ${colorScheme} scrollbar theme`, async ({
+      onTestFailed,
+    }) => {
       const context = await browser.newContext({
         colorScheme,
         permissions: ["local-network-access"],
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b453c838fd4211c00c81657a8d56f73224a28415](https://github.com/openclaw/openclaw/pull/140626/commits/b453c838fd4211c00c81657a8d56f73224a28415)
- Landed commit: [e3a7948de814851b077010ae06e877cc9eac6d94](https://github.com/openclaw/openclaw/commit/e3a7948de814851b077010ae06e877cc9eac6d94)

PR #140638, opened 2026-09-07
refactor(tui): test the displayed session filter

steipete:
## What Problem This Solves

Session-picker regression tests read the filter through an otherwise unused getter, keeping a test-only observation method in the component.

## Why This Change Was Made

Assert the exact displayed filter row through the existing render boundary and remove `getFilterText()`. The three migrated assertions still check typed text and clearing; selected-value assertions remain separate because displayed labels cannot recover opaque values. All tests remain. Production decreases by four lines; tests change by +4/-3.

## User Impact

No intended user-visible behavior change. Keyboard filtering, clear-before-cancel behavior, navigation, cursor markers and Unicode rendering are unchanged.

## Evidence

- Baseline and candidate picker component suites: **72/72** each, identical case inventories. All four cancellation encodings, selection, cursor, Unicode/column-width and terminal-sanitization coverage remain.
- Independent applied review: no findings through P2; pinned Pi TUI input rendering and package export boundaries inspected.
- All **29 required changed-file checks passed** on [Testbox](https://github.com/openclaw/openclaw/actions/runs/34074319075); the one-shot command returned 0 and its lease/source capsule were cleaned up.
- Additional existing PTY suite: candidate **5/8**, unchanged baseline **4/8**. Failures on both sources time out waiting for the fixture's `local ready`, before any picker command/key is sent. The baseline includes model-picker failures that do not construct this component. This does **not** claim a passing end-to-end run or a diagnosed startup cause. The separate follow-up is TUI PTY pre-readiness diagnostics, including retaining the existing fixture action log before cleanup; no timeout, retry or assertion changes are included here.

The PTY fixture uses the real TUI with a fake backend; no Gateway, provider or persistence-transport proof is claimed.

### DIFF
--- a/src/tui/components/filterable-select-list.test.ts
+++ b/src/tui/components/filterable-select-list.test.ts
@@ -1,6 +1,7 @@
 // Filterable select list tests cover keyboard filtering and cursor behavior.
 import { CURSOR_MARKER, visibleWidth } from "@earendil-works/pi-tui";
 import { describe, expect, it } from "vitest";
+import { stripAnsi } from "../../../packages/terminal-core/src/ansi.js";
 import { FilterableSelectList, type FilterableSelectItem } from "./filterable-select-list.js";
 
 type FilterableSelectListTheme = ConstructorParameters<typeof FilterableSelectList>[2];
@@ -92,7 +93,7 @@ describe("FilterableSelectList", () => {
 
     list.handleInput(query);
 
-    expect(list.getFilterText()).toBe(query);
+    expect(stripAnsi(list.render(80)[0] ?? "").trimEnd()).toBe(`>Filter: <> ${query}`);
     expect(list.getSelectedItem()?.value).toBe(expectedValue);
   });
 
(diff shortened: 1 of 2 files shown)


steipete:
Dependency-inspection gap addressed for exact head `3ecafeb6cfbc6180701f6156fa1289164808a66f`.

I personally inspected the installed, pinned `@earendil-works/pi-tui@0.84.4` implementation in `dist/components/input.js:316–375` (SHA-256 `4762edfaa75de102aabc00f8660c591f2d7de1ba6e7e212900dd81c48f231e63`). `Input.render` adds its prompt, preserves these short values without scrolling, then adds an inverse-video cursor and right padding. The component adds the themed filter label. The existing canonical ANSI stripper and `trimEnd()` remove presentation codes and padding for the exact row comparison.

A direct execution of the exact PR component with the existing test theme and width 80 returned:

```json
[
  {"input":"j","row":">Filter: <> j"},
  {"input":"k","row":">Filter: <> k"},
  {"input":"beta","row":">Filter: <> beta"},
  {"input":"","row":">Filter: <>"}
]
```

Repeated renders returned the same rows. The independent applied review separately inspected this dependency source and rendering behavior. Baseline/candidate component suites retain all 72 cases, and [exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34075720619) is now green.

This closes the review's missing dependency-inspection evidence. The separately documented pre-readiness PTY failures on both baseline and candidate remain an explicit limitation and follow-up.

steipete:
Merged via squash.

- Prepared head SHA: [3ecafeb6cfbc6180701f6156fa1289164808a66f](https://github.com/openclaw/openclaw/pull/140638/commits/3ecafeb6cfbc6180701f6156fa1289164808a66f)
- Landed commit: [d53e916b066ecf9603d075357732a992386cbf0f](https://github.com/openclaw/openclaw/commit/d53e916b066ecf9603d075357732a992386cbf0f)

PR #140640, opened 2026-09-07
fix(gateway): preserve WebSocket close causes in file logs

steipete:
## What Problem This Solves

Fixes an issue where operators investigating WebSocket disconnects could see a close code but lose the Gateway's recorded cause at the default log level.

## Why This Change Was Made

Preserve the existing `cause` and `durationMs` on the two connected-client info-level file records. Unknown causes remain omitted. Keepalive, authentication and transport behavior stay unchanged.

## User Impact

Operators can distinguish a recorded missed-pong decision from a close with no known Gateway cause. The diagnostics guide explains that `heartbeat-timeout` does not prove ping delivery or remote-peer fault.

## Evidence

The baseline failed the two extended regression cases and a temporary native WebSocket/file-log case. The candidate passed all 30 checks, including normal close with omitted cause and missed pong with `heartbeat-timeout` after the unchanged 25-second ping/50-second termination sequence. Native sockets, listener and temporary files were cleaned up. The native proof uses fixture registration; it does not attribute a browser or proxy failure.

Independent P2 review found no actionable issues. The full local changed gate passed; exact-head CI follows publication. An earlier Crabbox sync attempt returned no test evidence; the qualified proof ran locally on Node 24.20.0/macOS.

### DIFF
--- a/docs/gateway/diagnostics.md
+++ b/docs/gateway/diagnostics.md
@@ -96,6 +96,15 @@ hostnames, and local usernames.
 When a log message looks like user, chat, prompt, or tool payload text, the
 export keeps only that a message was omitted plus its byte count.
 
+## WebSocket disconnect logs
+
+Connected webchat and authenticated-user disconnects include `durationMs`
+(connection lifetime in milliseconds) in default info-level file logs. The
+`cause` field contains the Gateway's recorded close cause, when known; otherwise
+it is omitted. `heartbeat-timeout` records the Gateway's missed-pong decision.
+It does not prove that a ping reached the remote peer or that the peer caused
+the transport failure.
+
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1f81dcad69975f0fdc8041fd95fe904f7a9681dc](https://github.com/openclaw/openclaw/pull/140640/commits/1f81dcad69975f0fdc8041fd95fe904f7a9681dc)
- Landed commit: [8e41bafff836a11c328d6d7eb644f5450604d854](https://github.com/openclaw/openclaw/commit/8e41bafff836a11c328d6d7eb644f5450604d854)

PR #140644, opened 2026-09-07
refactor(searxng): test category retries through requests

steipete:
## What Problem This Solves

SearXNG category-fallback tests checked a private predicate separately from the request path, leaving some category combinations untested at the point where retries actually happen.

## Why This Change Was Made

Move the same five category inputs into the existing request tests and remove the predicate's testing export. The retry cases assert the original request followed by exactly one `general` request and retain the result checks. The non-retry cases assert one request, its category parameter (or absence), and the complete stable empty result.

The runtime predicate and call site are unchanged. Cache resets, URL/DNS validation, response parsing, cancellation and error handling stay as they are. Production decreases by one line; test lines are net neutral.

## User Impact

No intended behavior change. Regression coverage now detects a broken retry call site as well as predicate errors, while keeping all five existing category subjects.

## Evidence

- Full plugin suite: **29/29 before, 31/31 after**. The same five category subjects move from three cases to five request cases; all other 26 case names remain unchanged, including real loopback transport, cache TTL and cancellation tests.
- Fault control: limiting the runtime call site to the single `weather` category still passes the old client tests, but fails only the new `weather,news` request case. Both source versions were restored exactly afterward.
- Independent applied review: no findings through P2.
- All **25 required changed-file checks passed** on [Testbox](https://github.com/openclaw/openclaw/actions/runs/34075382357); one-shot command exit 0, lease stopped, source capsule removed.

The request-policy cases use the existing mocked endpoint boundary. Real transport cases use a local HTTP server; no authenticated or deployed SearXNG service proof is claimed, and no external API contract changes.

### DIFF
--- a/extensions/searxng/src/searxng-client.test.ts
+++ b/extensions/searxng/src/searxng-client.test.ts
@@ -93,85 +93,93 @@ describe("searxng client", () => {
     ).toEqual([{ title: "One", url: "https://example.com/1", content: "A" }]);
   });
 
-  it("retries an empty category search with general results", async () => {
-    endpointMockState.responses.push(
-      new Response(JSON.stringify({ results: [] }), { status: 200 }),
-      new Response(
-        JSON.stringify({
-          results: [
-            {
-              title: "Beijing hourly weather",
-              url: "https://example.com/weather",
-              content: "Hourly forecast",
-            },
-          ],
-        }),
-        { status: 200 },
-      ),
-    );
+  it.each(["weather", "weather,news"])(
+    "retries an empty category search with general results (%s)",
+    async (categories) => {
+      endpointMockState.responses.push(
+        new Response(JSON.stringify({ results: [] }), { status: 200 }),
+        new Response(
+          JSON.stringify({
+            results: [
+              {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via the normal native squash as [4b547c680cd2](https://github.com/openclaw/openclaw/commit/4b547c680cd2127a108c90abcf41c62359c3928f), from reviewed head `bce843fd2aade41353a977bf92008fc21fe28b7d`.

The original merge was accepted, then stopped during local reconciliation because the reported landed commit was unavailable against the observed main snapshot. Native reconciliation subsequently verified the exact landed tree. No second merge was dispatched. This is the first completion comment for that accepted attempt; the native journal retains the reconciled `merged` phase, and the owned native worktree and temporary refs were removed through `scripts/pr gc` after preserving the native evidence.

All 31 plugin cases, the retry fault control, 25 changed-file checks, and [exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34076091542) passed. Category fallback behavior, cache isolation, validation and cancellation coverage are preserved.

PR #140657, opened 2026-09-07
fix: stop sending missing env refs as catalog credentials

shakkernerd:
## What Problem This Solves

Fixes an issue where users selecting a stored MiniMax credential backed by a custom environment SecretRef would send the environment-variable name as the credential when its value was unavailable. Catalog discovery then reported a provider authentication rejection instead of the selected account being locally unavailable.

## Why This Change Was Made

Prepare actual environment-backed material for the selected profile, preserving existing cold command and worker discovery. If that material is missing, use the same canonical profile owner as other SecretRefs: it supplies the selected profile's valid published value or fails before catalog HTTP, without borrowing another account. Provider endpoints, Bearer versus `X-Api-Key` behavior, profile ordering, literal credentials, and synchronous plugin callbacks are unchanged.

This follows MiniMax's documented [OpenAI-compatible](https://platform.minimax.io/docs/api-reference/models/openai/list-models) and [Anthropic-compatible](https://platform.minimax.io/docs/api-reference/models/anthropic/list-models) model-list authentication contracts. No vendor API changes are needed.

## User Impact

Missing environment-backed credentials remain configured but unavailable, and compatible model names remain visible. Restoring the environment value makes cold commands usable again; a running Gateway uses `openclaw secrets reload` to activate the restored secret. Real uppercase literal credentials continue to work; reference names are not inferred from their spelling.

## Evidence

- Exact-head Testbox verification at `fc31ed5c2d8409e9a12b752b31e5a0d577046ac6` passed all 224 cases in ten affected and sibling files: shared auth, the real catalog worker and secret serialization, MiniMax, Chutes, OpenAI catalog contracts, Copilot, Ollama, and configured environment handling.
- All 12 added env-ref regressions fail on the original base for the intended outgoing-header mismatch. They cover `keyRef` and `tokenRef`, both catalog callbacks, published values, missing cold/unpublished material, and selected-account identity. Existing non-env, plain, OAuth, and healthy unactivated-env controls remain intact.
- Five real MiniMax HTTP controls passed: missing built-in/custom refs sent no HTTP, including with another healthy account present; a present custom ref and a genuine uppercase literal each sent the correct credential. The original base sent the custom reference name and received HTTP 401.
- The original ref-only real-worker integration test is unchanged. `pnpm openclaw models list --all --provider minimax-portal --json` with a healthy env-backed profile produced identical model JSON and the same successful authenticated request on the base and final head.
- An isolated real Gateway with a persisted env-backed token account reported it unavailable at cold startup without HTTP. Normal `config set` and `secrets reload` commands recovered the same account; full `models.list` then authenticated discovery and reported it ready. A stopped and restarted Gateway repeated that success from persisted configuration. Temporary processes, servers, state, and the comparison worktree were cleaned up.
- The complete remote changed gate passed, including core and test type graphs, typed changed-path lint, formatting, ratchets, source boundaries, and dependency guards. [Hosted CI](https://github.com/openclaw/openclaw/actions/runs/34079791458) is green, and the [exact-head ClawSweeper review](https://github.com/openclaw/openclaw/pull/140657#issuecomment-5564255604) is Platinum with no before-merge blockers.

No authenticated live-vendor result is claimed; the regression concerns requests that must never leave the process.

### DIFF
--- a/docs/auth-credential-semantics.md
+++ b/docs/auth-credential-semantics.md
@@ -91,6 +91,12 @@ catalog discovery. Configured subscription modes remain attached to direct
 credentials, and successful OAuth preparation supplies the resolved current token
 to its catalog consumer rather than the captured store's older token.
 
+Static profile SecretRefs, including custom environment refs, use the activated
+secret snapshot. An unavailable selected ref reports `unavailable` before catalog
+HTTP; its reference name is never sent as a credential or replaced with another
+profile's credential. Restore the secret and run `openclaw secrets reload` before
+retrying discovery.
+
 When every eligible OAuth candidate fails preparation, discovery reports
 `unavailable` with the attempted profile identities instead of treating the
 provider as unconfigured. Compatible prior inventory remains available. A usable
--- a/docs/auth-credential-semantics.md
+++ b/docs/auth-credential-semantics.md
@@ -91,11 +91,12 @@ catalog discovery. Configured subscription modes remain attached to direct
 credentials, and successful OAuth preparation supplies the resolved current token
 to its catalog consumer rather than the captured store's older token.
 
-Static profile SecretRefs, including custom environment refs, use the activated
-secret snapshot. An unavailable selected ref reports `unavailable` before catalog
-HTTP; its reference name is never sent as a credential or replaced with another
-profile's credential. Restore the secret and run `openclaw secrets reload` before
-retrying discovery.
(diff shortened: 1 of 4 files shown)


shakkernerd:
Fixed: catalog discovery no longer sends missing environment SecretRef names as credentials, while preserving working cold-reader authentication.

PR #140658, opened 2026-09-07
fix: keep desktop cleanup retryable after shutdown failures

steipete:
## What Problem This Solves

Fixes an issue where stopping or reopening a worker desktop could report successful cleanup after SSH termination failed, while the transport was still running and its temporary connection files had already been deleted. A filesystem cleanup failure also made later disposal attempts incorrectly succeed without removing the directory.

## Why This Change Was Made

The desktop session registry now owns final resource disposal after startup and transport teardown have joined. Failed cleanup retains the stopped session and blocks replacement until a later attempt succeeds. This removes the SSH desktop hook's duplicate startup/stop flags and disposal branches. The SSH runner permits a new stop attempt after failure, including recovery after a late close, and SSH preparation marks disposal complete only after removal succeeds.

Pending acquisitions remain registered while predecessors drain; older app launches are cancelled immediately after registration and joined even if predecessor teardown fails, and worker managers join all sibling cleanup through the existing worker stop helper before reporting failure.

Readiness markers, output-drain policy, signal escalation deadlines, observer limits, and public configuration are preserved. Owner epochs continue to fence desktop access. No persistence, protocol, dependency, or public SDK changes.

## User Impact

Desktop cleanup failures remain visible and retryable. Reopening a desktop cannot replace an unresolved owner, and temporary SSH files survive an unconfirmed shutdown. A late transport exit triggers cleanup of that exact retained owner, without cancelling a same-epoch retry.

## Evidence

- The original 48 focused tests passed on the baseline. Four added/extended regressions failed before the production change for hidden stop errors, lost resources, failed disposal retry, and cached failure after real closure.
- 115 unique owner and sibling tests pass (114-case owner/sibling run followed by the final 24-case desktop suite), covering host and paired-node desktops, SSH preparation, worker desktop and workspace tunnels, cancellation before spawn and during readiness/password reads, cleanup retry, and sibling joining.
- Real synthetic subprocess fault injection: stop rejects and retains the directory while fresh child output proves the child is still running. The test then restores signal delivery, reopens in the same owner epoch, and proves predecessor exit does not cancel the replacement before joining closure and completing cleanup.
- Real loopback OpenSSH proof: generated test keys, pinned host key, remote readiness, Unix-socket RFB forwarding, natural remote exit, and directory disposal. The fixture daemon's exit was joined and its resources were removed. The macOS SSH client emitted its TCP_NODELAY warning for the Unix socket; forwarding and exit remained successful.
- Production delta: +169 / -133, net +36. Tests: +376 / -0. Test support: +8 / -5. Docs: +7. No generated output, snapshots, lockfiles, or moved files.

- Final core typecheck, all 16 selected core test type graphs, and lint across all changed TypeScript files pass. A shadowed callback name found by the remote changed gate was corrected without changing its binding.
- Isolated independent review is clean through P2; the final follow-up was only the callback rename for lint. Exact-head required CI remains a landing gate.

### DIFF
--- a/docs/gateway/config-browser-ui-desktop.md
+++ b/docs/gateway/config-browser-ui-desktop.md
@@ -212,6 +212,13 @@ managed mode. Managed mode requires `Xtigervnc`, `tigervncpasswd`, and
 fresh temporary VNC password for each managed session, never persists it, and
 supervises both the VNC server and XFCE session.
 
+If desktop teardown fails, the Gateway retains the session's cleanup owner and
+reports the failure in its logs. A new observation retries cleanup before
+starting a replacement. For SSH-backed worker desktops, temporary connection
+files remain until the transport has closed; a late process exit triggers
+another cleanup attempt. Check the reported process or filesystem error before
+retrying an observation that cannot finish cleanup.
+
 Without managed mode, configure third-party servers to listen on loopback when
 they support it. On Linux, use loopback-only TigerVNC or `x11vnc`; GNOME Remote
 Desktop's VeNCrypt mode is not supported. On Windows, enable VNC authentication
--- a/src/gateway/desktop/attachment.test.ts
+++ b/src/gateway/desktop/attachment.test.ts
@@ -203,6 +203,106 @@ describe("RFB attachments", () => {
     await registry.stopAll();
   });
 
+  it("retains failed teardown and prevents replacement until cleanup succeeds", async () => {
+    const registry = createDesktopSessionRegistry();
+    const failure = new Error("transport still running");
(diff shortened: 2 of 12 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ed00a539c8ec7ec5979063a935377213c7ac359c](https://github.com/openclaw/openclaw/pull/140658/commits/ed00a539c8ec7ec5979063a935377213c7ac359c)
- Landed commit: [23c1fffa16c1b46ea7f6ff875ae7f3246f4116a7](https://github.com/openclaw/openclaw/commit/23c1fffa16c1b46ea7f6ff875ae7f3246f4116a7)

PR #124648, opened 2026-08-16
fix(ui): surface command-palette chat search failures

steipete:
## What Problem This Solves

Found by live Control UI QA: opening the command palette and typing any chat query rendered "No results" even when every visible session title matched. The gateway was actually rejecting `sessions.list` (store pending a doctor canonical-key migration), but the palette's `searchSessions` catch swallowed the error and the empty state rendered as if the search succeeded with zero matches — a silent failure on a default path, the worst bug class per the repo doctrine.

## Why This Change Was Made

A failed search and an empty search are different facts and must render differently. The palette now records the failure (`sessionSearchFailed`, scoped to the current request id while open, cleared on every new query) and the empty state says "Chat search failed — check the gateway logs and retry" instead of "No results". Navigation commands stay usable throughout — the search remains best-effort.

## User Impact

Operators with a broken session store (or any gateway-side search failure) now see that search is failing instead of concluding their chats are gone.

## Evidence

- Live repro on the QA gateway: `sessions.list` rejected with `SessionCanonicalKeyMigrationRequiredError` while the palette showed plain "No results" (gateway log confirms 6 consecutive rejections).
- New regression "shows a search failure instead of a false empty result" fails pre-fix (stash-verified), passes post-fix; also covers failure-state clearing + clean retry on the next keystroke.
- `node scripts/run-vitest.mjs run ui/src/components/command-palette.test.ts` — 8/8 pass.
- `pnpm ui:i18n:baseline` + `pnpm ui:i18n:verify` clean (new `palette.searchFailed` key, English source only per ui/AGENTS.md).
- tsgo ui lane green.
- LOC: prod +18/−2, test +26.
- Codex autoreview (gpt-5.6-sol, high): clean, 0.99.

PR #140662, opened 2026-09-07
fix(ui): palette hides chat search failures beside navigation results

steipete:
Related: #124648

<details>

This branch is in the main repository and remains editable by maintainers.

</details>

## What Problem This Solves

Fixes an issue where users searching chats in the command palette would see no failure notice when chat search failed and the query also matched navigation or catalog entries. For example, searching for `plugins` left the Plugins destination visible but silently omitted the failed Chats search.

## Why This Change Was Made

Render the recorded chat-search failure in the existing search-status region, alongside partial and incomplete transcript-search notices. Remove the empty-state-only failure branch and reuse the existing translated message. Successful results remain available, and failed empty searches do not display a contradictory “No results” message.

## User Impact

Users can distinguish a failed chat search from a successful search with no matching chats, while continuing to use navigation and catalog results. Editing the query clears the failure and retries normally.

## Evidence

The existing failure-and-retry regression now covers both an unmatched query and `plugins`. Before the fix, the matching-query case fails because the error is absent; afterward, all 44 palette and catalog tests pass.

Synthetic Chromium runs through the real Control UI with a mocked Gateway confirm the error changes from absent to visible. Clicking Plugins reaches `/settings/plugins` both before and after the fix. These runs used isolated browser contexts and a source-mode UI server, without a live provider or operator Gateway.

Owning UI test types, full-file lint/format/style checks, and `check:changed` pass. Independent P2 review found no actionable issues. Integration onto current main is patch-identical, with no upstream changes to the palette owner, its tests, or search helpers.

Production: 8 added / 10 removed lines (net −2). Tests: 5 added / 4 removed lines (net +1).

Before:

Failed chat search hidden behind navigation and catalog results

After:

Chat search failure visible above usable navigation and catalog results

### DIFF
--- a/ui/src/components/command-palette.test.ts
+++ b/ui/src/components/command-palette.test.ts
@@ -874,22 +874,23 @@ describe("CommandPalette lifecycle", () => {
     expect(palette.textContent).toContain("Unrelated title");
   });
 
-  it("shows a search failure instead of a false empty result", async () => {
+  it.each(["zzz-unmatched", "plugins"])("shows a chat search failure for %s", async (query) => {
     const { gateway } = createGateway(true);
     const list = vi
       .fn<ApplicationContext<RouteId>["sessions"]["list"]>()
       .mockRejectedValueOnce(new Error("store needs doctor migration"))
       .mockResolvedValueOnce(createSessionResult("agent:main:zz", "Recovered chat"));
     const { palette } = await mountPalette(createContext(gateway, list));
-    // The query matches no navigation item, so a swallowed search failure
-    // would render the plain "No results" empty state.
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [10ca2af64a7a21a90e019b0a204268d2f0b22c38](https://github.com/openclaw/openclaw/pull/140662/commits/10ca2af64a7a21a90e019b0a204268d2f0b22c38)
- Landed commit: [3b610841b0f7f0d52f6b6cb7f99bbfb2efeafc29](https://github.com/openclaw/openclaw/commit/3b610841b0f7f0d52f6b6cb7f99bbfb2efeafc29)

PR #135359, opened 2026-09-01
fix(matrix): preserve text for unknown message types

teddytennant:
Matrix messages with inherited JavaScript property names as `msgtype` could lose filename-shaped text from quoted replies, thread roots, pending history, and message-read results.

Check declared message types at the shared media-kind lookup. Unknown types stay ordinary text, while the five declared attachment kinds keep their existing caption and filename behavior. The regression tests cover the inherited names without type casts. This retains the original contributor's change and ancestry.

Validation on the repaired commit:

- Real disposable Matrix events reproduced the failure on pinned main and recovered with the same inputs, including encrypted send/decrypt, reply and thread context, pending history, and message reads.
- Real uploads covered all five declared media kinds, caption and filename combinations, scoped unavailable downloads, and files below and above the configured size limit.
- All 55 focused tests passed; the same tests produced five expected regression failures on the original code.
- Independent source review and fresh source-blind acceptance completed. Full CI remains required before landing.

The size case covers the announced-size gate. Media proof covers classification and download behavior. The sticker control uses the declared message content type. Private runtime evidence is retained for review.

PR #140663, opened 2026-09-07
fix(matrix): reject prototype-named msgtypes in media kind lookup

xialonglee:
## What Problem This Solves

Fixes an issue where a remote Matrix user sending a message whose `msgtype` collides with an `Object.prototype` property name (`__proto__`, `constructor`, `toString`, `hasOwnProperty`) would have that text message rendered as a corrupted media marker — e.g. `[matrix [object Object] attachment]` or `[matrix function Object() { [native code] } attachment]` — instead of the message body. Any unencrypted Matrix room peer controls `msgtype`, so this is remotely triggerable on the default inbound message path (`monitor/handler-helpers.ts`) and the message summary path (`actions/summary.ts`).

## Why This Change Was Made

`resolveMatrixMediaKind` indexed the `MATRIX_MEDIA_KINDS` plain-object lookup table with the remote-controlled `msgtype` string. Prototype-named keys resolve to inherited `Object.prototype` values, which are truthy and bypass the `?? null` fallback, so the function's declared `MatrixMessageAttachmentKind` return type was a lie and downstream label formatting stringified the inherited value. The fix guards the lookup with `Object.hasOwn(MATRIX_MEDIA_KINDS, key)` so only own `m.*` keys resolve, with a comment recording the remote-input contract. Regression coverage is a new `media-text.test.ts` exercising all four prototype-name collisions through both the attachment resolver and the end-to-end text formatter.

## User Impact

Matrix text messages with prototype-named `msgtype` values now render as plain text. Legitimate `m.audio`/`m.file`/`m.image`/`m.sticker`/`m.video` attachments are unaffected.

### Before fix
Running the new `media-text.test.ts` against the pre-fix `media-text.ts` (restored from `origin/main`):

```
Tests  6 failed (6)
[test] failed 1 Vitest shard in 4.28s
```

Reproduced symptom: `msgtype: "__proto__"` renders `[matrix [object Object] attachment]`; `msgtype: "constructor"` renders `[matrix function Object() { [native code] } attachment]`.

### After fix
```
media-text.test.ts        Tests  6 passed (6)
summary.test.ts + media.test.ts + handler.media-failure.test.ts  Tests 22 passed (22)
```

- `node scripts/run-oxlint.mjs --tsconfig extensions/tsconfig.json <changed files>`: 0 warnings, 0 errors
- `oxfmt --check`: clean
- `tsgo -p tsconfig.extensions.json --incremental`: pass
- Local structured review (`openclaw-pr-review`): PASSED, no findings; best-fix verdict `best`; bug introduced by #127417 (clear)

### CI proof: prototype-named msgtypes through the real Matrix gateway flow

Live Matrix QA lane on a real disposable homeserver (tuwunel): hostile `msgtype` events (`toString`, `__proto__`) are sent raw (the QA text client pins `m.text`), become thread roots, and are ingested by the SUT gateway through the normal matrix-js-sdk sync monitor path; the assertion scans the exact text the mock provider recorded for the agent turn, so the thread-starter summary under test is real gateway output, not a mocked projection. A legitimate `m.image` thread root is the control:

- Before (main): `f9f4aa93`
- After (PR head): `6f7efd2c`
- Proof run: https://github.com/xialonglee/openclaw/actions/runs/34109358545
- Artifact download URLs: https://github.com/xialonglee/openclaw/actions/runs/34109358545/artifacts/10014004517 (before), https://github.com/xialonglee/openclaw/actions/runs/34109358545/artifacts/10014093040 (after)

Verdict markers (9/9 checks pass, `[proof] VERDICT: PASS`):

```
# BEFORE (main f9f4aa93) — lane exited 1, bug reproduced for the intended reason:
scene=before-lane status=pass detail=lane exited 1 (bug reproduced)
scene=before-corrupted status=pass detail=corrupted media markers reached the recorded model input
scene=before-corrupted-marker status=pass detail=[matrix function toString() { [native code] } attachment]
scene=before-corrupted-marker-proto status=pass detail=[matrix [object Object] attachment]

# AFTER (PR head 6f7efd2c) — lane exited 0, same scenario passes cleanly:
scene=after-lane status=pass detail=lane exited 0 (fix verified)
scene=after-thread-context status=pass detail=scene=thread-context status=pass
scene=after-control-image status=pass detail=scene=control-image status=pass
scene=after-clean status=pass detail=no corrupted marker after fix
```

The recorded model input for the after head kept all three probe bodies as plain text (`provider-requests count=3`, hostile bodies + control caption) and preserved the `[matrix image attachment]` control marker.

### DIFF
--- a/extensions/matrix/src/matrix/media-text.test.ts
+++ b/extensions/matrix/src/matrix/media-text.test.ts
@@ -1,0 +1,23 @@
+// Matrix tests cover media attachment text behavior.
+import { describe, expect, it } from "vitest";
+import { formatMatrixMessageText, resolveMatrixMessageAttachment } from "./media-text.js";
+
+// Remote peers control msgtype; names that collide with Object.prototype keys
+// must stay plain text, not become inherited lookup values misclassified as media.
+describe("resolveMatrixMessageAttachment", () => {
+  it.each(["constructor", "toString", "hasOwnProperty", "__proto__"])(
+    "does not treat Object.prototype names as media kinds (%s)",
+    (msgtype) => {
+      expect(resolveMatrixMessageAttachment({ body: "hi", msgtype })).toBeUndefined();
+    },
+  );
+});
+
+describe("formatMatrixMessageText", () => {
+  it.each(["constructor", "__proto__"])(
+    "keeps prototype-named msgtypes as plain text (%s)",
+    (msgtype) => {
+      expect(formatMatrixMessageText({ body: "hi", msgtype })).toBe("hi");
+    },
+  );
+});
--- a/extensions/matrix/src/matrix/media-text.ts
+++ b/extensions/matrix/src/matrix/media-text.ts
@@ -20,7 +20,10 @@ const MATRIX_MEDIA_KINDS: Record<string, MatrixMessageAttachmentKind> = {
 };
 
 function resolveMatrixMediaKind(msgtype: string | undefined): MatrixMessageAttachmentKind | null {
-  return MATRIX_MEDIA_KINDS[msgtype ?? ""] ?? null;
+  const key = msgtype ?? "";
+  // Remote events control msgtype; prototype-named keys must not resolve to
+  // inherited Object.prototype values (they would misclassify text as media).
+  return Object.hasOwn(MATRIX_MEDIA_KINDS, key) ? MATRIX_MEDIA_KINDS[key]! : null;
 }
 
 function resolveMatrixMediaLabel(


xialonglee:
Closing as superseded: the fix is already on main via #135359 (commit 4023edcbf194, "fix(matrix): preserve text for unknown message types"). That change applies the same `Object.hasOwn` guard at the shared media-kind lookup so prototype-named msgtypes (e.g. `constructor`, `toString`, `__proto__`) are treated as plain text, and it retains this PR's regression coverage. Rebasing this branch onto current main leaves no effective diff, so keeping it open would only add a redundant commit for maintainers. Thanks for the fix!

PR #140669, opened 2026-09-07
fix(memory): avoid duplicate snapshot allocation during flush preparation

### LINKED ISSUE #140668: Memory-flush preparation duplicates an already-owned bounded snapshot
## Problem

Memory-flush preparation deep-clones an already-owned bounded transcript snapshot when building its detached working session. Large snapshots therefore allocate a second copy of their normalized message graph during preparation.

## Evidence

The actual preparation path, using isolated canonical SQLite transcripts, returns identical expected context and stored event bytes before and after a private owned-snapshot acquisition path. In the 56 MiB serialized fixture, sampled JavaScript allocation falls from 207.37 to 147.10 MiB per preparation, removing about 60 MiB of transient allocation. The 4 MiB fixture falls from 21.85 to 17.26 MiB.

Timing is mixed: near-cap median preparation time changes from 90.22 to 116.98 ms with wide candidate batches, while the medium fixture improves. These consecutive observations do not establish a speed improvement. Both variants release held heap after the returned result is dropped, so this is not a leak report. Native SQLite allocation is outside the sampling.

## Expected behavior

Let the owner acquire a fresh bounded read, normalize its branch, and privately adopt those entries into a detached session. Public construction from caller-owned entries must still clone. Keep both managers/index passes, admission, readiness, cancellation, truncation rejection, cwd behavior and persistent transcript isolation unchanged.

## Scope

This removes one redundant transient clone. It adds no schema, configuration, retention, persistent projection or caller-owned adoption flag. Existing navigation and model-only acquisition optimizations concern separate work.

steipete:
Closes #140668

<details>

</details>

## What Problem This Solves

Resolves redundant transient allocation when memory-flush preparation clones the freshly read, already-owned bounded transcript before creating its detached working session.

## Why This Change Was Made

Add an acquisition factory that performs the existing bounded read, obtains the normalized header and branch, and privately adopts those owned entries into the detached manager. Public `fromEntries` still clones caller-owned input before using the same private constructor path. There is no public no-copy flag.

Both managers and index passes remain: the first normalizes opaque parents and retained cuts, while the second discards persistence and bounded-reader state. Admission, readiness, aborts, truncation rejection, cwd handling and detached run identity retain their existing contracts.

## User Impact

Large memory-flush snapshots require less redundant JavaScript allocation during preparation. The 56 MiB serialized fixture falls from 207.37 to 147.10 MiB sampled allocation per preparation, about 60 MiB less. The 4 MiB fixture falls from 21.85 to 17.26 MiB. These input sizes are serialized budgets, not heap-size claims.

Timing is mixed: small median 1.15 to 1.41 ms, medium 12.70 to 10.85 ms, and near-cap 90.22 to 116.98 ms with a wide candidate range. No general speed improvement is claimed. Both variants release held heap after the returned payload is released, so this removes a transient copy rather than repairing a leak.

## Evidence

- All 48 focused tests passed, covering detached preparation, bounded/general manager compatibility and completed/interrupted memory inference. Public frozen-input copying, admission/stale admission, aborts, rewrites and truncation checks remain covered.
- Nested mutation, append and compaction of the prepared manager preserve the original raw SQLite event bytes. Opaque-parent cuts, retained tool pairs and empty/missing/legacy headers retain their expected behavior.
- The real source-owner probe used canonical isolated SQLite state without mocks or live Gateway/provider access. All 12 matched runs preserve expected context and stored event hashes.
- Complete check:changed and isolated native P2 autoreview passed on the final three source hashes. No package/barrel/lazy-import boundary changed; a separate full build was not run locally.

Allocation uses Inspector sampling and excludes native SQLite allocation. Near-cap endpoint RSS was lower in the observed batches, but no peak or fixed whole-process memory guarantee is claimed. Both retained-heap paths return within 0.03 MiB of their starting values after release. Production grows by 14 lines to establish the explicit ownership path; tests grow by 113 lines.

### DIFF
--- a/src/agents/sessions/session-manager.ts
+++ b/src/agents/sessions/session-manager.ts
@@ -135,6 +135,19 @@ export class SessionManager extends SessionManagerBranching {
     });
   }
 
+  /** Consumes a fresh bounded read as a detached branch, without copying its owned payloads. */
+  static openDetachedBounded(
+    target: SessionTranscriptRuntimeTarget,
+    options: Parameters<typeof SessionManager.openBounded>[1],
+  ): SessionManager {
+    const source = SessionManager.openBounded(target, options);
+    // Normalize opaque parents and retained cuts before discarding persistence and bounded state.
+    return SessionManager.fromOwnedEntries(
+      [source.getHeader(), ...source.getBranch()],
+      source.getCwd(),
+    );
+  }
+
   /** Detached model view: selected payloads plus lightweight ancestry, never raw replay evidence. */
   static openModelContext(
     target: SessionTranscriptRuntimeTarget,
@@ -230,7 +243,14 @@ export class SessionManager extends SessionManagerBranching {
   }
 
   static fromEntries(entries: readonly unknown[], cwdOverride?: string): SessionManager {
-    const fileEntries = structuredClone(entries) as FileEntry[];
+    return SessionManager.fromOwnedEntries(structuredClone(entries), cwdOverride);
+  }
+
+  private static fromOwnedEntries(
+    entries: readonly unknown[],
+    cwdOverride?: string,
+  ): SessionManager {
+    const fileEntries = entries as FileEntry[];
     const header = fileEntries.find(
       (entry) => typeof entry === "object" && entry !== null && entry.type === "session",
     );
--- a/src/auto-reply/reply/memory-flush-session.test.ts
+++ b/src/auto-reply/reply/memory-flush-session.test.ts
@@ -123,9 +123,12 @@ it.each([false, true])(
       expect(checkpoint.sessionPersistence).toBe("detached");
       const first = checkpoint.sessionManager.getBranch()[0];
       if (first) {
+        if (first.type === "message" && first.message.role === "user") {
+          first.message.content = "Checkpoint-only edit";
+        }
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2fcbd68857321c8dcf904cd223d3571836fc1d1e](https://github.com/openclaw/openclaw/pull/140669/commits/2fcbd68857321c8dcf904cd223d3571836fc1d1e)
- Landed commit: [acba0c658dc9167611752408c22362935dbc9d19](https://github.com/openclaw/openclaw/commit/acba0c658dc9167611752408c22362935dbc9d19)

VACInc:
Integration follow-up landed in https://github.com/openclaw/openclaw/pull/138984, commit b0ae5b6793e5e1e2ba55b7e1d632e15e0fd48d00.

The new detached bounded reader could lose rewritten context when selected side-entry payloads were reconstructed without their omitted leaf controls. The merged repair routes detached snapshots through the shared selected-path constructor while preserving this PR's owned-payload/no-extra-copy optimization.

The reset/no-reset regressions now cover detached reads, detached append isolation, subsequent persisted turns, and branching. This is a complementary integration repair, not a reversal of the allocation optimization.

PR #140675, opened 2026-09-07
fix(ui): dashboard pane shows unstyled "Loading…" text while the board loads

steipete:
## What Problem This Solves

Fixes an issue where users opening a session on its dashboard face (`/dashboard/<agent>/<session>`) would see a bare, unstyled "Loading…" string stuck to the top-left of an otherwise empty pane until the Gateway answered `board.get`. On slower connections or larger boards that sat there long enough to look broken, and it stood out next to the rest of the Control UI, which shows shimmer skeletons for pending surfaces.

## Why This Change Was Made

The dashboard face's pending state rendered through a `rail-empty` class that no stylesheet defines, so the label had no layout, color, or spacing at all. Rather than style a text label, this adds a `board` variant to the shared panel loading skeleton (a tab strip over a 12-column widget grid mirroring `board.css`, collapsing to a single column under 560px like the real grid) and renders it while the board snapshot is pending. The load-error branch gets a centered, styled alert. The standalone fullscreen board document and the dashboard side-panel lazy placeholder switch to the same variant so every dashboard loading state looks alike.

## User Impact

Opening a session dashboard now shows a board-shaped shimmer immediately, in both the expanded face and the side-panel layout, until the widgets arrive. Load failures render as a readable centered message instead of raw text. Nothing changes once the board has loaded.

## Evidence

Before/after on the expanded dashboard face against the mocked Gateway with `board.get` held (fixture data only):

Before: bare Loading text at the top-left of an empty pane

After: board-shaped shimmer skeleton

- New `ui/src/pages/chat/chat-pane-board.loading.test.ts` asserts the skeleton renders while the snapshot is pending and the styled alert renders on load failure. Both tests fail on the previous code (`expected undefined to be 'board'`; alert class missing).
- `panel-loading-skeleton.test.ts` now covers the new variant (structure plus shimmer-primitive parity with `base.css`); `chat-pane-embedded-panels.test.ts` updated for the dashboard placeholder variant.
- `node scripts/check-changed.mjs` green (typecheck core/UI, i18n catalog, dead-export scan, stylelint, oxlint). Codex autoreview scoped-clean at P2.

Known gap: the still capture shows the skeleton blocks but not the animated shimmer band; that is the shared `.skeleton::after` animation already used by the other panel skeletons.

### DIFF
--- a/ui/src/components/board/board-document.ts
+++ b/ui/src/components/board/board-document.ts
@@ -262,7 +262,7 @@ export class OpenClawBoardDocument extends OpenClawLightDomElement {
 
   private renderState() {
     if (this.documentState === "loading") {
-      return renderPanelLoadingSkeleton("discussion", t("common.loading"));
+      return renderPanelLoadingSkeleton("board", t("common.loading"));
     }
     if (this.documentState === "missing-session") {
       return html`<div class="board-document__state" role="status">
--- a/ui/src/components/panel-loading-skeleton.test.ts
+++ b/ui/src/components/panel-loading-skeleton.test.ts
@@ -7,6 +7,7 @@ import {
 } from "./panel-loading-skeleton.ts";
 
 const variants = [
+  "board",
   "browser",
   "chat",
   "desktop",
--- a/ui/src/components/panel-loading-skeleton.ts
+++ b/ui/src/components/panel-loading-skeleton.ts
@@ -3,6 +3,7 @@ import { property } from "lit/decorators.js";
 import { OpenClawLitElement } from "../lit/openclaw-element.ts";
 
 export type PanelLoadingSkeletonVariant =
+  | "board"
   | "browser"
   | "chat"
   | "desktop"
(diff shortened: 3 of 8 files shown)


steipete:
Re the two "before merge" items: both are about the reviewer not being able to fetch the attached captures, not about the code. Maintainer note for the record: I captured and inspected both PNGs before uploading. They come from the mocked-Gateway Playwright harness (`installMockGateway` with `board.get` held), so they contain fixture data only (session "Main", owner label, mock nav), no real sessions, identifiers, or model routes. The "before" shot reproduces the unstyled `Loading…` text at the top-left of the empty pane; the "after" shot shows the board skeleton in the same layout. Treating these as addressed for landing.

steipete:
Merged via squash.

- Prepared head SHA: [e60209cb76c39d2f0b09410c8424a0a938c29769](https://github.com/openclaw/openclaw/pull/140675/commits/e60209cb76c39d2f0b09410c8424a0a938c29769)
- Landed commit: [238369c981bf9c5dcb32d0c7553616b768629e27](https://github.com/openclaw/openclaw/commit/238369c981bf9c5dcb32d0c7553616b768629e27)

PR #140678, opened 2026-09-07
fix(file-transfer): roll back bound overwrites interrupted mid-write

zenglingbiao:
## Summary

`extensions/file-transfer` bound overwrites (preflight-approved, `expectedBinding` pinned to device+inode) destroyed the original host file before writing the replacement: `truncate(0)` then a single `writeFile`. Any mid-write failure (ENOSPC/EFBIG) left the target truncated or half-written with no recovery. The overwrite now snapshots the original prefix, extends the tail first, touches original bytes last, and rolls the prefix + size back on failure — the same shape as the already-merged `src/agents/host-file-write.ts` in-place overwrite.

## What Problem This Solves

When a file-transfer client writes with `overwrite: true` against a preflight binding, `writeBoundTarget` opens the approved target with `r+` and immediately truncates it to zero bytes before writing the new payload. If the write dies partway — disk full, file-size limit, quota — the original content is already gone. The operation fails *and* the pre-existing host file is permanently destroyed, which is the worst possible outcome for an overwrite of a user file.

- Trigger: `handleFileWrite({ overwrite: true, expectedBinding, ... })` on an existing file, with the write failing mid-copy (ENOSPC/EFBIG/quota).
- Symptom: the API reports an error, but the target file is left empty or half-overwritten; the user's original bytes are unrecoverable.
- Blast radius: every bound overwrite of an existing file on a nearly-full filesystem.

**Code evidence**: in `extensions/file-transfer/src/node-host/file-write.ts`, `writeBoundTarget` ran `await handle.truncate(0); await handle.writeFile(input.buffer)` with no snapshot and no rollback. The sibling implementation `src/agents/host-file-write.ts` (`overwriteHostFileInPlace`, merged earlier for the same bug shape) keeps the original prefix and restores it on failure; this write point was missed.

## Root Cause

- Introduced by commit `424521a3d39` (#129762, 2026-08-26), which added the preflight-binding fast path and overwrote in place via `truncate(0)` + `writeFile`.
- Violated invariant: an in-place overwrite must not destroy bytes it cannot restore. `truncate(0)` violates it up front — after it, any failure leaves zero or partial bytes.
- The binding pins device+inode, so the usual tmp+rename atomic swap is impossible here; the fix has to be an in-place ordered write with rollback.

## Why This Fix

- Mirrors the proven `overwriteHostFileInPlace` ordering from `src/agents/host-file-write.ts` (same repo, same bug class, already reviewed and merged): read the original prefix (capped at `min(payload, currentSize)` ≤ the 16 MB payload cap), extend the tail beyond the original size first, overwrite the prefix region last, then shrink-truncate if needed.
- Failure ordering makes rollback exact: if the tail extension fails, original bytes were never touched (just truncate back to `currentSize`); if the prefix write fails, the snapshotted prefix is rewritten and the size restored.
- Rejected alternative: tmp+rename would break the device+inode binding that `writeBoundTarget` exists to enforce. Whole-file backup copies are unbounded (host file can exceed memory); the prefix cap works because at most `min(payload, currentSize)` bytes are ever endangered.
- Best-effort rollback (`.catch(() => undefined)`) matches the reference implementation: a failed restore must not mask the original write error.

## User Impact

- Affected operation: agent-driven file-transfer overwrites of existing host files (the preflight → bound-write flow used by the file-transfer node host).
- After the fix, a failed overwrite returns the same error but leaves the original file byte-for-byte intact; users no longer lose the pre-existing file when a disk fills up mid-write.

## Evidence

- **Before**: on `main`, a real `EFBIG` injected mid-overwrite (child process under `ulimit -f 4`, original 3072 bytes, payload 6144 bytes) left the target at 4096 bytes of new payload — the original content was destroyed (`FAIL: original file destroyed by failed overwrite`).
- **After**: same setup on the fix branch — the write still fails with `EFBIG`, but the target is rolled back to the original 3072 bytes (`PASS: original file content fully preserved after failed overwrite`).

### Before (on main)

```bash
$ node --import tsx proof.mjs
original bytes: 3072 | payload bytes: 6144
child overwrite error: EFBIG
after bytes: 4096 | first byte: 62
FAIL: original file destroyed by failed overwrite (data loss)
```

### After (with fix)

```bash
$ node --import tsx proof.mjs
original bytes: 3072 | payload bytes: 6144
child overwrite error: EFBIG
after bytes: 3072 | first byte: 61
PASS: original file content fully preserved after failed overwrite
```

## Tests and Validation

- Regression tests added in `extensions/file-transfer/src/node-host/file-write.test.ts`:
  - `restores the original file when the tail extension runs out of space` — fails bound writes at/after the original size; asserts the original bytes survive.
  - `restores the original prefix when a shrinking overwrite fails mid-write` — fails after half the prefix landed; asserts the original bytes survive.
- `node scripts/run-vitest.mjs extensions/file-transfer/src/node-host/file-write` → 33 passed.
- `pnpm check` passed.
- `git diff --check` clean.


jalehman:
Merged via squash.

- Prepared head SHA: [575bdf8069187657a533d74435cf67702591b1ce](https://github.com/openclaw/openclaw/pull/140678/commits/575bdf8069187657a533d74435cf67702591b1ce)
- Landed commit: [f34f142683ce26e16ef548c1d0ee16cdbda3d3b4](https://github.com/openclaw/openclaw/commit/f34f142683ce26e16ef548c1d0ee16cdbda3d3b4)

PR #140682, opened 2026-09-07
fix(google-meet): serialize caption state without crashing on __soy circular JSON

### LINKED ISSUE #140455: google-meet 2026.9.2: agent voice broken on current Meet — circular-JSON in-call crash + audio-routing verification
# Bug report: @openclaw/google-meet 2026.9.2 — Meet-DOM/UI incompatibilities break agent voice sessions

**Environment**
- OpenClaw 2026.9.2, `@openclaw/google-meet` 2026.9.2 (global extension)
- macOS (Apple Silicon), Chrome via `chrome-node` transport, `mode: agent`
- Google Meet current web UI (uses Google Closure/Soy `__soy` DOM objects)
- Audio: BlackHole 2ch virtual device (verified working via sox loopback)

**Summary:** An agent joins the Meet call successfully but stays silent. Three distinct Meet-compatibility defects in `google-meet-platform-adapter-*.js` block the audio path. Fixes/workarounds below; #1 is a clean fix.

---

## Bug 1 (CRITICAL, clean fix) — in-call inspection crashes on circular JSON

The injected in-call inspection script ends with `return JSON.stringify({...})`. One of the serialized fields (caption/transcript extraction) holds a live Meet DOM node. Google Meet's DOM attaches a `__soy` property whose object closes a cycle back to the node, so `JSON.stringify` throws:

```
TypeError: Converting circular structure to JSON
  --> starting at object with constructor 'HTMLDivElement'
  |     property '__soy' -> object with constructor '_.AGx'
  --- property 'node' closes the circle
```

**Effect:** `page.evaluate` returns HTTP 500; the plugin can never confirm in-call (`inCall` reported `no` despite the agent being a participant), never establishes the audio bridge, and reports `manualAction: browser-control-unavailable`. `recover-tab` fails with the same error.

**Fix (verified):** make the final serialization circular-safe and drop DOM nodes. At the end of the in-call inspection object in `google-meet-platform-adapter-*.js` (~line 507):

```diff
-  });
+  }, (function(){var seen=new WeakSet();return function(k,v){if(v&&typeof v==='object'){if(typeof v.nodeType==='number')return undefined;if(seen.has(v))return undefined;seen.add(v);}return v;};})());
```

After this, `inCall` is correctly detected. (Better long-term: never place raw DOM nodes into the returned object — extract `.textContent` for caption fields.)

## Bug 2 (HIGH) — audio-routing verification depends on a fragile Meet Settings-UI scrape

`routeMeetAudioInput()` finds the correct virtual mic via `enumerateDevices()` (`audioInputDeviceLabel: "BlackHole 2ch (Virtual)"`), but then only sets `audioInputRouted = true` if `selectedMicrophoneLabel()` matches — and `selectedMicrophoneLabel()` relies on `findMicrophoneDeviceControl()` scraping Meet's Settings→Audio panel, which returns nothing on the current Meet UI. So `audioInputRouted`/`audioOutputRouted` stay `false` even when BlackHole 2ch is correctly selected as both mic and speaker, which keeps the mic muted and blocks the bridge (`meet-audio-choice-required`).

**Workaround applied:** trust a recognized virtual device from `enumerateDevices()` instead of the UI scrape (set `audioInputRouted`/`audioOutputRouted` true when a recognized device family is present). A robust fix should verify selection via the active `getUserMedia` track / `sinkId`, not the settings DOM.

## Bug 3 (HIGH, unresolved) — realtime bridge never starts; audio-choice check is non-deterministic

Even with Bugs 1–2 patched and both Meet devices on BlackHole 2ch:
- `googlemeet status` reports `audioInputRouted: True`, `audioOutputRouted: True`, `manualAction: None`, `meet output routed: yes`.
- but `googlemeet test-speech` on the **same** session returns `speechBlockedReason: meet-audio-choice-required`.
- `audio bridge: none`, `last input/output: 0 bytes`, `provider connected: unknown`, and no realtime connection appears in gateway logs.
- A related log line recurs: `unknown session key "agent:meet-consult:main"`.

The per-action re-inspection returning different routing answers points to remaining Meet-DOM assumptions. Net result: the realtime provider is never connected and no audio flows, so the agent is a silent participant.

---

## Repro
1. Configure a `meet-consult`-style agent, `chrome-node` transport, `mode: agent`, BlackHole 2ch.
2. `openclaw googlemeet create` / `join` a Meet space.
3. Observe: agent joins (visible as a participant) but is silent; `googlemeet doctor` shows `audio bridge: none`, `in call: no` (pre-Bug-1-fix) or `meet-audio-choice-required` / no bridge (post-fix).

## Impact
Agent voice mode in Google Meet is non-functional on current Meet with plugin 2026.9.2. BlackHole and OpenClaw config are not at fault (loopback verified).

gaoanze888:
## What Problem This Solves

Closes #140455 (Bug 1 — CRITICAL, clean fix).

Google Meet's DOM attaches a `__soy` property whose object closes a cycle back to the node. The in-call status script (`meetStatusScript`) stored the live DOM node in each visible caption row for lifecycle tracking, then passed those rows into `JSON.stringify` via `recentTranscript` — throwing:

```
TypeError: Converting circular structure to JSON
  --> starting at object with constructor 'HTMLDivElement'
  |     property '__soy' -> object with constructor '_.AGx'
  --- property 'node' closes the circle
```

**Effect:** `page.evaluate` returns HTTP 500; the plugin can never confirm in-call (`inCall` reported `no` despite the agent being a participant), never establishes the audio bridge, and reports `manualAction: browser-control-unavailable`.

## Root Cause

In `scrapeCaptions()`, each visible caption row is pushed as `{ ...row, node: region }` where `region` is a live DOM element. When `recentTranscript = lines.slice(-5)` copies these rows (including the `node` property) into the serialized payload, `JSON.stringify` encounters the `__soy` → `node` cycle and throws.

## Fix

**1. Root cause — strip the live DOM node from `recentTranscript` entries:**

```diff
- recentTranscript = lines.slice(-5);
+ // Visible caption rows carry a live DOM node for lifecycle tracking;
+ // strip it so JSON.stringify never encounters a __soy circular reference. (#140455)
+ recentTranscript = lines.slice(-5).map((entry) => ({
+   at: entry.at,
+   speaker: entry.speaker,
+   text: entry.text,
+ }));
```

This keeps only the serializable `{ at, speaker, text }` fields and drops the `node`, `seenAt`, etc. The `node` property remains on the internal `captionState.visible` array for lifecycle tracking — only the serialized copy is cleaned.

**2. Defensive — add a circular-safe replacer to `JSON.stringify`:**

```diff
- return JSON.stringify({ ... });
+ return JSON.stringify({ ... }, (function(){
+   var seen = new WeakSet();
+   return function(k, v) {
+     if (v && typeof v === 'object') {
+       if (typeof v.nodeType === 'number') return undefined;  // drop DOM nodes
+       if (seen.has(v)) return undefined;                      // break cycles
+       seen.add(v);
+     }
+     return v;
+   };
+ })());
```

This is a belt-and-suspenders guard that handles any future edge case where a DOM node or circular reference might sneak into the serialized payload.

### Regression test

Added a test in `google-meet-platform-adapter.test.ts` that:
- Creates a mock caption region with `__soy = { node: <self> }` (circular reference) and `nodeType: 1` (DOM element)
- Enables caption capture (`captureCaptions: true`) and in-call state
- Calls `meetStatusScript` in a sandboxed VM context
- Verifies `JSON.parse(result)` succeeds, `inCall` is `true`, and `recentTranscript` entries have no `node` property

**Before the fix (test catches the bug):**
```
FAIL: TypeError: Converting circular structure to JSON
```

**After the fix:**
```
✓ serializes caption state without crashing on __soy circular references (#140455)
Test Files  1 passed (1)
Tests  6 passed (6)
```

### All existing tests pass

The existing 5 tests (caption capture, audio routing) continue to pass with no regression.

### DIFF
--- a/extensions/google-meet/src/transports/google-meet-page-scripts.ts
+++ b/extensions/google-meet/src/transports/google-meet-page-scripts.ts
@@ -483,7 +483,13 @@ export function meetStatusScript(params: {
     lastCaptionAt = last?.at;
     lastCaptionSpeaker = last?.speaker;
     lastCaptionText = last?.text;
-    recentTranscript = lines.slice(-5);
+    // Visible caption rows carry a live DOM node for lifecycle tracking;
+    // strip it so JSON.stringify never encounters a __soy circular reference. (#140455)
+    recentTranscript = lines.slice(-5).map((entry) => ({
+      at: entry.at,
+      speaker: entry.speaker,
+      text: entry.text,
+    }));
   }
   const lobbyWaiting = !inCall && /asking to be let in|you.?ll join when someone lets you in|waiting to be let in|ask to join/i.test(pageText);
   const leaveReason = !inCall && /you left the meeting|you.?ve left the meeting|removed from the meeting|you were removed|call ended|meeting ended/i.test(pageText)
@@ -530,7 +536,7 @@ export function meetStatusScript(params: {
     title: document.title,
     url: pageUrl,
     notes
-  });
+  }, (function(){var seen=new WeakSet();return function(k,v){if(v&&typeof v==='object'){if(typeof v.nodeType==='number')return undefined;if(seen.has(v))return undefined;seen.add(v);}return v;};})());
 }`;
 }
 
--- a/extensions/google-meet/src/transports/google-meet-platform-adapter.test.ts
+++ b/extensions/google-meet/src/transports/google-meet-platform-adapter.test.ts
@@ -210,3 +210,96 @@ describe("GOOGLE_MEET_PLATFORM_ADAPTER audio routing", () => {
     expect(MeetingPlatformAdapter.isRealtimeRouteReady("agent", health)).toBe(false);
   });
 });
+
+describe("meetStatusScript circular JSON safety", () => {
+  it("serializes caption state without crashing on __soy circular references (#140455)", async () => {
+    // Google Meet DOM nodes carry a __soy property whose object closes a cycle
+    // back to the node, causing JSON.stringify to throw “Converting circular
+    // structure to JSON”.
+    const captionRegion: Record<string, unknown> = {
+      nodeType: 1,
+      textContent: "Alice\nHello world",
+      innerText: "Alice\nHello world",
+      getAttribute: () => "Captions",
+    };
+    captionRegion["__soy"] = { node: captionRegion };
+
+    const leave = pageNode("Leave call");
+    const media = {
+      sinkId: "",
+      setSinkId: vi.fn(async () => {}),
+    };
+    const document = {
+      body: { textContent: "" },
+      title: "Meet",
+      querySelector() {
+        return null;
+      },
+      querySelectorAll(selector: string) {
+        if (selector === "button") {
+          return [leave];
+        }
+        if (selector === "input") {
+          return [];
+        }
+        if (selector === "audio, video") {
+          return [media];
+        }
+        if (selector.includes("aption") || selector.includes("aria-live")) {
+          return [captionRegion];
+        }
+        if (selector.includes("button") || selector.includes('[role="')) {
+          return [leave];
+        }
+        return [];
+      },
+    };
+
+    const result = await runInNewContext(
+      `(${meetStatusScript({
+        allowMicrophone: true,
+        autoJoin: false,
+        captureCaptions: true,
+        guestName: "OpenClaw Agent",
+      })})()`,
+      {
+        Event: globalThis.Event,
+        JSON,
+        String,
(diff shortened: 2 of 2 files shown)


gaoanze888:
Closing to free open-PR slots. This fix is blocked on real-environment proof (Google Meet/iMessage/Telegram/Feishu/etc) that I cannot produce from a dev machine. Will reopen when the environment is available. Code is verified locally; CI was green where it ran.

PR #140698, opened 2026-09-07
fix(agents): keep tool schema ordering deterministic for prompt caching

steipete:
<details>
<summary>Additional instructions</summary>

Maintainers can update this branch in the OpenClaw repository.

</details>

## What Problem This Solves

Fixes an issue where users switching message channels or changing tool discovery order would lose a stable prompt-cache prefix even when the available capabilities stayed identical. Message action enums and native Ollama tool lists could serialize in different orders between requests.

## Why This Change Was Made

Sort and deduplicate the final message action list with the helper already used by its description, preserving allowlist filtering and the existing empty-intersection fallback. Native Ollama now uses the shared tool-name sorter through its existing public `openclaw/plugin-sdk/provider-transport-runtime` import.

## User Impact

Equivalent capabilities produce stable message-tool schema bytes and stable native Ollama tool payloads across discovery-order changes. Action membership and tool descriptors are preserved. Production code changes total +4/-3 lines (net +1); no new configuration, storage, dependency, or SDK surface is introduced.

## Evidence

Both regressions failed before the production fixes: message schemas differed in action enum order (three failing cases, including policy fallback), and reversed Ollama descriptors produced different serialized request `tools` (one failing case).

```text
node scripts/run-vitest.mjs src/agents/tools/message-tool-discovery.test.ts extensions/ollama/src/stream-runtime.test.ts extensions/ollama/src/stream.test.ts extensions/ollama/src/stream-plain-text-tool-call.transport.test.ts
Test Files: 4 passed
Tests: 234 passed (3 message discovery + 231 Ollama)
[test] passed 2 Vitest shards in 94.84s
```

After test fixture typing was corrected, the message discovery suite was rerun: 3/3 passed.

[...]

PR #140705, opened 2026-09-07
fix(ollama): preserve prompt cache across tool catalog reordering

steipete:
## What Problem This Solves

Fixes an issue where local Ollama users could pay repeated prompt-processing costs when tool discovery returned an unchanged catalog in a different order.

## Why This Change Was Made

Sort native Ollama tool definitions with the existing prompt-cache stability helper before constructing the request. This matches the other provider adapters and preserves every tool definition and the caller's original array.

## User Impact

An unchanged tool catalog retains a stable prompt prefix across discovery-order changes. Actual changes to tool descriptions or schemas still reach the model.

## Evidence

- The regression fails on the original implementation because reversed discovery order changes the outbound request; the complete native stream suite passes with the fix (191 tests).
- Full changed-file checks and independent review pass.
- Native Ollama 0.33.3, `qwen3.5:4b-q4_K_M`, 32K context, eight synthetic tools, identical prompt and sampling controls. Two fresh-server arms each submitted caller order `A, A, B, B, A`; the fixed arm canonicalized both orders. All 10 requests returned the expected exact tool call. Native request captures matched every submitted body, and both owned runtimes were cleaned up.

At the first changed-order request (call 3):

| Native metric | Original order | Stable order |
| --- | ---: | ---: |
| Prompt tokens | 6,026 | 6,026 |
| Cached prompt tokens | 0 | 6,022 |
| Newly processed tokens | 6,026 | 4 |
| Prompt processing | 3.126 s | 0.038 s |
| Request wall time | 6.007 s | 0.931 s |

This is one synthetic run on one model, not a general speedup claim. The original arm reused its saved cache when it returned to order A on call 5; not every reorder necessarily misses. Cold-load and whole-arm times are excluded because background source scanning and filesystem warming confounded them.

Direct production-adapter validation also passes: two real streaming requests through `createOllamaStreamFn`, preserving the existing native live case's custom provider prefix, tool-schema normalization, keep-alive, completion, and usage assertions. Both server-captured request bodies exactly match the adapter payloads. The warm request reported 331 cached / 4 uncached tokens. Source fingerprints were unchanged and all owned processes were cleaned up.

The ordinary native Vitest invocation was attempted twice and hit the existing startup watchdog before sending any model request. The diagnostic run located the wait in shared compiled-subprocess preparation (8,529 inputs / 4,268 outputs), before collection of this test. Neither source nor test deadlines were changed; both failed runs were retained and their owned processes cleaned up.

### DIFF
--- a/docs/reference/prompt-caching.md
+++ b/docs/reference/prompt-caching.md
@@ -188,6 +188,7 @@ If you see unexpected `cacheWrite` spikes after a config or workspace change, ch
 
 - Delivery instructions live after the system-prompt cache boundary. Native Codex carries the current delivery and target policy in late turn context, so alternating delivery modes does not rebuild its static prompt or message tool catalog when the available capabilities remain unchanged. Actual capability changes still update the catalog.
 - Bundled MCP tool catalogs are sorted deterministically (by server name, then tool name) before tool registration, so `listTools()` order changes do not churn the tools block and bust prompt-cache prefixes.
+- Native Ollama requests sort tool definitions by name, so callers returning the same catalog in a different order do not change its prompt prefix. Changed tool descriptions or schemas still update the request.
 - Legacy sessions with persisted image blocks keep the **3 most recent completed turns** intact (counting all completed turns, not just image-bearing ones). Older already-processed image blocks are replaced with a text marker so image-heavy follow-ups do not keep re-sending large stale payloads.
 
 ## Tuning patterns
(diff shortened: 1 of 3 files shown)


steipete:
The native Ollama ordering fix is already on main through #140698, commit d448f2bbb552c184d5cf13a83c227070e72b2638. I verified that it imports the same shared helper and sorts at the same `extractOllamaTools` projection boundary; it also includes the corresponding request-order regression and documentation.

Closing this overlapping PR as superseded. The native cache measurements and direct-adapter evidence in this PR remain useful supporting evidence for the landed behavior.

PR #140720, opened 2026-09-07
improve: reduce Google tool-history preparation overhead

steipete:
<details>

Keep **Allow edits from maintainers** enabled where applicable.

</details>

## What Problem This Solves

Preparing Google requests with tool-call history does unnecessary work even when earlier thought signatures cannot be reused. This adds local request-building cost for unsigned Gemini 2.5 calls, foreign-provider history, and calls with no prior signature to replay.

## Why This Change Was Made

Reuse the already prepared argument record when constructing a replay key, and construct that key only when a managed call can record its own signature or consult an eligible replay map. Signature bytes, per-assistant staging, argument identity, route checks, and the separate signed-parts policy remain unchanged.

## User Impact

Google request preparation does less work for these histories, with no intended change to request contents or configuration. This is a local request-building improvement; it does not establish faster provider responses or Gateway turns.

## Evidence

282 tests passed across four complete Google conversion/transport test files; the normal changed-file gate passed all 28 stages. Independent P2 review reported no findings (0.98 confidence). The formatted diff adds 4 production lines and 106 test lines; the added tests protect preserved behavior rather than assert implementation counters.

A fixed 18-history corpus exercised both actual request builders, each with and without JSON serialization, in baseline/candidate/candidate/baseline order. Each row/mode used eight warmups and 24 batches of 16 calls. All 56,448 returned wire outputs matched independently authored full expectations; 64 deduplicated full JSON payload definitions plus every return's reference were retained. Full-object equality, original argument-record references, own undefined fields, and unchanged frozen inputs were checked in-process; serialized evidence cannot independently reconstruct JavaScript identity.

Selected medians of batch means, milliseconds per call (forward / reverse pair):

| Managed history | Builder before → after | Builder + JSON before → after |
| --- | --- | --- |
| Unsigned Gemini 2.5, stored JSON arguments | 0.2020 → 0.0991 / 0.2122 → 0.1016 | 0.2265 → 0.1252 / 0.2485 → 0.1236 |
| Foreign history | 0.4644 → 0.0390 / 0.5113 → 0.0384 | 0.5623 → 0.1364 / 0.6049 → 0.1351 |
| Empty replay map | 0.4639 → 0.0344 / 0.4986 → 0.0349 | 0.5635 → 0.1375 / 0.6185 → 0.1315 |
| Mixed malformed arguments | 0.0926 → 0.0624 / 0.0938 → 0.0615 | 0.0966 → 0.0727 / 0.0970 → 0.0748 |

Results are not uniformly faster: 30/72 median, 34/72 mean, and 35/72 maximum batch-mean comparisons were adverse. Reverse-pair tiny Vertex/alias medians regressed 22.28–38.83% (0.00127–0.00158 ms); assistant staging rose 2.94% for builder-only. Signed-parts controls ranged from 4.27% faster to 18.04% slower. The largest mean regression was tiny Vertex builder, 0.004920 → 0.010631 ms (+116.05%); the largest maximum batch-mean regression was tiny Gemini 2.5 with JSON, 0.007536 → 0.141320 ms (+1775.16%). Warmup median/mean/maximum comparisons were adverse in 24/28/31 of 72 cases, and first warmups in 29/72; no warmup is claimed as a cold start, and batch maxima are not individual-call tail latency.

Whole proof-process wall time fell 13.83% / 14.74%, including imports, checking, and journaling. Memory did not improve: kernel child maximum RSS rose 3.27% / 3.00%, sampled process-tree peak 2.88% / 2.89%, and end-workload RSS 3.19% / 2.84%. All measured children exited successfully and closed, and the candidate source was restored. No external Google requests, provider-latency, Gateway, or memory-benefit claim is made.

### DIFF
--- a/packages/ai/src/providers/google-messages.ts
+++ b/packages/ai/src/providers/google-messages.ts
@@ -50,16 +50,12 @@ function sanitizeGeminiThoughtSignature(value: string | undefined): string | und
     : undefined;
 }
 
-function toolCallThoughtSignatureReplayKey(block: {
-  id: string;
-  name: string;
-  arguments: unknown;
-}): string {
-  return [
-    block.id,
-    block.name,
-    stableStringifyGoogleToolCallValue(coerceTransportToolCallArguments(block.arguments)),
-  ].join("\u0000");
+function toolCallThoughtSignatureReplayKey(
+  id: string,
+  name: string,
+  args: Record<string, unknown>,
+): string {
+  return [id, name, stableStringifyGoogleToolCallValue(args)].join("\u0000");
 }
 
 export function requiresGoogleToolCallId(modelId: string): boolean {
@@ -192,7 +188,15 @@ export function projectGoogleMessages(params: {
           const ownSignature = isSameProviderAndModel
             ? signature(block.thoughtSignature)
             : undefined;
-          const replayKey = managed ? toolCallThoughtSignatureReplayKey(block) : "";
+          // Keys serve managed signature recording or a possible earlier-turn replay lookup.
+          const replayKey =
+            managed &&
+            (ownSignature ||
+              (params.requiresToolCallSignature &&
+                isSameProviderAndModel &&
+                replaySignatures.size > 0))
+              ? toolCallThoughtSignatureReplayKey(block.id, block.name, args)
+              : "";
           if (managed && ownSignature) {
             nextReplaySignatures.set(replayKey, ownSignature);
           }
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b6554e3b1f44a2f9c41d09bcfeea2b44017540d0](https://github.com/openclaw/openclaw/pull/140720/commits/b6554e3b1f44a2f9c41d09bcfeea2b44017540d0)
- Landed commit: [961d31789073b44fe2936b3260d10311f6707655](https://github.com/openclaw/openclaw/commit/961d31789073b44fe2936b3260d10311f6707655)

PR #140618, opened 2026-09-07
refactor(cli): use shared startup policy for routed commands

steipete:
Closes #140611
Related: #140480

## What Problem This Solves

Maintainer-requested CLI startup cleanup: route-first dispatch has a separate plugin-preload policy path that no real command enables. Mocked route tests exercise that unused override instead of the catalog's actual behavior, making startup policy harder to maintain consistently.

## Why This Change Was Made

Remove the route-only preload evaluator and override fields, and rely on the existing shared startup policy and bootstrap owner. Replace synthetic route-policy tests with real route selection and argument parsing while retaining plugin-loader and Commander coverage. Production source is +1/-20, net -19 lines; tests are +68/-87, net -19.

## User Impact

No intended user-visible behavior change. Config readiness still precedes plugin activation where required; passive commands, help/version laziness, structured failures and JSON stdout retain their contracts. Config inspection, migrations, metadata discovery, activation and build-time artifact generation remain distinct. No public CLI/config, persistence, dependency or build-policy changes, and no performance claim.

## Evidence

- 176 focused tests passed across route dispatch, route definitions, startup policy/bootstrap, Commander preaction and plugin registry loading.
- Full `pnpm build` passed, including the CLI bootstrap import guard, SDK exports and generated startup metadata.
- Configured `node scripts/check-changed.mjs` passed on Testbox: https://github.com/openclaw/openclaw/actions/runs/34072965912. The native delegated result was exit 0; the one-shot lease stopped successfully.
- Eight isolated built CLI cases passed: help/version with invalid config, identical route-first/Commander JSON config reads, plugin listing, config-unset dry-run, invalid config and invalid log-level failures. All config files remained byte-identical. Plugin listing and dry-run may create command-owned SQLite state; no broader read-only claim is made.
- Independent autoreview was scoped-clean through P2; `git diff --check` passed.

The local proof used synthetic HOME/state/config directories. Live Gateway/channel/provider actions and performance benchmarking were not run.

PR #140724, opened 2026-09-07
refactor(cli): prepare routed commands once before startup

### LINKED ISSUE #140721: Prepare CLI route arguments and startup facts once
## Summary

Simplify CLI startup by preparing route arguments once and removing redundant startup wrappers.

## Problem

After #140618, a successful route still validates/parses the same argv three times through generated route objects. A startup-context wrapper also reconstructs invocation facts already owned by the route or Commander caller. This creates unnecessary runtime structure and type-erased dispatch while tests preserve internal wrapper shapes.

## Proposed solution

Bind typed parser results to a deferred action in the existing command-definition owner, remove the generated route-spec module, and pass each caller's authoritative command path directly to the shared startup policy. Preserve unsupported-argument fallback, lazy imports, configuration-before-action ordering and structured output.

## Scope and evidence

Maintainer-requested behavior-neutral follow-up. No public CLI/config, persistence, dependency or plugin-activation policy change. The reviewed candidate removes 98 production lines and 209 test lines while keeping real entrypoint coverage. Focused tests, full build, changed checks, isolated CLI cases, independent review and live CLI queries against a disposable Gateway passed.

Related: #140618

steipete:
Closes #140721
Related: #140618

## What Problem This Solves

Maintainer-requested follow-up to the CLI startup cleanup: fast-route execution still parses the same arguments three times through generated route wrappers, and startup-context resolution reconstructs command facts the caller already owns. The extra runtime layer also requires type-erased dispatch and wrapper-shaped tests.

## Why This Change Was Made

The existing command-definition owner now binds typed parser results to a deferred action once. Remove the generated route-spec module and redundant startup-context/preparation wrappers; route-first and Commander callers pass their authoritative command paths directly to the existing shared policy. Tests exercise the real route entrypoint, preserving payload, fallback, ordering, error and cold-import coverage.

Production source is +45/-143, net -98 lines; tests +96/-305, net -209. The deleted module's one unsafe-assertion allowance is also removed, with no baseline expansion.

## User Impact

No intended public CLI/config, persistence, dependency or plugin-activation policy change. Unsupported arguments still reach Commander before startup, command imports/actions stay deferred until readiness completes, and action failures propagate rather than starting a second dispatch. JSON stdout and configuration-before-plugin ordering are preserved. No latency or exact microtask-equivalence claim is made.

## Evidence

- 180 focused tests passed across 12 CLI files; the final 30-test route suite passed after adding action-failure coverage and retaining the direct guard-path assertion.
- Full `pnpm build` passed, including bootstrap import restrictions, runtime postbuild, SDK exports and startup metadata.
- Final `node scripts/check-changed.mjs` passed on Testbox, including core/core-test types, full core/script lint, dead exports and runtime import cycles: https://github.com/openclaw/openclaw/actions/runs/34078375937. The initial obsolete-allowance ratchet failure was corrected by the exact shrink-only baseline change.
- Eight isolated built CLI cases passed: help/version, route-first/Commander JSON config reads, plugin listing, dry-run mutation and structured failures, with unchanged config bytes.
- Live proof used an unmocked disposable loopback Gateway with isolated HOME/state and no configured channels or inference providers. Routed `gateway health --json`, Commander fallback health, `health --json`, `status --json`, and `gateway call health --json` succeeded with one JSON document and no stderr. An invalid RPC returned one JSON failure with exit 1. Status reached the fixture's exact endpoint, stable health fields matched across dispatch paths, and config bytes stayed unchanged. Shutdown left no listener or process-group members.
- Final independent autoreview was scoped-clean through P2; `git diff --check` passed.

Live proof covers CLI/Gateway interaction, not provider inference or real channel sends. Plugin listing and dry-run retain their existing command-owned SQLite state creation.


steipete:
Merged via squash.

- Prepared head SHA: [f47f8cc6c2fe790e0244c385745019415440d197](https://github.com/openclaw/openclaw/pull/140724/commits/f47f8cc6c2fe790e0244c385745019415440d197)
- Landed commit: [a30572a99f91ca45b4dd0dc615ee3325723be244](https://github.com/openclaw/openclaw/commit/a30572a99f91ca45b4dd0dc615ee3325723be244)

Issue #140681, opened 2026-09-07
[Bug]: memory-core: since 2026.8.1, memory.search spends ~2.3 s per query booting a throwaway KNN child process — KNN itself is ~70 ms (was in-process in 2026.7.1)


justinkirklin-gif:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

Since 2026.8.1, every `memory.search` call spawns a fresh `memory-search-knn.child.js` process whose module load alone takes ~2.3 s, so server-side search is ~3 s median even though the KNN query itself takes ~70 ms and the embedding call ~0.3–0.7 s.

### Steps to reproduce

1. Run OpenClaw 2026.9.2 with `memory.search.provider=openai`, `sources=["memory"]`, sqlite-vec available (defaults otherwise). Index here: 13,965 chunks, `text-embedding-3-small` (1536-d), DB 1.0 GB.
2. Call `memory.search` (any query) and read the gateway journal line `res ✓ memory.search NNNms` — 2.9–3.5 s median over 10 runs, p95 3.2–4.7 s.
3. Time the child alone with empty stdin (it exits on EOF):
```
cd /usr/lib/node_modules/openclaw
time node dist/extensions/memory-core/memory-search-knn.child.js < /dev/null   # ≈2.3 s
time node -e ""                                                                  # 0.03 s
```
4. Run the same KNN SQL (`vec_distance_cosine`, k=48, oversample 8) in-process against the same DB: ~68 ms.

### Expected behavior

In 2026.7.1 the KNN (`vec_distance_cosine`) ran in-process (`dist/manager-*.js`); a search cost roughly one embedding round-trip plus tens of ms. Expected today: ~0.6 s per search on this box.

### Actual behavior

Each search spawns `node dist/extensions/memory-core/memory-search-knn.child.js` via `runVectorKnnInSubprocess` (`dist/extensions/memory-core/manager-runtime.js`). Per search:
- child lifetime 2.9 s, ~4 s CPU (multi-threaded startup), 109 MB read
- the KNN SQL it runs takes 68 ms in-process
- the child costs 2.27 s with empty stdin — pure module load: it side-effect-imports `memory-core-host-engine-foundation-*.js` → `memory-search-*.js` (2.19 s alone). strace: 3,967 `openat`, 1,366 distinct .js files, 147 `package.json` reads. CPU profile: `package_json_reader` ~750 ms, GC ~300 ms, zod schemas ~120 ms.
[...]

PR #140730, opened 2026-09-07
fix(memory): avoid repeated vector search startup delays

NianJiuZst:
Closes #140681

## What Problem This Solves

Fixes an issue where users repeatedly searching memory wait for unrelated configuration, schema and query-builder modules to load in each native vector-search child.

## Why This Change Was Made

Use one private host entrypoint for the existing SQLite opener, sqlite-vec loader and UTF-16 truncation helper. Both the child and its query owner use that entrypoint. SQL, isolation, cancellation, admission and index ownership stay with their existing implementations. The entrypoint is private-local and has no published TypeScript declaration; it adds no supported public Plugin SDK contract or dependency.

## User Impact

Repeated semantic memory search returns the same results with less child startup overhead. The change preserves the contributor's repair and merges the tested main baseline to keep dependencies matched.

## Evidence

Matched Linux proof used one idle machine, Node 24.19.0, SQLite 3.53.3, a normally published 4,097-row synthetic index, fixed 32-dimensional local embeddings, two alternating warmups and ten measured calls per build. Baseline source: 91c864ca19b2ffcad0dd82d80f471192ec701e18. Candidate source tree: 9581344fc64f5a8b4707392c8f9a60aeaf7c0ca9, exactly preserved by commit 8569f6de53370c852ede72855d083eb0794ed88f.

| Measured median | Baseline | Candidate |
| --- | ---: | ---: |
| Exact native child, uninstrumented |443.123 ms|60.508 ms|
| Real Gateway memory.search server response |478.5 ms|126 ms|
| Complete Gateway CLI call |1193.843 ms|845.074 ms|

The child improves 86.3%. All complete child and Gateway results match, including order, scores, paths, line ranges, text and provenance. Gateway results have zero text scores and positive vector scores, so this exercises semantic retrieval. Separate import traces load 1,916 versus 151 modules; the candidate loads no Kysely or TypeBox modules. Embedding handler times are retained separately and are not the complete HTTP round-trip.

The regression fails on baseline for `KNN child loaded unrelated runtime: kysely` and passes on the candidate. All 56 memory owner tests pass, including real WAL/source-filter queries, cancellation/reaping, retained admission after cleanup timeout, concurrent reindex publication, fallback/limit behavior and oversized UTF-16 snippets. All 30 SQLite opener tests and 14 vector-loader/redaction tests pass. Full build, canonical generated-export check and formatting pass. An actual package archive includes the private runtime entrypoint and excludes its declaration. Independent code review found no actionable P0–P2 defects.

Independent behavioral validation measured an 85.98% native improvement and matched all 17 paired mutation cases, including committed/uncommitted WAL, intersecting model/source filters, Unicode snippets and both fallback decisions. Its fresh Gateway follow-up measured a 500 ms to 94 ms server median with identical results. The validator leaves internal implementation and unavailable operator-control clauses to the separate source review and owner tests; native Windows execution is not claimed.

The restored baseline initially lacked a generated bundled-plugin alias, causing an unrelated startup migration warning. The normal staging owner restored that alias without changing the built CLI or child bytes. The Gateway table above is the repeated, warning-free comparison after this correction; all earlier attempts remain retained. Exact-head hosted CI and the final landing review remain pending.

Full private command output, build/source identities, module traces and original failures are retained for the final review. No production data, external embedding provider or Telegram credentials were used for this proof.


obviyus:
Thanks @NianJiuZst!

Merged as 439a4310ef3858036b0547a045cca91351331a81.

**Proof:** real Gateway `memory.search` and the exact native child used the same normally published 4,097-row synthetic index. Ten alternating measured calls per build returned identical complete results.

- **Bug:** each vector query loaded unrelated configuration, schema and query-builder modules.
- **Fix:** one private host entrypoint supplies the existing SQLite, vector-loader and text helpers.

| Measured median | Before | After |
| --- | ---: | ---: |
| Native child |443.123 ms|60.508 ms|
| Gateway server response |478.5 ms|126 ms|

The child improved 86.3%. Independent behavior validation confirmed an 85.98% improvement and matching WAL, model/source filters, Unicode snippets and both fallback decisions. All 100 focused tests, build/package checks, final ClawSweeper review and exact-head CI passed.

PR #140740, opened 2026-09-07
fix(cli): preserve plugin manifests when builds fail

zenglingbiao:
## What Problem This Solves

A failed `openclaw plugins build` could truncate the user's hand-maintained `package.json`, leaving invalid JSON and breaking the plugin project.

## Solution

Publish the complete package manifest through the existing atomic replacement helper. Preserve the file mode and all parent-directory permission bits, including setgid and sticky bits. A failed rename reports an error without falling back to an in-place copy. Resolve a selected symlink project root for publication while preserving entry validation and displayed paths.

## User Impact

A failed package-manifest write preserves the original file. After the filesystem problem is resolved, the same build can succeed. Successful JSON output, `--check`, initialization, and adjacent plugin-manifest publication retain their behavior.

## Evidence

- Real CLI under a file-size limit: the original publication code truncated a 5,164-byte manifest to 2,048 bytes of invalid JSON. The repaired runtime preserved all 5,164 bytes through both physical and symlink project roots.
- Partial disk-full writes and rename permission failures preserve bytes, file mode `02640`, directory mode `03770`, and remove owned temporary files. Successful retries generate both manifests.
- Staged-file symlink, hardlink, file replacement, and directory replacement are rejected without modifying unrelated targets. Entries whose ownership can no longer be verified remain protected instead of being deleted.
- Fresh independent CLI checks cover relative and absolute entries through symlink roots, failures, retries, modes, checks, and initialization. Adjacent-manifest failure remains visible; no transaction across both manifests is promised.
- All 28 focused authoring tests passed, including the symlink-root regression. The two interrupted-publication cases fail against the original publication code for the expected reasons.
- Formatting, lint, source review, build, type checks, and [CI for this head](https://github.com/openclaw/openclaw/actions/runs/34087634188) passed. Runtime proof used the compiled merge artifact from that run.

### DIFF
--- a/src/cli/plugins-authoring-command.test.ts
+++ b/src/cli/plugins-authoring-command.test.ts
@@ -687,7 +687,7 @@ describe("plugin authoring commands", () => {
       const originalPackage = fs.readFileSync(packagePath);
       const originalMode = fs.statSync(packagePath).mode & 0o7777;
       const originalDirectoryMode = fs.statSync(tmpDir).mode & 0o7777;
-      const originalEntries = fs.readdirSync(tmpDir).sort();
+      const originalEntries = fs.readdirSync(tmpDir).toSorted();
       const error = Object.assign(new Error("publication failed"), {
         code: failure === "write" ? "ENOSPC" : "EPERM",
       });
@@ -734,7 +734,7 @@ describe("plugin authoring commands", () => {
         expect(fs.readFileSync(packagePath)).toEqual(originalPackage);
         expect(fs.statSync(packagePath).mode & 0o7777).toBe(originalMode);
         expect(fs.statSync(tmpDir).mode & 0o7777).toBe(originalDirectoryMode);
-        expect(fs.readdirSync(tmpDir).sort()).toEqual(originalEntries);
+        expect(fs.readdirSync(tmpDir).toSorted()).toEqual(originalEntries);
         expect(log).not.toHaveBeenCalled();
       } finally {
         vi.restoreAllMocks();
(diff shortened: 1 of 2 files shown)


obviyus:
Thanks @zenglingbiao! Merged as 3be8b660e2278e5a0055004ac9eb20bc5b445d2e.

**Proof:** Real `openclaw plugins build` runs preserve the original manifest and full permission bits after partial writes or rejected renames, then succeed on retry through physical and symlink project roots.

- **Bug:** An interrupted in-place write could corrupt the user's package manifest.
- **Fix:** Stage and publish through the existing atomic helper, preserve full modes, disable copy fallback, and resolve symlink roots at publication.
- **Validation:** All 28 authoring tests and [CI for the merged head](https://github.com/openclaw/openclaw/actions/runs/34087634188) passed. Fresh independent behavior validation and final source review found no remaining defect.

| Before | After |
| --- | --- |
| A file-size limit truncated a 5,164-byte manifest to 2,048 invalid bytes. | The same failure preserves all 5,164 bytes; removing the limit allows a successful build. |
| Unsafe options could strip directory special bits or permit an in-place copy after rename failure. | File mode `02640` and directory mode `03770` remain intact, and rename failures remain visible. |

Owned temporary files are removed. Substituted, multiply linked, or relocated entries remain protected when cleanup ownership cannot be verified. The original literal cleanup failure, its clarified ownership interpretation, and all raw observations are retained; no unconditional deletion or cross-manifest transaction is claimed.

PR #140743, opened 2026-09-07
fix(agents): plugin revisions fail while session history rebuilds

petroff:
## What Problem This Solves

Fixes: a plugin returning `revise` from `before_agent_finalize` can fail a SQLite-backed turn with “Session transcript projection is rebuilding” instead of producing the revised answer.

## User Impact

The requested revision completes with one accepted answer. The rejected draft and internal revision instruction remain outside active conversation history. Normal completion, cancellation, and the existing revision limit remain intact; no configuration or migration is required.

## Why This Change Was Made

Rejecting the draft rewinds the active transcript through its existing append-only control. The revision now marks that run as an owned-transcript retry so the existing abort-aware projection wait completes before the next attempt reopens the session. There is no new retry loop, storage field, schema, or recovery path.

The related yield-handoff repair in #141718 remains separate; this change fixes the finalization-revision handoff.

## Evidence

A real local Gateway, registered finalization plugin, SQLite sessions, and deterministic HTTP provider exercised the current code:

| Scenario | Observed result |
| --- | --- |
| Main `212313307870` | One draft and a real `revise` decision, then the projection error; no final answer. Ordinary completion passed. |
| Repaired revision | Two model calls, exactly one revised final answer, and active history containing only the original user turn and accepted answer. |
| Independent revision instruction | The next request receives the complete changed instruction and produces its different expected answer. |
| Cancellation during the hidden pass | After real `chat.abort`, the held response is released and both provider connection and handler settlement are joined. Execution remains aborted, with no late final or assistant entry in active history. |

- The wait-owner regression fails before the fix and passes afterward on Node 24.19.0 / Linux aarch64 in a secretless local Docker container. All 67 selected tests across three affected files pass; 134 unrelated shared-harness cases were filtered.
- Candidate runtime build and structural checks passed. Production source is byte-identical to the runtime-proof inputs; the final type-only test cleanup preserves the same emitted JavaScript.
- A separate [community reproduction on released 2026.9.4 with Fireworks](https://github.com/openclaw/openclaw/pull/140743#issuecomment-5742520348) reports the same failure in three fresh sessions. That external-provider observation is not claimed as our candidate run.


VACInc:
Related owner-wait fix merged in #141718 ([merge commit](https://github.com/openclaw/openclaw/commit/d12106ef815f9e36585cb53db639a1e447056cb1)), covering `sessions_yield` cleanup before the next queued turn, plus preflight compaction accounting.

This PR remains distinct and needed for review: #141718 does not mark the `before_agent_finalize` revision as an owned transcript retry. Current main still activates the hidden revision prompt and returns `retry` without the `markOwnedTranscriptRetry` call in this PR. The same projection error can arise at different mutation owners; the yield repair does not supersede your finalization regression or Gateway proof. Leaving this PR open.

petroff:
Revalidated and rebuilt this PR on current `openclaw/main` (`ecc4c992fb1b956a901ffb3fe13676f81a315b21`). The updated PR head is `44509a9b28af5fdc28d21d9606171ff451182fbc`; all three original commits and authors are preserved.

Current validation:

- `pnpm test src/agents/embedded-agent-runner/run/terminal-resolution.test.ts`: 51/51 passed.
- `git diff --check origin/main...HEAD`: passed.
- Independent autoreview: no P0/P1 findings; the revision path marks the owned transcript retry once, while ordinary completion remains unchanged.
- `pnpm check:changed`: guards, formatting, and core production typecheck passed. The core test-type shard reaches the same pre-existing `payloads.test.ts:626` missing-`finalMessageStart` error on a clean checkout of this exact `main`, so it is not introduced by this PR.

This remains a runtime retry/projection-lifecycle fix only: no persisted schema, state shape, migration, or protocol change.

@clawsweeper re-review

petroff:
Rebased onto current main `c9ab1cff77ecd3986b46078243a5c7c8498e9f42` and force-pushed exact head `a3da681c616d2c2a5cb9dfecf76dd7908db04e31`. The issue remains on main; the final diff is now limited to the owned-transcript retry wiring plus consolidated regression coverage. Focused terminal-resolution suite passes 51/51, and local changed-file guards plus production typecheck passed. The previous unrelated plugin-boundary CI failure was fixed on current main. @clawsweeper re-review

petroff:
The requested data-model compatibility proof is not applicable: this patch introduces no persisted or serialized field. `run-loop.ts` only passes the already-existing `sessionPromptState.markOwnedTranscriptRetry` callback into terminal resolution. Its backing value is the runtime-local closure boolean `settleOwnedTranscriptProjection` declared in `session-prompt-state.ts`; it is set, consumed, and cleared within the current run. No schema, encoder/decoder, database column, transcript record shape, config, or migration changes. Existing stored sessions therefore have exactly the same representation before and after upgrade. Please clear the stored-data-model blocker. @clawsweeper re-review

rmyers64:
Independent reproduction on the released **2026.9.4** package with a real external provider, in case it helps with the live-provider gap noted in the PR evidence.

**Environment:** OpenClaw 2026.9.4 (3a9d69d), npm package on Node 24.16.0, Ubuntu Linux (x64), gateway as a systemd service, embedded runtime, SQLite transcripts. Model `fireworks/accounts/fireworks/models/deepseek-v4p1-flash` (route: openclaw -> Fireworks, `openai-completions` API).

**Steps to reproduce**

1. Create a local plugin `my-guard` (a reduced form of the probe plugin we ran, minus logging):

   `package.json`
   ```json
   { "name": "my-guard", "version": "1.0.0", "type": "module", "openclaw": { "extensions": ["./index.js"] } }
   ```
   `openclaw.plugin.json`
   ```json
   { "id": "my-guard", "name": "My Guard", "activation": { "onStartup": true }, "configSchema": { "type": "object", "additionalProperties": false } }
   ```
   `index.js`
   ```js
   import { definePluginEntry } from "openclaw/plugin-sdk/plugin-entry";
   const revised = new Set();
   export default definePluginEntry({
     id: "my-guard",
     name: "My Guard",
     register(api) {
       api.on("before_agent_finalize", (event, ctx) => {
         const runId = event.runId ?? ctx.runId;
         if (revised.has(runId)) return; // revise once per run
         revised.add(runId);
         return { action: "revise", reason: "probe: please restate the answer" };
       });
     },
   });
   ```
2. `openclaw plugins install --link ./my-guard --force && openclaw plugins enable my-guard`, and set `plugins.entries.my-guard.hooks.allowConversationAccess: true`.
3. On a fresh session: `openclaw agent --agent main --session-key test-1 --message "What is 2+2?" --json`

Adding `retry: { instruction, idempotencyKey, maxAttempts: 1 }` to the result made no difference.

**Expected:** one more model pass, then one final reply (docs/plugins/hooks/prompt-and-session.md, `before_agent_finalize`).

**Actual:** 3 of 3 fresh sessions failed. The same prompt succeeds with the plugin disabled.

```text
[agent/embedded] before_agent_finalize requested one more pass: runId=<id> sessionId=<id> attempt=1/3
# ~170 ms later
lane task error ... error="Session transcript projection is rebuilding: <sessionId>"
```
The `agent` RPC returns `UNAVAILABLE` and there is no reply.

**Dist reading (2026.9.4) agrees with this PR:** the revise branch rewinds the rejected answer with a leaf-control append, which sets `needs_rebuild` and queues a background rebuild. It then returns `retry` without `markOwnedTranscriptRetry()`, so the next attempt opens the session through the projection snapshot without waiting and throws at once. The run loop recovers only auth errors there. The other rewind paths (orphan repair, yield rewrite, compaction and truncation retries) wait for the rebuild. The revise branch in `run/terminal-resolution.ts` on `main` (bedc7d8) still returns `retry` without the wait.

**Impact:** any plugin that uses `revise` loses the whole turn on SQLite transcripts, so revise is unusable for us on 2026.9.4. We disabled it and used a fallback.

obviyus:
landpr-approve-unrelated-red
pr: 140743
head: 6c8c4eab0e83a1e27d51d8156fd69d509f392a79
run: 35457160937 attempt: 2
failed: checks-node-changed-extensions-bundle-36 job=105940869529 conclusion=failure
check: openclaw/ci-gate

obviyus:
Landed. Plugin-requested finalization revisions now wait for their own transcript projection before reopening the session, so the revised answer completes without exposing the rejected draft or hidden instruction in active history.

Verified current-main red/green through a real Gateway, registered plugin, SQLite, and HTTP provider, including cancellation with joined provider work and no late final. The 67 selected Linux regression checks passed. The production-dependency audit passed after npm recovered; landing used the [documented unrelated Buzz test exception](https://github.com/openclaw/openclaw/pull/140743#issuecomment-5744166085), not a claim that every test was green.

Thanks @petroff for the original repair and @rmyers64 for the independent Fireworks reproduction.

PR #140756, opened 2026-09-07
refactor(macos): rebuild the Connection window as a standard Settings window

steipete:
## What Problem This Solves

The native **Connection** window (menu bar → Connection…) was the last hand-styled surface in the Mac app: custom card rows with translucent fills and hairline strokes, a large tinted status tile, pill badges, a segmented picker in the title bar, and an 840×640 window that still scrolled. It looked like a web page rather than a Mac settings window, and it did not match the Dashboard, which now owns every other setting.

## Why This Change Was Made

The window is now a standard macOS Settings window: a SwiftUI `Settings` scene with toolbar tabs (Connection, Gateways, and Debug while the developer toggle is on), a fixed 660pt width, a content-sized height per tab, and the window title following the selected tab. Content uses the native grouped `Form` with `LabeledContent` rows, section footers for guidance and validation, bordered fields, and a compact status row. The custom card components are deleted; Tailscale, discovery, config-conflict, Gateways, and Debug views emit plain form rows and sections. `Cmd-,` still opens Dashboard settings; only the offline Connection window stays native, per `apps/macos/AGENTS.md`.

Non-goals: no settings moved between native and Dashboard, no new configuration, no changes to connection behavior. Copy changes are limited to shorter row labels and footer text; the i18n inventory is regenerated. The native string scanner now recognizes SwiftUI `Tab("…", systemImage:)` titles, so the toolbar tab names keep their catalog entries and translations (the first CI run caught that gap).

## User Impact

The Connection window looks and behaves like every other Mac settings window: familiar toolbar tabs, native light/dark rendering, real text fields, and no scrolling in the common remote configurations. Nothing about connecting to a Gateway changes.

## Evidence

Before (current main, same isolated profile):

Before: card-styled Connection window

After, SSH tunnel:

After: standard Settings window, SSH tunnel

After, Gateways tab:

After: Gateways tab

Captures come from an isolated app profile pointed at a demo target, so the status reads "Connecting…"; the discovered-gateway row is masked. Local proof: `swift build` for the app and test target, `scripts/lint-swift.sh macos`, `pnpm native:i18n:verify`, `node scripts/check-changed.mjs` (7 Vitest shards green), and `swift test --filter ConnectionWindowTests`. `GatewaySettingsSmokeTests` and `HealthStoreStateTests` host these views and run in the macOS CI lane. Codex autoreview: scoped-clean at P0/P1, no actionable findings.

Known gaps: the local-mode sections (Local Gateway, Tailscale) were compiled and reviewed but not captured, since the rig cannot run local mode without touching the operator Gateway; a connected-state capture is also pending a dev Gateway.


steipete:
Merged via squash.

- Prepared head SHA: [cb277c3660dd1fe17118b00a35e1b34ca546adfe](https://github.com/openclaw/openclaw/pull/140756/commits/cb277c3660dd1fe17118b00a35e1b34ca546adfe)
- Landed commit: [adf58186567612d3bcca28776ff6d9001c263cad](https://github.com/openclaw/openclaw/commit/adf58186567612d3bcca28776ff6d9001c263cad)

PR #134939, opened 2026-09-01
feat(android): align chat, sidebar, and appearance with the web UI

IWhatsskill:
## What Problem This Solves

Android chat and navigation lagged behind mobile web and iOS: the composer crowded out the conversation, model controls were hard to reach, and sidebar/session and appearance behavior was inconsistent. This keeps the compact composer already landed in #135432 and completes the larger Android parity work.

The maintainer rewrite also repairs stale settings replies, native-thread action eligibility, terminal error visibility, and profile appearance changes lost during offline/read-only profile switching. Final live testing also caught and repaired global New session being blocked after a native chat.

## Why This Change Was Made

Settings updates now use one ordered, session-owned flow, checked again when the Gateway request is actually enqueued. Native model ownership comes from canonical Gateway metadata rather than guesses; effort, Fast Mode, and permissions remain available where supported. A current run's diagnostic owner survives spinner cleanup, so a later terminal error stays visible without showing errors from an old run.

Plain New from a locked native conversation now starts an independent session using the selected agent’s defaults. Normal mutable chats retain their parent hooks, while explicit generic fork/worktree actions keep the native-parent restriction and enqueue-time checks.

Appearance uses the existing preference store and synchronization owner. A device-local choice no longer deletes another profile's queued changes, and an acknowledgement retires only its matching queued value without overwriting newer local choices. The three setters share one edit path; this last repair removes 24 production lines. Sidebar navigation has one owner, with the obsolete adaptive bottom-bar path removed.

Contributor history and credit are retained. Thanks @IWhatsskill and Ayaan Zaidi.

## User Impact

- A full-width editor and compact, single-row controls; readable model names, aligned voice icon, 48dp touch targets, and only one filled primary action.
- Searchable model controls with authentication repair, session settings, and clear native-model restrictions. A blocked draft keeps a disabled Send action instead of silently changing into Talk.
- Native session/catalog navigation, working global New after a native conversation, reliable sidebar ordering, and theme choices with readable light/dark contrast and visible selection states.
- Offline appearance edits reach their original profile once it reconnects; newer local choices survive app recreation and acknowledgement.

## Evidence

Candidate: `beb9db2d5e2617dd08883e8284cab88fcde6f9f2`, integrating main `c4b8d42bd42cdf4ef447bd48dc96abc1fa33c27f`.

[...]

PR #140759, opened 2026-09-07
fix(android): keep your reading position as replies grow

steipete:
Related: #134939. Persistent reading bookmarks in #137674 remain separate work.

## What Problem This Solves

Fixes an issue where Android users reading an earlier part of a reply would lose their place as more text arrived or the window changed size. Expanding long messages and using Start/End could also leave the requested part clipped behind the conversation viewport.

## Why This Change Was Made

A session-owned navigation job now coordinates following, gestures, disclosure, and code navigation. The reader retains an actual rendered glyph or content point, rather than guessing from whole-row height changes. Its preferred position survives temporary scroll limits and width/height changes; explicit navigation, replacement, and disposal release the old anchor and its pinned rows.

This replaces the earlier whole-row correction flagged in review. Appended text, content appearing above the reading point, neighboring-row growth, and reflow now use the same owner. Current main's compact Details, tabletop, IME, and model-picker layout work is preserved. No configuration, dependency, database, protocol, authentication, or persisted-bookmark behavior changes.

## User Impact

Readers keep the paragraph they chose as replies grow and can deliberately return to the latest text. View all and code Start/End reveal the requested content within the actual conversation viewport. Resizing an idle conversation so everything fits resumes following.

## Evidence

The integrated candidate passed **485 tests across 35 suites**, both Play and ThirdParty Kotlin compilation, Android ktlint, and both lint variants. The repository changed-file checks also passed locally, including app/Wear formatting and the native-state schema guard. One existing saveable-state autoboxing hint is retained to avoid an unproven restoration-compatibility change; there are no lint errors or warnings.

Rendered regressions cover deferred-prefix and streaming-tail growth, unrelated rows, navigation cancellation, touch/wheel/accessibility scroll, replacement/disposal, compact/tabletop layouts, and resize restoration. The new width-only case first failed at **582 → 1114 → 1114 px**, then passed at **582 → 1114 → 582 px** with unchanged glyph height and genuine scroll room after restoration. The height-round-trip case similarly returns to its original position. An existing question-interaction fixture now waits for its actual database initialization before beginning UI assertions; no deadline or behavior assertion was weakened.

Installed comparison: the baseline is main `823ff3ed9ad61e03aecd0bec0421a5463d80eccf`, built and installed on the task-owned API 36 emulator. A synthetic local TLS Gateway supplies the same long-code and streamed-reply scenario without a real provider or user data. The baseline hides the opening after View all, clips R699 after End, and moves from S001 to S021 on the second streaming update without intervening input. The after APK is exact `1281264fda33b0d0b98192d45fd7bd568df2b04e`, with verified embedded commit and the same signing certificate. It reveals the opening, fully shows R699 after End and R000 after Start, and restores the collapsed preview with Close. After one real Send and one swipe to pause, S001 remains at the same paragraph origin **(42, 740)** through the second update to S048, with no intervening UI input. A separate Jump capture shows S048 before the final update is released; the final reply then appears. These are real installed-app/synthetic-Gateway checks, not live-provider proof. The fixture and emulator stopped normally and display settings were restored.

A configured remote changed-check attempt stopped with SSH transport exit 255 before checks started; its lease and temporary checkout were cleaned. The successful changed checks above are local, not a claim that this remote attempt passed. The complete three-pass independent review is clean, with no P0–P2 findings. Exact-head [hosted CI](https://github.com/openclaw/openclaw/actions/runs/34183296478) is green. The current [hosted review](https://github.com/openclaw/openclaw/pull/140759#issuecomment-5564878347) also has no actionable findings; it confirms the prior blocker is resolved. Native landing follows these checks.

Production is +681/−168 (net +513), tests +1367/−37 (net +1330), and docs net +1. Most growth establishes the single rendered-coordinate/navigation lifetime and meaningful regression coverage; two nearby non-saveable scalar states also use the existing primitive Compose factories. The old callback/counter and whole-row approximation paths are removed.

### Installed before/after

Both sides use the same isolated API 36 emulator, synthetic content, and original display density/font scale. The code-control case uses the same compact viewport.

| Scenario | Baseline | Repaired app |
| --- | --- | --- |
| View all | Opening hidden after View all | Opening and R000 revealed |
| End of code | R699 remains clipped | R699 fully visible |
| More text while reading | Chosen paragraph lost after update | S001 stays in place after update |

Installed checks cover these settled controls and streaming geometry. Rendered tests separately cover deferred content, resizing, and cancellation interleavings. Persistent bookmark restoration and wheel interruption of an already-active manual reveal remain separate follow-ups, not claimed by this proof.


steipete:
Merged via squash.

- Prepared head SHA: [1281264fda33b0d0b98192d45fd7bd568df2b04e](https://github.com/openclaw/openclaw/pull/140759/commits/1281264fda33b0d0b98192d45fd7bd568df2b04e)
- Landed commit: [9c852138412369c25ac25346b401ec1eca81a575](https://github.com/openclaw/openclaw/commit/9c852138412369c25ac25346b401ec1eca81a575)

PR #140760, opened 2026-09-07
fix(daemon): preserve service environment after failed writes

zenglingbiao:
## What Problem This Solves

Fixes an issue where a failed service environment rewrite could truncate existing credentials or operator settings. LaunchAgent installation and repair rewrote the live environment file directly. Node service uninstall did the same when removing managed credentials from the systemd environment file.

## Why This Change Was Made

Both writers now stage complete content in an exclusively created private file and publish it by rename. Failed creation, writing, or rename leaves the previous environment intact and cleans up the temporary file. The systemd writer uses the existing strict atomic replacement primitive, preserves the full parent-directory mode, and supports state-directory aliases.

Serialization and readers are unchanged. Existing environment files, quoted shell references, wrapper arguments, and unrelated settings retain their existing format. No schema, configuration, or migration change is needed.

## User Impact

A failed rewrite preserves the previous complete service environment. Node uninstall removes only its managed credentials on success and retains operator settings. LaunchAgent services can still authenticate with their preserved credentials after a failed rotation.

## Evidence

- Real Linux `openclaw node uninstall --json` against the task account's native systemd manager: current main truncated 2,357 bytes to 512 under `EFBIG`; the candidate retained all 2,357 bytes. A fresh command without the fault completed uninstall and retained all operator values.
- Both launchd staging regression tests fail against current main at the original-byte assertions. The revised launchd suite passed 160 tests with 25 existing skips.

- Current main `d5bbb4368700cce040dbb85622748e6e006ad611`: a real kernel file-size limit produced `EFBIG` and truncated both environment files to 512 bytes.
- Repaired candidate: the same failures preserved every original byte, mode `0600`, parent mode `03700`, and left no temporary file.
- Creation and `ENOSPC`/`EPERM` rename failures preserve existing files. Success through both ordinary directories and directory aliases preserves permissions and serialized values.
- A compiled Gateway started from the preserved LaunchAgent environment accepted the synthetic credential at a protected HTTP endpoint; wrong and missing credentials returned `401`.
- Focused launchd and systemd suites: 391 passed, 25 existing skips. Five added systemd cases fail against the original candidate for the intended mode, alias, and fallback defects.
- Changed-file formatting, lint, assertion-safety, and line-count checks passed.

File-publication proof ran on isolated Linux with synthetic credentials. Native macOS service activation and production services were not exercised.

The existing LaunchAgent private-directory hardening remains unchanged: it removes group/other access before publication. Independent validation confirmed this on both main and the candidate; unconditional preservation of a previously permissive LaunchAgent directory mode is not claimed.


zenglingbiao:
@clawsweeper re-review

PR #138870, opened 2026-09-05
fix(plugins): retain install files when ownership expires

steipete:
Closes #138815. Builds on the private npm staging already landed in #140145.

## What Problem This Solves

Fixes an issue where plugin installation could continue deleting copied source files after its installation owner expired. Retained transactions could also trust a substituted backup or unknown Windows filesystem identity, and a failure after directory publication could leave a new install in place without reporting that published state.

## Why This Change Was Made

Adopt published `@openclaw/fs-safe` 0.8.3 and use its synchronous per-mutation authority hook and exact bigint destination receipts. The directory publisher owns backup, forward publication, immediate recovery, and retained rollback through one recovery path. It records successful filesystem publication before later verification or source cleanup can fail, and rechecks current lifecycle authority and identity before recovery.

The npm preparation, generation selection, peers, archives and consent flow remain with the existing owners. The bundled 1Password plugin uses the same released dependency. No configuration option, SQLite schema, or SDK budget change is introduced.

## User Impact

Windows EPERM and cross-device copy fallback remain available. Caller cancellation can still restore the original install while the lifecycle lease remains owned. Once that lease closes, recovery preserves the trees and reports the published destination and backup path. Substituted backups and successor installations are rejected without mutation. A retry after successful restoration only finishes remaining backup cleanup after verifying the restored identity.

## Evidence

- Seven ownership/identity cases failed on unchanged current main before this repair; the Windows EPERM fallback control passed. Additional causal tests caught six immediate-publication recovery cases and an already-restored cleanup retry during development, then passed after sharing the rollback owner.
- Final registry dependency: 266 local tests passed across directory publication/rollback, npm/npm-pack contracts, and the bundled 1Password plugin; one platform-specific case was skipped. Plugin credential/process fixtures are synthetic.
- Final independent Codex P0–P2 review is scoped-clean. The initial publication-failure finding is fixed and covered by the causal regressions.
- Published dependency proof: [fs-safe 0.8.3](https://github.com/openclaw/fs-safe/releases/tag/v0.8.3), with all eight tarball identities/signatures/provenance verified by its release workflow. [Cross-platform CI](https://github.com/openclaw/fs-safe/actions/runs/34071393565) includes an actual Windows open-file EPERM fallback, and a packaged consumer run proves receipt retention and cleanup refusal using real files without filesystem mocks.
- The complete changed-code gate passed, including all type lanes, package-lock validation, dependency pins, SDK budgets, boundary checks, dead-export scans, lint, and runtime import cycles.
[...]

PR #140764, opened 2026-09-07
fix(openshell): report preserved shadows after mirror failures

### LINKED ISSUE #140703: OpenShell mirror failures omit published shadow recovery paths
OpenShell preserves local workspace shadows while importing a downloaded mirror. A cross-filesystem move can publish a complete backup and then fail while cleaning up the source. The current caller records the backup only after the entire move resolves, so this failure leaves a recovery copy on disk without reporting its location.

Two real-filesystem reproductions demonstrate the problem: one source entry is removed before the next cleanup operation refuses, leaving a partial source and complete backup; and a source replacement after publication causes ESTALE, preserving the replacement plus the original backup. Both errors omit the backup location. Normal cross-filesystem and file-parent restoration controls pass.

Register the exact publication receipt immediately. Keep incomplete source cleanup separate from a completed move: retain and report both locations when cleanup is incomplete, and restore only completed moves with verified backup identity. Continue recovering earlier records when a later restoration fails. If restoration succeeds but removing the empty preservation directory fails, report the restored workspace and leftover directory accurately.

The plugin can own the compatible @openclaw/fs-safe 0.8.3 dependency so older supported hosts do not silently ignore the publication callback. Keep the existing host compatibility floor and remaining SDK seams unchanged. This relates to the installer adoption in #138870 but concerns OpenShell's separate mirror lifecycle.

steipete:
Closes #140703

## What Problem This Solves

Fixes an issue where OpenShell mirror synchronization could leave a complete local-shadow backup on disk but omit its recovery location when source cleanup failed after publication. A partial source or concurrent replacement could remain beside that backup, making recovery unclear.

## Why This Change Was Made

Register each shadow's exact identity as soon as fs-safe publishes its backup. Treat incomplete source cleanup separately from a completed move: retain and report both paths for incomplete moves, and restore completed moves in reverse order with verified backup identity. Continue restoring earlier shadows after an independent failure, preserving all failure causes.

When restoration succeeds and only removal of the empty preservation directory fails, report the restored workspace and leftover directory accurately. The plugin owns exact `@openclaw/fs-safe@0.8.3`, so supported older hosts cannot silently ignore its publication callback. Existing SDK seams and the minimum host version remain unchanged. The root dependency adoption already landed in #138870.

## User Impact

Mirror failures identify the recovery copies and remaining workspace entries. Concurrent replacements and unverified backups are preserved. Operators can distinguish incomplete restoration from successful restoration followed by directory-cleanup failure. This adds necessary recovery logic rather than claiming a net production-code reduction.

## Evidence

The original source failed two real-filesystem cases while two controls passed: partial source cleanup after a completed publication, and a source replacement rejected with ESTALE. Both retained a complete backup but omitted its path. A separate regression reproduced the misleading missing-backup diagnostic after successful restoration and failed empty-directory cleanup.

The final OpenShell sibling group passed 141 cases; seven focused failure/control cases also passed. The generic extension runtime-dependency contract passed 451 cases. The final plugin runtime build, manifest-overlay npm pack, transient npm lock validation and normal managed plugin installation passed in isolated state. Importing the actual installed runtime verified the final tarball bytes, its unchanged host minimum and its own fs-safe 0.8.3 dependency; a real staged move returned the expected publication receipt.

The tests use synthetic workspaces, actual local filesystem operations and a mocked OpenShell CLI. No live OpenShell environment or operator Gateway was used; no native Windows OpenShell run is claimed. The final source/test/docs gate passed without skip flags; unchanged manifest and dependency checks passed in the preceding validation runs. Formatting and targeted type-aware lint passed. Review triage left no accepted/actionable findings: one automated suggestion was rejected against the caller flow and its existing passing reverse-cleanup regression. After alignment onto current main, all five reviewed owner files remain byte-identical; the lockfile adds only this plugin's existing 0.8.3 dependency entry.

Published dependency contract: fs-safe v0.8.3 defines the [frozen bigint receipt and synchronous callbacks](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/move-path.ts#L21), captures the callbacks and [reports publication before source cleanup](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/move-path.ts#L299), and [guards copied-source cleanup](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/move-path-cleanup.ts#L192). The move retains `Promise<void>` and EXDEV/Windows EPERM fallback. The [verified 0.8.3 release](https://github.com/openclaw/fs-safe/releases/tag/v0.8.3) includes the cross-platform proof.

The maintainer-approved import exception is now implemented: only the named atomic move and type-only publication receipt are allowed in OpenShell's backend. Existing prefix exemptions and detection remain. The boundary suite passes 21 cases covering allowed aliases, prohibited broader imports, other owners and additional direct imports; fresh P2 review found no actionable issues. All selected changed checks passed except a parameter-reassignment lint issue, corrected with an equivalent local string variable; final type-aware lint and the 21-case rerun pass. No production code changed in this follow-up.

### DIFF
--- a/docs/gateway/openshell.md
+++ b/docs/gateway/openshell.md
@@ -536,6 +536,15 @@ openclaw logs --follow
 - **An image or attachment cannot be sent:** Use a path under the configured
   `remoteWorkspaceDir`, such as `/sandbox/report.png`, rather than assuming
   every backend uses Docker's `/workspace` directory.
+- **Mirror synchronization reports recovery paths:** OpenClaw kept a host shadow
+  outside the workspace because its move or restoration could not finish. The
+  error identifies the preserved path and the workspace path and retains the
+  original failure. Compare both paths and recover the needed files before
+  deleting either copy. A partial move can leave different files in each path;
+  OpenClaw preserves remaining workspace entries instead of overwriting them
+  with an incomplete or unverified backup. If restoration completed and only
+  cleanup of the preservation directory failed, the error confirms the restored
+  workspace path and identifies the leftover directory instead.
 - **Recreate or prune cannot delete a sandbox:** Restore access to the original
   OpenShell gateway and workspace, confirm the sandbox still exists with
   `openshell --workspace <workspace-name> sandbox get <sandbox-name>`, and retry
--- a/extensions/openshell/package.json
+++ b/extensions/openshell/package.json
@@ -8,6 +8,7 @@
   },
   "type": "module",
   "dependencies": {
+    "@openclaw/fs-safe": "0.8.3",
     "p-limit": "7.3.1",
     "zod": "4.4.3"
   },
--- a/extensions/openshell/src/backend.ts
+++ b/extensions/openshell/src/backend.ts
@@ -1,7 +1,12 @@
 // Openshell plugin module implements backend behavior.
 import { createHash } from "node:crypto";
+import fsSync from "node:fs";
 import fs from "node:fs/promises";
 import path from "node:path";
+import {
+  movePathWithCopyFallback,
+  type MovePathPublicationReceipt,
+} from "@openclaw/fs-safe/atomic";
 import { createDeferred } from "openclaw/plugin-sdk/extension-shared";
 import { KeyedAsyncQueue } from "openclaw/plugin-sdk/keyed-async-queue";
 import type {
@@ -38,7 +43,6 @@ import { resolveOpenShellPluginConfig, type ResolvedOpenShellPluginConfig } from
 import { createOpenShellFsBridge } from "./fs-bridge.js";
 import {
   DEFAULT_OPEN_SHELL_MIRROR_EXCLUDE_DIRS,
-  movePathWithCopyFallback,
   replaceDirectoryContents,
   stageDirectoryContents,
 } from "./mirror.js";
@@ -1077,6 +1081,7 @@ class OpenShellSandboxBackendImpl {
             throw new Error(result.stderr.trim() || "openshell sandbox download failed");
           }
           const preservedShadows: PreservedLocalShadow[] = [];
+          const failures: unknown[] = [];
           try {
             for (const shadowedRoot of roots.slice(index + 1)) {
               if (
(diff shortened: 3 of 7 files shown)


steipete:
Maintainer dependency inspection is complete. I inspected the released fs-safe v0.8.3 source and its installed-package behavior:

- The [public options and receipt](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/move-path.ts#L21) preserve `Promise<void>` and expose synchronous mutation/publication hooks with a frozen bigint filesystem identity.
- The [move owner](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/move-path.ts#L299) captures those callbacks before awaited work. Both direct and staged renames report publication before post-rename checks or copied-source cleanup.
- The [cleanup owner](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/move-path-cleanup.ts#L192) invokes the mutation assertion before each source unlink/rmdir. EXDEV and Windows EPERM fallback remain intact; the release's actual Windows open-file fallback case passed.

The final installed OpenShell tarball resolved its own fs-safe 0.8.3 and returned a real publication receipt. The caller regressions cover both outward and reverse partial cleanup. This resolves the dependency-inspection concern without treating the reviewer's network failure as a code finding.

CI has one attributable failure: the existing TypeScript fs-safe import-boundary guard rejects the plugin-owned import. A narrowly scoped exception is prepared and under maintainer approval; the repository guard remains unchanged, and this PR will stay unmerged until that policy decision and green exact-head CI are complete.

steipete:
Merged via squash.

- Prepared head SHA: [4a4f842da20570e5307469246a855f1d0672fa27](https://github.com/openclaw/openclaw/pull/140764/commits/4a4f842da20570e5307469246a855f1d0672fa27)
- Landed commit: [170ff662b64afeb99f2baae1be93445050860c6b](https://github.com/openclaw/openclaw/commit/170ff662b64afeb99f2baae1be93445050860c6b)

PR #138656, opened 2026-09-04
fix(terminal): stream wrapped table rows without argument limits

steipete:
## What Problem This Solves

Fixes an issue where CLI tables can crash with `RangeError: Maximum call stack size exceeded` when one cell wraps into many physical rows or retains a long zero-width suffix at a word break.

## Why This Change Was Made

Consume the existing ANSI iterator directly into the active wrap buffer, retain the unconsumed suffix in that buffer, and append physical rows directly to the table output. This removes the whole-cell token collection, suffix copy, temporary per-row output, and both unbounded argument spreads.

## User Impact

Very long wrapped cells render without depending on the JavaScript argument-count limit. Grapheme widths, ANSI styles and links, line breaks, padding, and borders retain their behavior. Production LOC: +49/-54 (net -5); tests: +36/-26 (net +10).

## Evidence

Both failures were reproduced before the fix in normal Node processes: a 450,000-character cell at width 3 and a 150,000-character zero-width suffix. The final owner passes both; the long cell emits all 150,000 data rows. The regression tests also fail on the original code.

All 174 focused table, ANSI, and hooks tests passed. Exact output parity passed for 375 cases across Unicode/control fixtures, three borders, and five widths. The registered source CLI ran `openclaw hooks list --agent proof` against synthetic filesystem-discovered hooks in an isolated workspace; both arms exited successfully with all 698,248 stdout bytes identical.

The complete 28-stage changed-file gate passed on the final source. Independent managed Codex review through P2 found no actionable issues. The removed storage and copies are source-level improvements; shared-load timing and RSS observations are not claimed as controlled performance results.

The integrated built `openclaw.mjs hooks list --agent proof` comparison now passes. The candidate build contains this PR’s exact table postimage; it is an integrated build, not an isolated build of the PR branch. A synthetic hook with a long styled description produced byte-identical 377,781-byte outputs, SHA-256 `4d80f5705d3e0dc020afa1133391ddd6c78754de9d3295156cc35b42db5860a5`. Both outputs retained all 2,000 description repetitions, CJK, emoji and combining text, ANSI styling, and three 120-character borders across 2,505 lines. Complete source, build, dependency, and Node identities matched before and after each arm. Captures were untruncated, every owned command exited successfully, and observed processes and temporary runtimes closed. This verifies built-entry functional parity; it is not a timing or memory measurement.

PR #135918, opened 2026-09-02
improve: reduce memory and latency when truncating terminal text

steipete:
## What Problem This Solves

Truncating a long terminal line indexed every grapheme and allocated an offset array even when only a few display columns were needed. A 1 MiB ASCII line required roughly 35 MiB of immediate heap growth just to keep 40 columns.

## Why This Change Was Made

The terminal width owner now finds complete grapheme boundaries with `Intl.Segments.containing()`, brackets the estimated cut and measures the shorter side. It reuses a cluster's measured width while searching within that cluster and keeps the search bounded. ANSI scanning, reset handling and display-width measurement retain their existing owners.

The unreachable optional-Segmenter fallback is removed: the locked `string-width` dependency already constructs `Intl.Segmenter` during module initialization. Production changes are +61/−72 lines (net −11); tests add 14 lines. No dependency or configuration changes.

## User Impact

Terminal formatting keeps identical text, styling and Unicode boundaries while using less temporary memory for long lines. Across three alternating Node 26.8.1 process pairs, 1 MiB ASCII truncation fell from a median 59.8 ms to 2.2 ms and immediate heap growth from 36,494,848 bytes to 2,256 bytes. Near-end cuts showed similar improvements; the mixed Unicode fixture fell from 23.1 ms to 14.1 ms.

The oversized single-cluster control was effectively unchanged (1.52 ms versus 1.54 ms). These are synthetic owner measurements; immediate heap/RSS observations include allocator effects and do not establish a whole-Gateway or retained-memory reduction.

## Evidence

- 55,824 Unicode/ANSI input-budget cases produced identical before/after digests and respected display-width limits. A real rendered table also matched byte for byte.
- Added contracts cover uneven-width runs, invisible characters, combining marks, flag boundaries, an oversized suffix cluster and invalid-width termination.
- Canonical ANSI, table and command text-format suites: 154 tests passed. Full changed-file gate passed on the final source.
- Independent source/dependency review checked callers and the V8 grapheme-boundary contract. Reported issues were repaired and rechecked; final Codex autoreview has no accepted findings.

Baseline: `173f41d682b0d4a05674820a3d390d934da8b066`. The boundary operation follows [ECMA-402's Segments containing contract](https://tc39.es/ecma402/#sec-%segmentsprototype%.containing).

PR #140773, opened 2026-09-07
perf(terminal): avoid eager grapheme arrays during wrapping

### LINKED ISSUE #140772: Terminal wrapping retains eager grapheme arrays during sequential iteration
Terminal table and note wrapping consume graphemes sequentially, but call splitGraphemes first and retain its full Array during iteration. The remaining intermediate is unnecessary even though table token processing already streams.

An actual renderTable call paused at the first accepted character of a 450,000-character cell retains a 450,000-element Array through its Array Iterator, with a 5,171,424-byte backing store. A sequential generator over the same Intl.Segmenter removes that intermediate at the identical execution boundary while preserving the complete rendered output.

Keep the existing array API and its callers intact; migrate only the two sequential wrappers. ANSI run boundaries, Unicode clusters, newlines, visible width and copy-sensitive note text must remain identical. This is transient allocation work, not a post-return leak or constant-memory output promise.

Related #138656 streams table tokens/rows and #135918 optimizes truncation lookup; neither removes these remaining grapheme arrays. A possible small-render CPU tradeoff must be measured and stated rather than assuming a speed improvement.

steipete:
Closes #140772

<details>

</details>

## What Problem This Solves

Table and note wrapping build and retain complete grapheme arrays even though both consumers only iterate sequentially. Large text runs keep that intermediate alive throughout wrapping.

## Why This Change Was Made

Add a string generator over the existing shared Intl.Segmenter and use it in the two sequential wrappers. The existing splitGraphemes array API remains unchanged, as do ANSI run boundaries, width calculation and final output construction.

## User Impact

In actual renderTable execution paused at the first accepted character of the same 450,000-character cell, a 450,000-element Array with a 5,171,424-byte backing store is absent from the candidate heap. Both resumed renders produce the same 1,200,032 output bytes. Table and note appearance and copy behavior stay identical.

This removes a concrete temporary array. It does not make complete table output streamed or constant-memory, establish a post-return leak fix or imply a general CPU improvement. The 4 KiB plugin-description fixture was approximately 0.12–0.37 ms slower per render in the measured children; large-cell baseline variance prevents a stable speedup claim. Production incidence of huge descriptions is unknown.

## Evidence

- Nine complete output fixtures match across separate baseline/candidate/candidate/baseline children: large and small plugin-shaped cells, colored rows, Unicode/OSC-8/CRLF, fitting ASCII, borderless tables, generic/CJK notes and copy-sensitive URLs.
- External Inspector receipts bind the same first-token execution state in both actual table renders. The retired Array and its backing store are identified through retainers, rather than heap-snapshot file size or post-return totals.
- 172 existing ANSI/table/note tests, the complete selected changed-file gate and fresh isolated P2 review passed on the exact final source. No full build was required by that gate and none is claimed.
- Three production files change by +11/−4 lines. No tests, configuration or existing array API behavior changes.

Related #138656 and #135918 cover distinct earlier terminal work. Timing limits remain explicit; no extra repetitions were selected to obtain a nicer result.

### DIFF
--- a/packages/terminal-core/src/ansi.ts
+++ b/packages/terminal-core/src/ansi.ts
@@ -147,6 +147,13 @@ export function stripAnsiForStreamChunk(
   });
 }
 
+/** Let sequential renderers consume graphemes without retaining a full-run array. */
+export function* iterateGraphemes(input: string): Generator<string, void> {
+  for (const { segment } of graphemeSegmenter.segment(input)) {
+    yield segment;
+  }
+}
+
 export function splitGraphemes(input: string): string[] {
   if (!input) {
     return [];
--- a/packages/terminal-core/src/note.ts
+++ b/packages/terminal-core/src/note.ts
@@ -2,7 +2,7 @@
 import { AsyncLocalStorage } from "node:async_hooks";
 import { note as clackNote } from "@clack/prompts";
 import { normalizeLowercaseStringOrEmpty } from "@openclaw/normalization-core/string-coerce";
-import { splitGraphemes, visibleWidth } from "./ansi.js";
+import { iterateGraphemes, visibleWidth } from "./ansi.js";
 import { stylePromptTitle } from "./prompt-style.js";
 
 const MIN_NOTE_COLUMNS = 80;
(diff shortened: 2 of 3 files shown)


RaminduDul (commented):
This is good example of a performance change that fixes a specific allocation problem without unnecessarily redesigning the terminal rendering architecture.

steipete:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [70fbd4fe80873880d4759284691cc2beaf62a84e](https://github.com/openclaw/openclaw/pull/140773/commits/70fbd4fe80873880d4759284691cc2beaf62a84e)
- Landed commit: [3de516d21fc4bb05e4c9166c3be9a985b53f5a92](https://github.com/openclaw/openclaw/commit/3de516d21fc4bb05e4c9166c3be9a985b53f5a92)

PR #140774, opened 2026-09-07
fix(whatsapp): retry pre-delivery outbound failures by typed signal

zenglingbiao:
## Summary

WhatsApp outbound retry decided "should we resend?" with a single text regex over the formatted error (`closed|reset|timed\s*out|disconnect`). Typed pre-delivery failures whose text contains none of those words — Node socket/DNS errors (`write EPIPE`, `getaddrinfo EAI_AGAIN`) and Baileys boom errors like a 428 whose message is "Connection Terminated" — were abandoned after one attempt, silently dropping the user's reply. The predicate now checks typed signals first (Node `error.code` and the Baileys `output.statusCode` read from the direct error only), keeping the text regex as the fallback.

## What Problem This Solves

When a WhatsApp reply send fails before the message ever leaves the host — the socket is already dead (`EPIPE`), DNS can't resolve (`EAI_AGAIN`), or Baileys reports the connection closed with wording that happens to miss the regex — the current code retries nothing:

- Trigger: transient transport failure on an outbound WhatsApp send whose error text lacks `closed|reset|timed out|disconnect`.
- Symptom: the reply is dropped after a single attempt; the user only sees a log line, the recipient never gets the message.
- Blast radius: every outbound WhatsApp send during reconnect churn, DNS hiccups, or socket handover.

**Code evidence**: `extensions/whatsapp/src/outbound-retry.ts` — `isRetryableWhatsAppOutboundError` ended at the text regex, while the same package already extracts the authoritative typed signal (`getStatusCode`, used by `session.ts`/`connection-controller.ts` for reconnect decisions).

## Root Cause

- The text-only predicate dates to commit `0934c445d3c` (#105579, 2026-07-13), which moved outbound retry onto the shared runner without typing the classification.
- Violated invariant: retry classification should key on *what* failed, not on whether a human-readable message happens to contain a keyword. Error message text is not a stable contract across Baileys/Node releases.
- The adjacent inbound sibling (`inbound/socket-session.ts`) already recognizes a wider reconnect vocabulary (`no active socket`), confirming the outbound pattern was the stale one.

## Why This Fix

- Typed-first classification, text as fallback: Node `error.code ∈ {EPIPE, ECONNREFUSED, EAI_AGAIN, ENOTFOUND, ENETUNREACH}` and Baileys `output.statusCode ∈ {428 connectionClosed, 440 connectionReplaced, 515 restartRequired}` are all *pre-delivery* failures — the send provably never reached WhatsApp, so replaying cannot duplicate a delivered message (the non-idempotency concern that the existing `isChannelPartialDeliveryError`/`WhatsAppSocketOperationTimeoutError` exclusions already guard).
- Deliberately **excluded**: 408 (`timedOut`/`connectionLost`) and any timeout spelling — delivery is unknown after a timeout, so retrying risks duplicates. This keeps the change complementary to #116232 (which narrows timeout retrying); the two do not overlap.
- Checks stay on the direct error (never the `cause` chain and never a nested wrapper's `error.error.output.statusCode`), preserving the existing "don't inspect wrappers" anti-duplicate-delivery rule. A disconnect status carried only by an inner error does not admit a retry.
- Rejected alternative: widening the regex (e.g. adding `terminated`) keeps chasing message text across dependencies; the typed checks cover the whole class.

## User Impact

- Affected operation: outbound WhatsApp replies during transient transport failures.
- After the fix, a dead-socket/DNS/restart-required send is retried with the existing 3-attempt backoff instead of being dropped silently; genuinely unknown-delivery outcomes (timeouts, partial delivery, disconnect signals carried only by nested wrapper errors) still never retry.

## Evidence

- **Before**: on `main`, a send failing with `write EPIPE`, Baileys 428 "Connection Terminated", or `EAI_AGAIN` was attempted exactly once and dropped.
- **After**: the same pre-delivery failures are retried and the send recovers; a Baileys 428 carried only by a nested `error.error.output.statusCode` is still attempted exactly once, preserving the adapter's direct-error safety contract.

## Real behavior proof

- **Behavior or issue addressed:** Outbound WhatsApp replies were dropped after a single attempt when the send failed pre-delivery with a typed signal whose text misses the retry regex (`write EPIPE`, `getaddrinfo EAI_AGAIN`, Baileys 428 "Connection Terminated"). The retry owner now classifies those typed signals on the direct error only, and a 428 seen solely through a nested wrapper error (`error.error.output.statusCode`) still stops after one attempt.
- **Real environment tested:** Linux, Node v24.18.1 — the real `sendWhatsAppOutboundWithRetry` imported from `extensions/whatsapp/src/outbound-retry.ts` at this branch head, driven end to end by `node --import tsx` with transport faults injected in the sender at the exact send boundary the production auto-reply delivery path passes its sends through.
- **Exact steps or command run after this patch:** `node --import tsx proof.mjs` against this head (fault injection: EPIPE once then success; nested-wrapper 428; direct Baileys 401; direct Baileys 428 then success). The same script was also run against the merge-base (`a58d1245e9d`, checked out in a detached worktree) for the before behavior.
- **Evidence after fix:**

```text
Proof: WhatsApp outbound retry owner — fault injection through the real sendWhatsAppOutboundWithRetry
node v24.18.1, 2026-09-07T07:58:43.579Z

== Scenario 1: EPIPE transport write failure, then recovery ==
  final result: "delivered:msg-1"
  send attempts: 2, onRetry calls: 1
  onRetry info: {"attempt":1,"maxAttempts":3,"backoffMs":500,"errorText":"write EPIPE"}
  PASS recovered with the delivered result
  PASS exactly 2 send attempts (no duplicate dispatch)
  PASS exactly 1 onRetry call
  PASS onRetry error identity is the injected EPIPE

== Scenario 2: baileys 428 only on a NESTED error (error.error.output.statusCode)
  threw: "request failed"
  send attempts: 1, onRetry calls: 0
  PASS original nested error rethrown unchanged
  PASS send attempted exactly once (no replay from an inner error)
  PASS onRetry never called

== Scenario 3: direct baileys 401 loggedOut
  send attempts: 1, onRetry calls: 0
  PASS original 401 error rethrown unchanged
  PASS send attempted exactly once
  PASS onRetry never called

== Scenario 4: direct baileys 428 connectionClosed, then recovery
  final result: "delivered:msg-4"
  send attempts: 2, onRetry calls: 1
  PASS recovered with the delivered result
  PASS exactly 2 send attempts (no duplicate dispatch)
  PASS exactly 1 onRetry call

ALL CHECKS PASSED
```

- **Observed result after fix:** The real `sendWhatsAppOutboundWithRetry` drives the injected sender; an EPIPE on the first attempt followed by success recovers with exactly 2 send attempts and a single `onRetry` call (`attempt=1, maxAttempts=3, backoffMs=500`) — the reply is delivered and never dispatched twice. A 428 carried only by a nested wrapper error and a direct Baileys 401 are each attempted exactly once and the original error object is rethrown unchanged, while a direct Baileys 428 recovers through the same bounded backoff.
- **What was not tested:** A live WhatsApp/Baileys socket session (requires a real WhatsApp account and phone number, unavailable in this environment). The proof exercises the real retry owner at the same send boundary production delivery uses; only the transport fault itself is injected.
- **Repro confirmation:** On the merge-base the identical fault-injection script shows the gap — `write EPIPE` is thrown back after exactly 1 attempt with zero `onRetry` calls:

```text
BEFORE: merge-base a58d1245e9d (text-only classification) — EPIPE does not recover
node v24.18.1, 2026-09-07T07:59:21.389Z

  threw: "write EPIPE"
  send attempts: 1, onRetry calls: 0
  PASS EPIPE is thrown back to the caller (no recovery)
  PASS send attempted exactly once (no retry)
  PASS onRetry never called

ALL CHECKS PASSED
```

## Tests and Validation

- Regression tests added in `extensions/whatsapp/src/outbound-retry.test.ts`:
  - `retries a pre-delivery failure: ...` — EPIPE, EAI_AGAIN, Baileys 428 ("Connection Terminated", no regex keyword), Baileys 515: each is retried and the send eventually resolves.
  - `does not retry ...` — Baileys 401 (logged out), 408 (timeout class), an `ECONNABORTED` send that may have delivered, and a disconnect status (428) carried only by a nested wrapper error: each is attempted exactly once with no `onRetry`.
- The focused retry-owner suite `extensions/whatsapp/src/outbound-retry.test.ts` runs green locally: 15/15 cases.
- `oxlint` and `oxfmt --check` clean on touched files; `pnpm check:assertion-safety --base upstream/main` passes; `pnpm tsgo:extensions` clean; `git diff --check` clean.

## Review Findings Addressed

- **[P2] Keep status classification on the direct error** (`outbound-retry.ts:41-46`): retry classification no longer reuses `getStatusCode`, which also unwraps `error.error.output.statusCode`. A local `getDirectStatusCode` reads only the direct error's own `output.statusCode` / `status`, so `Object.assign(new Error("request failed"), { error: { output: { statusCode: 428 } } })` is attempted exactly once and never replayed — pinned by the new `does not retry a disconnect status carried only by a nested error` regression case and by scenario 2 of the fault-injection proof above.
- **CI `checks-fast-baseline-ratchets`**: the new non-const type assertions in `outbound-retry.ts` now carry `// SAFETY:` invariants (arbitrary thrown send failure; fields are read then narrowed by `typeof` before use), clearing the `1 > 0` baseline overrun.
- Branch was 207 commits behind `main`; merged current `main` (`6ed74d66d5c`). New head: `087f199d990`.


zenglingbiao:
@clawsweeper re-review

PR #139897, opened 2026-09-06
improve: skip bind retries in occupied-port startup fixtures

steipete:
## What Problem This Solves

Resolves unnecessary bind retries in native Gateway startup-cleanup tests. Two fixtures intentionally occupy a port for the entire scenario, but the listener still waits through its default twenty 500-millisecond retries before testing cleanup.

## Why This Change Was Made

The existing recording spy delegates to the real listener with its supported `retryEaddrinuse: false` option. It still attempts a native bind and observes the actual EADDRINUSE error. All three isolated fixture scenarios, error identity, journal, retained-owner, cleanup, and successor-admission assertions remain unchanged.

Default retry policy is independently covered by listener tests, including transient contention, exhaustion, explicit disablement, and other failures. No production defaults, retry behavior, test timeout, or shard count change.

## User Impact

Less deliberate waiting in developer and CI startup-cleanup tests. Runtime behavior is unchanged.

## Evidence

All three scenarios passed in the original baseline, candidate, and original-code control. Candidate command time was 80.96 seconds, versus 104.30 seconds initially and 95.90 seconds in the control. The two changed cases totaled 53.387 seconds versus 64.141 seconds in the control.

Cold-process costs varied, including a 3.321-second difference in the unchanged TLS case. The source removes two configured ten-second retry sequences, but these measurements do not establish a precise twenty-second whole-suite saving.

- Initial paired comparisons used Node 26.8.1 and matching dependencies; all native child-exit and journal checks passed.
- Final contained proof on Node 24.19.0 with freshly frozen-installed dependencies passed 21 tests across startup fixtures, HTTP listener policy, and runtime-state integration.
- Full changed-file gate, formatting, and independent P2 review passed.
- Production delta zero; test support +3/-2. No new mocks or removed assertions.

Validation: `node scripts/run-vitest.mjs src/gateway/server.startup-fixture-lifetime.test.ts src/gateway/server/http-listen.test.ts src/gateway/server-runtime-state.test.ts --maxWorkers=1` and the changed-file gate for `server.fixture-lifetime.test-support.ts`.

Exact-head [CI 34019293357](https://github.com/openclaw/openclaw/actions/runs/34019293357) is green. Its startup cases took 11.994, 11.662, and 11.035 seconds (34.691 seconds summed), with all three cases and eight fixture-lifecycle siblings passing. The earlier main sample totaled 57.987 seconds for the same three names; this is a cross-head observation, separate from the controlled local pair.

PR #122595, opened 2026-08-12
fix(gateway): close startup trace on terminal failure

vincentkoc:
Supersedes #80392

## What Problem This Solves

Gateway startup failures could leave the startup trace's event-loop delay monitor active when startup terminated before readiness.

## Solution

The concrete result of `createGatewayStartupTrace` now has an idempotent close operation without widening the exported `GatewayStartupTrace` interface.

Ownership is explicit:

- Bootstrap owns cleanup until it returns successfully.
- Kernel owns cleanup until `prepareGatewayLifecycle` succeeds.
- Lifecycle then owns cleanup and closes tracing before awaited terminal-failure teardown.
- Recovered channel startup errors remain non-terminal and later startup trace phases are preserved.

`server-start.ts` is unchanged. The final patch changes only four production files and three focused test files.

## Evidence

- Frozen defect-reproduction main: `b57c5f25a9142ac8e1195e8552c54c268c2e03f8`.
- Refresh main: `faec7a7ca239459da42cdae91a1ac481bb107215`.
- Published head: `18e8cc2ce82db5f03af664da58cedfd397d25aa6`.
- GitHub verifies the published merge-main commit signature. Its parents are the hosted PR head `beec93e85c6712ffe49e2dc734db1aa497de912b` and refresh main `faec7a7ca239459da42cdae91a1ac481bb107215`.
- The signed original commit `a8b10183f3cf38f65f4f5eb61d2cdc9be6666d8c` remains in ancestry with `Co-authored-by: Evan Newman <180847219+ejames-dev@users.noreply.github.com>`.
- Pre-fix invalid-config reproduction against frozen main failed for the intended reason: expected one monitor disable call, received zero.
- Focused Vitest passed 5 files and 136 tests across startup trace, lifetime, post-attach, unchanged kernel, and unchanged startup-fixture lifetime behavior.
- Idempotent close, invalid config, TLS failure, public-listener finalization failure, measured errors, recovered channel startup, and later trace measurement are covered.
- `check:changed --dry-run` selected core and core-test lanes. The full plan completed as a clean composite: expensive typecheck/lint/dead-export lanes passed; wrapper-terminated terminal guards were rerun directly and passed.
- Oxfmt, `git diff --check`, test-audit, and exact-head max-P1 autoreview passed.
- Production delta: +22/-2, net +20. Total: 7 files, +237/-3.
- Prior Blacksmith Testbox changed gate passed ([public Actions run](https://github.com/openclaw/openclaw/actions/runs/31603197621)).

[...]

PR #140775, opened 2026-09-07
chore(tests): avoid retry waits in startup tracing coverage

steipete:
## What Problem This Solves

The public-startup tracing regression retries an occupied port twenty times even though its own blocker remains open until cleanup. The case took 10.361 seconds in [main CI](https://github.com/openclaw/openclaw/actions/runs/34070027186).

## Why This Change Was Made

For this case only, delegate to the real listener with its existing `retryEaddrinuse: false` option. The real TCP blocker, public Gateway startup, native bind failure, exact-once trace shutdown assertion, and awaited cleanup remain intact. A case-local spy is restored in `finally` before the blocker and state are cleaned up.

Default retries retain their independent listener tests. This follows the same permanently occupied fixture-port approach already used in #139897. Related tracing repair: #122595.

## User Impact

Faster tracing regression coverage, with no production behavior, timeout, dependency, shard-count, or test-inventory change. The test-only diff is +20/-10 lines, including indentation; production delta is zero.

## Evidence

- Full startup-lifetime file: all six cases passed before and after. Four standalone listener-policy tests also passed. Cleanup was subsequently spelled with explicit `finally` to satisfy lint; the listener delegation and original assertions are unchanged.
- Existing startup trace: `http.listen` **10,145.4 ms → 0.7 ms**, preserving the native failure and trace-close assertion.
- Local whole-command timing was noisy: 328.69 → 352.83 seconds, RSS 2.213 → 2.167 GB. No whole-command speedup is claimed from that comparison; CI will provide the representative case timing.
- Earlier local attempts stalled during setup or timed out in the preceding, unchanged TLS case. The full file subsequently passed without changing its deadline. An optional trace-close counterfactual did not reach an assertion and is not counted as regression proof; the original lifecycle source was restored and hash-verified.
- Required changed checks passed through typecheck and all dead-export guards; final lint passed after the naming/cleanup follow-up. Fresh independent P0–P2 review of the final bytes found no actionable defect.

### DIFF
--- a/src/gateway/server-startup-lifetime.test.ts
+++ b/src/gateway/server-startup-lifetime.test.ts
@@ -123,16 +123,26 @@ describe("Gateway startup lifetime", () => {
     });
     state.applyEnv();
     try {
-      const { startGatewayServerCore } = await import("./server-start.js");
-      await expect(
-        startGatewayServerCore(port, {
-          auth: { mode: "token", token },
-          bind: "loopback",
-          controlUiEnabled: false,
-          sidecarStartup: "defer",
-        }),
-      ).rejects.toThrow("another gateway instance is already listening");
-      expect(startupTraceEventLoopDelay.instances[0]?.disable).toHaveBeenCalledOnce();
+      const listenModule = await import("./server/http-listen.js");
+      const listen = listenModule.listenGatewayHttpServer;
+      // The owned blocker cannot leave; retry policy has its own listener tests.
+      const listenSpy = vi
+        .spyOn(listenModule, "listenGatewayHttpServer")
+        .mockImplementation((params) => listen({ ...params, retryEaddrinuse: false }));
+      try {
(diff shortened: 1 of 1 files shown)


steipete:
The exact patch is already on main in [a681f6f42676](https://github.com/openclaw/openclaw/commit/a681f6f4267682ee9743beca30979e3431af0355). The sole changed file has the identical Git blob (`c20b2eeceea228fb88ef1481ee36199d30c767e9`) at main and the reviewed head `f56e727ddb6070af5d5bd1968eadc348d9cd2e2c`.

GitHub still reports this PR as open without a merge association, and the native recovery command confirms there is no net change to merge. Closing it as already applied. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34082506199) passed in 5m42s with all six cases; the real listener phase fell from 10008.2ms to 0.8ms while preserving native bind failure and trace cleanup.

PR #136274, opened 2026-09-02
feat(msteams): resolve team SharePoint site dynamically for channel file uploads

a-tokyo:
Closes #136271

## What Problem This Solves

Fixes: Teams channel members see "request access" when opening files the bot sent. The bot uploaded every file to one configured SharePoint site (`sharePointSiteId`). The native file card links via `webDavUrl`, which requires membership of that site — not membership of the Teams team.

In multi-team deployments this is a production failure: members of a team can read the channel message but cannot open the attached file unless they also have access to the bot's central SharePoint site.

## User Impact

User impact: members of a standard Teams channel can open files the bot sends, because those files land on that team's own SharePoint site. Operators can omit `sharePointSiteId` for standard channels. Group chats and private/shared channels still require an explicit site. Existing deployments with `sharePointSiteId` set are unchanged.

New optional Graph permission for automatic discovery: `Channel.ReadBasic.All` (so private/shared channels are rejected before parent-team upload). Uploads still need `Sites.ReadWrite.All`.

`sharePointFolder` stays in this change with a strict contract: one folder name, no path separators, URL-encoded. Default remains `OpenClawShared`.

## Why This Change Was Made

When `sharePointSiteId` is omitted, a standard channel's file send resolves the team's backing site with `GET /groups/{groupId}/sites/root` and uploads there. Team membership already grants read access on that site, so the file card works.

Resolution is lazy at the upload boundary in `send.ts` and `messenger.ts`. Text-only messages never call Graph for a site ID.

Cold-start after restart: both upload entry points pass the Teams SDK `app.api.teams.getById` lookup into `resolveUploadSiteId` → `resolveTeamGroupId`, so a stored conversation can still resolve the AAD group when the process cache is empty.

Private and shared channels are not auto-discovered. Graph `membershipType` is checked first; anything other than `standard` fails closed unless an explicit `sharePointSiteId` is set. Group chats still require `sharePointSiteId` (no team site).

Site discovery uses the abortable request owner so a stalled `/sites/root` fetch is cancelled when the deadline expires. Folder names are validated as a single segment and encoded before interpolation.

## Evidence

[...]

PR #135162, opened 2026-09-01
perf(media): avoid intermediate multipart upload copies

steipete:
## What Problem This Solves

Upload builders copy each Buffer into a new typed array before Blob or File immediately snapshots it again. Large audio, image, video, workflow and attachment uploads therefore hold an avoidable second full-size copy during construction.

## Why This Change Was Made

Use one dependency-free `bufferToBlobPart` implementation for the core audio adapter and eleven official plugins (thirteen upload owners). Ordinary ArrayBuffer-backed Buffers become exact offset/length views; SharedArrayBuffer backing is copied into an ordinary ArrayBuffer. Every caller constructs its Blob or File synchronously, preserving the existing snapshot boundary. Delete the duplicate Comfy, Feishu and OpenAI video conversion helpers and the other intermediate copies.

The implementation lives directly in the private `blob-runtime` SDK entrypoint. Importing broader provider/fetch facades added unrelated runtime dependencies in measured plugin imports, so this leaf has no imports. Its built runtime closure is two files totaling 451 bytes. Canonical SDK generation adds its JavaScript-only package export and local declaration aliases; the typed public SDK counts remain 147 entrypoints, 4,359 exports and 2,594 callables, with no budget changes.

## User Impact

A controlled 32 MiB upload through the actual core multipart builder reduces ArrayBuffer growth during Blob construction from 67,111,808 to 33,557,376 bytes: one 33,554,432-byte intermediate copy removed. The wire bytes, filename, MIME type, form-field order and mutation-after-construction behavior remain identical. This measures upload-construction allocation, not a whole-Gateway resident-memory reduction.

Production source is net -8 lines; the two maintained SDK classification entries bring production plus maintained metadata to net -6. Generated package/path metadata is +10, documentation +7 and tests +8. Existing OpenAI video tests now protect offset Buffers and snapshot timing; the existing Mattermost HTTP loopback test also asserts exact uploaded bytes.

## Evidence

- Actual-source baseline/candidate probes cover nonzero-offset Buffers, sentinels outside their visible ranges, Blob/File snapshots and SharedArrayBuffer input. A normalized multipart comparison produces identical 466-byte wire payloads. The extracted OpenAI video reference owner independently drops ArrayBuffer growth from 64 MiB to 32 MiB for a 32 MiB input; both image and video metadata/validation remain identical. Source-extracted LINE, Mattermost, PixVerse and Tlon owners preserve bytes and metadata while removing each adapter allocation.
[...]

PR #123244, opened 2026-08-13
fix(msteams): validate Graph upload responses against malformed envelopes

coaiMax:
## What Problem This Solves

MS Teams Graph API file upload operations could throw a raw TypeError when a successful response body was null or a non-record value (e.g., a primitive or array) instead of the expected JSON object. All four Graph response handlers in graph-upload.ts were affected: SharePoint upload, driveItem properties, chat members, and sharing link creation.

## Why This Change Was Made

Each `readProviderJsonResponse` call previously used a typed generic (e.g., `{ id?: string; ... }`) that assumed the parsed JSON was always an object. At runtime, a null or primitive response would pass through without error, and the subsequent property access (`data.id`, `data.value`, `data.link`) would throw `TypeError: Cannot read properties of null (reading 'id')`.

Following the same pattern as the OpenRouter video catalog fix (PR #111637), all four handlers now:
1. Read the response as `unknown` via `readProviderJsonResponse<unknown>`
2. Validate the payload with `isRecord()` before accessing any property
3. Use explicit `typeof` checks for string fields and `Array.isArray` for array fields

## User Impact

MS Teams file sending continues to work normally for well-formed Graph responses. A malformed Graph response now produces a clear, actionable error message (e.g., "SharePoint upload response missing required fields") instead of a raw TypeError stack trace.

## Evidence

- Before fix: A null Graph success body would crash with `TypeError: Cannot read properties of null (reading 'id')` at `data.id` in `uploadToSharePoint`.
- After fix: `isRecord(data)` returns `false` for null, and the existing "missing required fields" error path is taken instead.
- The same defensive pattern is applied to `getDriveItemProperties`, `getChatMembers`, and `createSharePointSharingLink`.
- `isRecord` is imported from `openclaw/plugin-sdk/string-coerce-runtime`, the same module used by the OpenRouter, BytePlus, Runway, xAI, and Google video/image generation providers.
- All 48 CI checks pass on the latest commit (`061c480`), including `check-lint`, `check-test-types`, and `check-prod-types`.

## Regression Test Plan

Added 3 regression tests to `extensions/msteams/src/graph-upload.test.ts` (commit `061c480`):

1. **`rejects a null upload response body`** — sends `null` as the JSON response body, verifies the error "SharePoint upload response missing required fields" is thrown instead of a TypeError.
2. **`rejects a primitive string upload response body`** — sends `"not-an-object"` as the JSON response body, verifies the same error.
[...]

PR #140787, opened 2026-09-07
perf(channels): snapshot raw Slack and Teams uploads as blobs

### LINKED ISSUE #140786: Raw Slack and Teams uploads retain avoidable BufferSource copies
Slack external file upload, Teams Graph upload and Teams file-consent upload create a Uint8Array snapshot for a raw fetch body. Fetch treats that as BufferSource and copies its bytes again; the queued byte-stream path can retain another copied representation before connection admission.

These sites already require an immediate immutable snapshot after their existing prerequisites: Slack upload-URL acquisition, Graph token acquisition and consent URL/DNS validation. The existing public bufferToBlobPart helper can supply the exact Buffer slice to a Blob constructor at those same expressions. Explicit MIME/range headers, mutable-input timing, shared backing, validation, deadlines and completion must remain unchanged.

An actual 16 MiB Slack owner run records 16,777,216 bytes copied by package Undici's BufferSource helper; the Blob candidate records zero for that helper and produces identical real local HTTP bytes. A separate gated-connection transport experiment shows a 32 MiB ArrayBuffer difference for a 16 MiB request. That component result is not a fixed whole-upload peak/RSS saving; Blob still owns its immutable snapshot and shared backing can require a temporary helper copy.

Related #135162 introduced the existing helper and multipart reuse, excluding these raw Slack/Teams bodies. This issue concerns only the three raw-body snapshot expressions, not structured message preparation or provider API behavior.

steipete:
Closes #140786

<details>

</details>

## What Problem This Solves

Slack and Teams raw file uploads pass Uint8Array snapshots to fetch, which copies those BufferSource bytes again. The queued transport can retain additional copied representations before connection admission.

## Why This Change Was Made

Use immediate Blob snapshots through the existing public bufferToBlobPart helper at the same three expressions: Slack after upload-URL acquisition, Graph after token acquisition, and file consent after URL/DNS validation. Exact Buffer slices and SharedArrayBuffer-backed inputs remain protected. Blob.type stays empty, leaving existing explicit MIME/range headers authoritative.

## User Impact

The actual 16 MiB Slack owner records 16,777,216 bytes copied by the real package BufferSource helper before, and zero afterward. Twenty real local HTTP upload scenarios plus two consent rejection controls preserve complete payloads, headers, result values, failures, completion order and deadline values across Slack, Graph and consent owners.

Separately, native/package transport probes at gated connection admission measured 32 MiB fewer retained ArrayBuffer bytes for a 16 MiB Blob request. Response-wait deltas differ and Blob native storage accounting is not the same as external-memory accounting; no fixed per-upload peak/RSS or whole-transfer saving is claimed. Blob still owns one immutable snapshot, and shared backing can need a temporary helper copy. Small transport timing controls showed no material regression, without establishing general throughput gains.

## Evidence

- Actual source owners run through real media loading/guards and a synthetic local HTTP server. Accepted responses traverse the real fetch transport; cases include ordinary/sliced/shared backing, mutation during prerequisite awaits and after snapshot, raw 503 failures, late completion errors, timeouts and rejected consent destinations.
- Long Slack/Graph deadlines use clock-controlled invocation of the already-armed owner timer after the real request reaches the server; this does not claim that full 120/301-second deadlines elapsed. Consent uses its real configurable short timeout.
- 169 focused tests across six files, the complete changed-file gate, normal build with SDK/bundled-plugin validation, and final isolated P2 review passed on the final seven-file candidate.
- Tests protect bytes, exact slice boundaries, snapshot timing, explicit/absent MIME, completion and cancellation. The old Slack send suite is unchanged; two focused cases live beside the upload owner.
- Production is net +3 lines, tests net +108. The assertion baseline shrinks only client-delivery.ts from five to four after removal of its BodyInit assertion. No dependency or configuration change.

Credit #135162 for the existing Blob helper. Active #119737, #123244 and #136274 address distinct receipt, response-validation and SharePoint-resolution behavior; their independent contracts must be preserved if they land first.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -1105,7 +1105,7 @@ extensions/slack/src/blocks-render.ts	1
 extensions/slack/src/channel-actions.ts	1
 extensions/slack/src/channel-type.ts	1
 extensions/slack/src/channel.ts	7
-extensions/slack/src/client-delivery.ts	5
+extensions/slack/src/client-delivery.ts	4
 extensions/slack/src/client-options.ts	5
 extensions/slack/src/directory-live.ts	2
 extensions/slack/src/doctor.ts	1
--- a/extensions/msteams/src/file-consent.test.ts
+++ b/extensions/msteams/src/file-consent.test.ts
@@ -325,7 +325,7 @@ describe("uploadToConsentUrl", () => {
       "Content-Type": "application/octet-stream",
       "User-Agent": buildUserAgent(),
     });
-    expect(opts?.body).toEqual(new Uint8Array(Buffer.from("hello")));
+    expect(Buffer.from(await new Response(opts?.body).arrayBuffer())).toEqual(Buffer.from("hello"));
   });
 
   it("aborts consent uploads that do not finish before the request timeout", async () => {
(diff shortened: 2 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [69f8e39d509e5f947752f8684da5448841a95036](https://github.com/openclaw/openclaw/pull/140787/commits/69f8e39d509e5f947752f8684da5448841a95036)
- Landed commit: [fdd1da1e5b81681f09b14ff9497390abecdf7fc5](https://github.com/openclaw/openclaw/commit/fdd1da1e5b81681f09b14ff9497390abecdf7fc5)

PR #140792, opened 2026-09-07
refactor(google-meet): test platform guards through the tool

steipete:
## What Problem This Solves

Google Meet's platform test called a private predicate directly. It stayed green when the registered tool stopped invoking the platform guard, leaving the unsupported-host error path unprotected.

## Why This Change Was Made

Exercise the same four platform/mode/transport cases through the registered tool. Allowed cases assert their result and Gateway dispatch; the Windows case asserts the existing actionable error and no additional dispatch. Remove the private predicate's testing export and its test-only platform argument, using the existing platform producer and fixture setters.

## User Impact

No runtime behavior, supported platform, configuration, or API changes. The production diff removes two lines; the test grows by 22 lines to preserve all four cases at the tool boundary. Existing routing, authorization, browser ownership, lifecycle, transcript, and audio coverage stays intact.

## Evidence

- Both owning suites passed before and after: 178/178 tests, with every test name retained.
- Caller regression control: removing the tool's guard call leaves all 178 old cases green; the revised suite fails exactly the Windows case because it returns success instead of the existing platform error. The fault was removed and candidate bytes restored exactly.
- Read the original platform-guard history and Linux support change (#118451), the complete owning suites, registration/dispatch/config owners, and existing fixture lifecycle. Tests use the existing synthetic Gateway stub; no real meeting, provider call, microphone, or browser is involved.
- Independent applied review is clean through P2. All 25 required changed-file checks passed through the configured Blacksmith Testbox route ([run 34082784920](https://github.com/openclaw/openclaw/actions/runs/34082784920)); command exit 0 in 10m49s. The lease completed and its temporary checkout was removed. PR-head CI is tracked separately.

### DIFF
--- a/extensions/google-meet/index.test.ts
+++ b/extensions/google-meet/index.test.ts
@@ -1467,41 +1467,63 @@ describe("google-meet plugin", () => {
     });
   });
 
-  it("keeps local Chrome talk-back available on Linux and blocks unsupported hosts", () => {
+  it("keeps local Chrome talk-back available on Linux and blocks unsupported hosts", async () => {
     const { cliRegistrations, methods, tools } = setup(undefined, { registerPlatform: "linux" });
+    const tool = getMeetTool({ tools });
+    const callGatewayFromCli = vi.fn(async () => ({ ok: true }));
+    googleMeetPluginTesting.setCallGatewayFromCliForTests(callGatewayFromCli);
 
     expect(tools).toHaveLength(1);
     expect(cliRegistrations).toHaveLength(1);
     expect(methods.has("googlemeet.setup")).toBe(true);
-    expect(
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [976594bbdc19c68e1542992dcbf4515ebc003f9e](https://github.com/openclaw/openclaw/pull/140792/commits/976594bbdc19c68e1542992dcbf4515ebc003f9e)
- Landed commit: [613e2f4eb72e3796f0b67466a07f4a1d0cc092c7](https://github.com/openclaw/openclaw/commit/613e2f4eb72e3796f0b67466a07f4a1d0cc092c7)

PR #140793, opened 2026-09-07
feat(sessions): split deleted archive retention

jjjhenriksen:
## What Problem This Solves

`session.maintenance.resetArchiveRetention` previously applied the same age cutoff to both reset and deleted transcript archives. This adds `session.maintenance.deletedArchiveRetention` as an independent duration-or-`false` control.

When omitted, `deletedArchiveRetention` inherits `resetArchiveRetention` for backward compatibility. An explicit duration or `false` overrides it independently. Both settings age archive artifacts; they do not age retained SQLite reset history or transcript rows.

## Implementation

- Resolves independent deleted/reset rules from configuration.
- Applies those rules through legacy file-backed maintenance, cron reaping, and ordinary SQLite maintenance.
- Ordinary SQLite maintenance runs archive cleanup even when no session rows are removed.
- Removes a derived archive file before deleting its published SQLite recovery row; failed file cleanup leaves the recovery row intact.
- Preserves inherited deleted retention for older resolved-maintenance callers while retaining explicit disablement.
- Fences archive-row pruning against maintenance-owner invalidation before and inside each SQLite writer callback.
- Updates schema help, examples, and session-management documentation.

## Evidence
- Exact pushed head: `5460eaf7b0ab2221938f1b7a4dbfc052e69d9588`.
- `git diff --check` passed.
- Focused runtime command:
  `pnpm exec vitest run src/config/sessions/session-accessor.sqlite-retention.integration.test.ts src/config/sessions/session-accessor.sqlite-maintenance-writer.test.ts src/config/sessions/store.pruning.archive-retention.test.ts src/config/sessions/cleanup-service.fix-missing.test.ts --reporter=verbose`
  Result: **4 test files passed, 22 tests passed**.
- The unmocked integration proof loads a real config file and exercises both paths:
  - a fresh configured store is initialized (`configuredFreshStore: true`);
  - an existing configured store is maintained with independent cutoffs (`configuredExistingStore: true`);
  - the expired deleted archive file and its SQLite recovery row are absent after cleanup (`deletedArchiveFile: false`, `deletedArchiveRow: false` — `false` here means absent);
  - the reset archive remains (`resetArchiveFile: true`).
- The focused tests additionally prove inherited retention for older resolved callers, explicit `false` disablement, ordinary SQLite archive cleanup with no session removals, maintenance-owner invalidation without reopening the store, file-before-row deletion ordering, and lifecycle-deletion failure recovery.
- This proof is on the exact pushed head above; no claim is made that it resolves the separate current-base merge conflict or pending maintainer acceptance.
## Known Gate

This change affects persistent retention and upgrade behavior. Maintainer acceptance of the independent retention contract, ordinary-maintenance deletion timing, and the accessor cleanup boundary remains required before merge.

## Worked on by

- Jacqueline Henriksen (@jjjhenriksen)


jjjhenriksen:
@clawsweeper re-review

jjjhenriksen:
@clawsweeper re-review

jjjhenriksen:
Maintainer-readiness follow-up: the implementation findings are addressed, but the remaining gate is upgrade safety for the retention contract. The missing proof is a pre-change store upgraded in place, plus explicit-false behavior and recovery after a failed file removal; the cleanup-boundary/config-compatibility decision also needs maintainer acceptance. I’ll target those exact cases on the next update, then request a fresh ClawSweeper review on the resulting head.

jjjhenriksen:
Exact-head integration proof pushed at `8f4f52bc771`.

The SQLite/file integration now gives deleted and reset archives the same one-day age, applies independent cutoffs (`deletedArchiveRetention: 1ms`, `resetArchiveRetention: 7d`), and proves:
- a pre-existing configured store is upgraded in place
- the deleted derived file and canonical recovery row are removed
- the equal-age reset archive remains

The targeted retention set also proves explicit disablement remains disabled and a canonical row is retained while its derived file still exists (the failed/not-yet-completed removal case).

Validation: 3 files, 13 tests passed.

This supplies behavior evidence only; it does not substitute for the requested maintainer decision on the retention contract/deletion timing.

PR #140130, opened 2026-09-06
fix(android): show model status for the selected agent

obviyus:
## What Problem This Solves

Android model settings could show the default agent’s credentials after another agent was selected, and delayed responses could restore stale state after switching away and back.

## Why This Change Was Made

Model readers now carry explicit agent selection. The main-session owner clears state and invalidates pending reads before publishing the change; Refresh requests fresh discovery through the existing API. Default requests preserve older-Gateway compatibility.

## User Impact

Provider and model settings follow the selected agent. Late replies cannot overwrite the new selection, and existing upgrade guidance and reconnect behavior remain.

## Evidence

A real Android emulator and isolated Gateway reproduced the wrong credential row before repair and showed the selected agent afterward. Held success/error replies, away/back switching, explicit refresh, and recovery retained correct visible state. Simulated older-Gateway rejection kept upgrade guidance and independent credential status. All 65 focused tests, Android lint, and the debug build passed. Compatibility was simulated rather than run against an old server.

Related: #136257.

PR #136257, opened 2026-09-02
feat(models): direct model lists and provider login across surfaces

obviyus:
## Session-Only Model Selection

An ordinary admin model choice in a chat could also rewrite the agent or shared default. The omitted scope used a legacy `effective` policy that inferred a configuration target from the agent's current setup.

Commit `9de42a0e26804c4e04d3f02a18c03e03a77d65a7` makes ordinary selections session-only across the shared command and Gateway policy. It removes the effective-scope branch and requires an explicit target in the internal default writer. Explicit `-a`/`-g` commands and configured agent/global scope retain their authorization and behavior. Public SDK callers that explicitly request persistence keep the shipped v2026.9.1 contract.

This intentionally changes the unset `modelSelectionScope` behavior. Owners who want broad persistence must request an agent/global scope or configure it explicitly. No new configuration key, schema, or protocol field is added.

The writer audit covered direct and mixed commands, Gateway patches and projections, channel pickers, SDK entry points, setup, migration, and CLI default-setting commands. Source review found no P0-P2 defects in the correction. Automated tests, CI, rebasing, and merge validation remain deferred. This correction removes 25 production lines, 5 test lines, and 1 documentation line net.

The corrected package is deployed. The same read-only admin request reports `modelSelectionTarget: "global"` before the correction and `"session"` afterward; the configured default remains unchanged. The Gateway is healthy and the served Control UI identifies this commit. A verified state backup preceded the graceful service switch. Full manual interaction checks remain pending.

## Catalog Cutover Checkpoint

Ordinary model-list requests read the published catalog. Startup, lifecycle changes, and explicit refresh own discovery. Existing chats, new chats, commands, and native pickers use the same inventory and availability decisions.

- The online CLI reads the running Gateway; offline listing identifies its cached source. `models list --refresh` requests discovery explicitly.
- Generated inventory no longer supplies credentials or request headers. Current credentials, explicit restrictions, and authored model routes remain authoritative.
- Bundled provider discovery reports successful empty results and acquisition failures separately. The shared catalog owner retains compatible last-good inventory after failure.
- Provider aliases, session profile locks, and runtime choices use shared core decisions. Web and native clients consume publication changes; Android negotiates command metadata and model-catalog capabilities independently.
- Published external SDK contracts remain at their adapters. Bundled callers use the new strict discovery path.

[...]

PR #140796, opened 2026-09-07
fix(android): keep open model settings current

obviyus:
Related: #136257, #140130.

## What Problem This Solves

Android Providers & Models stayed stale after the Gateway accepted model-name or credential changes. Manual Refresh worked, but the open page ignored publication events.

## Why This Change Was Made

The current connection's configuration and metadata events refresh existing model and provider readers through their selected-agent and connection guards. Event-driven reads remain passive; explicit Refresh retains discovery behavior.

## User Impact

The open page follows accepted configuration changes and provider sign-out without navigation or reconnecting. Agent switching, visible errors, and reconnect-to-Gateway-default behavior remain intact. No configuration, protocol, or authentication-policy change is added.

Consumers: Android model and provider settings now reread published state on current-connection events.

## Evidence

The original client remained stale until manual refresh. A real emulator showed sign-out and two model-name changes on the open page within 15 seconds. Native checks covered passive versus explicit requests, selected-agent reads, event bursts, stale responses, rejection recovery, and reconnect behavior. All 36 focused owner/request tests passed; the event regression failed before the fix. Provider data was synthetic, with no inference. Sampled captures do not prove every frame or replacement-socket callback race.

### DIFF
--- a/apps/android/app/src/main/java/ai/openclaw/app/NodeRuntime.kt
+++ b/apps/android/app/src/main/java/ai/openclaw/app/NodeRuntime.kt
@@ -5507,6 +5507,10 @@ class NodeRuntime private constructor(
     if (event == GatewayEvent.VoicewakeChanged.rawValue) {
       applyVoiceWakeWords(payloadJson)
     }
+    if (operatorConnected && (event == "config.changed" || event == "chat.metadata.changed")) {
+      refreshModelCatalog()
+      refreshProviderModels()
+    }
     if (event == "config.changed" || event == GatewayEvent.UsersPrefsChanged.rawValue) {
       // Config changes invalidate the snapshot; profile changes are targeted by
       // the gateway to connections bound to our own profile.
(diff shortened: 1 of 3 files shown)


obviyus:
Thanks @obviyus!

**Proof:** On an Android API 36 emulator, real Gateway publications update the open Providers & Models page without navigation, reconnect or manual Refresh. Synthetic credential removal and two distinct model-name changes appeared within 15 seconds. All 36 focused Android tests pass.

Exact-head CI passed. The final independent review found no actionable defects, cleared security, and rated the repair platinum with sufficient native proof.

- **Bug:** The open page kept old model and credential state after Gateway changes.
- **Fix:** Publication events call the existing guarded model and provider readers. Passive reads do not force discovery; explicit Refresh still does.

| Before | After |
| --- | --- |
| Removed credential remains visible after publication. | The open page removes the credential row. |
| Changed model name stays stale until Refresh. | The open page shows each newly published name. |

The original reconnect test remains recorded as five passes and one failed expectation. Independent shipped-source review justified correcting that expectation: a replacement connection uses its advertised default agent. A fresh native probe passed that clause within the original deadlines. Releases addressed the closed old socket; replacement-socket callback delivery and reconnect injected-error coverage are not claimed. Tests use isolated synthetic credentials, not live vendor authentication.

PR #140820, opened 2026-09-07
fix(ai): prevent truncated Anthropic streamed replies

steipete:
<details>

Keep **Allow edits from maintainers** enabled so maintainers can help update the branch.

</details>

## What Problem This Solves

Fixes an issue where developers consuming Anthropic streams through the standalone adapter received truncated or empty replies when the response included text or thinking in its initial stream events. A response starting with `Hello` and continuing with ` world` returned only ` world`.

## Why This Change Was Made

Both Anthropic adapters now preserve initial content through their shared stream reducer. They also share the completed-signature accumulation path: an initial signature survives when no replacement arrives, and streamed signature data replaces that initial value. Production code shrinks by four lines.

## User Impact

Standalone Anthropic consumers retain the complete streamed reply and thinking content, matching the managed adapter and native SDK. The usual empty-start sequence continues to work.

## Evidence

The two seeded regression cases failed before production changes; the empty-start control passed. After repair, both adapters matched the pinned Anthropic SDK 0.120.0 across three event sequences and nine real local HTTP requests. All 278 related tests passed, followed by a final 18-test parity run. The full package test typecheck, full-file lint, formatting, max-lines ratchet, and independent P2 review passed. Regression tests add 70 lines, and the corrected snapshot adds 10 net lines.

The HTTP proof uses synthetic local SSE responses. Current first-party frequency of nonempty block starts was not measured: request parameters cannot force whether Anthropic places content in a start event or a later delta, so a live smoke would not reliably exercise this case. The regression compares the actual native SDK parser and both adapters against seeded, seed-only, and empty-start sequences.

### DIFF
--- a/packages/ai/src/provider-transport-parity.test.ts
+++ b/packages/ai/src/provider-transport-parity.test.ts
@@ -1,4 +1,5 @@
 import path from "node:path";
+import Anthropic from "@anthropic-ai/sdk";
 import { APIError as AnthropicAPIError } from "@anthropic-ai/sdk/core/error.js";
 import type { AssistantMessageEventStreamLike, Model } from "@openclaw/llm-core";
 import { describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [843ea40892f2f4b6f22275c49734a8b450fb5cea](https://github.com/openclaw/openclaw/pull/140820/commits/843ea40892f2f4b6f22275c49734a8b450fb5cea)
- Landed commit: [f02c72916aeb374a5a5da9e36df2751fb38f2221](https://github.com/openclaw/openclaw/commit/f02c72916aeb374a5a5da9e36df2751fb38f2221)

PR #140823, opened 2026-09-07
refactor(time): isolate browser-safe duration parts

steipete:
## What Problem This Solves

Pure duration quantities currently share a module with the `pretty-ms` presentation adapter. A browser-safe owner is needed before consolidating Workboard's separate duration decomposition through a supported plugin contract.

## Why This Change Was Made

The existing constants, types, and resolver bodies move unchanged into `src/infra/format-time/duration-parts.ts`. Core formatters and the Dashboard import that owner; `pretty-ms` remains in the formatting adapter. This is the bounded preparatory part of the duration consolidation.

## User Impact

None intended. Existing duration strings and arithmetic are unchanged. Workboard's adjacent-unit behavior and the chat recap formatter remain unchanged. No SDK export, configuration, dependency, schema or protocol change.

## Evidence

- `node scripts/run-vitest.mjs src/infra/format-time extensions/workboard/browser/lib ui/src/lib/format`: 433 tests passed across eight files and three shards in 26.20s.
- Moved arithmetic blocks are byte-identical to their original implementations.
- Autoreview: P0-P2 scoped-clean, zero findings.
- `node scripts/check-changed.mjs --base abc35e08c1a2efd29f356354b7fa0d8374429889`: passed.
- `pnpm ui:build` and `pnpm ui:check-performance`: passed; startup JS 346,454 / 350,953 B gzip with 8 / 18 requests; CSS 43,989 / 51,200 B. Baseline unchanged.
- Full root/package build and full Workboard/SDK integration were not run or implemented for this preparatory draft.
- Production +61/-59 (net +2); no test changes. The small growth is import wiring for the browser-safe owner.

## NEEDS-MAINTAINER-DECISION

Keep this PR as a draft. A focused supported duration SDK entrypoint would increase the verified public entrypoint count from 152 to 153 against the existing cap of 152 in `scripts/plugin-sdk-surface-report.mts`. This lane has no authority to raise that ratchet. The existing `time-runtime` and `number-runtime` seams are private-local; `time-runtime` also imports Node-dependent filesystem code. A maintainer must choose the supported browser-safe duration contract and allocate its API budget before Workboard integration can proceed.

Workboard must retain its adjacent-unit cutoff: 3,630,000 ms displays as `1h`, while core/UI display `1h 30s`; 86,460,000 ms stays `1d`, versus `1d 1m`. It also decomposes rounded seconds through nested division, whereas core uses direct millisecond division and BigInt days. Those arithmetic differences must remain explicit in any shared resolver.

The chat recap loop in `chat-working-indicator.ts` requires body changes owned by lane L10; L09 is permitted only one import-line edit there, which cannot remove that loop. No chat page was edited. Full Workboard/SDK integration is not implemented in this draft.

### DIFF
--- a/src/infra/format-time/duration-parts.ts
+++ b/src/infra/format-time/duration-parts.ts
@@ -1,0 +1,52 @@
+// Browser-safe duration quantities; formatting remains with each presentation owner.
+export const durationUnitMs = {
+  year: 31_536_000_000,
+  week: 604_800_000,
+  day: 86_400_000,
+  hour: 3_600_000,
+  minute: 60_000,
+  second: 1_000,
+  millisecond: 1,
+} as const;
+
+export type DurationPart = { value: number | bigint; unit: keyof typeof durationUnitMs };
+
+function resolveDurationParts(ms: number, unitCount: number, showYears = false): DurationPart[] {
+  const days = BigInt(Math.trunc(ms / durationUnitMs.day));
+  const parts: DurationPart[] = [
+    { value: showYears ? days / 365n : 0n, unit: "year" },
+    { value: showYears ? days % 365n : days, unit: "day" },
+    { value: Math.trunc((ms / durationUnitMs.hour) % 24), unit: "hour" },
+    { value: Math.trunc((ms / durationUnitMs.minute) % 60), unit: "minute" },
+    { value: Math.trunc((ms / durationUnitMs.second) % 60), unit: "second" },
+    // Large floats can retain a remainder after second-rounding; only subsecond input uses ms.
+    { value: ms < 1_000 ? Math.trunc(ms) : 0, unit: "millisecond" },
+  ];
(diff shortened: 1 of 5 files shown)


steipete:
Closing by maintainer decision. The browser-safe duration-parts leaf would add a 153rd public plugin-SDK entrypoint against the surface cap of 152 (`scripts/plugin-sdk-surface-report.mts`), and `src/plugin-sdk/AGENTS.md` says the surface is already too large and should be replaced rather than grown. Workboard's adjacent-unit cutoff and rounded-seconds decomposition are contractual, so the consolidation could not remove them either. The remaining duplicate (about 20–35 lines across `src/infra/format-time`, `ui/src/lib/format.ts`, and the Workboard formatter) is recorded in the de-duplication survey as a known, accepted difference.

PR #137128, opened 2026-09-03
improve: reuse plugin activation preparation per operation

steipete:
## What Problem This Solves

Provider-auth and plugin-contribution lookups repeatedly prepared the same activation policy for each installed owner during one synchronous operation.

## Why This Change Was Made

The installed-index owner now prepares at most two lazy activation sources per operation: original policy and bundled-provider compatibility policy. Keeping them separate preserves non-provider eligibility when compatibility expands an allowlist. Batch callers reuse those prepared facts, and the contribution loader's old one-use wrapper disappears. Singleton callers use the same evaluator with only their selected record.

## User Impact

Eligibility, trust, root selection, current configuration and disabled-owner inspection stay unchanged. There is no new cache, option or persistence surface. Production +13 LOC creates the operation-owned preparation boundary; the existing test change is net zero. Singleton calls add a small one-record array/closure; the measured benefit below applies to batch callers.

## Evidence

- All 39 fixed actual-source observations match across configuration, allowlist, bundled mode, root A/B/A, invalidation, contribution and provider-auth cases.
- Across 40 composed operations, canonical normalizations fell from 3,960 to 280 and compatibility memo reads from 2,440 to 160. Separate sampled allocation estimates fell from 33,932,680 to 15,792,240 bytes; these are not retained heap or Gateway RSS.
- 94 existing tests across seven files pass, as does the canonical changed-file gate. Managed Codex review through P2 is clean.
- Real Gateway verification with MCP and session features passed on a composite containing this exact patch; details below.

## Real Gateway integration (2026-09-03)

A prebuilt, isolated Gateway with a real OpenAI API key passed four fixed turns: ordinary completion, real `web_fetch` of example.com, a 1.31 MB Unicode return through Code Mode, and a real MCP stdio tool with default and explicit arguments. Fifty-seven Gateway RPCs verified three session identities, full/repeated/paginated/subscribed list rows, descriptions and final previews, transcript completion, an explicit reconnect, empty approval replay and unsubscribe.

The MCP server observed the default `id: alpha` before server-side validation, followed by explicit `id: beta`; both nested call identities/order, structured results and metadata projection matched. The Unicode prefix remained well formed and within its byte budget with exact omitted-byte accounting. All four turns completed without retries.

[...]

PR #140824, opened 2026-09-07
perf(plugins): speed up tool preparation with installed plugins

steipete:
## What Problem This Solves

Resolves a problem where preparing agent tools repeats a full plugin configuration scan for every installed plugin. With many installed tool plugins, this adds avoidable work to each turn even when their runtime registry is already prepared.

## Why This Change Was Made

Reuse the existing installed-plugin eligibility predicate for one synchronous tool selection, carrying it through the existing manifest eligibility check. Owner policy and bundled-plugin behavior still take precedence, and each new selection prepares fresh policy. Also carry forward the already-built loader options instead of constructing them twice.

This extends the operation-owned preparation introduced in #137128 to tool selection. No cross-turn cache, configuration, database, dependency or public plugin API change. Production net +4 lines.

## User Impact

Lower tool-preparation overhead when multiple installed plugins contribute tools. Plugin enablement, trust, tool availability, factory context and execution remain unchanged.

## Evidence

- The new regression fails on the original source: eight configuration scans instead of one. Existing behavior controls pass.
- Focused eligibility, tool-resolution and real SQLite compatibility tests pass. The full four-file changed gate and independent P2 review passed.
- Actual native CommonJS plugin loader and tool execution on Linux Node 24.19.0, using 100 generated no-op plugins and A/B/B/A order: 40 warm observations per variant measured baseline medians 2.39/2.30 ms versus 0.84/0.77 ms. Configuration scans fell from 101 to 2. Both variants selected all 100 tools, and disabling/re-enabling one plugin produced 99 then 100 tools.
- Independent macOS Node 26.8.1 comparison also improved. These synthetic measurements establish the removed repeated work, not a claim to resolve unrelated production stalls.

The first full local gate caught an incomplete contribution type in the new test fixture. The fixture now supplies the required fields and uses the required nonmutating clone style; no type or lint checks were weakened. The Linux proof used identical production bytes before that test-only correction and additional compatibility coverage.

### DIFF
--- a/src/plugins/installed-plugin-index.compat.test.ts
+++ b/src/plugins/installed-plugin-index.compat.test.ts
@@ -6,7 +6,11 @@ import { closeOpenClawStateDatabaseForTest } from "../state/openclaw-state-db.js
 import { clearBundledDiscoveryModeMemo } from "./bundled-discovery-state.js";
 import type { PluginCandidate } from "./discovery.js";
 import { refreshPersistedInstalledPluginIndex } from "./installed-plugin-index-store-write.js";
-import { isInstalledPluginEnabled, loadInstalledPluginIndex } from "./installed-plugin-index.js";
+import {
+  createInstalledPluginEnabledPredicate,
+  isInstalledPluginEnabled,
+  loadInstalledPluginIndex,
+} from "./installed-plugin-index.js";
 import { clearPluginMetadataLifecycleCaches } from "./plugin-metadata-lifecycle.js";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e87d8a48df4df460587353dcb81b1d8cebe23a84](https://github.com/openclaw/openclaw/pull/140824/commits/e87d8a48df4df460587353dcb81b1d8cebe23a84)
- Landed commit: [a77442bcfb505b44c7e9ec78cf26ed37975447b5](https://github.com/openclaw/openclaw/commit/a77442bcfb505b44c7e9ec78cf26ed37975447b5)

PR #140832, opened 2026-09-07
fix: identify embedded stage timing log emitters

steipete:
## What Problem This Solves

Resolves a problem where operators reading shared logs could not identify the process and thread that emitted an embedded-run stage timing summary.

## Why This Change Was Made

The existing embedded summary formatter now records the emitter's `pid`, `threadId` and `isMainThread`. This covers startup, prep, core-plugin-tool and auth summaries without changing their timing, warning thresholds or event count. The generic SDK timing helper stays unchanged.

## User Impact

Operators can distinguish emitters sharing a log file. The logging guide explains that these fields identify the summary emitter: elapsed stages can include asynchronous waits and do not establish where every operation ran or how much CPU it used.

## Evidence

- Before the change, the three updated formatter expectations failed while the three existing duration/threshold controls passed.
- A native Node 26.7 baseline used the actual subsystem logger and one shared JSONL file from a true main thread and a joined Worker. All six trace/warning records lacked the emitter tuple; their levels, stage text and quiet control matched the contract.
- Candidate: all six native main/Worker log records include the correct tuple; all 16 focused/sibling tests and the full changed-scope gate pass. Every managed child/group/output is joined. The committed tree exactly matches the tested source carrier.
- Independent Codex review through P2 found no actionable findings. Production delta: +1 line.

The proof uses synthetic identifiers and a controlled clock; it makes no startup-speed or CPU-blocking claim.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -327,6 +327,11 @@ logs each completed slow stage immediately, including failures, and emits a
 `native-turn-handoff` summary before submitting the native turn. Timing records
 contain stage names and identifiers, not prompts or tool arguments.
 
+Embedded-run startup, prep, core-plugin-tool and auth stage summaries include
+`pid`, `threadId` and `isMainThread` in the message to distinguish emitters sharing
+a log file. These identify the summary emitter, not where every timed operation
+ran. Elapsed stage time can include asynchronous waits and is not CPU time.
+
 Use the first `turn_accepted`, `model_call_started`, `tool_execution_started`, and
 `assistant_output_started` milestones to separate startup from later activity.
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3494d15e80f4af4ac7803589c8860c133fffa45e](https://github.com/openclaw/openclaw/pull/140832/commits/3494d15e80f4af4ac7803589c8860c133fffa45e)
- Landed commit: [8507298d1a9d0a8848a99413abd81098bb7f7fb2](https://github.com/openclaw/openclaw/commit/8507298d1a9d0a8848a99413abd81098bb7f7fb2)

PR #140833, opened 2026-09-07
fix(gateway): preserve paragraph boundaries in streamed replies

leapdragon:
## What Problem This Solves

When a tool-calling turn produces distinct assistant messages, public Gateway chat events and OpenAI-compatible HTTP streams join their text without a paragraph boundary. A second message starting with a list or table becomes part of the preceding paragraph in the combined text.

## Why This Change Was Made

The shared text merger now keeps inserted separator bytes separate from each item's text. Snapshots and deltas use the same boundary calculation, including an initially empty item. This fixes the reviewed split-newline case without retracting output already sent over HTTP. Display truncation also retires separator state so a later correction cannot recreate discarded padding.

## User Impact

Distinct assistant items keep their paragraph boundaries in public chat text, the embedded terminal backend, Chat Completions, and OpenResponses. Existing blank lines do not gain extra padding. Same-item updates, provisional replacement, legacy unkeyed streams, media handling, and terminal fallback retain their contracts.

The tested current WebChat already displayed separate paragraph, tool, and list blocks. That layout and transcript order remain correct. Its public combined text was still affected, including the final event.

## Evidence

- Baseline: `18f452584417a122747415fefc3089e296dc22cf`. Runtime/public proof: `adb13ba9a810357878b3829d50808c954b3f887e`. Current head `053b12bae1a4d1be0fcbacc16e1de18a34c498f3` adds only two test-file corrections; all production and fixture inputs are unchanged.
- A real isolated Gateway and bundled WebChat used a controlled Responses provider and an actual read tool. The same normal prompt produced distinct assistant items on both builds.
- Baseline public chat and both HTTP streams returned `First paragraph.- Second item\n- Third item`. Candidate live and final text is `First paragraph.\n\n- Second item\n- Third item`.
- All 12 candidate HTTP runs passed: list, table, trailing newlines, leading newlines, both edges, and split leading-newline provider deltas, through both endpoints. Each stream completed normally with consistent terminal text.
- The real producer normalizes edge whitespace. Separate official-SDK endpoint regressions cover raw delta-only events: all four new cases failed on the contributor patch and passed after repair.
- A display-limit correction also has a verified failing control and passing regression coverage. The original 713 focused Gateway, abort, and terminal tests passed. CI then identified omitted worker expectations and a shutdown fixture; both were corrected, and all 82 tests in those affected files passed. Formatting, syntax/boundary lint, source-size and assertion checks passed.
- An optional fresh-session browser probe hit an auxiliary title-request limitation in the controlled provider and was stopped. It is excluded from the successful proof claims; its failure is retained with the full evidence.

Complete private traffic, provider requests, inspected screenshots, command logs, and review history are retained. No claim is made that the baseline browser layout was broken or that its final event repaired the missing boundary.

### DIFF
--- a/src/gateway/agent-event-assistant-text.ts
+++ b/src/gateway/agent-event-assistant-text.ts
@@ -78,6 +78,16 @@ export function resolveAssistantTextInput(data: unknown): AssistantTextInput | u
   };
 }
 
+/** Paragraph separator owed between a retained item prefix and the next item's text. */
+function missingItemBoundarySeparator(previousText: string, nextText: string): string {
+  if (!previousText || !nextText) {
+    return "";
+  }
+  const trailing = previousText.slice(-2).match(/\n*$/u)?.[0].length ?? 0;
+  const leading = nextText.slice(0, 2).match(/^\n*/u)?.[0].length ?? 0;
+  return "\n".repeat(Math.max(0, 2 - trailing - leading));
+}
+
 /** Merge item snapshots without imposing a transport's display or wire limit. */
 export function mergeAssistantText(
   previous: AssistantTextSnapshot,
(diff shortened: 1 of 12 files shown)


obviyus:
Thanks @leapdragon! Merged in 0a202622ded657fe08f901b3d91c072c6fd68e0b.

- **Bug:** Combined streamed text joined distinct assistant messages without a paragraph boundary.
- **Fix:** The shared merger tracks inserted padding separately, so updates and display truncation preserve the boundary without retracting emitted text.
- **Proof:** A real Gateway, WebChat, and read tool exercised the public path. All 12 HTTP scenarios completed consistently, and browser layout and transcript order stayed correct.
- **Validation:** 795 focused tests passed. Final review found no actionable issue, and all required [CI checks](https://github.com/openclaw/openclaw/actions/runs/34428686923) passed. The final correction changed only tests, preserving the production code used for the live proof.

PR #140835, opened 2026-09-07
fix(ui): keep pre-final assistant text visible after a completed turn

leapdragon:
## What Problem This Solves

Fixes an issue where users chatting in the Control UI WebChat would watch a long assistant reply disappear once the turn completed, leaving only a short closing message visible, when the agent ran on an `openai-completions` transport and a tool call followed the reply text.

## Why This Change Was Made

`openai-completions` transports retag already-streamed answer text as commentary once a tool call follows it. The WebChat's completed-turn grouping then folded that text behind the "worked for" row together with the tool activity, so the substantive reply vanished from view.

The fix changes `chat-thread-grouping` to fold only tool groups and text-less assistant groups when a turn completes. Assistant groups that carry text stay visible regardless of how the transport phased them.

## User Impact

Long replies that precede a tool call remain visible after the turn completes. Tool activity still collapses as before.

## Evidence

New cases in `chat-thread-grouping.test.ts` and `chat-thread.test.ts` cover a text-bearing commentary group before a tool call, a text-less group, and the existing tool-only collapse.

Focused runs on this branch (rebased on current `main`):

```
node scripts/run-vitest.mjs ui/src/pages/chat/chat-thread-grouping.test.ts ui/src/pages/chat/chat-thread.test.ts
  Test Files  2 passed (2)   Tests  277 passed (277)
node scripts/run-oxlint.mjs <changed files>   # clean
```

Observed originally on the 2026.9.2 release with an `openai-completions` transport in the WebChat.

### DIFF
--- a/ui/src/pages/chat/chat-thread-grouping.test.ts
+++ b/ui/src/pages/chat/chat-thread-grouping.test.ts
@@ -188,7 +188,7 @@ describe("cached group content classification", () => {
     },
   );
 
-  it("keeps media visible and folds commentary after the same message changes in place", () => {
+  it("keeps media visible and folds text-less activity after the same message changes in place", () => {
     const content: Record<string, unknown>[] = [
       { type: "image", url: "https://example.com/diagram.png" },
     ];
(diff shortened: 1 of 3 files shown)


VACInc:
Related merge: #146554 preserves continuous live text through tool activity. It does not change completed-turn grouping or the rule that folds text retagged as commentary behind the worked-for row. This PR should remain open for its separate post-completion visibility scope.

PR #140839, opened 2026-09-07
perf(media): bound playback input staging buffers

### LINKED ISSUE #140838: Playback transcoding duplicates the complete source during input staging
Playback transcoding reads the complete admitted source into a Buffer and sends it through a private workspace write that copies the payload again. Those full buffers are temporary input staging rather than required conversion output.

Keep private-store admission, but copy the admitted source descriptor to a private staging descriptor with bounded scratch. Preserve short reads/partial writes, overflow detection, source identity, sync/close, guarded rename and verified reopen before ffmpeg. Output buffering and cleanup-before-cache-publication must remain unchanged.

The actual resolver/CAF conversion probe with real ffmpeg and ffprobe observes lower Node ArrayBuffer usage at a 15.36 MB input, with a small elapsed-time increase. This is bounded input staging; it should not claim HTTP/video coverage, fixed process peaks, faster conversion or changed cache semantics.

steipete:
Closes #140838

<details>

</details>

## What Problem This Solves

Playback input staging holds a complete source Buffer and an additional private-write payload copy before conversion. The resolver only needs a private input file for ffmpeg.

## Why This Change Was Made

Create an empty private staging leaf through the existing store, then copy between admitted descriptors with at most 64 KiB scratch and explicit positions. Preserve overflow/short-read handling, partial writes, source revalidation, sync, guarded rename and writer-identity/size/mode checks while the writer stays open. Both staging handles close before ffmpeg. Output buffering and cleanup-before-cache-publication stay unchanged.

## User Impact

Eight matched actual resolver runs with deterministic CAF input and real ffmpeg/ffprobe preserve source/output/duration/readiness hashes and private staging permissions. Near a 15.36 MB source, input scratch changes from 15,360,130 to 65,536 bytes and the full private-write payload copy is removed.

With tested fs-safe 0.8.1, observed Node ArrayBuffer rise changes 31.847→2.621 MiB; sampled ffmpeg RSS stays about 15.6 MiB. These sampled maxima are not exhaustive peaks or process-memory ceilings. The output Buffer remains required. Near-cap elapsed time increases 851.7→864.0 ms (+1.4%), and small/medium medians rise 2.6–5.6%; no speedup is claimed.

## Evidence

- On Node 26.8.1 and installed fs-safe 0.8.3, 104 focused tests, the full changed-file gate and independent P2 review pass. Cases cover positioned partial I/O, overflow/truncation/cap rejection, zero-progress/original failures, changed source identity, substituted staging input and cleanup failure blocking publication.
- Actual resolver conversion includes source inspection, staging, conversion, output probing, cleanup and cache publication. HTTP ticket/Range/HEAD flows and real video encoding were not replayed.
- A fresh fs-safe 0.8.3 resolver/ffprobe/ffmpeg smoke converts a synthetic 15,360,129-byte CAF to valid 160-second AAC/M4A. It verifies exact staged bytes, private modes, writer lifetime through the final identity check, closure before ffmpeg, unchanged source and cleanup before cache readiness. Original comparative measurements above remain labeled 0.8.1; this smoke is current-dependency integration proof.
- Production net+69 lines handles the explicit I/O/descriptor ownership formerly delegated to a whole-buffer write; tests net+258, including extracted staging cases. The old test-only reader export is removed. No dependency, configuration, cache-key/quota/TTL, concurrency or persistent-state semantics change.

The private input file and OS page-cache work remain. Full output ownership is preserved; no leak, fixed RSS or whole-Gateway claim.

## CI Follow-up

The original Linux run exposed two assumptions in the new staging tests. The same-length rewrite fixture now explicitly changes and verifies its modification timestamp. The staging writer now remains open through the guarded rename and reopened-file identity check, preventing inode reuse from defeating that comparison. The moved-substitution regression observes the actual production writer; it fails on the former close-before-move implementation and passes with the repair. The existing changed/mismatch rejection and cleanup assertions remain intact.

The production repair changes 19 lines each way and adds no payload buffers or filesystem calls. Fresh Linux CI on the repaired head is required before landing. Windows delete-sharing compatibility was checked against [Node 26.8.1 source](https://github.com/nodejs/node/blob/7be6d3af31a65adea57c94c41e50c2b071ed0b3a/deps/uv/src/win/fs.c#L480-L513); no live Windows execution is claimed.

### DIFF
--- a/src/media/playback-input.test.ts
+++ b/src/media/playback-input.test.ts
@@ -1,0 +1,101 @@
+import { describe, expect, it, vi } from "vitest";
+import { copyPlaybackInputBounded } from "./playback-input.js";
+
+function reader(contents: Buffer, maxRead = Number.POSITIVE_INFINITY) {
+  return {
+    read: vi.fn(async (buffer: Buffer, offset: number, length: number, position: number) => ({
+      bytesRead: contents.copy(buffer, offset, position, position + Math.min(length, maxRead)),
+    })),
+  };
+}
+
+describe("bounded playback input copy", () => {
+  it("preserves every byte across short reads and partial positioned writes", async () => {
+    const contents = Buffer.from(
+      "A short read must finish all its writes before reusing the scratch buffer. 🦞",
+    );
+    const source = reader(contents, 7);
+    const actual = Buffer.alloc(contents.length);
+    const target = {
+      write: vi.fn(async (buffer: Buffer, offset: number, length: number, position: number) => ({
+        bytesWritten: buffer.copy(actual, position, offset, offset + Math.min(length, 3)),
+      })),
+    };
+    await copyPlaybackInputBounded(source, target, contents.length, contents.length);
+    expect(actual).toEqual(contents);
+    expect(source.read.mock.calls.at(-1)?.[3]).toBe(contents.length);
+    expect(target.write.mock.calls.length).toBeGreaterThan(source.read.mock.calls.length);
+  });
+
+  it("copies a file larger than the scratch buffer without dropping a tail", async () => {
+    const contents = Buffer.from("0123456789".repeat(20_000) + "tail");
+    const source = reader(contents);
+    const actual = Buffer.alloc(contents.length);
+    await copyPlaybackInputBounded(
+      source,
+      {
+        write: async (buffer, offset, length, position) => ({
+          bytesWritten: buffer.copy(actual, position, offset, offset + length),
+        }),
+      },
+      contents.length,
+      contents.length,
+    );
+    expect(actual).toEqual(contents);
+    expect(source.read.mock.calls.length).toBeGreaterThan(2);
+  });
+
+  it.each([
+    { name: "growth", contents: "123456789", expected: 5, cap: 5, readBytes: 6 },
+    { name: "truncation", contents: "1234", expected: 5, cap: 5, readBytes: 4 },
+    { name: "byte cap", contents: "123456789", expected: 8, cap: 5, readBytes: 6 },
+  ])(
+    "rejects $name with only the bounded overflow probe",
+    async ({ contents, expected, cap, readBytes }) => {
+      const source = reader(Buffer.from(contents));
+      const target = {
+        write: async (_buffer: Buffer, _offset: number, length: number) => ({
+          bytesWritten: length,
+        }),
+      };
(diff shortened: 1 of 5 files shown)


steipete:
The fs-safe 0.8.3 source gap is covered by the published archive audit: its integrity matches the frozen lock, and the archive SHA-256 is `84cce0572e654374e744096132fbeed18b9b2745dabc5fd0f998e9c0241186d5`. These source links pin the commit behind v0.8.3:

- [The private writer still copies Buffer input](https://github.com/openclaw/fs-safe/blob/bc2ef952cfb54ca75340f933540cb51e6ab47ea9/src/secret-file.ts#L356); streaming into the admitted descriptor removes that copy.
- [Root.openWritable](https://github.com/openclaw/fs-safe/blob/bc2ef952cfb54ca75340f933540cb51e6ab47ea9/src/root-impl.ts#L493) preserves the existing file in update mode. [Root.move](https://github.com/openclaw/fs-safe/blob/bc2ef952cfb54ca75340f933540cb51e6ab47ea9/src/root-impl.ts#L1493) uses guarded rename, without a payload-copy fallback.
- [File identity compares device/inode](https://github.com/openclaw/fs-safe/blob/bc2ef952cfb54ca75340f933540cb51e6ab47ea9/src/file-identity.ts#L33), retaining its existing Windows unknown-value behavior. The repaired caller holds its writer until the reopened file passes the separate identity, size and mode checks; it adds no stronger timestamp/content guarantee to that helper.

Head `715b2ab60b79ce7ec3363025794bec138717bfaf` has current-0.8.3 real CAF resolver/ffprobe/ffmpeg proof, 104 focused tests and a passing [required Linux CI run](https://github.com/openclaw/openclaw/actions/runs/34090644634). The unrelated unchanged CLI exit-timeout job passed its one exact rerun. Original comparative allocation/timing results remain labeled 0.8.1; the new smoke establishes current-dependency behavior and writer lifetime.

steipete:
Merged via squash.

- Prepared head SHA: [715b2ab60b79ce7ec3363025794bec138717bfaf](https://github.com/openclaw/openclaw/pull/140839/commits/715b2ab60b79ce7ec3363025794bec138717bfaf)
- Landed commit: [149e97d4dbcadfaf2eda0c463aa7c7815da393d3](https://github.com/openclaw/openclaw/commit/149e97d4dbcadfaf2eda0c463aa7c7815da393d3)

PR #137440, opened 2026-09-03
fix(memory): avoid premature maintenance on cold gateways

shojikumaru:
## What Problem This Solves

Resolves a problem where a cold Gateway could flush memory or compact history too early: maintenance could use a 200k fallback even though the selected model was already prepared with a 1M context window.

## Why This Change Was Made

Carry provider-qualified prepared limits through the existing context resolver and reuse that budget for provider-floor planning. Remove the duplicate resolver while preserving configured caps, native windows, smaller cached limits and fallback behavior.

Production LOC: +25/−32 (net −7) | Tests: +136/−11.

## User Impact

Cold sessions plan maintenance against their selected model's budget. Explicit smaller caps still trigger normal maintenance, and optional flushing preserves the current private-session and post-delivery behavior.

## Evidence

- [Fresh lint, build and real Gateway proof](https://github.com/steipete/openclaw/actions/runs/34076206121): four cells, 20 settled maintenance cycles and 20 exact replies. Cold 177k/190k sessions use 1M budgets without unnecessary maintenance; the 100k-cap control completes one healthy detached flush while preserving source usage and history; the naturally warmed control remains correct.
- 261 owner/sibling tests and 20 changed checks passed on the equivalent candidate before the six-token local variable rename. The successful continuation revalidates those immutable results and the exact rename equivalence, then runs fresh lint, build and Gateway proof.
- Before evidence includes two intended unit failures and an actual cold 200k/176k decision at 177,032 tokens followed by maintenance dispatch. That baseline fixture stopped after dispatch; it did not complete a healthy full comparison.

The Gateway flow uses synthetic conversations and a loopback provider; no authenticated external-provider API or production Gateway update is claimed. [Published-head CI is green](https://github.com/openclaw/openclaw/actions/runs/34079096462).

Refs #127239. Thanks @shojikumaru.

PR #140841, opened 2026-09-07
fix(tests): avoid model discovery in memory budget fixtures

steipete:
## What Problem This Solves

Resolves a problem where memory-maintenance budget tests entered unrelated model-capability discovery, coupling otherwise synthetic CI cases to discovery startup.

## Why This Change Was Made

Declare the fixture model's text input in config while leaving context-window metadata under each table case's control. Prepared-window, authored-cap, provider-mismatch, and missing-catalog cases retain their original budget and flush assertions. This follows the capability-fixture boundary established in #133532 for the table added in #137440.

## User Impact

Test reliability only. Production behavior, assertions, and timeouts are unchanged.

## Evidence

- Observation-only `vi.spyOn` preserving the real discovery implementation: all original budget assertions passed on the baseline, but the three runnable cases each made one unrelated vision-catalog call. With configured text input, all five cases passed with zero discovery calls. The temporary observer was removed from this patch.
- Final uninstrumented memory and vision coverage: 172 tests passed across six files on Blacksmith Testbox, Node 24.19.0.
- Standard changed-file gate passed, including core test types, formatting, lint, and repository guards.
- Independent P2 review: no actionable findings.

The original 120-second CI timeout did not capture stage timing. The observation proves the unintended discovery path and its removal; it does not attribute that timeout to a specific discovery stage or constitute a cold-import benchmark.

### DIFF
--- a/src/auto-reply/reply/agent-runner-memory.test.ts
+++ b/src/auto-reply/reply/agent-runner-memory.test.ts
@@ -609,6 +609,27 @@ describe("runMemoryFlushIfNeeded", () => {
         : undefined,
     });
     const overrides = {
+      cfg: {
+        agents: { defaults: { compaction: { memoryFlush: {} } } },
+        models: {
+          providers: {
+            [provider]: {
+              baseUrl: "https://catalog-fixture.invalid/v1",
+              // Keep input preparation local while each case owns its catalog budget facts.
+              models: [
+                {
+                  id: model,
+                  name: model,
+                  reasoning: false,
+                  input: ["text"],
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [cf93a9b98c1e3606d62b88d787a5905d99b44ec1](https://github.com/openclaw/openclaw/pull/140841/commits/cf93a9b98c1e3606d62b88d787a5905d99b44ec1)
- Landed commit: [dbe88558c6141f17d49ed20ae91305029a80d572](https://github.com/openclaw/openclaw/commit/dbe88558c6141f17d49ed20ae91305029a80d572)

PR #140862, opened 2026-09-07
fix: yield after costly session list preparation

steipete:
## What Problem This Solves

Fixes an issue where loading a filtered Activity session list could keep other Gateway work waiting after expensive preparation, even when the resulting page contained only one row or no rows.

## Why This Change Was Made

Carry the existing 12 ms work budget from store loading through list preparation, then yield after synchronous selection and transcript capture when that budget is spent. This preserves the sharing facts' synchronous lifetime, workspace pin, transcript read order, and fresh selected-row authorization checks. The underlying store load and preparation remain atomic synchronous phases.

## User Impact

Other Gateway work gets an event-loop opportunity before row projection after costly preparation. Cheap lists retain their existing behavior. This is a cooperative fairness improvement; it does not promise a 12 ms maximum stall or reduce the underlying SQLite work.

## Evidence

- Final regression table: four expected failures on unchanged production code, two existing row cases passing, then all six cases passing with the repair. The loading-only case also proves that visibility filtering starts before the queued callback.
- Linux Node 26.7: the unchanged production diff passed 103 tests across eight focused suites covering the work budget, visibility refresh, response-cache invalidation, catalog scope, transcript ownership, plugin runtime, and single-row projections.
- Native canonical-writer proof kept retained shallow-copied session metadata stable across entry replacement and participant/owner updates. A separate real lifecycle listener/event-bus/index projection proof preserved late-bound run start timestamps.
- Ordinary production build and loopback Gateway RPC flow passed with synthetic sessions, ordinary metadata mutations before measured calls, four list shapes, 12 measured list requests, 80 concurrent successful health requests, and clean Gateway/seeder shutdown. Request latency remained similar to baseline; no overall speedup claim is made.
- Independent autoreview through P2: no actionable findings, including the final test-only extension. All 29 changed-file checks passed before that extension; [final exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34088725588) passed.

Production delta: +8 lines for carrying the existing work clock and the additional checkpoint; tests: +62 net lines extending the existing behavioral table.

### DIFF
--- a/src/gateway/server-methods/sessions-read.ts
+++ b/src/gateway/server-methods/sessions-read.ts
@@ -1,4 +1,5 @@
 // Read-only session queries.
+import { performance } from "node:perf_hooks";
 import { setImmediate as yieldToEventLoop } from "node:timers/promises";
 import { expectDefined } from "@openclaw/normalization-core";
 import { normalizeOptionalString } from "@openclaw/normalization-core/string-coerce";
@@ -247,6 +248,7 @@ export const sessionReadHandlers: GatewayRequestHandlers = {
             rowRepairAttempted?: boolean;
           } = {},
         ): Promise<Awaited<ReturnType<typeof listSessionsFromStoreAsync>>> {
+          const workStartedAt = performance.now();
           let loaded = options.loaded;
           if (!loaded) {
             const loadedStore = measureDiagnosticsTimelineSpanSync(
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6a0b8e627586cf209bc64c01fc87987c7ac3436d](https://github.com/openclaw/openclaw/pull/140862/commits/6a0b8e627586cf209bc64c01fc87987c7ac3436d)
- Landed commit: [d8b597be60c4b3c0d00958ecca807eeb2eac4c8c](https://github.com/openclaw/openclaw/commit/d8b597be60c4b3c0d00958ecca807eeb2eac4c8c)

PR #140863, opened 2026-09-07
fix(macos): align menu cost totals with local days

steipete:
## What Problem This Solves

Fixes an issue where Mac users would see another day's spending or **n/a** in the menu's **Today** cost total when the local date differed from UTC. Daily cost ranges also used 24-hour UTC days across local daylight-saving transitions.

## Why This Change Was Made

Request cost buckets in the Mac's local time zone using the existing Gateway API, including its UTC-offset fallback for older Gateway timezone data. Keep the requested date formatter with the cached response so Today and chart labels use the same day boundaries. This replaces the process-global formatter without adding configuration or changing the wire schema.

## User Impact

The menu's Today total and daily chart use local calendar days, including 23- and 25-hour daylight-saving days. Gateways that do not recognize a newer client timezone can still return costs through the existing fixed-offset fallback.

## Evidence

- Compiled native request-method and wire-model proof failed before the change and passes afterward. At `2026-09-07T01:00Z`, the Los Angeles display and server now agree on September 6; the Tokyo midnight case also matches.
- Eight Foundation tests covering 65 cases pass, including five local/UTC/DST cases and seven whole-, half-, and quarter-hour offsets. The new offset cases failed before the compatibility correction.
- The actual server date resolver accepts the native serialized parameters, preserves IANA timezone precedence, and handles the synthetic unsupported-zone fallback. This is isolated boundary proof; no live pair of mismatched timezone databases was used.
- Swift 6 cost view/model typechecking, pinned SwiftFormat and SwiftLint checks on all five changed Swift files, and P0–P2 autoreview pass. The [initial native CI run](https://github.com/openclaw/openclaw/actions/runs/34086866482) caught a non-Sendable JSON dictionary captured in the new test fixture. The correction constructs the existing Sendable request record before the lock closure; an exact pinned-dependency Swift 6 reproduction now compiles and preserves the wire fields. The [corrected exact-head CI run](https://github.com/openclaw/openclaw/actions/runs/34088766421) passed, including both changed suites, 2,082 default native tests, two isolation tests, and 1,674 shared-package tests.

The change affects date/data behavior without redesigning the menu's appearance. Production delta is +16 lines, with +47 test lines; the added ownership keeps cached data and its date interpretation together.

### DIFF
--- a/apps/macos/Sources/OpenClaw/CostUsageMenuView.swift
+++ b/apps/macos/Sources/OpenClaw/CostUsageMenuView.swift
@@ -3,6 +3,7 @@ import SwiftUI
 
 struct CostUsageHistoryMenuView: View {
     let summary: GatewayCostUsageSummary
+    let dates: CostUsageMenuDateParser
 
     var body: some View {
         VStack(alignment: .leading, spacing: 10) {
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [758b969e0203a8894bdc87d4645cb325c2297dc0](https://github.com/openclaw/openclaw/pull/140863/commits/758b969e0203a8894bdc87d4645cb325c2297dc0)
- Landed commit: [a2770e4592fe7127c12fe594edddd1e26cfca1c4](https://github.com/openclaw/openclaw/commit/a2770e4592fe7127c12fe594edddd1e26cfca1c4)

PR #128289, opened 2026-08-23
[codex] Repair Browser Session Credential Steward MVP

SnowBelt:
## What Problem This Solves

The Browser / Session / Credential Steward MVP needed one reliable authority path across Browser node routing, session targeting, approval provenance, and credential redaction. This PR prevents malformed or cross-agent session requests, keeps approved Browser node routes bound to the authorized node lifecycle, propagates Gateway run-authority revocation into in-flight paired-node execution, prevents host fallback after a failed or revoked node route, and keeps credential-like material out of policy decisions, tracked metadata, diagnostics, and browser output.

## Evidence

Behavior addressed: Browser-owned approval authority, Gateway-to-node run-authority revocation, node/pairing lifecycle revocation, session-boundary validation, credential redaction, response-body safety, snapshot safety, and guarded Browser delegation.

Real environment tested: Mac Studio disposable OpenClaw verification clone with the repository Node/pnpm toolchain and temporary local Browser/Gateway/node test harnesses. No credentials, production services, real external messages, or user data were used.

Exact steps or commands run after the latest rebase and fix:
- Focused Vitest proof for steward policy, delegated authority, Gateway, Browser, node-host, credential-redaction, meeting-runtime, and eval-harness tests.
- `pnpm tsgo:prod`
- `pnpm tsgo:extensions:test`
- `pnpm lint:core` and touched-file oxlint checks
- `pnpm docs:list`, `pnpm docs:check-mdx`, `pnpm docs:check-links`, and the repository glossary check
- `pnpm build`
- contracts-only agent-role eval
- TruffleHog scan of the branch diff
- fresh sanitized autoreview against current upstream main

Evidence after fix:
- After the verified rebase onto current upstream `main`, the focused steward regression suites passed, including Session Steward policy/boundary/gateway coverage, delegated authority and node-host coverage, Browser runtime/approval/redaction coverage, session-tab metadata coverage, meeting-runtime compatibility coverage, and the eval harness. The Browser suite was run as isolated serial files because the repository's cross-file mock state requires isolation.
- TypeScript, core lint, touched-file extension/script lint, documentation MDX/internal-link/config-example/format/markdownlint checks, build, contract catalog, secret scan, and fresh sanitized branch autoreview passed. The final rebased focused steward proof covered 357 core/Gateway tests and 689 Browser/plugin/node-host tests. The repository glossary check remains a pre-existing baseline failure with many unrelated missing zh-CN entries on current `main`; no glossary baseline files were changed.
[...]

PR #140882, opened 2026-09-07
perf(browser): count encoded envelopes without a second serialization

### LINKED ISSUE #140881: Browser node responses allocate a second string only to count encoded bytes
Browser node responses serialize their payload to JSON, then serialize that string again only to measure its size inside the node result frame. The second full encoded string is discarded after the 24 MiB admission check.

Count the exact second-level encoded bytes from native compact JSON output while keeping the original serialization and final transport framing. Preserve quote/backslash expansion, C0 rules, lone surrogate handling, native serialization errors, file limits and the exact boundary. The helper should remain private to this producer domain, rather than becoming a generic serializer.

Actual delegated-command and framing/receiver proof shows lower sampled allocation for large PDF/file envelopes with identical results. Owner retention remains baseline-equivalent. Related #103328 established the existing nested-JSON budget; this work must preserve it, not raise it.

steipete:
Closes #140881

<details>

</details>

## What Problem This Solves

Browser node responses build a second full JSON string solely to measure the serialized payload inside the outer node frame. That allocation is discarded after budget admission.

## Why This Change Was Made

Keep native serialization once, then count its exact quoted byte length with UTF-8 byteLength, statics-free quote/backslash scans and typed toUSVString handling for raw lone surrogates. The private helper accepts native compact JSON output; actual final framing and receiver validation stay unchanged.

## User Impact

Twenty actual delegated-command/framing/receiver flows preserve returned and wire hashes, native errors, file handling and exact 24 MiB admission. Sampled allocation falls by about 22.55 MB per owner operation for an 8 MiB PDF and 44.79 MB for 16 MiB aggregate files; corresponding framing/receiver observations are consistent.

Owner retention matches baseline. Malformed raw JSON can still require a replacement string, with no demonstrated allocation saving in that control. No small-output throughput, strict non-regression, peak/RSS or whole-Gateway claim is made.

## Evidence

- 63 focused tests, full changed-file checks and final P2 review pass. Three exact-limit cases cover minus one/equal/plus one with controls, Unicode, escapes and lone surrogates.
- 27 private counter oracle cases and 134 native serialization-domain checks preserve native JSON/rawJSON behavior. Native serialization still owns unknown-object traversal and its errors.
- Actual control service, routes, file read/MIME/base64, Gateway request framing and receiver validation/persistence are exercised. Playwright/CDP and RPC transport/ack boundaries are synthetic.
- The rejected sticky-RegExp prototype is excluded; no gain against that prototype is credited. A longer small-output control failed fixture validation and is excluded entirely, so it supplies no throughput or parity claim.
- Production net +34 lines; tests net zero. Per-file/aggregate/encoded limits, routes and authorization remain unchanged. No new public serializer, SDK/configuration or full build requirement.

Related #103328 established the budget. Active #128289 and #126870 change independent invocation authority/lifecycle behavior; preserve those guards if they land first.

### DIFF
--- a/extensions/browser/src/node-host/invoke-browser.test.ts
+++ b/extensions/browser/src/node-host/invoke-browser.test.ts
@@ -665,10 +665,7 @@ describe("runBrowserProxyCommand", () => {
 
       const error = await runBrowserProxyCommand(
         JSON.stringify({ method: "POST", path: "/act" }),
-      ).then(
-        () => null,
-        (err: unknown) => err,
-      );
+      ).catch((err: unknown) => err);
       expect(error).toBeInstanceOf(Error);
       expect((error as Error).message).toBe(
         `browser proxy file read failed for ${secondPath}: Error: browser proxy files exceed 16 MiB aggregate limit`,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [90e3cada354bef4ded98823b55ae9c37eeb38378](https://github.com/openclaw/openclaw/pull/140882/commits/90e3cada354bef4ded98823b55ae9c37eeb38378)
- Landed commit: [1a970dc06ac5cd06595055d7748c7d715df8a0f4](https://github.com/openclaw/openclaw/commit/1a970dc06ac5cd06595055d7748c7d715df8a0f4)

PR #140888, opened 2026-09-07
refactor(tests): simplify the serial UI source inventory

steipete:
Related: #139428

## What Problem This Solves

The Web Awesome ownership guard repeatedly scans local source files to enforce UI architecture contracts. Its serial checks pay asynchronous filesystem and Promise overhead without needing concurrent execution.

## Why This Change Was Made

Uses synchronous filesystem reads and a recursive flatMap walk, removing unnecessary async/await machinery. All three tests, five scans, regular expressions, file filters, sorted paths, and independent expected inventories remain unchanged. There is no cache or retained source snapshot.

## User Impact

No product behavior change. This maintainer-requested cleanup removes four net test lines and reduces local source-scan overhead; it does not claim a full-suite runtime improvement.

## Evidence

- All three actual Vitest cases passed before and after with identical names and results. Single local shard runs took 1.58 seconds before and 1.06 seconds after; these are observations, not a controlled full-suite benchmark.
- An alternating source-only diagnostic using the actual case bodies and assertions measured median scan work of 222 ms versus 133 ms. Both variants performed all 7,940 file reads and returned identical match arrays. This excludes Vitest startup and does not measure global test performance.
- File and directory failure probes preserved filesystem errors at all three case boundaries.
- Full mapped changed gate, deslop, and independent P2 review passed.

### DIFF
--- a/ui/src/components/web-awesome-migration.node.test.ts
+++ b/ui/src/components/web-awesome-migration.node.test.ts
@@ -1,31 +1,27 @@
 // @vitest-environment node
-import { readdir, readFile } from "node:fs/promises";
+import { readdirSync, readFileSync } from "node:fs";
 import path from "node:path";
 import { describe, expect, it } from "vitest";
 
 const sourceRoot = path.resolve(import.meta.dirname, "..");
 
-async function productionTypeScriptFiles(dir = sourceRoot): Promise<string[]> {
-  const entries = await readdir(dir, { withFileTypes: true });
-  const files = await Promise.all(
-    entries.map(async (entry) => {
-      const filePath = path.join(dir, entry.name);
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [674bbbd0b510cf5a47a44a65ae7f30a08f8cf1d9](https://github.com/openclaw/openclaw/pull/140888/commits/674bbbd0b510cf5a47a44a65ae7f30a08f8cf1d9)
- Landed commit: [7ab1c97cb7fee1a5bc72da1cf913ea25873c3497](https://github.com/openclaw/openclaw/commit/7ab1c97cb7fee1a5bc72da1cf913ea25873c3497)

PR #140893, opened 2026-09-07
fix(ui): preserve newer copies and restore browser URL copying

steipete:
## What Problem This Solves

Fixes an issue where users copying code, table data, or Mermaid source could end up with older text after the source disappeared or a newer copy action started. Browser-tab **Copy URL** also failed on plain-HTTP dashboards where the browser does not expose its Clipboard API.

## Why This Change Was Made

All known UI copy actions now record shared copy intent, and the existing text fallback checks that the intent and source are still current. Browser-tab URLs use that shared fallback instead of a separate direct write. Managed-image and widget actions only register intent; their existing `ClipboardItem` promises and immediate binary writes are preserved.

## User Impact

A delayed text fallback no longer overwrites a newer text or image copy request. Removed or replaced code, table, Mermaid, and browser-card sources cannot start a stale fallback, and **Copy URL** works on plain HTTP. Native clipboard writes already accepted by the browser are not canceled.

## Evidence

- 78 focused tests passed, including 21 real-Chromium cases and existing clipboard, code-block, table, Mermaid, browser-card, and widget owner tests. Regressions were reproduced before their repairs, including empty copy requests and binary copy intents with the API available or unavailable.
- A genuine insecure-origin Chromium capture reported `isSecureContext=false` and no Clipboard API. Copy URL threw before the fix and copied the exact URL without errors afterward. Clipboard transports were captured into synthetic page storage; no real clipboard, account, or Gateway was used.
- UI production and owning test type graphs, full-file typed lint, scoped stylelint, formatting, and repository ratchets passed. Fresh independent review through P2 found no actionable issues. The refresh onto current main preserves the reviewed source bytes/modes and the unchanged documentation patch.
- Initial Linux UI CI exposed three parser-copy fixtures that clicked detached fragments. The fixture now mounts its existing fragment only during the interaction and removes it in `finally`; the retirement guard and assertions are unchanged. All 149 selected Markdown and clipboard lifecycle tests, owning test types, typed lint, and fresh P2 review pass after that correction.
- The browser tests prebundle four existing dependencies to avoid a reproduced cold optimizer reload and duplicate custom-element registration.

| Scenario | Before | After |
| --- | --- | --- |
| Delayed text copy | An older fallback overwrites the newer copied message | The newer copied message survives with no stale fallback |
| Browser-tab Copy URL over plain HTTP | Copy URL fails when the browser Clipboard API is unavailable | Copy URL succeeds through the existing fallback |

### DIFF
--- a/docs/web/control-ui/chat.md
+++ b/docs/web/control-ui/chat.md
@@ -156,6 +156,14 @@ Chat error banners, including cloud runner failures, show short messages in full
 Decoded text artifacts use the same literal preview. **Copy code** preserves the
 code's leading whitespace and final newline when present.
 
+**Copy URL** in browser tab cards also works on plain HTTP connections where the
+browser does not provide its Clipboard API.
+
+If the browser rejects a text clipboard write, starting another text or image
+copy cancels its delayed fallback. Code blocks replaced during streaming and
+Mermaid diagrams whose source changes also cancel that fallback. This does not
+cancel native clipboard writes that the browser has already accepted.
+
 ### Markdown tables
 
 Markdown tables scroll horizontally within the conversation. **Copy table** copies
--- a/ui/src/components/markdown-clipboard.browser.test.ts
+++ b/ui/src/components/markdown-clipboard.browser.test.ts
@@ -1,0 +1,353 @@
+import { html, nothing, render } from "lit";
+import { unsafeHTML } from "lit/directives/unsafe-html.js";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { createDeferred } from "../../../test/helpers/promise.js";
+import { renderMessageImages } from "../pages/chat/components/chat-message-images.ts";
+import { renderMessageMarkdown } from "../pages/chat/components/chat-message-text.ts";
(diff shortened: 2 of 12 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b60d912651c4415b7456f5cd693da801b3eb924d](https://github.com/openclaw/openclaw/pull/140893/commits/b60d912651c4415b7456f5cd693da801b3eb924d)
- Landed commit: [1b0186412e20ff7e248b3228fc8d4c2f7a287e5e](https://github.com/openclaw/openclaw/commit/1b0186412e20ff7e248b3228fc8d4c2f7a287e5e)

VACInc:
Related follow-up: https://github.com/openclaw/openclaw/pull/141574 merged into main on September 7, 2026 and fixes code-copy payloads being removed by HTML sanitization before the clipboard operation begins.

The shared copy-intent tracking, retired-source protection, and HTTP fallback behavior implemented here remain unchanged. These are complementary fixes: #141574 repairs the text supplied to the copy operation, not clipboard ordering or transport availability.

PR #140901, opened 2026-09-07
refactor: simplify group reply guidance

obviyus:
## What Problem This Solves

Group prompts repeated when-to-reply and silence instructions on every turn.

## Why This Change Was Made

The prompt keeps one participation rule and one complete automatic-silence instruction, including reaction-only outcomes. Message-tool-only guidance states private final delivery once and keeps the rule to skip sending when no visible response is needed. Runtime conditions, the exact silence token, routing exceptions, formatting, activation, and direct-chat wording remain unchanged.

## User Impact

Group prompts carry less repeated text with the same reply and silence rules. The two measured automatic-group prompts shrink by 23% and 22%. Direct-chat and ineligible-silence variants remain byte-identical.

## Evidence

All 20 targeted group, composition, and channel-prompt tests passed. Seven generated snapshots passed, with only the group snapshot changed. Semantic review compared nine complete before/after variants and found no lost instruction. Formatting, syntax lint, and whitespace checks passed. No live-channel behavior improvement is claimed.

### DIFF
--- a/src/auto-reply/reply.triggers.group-intro-prompts.cases.ts
+++ b/src/auto-reply/reply.triggers.group-intro-prompts.cases.ts
@@ -51,9 +51,7 @@ export function registerGroupIntroPromptCases(): void {
     const telegramGroupParticipationNote =
       "Be a good group participant: mostly lurk and follow the conversation; reply only when directly addressed or you can add clear value. Emoji reactions are welcome when available. Write like a human. Minimize empty lines and use normal chat conventions, not document-style spacing. Don't type literal \\n sequences; use real line breaks sparingly.";
     const groupSilentNote =
-      'If no response is needed, reply with exactly "NO_REPLY" (and nothing else) so OpenClaw stays silent.';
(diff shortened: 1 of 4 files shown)


obviyus:
Merged [a8d63e54ec6b](https://github.com/openclaw/openclaw/commit/a8d63e54ec6b461546c8f8514c4007bc358e91d1) as [2b6f611022d4](https://github.com/openclaw/openclaw/commit/2b6f611022d4218afda9449a4c42bc60ce364cff).

**Proof:** 20 targeted tests, seven prompt snapshots, independent semantic review, clean exact-head local review, and required CI passed. This is a wording refactor with no intended behavior change.

| Before | After |
| --- | --- |
| Repeated participation, silence, and private-final guidance | One complete rule per delivery mode; exact silent tokens and reaction-only outcomes preserved |

The automatic group-guidance block is 22–23% shorter. Direct-chat and disabled-silence variants are unchanged.

PR #140906, opened 2026-09-07
refactor(tests): use restorable status reply spies

steipete:
## What Problem This Solves

Three status reply tests reload the command module graph just to substitute rendering results and capture diagnostic logs. That adds unnecessary setup and makes the tested error-handling boundary harder to see.

## Why This Change Was Made

Use restorable namespace spies with the existing command imports. The same generic error replies, diagnostic details, rich presentation payload, and whole-case mock behavior remain covered. The separate test of lazy import failure and retry through the same module instance stays unchanged. Removes 23 net test lines and three command reevaluations without production changes.

## User Impact

No user-visible behavior change. The tests retain their error-containment and presentation guarantees with narrower setup.

## Evidence

- All 75 tests passed before, after, and on final formatted source: 55 command cases and 20 status-text cases, including ordinary status calls after spy restoration and the real lazy-loader retry.
- External comparison confirms all 149 assertion expressions, both command-call input objects, surrounding code, and reported case names/order remain unchanged. The status-text sibling is byte-identical.
- Full mapped changed gate, diff-scoped deslop, and fresh P2 review passed. The first gate caught new-import ordering; normal formatting corrected it.
- In one local comparison, the three selected cases totaled 303ms before and 3ms after; the two-shard wrapper took 34.90s and 21.96s. Transform/cache conditions differed, so the wrapper difference is not attributable savings or an aggregate runtime claim. The final formatted-source run took 23.81s.

### DIFF
--- a/src/auto-reply/reply/commands-status.test.ts
+++ b/src/auto-reply/reply/commands-status.test.ts
@@ -19,7 +19,9 @@ import {
   replaceSessionEntry,
 } from "../../config/sessions/session-accessor.js";
 import type { ModelDefinitionConfig } from "../../config/types.models.js";
+import * as logger from "../../logger.js";
 import type { ProviderThinkingProfile } from "../../plugins/provider-thinking.types.js";
+import * as statusText from "../../status/status-text.js";
 import {
   completeTaskRunByRunIdCore,
   createQueuedTaskRunCore,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1db980050b198757ddcd96542b812d355b42fdfe](https://github.com/openclaw/openclaw/pull/140906/commits/1db980050b198757ddcd96542b812d355b42fdfe)
- Landed commit: [c78e6aea330f58982252c15348341d34645a0ed5](https://github.com/openclaw/openclaw/commit/c78e6aea330f58982252c15348341d34645a0ed5)

PR #140924, opened 2026-09-07
fix(ui): preserve rejected enum state and nullable choices

steipete:
<details>

This is a same-repository branch; repository maintainers can update it.

</details>

## What Problem This Solves

Fixes an issue where users choosing a rejected value in a Settings enum control would leave its accessibility and keyboard state on the unaccepted value while the painted highlight still showed the previous choice. For example, after a rejected duplicate, beta remained highlighted but alpha was announced as checked, and ArrowRight only returned the internal selection to beta.

Also fixes small nullable choice fields that omitted their explicit `null` option, and the Appearance message-width input being named only by its `48rem` placeholder in the accessibility tree.

## Why This Change Was Made

The shared segmented control now restores its selected value when the existing parent validation rejects a patch. Small nullable enums use the existing dropdown, preserving explicit null separately from an unset optional setting. The width input uses its existing visible label as its accessible name.

Validation remains with the existing array/object owners. No config schemas, runtime defaults, or persistence behavior change. The four production files grow by three net lines; the added browser scenarios exercise the full Settings form.

## User Impact

Rejected choices keep the saved value, accessibility state, and keyboard navigation aligned. After a rejected duplicate, ArrowRight now advances from beta to gamma and saves once. Operators can select explicit null in small choice fields and find the message-width textbox by its visible name.

## Evidence

All three regressions were reproduced on the original source and pass with the repair. The production-bundle Chromium suite covers pointer and keyboard rejection, a subsequent accepted config write, required/optional/literal-union/large nullable enums, and message-width naming with persistence across reload. The 92 existing scalar, rejection, and Settings browser tests, full owning UI E2E test type graph, full-file typed lint, and formatting pass. Fresh independent P2 review found no actionable issues. The current-main integration preserves the reviewed patch and the newer Settings docs. CI also caught the expected assertion-count decrease from reusing the event target; the baseline now records six assertions instead of seven, and its owning ratchet passes.

A separate bounded Chromium continuation probe captured checked state, `aria-checked`, tab stops, focus, and config writes after the rejected click. Both rejected-click pictures paint beta; the meaningful visual difference appears on the **next ArrowRight**:

| Before: ArrowRight returns to beta, no write | After: ArrowRight selects gamma and saves once |
| --- | --- |
| Before: next ArrowRight leaves beta selected | After: next ArrowRight advances to gamma |

The nullable fields now expose their explicit null choice:

| Before: small enums omit null | After: null is selectable |
| --- | --- |
| Before: small nullable enum controls have no null option | After: nullable enum dropdowns show null |

The actual accessibility snapshot for message width changed from `textbox "48rem"` to `textbox "Message width"`, with `48rem` retained separately as its placeholder. Screenshots and fixtures are synthetic; no operator Gateway or real account state was used.


steipete:
Merged via squash.

- Prepared head SHA: [377301bdd966a2a4be4592f42f84fdbe6a69732c](https://github.com/openclaw/openclaw/pull/140924/commits/377301bdd966a2a4be4592f42f84fdbe6a69732c)
- Landed commit: [a3cab57127e0ba5f723709856e1a0b4e15effdab](https://github.com/openclaw/openclaw/commit/a3cab57127e0ba5f723709856e1a0b4e15effdab)

PR #140870, opened 2026-09-07
refactor(tests): import benchmark helpers from their owners

steipete:
## What Problem This Solves

Gateway benchmark tests still reach shared helpers through test-only forwarding members on the startup and restart CLI modules. This leaves duplicate ownership seams after the shared benchmark runtime extraction.

## Why This Change Was Made

Remove eleven forwarded members and one unused import. The same tests import the existing runtime, probe, and child-process helpers directly. CLI-local helpers stay on their existing testing surfaces, and both CLI modules remain imported and exercised through their real help/error paths.

## User Impact

No benchmark or Gateway behavior changes. Tooling loses 12 lines; test imports replace the forwarded access without removing cases or assertions. The child-stop suites retain their distinct exit-code scenarios and lifecycle checks.

## Evidence

- The startup, restart, and discovery-isolation suites pass all 67 cases before and after, with identical case names and outcomes.
- Trace buffering, probe transitions, malformed RSS, restart persistence, and bounded child cleanup assertions remain unchanged.
- Targeted formatting and `git diff --check` pass.
- Independent Codex review is clean through P2.
- All 19 normal changed-file checks passed on the configured Testbox (typechecks, lint, formatting, and structural guards); the one-shot lease and source capsule are closed.

No live Gateway benchmark or performance result is claimed: helper implementations and benchmark behavior are unchanged. A separate formatting follow-up tracks counts displayed with an ms suffix.

PR #140936, opened 2026-09-07
fix(bench): report startup counts without time units

steipete:
Related: #140870

## What Problem This Solves

Gateway startup benchmark reports label agent, workspace, and catalog counts as milliseconds. For example, a median of 11.5 agents appears as `11.5ms`, even though the numeric JSON is correct.

## Why This Change Was Made

Let the existing statistics formatter accept a value formatter while keeping milliseconds as its default. Count fields use unitless numbers, preserving fractional medians and averages. Memory and ratio summaries reuse that same owner, removing two duplicated formatting wrappers (net 11 fewer tooling lines).

## User Impact

All fourteen count fields display unitless values. Duration, memory, ratio, unavailable-value formatting, and numeric JSON retain their existing meaning.

## Evidence

- Four new regression cases fail on the original formatters; the repaired candidate passes 74 startup, restart, and shared-runtime cases.
- Actual CLI runs with synthetic HEAD-probe children verify all fourteen count fields and the same numeric trace summaries. Noncount fields in the prepared-runtime row and fixed memory rows match the baseline.
- Cancelling the test-owned CLI proves its detached fixture closes on stdin EOF without signalling retired PIDs; normal and cancellation paths release owned processes, ports, and temporary roots.
- Independent Codex review is clean through P2. The CLI fixture is POSIX-only because the managed helper does not support strict process-tree verification on Windows; no Windows execution or performance result is claimed.

All 19 normal changed-file checks passed on the final source. The first run stopped on three mechanical test-lint findings after 18 passing checks; those were corrected, checked locally, and the full gate rerun passed. Both one-shot Testboxes and source capsules are closed.

### DIFF
--- a/scripts/bench-gateway-startup.ts
+++ b/scripts/bench-gateway-startup.ts
@@ -569,20 +569,6 @@ function formatRatio(value: number | null): string {
   return value.toFixed(3);
 }
 
-function formatMemoryStats(stats: SummaryStats | null | undefined): string {
-  if (!stats) {
-    return "n/a";
-  }
-  return `p50=${formatMb(stats.p50)} avg=${formatMb(stats.avg)} min=${formatMb(stats.min)} max=${formatMb(stats.max)}`;
-}
-
-function formatRatioStats(stats: SummaryStats | null): string {
-  if (!stats) {
-    return "n/a";
-  }
-  return `p50=${formatRatio(stats.p50)} avg=${formatRatio(stats.avg)} min=${formatRatio(stats.min)} max=${formatRatio(stats.max)}`;
-}
-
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1a6a430af6f08f133c7ff353eb697659c5fc74d4](https://github.com/openclaw/openclaw/pull/140936/commits/1a6a430af6f08f133c7ff353eb697659c5fc74d4)
- Landed commit: [47042ffd20bedc35efc88f8e437d23eaf8df6fc7](https://github.com/openclaw/openclaw/commit/47042ffd20bedc35efc88f8e437d23eaf8df6fc7)

PR #128078, opened 2026-08-23
fix(memory): isolate sqlite-vec KNN from the event loop

anyech:
## What Problem This Solves

Fixes an issue where builtin memory search can freeze the Gateway while a cold sqlite-vec query runs. The native KNN/count/widening sequence uses synchronous SQLite, so the same event loop cannot service the search deadline until it returns. Related: #81172; #83758 repaired the separate JavaScript fallback, not native KNN.

Review also reproduced a reader-lifetime regression exposed by asynchronous KNN: forced reindex temporarily made its shadow database the manager's global database. A query could receive real child results and then lose them when provenance enrichment touched the now-closed shadow connection.

## Why This Change Was Made

The native query runs in a bounded, one-shot read-only child. The Gateway retains its event loop and can terminate the owned child on cancellation. There are at most two admitted children; only the authoritative close event releases admission, including when the caller's cleanup timeout expires. The child launches no descendants. Detached process groups, graceful tree termination, recycled-PID probes, and production-only test hooks were removed.

Published database state now has one stable owner. An owner-scoped async context selects the shadow database only inside the awaited rebuild; public operations leave that context. Closed shadow work fails rather than redirecting writes into the published database. Reindex publishes atomically under the existing workspace lock and invalidates the published metadata/vector/FTS state. No schema, schema version, config, migration, or provider API changes are introduced. The maintainer explicitly accepted this concurrency design.

The child reuses the runtime-worker resolver and SQLite opener, uses no shell, excludes parent credentials/Gateway variables, bounds its input/output and result rows, and never retries native KNN on the Gateway thread. Non-abort failures are logged and leave an empty vector stage so existing keyword/hybrid retrieval can continue.

## User Impact

Search cancellation can run while native SQLite is busy, and concurrent reindexing no longer exposes temporary shadow connections to public search/status operations. Existing source filtering, provenance, hybrid search, keyword/path search, and JavaScript fallback remain in place. One-shot process startup adds latency; this is cancellation isolation, not a claim that cold native queries become faster.

## Evidence

[...]

PR #140939, opened 2026-09-07
perf(memory): slim memory-search-knn child imports to cut ~2.3s per-query boot

gaoanze888:
## What Problem This Solves

Since 2026.8.1 (#128078), every `memory.search` query spawns a fresh `memory-search-knn.child.js` process to isolate synchronous sqlite-vec KNN work from the parent event loop. The child imported through heavy plugin-sdk barrels (`memory-core-host-engine-schema`, `sqlite-runtime`, `memory-core-host-engine-foundation`) whose `export ... from` statements eagerly load **1,000+ transitive `.js` files** into the child's module graph — adding **~2.3s of pure module resolution and evaluation** per query, while the KNN query itself takes ~70ms.

Closes #140681.

## Root Cause

In ESM, `export ... from` barrel re-exports cause ALL re-exported modules to load eagerly (not lazily). The child only needs 3 symbols:
- `loadSqliteVecExtension` (from `memory-core-host-engine-schema` → pulls in memory-schema, agent scope, config, paths)
- `openNodeSqliteDatabase` (from `sqlite-runtime` → pulls in agent DB, maintenance modules)
- `truncateUtf16Safe` (from `memory-core-host-engine-foundation` → pulls in agent scope, config, paths)

But importing these 3 symbols through the heavy barrels loads **8,501 module load events** at runtime.

## Fix

Replace the three heavy barrel imports with **narrow re-export barrels** that pull in only what the child needs:

| New barrel | Re-exports | Replaces |
|---|---|---|
| `src/plugin-sdk/sqlite-vec-runtime.ts` | `loadSqliteVecExtension` only | `memory-core-host-engine-schema` |
| `src/plugin-sdk/node-sqlite-runtime.ts` | `openNodeSqliteDatabase` only | `sqlite-runtime` |
| `src/plugin-sdk/utf16-slice-runtime.ts` | `truncateUtf16Safe` only | `memory-core-host-engine-foundation` |

Each new barrel is a single-line re-export from the direct source module, bypassing the heavy barrel's transitive dependency chain. The new barrels are registered in `package.json` exports and `plugin-sdk-entrypoints.json`.

### Performance measurement (from source via tsx, same machine, 3 runs each)

| Metric | Before (heavy barrels) | After (slim barrels) | Reduction |
|---|---|---|---|
| Cold start | 2.89s | 0.58s | **80%** |
| Warm start | 1.35s | 0.34s | **75%** |
| Module load events | 8,501 | 5,333 | **37%** |

The dist build is expected to show an even larger improvement because the eliminated barrels pull in agent-scope, config, paths, schema, and agent-database modules that are absent from the narrow barrels.

### Test verification

All existing memory-core KNN tests pass (33/33), including the subprocess boundary test that spawns the **real child process** and runs a KNN query:

```
✓ memory vector KNN subprocess boundary > queries the real source child across WAL writer visibility and source filters 384ms
✓ memory vector KNN subprocess boundary > keeps the parent event loop responsive during synchronous child work 287ms
✓ memory vector KNN subprocess boundary > hard-kills and reaps a busy child on caller abort, then admits another query 83ms
```

### Lint & format

- `oxlint` passes on all 5 modified/new files
- `oxfmt` clean
- `check-plugin-sdk-subpath-exports.mts` passes: "all referenced openclaw/plugin-sdk/<subpath> imports are exported"

## Why not a worker pool?

A persistent child worker pool (option A) would be more effective for repeated queries, but it adds significant complexity (lifecycle management, crash recovery, admission slot semantics) and changes the hard-cancellation contract that the subprocess approach provides. This PR takes the **minimal-risk approach** of trimming the import chain — the child still spawns per query (preserving the SIGKILL cancellation guarantee), but it boots 80% faster.


gaoanze888:
Replacing with v8b: fixed manager-search-knn.ts import (was missing from diff due to wrong base import path).

PR #140941, opened 2026-09-07
fix(cron): preserve prepared runtime and error ownership

### LINKED ISSUE #140940: [Bug]: isolated cron runs fail with "plugin generation was superseded" when a fallback provider's plugin is disabled (2026.9.x regression)
## Bug type

## Summary

Every isolated cron run that selects a provider outside the agent's primary route fails within ~150 ms with

```
prepared model runtime plugin generation was superseded for <state>/agents/main/agent
```

when `agents.defaults.model.fallbacks` (or any configured model chain entry) points at a provider whose owner plugin is **disabled**. Interactive turns and cron jobs on the primary provider keep working, which hides the cause.

This looks like #133692 but is a different repair boundary: #133693 (applied locally on 2026.9.2, its regression tests pass) does **not** fix it, because the failure is not the timing of the published-runtime read; it is the borrow predicate `preparedPluginGenerationSupportsSelections` rejecting a plan plugin recorded as `disabled`.

## Steps to reproduce

1. Primary model on one provider (here `nonstop/claude-opus-5`, anthropic-messages), `agents.defaults.model.fallbacks = ["opencode-go/glm-5.3", "<alias of bailian-token-plan/qwen3.7-max>"]`.
2. Leave the bundled `qwen` plugin (owner of provider `bailian-token-plan`) disabled; define `bailian-token-plan` and `opencode-go` as manual `models.providers` entries.
3. Create an isolated cron job (`agentTurn`, no delivery) with `model: "opencode-go/glm-5.3"` (or `openai/gpt-5.4-mini`).
4. `openclaw cron run <jobId>`.

## Expected behavior

The run is admitted and dispatched to the selected provider (or fails on that provider with a provider error), as on 2026.8.1.

## Actual behavior

Fails in lease admission before any provider call. With the lease's throw instrumented to print its predicates (local build), the failing run reports:

```
needsRefresh=false ownerGen=g1 publishedGen=g3 borrowed=true
metaMatch=true configMatch=true agentMatch=true flags=ro:false,lrp:false,sc:false,env:false,subBind:true
selectionsSupported=false
plan=["lossless-claw","memory-core","opencode-go","qwen"]
registry=[... "qwen:disabled", "opencode-go:loaded", "anthropic:loaded" ...]
selections=["bailian-token-plan/qwen3.7-max@auto","opencode-go/glm-5.3@auto"]
```

Timeline in the same process: boot publishes owner `main=g1` and dispatch publication `main=g1` (consistent); cron's first admission publishes a run-provenance owner `g3`; the nested admission (lane `cron-nested`) compares `g1 !== g3`, borrows, and `preparedPluginGenerationSupportsSelections` returns false only because `qwen` is `disabled` (it tolerates `status === "error"`, not `"disabled"`). The primary admission had already accepted that same disabled outcome when it built `g3`.

Jobs whose selections need only plugins in the base generation (anthropic family here) pass through the same borrow and succeed; jobs needing `openai`/`opencode-go` + the disabled `qwen` fail. Gateway restart does not help. No config hot reload is involved (unlike #128515).

## Regression range

- 2026.8.1 (`ea806575e64`): job succeeded daily through 2026-09-04 with the same config (same fallbacks, same disabled plugin).
- 2026.9.1 / main `bf9cbcfb064` / 2026.9.2 (`3928bad9bad`): fails every run from 2026-09-06 02:00 onward.
- The borrow predicate itself predates 2026.8.1 (#132804); what routes cron's nested admission into the borrow path on 2026.9.x was not bisected.

## Environment

- OpenClaw 2026.9.2 (`3928bad9bad`), git-mode install, macOS (Darwin 25.6, arm64), Node 26.5.
- Also reproduced on main `bf9cbcfb064` and 2026.9.1.

## Proposed fix

Treat `disabled` like `error` in `preparedPluginGenerationSupportsSelections` (one line + unit test). Verified live: after the change the same cron run is admitted and reaches the provider. PR to follow.

AI-assisted (Claude Code + Codex review), human-supervised session.

Marvinthebored:
Isolated cron runs can reject a usable non-primary model before dispatch when another plugin in the fallback plan is disabled. The prepared-generation borrow check treated that recorded outcome as a missing owner. Accept disabled outcomes alongside recorded load errors, while preserving the live parent lease, identity checks, and downstream model availability.

The provider-error control also exposed a missing handoff in cron's shared execution path: each failed candidate settled a separate error recorder against the same terminal transcript key. Pass the existing logical-run error recorder into the embedded runner so exhausted fallbacks report the provider failures once.

Validation:

- Current main `252158295ee487c4145404367fa15e3db588989b`: the real Gateway `cron run` path fails with generation supersession in 168 ms before any model POST.
- Candidate: the same configuration and original job return the normal provider-B result with exactly one B model POST and no C model POST.
- Disabled plugin-only model selection still returns `model_not_found` without model I/O. Explicitly configured generic provider transports retain their existing availability; disabling plugin code does not remove that separate configuration.
- Injected B and C provider failures now report both original errors instead of a transcript idempotency conflict. Restoring B succeeds in the same Gateway process, without a restart or an intervening primary run.
- Primary cron, interactive turns, first usable fallback, and disabled-plugin state pass. Earlier independent source-blind CLI acceptance remains recorded separately from owner-level checks.
- The new borrowed-lease cases cover loaded, disabled, error, absent, and not-imported outcomes. Reverting only the predicate makes the disabled case fail; the repaired owner/lifecycle set passes 99 tests. Cron fallback and shared-entry siblings pass 40 tests.
- Fresh independent P0–P2 review is clean. Hosted CI checks the published head.

The current native proof used candidate tree `6a051fda57b16deecc389828b52ba9605cb68e38`, now committed as `55a440123a0850b48f14213e38129cb74e27c34e`. Its pre-commit build stamp retains the original checkout base; no artifact was restamped. Earlier baseline/candidate proof and failed CI attempts are preserved with their original identities.

Fixes #140940.


obviyus:
Thanks @Marvinthebored! Merged as `855c9de4a2a7c8015d4fa78b28d00f330c0649ba`.

**Proof:** The real Gateway cron path now reaches the usable non-primary provider while the fallback plugin remains disabled. Exhausted fallbacks report the original provider errors, and restoring the provider succeeds without restarting the Gateway.

- **Before:** Generation supersession blocked dispatch; multiple provider failures could instead produce a transcript idempotency conflict.
- **After:** Borrowing preserves recorded disabled outcomes, and fallback attempts share the logical run's error recorder. Plugin-only unavailable models still reject normally.
- **Validation:** 139 focused owner/sibling tests passed; the disabled regression fails with the original predicate. Independent review and the final ClawSweeper gate passed. [Full CI passed on the published head](https://github.com/openclaw/openclaw/actions/runs/34185932100).

PR #140945, opened 2026-09-07
fix(workers): keep CLI preflight failures out of lease cleanup

steipete:
## What Problem This Solves

A missing or unusable Crabbox CLI can leave a cloud session stuck behind cleanup for a lease that was never requested. Core recorded allocation as possible before the plugin ran its local CLI/configuration checks; a subsequent Stop correctly refused to treat the broker's missing record as proof of deletion.

## Why This Change Was Made

Providers can prepare an allocation function while a fresh environment remains requested. Core records provisioning immediately before invoking it, carrying the prepared facts forward under the existing timeout and settlement owner. Fresh preparation failures need no provider cleanup; replay and post-allocation failures preserve their existing uncertainty. An invocation-local fence prevents late preparation from allocating after timeout or cancellation.

Crabbox uses this owner boundary for its CLI/config checks and removes its duplicate preflight cleanup path. The required public `WorkerProvider.provision` contract, verified at stable v2026.9.2, remains supported through the same prepared flow.

## User Impact

Fresh CLI/configuration failures remain actionable: operators can fork, retry or reclaim without a nonexistent lease blocking the session. Existing unresolved environments remain preserved. There is no schema bump, stored-format change, new configuration option or increased timeout.

## Evidence

The real non-executable CLI failure was reproduced. In an isolated production RPC-handler harness with the real Crabbox plugin and real stores, dispatch reported the process-launch failure; fork, same-session redispatch and reclaim then succeeded. Both environments stayed failed with no lease or destroy intent. There were zero provider destruction, enrollment or cloud requests. This exercises the real RPC handlers and process launch, not a network WebSocket server.

106 lifecycle/replay cases, 47 provisioning/node cases, 249 Crabbox cases and 22 registry cases pass. A late replay preparation regression fails before its invocation fence and passes afterward. The full changed gate, including schema guard, and independent P2 review pass.

Production delta is −21 lines; tests are net-neutral. Mutable image selection remains within the allocation owner because it can change provider resources; this repair specifically establishes the local preparation boundary. No old row was fabricated as deleted to produce the proof.

### Maintainer contract decision

Accept the optional allocation-free `prepareProvision` contract identified in review. Preparation may validate/read local configuration but must not allocate, renew, mutate provider resources, enroll nodes, or prepare a project. Its returned operation is the sole allocation entry, invoked only after core records allocation possibility and revalidates the invocation. The existing required `provision` SDK entry remains supported, and providers without the new hook retain their current path. Fresh-failure, timeout/abort, and unresolved-replay tests enforce the core side of this boundary; the real Crabbox preflight is allocation-free.

### CI follow-up

The first Linux compact-small-23 run stalled in the pre-existing CLI dotenv/JSON entry-validation process test after printing the expected `--container requires a value` diagnostic; its unchanged 240-second deadlock guard killed the child. The entry, test harness, dotenv and logging/exit owners have identical source blobs at this PR head and the tested main base. A real Node 24.19.0 entry probe on macOS imported none of this PR's changed production modules and exited with code 2. The same eight-file execution order passed locally (48 tests), and the individual case passed with both Node 24.19.0 and 24.20.0.

This remains an unconfirmed intermittent Linux CLI exit hang; no speculative source fix, weaker assertion, or timeout change was made. The exact failed job was rerun once and passed; the full [CI run](https://github.com/openclaw/openclaw/actions/runs/34093294208) and its aggregate gate are now green.


steipete:
Merged via squash.

- Prepared head SHA: [1ab02820fb6505f3a10e984adc48b7872977beea](https://github.com/openclaw/openclaw/pull/140945/commits/1ab02820fb6505f3a10e984adc48b7872977beea)
- Landed commit: [bf1ba1f3a27633e418d65f4e932eb6844fed6ecb](https://github.com/openclaw/openclaw/commit/bf1ba1f3a27633e418d65f4e932eb6844fed6ecb)

PR #140951, opened 2026-09-07
fix(plugins): install Git plugins from non-default branches

steipete:
<details>

</details>

## What Problem This Solves

Fixes an issue where users installing a Git plugin from a non-default branch, such as `git:owner/plugin#feature/demo`, would receive a Git checkout error even though the requested branch existed.

## Why This Change Was Made

Select the existing remote-ref resolution mode in the shared Git acquisition helper. It resolves the requested revision to a commit before detached checkout, including branches that the clone knows only through `origin`. The plugin installer continues to own package validation, installation, and recorded commit metadata.

## User Impact

Users can install and update plugins from non-default Git branches. Lightweight and annotated tags, full and abbreviated commit IDs, and default-branch installs retain their existing behavior.

## Evidence

The real installer regression failed on current main before the repair with `fatal: '--detach' cannot be used with '-b/-B/--orphan'`. After the one-line mode change, all 40 focused tests passed: both plugin Git suites and the shared Git-source suite. Actual installations use controlled local repositories containing a minimal plugin with no external dependencies. Coverage includes a non-default branch during both installation and update, lightweight and annotated tags, full and abbreviated commit IDs, default HEAD, installed file contents, and recorded commit identity.

Validation: `node scripts/run-vitest.mjs src/plugins/git-install-target.test.ts src/plugins/git-install.test.ts src/infra/git-source.test.ts --maxWorkers=1`; the owning `test/tsconfig/tsconfig.core.test.plugins-platform.json` type graph; full-file typed lint; formatting; and `git diff --check`.

The production diff replaces one line and adds no net lines. Tests add 59 net lines. The existing Git install documentation already describes the repaired behavior.

### DIFF
--- a/src/plugins/git-install-target.test.ts
+++ b/src/plugins/git-install-target.test.ts
@@ -61,6 +61,59 @@ describe("git install target ownership", () => {
     await state.cleanup();
   });
 
+  it.each(["branch", "tag", "annotated-tag", "commit", "short-commit", "default"] as const)(
+    "installs the requested %s revision from a real Git repository",
+    async (kind) => {
+      const defaultCommit = await git(sourceDir, "rev-parse", "HEAD");
+      await git(sourceDir, "switch", "-c", "feature/demo");
+      const featureCommit = await commitPlugin("demo", "2.0.0");
+      await git(sourceDir, "tag", "v2.0.0");
+      await git(sourceDir, "-c", "tag.gpgsign=false", "tag", "-a", "release", "-m", "release");
+      await git(sourceDir, "switch", "main");
+      const refs = {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7cc7e8a7929b0de88cf487db35ad3a88472a4df5](https://github.com/openclaw/openclaw/pull/140951/commits/7cc7e8a7929b0de88cf487db35ad3a88472a4df5)
- Landed commit: [78cbe2999b3974143d6de8ad2256aad016d02a19](https://github.com/openclaw/openclaw/commit/78cbe2999b3974143d6de8ad2256aad016d02a19)

Issue #140819, opened 2026-09-07
Queued preflight compaction repeats when its accounting is refused after the compactor releases the session lane ("Session changed before compaction maintenance could be recorded")


joeykrug:
### Summary

When a token-triggered preflight compaction **succeeds** (the `compaction` event is appended to the transcript and the checkpoint is persisted) but the session-row accounting write is refused afterwards, OpenClaw throws `Session changed before compaction maintenance could be recorded`, the queued message that triggered the preflight fails, and the session entry keeps its **pre-compaction** `totalTokens` marked fresh. The next queued turn's preflight projects pressure from that stale total and compacts again. On a busy session this repeats until an accounting write happens to land.

Observed: **4 committed compactions in 14 minutes** on one long-lived Signal direct session; three of them ran against an actual context of about 84k / 120k / 91k tokens with a 252k budget, and each showed a "Context compacted · saved Nk tokens" card in the Control UI.

### Environment

- OpenClaw `2026.9.2`, Node 24.15.0, Linux
- Effective agent harness `openclaw` (not the Codex app-server harness); model route: OpenAI subscription Responses (ChatGPT plan), context 272,000, reserve 20,000 → prompt budget 252,000
- `agents.defaults.compaction`: `{ mode: "safeguard", timeoutSeconds: 900, maxActiveTranscriptBytes: "25mb", memoryFlush.forceFlushTranscriptBytes: "16.67mb" }`. The byte fuse did not fire; every checkpoint has `reason: "auto-threshold"` and every trigger was the token path.
- Compaction ran inside the **followup queue drain** (queued turn preflight); the errors are logged as `followup queue drain failed for <session>: Error: Session changed before compaction maintenance could be recorded`.

### Timeline (UTC, 2026-09-07)

| compaction commit | checkpoint `tokensBefore` (caller projection) | transcript `tokensBefore` (actual) | accounting |
|---|---|---|---|
| 04:39:07 | 269,654 | 271,995 | refused 04:39:07.860 |
| 04:43:19 | 253,274 | 84,154 | refused 04:43:20.490 |
| 04:47:13 | 256,472 | 120,500 | refused 04:47:14.746 |
| 04:48:36 | 259,042 | 91,468 | landed |

Gateway log around each refusal (same session lane):

```
04:39:07.812 WARN lane wait exceeded: lane=session:… waitedMs=270836 queueAhead=0 activeAhead=1 activeNow=0 queueBehind=1
[...]

PR #140955, opened 2026-09-07
fix(auto-reply): queued turn re-compacts a small context when preflight accounting is refused

joeykrug:
Related: #140819

## What Problem This Solves

Fixes an issue where a busy chat session would compact its context again on the very next queued turn, even though the context was far below the compaction budget. Each spurious pass spent minutes summarizing, appended another summary on top of the previous one, blocked the queued message behind the session lane, failed that queued message with `Session changed before compaction maintenance could be recorded`, and surfaced another "Context compacted" card. On one Signal session this repeated four times in fourteen minutes; three of the four ran against a context of roughly 84k, 120k, and 91k tokens against a 252k budget.

## Why This Change Was Made

Token-triggered preflight compaction committed the compaction to the transcript inside the queued compactor's session-lane task, but recorded the session-row accounting (count, `totalTokens`, byte latch) only after that task returned and the lane was released. A run waiting on the lane claims the session writer as its first lane write, so the accounting compare-and-swap was correctly refused, the row kept its pre-compaction fresh total, and the next preflight projected pressure from that stale total and compacted again. The queued compactor already exposes a host commit callback that runs while the lane and writer fence are still held; the Codex transcript-byte path used it and the token path did not. This change routes accounting through that callback for every runtime, and supplies the transcript-settled callback for byte-triggered compaction on every runtime so the byte latch keeps measuring the post-maintenance transcript. Nothing about the writer, lifecycle, or session guards changes: a writer that replaces the row before the host commit is still refused, and the after-return path remains only for compactors that emit no host commit.

Non-goals: no retry against a replacement writer, no schema or config change, no change to projection inputs or context limits.

## User Impact

After a preflight compaction, the session's recorded context and compaction count match the transcript, so the following queued turn no longer re-compacts a small context, the queued message is delivered instead of failing, and users stop seeing back-to-back "Context compacted" cards for a session that was not over budget.

## Evidence

- Regression test `lands before a run waiting on the session lane claims the writer` (`src/auto-reply/reply/agent-runner-memory.test.ts`) models the queued-execution contract against a real SQLite session store: successor acceptance and host commit inside the lane, then a replacement writer claim once the compactor returns. Red/green on this head: with the pre-change `agent-runner-memory.ts` from `upstream/main` swapped in, that test fails with `Session changed before compaction maintenance could be recorded` thrown from `recordCompactionAccounting` (1 failed, 125 passed); with the change it passes, the count and `totalTokens` land, the replacement writer's claim is intact, accounting runs exactly once, and the next preflight does not compact again.
- Guard test `does not account against a writer that replaced the row before the host commit` keeps the newer-writer protection on the in-lane path.
- The byte-latch test now runs for both the Codex and the embedded runtime and asserts the persisted latch right after the first compaction equals the settled (post-maintenance) transcript size. The embedded case would fail if the settled-latch callback were supplied only for Codex, because the in-lane host commit measures the pre-maintenance size.
- Focused suites on this head (`node scripts/run-vitest.mjs`, one invocation): `agent-runner-memory` (126), `session-updates.lifecycle` (17), `compact.queued-successor` (25), `agent-runner-memory.preflight-stale-tokens` (7), `agent-runner-execution-compaction` (14), `commands-compact.lifecycle` (8): 197 passed. `compact.queued-successor` already pins that `onHostCompactionCommitted` is awaited before the checkpoint and maintenance steps.
- Changed-path gate on this head (`node scripts/check-changed.mjs --base upstream/main`): all 29 checks pass, including `tsgo:core`, `tsgo:core:test`, oxlint and `oxfmt` on the changed files, the max-lines and assertion-safety ratchets, dead-export scan, and import-cycle check; `git diff --check` is clean. A second red check confirms the byte-latch asymmetry: with the settled-latch callback supplied only for Codex, the embedded case fails (`activeBytes` 314 persisted instead of the settled 186).
- Production correlation (sanitized, from the reporter's gateway): each refused accounting was preceded by `lane wait exceeded` on the same session lane 50 to 330 ms before the accounting error, and the one compaction with no lane waiter recorded its accounting normally. Effective harness was `openclaw`; the model route was an OpenAI subscription Responses route.
- Independent review: repository `autoreview` helper on the Claude engine (`claude-fable-5-1`, xhigh), branch mode against `upstream/main`, P2 threshold: `scoped-clean`, no findings. An earlier Codex-engine autoreview of the pre-amend commit (P1 threshold) and an adversarial read-only concurrency review of the diff and tests were also clean at P0/P1; that review's P2 notes (byte-latch timing on non-Codex byte triggers, comment precision, two test-fidelity cleanups) are addressed in this head.

Known gap: this is a source-only repair validated with focused tests; it has not been applied to a live Gateway. The branch includes a merge of current `upstream/main`; upstream did not touch the changed files.


VACInc:
The central accounting race tracked in #140819 is now fixed on main by #141718 ([merge commit](https://github.com/openclaw/openclaw/commit/d12106ef815f9e36585cb53db639a1e447056cb1)). Comparing this PR's production diff with merged main confirms that both move `onHostCompactionCommitted` outside the native byte-preflight condition. Your report and regression coverage capture the same failure; the issue can now close against the landed repair.

This PR is **not fully superseded**: it additionally supplies `onHostCompactionTranscriptSettled` for transcript-byte compaction on every runtime. Merged main still supplies that callback only in the Codex byte-preflight branch. Your post-maintenance latch coverage therefore protects a remaining distinction that #141718 did not cover.

Leaving this PR open for that residual byte-latch correction and its regression coverage. No claim that the entire diff has landed.

PR #139976, opened 2026-09-06
fix(telegram): preserve styled links inside rich message containers

steipete:
## What Problem This Solves

Fixes literal HTML link tags and lost formatting when HTML containers and Markdown styles are mixed inside Telegram rich-message disclosures, lists, or quotations. Escaped tag examples and image alternative text stay literal, while supported authored HTML retains its attribute values and formatting.

## Why This Change Was Made

HTML containers, atomic inline content, and Markdown formatting now compose over the same parsed text ranges. One renderer owns that composition, replacing `applyInlineHtmlIslands` / `parseInlineHtmlIslands` and their per-string-leaf reparsing. The renderer preserves Markdown link/spoiler wrappers around formulas and custom emoji; transcript-role annotations remain outside surrounding active formatting. Native acceptance below uses the existing disabled-preview scenario setting.

The Markdown parser generalizes its existing underline lexical rule to recognize complete authored HTML lexemes before entity decoding or text joining. It records complete tags and their original bytes in non-enumerable `htmlTags` metadata, carrying those facts through token transformations, IR slicing/appending, construct fallbacks, and table cells. Partial tag fragments do not inherit a whole tag’s meaning. Telegram consumes these recorded facts rather than rescanning decoded text: there are no decoded `literalRanges`, post-decoding admission callbacks, sentinel strings, or second decoder. The private token leaves `html: false` and HTML block parsing unchanged, preserves ordinary anchor-body autolinks, and retains the existing underline opt-in.

This also preserves valid attribute grammar such as unquoted `x&#61;1`, quoted entity-delimiters, and Markdown punctuation inside URL values. The HTML mapper decodes supported attribute values after parsing their authored syntax; both entity-detection modes are covered. The obsolete tag-prefix link-suppression heuristic is removed, so a real Markdown link within incomplete tag-shaped text remains a link without manufacturing an HTML tag. Image alternatives use one plain-text owner; existing role-annotation metadata remains intact.

Unsupported HTML wrappers and their HTML descendants remain literal without discarding independently authored Markdown links or bold text. Their literal ownership is retained across Markdown blockquote/table slices, including inside disclosures. A table preceding a wrapper is not incorrectly assigned to it. The mapper reuses its text-range recorder and existing serializer instead of flattening unsupported subtrees into formatting-free atoms.

[...]

PR #140957, opened 2026-09-07
fix(plugins): restore source startup for Gateway error imports

steipete:
## What Problem This Solves

Fixes source-backed plugin startup when its SDK runtime imports Gateway error details. An isolated Telegram Gateway hit a module-not-found path ending in `gateway-protocol/src/index.ts/gateway-error-details` before its channel could start.

Related: #139976. This is a separate startup repair; it does not contain that PR's rich-message changes.

## Why This Change Was Made

The public protocol subpath and TypeScript mapping already exist, but the plugin loader's curated workspace map omitted the subpath. Jiti consequently applied the package-root alias and appended the subpath to `index.ts` or `index.mjs`.

Add that existing export to the same canonical map. Package manifests can be absent in installed layouts, so this preserves the established resolution contract without adding another manifest reader, resolver, or fallback. Protocol schemas, dependencies, configuration, and private SDK permissions are unchanged.

## User Impact

Source-backed plugins can load the existing Gateway error-details module. The same exact alias works for source and built workspace artifacts.

The production change is one necessary alias line; regression coverage adds 33 lines in the existing test file.

## Evidence

- Both new real-Jiti cases fail on the main baseline with the malformed `index.ts`/`index.mjs` paths and pass with the repair.
- The full alias and loader sibling run passed 80 tests, with one existing Windows-only skip.
- Full build, required changed-code gate, and fresh independent P0-P2 review passed.
- The built Telegram entrypoint import profile passed (161.09 MiB peak RSS); this is import profiling, not a live Telegram message test.
- Literal SDK/runtime import-graph inspection found the same missing specifier at two value imports; computed dynamic imports remain outside that static evidence.

An earlier cold source-only legacy SDK bridge test timed out on both the original code and the candidate at its unchanged 120-second limit. The fresh built-layout sibling run passes; this PR does not claim to repair or erase that earlier limitation.

The separate native custom-emoji failure in #139976 remains unresolved and is not counted as proof for this fix.

### Remaining runtime proof

A bounded actual source-Gateway run using the existing QA Lab/Crabline owners did not reach healthy channel readiness. The Gateway logged its HTTP listener, but the existing HEAD health check timed out after 120 seconds; zero Telegram API calls were recorded. Owned cleanup completed. The cause is not established, and this failed attempt is not replaced by the passing import tests or built-entry profile. This PR remains a draft pending that readiness investigation and exact-head CI.

### DIFF
--- a/src/plugins/sdk-alias.test.ts
+++ b/src/plugins/sdk-alias.test.ts
@@ -1370,6 +1370,39 @@ describe("plugin sdk alias helpers", () => {
     expectWorkspaceAliasTargets(aliases, workspaceAliases, "expectedDistFile");
   });
 
+  it.each(["src", "dist"] as const)(
+    "loads gateway error details through Jiti from %s workspace artifacts",
+    async (resolution) => {
+      const fixture = createPluginSdkAliasFixture();
+      const workspaceAliases = writeWorkspaceAliasFixtures(fixture.root, [
+        ["@openclaw/gateway-protocol", "gateway-protocol", "index"],
+        [
+          "@openclaw/gateway-protocol/gateway-error-details",
+          "gateway-protocol",
+          "gateway-error-details",
+        ],
+      ]);
+      const details = expectDefined(workspaceAliases[1], "gateway error details fixture");
+      fs.writeFileSync(details.srcFile, 'export const fallback = "source fallback";\n');
+      fs.writeFileSync(details.distFile, 'export const fallback = "built fallback";\n');
+      const entry = writePluginEntry(fixture.root, bundledPluginFile("demo", "index.ts"));
+      fs.writeFileSync(
+        entry,
+        'export { fallback } from "@openclaw/gateway-protocol/gateway-error-details";\n',
+      );
+      const aliases = buildPluginLoaderAliasMap(entry, undefined, undefined, resolution);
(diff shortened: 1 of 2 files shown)


steipete:
The alias repair is superseded by #140973 (`cb1b29d9a1e6e707fa469e394e4348d35a2fbedc`). Current main retains the exact `@openclaw/gateway-protocol/gateway-error-details` alias, its public package export, and source/dist coverage.

Verified on `cecf47a2d92e79c50bbc2c7949eebc871929e257`: **121 SDK alias, normalization, and loader-boundary tests passed**, with one existing Windows-only skip on macOS. A separate real-Jiti synthetic probe loaded the expected source and built modules; omitting only the exact alias reproduced the old `index.ts/gateway-error-details` and `index.mjs/gateway-error-details` failures.

This confirms the import-resolution fix. It does **not** establish that the historical source-Gateway HEAD-health timeout or Telegram readiness is resolved; that investigation remains separate from this duplicate alias patch.

The relevant owners and tests remain unchanged at main `566863b21cf661f1142e9504ec6a2bf820d39a47`, which contains the canonical fix. Reviewed proposal head: `70d3ed34b44768f812c5f1d215d47ddf0d57d8dc`.

PR #140958, opened 2026-09-07
fix: avoid retry delays for absent ClawHub tracking

steipete:
## What Problem This Solves

Fixes an issue where first-time ClawHub skill installs pause for about three seconds while checking tracking files that do not exist yet. The same waits made the ClawHub test file take 145 seconds locally.

## Why This Change Was Made

Recognize confirmed initial absence through the existing OpenClaw filesystem facade before invoking the strict JSON reader for optional workspace tracking. Present files and uncertain probe results retain strict parsing, errors, and replacement-race retries. This preserves rejection of malformed JSON, JSON null, and symlinks; a nullable JSON helper would conflate actual JSON null with absence.

Both the initial read and the publication-time reread remain. The tracking file is metadata; existing workspace mutation leases still own writer coordination. No metadata format, dependency, schema, lease, or trust-policy changes. The six added production lines use the existing filesystem primitive rather than introducing another reader or cache.

## User Impact

Fresh installs and operations on untracked workspaces respond promptly. Existing damaged tracking is still rejected and preserved. An optional read reports initial absence immediately rather than waiting for a future file; present-file replacement handling remains intact.

## Evidence

- Whole ClawHub file: **145.27s → 18.33s** with the same pinned Node/pnpm dependency install. All **117 original cases** remain, plus three filesystem preservation cases; candidate **120/120** passed. Case-time sum: **119.467s → 3.077s**.
- Representative fresh install: **3,559ms → 246ms**; early missing-resolution error: **1,733ms → 6ms**. These are local measurements, not workflow-time claims.
- Canonical and legacy metadata symlink rejection, plus real atomic replacement between inspection and open, passed against both the old reader and candidate. New symlink cases skip only when Windows actually denies the required file-symlink capability with EPERM.
- Uninstall rollback and workspace lease siblings: **15/15 passed**. Changed-code gate passed; the final capability-only test refinement passed focused tests, typecheck, formatting, and lint.
- Final wrapper repair: **122/122** ClawHub and filesystem-boundary cases passed, with focused typecheck/lint and independent P0–P2 review clean. The test mock now preserves real filesystem exports.
- Exact fs-safe **0.8.3** runtime and upstream source were inspected; [dependency evidence](https://github.com/openclaw/openclaw/pull/140958#issuecomment-5566765968) resolves both review inspection requests.
- The initial CI head passed all **120 ClawHub cases** (117 originals retained), totaling **687ms**, but failed the repository import-boundary guard. The follow-up routes the same primitive through the required policy facade; the guard is preserved.

No live registry write was performed; the change is local filesystem handling.

### DIFF
--- a/src/skills/lifecycle/clawhub-store.ts
+++ b/src/skills/lifecycle/clawhub-store.ts
@@ -1,5 +1,6 @@
 import fsSync from "node:fs";
 import path from "node:path";
+import { statRegularFile } from "@openclaw/fs-safe/advanced";
 import { normalizeOptionalString as normalizeOptionalStringValue } from "@openclaw/normalization-core/string-coerce";
 import type { ClawHubDownloadResult } from "../../infra/clawhub-artifacts.js";
 import {
@@ -286,6 +287,11 @@ export async function readClawHubSkillsLockfile(
 ): Promise<ClawHubSkillsLockfile> {
   for (const candidate of metadataPaths(workspaceDir, "lock.json")) {
     try {
+      // Missing metadata is normal before installation. Leave present-file races
+      // and uncertain paths to the strict reader, including its error diagnostics.
+      if ((await statRegularFile(candidate).catch(() => undefined))?.missing) {
+        continue;
+      }
       return parseClawHubSkillsLockfile(await readJson<Partial<ClawHubSkillsLockfile>>(candidate));
     } catch (err) {
       if (err instanceof JsonFileReadError && hasErrnoCode(err.cause, "ENOENT")) {
--- a/src/skills/lifecycle/clawhub-store.ts
+++ b/src/skills/lifecycle/clawhub-store.ts
@@ -1,6 +1,5 @@
 import fsSync from "node:fs";
 import path from "node:path";
-import { statRegularFile } from "@openclaw/fs-safe/advanced";
 import { normalizeOptionalString as normalizeOptionalStringValue } from "@openclaw/normalization-core/string-coerce";
 import type { ClawHubDownloadResult } from "../../infra/clawhub-artifacts.js";
 import {
@@ -10,6 +9,7 @@ import {
   type ClawHubSkillsShTrustState,
 } from "../../infra/clawhub-skills.js";
 import { formatErrorMessage, hasErrnoCode } from "../../infra/errors.js";
+import { statRegularFile } from "../../infra/fs-safe.js";
 import {
   JsonFileReadError,
   readJson,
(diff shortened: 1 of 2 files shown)


steipete:
Both dependency-inspection items in the review are now verified against the exact pinned **@openclaw/fs-safe 0.8.3**, including the installed runtime files used by the before/after runs and the upstream v0.8.3 source:

- [`statRegularFile`](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/regular-file.ts#L62) uses `lstat`, reports missing only for the [ENOENT/ENOTDIR set](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/path.ts#L16), rejects symlinks/nonregular files, and propagates other errors.
- [`readJson`](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/json.ts#L341) enables the existing [five-attempt retry loop](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/json.ts#L14). Initial ENOENT therefore waits 50+100+200+400ms per absent file. The dependency's [optional-read helper](https://github.com/openclaw/fs-safe/blob/v0.8.3/src/json.ts#L50) already performs the same initial-absence check before retaining replacement-race retries for observed files.
- This patch keeps strict `readJson` and lock-shape parsing after every present or uncertain probe, preserving JSON-null rejection and existing diagnostics. The real symlink/replacement cases passed on both implementations, and all 117 original cases remain.

Installed `dist/json.js` SHA-256: `b7d28e3e7ae92d76fe58bc684088fbbd3fabd4069f0a24b29d51de6d3b135d5c`; `dist/regular-file.js`: `c1d34f464150b69516db3ed3bb9663c7e1ca36fd35637fee6c54f15d61f7aff8`. The independent review received these installed implementations. No source change was needed for this evidence clarification; exact-head CI remains the final landing gate.

steipete:
@clawsweeper re-review

The follow-up routes statRegularFile through the existing OpenClaw filesystem facade and retains real exports in the test partial mock. All122 ClawHub/boundary cases, focused typecheck/lint, and final independent review pass. The exact0.8.3 dependency inspection requested earlier is documented in https://github.com/openclaw/openclaw/pull/140958#issuecomment-5566765968. Waiting for CI on the repaired head.

steipete:
Merged via squash.

- Prepared head SHA: [00033b383127d0944fe0e9e39dc20b2ef383d1ab](https://github.com/openclaw/openclaw/pull/140958/commits/00033b383127d0944fe0e9e39dc20b2ef383d1ab)
- Landed commit: [4d10c9dec8d9fa9de8297066a011c26b109851a8](https://github.com/openclaw/openclaw/commit/4d10c9dec8d9fa9de8297066a011c26b109851a8)

PR #140961, opened 2026-09-07
fix(markdown): preserve block boundaries inside list items

steipete:
## What Problem This Solves

Fixes an issue where users reading formatted replies would see separate paragraphs, headings, or commands joined together when they appeared inside a list item. An instruction followed by a fenced command could become `Run this:echo hello`, and separate quoted paragraphs could lose their boundary.

## Why This Change Was Made

The shared Markdown parser suppressed separators for tight or indirectly nested list paragraphs before channel rendering. Use its existing next-block and item source boundaries to preserve separators between blocks in the same item, while retaining the established spacing between tight and loose list items.

## User Impact

Lists containing instructions, code blocks, headings, or quoted paragraphs remain readable in replies that use the shared Markdown formatter. The formatting documentation now states this behavior.

## Evidence

Three regression cases failed before the repair while the existing 27 list cases passed. Afterward, 200 focused IR and renderer/chunk tests pass, including existing list nesting, quote spacing, block metadata, links, and source-spacing cases.

Direct calls through the Markdown parser and both marker and attributed-range renderers confirm that `Run this:` and `echo hello` appear on separate lines, headings keep separate surrounding prose, and quoted paragraphs retain their paragraph break. Code, link, and bold ranges remain aligned. These are formatter API probes; no live channel delivery is claimed.

Scoped Markdown production and test types, full-file lint and formatting, the max-lines check, and diff checks pass. Independent review found no actionable findings through P2. The candidate adds 8 production lines, 20 test lines, and 2 documentation lines. Required CI must pass on the published head before landing.

### DIFF
--- a/docs/concepts/markdown-formatting.md
+++ b/docs/concepts/markdown-formatting.md
@@ -95,6 +95,8 @@ channels:
   channels render the closing fence correctly.
 - List and blockquote prefixes are part of the IR text, so chunking never
   splits mid-prefix.
+- Paragraphs, headings, and code blocks inside a list item stay separated in
+  the IR, including paragraphs nested inside a quoted list item.
 - Inline styles never split across chunks; the renderer reopens an open
   style at the start of the next chunk.
 
--- a/packages/markdown-core/src/ir.nested-lists.test.ts
+++ b/packages/markdown-core/src/ir.nested-lists.test.ts
@@ -300,6 +300,26 @@ describe("Nested Lists - Edge Cases", () => {
 describe("list paragraph spacing", () => {
   it.each([
     {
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [93ec1f2194b1fa9662289cfb7fe366f1c90b7e44](https://github.com/openclaw/openclaw/pull/140961/commits/93ec1f2194b1fa9662289cfb7fe366f1c90b7e44)
- Landed commit: [934dc2028c09b4de67b9a2ad960f8dd2ae90f8af](https://github.com/openclaw/openclaw/commit/934dc2028c09b4de67b9a2ad960f8dd2ae90f8af)

PR #140969, opened 2026-09-07
fix(matrix): include alternative aliases in room info

steipete:
<details>

This PR uses a branch in the same repository, so maintainers can update it directly.

</details>

## What Problem This Solves

Fixes an issue where users requesting Matrix channel information would receive an empty alternative-alias list even when the room advertised additional addresses.

## Why This Change Was Made

Return `alt_aliases` from the existing `m.room.canonical_alias` state response using the shared string-entry helper. This removes the hardcoded empty result and the stale assumption that a separate request was required.

## User Impact

Channel information includes the room's advertised alternative addresses alongside its canonical alias. Other room metadata remains available as before, with no additional network request.

## Evidence

- The regression failed before the repair when the server response contained two alternative aliases; the other three room-action tests passed. All four tests pass after the repair, covering both absent and populated aliases and the existing room/member fields.
- A synthetic loopback HTTP homeserver exercised the real Matrix SDK, OpenClaw adapter, and `getMatrixRoomInfo`. Before: `altAliases` was `[]`. After: it contained `#incidents:example.org` and `#team:example.org`, with exactly one canonical-alias state request. The fixture did not start sync or use saved accounts; final cleanup awaited `stopWithoutPersist()`. This is local adapter/protocol evidence, not a live public homeserver test.
- Full extension production and test typechecks, full-file lint, formatting, and `check:changed` passed. Independent review found no actionable P0–P2 issues.
- Production delta: +4/-1 lines for projecting the already-fetched state through a shared helper; test delta: +11/-5 lines. No new settings or persistence behavior.

### DIFF
--- a/extensions/matrix/src/matrix/actions/room.test.ts
+++ b/extensions/matrix/src/matrix/actions/room.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, it, vi } from "vitest";
 import type { MatrixClient } from "../sdk.js";
 import { getMatrixMemberInfo, getMatrixRoomInfo } from "./room.js";
 
-function createRoomClient() {
+function createRoomClient(altAliases?: string[]) {
   const getRoomStateEvent = vi.fn(async (_roomId: string, eventType: string) => {
     switch (eventType) {
       case "m.room.name":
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3897df2fbd260b2377436f292428d2d80e4af724](https://github.com/openclaw/openclaw/pull/140969/commits/3897df2fbd260b2377436f292428d2d80e4af724)
- Landed commit: [58ad5780b53c2f793aef8641ba27f4095aacc717](https://github.com/openclaw/openclaw/commit/58ad5780b53c2f793aef8641ba27f4095aacc717)

PR #140970, opened 2026-09-07
refactor(bench): require process metrics directly

steipete:
## What Problem This Solves

Gateway startup benchmark validation retains an unused switch for allowing missing CPU/RSS samples. The benchmark already requires these metrics on every platform, while five tests redundantly override the switch to the same value.

## Why This Change Was Made

Remove the unused optional argument and make the existing CPU/RSS checks direct. The tests now exercise the same one-argument path as the benchmark entry point. No cases, fixtures, or assertions are removed.

## User Impact

No behavior change: missing process metrics still fail the benchmark, zero values remain valid, and failure ordering, child-exit handling, incident budgets, and report formatting are preserved. This removes six tooling lines without adding a replacement seam.

## Evidence

All 74 startup, restart, and shared-runtime cases pass before and after with identical case/status inventories. Independent Codex review found no actionable findings. History confirms the all-platform requirement predates this cleanup; no false caller remains. No new platform execution or performance result is claimed.

All 19 normal changed-file checks passed on the final source in Linux Testbox. The one-shot lease and source capsule are closed.

### DIFF
--- a/scripts/bench-gateway-startup.ts
+++ b/scripts/bench-gateway-startup.ts
@@ -473,11 +473,7 @@ function summarizeCase(benchCase: GatewayBenchCase, samples: GatewaySample[]): C
   };
 }
 
-function collectResultFailures(
-  results: CaseResult[],
-  options: { processMetricsRequired?: boolean } = {},
-): BenchmarkFailure[] {
-  const processMetricsRequired = options.processMetricsRequired ?? true;
+function collectResultFailures(results: CaseResult[]): BenchmarkFailure[] {
   const failures: BenchmarkFailure[] = [];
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f79fc9ec78e106450a7dfca3832fa77575efeee0](https://github.com/openclaw/openclaw/pull/140970/commits/f79fc9ec78e106450a7dfca3832fa77575efeee0)
- Landed commit: [24707ee6827749a376e88f314e73e4b39076d4f8](https://github.com/openclaw/openclaw/commit/24707ee6827749a376e88f314e73e4b39076d4f8)

PR #135862, opened 2026-09-02
fix(gateway): keep service-derived status targets off the explicit-URL credential path

SunnyShu0925:
Closes #135471

## What Problem This Solves

Resolves a problem where operators running `gateway status --require-rpc` against an auth-none local Gateway receive an explicit-credential error even though they did not supply `--url`.

Service diagnostics could also select credentials or connection details from a configured remote Gateway instead of the service being inspected. An empty target-scoped credential result could fall back to unrelated config or environment credentials later in the RPC path.

## Why This Change Was Made

Service-derived targets now remain distinct from explicit URL overrides throughout the status probe. The selected service target owns its local credentials, merged service environment, and certificate, regardless of the configured client remote mode.

- `serviceTargetUrl` carries the service URL without classifying it as CLI `--url`.
- Status uses a nonpersisted command-local config view so downstream resolution cannot borrow remote Gateway credentials or conflicting target settings. Service auth mode remains available for RPC preflight.
- The view preserves only the existing remote URL scope, `edgeAuth`, and `tlsFingerprint`. Existing consumers retain their endpoint-scoped transport policy: proxy headers use the existing origin/path selector, and ordinary connectivity probes retain the saved pin only on an exact configured endpoint. Unrelated token/password fallback and service certificate inheritance remain excluded.
- `skipImplicitAuth` preserves the already-resolved credential set, including an empty set for auth-none.
- Explicit `--url` uses only explicit RPC token/password and does not inherit the service certificate.
- Saved pins remain configuration input to their existing consumer; they are not promoted to an explicit fingerprint argument, and RPC TLS selection is unchanged.
- Connection details report the same service URL that the probe uses.
- The command-local projection lives with the existing status target helpers, keeping the gather function below its existing size limit.
- Advisory plugin install-record reads use the existing artifact-preserving read-only mode, so status and diagnostics do not create SQLite sidecars beside the inspected source database.

Generic credential precedence, persisted configuration, schemas, and protocols are unchanged. Explicit URL credential requirements and read-only device-auth behavior remain intact.

## User Impact

Operators can inspect an auth-none service without inventing a token or password. Service status remains tied to the selected service even when remote client settings or an environment URL are present. Explicit remote probes retain their credential isolation, while configured proxy headers and saved certificate restrictions remain available under their existing endpoint-scoping rules.

## Evidence

- Before main integration, 267 focused tests passed: 154 Gateway call tests and 113 daemon status/probe tests.
[...]

PR #140975, opened 2026-09-07
fix: restore local password auth in gateway status

steipete:
Related: #135862

## What Problem This Solves

Resolves a problem where `openclaw gateway status` reported an authentication failure for a local Gateway in trusted-proxy mode even though its local-direct password was configured. The JSON command could exit successfully with `rpc.ok: false`, causing callers that inspect that field to reject an otherwise reachable Gateway.

## Why This Change Was Made

Status credential selection grouped trusted-proxy with credential-free `none` mode. Trusted-proxy already supports password authentication for local-direct clients, so it must use the existing probe credential resolver. This restores configured, environment and SecretRef password selection while preserving explicit URL/auth precedence, remote credential isolation and disabled-exec restrictions. Server authorization is unchanged.

## User Impact

Local status checks can authenticate with the existing password configuration, including under an external supervisor. Explicit URL targets still require their own credentials. The documentation clarifies the default local-target behavior.

## Evidence

- Baseline: four intended credential-contract failures, with 16 existing controls passing.
- Candidate: 199 focused/sibling tests passed; two existing Windows/macOS-only cases skipped. This includes all four regressions, status JSON projection, target/probe behavior, and canonical probe auth.
- Independent P2 review found no actionable findings; unchanged credential owners and the server's local-direct contract were also directly audited.
- On `b6fe8d2454d2`, full changed-scope checks passed on Linux/Node 26.7. An ordinary source-entry `gateway status --json` call against a real isolated Gateway returned `rpc.ok: true` with external-supervisor context, fresh client state and normal password configuration. No credential/URL overrides or model turns were used; the client process, Gateway and temporary state closed cleanly. This is actual Vitest minimal Gateway startup plus the source-entry CLI, not compiled packaging or full startup validation.
- OpenGrep reported an existing TLS test server missing an explicit payload bound. The follow-up changes only that fixture to its existing advertised 1,000,000-byte limit; the two matching/mismatched TLS cases and independent incremental review cover that line. Production source and documentation remain identical to the completed b6 proof.
- A preliminary proof preflight refused an incorrectly assumed source-carrier parent before any workload ran. The successful run verifies the canonical carrier base and exact published source tree, retaining the original commit and managed closure receipts. No source-trust check was bypassed.

### DIFF
--- a/docs/gateway/trusted-proxy-auth.md
+++ b/docs/gateway/trusted-proxy-auth.md
@@ -105,7 +105,7 @@ The `deviceAutoApprove` examples below target beta/current-main builds. Stable `
 
 `allowLoopback` trusts local processes on the Gateway host to the same degree as the reverse proxy. Enable it only when the Gateway is still firewalled from direct remote access and the local proxy strips or overwrites client-supplied identity headers.
 
-Internal Gateway clients that do not travel through the reverse proxy should use `gateway.auth.password` / `OPENCLAW_GATEWAY_PASSWORD`, not trusted-proxy identity headers. Non-loopback Control UI deployments still need explicit `gateway.controlUi.allowedOrigins`.
+Internal Gateway clients that do not travel through the reverse proxy should use `gateway.auth.password` / `OPENCLAW_GATEWAY_PASSWORD`, not trusted-proxy identity headers. `openclaw gateway status` selects this local password automatically when no `--url` override is supplied, including with `--json`. Non-loopback Control UI deployments still need explicit `gateway.controlUi.allowedOrigins`.
 </Warning>
 
 ### Configuration reference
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e62e97bfb50ddc499dfc0832c33eb0da1269901d](https://github.com/openclaw/openclaw/pull/140975/commits/e62e97bfb50ddc499dfc0832c33eb0da1269901d)
- Landed commit: [b2d7fe9fe332c9f6c043bc00ec3f3ec07ddce0f5](https://github.com/openclaw/openclaw/commit/b2d7fe9fe332c9f6c043bc00ec3f3ec07ddce0f5)

PR #122221, opened 2026-08-11
fix(auto-reply): respect fast-auto progress visibility

vincentkoc:
## Summary

- make fast-auto tool progress follow the canonical hidden, enabled, and forced visibility policy
- apply the same callback/fallback contract to direct and queued follow-up execution
- preserve Telegram draft ownership and one-shot durable fallback
- preserve iMessage native typing without creating a durable progress bubble

Fixes #97601.

This replaces #120652. That PR's exact head `730963df60a65fc60858847428cee953e78e0c32` is stale/conflicting and `maintainerCanModify=false`, so maintainers cannot repair it in place. Its head must not land.

Credit to @joeykrug for the report and original repair direction. The implementation history preserves co-author credit.

## Root cause and repair

Fast-auto status payloads crossed the auto-reply visibility boundary through separate direct and queued paths. Hidden progress could reach a channel callback, while callback rejection had inconsistent fallback behavior.

The auto-reply dispatcher owns visibility authorization. Channel callbacks report whether they produced an operator-visible result:

- `true` or legacy `void` accepts ownership and prevents duplicate delivery
- `false` declines visibility
- hidden fast-auto progress remains suppressed after a decline
- enabled or forced fast-auto progress falls back to one durable delivery after a decline

The same contract now applies to initial dispatch and queued follow-ups. Source-suppressed room events remain silent. Telegram renders an authorized progress draft or sends one fallback message; iMessage may refresh native typing through its lifecycle-only opt-in without sending a progress bubble.

The candidate is based on `a7da4314609485e5296de0f5e8a56486122706dd`. Before publication on September 4, 2026, latest `main` was `e0c230789b33d4a59a8cda1f7b0858c4af0d5e7a`; the candidate and intervening main changes had no changed-path overlap, and a synthetic merge reported no conflict, so no churn-only rebase was performed.

## Current candidate

- Exact head: `97109f7bc0f1ede817c36a5c5f7113ce83e11e5f`
- Production LOC: `+87/-41` (net `+46`)
- Tests/test support: `+498/-22` (net `+476`)
- Total: `+585/-63` (net `+522`)

Production growth applies one visibility/ownership invariant across the direct dispatcher, queued executor, Telegram callback boundary, and iMessage lifecycle-only path. The additional test support covers the direct/queued policy matrix and sibling channel behavior.

## Local validation

All commands below ran against exact head `97109f7bc0f1ede817c36a5c5f7113ce83e11e5f`.

- focused direct + queued fast-auto matrix: 29 passed
[...]

PR #140988, opened 2026-09-07
feat(macos): open links as native Mac tabs in the Browser panel

steipete:
## What Problem This Solves

Resolves a problem where the macOS app kept two separate browsers. External links clicked in the dashboard opened in a native WKWebView split-view sidebar beside the dashboard, while the Control UI's Browser panel showed the Gateway-controlled Chromium (the browser agents drive) as streamed screenshots. With **Open links in Control UI browser** enabled, a link the user only wanted to read went to the remote browser and arrived as a blurry, polled image inside the app.

## Why This Change Was Made

The Browser tab of the unified side panel (and the shell-level Browser dock on non-chat routes) becomes one chrome with two tab kinds. **Mac tabs** are real WKWebViews owned by the macOS app and overlaid on the panel's stage: the web panel reports geometry, visibility, and occlusion, and the app owns tab lifetime and navigation state and pushes it back through the new `openclawBrowser` bridge (`ui/src/app/native-browser-bridge.ts`, mirrored by `DashboardBrowserMessageHandler.swift`). **Agent browser tabs** keep the Gateway browser behavior, including the live screencast that #141031 introduced; a Mac tab never starts or acks a screencast, and switching between the two kinds hands the stage between the native view and the stream. Mac tabs are window-owned and survive chat session switches. The split-view sidebar, its divider, width memory, and the app-side `inline` link route are removed; an older Control UI bundle that still posts `inline` opens the default browser instead. Annotate and Inspect on a Mac tab take a one-shot snapshot; top-layer surfaces (command palette, modal dialogs, menus, hovercards) hide the native view while open, with a stage hit-test fallback. Same-link clicks reuse the existing tab, including a retained original URL after a redirect. Downloads and non-displayable responses hand off to the default browser only for a user-activated main-frame navigation. The Inspect page script is generated into Swift from the UI source (`scripts/generate-browser-inspect-script-swift.mts`, verified in CI). No Gateway protocol, schema, or config changes.

## User Impact

In the macOS app, links open as crisp native Mac tabs in the Browser panel next to Agent browser tabs, with a tab strip, URL bar, back/forward/reload/stop, open-in-default-browser, Annotate, and Inspect. The panel is available in the app even when the Gateway does not advertise `browser.request`. Right-click a link for **Open in Browser Panel**, **Open in Default Browser**, or **Copy Link**. The **Open links in Control UI browser** preference is hidden inside the macOS app because links always open as Mac tabs there; while a Settings page has the viewport, a clicked link opens in the default browser instead of an invisible dock. Regular browsers are unchanged. Only the one-line bridge probe sits on the startup path; the bridge itself loads with the Browser panel chunk, and the native link context menu loads on first use inside the app, so regular-browser startup JavaScript ends up about 7 KiB gzip smaller than `main` rather than larger, with the startup sequence itself unchanged. Docs: `docs/platforms/macos.md`, `docs/web/control-ui/panels.md`, `docs/platforms/mac/webchat.md`, `apps/macos/AGENTS.md`.

Release-note context: macOS app link browsing now lives in the Browser panel as native Mac tabs; the split-view link sidebar is gone.

## Evidence

Live run on a Developer ID signed package of this branch, launched under a throwaway named app profile against an isolated Gateway (remote-direct mode, fresh state dir, free port):

| Link click opens a Mac tab in the side panel | Command palette hides the native view |
| --- | --- |
| Mac tab | Palette occlusion |

| Annotate captures a one-shot snapshot | Link context menu (native view hidden while open) |
| --- | --- |
| Annotate | Context menu |

Also verified live: the view returns when the palette or menu closes, a second click on the same link reuses the single tab, form interaction inside the Mac tab goes straight to WebKit, and closing the tab returns the panel to its empty state with a cleared address bar. No "before" capture of the removed split sidebar is included; the previous behavior is the sidebar documented in `docs/platforms/macos.md` on `main`.

Proof runs: `node scripts/check-changed.mjs` green on the final head (its lint lane run in a flat checkout of the same head, because the declaration-boundary guard rejects nested worktrees by design); focused Vitest for the bridge, panel, occlusion, routing, menu, and modal suites; `swift build` for the product and `--build-tests`; pure-logic Swift suites (`DashboardNativeBrowserContractTests`); `scripts/format-swift.sh` and `scripts/lint-swift.sh`; `pnpm ui:i18n:baseline` and `verify`; `pnpm check:browser-inspect-script:swift`. Six Codex autoreview passes at P2: three accepted findings on the first pass, one on the second (crypto.randomUUID on insecure origins), one on the third (a superseded new-tab request clearing the selected address), two on the fourth (automatic fallback selection cancelling a newer pending open; explicit blank new tabs collapsing onto an existing blank tab), each fixed with a regression test that failed before the fix; passes five and six clean, the last on the head that also fixed the four exact-head CI failures (stale native-link tests and a type-only import cycle) and passes `pnpm check:architecture`. Older app builds without the `openclawBrowser` bridge now open external links in the default browser. One unrelated hosted failure on the final head (`chat-permission-refresh.e2e.test.ts` timing out on the refresh warning) is the intermittent case fixed separately in #141020; it passes locally on this head and its job was rerun once. Two later merge-ref failures came from `main` itself, not this branch: the Telegram progress-updates test crossed the max-lines cap in #122221, and the iOS Periphery scan flags an unused `description` from #141003; both are routed to separate repairs. Window-level Swift tests compile locally and run in CI. Two shared-lane test fixtures were hardened along the way: the command palette occlusion test stubs the real `window.webkit` boundary instead of a per-file module mock, and the lazy-recovery e2e seeds its collapsed-navigation intent at document init instead of racing the router before Retry.


steipete:
The frozen-target guard for `check:browser-inspect-script:swift` is fixed in `734832e55d58defb1f9500ad53482d5dfb8f3cef`: older frozen targets skip the absent script, while current targets still require it.

Native routing taking precedence over the old link preference, hiding that preference in the macOS app, and legacy `inline` requests opening the default browser are intentional. The behavior is documented in `docs/platforms/macos.md` and `docs/web/control-ui/panels.md`; the evidence captures are embedded in the PR body.

steipete:
Merged via squash.

- Prepared head SHA: [dc0ce11edf9efb4dc74c5699c9558f4fc37931e1](https://github.com/openclaw/openclaw/pull/140988/commits/dc0ce11edf9efb4dc74c5699c9558f4fc37931e1)
- Landed commit: [fea46682a940cae0077d5e7af01058935defa739](https://github.com/openclaw/openclaw/commit/fea46682a940cae0077d5e7af01058935defa739)

PR #140842, opened 2026-09-07
fix(plugins): avoid runtime loading during bundled enablement

steipete:
## What Problem This Solves

Enabling a bundled plugin without an exclusive kind can execute its runtime solely to discover that it owns no slot. The CLI also reconstructs metadata after it has already prepared the same bundled manifest. Cloud-node bootstrap calls this path during activation; an earlier warm-start observation spent 41.201 seconds in that phase.

## Why This Change Was Made

Bundled manifests are authoritative for exclusive slot kinds. Reuse their prepared metadata under the existing lifecycle lease and avoid runtime inspection when no kind is declared. External CLI plugins retain fresh inspection against the enabled config and the documented legacy runtime-kind fallback. The discarded selected-plugin projection is removed instead of adding a cache or another activation path.

## User Impact

Bundled enablement persists activation policy without executing the plugin just to resolve its slot. Slot selection, capability consent, restrictive allowlists, missing-plugin results, config-write ownership, and registry refresh keep their existing behavior. There are no added, removed, or changed config keys/defaults, CLI flags, schemas, or stored representations. Existing external plugins with runtime-only kinds retain their compatibility path; bundled memory owners already declare their kinds in their manifests.

## Evidence

- The real config-mutation regression failed before the fix because a bundled plugin's module executed. With the fix it verifies persisted enablement, declared memory-slot selection, and no bundled runtime import. External legacy CLI enablement still imports the required runtime and selects its memory slot. All 58 focused tests pass across `src/plugins/slots.test.ts`, `src/plugins/status.runtime-inspection.test.ts`, and `src/cli/plugins-cli.policy.test.ts`.
- Changed-surface gates and independent managed review passed. [Exact-head CI run 34085636778](https://github.com/openclaw/openclaw/actions/runs/34085636778) is successful for `4f2b533c468c0d6d8021fcc02d481558f636bba5`.
- `pnpm build` passed in 245.7 seconds on the frozen combined validation source: captured main `28a4d41744af04bc68c6877fdc19753f7ef96b77` plus this plugin patch, the separate warm-readiness patch, and private diagnostic instrumentation. This is an instrumented combined candidate, not unmodified main or an isolated timing experiment for this PR.
[...]

PR #140990, opened 2026-09-07
fix(codex): avoid first-turn failures on cold cloud workers

steipete:
## What Problem This Solves

Fixes an issue where users starting a Codex cloud session could lose the first turn when the worker's native executable was cold. A live snapshot-backed worker accepted the initialize frame while its native process was still starting: native listening took 10.544 seconds after spawn, while the Gateway-side initialize connection closed after 10.002 seconds.

## Why This Change Was Made

The node command advertised its duplex receiver immediately after spawning the process. It now announces readiness after the managed native process starts listening, keeping native startup inside the existing cancellable launch lifetime and preserving the single Codex initialize exchange. Stderr and exit observers are installed before process registration yields, and execution authority is revalidated before receiver publication. Process cleanup, ordered framing, and the existing handshake timeout remain unchanged.

This deliberately uses an internal contract of the pinned `@openai/codex` 0.153.4 executable: its transport emits a listening line immediately before processing stdio. The launch owns the exact diagnostic filter, and the real pinned-binary integration guards the contract. This is not a new public Codex readiness API; package updates must continue to satisfy that integration. The node path already permits only the managed package's native executable, including in OpenClaw's v2026.9.2 release; custom Gateway app-server commands are unaffected.

The production delta is +44 lines: bounded, fragmented startup-line handling plus the readiness/exit race and post-wait authority check. Existing stderr and exit handling moved into the same owner rather than adding another process, protocol proxy, or preparation path.

## User Impact

The first turn can wait for a cold native process to become usable instead of spending the protocol handshake budget on process startup. Cancellation still terminates the owned process and removes its private home.

The fixed startup diagnostic `RUST_LOG` filter is inherited by native task processes under Codex's default `inherit=all` environment policy. This is an operational tradeoff: an explicit command-level `RUST_LOG` value can select the log level a program needs. The docs state this behavior; the change does not rewrite task environment policies or request payloads.

## Evidence

- Live pre-fix trace distinguished native startup delay from transport delivery: initialize reached node stdin before the native listening marker, and the Gateway closed at its existing 10-second deadline.
- A separate diagnostic on a fresh fork read and hashed the same 258,659,424-byte executable before launch. Reading took 39.943 seconds; the following native initialize took 91 ms, with a 12 ms repeat. This supports cold executable I/O as the startup delay; a full-file prefetch is not included in this fix.
- Controlled delayed-startup cases failed on the base because the receiver was published prematurely, then passed with the fix. Coverage includes fragmented readiness, a marker emitted during process registration, first-initialize preservation, cancellation, authority revocation, early exit, and oversized diagnostic lines.
- Existing real pinned-binary, relay, duplex-framing, and node-host suites passed. Production and test extension types and the full changed gate passed. Independent P2 review found no actionable findings.
- Exact dependency contracts inspected: [stdio listening](https://github.com/openai/codex/blob/rust-v0.153.4/codex-rs/exec-server/src/server/transport.rs#L149), [initialize deadline](https://github.com/openai/codex/blob/rust-v0.153.4/codex-rs/exec-server/src/client.rs#L697), and [default environment policy](https://github.com/openai/codex/blob/rust-v0.153.4/codex-rs/protocol/src/config_types.rs#L261).

The full combined candidate build passed in 303.8 seconds with this exact repair patch included.

Exact-head [CI run 34097560695](https://github.com/openclaw/openclaw/actions/runs/34097560695) passed on the first attempt: 84 jobs succeeded and 16 were skipped.

### After-fix live verification

The real cold-capture → warm-clone pair passed with the same runtime archive (`0f76db37c27866f48dff208da9864b3390c6d7e015611a99a1dc0d0f8ffdd034`). The combined build used captured main `86a3c7267be4` plus this exact repair and the now-merged plugin-activation/preallocation repairs (#140842, #140945). All three files in this PR were byte-compared with head `15acedda2070`; the tested build contained no private diagnostic instrumentation.

| Flow | Real tool turn | Repository continuity | Runtime setup |
| --- | --- | --- | --- |
| Cold capture | Passed, 58.929s | Restored 8 retained files and created the ninth | Downloaded and installed the runtime, then captured an available native image |
| Warm clone | Passed, 65.872s | Restored all 9 files and created the tenth | Used that same image/archive; no runtime download or npm installation |

Both flows completed normal Stop. Authoritative provider state was released with cleanup complete, canonical environments were destroyed with no attachments, and all 34 cold / 32 warm recorded process birth identities were absent. Gateway ports were closed, controllers joined, preexisting owners were preserved, and no coordinator version drift was observed. Neither flow created an additional Gateway repository checkout.

This completes ClawSweeper's requested after-fix cold/warm proof. It establishes the first-turn correctness repair, not an overall startup-speed claim: warm dispatch still took 325.591s, including 36.649s in plugin activation. That variable startup latency remains a separate performance follow-up.

### DIFF
--- a/docs/plugins/codex-harness.md
+++ b/docs/plugins/codex-harness.md
@@ -248,7 +248,10 @@ openclaw gateway call sessions.dispatch \
 ```
 
 The node starts the same managed, pinned Codex binary with
-`codex exec-server --listen stdio` in the placement workspace. The Gateway
+`codex exec-server --listen stdio` in the placement workspace. The node waits for
+the native process to start listening before announcing execution readiness, so
+cold disk startup does not consume Codex's initialize-handshake budget. Startup
+remains cancellable through the existing attempt lifecycle. The Gateway
 relays complete Codex JSON-RPC messages through the existing authenticated,
 approval-gated duplex node channel, with a 64 MiB limit per message. It does not
 start an OpenClaw worker child, open a reverse tunnel, or copy provider, cloud,
@@ -261,6 +264,10 @@ credential-free endpoint. The node process uses a fresh private
 environment and requested child-process environments are sanitized. Completed
 filesystem changes reconcile into the Gateway-owned managed worktree or, for
 repository-only sessions, an immutable checkpoint retained by the Gateway.
+
+Native task processes inherit the exec-server's startup `RUST_LOG` filter under
+Codex's default environment policy. Set `RUST_LOG` explicitly in a command when
+that program needs a different log level.
 
 Disconnecting the node, closing the app-server connection, cancelling the turn,
 or retiring the plugin ends that Codex attempt visibly and terminates its remote
--- a/extensions/codex/src/node-exec-server.readiness.test.ts
+++ b/extensions/codex/src/node-exec-server.readiness.test.ts
@@ -1,0 +1,176 @@
+import type { ChildProcessWithoutNullStreams } from "node:child_process";
+import { once } from "node:events";
+import { access } from "node:fs/promises";
+import { fileURLToPath } from "node:url";
+import type { OpenClawPluginNodeHostCommandIo } from "openclaw/plugin-sdk/node-host";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { setManagedCodexPluginRoot } from "./app-server/managed-binary.js";
+import * as transport from "./app-server/transport-stdio.js";
+import { runCodexNodeExecServer } from "./node-exec-server.runtime.js";
+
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [15acedda207068cf9c16359a2fd0f1fb84f7adbd](https://github.com/openclaw/openclaw/pull/140990/commits/15acedda207068cf9c16359a2fd0f1fb84f7adbd)
- Landed commit: [0d30dcbbf4d7a15936f3d71a0fda888d77df01ea](https://github.com/openclaw/openclaw/commit/0d30dcbbf4d7a15936f3d71a0fda888d77df01ea)

PR #140994, opened 2026-09-07
feat(cron): report slow list page timings

steipete:
## What Problem This Solves

Resolves a problem where operators investigating a slow cron list cannot tell whether a source page waited before entering cron's serialized callback or spent that time inside it. Gateway RPC timing includes both and does not expose the internal cron queue boundary.

## Why This Change Was Made

The existing list-page owner records four monotonic timestamps and emits one warning only when total elapsed time reaches one second. The record separates wait-to-callback, callback and subsequent settlement time, retains existing trace context, and includes aggregate row counts, outcome and emitter process/thread identity. Queue ordering, filtering, returned pages and original exceptions remain unchanged; logging failures cannot replace an operation outcome.

## User Impact

Operators can read `cron: slow list page` records in the existing file log to narrow a slow page's next investigation. These are wall times, not CPU time or pure lock contention. Caller visibility filtering and delivery previews outside the service page are not measured. No new configuration, RPC, dependency or database change, and no claimed latency improvement.

## Evidence

- Focused regression and read sibling suites: 41 tests passed across four files after refreshing onto the current cron owner-resolution fix. Baseline failed four missing-warning assertions; existing page and queue behavior passed.
- Real source-runtime queue/file-logger proof on Node 26.8.1 on macOS: baseline produced no slow records; candidate main thread and Worker each produced exactly three, with caller trace, actual emitter identity, and no job/error text. Held-queue and callback-only work were distinguished; original callback errors and later reads were preserved.
- A separate real microtask-delay control reported 0ms before entry, 1ms callback and 1100ms subsequent settlement, avoiding false attribution of that delay to callback work.
- Independent review found no actionable findings. Initial head `0876b965` passed all 28 blocking changed checks (plus the warning-only report), including all 16 core-test type graphs. The required refresh preserves main’s new unresolved-owner behavior; its 41 focused tests and main/Worker/settlement source-runtime proofs passed again. [Final CI](https://github.com/openclaw/openclaw/actions/runs/34098571394) passed on `e695605b`; native prepare and merge completed with that head unchanged.

The native proof exercises the real source owners and file logger with synthetic already-loaded scheduler state; it is not a compiled Gateway or hydration benchmark. Production adds 51 lines for this bounded diagnostic capability; no generic timing framework or queue change.

### DIFF
--- a/docs/gateway/logging.md
+++ b/docs/gateway/logging.md
@@ -51,6 +51,23 @@ openclaw logs --follow
 - Embedded-run `continue_normal` decisions log at `debug`; retry, profile-rotation, model-fallback, and error decisions remain warnings.
 - Trace logging also includes diagnostic timing summaries for selected hot paths, such as plugin tool factory preparation. See [/tools/plugin#slow-plugin-tool-setup](/tools/plugin#slow-plugin-tool-setup).
 
+### Slow cron list pages
+
+A cron list page taking at least one second emits `cron: slow list page` through
+its existing logger, subject to the file log level. The structured record names
+`operation: "cron.listPage"` and reports `elapsedMs`, `waitToCallbackMs`,
+`callbackMs`, and `completionDelayMs`, plus available source, matched, and returned
+row counts, the outcome, and emitter `pid`, `threadId`, and `isMainThread`. Fast
+pages emit no such record.
+
+These are wall times, not CPU time: waiting includes scheduling delays, callback
+time includes awaited work, and completion delay covers settlement after the
+callback finishes. Each source page is measured separately; caller visibility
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e695605b3d1de872c838edd4003f4eafa4eb8b02](https://github.com/openclaw/openclaw/pull/140994/commits/e695605b3d1de872c838edd4003f4eafa4eb8b02)
- Landed commit: [ba0ea7424c43785e3173f00a31db489559747079](https://github.com/openclaw/openclaw/commit/ba0ea7424c43785e3173f00a31db489559747079)

PR #141001, opened 2026-09-07
fix(tasks): return expired-page errors promptly during updates

steipete:
## What Problem This Solves

Fixes an issue where users paging through Tasks could wait for a full obsolete scan after concurrent task activity had already invalidated their pagination cursor.

## Why This Change Was Made

Check the carried registry revision immediately after each existing event-loop yield, before reading another batch. Cursorless scans retain their complete-attempt retry behavior, and the final revision and authorization checks remain in place.

An already-invalid cursor deliberately returns the existing restart-pagination error before an unrelated later batch can fail, matching stale-at-entry behavior. Valid cursors still surface those later errors.

## User Impact

Expired task pages return restart-pagination guidance sooner during concurrent task activity.

## Evidence

- Regression: the stale continuation examines 32 records instead of 1,024 and returns the same cursor error.
- All 14 query tests pass, including cursorless convergence, valid-cursor error propagation, combined invalidation/error priority, and existing 32/33/64/65-record yield boundaries. The original implementation fails the two new stale-cursor cases.
- Exact-head compiled Gateway proof passed in an isolated Linux environment: 23 ordinary request operations and 11 observed operations cover UI request pairs, valid pagination, cursorless convergence, and access-only cursor invalidation. The instrumented continuation examined 32 rows and skipped 992, returning the existing restart error. Both Gateway processes shut down normally with closed streams; source, build, runtime, and dependency seals remained stable.
- The full changed-file gate and exact-head CI pass. Independent review found no actionable issues.
- This proves reduced discarded scan work, not a whole-Gateway latency improvement or a cure for historical stalls. No provider workload was exercised.

### DIFF
--- a/src/tasks/task-registry-query.test.ts
+++ b/src/tasks/task-registry-query.test.ts
@@ -134,6 +134,94 @@ describe("listTaskRecordPage", () => {
     },
   );
 
+  it.each([
+    { name: "stale cursor", continuation: true, mutate: true, failLater: false },
+    { name: "cursorless retry", continuation: false, mutate: true, failLater: false },
+    {
+      name: "stale cursor before a later failure",
+      continuation: true,
+      mutate: true,
+      failLater: true,
+    },
+    {
+      name: "valid cursor with a later failure",
+      continuation: true,
+      mutate: false,
+      failLater: true,
+    },
+  ])("handles yielded task pages with $name", async ({ continuation, mutate, failLater }) => {
+    const tasks = Array.from({ length: 1_024 }, (_, index): TaskRecord => ({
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e721ba9e31475725ff5855f4237311df27e01034](https://github.com/openclaw/openclaw/pull/141001/commits/e721ba9e31475725ff5855f4237311df27e01034)
- Landed commit: [278c8dec6a37e829bd339e176390f0f61ee11241](https://github.com/openclaw/openclaw/commit/278c8dec6a37e829bd339e176390f0f61ee11241)

PR #139689, opened 2026-09-06
fix(sessions): prevent cleanup failures during other database writes

steipete:
Related: #139469

## What Problem This Solves

Fixes an issue where session cleanup fails or other session and dashboard writes stall when reclamation is waiting for commit authorization. The earlier repair covered transcript appends and replacements, but synchronous session-entry updates and first-use board transactions still blocked the parent thread on the worker's SQLite write lock.

## Why This Change Was Made

Move the existing cooperative admission into the shared `BEGIN IMMEDIATE` boundary. Every transaction writer using that boundary can service the pending worker request before retrying admission; transaction callbacks, mutations, and commit are never replayed. Remove the transcript-only wrapper and return its four callers to the canonical agent database writer.

The reclamation owner still captures its own async context, validates current authority once, and synchronously joins COMMIT, rollback, or connection-close settlement after granting approval. Registration uses the already-open connection's native database location, covering separate and reopened connections without filesystem freshness checks. Deferred reads and nested savepoints retain their existing paths, and reclamation without an authorization callback does not acquire an extra parent connection.

### Accepted design

The maintainer explicitly approved this writer-coordination repair, live testing, and landing before implementation. It extends the cooperative design in #139469 to the shared transaction owner. Schemas, persisted representations, retention, deletion atomicity, rollback, and commit-time authorization remain unchanged. No configuration option or dependency is added. Long deletion or COMMIT operations and unrelated external-lock contention remain subject to their existing costs; this does not claim a universal latency bound.

## User Impact

Session-entry and dashboard writes can finish without making the cleanup worker lose its commit checkpoint. The common admission owner also covers delivery records and first-use Goal/pending-input schema transactions. The production change is net +12 lines after removing the feature-specific path; the additional scope registration provides the shared ownership boundary. Tests add 40 lines.

## Evidence

- Failing-first on unchanged production: synchronous entry and cold board writers caused `SQLite session reclamation commit checkpoint expired`; two cases failed and seven controls passed.
- The repaired focused suite passed 39 tests, including grant/rejection, owner-context preservation, abrupt worker exits, mandatory commit settlement, native busy timeouts, savepoints, and rollback.
- Final-source unmocked Node 24 SQLite/Worker probes passed four separate/reopened-connection cases on both macOS and Linux. The parent entered a real synchronous write while the worker held its lock and the ordinary authorization callback remained queued. They verified exactly one body and authority check, durable commit/rollback results, and timeout restoration; Linux competing writes took 27–28 ms.
[...]

PR #139469, opened 2026-09-05
fix(sessions): prevent database lock failures during reclamation

steipete:
## Problem

Fixes an issue where concurrent Gateway runs fail with `database is locked` while session reclamation is committing. This was reported on the Team Gateway and reproduced with two synchronous transcript writers against an isolated SQLite store.

The reclamation worker holds `BEGIN IMMEDIATE` and waits for the parent's commit-authorization message handler. A synchronous transcript append on the parent waits in SQLite's native busy handler for that same writer lock, preventing the authorization handler from running. The existing settlement barrier cannot satisfy that event-loop dependency.

## Solution

Keep archive preparation, deletion, and COMMIT on the worker. Publish a distinct authorization-requested state in the existing shared memory. While reclamation is active, synchronous transcript writers use short admission attempts and service that pending request with the existing authority validator before retrying admission. The same admission path covers synchronous transcript replacement and compaction. Once the write transaction begins, its original busy timeout and diagnostics are restored; append callbacks and mutations are never replayed.

Reclamation performs a PASSIVE checkpoint and caps incremental vacuum at 512 pages per pass. Before pruning retained archives, the disk-budget owner drains its initially observed free pages in bounded units, yielding between them and reacquiring the connection after each yield. It retains physical checkpointing before pressure measurements so pending free pages do not cause unnecessary archive deletion. Duplicate history-deletion and initial sweep logic is consolidated while retaining existing snapshot and retention checks.

## Design

The parent registers an operation-scoped authorization service before starting the worker and removes it after the worker settles. The service captures the reclamation owner's async context: the admission checks depend on that context to distinguish the owner's own work from competing admissions. The synchronous append invokes the same live authority checks in that captured context, then grants or rejects the shared request.

There is no parent event-loop dependency between authorization and worker COMMIT. The existing synchronous settlement join remains: after granting approval, the parent must not allow queued owner retirement before COMMIT, rollback, or connection-close settlement, including abrupt worker exit. A later queued request returns the already recorded result or failure, so it cannot authorize twice. Authorization rejection belongs to reclamation; unrelated transcript writers can continue after rollback.

The budget drain is limited to its initial freelist, stops when SQLite makes no progress, and never holds a transaction across a yield. Budget measurements retain TRUNCATE checkpointing because PASSIVE alone leaves WAL allocation counted as disk pressure. Archive pruning starts only after this physical reclamation step; the existing archive-selection and retention rules are unchanged.

[...]

PR #141006, opened 2026-09-07
fix(sqlite): prevent cleanup timeouts during periodic vacuum

steipete:
Related: #139689, #139469

## What Problem This Solves

Fixes an issue where session deletion or cleanup could fail when periodic database maintenance ran while cleanup was awaiting commit approval. The maintenance callback could block the Gateway thread for the SQLite busy timeout, preventing the cleanup worker from receiving approval.

## Why This Change Was Made

Periodic incremental vacuum now uses the existing shared SQLite write-admission helper, which services pending cleanup approval before acquiring the writer lock. Passive checkpoints remain outside the transaction. This extends the existing admission design from #139689 while preserving the timeout policy, 512-page maintenance bound, and cleanup authorization and commit-settlement guarantees. Vacuum transaction diagnostics also identify the operation and database.

## User Impact

Background space reclamation can run alongside session cleanup without causing this approval timeout. Session cleanup still rejects a retired owner, and maintenance retains its existing page limit.

## Evidence

- Added a regression using the real periodic maintenance callback and a real reclamation worker, covering both accepted and revoked cleanup authority. Both cases failed on the original code with native SQLite error 5 (`database is locked`) and pass with this fix.
- The regression verifies that approval is serviced exactly once during maintenance, the expected cleanup outcome is preserved, and vacuum releases at most 512 pages.
- Focused SQLite WAL, transaction, reclamation, and commit-settlement suites: 94 passed; 7 Linux-only checks skipped on macOS.
- Updated the database reference to document the maintenance admission boundary.
- `node scripts/check-changed.mjs` passed, including core/test typechecking, lint, schema, and import-cycle guards.
- Documentation MDX validation passed; independent autoreview was clean through P2.

### DIFF
--- a/docs/reference/database-schemas.md
+++ b/docs/reference/database-schemas.md
@@ -415,6 +415,10 @@ The original lock-admission deadline is retained. After granting approval,
 the parent synchronously joins transaction settlement before allowing owner retirement;
 that mandatory join cannot be abandoned at the append deadline.
 
+Periodic incremental vacuum uses the same write-admission boundary, so it can
+service reclamation approval before taking the writer lock. Its 512-page limit
+is unchanged; passive checkpoints remain outside the write transaction.
+
 Reclamation page maintenance uses a PASSIVE checkpoint and at most 512 pages of
 incremental vacuum per pass. PASSIVE does not wait for readers, but does not cap
 the number of WAL frames copied. Before pruning retained archives, disk-budget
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f1c95894d6f81f909eec1f79743b836bcd262e8a](https://github.com/openclaw/openclaw/pull/141006/commits/f1c95894d6f81f909eec1f79743b836bcd262e8a)
- Landed commit: [e58b5bde5819ca825021de42712d50c228ef467a](https://github.com/openclaw/openclaw/commit/e58b5bde5819ca825021de42712d50c228ef467a)

PR #141015, opened 2026-09-07
refactor(scripts): use canonical bounded response readers

steipete:
## What Problem This Solves

Two diagnostic scripts retain private wrappers around the canonical bounded response reader. Claude's overflow tests call that adapter directly, so they can miss a request caller that accidentally skips bounded reading.

## Why This Change Was Made

Firecrawl comparison and Claude usage now call the shared reader with the same limits and signals. The two Claude overflow cases exercise the existing usage-request function with synthetic responses and an independent 256 KiB expectation; the test-only reader and constant exports disappear.

The former planner follow-up is already covered by merged #141137 and has been removed from this PR. The refreshed branch preserves main's planner fixture, discovery caches, and later assertions unchanged.

## User Impact

Response limits, timeouts, cancellation, returned fields, and displayed errors are unchanged. Raw overflow-error stacks now originate in the shared helper. Authentication, credential discovery, CLI behavior, and production planning policy are untouched. Tooling shrinks by 25 lines and tests by six; no test case is removed.

## Evidence

- Fresh recovery runs pass all 23 selected reader/usage cases before and after. All 52 other safety cases remain unchanged and unselected.
- A deliberate caller bypass still passes both original cap tests and fails both migrated tests because their promises resolve instead of rejecting. The control changes are removed and source hashes verified.
- The current complete planner file passes all 53 tests in original order. Both canonical cron browser cases also pass. These validate the main integration; the planner/UI repairs are not claimed as this PR's contribution.
- Fresh independent residual review is clean through P2, and the exact resulting tree preserves every other main entry.

The refreshed three-file configured changed gate passed all 19 checks on Linux, including both TypeScript graphs and scoped lint. Its one-shot runner stopped normally and the temporary source capsule was removed. The reported remote tree matches the signed candidate tree; a post-success portal synchronization warning is retained separately.

### DIFF
--- a/scripts/debug-claude-usage.ts
+++ b/scripts/debug-claude-usage.ts
@@ -8,7 +8,7 @@ import { pathToFileURL } from "node:url";
 import { expectDefined } from "../packages/normalization-core/src/expect.js";
 import { normalizeOptionalString } from "../packages/normalization-core/src/string-coerce.js";
 import { requireOptionArgument } from "./lib/arg-utils.mts";
-import { readBoundedResponseText as readBoundedResponseTextWithLimit } from "./lib/bounded-response.mjs";
+import { readBoundedResponseText } from "./lib/bounded-response.mjs";
 import {
   maskIdentifier,
   parseStrictIntegerOption,
@@ -177,17 +177,6 @@ const withFetchTimeout = async <T>(
   }
 };
 
-const readBoundedResponseText = (
-  response: Response,
-  label: string,
-  signal: AbortSignal,
-  maxBytes = FETCH_RESPONSE_MAX_BYTES,
-): Promise<string> =>
-  readBoundedResponseTextWithLimit(response, label, maxBytes, {
-    createTooLargeError: (message: string) => new Error(message),
-    signal,
-  });
-
 const fetchText = async (
   label: string,
   url: string,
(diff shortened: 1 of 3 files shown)


steipete:
The same inventory-growth timeout is now the sole failed test blocking [PR #141095, shared Git-ref mutation ownership](https://github.com/openclaw/openclaw/pull/141095).

[CI run 34112666847, compact-small-29](https://github.com/openclaw/openclaw/actions/runs/34112666847/job/101712460753) reports `keeps hosted tooling within the GitHub job cap when its inventory grows` timing out at 121,778 ms against its unchanged 120,000-ms deadline. It reports no job-cap assertion failure. The first logical group passed; the second reported 535 passes, 72 skips, and this single failure.

The tested merge was `ccbbe6e68ee58d2ebcf2baeae4766b3fab138978`, combining head `113f5b4ae5ba60b3b0baf7e2756be8d35ba674cd` with main `e3e2953f4b9a3ef1cd0568763c74c42f33067410`. The Git-ref patch does not change the planner, growth fixture, or unit-fast classifier. The fixture still resets modules and reimports the planner for all ten variants; this PR's retained real unit-fast namespace addresses that repeated setup without changing the assertions or timeout.

Recording the dependent failure here so we do not create a competing repair or retry unchanged code to hide it. This is additional failure evidence, not independent verification of this PR's reported performance improvement. The Git-ref landing remains blocked until the relevant repair and exact-head CI are verified.

PR #141031, opened 2026-09-07
feat(browser): stream the Control UI browser panel live instead of polling screenshots

steipete:
## What Problem This Solves

Resolves a problem where the Control UI Browser panel felt slow and stale: it showed the agent's browser as one-off screenshots, each refresh costing three gateway round trips (a screenshot written to the media store, an HTTP fetch of that PNG, and an evaluate call for viewport metrics), and it only refreshed after a user action, so page loads that finished a moment later never appeared.

An iframe is not a viable alternative: most sites block embedding with `X-Frame-Options` / `frame-ancestors`, and an iframe would be a separate browser with its own cookies rather than the agent's browser.

## Why This Change Was Made

The browser plugin now streams the active tab with Chrome's own `Page.startScreencast`. `POST /screencast` (via `browser.request`) validates the tab exactly like `/screenshot` and mints a one-time 60 s token; the panel opens a plugin-owned binary WebSocket at `/browser/screencast?token=…` and receives JPEG frames only when the page repaints, ack-paced to about 20 fps, with slow viewers skipping frames instead of queueing.

Navigation policy keeps parity with screenshots: each approved document gets its own CDP session. A main-frame navigation retires the capture synchronously and a new session starts only after the address passes the navigation policy, so frames of a rejected document can never reach a viewer (Chromium encodes frames asynchronously, so a plain stop/restart would not close that race). A rejected navigation closes the stream with `4003 navigation_blocked`; the panel shows the existing blocked-tab state. Sessions end when the page closes, the last viewer leaves, the profile lifecycle changes, or the gateway shuts down. Gateway-minted tickets and individual viewers also follow the requesting Gateway connection: disconnect or revocation deletes pending tickets and closes that requester’s viewers with `4006 authority_revoked`, without interrupting other live viewers sharing capture. Minting for an ended requester returns HTTP 401 / `SCREENCAST_REQUESTER_GONE`. Loopback HTTP control-API tickets retain the documented TTL-only boundary.

The screenshot path stays as an automatic fallback for node-routed browsers, Chrome MCP existing-session profiles, missing Playwright, and stream failures. Stream frames own the view while live, a late screenshot never overwrites a newer frame, and only failed attempts back off, so URL-bar navigation and resizes restart the stream immediately. Malformed mint replies also fall back to screenshots. Annotate and Inspect pin the capture image and URL while holding the latest incoming frame, which is published when capture mode ends. Node-side streaming for paired-node browsers is a follow-up.

## User Impact

The Browser panel now shows the agent's browser live: page loads, scrolling, and clicks render as they happen instead of after the next action, and idle pages cost no bandwidth. Existing interactions (click, type, scroll, annotate, inspect) work unchanged on top of the live frames. Docs: `docs/tools/browser.md` (panel behavior) and `docs/tools/browser-control.md` (route, wire format, close codes).

## Evidence

Live proof on an isolated gateway with the managed headless profile (Playwright-driven Control UI):

- Stream live on example.com over `ws://…/browser/screencast`; URL-bar navigation to iana.org followed via frames with **zero** `assistant-media` (screenshot) fetches for the whole session.
- A page-initiated navigation to a private loopback address closed the stream with `4003` and the tab showed the blocked state.

Browser panel streaming iana.org live after a URL-bar navigation

Blocked-tab state after a policy-rejected in-page navigation closed the stream

Tests and checks:

- Plugin: `node scripts/run-vitest.mjs extensions/browser/src/browser/screencast extensions/browser/src/browser/routes/agent.screencast.test.ts extensions/browser/src/gateway/browser-request.profile-from-body.test.ts extensions/browser/plugin-registration.shutdown.test.ts` → 77 passed, 1 skipped (chromium-gated).
- Chromium e2e (`OPENCLAW_BROWSER_SCREENCAST_E2E=1`, Google Chrome): JPEG `FF D8`, header CSS size equals `window.innerWidth/innerHeight`, page close → `4004 target_closed`.
- Regression `never forwards delayed blocked-document pixels after a later allowed navigation` fails on the pre-fix session module and passes after.
- UI: `node scripts/run-vitest.mjs ui/src/components/browser` → 139 passed.
- `node scripts/check-changed.mjs`: 28 lanes ok; the extension-lint lane cannot run inside a nested worktree (its declaration-boundary preflight resolves the parent checkout's `node_modules`), so it ran from a neutral checkout with the same diff → clean.
- Codex autoreview (P0/P1): scoped-clean after the session-rotation fix.

Review-fix validation for `1a46746cad7`:

- Focused browser/plugin, browser UI, and route-handoff e2e suites: 246 passed, with the Chromium-gated test run separately and passing. The standalone route-handoff rerun also passed all 4 tests.
- Regression coverage reproduced the missing requester revocation, capture pinning, and malformed-response behavior before the fixes. Requester tests cover pending-token deletion, ended-requester mint refusal, and independent viewer revocation while another viewer continues receiving frames. Live proof reported in the landing handoff covers the `4003 navigation_blocked` and `4006 authority_revoked` paths.
- The route-handoff mock now explicitly returns `SCREENCAST_UNSUPPORTED`; the client rejects malformed mint replies before opening a WebSocket.
- Changed-file checks passed except for the known nested-worktree declaration-boundary environment error. The affected extension lint passed in a separate byte-verified neutral checkout, and the remaining guards passed separately.
- Final Codex autoreview through P2: scoped-clean, with no actionable findings and no subsequent product edits.

These changes use in-memory session, token, and viewer state; they introduce no persistent schema or stored-data migration.

### DIFF
--- a/docs/tools/browser-control.md
+++ b/docs/tools/browser-control.md
@@ -23,7 +23,7 @@ agent tools, but nothing listens on the loopback control port.
 - Status/start/stop: `GET /`, `GET /doctor`, `POST /start`, `POST /stop`, `POST /reset-profile`
 - Profiles: `GET /profiles`, `POST /profiles/create`, `DELETE /profiles/:name`
 - Tabs: `GET /tabs`, `POST /tabs/open`, `POST /tabs/focus`, `DELETE /tabs/:targetId`, `POST /tabs/action`
-- Snapshot/screenshot: `GET /snapshot`, `POST /screenshot`
+- Snapshot/screenshot/stream: `GET /snapshot`, `POST /screenshot`, `POST /screencast`
 - Actions: `POST /navigate`, `POST /act`
 - Hooks: `POST /hooks/file-chooser`, `POST /hooks/dialog`
 - Downloads: `POST /download`, `POST /wait/download`
@@ -84,6 +84,53 @@ Notes:
   Tailscale Serve identity headers.
 - If `gateway.auth.mode` is `none` or `trusted-proxy`, these loopback browser
   routes do not inherit those identity-bearing modes; keep them loopback-only.
+
+### Screencast stream
+
+`POST /screencast` mints a single-use token for a live view of the selected tab.
+Pass an optional `targetId`, `maxWidth`, `maxHeight`, and `quality` in the JSON
+body. Dimensions default to 1280 and are clamped to integers from 320 to 2000;
+JPEG quality defaults to 70 and is clamped from 30 to 90.
+
+The response contains `token`, `wsPath`, `expiresAtMs`, `targetId`, and `url`.
+Resolve `wsPath` against the Gateway URL and open a WebSocket there:
+`/browser/screencast?token=<token>`. The 48-character hexadecimal token expires
+after 60 seconds and can be consumed once. Invalid, expired, and reused tokens
+are rejected with HTTP 401 before upgrade. Viewers send no application messages;
+binary viewer messages close the connection.
+
+The plugin shares one CDP screencast per profile and tab. Chrome sends JPEG
+frames on repaint, paced to approximately 20 frames per second. Slow viewers
+skip frames instead of building a queue. Navigation immediately retires the
+capture session. A new CDP session starts only after the address is allowed,
+so delayed frames from the previous document cannot enter the new stream.
+A rejected navigation stops the stream.
+
+Text messages are JSON with `type` in `ready`, `meta`, or `error`. A `ready`
+message includes `targetId`, `url`, and `title`; `meta` updates `url` and `title`
+after allowed navigation and page load. Binary messages contain:
+
+1. A four-byte unsigned big-endian JSON header length.
+2. That many bytes of UTF-8 JSON: `{ "url", "cssWidth", "cssHeight", "scrollX", "scrollY", "ts" }`.
+3. The JPEG bytes.
+
+`cssWidth` and `cssHeight` come from CDP's `deviceWidth` and `deviceHeight`
+metadata and describe the layout viewport in CSS pixels. `scrollX` and `scrollY`
+come from `scrollOffsetX` and `scrollOffsetY`; `ts` is the CDP frame timestamp.
+
+| Close code | Meaning                                                                   |
+| ---------- | ------------------------------------------------------------------------- |
+| 4001       | Token invalid or expired (normally rejected before upgrade with HTTP 401) |
+| 4003       | `navigation_blocked`                                                      |
+| 4004       | `target_closed`, including a profile lifecycle change                     |
+| 4005       | Unsupported streaming                                                     |
+| 1012       | Gateway shutting down                                                     |
+
+Chrome MCP existing-session profiles and missing Playwright return HTTP 501 with
+`code: "SCREENCAST_UNSUPPORTED"` and `reason: "existing-session"` or `"playwright"`.
+Node-routed requests fail before proxying with `INVALID_REQUEST` and details
+`{ "code": "SCREENCAST_UNSUPPORTED", "reason": "node" }`. The Control UI falls
+back to the existing screenshot route when streaming is unavailable.
 
 ### `/act` error contract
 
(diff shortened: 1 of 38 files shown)


steipete:
The P1 requester-authority finding in https://github.com/openclaw/openclaw/pull/141031#issuecomment-5567869650 is fixed in 1a46746cad7. Tickets and viewers are now bound to the requesting Gateway connection through `client.connectionSignal`. Unconsumed tokens are deleted when that connection aborts, and `POST /screencast` refuses an already-ended requester with HTTP 401 and `SCREENCAST_REQUESTER_GONE`. Established viewers close with `4006 authority_revoked`, while shared capture sessions continue serving other live viewers. Loopback HTTP control-API tickets remain TTL-only, as documented. The verified screencast suites and live proof cover pending-token revocation and established-viewer revocation, alongside the existing `4003 navigation_blocked` path.

The P2 annotation finding is also fixed in 1a46746cad7. While Annotate or Inspect is active, the stream holds incoming frames without rewriting the pinned view or its URL. On exit, it publishes the latest held frame, keeping annotation pixels and page context attached to the capture they describe.

The CI failure in `browser-route-handoff.e2e.test.ts` came from the mock Gateway answering `{}` for `/screencast`. The client now validates the mint response and falls back to screenshots for a malformed reply, and the fixture explicitly declares `SCREENCAST_UNSUPPORTED`. The route-handoff e2e and focused browser client, stream, controller, Gateway handler, route, and lifecycle suites pass with these fixes.

steipete:
The P1 finding, “Revoke screencast authority before the Gateway socket finishes closing,” in the second review at https://github.com/openclaw/openclaw/pull/141031#issuecomment-5567869650 is fixed in 6355548702e. Requester authority is now a live predicate (`client.invalidated !== true && !connectionSignal.aborted && hasCurrentClientAuthority() !== false`), evaluated at mint before and after awaited tab resolution, at ticket consumption, and before every frame or text send to each viewer. A revoked-but-not-yet-closed Gateway connection can neither redeem a pending ticket nor receive another frame; its viewer closes with `4006 authority_revoked`, while other authorized viewers of the shared capture keep streaming. The passing session regression test invalidates the requester while its socket close never fires: the pending ticket is refused with HTTP 401, frame sends cease, and a second authorized viewer remains unaffected. The Gateway handler test also verifies that `requester.isCurrent()` flips immediately when `client.invalidated` is set.

The P2 finding, “Keep the displayed image's URL until its replacement frame arrives,” is also fixed in 6355548702e. Metadata messages no longer rewrite `view`: the displayed URL and metrics belong to the displayed frame, while tab-strip and address-bar metadata update separately. The passing regression test delivers `meta` for the next document and then enters Annotate, verifying that annotation composition still uses the displayed frame and its own URL.

No storage or schema changes were made in this PR. Session, ticket, and viewer state are in memory; the review's data-model compatibility flag has no basis in the diff.

steipete:
The P2 finding, “Keep incoming frames from postponing viewport synchronization,” in the third review at https://github.com/openclaw/openclaw/pull/141031#issuecomment-5567869650 is fixed in 4a79e9548b9127e12697e588a08a64594f025dca. Mismatched frames now schedule one debounced viewport sync, and later frames join the pending sync instead of resetting its deadline. The pending flag clears when the sync runs or the stream is retired, so a persisting mismatch can schedule another sync afterwards.

The passing stream regression test delivers a frame every 100 ms for 600 ms and asserts that the remote resize still fires at 300 ms. It fails against the previous code, and the focused UI suites pass all 153 tests.

The previous CI run’s `check-lint` failure was main’s own max-lines budget on an untouched Telegram test. That was fixed on main by #141139, and this push rebases onto that fix. All four commits remain in their original order, and the rebase is patch-identical.

steipete:
Merged via squash.

- Prepared head SHA: [4a79e9548b9127e12697e588a08a64594f025dca](https://github.com/openclaw/openclaw/pull/141031/commits/4a79e9548b9127e12697e588a08a64594f025dca)
- Landed commit: [a7175b2af3aa297143aa62117644d03d04df443a](https://github.com/openclaw/openclaw/commit/a7175b2af3aa297143aa62117644d03d04df443a)

PR #141039, opened 2026-09-07
fix: restore local password auth for trusted-proxy gateway status

steipete:
## What Problem This Solves

Fixes an issue where `openclaw gateway status` reports an authentication failure for a healthy local Gateway using trusted-proxy authentication with an existing local password. Managed update verification can consequently time out and roll back even while normal Gateway RPC calls succeed.

## Why This Change Was Made

Status now lets the existing credential resolver load the service's local password, including file and environment SecretRefs. The status-specific shortcut introduced in #135862 incorrectly grouped trusted-proxy with auth mode `none`, discarding the documented password path for same-host clients.

The change removes that shortcut for trusted-proxy. Service-derived target isolation, explicit-URL credential rules, read-only diagnostics, and Gateway authorization policies remain intact.

## User Impact

Operators can verify their trusted-proxy Gateway through its existing local password without copying credentials into command arguments or changing authentication policy.

## Evidence

- Before the production fix, the trusted-proxy environment SecretRef regression failed with an undefined password; both real WebSocket probes using a synthetic file SecretRef failed with close code 1008 (`password required`). The neighboring password/unspecified-mode cases passed.
- After the fix, 136 tests passed across `status.gather.test.ts`, `probe.test.ts`, and `probe.reachability.test.ts`. These include default connectivity and `--require-rpc` wire behavior, existing TLS/explicit-target isolation, no pairing writes, and an isolated real Gateway authentication-rejection/recovery test.
- The synthetic file credential is stored with mode 0600, delivered to the selected loopback peer, and absent from serialized status output. The fixture exercises actual status gathering, SecretRef resolution, client connection, and read RPC; it does not simulate a proxy identity.
- The existing local-direct password contract is documented in `docs/gateway/trusted-proxy-auth.md` and enforced by `src/gateway/auth.ts`. No configuration or server-policy change is included.

### DIFF
--- a/src/cli/daemon-cli/status.gather.test.ts
+++ b/src/cli/daemon-cli/status.gather.test.ts
@@ -1622,29 +1622,113 @@ describe("gatherDaemonStatus", () => {
     );
   });
 
-  it("resolves daemon gateway auth password SecretRef values before probing", async () => {
-    daemonLoadedConfig = {
-      gateway: {
-        bind: "lan",
-        tls: { enabled: true },
-        auth: {
-          password: { source: "env", provider: "default", id: "DAEMON_GATEWAY_PASSWORD" },
+  it.each([undefined, "password", "trusted-proxy"] as const)(
+    "resolves daemon gateway auth password SecretRef values before probing in %s mode",
+    async (mode) => {
+      daemonLoadedConfig = {
+        gateway: {
+          bind: "lan",
+          tls: { enabled: true },
+          auth: {
+            mode,
+            password: { source: "env", provider: "default", id: "DAEMON_GATEWAY_PASSWORD" },
+          },
         },
-      },
-      secrets: {
-        providers: {
-          default: { source: "env" },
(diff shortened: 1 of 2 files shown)


steipete:
Closing as superseded by #140975, which landed while this independent reproduction and review were running.

Verified that [b2d7fe9fe332c9f6c043bc00ec3f3ec07ddce0f5](https://github.com/openclaw/openclaw/commit/b2d7fe9fe332c9f6c043bc00ec3f3ec07ddce0f5) is on current main and restores the same canonical password-resolution branch. Its implementation differs from this production patch only by an explanatory comment; it also covers configured/service-environment passwords, SecretRefs, and disabled exec resolution.

The independent proof here captured the failing environment-SecretRef case and both actual WebSocket file-SecretRef modes before the fix, then passed 136 focused/sibling tests with the equivalent producer correction. Independent review through P2 found no actionable findings. No additional production change is needed from this PR.

PR #141041, opened 2026-09-07
fix(ai): honor compat.sendSessionIdHeader for OpenAI Responses proxies

### LINKED ISSUE #140918: OpenAI Responses proxy: missing session affinity header causes repeated prompt-cache misses
## Proxy case still affected on 2026.9.2

The native-host restriction in #100233 leaves the same cache-affinity problem on an OpenAI Responses proxy that forwards to the ChatGPT Codex backend. Related work #90033 concerns Chat Completions and does not establish a fix for this managed Responses path.

Environment: OpenClaw 2026.9.2, built-in `openclaw` runtime, custom provider using `openai-responses`, loopback OpenCodex 2.44.0 proxy, GPT-6 Astra with medium reasoning. Proxy uses its OpenAI Responses forward adapter and a Codex OAuth account pool. All sampled requests resolve to the same account label.

`compat.supportsPromptCacheKey: true` is necessary to preserve the body cache key for this proxy. After setting it, a local diagnostic relay observed a stable session-derived `prompt_cache_key` on every request, an identical first input item hash, but **no `session_id` header**.

A three-request sequential tool loop then reported:

| Call | Total input tokens | Cached input tokens |
|---|---:|---:|
| 1 | 17955 | 0 |
| 2 | 18001 | 0 |
| 3 | 18039 | 0 |

Diagnostic control: a temporary loopback relay forwarded the same request shape and added only `session_id` from that request's own session-derived `prompt_cache_key`. A four-request loop reported:

| Call | Total input tokens | Cached input tokens |
|---|---:|---:|
| 1 | 17980 | 0 |
| 2 | 18025 | 17792 |
| 3 | 18062 | 17920 |
| 4 | 18100 | 17920 |

These are separate short live runs, not a large statistical sample. The supported Gateway route still misses; the control relay is not deployed as a workaround. No OpenClaw package source was modified.

The installed `buildOpenAIClientHeaders` in `packages/ai` emits `session_id` only under `usesNativeOpenAICodexResponsesBackend(model)`, which excludes this proxy. Although `sendSessionIdHeader` is in the config schema, this boundary-aware implementation does not consult it.

Expected: honor the existing `compat.sendSessionIdHeader: true` option in the managed Responses header builder. Derive the header from the real session identity; preserve unconfigured endpoint behavior, independent-session isolation, case-insensitive caller header precedence, and cache-disable semantics. No new configuration option, unconditional proxy headers, or static shared session ID is needed.

Reproduction: configure a loopback OpenAI Responses forward proxy targeting the Codex backend; enable `supportsPromptCacheKey`; invoke an isolated `openclaw agent exec` on that provider with sequential read-only shell calls; capture only the presence/value identity of the cache key and session header, then compare provider-reported input/cache counters. Repeat with a diagnostic relay injecting the per-session header. Exact private prompts/config/credentials are intentionally omitted.

## Bot review follow-up

[ClawSweeper review](https://github.com/openclaw/openclaw/issues/140918#issuecomment-5566040815) independently confirmed the header defect on current main and kept this issue open for maintainer follow-up. It did not rerun the live cache-performance comparison. The expected behavior above now names the existing option and the preservation requirements identified by that review. After-fix acceptance still requires a normal Gateway run through the compatible proxy, without the diagnostic relay.

gaoanze888:
## What Problem This Solves

Closes #140918.

OpenAI Responses proxies that forward to the ChatGPT Codex backend never received a `session_id` header, so the backend could not correlate requests to a session and prompt-cache reads missed on every call. A three-request tool loop reported 0 cached input tokens across all calls.

**Root cause:** `buildOpenAIClientHeaders` only emitted `session_id` when `usesNativeOpenAICodexResponsesBackend(model)` was true (i.e. `baseUrl` is the native `chatgpt.com/backend-api/codex`). The proxy path — a custom provider using `openai-responses` with a non-native `baseUrl` — was excluded even though the config schema has a `compat.sendSessionIdHeader` option specifically for this case. The boundary-aware header builder never consulted it.

## Evidence

**The fix** (in `packages/ai/src/transports/openai-transport-params.ts`): extend the session_id guard to also fire when the model explicitly opts in via `compat.sendSessionIdHeader === true`, while preserving the prior behavior for unconfigured endpoints (native backend only):

```ts
// Before
if (
  sessionId &&
  !Object.keys(resolvedHeaders).some(
    (key) => normalizeLowercaseStringOrEmpty(key) === "session_id",
  ) &&
  usesNativeOpenAICodexResponsesBackend(model)
) {

// After
if (
  sessionId &&
  !Object.keys(resolvedHeaders).some(
    (key) => normalizeLowercaseStringOrEmpty(key) === "session_id",
  ) &&
  (usesNativeOpenAICodexResponsesBackend(model) ||
    model.compat?.sendSessionIdHeader === true)
) {
```

**Preserved behavior:**
- Native Codex backend still always emits `session_id` (unchanged).
- Unconfigured proxy endpoints (no `compat.sendSessionIdHeader`) still do **not** emit `session_id` (unchanged).
- Caller-wins case-insensitive header precedence is preserved (the `some()` guard still prevents overwriting a caller-supplied `session_id`).
- Cache-disable semantics and independent-session isolation are preserved (the header derives from the real per-session `sessionId`).

**Tests** (`packages/ai/src/transports/openai-transport-params.session-header.test.ts`): 3 new regression tests added, 9/9 pass:

| Test | Expected | Result |
|---|---|---|
| Non-native proxy with `compat.sendSessionIdHeader: true` | emits `session_id` | ✓ pass |
| Non-native proxy with `sendSessionIdHeader` unset | does **not** emit `session_id` | ✓ pass |
| Native endpoint with `compat.sendSessionIdHeader: false` | does **not** emit `session_id` | ✓ pass |

**No regression:** 75 OpenAI Responses provider tests pass (openai-responses, openai-chatgpt-responses, protocol, cache).

**Local verification:**
- `npx vitest run packages/ai/src/transports/openai-transport-params.session-header.test.ts` → 9/9 pass
- `npx vitest run` (openai-responses + chatgpt-responses provider tests) → 75/75 pass


er-s-an:
Thanks for taking this on. I was independently investigating #140918 and have stopped preparing a separate PR to avoid overlap.

I reproduced the missing-header case on main and have a local regression-tested patch that also threads `cacheRetention` through the managed Responses executor/client to the header builder. The focused header and managed SSE suites pass 37 tests locally, including `cacheRetention: "none"` suppressing the generated proxy header while preserving the transport's session identity. This is local boundary coverage, not live backend cache-performance proof.

Would you welcome the focused cache-disabled regression case or a small follow-up patch on your branch? Happy to coordinate here.

PR #141046, opened 2026-09-07
fix(cli): keep shared credentials for trusted-proxy status probes

steipete:
Related: #135862

## What Problem This Solves

Fixes an issue where operators running `openclaw gateway status --json` against a trusted-proxy Gateway with a shared loopback password would receive `rpc.ok: false` and `AUTH_VERIFIED_USER_REQUIRED`, even though the same Gateway accepted `gateway call system.info`.

The regression started with 34a6d38b187 (#135862). In the reported production deployment, candidate verification required `rpc.ok === true` and `rpc.server.buildId`; every hourly deployment failed verification and rolled back, causing two Gateway interruptions per hour.

## Why This Change Was Made

Only `gateway.auth.mode="none"` now short-circuits to empty probe credentials. Trusted-proxy status probes use the existing local credential resolver, including password SecretRefs and the exec-SecretRef gate, because they reach the loopback listener directly and the proxy cannot supply a verified identity.

Service-derived targets still use their dedicated target path. The sanitized probe config still strips credentials, `skipImplicitAuth` remains enabled, and only the resolved explicit credentials reach the probe. Explicit `--token`/`--password` keep precedence, and explicit `--url` targets consume only explicitly supplied credentials. Existing trusted-proxy password-only fallback semantics are unchanged.

## User Impact

Local status checks can authenticate to trusted-proxy Gateways using the configured shared password again, restoring candidate verification and build identity reporting. Trusted-proxy installations without shared credentials and auth-none installations still send no configured credentials. The CLI reference now explains the direct-loopback password behavior.

## Evidence

- Regression reproduced before the production edit: the new plaintext and env-provider SecretRef password cases received `undefined` instead of the configured password. The disabled exec-SecretRef case also failed its existing diagnostic contract. Result: 3 failed, 109 passed.
- After the fix: `node scripts/run-vitest.mjs run src/cli/daemon-cli/status.gather.test.ts` passed all 112 tests.
- `node scripts/run-vitest.mjs run src/cli/daemon-cli src/gateway/call.test.ts src/gateway/probe-auth.test.ts`: 29 files passed; 779 tests passed, 1 skipped (Windows-only home-path casing), across 3 routed suites. The run used a separate filesystem module cache while the changed-file gate ran.
- `pnpm check:changed` passed, including production/test typechecks, changed-file lint, dead-export scans, and repository guards.
- `pnpm docs:list` and `git diff --check` passed.
- Source tracing confirmed `probeGatewayStatus`, `callGateway`, and client bootstrap retain resolved explicit credentials for both probe modes. Existing service-target assertions remain intact.
- Production symptoms above are operator-reported; validation uses synthetic local fixtures and does not restart or modify a live Gateway. Autoreview is reserved for the lead per the work order.


steipete:
Superseded by #140975 (`b2d7fe9fe33`, landed on main at 08:37 UTC), which restores the same behavior: only `gateway.auth.mode: "none"` skips probe credentials and trusted-proxy probes resolve the configured local password. This branch was cut before that commit reached main; closing as a duplicate.

PR #141058, opened 2026-09-07
fix(pr): keep trusted wrappers working when package aliases change

steipete:
## What Problem This Solves

Resolves a problem where native PR workflows using a materialized trusted wrapper can lose their tooling dependencies when the canonical checkout's top-level package aliases disappear or change. The hosted-gates verifier can then fail with `ERR_MODULE_NOT_FOUND` for `yaml`, despite the dependency already being included in the wrapper's package list.

## Why This Change Was Made

Resolve and validate the four selective third-party package directories before creating the materialized anchor's dependency links. Link to those real installed directories instead of mutable top-level `node_modules` aliases. If a required directory is unavailable, report the missing dependency and recovery guidance, remove the incomplete anchor, and stop before handoff or lock acquisition without installing anything.

This preserves source-blob verification, trusted-wrapper selection, selective dependency access, Git locks and retained outcomes, and exact-head CI enforcement. It does not link the whole workspace, add a dependency, or change gate policy. The underlying package installation must remain intact: pinning directory paths does not copy or freeze package contents.

## User Impact

Maintainers can continue using an already-materialized wrapper when canonical package aliases are removed or redirected. Incomplete installations fail early with actionable guidance instead of producing a late helper-import failure. There is no change to the packaged OpenClaw runtime.

## Evidence

- Before the repair, the regression reproduced `ERR_MODULE_NOT_FOUND` for `yaml` imported by the actual materialized `verify-pr-hosted-gates.mts` owner after removing the synthetic canonical alias. Redirecting that alias also failed. Both scenarios now pass in fresh Node processes using real installed dependencies.
- Missing-dependency cases cover all four tooling packages and verify no dependency installation, PR refs, or incomplete anchor directory is left behind.
- 170 tests passed across the wrapper, hosted-verifier, and native hosted-merge handoff suites, including existing anchor-integrity and exact-head gate enforcement coverage.
- Scoped changed checks, shell syntax, and `git diff --check` passed.

```bash
node scripts/run-vitest.mjs test/scripts/pr-wrappers.test.ts test/scripts/verify-pr-hosted-gates.test.ts test/scripts/pr-merge-hosted.test.ts
```

```bash
node scripts/check-changed.mjs -- scripts/pr scripts/AGENTS.md test/scripts/pr-wrappers.test.ts
```

The regression proves the dependency-delivery failure mode; it does not establish the exact timing of the original operator incident's alias state. The alias-lifetime scenario is POSIX-only; the production path preserves the existing Windows junction behavior.


steipete:
Merged via squash.

- Prepared head SHA: [4c3bbc2a5f7c662caf7cf7574a8a8c2f5528c324](https://github.com/openclaw/openclaw/pull/141058/commits/4c3bbc2a5f7c662caf7cf7574a8a8c2f5528c324)
- Landed commit: [daebe66972eb15a24a62df1d7d2a304b25d75300](https://github.com/openclaw/openclaw/commit/daebe66972eb15a24a62df1d7d2a304b25d75300)

PR #141003, opened 2026-09-07
feat(ios): start live voice from Siri and Shortcuts

steipete:
Closes #140998
Related: #47584 (broader background Siri question-and-answer support is not part of this change).

## What Problem This Solves

Starting live voice on iOS required opening OpenClaw, finding Chat, and tapping Talk. There was no native action for starting the existing voice conversation from Shortcuts, Siri, the Action Button, or a Home Screen shortcut.

## Why This Change Was Made

Adds a parameterless **Start Live Voice** App Shortcut with the Siri phrase **“Start live voice with OpenClaw.”** It foregrounds the app, preserves the selected chat, and uses the existing Talk owner for Gateway admission, microphone permissions, audio arbitration, and connection recovery. App Intents metadata generation is enabled for the iPhone app so the action is actually discoverable.

Early launch requests are coalesced and wait for the active root UI and initial session selection. Repeated runs enable rather than toggle or restart Talk. Setup rejection consumes the request so completing onboarding cannot unexpectedly start an old microphone request. Immediate failures use the existing dismissible notice banner; simulator testing caught an alert/onboarding-dismissal collision and the banner avoids that modal race.

No new voice backend, public URL route, Gateway option, intent extension, or background Siri conversation path is introduced.

## User Impact

Users can choose **Shortcuts → Apps → OpenClaw → Start Live Voice**, invoke the Siri phrase, or assign a saved shortcut to the Action Button or Home Screen. The existing Gateway setup and iOS foreground, unlock, and permission requirements remain in place. Usage and limitations are documented on the iOS platform page.

## Evidence

- Full iOS Simulator app build passed, including the embedded Watch companion and App Intents extraction.
- **24 focused tests passed; 0 failures and 0 skipped** across `OpenClawAppDelegateTests` and `OpenClawTypographyTests`. Coverage includes cold-launch buffering, inactive-scene admission, selected-session preservation, repeated invocation, setup rejection without later replay, and canonical capture rejection.
- `node scripts/check-changed.mjs`, Swift formatting/lint, and `git diff --check` passed.
- `pnpm native:i18n:verify` passed after regenerating the two new source inventory entries. The first CI attempt caught the missing generated catalog; no runtime change or translated locale update was needed.
- Fresh independent pre-commit review found no actionable P0–P2 findings.
[...]

PR #141069, opened 2026-09-07
fix: node battery level reads as 1% at full charge because the 0.0-1.0 fraction is undocumented

### LINKED ISSUE #141066: [Bug]: device.status battery 'level' is an undocumented 0.0–1.0 fraction — full charge reads as '1%'
## Summary
`device.status` reports `battery.level` as a normalized 0.0–1.0 **fraction**, but the unit is undocumented in the payload struct, node docs, and tests. Every consumer — human or agent — naturally reads `level: 1` as "1%". **Live-observed:** an OpenClaw agent queried a paired Samsung SM-S938U1 at ~100% charge and reported to its operator that the phone was at "1%"; the operator's verbatim reaction was "1% isn't even right, something else seems broke." Full transcript of the interpretation failure in [this comment](https://github.com/openclaw/openclaw/issues/141066#issuecomment-5568155800) (attached below the review).

## Environment
- Android app 2026.7.4 (Play release), Samsung SM-S938U1, Android 16
- Affects `main` and latest release v2026.9.2 (source verified 2026-09-07); iOS has the identical behavior

## Repro
1. Pair any Android or iOS node
2. Invoke `device.status` while the battery is above ~1% (e.g. 95%)
3. Observe `battery.level: 0.95` (or `1.0` at full charge) with no unit indicated anywhere

## Expected
Either the fraction contract is documented unambiguously, or the payload exposes an explicit percentage.

## Root cause
`apps/android/app/src/main/java/ai/openclaw/app/node/DeviceHandler.kt`:
```kotlin
private fun batteryLevelFraction(intent: Intent?): Double? {
    val rawLevel = intent?.getIntExtra(BatteryManager.EXTRA_LEVEL, -1) ?: -1
    val rawScale = intent?.getIntExtra(BatteryManager.EXTRA_SCALE, -1) ?: -1
    ...
    return rawLevel.toDouble() / rawScale.toDouble()   // 0.0–1.0
}
...
battery.levelFraction?.let { put("level", JsonPrimitive(it)) }  // emitted raw
```

`apps/ios/Sources/Device/DeviceStatusService.swift`:
```swift
let level = device.batteryLevel >= 0 ? Double(device.batteryLevel) : nil  // UIDevice.batteryLevel is 0.0–1.0
```

`apps/shared/OpenClawKit/Sources/OpenClawKit/DeviceCommands.swift`:
```swift
public struct OpenClawBatteryStatusPayload: Codable, Sendable, Equatable {
    public var level: Double?   // no unit documented
```

Both platforms agree on the fraction, and `DeviceHandlerTest.handleDeviceStatus_returnsExpectedShape` asserts `level in 0.0..1.0` — so the fraction is the intended cross-platform contract, but nothing tells consumers that. Gateway `src/` has no battery consumers, so the raw fraction reaches agents/CLIs unannotated.

## Impact
Any automation or agent keyed on battery level (low-battery alerts, charge-management routines, device checks) receives a value that reads 100× off from intuition, with nothing in the payload to signal the unit. A phone at 95% charge reports `level: 0.95` → read as "0.95% ≈ dead"; at 100% it reports `level: 1` → read as "1%".

## Proposed fix (non-breaking)
1. Document `level` as "normalized fraction of full charge, 0.0–1.0" on `OpenClawBatteryStatusPayload`
2. Add optional `levelPercent: Int?` (0–100, rounded) populated by both apps
3. Assert fraction + percent semantics in the Android Robolectric test

Alternative (breaking, maintainer decision): rename to `levelFraction` or switch to percent — both break the existing payload shape.

Fix in progress: #141069

NullArbitrage:
A fully charged device reports `battery.level: 1` because `level` is a fraction from 0 to 1. Issue #141066 records that value being reported to the operator as 1%.

This change preserves `level` and adds optional integer `levelPercent`, rounded from the same captured level, to Android, iOS, and watchOS `device.status` responses. Unknown readings still omit both numeric fields. Battery state, low-power mode, monitoring ownership, and command routing stay unchanged. Older consumers can keep using `level`.

Validation:

- Retained contributor evidence from September 7 shows the same paired Android phone returning `{"level":1,"state":"full","lowPowerModeEnabled":false}` before the change and `{"level":1,"levelPercent":100,"state":"full","lowPowerModeEnabled":false}` afterward. The reported operating-system charge was 100%; the retained build screenshot also shows 100% in the status bar. The four changed production files match that tested build. This is retained contributor evidence, not a new physical-device run.
- All 13 Android handler tests pass, including fractional values, nearest rounding, zero/full charge, and unavailable readings. With the pinned main battery producer, the same tests fail only because `levelPercent` is missing. Kotlin lint passes.
- Shared Swift tests cover older payloads, integer percentage encoding, unavailable omission, and older consumers. Hosted CI covers these tests and the Android, iOS, and Watch build paths. Physical iOS/watchOS execution is not claimed.

The explicit field clarifies the unit; it does not guarantee how every consumer or model will interpret the result.

Fixes #141066.


NullArbitrage:
**Note on the `Scan iOS dead code` failure at `60ba01d`:** this finding is inherited from `main`, not introduced by this PR — and it's now fixed on the branch head.

- The single Periphery finding was `StartLiveVoiceIntent.swift:5` (static `description`), introduced on `main` by #141003 (merged 2026-09-07 01:50 PT — after this workflow's 57 most recent green runs on other branches). This branch inherited it via the update-branch merge.
- The `description` property is AppIntents framework metadata — consumed by the system at runtime for the Siri/Shortcuts presentation — so no static reference exists and Periphery flags it. The workflow is zero-tolerance (any finding fails the job), so every branch merging current `main` fails this lane until the finding is annotated.
- Fixed in `0ec9eab` with the repo's standard `// periphery:ignore` annotation (same convention as `RealtimeTalkRelaySession.swift`), with an inline comment explaining the framework-consumed nature of the property.
- This PR's own additions are not implicated: `OpenClawBatteryStatusPayload` is covered by `retain_codable_properties: true` in `apps/ios/.periphery.yml`, and the scan report contained exactly one finding (the `StartLiveVoiceIntent` one above), verified in the run artifact.

Happy to split the annotation into a separate PR if maintainers prefer to keep main's regression out of this diff — it's one line and this branch simply happened to be the first to merge `main` after #141003 landed.

NullArbitrage:
@clawsweeper re-review

obviyus:
landpr-approve-unrelated-red
pr: 141069
head: d5ec6732d0fcee3044b062e622cc208af416137e
run: 34412522293 attempt: 1
failed: android-test-play job=102670141635 conclusion=failure
check: openclaw/ci-gate

obviyus:
Thanks @NullArbitrage! Landed in 117b2b032cfe188782fd98a50909341425804bb2.

**Bug:** A full battery returns fractional `level: 1`, which the reported agent interpreted as 1%.

**Fix:** Android, iOS, and watchOS add optional integer `levelPercent` while preserving the fractional value and unavailable readings.

| Before | After |
| --- | --- |
| `{"level":1,"state":"full","lowPowerModeEnabled":false}` | `{"level":1,"levelPercent":100,"state":"full","lowPowerModeEnabled":false}` |

**Proof:** The table is retained contributor-reported output from the same paired Android phone on September 7. The production files remain identical to that tested build. All 13 focused Android handler tests passed; the same tests with the pinned main battery producer failed only for missing `levelPercent`. All three shared Swift compatibility tests passed. Exact-head source review passed. Hosted iOS/Watch builds, Android third-party and Wear tests, and Kotlin lint passed. Physical Apple execution is not claimed.

**Unrelated CI exception:** [Run 34412522293](https://github.com/openclaw/openclaw/actions/runs/34412522293), attempt 1, had one failure among 3,115 Android Play tests: [job 102670141635](https://github.com/openclaw/openclaw/actions/runs/34412522293/job/102670141635), `NodeForegroundServiceTest.stopDuringRuntimeConstructionRetiresBackgroundStartup`, expected `Offline` but observed `Connecting…`. The [aggregate gate](https://github.com/openclaw/openclaw/actions/runs/34412522293/job/102675828956) failed because of that child. The guarded merge used this [standing-approval record](https://github.com/openclaw/openclaw/pull/141069#issuecomment-5609953077).

The tested integration was `58e25896862db16898ae6b37e8666f960e4f8e10`, with main parent `e4a6d50073ceb4e97f5401e19d40ec3c02e0ff22`. The failing test and its service/runtime/session owners are unchanged from that parent. This PR's battery serialization is not invoked by the fixture, and its battery tests do not alter connection state, preferences, or dispatchers. The existing stop path publishes `Offline` before asynchronous socket cleanup; an admitted connection callback can publish `Connecting…` before cleanup completes. The test joins the service scope, not those callback tails ([runtime stop](https://github.com/openclaw/openclaw/blob/58e25896862db16898ae6b37e8666f960e4f8e10/apps/android/app/src/main/java/ai/openclaw/app/NodeRuntime.kt#L5169), [session cleanup](https://github.com/openclaw/openclaw/blob/58e25896862db16898ae6b37e8666f960e4f8e10/apps/android/app/src/main/java/ai/openclaw/app/gateway/GatewaySession.kt#L523), [callback](https://github.com/openclaw/openclaw/blob/58e25896862db16898ae6b37e8666f960e4f8e10/apps/android/app/src/main/java/ai/openclaw/app/gateway/GatewaySession.kt#L2091)).

This establishes causal separation from the battery change, not the exact callback order in this run. Sampled failed main runs skipped Android, so no matching main failure or first-introducing commit is claimed. Synchronizing the existing stop assertion with callback retirement remains a separate follow-up; no sleep, weaker assertion, or unrelated code change was added here.

PR #141070, opened 2026-09-07
fix(tui): preserve drafts after picker selections

steipete:
## What Problem This Solves

Fixes an issue where users choosing a model in the TUI could submit the same selection twice or lose their next draft when the backend update was slow. The picker stayed focused: repeated Enter sent another model update, and subsequent typing went into its search filter.

## Why This Change Was Made

Consume the existing picker request and close its exact overlay when an item is selected, before awaiting the backend. Model, agent, session, and context pickers share this owner. Existing session-incarnation checks and asynchronous error rendering remain; Settings continues to support multiple changes while open.

## User Impact

Users can start composing their next message immediately after choosing an item. Repeated Enter cannot resubmit the consumed selection, and delayed completion cannot close a newer picker.

## Evidence

The new regression failed on the original owner: repeated Enter caused two model updates and the next draft appeared in search. It passes with this change: one update, draft retained in the editor and subsequently sent.

- 304 focused command-handler, overlay, searchable-selector, and filterable-selector tests passed.
- The complete picker PTY file passed **9/9** on Linux x86_64, Node 24.19.0 and pnpm 12.3.4, in 18.31 seconds with original assertions/timeouts. The candidate and wrapper exited 0; the task-owned Blacksmith Testbox was stopped. [Testbox workflow run](https://github.com/openclaw/openclaw/actions/runs/34085243612).
- The canonical plugins-platform test type graph, full-file lint/format, and independent P0–P2 review passed.

The Linux source capsule verified base `9098b23255ab11743a5e9030d1291710e9e5679b` and candidate tree `4d03c2225592fcd46658700e3881033edc50b582`; all four reviewed file hashes match commit `62996a81145582d3011ae41ef641307b9fa617e9`. All four base files remain identical on captured main `58ad5780b53c2f793aef8641ba27f4095aacc717`. This is real TUI/native PTY proof with a synthetic backend, not live Gateway or provider validation.

Earlier macOS runs hit three startup/exit deadlines in existing cancellation cases. Their cause remains unestablished; all nine tests subsequently passed in the clean Linux run. Production change is +3 lines; tests and fixture support add 82 lines.

These inspected visuals faithfully render the captured PTY text rows with simplified styling; they are not desktop screenshots.

Before: the next draft is trapped in the old picker filter.

Before: draft remains in the model picker search

After: the next draft belongs to the editor while the backend update is still pending.

After: draft remains in the editor

### DIFF
--- a/src/tui/tui-command-handlers.test.ts
+++ b/src/tui/tui-command-handlers.test.ts
@@ -432,6 +432,36 @@ describe("tui command handlers", () => {
   });
 
   it.each([
+    { command: "/models", value: "fixture/model" },
+    { command: "/sessions", value: "agent:main:other" },
+    { command: "/agents", value: "other" },
+  ])(
+    "consumes $command selection before its asynchronous action finishes",
+    async ({ command, value }) => {
+      const pending = createDeferred();
+      const action = vi.fn(() => pending.promise);
+      const harness = createHarness({
+        listModels: vi.fn().mockResolvedValue([{ provider: "fixture", id: "model" }]),
+        listSessions: vi.fn().mockResolvedValue({ sessions: [{ key: "agent:main:other" }] }),
+        agents: [{ id: "main" }, { id: "other" }],
+        patchSession: action,
+        setSession: action,
+      });
+      await harness.handleCommand(command);
+      const selector = firstMockArg(harness.openOverlay, "openOverlay") as SelectableOverlay;
+
+      selector.onSelect?.({ value });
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [62996a81145582d3011ae41ef641307b9fa617e9](https://github.com/openclaw/openclaw/pull/141070/commits/62996a81145582d3011ae41ef641307b9fa617e9)
- Landed commit: [9fa180758af681d309a8dada4246847832ab6df5](https://github.com/openclaw/openclaw/commit/9fa180758af681d309a8dada4246847832ab6df5)

PR #141084, opened 2026-09-07
fix(telemetry): retain update results when state writes fail

vincentkoc:
## What Problem This Solves

Fixes an issue where operators would lose a valid update notice and send repeated update reports when the local state database rejected a write. A successful HTTP response was treated as a failed request if persisting it failed, allowing another request after the one-minute failure backoff instead of the normal daily interval.

## Why This Change Was Made

Keep an accepted response and its original timestamp in a bounded in-process fallback when persistence fails. Scope pending successes by endpoint and resolved database path, pin the write destination before HTTP, and retry the existing state write on a later ordinary call without another request or timestamp extension. The existing atomic machine-state update helper preserves an equal or newer durable success inside its transaction.

This retains at most 32 pending contexts and adds no durable store, identifier, timer, schema, dependency, consent change, collection change, or cadence change. It does not change the separate Gateway shutdown lifecycle or provider-name normalization.

## User Impact

Update notices remain available during local state-write failures, and repeated calls within the original 24-hour window reuse the successful result. When storage becomes writable, the original success is persisted without delaying the next eligible check. The fallback is process-local; it cannot survive a process restart while writes remain unavailable.

## Evidence

- Red proof with the real isolated SQLite database in `query_only` mode: two calls two minutes apart returned two `null` results and sent two HTTP requests despite valid version responses. The corrected test-only candidate produced five intended failures on unchanged production code.
- Green proof: all 34 telemetry tests passed; all 113 update-startup sibling tests passed on the same final production code. Targeted lint and `git diff --check` passed.
- Coverage includes the exact 24-hour boundary, persistence recovery without extra HTTP or throttle extension, endpoint and resolved state isolation, equivalent paths, request-time destination pinning, newer durable state, and unchanged invalid-response/transport failure behavior.
- HTTP was mocked; SQLite was real and task-isolated. No production ingestion or real Gateway state was used.
- Independent review identified an unconditional-write race. A separate red proof reproduced version/timestamp rollback when a newer success was persisted during HTTP. The correction reuses the existing atomic update primitive without modifying the shared store API or implementation.
- The changed gate passed its preliminary guards, formatting, plugin boundaries, and core graph-boundary check, then stopped because the local typecheck wrapper rejects the required shared-worktree dependency symlink. The dependency layout was not replaced or bypassed; exact-head CI must supply the remaining gate proof.
- Independent P2 review: accepted concurrency findings addressed; final targeted re-review is scoped-clean. The caller now uses the transaction-selected state for both the cached response and cadence decision, preventing an unnecessary request at an older pending timestamp's expiry.
- Exact-head hosted CI is required before landing.

### DIFF
--- a/src/infra/telemetry.test.ts
+++ b/src/infra/telemetry.test.ts
@@ -3,9 +3,13 @@ import type { OpenClawConfig } from "../config/types.openclaw.js";
 import { createEmptyPluginRegistry } from "../plugins/registry-empty.js";
 import * as pluginRuntime from "../plugins/runtime.js";
 import { createPluginRecord } from "../plugins/status.test-helpers.js";
+import { writeConfigMachineState } from "../state/config-machine-state-write.js";
 import { readConfigMachineState } from "../state/config-machine-state.js";
-import { closeOpenClawStateDatabaseForTest } from "../state/openclaw-state-db.js";
-import { setTestEnvValue } from "../test-utils/env.js";
+import {
+  closeOpenClawStateDatabaseForTest,
+  openOpenClawStateDatabase,
+} from "../state/openclaw-state-db.js";
+import { deleteTestEnvValue, setTestEnvValue } from "../test-utils/env.js";
 import { useMockHttp } from "../test-utils/mock-http.js";
 import {
   createOpenClawTestState,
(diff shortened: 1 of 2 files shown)


vincentkoc:
Merged via squash.

- Prepared head SHA: [f492a9e73f244d88f101edc163379d23e1227aff](https://github.com/openclaw/openclaw/pull/141084/commits/f492a9e73f244d88f101edc163379d23e1227aff)
- Landed commit: [46c1576f82bbd629e28fc0b5c34631d53becb0ea](https://github.com/openclaw/openclaw/commit/46c1576f82bbd629e28fc0b5c34631d53becb0ea)

Issue #140737, opened 2026-09-07
fix(file-transfer): preserve actual archive member identities in directory policy


steipete:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

File Transfer can reject a permitted directory containing a newline filename because archive policy evaluates the human-readable tar listing rather than the actual member identities used during extraction.

### Steps to reproduce

1. In an isolated source checkout, create a synthetic directory containing an ordinary file whose name includes a literal line-feed character.
2. Exercise the real node directory-fetch producer, File Transfer policy and guarded extraction with the source path permitted and a different, display-derived path denied.
3. Observe that source-tree preflight accepts the actual file, but the archive-listing stage evaluates the display-derived path and rejects the transfer.
4. As a control, deny the actual filename: source-tree preflight correctly rejects it.

### Expected behavior

The source preflight and final archive policy must evaluate exact, canonical identities admitted by the same archive owner used for guarded extraction. Human-oriented escaping must not change which policy path is checked. Actual archive members, including metadata sidecar files materialized by extraction, must not disappear from policy evaluation.

### Actual behavior

The node and Gateway archive checks run tar's display listing and normalize its output. On the observed macOS tar, newline names are escaped in that listing. The resulting policy path differs from the actual member name. The real regression shows a false denial. A separate synthetic macOS metadata case shows AppleDouble members materialized by extraction but omitted from the display-derived policy set.

An authorization bypass has **not** been demonstrated: separate source-tree preflight checks actual source names. This report concerns the confirmed identity mismatch and incomplete display-derived member set, not a demonstrated data-exfiltration path.

### OpenClaw version

Confirmed source: a86a1b505d335ce6b40cb4f9f1af4ece5103f0c5. Repair regression baseline: b6a2e5b4eff71070a02e755d55aea0dd53904a20.

### Operating system

macOS; Darwin 25.6.0. Observed producer: bsdtar 3.5.3 / libarchive 3.7.4.

### Install method

Source worktree; synthetic local producer → real policy → guarded extraction regression. No operator Gateway or live user files.

### Model

Not applicable; the regression invokes the File Transfer boundaries directly without model inference.

### Provider / routing chain

File Transfer policy → headless node directory-fetch producer → archive admission → guarded extraction. No provider route.

### Screenshots, recordings, and evidence

[...]

PR #141086, opened 2026-09-07
fix(file-transfer): authorize admitted archive identities in directory policy

steipete:
Related: #140737

## What Problem This Solves

Fixes an issue where File Transfer's directory-fetch archive policy evaluated human-readable `tar` listings instead of the actual member identities used for extraction. On macOS bsdtar, a filename containing a line feed is printed with display escaping, so policy checked a different path than the real file; AppleDouble records written by the producer also disappeared from the display listing while extraction materialized them. A real regression showed a permitted source tree falsely denied by its final archive check, and a synthetic xattr fixture produced members that policy never evaluated. No authorization bypass was demonstrated: source-tree preflight checks actual names. Tracked as [#140737](https://github.com/openclaw/openclaw/issues/140737).

## Why This Change Was Made

Both the node producer and the Gateway verifier stop parsing `tar -tzf` display output (the deleted listing/tail helpers) and use the fs-safe-owned `inspectTarArchive`, which shares complete TAR/gzip admission and the extraction policy planner. Byte-count/SHA-256 identity and guarded extraction are unchanged. The inspection API shipped upstream in [fs-safe 0.8.4](https://github.com/openclaw/fs-safe/releases/tag/v0.8.4) ([PR 239](https://github.com/openclaw/fs-safe/pull/239)); main has since adopted 0.8.5, which contains it, so this PR carries no dependency or cooldown changes.

Because archives can omit parent directory headers, File Transfer also authorizes the unique implicit-ancestor closure of the returned members through the same existing policy evaluation, counting shared parents once and enforcing the existing 5000-descendant cap before acceptance. Archive admission reserves one slot for the producer's root `./` header — a real producer tree at exactly 5000 descendants otherwise trips the member count after passing its source preflight (fail-first regression included). The result manifest stays member-only; the ancestor checks happen inside policy validation.

## User Impact

Directory fetches with newline, Unicode, BOM, or numeric-looking names now receive policy decisions on exact member identities, and producer-added AppleDouble members are authorized rather than silently unlisted. An exact deny on an implicit parent still rejects the whole transfer even when the archive omits its directory header. Default behavior, limits, link/traversal/collision checks, media saving, and the model-facing projection are unchanged. The Plugin SDK gains one narrow `inspectTarArchive` export for bundled plugins with the same semantics (public surface 4446 exports / 2628 callable, +1 each).

## Evidence

The upstream API passed its full suite (8076 tests), 531 focused boundary tests, independent review, and cross-platform CI/coverage before publication ([fs-safe 0.8.4 release](https://github.com/openclaw/fs-safe/releases/tag/v0.8.4), [release run](https://github.com/openclaw/fs-safe/actions/runs/34083374893)); this PR consumes it through main's existing 0.8.5 pin.

Against main's published 0.8.5 on latest main:

- Real system-tar producer → exact policy → guarded extraction: five cases pass in each backend, including the LF false-denial regression, the actual-LF deny control, and the synthetic macOS xattr → AppleDouble member authorization case.
- `node scripts/run-vitest.mjs extensions/file-transfer` — 314 passed, 2 existing skips, in both `FS_SAFE_NATIVE_MODE=off` and `require`.
- Boundary table — exact POSIX and Windows implicit-parent denies, 2501-pair overflow at the 5000-descendant cap, shared-parent counting, and a producer-shaped archive (root `./` header + exactly 5000 members) that failed before the root-slot reservation and passes after.
- Infra/SDK facade lanes — 88 passed/1 skipped infra, 7 unit-fast, 4 plugin-sdk; `pnpm build` and `node scripts/check-changed.mjs` pass (the extension lint lane runs in a standalone physical checkout because this nested worktree's declaration boundary refuses ancestor installs).
- Fresh isolated whole-PR autoreview through P2 is scoped-clean after its descendant-boundary finding was accepted and covered by the fail-first regression.

The intended isolated live Gateway `/tools/invoke` proof on landed main is recorded separately after merge.


steipete:
Decision items resolved:

- **Published fs-safe contract review:** complete and direct — [fs-safe PR #239](https://github.com/openclaw/fs-safe/pull/239) passed its full suite (8,076 tests), 531 focused boundary tests, independent whole-PR review (no actionable findings), and exact-head CI/coverage before merge; [0.8.4](https://github.com/openclaw/fs-safe/releases/tag/v0.8.4) was published with registry integrity and provenance verification ([release run](https://github.com/openclaw/fs-safe/actions/runs/34083374893)). Main's 0.8.5 adoption (#141165) carries the same published contract, and the published 0.8.5 dist was byte-verified and exports `inspectTarArchive`.
- **Enforcement correction:** intentional and kept — previously unaudited producer-added members (AppleDouble) and implicit parents now match the same deny rules as source-tree descendants, which source preflight already enforced. No new configuration is required.
- **Descendant boundary:** admission reserves one slot for the producer's root header; the 5000-descendant policy cap is unchanged (fail-first regression in the split dir.fetch suite).

Landing proceeds under the existing maintainer authorization with exact-head CI green ([run 34146344984](https://github.com/openclaw/openclaw/actions/runs/34146344984)).

steipete:
Merged via squash.

- Prepared head SHA: [af374bf55cd6958f243d9aa8c61aa9d7a62a63b4](https://github.com/openclaw/openclaw/pull/141086/commits/af374bf55cd6958f243d9aa8c61aa9d7a62a63b4)
- Landed commit: [7cab04806c95a8193555333f2fd8520806ed1080](https://github.com/openclaw/openclaw/commit/7cab04806c95a8193555333f2fd8520806ed1080)

PR #141087, opened 2026-09-07
improve(control-ui): lead the login gate with the fix instead of the credentials form

steipete:
## What Problem This Solves

Resolves a problem where the Control UI login gate answered the wrong question first. A browser blocked on device pairing, a rate limit, a trusted-proxy rejection, or a protocol mismatch was shown an editable WebSocket URL, an editable token, a "Password (not stored)" field, and a full-width red **Connect** button, with the actual reason and the one command that fixes it buried in a danger-styled callout underneath. The most common first-connection state, "device pairing required", looked like an error even though auth had already succeeded and nothing on the page could be typed to fix it.

## Why This Change Was Made

The gate now decides its layout by where the fix lives. When the fix is in the form (missing or rejected credential, unreachable Gateway) the form leads, the failure title becomes the page heading, the summary names the Gateway host, and the offending field is marked invalid. When the fix is on the Gateway host or in the browser (pairing, rate limit, proxy, origin, insecure context, protocol/build mismatch) the explanation leads, the single resolving command is the hero action, and the form folds into a one-line "Connecting to host · token saved · Change" summary that expands on demand. Pending states use the accent tone, retry-later states use warn, and only real rejections use danger.

The three-field form is reduced to Gateway URL plus one credential input with a Token / Password switch; the switch auto-selects password mode when the Gateway reports a password-mode error code. The pairing approve command is the hero (with the request ID when the Gateway supplies one; without one it shows `openclaw devices approve --latest`, which prints the exact approve command for the newest pending request, and the steps say to run that as well), and `openclaw dashboard` is demoted to the "prefer a link?" alternative. The oversized centered brand header shrinks to a small "OpenClaw" brand row so the state title is the primary heading; the "Gateway Dashboard" subtitle and the "Control UI" label are gone from this screen, and the idle heading reads "Connect to OpenClaw". The copy-button icon swap is now owned by the gate stylesheet, so the check glyph no longer renders beside the copy glyph when chat styles are not loaded.

Non-goals: no change to connect/reconnect behavior; the client still does not auto-retry after pairing approval (recorded as a follow-up below). The Settings → Connection page is unchanged.

## User Impact

Operators landing on the gate see what is blocking them and the one thing to do next, on desktop and phone, in light and dark themes. First-time pairing reads as a normal one-time approval instead of an error, with the exact approve command ready to copy. Credential problems highlight the field to fix and no longer show a second, unrelated credential field.

## Evidence

Before / after, pairing required (the reported state):

Pairing required before and after

Token needed (form-led layout, invalid field highlighted):

Token needed before and after

Idle gate (no failure yet), Gateway URL plus one credential field:

Idle connect screen

Phone (390px) and dark theme:

Pairing required on a phone

Pairing required in dark theme

Captures are real renders of the component from a Vite source server with the repo's mock Gateway harness; the host shown is the maintainer's own public Gateway hostname and the token value is a placeholder.

- Unit: `login-gate.test.ts` 24/24 + `vite-config.node.test.ts` 25/25 (two pairing tests rewritten to the new contract: hero approve command, two steps, folded connection summary, unsafe request IDs fall back to `--latest`; new cases pin the credential mode while a token is replaced, keep Connect beside Refresh on protocol mismatch, and select the password field when the Gateway reports a password-mode error despite a saved token).
- Playwright E2E locally: `login-gate.e2e.test.ts` 20/20 (new case covers the idle help disclosure), `browser-bootstrap.e2e.test.ts` 1/1, `control-ui-credentials.e2e.test.ts` 1/1. `control-ui-auth-transports.e2e.test.ts` needs a real Gateway build and its setup hook timed out locally under load; CI covers it.
- `node scripts/check-changed.mjs`: every lane green except `lint core`, which flagged `max-lines` on the first cut of `login-gate.ts`; the failure classification now lives in `login-gate-feedback.ts` and the lane was rerun (see the CI run on this head). Two shrink-only baseline updates: assertion-safety baseline for `login-gate.ts` 3 → 2, i18n raw-copy baseline for the changed placeholders. `pnpm check:architecture`: zero cycles.
- Codex autoreview (P2 threshold) on the full diff: two accepted findings fixed (credential mode flipping while clearing a token; Connect removed on mismatch layouts), rerun clean. The landing review then caught two more, also fixed: `approve --latest` only previews, so the no-ID steps now say to run the printed command; a password-mode error now selects the password field even when a token is saved.
- stylelint, oxfmt, `pnpm ui:i18n:verify`, `pnpm tsgo:ui`, and the core oxlint shard clean on the changed files.

Follow-up worth doing separately: the client stops reconnecting on a pairing-required close, so the last step is still "click Connect". The gate could poll `connect` while pairing is pending and admit the operator the moment approval lands.


steipete:
Merged via squash.

- Prepared head SHA: [4fbba0c7c6e91b8ce23bb46ed09e4054f53f19bb](https://github.com/openclaw/openclaw/pull/141087/commits/4fbba0c7c6e91b8ce23bb46ed09e4054f53f19bb)
- Landed commit: [ed9861f2e87bba69c91630b214c8783782460a2f](https://github.com/openclaw/openclaw/commit/ed9861f2e87bba69c91630b214c8783782460a2f)

PR #141096, opened 2026-09-07
fix(ui): show own avatar in New Session preview and focus composer quietly

steipete:
## Problem

Submitting a prompt from the New Session page had two visible glitches:

1. The immediate "Starting…" preview showed the prompt bubble without the sender's avatar. The avatar only appeared once the created session's transcript replaced the preview, so the row visibly changed shape a moment after send.
2. After navigating into the created session, the composer flashed with the accent-colored prefill attention animation (orange border + glow) even though nothing was prefilled. Started-session navigation and the assistant panel request composer focus through the same route hint as a draft prefill, and the pane treated every focus request as a prefill.

## Solution

- `ui/src/pages/new-session/draft-composer.ts` renders the own-user avatar in the submission preview using the same slot renderer as the chat thread. The placement decision (gutter when the viewer has a profile identity, footer otherwise) now lives in one shared helper, `resolveChatDefaultAvatarPlacement`, used by both the transcript projection and the preview, so the preview and the committed transcript row agree. `renderUserAvatarSlot` moved to `chat-author-avatar.ts` so both surfaces share it; chat's render contract is unchanged.
- `ui/src/pages/chat/chat-pane-lifecycle.ts` only plays the prefill attention cue when the focus request carries a non-empty draft. Bare focus requests still focus the textarea, silently.
- Docs: `docs/web/control-ui/sessions-and-sidebar.md` notes the immediate avatar and quiet focus.

## Impact

- Team-mode users see their avatar on the first paint of a new session's prompt, in the same position it will keep after the transcript loads.
- No composer flash on new-session navigation or the assistant panel's home shortcut. The attention animation still runs for navigation that prefills a draft.
- Net production change is small (shared helper + one guard); no config, schema, or i18n changes.

## Evidence

Local gateway from this branch's build (owner profile via token auth) driven with Playwright, trusted keyboard Enter. DOM timeline recorded with a MutationObserver:

- Before: preview `.chat-group.user` had no avatar element until the chat thread replaced it (~150 ms later); `agent-chat__input--prefill-attention` was added to the composer on navigation and removed 600 ms later.
- After: preview renders `.chat-avatar-slot > img` immediately; the attention class never appears; the textarea is still focused after navigation.

| Before: preview without avatar | After: avatar on first paint |
| --- | --- |
| before preview | after preview |

| Before: composer flash after navigation | After: quiet focused composer |
| --- | --- |
| before composer | after composer |

Tests:

- `node scripts/run-vitest.mjs ui/src/pages/chat/chat-pane-lifecycle.test.ts` (34 passed): draft prefill keeps the cue; plain focus and empty draft never add the attention class.
- `node scripts/run-vitest.mjs ui/src/pages/new-session/composer.test.ts` and sibling avatar/transcript suites (337 passed): the preview renders the own-user avatar in the gutter with a profile id and in the footer without one; both cases fail on the previous code.
- `node scripts/check-changed.mjs` passed; `pnpm check:architecture` passed; Codex autoreview (P0–P2) scoped-clean.

### DIFF
--- a/docs/web/control-ui/sessions-and-sidebar.md
+++ b/docs/web/control-ui/sessions-and-sidebar.md
@@ -156,6 +156,8 @@ If startup is rejected, the draft remains available to correct and retry.
 
 On a normal foreground OpenClaw Chat send, the submitted text and attachments appear immediately with a **Starting** indicator while the Gateway creates or adopts the session. This is a pending submission, not a Gateway acknowledgment. If creation is rejected, your prompt and attachments remain available to correct and retry. Once creation succeeds, the UI opens the session's chat.
 
+Attributed submissions show your avatar immediately, in the same position as the chat transcript. Opening the created session focuses the composer quietly; the attention cue is reserved for navigation that prefills a draft.
+
 The project picker refreshes after sign-in and reconnects. Gateway reconnects and Git verification retries preserve your edited base branch and worktree name. Choosing another folder or project clears those repository-specific details.
 
 For local worktree sessions, sending the first message opens the admitted session before naming, checkout, and setup finish. The chat shows the submitted message and preparation stages. A generated title is saved as soon as naming completes, independently of checkout and setup. Setup failures remain visible in that session; send a retry there after correcting the problem. The retry reuses the saved title. If naming itself fails, another attempt uses the original first prompt, including text attachments. Stopping during setup cancels preparation without starting the agent. Steering an active run keeps its progress visible, and delayed history cannot replace a newer startup stage or restore startup labels after activity begins.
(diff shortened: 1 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [60f1a088c2e49942c90f9af6d8b08e15b9fd3d42](https://github.com/openclaw/openclaw/pull/141096/commits/60f1a088c2e49942c90f9af6d8b08e15b9fd3d42)
- Landed commit: [78d00a580fc5bb2e0f811e6cc1de851fbfbb8dcf](https://github.com/openclaw/openclaw/commit/78d00a580fc5bb2e0f811e6cc1de851fbfbb8dcf)

PR #141098, opened 2026-09-07
improve(ui): show sidebar running state on the row icon instead of the trailing edge

steipete:
## What Problem This Solves

Resolves a problem where hovering a running session row in the Control UI sidebar crowded four trailing items together: the run spinner, the pin button, the session menu, and the collapsed-children chevron with its count. Running state and hover actions shared the same trailing slot, and since #128973 the spinner no longer yields to the actions, so a running parent with collapsed children got the worst of both.

The placement was also inconsistent across the sidebar: child session rows have always drawn their running ring around the leading glyph (#113645), while Home, top-level session rows, and catalog rows drew a spinner at the right edge (#119560).

## Why This Change Was Made

One rule now decides where "this is running" appears. When the row itself is running, its leading glyph carries the ring: Home, top-level sessions, child sessions, and Codex/Claude catalog rows alike. Rows with no icon, avatar, or owner chip draw a bare ring in the same always-reserved lead slot. Progress appears on the right only when a row summarizes rows you cannot currently see, which is what the collapsed section header spinner and the collapsed child toggle's accent tint and count already do; those are unchanged.

The ring reflects the row's own run only: an idle parent whose hidden child is running keeps a plain glyph, and its collapsed toggle carries the accent tint and count instead. The trailing slot of a session row therefore holds only hover-revealed actions plus the unread dot and PR badge, which already yield to the actions on hover. Accessible state is announced exactly once per row: the ring announces the run, the row description covers fork and unread, and Home keeps its composed link label. Unread placement is deliberately untouched.

## User Impact

Hovering a running session reveals pin and menu beside the chevron without a spinner squeezed between them, and nothing in the row moves. The running signal reads in one column, around each row's own icon, whether the row is Home, a session, a child session, or a catalog entry. A right-side spinner now always means "expand this to find the running row". Screen readers hear "Active run" or "Queued" once per row instead of twice.

## Evidence

Mocked-Gateway Chromium captures (synthetic fixtures only), 1200x800 dark theme. Left column before, right column after.

| State | Before | After |
|---|---|---|
| At rest | Before: running rows at rest | After: running rows at rest |
| Parent row hovered | Before: hovered running parent row | After: hovered running parent row |
| Home running | Before: Home running | After: Home running |

- Regression tests failed on the untouched tree for the intended reason (missing leading rings, remaining trailing state, double announcements) before the production change: `Tests 9 failed | 326 passed (335)` in the sidebar component suite, and the browser suite reported `Active run: expected length 1 but got 2` plus a missing leading ring.
- After: `node scripts/run-vitest.mjs run ui/src/components/app-sidebar.test.ts` → 336 passed (includes a new idle-parent case that fails on the aggregate predicate); the seven Playwright suites that anchor on the sidebar run indicator (`session-management.trailing-state`, `chat-flow.sidebar-presentation`, `session-management.sidebar`, `groups`, `unread`, `queue`, `sidebar-customization`) → 44 passed; `node scripts/check-changed.mjs` → exit 0; `node scripts/ui.js build` → startup CSS 43986 B gzip under the 51200 B ceiling.
- Accessible names and descriptions are asserted in real Chromium: "Active run", "Queued", and "Unread" each occur once per row across name plus description.
- Production delta +28/−55 lines across six UI source and style files (net −27); CSS drops the orphaned queued-spinner selector and rewrites two stale comments, three lines shorter. No config, i18n, snapshot, baseline, or budget changes.
- Autoreview (Codex) on the final tree: scoped-clean at P2 (an earlier pass flagged that the ring still aggregated child activity onto idle parents; fixed and covered by the new unit case before this final pass).

Follow-up worth a separate change: the unread dot on top-level rows could move to the glyph corner badge that child rows already use, which would finish the same left/right rule for unread. Kept out of this PR to keep the diff about one thing.


steipete:
CI status on the first run of `1ff77bd9a37` (re-anchor in `eb7ca507892`):

- `checks-ui-e2e (5/7)` was a real miss on my side: `chat-run-lifecycle.e2e.test.ts` waited for Home's "Active run" image inside the trailing state cluster, which this PR moves onto the Home glyph ring. Re-anchored to the glyph in the follow-up commit; the suite plus the draft-indicator suite pass locally.
- `checks-node-compact-small-29` failed in `test/scripts/ci-node-test-plan.test.ts` ("keeps hosted tooling within the GitHub job cap when its inventory grows"), which ran silently for 90 s on a hosted runner and hit the 120 s test timeout. This PR touches only `ui/` and `docs/`, so it cannot affect the shard planner. Filed as a separate repair of that test's runtime rather than a timeout bump.

steipete:
Addressed in 86959e4f32c: the collapsed child toggle now carries `aria-description="Active run"` while `runningChildCount > 0`, and drops it once the children are expanded and carry their own rings. `session-management.sidebar.e2e.test.ts` asserts both states through Chromium's accessibility tree (`Accessibility.getFullAXTree`), and the pre-fix run failed with `expected undefined to be "Active run"`. The idle parent's own glyph stays unringed, per the left/right rule in the PR body.

steipete:
Merged via squash.

- Prepared head SHA: [86959e4f32cda73f03271e803e2e13b58dcdb62a](https://github.com/openclaw/openclaw/pull/141098/commits/86959e4f32cda73f03271e803e2e13b58dcdb62a)
- Landed commit: [701687f6b7104dca7ba990a2557c840206af8ee8](https://github.com/openclaw/openclaw/commit/701687f6b7104dca7ba990a2557c840206af8ee8)

PR #141100, opened 2026-09-07
perf(plugins): defer installer loading for policy commands

steipete:
## What Problem This Solves

`openclaw plugins enable` and `disable` import the shared management mutation module. Its eager installer dependency and uninstall re-export also load package installation/removal code, even when the command only changes an installed plugin's policy. This remains after #140842 made bundled slot selection metadata-only.

## Why This Change Was Made

Load the existing installation implementation only when `installManagedPlugin` is invoked, before entering its existing lifecycle lease. Internal uninstall callers import the existing uninstall owner directly instead of reaching it through the mixed mutation module. This removes the unnecessary imports without adding another cache, wrapper, or activation path.

## User Impact

Enabling or disabling an installed plugin no longer requires loading installation/removal backends. External installation, legacy runtime-kind discovery, capability consent, restrictive policy, configuration ownership and uninstall behavior remain covered by their existing paths. No config/default, timeout, schema or persistence change is introduced. The removed re-export was an internal module path, not a Plugin SDK contract.

## Evidence

- An observable CLI regression makes the installation and removal implementations unavailable, then enables and disables a tiny bundled plugin and verifies persisted policy. Before the change it fails when `management-mutations.ts` imports the installation implementation; afterward it passes.
- All 156 focused tests pass across CLI policy, Gateway mutation handlers, external installation, uninstall ownership, capability consent, registry refresh, workspace inventory and runtime-kind inspection. Changed-surface checks and the staged temp-directory report pass. Full `pnpm build` passed; the prescribed Codex entrypoint import/RSS profile passed. Independent managed review through P2 is clean.
- **Real changed-surface proof:** the actual built `openclaw.mjs plugins enable codex --accept-capabilities` and `openclaw.mjs plugins disable codex` commands ran against an isolated HOME/config/state, with the existing lifecycle trace flag and a module-load observer. They exited 0 and persisted `enabled: true` then `enabled: false`. Neither installer nor removal owner loaded in either candidate command. This proof uses the built CLI; the separate sentinel-mock regression protects its import boundary. Before, the built enable trace included management installation/removal, ClawHub client/packages/skills, Git installation and marketplace modules. Those identified modules are absent in the rebuilt candidate; the observed enable module count fell from 2,275 to 2,190.
- Local diagnostic samples were 2.356 s and 1.836 s, with 93 major faults each and filesystem-operation counters reported as zero. These runs used different source snapshots and are not a controlled timing comparison. This PR does **not** attribute or claim to fix the separate 36.649-second remote plugin-activation observation or the complete cloud bootstrap latency. Remote phase attribution remains a separate investigation.

Production delta: **−4 lines**; tests **+66 lines**. Only the existing install owner's loading boundary and internal uninstall imports change.

The changed surface is internal import loading during local plugin policy commands. It does not alter an external API or provider operation. Remote cloud timing was not exercised for this PR and is not claimed; the broader latest-main cloud performance investigation continues separately.

### DIFF
--- a/src/gateway/server-methods/plugins-mutations.ts
+++ b/src/gateway/server-methods/plugins-mutations.ts
@@ -19,8 +19,8 @@ import { ManagedPluginLifecycleError } from "../../plugins/management-lifecycle-
 import {
   installManagedPlugin,
   setManagedPluginEnabled,
-  uninstallManagedPlugin,
 } from "../../plugins/management-mutations.js";
+import { uninstallManagedPlugin } from "../../plugins/management-uninstall.js";
 import { buildGatewayReloadPlan } from "../config-reload-plan.js";
 import { resolveGatewayReloadSettings } from "../config-reload-settings.js";
 import type { GatewayRequestHandlers } from "./types.js";
--- a/src/gateway/server-methods/plugins.test.ts
+++ b/src/gateway/server-methods/plugins.test.ts
@@ -27,6 +27,9 @@ vi.mock("../../plugins/management-service.js", () => ({
 vi.mock("../../plugins/management-mutations.js", () => ({
   installManagedPlugin: (...args: unknown[]) => managementMocks.install(...args),
   setManagedPluginEnabled: (...args: unknown[]) => managementMocks.setEnabled(...args),
+}));
+
+vi.mock("../../plugins/management-uninstall.js", () => ({
   uninstallManagedPlugin: (...args: unknown[]) => managementMocks.uninstall(...args),
 }));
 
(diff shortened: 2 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e221bf2fb0c2facbcf86ba0072dadf8cb9bf42e0](https://github.com/openclaw/openclaw/pull/141100/commits/e221bf2fb0c2facbcf86ba0072dadf8cb9bf42e0)
- Landed commit: [d4b2d73cf606c257a5ec4a0fbc5a3c7a6fd0dcce](https://github.com/openclaw/openclaw/commit/d4b2d73cf606c257a5ec4a0fbc5a3c7a6fd0dcce)

PR #141112, opened 2026-09-07
test(slack): close thread-resolution fixture reliably

steipete:
## What Problem This Solves

The Slack thread-resolution fixture can fail under Bun while closing its local HTTP server, reporting `ERR_SERVER_NOT_RUNNING` instead of completing the cache and request-coalescing check reliably.

## Why This Change Was Made

Use the existing `withServer` test helper to own listening, accepted sockets, and shutdown. The callback retains the same request handler, clients, cache/request/payload assertions, and test deadline. API URLs are derived from the helper's loopback base URL.

## User Impact

Developers can run this existing thread-resolution check under Node and Bun without the inline server-cleanup failure. There is no production or dependency change.

## Evidence

- Exact existing case: `coalesces Enterprise lookups and retains separate caches for replacement and default clients`.
- Before: Node passed; Bun failed in inline cleanup with `ERR_SERVER_NOT_RUNNING`. Because the error arose in `finally`, the original Bun report alone does not establish that every preceding body assertion passed.
- After: Node and Bun each passed the entire unchanged assertion body and helper cleanup. Runs used isolated temporary HOME/state, one worker, the original deadline, and a 4 GiB process-tree cap. All processes joined without watchdog intervention.
- Fresh P0–P2 review passed with no actionable findings; formatting and diff checks passed.
- The full changed-file gate passed on refreshed head 580e3ddd in an independently installed worktree, with all descendants joined. An earlier nested native-worktree run stopped at the declaration ancestor-install boundary before lint; moving proof to a proper isolated checkout resolved the environment failure without changing source or guards. Exact-head CI34140467149 and WorkflowSanity34140467082 passed, and hosted review is ready with no findings.

Only this loopback fixture was run locally; no live Slack service or broader suite was exercised. An initial zero-selection command was rejected by the selector audit and is not counted as proof.

The refreshed head is a GitHub-verified signed merge of fixed main into the original PR branch. The fixture blob is unchanged, and it remains the only file in the PR diff. Focused runtime proof preceded this base refresh; the full gate and hosted checks cover the refreshed head.


steipete:
Merged via squash.

- Prepared head SHA: [580e3ddd8278ba24a88f85352523125bc9f6f20c](https://github.com/openclaw/openclaw/pull/141112/commits/580e3ddd8278ba24a88f85352523125bc9f6f20c)
- Landed commit: [2f9bf3d88c8b0d280f8f9fafa58d1c1745c73ec2](https://github.com/openclaw/openclaw/commit/2f9bf3d88c8b0d280f8f9fafa58d1c1745c73ec2)

PR #141114, opened 2026-09-07
fix(agents): reuse warm prefixes for single-chunk compaction

steipete:
## What Problem This Solves

Fixes cold compaction requests when the same Anthropic Messages or OpenAI Responses route already has the conversation prefix cached. Eligible single-request summaries reuse the foreground system prompt, tool declarations, and verified native history. Invalid native summaries and genuinely changed history retain serialized fallback.

## Why This Change Was Made

The embedded live scenario on the previous head completed with `paths=serialized usageRecords=1` and no cache read. A deterministic replay now drives two real embedded runs and on-demand manual compaction with mocked streams, large stable prefixes, and the same model alias and 512-token target budget.

The no-tool case captured 1 then 3 foreground messages. Prefix resolution successfully matched the 2-message summary source against the last 3-message request. A later core eligibility check then planned hypothetical safeguard chunks against the capped 512-token context and discarded the verified prefix. Core compaction actually emits one history-summary request. Removing that redundant check restores reuse without changing budgets, split-turn handling, or actual safeguard chunking.

The tool case captured 1, 3, 5, then 7 messages. Its 4-message summary source failed at assistant index 1: replay had normalized the tool-call ID from `read-1` to `read1`. Prefix resolution now reuses the captured foreground ID-normalization policy and replay tool allowlist. It first checks the original projection for same-attempt turns; either projection must match every dispatched message fingerprint exactly. No transcript fields are ignored, and no transcript bodies or executable tools are retained in the bounded session cache.

## User Impact

Matching routes can reuse warm prefixes during compaction without new configuration. Changed history, missing snapshots, split turns, actual multi-chunk summaries, and previous-summary redistillation retain fallback. Foreground tools are never executed by summarization. Native output must satisfy the existing summary heading contract; provider rejection or malformed output triggers serialization.

Per-request usage diagnostics now include a serialization reason: missing snapshot, history length, digest mismatch with index and role, changed route/source, missing heading contract, or rejected/failed prefix request. Exception text is excluded. Both completed requests are accounted for when a native response is rejected. See `docs/concepts/compaction.md`.

## Evidence

- 450 focused tests across seven files passed, including real embedded execution, core/session compaction, safeguard, format propagation, and hook/accounting behavior.
- Both full embedded regressions failed before the repairs and now require exactly one `foreground-prefix` request. Final assertions compare the entire native history prefix and tool declarations with the real foreground request.
- Fallback cases verify changed-history index/role diagnostics, overlong history, changed model, actual chunking, rejected native output, and a throwing prefix request.
- The live compaction fixture now requires a final `foreground-prefix` request and positive cache read, rather than passing on serialization alone.

Follow-up head: `69378adab6bfed3ad4152bc4f961d1dd950b9717`. `node scripts/check-changed.mjs` passed, including production/test typechecks, lint, ownership/state guards, and zero runtime import cycles. Fresh P1 autoreview on the final diff is scoped-clean with no accepted/actionable findings. The five affected embedded/accounting cases also passed after moving usage collection into the existing diagnostics owner. Exact-head [CI run 34143177283](https://github.com/openclaw/openclaw/actions/runs/34143177283) completed successfully. The canonical watcher exited 0 with `pending=0` and `GREEN` after reconciling six superseded checks; no workflow rerun was needed. The requested [ClawSweeper re-review](https://github.com/openclaw/openclaw/pull/141114#issuecomment-5573398520) was posted after pushing. No live provider calls were made for this follow-up. Mocked request reuse is proven; a warm live cache hit remains for the coordinator to verify. No OpenAI live result is claimed.

Coordinator rerun with provider credentials supplied through the existing live-test setup:

```sh
OPENCLAW_LIVE_TEST=1 OPENCLAW_LIVE_CACHE_TEST=1 \
node scripts/run-vitest.mjs run --config test/vitest/vitest.live.config.ts \
  src/agents/embedded-agent-runner.cache.live.test.ts \
  -t 'preserves cache-safe shaping across compaction followup turns'
```

This PR remains open. No prepare or merge operation is requested.


steipete:
Follow-up pushed: `59d19c8e1ab76bc15a6a490d8b7110e2c7f2dbf9`.

Addressed the coordinator fixes: all three newly introduced cross-package type imports now use `@openclaw/agent-core`, and the summarizer reuses the exact source array already admitted by the host's digest check. Separate equivalent arrays retain the existing structural comparison, so eligibility and fallback behavior stay the same.

For ClawSweeper's review: there were no concrete correctness/security findings to reject. I accept the build and live-evidence requests. The full `pnpm build` passed. All 267 focused tests passed, and complete-candidate autoreview is P1-clean (`scoped-clean`).

A synthetic live Anthropic Messages probe exercised the production summarizer with both its default format and a custom format. Native requests received cache hits (the first default summary read 6,674 cached tokens); both formats retained all five exact identifiers, the 5% canary decision, the 0.1% threshold, and the pending trace inspection. The corrected probe passed. Its initial runs exposed two fixture issues: the warmup acknowledgment exhausted its small output budget, and a wording assertion rejected a correct passive-voice sentence. Those were corrected only in the local proof artifact.

This is limited evidence, not complete summary-quality clearance: the output also included an extraneous `ACK`, and the default summary inferred a completed build deployment beyond the explicit source facts. The full embedded compaction cache fixture was attempted twice and was terminated by its 120-second no-output watchdog before reaching a compaction result, including on a rerun after the build. No OpenAI live result is claimed. The broader embedded-flow and semantic-fidelity validation therefore remains a pre-landing requirement; I am not treating fixture responses or the narrow successful probe as proof of that broader requirement.

`node scripts/check-changed.mjs` also passed, including package boundaries, architecture/type checks, lint, and zero runtime import cycles. 

CI on `59d19c8e1ab76bc15a6a490d8b7110e2c7f2dbf9`: the requested watcher attached to [run 34112236488](https://github.com/openclaw/openclaw/actions/runs/34112236488) and exited 15 after `checks-node-compact-small-29` failed. Its only failing test was the unchanged `test/scripts/ci-node-test-plan.test.ts` case “keeps hosted tooling within the GitHub job cap when its inventory grows”, which timed out at 120,000 ms ([job log](https://github.com/openclaw/openclaw/actions/runs/34112236488/job/101711099209)). The test and its planner/list-inventory sources are unchanged by this PR; it exercises CI job packing rather than the compaction runtime. This CI-tooling timeout is recorded separately from the compaction repair. `check-lint`, both core lint lanes, and [runtime topology/architecture](https://github.com/openclaw/openclaw/actions/runs/34112236488/job/101711092183) passed. The earlier Telegram max-lines failure did not recur, and that file remains untouched. GitHub API timeouts delayed status/log retrieval. Final check rollup: 133 successful, 47 skipped, one neutral; the failed planner job and its aggregate `openclaw/ci-gate` are the only failures, with no checks still pending.

steipete:
@clawsweeper re-review

Updated head: `69378adab6bfed3ad4152bc4f961d1dd950b9717`. Fixed the core hypothetical-chunk rejection and replay tool-ID mismatch; added bounded fallback reasons and real embedded regression coverage. Local changed checks and P1 autoreview pass. The PR body has the diagnosis and proof.

steipete:
Parking this PR as a draft after three coordinator live runs of the Anthropic embedded "compaction followup" scenario against real keys (runs on heads 59d19c8e, dd8834ef, 69378ada):

- On 69378ada the foreground-prefix path is now eligible and used (`request=1 path=foreground-prefix reason=eligible`), but that request reported `cacheRead=0`: the reused prefix did not hit the provider cache, so there is no cost benefit yet.
- The model's reply on the prefix path was rejected by the new summary guard (`prefix-response-rejected`) and the compaction fell back to the serialized request, which is exactly the summary-contamination risk ClawSweeper raised.

Both need a different approach before this can land (a prefix request that is byte-identical to the last foreground request, and a summarization framing that reliably yields a summary rather than a conversational reply). Keeping the branch as an experiment; not landing in the prompt-cache program.

PR #141121, opened 2026-09-07
feat(ui): render sidebar and chat from cached state before the Gateway connects

steipete:
## Problem

A Control UI reload showed only the skeleton splash until the WebSocket hello completed, then an empty sidebar until the roster arrived, then the transcript. All of that data already lived in the browser: transcripts in the IndexedDB snapshot store (with delta cursors) and preferences in localStorage. Three gates hid it:

- `app-root.ts` rendered the connect splash for the whole document while the phase was `connecting`.
- `route-loader.ts` awaited the Gateway for `main` targets and issued `chat.startup` for short-id targets before the route could commit, so the chat pane never mounted early enough for the stored snapshot to matter.
- The session capability published `result: null` until the connect roster answered, so the sidebar drew zero rows.

## Solution

Render from cached state first, then reconcile against live data on hello. The only Gateway change is an additive optional hello-ok field, `auth.method`, that reports which authentication method accepted the connection.

- **Boot record** (`ui/src/app/boot-record.ts`, localStorage `openclaw.control.bootRecord.v1:<credential scope>`): the last live `agents.list`, custom groups, section order, and the profile id. Read synchronously at boot; written after each live roster (eager writer, flushed on page hide); 64 KB and 30-day bounds; avatar fields stripped. It is the single gate for every other warm read.
- **Session roster cache** (`ui/src/lib/sessions/session-roster-cache*.ts`, IndexedDB `openclaw-session-roster`): the primary sidebar list with volatile run state stripped (active run, run ids, status, runtime, observer digest, swarm, placement, subagent state, channel avatar URL); at most 200 rows / 1.5 MB, 6 scopes, 30 days. Accepted on boot only with a boot record and matching scope, selected agent, and query; profile validated on hello. Writers, eviction, and clearing live in a lazy `.runtime.ts` module.
- **Capabilities**: `agentsListCached` and `resultCached` flags. A live result always replaces the cached one; cached rows are never merged, appended to, or preserved into the live list, and never advance the canonical list revision.
- **Route loader**: skips the Gateway wait for `main` targets when defaults are cached (per-sender scope only; global scope keeps today's wait) and resolves short ids locally from cached rows before falling back to `sessions.resolve` / `chat.startup`.
- **Shell**: with a usable boot record the shell renders during the initial connect (`runtime.warmBoot`); a failed connect still swaps to the login gate. The sidebar footer shows "Connecting…" for the never-connected phase instead of the offline copy.
- **Transcript**: the routed session's stored snapshot is prewarmed at boot through a light IndexedDB reader (validated by the snapshot store on consume), and the first `chat.startup` after a warm mount waits up to 300 ms from connection readiness for that read so it carries the stored `deltaCursor`; the Gateway then answers with a delta instead of a full page. The wait is skipped when the loader already holds a startup handoff or the pane mounted after readiness without a prewarm.
- **Invalidation**: a credential change or a hello that reports a different profile clears the boot records synchronously (so a reload during the asynchronous IndexedDB cleanup fails closed), retires the in-memory cached agents and roster immediately (keyed on the gateway connection revision), and drops stored transcripts and the roster database.

Startup JS gzip is 350,195 B, 182 B under the committed baseline and 758 B under the enforcement limit; the first attempt validated with zod on the boot path and pulled the 29 KB `config-runtime` chunk into startup, which is why validation is hand-written narrowing and the prewarm reads through the small database module only.
- **Identity binding**: warm state is written only when hello reports a credential the browser itself holds (`token` or `device-token`) and is keyed to an FNV-1a fingerprint of the credential that actually authenticated (boot record v2; v1 records are rejected and removed). At boot the record is accepted only when the credential the browser can present matches that fingerprint. Trusted-proxy, Tailscale, password, and bootstrap identities persist nothing and always show the initial connection screen, so switching users behind a proxy on the same browser profile never shows the previous user's cached rows.
- **Boot cost**: the roster cache reader and the transcript snapshot reader load on demand off the boot path and `loadSkillStatusReport` was split from the lazy Skills management module. The roster cache excludes Incognito rows at write and restore, the roster wait releases on Gateway readiness, and the warm-boot writers stay inert without IndexedDB or window listeners. Startup JavaScript measures 350,389 B gzip locally against the unchanged 350,953 B limit; CI's Linux build measures about 1.1 KB above the local number for the same head.

## Impact

Reloads paint the real shell and the cached sidebar rows before the socket opens, the routed conversation renders from the stored snapshot, and the post-hello traffic shrinks to a transcript delta plus the live roster. Docs: `docs/web/control-ui/offline-and-reconnect.md` (new "Warm reload" section) and `docs/web/control-ui/sessions-and-sidebar.md`.

## Evidence

Isolated Gateway on loopback with the mock provider, 16 seeded sessions, reload into a 24-turn session, Playwright driving a persistent profile. Times are milliseconds after navigation start.

**Instrumented reload (IndexedDB `get` and WebSocket frames hooked from document start, no CDP), second warm reload, same machine minute:**

| build | shell | roster read | 10 sidebar rows | hello | transcript | first `chat.startup` |
| --- | --- | --- | --- | --- | --- | --- |
| before (8bd02ac5a39) | 74 (after hello) | discarded | 91 | 71 | 180 (live 60 KB page) | 74, short id, no cursor |
| after (this head) | 38 | 50 | 81 (before hello) | 83 | 143 (stored snapshot) | 151, `cursor` set |

**CDP harness (Playwright + `Network` domain, 5-run medians, emulated 150 ms handshake latency):** shell 191 ms vs 348 ms, skeleton visible 5 ms vs 169 ms, ten sidebar rows 378 ms vs 367 ms, transcript 458 ms vs 413 ms, `chat.startup` 59,962 B without cursor on both. Under that harness the IndexedDB reads settle a few milliseconds after the loopback hello, so it exercises the fallback path (Gateway short-id resolution), and run-to-run machine-load variance was larger than the transcript delta. On a real network hello arrives hundreds of milliseconds later than the boot-time reads, which is the case the cache targets.

Captures below are from the CDP harness at 150 ms latency, taken 150 ms after the reload committed (synthetic sessions and mock replies only).

Before: blank canvas 150 ms after reload commit

After: shell, cached sidebar rows, and connecting state 150 ms after reload commit

Local proof: `pnpm check:changed`; the new unit tests (boot record, roster cache, agent and session capabilities, route loader, chat pane, app root, warm-boot profile validation, credential-change retirement, transcript prewarm and startup ordering); the warm-reload e2e (`ui/src/e2e/warm-boot.e2e.test.ts`: cached shell, roster, and transcript visible while the hello is held, no `sessions.list` or `chat.startup` before release, `chat.startup` with `cursor` and live roster replacement after, previous profile's transcript dropped on a mismatching hello); `pnpm ui:build` budget report; `pnpm ui:i18n:verify`; Codex autoreview scoped-clean at P1 on the final head.

**Upgrade compatibility**: both stores are additive and versioned. Boot records live under `openclaw.control.bootRecord.v1:<scope>` with a `version` field; records from the earlier revision of this branch (version 1, no credential binding) are rejected on read and removed, which costs one cold boot. The IndexedDB roster database is version 1 with per-record shape validation and eviction bounds, and the transcript snapshot store is unchanged. A browser profile without either store boots exactly as today; a credential change or a hello reporting a different profile clears the boot records synchronously before the lazy IndexedDB cleanup. Covered by `ui/src/app/boot-record.test.ts` (version and fingerprint rejection) and the trusted-proxy scenario in `ui/src/e2e/warm-boot.e2e.test.ts`.

**Known gap (pre-existing, tracked in #141690)**: the transcript snapshot store keys records by agent and session without a Gateway scope, so a browser profile used against two Gateways can show one Gateway's cached transcript for a shared session key such as `agent:main:main` until the live delta replaces it. On `main` that happens after hello; with the prewarm it can happen before hello for the same credential holder. Scoping the snapshot key by Gateway is a persistent-store change and is deferred to the follow-up.


steipete:
Thanks for the review. Disposition of the findings against the rebased head (the reviewed head `a05df90ef92` predates the rebase onto current `main`):

- **[P1] Exclude Incognito sessions before persisting the roster** — accepted and fixed. `isPersistableSessionRow` now filters `incognito: true` rows at the write boundary (`session-roster-cache.runtime.ts`) and again on restore (`session-roster-cache.reader.ts`), so a record written by an older writer cannot resurface them; the regression test covers both boundaries. The transcript snapshot store is unchanged by this PR and already predates it.
- **[P2] Release the roster wait when the Gateway becomes ready** — accepted and fixed. `whenCachedRosterSettled` now races the IndexedDB read against Gateway readiness (and disposal); a late read still cannot publish cached rows after retirement. Regression test added.
- **[P2] Provide the required boot record in the sidebar fixtures** — fixed during the rebase (`test(ui): reconcile warm proof fixtures with main`); the fixtures now pass a boot record and the separate no-record admission test remains.
- **[P2] Validate short-link destinations beyond the cached subset** — not taken. Cached short-id resolution requires a unique match on the URL's short id (at least 8 hex characters of the trailing UUID) plus, when present, the slug, within the same 200-row primary page the in-memory sidebar roster already uses for the identical purpose on in-app navigation (`findUiSessionRow`); a wrong destination requires a second session whose UUID shares that prefix, which is a 2^-32 event per candidate. A stale cached row that no longer exists on the Gateway yields an empty transcript and is replaced by the live roster on hello. Keeping the same semantics for reload as for in-app navigation is deliberate; a post-hello `sessions.resolve` cross-check would add a round trip to every warm reload for that probability.
- **Merge risks** — the conflicts are resolved by the rebase (head now applies cleanly on `main`). The two captures are fixture-only (isolated Gateway, mock provider, UUID-named seeded sessions); they were inspected before upload. Compatibility with an existing browser profile: both stores are additive (`openclaw.control.bootRecord.v1:*` in localStorage, IndexedDB `openclaw-session-roster` v1) with versioned records that are removed on any shape mismatch, so a profile without them boots exactly as today, and credential change or a mismatching profile on hello clears them synchronously before the lazy IndexedDB cleanup.
- **Decision needed** — the bounded cache design (retention, pre-hello visibility, clearing, upgrade behavior) is the maintainer's accepted design: it was specified by the maintainer before implementation and this PR lands it on the maintainer's instruction.

steipete:
Follow-up on the remaining item and on a new finding from the independent review of this head:

- **Data-model compatibility proof** — recorded in the PR body under Evidence ("Upgrade compatibility"). Both stores are additive and versioned: boot records carry a `version` field and any record that fails validation is removed on read (a version-1 record from the earlier revision of this branch costs one cold boot), the IndexedDB roster database is version 1 with per-record validation and eviction bounds, and the transcript snapshot store is unchanged by this PR. A browser profile without either store boots exactly as today. `ui/src/app/boot-record.test.ts` covers version and fingerprint rejection.
- **Identity binding (new, from autoreview)** — under trusted-proxy or Tailscale authentication a user switch on the same browser profile leaves the Gateway URL and stored credentials unchanged, so the previous user's cached roster could have rendered before hello. Fixed: hello-ok now reports `auth.method`, warm state is written only for `token` and `device-token` connections and is bound to a fingerprint of the credential that authenticated, and boot accepts a record only when the credential the browser can present matches. Proxy, Tailscale, password, and bootstrap identities persist nothing; the e2e suite has a trusted-proxy scenario proving the cold path.

steipete:
Merged via squash.

- Prepared head SHA: [3fc9da4b0d75c33e5f66f1d78dd3a4e7d2fdc1e5](https://github.com/openclaw/openclaw/pull/141121/commits/3fc9da4b0d75c33e5f66f1d78dd3a4e7d2fdc1e5)
- Landed commit: [24ae0fc1984c9563ebd7b4417489ddf6abd88d78](https://github.com/openclaw/openclaw/commit/24ae0fc1984c9563ebd7b4417489ddf6abd88d78)

PR #141143, opened 2026-09-07
fix(chat): keep inter-tool commentary visible once across live stream and reload

### LINKED ISSUE #138735: [Bug]: WebChat renders assistant replies twice and drops live text after page refresh (2026.9 segmented streaming)
### Bug type

### Beta release blocker

No

### Summary

On 2026.9.1, the Control UI WebChat renders each assistant reply as two identical text blocks during a turn that streams multiple text segments, and after a page refresh some live-visible text disappears. The underlying SQLite transcript stores each text exactly once, so this is a history display/reconciliation issue, not data loss.

### Steps to reproduce

1. Run gateway 2026.9.1 (commit ad6fe23).
2. Open the Control UI WebChat chat tab.
3. Send a request that makes the agent emit intermediate commentary plus tool calls (e.g. a lookup that triggers a search tool), so the reply streams in multiple text segments.
4. During the run, observe the assistant's narration text shown twice.
5. Refresh the page.
6. Observe some text that was visible during the run is now missing from the reloaded history.

### Expected behavior

Each assistant text segment renders exactly once during the run and remains as exactly one copy after a refresh/history reload. Refreshing must not remove text that was already shown.

### Actual behavior

- During the run, assistant commentary/narration segments appear twice (duplicate text blocks).
- After a page refresh, some text that was visible during the run is missing from the reloaded history.
- Verified the durable SQLite transcript (transcript_events) holds exactly one copy of each assistant text, so the defect is in the history display projection / client reconciliation, not in persisted data.

### OpenClaw version

2026.9.1 (ad6fe23)

### Operating system

Windows 11 host + WSL2 (Linux 6.18.33.2-microsoft-standard-WSL2), x64

### Install method

npm global install (npm i -g openclaw)

### Model

deepseek/deepseek-v4-flash

### Provider / routing chain

deepseek (DeepSeek API, OpenClaw default) — api: openai-responses

_No response_

### Logs

```shell

```

_No response_

### Impact and severity

High for WebChat users: every tool-using turn duplicates assistant output, and refreshing can hide previously-visible text. No data is lost (transcript stores each text once); impact is confined to the chat UI.

### Additional information

Diagnostic notes (all observed directly):
- Each assistant commentary text segment carries a textSignature {phase:'commentary'} plus a unique segment id.
- The server history projection emits, for one such transcript row, both a pure-text 'segment fallback' record (openclawStreamFallback.source='segment') and the full assistant record; the WebChat client is meant to reconcile these by itemId but they are not always collapsed, producing the duplicate text.
- docs/web/webchat.md explicitly describes this failure mode: 'If live assistant text appears during a run but disappears after history reload, check ...' which matches the refresh symptom.
- No errors observed in gateway logs; issue reproduces in the Control UI WebChat chat tab.

admin-leo:
Closes #138735

## What Problem This Solves
Fixes an issue where WebChat showed inter-tool assistant commentary duplicated during a run and then missing after a page refresh, when a turn streams multiple text segments around a tool call.

## Why This Change Was Made
The commentary fallback row emitted by the history projection shares the owning assistant row's transcript id. Client reconciliation collapsed the two rows by id, discarding the commentary text that had been split out of the full row.

This change makes client reconciliation segment-aware: a keyed commentary fallback (`openclawStreamFallback.itemId`) is treated as distinct from the full row that owns its transcript id, so both survive. The `__openclaw.id` is left untouched, so `chat.message.get` recovery and reply lookups keep resolving.

## User Impact
WebChat now shows each inter-tool commentary segment exactly once, both during live streaming and after a history reload.

### Real-instance proof (real provider + real tool turn)
The fix was run as an isolated gateway and exercised through the real Control UI with DeepSeek. The scenario is a multi-step news search ("Search the web for the top 3 AI news stories today and summarize each in one sentence"): the agent attempts `web_search`, falls back to `web_fetch` (TechCrunch / The Verge / VentureBeat RSS) plus `exec`, producing several inter-tool commentary segments (e.g. "`web_search` is unavailable right now, so let me pull from news feeds directly", "The TechCrunch feed worked. Let me grab the rest of it and check another outlet"). Each commentary segment renders exactly once, both live and after reload:

**Live streaming** (commentary + tool card + answer, commentary appearing once):

commentary rendered once — live

**After page reload** (commentary still present once — previously dropped):

commentary still rendered once — after reload

**Patched build identity** — the operator menu shows the build at commit `e679b23` (`fix/webchat-du…@e679b23`):

operator menu showing patched commit

The gateway log corroborates it: `webchat connected … build=2026.9.2-e679b23d0ad4`.

### Automated regression tests (both fail on pre-fix code, pass with the fix)
1. **Unit test** (`src/gateway/chat-display-projection.commentary-dup.test.ts`) — projects a mixed commentary+tool+answer turn and delivers the resulting fallback + full rows incrementally through the session projection, asserting the commentary survives exactly once with the full row intact and the transcript id still resolvable.

2. **Mock-gateway E2E test** (`ui/src/e2e/chat-flow.commentary-dedup.e2e.test.ts`) — runs the real Control UI in Chromium against a mock gateway and exercises the full turn lifecycle: a commentary preamble item, a tool result event, the answer delta, persistence of the fallback + full rows (sharing the same transcript id) via `session.message`, and a reload (`chat.startup`). It asserts the commentary bubble renders exactly once alongside the answer, both live and after reload. On pre-fix `main`: **1 failed**; with the fix: **1 passed**.

Local suites, all green: session-projection (153), gateway display projection + server-methods (372), plus the E2E test above.

### DIFF
--- a/packages/gateway-client/src/session-projection.ts
+++ b/packages/gateway-client/src/session-projection.ts
@@ -236,7 +236,9 @@ function entryMatches(
   allowSnapshotPromotion = false,
 ): boolean {
   if (sameTranscriptIdentity(left.identity, right.identity)) {
-    return true;
+    // A keyed commentary fallback and the full row share an id but are distinct display segments.
+    const itemId = readAssistantStreamSegmentIdentity(left.message)?.itemId;
+    return itemId === readAssistantStreamSegmentIdentity(right.message)?.itemId;
   }
   const durableEntry = left.identity?.id ? left : right.identity?.id ? right : null;
   const provisionalEntry = durableEntry === left ? right : durableEntry === right ? left : null;
--- a/src/gateway/chat-display-projection.commentary-dup.test.ts
+++ b/src/gateway/chat-display-projection.commentary-dup.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from "vitest";
+import {
+  createSessionProjection,
+  reduceSessionProjection,
+  readSessionMessageIdentity,
+} from "../../packages/gateway-client/src/session-projection.js";
+import { projectChatDisplayMessages } from "./chat-display-projection.js";
+
+const runId = "run-commentary-dup";
+
+function textOf(message: unknown): string[] {
+  const content = (message as { content?: unknown }).content;
+  if (typeof content === "string") {
+    return [content];
+  }
+  if (!Array.isArray(content)) {
+    return [];
+  }
+  return content
+    .filter(
+      (block): block is { text: string } =>
+        typeof block === "object" &&
+        block !== null &&
+        typeof (block as { text?: unknown }).text === "string",
+    )
+    .map((block) => block.text);
+}
+
+const user = {
+  role: "user",
+  content: [{ type: "text", text: "Look it up for me." }],
+  __openclaw: { id: "user", seq: 1, idempotencyKey: `${runId}:user` },
+};
+
+// A mixed turn: inter-tool commentary text, then a tool call, then the final answer.
+const mixedCommentary = {
+  role: "assistant",
+  content: [
+    {
+      type: "text",
+      text: "Searching for the answer…",
+      textSignature: '{"v":1,"phase":"commentary","id":"commentary-0"}',
+    },
+    { type: "toolCall", id: "call-1", name: "web_search", arguments: {} },
+    {
+      type: "text",
+      text: "The answer is 42.",
+      textSignature: '{"v":1,"phase":"final_answer"}',
+    },
+  ],
+  stopReason: "stop",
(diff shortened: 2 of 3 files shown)


admin-leo:
@clawsweeper re-review

VACInc:
#146554 is merged, but it does not supersede this PR. #146554 removes tool-entry-triggered rollover so an unfinished live assistant sentence keeps one growing body. This PR fixes a separate persisted/display reconciliation boundary: keyed commentary fallbacks and their full transcript row share a durable transcript ID but must remain distinct across live delivery and reload. The patch and tests cover that identity/hydration behavior, which #146554 preserves rather than replaces.

PR #141147, opened 2026-09-07
fix(anthropic): refresh Desktop metadata when watch coverage is unknown

steipete:
## What Problem This Solves

Claude Desktop session titles can remain stale when the catalog watcher reports that its coverage is unknown. The cache fast path currently treats that state like a clean directory set and returns the cached overlay.

## Why This Change Was Made

Exclude the existing all-dirty state from the fast path so the normal refresh path rescans metadata. Clean watched entries still reuse their cached result, and an absent Desktop store retains the existing sixty-second negative-cache backstop.

## User Impact

Desktop catalog metadata refreshes when its watch state requires a full scan. There is no new watcher, polling interval, preload, or package dependency.

## Evidence

- Two focused owner tests use synthetic temporary catalog files and a controlled watch-state fixture. Before the fix, Node and Bun each fail the stale-title regression while passing the absent-store TTL case; both pass both cases afterward.
- The tests retain clean-cache identity, refreshed title, and the exact 59,999/60,000 ms absent-store boundary. All bounded, isolated runs joined without a cap or surviving process.
- Fresh independent P0–P2 review passed with no actionable findings. Final two-file source and patch hashes were verified before staging.
- The full changed-file gate passed on the final two-file candidate, including qualified Doctor architecture checks and all type/lint/static checks. An initial gate found an unbound-method assertion; the final test retains a named reference to the same close spy. Focused runtime proof preceded that mechanical reference correction and was not repeated; targeted lint, fresh full-candidate P0–P2 review, and the full gate cover the corrected test. Exact d465 CI34141622949 and pull_request WorkflowSanity34141622957 passed, and hosted review is ready with no actionable findings. No separate full build was run for this condition-only change to an existing module.

This is a cache-owner repair. The separate native filesystem watcher callback gap remains unresolved; these controlled-state tests do not prove native callback delivery or full catalog-suite compatibility.

### DIFF
--- a/extensions/anthropic/session-catalog-desktop.test.ts
+++ b/extensions/anthropic/session-catalog-desktop.test.ts
@@ -33,15 +33,17 @@ describe("Claude Desktop overlay cache", () => {
   let now: number;
   let dirty: "all" | Set<string>;
   let watch: DirtyDirectoryWatch;
+  let closeWatch = vi.fn();
 
   beforeEach(async () => {
     home = await fs.mkdtemp(path.join(os.tmpdir(), "openclaw-desktop-overlay-"));
     now = Date.UTC(2026, 0, 1);
     dirty = new Set();
+    closeWatch = vi.fn();
     watch = {
       takeDirty: () => dirty,
       observeChildDirectories: vi.fn(),
-      close: vi.fn(),
+      close: closeWatch,
     };
     createWatch.mockReset().mockReturnValue(watch);
     vi.spyOn(Date, "now").mockImplementation(() => now);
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d465eb74a8da03d048bf638e91dc8bd1530f2a11](https://github.com/openclaw/openclaw/pull/141147/commits/d465eb74a8da03d048bf638e91dc8bd1530f2a11)
- Landed commit: [7fbfcac305bb5dcdbb4bc4b30617ed7a42ac786f](https://github.com/openclaw/openclaw/commit/7fbfcac305bb5dcdbb4bc4b30617ed7a42ac786f)

PR #141139, opened 2026-09-07
fix(ci): unblock lint for Telegram progress tests

vincentkoc:
Related: #141128

<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers
can help update the branch when needed.

</details>

## What Problem This Solves

Resolves a problem where regular CI is blocked by the Telegram progress-updates test exceeding the existing 1,000-line lint limit. [The check-lint failure](https://github.com/openclaw/openclaw/actions/runs/34109152279/job/101701451795) reported 1,012 counted lines while validating https://github.com/openclaw/openclaw/pull/141128; the Telegram test and lint configuration were unchanged by that PR.

## Why This Change Was Made

Consolidate only the two duplicated post-final progress-restart cases into a two-row `it.each`. Each row creates its own fixture and preserves initial tool start, awaited final delivery, then the exact awaited late tool-start or command-output callback and every assertion.

The distinct pending-final ordering test remains separate. All later cases, including fast-mode coverage, are unchanged. No production code, helpers, lint rules, caps, ignores, or test inventory files change.

## User Impact

Removes duplicated test setup while preserving the same 37 test cases. No Telegram runtime behavior changes. The existing lint limit remains enforced.

## Evidence

- Fresh base: `de844c19cc1e26c0f616f5b0d8d7d349a41a1433`. Live file history still points to https://github.com/openclaw/openclaw/commit/3612625c4ca8074981c9ddeb47a06d5e92d3e5ef.
- Native CI baseline: 1,012 counted lines. The target-file and root lint-config blobs are identical at the failing perf head, its base, and this branch's base.
- **Derived count: 996**, from removing 16 nonblank lines in a comment-free region under the unchanged rule. This is a derived unchanged-rule count, **not a passing lint invocation**.
- **Exact-head hosted lint passed** at `55e7cc99e63d5ee21222cac14080c45706aaa0ae`: [check-lint, run 34110669208](https://github.com/openclaw/openclaw/actions/runs/34110669208/job/101706106471). This supplies the native lint proof beyond the derived count.
- Focused restart checks: **3 passed**, 34 filtered; full file: **37 passed, 0 skipped**.
- Byte comparison confirms the imports/suite prefix and every byte from the separate pending-final test onward are unchanged.
- Formatting, whitespace checks, max-lines suppression ratchet, assertion-safety ratchet, and the pre-typecheck changed-file checks passed.
[...]

PR #141085, opened 2026-09-07
test(ui): finish Mermaid rendering before clipboard teardown

steipete:
## What Problem This Solves

Finishes Mermaid rendering started by clipboard fixtures before their page is torn down. Otherwise a later browser suite can remove an active renderer frame and wait for its existing watchdog before its own rendering starts.

The original renderer-mock removal from this PR has independently landed in #141092. This PR now contains only the remaining fixture-lifetime repair; the final clipboard source is unchanged from the previously reviewed combined repair.

## Why This Change Was Made

The fixture waits for the real Mermaid preview and its image decode before interaction or final retirement. The connected-source-change case joins rendering after its existing clipboard race assertions. Final `pagehide` teardown disposes the settled renderer through its existing lifecycle.

Component disconnection already fences result publication, but does not finish rendering queued in the shared renderer. Joining that work at its test fixture prevents it from outliving the fixture.

## User Impact

Browser suites avoid waiting behind unfinished clipboard-fixture rendering. Clipboard transports, race assertions, deadlines, and production behavior are preserved.

## Evidence

- The original five-file CI order reproduced 100 passed / 6 failed with the module mock. Its removal fixes those failures, as landed in #141092.
- With mock removal alone, the first native case in that same order took 12.765 seconds while prior rendering waited for its watchdog. The fixture-lifetime repair passes all 106 cases with that first native case at 319 ms. This is an observed regression sequence, not a general performance benchmark.
- The minimal clipboard → native pair passes all 27 cases with the final source, on both the independently refreshed main base and the original PR base.
- Selected changed checks and fresh independent P0–P2 review pass. No new mock, production seam, timeout increase, or isolation setting was added. Clipboard stubs remain inside the browser page and do not write the OS clipboard.
- The rebase preserves the exact reviewed clipboard source. The original five-line removal is omitted because it is now on main.

CI run 34107394739 attempt 1 stopped in checkout. Attempt 2 passed all three UI browser shards, all thirteen mocked UI E2E shards, and the real-Gateway UI suite, but found a separate Telegram test-file line-limit failure on its main base. That lint repair independently landed in #141139 and is included in this branch. Final [CI run 34112456447](https://github.com/openclaw/openclaw/actions/runs/34112456447) passed on `a45a33a1feccf0ffdd4d1cba76b899e0be934306`: 38 successful jobs, 16 workflow skips, and the required CI gate green. The current-head ClawSweeper review has no findings or before-merge requests.

PR #141047, opened 2026-09-07
fix(ui): keep reopened select menus visible

steipete:
## What Problem This Solves

Fixes an issue where users reopening a Control UI select during its closing animation would see an expanded control with no visible options. This can affect the Automations model picker when its catalog refreshes between closing and reopening. An already-open select could also leave its menu visible after being disabled.

## Why This Change Was Made

Give each accepted transition ownership of its listeners, animation, focus updates, and completion. A replacement retires and joins the previous animation, so an old close cannot hide a reopened menu. Public `show()` and `hide()` calls join that work, including interrupted calls, vetoes, repeated requests, and first-render initialization.

Disabling invokes the same transition owner directly: its callback runs after the open watcher, so merely setting `open = false` during that update never closed the popup. Disabled teardown proceeds even when ordinary closing is vetoed; enabled cancellation still works.

This extends the existing, approved Web Awesome 3.12.0 patch. Both distributions receive the same repair, and the public declarations and metadata remain unchanged. The distributed runtime grows by six lines overall while replacing the paired listener-registration helpers.

## User Impact

Reopened options remain visible and selectable, and disabled menus close. Keyboard selection, focus return, multiple selection, clearing, and dismissal retain their existing behavior.

## Evidence

- Before/after component proof reproduces hidden reopening, unsettled method calls, stale completion, and disabled-menu visibility. Twelve lifecycle cases fail against the original package while its two ordinary behavior controls pass; the additional disabled-owner failure was reproduced separately before its repair.
- All 65 select and existing dropdown browser cases pass, plus four picker rendering tests and three mocked production-bundle Automations E2E cases. The original catalog-refresh/draft-preservation test is unchanged.
- The final visibility assertions inspect retained DOM elements. This fixes Vitest's inability to re-resolve a hidden option through a role locator while preserving the actual visibility check. No timeout or retry was increased.
- Existing patch hunks, both distributed implementations, public type/metadata parity, and the lockfile's patch-hash-only change are verified.

The captures show the same synthetic interrupted-close scenario in Chromium. The original CI interleaving is not independently established. Real-profile Chrome control was unavailable; the browser evidence is component and mocked application proof.

[...]

PR #141150, opened 2026-09-07
test(telegram): deduplicate post-final progress coverage

steipete:
## What Problem This Solves

The Telegram progress test file exceeds the existing 1,000-line lint limit, blocking unrelated PRs. The failure is visible in [CI run 34107394739, check-lint](https://github.com/openclaw/openclaw/actions/runs/34107394739/job/101699488888).

## Why This Change Was Made

Consolidates the two post-final progress cases into a two-row table. Both still start progress, finish delivery, send their original late callback, and assert that the draft is not restarted. The pending-final case stays separate because its delivery ordering differs.

This removes 19 test lines while preserving both callback payloads, all assertions, and the complete remainder of the file. No production code, lint limit, suppression, or timeout changes.

## User Impact

Restores the lint gate without changing Telegram behavior or removing regression coverage.

## Evidence

- Targeted lint reproduces the original 1,012 counted lines against the 1,000-line limit.
- All 48 progress-update and lifecycle sibling cases pass before and after consolidation.
- Targeted lint and all 19 selected changed-file checks pass, including extension test typechecking and dead-export scans.
- Fresh independent P0–P2 review is scoped-clean with no findings.

Exact-head hosted CI remains required before native landing.

### DIFF
--- a/extensions/telegram/src/bot-message-dispatch.progress-updates.test.ts
+++ b/extensions/telegram/src/bot-message-dispatch.progress-updates.test.ts
@@ -30,63 +30,44 @@ import {
 } from "./bot-message-dispatch.test-harness.js";
 
 describeTelegramDispatch("dispatchTelegramMessage progress-updates", () => {
-  it("does not restart progress drafts after final answer delivery", async () => {
-    const { answerDraftStream } = setupDraftStreams({ answerMessageId: 2001 });
-    dispatchReplyWithBufferedBlockDispatcher.mockImplementation(
-      async ({ dispatcherOptions, replyOptions }) => {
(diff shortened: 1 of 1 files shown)


steipete:
Fixed independently by @vincentkoc in #141139, merged as `908e5441f456ae327f059601b87a8b78ac32a395`. I verified it is on current main and compared the complete test files: the only difference from this PR is the table parameter name (`event` versus `lateCallback`). Both callback payloads, ordering, assertions, and the rest of the file match.

Closing this redundant PR. The independent landing resolves the lint prerequisite for #141047 and #141085; our before/after 48-case runs and 19 selected checks remain supporting evidence for the same consolidation. Thanks for landing the fix.

PR #141152, opened 2026-09-07
fix: expose the live voice shortcut description on iOS

shakkernerd:
## What Problem This Solves

Resolves a problem where the **Start Live Voice** shortcut description is missing when Apple reads its `AppIntent` metadata.

## Why This Change Was Made

Declare the description as `IntentDescription?` to satisfy Apple’s protocol requirement. The inferred nonoptional property instead left the default `nil` implementation in use. Add a regression assertion through `any AppIntent.Type` and run the existing app-delegate suite in the focused iOS CI job.

## User Impact

Apple’s shortcut metadata consumers receive the intended description: “Open the current chat in OpenClaw and start a voice conversation.” The shortcut continues using the existing current-chat voice-start path.

## Evidence

- Actual-source Apple SDK probe: the original declaration returns `nil` through `any AppIntent.Type`; the repair returns the intended description text.
- Swift 6 compilation and iOS 18-target typechecking against the iOS 26.5 SDK pass. Changed-file SwiftFormat and strict SwiftLint checks pass.
- The existing suite covers cold launch, active-scene admission, warm invocation, and setup/capture rejection. Native iOS CI, Periphery, and review status are reported on this PR.
- Shortcuts screenshots are omitted because the available environment has no iOS simulator. The before/after protocol probe verifies the metadata boundary directly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5070,6 +5070,7 @@ jobs:
             -only-testing:OpenClawTests/LocationServiceCallbackTests \
             -only-testing:OpenClawTests/LocationServiceOrderingTests \
             -only-testing:OpenClawTests/NodeAppModelInvokeTests \
+            -only-testing:OpenClawTests/OpenClawAppDelegateTests \
             -only-testing:OpenClawTests/WatchMessagingInboundTransportTests \
             -only-testing:OpenClawTests/WatchSessionActivationGateTests \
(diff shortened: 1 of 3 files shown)


shakkernerd:
Fixed: the iOS Start Live Voice shortcut now exposes its intended description through Apple’s AppIntent metadata contract.

PR #141166, opened 2026-09-07
refactor(plugins): share npm version queries

steipete:
## What Problem This Solves

Plugin installs and updates duplicate the npm version request, JSON parsing and semver filtering. The same registry handling has two implementation owners.

## Why This Change Was Made

Move the complete query/parser into the existing install-source utility module already used by both callers. Install continues to pass its AbortSignal and explicit process-tree termination; update continues to omit those options. Callers retain their version-selection and compatibility decisions. The npm environment helper becomes private now that all its callers are local.

Remove the impossible missing-process-result guard. Four existing test cases now supply a typed command-failure result instead of relying on an unspecified mock return; their assertions are unchanged.

## User Impact

No intended behavior change. Command arguments, timeout flooring, environment, error propagation, version order/filtering and caller-owned selection are preserved. The change removes 24 production code lines and adds 12 test-fixture lines, for a net reduction of 12 maintained code lines.

## Evidence

All 359 existing owner/caller tests pass before and after; they also pass after refreshing onto repaired main. An external differential probe matches 377 cases for output, rejection, command arguments and exact option presence, using the real semver predicate. The unsupported undefined-mock difference is explicitly excluded from production parity. All 29 selected gate stages passed and final independent Codex review was clean through P2. The refresh preserved all four changed file hashes and the complete changed-line payload. No live registry or package-build result is claimed for this source-only consolidation.

### DIFF
--- a/src/infra/install-source-utils.ts
+++ b/src/infra/install-source-utils.ts
@@ -13,7 +13,7 @@ import { resolveUserPath } from "../utils.js";
 import { resolveArchiveKind } from "./archive.js";
 import { pathExists } from "./fs-safe.js";
 import { applyNpmFreshnessBypassEnv, type NpmProjectInstallEnvOptions } from "./npm-install-env.js";
-import { resolveNpmJsonEntries } from "./npm-registry-spec.js";
+import { isExactSemverVersion, resolveNpmJsonEntries } from "./npm-registry-spec.js";
 import { withTempWorkspace } from "./private-temp-workspace.js";
 import { resolvePreferredOpenClawTmpDir } from "./tmp-openclaw-dir.js";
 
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9560cbd8f26867964e1fcaff81e86eb992279285](https://github.com/openclaw/openclaw/pull/141166/commits/9560cbd8f26867964e1fcaff81e86eb992279285)
- Landed commit: [4b1195fb3566a79a8a644e692b4b7f7f7ac26546](https://github.com/openclaw/openclaw/commit/4b1195fb3566a79a8a644e692b4b7f7f7ac26546)

PR #141137, opened 2026-09-07
fix(ci): avoid finalization fixture startup stalls

steipete:
Related: #140648
Related: https://github.com/openclaw/openclaw/pull/141128

## What Problem This Solves

Resolves repeated cold subprocess startup in the update-finalization process tests. On [main CI](https://github.com/openclaw/openclaw/actions/runs/34102454624/job/101680351564), the one-second prerequisite phase expired before the intended snapshot phase, then the test waited for a marker that could never arrive until its 100-second deadlock guard fired.

The refreshed CLI fix also encountered the existing planner growth-test timeout. Conversely, the separate planner fix encountered the CLI prerequisite failure. This PR now combines both already-reviewed repairs so the same CI run can validate them together.

## Why This Change Was Made

The test runner already compiles the real SQLite worker once per invocation. Forward its resolved entrypoint metadata through the private fixture arguments so nested workers reuse that generation instead of starting another TSX source graph. The fixture forwards metadata values because a bundled `currentModuleUrl` can identify a shared chunk, not a named-export entry module.

Observe the snapshot marker synchronously and assert the captured ledger state after child exit and both output EOFs. Observation no longer adds a separate promise that can outlive a failed child.

The planner fixture retains real discovery within its single test while rebuilding every planner snapshot. It caches only raw tracked-file reads, returns fresh filtered/injected arrays, and preserves all ten inventories and both mock cleanups. The production planner, job caps, weights, worker counts, prerequisites and 120-second timeout are unchanged.

## User Impact

Faster and more reliable CI. Production update behavior is unchanged. All 14 original scenarios, real SQLite ownership checks, phase budgets, subprocess cleanup, JSON routing, ledger assertions and reduced shard counts remain. The planner guard keeps its full assertions instead of raising its deadline or reducing coverage.

## Evidence

The first four bullets retain the original contributor's local evidence.

- Unmodified full suite: 12 passed / 2 failed, 568.26s command. `phase-hang` took 101.745s; both deadline cases failed in the earlier ownership-check phase.
- Repaired full suite, same original order: all 14 passed, 359.98s command. `phase-hang` took 24.778s, including cold CLI startup. These are observed local timings, not a controlled whole-CI forecast.
- Removing only worker-metadata forwarding, while retaining finite observation, restored both failures: missing snapshot-ledger observation in 17.631s and an earlier ownership-phase timeout in 17.221s. Restoring forwarding passed both intended phase contracts with the unchanged budgets.
[...]

PR #141170, opened 2026-09-07
fix(slack): keep progress updating after native streams expire

shannon0430:
<details>

</details>

## What Problem This Solves

Fixes an issue where users watching a Slack native progress card see frozen or failed-looking progress after Slack expires the stream, even though the agent continues working. Later text can fail or arrive in a separate reply.

## Why This Change Was Made

Keep a bounded projection of the streamed message and use `chat.update` on the same timestamp after an explicit `message_not_in_streaming_state` rejection. Subsequent progress, final text, and final metadata use that same message, following [Slack SDK recovery guidance](https://github.com/slackapi/node-slack-sdk/pull/2611).

The existing scoped writer and user Stop handling remain in control. Ambiguous writes are not replayed. If the full message cannot be reconstructed within the limits, or an update definitively rejects pending text, only the undelivered tail uses normal chunked delivery. This includes `edit_window_closed` and `cant_update_message`, so workspace editing restrictions do not discard an otherwise deliverable final reply. Once ordinary delivery succeeds, its retained recovery projection is retired, so native cleanup cannot replay that delivered tail onto the original card.

## User Impact

Progress and eligible final replies continue on the existing card after stream expiry. Task state, source links, narration order, and accumulated detail/output text are retained. Recovered task details and outputs use plain rich text, so embedded Markdown syntax is not interpreted.

## Evidence

- Before the fix, a real-SDK regression failed on the first expired append with `SlackStreamNotDeliveredError`.
- On final head `76bd69a7f9a3bb43d9befaf29aac2ab44225f23f`, all **301 tests** passed across streaming, real-SDK dispatch delivery traces, preview fallback, and progress blocks. The edit-restriction dispatch cases failed before the follow-up fix and now pass, with exactly one pending-tail delivery and no replay of the acknowledged prefix. Another pre-fix failing dispatch regression covers native rejection, successful ordinary tail delivery, and cleanup expiry; success and ambiguous cleanup variants also pass without replay. When an ordinary post is accepted but its response is lost, the session remains stopped, the original error propagates, and no subsequent native cleanup or update is attempted.
- The complete six-file `check:changed` gate passed (exit 0), including production/test types, lint, guards, dead-export scans, and documentation checks.
- The base includes upstream [#141137](https://github.com/openclaw/openclaw/pull/141137), which repairs the two unrelated fixture timeouts from the previous CI run.
- Slack `blocks.validate` accepted synthetic plan and timeline projections, including rich-text details/output and source elements.

### Real Slack recovery proof

Executed the actual compiled `streaming.ts`, `streaming-projection.ts`, and their client modules from measured revision `2fec06eb352ffb4fc89178b5130aa594c7c1a4d0` with the real Slack SDK **8.0.0** and installed public OpenClaw SDK modules. API responses were not mocked. A separate `chat.stopStream` call closed only each synthetic test stream on Slack, leaving the SDK stream object active, so the next real append exercised `message_not_in_streaming_state`.

```text
Baseline owner (4d7a33874):
  chat.startStream -> ok
  chat.stopStream  -> ok (test-induced server closure)
  chat.appendStream -> message_not_in_streaming_state
  readback: rejected tail absent; no recovery update

Candidate owner (2fec06eb):
  chat.startStream -> ok
  chat.stopStream  -> ok (test-induced server closure)
  chat.appendStream -> message_not_in_streaming_state
  chat.update -> ok; readback: same message, task in_progress, working text present
  chat.update -> ok; final answer and complete task appended
  chat.update -> ok; final metadata acknowledged

Independent native-CLI readback:
  expected bot identity: true
  same message: true
  all tasks complete: true
  acknowledged prefix occurrences: 1
  final answer occurrences: 1
  final metadata preserved: true
```

Task details, output, and source links also survived the recovery. The executed bundle hashes matched after the run. The follow-up change only retires the projection after ordinary fallback succeeds; the real-API recovery path measured here remains unchanged. The new fallback-cleanup boundary is verified by actual-dispatch tests with the real SDK and controlled API rejections. Only synthetic content was used; identifiers and private environment details are omitted here.

This proves the real server-closed-stream recovery contract and Slack message readback. It does not measure Slack's undocumented natural expiry duration or claim screenshot-based rendering verification. The installed production plugin and Gateway were not replaced or restarted.


shannon0430:
Automated author follow-up checkpoint:

- Current head: `b0c846f83b14a8f5cbb8465e70e89fdf8d50cdbd`.
- Current blocker: `CI/check failure: checks-node-compact-small-51, checks-node-compact-small-31, openclaw/ci-gate`.
- Labels: `docs, channel: slack, size: L, P2, rating: 🦐 gold shrimp, merge-risk: 🚨 message-delivery, status: 📣 needs proof`.
- `maintainerCanModify`: `true`.

For `fix(slack): keep progress updating after native streams expire`, author-side work remains pending. Follow-up and review requests belong in this pull request. This checkpoint does not modify source code.

shannon0430:
Addressed the review on `2fec06eb352ffb4fc89178b5130aa594c7c1a4d0`:

- `edit_window_closed` and `cant_update_message` now preserve ordinary delivery of the pending tail. Real-SDK dispatch regressions failed before the fix and pass afterward, with one tail delivery and no acknowledged-prefix replay.
- The PR body now includes real Slack API recovery evidence from the compiled owner modules: the baseline fails after server closure; the candidate updates the same message through working progress, final text, and metadata. Independent native-CLI readback confirmed completed tasks, the same bot/message, and one copy of the prefix and final answer. No API responses were mocked.
- All 297 owner/sibling tests and the complete changed-file gate passed. The base includes #141137, the upstream repair for the two prior CI fixture timeouts; fresh-head CI is running.

The proof uses an explicitly closed synthetic stream to obtain Slack's real `message_not_in_streaming_state` response; it does not claim a measurement of natural timeout duration or screenshot verification. Public evidence omits private identifiers.

@clawsweeper re-review

shannon0430:
Addressed the fallback-cleanup finding. Once normal tail delivery succeeds, its recovery projection is retired before native cleanup. The regression fails before the fix because cleanup expiry adds an extra chat.update; after the fix the tail appears only in the normal reply. Success, expiry, and ambiguous cleanup outcomes are covered.

Also verified the accepted-post/lost-response boundary: the session was already stopped before normal delivery, and without acknowledged fallback success it stays stopped. The original transport error propagates; no subsequent cleanup stop or recovery update is attempted.

The existing real Slack proof remains pinned to its measured revision; the follow-up changes only post-fallback retirement. Trace evidence helpers were moved without changing their behavior to preserve the existing file-size limit. Validation: 301 tests passed and the complete six-file check:changed gate passed.

@clawsweeper re-review

PR #141126, opened 2026-09-07
improve: reduce overhead when parsing images in replies

steipete:
## What Problem This Solves

Repeated replies containing Markdown images incur avoidable parsing work, particularly when processing many short messages. This is a performance cleanup; no broken image-extraction behavior is claimed.

## Why This Change Was Made

Prepare the private MarkdownIt configuration lazily once, while keeping the reference environment, source offsets, collector, and result array local to each document. The cached rule retains no document callback. Remove the repeated configuration path without changing image eligibility, URL spelling, reference handling, or media policy.

Production delta is +20 lines; the added boundary separates reusable parser configuration from ephemeral document state. The preservation test adds 18 lines and no production testing seam.

## User Impact

Image parsing does less repeated setup while preserving extracted attachments and visible text. The direct one-image scanner improves 26.35–49.63% in this fixture, but the complete media-split one-image case is mixed (+6.04% / -8.58%). Larger composed cases improve only 0.16–2.00%; those gains remain useful alongside the larger short-message and syntax-control improvements. This is not a claim of an overall Gateway or model-response speedup.

## Evidence

- The new A/B/A document-isolation test passes on the original scanner. It covers distinct reference environments, offsets, and earlier returned arrays remaining unchanged.
- Candidate: 128/128 tests in the complete image-span and media-parse files; 28/28 normal changed checks; independent managed P0–P2 review found no actionable issues.
- One initial normal check rejected the location of a safety comment. The comment was moved directly above its assertion; the fresh full check passed. No guard was weakened.
- Fixed B0/C0/C1/B1, Node 26.8.1 and installed MarkdownIt 15.0.1, using actual `findMarkdownImageSpans` and `splitMediaFromOutput`. All 7,632 calls across 6,784 operations match the frozen full-output fixtures, including every first call and warmup. An independent saved-data audit checked the complete graphs, A/B/A witnesses, arithmetic, source/dependency hashes, and process closure.
- No external image fetch, channel delivery, or Gateway turn is measured here. The broader campaign's separate live OpenAI Gateway baseline is not substituted for parser evidence.

The unchanged composed owner retains its line-length and per-line match limits. Numeric fixtures use separate lines for the 256 images and do not claim to exercise those thresholds; the full existing media tests remain intact.

<details>
<summary>All paired batch medians and tradeoffs</summary>

[...]

PR #141075, opened 2026-09-07
fix: avoid false CI placement failures when tooling splits

brokemac79:
Closes #141068. Related: #140512, #140664, #140929. Affected: #138953.

## What Problem This Solves

Resolves a problem where adding tests causes the compact CI planner's growth regression to report that a non-tooling job moved, even though the affected job is tooling and retains its execution requirements. An unsplit `core-tooling-1` group was incorrectly included in the non-tooling comparison; its valid split name was excluded.

## Why This Change Was Made

Use the existing numbered-tooling classification consistently for the shared test-only placement helper. Add one focused regression covering split/unsplit names, retained anchors, negative controls for anchor removal and policy changes, and input nonmutation.

Only `test/scripts/ci-node-test-plan.test.ts` changes. Production planning, the existing 52 tests and their invariant assertions, inventory filters and Mantis test membership remain unchanged. This follows the earlier helper and packing work by Vincent Koc (@vincentkoc) and Peter Steinberger (@steipete); #140929's upstream correction is preserved.

## User Impact

Contributors should no longer receive this false CI regression when tooling inventory grows. No application behavior, job cap, runner allocation or timeout changes.

## Evidence

Pinned base: `15e5cf111d5954d9b3b001b1dae5b1ff8e987a69`. Committed head: `dca59eb5a02b1fd3d9c8fa3b666f71c38fbae23a`, whose tree is byte-identical to the exercised candidate `a6ff29dd8d22cfda9f76e0fdaf7ebc0adfe41097`.

### Real Behavior Proof

Executed controls against the unmodified base:

- Main: 52/52 passed (`run_6d0b01ab4a45`). Main plus six byte-identical Mantis test files: 51/52 passed (`run_95f0d0e9d7a5`), failing the original placement comparison—not an 81/80 overflow.
- Bounded external diagnostic `run_87f00249dbdc` reproduced the original fixture filter without changing production or test source. Adding ten growth-probe files kept both plans at 80 jobs.
- The affected `pr-merge-outcome.test.ts` remained exact-once and alone: same 4-vCPU runner, two workers, configs/environment/prerequisites and 330-second estimate. Only its tooling split identity and check name changed.
- Comparison preserved all 150 genuine non-tooling anchors and their policies. Coverage increased by exactly ten files. Numbered prerequisite anchors retained their memberships and policies; canonical timing keys were recomputed against each complete sibling family, not ignored.

Fresh Docker-backed proof:

[...]

PR #140664, opened 2026-09-07
fix(ci): absorb tooling inventory growth under the compact job cap

steipete:
Related: https://github.com/openclaw/openclaw/pull/140622

## What Problem This Solves

Fixes an issue where adding tooling tests could exceed the compact GitHub planner's 80-job cap. Hosted tooling growth could also change which unrelated non-hosted groups share a job.

## Why This Change Was Made

Preserves @steipete's original bounded-tail repair: when the initial GitHub pull-request plan overflows, repartition final divisible tooling tails once into parts bounded by half the existing 150-second budget. Full chunks remain intact, and the splitter regenerates membership-bound timing identities before the existing packing checks run.

The maintainer follow-up constructs non-hosted anchors independently before placing hosted tooling. The generic first-fit algorithm, bounded tail retry, timing floors, build costs, runner requirements, worker limits, serial isolation, 80-job cap, and 10-group limit remain unchanged. Growth tests compare complete non-hosted placement.

The branch preserves the original contributor commits and integrates main snapshot `9d689fa1c21fc0dc9fc280cc6df95ae212c43f3d` through an ordinary merge. This includes the canonical Git-materialization test repair from https://github.com/openclaw/openclaw/pull/140771 (`6223cf8b74800e22b85f16c71cfd002965d426e0`), without duplicating that owner's repair.

The changed-plugin regression now compares sorted per-file policy multisets only for eligible GitHub hosted-tooling partitions, while validating complete timing families separately for each plan. It preserves coverage, runner/build/environment policy, and every other descriptor comparison without treating a valid repartition's physical allocation or timing keys as invariant. Timing keys must match the emitted group's family and ordinal as well as its regenerated membership.

## User Impact

Allows modest tooling inventory growth within the existing CI budget without shifting non-hosted placement. No runtime behavior changes.

## Evidence

Current published head: `bc2eb9705386029fde3415cdede538e8574e384c`.

### Original Bounded-Tail Contribution

@steipete reproduced the 81-job failure and recovered 80-job plans, including the actual npm-lock-report filename and ten-file growth. The original checkpoint reported all 51 owner tests passing, `pnpm check:changed` passing, and scoped-clean review through P2.

Original-head [CI 34076723175](https://github.com/openclaw/openclaw/actions/runs/34076723175) succeeded with 77 successful jobs and 16 skips; isolated native build/check also passed. These are historical results, not new-head gate claims.

### Independent-Anchor Follow-up

[...]

PR #140549, opened 2026-09-07
fix(ci): isolate compact planner inventory growth fixture

vincentkoc:
Related: https://github.com/openclaw/openclaw/pull/140512
Related: https://github.com/openclaw/openclaw/pull/137893

## What Problem This Solves

Fixes an issue where the compact planner inventory-growth regression duplicated its fixture filename when another PR added that file for real, causing a false `81 planned` failure even though the real merged inventory remained within the 80-job cap.

## Why This Change Was Made

The test now constructs its baseline and one-file growth inventories explicitly. Each inventory uses a separate dynamic import because the planner initializes `SPLIT_NODE_SHARDS` at module load: the baseline removes the growth path, and the growth case adds exactly one copy.

All existing planner policy assertions remain unchanged, including the 80-job and 150-second limits, exact-once coverage, family separation, runner rank, build/dist policy, non-tooling placement, and deterministic mixed-runner ordering.

This test-only PR is policy-compliant: `main` is green for this case, while #137893 turns the fixture's formerly synthetic filename into a real tracked path and exposes the duplicate-fixture bug.

## User Impact

Maintainers can validate legitimate tooling inventory growth without a false compact-plan cap failure. Production behavior and the planner algorithm are unchanged.

## Evidence

- Pre-fix controlled reproduction with the #137893 path present: focused test failed with `compact github node test plan exceeds 80 jobs (81 planned)`.
- Focused regression after the fix: 1 passed.
- Full planner suite: 51 passed.
- `oxfmt`, targeted `oxlint`, and `git diff --check`: passed.
- `check:changed`: all reachable guards passed; local test-root `tsgo` stopped only because linked-worktree declaration inputs resolve through the shared install outside the checkout. Exact-head CI covers that lane.
- Fresh autoreview through P2: scoped clean.
- LOC: production 0; tests +73/-65.
- Codex hard gate: sibling Codex `5f49aba876922d6f2f55caa153bbb0ed1b46feba` was inspected at `codex-rs/core/src/tools/registry.rs`, `codex-rs/core/src/session/mcp.rs`, and `codex-rs/core/src/tools/context.rs`; no planner or runtime coupling exists.

PR #141171, opened 2026-09-07
fix(ci): reduce timeouts in tooling inventory checks

steipete:
## What Problem This Solves

CI can time out while checking that hosted tooling remains within its job cap as the inventory grows. The growth test reached 121.863 seconds against its existing 120-second deadline on a two-CPU Linux runner, blocking unrelated PRs.

Related: #141126, #141047. This preserves the inventory-isolation, expanded-growth, and projection coverage from #140549, #140664, and #141075.

## Why This Change Was Made

Keep the real unit-fast discovery namespace alive for this one fixture's ten planner evaluations. That module reads its own unchanged Git/source snapshot; it does not consume the separately mocked tracked-file list that supplies the growth inputs. Re-importing the planner can therefore keep observing each new synthetic tooling inventory without repeatedly rebuilding unrelated discovery caches.

All ten real planner evaluations and their original assertions remain. The original per-evaluation mock cleanup stays, and an outer finally removes both module mocks and resets modules even on failure. There are no production changes, new testing seams, timeout increases, or changes to runner, job-cap, selection, or execution policy. The test grows by ten lines after formatting.

## User Impact

The inventory-growth check does less redundant work while continuing to verify complete coverage, placement, runner compatibility, concurrency, and job-cap constraints. Local proof reduces the affected case by 35.6%; the same Linux CI lane still needs to validate the resulting margin before landing.

## Evidence

The failed CI source at merge `85713c6ecf56db7e1e08951ae666b26f51e52161` was retrieved and verified byte-identical to this baseline test. The original failure is in [run 34108956395](https://github.com/openclaw/openclaw/actions/runs/34108956395), job `checks-node-compact-small-29` / `core-tooling-15-hosted-3`. Another unrelated PR independently encountered the same timeout in [run 34112443896](https://github.com/openclaw/openclaw/actions/runs/34112443896).

One baseline and one candidate whole-suite run used the same local Node 26.8.1 environment, original test order, and unchanged 120-second per-test deadline. Both passed all 53 tests. The existing unit-fast sibling suite passed all 14 tests; all 15 selected normal checks passed. Managed independent P0–P2 review found no actionable issues.

| --- | ---: | ---: |
| Affected growth case | 21.892 s | 14.104 s |
| Whole test command | 64.292 s | 55.542 s |
| Child user CPU | 70.431 s | 63.267 s |
| Child system CPU | 3.876 s | 2.819 s |
| Kernel child peak RSS | 1,376,436,224 B | 1,391,296,512 B |
| Sampled tree peak RSS | 1,857,175,552 B | 1,865,449,472 B |

The larger local Mac did not reproduce the Linux timeout, so these are diagnostic before/after observations, not a claim of equivalent capacity or a statistical speedup guarantee. All test names, statuses, durations, failures, and resources are retained. Thirty-one other test durations increased, and both RSS measures increased; no general memory benefit is claimed. The existing ten policy/coverage checks are preserved by source and execution, but this reporter does not serialize their internal plan objects.

### DIFF
--- a/test/scripts/ci-node-test-plan.test.ts
+++ b/test/scripts/ci-node-test-plan.test.ts
@@ -2249,75 +2249,35 @@ describe("scripts/lib/ci-node-test-plan.mts", () => {
         vi.resetModules();
       }
     };
-    const baseline = await createPlanWithInventory(false);
-    const baselineToolingFiles = baseline
-      .flatMap((job) => job.groups)
-      .filter(isNumberedToolingGroup)
-      .flatMap((group) => group.includePatterns ?? []);
-    const grown = await createPlanWithInventory(true);
-    const toolingGroups = grown.flatMap((job) => job.groups).filter(isNumberedToolingGroup);
-    const toolingFiles = toolingGroups.flatMap((group) => group.includePatterns ?? []);
-    const crossRunnerHostedJobs: CompactNodeTestShard[] = [];
-
-    expect(grown.length).toBeLessThanOrEqual(80);
-    expect(new Set(toolingFiles).size).toBe(toolingFiles.length);
-    expect(toolingFiles.toSorted()).toEqual(
-      [...baselineToolingFiles, inventoryGrowthFile].toSorted(),
-    );
-    expect(nonToolingPlacement(grown)).toEqual(nonToolingPlacement(baseline));
-
-    for (const job of grown) {
-      const hostedToolingGroups = job.groups.filter(isHostedToolingGroup);
-      if (hostedToolingGroups.length === 0) {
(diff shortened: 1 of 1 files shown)


steipete:
Superseded by #141137, which incorporates Vincent Koc’s earlier planner repair from #141128 together with the CLI fixture startup fix.

Verified merge ed86b491cce2d084dea375fb700c47e65f1b1913 on current main, exact equality of all three reviewed source files, retained contributor credit, and successful combined-head CI run 34114761822. This PR’s local experiment and its results remain documented above; landing another overlapping implementation is unnecessary.

PR #138839, opened 2026-09-05
fix(update): validate candidates before stopping the Gateway

steipete:
## What Problem This Solves

An already-current update could stop the managed Gateway, staging or plugin finalization could keep it unavailable for most of an update, and package rollback could restore files without restoring a running Gateway.

Closes #136997. Related: #137580, #124396.

PR D prepares and validates package/Git candidates while serving, skips unchanged artifacts before the stop boundary, and confirms activation or restores a verified previous generation when live state remains compatible. It follows merged #138690 and #138737; automatic repair integration belongs to PR E.

## Why This Change Was Made

The updater stopped the service before knowing whether a viable candidate existed, and restoring files did not prove the previous Gateway was serving. One transaction owner now controls activation, retained generations, verification and recovery.

## User Impact

Candidate validation, isolated state/config rehearsal and plugin download/sync run while serving. Required live Doctor migrations use exclusive maintenance ownership. Changed plugins need a second measured maintenance/restart window; unchanged plugins skip full Doctor but still receive fresh target config validation and lint. Unknown health stays unknown; inference is advisory and excluded from measured outage.

Package rollback requires unchanged config content and database schemas, a previously verified generation, and native manager safety. It preserves the previous Node executable/service definition and verifies the restored Gateway before reporting `serviceRunning`. Native late rollback refuses changed siblings. Git restores source/runtime only before live migrations; later failure retains the candidate. A native refusal before the live swap restarts the verified unchanged Gateway and preserves concurrent sibling changes; late rollback continues to refuse changed siblings.

Confirmed downgrades to targets that predate migration continuation retain the shipped CLI contract: runtime validation is recorded as unavailable, and the existing finalization path runs. A present but invalid continuation contract still fails, and target database-schema preflights remain in place. These older targets do not receive the new automatic schema-neutral rollback guarantee.

Bun on Windows retains its pre-stop refusal and now prints the resolved manual install, Gateway restart, and update-status commands. Bun is an opt-in runtime, so this uses actionable manual recovery without adding an in-place fallback stack.

Failure recovery without a managed-service context also preserves the invocation environment during config verification, keeping rollback on its intended state root.

[...]

PR #141175, opened 2026-09-07
fix(update): installed plugins fail candidate validation

steipete:
Related: #138839

## What Problem This Solves

Fixes an issue where users updating OpenClaw with installed plugins could fail candidate validation before activation because the staged host could not load those plugins. Flattening pnpm symlinks also broke transitive dependency resolution. On case-insensitive volumes, a configured filename with different casing could invalidate its copied entry.

## Why This Change Was Made

Prepare one disposable plugin generation alongside the SQLite snapshot, copying payloads with their dependency/package roots and carrying one locator map into copied install records and configuration. Preserve path aliases and public filenames, retain an existing same-identity file, and bind only copied host links to the staged candidate; serving files, links and database bytes remain unchanged. Preserve pnpm store topology with the existing runtime relocation owner. Complete stores or node_modules owners may include unused dependencies; ordinary external file links remain exact file copies.

## User Impact

Installed plugins can be validated against the candidate while the previous Gateway keeps serving. Candidate repairs operate on private copies, and post-swap schema inspection retains its existing response contract.

## Evidence

- 45 focused tests passed on the main forward-port: 28 projection/canary, 8 file-identity and 9 worker URL/argv cases, including real Node imports, dependency resolution, source immutability, aliases, case-insensitive filename preservation and worker launch from another directory.
- CI exposed an outdated snapshot-only CLI fixture. Updating that one producer line restored all 9 staged-repair cases; the complete 413-test update-cli owner suite then passed with assertions unchanged. Version inspection remains separately mocked with its array contract.
- Full changed checks and independent P2 reviews passed for the source repair and fixture correction. Main's newer absolute-tsx preload behavior is preserved.
- The original projection repair in rebuilt release candidate (`d9287bbc620390781584ce75d82572f6c37dcfc7`) passed migration rehearsal, Doctor lint, config validation, plugin resolution, migration-continuation checks and the packaged gateway canary. The canary exited successfully in 26.8 seconds.
- The topology follow-up passed 37 candidate/native-stage tests, then 38 after resolving the main schema-metadata conflict, including real-worker relative/absolute pnpm links, external stores, dependency cycles, executable launchers, copied transitive mutation isolation, and an external-file sibling-traversal regression. Each new regression failed its defective predecessor.
- Broader unchanged Git-staging fixtures passed 45/49 cases; four failed during Git setup before updater execution. One previously failing case passed unchanged in isolation with the original timeout. This limitation remains recorded rather than widening fixture deadlines.
- Packaged proof above covers the original projection repair; the topology follow-up has source-level proof and still needs rebuilt package qualification. This is not full-release green. Final PR head `8b8b3f368eff2b2a2e71b1d7363368c4b06a32f0` passed [CI run 34143258029](https://github.com/openclaw/openclaw/actions/runs/34143258029), including required `openclaw/ci-gate`, and landed through native review/prepare/merge as `85f79a5c0fada45db33dca7dd22455f19100f471`.

### DIFF
--- a/docs/cli/update.md
+++ b/docs/cli/update.md
@@ -484,8 +484,11 @@ For targets that support candidate validation, the old Gateway keeps serving thr
 plugin resolution and compatibility planning. It also rehearses migrations and
 boots a canary with copied configuration and verified SQLite snapshots in an
 isolated temporary state directory. The copied database registry points to the
-copied agent databases. Channels, cron, automatic updates, and other side
-services are suppressed in this canary.
+copied agent databases. Installed plugin payloads and their dependencies are also
+copied; the rehearsal install records point to those copies, and their OpenClaw
+host links target the staged candidate. The live plugin files and host links stay
+unchanged. Channels, cron, automatic updates, and other side services are
+suppressed in this canary.
 
 Schema checks also use private SQLite copies so inspection does not create or
 modify WAL sidecars beside live databases. Each schema inspection has a
--- a/src/cli/update-cli.test.ts
+++ b/src/cli/update-cli.test.ts
@@ -359,7 +359,7 @@ vi.mock("../process/exec.js", () => ({
   // Retain real rehearsal config projection and drift checks in this CLI fixture.
   runCommandBuffered: async () => ({
     code: 0,
-    stdout: Buffer.from("[]"),
+    stdout: Buffer.from(JSON.stringify({ versions: [], pluginPaths: {} })),
     stderr: Buffer.alloc(0),
   }),
   runCommandWithTimeout: vi.fn(),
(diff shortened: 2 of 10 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8b8b3f368eff2b2a2e71b1d7363368c4b06a32f0](https://github.com/openclaw/openclaw/pull/141175/commits/8b8b3f368eff2b2a2e71b1d7363368c4b06a32f0)
- Landed commit: [85f79a5c0fada45db33dca7dd22455f19100f471](https://github.com/openclaw/openclaw/commit/85f79a5c0fada45db33dca7dd22455f19100f471)

PR #141181, opened 2026-09-07
fix(agents): disable per-attempt AsyncLocalStorage on lifecycle dispose

### LINKED ISSUE #141122: Gateway heap leak: a per-run AsyncLocalStorage is never .disable()d, so storageList grows without bound and taxes every async resource
## Summary

`createEmbeddedAttemptTranscriptLifecycle` creates one `AsyncLocalStorage` **per agent run attempt** and never calls `.disable()` on it. Node's internal `storageList` holds a strong reference, so the instances are uncollectable and accumulate for the life of the gateway process.

This has two costs, and the second is the one that hurts: every enabled ALS stamps a property onto **every async resource created afterwards**, and legacy `async_hooks` ALS walks the whole `storageList` on each one. So the cost per await grows linearly and without bound as the process does more work.

I believe this is a distinct, unreported cause behind several of the long-running "gateway RSS grows to many GB" reports (#91588, #140443). It is **not** #121572 — that one is the CDP connection registry, which I fixed separately in my own code and which is genuinely eliminated in the measurements below.

## Evidence

Measured from a matched pair of V8 heap snapshots taken from a production gateway at **the same completed-task count** (174), so the two are directly comparable. Post-GC retained heap, since writing a snapshot forces a full GC.

Heap composition, post-fix snapshot (2,129 MB total retained):

| component | bytes | share |
|---|---|---|
| symbol-keyed property backing stores stamped on `ManualPromise` by leaked ALS | **1,371 MB** | **64.4%** |
| everything else (Node/module baseline, page graph, app data) | 758 MB | 35.6% |

`AsyncLocalStorage` instances in that snapshot: **243 total, 185 retained ONLY by a `storageList` element edge** — no application binding at all, i.e. the creating closure is long gone. An earlier snapshot from a younger process on the same instance has 74 ALS nodes and essentially no orphans, so it grows monotonically with runs.

Average `(object properties)` backing store per `ManualPromise`: **7,545 B → 11,768 B** between two snapshots (147 → 205 property slots each). Every slot is keyed by a per-instance ALS store symbol.

## Reproduction of the mechanism

Independent of this codebase, on `node v22.22.3`:

```js
const { AsyncLocalStorage } = require('node:async_hooks');
const keep = [];
for (let i = 0; i < 179; i++) { const a = new AsyncLocalStorage(); a.enterWith({i}); keep.push(a); }
// then time `await Promise.resolve()` in a loop
```

| enabled ALS instances | ns per await |
|---|---|
| 0 | 96 |
| 30 | 4,148 |
| 179 | **14,990** (measured on the affected host) |
| 300 | 35,937 |

And collectability: 50 leaked instances without `.disable()` → **0 of 50 collected**; with `.disable()` → **50 of 50**.

## The fix

`dispose()` already exists and is already called — it tears down the drain and sets `disposed = true`, but never touches the ALS. Adding `lifecycleOwner.disable()` there appears to be the whole change. `.disable()` does not appear anywhere in the installed bundle.

I checked the 2026.9.2 tarball: same construction site, same `dispose()`, still unfixed. 64 of the 65 ALS sites in the bundle are per-process and fine; this is the only per-run one I found.

## Workaround that works today

`NODE_OPTIONS=--experimental-async-context-frame` swaps the implementation for `AsyncContextFrame`, which is the shipped default from Node 24 (nodejs/node#55552) and which this project's own `engines` already permits.

Measured on the affected host, at 179 instances: **14,990 ns → 78 ns per await**, a ~192x improvement. It also makes the orphans collectable (0 of 50 → 50 of 50), so it addresses the retention as well as the speed.

Caveat, stated honestly: an ALS entered on a continuation that never unwinds still retains. The per-attempt shape here is the unwinding kind, which is why the flag helps this particular defect.

## Environment

- openclaw 2026.9.1 (also verified present in 2026.9.2), node v22.22.3, Linux, gateway as a systemd user service
- Continuous agent workload driving Chrome over CDP, ~240 page reads/hour, LLM streaming over SSE, transcripts to SQLite
- Heap analysis done with a streaming parser over the flat `nodes`/`edges`/`strings` arrays; happy to share methodology or numbers if useful

ooooo9x:
## What Problem This Solves

`createEmbeddedAttemptTranscriptLifecycle` creates a fresh `AsyncLocalStorage` for every agent run attempt and never calls `.disable()` on it. Node's internal `storageList` holds a strong reference to each enabled instance, so they accumulate for the life of the gateway process.

Two compounding costs (see #141122 for heap-snapshot + benchmark evidence):
1. Every leaked ALS stamps a per-instance store symbol onto **every async resource created afterwards**, and
2. legacy `async_hooks` ALS walks the whole list on each one.

Measured on the affected host, cost per `await` grows linearly and without bound: 96 ns at 0 enabled instances → ~14,990 ns at 179. A production snapshot showed 243 retained ALS instances, 185 of them kept alive only by a `storageList` edge.

## Why This Change Was Made

Release the owned AsyncLocalStorage by calling `.disable()` **after the actual serialized drain settles** (in `beginCleanup()`), not on dispose. This keeps bounded teardown while ensuring no admitted callback can still be observing the store when it is disabled:

- In the normal path, `beginCleanup()` drains every admitted transcript write, then the drain sentinel runs once and disables the store.
- If a transcript callback outlives the 30s teardown budget, the budget resolve lets `dispose()` return, but the store is **not** disabled early. The still-running callback's nested `withTranscriptWrite()` continues to be admitted as a descendant; the store is released only once that drain eventually settles. A regression test (`still admits a nested write from a callback that outlives the teardown budget`) locks this in and fails on the earlier fix where disable ran on the budget.

A small optional `deps.createLifecycleStore` seam (defaulting to an owned instance) lets the regression test observe the real per-attempt store.

## User Impact

Completed attempts release their transcript context after admitted writes drain. Long-running Gateways using legacy AsyncLocalStorage avoid accumulating these per-attempt stores, while callbacks that outlive the teardown budget retain nested-write admission.

## Evidence

Maintainer verification at `314cb2e39efe8d4b34dacae3559e802451d227e3`: [fresh sanitized AWS Crabbox proof](https://github.com/openclaw/openclaw/pull/141181#issuecomment-5599556871) on Node 24.19.0 collected 40/40 disposed stores versus 0/40 undisposed controls; all 23 focused lifecycle, preparation, and settlement tests passed.

Real runtime (legacy AsyncLocalStorage), `--expose-gc` retention child (`attempt-transcript-lifecycle.retention.test-support.ts`), 40 instances each:

```
retention ok: disposed collected=40/40 leaked collected=0/40
```

node v22.23.2, post-GC `heapUsed` delta over 500 attempts:

```
after 500 attempts (post-GC heapUsed delta): leaked(no dispose) +0.45 MB, disposed +0.18 MB
```

The retention child is launched with `--no-async-context-frame` on Node >= 24 so the "leaked stays retained" control remains meaningful on the repository's recommended runtime (Node 24+ defaults to AsyncContextFrame, where completed contexts are collected without `.disable()`).

Additional checks, all green locally:

- `attempt-transcript-lifecycle`, settle, prepare, replay, and cleanup suites pass.
- `pnpm tsgo:core` typecheck passes; oxlint + oxfmt clean on the changed files.

## ClawSweeper review follow-up

Both actionable P2 findings from the first review pass are addressed:

- **Disable only after the actual drain finishes** — implemented (see Fix); covered by the outlived-callback nested-write regression test.
- **Retention control valid on supported Node 24** — the child selects the legacy implementation on Node >= 24 via `--no-async-context-frame`; on Node < 24 the default already is legacy.
## Note for reviewers on the "real agent-run proof" exit criterion

The review bot's remaining pre-merge item asks for before/after time+memory output
from repeated **real agent/Gateway runs**. In this repository that criterion cannot be
satisfied while also exercising the code path this change touches, because the two
requirements are structurally mutually exclusive:

- A **complete** `runEmbeddedAgent` flow in the test harness mocks the attempt itself
  (`mockedRunEmbeddedAttempt` in `src/agents/embedded-agent-runner/run.overflow-compaction.harness.ts`, consumed by `runEmbeddedAgent` in `run-orchestrator.ts`), so it never
  constructs the per-attempt `AsyncLocalStorage` this PR changes.
- Constructing that store requires the real attempt life-cycle entry point
  `prepareEmbeddedAttemptTranscriptLifecycle` (`run/attempt-transcript-lifecycle-prepare.ts`),
  which only runs inside a full `runEmbeddedAttempt` (`run/attempt.ts`) — a much heavier,
  environment-coupled path (sandbox, MCP, context engine, session store) with no existing
  test fixture that drives it directly.

Consequently, "run the full agent and observe the ALS" is not reachable in the repo's
test surface without introducing a large new integration environment solely for this
narrow fix. The proof above therefore exercises the exact real functions the fix lives
in — `createEmbeddedAttemptTranscriptLifecycle` + `beginCleanup`/`dispose`
(`run/attempt-transcript-lifecycle.ts`, matching the settle sequence at
`run/attempt-session-settle.ts:151` → `:170`) — and demonstrates the retention property
(disposed 40/40 collected vs leaked 0/40) plus 500-attempt post-GC heap deltas on the
real runtime.

Maintainer judgment is requested: given the fix is a small, well-scoped ALS-release
change with the retention property proven on the exact functions it modifies, is the
full mock-agent-run measurement required, or is the existing evidence sufficient?

Closes #141122


ooooo9x:
@clawsweeper re-review

ooooo9x:
I've added a short "Note for reviewers on the real agent-run proof" section to the PR body explaining a structural constraint I verified in the codebase:

- The review bot's remaining pre-merge item asks for repeated **real agent/Gateway run** time+memory measurements, but in this repo a *complete* `runEmbeddedAgent` test flow mocks the attempt itself (`mockedRunEmbeddedAttempt` in `run.overflow-compaction.harness.ts`, consumed at `run-orchestrator.ts`), so it never constructs the per-attempt `AsyncLocalStorage` this PR changes.
- Constructing that store goes through `prepareEmbeddedAttemptTranscriptLifecycle` (`run/attempt-transcript-lifecycle-prepare.ts`), which only runs inside a real `runEmbeddedAttempt` (`run/attempt.ts`) — a heavy, environment-coupled path with no existing fixture that drives it directly.

So "run the full agent and observe the ALS" and "exercise the changed code path" are mutually exclusive in the current test surface. The retention proof (disposed 40/40 collected vs leaked 0/40) plus 500-attempt post-GC heap deltas run on the exact real functions the fix lives in (`createEmbeddedAttemptTranscriptLifecycle` + `beginCleanup`/`dispose`, matching the `run/attempt-session-settle.ts:151` → `:170` settle sequence).

Since the fix is a small, well-scoped ALS-release change with the retention property proven on the exact functions it modifies, I'd like maintainer judgment on whether the full mock-agent-run measurement is required or the existing evidence is sufficient. @clawsweeper re-review for the updated body.

ooooo9x:
@vincentkoc This fix for the per-attempt AsyncLocalStorage leak (#141122) is now fully green through the automated review path — ClawSweeper's latest pass reports **"Before merge: None"** (🐚 platinum hermit 4/6 across overall/proof/patch, `status: 👀 ready for maintainer look`), and it explicitly withdrew its earlier "full agent-run benchmark" request as "broader than necessary" for this narrow invariant.

Quick orienting summary:
- **Fix**: release each attempt's owned AsyncLocalStorage via `.disable()` after the actual serialized transcript drain settles (in `beginCleanup`), not on dispose — so completed attempts stop accumulating in Node's global `storageList` without ever disrupting still-running admitted write callbacks or their descendants.
- **Behavior proof**: real-runtime `--expose-gc` retention child — disposed stores collected 40/40 vs undisposed 0/40, plus 500-attempt post-GC heap deltas on node v22.
- **Why no full-agent-run benchmark**: a complete `runEmbeddedAgent` test flow mocks the attempt itself (`mockedRunEmbeddedAttempt` in `run.overflow-compaction.harness.ts`), so it never constructs the store this PR changes; driving the real store requires the heavy `run/attempt.ts` path which has no test fixture. ClawSweeper accepted this reasoning and cleared the blocker.

Would you be able to take a look when you have a moment? Happy to adjust if anything needs it.

jalehman:
## Exact-head runtime proof

Re-ran the production lifecycle retention check on **314cb2e39efe8d4b34dacae3559e802451d227e3** in a fresh sanitized AWS Crabbox with Node **24.19.0**. The trusted bootstrap verified the PR SHA, absent instance IAM credentials and package-manager pin. No credential hydration or Tailscale was used.

```sh
node --expose-gc --no-async-context-frame --import tsx \
  src/agents/embedded-agent-runner/run/attempt-transcript-lifecycle.retention.test-support.ts
```

Observed output:

```text
retention ok: disposed collected=40/40 leaked collected=0/40
```

The legacy ALS flag makes the undisposed control meaningful on Node 24. This runs the actual production transcript lifecycle with real AsyncLocalStorage and garbage collection; it does not establish whole-Gateway throughput or RSS.

Supplemental validation: all **23 tests across 3 files** passed in the same isolated run: `attempt-transcript-lifecycle.test.ts`, `attempt-transcript-lifecycle-prepare.test.ts`, and `attempt-session-settle.test.ts`. This includes descendant admission after the teardown budget expires. Independent P0–P2 autoreview found no actionable findings.

[Crabbox run `run_9ebd4de84c86`](https://crabbox.openclaw.ai/portal/runs/run_9ebd4de84c86), provider `aws`, lease `cbx_a2bd150117e4`, exit **0**.

The PR body headings now match the contribution template; [PR context and evidence passes](https://github.com/openclaw/openclaw/actions/runs/34334060704/job/102409187340). Code and contributor commits are unchanged.

jalehman:
@clawsweeper re-review

The remaining historical-proof equivalence gap is resolved by direct Git object comparison. All three proof-relevant files at recorded proof revision `314cb2e39efe8d4b34dacae3559e802451d227e3` and current head `47b55017364b3d4f1c23f5bb9dceb04a84c7e9e6` have identical blob IDs:

- `src/agents/embedded-agent-runner/run/attempt-transcript-lifecycle.ts`: `80757040e08660bce0f096ffc38567d3b76470c1`
- `src/agents/embedded-agent-runner/run/attempt-transcript-lifecycle.test.ts`: `f3dbc044972e1f06fa68da32957d209e6e16ac2b`
- `src/agents/embedded-agent-runner/run/attempt-transcript-lifecycle.retention.test-support.ts`: `4c8f11be9353735c00094971b0caa3749fae2126`

`git diff --exit-code 314cb2e39efe8d4b34dacae3559e802451d227e3 47b55017364b3d4f1c23f5bb9dceb04a84c7e9e6 -- <these three paths>` returned 0 with no differences. Both contributor commits also remained equal in the rebase range-diff. This preserves the existing narrow lifecycle/retention proof; no new runtime execution or whole-Gateway benchmark is claimed. Required CI currently has failures and remains a separate landing gate. Please reassess the byte-equivalence merge-risk item using this evidence.

jalehman:
Merged via squash.

- Prepared head SHA: [b5eda94b37f4865f4c089b68ca2d01804ebccca5](https://github.com/openclaw/openclaw/pull/141181/commits/b5eda94b37f4865f4c089b68ca2d01804ebccca5)
- Landed commit: [ea611460b5a2d2ddc56267184762d019d122b948](https://github.com/openclaw/openclaw/commit/ea611460b5a2d2ddc56267184762d019d122b948)

PR #141188, opened 2026-09-07
fix(pr): reject unexpected branch changes during publication

vincentkoc:
<details>

</details>

## What Problem This Solves

Resolves a problem where native PR preparation could adopt an unexpectedly changed branch head, retry against that new head, or report success for a replacement commit with the same tree. A newly reviewed preparation could also remain blocked by a receipt from the previous preparation.

## Why This Change Was Made

Keep publication authority pinned to the validated preparation source, or to a matching receipt from that preparation's prior publication; remote observations validate authority but never supply it. Remove automatic lease refresh, rebase, and retry, propagate the actual Git push status, and require both PR readback and the fetched ref to equal the authoritative returned publication OID before writing success artifacts.

GraphQL requests keep the existing `gh_plain` writer route and use a real input file. The existing routing guard and publisher fixture now reject stdin instead of preserving or accepting the obsolete transport.

## User Impact

- Unexpected advances, rewinds, and same-tree replacements stop publication while retaining state and evidence. An exact-candidate no-op must also preserve the authorized ancestry.
- Preserve legitimate GraphQL publication: its verified hosted OID may differ from the local candidate. The recorded local/hosted pair permits an exact no-op; subsequent fixups must extend the hosted commit.
- Only an actual Git permission failure may switch to GraphQL, with the same immutable lease. General Git failures and GraphQL failures are not retried.
- Explicitly validated new preparations preserve prior receipts, gate evidence, and result artifacts before retiring active stamps. Repeated sync continues to validate and use its matching receipt.

This six-file PR changes only native preparation helpers and their existing tests. It is a maintainer-workflow reliability repair; runtime behavior, configuration, dependencies, and release artifacts are unchanged by this PR.

## Evidence

- Both existing owner suites passed **118/118 tests** on the preserved repair commit `1217f0d`: `test/scripts/pr-main-refresh.test.ts` and `test/scripts/pr-prepare-gates.test.ts`.
- The earlier main integration passed **25/25 focused publication and lifecycle cases** on `d7c7c99`, with the original five-file repair unchanged.
- The stale stdin routing assertion was reproduced locally before correction. The test-only correction at `1f7097c4` passed the full **70/70 wrapper tests** plus **10/10 existing GraphQL and transport cases**, retaining file-input, commit-body, returned-OID, lease, ancestry, and fallback coverage.
- After merging pinned main `b07b7d8b`, which contains the separate UI fixture repair, **15/15 selected routing, GraphQL, and transport cases** passed on `ce8cec02`. The six-file PR delta is byte-identical to the reviewed candidate.
- Before their respective fixes, the fixtures demonstrated **14 publisher regression failures**, **1 retained-receipt lifecycle failure**, and **2 invalid no-op ancestry failures**. These exercise real temporary Git repositories and refs with file-only transport and a synthetic GitHub boundary; no real GitHub mutations occur during tests.
- Independent P0-P2 review of the six-file candidate found no remaining actionable issues. Targeted formatting, lint, Bash syntax, and diff-whitespace checks passed across the repair and correction; the final integration also passed signature and diff-whitespace verification.
- Local `check:changed` stopped at the unchanged compiler-containment guard because the shared compiler symlink resolves outside the checkout. That validation gap remains explicit; no guard was bypassed and no shared dependencies were modified.
- Exact-head [CI run 34123007538](https://github.com/openclaw/openclaw/actions/runs/34123007538) passed on `ce8cec02f9db1f70622e626247411ea11193ea28`, including production types. The formerly failing cached-read/writer-sensitive routing assertion passed in **1 ms** in job `101745376716`; its two test groups reported **739 passed, 2 skipped** and **550 passed**.


vincentkoc:
Merged via squash.

- Prepared head SHA: [ce8cec02f9db1f70622e626247411ea11193ea28](https://github.com/openclaw/openclaw/pull/141188/commits/ce8cec02f9db1f70622e626247411ea11193ea28)
- Landed commit: [5520a73e9115be31d5f710da23e9bd93ecff291f](https://github.com/openclaw/openclaw/commit/5520a73e9115be31d5f710da23e9bd93ecff291f)

PR #141199, opened 2026-09-07
fix(telegram): show compaction progress in partial and block previews

VACInc:
### High Level TLDR

Telegram could look idle during a long automatic compaction when answer previews used `partial` or `block` mode. Show the existing compaction status in those previews, respecting `streaming.preview.toolProgress`.

Related: #134826

## What Problem This Solves

Fixes an issue where users waiting for a reply would see no compaction progress in `partial` or `block` preview mode, even with preview tool progress enabled. A reported turn spent 111.42 seconds in Claude CLI automatic compaction before responding.

## Why This Change Was Made

Remove the Telegram-only `progress` mode restriction. The existing compositor already owns preview enablement and rendering; existing final-delivery guards still prevent late progress from changing a finished answer.

### Root Cause

The restriction was introduced by #134826 (`b7fa0f52c6a9`) and remains on upstream main at `701687f6b7104dca7ba990a2557c840206af8ee8`. `canPushCompactionProgress` rejected every `partial` and `block` turn before the compositor could honor the preview setting.

Read-only incident correlation separated transport and provider work: CLI execution began within 0.5 seconds of Telegram ingress; the native transcript records automatic compaction from 238,431 to 30,191 tokens in 111,420 ms. The first Telegram response followed after compaction. The applied #138984 session-rewrite overlay does not change this Telegram gate or Claude's native compaction; the resumed CLI turn did not reseed OpenClaw history.

## User Impact

Users with preview tool progress enabled see “Compacting context...” before model output, followed by completion or incomplete status. Disabled preview progress remains quiet, reactions still work, ambient room events remain invisible, and final answers retain their existing delivery behavior. This fixes the unexplained silence; it does not reduce native Claude compaction time.

### Evidence / Real behavior proof

The Telegram dispatch regression exercises compaction before any assistant or tool output, including an incomplete attempt followed by successful compaction. It covers `progress`, `partial`, and `block`, plus disabled previews and streaming off.

```sh
node scripts/run-vitest.mjs \
  extensions/telegram/src/bot-message-dispatch.draft-failures-progress.test.ts \
  extensions/telegram/src/bot-message-dispatch.status-reactions.test.ts \
  extensions/telegram/src/bot-message-dispatch.reasoning-room-events.test.ts
```

- Baseline: the two new preview-mode regression cases fail because no compaction preview is produced.
- Patched: 49 tests pass across three files on Node 24.15.0. Preview statuses are available before model output; partial/block finals update the existing answer stream, and progress-mode final delivery remains intact.
- `pnpm build`: passed.
- `node scripts/check-changed.mjs`: passed after relocating the worktree.
- Local review: verified setting precedence, disabled previews, final-delivery guards, and room-event suppression. Independent agent review was not run because delegation is disabled.

A local rendering benchmark on Linux x64 / Node 24.15.0 used 20 warmups and 100 fresh partial-preview turns per version. Baseline produced 0/100 compaction previews; patched produced 100/100. Handler duration median/p95 was 0.00057/0.00164 ms before and 0.01196/0.01901 ms after. Whole-process RSS was approximately 1,094/940 MiB; imports dominate these separate processes, so this is not a retained-heap comparison. No Telegram or provider latency is included in those measurements.

### What was not tested

A real Telegram Test Server run was blocked by the canonical doctor: `Telegram Test Server bot is not an active member of the test group.` Rechecked after marking this PR ready: Test Bot API `getChatMember` returns HTTP 200 / `ok: true` / status `left` for the leased bot, and TDLib reports `chatMemberStatusLeft` for the QA user. The leases were released and their temporary credential state cleaned up. The QA group membership must be restored before completing the required live preview-to-final-answer proof in partial and block modes. ClawSweeper reviewed this head again and found no actionable patch defect; its remaining blocker is that live proof. The operator Gateway was not changed or restarted. No claim is made that the native compaction itself becomes faster.

The first changed-check attempt was blocked by an ancestor `node_modules` symlink that resolved declaration metadata outside the worktree. The task checkout was moved outside that hierarchy and the canonical check was rerun without weakening the guard.

### DIFF
--- a/docs/channels/telegram.md
+++ b/docs/channels/telegram.md
@@ -380,7 +380,7 @@ curl "https://api.telegram.org/bot<bot_token>/getUpdates"
     - `streaming.progress.commentary` (default: `false`) opts into assistant commentary/preamble text in the temporary progress draft
     - legacy `channels.telegram.streamMode`, boolean `streaming` values, and retired native draft preview keys are detected; run `openclaw doctor --fix` to migrate them
 
-    Tool-progress lines are the short status updates shown while tools run (command execution, file reads, planning updates, patch summaries, Codex preamble/commentary in app-server mode). `partial` and `block` previews show them by default; the `progress` draft shows them only with `streaming.progress.toolProgress: true`.
+    Tool-progress lines are the short status updates shown while tools run (command execution, file reads, planning updates, patch summaries, Codex preamble/commentary in app-server mode). `partial` and `block` previews show them by default; the `progress` draft shows them only with `streaming.progress.toolProgress: true`. Compaction status follows the same settings and appears as soon as compaction starts, including before the first model output.
 
     Keep answer-preview edits but hide tool-progress lines:
 
--- a/extensions/telegram/src/bot-message-dispatch-progress.ts
+++ b/extensions/telegram/src/bot-message-dispatch-progress.ts
@@ -135,7 +135,6 @@ export function canPushToolProgress(turn: Turn): boolean {
 
 function canPushCompactionProgress(turn: Turn): boolean {
   return Boolean(
-    turn.streamMode === "progress" &&
     turn.answerLane.stream &&
     !turn.answerLane.finalized &&
     !turn.finalAnswerDeliveryStarted &&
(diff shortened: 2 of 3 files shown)


VACInc:
Related transcript repair https://github.com/openclaw/openclaw/pull/138984 is now merged into main as b0ae5b6793e5e1e2ba55b7e1d632e15e0fd48d00.

It fixes OpenClaw transcript publication and selected-history correctness. It does not change Telegram's partial/block compaction-preview gate or reduce native Claude compaction time. This preview change remains distinct and is not superseded by the merge.

Leaving this PR open; its own preview-to-final-delivery validation remains separate.

PR #141209, opened 2026-09-07
feat(agents): sharpen sessions_spawn research guidance and add stalled-child follow-up

hartmark:
## What Problem This Solves

`sessions_spawn`'s tool description told the model to spawn a hidden subagent for "research... throwaway side tasks", but that phrasing was too generic for the model to reliably recognize a long, multi-step, trial-and-error investigation against an external API/service (auth probing, endpoint discovery, iterative debugging) as that exact pattern. Separately, after spawning children, nothing reminded the model to check on ones that stalled.

## Why This Change Was Made

Live incident: during a heartbeat-triggered login investigation (probing an external service's undocumented auth API), the model ran ~35 minutes of consecutive `exec` calls inline in the main session -- dozens of curl/python attempts, credential combinations, and GitHub source greps -- instead of delegating that exploratory work to a hidden subagent. The run hit its own configured `agents.defaults.timeoutSeconds` (900s) mid-loop and the turn was aborted with "Request timed out before a response was generated." This is exactly the "research/throwaway side task" case the tool description already calls out for hidden-subagent delegation; the model just didn't recognize the pattern from the existing generic wording.

## User Impact

- `sessions_spawn`'s description now explicitly names "multi-step trial-and-error against external APIs/services (auth probing, endpoint discovery, iterative debugging)" as a hidden-subagent case, making the delegation guidance concrete instead of generic.
- After spawning, the model is reminded (once, gated on `subagents`/`sessions_history` actually being available -- the existing gated cross-tool-reference mechanism, not a new ungated static reference) to check its own spawned children for `failed`/`timeout`/`killed` status and follow up instead of leaving them stalled. Phrased as a single follow-up check, consistent with the existing "never loop-poll status" guidance elsewhere in the system prompt.
- No production behavior change beyond model-facing prompt text; no new config surface.

## Evidence

- Root cause traced live: transcript shows the exact inline tool-call sequence (`docs/... n/a` -- internal incident, not filed as a public issue) hitting `[agent/embedded] embedded run timeout: ... timeoutMs=900000`.
- Confirmed the false-idle-timeout regression (#117884) is unrelated and already fixed upstream; this is the model's own delegation-judgment gap, addressed via clearer prompt text.
- Tests: `src/agents/agent-tools.deferred-followup-guidance.test.ts` (35/35), `src/agents/tool-description-presets.test.ts` + `src/agents/system-prompt.test.ts` (154/154), all passing pre- and post-change.
- `oxfmt --check` clean on both touched files.

https://claude.ai/code/session_01Vk2cBRQdErwRtPea8VuhG6


hartmark:
Fixed the P2 finding (commit ad9ad7431bd):

> Gate terminal-status guidance on an actual status tool

Confirmed directly against source: `sessions_history` (`src/agents/tools/sessions-history-tool.ts`) returns only `{sessionKey, messages, truncated, droppedMessages, contentTruncated, contentRedacted, bytes, pendingInputs}` -- no run-status field, ever. Only `subagents` reports terminal status, via `STATUS_MAP` (`failed`/`timed_out`/`cancelled`, not the `sessions_list`-style `failed`/`timeout`/`killed` vocabulary the original sentence borrowed). The new follow-up sentence is now gated on `subagents` specifically and uses its real status vocabulary; `sessions_history`'s existing, unrelated "Check spawns via ..." lookup mention is untouched.

Added a regression test proving the sentence is absent when only `sessions_history` (no `subagents`) is available, and present with the correct wording when `subagents` is available. All 36 tests in the affected suite pass.

On "Add real behavior proof": this PR changes model-facing prompt text only (no runtime logic), so there's no deterministic pass/fail behavior to demonstrate the way a code-path test would -- it's a probabilistic nudge to model judgment, not a new capability. What *is* verifiable and now proven end-to-end (not just via synthetic test objects) is that the real assembly pipeline renders the intended text. Ran the actual production factory + `applyToolAvailabilityDescriptions` together:

```
$ node --import ./scripts/tsx.mjs -e '... createSessionsSpawnTool + applyToolAvailabilityDescriptions ...'
=== live-rendered sessions_spawn description ===
...Hidden child: research, parallel/batch reads, throwaway side tasks -- including multi-step
trial-and-error against external APIs/services (auth probing, endpoint discovery, iterative
debugging)... Before moving on, check own spawned children via `subagents` for failed/timed_out/
cancelled status and follow up instead of leaving them stalled.
```

Confirms both additions render correctly together through the real construction path a live agent actually sees, with the (now-corrected) `subagents`-only gate applied.

@clawsweeper re-review

hartmark:
Regenerated the prompt snapshot fixtures (`pnpm prompt:snapshots:gen`) — they'd drifted after the rebase onto current `main`. `prompt:snapshots:check` is clean now.

@clawsweeper re-review

hartmark:
Fixed the P2 finding: reworded the added sentence from an unconditional "Before moving on, check own spawned children via `subagents`..." to "If an announced child hasn't reported back when expected, check it via `subagents`..." — ties the status check to an actual outcome gap instead of contradicting the existing push-based completion contract (`delegation-guidance.ts`'s "Never busy-poll" / "only for requested status/debug", `system-prompt.ts`'s status-check rule).

Real rendered production output confirming the new wording:

```
...No spawn for quick lookup/single read. Check spawns via `subagents`/`sessions_history`. After spawn, do non-overlap work; follow the receipt's completion mode. If an announced child hasn't reported back when expected, check it via `subagents` for failed/timed_out/cancelled status and follow up instead of leaving it stalled.
```

Updated the test assertions and regenerated the prompt snapshot fixtures to match.

@clawsweeper re-review

PR #141216, opened 2026-09-07
fix(ui): count selected usage tool invocations accurately

steipete:
<details>

</details>

## What Problem This Solves

Fixes an issue where users inspecting a selected interval in the Usage timeline would see incorrect tool-call counts when an assistant called the same tool repeatedly and received results. Two read calls followed by their two results displayed as three calls.

## Why This Change Was Made

The selected-range summary now counts assistant invocations using the multiplicity already preserved by the compact-log parser. Result markers remain available for conversation filtering. The repair replaces the inaccurate accumulation without increasing production lines (+4/-4).

## User Impact

The tool-call total and top-tool rows reflect invocations in the selected interval of the loaded conversation. The Usage feature reference explains the behavior.

## Evidence

The regression failed on the original source for both supported result-row dialects, with an expected count of 2 and an actual count of 3. The repaired candidate passes 20 focused detail/parser tests. An isolated Chromium flow used the real timeline handle: it displayed 3 calls before the fix and 2 afterward, checked the per-tool count, then reset the range successfully.

Before: selected interval incorrectly shows 3 tool calls
After: selected interval correctly shows 2 tool calls

The validated candidate also passes the production UI type graph, both owning test type graphs, full-file lint, stylelint, i18n verification, and formatting. Independent review is scoped-clean through P2.

After refreshing onto current main, the same 20 detail/parser tests, both owning test type graphs, full-file lint, and Chromium drag/reset flow pass. A fresh after screenshot confirms 2 tool calls, read 2, and exec 0 in the selected interval. Final independent review of the refreshed integration is scoped-clean through P2.

Known separate follow-ups: whole-session aggregation deduplicates repeated tool names within a message, and selected-range user/assistant message counts are inferred from sampled token points. Those remain outside this repair. Counts here use the existing bounded conversation data; no persistent aggregate, schema, API, or retention change is included.

### DIFF
--- a/docs/web/control-ui/feature-reference.md
+++ b/docs/web/control-ui/feature-reference.md
@@ -83,6 +83,7 @@ Control UI capabilities grouped by area, each with the Gateway RPC methods behin
     - Session-derived token and estimated-cost analysis stays separate from provider billing.
     - Filter sessions with the provider, model, channel, or tool menus, or type case-insensitive `key:value` terms. Values within one category match as alternatives. Toggling a menu option preserves the other filters and their quoted text.
     - Selecting days narrows token and cost totals to those days within the active session filters. Daily charts and exports retain that session scope. Provider/model/tool queries select matching sessions, including all usage within each matched session; hour filters select sessions active in those hours.
+    - Drag the handles in a session’s usage timeline to inspect an interval. Tool-call counts use assistant invocations in the loaded conversation, including repeated calls to the same tool; tool results do not add calls.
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [44d89cd9b60262fd20720ca1a5b252faccd90799](https://github.com/openclaw/openclaw/pull/141216/commits/44d89cd9b60262fd20720ca1a5b252faccd90799)
- Landed commit: [5e032e5d25b58942f3a0301a66af2d7018984ca0](https://github.com/openclaw/openclaw/commit/5e032e5d25b58942f3a0301a66af2d7018984ca0)

PR #140725, opened 2026-09-07
refactor(update): verify updates without model inference

steipete:
## What Problem This Solves

Fixes an issue where an otherwise healthy OpenClaw update could fail, enter repair, or roll back because configured model inference was unavailable or a verification turn could not be saved.

Related: #140274, #140419, #124396.

## Why This Change Was Made

Per maintainer decision, update verification uses managed-service liveness and port ownership, the Gateway hello handshake and expected version/build identity, plugin and channel health, `/readyz`, and existing config/schema/state checks. It runs no model inference, including no advisory probe. This retires the serving verifier introduced by @fuller-stack-dev in #140274 while preserving Doctor's canonical run admission and the independent rollback and recovery fixes from #140419. The bounded repair loop still uses configured inference after a failure.

## User Impact

`openclaw update`, auto-update, and Doctor update can complete without a model provider serving a turn. The Control UI update view no longer displays an inference indicator. The optional `inferenceProbe` verification fact is removed from `update.runs`; omitting this optional server-to-client field is safe for older clients and does not require a protocol version bump. No configuration or SQLite schema changes.

## Evidence

- Focused CLI, daemon, Doctor, flows, ledger/report, protocol, and UI proof: 520 files, 8,617 tests passed; five existing skips. The update browser flow also passed on the previous UI source for the comparison below.
- The full `pnpm check:changed` plan, `pnpm build`, `pnpm protocol:check`, and `pnpm plugin-sdk:surface:check` passed. The explicit build emitted zero `[INEFFECTIVE_DYNAMIC_IMPORT]` warnings.
- The top-level update CLI test passed all 409 tests; the final service-verification test passed all 9 tests.
- Committed-branch Codex review passed with no actionable P0/P1 findings.
- `pnpm test:changed` failed twice locally in the unchanged `src/worker/worker.runtime.test.ts` suite with `ENOENT` reading its own `retained-profile.txt` workspace file, including a rerun without concurrent checks. The four-file focused rerun passed all 229 tests. The first broad run also had failures in unchanged fleet, transcript, and Doctor cases covered by that successful rerun. Those suites are untouched by this PR.
- All three protocol generators ran. Swift uses a dictionary for verification facts and Kotlin does not emit an update-run model, so no tracked generated client changes are needed. The JSON schema is build output.
- The retired verification symbols are absent; the repair loop, Doctor admission, and independent rollback fixes remain intact.

Synthetic Control UI screenshots:

| Before | After |
| --- | --- |
[...]

PR #140419, opened 2026-09-06
fix(update): preserve rollback and truthful serving reports

steipete:
Related: #140274

<details>
<summary>Additional instructions</summary>

This is a same-repository branch; maintainers can push updates.

</details>

## What Problem This Solves

Fixes an issue where operators updating a fresh installation could lose automatic rollback and receive misleading verification and recovery reports. In the AWS packaged update lane on `7d95c2f50bb`, the candidate reached verification and persisted its agent turn, but the report called the failure `persistence-missing`, treated the first agent database as migrated state, and told the operator the healthy running candidate remained stopped.

## Why This Change Was Made

The schema comparison now distinguishes first-use database creation from changes to existing databases. New stores at the candidate's supported schema are neutral; real migrations, lost databases, and foreign versions still block rollback. A separate guard checks the previous package's recorded schema support before stopping the candidate or restoring code, so neutral creation cannot authorize an incompatible rollback.

Fresh transcript readback distinguishes absent persistence, incomplete evidence, and a completed response that does not match. It accepts the run-specific token as a whole word within punctuation or a short sentence while preserving run identity, lineage, provider/model metadata, stop reason, bounded reads, and the final snapshot checks. Final recovery guidance uses the latest service observation and retains the verified version only while the same process remains running.

The served-and-persisted check remains mandatory with its existing time budget. No configuration keys, database schemas, or release files change.

## User Impact

Fresh installations retain rollback when both generations support the new database. Operators see the actual serving failure and an actionable repair command in the report, completion notice, and update status. A running but unverified candidate is identified as running, with its version when known.

## Evidence

- Reproduced all three defects before production edits: first persisted agent turn blocked rollback, saved mismatched/punctuated replies reported missing persistence, and running-candidate guidance claimed the service remained stopped. Real shared migrations and foreign-schema controls continued to reject rollback.
- Independent review found the previous-package compatibility edge case. Two additional regressions reproduced incompatible/unknown prior schema support incorrectly permitting rollback; the guard now refuses before stopping the candidate.
- `pnpm test src/cli/update-cli src/infra/update-serving-verification.test.ts src/infra/update-candidate-state.test.ts src/infra/update-run-report.test.ts`: 587 passed, one existing skip. The test runner rebuilt the runtime for fresh-process proof.
- Full `pnpm check:changed` passed, including production/test types, lint, all three unused-export scans, schema/import guards, and unchanged ratchets.
[...]

Issue #124396, opened 2026-08-16
[Feature]: Make `openclaw update` one transaction: prove candidate before activation, verified state checkpoint, whole-product rollback, no crash-loop on migration warnings


obviyus:
## Problem

`openclaw update` changes the package but never proves the install works. Operators experience an upgrade as one transaction; we execute it as independent operations across updater, state migrations, plugins, service manager, and agent runtime. Any one can half-succeed, and nothing owns the outcome.

2026.8.1 shipped to `latest` on 2026-08-31. Since then the update path is the largest source of crash-loop reports in the tracker.

| Signal (since 2026-08-31) | Count |
|---|---|
| New issues, repo-wide | 415 |
| Labeled `impact:crash-loop` | 37 (95 more in the 08-16→08-30 beta window) |
| Body mentions `doctor --fix` | 68 |
| Mentions `exec-approvals` | 34 |

These are not 415 defects. The exec-approvals gate alone produced 19+ separate issues. Six structural faults generate the entire volume.

## Missing invariant

An update must have **exactly two durable outcomes**: the candidate running and verified, or the previous version running with its matching state restored. Today there is a third — new core, unhealthy product — that `openclaw update repair` and `openclaw doctor --fix` frequently cannot converge, and package rollback does not restore. That third state is what kills a working setup.

Completion needs a real definition too. **An upgrade is complete only when the new generation has served a real agent turn and persisted it.** Today "complete" means the package swapped. Process health is not capability: #134085 boots healthy and silently drops every cron delivery and channel message; #134156 produces a host/plugin mismatch that "looks completely healthy."

## Failure classes

Issue numbers are representative, not exhaustive. Each class recurred in 2026.8.1 after being documented in the 2026.7.x line.

**A. Fail-closed startup-migration gate turns benign conditions into a permanent crash-loop.**
Any non-empty warning list blocks boot. 8.1: one stale-plugin warning crash-looped a gateway 116 times (#135150); also #135038, #134616, #133813. Earlier triggers: archived legacy sidecars (#110199, #107223), a migration lease held by a dead process while the supervisor refreshes it before each crash (#111972), a failure path exiting via SIGSEGV so supervisors loop (#109831), plugin inventory changing mid-startup leaking generation dirs until the disk fills (#122321). `openclaw doctor` is the prescribed remedy and does not converge these.

**B. Plugin convergence happens after core is live and can fail there.**
[...]

PR #141219, opened 2026-09-07
fix(update): roll back a fresh install's first broken update

steipete:
Related: #124396, #140419, #140725

## What Problem This Solves

Fixes an issue where a fresh install's first broken update could leave the managed Gateway stopped instead of restoring the previous version. The candidate's activation Doctor added config migrations, which the rollback guard mistook for outside edits. If startup failed before verification began, the live repair worker also rejected the still-running update because its ledger phase remained `restarting`.

## Why This Change Was Made

The updater captures the restore bytes immediately before Doctor. Edits already present in that capture are preserved; subsequent edits refuse restoration. Doctor reports the hash of its consumed input snapshot and the hash of its serialized output through the existing post-install result. The parent requires the input to match its pre-Doctor capture and the output to match the observed file. Operator saves before Doctor reads the config or after its last write therefore refuse restoration. Intervening writes between Doctor passes invalidate input attribution; missing input evidence permits only a byte-identical unchanged file. Rollback checks that hash before and after stopping the candidate, acquires the normal config writer lock, and rechecks before atomically restoring the config with mode `0600`. The original include graph remains guarded, and unchanged config needs no write. This replaces the semantic config comparison and writer-stamp-only rewrite. Database schema comparison, retained-package support, service ownership, and previous-runtime verification guards remain intact.

The single restart-failure producer in `maybeRestartService` records `verifying` before returning failure, so live repair can enter `repairing` and the existing strict worker ownership check accepts the admitted run. The compensating phase transition was removed from the repair consumer. Ledger reads retain the run's environment; a read error remains distinct from a missing or finished run.

## User Impact

Fresh installs that have never updated can roll back a broken first update through `openclaw update`, auto-update, and chat/UI-originated updates. Doctor's own migrations in the main config file no longer block recovery. Operator edits after activation remain protected, and refusal guidance names the changed config file. Successful recovery uses the original headline `↩️ OpenClaw update rolled back to <version>: <reason>.` and preserves the failed update's nonzero exit and original diagnostic reason.

## Evidence

Both accepted ClawSweeper P1s are closed at the Doctor result boundary: the input snapshot fences the pre-read window, and the committed output hash fences the post-write window. Regression coverage reproduces both update failures on the original implementation and covers exact config bytes, owner-only permissions, operator edits before Doctor’s read/between Doctor’s write and parent capture/before rollback/during stop/during restoration, unchanged schema guards, durable rollback reporting, live repair ownership, and read-error propagation with a successful subsequent ownership check. Existing lifecycle/version-guard coverage remains in place.

- `pnpm test src/cli/update-cli src/cli/update-cli.test.ts src/flows/doctor-health.test.ts src/infra/update-doctor-result.test.ts`: 1,065 passed, 5 skipped.
- The input-window regression failed before this fix and now preserves the operator edit. Coverage also includes missing input receipts, unchanged files, and an operator write between three Doctor write passes.
- A real isolated reader/writer probe verified that creating a previously absent config preserves its canonical missing-file input hash.
- `pnpm check:changed`: passed for the fix and again after main integration, including production/test typechecks, lint, and all selected guards.
- The final focused run rebuilt the runtime with zero `[INEFFECTIVE_DYNAMIC_IMPORT]` warnings. Full `pnpm build` and the config-writer/schema/recovery/report suites passed in the preceding accepted follow-up.
- Codex autoreview: final local and full committed-branch reviews are scoped-clean with no actionable P0/P1 findings at `394d3a2f39a723c29974b9c1a55d25e853e34232`.

`pnpm test:changed` passed 17,622 tests before stopping on four failures in untouched fleet, transcript-occupancy, and filesystem-watcher tests. The fleet and transcript files passed an independent rerun; `src/skills/runtime/refresh.missing-root.integration.test.ts:58` still fails its event wait on this macOS host. That failure is tracked by #138642 and the file is untouched by this PR. Those other files are also unchanged. The supplied packaged-update evidence was analyzed; the AWS packaged-update lane was not rerun.

After main advanced, a merge commit preserved the fix commit and reconciled one rollback-test overlap. All 40 rollback cases passed, including main’s deferred shared-schema case and both Doctor edit windows.

The initial CI run exposed three older update CLI fixtures whose snapshots claimed a nonexistent config file. The follow-up gives those cases a real selected config file and matching snapshot, with per-test cleanup; all original recovery assertions remain intact.


steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
Landing proof from the AWS Crabbox packaged-update lane (user-systemd managed helper, fresh 2026.8.90 install by `npm install -g`, config written by first boot and never migrated).

**Before (main `24707ee6827`, run 15b):** the first-ever update targeted a candidate that crashes on every real boot. The candidate's activation doctor pass rewrote the config (added `meta.migrations.modelPolicyAllowlist`, `agents.defaults.modelPolicy`, a `wizard` block, and the writer stamp), rollback was refused with `state-migrated-no-rollback`, the repair attempt was skipped with "Repair no longer owns the update attempt." while the ledger row was still running, and the run finished `failed` with the managed gateway stopped ("Managed gateway remains stopped because update recovery could not prove a runnable installation").

**After (this PR at `ed8e0606baf`, run 16b, same scenario as the first-ever update):** the ledger shows `activating` → `openclaw doctor` → `restarting` failed → `package rollback` completed → `previous generation restoration` completed → `verifying` completed → `gateway verification` completed; the run finished `rolled-back` (reason `restart-unhealthy`) with `repair: []`, the service was back on 2026.8.90 and healthy, and the mock provider saw no update-bound request. The later commit on this PR is test-only (fixtures use real config files), so the production code under test is unchanged.

**Repair ownership:** the same lane on main showed the live repair starting at 09:50:49.986 and stopping one second later with the ownership error while observations still reported `restarting/running`; the ledger only admits `repairing` from `verifying`, which a thrown restart had never recorded. With this PR the repair phase is entered from `verifying`, so a run that reaches live repair keeps ownership; in run 16b rollback succeeded first, so no repair was needed.

**ClawSweeper P1 (capture at Doctor's write boundary):** addressed in the third commit. The Doctor subprocess records the hash of the config payload it publishes (in the config writer, at write time, scoped to its own config path) and returns it in the post-install Doctor result; the parent marks the activation snapshot as Doctor-owned only when the file it observes matches that hash, or matches the pre-Doctor bytes when Doctor wrote nothing. An operator save between Doctor's last write and the parent's read therefore refuses rollback and keeps the edit (regression test added). The `configHash` field lives in the transient parent-to-child Doctor result file of one update run, so there is no persisted data-model change; the ledger's `config rollback` step is an ordinary result step.

**Full lane on `c41ba4a8b1e` (run 17):** success, chat-originated, no-op, poisoned staging, crash-on-boot canary (repair budget one), and rollback all passed (94 product assertions); the rollback scenario finished `rolled-back` with no repair attempt and no update-bound model request. The one model request in that window was the updater's existing automatic triage turn after service recovery, which main already runs after a rolled-back run; the lane now records it separately.

**Final head `394d3a2f39a` (run 18, rollback as the first-ever update on a fresh install):** `rolled-back` with reason `restart-unhealthy`, previous version 2026.8.90 running, no repair attempt, no update-bound model request (only the heartbeat and the existing post-recovery triage turn). The lane's follow-on "success" scenario in that run accidentally targeted the crash fixture again because the lane had published it as `latest`; the updater rolled that back correctly as well, which is a second pass of the same path on this head rather than a product failure.

**Second ClawSweeper P1 (fence the config input Doctor consumed):** closed in the follow-up commit fe520959019: Doctor also reports the hash of the config it loaded at start, and the parent marks the activation snapshot Doctor-owned only when that input hash equals its own pre-Doctor capture and the observed output hash equals Doctor's reported output hash; an operator save between the parent's capture and Doctor's read therefore refuses rollback and keeps the edit (regression test added).

steipete:
Merged via squash.

- Prepared head SHA: [394d3a2f39a723c29974b9c1a55d25e853e34232](https://github.com/openclaw/openclaw/pull/141219/commits/394d3a2f39a723c29974b9c1a55d25e853e34232)
- Landed commit: [ee748b6c86d5c1052ff9c540acecc4ae7af096cf](https://github.com/openclaw/openclaw/commit/ee748b6c86d5c1052ff9c540acecc4ae7af096cf)

PR #141212, opened 2026-09-07
fix(test): finish service worker setup before deployment

steipete:
## What Problem This Solves

Resolves an intermittent production service-worker E2E failure where the simulated deployment completes without a build-B worker or cache, leaving the draft-preservation test waiting for an update that never started.

Original failure: https://github.com/openclaw/openclaw/actions/runs/34111989955/job/101714057877

## Why This Change Was Made

The test could replace the served build while a startup update check still held build-A bytes. Chromium can coalesce a concurrent deployment update with that request. The existing controlled-page helper now completes the initial native update check before the test builds and serves the replacement.

## User Impact

More reliable service-worker deployment coverage. No production behavior or UI appearance changes. Existing cache, activation, draft, build-identity, URL, and reload assertions remain unchanged.

## Evidence

A private native HTTP probe held a real build-A worker response through the build-B directory swap. The original test failed **0/1** at its existing 15-second missing-B-cache assertion: the deployment update resolved with active A, no replacement, and only the A cache. With CI’s two-worker ceiling, the original failed **0/1** and the exact five-line repair passed **5/5**; an earlier one-worker control also failed **0/1** before and passed **5/5** after. Each trace confirms the setup update settled before build B was served; the original assertions then passed. The probe preserved worker response bytes and native promise identity; no diagnostic proxy code is included in this PR.

```sh
# Controlled before/after runs use CI's two-worker ceiling:
CI=1 OPENCLAW_VITEST_MAX_WORKERS=2 OPENCLAW_UI_E2E_SKIP_REAL_GATEWAY=1 node scripts/run-vitest.mjs run --config test/vitest/vitest.ui-e2e.config.ts --configLoader runner ui/src/e2e/service-worker-update.e2e.test.ts -t 'recovers a missed worker activation without losing a chat draft'
# Clean full-file sibling verification:
CI=1 OPENCLAW_VITEST_MAX_WORKERS=1 OPENCLAW_UI_E2E_SKIP_REAL_GATEWAY=1 node scripts/run-vitest.mjs run --config test/vitest/vitest.ui-e2e.config.ts --configLoader runner ui/src/e2e/service-worker-update.e2e.test.ts
node scripts/check-changed.mjs
```

Known gap: the historical CI artifact lacks request/lifecycle traces. This proves a reachable matching race, not the exact scheduling of that prior run. All endpoints and data are synthetic; no live-provider claim.

[...]

PR #141225, opened 2026-09-07
test(ui): settle startup worker updates before replacing builds

steipete:
Related: #141047

## What Problem This Solves

Resolves a race where the Control UI missed-activation tests can fail before exercising recovery: an intended update to build B can join an unfinished startup check against build A, leaving only A cached. This fixture repair follows the failure encountered while validating #141047.

## Why This Change Was Made

Wait for the active registration’s startup update while A is still served, before compiling and substituting B. The existing chat/config cases then make their single B update and retain all activation, reload, route and draft assertions. The change adds four test lines and uses the public browser API; production behavior and test deadlines are unchanged.

## User Impact

No runtime behavior change. The production-build tests reliably reach the recovery behavior they are intended to check.

## Evidence

- A controlled Chromium run held a real A validation response, swapped in B and called update: the call fulfilled without fetching B, and B remained absent from the cache for 15 seconds. A fresh-context control joined the A update before swapping; its single B update fetched the real B worker and created B’s cache. The original CI network sequence was not captured, so this demonstrates a concrete fixture race without claiming to reconstruct that run.
- The existing deep-link → chat-draft → config-draft prefix passes all three cases with the fix. The later terminal case was filtered from this focused run.
- All 18 selected repository checks pass; independent P0–P2 review found no actionable findings.

The controlled response experiment proves request/cache ordering. The unchanged application test assertions separately cover activation, reload and draft recovery. Hosted CI for this PR supplies the combined-main validation.

### DIFF
--- a/ui/src/e2e/service-worker-update.e2e.test.ts
+++ b/ui/src/e2e/service-worker-update.e2e.test.ts
@@ -352,6 +352,10 @@ describe("Control UI service-worker production update E2E", () => {
         const draft =
           mode === "chat" ? "keep my draft through the missed update" : '{ "count": 2 }';
         await editor.fill(draft);
+        // Join startup update checks before changing the served worker bytes.
+        await page.evaluate(async () => {
+          await (await navigator.serviceWorker.ready).update();
+        });
         await buildProductionControlUiE2e(nextDir, buildB);
         await page.evaluate(() => sessionStorage.setItem("test-missed-activation", "1"));
         await rename(outDir, previousDir);


steipete:
Closing as independently fixed by #141212, merged as f5f032474c5fc38734a0d378a24b143c922bbabb.

I verified the merged source and current main: the shared `ensureControlledPage` helper now awaits the identical native `registration.update()` after activation/control and before every later build swap. Both chat/config cases already await that helper; it also covers the terminal flow. This PR’s later per-case check would duplicate that synchronization.

The merged tree matches the actual parent plus prepared head, and the test source matches on head, merge and current main. Exact-head CI [34121720362](https://github.com/openclaw/openclaw/actions/runs/34121720362) passed, including all four service-worker cases and all 288 cases in the owning E2E shard. Existing recovery assertions and deadlines remain unchanged. The original CI network interleaving remains unobserved; the controlled proof establishes the matching reachable race.

No additional patch is needed here.

PR #141227, opened 2026-09-07
refactor(memory-wiki): share imported source orchestration

steipete:
## What Problem This Solves

Bridge and unsafe-local wiki imports maintain separate copies of vault initialization, stale-page pruning, sync-state publication, result counting, and ingest logging. That duplication makes authority and failure-ordering guarantees harder to maintain consistently.

## Why This Change Was Made

A plugin-local `syncImportedSourcePages` operation beside the existing shared page writer owns those common phases and their private result type. Each importer keeps its source selection, capacity checks, and execution loop. Bridge pruning authority is read after awaited writes; unsafe-local keeps its 16-worker loop and immediate failure propagation. A failed source batch cannot proceed to prune, publish state, or log success.

## User Impact

No intended behavior change. Human Notes preservation, unreadable-scope retention, abort checks, state publication before logging, and sorted result paths remain unchanged. No schema, stored representation, retention policy, transaction boundary, configuration option, dependency, or public export changes.

Production shrinks by 6 lines; tests add 91 lines. Preserving the two source-specific policies limits the line reduction, while the common lifecycle now has one owner.

## Evidence

- `node scripts/run-vitest.mjs extensions/memory-wiki/src/bridge.test.ts extensions/memory-wiki/src/unsafe-local.test.ts extensions/memory-wiki/src/source-page-shared.test.ts extensions/memory-wiki/src/source-sync.test.ts extensions/memory-wiki/src/source-sync-state.test.ts`: **67 passed**.
- A capability-closes-during-read case proves stale bridge pages are retained until pruning authority returns. A failed-source-read case checks real page/log files and published state remain unchanged. That state assertion uses the existing in-memory fallback; the separate source-sync-state suite covers SQLite persistence.
- CI detected one type-only Madge cycle because the operation imported its result type from its bridge caller. The canonical check reproduced it locally; relocating the private type into the operation owner and migrating all local callers yields **0 Madge cycles**. No public barrel exports this type. All **67 tests** passed again after the move. `pnpm check:architecture` and extension production/test type checks passed, including **0 runtime cycles / 0 Madge cycles**.
- Autoreview is scoped-clean through P3, with zero accepted/actionable findings.
- Changed checks passed ratchets, formatting, extension production/test types, guards, and all three dead-export scans. Extension lint preparation stops at the documented nested-checkout ancestor `punycode` declaration guard; hosted CI must settle that boundary. No ancestor install or guard is modified.
- `pnpm build` passed in **7m12.4s**, including plugin packaging, SDK declarations, and runtime import checks, before the type-only relocation. That correction changes no runtime import graph and has the repeated focused proof above.

The existing page writer and state/concurrency owners remain unchanged. No live provider or Gateway is involved in this filesystem/state orchestration change.

The data-model compatibility request does not require a migration here. The source-state writer/reader (`source-sync-state.ts`), guarded page writer (`source-page-shared.ts`), vault mutation coordinator, config definitions, and core plugin-state store are byte-identical to the PR base. State remains version 1 with the same keys and entry fields, and page generation remains in the same writer. The existing sequential bridge and 16-worker unsafe-local loops produce the same rows; pruning, state publication, and logging retain their order. The existing SQLite-backed state tests passed. The `source-sync.ts` follow-up changes only the private result-type import, with the exact same fields; it does not change a persisted projection. No vector or embedding metadata is modified.

Final exact-head CI passed for `4e3eaaa22cb0e06e1c239dccd88a7fb66fe1e578`: https://github.com/openclaw/openclaw/actions/runs/34125385543.

### DIFF
--- a/extensions/memory-wiki/src/bridge.test.ts
+++ b/extensions/memory-wiki/src/bridge.test.ts
@@ -558,6 +558,56 @@ describe("syncMemoryWikiBridgeSources", () => {
     }
   });
 
+  it("keeps stale pages when the memory capability closes during a source read", async () => {
+    const workspaceDir = await createBridgeWorkspace("closing-capability-workspace");
+    const { rootDir: vaultDir, config } = await createVault({
+      rootDir: nextCaseRoot("closing-capability-vault"),
+      config: { vaultMode: "bridge", bridge: { enabled: true, indexMemoryRoot: true } },
+    });
+    const artifacts: MemoryPluginPublicArtifact[] = ["previous.md", "current.md"].map(
+      (relativePath) => ({
+        kind: "memory-root",
+        workspaceDir,
+        relativePath,
+        absolutePath: path.join(workspaceDir, relativePath),
+        agentIds: ["main"],
+        contentType: "markdown",
+      }),
+    );
+    for (const artifact of artifacts) {
+      await fs.writeFile(artifact.absolutePath, `# ${artifact.relativePath}\n`, "utf8");
+    }
+    registerBridgeArtifacts(artifacts.slice(0, 1));
+    const appConfig: OpenClawConfig = {};
+    const first = await syncMemoryWikiBridgeSources({ config, appConfig });
+    const previousPage = path.join(vaultDir, first.pagePaths[0] ?? "");
+    const previousContent = await fs.readFile(previousPage, "utf8");
+
+    registerBridgeArtifacts(artifacts.slice(1));
+    const readFile = fs.readFile.bind(fs);
+    const readSpy = vi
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4e3eaaa22cb0e06e1c239dccd88a7fb66fe1e578](https://github.com/openclaw/openclaw/pull/141227/commits/4e3eaaa22cb0e06e1c239dccd88a7fb66fe1e578)
- Landed commit: [38bca2a4755554c69148a2127871755d409cda08](https://github.com/openclaw/openclaw/commit/38bca2a4755554c69148a2127871755d409cda08)

PR #141229, opened 2026-09-07
refactor(plugins): share setup presentation URL policy

steipete:
## What Problem This Solves

Plugin auth choices and recommended installs maintain two copies of the same HTTPS presentation URL policy. Keeping the canonicalization, credential rejection and length limit in sync currently requires editing both implementations.

## Why This Change Was Made

Share that policy in one private plugin helper and remove both copies. Each caller keeps its existing record handling: auth choices omit invalid optional URLs, while recommended installs reject entries missing valid required URLs and retain their ordering and duplicate-ID rules.

## User Impact

Behavior is unchanged. URLs still require HTTPS, a hostname, no credentials, and at most 2,048 UTF-16 code units after canonicalization. This maintainer-requested consolidation removes 16 production code lines (17 physical lines), with no test changes or new public API.

## Evidence

All 142 existing manifest, recommended-install and management tests pass before and after. Differential checks using the final helper match 106 URL comparisons and 2,809 cases for each caller, including invalid entries followed by a valid duplicate. The selected changed gate and independent P2 review pass. These are metadata parsing checks; no live fetching or plugin activation is involved.

### DIFF
--- a/src/plugins/manifest-setup-normalizers.ts
+++ b/src/plugins/manifest-setup-normalizers.ts
@@ -28,6 +28,7 @@ import type {
   PluginManifestSetupProviderAuthEvidence,
   PluginConfigUiHint,
 } from "./manifest-types.js";
+import { normalizeSetupPresentationHttpsUrl } from "./setup-presentation-url.js";
 
 export function normalizeManifestActivation(value: unknown): PluginManifestActivation | undefined {
   if (!isRecord(value)) {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [62f0567d7023fb1d7a6e2d8280e7f9e68e8d6ee5](https://github.com/openclaw/openclaw/pull/141229/commits/62f0567d7023fb1d7a6e2d8280e7f9e68e8d6ee5)
- Landed commit: [0ffd99975726cfc2028c9ec7fc2e7c8040c69b9c](https://github.com/openclaw/openclaw/commit/0ffd99975726cfc2028c9ec7fc2e7c8040c69b9c)

PR #141235, opened 2026-09-07
refactor(codex): share node catalog record lookup

steipete:
## What Problem This Solves

Continuing a Codex session on a paired node and opening its terminal maintain separate copies of the same bounded catalog lookup. Changes to RPC arguments, response parsing or cursor handling currently have to remain synchronized across both copies.

## Why This Change Was Made

One private plugin helper now owns the sequential request, real page parser, first matching record and bounded cursor traversal. Both old lookup functions are removed, along with a private parser parameter whose sole production caller always supplied the canonical parser. Continuation and terminal operations retain their distinct eligibility checks and error messages.

## User Impact

Behavior is unchanged. Lookup keeps the same node command, scopes, 100-item pages, 100-page maximum, 65-second request timeout, cursor-cycle handling and first-match ordering. Active interactive sessions remain eligible for terminal opening; continuation keeps its existing state restrictions. The maintainer-requested consolidation removes 27 production code lines (28 physical lines), with no test changes or new public API.

## Evidence

All 40 existing node-listing and continuation tests pass before and after. A further 666 comparisons use the real helper, parser and response unwrapping to verify both callers across direct, payload and payloadJSON responses, including page 100, malformed or archived rows, cycle/match ordering and exact error causes. All 24 selected checks and independent P2 review pass.

The relevant upstream Codex listing request/response contract was inspected directly. Verification here uses synthetic transport plus existing provider-boundary tests; it does not claim a live Gateway, paired-node or Codex request. No runtime dependency or package-export change is introduced.

### DIFF
--- a/extensions/codex/src/session-catalog-node-continue.ts
+++ b/extensions/codex/src/session-catalog-node-continue.ts
@@ -21,17 +21,16 @@ import {
   type CodexNodeHistory,
   type CodexSessionDisposition,
 } from "./session-catalog-node-adoption.js";
+import { lookupNodeCodexCatalogRecord } from "./session-catalog-node-lookup.js";
 import {
   catalogError,
   CatalogParamsError,
   CODEX_APP_SERVER_THREADS_LIST_COMMAND,
   CODEX_APP_SERVER_THREAD_TURNS_LIST_COMMAND,
-  CODEX_SESSION_CATALOG_MAX_PAGE_LIMIT,
   MAX_SESSION_ID_LENGTH,
   filterCatalogPageByTitle,
   isInteractiveThreadSource,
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [26654e69d624f8e2da81d74c007b49af11e5982d](https://github.com/openclaw/openclaw/pull/141235/commits/26654e69d624f8e2da81d74c007b49af11e5982d)
- Landed commit: [5260a2fe594bbdc475d317983984cfa915bbc2fc](https://github.com/openclaw/openclaw/commit/5260a2fe594bbdc475d317983984cfa915bbc2fc)

PR #141237, opened 2026-09-07
fix(voice-call): flush media-stream upgrade rejects before destroy

hugenshen:
## What Problem This Solves

Voice-call media-stream upgrade rejects (shutdown and connection-cap) could destroy the HTTP socket before rejection bytes flushed, so clients saw a hard disconnect instead of HTTP 503. Peer reset during that write could also become `uncaughtException`.

## Why This Change Was Made

`rejectMediaStreamUpgrade` attaches `socket.once("error", () => {})` then reuses `rejectWebSocketUpgrade`, which calls `socket.end(response, () => destroy())`. Cap and shutdown rejects share that owner.

## User Impact

**Before:** Capacity/shutdown upgrade rejects could reset the client with no HTTP status, and a reset peer during write could crash the process.

**After:** Clients receive `HTTP/1.1 503` with the reject body; peer-reset errors during flush are swallowed.

## Evidence

- Production: `extensions/voice-call/src/media-stream.ts` `rejectMediaStreamUpgrade`
- Tests: `extensions/voice-call/src/media-stream.test.ts` — real HTTP 503, injected-hold flush on a real TCP socket, and RST during pending `socket.end`

## Real behavior proof

- **Behavior or issue addressed:** Media-stream upgrade rejects could drop the HTTP status on buffered/reset sockets.
- **Canonical reachability path:** `MediaStreamHandler.handleUpgrade` → `rejectMediaStreamUpgrade` → `rejectWebSocketUpgrade` against a live loopback upgrade.
- **Boundary crossed:** Real `node:http` / `net.connect` transport. Buffering is injected by holding `socket.end` on that real socket, then releasing so the 503 bytes flush. RST destroys the client only after the server has entered `socket.end`.
- **Shared helper / provider constraint check:** Reused `openclaw/plugin-sdk/websocket-runtime` `rejectWebSocketUpgrade`.
- **Real environment tested:** Local Vitest extension-voice-call lane, loopback TCP. No carrier/video.

```text
$ git diff --check
$ node scripts/run-vitest.mjs extensions/voice-call/src/media-stream.test.ts --reporter=verbose -t "returns HTTP 503 over a real upgrade"
$ node scripts/run-vitest.mjs extensions/voice-call/src/media-stream.test.ts --reporter=verbose -t "flushes a buffered 503 over real transport"
$ node scripts/run-vitest.mjs extensions/voice-call/src/media-stream.test.ts --reporter=verbose -t "peer resets during a 503 reject"
```

- **Evidence after fix:**

```text
✓ |extension-voice-call| ... > returns HTTP 503 over a real upgrade when the connection cap is reached 13ms

stdout | ... > flushes a buffered 503 over real transport after an injected hold
[media-stream buffered transport proof] pending_before_release=true status=503 body=Too many media stream connections
 ✓ ... > flushes a buffered 503 over real transport after an injected hold 65ms

stdout | ... > does not surface uncaught errors when the peer resets during a 503 reject
[media-stream rst proof] rejection_started=true uncaught=0
 ✓ ... > does not surface uncaught errors when the peer resets during a 503 reject 65ms
```

- **Observed result after fix:** Real transport returns HTTP 503. Holding `socket.end` keeps the reject pending until release, then the client reads the 503 body. Destroying the client at `socket.end` does not emit `uncaughtException`.
- **What was not tested:** Live Twilio/Telnyx media-stream against a full carrier call.

- [x] AI-assisted (Cursor)


hugenshen:
@clawsweeper re-review

hugenshen:
@clawsweeper re-review

PR #141239, opened 2026-09-07
fix(gateway): trim device.scopes.waitUpgrade requestId

hugenshen:
## What Problem This Solves

A device permission upgrade could fail after copying its request ID with surrounding whitespace: approval succeeded, but the waiter reported that the request had expired or did not exist. Real Control UI validation also found that approval discarded an existing token issuer binding, so the returned upgraded token could not reconnect.

## Why This Change Was Made

Normalize only the incoming wait ID before the existing owner-bound lookup. Keep stored IDs, returned IDs, approval IDs, and device identity checks unchanged.

Normal pairing approval now uses the existing token constructor and preserves the existing issuer binding, as ordinary token rotation already does. It retains the original generation rather than assigning a new one. Old tokens, revoked tokens, and tokens from an old shared-auth generation remain rejected. Bootstrap approval and the verifier are unchanged.

The contributor's commit remains in ancestry. The repair removes duplicate token construction and keeps the real public regression and omits the redundant mock-only test. Production code decreases by two lines; the two regression tests add 87 lines.

## User Impact

Copied request IDs with surrounding whitespace resolve the same approved request. The returned Control UI token reconnects with its approved permissions while retaining the normal revocation and role restrictions. This changes no schema, configuration option, stored-ID format, or public API.

## Evidence

- Pinned main `e53b3dea864cf8eee366267a75847c3db4bd43c0`: a real isolated Gateway accepted the device's request and administrator approval, but the padded waiter failed with `scope upgrade expired or not found`.
- Independent validation of the first adapted candidate found that both canonical and padded approval returned tokens that failed reconnect. The complete failed report remains retained.
- The issuer regression failed with `legacy-browser-token` before the producer repair. It now proves same-generation token use, rejection after generation change, and rejection of the replaced token.
- All 79 pairing tests and 12 real WebSocket cases pass on the corrected source. Formatting and ordinary lint pass.
- Corrected runtime: commit `e884265bccdf012da56349cdcff64ce2f1aae291`, build `2026.9.3-e884265bccdf-2026-09-10T00-38-31.313Z`. Fresh independent compiled acceptance passed: canonical and padded approval reconnect with the returned scopes; replaced and revoked tokens are refused; current-role checks deny elevated results after approval; actual shared-auth rotation invalidates the previously usable upgraded token. Same-device invalid-ID controls also pass. Final exact-head review and CI remain merge gates.
- The retained natural-expiry control observed the real five-minute deadline without changing clocks or stored requests. Source/process guarantees remain supported by the independent code-aware review; no source-blind storage inspection is claimed.

Complete private observations preserve all attempts and fixture corrections. Published evidence excludes tokens, signing keys, private infrastructure, and execution metadata.

AI-assisted.


hugenshen:
@clawsweeper re-review

hugenshen:
@clawsweeper re-review

PR #141246, opened 2026-09-07
fix: stop idle remote sessions without waiting for unrelated provisioning

steipete:
## What Problem This Solves

Fixes an issue where users stopping an idle remote session would wait for a different session's slow cloud provisioning before cleanup could begin.

## Why This Change Was Made

The placement coordinator used the same readiness promise for running cleanup and releasing its scheduling fence. Stop now proceeds after its own preparation and existing predecessor fence settle, without waiting for unrelated active dispatches. The original global-idle obligation stays in the fence, preserving admission ordering for later dispatch, move, sweep, and recovery operations.

The existing same-session test now models the real preparation dependency and checks it after yielding. Both shared test-harness defaults also settle that supplied dependency before running reclaim, while preserving explicit custom preparations. No authorization, cancellation policy, placement storage, protocol, or configuration contract changes.

## User Impact

An idle session can finish Stop while another session is still provisioning. Earlier queued maintenance can still delay Stop, and later Starts still wait behind the retained fence. This does not solve all provisioning delays or establish a cold-start latency target.

## Evidence

- Applied the new ordinary disjoint-session regression before the production change: it failed on current main at the progress assertion, not from setup or timeout. Its cleanup releases and joins every pending operation.
- With the fix, 16 admission cases, 11 related same-session/move/sweep/recovery ordering cases, and the real service's ordinary active-placement teardown case pass.
- The first hosted run exposed a test harness that skipped the real preparation dependency. That failure was reproduced locally; the fixture-only repair preserves all assertions and adds complete promise cleanup. Seven reclaim cases and eleven recovery/shutdown cases pass afterward.
- Canonical changed-file checks pass, including core and core-test typechecking, lint, formatting, dead-export scans, and current repository guards.
- Independent source review and canonical isolated P0–P2 autoreview found no actionable findings.

This candidate has not been deployed or live-tested. The focused tests exercise the changed scheduling boundary; live Gateway verification is a separate follow-up. Production changes are +23/-13 lines, separating operation readiness from the existing fence obligation in one private helper.

### DIFF
--- a/src/gateway/worker-environments/placement-dispatch-coordinator-admission.test.ts
+++ b/src/gateway/worker-environments/placement-dispatch-coordinator-admission.test.ts
@@ -14,6 +14,56 @@ import {
 import type { WorkerPlacementDispatchRequest } from "./service-contract.js";
 
 describe("worker placement maintenance admission", () => {
+  it("reclaims an idle session before a disjoint dispatch finishes while preserving its fence", async () => {
+    const cloudStarted = createDeferredCore();
+    const releaseCloud = createDeferredCore();
+    let reclaimed = false;
+    const dispatch = vi.fn(async (request: WorkerPlacementDispatchRequest) => {
+      if (request.sessionId === "cloud") {
+        cloudStarted.resolve();
+        await releaseCloud.promise;
+      }
+      return { ...ACTIVE_PLACEMENT, ...request };
+    });
+    const service = createCoordinatorTestService({
+      dispatch,
+      reclaim: async (_request, _authorize, _beforeDrain, serialize) => {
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d5a5b5d253337809830d83896863b503355657e5](https://github.com/openclaw/openclaw/pull/141246/commits/d5a5b5d253337809830d83896863b503355657e5)
- Landed commit: [dcc68718f57e196bef10a07178fc8061b896140c](https://github.com/openclaw/openclaw/commit/dcc68718f57e196bef10a07178fc8061b896140c)

PR #141045, opened 2026-09-07
feat: show active sessions in Live activity

shakkernerd:
## What Problem This Solves

Fixes an issue where users opening Live activity could see no current work even while sessions were already running or queued. The received-event feed cannot reconstruct activity that happened before the browser connected.

## Why This Change Was Made

An **Active sessions** section loads the Gateway's current session state on entry and reconnect. The existing session list API selects active work before pagination, using live runtime owners and current visibility checks. Each row retains its agent, exact session key, and session generation so sessions from different agents cannot replace one another.

The browser reconciles current state with session updates and refreshes when an in-flight response cannot safely represent newer run identities. Agent-owned `global` and `unknown` sessions remain distinct from ordinary sessions with similar names. The new active-only projection omits ambiguous optional descendant metadata on those reserved rows.

## User Impact

Operators can see up to 100 permitted running or queued sessions immediately, with a limit notice when more match. The bounded section stays visible as the received-event feed grows. Linked rows open their session. Raw `unknown` sessions show status without a conversation link. Raw `global` rows also remain status-only outside global session scope, where Home would open a different session. Literal qualified keys keep their normal links.

The received-event feed keeps its existing bounds and Clear behavior. Clear does not hide or stop active sessions. Loading, disconnection, and query failure have visible states. This change adds no session-filter control or historical tool-event replay.

## Evidence

- Completed affected proof covers 606 Gateway/storage/observer/Activity checks and all 17 bundled Activity browser cases, plus protocol, type, translation and stylesheet gates.
- The busy-feed regression checks physical clipping, alongside the existing scroll and auto-follow behavior. Original-shrink negative controls fail on desktop and short viewports; the final stylesheet preserves running/queued rows and empty feedback across all four viewport controls.
- Full code review and the focused CSS correction review found no actionable P0–P2 defects. [Current CI results](https://github.com/openclaw/openclaw/pull/141045/checks) are attached to this PR. The iOS scan finding was independently reproduced on `main`; its separate repair is [#141152](https://github.com/openclaw/openclaw/pull/141152).
- Both complete recordings and the paired screenshots below were inspected. The required documentation rebase preserves every affected non-doc source file, the captured Activity modules and stylesheet; capture heads remain explicit.

## Visual evidence

| View | Before | After |
| --- | --- | --- |
| Opening | Live Activity has no current-work overview when the page opens | Active sessions shows one running and one queued session on opening |
| Completion | Only the received completed tool is visible | Queued work remains current above the received completed tool |
[...]

PR #141247, opened 2026-09-07
fix: keep active sessions scoped to their agent

shakkernerd:
Related: #141045

## What Problem This Solves

Fixes an issue where Activity and other session snapshots could mark an idle agent’s session as running when different agent stores contained the same session ID.

## Why This Change Was Made

The session presenter now matches retained embedded and reply activity to the owning agent. Reply admission carries that identity through queued work and command continuation, including an unchanged raw global key. Codex and Copilot pass their resolved owner when registering a run; legacy registrations retain the agent from an already validated host binding.

## User Impact

Only the owning agent’s session appears active, including after its transient run context has been released.

## Evidence

- 479 focused tests passed: 407 retained core/caller regressions and 72 final-head runtime, Codex, and Copilot tests. These cover the original SQLite/session-list collision, queued owner adoption, retained ownership, aliases, settlement, scoped abort, ordinary cached reads, and existing authority controls.
- Full [CI at 0fd43e9b](https://github.com/openclaw/openclaw/actions/runs/34132418505) passed, including production and test type checks. Changed-file formatting, lint, source limits, and SDK documentation validation passed.
- Actual Copilot registration and four-argument legacy SDK registration inside a live host binding failed before the producer correction and passed afterward.

## Visual evidence

| Before | After |
| --- | --- |
| Before: both the saved main session and Operations work are shown as running | After: only Operations work is shown in Active sessions |

### Final-head response and visible owning session

https://github.com/user-attachments/assets/206eb1e6-2fe2-488b-8d5e-17757e3fe433

### Visual verification

- Baseline: 5d895d3d52a78751413eac0e59399da9b91cb026
- Exact head: `0fd43e9b37c6243de9ed77d57762e6008ac57035`
- Scenario: Two agents have raw global sessions with the same restored session ID, while only ops owns an active embedded run. The real SQLite-backed sessions.list handler produces each response; the unchanged Control UI displays the baseline's incorrect main+ops membership and the fixed ops-only membership.
- Runtime/build: Real Control UI source-mode runtime at 0fd43e9b, Node 24.19.0 and Chromium 144.0.7559.0. The baseline response is from 5d895d3d and was captured with byte-identical UI code at 6e2c78db. Both use 1440x1000, dark theme, en-US, UTC, reduced motion, blocked service workers and synthetic data.
- Active owners returned and rendered: Before: main,ops; after: ops
- Projected-only control: Only ops on both versions before the embedded handle is attached
- Backend/query fidelity: Real isolated SQLite stores and sessions.list with the same activeOnly query used by Live activity
- Final recording: 5.32 seconds, 133 frames; all frames decode; no page errors
- Focused native tests: 407 retained core/caller passes plus 72 current-head runtime/Codex/Copilot passes
- All 22 final changed-file hashes and the producer were checked against 0fd43e9b; the tested sources remained unchanged.
- The UI tree is identical at baseline 5d895d3d, the preserved baseline renderer 6e2c78db, and final head 0fd43e9b.
- The original recordings were inspected from initial load through roster settlement. Both scenarios leave the existing Activity history empty.
- The recorder uses the unchanged repository mock Gateway with actual server-produced session membership; the evidence captions do not force rendered rows.

### DIFF
--- a/docs/plugins/sdk-agent-harness.md
+++ b/docs/plugins/sdk-agent-harness.md
@@ -514,6 +514,11 @@ enforce the assertion, leave V2 unsupported.
 
 The V1 `messageInjection`, queue options, `queueAgentHarnessMessage`, and
 `setActiveEmbeddedRun` signatures shipped in v2026.8.1 remain source-compatible.
+Pass the resolved agent ID as the fifth `setActiveEmbeddedRun` argument so raw
+`global` and `unknown` keys retain their owner. Legacy calls inside a matching
+live host binding inherit its validated agent; an ambient caller alone does not
+supply ownership. Outside that binding, omitted ownership uses the qualified
+session key or the configured default agent for session activity.
 Unscoped V1 injection retains its existing behavior. Source-bound controls
 require V2 and reject visibly before queue or I/O when only V1 is available;
 they never fall back to an unchecked V1 callback. Existing deprecation windows
--- a/extensions/codex/src/app-server/run-attempt-active-turn.ts
+++ b/extensions/codex/src/app-server/run-attempt-active-turn.ts
@@ -547,7 +547,13 @@ export function activateCodexAttemptTurn(
     abort: () => abortExplicitly("aborted"),
   };
   params.replyOperation?.attachBackend(handle);
-  setActiveEmbeddedRun(params.sessionId, handle, params.sessionKey, params.sessionFile);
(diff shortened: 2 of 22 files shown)


shakkernerd:
Fixed: session snapshots keep embedded and queued reply activity with the correct owner when saved session IDs collide.

Issue #137116, opened 2026-09-03
[Bug]: Terminal upload batches have no aggregate staging bound


shakkernerd:
## Summary

Terminal uploads enforce a 16 MiB per-file limit but have no aggregate retained-byte or directory limit for multi-file batches.

Each accepted file receives a separate temporary directory and cleanup timer and remains staged for up to 24 hours. The supported multi-file upload flow can therefore grow temporary storage, directory count, and timer count without a process-level or session-level bound.

Reproduced on `204431bb8004a2da9d5b49a680a68a31624a4214`. The affected paths remained unchanged through `6d763ee148d4ee62c2c1a39fd5726e1a4d0b94a2`.

## Steps to reproduce

1. Start the terminal-upload staging owner with an isolated temporary root and shortened retention interval.
2. Submit 65 files through the production staging function.
3. Make each file slightly larger than 4 MiB and smaller than the supported 16 MiB per-file limit.
4. Count the retained files, bytes, directories, and cleanup timers before expiration.
5. Observe that all 65 uploads are accepted, retaining 285,212,672 bytes across 65 directories.
6. Wait for the shortened retention interval and confirm that the directories are eventually removed.

## Current behavior

`packages/gateway-protocol/src/schema/terminal-constants.ts` defines a 16 MiB maximum for each file.

The authenticated `terminal.upload` RPC reaches `src/infra/terminal-file-upload.ts`, which creates a new temporary directory, writes the decoded file, and schedules an independent 24-hour cleanup timer for every successful upload.

The multi-file picker in `ui/src/components/terminal/terminal-panel-upload.ts` accepts an arbitrary `File[]` and uploads every selected file sequentially.

No aggregate byte, directory-count, batch, session, or process limit is enforced. Startup recovery also creates one timer for every unexpired directory it discovers.

## Expected behavior

The terminal staging owner should enforce an aggregate retained-byte and directory limit across supported upload batches.

New uploads should be rejected with a clear error when accepting them would exceed that budget.

## Impact

An authenticated `operator.admin` using the ordinary multi-file upload flow can cause temporary disk usage, directory count, and cleanup-timer count to grow linearly for 24 hours.

The reproduction retained approximately 272 MiB across 65 directories, exceeding the 256 MiB and 64-directory limits already used by the Browser staging implementation. Larger batches can exhaust the Gateway host’s temporary volume and disrupt uploads or other disk-dependent operations.

[...]

PR #141251, opened 2026-09-07
fix: preserve recovered terminal upload expiry

shakkernerd:
Related: #137116

## What Problem This Solves

Fixes an issue where recovered terminal uploads could remain for an extra day after a file rename or a partially failed cleanup.

## Why This Change Was Made

The cleanup owner retains each observed directory's original expiry and full-precision filesystem identity. Records are removed when uploads disappear, and one timer continues to manage the staging root, including existing inventories above the admission limit. Timestamp conversion preserves fractional milliseconds.

## User Impact

Renaming a staged file and retrying failed cleanup preserve its scheduled deadline. A replacement directory at the same path receives its own deadline, including on filesystems with large 64-bit identifiers. The existing 24-hour retention and upload storage limits are unchanged.

## Evidence

- All 20 upload-owner tests pass locally and on native Windows Server 2022, covering renames in one and 65 directories, partial cleanup, directory replacement, large file identifiers, future timestamps, byte/count admission, unreadable inventory and write failures.
- Real filesystem probes confirm deletion at the original recovered deadline and the scheduled cleanup retry after permission recovery. On native Windows, a replacement with an identifier beyond JavaScript Number precision survives the old deadline and expires at its own deadline.
- Linux probes retain 1,024 legacy directories with one unreferenced timer; the timer retires when the removed inventory is observed.
- All 119 Linux owner/Gateway/node-invoke tests, the complete three-path `check:changed` gate, and exact-head hosted CI pass. Independent review is clean through P2.


shakkernerd:
Fixed: recovered terminal uploads keep their scheduled expiry through renames and partial cleanup while replacement directories retain their own deadlines.

PR #141258, opened 2026-09-07
refactor(tts): share readiness fallback fixtures

steipete:
## What Problem This Solves

The speech and telephony readiness-fallback tests repeat the same fixture setup and result assertions. Keeping that shared contract aligned currently requires editing both copies.

## Why This Change Was Made

Run the common fixture and assertions through two explicit table rows. Both public APIs still execute independently, retaining their original test names, provider identities, input text, whitespace variants and telephony capability difference. Each row still checks all 20 result assertions, including fallback order, attempt classification, latency and error redaction.

## User Impact

Production behavior is unchanged. This maintainer-requested test cleanup removes 43 test/support code lines (46 physical lines). The helper installs Vitest suites and is called only by the contract test; its filename does not make this a production-code reduction. Public TTS testing exports remain unchanged.

## Evidence

The complete contract and boundary suites pass all 48 tests before and after with identical test names. All 31 selected checks and independent P2 review pass. Source outside the two fixture bodies is unchanged, and normal speech still receives no telephony callback.

### DIFF
--- a/src/plugins/contracts/tts-contract-suites.ts
+++ b/src/plugins/contracts/tts-contract-suites.ts
@@ -1008,148 +1008,102 @@ export function describeTtsProviderRuntimeContract() {
     });
 
     describe("fallback readiness errors", () => {
-      it("continues synthesize fallback when primary readiness checks throw", async () => {
-        await withIsolatedSpeechProviderEnvAsync({}, async () => {
-          const throwingPrimary: SpeechProviderPlugin = {
-            id: "openai",
-            label: "OpenAI",
-            autoSelectOrder: 10,
-            resolveConfig: () => ({}),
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1f4faba832a85b43ee13f087ab18bb4b90898fe8](https://github.com/openclaw/openclaw/pull/141258/commits/1f4faba832a85b43ee13f087ab18bb4b90898fe8)
- Landed commit: [3b7bc9cb9de3098c2ae67cf42ec59556fb02855d](https://github.com/openclaw/openclaw/commit/3b7bc9cb9de3098c2ae67cf42ec59556fb02855d)

PR #140422, opened 2026-09-06
fix(codex): restore installed skills with catalog-backed models

Colton-Harris:
Related: #95058 (Codex/OpenClaw skill prompt-contract differences).

## What Problem This Solves

An ordinary Codex-backed OpenClaw agent could not use installed skills when the active model supplies its own collaboration-mode instructions. The OpenClaw skill snapshot and the system-prompt report could both show the complete catalog while the model received none of it.

A second, narrower failure is fixed in this revision, reported by review at the previously reviewed head `164d85a4`: on an **incognito** conversation whose catalog was refreshed mid-session (an edited or deleted skill), a context compaction silently reverted the model to the catalog the conversation started with. The user saw an edited skill keep its stale description, or a deleted skill stay visible, with nothing in the UI indicating it.

## Why This Change Was Made

Codex intentionally prefers `model_messages.collaboration_modes` over caller-supplied `collaborationMode.settings.developer_instructions` (openai/codex#36351). OpenClaw was using that replaceable field to transport its skill catalog, so a catalog-backed model silently replaced it.

The eligible skill catalog now travels with the ordinary **thread** developer instructions, after the generic policy. Model-owned collaboration policy is preserved, and the catalog is no longer replaceable.

## Skill catalog lifetime

The first revision folded the catalog into the generic policy string. That made an ordinary skill edit look like a generic-policy change, and a live incognito thread would refuse the next turn with `CodexIncognitoPolicyChangeError`. That was a real regression and it is fixed.

The catalog is now a **separate lifecycle input** (`skillsInstructions`), joined onto the wire developer carrier at request build rather than mixed into the policy:

- **Incognito / ephemeral threads** — the retained record keeps the immutable generic policy and the catalog last delivered separately. Warm reuse compares **only** the generic policy, so the guard is unchanged and still fail-closed. A changed catalog is delivered to the live thread with `thread/inject_items`, and an unchanged catalog is not re-sent. Editing a skill no longer ends an incognito conversation; changing generic policy still refuses the turn.
- **Persistent threads** — the catalog rides the thread developer carrier, so it participates in the thread-config fingerprint. A changed catalog invalidates warm reuse and cold-resumes the **same** thread with the new catalog through the existing policy handoff. The conversation is preserved.

### Catalog lifetime across compaction (review finding, fixed)

[...]

PR #125783, opened 2026-08-18
fix(codex): inherit agent workspace instructions in native subagents

obviyus:
## What Problem This Solves

Follow-up to #125675. In external-cwd sessions, the agent workspace's `AGENTS.md` rode the turn-scoped collaboration developer-instructions channel — which native Codex subagents deliberately do not inherit. Subagents spawned in worktree sessions saw the visited repo's `AGENTS.md` but never the agent's own standing instructions.

## Why This Change Was Made

Move the agent-workspace `AGENTS.md` snapshot from the turn-scoped channel to **thread-level** `developer_instructions`, set once at Codex thread creation and only when the session executes outside the configured agent workspace. Verified against sibling `../codex` source before building:

- thread creation accepts `developer_instructions` and lands in thread config: `codex-rs/app-server-protocol/src/protocol/v2/thread.rs:101`, `codex-rs/app-server/src/request_processors/thread_processor.rs:1121,1181,1603,1621`, `codex-rs/core/src/config/mod.rs:633`
- subagents inherit thread developer instructions: `codex-rs/core/src/tools/handlers/multi_agents_common.rs:204,213`, `codex-rs/core/src/agent/control/spawn.rs:639,650`, child context `codex-rs/core/src/session/turn_context.rs:675`
- turn collaboration instructions stay separate and un-inherited: `codex-rs/core/src/session/turn_context.rs:172,174,204,210`
- both layers coexist in initial context: `codex-rs/core/src/session/mod.rs:3474,3494,3612`

`SOUL.md`/`IDENTITY.md`/`USER.md` remain turn-scoped (their subagent exclusion is intentional). The snapshot is bounded (existing truncation caps), provenance-labeled, frozen in the session binding so resumes reuse identical bytes (prompt-cache-friendly; workspace edits take effect next session per repo prompt-state policy). Canonical same-dir sessions are unchanged — native project-doc discovery owns the workspace file there.

## User Impact

Native Codex subagents spawned from worktree/external-cwd sessions now carry the agent's standing `AGENTS.md` rules instead of being rule-blind.

## Evidence

- Verification citations above (direct `../codex` inspection @ 32a383c0ba by the acting worker; independently re-verified for spawn inheritance and single-injection at thread create).
- Focused Codex extension suite: 886 passed; 3 pre-existing `approval-bridge.test.ts` failures unrelated to this diff (also present before the change; exact-head CI will adjudicate).
- `pnpm tsgo` + `pnpm check:test-types`: pass.
[...]

PR #96795, opened 2026-06-25
fix(codex): move turn-scoped workspace context to thread-level developer instructions

evan-YM:
## Summary

DeepSeek KV cache alternates between ~20% and ~99% on consecutive API calls within the same user turn because workspace soul/identity/user files were sent through a channel that Codex app-server injects at variable positions. Move them to a stable channel.

- **Problem**: Turn-scoped workspace files (SOUL.md, IDENTITY.md, USER.md) rendered as `turnScopedDeveloperInstructions` and placed in `collaborationMode.settings.developer_instructions`. This channel is injected at the front of the messages array on full-turn calls but differently (or not at all) on tool-continuation calls, creating two different message prefixes that prevent DeepSeek KV cache reuse.
- **Solution**: Move `turnScopedDeveloperInstructions` from per-turn `collaborationMode.settings.developer_instructions` into thread-level `developerInstructions` (via `baseDeveloperInstructions`), where they join other workspace files (AGENTS.md, TOOLS.md) in a stable position that is consistent between full-turn and tool-continuation calls.
- **What changed**:
  - `extensions/codex/src/app-server/run-attempt.ts`: Added `workspaceBootstrapContext.turnScopedDeveloperInstructions` to `baseDeveloperInstructions`
  - `extensions/codex/src/app-server/thread-lifecycle.ts`: Removed `options.turnScopedDeveloperInstructions` from `buildTurnScopedCollaborationInstructions` context instructions
  - `extensions/codex/src/app-server/run-attempt.test.ts`: Updated assertions — thread/start now includes soul/identity/user, turn/start collaborationMode no longer includes them
  - `extensions/codex/src/app-server/thread-lifecycle.test.ts`: Updated assertions to match new routing
- **What did NOT change**: Memory and skills collaboration instructions remain in the turn-scoped channel (they are truly dynamic). Default/Plan collaboration mode protocol (`mode`, `model`, `reasoning_effort`) is untouched. Workspace file rendering and content are identical — only the delivery channel changed.

### Change Type (multi-select)

- [x] Bug fix
- [ ] Feature
- [ ] Refactor (must be explicitly requested by maintainer)
- [ ] Docs
- [ ] Security hardening
- [ ] Chore/infra

### Scope (multi-select)

- [ ] Gateway / orchestration
- [ ] Skills / tool execution
- [ ] Auth / tokens
- [ ] Memory / storage
- [x] Integrations
- [ ] API / contracts
- [ ] UI / DX
- [ ] CI/CD / infra

[...]

Issue #96773, opened 2026-06-25
[Bug]: Turn-scoped developer instructions (SOUL.md/IDENTITY.md/USER.md) create different message prefixes between full-turn and tool-continuation calls, breaking DeepSeek KV cache (20%/99% oscillation)


StewartXiang:
### Bug type

Performance regression — prompt cache hit rate degradation

### Summary

In long conversational sessions, DeepSeek KV cache hit rate alternates between **~20% and ~99%** on consecutive API calls within the **same user message turn**. This wastes ~15M input tokens per long session (~$27 at DeepSeek pricing).

Root cause traced to **`CODEX_TURN_SCOPED_WORKSPACE_DEVELOPER_CONTEXT_BASENAMES`** (SOUL.md, IDENTITY.md, USER.md) being injected as developer instructions differently between full-turn calls and tool-continuation calls, creating two different message prefixes that prevent cache reuse.

### Affected versions

- First observed: 2026.6.5
- Still present: **2026.6.10** (latest)

### Mechanism (code trace)

In `run-attempt-*.js`:

```javascript
// Line 211-215: These files are marked as turn-scoped
const CODEX_TURN_SCOPED_WORKSPACE_DEVELOPER_CONTEXT_BASENAMES = new Set([
    "identity.md",
    "soul.md", 
    "user.md"
]);

// Line 8285-8291: Turn collaboration developer instructions differ between calls
const buildCodexTurnCollaborationDeveloperInstructions = () => 
    buildTurnCollaborationMode(params, {
        turnScopedDeveloperInstructions: workspaceBootstrapContext.turnScopedDeveloperInstructions,
        ...
    }).settings.developer_instructions ?? void 0;

const buildRenderedCodexDeveloperInstructions = () => 
    joinPresentSections(promptBuild.developerInstructions,
        buildCodexTurnCollaborationDeveloperInstructions());
```

`buildTurnCollaborationMode()` returns different developer instructions for full-turn calls vs tool-continuation calls. These instructions are injected at/near the **front** of the messages array, so the two call types produce different prefixes → DeepSeek KV cache cannot serve the conversation history on full-turn calls.

### Observed pattern

Every consecutive API call within the same user message alternates:

```
Turn N   (full call):        hit=20.0%  in=108k  cache=27k   ⚠️ only system prompt cached
Turn N+1 (tool continuation): hit=99.9%  in=132   cache=138k  ✅ full history cached
[...]

PR #141270, opened 2026-09-07
fix(codex): carry workspace persona and memory on the thread developer carrier

mbelinky:
Builds on #140422 (skill catalog carrier) and includes its commits, merged with current `main` (that PR currently conflicts with `main` in `run-attempt-context.ts` and `run-attempt-prompt.ts`; the resolution is in the merge commit here). The persona change itself is the single commit `fix(codex): carry workspace persona and memory on the thread developer carrier`. If #140422 lands first, this rebases onto it and shrinks to that one commit. Related: #95058.

## What Problem This Solves

A Codex-backed OpenClaw agent silently loses its persona when the active model supplies its own collaboration-mode instructions. `SOUL.md`, `IDENTITY.md`, `USER.md`, and the workspace-memory pointer were carried in `turn/start.collaborationMode.settings.developer_instructions`. Codex prefers `model_messages.collaboration_modes` over that field (openai/codex#36351), so on a catalog-backed model such as `gpt-6-astra` the model never sees them, while the OpenClaw prompt report and the recorded turn context both show them as loaded.

Observed on two gateways. On the affected agent's recorded Codex rollout, every `turn_context.collaboration_mode.settings.developer_instructions` is 30,364 chars and contains `<AGENT_SOUL>`, yet no developer-role item in the model-facing history contains it. Switching the same thread to `gpt-5.6-sol` (no catalog collaboration message) makes the first developer message 37,320 chars and it contains the block. Switching a second, unrelated production agent from `gpt-5.6-terra` to `gpt-6-astra` reproduced the loss; the model reported USER, IDENTITY, SOUL, and the skills list as absent with zero tool calls, while tools stayed available.

#140422 fixes the same mechanism for the skill catalog and explicitly leaves persona and memory for a follow-up. This is that follow-up.

Prior art on the same carrier: #96773 traced the turn-scoped SOUL/IDENTITY/USER channel as the cause of an unstable message prefix (DeepSeek prompt-cache hit rate alternating between ~20% and ~99%), and #96795 proposed the same move to thread-level developer instructions before being withdrawn by its author; that issue was later auto-closed as stale, not fixed. #125783 already moved the agent-workspace `AGENTS.md` to the thread carrier for the same reason. This PR completes that move for the remaining workspace files, so it also gives those files a stable position in the prompt prefix.

## How to reproduce

1. Take any Codex-backed OpenClaw agent whose workspace has `SOUL.md`, `IDENTITY.md`, and `USER.md`. Set its model to a Codex catalog model that ships its own collaboration-mode text. `openai/gpt-6-astra` does; `openai/gpt-5.6-sol` and `gpt-5.6-terra` do not. You can check the agent's Codex home:

   ```bash
   jq '.. | objects | select(.slug? == "gpt-6-astra") | .model_messages.collaboration_modes' <codex-home>/models_cache.json
   ```

   Non-null means the model will replace caller collaboration instructions.
2. Start a fresh session (`/new`) and send:

   ```text
   Without opening files or using tools, quote one short line each from the SOUL, IDENTITY and USER content already in your instructions.
   ```

3. On `main` the model reports the files as missing (or paraphrases the chat history), while `openclaw system-prompt` and the OpenClaw prompt report both list the files as loaded. The recorded rollout confirms where they vanish:

   ```bash
   F=$(ls -t <codex-home>/sessions/$(date +%Y/%m/%d)/rollout-*.jsonl | head -1)
   jq -c 'if .type=="turn_context" then {model:.payload.model, collab_has_soul:(.payload.collaboration_mode.settings.developer_instructions//""|contains("<AGENT_SOUL"))}
          elif (.type=="response_item" and .payload.role=="developer") then {dev_chars:(.payload.content|map(.text//"")|join("")|length), dev_has_soul:(.payload.content|map(.text//"")|join("")|contains("<AGENT_SOUL"))}
          else empty end' "$F"
   ```

   Expected on `main` with `gpt-6-astra`: `collab_has_soul: true` on every `turn_context`, `dev_has_soul: false` on every developer item.
4. Switch the same agent to `gpt-5.6-sol` (no gateway restart needed) and repeat step 2: the quotes come back and the first developer item now has `dev_has_soul: true`.
5. With this branch, repeat step 2 on `gpt-6-astra`: the quotes come back with zero tool calls, the turn's `collaboration_mode.settings.developer_instructions` no longer carries any OpenClaw workspace text, and the first developer item contains `<AGENT_SOUL>` and `<available_skills>`.

Automated form: the added cases in `run-attempt.skills.test.ts`, `thread-lifecycle.binding.test.ts`, and `run-attempt.skills.native.test.ts` fail against the pre-change plugin code and pass with it.

## Why This Change Was Made

The persona and memory pointer now ride the thread developer carrier (`thread/start` and `thread/resume` `developerInstructions`) after the generic policy, exactly like the skill catalog. The existing refreshable section from #140422 is generalised from `skillsInstructions` to `refreshableInstructions` = skills catalog + persona block + memory pointer, in that order, so there is one refresh path and one post-compaction restore path instead of two. The collaboration-mode slot no longer carries any OpenClaw workspace text; only the cron collaboration text remains there.

Behavior change, intentional: native Codex sub-agents now inherit the persona and memory pointer through the thread carrier, the same way they already inherit `AGENTS.md` and the skill catalog. The previous "turn-only, never inherited" property was a side effect of using a field the model can replace, not a designed isolation boundary. Docs say so explicitly.

The file loader, privacy filters, size budget, and `<AGENT_SOUL>` rendering in `attempt-context.ts` are unchanged; only the carrier moved.

## Maintainer decision (carried over from #140422)

ClawSweeper blocked #140422 on a product decision: whether native Codex children may inherit the thread-carried catalog. This PR makes the same choice for persona and memory and documents it as intentional. The previous "turn-only, never inherited" behavior was a side effect of using a field the model can replace, not a designed isolation boundary; the same thread carrier already hands children `AGENTS.md`.

## User Impact

- Agents on catalog-backed models (`gpt-6-astra` today) get their SOUL, IDENTITY, USER, and memory guidance back without switching models.
- Editing SOUL/IDENTITY/USER reaches a live conversation on the next turn, including incognito threads, without `/new` and without a policy-change refusal.
- After a native compaction the current persona is re-delivered, matching the skill catalog behavior.
- Native Codex sub-agents now see the parent agent's persona and memory pointer.

## Evidence

- New regressions (fail on the pre-change code, pass after):
  - `run-attempt.skills.test.ts`, `thread-lifecycle.binding.test.ts`: thread start and resume `developerInstructions` contain `<AGENT_SOUL>` and the memory pointer after the generic policy and skills; collaboration settings no longer carry them.
  - `run-attempt.skills.native.test.ts` (real Codex binary against a recorded Responses wire): a model whose catalog owns collaboration instructions still receives the persona; an edited SOUL reaches a live incognito thread; after compaction the current persona is delivered again.
- Existing suites in `extensions/codex/src/app-server` for run-attempt, thread lifecycle, compaction, turn params, attempt context, and client runtime pass.
- `pnpm tsgo:extensions`, `pnpm tsgo:extensions:test`, `pnpm tsgo:test:root`, `oxfmt --check`, and `oxlint` on the changed files pass.
- Live gateway verification: with this branch deployed and the affected agent switched back to `gpt-6-astra`, a fresh Telegram turn's first developer message was 38,889 chars and contained SOUL, IDENTITY, USER, and the skills list; the model quoted all three files with zero file reads (its only tool call was the message send).
- Prompt snapshot fixtures regenerated with `pnpm prompt:snapshots:gen`; the Codex happy-path fixture now shows `developer_instructions: null` in the turn collaboration settings and the persona in the thread developer instructions.
- Docs updated: `docs/concepts/system-prompt.md`, `docs/plugins/codex-harness-reference.md`, `docs/plugins/codex-harness-runtime.md`, `docs/reference/token-use.md`, plus the in-code bootstrap notice.


VACInc:
#153966 has merged as d61156bb5cfa22e15d19121d6ab4c073bfeac33e and intersects this Codex carrier work. It forwards authenticated `bootstrapUserProfileId` into workspace-context preparation and deliberately keeps the selected personal USER.md overlay turn-scoped, outside the reusable thread snapshot. It does not fix the catalog-owned collaboration-message loss addressed by this PR. When rebasing the thread-carrier change, preserve that boundary: any shared USER section moved to thread-level instructions must exclude `users/<canonical-profile-id>/USER.md`, and native-child inheritance must not gain that personal overlay. This PR is not superseded by #153966.

PR #141293, opened 2026-09-07
fix(cli): preserve inherited primary in model pickers

obviyus:
Related: #136257.

## What Problem This Solves

Configuring an agent with only model fallbacks showed the built-in default instead of the inherited global primary. Manual entry and scoped picker selection also started with the wrong model.

## Why This Change Was Made

The temporary picker model uses the existing effective-primary resolver while retaining the agent's fallback list and model metadata.

## User Impact

Keep and manual entry show the effective primary. Keep makes no edit, so later global-primary changes still apply. Explicit agent primaries and scoped selections keep their existing precedence.

Consumers: configure's Keep label, manual-entry seed, and scoped picker selection now use the effective primary.

## Evidence

The registered configure command reproduced the incorrect default in a real terminal. The corrected command preserved fallback-only configuration after Keep, followed a later global-primary change, and saved an explicit manual primary without changing the global value or fallbacks. Saved-file reads and public configuration reads agreed. All seven targeted picker checks, formatting, whitespace checks, and required continuous integration passed. The proof used the source CLI with synthetic settings; no rebuilt distribution or live provider was exercised.

### DIFF
--- a/src/commands/model-picker.default-agent.test.ts
+++ b/src/commands/model-picker.default-agent.test.ts
@@ -1,5 +1,6 @@
-// Model picker tests read the configured target agent without rewriting global defaults.
+// Model picker tests read the target agent without rewriting global defaults.
 import { describe, expect, it, vi } from "vitest";
+import type { AgentModelConfig } from "../config/types.agents-shared.js";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import type { WizardPrompter } from "../wizard/prompts.js";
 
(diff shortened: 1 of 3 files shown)


obviyus:
**Proof:** The real `configure --section model` command and saved-config readback pass for a fallback-only agent. Seven focused regression/control checks and required CI also pass.

- **Bug:** Keep and manual entry showed the built-in default instead of the agent's inherited primary.
- **Fix:** The picker uses the existing effective-primary resolver and preserves the agent's fallback list.

| Before | After |
| --- | --- |
| Keep and manual entry show the wrong default. | Both show the inherited primary. |
| Keep already preserves the saved fallback-only model. | It still preserves inheritance, including a later global-primary change. |
| Manual replacement uses the existing scoped writer. | Saved readback confirms the selected agent changes while its fallback and the global primary remain intact. |

Command evidence uses the registered source CLI with isolated settings and real terminal input; no live-provider or rebuilt-distribution result is claimed. The fresh local ClawSweeper judgment covers the complete command evidence and exact merged PR head.

PR #141296, opened 2026-09-07
fix(config): apply environment changes after in-process restart

obviyus:
## What Problem This Solves

Fixes an issue where changing or removing configured environment values after an in-process Gateway restart left the process using old values.

## Why This Change Was Made

Keep environment ownership through snapshot cleanup so the next startup does not mistake surviving configured values for inherited settings. Fence old rollback transactions while retaining explicit full-reset behavior.

## User Impact

Accepted configuration changes replace or remove owned environment values after restart. Inherited values keep precedence, and configuration-selected locations stay stable.

## Evidence

A real Gateway reproduced the stale value after a same-process restart. The candidate completed restart, replacement and removal: public reads showed the new value, then the unresolved placeholder, while the inherited control stayed unchanged. Fifty-eight focused tests covered rollback, precedence, aliases, casing, blocked variables and reset. Required checks passed; no provider call was used.

### DIFF
--- a/src/config/config-env-vars.ts
+++ b/src/config/config-env-vars.ts
@@ -362,8 +362,15 @@ export function initializePublishedConfigRuntimeEnv(
   pendingConfigRuntimeEnvPublication = null;
 }
 
-export function resetPublishedConfigRuntimeEnv(): void {
-  publishedConfigRuntimeEnvState = { generation: 0, ownedEnv: {}, sourceConfig: null };
+export function resetPublishedConfigRuntimeEnv(
+  options: { preserveOwnership?: boolean } = {},
+): void {
+  publishedConfigRuntimeEnvState = options.preserveOwnership
+    ? {
+        ...publishedConfigRuntimeEnvState,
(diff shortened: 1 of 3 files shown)


obviyus:
Landed in c2affa16cb1ca96e3e460c18997dc4b1d855155a.

- **Bug:** After an in-process restart, the Gateway accepted a configured environment replacement but kept reading its old value.
- **Fix:** Snapshot cleanup retains ownership of the environment values that remain in the process, while fencing old rollback transactions.
- **Proof:** The real Gateway now applies replacement and removal after restart, with inherited values unchanged. Both writes and final shutdown exit successfully. The 58 focused tests and [required CI](https://github.com/openclaw/openclaw/actions/runs/34136580481) pass; fresh local ClawSweeper review is clean at tier B.

| Before | After |
| --- | --- |
| Accepted replacement still reads `one` after restart. | Replacement reads `two`; removal leaves the normal unresolved placeholder. The inherited control is unchanged. |

PR #141297, opened 2026-09-07
fix(gateway): retain owned run loop through failed restarts

obviyus:
## What Problem This Solves

After a foreground Gateway closed its listener for restart and startup failed, it promised to remain alive but could exit with code 13 before the operator could retry.

## Why This Change Was Made

The CLI run loop keeps a referenced message port for the signal lifetime it already owns. Existing cleanup closes both ports. Restart timing, retry policy, and initial-start rejection stay with their current owners.

## User Impact

The foreground Gateway remains available for another restart attempt after refused startup and exits normally when stopped. No configuration or migration is required.

## Evidence

The real source CLI reproduced exit 13 after restart with missing encryption certificate files. With the fix, the same process survived two refused restarts, recovered after correcting configuration, returned authenticated health, and exited 0 on termination. Four subprocess tests cover repeated recovery, both termination signals, and initial-start rejection. Formatting, lint, whitespace checks, and required continuous integration passed. Packaged installation and Windows signals were not exercised.

### DIFF
--- a/src/cli/cli-entrypoint.test-support.ts
+++ b/src/cli/cli-entrypoint.test-support.ts
@@ -44,6 +44,11 @@ export const gatewayDirectStopEntrypoints = {
     sourceWorkerName: "gateway-cli/run-loop",
     distWorkerPath: "cli/gateway-cli/run-loop.js",
   },
+  restartPolicy: {
+    currentModuleUrl: import.meta.url,
+    sourceWorkerName: "../infra/restart",
+    distWorkerPath: "infra/restart.js",
+  },
   workAdmission: {
     currentModuleUrl: import.meta.url,
     sourceWorkerName: "../process/gateway-work-admission",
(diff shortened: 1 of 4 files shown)


obviyus:
Landed with a clean exact-head local review (tier B, zero findings) and [green required CI](https://github.com/openclaw/openclaw/actions/runs/34138003289).

- **Bug:** the foreground Gateway could exit13 after announcing a failed-restart recovery wait.
- **Fix:** the run-loop process owner retains and releases the lifetime reference with its signal handlers.
- **Proof:** the real source-mode CLI baseline exited13 after a TLS refusal. The candidate survived two refusals, recovered in the same process, returned authenticated health with `ok:true`, and exited0 on SIGTERM.
- **Controls:** focused subprocess coverage passed for repeated recovery, SIGTERM, SIGINT and initial-start rejection, including the corrected compiled CI path.

Packaged installation and Windows signal behavior were not exercised. Earlier setup failures and the first insufficient-proof review remain preserved in the private verification record.

PR #141301, opened 2026-09-07
fix(line): channel access token and secret cannot be secret references

### LINKED ISSUE #133625: [Bug]: LINE credentials cannot be SecretRefs, so a LINE token cannot live in a secret provider
### Bug type

### Beta release blocker

No

### Summary

LINE credentials cannot be SecretRefs: `channels.line.channelAccessToken` rejects the same store reference `channels.telegram.botToken` accepts, so a LINE token cannot live in the secrets store, a file provider, or 1Password the way every other configured channel's can.

### Steps to reproduce

Two `config patch` calls against the same config, same ref shape:

```
$ echo '{"channels":{"line":{"channelAccessToken":{"source":"store","provider":"default","id":"LINE_TOKEN"}}}}' \
    | openclaw config patch --stdin
Config validation failed: channels.line.channelAccessToken: invalid config: must be string

$ echo '{"channels":{"telegram":{"botToken":{"source":"store","provider":"default","id":"TG_TOKEN"}}}}' \
    | openclaw config patch --stdin
Applied 1 config update(s). Restart the gateway to apply.
```

### Expected behavior

The LINE token and channel secret accept a SecretRef, resolve through the configured provider at startup, and participate in the channel-account secret lifecycle the same way other channels' credentials do. `docs/gateway/secrets.md:25` lists channel accounts among the owner classes that isolate a retryable SecretRef failure into a configured-unavailable account instead of failing Gateway startup, and `:27` describes reload reconciling channel connections when a secret-provider edit changes resolved credentials. Neither applies to LINE today.

### Actual behavior

The config is rejected at validation. There is no spelling of a LINE credential that reaches a secret provider — the operator has to inline the token or point at a file with `tokenFile` / `secretFile`.

### OpenClaw version

2026.8.1 (built from `main`)

### Where it comes from

Two halves are both absent.

The schema declares plain strings inside a `.strict()` object, so a SecretRef object cannot validate:

```
extensions/line/src/config-schema.ts:28-29
  channelAccessToken: z.string().optional(),
  channelSecret: z.string().optional(),
```

Compare `extensions/telegram/src/config-schema.ts:181`:

```
  botToken: registerSensitiveConfigSchema(SecretInputSchema.optional()),
```

And the secret contract is an explicit empty stub, so nothing registers LINE's fields as secret targets or resolves them:

```
extensions/line/secret-contract-api.ts
  // Line does not expose secret-contract surfaces.
  export const secretTargetRegistryEntries: readonly [] = [];
  export function collectRuntimeConfigAssignments(): void {}
```

`collectRuntimeConfigAssignments` is what writes a resolved secret back over the config field (`src/secrets/channel-secret-basic-runtime.ts`, the `apply` callback), and `secretTargetRegistryEntries` is what puts a channel in the `openclaw secrets` target list (`src/cli/command-secret-targets.ts:650`). With both empty, LINE has neither.

The host fallback does not cover it either: `loadOfficialExternalChannelSecretContractApi` builds a contract from a plugin manifest's `openclaw.channelSecrets.fields` (`src/plugins/official-external-plugin-catalog.ts:1523`), and `extensions/line/package.json` declares no `channelSecrets` block. The `"sensitive": true` flags it does declare are setup-prompt masking, not secret resolution.

### What a fix has to decide

Filing rather than patching, because the shape is a maintainer call rather than a mechanical port. The 14 channels that have `src/secret-contract.ts` do not agree on the parts LINE would have to choose between:

- Inheritance. LINE's resolver only lets the default account inherit top-level credentials (`extensions/line/src/accounts.ts:46`: `accountId === DEFAULT_ACCOUNT_ID ? [accountConfig, baseConfig] : [accountConfig]`), so the collector's `topLevelInheritedAccountActive` predicate is LINE-specific rather than one of the stock modes.
- Precedence against `tokenFile` / `secretFile`, which LINE already resolves with its own `configured_unavailable` status and diagnostic. Which one wins, and whether an unavailable ref degrades the account the way an unreadable file does, is a product decision.
- How much of the read-side handling to mirror: `extensions/clickclack/src/accounts.ts:146-190` carries roughly sixty lines of ref handling, including env-provider allowlist errors and a read-only-path branch. Some of that may not apply to LINE.

### Scope check

LINE is not alone: `whatsapp`, `signal` and `imessage` also have no secret contract. This issue is scoped to LINE because that is the channel whose credential resolution I verified end to end; the same two halves would need checking separately for each of the others, and their credential shapes differ.

edenfunf:
Closes #133625

## What Problem This Solves

`channels.line.channelAccessToken` and `channels.line.channelSecret` reject the structured secret reference that `channels.telegram.botToken` and the credentials of eleven other channels accept, so a LINE deployment has to keep its bot credentials as plaintext in `openclaw.json`. Same reference shape, same config, two different answers.

Three things had to be true at once for that to work, and LINE had none of them:

- `extensions/line/src/config-schema.ts:28` declared both credentials as `z.string()`, so the schema rejected a reference before anything else ran.
- `extensions/line/secret-contract-api.ts` was an explicit stub — `// Line does not expose secret-contract surfaces.` — so even a reference that validated would never be collected for resolution. Four other channels ship the same stub (iMessage, Signal, WhatsApp, Zalo User), and for them it is accurate: none of the four declares a credential field in its config schema, because their auth lives in a linked-device store. LINE declares four (`channelAccessToken`, `channelSecret`, `tokenFile`, `secretFile`) and shipped the stub anyway.
- `resolveLineCredential` read the raw config value with `?.trim()`, so a reference object would have thrown rather than degraded.

The operator-visible result is that a LINE deployment cannot keep its bot credentials in the shared secret store, a file provider, or an external manager, and cannot bind them to an egress host — while a Telegram or Slack deployment beside it can.

## Why This Change Was Made

The fix joins LINE to the contract that already exists rather than adding a LINE-specific credential path.

**Schema.** Both credentials become `registerSensitiveConfigSchema(SecretInputSchema.optional())`, the same declaration Telegram, Discord, Slack, Matrix and eight others use. Registering the schema is what makes host config projections redact the value, so this also stops LINE credentials from being echoed back in config reads.

**Contract.** `extensions/line/src/secret-contract.ts` publishes both credentials at the channel and account scopes, which is four target paths, and collects runtime assignments, so core resolves the reference and writes the literal back before the plugin reads it. The activity predicates encode LINE's own precedence rather than copying Telegram's:

- LINE reads the inline credential *before* its credential file, so an enabled account always consumes its own inline value. Telegram reads `tokenFile` first and therefore deactivates `botToken` when a file is configured; that condition would be wrong here.
- Only the default account falls back to channel-level credentials, and only while it supplies neither the inline value nor its own file.
- `resolveChannelAccountSurface` omits the default account when the accounts map names only other accounts, but LINE still resolves a default account from channel-level credentials in exactly that config. The contract adds that consumer back, otherwise a deployment with a channel-level reference plus one named account would leave its primary account permanently unresolved.

**Reader.** `resolveLineCredential` now reads through `resolveSecretInputString(..., mode: "inspect")`. A reference the runtime has not resolved returns `configured_unavailable` and stops there instead of falling through to the credential file or the environment: falling through would authenticate with a different credential than the operator named. `configured_unavailable` is the status `hasConfiguredUnavailableCredentialStatus` already feeds to channel status, gateway health, outbound target selection and the security audit, so the degraded state surfaces through the paths that exist rather than a new one.

**Setup wizard.** Two of its declarations were written while the schema was narrow, and neither survives a credential that can be a reference:

- `set: { value: "resolved" }` wrote the value the wizard had resolved a reference *to*, so choosing "Use external secret provider" stored the plaintext. Writing the reference itself was not possible before, which is why the declaration was there; it is now `set: { clearFields: [...] }`, matching Slack. Plaintext entry is unchanged, because that prompt returns the same string in both halves.
- `accountConfigured` asked whether the account as a whole was configured, which needs both credentials to resolve. Once a credential can be a reference, one reference the provider has not answered marks the other credential as missing too, and the wizard asks the operator to retype a value that works. It now answers per credential from what that credential resolved to: a resolved value, or a reference the operator named on purpose, can be kept; a credential file that cannot be read cannot. Whether the config names a value at all stays with `defineTokenCredential`, so a named file still suppresses the environment-variable offer and the operator's file is never dropped without a word.

One thing this deliberately does not do.

The default-account match uses `normalizeOptionalAccountId`, which leaves an id the router cannot read as `undefined`, because that is the answer the reader's own lookup gives: `resolveAccountEntry` does not match `"!!!"` for `"default"` either. The shared owner builder still normalizes non-optionally, so an `accounts` map naming only an unreadable id produces two config paths under one owner id (`line:default`), and a failure on the unreachable one isolates the reachable default account with it. That collapse is in `createChannelAccountSecretOwner`, not here, and matching it in LINE would put the wrong answer back.

One behavior note: a credential written as `$NAME` or `${NAME}` is now collected as an environment reference, which is how every other channel treats that spelling. The two spellings are matched by `^\$\{([A-Z][A-Z0-9_]{0,127})\}$` and `^\$([A-Z][A-Z0-9_]{0,127})$`; a LINE channel access token is a ~170-character mixed-case string, so no real credential can collide with it. The docs describe the object form, which is the spelling the reader and the collector agree on.

This PR touches `src/secrets/`, which CODEOWNERS assigns to `@openclaw/openclaw-secops`, with two files and no production code: `runtime-line-credential-inheritance.test.ts` and `runtime-line.test-support.ts`. The second follows the shape already there for Discord, Matrix, Telegram, Zalo and Nextcloud Talk, and the first uses it to run the real `prepareSecretsRuntimeSnapshot` against LINE's contract artifact rather than a hand-written double. Everything that runs in production is under `extensions/line/`.

No dependency, lockfile, or CHANGELOG changes.

## User Impact

A LINE operator can now keep both credentials outside `openclaw.json`:

```jsonc
{
  "secrets": { "providers": { "default": { "source": "store" } } },
  "channels": {
    "line": {
      "channelAccessToken": { "source": "store", "provider": "default", "id": "LINE_TOKEN" },
      "channelSecret": { "source": "store", "provider": "default", "id": "LINE_SECRET" }
    }
  }
}
```

Existing plaintext, `tokenFile`/`secretFile`, and `LINE_CHANNEL_ACCESS_TOKEN`/`LINE_CHANNEL_SECRET` configurations keep working unchanged, and their precedence is unchanged.

A configured reference owns its credential: when the runtime cannot resolve it at startup the account stops rather than quietly authenticating with `tokenFile`/`secretFile` or the environment variable the operator did not name. A reference that stops resolving later is a different case, and the Evidence below walks all three of them. `openclaw channels status` shows the account stopped with the reason and `openclaw doctor` points at that account's references rather than only the one that failed, with the retry command, instead of the account looking unconfigured. Rotating a stored value needs no config edit: `openclaw secrets reload` re-resolves it and restarts the LINE channel.

Guided setup answers per credential as a consequence, which reference support requires and which an operator on a plaintext config sees today:

- An account that holds one credential and not the other is no longer asked to retype the one it holds. A config with a channel access token and no signing secret used to prompt for both.
- A credential the runtime cannot resolve right now is judged by what it is. A reference the provider has not answered is still the operator's choice and can be kept; a credential file that cannot be read leaves nothing to keep, and the wizard asks for a credential, as it does today on `main`.

## Evidence

Run on a desktop LINE test channel against a real gateway built from this branch, with both credentials moved into the shared secret store and the config carrying only references. Three sittings are quoted below and every block says which one it belongs to. The blocks down to the audit pair come from the startup sitting, where each half restarts the gateway because the config changes between them; its platform verifications and tunnel rows carry matching timestamps. The rotation block comes from a separate sitting on one long-lived gateway, run earlier the same evening on the same build. The wizard and check blocks are neither: they are test runs, labelled with the commands that produced them.

```
$ openclaw --version
OpenClaw 2026.9.2 (5ccb173)
```

**Store, not config.** The credentials live in the store; the config names them:

```
$ openclaw secrets store list
LINE_STORE_SECRET [secret] (write-only); allowed hosts: none
LINE_STORE_TOKEN [secret] (write-only); allowed hosts: none

$ jq '.channels.line.accounts.default | {channelAccessToken, channelSecret}' openclaw.json
{
  "channelAccessToken": { "source": "store", "provider": "default", "id": "LINE_STORE_TOKEN" },
  "channelSecret": { "source": "store", "provider": "default", "id": "LINE_STORE_SECRET" }
}
```

**The command from the issue now succeeds**, and what lands in the config is the reference rather than a value resolved from it:

```
$ echo '{"channels":{"line":{"channelAccessToken":{"source":"store","provider":"default","id":"LINE_STORE_TOKEN"}}}}' | openclaw config patch --stdin
Config warnings: plugins.entries.codex: plugin disabled (bundled (disabled by default)) but config is present
Applied 1 config update(s). Restart the gateway to apply.

$ jq '.channels.line' openclaw.json
{
  "channelAccessToken": { "source": "store", "provider": "default", "id": "LINE_STORE_TOKEN" }
}
```

On `origin/main` that input cannot validate at all: `extensions/line/src/config-schema.ts` declares `channelAccessToken: z.string().optional()`, so an object is rejected before anything reads it.

**The account starts on the resolved credentials.** The startup probe reads the bot's display name back from the Messaging API using the resolved token:

```
2026-09-07T21:44:44.584+08:00 [line] [default] starting LINE provider (openclaw)
2026-09-07T21:44:44.592+08:00 line: registered webhook handler at /line/webhook

$ openclaw channels status
- LINE default: enabled, configured, running, connected, mode:webhook, token:***, signing:config
```

**LINE's own platform signs a request that the resolved signing secret validates.** This is the channel's webhook verification, so LINE computes the signature and reports the status it observed at our endpoint:

```
$ curl -X POST https://api.line.me/v2/bot/channel/webhook/test -H 'Authorization: Bearer ...'
{"success":true,"timestamp":"2026-09-07T13:44:53.826530Z","statusCode":200,"reason":"OK","detail":"200"}
```

**Negative half: the same config with the signing-secret reference pointed at a store key that does not exist, on a gateway restarted to pick it up.** Nothing else changed.

```
[secrets] [SECRETS_OWNER_UNAVAILABLE] Secret owner account:line:default is configured-unavailable;
          paths: channels.line.accounts.default.channelAccessToken,
                 channels.line.accounts.default.channelSecret
[secrets] [SECRETS_DEGRADED] cold account:line:default: secret reference was not found.
          Retry: openclaw secrets reload.
[line] channel startup failed: Secret owner account:line:default is configured but unavailable
       (secret reference was not found).

$ openclaw channels status
- LINE default: enabled, configured, reason:Secret owner account:line:default is configured but unavailable (secret reference was not found)., error:Secret owner account:line:default is configured but unavailable (secret reference was not found)., stopped
Warnings:
- line default: Channel runtime is blocked and needs operator action. (resolve the reported channel error, then restart the channel)

$ curl -X POST https://api.line.me/v2/bot/channel/webhook/test -H 'Authorization: Bearer ...'
{"success":false,"timestamp":"2026-09-07T13:45:34.726475Z","statusCode":403,"reason":"ERROR_STATUS_CODE","detail":"403"}
```

One unresolved reference degrades the account that owns it, so both of its config paths are named. `openclaw doctor` says the same against that gateway:

```
Secret runtime degradation
- cold account:line:default
  (channels.line.accounts.default.channelAccessToken,
   channels.line.accounts.default.channelSecret): secret reference was
   not found
  Retry: openclaw secrets reload
```

**Restoring the reference restores the channel**, with nothing else changed back:

```
$ openclaw channels status
- LINE default: enabled, configured, running, connected, mode:webhook, token:***, signing:config

$ curl -X POST https://api.line.me/v2/bot/channel/webhook/test -H 'Authorization: Bearer ...'
{"success":true,"timestamp":"2026-09-07T13:51:55.285903Z","statusCode":200,"reason":"OK","detail":"200"}
```

The tunnel in front of the gateway recorded all three of those deliveries, which is the whole A/B at the wire: same platform, same endpoint, same signature shape, and the only thing that differed between them was whether the signing secret could be resolved.

```
$ curl -s 'http://127.0.0.1:4040/api/requests/http'          # newest first
21:51:55  POST /line/webhook -> 200 OK         reference restored
21:45:34  POST /line/webhook -> 403 Forbidden  reference pointed at a missing store key
21:44:53  POST /line/webhook -> 200 OK         reference resolved

all three: User-Agent LineBotWebhook/2.0, X-Line-Signature 44 chars
```

**Outbound uses the resolved access token.** Sent through OpenClaw rather than the raw API, and delivered to a real LINE account:

```
$ openclaw message send --channel line --target U4a… -m "…" --json
"deliveryStatus": "sent",
"payloadOutcomes": [ { "index": 0, "status": "sent", "resultCount": 1 } ]
```

**The audit can see LINE credentials for the first time.** Same build, two configs that differ only in whether those two fields hold references or literals:

```
# config with SecretRefs
Secrets audit: findings. plaintext=2, unresolved=0, shadowed=0, storeResidue=0, legacy=0.
- [PLAINTEXT_FOUND] gateway.auth.token is stored as plaintext.
- [PLAINTEXT_FOUND] models.providers.openrouter.apiKey is stored as plaintext.

# same build, credentials written inline
Secrets audit: findings. plaintext=4, unresolved=0, shadowed=0, storeResidue=0, legacy=0.
- [PLAINTEXT_FOUND] gateway.auth.token is stored as plaintext.
- [PLAINTEXT_FOUND] models.providers.openrouter.apiKey is stored as plaintext.
- [PLAINTEXT_FOUND] channels.line.accounts.default.channelAccessToken is stored as plaintext.
- [PLAINTEXT_FOUND] channels.line.accounts.default.channelSecret is stored as plaintext.
```

On `origin/main` the last two lines cannot appear at all: `extensions/line/secret-contract-api.ts` publishes `readonly []`, so the registry holds no LINE paths to audit.

**Rotating the stored value needs no config edit, and the new value is the one that gets used.** The store entry was changed twice on a running gateway; the config file was not touched:

```
$ openclaw secrets store set LINE_STORE_SECRET --value-file <a secret LINE will not accept>
Stored LINE_STORE_SECRET (secret).
A gateway is running. Run `openclaw secrets reload` for config-referenced values to take effect.

$ openclaw secrets reload
Secrets reloaded.

$ openclaw channels status
- LINE default: enabled, configured, running, connected, mode:webhook, token:***, signing:config

$ curl -X POST https://api.line.me/v2/bot/channel/webhook/test -H 'Authorization: Bearer ...'
{"success":false,"timestamp":"2026-09-07T13:34:20.193275Z","statusCode":401,"reason":"ERROR_STATUS_CODE","detail":"401"}

$ openclaw secrets store set LINE_STORE_SECRET --value-file <the real one>
$ openclaw secrets reload
$ curl -X POST https://api.line.me/v2/bot/channel/webhook/test -H 'Authorization: Bearer ...'
{"success":true,"timestamp":"2026-09-07T13:34:30.880093Z","statusCode":200,"reason":"OK","detail":"200"}

$ jq '.channels.line.accounts.default.channelSecret' openclaw.json   # unchanged throughout
{ "source": "store", "provider": "default", "id": "LINE_STORE_SECRET" }
```

The account stays up across the rotation because the reference still resolves; what changes is the value it resolves to, which is why LINE's signature check reports `401` rather than the `403` an unresolvable reference produces.

Removing the store entry outright, on the same running gateway, takes the third path the runtime has for a reference. This reload does not restart the channel at all, unlike the two rotations above it: the account keeps serving on the last value it resolved and the runtime says so, rather than dropping a live channel because a provider went away.

```
$ openclaw secrets store rm LINE_STORE_SECRET
Removed 1 team store entry.

$ openclaw secrets reload
Secrets reloaded with 1 warning(s).

[secrets] [SECRETS_OWNER_UNAVAILABLE] Secret owner account:line:default is using last-known-good;
          paths: channels.line.accounts.default.channelAccessToken,
                 channels.line.accounts.default.channelSecret
[secrets] [SECRETS_DEGRADED] stale account:line:default: secret reference was not found.
          Retry: openclaw secrets reload.

# and when the value comes back
[secrets] [SECRETS_RELOADER_RECOVERED] Secret resolution recovered; runtime remained on
          last-known-good during the outage.
```

So a LINE account now follows the same three states every other referenced credential has: `cold` and stopped when the reference cannot resolve at startup, `stale` and serving the last known good value when it stops resolving later, and recovered when it comes back. None of that is new machinery; LINE simply had no way to reach it.

**Guided setup.** The eight cases this PR adds are what hold the wizard's answer in place; seven of them fail if the answer is put back the way it was. Replacing the override with the whole-account question `main` asks reddens four of them; replacing it with the first attempt at the credential-file fix reddens the two that watch an unreadable file; removing it entirely reddens four:

```
$ vitest run --config test/vitest/vitest.extension-line.config.ts extensions/line/src/setup-surface.test.ts
      Tests  18 passed (18)

# with accountConfigured replaced by main's Boolean(token && secret)
 × answers each credential from what that credential resolved to
 × keeps a plaintext credential while the other one is a reference it cannot resolve
 × does not ask for the credential the account already has while the other is missing
 × treats a configured reference as configured while it stays unresolved
      Tests  4 failed | 14 passed (18)

# with the override moved back onto hasConfiguredValue
 × treats a credential file it cannot read as nothing to keep
 × asks for the credential behind an unreadable file even when the env var is set
      Tests  2 failed | 16 passed (18)

# with the override removed
 × answers each credential from what that credential resolved to
 × treats a credential file it cannot read as nothing to keep
 × asks for the credential behind an unreadable file even when the env var is set
 × asks for a token instead of offering to keep one behind an unreadable file
      Tests  4 failed | 14 passed (18)
```

**Checks.** All of the following ran on Windows at `4a7d6c0b92c`, on a branch rebased onto `233dabe750a`. One commit follows it, and it rewrites a single sentence in `docs/channels/line.md`; the format check below was re-run after it:

```
$ vitest run --config test/vitest/vitest.extension-line.config.ts extensions/line
 Test Files  2 failed | 53 passed (55)
      Tests  57 failed | 807 passed | 1 skipped (865)
```

Every failure is in `bot-message-context.test.ts` (45) and `bot-mention-context.test.ts` (12), from SQLite `EBUSY` unlinks and 120s timeouts that this branch does not touch. Running those two files alone in a pristine worktree checked out at this branch's base gives the same result, and the failing test names are identical in both directions:

```
$ vitest run --config test/vitest/vitest.extension-line.config.ts \
    extensions/line/src/bot-message-context.test.ts extensions/line/src/bot-mention-context.test.ts
 Test Files  2 failed (2)
      Tests  57 failed | 2 passed (59)

$ comm -3 branch-failures.txt baseline-failures.txt
(no output; 56 distinct names on each side)
```

The rest:

```
vitest --config test/vitest/vitest.secrets.config.ts src/secrets/{target-registry.docs,
        runtime-line-credential-inheritance,runtime.coverage}.test.ts   92 passed

tsgo -p tsconfig.extensions.json --noEmit                     0 errors
tsgo -p test/tsconfig/tsconfig.extensions.test.json --noEmit  0 errors
pnpm lint:extensions                                          pass
oxfmt --check <all 16 changed .ts and .md files>              pass  (at 5ccb173f807)
pnpm check:assertion-safety  --base origin/main               3980 files, no new assertions
pnpm check:max-lines-ratchet --base origin/main               pass
pnpm config:channels:check                                    pass
```

**Not exercised.** A message typed by a person in the LINE client: the Windows LINE app ignores synthetic mouse and keyboard input, so the inbound half here is LINE's own platform verification rather than a user-authored message. That path is unchanged by this PR, and both credentials it depends on are exercised above.


Patrick-Erichsen (commented):
The LINE SecretRef implementation is technically qualified, but landing is pending security-owner involvement for the changes under `src/secrets/` and the credential-surface documentation. Current `CODEOWNERS` assigns those paths to the security team, and current repository policy requires listed-owner involvement; there is no such review on this head yet. The PR also needs integration with current main and refreshed exact-head gates after that review. Leaving it open for that prerequisite; no change request on the implementation.

PR #141312, opened 2026-09-07
fix(skills): honor Content-Disposition filenames for skill installs

hugenshen:
## What Problem This Solves

Extensionless skill dependency downloads fail with `extract requested but archive type could not be detected`, even when the final response names a supported archive. The download owner now uses explicit archive metadata first, a valid response filename second, and the original URL suffix last.

The parser supports quoted filenames, UTF-8 extended filenames with language tags, and quoted semicolons. It rejects malformed parameters, duplicate names, invalid language tags, and control characters so they cannot override a valid URL fallback. Empty valid extended filenames also preserve their precedence and permit URL fallback. Response metadata selects the archive format only; destination names and extraction safety remain with their existing owners.

## Evidence

- Production Skills UI → authenticated `skills.install` → guarded public HTTPS redirect → ZIP extraction. Pinned main `219d68b83530` returned the inference error; the repaired tree returned success and extracted the expected 13-byte README to the configured tools directory.
- The harmless archive contains a README, not an executable. The fixture's missing-bin badge therefore remains; this proof establishes extraction and the visible install result.
- Focused source, Gateway workspace, and real HTTP/archive controls cover all five archive suffixes, explicit/header/URL precedence, malformed and UTF-8 names, non-extracting downloads, checksum/size/link/strip/permission/root boundaries, and redirect/request safety.
- Independent source-blind acceptance repeated the public UI install with its own fixture and verified read-only, unknown-agent, and unknown-installer refusals without filesystem changes.
- Public CLI policy controls retain warning acknowledgement and fail closed on blocks or invalid policy output. Public Gateway controls retain admin and configured-workspace requirements.
- Original contributor HTTP proof and prior URL-fallback/quoted-semicolon corrections are preserved in the PR history. Additional malformed-header and language-tag failures were reproduced before repair.

Candidate UI and Gateway were built together from tested tree `e698cdeb0ff72de2b76564fb6fe082746bbe8759`, before its signed integration commit. Their original build stamp remains unchanged. The final correction affects only an empty extended filename; these public UI responses use ordinary filenames and do not enter that branch. Separate real HTTP regressions prove the final empty-value correction. Private captures preserve full requests, responses, failures, and filesystem evidence; public text omits credentials, URLs, headers, private paths, and unrelated skill data.

Thanks @hugenshen for the original fix.

- [x] AI-assisted
- [x] Change is focused and preserves existing installer boundaries


hugenshen:
@clawsweeper re-review

PR #141315, opened 2026-09-07
fix(voice-call): flush unmatched webhook upgrade rejects before destroy

hugenshen:
## What Problem This Solves

Unmatched WebSocket upgrades to the voice-call webhook server close without an HTTP response. Clients see a reset or empty connection close instead of a readable rejection. Once the server writes a rejection, it must also own socket errors during that write because Node has transferred the upgraded socket out of HTTP error handling.

## Why This Change Was Made

Use the existing SDK rejection helper to flush HTTP 404, with an error guard installed first and only for unmatched upgrades. Recognized realtime and media streams retain their existing handlers and error ownership.

Replace the peer-disconnect-and-sleep test with a deterministic fault at the real server socket. A non-handling error monitor observes the actual error while the rejection write is pending. Removing only the production guard makes the same regression fail with that uncaught error. Test fixtures now validate separate, supported audio modes.

## User Impact

Clients receive a readable HTTP 404 for unknown voice stream paths. Socket errors during rejection do not escape as uncaught process errors. Existing realtime, media, and ordinary webhook behavior stays intact.

## Evidence

**Normal Gateway/plugin startup**, tested separately for realtime and classic streaming. Both configurations pass `openclaw voicecall setup --json`; service logs report the voice-call runtime initialized. Tests use isolated loopback networks, a mock carrier for realtime, and inert synthetic carrier settings for streaming. No real carrier request, credential, or public exposure is involved.

Redacted output from both modes:

```text
Baseline c7cb45dfd5db:
  voicecall setup: ok=true
  unmatched upgrade: response="", closed=true

Candidate runtime 7efebf957152:
  voicecall setup: ok=true
  unmatched upgrade:
    HTTP/1.1 404 Not Found
    Connection: close
  complete headers received before closure

Before and after controls:
  recognized realtime rejection: 401
  recognized media handshake: 101
  malformed media upgrade: 400
  ordinary webhook GET: 405
  unknown HTTP path: 404
```

The final revision changes only tests, so the verified candidate runtime retains its actual build identity. A fresh independent validator passed **24 TCP/HTTP checks** across both Gateway-started modes, including random paths, fragmented requests, repeated requests, fresh WebSocket handshake verification, and both setup commands.

Deterministic error regression:

```text
guard present: pending-write socket-error test passed
guard removed: 1 failed test, 1 uncaught exception
Error: synthetic rejection transport failure
code: ECONNRESET
exit: 1
guard restored: tests passed
```

This injects a fault through Node's real `destroy(error)` path. It does not claim that a peer sent an operating-system reset packet. The monitors observe without handling the error.

- **91 focused tests** passed with valid single-mode fixtures; all four final transport cases passed after the typed socket-chunk correction.
- Formatting and focused lint passed. Fresh source review found no actionable findings.
- The previous direct-server proof and invalid dual-mode fixture are retained as failure history; the Gateway-started runs above address that review finding.
- The first CI run found a typed-lint error in test response collection; the final revision converts the actual string-or-buffer chunk without a cast.
- Source-only guard order and ownership requirements have separate source review. Private paths, identities, request bodies, and credentials are omitted from these excerpts.

Retains @hugenshen's original contribution and commit ancestry. AI-assisted.


hugenshen:
@clawsweeper re-review

hugenshen:
@clawsweeper re-review

obviyus:
Thanks @hugenshen! Landed in 0b4864b3a95629ad2033c83feb66a882328057ff with your contributor credit preserved.

**Proof:** Normal Gateway/plugin startup passed in separate realtime and streaming configurations. The PR body retains redacted before/after transport output and the deterministic socket-error regression. [Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34299013234), with fresh independent behavior and source reviews.

- **Bug:** Unknown voice stream paths closed without an HTTP response.
- **Fix:** The existing helper flushes HTTP 404, with socket-error ownership limited to the rejected upgrade.

| Before | After |
| --- | --- |
| Unmatched upgrades closed with no HTTP bytes. | Both audio modes return a complete HTTP 404 before closing. |
| The peer-disconnect test did not prove the guard ran. | A real-server pending-write fault is observed, and removing the guard causes the expected uncaught error. |

Recognized realtime/media and malformed-upgrade controls retain their behavior. Validation used isolated loopback traffic and no real carrier calls or credentials. All earlier failed proof and CI attempts remain recorded separately.

PR #141326, opened 2026-09-07
fix(tlon): honor Content-Disposition filenames on outbound image upload

hugenshen:
## What Problem This Solves

Tlon outbound image uploads named the stored file from the source URL path only. A CDN URL like `/download?id=123` with `Content-Disposition: attachment; filename="photo.png"` uploaded as `download` (or a generated default) instead of `photo.png`.

## Why This Change Was Made

`uploadImageFromUrl` already receives `fileName` from `readRemoteMediaBuffer`, which parses Content-Disposition. Use that fetched name first, then the URL basename, then a generated default. Byte caps, SSRF policy, and upload fallback are unchanged.

## User Impact

**Before:** Outbound images from extensionless download URLs lost the publisher filename in Tlon storage.

**After:** The Content-Disposition filename is the stored Tlon upload name when the downloader provides one.

## Evidence

- Changed: `extensions/tlon/src/urbit/upload.ts`
- Production path: live loopback HTTP → production `readRemoteMediaBuffer` → `uploadImageFromUrl` → captured `uploadFile({ fileName })`
- Tests: `extensions/tlon/src/urbit/upload.content-disposition.http.test.ts`, `extensions/tlon/src/urbit/upload.test.ts`

## Real behavior proof

- **Behavior or issue addressed:** Outbound Tlon uploads ignored Content-Disposition filenames from the shared media downloader.
- **Canonical reachability path:** Production `uploadImageFromUrl` against a real `node:http` server that sends `Content-Disposition: attachment; filename="photo.png"` on `/download?id=123`.
- **Boundary crossed:** Bytes on the wire through production `readRemoteMediaBuffer`. Tlon `uploadFile` is captured at the storage boundary (no live Tlon account). Loopback is allowlisted only so the local reflector is reachable; public CDN fetches use the same downloader.
- **Shared helper / provider constraint check:** Filename selection stays in the Tlon upload owner; media fetch parsing is unchanged.
- **Real environment tested:** Local Vitest extension-messaging lane with loopback HTTP and a 1×1 PNG body.

```text
$ git diff --check
$ node scripts/run-vitest.mjs extensions/tlon/src/urbit/upload.content-disposition.http.test.ts --reporter=verbose
$ node scripts/run-vitest.mjs extensions/tlon/src/urbit/upload.test.ts --reporter=verbose
```

- **Evidence after fix:**

```text
stdout | extensions/tlon/src/urbit/upload.content-disposition.http.test.ts > uploadImageFromUrl Content-Disposition HTTP > stores the Content-Disposition filename from a real HTTP download
[tlon content-disposition proof] stored_filename=photo.png source=http://127.0.0.1:52996/download?id=123

 ✓ |extension-messaging| extensions/tlon/src/urbit/upload.content-disposition.http.test.ts > uploadImageFromUrl Content-Disposition HTTP > stores the Content-Disposition filename from a real HTTP download 91ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
[test] passed 1 Vitest shard in 7.56s
```

- **Observed result after fix:** `/download?id=123` with `filename="photo.png"` is stored as `photo.png`, not `download`.
- **What was not tested:** A live Tlon/Urbit ship upload session.

- [x] AI-assisted (Cursor)

### DIFF
--- a/extensions/tlon/src/urbit/upload.content-disposition.http.test.ts
+++ b/extensions/tlon/src/urbit/upload.content-disposition.http.test.ts
@@ -1,0 +1,98 @@
+// Production uploadImageFromUrl against a real Content-Disposition HTTP response.
+import { createServer, type IncomingMessage, type ServerResponse } from "node:http";
+import type { AddressInfo } from "node:net";
+import { readRemoteMediaBuffer } from "openclaw/plugin-sdk/media-runtime";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { uploadFile } from "../tlon-api.js";
+import { uploadImageFromUrl } from "./upload.js";
+
+const actualReadRemoteMediaBuffer = vi.hoisted(() => ({
+  current: null as typeof readRemoteMediaBuffer | null,
+}));
+
+vi.mock("openclaw/plugin-sdk/media-runtime", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("openclaw/plugin-sdk/media-runtime")>();
+  actualReadRemoteMediaBuffer.current = actual.readRemoteMediaBuffer;
+  return {
+    ...actual,
+    readRemoteMediaBuffer: vi.fn((opts: Parameters<typeof actual.readRemoteMediaBuffer>[0]) => {
+      const origin = new URL(opts.url).origin;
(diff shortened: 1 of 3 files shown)


hugenshen:
@clawsweeper re-review

PR #141330, opened 2026-09-07
feat: choose agent fallback models from a searchable dropdown

### LINKED ISSUE #137410: [Feature]: Model selection dropdown for fallback.
### Summary

Please add a drop-down

Image

 model selection for failback model. 

### Problem to solve

When you go to the Agent configuration, there is a nice model dropdown selection for the primary model.  But the fallback model you need to type in manually.  

### Proposed solution

Just add same kind of selection as primary model. 

### Alternatives considered

_No response_

### Impact

More user friendly and consistent. 

### Evidence/examples

_No response_

### Do you plan to open a PR for this?

No, proposing the idea only

### Additional information

_No response_

shakkernerd:
Fixes #137410

## What Problem This Solves

Agents → Overview offers a catalog picker for the primary model, but configuring fallbacks requires typing model references manually. Operators need the same searchable choices when building a fallback chain.

## Why This Change Was Made

Replace the fallback text field with a reusable, full-width multi-select using the existing model catalog and config save path. Chosen models appear as removable chips in priority order; the dropdown excludes the primary model and choices already in the chain.

## User Impact

Operators can search and select several fallback models, keep custom references through the explicit Add row or typed entry, and remove individual choices. Keyboard selection stays visible when the list scrolls or refreshes, and focus remains visible in forced-colors mode. Enter selects the highlighted option; comma or leaving the field accepts typed references.

## Evidence

- Remote UI and UI-test type checks passed.
- 52 focused unit tests passed across the component, Agents Overview, agent view, and config staging; all three shared-control ownership checks passed.
- Nine browser E2E cases passed across the three affected suites, including the real-Gateway catalog and config-save lifecycle.
- Equivalent browser scenarios reproduced the original input/selection failures and passed with the corrected component.
- An isolated local Gateway serves the remotely built feature for manual testing.

## Visual evidence

| View | Before | After |
| --- | --- | --- |
| Model selection | Original fallback field with two manually entered model references | Searchable fallback field with removable catalog-labeled chips and dropdown |
| Forced colors | Focused fallback field and active dropdown option without visible outlines in forced colors | Focused fallback field and active dropdown option with two-pixel outlines in forced colors |

### Catalog selection, explicit custom Add row, removal, autosave, and forced-color keyboard focus

https://github.com/user-attachments/assets/93660761-f1a8-413b-98f4-a5dd07edce9e

### Visual verification

- Baseline: 01e307c2c2956a9550525243d135b9a93eb01338; preserved main panel/styles and original component CSS
- Exact head: `8afc6b6a9833908db50a164ffb67c4df08120fde`
- Scenario: Choose a catalog model, explicitly add an unlisted reference that also prefixes catalog IDs, remove a selected model, observe the saved ordered chain, and show keyboard targets in forced colors.
- Runtime/build: OpenClaw 2026.9.2; UI build 2026.9.2-12656d88c30a-2026-09-07T16-01-59.841Z; matching UI sources in Chromium 144.0.7559.0 with the maintained mock Gateway fixture
- Saved order: fixture/model-004 → fixture/model-014 → fixture/model-00
- Custom prefix collision: Explicit Add row preserves fixture/model-00 while fixture/model-000 also matches
- Forced-color outlines: Field and active option: 0px before → 2px after; active option offset -2px
- Keyboard scroll: Original active row was outside the menu; repaired scrollTop 487 keeps it fully visible
- Identical 1440×1000 desktop fixture, theme, locale, route and initial chain; each before/after pair uses matching crop coordinates
- All eight focused repaired interaction scenarios passed; original component bytes reproduce six distinct interaction defects
- Readable recording includes factual input cues, resting states, an explicit custom Add-row click, saved-result assertion and forced-color keyboard outline assertions
- Original full frames and recording retained; media contains only synthetic model/config data
- Normal-theme images are behavior-equivalent across the redundant return removal, documentation update and forced-colors-only styling


shakkernerd:
Added: searchable fallback model selection with ordered, removable choices and custom entries.

PR #141333, opened 2026-09-07
test: make doctor archive checks portable across runtimes

steipete:
## What Problem This Solves

Three positive doctor migration fixtures fail under Bun when an archive exists because they expect the successful return value of native filesystem access to be `undefined` rather than `null`.

## Why This Change Was Made

Await access to the expected archive directly in three assertions across the Codex and memory-core fixtures. Existing migrated content, source-absence, ownership results, and cleanup assertions remain unchanged. Seventeen other access sites are untouched.

## User Impact

These selected archive checks validate successful file access under both Node and Bun. No production, data format, schema, dependency, timeout, or protocol changes.

## Evidence

- Three existing positive migration cases passed under Node before the repair; Bun failed all three specifically on `null` versus `undefined`. All three pass under both runtimes afterward.
- Runs used isolated temporary HOME/state, the established macOS SQLite preload for Bun, original deadlines, and bounded process monitoring. All processes joined without caps. Other cases were not run locally.
- The full two-file changed gate passed, including all selected static/type/lint checks, with no additional runtime tests or full build. Fresh independent P0–P2 review passed with no actionable findings.
- The final two source hashes were verified before staging. Initial zero-selection startup attempts and the fail-fast baseline routing are retained and are not credited as passing proof.

This is limited fixture portability evidence, not a full doctor-suite or upstream runtime compatibility claim. Exact-head CI34141420084 and pull_request WorkflowSanity34141352194 passed, and hosted review is ready with no actionable findings.

### DIFF
--- a/extensions/codex/doctor-contract-api.test.ts
+++ b/extensions/codex/doctor-contract-api.test.ts
@@ -527,7 +527,7 @@ describe("codex doctor contract", () => {
         storePath: fixture.storePath,
       }),
     ).toMatchObject({ agentHarnessId: "codex" });
-    await expect(fs.access(`${fixture.sidecarPath}.migrated`)).resolves.toBeUndefined();
+    await fs.access(`${fixture.sidecarPath}.migrated`);
 
     await removeCodexDoctorFixture(fixture.stateDir);
   });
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ae60b4ff1f2eaf101287fc2481f02bda82d5164f](https://github.com/openclaw/openclaw/pull/141333/commits/ae60b4ff1f2eaf101287fc2481f02bda82d5164f)
- Landed commit: [b4ce0df975fab5f1a5bc234ad9643c6de40c9eb0](https://github.com/openclaw/openclaw/commit/b4ce0df975fab5f1a5bc234ad9643c6de40c9eb0)

PR #141335, opened 2026-09-07
improve(ui): show sidebar unread state on the row icon instead of the trailing edge

steipete:
## What Problem This Solves

Resolves the last inconsistency in where the Control UI sidebar shows row state. After #141098, a running session rings its own icon on the left, and the trailing edge of a session row is for hover-revealed actions. Unread was deliberately left behind in that change: child rows and Home already draw unread as a corner badge on their icon, while top-level session rows still drew a trailing dot that competed with the same pin and menu controls on hover.

## Why This Change Was Made

Top-level session rows now draw unread the way child rows and Home already do: a corner badge on the row's own icon, or the dot centered in the icon slot when a row has no icon, avatar, or owner chip. The trailing state slot is gone from session rows entirely, so their trailing edge holds only hover actions plus the PR, count, and draft badges. Precedence is unchanged: while a row runs, only the ring is drawn and unread is announced rather than shown. Each state is announced once per row: the badge announces unread; the row description covers fork, and unread only while the badge is hidden by a run.

Other unread surfaces are untouched: the Sessions page list, the agent card and menu, and the collapsed catalog header dot, which is a right-side summary of hidden rows by design.

## User Impact

Unread reads in the same place for every row in the sidebar, next to the icon, and it stays visible while you hover to pin or open the menu. Hovering an unread row no longer hides the unread mark under the actions. Screen readers hear "Unread" once per row.

## Evidence

Mocked-Gateway Chromium captures (synthetic fixtures only), 1200x800 dark theme. Left before, right after.

| State | Before | After |
|---|---|---|
| At rest | Before: unread rows at rest | After: unread rows at rest |
| PR row hovered | Before: hovered unread PR row | After: hovered unread PR row |

- Regression tests failed on the untouched tree for the intended reason before the production change: `4 failed | 335 passed` in the sidebar component suite (missing glyph badge, missing lead-slot dot, empty lead slot), and `5 failed | 11 passed` across the Playwright suites (missing leading badge/dot locators, trailing state still present).
- After: `node scripts/run-vitest.mjs run ui/src/components/app-sidebar.test.ts` → 339 passed; the Playwright suites that anchor on unread (`chat-flow.sidebar-presentation`, `session-management.trailing-state`, `session-management.unread`, `session-management.queue`) plus the unchanged `chat-run-lifecycle` suite → 27 passed; `node scripts/check-changed.mjs` → exit 0; `node scripts/ui.js build` → startup CSS 44004 B gzip under the 51200 B ceiling. Chromium's accessibility tree confirms exactly one "Unread" announcement for owner-chip, glyph-less, running, and PR rows.
- Production delta: TypeScript +28/−50 and CSS +1/−17 (the trailing state rule, its endcap override, and the unread hover selectors are gone). No config, i18n, snapshot, baseline, or budget changes.
- Autoreview (Codex) on the final tree: scoped-clean at P2 (patch judged correct, 0.88).


steipete:
I inspected all four before/after captures locally before uploading them. They are mocked-Gateway Chromium fixtures with synthetic data only; the unread badge/dot stays visible in the leading icon slot at rest and on hover, and running rows retain ring-only visual precedence.

steipete:
Merged via squash.

- Prepared head SHA: [6c4b6f00c964162483ff888aca9cec0253ca045a](https://github.com/openclaw/openclaw/pull/141335/commits/6c4b6f00c964162483ff888aca9cec0253ca045a)
- Landed commit: [64e65c816044aa02d617a798af6344a01679741f](https://github.com/openclaw/openclaw/commit/64e65c816044aa02d617a798af6344a01679741f)

PR #141336, opened 2026-09-07
fix(desktop): bind observer tickets and viewers to the requesting Gateway connection

steipete:
## What Problem This Solves

Resolves a problem where a Desktop panel observer could outlive its operator's authority: `desktop.observe` minted a one-shot token and relayed RFB bytes on a separate WebSocket, but neither the unconsumed token nor the attached observer was bound to the Gateway connection that requested it. A device whose token was revoked, or any client that disconnected, could still redeem the token within its 60 s TTL and keep receiving framebuffer bytes (and, with `control`, keep sending input). PR #141031 closed the same class of gap for the browser screencast; this brings the desktop observer to parity.

## Why This Change Was Made

The `desktop.observe` handler now derives a requester from the calling client (`connectionSignal` plus a live predicate over the synchronous `invalidated` marker and current client authority) and threads it through the host, node, and worker-environment sources into the observer bridge. Unconsumed tokens are revoked when the connection ends (via the shared ticket store's `revokeSignal`), redemption is refused when the requester is no longer current, and both relay directions are fenced per chunk: a revoked requester's observer closes with `4006 authority_revoked` before the next frame is forwarded and before any further input reaches the desktop. Internal callers without a connection keep TTL-only tokens, as documented.

## User Impact

Revoking a device token or disconnecting an operator now ends that operator's desktop observation immediately, including observers whose socket close handshake is still pending. No change for ordinary sessions.

## Evidence

- Suites: `node scripts/run-vitest.mjs src/gateway/desktop src/gateway/worker-environments/environment-access.test.ts src/gateway/worker-environments/node-desktop-carrier.test.ts src/gateway/server-methods/environments.desktop.test.ts src/shared/one-time-ticket-store.test.ts` → 182 + 17 passed. New cases: stale requester refused at redemption before its signal aborts and within TTL; signal abort revokes an unconsumed token; an attached observer whose requester is invalidated while its socket close never fires gets no further desktop chunk, forwards no input, and closes with `[4006, "authority_revoked"]`; a current requester keeps relaying; requester reaches the mint for host, node, and environment sources.
- `node scripts/check-changed.mjs` clean (core-only change).
- Live proof (loopback RFB 3.8 stub with VncAuth, isolated Gateway built from this branch, repository Gateway client): (1) mint via a connection, close it, then attempt the observer upgrade → **HTTP 401**; (2) mint on a live connection, attach, receive `RFB 003.008` plus five more frames, then close the minting connection → observer closed with **4006 `authority_revoked` in 1 ms**, **0 frames** during the following 1.1 s quiet window.
- Codex autoreview (P0/P1): scoped-clean, no P0/P1 findings.

### DIFF
--- a/docs/gateway/config-browser-ui-desktop.md
+++ b/docs/gateway/config-browser-ui-desktop.md
@@ -180,6 +180,11 @@ machine. It can attach to an existing loopback RFB server, or supervise a
 headless TigerVNC/XFCE desktop on Linux. It is a Labs feature and is off by
 default.
 
+Observer tokens and observer connections are bound to the Gateway connection
+that requested them. Ending or revoking that connection refuses unused tokens
+and closes its observers with `4006 authority_revoked`. Internal callers without
+a Gateway connection keep TTL-only tokens.
+
 ```json5
 {
   desktop: {
--- a/src/gateway/desktop/host-source.ts
+++ b/src/gateway/desktop/host-source.ts
@@ -10,6 +10,7 @@ import {
   type ManagedLinuxDesktopStatus,
 } from "./managed-linux.js";
 import { mintDesktopObserverToken } from "./observe-bridge.js";
+import type { DesktopObserveRequester } from "./observe-requester.js";
 import { classifyRfbSecurity, probeRfbServer, type RfbProbeResult } from "./rfb-probe.js";
 import type { DesktopSessionRegistry } from "./session-registry.js";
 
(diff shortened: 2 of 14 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2ec77047446ecf0a8773b7aa612fae2e4cb61b9c](https://github.com/openclaw/openclaw/pull/141336/commits/2ec77047446ecf0a8773b7aa612fae2e4cb61b9c)
- Landed commit: [922570204cc55d5c03ad7577ea4f5d7f3dcb404f](https://github.com/openclaw/openclaw/commit/922570204cc55d5c03ad7577ea4f5d7f3dcb404f)

PR #141174, opened 2026-09-07
feat(agents): add role templates and a team preset for onboarding

steipete:
## What Problem This Solves

Creating a purpose-built agent means hand-writing `AGENTS.md`, `SOUL.md`, and `IDENTITY.md`, then wiring delegation config by hand. Setting up a small coordinated team (one agent you talk to, a few specialists it delegates to) requires knowing `subagents.allowAgents`, `delegationMode`, workspace layout, and the explicit-ownership rules. Most people never get past a single agent, and the ones who try teams get the delegation wiring wrong.

## Why This Change Was Made

Ships a bundled role catalog and a team preset, using the existing agent creation, workspace seeding, identity, and config-mutation owners:

- **Role templates** under `docs/reference/templates/roles/` for `coordinator`, `researcher`, `writer`, and `reviewer`. Each role is a Claw source (RFC 0016): a `CLAW.md` whose frontmatter carries the agent id, display name, one-line description, and identity, whose Markdown body becomes `SOUL.md`, and whose `workspace/AGENTS.md` is the operating program. Roles are read through the shared Claw reader and schema, so `openclaw claws inspect docs/reference/templates/roles/<role>` validates them and `claws add` on the same source works once the experimental Claws surface is enabled. There is no second manifest format (the first version of this PR had one; it was replaced, and the separate export/import PR #141339 was closed in favor of Claws). Each program covers: scope, what to do on a task, hand-off contract (return verifiable artifacts with paths/links), approval gates (never send, publish, purchase, delete, or change production without approval), escalation rules, and memory hygiene. The coordinator program keeps the human's single point of contact, discovers specialists with `agents_list`, delegates with `sessions_spawn`/`sessions_send`, verifies results, and never lets two specialists loop on each other.
- **`openclaw agents add <id> --role <role>`** seeds the workspace from the role, sets config identity from the manifest, and completes the identity ceremony (no `BOOTSTRAP.md`). Works interactively, with `--non-interactive` (no `--workspace` needed), and with `--json`.
- **`openclaw agents team create`** (`--preset team`, `--coordinator`, `--prefix`, `--workspace-root`, `--non-interactive`, `--json`) creates the coordinator plus researcher, writer, and reviewer with separate workspaces, sets the coordinator's `subagents.allowAgents` to the specialist ids with `delegationMode: "prefer"`, and gives specialists `allowAgents: []`. Id collisions are rejected before anything is written.
[...]

PR #141339, opened 2026-09-07
feat(agents): export and import portable agent templates

steipete:
Stacked on #141174 (role templates and team preset); this PR targets that branch and will retarget to `main` once it lands.

## What Problem This Solves

There is no way to hand a configured agent to someone else, or to move one between installs, without copying workspace files by hand and re-deriving identity, skills, model, and delegation config. Anything copied by hand also risks carrying credentials, memory, or personal files along.

## Why This Change Was Made

Adds a portable, secret-free agent template bundle and two commands built on the shared creation path and the role-template manifest from #141174:

- **`openclaw agents export <id> --out <dir> [--force] [--json]`** writes `manifest.json` (identity, title/summary, explicitly configured skills allowlist, agent-level model, `subagents.allowAgents` as ids plus `delegationMode`), `workspace/{AGENTS.md,SOUL.md,IDENTITY.md}`, and optional `automations.json` with sanitized agent-turn jobs owned and executed by that agent (name, schedule, prompt, model/thinking/timeout only). It never includes `USER.md`, `MEMORY.md`, `memory/`, `BOOTSTRAP.md`, auth profiles, sessions, agentDir, delivery targets, tool policy, or command/script/system-event jobs, and it prints an omissions report saying so. Export refuses when the secret scanner finds a probable literal secret (no bypass flag); embedded absolute paths are reported as review warnings, not refusals. Output never overlaps the source workspace, agent state, or state dir; a non-empty output needs `--force` and is replaced atomically.
- **`openclaw agents import <dir> [--id <newId>] [--workspace <dir>] [--non-interactive] [--json]`** validates the bundle (strict manifest, exact file set, 32 files / 256 KiB per file / 2 MiB total, no symlinks, no path escape), creates the agent through `createAgent` with identity completed (no `BOOTSTRAP.md`), seeds the workspace program files, applies skills/model/delegation (only delegation ids that exist in the target roster are kept; dropped ids are listed), and creates automations disabled and retargeted to the new id with a "review before enabling" line. Interactive runs confirm the summary first.

No new config keys. The bundle format is documented for tool authors in `docs/reference/agent-templates.md`.

## User Impact

Share an agent with a link to a folder, install one from someone else, or move an agent between machines with `export`/`import`, without ever copying credentials or memory. Role agents from #141174 round-trip unchanged.

## Evidence

| Before | After |
| --- | --- |
| before | after |

Round trip in an isolated state dir (role agent → export → import as a new id):

round trip

- Focused tests (templates command suite, template catalog, CLI registration, command JSON classification): 71 passed on the final head; the full templates suite covers exact exported file set and exclusions, secret refusal on a planted fake key, identity/skills/automation round trip with disabled retargeted jobs, path-escape and oversize rejection, and existing-id refusal. `node scripts/check-changed.mjs` passed.
- Independent review pass on the branch: no actionable findings.

Line delta (task only, on top of #141174): production +749, tests +469, docs +217. The growth is the bundle IO/validation, cron conversion, and two commands; nearby cleanup removed duplicate catalog file loading and reused the existing identity parser and cron validators.

## Known Gaps

- Pattern-based secret detection is conservative, not proof; the docs tell operators to inspect a bundle before sharing.
- Import is not a single transaction across workspace, config, and automations: an automation write failure after the agent exists reports per-item outcomes and exits non-zero rather than rolling back.
- Role origin is not persisted, so delegation targets travel as ids, not role names.

### DIFF
--- a/docs/.i18n/glossary.zh-CN.json
+++ b/docs/.i18n/glossary.zh-CN.json
@@ -1308,6 +1308,10 @@
     "target": "智能体"
   },
   {
+    "source": "Agent templates",
+    "target": "智能体模板"
+  },
+  {
     "source": "fs-safe Cleanup Plan",
     "target": "fs-safe Cleanup Plan"
   },
--- a/docs/cli/agents.md
+++ b/docs/cli/agents.md
@@ -3,6 +3,7 @@ summary: "CLI reference for `openclaw agents` (roles, teams, workspaces, routing
 read_when:
   - You want multiple isolated agents (workspaces + routing + auth)
   - You want to create an agent from a role or set up a coordinated team
+  - You want to export or import a portable agent template
 title: "Agents"
 ---
 
@@ -25,6 +26,8 @@ openclaw agents add work --workspace ~/.openclaw/workspace-work
 openclaw agents add work --workspace ~/.openclaw/workspace-work --bind telegram:*
 openclaw agents add ops --workspace ~/.openclaw/workspace-ops --bind telegram:ops --non-interactive
 openclaw agents add research --role researcher --non-interactive
+openclaw agents export research --out ./research-template
+openclaw agents import ./research-template --id research-copy --non-interactive
 openclaw agents team create --non-interactive
 openclaw agents bindings
 openclaw agents bind --agent work --bind telegram:ops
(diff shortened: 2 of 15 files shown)


steipete:
Closing without merging: this duplicates the Claws format from RFC 0016 (`openclaw claws` in `docs/cli/claws.md`, `src/claws/`), which already defines a portable one-agent package with identity, workspace files, skill/plugin/MCP/cron declarations, provenance, and `claws export` of an installed agent. Shipping a second portable-agent format is the wrong call.

The pieces this PR carried that Claws lack (an agent-level model hint and `subagents` delegation targets) move into the Claw OpenClaw profile instead, and the role catalog from #141174 is being reworked to use the Claw format so there is one schema. Follow-up PRs will reference this one.

PR #141345, opened 2026-09-07
fix(agents): rate-limit copy drops the provider reset hint on a failed chain

### LINKED ISSUE #141260: [Bug]: failover rate-limit copy drops the provider reset hint - aggregate "All models failed" string exceeds the 300-char guard
> **Housekeeping:** #141259 is a stray empty issue accidentally opened by a permission probe from this same integration, which turns out to have create-only rights and cannot edit or close it. Please close #141259 — this issue supersedes it.

### Bug type

### Beta release blocker

No

### Summary

When a model-fallback chain exhausts, the user-facing copy is built from the **aggregate** "All models failed (N): …" string rather than from an individual leg's message. That aggregate reliably exceeds the 300-character guard in `extractProviderRateLimitMessage`, so a perfectly good, specific retry hint is discarded and the user is shown the generic `RATE_LIMIT_ERROR_USER_MESSAGE` instead.

The provider told us exactly when the limit resets. The user is told "try again later".

### Steps to reproduce

1. Configure a three-leg chain, e.g. `anthropic/claude-opus-5` → `claude-cli/claude-sonnet-5` → `openai/gpt-5.6-sol`.
2. Exhaust all three so the chain fails with a `rate_limit` classification.
3. Observe the raw aggregate handed to `renderRateLimitBaseCopy`:

```
All models failed (3): anthropic/claude-opus-5: You've hit your session limit · resets 6:20pm (Europe/London) (unknown) | claude-cli/claude-sonnet-5: You've hit your session limit · resets 6:20pm (Europe/London) (unknown) | openai/gpt-5.6-sol: Codex error: The usage limit has been reached (rate_limit)
```

4. Observe the user sees only: `⚠️ API rate limit reached. Please try again later.`

### Root cause

In `src/agents/failover/user-copy.ts`:

```js
function renderRateLimitBaseCopy(context) {
  const raw = context.raw ?? "";
  if (MODEL_CAPACITY_ERROR_RE.test(raw)) return MODEL_CAPACITY_ERROR_USER_MESSAGE;
  return extractProviderRateLimitMessage(raw) ?? RATE_LIMIT_ERROR_USER_MESSAGE;
}

function extractProviderRateLimitMessage(raw) {
  // …
  if (!candidate || !RATE_LIMIT_SPECIFIC_HINT_RE.test(candidate)) return;
  if (isCloudflareOrHtmlErrorPage(withoutPrefix)) return;
  const trimmed = candidate.trim();
  if (trimmed.length > 300 || trimmed.startsWith("{") || /^(?:<!doctype\s+html\b|<html\b)/i.test(trimmed)) return;
  return `⚠️ ${trimmed}`;
}
```

Measuring the aggregate above:

| | value |
|---|---|
| aggregate length | **302** |
| `RATE_LIMIT_SPECIFIC_HINT_RE` matches aggregate | **true** (`\bresets?\b`) |
| first leg alone (`You've hit your session limit · resets 6:20pm (Europe/London)`) | **61** chars |
| hint matches first leg | **true** |

It misses the cap by **two characters**. The hint regex already matched — the guard then throws the match away. Because the aggregate grows with the number of legs and the length of each provider message, any chain of 2–3 legs with reasonably verbose provider text will cross 300 and permanently degrade to the generic copy. A longer chain makes it *more* certain, not less.

### Expected behavior

Extract the retry hint from an individual leg's message — most usefully the **primary/first** leg, which is the one the user cares about — rather than from the concatenated summary. The user should see something like:

```
⚠️ You've hit your session limit · resets 6:20pm (Europe/London)
```

The 300-char guard is sensible for a single provider message; it just should not be applied to a string that is structurally guaranteed to be a concatenation of several of them.

### Actual behavior

All specificity is lost. The user sees a bare "API rate limit reached", which:

- gives no reset time even though one was returned,
- misattributes the failure — here the only genuinely exhausted leg was the third (Codex), while the message reads as though the user's primary provider quota is gone,
- gives no signal about which leg actually failed, so the user cannot tell whether to wait, switch model, or fix a credential.

### Suggested fix

In `renderRateLimitBaseCopy`, prefer per-leg extraction before falling back to the aggregate:

- if the raw is recognisably an `All models failed (N): …` aggregate, split on the leg separator and run `extractProviderRateLimitMessage` over each leg in order, returning the first hit;
- otherwise behave as today.

That keeps the existing guard semantics for genuine single-provider payloads while restoring the hint for the multi-leg case.

### OpenClaw version

2026.9.2 (`3928bad`)

### Operating system

macOS 26.5.2 arm64

### Install method

npm global; Gateway managed as a macOS LaunchAgent

### Model

Chain: `anthropic/claude-opus-5` → `claude-cli/claude-sonnet-5` → `openai/gpt-5.6-sol`

### Related

- #113159 — usage-limit cooldown ignores the real reset time for non-OpenAI providers. Related but distinct: that one is about the cooldown *state*, this one is about the *user-facing copy* discarding a hint it already parsed.
- #115642 — billing cooldown outlives the outage. Adjacent; a user hitting that issue will see this generic message and be unable to diagnose it.

ayaangazali:
## What Problem This Solves

Fixes #141260.

When a model fallback chain exhausts, the user is told "try again later" even though the provider said exactly when the limit resets.

The rate-limit copy is rendered from the aggregate summary rather than from any single leg:

`src/agents/failover/user-copy.ts`

```ts
function renderRateLimitBaseCopy(context: FailoverUserCopyContext): string {
  const raw = context.raw ?? "";
  ...
  return extractProviderRateLimitMessage(raw) ?? RATE_LIMIT_ERROR_USER_MESSAGE;
}
```

`extractProviderRateLimitMessage` bounds provider text at 300 characters so an HTML error page or a JSON blob never reaches the user. That bound is right, but the aggregate is not provider text, it is every leg concatenated:

```
All models failed (3): anthropic/claude-opus-5: You've hit your session limit · resets 6:20pm (Europe/London) (unknown) | claude-cli/claude-sonnet-5: ... | openai/gpt-5.6-sol: ...
```

That is 302 characters for three legs, so the hint regex matches and the length check then throws the match away. The first leg on its own is 61 characters and carries the whole hint. Longer chains and more verbose provider text make this more certain, not less.

## Why This Change Was Made

The legs are read individually when the summary carries the aggregate prefix, preferring the first, which is the route the user actually chose.

The full string is still tried first, so a single-leg failure behaves exactly as before and nothing about non-aggregate copy changes. The 300 character bound still applies to each candidate, so the guard against HTML and JSON payloads is intact; the change only stops one long string from hiding the short messages inside it.

Splitting is anchored on the `All models failed (N):` prefix rather than on separators alone, so an ordinary provider message that happens to contain a pipe is never split. Both separators that appear in existing fixtures are handled, and the `provider/model:` route label is stripped from each leg so the user sees the provider's sentence rather than routing detail.

## User Impact

A user whose chain exhausts on a rate limit now sees the provider's own hint:

```
⚠️ You've hit your session limit · resets 6:20pm (Europe/London)
```

instead of `⚠️ API rate limit reached. Please try again later.` When no leg carries a hint, the generic message is still what they get.

## Evidence

Production change is +33/-1 in one file.

The regression test fails on unfixed code. Restoring only `user-copy.ts` from main:

```
× keeps the provider reset hint when the summary exceeds the length guard
  → expected '⚠️ API rate limit reached. Please try…' to contain 'resets 6:20pm (Europe/London)'
  Tests  1 failed | 17 passed (18)
```

With the change, `src/agents/failover/user-copy.test.ts` passes 18. The second new test, that a chain with no hint in any leg still returns the generic copy, passes both before and after, so it shows the fallback path was not weakened rather than merely following the change.

```
node scripts/run-vitest.mjs src/agents/failover/user-copy.test.ts        18 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                              clean
./node_modules/.bin/oxfmt --check <both touched files>                       clean
```

One gap I should state rather than paper over: I could not run the test-types lane or oxlint locally. Both die on this machine for environmental reasons, `tsgo -p test/tsconfig/tsconfig.core.test.json` with exit 137 and oxlint's `tsgolint` crashing, and they behave the same way on an unmodified checkout of main. CI is authoritative for those two. The test file adds no type assertions and calls an existing exported function with a literal `reason`, so the type surface it introduces is small, but I have not verified that locally and did not want to imply otherwise.

### Real reply-path proof, and a correction

Added coverage at the path this change actually serves: the mid-turn surfacing branch in `src/auto-reply/reply/agent-runner-execution.ts:462-480`, where a run produces no usable text and the raw upstream error is rendered for the user through `renderRateLimitOrOverloadedCopy`. The new case in `agent-runner-execution-terminal-failures.test.ts` drives `executeAgentTurn` with an exhausted three leg chain and asserts the delivered reply carries the hint.

It fails on unfixed source for the right reason. Restoring only `user-copy.ts` and `message-patterns.ts` from main:

```
× keeps the provider reset hint when the chain summary exceeds the length guard
  → expected '{"kind":"success","runId":"16442fdd-c…' to contain 'resets 6:20pm (Europe/London)'
  Tests  1 failed | 27 passed (28)
```

With the change, that file passes 28 and `user-copy.test.ts` passes 18.

The correction: my earlier User Impact claim was broader than what this fixes. While building the proof I found there are two rate-limit reply paths, and they do not share a renderer.

- `agent-runner-execution.ts:477` renders through `renderRateLimitOrOverloadedCopy`, so it reaches the guarded extractor. **This is the path this PR fixes, and the one now proven.**
- `agent-runner-failure-reply.ts:110-131` uses `renderRateLimitReplyCopy` for `rate_limit`, reserving `renderRateLimitOrOverloadedCopy` for `overloaded` only. `renderRateLimitReplyCopy` never consults the provider text at all; it returns cooldown or "all attempted models were rate-limited" copy from the attempts list. **That path is unchanged by this PR and still shows no provider hint.**

So this repairs the reported renderer and the path that produced the reported string, but it does not make every rate-limit reply carry the hint. Whether the second path should also surface it looks like a separate decision, and I did not want to widen this diff to make that call. Happy to follow up if you want them unified.

Line budgets checked since CI flagged one earlier: `user-copy.ts` is 692 counted against its 700 limit, and the touched test files are 765 and 306 against the 1000 limit that applies to `*.test.*`.

### DIFF
--- a/src/agents/failover/message-patterns.ts
+++ b/src/agents/failover/message-patterns.ts
@@ -408,3 +408,25 @@ export function isServerErrorMessage(raw: string): boolean {
   }
   return matchesErrorPatterns(scrubbed, ERROR_PATTERNS.serverError);
 }
+
+/** `All models failed (3): provider/model: text | provider/model: text` */
+const FAILOVER_AGGREGATE_PREFIX_RE = /^all(?: [\w-]+)? models failed \(\d+\):\s*/i;
+const FAILOVER_AGGREGATE_LEG_SPLIT_RE = /\s\|\s|;\s(?=\S+\/\S+:\s)/;
+/** `anthropic/claude-opus-5: You've hit your session limit` */
+const FAILOVER_LEG_ROUTE_PREFIX_RE = /^\S+\/\S+:\s+/;
+
+/**
+ * A chain summary concatenates every leg, so it outgrows the length guard that keeps
+ * provider text bounded, and a hint the provider did give is thrown away. Read the legs
+ * individually, preferring the first, which is the route the user actually chose.
+ */
+export function splitFailoverAggregateLegs(raw: string): string[] {
+  const withoutPrefix = raw.trim().replace(FAILOVER_AGGREGATE_PREFIX_RE, "");
+  if (withoutPrefix === raw.trim()) {
+    return [];
+  }
+  return withoutPrefix
+    .split(FAILOVER_AGGREGATE_LEG_SPLIT_RE)
+    .map((leg) => leg.trim().replace(FAILOVER_LEG_ROUTE_PREFIX_RE, "").trim())
+    .filter(Boolean);
+}
--- a/src/agents/failover/user-copy.test.ts
+++ b/src/agents/failover/user-copy.test.ts
@@ -145,3 +145,27 @@ describe("failover user copy", () => {
     );
   });
 });
+
+describe("rate limit copy from a failover chain summary", () => {
+  const aggregate =
+    "All models failed (3): anthropic/claude-opus-5: You've hit your session limit · resets 6:20pm (Europe/London) (unknown) | " +
+    "claude-cli/claude-sonnet-5: You've hit your session limit · resets 6:20pm (Europe/London) (unknown) | " +
+    "openai/gpt-5.6-sol: Codex error: The usage limit has been reached (rate_limit)";
+
+  it("keeps the provider reset hint when the summary exceeds the length guard", () => {
+    // The summary is over the 300 char bound, so reading it whole discards a hint the
+    // provider did give. The first leg is the route the user picked.
+    expect(aggregate.length).toBeGreaterThan(300);
+    const copy = renderRateLimitOrOverloadedCopy({ reason: "rate_limit", raw: aggregate });
+    expect(copy).toContain("resets 6:20pm (Europe/London)");
+    expect(copy).not.toBe("⚠️ API rate limit reached. Please try again later.");
+  });
+
+  it("still falls back to the generic message when no leg carries a hint", () => {
+    const copy = renderRateLimitOrOverloadedCopy({
+      reason: "rate_limit",
+      raw: "All models failed (2): anthropic/claude: 429 (rate_limit) | openai/gpt-5.4: 429 (rate_limit)",
+    });
+    expect(copy).toContain("API rate limit reached");
+  });
+});
--- a/src/agents/failover/user-copy.ts
+++ b/src/agents/failover/user-copy.ts
@@ -18,6 +18,7 @@ import {
   isPeriodicUsageLimitErrorMessage,
   isProviderCompletedErrorFinishReasonMessage,
 } from "./classify.js";
+import { splitFailoverAggregateLegs } from "./message-patterns.js";
 import {
   classifyProviderRequestFacets,
   type ProviderRequestFacet,
@@ -122,28 +123,6 @@ function extractProviderRateLimitMessage(raw: string): string | undefined {
     return undefined;
   }
   return `⚠️ ${trimmed}`;
-}
-
-/** `All models failed (3): provider/model: text | provider/model: text` */
-const FAILOVER_AGGREGATE_PREFIX_RE = /^all(?: [\w-]+)? models failed \(\d+\):\s*/i;
-const FAILOVER_AGGREGATE_LEG_SPLIT_RE = /\s\|\s|;\s(?=\S+\/\S+:\s)/;
-/** `anthropic/claude-opus-5: You've hit your session limit` */
-const FAILOVER_LEG_ROUTE_PREFIX_RE = /^\S+\/\S+:\s+/;
-
-/**
- * A chain summary concatenates every leg, so it outgrows the length guard that keeps
- * provider text bounded, and a hint the provider did give is thrown away. Read the legs
- * individually, preferring the first, which is the route the user actually chose.
- */
-function splitFailoverAggregateLegs(raw: string): string[] {
-  const withoutPrefix = raw.trim().replace(FAILOVER_AGGREGATE_PREFIX_RE, "");
-  if (withoutPrefix === raw.trim()) {
-    return [];
-  }
-  return withoutPrefix
-    .split(FAILOVER_AGGREGATE_LEG_SPLIT_RE)
-    .map((leg) => leg.trim().replace(FAILOVER_LEG_ROUTE_PREFIX_RE, "").trim())
-    .filter(Boolean);
 }
 
 function renderRateLimitBaseCopy(context: FailoverUserCopyContext): string {
--- a/src/agents/failover/user-copy.ts
+++ b/src/agents/failover/user-copy.ts
@@ -124,12 +124,44 @@ function extractProviderRateLimitMessage(raw: string): string | undefined {
   return `⚠️ ${trimmed}`;
 }
 
+/** `All models failed (3): provider/model: text | provider/model: text` */
+const FAILOVER_AGGREGATE_PREFIX_RE = /^all(?: [\w-]+)? models failed \(\d+\):\s*/i;
+const FAILOVER_AGGREGATE_LEG_SPLIT_RE = /\s\|\s|;\s(?=\S+\/\S+:\s)/;
+/** `anthropic/claude-opus-5: You've hit your session limit` */
+const FAILOVER_LEG_ROUTE_PREFIX_RE = /^\S+\/\S+:\s+/;
+
+/**
+ * A chain summary concatenates every leg, so it outgrows the length guard that keeps
+ * provider text bounded, and a hint the provider did give is thrown away. Read the legs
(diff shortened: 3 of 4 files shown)


ayaangazali:
Three attempts at the running-chain evidence, and I have hit an environment wall rather than a code problem. Reporting specifics instead of quietly stalling.

**Attempt 1, real gateway e2e.** Built a proof test on the `test/pr*-proof.e2e.test.ts` convention: isolated state, a mock provider answering 429 with the hint, and a real `chat.send` turn over a two and three leg chain. The chain genuinely ran, my mock counted 8 dispatched requests, but `agent.wait` returned `status: "timeout"` and the transcript held only the user message. The run was still cycling rate-limit retries when the window closed. Capping `OPENCLAW_SDK_RETRY_MAX_WAIT_SECONDS` to 1 did not shorten it.

**Attempt 2, the test now in the PR.** Drives `executeAgentTurn` so the changed renderer is reached through the real mid-turn surfacing branch, and it fails on unfixed source with the generic copy. Your review is right that it supplies the runner error directly, so it demonstrates the renderer in place rather than a live chain.

**Attempt 3, the e2e again with `retry-after: 0`** to stop the transport sitting in backoff. This time the e2e build prerequisite itself failed: `tsdown` dies at invocation 4 of 11 with no diagnostic, which on this machine is the OOM signature I also get from `tsgo -p test/tsconfig/tsconfig.core.test.json` (exit 137) and from oxlint's `tsgolint`. Roughly 3 GB free with other long-running services on the box. `scripts/run-node.mjs --version` succeeds on its own, so it is resource pressure during the parallel build, not a defect in the branch.

So I cannot produce the live-chain trace here. Two honest options, your call:

1. If the reply-path test is acceptable given it exercises the real branch and fails without the fix, say so and I will leave the PR as is.
2. If a live chain is required, close this. I would rather it close than sit at `needs proof` indefinitely, and the diagnosis plus the two-path finding are recorded on #141260 for whoever has an environment that can run it.

I am not going to claim proof I did not obtain. Also worth repeating from the body: this fixes the mid-turn path only. `renderRateLimitReplyCopy`, which serves `rate_limit` in `agent-runner-failure-reply.ts`, never consults provider text at all, so that path still shows generic copy either way.

ayaangazali:
Following up on my 2026-09-07 note, because I can now replace "I hit an environment wall" with a measurement.

The e2e lane is not runnable on this machine, and the reason is structural rather than a bad run. Any `test/*.e2e.test.ts` pulls in `build-all`, which runs tsdown across eleven invocations:

```
[build-all] tsdown
[tsdown-build] invocation 1/11 finished in 4.9s
[tsdown-build] invocation 2/11 finished in 22.2s
[tsdown-build] invocation 3/11 finished in 19.9s
[tsdown-build] invocation 4/11 finished in 20.8s
```

At invocation 4 of 11 this box was at 3.6 GB of swap against a hard 3 GB working ceiling I am held to, so I stopped it rather than push into another out-of-memory event. On an earlier attempt the same build died at the same invocation on its own. That is twice, at the same place, for the same reason, so I am treating the running-chain proof as something I cannot produce here rather than something to keep retrying.

Nothing about the change has moved. No code findings remain on this PR, the fix still reproduces on current main, and the test that ships with it drives `executeAgentTurn` so the changed renderer is reached through the real mid-turn surfacing branch and fails on unfixed source with the generic copy. Your review is right that it supplies the runner error directly rather than letting a live chain produce it.

So the same two options from a week ago, with better evidence behind them:

1. If the reply-path test is acceptable given it exercises the real branch and fails without the fix, say so and I will leave this as is.
2. If a live chain is required, close it. I would rather that than have it sit at `needs proof` indefinitely, and the diagnosis plus the two-renderer finding are recorded on #141260 for whoever has an environment that can run it.

I am not going to claim proof I did not obtain, and I would rather you close this than have it linger. If I do not hear either way I will close it myself on the next sweep and leave the analysis on the issue.

Still worth repeating from the body: this fixes the mid-turn path only. `renderRateLimitReplyCopy`, which serves `rate_limit` in `agent-runner-failure-reply.ts`, never consults provider text at all, so that path shows generic copy either way.

Issue #141202, opened 2026-09-07
[Bug]: regression race after #120575: terminal process.poll can still trigger duplicate notifyOnExit wake


wpu911:
﻿### Bug type

Regression / race condition in background exec completion acknowledgement

### Summary

On OpenClaw `2026.9.2`, a background `exec` completion can still trigger a second `exec-event` agent wake even when the originating turn has already observed the terminal result through `process poll`.

This appears to be a remaining timing race after #120488 / #120575. The one-use completion receipt machinery exists, but the terminal poll acknowledgement can happen after `notifyOnExit` has already scheduled/emitted its system event and heartbeat wake.

For local single-slot LLM backends, the duplicate wake is especially expensive because it can cause a full long-context prefill with no reusable prefix.

### Live observation

Environment:

- OpenClaw: `2026.9.2`
- Windows
- Node `22.23.2`
- local OpenAI-compatible provider -> llama-swap -> llama-server
- recurring heartbeat disabled (`every: "0m"`)
- `tools.exec.notifyOnExit = true`

One interactive turn started at `18:57:46` with run id:

```text
50fef049-ae09-42bf-8e0b-ce28b79ba7a3
```

The first model phase took ~265 s before tool execution. After several tool/model continuations, OpenClaw started another embedded run at `19:05:08`:

```text
34c40114-86dc-4c7e-bf08-5152d69c6f2e
```

There was no new `chat.send` before this second run. It shared the same trace/parent chain as the original turn.

On the llama backend, that second wake corresponded to a fresh full prompt load rather than continuation cache reuse:

```text
Prompt: 191,557 tokens
Cached: -
Generated: 25 tokens
Duration: ~269 s
```

The previous continuation had ~194K cached tokens. So this background wake converted a tiny completion/control event into another multi-minute full-context prefill.

### Expected behavior

If a terminal `process poll` has already observed a background process completion, the matching `notifyOnExit` path must be suppressed/consumed before any duplicate system event or heartbeat wake is emitted.

Unobserved background completions should continue to notify normally.

### Actual behavior

A completion can still be consumed by `process poll` and also cause a separate async wake, depending on ordering.

This produces duplicate agent work and, on local stateful backends, can invalidate or bypass a large cached prefix.

### Local fix validated

I tested a narrow race fix in the installed `dist` runtime:

1. Add a short `notifyOnExit` grace period (`1000 ms`) after process exit.
[...]

PR #141352, opened 2026-09-07
fix: avoid duplicate agent turns for acknowledged completions

maximilliangrand:
Related: #141202

<details>

Allow edits from maintainers is enabled.

</details>

## What Problem This Solves

Resolves a problem where reading a completed background command can still be followed by another agent turn if its completion is acknowledged while heartbeat preparation awaits routing or runtime setup. The extra turn receives an obsolete completion notice and can repeat expensive long-context inference.

## Why This Change Was Made

Build the event prompt after heartbeat preparation and revalidate the original queue occurrences immediately before generic reply dispatch. Outcome classification and successful handling consume that same executed selection. Identical successor events keep their own identities and remain queued; no grace timer or early acknowledgement is added.

## User Impact

Already-observed completions no longer trigger this redundant follow-up. Unobserved completions, tagged cron work, due tasks, and scheduled cadence continue to run. Dispatch and delivery failures retain unread events for recovery.

Thanks to @wpu911 for the report and ordering examples in #141202.

## Evidence

- On unmodified main `5bba4d8b8802304968116fac4e44c6500095b82b`, the final regression file has **7 failures / 12 passes**: consumed completion text survives routing, and acknowledgement during published-runtime preparation still dispatches. With the fix, **19/19 pass**.
- **230 focused tests passed**, covering heartbeat event routing, isolated sessions, structured delivery, model/tool settings, queue receipts, terminal poll persistence, and direct bash polling.
- Independent `autoreview` completed with no actionable P0–P2 findings.
- Full `pnpm build` passed on macOS / Node 26.7.0.
- Full `node scripts/check-changed.mjs` passed: production and all sixteen core-test typecheck groups, formatting, lint, dead-export checks, import-cycle checks, and repository guards.
- `git diff --check` passed.

Proof uses the real heartbeat runner, queue receipts, temporary SQLite session state, and heartbeat runtime facade, with controlled routing/runtime/model/transport boundaries. It does not reproduce the reporter's live Windows/provider trace or claim their timing measurement. Existing route selection is preserved; this does not retract a turn after generic reply dispatch has begun.

### DIFF
--- a/src/infra/heartbeat-runner-delivery.ts
+++ b/src/infra/heartbeat-runner-delivery.ts
@@ -265,6 +265,7 @@ export async function finalizeHeartbeatOutcome(params: {
   opts: HeartbeatRunOptions;
   wake: ReadyHeartbeatWake;
   prepared: PreparedHeartbeatRun;
+  inspectedSystemEventsToConsume: CompletedHeartbeatAgentRun["inspectedSystemEventsToConsume"];
   outcome: ClassifiedHeartbeatOutcome;
   replyPayloadSource: CompletedHeartbeatAgentRun["replyPayload"];
   maybeSendHeartbeatOk: () => Promise<boolean>;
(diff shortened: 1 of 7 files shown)


maximilliangrand:
Rebased onto `ebfe25f8` and updated the description with actual runtime before/after proof. The committed regression launches a background process and exercises process polling, the production acknowledgement hook, real reply/agent dispatch, a deterministic loopback provider, and transcript persistence. An acknowledgement during preparation now produces zero model requests and zero assistant turns; the unobserved control still produces one of each.

The description also states the precise preparation boundary and distinguishes this reproduction from the reporter's Windows/provider timing trace. Coalesced work retains its own event identities and delivery policy, with regression coverage for WebChat publication and route changes.

PR #140766, opened 2026-09-07
fix(crabbox): skip redundant inspection after warm capture

steipete:
## What Problem This Solves

Completed warm-image captures appeared pending, so the first subsequent cloud session performed an unnecessary provider inspection before reusing them. A real AWS capture returned native state `completed` with coordinator state `ready`; the Gateway stored `pending` because it recognized only the native word `available`.

## Why This Change Was Made

The capture owner always requests `checkpoint create --mode native --wait --json`. Successful completion is the portable readiness contract, so the owner now records availability directly. JSON, checkpoint identity, lease identity and native-record validation remain intact; failed or uncertain capture outcomes remain owned for recovery.

The dependency was directly inspected at Crabbox commit [4f2ccff7ff88](https://github.com/openclaw/crabbox/commit/4f2ccff7ff8875296064d972477fb4f62ff2518d), the CLI source used for fresh proof: [waited readiness and identity validation](https://github.com/openclaw/crabbox/blob/4f2ccff7ff8875296064d972477fb4f62ff2518d/internal/cli/checkpoint_native.go#L230), and [successful waited result publication](https://github.com/openclaw/crabbox/blob/4f2ccff7ff8875296064d972477fb4f62ff2518d/internal/cli/checkpoint.go#L269). Managed readiness accepts the provider's `available`, `ready`, `succeeded` and `completed` terminal states; Machine0's image owner independently waits for its underlying snapshot readiness. This follows the existing successful-wait contract rather than adding another provider-state alias list.

## User Impact

The first reuse after a successful capture skips the redundant readiness inspection. AWS and Machine0 native state labels work consistently. Existing pending images still use verified recovery; successful forks already avoid repeated inspections on later reuse. No schema bump, table change, stored representation, configuration option or retention change.

## Evidence

- AWS `completed` and Machine0 `ACTIVE` lifecycle regressions fail before the fix and pass afterward. All 132 lifecycle/project tests and the full changed-surface gate pass. The final cancellation fixture explicitly seeds a persisted pending image so inspection cancellation remains covered.
- [Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34084192603) on `fa8062e640b3859469a7fd25ab90471d498e7e84`.
- Fresh AWS cold-session proof passed: a synthetic repository session completed a real tool turn, retained all eight files, captured an EBS checkpoint and released its lease with provider cleanup complete. The Gateway stored the captured image as `available`; independent provider verification returned `completed` and `fork_or_delete`.
[...]

PR #134931, opened 2026-09-01
feat(cloud-workers): reuse project snapshots and keep workers ready

steipete:
## Superseded by landed split PRs

The useful snapshot-preparation, ready-worker reuse, public/private repository and lifecycle changes have landed through smaller PRs across OpenClaw and Crabbox, followed by the managed-CLI rollout in #145109. Each piece was improved against current main, independently reviewed and validated through its relevant runtime flow. This combined prototype is superseded; its branch is preserved rather than merged wholesale. Optional prototype differences and the separate release rollout are recorded below.

| Piece | State |
| --- | --- |
| Preserve lease renewal after temporary heartbeat failures | [#142194](https://github.com/openclaw/openclaw/pull/142194) **merged**; focused checks, independent review, actual CLI/process proof and required CI passed |
| Support workspace paths admitted by project preparation | [#142198](https://github.com/openclaw/openclaw/pull/142198) **merged**; actual child-worker/IPC proof, checks, review and CI passed |
| Recognize checkpoint creation rejected before provider submission | [#142221](https://github.com/openclaw/openclaw/pull/142221) **merged**; 420 tests, full checks, fresh independent review and exact-head CI passed. Required-project and optional-teardown failure/cleanup paths passed real AWS proof; native landing and resulting source are verified. Producer follow-up [Crabbox #2042](https://github.com/openclaw/crabbox/pull/2042) is also **merged** after all required/relevant CI passed; all six landed files and main ancestry verified |
| Settle cancelled workspace uploads | [#142237](https://github.com/openclaw/openclaw/pull/142237) **merged**; real HTTP cancellation/control proof, independent review and fresh required CI passed |
| Compress negotiated workspace manifests | [#142242](https://github.com/openclaw/openclaw/pull/142242) **merged**; 235 tests plus one existing Windows-only skip, full checks, independent review and real HTTP proof passed (68.73% fewer manifest bytes). A cache-service delay required one unchanged-head artifact-job retry; required CI and native landing completed |
| Keep stream diagnostics in their owning test | [#143133](https://github.com/openclaw/openclaw/pull/143133) **merged**; one-file fixture repair, original-order leakage 25/25 before to 0/25 after, 19 existing tests and full checks/review/CI passed |
| Avoid stale sockets in Telegram callback fixtures | [#143197](https://github.com/openclaw/openclaw/pull/143197) **merged**; reproduced HTTP failure, 57 focused tests, full checks/review and exact-head CI passed |
[...]

PR #141375, opened 2026-09-07
fix(crabbox): reuse updated runtimes in warm cloud sessions

steipete:
Related: #140766.

## What Problem This Solves

Fixes an issue where cloud sessions reinstall the same updated runtime on every start because teardown keeps a younger image containing the previous runtime. Repository-only sessions can therefore repeatedly pay for download and npm installation even after a successful upgrade.

## Why This Change Was Made

Core carries the already-prepared runtime identity into provisioning and verifies the actual artifact before enrollment or capture. Crabbox records that identity in the existing allocation and image metadata and refreshes an image when its runtime differs. An older image remains a useful setup base until its replacement is captured. Existing replay choices, capture ownership, pinned-image retirement, profile limits and age policies remain in force.

The maintainer approved the cache metadata and freshness change with no database schema change. This adds no SQL table, column, index, schema-version bump or configuration option. It does not include the separate prepared-worker pool in #134931.

## User Impact

After a runtime update, the first eligible session captures the new installation for subsequent sessions. Repository-only cloud sessions continue to work without a local checkout. An interrupted allocation cannot silently switch its recorded runtime identity; Stop and reprovision resolve a changed runtime.

## Evidence

The original failing reproduction uses the real warm-image manager, generated bootstrap script, local HTTP archives and dependency-free npm installs. It installed the new runtime in one child, retained the old image, then repeated npm installation in the next child. The repaired regression captures a replacement and the next child skips installation while keeping one profile record.

On candidate `8ad3b780915b4ce6c7d4198761fea5fee0c80fcb`, 447 Crabbox tests passed (one skipped), 114 core lifecycle tests passed, and the canonical changed-file gate passed. Managed independent review through P2 is clean. Coverage includes changed node/worker artifacts, execution modes, missing metadata, replay drift, stale capture attempts, grant closure, restart failure reporting, and existing image pins/retirement.

An intermittent real-child stop fixture missed its filesystem release notification. A phase-instrumented reproduction showed the child still alive after the release rename; replacing only that fixture signal with real-child marker polling passed the same virtual deadline and custody assertions. Production timeouts are unchanged, and the orphaned test processes were confirmed gone.

Production delta is +134/-32 (net +102), including a 17-line pure move of the existing profile-key helper to its profile owner. Growth carries the previously missing prepared/attested runtime fact across the existing lifecycle; duplicate provider checks and an unnecessary SDK export were removed. Tests/support are +399/-8; docs are +7/-3. No schema, configuration, SDK budget or line-limit suppression was added.

### Native AWS update and reuse

Both stages passed on the exact committed/full-built candidate, using the same CLI and coordinator version. The first restored 11 files, completed a real tool turn and saved the twelfth, then captured a replacement image. Ordinary Stop completed before the last image pin was released and the predecessor was deleted. The second restored all 12 files, completed another real tool turn and saved the thirteenth, reused the exact runtime digest, and retained the same replacement image without recapture.

| Observed step | Runtime update | Same-build reuse |
| --- | ---: | ---: |
| Node runtime download | 12.317 s | skipped |
| npm installation | 36.567 s | skipped |
| Plugin activation | 2.837 s | 33.624 s |
| Lease allocation | 65.569 s | 147.528 s |
| Dispatch to active | 174.248 s | 293.915 s |
| Real tool turn | 56.345 s | 72.979 s |

The cache regression is fixed, but these samples do **not** establish faster overall startup. AWS request time rose from 7.551 s to 79.426 s, and plugin activation also slowed; ingress waiting was zero in both stages. Those remaining owners are being investigated separately. No in-path verification command was added to either dispatch measurement.

The first replacement capture took 121.515 s during teardown. Its source lease Stop completed before predecessor retirement; the second stage did not recapture. Both leases are released with provider cleanup complete, environments destroyed and attachments empty. All 41 first-stage and 34 second-stage process birth identities are absent, and both Gateway ports are closed. Full SQLite schema definitions and version markers are identical before/after both stages (shared 16, agents 19). No coordinator version drift was observed.

### Interrupted allocation disposition

Cloud-session backward compatibility is explicitly not a maintainer goal for this repair. An older interrupted allocation without runtime identity is rejected with Stop/reprovision guidance; it is not silently relabeled or resumed.

A separate **local synthetic-backend ownership proof** executes the actual older provider/warm-image writer and the current provider. Replay rejects the missing identity without allocating or changing its owner. Current provider destroy retains the original checkpoint pin while a real SDK-owned Stop subprocess is pending, unpins only after that subprocess exits successfully, and admits a fresh allocation with current identity using the existing image base. Final destroy releases all synthetic resources and pins; the child is joined and absent. This qualifies provider ownership and recovery ordering, not a native interrupted-upgrade cloud test. The native pair above separately exercises actual provider Stop, image retirement and fresh provisioning.


steipete:
Merged via squash.

- Prepared head SHA: [8ad3b780915b4ce6c7d4198761fea5fee0c80fcb](https://github.com/openclaw/openclaw/pull/141375/commits/8ad3b780915b4ce6c7d4198761fea5fee0c80fcb)
- Landed commit: [eb439c0d6e9b6a77403c64a96ff96e860ef44fcf](https://github.com/openclaw/openclaw/commit/eb439c0d6e9b6a77403c64a96ff96e860ef44fcf)

PR #141376, opened 2026-09-07
improve(memory): reduce tool preparation cost in large agent fleets

steipete:
## What Problem This Solves

Memory tools take unnecessary CPU time to prepare in gateways with large agent fleets, even when the request already selects its agent.

## Why This Change Was Made

Strict scalar session-agent resolution now shares the existing owner validation and only looks up a legacy/default owner when no explicit, session-scoped, persisted, or prepared owner was selected. Four memory-core consumers use that existing scalar SDK helper. The paired API still computes and returns its compatibility default, and fixed-store, conflicting-owner, retired-owner, and legacy fallback behavior stay intact. This is request-time selection work: no stored fields, vector/embedding metadata, schema, store writes, or migration behavior change.

## User Impact

Memory tool preparation and other strict scalar callers avoid projecting the full roster solely to compute an unused default. Existing recall-policy and per-agent configuration lookup costs remain; this change does not make the complete memory flow constant-time.

## Evidence

The new scalar regression cases fail on the original source with two unrelated roster reads each; the memory factory/prompt regression fails with three unrelated legacy-default reads. After the fix, all 58 selector/SDK cases and 29 memory-core index cases pass. Full changed checks pass, including all four typecheck lanes, lint, plugin boundaries, and repository guards. Independent Codex P0–P2 review found no actionable findings.

Fresh paired synthetic measurements use actual `createMemoryGetTool`, canonical explicit ownership, and nine rounds on the same Linux x64 Testbox with Node 24.19.0, four visible CPUs, and 16.56 GB RAM. Median time per factory call:

| Agents | Owner position | Before | After |
| --- | --- | ---: | ---: |
| 1 | first | 3.482 µs | 3.611 µs |
| 100 | first | 25.501 µs | 14.824 µs |
| 100 | last | 31.780 µs | 22.701 µs |
| 1,000 | first | 218.114 µs | 122.917 µs |
| 1,000 | last | 292.374 µs | 189.249 µs |

The 1,000-agent factory savings are 43.6% for the first owner and 35.3% for the last; the one-agent measurement does not show a factory improvement. Strict scalar lookup itself drops from 111.287 to 0.311 µs for the first owner and 113.773 to 0.315 µs for the last. The complete benchmark command took 11.299 s before and 9.427 s after; whole-process maximum RSS was 657,088 KiB before and 632,256 KiB after. These are synthetic source benchmarks and whole-process observations, not end-to-end latency or per-call allocation measurements.

Real isolated Gateway `/tools/invoke` comparisons pass with 64 agents on both revisions: 28 verified requests total (14 per revision), covering distinct first/last owner memory contents, missing-file `not_found`, and exact conflicting-owner HTTP 400 rejection. The three changed production file hashes independently match this PR's head. Both `pnpm openclaw --version` runtime builds and benchmark phases pass; every owned runtime descendant is verified closed and the [Testbox lease](https://github.com/openclaw/openclaw/actions/runs/34144260735) is stopped.

A separate 1,000-agent startup diagnostic remains unresolved: both baseline and candidate miss the unchanged 180-second readiness budget with sustained runtime-process CPU and no ready socket. The composite remote command deliberately retains exit 1 for that failed diagnostic; it does not represent an aggregate runtime pass. The 64-agent functional receipts are separate successful phases. No startup fix or production-stall causal claim is made. Earlier pre-request harness failures and all startup outcomes are preserved.

### DIFF
--- a/extensions/memory-core/index.test.ts
+++ b/extensions/memory-core/index.test.ts
@@ -108,6 +108,31 @@ function captureMemoryModelContract(initialConfig: OpenClawConfig) {
 }
 
 describe("buildPromptSection", () => {
+  it("prepares explicitly owned memory tools without resolving unrelated legacy defaults", () => {
+    let unrelatedDefaultReads = 0;
+    const config: OpenClawConfig = {
+      agents: {
+        list: [
+          { id: "main" },
+          {
+            id: "unrelated",
+            get default() {
+              unrelatedDefaultReads += 1;
+              return false;
+            },
+          },
+        ],
+      },
+    };
+    const { search, get, promptBuilder } = captureMemoryModelContract(config);
+    expect(search.description).toContain("MEMORY.md");
+    expect(get.description).toContain("MEMORY.md");
+    expect(
+      promptBuilder({ availableTools: new Set(["memory_search", "memory_get"]), agentId: "main" }),
+    ).toContain("## Memory Recall");
+    expect(unrelatedDefaultReads).toBe(0);
+  });
+
   it("returns empty when no memory tools are available", () => {
     expect(buildMemoryPromptSection({ availableTools: new Set() })).toStrictEqual([]);
   });
--- a/extensions/memory-core/index.ts
+++ b/extensions/memory-core/index.ts
@@ -1,4 +1,4 @@
-import { resolveSessionAgentIdsStrict } from "openclaw/plugin-sdk/agent-scope-runtime";
+import { resolveSessionAgentIdStrict } from "openclaw/plugin-sdk/agent-scope-runtime";
 import { createLazyRuntimeModule } from "openclaw/plugin-sdk/lazy-runtime";
 // Memory Core plugin entrypoint registers its OpenClaw integration.
 import {
@@ -109,7 +109,7 @@ function createLazyStandingIntentTool(
     reportUnavailable("runtime config is unavailable for this turn");
     return null;
   }
-  const { sessionAgentId: agentId } = resolveSessionAgentIdsStrict({
+  const agentId = resolveSessionAgentIdStrict({
     sessionKey: ctx.sessionKey,
     config: cfg,
     agentId: ctx.agentId,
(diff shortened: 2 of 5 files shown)


steipete:
Maintainer disposition for the completed review of `1a0ffcb34241d71251df20340ff65684064a71d6`:

The performance checklist used the earlier pending-evidence body. The updated PR body now contains comparable, canonical explicit-owner measurements from the actual memory tool factory: 218.114→122.917 µs for the first owner in a 1,000-agent fleet and 292.374→189.249 µs for the last. The complete paired benchmark commands took 11.299→9.427 s; whole-process maxRSS was 657,088→632,256 KiB. The one-agent factory result (3.482→3.611 µs) does not show an improvement, and the RSS values are not per-call allocation measurements.

Both 64-agent real Gateway API flows passed all 28 requests combined, including distinct owner-specific contents, missing-file handling, and exact conflicting-owner rejection. Production-file hashes match this head; runtime descendants and the Testbox lease are closed. The separate 1,000-agent startup failures remain unresolved on both revisions, and the composite remote command retains exit 1 for that diagnostic. No startup or production-tail improvement is claimed.

The stored-data-model checklist item is a false positive from the two memory file paths. Their production changes only switch existing call sites to the existing strict scalar helper; the core change extracts shared owner validation and defers unused default lookup. No vector/embedding metadata, SQLite schema, persistent-store semantics, writes, or migration code changed. Existing configured/retired fixed-store owner checks and raw/retained legacy fallback tests pass, so no data migration is required by this diff.

The completed review found no correctness or security defects. Exact-head CI is green, and the updated evidence addresses the performance checklist without changing the reviewed production bytes.

steipete:
Merged via squash.

- Prepared head SHA: [1a0ffcb34241d71251df20340ff65684064a71d6](https://github.com/openclaw/openclaw/pull/141376/commits/1a0ffcb34241d71251df20340ff65684064a71d6)
- Landed commit: [0fd838b9c788e257d01ced8b4b40bf195409f021](https://github.com/openclaw/openclaw/commit/0fd838b9c788e257d01ced8b4b40bf195409f021)

PR #141378, opened 2026-09-07
fix: disable desktop keyboard until connected control is ready

steipete:
## What Problem This Solves

Fixes an issue where users could open the desktop keyboard and type while the desktop was still connecting. The text changed the hidden mobile input locally but never reached the desktop.

## Why This Change Was Made

The Keyboard button and its textarea now remain disabled until the desktop is connected and in control mode. The button reuses the existing disabled appearance so that unavailable input is visible. Connected control typing follows the same input path; nothing is buffered or replayed.

## User Impact

Users can see when the desktop keyboard is ready. It stays unavailable while connecting and in view-only mode, and works normally after connected control is ready. This changes no credentials, desktop permissions, configuration, or protocol.

## Evidence

- Extended the existing browser flow to click Keyboard and type synthetic text while the mock desktop observation is deferred. Before the fix, this changed the local input; afterward, the input stays unchanged until connected control is ready.
- All 22 keyboard and desktop-document E2E tests pass. Real noVNC still emits the expected balanced events for 32 ASCII characters, Backspace, BMP/supplementary Unicode, and line endings.
- Readiness assertions now use the actual status role and accessible label; the previous text-based hidden wait could pass without observing that status.
- Full changed-file checks and fresh isolated P0–P2 review pass.

The attachments show the same synthetic connecting state before and after, in that order. No real desktop, credentials, or user data was used. This proof does not establish why a separate OS sign-in failed; live deployment and verification remain a follow-up.

Before: Keyboard appears enabled while the desktop is connecting

After: Keyboard is disabled until connected control is ready

### DIFF
--- a/ui/src/components/desktop/desktop-document-view.ts
+++ b/ui/src/components/desktop/desktop-document-view.ts
@@ -56,6 +56,7 @@ export function renderDesktopDocumentView(options: DesktopDocumentViewOptions) {
         spellcheck="false"
         tabindex="-1"
         aria-label=${t("desktop.keyboardInput")}
+        ?disabled=${options.state !== "connected" || !options.controlling}
         .value=${options.keyboardInputValue}
         @keydown=${options.onKeyboardEvent}
         @keyup=${options.onKeyboardEvent}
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7f6ab93a0dcc5090c8a4e1c97cc44fb6bab2dc05](https://github.com/openclaw/openclaw/pull/141378/commits/7f6ab93a0dcc5090c8a4e1c97cc44fb6bab2dc05)
- Landed commit: [b858b7e34b9e2b96663be8742397fc867e8ef702](https://github.com/openclaw/openclaw/commit/b858b7e34b9e2b96663be8742397fc867e8ef702)

PR #141381, opened 2026-09-07
fix(ollama): preserve reasoning in tool follow-ups

### LINKED ISSUE #141356: Ollama tool follow-ups discard retained assistant reasoning
Native Ollama tool follow-ups lose the reasoning returned with the preceding assistant tool call.

On current main (`871668906e1`), `convertToOllamaMessages` copies assistant text and tool calls but drops retained thinking blocks. The existing history cleanup preserves reasoning for this first current-turn tool exchange, so the loss occurs at the native transport projection.

Reproduced with isolated Ollama 0.33.3 and `qwen3:0.6b`: ask the model to call a synthetic temperature tool for Paris and report the result. The real first response produced 905 characters of thinking plus tool calls. The second native `/api/chat` request contained the calls and synthetic result but omitted the assistant thinking entirely. Preserving that field returns the same correct answer; this is a protocol/state preservation defect, not a claim that the baseline always answers incorrectly.

Expected: forward available non-redacted thinking separately from visible content, alongside tool calls, consistent with Ollama's documented streaming tool loop: https://docs.ollama.com/capabilities/tool-calling#tool-calling-with-streaming

A regression added to the existing conversion test fails on main with the missing `thinking` field (194 other tests pass). A candidate adapter repair passes all 195 and a real two-call production-stream test with normal history cleanup and a read-only payload observer.

steipete:
Closes #141356.

## What Problem This Solves

Native Ollama tool follow-ups discard the reasoning retained from the preceding assistant tool call. This removes part of the model's own response before it receives the tool result, even though Ollama's tool-calling protocol supports a separate `thinking` field.

## Why This Change Was Made

Preserve non-redacted assistant thinking alongside visible text and tool calls when projecting native requests. Include it in fallback prompt-usage estimates when the server omits counters. Existing session history policy still decides which reasoning is retained.

## User Impact

Reasoning-capable Ollama models receive the retained context of their tool calls without mixing it into visible answer text. No configuration change is required.

## Evidence

- Extended the existing replay regression: baseline fails with the missing `thinking` field while 194 other tests pass; final candidate passes all 233 tests across the transport and sibling stream suites.
- Real isolated Ollama 0.33.3 with `qwen3:0.6b`: the first production-stream response emitted 905 reasoning characters plus tool calls. The patched follow-up preserved those bytes exactly in the native request, with ordinary history cleanup and a read-only payload observer, and returned the correct synthetic temperature answer.
- The controlled trial generated 198 follow-up tokens before and 117 after, with increased prompt input and effectively unchanged cache counters. This single prompt does not establish a general speed or quality improvement.
- Owned model, daemon and runner were stopped; the test port is closed. Full `node scripts/check-changed.mjs` passed, including production/test typechecking, lint, formatting, unused-export analysis and repository guards. Final independent review found no actionable P0–P2 findings.

### DIFF
--- a/docs/providers/ollama.md
+++ b/docs/providers/ollama.md
@@ -1069,6 +1069,11 @@ For full setup and behavior, see [Ollama Web Search](/tools/ollama-search).
     `thinking` capability expose `/think low`, `/think medium`, `/think high`,
     and `/think max`; non-thinking models expose only `/think off`.
 
+    When replaying an assistant message, native requests retain its available
+    reasoning in Ollama's separate `thinking` field alongside text and tool
+    calls. This lets tool follow-ups reuse reasoning retained by the session's
+    history policy without mixing it into visible answer text.
+
     ```bash
     openclaw agent --model ollama/gemma4 --thinking off
     openclaw agent --model ollama/gemma4 --thinking low
--- a/extensions/ollama/src/stream-runtime.test.ts
+++ b/extensions/ollama/src/stream-runtime.test.ts
@@ -664,16 +664,20 @@ describe("convertToOllamaMessages", () => {
     });
   });
 
-  it("converts assistant messages with toolCall content blocks", () => {
+  it("preserves assistant thinking alongside text and tool calls", () => {
     const result = convertAssistantContent([
+      { type: "thinking", thinking: "Check the directory.\n" },
+      { type: "thinking", thinking: "Then report its contents." },
+      { type: "thinking", thinking: "redacted reasoning", redacted: true },
       { type: "text", text: "Let me check." },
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [288167be807cf1a5d670bc46de85ed445f0b075f](https://github.com/openclaw/openclaw/pull/141381/commits/288167be807cf1a5d670bc46de85ed445f0b075f)
- Landed commit: [4d4dac1cb63aadfda77cbf57a00838c3e097409e](https://github.com/openclaw/openclaw/commit/4d4dac1cb63aadfda77cbf57a00838c3e097409e)

PR #141383, opened 2026-09-07
improve(ui): shrink the sidebar running ring on rows without an icon

steipete:
Related: #141098

## What Problem This Solves

Resolves a problem where sidebar sessions without an icon, avatar, or owner chip showed their running state as a large empty circle beside the label. #141098 moved the running ring onto each row's own glyph, and rows with nothing inside the ring inherited the full 24px glyph size, which reads as a missing icon rather than as progress.

## Why This Change Was Made

The shared glyph renderer now marks a content-less glyph as bare, and one CSS rule shrinks that ring to the 12px size the old trailing spinner had, centered in the same 20px lead slot so it stays aligned with neighboring avatars and unread dots. Rows with artwork keep the full ring around it. All bare-ring callers (top-level sessions, child rows, and the Codex/Claude catalog rows) share the renderer, so they change together.

## User Impact

Running sessions with no icon show a compact spinner in the icon column instead of an oversized empty ring. Rows with icons or avatars are unchanged.

## Evidence

Mocked Control UI dev server, Playwright measuring every `.session-glyph__ring` on the page: before the change every ring measured 24px; after it, bare rings measure 12px and rings around artwork still measure 24px. Captures below are fixture data from the mock scenario; the "before" is the same build with the new rule overridden back to the old inset, which is the only visual difference.

Before:

Before: full-size empty ring on icon-less rows

After:

After: compact 12px ring on icon-less rows

- Codex autoreview at P2 threshold: scoped-clean, no findings.
- `node scripts/check-changed.mjs`: lint, format, typecheck, dead-export scans passed.
- `chat-flow.sidebar-presentation.e2e.test.ts` and `session-management.sidebar.e2e.test.ts`: 15/15 passed.


steipete:
Maintainer note on the two before-merge items: the before/after captures were produced on the mocked Control UI dev server (fixture sessions only, no real data) and inspected before upload. The review environment cannot fetch GitHub user-attachments, which is why they could not be verified there. The measurements in the Evidence section were taken from the same Playwright run that produced the images: every bare ring measures 12px after the change, every ring around artwork still measures 24px.

steipete:
Merged via squash.

- Prepared head SHA: [8bff85d95c3e2d4714e0280d0ce839e746ac5392](https://github.com/openclaw/openclaw/pull/141383/commits/8bff85d95c3e2d4714e0280d0ce839e746ac5392)
- Landed commit: [179694d2f5d1c2aa7f96b1bce4c5ad3b891aa544](https://github.com/openclaw/openclaw/commit/179694d2f5d1c2aa7f96b1bce4c5ad3b891aa544)

PR #141327, opened 2026-09-07
feat(team-reports): adopt Carapace visual contract and GitHub avatars

steipete:
## What Problem This Solves

The Team Reports pages shipped with a generic, ad-hoc stylesheet that looked nothing like the rest of OpenClaw's surfaces, and people appeared as bare logins with no avatars, which makes a roster of 60 members slow to scan.

## Why This Change Was Made

The report, people, and person pages now follow Carapace, OpenClaw's shared visual contract (openclaw/carapace v0.6.2). Because the plugin serves self-contained HTML behind a nonce-only `style-src` policy and Carapace is a git-tag package rather than an npm dependency, the plugin vendors a faithful subset of its tokens (palette, semantic dark and light themes, status colors, spacing, radii, type scale, embed font stacks) with the source files named in the stylesheet header, and re-implements the primitives it uses under their Carapace class names: app surface toolbar, sections and eyebrows, stat cards, pills, status badges, banners, tables, resource lists, and sparkline-style chart lines. Dark is the default, light follows the OS preference or an explicit `data-theme`. Every roster member, other GitHub actor, people-table row, and person page gets a GitHub avatar from `avatars.githubusercontent.com` (already permitted by the page's `img-src`), rendered over CSS initials so a blocked image degrades without JavaScript; only syntactically valid GitHub logins produce an image URL.

## User Impact

Same routes, data, Markdown, JSON, and security headers; only the HTML and CSS changed. Reports are denser and scan faster, status is conveyed with Carapace status colors, and people are recognizable at a glance. No configuration options were added.

## Evidence

- `node scripts/run-vitest.mjs run extensions/team-reports`: 13 files, 204 tests passed (new cases: GitHub avatar for valid logins at each size, unsafe identities escaped with initials fallback, hostile display name never reaches the avatar URL, closed/partial status badge variants, dark and light token blocks with CSS-only initials).
- oxlint, oxfmt, and dead-export scans clean; both `tsgo` extension lanes report only pre-existing errors in `extensions/openshell` and `src/infra/install-package-dir.ts` from a stale local `@openclaw/fs-safe` build, none in this plugin; CI is the authoritative typecheck.
- Codex autoreview (branch vs `origin/main`, P0–P1): scoped-clean.
- Before/after screenshots from an isolated Gateway with public data only (GitHub team roster, one public Discord channel, no excerpts, no summaries) are attached below.

Overview, before (dark):

Overview before

Overview, after (dark):

Overview after

Day report, before (dark, top of page):

Day report before

[...]

PR #141384, opened 2026-09-07
feat(team-reports): match the maintainer report site layout

steipete:
## What Problem This Solves

The Team Reports plugin replaces the internal maintainer report site, but after the Carapace token pass (#141327) its pages still looked like a generic plugin surface: a toolbar, three stat cards, a line chart, and tables. Operators who use today's report site get a very different product in the Control UI tab: no activity dateline, no today/week/month cards with deltas and the open-day countdown, no history rows with trailing trends, no people timelines, and no per-person calendar.

## Why This Change Was Made

The plugin pages now reproduce the maintainer report site's composition with the plugin's own data: a brand banner with the Carapace crab artwork and a generation stamp, a full-history dateline, open-window and source-coverage notices, today/week/month quick cards with period-over-period deltas (open aggregates compare against the same number of elapsed prior days), People and Generation panels, and day/week/month history rows with trailing sparklines. Report pages get the site's metric strip with deltas and sparklines against prior periods, the activity mix with a composition bar and ranked member distribution, and the three-column member rows with a filter and the full retained item list. The people page shows 28-day activity strips ranked by activity and grouped into active, quiet, read-only, and archived; the person page shows the dual-axis 30-day chart, a monthly calendar linking into day reports, and active-day rows.

Design decisions and boundaries:

- The site's Carapace primitives (brand banner, summary strip, sparkline, delta, split bar, segmented nav, switch) are vendored into the plugin stylesheet under their Carapace class names, using the product theme's rounded geometry. No dependency is added; the crab hero is byte-identical to Carapace's MIT-licensed home artwork and ships with the plugin as a declared static asset.
- One inline script per page, allowed by a `script-src` nonce, adds the theme toggle, live relative times and the day countdown, history toggles, the member filter, and the quiet switch. Every page still works without JavaScript: all rows and members are in the HTML and timestamps have server-rendered text.
- The Control UI embeds the tab with `allow-scripts` only, so the frame has no storage. A theme choice therefore travels through in-tab navigation as a `#theme=` page fragment and persists per browser only where storage exists (a separate window); the docs say exactly that.
- Artwork and the plugin icon are served from a new allow-listed `assets` route under the same scope check as report pages, with a private one-day cache, resolved from the plugin root so source checkouts, the flattened `dist` bundle, and installed packages all work; `img-src` gains `'self'`.
- Completeness on the overview comes from the stored status: an open day generated before close shows `Intraday`, a stored partial entry whose window has closed shows `Incomplete`, closed entries show nothing. Overview organizations come from the displayed report or the active scheduler configuration, not from registration time.
- Period totals for datelines, cards, and history come from SQLite `json_extract` on the stored report documents (no schema change). A 400-row query of 200 KB reports measured about 75 ms, so no memo was added.
- Routes, Markdown and JSON exports, escaping, HEAD handling, and the operator scope check are unchanged.

## User Impact

The Reports tab now looks and reads like the maintainer report site operators already know, with the same dateline, cards, history, timelines, and calendar idioms, plus a light/dark toggle. Nothing changes for the CLI, Gateway methods, exports, or configuration.

## Evidence

- `node scripts/run-vitest.mjs run extensions/team-reports`: 15 files, 233 tests passed. New behavior tests cover the dateline threshold, UTC open-window guidance and countdown, the "to date" comparison for open aggregates, stored-status completeness badges, history toggles, metric deltas and in-progress labels, ranked distribution order, the full GitHub item list, bounded warning banners with hostile input escaped on every page, people grouping and activity buckets, calendar links and chart gaps, fragment/storage/OS theme resolution and link carrying, json_extract totals, `listPersonDaysSince`, scheduler health, current organizations, and the asset route (content types, caching, HEAD, unknown names, scope).
- oxlint, oxfmt, `git diff --check`, the dead-export scans, and the runtime-dependency contract test (which validates the `staticAssets` declaration) are clean locally. Both `tsgo` extension lanes pass locally after reconciling the worktree install, and CI typecheck is green on the head.
- Codex autoreview (P0–P1) on the initial change, the review-findings fix, and the asset fix: scoped-clean, no findings.
- Live proof on an isolated Gateway running the built `dist` plugin with public data only (openclaw org roster, one public Discord channel, no excerpts, no summaries): five closed days plus today's intraday snapshot backfilled through the CLI; both assets serve `200` from the packaged plugin (the first build exposed an ENOENT from module-relative asset paths, fixed by resolving from the plugin root and declaring the artwork as a static asset); CSP header as described; no route failures in the Gateway log.

Overview, before (current main, dark):

Overview before

Overview, after (dark, full page):

Overview after

Overview, after (light, top):

Overview after light

Day report, before (current main, dark, top of page):

Day report before

Day report, after (dark, top of page; deterministic summary because model summaries are disabled on the rig):

Day report after

People, after (dark, header; the full roster grid is omitted from this public description):

People after

Person page, before (current main, dark):

Person page before

Person page, after (dark, top):

Person page after


steipete:
Addressed the five findings in the follow-up commit on this branch (`fix(team-reports): address review findings on the site layout`), each with a behavior test:

- **Theme persistence in the sandboxed tab:** the head script now resolves the theme from the page fragment (`#theme=light|dark`) first, then browser storage when it is readable, then the OS preference. Toggling rewrites the current fragment and every same-origin report link so the choice follows in-tab navigation inside the `allow-scripts` frame; storage is still used where available (separate window). The docs now state exactly that. Verified in Chromium inside an `allow-scripts` sandbox with storage blocked.
- **GitHub items beyond the first three:** member cards keep the three-item preview and again expose the complete retained list in an expandable "N GitHub items" details block.
- **Incomplete snapshots after the window closes:** overview cards and history rows derive completeness from the stored status. An open day generated before close shows `Intraday`; any stored partial entry whose window has closed shows an `Incomplete` warning badge; closed entries show none.
- **Person-chart axis:** the axis rule now sets `fill: none`.
- **Overview organizations:** the HTTP handler resolves organizations from the active scheduler configuration (falling back to the initial config only before the scheduler exists), and the overview prefers the displayed report's organizations.

On the two remaining items: the "stored data model" flag comes from the localStorage read in the inline script and the in-memory asset cache in the HTTP handler; there is no persisted-state or SQLite change in this PR (period totals are read with `json_extract` from the existing report documents). Sanitized before/after browser captures from an isolated Gateway with public data are now in the PR description.

steipete:
Merged via squash.

- Prepared head SHA: [5a0826be2035948d5f6f17f6fc32727ed717d3b6](https://github.com/openclaw/openclaw/pull/141384/commits/5a0826be2035948d5f6f17f6fc32727ed717d3b6)
- Landed commit: [b7528507af5a4ea04b5165ac64d30f504e898f19](https://github.com/openclaw/openclaw/commit/b7528507af5a4ea04b5165ac64d30f504e898f19)

PR #141390, opened 2026-09-07
fix(lmstudio): route chat to the prepared model instance

### LINKED ISSUE #141357: LM Studio inference can select a smaller loaded instance than its context budget
LM Studio model discovery can report a larger context than the model instance actually selected for inference.

Reproduced on main `ee748b6c86d5` with headless LM Studio 0.0.23-1 and `Qwen3.5-0.8B-Q4_K_M`. Load the same model twice: a canonical instance with 4,096 context tokens and a separately named instance with 16,384. Discovery correctly sees the larger instance and records `contextTokens: 16384`, but preload returns the canonical model key, so the chat request addresses the smaller instance.

The production preload/stream wrapper rejects a 5,777-token synthetic prompt with HTTP 400 and reports only 4,096 available context tokens. Addressing the identical prompt directly to the larger instance succeeds with `OK` and usage of 5,777 input / 1 output token.

Expected: inference should address the loaded instance whose capacity was selected, or the instance returned by a new load request. Keep the stable canonical model key for configuration, transcript identity and embedding cache identity; loaded instance IDs are transient request routing data.

The before-fix proof used an isolated Testbox with real native discovery/load and completion APIs. All test daemon processes were stopped after the run. A bounded plugin-owned repair and after-fix live proof are being prepared.

steipete:
Closes #141357.

## What Problem This Solves

When the same LM Studio model has multiple loaded instances, OpenClaw can budget against the larger instance but send chat to the smaller one. A real 5,777-token prompt failed with HTTP 400 against the canonical 4,096-token instance even though discovery reported an available 16,384-token instance.

## Why This Change Was Made

Select a loaded instance with enough context, or use the instance ID returned by a new native load. Apply that ID only to the effective completion payload, including replacement payloads from asynchronous callbacks. Keep the canonical model key for configuration, conversation identity and the existing public loading helper used by embeddings.

## User Impact

With preload enabled, chat uses the capacity selected during discovery instead of failing against another smaller instance. Existing model references and preload opt-out behavior remain intact.

## Evidence

- Real before-fix Testbox proof with headless LM Studio 0.0.23-1 and Qwen3.5-0.8B Q4_K_M: 4,096 and 16,384-context instances loaded simultaneously; the production wrapper failed against 4,096, while the identical prompt addressed directly to the larger instance succeeded with `OK` (5,777 input / 1 output token).
- New wrapper regression failed on the original source for both an already loaded suitable instance and a new load. A separate regression reproduced an independently constructed replacement callback discarding the selected ID; the finalizer now preserves it.
- 100 focused model, stream and embedding tests passed. The final fixture-only type narrowing passed its three routing cases. Final independent review found no actionable P0–P2 findings.
- After-fix live Testbox proof passed on the exact changed production-file hashes: the wrapper routed the 5,777-token prompt to the existing 16K instance and returned `OK`. After that instance was evicted, it loaded at 16K and used the new `instance_id` returned by the native API; the second completion also returned `OK` (5,777 input / 1 output token each). The callback deliberately returned a fresh payload containing the canonical model key; final request routing still selected the correct instance. Conversation model identity stayed canonical. All 61 observed task processes stopped.
- Changed-file validation completed: production/test typechecks and unused-export scans passed; after correcting a fixture-only lint issue, targeted lint, database/media/sidecar/import-cycle guards, six-file formatting and diff checks all passed. The live snapshot differs from this commit only by that test-fixture string narrowing; both changed production-file hashes match exactly.

### DIFF
--- a/docs/providers/lmstudio.md
+++ b/docs/providers/lmstudio.md
@@ -146,6 +146,13 @@ still has `off`/`on` reasoning maps.
 }
 ```
 
+### Model instances and context
+
+With preload enabled, OpenClaw routes chat requests to a loaded instance with
+enough context for the selected model budget. A newly loaded instance is addressed by the
+identifier returned by LM Studio. Your configured model reference and conversation model identity
+keep the canonical model key.
+
 ### Disabling preload
 
 LM Studio supports just-in-time (JIT) model loading, loading models on first request. OpenClaw
--- a/extensions/lmstudio/src/models.fetch.ts
+++ b/extensions/lmstudio/src/models.fetch.ts
@@ -45,6 +45,7 @@ function redactLmstudioLoadError(value: string, headers: Record<string, string> 
 
 type LmstudioLoadResponse = {
   status?: string;
+  instance_id?: string;
 };
 
 type LmstudioResolvedModelKeyError = {
@@ -233,8 +234,7 @@ export async function discoverLmstudioModels(
     .filter((entry): entry is ModelDefinitionConfig => entry !== null);
 }
 
-/** Ensures a model is loaded in LM Studio before first real inference/embedding call. */
-export async function ensureLmstudioModelLoaded(params: {
+type LmstudioModelLoadParams = {
   baseUrl?: string;
   apiKey?: string;
   headers?: Record<string, string>;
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [66c3d5bf4db1cc607fc4e2fa6a1f6d88f86d9675](https://github.com/openclaw/openclaw/pull/141390/commits/66c3d5bf4db1cc607fc4e2fa6a1f6d88f86d9675)
- Landed commit: [2ddc37e61d2011b7abb03dcd13444ec62c02808b](https://github.com/openclaw/openclaw/commit/2ddc37e61d2011b7abb03dcd13444ec62c02808b)

PR #141393, opened 2026-09-07
fix(gateway): trim wizard next/cancel/status sessionId before Map lookup

nocodet888-arch:
## What Problem This Solves

`wizard.next` / `wizard.cancel` / `wizard.status` forwarded the raw `sessionId` into an exact Map lookup. Clipboard-padded ids missed a live setup session. Unconditional trim-first would also collide with a caller-supplied key such as `" login "` when `"login"` is a different live session.

## Why This Change Was Made

Keep the caller-supplied sessionId. Look up the exact Map key first; trim only when that exact key is missing. If the exact session exists but `canAccessWizardSession` fails, do not fall back. Purge/cancel use the selected key. Proof is a real Gateway WebSocket client plus handler coverage for colliding keys.

## User Impact

**Before:** Pasting a wizard session id with surrounding whitespace returned not-found and left the setup session running.

**After:** Padded status/cancel hit the generated live session. A caller-supplied padded key stays distinct from its trimmed neighbor; an unauthorized exact session does not leak the trimmed one.

## Evidence

- Changed: `src/gateway/server-methods/wizard.ts`, `src/gateway/server-methods/wizard.session-id-trim.test.ts`, `src/gateway/gateway-wizard-cancel.e2e.test.ts`
- Production path: `connectGatewayClient` → `wizard.start` → padded `wizard.status` / `wizard.cancel`
- Compatibility: generated session ids stay unpadded; caller-supplied Map keys remain exact; only missing exact keys use trim fallback

## Real behavior proof

- **Behavior or issue addressed:** A running Gateway accepts clipboard-padded wizard session ids for status and cancel without collapsing distinct caller-supplied keys.
- **Canonical reachability path:** Real `GatewayClient` token auth → `wizard.start` → padded `wizard.status` (`running`) → padded `wizard.cancel` (`cancelled`).
- **Boundary crossed:** Real Gateway WebSocket RPC + live wizard runner (not a mocked handler callback).
- **Shared helper / provider constraint check:** `findWizardSessionOrRespond` is exact-first; unauthorized exact sessions return `WIZARD_NOT_FOUND` without trim fallback. Supplemental handler tests cover `" login "` vs `"login"` and owner mismatch.
- **Real environment tested:** Local trusted source checkout; Vitest e2e project; `startGatewayServer` + real `connectGatewayClient` on loopback.

```bash
OPENCLAW_E2E_SKIP_BUILD=1 node scripts/run-vitest.mjs src/gateway/gateway-wizard-cancel.e2e.test.ts -t "padded sessionId" --reporter=verbose
```

- **Evidence after fix:**

```text
 src/gateway/gateway-wizard-cancel.e2e.test.ts > gateway wizard cancellation lifecycle > cancels a live wizard when wizard.cancel receives a padded sessionId 4373ms

 Test Files  1 passed (1)
      Tests  1 passed | 3 skipped (4)
```

- **Observed result after fix:** Padded status returns `running`; padded cancel returns `cancelled`; colliding exact/trimmed keys stay distinct.
- **What was not tested:** `wizard.next` answers through a real client.

- [x] AI-assisted (Cursor)

### DIFF
--- a/src/gateway/gateway-wizard-cancel.e2e.test.ts
+++ b/src/gateway/gateway-wizard-cancel.e2e.test.ts
@@ -128,6 +128,43 @@ async function withWizardGateway(
 
 describe("gateway wizard cancellation lifecycle", () => {
   it(
+    "cancels a live wizard when wizard.cancel receives a padded sessionId",
+    { timeout: GATEWAY_E2E_TIMEOUT_MS },
+    async () => {
+      const runnerSettled = createDeferred();
+      await withWizardGateway(
+        async (_opts, _runtime, prompter) => {
+          prompter.progress("working");
+          await runnerSettled.promise;
+        },
+        () => {
+          runnerSettled.resolve();
+        },
+        async ({ connect }) => {
+          const client = await connect();
+          const start = await client.request<WizardStartResult>("wizard.start", { mode: "local" });
+          expect(start).toMatchObject({ done: false, status: "running" });
+          expect(start.sessionId).toBeTruthy();
+
+          const paddedSessionId = ` ${start.sessionId} `;
+          const status = await client.request<{ status: string }>("wizard.status", {
+            sessionId: paddedSessionId,
+          });
+          expect(status).toMatchObject({ status: "running" });
+
+          await expect(
+            client.request("wizard.cancel", { sessionId: paddedSessionId }),
+          ).resolves.toMatchObject({ status: "cancelled" });
+          console.log(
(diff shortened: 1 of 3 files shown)


nocodet888-arch:
@clawsweeper re-review

Rebased onto current main (kept canAccessWizardSession + closeInput on the trimmed id). Added real Gateway WebSocket proof: wizard.start, then padded wizard.status / wizard.cancel.

```text
✓ cancels a live wizard when wizard.cancel receives a padded sessionId 14510ms
Test Files  1 passed (1)
Tests  1 passed | 3 skipped (4)
```

nocodet888-arch:
@clawsweeper re-review

PR #141395, opened 2026-09-07
fix(cli): explain failed local secret resolution

obviyus:
When Gateway secret resolution is unavailable and local resolution also fails, strict commands only advised starting the Gateway. The final error now identifies the local failure too and advises checking configured secret sources and Gateway access. Strict rejection, the JSON error shape and the existing Gateway context stay intact; no local error, reference or secret value is added to output.

Validation:

- The new regression assertion fails on main `4b1f467dff76`; five focused strict, local-success and incomplete-snapshot controls pass with the fix.
- Real source CLI `models status --json --check`: a missing secret file still exits 1 with both resolution failures explained; a valid local file still exits 0 without a Gateway. Both outputs preserve secret-value masking.
- Pinned formatting, syntax-only lint, assertion-safety and max-lines ratchets pass. Full type checks remain CI-owned.

Production changes one message with no net line growth. AI-assisted.

### DIFF
--- a/src/cli/command-secret-gateway.test.ts
+++ b/src/cli/command-secret-gateway.test.ts
@@ -688,17 +688,21 @@ describe("resolveCommandSecretRefsViaGateway", () => {
     const envKey = "TALK_API_KEY_FAILFAST";
     await withEnvValue(envKey, undefined, async () => {
       callGateway.mockRejectedValueOnce(new Error("gateway closed"));
-      await expect(
-        resolveCommandSecretRefsViaGateway({
-          config: buildTalkTestProviderConfig({
-            source: "env",
-            provider: "default",
-            id: envKey,
-          }),
-          commandName: "memory status",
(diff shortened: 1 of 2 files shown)


obviyus:
Verified landing: `d4f89aede2689c8d9eb90645b758c80a4d8cbc52`.

**Proof:** Real source CLI `models status --json --check` keeps exit 1 for a missing secret file and exit 0 for an available local file without a Gateway. The failure now explains both resolution routes without adding local error details or secret values.

| Before | After |
| --- | --- |
| Failed local resolution ended with Gateway-only advice. | The error also names local resolution failure and advises checking configured secret sources and Gateway access. |

Five focused controls passed. Exact-head local ClawSweeper passed at tier B with no findings. [Required CI passed](https://github.com/openclaw/openclaw/actions/runs/34147837131/job/101825504393).

AI-assisted.

PR #141396, opened 2026-09-07
fix: show completed fallback models in session diagnostics

obviyus:
## What Problem This Solves

Fixes an issue where chat model controls could omit the model that answered after the selected primary failed. The selected model stayed correct, but the existing active-model fields were empty even after a visible, successful fallback.

## Why This Change Was Made

Fallback accounting deliberately preserves the session's selection fields. Session diagnostics now read the answering model from the latest canonical transcript message for the completed session and exact run, then apply the existing fallback-notice and alias checks. The read uses the existing one-message, 256 KiB materialization bound and requires a visible terminal assistant message.

## User Impact

Users can distinguish the selected primary from the fallback that answered. The next turn still starts from the selected primary, and successful primary recovery clears the active fallback. Selection, thinking defaults, model metadata, authentication, and persistent-store semantics are unchanged.

## Evidence

- Current-main regression reproduced the missing active fields through real accounting, transcript persistence, and lightweight session-row projection, starting without a stored runtime model pair.
- Four focused tests pass: fresh and previous-primary fallback sequences plus existing active/stale projection controls. The sequences cover changed selection, other runs/sessions, running state, hidden output, and primary recovery while an old notice remains.
- An authenticated source-mode Gateway with a synthetic loopback provider made exactly three model requests: primary 404, fallback 200, primary recovery 200. Public settled receipts, history, and session lists agreed on the answering model. Selection, thinking defaults, and model metadata stayed unchanged; recovery cleared both active fields.
- Scoped formatting, lint, and diff checks pass. Typechecking and required CI run on the PR head.

Production changes are +41/-1 lines. The additional reader preserves exact run and visibility checks without changing stored selection state. A focused regression file keeps the accounting fixture and failure/recovery sequence together. Diagnostics remain unavailable when the bounded latest-message read cannot establish the answering model.

### DIFF
--- a/src/auto-reply/reply/agent-runner-result-accounting.fallback.test.ts
+++ b/src/auto-reply/reply/agent-runner-result-accounting.fallback.test.ts
@@ -1,0 +1,226 @@
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { afterAll, beforeAll, expect, it } from "vitest";
+import {
+  loadSessionEntry,
+  persistSessionTranscriptTurn,
+  replaceSessionEntry,
+} from "../../config/sessions/session-accessor.js";
+import { drainSessionStoreWriterQueuesForTest } from "../../config/sessions/store-writer-state.js";
+import type { InternalSessionEntry } from "../../config/sessions/types.js";
+import type { OpenClawConfig } from "../../config/types.openclaw.js";
+import { buildGatewaySessionRow } from "../../gateway/session-utils-row.js";
+import { disposeOpenClawAgentDatabaseByPath } from "../../state/openclaw-agent-db.js";
+import { accountAgentTurn } from "./agent-runner-result-accounting.js";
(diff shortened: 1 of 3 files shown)


obviyus:
**Proof:** An authenticated source-mode Gateway completed primary failure, fallback success, and primary recovery in the same session.

- **Bug:** Session diagnostics omitted the answering fallback even after a visible, successful reply.
- **Fix:** Read the completed run's model from its bounded canonical transcript while preserving the selected primary and existing notice checks.
- **Validation:** Four focused cases passed, along with scoped formatting and lint. Fresh local ClawSweeper review found no blocking correctness or security issues and accepted the retained product proof.

| Before | After |
| --- | --- |
| Selected primary remained correct, but both active-model fields were absent after fallback. | Diagnostics show the fallback that answered. Primary recovery clears both active fields without changing selection. |

Source-mode proof used a synthetic loopback provider. Required CI: https://github.com/openclaw/openclaw/actions/runs/34148478767

PR #141402, opened 2026-09-07
fix(models): use resolved runtime headers for catalog hooks

obviyus:
## What Problem This Solves

Fixes an issue where valid provider-header secret references were reported unresolved during catalog discovery even after runtime secret resolution.

## Why This Change Was Made

The shared catalog context gives the active provider hook its matching runtime headers and request settings. Source credential selection and other providers’ configuration remain separate; generated planning preserves authored secret markers.

## User Impact

Catalog discovery can use resolved headers without replacing authored references or changing credential ordering, validation or endpoint policy.

## Evidence

The actual planning/discovery callback reproduced the failure. The resolved-runtime case and genuine-unresolved control passed after the fix, along with 95 authentication/provenance tests, 11 persistence tests and four header controls. A source command made authenticated requests to a synthetic service and listed the returned model while preserving references. Earlier Gateway attempts stopped at loader setup; no external provider account or inference was tested.

### DIFF
--- a/src/agents/models-config.providers.catalog-context.ts
+++ b/src/agents/models-config.providers.catalog-context.ts
@@ -1,4 +1,7 @@
-import { normalizeProviderId } from "@openclaw/model-catalog-core/provider-id";
+import {
+  findNormalizedProviderValue,
+  normalizeProviderId,
+} from "@openclaw/model-catalog-core/provider-id";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import type {
   ProviderCatalogOutcome,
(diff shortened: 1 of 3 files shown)


obviyus:
**Proof:** The real source CLI `models list --provider github-copilot --all --json` completed against a synthetic HTTPS account/catalog fixture. It sent the resolved nested integration header, listed the returned model, and kept the authored secret references. Targeted auth, persistence, unresolved-input, and header-precedence controls passed. Exact-head local ClawSweeper review was clean.

- **Bug:** Valid header references reached catalog hooks unresolved after runtime materialization.
- **Fix:** The shared catalog context gives each hook its matching runtime request settings while preserving source credential selection and saved references.

| Before | After |
| --- | --- |
| A valid integration header reference caused catalog discovery to fail as unresolved. | The public CLI completes authenticated catalog discovery with the resolved header. |

The scoped CLI proof uses synthetic data, not a real provider account or inference. Failed source Gateway setup attempts remain failed evidence; their separate loader issue is tracked outside this repair.

PR #141081, opened 2026-09-07
fix: correct missing and misdirected Control UI notifications

steipete:
Closes #141035

## What Problem This Solves

Fixes an issue where Control UI users miss automation failures beyond the first 50 jobs, receive premature completion or duplicate failure alerts, and open an unrelated chat when selecting a notification. First messages sent from New Session also skipped notification permission onboarding.

## Why This Change Was Made

The Inbox now publishes a complete automation inventory, retaining its prior state while later pages load. Browser alerts use prepared session scope or the existing question/automation routes, preserve the originating Gateway, wait for yielded parent runs to finish, and classify cron tracking records only through scheduled-failure delivery. New Session uses the existing notification prompt owner within the original input gesture.

## User Impact

Users can find failures across the full automation inventory and open the question, conversation, or automation run needing attention. Normal and background New Session starts participate in one-time notification onboarding. Existing category preferences, intentional unsubscribes, profile/session authorization, and temporary mention retention remain intact.

## Evidence

- Original regressions failed before repair: missing later-page automation, yielded completion, four notification destinations, duplicate cron categories, and web/native foreground/background onboarding.
- Follow-up regression proof protects complete-page publication, prepared-scope navigation, and cancellation of internal snapshot recovery after hide/disposal. The two recovery cases failed before the final guard; all 239 store/cron tests passed afterward. Focused Gateway/approval and UI suites passed; the final UI follow-up passed 57 tests.
- Four Chromium E2E cases passed, including native-bridge permission requests with live click/keyboard user activation.
- The initial CI run passed every selected lane except six existing native Mermaid browser tests. The clipboard suite leaked its module mock into the native suite on the same non-isolated worker; that baseline defect was repaired separately in merged PR #141092. The next CI attempt passed all selected test lanes but hit an existing Telegram test-size lint failure. That baseline is repaired separately in #141139 and #141148, now included here. Both notification patches are unchanged across the refreshes; the latest refresh passed 91 focused Gateway/UI tests.
- The latest full CI attempt cleared lint and UI but failed two CLI/tooling shards; GitHub log-storage downloads were inaccessible from two networks, so the exact failure diagnostics could not be recovered. The locally reconstructed groups passed 1,735 tests with nine existing platform/fixture skips. The unchanged-head retry passed all required CI ([run34115498598, attempt2](https://github.com/openclaw/openclaw/actions/runs/34115498598/attempts/2)). The original failures remain recorded, with cause unverified.
[...]

PR #141404, opened 2026-09-07
fix: keep the Mentions Inbox across Gateway restarts

steipete:
## What Problem This Solves

Fixes an issue where users lose still-relevant Mentions Inbox entries and acknowledgement/replay state when a Gateway restarts, including during upgrades. Retained mentions and dismissals now survive restarts with their original identifiers and expiry times.

## Why This Change Was Made

Use typed records in the existing `config_machine_state` table: one bounded record per committed source, plus an atomic revision/arrival sequence record. In-memory recipient indexes remain disposable projections. Each mutation rereads the current revision inside the existing synchronous state transaction, writes only changed records, and publishes after commit. Failed writes return unavailable or withhold new delivery; they do not publish uncommitted state.

Maintainer-approved scope is persistence without a SQLite schema change. There are no new tables, columns, indexes, triggers, schema/database versions, configuration options, or dependencies. The existing machine-state owner already supports typed runtime snapshots and notification deduplication. Retention remains seven days, 100 entries per profile, 10,000 entries globally, and 10,000 consumed sources. Empty sources retain only replay receipts; expiry resumes at startup and retries transient storage failures while its owner remains active. Incognito sources never enter durable storage. Older compatible builds that lack this feature ignore its keys; their in-memory changes are not backfilled.

## User Impact

Restarting or upgrading no longer clears the retained Inbox or revives dismissed/evicted mentions. Loading saved state does not send old browser alerts or scan old transcripts. Current profile, role, session visibility, and session-incarnation checks still control every read and push. The UI and documentation explain the retained Inbox and preserve the existing documentation anchor.

Follow-up to #141081. This retains the existing best-effort boundary after transcript commit; it does not add a transactional chat-delivery outbox.

## Evidence

- Six initial regressions failed on the original implementation. Additional red/green checks cover startup expiry without readers, incognito storage exclusion, and retry/cancellation after failed expiry writes.
- The broader Gateway proof passed 105 tests and hit one cold kernel-start timeout. That exact test subsequently passed unchanged in a focused run (36.42 seconds total, including transforms), with no timeout increase. The Inbox-containing startup phase measured 2.8 ms. Both profile-merge recipient orders and restart persistence pass in focused proof.
- `sqlite_schema` and `PRAGMA user_version` remain byte-for-byte/equivalently unchanged after admission, dismissal, and reopening. Cross-instance tests protect against stale writes resurrecting dismissals or losing new inputs; temporary abort triggers prove write rollback without push or invalidation publication.
- Real source Gateway proof used two synthetic authenticated profiles and a deterministic local model. Two chat sends created mentions; one was dismissed. The owned Gateway was killed with SIGKILL and replaced with a new PID/boot ID over the same state. The retained item kept its exact ID and expiry, the dismissal remained effective, and a fresh mention and dismissal worked after recovery. The fixture was cleaned up afterward.
- Control UI production build, core/UI/test typechecks, localization verification, changed-file lint and `node scripts/check-changed.mjs` passed. Required CI passed on the published head in [run34148249455](https://github.com/openclaw/openclaw/actions/runs/34148249455), attempt1.
- Independent Codex review raised a profile-merge ordering concern. The existing collision branch explicitly makes a consumed receipt win in either order; both orders now have restart regression proof. No production change was needed for that finding.
- Browser automation repeatedly lost its debugger attachment, so no new GUI screenshots or native-app delivery proof are claimed. Live evidence above uses the actual Gateway chat/Inbox RPC flow; restored push suppression is covered at its delivery callback boundary.

### DIFF
--- a/docs/concepts/multi-user.md
+++ b/docs/concepts/multi-user.md
@@ -135,13 +135,15 @@ The picker includes known Gateway profiles eligible to read the session, includi
 
 Mentions work for ordinary messages, queued or steered input, and the first message of a new session, including a remotely placed session. They are unavailable in incognito, Goal, catalog, suggestion-only, command-send, or terminal-launch modes. If selected mentions remain after switching to an unsupported mode, the composer blocks the send and asks you to remove them or return to a normal chat. It does not silently discard selected recipients.
 
-## Temporary mentions Inbox
+<a id="temporary-mentions-inbox" />
+
+## Mentions Inbox
 
 Open **Inbox → Mentions** to see messages addressed to your signed-in profile across accessible agents. Opening a mention opens its session without dismissing it. Select **Dismiss** to remove the entry from your Inbox; that change follows the same profile across connected browsers, without deleting the chat message.
 
-Mentions are kept in Gateway memory for **up to seven days**, with at most **100 entries per profile**. Older entries can be evicted earlier by capacity limits. Refreshing or reconnecting to the same running Gateway reloads its current Inbox. A Gateway restart, including one during an upgrade, clears it. Old transcript messages do not repopulate the Inbox after restart. This is a temporary attention list, not a durable notification archive or delivery guarantee.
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [965cc1560d63bafb84c08a74ef67a748361e3b85](https://github.com/openclaw/openclaw/pull/141404/commits/965cc1560d63bafb84c08a74ef67a748361e3b85)
- Landed commit: [8184a1107dfd26c2a7003112cdafbad1571d182a](https://github.com/openclaw/openclaw/commit/8184a1107dfd26c2a7003112cdafbad1571d182a)

PR #141405, opened 2026-09-07
fix(ui): selected owner overlaps the Specific owner label in the sessions filter menu

steipete:
## What Problem This Solves

Fixes an issue where users who picked a specific owner in the sessions sidebar **Filter & sort** menu would see that owner's avatar and name drawn on top of the **Specific owner** row label, so the label read as "Specific ow" with the avatar covering the rest.

## Why This Change Was Made

The selected-owner rail sits in Web Awesome's `details` column, which is fixed-size and end-justified. The rail inherited `flex: 0 0 auto` from the keycap shortcut idiom and refused to shrink, so any name longer than the column spilled leftward across the label. The label part is a plain block in the shadow row, so the label's own ellipsis rules never applied either.

The row now gives the label part its own width (capped so it can still truncate), lets the details part absorb the rest of the row, and lets the rail shrink into it so the owner name truncates before the label does. The details part is pulled back by Web Awesome's forced 2em slotted margin so the row keeps its normal 8px gutter instead of stacking a ~30px one. Desktop menu only; the compact/mobile variant already lays out its details inside the label part and was unaffected.

## User Impact

The **Specific owner** row stays fully readable with the selected owner's avatar and truncated name to its right. Nothing else in the menu moves; sibling row labels keep the same left edge and the chevron keeps its spacing.

## Evidence

Mock Gateway, 220px desktop filter menu, owner "Peter Steinberger" selected.

| Before | After |
| --- | --- |
| Specific owner row with avatar and name overlapping the label | Specific owner row with label intact and truncated owner name |

Measured on the row before the fix: the label text ran 17.7px underneath the selection rail. After: the rail starts 6px right of the label, the name truncates, and the chevron gap is unchanged.

- `ui/src/e2e/session-owner-filter-keyboard.e2e.test.ts` now selects a long owner name ("Ada Lovelace Byron") and asserts the row label is not clipped and its text bounds end before the selection rail begins. The new assertion fails on the original CSS (overlap 17.6px) and passes with the fix. The existing label-alignment, trailing-gap, large-roster, and RTL assertions still pass.
- `node scripts/check-changed.mjs`: green (typecheck, UI stylelint, i18n verify, knip).
- Codex autoreview (P0–P2): scoped-clean.

### DIFF
--- a/ui/src/e2e/session-owner-filter-keyboard.e2e.test.ts
+++ b/ui/src/e2e/session-owner-filter-keyboard.e2e.test.ts
@@ -32,7 +32,7 @@ suite.define(() => {
             sessionRow("agent:main:bob", "Bob operations", 1),
           ]),
           owners: [
-            { type: "human", id: "profile-ada", label: "Ada" },
+            { type: "human", id: "profile-ada", label: "Ada Lovelace Byron" },
             { type: "human", id: "profile-bob", label: "Bob" },
             { type: "human", id: "profile-carol", label: "Carol" },
             { type: "human", id: "profile-dave", label: "Dave" },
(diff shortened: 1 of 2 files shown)


steipete:
Maintainer note on the remaining merge-risk item (visual inspection and the Web Awesome shadow contract):

The before/after captures in the PR body were inspected locally before upload; they come from the mocked Control UI e2e harness with synthetic data only. Measured on the same row in that harness (220px menu, owner "Peter Steinberger" selected):

- Before: label text ended at x=316.7 while the details part content started at x=299, so the label ran 17.7px under the avatar. The slotted rail was 94px wide inside a 66px details column.
- After: label 231–316.7 unchanged, avatar starts at 322.7, name truncated to 48px, chevron gap unchanged at 6.5px, sibling label left edges unchanged.

The Web Awesome 3.12 `dropdown-item` shadow rules the fix works against (from the installed package):

```css
#label { flex: 1 1 auto; min-width: 0; }
#details { display: flex; flex: 0 0 auto; align-items: center; justify-content: end; }
#details ::slotted(*) { margin-inline-start: 2em !important; }
:host(:state(has-submenu)[submenu-adjacent]) #details { padding-inline-end: 1.75em; }
```

The light DOM cannot outrank the `!important` slotted margin, which is why the details part is pulled back by `-2em` rather than the rail's own margin being changed. The new e2e assertion fails on the pre-fix stylesheet with overlap 17.6px and passes with the fix; the full owner-filter e2e file passed locally.

steipete:
Merged via squash.

- Prepared head SHA: [bfbba50b1c1006d07e709749e939c6708af6eee8](https://github.com/openclaw/openclaw/pull/141405/commits/bfbba50b1c1006d07e709749e939c6708af6eee8)
- Landed commit: [e3376f9a5aad9722736e1134ec5ca0dd22d5bf31](https://github.com/openclaw/openclaw/commit/e3376f9a5aad9722736e1134ec5ca0dd22d5bf31)

Issue #139105, opened 2026-09-05
[Bug]: In-place upgrade 2026.7.1-2 → 2026.8.2 silently drops/rewrites config the new runtime cannot use (Anthropic auth 401, memorySearch, Gmail hook 400, workshop jobs) — no warning from doctor/validate


jerabinovich:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

After an in-place upgrade from 2026.7.1-2 to 2026.8.2 the migration completed without errors, yet within 16 minutes every Anthropic request returned HTTP 401, memory search silently switched from the configured Ollama embedder to the OpenAI default, every Gmail webhook push got HTTP 400, and 10 model-backed `skill-collection-review-*` jobs were created — because 7.x config (`${VAR}` profile values, `agents.defaults.memorySearch`, hook mappings without `agentId`, default-agent marker) was dropped or reinterpreted instead of migrated, and neither the startup log, `openclaw doctor` nor `config validate` reported it.

### Steps to reproduce

1. On 2026.7.1-2, run a multi-agent config (15 agents in `agents.list`) with: `ANTHROPIC_API_KEY` in the systemd `EnvironmentFile` and no `apiKey` in `models.providers.anthropic`; an OpenAI auth profile created with `openclaw models auth paste-api-key` using the `${OPENAI_API_KEY}` syntax; `agents.defaults.memorySearch = {provider: "ollama", model: "nomic-embed-text:latest", sources: ["memory","sessions"], extraPaths: [...]}`; `hooks.mappings = [{match:{path:"gmail"}, action:"agent", sessionKey:"hook:ingress"}]` (no `agentId`) fed by `gog gmail watch serve`; no `skills.workshop` block.
2. `npm i -g openclaw@2026.8.2`, restart the gateway, let the state/agent DB migrations run (schema v15 / v19).
3. Send any Anthropic turn; wait for the next Gmail push; run `openclaw memory status`; run `openclaw cron list`.
4. Observe: `[agent/embedded] ... HTTP 401 invalid x-api-key` then `decision=rotate_profile`; on the Claude CLI backend `cli terminal failure ... error=Invalid bearer token`; `[memory] sync failed ... embedding provider "openai" is configured but unavailable`; `[gmail-watcher] [gog] watch: hook failed: hook status 400`; 10 new `skill-collection-review-<agent>` jobs, one failing with `Shared workspace agents use different collection-review identities`.
[...]

Issue #138892, opened 2026-09-05
doctor --fix --non-interactive applies a varying subset of repairs per run; some startup-blocking migrations have no working non-interactive path


TreyLawrence:
Found while migrating hosted 2026.7.1 state to 2026.8.1 (both official images, `OPENCLAW_STATE_DIR` on a persistent volume). The 2026.8.1 gateway hard-refuses to start on 2026.7.1 state (as designed — `OpenClawStateDatabaseSchemaMigrationRequiredError`, legacy session stores, legacy device identity, legacy exec approvals), and each refusal says *run `openclaw doctor --fix`*. Automating exactly that surfaced three issues:

**1. `doctor --fix --non-interactive` covers a varying prefix of its repair list per invocation.** Across repeated runs on identical state we observed: device identity migrated on some runs but not others; the config-shape migration (`agents.list` → `agents.entries` + `ownership: explicit`) applied in some environments but not others with the same input config. It looks like a mid-list repair failure (or an advisory-check failure) ends the pass early, so which migrations land is a per-run lottery. Running the identical command 2–3 times converges — the repairs themselves are idempotent — but nothing documents that a single pass isn't complete.

**2. Exit code conflates advisory findings with repair failure.** In a headless environment some advisory check always fails (no gateway listening yet, no memory system installed), so `doctor --fix` exits non-zero even when every migration succeeded. There's no way to distinguish 'migrations done, advisories failed' from 'migration failed' without parsing the human-oriented output.

**3. The exec-approvals refusal has circular fix advice.** A legacy `exec-approvals.json` (v1, socket+token, empty `defaults`/`agents`) makes plugin activation fail with *Run "openclaw doctor --fix" against the same state/config* — but `doctor --fix --non-interactive` does not migrate or remove that file; it prints the same advice back. Reproduced against `ghcr.io/openclaw/openclaw:2026.8.1@sha256:e7849cb6c1ef1ead39ab4be7d85edb2df89611f486e283284c7cf35ce39a20d4`.

Also worth noting: the session-store import (`doctor --session-sqlite import`) refuses a multi-agent roster without `--session-sqlite-all-agents` or an explicit agent, which is reasonable but not mentioned by the gateway's *run doctor --fix* refusal message, and `--fix` alone never runs it.

[...]

Issue #134896, opened 2026-09-01
[Bug] 2026.8.1 update: 5-blocker gateway restart cascade + doctor --fix self-referential failure on legacy workspace state


WoodyKim554:
## Summary

Updating 2026.7.1-2 → 2026.8.1 caused a cascade of gateway-restart failures — 5 sequential blockers plus a separate `doctor --fix` bug — that required manual dist-source inspection to fully resolve. Filing as first-hand user feedback from a real update, not a synthetic repro.

## Part 1 — Gateway restart: 5 sequential blockers, each hidden until the previous was fixed

Each fix required its own `gateway restart` cycle to discover the next blocker:

1. **Codex plugin version mismatch** (2026.7.1-1 plugin vs 2026.8.1 gateway)
   - Error: `Failed detecting Codex app-server thread bindings: AgentSelectionRequiredError`
   - Fix: `openclaw plugins update codex --accept-capabilities`
2. **duckduckgo plugin capability consent not granted** — blocked gateway boot entirely
   - Error: `Plugin duckduckgo requires capability consent`
   - Fix: `openclaw plugins install @openclaw/duckduckgo-plugin --accept-capabilities`
3. **Legacy session store needing migration**, discovered per-agent one at a time (main → wrench → ...)
   - Error: `Legacy session store requires migration: .../sessions/sessions.json`
   - Fix: `openclaw doctor --session-sqlite import --session-sqlite-all-agents` (this flag handles all agents at once, but default `doctor --fix`/`--repair` does not run it)
4. **Tailscale 443 route ownership conflict** — was a stale route from a previous OpenClaw version pointing at the same gateway
   - Error: `Tailscale HTTPS port 443 is already owned by a route whose ownership OpenClaw cannot prove`
   - Fix: `tailscale serve --yes --https=443 --set-path=/ off`, then restart
5. **Multi-agent ambient owner unset**
   - Error (on `openclaw status --all` etc.): `Multiple agents are configured, but this operation has no explicit owner`
   - Fix: `openclaw config set agents.defaults.systemAgent.agentId main`

**Ask:** have `doctor --fix`/`--repair` handle all of these in one pass, and have the diagnostic step surface every blocker up front instead of one per restart attempt. An automated post-`npm update` health check (doctor --repair + session-sqlite import + plugin capability consent) would prevent this entirely.

## Part 2 — `doctor --fix` fails on its own error message (self-referential loop)

[...]

Issue #133984, opened 2026-08-31
[Bug]: 2026.7.1-2 → 2026.8.1 leaves Gateway unstartable; startup and `doctor --fix` both skip config-key migrations non-interactively


Lendersmark:
## Summary

Upgrading `2026.7.1-2` → `2026.8.1` (`ea80657`) left the Gateway unstartable and required roughly a dozen manual repair steps across five independent defects. The documented recovery path — `openclaw doctor --fix` — could not repair the config, because it applies **no config-key migrations at all when stdout is not a TTY**, while `openclaw config set/unset` simultaneously refuses to write *because* the config is invalid. That is a deadlock for anyone running the Gateway as a supervised service.

This is filed separately from #133813 (same version path, same crash-loop, macOS) which covers only the `exec-approvals.json` / legacy session-store gate. Defects 1–5 below are not mentioned there. I hit the exec-approvals gate too and can confirm it on Linux.

## Environment

| | |
|---|---|
| OpenClaw | `2026.8.1` (`ea80657`), upgraded from `2026.7.1-2` |
| Install | npm global, `~/.npm-global` |
| Node | v24.18.0 |
| OS | Fedora 44, kernel 7.1.10-200.fc44.x86_64 |
| Supervisor | systemd `--user` unit, `Restart=` enabled |
| Config | single-file `openclaw.json`, no `$include`, not Nix-managed, 17 agents, 2 channels (Telegram ×7 accounts, WhatsApp ×2) |

`openclaw update` itself reported success for the package step and then `Running doctor checks ✗` — but the actual state was a dead Gateway, which the update output did not make clear.

---

## 1. Gateway startup does not apply the documented prompt-free legacy-key migrations

[The configuration docs](https://docs.openclaw.ai/gateway/configuration) state:

> Gateway startup first applies safe legacy-key migrations to eligible single-file configs.

described as *"deterministic, prompt-free transforms"* that *"operate the same way as `openclaw doctor --fix`"*, writing *"only when the entire migrated config validates"*.

My config is eligible by every stated criterion (single file, no `$include`, not Nix, not written by a newer version). Startup nevertheless refused to boot with `status=78/CONFIG`:

```
openclaw.json:35 — meta: Unrecognized key: "lastTouchedAt"
openclaw.json:56 — agents.defaults: Unrecognized keys: "memorySearch", "mediaGenerationAutoProviderFallback", "imageGenerationModel"
[...]

PR #141412, opened 2026-09-07
fix(doctor): admit maintenance before database inspection

steipete:
## What Problem This Solves

Fixes an issue where operators running `openclaw doctor --fix` after an upgrade could wait on database snapshots before learning that the Gateway or an agent still owned the state. An unreadable legacy workspace file also lacked an actionable recovery step, leaving startup blocked after Doctor returned.

## Why This Change Was Made

Doctor now verifies or stops its managed service, acquires Gateway and state coordination, and checks agent leases before taking schema snapshots. The diagnostic lease reader neither migrates state nor consults a persisted quarantine receipt that Doctor needs to repair. A live Gateway owns its coordinator for its lifetime, so admission refuses it immediately. Optional diagnostic imports follow admission.

Recognized canonical-index drift reaches the schema repair owner; strict readiness validation runs after repair. Ordinary read-only close remains retryable after a transient snapshot cleanup failure.

Snapshot and initial integrity children have a 30-second execution fence. Timeout and cancellation kill and join the child before releasing ownership, and the parent cleans unpublished snapshot copies. Integrity uses a process because a Worker thread cannot reliably interrupt native SQLite. The recovery docs consolidate downgrade guidance, distinguish retained migration originals from package-transaction backups, and remove reverse-schema SQL recipes. Unreadable workspace sources remain intact and name backup or retained-file recovery.

Main's QA catalog breakage from #141514 was reproduced and repaired during landing. Our YAML correction passed all 20 native scenario-runner tests, then #141555 landed an equivalent fix. This branch now includes that upstream repair and drops the redundant YAML hunk. The effective selector matches this exact current line in `ui/src/e2e/control-ui-credentials.e2e.test.ts:114`:

```ts
it("submits one secret for both authentication modes with visible acceptance and rejection recovery", async () => {
```

The scenario retains its token/password-mode ID and coverage key; its title and criteria describe the single Gateway secret field supplying both connect auth values. The existing alignment test failed before the correction (one failure, 19 passes).

### Production footprint

An absorption pass removed 9 production lines; the current net is **+127**. Both lease checks now use one typed lease query and the same PID/start-time predicate. Integrity timeout/cancellation use Node's native `fork` options, removing the custom timer and abort listener. The remaining growth protects these boundaries:

| Owner | Net lines | Required boundary and simpler alternatives considered |
| --- | ---: | --- |
| `src/commands/doctor-maintenance.ts` | −33 | Removed the duplicate schema preflight; this owner now only establishes service/coordinator/lease authority before inspection. |
| `src/commands/doctor.ts` | +7 | Defers optional session-maintenance and post-upgrade imports until those modes run. Eager imports load their state graphs before normal repair can reject a live owner. |
| `src/flows/doctor-health.ts` | +25 | Absorbs configured-path schema admission from maintenance, preserves diagnostic Doctor's update offer, and reports admission failure through the updater's existing result envelope. This offsets the maintenance deletion; a second detector would restore the original ordering defect. |
| `src/state/openclaw-agent-db-lease.ts` | +56 | Adds read-only admission without schema creation, stale-row deletion, snapshots, or a fresh quarantine-store read. It borrows an existing handle without closing it, bounds/restores its busy policy, and closes only its own connection. The canonical query and liveness predicate are shared. The writable checker must retain transactional maintenance/deletion fencing and stale-lease cleanup, so calling it before admission would mutate the state being protected. |
| `src/infra/sqlite-readonly-worker.ts` | +35 | Bounds synchronous/asynchronous snapshot processes and joins `close` before cleanup or lease release. `execFile` does not forward `killSignal` on AbortSignal cancellation, so its explicit SIGKILL listener is necessary. Native timeout options and the error formatter are shared with integrity inspection. |
| `src/infra/sqlite-readonly-location.ts` | +14 | The parent owns staging before spawn, allowing removal when a killed child never publishes its copy path. Ordinary handles retain retryable boolean cleanup; cancellable maintenance retains throwing cleanup. Combining those policies reproduced a close-contract regression. |
| `src/infra/sqlite-integrity-worker.ts` | +12 | A process can interrupt native SQLite; a Worker thread cannot reliably do so. Native `fork` options now own timeout/cancellation. The remaining code sends typed IPC input, waits for `close`, and preserves file identity plus native error details. Reusing the stdout snapshot transport would require adapters for BigInt file identities and the existing error envelope. |
| `src/infra/sqlite-integrity.worker.ts` | +5 | Receives private process IPC and binds send/disconnect after checking availability. These replace Worker-only input/result plumbing; the diagnostic envelope remains unchanged. |
| `src/infra/state-migrations.workspace-setup.ts` | +6 | Uses the existing recovery formatter to tell the operator how to retain or restore an unreadable source/claim and retry. Removing the recovery action would reinstate the Doctor/startup dead end. |

A custom common subprocess wrapper was considered. It would need separate adapters for captured stdout/stderr and advanced IPC, while still retaining each transport's failure rules. The shared boundary is Node's process lifecycle; adding that wrapper would obscure ownership without reducing the required work.

## User Impact

Explicit repair refuses live ownership before copying databases and reports which service or process must stop. Supported startup-blocking migrations run without prompts with `--fix --non-interactive`; malformed or conflicting inputs require the manual action in the diagnostic. Stalled snapshot or asynchronous integrity inspection ends with a database-specific recovery message.

## Evidence

- Reinstating the original Doctor flow and maintenance owner makes all three admission regressions fail: Gateway/state/agent cases spawn 3/3/4 snapshot children before refusal. The repaired flow spawns none. Existing maintenance tests also verify the fence survives until cancelled native work closes.
- Real stalled snapshot and integrity processes fail on the original code at the fixture's 35-second safety exit and pass after the repair at approximately 30 seconds. A 300 MiB synthetic database took 1.13–3.15 seconds to snapshot and under one second to check; the 30-second fence leaves storage headroom after ownership admission.
- Shipped-package reproduction: 7.1-2 created shared schema 1; raw 9.2 startup exited 1 on legacy workspace state after changing the SQLite/WAL/SHM family to schema 15 and adding startup-failure/lock artifacts. The fixture's configuration bytes were unchanged.
- [Clean Linux rig](https://github.com/openclaw/openclaw/actions/runs/34155452783), Node 24.19.0: the built candidate refused the live shipped 9.2 Gateway with exit 1 in **0.844 seconds**. With writers stopped, it repaired actual 7.1-2 state with exit 0, migrated shared schema 1→16, preserved the legacy audit event, retired workspace/session/exec-approval sources, and booted the candidate Gateway. The native Testbox wrapper exited 0 and stopped its lease; the backing Actions step can show cancellation during delegated cleanup.
- CI follow-up regression proof: `lets the schema repair owner decide current shared-state %s` repairs missing/drifted indexes while a missing required table stays absent and fails readiness; `reports success when retrying transient read-only snapshot cleanup` preserves retryable close. All four fail on the prior pushed production for the intended reasons. The newer-agent-schema cases now exercise `runDoctorHealthFlow` and retain byte-preservation assertions. The IPC WAL-commit fixture and Windows directory-sync cases retain their real filesystem/database checks.
- The broad focused run passed 601 tests (two existing skips); the relocated eight-case schema-preflight suite and targeted lint passed. After the absorption refactor, 86 focused lease, deletion, admission, integrity, and deadline tests passed, and the shipped-package rig was rebuilt and rerun. Rebasing onto the merged main UI repair and routing #141416 preserved both routing changes; all 186 focused integration tests and a fresh branch-mode P0–P2 review passed. Doctor, workspace, SQLite, cancellation, and agent-maintenance tests passed. `pnpm build`, `node scripts/check-changed.mjs`, documentation anchor validation (9,510 links; zero broken), `git diff --check`, and fresh P0–P2 autoreview passed. The refreshed rig home-marker assertion and the earlier local real-state guard both reported zero changed files.

The [fresh physical-checkout gate](https://github.com/openclaw/openclaw/actions/runs/34161842804) completed all typechecks and lint with exit 0 and stopped its Testbox. The local nested checkout had correctly refused ancestor `node_modules` declaration inputs; the guard was preserved and validation moved to an isolated checkout.

Current main already authorizes all four named migration families in explicit non-interactive repair. The combined rig verifies that behavior; no redundant migration switches were added. No schema, configuration option, environment variable, dependency, or validation baseline was added or changed.

Related: #133984 #139105 #119583 #138892 #134896

Reported by @Rocky_Chuck: https://x.com/Rocky_Chuck/status/2096970154622267418


steipete:
Addressed both findings and rank-up moves. Doctor now lets recognized canonical-index drift reach its existing schema repair owner while retaining header/readability/newer-version refusal and final strict readiness. The missing-index and drifted-index cases repair and complete; a missing required table remains absent and fails readiness. The four newer-agent-schema refusal cases now run through `runDoctorHealthFlow` and retain unchanged config/state/agent bytes assertions.

The schema/cleanup regressions failed on the prior pushed production for the expected reasons and pass after the repair. The broad focused run passed 601 tests with two existing skips; the relocated eight-case schema-preflight suite also passed. CI fixture repairs preserve independent WAL visibility, Windows fsync policy, and retryable read-only cleanup.

steipete:
Compatibility proof confirmed for this PR. The refreshed built-runtime run is https://github.com/openclaw/openclaw/actions/runs/34155452783 (Node 24.19.0): Doctor refused the live shipped 2026.9.2 Gateway in 0.844 seconds, then repaired actual shipped 2026.7.1-2 state with the writers stopped. It migrated shared schema 1→16, preserved the audit event, retired the workspace/session/exec-approval sources, exited 0, and the candidate Gateway reached readiness. The home marker reported zero changed files; the wrapper exited 0 and stopped its Testbox.

The tested tree was `8c8d1d5e5eaefc7b8c9b3e28a1eba0b90b50075b`. The subsequent rebase preserved the production patch and added the merged routing repair; 186 integration tests passed, followed by the complete changed-file gate and clean independent reviews. This PR changes no SQLite schema. Read-only admission of a restored quarantined primary, supported canonical-index repair, and unchanged-file refusal for newer schemas are covered by the focused tests.

This confirms the compatibility evidence already acknowledged in the latest review. The separate CI failure is the stale QA scenario title/criteria introduced on main by #141514; it is being handled at its ownership boundary. Prepare and merge remain gated on green exact-head CI.

steipete:
Merged via squash.

- Prepared head SHA: [5a09bb942a27a10f321a3135746b29e02cafd3be](https://github.com/openclaw/openclaw/pull/141412/commits/5a09bb942a27a10f321a3135746b29e02cafd3be)
- Landed commit: [662080ac830ea15d147a498d72a1abf884e48b37](https://github.com/openclaw/openclaw/commit/662080ac830ea15d147a498d72a1abf884e48b37)

PR #141416, opened 2026-09-07
fix(gateway): stop restart loops for required maintenance

steipete:
## What Problem This Solves

Fixes an issue where Gateway startup refusals requiring offline migration were treated as crashes after an upgrade. Systemd could restart the same refusal repeatedly, and the crash-loop breaker could subsequently suppress channels.

## Why This Change Was Made

Use one typed maintenance family from the state, agent-media, workspace, session-store, and newer-schema producers through config reading and Gateway startup. Maintenance exits use 78 and retain LaunchAgent parking. Boot records carry a maintenance reason that the breaker excludes; Doctor marks those records repaired only after all required readiness checks succeed and before releasing maintenance ownership.

## User Impact

Recovery guidance gives the ordered steps: stop the service, run `openclaw doctor --fix`, then start it again. A newer-schema refusal instead directs operators to restore their pre-update `openclaw backup`. Maintenance refusals no longer contribute to channel suppression when recorded in the boot lifecycle.

## Evidence

Closed all three classification/swallow sites:

- `src/cli/gateway-cli/startup-maintenance.ts:9`: replace the narrow resolver with the shared typed family, including legacy workspace/session refusals.
- `src/config/io.health-state.ts:158`: propagate maintenance failures instead of logging and continuing.
- `src/config/io.snapshot.ts:364`: propagate maintenance failures instead of flattening them into an invalid-config snapshot.

Regression evidence includes a real temporary legacy workspace exiting 1 before and 78 after; audit health-write and public config-snapshot propagation; maintenance boot accounting; in-process restart refusal; and Doctor's unsuccessful/successful repair boundary. A mixed-error regression also preserves newer-schema precedence: even when a repairable error occurs first, the recovery guidance remains stop → restore backup → start. The original Doctor code leaves `startup_failed` after successful repair; the changed code marks `startup_failure_repaired`. With the old snapshot catch restored, the new snapshot assertion resolves an invalid-config object instead of rejecting with the typed error.

Focused regression names:

- `run.option-collisions.test.ts`: “exits 78 when the only startup blocker is legacy workspace setup state”; “stops newer-schema retries” includes mixed maintenance errors.
- `io.health-state.test.ts`: “propagates audit migration required from health writes and config snapshots”.
- `gateway-boot-lifecycle.test.ts`: “does not count maintenance refusals toward channel suppression”; “repairs maintenance history without clearing real startup crashes”.
- `doctor-health.test.ts`: “refuses blocked required migration … then completes after the writer releases”, for both repair and yes modes.
- `run-loop.test.ts`: typed maintenance reasons and maintenance failure during deferred restart; both Gateway CLI suites pass 178 tests.

The shipped-package rig starts real 2026.7.1-2, then reproduces shipped 2026.9.2 exiting 1 after shared schema 1→15 with an untagged startup failure. The built candidate exits 78 on three untouched schema-1 attempts. Its own `doctor --fix --non-interactive` then exits 0 and prepares the database. After the synthetic legacy workspace marker is reseeded, four further built-worktree attempts each exit 78 and record `gateway.maintenance_required`; the fourth attempt does not trip the breaker. Current Doctor on that mature fixture exits 0 and changes all four records to `startup_failure_repaired`. The final rig pins both Gateway and Doctor probes to the assigned scratch port range; the real-home guard reports zero changed files.

`pnpm build`, an updated `pnpm build cliStartup` for the isolated Doctor caller change, focused regression tests, production/test typechecks, import-cycle checking, and scoped P0/P1 autoreview passed. `node scripts/check-changed.mjs` completed with exit 0.

CI follow-through corrected the shared built-session lifecycle proof to expect exit 78 while retaining its message, path, and source-byte assertions. A separate inherited-clock CI flake in `session-state-events.test.ts` was reproduced with a pending exec wake from an earlier fake-clock origin; setup now drains inherited timers before routing assertions. The regression fails before that setup fix and all 29 tests pass after it. Behavioral timing and assertions are unchanged. The unrelated update-cleanup process timeout passed all 18 cases on local reproduction and passed the next hosted run without a timeout change.

ClawSweeper’s latest review reports no actionable patch findings. Its data-model compatibility proof checkbox is satisfied by the existing-database rig above; its automatic SQLite-table warning is inapplicable because this PR adds no table, column, or schema version.

A later merged-tree lint run exposed a Markdown test file on main exceeding its line limit. Upstream #141466 split those tests while this PR was being validated; the PR is rebased over that fix and carries no UI edits. The four routing commits were patch-identical across the rebase.

The full hosted Doctor suite also caught a no-create edge case in repair accounting: an absent state database was opened solely to update nonexistent history. The lifecycle owner now returns zero for a definitely absent state path. The existing no-create test reproduces the failure before the correction; all 87 Doctor tests and 16 boot-lifecycle tests pass after it. Production remains net −69 lines.

`src/commands/doctor-config-flow.ts` has a deliberately isolated one-line `observe: false` addition to `runDoctorConfigPreflight`. Doctor reads config without writing health state before its migration stage. This is the only change in that file, so the companion Doctor/preflight work can rebase over it directly.

Legacy untagged-history compatibility was checked against shipped writers. Commit `367ff3cea190a3f725627adc6010e6016e9d0f3b` introduced the untagged media refusal; `f6e874835c603e7e20d3da9bc49c51657cf6bb6f` added tagging. The only release tag containing the first but not the second is `v2026.7.2-beta.5`. Stable [v2026.8.1](https://github.com/openclaw/openclaw/blob/v2026.8.1/src/cli/gateway-cli/run-loop.ts#L1059) and [v2026.9.2](https://github.com/openclaw/openclaw/blob/v2026.9.2/src/cli/gateway-cli/run-loop.ts#L1221) both tag these failures. The old `gateway.agent_media_migration_required` tag remains supported. ClawSweeper's rank-up to restore untagged matching is explicitly skipped under the maintainer decision to remove beta-only compatibility; a later reader's beta-history fallback does not establish a stable writer. No SQLite table, column, or schema version changes in this PR.

The rig still observes shared SQLite file writes on early schema refusals. On workspace refusals it also observes WAL/SHM files, startup stability bundles, and the agent CLI shim; config bytes remain unchanged in this fixture. Readiness-before-write ordering remains separate startup-preflight work. This change classifies and accounts for refusals; it does not move migration gates or make startup writes atomic.

Related: #133984 #139105 #134896

Reported by @Rocky_Chuck.


steipete:
Merged via squash.

- Prepared head SHA: [c8b137d4792e9702ab0ff26c812aa5e7854b4ff3](https://github.com/openclaw/openclaw/pull/141416/commits/c8b137d4792e9702ab0ff26c812aa5e7854b4ff3)
- Landed commit: [bdf5d9939abf8262a247e73a1fa8b6c202bd77e8](https://github.com/openclaw/openclaw/commit/bdf5d9939abf8262a247e73a1fa8b6c202bd77e8)

PR #141280, opened 2026-09-07
fix: close prepared model runtimes before auth teardown

steipete:
Gateway shutdown could clear auth snapshots while prepared model owners were still listening, starting fresh discovery during teardown. Shutdown could also finish before an admitted catalog worker’s parent auth probe settled.

Close and restart now fence model admission, retire pending publications, and join the existing model builds and native worker cleanup before completing. Model snapshots and auth bindings stay in their original module; a process coordinator holds the actual cleanup callbacks. Test cleanup uses the same awaited path, replacing the duplicated reset that discarded pending queues.

Validation: causal lifecycle and native-worker failures before the fix; 283 owner/native/Gateway tests; 47 selector/cancellation tests after an equivalent helper extraction; the full changed-file gate; build; and an isolated built Gateway proving health, configured model visibility, graceful SIGTERM, and complete process cleanup. Independent P2 review found no actionable issues; the final helper extraction preserves the reviewed selection body exactly and has separate focused verification.

Production LOC: +102 for the missing process-close, admission, and cleanup ownership, including the owner-selection extraction; tests: +157. This is an independently useful prerequisite for the plugin lifecycle work in #135599.

PR #141422, opened 2026-09-07
improve: avoid repeated cold imports in catalog worker tests

steipete:
## What Problem This Solves

Prepared model-catalog integration tests repeatedly cold-load the same TypeScript dependency graph in fresh worker threads. A phase profile measured 50.2 seconds in post-start static imports across 25 workers, versus 1.6 seconds preparing their generations.

## Why This Change Was Made

Declare the catalog worker in the shared runtime entrypoint map so the existing Vitest invocation compiler prepares it once. Remove the now-duplicate package build entry. Each prepared generation still owns a separate worker thread; plugin artifact selection, auth refresh, cancellation, and process-close ownership remain unchanged. Source and packaged entrypoint paths also remain unchanged.

This reuses the existing compiler and cleanup owner instead of adding a worker pool, cache, or more shards. Runtime source grows by two lines; build configuration shrinks by one. Tests and assertions are unchanged.

## User Impact

Faster developer and CI catalog tests. No configuration, schema, dependency, or user-facing behavior change.

## Evidence

- On the same main snapshot including #141280, all 26 existing integration cases passed with identical names and order before and after. Source-worker command: 682.62s; candidate: 82.30s. Case sums: 500.39s and 49.71s. Host load and filesystem warmth varied substantially, so these are observed local timings, not a portable speedup ratio.
- The existing cases cover configured/standalone plugin artifacts, metadata scope, auth refresh, native routes, generation replacement, and all three current cancellation/process-close scenarios.
- Earlier actual-path instrumentation verified 25 source TypeScript workers became 25 JavaScript workers from the private invocation generation. A control retained the augmented compiler graph but forced source workers; all original 24 cases passed and import costs returned. Instrumentation was removed.
- Earlier standalone source-mode selection/auth cases and adjacent catalog/worker-URL tests passed. A matched compiler-graph probe added 16,160 bytes of output.
- Independent P2 review: no actionable findings.
- Full `pnpm build` passed, including built plugin loading and all 152 public SDK exports; the shared declaration resolves to the emitted `dist/agents/prepared-model-catalog.worker.js`.
- Full `check:changed` passed: all typecheck lanes, lint, static guards, and zero runtime import cycles.
- [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34149947557) passed in 8m36s. The [actual catalog job](https://github.com/openclaw/openclaw/actions/runs/34149947557/job/101830093505) ran all 26 cases with identical names/order: 21.289s total, 608–2072ms per case. Its model-test invocation prepared the shared generation in 5.793s. These CI logs do not emit individual worker URLs; actual-path proof comes from the local instrumentation described above.

### DIFF
--- a/docs/reference/test.md
+++ b/docs/reference/test.md
@@ -321,7 +321,7 @@ and runtime parents on TypeScript. Importing a declared subprocess entrypoint
 compiles the fixed test entry set and its workspace dependencies into one fresh
 invocation directory under `.artifacts/vitest-workers/`.
 
-The ten declared application entries run as plain Node JavaScript without a
+The declared application entries run as plain Node JavaScript without a
 TypeScript loader: SQLite read-only snapshots, database verification, Tailscale
 route ownership, the service relay, its POSIX and Windows anchors, the memory
 plugin's KNN child, session transcript archive and reconciliation workers, and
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2749a7a3dc66314eb1b7dc7925720ab583074c11](https://github.com/openclaw/openclaw/pull/141422/commits/2749a7a3dc66314eb1b7dc7925720ab583074c11)
- Landed commit: [26d322a67582b67bf47544c413430782397fed6f](https://github.com/openclaw/openclaw/commit/26d322a67582b67bf47544c413430782397fed6f)

PR #141427, opened 2026-09-07
improve(control-ui): fold Gateway settings into one connection section and explain Gateway switches

steipete:
## What Problem This Solves

Follow-up to #141087. Two connection surfaces still looked like the old design and used its vocabulary. Settings → Gateway showed "Gateway Access / Where the dashboard connects", a WebSocket URL, a Gateway Token and a "Password (not stored)" field always side by side, a Default Session Key, a Connect/Refresh pair, and a separate "Snapshot" table of Status, Tick Interval, and Last Channels Refresh. The "Change Gateway URL" confirmation that appears when a link points the browser at another Gateway was a danger-styled exec-approval card saying "Malicious URLs can compromise your system" without saying what was actually at stake.

## Why This Change Was Made

Settings → Gateway now has one Connection section: a status pill in the header, one fact line under the heading (host · auth mode · tick), and rows for Gateway URL, Credential, and Default session, each with a one-line hint. The credential row is a single secret input with the same Token / Password switch as the login gate; the mode resolution and the switch live in a new shared `credential-mode.ts` so the two surfaces cannot drift. Connect turns primary only when the draft differs from the live connection and says so; the Refresh button is gone because it refreshed the channel list, which Channels owns. The last-error row remains while disconnected. The Gateway host card is unchanged.

The Gateway-switch dialog now shows the current URL next to the one the link wants, and its note states the real rules: anything entered after switching goes to that host, the link carries its own token when it does, and, only when the origins differ, the saved credential for the current Gateway stays here and is not sent (mirrors the origin-scoped token rule in `resolveGatewayCredentialsForUrlEdit`). Buttons read "Switch to <host>" and "Keep current Gateway". Confirm/cancel wiring and the guard itself are unchanged.

Non-goals: no change to connect semantics, credential storage, or the guard's trigger conditions. Native app connection screens and the CLI wizard keep their own copy.

## User Impact

Operators editing the connection see one credential field that matches the login gate, hints that say where values come from, and a Connect button that only lights up when something changed. A pasted or crafted link that would move the browser to another Gateway gets a clear from → to view and an explanation of what travels with the switch, instead of a generic warning.

## Evidence

Before / after, Settings → Gateway (desktop):

Settings Gateway before and after

Gateway-switch confirmation (light and dark):

Switch dialog before and after

Switch dialog dark before and after

Phone (390px):

Settings Gateway on a phone

Captures are real renders from `pnpm dev:ui:mock`; the host shown in the dialog is the maintainer's own public Gateway hostname.

- Unit: `view.render.test.ts`, `connection-page.test.ts`, `login-gate.test.ts`, `vite-config.node.test.ts`: 74/74 (new cases: the switch changes which value the single input edits, a password-mode error selects Password, dirty-state Connect styling, and the header names the live Gateway rather than an edited draft URL).
- Playwright e2e locally: `sidebar-settings.e2e.test.ts` + `config-safe-write.e2e.test.ts` 19/19; other e2e files changed selectors only (dialog button "Confirm" → "Switch to <host>", "WebSocket URL" → "Gateway URL"). `control-ui-auth-transports` needs a real Gateway build; CI covers it.
- `node scripts/check-changed.mjs`: all lanes green; `pnpm tsgo:ui`, `pnpm ui:i18n:verify`, stylelint clean.
- Startup bundle: the first CI run tripped `control-ui-performance` (the switch dialog had pulled the login failure classifier into the startup chunk). Fixed by moving `formatGatewayHost` into a dependency-free `ui/src/lib/gateway-host.ts` and moving the Settings → Gateway copy from the eager catalog to the lazy `en-settings.ts` catalog; `pnpm ui:build` + `check-control-ui-performance --report-only` pass locally.
- Codex autoreview (P2): one accepted finding fixed (the connection summary used the editable draft URL instead of the live connection), rerun scoped-clean.


steipete:
@clawsweeper re-review

PR #141131, opened 2026-09-07
fix(test): prevent Vitest profiling startup hangs

steipete:
<details>
<summary>Additional instructions</summary>

Keep **Allow edits from maintainers** enabled so maintainers can update this branch.

</details>

## What Problem This Solves

Resolves startup hangs in Vitest profiling on Node 24.19.0. Main-mode help and invalid arguments could reach the 120-second CI deadline before returning an outcome. Native sampling during an unchanged-suite reproduction found OpenSSL hash enumeration and Node's background bundled-certificate initialization waiting on locks.

## Why This Change Was Made

The shared profiler child now completes Node's cached hash enumeration before importing Vitest/Vite and their TLS dependencies. Main CPU capture starts first, preserving startup measurement. The code records the removal condition: supported Node/OpenSSL versions must make concurrent initialization safe.

The process-cleanup repair already landed in #141049; its managed fixture lifetime and cancellation regression are preserved.

## User Impact

Developers retain the existing main/runner commands, pool selection, validation errors, and CPU/heap outputs, with deterministic startup ordering. There are no certificate-trust, configuration, timeout, dependency, or runtime-pin changes.

## Evidence

- The four existing help cases observe the first actual hash-enumeration call through a Node preload. Against the original child, all four deterministically fail with TLS already loaded; the observer stops before entering the native race. With the fix, they verify enumeration precedes TLS and main CPU capture is already active.
- `node scripts/run-vitest.mjs test/scripts/run-vitest-profile.test.ts --reporter=verbose`: **35/35 passed** locally, including the previously landed cancellation case.
- Real main/runner commands with both forks and threads produced valid CPU/heap profiles and left no recorded profiling processes alive.
- `node scripts/check-changed.mjs -- scripts/run-vitest-profile-child.mts test/scripts/run-vitest-profile.test.ts` passed, including script/root-test types and targeted lint. Independent review through **P2** was scoped-clean; `git diff --check` passed.
- [Linux CI run 34109753909](https://github.com/openclaw/openclaw/actions/runs/34109753909) passed all **35 profiler cases** in [owner job 21](https://github.com/openclaw/openclaw/actions/runs/34109753909/job/101703264999) and **9 sibling profiling flows** in [owner job 20](https://github.com/openclaw/openclaw/actions/runs/34109753909/job/101703264887), on Node 24.19.0. The sibling flows execute main/runner commands with forks/threads, validate CPU/heap profile contents, and check namespace cleanup.
[...]

PR #141429, opened 2026-09-07
fix(ui): show link chips for URLs written in inline code

steipete:
## What Problem This Solves

Fixes an issue where a URL written in backticks in a chat message, such as `` `https://github.com/openclaw/openclaw/pull/141131` ``, rendered as a plain monospace code span with the full URL. The same URL written bare shows the compact GitHub chip with an icon, a shortened label, and the hover card. Models frequently wrap URLs in backticks, so the chips were missing exactly where they were most useful.

## Why This Change Was Made

markdown-it never linkifies inside `code_inline`, and the Control UI's link classifier only walks `link_open` tokens, so a backtick-wrapped URL never reached the chip logic. A code span whose entire content is a single absolute HTTP(S) URL is now wrapped in a link inside that same classifier and revisited by it, so GitHub issue and pull-request links get the `#number` chip and hover card, other GitHub links get the shortened label, and non-GitHub links get their favicon. Session URLs written in backticks are left for the session-link rule that already owns them, and fenced or indented code blocks are untouched. Classification wins over authoring syntax: the inner `<code>` drops its box and inherits the link typography so the result is indistinguishable from a bare URL.

## User Impact

Backtick-wrapped GitHub issue and PR URLs in chat now show the same compact chip, icon, tooltip, and hover card as bare URLs. Other backtick URLs become normal links with a favicon. Code spans containing anything besides a single URL, and all code blocks, render exactly as before.

## Evidence

Before and after, light theme (synthetic fixtures):

Before: backtick URLs render as plain code spans
After: backtick URLs render as GitHub chips and favicon links

- `node scripts/run-vitest.mjs ui/src/components/markdown-links.test.ts ui/src/components/markdown.test.ts ui/src/components/github-link-hovercard.test.ts`: 3 files, 278 tests passed. The five new inline-code cases failed before the parser change (expected anchor absent) and pass after.
- `ui/src/styles/chat-github-link-presentation.browser.test.ts` (Chromium): 5 tests passed, including inherited typography, no inner code box, and identical geometry to the bare-URL chip in light and dark.
- `pnpm check:changed`: passed.
- Independent code review: clean at P2, including the CI repair.
- CI exposed a selector-dependent contrast check in two jobs. Restoring the ordinary-code selector keeps that coverage intact while the more specific link rule owns the code-URL styling. The contrast and Chromium presentation suites then passed together: 2 files, 10 tests.

## Known Gaps

Hover-card fetches were verified against the rendered anchors, not a live Gateway round trip; the hover card keys on `href`, which is unchanged in shape.


steipete:
Superseded by #141446, which landed the same backtick-URL promotion for `github.com` links (chip, tooltip, hover card) with a simpler text-token approach that needs no CSS change. Closing this PR rather than rebasing it.

The one behavior this branch had beyond #141446 was promoting non-GitHub backtick URLs to favicon links while deferring session URLs to the session-link rule. #141446 deliberately keeps non-GitHub code spans as code, so that stays a separate product decision.

PR #141433, opened 2026-09-07
fix: restore provider plugin loading in source catalog workers

obviyus:
## What Problem This Solves

Fixes an issue where source Gateway catalog refresh lost discovered models because an import-only compiled dependency failed inside the catalog worker.

## Why This Change Was Made

TypeScript workers use the same ESM preload as the supported source command. The previous general preload installed a CommonJS hook that rewrote native ESM loading incompatibly. The existing pool still owns default preload selection.

## User Impact

Affected provider plugins load during source catalog refresh. Explicit worker options and existing CommonJS/compiled behavior remain unchanged.

## Evidence

The real-worker regression failed at the dependency export while three compatibility controls passed. After the fix, all 19 regression/lifetime tests and three plugin-loader controls passed. A source command started a Gateway, refreshed through synthetic account/catalog endpoints and returned the fixture model. No external authentication, inference or channel operation was tested.

### DIFF
--- a/src/infra/worker-task-pool.source-loader.test.ts
+++ b/src/infra/worker-task-pool.source-loader.test.ts
@@ -1,0 +1,82 @@
+import fs from "node:fs";
+import path from "node:path";
+import { pathToFileURL } from "node:url";
+import { afterEach, describe, expect, it } from "vitest";
+import { createTempDirTracker } from "../../test/helpers/temp-dir.js";
+import { WorkerTaskPool } from "./worker-task-pool.js";
+
+const tempDirs = createTempDirTracker();
+afterEach(tempDirs.cleanup);
+
+async function runWorker(directory: string, filename: string, source: string) {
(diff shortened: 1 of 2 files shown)


obviyus:
Landed after a clean exact-head local review and successful CI.

**Proof:** The source Gateway completed `models.list` refresh, reached both local synthetic HTTPS endpoints, and returned the unique fixture model with a ready provider outcome. The Gateway and fixture stopped afterward. All 19 worker tests and three focused native plugin-loader controls passed.

| Before | After |
| --- | --- |
| Source catalog workers failed loading compiled dependencies with import-only exports. | The existing ESM-only preload preserves those imports and returns discovered models. |

CommonJS TypeScript, compiled CommonJS/ESM, cancellation and shutdown controls remain green. The proof used the declared installed dependency versions; it does not claim vendor inference or a fresh full lockfile installation.

PR #141441, opened 2026-09-07
fix(ollama): reject embedding-only chat setup

### LINKED ISSUE #141424: Ollama onboarding accepts embedding-only models as chat defaults
Non-interactive Ollama onboarding accepts an explicitly selected embedding-only model as the agent's chat default, even when the server advertises that it cannot complete chat requests.

Reproduced with real Ollama 0.33.3 and `all-minilm:22m` (384-dimensional embeddings):

1. `/api/show` and `/api/tags` report `capabilities: ["embedding"]`.
2. The actual reset-validation owner accepts `--custom-model-id all-minilm:22m`.
3. The actual setup owner returns configuration replacing a synthetic chat default with `ollama/all-minilm:22m`, logs success, and reports no error.
4. `/api/chat` returns HTTP 400: `"all-minilm:22m" does not support chat`.
5. `/api/embed` succeeds, confirming a reachable, working embedding model rather than an unavailable server.

Reset validation also accepts an inventory that is positively identified as containing only embedding models. The fallback model-selection path already excludes embedding-only models; explicit selection bypasses that check.

Expected: reject known embedding-only chat selections before configuration application or reset, with guidance to choose a chat model. Preserve support for models advertising both completion and embeddings, and for older servers without capability metadata.

Source checked on current main; setup owner SHA256 `5c8e87f95a257ce411e8f0ec5e8f80f415e7ce3a9779bcb82b353c99ec3dce5d`. Proof used an isolated daemon and synthetic configuration; no operator state was changed.

steipete:
Closes #141424.

## What Problem This Solves

Non-interactive Ollama setup can select an embedding-only model as the agent's chat default and report success, even though the server rejects every chat request for that model. Reset validation also accepts the explicit invalid choice or an inventory advertised as entirely embedding-only.

## Why This Change Was Made

Reuse the existing embedding-only capability predicate before configuration application and reset. Inspect an explicitly selected installed model using its canonical name, prefer authoritative model metadata, and retain list metadata if inspection is unavailable. Missing capability metadata remains compatible with older servers; models advertising both completion and embeddings remain eligible.

## User Impact

Invalid local chat selections stop with an actionable error before setup changes the configuration. Existing chat selections, cloud checks, model aliases and failed-download fallback behavior are preserved.

## Evidence

- Four regressions fail on the original source for the invalid choices. Final setup/registration suites pass 182 tests, and the onboarding ingress suite passes 95 tests.
- Real isolated Ollama 0.33.3: all-minilm:22m reports embedding-only, chat returns HTTP 400, embeddings return HTTP 200 with 384 dimensions. Patched production setup and reset-validation owners reject it; synthetic configuration remains unchanged. A real qwen3:0.6b positive control validates, selects and completes chat with HTTP 200.
- Full changed-code checks passed, including production/test typechecks, formatting, lint, unused-export analysis and repository guards. Independent P2 review found no actionable findings.
- Both test daemons and model runners stopped; ports are closed. Live proof exercised the production provider owners with real servers, without running a destructive reset or writing operator configuration.

### DIFF
--- a/docs/providers/ollama.md
+++ b/docs/providers/ollama.md
@@ -98,6 +98,12 @@ OpenAI-SDK-style examples, but new config should use `baseUrl`.
 
     `--custom-base-url` and `--custom-model-id` are optional; omitting them uses the local default host and the `gemma4` suggested model.
 
+    A local model advertised as embedding-only cannot be selected as the chat
+    default. Setup reports an error and leaves the existing configuration intact;
+    reset preflight also rejects an explicitly selected embedding-only model or
+    an inventory advertised as entirely embedding-only. Models that support both
+    completion and embeddings remain eligible.
+
   </Tab>
 
   <Tab title="Manual setup">
--- a/extensions/ollama/index.test.ts
+++ b/extensions/ollama/index.test.ts
@@ -465,6 +465,41 @@ describe("ollama plugin", () => {
       models: ["qwen2.5-coder:7b"],
     },
     {
+      name: "rejects an explicitly selected embedding-only model before reset",
+      models: ["embedding-model"],
+      customModelId: "embedding-model",
+      capabilities: ["embedding"],
+      omitListedCapabilities: true,
+      error: "Ollama model embedding-model only supports embeddings. Choose a chat model instead.",
+    },
+    {
+      name: "retains known embedding-only metadata when inspection is unavailable",
+      models: ["embedding-model"],
+      customModelId: "embedding-model",
+      capabilities: ["embedding"],
+      inspectionFailed: true,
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ccdb704a187a162c242da9205e41fdb31dce858f](https://github.com/openclaw/openclaw/pull/141441/commits/ccdb704a187a162c242da9205e41fdb31dce858f)
- Landed commit: [6979589c620005eac74a45abe1af59a78cfe5bd0](https://github.com/openclaw/openclaw/commit/6979589c620005eac74a45abe1af59a78cfe5bd0)

PR #1234, opened 2026-01-19
Gateway: fix ws checkServerIdentity typing

chrisrodz:
## Summary
- Fixes a TS2322 build error in `src/gateway/client.ts` by aligning `checkServerIdentity` with the `@types/ws` boolean signature.
- Keeps the TLS fingerprint validation behavior but now throws on mismatch and returns `true` on success.

## Issue
`wsOptions.checkServerIdentity` is typed by `@types/ws` as `(servername: string, cert: CertMeta) => boolean`, but our earlier implementation returned `Error | undefined`, triggering `TS2322` during `pnpm build`.

## Fix
- Update `checkServerIdentity` to throw `Error` on missing/mismatched fingerprint and return `true` when validation succeeds, matching the expected boolean signature.

## Testing
- `pnpm build` (user verified; build succeeds).

## Prompts
- "help me debug this build error I'm getting on pnpm build: ➜  clawdbot git:(main) pnpm build ... error TS2322 ..."
- "pnpm build works as expected. should we open a PR to the upstream project with this fix?"
- "I meant to the clawdbot repository"
- "Yes, commit, push and open a PR. Make sure to include the prompts we used, your explanation of the issue and our testing (pnpm build works) as well as anything else useful for the maintainer of the project to review"
- "seems like the fix you implemented is just masking the problem. let's apply the diff I pasted instead"
- "try again I switched back to the branch, here's the fresh diff for you"
- "update the PR and it's description to reflect this change"

PR #141446, opened 2026-09-07
fix(ui): show backtick-quoted GitHub links as chips with hover cards

steipete:
## What Problem This Solves

Fixes an issue where users reading chat messages would see a GitHub pull request or issue link rendered as a plain grey code box when the model quoted the URL in backticks (`` `https://github.com/openclaw/openclaw/pull/141131` ``), while the same URL written bare showed the compact `#141131` chip with its icon and hover card. Models quote links this way often, so a common message shape lost the reference affordance.

## Why This Change Was Made

markdown-it never linkifies inside code spans, so the existing GitHub link classification never saw these URLs. The parser now promotes a top-level code span whose CommonMark-normalized content is exactly one http(s) `github.com` URL, with no whitespace or control characters, into the same link tokens linkify produces, so the existing chip, label, tooltip, and hover-card path applies unchanged and the emitted markup is identical to a bare URL. Fenced and indented code blocks, code spans with surrounding text, non-GitHub URLs, and code spans inside authored `[label](url)` links stay as code.

## User Impact

Backtick-quoted GitHub issue and pull request links in chat now show as `#number` chips with the issue or pull-request icon, keep the full URL in the tooltip, and open the hover card. Repository and file links quoted the same way get their short label with the GitHub mark. No change for other code spans or code blocks.

## Evidence

- New parser tests cover a quoted pull request, issue (with symmetric padding), and repository URL, plus negative cases: non-GitHub URL, leading prose, trailing prose, prose after a fragment, one-sided padding, an adjacent control character, and a code span inside an authored link. The positive cases fail on the previous parser; the trailing-prose cases fail without the whitespace guard.
- Review follow-ups: ClawSweeper's P2 (trailing prose folded into the href because the URL parser percent-encodes spaces) and two Codex autoreview findings (one-sided padding bypassing a trimmed guard, C0 control characters) are fixed in follow-up commits. Final Codex autoreview at P2: scoped-clean.
- Rebased onto main after #141431 (bare `#1234` references) landed; that rule runs before this one and only touches text tokens, so the two do not overlap.
- `ui/src/components/markdown*.test.ts` and `github-link-hovercard.test.ts`: 361 tests pass across 10 files, plus the hover-card provider suite.
- `check:changed` clean (Control UI i18n, UI typecheck, dead-export scan, lint lanes).
- Docs: `docs/web/control-ui/feature-reference.md` updated; it previously stated that GitHub marks never appear inside code spans.
- Known gap: no live screenshot. The mock Gateway does not process chat sends, and the parser emits byte-identical markup to the bare-URL case, which the existing presentation browser tests cover.


steipete:
Status: the `check-lint-core-2` failure on head `1b7c4429ef6` is `ui/src/components/markdown.test.ts` exceeding the `max-lines` limit. That file is untouched here; it crossed the limit on `main` after #141369 and #141431 landed, so every PR rebased onto current `main` hits it. Repair is #141485 (verbatim test split). This PR will be rebased and re-verified once that lands.

steipete:
Merged via squash.

- Prepared head SHA: [afb1bc9a0729317a4a2923feadadf360e919e663](https://github.com/openclaw/openclaw/pull/141446/commits/afb1bc9a0729317a4a2923feadadf360e919e663)
- Landed commit: [c688a95c34f117692b33e45676822fea0db0d9ab](https://github.com/openclaw/openclaw/commit/c688a95c34f117692b33e45676822fea0db0d9ab)

PR #141449, opened 2026-09-07
fix(workers): explain blocked archives and preserve failure causes

steipete:
## What Problem This Solves

Fixes an issue where users archiving a failed cloud-worker session were told to retry the archive even though unresolved provider cleanup prevented the archive from starting. Worker setup and cleanup errors could also lose their actionable final diagnosis when an already-bounded provider error was truncated again.

## Why This Change Was Made

Preserve the typed placement-stop refusal at the archive boundary. Failed placements requiring cleanup direct the operator to **Stop cloud worker** and resolve its provider error instead of advertising an automatic archive retry. Transitional placement states remain retryable. The existing cancellation, reconciliation, and confirmed-teardown requirements are unchanged.

The shared worker-error formatter retains both operation context and terminal diagnosis with an explicit omission marker inside the existing size budget. Redaction still precedes UTF-16-safe slicing. This does not treat a released lease or a missing instance as proof of completed provider cleanup.

Related provider cleanup repair: https://github.com/openclaw/crabbox/pull/1904

## User Impact

Operators can distinguish a pending cleanup prerequisite from an actual stop failure or a transient worker transition. Bounded errors retain useful setup and cleanup causes across persistence and Gateway restart, without expanding logging or exposing credentials.

## Evidence

- Before the repair, nine archive-state cases failed the new actionable-error assertions; the existing safe-archive and reclaim cases passed. Both long provider-diagnostic cases also failed because their terminal causes disappeared; the short-error siblings passed.
- After the repair, 315 focused tests passed across archive placement, provisioning replay, inference, and the Crabbox provider owner. These cover untouched cleanup safeguards, no cancellation on pre-stop refusal, successful archive after proven cleanup, durable replay, redaction, Unicode preservation, and existing length bounds.
- `node scripts/check-changed.mjs` passed, including production/test typechecks and the selected lint/docs checks.
- The provisioning replay suite passed again after changing its Unicode assertion to a UTF-8 round trip compatible with the repository's TypeScript library target.
- Independent isolated review found no actionable P0–P2 issues. `git diff --check` passed.

No visual layout or styling changes. This PR fixes diagnosis and retry guidance; it intentionally does not force-archive an unresolved worker or repair provider resources itself.


steipete:
Merged via squash.

- Prepared head SHA: [004286f13468fee373c7af2d1419cdbdedf7a445](https://github.com/openclaw/openclaw/pull/141449/commits/004286f13468fee373c7af2d1419cdbdedf7a445)
- Landed commit: [16e478c38770093e12e6d10146aecddda3bb63b2](https://github.com/openclaw/openclaw/commit/16e478c38770093e12e6d10146aecddda3bb63b2)

PR #141450, opened 2026-09-07
fix(lmstudio): reload evicted embedding models

### LINKED ISSUE #141430: LM Studio embeddings fail after model eviction with JIT disabled
LM Studio memory embeddings stop working after their model is evicted when the server's JIT loading is disabled, despite OpenClaw preload being enabled.

Real reproduction on main `803067d4b9a` using an isolated headless LM Studio 0.0.23-1 server and Nomic Embed Text v1.5 Q4_K_M:

1. Disable the server's existing `justInTimeModelLoading` setting; leave OpenClaw preload enabled.
2. Construct the actual LM Studio embedding provider. Its initial preload succeeds, and an embedding returns 768 dimensions.
3. Unload that model; verify native discovery reports no loaded instance.
4. Reuse the same provider for another embedding. It sends no native load request and receives HTTP 400: `No models loaded. Please load a model in the developer page or use the 'lms load' command.`
5. Manually load the same model at 2,048 context tokens and repeat through the same provider. It succeeds with 768 dimensions.

Control: with default server JIT enabled, the same eviction sequence succeeds because the server loads the model itself.

The embedding provider currently prepares the model only at construction. Its request-time lifecycle needs to honor enabled preload after eviction, while preserving the preload opt-out, stable embedding model identity, cancellation and local-service lease ownership. Chat already checks readiness per request.

All prompts and state were synthetic. Each probe unloaded models and stopped its owned daemon; no operator service was changed.

steipete:
Closes #141430.

## What Problem This Solves

LM Studio memory embeddings fail after their model is evicted when server JIT loading is disabled. OpenClaw preloads at provider construction but previously sent later embedding requests without checking readiness, producing HTTP 400 even though preload remained enabled.

## Why This Change Was Made

Recheck model readiness before embedding requests. Serialize preparation within the provider while leaving inference concurrent, propagate cancellation through native discovery/load, and keep the service lease until an active preparation has settled. Cancelled queued requests never start a load. The canonical model identity is resolved only at construction, preserving embedding cache identity.

## User Impact

Memory embedding requests recover after eviction without a restart or manual model load. Single requests and query/document batches are covered. The existing preload opt-out and empty-batch behavior remain intact.

## Evidence

- Real isolated headless LM Studio 0.0.23-1 with Nomic Embed Text v1.5 Q4_K_M: initial embeddings succeeded with 768 dimensions, eviction caused the old provider to return HTTP 400, and a manual native load restored it. Default server JIT was also tested as a working control.
- The final live run matched this PR source tree and recovered single requests and both batch types with 768-dimensional vectors, with exactly one native reload per batch. The preload opt-out remained delegated. All 45 observed task processes stopped; both test leases were released.
- The eviction regression failed on the original source for single/document/query requests. Final isolated validation passes all 58 focused tests, including active and queued cancellation, lease lifetime, canonical identity and empty batches. Both typechecks, targeted lint, and all remaining selected repository guards passed after correcting four lint findings.
- Final independent review found no actionable P0–P2 findings, including the error normalization and cancellation paths. Test services are isolated from operator state.

## Data compatibility

The stored-data warning is a false positive for this diff: no schema, serialization, migration, cache-key producer, model default, embedding dimension, or vector conversion algorithm changes. `packages/memory-host-sdk`, `extensions/memory-core`, `src/state`, and `src/config` are unchanged from the PR base.

The unchanged `extensions/lmstudio/memory-embedding-adapter.ts` still builds `cacheKeyData` from provider ID, base URL, canonical model, and sanitized headers. Construction resolves the canonical model exactly as before; request-time preparation never changes `client.model`. The existing canonical/variant/cache-identity tests pass, and the new eviction integration test verifies both `provider.model` and `client.model` remain unchanged after recovery. The shared embedding request/vector path is also unchanged.

Existing indexes and cached vectors therefore retain their identity and format; this lifecycle repair requires no migration or reindex. The final live run used the same model and returned the same 768-dimensional contract before and after eviction.

### DIFF
--- a/docs/providers/lmstudio.md
+++ b/docs/providers/lmstudio.md
@@ -153,6 +153,9 @@ enough context for the selected model budget. A newly loaded instance is address
 identifier returned by LM Studio. Your configured model reference and conversation model identity
 keep the canonical model key.
 
+With preload enabled, embedding requests also check that their model is loaded. This lets memory
+embeddings recover after model eviction even when LM Studio JIT loading is disabled.
+
 ### Disabling preload
 
 LM Studio supports just-in-time (JIT) model loading, loading models on first request. OpenClaw
--- a/extensions/lmstudio/src/embedding-provider.integration.test.ts
+++ b/extensions/lmstudio/src/embedding-provider.integration.test.ts
@@ -102,3 +102,163 @@ describe("LM Studio embedding request headers", () => {
     }
   });
 });
+
+it.each(["single", "documents", "queries", "cancelled"] as const)(
+  "reloads evicted embedding models before %s requests while holding the service lease",
+  async (kind) => {
+    vi.stubEnv("NO_PROXY", "127.0.0.1");
+    vi.stubEnv("no_proxy", "127.0.0.1");
+    let holdLoad = false;
+    let signalLoadStarted: () => void = () => {};
+    const loadStarted = new Promise<void>((resolve) => {
+      signalLoadStarted = resolve;
+    });
+    let loaded = false;
+    let leases = 0;
+    const requests: string[] = [];
+    const server = createServer((request, response) => {
+      let text = "";
+      request.on("data", (chunk: unknown) => {
+        if (!Buffer.isBuffer(chunk)) {
+          throw new Error("Expected request bytes");
+        }
+        text += chunk.toString("utf8");
+      });
+      request.once("end", () => {
+        requests.push(request.url ?? "");
+        expect(leases).toBeGreaterThan(0);
+        response.setHeader("content-type", "application/json");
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b9dbe9400e81bc239eb8a765b33516c0fbb93a5e](https://github.com/openclaw/openclaw/pull/141450/commits/b9dbe9400e81bc239eb8a765b33516c0fbb93a5e)
- Landed commit: [69c961ff0bc467e523fae6773ff7bcfb08595e95](https://github.com/openclaw/openclaw/commit/69c961ff0bc467e523fae6773ff7bcfb08595e95)

PR #141457, opened 2026-09-07
fix(clickclack): ingest message attachments

sercada:
## What Problem This Solves

ClickClack message events are fetched authoritatively, but the connector discarded attachment metadata. Image messages therefore reached the agent as text only, causing the bot to claim that no image was attached even though ClickClack stored it.

## Summary

- type attachment metadata returned with ClickClack messages
- download inbound uploads with the bot bearer token and reject credential-forwarding redirects
- stage responses through the managed channel media runtime with size and idle-read limits
- pass ordered media facts into agent turns
- preserve text-only `replyMode: "model"`: attachment messages receive a visible unsupported-media reply without download, LLM completion, agent dispatch, or tool authority
- turn permanent media-limit rejections into a visible terminal reply so the account cursor can advance, while transient failures remain retryable
- re-check cancellation after asynchronous staging so shutdown cannot dispatch a late turn
- apply the existing bot-pair loop budget to unsupported-media and size-limit notices while preserving same-message retries
- propagate gateway cancellation into attachment downloads

The companion ClickClack server change makes new message and attachment creation atomic: https://github.com/openclaw/clickclack/pull/239

## Evidence

Run locally from this PR head against the companion ClickClack branch and the repository QA OpenAI provider:

1. Uploaded the incident screenshot as a 215,174-byte PNG and created a channel message with `upload_id` in one request.
2. The returned `message.created` message already contained that upload.
3. The real OpenClaw gateway consumed the ClickClack event, stored a managed inbound file of exactly 215,174 bytes, and staged the same file into the agent workspace.
4. The QA provider `/debug/requests` recorded one request whose user content contained both `input_text` and `input_image`; the image was a data URL and the media fact named the staged PNG.
5. ClickClack received bot message `CLICKCLACK_MEDIA_PROOF_OK`, quoting the exact image-bearing source message.

This exercises actual ClickClack event consumption, authenticated attachment retrieval, managed-media staging, multimodal agent input, provider dispatch, and ClickClack reply delivery. All resources and credentials were synthetic and local.

### Checks

- complete ClickClack extension suite: 36 files, 412 tests passed
- bot-loop regressions cover second-notice suppression for text-only model mode and permanent media-limit rejection, plus same-message notice retry accounting
- extension production typecheck passed
- extension test typecheck passed
- extension lint passed
- focused formatting check and `git diff --check` passed
- prior focused plugin-SDK boundary check passed; the current package-boundary run compiled ClickClack successfully before the unrelated all-plugin sweep was stopped to avoid redundant work
- companion server HTTP proof verified atomic create, exact nonce replay without a duplicate event, and rollback on invalid upload

The production incident showed `message.created` arriving about 68 ms before the attachment update; the authoritative message contained no attachment at the moment the connector dispatched the turn. The server-side companion PR removes that producer race, while this PR ensures attachments present on the authoritative message are delivered to the agent.

### DIFF
--- a/docs/channels/clickclack.md
+++ b/docs/channels/clickclack.md
@@ -300,7 +300,7 @@ safety bound can omit an older active thread.
 ## Reply modes
 
 - `replyMode: "agent"` (default) dispatches inbound messages through the normal agent pipeline, including session recording and tool policy.
-- `replyMode: "model"` skips the agent pipeline and uses the plugin runtime's `llm.complete` for direct bot replies, optionally shaped by `model` and `systemPrompt`. The selected provider and model own the completion budget.
+- `replyMode: "model"` skips the agent pipeline and uses the plugin runtime's `llm.complete` for direct text-only bot replies, optionally shaped by `model` and `systemPrompt`. Because that completion API has no media input, attachment messages receive a visible unsupported-media notice and never gain agent or tool authority. The selected provider and model own the completion budget.
 
 Both modes honor `responsePrefix` at the channel or account level. Account
 values win, including `""` to disable an inherited prefix. Use `"auto"` for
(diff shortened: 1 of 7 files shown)


sercada:
@clawsweeper re-review

Issue #136183, opened 2026-09-02
[Bug]: Command executor hangs when spawning ssh — SIGTERM while waiting for server banner (regression in 2026.8.1, persists in 2026.8.2)


manumadrigal09:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

ssh spawned by the command executor never completes the protocol banner exchange. The client sends its version string and then hangs until the executor kills it with SIGTERM. The exact same command succeeds from an interactive shell on the same machine, as the same user, at the same moment.

### Steps to reproduce

From a shell: ssh -o BatchMode=yes -i ~/.ssh/key user@host 'echo OK' → returns OK in ~1s.
Have an agent run the identical command through the executor → Command aborted by signal SIGTERM after 15s.
Run both simultaneously: the shell one succeeds while the executor one hangs.

Debug output from the failing run — last lines of ssh -vvv:

debug1: Connection established.
debug3: timeout: 15000 ms remain after connect
debug1: loaded pubkey from .../key: ED25519 SHA256:...
debug1: identity file .../key type 2
debug1: Local version string SSH-2.0-OpenSSH_10.2
Command aborted by signal SIGTERM

It never reaches Remote protocol version. Server-side auth.log shows Connection closed by <ip> [preauth] — TCP established, no protocol progress.

What we ruled out

Network/server: nc -G 5 -vz host 22 → succeeded! and curl -sI https://host → 200 OK, both from the same executor process.
Credentials/HOME: agent reports same $HOME, same user, key and known_hosts present and readable.
Interactive prompts: fails with BatchMode=yes, with -T, and with < /dev/null.
Timeout too short: successful runs take 1.1s; SIGTERM arrives at 15s. Process makes zero progress, so a longer timeout only delays the kill.
Shell wrapper: fails both directly and wrapped in a shell script; bash -c '...' succeeds from a terminal.

### Expected behavior

Prior to 2026.8.1, the agent's command executor spawned `ssh` successfully and returned remote command output. Monitoring jobs relying on SSH ran reliably for months.

### Actual behavior

[...]

PR #141463, opened 2026-09-07
fix(process): do not kill roots on lineage EOF

jjjhenriksen:
## What Problem This Solves

Fixes an issue where service-managed exec commands were killed after closing inherited file descriptors. Bare OpenSSH clients could receive SIGTERM before completing a connection or command, even though their root process was still alive.

Related: #136183

## Why This Change Was Made

The existing lineage observer now waits for the direct root's exit before reclaiming remaining descendants. It keeps the existing output-drain, cancellation, process-group cleanup and result owners. This removes the short observation timer, the separate root-exit cleanup callback, and the single-use policy module.

The contributor's real-process regressions remain. The corrected delayed-root-exit case now has a direct failing-without-callback and passing-with-callback receipt, followed by a pass with the simpler observer.

## User Impact

Commands that close inherited descriptors can finish and return their real exit status. Descendants that retain output after the root exits are still cleaned up. Explicit cancellation and configured timeouts keep their existing behavior.

## Evidence

Matched native Linux proof used a compiled service-managed Gateway, authenticated CLI `chat.send`, the documented `/bash` command, delivered `chat.history`, and a task-owned loopback SSH server with strict host-key checking. Process snapshots confirmed that the actual Node/OpenSSH root ran under the compiled relay and anchor.

| Command through `/bash` | Pinned main | Repaired candidate |
| --- | --- | --- |
| Root closes fd3, waits, prints `SURVIVED` | SIGTERM | `SURVIVED`, exit0 |
| Bare SSH with delayed remote output | SIGTERM | `SSH_STARTED SSH_SURVIVED`, exit0 |
| Root or SSH command exits7 | — | Real exit7 retained |
| Root writes stdout and stderr | — | Both outputs delivered, exit0 |

- 148 focused and sibling tests passed; one Windows-specific case was skipped on Linux.
- The exact corrected delayed-root-exit regression timed out without the callback, then passed with the callback restored and with the final observer. A separate native process capture confirmed that the root had exited0 while the descendant and unsettled result remained until explicit harness cleanup in the negative control.
- Formatting, targeted lint, maximum-line/assertion ratchets and staged-content guard passed.
- Candidate proof uses tested tree `df30e161642dbd29394893c93d45668c83f343bf`. The build stamp remains `449a108c673c938910e0691d85eef024e478f639` because the build preceded the final commit; exported source, patch and worker/build-file receipts bind the tested bytes.
- Fresh source-blind acceptance exercised 39 public Gateway requests, including bare SSH, default and legacy SCP, SFTP with byte readback, authentic signals/codes, cooperative and stubborn descendants, silent descendants, output drainage, graceful stop and timeout. Internal API, concurrency and unexecuted-platform limits are kept separate from the CLI results and covered by the applicable owner tests/source/hosted checks.
- Fresh cumulative source review is clean at P0–P2. The earlier silent-descendant concern was checked against the full cleanup owner and new native/public silent-descendant evidence; no source change was needed. Exact-head hosted CI and final landing review remain required before merge.

The native proof establishes the shared POSIX lifecycle repair and the reported Linux route. It does not independently certify every macOS SSH-banner symptom. Earlier predecessor proof, fixture failures and review history remain preserved. Initial HTTP `/tools/invoke` probes returned404 because that surface does not construct generic exec; those are setup failures, not bug reproductions.

The timeout probe preserved an existing diagnostic limit: when the root result is unavailable after forced cleanup, `/bash poll` reports `unknown exit code`. The formatter, polling owner and relay host are unchanged from pinned main; the full observation and causal limits are retained as a separate follow-up. No final signal or success was inferred.


obviyus:
Thanks @jjjhenriksen!

**Proof:** Authenticated `/bash` commands through the compiled service-managed Gateway now preserve live roots and bare SSH. The loopback SSH/SCP/SFTP checks retained strict host-key verification and verified transferred bytes.

- **Bug:** Closing inherited descriptors caused premature SIGTERM while the command was still running.
- **Fix:** The existing lineage observer waits for the direct root's exit, then uses the existing bounded cleanup owner.
- **Validation:** 148 focused/sibling tests and 39 independent public requests, including nonzero codes, signals, output drainage, graceful stopping, and output-holding and silent descendants. The exact delayed-root-exit test fails without the prior callback and passes with the restored and simplified implementations. Fresh cumulative and final review are clean.

| Before | After |
| --- | --- |
| Descriptor-closing root and bare SSH receive SIGTERM | Commands finish with their real output and exit status; descendants are reclaimed |

The linked issue remains open because the Linux proof does not independently establish every reported macOS SSH-banner symptom. The existing unknown-exit diagnostic after forced timeout cleanup is retained as a separate follow-up; no unobserved signal is claimed.

Merged as `fb380b6a474b7cd882009f7a3b3485502d17ef3d`. Exact-head CI passed on [attempt 2](https://github.com/openclaw/openclaw/actions/runs/34215279688/attempts/2). Attempt 1 retains the dependency-download failures and an updater-CLI timeout; the debug retry passed without source or assertion changes, and the timeout cause remains unproven.

PR #141466, opened 2026-09-07
perf(plugins): reduce bundled activation imports

steipete:
## What Problem This Solves

`openclaw plugins enable` still eagerly loaded runtime diagnostics and plugin installation persistence even when bundled manifest metadata already supplied the selected plugin's slot information. This adds file reads and module initialization to cloud worker bootstrap, including warm image reuse where runtime download and npm installation are already skipped.

## Why This Change Was Made

Move the existing plugin/hook config admission checks and write-option facts into a lightweight shared owner, with the checks unchanged. Load runtime diagnostics only for external plugins that need runtime-only kind inspection. Both callers await slot selection, and the existing install or lifecycle owner is revalidated after the import before plugin code executes. Enablement restrictions, capability consent, exclusive slots, config include/CAS checks, and registry refresh retain their existing owners. No configuration, CLI flag, database schema, or persistence-semantic change.

## User Impact

Bundled plugin policy changes load less unrelated implementation code. Existing external plugin runtime-kind inspection, installation, capability consent, and config write checks keep their behavior. No cloud latency improvement is claimed.

## Evidence

- CLI policy boundary regression failed on unchanged main and passes with the repair; enable and disable persist through the normal config owner without importing runtime diagnostics or install persistence.
- 115 focused plugin tests and 383 CLI/chat/config-include sibling tests passed; 47 final owner/capability tests passed after the lease callback change.
- External runtime-only slot selection and owner revocation during the lazy import or between package entries are covered. Existing between-entry behavior passed on the base; the stronger during-import expectation failed on the base and passes on the candidate.
- Full changed gate passed, including core/core-test types, lint, zero runtime import cycles, and unchanged schema v16. Both baseline and candidate full builds passed.
- Prescribed affected Codex entrypoint import/RSS profile passed on the existing candidate build: 151.31 MB peak RSS (43.39 MB empty-process baseline); this is an import-memory measurement, not activation timing. No rebuild was needed.
- Independent P0–P2 review of the exact final patch: clean (0.95).

The observed warm cloud bootstrap plugin phase was 33.624 seconds with runtime download/install skipped. Local profiling is used to measure import and CPU work, not to attribute that whole cloud duration to EBS reads. The same-base packaged CLI loaded 2,150 modules before and 1,989 after (161 fewer); the installer, runtime diagnostics, and runtime loader are absent from the candidate import graph. Loaded module file sizes fell from 12.35 MB to 10.42 MB. Both enable commands exited successfully and the canonical CLI read back `enabled: true`. All observed child processes exited and were verified absent.

The single local instrumented comparison measured policy CPU 1.827 s before and 1.854 s after, so it does not establish a CPU speedup. Overall wall time was 4.715 s before and 3.565 s after, with uncontrolled host/OS-cache conditions. Candidate cloud timing remains pending the ordinary live worker flow.

Production delta is +10 lines after discounting the unchanged 222-line owner move; the growth carries the asynchronous inspection boundary and existing owner assertion. Existing exports at the heavy persistence owner are removed, without compatibility forwarding exports.

CI initially failed the unrelated Markdown test file's existing max-lines rule (1,011 counted lines). The branch was mechanically refreshed and its streaming/indented-code groups moved into a separate test file: 272 unchanged lines, no assertion or UI runtime changes. All 115 Markdown tests, the full changed UI gate, and independent P2 review (0.99) pass. The plugin patch retains its stable patch identity, and all 13 production files remain byte-identical to the packaged CLI proof candidate.


steipete:
Merged via squash.

- Prepared head SHA: [3ac3292395eb48a7b5ef2e6c6a3a786645622bd2](https://github.com/openclaw/openclaw/pull/141466/commits/3ac3292395eb48a7b5ef2e6c6a3a786645622bd2)
- Landed commit: [7a1f677e3cb5d4882c0cff0d76dd4af196c27c58](https://github.com/openclaw/openclaw/commit/7a1f677e3cb5d4882c0cff0d76dd4af196c27c58)

PR #141467, opened 2026-09-07
feat(beam): delete beamed sessions from the sidebar row menu

steipete:
## What Problem This Solves

Beamed sessions could not be removed from the Control UI. The sidebar row menu under **Beam** only offered "Open in OpenClaw" and "Open in terminal", so a beamed transcript stayed listed until its seven-day retention expired or the 500-entry catalog evicted it. Operators asked how to delete or archive a beam and there was no answer.

## Why This Change Was Made

The Gateway already exposes `sessions.catalog.archive` for external session catalogs, and the Codex catalog implements it, but the Beam catalog advertised `canArchive: false` and registered no hook, and no Control UI surface called the method at all.

- **Beam plugin:** the catalog now advertises `canArchive` and implements `archive` as a permanent delete of the stored row through the existing keyed plugin-state store. Unknown hosts and unknown ids fail with the same error shape the read path uses. No retention, bounds, schema, or protocol changes; the Gateway's `operator.write` authorization and parameter validation stay authoritative.
- **Control UI:** the external-session row menu gains a destructive **Delete** item, shown only when both the row's `canArchive` and the catalog's `archive` capability are true, so Claude Code and acpx rows are unchanged. Delete goes through the shared confirm dialog, calls `sessions.catalog.archive`, invalidates the catalog snapshot so an in-flight list response cannot re-show the deleted row, queues an immediate refresh behind any active request, and navigates to the plain chat route if the deleted session was open. Failures surface as a toast. Stale connection scopes abort the mutation, matching the sidebar's existing session-mutation guard.
- **Docs:** `docs/plugins/beam.md` documents deletion, that re-uploading the same `beamId` (manual skill or a still-active mirror on its next changed upload) recreates the row, and that existing continuations are unaffected.

The generic menu item also becomes the first UI surface for the Codex catalog's existing archive hook, which archives the local Codex rollout rather than erasing it. The confirm copy states that beamed sessions are deleted permanently while sessions kept by another tool are archived there and may be restorable, and asks the operator to make sure no other runner is using the session, matching the `confirmNoOtherRunner` contract.

## User Impact

Operators with `operator.write` can delete a beamed session from its sidebar row menu. Deletion is immediate and permanent. Rows from catalogs that cannot archive show no Delete item.

## Evidence

Before/after from `pnpm dev:ui:mock` with a synthetic Beam fixture (the "before" row is the same fixture with the archive capability off, which renders exactly as main does):

| Before | After |
| --- | --- |
| Beam row menu before: only Open in OpenClaw and Open in terminal | Beam row menu after: adds a red Delete item |

Confirm dialog names the session title:

Delete confirmation dialog for a beamed session

- `extensions/beam/src/beam.test.ts`: new regression against the real SQLite-backed keyed store covers wrong host, unknown id, delete removes the row from list and read, and re-upload recreates it. It failed on main for the intended reason (`archive` undefined, `canArchive` false). 14 passed.
- `ui/src/components/app-sidebar.test.ts` + `app-sidebar-catalog-menu.test.ts`: Delete item gated on row and catalog capability, cancel sends no RPC, confirm sends exact `sessions.catalog.archive` params then refreshes, navigation only when the deleted session is open. 347 passed.
- `pnpm ui:i18n:baseline`, `git diff --check`, oxfmt clean. `pnpm check:changed` passes every lane except the extension lint lane, which fails in this nested worktree on the known environmental "Declaration input escapes checkout" boundary; CI is authoritative for that lane.
- ClawSweeper's two P2 findings on the first head (stale in-flight catalog snapshot after delete; permanent-deletion wording for recoverable Codex archives) are fixed in the follow-up commit with a regression covering the list-starts, delete-completes, stale-list-arrives ordering.
- Codex autoreview on the final diff: no actionable findings.

Known gap: no live Gateway run; proof is unit and mock-Gateway UI. The native apps do not expose the catalog row menu and are unchanged.

### DIFF
--- a/docs/plugins/beam.md
+++ b/docs/plugins/beam.md
@@ -134,11 +134,19 @@ A continuation belongs to the authenticated operator who creates it. From then o
 
 User turns are attributed to the verified publisher of the current snapshot, using their current profile name and avatar, including merged profiles. Beam's upload format does not identify individual authors within a multi-user transcript. The uploader reference shares the snapshot's seven-day retention and is replaced on each upload. Shared-token uploads, failed profile resolution, and older snapshots without a recorded uploader display **User**; they never inherit the viewer's identity or a previous uploader's identity. Reupload an older snapshot through personal authentication to attribute it.
 
+### Delete
+
+Any operator with `operator.write` can delete a Beam from its sidebar row menu.
+Deletion is permanent and immediate. Re-uploading the same `beamId` recreates the
+row, whether through the manual skill or a still-active mirror's next upload.
+Mirrors skip unchanged snapshots, so recreation does not necessarily happen on
+the next poll. Deleting a Beam does not affect continuations already created from it.
+
 ## Security boundary
 
 Beam publication is not remote control.
 
-- Continuing makes an independent Gateway-owned session. Beam itself has no archive, terminal, tool, or node capability.
+- Continuing makes an independent Gateway-owned session. Beam itself has no filesystem, terminal, tool, or node capability.
 - It accepts text-only normalized transcript items, not HTML, scripts, archives, attachments, or server-fetched URLs.
 - The official skill removes raw tool results, reasoning, prompts, local paths, credentials, cookies, and auth material before upload.
 - The receiver treats every transcript as untrusted text. The first message in the Beam composer is the explicit operator action that copies it into a new session.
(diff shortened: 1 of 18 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f43640c630132a2e36f990fbcf7f1d4d8a85fb02](https://github.com/openclaw/openclaw/pull/141467/commits/f43640c630132a2e36f990fbcf7f1d4d8a85fb02)
- Landed commit: [893e59465a59841b2de6ce150ac32b8aff28571a](https://github.com/openclaw/openclaw/commit/893e59465a59841b2de6ce150ac32b8aff28571a)

PR #137022, opened 2026-09-03
fix(gateway): owner profile follow-ups

steipete:
## What Problem This Solves

Fixes an issue where administrators could merge the shared local owner profile with a personal profile or assign it an operator role, and where adding one identified person to an owner-only Gateway incorrectly activated multi-person catalog restrictions. The owner's Identity page also showed an empty Linked emails row and misleading GitHub retry instructions.

Related: #136819 (durable owner profiles for single-user Gateways).

## Why This Change Was Made

The shared owner is attribution and display identity only. This follow-up keeps merge and role rejection at the profile store, maps its typed errors to `INVALID_REQUEST`, and applies the existing owner exclusion to catalog counts and visibility. The reserved ID now comes from the public Gateway protocol contract, shared by the store and Control UI.

Three follow-ups:

1. Reject email links/merges into or out of the owner, including tombstoned targets, and reject operator-role assignment. An email attached by the old `linkEmail` path could later make GitHub sign-in attach an identity to the owner or merge it in either direction; the verified-identity writer now rejects that demonstrated path too.
2. Count personal identities only and treat attached owner attribution as no sharing identity in catalog visibility/cache keys. Owner plus one person stays single-user; multi-person visibility and role caps remain enforced.
3. Explain the GitHub-backed Cloudflare Access/Tailscale Serve sign-in path on the owner's Identity rows and omit Linked emails. Document the merge/role restrictions in the user model guide.

The first commit is behavior-neutral cleanup limited to the landed hunks: shorten the misleading personal-GitHub comment (`src/gateway/server-methods/github-personal-authorization.ts:33`) and consolidate two identical provider/owner test fixtures (`src/gateway/server-methods/users.test.ts:178`, used at `:196` and `:674`). No database schema, migration, configuration option, authorization grant, sandbox, recovery, budget, baseline, or `CLAUDE.md` changes.

[...]

PR #141468, opened 2026-09-07
fix(codex): show local native sessions to admins

goldmar:
## What Problem This Solves

Effective `operator.admin` callers can receive an empty Codex catalog on an isolated or named Gateway even when the Gateway process home contains valid native Codex CLI sessions. The provider still returns other connected local hosts and continuation cursors, which made this look like a visibility-projection failure.

A real isolated Gateway reproduction seeds a non-sensitive native `source=cli` rollout in its temporary `$HOME/.codex`, starts the bundled Codex app server, and connects authenticated admin and read-only clients. Before the fix, admin list and exact-title search cannot find the native row because the source home is removed before Codex calls `thread/list`; transcript read is blocked at the same boundary.

Related: #123438, #137022.

## Root Cause and Fix

Core catalog visibility already treats effective `operator.admin` as unrestricted. The row disappears earlier: isolated Gateway state makes `allowsProcessHomeSessionScan()` false, and Codex filters every `usesProcessHomeFallback` home before app-server enumeration.

This change adds a provider opt-in for admin process-home reads. Codex opts in, and the Gateway allows that fallback only for list/search/read requests from callers whose effective scopes satisfy `operator.admin`. Non-admin callers retain the existing process-home isolation and privacy projection.

The override is read-only. The Gateway carries its startup-time HOME-isolation decision into catalog action projection, so rows admitted only through the admin read exception advertise no continue, archive, or terminal actions. Continue, archive, terminal, and other mutation handlers retain their existing authority checks. Ordinarily configured Codex homes retain their existing actions. Paired-node enumeration is unchanged because the capability affects only the provider's local process-home source. Existing adopted-session creator checks, draft/incognito privacy, sharing roles, method scopes, and caller-partitioned catalog caching remain unchanged.

## User Impact

The requested product policy is that effective `operator.admin` users may read these Gateway-local native sessions while non-admin privacy remains unchanged.

Admins can list, exact-title search, and read Gateway-local native Codex sessions from the Gateway process home. Non-admin Gateway owners, ordinary profiles, and generic unprofiled operators gain no transcript access.

## Evidence

- Real isolated Gateway + bundled real Codex app-server boundary: admin list, exact-title search, and transcript read pass for a temporary native CLI rollout; a read-only caller cannot list or read it. The admin and non-admin calls occur inside the catalog cache share window, proving the row does not cross caller partitions.
- The same boundary proof verifies process-home rows admitted for admin reading advertise `canContinue=false`, `canArchive=false`, and `canOpenTerminal=false`.
- Rebased focused Gateway isolation and Codex listing/transcript suites: 27 tests pass. The broader focused suites passed earlier with 61 tests.
- The isolation tests prove opted-in admin list/read receive the process-home source while read-only callers remain denied. Admin continue/archive still receive `allowProcessHomeFallback=false`; configured homes remain actionable.
- Production core and extension typechecks, focused core/extension lint, formatting, diff checks, and a full repository build pass locally.
- Hosted CI is green on exact head `39f9097066f9ec21cc201 silhouette`.
- On the rebased head, `check:changed` passed conflict, max-lines, assertion, attribution, plugin-boundary, wrapper, package-patch, and temp guards. Its core tsgo graph stage matched the previously observed baseline-slow Gateway-root validation path and was stopped after a bounded silent interval; targeted production and Gateway-root typechecks pass, and hosted CI supplies the authoritative full-shard proof.
- Duplicate open issue/PR search was refreshed after rebasing and found no equivalent open fix.

No Gateway deployment, installation, restart, live configuration/session mutation, migration, or OCA change was performed.

Worked on by @goldmar


goldmar:
@clawsweeper re-review\n\nThe PR has been replaced with the corrected admin-only fix and a real isolated Gateway + bundled Codex app-server regression. The non-admin Gateway-owner authorization exception and its public audience callback were removed; non-admin behavior is unchanged.

goldmar:
@clawsweeper re-review\n\nHosted CI exposed that the real Gateway regression was placed in a compact unit shard that does not own a built runtime. The same proof now runs in the existing session-catalog E2E root; local E2E passes both tests, and no production behavior changed in this follow-up.

goldmar:
@clawsweeper re-review

The requested action-capability hardening is now on the rebased exact head. Gateway startup-time HOME isolation is carried into the Codex list projection, and native process-home rows admitted solely for admin reading now advertise `canContinue=false`, `canArchive=false`, and `canOpenTerminal=false`. Ordinary configured homes retain their existing actions. The real isolated Gateway + bundled Codex app-server E2E asserts the read-only action projection alongside admin list/search/read and non-admin denial.

PR #141487, opened 2026-09-07
refactor(tests): share complete CI scope expectations

steipete:
## What Problem This Solves

The changed-scope tests repeat the same ten-field expected lane maps across 22 independent path checks. That repetition makes the intended Node-only versus Node-and-smoke outcomes harder to scan and maintain.

## Why This Change Was Made

Share two complete, test-owned literal expectations within the existing suite. Every changed-path input, full-object assertion, case name, and additional check stays intact. The expectations are used only on the assertion side and do not derive from production values. Removes 184 net test lines without changing workflows or routing policy.

## User Impact

No user-visible behavior change. The existing positive and negative CI routing contracts remain covered with less repeated expected data.

## Evidence

- All six scope suites passed before and after: 247 tests, including native/platform routing and the real zero-install script boundary.
- Expanding the two constants reproduces the entire original test AST, including all 81 full `expect` argument expressions. All 22 constant references are exact `toEqual` expected arguments; both ten-field maps retain their property order and values.
- The same 247 case names and per-file order passed; cross-file reporter completion order differed normally.
- Full changed gate, diff-scoped deslop, and fresh P2 review passed. No production, workflow, timeout, or capacity changes; no runtime saving claimed.

### DIFF
--- a/src/scripts/ci-changed-scope.test.ts
+++ b/src/scripts/ci-changed-scope.test.ts
@@ -99,6 +99,32 @@ describe("parseArgs", () => {
 });
 
 describe("detectChangedScope", () => {
+  const expectedNodeOnlyScope = {
+    runNode: true,
+    runMacos: false,
+    runMacosNode: false,
+    runIosBuild: false,
+    runAndroid: false,
+    runWindows: false,
+    runSkillsPython: false,
+    runChangedSmoke: false,
+    runControlUiI18n: false,
+    runUiTests: false,
+  };
+
+  const expectedNodeAndChangedSmokeScope = {
+    runNode: true,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [59aa981d1a2a12c06a0762ff535e45757bf330f3](https://github.com/openclaw/openclaw/pull/141487/commits/59aa981d1a2a12c06a0762ff535e45757bf330f3)
- Landed commit: [bc79a198dc0de4b77f63c67c02cfc86400386e6e](https://github.com/openclaw/openclaw/commit/bc79a198dc0de4b77f63c67c02cfc86400386e6e)

PR #140674, opened 2026-09-07
refactor(plugins): release scoped resources after actual work

steipete:
## Split complete

All sixty-three focused replacements have landed with separate review, proportional checks, applicable runtime proof, and required CI. This aggregate remains closed and unmerged as the preserved historical reference.

- **Merged:** #141491 — Codex callback context.
- **Merged:** #141492 — Memory capability reads.
- **Merged:** #141688 — Cron result parser move.
- **Merged:** #141692 — Additive owned plugin inspection.
- **Merged:** #141695 — Transcript summary owner move.
- **Merged:** #141706 — Talk connection cleanup.
- **Merged:** #141711 — Prepared snapshot owner move (P1).
- **Merged:** #141728 — Prepared retention/auth invalidation move (P3).
- **Merged:** #141780 — CLI runtime plugin inspection ownership.
- **Merged:** #141810 — Catalog source preparation move (P2).
- **Merged:** #141819 — Chat plugin inspection ownership.
- **Merged:** #141824 — Gateway dedupe replay owner move.
- **Merged:** #141845 — Plugin doctor inspection ownership.
- **Merged:** #141858 — Workboard registration disposal.
- **Merged:** #141938 — Retain prepared runtimes through direct plugin completions.
- **Merged:** #142006 — Preserve search setup configuration types.
- **Merged:** #142197 — Retain prepared runtimes through actual transport callback and cancellation work.
- **Merged:** #142214 — Separate inspection authority from retained registration resources.
- **Merged:** #142286 — Keep shared Gateway state until actual active-registry cleanup finishes.
- **Merged:** #142388 — Dispose executable CLI plugin resources after actual invocation cleanup.
- **Merged:** #142423 — Release one-shot exporter resources after actual service cleanup.
- **Merged:** #142519 — Retain replaced registries through cleanup and reentrant publication.
- **Merged:** #142632 — Drain admitted MCP tool work before shutdown cleanup.
- **Merged:** #142636 — Retain registration work through final resource cleanup.
- **Merged:** #142787 — Join actual node execution cleanup through queued handoffs.
- **Merged:** #142738 — Own migration providers through operation completion and cleanup.
- **Merged:** #142814 — Finish admitted cleanup before automatic macOS CLI exit.
- **Merged:** #142834 — Keep Music providers alive through generation and result processing.
- **Merged:** #142843 — Keep Image providers alive through generation and result processing.
- **Merged:** #142838 — Give synchronous SDK provider lookups an explicit host lifetime.
[...]

PR #141491, opened 2026-09-07
fix(codex): preserve turn context on reused sessions

steipete:
Related: #140674

## What Problem This Solves

Callbacks on a reused Codex session can inherit an earlier attempt's async context. Later requests or notifications then run with stale run-local state. This is a two-file independent extraction from the larger cleanup campaign; it does not require that framework.

## Why This Change Was Made

Bind each route's callbacks when its attempt activates. The shared transport keeps its existing lifetime, while request and notification callbacks preserve their receivers, routing guards, and original returned promises. This follows Node's [documented event-listener context binding](https://nodejs.org/api/async_context.html#integrating-asyncresource-with-eventemitter).

## User Impact

Reused sessions keep callbacks associated with the current attempt. No tool schema, prompt, protocol, configuration, dependency, or host-floor change is included.

## Evidence

- Both registration paths failed on current main: the second generation observed the first generation's async context.
- 31 router/lifecycle tests passed, covering both paths, receivers, notifications, and released routes.
- The unchanged real three-turn Gateway/Codex delivery-cache test passed: three delivered messages, one native session, stable instruction/tool bytes. The case took 22.830 seconds; its maintained wrapper, including QA preparation, took 120.408 seconds.
- All selected local checks passed after one test-only `const` correction. Deslop and independent P0–P2 review are clean.
- Proof uses isolated synthetic state and a scripted provider through the real Gateway and native-session flow; it does not claim external-model compliance.

The branch was rebased onto `ca10e17c` to include the already-merged unrelated UI test split from #141466. The patch, checked runtime contracts, and dependency manifests are unchanged; the existing local and live evidence retains its original source pin. Native interrupt/unsubscribe behavior was also checked directly in upstream [turn processing](https://github.com/openai/codex/blob/ade0ccacf9182e665cf784de041c51a79e35bc31/codex-rs/app-server/src/request_processors/turn_processor.rs#L1555) and [thread processing](https://github.com/openai/codex/blob/ade0ccacf9182e665cf784de041c51a79e35bc31/codex-rs/app-server/src/request_processors/thread_processor.rs#L1011); this changes no wire contract.

### DIFF
--- a/extensions/codex/src/app-server/turn-router.test.ts
+++ b/extensions/codex/src/app-server/turn-router.test.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage, AsyncResource } from "node:async_hooks";
 import { embeddedAgentLog } from "openclaw/plugin-sdk/agent-harness-runtime";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import type { CodexAppServerClient } from "./client.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f71a2362263132a09f1718b424047da50746dfa7](https://github.com/openclaw/openclaw/pull/141491/commits/f71a2362263132a09f1718b424047da50746dfa7)
- Landed commit: [ad38464e53975154fcb6cac17f5e177abb06a329](https://github.com/openclaw/openclaw/commit/ad38464e53975154fcb6cac17f5e177abb06a329)

PR #141492, opened 2026-09-07
fix(memory): keep cleared plugin registries absent during lookup

steipete:
Related: #140674

## What Problem This Solves

Memory capability queries recreate an empty plugin registry after the registry has been cleared. During cleanup this can restore process state that shutdown has already removed.

## Why This Change Was Made

Use the existing non-creating registry lookup for capability reads, preserving registration/request/active precedence. Registration paths still create their registry as before. This independent two-file fix contains none of the larger resource-ownership framework.

## User Impact

Cleanup and cold memory queries leave an absent registry absent. Registered memory behavior, storage, and SDK contracts are unchanged.

## Evidence

- The new regression failed on original main because a cold query recreated the registry.
- 43 memory-state/runtime tests passed.
- Native SQLite probes passed on Node 24.20 and 26.8: the registered runtime closed exactly once, its file reopened read-only with preserved data, and subsequent cold cleanup/queries left the registry absent.
- Canonical changed checks, deslop, and independent P0–P2 review passed.

The branch was rebased onto `ca10e17c` to include the already-merged unrelated UI test split from #141466. The two-file patch, checked memory/runtime contracts, and dependency manifests are unchanged; the existing proof retains its original source pin.

### DIFF
--- a/src/plugins/memory-state.test.ts
+++ b/src/plugins/memory-state.test.ts
@@ -19,7 +19,11 @@ import {
   type MemoryPluginPublicArtifact,
 } from "./memory-state.test-fixtures.js";
 import { createEmptyPluginRegistry } from "./registry-empty.js";
-import { withPluginRegistrationContext } from "./runtime.js";
+import {
+  clearActivePluginRegistry,
+  getActivePluginRegistry,
+  withPluginRegistrationContext,
+} from "./runtime.js";
 import { createPluginRecord } from "./status.test-helpers.js";
 
 function createMemoryRuntime() {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [277b86162272be7f7c32b59c7760df34342ba636](https://github.com/openclaw/openclaw/pull/141492/commits/277b86162272be7f7c32b59c7760df34342ba636)
- Landed commit: [42ecdff416fe0ee6e2b1be0112b847d3aaeb76a1](https://github.com/openclaw/openclaw/commit/42ecdff416fe0ee6e2b1be0112b847d3aaeb76a1)

PR #141493, opened 2026-09-07
fix(scripts): preserve multibyte command output

steipete:
## What Problem This Solves

Resolves corrupted non-ASCII command output in package-candidate preparation, plugin memory profiling, and the bundled-plugin/Kitchen Sink smoke helpers when a pipe chunk ends inside a UTF-8 character. Valid filenames or diagnostics could contain replacement characters even though the child wrote valid text.

## Why This Change Was Made

Enable Node’s streaming UTF-8 decoder on stdout and stderr before capturing text in each affected owner, matching the existing compiler-boundary helper. The shared bounded-tail helper retains its character limits, truncation accounting and UTF-16 semantics. No new buffering abstraction or dependency is needed.

## User Impact

Maintainers receive intact multilingual command output and package metadata. Existing overflow failures, RSS extraction and child cleanup behavior remain unchanged.

## Evidence

A real child writing `café 🦞 終` in separate bytes produced replacement characters on both streams before the fix. The regression uses real Node pipes with a byte-splitting transform, so its chunk boundaries are deterministic and require no sleeps. It failed before on text alone while normal exit and RSS accounting matched.

With the repair, the regression and seven existing capture/overflow/RSS tests pass across five files. All 19 selected repository checks pass, and independent review through P2 found no actionable findings. Tooling LOC is unchanged; regression and fixture support add 48 test lines. This validates command-output collection, not a full smoke deployment or live provider flow.

### DIFF
--- a/scripts/e2e/kitchen-sink-rpc-walk.mts
+++ b/scripts/e2e/kitchen-sink-rpc-walk.mts
@@ -569,10 +569,10 @@ export function runCommand(
       );
       forceKillTimer.unref();
     }, resolvedTimeoutMs);
-    child.stdout?.on("data", (chunk) => {
+    child.stdout?.setEncoding("utf8").on("data", (chunk) => {
       stdout = appendBoundedTail(stdout, chunk, outputCaptureChars);
     });
-    child.stderr?.on("data", (chunk) => {
+    child.stderr?.setEncoding("utf8").on("data", (chunk) => {
       stderr = appendBoundedTail(stderr, chunk, outputCaptureChars);
     });
     child.on("error", (error) => {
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ffa6cad5949171de68f3faab0d39f4e7a9558139](https://github.com/openclaw/openclaw/pull/141493/commits/ffa6cad5949171de68f3faab0d39f4e7a9558139)
- Landed commit: [631267364649d1e47b331be20388bd8f80c313d3](https://github.com/openclaw/openclaw/commit/631267364649d1e47b331be20388bd8f80c313d3)

PR #141507, opened 2026-09-07
fix(sessions): retain SQLite write failure details in logs

steipete:
## What Problem This Solves

Fixes an issue where operators investigating failed SQLite session writes see an empty `error: {}` in structured file logs instead of the failure message and cause. An enumerable error code can survive alone, while the useful explanation disappears.

## Why This Change Was Made

The existing write owner now formats its caught error with the shared redaction-aware message/code formatter and caps the summary at 2,048 UTF-16 code units without splitting surrogate pairs. This fixes the producer's nested-Error serialization mismatch. The original thrown value, timing and store fields, FIFO behavior, and global logger/redactor remain unchanged. There are no database-schema, persistence, retention, or concurrency changes.

## User Impact

Future failed writes include a bounded, redacted summary with available cause and code details. This improves diagnosis; it does not recover details already lost from historical records or resolve the underlying write failures or stalls. The logging documentation describes the summary and truncation.

## Evidence

- Three real owner-to-file-logger regressions fail on the original source for code-only/empty error output. All six cases in the existing owner suite now pass, including standard and Worker-originated errors, cause/code retention, credential masking, UTF-safe truncation, original rethrow identity, and existing timing/FIFO behavior.
- Documentation MDX validation passed for 864 files. Independent final Codex P0–P2 review found no actionable findings. The full changed gate passed, including both typecheck lanes, lint, import-cycle checks, and repository guards.
- The [Linux Testbox boundary run](https://github.com/openclaw/openclaw/actions/runs/34155326332) used Node 24.19.0, normal isolated logging configuration, the actual write owner and file sink, and no logger mocks or overrides. The baseline genuinely exited 1 on the diagnostic-string assertion after its known `{}`/code-only loss was recorded. The candidate exited 0, preserving standard and actual Worker causes, `SQLITE_BUSY`, credential redaction, the 2,048-character ASCII cap, and the 2,047-character result when a surrogate pair crosses the cap.
- All five original-value identity checks and all five already-queued writer continuations passed on both Linux variants. Timing/store metadata survived; no database or Gateway was opened. The orchestrator required the exact expected baseline failure plus the candidate pass and exited 0; the lease is confirmed closed. Source hashes and all 16 retained artifacts were independently verified.

### DIFF
--- a/docs/gateway/logging.md
+++ b/docs/gateway/logging.md
@@ -51,6 +51,13 @@ openclaw logs --follow
 - Embedded-run `continue_normal` decisions log at `debug`; retry, profile-rotation, model-fallback, and error decisions remain warnings.
 - Trace logging also includes diagnostic timing summaries for selected hot paths, such as plugin tool factory preparation. See [/tools/plugin#slow-plugin-tool-setup](/tools/plugin#slow-plugin-tool-setup).
 
+### SQLite session writes
+
+Failed SQLite session writes include a bounded, redacted `error` summary in
+their structured file-log record, with cause and error-code details when
+available. Long summaries are truncated; the record retains its write timing
+and store fields.
+
 ### Slow cron list pages
 
 A cron list page taking at least one second emits `cron: slow list page` through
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3e3c7af1f40e1b8508f655353c8a5ebd5f579cbf](https://github.com/openclaw/openclaw/pull/141507/commits/3e3c7af1f40e1b8508f655353c8a5ebd5f579cbf)
- Landed commit: [79dcb093e50d521e5cdd5d0cfe967a9ef2d66698](https://github.com/openclaw/openclaw/commit/79dcb093e50d521e5cdd5d0cfe967a9ef2d66698)

PR #141513, opened 2026-09-07
fix(lmstudio): prevent embedding truncation through smaller instances

### LINKED ISSUE #141510: LM Studio embeddings silently truncate through a smaller loaded instance
## Problem

LM Studio memory embeddings can silently use a smaller loaded model instance than the configured context length. Text beyond that instance's limit is truncated even though a suitable larger instance is already loaded.

## Reproduction and observed behavior

On current main c20600ad9394a18aa5830e4d2cfb0f91c32390b8, using LM Studio llmster 0.0.23-1 and Nomic Embed Text v1.5 Q4_K_M:

1. Load instances of the same embedding model with context lengths 1,024 and 2,048.
2. Configure the LM Studio embedding provider for 2,048 tokens.
3. Embed two synthetic inputs sharing a long prefix but different suffixes, each measuring 1,206 tokens.

The daemon logs show input truncated from 1,206 to 1,024. OpenClaw returns identical vectors for the distinct suffixes and exactly matches the explicit 1,024-instance control. The explicit 2,048-instance control preserves the suffix difference (maximum component delta 0.023939330130815506). Removing the smaller instance makes the unchanged provider match the larger control.

## Expected behavior

Each embedding operation should address the prepared instance that satisfies its configured context, while keeping the canonical model key for provider and cache identity.

## Cause and scope

Embedding readiness prepares the appropriate model but discards its instance identifier. Requests subsequently use the canonical key, allowing LM Studio to choose the smaller instance. This affects single embeddings and native document/query batches. Existing cached vectors are not retroactively recomputed by correcting request routing.

steipete:
Closes #141510

## What Problem This Solves

Fixes an issue where memory embeddings silently lose trailing text when LM Studio has multiple instances of the same model loaded with different context lengths. A configured 2,048-token budget could still send the request through a 1,024-token instance.

## Why This Change Was Made

Embedding readiness already prepares a suitable instance. Each operation now carries that identifier into a request-local embedding client. The public provider and cache keep the canonical model identity, and eviction recovery continues to select the newly prepared instance.

## User Impact

New embedding requests use the prepared context capacity, including single requests and document/query batches. Existing cached vectors are not automatically recomputed. Preload opt-out, cancellation, and managed-service lifetime retain their existing behavior.

## Evidence

- Four extended HTTP lifecycle cases fail before the fix for wrong-instance output; all 58 focused and sibling tests pass afterward.
- Complete changed-code gate and whitespace checks pass.
- Real LM Studio llmster 0.0.23-1 with Nomic Embed Text v1.5 Q4_K_M, loaded at 1,024 and 2,048 tokens: daemon logs confirm baseline truncation from 1,206 to 1,024 tokens. Distinct suffixes produce identical vectors through the baseline but different vectors through the explicit 2,048-instance control.
- Patched production requests exactly match the 2,048-token control, retain suffix differences in single/document/query operations, and reload at the requested capacity after eviction. Preload opt-out control remains unchanged.
- [Baseline live run](https://github.com/openclaw/openclaw/actions/runs/34154843769) and [candidate live run](https://github.com/openclaw/openclaw/actions/runs/34155672988). Final source differs from the live-tested source only by an explicit return type and equivalent `return undefined` normalization; final focused tests and changed checks cover those bytes.
- All owned test processes and leases were stopped. No operator configuration, model service, or production index was changed.

### Stored-data compatibility

No stored-data model changes or migration are required. The diff changes only the LM Studio request owner, its tests, and provider documentation; database schema, cache key generation, and vector metadata are unchanged. The long-lived provider/client retain the canonical model key, which the HTTP lifecycle regressions and real-runtime identity observations verify before and after eviction. Only a temporary per-operation client receives the instance ID. Previously cached vectors remain readable and are not automatically recomputed. This addresses the automated compatibility prompt without introducing an unnecessary migration.

### DIFF
--- a/docs/providers/lmstudio.md
+++ b/docs/providers/lmstudio.md
@@ -153,8 +153,10 @@ enough context for the selected model budget. A newly loaded instance is address
 identifier returned by LM Studio. Your configured model reference and conversation model identity
 keep the canonical model key.
 
-With preload enabled, embedding requests also check that their model is loaded. This lets memory
-embeddings recover after model eviction even when LM Studio JIT loading is disabled.
+With preload enabled, embedding requests also check that their model is loaded and route to the
+instance prepared for the configured context length. This avoids truncating input through a smaller
+loaded instance and lets memory embeddings recover after model eviction even when LM Studio JIT
+loading is disabled. Embedding model and cache identity keep the canonical model key.
 
 ### Disabling preload
 
--- a/extensions/lmstudio/src/embedding-provider.integration.test.ts
+++ b/extensions/lmstudio/src/embedding-provider.integration.test.ts
@@ -104,7 +104,7 @@ describe("LM Studio embedding request headers", () => {
 });
 
 it.each(["single", "documents", "queries", "cancelled"] as const)(
-  "reloads evicted embedding models before %s requests while holding the service lease",
+  "routes %s embeddings to sufficient-context instances across eviction while holding the service lease",
   async (kind) => {
     vi.stubEnv("NO_PROXY", "127.0.0.1");
     vi.stubEnv("no_proxy", "127.0.0.1");
@@ -113,7 +113,8 @@ it.each(["single", "documents", "queries", "cancelled"] as const)(
     const loadStarted = new Promise<void>((resolve) => {
       signalLoadStarted = resolve;
     });
-    let loaded = false;
+    let loaded = true;
+    let instanceNumber = 2;
     let leases = 0;
     const requests: string[] = [];
     const server = createServer((request, response) => {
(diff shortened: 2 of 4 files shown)


steipete:
The data-model compatibility prompt is addressed in the PR body. This is a request-routing repair: no schema, cache-key, vector metadata, or stored record changes. The existing provider/client canonical identity remains stable; only a request-local client receives the prepared instance ID. HTTP lifecycle tests and real-runtime observations verify that identity before and after eviction. Existing cached vectors remain readable and are not recomputed. No migration is required.

The independent review and ClawSweeper source review both found no actionable code defects; exact-head CI is green.

steipete:
Merged via squash.

- Prepared head SHA: [440ac97cd391a674a6feed1e4e900d436704b272](https://github.com/openclaw/openclaw/pull/141513/commits/440ac97cd391a674a6feed1e4e900d436704b272)
- Landed commit: [a02a5399989cfd89feb70e4561e41418d963ed21](https://github.com/openclaw/openclaw/commit/a02a5399989cfd89feb70e4561e41418d963ed21)

PR #141515, opened 2026-09-07
fix(update): validate workspace plugins against the staged host

steipete:
Follow-up to #141175.

## What Problem This Solves

Workspace plugins can load successfully before an update but fail candidate validation when their dependencies are hoisted into an ancestor directory. Shared or symlinked module directories could also leave the candidate using the serving host, depending on configured plugin order.

## Why This Change Was Made

Record the actual module owners, copy those directories with their lookup relationships, and preserve private directory aliases. Derive host edges after owner discovery, then bind them directly to the staged candidate. Canonical temporary paths keep filesystem aliases consistent across copying and verification.

Use one exclusion policy for both pruning and expansion. Links requiring an excluded host package now fail promptly with guidance instead of repeatedly adding and removing the same owner. Reached module directories may include unused dependencies; unrelated repository files are excluded.

## User Impact

Candidate validation uses the staged host while serving plugin files and links remain unchanged. Hoisted dependencies, local shadowing, directory aliases, and optional-only installations retain their existing behavior.

## Evidence

- Real snapshot-worker and Node import regressions cover hoists, both shared-owner scan orders, staged-versus-serving host identity, source immutability, optional dependencies, and aliased module/temporary directories.
- The complete final owner/native-stage suite passed 47 tests on this exact head. The nontermination predecessor hit its 10-second test deadline; the repaired worker returned its explicit error in 754 ms.
- Full changed checks passed for the repair, with final affected test, type and lint checks rerun after policy unification. Independent P2 review is clean.
- A standalone Mac proof passes through the real `/var/folders` filesystem alias. Exact rebuilt-package qualification remains part of release validation; it is not claimed by these source-level checks.

### DIFF
--- a/src/infra/update-candidate-plugin-tree.ts
+++ b/src/infra/update-candidate-plugin-tree.ts
@@ -1,6 +1,9 @@
 import fs from "node:fs/promises";
 import path from "node:path";
+import { isRecord } from "@openclaw/normalization-core/record-coerce";
 import { resolvePathViaExistingAncestorSync } from "./boundary-path.js";
+import { tryReadJson } from "./json-files.js";
+import { parseRegistryNpmSpec } from "./npm-registry-spec.js";
 import { hasNodeErrorCode, isPathInside } from "./path-guards.js";
 import {
   readRuntimeModulesManifest,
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [631f79ca260f9787f13cc61cbe1ae87b261b8393](https://github.com/openclaw/openclaw/pull/141515/commits/631f79ca260f9787f13cc61cbe1ae87b261b8393)
- Landed commit: [fd3ae9af9c9a1cbc82e94f3556745ed68dcb86b2](https://github.com/openclaw/openclaw/commit/fd3ae9af9c9a1cbc82e94f3556745ed68dcb86b2)

PR #141517, opened 2026-09-07
refactor(reply): narrow pending dispatcher registration

steipete:
## What Problem This Solves

Reply-dispatch registration generated an ID and retained an idle-wait callback even though the registry’s only consumers read pending counts. The dispatcher already owns actual delivery and settlement waiting.

## Why This Change Was Made

Narrow registration to the live pending-count getter and its unregister function. Remove the unused ID, counter, callback and return wrapper. Each registration still has distinct object identity, and the shared registry keeps its existing close-only lifetime so restart checks retain pending work.

## User Impact

Reply delivery and Gateway restart behavior stay the same, with one smaller internal registration contract. This maintainer-requested cleanup removes 15 production lines without changing tests or public APIs.

## Evidence

The same 19 existing tests passed before and after across Gateway restart/reservation, projected reply dispatch and singleton lifecycle coverage. All 28 selected repository checks pass, including core/test types, lint, dead exports and import-cycle checks. Independent review through P2 found no actionable findings. No live Gateway/provider run or performance gain is claimed.

### DIFF
--- a/src/auto-reply/reply/dispatcher-registry.ts
+++ b/src/auto-reply/reply/dispatcher-registry.ts
@@ -5,38 +5,26 @@
 import { resolveGlobalSet } from "../../shared/global-singleton.js";
 
 type TrackedDispatcher = {
-  readonly id: string;
   readonly pending: () => number;
-  readonly waitForIdle: () => Promise<void>;
 };
 
 const activeDispatchers = resolveGlobalSet<TrackedDispatcher>(
   Symbol.for("openclaw.activeReplyDispatchers"),
   "close-only",
 );
-let nextId = 0;
 
 /**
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [498a9e202fbe0c971dbb0e41d9259d7b6054c07c](https://github.com/openclaw/openclaw/pull/141517/commits/498a9e202fbe0c971dbb0e41d9259d7b6054c07c)
- Landed commit: [0c55a0a1ad048788025a5561486f170f0122c7f5](https://github.com/openclaw/openclaw/commit/0c55a0a1ad048788025a5561486f170f0122c7f5)

PR #141518, opened 2026-09-07
fix(android): keep a stale operator hello from overwriting the handoff-issued device token

steipete:
## What Problem This Solves

Fixes an issue where Android users who reconnected with a saved setup code while an older stored operator token existed could lose operator access on the next reconnect. The runtime starts the operator session on the stored operator token at the same time as the node session redeems the bootstrap token. Both hellos wrote the operator token slot, so a delayed old operator hello could overwrite the fresher token that the bootstrap handoff had just stored. The Gateway had already rotated that token away, so the next operator reconnect failed with a device-token mismatch and the app needed re-pairing. This race predates #141013 and was recorded there as a follow-up.

## Why This Change Was Made

`DeviceAuthStore` now owns admission of role-token writes. A write derived from a stored token (a stored-token hello, a recovery recommit of a loaded entry, or a mismatch-driven clear of the token that was tried) is admitted only while the slot still holds exactly that token, compared and committed under the store's own lock. Freshly issued grants (bootstrap handoff tokens and device tokens issued to shared-token or password connects) keep writing unconditionally, so the saved-bootstrap recovery from #141013 is unchanged. A stale hello keeps its live socket; only its persistence is refused.

There are no retries, timeouts, or second stores. Persisted preference keys, the serialized metadata format, configuration options, and the wire protocol are unchanged. The retained Android operator-startup policy stays as documented: the operator session may still start on the old stored token while bootstrap is present.

## User Impact

Reconnecting with a saved setup code no longer risks replacing the freshly issued operator token with a stale one. Operator access survives the next reconnect and process relaunch. No visible UI changes.

## Contract Notes

`apps/android/GATEWAY_AUTH_POLICY.md` records the write-owner rule under "Android durable handoff" and keeps the operator-startup difference accurate. `DeviceAuthTokenStore.saveToken` gains an optional `replacesStoredToken` expectation and `clearToken` an optional `onlyIfToken` expectation; both default to unconditional. No compatibility alias, schema change, release, or Google Play upload is included.

## Evidence

Both wire regressions run two `GatewaySession` instances against one real `DeviceAuthStore` and sequence the race with server-side latches, not timing.

`bootstrapHandoff_staleStoredOperatorHelloCannotOverwriteFreshOperatorToken` on unchanged production code (the intended failure, at the operator-slot assertion after the delayed operator hello):

```text
GatewaySessionInvokeTest > bootstrapHandoff_staleStoredOperatorHelloCannotOverwriteFreshOperatorToken FAILED
    org.junit.ComparisonFailure: old-operator expected:<[new]-operator> but was:<[old]-operator>
```

`bootstrapHandoff_staleOperatorRetryMismatchCannotClearFreshOperatorToken` on unchanged production code (a delayed `AUTH_DEVICE_TOKEN_MISMATCH` cleared the fresh token):

```text
GatewaySessionInvokeTest > bootstrapHandoff_staleOperatorRetryMismatchCannotClearFreshOperatorToken FAILED
    java.lang.AssertionError: expected:<DeviceAuthEntry(token=new-operator, role=operator, scopes=[operator.read], ...)> but was:<null>
```

Both pass with the repair, for the re-issued and the rotated operator token cases. `DeviceAuthStoreTest.tokenDerivedMutationsRequireTheCurrentStoredToken` covers the owner directly.

| Proof | Result |
| --- | --- |
| `ANDROID_HOME=/opt/homebrew/share/android-commandlinetools node --import ./scripts/tsx.mjs scripts/run-android-gradle.mts :app:testPlayDebugUnitTest :app:testThirdPartyDebugUnitTest` | Play 2,936 and third-party 3,067 tests; zero failures, errors, or skips |
| Touched lifecycle suites per flavor | GatewayBootstrapAuthTest 55, GatewaySessionReconnectTest 42, GatewaySessionCustomHeadersTest 7, GatewaySocketPublicationTest 2, all passing |
| `pnpm android:i18n:check` | passed |
| `ANDROID_HOME=/opt/homebrew/share/android-commandlinetools node scripts/check-changed.mjs` | passed, including Android ktlint |
| `git diff --check` | clean |

Independent Codex autoreview (branch mode against origin/main, P1 threshold): clean.

No emulator or live-Gateway proof was run for this change; the unit-level wire proof covers the changed boundary, and the live-proof requirement recorded in the policy doc for the original handoff change is untouched.

### DIFF
--- a/apps/android/GATEWAY_AUTH_POLICY.md
+++ b/apps/android/GATEWAY_AUTH_POLICY.md
@@ -76,6 +76,12 @@ write result. The session records the final write result for each role in this
 hello; receipt alone and preexisting tokens do not establish fresh handoff.
 Both node and operator writes must succeed before bootstrap retirement.
 
+`DeviceAuthStore` admits a write derived from a stored token only while the
+slot still holds that token. Freshly issued grants write unconditionally.
+A stale hello keeps its socket, but its persistence is refused; mismatch clears
+and recovery recommits follow the same stored-token rule. The wire regression is
+`bootstrapHandoff_staleStoredOperatorHelloCannotOverwriteFreshOperatorToken`.
+
 [`SecurePrefs.kt`](app/src/main/java/ai/openclaw/app/SecurePrefs.kt) then commits
 the existing credential bundle with only the consumed bootstrap removed. Token
 and password fields are preserved. A failed commit restores the previous
@@ -130,8 +136,9 @@ would change other Android authentication paths, outside this fix's scope:
   Gateway reports `not-paired` and explicitly advises waiting. Swift's channel
   classifies pairing-required as nonrecoverable.
 - Android may start an operator session using an old stored operator token
-  while bootstrap is present. iOS waits for the bootstrap handoff before
-  starting stored-only operator access.
+  while bootstrap is present. The store fence prevents a delayed old operator
+  hello from replacing the handoff-issued token. iOS avoids the race by waiting
+  for the bootstrap handoff before starting stored-only operator access.
 
 ## Required implementation proof
 
--- a/apps/android/app/src/main/java/ai/openclaw/app/gateway/DeviceAuthStore.kt
+++ b/apps/android/app/src/main/java/ai/openclaw/app/gateway/DeviceAuthStore.kt
@@ -35,20 +35,26 @@ interface DeviceAuthTokenStore {
     role: String,
   ): String? = loadEntry(gatewayId, deviceId, role)?.token
 
-  /** Returns true only after the token and its scope metadata are durably committed. */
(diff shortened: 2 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6b5d05250cbbb74a72f6ccdcccbc9df54da92ade](https://github.com/openclaw/openclaw/pull/141518/commits/6b5d05250cbbb74a72f6ccdcccbc9df54da92ade)
- Landed commit: [11c2a46c18fb78a2f549d501f73b2ff10daf2846](https://github.com/openclaw/openclaw/commit/11c2a46c18fb78a2f549d501f73b2ff10daf2846)

PR #141485, opened 2026-09-07
test(ui): split streaming markdown tests to restore the core lint lane

steipete:
## What Problem This Solves

Resolves a problem where every pull request rebased onto current `main` fails the `check-lint-core-2` lane: `ui/src/components/markdown.test.ts` grew past the `max-lines` limit (1011 counted lines against a maximum of 1000) after #141369 and #141431 landed. No source behavior is involved; the lane is red for unrelated work.

## Why This Change Was Made

Move the `toStreamingMarkdownParts` and `indented Markdown source` describe blocks, which both exercise the streaming split, into a new `markdown-streaming-parts.test.ts` beside the existing streaming highlight tests. The remaining file drops to 902 lines and its now-unused imports are removed. Tests are moved verbatim; the new file carries only the imports and the small `htmlFragment` helper it needs.

## User Impact

None for users. Contributors get a green core lint lane back on `main`.

## Evidence

- `pnpm exec oxlint` on `markdown.test.ts` reproduced `eslint(max-lines): File has too many lines (1011)` on current `main`; both files lint clean after the split.
- `node scripts/run-vitest.mjs` on `markdown.test.ts` and `markdown-streaming-parts.test.ts`: all tests pass, same count as before the split.
- Observed on #141446 at head `1b7c4429ef6`, which does not touch `markdown.test.ts`.

PR #141519, opened 2026-09-07
test(ui): remove duplicate streaming Markdown coverage

steipete:
## What Problem This Solves

Two main-branch repairs created byte-identical streaming Markdown test suites under different filenames. #141466 moved the streaming and indented-source coverage into `markdown-streaming.test.ts`; #141485 then added the same content as `markdown-streaming-parts.test.ts`. Both match Git blob `c51834438a442fb5d8c80ff531bae3ec9677cc68` and are discovered by the same UI unit-test glob.

## Why This Change Was Made

Delete only the later redundant `markdown-streaming-parts.test.ts`. The canonical streaming file retains all 32 cases, assertions, helpers, mocks, and internal ordering. There are no explicit references to the redundant path and no separate isolation or coverage contract.

## User Impact

No product behavior changes. The UI test suite stops discovering the same coverage twice. This is supporting CI cleanup, not a measured performance optimization.

## Evidence

- Fresh parent `ca10e17c129eda1548de070010bc82d731d3dcc4` contains both identical blobs. The actual parent-relative change of #141485 (`18eabd4c`) adds only the redundant file because #141466 (`7a1f677e`) already performed the move.
- Whole-tree reference inspection found no explicit dependency on either test filename. The retained canonical file is unchanged and remains covered by the existing unit-test glob.
- The original max-lines failure is already fixed: `markdown.test.ts` is 773 counted lines and the retained streaming suite is 246, below the unchanged 1,000-line test limit.
- Existing exact-head CI for the split passed: [#141485 CI](https://github.com/openclaw/openclaw/actions/runs/34155567885). No additional local runtime run is needed for deleting a byte-identical duplicate; this does not claim that an earlier unrelated test run executed this deletion.
- Diff checks passed and fresh P2 review found no actionable issue. The new PR's exact-head CI remains required before landing.

Production delta: zero. Test delta: one redundant 284-line file deleted. No baseline, timeout, assertion, or isolation changes.


steipete:
Merged via squash.

- Prepared head SHA: [21c6a1c0f6d204168c796945e9c946128f7a644e](https://github.com/openclaw/openclaw/pull/141519/commits/21c6a1c0f6d204168c796945e9c946128f7a644e)
- Landed commit: [5180c4569c5d7ada2864cafffe511b116935148a](https://github.com/openclaw/openclaw/commit/5180c4569c5d7ada2864cafffe511b116935148a)

PR #141521, opened 2026-09-07
fix(memory-lancedb): recall against the current prompt

steipete:
## What Problem This Solves

Fixes an issue where LanceDB auto-recall could search for memories about the previous user turn instead of the current question once a conversation had history. The prompt hook supplies the current request separately from prepared prior messages.

## Why This Change Was Made

Use the hook’s current prompt as the recall query, preserving media-note removal, whitespace normalization and the configured character bound. Remove the unused reverse-history reader and correct the existing hook fixture and documentation. Agent scope, tool policy, cooldown, timeouts, stored vectors and search/result behavior remain unchanged.

## User Impact

Automatic memory recall follows the current question when a conversation changes topic. No configuration or data migration is required.

## Evidence

The existing registered-hook test now matches production ordering: prior travel question/answer in history, current editor question in the prompt. Before the repair, the embedding-request assertion received the prior travel question. The identical case passes after the repair, including media-note removal, the 120-character limit and normal memory injection.

All 189 plugin tests and two existing current-user preparation boundary cases pass. All 25 selected repository checks pass, and independent review through P2 found no actionable findings. This is synthetic hook/embedding-request boundary proof; no live embedding provider or operator database was used. Production shrinks by 13 lines, tests by 4, and documentation by 1.

### DIFF
--- a/docs/plugins/memory-lancedb.md
+++ b/docs/plugins/memory-lancedb.md
@@ -211,10 +211,9 @@ local server returns context-length errors.
 
 `recallMaxChars` bounds the `before_prompt_build` auto-recall query, the
 `memory_recall` tool, the `memory_forget` query path, and `openclaw ltm search`.
-Auto-recall embeds the latest user message from the turn and falls back to the
-full prompt only when no user message is present, keeping channel metadata and
-large prompt blocks out of the embedding request. It also bounds each recalled
-item after prompt escaping before that text reaches the model.
+Auto-recall embeds the current turn's prompt after removing media attachment
+notes and normalizing whitespace. The same limit bounds each recalled item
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9be5adf1f8e8f191ea4cf6788dd7025836a4e512](https://github.com/openclaw/openclaw/pull/141521/commits/9be5adf1f8e8f191ea4cf6788dd7025836a4e512)
- Landed commit: [0f22cb304de7760de58ecc4d277de6a19d4c3fac](https://github.com/openclaw/openclaw/commit/0f22cb304de7760de58ecc4d277de6a19d4c3fac)

PR #141522, opened 2026-09-07
refactor(copilot): forward the prepared spawn workspace

steipete:
## What Problem This Solves

The Copilot tool bridge recomputes a spawn workspace when its caller omits one, even though every production attempt already prepares that decision from the original workspace. An obsolete private fixture kept the second calculation alive and expected the sandbox copy instead.

## Why This Change Was Made

Forward the attempt's prepared value directly and require the private input property as `string | undefined`. An explicit `undefined` remains the completed decision when no alternate spawn workspace is needed. Remove the old calculation and its obsolete fixture; the existing test helper supplies the explicit property. No runtime shim or new helper is introduced.

The complete change removes 9 production code lines and 11 test code lines, including the three lines of fixture setup needed for the explicit input.

## User Impact

No intended behavior change. Workspace selection stays with the existing attempt preparation owner. Read-only and no-access sandboxes retain the original workspace for subagent inheritance; absent and read-write sandboxes retain their normal wiring. Public SDK, configuration, and sandbox policy are unchanged.

## Evidence

- Baseline: 261 passing focused tests. Final: all 260 remaining tests pass; the only removed case asserts the obsolete private omitted-input fallback.
- Five synthetic cases pass before and after through actual attempt preparation/execution and the real bridge: no workspace, null sandbox, read-write, read-only, and no-access. They verify property presence, the original/effective/spawn values, terminal success, and the SDK request's working-directory value. The SDK is a fixture and final tool construction is captured; this is not live sandbox or Copilot service proof.
- All 25 selected check stages pass, including production/test types, lint, dead exports, and runtime import cycles.
- Fresh independent P2 review is scoped-clean. The refresh onto repaired main preserves both source files and the complete reviewed patch byte-for-byte.

### DIFF
--- a/extensions/copilot/src/tool-bridge.test.ts
+++ b/extensions/copilot/src/tool-bridge.test.ts
@@ -31,9 +31,10 @@ type CopilotToolBridgeInput = Parameters<typeof createCopilotToolBridgeImpl>[0];
 type CopilotToolBridgeAttemptParams = NonNullable<CopilotToolBridgeInput["attemptParams"]>;
 type CopilotToolBridgeTestInput = Omit<
   CopilotToolBridgeInput,
-  "agentId" | "attemptParams" | "modelId" | "modelProvider" | "sessionId"
+  "agentId" | "attemptParams" | "modelId" | "modelProvider" | "sessionId" | "spawnWorkspaceDir"
 > &
   Partial<Pick<CopilotToolBridgeInput, "agentId" | "modelId" | "modelProvider" | "sessionId">> & {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d023c51e22c4233a6996aed085d916fd40deea1d](https://github.com/openclaw/openclaw/pull/141522/commits/d023c51e22c4233a6996aed085d916fd40deea1d)
- Landed commit: [89619426bea7f2bf63d33c6c07183a4e6d0e1f7d](https://github.com/openclaw/openclaw/commit/89619426bea7f2bf63d33c6c07183a4e6d0e1f7d)

PR #141525, opened 2026-09-07
perf(cli): avoid diagnostic imports in successful respawns

steipete:
## What Problem This Solves

Successful CLI respawns prepared an error writer after the child environment was already snapshotted. That unnecessarily loaded dotenv, provider/plugin metadata, and diagnostic modules in the parent before starting the child, which then loaded most of the same modules again.

## Why This Change Was Made

Defer that preparation until a spawn failure needs it. Explicit startup tracing keeps its existing eager dotenv and formatting order before spawn. The shared respawn owner waits for asynchronous diagnostics before exit 1, including a rejected formatter, while synchronous reporting, synchronous spawn exceptions, and signal/TTY behavior retain their existing semantics. Root-help failures now use the same diagnostic writer instead of duplicating its setup.

No new config, environment controls, CLI flags, dependencies, schema, or persistence behavior. Production delta is +7 lines: the existing process owner now retains asynchronous error reporting through settlement.

## User Impact

Successful untraced respawns avoid loading diagnostic dependencies in the parent. Failed starts retain formatted, redacted error output before exit; explicitly enabled startup tracing keeps its existing order.

## Evidence

- Regressions failed on unchanged main for both untraced entry paths and for premature exit before asynchronous diagnostics settled. All 103 focused tests passed, including existing signal/escalation and synchronous error cases; final targeted retests passed.
- Core and core-test types, formatting, dead exports, lint, and the complete remaining changed-check guard sequence passed. Schema remains v16.
- Untouched baseline and candidate full builds share base `3fbd8f5739ba6d0622c8bbecad83ad54b90d5030`. Real packaged CLI enablement and canonical readback passed on both; observed processes exited and were verified absent.
- Parent module loads fell from 635 to 169; the child still loaded 2,150. Most removed parent imports are also needed by the child, so this is not a claim of 466 fewer unique cloud file reads.
- A fixed local comparison used one recorded warmup each and four balanced AB/BA pairs, fresh process/state/compile cache per run, and no profiler/import hook. Median wall time was 3.818 s before and 3.568 s after; recorded child CPU medians were 4.067 s and 3.818 s. All samples are retained. Host load/OS cache was uncontrolled, and this is not cloud/EBS proof.
- Real local failed-spawn probes recorded ENOENT with no child PID. Packaged and source compile-cache paths preserve empty stdout, JSON error output, dotenv-backed custom redaction, and parent trace-before-error ordering. A subsequent diagnostic import rejection exits 1 without raw fallback output or an unhandled rejection.
- Managed P0–P2 review was clean (0.94). The only later edit omitted a redundant default `void` type argument in a test; production bytes match the reviewed and built input, and final test/type/lint checks passed.

The failed-spawn probes use synthetic local executables and temporary state created by the normal config CLI. No live cloud state or credentials were involved. The changed CLI flow was live-tested locally. Candidate cloud latency remains unmeasured and no cloud speedup is claimed; the broader latest-main cloud flow is separate follow-up work.

### DIFF
--- a/src/entry.compile-cache.ts
+++ b/src/entry.compile-cache.ts
@@ -105,7 +105,7 @@ type OpenClawCompileCacheRespawnPlan = {
 };
 
 type OpenClawCompileCacheRespawnRuntime = RespawnChildRuntime & {
-  writeError: (message: string) => void;
+  writeError: (message: string) => void | Promise<void>;
 };
 
 function buildOpenClawCompileCacheRespawnPlan(params: {
@@ -145,7 +145,7 @@ function buildOpenClawCompileCacheRespawnPlan(params: {
 export async function respawnWithoutOpenClawCompileCacheIfNeeded(params: {
   currentFile: string;
   installRoot: string;
-  prepareWriteError?: () => Promise<(message: string) => void>;
+  prepareWriteError?: () => Promise<(message: string) => void | Promise<void>>;
 }): Promise<boolean> {
   const plan = buildOpenClawCompileCacheRespawnPlan({
     currentFile: params.currentFile,
@@ -176,7 +176,9 @@ function runOpenClawCompileCacheRespawnPlan(
     spawn,
     attachChildProcessBridge,
     exit: process.exit.bind(process) as (code?: number) => never,
-    writeError: (message: string) => process.stderr.write(message),
+    writeError: (message: string) => {
+      process.stderr.write(message);
+    },
   },
 ): ChildProcess {
   return runRespawnChildWithSignalBridge({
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0b814afaf27554e071c9efa436a5c217e59fbadc](https://github.com/openclaw/openclaw/pull/141525/commits/0b814afaf27554e071c9efa436a5c217e59fbadc)
- Landed commit: [8da3693041f3b1877cd482c272e8e409554dbe62](https://github.com/openclaw/openclaw/commit/8da3693041f3b1877cd482c272e8e409554dbe62)

PR #141526, opened 2026-09-07
test(ui): remove duplicate markdown streaming coverage

steipete:
## What Problem This Solves

Two files run the same 32 Markdown streaming and indentation cases. The original split in #141466 already moved the tests out of the oversized Markdown suite; #141485 subsequently added an identical copy under another filename.

## Why This Change Was Made

Delete only the later `markdown-streaming-parts.test.ts`. Its complete 284 lines are byte-identical to the retained `markdown-streaming.test.ts`, including imports, fixtures and assertions. No filename-specific routing or reference depends on the deleted copy.

## User Impact

No application behavior change. CI keeps every distinct streaming assertion while avoiding the duplicate execution. This removes 284 test lines; production and tooling are unchanged.

## Evidence

Exact file comparison confirms identical contents, and the retained original Markdown file stays below its line cap. One focused invocation passed all 32 retained streaming cases plus 69 ordinary Markdown cases. Fourteen unchanged security/ReDoS cases were explicitly filtered from that focused run. All 18 selected repository checks passed, including UI/core-test types, i18n, dead exports and full core lint. Independent review through P2 found no actionable findings.


steipete:
The same one-file deletion has landed through the earlier #141519 as 5180c4569c5d7ada2864cafffe511b116935148a, so this PR is now redundant.

I verified the actual merge removes only `markdown-streaming-parts.test.ts`. All 32 cases remain in the canonical streaming suite, including main's subsequent extraction of the identical `htmlFragment` helper. Exact-head CI and review passed for #141519. Closing this duplicate while preserving its branch and evidence.

PR #141529, opened 2026-09-07
refactor(tests): consolidate WebSocket ping fixtures

steipete:
## What Problem This Solves

Four WebSocket ping regressions use three private fake tasks that repeat every protocol member except pong delivery. The repeated transport scaffolding makes their actual timing and callback differences harder to see.

## Why This Change Was Made

Use one private fake with an immutable, explicit delayed/omitted/callback-list behavior. Preserve asynchronous delayed pongs, discarded omitted handlers, synchronous callback order, fresh task instances, and all four existing test bodies. Removes 58 net test lines with no production changes.

## User Impact

No user-visible behavior change. Missing-pong timeout, healthy delayed pong, duplicate success, and first-error handling remain independently covered. The healthy-pong case retains its 20 repetitions and every deadline stays unchanged.

## Evidence

- Static comparison verifies all six shared protocol members, all three behavior branches, and the complete remaining file/four test bodies after constructor normalization.
- SwiftFormat passed without changes. The normal mapped changed gate passed, including 1,132 Node/Vitest macOS script-fixture tests; these are not Swift runtime proof.
- Direct SwiftLint reports the same six existing findings before and after, shifted by the removed lines: file/type size, for-where, and three multiline argument findings. No new violation or suppression was introduced.
- Fresh P2 review passed. Exact-head macOS CI compiled the target in fresh scratch and ran the unfiltered OpenClawKit suite: 1,668 tests across 128 suites passed, including GatewayNodeSessionTests. The raw log records three ping passes but omits the missing-pong case and ten other test names; all four unconditional test registrations and bodies remain unchanged. Full-suite execution plus static equivalence supplies the native proof; no fourth individual console receipt is claimed.
- No native app, Swift test, signing, or credential operation ran on the operator desktop. No runtime improvement is claimed.

### DIFF
--- a/apps/shared/OpenClawKit/Tests/OpenClawKitTests/GatewayNodeSessionTests.swift
+++ b/apps/shared/OpenClawKit/Tests/OpenClawKitTests/GatewayNodeSessionTests.swift
@@ -36,13 +36,17 @@ private actor StringCapture {
     }
 }
 
-/// Delivers a pong asynchronously, well before the deadline, so a cancelled deadline
-/// task racing the gate would surface as a spurious timeout.
-private final class DelayedPongWebSocketTask: WebSocketTasking, @unchecked Sendable {
-    private let delay: Duration
+private final class PingWebSocketTask: WebSocketTasking, @unchecked Sendable {
+    enum Behavior {
+        case delayed(Duration)
+        case omitted
+        case callbacks([Error?])
+    }
 
-    init(delay: Duration) {
-        self.delay = delay
+    private let behavior: Behavior
+
+    init(_ behavior: Behavior) {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [02529772e92de6b219dec48c7595250f263d432e](https://github.com/openclaw/openclaw/pull/141529/commits/02529772e92de6b219dec48c7595250f263d432e)
- Landed commit: [9f77e922c8de45507e162d1b8dc4b758d20d9256](https://github.com/openclaw/openclaw/commit/9f77e922c8de45507e162d1b8dc4b758d20d9256)

PR #141531, opened 2026-09-07
fix(setup): retain consent choices through roster reconciliation

obviyus:
Classic setup could discard an answered telemetry preference when reconciling an existing bare `main` agent roster. The merge compared the candidate against a baseline that already contained the new choice, so the preference was missing from saved configuration even though onboarding completed.

Use the retained configuration snapshot as the roster-merge baseline. This preserves pending choices while keeping concurrent saved settings and existing agent ownership.

Validation:
- The current-main owner regression failed because the saved telemetry choice was absent.
- Four targeted checks pass: decline, accept, concurrent Gateway settings, and existing default-agent ownership.
- One isolated source-mode onboarding run saved the declined preference and its timestamp while preserving authored model rows, aliases, primary/fallback and the synthetic auth profile. No inference or repeated-prompt claim.
- Focused formatting, lint, boundary guards, ratchets and `git diff --check` pass. Local build and typecheck were not run.

### DIFF
--- a/src/wizard/setup.default-agent.test.ts
+++ b/src/wizard/setup.default-agent.test.ts
@@ -108,7 +108,10 @@ vi.mock("../commands/onboard-helpers.js", () => ({
 }));
 
 vi.mock("../commands/onboard-skills.js", () => ({ setupSkills: mocks.setupSkills }));
-vi.mock("../config/config.js", () => ({ resolveGatewayPort: () => 18789 }));
+vi.mock("../config/config.js", () => ({
+  resolveGatewayPort: () => 18789,
+  readConfigFileSnapshot: mocks.readSnapshot,
+}));
 vi.mock("../config/logging.js", () => ({ logConfigUpdated: vi.fn() }));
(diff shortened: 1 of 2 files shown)


obviyus:
Landed as 9a4114bc4691bdb1293fd48ef1ed1b1790485d12.

- **Bug:** Classic setup could lose an answered telemetry preference when reconciling an existing bare main roster.
- **Fix:** Use the saved configuration snapshot as the merge baseline, preserving pending choices and concurrent settings.
- **Proof:** The current-main owner regression failed before the fix; all four targeted checks pass. A real isolated source-mode onboarding run now saves the declined answer and timestamp. Independent behavior acceptance and exact-head local review pass; [required CI is green](https://github.com/openclaw/openclaw/actions/runs/34158793930/job/101857645566).

| Before | After |
| --- | --- |
| Setup completed but the answered telemetry preference was absent. | The declined answer and timestamp remain saved through setup completion. |

No later prompt-suppression or vendor-inference observation is claimed. Earlier failed captures remain preserved.

PR #141532, opened 2026-09-07
fix(docker): avoid hanging on resource-limit diagnostics

steipete:
Related: #139428

## What Problem This Solves

Fixes an issue where Docker E2E resource-limit diagnostics can hang on Linux systems using uutils coreutils 0.8.0. After Docker emits a short resource-limit error, tail can wait for another FIFO writer instead of completing the diagnostic.

## Why This Change Was Made

Pass the capture FIFO directly to tail instead of redirecting it through stdin. This avoids the implementation's pathname-resolution/reopen path while preserving the 65,536-byte tail limit, full stderr forwarding, Docker exit status, both child waits, and temporary-directory cleanup.

The existing short-error fixture verifies the limit and one real FIFO operand without assuming its pathname. Its assertions and 15 adjacent resource assertions now fail explicitly on Bash 3.2, where a failed standalone `[[ … ]]` can otherwise be masked despite `set -e`.

## User Impact

Resource-limit failures finish with their intended diagnostic and exit status on the affected Linux utility implementation. No dependency, timeout, resource-limit, or command-line option changes are introduced. This correctness repair adds one production line and one test line; it carries no aggregate runtime-improvement claim.

## Evidence

On the same Linux VM with uutils 0.8.0, the original targeted case hit the 120-second watchdog (121.4 seconds observed), with stable pre-signal samples showing Bash waiting for tail and tail waiting for a FIFO partner. The patched case passed, followed by all 276 tests selected by the frozen Docker test owner. The comparison driver still returned 1 because its inventory check incorrectly expected two files: that source revision routes the suppression sibling through the unit-fast owner. That wrapper failure is retained rather than reported as a passing comparison driver.

The suppression sibling then passed all four original cases with zero skips under its unit-fast owner. The native run exited 0 and joined its children; source/tree hashes remained unchanged from the patched comparison.

Locally, five resource/package cases passed, the old-helper/new-guard control failed as intended, and 16 synthetic shell controls demonstrated that four regressions masked by the original assertions are rejected by explicit exits. The full mapped changed gate, shell syntax, formatting, deslop, and fresh independent P2 review passed. Source and C44-derived patch/tree hashes are bound in the retained proof. No real Docker daemon was needed for the synthetic fixture cases.

### DIFF
--- a/scripts/lib/docker-e2e-resource-diagnostics.sh
+++ b/scripts/lib/docker-e2e-resource-diagnostics.sh
@@ -121,7 +121,8 @@ docker_e2e_docker_run_with_resource_diagnostics() {
     return
   fi
 
-  "$tail_bin" -c 65536 <"$capture_fifo" >"$stderr_file" &
+  # Some tail implementations reopen named FIFOs passed through stdin and wait for a new writer.
+  "$tail_bin" -c 65536 "$capture_fifo" >"$stderr_file" &
   local tail_pid="$!"
   "$tee_bin" "$capture_fifo" <"$stderr_fifo" >&2 &
   local tee_pid="$!"
--- a/test/scripts/docker-build-helper.test.ts
+++ b/test/scripts/docker-build-helper.test.ts
@@ -1429,7 +1429,8 @@ mktemp() {
 }
 
 tail() {
-  printf "%s\\n" "$*" >"$TMPDIR/tail-seen"
+  [[ "$#" = "3" && "$1" = "-c" && "$2" = "65536" && -p "$3" ]] || return 1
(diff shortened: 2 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [63b27f3fdfe1096b06c193d0a016c3ec3b9370a6](https://github.com/openclaw/openclaw/pull/141532/commits/63b27f3fdfe1096b06c193d0a016c3ec3b9370a6)
- Landed commit: [0ad318398e906aa710af94d61231dfe65946c02d](https://github.com/openclaw/openclaw/commit/0ad318398e906aa710af94d61231dfe65946c02d)

PR #141535, opened 2026-09-07
fix(gateway): don't fail the whole turn on a skill-authoring run-replacement race

hartmark:
## What Problem This Solves

Opening the same "main" agent chat from a second device (mobile app) while a turn was already in flight on another device caused the *entire embedded agent run to fail* with a confusing internal error -- `Personal authoring cannot move to a replacement run. Send a fresh message.` -- even on turns that never used the skill-authoring feature.

## Why This Change Was Made

Root-caused live against a running Gateway. The exact sequence, from the journal:

```
19:36:44  [gateway] device pairing approved device=0b5b39db... role=operator
19:38:00  a dashboard session's turn starts (tool_execution_started)
19:38:53  "Personal authoring cannot move to a replacement run. Send a fresh message."
          "Embedded agent failed before reply: ..."
19:42:08  same error again, same session
```

`prepareGatewaySkillAuthoring`'s returned capability is bound to the run's `AdmittedRunContext` unconditionally in `onAdmittedRunContext`, on every human main-turn admission -- regardless of whether `skill_workshop` is ever invoked that turn. Its `bind()` deliberately throws if called a second time with a *different* context (a legitimate security guard against carrying stale personal-authoring authority across a replaced admission -- see `src/gateway/skill-library-authoring.ts`). But nothing caught that throw, so it propagated out of `onAdmittedRunContext` and failed the whole admission, killing the reply for a turn that had nothing to do with skill authoring.

## User Impact

- A run-replacement race (e.g. a second device attaching to the same in-flight agent turn) no longer kills the whole reply.
- The security boundary is unchanged: if `skill_workshop` is actually invoked later in that same (replaced) turn, its own `assertCurrent()` check still fails closed with the same authority-expired error -- narrowly, on that one tool call, instead of on every unrelated reply.
- No config surface, no protocol change, no weakening of the personal-authoring authority model -- only the caller's error handling around a capability most turns never touch.

## Evidence

- New regression test: `src/gateway/agent-turn/agent-run-execution-phase.owner.test.ts` -- "does not fail the whole run when a mid-admission run replacement rejects the skill-authoring bind". Fails on pre-fix code (uncaught throw), passes post-fix.
- `src/gateway/skill-library-authoring.test.ts`: 7/7 unchanged, confirming the actual authority semantics are untouched.
- `oxfmt --check` clean on both touched files.
- Live-reproduced on an operator's own deployment (journal excerpt above); this PR's fix has not yet been re-verified against a live two-device reproduction -- that's a planned follow-up, not blocking this narrow resilience fix.

https://claude.ai/code/session_01Vk2cBRQdErwRtPea8VuhG6


hartmark:
Rebased onto current `main` (was conflicting — main independently added `prepareGatewaySkillAuthoring`/`getAdmittedRunDelegatedAuthority` mock resets to the same shared `beforeEach` this PR's own tests extend; merged both sets of resets, kept every test from both sides). Full `agent-run-execution-phase.owner.test.ts` + `skill-library-authoring.test.ts` suite green (25/25), including this PR's own new regression.

@clawsweeper re-review

hartmark:
On the remaining ask ("after-fix runtime evidence... recovery in a running Gateway"): the existing regression test (`does not fail the whole run when a mid-admission run replacement rejects the skill-authoring bind`) is a mock-Gateway boundary harness that drives the real, unmocked `startAgentRunExecution` production function through the exact failure path this PR fixes (a rejected `prepareGatewaySkillAuthoring` bind during run replacement), only mocking the outer `dispatchAgentRunFromGateway`/task-registry seams around it — per this repo's own testing doctrine (`docs/help/testing/writing-tests.md` / root `AGENTS.md`: "an isolated mock-Gateway harness covering the changed path is valid boundary proof"), that already clears the required bar; live-Gateway proof is called out there as *stronger*, not required.

Standing up a real local Gateway process specifically to reproduce a two-device concurrent-chat race is a reasonable next step if a maintainer wants the stronger evidence before merge, but it's beyond what the project's own policy requires for this fix, so I'm not doing it unprompted as part of a routine pass. No code defect was found in this round (patch quality 4/6, no actionable findings) — happy to add the live trace on request.

@clawsweeper re-review

Issue #112747, opened 2026-07-22
[Bug]: Arcee OpenRouter onboarding stores unusable auth profile


steipete:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

The documented `arceeai-openrouter` onboarding flow stores an OpenRouter auth profile, but `arcee/trinity-large-thinking` runtime auth resolves only the `arcee` provider and cannot use the stored credential.

### Steps to reproduce

1. Start from an isolated OpenClaw home/state directory on current `main`.
2. Install the Arcee provider plugin from the current source checkout.
3. Run non-interactive onboarding with `--auth-choice arceeai-openrouter`, a valid OpenRouter API key, and `--secret-input-mode ref`.
4. Run `openclaw agent --local --agent main --session-key agent:main:arcee-live --model arcee/trinity-large-thinking --message 'Reply with exactly: OK' --thinking low --json`.
5. Observe `No API key found for provider "arcee"` even though onboarding succeeded and stored `openrouter:default`.
6. As a diagnostic only, add the same credential under `arcee:default` while preserving the OpenRouter-backed Arcee model config.
7. Observe the next ownership mismatch: `Unable to rematerialize arcee/trinity-large-thinking for its resolved auth profile.`

### Expected behavior

The documented Arcee-via-OpenRouter onboarding flow should make `arcee/trinity-large-thinking` immediately runnable through OpenRouter without requiring a duplicate credential, manual profile editing, or a different model ref.

### Actual behavior

Onboarding succeeds, writes the Arcee OpenRouter model config, and stores the credential as `openrouter:default`, but the runtime plan requests auth for provider `arcee` and rejects the stored profile. A manually duplicated `arcee` profile passes initial auth selection but causes credential-scoped model rematerialization to choose the wrong Arcee catalog route.

### OpenClaw version

Current `main` at `6bf03cd6741ca8` (2026-07-22)

### Operating system

macOS 27.0.0, arm64

### Install method

Source checkout (`pnpm openclaw`)

### Model

`arcee/trinity-large-thinking`

### Provider / routing chain

OpenClaw local agent → Arcee provider plugin → OpenRouter → `arcee-ai/trinity-large-thinking`

### Additional provider/model setup details

- Plugin install succeeds.
- `arceeai-openrouter` onboarding succeeds.
- The valid OpenRouter credential is injected through an environment SecretRef; no credential value is present in logs.
[...]

PR #141541, opened 2026-09-07
fix(arcee): preserve model rows and use the selected credentials

obviyus:
**Verification in progress:** This revision stages the installed-plugin identity repair for CI artifacts. Its source regression and caller checks pass. Packaged OpenClaw with separately installed Arcee still needs the captured picker-to-first-reply proof before final review and merge.

Related: #112747, #113097

## What this fixes

Arcee setup through OpenRouter could save a working account while the first reply failed with missing Arcee credentials. Registered setup also copied generated catalogs into config. Mixed-endpoint sessions could lose a healthy account pin, and wire model IDs could produce duplicate picker rows or false fallback notices. Large configured catalogs could block Gateway responses past the request deadline.

## Changes

- Registered setup saves connection settings and retains authored model rows. Explicit replace mode and public preset helpers keep their eager catalog behavior.
- Credential readiness, session creation, same-model pin preservation and execution use the selected model endpoint. Stored credential realms, explicit bindings, linked accounts, account order and plugin trust remain authoritative.
- Provider-owned logical identity governs catalog joins, authored metadata, aliases, default tags, terminal receipts and persisted fallback state. Real provider/model changes remain distinct.
- The shared configured-model matcher preserves an authored row's endpoint, headers and limits after a shortened picker ID is selected. Exact authored rows take precedence over equivalent wire spellings in both catalog metadata and execution. Unknown case and foreign-provider namespaces retain their existing behavior.
- Identity lookup uses the selected trusted installed-plugin metadata when the provider is outside the core package. Captured and explicitly empty metadata retain their authority, and loader failures remain visible.
- Catalog operations prepare identity sets within the existing visibility owner and use the existing nullable artifact loader for providers without a policy artifact. The same path, hardlink and module validation remain. No global cache, configuration field, store schema, credential copy or provider migration is added.

## Validation

- Unmodified main reproduced the OpenRouter-only credential failure. The candidate completes through the synthetic OpenRouter route with its stored profile. Direct, both-account, mixed-endpoint and unknown-endpoint controls retain their actual results.
- Authenticated Gateway before/after proof links a work account, creates a session, changes the default for future sessions and reaches the first reply. The baseline sends the other account; the candidate retains and sends the linked work account. Independent stored/received fingerprints distinguish the credentials. An incompatible OpenRouter session uses its own account. Normal terminal-client identity and ownership checks remain active.
- Public picker-to-session-to-reply proof preserves an authored OpenRouter endpoint, header, account and 32,768-token run budget. A separate competing-row control proves catalog metadata and execution both prefer the exact authored 65,536-token row. Config and stored credentials remain unchanged.
- Corrected public `models.list` returns three logical rows with authored metadata and alias/default/configured tags. Pin, later-session selection, false-notice cleanup and genuine fallback/recovery observations retain their original source and transport identities.
- Six compiled public-helper controls, both real interactive setup routes, stored-profile and flag/environment/stored precedence controls, and first-prompt Ctrl-C pass at their recorded scopes. The same source-blind verifier passed all seven original and four supplemental clauses. Later linked-account, authored-row, collision and catalog observations extend that evidence for the final source-aware review.
- The 400-model, 11-agent source baseline timed out at the unchanged 30-second request limit. Complete Gateway/client profiles identified repeated missing-artifact exceptions and catalog scans. The final source candidate returns all 4,400 available rows in 11 requests, with 18.75 seconds total request time and a 2.49-second maximum. It makes no provider request, preserves config and shuts down cleanly. The earlier nine-agent partial result and every failed trace remain recorded.
- Fifteen provider-policy checks, eleven changed/existing visibility checks and the earlier focused collision, missing-artifact, auth-rotation and prepared-selection controls pass at their exact inputs. Lint, pinned formatting, import-cycle and whitespace checks pass. The installed-source regression, captured/empty-owner controls and authored catalog/auth caller check also pass. The image-policy fixture supplies the nullable loader export; its four failing CI cases now pass. Packaged installation proof, fresh exact-head review and full required CI remain the landing gates.

Proof combines separately bound source executions, compiled runtime `621dcb063135e31ea0c41878bcd57792ee2f8771` from raw `94015ca066a802e19a44c9a540fe9fae23f84a12`, and retained historical fallback/recovery controls. Later source repairs are not part of that older build. The unrelated full-layout refusal remains recorded. Traffic is isolated and synthetic; no real vendor authentication, graphical banner, separate nondefault-agent onboarding or personal-enrollment policy is claimed. The source-blind report remains unchanged at its earlier scopes. Full original failures, incomplete attempts and source/build distinctions remain available to the reviewer.


obviyus:
landpr-approve-admin-merge
pr: 141541
head: b0c8a9cab8e8843a9c9daa12cf20dd4d92c36aeb
scope: owner-review

obviyus:
Superseded by #142898, verified merged as 8bc0bace796a14c5a8e3aeab02149ea06caa4485 after clean exact-head review and successful required CI. The replacement preserves the signed contribution history, scoped proof and earlier failures.

PR #141551, opened 2026-09-07
fix(msteams): preserve every progress reply

steipete:
## What Problem This Solves

Fixes an issue where Microsoft Teams users could lose earlier reply text when progress mode produced more than one final payload. Formatting the last payload replaced the accumulated stream text, even though delivery could be reported as successful.

## Why This Change Was Made

One native stream now owns its first final segment. Later replies use the existing block delivery path, and the dispatcher waits for the earlier stream close and any fallback send to settle before later output can overtake it. That shared settlement also prevents overlapping closes and preserves a Stop discovered during close.

The controller reuses the existing partial-mode suppression and segment fallback ownership. The dispatcher adds one in-flight settlement promise; production grows by 10 lines to cover that lifecycle.

## User Impact

Multi-part progress replies retain each part in order. Later partial replies remain whole after native finalization, and cancellation continues to suppress later text and attachments.

## Evidence

- Before the repair, the pinned Microsoft Teams SDK with a synthetic localhost transport lost the first of two final texts. A second regression through the real core dispatcher reproduced the incorrect delivery result. Both original cases pass after the repair.
- All 134 cases in the three affected owner suites passed. Eight affected settlement/cancellation cases passed again after test typing cleanup.
- All 25 selected changed checks passed across recorded resumes. Initial test TypeScript/lint errors were corrected before the final checks passed.
- Independent managed review completed with no actionable P0–P2 findings. The pinned Teams SDK source was included in that review context.
- CI exposed two trace expectations for the old ordering. Both were regenerated after source and independent review; all four exact-order trace scenarios and 24 additional selected checks passed. The follow-up changes only trace fixtures and their explanation.

The SDK test exercises the actual stream implementation and HTTP wire boundary; it does not send to a real Teams tenant.

The installed `@microsoft/teams.apps@2.0.15` HttpStream implementation was inspected directly. Its string emits accumulate text, while final formatting can replace that buffer. The one-segment owner and ordered settlement were checked against that implementation and exercised through the real SDK HTTP fixture.

### DIFF
--- a/extensions/msteams/src/__traces__/stream-failure-redeliver-full.trace.jsonl
+++ b/extensions/msteams/src/__traces__/stream-failure-redeliver-full.trace.jsonl
@@ -7,7 +7,6 @@
 {"seq":7,"at":600,"dir":"in","kind":"block-final","data":{"text":"Deploy status: build is green."}}
 {"seq":8,"at":600,"dir":"in","kind":"partial","data":{"text":"Rolling out to production now."}}
 {"seq":9,"at":900,"dir":"in","kind":"final","data":{"text":"Deploy status: build is green.\n\nRolling out to production now."}}
-{"seq":10,"at":900,"dir":"in","kind":"idle"}
-{"seq":11,"at":900,"dir":"out","kind":"context.sendActivity","data":{"payload":{"channelData":{"feedbackLoopEnabled":true},"entities":[{"@context":"https://schema.org","@id":"","@type":"Message","additionalType":["AIGeneratedContent"],"type":"https://schema.org/Message"}],"text":"Deploy status: build is green.","type":"message"},"result":{"id":"activity-2"},"target":"a:1dm-trace-conversation"}}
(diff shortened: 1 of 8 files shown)


steipete:
The SDK-inspection concern is resolved in the maintainer review: I inspected the installed `@microsoft/teams.apps@2.0.15` HttpStream implementation directly, and supplied its complete source to the independent review. The actual SDK localhost test reproduced the lost first reply before the repair and preserves both replies afterward. The mitigation is the one-segment native owner plus ordered close/fallback settlement already implemented here.

CI also caught two old trace expectations. The follow-up updates their exact event order and removes only a guaranteed-failing write to a retired stream; all message payloads, targets and IDs remain unchanged. All four trace scenarios pass in normal assertion mode, including unchanged channel and cancellation controls. No production code changed in that follow-up. Exact-head CI is running again.

steipete:
Merged via squash.

- Prepared head SHA: [eb54bc77f5bb9ffe95ed13660924658e92a7446d](https://github.com/openclaw/openclaw/pull/141551/commits/eb54bc77f5bb9ffe95ed13660924658e92a7446d)
- Landed commit: [fb60915a3f9fb6f5961102df33eda5d5bc30941a](https://github.com/openclaw/openclaw/commit/fb60915a3f9fb6f5961102df33eda5d5bc30941a)

PR #141552, opened 2026-09-07
fix(baseten): keep generated models out of ordinary setup config

obviyus:
Related: #141531.

## What Problem This Solves

Fixes an issue where ordinary provider onboarding copied bundled catalog models into saved configuration, making generated inventory indistinguishable from authored settings.

## Why This Change Was Made

Registered setup uses a private callback that adds no catalog rows in ordinary mode. Keep the shared configuration applier, authored rows, aliases and caller-specific defaults. Explicit replace mode and the public configuration helper still seed the catalog.

## User Impact

Ordinary onboarding preserves authored configuration without saving generated inventory. Explicit replacement and public-helper behavior remain available.

Consumers: registered provider setup becomes passive about catalog rows; explicit replacement retains seeding.

## Evidence

The public onboarding regression saved nine unwanted rows before the fix. Sixteen candidate tests and a plugin build passed. Source-mode setup and public-helper checks covered authored, repeated and replace modes. Real Gateway checks covered failure, empty results, recovery and authored-row retention without external inference.

### DIFF
--- a/docs/providers/baseten.md
+++ b/docs/providers/baseten.md
@@ -51,6 +51,8 @@ export BASETEN_API_KEY=...
 
     </CodeGroup>
 
+    Onboarding saves the connection settings without copying the generated catalog into your config. Existing model rows stay unchanged. If you use `models.mode: "replace"`, onboarding also adds the bundled catalog because that mode disables implicit discovery.
+
   </Step>
   <Step title="Verify the live catalog">
     ```bash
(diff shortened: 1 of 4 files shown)


obviyus:
Landed in f166d5c43df68dd366ea4984e8e1d95d1691d11d.

- Ordinary Baseten setup now saves zero generated model rows. It previously saved nine.
- Authored rows, aliases, defaults and auth settings remain intact. Replace mode and the public catalog helper retain catalog seeding.
- Independent behavior validation passed all seven clauses through CLI and Gateway checks with synthetic provider responses. Consent remembrance is covered by merged #141531.
- The retained 16 tests and plugin build passed. After the first CI run found two incomplete test contexts, all five affected fixture cases passed; the corrected head passed CI and local ClawSweeper review with no findings.
- The proof does not claim vendor inference or later wizard prompt suppression. Earlier failed setup, oracle and CI results remain in the retained evidence.

PR #120340, opened 2026-08-07
perf(hooks): reduce native relay startup memory and latency

sightingsstellar-beep:
Related: #119565, #91009. These incidents remain open; this PR makes a measured startup improvement without claiming to resolve incident-scale exhaustion.

## What Problem This Solves

Concurrent Codex native hooks pay for the general OpenClaw CLI startup graph in each short-lived relay process. This change reduces that repeated startup work while keeping ordinary hook delivery, output, deadlines and process cleanup intact.

## Why This Change Was Made

Prefer a dedicated relay executable built inside the existing unified runtime graph. This shares lazy Gateway dependencies and canonical SQL-schema inlining instead of emitting a second runtime graph. Use the existing lightweight error and scope owners, share the unchanged bounded output drain, and keep the existing state-version reader independent of updater publication policy.

Explicit executable/CLI overrides and older-package entry selection remain supported. The 512 KiB cold static graph budget, server exclusions and external-package allowance stay enforced. No schema, stored representation, publication policy, configuration or dependency version is changed by this PR.

## User Impact

Native hook subprocesses start with less memory and latency. No setup change is required. The existing general CLI path also benefits from the narrower imports.

## Evidence

[Secretless paired qualification](https://github.com/steipete/openclaw/actions/runs/34150208951): baseline `5f661669b2c2979f2a61e23d4addb3bdd4545469`, candidate `4a87fd3ce171f824fc04fcdf5b91f0dfb150fe98`, tooling `6a0013f2dc48b8bcc56c9651c444cc39e3bc41cc`. Both used Node 24.20.0, integrity-verified pnpm 12.3.4 and the canonical `qaRuntime` build.

The table reports the median of three alternating 20-child cohort measurements per arm. Each latency measurement is that cohort's median child latency; these are not pooled 60-child medians:

| Arm | Sampled aggregate peak RSS | Median child latency |
| --- | ---: | ---: |
| Baseline general CLI | 2,458,628 KiB | 3,914.88 ms |
| Candidate general CLI | 1,817,108 KiB | 2,112.13 ms |
| Candidate dedicated relay | 1,704,588 KiB | 1,940.56 ms |

The complete candidate reduced sampled aggregate RSS by 30.67% and latency by 50.43% against baseline. The dedicated entry improved another 6.19% RSS and 8.12% latency compared with the same candidate's general CLI. These are bounded same-host observations, not unique/private-memory measurements, statistical guarantees or incident-scale Gateway responsiveness proof.

[...]

PR #141553, opened 2026-09-07
fix(worker): keep cloud bootstrap build closure self-contained

steipete:
## What Problem This Solves

Fixes an issue where creating a cloud session failed before allocating a machine after a successful fresh build. Node bootstrap rejected an incomplete import closure involving `dist/update-candidate-plugins-*.mjs` and the separately staged `dist/worker/worker.mjs`.

## Why This Change Was Made

Plugin projection imported pure path helpers from the snapshot owner that dynamically imports it. Moving the unchanged helpers into one dependency leaf removes that cycle and keeps the deploy worker in its single staged file. The existing unstaged-runtime rule now also checks the bundler's complete output before publication, so an escaped root chunk fails the build immediately.

## User Impact

Cloud sessions can package the slim node runtime from a fresh build again. Candidate state and plugin paths keep their existing semantics, and the worker's three public artifact paths and schema version 16 are preserved.

## Evidence

- Reproduced the exact failure on a clean full build of `7a1f677e3cb5d4882c0cff0d76dd4af196c27c58` using the canonical node-bootstrap artifact provider.
- Rechecked fresh main `0f22cb304de7760de58ecc4d277de6a19d4c3fac`, including the separate plugin-ownership repair in #141515: the actual worker build still emitted a root facade importing `worker/worker.mjs`. The fix preserves that plugin-ownership repair.
- The original actual worker build emitted an extra root facade. The real-build regression fails on that output; the repaired actual worker build emits only `worker/worker.mjs`.
- Started the candidate checkout with the original built output still present. A canonical full rebuild passed in 4m 49.3s, removed the stale facade through normal build cleanup, and passed node bootstrap packaging. Minimal node archive (no selected plugins): 23,185,354 bytes; SHA-256 `0c156b843ada6033006c448ff1008af3dd593ff8d50f01ae2d485c00efa82eda`.
- Canonical worker staging still contains exactly `worker.mjs`, `github-exec-launcher.mjs`, and `workspace-rsync-receiver.mjs`.
- 79 focused tests passed on the original proof; 88 worker/build-guard and updater/state/canary tests passed after integration onto current main, including workspace hoists and aliased state directories. Independent P2 review of the integrated patch is clean.
- Full seven-file changed gate passed on the integrated source: core and test type checks, script/root-test type checks, lint, dead exports, runtime import cycles, and architecture guards.
- Actual cloud flow passed on committed candidate `ac21783909a0674cca4c5f11aaf592f67e30f3cb`: dispatch completed in 261.855 s, a real tool turn completed in 57.378 s, and the session restored 13 files then wrote the 14th without a new Gateway checkout. State/agent schemas remained 16/19. Cleanup verified the lease released, environment destroyed, no attachments, all 33 original process identities absent, and the Gateway port closed.

Runtime source is +3 lines after moving the helpers; build tooling is +25 lines to enforce the existing artifact contract at publication; tests are +57 lines. No dependency or configuration option was added.

The final branch is mechanically integrated onto main `bd62b672d9ab8556b72fa1668392192bbc8dce6a`. It preserves #120340's schema-reader import relocation and new shared bootstrap checks, and takes the already-landed QA scenario repair from #141555 rather than retaining a duplicate metadata commit. Five of the seven repair files remain byte-identical to the cloud candidate; the state owner's non-import body is unchanged and the shared checker preserves all intervening main content. This is not a claim that the refreshed whole build is identical to the earlier cloud build.

The current-main premise was rechecked with the actual worker configuration: it still emits `worker/worker.mjs` plus an unstaged non-entry plugin facade. On the final branch, all 93 affected worker/build-guard and updater state/canary tests, the complete seven-file changed gate, and independent P2 review (0.98) pass. The selected real Chromium credential scenario and 20 catalog-owner tests also passed while diagnosing the upstream QA fix. A single final main smoke follows landing; no additional performance claim is made.

### DIFF
--- a/scripts/check-cli-bootstrap-imports.mts
+++ b/scripts/check-cli-bootstrap-imports.mts
@@ -12,6 +12,7 @@ import {
 } from "../src/shared/worker-bundle-hash.js";
 import { isDirectRunUrl } from "./lib/direct-run.mjs";
 import { readGatewayRunChunks } from "./lib/gateway-run-chunk-metadata.mts";
+import { isUnstagedWorkerDeployRuntimeArtifact } from "./lib/worker-deploy-build-plugin.mts";
 
 const DEFAULT_ENTRYPOINTS = ["dist/entry.js", "dist/cli/run-main.js"];
 const DEFAULT_NATIVE_HOOK_RELAY_ENTRYPOINT = "dist/native-hook-relay/entry.js";
@@ -483,7 +484,7 @@ export function collectWorkerDeployArtifactErrors(params: CliBootstrapCheckParam
         );
       } else if (entry.name === "node_modules") {
         errors.push("Worker deploy artifact must not contain materialized dependencies.");
-      } else if (/\.(?:mjs|node|wasm)$/u.test(entry.name)) {
+      } else if (isUnstagedWorkerDeployRuntimeArtifact(entry.name, artifactNames)) {
         errors.push(
           `Worker deploy artifact emits unstaged runtime asset ${path.relative(
             rootDir,
--- a/scripts/lib/worker-deploy-build-plugin.mts
+++ b/scripts/lib/worker-deploy-build-plugin.mts
@@ -40,6 +40,14 @@ export function resolveWorkerDeployGeneratorInputs(rootDir = process.cwd()) {
   ] as const;
 }
 
+/** The worker archive stages entry files only; emitted runtime auxiliaries have no owner. */
+export function isUnstagedWorkerDeployRuntimeArtifact(
+  fileName: string,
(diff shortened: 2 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1028ae17c96f3e8ca53ca195634b5beedcec32de](https://github.com/openclaw/openclaw/pull/141553/commits/1028ae17c96f3e8ca53ca195634b5beedcec32de)
- Landed commit: [88c37424d8c20f7164cfb844e13188d6b1442b68](https://github.com/openclaw/openclaw/commit/88c37424d8c20f7164cfb844e13188d6b1442b68)

PR #139023, opened 2026-09-05
fix(ui): keep Refresh stable during background session updates

PollyBot13:
Related: #138988

## What Problem This Solves

Fixes an issue where users viewing the Sessions page would see the manual Refresh button switch to Loading and become disabled whenever background session activity refreshed the roster.

## Why This Change Was Made

The Sessions page now tracks requests it initiates separately from the managed roster's background loading state. Background refreshes still update rows and retain the existing loading protections for roster operations, while only a page-owned request changes the manual Refresh control. Request scheduling, query ownership, and coalescing are intentionally unchanged.

## User Impact

The Refresh button remains stable and usable while session events update the roster in the background. Clicking Refresh still immediately shows Loading and disables the control until that request completes.

## Evidence

- The focused regression fails on the previous behavior when a managed background refresh publishes `loading: true`; it passes after the fix and verifies that an explicit Refresh still shows Loading.
- Maintainer validation after rebase: 73 focused unit tests and 2 Chromium/mock-Gateway browser scenarios passed; the UI bundle built successfully.
- Real Chromium with the bundled Control UI and mock Gateway protocol: 2 scenarios passed. The background regression now consistently selects the Sessions-page query (`includeUnknown: false`) for request deferral, counting, and waiting. Restoring the original loading binding fails with `Loading…` instead of `Refresh`; restoring the fix passes both scenarios.
- `node scripts/check-changed.mjs --base upstream/main`: passed all selected repository gates, including formatting, typechecks, dead exports, lint, and runtime import cycles.
- Maintainer review: fresh local autoreview and ClawSweeper review found no blocking findings in the Refresh change. The original button-binding regression was independently reproduced and verified fixed.

### Real-Gateway visual proof

Disposable loopback Gateways used synthetic session data only. Exact upstream main `41651036f27f672e17dfb0f2b81aa084f7715805` is compared with PR head `02630a3e77a978678786d3eb3698bb212b0b1c16`. The isolated browser delayed resolution of the real `sessions.list` response by 1.5 seconds so the transient state remained capturable; no Gateway response or roster payload was mocked. [Proof manifest](https://github.com/PollyBot13/openclaw/tree/proof-139023)

The displayed images are matching 942×480 crops of those existing captures, focused on the Sessions area to exclude the Discord invitation card as required by `ui/AGENTS.md`. No new runtime capture or image content alteration was performed.

Upstream main: background refresh incorrectly takes over the manual control:

Upstream main background refresh shows Loading

[...]

Issue #138988, opened 2026-09-05
Control UI session events fan out into sessions.list request storm and Refresh/Loading flicker


lakemike:
### Bug type

Performance / behavior regression

### Summary

On OpenClaw 2026.9.1, one Web Control UI connection can issue hundreds of `sessions.list` requests while the Sessions page is open and a session is active. The Refresh button rapidly alternates between Refresh and Loading because every managed-list request publishes its loading state to the page.

### Reproduction

1. Open Control UI and navigate to Sessions.
2. Keep a dashboard or another session actively producing session lifecycle events.
3. Watch the Refresh button flicker between Refresh and Loading.
4. Inspect Gateway logs for `sessions.list` responses grouped by WebSocket connection.

### Observed evidence

Environment:

- OpenClaw 2026.9.1, local build `f0d35b50421` based on `v2026.9.1`
- Node 26.8.1
- Fedora Linux
- Browser Control UI
- 50 rows displayed, one active session

One WebSocket connection, `conn=38c5af28…8c31`, produced:

- 641 successful `sessions.list` calls in five minutes
- 2.14 calls per second on average
- bursts up to 14 calls per second
- typical response latency 50–120 ms
- no `sessions.list` errors, SQLite lock warnings, event-loop degradation, or WebSocket reconnects in the measured window

The request rate dropped substantially after the active dashboard turn completed.

### Root cause in current source

The request amplification is event-driven:

- `ui/src/lib/sessions/index.ts` handles `sessions.changed` and terminal `session.message` events and calls `roster.scheduleEvent(...)`.
- `ui/src/lib/sessions/session-roster-refresh.ts` schedules the canonical roster and every matching or unscoped managed list.
- Every managed list owns an independent `createSessionEventRefreshCoordinator` with a 200 ms debounce and 1 s maximum wait.
- The Sessions page subscribes to a filtered managed list, while the sidebar and other visible surfaces can retain separate list subscriptions.
- One logical session event can therefore fan out into multiple concurrent `sessions.list` requests. The debounce coalesces each list independently, but does not coalesce requests across list owners.
- `primarySnapshotApplied` suppresses only the canonical refresh for eligible terminal snapshots; managed filtered lists still reload.
[...]

PR #141570, opened 2026-09-07
fix(ui): reduce repeated session-list reads during sustained activity

PollyBot13:
Related: #138988

## What Problem This Solves

Fixes repeated background session-list reads while sessions are busy.

## User Impact

Busy sessions generate less background traffic. Manual refresh and returning to a hidden page still catch up immediately. Different filtered lists remain independent; this is not a global request cap.

## Why This Change Was Made

The existing refresh coordinator spaces event-driven reads by one second per effective query, anchored to actual request starts, including delayed in-flight drains. Initial debounce, query ownership and final authoritative updates are preserved. A held background invalidation now uses the managed list's existing queued intent so a completed explicit refresh can satisfy it without an extra read; pending-RPC invalidations still receive their trailing refresh. This is separate from the Refresh-button state fix in #139023.

## Evidence

- September 14 CI refresh: rebased onto `10d67c8030`, incorporating upstream fixes #148107 (post-login catalog observation) and #148129 (configurable ClawHub fixture assertion). All seven PR commits remain patch-equivalent; all four modified production owners are byte-identical to the prior reviewed head. Scope remains 17 files, with no dependency, threshold, baseline, layout or live-runtime change.
- Reproduced the hosted held-admission regression (four reads instead of three), then repaired the owner without changing its existing assertions. 128 relevant tests passed before base refresh; 121 focused tests passed after it.
- Fresh production-UI/real disposable-Gateway Chromium proof passed: three synthetic 42-event bursts produced 8 hot-session reads, 6 many-session reads and zero unrelated-agent reads. Same-query gaps remained at least one second. Final labels and the invitation-free capture were inspected. This proves request pacing and settled content, not a visual redesign. Command: `node scripts/run-vitest.mjs run --config test/vitest/vitest.ui-e2e.config.ts ui/src/e2e/session-roster-request-rate.real-gateway.e2e.test.ts`.
- The cloud-dispatch test incorrectly equated immediate parent-detail reads with now-cooled child-list reads. It now asserts exactly one initial parent read plus one per placement update, retaining child-read, placement-label and send-order checks. Both browser tests pass; no parent throttling was added. Chat-resize browser and model-auth-refresh catalog regressions also pass locally after the base refresh.
- Current rebased-head validation: 59 focused session tests, all six scrolling browser cases, both model-login catalog cases, and 354 Docker-helper tests passed (two skipped). The current scrolling case passed locally; hosted recurrence is still checked separately. Paired Gateway/UI ciArtifacts build and performance budgets passed. Changed-file validation passed (core-test/UI/test-root typechecks, formatting and scoped lint). The normal dead-export launcher encountered missing registry timestamp metadata; all three equivalent scans passed using the already-cached exact Knip 6.32.2 version, with no installs or dependency changes. Diff checks passed. Fresh real-Gateway request-rate and both cloud-dispatch browser cases also passed on this rebased head.
- Independent native reviews of the held-admission repair, rebase equivalence and cloud-test contract found no actionable P0–P2 defects. CLI autoreview was attempted but its engine failed at startup on a model-cache schema error; that attempt provides no review coverage. A fresh independent rebase review also found no actionable defects. ClawSweeper revision 9 accepted prior head `ad89def11b`; refreshed hosted CI and new-head review remain pending.
- Original before/after synthetic proof reduced hot-session reads from 12 to 8 across two queries and rolling peak from 4/s to 2/s. This does not reproduce the originally reported 14/s workload or establish a global cap across arbitrary queries.

No live operator Gateway or user session data was used. Rollback: revert the focused coordinator/caller repair. AI-assisted: yes.


PollyBot13:
@clawsweeper re-review

PollyBot13:
@clawsweeper re-review

Head 558f56dd809f7f14f5d7f5dc06f3313035c7c9df rebases onto the new upstream managed-list owner, retaining actual-request-start cooldown accounting there. Fresh independent source review, affected tests, real-Gateway browser proof and the CI-matching fixed-identity performance gate passed. The PR body records current measurements and the remaining local infrastructure/type and unrelated workflow-fixture limitations. Fresh hosted CI is running; please review this head.

RomneyDa (dismissed):

PR #141572, opened 2026-09-07
fix(daemon): preserve service inspection identity and deadlines

steipete:
## What Problem This Solves

Fixes an issue where Linux status could report an installed node service as absent when the service manager was unavailable. Also fixes LaunchAgent enabled-state inspection exceeding its requested deadline, including the second inspection during update preparation.

## Why This Change Was Made

The node adapter now supplies its existing node-specific service identity to the inherited absence check. The LaunchAgent read and update preparation forward the existing timeout through their missing call sites. The repair keeps inspection context with its current owner and adds three production lines overall.

## User Impact

A readable installed node service stays visible even when its runtime cannot be queried. macOS enabled-state inspection respects the supplied deadline instead of delaying status or update preparation indefinitely.

## Evidence

Three regressions failed on unchanged production: the enabled-state read ignored its 100 ms deadline, node inspection returned installed=false for a node-only fixture, and update preparation dropped its 200 ms timeout. The same regression source passed after repair, along with two existing controls.

All 323 owner/sibling cases across six files passed, as did all 29 selected changed checks and the docs checks. Independent managed P0–P2 review found no actionable defect. Tests use an owned fake launchctl command and isolated service definitions; no live services were stopped or modified.

### DIFF
--- a/docs/cli/status.md
+++ b/docs/cli/status.md
@@ -91,6 +91,8 @@ and `openclaw memory status --deep`.
 
 - Overview includes Gateway + node host service install/runtime status when
   available, plus compact Gateway process uptime and host system uptime.
+- On Linux, a readable installed node service remains listed when the service
+  manager is unavailable; its runtime status stays unknown.
 - Overview includes update channel + git SHA (for source checkouts).
 - Update info surfaces in the Overview; if an update is available, status
   prints a hint to run `openclaw update` (see [Updating](/install/updating)).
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1b3224337e896a27e248ac29fcf5c6cff9a7700d](https://github.com/openclaw/openclaw/pull/141572/commits/1b3224337e896a27e248ac29fcf5c6cff9a7700d)
- Landed commit: [b08fbeade50734f918dd71c9c20f3038e4a317da](https://github.com/openclaw/openclaw/commit/b08fbeade50734f918dd71c9c20f3038e4a317da)

Issue #141191, opened 2026-09-07
System-expert inference probe rejects documented store SecretRef on `models.providers.*.apiKey` — "Couldn't sign in to zai / No API key found" while normal sessions work


wenbiyou:
## Version / Environment

- OpenClaw 2026.9.2 (3928bad), gateway via systemd user service
- Linux (WSL2, headless, no desktop keychain), node v22
- Custom provider `zai` (Zhipu bigmodel.cn, `api: openai-completions`) defined in `openclaw.json` `models.providers`, default agent model `zai/glm-5.3`
- Provider credential: `models.providers.zai.apiKey = {"source":"store","provider":"default","id":"ZAI_API_KEY"}` — created by the official Control UI protected-store import (2026-09-06, with `allowedHosts`); the config's plaintext string was replaced by this SecretRef around the same time (see Evidence 4 for the exact backup timeline)

## Summary

After the Control UI imports a provider API key into the protected store (converting `models.providers.<id>.apiKey` to a store SecretRef), the **system-expert inference gate** (`openclaw` tool / fresh `openclaw.chat` sessions) permanently fails with:

```
OpenClaw requires working inference: Couldn't sign in to zai. ... 
(No API key found for provider "zai". Auth store: <agentDir>/openclaw-agent.sqlite ...)
```

while **normal agent sessions on the same provider/model keep working** (the gateway model runtime resolves the store SecretRef via egress substitution). This split-brain state is very hard to debug because:

- the error names whichever agent store was attempted **last** in the fallback cascade (for us: `ui-designer`), not the root cause;
- the officially suggested remediation commands cannot satisfy the gate (see below).

## Evidence

1. **Docs declare the field supports store refs.** `docs/reference/secretref-credential-surface.md` lists `models.providers.*.apiKey` on the canonical SecretRef credential surface, and documents store refs (`{"source":"store",...}`, gateway-wide scope) as a valid form.
2. **Code rejects store refs on the probe path.** `resolveUsableCustomProviderApiKey` (dist `model-auth-provider-config-*.js:71`) returns `null` for any SecretRef whose source is not `env`. The probe chain `openclaw.chat` → `verifySystemAgentInferenceWithFallback` → `verifySetupInference` → `resolveApiKeyForProviderCore` hits this and throws `missing-provider-auth`.
[...]

PR #140392, opened 2026-09-06
fix(system-agent): show inference verification failure causes

ayaangazali:
## What Problem This Solves

Setup discarded useful plugin-loading and inference-owner errors, leaving operators with a generic failure instead of the underlying cause. Related: #134471.

## Why This Change Was Made

Return the cause through the existing diagnostic redactor and share activation's existing error coercion. Failure statuses, owner checks, callback order and cleanup stay unchanged. Production code is net −2 lines.

## User Impact

One-shot text and JSON errors retain useful detail alongside onboarding guidance. This repairs diagnostics; the original reporter's underlying Codex runtime failure remains undiagnosed and open. Preserves Ayaan Gazali's (@ayaangazali) contribution and original commit ancestry.

## Evidence

- [Installed CLI proof](https://github.com/steipete/openclaw/actions/runs/34154144688/job/101842366877) passed on source matching this head: two normal package builds and npm installs, with text and JSON tested before and after the fix. All four CLI cases reached the owned unreadable file and exited 1. Baseline output omitted the cause; repaired output included:

  ```text
  plugin runtime artifact file is not readable: pr140392-unreadable.txt
  ```

  Full raw output and syscall traces are retained. CLI traces contained no IPv4/IPv6 operations; configuration and package files stayed unchanged. All 18 commands joined, cleanup reported no errors, and owned scratch was removed. All 66 baseline/67 candidate source hashes and the complete repair delta were independently verified. This job passed; the overall workflow failed in unrelated rows.
- [Regression proof](https://github.com/steipete/openclaw/actions/runs/34090441998/job/101642662487) passed nine cases/eight receipts through the verifier and command function with synthetic dependencies. [Sibling proof](https://github.com/steipete/openclaw/actions/runs/34084842082/job/101626833548) passed 369 tests on their original attempts.
- [Static completion](https://github.com/steipete/openclaw/actions/runs/34103844505/job/101684262283) passed formatting, scoped lint, core-test typing and 30 final source hashes after narrow fixture corrections. The earlier regression/sibling jobs remain failed from later lint/type gates; their reused runtime evidence is explicitly source-bound.
- [Required CI](https://github.com/openclaw/openclaw/actions/runs/34152682301) passed on the current head, including the aggregate gate. Fresh autoreview and a separate committed-head review found no actionable P0–P2 issues.

[...]

PR #141586, opened 2026-09-07
fix(setup): New Agent fails with protected provider credentials

miguelbranco80:
Related: #141191

<details>

Allow edits from maintainers is enabled. The fork has no repository Actions
secrets or environments; maintainer branch edits are acceptable.

</details>

## What Problem This Solves

Fixes an issue where users opening **New Agent** or another fresh system-agent
setup conversation receive “No API key found” when their configured provider key
is held in the protected store, even though ordinary agent conversations can
use that same provider. This repairs the reproduced store-SecretRef probe failure
in #141191, not its separately unconfirmed custom-provider/profile interactions.

## Why This Change Was Made

Reuse the correctly scoped prepared runtime before setup projects its reserved
execution agent, carrying the paired canonical-source and default-materialized
configuration views from the authoritative read. Fresh reads compare complete
resolution-fact provenance structurally; ownership revalidation uses the
unprepared logical configuration and the same credential-selection policy as
the successful probe, while fallback-owner availability uses its prepared route.

This retains protected storage, sentinel/egress handling, candidate isolation,
and credential-rotation rejection. It does not copy keys into auth profiles,
change the env-only helper, add a plaintext fallback, or change a schema,
dependency, configuration option, or permission. This remains credential-sensitive
work for the inference/credential owners to review.

## User Impact

Operators can start setup conversations using their existing protected provider
key without moving it back into plaintext configuration. A changed provider
destination or SecretRef cannot borrow the active key, and a previously verified
session cannot silently switch credentials after rotation.

## Evidence

- **Before/after real Gateway proof:** on untouched `eb439c0d6e`, the authenticated
  New Agent RPC (`openclaw.chat`, `welcomeVariant: "new-agent"`) fails with
  `system_agent_inference_unavailable` / `No API key found for provider "fixture"`.
  The same acceptance gate passes with this repair. The fixture uses a protected
  SQLite-store key, the real embedded runner and Gateway handlers, and a local
  HTTP Responses provider that requires the correct synthetic bearer key; no
  runner, auth resolver, or Gateway handler is substituted.
- **Lifecycle proof:** the extended test verifies a successful follow-up turn,
  rotation through the real `secrets.store.set` RPC, rejection of the old session
  before another provider request, successful verification in a fresh session,
  and no key bytes in persisted config, public replies, or Gateway logs.
- **Local validation on `da3e876aaef5`:** 818 tests across 38 files passed, including repeated reads,
  alternate profiles, fallback owners, completion/activation, rotation during
  verification, staged versus saved-but-not-activated destination/reference
  changes, and absence of prepared runtime. Full build, changed-file checks, and
  the full aggregate check with `--include-test-types` passed on macOS with
  Node 24.19.0 and pinned pnpm 12.3.4. The patch-identical rebase onto
  `0bec2710f683` also preserves #140392 and passes its nine error-reporting tests.
- Repository autoreview completed. Its sole proposed canonical/logical-view
  change was checked independently and disproved by applying it to the existing
  revalidation test, which failed; restoring the intended view passed all 12
  boundary cases. No accepted actionable finding remains.

Focused reproduction:

```sh
node scripts/run-vitest.mjs run src/config/runtime-snapshot.test.ts src/system-agent/inference-route-runtime.test.ts
node scripts/run-vitest.mjs run test/system-agent-managed-auth.e2e.test.ts
node scripts/check-changed.mjs
node --import ./scripts/tsx.mjs scripts/check.mts --include-test-types
```

No external paid provider, rendered browser-click flow, Linux/Windows execution,
or Docker upgrade-survivor result is claimed by this local proof. The latter is
selected by CI because an existing update-repair test fixture must carry the new
route field. Hosted CI results belong to the exact PR head.

Worked on by:
- @IWhatsskill

### DIFF
--- a/src/commands/triage-target.test.ts
+++ b/src/commands/triage-target.test.ts
@@ -182,6 +182,7 @@ describe.skipIf(process.platform === "win32")("embedded triage installation targ
             agentId: "diagnostic",
             agentDir: state.statePath("agents", "diagnostic", "agent"),
             runConfig: config,
+            sourceConfig: config,
           },
           modelFallbacks: ["fixture/blocked", "blocked-provider/model", "fixture/fallback"],
           prompt: "Check the installation.",
@@ -230,6 +231,7 @@ describe.skipIf(process.platform === "win32")("embedded triage installation targ
           agentId: "diagnostic",
           agentDir: state.statePath("agents", "diagnostic", "agent"),
           runConfig: config,
+          sourceConfig: config,
         },
         modelFallbacks: [],
       });
--- a/src/config/runtime-snapshot.test.ts
+++ b/src/config/runtime-snapshot.test.ts
@@ -200,6 +200,52 @@ describe("runtime snapshot state", () => {
     },
   );
 
+  it("matches independently loaded config with equivalent resolution facts", () => {
+    const source = createProviderConfigFixture();
+    const freshRead = structuredClone(source);
+    const facts = () =>
+      createConfigResolutionFacts(
+        [],
+        new Map([["models.providers.openai.apiKey", "PROVIDER_KEY"]]),
+      );
+    setConfigResolutionFacts(source, facts());
+    setConfigResolutionFacts(freshRead, facts());
+    const runtime = createProviderConfigFixture("synthetic-runtime-key");
+    setRuntimeConfigSnapshot(runtime, source);
+
+    expect(getConfigResolutionFacts(freshRead)).not.toBe(getConfigResolutionFacts(source));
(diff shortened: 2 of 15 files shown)


VACInc (approved):
The protected-store credential failure in setup is valid, and reusing the matching prepared runtime addresses the correct boundary. Reviewed the source changes and coverage for changed destinations, secret references, and credential rotation; the CI gate is green. The author provides real Gateway/runner proof with a local provider fixture; I did not independently rerun that proof.

PR #141591, opened 2026-09-07
fix: preserve plugin test process limits with release flags

steipete:
## What Problem This Solves

Resolves a problem where Plugin Prerelease runs bypassed the existing plugin test process limits when forwarding retry and exact-file exclusion options. In [run 34159135860](https://github.com/openclaw/openclaw/actions/runs/34159135860/job/101857400767), the Codex batch put 232 files into one process and eventually exited 137.

## Why This Change Was Made

Preserve the existing sequential process limits for numeric per-test retry and exact-file exclusions. Codex keeps its 12-file limit. Unknown, malformed, watch, bail, report, shard, and broad-pattern options—including extglobs—keep their existing unsplit semantics. Heap sizes, runner classes, job parallelism and test inventory stay unchanged.

## User Impact

Release operators retain the established process boundaries when using the release command's retry and exclusion flags. The repair must be included in the selected release Code revision because Plugin Prerelease executes the planner from that checkout.

## Evidence

The exact release-command planning regression failed before the repair: one invocation instead of 20. After repair, all 232 selected files appear exactly once across 20 sequential invocations, each containing at most 12 files; the same retry and exclusion arguments reach Vitest.

- 179 owner tests pass, including installed Vitest fixtures that deliberately fail their first attempt, pass on retry, and preserve exclusions.
- Three sibling routing checks and the changed-file gate pass.
- Both extglob cases fail before the follow-up and pass afterward.
- Independent P2 reviews found no actionable findings.

The failed job retained no heap profile or RSS telemetry. This fixes the independently reproduced process-limit bypass; it does not claim proof of a specific application memory leak or a completed release rerun.

### DIFF
--- a/docs/ci/release-validation.md
+++ b/docs/ci/release-validation.md
@@ -426,6 +426,8 @@ The scheduled live/E2E workflow runs the full release-path Docker suite daily an
 
 ## Plugin Prerelease
 
+Plugin batch execution preserves existing process limits when the only forwarded options are numeric `--retry` and exact-file `--exclude` selections. Codex retains at most 12 files per sequential process, including release runs with those options. Watch mode, suite-wide bail or sharding, reports, broad exclusion patterns, and other options retain their single-process semantics.
+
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a16e704bc07874cf2ac52bfa14b711adfe82b6a9](https://github.com/openclaw/openclaw/pull/141591/commits/a16e704bc07874cf2ac52bfa14b711adfe82b6a9)
- Landed commit: [26c194333be5222ecfc25279c6417a309c5701e0](https://github.com/openclaw/openclaw/commit/26c194333be5222ecfc25279c6417a309c5701e0)

PR #141512, opened 2026-09-07
refactor(tests): isolate repair effects from Slack activation

steipete:
Related: #139428

## What Problem This Solves

The generic repair-effect revocation tests configure Slack even though they exercise requester authority around real exec/write effects, causing repeated Slack plugin activation during repair admission and rehearsal setup.

## Why This Change Was Made

Uses the established synthetic external requester and explicit bare owner in these four fixtures, omitting their incidental Slack configuration. The requester remains external, fresh policy checks still revoke subsequent effects, and both validating and verifying paths retain their original assertions, response gates, isolated state, child lifetimes, and source/compiled-runtime selection.

Separate handoff controls retain real Slack revoked/unchanged ownership coverage, alongside internal and channel-less controls. Requester-authority and fresh-worker tests retain configuration/load-failure handling and cross-process authority checks. No production authorization policy, registry mock, or activation flag changes.

## User Impact

No product behavior change. Removes one net test/support line and avoids incidental transport activation in generic repair fixtures.

## Evidence

- Focused baseline and candidate each passed 16 cases with identical names and results: four repair-effect cases, four ownership controls, three requester-authority cases, and five fresh-worker cases.
- Full lifecycle/command proof passed all 171 retained cases with no skips; the mapped changed gate, deslop, and independent P2 review passed.
- In the local focused comparison, the four repair cases decreased from 112.34 to 79.78 seconds (32.56 seconds, about 29%). Existing CLI registry-load spans decreased from 30.99 to 3.36 seconds in aggregate. Build time is excluded; these observations are not a global test-suite improvement claim, and no cross-version full-suite timing comparison is used.

PR #140458, opened 2026-09-06
fix(infra): start source workers outside the package directory

gaoanze888:
Closes #140416.

## What Problem This Solves

Fixes an issue where users running OpenClaw from source outside its package directory would see worker startup fail with `ERR_MODULE_NOT_FOUND` for `tsx`. Direct SQLite and update-state workers inherit or deliberately use an unrelated working directory. Reported by @mbundy; this PR retains @gaoanze888's contributor commits and owner-level repair.

## Why This Change Was Made

`src/infra/runtime-worker-url.ts` now resolves the TypeScript preload relative to its owning module before constructing child arguments. Node receives an absolute URL while the caller's working directory, JavaScript launches, and Bun behavior remain intact. The implementation uses native module resolution and removes the candidate's redundant resolver scaffolding; the regression launches real TypeScript workers rather than only checking the argument string.

Changing every caller's cwd would alter caller behavior and leave future callers exposed. Removing CLI-specific parent-preload pinning is also inappropriate: that code still owns inherited runtime flags and target-package selection. No additional safe connected cleanup remains.

Compatibility review: this four-file patch changes only worker argument construction and its tests. It changes no configuration/default surface, SQLite schema, or persistent-store semantics, so migration proof is not applicable. ClawSweeper's two compatibility checklist items were triggered by `unknown-truncated-pull-files` while GitHub held an older base snapshot. Refreshing the existing `main` base restored the complete four-file diff; the head and code did not change.

## User Impact

Source-mode maintenance workers can start from unrelated directories, including temporary directories used by state inspection. This does not establish or claim a defect in packaged JavaScript workers.

## Evidence

- The new `.ts`, `.mts`, and `.cts` subprocess regressions all fail against the original resolver with `Cannot find package 'tsx'` from the temporary child cwd. Each fixture requires TypeScript enum transformation and must complete with output `42`.
- Final production delta: 4 added / 1 removed = **+3 lines**. Tests/support: 29 added / 3 removed = **+26 lines**. Growth is within the explicitly approved +15 production-line allowance.
- Head: `6624c18bfb4046bb54ae898ea503dd42d55a387e`.
- `node scripts/run-vitest.mjs src/infra/runtime-worker-url.test.ts src/infra/openclaw-cli-invocation.test.ts src/infra/sqlite-readonly-location.test.ts src/infra/sqlite-readonly-location.worker.test.ts`: **91 passed, 1 skipped**.
[...]

PR #141594, opened 2026-09-07
improve: avoid repeated snapshot compilation in handoff tests

steipete:
## What Problem This Solves

Resolves a problem where the managed-update lifecycle tests repeatedly compile the snapshot worker from source even though their fixture intends to use the already-built worker. The redirect still expected the bare `tsx` argument after the shared launcher began pinning it to an absolute file URL in #140458.

## Why This Change Was Made

Derive the exact source-worker argument vector with `resolveRuntimeWorkerArgv`, then match that vector in the existing fixture preload. The snapshot remains a fresh real process using the same built worker as packaged updates. Source admission and orchestration, candidate isolation, and the existing repair-effect checks remain intact.

## User Impact

Faster tests with no production behavior change. This changes one test-support file (+8/-2), with no added cache, dependency, shard, timeout, or test removal.

## Evidence

- A current-main baseline at `bd62b672` passed all four existing repair-effect cases. An observed child command still ran `src/infra/update-candidate-state.worker.ts` with an absolute TSX loader URL, confirming the stale match.
- Same four-case order on the candidate: all pass. Validating cases decreased from **55.634s / 29.806s** to **30.099s / 10.112s**. Verifying cases took **9.347s / 9.768s**, versus **12.470s / 8.574s** before. These sequential local samples show a useful improvement, not a portable CI ratio.
- Whole commands include the normal runtime prerequisite build: **211.56s before, 124.56s after**. Build/load variation is included in those totals, so it is not all attributed to the fix.
- A prior phase probe attributed **16.9–23.7s** to validating rehearsal and only **0.52–0.98s** to compiled repair-runtime import. Its separate Slack startup cost was already addressed by #141512, which is included in this baseline and is not part of this patch.
- All **173 tests** passed across the full lifecycle file and runtime-URL suite. Changed-file gates passed, including types, lint, and unused-export scans. Managed P2 review found no actionable issues.

A focused passing case with temporary launch logging recorded an actual child PID using `dist/infra/update-candidate-state.worker.js` with no TSX preload. The logger was removed by restoring the exact candidate bytes already tested and reviewed; recorded child and parent PIDs had exited. The earlier short process samplers missed that short-lived child and are not used as proof of the redirect.

Exact-head [CI run 34164332059](https://github.com/openclaw/openclaw/actions/runs/34164332059) passed in **7m30s**, including Docker seed tests and the required gate. The [Node job](https://github.com/openclaw/openclaw/actions/runs/34164332059/job/101872381444) passed all **164 lifecycle cases** in 182.95s Vitest time. The four target cases ran in their original order at **14.035s, 7.567s, 6.612s, and 6.552s**; compiler setup took 4.650s. Broad main CI still has runner-queue delays, so this PR run does not establish a consistent eight-minute result for the entire campaign.

### DIFF
--- a/src/infra/update-managed-service-handoff-repair.test-support.ts
+++ b/src/infra/update-managed-service-handoff-repair.test-support.ts
@@ -11,6 +11,7 @@ import {
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import { withServer } from "../plugin-sdk/test-helpers/http-test-server.js";
 import { runtimeProcessEntrypoints } from "./runtime-process-entrypoints.js";
+import { resolveRuntimeWorkerArgv } from "./runtime-worker-url.js";
 import type {
   ManagedRepairBoundary,
   ManagedServiceManagerBoundaryRunner,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [956e29f911bc18ab89d28ef081fcf13c2e84618a](https://github.com/openclaw/openclaw/pull/141594/commits/956e29f911bc18ab89d28ef081fcf13c2e84618a)
- Landed commit: [4f2aca64b82a9bb509e9f655f72755ba77d3e8d8](https://github.com/openclaw/openclaw/commit/4f2aca64b82a9bb509e9f655f72755ba77d3e8d8)

PR #141595, opened 2026-09-07
refactor(tts): retire copied OpenAI test provider

steipete:
## What Problem This Solves

The generic TTS contract suite carried its own OpenAI voice/model policy and HTTP implementation. Some tests therefore verified that copy, including successful synthesis from an audio response the fake deliberately discarded, rather than the shipped provider.

## Why This Change Was Made

Retire the copied OpenAI provider and transport. Generic core tests use a small synthetic speech provider for routing, readiness and auto-apply decisions. OpenAI-owned integration tests exercise the same public TTS APIs with the actual provider in Vitest's existing module graph, retaining endpoint/config/directive/instruction checks and proving cancellation against real unfinished loopback responses.

The four timeout cases cover ordinary and telephony synthesis while waiting for headers or audio body, with unchanged timeout/watchdog budgets and connection-close assertions. Reuse the existing environment helper to restore absent TTS preference variables correctly.

## User Impact

No production behavior changes. The test suite maintains less copied provider code and tests the real OpenAI implementation through the public synthesis boundary. Existing readiness fallback, error redaction, summarization and auto-apply contracts remain covered.

## Evidence

129 distinct tests pass across the revised core/OpenAI suites and their existing siblings. The four real header/body timeout cases also pass alone from a fresh cache, with the existing 100 ms request timeout, 1 s watchdog and 2 s test budget. All 34 selected checks pass, and fresh independent review through P2 found no actionable issue. Actual Vitest source maps were checked against the provider and SDK sources.

The combined owner suites ran 48 cases in 47.8 seconds, versus 44 original cases in 44.0 seconds on the same host. This is a bounded cold-run observation, not a general performance claim. Proof uses synthetic data, mocked successful HTTP responses and real loopback timeout requests; no live OpenAI service result is claimed.

The complete change removes 64 test/support code lines / 62 physical lines, including the new integration suite. Production source is unchanged. The existing assertion baseline shrinks from 31 to 19; no limit, assertion or timeout allowance increases.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3317,7 +3317,7 @@ src/plugins/config-normalization-shared.ts	1
 src/plugins/config-schema.ts	6
 src/plugins/contracts/inventory/bundled-capability-metadata.ts	3
 src/plugins/contracts/registry.ts	1
-src/plugins/contracts/tts-contract-suites.ts	31
+src/plugins/contracts/tts-contract-suites.ts	19
 src/plugins/conversation-binding.ts	1
 src/plugins/current-plugin-metadata-snapshot.ts	1
 src/plugins/dashboard-capabilities.ts	1
--- a/extensions/openai/speech-runtime.contract.test.ts
+++ b/extensions/openai/speech-runtime.contract.test.ts
@@ -1,0 +1,353 @@
+// Exercise the public TTS runtime with the actual OpenAI speech provider in one module graph.
+import { randomUUID } from "node:crypto";
+import { tmpdir } from "node:os";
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3da1a789a49e13da906bd17c982524ad35437820](https://github.com/openclaw/openclaw/pull/141595/commits/3da1a789a49e13da906bd17c982524ad35437820)
- Landed commit: [3865413f4f63e6c1560d634751f275794028ad37](https://github.com/openclaw/openclaw/commit/3865413f4f63e6c1560d634751f275794028ad37)

PR #141603, opened 2026-09-07
fix(android): show expanded catalog refresh failures

steipete:
## What Problem This Solves

Fixes an issue where Android could hide a session-catalog refresh failure after the user had loaded more pages. Previously loaded rows stayed visible, making the failed refresh look successful.

## Why This Change Was Made

Remove the inner catch that converted a failed page request into a successful missing-host result. The existing refresh error handler now handles that failure, preserving loaded rows and page depths while clearing loading and publishing the existing error message. Successful missing-host responses retain their existing behavior.

## User Impact

An expanded catalog reports a failed refresh while keeping its previous contents available.

## Evidence

The existing paging test first proves successful refresh/load-more, then fails a later-page request. On unchanged production, only the visible-error assertion failed among 11 selected cases. The identical regression and all existing catalog siblings passed after repair: 15 tests, zero skips.

All 14 canonical changed checks passed, including Gradle ktlint. Independent P0–P2 review found no actionable defect. Proof ran as hermetic Robolectric/JVM tests on a disposable Linux worker; raw JUnit results were inspected and the worker was stopped. Device rendering and a live Gateway were not exercised.

CI recovery: the original run failed in an unrelated Voice Wake save-notice test. Its synchronization fix landed separately in #141635. This branch includes that prerequisite through a rebase; both changed file contents and the complete introduced patch remain byte-identical to the reviewed version. Exact-head CI34174157023 passed8jobs with26skipped;the required gate is green.

### DIFF
--- a/apps/android/app/src/main/java/ai/openclaw/app/NodeRuntime.kt
+++ b/apps/android/app/src/main/java/ai/openclaw/app/NodeRuntime.kt
@@ -5945,26 +5945,20 @@ class NodeRuntime private constructor(
           loadedPageDepthsByHost = previousPageDepths,
           isCurrent = { sessionCatalogRefreshSeq.get() == requestSeq },
         ) { catalogId, hostId, cursor ->
-          try {
-            val pageResponse =
-              requestGatewayData(
-                gatewayScope,
-                "sessions.catalog.list",
-                sessionCatalogPageParams(
-                  normalizedAgentId,
-                  catalogId,
-                  mapOf(hostId to cursor),
-                ),
-              )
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0c81c4aa7544037c63929ccadf7d4338e24ac0c0](https://github.com/openclaw/openclaw/pull/141603/commits/0c81c4aa7544037c63929ccadf7d4338e24ac0c0)
- Landed commit: [904422ac4201c526449466b4003544daa723fc49](https://github.com/openclaw/openclaw/commit/904422ac4201c526449466b4003544daa723fc49)

Issue #121823, opened 2026-08-11
[Bug]: memory-core dreaming promotion parks the event loop (minutes on native FS, 90+ min on FUSE state): healthz dark, in-process guards starve; liveness monitor observes but cannot act


BMA-Aaron:
## Summary

The nightly memory-core **"Memory Dreaming Promotion"** job runs heavy synchronous work (`node:sqlite` `DatabaseSync` + `writeFileSyncAtomic` + scoring) on the gateway's main thread. On our install this parks the event loop hard: healthz stops answering, log flushes stop, and every in-process protection (liveness monitor, lane timeouts, rescue crons) is starved along with it — they all share the loop they're supposed to protect.

Two eras, same mechanism:

- **State on FUSE-backed storage** (Unraid `/mnt/user` shfs bind): the park lasted **89–95 minutes** nightly.
- **State moved to native btrfs**: the same job still parks the loop **~3–6 minutes**, now purely CPU-bound + fsync. The filesystem only amplified; the synchronous work itself scales with store size.

Related open reports (different reproducers, same family): #78805 (sync I/O blocking), #85366 (startup sidecars saturate loop), #112698 (notification path starves main thread). This report adds a nightly in-product reproducer (memory-core dreaming), `/proc`-level evidence, and four concrete asks.

## Environment

- OpenClaw **2026.7.1** (docker, `ghcr.io/openclaw/openclaw`, tini init)
- Host: Unraid 7.x; gateway state initially bind-mounted from `/mnt/user` (shfs/FUSE), later remounted to native btrfs — both reproduce, FUSE amplifies ~15-30×
- memory-core dreaming enabled (`0 3 * * *`), memory index ~1.9 GB sqlite

## Observed — era 1 (FUSE-backed state)

- The dreaming promotion cron failed to complete for 6+ consecutive nights (killed/interrupted every run). Worst case: event loop dark **03:05:38→04:34:54 (89 min, zero log lines)**.
- A custom heartbeat cron scheduled mid-window fired 25 minutes late into the stall and died to its own starved timers (its lane error reported 288883 ms = a 210 s timeout + 79 s of timer starvation).
- Liveness fingerprint while parked in syscalls: `eventLoopUtilization=1.0` with `cpuCoreRatio≈0.09`; boot phases `bootstrap-context` 60 s / `session-resource-loader` 56 s vs ~5 s / 2.5 s healthy.
[...]

PR #111392, opened 2026-07-19
fix(memory-core): skip unchanged Dreaming workspace state writes

wuqxuan:
Closes #111376

## What Problem This Solves

Fixes an issue where memory-core Dreaming would rewrite every workspace entry in the dreaming-session-ingestion-files plugin-state namespace even when no values changed. On long-lived workspaces this full rewrite can stall the gateway event loop for tens of seconds and coincide with control-plane / channel instability during a Dreaming run.

## Why This Change Was Made

writeMemoryCoreWorkspaceEntries now reads existing workspace rows once, compares each desired canonical value with isDeepStrictEqual, and calls register() only for new or changed rows. Rows missing from the desired set are still deleted.

The comparison snapshot is updated after each actual write. This preserves the writer's sequential last-write-wins behavior when duplicate logical keys appear in one input batch.

After any real write, the writer rereads authoritative plugin-state and restores desired rows that capacity eviction removed mid-pass (skipped equal rows). True no-op passes still perform zero register() calls.

Fix quality: compatibility — same public API; final desired set is restored after eviction-capable writes; performance — skips unchanged register() work on no-op rewrites; usability — no new config; security — no trust-boundary change.

## User Impact

Operators running managed Dreaming on large workspaces should no longer pay a full plugin-state rewrite when workspace ingestion state is unchanged. Gateway event-loop pressure from redundant Dreaming state writes should drop substantially on stable workspaces.

## Evidence

Latest head (post-reconcile + main merge + scale proof): `8559a8dff6e28005d348bccb9bfb6603d5cffced`

Scale proof (2,000 rows, real SQLite test store):
- first pass register=2000
- identical second pass register=0
- third pass with 7 changed rows register=7
- focused suite: 11 passed

Previous reconcile head: `e4a9274d7f3d3cdb5eac84ecf9194f45cf22067f`

Focused local proof:

~~~text
extensions/memory-core/src/dreaming-state-write.test.ts
Test Files  1 passed (1)
Tests       10 passed (10)
Duration    ~4.6s
~~~

Verification host: secretless canonical GitHub Actions on Linux for the untrusted contributor head 423f6f6020ef8a8f41a212c464c74dd312d713b9.

Focused latest-head result:

~~~text
extensions/memory-core/src/dreaming-state-write.test.ts
Test Files  1 passed (1)
Tests       6 passed (6)
Duration    8.23s; test body 276ms
~~~

[...]

Issue #111376, opened 2026-07-19
[Bug]: memory-core Dreaming rewrites unchanged workspace state and stalls the gateway event loop


khenn:
## Summary

In OpenClaw 2026.7.1, memory-core Dreaming can synchronously rewrite every workspace entry in the `dreaming-session-ingestion-files` plugin-state namespace even when no values changed. On a long-lived workspace, the rewrite duration closely matches a severe gateway event-loop stall and causes control-plane/channel instability during the run.

## Environment

- OpenClaw 2026.7.1 (`2d2ddc4`)
- npm-global installation
- Linux, systemd user service
- SQLite plugin-state backend
- Two Dreaming workspaces: one long-lived, one smaller

## Root cause

The bundled `writeMemoryCoreWorkspaceEntries(params)` implementation builds each desired row and unconditionally awaits:

```js
await store.register(stateKey, {
  version: 1,
  workspaceKey,
  workspaceDir: path.resolve(params.workspaceDir),
  key: entry.key,
  value: entry.value
});
```

It calls `store.entries()` only afterward to find rows to delete. There is no comparison with the current value, so unchanged state is serialized and synchronously persisted again.

## Observed evidence

During one Dreaming run:

- Long-lived workspace: 6,923 rows were rewritten.
- Its row `created_at` timestamps spanned about 130.7 seconds.
- Gateway maximum event-loop delay in the same run was about 130.8 seconds.
- Smaller workspace: 462 rows spanned about 10.7 seconds.
- The run coincided with stale-lane recovery, Discord disconnect/restarts, and a REM narrative fallback.

The nearly identical large-workspace write span and event-loop delay make the synchronous full rewrite the likely primary stall source.

## Expected behavior

Dreaming should persist only new or changed workspace-state rows and delete rows no longer present. An unchanged second run should perform zero writes.

## Proposed compatible fix

1. Read the namespace entries once.
2. Index current rows for the target workspace by key.
3. Construct the same desired value object and compare it with the existing value.
4. Call `register()` only for new or changed rows.
5. Delete current rows absent from the desired key set.

A batched/off-thread state-store write path would also address event-loop starvation, but skipping unchanged rows is a small compatibility-preserving fix for the common case.

## Focused verification

A contract test can cover:

- First run writes all rows.
- Identical second run writes zero rows and deletes zero rows.
- One changed row produces one write.
- One removed row produces one delete.
- Final serialized state is identical to the current algorithm.

## Duplicate check

[...]

PR #141613, opened 2026-09-07
fix(memory): keep workspace state updates responsive

steipete:
Related: #111376, #111392, #121823.

## What Problem This Solves

Fixes an issue where users could see delayed Gateway I/O while background memory workspace state was replaced or cleared. Large loops await keyed-store methods that finish their SQLite work synchronously, so those awaits can keep running through microtasks without servicing other event-loop work.

## Why This Change Was Made

Workspace replacement and clearing now yield to the event loop after every ten completed scalar store operations, following the existing memory source-sync convention. Registrations still finish before obsolete-row cleanup; each operation keeps its existing transaction, retention, error and workspace/namespace boundaries. The separate scalar writer is unchanged.

This is complementary to #111392, which proposes reducing unchanged writes and selecting a different retention policy. This change preserves every existing scalar write and SQL count and does not claim to close the broader write-amplification or Dreaming-stall issues.

Storage compatibility is unchanged: workspace records retain the same `version: 1` value shape, hashed keys, namespaces, encoding and readers. The keyed-store API, SQLite backend, schema, indexes, SQL, per-operation transactions, TTLs, limits and retention policy are untouched. Existing records remain readable and writable without migration; the only new state is a function-local completed-operation counter. The real-store regressions verify replacement and clearing against existing records, correct final contents, foreign workspace/namespace isolation, and unchanged committed-prefix behavior after rejection. The generic `unknown-data-model-change` review flag therefore does not identify a persistent-model change in this patch.

## User Impact

Queued work can run before a large memory-state operation finishes. This improves scheduling responsiveness; total duration can be slightly longer, and an individual expensive native operation still runs to completion before yielding.

## Evidence

Verified candidate: `47f790b3005f8f2fdf7c84efbd273b896c311a3a`.

- [Exact-head required CI](https://github.com/openclaw/openclaw/actions/runs/34165996815) passed, including `openclaw/ci-gate`.
- Four real-store regression cases pass. Replacement and cleanup progress failed on unchanged main; the equivalent clear case failed before its matching fix. Each progress test observes partial committed work outside a SQLite transaction, preserves foreign workspace/namespace rows, and checks final state. Rejected registration retains its committed prefix and skips cleanup; scalar recovery remains intact.
- All 164 tests in six existing ingestion, promotion, consolidation, repair, forget and workspace-lock suites passed without test or fake-timer changes.
- Full changed checks passed, including production/test types, lint, exports, storage/runtime guards and zero runtime import cycles. Managed independent review through P2 found no accepted/actionable findings; the ownership and test review also found no issue.
- [Blacksmith Testbox proof](https://github.com/openclaw/openclaw/actions/runs/34165089145) completed normal `pnpm build`, all four regressions and 36 ordinary real-owner measurements with payload exit 0. The source was verified, and the one-shot Testbox is independently confirmed completed. Its external portal sync timed out afterward; that warning is retained separately. The delegated workflow may show cancellation during teardown and is not represented as green PR CI.

The Linux probe used Node 24.19.0 / SQLite 3.53.3, 64 overwrites with 1 KiB payloads, and three alternating baseline/candidate pairs per shape. Sparse namespaces contain 64 rows; dense namespaces contain all listed plugin rows. A timer and immediate callback are queued before each replacement, and a 5 ms heartbeat records the largest observed service gap. All candidate immediate callbacks ran before completion with a partial write count and no active transaction; baseline callbacks ran after completion. Final contents and row counts were verified.

| Plugin rows / namespace | Median first timer, before → after | Worst heartbeat gap, before → after | Median total, before → after |
| --- | --- | --- | --- |
| 512 / sparse | 14.5 → 2.4 ms | 19.2 → 6.8 ms | 14.4 → 15.8 ms |
| 512 / dense | 15.7 → 2.4 ms | 17.3 → 7.0 ms | 15.7 → 15.6 ms |
| 2,048 / sparse | 33.9 → 4.6 ms | 35.2 → 9.2 ms | 33.9 → 29.0 ms |
| 2,048 / dense | 72.0 → 10.7 ms | 73.7 → 13.0 ms | 71.9 → 72.7 ms |
| 8,192 / sparse | 91.2 → 14.3 ms | 96.7 → 14.6 ms | 91.1 → 87.0 ms |
| 8,192 / dense | 259.6 → 37.9 ms | 261.6 → 42.0 ms | 259.5 → 256.4 ms |

A separate counter pass verified identical query executions: 64 plugin counts, 64 namespace counts, 64 point reads/upserts/expiry sweeps and one final listing. Each COUNT returned one aggregate row; this is a scheduling fix, not a row-materialization optimization. The final namespace listing still materializes its rows.

Retained macOS pairs of the same replacement implementation had a slower dense median total (685.5 → 728.9 ms), while the worst heartbeat gap improved from 715.4 → 227.8 ms. Those measurements preceded the equivalent clear-sibling addition; the final-head Linux proof above includes it. All samples and qualifications are retained. These small synthetic samples establish progress, not a universal latency bound, reduced SQL work, or sustained production acceptance.

Production: **+14 lines**. Tests: **+150 lines**. No schema, index, configuration, dependency, row-limit, TTL, retention or batch-atomicity change is included.

### DIFF
--- a/extensions/memory-core/src/dreaming-state.test.ts
+++ b/extensions/memory-core/src/dreaming-state.test.ts
@@ -1,0 +1,150 @@
+import path from "node:path";
+import { setImmediate } from "node:timers";
+import type { OpenKeyedStoreOptions } from "openclaw/plugin-sdk/plugin-state-runtime";
+import {
+  createPluginStateKeyedStoreForTests,
+  openOpenClawStateDatabase,
+  resetPluginStateStoreForTests,
+} from "openclaw/plugin-sdk/plugin-state-test-runtime";
+import { useAutoCleanupTempDirTracker } from "openclaw/plugin-sdk/test-env";
+import { afterEach, expect, test } from "vitest";
+import {
+  clearMemoryCoreWorkspaceNamespace,
+  configureMemoryCoreDreamingState,
+  readMemoryCoreWorkspaceEntries,
+  writeMemoryCoreWorkspaceEntries,
+  writeMemoryCoreWorkspaceEntry,
+} from "./dreaming-state.js";
+
+const tempDirs = useAutoCleanupTempDirTracker((cleanup) =>
+  afterEach(() => {
+    configureMemoryCoreDreamingState(() => {
+      throw new Error("memory workspace test store is closed");
+    });
+    resetPluginStateStoreForTests();
+    cleanup();
+  }),
+);
+
+function createFixture() {
+  const root = tempDirs.make("memory-workspace-state-");
+  const env = { OPENCLAW_STATE_DIR: root };
+  configureMemoryCoreDreamingState(<T>(options: OpenKeyedStoreOptions) =>
+    createPluginStateKeyedStoreForTests<T>("memory-core", { ...options, env }),
+  );
+  const scope = { namespace: "workspace-progress", workspaceDir: path.join(root, "workspace") };
+  return {
+    scope,
+    foreignWorkspace: { ...scope, workspaceDir: path.join(root, "foreign-workspace") },
+    foreignNamespace: { ...scope, namespace: "workspace-other" },
+    database: openOpenClawStateDatabase({ env }),
+  };
+}
+
+async function readValues(scope: ReturnType<typeof createFixture>["scope"]) {
+  return Object.fromEntries(
+    (await readMemoryCoreWorkspaceEntries(scope)).map(({ key, value }) => [key, value]),
+  );
+}
+
+test.each(["replacement", "cleanup", "clear"] as const)(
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [47f790b3005f8f2fdf7c84efbd273b896c311a3a](https://github.com/openclaw/openclaw/pull/141613/commits/47f790b3005f8f2fdf7c84efbd273b896c311a3a)
- Landed commit: [816a81d737b102d49f0e742e364e5fd4eb9344ec](https://github.com/openclaw/openclaw/commit/816a81d737b102d49f0e742e364e5fd4eb9344ec)

PR #141629, opened 2026-09-07
improve: reduce session recovery import overhead

steipete:
## What Problem This Solves

Resolves a problem where session recovery and admission load outbound payload normalization and provider runtime just to sanitize or clear stored reply text. The module-identity regression test exposed this unnecessary work during cold SDK transformation.

## Why This Change Was Made

Move the existing sanitizer and clear patch into a small text-only state module and redirect all five lightweight consumers to it. Payload normalization keeps its existing owner and imports the same sanitizer when it actually needs it. The old internal exports are removed; no public SDK or package export changes.

The helper implementations are byte-identical. Recovery decisions, persisted record shape, replay ordering, silence handling, media handling, and ownership checks are unchanged.

## User Impact

Less cold-import work when admitting or recovering sessions, with the same reply and recovery behavior. No cache, dependency, configuration option, timeout, or shard change.

## Evidence

- Rebuilt admission graph: **666 → 520 statically reachable chunks**, **9.16 MB → 7.02 MB**. The provider-loader path is no longer reachable through recovery state. This removes about **2.14 MB** from the evaluated graph; total emitted bundle size is nearly unchanged because actual delivery still owns that code.
- The unchanged module-identity test passes with native graph equality, transformed graph inequality, and all three real storage/admission scenarios retained. Local case time: **17.028s → 8.989s**; whole command: **24.87s → 10.33s**. These are sequential local observations with host/cache variation, not a portable speedup ratio.
- **92 existing tests passed**: the identity case, 90 sanitizer/builder/recovery/core/heartbeat cases, and the named real resume-prompt sanitation case. No assertions or test deadlines were weakened.
- Original helper bodies and the identity test were independently checked byte-for-byte. Temporary profiling/inventory instrumentation was removed. Managed P2 review found no actionable issues. The full build and changed-file gate passed, including core/test types, lint, unused-export scans, and runtime import-cycle checks.

Production LOC delta is **+2**, entirely module-boundary/import/comment overhead; implementation is moved. Test LOC is net zero, with one import updated.

Additional sibling cross-check: the exact patch also passed all **38 migration caller-mode/refusal cases** on the same baseline, preserving their observed order. This brings local coverage to **130 existing tests**. These later timing samples are not used to claim a separate migration speedup.

Final exact-head [CI passed in 8m24s](https://github.com/openclaw/openclaw/actions/runs/34169571333), with the unchanged module-identity case at **10.175s**. The current-head ClawSweeper review has no actionable findings.

Two earlier CI runs exposed separate process-containment fixture failures (`reparented` and `root-resumed`). [PR #141634](https://github.com/openclaw/openclaw/pull/141634) independently landed the proven PID-selection forwarding repair. This branch was refreshed onto that canonical fix; its eight authored file blobs remain byte-identical to the reviewed refactor, with no containment changes in this PR. The second failure occurred despite forwarding, so the remaining OS/fixture race is being investigated separately with Linux process-state diagnostics; the final green run is not claimed as proof that the flake is resolved.

### DIFF
--- a/src/agents/main-session-recovery/main-session-recovery-state.ts
+++ b/src/agents/main-session-recovery/main-session-recovery-state.ts
@@ -1,7 +1,7 @@
 import {
   PENDING_FINAL_DELIVERY_CLEAR_PATCH,
   sanitizePendingFinalDeliveryText,
-} from "../../auto-reply/reply/pending-final-delivery.js";
+} from "../../auto-reply/reply/pending-final-delivery-state.js";
 import type {
   InternalSessionEntry as SessionEntry,
   MainRestartRecoveryState,
--- a/src/agents/main-session-recovery/main-session-restart-dispatch.ts
+++ b/src/agents/main-session-recovery/main-session-restart-dispatch.ts
@@ -2,7 +2,7 @@ import { randomUUID } from "node:crypto";
 import { normalizeOptionalString } from "@openclaw/normalization-core/string-coerce";
 import { GatewayClientRequestError } from "../../../packages/gateway-client/src/index.js";
 import { isExecutionIdentityCollectionEnabled } from "../../audit/audit-config.js";
-import { sanitizePendingFinalDeliveryText } from "../../auto-reply/reply/pending-final-delivery.js";
+import { sanitizePendingFinalDeliveryText } from "../../auto-reply/reply/pending-final-delivery-state.js";
 import type { SessionEntry } from "../../config/sessions.js";
 import {
   buildRestartRecoveryClaimCleanupPatch,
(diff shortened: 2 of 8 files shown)


steipete:
@clawsweeper re-review

The recovery import refactor is unchanged. A separate test-only CI repair now preserves the PID selection requested by containment when its fixture delegates to real OS inspection. A focused argument-capture check failed before forwarding and passed afterward; temporary diagnostics were removed. All 51 containment/snapshot cases, the changed-file gate, and refreshed P2 review pass. Production cleanup rules, assertions, and deadlines are unchanged. The PR body records the original failure and the remaining attribution limit.

steipete:
Merged via squash.

- Prepared head SHA: [cbad65aa280374ee50b292ac327dcd26d6ad5aec](https://github.com/openclaw/openclaw/pull/141629/commits/cbad65aa280374ee50b292ac327dcd26d6ad5aec)
- Landed commit: [9fb6b2b5ec3b2cf1889912c5d6f030ffe31d2573](https://github.com/openclaw/openclaw/commit/9fb6b2b5ec3b2cf1889912c5d6f030ffe31d2573)

PR #141632, opened 2026-09-07
fix(ui): refresh selected usage details

steipete:
## What Problem This Solves

Fixes an issue where users selecting Refresh in Usage saw updated totals while the selected session’s timeline and conversation stayed stale.

## Why This Change Was Made

Carry the selected session key with the accepted manual refresh request and reload its details through the existing detail owner. Automatic refreshes retain their current behavior, and a late response cannot transfer manual refresh intent to a different selection. This also removes a redundant private reload wrapper.

## User Impact

Refresh updates the selected timeline, conversation, and system-prompt breakdown alongside the overview. The existing query controls, refresh errors, and bounded automatic retry policy remain intact.

## Evidence

The existing mounted regression fails on the original code with stale timeline and conversation values, then passes unchanged after the fix. All 42 related mounted cases and 20 canonical changed checks pass, including automatic polling and replacement-selection controls. Independent Codex P0–P2 review found no actionable findings.

The same bundled Chromium scenario with a synthetic Gateway shows overview480/timeline360/five conversation entries before, then overview480/timeline480/six entries after; each detail request count changes from one to two. Existing tool-count, drag-selection, and Reset checks are retained. The inspected screenshots below use synthetic data. This validates the client request/render flow; it does not claim an atomic server snapshot or live provider billing freshness.

Before Refresh leaves the selected timeline and conversation count stale

After Refresh updates the selected timeline and conversation count

Before the newly completed reply is missing

After the newly completed reply appears

### DIFF
--- a/docs/web/control-ui/feature-reference.md
+++ b/docs/web/control-ui/feature-reference.md
@@ -89,6 +89,7 @@ Control UI capabilities grouped by area, each with the Gateway RPC methods behin
     - Provider cards call `usage.status` and show live plan names, quota windows, balances, spend, and budgets reported by configured provider plugins.
     - A provider usage failure does not block the session/cost dashboard; unavailable provider cards show their own error state.
     - Incomplete session/cost totals stay readable while the visible, focused page checks for updates. Automatic checks are bounded; if they pause, select **Refresh** to check again.
+    - **Refresh** also reloads the selected session's timeline, conversation, and system-prompt breakdown.
     - The overview loads session summaries first. Full system-prompt breakdowns load when you select a session; the `has:context` filter still works before opening details.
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [44e98fb05ddba19de36e45990ba80f8dbdec40b6](https://github.com/openclaw/openclaw/pull/141632/commits/44e98fb05ddba19de36e45990ba80f8dbdec40b6)
- Landed commit: [8ca8c8ef0e382df25e0c648f1337d78f19a1d463](https://github.com/openclaw/openclaw/commit/8ca8c8ef0e382df25e0c648f1337d78f19a1d463)

PR #141620, opened 2026-09-07
fix(btw): report usage for direct-provider answers

steipete:
## What Problem This Solves

Fixes an issue where users received successful direct `/btw` answers but their usage was missing from configured model diagnostics and the optional reply-usage handoff. The harness side-question path already reported that usage.

## Why This Change Was Made

Direct and harness completions now share the existing BTW accounting owner. Direct completions normalize the returned usage once, and three identical diagnostic token projections use one pure helper. This preserves recorded cost, explicit totals, correlation IDs, diagnostics controls, and the ephemeral conversation contract.

## User Impact

Configured diagnostics include usage from successful direct side questions, including completed block streams. Questions and answers remain outside session history and session-derived `/usage cost` totals; blocks already delivered are not retroactively decorated.

## Evidence

- The same synthetic regression and sibling selection went from **20 passed / 6 failed** before the fix to **26 passed** afterward, with unchanged test bytes. The six failures covered direct final-text and block completion with diagnostics disabled, enabled, and at its default. Harness, ingress, and normal-reply controls passed throughout.
- Production **−3 lines**, existing tests **+44**, docs **+4**. No configuration, storage, audit/identity, protocol, or public SDK change.
- All **29 classified changed checks passed**, including core/test typechecks, focused lint, formatting, dead exports and architecture guards. Diff-scoped deslop review found no further change needed. Managed review is complete; its block-timing finding was rejected after source inspection confirmed the existing queued-delivery and best-effort handoff contract. Exact-head CI34170228202 passed108jobs with16skipped;the required gate is green.
- No live model/provider/exporter calls were made; proof uses existing synthetic stream events and isolated test state.

Reply hooks read the live snapshot when queued delivery starts, so queued blocks and later TTS replies can consume completion usage. Already-delivered blocks keep the documented best-effort limitation; this change adds no streaming delay.

CI recovery: the first run failed in an unrelated process-test adapter. That repair landed independently in #141634. This branch now includes it through a rebase; the complete introduced patch and all six changed file contents are byte-identical to the reviewed version. Final CI passed for the rebased head.

PR #141637, opened 2026-09-07
fix(test): preserve process snapshot selection

steipete:
Related: #141620

## What Problem This Solves

Resolves a test-adapter mismatch found while investigating unrelated CI failures blocking #141620. Cleanup requested a selected process snapshot, but the adapter dropped that argument and delegated a full-host snapshot instead.

## Why This Change Was Made

Forward the existing optional selector through the existing adapter to the real snapshot reader. All scenarios, expected cleanup results, exact assertions, deadlines, and fixture cleanup are preserved.

## User Impact

No product behavior changes. The fixture now exercises the same snapshot scope requested by its production caller.

## Evidence

The production caller supplies the selector, and the reader already accepts it. Reversing only the argument-forwarding edits reproduces the entire original test byte-for-byte. The canonical extension-test typecheck, targeted formatting, applicable Oxlint, and diff check pass; independent Codex P0–P2 review of the final staged patch found no actionable findings. The first CI run caught a default-parameter-order lint error in the new adapter signature; the follow-up keeps the same deadline default inside the call.

Runtime validation remains pending ordinary exact-head CI. No local process tests or probes were run. The earlier CI log does not identify the failed observation, so this fixes a verified adapter defect without claiming that the historical runtime cause has been conclusively reproduced.

### DIFF
--- a/extensions/codex/src/app-server/transport.process.test.ts
+++ b/extensions/codex/src/app-server/transport.process.test.ts
@@ -367,8 +367,8 @@ process.stdin.on("end", () => process.exit(0));
       };
       const snapshotSpy = vi
         .spyOn(processSnapshot, "readCodexAppServerProcessSnapshot")
-        .mockImplementation((inspectionDeadline = Date.now() + 2_000, pids) =>
-          readSnapshot(inspectionDeadline, pids),
+        .mockImplementation((inspectionDeadline, pids) =>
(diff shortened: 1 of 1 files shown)


steipete:
Superseded by #141634, merged as f73f03c697d695824d725a9228febd29112d732f. I verified that current main includes this PR’s selector forwarding and deadline-default correction, and also preserves the selected PID for scalar reads. Closing this duplicate; no additional production or test change is needed here.

PR #141609, opened 2026-09-07
perf(gateway): speed up suggestion delivery in large stores

steipete:
## What Problem This Solves

Fixes an issue where delivering a suggestion gets slower as unrelated sessions accumulate and more operators are connected.

## Why This Change Was Made

A single-target event check now uses the existing exact metadata lookup. Multiple-target events retain their shared store view, and each recipient still gets fresh authorization checks with the existing lazy visibility, author, creator, alias and incognito behavior.

## User Impact

Suggestion delivery requires less metadata work in larger session stores. No configuration or API change is needed.

## Evidence

- Real-store RED/GREEN regression: unmodified main materialized 64 rows during a warm singleton suggestion check; the candidate avoids that whole-store work. Dense 2/32-target checks, snapshot laziness and membership revoke/restore remain covered. All 52 owner and policy/creator/snapshot sibling tests passed.
- Complete candidate changed-file gate passed on an independently installed Blacksmith Testbox, including core and core-test typechecks, lint and owner guards. Both revisions built successfully before timing.
- [Paired ordinary Gateway run](https://github.com/openclaw/openclaw/actions/runs/34163315778), baseline `0bec2710f6836388b58974c3a8310dfcbc21890e` versus candidate `e0daf0a1265d03c5e68c5cfd84b2c1aa8ab0e913`: fresh state per cell, alternating baseline/candidate order, 64/512/2,048 metadata rows, 1/8/32 explicit members plus five fixed profiles, three warm-ups and ten timed additions per cell, at least 100 ms pacing.
- At 2,048 rows and 37 connected profiles, completion median improved **39.671 → 8.801 ms (77.8%)**, p90 **40.501 → 9.109 ms**, and maximum **40.900 → 9.264 ms**. Larger adjacent cells also improved. Two 64-row medians increased by **1.035 ms** and **0.362 ms**; their tail increases stayed under 1 ms. This is a bounded synthetic comparison, not a p99 or universal-speedup claim.
- Both independent inspections reconciled all **1,408 request/response pairs**, **3,928 suggestion frames**, **384 list barriers**, **216 pacing intervals**, role/membership changes, closed copied SQLite state and all 18 Gateway/fixture shutdowns. No hold, unexpected recipient, duplicate delivery or state mismatch occurred.

PR #141638, opened 2026-09-07
fix: keep Codex containment tests scoped to owned processes

steipete:
Related: #141609

## What Problem This Solves

Resolves a problem where Codex process-containment tests can report uncertain cleanup because an unrelated process is unreadable, even when the production cleanup reader requests only the owned processes.

## Why This Change Was Made

The test adapter now preserves the snapshot reader's selected PIDs, including individual identity reads. It retains full enumeration when requested and preserves all cleanup assertions, signals, identity checks, and deadlines.

## User Impact

No runtime behavior changes. This makes the containment test exercise the production inspection scope accurately.

## Evidence

- Linux/Node 24 controlled proof preserved the original reparented execution order and added an unreadable unrelated procfs neighbor after the real descendant SIGKILL. The original adapter failed with `cleanup: uncertain`; the forwarding adapter passed with `cleanup: closed`. Both confirmed root exit 0, selected-owner read success, and full-scan refusal.
- Linux/Node 24 focused transport, process-snapshot and procfs registration tests: 70 passed, seven platform skips.
- The full changed gate passed on the isolated Linux Testbox; independent review found no actionable P0–P2 findings.
- The original failure investigated on #141609 remains causally unclassified; this separate controlled proof establishes the adapter defect.

### DIFF
--- a/extensions/codex/src/app-server/transport.process.test.ts
+++ b/extensions/codex/src/app-server/transport.process.test.ts
@@ -346,11 +346,14 @@ process.stdin.on("end", () => process.exit(0));
         }
         return signalled;
       });
-      const readSnapshot = async (inspectionDeadline: number): Promise<PosixProcess[]> => {
+      const readSnapshot = async (
+        inspectionDeadline: number,
+        pids?: readonly number[],
+      ): Promise<PosixProcess[]> => {
         // Identity faults exercise signalling; cleanup needs a separate OS
         // observation. A queued kill alone must not certify an uninterruptible child.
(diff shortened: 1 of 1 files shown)


steipete:
Closing as already fixed by #141634, merged in f73f03c697d695824d725a9228febd29112d732f.

I verified the landed source preserves the same selected-PID forwarding through both fixture adapters. The remaining difference is equivalent handling of an omitted inspection deadline, so this PR adds no further repair. Its independent Linux proof and successful CI remain supporting evidence; they do not establish the cause of the original failure investigated on #141609.

PR #141639, opened 2026-09-07
fix(opencode-go): send stable session routing header

RomneyDa:
## Summary
- attach a stable `x-opencode-session` header to native OpenCode Go requests
- derive it from the conversation session id, falling back to the turn id
- leave custom proxy transports untouched

## Why
The 2026.7.33 live provider gate now receives HTTP 400 because OpenCode Go requires this routing header.

## Validation
- `node scripts/run-vitest.mjs run extensions/opencode-go/index.test.ts` (39 passed)


RomneyDa:
Closing as obsolete after review: current main already owns OpenCode session affinity centrally in packages/ai and preserves explicit header precedence. The 7.33-only gap is repaired in #141643 through the provider stream wrapper used by both affected transports.

PR #141627, opened 2026-09-07
fix: avoid Linux Gateway test port collisions

steipete:
Related: #141575

## What Problem This Solves

Resolves a problem where Linux Gateway tests intermittently fail to bind even though their selected port passed an availability probe. An outbound startup connection can claim that port and leave it unavailable in `TIME_WAIT`.

## Why This Change Was Made

Cache the actual Linux kernel ephemeral range and select complete nonprivileged listener blocks outside it, including fallback and derived ports. Preserve the existing non-Linux pool and explicit bind-permission handling. Exclude Fetch-blocked ports so the lower Linux pool remains usable by HTTP and WebSocket clients.

## User Impact

More reliable Gateway verification. This changes test tooling, with no end-user configuration change. Availability probing remains best-effort; it is not an exclusive port lease.

## Evidence

- A controlled Linux socket experiment reproduced `EADDRINUSE` for an established outbound client and its subsequent `TIME_WAIT` state. The original CI occupant could not be identified retrospectively.
- On the actual kernel range, the original allocator selected an unsafe port; the repair allocated 64 complete blocks and served 320 real HTTP requests successfully.
- All 13 allocator regressions and 12 unchanged Gateway startup owner tests pass locally and on Linux. The range and HTTP-compatibility regressions fail for their intended reasons before the corresponding fixes.
- Full changed checks and independent P2 review pass on the identical reviewed patch. Both files and their direct owners are unchanged by this main backport.

[Final Linux proof](https://github.com/openclaw/openclaw/actions/runs/34166422702)

PR #141645, opened 2026-09-07
fix(test): keep denied Linux port probes in the safe pool

steipete:
Related: #141627

## What Problem This Solves

Resolves a problem where Linux test fixtures could still select ports inside the kernel's ephemeral client range when socket probes were denied with EPERM or EACCES. This corrects a permission-fallback gap in #141627.

## Why This Change Was Made

Permission-tolerant allocation now retains the same complete, HTTP-valid, non-ephemeral candidate selected by the normal allocator, without trying additional denied socket probes. Strict allocation still reports permission errors. Non-Linux explicit fallback behavior and fail-closed handling of unreadable Linux kernel facts remain unchanged.

## User Impact

No runtime product behavior changes. Permission-constrained test fixtures retain the Linux listener-pool guarantees introduced by #141627.

## Evidence

Both Linux permission regressions fail on the previous implementation because the fallback is inside the client range. All 16 allocator cases pass with the repair, including blocked derived HTTP ports, Darwin/Windows fallback, and denied kernel facts. The unchanged real Gateway startup SecretRef owner-isolation suite passes 12/12. Full changed-file checks pass on the reviewed source; the main backport is identical in both changed files. Fresh independent P2 review found no actionable issues.

The original CI collision's actual occupant remains unknown; this follow-up addresses the independently identified permission-path gap.

### DIFF
--- a/src/test-utils/ports.test.ts
+++ b/src/test-utils/ports.test.ts
@@ -185,17 +185,47 @@ describe("deterministic test port blocks", () => {
   });
 
   it.each(["EPERM", "EACCES"])(
-    "preserves the explicit permission fallback for %s",
+    "keeps permission-denied Linux blocks outside the client range for %s",
     async (code) => {
       host.range = "32768\t60999\n";
       host.rejectPort.mockReturnValue(code);
-      const { getFreePortBlockWithPermissionFallback, isPortFree } = await import("./ports.js");
-      await expect(isPortFree(41032)).resolves.toBe(false);
-      await expect(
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0e0dde2b82d5b3ba3d5cac8e6fd96e028d805816](https://github.com/openclaw/openclaw/pull/141645/commits/0e0dde2b82d5b3ba3d5cac8e6fd96e028d805816)
- Landed commit: [420992d2ced72ad3dffb5776f48bdf2b27b29c96](https://github.com/openclaw/openclaw/commit/420992d2ced72ad3dffb5776f48bdf2b27b29c96)

PR #141652, opened 2026-09-07
fix(ci): accept frozen legacy same-version updates

RomneyDa:
## Summary
- allow the explicit frozen-target lane to recognize the legacy 7.33 same-version updater contract
- forward that compatibility flag through the Docker wrapper's existing environment allowlist
- retain strict `skipped` / `already-current` behavior by default
- require the legacy path to remain same-version and complete its normal doctor checks

## Validation
- `node scripts/run-vitest.mjs run test/scripts/test-install-sh-docker.test.ts` — 120 passed
- `bash -n scripts/test-install-sh-docker.sh`
- `oxfmt --check test/scripts/test-install-sh-docker.test.ts`
- `git diff --check`

## Release context
The 2026.7.33 updater predates the modern same-build no-op response. It safely reapplies the same package and returns `status: ok`; frozen Full Validation had begun requiring the later `skipped` contract unconditionally.

The compatibility remains opt-in from trusted frozen-target evidence, crosses the host/container boundary through the explicit smoke-runner allowlist, and fails closed for every normal run.


RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

Issue #141604, opened 2026-09-07
[Bug]: model fallback chain aborts on a non-final hop and logs nothing, skipping remaining candidates


von8794:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

When a fallback candidate fails with an error that is not recognized as a `FailoverError`, the chain aborts while candidates remain and emits no `[model-fallback/decision]` record, so a configured fallback model is silently skipped and the run ends with no log line explaining why.

### Steps to reproduce

Observed, not synthesized. The two runs below are the same agent with the same configured chain, seven minutes apart; the first is the control.

1. Configure an agent with three candidates:

       primary   anthropic/claude-opus-5
       fallback  anthropic/claude-sonnet-5
       fallback  ollama/gemma3:27b

2. Run a turn where the primary fails with an error the ladder recognizes (`empty_response`). All three hops resolve and each is logged — this is the control run at 12:47, and it proves `ollama/gemma3:27b` is present in the resolved candidate list.
3. Run a turn long enough to exceed `agents.defaults.timeoutSeconds` (1320 here). The primary is terminated and correctly fails over to hop 2.
4. Observe hop 2 fail with a bare `TimeoutError` rather than a `FailoverError`.
5. Observe that hop 3 is never attempted and that nothing is logged for hop 2's failure.

I could not narrow this further from outside: see "Actual behavior" for why the deciding branch is not observable.

### Expected behavior

Grounded in the control run of the same chain, same agent, 12:47:08-12:47:12 the same day, where every hop was logged and the chain reached its final candidate:

    decision=candidate_failed    candidate=anthropic/claude-opus-5    reason=empty_response  next=anthropic/claude-sonnet-5
    decision=candidate_failed    candidate=anthropic/claude-sonnet-5  reason=empty_response  next=ollama/gemma3:27b
    decision=candidate_succeeded candidate=ollama/gemma3:27b          reason=unknown         next=none

Expected: hop 2's failure is recorded with a decision line, and the chain either continues to `ollama/gemma3:27b` or records the reason it is stopping.

### Actual behavior

The chain stopped at hop 2. `ollama/gemma3:27b` was never attempted, no decision line was emitted for hop 2's failure, and no `run <id> ended with stopReason` line exists for the run at all — the only two run-end lines in the surrounding 35 minutes belong to other agents.

[...]

PR #141658, opened 2026-09-07
fix(agents): explain terminal model fallback stops

von8794:
## What problem this solves

Local or terminal failures can stop a model fallback chain without a diagnostic that names the stopping condition. They correctly bypass provider-failure handling, but operators cannot tell why no next candidate ran.

## Why this change was made

Name the first matching condition in the existing candidate-catch guard and emit one `model_fallback_chain_stopped` diagnostic through the existing logger. Preserve condition order and rethrow the same error. The record is separate from `candidate_failed`, so a local stop does not blame the provider.

## User impact

With warning logging enabled, operators can see the stop reason, candidate context, available session/lane attribution, and available error name. An error name that cannot be read or is not a string is omitted so logging preserves the original stop error. Console fields use the existing sanitizer; structured logs retain the existing encoding and redaction contract. No fallback, retry, timeout, profile, suspension, or logging-retention policy changes.

This covers stops inside the candidate catch. Whole-run deadlines, idle breakers, and other stops outside that catch remain outside this repair; it does not fully resolve the broader timeout incident in #141604.

## Evidence

Proof uses a real compiled `openclaw agent exec`, an explicit retained state directory, synthetic model configuration, the real HTTP client, and a deterministic loopback provider in an isolated network namespace. No live provider, channel, real credential, or production state is used.

- Pinned main: a SIGINT delivered after primary request entry preserves exit 130 and prevents fallback, but emits no chain-stop diagnostic.
- Candidate: the same flow records `caller_signal_aborted` once, retains the actual session attribution and `AbortError` name, preserves exit 130, and makes no fallback request.
- Ordinary HTTP 500 recovery, initial success, and candidate exhaustion retain their existing outcomes and diagnostics. Request counts reflect the existing retry window; they are not forced to match across runs.
- The 26-case producer/runner suite covers all eight reasons, precedence, frozen error identity, no next candidate or failure/step callback, optional attribution, warning-disabled behavior, safe console/structured representations, and ordinary controls. These are owner-boundary tests, not eight different CLI fault-injection flows.
- Corrected regression suite: 18 missing-diagnostic failures on baseline; all 224 new and existing focused tests passed, and all 26 cases in the changed file passed again after the final test-setup correction. Formatting and type-aware lint pass. Hosted CI owns full typechecks and checks.

- A throwing-name timeout regression fails by replacing the original error before the logging fix, then passes after guarded name extraction. The compiled ordinary cancellation check also passes. A later non-string name regression also fails before its type guard and passes after it.
- Independent CLI acceptance reran cancellation, success, recovery, and exhaustion on the prior build. Source review verified reuse for the unchanged ordinary paths; the compiled predecessor directly reconfirmed cancellation. The final metadata type guard has fresh owner/sibling tests and source-reviewed equivalence for these ordinary CLI inputs.


von8794:
Real-run proof added to the PR body: an isolated `agent exec` run against a local stub provider, interrupted with SIGINT, emitting

```
[model-fallback/decision] model fallback chain stopped: reason=caller_signal_aborted candidate=hangprov/hang-primary errorName=AbortError
```

with the terminal outcome preserved (exit 130, AbortError still propagates, fallback correctly not attempted), plus a same-build control on the retryable path showing the existing decision logging unchanged.

Also recorded there is a negative result I think matters: instrumenting the catch showed the whole-run agent deadline and the idle-timeout circuit breaker both stop the chain **outside** this layer, so they stay silent even with this patch. That is not in scope here and I did not widen the PR for it, but it means these diagnostics are necessary rather than sufficient for #141604.

@clawsweeper re-review

obviyus:
landpr-approve-unrelated-red
pr: 141658
head: 3e417d954ed1518002c57fdc1130bfcb121f7da7
run: 34212536521 attempt: 1
failed: checks-node-compact-large-13 job=102016905158 conclusion=failure
check: openclaw/ci-gate

obviyus:
The completed run has one failed leaf: [the session-mutation performance test](https://github.com/openclaw/openclaw/actions/runs/34212536521/job/102016905158). The corrected test-type and lint checks pass, and the exact-head review is clean.

The failure is independent of this diagnostic repair. In the [exact tested merge](https://github.com/openclaw/openclaw/commit/6edc552277aa586c2645a826dd6009ffa4aabe9c), the session mutation handler, transaction helper, failing test, counters, and Gateway-methods test configuration are unchanged from main parent `c667fa4cd0928f4ce02e394a8544e08957af170d`. The new test runs in the separate agents-core project. The archive fixture does not execute model fallback, and the changed production functions add no import-time side effects.

The [assertion at line 297](https://github.com/openclaw/openclaw/blob/6edc552277aa586c2645a826dd6009ffa4aabe9c/src/gateway/server-methods/sessions-mutations.perf.test.ts#L297) expects one `session.entry-replacements` observation from a partial module mock and receives zero. Earlier assertions passed for all 30 handler outcomes, zero transcript hydration, and one native BEGIN/COMMIT. Later assertions were not reached.

None of the six failure-conclusion runs within the inspected newest 100 completed main-push CI runs matched this failure. The underlying instrumentation cause and introducing commit remain unconfirmed. This record uses the standing authority for a source-proved unrelated CI failure; the failed test and its expectations remain unchanged for separate investigation.

obviyus:
Thanks @von8794! Landed as 71002a02253c4524c9fffdcdc0a7a186f9cc50ea with contributor credit preserved.

**Proof:** compiled `openclaw agent exec` cancellation after HTTP request entry, plus success, HTTP 500 recovery, and exhaustion controls. The 224 focused cases passed; all 26 changed-file cases passed again after the test-setup correction. Formatting, type-aware lint, hosted test-type checks, and the exact-head review passed.

- **Bug:** local or terminal fallback stops did not name their stopping condition.
- **Fix:** record the first existing condition, preserve the original error and stop order, and omit unreadable or non-string error names.

| Before | After |
| --- | --- |
| Request-entry SIGINT exits 130 without a stop reason. | One `caller_signal_aborted` diagnostic, exit 130, and no fallback request. |
| Ordinary provider failures follow existing recovery/exhaustion behavior. | The same behavior and decision records remain. |

**CI exception:** the completed [run 34212536521](https://github.com/openclaw/openclaw/actions/runs/34212536521) has one unrelated failed leaf, [checks-node-compact-large-13](https://github.com/openclaw/openclaw/actions/runs/34212536521/job/102016905158). The session-mutation performance test expected one `session.entry-replacements` mock observation and received zero after its native BEGIN/COMMIT and initial response assertions passed. The exact tested main parent leaves all affected session, transaction, observer, and test-project source unchanged; this PR's fallback code does not execute in that fixture. The six failure-conclusion runs inspected within the newest 100 completed main-push runs did not provide a matching runtime failure. The underlying instrumentation cause and introducing commit remain unconfirmed. See the [complete causal assessment](https://github.com/openclaw/openclaw/pull/141658#issuecomment-5583708611) and [exact-head approval record](https://github.com/openclaw/openclaw/pull/141658#issuecomment-5583670141). The guarded merge used that standing unrelated-failure exception; no test or check was weakened.

Whole-run deadlines and idle breakers outside this catch remain outside the repair. This does not close the broader #141604 incident.

PR #141662, opened 2026-09-07
refactor(tasks): require atomic store writes

### LINKED ISSUE #141661: Retire task-store snapshot write fallbacks
This is a maintainer-requested codebase reduction for the task and TaskFlow persistence owners.

The shipped SQLite stores already provide atomic row operations, but the internal adapter contracts also allow partial stores. Supporting those test fixtures keeps a second full-snapshot write strategy, projection fallbacks, temporary-table pruning and orphan-prune machinery in production modules. Repository callers do not expose a public custom-store or snapshot-write API. Gateway observer configuration and the benchmark's deliberate no-op persistence mode are real uses and must remain.

Consolidate the internal contract on atomic task-plus-delivery writes and flow row writes, retire the now-unused snapshot writers, and use shared complete memory fixtures plus narrow test-owned SQLite setup. Preserve snapshot reads, optional owner lookup and close, all existing SQLite schemas, lifecycle bindings, exact-owner metadata deletion, and public task behavior.

Verification should cover real SQLite create/update/delete rollback and publication ordering; retained restore, corruption, concurrent-read, owner/access, cancellation and maintenance contracts; isolated reset scope and optional-table absence; both benchmark persistence modes; selected checks and independent review. A temporary transaction bypass should make the rollback proof fail for partial durable changes.

The prepared [commit](https://github.com/openclaw/openclaw/commit/a23ae70da2e08e340ef207ca4cd2538dcb491973) removes 471 maintained code lines, including all fixture costs. Local proof is complete: 589 distinct tests, all 41 selected checks, both benchmark persistence modes and independent P2 review. Hosted CI and native landing are next.

steipete:
Closes #141661

## What Problem This Solves

Maintaining task persistence required two write strategies even though the shipped stores already support atomic row operations. Partial test adapters kept full-snapshot fallbacks, temporary-table reconciliation and orphan-pruning code in production, while some storage-failure tests only threw SQLite-shaped errors from memory stores.

## Why This Change Was Made

Require atomic task-plus-delivery writes and flow row writes, then remove the unused snapshot writers and their pruning helpers. Shared complete memory fixtures preserve fast tests; narrow test-owned reset and seeding helpers replace the former production setup API. Canonical snapshot readers, SQLite transactions, lifecycle binding and exact-owner deletion remain with their existing owners.

## User Impact

No public API, configuration, schema or task behavior changes. Gateway observers, optional owner reads and the benchmark's no-op persistence mode remain supported. The maintenance surface is smaller, and the retained persistence contract is tested against real SQLite failures.

## Evidence

589 distinct cases pass across 25 owner and consumer files. Three create/update/delete cases fail the second SQLite statement, verify durable and in-memory state remain unchanged with no observer event, then verify successful publication sees committed rows through the read-only loader. A temporary transaction bypass makes all three fail for the intended partial-write differences; original transaction code was restored.

Existing two-connection read-snapshot proof is unchanged. Scoped reset tests preserve other-family metadata and leave optional tables absent; lifecycle tests retain exact deletion, surviving bindings and read-only parity. All 41 selected check stages pass. The synthetic size-four benchmark passes in memory and durable modes: no SQLite state in memory mode, four persisted task/subagent rows in durable mode, and clean teardown in both.

The complete patch removes 270 production code lines, 196 test/support lines and five tooling lines: 471 maintained code lines / 536 physical lines, including both new test helpers. Fresh independent review through P2 is scoped-clean with no actionable findings. Hosted CI will run on the published head before landing. No live provider or remote-platform result is claimed.

### DIFF
--- a/scripts/bench-agent-concurrency-worker.ts
+++ b/scripts/bench-agent-concurrency-worker.ts
@@ -228,20 +228,15 @@ async function configureSpawnRuntime(
     taskStore.configureTaskRegistryRuntime({
       store: {
         loadSnapshot: () => ({ tasks: new Map(), deliveryStates: new Map() }),
-        saveSnapshot: () => {},
         upsertTaskWithDeliveryState: () => {},
-        upsertTask: () => {},
         deleteTaskWithDeliveryState: () => {},
-        deleteTask: () => {},
         upsertDeliveryState: () => {},
-        deleteDeliveryState: () => {},
         close: () => {},
       },
     });
     flowStore.configureTaskFlowRegistryRuntime({
       store: {
         loadSnapshot: () => ({ flows: new Map() }),
-        saveSnapshot: () => {},
         upsertFlow: () => {},
         deleteFlow: () => {},
         close: () => {},
--- a/src/audit/execution-owner-lifecycle-binding-store.ts
+++ b/src/audit/execution-owner-lifecycle-binding-store.ts
@@ -18,7 +18,7 @@ type ExecutionOwnerLifecycleKind = "cron" | "task" | "flow";
 
 type ExecutionOwnerLifecycleDatabase = Pick<
   OpenClawStateDatabase,
-  "cron_run_receipts" | "execution_owner_lifecycle_bindings" | "flow_runs" | "task_runs"
+  "execution_owner_lifecycle_bindings"
 >;
 
 const SCHEMA_START = `CREATE TABLE IF NOT EXISTS ${EXECUTION_OWNER_LIFECYCLE_BINDING_TABLE} (`;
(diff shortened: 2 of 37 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a23ae70da2e08e340ef207ca4cd2538dcb491973](https://github.com/openclaw/openclaw/pull/141662/commits/a23ae70da2e08e340ef207ca4cd2538dcb491973)
- Landed commit: [16b031d64ec3643d441c53e2643209640181ee43](https://github.com/openclaw/openclaw/commit/16b031d64ec3643d441c53e2643209640181ee43)

PR #141663, opened 2026-09-07
test(release): require completed managed survivor evidence

steipete:
## What Problem This Solves

Resolves a false-green managed upgrade survivor result: a guarded Bash function could swallow a state assertion failure and continue reporting success. The post-startup check also expected legacy skills-cache bytes to remain unchanged after runtime turn preparation had legitimately refreshed that cache.

## Why This Change Was Made

Propagate assertion failures explicitly through the guarded shell calls. Preserve strict migration equality before startup, then require an authenticated installed-CLI serving turn to complete and its unique assistant reply to be persisted in the original migrated session before accepting a fresh versioned runtime skills snapshot.

The serving probe waits on the same run within a fixed two-minute deadline; it does not retry model turns. Result-based child execution retains timeout and output bounds while keeping credential-bearing stderr, argv, and error objects out of failures. Session IDs, imported transcripts, retired-field absence, and exclusion of heavy resolved-skills data remain checked. Companion watchOS and mobile scenarios retain their separate contracts.

## User Impact

Release verification can no longer report managed survivor success after these assertion failures, and it distinguishes preserved migration data from an intentionally refreshed runtime cache. Product behavior and storage formats are unchanged.

## Evidence

- All 371 survivor tests passed after updating the companion bootstrap fixture for the new authenticated serving calls. The final result-based child-process refinement passed 95 related assertion, bootstrap, and unchanged lint-inventory tests.
- The downstream SQLite-volume idempotence assertion now receives the current serving stage. Its managed-serving-to-idempotence regression fails before the fix and passes afterward; 110 focused tests and scoped checks pass, with independent P2 review clean.
- Scoped changed checks passed. Fresh integration review caught nonterminal agent.wait timeout handling; the regression failed before correction and passes afterward. Final P2 review is clean with the full deadline-enforcing helper context.
- Causal regressions cover swallowed failures, invalid cache acceptance, queued turns, nonterminal wait timeouts, and failed-CLI credential redaction. The redaction test fails against the prior throwing child-process call and passes with result-based execution, without adding a lint-inventory exception. Final component proof used private dependencies; invalid earlier borrowed-dependency results are not relied upon.
- The actual prepared-package managed Docker replay passed in157seconds with no hidden assertion errors. Its retained receipt proves successful completion and a matching assistant reply in the original migrated session; strict migration and post-inference checks passed. This used a synthetic local provider, not live credentials.

Full release qualification remains a separate campaign with explicitly selected Code and Tooling identities; no publication occurs in this PR.

### DIFF
--- a/scripts/e2e/lib/upgrade-survivor/assertions.mjs
+++ b/scripts/e2e/lib/upgrade-survivor/assertions.mjs
@@ -1,6 +1,6 @@
 import assertStrict from "node:assert/strict";
 import { execFileSync } from "node:child_process";
-import { createHash } from "node:crypto";
+import { createHash, randomUUID } from "node:crypto";
 // Assertions for upgrade-survivor E2E scenarios.
 import fs from "node:fs";
 import path from "node:path";
@@ -677,7 +677,7 @@ function assertStateSurvived() {
     "legacy session file missing",
   );
   if (stage !== "baseline") {
-    assertSessionMetadataMigrated(stateDir);
+    assertSessionMetadataMigrated(stateDir, stage);
   }
   if (scenario === "meeting-transcripts-sqlite") {
     assertMeetingTranscriptsMigrated(stateDir, stage);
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9ddbf83a3facf3259d151cf97b7ea6795947a031](https://github.com/openclaw/openclaw/pull/141663/commits/9ddbf83a3facf3259d151cf97b7ea6795947a031)
- Landed commit: [135d2700e7b62fe419c15b1688b0147e56589ef7](https://github.com/openclaw/openclaw/commit/135d2700e7b62fe419c15b1688b0147e56589ef7)

PR #141671, opened 2026-09-07
refactor(update): simplify constant downgrade confirmation text

steipete:
Related: #141663

## What Problem This Solves

Resolves a problem where main-based CI fails the update CLI's existing 700-line limit, blocking unrelated changes such as #141663.

## Why This Change Was Made

Replace a constant two-element array joined with a newline with its identical string literal. This removes unnecessary allocation and three counted lines without changing the confirmation message or the lint rule.

## User Impact

No user-visible behavior change. The downgrade warning has exactly the same bytes.

## Evidence

- Original core lint failure: https://github.com/openclaw/openclaw/actions/runs/34170481334/job/101889773104
- Focused core lint and formatting passed.
- Compared the original joined strings with the replacement literal: exact byte equality.
- Independent P2 review found no actionable issues.
- Full scoped changed check passed, including core/test typechecks and repository guards.
- Exact-head CI is required before merge.

PR #141673, opened 2026-09-07
fix: prevent multi-select tests leaking their resize observer

steipete:
Related: #139428

## What Problem This Solves

Resolves a problem where the shared UI test run fails the marquee resize case after the multi-select suite runs.

## Why This Change Was Made

The multi-select fixture assigned a synthetic `ResizeObserver` directly to the global object and never restored it. The marquee suite then cached that observer before its resize case could install its own callback. Track the fixture with `vi.stubGlobal` and restore globals after the suite.

## User Impact

No product behavior changes. All 13 multi-select cases and seven marquee cases remain, with unchanged production code and assertions.

## Evidence

- The complete suites, forced into multi-select → marquee order in one non-isolated worker with the repository UI configuration, reproduce the original missing-callback failure: 19 passed, one failed.
- The same ordered run passes all 20 cases after the repair; marquee alone passes all seven.
- Changed-file gate and fresh P2 review pass.

The initial CI run failed on an inherited `update-command.ts` maximum-line violation. After #141671 repaired that baseline, this branch was refreshed onto the fixed main with an identical patch; new exact-head CI supplies the landing check.

### DIFF
--- a/ui/src/components/multi-select.test.ts
+++ b/ui/src/components/multi-select.test.ts
@@ -1,5 +1,5 @@
 /* @vitest-environment jsdom */
-import { afterEach, beforeAll, expect, it, vi } from "vitest";
+import { afterAll, afterEach, beforeAll, expect, it, vi } from "vitest";
 import { MultiSelect, type MultiSelectOption } from "./multi-select.ts";
 
 const MULTI_SELECT_TEST_TAG = `test-openclaw-multi-select-${crypto.randomUUID()}`;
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d2f85602be7428bcf4ad7da70cb291f720e36dd0](https://github.com/openclaw/openclaw/pull/141673/commits/d2f85602be7428bcf4ad7da70cb291f720e36dd0)
- Landed commit: [381fd7239725c55cd960d15d0d590a3444cdb8e0](https://github.com/openclaw/openclaw/commit/381fd7239725c55cd960d15d0d590a3444cdb8e0)

PR #141678, opened 2026-09-08
fix: keep update refusal exits within the lint limit

steipete:
## What Problem This Solves

Main CI fails because the update command reached 701 counted lines after the staged-target admission change, exceeding the existing 700-line lint limit. [Failing run](https://github.com/openclaw/openclaw/actions/runs/34170545757).

## Why This Change Was Made

Combine eight `await refuseUpdate(...); return;` exits into `return await refuseUpdate(...)`. The failure owner returns `Promise<never>` and always throws after its existing reporting and recovery work. Every refusal condition, argument, await, and failure remains; the explicit return preserves TypeScript narrowing. No lint suppression or packed formatting.

## User Impact

Restores CI without changing update admission, error reporting, schema checks, configuration, or service behavior. Production is eight lines smaller; no test changes.

## Evidence

- Reproduced the exact 701-line lint failure on the original file; the same targeted lint passes after the cleanup.
- Four existing refusal cases pass: exact target metadata, explicit and stored extended-stable resolution, and extended-stable Git rejection. Their error and no-mutation assertions are retained.
- Formatting and diff checks pass. Managed P2 review found no actionable issues; the reviewed update-command file is byte-identical to this patch.

### DIFF
--- a/src/cli/update-cli/update-command.ts
+++ b/src/cli/update-cli/update-command.ts
@@ -197,8 +197,7 @@ async function updateCommandInternal(
     });
 
   if (requestedChannel === "extended-stable" && installKind === "git") {
-    await refuseUpdate("unsupported_git_channel");
-    return;
+    return await refuseUpdate("unsupported_git_channel");
   }
 
   const preparedConfig = await createConfigIO({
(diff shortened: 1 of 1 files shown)


steipete:
The main CI blocker is already fixed by #141671, merged as 4140743e31f8c3500ef9ab972105d5f273107b4e. Its constant-string simplification removes three counted lines while preserving the exact warning bytes; the original 701-line file is now below the unchanged 700-line limit. I verified the merged diff and reported focused lint proof.

Closing this redundant CI repair rather than landing another change for the same blocker. The eight awaited-refusal simplifications here were locally validated, but they are not needed to unblock the separate catalog-test performance improvement.

PR #141610, opened 2026-09-07
fix(update): allow artifact updates with matching configured plugins

steipete:
## What Problem This Solves

Fixes an issue where operators updating from an explicit package artifact could be refused before staging when enabled npm plugins required a matching core version. The updater could not resolve that version from registry metadata even though the artifact contained it.

## Why This Change Was Made

Explicit artifacts now provide their version from the private staged package. Configured plugin admission runs before rehearsal, live-state preparation, or activation, with service and database ownership revalidated around the metadata request. Registry targets keep their early admission. Same-version no-ops avoid mutable preparation, and artifact dry runs explain that plugin checking is pending because previews do not stage packages.

## User Impact

Operators can update from supported explicit artifacts when matching configured plugins are available. Missing plugins still refuse the update before serving state changes. Version pins, consent requirements, and ownership checks remain enforced. The CLI and installation guides describe both the registry and artifact paths.

## Evidence

- The two CLI/execution owner suites passed 433 tests on the original repair; the full changed gate passed and independent P0–P2 review reported no findings.
- A fresh packaged plugin-update Docker lane passed in 181 seconds using the original repair commit `77d12800edc4533ffbf70c54e0846f0c177838ae`: staged admission, rehearsal and activation; denied consent with healthy rollback; accepted fresh core; no future permission; missing payload cases.
- All five production/test files are byte-identical to that reviewed and packaged repair after applying it to main `635089846dcb`; only the docs merged surrounding main content.
- The same sealed repair package also passed corrupt-plugin recovery (75 seconds), plain upgrade survival (57 seconds), and managed restart/auth survival (157 seconds). The managed lane verified a completed serving turn and strict migrated-state checks using a synthetic local provider. These later runs used corrected test fixtures and failure propagation; earlier fixture/infrastructure failures were not counted as passing proof.
- On this main-based branch, the full changed gate passed and the two owner suites passed all 433 tests (76.72 seconds). Exact-head [CI passed](https://github.com/openclaw/openclaw/actions/runs/34165776149), and ClawSweeper reported no findings. These scoped results do not claim that the complete release validation campaign is green.

PR #141680, opened 2026-09-08
fix(lint): restore update command line budget

RomneyDa:
## Summary

- resolve the CLI display name at its single use site instead of retaining a module-level constant
- restore `update-command.ts` to the enforced 700 executable-line budget
- preserve the exact user-facing update guidance

## Why

#141610 added one required package-admission input and moved this file from 700 to 701 counted lines. The resulting `max-lines` failure reproduces on current `main` and blocks otherwise unrelated PR CI.

## Validation

- exact hosted core lint stripe `1/5` passes
- `src/cli/update-cli.test.ts` and `update-command-execution.test.ts`: 433 passed
- `oxfmt --check`
- `git diff --check`

### DIFF
--- a/src/cli/update-cli/update-command.ts
+++ b/src/cli/update-cli/update-command.ts
@@ -76,7 +76,6 @@ import {
 import type { UpdateCommandRecoveryState } from "./update-command-service.js";
 import { withUpdateFailureTriage } from "./update-command-triage.js";
 
-const CLI_NAME = resolveCliName();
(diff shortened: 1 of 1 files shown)


RomneyDa:
Closing as superseded: current `main` is already at 698 counted lines under the unchanged 700-line limit through a separate simplification. The original cross-PR lint blocker no longer exists, so this cleanup is no longer needed.

PR #141686, opened 2026-09-08
fix(memory): split embedding requests at provider input-array limits

### LINKED ISSUE #139040: [Bug]: [memory] sqlite-backed corpus chunks are not split to the 64-item embedding limit (HTTP 400, Zhipu code 1214)

vildevev:
## What Problem This Solves

Resolves a problem where agents with sqlite-backed session corpora could never build their memory vector index when using remote embedding providers that cap how many texts one request may contain. Zhipu embedding models accept at most 64 texts per request and reject larger arrays with HTTP 400 / code 1214, so every memory sync failed for these agents (#139040, observed on 2026.9.1), while file-backed agents only worked by coincidence: their markdown chunks are large enough that the existing byte-only batch budget kept item counts under the provider cap.

## Why This Change Was Made

Memory request batching (`buildMemoryEmbeddingBatches`) split requests by estimated bytes only and had no notion of the provider's per-request input-array limit. The fix models that limit where the byte budget already lives:

- `EmbeddingProvider` gains an optional `maxInputsPerRequest` declaration, re-exported through the plugin SDK so plugin-registered providers can declare it too.
- `resolveEmbeddingMaxInputsPerRequest` resolves a conservative 64-item default for providers that declare nothing.
- The native OpenAI provider declares its documented 2048-item array cap **only for the native `api.openai.com` endpoint**; custom endpoints behind the OpenAI adapter keep the conservative 64-item fallback. `createRemoteEmbeddingProvider` passes declarations through for other remote providers.
- `buildMemoryEmbeddingBatches` closes a batch at the resolved cap in addition to the byte budget, so memory sync can send more requests but never an oversized array.

No configuration surface changes; the field is optional provider-contract surface and existing adapters keep working with the conservative default.

## User Impact

Agents with sqlite-backed session corpora can build vector indexes against capped providers (Zhipu embedding-2/3, any OpenAI-compatible endpoint with a ≤64-item array limit) instead of failing every sync and losing vector recall. File-backed and local providers keep working, bounded by the same conservative cap. Native OpenAI API users keep their previous request granularity; custom endpoints through the OpenAI adapter now get the safe 64-item default.

Throughput note, benchmarked: for a tiny-chunk corpus (1,200 curated one-line entries, ~45 bytes per chunk) fresh indexing sent **19 embedding requests** (18 at the 64-item cap plus a 48-item tail) and completed in **3.8s** against a loopback endpoint; the byte-only policy would have grouped the same corpus into ~7 requests of ~175 items, but every one of those is rejected outright by a 64-capped provider — that failure is exactly the bug being fixed. For typical markdown corpora (hundreds of bytes per chunk) the 8,000-byte budget binds before the item cap, so request counts are unchanged. Native OpenAI endpoints declare 2048 items and keep their previous grouping, and the OpenAI batch-API path is untouched. 429 rate-limit responses already retry with backoff through the existing memory embedding retry policy.

## Evidence

**Live memory indexing against a Zhipu-style capped endpoint** (mock embeddings server rejecting `input.length > 64` with `HTTP 400 {"error":{"code":"1214","message":"input array max 64"}}`, isolated `OPENCLAW_STATE_DIR`, loopback only):

- Workspace `MEMORY.md` with 150 short curated entries → 150 indexed chunks.
- `pnpm openclaw memory status --index --agent main` after the fix: `Embeddings: ready`, `memory · 1/1 files · 150 chunks`, `Vector store: indexed`, `FTS: ready`, embedding cache 150 entries — no 400s.
- Captured HTTP requests (item counts per request): `1, 64, 64, 22` — the leading 1 is the provider availability probe; **max 64 items per request**, exactly at the declared cap.
- Re-running `memory status --index`: zero new embedding requests (only a fresh 1-item probe) — the embedding cache and existing index rows are reused unchanged, demonstrating fresh-indexing and existing-index compatibility.

**Data model and upgrade compatibility:** no SQLite schema or stored-format change — no new tables or columns, no index metadata or chunking-version bump, and identical chunk texts, hashes, index rows, and embedding-cache keys with and without this change. Batching only groups provider HTTP calls. Concretely: an index and embedding cache built before this change load and serve as-is, and re-indexing unchanged content after the change issues **zero** provider requests because every chunk hits the existing cache (observed above); only uncached chunks are batched under the new cap. Downgrade is symmetrical: index and cache files written under this change carry the same schema and chunking version, so an earlier runtime reads them unchanged. The repository's own `native state schema version guard` passes on this branch at the unchanged version (v16), confirming no stored-schema bump.

**Tests:** `buildMemoryEmbeddingBatches` item-cap boundary (`[64, 36]`, per-item, oversized-chunk, and uncapped cases — the uncapped case produces the unbounded 100-item batch shape that triggers the reported 400), `resolveEmbeddingMaxInputsPerRequest` fallbacks, `createRemoteEmbeddingProvider` declaration passthrough, and the OpenAI adapter declaring 2048 only for the native endpoint (`undefined` for custom endpoints). Adjacent suites green: `manager-embedding-ops.retry.test.ts`, `embeddings.test.ts`, generic-provider bridge, `openai-compatible-embedding-provider.test.ts`, embedding-provider contract tests.

**Checks:** `pnpm check:changed` green locally (typecheck lanes, oxlint, plugin-sdk boundary dts, import-cycle and store guards); `oxfmt --check` and `git diff --check` clean.

Closes #139040


vildevev:
Two required shards failed for reasons unrelated to this diff; both pass locally on this branch:

- `checks-node-compact-large-11` — `src/agents/subagents/spawn/acp-spawn.authority.test.ts` "transfers initialized ACP work only from its live parent" hit a soft timing assertion ("cleanup completes before spawn returns"). The full file passes locally on this branch (10/10), and the change here touches only memory embedding batching.
- `checks-node-compact-small-36` — the shard died in checkout setup: `Release ancestry exceeded its total time budget` / `Git ownership/setup failed (RuntimeError)`, before any test ran.

I don't have rerun rights on upstream workflow runs, so a maintainer rerun (or branch update) should clear both.

vildevev:
Pushed e54f407e9eb addressing the review findings: the 2048-item declaration now applies only to the native api.openai.com endpoint (custom endpoints through the OpenAI adapter keep the 64-item fallback, covered by adapter tests), and the PR body now carries the live indexing proof (150 chunks → requests of 1/64/64/22 against a 64-item-capped endpoint, cache reuse verified on re-run). @clawsweeper re-review

vildevev:
Rebased onto current `main` (`e321e32abde`, head `9e1ef4608da`) — the earlier merge conflict is resolved, and the docs paragraph moved with main's doc restructure into `docs/plugins/sdk-provider-plugins/media-and-search.md`. No code changes versus the reviewed head `e54f407e9eb` beyond that rebase.

CI on the new head: 167 pass / 38 skipped, one shard failure unrelated to this diff — `checks-node-changed-extensions-bundle-29` timed out after 120s in `extensions/telegram/src/sticker-cache.selection.test.ts` ("keeps minimax on VLM…"), which passes locally on this branch in ~23s and touches nothing this PR changes (memory embedding batching). The `openclaw/ci-gate` failure is just that shard's aggregate.

vildevev:
Benchmarked the tradeoff called out in the latest review, on the same isolated loopback harness with a 1,200-entry corpus (~45-byte chunks): fresh embed = **19 embedding requests** (18×64 + 48) in **3.8s**; the byte-only policy would have sent ~7 requests of ~175 items, all rejected by a 64-capped provider. Re-indexing unchanged content = **0 provider requests** (full cache reuse), which is also the data-model compatibility proof: same schema, same chunking version, same cache keys — pre-change indexes and caches load as-is, downgrades read new files unchanged. Numbers are in the PR body. @clawsweeper re-review

vildevev:
Two more required shards failed on tests unrelated to this diff; both files pass locally on this branch:

- `checks-node-changed-extensions-bundle-13` — `extensions/discord/src/monitor/native-command.reset.test.ts` ("resets the channel and acknowledges a matching sender entry") timed out after 120s.
- `checks-node-changed-extensions-bundle-3` — `extensions/signal/src/monitor.tool-result.sends-tool-summaries-responseprefix.test.ts` ("skips tool summaries with responsePrefix") failed its drain wait ("Signal tool-result ingress still active: 0 pending, 1 claimed").

This is now the third consecutive run where a *different* set of timing-sensitive channel-extension tests has failed (telegram sticker-cache, then discord + signal), each passing locally and touching nothing this PR changes — memory embedding batching. `ci-gate` fails only as the aggregate of these shards.

PR #141689, opened 2026-09-08
fix(e2e): preserve typed onboarding install diagnostics

vincentkoc:
## Root cause

The shared package-install helper exits the scenario process itself when npm fails. That bypasses the caller-owned ERR trap, so scenario cleanup removes its temporary logs before the typed-onboarding diagnostics can publish.

## Fix

Return the original npm status from `openclaw_e2e_install_package`. The scenario remains fail-fast, but its existing ERR/EXIT lifecycle now owns diagnostics and cleanup in the intended order. No new helper, protocol, environment variable, or compatibility branch.

## Proof

- Pre-fix focused regression: status collapsed to 1 and caller diagnostics did not run.
- Post-fix: `test/scripts/openclaw-e2e-instance.test.ts` 40/40.
- Bash syntax, Oxfmt, OXLint, diff check pass.
- AutoReview P1 scoped-clean (0.97).

## LOC

- Production/tooling: +1/-1
- Tests: +26/-1

Rewrites the former +833-line diagnostic subsystem into the existing lifecycle owner.

### DIFF
--- a/scripts/lib/openclaw-e2e-instance.sh
+++ b/scripts/lib/openclaw-e2e-instance.sh
@@ -267,7 +267,7 @@ openclaw_e2e_install_package() {
     if [ -f "$log_file" ]; then
       openclaw_e2e_print_log "$log_file" >&2
     fi
-    exit 1
+    return "$install_status"
   fi
 }
 openclaw_e2e_find_dep_package() {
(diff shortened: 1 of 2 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
@clawsweeper re-review

vincentkoc:
@clawsweeper re-review

Addressed the exact-head P2 at `7d7de56f1c05c53e527a94d525130f21b442b184`: capture now flushes already-read bounded bytes on SIGTERM/SIGINT before re-raising the signal. The regression preserves timeout status 124 and verifies one published retained tail with no omission marker. Scoped suites pass 45/45 and 24/24.

RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

PR #141692, opened 2026-09-08
fix(doctor): release plugin resources after inspection

steipete:
Related: #140674

## What Problem This Solves

Doctor can execute plugin registration while inspecting a configured context engine, but its existing raw loader result has no release boundary for resources captured during registration.

## Why This Change Was Made

Add an explicit uncached, non-activating inspection acquisition and release it when Doctor discovery finishes. Optional resource-only disposers remain owned through rollback and actual invalid asynchronous registration work; release waits for cleanup. Prototype methods retain their original receiver and normalized registration ID.

Existing raw loader returns, cache behavior, Gateway cleanup notifications, and active runtime engine handling stay unchanged. Discovery-only context engines still report offline inspection unavailable, and their factories are never invoked. This is the first bounded ownership slice from the aggregate PR.

## User Impact

Plugins can release registration-owned resources after an explicitly owned inspection by providing the documented optional disposer. Doctor waits for that cleanup. Disposal is separate from disabling a plugin or deleting durable state; older raw loader lifetimes still use their existing cleanup contract.

## Evidence

- 59 focused native, Doctor, raw-loader, and laziness cases passed. The final mechanical cleanup was rechecked with 23 native/Doctor cases and 13 format, type, lint, and guard phases.
- Initial canonical checks caught two missing-braces errors in test helpers. Those were fixed; the affected checks and previously unrun guard steps all passed. Production dead-export scans passed.
- Full build passed in 285.793 seconds with source and index unchanged.
- The original prototype-capture defect failed the native rollback regression before correction.
- A synthetic probe invoked the actual built Doctor owner with a prototype/private-field disposer and real SQLite file. Doctor stayed pending until disposal finished; the probe observed exactly one disposal, no cleanup notification or factory invocation, native closure, and successful read-only reopen with its row preserved. This covers the built owner entry, not a full CLI diagnostics run or real provider execution.
- Deslop and independent P0–P2 Codex review completed without actionable findings.

No schema, data representation, dependency, configuration option, protocol version, or host floor changes.

## Review Disposition

Retain the documented inspection-only callback contract. This is the accepted resource-only registration cleanup scope: disposers own resources captured by that registration and must not close resources belonging to a raw-loader or Gateway registration. The native isolation case keeps the cached raw registration usable while the separate inspection closes, and the existing cleanup notification remains untouched.

The automatic data-model flag is a path-classification false positive. The complete diff changes no schema, migration, stored value, query, retention, or repair logic; Doctor's existing configuration-repair function is unchanged. Its inspection path gains a finite resource lifetime. Native closure plus read-only reopen verifies preserved rows; there is no data conversion needing an upgrade migration.

### DIFF
--- a/docs/plugins/sdk-overview.md
+++ b/docs/plugins/sdk-overview.md
@@ -498,6 +498,14 @@ plugins.
 | `api.session.controls.registerSessionAction(...)`                                    | Typed session actions clients can dispatch through the Gateway                                                                                             |
 | `api.registerBoardWidgetContentKind(...)`                                            | Sandboxed board widget source validation, renderer resources, and document composition                                                                     |
 
+Runtime lifecycle registrations can also supply `dispose()` for resources captured
+by that registration. Explicitly owned, uncached plugin inspections invoke it on
+rollback or release and await cleanup, including invalid asynchronous registration
+work. Doctor uses this ownership when loading a context engine for discovery; it
+still does not invoke a discovery-only engine factory. `dispose()` must not delete
+durable state or disable another registration. Existing raw loader and Gateway
+lifetimes do not gain automatic disposal: keep their `cleanup(ctx)` behavior.
+
 `registerBoardWidgetContentKind(...)` is for plugins that own a declarative
 widget source format. The registration supplies a globally unique lowercase
 `kind`, a short label, one capability-scoped plugin surface plus its renderer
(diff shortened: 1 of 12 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9e0b2838bed7871868e23ce90252067ec253c89b](https://github.com/openclaw/openclaw/pull/141692/commits/9e0b2838bed7871868e23ce90252067ec253c89b)
- Landed commit: [5605b92d2e6be08c9c1fd0347e69f9e2a1d0dff9](https://github.com/openclaw/openclaw/commit/5605b92d2e6be08c9c1fd0347e69f9e2a1d0dff9)

PR #141704, opened 2026-09-08
fix(test): make TUI startup frame ordering deterministic

steipete:
Related: #139428

## What Problem This Solves

The PTY startup test could finish its artificial 400ms history delay without observing a startup frame, then fail while searching raw terminal output. Elapsed time alone does not establish that a frame was rendered while initialization remained pending.

## Why This Change Was Made

Hold fixture history initialization behind an owned release barrier. Require a completed startup frame, reject any earlier ready frame using the exact same captured snapshot, then release history and require a completed ready frame. The fixture's disposal also releases the barrier when the case is filtered out or setup fails.

The existing startup test-support owner holds the generated fixture fragment and assertions. Production rendering, the 20s/25s deadlines, and line limits remain unchanged.

## User Impact

More deterministic startup regression coverage; no product behavior change or overall runtime improvement claim. This adds 73 net test/support lines.

## Evidence

- The original case passed locally with complete PTY and phase captures. That pass does not establish full-load reliability or identify the exact mechanism of the earlier frozen-run failure.
- The full harness and session-identity sibling passed all 79 cases before the final historical-frame assertion was added. After that adjustment, the precise startup plus three existing frame-reader contracts passed (4 cases).
- Suppressing real startup rendering failed the completed-frame oracle. Emitting a real premature ready frame failed specifically at the new historical-frame assertion. Production was restored after both controls.
- Filtered-case and setup-failure controls verified barrier release during disposal and termination of all four fixture PTYs.
- Hosted architecture validation exposed a type-only dependency cycle; using the existing leaf fixture type removed it without changing emitted JavaScript. Madge, focused tests, mapped checks, and fresh review passed after the correction.
- The initial changed gate failed only file line limits. Moving the unchanged generated fragment and assertion exercise into the existing startup support file resolved this without weakening limits. Final mapped checks and fresh independent P2 review passed.

These are local Node 26 validation scopes, not a Linux full-suite or performance acceptance result.

### DIFF
--- a/src/tui/tui-pty-harness-fixture-test-support.ts
+++ b/src/tui/tui-pty-harness-fixture-test-support.ts
@@ -31,10 +31,15 @@ export async function disposeActiveTuiFixtures(): Promise<void> {
   }
 }
 
-export async function startTuiFixture(opts: { env?: NodeJS.ProcessEnv; execPath?: string } = {}) {
+export async function startTuiFixture(
+  opts: { env?: NodeJS.ProcessEnv; execPath?: string; holdStartupHistory?: boolean } = {},
+) {
   const tempDir = await mkdtemp(path.join(tmpdir(), "openclaw-tui-pty-"));
   const scriptPath = await writeTuiPtyFixtureScript(tempDir);
   const logPath = path.join(tempDir, "fixture-log.jsonl");
+  const startupHistoryReleasePath = opts.holdStartupHistory
+    ? path.join(tempDir, "startup-history.release")
+    : undefined;
   const execPath = opts.execPath ?? process.execPath;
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [72e94652fba0b66051decc470cb7b474f0d64542](https://github.com/openclaw/openclaw/pull/141704/commits/72e94652fba0b66051decc470cb7b474f0d64542)
- Landed commit: [cad71bedd44667e2d481fb68e7e88159a8d3f048](https://github.com/openclaw/openclaw/commit/cad71bedd44667e2d481fb68e7e88159a8d3f048)

Issue #128368, opened 2026-08-23
Settled-tool-no-answer turns skip the tools-enabled continuation retries that reasoning-only/empty-response turns already have


adinballew:
## Summary

`incomplete-turn-recovery.ts` already has three bounded, tools-enabled continuation retries for turns that end without visible text — `reasoningOnlyAttempts` (max 2), `emptyResponseAttempts` (max 1), `compactionContinuationAttempts` (max 1) — each using `activateInternalPrompt(...)` + loop `continue` to let the model keep working normally, with tools available, governed by the same timeout/loop-detection safety nets every turn already has.

The one exception is the "settled tool calls, no final answer" case (`resolveSettledToolTerminalContinuationInstruction` / `SETTLED_TOOL_TERMINAL_CONTINUATION_INSTRUCTION`): it skips straight to the isolated, `disableTools: true` finalization instead. There's no equivalent bounded tools-enabled retry for it.

## Impact

When a turn ends after tool calls with no visible answer — a very common case in long tool-heavy sessions (the model just needs one more `read`/`exec`/etc. to actually finish) — the isolated finalizer runs with `disableTools: true`, but its instruction text only says "do not repeat completed tool calls," never that tools are unavailable at all. The model naturally attempts a *new* tool call to keep working, which resolves against the empty tool list and returns `Tool <name> not found`. That denial gets narrated back into the visible transcript by the model itself (e.g. "The tool layer returned transient errors. Let me retry."), and enough of it in one completion trips `"reported capability activity"`, closing the turn and triggering the bounded finalization retry.

Net effect from a user's perspective: a multi-minute visible "outage" where tool calls appear broken, when in fact the *session's* tools are healthy — only the finalizer's own placeholder completion (deliberately tool-free) is being denied a tool it wasn't told it couldn't use.

## Confirmed via live trace evidence

Added temporary trace logging at the tool-resolution site (`Tool <name> not found`, in `validateToolCallForBatchAdmission`) tagging each `currentContext` object via a `WeakMap`. Across 2 independent live incidents (9 samples total):

- Every trace had a **distinct** `currentContext` object (only one repeat, from 2 tool calls in the same completion batch) — ruling out a shared/leaked context bug.
- Every trace had `ctxToolsLen=0` and the same minimal shape (`systemPrompt|messages|tools`) — consistent with a small, freshly built, deliberately tools-free completion request (the isolated finalizer), not a large real-session context that transiently emptied.
[...]

PR #141705, opened 2026-09-08
fix(agents): tell the settled-turn finalizer that tools are disabled

fabiolr:
## What Problem This Solves

The settled-turn finalization pass runs with `disableTools: true`, but the instruction it is handed never says so:

```ts
// src/agents/embedded-agent-runner/run/incomplete-turn-recovery.ts
const SETTLED_TOOL_TERMINAL_CONTINUATION_INSTRUCTION =
  "The previous assistant turn completed its tool calls but did not produce a user-visible answer. Continue from the current transcript and produce the final user-visible answer now. Do not repeat completed tool calls or restart from scratch.";
```

"Do not repeat completed tool calls" reads as *don't redo work*, not *you have no tools*. So the model reaches for one, the call resolves against an empty tool list and returns `Tool <name> not found`, and that **denied** call registers as capability activity. `projectSettledTurnFinalizationAttemptResult` then throws `Settled-turn finalization attempt reported capability activity` and discards the whole result — **including a complete, correct answer written in the same completion**. The user receives only:

> The tool run finished, but no final summary was produced. I did not repeat any completed actions.

The work had already succeeded. Only the report is lost, which makes it look to the user like the task failed.

This is the discarded-answer half of #128368.

## Evidence

Running as a local patch on a live 2026.9.1 deployment since 2026-09-05 19:16 ET, measured from the gateway journal:

| window | finalizer triggered | `reported capability activity` | failed |
|---|---|---|---|
| before | 3 | **3** | 3 |
| after | **12** | **0** | 1 |

The guard fired on **every** finalization before the change and on **none of twelve** after.

```
# since the patch
$ journalctl -u openclaw-gateway --since "2026-09-05 19:16" | grep -c "reported capability activity"
0

# the only post-patch failure, a different error:
$ journalctl -u openclaw-gateway --since "2026-09-05 19:16" \
    | grep "settled-turn finalization failed" | sed -E 's/.*error=//' | sort | uniq -c
      1 Settled-turn finalization attempt did not complete successfully — using terminal fallback reply
```

That remaining failure is the earlier guard (terminal not ok / compaction / prompt timeout), not the tools guard, and it **discarded no answer** — the count of recoverable lost replies has not moved since the patch.

Before the change, the trace of a real occurrence (transcript event ids, one session):

```
[209] assistant  -> tool call
[210] toolResult "Tool automations not found"
[211] assistant  -> tool call
[212] toolResult "Tool exec not found"
[213] assistant  "<a complete, correct summary of the work>"   <- the real answer
[215] assistant  "The tool run finished, but no final summary was produced..."
[216] assistant  "The tool run finished, but no final summary was produced..."
```

Only `[216]` carried a delivery marker. `exec` is not a hallucinated tool name — it had 1458 successful calls in the same agent database; it "does not exist" only inside the finalization pass.

Full test suite not run locally (monorepo toolchain not installed here); CI covers that. I grepped the tree for every occurrence of the literal and updated each site that asserts equality — see below.

## Why this is safe

`resolveSettledToolTerminalContinuationInstruction` has exactly one caller, `resolveSettledTurnFinalizationRequest` (`terminal-resolution.ts`), whose only caller is `settled-turn-finalization.ts`, which always runs the attempt with `disableTools: true`. The added sentence is therefore true everywhere it is used and cannot mislead a tools-enabled retry.

**No security boundary changes.** Tools stay disabled, and the `toolMetas` guard in `agents/harness/settled-turn-finalization-result.ts` is untouched. This only stops the model walking into it.

That is deliberate. The review on #128368 blocked the tools-enabled-retry proposal as "changes a documented security boundary and needs maintainer direction", and I agree. I also floated excluding denied calls from `toolMetas` — on reading the guard, that understates it: the same denied call also lights `itemLifecycle.startedCount`, `itemLifecycle.completedCount` and `lastToolError`. That is a three-signal relaxation and still wants maintainer direction. This PR does not attempt it.

## Changes

- `incomplete-turn-recovery.ts` — the instruction gains one sentence.
- Four sites that assert the literal, updated to match: `settled-tool-evidence.test.ts`, `settled-turn-finalization.test.ts`, `run.prompt-timeout-fallback.test-support.ts`, `qa-lab/mock-openai/server.test.ts`.
- A new test asserting the *property* rather than the literal, so a future edit that drops the sentence fails with a reason rather than a string diff.

The qa-lab **needles** (`QA_SETTLED_TOOL_TERMINAL_CONTINUATION_NEEDLE`, `scenario-catalog.inference.test.ts`) contain only the unchanged opening sentence and still match, so they are intentionally left alone.

## CI note

`checks-node-compact-small-36` fails on `test/scripts/ci-git-owner.test.ts > accepts a newly proven relation when deepening only moves a boundary` (exit 125, `Release ancestry fetch completed without ancestry progress`, `warning: filtering not recognized by server`). That is release-ancestry git tooling and is unrelated to this diff, which touches only agent finalizer instruction strings and their assertions. The same shard passed 155/156, and the `Critical Quality (agent-runtime-boundary)` shard plus all six QA Smoke profiles are green. Happy to rebase if a re-run is wanted.

https://claude.ai/code/session_01LSQ1TiVbpzFfuU2p1LJX3p

### DIFF
--- a/extensions/qa-lab/src/providers/mock-openai/server.test.ts
+++ b/extensions/qa-lab/src/providers/mock-openai/server.test.ts
@@ -50,7 +50,7 @@ const QA_REASONING_ONLY_RETRY_INSTRUCTION =
 const QA_EMPTY_RESPONSE_RETRY_INSTRUCTION =
   "The previous attempt did not produce a user-visible answer. Continue from the current state and produce the visible answer now. Do not restart from scratch.";
 const QA_SETTLED_TOOL_TERMINAL_CONTINUATION_INSTRUCTION =
-  "The previous assistant turn completed its tool calls but did not produce a user-visible answer. Continue from the current transcript and produce the final user-visible answer now. Do not repeat completed tool calls or restart from scratch.";
+  "The previous assistant turn completed its tool calls but did not produce a user-visible answer. Continue from the current transcript and produce the final user-visible answer now. Do not repeat completed tool calls or restart from scratch. Tools are unavailable in this step: it is a text-only pass, so reply with plain text and do not attempt any tool call.";
 const QA_COMPACTION_RETRY_CODE_MODE_WRITE_RESULT = {
   status: "completed",
   value: {
--- a/src/agents/embedded-agent-runner/run.prompt-timeout-fallback.test-support.ts
+++ b/src/agents/embedded-agent-runner/run.prompt-timeout-fallback.test-support.ts
@@ -147,7 +147,7 @@ describe("runEmbeddedAgent prompt timeout fallback handoff", () => {
       disableTools: true,
       skipPreparedUserTurnMessage: true,
       prompt:
-        "The previous assistant turn completed its tool calls but did not produce a user-visible answer. Continue from the current transcript and produce the final user-visible answer now. Do not repeat completed tool calls or restart from scratch.",
+        "The previous assistant turn completed its tool calls but did not produce a user-visible answer. Continue from the current transcript and produce the final user-visible answer now. Do not repeat completed tool calls or restart from scratch. Tools are unavailable in this step: it is a text-only pass, so reply with plain text and do not attempt any tool call.",
     });
     expect(mockedGetApiKeyForModel).toHaveBeenCalledTimes(1);
   });
--- a/src/agents/embedded-agent-runner/run/incomplete-turn-recovery.ts
+++ b/src/agents/embedded-agent-runner/run/incomplete-turn-recovery.ts
@@ -36,7 +36,7 @@ const REASONING_ONLY_RETRY_INSTRUCTION =
 const EMPTY_RESPONSE_RETRY_INSTRUCTION =
   "The previous attempt did not produce a user-visible answer. Continue from the current state and produce the visible answer now. Do not restart from scratch.";
 const SETTLED_TOOL_TERMINAL_CONTINUATION_INSTRUCTION =
-  "The previous assistant turn completed its tool calls but did not produce a user-visible answer. Continue from the current transcript and produce the final user-visible answer now. Do not repeat completed tool calls or restart from scratch.";
+  "The previous assistant turn completed its tool calls but did not produce a user-visible answer. Continue from the current transcript and produce the final user-visible answer now. Do not repeat completed tool calls or restart from scratch. Tools are unavailable in this step: it is a text-only pass, so reply with plain text and do not attempt any tool call.";
 
 export function shouldRetrySilentErrorAssistantTurn(params: {
   attempt: Pick<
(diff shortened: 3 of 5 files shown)


fabiolr:
CI note, with the evidence rather than a claim of "flaky".

`PR context and evidence` was a real miss on my side — the body lacked the required authored sections. Fixed; it passes now.

The remaining two reds are both outside this diff, which touches only the finalizer instruction string and the four sites that assert it:

**1. `checks-node-compact-small-36` — reproducible, not flaky.** Failed identically on both runs:

```
FAIL tooling test/scripts/ci-git-owner.test.ts
  > accepts a newly proven relation when deepening only moves a boundary
AssertionError: ::error::Release ancestry fetch completed without ancestry progress.
warning: filtering not recognized by server, ignoring
: expected 125 to be +0
  ❯ expectPolicySuccess test/scripts/ci-git-owner.test.ts:221:63
Tests  1 failed | 155 passed (156)
```

Same test, same `exit 125`, same message on `aef664af` and `c6b8a2d6`. It exercises release-ancestry git tooling against synthetic `file://` origins it builds in `/tmp`, so it does not read this branch's history. The `warning: filtering not recognized by server` line suggests the local transport rejects the partial-clone filter in that environment. I have no way to reproduce or re-run it from a fork — `gh run rerun` needs admin on this repo.

**2. `checks-node-changed-extensions-bundle-11` — flaky.** Appeared on the second run only:

```
FAIL extension-discord extensions/discord/src/voice/transcripts-conversation.test.ts
  > records through stalled agent-proxy admission and retires only overflowing conversation audio
AssertionError: expected "vi.fn()" to be called 5 times, but got 4 times
Tests  1 failed | 3690 passed (3691)
```

A call-count timing assertion in Discord voice.

Green throughout on both runs: `Critical Quality (agent-runtime-boundary)`, all six `QA Smoke CI` profiles, every `Security High` shard, `build-artifacts`, `dependency-guard`.

Happy to rebase onto current main if that would help, or to leave it for a maintainer re-run.

RomneyDa:
@clawsweeper re-review

Rebased onto current `main` at `c226caba07b`; resolved only a stale test-layout conflict. The owner behavior remains the same at `4c74fbfba7b`: the sole tools-disabled finalization instruction now tells the model it is text-only. Focused owner, provider-error, integrated prompt-timeout, and QA mock-provider coverage passed; independent autoreview found no P0–P2 findings.

RomneyDa:
@clawsweeper re-review

Exact-head refresh at `f4b01527df4`: resolved the current-main canonical assertion moved to `terminal-resolution.settled-request.test.ts`. The previous queued review no-oped before the fork head advanced. Focused red/green: this test initially failed on the stale expected prompt and now passes with settled-tool evidence (41 tests); QA mock-provider 289 passed; formatting, core lint/type checks, diff check, and independent P0–P2 autoreview are clean.

PR #141706, opened 2026-09-08
fix(talk): await connection cleanup during shutdown

steipete:
Related: #140674

## What Problem This Solves

Talk connection cleanup deleted callback records before running them and ignored returned promises. Failures lost their cleanup owner, and shutdown could finish while browser provider cleanup was still pending.

## Why This Change Was Made

Retain one pending or failed cleanup owner per connection and relay kind. Join actual completion, run a queued replacement only after the original succeeds, and remove the record after successful cleanup. Browser disconnect now returns and joins its existing close-and-warn promises through the singleton shutdown drain.

## User Impact

Socket disconnect stays non-blocking while shutdown waits for returned cleanup work. Browser provider terminal errors, warnings, replacement, and late-adoption policy remain unchanged. This does not add physical browser-close retry or claim that relay callbacks which return no promise are fully drained.

## Evidence

- Both actual browser-owner variants failed before the callback change because shutdown resolved before provider close finished.
- 36 owner tests passed, including the complete browser-control suite.
- 11 native checks passed with all six SQLite instances closed by their actual owners and none left for manual fixture cleanup. The real browser-control owner was exercised through disconnect and singleton drain.
- Formatting, typed lint, and the full canonical changed gate passed; the gate took 399.95 seconds.
- Deslop and independent P0–P2 Codex review completed without actionable findings.
- Refreshed onto main to include its unrelated UI lint repair. Range comparison and all four source hashes are identical, runtime contracts and dependency lock are unchanged, and package changes are scripts only. Earlier local proof retains its original source/base attribution; exact-head hosted CI validates the refreshed branch.

No schema, data, configuration, dependency, or public protocol changes.

Hosted CI initially hit the existing 60-second child-process deadline in the unchanged update-cleanup refusal test. Its refusal path exits directly and does not await Talk drainage. Only that failed job was rerun; it and all required checks then passed on the same head, with no source, assertion, or timeout changes.

### DIFF
--- a/src/gateway/talk-client-gateway-control.test.ts
+++ b/src/gateway/talk-client-gateway-control.test.ts
@@ -1,6 +1,7 @@
 import { setImmediate as nextEventLoopTurn } from "node:timers/promises";
 import { describe, expect, it, vi } from "vitest";
 import { createDeferred } from "../../test/helpers/promise.js";
+import { drainGlobalSingletonLifecycleState } from "../shared/global-singleton.js";
 import type { RealtimeVoiceBridge } from "../talk/provider-types.js";
 import {
   closeTalkClientGatewayControlSession,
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [169626575b760e5bad01b1ff9d5ff8d919dce457](https://github.com/openclaw/openclaw/pull/141706/commits/169626575b760e5bad01b1ff9d5ff8d919dce457)
- Landed commit: [e94d0e97a6b927c9451814835cdd62a27e51e0ac](https://github.com/openclaw/openclaw/commit/e94d0e97a6b927c9451814835cdd62a27e51e0ac)

PR #141716, opened 2026-09-08
test(vitest): size every project against one host reading

steipete:
## What Problem This Solves

`pnpm test src/acp` — and `src/skills`, `src/plugins`, and in principle any lane — intermittently runs **zero tests** on a busy developer machine:

```
Test Files  no tests
Tests       no tests
[vitest] UNHANDLED ERRORS (1): Error: Projects "unit" and "unit-fast"
  have different 'maxWorkers' but same 'sequence.groupOrder'.
```

Measured at 4 failures in 6 runs on `src/acp`, and observed across three sweeps with three distinct project pairs (`unit`/`unit-fast`, `unit-fast`/`unit-fast-isolated`, `contracts-plugin`/`unit-fast`). It exits nonzero, so it is not silent green, but a developer sees a config error and no coverage.

Root cause: `resolveLocalVitestScheduling` and `resolveLocalFullSuiteProfile` take `system: VitestHostInfo = detectVitestHostInfo()` as a default parameter, and `detectVitestHostInfo()` reads `os.loadavg()` fresh on every call. Worker sizing derives from that load average (`loadRatio >= 1` halves the count, `>= 0.75` reduces it), so two Vitest projects that resolve their scheduling moments apart on a loaded host can land on different `maxWorkers`. Vitest 4 rejects a run whose projects share `sequence.groupOrder` but disagree on `maxWorkers`, and the whole selection collects nothing.

CI never sees this: `resolveSharedVitestWorkerConfig` short-circuits under `isCI` to a constant `maxWorkers: isWindows ? 2 : 3` with no load term. It is a local-developer-only defect, which is why it has survived.

## Why This Change Was Made

Take one host snapshot per process for scheduling. `schedulingHostInfo()` memoizes `detectVitestHostInfo()` and now backs both scheduling defaults, so every project in a process sizes against one reading. `detectVitestHostInfo()` itself stays live on purpose: `scripts/lib/vitest-resource-reporter.mts` consumes it for reporting current host state, and blanket memoization would freeze its output.

Explicit `system` arguments are unaffected; the existing table-driven tests keep passing their own host shapes.

## User Impact

Developers running `pnpm test <path>` on a loaded machine no longer intermittently get a zero-test run with a `maxWorkers`/`groupOrder` error. No runtime, CLI, config, or public-contract change; production LOC delta is 0 (both touched files are test infrastructure).

## Evidence

Mechanism, reproduced standalone with cpu and memory pinned so only load moves — two identical calls:

```
load=0  -> {"maxWorkers":12,"fileParallelism":true,"throttledBySystem":false}
load=64 -> {"maxWorkers":3, "fileParallelism":true,"throttledBySystem":true}
DIVERGES: true
```

Regression test `sizes every project against one host reading while the load average moves` pins cpu, total and free memory, moves only the load average between two resolutions, and guards against a vacuous pass by asserting the stub is driving the reading.

Pre-fix (lib reverted, test kept), on `53ff0867149`:

```
→ expected 3 to be 12 // Object.is equality
Tests  1 failed | 21 passed (22)
```

With the fix:

```
Tests  22 passed (22)
```

`node scripts/check-changed.mjs` exit 0 (oxfmt, lint, coercion/docker-e2e/http2 guards).

A note on evidence quality, because it shaped this test: my first regression test passed on pre-fix code. This host had 1.19 GB free memory at the time, so the memory-pressure clamp pinned workers to 1 and completely masked the load term. The test therefore pins memory explicitly rather than trusting the host.

Independent review: a Codex review pass was not available for this PR — the host's Codex configuration currently routes to a ChatGPT-backed provider that rejects the review model, and I do not modify that configuration. Given test-infrastructure-only scope, zero production LOC, a regression test that fails pre-fix for the stated reason, and green changed-file gates, I am landing on that basis and stating the gap rather than certifying a review that did not happen.

### DIFF
--- a/scripts/lib/vitest-local-scheduling.mts
+++ b/scripts/lib/vitest-local-scheduling.mts
@@ -70,7 +70,9 @@ export function detectVitestHostInfo() {
   };
 }
 
-let schedulingHostInfoCache: ReturnType<typeof detectVitestHostInfo> | undefined;
+// Vite bundles each project config on its own, so a module-level cache would be
+// per-project. The snapshot must live on globalThis to span every bundle in a process.
+const SCHEDULING_HOST_INFO = Symbol.for("openclaw.vitestSchedulingHostInfo");
 
 /**
  * Worker sizing reads the 1m load average, so re-detecting per Vitest project lets two
@@ -80,8 +82,11 @@ let schedulingHostInfoCache: ReturnType<typeof detectVitestHostInfo> | undefined
  * detectVitestHostInfo for the resource reporter.
  */
 function schedulingHostInfo(): ReturnType<typeof detectVitestHostInfo> {
-  schedulingHostInfoCache ??= detectVitestHostInfo();
-  return schedulingHostInfoCache;
+  const store = globalThis as Record<PropertyKey, unknown>;
+  if (!Object.hasOwn(store, SCHEDULING_HOST_INFO)) {
+    store[SCHEDULING_HOST_INFO] = detectVitestHostInfo();
+  }
+  return store[SCHEDULING_HOST_INFO] as ReturnType<typeof detectVitestHostInfo>;
 }
 
 function resolveMemoryPressureWorkerLimit(system: VitestHostInfo) {
--- a/scripts/lib/vitest-local-scheduling.mts
+++ b/scripts/lib/vitest-local-scheduling.mts
@@ -79,7 +79,7 @@ let schedulingHostInfoCache: ReturnType<typeof detectVitestHostInfo> | undefined
  * tests. Size every project in a process against one snapshot; live readings stay on
  * detectVitestHostInfo for the resource reporter.
  */
-export function schedulingHostInfo(): ReturnType<typeof detectVitestHostInfo> {
+function schedulingHostInfo(): ReturnType<typeof detectVitestHostInfo> {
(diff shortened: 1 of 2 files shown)


steipete:
The P2 was correct, and the probe that confirmed it is now the regression test. Two separately loaded module instances (`vi.resetModules()` between imports) resolved 12 then 3 workers on the module-level cache — per-bundle, exactly as you described. The snapshot now lives on `globalThis` under `Symbol.for("openclaw.vitestSchedulingHostInfo")`, matching the `scripts/lib` idiom in `vitest-profiler.mts` and `vitest-worker-bootstrap.mts`, and the test loads two distinct instances and guards that they are distinct before asserting agreement. Pre-fix: `expected 3 to be 12`, 1 failed / 21 passed. With the fix: 22 passed. deadcode:exports 0 entries; check-changed exit 0.

@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [b7a1898fad90c86c2a977cf6dbaa0af7cf2205d9](https://github.com/openclaw/openclaw/pull/141716/commits/b7a1898fad90c86c2a977cf6dbaa0af7cf2205d9)
- Landed commit: [3af7d292eed75b8af92f588838ac42e47d2a86e1](https://github.com/openclaw/openclaw/commit/3af7d292eed75b8af92f588838ac42e47d2a86e1)

PR #141719, opened 2026-09-08
fix(tooling): prevent accidental shared dependency installs

steipete:
## What Problem This Solves

Fixes an issue where developers running tools in a worktree without dependencies could later change another checkout's dependencies with a normal install. Tool bootstrap could silently attach that worktree to another checkout's `node_modules`.

## Why This Change Was Made

The TypeScript bootstrap and lint runners stop creating implicit dependency links. Explicit configured directories still fall back to an existing checkout installation, with pnpm's raw path values and environment precedence preserved. Existing readable borrows, explicit hydrated aliases, and relocated declaration output remain supported.

A dependency-free `pnpm:devPreinstall` check refuses borrowed checkout-root module directories and their `.pnpm` directories before reconciliation, with an exception for an exactly matching explicitly configured physical hydration target. The published file list includes the guard, and CI's audited dependency-cache contract records both its exact command and source bytes.

The guard does not inspect every workspace package, lock paths against concurrent replacement, or override script suppression and every alternate pnpm directory setting. No dependency, package-manager, runtime, or application configuration changes are included.

## User Impact

Missing dependencies produce actionable setup guidance. Ordinary installs through borrowed roots stop before changing their donor. Developers can preserve installations used by running tasks and prepare an owned installation separately. The maintainer accepts this transition and the documented admission-check limits for this repair.

## Evidence

- Real pnpm 12.3.4 default and frozen installs reproduced a synthetic donor changing from version 1.0.0 to 2.0.0. Five guarded borrowed variants reject with every donor byte, symlink target, and recorded filesystem identity unchanged; owned and explicitly matching hydrated installs succeed.
- Actual published-manifest preparation and npm file selection reproduced a retained hook whose script was omitted. The corrected archive includes the exact guard, and the extracted command succeeds for owned modules and refuses borrowed modules. This verifies file selection and command execution, not a full release build or installation.
- The original ownership/bootstrap regressions and environment-path corrections were reproduced before their fixes. The final original five-suite Linux run passed 127 tests, and the extended bootstrap/ownership run passed 98. Full original core/extension/script/test checks and full lint passed; later corrections passed appropriate scoped Linux checks. [Original Linux proof](https://github.com/openclaw/openclaw/actions/runs/34170559830).
- Initial PR CI exposed the omitted cache-contract entry, which was corrected and independently verified with fingerprint invalidation and 517 workflow-guard tests (two existing skips). ClawSweeper's stale primary-bootstrap success assertion was reproduced and corrected while preserving hydrated success and donor/link coverage.
- Subsequent CI exposed source-only Docker fixtures missing explicit TypeScript provisioning, outdated diagnostic/target-list expectations, and a planner test that equated incidental runner assignments across regrouping. Real release jobs already install their own harness. The fixtures now expose only the installed `tsx` package inside physical fixture-owned module directories. The planner test validates each group's actual compiler capacity while retaining negative mutation, coverage, timing, and execution-policy checks; production capacity policy is unchanged. All 604 tests in the five affected suites and the complete four-file changed gate passed in a fresh Linux environment with matching source hashes. [Final integration proof](https://github.com/openclaw/openclaw/actions/runs/34177408133).
- CI against newer main exposed test-file fragmentation into 81 jobs against the unchanged 80-job cap. The eight new bootstrap cases and fixture logic now live in the existing runtime-ownership suite (617 lines), with precise routing preserved. Independent AST/token comparison verifies every moved case; all three plans on the actual prospective merge fit 80 jobs. Integrated Linux validation passed 624 tests and complete scoped checks without changing capacity policy, estimates, or assertions. [Integrated consolidation proof](https://github.com/openclaw/openclaw/actions/runs/34180349881).
- Independent P0–P2 reviews are clean. Every task-owned Testbox lease is confirmed completed. The final command succeeded before an external portal-sync timeout; that warning is retained, and backing Actions lifecycle status is separate from delegated command proof.

A local dependency copy retained donor fallback text in generated launchers, and a local check encountered a missing browser-test dependency. Accepted comprehensive proof comes from independently installed Linux environments; the copy is not claimed to be fully independent. ClawSweeper could not retrieve an earlier proof run independently, which remains a reviewer access limitation. This prevents dependency incidents; it does not repair an already modified install or establish a fix for production request stalls.

Tooling/package: +69/−88 lines. Tests/support: +256/−33. Documentation: +23/−4. Existing shared installations and live services were not modified by this work.


steipete:
Merged via squash.

- Prepared head SHA: [7bf8adef13d97ee058b803158d60700b01fb6094](https://github.com/openclaw/openclaw/pull/141719/commits/7bf8adef13d97ee058b803158d60700b01fb6094)
- Landed commit: [62395281ff9c82f6ce121b248e7cd3fb09cf68bf](https://github.com/openclaw/openclaw/commit/62395281ff9c82f6ce121b248e7cd3fb09cf68bf)

PR #141721, opened 2026-09-08
fix(process): Windows claude-cli spawn picks the npm POSIX shim over claude.cmd

### LINKED ISSUE #141382: [Bug]: Windows — claude-cli backend fails with `write EPIPE` on every turn in 2026.9.2 (regression from 2026.9.1)
## Summary

On Windows, **every** `claude-cli` agent turn fails in 2026.9.2. The gateway boots fine and the port
listens, but no agent can ever reply.

The reported error is `write EPIPE`. That error is a **red herring** — it masks an asynchronous
`ENOENT` from `spawn()`. The actual defect is that OpenClaw's own executable resolver picks the
npm-generated **extensionless POSIX shim** (`claude`, a 308-byte `/bin/sh` script) instead of a real
Windows executable. `CreateProcess` cannot run that file.

This is a regression: 2026.9.1 delegated process launch to `@anthropic-ai/claude-agent-sdk` and was
unaffected.

## Environment

| | |
|---|---|
| OpenClaw | 2026.9.2 (3928bad) — broken; 2026.9.1 — works |
| OS | Windows 11 Home 10.0.26200 |
| Node | v24.19.0 |
| Claude Code CLI | 2.1.258 (installed via `npm -g`, unchanged before/after) |
| Gateway service | Windows Scheduled Task (`schtasks`) |
| Auth | Claude subscription OAuth (`loggedIn: true`), no API key |

## Symptom

User-facing (Telegram):

```
claude-cli/claude-fable-5-1 request failed (request timed out, HTTP 408)
```

Gateway log — identical for every model (`claude-fable-5-1`, `claude-sonnet-5`) and every trigger
(`user`, `cron`):

```
cli exec: provider=claude-cli model=claude-fable-5-1 promptChars=2460 trigger=user reuse=none
cli live session start: provider=claude-cli model=claude-fable-5-1 activeSessions=1
cli terminal failure: provider=claude-cli model=claude-fable-5-1 durationMs=3872 error=write EPIPE
lane task error: error="write EPIPE | EPIPE"
Embedded agent failed before reply: write EPIPE | EPIPE
```

Note `reuse=none` above — the `reason=auth-profile` / `invalidated:auth-profile` lines that often
accompany these failures are **not** causal; failures occur identically without them.

## Root cause

1. `executePluginOwnedProcess` resolves the command itself:
   `dist/execute.runtime-Dwf6akx9.js:1501`
   ```js
   const command = resolveExecutablePath(params.executionCommand, { cwd, env: params.env });
   ```

2. On Windows the resolver tries an **empty extension first**:
   `dist/executable-path-BK-aU8Pd.js:102`
   ```js
   // resolveWindowsExecutableExtensions(executable, env, includeExtensionless = true)
   return includeExtensionless ? ["", ...extensions] : extensions;
   ```
   and accepts it:
   ```js
   if (!ext || WINDOWS_NATIVE_EXECUTABLE_EXTENSIONS.has(ext)) return true;
   ```

3. A global npm install of `@anthropic-ai/claude-code` puts three shims in
   `%APPDATA%\npm` (which is on PATH):

   ```
   claude       <- 308-byte /bin/sh script   *** matched first (empty extension) ***
   claude.cmd
   claude.ps1
   ```

   The real binary is **not on PATH**; it lives at
   `%APPDATA%\npm\node_modules\@anthropic-ai\claude-code\bin\claude.exe`.

4. `spawn()` on the sh script emits `ENOENT` **asynchronously**. Meanwhile
   `createClaudeCliTransport` calls `initialize()` immediately after spawn and writes to
   `child.stdin`, which is already destroyed
   (`dist/extensions/anthropic/cli-transport-Do_poK4l.js`):
   ```js
   child.stdin.on("error", (error) => { fail(error); });
   // initialize() -> child.stdin.write(JSON.stringify(message) + "\n", cb)
   ```
   The `EPIPE` from that write wins the race and is surfaced as the terminal failure. **The real
   `ENOENT` is swallowed.**

## Reproduction (minimal)

```js
const { spawn } = require('node:child_process');
for (const p of ['claude',
                 String.raw`C:\Users\<you>\AppData\Roaming\npm\claude`,
                 String.raw`C:\Users\<you>\AppData\Roaming\npm\node_modules\@anthropic-ai\claude-code\bin\claude.exe`]) {
  const c = spawn(p, ['--version']);
  c.on('error', e => console.log(p, '-> ERROR', e.code));
  c.stdin.write('x\n', e => console.log(p, '-> write cb', e && e.code));
}
```

Observed:

| spawn target | result |
|---|---|
| `claude` (PATH) | `ENOENT` + `write EPIPE` |
| `...\npm\claude` (sh shim) | `ENOENT` + `write EPIPE` |
| `...\claude-code\bin\claude.exe` | works, exit 0 |

`write EPIPE` is byte-for-byte the gateway's reported error.

Verified separately that the 2026.9.2 transport protocol is fully compatible with claude 2.1.258:
spawning `claude.exe` with exactly the args `prepareClaudeCliTransportArgs()` builds and sending the
same `initialize` control_request returns `control_response {"subtype":"success"}` in ~905 ms and the
process stays alive. **There is no arg or version incompatibility — only the resolved path is wrong.**

## Why 2026.9.1 was unaffected

2026.9.1 used `dist/extensions/anthropic/agent-sdk.runtime.js`, passing
`pathToClaudeCodeExecutable: context.command` and letting `@anthropic-ai/claude-agent-sdk` own the
launch. 2026.9.2 removed that file and introduced `cli.runtime.js` + `cli-transport` + `cli-process`,
which spawn the resolved path directly.

## Suggested fixes

1. **Do not treat the extensionless candidate as executable on win32.** Either drop `""` from
   `resolveWindowsExecutableExtensions` on Windows, or verify the target is a real PE image (or at
   least reject files beginning with `#!`) before accepting it.
2. **Do not let `EPIPE` mask `ENOENT`.** Wait for the child's `spawn` event before the first
   `stdin.write()`, and prefer the `error` event over the stdin write error when both fire.
3. Consider preferring a package-local `bin/*.exe` when the resolved PATH entry is a shim.

## Secondary issue

The user-facing message translates this failure to `request timed out, HTTP 408`. Nothing timed out
and there is no HTTP 408 — this actively misleads diagnosis. Surfacing the underlying spawn error
would be far more useful.

## Workaround (confirmed working)

Put the directory holding the real executable ahead of the npm shim directory on PATH, then restart
the gateway (the resolver cache is process-local with a 60 s TTL, so a restart is required):

```
%APPDATA%\npm\node_modules\@anthropic-ai\claude-code\bin
```

After this, the same turn that previously failed returns normally, with no `EPIPE` and no
`auth-profile` invalidation.

ayaangazali:
## What Problem This Solves

Fixes #141382.

On Windows every `claude-cli` turn fails in 2026.9.2. The user sees a request timeout and the gateway logs `write EPIPE`, which is a red herring: it masks an asynchronous `ENOENT` from `spawn()`.

The real cause is PATH probe order. `resolveWindowsExecutableExtensions` returns the bare name before any PATHEXT candidate:

`src/infra/executable-path.ts`

```ts
const extensions = [...resolveWindowsExecutableExtSet(env)];
return includeExtensionless ? ["", ...extensions] : extensions;
```

and `resolveExecutableFromPathEnv` takes the first hit:

```ts
for (const entry of entries) {
  for (const ext of extensions) {
    const candidate = path.join(entry, executable + ext);
    if (isExecutableFile(candidate, { env })) {
      return candidate;
    }
```

A global npm install of `@anthropic-ai/claude-code` writes three files into a directory on PATH: `claude` (a 308 byte `/bin/sh` shim), `claude.cmd`, and `claude.ps1`. The bare shim matches first, and `CreateProcess` cannot execute it.

## Why This Change Was Made

The order is reversed so PATHEXT candidates are probed first and the extensionless entry becomes the last resort. On Windows an extensionless file is not directly executable, so it should never outrank a real `.cmd` or `.exe` of the same name.

Keeping the entry rather than dropping it preserves existing behaviour for callers that pass `includeExtensionless: true` and genuinely have only a bare file present; that case still resolves, it simply no longer wins when a PATHEXT sibling exists. Call sites that already pass `includeExtensionless: false` are untouched.

This is a one line ordering change. Nothing about acceptance, caching, or PATH parsing moves.

## User Impact

Windows installs where the CLI was installed through global npm can spawn it again, which is every `claude-cli` turn on that setup. The issue reports 2026.9.1 was unaffected because launching was delegated to the SDK.

## Evidence

Production change is +4/-1 in one file.

The regression test fails on unfixed code for the right reason. Restoring only `src/infra/executable-path.ts` from main:

```
× prefers a PATHEXT match over an extensionless shim of the same name
  → expected '…/claude' to be '…/claude.cmd'
  Tests  1 failed | 26 passed | 2 skipped (29)
```

It runs on any platform: the suite already has a `withMockedPlatform("win32", …)` helper, and under that mock `isExecutableFile` only checks the file and its extension, so no Windows host is needed.

```
node scripts/run-vitest.mjs src/infra/executable-path.test.ts        27 passed | 2 skipped
node scripts/run-vitest.mjs src/infra src/process src/tui     47 files, 688 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                                clean
./node_modules/.bin/oxfmt --check <both touched files>                         clean
```

One failure in that wider run, `btw-inline-message > sanitizes the question and successful Markdown result on every update`, reproduces identically with this file restored from main, so it is pre-existing and unrelated.

The existing `keeps extensionless lookup explicit with PATHEXT` cases still pass unchanged. They cover a bare file with no sibling, which this ordering does not affect; the new case is the ambiguous one they did not cover.

Proof gap worth stating: I have no Windows host, so this is verified through the mocked-platform path rather than by spawning the real CLI on Windows. The mock covers the resolver decision that the issue identifies, not the downstream `CreateProcess` behaviour.

### DIFF
--- a/src/infra/executable-path.test.ts
+++ b/src/infra/executable-path.test.ts
@@ -158,10 +158,10 @@ describe("executable path helpers", () => {
     },
   );
 
-  it("prefers a PATHEXT match over an extensionless shim of the same name", async () => {
+  it("spawn resolution prefers a PATHEXT match over an extensionless npm shim", async () => {
     // A global npm install drops a POSIX shim next to the real launcher, so `claude` and
     // `claude.cmd` share a directory on PATH. Windows cannot CreateProcess the bare shim,
-    // so probing it first made every spawn fail.
+    // so the spawn resolver probes PATHEXT first and keeps the bare name as a fallback.
     await withMockedPlatform("win32", async () => {
       await withTestDir({ prefix: "openclaw-exec-shim-" }, async (binDir) => {
         const shimPath = path.join(binDir, "claude");
@@ -170,8 +170,25 @@ describe("executable path helpers", () => {
         await fs.writeFile(commandPath, "@echo off\n");
 
         expect(
-          resolveExecutableFromPathEnv("claude", binDir, { PATHEXT: ".COM;.EXE;.BAT;.CMD" }),
+          resolveExecutablePath("claude", {
+            env: { PATH: binDir, PATHEXT: ".COM;.EXE;.BAT;.CMD" },
+          }),
         ).toBe(commandPath);
+      });
+    });
+  });
+
+  it("spawn resolution still finds a host that only ships a bare executable", async () => {
+    await withMockedPlatform("win32", async () => {
+      await withTestDir({ prefix: "openclaw-exec-bare-" }, async (binDir) => {
+        const barePath = path.join(binDir, "bare-host");
+        await fs.writeFile(barePath, "#!/bin/sh\nexit 0\n");
+
+        expect(
+          resolveExecutablePath("bare-host", {
+            env: { PATH: binDir, PATHEXT: ".COM;.EXE;.BAT;.CMD" },
+          }),
+        ).toBe(barePath);
       });
     });
   });
--- a/src/infra/executable-path.test.ts
+++ b/src/infra/executable-path.test.ts
@@ -158,6 +158,24 @@ describe("executable path helpers", () => {
     },
   );
 
+  it("prefers a PATHEXT match over an extensionless shim of the same name", async () => {
+    // A global npm install drops a POSIX shim next to the real launcher, so `claude` and
+    // `claude.cmd` share a directory on PATH. Windows cannot CreateProcess the bare shim,
+    // so probing it first made every spawn fail.
+    await withMockedPlatform("win32", async () => {
+      await withTestDir({ prefix: "openclaw-exec-shim-" }, async (binDir) => {
+        const shimPath = path.join(binDir, "claude");
+        const commandPath = path.join(binDir, "claude.cmd");
+        await fs.writeFile(shimPath, "#!/bin/sh\nexec node run.js\n");
+        await fs.writeFile(commandPath, "@echo off\n");
+
+        expect(
+          resolveExecutableFromPathEnv("claude", binDir, { PATHEXT: ".COM;.EXE;.BAT;.CMD" }),
+        ).toBe(commandPath);
+      });
+    });
+  });
+
   it("slides PATH hit and miss expiry for steady pollers", async () => {
     await withTestDir({ prefix: "openclaw-exec-path-" }, async (base) => {
       const binDir = path.join(base, "bin");
--- a/src/infra/executable-path.ts
+++ b/src/infra/executable-path.ts
@@ -46,7 +46,10 @@ function resolveWindowsExecutableExtensions(
     return [""];
   }
   const extensions = [...resolveWindowsExecutableExtSet(env)];
-  return includeExtensionless ? ["", ...extensions] : extensions;
+  // Windows cannot CreateProcess an extensionless file, and a global npm install drops a
+  // POSIX shim beside the real launcher (`claude` next to `claude.cmd`). Probing the bare
+  // name first matched that shim and the spawn failed, so PATHEXT candidates come first.
+  return includeExtensionless ? [...extensions, ""] : extensions;
 }
 
 function resolveWindowsExecutableExtSet(env: NodeJS.ProcessEnv | undefined): Set<string> {
--- a/src/infra/executable-path.ts
+++ b/src/infra/executable-path.ts
@@ -46,10 +46,7 @@ function resolveWindowsExecutableExtensions(
     return [""];
   }
   const extensions = [...resolveWindowsExecutableExtSet(env)];
-  // Windows cannot CreateProcess an extensionless file, and a global npm install drops a
-  // POSIX shim beside the real launcher (`claude` next to `claude.cmd`). Probing the bare
-  // name first matched that shim and the spawn failed, so PATHEXT candidates come first.
-  return includeExtensionless ? [...extensions, ""] : extensions;
+  return includeExtensionless ? ["", ...extensions] : extensions;
 }
 
 function resolveWindowsExecutableExtSet(env: NodeJS.ProcessEnv | undefined): Set<string> {
@@ -212,10 +209,19 @@ export function resolveExecutablePath(
     resolveEnvironmentValue(options?.env, "PATH") ??
     resolveEnvironmentValue(process.env, "PATH") ??
     "";
-  return resolveExecutableFromPathEnv(candidate, envPath, options?.env, {
-    cwd: options?.cwd,
-    useCache: options?.useCache,
-  });
+  // This resolves a command that is about to be spawned. Windows cannot CreateProcess an
+  // extensionless file, and a global npm install drops a POSIX shim beside its launcher
+  // (`claude` next to `claude.cmd`), so probe PATHEXT first and keep the bare name as the
+  // fallback for hosts that genuinely ship one. resolveNodeHostExecutable does the same.
+  // Both probes carry the caller's cwd and cache choice, and includeExtensionless is part
+  // of the cache key, so the two phases cannot read each other's entry.
+  const probe = { cwd: options?.cwd, useCache: options?.useCache };
+  return (
+    resolveExecutableFromPathEnv(candidate, envPath, options?.env, {
+      ...probe,
+      includeExtensionless: false,
+    }) ?? resolveExecutableFromPathEnv(candidate, envPath, options?.env, probe)
+  );
 }
 
 /**


ayaangazali:
Windows CI caught a real problem with my first attempt, and it was mine. Pushed a corrected approach rather than arguing with it.

The failure was `resolveNodeHostExecutable > preserves an explicit extensionless Windows override`. My change reordered PATHEXT ahead of the bare name inside `resolveWindowsExecutableExtensions`, which quietly changed what `includeExtensionless: true` means. That flag is a contract: a caller passing it explicitly wants the bare file to win, and `node-host.ts` depends on that.

Worse, I had missed that `resolveNodeHostExecutable` already solves this exact npm shim problem, and documents it:

```ts
// npm installs a non-runnable bare shim beside its .cmd launcher. Search every
// PATH source for PATHEXT launchers before retaining bare-only native hosts.
return resolve(false) ?? resolve(true);
```

So the correct shape already existed one layer up; I had reached for a global reorder instead of reusing it.

The new commit leaves `resolveWindowsExecutableExtensions` byte-identical to main and applies that same two-phase lookup where a command is actually about to be spawned. `resolveExecutablePath` now probes with `includeExtensionless: false` first and falls back to the bare name, so:

- `claude` beside `claude.cmd` resolves to `claude.cmd`, which is the reported failure.
- a host that genuinely ships only a bare executable still resolves, which the second new test covers and which passes both before and after the change.
- `includeExtensionless` keeps its meaning for every caller that sets it, so the node-host override behaves exactly as before.

Local runs: `executable-path` 28 passed, `node-host` and `src/process` green, core typecheck and format clean. The node-host Windows cases are `runIf(win32)` so they skip on my machine; I am relying on this CI lane for them, which is the same gap I flagged in the PR body and which is exactly what caught the first attempt.

ayaangazali:
Accepting the P1 finding rather than arguing it. You are right that this is an incomplete candidate fix.

Picking `claude.cmd` over the bare shim fixes which file gets selected, but it does not make that file launchable by this consumer. The repo already draws that distinction: `src/process/windows-command.ts` resolves with `includeExtensionless: false`, restricts to `.exe/.com/.cmd/.bat`, and otherwise throws "use an explicit executable or shell wrapper", which is exactly the boundary my change does not cross. `executePluginOwnedProcess` calls `resolveExecutablePath` directly and never goes through that support.

So the honest description of this PR is: it corrects the lookup order, and it leaves the launch boundary unchanged. On the reported install that likely turns one failure into a different one rather than into a working turn.

Where that leaves me. The better solution you describe, preparing a runnable executable or Node entrypoint at the launch boundary through the existing wrapper support, is a change I can write but cannot verify: I have no Windows host, and the earlier CI failure on this PR showed that reasoning about Windows launch semantics without one is how I got the first attempt wrong. Shipping a second unverified guess at a P0 launch path is worse than shipping nothing.

Two options, and I am fine with either:

1. Close this. The lookup-order detail and the pointer to `windows-command.ts` as the right layer are recorded here and on #141382 for whoever has a Windows environment.
2. If a narrowed version is still useful, say so and I will scope it to exactly that, with the incompleteness stated in the body rather than implied.

I would rather it close than sit as a partial fix on a release blocker. Not going to claim Windows proof I cannot obtain.

ayaangazali:
This had gone into merge conflict with main. Rebased and resolved, and the resolution changed the diff slightly, so here is exactly what happened.

#141987 added `cwd` and `useCache` to the same call this PR rewrites:

```
<<<<<<< HEAD
  return resolveExecutableFromPathEnv(candidate, envPath, options?.env, {
    cwd: options?.cwd,
    useCache: options?.useCache,
  });
=======
  return (
    resolveExecutableFromPathEnv(candidate, envPath, options?.env, {
      includeExtensionless: false,
    }) ?? resolveExecutableFromPathEnv(candidate, envPath, options?.env)
  );
>>>>>>>
```

Taking either side alone loses something. Mine would have dropped the new `cwd` and `useCache` threading; theirs would have dropped the two-phase probe. Both probes now carry the caller's options:

```ts
const probe = { cwd: options?.cwd, useCache: options?.useCache };
return (
  resolveExecutableFromPathEnv(candidate, envPath, options?.env, {
    ...probe,
    includeExtensionless: false,
  }) ?? resolveExecutableFromPathEnv(candidate, envPath, options?.env, probe)
);
```

The new cache raised a question I went and checked rather than assumed, because a two-phase probe against a shared cache is exactly where this would break. `executablePathCacheKey` includes `options?.includeExtensionless` (`src/infra/executable-path.ts:147-153`), so the PATHEXT-only probe and the fallback probe occupy different keys and cannot read each other's entry. If that argument were not part of the key, the first phase's miss would have been served to the second and the fallback would never have worked.

Production diff is +13/-4 in one file, up from +9/-1, and the whole increase is the option threading plus the comment. `resolveWindowsExecutableExtensions` is still untouched, which was the point of the correction I pushed on 2026-09-08 after Windows CI caught my first attempt.

`oxfmt --check` is clean on both touched files.

Proof gap, stated plainly: I did not re-run tests at this head. This machine went over its memory ceiling partway through the rebase work and I stopped rather than push it further, so the local evidence here is the conflict resolution and the cache-key check, not a test run. The Windows behavior was never provable on this box anyway, so CI has always been the authority for this PR; I will fix whatever it reports.

The accepted P1 from the review is unchanged and still open: picking `claude.cmd` over the bare shim fixes which file gets selected, and I said at the time that it is a candidate fix rather than a complete one. Still waiting on your call about whether that is the shape you want, or whether the launcher selection belongs somewhere else entirely.

ayaangazali:
Confirmed at the source, and you are right that this PR alone does not fix the reported turn.

`extensions/anthropic/cli-process.ts:32`:

```ts
const child = spawn(options.command, options.args, {
  argv0: options.argv0,
  cwd: options.cwd,
  detached: process.platform !== "win32",
  env: options.env,
  signal: options.signal,
  stdio,
  windowsHide: true,
});
```

No `shell`, no batch handling. So once my first probe returns `claude.cmd`, that path goes straight into a `spawn` that cannot execute a Windows batch file, and the turn fails exactly as reported. Selecting the right file and being able to launch it are two different problems, and I only fixed the first.

This is the third time on this PR that the real fix has turned out to sit a layer above the symptom, so I want to stop guessing at the scope and ask rather than push a fourth revision.

The lookup correction in `resolveExecutablePath` is independently right: `resolveNodeHostExecutable` already probes PATHEXT before the extensionless name, and this makes the spawn path agree with it. What it cannot do is make the selected `.cmd` launchable, because that decision belongs to whoever hands the command to `spawn`.

Two ways to finish it, and the choice is yours because it crosses an ownership line:

1. **Extend this PR into the plugin.** Prepare the invocation in `extensions/anthropic/cli-process.ts` before spawning, so a resolved `.cmd` is launched through the Windows entrypoint path rather than directly. Keeps the whole user-visible repair in one PR, but puts core-infra and plugin changes in the same diff.
2. **Keep this PR as the bounded lookup correction** and track the launch preparation separately against the anthropic plugin, since `cli-process.ts` is its owner and the same gap would affect any batch-resolved CLI it spawns, not just `claude.cmd`.

I lean toward 2, because the launch defect is not specific to this resolver and fixing it inside the plugin protects every command it spawns. But option 1 lands the reported symptom in one piece, and if that is what you want I will do that instead. Say which and I will build it.

On proof, my position has not changed and I would rather repeat it than let it look unaddressed: I have no Windows host, so I cannot produce the after-fix terminal output you are asking for. Everything I have shown is mocked-platform resolver selection, which is what this box can honestly demonstrate. If neither option can be verified without that proof, that is a fair reason to leave this parked until someone with Windows can pick it up, and I would rather hear that than keep revising blind.

PR #141725, opened 2026-09-08
fix: report silent fallback outcomes accurately

### LINKED ISSUE #141694: [Bug]: Silent-fallback reply hardcodes "I couldn't reach the configured model backend" regardless of failure reason, naming a provider that returned HTTP 200 to every request
### Bug type

### Beta release blocker

No

### Summary

`buildSilentFallbackFailurePayload` emits a fixed user-facing string asserting the configured model backend was unreachable whenever a fallback model ran and the final reply was empty — the failure reason is never consulted, so a provider that answered HTTP 200 to every request in the window is named to the user as unreachable.

### Steps to reproduce

Grounded in source inspection of the shipped dist plus one observed occurrence. Line references are to the published `openclaw@2026.9.1` build; filenames are content-hashed so they will differ in other builds.

1. Configure a primary model and at least one fallback.
2. Cause the primary to fail with a non-transport reason — any reason that `resolveFailoverStatus` maps to a 4xx, e.g. `format` (400). In the observed case the provider returned HTTP 200 and the run was classified `incomplete turn detected` with `candidate_failed reason=format`.
3. Let the fallback candidates also produce no visible reply.
4. Read the reply delivered to the channel.

The emitting path is unconditional on reason, so any reason class reproduces it — see Actual behavior.

### Expected behavior

Wording that reflects the observed failure class, with "couldn't reach" reserved for transport-class reasons.

`resolveFailoverStatus` (`dist/failover-error-B46Ke6VT.js:110-126`) already partitions the vocabulary:

| Maps to 5xx / timeout — genuinely unreachable | Maps to 4xx — provider responded |
|---|---|
| `server_error` 500, `overloaded` 503, `timeout` 408, `tls_certificate` 502 | `format` 400, `context_overflow` 413, `model_not_found` 404, `auth` 401, `auth_permanent` 403, `billing` 402, `rate_limit` 429, `session_expired` 410 |

Reason-aware user-facing wording is established behavior elsewhere in the same module — `isPureOverloadSummary`, `isBilling` and `isPureTransientSummary` all branch on `attempt.reason` at `dist/agent-runner.runtime-CAdsMsQt.js:990`, `:993` and `:1030`. This one path is the outlier.

### Actual behavior

A single hardcoded string, at `dist/agent-runner.runtime-CAdsMsQt.js:379-384`:

```js
function buildSilentFallbackFailurePayload(params) {
  if (params.isHeartbeat || params.allowEmptyAssistantReplyAsSilent === true
      || params.silentExpected === true || params.hasSuccessfulTerminalDelivery
      || !params.fallbackTransition.fallbackActive || !params.fallbackFailureKnown) return;
  return markReplyPayloadForSourceSuppressionDelivery({
    text: `⚠️ I couldn't reach the configured model backend ${params.fallbackTransition.selectedModelRef}. Fallback used ${params.fallbackTransition.activeModelRef}, but it produced no visible reply.`,
    isError: true });
}
```

Six guard conditions, and the only two that concern the models test neither reachability nor error class:

- `fallbackActive` (`:242`) is a ref comparison — `!areRuntimeModelRefsEquivalent(selectedModelRef, activeModelRef)`. It means "a different model ended up running."
- `fallbackFailureKnown` (`:5558`) is `fallbackAttempts.length > 0 || configuredFallbackModel.persistedAutoFallback`. It means "at least one attempt was recorded."

So the first clause is emitted on the condition *a different model ran and the final reply was empty*, and asserts unreachability of a specific named backend on that basis alone.

The reason data is in scope but discarded one frame before the message is built. `fallbackAttempts` is destructured at `:5533` and reduced to the boolean above at `:5558`; the call site at `:5658-5666` then passes only `fallbackTransition`, `fallbackFailureKnown`, `isHeartbeat`, `hasSuccessfulTerminalDelivery`, `allowEmptyAssistantReplyAsSilent` and `silentExpected`. The array of per-attempt reasons never reaches the builder.

The second clause ("Fallback used *X*, but it produced no visible reply") is accurate in the observed case. The defect is confined to the first.

### OpenClaw version

2026.9.1

### Operating system

macOS 26.5.2

### Install method

npm

### Model

Primary `openrouter/z-ai/glm-5.3`; fallbacks `amazon-bedrock/us.anthropic.claude-sonnet-5` then `amazon-bedrock/us.anthropic.claude-haiku-4-5`

### Provider / routing chain

`openclaw -> openrouter -> z-ai`, with fallbacks `openclaw -> amazon-bedrock`

Two agents, separate gateway processes, one host, identical routing configuration. No proxy or AI gateway in front of either provider.

### Logs

```shell
# Aggregates over one window; two agents on one host, openclaw 2026.9.1.
# Window: 2026-09-07 13:29-14:19 PT

Provider HTTP results (openrouter):
  agent A   72 requests   72 status=200   0 non-200
  agent B   40 requests   40 status=200   0 non-200

Run outcomes, same window:
  "incomplete turn detected"   agent A: 1   agent B: 3
  candidate_failed             every attempt reason=format

One agent-B turn walked the whole chain within a single run id:
  openrouter/z-ai/glm-5.3               -> incomplete turn, stopReason=stop, reason=format
  amazon-bedrock/...claude-sonnet-5     -> incomplete turn, stopReason=stop, reason=format
  amazon-bedrock/...claude-haiku-4-5    -> incomplete turn, stopReason=stop, reason=format

Reply delivered to the channel named only the first candidate, as unreachable.
```

### Impact and severity

- **Affected:** any deployment with a configured fallback chain where a run exhausts it on a non-transport reason. Not provider-specific — both Bedrock candidates in the observed run failed identically to the OpenRouter primary.
- **Severity:** low as a functional matter (the turn had already failed), meaningful as a diagnostic one. The message is the only signal most operators see, it is confidently worded, and it names one component.
- **Frequency:** deterministic on that path — the guard has no reason-dependent branch. Observed 4 times in one hour across two agents.
- **Consequence:** misdirected investigation. The message named the single component that was demonstrably healthy — 112 of 112 requests answered 200 — and did so while two candidates from a different provider had failed the same way. Acting on it means auditing provider credentials, credit balance and connectivity for a provider with no fault.

### Additional information

Not a regression as far as observed — no prior version was tested for this path, so no last-known-good is claimed.

The narrowest fix is to pass `fallbackAttempts` through to `buildSilentFallbackFailurePayload` and select wording by reason class, reusing the partition `resolveFailoverStatus` already encodes: keep the current sentence when every attempt reason maps to 5xx/timeout, and use something like "the configured model backend `X` responded but produced no usable reply" otherwise. The helper shape already exists in the same module (`isPureOverloadSummary` at `:990` is the same "every attempt has reason R" reduction).

`replyOperation.fail(...)` at `:5668` carries a parallel string, but "failed" there is defensible — the run did fail. Only the user-facing "couldn't reach" asserts a cause.

leilei3167:
Closes #141694

## What Problem This Solves

When fallback produced no visible reply, OpenClaw could claim the selected backend was unreachable even after a response. A local cooldown skip could also reuse an old format or timeout reason and imply that a new request occurred.

The existing warning now uses complete evidence for the selected backend through the existing runtime reference comparator. Local skips carry a structured marker in the existing error-code field and keep cause-neutral wording. Missing, mixed, unknown, authentication, billing, and rate-limit evidence also stays neutral.

Finalization retains an actual buffered `NO_REPLY` marker so the existing silence checks can recognize it. This removes the erroneous missing-output retry and error on the documented silent path. Fallback order, retry rules, cooldown policy, silence permissions, and channel delivery owners remain unchanged.

## Evidence

- Matched main/candidate Telegram Test Server turns reproduce the false unreachable warning and verify the corrected response wording, with two provider requests and one warning in each case.
- A local-cooldown control skips the selected backend without a primary request. The earlier candidate falsely claimed a response; the corrected candidate delivers one cause-neutral warning.
- 43 warning tests and 26 cooldown tests pass. The added local-skip cases fail on the prior code for the intended reason.
- 34 selected fallback integration tests, 21 subscription tests, two CLI empty-output producer tests, and targeted lint pass.
- Real channel controls cover supported causes, backend attribution, visible replies, duplicate updates, failed-send recovery, terminal message delivery, heartbeat suppression, and allowed group silence.
- Matched standalone group-silence proof fails on main and passes after the finalization repair. Fresh independent direct-message and group controls each finish after one fallback response with no assistant message.
- Fresh independent acceptance also verifies 22 negative, attribution, and suppression boundary controls. All fresh live runs completed owned-message cleanup.

| Selected-backend evidence | Before | After |
| --- | --- | --- |
| Format failure | “I couldn't reach…” | “responded but produced no usable reply” |
| Local cooldown retaining a format reason | Claims a backend response | Cause-neutral warning |

The original contributor's commits and credit are preserved.


obviyus:
Thanks @leilei3167!

**Proof:** Real Gateway turns on Telegram's Test Server verify response-supported warnings, neutral local-cooldown warnings, and quiet standalone replies in direct and group chats. All fresh runs completed owned-message cleanup. Focused warning/cooldown tests and exact-head CI passed.

- **Bug:** Empty fallback output could invent a backend cause, and buffered silent replies could trigger an unnecessary retry and error.
- **Fix:** The existing owners retain actual failure and silence evidence while preserving retry and silence rules.

| Before | After |
| --- | --- |
| Format failure reported as unreachable | Response-supported warning |
| Local skip treated as a backend response | Neutral warning |
| Buffered `NO_REPLY` retried and failed | Quiet completion |

PR #141727, opened 2026-09-08
fix(gateway): update model availability after expiry or cooldown

obviyus:
## What Problem This Solves

Chat model metadata could show stale availability after a token expired or a credential cooldown ended.

## Why This Change Was Made

Bind cached authentication projections to their token and failure-window deadlines so the existing reader rebuilds expired state. Existing config and prepared-state ownership checks remain.

## User Impact

The next metadata read reflects expiry or recovery without another login, credential edit, or restart. No timer or public field is added.

## Evidence

Matched real CLI/Gateway runs with actual clock deadlines reproduced available-after-expiry and unavailable-after-recovery states. The fix matched fresh model-list controls in both cases, and all four instances stopped cleanly. Focused deadline and config checks, formatting, and lint passed. Tests used synthetic credentials and did not cover inference or graphical presentation.

### DIFF
--- a/src/gateway/server-methods/chat-metadata-runtime.auth-deadline.test.ts
+++ b/src/gateway/server-methods/chat-metadata-runtime.auth-deadline.test.ts
@@ -1,0 +1,116 @@
+import { describe, expect, test, vi } from "vitest";
+import type { AuthProfileStore } from "../../agents/auth-profiles.js";
+import type { OpenClawConfig } from "../../config/types.openclaw.js";
+import {
+  createChatMetadataHarness,
+  createChatMetadataOwner,
+} from "./chat-metadata-runtime.test-support.js";
+
+describe("gateway chat metadata auth deadlines", () => {
+  test.each([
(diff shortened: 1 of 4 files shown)


obviyus:
Merged after clean exact-head review and required CI.

**Proof:** Matched source-mode CLI/Gateway runs used a retained authenticated connection, synthetic credentials, and actual token/cooldown deadlines. Independent acceptance passed all five clauses, including serving-process and connection shutdown.

- **Bug:** Repeated chat metadata reads retained stale availability after credential deadlines.
- **Fix:** Expired host-auth projections are rebuilt through the existing reader, with config and owner checks preserved.

| Case | Before | After |
| --- | --- | --- |
| Token expires | Model still available | Model unavailable |
| Cooldown ends | Model still unavailable | Model available |

[Required CI](https://github.com/openclaw/openclaw/actions/runs/34175287672) passed. Earlier setup failures and proof limits remain retained; these observations do not claim graphical UI behavior or vendor credential validity.

PR #141737, opened 2026-09-08
refactor(approvals): infer lazy native runtimes

RomneyDa:
## What Problem This Solves

Native approval adapters duplicated casts around lazily loaded channel runtimes. The shared adapter contract erased each loader's inferred payload type, forcing every channel to restate types that the loader already proves.

## Why This Change Was Made

The shared approval adapter now preserves all five loader-inferred generic types for ordinary SDK callers. Bundled channels opt into one explicit `capabilityBoundary` erasure only where their typed runtime enters the non-generic `ChannelApprovalCapability` contract.

This removes nine duplicated double casts without changing lazy loading, required/optional hook behavior, or startup imports. Observation remains intentionally limited to already-loaded runtimes.

## User Impact

No user-visible behavior change. Third-party typed SDK consumers retain the shipped generic return contract, while bundled native approval handling remains lazy and channel-owned.

## Evidence

- Exact-head `pnpm check:changed` passed at `e9fce8a2975`, including core and extension production/test typechecks, boundary guards, lint, dead-export scans, and cycle checks.
- All 14 focused approval runtime tests passed.
- A typed consumer assertion verifies that a loader returning `{ text: string }` preserves that payload type when `capabilityBoundary` is omitted.
- Fresh exact-head P0–P2 autoreview was scoped-clean with 0.93 correctness confidence.
- Diff: 70 additions, 79 deletions; net −9 LOC.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -310,7 +310,6 @@ extensions/discord/src/activities/register.ts	1
 extensions/discord/src/activities/store.ts	1
 extensions/discord/src/api.ts	3
 extensions/discord/src/approval-handler.runtime.ts	4
-extensions/discord/src/approval-native.ts	1
 extensions/discord/src/audit-core.ts	2
 extensions/discord/src/channel-actions.ts	1
 extensions/discord/src/channel.ts	3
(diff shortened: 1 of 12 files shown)


RomneyDa:
@clawsweeper re-review

Issue #137813, opened 2026-09-04
[Bug]: Windows gateway never starts after 2026.9.1 update — new `--task-supervisor` flag exits 0 silently; child never spawns


akagifreeez:
## Environment

- OS: Windows 11 Home (10.0.26200)
- OpenClaw version: 2026.8.2 → 2026.9.1 (`ad6fe23`), npm global install
- Node: v24.20.0
- Service type: Scheduled Task (`OpenClaw Gateway`, gateway.vbs → gateway.cmd)

## Problem

The 2026.9.1 update regenerates `~/.openclaw/gateway.cmd` and adds a new `--task-supervisor` flag to the node launch line:

```bat
"...node.exe" --max-old-space-size=8192 ...\openclaw\dist\index.js gateway --port 18789 --task-supervisor < NUL
```

With that flag, the process exits **immediately with code 0** and the gateway never binds its port. There is no output anywhere:

- stdout is discarded (`< NUL`)
- `%TEMP%\openclaw\openclaw-YYYY-MM-DD.log` receives **zero** entries from the run
- the Scheduled Task's last-run result is 0, so Task Scheduler treats it as success and `RestartOnFailure` never fires

The updater itself notices (`Gateway did not become healthy after restart`), but the gateway stays down — a watchdog or manual `schtasks /run` just repeats the same silent exit. This is the same observability gap as #119882, but with a different trigger and a 100% repro.

## Steps to Reproduce

1. Windows install, gateway registered as Scheduled Task, healthy on 2026.8.2.
2. Update to 2026.9.1 (update regenerates `gateway.cmd` and `gateway.vbs`, adding `--task-supervisor`).
3. Gateway task runs → process exits instantly (code 0), port never binds.

## Isolation

All on the same machine:

1. **Flag present** (via Scheduled Task, via gateway.vbs, and by running the exact node command manually in a foreground shell): instant `exit 0`, zero log lines, port stays free.
2. **Flag removed, same command otherwise**: gateway starts normally (HTTP server, health monitor, channels, tailscale serve all come up).
3. **Fresh state dir** (`OPENCLAW_STATE_DIR` pointed at an empty scratch, port 18790): same contrast — with the flag, instant silent exit 0; without it, boots fine with `--allow-unconfigured`. So it is not state/config dependent.
[...]

PR #136980, opened 2026-09-03
fix(windows): keep managed Gateway startup alive and stoppable

steipete:
Related: #136203 — Windows upgrade maintenance, workspace and stale-Codex reports.

<details>
<summary>Additional instructions</summary>

Maintainers can update this same-repository branch.

</details>

## What Problem This Solves

Resolves a problem where starting a managed Gateway on native Windows could exit before the Gateway started, and stopping it immediately could report success while its supervisor later launched a Gateway. These are reproduced current-main startup/lifecycle defects discovered while investigating the Windows upgrade report.

## Why This Change Was Made

Keep the service relay's real child handle referenced through admission and lineage completion, removing the premature `unref()` and redundant retention map. Pass `NULL` to `CreateProcessW` when the command environment is omitted, preserving native inheritance; explicitly supplied environments, including `{}`, remain replacements. For pre-admission stop, recognize the exact installed supervisor argv while preferring the actual Gateway PID and preserving the full-argv, port and identity fences.

The native fixture now proves deliberate exit 23 and extinction of that exact child and supervisor, requests recovery explicitly, and reads current numeric Running state rather than treating historical LastTaskResult as liveness. It retains the complete service lifecycle and ownership assertions. Its external temporary cwd receives test-owned dependency resolution through the public tsx export and checkout tsconfig. No production retry or keepalive was added. Workspace migration instructions now identify the intended per-agent workspace and verify it without rerunning onboarding.

## User Impact

Native Windows operators can start, restart and stop the managed Gateway, including stopping before its child is admitted. Omitted child environments inherit normally; intentional sanitized environments and explicit Codex overrides retain their meaning. Production is **+17/-17 (net 0)**; tests/support **+232/-121 (net +111)**; docs **+3/-3**. There are no configuration, dependency, SQLite-schema or protocol changes.

Thanks to @hermannmetzker-ai and @DasX for the reports that prompted the investigation.

## Evidence

**266 distinct local passes and 63 distinct native Windows passes**; repeated subsets are not added. Local coverage includes POSIX cancellation, parent loss and descendant extinction. Final changed checks and formatting passed. A fresh, secret-scanned, isolated Codex review before commit found no actionable P0 findings; this is a P0-scoped result.

[...]

PR #141744, opened 2026-09-08
fix(windows): stop Gateway task descendants and log child failures

steipete:
Related: #137813 and #136980.

## What Problem This Solves

Fixes two Windows Scheduled Task failures: a supervised Gateway can fail without recording its child exit result or stderr, and ending the task can leave the Gateway alive and owning its state directory. The 2026.9.1 premature-host-exit fix from #136980 is already in 2026.9.2; this PR adds the missing diagnostics and repairs task teardown.

## Why This Change Was Made

The supervisor records the child exit code, signal, reason, and the last 8,192 stderr characters in the existing redacted Gateway log. It preserves failure results and intentional zero exits, flushes logging, and continues to discard stdout.

Native inspection found that the Gateway already belonged to a kill-on-close Job, but the surviving anchor process held its handle. Task Scheduler ended WScript, leaving the Node supervisor and anchor alive. The supervisor now validates and pins its original CMD/WScript ancestry, transfers the last non-inheritable handle of an outer kill-on-close Job to WScript, assigns itself, and closes its local copy before admitting a child. Ending WScript therefore takes down the supervisor and its descendants. The existing inner Job still owns cancellation and extinction joins.

The generated VBS uses the existing hidden-launcher setting as a consumed runtime marker. The generated CMD does not overwrite it with the install preference. Direct Startup launches retain their existing behavior. This changes the generated launcher contract; it adds no dependency, configuration option or log store. This PR makes no config-schema or SQLite/storage-format change. Existing installations receive the new launcher through service refresh or `openclaw gateway install --force`.

## User Impact

Operators can diagnose early child failures and end the Scheduled Task without leaving a Gateway behind. Both `schtasks /end` and `Stop-ScheduledTask` are covered directly, before any CLI cleanup, and a subsequent task start must create a new healthy process.

The Windows docs also clarify a platform limit: Task Scheduler's `RestartOnFailure` policy covers failed start conditions or action launches. A nonzero exit from an already launched Gateway does not guarantee automatic recovery; fix the logged cause and start it again.

## Evidence

Validation is native Windows Server 2022 x64 with Node 24.20.0, isolated task-owned accounts/state and loopback ports. Windows 11 and ARM64 were not directly exercised.

- Released 2026.9.2 and earlier candidate `9386638`: both raw stop methods kill the real Scheduler WScript action while leaving the Node supervisor, anchor and Gateway alive. Event 129 and EnginePID establish Scheduler origin. Specific `IsProcessInJob` queries prove Gateway membership in the anchor-held Job with `KILL_ON_JOB_CLOSE=0x2000`; the diagnostic inspector closes its duplicates before stopping the task.
- Exact final head `19e1f718aabb3807f4dd7d753b64c552bb683591`: the native integration suite passed all 5 tests. Raw `/End` extinguished all 10 recorded processes; `Stop-ScheduledTask` extinguished all 9 recorded processes. Each freed the port and allowed a fresh start. Restart, intentional zero-exit shutdown, and the 65-second no-restart observation passed. Generic direct and batch Startup fallback probes survived their launcher's exit (these native probes do not launch a real Gateway), and the default task definition remained unchanged. The Gateway-specific unit case also verifies that the install preference `1` does not require a WScript owner.
- The installed final tarball (SHA256 `165b20e8dd709ae04783f1775e6df53da90846a95b73156ffc5eb887238178de`) also passed both stop and fresh-start checks. Raw `/End`: WScript 2192 held non-inheritable Job handle 1516 with flags `0x2000`; Gateway 6156 was a confirmed member. All descendants disappeared, and the next `/Run` started Gateway 2624 with HTTP 200. `Stop-ScheduledTask`: WScript 4520 held handle 944 with the same flags; Gateway 6688 disappeared and the next `/Run` started Gateway 6484 with HTTP 200. Both left no listener or recorded survivor. The probe's aggregate status was nonzero because short-lived PowerShell/conhost startup helpers exited before handle enumeration (Win32 error 87); inspection of the WScript/supervisor/anchor/Gateway and the stop/restart assertions succeeded. These were normal generated launchers, without the diagnostic preload.
- Occupied-port retry comparison: the registered policy was `PT1M`, count 3. Demand-start, actual timer-triggered start, and explicit `wscript.exe` action each produced result `1` and the logged child failure, but no automatic second execution over 270.8–271.5 seconds after failure. This agrees with the [Task Scheduler protocol specification](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tsch/2ff4aa5a-7bc4-449f-bbb1-27475645867f). The timer experiment restored the original definition exactly. The final installed package independently returned task result `1` and logged `Gateway child failed` with child exit code `1` and an `EADDRINUSE` stderr tail; after 270.58 seconds it still had exactly one Scheduler-created action instance and no automatic retry.
- Positive retry control on the installed final package: changing only the Exec working directory to a nonexistent path caused Scheduler Event 203 / `ERROR_DIRECTORY` (`0x8007010B`) before WScript launched. Creating that directory triggered a new Scheduler-created WScript (PID 2180) automatically after 60.997 seconds, without a second user start; Gateway 5256 returned HTTP 200. The probe completed its 271.64-second window with exit `0`, stopped the task, and restored the original task XML exactly. This proves that the configured retry policy is active for an action-launch failure, while the occupied-port comparison proves that application exit `1` does not trigger it on this host. The initial live evidence-copy attempt encountered a Windows file-sharing conflict; the control was repeated without live output-file reads. The successful native probe outlived a Crabbox heartbeat failure; its completed JSON and cleanup receipt were retrieved afterward.
- Focused local tests: 183 passed, 1 platform-specific skip. The two new admission cases were first demonstrated failing against the original owner. `pnpm check:changed`, `git diff --check`, package build and tarball integrity passed. Independent P0–P2 review found no actionable code issue; the retry documentation correction also passed independent review.
- Exact-head CI is [green](https://github.com/openclaw/openclaw/actions/runs/34181787590).

Earlier issue proof remains relevant: 2026.9.1 reproduced silent task result `0`, no listener, and no added log lines. Shell `openclaw update --tag 2026.9.2 --yes` installed 9.2 but the old updater's restart step failed loading a removed dist module; a fresh CLI start recovered. The previous candidate's shell update completed and reached HTTP 200. No manual lock removal or stale-lock recovery was used. Issue #137813 remains open for maintainer review.

Thanks to @akagifreeez, @Devin-Linux, and @unknowbug for the native diagnosis in #137813.

### DIFF
--- a/docs/platforms/windows.md
+++ b/docs/platforms/windows.md
@@ -137,6 +137,12 @@ through a generated `gateway.vbs` WScript wrapper, so the background Gateway
 does not open a visible console window. If task creation is denied, OpenClaw
 falls back to a per-user Startup-folder login item.
 
+The hidden launcher owns the supervised Gateway process tree. Ending the task
+with `schtasks /end /tn "OpenClaw Gateway"`, `Stop-ScheduledTask`, or Task
+Scheduler's **End** action terminates the Gateway and its descendants. After
+updating an older installation, run `openclaw gateway install --force` to
+regenerate the launcher if the update did not refresh it.
+
 Gateway status and Doctor read the Scheduled Task's numeric current state, independently of the Windows display language or console code page. A previous task exit result does not prove whether it is running now. Queued or unknown tasks do not count as safely stopped for Doctor maintenance. Stop a queued task through its service owner; if inspection is inaccessible, restore Task Scheduler inspection permissions before retrying.
 
 Gateway startup creates private SQLite staging directories through Windows APIs,
--- a/docs/platforms/windows.md
+++ b/docs/platforms/windows.md
@@ -275,6 +275,15 @@ Notes:
 
 ## Troubleshooting
 
+### The Scheduled Task stops before the Gateway is ready
+
+Run `openclaw gateway status --json`, then inspect the local [Gateway log](/gateway/logging).
+Entries from `gateway/task-supervisor` record the child exit code, signal, and
+the last 8,192 characters of stderr, including failures before Gateway logging
+starts. Child stdout is discarded. A failed child or supervisor exits nonzero
+so Task Scheduler can apply its failure policy; an intentional clean stop still
+exits zero. A successful task result alone does not prove the Gateway is healthy.
+
 ### The tray icon does not appear
 
 Check Task Manager for `OpenClaw.Tray.WinUI.exe`. If it is running, open the
--- a/docs/platforms/windows.md
+++ b/docs/platforms/windows.md
@@ -286,9 +286,14 @@ Notes:
 Run `openclaw gateway status --json`, then inspect the local [Gateway log](/gateway/logging).
 Entries from `gateway/task-supervisor` record the child exit code, signal, and
 the last 8,192 characters of stderr, including failures before Gateway logging
-starts. Child stdout is discarded. A failed child or supervisor exits nonzero
-so Task Scheduler can apply its failure policy; an intentional clean stop still
-exits zero. A successful task result alone does not prove the Gateway is healthy.
+starts. Child stdout is discarded. A failed child or supervisor exits nonzero;
+an intentional clean stop still exits zero. A successful task result alone does
+not prove the Gateway is healthy.
+
+Task Scheduler's [`RestartOnFailure` policy](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tsch/2ff4aa5a-7bc4-449f-bbb1-27475645867f)
+retries failed start conditions or action launches. Do not rely on it to restart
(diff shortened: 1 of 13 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [19e1f718aabb3807f4dd7d753b64c552bb683591](https://github.com/openclaw/openclaw/pull/141744/commits/19e1f718aabb3807f4dd7d753b64c552bb683591)
- Landed commit: [64656c24fafb8a27e6ff64d764b1146147c70fab](https://github.com/openclaw/openclaw/commit/64656c24fafb8a27e6ff64d764b1146147c70fab)

PR #141746, opened 2026-09-08
fix(release): bind source admission and retain first-hop coverage

steipete:
## What Problem This Solves

Resolves three deterministic release-verification admission failures: the root-image network consumer validated the selected Code SHA against its Tooling checkout; ancestry deepening could create a hydration ref without advancing a shallow boundary; and the first-hop planner misclassified a supported mapped build declaration as unsupported.

## Why This Change Was Made

Use the existing trusted Git owner to check out the exact preflight-selected SHA without a token in a separate `.release-source` directory, then pass that directory through the network script's existing repository-root input. Executable tooling remains in the trusted harness; the network provenance and image-binding guards remain unchanged.

Seed the separate ancestry hydration ref at the already pinned target before deepening. Recognize the verified literal and mapped first-hop declaration blocks inertly, following the planner's existing content-addressed contract convention. Target build code is never executed during selection; missing or modified declarations remain unsupported, and zero-runnable execution still fails.

## User Impact

Release validation can use distinct selected Code and trusted Tooling revisions and retain required first-hop coverage without bypassing provenance, capability, or execution gates. No runtime or publication behavior changes.

## Evidence

- 159 combined workflow, planner, and postbuild tests pass, plus 12 selected real-Git ancestry tests and the scheduler's zero-runnable refusal regression.
- Actual workflow checkout uses the real Git owner against local Git transport, verifies exact HEAD and no persisted authentication, then runs the real network script. Correct source binding reaches a deliberate Docker inspection boundary; wrong or missing roots fail before Docker.
- Frozen candidate build metadata reproduces the original false first-hop omission; the corrected plan contains the required lane with no omissions. The postbuild writer and compatibility ABI tests remain green.
- Scoped type/lint/format/guard checks and workflow validation pass. Fresh combined P2 review is clean.

The entrypoint proof deliberately stops at a stubbed Docker inspection. Complete network and first-hop behavior against the exact release artifacts remains part of the release consumer runs; this PR does not claim full release verification is green.

### DIFF
--- a/.github/actions/git-owner/release-ancestry.py
+++ b/.github/actions/git-owner/release-ancestry.py
@@ -192,6 +192,9 @@ def establish_ancestry():
         return result
 
     previous_count = reachable_commit_count(source_sha, target_sha)
+    # Seed the separate destination: Git may only create a new ref on the first
+    # deepen, leaving existing shallow edges unchanged until the next fetch.
+    run_git(workspace, "update-ref", target_hydration_local_ref, target_sha, timeout=operation_timeout())
     chunk_index = 0
     while True:
         chunk = deepen_chunks[min(chunk_index, len(deepen_chunks) - 1)]
--- a/.github/workflows/install-smoke-reusable.yml
+++ b/.github/workflows/install-smoke-reusable.yml
@@ -489,6 +489,15 @@ jobs:
       - *trusted_workflow_subdir_checkout_step
       - *trusted_workflow_subdir_identity_step
 
+      - name: Checkout selected source for gateway network provenance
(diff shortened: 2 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4e1f07e3751cb6dd6d3fb75fa8f751991783eb85](https://github.com/openclaw/openclaw/pull/141746/commits/4e1f07e3751cb6dd6d3fb75fa8f751991783eb85)
- Landed commit: [009f929485cea13760d87bca66e5befb943765ad](https://github.com/openclaw/openclaw/commit/009f929485cea13760d87bca66e5befb943765ad)

PR #141754, opened 2026-09-08
fix(openrouter): release capture before cleanup waits

Alix-007:
## What Problem This Solves

With diagnostic HTTP capture enabled, OpenRouter can wait for a capture stream to reach EOF after music is complete or a usage HTTP error is already known. Awaiting one stream branch's cancellation prevents the existing request owner from aborting the transport that can settle the other branch.

## Why This Change Was Made

Music now requests reader cancellation once, releases its lock, and reaches the existing guarded request finalizer. Usage returns a known error status through that same existing release owner. Audio parsing, deadlines, billing, proxy routing, and capture policy are unchanged. No new retry, timeout, transport, or configuration is added.

The regression tests use real stream tees and explicitly settle their retained streams even when the release-order assertion fails. This preserves the contributor's repair and improves failure diagnostics.

## User Impact

Completed music becomes available without the capture-induced wait. Known usage errors return promptly. Early release can retain the existing aborted-capture terminal instead of a complete response body; each request still has one correlated terminal. Successful usage retains HTTP 200 captures.

## Evidence

Real public Gateway `tools.invoke`/`tasks.get` and cold `usage.status` calls ran on an isolated native Testbox with a loopback fixture proxy, synthetic credentials, real guarded HTTP, and capture SQLite records. The proxy never forwarded requests externally; this is controlled provider proof, not live OpenRouter inference or channel-delivery proof.

| Held-EOF case | Current main `757e482ab646` | Repaired source |
| --- | --- | --- |
| Music generation readiness after `[DONE]` | 3583 ms, after EOF | 1015 ms, before EOF; fresh independent input: 645 ms |
| Public usage HTTP 401 result | 3028 ms, after EOF | 30 ms, before EOF; fresh independent run: 23 ms |

These are fixture observations, not general benchmarks. Music's immediate background-task acceptance is not completion; proof uses generated-media progress, exact audio bytes, and eventual task status separately. Later completion-agent work remains separate.

Fresh independent inputs preserve 48-byte WAV data and lyrics. Controls cover capture disabled, immediate EOF, provider errors, malformed/interrupted audio, audio/transcript/event bounds, usage success/partial/malformed/timeout, redirects, and private-network denial. Usage balances remain $75 balance, $25 spend, and 25% key budget. Full correlated capture records and payloads are retained.

Four regression cases fail on original production because request release is never reached, then pass with the repair. All 55 provider tests and the shared request-release suite passed, as did formatting and lint. Fresh source review found no actionable P0–P2 issue. Source-only requirements and any additional unproved caller-cancellation interpretation are recorded separately from the observed public behavior.

Preserves @Alix-007's contributor commits and credit. AI-assisted verification.

The final independent assessment passes all 14 observable original obligations; the two structural requirements are covered by source review. The original interrupted-stream requirement is verified with failed early-EOF/no-partial-audio and timeout controls. The separately inferred public caller-cancellation question remains unproved and is not claimed as part of this repair. Earlier blocked judgments and export failures are preserved with the complete evidence.


obviyus:
Thanks @Alix-007!

**Proof:** Real Gateway music-task and usage-status calls ran through guarded HTTP and diagnostic capture against held-EOF fixtures. Current main waited for capture EOF; the repaired source exposed generated music and returned known usage errors before EOF.

| Before | After |
| --- | --- |
| Captured music readiness: 3583 ms | 1015 ms; fresh independent input: 645 ms |
| Captured public HTTP 401 result: 3028 ms | 30 ms; fresh independent run: 23 ms |

The existing request owner now gets to release transport without a provider-side cancellation wait. Exact WAV bytes, fresh lyrics, billing, timeout/error/early-EOF handling, size bounds, proxy routing, redirects, and private-network denial remain covered. Each captured request retains one terminal; early release records the existing abort terminal, while successful usage retains HTTP 200 captures.

All 55 provider tests and the shared release suite passed, along with source review and independent public behavior validation. These are controlled fixture observations, not general benchmarks, channel-delivery proof, or a new caller-cancellation claim. Contributor credit is preserved.

PR #141757, opened 2026-09-08
fix(matrix): avoid false readiness failures after unrelated config updates

steipete:
## What Problem This Solves

Resolves a problem where Matrix release verification rejects an already healthy, connected account after a configuration update that changes only provider model metadata. The release run had 81 passing scenarios and 12 failures at the same preparation gate; seven other scenarios needed a retry.

## Why This Change Was Made

The QA runner compared a redacted configuration snapshot with provisioned credentials and incorrectly inferred that Matrix must restart. `config.patch` now returns its canonical effective `changedPaths` (names only, including an empty list for a no-op), and Matrix preparation uses that receipt. Real Matrix credential changes still require a new account generation, alongside the existing applied-revision, health, connection, and deadline checks.

## User Impact

Matrix verification can proceed after unrelated configuration changes. Configuration API clients also receive authoritative change paths without exposing secret values. There are no configuration, protocol-version, or storage changes.

## Evidence

- Original Matrix failure artifacts: https://github.com/openclaw/openclaw/actions/runs/34175777625/artifacts/10037667475. Both representative failures show healthy accounts and Gateway writes changing only provider model arrays.
- Regression proof fails before the repair for the unrelated-write readiness timeout and missing effective token/no-op receipts, then passes with the repair.
- The five-file producer/consumer patch passed 104 owner/sibling tests (one existing skip), the full changed-file gate including all four type graphs, and independent P2 review. It applies byte-identically to current main with unchanged owner files. PR CI exposed one additional observation fixture missing the new no-op receipt; that fixture now matches production, and the observation plus environment suite passes all nine tests.
- A full Matrix transport run will qualify the integrated release candidate; the local proof covers the configuration and readiness boundaries.

### DIFF
--- a/docs/gateway/configuration.md
+++ b/docs/gateway/configuration.md
@@ -810,7 +810,11 @@ openclaw gateway call config.patch --params '{
 ```
 
 `config.patch` records explicitly supplied values in the config file even when
-they equal the current runtime defaults. Unchanged runtime defaults stay omitted.
+they equal the current runtime defaults. Unchanged runtime defaults stay omitted. Its
+successful response includes `changedPaths`, the effective runtime paths changed
+after validation and secret restoration, or `[]` for a no-op. These paths contain
+no configuration values; clients can use them to distinguish a channel change
+from an unrelated write even when secret values are redacted.
 
 Both `config.apply` and `config.patch` accept `raw`, `baseHash`, `sessionKey`,
 `note`, and `restartDelayMs`. `baseHash` is required for both methods once a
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [dfdaea5214ad7e5064f7e6c403564545b1178a17](https://github.com/openclaw/openclaw/pull/141757/commits/dfdaea5214ad7e5064f7e6c403564545b1178a17)
- Landed commit: [18b8a08c09824028fdb119148a2886f2de39be80](https://github.com/openclaw/openclaw/commit/18b8a08c09824028fdb119148a2886f2de39be80)

PR #139020, opened 2026-09-05
fix: fence pending child spawns with parent authority

steipete:
<details>
<summary>Additional instructions</summary>

Keep **Allow edits from maintainers** enabled for this PR.

</details>

Closes #139001

## What Problem This Solves

Fixes an issue where stopping a parent while `sessions_spawn` was preparing a child could still let that pending operation create session state or start a child later. ACP preparation also split failure cleanup across several owners, so a refusal could leave a provisional session or close a successor's runtime. The reproduction holds the real Gateway facade before admission, closes the exact parent invocation, and then releases the request.

## Why This Change Was Made

Carry the existing private source-invocation assertion through both ACP and native preparation to the owner of each asynchronous mutation and final Gateway admission. Recheck after actor queues, hooks, authorization, facade creation and work admission; check immediately before authoritative session and fork writes. The canonical Gateway clears the source callback after accepting the child. An accepted child's later execution remains independent of ordinary parent completion, while explicit cancellation keeps its existing cascade.

The shared spawn pipeline now retains prepared resources before checking authority and transfers ownership once. ACP failures use canonical `sessions.delete` to drain work, release the runtime, delete the provisional row and metadata, and unbind the session. A dispatch refused before entering that owner releases only the exact cached runtime captured by the failed initialization. Session ID, lifecycle revision, and handle identity protect replacements; a cleanup deadline fences late entry. `/acp` uses the same cleanup owner. Native preparation removes a redundant model write and fences queued forks, attachments and thread binding at their existing owners.

No configuration option, environment variable, public RPC field, protocol-version bump, dependency, SQLite schema, migration or stored representation changes. The callback is private, ephemeral authority from the original invocation, not a copied token, diagnostic execution identity or inferred session ownership.

## User Impact

Stopping a parent while child creation is still pending prevents that stale operation from starting a new child. Refused ACP spawns remove their own provisional resources without affecting a replacement session. Children already accepted by the Gateway continue normally when their parent completes.

## Evidence

Original 41-file qualification commit: `242f507683da94d8f2b3b83a1b1b4a69a7ae05d4`, tree `ec799c29b5a86048dece4ef0fc27b981ab75a7cb`. The commit tree exactly matches the independently tested Linux candidate.

[...]

PR #141760, opened 2026-09-08
fix(auto-reply): make silent-fallback failure text reflect the actual reason (#141694)

xiaoguomeiyitian:
## What Problem This Solves

Fixes the silent-fallback failure payload in `buildSilentFallbackFailurePayload` (`src/auto-reply/reply/agent-runner-core.ts`), which hardcoded "I couldn't reach the configured model backend" for every failure class. Operators whose provider returned HTTP 200 with a 4xx classifier (format 400, auth 401, rate_limit 429, etc.) were told the backend was unreachable — directly contradicting the partition that `resolveFailoverStatus` already makes between transport-class (5xx) and provider-responded (4xx) reasons.

## Why This Change Was Made

`ResolvedFallbackTransition` now retains the original `RuntimeFallbackAttempt`s so the failure text can consult the originating reason instead of guessing. `buildSilentFallbackFailurePayload` picks the attempt against the *selected* model ref (the run that triggered the fallback) and maps its reason to a class-appropriate sentence. The existing `resolveFailoverStatus` table is the single source of truth for which reasons are transport-class; that membership drives the reserved "couldn't reach" wording. Specific reasons (auth, auth_permanent, billing, rate_limit, context_overflow, model_not_found, format, session_expired, empty_response, no_error_details) get dedicated copy with the HTTP status where one is already mapped. Anything else falls back to `failed (${reason})` so we never claim unreachable without evidence. The heartbeat / no-fallback / explicit-silent / has-successful-delivery early returns are preserved.

## User Impact

When a primary model fails for a 4xx reason and the fallback chain returns no visible reply, the user sees a sentence that names the actual failure class (e.g. "authentication for the configured model backend x was rejected (401)", "the configured model backend x is rate-limited (429)") instead of a blanket unreachable claim. The downstream transport error — `configured model backend <selected> failed and fallback <active> produced no visible reply` — is unchanged for operators who consume it.

## Evidence

* New `buildSilentFallbackFailurePayload` describe-block in `src/auto-reply/reply/agent-runner-core.test.ts` covers every reason class (rate_limit, auth, auth_permanent, billing, model_not_found, context_overflow, format, session_expired, empty_response, no_error_details, overloaded, timeout, server_error, tls_certificate, unclassified) plus the heartbeat / no-fallback early returns.
* Full file: **25 passed (25)** in `agent-runner-core.test.ts`.
* Neighbouring `fallback-state.test.ts` and `agent-runner-result-accounting.test.ts`: **27 passed (27)** — no regressions.
* `formatProviderModelRef` is the existing helper used to compare attempt refs to the selected ref, so an attempt against a fallback candidate cannot be picked as the "selected model" reason.
* The transport-class set (`server_error`, `overloaded`, `timeout`, `tls_certificate`) is the exact 5xx/timeout partition already enumerated by `resolveFailoverStatus`, so no new vocabulary is introduced.

## CI note

`checks-node-compact-large-11` fails on the pre-existing `acp-spawn.authority.test.ts` flaky (the code being tested is identical to `main`; this PR touches ~zero ACP code). Root-cause hypothesis + follow-up plan: [comment](https://github.com/openclaw/openclaw/pull/141760#issuecomment-5578602171).

### DIFF
--- a/src/auto-reply/fallback-state.ts
+++ b/src/auto-reply/fallback-state.ts
@@ -128,6 +128,8 @@ type ResolvedFallbackTransition = {
   fallbackCleared: boolean;
   reasonSummary: string;
   attemptSummaries: string[];
+  /** Original runtime fallback attempts, in order. */
+  attempts: RuntimeFallbackAttempt[];
   previousState: {
     selectedModel?: string;
     activeModel?: string;
@@ -208,6 +210,7 @@ export function resolveFallbackTransition(params: {
     fallbackCleared,
     reasonSummary,
     attemptSummaries,
+    attempts: params.attempts,
     previousState,
     nextState,
     stateChanged,
--- a/src/auto-reply/reply/agent-runner-core.test.ts
+++ b/src/auto-reply/reply/agent-runner-core.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it } from "vitest";
+import { resolveFallbackTransition } from "../fallback-state.js";
 import type { TemplateContext } from "../templating.js";
 import {
+  buildSilentFallbackFailurePayload,
   resolveAdmittedRunSessionFile,
   resolveReplyRunDeliveryContext,
 } from "./agent-runner-core.js";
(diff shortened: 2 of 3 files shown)


xiaoguomeiyitian:
Following up on the CI signal: the only remaining failures are in the pre-existing CI infrastructure, not in this PR.

| Failed check | File / cause | Reproducible on main? |
| --- | --- | --- |
| `checks-node-compact-small-36` | `test/scripts/ci-git-owner.test.ts` — "Release ancestry fetch completed without ancestry progress." The check-out filter is not recognised by the embedded server in the CI runner. | Yes — same file is unmodified on main (`5326e8ff8e6`) and fails the same way locally. |
| `checks-node-compact-large-11` | `src/agents/subagents/spawn/acp-spawn.authority.test.ts` — "transfers initialized ACP work only from its live parent: runtime / abort" — async cleanup ordering race. | Yes — same file is unmodified on main (`b6a2e5b4eff`) and fails the same way locally. |
| `openclaw/ci-gate` | composite gate that follows the two above. | — |

`git diff main HEAD` for both files is empty. My change only touches `src/auto-reply/fallback-state.ts`, `src/auto-reply/reply/agent-runner-core.ts` and the new tests in `src/auto-reply/reply/agent-runner-core.test.ts`. None of those files are on the failing shards' path.

Re-running the same shards on a fresh main branch in this environment reproduces the same failures. They look like pre-existing flakiness, not regressions from this PR.

xiaoguomeiyitian:
## Note on `checks-node-compact-large-11` (acp-spawn authority)

That test was added in #139020 (`b6a2e5b4eff fix: fence pending child spawns with parent authority`) and the failing assertion is at `src/agents/subagents/spawn/acp-spawn.authority.test.ts:453`:

```
expect.soft(sourceBoundary.entry, "cleanup completes before spawn returns").toBeUndefined();
```

`sourceBoundary.entry` is captured **after** `await forwarded`, `await wrappedOutcome`, and `await work.drain()` (lines 422-428), so the assertion is asking the production cleanup path to guarantee the child session row is gone by the time the spawn tool promise resolves, even when the parent was aborted mid-runtime-initialisation (the `runtime / abort` parameterisation).

I have a hypothesis but want to confirm with a maintainer before patching the production path. The race window is: when `chat.abort` fires during `acpSpawn.ensureSession → pause()`, the runtime-cleanup branch in `src/agents/subagents/spawn/acp-spawn.ts` may not `await` the child entry row deletion before the spawn tool resolves (e.g. via `fire-and-forget` / `void` `deleteSessionEntryLifecycle`). Locally the test is a 1/9 miss, which matches "post-await cleanup still pending".

If a maintainer confirms that scope (or points me at the actual delete path that owns cleanup), I'm happy to file a focused follow-up PR that ties the abort path's `await` chain to the child-row deletion. I'd rather not push that change into #141760 because it's an independent bug and the diagnostic surface is different.

xiaoguomeiyitian:
Force-pushed '74f4877ddd4' (amend --no-edit; no content change, just new SHA) to retrigger the CI suite, since the pre-existing `acp-spawn.authority.test.ts` flake in `checks-node-compact-large-11` was failing the previous run. `docs/pr-context-bot` and `openclaw/ci-gate` are both going to re-run on this SHA — if the race hits again, that is pre-existing flakiness on `main` (zeroed against `src/acp/control-plane/spawn.ts` in my diagnostic comment above), unrelated to the copy/text change here.

Issue #141564, opened 2026-09-07
[Bug]: Browser panel stays stale after screencast socket disconnects


safzanpirani:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

The Browser panel keeps its last streamed image after the screencast socket disconnects, even when the Gateway client remains available.

### Steps to reproduce

1. Mount the production Browser panel in the source-mode Control UI E2E server. Provide Gateway request fixtures for tabs, viewport actions, screenshots, and screencast connection details. Connect screencast traffic to a disposable WebSocket server.
2. Deliver an image and wait for initial viewport synchronization to finish.
3. Terminate only the server-side screencast connection.
4. Wait 12 seconds without refreshing, navigating, resizing, or clicking the panel.
5. Record screenshot requests, new screencast requests, and the displayed image.

The reproduction uses Chromium 151 and a real WebSocket connection. Gateway request responses are synthetic; it does not run a real Gateway backend.

### Expected behavior

The panel should request a screenshot after the stream fails and retry streaming after its existing ten-second backoff. `docs/tools/browser.md` documents screenshot fallback for stream connection failures.

### Actual behavior

The panel retains the original image for the entire 12-second observation. It makes no screenshot request and opens no replacement socket. The Gateway client remains available.

```text
initial streamed image: displayed
screencast connection: terminated
observation: 12 seconds
screenshot requests after disconnect: 0
replacement sockets: 0
final displayed image: unchanged
Gateway client: unchanged and available
```

### OpenClaw version

Source checkout at `631267364649d1e47b331be20388bd8f80c313d3`.

### Operating system

Ubuntu 24.04.4 LTS, ARM64. Chromium 151.0.7922.34. Node 26.1.0.

### Install method

Source checkout with pnpm 12.3.4; source-mode Control UI E2E server.

### Model

Not applicable. This reproduction makes no model calls.

### Provider / routing chain

Production Control UI component → synthetic Gateway request responses and a disposable WebSocket server.

### Impact and severity

The affected Browser panel displays a stale image after an established stream fails. The controlled reproduction observed this once over 12 seconds. Production frequency is unknown.

### Additional information

Related to #141031. The ordinary socket-close handler records a failure and closes stream ownership but schedules no recovery. The existing stream-failure test calls `refreshAll()` manually, so it does not establish automatic recovery.

[...]

PR #141576, opened 2026-09-07
fix(browser): recover stale panel after stream disconnects

safzanpirani:
Closes #141564

## What Problem This Solves

Fixes an issue where the Browser panel would keep displaying a stale image after an established screencast connection failed. The Gateway client could remain available while the panel requested neither a screenshot nor a replacement stream.

## Why This Change Was Made

The stream owner now requests screenshot fallback and retries streaming through its existing ten-second backoff. Recovery cancels when ownership changes. The screenshot controller preserves captured annotation and inspection images.

Review follow-up in `68e57c98`:

- A retry reconnects the stream directly and requests a new screenshot only when no screenshot is still in flight. A failed reconnect no longer discards a pending fallback screenshot.
- A large width change during annotation or inspection no longer closes the stream. The restart waits until interaction mode resumes.

## User Impact

The panel recovers without a refresh, resize, or navigation action. Users can finish annotating or inspecting their captured image and return to interaction mode to resume streaming, including after resizing the panel.

## Evidence

- All 202 tests pass across the 14 browser panel test files. The 36 stream tests include three new regressions for the review findings; all three failed on `d05d51d6` before the fix.
- Formatting, lint, UI typecheck, and `git diff --check` pass. `check:madge-import-cycles` reports zero cycles.
- An independent reviewer checked the follow-up commit. It found no defect in either fix and confirmed that the three regressions fail on the previous head for the traced reasons. Its one remaining note concerns the open-socket case below.

The Chromium fixture loads the production Browser panel and terminates a real disposable WebSocket connection. Gateway request responses are synthetic; this does not test a real Gateway backend.

| Observation | Baseline `63126736` | PR head `68e57c98` |
| --- | --- | --- |
| Screenshot after disconnect | None within 12 seconds | Requested after 4 ms |
| Replacement stream | None within 12 seconds | Connected after 10.004 seconds |
| Display | Original image remains | Fallback screenshot, then a new streamed frame |
| Gateway client | Available throughout | Available throughout |

The timing values describe one fixture run. No manual refresh occurs.

A socket that stays open without sending its first frame remains outside this fix. The independent review of the follow-up commit flagged the same case; it behaves as it did on the previous head and on `main`.

The `checks-ui-e2e (7/13)` failure on `d05d51d6` was `chat-flow.image-handoff.e2e.test.ts`, which this change does not touch.

AI-assisted implementation and review.

PR #141763, opened 2026-09-08
fix(ui/browser): refresh screenshot and reschedule stream after screencast disconnect (#141564)

xiaoguomeiyitian:
## What Problem This Solves

Fixes the Control UI browser panel in `ui/src/components/browser/browser-panel-stream.ts`. When the WebSocket carrying the live screencast disconnected while the panel was still showing a usable frame, the panel kept the stale frame indefinitely: no screenshot fallback was issued, and the existing 10-second retry window only triggered when something else later called `ensure()`. `docs/tools/browser.md` explicitly documents a screenshot fallback for stream connection failures, so the behaviour was a regression against the documented contract.

## Why This Change Was Made

`BrowserPanelStream.onClose` now distinguishes three close classes. **4003 navigation_blocked** marks the tab unavailable and clears the visible view, unchanged. **4004 target restart** calls `refreshAll()` so the regular flow reconnects at once, unchanged. **Other closes** (the reported regression) preserve the last frame so the user keeps a view while the fallback resolves, request a fresh screenshot via `refreshView(targetId, epoch)`, and schedule an `ensure()` retry after the existing `RETRY_DELAY_MS` (10s) window so the screencast reconnects on its existing backoff.

`close()` also clears any pending retry timer + the retry target so a forced teardown cannot leave a stale reconnect behind. The `BrowserPanelStreamHost` operations subset now exposes `captureClient()` so the retry can bind a client.

## User Impact

After a non-handshake screencast disconnect the panel refreshes via a one-shot screenshot within the same idle window, and the live screencast reconnects on the existing 10s backoff instead of waiting for the next user-initiated action. The 4003 / 4004 paths are unchanged.

## Evidence

* New tests in `ui/src/components/browser/browser-panel-stream.test.ts`:
  * "requests a screenshot fallback when a non-handshake stream disconnects after presentation" — asserts `/screenshot` is called exactly once after a `disconnect(1006)` while a frame is visible.
  * "schedules a screencast retry after the disconnect backoff" — asserts `/screencast` is still 1 after 9999ms and becomes 2 after 10001ms.
  * "cancels a pending retry when the active target changes" — asserts no extra `/screencast` request fires after `setState("activeTargetId", null)`.
* Updated existing "discards a fallback screenshot overtaken by a frame (socket then closes: %s)" so the `closes=true` arm reflects the new contract (disconnect now pulls a screenshot; the previously-pulled one is still the one expected to be discarded).
* Full `ui/src/components/browser/` suite: **189 passed (189)** locally.
* `ensure()` already enforces `RETRY_DELAY_MS` backoff via `lastFailures.get(targetId)`, so the retry path is rate-limited even if multiple disconnects fire rapidly.

### DIFF
--- a/ui/src/components/browser/browser-panel-stream.test.ts
+++ b/ui/src/components/browser/browser-panel-stream.test.ts
@@ -347,17 +347,59 @@ describe("Browser panel stream ownership", () => {
       sockets[0]!.receive(screencastFrame());
       await flush();
       const streamedView = controller.view;
-      expect(streamedView?.dataUrl).toBe("blob:frame-0");
+      expect(streamedView?.dataUrl).toMatch(/^blob:frame-/);
       expect(controller.loading).toBe(false);
       if (closes) {
         sockets[0]!.disconnect(1006);
+        await flush();
+        // Stream disconnect requests a fresh screenshot to replace the stale
+        // frame; the previously-pulled screenshot is still the one resolving.
+        expect(calls("/screenshot").length).toBeGreaterThanOrEqual(1);
       }
       screenshot.resolve({ path: "/old.png", targetId: "raw-a", url: PAGE_URL });
       await pending;
       expect(controller.view).toBe(streamedView);
-      expect(fetch).not.toHaveBeenCalled();
     },
   );
+
+  it("requests a screenshot fallback when a non-handshake stream disconnects after presentation", async () => {
+    const { controller, calls } = setup();
+    const socket = await start(controller);
+    socket.receive(screencastFrame());
+    await flush();
+    const streamedView = controller.view;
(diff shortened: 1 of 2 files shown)


xiaoguomeiyitian:
Closing: #141564 was already fixed by the merged #141576 (`fix(browser): recover stale panel after stream disconnects`, merged 2026-09-08T04:10Z, author @safzanpirani). The upstream fix covers the same root cause (screenshot fallback on screencast close + recovery pinned to Retry delay) with a different architecture, and is already on main. My rebase work and the conflict resolution I did this morning is available in commit log for reference, but no longer useful to land. Picking this up elsewhere if it still shows value.

PR #141766, opened 2026-09-08
fix(opencode): backport required session identity

RomneyDa:
## Summary

- backport OpenCode's required stable `x-opencode-session` conversation header to the 2026.6 provider stack
- share one HTTPS endpoint/header resolver across Anthropic, Gemini, OpenAI Completions, and OpenAI Responses
- preserve explicitly configured headers and keep unrelated endpoints unchanged

## Root cause

OpenCode began requiring `x-opencode-session`, but the 2026.6 runtime predates the current transport-level fix. Full release validation reached the real OpenCode API and received HTTP 400: `Request is missing x-opencode-session`.

The header is conversation identity, not prompt-cache metadata, so it must use the original stream `sessionId` even when cache retention is `none`. This is the candidate-tree equivalent of main's #137464, adapted to the older `src/llm/providers` owner rather than backporting the newer `packages/ai` architecture.

## Verification

- failing hosted reproduction: Full Release Validation run 34174046853, OpenCode live provider lane
- focused resolver test: 7/7 passing
- formatter, targeted OXLint, and `git diff --check` passing
- production TypeScript: +24/-4 existing files plus one 25-line shared resolver; tests: one focused boundary file

No release, tag, or npm mutation.

### DIFF
--- a/docs/providers/opencode.md
+++ b/docs/providers/opencode.md
@@ -17,6 +17,9 @@ Both catalogs use the same OpenCode API key. OpenClaw keeps the runtime provider
 split so upstream per-model routing stays correct, but onboarding and docs treat them
 as one OpenCode setup.
 
+OpenClaw sends the stable conversation identity required by OpenCode on requests to
+`https://opencode.ai`. Direct runtime callers should provide `sessionId` in stream options.
+
 ## Getting started
 
 <Tabs>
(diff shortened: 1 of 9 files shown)


RomneyDa:
@clawsweeper review

RomneyDa:
@clawsweeper re-review

Accepted P1 fixed at b7ac49b05fc. The shared resolver is now applied once at the managed transport selector, so OpenAI Completions, Responses, Anthropic, and provider-owned Google all receive stable OpenCode conversation identity before request construction. Focused request-boundary coverage: 15/15, including cacheRetention=none.

Issue #122105, opened 2026-08-11
Gateway restarts repeat full SQLite integrity snapshots after five minutes, causing GB-scale SSD writes


keshavbotagent:
## Summary

The gateway starts `startOpenClawDatabaseIntegrityVerifier()` after becoming ready. The verifier always schedules its first pass after a fixed five-minute delay (`OPENCLAW_DATABASE_VERIFY_INITIAL_DELAY_MS`), then repeats every 24 hours.

Because the successful-run time is not persisted, every gateway restart resets that schedule. A host that restarts or upgrades the gateway repeatedly performs a fresh full-database verification five minutes after every start, even if the same databases passed minutes earlier.

This is not a steady-state five-minute interval: it is a **five-minutes-after-every-start** interval. Frequent restarts make the resulting I/O bursts appear every few minutes.

## Observed impact

On a real OpenClaw host on 2026-08-11 (timestamps normalized to Asia/Kolkata):

- Gateway became ready at approximately 20:25.
- Four database verification passes completed at approximately 20:30, matching the fixed five-minute startup delay.
- The four target databases occupied about 814 MB in aggregate.
- The gateway process reported about 1.13 GB of writes during the 15-minute startup window, despite no comparable user-message workload.
- Repeating the restart repeats the burst.

The byte count exceeds final file growth because verification prepares private, consistent SQLite snapshots. Depending on journal state, `prepareSqliteReadOnlyLocation()` either uses SQLite's online backup API or copies the main database and sidecars into a private temporary directory. Those temporary bytes are then deleted, so ordinary disk-usage checks do not reveal the write volume.

### SSD endurance and storage pressure

- Each redundant pass writes approximately the aggregate size of all registered state/agent databases, plus snapshot/WAL/filesystem overhead.
- On the observed host, ten restarts per day would generate roughly 8.1 GB of snapshot payload and approximately 11 GB of process-reported writes per day from this maintenance path alone. That is approximately 0.33 TB per month before SSD-controller write amplification.
- Larger or more numerous agent databases scale this cost linearly.
- Temporary snapshots require peak free space close to the database family being verified and can cause avoidable `ENOSPC` risk on constrained hosts.
- Repeated large sequential writes consume SSD program/erase cycles for no additional integrity coverage when the previous successful pass is still within the intended 24-hour cadence.

### Gateway and server performance

[...]

PR #141773, opened 2026-09-08
fix(sqlite): reduce temporary space for database snapshots

steipete:
Related #122105 for snapshot write volume; verification scheduling is a separate issue.

## What Problem This Solves

Fixes an issue where database inspection can exhaust temporary disk space even when one complete snapshot fits. Artifact-preserving snapshots retained two copies of a stable database before inspecting it.

## Why This Change Was Made

Keep the first fsynced private copy and compare a complete second pinned source read directly against it. Both streams must reach the same EOF, and the source descriptor and pathname identity are revalidated before promotion. Sidecar ordering, private rollback recovery, retries, cleanup, and the native worker deadline remain unchanged.

## User Impact

A stable database without sidecars needs one complete staging copy instead of two, removing one full scratch write and one scratch read. WAL and rollback-journal snapshots likewise avoid the duplicate sidecar copy.

## Evidence

- A real SQLite fixture with a 16 MiB payload is exercised under an injected synthetic per-staging-descriptor `writeSync` budget of one complete copy (not an OS quota): the original producer throws injected ENOSPC and the candidate succeeds. Byte preservation, short reads, mutation/replacement races, private permissions, and descriptor cleanup are covered.
- Focused readonly, worker, cancellation, rollback-recovery, and preflight-artifact suites: 144 passed, 2 platform skips. The final copy suite passed 11/11; the full changed gate and independent review passed.
- Both normal Linux builds passed. All 8 built snapshot/schema-preflight cases completed, with all 12 native children closing before API settlement and source families/cleanup preserved. [Testbox hosting run](https://github.com/openclaw/openclaw/actions/runs/34177418845); the successful command receipt is distinct from hosting-job cleanup state.
- On the same 9,368,645,632-byte synthetic database, one ordered warm-cache pair measured snapshot invocation at 23.612 → 12.085 seconds and schema preflight at 19.270 → 11.277 seconds. Observed large-file logical staging peak was 18,737,291,264 → 9,368,645,632 bytes for both operations.

The paired source was 9fb6b2b5ec3b2cf1889912c5d6f030ffe31d2573 plus this patch. Current main retains the same snapshot producer and native worker. Sampling provides lower bounds, and this single synthetic pair does not establish cold-cache performance, complete Gateway startup, or production latency. Full-file verification and the 30-second worker deadline were preserved.

### DIFF
--- a/src/infra/sqlite-readonly-location.copy.test.ts
+++ b/src/infra/sqlite-readonly-location.copy.test.ts
@@ -1,0 +1,316 @@
+import fs from "node:fs";
+import path from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
+import { requireNodeSqlite } from "./node-sqlite.js";
+import { prepareSqliteReadOnlyLocationSyncInProcess } from "./sqlite-readonly-location.js";
+
+const MIB = 1024 * 1024;
+const tempDirs = useAutoCleanupTempDirTracker((cleanup) => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+    cleanup();
+  });
+});
+
+function createFixture(bytes: Buffer) {
+  const sourceRoot = tempDirs.make("openclaw-readonly-copy-source-");
+  const sourcePath = path.join(sourceRoot, "source.sqlite");
+  const stagingRoot = tempDirs.make("openclaw-readonly-copy-staging-");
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5477ea0660004c51d84830e8a33558c2f51d60f3](https://github.com/openclaw/openclaw/pull/141773/commits/5477ea0660004c51d84830e8a33558c2f51d60f3)
- Landed commit: [eb9d551af5a1b8fb52fc86109b50c43205c93fbb](https://github.com/openclaw/openclaw/commit/eb9d551af5a1b8fb52fc86109b50c43205c93fbb)

PR #140954, opened 2026-09-07
fix(ui): restore authenticated agent picker avatars

LiuwqGit:
> **AI-assisted: yes**

Closes #140879

## What Problem This Solves

Fixes broken avatar images when users open the Control UI agent picker on a token-authenticated Gateway with workspace-relative or inline avatars.

## Why This Change Was Made

The picker now uses the existing authenticated avatar loader, matching the selected-agent header and standalone selector. These views also retain their text or emoji fallback when an image cannot decode, and retry presentation when its source changes. Gateway authorization and shared image caching stay intact.

## User Impact

Configured agent images render in the picker. Missing or malformed images show a fallback. Switching agents and replacing images preserve each agent's identity without putting credentials in image URLs.

## Evidence

- A real token-authenticated Gateway and Chromium reproduced the bug on main at `ade9cce816fabe84c268933da2cd8e2bba2db6a4`: both picker images failed while the selected-agent header decoded correctly.
- Matching Gateway and Control UI builds at `a4aacc5b6fa96ec19bcaca29d7a11114c4f15c06` render the expected red, blue, and inline green pixels. Missing, unreadable, malformed GIF, and unsupported files settle to their text or emoji fallback.
- Independent browser acceptance covered all 16 original clauses, including agent switching, version changes after the documented 60-second freshness window, credential/origin invalidation, cache eviction, stable rerenders, and page teardown. A 132-avatar fixture exercised image-lease release; all temporary fixtures were restored or removed.
- Unauthenticated, invalid-token, and query-token avatar requests remain `401`; valid bearer requests return `200`. Images contain no credential-bearing URLs.
- Pin order and read/unread session behavior match baseline. The aggregate picker unread dot did not appear on either build in these fixtures, so its positive display remains unverified. Arbitrary external image hosts remain blocked by the existing image policy.
- Both regression tests failed for their intended reasons before the corresponding repair. All 414 focused sidebar, selector, identity-view, and image-loader tests pass. Changed-file formatting, lint, line-limit, and assertion-safety checks pass. Full CI runs on GitHub and is required before landing.

The final follow-up replaces a duplicated TypeScript renderer declaration with the existing module type. Its emitted JavaScript is byte-identical to the browser-tested build, and lint passes against the exact CI merge that exposed the line-limit failure. The retained browser evidence therefore keeps its original build identity.

PR #141779, opened 2026-09-08
fix(ui): load sidebar agent-menu avatars through authenticated blob loader

slzszyt-ux:
## What Problem This Solves

After the sidebar agent-menu refactor, agent avatars in the agent-switcher grid (`sidebar-agent-menu__agent-grid`) fail to load when gateway token auth is active. Every `?v=...">` request returns **401** and renders as a broken/blank avatar.

Root cause: `renderAgentRow()` passes the raw local avatar URL straight to ``. The browser cannot attach the gateway `Authorization` header to an `` request, while the `/avatar/*` route requires gateway authentication. Other surfaces (chat roster, agent-select dropdown, viewer avatars) already fetch the same URLs through the authenticated identity-avatar loader and render a CSP-safe blob URL. The pre-refactor implementation also resolved these avatars through the same authenticated path, so this is a regression.

## Change

Route agent-menu row avatars through the existing `identityAvatarImage` directive (the same authenticated fetch → blob-URL pipeline used by `identity-avatar-view.ts`):

- `ui/src/components/app-sidebar-agent-menu.ts`: resolve each agent's avatar URL with `resolveAgentAvatarUrl()` and render it via `identityAvatarImage(avatarUrl, avatarUrl)` instead of the unprotected plain fallback; keep the text/emoji fallback when no image URL exists.

## Evidence

- `pnpm lint:ui:no-raw-window-open` (CI gate): 0 warnings, 0 errors
- oxlint on the changed file: 0 errors
- Unit tests: `agent-select.test.ts` 23/23 pass; `app-sidebar.test.ts` + `sidebar-menus-controller.test.ts` 347/347 pass
- Manual browser verification against a token-authenticated gateway: avatars in the sidebar agent menu now render via blob URLs; before the fix all 8 grid images returned 401 (`resourceType: image`) and rendered blank

### DIFF
--- a/ui/src/components/app-sidebar-agent-menu.ts
+++ b/ui/src/components/app-sidebar-agent-menu.ts
@@ -9,6 +9,7 @@ import type { ApplicationNavigationOptions } from "../app/context.ts";
 import type { ThemeMode } from "../app/theme.ts";
 import { t } from "../i18n/index.ts";
 import { normalizeAgentLabel } from "../lib/agents/display.ts";
+import { resolveAgentAvatarUrl } from "../lib/avatar.ts";
 import { buildExternalLinkRel, EXTERNAL_LINK_TARGET } from "../lib/external-link.ts";
 import {
   formatKeyboardShortcutCombo,
@@ -23,6 +24,7 @@ import {
 } from "../pages/debug/debug-overlay-contract.ts";
 import { renderAgentSelectAvatar, renderAgentSelectCopy } from "./agent-select.ts";
 import { icons, type IconName } from "./icons.ts";
+import { identityAvatarImage } from "./identity-avatar-view.ts";
 import "./sidebar-build-chip.ts";
 import "./viewer-facepile.ts";
 import {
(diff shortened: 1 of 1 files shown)


slzszyt-ux:
Superseded by #140954 (commit c569902a) — closing in favor of the upstream fix.

I opened this PR against `2634c04a` after diagnosing the same regression independently (sidebar agent-menu rows rendered raw `/avatar/<agentId>` URLs in ``, which 401 under token auth because `` cannot carry the Authorization header). My change routed the rows through `resolveAgentAvatarUrl()` + `identityAvatarImage`.

Upstream #140954 covers the same paths and goes further (authenticated loader wiring via the menus controller, an avatar error handler restoring text/emoji fallbacks, and test coverage), so this PR is redundant. Nothing unique to preserve here.

Verification notes that may still be useful:
- 2026.9.3 (`1391f7c`) does **not** include the fix: the shipped `control-ui-boot-shared-*.js` still calls the avatar renderer with the raw-URL fallback in the agent-menu grid, so agent-menu avatars 401 on token-authenticated gateways. Worth flagging in case other 9.3 users hit this; the fix should land in the next release.
- Until then, a server-side workaround is serving the read-only `/avatar/*` route unauthenticated.

Thanks @weiqinl @hi@obviy.us for the thorough fix.

PR #140748, opened 2026-09-07
fix(sessions): keep pins on root sessions only

steipete:
## Problem

A session pin is a bare `pinnedAt` timestamp on the session entry, and the Gateway stamped it on any key. Nothing enforced "pins are for root sessions": the agent `sessions` tool, the Sessions page menu (which lists raw Gateway rows), promoted child rows in the sidebar, and drag-to-pin all reached `sessions.patch { pinned: true }` for child sessions. On a team deployment a spawned child dashboard session was pinned through a batch `sessions.patch`, and because the sidebar's pinned zone is built from any projected root row with `pinned: true` (the active session is always promoted to a root row), the child rendered as a top-level pinned card next to the navigation routes.

## Solution

- One config-layer helper decides pinnability: not a `subagent:*` key, no non-empty `spawnedBy`, no non-empty `parentSessionKey`.
- `sessions.patch` rejects `pinned: true` on such entries with `cannot pin a child session; pin its parent session instead`, and drops a stale `pinnedAt` on any successful patch to an unpinnable entry.
- Row projection (`sessions.list` and session events), list ordering, and store-maintenance preservation ignore child pins, so existing child pins disappear without a migration.
- Control UI: sidebar rows, the session menu, the Sessions page, the chat pane header, and drag-to-pin gate on a lineage-derived `pinnable` fact (shared helper) instead of the sidebar's child-projection flag, with a localized disabled reason.
- The agent tool schema describes the root-only rule; docs updated.

## Impact

- Child/subagent sessions can no longer be pinned from the UI or the agent tool; the error text says to pin the parent instead.
- Existing child pins stop showing and no longer protect the entry from maintenance; unpinning a child still works.
- No schema, config, or doctor changes.

## Evidence

- `pnpm check:changed`: all lanes green after fixing a test-fixture type error and trimming the pruning test under the file line cap (final lint on the touched file green).
- `node scripts/run-vitest.mjs src/gateway/sessions-patch.test.ts src/gateway/session-utils.test.ts src/config/sessions/store.pruning.test.ts src/agents/tools/sessions-tool.batch.test.ts`: 3 shards passed (366 + 60 + 12 tests).
[...]

PR #141788, opened 2026-09-08
perf(gateway): skip absent pin checks in session lists

steipete:
Related: #140748

## What Problem This Solves

Large lists of unpinned sessions repeat child-session classification while sorting entries and preparing displayed rows. Those checks cannot affect an absent pin and add work to a frequent session-list request.

## Why This Change Was Made

Reuse the existing absent-pin guard at the two listing consumers. Entries with a pin still go through the unchanged root-only policy; mutation admission, sorting, filtering, paging and cache policy retain their current contracts.

## User Impact

Fresh metadata lists with 512–2,048 stored rows had 21–34% lower median response times in the paired test below, with the same results and pin behavior. The smallest 64-row/page-30 case was 0.44 ms slower; this is not a universal speedup claim.

## Evidence

A frozen common-base pair (`b1f07a0edeff` → `02a6ac142697`) ran on one fresh Linux Blacksmith Testbox. The published commit has the identical source tree to that measured candidate; only commit metadata differs. Both builds finished before timing; variant order alternated by state size. Each row below has ten fresh-list samples, timed to actual complete WebSocket response-frame receipt. No sample was removed.

| Stored rows | Page | Baseline median | Candidate median |
| ---: | ---: | ---: | ---: |
| 64 | 30 | 6.971 ms | 7.411 ms |
| 64 | 100 | 11.209 ms | 10.552 ms |
| 512 | 30 | 10.304 ms | 8.172 ms |
| 512 | 100 | 19.676 ms | 14.279 ms |
| 2048 | 30 | 20.366 ms | 15.262 ms |
| 2048 | 100 | 30.921 ms | 20.424 ms |

For the 2,048-row/page-100 shape, p90 decreased from 39.222 to 21.280 ms and maximum from 42.398 to 22.409 ms. The 64-row/page-30 candidate retained a 23.048 ms sample versus an 8.546 ms baseline maximum. These results cover sequential metadata requests, not concurrent turns or transcript-heavy sessions.

- Both revisions passed 446 tests across five focused files; candidate full changed-file checks and both full builds passed.
- All 460 request/response pairs, ordering/filter/paging/pin/error contracts, eight clean Gateway lifetimes and persisted state audits were independently verified. No admission holds occurred.
- Existing stale-child coverage now exercises limits 2 and 201, and the existing missing-store test verifies public projection of an undefined entry. No call-count test or new production test seam was added.
- P2 autoreview found no actionable issue; the later formatter-only correction was independently reviewed.
- Native Testbox command returned exit 0 ([hydration run](https://github.com/openclaw/openclaw/actions/runs/34178788952)). Complete raw artifacts were retained and hash-verified.

An earlier paired attempt stopped at formatting before any benchmark request; it remains recorded as failed. All reported timings come from the successful paired run.

### DIFF
--- a/src/gateway/session-list-order.ts
+++ b/src/gateway/session-list-order.ts
@@ -14,8 +14,10 @@ function compareSessionEntryPairs(
   sortBy: SessionsListParams["sortBy"] = "updatedAt",
 ): number {
   if (sortBy !== "lastInteractionAt") {
-    const aPinnedAt = isPinnableSessionEntry(a[0], a[1]) ? (a[1]?.pinnedAt ?? 0) : 0;
-    const bPinnedAt = isPinnableSessionEntry(b[0], b[1]) ? (b[1]?.pinnedAt ?? 0) : 0;
+    const aPinnedAt =
+      a[1]?.pinnedAt !== undefined && isPinnableSessionEntry(a[0], a[1]) ? (a[1].pinnedAt ?? 0) : 0;
+    const bPinnedAt =
+      b[1]?.pinnedAt !== undefined && isPinnableSessionEntry(b[0], b[1]) ? (b[1].pinnedAt ?? 0) : 0;
     if (aPinnedAt !== bPinnedAt) {
       return bPinnedAt - aPinnedAt;
     }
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f5cc96ca2632c56b78c08c35c7776392a1581cea](https://github.com/openclaw/openclaw/pull/141788/commits/f5cc96ca2632c56b78c08c35c7776392a1581cea)
- Landed commit: [9fa3c672dae4e27e0efcce23817388adf689be15](https://github.com/openclaw/openclaw/commit/9fa3c672dae4e27e0efcce23817388adf689be15)

PR #141794, opened 2026-09-08
fix(ci): provision ripgrep for filesystem contract tests

steipete:
## What Problem This Solves

Resolves a problem where Linux CI ran real filesystem output-contract grep tests without requesting the existing ripgrep installation step. Those tests could fail before search with “ripgrep (rg) is not available and could not be downloaded,” depending on the runner image or automatic download availability.

## Why This Change Was Made

The existing ripgrep prerequisite rule now recognizes both real grep test owners through exact selections, glob selections, grouped plans, and their unfiltered owning shards. Explicitly unrelated selections stay excluded. The installation step and all test assertions remain unchanged.

## User Impact

CI provisions the binary needed by the filesystem output-contract tests instead of relying on an ambient installation or a silent runtime download. No product runtime behavior changes.

## Evidence

The new test executes the workflow manifest and checks 12 routing cases. It fails against the previous rule and passes with this change; existing Go and ripgrep setup checks also pass. The observed CI failure contained four grep prerequisite errors while 3,475 sibling tests passed. Its silent resolver did not expose the actual download failure, so no particular external outage is claimed.

Full workflow-guard validation passes: 518 tests passed, two existing tests skipped. The full changed-file gate passes. Fresh independent P2 review found no actionable issues. A local Homebrew Bash heredoc stall was isolated and the unchanged workflow suite was rerun with system Bash; no workflow behavior or timeout was altered.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1665,18 +1665,18 @@ jobs:
                   // current isolated and Docker catalogs exclude the Go owner.
                   return (plan.shard_name ?? plan.shardName).startsWith("core-tooling");
                 }),
-                requires_ripgrep:
-                  shard.targets?.includes("src/agents/sessions/tools/index.test.ts") ||
-                  shard.includePatterns?.includes("src/agents/sessions/tools/index.test.ts") ||
-                  (shard.shardName === "agentic-agents-support" && !shard.includePatterns) ||
-                  shard.groups?.some(
-                    (group) =>
-                      group.includePatterns?.includes(
-                        "src/agents/sessions/tools/index.test.ts",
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [be954fd8836cc7d2a33cf79a641e1b29114cac57](https://github.com/openclaw/openclaw/pull/141794/commits/be954fd8836cc7d2a33cf79a641e1b29114cac57)
- Landed commit: [749f1382c0cbede6900f08ae4707e34f78487496](https://github.com/openclaw/openclaw/commit/749f1382c0cbede6900f08ae4707e34f78487496)

PR #141805, opened 2026-09-08
feat(cron): explain slow automation list requests

steipete:
Related: #140994

## What Problem This Solves

Resolves a problem where operators investigating slow automation lists can see individual source-page timings but cannot locate delays in the rest of the request. A scoped listing can exceed one second across several individually fast pages, and job projection or delivery previews can add further time.

## Why This Change Was Made

Adds one request-owned summary for `cron.list` handlers taking at least one second, using the existing diagnostics policy, Gateway logger and trace context. Fixed outer phases and scalar page/attempt totals cover listing, scope processing, projection, previews and the response callback without retaining per-job or per-page diagnostic records.

## User Impact

Operators can distinguish these elapsed-time boundaries while existing request parameters, response payloads, pagination, authorization and preview defaults retain their behavior. The summary contains no job identifiers, content, targets, query strings or error text. Timings include awaits and scheduling; callback completion does not establish client receipt.

## Evidence

- Regression cases fail on unchanged main solely because the whole-request warning is absent; response, preview-laziness and original-error assertions still pass.
- Focused cron handler, diagnostics-owner and existing source-page suites: 262 tests passed. Final assertion-label correction: all seven diagnostics-owner tests passed.
- Complete changed-source gate passed: production/core-test types, formatting, lint, all dead-export scans, import cycles and applicable owner guards.
- [Blacksmith Testbox RPC proof](https://github.com/openclaw/openclaw/actions/runs/34181383881): six authenticated synthetic operator requests compare complete payloads with diagnostics off/on, verify preview bypasses, and correlate the real structured warning with native RPC trace/span. The two full-preview requests use a fixed awaited 1,100-ms delay after a pure webhook preview result; this proves diagnostic attribution, not production latency, CPU cost or a live upstream contract.
- The remote test and wrapper exited 0, artifacts were verified, and the owned Testbox stopped. The backing Actions job can show cancellation after delegated cleanup; a later portal-sync warning is retained separately from successful command/export/stop.
- Independent autoreview found no actionable P0–P2 issue; the RPC artifacts also passed independent verification.

### DIFF
--- a/docs/gateway/logging.md
+++ b/docs/gateway/logging.md
@@ -97,6 +97,35 @@ trace context is retained when present. Emitter identity identifies the logging 
 awaited by the callback. The diagnostic adds no job identifiers,
 job contents, or request parameters.
 
+### Slow cron list requests
+
+With `diagnostics.enabled` active, a `cron.list` handler taking at least one
+second emits `cron: slow list request` through the Gateway logger. The record
+uses the existing request trace/span and reports `elapsedMs` plus fixed
+`phaseDurationsMs` for `setup`, `listing`, `projection`, optional `previews`,
+`response`, and `handlerExit`. Unentered phases are absent.
+
+`sourcePageMs` and `sourcePageCount` aggregate source-page calls, including
+failed calls. `returnedCount` appears once a page is selected.
+`scopeAttemptCount` is zero for direct lists; scoped lists allow
+three total attempts. For scoped lists, `scopeProcessingMs` is listing time
+minus source-page time: it includes visibility filtering, snapshot processing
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [68548d4cdef754073c01340a151a7504c512b8b0](https://github.com/openclaw/openclaw/pull/141805/commits/68548d4cdef754073c01340a151a7504c512b8b0)
- Landed commit: [0e7607648b329762405aee9db795ad761d825bd8](https://github.com/openclaw/openclaw/commit/0e7607648b329762405aee9db795ad761d825bd8)

PR #141774, opened 2026-09-08
fix(release): run frozen targets against shipped scenario contracts

RomneyDa:
## Summary

- qualify frozen upgrade, onboarding, and Codex package behavior with the selected release's shipped assertion modules
- run the selected release's legacy gateway client when the newer suspension client is absent
- preserve strict current-target behavior and require the existing frozen-target authorization/identity boundary
- let legacy update CLIs prove dirty-worktree rejection from their structured result even when they historically exited zero
- bind the root Docker install smoke to the selected source checkout

## Root cause

Full Release Validation used current-main scenario assertions and gateway setup against the immutable 2026.6.35 candidate. Those assertions require later SQLite exec-approval migration, current model defaults, plugin installation phases, and `admin-http-rpc`; the selected release shipped different valid contracts. This produced false failures after the package had completed its own behavior correctly.

The fix reuses target-owned, already-shipped scenario artifacts behind the canonical frozen-target authorization check. It does not add versions, SHA allowlists, public config, or a compatibility profile.

## Evidence

Failing full validation: https://github.com/openclaw/openclaw/actions/runs/34174003947

Local:
- `bash -n` on all changed runners
- focused Docker harness contracts: 2/2
- frozen target-file fallback/fail-closed regression: 1/1
- broader affected tooling suite before the final assertions: 770 passed, 2 skipped
- exact-head autoreview: scoped-clean (0.99), no actionable findings
- `git diff --check`

Production/workflow: +90/-10. Tests: +51/-3. Application runtime: 0. The six scenario selectors share one fail-closed target-file resolver; no per-scenario authorization logic remains.

PR #141814, opened 2026-09-08
fix(release): preserve frozen gateway client imports

RomneyDa:
## Summary

- mount the authorized frozen target’s complete `scripts/e2e/lib` tree read-only
- run its Gateway client at the original relative path so sibling imports resolve
- leave the current-target Gateway client and suspension proof unchanged

## Root cause

Merged #141774 relocated only `scripts/e2e/lib/gateway-network/client.mjs` to `/tmp/openclaw-gateway-network-client.mjs`. The exact 2026.6.35 client imports `../websocket-open.mjs`, `./limits.mjs`, and `./ws-frames.mjs`, so Node resolves those under `/tmp` and fails before connecting.

## Verification

- regression failed before the repair and passes after it
- exact 2026.6.35 client resolves its three relative imports from the preserved tree and reaches the expected missing-runtime-env boundary
- focused Vitest: 1 passed
- `bash -n scripts/e2e/gateway-network-docker.sh`
- changed-file checks passed, including test-root typecheck and scripts/test lint
- autoreview scoped-clean, no P0/P1 findings (0.94)

Production/tooling: +6/-6. Tests: +3/-1. No config, schema, protocol, dependency, or runtime-product changes.

### DIFF
--- a/scripts/e2e/gateway-network-docker.sh
+++ b/scripts/e2e/gateway-network-docker.sh
@@ -10,11 +10,11 @@ openclaw_prepare_frozen_target_context "$SOURCE_ROOT" && FROZEN_CONTEXT=1 || {
   context_status=$?
   [ "$context_status" -eq 1 ] || exit "$context_status"
 }
-LEGACY_GATEWAY_CLIENT=""
+LEGACY_GATEWAY_LIB=""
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper review

PR #140672, opened 2026-09-07
fix(runtime): require Node builds with lossless SQLite reads

steipete:
## What Problem This Solves

Related: #140465. Reported by @aeoess.

Previously supported Node releases silently truncate SQLite TEXT at embedded NUL characters. SQLite retains the correct bytes, but Node loses the suffix when converting to JavaScript strings. This breaks session identity validation and suggestion accounting and reproduces four existing session regressions.

## Why This Change Was Made

Requires `engines.node = >=24.16.0 <25 || >=26.1.0` consistently across package metadata, runtime diagnostics, installers, native runtime discovery, worker defaults, documentation, and compatibility CI. Node 22.23.x, 24.15.0, 25.9.0, and 26.0.0 contain the broken decoder; 24.16.0 and 26.1.0 are the first fixed releases on their lines. [Node #61954](https://github.com/nodejs/node/pull/61954) makes the conversion length-bearing. The separate SQLite WAL-version gate remains intact.

Peter approved this compatibility change and the platform losses below on September 7, 2026. Corrected native bindings address every SQLite text-read boundary without changing schemas or persistent-store semantics.

The worker default becomes `node:24.19.0-slim`; explicit worker images remain operator-owned. The sealed worker entry checks the existing runtime contract. The rootless installer validates a downloaded runtime before activating its alias, preserving an existing runtime when validation fails. The macOS Gateway-error fixture uses supported Node 24.16.0 while retaining its error assertions.

Windows test temporary roots use native path resolution before fixture allocation and child-environment publication. This expands inherited 8.3 aliases before Node 24 filesystem watchers encounter them. Explicit custom-root spelling and cleanup ownership are preserved. Both regression files are included in the existing Windows CI lane without adding jobs or changing timeouts.

## User Impact

- Node 22 and 25 users, users on Node 24 before 24.16.0, and users on Node 26.0 must upgrade Node before updating OpenClaw. Node 22 remains maintenance LTS; this support change is driven by correctness.
- Official Linux ARMv7 provisioning is dropped because Node 24+ has no official ARMv7 binaries. Use a supported 64-bit OS or host.
- Supported Node-based CLI/Gateway installation on macOS 11 through 13.4 is dropped because official Node 24/26 binaries require macOS 13.5. The native app independently requires macOS 15.0, documented by #140578.
- Replace explicitly configured older worker images before upgrading. Older Node syntax support can cause startup failure before the runtime diagnostic executes.
[...]

Issue #140465, opened 2026-09-06
Node 22.23.2 is within engines, but four session tests fail on embedded-NUL SQLite reads


aeoess:
Bug type: Test suite / developer environment
Beta release blocker: No

Summary

`pnpm test src/config/sessions` fails on Node 22.23.2 with 4 failed / 1536 passed. The same commit, same tree and same command pass 1540/1540 on Node 24.20.0, which is what `ci.yml` runs.

22.23.2 is inside the repo's own `engines` range and CONTRIBUTING says to run the Vitest suite on Node 22.22.3+. So a contributor on Node 22 sees red on main for tests that CI never runs on that runtime. The `checks-node-compat-node22` job is pinned to 22.23.2 but only runs on `workflow_dispatch` and does build, `--help`, `status --json` and the build singleton test, not vitest.

Steps to reproduce

1. `n 22.23.2` (or any Node 22 build with the same node:sqlite behaviour)
2. `pnpm install --frozen-lockfile && pnpm build`
3. `pnpm test src/config/sessions`

Direct probe of the mechanism, same script on both runtimes:

```text
node v22.23.2 | sqlite 3.51.3
  wrote 6 chars, read back 2 chars -> TRUNCATED
  stored bytes: 206100F09FA69E20

node v24.20.0 | sqlite 3.53.4
  wrote 6 chars, read back 6 chars -> PASS
  stored bytes: 206100F09FA69E20
```

The bytes in the database are identical. Only the TEXT to JS string conversion truncates at the first NUL on the 22 build.

Expected behavior

The four tests pass, or the suite makes it visible that they need Node 24.

Actual behavior

```text
node v22.23.2   pnpm test src/config/sessions
  exit 1 | Test Files 3 failed | 123 passed (126) | Tests 4 failed | 1536 passed (1540)

node v24.20.0   pnpm test src/config/sessions
  exit 0 | Test Files 126 passed (126) | Tests 1540 passed (1540)
```

The four:

- `session-suggestion-store.test.ts` > bounds pending suggestions for embedded NUL author IDs
- `session-sharing-store.test.ts` > preserves membership identity checks for opaque identity
- `session-sharing-store.test.ts` > preserves membership identity checks for literal NUL
- `session-accessor.creation-snapshot.test.ts` > preserves warm listing behavior for a nul target

[...]

PR #141822, opened 2026-09-08
fix: keep Node 22 supported for SQLite session IDs

RomneyDa:
Related: #140465
Related: #140672

## What Problem This Solves

Fixes an issue where OpenClaw users on supported Node 22 could get incorrect session identity, sharing, suggestion-accounting, and warm-listing behavior when stored text contained an embedded NUL byte. The earlier mitigation raised the entire OpenClaw runtime floor to Node 24.16+, which forced unrelated users to upgrade and dropped Node 22/ARMv7 support.

## Why This Change Was Made

The failure is in affected `node:sqlite` TEXT decoding, not in OpenClaw's schema or SQLite itself. At the exact session read boundaries that need lossless text, this change projects TEXT as BLOB before `node:sqlite` can truncate it, then decodes it in JavaScript according to the database's UTF-8/UTF-16 encoding. A code comment at those projections documents why the apparently unusual cast is required.

This restores the previous supported ranges (`>=22.22.3 <23`, `>=24.15 <25`, and `>=25.9`) across runtime guards, installers, workflows, Docker/worker defaults, macOS runtime discovery, and public docs. The independent minimum versions that contain SQLite's WAL-reset corruption fix remain enforced.

## User Impact

Node 22 users can continue installing and running OpenClaw without a forced major-version upgrade. Session identifiers and JSON containing literal NUL bytes remain lossless on affected Node 22 and Node 24 builds. ARMv7 worker images can continue using the available Node 22 runtime.

## Evidence

- Reproduced on the checksum-verified official Node 22.23.2 macOS arm64 binary before the fix: **4 failed, 34 passed** across the three session regression files, matching #140465.
- Ran those same files with the same Node 22.23.2 binary after the fix: **38 passed, 0 failed**.
- Ran the same regression files on affected Node 24.15.0 after the fix: **38 passed, 0 failed**.
- The expanded Node 22.23.2 proof passes **118 tests, 0 failed**, covering UTF-8, UTF-16le, and UTF-16be databases; malformed literal-NUL JSON; valid escaped NULs; ordinary/exact reads; maintenance/reclamation projections; sharing; suggestions; warm listing; and sealed-worker runtime admission.
- The wider runtime/installer/workflow regression set completed **1,052 passed, 30 skipped**, with one expected guard assertion updated for the focused Node 22 suite; its focused rerun passes.
- `pnpm build` passed as part of the wider regression run.
- `pnpm check:changed` passed, including all core/UI/extension/script/test typechecks, lint, dead-code, architecture, Kysely guardrails, formatting, generated baselines, and dependency checks.
- The retained installer activation-safety regression matrix passes all four fresh/existing and launchable/unlaunchable cases.
- Manual Node 22 compatibility CI now runs the embedded-NUL identity regression alongside the existing compatibility regression files so removal of the mitigation cannot silently regress.


RomneyDa:
Decided against this

RomneyDa:
@clawsweeper re-review\n\nAddressed both findings on the latest head: all ordinary/exact/maintenance session readers now use encoding-aware lossless BLOB projections, with Node 22 coverage for embedded-NUL identities and UTF-8/UTF-16 parser parity; the sealed worker entry again enforces the restored supported runtime range before starting work. Official Node 22.23.2 targeted proof passes 118/118.

PR #141824, opened 2026-09-08
refactor(gateway): keep cached turn replay with dedupe

steipete:
## What Problem This Solves

Cached agent-turn replay lives in the turn orchestration module, separate from its existing dedupe reader and payload classifier. Moving it to that owner makes the later lifetime changes in #140674 easier to review independently.

## Why This Change Was Made

Move the unchanged replay helper into the existing dedupe module and import it directly from the turn service. Its parameters describe the two preflight fields it reads and use the existing Gateway context type. The call remains before lifecycle generation, reservations, and awaited preparation.

## User Impact

No user-visible behavior change. Accepted, completed, and failed cached responses keep their payload and metadata. Cache misses continue normal admission. No cache policy, resource lifetime, schema, protocol, or SDK change.

## Evidence

The complete helper body is byte-identical to the original, and all other statements remain unchanged. The new type import is erased; both normalization helpers already belong to the module's runtime dependency graph. No lazy or bundled-plugin import boundary changes.

All 22 existing preflight tests passed before and after the move. Wall times were 35.39 and 14.48 seconds; the second run used warmed caches, so these are validation timings, not a speedup claim. Two existing isolated Gateway WebSocket cases passed in 29.03 seconds, covering completed-response dedupe and replay after reconnect. No tests were changed or added.

Canonical changed checks passed in 405.53 seconds, including core and test types, lint, and zero runtime import cycles. Independent Codex review found no actionable findings. The diff touches two production files (+44/-42 lines), with no compatibility alias or duplicated helper left behind.

### DIFF
--- a/src/gateway/agent-turn/agent-dedupe.ts
+++ b/src/gateway/agent-turn/agent-dedupe.ts
@@ -1,6 +1,9 @@
+import { asOptionalRecord } from "@openclaw/normalization-core/record-coerce";
+import { normalizeOptionalString } from "@openclaw/normalization-core/string-coerce";
 import { uniqueStrings } from "@openclaw/normalization-core/string-normalization";
 import type { GatewayRequestContext } from "../server-methods/types.js";
 import { setGatewayDedupeEntry } from "./agent-job.js";
+import type { AgentTurnIo } from "./types.js";
 
 export function resolveAgentDedupeKeys(params: {
   idempotencyKey: string;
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [51f4ca6e3e012a2201d31adbd089694ea54765ec](https://github.com/openclaw/openclaw/pull/141824/commits/51f4ca6e3e012a2201d31adbd089694ea54765ec)
- Landed commit: [98c78e2573b10078def6c74518240a904d584aeb](https://github.com/openclaw/openclaw/commit/98c78e2573b10078def6c74518240a904d584aeb)

PR #141827, opened 2026-09-08
fix(release): preserve frozen gateway package resolution

RomneyDa:
## Summary

- mount the authorized selected-release gateway client at its canonical `/app/scripts/e2e/lib` path
- preserve both relative ESM imports and bare package resolution through `/app/node_modules`
- leave the strict current-harness path unchanged

## Root cause

Full validation job https://github.com/openclaw/openclaw/actions/runs/34184637525/job/101931606884 failed because the selected client was copied under `/tmp`. Its relative imports resolved there, but the bare `ws` import could not walk to `/app/node_modules`. The later, client-only bind now shadows the broad harness bind at the original package root.

## Validation

- red before fix: focused Docker contract expected the canonical mount and failed on the `/tmp` path
- green after fix: focused Docker contract 1/1
- `bash -n scripts/e2e/gateway-network-docker.sh`
- Oxfmt, OXLint, diff check
- changed-file gate
- AutoReview: correct, no P0/P1 findings (0.94)

## LOC

- release tooling: +2/-2
- tests: +3/-2
- application runtime: 0

### DIFF
--- a/scripts/e2e/gateway-network-docker.sh
+++ b/scripts/e2e/gateway-network-docker.sh
@@ -106,11 +106,11 @@ if [[ -n "$LEGACY_GATEWAY_LIB" ]]; then
     "${DOCKER_E2E_HARNESS_ARGS[@]}" \
     --network "$NET_NAME" \
     "${CLIENT_LIMIT_ENV_ARGS[@]}" \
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper review

PR #141849, opened 2026-09-08
feat(google): add Gemini 3.8 Flash to static catalog

spiderman-02:
## What Problem This Solves

Google's static fallback catalog omits Gemini 3.8 Flash. Catalog consumers that cannot use authenticated runtime discovery therefore cannot expose the direct Google model.

## Why This Change Was Made

Adds the documented Gemini 3.8 Flash catalog entry to the canonical Google plugin manifest and covers both static-catalog and authenticated `models.list` discovery behavior.

## User Impact

Google AI Studio and Vertex users can select Gemini 3.8 Flash through the static catalog. Downstream catalog mirrors can publish the matching direct Google model route without a hand-maintained provider entry.

## Evidence

- Google documents `gemini-3.8-flash` with a 1,048,576-token context window, 65,536 maximum output tokens, and no `minimal` thinking level: https://ai.google.dev/gemini-api/docs/models/gemini-3.8-flash?authuser=2
- Static catalog regression was demonstrated before the change (`10 !== 11`); the post-change JSON contract assertion and `git diff --check` pass.
- Targeted Vitest and changed-file checks could not run because this checkout has no `node_modules`; the repository-pinned pnpm 12.3.4 launcher fails with `ENOEXEC` in this environment. CI remains the required test gate.
- Required local `autoreview` was invoked but could not send its review pack because TruffleHog is unavailable in this environment.

### DIFF
--- a/extensions/google/manifest.test.ts
+++ b/extensions/google/manifest.test.ts
@@ -134,7 +134,8 @@ describe("google manifest model catalog", () => {
       api: "google-generative-ai",
       baseUrl: "https://generativelanguage.googleapis.com/v1beta",
     });
-    expect(modelIds).toHaveLength(10);
+    expect(modelIds).toHaveLength(11);
+    expect(modelIds).toContain("gemini-3.8-flash");
     expect(modelIds).not.toContain(undefined);
     expect(new Set(modelIds).size).toBe(modelIds.length);
   });
--- a/extensions/google/openclaw.plugin.json
+++ b/extensions/google/openclaw.plugin.json
@@ -180,9 +180,6 @@
             "maxTokens": 65536,
             "thinkingLevelMap": {
               "minimal": null
-            },
-            "compat": {
-              "codeMode": "preferred"
             }
           },
           {
--- a/extensions/google/openclaw.plugin.json
+++ b/extensions/google/openclaw.plugin.json
@@ -163,6 +163,29 @@
             }
           },
           {
+            "id": "gemini-3.8-flash",
+            "name": "Gemini 3.8 Flash",
+            "reasoning": true,
+            "input": [
+              "text",
+              "image"
+            ],
+            "cost": {
+              "input": 0,
+              "output": 0,
+              "cacheRead": 0,
+              "cacheWrite": 0
+            },
+            "contextWindow": 1048576,
+            "maxTokens": 65536,
+            "thinkingLevelMap": {
+              "minimal": null
+            },
+            "compat": {
+              "codeMode": "preferred"
+            }
+          },
+          {
             "id": "gemini-3.7-flash",
             "name": "Gemini 3.7 Flash",
             "reasoning": true,
--- a/extensions/google/provider-catalog.test.ts
+++ b/extensions/google/provider-catalog.test.ts
@@ -36,7 +36,6 @@ describe("google provider catalog", () => {
       reasoning: true,
       input: ["text", "image"],
       thinkingLevelMap: { minimal: null },
-      compat: { codeMode: "preferred" },
     });
     expect(provider.models.find((model) => model.id === "gemini-3.7-flash")).toMatchObject({
       contextWindow: 1_048_576,
@@ -172,7 +171,6 @@ describe("google provider catalog", () => {
         contextWindow: 1_048_576,
         maxTokens: 65_536,
         input: ["text", "image", "video"],
-        compat: { codeMode: "preferred" },
         thinkingLevelMap: { minimal: null },
       }),
       expect.objectContaining({
--- a/extensions/google/provider-catalog.test.ts
+++ b/extensions/google/provider-catalog.test.ts
@@ -27,8 +27,17 @@ describe("google provider catalog", () => {
         "gemini-3.5-flash-lite",
         "gemini-3.6-flash",
         "gemini-3.7-flash",
+        "gemini-3.8-flash",
       ]),
     );
+    expect(provider.models.find((model) => model.id === "gemini-3.8-flash")).toMatchObject({
+      contextWindow: 1_048_576,
+      maxTokens: 65_536,
+      reasoning: true,
+      input: ["text", "image"],
+      thinkingLevelMap: { minimal: null },
+      compat: { codeMode: "preferred" },
+    });
     expect(provider.models.find((model) => model.id === "gemini-3.7-flash")).toMatchObject({
       contextWindow: 1_048_576,
       maxTokens: 65_536,
@@ -93,6 +102,14 @@ describe("google provider catalog", () => {
                     thinking: true,
                   },
                   {
+                    name: "models/gemini-3.8-flash",
+                    displayName: "Gemini 3.8 Flash",
+                    inputTokenLimit: 1_048_576,
+                    outputTokenLimit: 65_536,
+                    supportedGenerationMethods: ["generateContent"],
+                    thinking: true,
+                  },
+                  {
                     name: "models/gemma-3-1b-it",
                     displayName: "Gemma 3 1B",
                     inputTokenLimit: 32_768,
@@ -151,6 +168,16 @@ describe("google provider catalog", () => {
       expect.objectContaining({
         id: "gemini-3.7-flash",
         name: "Gemini 3.7 Flash",
+        reasoning: true,
+        contextWindow: 1_048_576,
+        maxTokens: 65_536,
+        input: ["text", "image", "video"],
+        compat: { codeMode: "preferred" },
+        thinkingLevelMap: { minimal: null },
+      }),
+      expect.objectContaining({
+        id: "gemini-3.8-flash",
+        name: "Gemini 3.8 Flash",
         reasoning: true,
         contextWindow: 1_048_576,
         maxTokens: 65_536,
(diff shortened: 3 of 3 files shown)


fenglanhua:
Real behavior proof (before fix) on OpenClaw 2026.9.3 (1391f7c), Node v26.8.1, Rocky Linux 10, Google AI Studio authenticated:

$ openclaw models list --provider google
Model                                      Input      Ctx         Local Auth  Tags
google/gemini-2.5-flash                    text+image 1049k       no    yes
google/gemini-2.5-flash-lite               text+image 1049k       no    yes
google/gemini-2.5-pro                      text+image 1049k       no    yes
google/gemini-3-flash-preview              text+image 1049k       no    yes   default,configured
google/gemini-3.1-flash-lite               text+image 1049k       no    yes
google/gemini-3.1-pro-preview              text+image 1049k       no    yes
google/gemini-3.5-flash                    text+image 1049k       no    yes
google/gemini-3.5-flash-lite               text+image 1000k       no    yes
google/gemini-3.7-flash                    text+image 1000k       no    yes
google/gemini-3.6-flash                    text+image 1049k       no    yes

google/gemini-3.8-flash is absent from the authenticated live catalog on current 9.3. Manually adding it to modelPolicy.allow does not resolve it — the model remains unknown at runtime.

potterdigital:
Independent after-fix proof for this change, from a production OpenClaw deployment with an authenticated Google AI Studio key. I hit this gap on my own install before finding this PR, so this is corroboration plus the after-fix half that the review is still asking for.

The catalog entry in this PR is absent from `main` as of `912b21b98`, and also absent from the published `2026.9.5` tarball (checked with `npm pack openclaw@2026.9.5` and reading `package/dist/extensions/google/openclaw.plugin.json` without installing). So the change is still needed.

Without it, the only way to reach the model is a hand-maintained provider entry in user config, which is the exact workaround this PR's User Impact section says users should not need. In my case that meant registering the model under `models.providers.google.models[]` and setting `models.providers.google.baseUrl`, because a model that exists only in user config takes the OpenAI-compatible transport and stops at `Provider "google" requires an explicit base URL before using an OpenAI-compatible API`. With this PR's catalog entry in place, none of that is required.

## Real behavior proof

**Behavior addressed:** `google/gemini-3.8-flash` cannot be selected from the static Google catalog, so an agent run against it fails with `Unknown model` even on an authenticated Google AI Studio account.

**Real environment tested:** OpenClaw 2026.9.4 (`3a9d69d`), Node v24.18.0, Debian GNU/Linux 12 (bookworm), unprivileged LXC container, gateway running as a systemd user unit. Google AI Studio API key authenticated through a store SecretRef. Nine live agents on the gateway. `models.providers.google.models` empty and `models.providers.google.baseUrl` unset for both the before and after runs, so the catalog entry is the only variable.

**Exact steps or command run after this patch:**

1. Restored the pristine `dist/extensions/google/openclaw.plugin.json` from the installed 2026.9.4 and confirmed it carries 7 `gemini-3.x` ids with no `gemini-3.8-flash`.
2. Cleared all local workarounds: `openclaw config set models.providers.google.models '[]' --strict-json --replace` and `openclaw config set models.providers.google.baseUrl ''`.
3. Before run on gateway pid 447: `openclaw agent --agent main --model google/gemini-3.8-flash -m 'Reply with exactly: PROOF-OK'`
4. Inserted this PR's `gemini-3.8-flash` catalog object into the same manifest ahead of `gemini-3.7-flash`, with the same ids, values and field order as the diff. The surrounding file was re-serialized by my insert script, so it is not byte-identical to the branch; the entry's contents are.
5. Restarted the gateway with the deployment's own `safe-gateway-restart.sh`, which reported `All clear` and completed in 12 seconds. New gateway pid 184108.
6. After run, same command as step 3, plus a content prompt and a regression check on `google/gemini-3.7-flash`.

**Evidence after fix:**

Before, pristine catalog, pid 447:

```
$ openclaw agent --agent main --model google/gemini-3.8-flash -m 'Reply with exactly: PROOF-OK'
Unknown model: google/gemini-3.8-flash
```

After, this PR's catalog entry applied, pid 184108, user config still empty:

```
$ openclaw config get models.providers.google.models
[]
$ openclaw config get models.providers.google.baseUrl

$ openclaw agent --agent main --model google/gemini-3.8-flash -m 'Reply with exactly: PROOF-OK'
PROOF-OK
[exit=0]

$ openclaw agent --agent main --model google/gemini-3.8-flash \
    -m 'In one sentence: name the capital of Burkina Faso and give 37 x 43.'
The capital of Burkina Faso is Ouagadougou, and 37 × 43 is 1,591.

$ openclaw agent --agent main --model google/gemini-3.7-flash -m 'Reply with exactly: OK'
OK
```

To rule out a fallback answering in place of the requested model, I counted `model_fallback_decision` events in the gateway JSON log for the window covering these runs: 0. The replies came from `gemini-3.8-flash` itself.

**Observed result after fix:** `google/gemini-3.8-flash` resolves and serves real traffic through the native Google transport with zero user-side model or endpoint configuration. Content is correct on a non-trivial prompt. `gemini-3.7-flash` still works, so the inserted entry does not disturb the neighboring catalog entries.

**What was not tested:** I did not build this PR's branch or run its Vitest changes. I applied the catalog object from the diff to the built manifest of an installed 2026.9.4, because the assertion under review is catalog data and the runtime reads that manifest directly. The two test files in this PR are therefore unverified by me and still need CI. I also did not test Vertex, only AI Studio, and I did not exercise image input or the `thinkingLevelMap` behavior, only text generation and tool-free replies.

potterdigital:
Follow-up to my proof above. I closed most of the gaps I listed, including the two test files.

## Vitest on this branch, which the PR body lists as unrun

Ran on branch head `017230e784` in a clean clone. The `ENOEXEC` the author hit looks environmental: `corepack enable && corepack prepare --activate` provisioned the branch-pinned pnpm 12.3.4 and installed in 25.5s. macOS 15, Node v24.18.0.

```
$ node scripts/run-vitest.mjs run extensions/google/manifest.test.ts extensions/google/provider-catalog.test.ts
 Test Files  2 passed (2)
      Tests  23 passed (23)
   Duration  7.01s
```

Both files pass as submitted.

## The assertions genuinely gate the data

To confirm the new assertions are not tautological, I reverted only `extensions/google/openclaw.plugin.json` to its merge-base state, kept the new test code, and re-ran:

```
 FAIL  extensions/google/manifest.test.ts > owns the canonical Google model catalog and runtime discovery declaration
AssertionError: expected [ 'gemini-2.5-pro', …(9) ] to have a length of 11 but got 10
 FAIL  extensions/google/provider-catalog.test.ts > registers current Gemini rows for the Google Vertex provider
 FAIL  extensions/google/provider-catalog.test.ts > builds the authenticated text catalog from Google models.list metadata
 Test Files  2 failed (2)
      Tests  3 failed | 20 passed (23)
```

Restoring the file returns all 23 to green. So the catalog data is what those three tests measure.

## thinkingLevelMap

The entry sets `thinkingLevelMap: { minimal: null }`. Exercised every level the CLI accepts against `google/gemini-3.8-flash` on the live gateway. All five return a correct reply, and `minimal` in particular does not error:

```
--thinking off      T-OK
--thinking minimal  T-OK
--thinking low      T-OK
--thinking medium   T-OK
--thinking high     T-OK
```

I did not inspect the outbound request payload, so this establishes acceptance rather than the exact field sent.

## Tool-bearing turn

Earlier I only tested tool-free replies. Wrote an unguessable token to a file on the container and asked the model to read it with its file tool:

```
$ openclaw agent --agent main --model google/gemini-3.8-flash \
    -m 'Use your file-reading tool to read /tmp/tool-secret.txt and reply with only its exact contents.'
EEGFAVKL
```

The value was randomly generated and written only to that container, so returning it exactly means a real tool call ran and its result came back through the Google transport intact.

## Image input

Closed. The catalog entry declares `input: ["text", "image"]`, so I checked the model actually receives image content through OpenClaw.

I rendered a PNG containing a random 10 character code and a number of triangles, put it on the container, and confirmed no text copy of either value existed anywhere on the box. The only place that code exists is as pixels.

```
$ openclaw agent --agent main --model google/gemini-3.8-flash \
    -m 'Open the image file /tmp/v2.png and look at it. Transcribe the printed code exactly
        and count the orange triangles. Reply as: CODE=<code> TRIANGLES=<n>'
CODE=DT4HTQGKVC TRIANGLES=7
```

Ground truth was `DT4HTQGKVC` and 7. A first attempt at this used a code that also existed in a text file from the tool test above, which would have let a file read stand in for vision, so I discarded it and regenerated with a pixel-only value.

Worth noting for anyone trying to reproduce this through the media-understanding CLI instead: `openclaw capability image describe` fails on this deployment with `AgentSelectionRequiredError` no matter what `--agent` is passed, and it fails identically for `google/gemini-3.7-flash`, so that is a separate multi-agent CLI issue on 2026.9.4 and not related to this PR. The agent path above works fine.

## Still not tested, and why

**Live Vertex inference.** I authenticated gcloud and refreshed ADC to try this, but the Google Cloud project reachable from this account has no active billing account, and Vertex requires one. So a live Vertex call is not something I can produce. The Vertex side is not unverified in this PR either: `registers current Gemini rows for the Google Vertex provider` is one of the three tests I showed failing without the catalog data, so the Vertex catalog rows are covered at test level. Someone with a billed Vertex project would still be the right person to confirm `gemini-3.8-flash` is actually served there, since AI Studio availability does not guarantee Vertex availability.

potterdigital:
@clawsweeper re-review

Two proof comments were added since the last review: after-fix behavior on a real authenticated deployment, and the branch test run the PR body lists as unrun (23 passing, plus a revert showing the new assertions fail without the catalog data).

PR #141104, opened 2026-09-07
fix(memory): use fallback search when SQLite extensions are unavailable

steipete:
## What Problem This Solves

Memory search under macOS Bun can omit vector-backed results when its isolated KNN child uses SQLite without extension-loading support. Selecting a custom SQLite library in the parent does not configure that child.

## What Changed

After validating its request and opening the database read-only, the child checks the existing extension capability and requests the existing batched embedding scan when native KNN is unavailable. Native-capable extension and query failures retain their existing behavior. The existing provider/source filters, cancellation, query-only mode, and close remain intact.

The final resolution preserves the lightweight KNN import boundary introduced by #140730: the child, query helper, and existing capability query use the narrow memory-core-host-engine-knn SDK barrel. The generic SQLite barrel is unchanged. The Bun guide explains the degraded scan and its cost; Node remains recommended for native sqlite-vec KNN on macOS.

## Evidence

- On the refreshed current-main donor, the exact synthetic hybrid case passed under Node and failed under Bun with the expected missing result before the repair; it passed under both afterward.
- Two existing fallback yielding/cancellation cases passed under each runtime. The supported Node WAL/source-filter child case also passed, including its current upstream import guard.
- Focused runs used isolated secretless HOME/state, original deadlines, and a 4 GiB process-tree limit. All descendants joined without a cap. No tests were changed.
- Fresh independent review, the complete four-path changed gate, and the full build passed on the refreshed donor. Emitted-source inspection confirmed the narrow SDK chain and preserved read-only/fallback/close behavior.
- Final signed commit `7d603675e01146afb00f73c5a39de0eb7d1c1d42` has all four donor file hashes. The generic SQLite barrel equals its actual main parent. Final-head CI34158721753 and pull_request Workflow Sanity34158600721 passed. The completed20:21 hosted review at this exact head confirms the prior source finding is resolved and reports no actionable code or security finding. Its generated data-model checklist remains rejected on the source evidence below; no marker was edited.

The temporary draft bridge is complete. It preserved ancestry while integrating the conflicting upstream import refactor; no intermediate head was ready for merge.

## Data-model Compatibility

There is no schema, migration, metadata producer, index format, embedding serialization, or persistent write change. The four-file diff contains the child capability branch, shared existing validator export, narrow SDK capability export, and documentation. The existing fallbackScanRequired response and parent fallback reader remain unchanged.

[...]

PR #141854, opened 2026-09-08
feat(sqlite): select an extension-capable SQLite library for Bun on macOS

steipete:
## What Problem This Solves

Under Bun on macOS, `sqlite-vec` never loaded, so memory search silently ran the batched embedding scan (after #141104) or failed outright (before it). Bun dlopens Apple's system SQLite so exactly one SQLite library lives in the process, and Apple's build omits extension loading. Our docs told operators to preload a custom SQLite themselves, which nobody does, and even a parent-process preload never reached the KNN child.

## Why This Change Was Made

Bun's only hook is `bun:sqlite` `Database.setCustomSQLite(path)`. It is one-shot, must run before the first database open, and a bad path breaks every later open in the process. This change turns that fragile hook into safe automatic behavior owned by the SQLite runtime boundary.

- New `src/infra/bun-sqlite-library.ts` validates each candidate through `bun:ffi` (`sqlite3_libversion` passes the WAL-reset safety floor, `OMIT_LOAD_EXTENSION` absent) before committing, then memoizes process-wide. Resolution order: explicit path, then `OPENCLAW_SQLITE_LIBRARY`, then discovery under `$HOMEBREW_PREFIX`, `/opt/homebrew`, `/usr/local`, and MacPorts. A capable candidate wins on capability, not raw version: Homebrew 3.53.4 beats Apple 3.54.0 because Apple's cannot load extensions.
- Selection runs at both first-open owners: the startup runtime guard probe and `requireNodeSqlite()`. An unusable override is reported as one clean runtime-guard diagnostic with exit 1.
- The memory KNN child receives the selected library through its typed stdin input and selects before opening. `buildChildEnv` stays stripped. The #141104 scan fallback remains the path when no capable library exists.
- Gateway startup logs the selected library; doctor reports it for the doctor process.
- Node, and Bun on Linux and Windows, are intentional no-ops. Bun statically links its own SQLite 3.53.2 with upstream defaults there, so extension loading already works; `setCustomSQLite` is a documented no-op on those platforms.
- The undocumented legacy alias `OPENCLAW_CLAUDE_CLI_LOG_OUTPUT` (superseded by `OPENCLAW_CLI_BACKEND_LOG_OUTPUT` in the March 2026 CLI-backend rename, present in no docs) is retired so the `OPENCLAW_*` name budget stays at 493. `docs/help/environment.md` states undocumented names are internal and may disappear.

No configuration option, schema, persistent-store, WAL floor, or child-environment change. `src/daemon/runtime-paths.ts`'s raw runtime probe is unchanged: it checks the runtime default's WAL floor, which selection never lowers.

## User Impact

On macOS Bun, `brew install sqlite` is now all that is needed for native `sqlite-vec` KNN memory search; OpenClaw picks the library up automatically in the Gateway and in the KNN child. Operators can pin a library with `OPENCLAW_SQLITE_LIBRARY`; an invalid value fails loudly with the fix. Without a capable library, behavior is unchanged from #141104.

Existing custom SQLite preload scripts must be removed and replaced with `OPENCLAW_SQLITE_LIBRARY` pointing to the same library. The Bun install guide documents this migration because Bun allows the selector hook only once per process.

## Evidence

- The KNN barrel export follows main's narrow child import boundary; the parent and subprocess mock use the same selector import. SDK registration and surface checks pass.

- Live on macOS with Bun 1.4.2 and Homebrew SQLite 3.53.4: discovery selects the Homebrew library, `sqlite-vec` v0.1.9 loads, and a real `runVectorKnnInSubprocess` round trip returns both synthetic rows with `fallbackScanRequired: false`.
- Built CLI: `bun openclaw.mjs --version` unchanged (metadata fast path); `OPENCLAW_SQLITE_LIBRARY=/nonexistent.dylib bun openclaw.mjs status` prints one diagnostic block (`Cannot use SQLite library /nonexistent.dylib: missing file. Fix or unset OPENCLAW_SQLITE_LIBRARY; install a supported library with brew install sqlite.` plus the Detected line) and exits 1; the same override under Node is ignored and `--version` succeeds.
- Linux verified on a Blacksmith Testbox lease (`tbx_01m1yd3xnasmcm5h2qq3rec2kg`, x64, fresh Bun 1.4.2): `node:sqlite` reports SQLite 3.53.2, `OMIT_LOAD_EXTENSION` = 0, `enableLoadExtension(true)` works, and `setCustomSQLite` is a no-op. Windows shares Bun's static build path.
- Tests: selector resolution, validation, memoization, one-shot failure, and platform no-op cases in `src/infra/node-sqlite.test.ts` (consolidated to respect the 720-root infra shard limit); runtime-guard diagnostic path; KNN subprocess forwarding for runtime/env/discovered selections with the stripped child env asserted; 83 infra + 10 subprocess + 26 orchestration tests pass on the rebased head.
- `pnpm build` passes; the bundled chunk keeps `bun:ffi` and `bun:sqlite` as runtime `createRequire` calls with no inlining. Import cycles 0. `OPENCLAW_*` count 493/493. `git diff --check` clean. Independent review reported no actionable findings.
- `check:changed` passed every gate up to the known nested-worktree extension declaration lint boundary; the changed core files pass the core lint configuration directly. CI is authoritative for that lane.

- Preload migration verified on Bun 1.4.2: a legacy `--preload` script calling `Database.setCustomSQLite()` conflicts with automatic selection (`SQLite already loaded`), even for the same Homebrew path. Removing that preload and setting `OPENCLAW_SQLITE_LIBRARY` to the same path makes the built CLI `memory status --json` exit 0 in an isolated state directory; the real parent/child KNN round trip returns both synthetic rows with `fallbackScanRequired: false`.
- Initial hosted CI found test formatting and an unbound `stdin.end` method reference. A test-only follow-up retains the spy handle directly; all 10 subprocess tests, formatting, and fresh independent review pass.

- Review resolution: the supported upgrade path for custom SQLite preload users is the documented and live-verified preload-to-environment transition. The data-model heuristic does not identify a persisted-format change: `sqliteLibraryPath` exists only in the bounded per-query stdin message, and the doctor addition reports the selected runtime library. Database paths, schemas, stored vectors, and existing fallback search remain unchanged, so no database migration or backfill is required.

### DIFF
--- a/docs/help/environment.md
+++ b/docs/help/environment.md
@@ -28,15 +28,16 @@ The variables below are the supported environment contract for operators. Undocu
 
 ### Paths and instances
 
-| Variable                 | Purpose                                                           |
-| ------------------------ | ----------------------------------------------------------------- |
-| `OPENCLAW_HOME`          | Override the home directory used for OpenClaw path defaults.      |
-| `OPENCLAW_STATE_DIR`     | Override the mutable state directory.                             |
-| `OPENCLAW_CONFIG_PATH`   | Override the active config file path.                             |
-| `OPENCLAW_WORKSPACE_DIR` | Override the default agent workspace.                             |
-| `OPENCLAW_PROFILE`       | Select a named profile and its isolated defaults.                 |
-| `OPENCLAW_GIT_DIR`       | Override the source checkout used by development-channel updates. |
-| `OPENCLAW_INCLUDE_ROOTS` | Allow `$include` to resolve from additional roots.                |
+| Variable                  | Purpose                                                               |
+| ------------------------- | --------------------------------------------------------------------- |
+| `OPENCLAW_HOME`           | Override the home directory used for OpenClaw path defaults.          |
+| `OPENCLAW_STATE_DIR`      | Override the mutable state directory.                                 |
+| `OPENCLAW_CONFIG_PATH`    | Override the active config file path.                                 |
+| `OPENCLAW_WORKSPACE_DIR`  | Override the default agent workspace.                                 |
+| `OPENCLAW_PROFILE`        | Select a named profile and its isolated defaults.                     |
+| `OPENCLAW_GIT_DIR`        | Override the source checkout used by development-channel updates.     |
+| `OPENCLAW_INCLUDE_ROOTS`  | Allow `$include` to resolve from additional roots.                    |
+| `OPENCLAW_SQLITE_LIBRARY` | Override the SQLite library for [Bun on macOS](/install/bun#caveats). |
 
 ### Gateway and authentication
 
--- a/docs/install/bun.md
+++ b/docs/install/bun.md
@@ -95,6 +95,13 @@ floor and supports extension loading. An invalid override fails with an error
 explaining how to fix or unset it. Node and Bun on platforms other than macOS
 ignore this override; Gateway startup logs a warning when it is ignored.
 
+If you previously used a Bun preload script that calls `Database.setCustomSQLite()`,
+remove that preload and set `OPENCLAW_SQLITE_LIBRARY` to the same library path
+instead. Bun permits that hook only once per process; keeping the preload causes
+startup to fail with `SQLite already loaded`, even when both selections name the
+same library. The environment override lets OpenClaw select the library before
+opening databases and forward the path to the memory KNN child.
+
 Without a suitable library on macOS, Bun keeps Apple's system SQLite, which omits
 native extension loading. OpenClaw can open ordinary agent databases when that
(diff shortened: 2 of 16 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1c1dcb247e8898352ff578795e7e175417e2d65c](https://github.com/openclaw/openclaw/pull/141854/commits/1c1dcb247e8898352ff578795e7e175417e2d65c)
- Landed commit: [757e482ab646421f45b3eb9c069beb3eb1f7027c](https://github.com/openclaw/openclaw/commit/757e482ab646421f45b3eb9c069beb3eb1f7027c)

PR #141723, opened 2026-09-08
fix(auth): reconcile Gateway secret guidance and macOS password prompts

steipete:
Related: #141456, #141514, #141511

## What Problem This Solves

Fixes an issue where users following Gateway setup and connection guidance would be told to choose token versus password or use separate fields after onboarding and Control UI moved to one Gateway secret.

## Why This Change Was Made

Align protocol, setup, Control UI, remote-client, and schema help with the existing contract: the Gateway auth mode selects its configured secret, and clients may send that secret in either connect field. First-run onboarding generates a secret by default, while explicit password flags, existing password configurations, and Funnel's password requirement remain documented. Targeted local configure still has its auth-mode picker.

Rotation guidance distinguishes same-mode hot reload from mode changes or process-environment updates requiring restart. Config baselines were regenerated with their owning tool.

Gateway configuration fields, defaults, protocol, and storage contracts are unchanged; no configuration or data migration is required. The baseline hashes changed because help text changed, while the config entry counts remain identical.

## User Impact

Users can follow consistent setup and connection instructions and rotate credentials with accurate restart guidance. Control UI storage guidance preserves the rule that only successful token-mode connections remember the secret per tab and Gateway; passwords remain memory-only.

## Evidence

- Source-checked the Gateway's shared-secret selection, onboarding, local configure, remote configuration, and Control UI persistence paths.
- `git diff --check` and `pnpm docs:list` passed.
- `pnpm config:docs:gen` / `pnpm config:docs:check` passed.
- Codex autoreview of the original patch and the test-selector follow-up: `scoped-clean`, “patch is correct,” no accepted/actionable P0–P2 findings.
- Searched removed copy across `ui/src`, `extensions`, `apps`, `src`, and `docs`; no selector changes or stale assertions for the removed unsupported-auth text.
- Before the context-only rebase, `node scripts/check-changed.mjs` passed for the core/docs changes, including config-baseline, typechecking, lint, and guards.
- [Initial CI](https://github.com/openclaw/openclaw/actions/runs/34175136437/job/101903026942) caught two stale config-help test selectors. The baseline test now requires both configured-secret selection and either-field guidance; the remote-token quality selector requires the full storage-and-rotation instruction. All keys, minimum lengths, and other selectors are preserved. Both affected files pass locally: 34 tests.
- The rebase retained main's new invalid-secret guidance and setup-code hints while preserving every added and removed line of this PR. The config baseline and all 34 focused tests passed again afterward.

[...]

PR #141857, opened 2026-09-08
fix(macos): let remote onboarding accept password-mode Gateways

steipete:
## What Problem This Solves

Fixes an issue where users connecting the macOS app to a password-mode Gateway would reach an unsupported-auth message with no credential field. The app asked them to reconfigure the Gateway for token auth even though the Gateway already accepts its configured secret from either connect field.

## Why This Change Was Made

Password-required errors now expose the existing Gateway token field and explain that the Gateway password goes there. The banner, detailed guidance, short status, secure-field caption, and generated native English source inventory agree. The existing onboarding field-visibility test includes the password-required case. This uses the Gateway authentication behavior merged in #141456; no protocol or Gateway changes are needed.

## User Impact

Users can enter their Gateway password and retry remote onboarding without switching the host to token auth. If no password is configured on the host, the prompt identifies the existing configuration and environment keys needed to set one. Password entry through this field requires a Gateway build containing #141456; older Gateways must be updated. Compatibility with older Gateways that lack that change is not claimed.

## Evidence

- Native test build passed: `swift build --package-path apps/macos --build-system native --build-tests`, after `node scripts/prepare-apple-mermaid.mjs` generated the required local resources.
- All 10 `OnboardingRemoteAuthPromptTests` passed locally. The built test bundle ran through the SwiftPM testing helper with `--testing-library swift-testing --filter OnboardingRemoteAuthPromptTests` inside a verified OS sandbox blocking operator files, preferences/Keychain services, network, and helper processes.
- `pnpm native:i18n:baseline` regenerated the inventory after the field-caption review fix; `pnpm native:i18n:verify` then passed with `changed=false`.
- `node scripts/check-changed.mjs` passed on the original three-file candidate, including Swift lint, selected macOS script tests, and the native schema guard. The final run used system Bash first on PATH and the repository-pinned Swift tools; an earlier run was interrupted after Homebrew Bash stalled on a packaging-test heredoc.
- Final caption follow-up: native test rebuild, all 10 filtered onboarding tests, and scoped macOS Swift lint passed.
- Codex autoreview against the branch's merge base found no actionable P0–P2 findings; the complete final candidate was reviewed again after the caption correction.
- **Proof gap:** no before/after app captures. Screen capture on the maintainer's Mac is currently blocked (the Peekaboo host reports Screen Recording as not granted despite the System Settings toggle; being debugged separately). The change is copy plus one enum-case switch, so the unit test is the behavioral proof.
- The full native suite and live Gateway onboarding were not run locally; exact-head macOS CI supplies full-suite proof. The regression was not executed against pre-fix code.

Related: #141456 and #141723. The macOS portion was split out of #141723 because captures could not be produced.


steipete:
Merged via squash.

- Prepared head SHA: [15aa8da43f187464c59c5b1d21bede3c0d263d1c](https://github.com/openclaw/openclaw/pull/141857/commits/15aa8da43f187464c59c5b1d21bede3c0d263d1c)
- Landed commit: [74d3f22371209c30756b6ca6d4f6bc6c3bd89a45](https://github.com/openclaw/openclaw/commit/74d3f22371209c30756b6ca6d4f6bc6c3bd89a45)

PR #141858, opened 2026-09-08
fix(workboard): release stores from owned plugin inspections

steipete:
## What Problem This Solves

Fixes Workboard database connections remaining open after an owned plugin inspection. Registration could also fail after opening the store but before registering its cleanup.

## Why This Change Was Made

Register store cleanup immediately after acquisition and add a resource-only disposer using the existing close path. Track each successfully constructed service so disposal stops those producers before the existing store owner drains admitted work and closes its connection.

Related: #140674. This three-file slice uses the landed inspection owner and leaves general cached/prepared registry lifetimes unchanged.

## User Impact

Temporary inspections close their own Workboard connection while active registrations remain usable. Existing restart/disable and session/run cleanup rules are preserved, as are stored cards, schemas, queries, and transaction behavior.

## Evidence

Two intended regressions failed before production edits. All 82 entry/store/service tests, canonical changed checks, full build, and Codex review passed. Build time was 268.93 seconds.

An instrumented integration run used the actual compiled loader, bundled Workboard, and native SQLite with synthetic host context. It proved inspection release waits for admitted work, partial registration rollback joins cleanup, repeated release closes once, and the active sibling stays readable/writable with its change service intact. Each of three distinct Workboard connections closed once at its own retirement; three cards remained on reopen. This run took 2.267 seconds. The observer preserves native constructor argument arity; an earlier observer-only failure is retained separately.

A separate uninstrumented shipped `node openclaw.mjs workboard list --json` process read all three cards from that same final synthetic state in 3.218 seconds. That is compatibility and durable-data readback proof; deterministic disposal is established by the instrumented loader run, not by CLI process exit.

Source remained unchanged throughout validation. Production delta is +16/-8 lines; existing entry tests add the regression coverage. No new SDK export, configuration option, database schema, or persistence policy.

### DIFF
--- a/extensions/workboard/index.lifecycle.test.ts
+++ b/extensions/workboard/index.lifecycle.test.ts
@@ -63,6 +63,11 @@ function registerGeneration(register: (api: OpenClawPluginApi) => void = plugin.
       await lifecycle.cleanup?.(context);
     }
   };
+  const dispose = async () => {
+    for (const lifecycle of captured.runtimeLifecycles) {
+      await lifecycle.dispose?.();
+    }
+  };
   const run = async (...args: string[]): Promise<unknown> => {
     const program = new Command().exitOverride();
     for (const registration of captured.cliRegistrars) {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ff1c541af68d658388a07fa6c6986201cad6aa53](https://github.com/openclaw/openclaw/pull/141858/commits/ff1c541af68d658388a07fa6c6986201cad6aa53)
- Landed commit: [b510ebd80c5098eced273a1ed21167cd9eb966a1](https://github.com/openclaw/openclaw/commit/b510ebd80c5098eced273a1ed21167cd9eb966a1)

PR #141859, opened 2026-09-08
fix(cli): honor the selected agent when inspecting models

obviyus:
## What Problem This Solves

Fixes an issue where `openclaw infer model --agent work list --json` tells operators that no agent was selected in a multi-agent configuration. Model inspection and case-only model-name normalization also lose the selected agent when reading the catalog.

## Why This Change Was Made

Pass the inherited agent option through the existing catalog reader and validate explicit agent IDs with the existing CLI owner resolver. Model runs keep their already-resolved agent during catalog normalization. Read-only acquisition, omitted-agent behavior and aggregate inventory remain unchanged.

## User Impact

Operators can list and inspect models for a selected agent without configuring a separate system agent. An unknown agent is rejected instead of reading another agent's catalog.

## Evidence

- Real CLI baseline: `infer model --agent work list --json` exits 1 with “this operation has no explicit owner” despite the supplied selector.
- With this change, the same command exits 0. `infer model --agent work inspect --model catalog-fixture/work-model --json` returns the expected model with a 32768-token context window and text input. An unknown `retired` agent exits 1 with the correct diagnostic.
- Five added regression cases fail before the fix. All 14 selected regression and neighboring checks pass after it, including local/Gateway case normalization and auth-profile preservation. These dispatch checks use the existing test doubles; no live inference is claimed.
- Pinned formatting and `git diff --check` pass. Public command checks use isolated synthetic state, no credentials and no external network.

### DIFF
--- a/src/cli/capability-cli.test.ts
+++ b/src/cli/capability-cli.test.ts
@@ -104,7 +104,9 @@ const mocks = vi.hoisted(() => ({
   resolveMemorySearchConfig: vi.fn<
     typeof import("../agents/memory-search.js").resolveMemorySearchConfig
   >(() => null),
-  loadModelCatalog: vi.fn(async () => []),
+  loadModelCatalog: vi.fn<
+    typeof import("../agents/prepared-model-catalog.js").loadPreparedModelCatalog
+  >(async () => []),
   prepareSimpleCompletionModelForAgent: vi.fn(async () => ({
     selection: {
       provider: "openai",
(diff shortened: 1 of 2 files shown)


obviyus:
Landed the selected-agent catalog repair.

**Proof:** Real isolated CLI runs confirm that explicit-agent list and inspect now succeed, and an unknown agent is rejected. All 14 focused regression and neighboring checks pass. Exact-head CI run [34188746737](https://github.com/openclaw/openclaw/actions/runs/34188746737) passed; the fresh independent local review found no code or security issue.

| Before | After |
| --- | --- |
| `infer model --agent work list --json` reports no explicit owner. | The same command succeeds without a separate system agent. |
| Inspection and case-only normalization lose the selected catalog owner. | Both preserve the selected agent through the existing read-only catalog path. |

Aggregate inventory and omitted-agent behavior are preserved. Normalization has dispatch-test coverage; live inference is not claimed.

Issue #141838, opened 2026-09-08
[Bug]: configured model fallback is skipped silently when a run ends on the agent deadline or the idle-timeout breaker


von8794:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

A configured model fallback is never attempted, and nothing is logged about the chain stopping, when a run ends via the whole-run agent deadline or the idle-timeout cost-runaway breaker — both of which unwind outside the model-fallback layer.

### Steps to reproduce

Deterministic, with a local stub provider and no real API. The stub accepts connections and never responds.

```js
// 127.0.0.1:8899 — accepts and never responds
import http from "node:http";
http.createServer(() => {}).listen(8899, "127.0.0.1");
```

Provider config declares `hangprov` (`openai-completions`, `baseUrl http://127.0.0.1:8899/v1`) with two models, `hang-primary` and `hang-fallback`.

**Scenario A — whole-run agent deadline:**

1. `openclaw agent exec "say hi" --model hangprov/hang-primary --fallback hangprov/hang-fallback --timeout 25 --json`
2. Observe the run end on the agent deadline.
3. Observe that `hang-fallback` is never requested and that no `[model-fallback/decision]` line is emitted.

**Scenario B — idle-timeout circuit breaker:**

1. Add `timeoutSeconds: 8` to the provider so each request idles out well inside the agent deadline.
2. `openclaw agent exec "say hi" --model hangprov/hang-primary --fallback hangprov/hang-fallback --timeout 120 --json`
3. Observe 5 consecutive idle timeouts trip the breaker.
4. Observe that `hang-fallback` is never requested and that no `[model-fallback/decision]` line is emitted.

### Expected behavior

Grounded in the same harness on the same build: when the primary fails with a *retryable* error (stub returning HTTP 500 instead of hanging), the chain does advance and every hop is logged.

```text
[model-fallback/decision] decision=candidate_failed candidate=hangprov/hang-primary  reason=timeout next=hangprov/hang-fallback detail=500 upstream exploded
[model-fallback/decision] decision=candidate_failed candidate=hangprov/hang-fallback reason=timeout next=none                 detail=500 upstream exploded
```

Expected in A and B: either the configured fallback is attempted, or the run records why the chain stopped without trying it.

### Actual behavior

In both scenarios exactly one model is attempted and the chain stops with no fallback diagnostic at all.

Scenario A — only `hang-primary` appears; `[model-fallback/decision]` count is 0:

[...]

Issue #42713, opened 2026-03-11
Failover wrapper errors missing provider and session_id in error logging


theo674:
## Problem

When the gateway's failover system exhausts all auth profiles and model fallbacks, the resulting error log entries are missing `provider` and `session_id` metadata. This makes it difficult to attribute errors during post-mortem analysis.

## Evidence

Analyzing the last 100 errors from our `api_health_log` table (structured ingestion from `gateway.err.log`):

- **18 of 100 errors** (18%) had no `provider` or `session_id` populated
- These are all failover/wrapper errors — the final `FailoverError` after all profiles/models have been tried
- The underlying per-profile errors DO have provider attribution, but the wrapper error that gets logged does not carry it through

## Expected behavior

Failover wrapper errors should include:
1. The **provider(s) attempted** (even if multiple)
2. The **session_id / lane** that originated the request
3. Consistent `session_id` format (currently we see both `session:browser-XXXX` and `browser-XXXX` for the same session)

## Impact

When triaging rate limit storms or auth failures, the missing attribution means you have to manually correlate wrapper errors with nearby per-profile errors by timestamp. At scale (we had 6,000+ rate limit errors in 12 days), this makes automated analysis unreliable.

## Environment

- OpenClaw version: 2026.3.7
- Providers: anthropic (token), openai-codex (OAuth)
- Config: 3 anthropic profiles, 1 openai-codex profile, 2 model fallbacks

## Suggestion

In the `FailoverError` constructor or the error logging path, propagate the `provider` from the last-attempted profile and the `sessionId` / `lane` from the originating request. For the session_id normalization, ensure the `session:` prefix is consistently applied.

PR #141869, opened 2026-09-08
fix: configured model fallbacks are skipped when the primary model times out

von8794:
Related: #141838

## What Problem This Solves

Fixes an issue where an operator who has configured model fallbacks gets no
reply at all — no answer, no error — when the primary model times out.

The configured chain is never walked past the first hop. A run that should have
fallen back to a second or third model instead ends in silence, and the local
tail an operator configured as their last resort is unreachable by construction
whenever the primary times out.

Observed on a live gateway (2026-09-07), four times in fourteen minutes:

```
18:26:09  turn failed model=claude-opus-5   durationMs=1320017 error=FailoverError
18:26:10  [model-fallback/decision] candidate_failed opus reason=timeout next=claude-sonnet-5
18:27:11  turn failed model=claude-sonnet-5 durationMs=60537   error=TimeoutError
          (no decision line; chain stops; the configured third candidate is never attempted)
```

`agents.defaults.timeoutSeconds` was 1320. The primary always died at exactly
1320s. The second candidate then died after 60.5s / 79.9s / 102.9s across
repeats — not a timer, just the leftover abort grace.

## Why This Change Was Made

A run's whole-run deadline is sized for **one** attempt.
`registerChatAbortController` sets `expiresAtMs` to `timeoutMs` plus a 60s abort
grace, and `resolveAgentTimeoutMs` feeds that same `timeoutMs` to the
per-attempt timer. So a primary candidate that times out has already spent the
entire run budget, and every later candidate inherits only the grace window.
`runGatewayMaintenance` then sweeps the entry and aborts the controller with
`createChatAbortSignalReason("timeout")` — an external, bare `TimeoutError` —
which stops the chain without a decision line.

`markExecutionStarted` refreshes that deadline exactly once, and returns early
for `kind !== "agent"` (chat-abort.ts), so a `chat-send` run never gets even
that one refresh.

This renews the deadline as each attempt actually starts, in the attempt layer
next to the existing "later attempts must not start after an abort" guard. The
run owner registers a renewer keyed by run id; a run with no registered owner
(CLI, cron, unit tests) keeps today's behavior exactly.

Deliberate boundaries:

- **Abort attribution is untouched.** This changes *when* the run deadline
  fires, never *how* an abort is classified. No candidate is blamed for a
  local or terminal abort.
- **The once-only guard in `markExecutionStarted` is untouched.** Renewal is a
  separate entry point, so `session-active-runs.test.ts` still pins
  `true` then `false`.
- Renewal only ever **extends**, never shortens, and declines for a
  registration that is missing, superseded, aborted, not yet executing, or
  already past its deadline. An entry the sweep has condemned is never revived.
- The deadline resolvers move to `src/gateway/chat-run-deadline.ts` alongside
  the renewal, because `chat-abort.ts` is at its 700-line cap.
  `chat-abort.ts` re-exports them, so no existing import path changes.

**The tradeoff, stated plainly:** worst-case run wall clock becomes
`attempts x timeoutSeconds + grace` instead of `timeoutSeconds + grace`. That is
the point — the configured chain cannot otherwise complete — but it does change
what `timeoutSeconds` bounds, so it is a judgement call I would rather you make
explicitly than discover. No new config knob is added; `timeoutSeconds: 0`
already exists for callers who want no run budget at all. Stall detection is
unaffected: `resolveCliNoOutputTimeoutMs` is clamped independently
(`CLI_FRESH_WATCHDOG_DEFAULTS` maxMs 600s), so a genuinely hung attempt is still
killed on the same schedule and still raises a `FailoverError` that walks the
chain normally.

## User Impact

An operator whose primary model times out now actually reaches their configured
fallbacks, including the local model many people configure as the last hop.
Before this, a timed-out primary meant the run died silently at the second
candidate. Nothing changes for runs that succeed, for runs with no fallbacks
configured, or for CLI and cron runs.

## Evidence

**New regression test, chain level, fails without the fix.**
`src/agents/model-fallback.run-deadline.test.ts` registers a real
`kind: "chat-send"` controller admitted a full `timeoutMs` ago — the exact state
after a primary burns the budget — then runs a two-candidate chain and measures
the budget the fallback candidate actually receives.

With the renewal removed (`renewAgentRunDeadline` call neutralized, everything
else identical):

```
 × src/agents/model-fallback.run-deadline.test.ts > gives a fallback candidate
   its own run budget after the primary spent it
   → expected 59892 to be greater than 60000
```

59.9s is the leftover grace. With the fix, the candidate gets a full fresh
budget and both tests pass:

```
 ✓ gives a fallback candidate its own run budget after the primary spent it
 ✓ leaves runs without a registered owner on their existing deadline
 Test Files  1 passed (1)   Tests  2 passed (2)
```

**Renewal guards** — `src/gateway/chat-run-deadline.test.ts`, 6 cases: extends
an executing run, declines before execution starts, declines once aborted,
never revives a condemned entry, never shortens, declines for a superseded or
removed registration. Run together with the existing `chat-abort.test.ts`:
`Test Files 2 passed (2), Tests 56 passed (56)`.

**Existing behavior that had to stay pinned:**

- `session-active-runs.test.ts` + `sessions.abort-embedded.test.ts`, unmodified:
  `36 passed`.
- `model-fallback.test.ts`, `model-fallback.terminal-boundary.test.ts`,
  `model-fallback.probe.test.ts`, `model-fallback-observation.test.ts`,
  unmodified: `Test Files 4 passed (4), Tests 228 passed (228)`.

**Repo gates:** `pnpm check` exit 0 (includes typecheck, oxlint, oxfmt, and the
import-cycle guard, which reports `0 runtime value cycle(s)`), `pnpm build`
exit 0.

**Unit lane:** `13101 passed | 2 failed`. The two failures are
`src/state/openclaw-state-db.permissions.test.ts` ("rethrows EPERM/EROFS when
existing permissions are too broad"). I checked out clean upstream with this
branch's code absent and re-ran that file: the same two fail there
(`Tests 2 failed | 11 passed`). They are pre-existing on this machine and
unrelated to this change.

**What I could not run:** the full `pnpm test` on this host. It forks up to ten
concurrent Vitest processes and was SIGKILLed (exit 137). Single-lane
single-process runs are what is reported above; I am not implying wider coverage
than that.

### After-fix runtime capture, with a matched control

Re-review asked for real behavior rather than mocked candidate execution. Below
is a runtime capture of the same branch at `cc8a7e96`, node v24.20.0, taken in an
isolated dev sandbox: no live Gateway, no provider account, no API key.

**Real in this capture:** `registerChatAbortController` /
`removeChatAbortControllerEntry`, `startGatewayMaintenanceTimers` (the actual
expiry sweep, on its real 60s cadence), `runWithModelFallback` /
`runFallbackAttempt`, `registerAgentRunDeadlineRenewer` / `renewAgentRunDeadline`,
plus real sockets, real elapsed wall time and real `AbortSignal` propagation.

**Substituted:** the agent-turn executor. The `run` callback performs the provider
HTTP call itself against a local stub instead of building a full agent turn.
Candidate execution is real I/O with real cancellation; it is not a full gateway
turn, and I am not claiming it is.

Setup: run kind `chat-send` (the path that gets no deadline refresh at execution
start), `timeoutMs` 120000 so the whole-run deadline lands at t+180s, a primary
that hangs until its 120s attempt timeout, and a fallback that needs 100s of work
— more than the 60s abort grace.

**Control** (renewal hook removed, everything else identical):

```
[proof +0.0s]   mode=CONTROL (renewal neutralised)
[proof +0.0s]   registered kind=chat-send; run deadline is 180.0s away
[proof +0.6s]   ATTEMPT stubprimary/hang-primary starting; run deadline 179.4s away
[proof +120.7s] ATTEMPT stubprimary/hang-primary TIMED OUT after 120000ms
[model-fallback/decision] decision=candidate_failed candidate=stubprimary/hang-primary reason=timeout next=stubfallback/ok-fallback
[proof +120.7s] ATTEMPT stubfallback/ok-fallback starting; run deadline 59.3s away
[proof +180.1s] RUN ABORTED by owner: TimeoutError: chat run timed out
[proof +180.1s] ATTEMPT stubfallback/ok-fallback FAILED: TimeoutError: chat run timed out
[proof +180.1s] THREW TimeoutError: chat run timed out
```

Effective cancellation, from the stub provider's own log in that same run:

```
[ok-fallback] request /ok
[ok-fallback] CLIENT CANCELLED /ok
[ok-fallback] connection closed before reply /ok
```

The fallback entered with **59.3s** — the abort grace and nothing else, the same
leftover the unit test pins at 59892ms.

**Fixed** (this PR, unmodified):

```
[proof +0.0s]   mode=FIXED (renewal active)
[proof +0.0s]   registered kind=chat-send; run deadline is 180.0s away
[proof +0.7s]   ATTEMPT stubprimary/hang-primary starting; run deadline 180.0s away
[proof +120.7s] ATTEMPT stubprimary/hang-primary TIMED OUT after 120000ms
[model-fallback/decision] decision=candidate_failed candidate=stubprimary/hang-primary reason=timeout next=stubfallback/ok-fallback
[proof +120.8s] ATTEMPT stubfallback/ok-fallback starting; run deadline 180.0s away
[proof +220.8s] ATTEMPT stubfallback/ok-fallback SUCCEEDED
[model-fallback/decision] decision=candidate_succeeded candidate=stubfallback/ok-fallback next=none
[proof +220.8s] RESULT outcome=completed
```

Identical primary timeout at the same moment; the fallback starts with a full
180s instead of 59.3s and completes at +220.8s, which the control never reaches.
Both captures are unedited apart from trimming the sandbox launcher's banner
lines. The 09-07 excerpt at the top of this PR remains the original pre-fix
failure.


von8794:
Two implementation notes a reviewer would otherwise have to ask about.

**Branch base.** This targets `de9fce557` rather than current `main`. Pushing a branch based on a newer commit is rejected by my fork with `Unable to determine if workflow can be created or updated due to timeout; \`workflows\` scope may be required.`, and `merge-upstream` fails the same way — my token has no `workflow` scope. Rebasing onto a commit the fork already holds was the way through. The three intervening commits are unrelated to the touched files, and every gate reported in the PR body was re-run on this exact base. **Allow edits from maintainers** is on, so rebase away if you prefer the tip.

**Why `runId` rides on `runAttribution`.** `src/agents/model-fallback-runner.ts` is at exactly the 700-line `max-lines` cap, so it cannot take a single added line — that is why the run id travels on the existing `runAttribution` object rather than as a new argument to `runFallbackAttempt`. `runAttribution` also spreads into two `FailoverError` constructors; the constructor ignores the extra key, so no error shape changes. If you would rather `FailoverError` carried `runId` alongside `sessionId`/`lane` (it already documents propagating attribution for log ingestion, #42713), say so and I will add it.

Same reason `chat-abort.ts` gave up its deadline resolvers to a new file: it was over the cap once the renewal landed. The re-export keeps every existing import path working.

von8794:
Pushed three commits to clear the four red checks. No change to the fix itself: abort attribution and the once-only `markExecutionStarted` guard are still untouched.

**`check-additional-runtime-topology-architecture`** — `check:madge-import-cycles` found `chat-abort.ts -> chat-run-deadline.ts -> chat-abort.ts`. The back-edge was a single type-only import, which is why `check:import-cycles` still reported 0 runtime value cycles; only the type graph was circular. `renewChatRunExecutionDeadline` reads `controller` and `executionStarted` and writes `expiresAtMs`, so I declared exactly that shape locally as `ChatRunDeadlineEntry` and dropped the import. `ReadonlyMap` now also states that the entry map is never structurally mutated there. No new file.

**`check-dependencies`** — knip flagged `resetAgentRunDeadlineRenewersForTest` as an unused export, correctly: it runs `--production` and only a test imported it. It turned out to be unnecessary. `removeChatAbortControllerEntry` already calls `unregisterAgentRunDeadlineRenewer`, so the test retires its registration through the same path the gateway uses, from `afterEach` — which also makes cleanup survive a failing assertion, unlike the inline reset it replaces. The export is gone rather than exempted.

**`check-additional-boundaries`** — this one is worth flagging beyond this PR. The failing sub-check was `lint:tmp:tsgo-core-boundary`: `gateway-root: 721 test roots exceeds the 720 limit`. **`gateway-root` is at exactly 720 on current main**, so any PR that adds a new `src/gateway` test file fails this gate. Verified by moving my file aside on this branch: the check passes at 720 and fails at 721. I did not touch `TSGO_CORE_TEST_MAX_ROOTS` or the shard table, since rebalancing shared CI infra is your call, not something to slip into a bug fix. Instead the six renewal cases moved into `chat-abort-lifecycle-internal.test.ts`, which already covers registration lifecycle and already imports `registerChatAbortController`, `removeChatAbortControllerEntry` and `ChatAbortControllerEntry`. I chose it over `chat-abort.test.ts` because that file is near its own 1000-line `max-lines` cap — folding there tripped it at 1077.

**`openclaw/ci-gate`** was the aggregate of the above.

Verified locally on node 24.20.0 before pushing, all exit 0: `pnpm lint`, `pnpm build`, `pnpm deadcode:full`, `pnpm check:architecture`, `pnpm run lint:tmp:tsgo-core-boundary`, and typecheck (prod, scripts, test root). Tests: 70 passed across `chat-abort-lifecycle-internal.test.ts` and `chat-abort.test.ts`, 2 passed in `model-fallback.run-deadline.test.ts`.

Still not re-run here: the full `pnpm test` fork-parallel lane, for the same memory reason noted earlier in this PR.

von8794:
Pushed `f75619f2f`. It resolves the two review items that did not need a maintainer, and leaves the one that does.

**Renewal is now recovery-only.** The call is gated on `attempt > 1`, so the primary candidate never extends the deadline. A configuration with no fallback candidates, or one whose primary succeeds, keeps exactly the whole-run budget its operator configured, unchanged from current main. The review's lead evidence — "the pinned introduced patch calls renewal before every candidate, including the primary" — no longer describes the branch.

**That also closes the regression-test hole**, and not only by assertion. With the primary excluded from renewal, "renewing only the primary would still satisfy the fallback-budget assertion" is no longer expressible. The test now also records the deadline inside the primary callback and asserts it is identical to the admission value, then asserts the fallback candidate sees more than a whole timeout remaining. Written failing first: `expected 1788909542268 to be 1788909482056` before the gate, green after.

Verified on Node 24.20.0: 2 passed in `model-fallback.run-deadline.test.ts`, 313 passed across the 11 `model-fallback` / `chat-abort` / `session-active-runs` suites, `pnpm build` exit 0, `pnpm check` exit 0.

**On the decision you asked for, with the trade-off stated plainly rather than argued.**

The narrowing shrinks the compatibility question but does not remove it. Worst case is still `attempts x (timeout + grace)` of wall clock, now reachable only by a run that actually enters recovery.

The reason I have not simply implemented option 1 as written is that "bound candidate recovery within the existing budget" and "a configured fallback chain works after a primary timeout" cannot both be true. `timeoutMs` is the per-attempt provider timeout *and* the whole-run bound. When the primary consumes it, there is no remaining budget to fit a candidate inside, so the honest behavior under a preserved whole-run bound is that the configured tail is never tried — the same user-visible outcome as today, now merely explained by the diagnostics merged in #142278.

So it is a real choice, not a formality:

1. **Merge this as recovery-only renewal.** Configured fallbacks work after a primary timeout. Finite-timeout configurations that reach recovery can run longer than their stored value.
2. **I rework it to clamp instead of renew.** Each recovery attempt's own timeout is capped at the remaining run budget, the absolute deadline is never moved, and a candidate with no room fails fast and attributably instead of being swept by the maintenance timer. Documented semantics are preserved exactly; fallback after a full-budget primary timeout stays impossible by design.
3. **Neither, without a separate opt-in setting.** I have deliberately not invented a config flag here, since that answers the policy question rather than asking it.

Say which and I will implement it. I have no stake in the outcome beyond the chain no longer failing silently, and #142278 already delivered the silent half.

von8794:
@clawsweeper re-review

Two of the open items are addressed in `f75619f2f`: renewal is now gated on `attempt > 1`, so the primary candidate never extends the deadline, and the regression test records the primary's deadline and asserts it is unchanged from admission before asserting the fallback renewal. CI on the new head is 165 success, 0 failed, 0 pending.

The remaining P1 is the whole-run timeout policy question, which I have answered with three options in the comment above rather than deciding it for you.

PR #141876, opened 2026-09-08
fix(npm): verify plugin-owned runtime chunks

RomneyDa:
## What Problem This Solves

Release verification must distinguish plugin-owned shared chunks from undeclared root dependencies. Generated region comments are fragile evidence, and a package-wide dependency exemption can hide root `createRequire()` imports or imports from another build output.

## Why This Change Was Made

Record plugin ownership for exact emitted chunk filenames and SHA-256 hashes. The verifier scans the actual installed imports, requires the owning plugin manifests to declare dependencies, and follows root-relative imports with Node's resolution rules. Postbuild carries verified ownership through its import rewrites and generated stable aliases.

Remove the region parser, import offsets, version-specific fallback, and dependency-wide ownership schema. Rebuilt older release branches use the same artifact contract; no dated compatibility list is introduced. Keep the real build tests in the existing verifier test file so the GitHub test plan stays within its existing 80-job cap.

Sister producer/lifecycle backport: #141922 (`extended-stable/2026.7.33`). Its full build passes and the revised main verifier reports zero dependency-manifest errors against the rebuilt output.

## User Impact

Release qualification catches undeclared root dependencies while accepting dependencies owned by the correct plugin. This works for current builds and rebuilt extended-stable releases, including `2026.7.33`.

## Evidence

- 254 focused verifier, release-check, real bundler, and postbuild tests passed.
- Reproduced the original root `createRequire()` false acceptance; the revised verifier rejects it.
- Extensionless and directory CommonJS regressions failed before the resolver repair and pass afterward.
- Real postbuild regression fails without ownership preservation and passes with it.
- GitHub compact planner: 80 jobs; no CI limits or packing rules changed.
- Independent autoreview through P2: clean.

- Full `pnpm build` passed before the mechanical merge of the newer PR head; all 254 focused tests passed again afterward.
- Final built ownership artifact: 705 chunks, zero hash mismatches, zero dependency-manifest errors for both main and the rebuilt backport.
- Core, UI, core-test, extension, script, and root-test typechecks passed; core, extension, and script lint lanes passed.
- Final fixture layout also passes all 164 verifier tests and all 10 extension-boundary tests; independent follow-up review is clean through P2. Root-test lint and targeted producer/postbuild lint passed.


RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

RomneyDa:
Compatibility proof update:

- Matching frozen producer backport #141922 is now CI-green (64/64), mergeable, and emits the same versioned chunk/hash ownership artifact for `extended-stable/2026.7.33`.
- The rebuilt frozen output was verified by this PR's consumer with zero ownership hash mismatches and zero dependency-manifest errors.
- `dist/runtime-dependency-ownership.json` is build/package verification metadata, not runtime persisted user state; no migration or upgrade path reads old application data. Existing published artifacts remain unchanged, while qualification of rebuilt branches uses the matching producer.

@clawsweeper re-review

PR #141877, opened 2026-09-08
refactor(scripts): test default HTML response cap

steipete:
## What Problem This Solves

The HTML comparison script's tests supplied a private 1 KiB override, so they did not exercise its real 5 MiB response limit. Changing that default could leave every test green.

## Why This Change Was Made

Remove the test-only byte-limit override and its options bag. Tests inject fetch directly and check the fixed default using an independent 5 MiB expectation. The shared bounded reader, abort timer, CLI URL arguments, vendor comparison path, and UTF-16 output handling remain unchanged.

## User Impact

The operational default remains 5 MiB. The tests now detect drift in that default. All four script cases remain, with three titles clarified, alongside the unchanged shared reader coverage. Tooling shrinks by nine lines and tests by two.

## Evidence

- Baseline script and shared-reader suites: 15 passed.
- Change only the production default to 6 MiB: old tests still pass all four cases; new tests fail exactly the declared-size and streamed-overflow cases, while two other cases pass.
- Restore the correct candidate: all 15 cases pass. The subsequent formatter change only reflows one reader call; that whitespace-only equivalence was verified separately.
- Independent Codex review through P2 is clean, and scoped formatting/diff checks passed.
- All 19 normal changed-file checks passed on the final formatted source, including script/root-test types and scoped lint ([Linux Testbox run](https://github.com/openclaw/openclaw/actions/runs/34189307023)). The one-shot lease and temporary capsule are cleaned up.

All HTML responses in this proof are injected. No live vendor requests were made, and no temporary control change remains.

### DIFF
--- a/scripts/firecrawl-compare.ts
+++ b/scripts/firecrawl-compare.ts
@@ -24,11 +24,6 @@ const userAgent =
 const timeoutMs = 30_000;
 const FETCH_HTML_MAX_BYTES = 5 * 1024 * 1024;
 
-type FetchHtmlOptions = {
-  fetchImpl?: typeof fetch;
-  maxBytes?: number;
-};
-
 function truncate(value: string, max = 180): string {
   if (!value) {
     return "";
@@ -38,7 +33,7 @@ function truncate(value: string, max = 180): string {
 
 async function fetchHtml(
   url: string,
-  options: FetchHtmlOptions = {},
+  fetchImpl: typeof fetch = fetch,
 ): Promise<{
   ok: boolean;
   status: number;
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [88d2f841b57b01f14ba400b5c81636e75075140b](https://github.com/openclaw/openclaw/pull/141877/commits/88d2f841b57b01f14ba400b5c81636e75075140b)
- Landed commit: [7520fe674f518389984c8fe4eae5461ed19597a8](https://github.com/openclaw/openclaw/commit/7520fe674f518389984c8fe4eae5461ed19597a8)

PR #141878, opened 2026-09-08
refactor(agents): derive conversation capability profile

RomneyDa:
## What Problem This Solves

Conversation capability resolution maintained a 97-line result interface separately from the function that constructs every returned field, allowing the two to drift.

## Why This Change Was Made

Make the resolver the contract owner and derive its exported result type from the producer. The cleanup also makes its now-internal requester-policy source type private.

## User Impact

No runtime behavior change. Capability additions and removals can no longer silently diverge from a hand-maintained parallel result model.

## Evidence

- Full `pnpm check:changed` passed
- P0–P2 autoreview scoped-clean
- Production diff: +7/−111 (net −104)

### DIFF
--- a/src/agents/conversation-capability-profile.ts
+++ b/src/agents/conversation-capability-profile.ts
@@ -4,7 +4,6 @@
  * new public access-profile config surface.
  */
 import { uniqueStrings } from "@openclaw/normalization-core/string-normalization";
-import type { ChatType } from "../channels/chat-type.js";
 import { normalizeChatType } from "../channels/chat-type.js";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import type { GroupToolPolicyConfig } from "../config/types.tools.js";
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper re-review

Data-model compatibility proof for exact head `577d9430f89d1a87ecd69d84d559b24e25de34e2`:

- The two-file patch only replaces a handwritten internal result declaration with `ReturnType<typeof resolveConversationCapabilityProfile>` and removes one unused type export.
- No executable expression, returned value, serializer, parser, storage key/path, schema, migration, or persisted version changes.
- `RequesterToolPolicySource` was type-only and had no external/plugin SDK/package/docs consumers; making it private changes no runtime or stored representation.
- Exact-head hosted CI is fully green, as are the full local changed-file gate and scoped P0–P2 autoreview.

The prior review itself confirmed runtime policy production/consumption and returned values are unchanged. Please reassess the mechanically detected serialized-state blocker using this compatibility evidence.

PR #141879, opened 2026-09-08
fix(memory): recover searches after manager replacement

steipete:
## What Problem This Solves

Fixes an issue where users searching memories receive “Memory Search failed” and misleading index-rebuild guidance when the memory manager closes or is replaced during retrieval. A valid semantic index could be reported as incompatible with keyword-only search.

## Why This Change Was Made

A retired memory manager can still read its shared SQLite connection after its embedding provider has closed. Status now rejects retired managers, and the tool captures post-search status inside its existing single-reacquisition boundary before finalizing results. The recovery matcher now recognizes the manager's actual closed error. Shared query context and removal of a redundant intermediate value keep the production diff at **−2 lines**.

## User Impact

Memory searches recover from manager replacement without rebuilding a healthy index. Genuine embedding-model mismatches still return their diagnostic and repair guidance. The existing public search API and visibility filtering are preserved.

## Evidence

- Before the fix, both real-manager regression cases failed: closing before retrieval returned “Memory index manager is closed”; closing after retrieval returned “index was built for model mock-embed, expected fts-only”. Both use real manager teardown while confirming shared SQLite remains readable.
- After the fix, all **67 tests** passed across the tool, real-manager integration, and provider-lifecycle suites. Recovery reacquires once and makes no indexing-embedding calls; the genuine model-change control still fails correctly.
- Independent autoreview through P2: scoped-clean, no actionable findings.
- Live tool proof passed on Linux/Node 24.19.0 with real OpenAI `text-embedding-3-small`: one synthetic file/chunk and a complete 1536-dimensional vector index; real search followed by real manager teardown; the tool reacquired and returned the match while shared SQLite stayed open. A SHA-256 comparison of every chunk row plus index metadata was unchanged. A true change to `text-embedding-3-large` remained unavailable without changing the index. [Blacksmith Testbox run](https://github.com/openclaw/openclaw/actions/runs/34189812188).
- `check:changed` passed on Blacksmith Testbox (Linux/Node 24.19.0), including production/test typechecks, lint, formatting, plugin boundaries, and runtime import-cycle checks: [run](https://github.com/openclaw/openclaw/actions/runs/34189874690). Delegated command exit codes were zero; stopped Testbox workflows may display cancelled after lease cleanup.

### DIFF
--- a/extensions/memory-core/src/memory-search-tool-query.ts
+++ b/extensions/memory-core/src/memory-search-tool-query.ts
@@ -63,7 +63,7 @@ function isClosedMemoryStoreError(error: unknown): boolean {
     message.includes("database is not open") ||
     message.includes("database connection is not open") ||
     message.includes("database handle is closed") ||
-    message.includes("memory search manager is closed")
+    message.includes("memory index manager is closed")
   );
 }
 
@@ -93,6 +93,7 @@ export async function executeMemorySearchToolQuery(params: {
           ? query.indexedSources
           : query.defaultSources
         : undefined);
+  const queryContext = { query, visibility, searchSources, startedAt };
 
   const searchOnce = async () => {
     const allowedSources = searchSources ? new Set(searchSources) : undefined;
@@ -132,11 +133,8 @@ export async function executeMemorySearchToolQuery(params: {
             void finalizeMemorySearchToolQuery({
               active,
               searched: { candidates: memoryCandidates, searchWindow },
-              query,
-              visibility,
-              searchSources,
+              ...queryContext,
               runtimeDebug: [...runtimeDebug],
-              startedAt,
               effectiveMode: "keyword-only",
             }).then(
               (result) => {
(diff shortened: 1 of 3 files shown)


steipete:
Reviewed the data-model compatibility flag against the exact patch: this PR changes no SQLite schema, migration, stored format, embedding metadata, or index-write path. It changes manager-lifetime validation and the timing of a status read, preserving the existing index identity checks and public search contract. No data migration is needed.

The live compatibility control indexed synthetic data with real OpenAI `text-embedding-3-small`, closed the real manager after retrieval, and successfully reacquired against that same existing index. SHA-256 over every chunk row plus all index metadata was identical before and after recovery. An intentional switch to `text-embedding-3-large` still returned the correct incompatibility error without modifying the index. The two new local regressions additionally assert that recovery makes no indexing-embedding calls.

The filename/embedding-metadata classifier flag does not apply to this lifecycle-only patch. Independent review found no actionable findings, and exact-head hosted CI is green.

steipete:
Merged via squash.

- Prepared head SHA: [ec4b2d093ecb35ba204ae062079148e1a0f30605](https://github.com/openclaw/openclaw/pull/141879/commits/ec4b2d093ecb35ba204ae062079148e1a0f30605)
- Landed commit: [6627766f5bfa403d5425e7c1da40f919fa924b45](https://github.com/openclaw/openclaw/commit/6627766f5bfa403d5425e7c1da40f919fa924b45)

PR #139944, opened 2026-09-06
fix(test): avoid false presence waiter timeouts

steipete:
## What Problem This Solves

Presence tests could report a timeout even after the expected clear or activity send had completed. Their waiters suspended, then unconditionally reported failure when the final wait interval ended.

## Why This Change Was Made

Each of the three waiters now checks its existing completion predicate after the final suspension. The original clock, four-second deadline, ten-millisecond sleep, yield limits, caller assertions and cleanup stay unchanged. All temporary clock controls and diagnostic tracing used to prove the race have been removed.

## User Impact

This corrects false failures in test support. Production behavior is unchanged; the final PR changes one test file by six added and three removed lines, with no new permanent test case or runtime hook.

## Evidence

A controlled before-fix [CI run 34022454546](https://github.com/openclaw/openclaw/actions/runs/34022454546) completed the second clear before advancing its logical clock past the deadline, then recorded the incorrect timeout. The unchanged controlled scenario [passed after the repair in CI 34024499221](https://github.com/openclaw/openclaw/actions/runs/34024499221): `calls=2 elapsedSeconds=5`, followed by the named case passing. All 14 ordinary presence functions (16 cases) also passed. These results prove the helper gap; they do not attribute every historical presence-test failure to it.

After removing the temporary controls, the complete current test file passed typechecking, formatting, lint and parsing, plus all 12 selected canonical check stages. Reversing only the three predicates and their comment reconstructs the original base file exactly. One independent managed P2 review found no actionable issue. The cleaned final head `a929799a35c6ad15331d9e61da92abd2a03ad376` passed [CI 34026464893](https://github.com/openclaw/openclaw/actions/runs/34026464893), including all 14 ordinary presence functions / 16 cases, the Mac Node shards and release build. Actual tested merge: `2bb527fda398bf459b26c671d30fc4b46bef480c`. This final run complements the retained controlled before/after proof.

PR #141884, opened 2026-09-08
fix(scripts): run macOS tooling under /bin/bash and make the framework-merge fixture self-contained

steipete:
## Problem

Fixes an issue where macOS packaging, signing, and PR tooling can hang before producing output when PATH selects Homebrew Bash 5.3+. Under pipe-buffer pressure, Bash can block writing a heredoc before its reader starts. Explicit `bash script.sh` callers also bypass a script's shebang.

The framework-merge test additionally borrowed `/bin/ls`, making its architecture set depend on the host OS. The reported merge failure already passes on this base: #141056 fixed the early-exiting `grep` pipelines. This change preserves that repair.

## Solution

The Bash 3.2 compatibility audit covers 205 scripts and sourced helpers. Replace Telegram package metadata `mapfile` with a read loop, replace three fleet tenant lowercase expansions with `tr`, and safely expand optional empty arrays under `set -u` across Docker/Podman launchers, installers, publishing cleanup, and Android emulator arguments. Add a static runtime-construct scan plus a bind-ownership regression test. No re-exec guard exclusions or compatibility exceptions are needed.

Pin 44 native Apple/macOS scripts and helpers, their nested Bash invocations, and the Apple/Parallels package commands to `/bin/bash`. Add one uniform inline Darwin re-exec guard to 91 portable entrypoints, including callers of heredoc-bearing libraries. Inlining preserves standalone scripts and the PR wrapper's materialized dependency closure. Keep Podman's boolean parsing compatible with Bash 3.2.

File-based and streamed installers re-exec under system Bash on affected macOS versions. Streamed input is captured with `mktemp` in a private mode-0600 file and unlinked immediately after re-entry; only sourcing retains an actionable rejection. Public install/recovery commands and emitted guidance keep portable `| bash`. The shebang/Bash 3.2 scan now lives in the existing `build-and-run-mac.test.ts` tooling file, leaving CI planner packing and capacity policy unchanged.

Generate inert Mach-O fixtures with deterministic x86_64, arm64, and arm64e headers. Real `file` and `lipo` tools classify and merge them; assertions verify the architecture set and exact slice bytes under a checkout path containing glob metacharacters. Add interpreter policy and re-exec tests, and document the rule and empty-log symptom.

## Impact

macOS tooling selects the system Bash without depending on Homebrew Bash. Portable scripts retain their interpreter on other platforms. Framework fixture coverage no longer depends on a system executable's slices or on a compiler SDK.

## Evidence

Correction proof: both new streamed-installer regressions fail on the previous guard and pass on Homebrew Bash 5.3.15 with no temp files left. Scan/routing: 642 passed; selected installer cases: 23 passed under each PATH order; unchanged planner: both coverage/cap tests passed. All 139 shell syntax checks, installer ShellCheck, 1,135 changed-gate tests, corrected lint/focused changed checks, and fresh P0/P1 autoreview pass. Exact-head CI on `8541789c8718` is blocked by [Swift presence retry test failure](https://github.com/openclaw/openclaw/actions/runs/34198825215/job/101972882916): one clear call instead of two. The test, reporter, scheduling helper, native runner, and CI workflow are byte-identical to main `8ac88a1a335a`; the prior #139944 waiter fix is already included. The other 126 CI jobs passed, including agents coverage; Swift also failed the required aggregate CI gate. No merge was attempted.

Compatibility follow-up: all 140 shell files in the PR pass `/bin/bash -n`; the five requested packaging/signing/shebang/iOS/Parallels test files pass **481 tests with default PATH and 481 with `/bin` first**. Broader Docker/installer/helper coverage reports **798 passed**, six requested tar-case skips, two existing Linux-only skips, and two baseline npm 12 workflow timeouts. Both timeout cases reproduce in isolation; their seven test/workflow/copied source files match the prior head and `origin/main` byte-for-byte. The documented validator-diagnostic case was not rerun. `node scripts/check-changed.mjs` and `git diff --check` pass. Fresh Codex autoreview reported six INDEX-only findings, each explicitly fixed in the reviewed working tree; staging those exact files resolved all six, with no working-tree P0/P1 findings.

- `node scripts/run-vitest.mjs test/scripts/package-mac-app.test.ts test/scripts/codesign-mac-app.test.ts test/scripts/mac-script-shebangs.test.ts` with the default PATH (`/opt/homebrew/bin/bash` resolves first): **311 passed** on the final files.
- `PATH=/bin:$PATH node scripts/run-vitest.mjs test/scripts/package-mac-app.test.ts test/scripts/codesign-mac-app.test.ts test/scripts/mac-script-shebangs.test.ts`: **306 passed** before the additional package-command/Parallels guards; final packaging and signing coverage also passed in the full changed gate below.
- `/bin/bash -n <each changed shell script>`: **all 137 passed**. The new policy assertions reject the original entrypoint choices; the real Homebrew Bash re-exec probe preserves arguments, including empty arguments and glob characters.
- `scripts/codesign-mac-app.sh --help` with default PATH: **exit 0**.
- `PATH=/bin:$PATH node scripts/check-changed.mjs`: **exit 0**, including **1,135 passing tests**, formatting, typechecking, script lint, and Docker scheduler dry-run. Output was captured to a file. The initial run caught four test typing errors, fixed before this successful final run.
- `PATH=/bin:$PATH node scripts/run-vitest.mjs test/scripts/ios-release-fastlane-gates.test.ts test/scripts/parallels-smoke-model.test.ts`: **147 passed**. Broader sibling coverage also exercised installers, Docker helpers, native scripts, and PR wrapper lifecycle paths.
- Fresh Codex autoreview after the final edits: **no actionable P0/P1 findings**. `git diff --check`: **passed**.

## Notes/Gaps

Two broader sibling failures were reproduced after restoring the relevant script to its original HEAD bytes, then restoring the reviewed changes byte-for-byte:

- `PATH=/bin:$PATH node scripts/run-vitest.mjs test/scripts/docker-build-helper.test.ts -t 'retains self-upgrade RPC evidence through the actual launcher'`: **six failures on both versions**. A standalone `/usr/bin/tar -C "$src" -cf - . | /usr/bin/tar -xf - -C "$dst"` also fails on this host: writer exit 1, reader exit 0, with `tar: Write error`; a file-based archive round-trip passes. These cases fail before their intended RPC assertions.
- `PATH=/bin:$PATH node scripts/run-vitest.mjs test/scripts/find-reusable-release-validation.test.ts -t 'surfaces bounded validator rejection'`: **one failure on both versions**. The expected bounded detail is replaced by the generic validator rejection diagnostic. That test already invokes `/bin/bash`.

These are separate follow-ups; the required changed-file gate passes. Full native packaging and the full Swift suite were intentionally left to the coordinator. `package-mac-app.sh` has no help option, so it was not invoked with `--help`. No release or merge is included.


steipete:
Coordinator live proof on the host where the original stall was observed (macOS 27.0, Homebrew GNU bash 5.3.15 first on PATH):

- Before this branch, `scripts/package-mac-app.sh` invoked through PATH bash sat for 11h51m with a 0-byte log, blocked in `heredoc_write → write` on its first heredoc command substitution; a second attempt stalled the same way inside `scripts/codesign-mac-app.sh` after the Sparkle signing step.
- With this branch checked out, `OPENCLAW_SKIP_MLX_TTS=1 ./scripts/package-mac-app.sh` run via its shebang with `/opt/homebrew/bin/bash` (5.3.15) resolving first on PATH completed end to end: JS build, Swift build for `OpenClaw` and `openclaw-mac`, node-worker staging, signing of the bundle and helpers, and `codesign --verify --deep --strict` all passed, producing `dist/OpenClaw.app` with both binaries.
- The reviewed follow-up commit replaced the one `mapfile` and the case conversions the guard would have routed into bash 3.2, and its static scan over the guarded scripts and their sourced libraries reports no remaining bash-4-only constructs.

steipete:
Merged via squash.

- Prepared head SHA: [8541789c8718a53c40671183aa6647bc4302056b](https://github.com/openclaw/openclaw/pull/141884/commits/8541789c8718a53c40671183aa6647bc4302056b)
- Landed commit: [299ea7c393b840cc57782c848f80a24d9d03067a](https://github.com/openclaw/openclaw/commit/299ea7c393b840cc57782c848f80a24d9d03067a)

PR #141886, opened 2026-09-08
fix(cli): name the System Agent remedy for message ownership

### LINKED ISSUE #141835: [Bug]: `openclaw message send` unusable on explicit multi-agent fleets — AgentSelectionRequiredError advises a `--agent` flag the command does not have
### Bug type

### Summary

On an explicit multi-agent fleet (`agents.ownership: "explicit"`), `openclaw message send` fails for **every** channel with `AgentSelectionRequiredError`. The error's remediation text tells CLI callers to pass `--agent <id>`, but `message send` registers no such option — and neither does its parent `message` command, nor the global CLI. There is no per-invocation way to satisfy it.

This is the same shape as #124679 (`openclaw directory`) and #134421 (`channels login/logout`), on a different surface. It is not covered by #128637, which is about the exec auto-reviewer and the Control UI channel wizard.

The practical impact is larger than a single command being unusable: `openclaw message send` is the documented way for scripts and cron jobs to post notifications. After upgrading, every alerting script on the host went silent at once. Scripts that guard the call with `|| true` (a reasonable thing to do, so a failed notification doesn't abort a backup job) log success while delivering nothing, so the outage is invisible until someone notices alerts stopped arriving.

### Steps to reproduce

1. Configure two or more agents with `agents.ownership: "explicit"` and leave `agents.defaults.systemAgent.agentId` unset.
2. Run any send, e.g.:
   ```
   openclaw message send --channel slack --target C01234567 -m "test" --dry-run
   ```

### Expected behavior

Either `message send` accepts `--agent <id>` as the error suggests, or the error names a remedy that actually exists on this surface.

### Actual behavior

```
Multiple agents are configured, but this operation has no explicit owner. Select an agent explicitly; CLI callers can pass --agent <id>, channels can add a binding, and ambient services can set their agentId target.
```

Following the hint fails:

```
$ openclaw message send --agent shanks --channel slack --target C01234567 -m "test"
OpenClaw does not recognize option "--agent".

$ openclaw --agent shanks message send --channel slack --target C01234567 -m "test"
Try: openclaw --help
```

`OPENCLAW_AGENT`, `OPENCLAW_AGENT_ID` and `OPENCLAW_DEFAULT_AGENT` have no effect. `openclaw doctor` reports no routing-owner finding for the affected channels. Reproduced on both `--channel slack` and `--channel telegram`, and with `--dry-run`, so it is not channel- or delivery-specific.

### Root cause (from the shipped bundle)

`dist/register.message-*.js` resolves the owner with no requested id:

```js
const agentId = resolveAmbientOwnerAgentId(cfg);
```

`resolveAmbientOwnerAgentId(cfg, requestedAgentId, context)` accepts a `requestedAgentId` — the CLI simply never passes one and registers no flag to supply it. With `ownership: "explicit"` the legacy-compatibility fallback is skipped, so resolution falls through to `resolveSoleAgentId()`, which throws whenever more than one agent exists.

### Workaround

Setting the ambient owner in config works and applies without a gateway restart:

```
openclaw config set agents.defaults.systemAgent.agentId <agent>
```

This is worth surfacing in the error text — it is the only remedy that actually works for this surface, and it is not one of the three the message currently lists.

### Suggested fix

Register `--agent <id>` on `message send` and thread it through to `resolveAmbientOwnerAgentId` as `requestedAgentId` (the resolver already supports it). Failing that, pass a `context.hint` naming `agents.defaults.systemAgent.agentId` so the generic `AgentSelectionRequiredError` text isn't shown on a surface where none of its three suggestions apply.

### OpenClaw version

2026.9.2 (3928bad)

### Operating system

Ubuntu 24.04 (noble)

### Install method

npm (global)

Dreamer-HIT:
## What problem this solves

`openclaw message send` stops when an explicit multi-agent fleet has no System Agent, but its error recommends `--agent`, which the message command does not support. This also affects dry runs.

## Why this change was made

Pass the existing message CLI context to the owner resolver and name the supported remedy: `openclaw config set agents.defaults.systemAgent.agentId <id>`. The existing formatter preserves active profile/container routing context. Document that System Agent selection also owns other ambient system work.

## User impact

Operators receive a command they can use with a configured agent ID. Owner selection, legacy and sole-agent behavior, permissions, config schema, channel delivery, and JSON error structure stay unchanged. This adds no per-invocation agent selector.

Fixes #141835.

## Evidence

Real compiled CLI commands used an isolated two-agent configuration, synthetic Slack metadata, a canonical synthetic target, and a network namespace without external access. Process-tree network observations and config snapshots accompanied each invocation. No live channel or real credential was used.

- Pinned main: normal and dry-run calls exited 1 with unsupported `--agent` advice.
- Candidate: both calls identify `message CLI` and print the existing System Agent config command, with unchanged config and no observed network request.
- Running the printed remedy succeeds; the same dry-run then reaches the existing message action. The profile variant preserves the profile and applies its remedy successfully.
- Existing per-agent action policies distinguish the exact selected owner. System Agent precedence, retained legacy ownership, sole-agent behavior, invalid config/action/option/target/channel/account, and unresolved SecretRef controls remain covered.
- The two regression cases fail for the intended hint mismatch on baseline. All 25 message-command tests and 10 CLI registration/formatter tests pass on the candidate. Focused formatting and lint pass.
- The new documentation anchor passes the canonical audit. The full local audit retains 39 unrelated ClawHub mirror failures; no docs exception was changed. GitHub CI owns the full checks.

Fresh source-blind acceptance exercised 17 scenarios and 25 CLI calls with no behavioral failure. Source-only mechanisms and fixture lineage were reviewed separately. Fresh source review found no actionable defects; exact-head CI and the final landing review remain required.


obviyus:
Thanks @Dreamer-HIT!

**Proof:** Real compiled CLI commands reproduced the unsupported `--agent` advice on pinned main. The repaired normal and dry-run errors name the existing System Agent setting. Applying the printed command enables the identical dry-run, including under a real profile.

- **Bug:** An ownerless explicit fleet received a remedy unavailable to `openclaw message`.
- **Fix:** Pass message-specific context to the existing owner error and document the supported configuration remedy.

| Before | After |
| --- | --- |
| Normal and dry-run commands suggested unsupported `--agent` | Both print `openclaw config set agents.defaults.systemAgent.agentId <id>` |
| The error omitted the working profile-specific remedy | Printed profile command updates that profile and enables the same preview |

All 35 focused tests pass, along with independent behavior acceptance and source review. Existing owner precedence, per-agent policy, config/secret errors, and JSON shape remain intact. Proof used synthetic metadata and isolated networking; no live channel delivery is claimed. The changed documentation anchor passed; the local full audit's 39 absent external ClawHub mirror links were retained as an unrelated environment limitation.

PR #141887, opened 2026-09-08
fix(tests): isolate plugin artifact loader fixtures

steipete:
## What Problem This Solves

Fixes two plugin artifact tests that fail on Linux when setup modules observe uninitialized shared state after the artifact entry has already passed its native-loading checks.

## Why This Change Was Made

Extract the same archive into a separate directory for forced-source loading, following the isolation pattern already used by the neighboring test. Native module caches can otherwise supply the entry while the source loader evaluates setup against a different shared-module instance. Entry and setup still use one source loader and extraction together.

## User Impact

Test-only repair; no plugin runtime behavior changes. Native loading, source entry/setup initialization, metadata and shared-chunk assertions remain intact. The fixture adds 14 net test lines.

## Evidence

- On Linux x64 with Node 24.19.0, the unchanged full suite reproduced 14 passing tests and the same two failures. The isolated patched copy passed all 16 tests with no skips and the same ordered case inventory.
- Local Node 24.20.0: all 16 tests passed after the change.
- Full mapped changed-file checks and independent P2 review passed.

The Linux control and comparison source trees were preserved. No runtime performance claim is made.

### DIFF
--- a/src/cli/plugins-feature-artifact.test.ts
+++ b/src/cli/plugins-feature-artifact.test.ts
@@ -254,22 +254,36 @@ export default Object.assign(defineToolPlugin({ id: ${JSON.stringify(id)}, name:
         tools: [{ name: "artifact_echo", optional: true }],
       });
       if (setup) {
-        const setupPath = resolvePackageSetupSource(resolution);
+        // Keep native module caches out of the independent source-loader graph.
+        const sourceExtracted = path.join(parent, "source-extracted");
+        await fs.mkdir(sourceExtracted);
+        await extract({ file: archive, cwd: sourceExtracted, strict: true });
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a72009cd3a8cb760f2a3d2f4ed5ef8084b30eb6e](https://github.com/openclaw/openclaw/pull/141887/commits/a72009cd3a8cb760f2a3d2f4ed5ef8084b30eb6e)
- Landed commit: [42fa00e04e78f7ada936590b5d1fa95cd3004371](https://github.com/openclaw/openclaw/commit/42fa00e04e78f7ada936590b5d1fa95cd3004371)

PR #141905, opened 2026-09-08
fix(qa-channel): preserve inbound references on queued replies

steipete:
## What Problem This Solves

Fixes an issue where QA Channel replies lose their inbound message reference when a new message is queued behind an active turn. A restart-recovery scenario delivered both the recovery answer and the new answer, but the queued answer had no `replyToId`.

## Why This Change Was Made

Carry the queued external inbound message ID through the existing transport resolver as optional context. QA alone uses it for implicit all-mode correlation, matching its ordinary inbound callback. Core does not force a reply target onto other channels.

Explicit targets, null/empty opt-outs, and first/off/batched filtering remain unchanged. Internal and inter-session turns do not supply synthetic IDs; the handoff uses typed provenance and requires the originating channel to match.

## User Impact

Queued QA answers retain the same inbound correlation as ordinary QA replies. Other channel policies and explicit reply targets are preserved. The optional resolver field is documented; there are no configuration, schema, or new SDK-subpath changes.

## Evidence

- A controlled current-main test using the real followup router and QA plugin reproduced the missing reference; the explicit-target control passed. The same probe passed after the repair.
- All 246 affected routing, threading, delivery-custody, and QA inbound tests passed. Final retained tests follow the repository's type-graph boundaries: 39 core handoff and QA-owned policy controls passed after separating the initial combined probe.
- Full scoped checks passed, including core/plugin production and test types, graph boundaries, lint, formatting, exports, and applicable guards. The initial combined test's forbidden extension import was removed without adding a guard exception.
- Independent P2 review and lead review found no actionable issues.

This is transport-boundary proof, not an authenticated live-channel run. The change deliberately leaves first/batched semantics alone and is independent of release qualification.

### DIFF
--- a/docs/plugins/sdk-channel-plugins.md
+++ b/docs/plugins/sdk-channel-plugins.md
@@ -889,6 +889,13 @@ native API requires a thread root can resolve a current-message reply against
 the admitted thread without redirecting arbitrary explicit `replyToId` targets.
 Omitted intent keeps the existing explicit-target behavior.
 
+For queued replies to the originating channel, the hook may also receive
+`currentMessageId`, the external inbound message ID captured by the queue owner.
+It is optional context, not an explicit reply target: core does not attach it as
+`replyToId`. The channel decides whether to use it for implicit correlation,
+while preserving explicit targets, null opt-outs, and reply-mode filtering.
+Internal and inter-session turns do not supply this external message fact.
+
 Auth-only channels can usually stop at the default path: core handles
 approvals and the plugin just exposes outbound/auth capabilities. Native
 approval channels such as Matrix, Slack, Telegram, and custom chat transports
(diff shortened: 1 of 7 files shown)


steipete:
Compatibility follow-up for the review note: a resolver using the pre-change parameter shape remains assignable to `ChannelThreadingAdapter.resolveReplyTransport`. I typechecked that legacy callback against the current contract and invoked it with the added context; its explicit reply target and thread were unchanged. The normal full core/plugin production and test typechecks also pass.

The field is optional context, and existing argument values and callback return semantics are unchanged. This checks the legacy callback contract rather than claiming every third-party implementation was run. No product change or compatibility shim is needed. Exact-head CI is green.

steipete:
Merged via squash.

- Prepared head SHA: [7bdcafa67c87f568baf6e2c995ad19f32175734d](https://github.com/openclaw/openclaw/pull/141905/commits/7bdcafa67c87f568baf6e2c995ad19f32175734d)
- Landed commit: [386e7091065f70b633ded81158ea21396cc8c193](https://github.com/openclaw/openclaw/commit/386e7091065f70b633ded81158ea21396cc8c193)

PR #141908, opened 2026-09-08
refactor: give root project validation one test owner

steipete:
Related: #139428

## What Problem This Solves

Resolves a problem where reporter-contract tests repeatedly initialize the full root project graph even though their assertions only inspect reporter settings. Their child processes also inherit focused test-selection filters, making focused and full-suite configuration resolution differ.

## Why This Change Was Made

Resolve the root reporter settings without expanding projects for both default and CLI reporter checks. Preserve real full-root loading once in the project-config suite, with a successful subprocess exit and a nonempty resolved-project assertion. Both child paths clear inherited include-file filtering; the graph smoke deliberately uses the CI environment.

## User Impact

No product behavior changes. All 19 reporter resolutions per environment and both environment rows remain, with unchanged reporter assertions. The project suite now explicitly protects the root-watch startup contract.

## Evidence

- Complete suites: 26 original cases passed; all 27 cases passed after the change.
- A temporary duplicate name in a real child config made the new smoke fail with Vitest's duplicate-project error. Restoring the config restored the complete passing suite.
- Changed-file gate and fresh P2 review passed.
- This adds 30 test lines. The local run is not a measured full-suite speedup or proof that the historical timeout is fixed. Existing comparison controls must be updated explicitly for the new ownership before another campaign comparison.

### DIFF
--- a/test/vitest-projects-config.test.ts
+++ b/test/vitest-projects-config.test.ts
@@ -1,5 +1,6 @@
 // Vitest project config tests validate aggregate Vitest project wiring.
 import { afterEach, describe, expect, it } from "vitest";
+import { spawnNodeEvalSync } from "../src/test-utils/node-process.js";
 import { createPatternFileHelper } from "./helpers/pattern-file.js";
 import { normalizeConfigPath, normalizeConfigPaths } from "./helpers/vitest-config-paths.js";
 import { auditFullSuiteTestFileOwnership } from "./vitest-projects-config.test-support.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [cc17f62c7bcc3e0fc34ca4727d781be03784ca41](https://github.com/openclaw/openclaw/pull/141908/commits/cc17f62c7bcc3e0fc34ca4727d781be03784ca41)
- Landed commit: [91d96f5a5a2611f634ad1e89bca3310b14b0e8d2](https://github.com/openclaw/openclaw/commit/91d96f5a5a2611f634ad1e89bca3310b14b0e8d2)

PR #141669, opened 2026-09-07
fix(sqlite): report observed integrity timeout phases

steipete:
Related: #141412.

## What Problem This Solves

Fixes an issue where operators diagnosing a database integrity timeout during startup or maintenance could not tell whether the child had opened the database, reached validation, or begun closing. The existing 30-second timeout covers the child's whole lifetime.

## Why This Change Was Made

The child reports three bounded phases through its existing private IPC channel: `opening`, `checking`, and `closing`. The parent retains the last observed phase, including `starting` and `result-received`, and appends it to timeout and incomplete-exit errors. Messages are flushed before blocking work; progress cannot count as a final result or overwrite one.

The existing deadline, SIGKILL cancellation, full integrity and foreign-key checks, file-identity checks, native error envelope, and ownership until child closure are preserved. Failed closing diagnostics still close the native database and retain an existing integrity failure. No SQLite schema, stored representation, SQL, retention, configuration, or execution-policy change is included.

## User Impact

Timeout errors now include a field such as `lastObservedPhase=checking`. This records the last message the parent received; it does not identify the child's exact current location, distinguish integrity from foreign-key work, establish CPU use, or explain the underlying stall. The reference documentation defines each value.

## Evidence

Verified candidate: `d32cb57583a774a4a55c829e0342c756799a8398`.

- All **68 focused tests pass**: 15 integrity/protocol cases, 38 database-admission cases, and 15 maintenance cases, including real cancellation and child-closure fencing.
- Pre-fix protocol checks failed for missing timeout phase detail, missing-completion diagnostics, and a final result overwritten by a later phase. Real FK cases preserved their native failure but lacked the expected phases. The candidate passes those checks, including a failed closing-phase send with native close independently observed.
- The existing WAL-commit observer initially failed because an unrelated closing phase erased its confirmed commit flag. It now latches that actual observation; the same foreign-key and ownership assertions pass.
- Full changed checks passed, including production/test types, lint, unused exports, storage guards, and zero runtime import cycles. Independent review through P2 found no actionable findings.
[...]

PR #141918, opened 2026-09-08
fix: gateway boot fails with SQLite integrity check timeout on large agent databases

steipete:
Related: #141412, #141669.

## What Problem This Solves

Fixes an issue where a Gateway with a large agent database would fail every boot with "SQLite integrity check timed out after 30 seconds" whenever a schema migration was pending, restarting in a loop until the restart-loop breaker suppressed channels. The team Gateway with a 9.4 GB agent database was down for two hours on 2026-09-07 and every hourly release since then has failed its first boot the same way.

## Why This Change Was Made

#141412 moved the boot-time full-file integrity check (`PRAGMA integrity_check` plus `foreign_key_check`) into a forked read-only child that inherits the fixed 30-second inspection deadline sized for a 300 MiB synthetic database, and SIGKILLs it on expiry. The child now receives a budget derived from the database file size, which the identity stat already provides: 30 seconds for startup and shutdown plus one second per 32 MiB, rounded up and capped at 30 minutes, so a 9.4 GB file gets 331 seconds. Budgets above 30 seconds are logged with path and size, and the timeout error names the applied budget and size while keeping the remediation text and the `lastObservedPhase` suffix. Read-only snapshot and inspection paths keep their 30-second limit. Non-goals: when the integrity check runs is unchanged; whether multi-gigabyte stores should pay a full scan on every migration-pending boot is a separate design question for the database owners.

## User Impact

Large installs boot again after upgrades that carry schema migrations, instead of crash-looping. Operators see the applied budget in logs and in any timeout error.

## Evidence

Team host measurements (warm cache): `integrity_check` 8 s, `foreign_key_check` 2 s on the 9.4 GB database; under I/O and memory pressure the same work exceeded 30 s on every boot for two hours. Regression tests fail on the pre-fix code for the intended reason: fixed 30 s fork deadline and old message captured, then real empty, small, and blob-padded databases verify base and size-derived budgets and the new error text. `node scripts/check-changed.mjs` passed; 61 focused tests pass. Codex-backed autoreview at P1 before landing.

## Known Gaps

Cold-cache behavior on the production host is not reproduced in tests; the budget is a conservative floor, not a measurement.


steipete:
Merged via squash.

- Prepared head SHA: [1c2c0ae0aa85dde2711f949764952f4b86086194](https://github.com/openclaw/openclaw/pull/141918/commits/1c2c0ae0aa85dde2711f949764952f4b86086194)
- Landed commit: [a6a4e88da29863c8527bef2710f7090e3a0d95d8](https://github.com/openclaw/openclaw/commit/a6a4e88da29863c8527bef2710f7090e3a0d95d8)

PR #141919, opened 2026-09-08
refactor(tests): drop implementation-coupled config assertions

steipete:
Related: #139428

## What Problem This Solves

Config footprint tests enforce private streaming helper names and the declaration order of two erased TypeScript properties. Those assertions can fail after behavior-preserving refactors without detecting a change in accepted configuration or runtime precedence.

## Why This Change Was Made

Remove the Matrix property-order case and four exact private helper-name assertions. Keep every retired streaming alias guard and source read, and rename that case to describe its remaining alias-exclusion contract. Public schema, compatibility and architecture checks remain unchanged.

## User Impact

No production or schema behavior changes. The test suite has less implementation coupling and shrinks by 13 net lines. No speed or aggregate coverage claim.

## Evidence

- Contract suite before: 9 passing cases. After: 8 passing cases, with only the approved deletion and case rename.
- Existing Telegram, Discord, Slack and MSTeams schema tests plus Matrix setup tests: 173 cases passed, retaining nested/optional streaming and canonical Matrix opt-in behavior.
- All retired-alias assertions and their four source reads remain unchanged. No SSRF tests or production code changed.
- Formatting, diff checks, the full changed gate and fresh P2 review passed.

### DIFF
--- a/src/plugins/contracts/config-footprint-guardrails.test.ts
+++ b/src/plugins/contracts/config-footprint-guardrails.test.ts
@@ -139,16 +139,12 @@ describe("config footprint guardrails", () => {
     }
   });
 
-  it("keeps canonical nested streaming paths in channel-owned schemas", () => {
+  it("keeps retired flat streaming aliases out of channel-owned schemas", () => {
     const telegramSource = readSource("extensions/telegram/src/config-schema.ts");
     const discordSource = readSource("extensions/discord/src/config-schema.ts");
     const msTeamsSource = readSource("extensions/msteams/src/config-schema.ts");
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [11684f44f419a413a752bbc59af4f69d8449d271](https://github.com/openclaw/openclaw/pull/141919/commits/11684f44f419a413a752bbc59af4f69d8449d271)
- Landed commit: [a86dd62e987f271c2f06fe34b6529cd8a0c40d61](https://github.com/openclaw/openclaw/commit/a86dd62e987f271c2f06fe34b6529cd8a0c40d61)

PR #141920, opened 2026-09-08
fix(release): backport runtime dependency ownership artifact

RomneyDa:
## Summary
- emit `dist/runtime-dependency-ownership.json` from the frozen 2026.7.33 tsdown build
- record extension-only ownership without parsing generated bundle comments
- provide the versioned artifact consumed by the mainline verifier from #141876

## Why
The frozen package bundles Discord and Teams runtime dependencies into root dist while their manifests live with the companion extensions. Mainline validation can consume explicit build provenance, but the 2026.7.33 producer must emit it.

The collector uses TypeScript preprocessing over modules actually included by tsdown. Any dependency also imported by root-owned source is omitted, preserving fail-closed root dependency validation.

## Validation
- frozen tsdown build succeeds
- ownership manifest contains `@discordjs/voice` → `discord` and `@microsoft/teams.apps` → `msteams`
- focused tsdown/ownership tests: 17/17
- npm dry-run includes `dist/runtime-dependency-ownership.json`
- formatting and `git diff --check` pass

### DIFF
--- a/scripts/lib/runtime-dependency-ownership-build-plugin.mts
+++ b/scripts/lib/runtime-dependency-ownership-build-plugin.mts
@@ -1,0 +1,95 @@
+import { readFileSync } from "node:fs";
+import { builtinModules } from "node:module";
+import path from "node:path";
+import type { TsdownPlugin } from "tsdown";
+import ts from "typescript";
+import { packageNameFromSpecifier } from "./plugin-package-dependencies.mts";
+import {
+  RUNTIME_DEPENDENCY_OWNERSHIP_ASSET_NAME,
+  RUNTIME_DEPENDENCY_OWNERSHIP_FORMAT_VERSION,
+  type RuntimeDependencyOwnership,
+} from "./runtime-dependency-ownership-contract.mts";
+
+type RuntimeOwner = "root" | `extension:${string}`;
+const NODE_BUILTIN_MODULES = new Set(builtinModules.map((name) => name.replace(/^node:/u, "")));
+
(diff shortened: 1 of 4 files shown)


RomneyDa:
Superseded by #141922, the matching producer/postbuild backport for the revised #141876.

The revised pair binds plugin ownership to emitted chunk filenames and hashes, verifies actual installed imports and root reachability, and preserves that evidence through postbuild rewrites and stable aliases. This avoids the package-wide dependency exemption in this PR, which can also exempt root imports from another build output or miss root `createRequire()` usage.

#141922 covers the existing `extended-stable/2026.7.33` branch with the same contract used by current builds; it does not introduce a release-specific compatibility list. Closing this alternative to keep one compatible main/backport pair. Neither PR has been merged or released.

RomneyDa:
Compatibility proof for the generated ownership artifact:

- `dist/runtime-dependency-ownership.json` is regenerated during package build; it is not runtime/user state and requires no migration.
- Existing 2026.7.33 runtime code never reads this file, so adding it cannot change upgrade/runtime behavior.
- The mainline verifier remains backward compatible: artifact-free frozen packages continue through its legacy ownership fallback.
- When the artifact exists, the verifier validates `formatVersion: 1` and the complete ownership shape before use; malformed or unknown formats fail closed rather than being partially interpreted.
- The frozen build/package proof confirms the versioned artifact is emitted into the npm tarball with the expected Discord and Teams ownership entries.

Therefore upgrades in either order are covered: old package/new verifier uses the fallback; new package/new verifier uses v1; old verifier ignores the extra package file.

RomneyDa:
@clawsweeper re-review

PR #141921, opened 2026-09-08
feat(claws): carry model and delegation settings in the Claw profile

steipete:
Related: closes the gap that #141339 tried to fill with a second format; that PR is closed in favor of this one.

## What Problem This Solves

A Claw (RFC 0016) describes one agent's portable setup, but the OpenClaw profile cannot say which model the agent prefers or which agents it may delegate to. Sharing a coordinator-style agent, or exporting one that has delegation targets, silently drops that configuration.

## Why This Change Was Made

Extends the conventional `profiles/openclaw.yml` `agent` object with two strict, optional objects:

- `model { primary, fallbacks? }`, each a `provider/model` reference.
- `subagents { allowAgents?, delegationMode? }`, ids following the Claw agent id rules; an empty `allowAgents` is an explicit "no delegation".

`claws add` and `claws update` disclose both in the plan (a Model line and a Delegation line) and bind them into the plan-integrity digest; a model missing from the local catalog or a delegation target missing from the local roster is a notice, not a blocker, and the values are applied as declared so a team can be installed one Claw at a time. `claws status` reports drift through the existing agent-config digest, and `claws export` writes both when the installed agent has explicit values (inherited defaults are not exported). Claws without the fields behave exactly as before; the experimental flag, CLI surface, and ClawHub track are unchanged.

## User Impact

A Claw can now carry "prefers this model" and "may delegate to these agents", and exporting an agent that has them round-trips. This is the piece needed for a shareable chief-of-staff Claw and for the role catalog in #141174 to express its delegation wiring in the Claw format later.

## Evidence

| Before: profile with the new fields is rejected | After: plan discloses model and delegation with notices |
| --- | --- |
| before | after |

- Tests: 603 passed across the Claw schema, application-schema, add, export, update, and CLI project suites (2 pre-existing skips); new cases cover accept/reject shapes, plan disclosure and notices, apply, drift, and export round trip. `node scripts/check-changed.mjs` passed; `pnpm check:architecture` reports zero runtime and madge cycles.
- Offline round trip in an isolated state dir with the extended fixture: zero blockers, notices for the example `acme/*` models and for the two absent delegation targets.
- Independent review pass on the branch: no actionable findings.

Line delta (task only): production +131, tests +276, fixtures +28, docs +23.

## Known Gaps

- Model availability is checked against the local catalog at plan time only; a model that later disappears is reported by ordinary agent diagnostics, not by `claws status`.


steipete:
Merged via squash.

- Prepared head SHA: [b312e170b856dd693ca6057c394189e4f4a6f906](https://github.com/openclaw/openclaw/pull/141921/commits/b312e170b856dd693ca6057c394189e4f4a6f906)
- Landed commit: [21d7660630aefa0df12cb7f59c11c2bfbc4f12d9](https://github.com/openclaw/openclaw/commit/21d7660630aefa0df12cb7f59c11c2bfbc4f12d9)

PR #141922, opened 2026-09-08
fix(npm): backport runtime chunk ownership to 7.33

RomneyDa:
## What Problem This Solves

Rebuilt `2026.7.33` packages need emitted chunk ownership evidence so current trusted release tooling can verify plugin dependencies without recovering ownership from generated region comments.

## Why This Change Was Made

Backport the producer and postbuild lifecycle changes from #141876 to `extended-stable/2026.7.33`. Emit exact chunk filenames, hashes, and plugin owners, and preserve verified ownership through the branch's actual `runtime-postbuild.mjs` implementation. Keep the real build tests in its existing postbuild test file.

The matching trusted verifier is now published in #141876 at `ac5048ec7b2b4cc5de809eae08d5518cb77ef523`; its chunk contract and producer match this backport exactly. That verifier reports zero dependency-manifest errors against the locally rebuilt branch artifact. Current trusted qualification tooling supplies the verifier. This backport does not require importing the branch's unrelated newer verifier dependencies.

This is the canonical sister backport for #141876 and supersedes the alternative #141920.

## User Impact

The next rebuilt `2026.7.33` artifact can use the same ownership verification as current releases. Existing frozen tarballs must be rebuilt to contain the evidence.

## Evidence

- Full `pnpm build` passed.
- 35 focused real-build and postbuild tests passed, including rejection of altered input bytes.
- All 1,570 final ownership entries match their emitted files and SHA-256 hashes.
- The revised main verifier reports zero dependency-manifest errors against the rebuilt output.
- Scoped production/test lint, formatting, and `git diff --check` passed.
- Strict producer/contract/build-config typecheck passed. The existing postbuild test typecheck has one confirmed baseline error: TypeScript resolves the older `.mjs` import to its incomplete `.mts` sibling; the actual `.mjs` tests pass.
- The branch's changed-check wrapper could not start its configured Crabbox backend. Whole-config lint also reports an existing, unchanged ternary at `tsdown.config.ts:44`; scoped changed production/test files are clean.

- Independent autoreview completed. Its Node 22.12–22.17 concern does not apply: the branch requires Node >=22.22.3. Loading the actual postbuild and `.mts` contract on Node 22.22.3 passed.
- Fixed the CI fixture/boundary collision by moving the generated shared module under the temporary `src/` tree. All 11 extension-boundary tests and 35 postbuild/build tests pass; fresh follow-up autoreview is clean through P2. No boundary rules were weakened.

### DIFF
--- a/scripts/lib/runtime-dependency-ownership-build-plugin.mts
+++ b/scripts/lib/runtime-dependency-ownership-build-plugin.mts
@@ -1,0 +1,59 @@
+import { createHash } from "node:crypto";
+import path from "node:path";
+import type { TsdownPlugin } from "tsdown";
+import {
+  RUNTIME_DEPENDENCY_OWNERSHIP_ASSET_NAME,
+  type RuntimeDependencyOwnership,
+} from "./runtime-dependency-ownership-contract.mts";
+
+export function createRuntimeDependencyOwnershipBuildPlugin(rootDir = process.cwd()): TsdownPlugin {
+  return {
+    name: "openclaw:runtime-dependency-ownership",
+    generateBundle: {
+      order: "post",
+      handler(_options, bundle) {
+        const chunks = Object.values(bundle).filter((output) => output.type === "chunk");
+        const ownersByChunk = new Map<string, Set<string>>();
+        for (const entry of chunks.filter((chunk) => chunk.isEntry)) {
+          const source = entry.facadeModuleId?.split("?", 1)[0];
+          const relative = source ? path.relative(rootDir, source).split(path.sep).join("/") : "";
+          // An empty owner represents a root entry and prevents plugin authorization.
+          const owner = /^extensions\/([a-z0-9][a-z0-9-]*)\//u.exec(relative)?.[1] ?? "";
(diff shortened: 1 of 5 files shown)


RomneyDa:
Compatibility proof for the generated ownership artifact:

- `dist/runtime-dependency-ownership.json` is regenerated during package build; it is not runtime or user state and requires no migration.
- Existing 2026.7.33 runtime code does not read this file, so adding it cannot change runtime or upgrade behavior.
- Older release tooling ignores the extra packaged file. The coordinated verifier in #141876 consumes the versioned chunk-and-hash contract and rejects malformed, unknown-version, or altered-byte evidence fail-closed.
- Rebuild ordering is safe: an old package remains artifact-free and follows the verifier legacy path; a rebuilt package carries the new evidence; runtime behavior is identical in either case.
- The rebuilt-package proof covers all 1,570 entries, matching emitted bytes and SHA-256 hashes, with zero dependency-manifest errors.

This is generated release-verification metadata, not a persisted application data-model change.

RomneyDa:
@clawsweeper re-review

PR #141925, opened 2026-09-08
fix(backup): preserve symbolic links to external targets

### LINKED ISSUE #141923: [Bug]: backup create fails permanently when the state directory contains any absolute symlink an asset doesn't own
**Bug type:** Product bug / behavior defect
**Beta release blocker:** No

## Summary

`openclaw backup create` aborts, publishing nothing, when the state directory contains an ordinary absolute symbolic link whose target no backed-up asset owns — a virtualenv interpreter link, a tool-cache link, an operator convenience link. The failure is permanent rather than transient: the link stays on disk, so every subsequent `backup create` fails identically. There is no flag or config key to exclude the path, while `openclaw doctor` continues to recommend the same command.

## Steps to reproduce

1. Create an absolute symlink under the state directory whose target is not itself part of the backup — e.g. a Python virtualenv, whose `bin/python3.x` is an absolute link to the interpreter that created it:

   ```
   python3 -m venv "$OPENCLAW_STATE_DIR/venv"
   ```

   A minimal equivalent: `ln -s /usr/bin/python3 "$OPENCLAW_STATE_DIR/somelink"`

2. Run `openclaw backup create --verify --json`

## Expected behavior

Creation completes and publishes an archive. An absolute link that no backed-up asset owns is omitted and reported, since for such a target "abort" and "omit" put byte-for-byte the same content in the archive — nothing. Links whose target *is* a backed-up asset continue to be rewritten to portable relative archive links.

## Actual behavior

Creation aborts and publishes nothing, naming whichever absolute link the traversal reaches first:

```
Backup archive write failed: Archive symbolic link target must be relative: <archive entry path> -> <link target> (after 1 attempt)
```

The output directory is empty afterwards.

`backup create` accepts only `--output`, `--json`, `--dry-run`, `--verify`, `--only-config`, and `--no-include-workspace` (`src/cli/program/register.backup.ts:42-47`), and there is no `backup` section in the config types — so there is no supported way to exclude the offending path. Meanwhile `src/commands/backup-health.ts:67-68` keeps emitting *"No successful backup is recorded. Create one now with `openclaw backup create`."*

## OpenClaw version

`2026.9.2 (3928bad)`. Reproduced against `upstream/main`.

## Operating system

macOS 26.6.2 (arm64)

## Install method

npm global (`npm install -g openclaw`)

## Model

Not model-dependent — the defect is in archive traversal and never reaches a model. For completeness, the host's model is `anthropic/global.anthropic.claude-opus-5[1m]`.

## Provider / routing chain

Not provider-dependent, for the same reason. For completeness: `claude-cli` runtime on Bedrock.

## Impact and severity

Any operator who has ever created a virtualenv, or run any tool that leaves an absolute link, under the state directory loses `backup create` entirely and permanently — with no diagnostic pointing at the cause and no flag to work around it.

It also has a small hardening dimension in the current direction. Because aborting is triggered by *any* absolute link, a single local write primitive inside the state directory (`ln -s /tmp/x ~/.openclaw/x`) is enough to permanently deny backups until a human finds the link. Omitting-and-reporting removes that leverage without weakening any restore-side check: `assertArchiveSymbolicLinkTarget` still runs on the write path and again over every archived link at verification, which restore performs before extracting.

## Additional information

Narrowing this rule has been attempted three times without converging:

- [`8c7e86a2ead`](https://github.com/openclaw/openclaw/commit/8c7e86a2eadad6ec444e65f5bfaac68ce0571020) (#123672) introduced the rejection.
- [`0db1d89d05c`](https://github.com/openclaw/openclaw/commit/0db1d89d05cf0d3a3275d8ddd0cad29fc41ae821) (#124821, closing #124769) excluded four managed roots after real installs hit it.
- [`0aa9ae9f3e3`](https://github.com/openclaw/openclaw/commit/0aa9ae9f3e3ee94fbc3233f07b9653ef1402e7e5) (#136343, closing #136326) remapped Nix links.

The set of absolute links a user may create under a state directory isn't enumerable, which is why per-root exclusions keep needing another entry.

This shape is also already OpenClaw policy on a sibling backup write path. `src/fleet/backup.runtime.ts` detects a symbolic link inside its tar `filter`, increments `skippedSymlinks`, and returns `false`; `src/cli/fleet-cli/commands.runtime.ts` reports the count to the operator. Fleet skips *every* symbolic link unconditionally, with no ownership test — so "omit the unarchivable link and report it" is not a new posture here, and this request is narrower than what already ships: it applies only to absolute targets that no declared asset owns, and leaves relative escapes failing closed.

Scope note: this concerns absolute targets that no declared asset owns. A **relative** link escaping the archive root or the declared assets should keep failing creation — unlike an absolute target, a relative escape would resolve to a path inside or beside the restored tree, which is the substitution hazard the guard exists to prevent.

adele-with-a-b:
## What Problem This Solves

An absolute symbolic link to an external file or directory could prevent `openclaw backup create` from publishing any archive.

Closes #141923

Credit: @adele-with-a-b for the report, reproduction, and original contribution. Recovery-policy context: @steipete and @vsumner.

## Why This Change Was Made

The backup archive policy now preserves original link targets and reports targets outside state. It replaces absolute-target rejection and relocation rewriting. Restore extracts files and creates parent directories before recreating verified links.

## User Impact

Backups complete with absolute, relative, and dangling links without copying content through those links. External targets appear in the backup summary and restore warnings. Absolute links retain their original locations, so review them before activating a relocated backup.

Existing v2026.9.4 archives remain readable. New archives with absolute or escaping link targets require the updated reader; v2026.9.4 rejects them.

## Evidence

- Before the fix, production `backup create --verify --json` failed on an external link and published no archive.
- Repaired production create, verify, and restore commands preserved literal file, directory, relative, and dangling links and reported outside-state targets, including separately backed-up assets.
- Crafted entries beneath ordinary and space-suffixed symbolic links were rejected before extraction; external content remained unchanged.
- The updated reader restored v2026.9.4 archives. The v2026.9.4 reader restored an ordinary new archive but rejected a new external-link archive before extraction.
- Recorded execution covers macOS; native Linux and Windows restoration remain unproven.

## Consumers

- Backup creation reports external symbolic links in the archive metadata and command output.
- Backup verification checks link records and rejects entries beneath links.
- Backup restore recreates verified links after file extraction and warns about external targets.


obviyus:
Thanks @adele-with-a-b for the report, reproduction, and original contribution. Your authorship is preserved in the merged commit ed0df5bf09b778cf1175aafa770be561527ee02f. This closes #141923.

- **Bug:** An external absolute symlink prevented the entire backup from being published.
- **Fix:** Backups preserve literal symlink targets, report targets outside state in the manifest and summary, and recreate verified links after extracting files. Targets are never copied through links.
- **Proof:** Production CLI create/verify/restore runs preserve file, directory, relative, and dangling links. Separately included outside-state targets remain reported. Crafted link descendants are refused before extraction. All 15 required validation lanes passed, including both real CLI process consumers.
- **Recovery limits:** Existing v2026.9.4 backups restore with the updated reader. v2026.9.4 rejects new archives containing absolute or escaping link targets. Absolute links keep their original locations; review them before activating a relocated backup.

PR #141893, opened 2026-09-08
fix(e2e): keep frozen onboarding fixtures aligned

RomneyDa:
## Summary

- keep each selected release scenario internally generation-consistent
- mount npm onboarding assertions with their imported mock-model config producer
- mount the complete selected upgrade-survivor runner/assertion/fixture directory
- run Codex doctor qualification only when the selected release shipped that contract
- accept historical zero-exit dirty rejections only after the structured assertion proves no update occurred
- keep current-target behavior unchanged

## Root cause

Merged #141774 selected individual frozen assertion files while retaining current-main producers, helpers, and the upgrade runner. That created mixed-generation scenarios:

- npm onboarding: the 2026.6.35 assertion expected `openai/gpt-5.5`, while current tooling generated `openai/gpt-5.6-luna`
- upgrade survivor: current runner invoked `assert-exec-approvals`, which the selected release assertion did not expose; the same mismatch affected ClawHub fixture expectations
- Codex on demand: current tooling invoked `codex/managed-app-server`, a doctor check the selected release never shipped

The repair uses the existing fail-closed frozen-target resolver. No version/SHA allowlist, new environment mode, or compatibility parser is added. Current and unauthorized targets continue to run the current doctor check strictly.

## Red proof

- [npm onboarding](https://github.com/openclaw/openclaw/actions/runs/34189421223/job/101944733208): expected `openai/gpt-5.5`, got `openai/gpt-5.6-luna`
- [upgrade survivor](https://github.com/openclaw/openclaw/actions/runs/34189437057/job/101944965118): unknown assertion command `assert-exec-approvals`
- [Codex on demand](https://github.com/openclaw/openclaw/actions/runs/34189437057/job/101944965138): unknown health check `codex/managed-app-server`

## Verification

- selected-contract and frozen-resolution regressions: 12 passed
- `bash -n` on all changed runners
- Oxfmt and `git diff --check`
- final AutoReview: scoped-clean, no P0/P1 findings (0.90)

Production/tooling: +38/-16 (net +22). Tests: +57/-3. Application runtime: 0.

Codex dependency gate: inspected `../codex/codex-rs/app-server/README.md:224-239`, `../codex/codex-rs/app-server-protocol/src/protocol/common.rs:705,812`, and `../codex/codex-rs/app-server/src/request_processors/plugins.rs:1481-1908`. Upstream owns `plugin/install`; the failing `codex/managed-app-server` doctor identifier is an OpenClaw release qualification contract, not an upstream Codex API.

PR #141930, opened 2026-09-08
fix(plugins): show compatibility notices after raw plugin loading

steipete:
## What Problem This Solves

Fixes an issue where operators checking plugin compatibility could miss notices about registered legacy hooks after a plugin had been loaded from raw options. The runtime registration existed, but a later prepared status read did not recognize it as the same active plugin selection.

## Why This Change Was Made

The loader now records the resolved physical selection identity alongside its original request identity after manifest discovery. Compatibility reads may use that loader-owned identity only while the original request identity still matches the active registry. Metadata rebinding cannot replace the captured identity, and different source selections or stale active keys remain incompatible.

Raw cache lookup, in-flight tracking, activation, cache capacity, and retirement behavior remain unchanged. This adds no cache aliases or cross-representation load deduplication.

## User Impact

Status and doctor can report compatibility notices from the already active plugin registry without loading cold plugin modules or accepting a registry from a different physical source selection.

## Evidence

- The unchanged status integration suite went from 2/3 passing to 3/3 passing; the original failure was the missing hook-only notice.
- Six focused owning suites passed 90/90 cases with no skips. The existing CJS/TS cache test also checks repeated raw cache reuse followed by prepared reads, physical-source mismatch, metadata rebinding, and stale active-key rejection.
- After fixing a test-only map-spread lint finding with an equivalent fresh clone, all 23 cases in the affected owner suite passed again.
- An external diagnostic compared extracted actual old/new key builders across 128 request-key cases and 128 captured-fact mutation cases. All matched; this complements the real loader tests rather than replacing them.
- The full mapped changed gate, `pnpm build`, and fresh independent P2 review passed.

The first CI run exposed an incomplete whole-module mock in the prepared model catalog test that leaked into a later real transport test in the shared worker. A separate focused commit replaces that invalid fixture with the existing complete metadata builder and removes the mock. All 11 producer assertions and the transport test remain unchanged. The original producer→consumer order reproduced the failure; the combined source now passes both cases in that order in one shared worker. Normal project routing also passed in the fixture owner proof. Fresh combined P2 review and the proportional changed gate passed.

The combined PR adds 28 production lines and 21 test lines relative to current main. No runtime improvement is claimed.

Current main later changed the same incomplete mock to a partial mock in #141893. The merge conflict was resolved by retaining the reviewed complete fixture, which removes the mock altogether and preserves the upstream export-availability fix. The original ordered pair, seven-file changed gate, and fresh combined P2 review pass after that resolution; all seven reviewed source files remain byte-identical.


steipete:
Merged via squash.

- Prepared head SHA: [8b8820ff9e0cbab7109835e1223598c6e9ce7ce4](https://github.com/openclaw/openclaw/pull/141930/commits/8b8820ff9e0cbab7109835e1223598c6e9ce7ce4)
- Landed commit: [db25ae4e56e978e98ddef4fbc425563b34f3ef6f](https://github.com/openclaw/openclaw/commit/db25ae4e56e978e98ddef4fbc425563b34f3ef6f)

PR #128793, opened 2026-08-24
fix(chat): show native commentary progress

obviyus:
## Problem

Native phased commentary was emitted as assistant events, while Control UI progress consumes keyed preamble items. Tool calls appeared, but commentary around them disappeared live and after history reload.

## Root cause

Commentary had consumer-specific representations instead of one presentation contract. Gateway history also removed commentary before the UI could reconcile it. The first history fallback implementation then omitted transcript sequence metadata, which could strand older cursor pages.

## Fix

- Normalize native commentary once into cumulative keyed preamble items at embedded reply delivery.
- Route channels and Control UI through the existing item path; remove the channel-only adapter.
- Project keyed commentary into bounded chat-history fallbacks while preserving source transcript metadata and structured trace.
- Keep exact phased transcript content for model replay and prompt-context consumers.

This exposes explicit provider commentary/preambles, not hidden chain-of-thought.

## Proof

Exact head: `68ce4369635f7b19d1e250b526ab5171bf2c97b3`

- owner regression without the metadata fix: `hasMore: true`, `nextCursor: undefined`
- same regression with the fix: cursor `2`; older page reachable
- `pnpm check:changed`
- production tsgo + `pnpm check:test-types`
- 499 Gateway tests
- 4 cron prompt-context tests
- 25 auto-reply tests
- 41 embedded-agent tests
- 21 embedded-Gateway tests
- 4 Control UI E2E tests; exact-head [before/after screenshots](https://github.com/openclaw/openclaw/pull/128793#issuecomment-5399042361)
- `$distill`: one normalized metadata representation; no extra compatibility path
- `$greybeard`: bounded linear projection scan; no new I/O or repeated transcript scan

Production LOC: +134/-56, net +78. Growth owns the missing generic live/history presentation contract and transcript identity; the channel-only adapter was removed. Tests: +374/-72.

## Exact-head review

ClawSweeper local review: PASS — patch correct; 0 findings; 0 security concerns; no maintainer decision; 0 Rank-up moves.

## Telegram proof

Final exact-head real-user DM used the deterministic two-turn OpenAI commentary fixture with progress commentary enabled.

- 2 provider requests
- 4 typing events
- temporary progress showed `Working`, `Checking the workspace before answering.`, and `Exec blocked`
- persistent final contained `OPENCLAW_E2E_DRAFTPROOF`
- temporary progress was deleted after final delivery; no intermediate Telegram edit was observed

[...]

PR #141932, opened 2026-09-08
fix(agents): keep commentary line breaks in preamble progress text

leapdragon:
## What Problem This Solves

Fixes a very visible rendering glitch in the Control UI WebChat during tool-calling turns. The assistant's text streams in correctly formatted (paragraphs, bullet lists, tables), and then at the moment a tool call begins it snaps into one long paragraph. Bold text and links survive, but every line break, list marker, and table row is flattened into a single line.

## Why This Change Was Made

When the provider resolves commentary at the tool boundary, `createReplyDelivery` in `src/agents/embedded-agent-subscribe.reply-delivery.ts` emits the commentary-phase text as a keyed `preamble` item whose `progressText` had all whitespace collapsed to single spaces (`text.replace(/\s+/g, " ").trim()`). The WebChat's `handlePreambleProgress` then takes that item as the owner of the text it had already streamed and renders `progressText` as Markdown in its place. Markdown without newlines has no paragraphs, list items, or table rows, so the block structure disappears exactly when the item replaces the stream.

The collapse was introduced with the native commentary progress work (#128793). No consumer needs it: the CLI runner path already emits multi-line commentary as `progressText`, the channel progress compositors (`pushPreambleHeadline`, `pushCommentaryProgress`) and the session-observer digest normalize their own single-line headlines, and the WebChat and history snapshot store the text verbatim. This change sends the trimmed text with its line breaks intact.

Non-goal: this does not change how commentary is phased, deduplicated, or persisted; the commentary signature used to suppress duplicate emissions still keys on the (now uncollapsed) text.

## User Impact

Pre-tool commentary in the WebChat keeps its paragraphs, lists, and tables through the tool boundary instead of collapsing to a single line. Other channels are unchanged because they already normalize the text themselves.

## Evidence

The existing "preserves aggregate commentary for providers without Responses item boundaries" expectation encoded the collapse (`"First. Second."`) and now expects the actual `\n` join. A new test streams multi-line Markdown commentary (paragraphs and a bullet list) through an Anthropic-style tool handoff and asserts both the update and end preamble events carry the text unchanged. Both fail without the fix.

```
node scripts/run-vitest.mjs run src/agents/embedded-agent-subscribe
  Test Files  34 passed   Tests  866 passed
node scripts/run-vitest.mjs run src/gateway/server-chat src/gateway/session-observer src/auto-reply/reply/agent-runner-execution-progress.test.ts
  Test Files  8 passed    Tests  348 passed
node scripts/run-oxlint.mjs <changed files>   # clean
oxfmt --check <changed files>                 # clean
```

Observed on the 2026.9.2 release with an `openai-completions` transport driving tool-calling turns in the WebChat.

https://claude.ai/code/session_011LRND53x3eTxf5wPGYVFro


JosephNatsu:
I reproduced this independently on current upstream main `68b7893b3ca977a238af165717f7bc2a40e1a8f0` and tested the exact production change proposed here (`eventData.text.trim()`).

The repro sends a synthetic tool-handoff preamble containing a fenced two-line command block and a nested list through the real embedded-agent reply-delivery producer, then forwards its emitted commentary into the built Control UI using the repository's mocked-Gateway Chromium harness.

**Before:** the producer collapses whitespace, so the UI receives a single line; no `pre > code` block is rendered.

Before: commentary loses code-block newlines and list indentation

**After this PR's change:** the code block retains both lines and the nested list renders correctly.

After: multiline commands and nested list are preserved

Verification:
- Producer regressions cover both `anthropic-messages` and `openai-completions` API routes. On the unpatched source, the aggregate-newline expectation and both structured-Markdown cases fail; with this change, 54 tests across the three focused reply-delivery suites pass.
- The browser regression checks the actual rendered code text and nested-list element. It fails on the unpatched source and passes with this change.

Scope of this evidence: **real producer + real built Control UI + mocked Gateway**, not a live model-provider or Telegram run. The model name visible in the synthetic browser fixture is just the harness's default UI label; no model/API was called. I have not verified Telegram behavior and am not asserting that other channels are unchanged.

This matches a problem we had patched in our downstream 8.2 build. Sharing the WebChat evidence here instead of opening a duplicate PR.

VACInc:
#146554 is not a fix for this PR. #146554 changes when live text is retired around tool activity; this PR addresses whitespace/newline loss when commentary becomes keyed preamble progress text. The symptoms can co-occur, but the changed owner, invariant, and regression assertions are different.

PR #141935, opened 2026-09-08
fix(release): scope status CPU allowance to 2026.7.33

RomneyDa:
Supersedes: #141928
Related: openclaw/Kova#116
Companion Kova revert: openclaw/Kova#117

## What Problem This Solves

Full Release Validation for OpenClaw 2026.7.33 can fail the status CLI CPU gate even though current OpenClaw versions remain below Kova's restored 200% ceiling.

## Why This Change Was Made

The performance workflow now owns the compatibility mapping beside its existing reviewed Kova pins. It inspects the frozen target's `package.json` and selects the calibrated openclaw/Kova#116 revision only for exactly 2026.7.33. Explicit `kova_ref` inputs still take precedence.

Full Release Validation no longer contains or passes a duplicate SHA, and no 2026.7.33-specific behavior is added to Kova main. Every other target continues using the normal fixture-compatible Kova pin and restored 200% ceiling.

## User Impact

Release validation can finish for 2026.7.33 without weakening CPU regression detection for current releases or freezing unrelated Kova development.

## Evidence

- Performance and extended-stable workflow tests: 45/45 passed
- `oxfmt --check` passed
- `git diff --check` passed

### DIFF
--- a/.github/workflows/openclaw-performance.yml
+++ b/.github/workflows/openclaw-performance.yml
@@ -119,7 +119,9 @@ jobs:
         with:
           ref: ${{ inputs.target_ref || github.sha }}
           path: .artifacts/performance-target
-          sparse-checkout: src/config/zod-schema.agent-defaults.ts
+          sparse-checkout: |
+            package.json
+            src/config/zod-schema.agent-defaults.ts
           sparse-checkout-cone-mode: false
           fetch-depth: 1
           persist-credentials: false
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper re-review

Addressed the P1 on current head: candidate `package.json` is now read as text and parsed with `JSON.parse`, never loaded as a module. Added a regression using symlinked executable metadata and verifying no side effect occurs.

RomneyDa:
@clawsweeper re-review

Current head addresses the P1 by parsing candidate metadata as JSON text rather than loading it as a module. The compatibility SHA is now inline once at the exact-version decision point.

PR #141537, opened 2026-09-07
fix: surface subagent worktree setup failures and stop per-blob fetches during worktree sizing

steipete:
Supersedes #141504 (native merge-commit request was rejected by repository policy; identical head).

## What Problem This Solves

Fixes an issue where a spawned child session whose managed worktree could not be prepared would show only a warning glyph in the Control UI sidebar, with "Unknown error" on hover, an empty chat pane, and no trace of the failure in its transcript. The actual reason lived only in the subagent run record. On a blobless partial clone the preparation itself failed because the disk-space size estimate fetched every missing blob one network round trip at a time until the two-minute Git budget expired, and its hint blamed disk space.

## Why This Change Was Made

Failed and timed-out child runs now project their bounded reason into the session entry and append the same per-run "This turn did not run" transcript notice the gateway already uses, through one shared session-layer formatter. The chat.send dispatch-error owner now publishes its fallback terminal only after its lifecycle write settles, because publishing first let the registry record a later start time and the gateway rejected its own failure as stale. Command-execution errors keep their command and termination reason instead of being rendered as provider timeouts. The worktree size estimate inventories missing objects without lazy fetching, hydrates them from the promisor remote in one batch under the checkout budget, and forbids per-object lazy fetches during sizing. The Git timeout hint reports the applied budget and points to reachability, locks, and clone shape. Non-goals: no UI changes (the UI already renders `lastRunError`), no changes to provider failover classification, no schema or config changes. It also realigns the QA scenario selector for the credentials E2E test renamed by #141514, which was failing on main.

## User Impact

Operators see why a child session failed before its first reply: the sidebar warning carries the reason, the chat pane shows the error, and the transcript keeps a durable notice. A Git setup timeout now reads as a Git timeout with the failed command, not "LLM request timed out." Worktree creation from partial clones no longer stalls on per-blob fetches.

## Evidence

Production incident (team Gateway, 2026-09-07): child session failed after 123 s in `git ls-tree -r --format=%(objectsize)` on a shallow blobless clone; entry had `status: failed` and no `lastRunError`; measured lazy fetch cost ~0.3 s per missing blob, 0.7 s for the full inventory once objects are local.

Regression tests, each shown to fail on the pre-fix code for the intended reason:
- `src/agents/worktrees/capacity.test.ts`: real `--filter=blob:none` clone, one batch prefetch, no fetch when nothing is missing, actionable no-promisor error, `GIT_NO_LAZY_FETCH=1` on the inventory.
[...]

PR #141936, opened 2026-09-08
fix: snapshot restore fails silently on shallow clones and doctor now flags partial project clones

steipete:
Related: #141537.

## What Problem This Solves

Fixes two gaps exposed when a registry-owned project clone on the team Gateway was turned into a shallow, blobless partial clone by agent-run `git fetch --filter=blob:none --depth` commands. Restoring a managed-worktree snapshot whose commit had become shallow failed with git's raw "unknown revision" text, and nothing in OpenClaw reported that the clone had that shape until the next failure.

## Why This Change Was Made

Snapshot restore keeps its normal parent lookup; only when that lookup fails does it check whether the repository is shallow and the snapshot commit exists, and it then reports the repository, the snapshot commit, and the recovery step (`git fetch --unshallow`, or recovering the parent from the original repository) instead of the raw git error. A new doctor check inspects registry-owned cloned projects with three bounded read-only git reads, warns when any promisor or partial-clone filter key (including URL-keyed twins) or shallow state is present, and prints the exact manual repair sequence. Doctor never runs the network-heavy repair itself, adds no CLI or config option, and skips unreadable roots with a warning. Docs for managed worktrees and doctor describe both. Non-goals: an exec-policy warning when an agent runs a filtered fetch inside a registry clone is left as a follow-up; project cloning is unchanged.

## User Impact

Operators learn that a project clone is partial or shallow from `openclaw doctor`, with a copyable repair sequence, before it breaks a session. A snapshot restore that cannot proceed says why and what to do.

## Evidence

Real git fixtures: a bare origin with a two-commit history, `--depth=1` and `--filter=blob:none` clones, a URL-keyed promisor twin, an ignored non-registry workspace, and a missing clone. The restore regression reproduces the raw "unknown revision" failure by making the snapshot itself shallow with `git fetch --depth=1` and fails on the pre-fix code for that reason; the doctor tests verify every finding shape and execute the printed repair commands against a local origin. `node scripts/check-changed.mjs` passed; focused suites pass. Codex-backed autoreview at P1 before landing. Team host: the affected clone was normalized by hand with the same sequence on 2026-09-07.

## Known Gaps

Restore reports rather than deepens: origin cannot deepen a local-only snapshot that a later fetch made shallow. Doctor detection needs a stored registry row; workspace directories that are not registry clones are intentionally not inspected.

### DIFF
--- a/docs/cli/doctor.md
+++ b/docs/cli/doctor.md
@@ -16,6 +16,11 @@ When Gateway status reports degraded SecretRef owners, doctor prints a **Secret 
 
 When channel ingress events are dead-lettered, doctor names each affected channel account and points to [`openclaw channels dead-letters list`](/cli/channels#inbound-dead-letters) for inspection and recovery.
 
+Doctor warns when a registry-owned project clone is partial or shallow. It names
+the clone, shallow state, and partial-clone config keys, including URL-keyed
+remote twins. It prints manual repair commands; `--fix` does not fetch or repack
+these clones. Agent workspaces and manually registered checkouts are excluded.
+
 When the Gateway has exporter health facts, doctor reports the latest trusted
 per-signal state and transport under **Telemetry exporters**. The summary is
 redacted and does not include endpoint values, headers, certificates, payloads,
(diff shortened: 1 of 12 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [cb608a5a7c326b87b3f30bf3b8e3306020354233](https://github.com/openclaw/openclaw/pull/141936/commits/cb608a5a7c326b87b3f30bf3b8e3306020354233)
- Landed commit: [9911c71a5734d6be87aaba41cb20e40dad67e9b2](https://github.com/openclaw/openclaw/commit/9911c71a5734d6be87aaba41cb20e40dad67e9b2)

PR #141938, opened 2026-09-08
fix(llm): retain prepared runtimes through direct completions

steipete:
## What Problem This Solves

Overlapping direct plugin LLM completions rebuild prepared model facts because preparation releases its lease before execution finishes. In Gateway mode, that also makes an active completion's owner eligible for LRU eviction. A request-store construction failure could instead leak the counted lease.

## Why This Change Was Made

An internal acquisition function transfers the existing lease to the direct `runtime.llm.complete` operation, which releases it after completion and result processing. Failed construction or preparation releases internally. The public SDK preparation adapter retains its existing return shape and immediate-release behavior.

Related: #140674. This eight-file slice changes two production owners and their test fixtures. It adds no physical database/registration disposal, cache key or size, SDK export, configuration option, or protocol change. Generation scope still covers preparation only; isolated and worker inference retain their existing outer owners.

## User Impact

Overlapping completions reuse their prepared generation while keeping independent mutable request stores. Configuration, credential publication, policy checks, provider errors, cancellation, and returned results retain their behavior.

## Evidence

The original production code failed three native cases: overlap reuse, in-use LRU eviction, and a store-fork failure leaking ownership. Six configuration/authentication/error/abort controls passed. All nine pass with the repair, alongside 129 runtime/command tests, 49 core/SDK compatibility tests, and the complete canonical changed checks. Codex review found no actionable findings. The full build passed on the identical production bytes; subsequent corrections were test-only.

Matched compiled probes called the unchanged `createPluginRuntime().llm.complete` entry against a local synthetic SSE provider. Four separate Node 24 processes ran in baseline/candidate/candidate/baseline order, each with a warm-up and five gated request pairs. Pass-through observers preserved the original public registry methods and receivers.

| --- | ---: | ---: |
| Generation builds per overlapping pair | 2 | 1 |
| Independent request-store forks per pair | 2 | 2 |
| Median second-request preparation, 10 samples | 5.94 ms | 4.24 ms |

Preparation timing ends when the second provider request arrives while the first response is held. These are local synthetic observations, not an end-to-end inference speed guarantee. Source and compiled artifacts stayed unchanged throughout the comparison.

The first CI run exposed one remaining completion mock using the old internal entrypoint. Its typed fixture now follows the acquisition contract while preserving the refusal-before-acquisition assertion; all 13 overflow cases pass. CI also exposed a pre-existing partial metadata mock leaking across shared-worker tests. The unchanged base reproduced that failure, and the same ordered worker passed after replacing the mock with the existing complete fixture. Main independently received that repair in #141930; this PR was rebased and carries no metadata-test change. Production code and the compiled benchmark are unchanged.

### DIFF
--- a/src/agents/embedded-agent-runner/run.overflow-compaction.test.ts
+++ b/src/agents/embedded-agent-runner/run.overflow-compaction.test.ts
@@ -38,7 +38,8 @@ afterEach(() => {
   }
 });
 const completionMocks = vi.hoisted(() => ({
-  prepareSimpleCompletionModelForAgent: vi.fn(),
+  acquireSimpleCompletionModelForAgent:
+    vi.fn<typeof import("../simple-completion-runtime.js").acquireSimpleCompletionModelForAgent>(),
   completeWithPreparedSimpleCompletionModel: vi.fn(),
   resolveSimpleCompletionSelectionForAgent: vi.fn(),
 }));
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a87f5051325db047df6da1fcba5d501bc9d0f29d](https://github.com/openclaw/openclaw/pull/141938/commits/a87f5051325db047df6da1fcba5d501bc9d0f29d)
- Landed commit: [e4ee9d681e142c5171ffcf73b27c3d588c0808bd](https://github.com/openclaw/openclaw/commit/e4ee9d681e142c5171ffcf73b27c3d588c0808bd)

PR #141939, opened 2026-09-08
fix(canvas): reject show_widget scripts that do not parse

steipete:
## Problem

`show_widget` hosted whatever `widget_code` the model sent. When an inline `<script>` had a syntax error, the widget was still hosted and rendered as a dead, non-interactive card, and the model got no signal that anything was wrong.

Real case on a team Gateway: the agent built the widget HTML inside a JavaScript template literal, `\n` inside a single-quoted string became a raw line break, the browser reported `Unexpected identifier 'os'`, and the first card was inert. The agent only noticed after a manual `vm.Script` check, fixed the escaping, and called `show_widget` again. The transcript ended up with two widgets: a broken one and a working one. Inline widgets have no replace semantics, so the first card could not be retracted.

## Solution

Core now parses every inline JavaScript `<script>` in HTML `widget_code` with acorn before any side effect and rejects the call with the script index, line, column, and offending line when the first one fails to parse. A widget with a broken script is never hosted, pinned, or presented on any surface.

- `src/canvas/widget-script-syntax.ts`: a small position-tracking HTML tokenizer that skips comments, honors quoted attributes, treats `style`/`textarea`/`title`/`xmp`/`iframe`/`noembed`/`noframes`/`noscript` as raw text, ends script bodies per the script-data escaped and double-escaped states (so `"<!--<script></script>-->"` inside a string does not end the element), and handles SVG foreign content (CDATA-wrapped scripts, opaque CDATA sections, self-closing elements, `foreignObject` back to HTML rules). Classic vs `type="module"` after entity-decoding the attribute; scripts with `src` or a non-JavaScript `type` are skipped per the HTML spec JavaScript MIME essence list. Acorn parse with `ecmaVersion: "latest"`; the error location is mapped back into the untrimmed `widget_code`.
- `src/canvas/widget-tool.ts`: gate wired right after the size check for `kind === "html"` and before pinning, presenters, or canvas document creation. Registered content kinds keep their own `validateSource`; reports are untouched.
- Tool description: "Update HTML by name" now reads "Update pinned HTML by name" (name only affects pinned widgets; the unqualified wording is what led the model to expect an inline replace), plus one sentence stating that inline scripts must parse.
- Docs: `docs/tools/show-widget.md` describes the validation in the "How widgets work" section and the `widget_code` parameter.

Runtime errors (a typo'd element id, a handler that throws on click) are out of scope here; they need a client-side error bridge and will be a separate change.

Declined review findings, recorded in the module doc comment: character references inside SVG script text, CDATA sections interleaved with script text, and SVG nested inside `foreignObject` are not modeled. None of these appear in realistic widget markup, and in each case the gate fails loudly with an actionable error instead of hosting a widget whose script cannot run. Inline widgets still have no replace-by-name semantics; that stays a follow-up.

## Impact

Model-facing: a broken widget becomes a tool error with an actionable location instead of a silently dead card. No config, schema, CSP, size, retention, or dashboard behavior changes. Valid JavaScript is unaffected; the parser is the same acorn build already used for code-mode scripts.

## Evidence

- `node scripts/run-vitest.mjs src/canvas/widget-script-syntax.test.ts src/canvas/widget-tool`: 8 files, 148 tests passed. The four routing regression tests fail without the gate (`promise resolved ... instead of rejecting`).
- `pnpm check:changed`: all lanes green on the final tree, including typecheck core, core tests, boundary, format, and the production unused-export scan.
- Five Codex autoreview rounds; accepted findings (MIME list, tokenizer context, end-tag forms, hashbang, escape states, SVG CDATA, entity-decoded `type`, self-closing foreign elements, `foreignObject`) are all in; the three declined ones are listed above.
- Live proof on an isolated token-auth Gateway built from this branch (`pnpm build`, `OPENCLAW_STATE_DIR`/`OPENCLAW_CONFIG_PATH` scratch dirs, `anthropic/claude-sonnet-4-6`), driven through the Control UI with Playwright:
  1. Agent instructed to call `show_widget` with the exact original bug shape (raw newline inside a single-quoted string). Tool returned `widget_code has a JavaScript syntax error in inline script 1 at line 3, column 14: Unterminated string constant. Offending line: const label = 'sessions.dispatch. ...`. No canvas document was created and no widget card rendered.
  2. Agent asked to escape the newline and retry. Widget hosted; the sandboxed frame's text read `sessions.dispatch {"os":"linux"}`, proving the script executed.

Rejected broken script, then the corrected widget rendering with its script executed

### DIFF
--- a/docs/tools/show-widget.md
+++ b/docs/tools/show-widget.md
@@ -16,7 +16,7 @@ For a pinned data report, provide a structured `report` with `pin: true`. Report
 
 ## How widgets work
 
-For HTML widgets, OpenClaw core validates `widget_code` and wraps it once in the canonical HTML document. For an inline client, core stores that document as a Canvas document and returns a preview handle. The Control UI reads the document over its authenticated Gateway connection and renders it through the dedicated-origin, double-iframe sandbox used by dashboard widgets and MCP Apps. The widget frame does not need its own login session. iOS, Android, macOS, and Linux Quick Chat use isolated web views. Full chat clients restore the widget after history reload; Quick Chat keeps the widget for its active reply.
+For HTML widgets, OpenClaw core validates `widget_code` by parsing every inline JavaScript `<script>` (classic and module), skipping scripts with `src` or a non-JavaScript `type`, then wraps it once in the canonical HTML document. Core rejects the call with the line and column of the first syntax error, so a widget with a broken script is never hosted. For an inline client, core stores that document as a Canvas document and returns a preview handle. The Control UI reads the document over its authenticated Gateway connection and renders it through the dedicated-origin, double-iframe sandbox used by dashboard widgets and MCP Apps. The widget frame does not need its own login session. iOS, Android, macOS, and Linux Quick Chat use isolated web views. Full chat clients restore the widget after history reload; Quick Chat keeps the widget for its active reply.
 
 Channel plugins can register a contextual presenter behind the same core tool. In a configured Discord session, core hands the composed document to the Discord presenter, which stores it and posts the Activity button in the current channel. The model still makes one `show_widget` call; there is no transport-specific widget tool or content kind.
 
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3ef2b7f3fb5c1f34ccb414c3b2e5e0fe312ab5d1](https://github.com/openclaw/openclaw/pull/141939/commits/3ef2b7f3fb5c1f34ccb414c3b2e5e0fe312ab5d1)
- Landed commit: [2d6dcabc3ecfa90cf9b44cb22922f4f2d31b0dff](https://github.com/openclaw/openclaw/commit/2d6dcabc3ecfa90cf9b44cb22922f4f2d31b0dff)

PR #141944, opened 2026-09-08
refactor(media): exercise quota through the cache writer

steipete:
## What Problem This Solves

The playback-cache eviction test called a private quota hook, so it could pass even if writing a new rendition stopped enforcing the cache budget.

## Why This Change Was Made

Exercise the existing cache writer directly, then remove the unused private enforcement wrapper and its three test-global members. The fixture starts exactly at an independently stated 512 MiB budget; inserting eight bytes must evict the oldest file before the writer resolves. The retention case keeps checking the fixed seven-day policy independently of production constants.

## User Impact

Runtime cache policy and public APIs are unchanged. Both existing cases and all five original expectations remain, with an inserted-content check added. Production code shrinks by eight lines and tests by eleven. The network-test hook and the shared publication remain intact.

## Evidence

- Original and final store/cache/transcode/staging suites: 152 passed each, with identical case inventories.
- Controlled removal of only the writer's quota call: old tests still pass all 152; migrated tests fail exactly the oldest-file eviction assertion, while 151 pass.
- Correct code restored before final proof; no control change remains.
- Independent Codex review through P2 is clean; formatting and diff checks passed.
- All 29 normal changed-file checks passed, including production and core-test types, dead-export scans, scoped lint, and remaining guards ([Linux Testbox run](https://github.com/openclaw/openclaw/actions/runs/34193344856)). The one-shot lease, temporary capsule, and recorded processes are cleaned up.

These tests use isolated temporary state and synthetic media. No operator cache, live provider, or actual ffmpeg process was exercised by this cleanup. Sparse-file logical sizes are known; physical allocation was not measured.

### CI integration

The first exact-head CI run failed because an earlier catalog-worker test left a partial registry mock behind. The same missing-export failure reproduced on the base without this media patch. Main now contains the canonical fixture repair in #141930; the branch was refreshed onto it with the two-file media patch byte-identical. All 152 media/store/transcode cases passed again after that refresh.

An ordered control confirmed that the canonical repair changes the failing sentinel integration to passing. That local cohort also exposed a separate keyless-auth fixture failure under ambient credentials, retained for follow-up; it is not reported as a wholly green cohort. The earlier 29-check gate remains evidence for the unchanged media patch, and fresh exact-head CI is required on this revision.

### DIFF
--- a/src/media/store.playback-cache.test.ts
+++ b/src/media/store.playback-cache.test.ts
@@ -6,22 +6,6 @@ import { createTempHomeEnv, type TempHomeEnv } from "../test-utils/temp-home.js"
 let store: typeof import("./store.js");
 let tempHome: TempHomeEnv;
 
-type PlaybackCacheTestApi = {
-  enforcePlaybackTranscodeCacheLimit(): Promise<void>;
-  PLAYBACK_TRANSCODE_MAX_CACHE_BYTES: number;
-  PLAYBACK_TRANSCODE_TTL_MS: number;
-};
-
-function getPlaybackCacheTestApi(): PlaybackCacheTestApi {
-  const api = (globalThis as Record<PropertyKey, unknown>)[
-    Symbol.for("openclaw.mediaStoreTestApi")
-  ];
-  if (!api) {
-    throw new Error("media store test API is unavailable");
-  }
-  return api as PlaybackCacheTestApi;
-}
-
 beforeAll(async () => {
   tempHome = await createTempHomeEnv("openclaw-playback-cache-");
   store = await import("./store.js");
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9f155bc4214247928f33355e32c678ecce6e8743](https://github.com/openclaw/openclaw/pull/141944/commits/9f155bc4214247928f33355e32c678ecce6e8743)
- Landed commit: [dc9421a09bf072713cec560b9c2d3b041078c368](https://github.com/openclaw/openclaw/commit/dc9421a09bf072713cec560b9c2d3b041078c368)

PR #141946, opened 2026-09-08
fix(agents): retry exec auto-review once when the endpoint requires reasoning

leapdragon:
## What Problem This Solves

The model-backed exec auto-reviewer (`tools.exec.reviewer.model`) fails closed on every request when it is pointed at a reasoning-only endpoint. With OpenRouter's `openai/gpt-oss-safeguard-20b`, a model built for exactly this kind of policy review, each review returned:

```
exec reviewer completion failed: 400 Reasoning is mandatory for this endpoint and cannot be disabled.
```

Every auto-reviewed command fell back to human approval, so the reviewer did no work while looking configured.

## Why This Change Was Made

`createModelExecAutoReviewer` calls the simple-completion runtime with `maxTokens` and `temperature` only and never sets a reasoning level. The OpenAI-compatible completions transport maps an unset level to the model's off-effort, and for OpenRouter-format reasoning models that becomes a literal `reasoning: { effort: "none" }` in the payload. Models that allow reasoning to be disabled accept it; reasoning-only models reject the request outright.

The compaction path already handles this case: when a provider reports reasoning as mandatory it retries once at `minimal` via `pickFallbackThinkingLevel`. The reviewer had no equivalent. This change gives it one:

- On a completion failure whose message matches `isReasoningConstraintErrorMessage`, retry the review once with `reasoning: "minimal"`.
- The retry only runs when `tools.exec.reviewer.thinking` (#147192) is unset. An operator who configured a level, including `off`, asked for exactly that request and gets the provider's answer unchanged.
- The retry reuses the reviewer's abort signal and gets a fresh reviewer timeout, matching how the existing code already treats preparation and completion as separate timeout windows.
- A second failure returns the existing "exec reviewer completion failed" ask decision; nothing loops.

No behavior changes for reviewers whose first request succeeds, and no config is added.

## User Impact

Operators can point the exec reviewer at reasoning-only models without every review degrading to a manual prompt. Failure handling is unchanged: an error other than a reasoning constraint, or a second reasoning-constraint error, still produces an "ask" decision with the provider message.

## Evidence

New file `exec-auto-reviewer.reasoning-fallback.test.ts` (the existing suite is at the 1000-line lint cap):

- "retries once with minimal reasoning when the endpoint requires reasoning" asserts the first call carries no `reasoning` option, the second carries `"minimal"`, and the parsed allow decision is returned.
- "does not retry a reasoning-constraint failure more than once" asserts exactly two calls and the ask decision with the provider message.
- "does not retry when a reviewer thinking level is configured" asserts a single call carrying the configured level and the ask decision.

```
node scripts/run-vitest.mjs run src/agents/exec-auto-reviewer.test.ts src/agents/exec-auto-reviewer.reasoning-fallback.test.ts
  Test Files  2 passed   Tests  87 passed
pnpm tsgo:core   # clean
oxlint src/agents/exec-auto-reviewer.ts src/agents/exec-auto-reviewer.reasoning-fallback.test.ts   # clean
```

Observed on 2026.9.2 with `tools.exec.reviewer.model` set to `openrouter/openai/gpt-oss-safeguard-20b`: every exec approval surfaced the 400 above in the Guardian review and fell back to a manual prompt.

https://claude.ai/code/session_011LRND53x3eTxf5wPGYVFro


VACInc:
PR #147192 is now merged, but this remains a complementary bounded retry change for mandatory-reasoning cases. Please keep this PR open for its distinct scope.

PR #141954, opened 2026-09-08
fix(release): route Gateway load and preserve assertion ownership

steipete:
## What Problem This Solves

Resolves two release-validation problems: the Gateway worker-load case exceeds its unchanged deadline on GitHub-hosted capacity, and a blanket frozen-target assertion mount replaces the trusted runner's serving-turn oracle with an older dispatcher that does not implement it.

## Why This Change Was Made

Add a default-true Gateway-specific hosted-runner option, combined with the existing global option. Only the release caller disables it. The existing full Gateway build and four test shards use `blacksmith-32vcpu-ubuntu-2404`; other UI, plugin, and live routing is unchanged. This moves five existing registrations per campaign, with at most four Gateway test runners, without adding jobs or changing workers, coverage, deadlines, or artifact identity.

Select upgrade assertion ownership from the immutable target package version through the existing release-train classifier after source validation. Extended-stable targets retain the shipped scenario directory and runner, including assertions and fixtures; regular targets retain the coherent trusted scenario. Historical baselines and dirty working-tree metadata do not choose the owner. Invalid versions and unsupported extended-stable corrections fail before Docker.

## User Impact

Release validation gets the measured capacity needed for the existing Gateway workload and preserves both regular serving-turn proof and extended-stable state contracts. Release Code and test thresholds remain unchanged.

## Evidence

- Matched diagnostic test/cohort/stages: GitHub-hosted timed out at 240.261 seconds; the existing Blacksmith profile completed the target in 67.140 seconds, including all three 12-turn waves, control latency checks, restart, and cleanup. The latter reported 8 CPUs and about 31 GiB RAM; the configured label is not a claim of 32 measured CPUs. Both overall diagnostic tails were cancelled after the target finished, so this is routing evidence, not full-cohort or release qualification.
- 220 full release/upgrade owner tests, 18 focused track/serving cases, five routing cases, and the existing repo-build workflow guard pass. The original routing and blanket-mount failures were reproduced before their fixes.
- Direct proof against the exact frozen Code checkout selects the trusted regular oracle and produces a completed, persisted synthetic serving reply. Extended-stable runner/directory mounts, baseline independence, immutable metadata, and invalid-version rejection are covered. The bounded main conflict resolution preserved the upstream whole-scenario ownership improvement and passed 18 policy/serving cases plus a fresh independent review.
- Workflow validation, scoped type/script/format checks, and final targeted lint pass. Independent P2 review is clean.
- Broad Docker-helper testing: 277 passed, two platform skips, and six unchanged macOS tar-stream fixture failures before their self-upgrade RPC checks. Those failures were recorded, not fixed or skipped; required hosted CI remains a landing gate.

No frozen release ref, validation plan, artifact, or publication surface is changed. A replacement Tooling selection and fresh qualification remain explicit release-owner actions.

### DIFF
--- a/.github/workflows/openclaw-live-and-e2e-checks-reusable.yml
+++ b/.github/workflows/openclaw-live-and-e2e-checks-reusable.yml
@@ -109,6 +109,11 @@ on:
           - beta
           - stable
           - full
+      gateway_repo_e2e_use_github_hosted_runners:
+        description: Use GitHub-hosted Gateway repo E2E when the overall hosted-runner option is enabled
+        required: false
+        default: true
+        type: boolean
       use_github_hosted_runners:
         description: Use GitHub-hosted runners where supported, including OpenWebUI
         required: false
@@ -461,6 +466,11 @@ on:
         required: false
         default: stable
         type: string
+      gateway_repo_e2e_use_github_hosted_runners:
+        description: Use GitHub-hosted Gateway repo E2E when the overall hosted-runner option is enabled
+        required: false
+        default: true
+        type: boolean
       use_github_hosted_runners:
         description: Use GitHub-hosted runners where supported, including OpenWebUI
         required: false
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [faca632e4e2a0d132e5ff0df6adee0b089fc324f](https://github.com/openclaw/openclaw/pull/141954/commits/faca632e4e2a0d132e5ff0df6adee0b089fc324f)
- Landed commit: [830c2eed4302280783e70a4158b2ff978c677bc1](https://github.com/openclaw/openclaw/commit/830c2eed4302280783e70a4158b2ff978c677bc1)

PR #141967, opened 2026-09-08
fix: preserve primary inheritance when editing agent fallbacks

shakkernerd:
Related: #137410

## What Problem This Solves

Fixes an issue where users editing an agent’s fallback models in Settings → Agents → Overview would also pin its inherited primary. Removing the last fallback could restore shared fallbacks when no primary was explicitly configured.

## Why This Change Was Made

Fallback edits now preserve the agent’s existing primary choice and save an explicit array, including an empty chain. The change uses the existing fallback-only model shape and leaves shared defaults and other agents intact.

## User Impact

Agents keep following changes to their shared primary after fallback edits. Clearing every fallback disables the chain, including for agents that use an implicit primary.

## Evidence

- All 11 focused writer tests pass. The original owner fails the inherited-primary and explicit-empty regression cases.
- Built Control UI at `609ca7ae74533c537145af01af37e10648d60e72` with the maintained stateful mock Gateway: adding B saves only `[A, B]` and follows shared Q after a page reload; clearing the final fallback saves `[]` and stays empty after config reload. Each edit sends one `config.set`, with defaults and unrelated agent metadata preserved.
- Remote formatting, lint, UI types, assertion/line ratchets, and MDX pass. [CI and required aggregate](https://github.com/openclaw/openclaw/actions/runs/34198332549) pass on the same head.

## Visual evidence

| View | Before | After |
| --- | --- | --- |
| Inherited primary survives fallback edits | Original behavior after the same edit and reload | Corrected writer behavior after the same edit and reload |
| Explicit empty fallback list survives reload | Original behavior after the same edit and reload | Corrected writer behavior after the same edit and reload |

### Exact-head fallback inheritance and explicit-empty interactions

https://github.com/user-attachments/assets/d4580016-2040-44b9-a78e-f212f3d51798

### Visual verification

- Baseline: f310ecc64128e33b3ace61e5a66f20ad8414be1f
- Exact head: `609ca7ae74533c537145af01af37e10648d60e72`
- Scenario: Add B to an inherited P+A chain, change the shared primary to Q and reload the page; then clear the last fallback with no authored primary and reload config.
- Runtime/build: Node v24.19.0; Chromium 144.0.7559.0; UI index SHA256 686e8b37e7a50565ad0d12ff70730ab9cc12f670bad52e3c87a15bd91caeb82c
- Writer primary ownership: Before: edits pin P. After: saved model has only fallbacks, then full page reload visibly inherits Q.
- Clearing final fallback: Before: inherited A returns. After: saved fallbacks [] remains empty after reload.
- Config writes: One config.set for each edit; unchanged defaults and unrelated agent metadata verified.
- Clean exact-head source and unchanged reused UI bundle hashes verified
- Equivalent baseline and fixed fixtures at 1440x1000, dark theme, en-US
- Actual stateful mock config.set payload plus config/raw readback asserted
- Source recordings retained; input cues do not cover controls
- In-place WaSelect label caching reproduced without edits on both base and fixed; full page reload verified with Q retained in mock config/raw
- Lead inspected all four original-resolution screenshots and the complete 27.16-second decoded recording sequence; synthetic fixtures only, with readable controls and input cues.

### DIFF
--- a/docs/concepts/model-failover.md
+++ b/docs/concepts/model-failover.md
@@ -61,6 +61,12 @@ The selection source controls whether the fallback chain is allowed:
 - **Legacy session override**: older session entries may have `modelOverride` without `modelOverrideSource`. OpenClaw treats those as user overrides so an explicit old selection is not silently converted into fallback behavior.
 - **Cron payload model**: a cron job `payload.model` / `--model` is a job primary, not a user session override. It uses configured fallbacks unless the job provides `payload.fallbacks`; `payload.fallbacks: []` makes the cron run strict.
 
+An agent can override only its fallback chain with `model: { fallbacks: [...] }`
+and keep inheriting the shared primary. Setting `fallbacks: []` explicitly disables
+fallbacks without pinning that primary. In **Settings → Agents → Overview**, editing
+fallback chips preserves primary inheritance; removing every chip saves an empty
+chain instead of restoring the shared fallbacks.
+
 Outside group and channel conversations, OpenClaw sends a visible notice when a turn moves onto fallback and another notice when a later turn succeeds on the selected primary. Group and channel conversations keep the same fallback state and lifecycle events without posting these notices. Persisted notice state prevents repeated notices when consecutive turns use the same selected/active pair, while model selection itself remains unchanged.
 
 ## Auth failure skip cache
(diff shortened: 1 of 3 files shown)


shakkernerd:
Fixed: editing agent fallbacks preserves primary inheritance and keeps explicitly cleared chains empty.

PR #141978, opened 2026-09-08
test(qa): bind Claude compaction fixture to active turn

RomneyDa:
## Summary

- make the Claude compaction fixture acknowledge live-session initialization
- wait for the active user record before emitting source-shaped turn output
- preserve the original commentary, compaction, tool, final, and terminal-result sequence

## Root cause

The fixture emitted its entire stream at process startup. The current Claude live-session runtime establishes an initialize/user boundary first, so those records preceded the active turn and were discarded; the process then ended with `exited unexpectedly without a terminal result`.

## Maturity failure addressed

- `telegram-claude-cli-compaction-final-priority`

## Testing

- executed the YAML fixture as a child, sent initialize and user protocol records, and observed control response, compaction completion, and terminal success result
- `pnpm format:check -- qa/scenarios/channels/telegram-claude-cli-compaction-final-priority.yaml`
- `git diff --check`
- autoreview P0-P2: scoped clean
- authenticated Telegram delivery remains the scorecard rerun proof

## Worked on by

- @RomneyDa

### DIFF
--- a/qa/scenarios/channels/telegram-claude-cli-compaction-final-priority.yaml
+++ b/qa/scenarios/channels/telegram-claude-cli-compaction-final-priority.yaml
@@ -41,6 +41,7 @@ scenario:
       finalMarker: CLAUDE-COMPACTION-TELEGRAM-FINAL-OK
       fixtureSource: |-
         #!/usr/bin/env node
+        const { createInterface } = require("node:readline");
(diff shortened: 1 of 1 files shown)


RomneyDa:
@clawsweeper re-review

PR #141979, opened 2026-09-08
fix(skills): watch status workspaces for hot installs

RomneyDa:
## What Problem This Solves

Fixes an issue where users opening skill status before the first agent turn could keep seeing stale skill availability after installing or editing a workspace skill. It also closes the initialization window where a status read could cache discovery before native file watches were registered.

## Why This Change Was Made

`skills.status` establishes the existing watcher after workspace resolution and session authorization. The watcher invalidates cached discovery after its initial scan completes, using the existing debounce, snapshot version, and shared-subscriber lifecycle. This reconciles changes that Chokidar suppresses with `ignoreInitial` during registration.

The repair stays in the watcher owner. Restarting the Gateway or increasing the status-test timeout would not fix a missed invalidation. Agent ownership, watch configuration, shared watch handles, capacity handling, and shutdown behavior stay on their existing paths. No config, protocol, dependency, or persistent-store changes.

## User Impact

New and edited file-backed skills become visible on subsequent status reads without restarting the Gateway, including changes made while watching starts. Status requests remain nonblocking. This addresses `skill-install-hot-availability`.

## Evidence

Validated in a fresh worktree with its own dependency install on macOS and Node 24.16.0.

- Before the watcher-owner repair, the full Gateway file failed its new hot-status assertion once while focused and later full runs passed. A real-watcher edit-during-registration test retained the old description, and a deterministic shared-readiness test emitted no invalidations; the ordinary change control passed.
- After the repair, 86 selected tests passed, with one Windows-only case skipped. The full Gateway file passed three consecutive runs (18 tests per run), preserving its original execution order.
- Coverage includes real filesystem creation/edits during registration, initially missing roots, shared subscribers, disabled watching, native watch capacity, shutdown/lifetime, and session snapshots.
- Independent autoreview through P2: scoped clean. The complete changed-file checks passed, including core/test typechecks, lint, formatting, architecture/dead-export scans, and state/auth guards. `git diff --check` passed.

Commands:

```sh
node scripts/run-vitest.mjs src/skills/runtime/refresh.test.ts src/skills/runtime/refresh.capacity.test.ts src/skills/runtime/refresh.missing-root.integration.test.ts src/skills/runtime/refresh.windows.test.ts src/skills/runtime/session-snapshot.test.ts src/gateway/server.models-voicewake-misc.test.ts
node scripts/run-vitest.mjs src/gateway/server.models-voicewake-misc.test.ts
node scripts/check-changed.mjs -- src/gateway/server-methods/skills.ts src/gateway/server.models-voicewake-misc.test.ts src/skills/runtime/refresh.ts src/skills/runtime/refresh.test.ts src/skills/runtime/refresh.missing-root.integration.test.ts src/skills/runtime/refresh.watcher.test-support.ts
```

Production LOC: +9/-0 total, including the original six-line status wiring and three-line readiness reconciliation. Tests and test support: +95/-23, including formatting of an existing shared-subscriber case extended to cover readiness.

Direct Chokidar 5.0.0 source inspection confirmed that counted initial directory scans and native watch registration precede `ready`; the existing-prefix watch strategy avoids the normal missing-path fallback. This repair reconciles initialization, rather than claiming atomic filesystem observation or immutable library-revision refresh.

## Worked on by

- @RomneyDa

### DIFF
--- a/src/gateway/server-methods/skills.ts
+++ b/src/gateway/server-methods/skills.ts
@@ -52,6 +52,7 @@ import {
 import { installSkill } from "../../skills/lifecycle/install.js";
 import { installUploadedSkillArchive } from "../../skills/lifecycle/upload-install.js";
 import { loadWorkspaceSkills } from "../../skills/loading/workspace-skill-loader.js";
+import { ensureSkillsWatcher } from "../../skills/runtime/refresh.js";
 import { getRemoteSkillEligibility } from "../../skills/runtime/remote.js";
 import {
   collectClawHubVerdictTargets,
@@ -273,6 +274,11 @@ export const skillsHandlers: GatewayRequestHandlers = {
         return;
       }
     }
+    ensureSkillsWatcher({
+      workspaceDir: resolved.workspaceDir,
+      config: resolved.cfg,
+      agentId: resolved.agentId,
+    });
     const report = buildRemoteAwareWorkspaceSkillStatus(
       resolved,
       target?.entry.skillLibrarySelections,
--- a/src/gateway/server.models-voicewake-misc.test.ts
+++ b/src/gateway/server.models-voicewake-misc.test.ts
@@ -713,6 +713,26 @@ describe("gateway server models + voicewake", () => {
           agentId: "ops",
           workspaceDir: path.join(workspaceRoot, "ops-workspace"),
         });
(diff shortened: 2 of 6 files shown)


RomneyDa:
@clawsweeper re-review

PR #141980, opened 2026-09-08
test(qa): select compact Slack progress proof

RomneyDa:
## Summary

- run the Slack omitted-commentary maturity scenario on the compact progress surface
- preserve `toolProgress: true` while making the editable draft visible in `chat.update` history
- cover the scenario-specific progress configuration in the runtime contract test

## Why

The maturity run observed commentary but could not find a history-visible tool-progress row. The default progress surface is now card-based, while this scenario explicitly proves the portable editable-text draft contract. Selecting `style: compact` keeps the scenario aligned with the surface it inspects without changing defaults for other Slack QA scenarios.

## Validation

- `pnpm vitest run extensions/qa-lab/src/live-transports/slack/slack-live.runtime.test.ts` (62 passed, before rebase)
- `pnpm prettier --check ...` (passed)
- `pnpm check:diff` (passed)
- local autoreview P0-P2: clean

Post-rebase local Vitest initialization is currently blocked by the shared checkout's dependency drift (`@vitest/browser` startup); hosted exact-head CI is authoritative.

Maturity failure: `slack-progress-commentary-omitted`

### DIFF
--- a/extensions/qa-lab/src/live-transports/slack/slack-live.config.ts
+++ b/extensions/qa-lab/src/live-transports/slack/slack-live.config.ts
@@ -302,6 +302,7 @@ export function buildSlackQaConfig(
                         commandText: "raw" as const,
                         label: false,
                         maxLines: 4,
+                        ...(progressOverrides.style ? { style: progressOverrides.style } : {}),
                         toolProgress: progressOverrides.toolProgress,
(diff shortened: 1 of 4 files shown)


RomneyDa:
@clawsweeper re-review

PR #141985, opened 2026-09-08
test(qa): scope Teams timeout fault to messages

RomneyDa:
## Summary

- limit the Teams accepted-504 QA fault to actual `message` activities
- let progress/typing activities that happen to echo the marker pass normally
- cover both activity types in the loopback Connector regression

## Why

The maturity scenario's marker is present in the message-tool arguments. Teams progress mode can surface those arguments through a native `typing` activity before the actual message tool dispatch. The loopback fault injector matched any activity text, so it injected 504 into the Microsoft SDK stream, whose independent retry loop produced repeated outbound events and prevented final settlement. The scenario is intended to prove OpenClaw does not replay an ambiguous non-idempotent message POST, not to fault an informative typing update.

## Validation

- `node scripts/run-vitest.mjs extensions/msteams/src/qa/bot-framework-server.test.ts` — 4 passed
- `oxfmt --check` — passed
- local autoreview P0-P2 — clean (0.98)

Maturity failure: `msteams-ambiguous-gateway-timeout`

### DIFF
--- a/extensions/msteams/src/qa/bot-framework-server.test.ts
+++ b/extensions/msteams/src/qa/bot-framework-server.test.ts
@@ -53,7 +53,7 @@ describe("Microsoft Teams QA Bot Framework server", () => {
     }
   });
 
-  it("models an ambiguous gateway timeout after accepting the marked activity", async () => {
+  it("models an ambiguous gateway timeout only for a marked message activity", async () => {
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper re-review

PR #141679, opened 2026-09-08
fix: prevent stale provider hooks and incorrect model alias selection

steipete:
Related: #130706.

## What Problem This Solves

Fixes an issue where users changing model defaults, aliases, or fallbacks from a fresh CLI could save a provider's legacy model ID instead of its canonical ID. Reusing a warm registry could also select hooks from another plugin copy or workspace, keep a disabled provider in a mixed selection, or let another provider's runtime alias take over after a declared provider failed to load.

## Why This Change Was Made

Provider parsing and hooks now share one registry selection with separate required load owners and eligible receivers. Declared providers reserve their names even when registration fails; activation helpers can still run beside them, and a setup-only registration does not satisfy a runtime registration requirement. Retained generations keep their captured selection, including an explicitly empty selection.

The loader owns the physical source, artifact-selection, workspace, and registration-input facts used for reuse. Ordinary lookups validate the selected origin, root, entry, source/build choice, and paired runtime/source configuration against those records. Current enablement independently narrows eligible receivers before warm reuse, so disabling one provider cannot leave its hooks in an otherwise valid mixed selection. The shared selection replaces duplicate provider-hook caches and normalization paths.

Model mutations prepare the selected providers and required helpers through the canonical loader, then resolve aliases under that registry and metadata scope. Runtime-only aliases can identify an eligible owner to load, while exact command preparation still follows the normal loader policy. The original config snapshot/hash fences the write against concurrent replacement. The legacy public SDK `loadModelCatalog` metadata input remains source-compatible when callers omit the new internal owner index; internally built snapshots still provide complete owner facts.

## User Impact

Default-model, image-model, alias, and fallback commands agree on the canonical provider model. Existing model-entry settings move to the canonical key, with existing canonical settings taking precedence; adding an alias replaces the model's previous alias. Unrelated authored values, including environment references, remain in source form. A failed declared provider cannot silently hand normalization to another provider's alias, and ordinary lookups no longer reuse a disabled or physically unrelated provider's callbacks.

## Evidence

- Fresh rebased head: `d0b317508a6d76e754c931269280ba15a1f4af6b`, based on main `53ff0867149e1fd753cbfc9f67f79a28e6318f58`. The stale-base lint failure is absent: all 95 changed paths, 24 large files changed by main, and explicit update-command/markdown lint controls pass `pnpm exec oxlint`.
[...]

PR #140187, opened 2026-09-06
perf(plugins): avoid full inspection for compatibility notices

steipete:
## What Problem This Solves

Status and Doctor construct full plugin inspection reports just to collect compatibility notices. That also prepares unused runtime policy context and inspects MCP/LSP support, adding work to diagnostic paths with larger plugin registries.

## Why This Change Was Made

Project notices directly from the supplied registry, or build the diagnostics registry from the existing options when none is supplied. The projection reuses the established shape classifier, exact plugin diagnostic attribution, and notice builder. Full inspection and snapshot/live-registration merging retain their existing owners.

## User Impact

Compatibility warnings require less inspection work and preserve the same notice content and order. A supplied report remains authoritative even when the ambient registry differs. This is a maintainer-requested performance cleanup with no new cache or configuration surface.

## Evidence

- Both existing status suites pass **38 tests** against baseline, candidate, and final source. The expanded existing fixture covers supplied/generated reports, both notice kinds, interleaved/global diagnostics, bundled errors, and unrelated plugins.
- Changed-check types, guards, boundaries, and dead-export stages passed. The final lint stage initially caught an oversized test file; shortening synthetic fixture IDs and compacting expected calls retained the cases, after which the exact lint command, 38 tests, and formatting passed. `git diff --check` passes.
- Independent source review and final exact-diff managed autoreview found no actionable issues.
- The actual owner produced **49 notices with identical complete JSON** in a 100-plugin fixture. Per supplied-report call, native MCP inspection fell 50 → 0, bundle MCP 50 → 0, bundle LSP 50 → 0, and ambient runtime-context preparation 1 → 0. Count probes called through to the real inspection implementations.
- Across five samples of 40 warm calls, median component time fell **1.399 → 0.168 ms**. Existing dependency mocks and concurrent checks limit that timing claim; eliminated calls and exact output parity are the primary evidence. This is not a whole-Gateway latency, allocation-byte, heap, or RSS measurement.
- Required hosted CI passed for exact head `04f5f38db4364dd908b5e9241b07b121e2221ac6`: https://github.com/openclaw/openclaw/actions/runs/34040631123.

PR #141989, opened 2026-09-08
test(plugins): activate the hook-only compatibility fixture through the runtime context

steipete:
## What Problem This Solves

`src/plugins/status.compatibility.integration.test.ts` ("reports actual hook-only registrations without activating cold plugin modules") fails deterministically on current `main` when run on its own:

```
expected [] to deeply equal [ ObjectContaining{ pluginId: "runtime-hook-only", code: "hook-only" } ]
```

`git bisect` over `aa947ec2812~1..main` lands on #141679 (`ccd3a4414e5`). That change made a supplied manifest registry's physical source facts part of the plugin load identity (`manifestSources` in `buildCacheKey`), which is the intended binding for provider reuse. The test activates its registry with the plain shape `loadOpenClawPlugins({ config, workspaceDir, env, cache: false })`; with no process-current metadata snapshot published in the test process, that identity carries no physical sources, while `buildPluginCompatibilitySnapshotNotices` resolves the runtime load context and hashes an explicit registry. `resolveCompatibleRuntimePluginRegistry` requires exact identity equality, so the live registry is not reused and hook shape is never observed.

Production is consistent with itself: Gateway startup (`src/gateway/server-plugins.ts`) and the runtime registry loaders activate through `buildPluginRuntimeLoadOptions(resolvePluginRuntimeLoadContext(...))`, the same path the snapshot notices use; the plain activating shape exists only in this test.

## Why This Change Was Made

The test now activates through the resolved runtime context exactly as Gateway and CLI loads do, with a comment naming the identity contract. No production change: the identity binding from #141679 is the correct owner behavior, and the notice path's exact-match reuse is its documented contract ("return the exact active registry without triggering a fresh load on cache miss").

## User Impact

None at runtime. Restores a deterministic local red on `main` for anyone running the plugins suite on a checkout at or after #141679.

## Evidence

- Pre-fix, isolated on `main` (`a6046641e91`) and on `d662d575372`: 2/2 and 3/3 runs fail with the assertion above; identical under default, canonical, and private-per-user `TMPDIR`, so not the macOS `/var` symlink.
- Probe of the hasher inputs for both option shapes: everything equal except `manifestSources` (`undefined` for the plain load, the resolved `[id, origin, rootDir, source, ...]` tuple for the snapshot path); the active registry key equals the plain-load key and differs from the snapshot key.
- `git bisect run` with this single test: good at `aa947ec2812~1` and `d12106ef815`, first bad `ccd3a4414e5` (#141679).
- Post-fix: the file passes 3/3 in isolation; the cold call still returns `[]` without creating the runtime marker; the `src/plugins/status` suites pass together; `check-changed` on the touched file passes.
- CI did not catch the break because none of the inspected node shard logs for #140187's and #141679's exact-head runs mention this file; the changed-target shard for this PR runs it directly.

Test-only change, production LOC 0.

### DIFF
--- a/src/plugins/status.compatibility.integration.test.ts
+++ b/src/plugins/status.compatibility.integration.test.ts
@@ -11,6 +11,8 @@ import {
   useNoBundledPlugins,
   writePlugin,
 } from "./loader.test-fixtures.js";
+import { buildPluginRuntimeLoadOptions } from "./runtime/load-context.js";
+import { resolvePluginRuntimeLoadContext } from "./runtime/load-context.resolve.js";
 import { buildPluginCompatibilitySnapshotNotices } from "./status.js";
 
 function addStartupActivation(pluginDir: string, onStartup: boolean): void {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f44ca3ec249acbb5b383078fb5ef27ba6aaf19fc](https://github.com/openclaw/openclaw/pull/141989/commits/f44ca3ec249acbb5b383078fb5ef27ba6aaf19fc)
- Landed commit: [8ac88a1a335aa5f6ec467e539f6bbce3df478d84](https://github.com/openclaw/openclaw/commit/8ac88a1a335aa5f6ec467e539f6bbce3df478d84)

Issue #92708, opened 2026-06-13
[P2] Feishu streaming card silently truncates long replies (rate-limit 230020/200850 mishandled, sequence gap poisons stream)


quick3210-gif:
# OpenClaw Bug Report — Feishu Streaming Card Truncation

**Date**: 2026-06-13
**Reporter**: Roy
**OpenClaw Version**: 2026.6.5
**Severity**: P2 (visible but confusing — message appears truncated/garbled in Feishu client; user thinks the agent is broken or hung)
**Affected Component**: `extensions/feishu/src/streaming-card.ts` (compiled to `dist/monitor.account-ky5Qt4fa.js`)

---

## Summary

When the OpenClaw Feishu plugin streams a long assistant reply as a live card (the default `streaming: true` behavior), **the message body in the Feishu client frequently ends up truncated or shows only a partial string**, even though the model output the full reply and the gateway log shows the full text being sent. The user sees "stuck mid-sentence" or "missing the last paragraph" responses. The work-around — disabling streaming entirely and using a single-shot `message` tool `announcement` mode — works but loses the typewriter effect and requires the agent to remember to bypass chat-reply.

This bug is reproducible with any reply longer than ~400 characters.

---

## Reproduction Steps

1. Configure OpenClaw with a Feishu account using the default `streaming: true` and `blockStreaming: true`.
2. Send a question that triggers a long reply (e.g., a 600-1500 character answer with several paragraphs and lists).
3. Observe the Feishu client as the card streams in. The reply frequently:
   - Stops at ~400-700 characters and the rest never appears
   - Shows a different (earlier) snapshot of the text — older content overwriting newer content
   - Displays the model output's first half but loses the second half
4. Check the OpenClaw gateway log: the **full reply is in the model output and was sent to the streaming endpoint** (PUT `/cardkit/v1/cards/<cardId>/elements/content`). The truncation is a **rendering / coalescing problem**, not a delivery problem.

---

### Timeline (UTC+8, 2026-06-09)

- **15:43** — Roy's LUCKY debugging reply (table + 6 bullet list + emoji + `**` bold + `→` arrows) → Feishu client only shows "哪个先？"
- **15:43** — Follow-up apology with similar structure → also truncated to first line
- **17:04** — Roy's "URL 模式" answer reformatted as 4 plain-text short lines (no markdown) → Feishu client shows only "?"
- **17:09** — Roy: "你应该在飞书上都用 announcement 来发送信息" → workaround command
[...]

PR #141991, opened 2026-09-08
fix(feishu): keep replies delivered when Feishu closes a streaming card

### LINKED ISSUE #139443: [feishu] Streaming card dies during long tool phases: 200850 idle timeout followed by ~1000 blind 300309 retries
### Summary

With `channels.feishu.streaming.mode = "partial"`, a turn that spends several minutes inside a **tool phase** (no assistant text increments) gets its streaming card closed server-side with code **200850** (`card streaming timeout`). The plugin then blindly retries every subsequent card update against the dead stream, logging **~978 consecutive 300309** (`streaming mode is closed`) errors at ~150–250 ms intervals until the turn finishes and the stream is finally closed by the plugin itself.

### Environment

- OpenClaw gateway `2026.9.2` (3928bad), plugin `@openclaw/feishu` `2026.9.2`
- `channels.feishu`: `connectionMode: "websocket"`, `renderMode: "card"`, `streaming.mode: "partial"`, `chunkMode: "newline"`
- Agent profile: reasoning enabled, heavy multi-tool turns, turn duration regularly > 10 minutes (long thinking + many tool calls)
- Provider: OpenAI-compatible completions API (Z.AI GLM), if that matters for pacing

### Timeline (one reproduced incident, 2026-09-06, systemd journal)

Turn started ~05:28. Long tool phase with no text deltas:

```
05:38:42.256 [feishu] feishu[default] Update failed: Error: Update card content failed: ErrMsg: card streaming timeout;  (code=200850)
05:38:42.447 [feishu] feishu[default] Update failed: ... ErrMsg: streaming mode is closed;  (code=300309)
05:38:42.638 [feishu] feishu[default] Update failed: ... ErrMsg: streaming mode is closed;  (code=300309)
05:38:42.833 [feishu] feishu[default] Update failed: ... ErrMsg: streaming mode is closed;  (code=300309)
... (repeating every ~150-250 ms)
05:41:26.819 [feishu] feishu[default] Update failed: ... ErrMsg: streaming mode is closed;  (code=300309)
05:41:27.247 [feishu] feishu[default] Note update failed: ... ErrMsg: streaming mode is closed;  (code=300309)
05:41:27.571 [feishu] feishu[default] Closed streaming: cardId=7682166285232278496
```

978 × 300309 logged between 05:38:42 and 05:41:27. The visible card freezes mid-turn ("dead stream"): users see a stalled typewriter for minutes, then the finalized content lands at turn end.

### Root cause analysis

1. Feishu's Card Kit streaming API closes a stream after an idle window with no card patches. During long tool phases the assistant produces **no text increments**, and the plugin sends no patches, so the idle timer fires (200850).
2. `STREAMING_START_FAILURE_BACKOFF` (introduced around 2026.9.x for the 300309 storms on stream *start*) only guards **start failures**. A mid-stream 200850 is not covered.
3. After 200850, every subsequent update attempt (content patches and note updates) fails with 300309, but the plugin keeps retrying each scheduled update until the turn completes — no terminal-state handling, no cap, no backoff.

### Requested behavior

1. **Tool-phase keepalive**: while a turn is inside a tool/thinking phase with no text deltas, emit a lightweight periodic patch (e.g. every 30–60 s) to keep the server stream alive. The existing `onToolStart` / status-line events are natural milestones — even a no-op note/content patch or a "🔧 tool running…" milestone update would reset the idle timer.
2. **Treat 300309 as terminal for that stream**: after the first `streaming mode is closed`, stop patching that card via the streaming API, mark the stream dead, and fall back to the non-streaming finalize path (full card update) so content is not lost and logs are not flooded.
3. Optionally reuse the start-failure backoff machinery for mid-stream failures as defense in depth.

### Workaround

`streaming.mode = "off"` (whole-card delivery). This is what we run now — but it forfeits streaming UX entirely for deployments whose turns legitimately contain long tool phases.

### Related

- #92708 (closed; 200850 mishandling on rate-limit/sequence gaps — different trigger, same error code)
- #108265, #91941 (streaming latency/rendering quality)

Happy to provide full journal excerpts or test a prerelease with keepalive on this workload (turns > 10 min, 20+ tool calls).

ylcn91:
Fixes #139443

## What Problem This Solves

Fixes an issue where Feishu users running long tool phases would see the streaming card freeze on a stale partial answer while the bot issued roughly a thousand rejected API calls, and in one branch never received the completed answer at all.

Feishu can close a card's streaming mode server-side (the reporter observed code `200850` after an idle tool phase). Every later request against that card is then rejected with `300309`. Today nothing recognizes that closure: the card keeps its stale text, the failure log fills for the rest of the turn, and if any preview text was accepted before the close, the completed reply is silently skipped because the card already counts as "visible".

## Why This Change Was Made

`assertSuccessfulCardKitResponse` flattened the CardKit numeric code into a message string, so no caller could tell a terminal stream closure from a retryable failure. The session kept reporting `isActive()` — the reply dispatcher's only liveness signal — and because `sentText` advances only on an accepted write, the frozen snapshot made every later delivery exceed the significant-delta threshold and bypass the 160 ms throttle, turning each provider chunk into one immediately-issued PUT.

The repair is owned by the Feishu session:

- Throw a typed `FeishuCardKitError` that keeps the numeric code. The rendered message is byte-identical, so operator log greps and existing log assertions are unaffected.
- Classify `200850` and `300309` as terminal in one small module, mirroring the existing `send-rate-limit.ts` shape. `19001`, HTTP 429, and transport failures stay retryable.
- On a terminal rejection the session retires itself via a dedicated `streamClosedByServer` flag, deliberately kept separate from `closed` (which means "we finalized" and makes `closeWithResult`/`discard` early-return without their receipt). `isActive()` then goes false, which alone ends the retry storm — the dispatcher's existing guards already honor it.
- Finalization on a retired stream skips the final write, the note update, and the settings PATCH instead of issuing three more guaranteed rejections, while still throwing `FeishuStreamingFinalizationError` with the terminal cause.
- A frozen card is visible but is not the delivered answer, so the close result records `finalTextAccepted` at the boundary that made the write, and the dispatcher carries it to `ensureVisibleStreamingDelivery` so a stale-partial card falls back to the static send.

Non-goals: no keepalive, no new configuration options, no retries. A tool-phase keepalive was considered and deliberately left out — the server idle window is unmeasured here, and whether a note or unchanged-content patch resets it is unverified, so any such timer would be a speculative fallback.

## User Impact

A Feishu streaming card that the server closes now stops updating instead of spraying rejected requests, and the completed answer is delivered as a static card or message instead of being lost behind a frozen preview.

Two things to expect:

- The visible prefix appears twice when the card froze mid-answer (stale partial on the card, then the full answer as a separate message). That is the intended trade against losing the answer, and matches the existing recovery-card behavior.
- The underlying server-side idle close is unchanged, so a long tool phase can still freeze the preview. This change makes that bounded and recoverable — one rejected request, then a clean static delivery — rather than a multi-minute storm.

Documented under **Streaming** in `docs/channels/feishu.md`.

## Evidence

Regression tests were confirmed to fail without the production change and pass with it, in the same run.

**Without the fix** (production files reverted, tests kept):

```
FAIL streaming-card.test.ts > retires a stream CardKit closed with 200850 instead of blind-retrying every later patch
  AssertionError: the session must stop patching a stream CardKit reported closed:
  expected [ …(6) ] to have a length of 1 but got 6

FAIL streaming-card.test.ts > does not PATCH close settings on a stream CardKit already reported as closed
  AssertionError: closeWithResult must not PATCH streaming settings on a stream CardKit
  reported closed: expected [ Array(1) ] to deeply equal []

FAIL reply-dispatcher.test.ts > delivers the complete answer statically when a server-closed stream kept only a stale partial
  AssertionError: a card frozen on a stale partial is not a delivered answer; the dispatcher
  must fall back to the static path: expected [] to include 'Working on it.\nHere is the complete …'
```

The first failure is the reported storm: six snapshots produced six PUTs inside a single 160 ms window. The test also pins the journal line from the issue byte-for-byte:

```
Update failed: Error: Update card content failed: ErrMsg: card streaming timeout;  (code=200850)
```

**With the fix:** full `extensions/feishu` suite green — 102 files, 2038 tests.

Over-classification guards are included and were confirmed to pass both with and without the fix, so the retryable contracts cannot silently regress: a `19001` body error and an HTTP 429 both leave `isActive()` true.

**Proof gaps, stated plainly:**

- `200850` is **not confirmed from an authoritative Feishu source.** Attempts to fetch the CardKit error-code reference returned no usable content, and the repository contains no occurrence of either code. The reporter's observation and OpenClaw issue #92708 (closed, different trigger) are the only evidence that `200850` is real on this endpoint. `300309` has independent third-party corroboration from another Feishu client that stopped calling the stream API after close for the same reason. The repair deliberately does not depend on the precise upstream semantics: both codes are terminal for this card's stream by observation, and the handling is identical.
- The `200850` idle close itself is **not reproduced** — that needs a live tenant and a real idle window. The tests inject the code at the transport boundary and prove everything downstream.
- No live Feishu tenant proof.

**Validation still running at the time this draft was opened:** `check-changed` (tsgo typecheck lanes) and `$autoreview`. Both were started before opening and results will be posted as follow-up commits if anything needs fixing. An earlier `check-changed` run on the same tree passed every lane except `oxfmt`, which has since been applied.

### DIFF
--- a/docs/channels/feishu.md
+++ b/docs/channels/feishu.md
@@ -346,6 +346,8 @@ Feishu/Lark supports streaming replies via interactive cards (Card Kit streaming
 
 Set `streaming.mode: "off"` to send the completed reply without streaming updates; long replies still split at the message limits above. `renderMode: "raw"` (plain text instead of cards) also disables streaming cards. `streaming.block.enabled` is off by default; enable it only when you want completed assistant blocks flushed before the final reply. Legacy boolean `streaming` and the flat `blockStreaming` / `blockStreamingCoalesce` / `chunkMode` keys migrate to this nested shape via `openclaw doctor --fix`.
 
+Feishu can close a card's streaming mode on its own — for example after a long tool phase with no card updates. When that happens the card stops accepting updates, so it keeps whatever text it last showed and the completed reply arrives as a separate static card or message. Expect the visible prefix to appear twice in that case; the full answer is the later message.
+
 Replies with controls use native cards for command buttons and HTTP(S) links, including when streaming is off. The card carries the reply text; attachments remain separate messages. Unsupported controls and cards that exceed Feishu's size limits keep their full labels in a readable fallback. That fallback remains a separate message when a later reply streams. A final controls reply replaces an active streaming preview without sending the preview text again; error controls after a completed answer remain separate. If Feishu cannot delete or clear a replaced preview, delivery reports a failure and retains the original message receipt.
 
 ### Quota optimization
--- a/extensions/feishu/src/reply-dispatcher.test.ts
+++ b/extensions/feishu/src/reply-dispatcher.test.ts
@@ -2716,6 +2716,66 @@ describe("createFeishuReplyDispatcher streaming behavior", () => {
     });
   });
 
+  // Regression coverage for #139443.
+  it("stops queueing streaming patches once the session retires a CardKit-closed stream", async () => {
+    const { result, options } = createDispatcherHarness();
+    await options.onReplyStart?.();
+    result.replyOptions.onPartialReply?.({ text: "Working on it." });
+    await new Promise((resolve) => setTimeout(resolve, 0));
+
+    const instance = requireStreamingInstance(0);
+    const patchesBeforeRetirement = instance.update.mock.calls.length;
+    // CardKit retires the stream on the next patch; a repaired session stops reporting
+    // itself active, which is the dispatcher's only liveness signal.
+    instance.update.mockImplementation(async () => {
+      instance.active = false;
+    });
+
+    for (let index = 0; index < 6; index += 1) {
+      result.replyOptions.onPartialReply?.({
+        text: `Working on it.\nstreamed answer paragraph ${String(index).padStart(2, "0")}`,
+      });
+    }
+    await new Promise((resolve) => setTimeout(resolve, 0));
+
+    expect(
+      instance.update.mock.calls.length - patchesBeforeRetirement,
+      "the dispatcher must not keep patching a stream whose owner reports it retired",
+    ).toBeLessThanOrEqual(1);
+  });
+
+  it("delivers the complete answer statically when a server-closed stream kept only a stale partial", async () => {
+    const { result, options } = createDispatcherHarness();
+    result.replyOptions.onPartialReply?.({ text: "Working on it." });
+    const finalText = "Working on it.\nHere is the complete answer the user must still receive.";
+    const delivery = await options.deliver({ text: finalText }, { kind: "final" });
+    // 200850 froze the card after "Working on it." was accepted, so the final write and
+    // the close were both rejected: the card is visible but the answer never landed.
+    requireStreamingInstance(0).closeWithResult.mockRejectedValueOnce(
+      new FeishuStreamingFinalizationError(
+        new Error("Update card content failed: ErrMsg: streaming mode is closed;  (code=300309)"),
+        {
+          visibleReplySent: true,
+          finalTextAccepted: false,
+          content: "Working on it.",
+          messageId: "om_stream",
+        },
+      ),
+    );
+
+    await Promise.resolve(options.onIdle?.()).catch(() => undefined);
+    await Promise.resolve(delivery?.finalization).catch(() => undefined);
+
+    const staticSends = [
+      ...sendStructuredCardFeishuMock.mock.calls,
+      ...sendMessageFeishuMock.mock.calls,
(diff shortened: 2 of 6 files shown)


ylcn91:
Closing this PR. I cannot produce the real Feishu behavior proof the review asks for (no live Feishu tenant available to me), so I am not able to carry it to a mergeable state. Leaving #139443 open for someone who can verify against a real Feishu card.

PR #141953, opened 2026-09-08
fix: discard abandoned fallback model searches

shakkernerd:
Related: #141330

## What Problem This Solves

Fixes an issue where users searching the fallback model field in Settings → Agents would save an unfinished query as a model reference when they pressed Tab or clicked away.

## Why This Change Was Made

Leaving the field now discards unconfirmed text. Explicit catalog selection, the custom Add row, Enter, and comma retain their commit behavior, and the field documentation describes cancellation.

## User Impact

Searching and dismissing the list leaves the agent’s fallback chain unchanged. Users can still add unlisted model references and paste ordered lists.

## Evidence

- All 14 focused selector tests pass; the original component fails the new Tab/blur cancellation regressions by saving the search text.
- Built Control UI at `0f40cc365d9b5d05c5c9352c55878a754bff9c52` with the maintained stateful mock Gateway: Tab and clicking away discard the query with zero config writes after the autosave interval. Explicit Add, comma, and Enter each preserve an intentional edit and match config/raw readback.
- Remote formatting, lint, UI production/test types, assertion/line ratchets, and MDX pass. [CI and required aggregate](https://github.com/openclaw/openclaw/actions/runs/34197614448) pass on the same head.

## Visual evidence

| Before | After |
| --- | --- |
| Original fallback behavior with the same synthetic model setup | Corrected fallback behavior with the same synthetic model setup |

### Exact-head fallback interaction recording

https://github.com/user-attachments/assets/4b7b9ee1-1983-479c-85b1-5d32994d9641

### Visual verification

- Baseline: 6627766f5bfa403d5425e7c1da40f919fa924b45
- Exact head: `0f40cc365d9b5d05c5c9352c55878a754bff9c52`
- Scenario: Type gem, leave with Tab and click-away without choosing a model, then explicitly choose Add; explicit comma and Enter still save.
- Runtime/build: Node v24.19.0; Chromium 144.0.7559.0; UI index SHA256 d956279579ca5405db32c177a55cbecfb3dd571005032850f48a92ab14ba4a18
- Unconfirmed search: Tab and click-away: no chip and zero config.set calls after the normal autosave interval.
- Explicit confirmation: Add, comma, and Enter each preserve an intentional fallback edit.
- Clean exact-head source and emitted UI index/assets verified
- Retained original audit baseline reused after all eight component/caller/config owner hashes matched the selected base
- Same viewport, dark theme, en-US and equivalent Model Selection framing
[...]

PR #141997, opened 2026-09-08
fix: preserve case-sensitive fallback model references

shakkernerd:
Related: #141330, #141953, #141967

## What Problem This Solves

Fixes an issue where users entering a case-sensitive fallback model reference could have it rewritten or hidden when the catalog contained a differently cased ID. The field could also offer the primary model again when it was selected through a configured alias.

## Why This Change Was Made

Search matching is separated from model identity. Catalog values and labels preserve model suffix case, typed references stay as authored, and primary exclusion uses the current form’s configured aliases with the distinct primary and fallback resolution rules.

## User Impact

References such as `custom/Model-A` and `custom/model-a` remain separate choices. Users can keep custom references and aliases without an automatic rewrite, while configured primary aliases exclude the matching fallback target.

## Evidence

- All 110 focused selector, agent-display, catalog-label, and overview tests pass. Original case-preservation regressions fail because the old component omits the case-distinct Add row.
- Built Control UI at `3fc9dd3f1e1a3cb04958f19cf7f3a11adcd9169f` with the maintained stateful mock Gateway: all 12 browser cases pass config/raw readback. These cover exact custom case through Add/Enter/comma, distinct catalog labels, primary aliases and per-agent overrides, explicit clipboard confirmation, retained alias bindings, and bare/profile-qualified alternatives.
- Remote formatting, lint, UI production/test types, assertion-safety, and line-ratchet checks pass. [CI and required aggregate](https://github.com/openclaw/openclaw/actions/runs/34200906453) pass on the same head.

## Visual evidence

| View | Before | After |
| --- | --- | --- |
| Custom model suffix case is preserved | Original fallback behavior with the same synthetic model setup | Corrected fallback behavior with the same synthetic model setup |
| Case-distinct fallback remains available beside the primary | Original fallback behavior with the same synthetic model setup | Corrected fallback behavior with the same synthetic model setup |

### Exact-head fallback interaction recording

https://github.com/user-attachments/assets/ae3934c1-bfd8-4cc9-acdc-8eb4e7151f31

### Visual verification

- Baseline: 6627766f5bfa403d5425e7c1da40f919fa924b45
- Exact head: `3fc9dd3f1e1a3cb04958f19cf7f3a11adcd9169f`
- Scenario: Explicitly add a case-distinct custom reference and add that same reference beside its lowercase primary; validate alias, catalog label, and clipboard alternatives.
- Runtime/build: Node v24.19.0; Chromium 144.0.7559.0; UI index SHA256 fda4aeb1dddf6ea9a0db4e9c61734bd9ffdd4fd86162d516f2b52dd3ba748960
- Authored suffix case: custom/Model-A is saved exactly through Add, Enter and comma.
- Distinct identities: Lowercase and uppercase catalog entries and labels remain distinct; case-distinct fallback is allowed beside primary model-a.
- Aliases and alternatives: Primary alias/profile exclusion, per-agent override, authored alias/CSV, bare and profile-qualified alternatives pass config.set readback.
- Clean exact-head source and emitted UI index/assets verified
- Retained original audit baseline reused after all eight component/caller/config owner hashes matched the selected base
- Same viewport, dark theme, en-US and equivalent Model Selection framing
- Assertions cover actual stateful mock config.set payload plus config/raw readback
- Original screenshots and source recordings retained
- All comparison screenshots and the complete recording sequence were inspected; inputs, cues and settled results are readable and contain only synthetic data.

### DIFF
--- a/ui/src/components/multi-select.test.ts
+++ b/ui/src/components/multi-select.test.ts
@@ -1,5 +1,7 @@
 /* @vitest-environment jsdom */
 import { afterAll, afterEach, beforeAll, expect, it, vi } from "vitest";
+import { normalizeAgentModelRefForConfig } from "../../../src/config/model-input.js";
+import { createPrimaryModelExclusion } from "../lib/agents/display.ts";
 import { MultiSelect, type MultiSelectOption } from "./multi-select.ts";
 
 const MULTI_SELECT_TEST_TAG = `test-openclaw-multi-select-${crypto.randomUUID()}`;
@@ -7,7 +9,8 @@ const MULTI_SELECT_TEST_TAG = `test-openclaw-multi-select-${crypto.randomUUID()}
 type MultiSelectElement = HTMLElement & {
   options: readonly MultiSelectOption[];
   value: readonly string[];
-  exclude: readonly string[];
+  isExcluded: (value: string) => boolean;
+  getValueKey: (value: string) => string;
   placeholder: string;
   allowCustom: boolean;
   disabled: boolean;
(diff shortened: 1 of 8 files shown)


shakkernerd:
Fixed: fallback model selection preserves exact references and correctly excludes configured primary aliases.

PR #141999, opened 2026-09-08
refactor(tui): remove unreachable stream boundary state

### LINKED ISSUE #141907: Remove unreachable block-boundary logic from TUI streaming
The TUI retains a separate heuristic and per-run state for streamed snapshots containing multiple text blocks around a tool block. The supported Gateway and embedded backends already combine streamed text into one cumulative text block, so this extra preservation algorithm cannot run on their output. Its original isolated fixtures injected provider-shaped arrays directly into the assembler and bypassed those producers.

This maintainer-requested reduction removes the complete unreachable heuristic, its state and option plumbing, and the three tests that exist only to preserve that unreachable shape. Ordinary text/thinking display, empty-final fallback, abort/error text, attachments, richer final arrays and live-run eviction remain covered and unchanged. The web UI's separate partial-error-message handling remains outside this change.

Compatibility was checked against the current protocol-v4 TUI handshake and both real backends, plus 150 locally tagged v4 versions with 27 distinct producer-constructor excerpts. All inspected delta factories emit one text block. The local version-tag inventory matches all 326 remote version-tag refs; this is source-history evidence, not a claim about release publication. The original February fixture lineage also already had a one-block Gateway producer.

Retain the existing real Gateway pre-tool preservation, embedded final-selection, and TUI abort/attachment coverage. Validation includes before/after owner suites, supported-stream differential comparisons, a producer-to-assembler check, bounded real runTui PTY proof, selected checks and independent review. Fake-backend PTY evidence must not be described as live Gateway or provider proof.

The proposed complete reduction is 104 production and 24 test code lines, plus the required one-line assertion-baseline shrink. No supported user-visible behavior, protocol, configuration, dependency or public API changes are intended.

steipete:
Closes #141907

## What Problem This Solves

The TUI retained extra streaming state and a preservation heuristic for input shapes its supported backends never emit. This maintainer-requested cleanup removes that obsolete complexity while retaining the visible streaming behavior.

## Why This Change Was Made

Gateway and embedded backends send one cumulative text block per streamed snapshot. The removed algorithm required at least two prior text blocks; its original isolated fixtures bypassed the real producer. Delete the complete heuristic, its state and option plumbing, and the three tests that only protected that unreachable shape.

Mixed-version compatibility was checked across 150 protocol-v4 version tags and both current backend owners. The complete change removes 104 production and 24 test code lines, plus one required assertion-baseline entry.

## User Impact

No supported user-visible behavior change. Ordinary text/thinking display, rich final arrays, empty-final fallback, abort/error text, attachments and protected live-run retention remain intact.

## Evidence

- Four owner files: 564 baseline tests and 561 retained tests pass; exactly the three identified heuristic cases disappear.
- 6,800 supported-stream scenarios produce 40,806 equal observations with unchanged input bytes. Actual Gateway producer-to-assembler final and abort replay also matches before/after.
- Three targeted real runTui PTY checks pass for streamed prefix/final text, tool-card ordering and attachments. These use a fake backend and do not claim live Gateway/provider or persistence proof.
- All selected checks pass: 29 gates plus one warning-only temp-file report, followed by fresh full-scope P2 review with no actionable findings.

No additional full build was needed for this deletion: it introduces no import, lazy-loading, package or dependency boundary. The existing focused tests compiled their normal subprocess fixtures.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3654,7 +3654,6 @@ src/tui/tui-session-actions.ts	5
 src/tui/tui-session-events.ts	1
 src/tui/tui-session-projection.ts	2
 src/tui/tui-session-run-coordinator.ts	1
-src/tui/tui-stream-assembler.ts	2
 src/tui/tui-task-suggestions.ts	6
 src/tui/tui.ts	5
 src/utils.ts	1
--- a/src/tui/tui-stream-assembler.test.ts
+++ b/src/tui/tui-stream-assembler.test.ts
@@ -14,8 +14,6 @@ const messageWithContent = (content: readonly Record<string, unknown>[]) =>
     content,
   }) as const;
 
-const TEXT_ONLY_TWO_BLOCKS = messageWithContent([text("Draft line 1"), text("Draft line 2")]);
-
 type FinalizeBoundaryCase = {
   name: string;
   streamedContent: readonly Record<string, unknown>[];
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0471a675bc3b89dd8862829bc5dc6a6ee23f2cb0](https://github.com/openclaw/openclaw/pull/141999/commits/0471a675bc3b89dd8862829bc5dc6a6ee23f2cb0)
- Landed commit: [712823ff8d77ed23fe4cee3efd9fec0090ee046e](https://github.com/openclaw/openclaw/commit/712823ff8d77ed23fe4cee3efd9fec0090ee046e)

PR #142007, opened 2026-09-08
fix(release): preserve frozen upgrade scenario closure

RomneyDa:
## Summary

- bind the authorized selected release’s complete upgrade-scenario import closure into Docker
- prevent a frozen scenario from importing incompatible current-main helper modules
- keep current and unauthorized targets on the trusted current harness

## Root cause

Full validation mounted the selected `scripts/e2e/lib/upgrade-survivor` directory, but its relative imports escaped that mount into current main. The 2026.6.35 scenario imported `parseReleaseVersion` from its shipped `npm-publish-plan.mjs`; current main no longer re-exports it, so every survivor lane failed before exercising the candidate.

## Verification

- original failure: full validation run `34200341661`, shared `SyntaxError` from `config-recipe.mjs`
- `pnpm exec vitest run test/scripts/docker-build-helper.test.ts --testNamePattern "uses each authorized frozen release.s shipped scenario contract files"`
- `bash -n scripts/e2e/upgrade-survivor-docker.sh`
- Oxfmt, OXLint, `git diff --check`

Production/tooling: +18/-0. Tests: +6/-0. The added lines represent the six existing transitive imports required to preserve one coherent selected-release module graph; no new compatibility mode or product runtime path is added.

### DIFF
--- a/scripts/e2e/upgrade-survivor-docker.sh
+++ b/scripts/e2e/upgrade-survivor-docker.sh
@@ -86,8 +86,26 @@ if [ "$UPGRADE_TARGET_TRAIN" = extended-stable ]; then
     exit 2
   fi
   UPGRADE_RUNNER="$UPGRADE_SCENARIO_DIR/run.sh"
+  UPGRADE_NPM_PUBLISH_PLAN="$(openclaw_resolve_frozen_target_file \
+    "$ROOT_DIR" scripts/lib/npm-publish-plan.mjs)"
+  UPGRADE_WINDOWS_HELPERS="$(openclaw_resolve_frozen_target_file \
+    "$ROOT_DIR" scripts/windows-cmd-helpers.mjs)"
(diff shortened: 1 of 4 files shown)


RomneyDa:
Fixed the actionable ClawSweeper P1 at `30bebd70452`: the frozen Windows helper now selects the source used by the shared harness mount instead of appending a duplicate destination. Added an executable combined-argument regression proving exactly one `/app/scripts/windows-cmd-helpers.mjs` mount, and updated the immutable-target fixture to include the complete selected helper closure. The exact old CI failures now pass locally.

@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

Fixed the exact-head compact-small-19 failure: removed the stale source-string assertion while retaining executable default and selected-source mount proofs. Full tooling-docker file passes (284 passed, 2 skipped).

PR #138545, opened 2026-09-04
fix: preserve tasks following text directives

steipete:
Closes #138530

<details>
<summary>Additional instructions</summary>

This branch is in `openclaw/openclaw`, available for maintainers to edit directly.

</details>

## Problem and fix

A text directive followed by a task could finish with an argument error or a settings acknowledgment instead of an agent reply. For example, `/exec security=deny ask=always Explain the output.` rejected `Explain`, while putting the task on the next line produced only an acknowledgment. The release Gateway regression also reproduced `/think high` followed by a multiline task completing without any provider request.

Text directives had acquired native commands' exclusive argument ownership and were parsed from canonical command text that discards later lines. This repair preserves complete text arguments during alias and bot-address normalization, then uses the existing inline directive parser. Native commands retain strict argument validation. The argument validator and metadata now describe command ownership rather than treating text as native.

Malformed text `/exec` arguments are classified from the original input before sibling directives can consume them. This prevents `/exec /think high` and `/exec /think high host=gateway` from applying sibling settings instead of rejecting `/think`. Leading spaces or newline indentation no longer produce an empty unexpected-argument diagnostic. Only the displayed diagnostic token is trimmed; task bytes remain intact.

## Behavior and scope

Valid text directives can precede same-line or multiline tasks, with code indentation, tabs, repeated spaces, and trailing whitespace preserved. Coverage includes colon syntax, aliases, addressed commands, multiple directives, disabled-text routing, per-turn overrides, and directives addressed to another bot. Invalid positional `/exec gateway` still receives a command error. No configuration, protocol, or persistence format changes.

Production grows by 20 lines and tests by 176 lines across the complete PR. The final follow-up contributes production +7 and tests +60: the parser adds six net lines to establish raw-argument ownership before destructive extraction, and the diagnostic expression wraps by one line. The preliminary parse reuses the canonical exec extractor and removes late classification; no separate argument policy is introduced. The QA scenario reference follows the validator rename to `directive-handling.arguments.ts`.

## Verification

Final source: `d0b816bc081317b60fe43d2a7fb9b608042fe6ba`. This is the exact tree tested before commit as base `5c6fbc77406dcb797bbc5c3af9ab65e73a7ef435` plus patch SHA-256 `4d394b0ef1b3097ec48430d86ca854d9b0f5f948d4c5462fcc82035bd52237bb`.

[...]

PR #142008, opened 2026-09-08
fix(telegram): preserve multiline text-directive commands at ingress

### LINKED ISSUE #141998: Telegram ingress drops later lines of multiline text-directive commands before the core boundary
## Summary

Named follow-up from #138545 ("preserve tasks following text directives"): "preserve multiline Telegram command projection at ingress. Telegram separately normalizes its command body before this core boundary (`extensions/telegram/src/bot-message-context.session.ts`). That preexisting producer behavior can discard later setting lines."

#138545 fixed the core boundary to re-normalize text-slash directive commands with `preserveArguments: true` (`src/auto-reply/reply/get-reply-directives.ts:301`), so `/think high\n<task>` keeps its task. On Telegram the fix cannot take effect: the ingress projection has already discarded the second and later lines before the core ever sees them.

## Evidence chain

1. `extensions/telegram/src/bot-message-context.session.ts:569` builds the command projection with `normalizeCommandBody(rawBody, { botUsername })` — no `preserveArguments`.
2. Without `preserveArguments`, `normalizeCommandBody` cuts the text at the first newline and resolves the multiline tail per command (`src/auto-reply/commands-registry-normalize.ts:33-43,87-93`): `skill`/`learn` keep the tail, `reset` flattens it — and every other registered command, including text directives like `/think`, **drops it**.
3. The truncated projection becomes `ctx.CommandBody` (same file, line 739), which the core prefers as `commandText` (`src/auto-reply/reply/inbound-context.ts:65-68`). The core's `preserveArguments` re-normalization in #138545 therefore receives `/think high` with the task already gone.

Reproduction: Telegram DM, send `/think high\n<task>` → the model receives only the acknowledgment-level directive; the task line never reaches the agent. Other channels that pass raw text through (or preserve arguments at ingress) are unaffected, which is why the core fix in #138545 could not cover Telegram.

## Native strictness is unaffected

The strict native-command view is computed independently at the core layer from the same text — `commands-context.ts:44` and `session-reset-command.ts:254` both call `normalizeCommandBody` without `preserveArguments`, so the per-command tail policy (including the `reset` flatten introduced in #117143) keeps applying exactly as today. Passing `preserveArguments: true` at the Telegram ingress changes only what the directive pipeline can see; it does not relax native argument ownership, matching #138545's "keep native argument ownership strict while preserving complete text directive arguments".

## Proposed repair

- `bot-message-context.session.ts:569`: pass `preserveArguments: true` — the same option the core boundary now uses for text-slash directives.
- Plugin regression test: multiline `/think high\n<task>` inbound context keeps the tail in `commandBody`; `/reset` multiline projection is preserved raw at ingress (flattening stays core-side), and command detection is unchanged.

Happy to open the PR.

Related: #138545 (named follow-up), #138530 (original report).

wangmiao0668000666:
## What Problem This Solves

With Telegram native commands disabled, `/think high` followed by a task on later lines could return only the settings acknowledgement. Telegram had discarded the task before core directive routing received it.

## Change

Preserve complete arguments at the existing Telegram command projection with `preserveArguments: true`. Core still owns directive removal, sender authorization, strict native arguments, and command-specific reset/skill/learn handling. No parser, configuration option, dependency or retry is added.

## Evidence

Built main `4f84c11f` and candidate `22f6d528` were exercised through Telegram's Test Server, a leased dedicated user and bot, the real Gateway, and a deterministic HTTP provider. Each paired plan retained one credential lease across its baseline/candidate phases.

| Scenario | Main | Candidate |
| --- | --- | --- |
| Text `/think high` plus multiline task | Settings acknowledgement, zero provider requests, no task transcript | One request containing the task once with high reasoning effort; one unique Telegram answer; matching user/assistant transcript |
| Native `/think high` with an extra task | Strict rejection, zero provider requests | Same strict rejection and zero provider requests |
| Directive only | Settings-only turn | Settings-only turn |

- Independent Telegram controls cover spacing, bot targeting, sender and mention gates, ordinary messages, reset, installed skill and learning inputs, photo captions, forum topics, replies, and stored history. Internal projection and unchanged-owner claims were checked separately against source.
- Telegram normalizes supplied CRLF to LF before OpenClaw receives it. The preserved task retains the recorded spacing, blank lines and LF endings. Reset keeps its strict core command view while preserving the raw task payload.
- Reply proof uses explicit per-turn mock responses with unique response IDs; the default mock could select an earlier conversation marker. Captured Telegram send confirmations bind temporary IDs to settled messages. Each paired run releases its credential and removes its temporary processes, ports and topic fixtures.
- 212 focused tests pass: 51 Telegram context/sibling cases and 161 shared normalizer, parser and routing cases. The new regression file fails twice on original production for the intended dropped-tail/flattened-projection assertions; its single-line control passes.
- Formatting, lint, line-count/environment-count and assertion-safety checks pass. Independent code review found no actionable P0–P2 defect. Exact-head CI run34201939949 completed successfully.

An exploratory overlapping-turn reply control exposed a missing reply target on the queued delivery path. A controlled early second turn reproduces the same failure on pinned main. The plain-text projection and downstream reply owners are unchanged; this existing defect is recorded as a separate follow-up. Ordinary reply proof waits for the first observed answer before sending the next turn.

The deterministic provider does not claim live model reasoning. Native commands remain distinct from text directives; this does not make task tails valid native arguments. Original loopback proof remains useful supporting evidence. AI-assisted maintainer validation preserves contributor authorship.

Fixes #141998.

### DIFF
--- a/extensions/telegram/src/bot-message-context.command-body.test.ts
+++ b/extensions/telegram/src/bot-message-context.command-body.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it } from "vitest";
+import { buildTelegramMessageContextForTest } from "./bot-message-context.test-harness.js";
+
+describe("buildTelegramMessageContext multiline command projection", () => {
+  const chat = { id: 999, type: "private" as const, first_name: "Alice" };
+  const sender = { id: 42, first_name: "Alice", is_bot: false };
+
+  it("preserves later lines of multiline text-directive commands", async () => {
+    const context = await buildTelegramMessageContextForTest({
+      message: {
+        message_id: 1,
+        chat,
+        from: sender,
+        text: "/think high\nsummarize the thread so far",
+        entities: [{ type: "bot_command", offset: 0, length: "/think".length }],
+      },
+    });
+
+    expect(context?.ctxPayload.CommandBody).toBe("/think high\nsummarize the thread so far");
+  });
+
+  it("keeps the raw multiline reset projection for core-side strict normalization", async () => {
+    const context = await buildTelegramMessageContextForTest({
+      message: {
+        message_id: 1,
+        chat,
+        from: sender,
+        text: "/reset\nextra context",
+        entities: [{ type: "bot_command", offset: 0, length: "/reset".length }],
+      },
+    });
+
+    // The ingress projection preserves the tail verbatim; the reset flatten policy
+    // is applied by the core command layer (session-reset-command / commands-context).
+    expect(context?.ctxPayload.CommandBody).toBe("/reset\nextra context");
+  });
+
+  it("keeps single-line command normalization unchanged", async () => {
+    const context = await buildTelegramMessageContextForTest({
+      message: {
+        message_id: 1,
+        chat,
+        from: sender,
+        text: "/status",
+        entities: [{ type: "bot_command", offset: 0, length: "/status".length }],
+      },
+    });
+
+    expect(context?.ctxPayload.CommandBody).toBe("/status");
+  });
+});
--- a/extensions/telegram/src/bot-message-context.session.ts
+++ b/extensions/telegram/src/bot-message-context.session.ts
@@ -568,6 +568,9 @@ export async function buildTelegramInboundContextPayload(params: {
   const hasGroupHistoryContext = isGroup;
   const commandBody = normalizeCommandBody(rawBody, {
     botUsername: normalizeOptionalLowercaseString(primaryCtx.me?.username),
+    // Preserve multiline text-directive arguments for the core boundary (#138545);
+    // native strictness is re-derived core-side from the same text.
+    preserveArguments: true,
   });
   const commandSource =
     options?.commandSource ??


wangmiao0668000666:
Proof gate addressed with a zero-credential loopback receipt (details in the updated Evidence section).

**Harness:** stub Bot API server (`getMe`/`deleteWebhook`/`getUpdates`/`sendMessage`) → real Telegram plugin ingress via `channels.telegram.apiRoot` → real gateway process (dist build) → `scripts/e2e/mock-openai-server.mjs` provider with an ndjson request log. One injected DM: `/think high\nPROOF_TASK_9F3E2D summarize the thread so far`. Synthetic token/user ids, `commands.native: false` for text-directive routing.

**Pre-fix build (`22d6219be8f`):** 0 provider requests; bot replied only `Thinking level set to high.` — the exact #138545 regression shape (acknowledgment, task silently gone).

**PR-head build:** `POST /v1/responses` contains `PROOF_TASK_9F3E2D summarize the thread so far`; the agent answered the task and the reply `OPENCLAW_E2E_OK` returned through the stub.

One disclosed delta from the same harness: with `commands.native: true`, the PR-head build now answers `Unexpected argument "PROOF_TASK_9F3E2D" for /think.` — strict native validation receives the complete input instead of never seeing the tail, matching #138545's "native commands retain strict argument validation".

@clawsweeper re-review

obviyus:
Thanks @wangmiao0668000666!

**Proof:** Real Telegram Test Server messages exercised the built Gateway on the baseline and exact PR head. A multiline text directive now reaches the provider once, produces one task answer, and retains matching history. Strict native argument rejection remains intact.

- **Bug:** With native commands disabled, Telegram discarded the task after `/think high` and returned only a settings acknowledgement.
- **Fix:** The Telegram producer preserves complete arguments through the existing option; core keeps command and authorization policy.

| Before | After |
| --- | --- |
| Settings acknowledgement; zero provider requests | Complete task once, high reasoning effort, one answer |

212 focused tests pass. Independent behavior validation and final source review passed for the repair. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34201939949) is green. The deterministic provider proves routing and delivery, not live model reasoning.

A separate reply-target failure under early overlapping turns was also reproduced on pinned main and remains a follow-up. Ordinary sequential replies pass on both versions. AI-assisted maintainer validation preserved contributor authorship.

wangmiao0668000666:
Thanks for the review and merge, @obviyus! Glad the ingress-side fix landed cleanly. Happy to follow up if anything related surfaces.

PR #136509, opened 2026-09-02
fix: doctor cannot load Teams state checker after a source build

steipete:
Closes #136499
Related: #135478, #124600

## What Problem This Solves

Fixes an issue where operators running doctor after a successful source-checkout build would see `core/doctor/channel-package-state-capabilities` report that the Microsoft Teams `configuredState` checker could not load, even when Teams was not configured.

## Why This Change Was Made

The checker was emitted as `configured-state.cjs`, but source-checkout package metadata still advertised the extensionless `./configured-state`. Node's native resolver does not infer `.cjs`. Main/setup entry metadata already followed the selected build format; the channel-state metadata did not.

The metadata producer now shares the standalone package transformation for both `configuredState` and `persistedAuthState`, naming the exact output selected by the canonical build plan. Checkout output uses the plugin root and standalone packages use plugin-local `dist/`. This keeps Docker-selected unified ESM, isolated CommonJS, env-only declarations, export names, and authored source manifests coherent. Unmapped module declarations fail at the build owner rather than advertising missing runtime files.

No runtime loader fallback, Teams authentication change, new dependency, configuration option, or schema change is introduced. The two standalone-only transformation layers become one shared implementation. The nearby manifest docs now accurately describe module-over-env precedence and emitted state-probe paths.

## User Impact

Doctor can load the Teams configured-state checker from a clean source build without a false missing-artifact warning. Other module-backed configured/auth-state capabilities retain their selected format, and standalone plugin packages retain their existing host-peer contract. Operators do not need a configuration workaround or manual dependency installation.

### Merged replacement

[...]

Issue #136499, opened 2026-09-02
fix: doctor cannot load the Teams checker after a clean source build


steipete:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

Clean source-checkout builds emit the Microsoft Teams configured-state checker as CommonJS but leave its package metadata extensionless, so doctor reports that the optional plugin's checker cannot load.

### Steps to reproduce

1. Build an affected OpenClaw source checkout with its normal dependency install and full package build (`pnpm install --frozen-lockfile`, then `pnpm build:package`).
2. Use an isolated HOME, `OPENCLAW_STATE_DIR`, and minimal synthetic configuration; Teams credentials or a running Gateway are not required.
3. Run `node openclaw.mjs doctor --lint --severity-min warning --json` against that isolated state.
4. Compare `dist/extensions/msteams/package.json` with the emitted `dist/extensions/msteams/configured-state.cjs`.

### Expected behavior

Every declared package-state checker resolves to its built artifact. Doctor should not report a missing Teams checker after a successful build. Standalone plugin packaging already maps both configured-state and persisted-auth metadata to their exact emitted files.

### Actual behavior

The built checker exists, exports the declared function, and returns the expected false/true results when loaded directly. However, the copied package metadata still declares `./configured-state`. Native Node resolution does not append `.cjs`, and doctor reports `core/doctor/channel-package-state-capabilities` for `msteams`.

### OpenClaw version

Source-checkout package metadata `2026.8.1`. The reproduced build used synchronized source whose affected producer and Teams manifest hashes match `f1ae53b7360f16250223f5edf74db727c9d96132`; those owners also match the reported source revision `b3fcdf31620d5c829554f65566a494385823fed6`. This is not a claim that the published stable host package reproduces it.

### Operating system

Linux, Node 24.19.0, isolated Blacksmith Testbox. Regression coverage also reproduced the resolution failure on macOS.

### Install method

Source checkout with pnpm; a clean full package build and the built CLI entrypoint.

### Model

Not applicable; no model invocation.

### Provider / routing chain

Not applicable; no provider or Teams API request.

### Logs

```text
emittedCheckerPresent: true
metadata specifier: ./configured-state
native resolution: MODULE_NOT_FOUND
checker results: empty=false, partial=false, complete=true

doctor checkId: core/doctor/channel-package-state-capabilities
target: msteams
Plugin msteams declared configuredState, but its checker failed to load:
[...]

PR #142016, opened 2026-09-08
fix(plugins): resolve extensionless specifiers to built .cjs runtimes

RomneyDa:
## What Problem This Solves

On any built install, the Microsoft Teams channel plugin cannot load its `configuredState` checker, and Doctor reports:

```
[warning] core/doctor/channel-package-state-capabilities - Plugin msteams declared
configuredState, but its checker failed to load: plugin module path not found:
<root>/dist/extensions/msteams/configured-state | ENOENT: no such file or directory,
lstat '<root>/dist/extensions/msteams/configured-state'
```

The artifact is present on disk as `configured-state.cjs`. The resolver cannot reach it.

## Why This Change Was Made

`resolvePluginModulePath()` resolves extensionless specifiers through `nodeRequire.resolve()`, which implicitly appends only `.js`, `.json`, and `.node` — never `.cjs`. It then falls back to source extensions (`.ts`, `.mts`, …), and otherwise returns the nonexistent extensionless path, which fails the boundary open as `plugin module path not found`.

All five plugins declaring `configuredState` use the same extensionless specifier `./configured-state`. Four build as ESM to `.js`, which `require.resolve` appends implicitly, so they resolve by virtue of their format. `msteams` is the only plugin with `build.runtimeFormat: "cjs"`, and `pluginRuntimeExtension()` deliberately emits `.cjs` for it — so it is the only plugin that lands in the gap.

The manifest convention (extensionless, format-agnostic) and the build output (`.cjs`) are both correct. The resolver is the piece that does not account for the extensions the plugin build can emit.

This reproduces only in built installs: from a source checkout the `.ts` fallback matches, which is why it does not surface in dev.

## User Impact

Operators running Teams get a persistent Doctor warning and degraded configured-state detection for the channel, with no available remedy — the file the resolver wants is present, under a name it cannot reach. Operators not running Teams see the warning as recurring noise in `openclaw doctor`.

## Evidence

Negative control — with the source fix reverted and the new tests kept, exactly the three new tests fail and the existing ten pass:

```
× resolves extensionless plugin module specifiers to built cjs runtimes
× resolves extensionless plugin module specifiers to built mjs runtimes
× loads a built .cjs runtime through an extensionless specifier
Tests  3 failed | 10 passed (13)
```

With the fix applied:

- `node scripts/run-vitest.mjs src/channels/plugins/module-loader.test.ts` — 13 passed
- `pnpm test:contracts:channels` — 13 files, 32 passed across 4 shards
- `oxfmt --check` — passed
- Node 24.16.0

Originally observed on a real 2026.8.1 git install on macOS via `openclaw doctor`, while recovering an unrelated failed update.

## Notes

- The new fallback runs *after* the existing source-extension fallback, so resolution precedence for every currently resolving specifier is unchanged. It only adds a path where resolution previously failed outright.
- `.mjs` is included alongside `.cjs` for symmetry, since Node implicitly appends neither. The plugin build currently emits only `.cjs` or `.js`.

### DIFF
--- a/src/channels/plugins/module-loader.test.ts
+++ b/src/channels/plugins/module-loader.test.ts
@@ -39,6 +39,35 @@ describe("channel plugin module loader helpers", () => {
     },
   );
 
+  it.each(["cjs", "mjs"])(
+    "resolves extensionless plugin module specifiers to built %s runtimes",
+    (extension) => {
+      const rootDir = tempDirs.make("openclaw-channel-module-loader-");
+      const expectedPath = path.join(rootDir, `configured-state.${extension}`);
+      fs.writeFileSync(expectedPath, "// built runtime artifact\n", "utf8");
+
+      expect(resolveExistingPluginModulePath(rootDir, "./configured-state")).toBe(expectedPath);
+    },
+  );
+
+  it("loads a built .cjs runtime through an extensionless specifier", () => {
+    const rootDir = tempDirs.make("openclaw-channel-module-loader-");
+    const expectedPath = path.join(rootDir, "configured-state.cjs");
+    fs.writeFileSync(
+      expectedPath,
+      "module.exports = { hasConfiguredChannelState: () => true };\n",
+      "utf8",
+    );
+
+    const modulePath = resolveExistingPluginModulePath(rootDir, "./configured-state");
+    expect(modulePath).toBe(expectedPath);
+
+    const loaded = loadChannelPluginModule({ modulePath, rootDir }) as {
+      hasConfiguredChannelState: () => boolean;
+    };
+    expect(loaded.hasConfiguredChannelState()).toBe(true);
+  });
+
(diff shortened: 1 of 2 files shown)


RomneyDa:
Closing as already fixed by #136509, merged as 16c07ee67e0e14d4acaaa9d06b88e796e95e1b8a and shipped in v2026.9.2. The canonical issue #136499 is also closed.

Current build metadata rewrites plugin catalog entries to the selected emitted runtime extension at `scripts/copy-bundled-plugin-metadata.mts:317`, so the reported Teams Doctor failure no longer reaches the loader in this form. Adding generic extensionless `.cjs`/`.mjs` loader fallback would now be a separate compatibility surface without a demonstrated supported contract.

If a distinct current workflow still requires extensionless built-module resolution, please open a new issue with a current-build reproduction.

PR #142017, opened 2026-09-08
fix(release): classify inconclusive Kova CPU evidence

RomneyDa:
## Summary

- recognize Kova-owned CPU intervals that straddle a threshold as harness-inconclusive rather than product regressions
- require the recorded lower bound to pass and reconcile every violation with its role measurements and blocking card
- keep true lower-bound violations, product-owned failures, other statuses, baseline regressions, and malformed evidence fail-closed

## Root cause

Kova deliberately emits `BLOCKED` when Linux process sampling can only bound CPU across a threshold. Full release validation treated that Kova-harness evidence gap as an OpenClaw product-performance blocker. The 9651.7% value was the interval upper bound from a short-lived process, while the independently recorded lower bound passed.

## Verification

- pre-fix adapter rejects the captured report: `gate verdict was not PARTIAL / DO_NOT_SHIP`
- post-fix exact captured report: `Tolerated Kova verdict: inconclusive-cpu-evidence`
- full `test/scripts/kova-report-gate.test.ts`: 124/124
- workflow owner regression, `tsgo:scripts`, Oxfmt, OXLint, diff check

Tooling/workflow: +88/-2. Tests: +69/-3. The validator is deliberately fail-closed and schema-bound; it does not alter product runtime or performance thresholds.

### DIFF
--- a/.github/workflows/openclaw-performance.yml
+++ b/.github/workflows/openclaw-performance.yml
@@ -651,7 +651,7 @@ jobs:
             then
               effective_status=0
               {
-                echo "Kova returned a non-zero release-gate verdict, but the trusted report adapter found only filtered coverage, profiling-affected resource thresholds, or a Kova-owned inconclusive CPU interval whose lower bound passed."
+                echo "Kova returned a non-zero release-gate verdict, but the trusted report adapter found only filtered coverage or profiling-affected resource thresholds with no baseline regression."
                 echo
               } >> "$GITHUB_STEP_SUMMARY"
             fi
--- a/.github/workflows/openclaw-performance.yml
+++ b/.github/workflows/openclaw-performance.yml
@@ -651,7 +651,7 @@ jobs:
             then
               effective_status=0
               {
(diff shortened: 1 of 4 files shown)


RomneyDa:
@clawsweeper re-review

Removed the workflow and workflow-test churn entirely. The repair is now owner-local to the trusted Kova report adapter: +78/-1 tooling, +68 tests. Focused adapter/workflow suite passes 166/166.

RomneyDa:
Closing this approach after owner-level investigation. Kova intentionally returns `BLOCKED` for an interval that cannot prove the threshold passed; OpenClaw must not reinterpret that as release success.

The producer defect is Kova’s CPU clock placement: it included the entire `ps` process census in `/proc` counter-scan uncertainty, widening the observed interval by ~79 percentage points. Owner fix: https://github.com/openclaw/Kova/pull/118 (production net -4). Once merged, OpenClaw should only update its reviewed Kova pin.

Issue #115908, opened 2026-07-29
Session transcript projection reconcile can livelock under sustained writes, blocking the main thread and stalling all channel transports


CanadaOrNaw:
Bug type: Core / sessions

Beta release blocker: No

## Summary

Under sustained transcript write load, the session transcript projection can enter a
non-converging rebuild cycle that occupies the Node main thread. Because every rebuild path is
synchronous, the event loop stalls for tens of seconds to minutes, and every long-poll channel
transport times out simultaneously and restarts.

Three properties combine:

1. **The read-side rebuild kick is unguarded at its call site.** `withCurrentProjectionSnapshot`
   calls `startSessionTranscriptIndexReconcile(...)` unconditionally whenever the projection is
   stale, then throws `SessionTranscriptProjectionUnavailableError`. The throw happens regardless
   of whether a rebuild is already in flight, so a polling reader gets an exception every time and
   re-issues the read.

2. **The coalescing guard is keyed per database file, not per session.** `startSessionTranscriptIndexReconcile`
   maintains a `runningReconciles` Map keyed by `resolveOpenClawAgentSqlitePath(params)`. A second
   call while one is running sets `pending = true` and returns. The loop is
   `while (true) { … if (state.pending) continue; … }`, so as long as *any* session in that
   database keeps going dirty during a pass, the shared loop restarts and re-scans the whole
   dirty set. One hot session can therefore hold the loop open across all sessions.

3. **A rebuild that races a write discards all of its work, with no backoff.** The prepared
   projection only commits if the source snapshot is unchanged:

   ```js
   function sourceSnapshotMatches(db, plan) {
     const session = …select("transcript_updated_at")…
     const latest  = …select("seq")…orderBy("seq","desc").limit(1));
     return session?.transcript_updated_at === plan.sourceTranscriptUpdatedAt
         && latest?.seq === plan.sourceIndexedSeq;
   }
   ```

   and `finalizePreparedSessionTranscriptProjectionInTransaction` returns `false` when it does not
   match. The claim is abandoned, `needs_rebuild` stays `1`, and the entire per-session
   `event_json` read plus `JSON.parse` must be redone from scratch. There is no backoff and no
   partial-progress retention between attempts.

Consequence: if the write rate for a hot session exceeds the completion rate of that session's
rebuild, the rebuild can never land. The work is re-done indefinitely while readers keep throwing,
and because none of it yields, the main thread is unavailable for the duration.

## Why it blocks the loop rather than just burning CPU

[...]

PR #142018, opened 2026-09-08
fix(sessions): converge projections during active writes

RomneyDa:
## Summary

- let transcript projection rebuilds claim a bounded, same-generation append-only tail instead of restarting whenever a live session advances
- reuse one append classifier for live indexing, suffix projection, and rebuild catch-up so branch changes, resets, side appends, legacy/canonical transitions, and parent mismatches still fail closed

This ports the convergence design from #116259 onto the current projection/reconcile architecture and addresses the sustained-write livelock described in #115908.

## Safety bounds

- at most 512 post-snapshot rows
- at most 256 KiB of serialized tail events
- transcript generation must be unchanged
- the full tail must be contiguous and append cleanly to the prepared cursor
- prepared sources with unresolved leaf controls cannot use catch-up
- partial base publication, rewrites, resets, branches, and oversized tails remain dirty for a fresh rebuild

## Compatibility and recovery

- no schema or stored-row format change; existing databases remain compatible
- exact-snapshot publication remains unchanged
- only bounded, same-generation append tails gain a successful publication path
- reverting this commit restores exact-snapshot-only publication; raw transcript events are never rewritten by this change

## Validation

- `pnpm check:changed` passed on the rebased tree
- 90 focused tests passed across projection catch-up/rebuild, replacement projection, memory reconcile, active events, search/cold archives, and exact suffix mutation
- the exact compact-memory CI regression passes for both null and unresolved initial leaf controls
- live recovery pass reconciled two previously stuck large projections (17,467 and 12,827 source events) to clean, exact-tip watermarks
- the final unrelated-main rebase passed the exact-head regression tests and full changed-file gate

## Review decision requested

Storage owner: please explicitly accept or reject the bounded same-generation catch-up publication contract above, including the compatibility, recovery, and rollback limits. This is the remaining material-store checkpoint identified by ClawSweeper.

## Worked on by

- Dallin Romney (@RomneyDa)
- CanadaOrNaw (@CanadaOrNaw)


RomneyDa:
Rebased and reconciled with current `main` at exact head `4745f5eeeeac2f86a2a5fafa5e72f71dc6c91fbd`.

@steipete @shakkernerd @vincentkoc: requesting storage-owner review and an explicit accept/reject decision on the bounded same-generation catch-up publication contract. The PR body now records the no-schema compatibility story, conservative rejection paths, recovery limits, rollback behavior, 86-test integration proof, and live large-session recovery evidence.

@clawsweeper re-review

RomneyDa:
Rebased the unchanged patch onto current `main` at exact head `9e907c2e762592e08ce3e2aafc4fdc2d677a6e90`. Patch ID is unchanged from the previously reviewed/green head; both leaf-control regressions and the full changed-file gate pass on this base. @clawsweeper re-review

PR #141775, opened 2026-09-08
fix(onboard): scope recommendations by agent

ChrisCantwell:
## What Problem This Solves

Onboarding recommendation commands told operators to pass `--agent <id>` when workspace selection was ambiguous, but did not accept that option. Add the shared selector and resolve its configured workspace before reading, acknowledging, retrying, or clearing the existing recommendation record.

Preserve explicit input through command registration. Reject blank and unknown IDs before store access, so an empty selector cannot fall through to the default workspace and a malformed ID cannot be rewritten into another configured agent. Omitted selection keeps the existing default behavior. The store schema, workspace keys, filtering, and concurrent-update checks stay with their existing owners.

## Evidence

Real CLI proof used two isolated configured agents, `writer` and `analyst`, with distinct workspaces and seeded records in the canonical store. Main was pinned at `f310ecc64128e33b3ace61e5a66f20ad8414be1f`.

Before:

```text
$ openclaw onboard recommendations --agent writer --json
OpenClaw does not recognize option "--agent".
Try: openclaw onboard recommendations --help
```

After:

```text
$ openclaw onboard recommendations --agent writer --json
[
  {
    "id": "writer-plugin",
    "source": "official-plugin",
    "tier": "recommended"
  },
  {
    "id": "@fixture/writer-skill",
    "source": "clawhub-skill",
    "tier": "optional"
  }
]
```

- Read, acknowledge, retry, and refresh select the requested workspace. Follow-up reads without reseeding confirm that mutations persist.
- The unselected workspace's exact record bytes and timestamps remain unchanged. These are workspace-keyed rows in a shared database, not separate per-agent recommendation files.
- Empty and whitespace-only IDs return `--agent must not be blank`; unknown IDs retain the `openclaw agents list` guidance. Invalid selectors cannot trigger legacy read-side cleanup or delete an offer.
- Safe install-ID filtering, selected legacy cleanup, parent-option rejection, structured JSON errors, ambiguous omission, and the unambiguous default remain covered.
- All 116 focused tests passed. Nine real-store invalid-selector cases and the empty-selector registration case fail on the original candidate for the intended reasons.
- Fresh independent CLI validation retained 47 complete captures: 12 observable clauses passed, none failed; source, concurrency, baseline identity, and final cleanup remain separately verified roles. Cumulative P0–P2 code review is clean.

[...]

PR #142019, opened 2026-09-08
fix(cli): accept agent selection after recommendation subcommands

obviyus:
## What Problem This Solves

Placing --agent after a recommendation subcommand failed with an unknown-option error even though the same selector worked before it.

## Why This Change Was Made

Register the selector for acknowledge and refresh, including retry. An explicit subcommand value wins; otherwise the existing option-source helper inherits the parent selection. Strict validation remains.

## User Impact

Users can place --agent beside the command they run. Only the selected agent’s recommendations change; blank or unknown selections cannot fall back to another agent.

## Evidence

Real compiled CLI refresh changed from exit 1 to exit 0 and cleared only the selected agent’s offer. Nine parser regressions failed before the fix; 133 focused registration, option-source, recommendation, and storage tests passed afterward. Independent command captures checked both placements, overrides, retry, invalid selection, omission, and unchanged other-agent records.

Related: #141775.

### DIFF
--- a/docs/cli/onboard.md
+++ b/docs/cli/onboard.md
@@ -48,6 +48,8 @@ openclaw onboard --skip-bootstrap
 openclaw onboard recommendations --json
 openclaw onboard recommendations --agent writer --json
 openclaw onboard recommendations --agent writer acknowledge
+openclaw onboard recommendations acknowledge --agent writer
+openclaw onboard recommendations refresh --agent writer
 openclaw onboard recommendations acknowledge
 openclaw onboard recommendations acknowledge --retry "<failed-id>"
 openclaw onboard recommendations refresh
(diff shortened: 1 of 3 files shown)


obviyus:
**Proof:** The compiled CLI now accepts `--agent` before and after recommendation subcommands. Both parent/leaf override directions select the intended workspace, invalid selections preserve stored offers, and subsequent reads confirm the changes persisted.

- **Bug:** `openclaw onboard recommendations refresh --agent analyst` failed with an unknown-option error. Acknowledge and retry had the same placement gap.
- **Fix:** Each applicable subcommand declares the selector and resolves an explicit leaf value before parent inheritance.
- **Validation:** 133 focused tests and 113 fresh independent CLI captures; clean final review. The prior parent-only coverage miss in #141775 remains recorded.

| Before | After |
| --- | --- |
| After-subcommand selector rejected, exit 1 | Same command succeeds, exit 0; only the selected offer changes |

Hosted CI passed: https://github.com/openclaw/openclaw/actions/runs/34202874319. Merged as `54eb0dce72507a9482adc041f3e3e81f357a19eb`.

PR #142024, opened 2026-09-08
fix: sync source assets with unusual filenames

### LINKED ISSUE #141996: Source CLI silently misses bundled skill edits with Unicode filenames
### Summary

A source checkout can silently leave a changed bundled skill out of its runtime artifacts when the filename contains Unicode. The ordinary command succeeds without synchronizing the file.

### Steps to reproduce

1. In an isolated trusted source checkout at `8954f104fb721c30d0f83c86a7cef6cd5f1df795`, run `pnpm openclaw --version` from genuinely absent build outputs. Let the normal runner finish its build and write its own stamps.
2. Add a small text file at `extensions/canvas/skills/canvas/qa-café.md`. Canvas already declares the containing skills directory.
3. Run `pnpm openclaw --version` again with Git's default `core.quotePath=true` behavior.
4. Inspect the matching files under `dist/extensions/canvas/skills/canvas/` and `dist-runtime/extensions/canvas/skills/canvas/`.

### Expected behavior

The runner recognizes the changed declared skill and runs its existing runtime synchronization so both artifact locations contain the exact source bytes. A skill-only edit should not require a full TypeScript rebuild.

### Actual behavior

The second command exits 0 and prints `OpenClaw 2026.9.2 (8954f10)`. Neither generated copy exists, and both genuine build stamps remain unchanged. The clean baseline and an unrelated file whose basename contains a literal POSIX backslash also exit successfully without rebuilding; these controls were run separately.

### Environment

Source checkout on macOS 27.0, Node 24.20.0, pnpm 12.3.4. No provider, live model, external channel, or operator Gateway is involved. The fixture and build outputs were isolated and retained.

### Cause and impact

The runner parses human-readable Git status as newline-delimited, trimmed paths. Git C-quotes unusual names in that output, so the classification no longer receives the original repository path. Existing real-Git regression cases also expose lost whitespace/newline/rename-side identity. This affects source development and bundled artifact freshness; it does not establish an installed npm-package failure or a general Chokidar filename-support claim.

The introducing commit and last known good version are unknown. The baseline and observed missing copies above are directly verified.

steipete:
Closes #141996

## What Problem This Solves

Fixes an issue where developers running a source checkout would silently keep stale bundled assets when changed filenames contain Unicode or other characters escaped by Git. For example, adding a declared Canvas skill named `qa-café.md` leaves both runtime copies absent even though `pnpm openclaw --version` succeeds.

## Why This Change Was Made

The runner now asks Git for NUL-delimited porcelain with rename detection disabled, preserving each original filename and both sides of a rename. The shared path normalizer converts only the native separator, and the watcher reuses it; redundant private normalization is removed. A connected mtime-path check now distinguishes actual parent segments from legitimate basenames beginning with `..`.

This keeps filename identity intact through the existing build/postbuild classifiers. It adds no parser fallback, configuration, state, dependency, schema or protocol surface. Production is +36/−34, net +2 across three scripts; tests are +176/−40, net +136. The small net growth includes explicit Git arguments and shared imports while removing duplicate normalization.

## User Impact

Declared skill edits with Unicode, whitespace or other unusual filename characters are recognized. Skill-only changes use the existing postbuild sync, and unrelated filenames do not trigger unnecessary work. JSON and CLI command behavior are unchanged. Chokidar still has its own upstream POSIX-backslash normalization in ignored callbacks; this does not claim to fix or live-verify every watcher filename.

## Evidence

- Original real-Git/classifier baseline: 13 intended failures and six controls. The NUL-only intermediate still failed the two backslash-boundary cases; the complete repair passed. A separate mtime case demonstrated the `..`-basename defect before its guard was corrected. Final runner, watcher and asset suites passed all 141 tests; complete changed-file checks passed.
- Six actual `pnpm openclaw --version` invocations all exited 0 and printed the expected version. The first created genuine baseline outputs/stamps through the normal missing-build path. Clean and outside-file controls stayed quiet. The baseline then missed the new declared Unicode file and left both generated copies absent.
- Normal final `pnpm build` passed in 5m37.4s. The independent outside-only command stayed quiet and preserved both stamps. Adding the Unicode source then triggered actual runtime postbuild, copied the exact 40-byte marker to both `dist` and `dist-runtime`, preserved the build stamp, and advanced only the runtime synchronization stamp.
- Both retained artifact snapshots were independently rehashed: 18 roots with 12,731 baseline entries and 15,427 final entries. Sources, existing private fixtures, owned synthetic files, command results and cleanup receipts were checked; the checkout returned to the exact five-file candidate diff. Exact nanosecond comparisons use original Python captures because JavaScript copies round those large integers.

- One fresh managed P0–P2 review was scoped-clean with no actionable findings; final source and all bound evidence stayed unchanged.

Proof is bound to base `8954f104fb721c30d0f83c86a7cef6cd5f1df795` plus patch `589730d909e1dec6a612e6fc4210863694c970cba12dfd08930ea256cf089a9c`, committed without source changes as `20c41847867af9a781ec8a1734bd5abeadaaa36d`. Baseline uses that base's production scripts with the final tests retained. The five touched files remain unchanged on inspected main `0604bbce80797482228b275eaacec34614942731`. This is real local source-runner/build/postbuild proof on macOS, with real Git; no live Gateway, external channel, installed npm package, all-platform or later-main execution is claimed. Introduction provenance is unknown.

### DIFF
--- a/scripts/run-node-watch-paths.mts
+++ b/scripts/run-node-watch-paths.mts
@@ -44,30 +44,23 @@ const ignoredRunNodeRepoPathPatterns = [
 ];
 const extensionSourceFilePattern = /\.(?:[cm]?[jt]sx?)$/;
 
-/** Normalizes watch paths to repository-style POSIX separators. */
+/** Canonicalizes native paths without treating POSIX filename backslashes as separators. */
 export const normalizeRunNodePath = (filePath: unknown): string =>
-  (typeof filePath === "string" ? filePath : "").replaceAll("\\", "/");
+  (typeof filePath === "string" ? filePath : "").replaceAll(path.sep, "/").replace(/^\.\/+/, "");
 
-const isIgnoredSourcePath = (relativePath: string): boolean => {
-  const normalizedPath = normalizeRunNodePath(relativePath);
-  return (
-    normalizedPath.endsWith(".test.ts") ||
-    normalizedPath.endsWith(".test.tsx") ||
-    normalizedPath.endsWith("test-helpers.ts")
-  );
-};
+const isIgnoredSourcePath = (relativePath: string): boolean =>
+  relativePath.endsWith(".test.ts") ||
+  relativePath.endsWith(".test.tsx") ||
+  relativePath.endsWith("test-helpers.ts");
 
-const isBuildRelevantSourcePath = (relativePath: string): boolean => {
-  const normalizedPath = normalizeRunNodePath(relativePath);
-  return extensionSourceFilePattern.test(normalizedPath) && !isIgnoredSourcePath(normalizedPath);
-};
+const isBuildRelevantSourcePath = (relativePath: string): boolean =>
+  extensionSourceFilePattern.test(relativePath) && !isIgnoredSourcePath(relativePath);
 
 const isRestartRelevantExtensionPath = (relativePath: string): boolean => {
-  const normalizedPath = normalizeRunNodePath(relativePath);
-  if (extensionRestartMetadataFiles.has(path.posix.basename(normalizedPath))) {
+  if (extensionRestartMetadataFiles.has(path.posix.basename(relativePath))) {
     return true;
   }
-  return isBuildRelevantSourcePath(normalizedPath);
+  return isBuildRelevantSourcePath(relativePath);
 };
 
 const isRelevantRunNodePath = (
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [20c41847867af9a781ec8a1734bd5abeadaaa36d](https://github.com/openclaw/openclaw/pull/142024/commits/20c41847867af9a781ec8a1734bd5abeadaaa36d)
- Landed commit: [2361821ec1ae73e0104843e7d9d41d79e9889720](https://github.com/openclaw/openclaw/commit/2361821ec1ae73e0104843e7d9d41d79e9889720)

PR #142039, opened 2026-09-08
fix(ui): update the expanded file preview tooltip

### LINKED ISSUE #142014: Agent file preview tooltip keeps saying Expand after expansion
### Summary

In Settings → Agents → Files, expanding a file preview leaves its visible action tooltip saying “Expand preview” even though the button now collapses it.

### Steps to reproduce

1. Open a Markdown file in an agent's Files settings and choose Preview.
2. Hover the expand action; its tooltip says “Expand preview.”
3. Click it to expand the preview, then hover the same action again.

### Expected behavior

The visible caption should say “Collapse preview” while expanded, matching the action and its accessible name. Closing or resetting the preview should restore “Expand preview.”

### Actual behavior and evidence

At source `8d11b06ba16e36d4893149a30976b1e26bcae093`, the expanded button has pressed state and accessible name “Collapse preview,” but its explicit tooltip retains “Expand preview.” A real headless Chromium run through the normal routed, production-bundled Control UI reproduced this with synthetic Gateway/file data and retained before/after captures. The expected-correct caption assertion failed for that mismatch; the focused component regression failed for the same reason.

The state helper updates native button title while the renderer owns an explicit tooltip component whose content is separate. The title adapter intentionally avoids creating another tooltip inside that wrapper.

### Environment and scope

macOS 27.0, Node 24.20.0, pnpm 12.3.4. No real agent files or Gateway were used. This is a caption-state defect, not a claim of a general positioning, clipping, or editor-persistence defect. The introducing commit and last known good version are unknown.

tooltip-collapse-expanded

steipete:
Closes #142014

## What Problem This Solves

Fixes an issue where users expanding an agent file preview would still see “Expand preview” in its tooltip, even though the button now collapses the preview. Its accessible name was already correct.

## Why This Change Was Made

The expand/reset owner now updates the explicit tooltip component's reactive content instead of the button's native title. The existing title adapter intentionally avoids competing with that wrapper. This replaces one production line and preserves the current modal, focus, editor, and translation behavior.

## User Impact

The visible hint says “Collapse preview” while expanded and returns to “Expand preview” after collapse, Close, Edit, or Escape resets. No setting, persistence, protocol, dependency, or generic positioning behavior changes. The attached before/after images show the same expanded state with synthetic file content.

## Evidence

- The existing component regression failed on the original caption mismatch; all 23 component tests pass after the repair. Both existing real-browser focus/typing tests also pass.
- A production-bundled Control UI in real headless Chromium reproduced the visible mismatch through the normal agent-files route. The final case verifies all four reset paths, the actual tooltip anchor, retained file/editor identity, and absence of file/config writes through mocked Gateway transport. Baseline and final were repeated against the final test assertions: one intended baseline failure, one passing final case, with two unrelated cases unselected.
- Complete changed-file checks passed. Earlier attempts caught a test fixture array-access type error and two text-reader lint errors; the assertions were corrected without suppressions. The final UI production source and captured behavior are unchanged by those test-only corrections.
- Independent media review inspected all 12 original PNGs and 60 sampled frames across both recordings. One managed P0–P2 review, automatically split into two passes, was scoped-clean with no findings. Production is +1/−1; tests add 146 net lines across the existing component and routed browser suites.

Evidence is bound to source base `8d11b06ba16e36d4893149a30976b1e26bcae093` and patch `edf31617f68ceb8c9294df757739972c13edc4d2d0d4612507eb92bba8a5a33f`. Committed head `5097c05ed1534d81bb3f112f1f41418a1f0adabc` preserves those exact reviewed source bytes. The affected tooltip, renderer and test sources are unchanged on inspected main `0604bbce80797482228b275eaacec34614942731`. This proves the bundled UI with synthetic Gateway/file data; it does not claim a live Gateway, general geometry repair, live locale-switch repair, or editor-persistence change. Existing wording is restored, so no documentation or translation change is needed.

before-expanded-tooltip

after-expanded-tooltip

### DIFF
--- a/ui/src/e2e/agent-file-lifecycle.e2e.test.ts
+++ b/ui/src/e2e/agent-file-lifecycle.e2e.test.ts
@@ -1,5 +1,7 @@
 // Control UI E2E tests cover visible agent-file save outcomes and agent ownership.
+import type WaTooltip from "@awesome.me/webawesome/dist/components/tooltip/tooltip.js";
 import { expect, it } from "vitest";
+import { waitForControlUiProofSurface } from "../test-helpers/control-ui-e2e-screenshot.ts";
 import { installMockGateway } from "../test-helpers/control-ui-e2e.ts";
 import {
   agentFileProofDir,
@@ -10,6 +12,7 @@ import { createControlUiE2eSuite } from "./control-ui-e2e-suite.test-support.ts"
 
 const suite = createControlUiE2eSuite({
   name: "Control UI agent file lifecycle",
+  browserLaunchOptions: { headless: true },
   startServerBeforeBrowser: true,
   unavailableMessage: (executablePath) =>
     `Playwright Chromium is not available at ${executablePath}`,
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5097c05ed1534d81bb3f112f1f41418a1f0adabc](https://github.com/openclaw/openclaw/pull/142039/commits/5097c05ed1534d81bb3f112f1f41418a1f0adabc)
- Landed commit: [85b9ad8827c278591fa4a0b0bf00ed9a095b84fb](https://github.com/openclaw/openclaw/commit/85b9ad8827c278591fa4a0b0bf00ed9a095b84fb)

PR #142043, opened 2026-09-08
fix(ui): avoid rejected read updates in restricted sessions

steipete:
## What Problem This Solves

Fixes an issue where viewing a restricted session sent background read acknowledgements that the session's participation rules reject. Refreshing the session could repeat the failed update even though the composer already correctly showed that the viewer could not act.

## Why This Change Was Made

Automatic read acknowledgements now consult the existing session participation tracker before latching an unread update. Read-only and suggestion viewers, and draft viewers or members, stay silent. Eligible members, owners, administrators and uncapped shared viewers keep their existing acknowledgement behavior.

## User Impact

Opening or refreshing a known restricted session preserves its unread marker without sending the rejected background update. If membership changes to permit updates, the same marker can be acknowledged normally. Manual unread reminders and suggestion controls retain their existing behavior.

## Evidence

- The unchanged producer failed four new owner regressions: each restricted case sent one `unread: false` patch. The bundled browser flow also failed by observing that outgoing request for a read-only viewer.
- After the fix, all 26 owner/tracker tests and 12 bundled-browser tests passed. Browser coverage includes fresh restricted rows, restored membership, uncapped shared viewers, manual unread reminders and suggestions.
- The complete changed-file gate passed, including core/UI typechecks, lint, i18n and dead-export checks. Independent P0–P2 review found no actionable issues.
- The Linux [Testbox run](https://github.com/openclaw/openclaw/actions/runs/34202889187) produced a report with all six unread browser cases passed on Chromium 151 / Node 24.19.0, with matching source and lockfile hashes. The wrapper's artifact-return protocol failed after the report was written; a separate successful download recovered all 21 files, which passed independent inventory, source-tree and capture audit. The failed wrapper result is retained separately from the six successful test results.

The attached before/after captures contain synthetic sessions. Outgoing-request assertions establish the behavior change; the visible read-only presentation and unread marker are intentionally preserved.

Before: restricted viewer sends a rejected read acknowledgement

After: restricted viewer preserves unread state without that request

### DIFF
--- a/docs/web/control-ui/sessions-and-sidebar.md
+++ b/docs/web/control-ui/sessions-and-sidebar.md
@@ -78,6 +78,8 @@ Enable **Hide empty groups** in the same menu to hide custom groups with no sess
 
 **Mark as unread** creates a reminder that remains unread while the current chat stays open, including while a run streams or completes. Leave and reopen the session, or choose **Mark as read**, to clear it.
 
+Opening a read-only or suggestion session as a viewer leaves its unread marker intact. Draft sessions acknowledge reads automatically only for their owner or an administrator.
+
 **Delete** removes the confirmed selection from loaded session lists immediately and leaves any deleted conversation that is open. The Gateway finishes deletion in the background, safely stopping and reclaiming an attached cloud worker first. If deletion fails, the affected session can reappear with an error; other successful deletions and any navigation you made in the meantime are preserved. Browser drafts are retired only after deletion is confirmed, not while the request is pending.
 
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c7c73bb3c1b5fe60565de20f48f51a152e5e0170](https://github.com/openclaw/openclaw/pull/142043/commits/c7c73bb3c1b5fe60565de20f48f51a152e5e0170)
- Landed commit: [3c79197f5d1d3efe9bb596e742946152219b4132](https://github.com/openclaw/openclaw/commit/3c79197f5d1d3efe9bb596e742946152219b4132)

PR #142060, opened 2026-09-08
fix(qa): finish stateful settled-tool continuations

RomneyDa:
## Summary

- recognize a settled-tool continuation before deciding that missing inline tool output means the write is still pending
- keep the exhaustion fixture empty and the ordinary first request write-producing
- cover the Responses API stateful continuation shape where prior tool output is provider state rather than repeated input

## Root cause

Full validation's Telegram scenario observed two write plans. The runtime correctly disabled tools and sent the settled-tool continuation instruction, but the mock provider first tested for inline tool output. Stateful Responses continuations need not repeat that output, so the fixture planned the write again.

## Verification

- exact pre-fix focused regression: 2 failures
- post-fix focused mock-provider suite: 3 passed
- Oxfmt, OXLint, diff check

Production: +7/-8 (net -1). Tests: +9. The live scenario's one-write assertion remains unchanged.


RomneyDa:
@clawsweeper review

PR #142061, opened 2026-09-08
fix(release): omit unsupported corrupt-plugin update proof

RomneyDa:
## Summary

- classify corrupt-plugin update admission as unsupported when an authorized frozen target lacks its owning preflight module
- consolidate the three source-qualified Docker omission rules through one planner loop
- keep current, supported, and omission-disabled targets strict

## Root cause

The current corrupt-plugin scenario requires the core-update admission behavior introduced by #140778. The 2026.6.35 target predates that owner. Its historical scenario also installs the moving latest baseline and is no longer a valid proof for the frozen product. Full validation therefore demanded a refusal that the selected release cannot implement.

Backporting #140778 would import a 673-line product/doctor/plugin-update feature neighborhood. This planner repair records the scenario as not applicable only under the existing authorized frozen-target omission contract.

## Verification

- exact pre-fix planner regression failed with the lane still scheduled
- post-fix Docker planner suite: 103/103
- supported target retains the lane
- omission-disabled target retains the lane
- Oxfmt, OXLint, diff check

Production/tooling: +30/-21 (net +9), including consolidation of two existing omission branches. Tests: +26.

### DIFF
--- a/scripts/lib/docker-e2e-plan.mts
+++ b/scripts/lib/docker-e2e-plan.mts
@@ -339,6 +339,13 @@ function supportsMobilePairingReconnectForTarget(targetRoot: string | undefined)
   );
 }
 
+function supportsCorruptPluginUpdateForTarget(targetRoot: string | undefined): boolean {
+  return (
+    !targetRoot ||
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper review

PR #142070, opened 2026-09-08
feat(cli): add --offset and --limit pagination to automations list

tzy-17:
## What Problem This Solves

Operators with many automations have no way to page through `openclaw automations list` (alias `cron list`) from the CLI. The command always fetches and renders the entire inventory by internally walking every bounded Gateway page, even when the caller only needs a slice or wants to walk the inventory incrementally. Scripts that want `N` jobs starting at an offset must download everything and slice it locally.

The underlying `cron.list` RPC already supports `offset`, `limit`, and returns `total`/`hasMore`/`nextOffset` pagination metadata, but none of that is reachable through the CLI.

## Why This Change Was Made

Add `--offset <n>` and `--limit <n>` (1-200) to `openclaw automations list`. When either flag is present, the CLI performs a single bounded Gateway page instead of auto-paginating across all pages, and the JSON output exposes `total`, `offset`, `limit`, `hasMore`, and `nextOffset` so scripts can walk the inventory manually.

- Default behavior is unchanged and safe: with no flags, the command still returns every matching automation exactly as before.
- Pure CLI plumbing onto an existing RPC — no new config, schema, protocol, storage, or dependency.
- `--offset` and `--limit` follow the same validation style as the existing `runs`/`show` command (`--limit` capped at 200, non-numeric values rejected).

## User Impact

Operators and scripts can now page the automation inventory without pulling the whole list:

```bash
openclaw automations list --offset 0 --limit 50   # first page
openclaw automations list --offset 50 --limit 50  # next page
openclaw automations list --limit 50 --json       # page + pagination metadata
```

## Evidence

Focused tests added against the real Gateway pagination contract (`src/cli/cron-cli/cron-pagination.gateway.test.ts`):

- returns a single bounded page when `--limit` is provided
- offsets the single returned page when `--offset` is provided
- rejects a non-numeric `--limit` / `--limit` above 200 / non-numeric `--offset`

Local validation (Node 26.8.1):

```
node scripts/run-vitest.mjs src/cli/cron-cli/cron-pagination.gateway.test.ts   # 30 passed
node scripts/run-vitest.mjs src/cli/cron-cli/register.cron-edit.test.ts src/cli/cron-cli/register.cron-simple.test.ts  # 123 passed
git diff --check  # clean
oxfmt --check     # clean
```

Known gap: the structured `$autoreview` helper could not produce a clean verdict in this environment (Codex engine output truncated), so this PR relies on the focused test suite, formatting, and manual review above; CI will run the full gate.

Related: none (no existing issue tracks CLI-side pagination on `automations list`; the neighboring `runs`/`show` commands already expose their own paging flags).

---

Fixes / addresses feature request: #142142

## Real Behavior Proof (running Gateway + real CLI)

A real, isolated OpenClaw Gateway process was spawned on a free localhost port
with an isolated state directory (`createOpenClawTestInstance`). Five real cron
jobs (`proof-job-0`…`proof-job-4`) were added through the **real** CLI
(`openclaw cron add …`), then the **real** `automations list` CLI was driven
against the same Gateway — no mocked transport.

Proof test: `test/automations-list-pagination.proof.e2e.test.ts` — ran green
twice (test, 30.8s each).

| CLI invocation (all `--json`) | jobs returned | metadata |
|---|---|---|
| `automations list` (no flags) | 5 (all) | unchanged full inventory |
| `automations list --limit 2` | 2 | total=5, limit=2, offset=0, hasMore=true, nextOffset=2 |
| `automations list --offset 1 --limit 2` | 2 | total=5, offset=1, hasMore=true, nextOffset=3 |
| `automations list --offset 4 --limit 2` | 1 (tail) | offset=4, hasMore=false, nextOffset=null |

This shows bounded pages and metadata from a running Gateway after the fix,
plus the unchanged default (no-flags) listing. All job ids are synthetic
(`proof-job-N`); no credentials, IP addresses, private endpoints, or personal
data appear.

### Additional ClawSweeper P2 fixes in this head

- **Accept the Gateway-clamped terminal offset**: the real Gateway clamps a
  requested offset above `total` to `total` and returns an empty terminal page;
  the CLI now accepts that contract-defined clamp instead of throwing.
  Regression: `accepts a Gateway-clamped terminal offset for an empty single page`.
- **Validate single-page continuation metadata**: a single page now enforces
  the same cursor/terminal invariants as the multi-page path.
  Regression: `rejects a single-page continuation page with a non-advancing cursor`.
- **Narrow first-job reads** with `expectDefined` for `noUncheckedIndexedAccess`.

Focused suite: `src/cli/cron-cli/cron-pagination.gateway.test.ts` — **34/34
passed** through the real `cronHandlers` and real `listPage`.

### Additional fix (Revision 5 finding: validate terminal row coverage)

A canonical terminal single page must cover the advertised inventory from its
offset — `offset + jobs.length` must reach `total` — with the sole exception of
the Gateway-clamped empty terminal page (`offset === total`, requested offset
beyond inventory end). Regression:
`rejects a truncated canonical terminal single page that does not reach total`.
Focused suite: `src/cli/cron-cli/cron-pagination.gateway.test.ts` — **35/35
passed**.

### Revision 7 fixes (defined offset + exact terminal coverage + fixture typing)

Two ClawSweeper P2 findings addressed:

- **Require a defined offset and exact terminal row coverage**: a canonical
  terminal single page now requires `page.offset + page.jobs.length === total`
  for every canonical terminal page. The previous `reachedTotal` shortcut
  (`offset === total`) could accept a malformed page whose rows exceeded the
  advertised inventory; a valid Gateway-clamped empty page (`offset === total`,
  `jobs.length === 0`) satisfies the equality naturally, so no special case is
  needed. `page.offset` is explicitly narrowed (it is guaranteed defined for a
  canonical page) so the arithmetic satisfies strict TypeScript.
  Regressions: `rejects a canonical terminal single page whose rows exceed the
  advertised total` and `rejects a truncated canonical terminal single page that
  does not reach total`.
- **Narrow the regression fixture's jobs before indexing**: the terminal-page
  fixture now narrows `jobs` with `Array.isArray` before indexing the first
  element, fixing the strict-TypeScript error on indexing an `unknown` value.

Focused suite: `src/cli/cron-cli/cron-pagination.gateway.test.ts` — **36/36
passed**; full `src/cli/cron-cli/` — **328/328 across 12 files passed**.

### DIFF
--- a/docs/cli/cron.md
+++ b/docs/cli/cron.md
@@ -318,6 +318,7 @@ Manual run and inspection:
 ```bash
 openclaw automations list
 openclaw automations list --agent ops
+openclaw automations list --offset 50 --limit 25
 openclaw automations get <job-id>
 openclaw automations get <job-id> --json
 openclaw automations show <job-id>
@@ -334,6 +335,12 @@ openclaw automations runs <job-id> --run-id <run-id>
 `--id <job-id>` form remain supported compatibility aliases.
 
 `openclaw automations list` shows enabled jobs across agents by default, including jobs whose owner cannot be resolved. Pass `--all` to include disabled jobs, or `--agent <id>` to filter by the effective normalized agent ID. Ownership resolves from the job's declared agent, its agent-scoped session key, then the configured system-agent owner. Unresolved jobs do not match an agent filter. The `cron list` alias has the same behavior.
+
+By default `automations list` returns every matching automation across all bounded
+Gateway pages. Pass `--offset <n>` to skip the first `n` automations and
+`--limit <n>` (1-200) to cap the result to a single page; the JSON output then
+also includes `total`, `offset`, `limit`, `hasMore`, and `nextOffset` so scripts
+can walk the inventory manually instead of pulling it all at once.
 
 The human-readable Agent ID column shows the effective owner. JSON list rows preserve the declared `agentId` and include `effectiveAgentId`, which is `null` when ownership is unresolved.
 
--- a/src/cli/cron-cli/cron-pagination.gateway.test.ts
+++ b/src/cli/cron-cli/cron-pagination.gateway.test.ts
@@ -333,12 +333,35 @@ describe("cron CLI with the real Gateway pagination contract", () => {
     installRealCronGateway(Array.from({ length: 201 }, (_, index) => createJob(index)), {
       transformListPage(page) {
         const response = page as Record<string, unknown>;
+        const responseJobs = Array.isArray(response.jobs) ? response.jobs : [];
+        const firstJob = responseJobs[0];
         // A canonical terminal page advertises total=2 but returns only one job
         // from offset 0; scripts must not treat a partial page as complete.
         return {
           ...response,
-          jobs: [response.jobs[0]],
+          jobs: [firstJob],
           total: 2,
+          hasMore: false,
+          nextOffset: null,
+        };
+      },
+    });
+
+    await expect(runCron(["list", "--json", "--limit", "50"])).rejects.toThrow("exit 1");
+  });
+
+  it("rejects a canonical terminal single page whose rows exceed the advertised total", async () => {
+    installRealCronGateway(Array.from({ length: 201 }, (_, index) => createJob(index)), {
+      transformListPage(page) {
+        const response = page as Record<string, unknown>;
+        const responseJobs = Array.isArray(response.jobs) ? response.jobs : [];
+        // A malformed page claims offset === total (3) but still returns a job
+        // beyond the inventory; offset + jobs.length (4) must not equal total.
+        return {
+          ...response,
+          jobs: responseJobs.slice(0, 1),
+          offset: 3,
+          total: 3,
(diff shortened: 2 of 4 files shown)


tzy-17:
Addressed ClawSweeper's code findings:

- Validate the single-page response before returning it, so a malformed modern response (e.g. missing `jobs`) fails closed instead of becoming an empty inventory.
- Preserve an explicitly supplied `--offset 0` — it now selects single-page mode instead of falling back to the full walk.
- Keep a legacy page's unavailable `total` unknown rather than reporting the page length as the inventory size.
- Narrow the first-job read before accessing its ID.

Added regressions for zero-offset single-page and legacy-`total` behavior.

@clawsweeper re-review

tzy-17:
@clawsweeper re-review

tzy-17:
@clawsweeper re-review

tzy-17:
@clawsweeper re-review

PR #142079, opened 2026-09-08
refactor(tooling): remove TUI watcher timing overrides

steipete:
## What Problem This Solves

The TUI PTY watcher's child-stop helper exposes two timing overrides used only by a test. That test already uses fake timers, so the overrides add tooling machinery while hiding the real defaults from the regression.

## Why This Change Was Made

Use the existing 500 ms and 5,000 ms grace constants directly, retain the useful signal callback as a direct parameter, and advance the existing test's fake clock through those defaults. Remove the one-property options bag along with the timing overrides.

## User Impact

The watcher's configured grace periods, signal order, cancellation, process ownership, mirror draining, and terminal restoration are unchanged. Tooling decreases by nine lines and tests by four; no cases or assertions are removed.

## Evidence

- All 60 dev-tooling cases pass before and after, with identical inventories.
- Both selected existing PTY namespace cases pass before and after; the other 73 namespace-suite cases are filtered out in these focused runs.
- Temporarily change only the SIGTERM default from 500 to 501 ms: the old named test passes, while the revised named test fails at the missing SIGTERM assertion after 500 ms. The other 59 dev-tooling cases are filtered out in each control. Exact source bytes were restored before the final passing suites.
- Formatting, diff checks, independent Codex review through P2, and all 19 normal changed-file checks passed ([Linux Testbox run](https://github.com/openclaw/openclaw/actions/runs/34209403001)). The retained source bundle matches the reviewed two-file tree.
- The delegated command passed and the lease completed; its capsule and recorded processes are gone. A later portal-sync timeout is recorded separately.

Fake-clock proof establishes default usage and signal ordering, not a wall-clock speedup or universal platform signal timing. The existing real watcher namespace cases retain process and cleanup coverage.

### DIFF
--- a/scripts/dev/tui-pty-test-watch.ts
+++ b/scripts/dev/tui-pty-test-watch.ts
@@ -130,23 +130,14 @@ function currentTerminalDimension(value: number | undefined, fallback: number): 
 
 function createChildStopper(
   child: KillableChild,
-  options: {
-    signalChild?: SignalChild;
-    sigtermGraceMs?: number;
-    sigkillGraceMs?: number;
-  } = {},
+  signalChild: SignalChild = (targetChild, signal) =>
+    terminateManagedChild(targetChild, signal, {
+      onChildSignalError(error) {
+        throw error;
+      },
+      taskkillTimeoutMs: null,
+    }),
 ): ChildStopper {
-  const signalChild =
-    options.signalChild ??
-    ((targetChild, signal) =>
-      terminateManagedChild(targetChild, signal, {
-        onChildSignalError(error) {
-          throw error;
-        },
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [51a4b50c40254abbd4baa1fb935f6bb3d0e614e5](https://github.com/openclaw/openclaw/pull/142079/commits/51a4b50c40254abbd4baa1fb935f6bb3d0e614e5)
- Landed commit: [bda64076747253e0e5f2c3e268f634b1c3599f65](https://github.com/openclaw/openclaw/commit/bda64076747253e0e5f2c3e268f634b1c3599f65)

PR #126240, opened 2026-08-19
refactor(update): simplify lifecycle transactions

steipete:
## What Problem This Solves

The update, uninstall, and installer lifecycle paths had accumulated duplicate transaction bookkeeping: repeated package-update failure objects, a one-use shim-backup abstraction, repeated uninstall error aggregation, unused cleanup `skipped` state, and three separate native parsers for npm lifecycle-script identity. That duplication made the lifecycle boundaries harder to audit and easier to drift while preserving no distinct product behavior.

## Why This Change Was Made

This maintainer-authorized refactor moves rollback and failure bookkeeping back into the transactions that own them, simplifies uninstall aggregation while retaining the `serviceSafe` and `workspaceBlocked` safety gates, removes unused cleanup result state, and gives all three native installers the same self-contained Node identity kernel. Strict npm version parsing, package identity policy, rollback behavior, and independent cleanup continuation remain unchanged.

This is AI-assisted maintainer work. No changelog entry is included because observable behavior is intentionally unchanged.

## User Impact

There is no intended user-visible behavior change. Package updates still roll back failed shim replacement, uninstall still protects workspace/state cleanup when service or state safety cannot be established, and Bash/PowerShell installers still reject malformed npm versions before mutation. The result is a smaller lifecycle implementation with the same package-boundary behavior.

Production LOC: +162/-291 (net -129). Tests/test support: +54/-29 (net +25). CI/tooling: +0/-7.

## Evidence

Local proof on the rebased patch:

- `node scripts/run-vitest.mjs src/infra/package-update-steps.test.ts src/infra/package-update-steps.pnpm11-guard.test.ts src/infra/update-global.test.ts src/infra/update-global.pnpm11-discovery.test.ts` — 268 passing across the package-update owner and sibling paths.
- `node scripts/run-vitest.mjs src/commands/uninstall.test.ts src/commands/cleanup-utils.test.ts` — 98 passing, including four live-state cleanup tests.
- `node scripts/run-vitest.mjs test/scripts/install-sh.test.ts test/scripts/install-cli.test.ts test/scripts/install-ps1.test.ts` — 434 passing across installer/package acceptance.
- `bash -n scripts/install.sh scripts/install-cli.sh` and ShellCheck — clean.
- PowerShell parser validation and the real `pwsh` installer suite — clean.
- `pnpm exec oxfmt --check` on the changed TypeScript files and `git diff --check` — clean.
- `node scripts/check-changed.mjs` — all 30 selected phases green.
[...]

PR #124013, opened 2026-08-15
fix(ui): keep the chat transcript anchored while the pane resizes

steipete:
## What Problem This Solves

Resizing the chat pane (window resize, sidebar/split drag, dashboard dock) made the transcript text jump wildly: the reader was teleported to different messages and the view kept shifting for several frames while row measurements trickled back in. Reported live on team.openclaw.ai running ~main.

**Before** (viewport resized 1280→1100→960→820→back, reader mid-transcript):

https://github.com/user-attachments/assets/78b7464a-58ad-4f96-94b3-c23954de6beb

**After** (same sequence — the anchor row holds still while text re-wraps):

https://github.com/user-attachments/assets/0cc071fc-40a0-4183-9681-433786c76487

## Why This Change Was Made

The width-change path in the transcript virtualizer host called `instance.measure()`, wiping the entire item-size cache. That wipe has no scroll compensation: every offscreen row collapses to the 120px default estimate, so the content above the viewport changes height wholesale while `scrollTop` stays fixed — the visible range now maps to different rows. The subsequent corrections (sync `resizeItem` seed pass, rAF repeat, per-row ResizeObserver) each shift content again, frame by frame.

The wipe dates to #115059 (session-switch measurement retention) and was refined by #117687 and #123713 without addressing anchoring. The repair deletes the wipe and keeps stale offscreen sizes as estimates: the existing synchronous `resizeItem` pass re-seeds connected rows *with* fold-based scroll compensation (rows entirely above the fold adjust `scrollTop` by their delta), so the anchor row holds still, and offscreen rows correct lazily as they connect — the standard dynamic-virtualizer model. End-pinned transcripts keep following the tail through the existing height-change `scrollToEnd` path.

## User Impact

Resizing the window, dragging the split, or docking the chat pane no longer loses the reading position. Mid-transcript readers stay on the same message; end-pinned sessions stay pinned.

## Evidence

- New regression e2e `ui/src/e2e/chat-resize-anchor.e2e.test.ts`: seeds 120 mixed-width messages, scrolls mid-transcript, resizes 1280→1000→820→1280. **Pre-fix**: the anchor row leaves the viewport entirely and `scrollTop` drifts 7021→7278. **Post-fix**: 0px drift at 1000, 42px at 820 (the anchor row itself spans the fold and re-wraps), byte-identical `scrollTop` back at 1280. Asserted bound: ≤60px.
[...]

PR #142088, opened 2026-09-08
test(infra): remove ephemeral port probe race

RomneyDa:
## What Problem This Solves

The ephemeral-port release test could fail under concurrent CI load with `Expected unavailable, received accepted` (`CI` run 34208988753, job 102005424124). After `tryListenOnPort` completes its `Server.close()` callback, the numeric port is deliberately unowned; a different listener can claim it before the test's follow-up connection.

## Why This Change Was Made

Replace the post-release numeric-port observation with an owner-boundary synchronization check. The test wraps the real server close callback, proves the helper promise remains unsettled while that callback is withheld, then releases it and verifies the allocated ephemeral port is returned. This directly protects the implementation contract without retries, sleeps, timeouts, or production changes.

No equivalent open or merged fix was found. #124013 introduced the follow-up connection after an earlier rebind assertion flaked; #126240 broadened refused-socket handling but retained the ownership race.

## User Impact

No runtime behavior changes. CI no longer attributes a replacement listener on a recycled ephemeral port to the already-closed probe server.

## Evidence

- Pre-fix deterministic ownership handoff: 20/20 replacement listeners produced `expected=unavailable received=accepted` after the original close callback.
- Pre-fix natural stress: 320 concurrent focused runs happened to pass, demonstrating the collision is scheduler/load dependent.
- Post-fix focused test: `node scripts/run-vitest.mjs src/infra/ports-probe.test.ts` (5 passed).
- Post-fix concurrent proof: 100/100 focused runs passed at concurrency 16.
- Sibling port suites: 114 passed, 1 platform-only skip.
- Targeted oxfmt, oxlint, and `git diff --check`: passed.
- `node scripts/check-changed.mjs`: all guards passed through core graph validation; local test typecheck could not use a borrowed cross-worktree compiler symlink. First-head CI exposed implicit mock callback types; exact Node types were added before this head.
- Autoreview on final candidate: scoped-clean, no P0 findings (0.99 confidence).
- LOC: production +0/-0; tests +46/-29 (net +17).

### DIFF
--- a/src/infra/ports-probe.test.ts
+++ b/src/infra/ports-probe.test.ts
@@ -1,6 +1,6 @@
 // Tests local port probing and availability detection.
 import net from "node:net";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { probePortUsage, tryListenOnPort } from "./ports-probe.js";
 
 async function withListeningServer(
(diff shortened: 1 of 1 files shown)


RomneyDa:
@clawsweeper review

PR #142092, opened 2026-09-08
fix(line): quote the message a reply answers

### LINKED ISSUE #142090: [Bug]: LINE replies never quote the message they answer, and channels.line.replyToMode is rejected
### Bug type

### Beta release blocker

No

### Summary

Reply threading is a shared OpenClaw setting: `channels.<channel>.replyToMode` decides
whether an outbound reply visibly references the message it answers, and the reply
pipeline resolves a reply target for every payload before it reaches a channel. LINE
does neither half. Setting `channels.line.replyToMode` is rejected by the config
schema, and the reply target the pipeline already resolved is dropped when the LINE
plugin builds its outbound messages.

In a group where several people are talking, the bot's answers arrive as unattached
messages. Nobody can tell which question a given answer belongs to. Telegram, Discord,
WhatsApp, Slack, Mattermost, and Google Chat each declare a threading adapter and read
`replyToMode`; LINE declares none, so it refuses the setting and drops the target.

The platform supports it. LINE's Messaging API attaches a `quoteToken` to every inbound
text, image, video, and sticker message, and accepts that token on the outbound
message types that carry text or a sticker.

### Steps to reproduce

1. Configure a LINE channel and add `replyToMode: "all"`:

   ```json5
   { channels: { line: { enabled: true, channelAccessToken: "…", channelSecret: "…", replyToMode: "all" } } }
   ```

2. Start the Gateway.

### Expected behavior

The Gateway starts, and each reply the bot sends quotes the message it is answering,
the way `replyToMode` behaves on the channels that support it.

### Actual behavior

The Gateway refuses to start:

```
Gateway failed to start: Invalid config at <state>/openclaw.json:
channels.line: invalid config: must not have additional properties: "replyToMode"
```

Remove the setting and the Gateway starts, but replies are never quoted. Captured from
the requests a running Gateway made to `https://api.line.me/v2/bot/message/*` while
answering an inbound LINE message (message text elided):

```
message/reply  status=400  [{"type": "text", "text": "A quote reply is …"}]
 message/push  status=200  [{"type": "text", "text": "A quote reply is …"}]
```

No `quoteToken` field, although the inbound webhook event that triggered this turn
carried one. (The 400 on `message/reply` is unrelated: this capture used a synthetic
reply token, and the push fallback is the designed degradation.)

### OpenClaw version

2026.9.2 (origin/main `76b61815623`)

### Operating system

Windows 11

### Install method

Source checkout

### Model

anthropic/claude-haiku-4-5

### Provider / routing chain

openrouter

Not relevant: the reply target is resolved before any model output reaches the channel.

### Logs

```
$ git grep -n -E "replyToId|replyToMode|threading" origin/main -- extensions/line/src
origin/main:extensions/line/src/channel.rich-messages.test.ts:43:        replyToId: "original-message",

$ git show origin/main:extensions/line/src/outbound.ts | grep -n "async ({"
44:  sendPayload: async ({ to, payload, accountId, cfg, onDeliveryResult }) => {
298:    sendMedia: async ({ cfg, to, text, mediaUrl, accountId }) =>
343:    text: async ({ cfg, to, text, accountId, onDeliveryResult }) => {
356:    media: async ({ cfg, to, text, mediaUrl, accountId, onDeliveryResult }) => {
```

One test fixture is the only mention in the whole plugin. Every outbound entry
point destructures the send context without `replyToId`, and `channel.ts` declares
no `threading` adapter, so `resolveReplyToMode` falls through to the generic config
reader, which looks for a key the LINE schema does not allow.

Platform capability, confirmed against the live Messaging API:

```
POST /v2/bot/message/validate/push  {"type":"text","quoteToken":"<invalid>"}
  -> 400 {"message":"Quote token is invalid"}
POST /v2/bot/message/validate/push  {"type":"flex","quoteToken":"…"}
  -> 400 {"message":"The message type 'flex' does not support quote message. (index: 0)"}
POST /v2/bot/message/push           {"type":"text","quoteToken":"<token from this chat>"}
  -> 200
```

### Impact and severity

Medium. Group chats are LINE's main multi-party surface, and this is where an
unattached answer is hardest to read. No data loss; the reply still arrives.

### Additional information

Related but different: #134217 covers the inbound direction — what the agent is told
about a message a person quoted. This issue is the outbound direction — whether the
bot's own reply quotes anything. #24004 asked for the inbound half and was closed as
stale. #120339 is the mirror case on Buzz — a channel that always threaded, with no
`replyToMode` to turn it off — and was fixed rather than documented.

edenfunf:
Closes #142090

## What Problem This Solves

Fixes an issue where people using OpenClaw in a LINE group could not tell which message a reply was answering. Replies arrive as unattached messages, and setting `channels.line.replyToMode` — the shared reply-threading setting other channels expose — stops the Gateway from starting at all, because the LINE config schema rejects the key.

## Why This Change Was Made

LINE addresses a quote by an opaque token it issues with each inbound message, not by message id, so the channel has to remember the token before it can honour the reply target the reply pipeline resolves. The plugin now records that token for each inbound message the agent is given, bounded per account and keyed by chat because LINE rejects a token used outside the chat that issued it, and every delivery path translates the resolved reply target back into it — including the two entry points that used to pick fields out of core's send context by hand and drop it: the message adapter, on the path core prefers for plain text, and `sendMedia`, which core routes a single-media reply through instead of the payload owner. `channels.line.replyToMode` (and the per-account override) selects the mode through the shared `scopedAccountReplyToMode` helper and defaults to `off`, matching every other channel that owns this slot, so no existing deployment changes behaviour until it opts in.

Deliberate boundaries: the token rides on the message object rather than on the request, because a LINE reply is split across a reply-token batch and push fallbacks that must not each carry their own quote; a reply quotes once, on the first message LINE accepts a quote on, so a reply made only of Flex cards or media goes out unquoted rather than being rejected wholesale; and the store stays in the running Gateway, so a reply that answers a message from before a restart is sent unquoted instead of guessing. `off` keeps the shared behaviour of honouring an explicit reply tag, as on Telegram, because a LINE quote is inline and hides nothing.

Two limits are enforced rather than documented away. `"batched"` is rejected by the LINE schema: it separates a reply to a coalesced turn from a reply to an immediate one, and nothing on the LINE path marks a turn as coalesced, so accepting it would leave an operator with a configured mode whose defining behaviour can never occur. When LINE rejects a request carrying a quote token with HTTP 400, the same reply is retried without that token; a 400 is a refusal, so that retry cannot duplicate an accepted message. Deleting or unsending the quoted message does not itself invalidate the token: LINE may instead show its content as unavailable, as documented in [LINE quote messages](https://developers.line.biz/en/docs/messaging-api/sending-messages/#send-quote-messages). Not done here: quoting the bot's own earlier messages, which has no consumer today.

## User Impact

In a busy LINE group, an operator can set `channels.line.replyToMode: "all"` and every answer visibly quotes the message it responds to, so participants can follow who the bot is talking to. Nothing changes for deployments that do not set it.

## Evidence

**Behavior addressed:** LINE replies carry no native quote, and `channels.line.replyToMode` is rejected by the config schema.  
**Real environment tested:** a real LINE Messaging API channel and a real Gateway on Windows 11, source checkout at the head below.  
**Exact steps or command run after this patch:** see the two blocks below.  
**Observed result after fix:** the reply the LINE API received carries the quote token of the message it answers; with the setting unset, it does not.  
**What was not tested:** the rendered quote bubble in a LINE client — this machine's LINE desktop app is not scriptable, so the platform's own acceptance of the token is used instead of a screenshot. `"first"` is covered at the resolver and filter level, not end to end.  
**Proof limitations or environment constraints:** the inbound webhook used in the Gateway run is a locally signed event carrying a quote token that the LINE API really issued for that chat; no second LINE account was available to send a genuine inbound message. `bot-message-context.test.ts` cannot report green on this machine — its teardown hits `EBUSY` on a Windows temp path, on `origin/main` as well — so its red/green is read from the failure message rather than the pass count; see Red/green below.

### Gateway A/B, real Messaging API

Two cold Gateway starts against the same state directory, differing only in `channels.line.replyToMode`, each answering one signed inbound webhook event. The requests the Gateway made to `https://api.line.me/v2/bot/message/*` were captured in process; the quote token is elided to its first 8 characters.

Unset (resolves to `off`):

```
message/reply  status=400  [{"type": "text", "text": "A quote reply is a message that references another message in the conversation by quoting or replying to it directly, showing context from the original message within your response."}]
 message/push  status=200  [{"type": "text", "text": "A quote reply is a message that references another message in the conversation by quoting or replying to it directly, showing context from the original message within your response."}]
```

`replyToMode: "all"`:

```
message/reply  status=400  [{"type": "text", "text": "A quote reply is a message that includes the text of a previous message, allowing you to respond to a specific point in the conversation with context visible.", "quoteToken": "Sc-yVFrq...<150 chars>"}]
 message/push  status=200  [{"type": "text", "text": "A quote reply is a message that includes the text of a previous message, allowing you to respond to a specific point in the conversation with context visible.", "quoteToken": "Sc-yVFrq...<150 chars>"}]
```

`Sc-yVFrq…` is the token LINE issued for the message the injected event quotes. The 400 on `message/reply` is present in both halves and is unrelated: the injected event carries a synthetic reply token, so both runs fall back to push — which is also why this capture shows the quote surviving that fallback.

### End-to-end against the live API, with controls

`probes/line-quote-reply-live.mts` drives the shipped modules — core's `resolveReplyToMode`, the LINE inbound context builder, and each LINE send entry point a reply can take — with the real Messaging API underneath:

```
1. core reply-threading resolves LINE through the plugin's threading adapter
  PASS core reads channels.line.replyToMode
    unset -> off; replyToMode:"all" -> all

2. anchor message in the operator DM (its quote token is a real one)
  anchor
    messageId=630865994952212777 quoteTokenLength=150 prefix=QT3qgSbC

3. the inbound context builder remembers the token LINE sent with the message
  PASS inbound builder records the quote token
    context built=true; stored token matches LINE's=true

4. the reply path attaches it, and LINE accepts the quoted reply
  PASS the reply LINE received carries the quote token
    status=200 sameTokenAsAnchor=true wire=[{"type":"text","text":"quote-reply probe: this reply should quote","quoteToken":"QT3qgSbC...<150 chars>"}]

5. controls
  PASS a reply that answers nothing carries no quote token
    status=200 wire=[{"type":"text","text":"quote-reply probe: this reply answers nothing"}]
  PASS an unknown message id is sent unquoted instead of guessing a token
    status=200 wire=[{"type":"text","text":"quote-reply probe: this reply answers a message this chat never saw"}]

6. the entry points core routes around the payload owner
  PASS the message adapter reply carries the quote token
    status=200 wire=[{"type":"text","text":"quote-reply probe: routed through the message adapter","quoteToken":"QT3qgSbC...<150 chars>"}]
  PASS the media reply carries the quote token on its caption
    status=200 wire=[{"type":"image","originalContentUrl":"https://raw.githubusercontent.com/.../1024.png","previewImageUrl":"https://raw.githubusercontent.com/.../1024.png"},{"type":"text","text":"quote-reply probe: media caption should quote","quoteToken":"QT3qgSbC...<150 chars>"}]

7. a quote token LINE refuses degrades to an unquoted reply
  PASS LINE refuses the quoted attempt and accepts the same reply unquoted
    400 [{"type":"text","text":"quote-reply probe: refused quote, the reply must still arrive","quoteToken":"not-a-re...<22 chars>"}] then 200 [{"type":"text","text":"quote-reply probe: refused quote, the reply must still arrive"}] -> delivered messageId=true
  PASS LINE reads the quoteToken field and rejects an invalid one
    POST /v2/bot/message/validate/push -> 400 {"message":"Quote token is invalid"}

9/9 checks passed
```

The last check is what makes the others mean something: LINE rejects the whole request for a token it does not accept, so the 200s above are LINE accepting the token rather than ignoring an unknown field. The media wire body also shows the split the platform forces: the image message carries no quote, the caption beside it does.

### Red/green

`bot-message-context.test.ts` cannot go fully green on this Windows machine: its per-test temp directory holds an open SQLite handle, so teardown fails with `EBUSY` after the assertions have already run. That is true on `origin/main` too — the unmodified file is `45 failed | 2 passed` here — so the red/green is read from the failure message rather than the pass count.

With the recording call in place, the two positive tests fail only in teardown:

```
 × remembers the token that quotes a text message it hands the agent
   -> EBUSY: resource busy or locked, unlink '...\openclaw-agent.sqlite-shm'
```

Reverting it moves the failure to the assertion:

```
 × remembers the token that quotes a text message it hands the agent
   -> expected undefined to be 'token-quotable' // Object.is equality
 × remembers the token that quotes a group message it hands the agent
   -> expected undefined to be 'token-quotable' // Object.is equality
```

The negative tests are pinned the other way: making `readLineQuoteToken` return a token for message kinds LINE does not attach one to turns one into

```
AssertionError: expected 'leaked-token' to be undefined
```

and the test for a message the agent never receives passes cleanly in both states, because it builds no session store and so never reaches the teardown that fails.

The other suites this change touches have no such problem and are green: `quote-tokens.test.ts`, `quote-reply-delivery.test.ts`, `outbound.send-media-quote.test.ts`, `channel.threading.test.ts`, `send.test.ts`, and `channel.sendPayload.test.ts`.

### Message adapter path

Core prefers `plugin.message.send.text` for a reply with no structured content, which is the path a plain Gateway-driven reply takes. Step 6 above drives that exported entry point against the real API. Reverting the context forwarding to the previous field-by-field pick makes the unit test for it fail with:

```
AssertionError: expected "vi.fn()" to be called once with arguments: [ 'line:group:Cadapter', …(2) ]
```

### Media path

A reply carrying one media url and no structured content does not reach `sendPayload` either: `deliver-core` takes it to `outbound.sendMedia`, which called `sendMessageLine` directly and dropped the reply target. Resolving the token there lets the send primitive put it on the caption, the one part of a media send LINE accepts a quote on. Removing that resolve turns the new test into:

```
AssertionError: expected { verbose: false, …(3) } to match object { …(2) }
```

The adapter also had to declare the `replyTo` durable-final capability: core answers `capability_mismatch` for any payload whose required capability an adapter does not declare, so without it a reply that names a target would have been pushed off durable final delivery the moment `replyToMode` was enabled.

### Config surface

`docs/.generated/config-baseline.counts.json` channel count goes 3710 -> 3712: exactly `channels.line.replyToMode` and its per-account override. `config:channels:gen` rewraps its output, so the line diff of `bundled-channel-config-metadata.generated.ts` is mostly reflow; the content delta is those same two keys and nothing else:

```
$ git show origin/main:src/config/bundled-channel-config-metadata.generated.ts | grep -o replyToMode | wc -l
31

$ grep -o replyToMode src/config/bundled-channel-config-metadata.generated.ts | wc -l
33
```

Both added occurrences read `"replyToMode":{"type":"string","enum":["off","first","all"]}` — the channel-level key and the per-account one.

### Checks

Windows 11, one run of each against this branch rebased on `76b61815623`:

```
tsgo -p tsconfig.extensions.json                     exit=0
tsgo -p test/tsconfig/tsconfig.extensions.test.json  exit=0
pnpm lint:extensions                                 exit=0
oxfmt --check (touched files)                        exit=0
pnpm check:assertion-safety --base origin/main       exit=0
pnpm check:max-lines-ratchet --base origin/main      exit=0
pnpm config:channels:check                           exit=0
pnpm config:docs:check                               exit=0
vitest threading.registry-backed-shard-a             exit=0   2 passed
vitest (extension-line)                              exit=1   60 failed | 825 passed | 1 skipped
```

The extension-line lane's 60 failures are the Windows temp-file problem described under Red/green, not assertions: 48 are `EBUSY` on teardown in `bot-message-context.test.ts` and 12 are the pre-existing 120s timeouts in `bot-mention-context.test.ts`, a file this change does not touch. The lane reports zero `AssertionError`, and every suite this change adds or edits passes.

`threading.registry-backed-shard-a` is the core contract shard LINE lands in; adding LINE to `threadingContractPluginIds` is what makes it assert `line threading contract` at all, so it is listed separately from the four standard checks.

### Maintainer integration

Source integration with pinned main `99a3f5152ae6f5858b561a9a4ef48682c5ace7fe` preserves the contributor root `63c7474fcb3ab92a1db76c06ded71b799605ada2`. The deleted-message example above and in the documentation was corrected against LINE’s official quote-message contract. Canonical metadata and config documentation were regenerated with trusted main tooling. All 20 changed paths match the maintainer reconstruction byte for byte. The main send path fails the refused-quote regression with HTTP 400; the integrated candidate passes 170 selected LINE tests plus the shared threading contract. Full changed-code checks, build (311.9 seconds), and LINE startup profiling pass for that integration. The earlier live proof above describes the contributor run.

After streaming PR #142145 landed, integrated main115ee1adf9b3402578a9e8cb723c9717cfb4210c to preserve both `replyToMode` and `streaming` in the LINE schema/types/docs and regenerated metadata. Quote runtime files are unchanged. Final head3d5307cf551ee00c74dd857b9b183094e32cf3a6 passes179 LINE tests (including all9 streaming cases),2 shared threading tests, both extension production/test typechecks, and fresh two-pass autoreview. All20 final changed paths match the tested reconstruction. Independent source/proof verification passes. Fresh hosted CI for this final integration remains the mechanical landing gate.


Patrick-Erichsen (approved):
Reviewed inbound quote ownership, account/chat scoping, the existing threading contract, and text/caption delivery entry points. A rejected quote token retries the same reply without the quote; unrelated failures remain failures. Trusted main fails that rejection regression with HTTP 400; the byte-matched integration passes 170 selected LINE tests plus the shared threading contract. Canonical generated metadata and config baselines are current. Fresh independent review of the final source tree is clean P0–P2. Code approved; changed-code checks/build/startup profile and new hosted CI remain in progress. Historical live provider proof and its synthetic-inbound/client-rendering limitations remain documented.

Patrick-Erichsen (approved):
Approved final quote-reply integration at 3d5307cf551ee00c74dd857b9b183094e32cf3a6. Retains already-landed streaming settings and regenerates shared config metadata.179 LINE tests plus2 shared threading tests, both extension typechecks, fresh autoreview and independent source/proof verification pass. Prior full build/profile evidence covers unchanged quote runtime. Fresh exact-head CI and native gates remain before landing.

Patrick-Erichsen:
@clawsweeper re-review

Please review final head 3d5307cf551ee00c74dd857b9b183094e32cf3a6. Final integration preserves already-landed streaming alongside replyToMode;179 LINE+2 shared tests, extension production/test types, fresh autoreview and independent source verification pass. Quote runtime matches the prior build/profile-proven integration. Body has current evidence.

Patrick-Erichsen:
Merged via squash.

- Prepared head SHA: [3d5307cf551ee00c74dd857b9b183094e32cf3a6](https://github.com/openclaw/openclaw/pull/142092/commits/3d5307cf551ee00c74dd857b9b183094e32cf3a6)
- Landed commit: [c71272163b5113f6308aead4df37b5b127feaeb8](https://github.com/openclaw/openclaw/commit/c71272163b5113f6308aead4df37b5b127feaeb8)

PR #142094, opened 2026-09-08
refactor(tests): share timestamped tool-result fixtures

steipete:
## What Problem This Solves

Agent, worker and provider tests repeat complete text-only tool-result input messages, adding boilerplate around the behavior each case protects.

## Why This Change Was Made

Share a pure test constructor at 49 input sites, removing 273 net test/support lines. All five values are required: call ID, tool name, text, error flag and timestamp. Each call creates a fresh message, content array and text block in the original property order, without defaults or production imports.

Existing sparse-message helpers, independent expected-result literals, casts, typed caller boundaries and shared input bindings remain intact. In particular, worker fixtures retain their ability to append image blocks, and replay fixtures retain their original append order and shared references.

## User Impact

No production behavior changes. The existing scenarios, assertions and deadlines remain intact.

## Evidence

- All 1,077 cases passed before and after in the same 32 complete suites, including consumers of the two changed support files. Per-file ordered case names and statuses match exactly, with no pending cases.
- Inline reconstruction matches the original syntax across all 29 consumers. Runtime checks verify literal values, ordered keys, ordinary prototypes and fresh nested allocations at all 49 sites.
- Independent P2 review and deslop are clean. Mapped changed gate passed.

No overall runtime improvement or coverage-percentage claim.

### DIFF
--- a/packages/ai/src/openai-completions-messages.test.ts
+++ b/packages/ai/src/openai-completions-messages.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it } from "vitest";
+import { makeTextToolResult } from "../../../test/helpers/text-tool-result.js";
 import { convertMessages } from "./openai-completions-messages.js";
 import type { ProviderContext, ProviderModel } from "./provider-types.js";
 import { resolveOpenAICompletionsCompat } from "./transports/openai-completions-compat.js";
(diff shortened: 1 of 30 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a22a47fb849e343b0d41d4ba184b670c391f8cdb](https://github.com/openclaw/openclaw/pull/142094/commits/a22a47fb849e343b0d41d4ba184b670c391f8cdb)
- Landed commit: [170a1151d0d90d7b3b3a31d26ab7d05bb06a8da4](https://github.com/openclaw/openclaw/commit/170a1151d0d90d7b3b3a31d26ab7d05bb06a8da4)

PR #142095, opened 2026-09-08
feat(cloud): choose an operating system for cloud sessions next to machine class

steipete:
## What Problem This Solves

Cloud sessions could pick a machine class ("size") per placement but not an operating system, even though Crabbox backends such as AWS and Azure already expose per-class targets (Linux, Windows normal/WSL2, macOS). Operators had no way to express or persist an OS choice in the Place picker, the Move session dialog, the cloud worker profile, or over RPC, and the Crabbox catalog silently discarded every non-Linux entry.

## Why This Change Was Made

Operating system becomes a second, provider-advertised choice that rides the same rails as machine class: `environments.list` profile summaries gain `operatingSystems` (id, label, default) and each machine option carries the `os` it belongs to; `sessions.dispatch` and the profile move target accept an optional `os` (profile targets only, like `machineClass`); the choice is carried through the placement destination, the durable provider-intent snapshot, and the move intent (new lazily ensured nullable `target_os` column beside `target_machine_class`, same schema version) into `WorkerProvider.provision`. Core stays provider-agnostic. The Crabbox plugin owns OS semantics: it parses every catalog target, adds `settings.target` as the profile default, maps the choice to `--target`/`--windows-mode`, includes the target in warm-image identity (Linux keys unchanged), and keeps desktop and warm images Linux-only.

The Control UI shows an **Operating system** section between the destination and machine sections whenever a profile advertises more than one OS, filters machine classes to the selected OS, extends the place chip (`aws · Windows (WSL2) · Fast`), and carries the OS through recovery and moves. The cloud worker settings editor offers only the operating systems the provider advertises.

Scope decision: the bundled Crabbox provider currently advertises **Linux only**. A live probe on Crabbox 0.51.0 showed Windows/WSL2 AWS leases warm up but `crabbox inspect` never reports `ready`, SSH lands in a Windows command shell, and `crabbox run` fails with an OpenSSH mismatch, so OpenClaw cannot enroll them yet; macOS enrollment is unverified and the enrollment replay reads `/proc`. Adding a target later is a one-line change in the plugin once Crabbox proves it. Windows normal mode needs a PowerShell enrollment path and is out of scope.

## User Impact

Operators can choose an operating system next to the machine size for new cloud sessions and moves, over RPC (`"os":"<id>"`) and in the Control UI, and set a per-profile default with `settings.target`. Existing profiles, moves, warm images, and clients are unaffected: absent `os` means the profile default, and Linux-only providers show no new UI. Generated Swift models are regenerated so the Mac app decodes the new move-target field.

## Evidence

- Live proof (isolated Gateway, Crabbox `aws` profile, tiny class): `environments.list` advertised OS options with per-OS machine catalogs; a Linux-default dispatch with no `os` reached `active` on lease `cbx_7ba160b5d29f` (`m7a.large`, kernel `7.0.0-1009-aws x86_64`), was reclaimed, and left zero task leases.
- Windows/WSL2 attempt: durable intent recorded `os: windows/wsl2`, AWS allocated `m8i.large` in WSL2 mode, but readiness never passed (Crabbox-side, see scope decision). Reclaimed cleanly.
- Focused suites: protocol schemas, gateway placement/dispatch/provider-intent/move-intent, state schema, Crabbox plugin, Control UI state/menus/chip/move dialog/settings; `pnpm protocol:check`, `pnpm tsgo:*`, `pnpm ui:check-performance`, `pnpm ui:i18n:verify`, `git diff --check`.
- Mocked Control UI captures (synthetic data) below.

### Before / after

| | |
| --- | --- |
| **Place picker before** Place picker before | **Place picker after** Place picker after |
| **Move dialog** Move dialog | **Settings editor** Settings editor |

## Notes for reviewers

- New config setting `cloudWorkers.profiles.<id>.settings.target` (Crabbox-owned; defaults to `linux`).
- New nullable `target_os TEXT` column on `worker_session_placement_moves`, first-use lazy like `target_machine_class`; documented in `docs/reference/database-schemas.md`.
- `WorkerMachineOptionsSchema.maxItems` 32 → 64 to hold per-OS catalogs; the plugin-sdk public export budget is unchanged.
- The extension lint lane fails only with the known nested-worktree "Declaration input escapes checkout" error locally; CI is authoritative.


steipete:
Recording the maintainer landing decision for the contract questions raised by ClawSweeper: this reviewed OS-selector implementation has explicit maintainer direction to land, including the additive provider contract, Crabbox-owned profile target, Linux-only rollout, and nullable, lazily ensured target_os move column at the existing schema version.

The compatibility scope is deliberate: older readers can reopen the database but do not implement the OS override. A rollback to an older build must not be understood as preserving an OS-specific move; it can use the profile default instead. The bundled Crabbox provider advertises and accepts Linux only in this rollout. Enabling another target requires enrollment proof first. The existing supportsProjectPreparation machine-class argument remains in its shipped second position, with OS added as an optional third argument.

The recorded storage proof covers first-use column addition, nullable shape, unchanged version metadata, and candidate reopening. It does not claim to run an older OpenClaw binary through an OS-specific move. This is the documented limit of the accepted design, rather than an unreported passing downgrade test.

The landing pass also fixed both initial CI failures (catalog/warm-image lint and the settings E2E merge-patch expectations). The follow-up unused-export failure is fixed in the latest commit; all three local dead-code export scans and extension typechecking pass. The cloud-worker browser suite passes all 8 tests, and the Crabbox suite passes 464 tests with 1 existing skip. The latest head still requires its own green CI before native preparation and merge.

steipete:
Merged via squash.

- Prepared head SHA: [b761451f9a93d3d0c3719105ef071f3f9e671d0b](https://github.com/openclaw/openclaw/pull/142095/commits/b761451f9a93d3d0c3719105ef071f3f9e671d0b)
- Landed commit: [8342d351b280e18936bfaacf85ab4b78f4e2476c](https://github.com/openclaw/openclaw/commit/8342d351b280e18936bfaacf85ab4b78f4e2476c)

PR #142103, opened 2026-09-08
fix(release): plugin previews fail with missing tooling dependencies

steipete:
## What Problem This Solves

Resolves a problem where plugin release previews fail before packaging because the independently checked-out trusted tooling has no dependencies, even though the selected release source was installed successfully.

## Why This Change Was Made

Initialize the frozen dependency graph in `.release-tooling` before each of its four npm and ClawHub preview/planning consumers, with lifecycle scripts disabled. The npm pack preview now checks out the full trusted workspace so its lockfile, workspace packages, and patches are available to that install. The existing checkout-local loader guard stays unchanged.

## User Impact

Release operators can complete plugin previews using the selected release source and trusted publisher tooling. This changes publisher preparation only; it does not rebuild core, alter the qualified release artifacts, or publish packages.

## Evidence

- Reproduced the missing-dependencies error from the actual trusted TSX loader before the independent install; the same loader succeeds afterward.
- Ran the real nonpublishing `--pack extensions/acpx` flow against unchanged release source `1391f7cd2d40ab5bbcf2f5f831d3a64f520e72d7`, with independently installed candidate/tooling checkouts and publication tokens removed. The package-local runtime build, lockfile check, and npm pack completed successfully on Node 24.20.0. Neither checkout built core or changed tracked candidate files.
- Four consumer regressions execute the install command and verify checkout ownership, ordering, frozen lockfile, and disabled lifecycle scripts. Focused workflow tests passed, as did the ClawHub and Git lifecycle sibling tests; the existing npm-command guard now distinguishes `npm install` from `pnpm install`.
- Workflow validation, scoped changed checks, affected test lint, and independent P2 review passed. Hosted exact-head CI remains required before landing.

### DIFF
--- a/.github/workflows/plugin-clawhub-new.yml
+++ b/.github/workflows/plugin-clawhub-new.yml
@@ -212,6 +212,13 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
           install-bun: "false"
 
+      - name: Install trusted plugin tooling dependencies
+        working-directory: .release-tooling
+        env:
+          CI: "true"
+        # Corepack reads this checkout's packageManager before pnpm processes arguments.
+        run: pnpm install --frozen-lockfile --prefer-offline --ignore-scripts
+
       - name: Validate publishable plugin metadata
         env:
           TSX_TSCONFIG_PATH: ${{ github.workspace }}/.release-tooling/tsconfig.json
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a15d0115c51044c2b86ee0b3abaf45aacfe084a6](https://github.com/openclaw/openclaw/pull/142103/commits/a15d0115c51044c2b86ee0b3abaf45aacfe084a6)
- Landed commit: [1a4d9a0340da6f2a79f6503b2ae488b043106bd7](https://github.com/openclaw/openclaw/commit/1a4d9a0340da6f2a79f6503b2ae488b043106bd7)

PR #142108, opened 2026-09-08
refactor(tests): remove unused database setup from recall timeout case

steipete:
## What Problem This Solves

The never-settling embedding test carries a complete mock database implementation even though both calls explicitly require that the database loader is never invoked.

## Why This Change Was Made

Replace that unused implementation with a fresh asynchronous spy. Embedding must finish before recall can search the database; this case instead times out and then returns the cooldown result. Both zero-load assertions remain, so an unexpected database load still fails the test.

## User Impact

No production behavior changes. Removes 11 lines of unused test setup while retaining every case, assertion, timeout, and cleanup action.

## Evidence

- The complete Memory-LanceDB index suite passes 189/189 before and after, with identical ordered case names and results and no skips.
- Every byte outside the single initializer replacement is unchanged.
- The mapped changed gate and fresh independent P2 review pass with no findings.

No runtime savings are claimed.

### DIFF
--- a/extensions/memory-lancedb/index.test.ts
+++ b/extensions/memory-lancedb/index.test.ts
@@ -1254,18 +1254,7 @@ describe("memory plugin e2e", () => {
     vi.useFakeTimers();
     const ensureGlobalUndiciEnvProxyDispatcher = vi.fn();
     const post = vi.fn(() => new Promise(() => {}));
-    const loadLanceDbModule = vi.fn(async () => ({
-      connect: vi.fn(async () => ({
-        tableNames: vi.fn(async () => ["memories"]),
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1e32db0d9a453dcf1f3cafe8144f3081b4818239](https://github.com/openclaw/openclaw/pull/142108/commits/1e32db0d9a453dcf1f3cafe8144f3081b4818239)
- Landed commit: [4dbb6b780759ac190b8795bb03d14260d7415fba](https://github.com/openclaw/openclaw/commit/4dbb6b780759ac190b8795bb03d14260d7415fba)

PR #142109, opened 2026-09-08
fix(cli): report JSON-mode command failures as command failures, not startup failures

steipete:
## What Problem This Solves

With `--json`, an ordinary command error is reported as a startup failure. Live on the built CLI at `53cac8a36e8` with two configured agents:

```
$ openclaw onboard recommendations --json
{ "ok": false, "error": { "type": "cli_error", "message": "Multiple agents are configured, but this operation has no explicit owner. …" } }
[openclaw] Could not start the CLI.
[openclaw] Reason: Multiple agents are configured, but this operation has no explicit owner. …
```

Without `--json` the same command prints only the message (exit 1). The CLI started fine; the operator is sent to `doctor` for a usage error.

## Why This Change Was Made

`runCommandWithRuntime` (`src/cli/cli-utils.ts`) deliberately rethrows in JSON mode so the entry writes the envelope centrally. The entry (`src/entry.ts`, `runMainOrRootHelp`) wraps both the `run-main` import and `runCli` in a single `onError` whose title is the startup one. The legacy entry in `src/index.ts` already distinguishes this case with "The CLI command failed.", and `failure-output.test.ts` pins that title's semantics. The fix records whether `runCli` began and picks the title accordingly; nothing else in the handler changes (envelope, debug hint, doctor hint, exit code).

Owner boundary: the entry handler owns phase classification; command-level JSON handling (`config`, `agents`, …) is untouched.

## User Impact

JSON-mode command errors now read "The CLI command failed." with the same reason line; failures before the command runs (a broken `run-main` import) still read "Could not start the CLI.". Exit codes and the JSON envelope are unchanged.

## Evidence

- Live before/after on the built CLI in an isolated state dir with two agents (`main`, `analyst`): `onboard recommendations --json` → startup banner before, command-failure banner after; the non-JSON form unchanged.
- `src/entry.run-main.test.ts`: two new cases — a command-phase throw must produce "The CLI command failed." and never "Could not start the CLI"; a `loadRunCli` rejection must produce "Could not start the CLI." and never the command title. The command-phase case fails on the unfixed `entry.ts`.
- `src/cli/failure-output.test.ts` and the entry tests pass; `check-changed` on the two files passes.

Production LOC: +4 (a flag, its set, the title ternary, one comment).

### DIFF
--- a/src/entry.run-main.test.ts
+++ b/src/entry.run-main.test.ts
@@ -16,6 +16,48 @@ describe("entry run-main boundary", () => {
     });
   });
 
+  it("frames a command-phase failure as a command failure, not a startup failure", async () => {
+    const previousExitCode = process.exitCode;
+    const errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    process.exitCode = undefined;
+    try {
+      await runMainOrRootHelp(["node", "openclaw", "onboard", "recommendations"], {
+        loadRunCli: async () => ({
+          runCli: vi.fn(async () => {
+            throw new Error(
+              "Multiple agents are configured, but this operation has no explicit owner.",
+            );
+          }),
+        }),
+      });
+      expect(process.exitCode).toBe(1);
+      expect(errorSpy).toHaveBeenCalledWith("[openclaw] The CLI command failed.");
+      expect(errorSpy).not.toHaveBeenCalledWith(expect.stringContaining("Could not start the CLI"));
+    } finally {
+      errorSpy.mockRestore();
+      process.exitCode = previousExitCode;
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [60ad7ed1458ed44bef19b62725a7d73aa2d35d88](https://github.com/openclaw/openclaw/pull/142109/commits/60ad7ed1458ed44bef19b62725a7d73aa2d35d88)
- Landed commit: [2f8ee468617d5633f69ba1937b2481cc64c7fdb7](https://github.com/openclaw/openclaw/commit/2f8ee468617d5633f69ba1937b2481cc64c7fdb7)

PR #142117, opened 2026-09-08
refactor: strengthen Feishu permission breaker coverage

steipete:
Related: #139428

## What Problem This Solves

The missing-scope breaker tests made ten lookup attempts at the same instant, but account-wide permission backoff admitted only one actual response. They could pass even if permission responses incorrectly counted toward the breaker threshold.

## Why This Change Was Made

Keep the rejected SDK error and resolved-envelope cases as two explicitly named table rows. Each now admits ten permission responses across expired backoff windows, verifies a distinct bot lookup stays suppressed within each window, and requires successful recovery afterward. The shared body removes duplication while strengthening the original contract.

## User Impact

No production behavior changes. All four original cases, names, order, module resets, and account lifetime remain. The change removes 11 test lines; no runtime improvement is claimed.

## Evidence

- Original suite: four passed. Adding failure accounting inside the permission branch while retaining backoff still passed all four old tests.
- Strengthened suite: four passed. The same mutation fails each missing-scope row separately at the expected recovery assertion.
- Production source restored byte-for-byte; final complete suite: four passed.
- Changed-file gate and fresh P2 review passed.

### DIFF
--- a/extensions/feishu/src/bot-name.test.ts
+++ b/extensions/feishu/src/bot-name.test.ts
@@ -46,18 +46,27 @@ describe("resolveFeishuBotName", () => {
     expect(requestMock).toHaveBeenCalledTimes(1);
   });
 
-  it("does not open the breaker for repeated missing-scope responses", async () => {
+  it.each([
+    ["does not open the breaker for repeated missing-scope responses", true],
+    ["does not open the breaker for resolved missing-scope envelopes", false],
+  ] as const)("%s", async (_name, rejects) => {
     vi.useFakeTimers();
     vi.setSystemTime(0);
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ff765bb32231d8649d510e257c6a90dc9d3a5d89](https://github.com/openclaw/openclaw/pull/142117/commits/ff765bb32231d8649d510e257c6a90dc9d3a5d89)
- Landed commit: [df1f3dbf38b1ed846e955c53624e6931f71a49d7](https://github.com/openclaw/openclaw/commit/df1f3dbf38b1ed846e955c53624e6931f71a49d7)

PR #142122, opened 2026-09-08
fix(telegram): isolate private-topic buffering and cancellation

### LINKED ISSUE #142107: [Bug]: Telegram DM topics share inbound buffers and cancellation
### Bug type

### Beta release blocker

No

### Summary

Messages in distinct private Telegram topics share one inbound debounce identity. Work waiting in one topic can block another topic, and a text `/stop` in one topic can discard buffered input from another.

### Steps to reproduce

The deterministic reproduction uses the real Telegram inbound-processing and buffering owners, with only downstream message processing held at its asynchronous boundary:

1. Use one private chat and sender, with `message_thread_id: 100` and `message_thread_id: 200`.
2. Start processing a message in topic 100 and hold its admission/completion promise.
3. Enqueue a second message in topic 100, then a message in topic 200.
4. Observe that topic 200 cannot reach downstream processing until topic 100 is released.

A second reproduction buffers input in both topics, then submits an authorized text `/stop` in topic 200. Input from topic 100 is discarded too. This reproduces in both normal and forwarded-message debounce lanes.

### Expected behavior

Distinct private topics have distinct inbound buffers. Waiting or stopping input in one topic should preserve input in another. This agrees with the existing topic-aware sequential lane and text-fragment/album collectors.

### Actual behavior

On the tested base, the blocked-topic case observed only `active A`, instead of `active A, topic B`. Both cancellation cases observed only `/stop`, instead of `/stop, keep topic A`.

The debounce-key producer calls `buildTelegramGroupPeerId`. Its conversation serializer intentionally handles forum and channel Direct Messages topics, while a private `scope: "dm"` falls back to the chat ID. Both enqueue and cancellation consume that identity.

### OpenClaw version

Source main `d3a596d97eca4b66a907d936adc5fc7a9b73a9b4` (package version 2026.9.2); the deployment that prompted the report runs 2026.9.1.

### Operating system

Ubuntu 24.04.4 LTS; Node 24.19.0.

### Install method

Source checkout with the pinned pnpm lockfile.

### Model

Not model-specific; the deterministic reproduction holds downstream message processing without calling a model.

### Provider / routing chain

Telegram inbound processing → keyed inbound debouncer → downstream message processing.

### Logs and evidence

Three focused regressions failed on the original producer and passed after retaining private topic identity. The same run passed 81 tests across inbound processing, key identity, provenance buffering, sequential lanes, and topic conversation parsing. The accompanying fix contains the executable regression tests.

### Impact and severity

Affected: private Telegram conversations with topic metadata. Consequences include cross-topic waiting and discarded pending input. Each of the three deterministic regressions reproduced once before the fix. The frequency across other deployments is unknown.

NianJiuZst:
Closes #142107

## What Problem This Solves

Fixes an issue where users messaging different private Telegram topics could have their input combined or blocked behind another topic, and an authorized `/stop` in one topic could discard another topic's buffered messages.

## Why This Change Was Made

Preserve the private topic ID in the Telegram inbound debounce key, which owns both enqueue and cancellation identity. Existing forum, channel Direct Messages, and unthreaded keys keep their behavior; public conversation IDs and persisted session routing are unchanged.

## User Impact

Private topics buffer and admit input independently. Stopping one topic preserves pending input in other topics, including forwarded messages.

## Evidence

- Reproduced four failures on main `8342d351b280e18936bfaacf85ab4b78f4e2476c` before changing production code: cross-topic coalescing, blocked admission in original message order, and cross-topic cancellation in default and forward lanes. The private-topic identity test also failed.
- After the repair, all four behavioral regressions pass using the real inbound-processing and buffer owners. The admission test holds downstream processing and verifies both independent topic progress and same-topic ordering after release.
- Focused validation: 81 passing tests across inbound processing, debounce identity, provenance buffering, sequential lanes, and topic conversation parsing. The five-suite run passed 80 tests; a final six-case identity run includes the added unthreaded-private control.
- `git diff --check` passes. Changed-scope checks passed production typechecking; the remaining local check run was stopped during test typechecking to publish at the requester's direction. Full check success is not claimed; GitHub CI is pending.
- Independent autoreview: scoped-clean, no actionable P0–P2 findings.
- Live Telegram proof is unavailable: the Test Server doctor could not access the QA broker through the Convex CLI. The local regressions are deterministic owner-boundary proof, not live Telegram evidence.
- Production delta: +5/-2 lines (net +3), retaining private-topic identity at its producer without broadening the shared conversation serializer. No configuration, schema, or dependency changes.

Thanks @aspalagin for the report and reproduction. The pre-existing ClawSweeper implementation run stopped before completion without publishing a PR.

Maintainer edits and access to this fork's workflow secrets are allowed.

### DIFF
--- a/extensions/telegram/src/bot-handlers.debounce-key.ts
+++ b/extensions/telegram/src/bot-handlers.debounce-key.ts
@@ -1,4 +1,3 @@
-// Telegram plugin module implements bot handlersebounce key behavior.
 import { buildTelegramGroupPeerId, type TelegramThreadSpec } from "./bot/helpers.js";
 export function buildTelegramInboundDebounceKey(params: {
   accountId?: string | null;
@@ -14,5 +13,9 @@ export function buildTelegramInboundDebounceConversationKey(params: {
   chatId: number | string;
   threadSpec: TelegramThreadSpec;
 }): string {
-  return buildTelegramGroupPeerId(params.chatId, params.threadSpec);
+  const { chatId, threadSpec } = params;
+  // Group peer IDs omit private topics; buffers and cancellation must retain them.
+  return threadSpec.scope === "dm" && threadSpec.id != null
+    ? `${chatId}:dm-topic:${threadSpec.id}`
+    : buildTelegramGroupPeerId(chatId, threadSpec);
 }
--- a/extensions/telegram/src/bot-handlers.inbound-pipeline.test.ts
+++ b/extensions/telegram/src/bot-handlers.inbound-pipeline.test.ts
@@ -28,44 +28,47 @@ describe("buildTelegramInboundDebounceKey", () => {
     ).toBe("telegram:default:12345:67890:forward");
   });
 
-  it("keeps scoped topic thread ids in the conversation key", () => {
-    const topic100 = buildTelegramInboundDebounceConversationKey({
-      chatId: 7,
-      threadSpec: { id: 100, scope: "forum" },
-    });
-    const topic200 = buildTelegramInboundDebounceConversationKey({
-      chatId: 7,
-      threadSpec: { id: 200, scope: "forum" },
-    });
-
-    expect(topic100).toBe("7:topic:100");
-    expect(topic200).toBe("7:topic:200");
-    expect(
-      buildTelegramInboundDebounceConversationKey({
+  it.each(["forum", "dm"] as const)(
+    "keeps %s topic thread ids in the conversation key",
+    (scope) => {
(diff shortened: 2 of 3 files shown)


VACInc:
Related follow-up context: #153829 is now merged, but it does not supersede this PR. #153829 bypasses buffering/FIFO only for authorized control commands. This PR repairs the separate private-DM-topic debounce identity used by ordinary buffered messages and cross-topic cancellation.

Keep this PR open.

PR #142126, opened 2026-09-08
fix(release): OpenShell preview fails on nested optional dependencies

steipete:
## What Problem This Solves

Resolves a problem where the OpenShell release preview fails with npm EOVERRIDE while assembling its portable dependency bundle. A dependency already installed beneath its parent was mistaken for missing and promoted into a conflicting direct install.

## Why This Change Was Made

Match dependency discovery to the shallow npm installation layout: resolve each dependency from its actual parent, stop at the real packaging root, and visit installed package directories rather than names. This preserves existing dependency overrides and still installs genuinely missing platform packages. Enclosing checkout dependencies and escaping symlinks cannot satisfy the packaged closure.

## User Impact

Plugin release previews can assemble complete portable bundles without changing dependency versions or hardening policy. The release source and qualified core artifacts remain unchanged.

## Evidence

- Reproduced the real OpenShell preview failure against frozen release source `1391f7cd2d40ab5bbcf2f5f831d3a64f520e72d7` and publisher tooling `1a4d9a0340da6f2a79f6503b2ae488b043106bd7`. The same nonpublishing pack succeeds with this owner repair.
- Inspected the resulting tarball: all seven fs-safe platform packages remain at 0.8.5, and jszip remains at 3.10.1.
- The new locked-optional regression fails on the original tooling with the same EOVERRIDE and passes after repair. An enclosing-checkout fixture confirms portable dependency inclusion without borrowing from an ancestor.
- All 51 package-manifest owner tests passed, including existing nested-version, patch, scoped override, source-preservation, and failure-cleanup cases. Scoped changed checks and independent P2 review passed. Exact-head hosted CI remains required before landing.

### DIFF
--- a/scripts/lib/plugin-npm-package-manifest.mts
+++ b/scripts/lib/plugin-npm-package-manifest.mts
@@ -346,32 +346,51 @@ export function generatePluginNpmPackageLockWithRetry(
   throw new Error(`package-lock generation retry loop exhausted for ${pluginDir}`);
 }
 
-function resolveInstalledPackageDir(packageDir: string, packageName: string) {
-  return path.join(packageDir, "node_modules", ...packageName.split("/"));
+function resolveInstalledPackageDir(
+  packageDir: string,
+  packageName: string,
+  fromDir = packageDir,
+): string | undefined {
+  const root = fs.realpathSync(packageDir);
+  let current = fs.realpathSync(fromDir);
+  while (true) {
+    const relative = path.relative(root, current);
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [916add3ac88f431e474747747684a5a9c235f124](https://github.com/openclaw/openclaw/pull/142126/commits/916add3ac88f431e474747747684a5a9c235f124)
- Landed commit: [01403169248346f2a6d6dd02955fc956fa9e1fe9](https://github.com/openclaw/openclaw/commit/01403169248346f2a6d6dd02955fc956fa9e1fe9)

PR #125715, opened 2026-08-18
fix(agents): report offline Trash failures in JSON

steipete:
## What Problem This Solves

Fixes an issue where operators running `openclaw agents delete <id> --force --json` without a reachable Gateway received success-shaped JSON when Trash cleanup failed. The agent was removed from config, but failed paths stayed on disk and the deletion journal remained unfinished without any machine-readable indication that manual cleanup or a retry was required.

## Why This Change Was Made

The offline deletion path now emits the same per-path `removed` and `failed` outcome shapes as Gateway deletion. The generic Trash helper returns structured outcomes from the shared cleanup owner, while the deletion journal remains incomplete until every required path reaches a terminal state so a later delete retries safely. The Gateway protocol and session-store purge contract are unchanged.

## User Impact

Scripts can now detect exactly which workspace, agent-state, or session paths failed to move to Trash and can retry deletion after fixing the host Trash condition. Human-readable deletion output remains unchanged.

Release note: Offline `agents delete --json` now reports per-path Trash failures instead of silently returning an incomplete cleanup result.

## Evidence

- Regression first failed on the pre-fix code because local JSON omitted `removed` and `failed`; the extended command-boundary test now verifies the failed workspace path and unfinished retry journal.
- Focused Vitest:
  - `src/commands/agents.delete.test.ts`: 22 passed
  - `src/commands/cleanup-utils.test.ts`: 25 passed
  - `src/commands/onboard-helpers.test.ts`: 51 passed
  - `src/claws/lifecycle-state.test.ts`: 29 passed
- `node scripts/check-changed.mjs -- <changed files>`: passed (`core`, `coreTests`, `docs`, `tooling`).
- `pnpm build`: passed locally after clearing exhausted build caches.
- Real local CLI proof with an isolated HOME/state and `.Trash` replaced by a regular file:
  - first delete exited 0 with `removed: []` and all three required paths in `failed`
  - failed paths remained on disk, config entry was removed, and journal `cleanup_completed` stayed `0`
  - after restoring Trash, the same command removed all three paths with `failed: []` and completed the journal
- The same clean-machine command passed on Blacksmith Testbox `tbx_01m09knqn2ckmqfp00bfb3s2wr`; the hosting workflow was canceled when the completed lease was stopped: https://github.com/openclaw/openclaw/actions/runs/32101010684
- Fresh branch autoreview: no accepted/actionable P0 or P1 findings.
[...]

PR #142130, opened 2026-09-08
fix(gateway): trash deleted-agent files from state dirs outside home and tmp

steipete:
## What Problem This Solves

`agents.delete` through the Gateway called `movePathToTrash` without allowed roots, so fs-safe fell back to its default roots (the home directory and the temp dir). Any installation whose `OPENCLAW_STATE_DIR` lives elsewhere, such as the documented Fly setup with `OPENCLAW_STATE_DIR=/data` or any volume-backed Docker deployment, saw every agent path refused with `Refusing to trash path outside allowed roots`. The agent left the config, but its workspace, agent dir and session directory stayed on disk, the deletion journal stayed incomplete, and the retained agent tree then tripped the shutdown session-store cleanup for every plugin (`plugin host cleanup failed ... OpenClaw agent database is unavailable while agent <id> is deleted`).

The offline CLI path had already moved to a parent-directory root policy in #125715; the Gateway sibling kept the fs-safe default.

## Why This Change Was Made

The Trash root policy for OpenClaw-owned paths now lives once with the cleanup owner in `src/commands/cleanup-utils.ts` (`trashAllowedRoots`): each declared path's own parent, plus the resolved parent when the moved path is a symlink (moving a link renames the link only). Both delete transports use it. The Gateway additionally keeps fs-safe's shipped home/tmp roots, so the directory behind a symlinked workspace keeps exactly the fence it had before (never widened to arbitrary external targets) while every path of a volume-backed state dir is now allowed. The helper deliberately stays off the fs-safe facade because the plugin SDK infra barrel re-exports that facade wholesale and the public SDK surface is budgeted. The Gateway feeds it the parent it already verified in `statAgentCleanupPath` and the recorded link target for symlink cleanup entries, so no second stat channel is introduced; the CLI path reuses its existing lstat and only resolves a link target when the path is a symlink. The private CLI-only copy is deleted.

## User Impact

Deleting an agent via the Control UI, `openclaw agents delete` with a reachable Gateway, or the `agents.delete` RPC now moves the agent's files to Trash on installations whose state dir is outside the home and temp directories, instead of reporting every path as failed and leaving the tree behind. Installations under `~/.openclaw` are unchanged.

## Evidence

- Live repro on a built `main` dist with an isolated `OPENCLAW_STATE_DIR` under a scratch directory: `agents delete writer --force --json` returned `failed` entries for the workspace, agent dir, session dir and every SQLite file, all with `Refusing to trash path outside allowed roots`; the gateway then logged `plugin host cleanup failed ... agent writer is deleted` for all 19 plugins on shutdown.
- New real-Gateway regression in `src/gateway/server.agent-database-recreation-product-proof.test.ts` (state dir outside `HOME` and `os.tmpdir()`): fails on `main` with the same refusal, passes with the fix (`failed: []`, workspace moved into `.Trash`).
- `src/gateway/server-methods/agents-mutate.test.ts` now asserts every Gateway Trash move carries the declared path's parent as an allowed root (the link's parent for symlink-derived targets); `src/commands/cleanup-utils.test.ts` keeps covering the CLI policy (symlink target parent, dangling links, symlinked parents).
- Local: `node scripts/run-vitest.mjs` over the product proof, `agents-mutate`, `cleanup-utils`, `agents.delete*`, `onboard-helpers`, `fs-safe`, `browser-maintenance` suites; `node scripts/check-changed.mjs` (core + coreTests lanes). Autoreview (Codex) scoped-clean on the final diff.

Production LOC delta: net +14 (policy helper exported from the cleanup owner, private copy removed, Gateway call carries declared-parent roots plus the shipped fence).

Accepted edge, named in the code: a symlink whose target sits beside the link in the same parent now has that target trashed by the Gateway too (the declared parent is a root). Transport divergence that predates this PR and stays out of scope: the Gateway also trashes the resolved directory behind a workspace symlink when it lies under home/tmp, the offline CLI only moves the link.

### DIFF
--- a/src/commands/cleanup-utils.ts
+++ b/src/commands/cleanup-utils.ts
@@ -77,7 +77,7 @@ export async function moveToTrashResult(
     const targetPath = path.resolve(pathname);
     const sourcePath = await resolveMoveToTrashSourcePath(targetPath);
     const allowedRoots = trashAllowedRoots(
-      sourcePath,
+      [sourcePath],
       isSymbolicLink ? await resolveSymlinkTargetPath(sourcePath) : undefined,
     );
     // Preparation can outlive its owner; revalidate immediately before Trash dispatch.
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [41b89706c61c0c8c2eb89fb66b1b7baf9f11ecfc](https://github.com/openclaw/openclaw/pull/142130/commits/41b89706c61c0c8c2eb89fb66b1b7baf9f11ecfc)
- Landed commit: [aefcc525baca6e2184df0c0a22799071deb3fc11](https://github.com/openclaw/openclaw/commit/aefcc525baca6e2184df0c0a22799071deb3fc11)

PR #142137, opened 2026-09-08
fix(qa): keep fresh write scenarios from inheriting completion

steipete:
Related: #142060.

## What Problem This Solves

Fixes an issue where a fresh mock QA recovery or exhaustion scenario could return final output before planning its required write when the conversation retained an earlier settled-tool continuation. An earlier exhaustion scenario could also determine a later recovery scenario's result.

## Why This Change Was Made

Use the existing scenario selector to identify one current recovery/exhaustion kind. Only a current continuation may reuse the earlier scenario prompt; recovery takes its marker from that selected prompt. This preserves the tool-free continuation behavior added by #142060 while preventing historical completion markers from satisfying a fresh kickoff.

## User Impact

Repeated QA scenarios plan their own initial write and retain their intended recovery or exhaustion behavior. This changes only the deterministic mock provider; it does not change the write tool or a real provider/channel.

## Evidence

Ordered requests first obtain an earlier write plan and settle that scenario, then start a fresh one. All four old/new recovery/exhaustion combinations failed at the fresh write-plan assertion before repair, while three controls passed. Two cases using the existing projected-input envelope separately reproduced the same omission. All nine pass after repair, preserving tool-free continuation, cron, heartbeat and finalization assertions.

There are 45 distinct passing focused cases including the related current-input, scenario-family and mock WebSocket coverage. All 25 canonical changed-check stages passed. Only a redundant test banner was deleted afterward; exact comment-only continuity and final formatting/diff checks are recorded. Independent managed review completed two passes with no actionable findings.

Synthetic requests inspect returned plans without executing the write tool, a Gateway, live QA/provider, native app or browser. The projected-input cases exercise the mock's accepted envelope; they do not certify an external runtime or arbitrary opaque transcript history. The two private scenario regexes become one kind matcher; net change is +2 mock implementation lines and +54 test lines.

### DIFF
--- a/extensions/qa-lab/src/providers/mock-openai/mock-openai-contracts.ts
+++ b/extensions/qa-lab/src/providers/mock-openai/mock-openai-contracts.ts
@@ -277,10 +277,8 @@ export const QA_THINKING_VISIBILITY_OFF_PROMPT_RE = /qa thinking visibility chec
 export const QA_THINKING_VISIBILITY_MAX_PROMPT_RE = /qa thinking visibility check max/i;
 export const QA_EMPTY_RESPONSE_RECOVERY_PROMPT_RE = /empty response continuation qa check/i;
 export const QA_EMPTY_RESPONSE_EXHAUSTION_PROMPT_RE = /empty response exhaustion qa check/i;
-export const QA_EMPTY_RESPONSE_SIDE_EFFECT_RECOVERY_PROMPT_RE =
-  /empty response after write recovery qa check/i;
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f942f91378bf5c0f8bfce54db941f8933ed80604](https://github.com/openclaw/openclaw/pull/142137/commits/f942f91378bf5c0f8bfce54db941f8933ed80604)
- Landed commit: [97e6f9ebfb612950d46681f748276b1822563a30](https://github.com/openclaw/openclaw/commit/97e6f9ebfb612950d46681f748276b1822563a30)

PR #142145, opened 2026-09-08
fix(line): accept the shared block streaming settings

### LINKED ISSUE #142144: [Bug]: channels.line.streaming is rejected, so LINE cannot use the shared block streaming controls
### Bug type

### Beta release blocker

No

### Summary

`agents.defaults.blockStreamingDefault` is off by default, and the docs tell an
operator to turn block replies on per channel with `*.streaming.block.enabled: true`.
LINE's config schema is strict and has no `streaming` key, so writing that setting
stops the Gateway from starting. The error suggests `openclaw doctor --fix`, and
running it restores the config from the last known good copy — silently discarding
the setting the operator just wrote instead of explaining that LINE cannot take it.

The result is that LINE has no way to turn block replies on or off for itself, and no
way to tune their coalescing, even though every block LINE sends is a separate message
counted against the channel's monthly quota.

### Steps to reproduce

1. Configure a LINE channel and add the setting the docs describe:

   ```json5
   { channels: { line: { enabled: true, channelAccessToken: "…", channelSecret: "…",
                         streaming: { block: { enabled: true } } } } }
   ```

2. Start the Gateway.
3. Run `openclaw doctor --fix` as the startup error suggests.
4. Re-read the config file.

### Expected behavior

The Gateway starts and LINE sends completed assistant blocks as they finish, the way
`*.streaming.block.enabled` behaves on the channels that accept it.

### Actual behavior

Step 2 fails:

```
[gateway] loading configuration…
[gateway] resolving authentication…
[gateway] starting...
Gateway failed to start: Invalid config at <state>/openclaw.json:
channels.line: invalid config: must not have additional properties: "streaming"
Run "openclaw doctor --fix" to repair, then retry.
```

Step 3 does not repair it — it rolls the whole file back:

```
Config auto-restored from last-known-good: <state>/openclaw.json (doctor-invalid-config);
Rejected validation details: channels.line: invalid config: must not have additional
properties: "streaming".
Updated config: <state>/openclaw.json
  Backup: <state>/openclaw.json.bak
```

After step 4, `channels.line` no longer has `streaming` at all. The operator's edit is
gone, and nothing said that LINE is the reason.

### OpenClaw version

2026.9.2 (origin/main `9911c71a573`)

### Operating system

Windows 11

### Install method

Source checkout

### Model

anthropic/claude-haiku-4-5

### Provider / routing chain

openrouter

Not relevant: the config is rejected before any model runs.

### Logs

```
$ git show origin/main:extensions/line/src/config-schema.ts | grep -n "streaming\|strict()"
21:  .strict();
57:}).strict();

$ git grep -l "streaming" origin/main -- 'extensions/*/src/config-schema.ts' \
    | sed 's#.*extensions/\([^/]*\)/.*#\1#' | sort -u | tr '\n' ' '
discord feishu imessage matrix msteams signal slack telegram
```

LINE is not in that list. `discord` and `telegram` use a preview-streaming schema,
which needs message editing; `signal` and `imessage` use the shared delivery-only
`ChannelDeliveryStreamingConfigSchema`, which is the shape LINE fits.

Asking the shipped LINE schema directly:

```
REJECT  channels.line.streaming.block.enabled            Unrecognized key: "streaming"
REJECT  channels.line.streaming.block.coalesce.minChars  Unrecognized key: "streaming"
REJECT  channels.line.streaming.chunkMode                Unrecognized key: "streaming"
REJECT  channels.line.accounts.x.streaming               Unrecognized key: "streaming"
```

The shared schema names LINE's case in its own doc comment
(`src/config/zod-schema.core.ts`):

```ts
/** Delivery-only nested streaming config for channels without preview modes. */
export const ChannelDeliveryStreamingConfigSchema = z.object({
  chunkMode: TextChunkModeSchema.optional(),
  block: ChannelStreamingBlockSchema.optional(),
}).strict();
```

LINE cannot edit a sent message, so it has no preview mode.

### Impact and severity

Medium. Nothing is lost from a conversation, but an operator following the
documentation is stopped at startup and then loses the edit to `doctor --fix`, with no
message naming LINE as the cause. Every LINE block is a separately billed message, so
this is also the control an operator most needs on this channel.

### Additional information

`docs/concepts/streaming.md` lists `*.streaming.block.enabled` as a channel override
that "forces block streaming per channel (and per account)", and
`docs/concepts/agent.md` says non-Telegram channels require it to enable block replies.
Related: #105940 was the same class of problem on Google Chat — `doctor --fix`
discarding a channel's block streaming config rather than preserving it.

edenfunf:
Closes #142144

## What Problem This Solves

Fixes an issue where an operator following the documentation could not turn block replies on or off for LINE. `channels.line.streaming.block.enabled: true` — the setting `docs/concepts/agent.md` and `docs/concepts/streaming.md` tell them to use — was rejected by LINE's strict config schema, so the Gateway refused to start; and `openclaw doctor --fix`, which the startup error recommends, restored the file from its last known good copy and discarded the edit.

## Why This Change Was Made

`ChannelDeliveryStreamingConfigSchema` is the shared shape for this, and its own doc comment says it is the "delivery-only nested streaming config for channels without preview modes". LINE cannot edit a message it has already sent, so it has no preview mode and is exactly that case; `signal` and `imessage` already take the same schema. Accepting it gives LINE `streaming.block.enabled`, `streaming.block.coalesce` and `streaming.chunkMode` at both channel and account scope in one step, with no new option invented for this channel.

Accepting the key alone would have been worse than rejecting it: the setting would be stored and never read, turning a loud startup failure into a silent no-op. So the channel's choice is also read into the turn, the way `imessage` and `feishu` do it — an explicit `true` sends `disableBlockStreaming: false` and an explicit `false` sends `true`, which is how core reads "this channel has decided", while an unset key sends nothing at all and leaves `agents.defaults.blockStreamingDefault` in charge.

The choice is resolved from the config admission produced for that event rather than the one the provider started with, because LINE re-reads its config per event; reading the startup snapshot would pin the streaming choice to whatever was configured when the provider was created.

Each scope is asked in turn rather than read off the merged account view. Account config is merged shallowly, so an account that sets nothing but `streaming.block.coalesce` replaces the channel's whole `streaming` object; reading only the merged view would let that account erase an explicit channel-level enable or disable. Core already resolves coalescing this way — channel scope and account scope separately, with the account overriding only the fields it names — so the enable flag now inherits by the same rule.

Not done here: a LINE-specific `blockStreamingCoalesceDefaults`. Picking a number on the operator's behalf is a product decision, and `streaming.block.coalesce` now lets them pick their own.

## User Impact

An operator can enable block replies for LINE alone, or keep LINE on whole replies while other channels stream, and tune the coalescing in between. Because every block LINE sends is a separate message counted against the channel's monthly quota, this is the control that matters most on this channel. Nothing changes for a deployment that does not set the key: it stays on `agents.defaults.blockStreamingDefault`, which is off by default.

## Evidence

**Behavior addressed:** `channels.line.streaming` was rejected by the LINE config schema, so the Gateway would not start and `doctor --fix` discarded the setting.  
**Real environment tested:** a real LINE Messaging API channel and a real Gateway on Windows 11, source checkout, four cold starts.  
**Exact steps or command run after this patch:** see the two blocks below.  
**Observed result after fix:** the Gateway starts, and the channel-scoped setting turns block replies on and off for LINE independently of the agent default.  
**What was not tested:** `streaming.chunkMode` end to end — it is carried by the same shared schema and read by core's chunker, and is covered here only at the schema level.  
**Proof limitations or environment constraints:** the inbound webhook in the Gateway runs is a locally signed event, so each reply shows a rejected `message/reply` followed by the `message/push` that carries it; no second LINE account was available to send a genuine inbound message.

### Before: the Gateway will not start, and doctor discards the setting

```
$ node dist/index.js gateway run
[gateway] loading configuration…
[gateway] resolving authentication…
[gateway] starting...
Gateway failed to start: Invalid config at C:\tmp\line-quote\state\openclaw.json:
channels.line: invalid config: must not have additional properties: "streaming"
Run "openclaw doctor --fix" to repair, then retry.

$ node dist/index.js doctor --fix
Config auto-restored from last-known-good: C:\tmp\line-quote\state\openclaw.json
  (doctor-invalid-config); Rejected validation details:
  channels.line: invalid config: must not have additional properties: "streaming".
Updated config: C:\tmp\line-quote\state\openclaw.json
  Backup: C:\tmp\line-quote\state\openclaw.json.bak
```

`channels.line` has no `streaming` key afterwards.

### After: the setting is accepted and decides the turn

Four cold Gateway starts, each answering the same signed inbound event. The requests the Gateway made to `https://api.line.me/v2/bot/message/*` were captured in process; the reply is the same three-paragraph answer every time.

| `agents.defaults.blockStreamingDefault` | `channels.line.streaming.block.enabled` | text messages LINE received |
| --- | --- | --- |
| `off` | unset | 1 (1070 chars) |
| `off` | `true` | 3 (374 / 370 / 324 chars) |
| `on` | unset | 3 (374 / 370 / 324 chars) |
| `on` | `false` | 1 (1070 chars) |

Row 2 is the channel override turning block replies on while the agent default is off; row 4 is it turning them off while the agent default is on. Rows 1 and 3 show the unset key leaving the agent default alone.

The wire bodies for row 2:

```
message/reply  400  [{"type":"text","text":"Webhook A webhook is a way for one application to send real-time notif…"}]
 message/push  200  [{"type":"text","text":"Webhook A webhook is a way for one application to send real-time notif…"}]
 message/push  200  [{"type":"text","text":"A message queue is a system that temporarily stores messages sent from…"}]
 message/push  200  [{"type":"text","text":"A rate limit is a restriction that caps how many requests or actions a…"}]
```

and for row 1, the same answer as one message:

```
message/reply  400  [{"type":"text","text":"Webhook A webhook is a way for one application to send real-time notif…"}]
 message/push  200  [{"type":"text","text":"Webhook A webhook is a way for one application to send real-time notif…"}]
```

### Unit coverage and mutation

Nine tests pin the translation. Each of these mutations reddens a different set, so no test is passing by accident:

| Mutation | Tests that fail |
| --- | --- |
| read the startup config instead of the turn's own | 5, including `forgets a choice the operator removed since the provider started` |
| pass `enabled` through without inverting it | 4, not including that one |
| let an unset key answer "disabled" | exactly the two `stays silent` tests |
| drop the account layer | exactly `lets a named account override the channel-wide choice` |
| read the merged account view only, without the channel fallback | exactly the two partial-override tests, with `expected undefined to be true` / `to be false` |

### Config surface

`docs/.generated/config-baseline.counts.json` channel count goes 3710 -> 3726: the eight keys the shared schema brings, at channel and account scope.

```
channels.line.streaming
channels.line.streaming.block
channels.line.streaming.block.coalesce
channels.line.streaming.block.coalesce.idleMs
channels.line.streaming.block.coalesce.maxChars
channels.line.streaming.block.coalesce.minChars
channels.line.streaming.block.enabled
channels.line.streaming.chunkMode
```

plus the same eight under `channels.line.accounts.*`.

### Checks

Windows 11, one run of each at this branch's head:

```
tsgo -p tsconfig.extensions.json                     exit=0
tsgo -p tsconfig.extensions.test.json                exit=0
pnpm lint:extensions                                 exit=0
oxfmt --check (touched files)                        exit=0
pnpm check:assertion-safety --base origin/main       exit=0
pnpm check:max-lines-ratchet --base origin/main      exit=0
pnpm config:channels:check                           exit=0
pnpm config:docs:check                               exit=0
vitest (extension-line)                              exit=1   57 failed | 787 passed | 1 skipped
```

The 57 failures are this Windows machine, not assertions: 45 are `EBUSY` on teardown in `bot-message-context.test.ts` and 12 are 120s timeouts in `bot-mention-context.test.ts`. Neither file is touched by this change, and the unmodified `bot-message-context.test.ts` reports `45 failed | 2 passed` on `origin/main` here as well. The lane reports zero `AssertionError`, and `monitor.streaming-config.test.ts` passes 9/9.


Patrick-Erichsen (approved):
Reviewed the integrated LINE streaming fix at `d14ffc32bcc68dd7228852003eea493ebfe6cf62`. Before/after proof on a trusted-main reconstruction of the reviewed source: six of nine streaming cases failed with the main monitor; all nine passed with this change. Tested source bytes match the prepared merge. Coverage includes explicit enable/disable, account inheritance, partial account settings, and live turn-config selection. Shared config metadata was regenerated with the canonical tools. Independent P0–P2 source review was clean.

Code approved for landing; fresh exact-head CI and the native final checks are still pending. The earlier live LINE evidence retains its stated synthetic-inbound and chunkMode limits.

Patrick-Erichsen:
Merged via squash.

- Prepared head SHA: [d14ffc32bcc68dd7228852003eea493ebfe6cf62](https://github.com/openclaw/openclaw/pull/142145/commits/d14ffc32bcc68dd7228852003eea493ebfe6cf62)
- Landed commit: [115ee1adf9b3402578a9e8cb723c9717cfb4210c](https://github.com/openclaw/openclaw/commit/115ee1adf9b3402578a9e8cb723c9717cfb4210c)

PR #142166, opened 2026-09-08
fix(crabbox): maintain warm images across mixed CLI binaries

steipete:
## Problem

`maintain()` in the Crabbox worker provider resolved one CLI binary per cloud-worker profile and, when more than one distinct executable resulted, warned `Crabbox warm-image maintenance requires one configured CLI executable` and returned without touching any retained warm image. On a team Gateway (2026-09-08) the `aws`/`machine0` profiles used a wrapper binary while the new `gcp`/`azure` profiles used `/usr/local/bin/crabbox`, so the warning fired every minute and idle-image cleanup and retained deletion retries silently stopped for every profile. Mixed binaries are valid configuration.

## Why the old code refused

Warm-image records are keyed by provider/target/setup/class/project, not by binary, so a retained record cannot be attributed to an executable. The Crabbox CLI (`internal/cli/checkpoint.go`, `checkpoint delete`) prints `checkpoint absent id=<id>` and exits 0 for an id outside its catalog. Running maintenance through the wrong executable would therefore have cleared the OpenClaw record while the snapshot survived in the other catalog, a leaked, billed image.

## Solution

Maintenance now receives every distinct resolved binary (de-duplicated, sorted for deterministic order). For each retained record, deletion is attempted through the executables in that order:

- non-zero exit or command error: the obligation is retained and warned once, exactly as before, and no further executable is consulted (an error is not evidence of absence);
- exit 0 with an exact `checkpoint absent id=<id>` line and another executable remaining: try the next one;
- exit 0 otherwise (real deletion, or absence reported by the last executable): the record is cleared as before.

A single-binary configuration behaves byte-for-byte as it did (one command; exit 0 clears; failure retains). All attempts share the existing one-minute maintenance budget; if it runs out between executables the obligation is retained silently, matching the existing deadline break. Cancellation (`dispose`, authority `assertCurrent`, abort signal) is checked between attempts.

The checkpoint command runner and the cross-catalog delete loop moved into `crabbox-worker-warm-image-checkpoint.ts` to keep the manager under the max-lines budget; call sites are unchanged.

## Docs

`docs/gateway/cloud-workers.md` no longer describes the single-executable requirement and documents the ordered-attempt rule.

## Tests

`extensions/crabbox/src/crabbox-worker-warm-image-maintenance.test.ts`: the old refusal test is replaced by behavior tests for two profiles with different binaries (records known to only one catalog, and to none, are all released in sorted executable order), first-executable errors retaining the obligation without consulting the second catalog, shared-deadline budgeting, exact-id absent matching, and the unchanged single-executable path. The mixed-catalog and error tests fail on the previous code (maintenance returned before issuing any deletion).

## Proof

- `pnpm test extensions/crabbox/src`: 14 files, 456 passed, 1 pre-existing skip.
- Focused: `crabbox-worker-warm-image-maintenance.test.ts` + `crabbox-worker-provider.test.ts`: 265 passed. Before the production change, the new maintenance tests failed for the intended reason (3 failed, 5 passed).
- `pnpm tsgo:extensions`, `pnpm tsgo:extensions:test`, `oxlint` and `oxfmt --check` on the changed files, `git diff --check`: clean. The manager file stays under the 700-line max-lines budget after the extraction (677 effective lines).
- `pnpm check:changed`: every lane passed up to `lint extension changed files`, which fails in this nested worktree with the known environmental `Declaration input escapes checkout` error (qrcode manifest resolved from the ancestor install); CI is authoritative for that lane.
- Independent Codex autoreview (P0–P2) on the final tree: scoped-clean.

Not covered: no live Gateway with two configured binaries was exercised; the CLI's `checkpoint absent`/`checkpoint deleted` lines were verified against the Crabbox source rather than a live run.


steipete:
Merged via squash.

- Prepared head SHA: [5524fe9a32b704a67b2f148beffc4fec5d31ed12](https://github.com/openclaw/openclaw/pull/142166/commits/5524fe9a32b704a67b2f148beffc4fec5d31ed12)
- Landed commit: [c95f48ed35e51f0e1b8f0937179aecdd890e5f3a](https://github.com/openclaw/openclaw/commit/c95f48ed35e51f0e1b8f0937179aecdd890e5f3a)

PR #126816, opened 2026-08-20
fix(updater): force-tolerant tag fetch so recreated upstream tags don't abort updates

ruel225:
Closes #123318

## What Problem This Solves

Fixes an issue where Git-checkout users could not update a stored stable or beta channel after upstream recreated an existing release tag. Git rejected the old tag with `would clobber existing tag`, and every update stopped at `fetch-failed`.

## Why This Change Was Made

Refresh branches with `--all --prune --no-tags --no-prune-tags`, without adding force to their configured refspecs. Then refresh tags using `+refs/tags/*:refs/tags/*` from one release remote, with pruning disabled. This recovers moved release tags without globally forcing branch updates or deleting local-only tags.

The release source is the declared remote retained in `branch.main.remote`, then `origin`, then the sole configured remote. Ambiguous non-origin remotes require an explicit tracking choice; their ordering cannot choose the release authority. Both the private inspection flow and the direct updater use the same fetch owner.

## User Impact

Stored stable/beta Git updates recover from recreated tags. A normal auxiliary fork remote cannot overwrite the upstream's same-named tag. Local-only tags survive even with `fetch.prune` and `fetch.pruneTags` enabled. Protected non-fast-forward branch mappings still fail before the tag refresh. The CLI's private inspection flow preserves installed refs until admission.

## Maintainer update

Built on @ruel225's existing three commits and integrated current main's private inspection flow without rewriting that ancestry. The branch already contained a single-remote selector; this update carries it into both current fetch paths, removes its arbitrary first-remote fallback, and replaces replayed Git-command fixtures with tests through `updateGitCheckout`.

Thanks to @zhangguiping-xydt (#128345) for the release-remote scoping idea; the maintainer commit includes their verified commit-author credit. The chosen design keeps the small two-fetch shape. It does not adopt the alternate PR's refspec classifier, negative refspecs, `--refmap`, or Git-version logic.

## Evidence

- Real Git fixture: bare upstream, installed clone with populated tags, and a fork with a colliding tag. The original `git fetch --all --prune --tags` exits 1 with `would clobber existing tag`.
- The same production-boundary regression fails against current main in both direct and private-inspection paths with `fetch-failed` and that rejection. With this patch, the updater resolves the corrected tag and returns the fixture's expected `already-current` outcome.
[...]

PR #142167, opened 2026-09-08
fix(tooling): drop machine squash credit and carry human co-author trailers

steipete:
## What Problem This Solves

Fixes an issue where landing a PR through `scripts/pr merge-run` refused GitHub's squash preview whenever any commit carried an agent attribution trailer (Claude, Codex, Cursor, Copilot, Amp, Trae), so maintainers had to hand-write a `--body-file` for every such PR. Landings that bypassed the wrapper, such as #126816 (665d4fadbd4), shipped `Co-authored-by: Claude <noreply@anthropic.com>` in the squash message, once in the trailer block and three more times inside the replayed commit bullets, against the repository rule to preserve human credit and never add agent attribution.

## Why This Change Was Made

The composer now drops machine credit from the default preview wherever GitHub replayed it (terminal trailer block, per-commit bullets, trailers folded onto indented continuation lines, indented key lines) and collapses the blank line it leaves behind. It still keeps preview credit backed by a non-merge commit author and appends every human `Co-authored-by` trailer found in the PR's commits, merge commits included. Reviewed `--body-file` bodies stay byte-preserved and are rejected when machine credit appears anywhere in them. Merge-queue PRs cannot take an override, so the composer refuses queue admission when the preview needs any edit or omits a required human trailer.

Machine identities are exact verified addresses plus GitHub App `[bot]` no-reply accounts, which cannot belong to a person. Copilot (three verified account ids), `claude[bot]`, `cursor[bot]`, the Codex connector bot, and `noreply@openai.com` join the existing list. Names and provider domains are never matched: people commit from `@openai.com` and `@anthropic.com` too. Commit authorship attribution is unchanged.

## User Impact

Maintainers can land PRs whose commits carry agent trailers without writing a body by hand, and the resulting squash message credits only humans. An operator who supplies a reviewed body gets a clear rejection when it still carries agent credit instead of a silently accepted message.

## Evidence

- Dry run of `scripts/pr-lib/merge-body.mjs compose` against the real four commits of #126816, using the actual merge body as the preview input. Before: the composer refused with `Squash message contains machine co-author credit; use --body-file`. After: all four Claude lines are removed and the trailer block is `ruel225` plus `zhang-guiping`.
- `pnpm test test/scripts/pr-merge.test.ts`: 103 passed (previously 78), driving the real `merge.sh` `prepare_squash_merge_body` path with fixture repositories, including a #126816-shaped fixture, folded, indented, and greedy-unfolding cases, explicit-body rejection, and merge-queue refusal.
- `pnpm check:changed` passes; `git diff --check` is clean.
- Four Codex autoreview rounds at P2 threshold: six findings fixed with regression tests; one rejected (a folded unsupported human credit inside GitHub's own terminal block, which GitHub writes one line per co-author, keeps the existing fail-closed error).

Known gap: agent accounts without a `[bot]` suffix, such as Copilot's user-type ids, are only caught by the explicit address list.

### DIFF
--- a/docs/ci/watching-runs.md
+++ b/docs/ci/watching-runs.md
@@ -102,7 +102,14 @@ regular UTF-8 file before verification. Empty files are valid. It preserves
 operator-provided text and trailers, appending any missing co-authors from the
 current GitHub preview and reviewed source commits. This option requires squash
 and a non-queue PR; all review, CI, exact-head, and admission checks still apply.
-Without the option, the existing GitHub preview behavior is unchanged.
+Without the option, the wrapper composes the message from the GitHub preview:
+it keeps credit backed by a non-merge PR commit author or a reviewed source
+trailer, appends human `Co-authored-by` trailers from the PR's commits that the
+preview omitted, and drops machine credit (Claude, Codex, Cursor, Copilot, Amp,
+Trae, and GitHub App `[bot]` accounts) wherever GitHub replayed it, including
+inside per-commit bullets. A reviewed body that still contains machine credit
+is rejected, and a merge-queue PR whose preview needs such edits stops before
+admission.
 
 `merge-recover` accepts the same option after its required outcome ID and
 `--confirmed-operator-recovery`. Repeating `merge-run` with a retained outcome
--- a/scripts/pr-lib/merge-body.mjs
+++ b/scripts/pr-lib/merge-body.mjs
@@ -55,45 +55,116 @@ function trailers(body) {
   return parsed.stdout.split("\n").filter(Boolean);
 }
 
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2573f30ed5c8dda466e61bc3259f013bcae7b541](https://github.com/openclaw/openclaw/pull/142167/commits/2573f30ed5c8dda466e61bc3259f013bcae7b541)
- Landed commit: [be6a59250caf2c01e7b901089206d74e16bdfbd3](https://github.com/openclaw/openclaw/commit/be6a59250caf2c01e7b901089206d74e16bdfbd3)

Issue #142149, opened 2026-09-08
[Bug]: rawError=terminated can bypass settled-turn recovery and surface as LLM timeout when timedOut=false


wpu911:
﻿## Summary

On OpenClaw `2026.9.2`, an OpenAI-compatible local model stream can end with `errorMessage="terminated"` even though the agent run itself did **not** time out. The UI then surfaces this as `LLM request timed out`, and a settled tool-driven turn can fail before producing its final visible reply.

In the reproduced run, all side-effecting tool work had already completed. The final assistant event contained internal reasoning only, no visible text and no new tool call, then ended with `stopReason="error"` / `errorMessage="terminated"`.

The run-level timeout flags were all false:

```text
timedOut=false
idleTimedOut=false
timedOutDuringToolExecution=false
timedOutByRunBudget=false
```

Yet the UI showed:

```text
The agent run failed before producing a reply.
LLM request timed out.
```

## Environment

- OpenClaw: `2026.9.2`
- OS: Windows 11
- Provider API: `openai-completions`
- Provider: local OpenAI-compatible endpoint through llama-swap / llama.cpp
- Model: local 35B MoE model
- Agent timeout: 7200 seconds
- Provider timeout: 7200 seconds

## Reproduction pattern

This was easiest to reproduce on a long tool-driven turn:

1. Agent starts a background download job.
2. The agent performs multiple tool calls while monitoring the job.
3. All requested downloads complete successfully.
4. The final assistant event contains reasoning such as `All 6 episodes downloaded` but no visible final answer.
5. The provider stream terminates with:

```json
{
  "role": "assistant",
  "stopReason": "error",
  "errorMessage": "terminated"
}
```

6. OpenClaw ends the run with `stopReason=error` and presents it as an LLM timeout.

The upstream local inference server remained healthy and returned HTTP 200 responses around the failure, so this was not a model/server crash or a 7200-second run timeout.

## Why this appears to happen

`message-patterns` currently classifies exact `terminated` as a timeout-class error:

```js
/^terminated$/i
```

That classification is useful for failover and is related to the earlier behavior discussed in #56875, but it also means a transport-level `terminated` can be surfaced as a timeout even when the actual run timeout flags are false.

There is an additional recovery gap for a settled turn. `resolveSettledTurnFinalizationRequest()` currently exits early for terminal assistant errors:

```js
const terminalAssistant = resolveCurrentAttemptAssistant(input.attempt);
[...]

PR #142168, opened 2026-09-08
fix(agents): recover final replies after terminated streams

LiuwqGit:
AI-assisted: yes

Related: #142149

## What Problem This Solves

Fixes an issue where a completed tool-driven turn loses its final reply when the provider stream ends with bare `terminated` after the ordinary retry window has elapsed. The turn reports `LLM request timed out` even though its run timeout did not fire.

## Why This Change Was Made

Reuse the existing anchored transport-message pattern when capturing settled-turn context. Preserve the caller's lifecycle callback and terminal deferral through the built-in finalizer, so the recovered answer ends the run successfully instead of retaining the prior attempt's error.

The normal transcript-continuation path, timeout-family classification, exact tool-settlement checks, cancellation, refusal and permanent-error guards remain unchanged. The finalizer still has no callable tools.

## User Impact

Eligible completed turns retain their final summary without repeating tool effects. Asynchronous or unfinished work, previously visible answers and unsettled delivery remain excluded.

## Evidence

The real authenticated Gateway proof uses `chat.send` and `agent.wait`, the built-in file tool, an actual interrupted HTTP response body and the built-in finalizer. An initial transient failure starts the existing retry window; the later post-tool stream closes after that window expires. The fixture does not disable retries or change timeout policy.

- Baseline `82d4e9a`: one completed write, persisted `terminated`, no final answer, and the existing timeout-family error.
- The ordinary single-cut continuation control still answers with one write.
- All three new context/transport regressions failed at the intended boundary before the fix; 834 focused checks passed after the context repair.
- The connected lifecycle regression failed because the callback was dropped; 35 lifecycle, cancellation and harness checks passed after forwarding the existing caller fields.

Candidate `96d2cfe4` passes the real late-cut Gateway flow: the existing finalizer sends one request with no callable tools, `agent.wait` returns success, history contains one final answer, the original message prefix is unchanged, and the only added message is the final answer. Normal cache bookkeeping remains separate from messages. The completed write is not repeated.

Fresh source-blind native acceptance independently passed both the late-cut recovery and ordinary-continuation controls on `96d2cfe4`. The full negative-state and source-mechanism gates are covered separately by focused tests and source review. Exact-head hosted CI and the final review gate remain required before landing.

Limits: the provider is a deterministic loopback service. This does not certify a live Windows llama.cpp/llama-swap deployment or make every background-download run eligible. Private internal flags/context are supported by source and observed lifecycle evidence, not claimed as directly exported fields.

### DIFF
--- a/src/agents/embedded-agent-runner/run/attempt-result.ts
+++ b/src/agents/embedded-agent-runner/run/attempt-result.ts
@@ -10,7 +10,10 @@ import {
 import { projectAgentRunAttemptTerminal } from "../../agent-run-terminal-outcome.js";
 import { isCloudCodeAssistFormatError } from "../../embedded-agent-helpers.js";
 import type { subscribeEmbeddedAgentSession } from "../../embedded-agent-subscribe.js";
-import { INCOMPLETE_ASSISTANT_STREAM_RE } from "../../failover/message-patterns.js";
+import {
+  INCOMPLETE_ASSISTANT_STREAM_RE,
+  TERMINATED_TRANSPORT_MESSAGE_RE,
+} from "../../failover/message-patterns.js";
 import type { AgentRuntimeModelAttempt } from "../../runtime-plan/types.js";
 import { markCoreTtsAttemptResult } from "../../tools/tts-tool-result-provenance.js";
 import { log } from "../logger.js";
@@ -117,7 +120,15 @@ function isTransientSettledTurnFailure(failure: unknown): boolean {
     typeof failure.message === "string"
       ? failure.message.trim()
       : "";
-  return INCOMPLETE_ASSISTANT_STREAM_RE.test(message);
+  // An exact `terminated` body-stream failure carries no retryable code, so the
+  // network classifier and the incomplete-stream wording both miss it. The
+  // failover timeout family already treats this transport failure as
+  // recoverable (#56875); the settled-turn context producer must agree so a
+  // completed tool batch keeps its isolated tool-free finalization instead of
+  // failing the whole run (#142149).
+  return (
+    INCOMPLETE_ASSISTANT_STREAM_RE.test(message) || TERMINATED_TRANSPORT_MESSAGE_RE.test(message)
+  );
 }
 
 function normalizeEmbeddedAttemptToolMetas(
--- a/src/agents/embedded-agent-runner/run/attempt.settled-turn-finalization-context.test.ts
+++ b/src/agents/embedded-agent-runner/run/attempt.settled-turn-finalization-context.test.ts
@@ -72,6 +72,19 @@ describe("settled post-tool turn finalization context", () => {
       error: new Error("Stream ended without finish_reason"),
       captures: true,
     },
+    {
+      // Undici ends a stream body with this exact bare transport message (#142149).
+      kind: "terminated-transport-body",
+      error: new Error("terminated"),
+      captures: true,
+    },
+    {
+      // Only the exact transport message is recoverable; variants that merely
+      // contain the word keep failing closed like any other provider failure.
+      kind: "terminated-word-variant",
+      error: new Error("the request was terminated by the server"),
+      captures: false,
+    },
   ])("$kind final provider failure captures context=$captures", async ({ error, captures }) => {
     const result = await runSettledTurnWithFailedFinalCall(
       "agent:main:telegram:direct:settled",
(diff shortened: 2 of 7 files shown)


LiuwqGit:
Added the missing real-behavior proof for the exact bare `terminated` path (follow-up to the review's proof ask): a new end-to-end case `finalizes after terminated failure` in `settled-turn-finalization.transport.test.ts`, pushed in 4512562d3e2e359f375a8f1ead56b66706804f9b.

**What it proves (real transport, no fixtures for the failure itself):**
- A loopback HTTP server speaks the real OpenAI Responses SSE protocol; on the final post-tool stream it writes a valid `response.output_item.added` event and then destroys the socket mid-body — Node/undici surfaces exactly `TypeError: terminated`, which reaches the transcript as `stopReason="error"` / `errorMessage="terminated"` with no retryable code (`isTransientNetworkError` returns `false`, confirming this is the previously-unclassified path — the failure is produced by the real transport, not injected).
- The production projector captures `settledTurnFinalizationContext` and the isolated tool-free finalizer issues the recovery request with `tools: []`, producing the visible final reply `"Note saved once."`.
- Completed side effects execute exactly once: `execute` called once, tool lifecycle `{startedCount: 1, completedCount: 1, activeCount: 0}`, note file contains a single `saved` line, provider request tool counts are `[1, 1, 0]`, and the failed-turn transcript prefix is preserved with the finalization answer as the only addition.

**Evidence after fix:**
```text
✓ finalizes after HTTP failure 1329ms
✓ finalizes after WebSocket failure 215ms
✓ finalizes after terminated failure 438ms
Test Files  1 passed (1)   Tests  3 passed (3)
```

**Evidence before fix** (same test with only the fix files reverted to main — the already-recognized HTTP/WebSocket cases still pass, the bare `terminated` case fails exactly at the classification gap):
```text
✓ finalizes after HTTP failure 1253ms
✓ finalizes after WebSocket failure 230ms
FAIL finalizes after terminated failure
AssertionError: expected { terminal: { kind: 'ok' }, …(48) } to match object
  { settledTurnFinalizationContext: { source: 'openclaw-transcript' } }
Test Files  1 failed (1)   Tests  1 failed | 2 passed (3)
```

The PR body Evidence section has been updated accordingly. Note on the earlier red checks at d5df610: `checks-node-compact-small-51`/`-52` were runner-infra failures (exactly 10-minute runs with zero recorded steps and no retrievable logs), and `checks-node-compact-large-17` failed in `ui/src/app/settings.preferences.node.test.ts` with a Vitest `EnvironmentTeardownError` — none touch this PR's `src/agents/**` files; the new push re-runs them.

obviyus:
landpr-approve-unrelated-red
pr: 142168
head: 96d2cfe4a62abb5cedb2427da009fb7644d65e05
run: 34347762454 attempt: 1
failed: checks-node-changed-extensions-bundle-27 job=102453618170 conclusion=failure
check: openclaw/ci-gate

obviyus:
Thanks @LiuwqGit! Landed as 56145576b232b367cb6063c51f70471681fb5119 with your authorship preserved.

**Proof:** The real Gateway baseline lost the final reply after a completed write and an interrupted HTTP response once the ordinary retry window had elapsed. The repaired run uses the existing tool-free finalizer, stores one answer, preserves the original messages, and completes successfully without repeating the write. Independent native acceptance confirmed both that recovery and the unchanged ordinary-continuation path. Focused checks and the exact-head security-aware review passed.

**CI disposition:** [Run 34347762454](https://github.com/openclaw/openclaw/actions/runs/34347762454) completed with 139 successful jobs, 17 skipped jobs, and one failed child plus its aggregate. [The failed child](https://github.com/openclaw/openclaw/actions/runs/34347762454/job/102453618170) reports a Telegram loopback model-confirmation `editMessageText` network error. That callback path uses grammY's HTTP client and does not execute this PR's embedded-attempt or finalizer changes. Main-side controls at `82d4e9a` passed in both standalone and canonical runner modes; they did not reproduce the failure. Its underlying fetch cause and introducing commit remain unknown. The failed shard's remaining selected tests are not claimed as passed. Landing used the [guarded completed-run exception](https://github.com/openclaw/openclaw/pull/142168#issuecomment-5601721253), not a green-CI claim or an unchanged rerun.

**Scope:** Existing asynchronous-work, unfinished-tool, cancellation, refusal, delivery and timeout safeguards remain. The proof uses a loopback provider; it does not certify the original Windows/background-download scenario. #142149 remains open for that broader reproduction.

PR #142129, opened 2026-09-08
fix(plugins): clarify Gateway restart guidance after installs

WOLIKIMCHENG:
Related: #141801

## What Problem This Solves

Plugin installation always told operators to restart manually, even when a Gateway-hosted install had already requested an automatic restart. The setup agent could print that instruction twice. The output also omitted that configuration reload can restart connected channels before the new plugin source is active.

## Solution

Keep one explanation in the shared install success path. Distinguish installs performed by the running Gateway from separate-shell installs and reload-off mode. Describe automatic restart as a request, preserve next-start source activation, and explain possible earlier channel reload. Remove the duplicate setup-agent instruction and retain the artifact-specific Control UI follow-up.

Real proof showed that managed-service presence and hybrid mode alone do not guarantee an automatic restart for a separate-shell install. The wording preserves that manual case. Install trust, persistence, restart scheduling, channel behavior, and approval policy are unchanged.

## Evidence

- All 71 focused tests passed, plus changed-file formatting and lint.
- Baseline real CLI and setup-agent output reproduced the misleading and duplicate instructions.
- Separate-shell, reload-off, and offline installs completed through the real CLI with the corresponding guidance.
- Gateway-hosted artifact installs accepted a restart plan, deferred until active work completed, restarted through the supervisor, and loaded the plugin after restart. The repaired result contains one shared explanation and the Control UI follow-up.
- Real trust, compatibility, archive-integrity, source-consent, and install-policy refusals reported failure without a success/restart promise. Injected config-publication failure rolled back the installed payload; removing the fault allowed recovery.
- Fresh independent behavior validation passed all 14 observable clauses with varied fixtures, stored-state checks, real restart observation, and negative controls. Source-only obligations were reviewed separately. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34228376739) and final review are enforced at landing.

This is an output correction. It does not claim to fix or reproduce the historical Telegram interruption.

PR #142023, opened 2026-09-08
fix(telegram): honor image defaults and preserve sticker descriptions

obviyus:
## What Problem This Solves

Static sticker descriptions could ignore the configured image default, and a new description could disappear before the first reply because finalization retained older canonical text.

## Why This Change Was Made

Sticker handling uses the shared image selector with agent scope, then updates canonical reply text and its existing alias. Duplicate catalog, credential, and provider policy code is removed. Raw and command text, cache reuse, native vision, and general-media fallback retain their contracts.

## User Impact

Sticker descriptions use the configured image selection and reach the first reply. Failed descriptions are not cached or retried through another sticker-description model; general-image handling remains available.

Consumers: sticker image selection and reply finalization now share the canonical description.

## Evidence

Selection and first-reply regressions failed before their fixes. Seven selection and five caller controls passed afterward. Compiled receiver → download → model → reply checks showed the fresh description in the first request, preserved quoted text, and stored the successful sticker. Missing credentials and a rejected request made no second sticker-description attempt. Formatting, lint, structural checks, and required continuous integration passed. Endpoints were synthetic; live messaging, inference, cache restart, and new native-vision behavior were not tested.

PR #142169, opened 2026-09-08
fix(gateway): config edits fail after successful agent mutations

ceckert:
## What Problem This Solves

Fixes an issue where clients creating, updating, or deleting an agent could immediately read an old configuration revision and then have their next `config.patch` rejected with `config changed since last load; re-run config.get and retry`, despite no competing writer.

## Why This Change Was Made

Invalidate the configuration read cache when the managed reloader observes persisted writes. This shared boundary covers agent mutations, configuration RPCs, and file-watcher observations. Remove the configuration-RPC-only exception while keeping committed notifications and applied-runtime tracking separate.

## User Impact

A fresh `config.get` exposes the persisted revision before runtime reload finishes. Clients can use that revision for their next conditional edit. Genuine stale edits still fail, and `config.patch` still waits for its exact runtime application before acknowledging success.

## Evidence

A real source Gateway and authenticated WebSocket client exercised both pinned main `d559bcda4f9d45ee0c610c8e109e80ffdbb8fcd5` and candidate `c3ce9d1dec75eeafba4dc4cae0e3439a09b26efe` on one isolated Testbox. A documented local SecretRef resolver held normal runtime preparation after persistence. No Gateway implementation was replaced, and no external provider, channel, service, or production credential was used.

| Operation | Baseline fresh read while preparation is held | Candidate fresh read while preparation is held |
| --- | --- | --- |
| `agents.create` | Old persisted revision; follow-up patch rejected | New persisted revision; follow-up patch succeeds after release |
| `agents.update` | Old persisted revision; follow-up patch rejected | New persisted revision; follow-up patch succeeds after release |
| `agents.delete` | Old persisted revision; follow-up patch rejected | New persisted revision; follow-up patch succeeds after release |
| External watched file edit | Old persisted revision | New persisted revision |
| Direct `config.set` | New persisted revision | New persisted revision |

Observed candidate RPC sequence, with opaque revisions renamed for privacy:

```text
config.get                         -> hash R0, applied A0
agents.create                      -> success; disk now R1
runtime preparation                -> held
config.get                         -> hash R1, resolved A1, applied A0
config.patch(baseHash: R0)          -> INVALID_REQUEST: config changed since last load
config.patch(baseHash: R1)          -> pending while preparation is held
release runtime preparation
config.patch                       -> success
config.changed                     -> final committed revision R2, once
config.get                         -> persisted and applied revisions converge
```

Update and delete produced the same fresh-read and guarded-write results. The public raw-revision tokens matched separately measured file bytes using the existing installation projection; the key was never exported. No `config.changed` event arrived during the hold. A superseded candidate emitted no committed event; only its replacement did.

Invalid configuration and an injected strict preparation failure did not become applied runtime revisions or emit committed events. Restoring valid preparation and making a new observed edit recovered normally. A genuine competing write still rejected the older token. Sensitive values and secret references stayed redacted in RPC output.

All 805 focused Gateway configuration, cache, reload, and lifecycle tests passed, including the 53-test configuration-RPC file. Formatting and changed-file lint passed. These checks also cover redaction, included-file preservation, conditional guards, and explicit replacement paths.

A fresh source-blind run independently confirmed the core persisted-read, stale-write, watcher, event, supersession, and failure-recovery outcomes. Its broader source-only and unexercised negative controls remain explicitly separated from live passes. Full private RPC, file-hash, resolver, event, and Gateway traces are retained. AI-assisted.

### DIFF
--- a/src/gateway/config-get-response.ts
+++ b/src/gateway/config-get-response.ts
@@ -52,8 +52,8 @@ export async function readConfigGetResponse(params: {
   }
   const appliedConfigHash = getRuntimeConfigAppliedHash();
   const pluginRegistryVersion = getActivePluginRegistryVersion();
-  // With an active watcher, cache hits never re-read the file. External edits
-  // become visible after its successful commit; the write path invalidates early.
+  // With an active watcher, cache hits never re-read the file. Candidate
+  // observation invalidates persisted bytes before runtime acceptance.
   if (
     configGetResponseCache?.getHotReloadStatus === getHotReloadStatus &&
     configGetResponseCache.revisionProjector === params.revisionProjector &&
@@ -87,7 +87,7 @@ export async function readConfigGetResponse(params: {
   }
 }
 
-/** Invalidates cached config.get work after the watcher accepts a config candidate. */
+/** Invalidates cached config.get work when the watcher observes or accepts a candidate. */
 export function invalidateConfigGetResponseCache(): void {
   configGetResponseCache = undefined;
 }
--- a/src/gateway/server-methods/config-write-flow.ts
+++ b/src/gateway/server-methods/config-write-flow.ts
@@ -24,7 +24,6 @@ import { captureGatewayRootWorkAdmissionContinuationScope } from "../../process/
 import { getActiveSecretsRuntimeSnapshotState } from "../../secrets/runtime-state.js";
 import { isRecord } from "../../utils.js";
 import { resolveGatewayAuth } from "../auth.js";
-import { invalidateConfigGetResponseCache } from "../config-get-response.js";
 import { buildGatewayReloadPlan, isNoopGatewayReloadPlan } from "../config-reload-plan.js";
 import { resolveGatewayReloadSettings } from "../config-reload-settings.js";
 import { formatControlPlaneActor, type ControlPlaneActor } from "../control-plane-audit.js";
@@ -268,9 +267,6 @@ export async function commitGatewayConfigWrite(params: {
     ),
     afterWrite: { mode: "auto" },
   });
-  // Watcher acceptance is debounced; clear now so the writer's immediate
-  // follow-up config.get observes the committed bytes before that hook runs.
-  invalidateConfigGetResponseCache();
   return {
     path: resolveGatewayConfigPath(params.snapshot),
     config: result.nextConfig,
--- a/src/gateway/server-reload-managed.ts
+++ b/src/gateway/server-reload-managed.ts
@@ -375,7 +375,11 @@ export function startManagedGatewayConfigReloader(
     readSnapshot: params.readSnapshot,
     promoteSnapshot: async (snapshot, _reason) => await params.promoteSnapshot(snapshot),
     subscribeToWrites: params.subscribeToWrites,
-    onConfigCandidateObserved: pauseGatewayRestartForConfigCandidate,
+    onConfigCandidateObserved: () => {
+      // Every writer must expose persisted revisions before runtime acceptance.
+      invalidateConfigGetResponseCache();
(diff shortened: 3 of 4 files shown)


obviyus:
landpr-approve-unrelated-red
pr: 142169
head: c3ce9d1dec75eeafba4dc4cae0e3439a09b26efe
run: 34230098743 attempt: 2
failed: checks-node-changed-extensions-bundle-29 job=102114765632 conclusion=failure
check: openclaw/ci-gate

obviyus:
Thanks @ceckert!

**Proof:** A real CLI-started Gateway and authenticated WebSocket client now return the persisted configuration revision after agent create, update, and delete while runtime preparation is still held. Old revisions are rejected. A patch using the fresh revision succeeds after normal application.

- **Bug:** Fresh reads returned the cached old revision after agent mutations, causing false concurrent-edit errors.
- **Fix:** The shared candidate-observation owner invalidates persisted reads for every writer; committed events and applied-runtime revisions remain separate.

| Before | After |
| --- | --- |
| Create/update/delete and external edits returned an old revision until acceptance | Fresh reads match persisted bytes before acceptance |
| A follow-up patch using that read failed | The fresh-hash patch succeeds after runtime preparation completes |

The baseline and candidate used the same isolated host and supported local SecretRef preparation fixture. Invalid, superseded, and failed candidates did not become applied revisions or emit early committed events. Fresh source-blind validation confirmed the central workflow; broader source-only and unexercised negative controls were kept separate. All 805 focused tests passed, including 53 configuration-RPC cases. Final exact-head review found no correctness or security issue.

The guarded merge uses a [new exact-run unrelated-CI approval](https://github.com/openclaw/openclaw/pull/142169#issuecomment-5587702110) for [run 34230098743, attempt 2](https://github.com/openclaw/openclaw/actions/runs/34230098743). Its sole failed child, [checks-node-changed-extensions-bundle-29 / job 102114765632](https://github.com/openclaw/openclaw/actions/runs/34230098743/job/102114765632), timed out in `sticker-cache.selection.test.ts`, “keeps minimax on VLM despite other available vision models.” The other failure was its dependent CI aggregate.

The exact tested merge was `2a661af124a95d1d33fad343773430a5cbb7633f`, based on main `c21fa540660255762d81053dcf06ee0f8385b357`. The same executed two-file prefix reproduced the same timeout both with the PR and on that exact main parent without it. Native execution coverage positively observed sticker/model selection and did not execute the changed Gateway cache/reloader functions. Those functions require a real Gateway write/reloader lifecycle, which this image-selection fixture does not create. The selector, provider registration, setup, dependencies, and cache-generation policy are unchanged by this PR.

Independent recurrence is also retained in [PR #142129 run 34228376739 / job 102069062439](https://github.com/openclaw/openclaw/actions/runs/34228376739/job/102069062439). Its tested merge `4f921a82e4c01b79774b09c4ea92b38aa4c55244` has all four of this PR's paths unchanged from our tested main parent, plus identical relevant selection/loader/setup/cache boundaries. That is a different-PR recurrence, **not** a matching main-push CI result.

The environments differ: current CI used Node 24.20.0 and restored caches; native controls used Node 24.19.0; the earlier recurrence used two CPUs and two workers. Exact elapsed-time attribution remains incomplete. The failure does not justify a timeout increase or unrelated source change here. [Commit 335e872](https://github.com/openclaw/openclaw/commit/335e87235d1426ec4ad862bdbebcb35829b0a02f), introduced by #142023, added this test/shared-selector call; that is source provenance, not blame for the timing failure. The loader-performance investigation remains separate.

The original preflight runner-loss failure and the single infrastructure recovery are preserved. No second hosted retry or false passing-CI claim is made. AI-assisted.

PR #142170, opened 2026-09-08
fix(gateway): recover stuck one-way restart drain after a failed SIGUSR1 handler

jai-assistant:
## What Problem This Solves

A gateway process can permanently refuse all newly submitted work while still looking healthy.

When the `SIGUSR1` restart handler in `runGatewayLoop` rejects *after* `markRestartDraining()` has already closed process-wide admission, no in-process restart arrives, so `resetAllLanes()` / `resetGatewayWorkAdmission()` never runs. The handler's recovery path called `rollbackGatewayRestartSignalAdmission()`, which only clears the **reversible** signal fence — `clearRestartSignalFence()` returns early and is documented as a "no-op while one-way restart drain owns admission".

The result: the surviving process keeps `restartDraining = true` for its entire lifetime.

Because already-admitted root work still completes, the failure is very hard to diagnose:

- `gateway call health` reports healthy, and interactive turns keep answering.
- Every *newly submitted* item — subagents, plugin relays, queued commands — fails with `Gateway is draining; new tasks are not accepted`.
- Only `admission closed: restart drain` appears in the log, with no matching reopen and no restart afterwards.
- Restarting the gateway is the sole workaround, since the flag is a `globalThis` singleton with no persistence.

Operationally this surfaced as a Google Meet realtime agent whose node-relay consults were refused for hours while the same agent answered correctly when invoked directly (`openclaw agent --agent <id>`), because that path was already-admitted root work.

The local mirror `restartDrainingMarked` **is** reset in that handler, and its comment states the intent ("a later signal must repeat the synchronous close transition even if this handler failed after marking the one-way drain") — but the global admission state has no matching recovery, so intent and behaviour diverge.

## What This Changes

- Adds `recoverGatewayRestartDrainAfterFailedSignal()` to `src/process/gateway-work-admission.ts`: an explicit escape hatch that clears a stuck one-way drain, replaces the aborted drain controller so recovered admissions observe a fresh signal (never a pre-aborted one), and falls back to the reversible-fence clear when no one-way drain is set.
- Wires it through `recoverGatewayRestartDrainAfterFailedSignalAdmission()` (`src/infra/restart.ts`) and the lifecycle runtime re-export, and calls it from the failed-`SIGUSR1` recovery path in `src/cli/gateway-cli/run-loop.ts`.
- Keeps the one-way guarantee intact for committed restarts: they continue to reset state through the existing `resetGatewayWorkAdmission()` boundary. Only failed-restart recovery may call the new function, which is stated in its doc comment.

## Evidence

Targeted tests added to `src/process/gateway-work-admission.test.ts`:

- `recovers stuck one-way drain when a restart handler failed after marking it` — asserts the old path cannot recover (`rollbackGatewayRestartSignalFence()` returns `false` while both `isGatewayWorkAdmissionClosed()` and `isGatewaySubordinateWorkAdmissionClosed()` stay `true`), then that recovery reopens admission, exposes a non-aborted drain signal, and admits new root work.
- `recovers a failed restart drain that superseded a suspension` — covers a drain that invalidated a live suspension.
- `reopens a stuck restart drain for a live suspension owner` — after recovery, normal suspension control still closes and releases admission.
- `leaves a committed restart drain one-way for the reset boundary` — regression guard that a committed restart stays closed until `resetGatewayWorkAdmission()`.

Local validation on Node v26.8.1 / pnpm 12.3.4:

```
pnpm vitest run src/process/gateway-work-admission.test.ts \
  src/infra/restart-suspension.test.ts \
  src/infra/restart.deferral-timeout.test.ts
  -> Test Files 3 passed (3) | Tests 57 passed (57)

pnpm vitest run src/infra/gateway-suspend-coordinator.test.ts src/process/command-queue.test.ts
  -> included in the 87-passing run across restart/queue lanes

pnpm run tsgo:prod        -> clean
pnpm run tsgo:test:root   -> clean
pnpm exec oxfmt --check <changed files>  -> all correctly formatted
npx oxlint <changed files>               -> 0 warnings, 0 errors
```

One iteration of `pnpm check` surfaced a real typing issue in an earlier draft (`AdmissionReopenReason` did not allow the new log reason); that is fixed by extending the union with `"failed restart drain"`, and `tsgo:prod` is clean. A full `pnpm check` run was not completed locally — it stalled at ~0.2% CPU on this machine after the typecheck lane, so the lanes covering the touched surface were run individually as shown above.

Diff is 5 files, +119/−2, with no behaviour change on the committed-restart path.

Refs #140455

### DIFF
--- a/src/cli/gateway-cli/lifecycle.runtime.ts
+++ b/src/cli/gateway-cli/lifecycle.runtime.ts
@@ -14,6 +14,7 @@ export {
   isGatewaySigusr1RestartExternallyAllowed,
   markGatewaySigusr1RestartHandled,
   peekGatewaySigusr1RestartReason,
+  recoverGatewayRestartDrainAfterFailedSignalAdmission,
   resetGatewayRestartStateForInProcessRestart,
   requestGatewayRestartWithSignalAdmission,
   rollbackGatewayRestartSignalAdmission,
--- a/src/cli/gateway-cli/run-loop.test.ts
+++ b/src/cli/gateway-cli/run-loop.test.ts
@@ -54,6 +54,7 @@ const resetGatewaySuspendCoordinatorForLifecycleRestart = vi.fn();
 const consumeGatewaySuspendHandoff =
   vi.fn<typeof import("../../infra/gateway-suspend-coordinator.js").consumeGatewaySuspendHandoff>();
 const disarmGatewaySuspendHandoff = vi.fn();
+const recoverGatewayRestartDrainAfterFailedSignalAdmission = vi.fn();
 const rollbackGatewayRestartSignalAdmission = vi.fn();
 const requestGatewayRestartWithSignalAdmission = vi.fn(() => ({ status: "emitted" as const }));
 const writeGatewayRestartHandoffSync = vi.fn(
@@ -197,6 +198,8 @@ vi.mock("../../infra/restart.js", async (importOriginal) => {
     peekGatewaySigusr1RestartReason: () => peekGatewaySigusr1RestartReason(),
     resetGatewayRestartStateForInProcessRestart: () =>
       resetGatewayRestartStateForInProcessRestart(),
+    recoverGatewayRestartDrainAfterFailedSignalAdmission: () =>
+      recoverGatewayRestartDrainAfterFailedSignalAdmission(),
     rollbackGatewayRestartSignalAdmission: () => rollbackGatewayRestartSignalAdmission(),
     requestGatewayRestartWithSignalAdmission,
     scheduleGatewaySigusr1Restart: (opts?: { delayMs?: number; reason?: string }) =>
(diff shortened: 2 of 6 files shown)


jai-assistant:
Withdrawing this PR.

I opened it while chasing a `Gateway is draining; new tasks are not accepted` error on the Google Meet voice consult path, and assumed a stuck one-way restart drain was the cause. That assumption was wrong.

The real cause turned out to be unrelated to restart handling: the meeting consult inherits the AsyncLocalStorage root-work context of the already-finished `googlemeet.create` request, and that context is `released`, so `isGatewaySubordinateWorkAdmissionClosed()` returns true and `enqueueCommandInLane` rejects the work with a misleading drain error. Instrumenting the function proved it:

```
DIAG subordinate-closed reason=rootWorkReleased origin=ws:googlemeet.create refs=0
```

Filed as #142610 with the actual root cause, a reproduction, and a verified fix (acquire an own root-work admission outside the stale context, mirroring the working browser Talk path).

The failed-SIGUSR1 recovery gap this PR describes may still be a real edge case in theory, but I never reproduced it and cannot supply the behaviour proof the review asked for. Closing rather than leaving an unverifiable change open — happy to revisit if someone hits it for real.

PR #142171, opened 2026-09-08
fix(tests): keep canonical transport fixtures on source plugins

steipete:
## What Problem This Solves

Resolves a problem where the canonical descendant integration suite fails after a normal runtime build, even though its focused run passes before building. The managed-desktop scenario unexpectedly starts a real Node process instead of using the fixture transport.

## Why This Change Was Made

Bundled discovery can select built plugin manifests, while the native fixture mocks source modules. The fixture now selects its checkout's source plugins through the existing test-state environment scope, which captures and restores the override after cleanup. All scenarios and assertions remain unchanged.

## User Impact

No production behavior changes. Developers can run the canonical integration suite with built artifacts present. This adds four test lines and makes no runtime-performance claim.

## Evidence

- The unchanged prebuilt cohort reproduced the same failure on Linux Node 24.19 and macOS Node 24.20: 71 passed, 1 failed.
- Diagnostic-only observations showed built plugin selection while the source transport mock remained active and unchanged. With the fix, source plugins were selected and all 72 cases passed; 71 fixture cleanups restored the environment.
- Final plain run: 72 passed, 0 failed or skipped, with identical ordered case names and all 180 existing expect expressions preserved. Prebuilt stamp hashes and modification times were unchanged.
- Mapped changed checks and fresh P2 review passed. Diagnostic code is excluded from this PR. Patched Linux validation is left to exact-head CI.

### DIFF
--- a/test/canonical-descendant.integration.test.ts
+++ b/test/canonical-descendant.integration.test.ts
@@ -1,5 +1,5 @@
 import http from "node:http";
-import { pathToFileURL } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { expectDefined } from "@openclaw/normalization-core";
 import { isRecord } from "@openclaw/normalization-core/record-coerce";
 import {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [93e3036fc52d1afc9bf499dccd70578930b7b132](https://github.com/openclaw/openclaw/pull/142171/commits/93e3036fc52d1afc9bf499dccd70578930b7b132)
- Landed commit: [7b2058c2ecc4bdbd6ab1d63b73af6b57a1249823](https://github.com/openclaw/openclaw/commit/7b2058c2ecc4bdbd6ab1d63b73af6b57a1249823)

PR #140637, opened 2026-09-07
fix(discord): restore session resets for sender aliases

steipete:
Related: #140535

## What Problem This Solves

Fixes an issue where Discord users invoking `/new` or `/reset` receive “No reply was generated” and retain their conversation context when the command allowlist uses the documented `user:<id>` form. The same normalization gap affects existing `discord:` and `pk:` aliases and Discord mentions. This reproduces the symptom reported by @chrismanderson in #140535; the report does not include its exact allowlist configuration.

## Why This Change Was Made

Discord's native command authorizer recognizes sender prefixes, but the shared Discord config adapter only lowercased the entries. Core reset authorization therefore compared a prefixed allowlist entry with the bare sender ID and rejected an already admitted command (`extensions/discord/src/shared.ts:93`, `src/auto-reply/command-auth.ts:359`, `src/auto-reply/reply/commands-reset.ts:44`). Normalize the existing aliases and mentions in the owning adapter, using the same SDK formatter option already used by Telegram. Preserve the existing Doctor-only migration of retired bundled-channel owner entries in the canonical core owner parser, without changing third-party native identities.

The reset target is already carried correctly, so changing slash-session routing or adding a retry/fallback would not repair this failure. #137313 is not the introduction: its merge commit is not an ancestor of the 2026.9.2 release commit. The adjacent group-system-prompt helper now returns its single trimmed value directly instead of constructing, filtering, and joining a one-element array.

## User Impact

Authorized Discord reset commands clear the routed channel context and return their normal acknowledgement. Durable session IDs remain stable; the lifecycle revision and reset event prove the context reset. Explicit denial still prevents reset. Retired global-owner entries still require `openclaw doctor --fix`, and native third-party owner identities remain supported. The command documentation now names the supported user-ID aliases.

## Evidence

Head: 5f4815fe901108ca489c1826c60019b0320672a0

- Baseline synthetic native interaction: bare sender ID passes; `user:`, `discord:`, and `pk:` each return the generic fallback with the original lifecycle revision unchanged.
- Fixed regression executes the real Discord native handler, channel dispatch, shared authorization/reset owner, SQLite session store, and reply pipeline. Only the interaction transport is synthetic. It checks both `/new` and `/reset`, durable reset events, normal acknowledgements, a mismatched sender, and an empty allowlist.
[...]

PR #142176, opened 2026-09-08
fix(agents): retry silent malformed tool rejections

### LINKED ISSUE #135111: [Bug]: Intermittent "Provider completed tool call with malformed JSON arguments" on v2026.8.1 (claude-sonnet-5), not tied to specific file/tool
### Bug type

### Beta release blocker

No

### Summary

Since upgrading from v2026.7.1-2 to v2026.8.1, agent runs intermittently fail with:

```
LLM request failed. rawError=Provider completed tool call with malformed JSON arguments
```

This has occurred ~6 times over a few hours in a single agent's coding sessions. It is **not** tied to a specific tool or file — I've seen it after:
- An MCP tool call (Serena `search_for_pattern`, searching Microsoft BaseApp AL source with CRLF line endings) — initially suspected as the cause, but ruled out by the next occurrence.
- The built-in `read` tool on a plain internal Markdown file (`AL-CODING-STANDARDS.md`, LF line endings, nothing unusual in content).

Both cases show the same pattern: the tool result content is stored as `"[Malformed diagnostic JSON redacted]"` in the transcript (OpenClaw's own export redaction, confirming the raw tool-result JSON itself was invalid), and the following assistant turn ends with `stopReason: "error"` / `errorMessage: "Provider completed tool call with malformed JSON arguments"`.

## Environment

- OpenClaw 2026.8.1 (ea80657)
- Model: `anthropic/claude-sonnet-5`
- Self-hosted gateway, systemd service, Ubuntu 26.04
- Agent runtime: embedded, sandboxed (Docker)
- Occurred both in a long-running session (with prior failed turns in history) and in a **brand-new session** with no prior context — ruling out "poisoned session context" as the sole explanation.

## Reproduction (partial — not fully deterministic)

1. Start an agent turn that calls a tool (either an MCP tool like Serena's `search_for_pattern`, or the built-in `read` tool) mid-task.
2. Intermittently, the tool result is stored as malformed JSON, and the subsequent assistant turn fails with `stopReason: error`.
3. Retrying (new prompt in same session, or brand-new session) sometimes succeeds, sometimes fails again — no clear deterministic trigger found yet.

## What I ruled out

- **File-specific CRLF/BOM content**: initially suspected because the first two failures both happened on Microsoft BaseApp source files with CRLF line endings and a UTF-8 BOM. A third failure occurred on a plain LF-only internal file via the built-in `read` tool, ruling this out as the sole cause.
- **Poisoned session context**: a brand-new session (no prior failed turns in its history) still failed the same way.
- **Session observer feature**: also noticed `[session-observer] session observer disabled after consecutive failures` recurring ~11x in the same window (separate background feature, `gateway.controlUi.sessionObserver`, uses a utility-model JSON completion). Disabled it via config as a precaution; the tool-call malformed-JSON errors persisted afterward, so this looks like a separate (but possibly related — both are JSON-parsing-of-model-output issues) symptom.

### Expected behavior

Tool calls should either succeed with valid JSON arguments/results, or fail with a specific, actionable error (e.g. identifying which tool, which argument, or a retry-with-backoff) — not a generic "malformed JSON arguments" that requires starting an entirely new session to recover from. Ideally the raw malformed payload (or at least byte-offset/length) would be logged server-side for diagnosis, since the redacted transcript currently only shows `[Malformed diagnostic JSON redacted]`.

### Actual behavior

Errior: LLM request failed. rawError=Provider completed tool call with malformed JSON arguments

### OpenClaw version

2026.8.1 (ea80657)

### Operating system

Ubuntu 26.04

### Install method

npm global

### Model

Anthropic Sonnet 5

### Provider / routing chain

Anthropic API (direct) — provider: anthropic, model: claude-sonnet-5, api: anthropic-messages Self-hosted OpenClaw gateway (systemd service, single instance, no proxy/LiteLLM/router in between) No custom base URL override; standard https://api.anthropic.com/v1/messages endpoint (confirmed via gateway logs: [provider-transport-fetch] url=https://api.anthropic.com/v1/messages)

## Diagnostics available

I have exported redacted trajectory bundles via `openclaw sessions export-trajectory` for two of the failing sessions, showing the exact point of failure (happy to attach on request — they're already redacted by OpenClaw's own export tool, but I'd rather not post internal repo paths/names publicly without review).

## Question

Is this a known issue with Anthropic tool-use streaming parsing in 2026.8.1, or something specific to my setup? Any pointers on what logging/flags to enable to capture the raw (pre-redaction) malformed payload for a proper bug report would be appreciated.

### Logs

```shell

```

[OpenClaw-Issue-Logs.zip](https://github.com/user-attachments/files/31686472/OpenClaw-Issue-Logs.zip)

### Impact and severity

_No response_

### Additional information

_No response_

lraesly:
## What Problem This Solves

A provider can consume output tokens before rejecting a malformed tool call. When that rejection leaves no visible answer or executable action, positive token usage prevented OpenClaw's existing silent-error retry. The user received a failed turn after one request even when a second attempt could succeed.

## Why This Change Was Made

Let the existing retry owner recognize the stable malformed-tool error code and five exact transport-owned rejection messages after its visibility, terminal-state and replay-safety checks. Message matching preserves case, whitespace and punctuation exactly. Parsing stays strict, and the existing budget remains three retries after the initial request.

## User Impact

Transient malformed tool output can recover through the same model or the existing configured fallback. Visible output, admitted actions, possible effects, active work, approvals and terminal failures retain their retry vetoes. A valid tool call from a later attempt still follows normal policy.

Fixes #135111.

## Evidence

- Real isolated built Gateway through public `chat.send`, `agent.wait` and `chat.history`, with local Messages/Responses streams and counted tool actions.
- Main: one failed request and zero rejected-call executions. Candidate: recovery in two requests; a later valid tool executes once across three requests; repeated rejection stops at four requests; configured fallback succeeds on request five.
- Independent public scenarios cover same-session replay, visibility, accepted work, real background activity, pending approval and owner cancellation, refusal, timeout, abort, authentication/context errors, and unrelated Responses errors. Additional real channel delivery and an potentially mutating action that made no domain change preserve the retry fence.
- 316 focused tests pass, including the exact-match regressions, retry/fallback owner, asynchronous settlement and permanent WebSocket-close producer. Formatting, lint, size/assertion guards and the native commit hook pass.

Native early asynchronous provider eligibility and typed terminal-code precedence are covered at their unchanged component boundaries. The local provider fixtures do not claim an official-provider live result.


lraesly:
Pushed 444b8843b8 and updated the PR body with after-fix runtime proof: a new `test/pr142176-real-runtime-proof.e2e.test.ts` boots a real gateway against a loopback Anthropic-format provider whose first streaming reply is a truncated tool call and whose second is plain text. On this branch `agent.wait` returns `status: ok` after two `POST /v1/messages`; with upstream/main sources the same test gets `status: error` ("Agent run failed") after one request. Also fixed the two CI items from the first run (typed usage fixture for tsgo; no casts in the gate helper for the assertion ratchet). The remaining red shard was a Telegram sticker image-model test this PR does not touch.

lraesly:
Pushed 848a0ecceb (test only): pins that once the three resubmits are spent and a fallback chain is configured, the rejection routes through the existing `exhaustedUnclassifiedSilentError` path to `fallback_model` instead of surfacing. PR body also gains a bound/replay check from the gateway harness with an always-rejecting provider: 4 requests per turn, then the same error as today, and the next turn's request carries nothing from the rejected turn (no tool block, no error text).

obviyus:
landpr-approve-unrelated-red
pr: 142176
head: b0e728bda06a925a9fa79f57855a8d0f6352c116
run: 34304870127 attempt: 1
failed: checks-node-changed-extensions-bundle-13 job=102319504622 conclusion=failure
check: openclaw/ci-gate

obviyus:
Thanks @lraesly! Landed in 6ac98274afd3ed493ba69edbb0f6dea1ef1d1c05.

**Proof:** 316 focused tests and 24 independently exercised Gateway scenarios passed, including strict rejection, counted tool effects, replay, pending approval and cancellation.

- **Bug:** positive output usage prevented an otherwise safe retry after malformed tool arguments were rejected.
- **Fix:** exact rejection signals use the existing bounded retry after all output, effect and terminal guards.

| Before | After |
|---|---|
| One failed request; rejected call never executes | Recovery in two requests; a later valid tool executes once |
| No recovery from repeated rejected calls | Four-request limit, then existing failure or configured fallback |

**CI exception:** [run 34304870127, attempt 1](https://github.com/openclaw/openclaw/actions/runs/34304870127) completed with one failed child, [checks-node-changed-extensions-bundle-13 / job 102319504622](https://github.com/openclaw/openclaw/actions/runs/34304870127/job/102319504622), and the resulting aggregate gate failure. The first bare Discord `/new` reset test timed out at 120 seconds; 3,695 other tests in that shard passed. This command returns its reset acknowledgement before model execution, so the changed provider retry branch is not reached. The test and reset/session/dispatch owners are unchanged in the exact tested merge.

The [main-parent run](https://github.com/openclaw/openclaw/actions/runs/34304824122) passed but did not select Discord tests; it is not claimed as a matching reproduction. No matching red main run or exact stalled await was established. The unchanged test came from [#140637](https://github.com/openclaw/openclaw/pull/140637), commit 93562ca9054dea9eda853d9c527190a19e8581ec; that is test provenance, not blame for the timeout. Used the [guarded standing approval](https://github.com/openclaw/openclaw/pull/142176#issuecomment-5595275915) for this causally separate failure. No timeout, assertion, source or CI rerun was changed to hide it.

lraesly:
Thank you @obviyus, both for landing it and for the write-up: the independent 24-scenario gateway run and the before/after table are a better summary of the change than my PR body was. Appreciate the care on the unrelated red shard too.

PR #142178, opened 2026-09-08
fix: join audio transcript completion in gateway tests

steipete:
Related: #139428

## What Problem This Solves

The agent-run audio transcript test asserted persistence after the chat acknowledgement, even though media materialization could still be running. It also supplied a four-byte placeholder to the real audio metadata probe.

## Why This Change Was Made

Await the existing admission release for the exact store and session before checking the transcript. Inline transcript updates complete before that release, so the original assertions can run directly. Use the existing valid MP3 fixture in the four audio cases sharing this playback contract.

## User Impact

No production behavior changes. The tests retain every transcript oracle and exercise valid audio while joining the operation that owns persistence. This adds two test lines overall.

## Evidence

- Original directive suite: 216 passed; final directive suite plus dispatch/media siblings: 267 passed across four files. All 216 original directive cases retain their names and order.
- A temporary test-only gate held the real metadata probe: the acknowledgement had returned, admission remained active, and no audio update existed. Releasing the probe and joining admission produced the expected persisted update and zero outstanding admissions. All diagnostic instrumentation was removed.
- Changed-file checks and independent P2 review passed.

The stage proof establishes the completion boundary; it does not establish the cause of the historical Linux timeout or a performance improvement. No deadlines, production code, or assertions were weakened.

### DIFF
--- a/src/gateway/server-methods/chat.directive-tags.test.ts
+++ b/src/gateway/server-methods/chat.directive-tags.test.ts
@@ -14,6 +14,7 @@ import {
 } from "../../../packages/gateway-protocol/src/client-info.js";
 import { ErrorCodes } from "../../../packages/gateway-protocol/src/index.js";
 import { CHAT_SEND_SESSION_KEY_MAX_LENGTH } from "../../../packages/gateway-protocol/src/schema.js";
+import { createPlaybackMediaFixture } from "../../../test/fixtures/media-playback.js";
 import { createDeferred } from "../../../test/helpers/promise.js";
 import {
   bindActiveCronCreatorAuthorityResolver,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [905410ff4e4c6f4542ae96cea050e85169ac251c](https://github.com/openclaw/openclaw/pull/142178/commits/905410ff4e4c6f4542ae96cea050e85169ac251c)
- Landed commit: [451a568c2e75284ed42aa8c861b153e649730238](https://github.com/openclaw/openclaw/commit/451a568c2e75284ed42aa8c861b153e649730238)

PR #142179, opened 2026-09-08
fix: doctor read-only database inspections time out on large agent databases

steipete:
Related: #141918, #141412.

## What Problem This Solves

Fixes an issue where read-only inspections of a large SQLite database, such as the memory-core readiness check in `openclaw doctor`, fail with "SQLite read-only snapshot timed out after 30 seconds" whenever the database is too large to copy in 30 seconds. On the team Gateway the 9.4 GB agent database fails this way on every doctor run.

## Why This Change Was Made

The read-only worker builds a private copy with SQLite's online backup API, so its cost grows with file size, but its child process had the fixed 30-second inspection deadline sized for a 300 MiB synthetic database. #141918 already gave the integrity child a size-derived budget; this change moves that budget into one shared helper and applies it to both read-only worker entry points: 30 seconds for startup and shutdown plus one second per 32 MiB, rounded up and capped at 30 minutes. The timeout error names the applied budget and file size, and budgets above the base are logged. Non-goals: when snapshots are taken is unchanged, and whether tiny metadata reads should use a live read-only connection instead of a full snapshot is a separate owner decision.

## User Impact

Doctor and other read-only inspections complete on large installs instead of reporting a spurious timeout; operators see the applied budget when a copy genuinely exceeds it.

## Evidence

Team host (2026-09-08): `openclaw doctor --lint --all --json` on the 9.4 GB agent database failed the memory-core check with the 30-second snapshot timeout; a niced online backup of the same file took 115 seconds niced (about 87 MB/s) on 2026-09-08 at 13:06 UTC, against which the size-derived budget for that file is 331 seconds. Regression tests fail on the pre-fix code for the intended reason and verify the size-derived budget for both the async and sync worker paths and the new error text. `node scripts/check-changed.mjs` passed; focused snapshot, cancellation, integrity, and read-only worker suites pass. Codex-backed autoreview at P1 before landing.

## Known Gaps

Each inspection still copies the whole database into the private staging directory; large installs pay that disk and I/O cost per doctor run.


steipete:
Merged via squash.

- Prepared head SHA: [60f4f8f5291d3a3cc779626541137e21ce7ec6cd](https://github.com/openclaw/openclaw/pull/142179/commits/60f4f8f5291d3a3cc779626541137e21ce7ec6cd)
- Landed commit: [c245214aa7178e3a4cdae2c39fb228cb7719bb1d](https://github.com/openclaw/openclaw/commit/c245214aa7178e3a4cdae2c39fb228cb7719bb1d)

PR #140742, opened 2026-09-07
refactor(logging): share portable redaction pattern mechanics

steipete:
## What Problem This Solves

Dashboard and backend redaction independently parse the same pattern literals, decode replacement callback arguments, and mask PEM blocks. Maintaining those mechanics twice risks drift while each surface intentionally has different privacy rules.

## Why This Change Was Made

A browser-safe leaf under the logging owner shares pattern source/flag parsing, concrete match context, and PEM framing. Backend compilation still uses the safe-regex validator and retains capture indexing, source-assignment preservation, bounded execution, registered secrets, and structured-field policy. Dashboard token hints, aggressive field/path masking, and literal replacement handling stay local. Each caller supplies its existing PEM marker.

The app-password path uses its existing token mask directly; its fixed pattern captures the entire unquoted password. No masking thresholds, suffixes, pattern order, or config semantics changed.

## User Impact

None intended. Existing redaction outputs and public entry points are preserved. No config option, dependency, protocol, or schema change.

## Evidence

Fresh validation at `e3da26703216b1de2c5dbc759a42d51d3415d5e7`, rebased onto main `4b1195fb3566a79a8a644e692b4b7f7f7ac26546`. `git range-diff` confirms an identical patch.

- `node scripts/run-vitest.mjs src/logging/redact ui/src/lib/browser-redact`: 3 files, 219 tests passed, including named captures, source preservation, app passwords, UTF-16 boundaries, literal replacements, and distinct browser masking.
- `node scripts/check-changed.mjs`: passed; no baseline edits.
- `pnpm build`: passed in 7m 50s, including declarations, 91 plugins, runtime closure checks, and Dashboard asset budgets. Startup JS was 349,600 / 350,953 B gzip, with 8 / 18 requests.
- Fresh Codex autoreview: scoped-clean at the requested command's default P0 threshold, zero findings. The original validation also recorded a clean P0–P2 review.
- Production +67/-96, net -29 lines; no test changes. No existing public export removed or renamed.
- [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34120339859) passed on the exact reviewed head without reruns.

The previous Workshop geometry failure is historical; this revision is being verified on current main. No redaction thresholds, regex order, configured-pattern semantics, or stored output policies change.

PR #136073, opened 2026-09-02
refactor(logging): reduce redaction capture overhead

steipete:
## What Problem This Solves

Secret redaction builds temporary collections for every regex match while processing logs and tool output.

## User Impact

Masking behavior is unchanged. The redactor avoids disposable capture objects and arrays, and skips fallback work when an exact capture position is already available.

## Why This Change Was Made

Select the last nonempty capture in one pass and keep the existing indexed-position precedence. This removes one production line without adding a cache or changing redaction policy.

## Evidence

All 246 selected tests passed across redaction, nested calls, and the logging transport. Formatting and independent source and managed P2 reviews passed. The selected changed-scope typechecks, lint, and repository guards also passed. A fixed synthetic workload through `redactSensitiveText` produced identical complete outputs in four independent processes. Across 3,000 calls per process, V8 sampled allocations averaged 163.09 MB before and 123.93 MB after, a 24.0% decrease (two samples per version, ABBA order, same Node and dependencies). This measures gross allocations in that workload; live Gateway memory and latency effects are not established by this probe.

Exact-head CI [run 34970864268](https://github.com/openclaw/openclaw/actions/runs/34970864268) passed.

PR #142183, opened 2026-09-08
refactor(logging): share assignment detection across redaction outputs

### LINKED ISSUE #142162: refactor(logging): share assignment detection between redaction outputs
The logging redactor independently detects the same four assignment forms for two consumers: replacing sensitive text and marking sensitive source positions. Eight private scanners repeat key admission, quote/suffix handling and value-offset extraction. This makes a future grammar change require paired edits even though the consumers intentionally use different output and pipeline-order contracts.

This maintainer-requested cleanup gives form, URL-query, encoded-key and body-context assignments one closed detector with separate replacement and bitmap sinks. Keep every existing grammar, mask, original offset and pipeline order: text passes continue operating on rewritten text, while coverage passes continue marking the original input. Keep public entry points, source-preservation policy, custom patterns, registered secrets, multiline behavior and approval display semantics unchanged.

The complete draft removes 76 production code lines (80 physical), including every replacement and caller. This is a bounded ownership cleanup within the larger code-reduction campaign. The earlier #140742 shared portable pattern mechanics; assignment detection remains separate from that work.

Validation will compare all affected public redaction surfaces and the approval-display consumer before/after, retaining existing tests and covering quoted/empty/encoded values, controls and Unicode, repeated calls, line arrays, exact bitmaps and bounded adversarial input sizes. No changes to redaction policy, dependencies, configuration, or persistent data are proposed.

Related work is separate: #136073 changes capture selection outside these scanners; #118211 expands the shared key classifier and separate Gateway-client parsers. This refactor retains those dependencies and does not absorb their behavior changes.

steipete:
Closes #142162

## What Problem This Solves

The logging redactor detects the same assignment forms twice: once to replace sensitive text and again to identify protected source positions for approval display. Eight private scanners duplicate key admission and value-span extraction, requiring paired edits when the grammar changes.

## Why This Change Was Made

Use one private detector for form pairs, URL queries, encoded keys and explicit body contexts, with separate text-replacement and bitmap sinks. Text passes still operate on rewritten text; bitmap passes still use original UTF-16 coordinates. Grammar, mask policy, quotes/suffixes, configured patterns and all public entry points remain unchanged.

This complete cleanup removes 74 production code lines and adds 13 test code lines, net 61 maintained lines. It retains the existing portable pattern helpers from #140742. Capture-selection work in #136073 and classifier changes in #118211 remain separate.

## User Impact

No intended user-visible behavior change. Logs, tool output and approval displays retain their masking and source-position behavior. Future assignment detection changes have one owner serving both outputs.

## Evidence

- All existing tests remain. The final affected redaction, approval-display and token-ordering suites pass 264 tests. One unchanged structured-auth log-tail case passed both the baseline and initial consolidation.
- One added public-boundary test protects original offsets when an empty value expands before a quoted secret after CRLF and a non-BMP prefix; it also passes on the original implementation.
- 1,987 synthetic inputs / 22,277 observations match the frozen original exactly, plus source-policy, custom-regex, registered-secret, flag and error cases. Comparisons include full bitmaps, unmarked positions, line arrays and the actual approval sanitizer.
- The complete selected changed gate and fresh independent P2 review pass. No new imports, dependencies, public APIs or production test seams; no additional full build or live-provider proof.
- Paired, warmed Node 24.20.0 measurements retain all 168 scenarios through the approval input cap. A seven-sample crafted short-line text case is about 6% slower in wall time and 3% in CPU; its sampled end-to-end approval case is about 3% slower. No pathological growth was observed across the measured sizes. The private regex loop resets state in `finally`; the larger initial iterator-cloning overhead was removed before publication.

### DIFF
--- a/src/logging/redact.test.ts
+++ b/src/logging/redact.test.ts
@@ -994,6 +994,21 @@ describe("redactSensitiveText", () => {
     expect(headerBitmap.slice(header.indexOf("=") + 1).every(Boolean)).toBe(true);
   });
 
+  it("keeps original bitmap offsets after empty values and Unicode line prefixes", () => {
+    const input = '😀safe\r\nbody: code=&safe=1\rclient%5Fsecret="abc";&safe=2';
+    const resolved = resolveRedactOptions({ mode: "tools" });
+    const bitmap = computeSensitiveRedactionBitmap(input, resolved);
+    const secretStart = input.indexOf('"abc"');
+
+    expect(redactSensitiveText(input)).toBe(
+      "😀safe\r\nbody: code=***&safe=1\rclient%5Fsecret=***;&safe=2",
+    );
+    expect(bitmap).toHaveLength(input.length);
+    expect(bitmap.slice(0, secretStart).some(Boolean)).toBe(false);
+    expect(bitmap.slice(secretStart, secretStart + 5).every(Boolean)).toBe(true);
+    expect(bitmap.slice(secretStart + 5).some(Boolean)).toBe(false);
+  });
+
   it("masks token prefixes embedded after adjacent text", () => {
     const token = `ghp_${"a".repeat(5_000)}`;
     const output = redactSensitiveText(`prefix-${token} suffix`, { mode: "tools" });
--- a/src/logging/redact.ts
+++ b/src/logging/redact.ts
@@ -322,29 +322,81 @@ function hasEncodedOrInvisibleFormKey(key: string): boolean {
   );
 }
 
-function redactFormEncodedPairs(
-  value: string,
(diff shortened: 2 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [fa127ce67408806ff6527ea73427073c8d3a2ed7](https://github.com/openclaw/openclaw/pull/142183/commits/fa127ce67408806ff6527ea73427073c8d3a2ed7)
- Landed commit: [044f77ad303f107b4fe87ee055fa1d18ab352f47](https://github.com/openclaw/openclaw/commit/044f77ad303f107b4fe87ee055fa1d18ab352f47)

PR #142184, opened 2026-09-08
refactor(tests): share timestamped assistant inputs

steipete:
## What Problem This Solves

Compaction and replay tests repeat timestamped assistant-text input objects around the behavior they protect.

## Why This Change Was Made

Reuse one required-text/timestamp constructor at 34 input sites. It preserves exactly the original role/content/timestamp fields and fresh message, content-array and text-block allocations. Existing casts, independent expected values and the single shared `ASSISTANT_MSG` initializer remain intact; all existing sparse helper implementations are unchanged.

The formatted change removes 128 net test/support lines. Four of those lines come from consequential collapse of the two enclosing thinking-repair call wrappers; the literal/import/helper reduction is 124 lines.

## User Impact

No production behavior changes. Existing scenarios, assertions and deadlines remain intact; the constructor adds no defaults or production imports.

## Evidence

- All 208 cases passed before and after in the same four complete suites, with identical per-file ordered case names and statuses and none pending.
- Inline expansion matches all 34 original inputs, including casts and the shared-constant reference graph. Actual-helper checks verify 68 fresh constructions and exact values, keys, descriptors and prototypes.
- Mapped changed gate passed. Deslop and fresh independent P2 review are clean.

No runtime improvement or coverage-percentage claim.

### DIFF
--- a/src/agents/agent-hooks/compaction-safeguard.test.ts
+++ b/src/agents/agent-hooks/compaction-safeguard.test.ts
@@ -16,6 +16,7 @@ import {
 import * as compactionModule from "../compaction.js";
 import { buildEmbeddedExtensionFactories } from "../embedded-agent-runner/extensions.js";
 import { castAgentMessage } from "../test-helpers/agent-message-fixtures.js";
+import { timestampedTextAssistant } from "../test-helpers/sparse-transcript.test-support.js";
 import { createZeroUsageFixture } from "../test-helpers/usage-fixtures.js";
 import { jsonResult } from "../tools/common.js";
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [19c9372d6615ac6c90d99d7b8ffb14e3f0ed06f2](https://github.com/openclaw/openclaw/pull/142184/commits/19c9372d6615ac6c90d99d7b8ffb14e3f0ed06f2)
- Landed commit: [efedb9452e617df4eec81fda3a08d23f4d1fc137](https://github.com/openclaw/openclaw/commit/efedb9452e617df4eec81fda3a08d23f4d1fc137)

PR #142186, opened 2026-09-08
fix(daemon): Bun runtime probe ignores the Gateway's SQLite library selection

steipete:
Related: #141854

## What Problem This Solves

Fixes an issue where `openclaw gateway install --runtime bun`, `openclaw doctor`, and the Gateway service audit would judge a Bun executable by Apple's system SQLite on macOS, while the Gateway itself (since #141854) selects an extension-capable Homebrew/MacPorts library or the `OPENCLAW_SQLITE_LIBRARY` override before its first `node:sqlite` open. On a macOS whose system SQLite is below the WAL-reset safety floor, daemon install rejected a Bun that the Gateway runs fine with Homebrew's library. In every case the probe reported the wrong SQLite version, and an invalid `OPENCLAW_SQLITE_LIBRARY` that makes the Gateway exit at startup was still judged "supported".

## Why This Change Was Made

The probe now applies the same selection semantics as startup: the selection module is registered as a stable runtime process entry (`dist/infra/bun-sqlite-library.js`, or the source module in dev checkouts, which Bun transpiles) and Bun candidates `require` it and call `ensureSqliteLibrarySelected()` before opening `node:sqlite`. Node probes never load it. The probe child gets a minimal environment (PATH, temp and Windows system keys, plus the two selection inputs `HOMEBREW_PREFIX` and `OPENCLAW_SQLITE_LIBRARY`) instead of the full shell environment. A rejected override is reported as an unsupported result carrying the exact startup error: preferred-Bun resolution throws it instead of advising a Bun upgrade, the service audit shows it as the issue detail, and start repair refuses to silently rewrite a Bun service to Node. A missing selection module fails the whole probe (broken install), not "unsupported". No config options were added.

Non-goal: the managed service environment allowlist still does not forward `OPENCLAW_SQLITE_LIBRARY`/`HOMEBREW_PREFIX` to launchd/systemd; the probe mirrors the invoking environment, matching what `openclaw gateway` does in that shell. Persisting the override into the service definition is a separate, approval-gated decision.

## User Impact

Bun daemon installs, repairs, doctor, and service audits now accept and report the SQLite library the Gateway will actually open. An invalid `OPENCLAW_SQLITE_LIBRARY` fails the install or repair with the same "Fix or unset OPENCLAW_SQLITE_LIBRARY" message the Gateway prints at startup, instead of a misleading "install Bun 1.4+" hint or a silent switch of the service to Node. The probe no longer inherits secrets or `NODE_OPTIONS` from the invoking shell.

## Evidence

Live reproduction on a Mac with Bun 1.4.2, Apple SQLite 3.54.0, Homebrew SQLite 3.53.4 (versions differ, so selection is observable without synthetic data). Startup guard under the same Bun selects Homebrew 3.53.4 and, with an invalid override, exits 1.

Before (`resolveBunRuntimeInfo("/opt/homebrew/bin/bun")`, source):

```
daemon probe (no override):      {"status":"supported","version":"1.4.2","sqliteVersion":"3.54.0","nodeSharedSqlite":false}
daemon probe (invalid override): {"status":"supported","version":"1.4.2","sqliteVersion":"3.54.0","nodeSharedSqlite":false}
```

After (source, and identically from the built `dist/runtime-paths-*.mjs` chunk after `pnpm build`):

```
daemon probe (no override):      {"status":"supported","version":"1.4.2","sqliteVersion":"3.53.4","nodeSharedSqlite":false}
daemon probe (invalid override): {"status":"unsupported","version":"1.4.2","sqliteVersion":null,"nodeSharedSqlite":false,"sqliteSelectionError":"Cannot use SQLite library /nonexistent/libsqlite3.dylib: missing file. Fix or unset OPENCLAW_SQLITE_LIBRARY; install a supported library with brew install sqlite."}
resolvePreferredBunPath (invalid override) threw: Cannot use SQLite library /nonexistent/libsqlite3.dylib: missing file. Fix or unset OPENCLAW_SQLITE_LIBRARY; ...
```

Node probe path with the minimal env (Homebrew Node 26.8.1, shared SQLite 3.53.4) still resolves as supported. `bun -e 'require("dist/infra/bun-sqlite-library.js")...'` loads the stable dist entry and selects Homebrew's library.

Tests: `src/daemon/runtime-paths.test.ts` (68), `src/daemon/service-audit.test.ts` (70), `src/cli/daemon-cli/start-repair.test.ts` (20) green, including new cases for the selected-library success (asserting the `require` line and the exact minimal env), Node never loading the module, the invalid-override result shape, the preferred-path throw, the audit detail, and the repair refusal. `check:changed` green (typecheck core + core tests, deadcode, lint after compacting the audit test file under the 1000-line cap), `check:import-cycles` and `check:madge-import-cycles` 0 cycles, Codex autoreview at P1 scoped-clean.

### DIFF
--- a/docs/install/bun-compatibility.md
+++ b/docs/install/bun-compatibility.md
@@ -54,6 +54,8 @@ Cannot use SQLite library <path>: <reason>. Fix or unset OPENCLAW_SQLITE_LIBRARY
 
 Node and non-macOS Bun ignore this override, with a warning in Gateway startup logs. When a library is selected, Gateway startup logs `SQLite: using <path> (<version>, extension loading enabled)`. `openclaw doctor` reports the selection for the doctor process.
 
+Daemon install, `openclaw gateway start` repair, `openclaw doctor`, and service audits probe candidate Bun executables through the same selection, so they judge and report the library the Gateway will actually open rather than Bun's runtime SQLite. An invalid override fails those probes with the message above instead of advising a Bun upgrade or switching the service to Node.
+
 If you previously used a preload that calls `Database.setCustomSQLite()`, remove it and set `OPENCLAW_SQLITE_LIBRARY` to the same path instead. The hook is one-shot: keeping the preload causes `SQLite already loaded`, even if both selections name the same library. OpenClaw's override also forwards the path to the KNN child.
 
 ## Memory search without an extension-capable library
@@ -73,6 +75,7 @@ See [Bun](/install/bun) for the workflow and lifecycle trust commands.
 
 | Release                            | Change                                                                                                                                                                                               |
 | ---------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Unreleased (main)                  | Daemon install, repair, doctor, and service audits probe Bun executables through the same SQLite library selection as Gateway startup, with a minimal probe environment. #142186                     |
 | Unreleased (main)                  | Automatically selects a WAL-safe, extension-capable macOS SQLite library and propagates it to the memory KNN child. Adds `OPENCLAW_SQLITE_LIBRARY`. #141854                                          |
 | Unreleased (main)                  | Documents Bun 1.4.2 retaining native statements and WAL/shared-memory files after close or disposal, with Node advised when prompt file release matters. #141846                                     |
 | Unreleased (main)                  | Adds batched embedding-scan fallback when the KNN child cannot load extensions, preserving provider/source filters and cancellation between batches. #141104                                         |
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [55ff0a77145b7dcd005a1a8d732ce52e32187bc8](https://github.com/openclaw/openclaw/pull/142186/commits/55ff0a77145b7dcd005a1a8d732ce52e32187bc8)
- Landed commit: [a8ac0975148fca79feabde82085500d04c94a6a5](https://github.com/openclaw/openclaw/commit/a8ac0975148fca79feabde82085500d04c94a6a5)

PR #142187, opened 2026-09-08
fix(gateway): serialize session timing clears in events

steipete:
## What Problem This Solves

A loaded session could keep the previous run's duration after a new run started. Lifecycle projection clears endedAt and runtimeMs in memory, but undefined properties disappear from JSON. Subscribers merge the remaining fields over their existing row, retaining the old timing until a replacement list arrives.

## Why This Change Was Made

Serialize absent terminal timing as explicit null tombstones in the existing Gateway event projection. Both flat events and nested session snapshots carry the clear. Known terminal timestamps and durations remain numeric, and existing consumers already interpret null as a clear instruction.

## User Impact

Starting another run clears its predecessor's end time and duration immediately, including when the next roster refresh is delayed. The repair restores the existing event contract without adding a client-side reset rule, configuration or persistent data change.

## Evidence

The existing lifecycle model-source table now exercises a completed row followed by a serialized start. Before repair, all three cases failed on missing timing tombstones while 11 controls passed. All 14 payload cases pass after repair; terminal controls also preserve numeric timing.

A real session-capability control receives a serialized null-tombstone start while the next list response is held. It clears previous timing immediately without losing the child title, parent, membership or sibling. This is an expected-passing consumer contract control; the Gateway JSON test owns causal before/after proof.

The focused total is 1,019 passing cases across 13 disjoint files: 14 payload, 27 consumer, 385 sibling, 559 RPC/broadcast and 34 queued-collector cases. Follow-up, recovery and collector event expectations require the explicit wire clears; ordinary history, list and stored-row expectations remain unchanged. Known `runtimeMs: 10` remains numeric in the follow-up fixtures. The shared helper preserves first-call ordering and exact routing while removing its manual mock-call cast.

Selected changed checks passed after each correction. Independent managed review completed two passes on the full initial repair, followed by a clean review of the final one-line collector assertion. These are synthetic boundary tests, with no live Gateway, provider or native app proof claimed.

CI exposed four old omission expectations, all reproduced and corrected locally. The branch now includes the separately landed Telegram runtime-preparation fix in #142234, UI event-log fixture teardown fix in #142314, and model-catalog API reconciliation in #142346. Each integration preserved the complete six-file timing patch and its source hashes. Exact-head [CI](https://github.com/openclaw/openclaw/actions/runs/34255307038) passed on the final head; earlier failed logs remain recorded, including the unresolved CLI child-process timeout.


steipete:
Merged via squash.

- Prepared head SHA: [b07c635d5695620df817e850bbe17dfa0f1d871b](https://github.com/openclaw/openclaw/pull/142187/commits/b07c635d5695620df817e850bbe17dfa0f1d871b)
- Landed commit: [c2d29e7678a8adeecd2c0207acb0d8c18b6d5156](https://github.com/openclaw/openclaw/commit/c2d29e7678a8adeecd2c0207acb0d8c18b6d5156)

Issue #132736, opened 2026-08-29
Auto-TTS supplement messages on Signal deliver as broken attachment (send hangs in signal-cli daemon)


reallyseth:
## Summary

Auto-TTS voice supplements on the **Signal** channel fail to deliver: the recipient's Signal client shows a broken attachment / "Original message not found" error where the audio supplement should be. Plain text replies deliver fine, and manually sending the *same audio* via `openclaw message send --media` delivers perfectly every time.

## Environment

- OpenClaw 2026.7.1-2 (npm), macOS 26.6.2 (arm64), Node 24.15.0
- Signal channel: signal-cli daemon (json-rpc, HTTP on 127.0.0.1:8080), single-account mode
- TTS: `messages.tts.auto: "always"`, provider `openai` (OpenAI-compatible) pointing at a local Kokoro server (mlx-audio) at `http://127.0.0.1:19280/v1`, model `mlx-community/Kokoro-82M-bf16`. Synthesis verified working via direct curl (HTTP 200, valid mp3/wav for all formats/voices tested).
- Tried both default format and `responseFormat: mp3` in the provider config — no change.

## Symptom detail

- Text reply arrives normally; the auto-TTS supplement (media-only voice message) arrives as a broken/unavailable attachment on the recipient's phone. Signal shows "Original message not found".
- Gateway log shows the signal-cli send for the supplement **starting but never completing**: e.g.

```
2026-08-29T12:12:10.343-05:00 [signal] delivered reply to +1XXXXXXXXXX   <- text reply, delivered
2026-08-29T12:13:22.848-05:00 [signal] signal-cli: WARN ManagerImpl - No profile name set...
(no "delivered reply" line for the supplement send — it hangs)
```

Before a gateway restart, both sends logged "delivered" per reply (two send/delivered pairs), but the recipient still saw the broken attachment. After restart, the supplement send frequently doesn't complete at all.

## What works (verified repeatedly)

Sending the same local audio file via CLI delivers playable audio every time, including:

1. `openclaw message send --channel signal --target <num> --media file.mp3 -m "text"` — ✅ plays
2. Media-only send (no text, mimicking the supplement) — ✅ plays
3. `--media` + `--reply-to <id>` (attachment + quote) — ✅ plays
4. `--delivery '{"audioAsVoice":true}'` — ✅ plays
5. wav, mp3, and m4a all deliver fine through the CLI path (one early wav test failed, but subsequent wav sends via CLI succeeded, so format is not the cause)

[...]

Issue #42539, opened 2026-03-10
Feature: TTS delivery mode option — separate text + voice messages on Telegram


holisticseo:
## Problem

When TTS is enabled on Telegram, the voice note and text are delivered as a single message (voice with text caption). Telegram renders voice note captions in a narrow bubble alongside the audio player, making longer text responses difficult to read.

## Current behavior

- Voice note + text caption delivered as one message
- Text renders in a narrow column next to the audio waveform
- This makes TTS impractical for anything beyond very short responses

## Workaround attempted

Using tagged mode (`auto: "tagged"`) and manually triggering TTS as a separate call works — text arrives full-width, then voice follows as a separate bubble. But this requires the agent to manage the split manually every time, which adds overhead and complexity.

There's also the existing `Telegram/Voice caption overflow fallback` which sends text separately when caption length is exceeded — but only on overflow, not as a configurable default.

## Proposed solution

Add a delivery mode config option, e.g.:

```json
{
  "messages": {
    "tts": {
      "delivery": "separate"
    }
  }
}
```

Options:
- `"bundled"` (current default) — voice + text caption in one message
- `"separate"` — send text as a normal message first, then voice note as a second message

This would give users the best of both worlds: full-width readable text AND a voice message, without requiring agent-side workarounds.

## Context

- Telegram is one of the most popular channels for OpenClaw
- This affects all TTS providers (OpenAI, ElevenLabs, Edge, sherpa-onnx, etc.)
- The narrow caption rendering is a Telegram UI constraint that can't be changed on their end
- Many users want TTS for accessibility/convenience but find the current UX unreadable for longer responses

PR #142190, opened 2026-09-08
fix(tts): speak an answer that carries a picture instead of dropping its voice

### LINKED ISSUE #142188: Auto-TTS is skipped entirely when a reply carries an attachment, so the answer arrives silent
## Summary

Auto-TTS is skipped for the whole reply when that reply carries an attachment, so an
answer with a picture arrives silent — no audio, no error, and no synthesis attempt.

## Steps to reproduce

1. Enable auto-TTS on a channel that supports voice notes (`tts.auto: "always"`,
   `messages.tts.mode: "final"`).
2. Ask the agent for something it answers with text **and** an image attachment in the
   same reply (for example a generated chart, or a picture with a caption).
3. Compare with the same answer text when no attachment is present.

Reduced repro without any channel, against the real synthesis path
(`maybeApplyTtsToPayload`, synthetic speech provider):

- `{ text: SPOKEN }` → audio is produced.
- `{ text: SPOKEN, mediaUrl: "…/chart.png" }` → payload is returned untouched and the
  provider is not called at all. Same for `.jpg`, `.pdf`, `.mp4`, several attachments,
  and a legacy `MEDIA:` line in the text.

## Expected behavior

The answer is spoken. The same text without an attachment is spoken, so the attachment
is the only difference.

## Actual behavior

No audio at all. `maybeApplyTtsToPayloadCore` returns before synthesis:

```ts
if (reply.hasMedia || hasLegacyFinalMediaDirective(text)) {
  return nextPayload;
}
```

Nothing is logged at the default level, `doctor` reports nothing, and the reply looks
normal apart from the missing voice.

Note the guard is not a mistake: on success the same function writes the audio into
`mediaUrl`, so removing the guard would overwrite the attachment. One message carries
one media set — `ttsSupplement` only holds `spokenText` and a flag, not a second set.
Speaking such an answer needs a second message.

## OpenClaw version

2026.9.2. The same line is present on current `main`.

## Operating system

macOS 15.6 (Darwin 25.6.0), Apple Silicon.

## Install method

npm global install, gateway run as a launchd service.

## Model

claude-opus-5 through the claude-cli backend. Speech provider: a local CLI synthesizer.

## Logs

Redacted; peer ids are hashed by the channel's diagnostics.

Reply with a picture — two attachments uploaded, no synthesis anywhere in the turn:

```
11:35:52  deliver kind=block len=1001 media=false
11:36:01  deliver kind=final len=758  media=true
11:36:02  vk upload ok kind=photo
11:36:03  vk upload ok kind=photo
```

The same install with the fix applied — the picture keeps its message and the voice
follows as a second one:

```
15:43:54  deliver kind=final len=621 media=true
15:43:55  vk upload ok kind=photo mime=image/jpeg bytes=210517
15:43:57  tts synthesize: format=ogg, size=385395
15:43:57  deliver kind=final len=0 media=true
15:43:59  vk upload ok kind=audio mime=audio/ogg bytes=402325
```

## Impact and severity

A user-facing feature disappears silently for a whole class of replies. Voice users get
no answer they can listen to whenever the agent illustrates it, and nothing tells them
why. Workaround is to instruct the agent to send pictures in a separate message, which
depends on the agent following instructions every time.

## Additional information

Not a regression: the guard has been there for as long as we have run this install.
Docs describe the current behavior in `docs/tools/tts.md` ("Skips TTS if the reply
already contains structured media"), so this is arguably a deliberate limitation — but
the outcome is a silent loss of speech, and a separate voice message avoids it without
touching the attachment.

A fix with tests is ready and will be linked from this issue.

Shagrat2:
Closes #142188

> **Rebased onto current `main` (2026-09-09).** The branch had gone stale by 585 commits
> and stopped being mergeable. One conflict, in documentation: `main` split
> `docs/tools/tts.md` into seven pages, so this PR's paragraph and decision tree now live
> in `docs/tools/tts/output.md`, next to the rule they amend. Source files merged cleanly.
> Re-verified on the new base: `scripts/check-changed.mjs` green, and the `auto-reply-reply`
> suite passes 5,649 tests across 226 files.

## What Problem This Solves

Fixes an issue where users with auto-TTS enabled would receive no voice message at all
when the assistant's answer carried an attachment. A reply that says something *and*
shows a chart arrives silent: no audio, no error, nothing in `doctor`. The same text
without an attachment is spoken normally, so the feature disappears exactly when the
assistant illustrates its answer.

## Why This Change Was Made

`maybeApplyTtsToPayloadCore` skips synthesis whenever the payload already carries media.
That guard is correct and stays: on success the same function writes the audio into
`mediaUrl`, so removing it would overwrite the picture. One message carries one media
set — `ttsSupplement` holds only `spokenText` and a flag, not a second set.

So the answer is spoken by a **second message** instead. A final that carries non-audio
media is delivered untouched; afterwards its text is synthesized separately and sent as
an audio-only payload marked
`markReplyPayloadAsTtsSupplement(…, { visibleTextAlreadyDelivered: true })`. This is the
mechanism `dispatch-from-config.finalize.ts` already uses for deferred final text; only
the entry condition is new.

Audio versus non-audio is decided by `hasNonAudioMediaReference`, built on the shared
media table (`isAudioFileName`), so a reply that already carries a voice note gets no
second one. Legacy `MEDIA:` directives participate in that classification and are
stripped from the spoken copy while staying in the visible message. Only speech-related
metadata (`tts`, `ttsExplicit`, `commandReply`) is inherited by the synthesized payload,
so command replies stay silent and delivery ownership is not transferred.

Boundaries: the ACP delivery coordinator has its own synthesis application and is not
touched here; the same branch can be added there if maintainers want it.

**Note for channel authors:** this makes a case reachable that previously only appeared
on channels with captioned-final delivery — a final that carries media and no text.
Such a payload is marked `visibleTextAlreadyDelivered` and must not be read as "no answer
was produced". Our own VK channel plugin did read it that way and deleted the message
holding the answer; the flag is what tells them apart.

## User Impact

An answer that comes with a picture is now also spoken: the picture arrives in its own
message, the voice follows in the next one. Answers that already carry audio, command
replies, and replies without attachments behave exactly as before.

## Evidence

**The guard is unconditional — real synthesis path, no channel, no TTS mocks**
(`src/tts/tts-entry-delivery.test.ts`, synthetic speech provider):

| payload | synthesis called |
|---|---|
| text + PNG / JPEG / PDF / MP4 / Opus | no |
| text + several attachments | no |
| text + legacy `MEDIA:` line | no |
| **same text, no attachment** | **yes** — control |

**Delivery, through `dispatchReplyFromConfig`** — 19 attachment formats plus edge cases:
mixed attachments, audio-only sets, a caption-less picture, `isError` payloads, non-final
TTS modes, two finals in one turn, a streamed answer ending in a picture, command replies
with and without an explicit speech request, and synthesis failure. Every case first
asserts the visible message survives untouched.

**The regression test fails without the fix:** stashing only `finalize.ts` turns
"speaks a picture answer as a separate voice message" red while the rest stay green.

**Writer authority is preserved and enforced.** A settled final can carry
`sessionWriterDeliveryAuthority`, and an absent authority reads as authorized at every
check, so the supplement inherits it — while transcript and completion ownership stay
with the visible message. Three dispatch tests pin the boundary:

| case | result |
|---|---|
| writer still holds the turn | picture, then its voice supplement |
| writer replaced while the supplement is being synthesized | picture only — the audio is rejected before transport |
| replaced writer, ordinary settled final (pre-existing test) | not dispatched, unchanged |

The middle case is the one this PR had to earn: the first attempt attached the authority
to the synthesis input instead of the delivered payload, and since that metadata lives by
object identity it never reached the send path. The test failed until the authority moved
onto the payload that is actually sent.

**Speech comes from the prepared text.** `sendFinalPayload` now reports the text it
prepared for delivery — dispatcher normalization applied, deferred block text merged, TTS
not yet run — and the supplement synthesizes that. A reply whose text still carries a
`NO_REPLY` token is stripped before delivery, and the recording no longer contains it; a
recording cannot be normalized after the fact.

**A supplement follows only a message the recipient can see.** Eligibility is decided by
the outcome of preparation and delivery, not by the raw reply:

| situation | supplement |
|---|---|
| final delivered (queued or routed, settled) | yes |
| final deduped against a block that was delivered | yes |
| block vetoed by the channel (`channel-transform`) | no |
| visible answer failed to route | no |
| preparation cleared the caption (`NO_REPLY`, channel transform) | no — an empty prepared text says "say nothing" |
| queued answer cancelled in `beforeDeliver` | no |
| queued answer suppressed as a channel transform | no |
| answer delivered but not visibly (`visibleReplySent: false`) | no |
| settlement could not complete within its bound | no |
| heartbeat run | no — those finals are sent with `skipTts` on purpose |

Admission is not visibility, so eligibility is decided after `turnLedger.settleQueued`,
the same bounded wait the no-visible-reply gate uses — but this gate fails closed. The
neighbouring gate treats admission as the strongest fact because uncertainty there must
not produce a duplicate message; here uncertainty must not produce a voice note that
arrives alone. Both answer "stay quiet"; only the reasoning differs, and the code says so.

**Runs** (Node 24.20.0):

| check | result |
|---|---|
| `pnpm check:changed` | pass |
| `auto-reply-reply` shard | 223 files, 5474 tests, 0 failures |
| `src/tts/tts-entry-delivery.test.ts` | 11 tests, 0 failures |
| `src/media/audio.test.ts` | 13 tests, 0 failures |
| `pnpm build`, `pnpm check` | pass |
| `autoreview` (claude, P2) | no accepted findings |

**Live traffic**, one install, before and after (redacted, hashed peer ids):

```
before   11:36:01  deliver kind=final len=758 media=true
         11:36:02  vk upload ok kind=photo
         11:36:03  vk upload ok kind=photo          ← no synthesis in the turn

after    15:43:54  deliver kind=final len=621 media=true
         15:43:55  vk upload ok kind=photo
         15:43:57  tts synthesize: format=ogg, size=385395
         15:43:57  deliver kind=final len=0 media=true
         15:43:59  vk upload ok kind=audio          ← voice as its own message
```

Docs updated: `docs/tools/tts.md` described the old behavior in both the rule list and
the decision tree.

## Plugin compatibility

Channel plugins see a new message shape: a final that carries media and no text. It is
marked `ttsSupplement.visibleTextAlreadyDelivered`, which is exactly the signal that tells
it apart from "no answer was produced".

Our own VK plugin got this wrong at first and deleted the message holding the answer,
because its progress-draft finalizer read an empty final as "no answer came". The repair
was to check the flag — not to work around the core — and it is a five-line guard plus
seven regression tests in that plugin. We report it because it is the upgrade risk a
reviewer should weigh: plugins that infer "no answer" from empty final text need that
one check. Plugins that only deliver what they are given need nothing.

## Open question for maintainers: durable recovery of the supplement

This is the one finding we have not implemented, because both ways to close it change a
contract that is yours to set. Everything below is what we traced; we are happy to build
whichever you choose.

**The gap.** `buildRecoveryDeliverParams` installs the writer fence
(`assertDirectAdapterHandoff` / `onPlatformSendDispatch`) only when the queued entry has a
`pending-final` `deliveryCompletion`, and reads the authority from that record alone. The
supplement carries its authority in payload metadata, which is process-local, and has no
completion of its own. If the process stops after the supplement is enqueued and the
session writer is replaced before recovery runs, the recovered audio sends unfenced.

**Why the supplement has no completion.** Reusing the visible answer's receipt would give
a second message the first one's delivery ownership — the mistake this PR was already
corrected for in an earlier round. So the choice is a new receipt or no recovery at all.

**Option 1 — give the supplement its own `pending-final` completion.** This is more than a
field: a completion is a durable record with a lifecycle (`prepared` → `queued` →
`delivered` / `suppressed` / `unknown`, settled through `settlePendingFinalDelivery`).
Minting one per supplement means new rows in that store and a new owner obligation to
settle them on every path, including failure. `AGENTS.md` asks contributors to agree
material persistent-store changes before implementing them, which is why we stopped here
rather than guessing at the shape.

**Option 2 — keep the supplement out of durable recovery.** Our preference, and the one
ClawSweeper now recommends ("Approve live-only supplements: add a narrowly scoped
non-recoverable delivery path"). A voice note lost to a restart is a small loss, while one
that outlives its writer is exactly what the fence exists to prevent.

We re-checked whether an existing signal already expresses it, so the ask is not for
something already there. `OutboundDeliveryQueuePolicy` is `required | best_effort`, and
`best_effort` is about how a failure is treated, not about persistence: recovery carries
the policy forward (`delivery-queue-recovery.ts` reads `entry.queuePolicy ?? "best_effort"`)
and replays such entries like any other. There is no admission bypass for an optional
send. So live-only needs a new signal, however small, and that is a contract addition.

**What is already fenced.** The live path: the supplement inherits
`sessionWriterDeliveryAuthority`, it is attached to the delivered payload rather than the
synthesis input, and a replacement during synthesis is covered by a regression test. Only
the recovered path is open.

Tell us which option you want — or a third we have not seen — and we will implement it in
this PR. If you approve option 2, we will add the narrowest signal that expresses "do not
recover this send", thread it from the supplement through admission, and cover it with a
restart test that asserts the recovered queue never replays the audio.

## Related

- #42539 asks for text and voice as separate messages as a general Telegram delivery
  option. Different problem — there the answer is spoken and the split is a preference;
  here synthesis never runs. The mechanism is the same one, so that option could be
  built on this branch.
- #132736 (closed) covers auto-TTS supplement messages on Signal — separate voice
  messages are an existing shape in the core, not a new one.


Shagrat2:
Rebased onto current `main` (the branch had fallen 5015 commits behind) and pushed as `bb83317514`. No conflicts; the behaviour change is unchanged.

The four failing jobs on the previous head were not the patch itself — they were the line-cap ratchet, which landed after this PR was opened. The suite added ~800 lines of tests to `dispatch-from-config.delivery-and-tts.test-utils.ts`, and that file is already over its grandfathered cap.

What changed to satisfy it, without removing coverage:

- The supplement tests now live in two sibling modules, split by subject: `dispatch-from-config.media-tts-supplement.test-utils.ts` (which media is spoken at all) and `dispatch-from-config.media-tts-supplement-delivery.test-utils.ts` (writer authority, what the voice repeats, and the silences a failed or invisible delivery must keep). Both are under the cap and are pulled in by the same `dispatch-from-config.delivery.test.ts` entrypoint.
- The predicate the mocked synthesizer uses to mirror the core guards moved into `dispatch-from-config.tts-guard.test-support.ts`, along with the small dispatch helper both halves need. That left `dispatch-from-config.shared.test-harness.ts` shorter than its baseline rather than longer, so the harness no longer grows either.

Verification on the rebased head: `pnpm exec vitest run src/auto-reply/reply/dispatch-from-config.delivery.test.ts` — 354 tests; `src/media/audio.test.ts` and `src/tts/tts-entry-delivery.test.ts` — 24 tests; `pnpm check:changed` green, including the line-cap ratchet, the dead-export scan, core typecheck and core test typecheck.

The open question from the last review is unchanged and still yours to settle: the supplement is live-only and is not restored, because nothing in the delivery contract marks a record as "do not restore" — `best_effort` describes failure handling, not retention.

Shagrat2:
The two remaining red checks on this head are not from this PR: `checks-node-compact-small-37` fails on `tsgo-core-test-shards.test.ts` with `messaging: 701 test roots exceeds the 700 limit`, and `openclaw/ci-gate` follows it.

That count comes from `main`: `origin/main` itself has 701 `*.test.ts` roots matching `test/tsconfig/tsconfig.core.test.messaging.json`, and this PR adds none — the new modules it introduces are `*.test-utils.ts`, which the shard config does not count. Tracked in #151577, fix in #151584.

The rebase work in the previous comment is unaffected: the line-cap ratchet, the dead-export scan, core typecheck and the delivery suite are green locally on this head.

Shagrat2:
Rebased onto current `main` (1426 commits behind) and pushed as `ca5cb95448f`. The behaviour is unchanged; the red checks on the previous head were `main`'s own shard count, fixed since by #151564.

The rebase touched only imports: `dispatch-from-config.finalize.ts` now takes `resolveOutboundMediaUrls` alongside `hasOutboundReplyContent` and `main`'s new `resolveReplyCompletion`, and the shared test harness dropped three type imports that `main` no longer uses there (`ReplyDispatchBeforeDeliver`, `ReplyDispatchKind`, `ReplyDispatchSettledCounts`) — the typecheck caught them, and the harness keeps `shouldSynthesizeFinalAudio` from the extracted guard support module.

### Verification

`node scripts/check-changed.mjs` passes end to end on this head, including the line-cap and suppression ratchets, the assertion SAFETY ratchet, plugin boundaries, the tsgo graph boundary, core typecheck and the config-doc baseline. `src/auto-reply/reply/dispatch-from-config.delivery.test.ts` — 359 tests pass.

Still running in production here: an answer that carries a picture arrives as the picture, then the voice as a separate message, with the guard that suppresses auto-TTS for media left intact.

The open question from the previous round is unchanged and is yours to decide: whether the durable restoration of that guard belongs in this PR or in a follow-up.

Shagrat2:
Note on the red checks on this head — they are `main`'s, not this branch's.

`check-test-types-core-2` fails on `src/state/openclaw-state-db-task-identifiers.test.ts(6,3): error TS2459` about `subagent-registry.store.sqlite.js`. This PR does not touch `src/state` or `src/agents/subagents`; #154052, an unrelated PR opened after mine, fails the same job the same way. The compact-node shards that went red are the tooling/stylelint lane case, also untouched here.

`node scripts/check-changed.mjs` passes end to end locally on this head, and `dispatch-from-config.delivery.test.ts` is 359/359.

Shagrat2:
Fixed the P1 in `4effeb43bf0`, taking the recommended contract: the supplement is now **live-only**.

### Why live-only rather than its own durable completion

The two options are not symmetric here. A `pending-final` completion is not a free-standing object: it is keyed by `deliveryId`/`intentId` against the session's pending final delivery, and that record belongs to the **visible answer**. Giving the supplement one would mean a second owner of the same session's pending final — a change to session delivery ownership for the sake of optional audio.

Live-only costs exactly one thing, and it is the right thing to lose: if the Gateway restarts between the picture and its voice, the audio is gone. The answer it repeats has already been delivered and is already in the transcript, so nothing the user needs disappears — while the authority hole does.

### What changed

`skipQueue` now reaches the routed reply path (`RouteReplyParams` → `routeReplyOperationToOriginating` → `sendFinalPayload`), and the media-final supplement sets it. `skipQueue` is the existing core switch for a send that owns no write-ahead row — task progress updates already use it — so no new durability concept enters the contract.

Consequences, stated plainly:

- no durable row is written for the supplement, so `buildRecoveryDeliverParams` has nothing to replay and cannot send past a replaced writer;
- the live writer fence is untouched: `isSessionWriterDeliveryAuthorized` still runs before the send, and the existing regression for a replaced writer still passes;
- the deferred-block supplement that `main` already had is deliberately **not** touched, even though it sets the same `visibleTextAlreadyDelivered` flag. It does not inherit `sessionWriterDeliveryAuthority`, so it has no fence to lose, and changing its durability is not this PR's business.

Regression: on the routed path the visible answer is sent with durable custody (`skipQueue` unset) and the voice supplement with `skipQueue: true`. `docs/tools/tts/output.md` now says the voice is live-only and can be lost on restart.

### Verification

`node scripts/check-changed.mjs` passes end to end on this head (including the line-cap ratchet — `dispatch-from-config.choose-route.ts` sat exactly at the cap, so the two identical "nothing left this lane" returns I had expanded are one helper now). `dispatch-from-config.delivery.test.ts`: 360 tests, including the new one.

### Still yours to decide

Two items on your list are decisions, not defects, and I do not want to presume either:

1. **Supplement lifetime.** I implemented live-only because it is what the review recommends and what the mechanism supports without moving session ownership. If you would rather have durable, independently fenced audio, say so and I will build that instead — it is a larger change and I would want your shape for it before writing it.
2. **Channel-plugin compatibility.** An audio-only final is newly reachable, and our own VK fork mishandled it at first: its step-progress draft treated a final without text as "no answer" and deleted the visible message. We fixed that in the fork by checking the `ttsSupplement.visibleTextAlreadyDelivered` flag the core already sets. That flag is the compatibility contract for channels; if you want it documented as such in the plugin docs, I will write that up in this PR.

Shagrat2:
You are right, and the fix was half a fix. `b6769cee465` completes it.

### What I got wrong

I threaded `skipQueue` as a call option down the routed path only. The channel-local path does not carry call options at all: `sendFinalPayload` hands the payload to `turnLedger.sendQueued("final", …)` → the dispatcher → `deliverInboundReplyWithMessageSendContextCore`, and the policy never reached the queue. Same hole, other lane.

### What it is now

The policy travels **on the payload**, because that is the only thing both lanes share: `ttsSupplement.liveOnly`, set by the finalizer on the supplement it builds. `sendMessageBatch` honors it once, at the single point both lanes pass through:

```ts
const liveOnlyPayload =
  pendingFinalCompletion === undefined &&
  params.payloads.some((payload) => payload.ttsSupplement?.liveOnly === true);
```

It sits right beside the `ephemeralWriterAuthorities` computation it belongs to: that array is exactly the fence that exists only in memory, and the flag says the payload holding it must not be persisted for replay. The guard on `pendingFinalCompletion` keeps a payload that *does* own durable completion authority durable.

The earlier option plumbing through `RouteReplyParams` / `routeReplyOperationToOriginating` / `sendFinalPayload` is reverted — one mechanism instead of two, and `route-reply.ts` is untouched again.

The pre-existing deferred-block supplement is still not affected: it does not set the new flag, and I am not changing its lifetime in this PR.

### Regressions at the boundary you named

In `src/channels/message/send.test.ts`, against the durable sender itself rather than a mock of routing:

- a payload with `ttsSupplement.liveOnly` reaches `deliverOutboundPayloads` with `skipQueue: true`;
- an ordinary answer with media reaches it with `skipQueue` unset.

The dispatch-level test now asserts the supplement payload carries the flag (and the visible answer does not), which is what the channel path consumes.

### Verification

`node scripts/check-changed.mjs` passes end to end, including the core and core-test typechecks and the line-cap ratchet. `dispatch-from-config.delivery.test.ts` 360 tests, `channels/message/send.test.ts` 23.

One housekeeping note: I briefly reformatted four files with a stray `npx prettier` before noticing the repository uses `oxfmt`. That is reverted — `pnpm format` output only, and the diff is my change alone.

### Still the maintainer's call

The two decisions from your list are unchanged and I have not presumed either: the supplement's lifetime (live-only versus an independently fenced durable one, which would need its own owner rather than the session's pending final), and whether the `ttsSupplement.visibleTextAlreadyDelivered` flag should be documented as the compatibility contract for channel plugins — our VK fork needed exactly that flag to stop treating an audio-only final as "no answer".

Shagrat2:
@clawsweeper re-review — the last review landed on `4effeb43bf0`; the same-channel propagation fix is `b6769cee465` (policy on the payload as `ttsSupplement.liveOnly`, honored once in `sendMessageBatch`, with regressions against the durable sender itself).

Shagrat2:
Fixed in `275c8092fec`. The finding is correct, and it is the same mistake as last round in a different place: I checked my own code instead of the whole path the payload travels.

`getReplyPayloadTtsSupplement` rebuilds the supplement field by field, so it dropped `liveOnly` — and `buildTtsSupplementMediaPayload` builds the delivered payload from that normalized object. Any path through the public helper therefore handed delivery a payload with no lifetime on it, and the queue boundary saw nothing to honor. The flag survived only as long as nothing normalized it.

The normalizer now carries it, with the reason written next to it: delivery reads the lifetime from the normalized supplement, so dropping it silently restores durable custody for audio whose writer fence recovery cannot rebuild.

Regressions in `src/auto-reply/reply-payload.test.ts`, on the helper pair rather than on a caller: a live-only supplement keeps the flag through both `getReplyPayloadTtsSupplement` and `buildTtsSupplementMediaPayload`; an ordinary supplement still has no flag. The existing dispatch-level assertion on the delivered supplement now includes it, which is how the regression surfaced locally.

### Verification

`node scripts/check-changed.mjs` passes end to end. `dispatch-from-config.delivery.test.ts` 360, `channels/message/send.test.ts` 23, `auto-reply/reply-payload.test.ts` 17.

The two open decisions are unchanged and still yours: the supplement's lifetime, and whether `ttsSupplement.visibleTextAlreadyDelivered` should be documented as the compatibility contract for channel plugins that treat an audio-only final as an empty answer.

Shagrat2:
Authority proof on a live installation, plus the layer fix the proof itself uncovered. Head `c6ca224bb79`.

### The policy moved to the queue boundary

Writing the proof showed my previous placement was wrong. I honored `ttsSupplement.liveOnly` in `sendMessageBatch`, which every *reply* lane passes through — but a caller reaching `deliverOutboundPayloads` directly bypassed it, and the integration test caught exactly that. The check now sits where `skipQueue` is consumed, in `deliver-queue.ts`:

```ts
const liveOnlyPayload =
  params.deliveryCompletion === undefined &&
  params.payloads.some((payload) => payload.ttsSupplement?.liveOnly === true);
const queued = params.skipQueue || liveOnlyPayload || …
```

One place, layer-independent, and `channels/message/send.ts` is untouched again.

### Non-replay, against the real queue

`src/infra/outbound/deliver.live-only-supplement.integration.test.ts` runs the real delivery path with a real queue state directory and the real plugin registry; only the channel adapter is a stub:

| payload | durable delivery intent |
|---|---|
| live-only voice supplement (carrying an in-memory writer authority) | none created |
| the answer it repeats | created |

The intent is the honest signal: a completed row is acked away, so asserting on leftover rows would pass for both.

### Replaced writer, before transport

`src/channels/message/send.test.ts` exercises the real guard at the hook this layer attaches, with the stub transport calling it exactly where the platform send begins. The authority points at a session store that holds no such session — the writer is gone, as after a replacement — and the hook rejects with `Session writer changed before final reply delivery`.

### Live installation, current head

Production install on OpenClaw 2026.9.5 with this change applied to `dist`, VK channel, auto-TTS in `final` mode. One user turn asking for a picture from a local gallery; identifiers and paths redacted:

```
12:08:43  message received: channel=vk … source=dispatchInbound
12:09:08  vk: step-progress draft sent  msgId=10078 len=171
12:09:08  tts-local-cli synthesize: text=«Нашла — голубое платье …»
12:09:09  vk: step-progress draft edited INTO final  msgId=10078 len=192
12:09:10  tts-local-cli synthesize: format=ogg, size=157010
          vk-diag  upload ok  kind=photo  mime=image/jpeg  bytes=360331
          vk-diag  media sent  messageId=10079
          vk-diag  deliver  kind=final  len=0  media=true
          vk-diag  upload ok  kind=audio  mime=audio/ogg  bytes=165692
          vk-diag  media sent  messageId=10080
```

Three messages, in order: the answer text, its picture, then the voice as its own message — the audio-only final with `len=0` is the supplement. The delivery queue database held no row for any of it afterwards, and its only rows are five unrelated failures from May.

The writer was current throughout, so this is the authorized half of the chain; the rejection half is the guard test above, and the non-replay half is the queue test. I cannot produce Telegram Test Server proof: that skill needs Convex-leased team credentials that belong to the project, not to a contributor.

### PR body

Refreshed to describe the current implementation — the payload-level lifetime, where it is honored, and what a restart costs — instead of the earlier routed-only variant.

### Verification

`node scripts/check-changed.mjs` passes end to end, including both typechecks and the line-cap ratchet. `deliver.live-only-supplement.integration.test.ts` 2, `dispatch-from-config.delivery.test.ts` 360, `channels/message/send.test.ts` 22, `reply-payload.test.ts` 17, `deliver.test.ts` 253.

Shagrat2:
Both items fixed in `6b96c7723ff`.

### [P2] Mixed batches — rejected, not silently downgraded

You are right, and you named the trap in the same breath: `some` strips durable custody from the ordinary reply, `every` would persist the unfenced supplement. Neither is a lifetime this boundary can honor, so a mixed batch is now refused before any I/O:

```ts
if (liveOnlyCount > 0 && liveOnlyCount !== params.payloads.length) {
  throw new Error(
    "Outbound batch mixes live-only supplemental audio with durable payloads; send them separately.",
  );
}
```

One batch, one lifetime. The finalizer already sends the supplement as its own final, so this is a caller mistake rather than a delivery choice, and failing closed keeps both payloads out of transport instead of quietly giving one of them the wrong guarantee.

Regression in the queue integration test: `[ordinaryAnswer, liveOnlySupplement]` rejects, the channel adapter is never called, and no delivery intent is created.

### [medium] Replaced-writer proof no longer stops at a mock

Replaced with the production send path — `sendDurableMessageBatchCore` over the real `deliverOutboundPayloads`, a real queue state directory and the real plugin registry, with only the channel adapter stubbed as the thing standing in for the network:

```
result.status === "failed"
result.error  → "Session writer changed before final reply delivery
                 | OPENCLAW_PLATFORM_MESSAGE_NOT_DISPATCHED"
adapter       → never called
```

The failure carries the platform-not-dispatched marker, so the stop is before transport, not after. The earlier `send.test.ts` case that consumed the rejection inside a mocked outbound callback is deleted rather than kept alongside — it implied coverage it did not have.

What this still is: a test, not a live capture. I cannot stage a writer replacement mid-turn on the production install without fabricating it, and the Test Server path needs project-owned credentials. The live evidence stays what it is — the authorized half, from a real VK turn.

### PR body

The obsolete recovery section is gone; the body now describes the payload-level lifetime, where it is honored, what a restart costs, and which decision is still open.

### Verification

`node scripts/check-changed.mjs` passes end to end, both typechecks included. `deliver.live-only-supplement.integration.test.ts` 4, `dispatch-from-config.delivery.test.ts` 360, `channels/message/send.test.ts` 21, `reply-payload.test.ts` 17, `deliver.test.ts` 253.

Shagrat2:
The two red shards on `6b96c7723ff` are timeouts in areas this branch does not touch. Details, so you can judge rather than take my word:

**`checks-node-compact-large-20`** — `test/scripts/bench-session-history.test.ts`:

```
AssertionError: expected SystemError: spawnSync … node ETIMEDOUT to be undefined
  errno: -110, code: "ETIMEDOUT"
  spawnargs: ["--import","./scripts/tsx.mjs","scripts/bench-session-history.ts", …]
```

The benchmark subprocess did not finish inside its 30s budget.

**`checks-node-compact-large-8`** — `src/node-host/node-worker-supervisor.recovery.test.ts:445`:

```
await vi.waitFor(() => expect(closed).toBe(true), { timeout: 1… })
AssertionError: expected false to be true
```

A worker-lifecycle wait expired.

The branch's 18 files are `src/auto-reply/**`, `src/infra/outbound/**`, `src/channels/message/send.ts`, `src/media/audio.ts`, `src/shared/reply-payload.types.ts`, `src/tts/tts-entry-delivery.test.ts` and one doc — nothing in `src/node-host/**` or `scripts/bench-session-history.ts`, and nothing that changes process spawning or worker supervision. `node scripts/check-changed.mjs` selects lanes from the changed files and passes end to end locally on this head.

I cannot re-run the jobs — rerun needs admin rights on the repository. If you re-run those two shards and either stays red, I will treat it as mine and dig in; both are timing-dependent, so I would expect them green.

Shagrat2:
Second round on `4ad7239e23e` (same tree, fresh sha): 172 checks green, two red again — but **different tests**, and neither is in this branch's files.

| round | shard | test that failed |
|---|---|---|
| 1 | `checks-node-compact-large-20` | `test/scripts/bench-session-history.test.ts` — `spawnSync … node ETIMEDOUT` |
| 1 | `checks-node-compact-large-8` | `src/node-host/node-worker-supervisor.recovery.test.ts:445` — `vi.waitFor` expired |
| 2 | `checks-node-compact-large-31` | `src/agents/main-session-recovery/main-session-restart-recovery.test.ts` — a spy expected twice, called zero times |
| 2 | `checks-node-compact-large-28` | `src/gateway/server-methods/models-list.native-lifecycle.integration…` — `expected undefined to be true` |

Four different tests across four areas this branch does not touch, and they do not repeat between rounds.

What they have in common is time. `main-session-restart-recovery.test.ts` runs 219 tests and takes **387 s** on my machine in one file alone; the pair of it and my new integration file took 190 s in the same session — the variance between runs of the same work is larger than anything my change adds. `bench-session-history` fails by exhausting a 30 s subprocess budget. Under a loaded hosted runner these land on the wrong side of their limits.

For calibration, the file this PR adds costs 5.4 s for 4 tests, and its slowest case — the durable-custody control that really writes a queue row — is 1.7 s. It passes locally, as do all four tests that CI reported, including the two from this round.

Not just this PR either: the three other open PRs I sampled are red on `ui/src/pages/chat/components/chat-position-rail.test.ts`.

I have no rerun rights. If you re-run these two shards and something red survives, or if you would rather I trim the 1.7 s control case to lighten that shard, say so and I will do it.

PR #142192, opened 2026-09-08
fix(plugins): name the changed facet in plugin registry differences

steipete:
## What Problem This Solves

A plugin registry difference is computed over the whole record, its install record and its per-plugin diagnostics, but the difference object only carried the persisted and derived *source paths*. Whenever the record or diagnostics changed while the source stayed put, both surfaces printed a self-contradicting line: the Gateway refused to start with `could not verify the persisted replacement (reread source was derived; differences: longcat (persisted source: "…/extensions/longcat/index.ts"; derived source: "…/extensions/longcat/index.ts"); diagnostics: persisted-registry-stale-source)` (observed live while a plugin package was being rebuilt), and `openclaw plugins registry` printed `longcat: persisted …/index.ts; derived …/index.ts`. Neither tells the operator what actually moved.

## Why This Change Was Made

`PluginRegistryDifference` now carries a non-empty `changed` list naming the differing facets (`record`, `install`, `diagnostics`), computed where the comparison already runs (`diffPluginRegistryRecords`). Both renderers print it: the startup refusal reads `longcat (record changed; persisted source: …; derived source: …)` and the CLI reads `longcat: record changed; persisted …; derived …`. The `plugins registry --json` output gains the same field additively.

## User Impact

Operators hitting a stale persisted registry see which part of the plugin changed instead of two identical paths. No behavior change beyond the message and the additive JSON field.

## Evidence

- Live: an isolated Gateway start on a built `main` dist failed with the identical-sources message while `pnpm test:extensions` rebuilt a plugin in the same checkout; the retry started cleanly.
- New case in `src/plugins/plugin-registry-inspection.test.ts` calls the comparison owner with identical sources and a changed install record plus diagnostic: fails on `main` (no `changed` field), passes with the fix (`["install", "diagnostics"]`); identical inputs yield no difference.
- The `plugins registry` report case in `src/cli/plugins-cli.list.test.ts` now uses identical sources and asserts `demo: install+diagnostics changed; persisted /plugins/demo/index.js; derived /plugins/demo/index.js`; the startup-refusal regression in `doctor-config-preflight.plugin-persistence.test.ts` now requires the facet.
- `node scripts/run-vitest.mjs` over plugin-registry-inspection (10), plugins-cli.list (42), doctor-config-preflight.plugin-persistence (15), status.registry-snapshot (30): pass. `node scripts/check-changed.mjs`: pass. Autoreview (Codex): scoped-clean.

Production LOC delta: +21 (facet computation replaces the single deep-equal; two format strings; one type).

### DIFF
--- a/src/cli/plugins-cli.list.test.ts
+++ b/src/cli/plugins-cli.list.test.ts
@@ -970,14 +970,16 @@ describe("plugins cli list", () => {
   });
 
   it("reports persisted plugin registry state without refreshing", async () => {
+    // Identical sources: only the changed facets tell the operator what moved.
     inspectPluginRegistryMock.mockResolvedValue({
       state: "stale",
       refreshReasons: ["stale-manifest"],
       differences: [
         {
           pluginId: "demo",
+          changed: ["install", "diagnostics"],
           persistedSource: "/plugins/demo/index.js",
-          derivedSource: "/plugins/demo/dist/index.js",
+          derivedSource: "/plugins/demo/index.js",
         },
       ],
       persisted: {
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [12627fd28df67e14b25611020240fe5360deebc7](https://github.com/openclaw/openclaw/pull/142192/commits/12627fd28df67e14b25611020240fe5360deebc7)
- Landed commit: [1db32b4f4d677c473eeb983829c74caaaf7dccb7](https://github.com/openclaw/openclaw/commit/1db32b4f4d677c473eeb983829c74caaaf7dccb7)

PR #142193, opened 2026-09-08
fix(mattermost): keep sibling accounts connected during secret-reference reloads

ceckert:
## What Problem This Solves

Fixes an issue where changing one Mattermost account's settings disconnected other accounts when bot tokens used secret references, even when those references resolved successfully at runtime.

## Why This Change Was Made

Reload planning now uses the existing account inspection hook, with strict resolution as the fallback when a plugin has no inspector. Mattermost exposes its existing inspector through the shared runtime/setup adapter. Restart execution, secret activation, and transport authentication retain their strict checks.

## User Impact

Operators can change a named account's non-secret settings without interrupting other working accounts. Global changes still restart the channel. Unavailable credentials retain the existing strict failure and recovery behavior.

## Evidence

Verified with the built Gateway, watched configuration writes, an isolated official Mattermost server, two bot accounts, and file-based secret references. Authentication used real REST responses and matched WebSocket acknowledgments. Connection continuity and replies were observed through the real transport; a deterministic local provider supplied only reply text.

| Scenario | Pinned main | Candidate |
| --- | --- | --- |
| First account-only change | Both connections replaced | Only the changed account replaced; both accounts replied |
| Second account-only change | Not repeated | Same sibling connection retained; both accounts replied |
| Global change | Not repeated | Both connections replaced; both accounts replied |

- Baseline planner regression failed on the unwanted whole-channel target. The setup-adapter regression failed on the absent inspection hook.
- 430 focused tests passed across planning, Mattermost adapters and authentication, secret publication, restart ordering, deferral, and failure handling.
- Missing, empty, unavailable, and invalid credentials did not create authenticated service. Existing valid service was distinguished from new startup. Valid token rotation used normal secret activation and real authentication.
- A held ordinary reply deferred the account reload. The reply completed before connection replacement, and the sibling connection remained unchanged.
- A cold-account negative initially had an incorrect same-sibling-connection expectation. The retained observations show the existing strict executor's whole-channel fallback: the unavailable account remained blocked while the valid sibling reauthenticated and replied. Independent source review confirmed this preserved behavior; the original failed harness result remains recorded.
- Fresh source-blind acceptance passed 13 independently exercised runtime cases and independently checked the raw baseline and cold-account records.
- Native builds, formatting, and focused lint passed. Full hosted checks and the final exact-head review are separate landing gates.

Private raw fixtures and observations retain the complete before/after evidence and failure history. Credentials, account and room details, messages, connection identities, and infrastructure metadata are omitted here.

Preserves @ceckert's original contribution and commit ancestry. AI-assisted repair and validation.

### DIFF
--- a/extensions/mattermost/src/channel-actions-setup-status.contract.test.ts
+++ b/extensions/mattermost/src/channel-actions-setup-status.contract.test.ts
@@ -5,7 +5,7 @@ import {
   installChannelStatusContractSuite,
 } from "openclaw/plugin-sdk/channel-test-helpers";
 import type { OpenClawConfig } from "openclaw/plugin-sdk/config-contracts";
-import { describe, expect } from "vitest";
+import { describe, expect, it } from "vitest";
 import { mattermostPlugin, mattermostSetupPlugin } from "../channel-plugin-api.js";
 
 describe("mattermost actions contract", () => {
@@ -138,5 +138,31 @@ describe("mattermost status contract", () => {
         },
       },
     ],
+  });
+});
+
+describe.each([
+  ["runtime", mattermostPlugin],
+  ["setup", mattermostSetupPlugin],
+] as const)("mattermost %s account inspection", (_name, plugin) => {
+  it("inspects source SecretRefs while strict account resolution rejects them", () => {
+    const cfg = {
+      channels: {
+        mattermost: {
+          baseUrl: "https://chat.example.com",
+          accounts: {
+            alpha: { botToken: { source: "env", provider: "default", id: "BOT_TOKEN" } },
+          },
+        },
+      },
+    } as OpenClawConfig;
+    expect(plugin.config.listAccountIds(cfg)).toContain("alpha");
+    expect(plugin.config.inspectAccount?.(cfg, "alpha")).toMatchObject({
+      accountId: "alpha",
+      configured: true,
+      botTokenStatus: "configured_unavailable",
+      botToken: undefined,
+    });
+    expect(() => plugin.config.resolveAccount(cfg, "alpha")).toThrow();
   });
 });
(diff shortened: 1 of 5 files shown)


ceckert:
@clawsweeper re-review

The PR body now includes after-fix live Gateway proof on unchanged head `7b6a269450a540c271aa58c51df3fe8f52817721`: two account-local reloads with authored file SecretRefs, zero sibling WebSocket closes/re-authentications, unchanged sibling lifecycle timestamps, and successful real Mattermost replies after each reload. A channel-global control reconnected both accounts. Sanitized Gateway logs, connection counts, methodology, and proof limitations are included under Evidence. No code changes were needed.

obviyus:
Thanks @ceckert! Landed in 7fdf62ec4b394821d0bab341c19ee5ae3d500f4d, preserving your authorship.

**Proof:** The built Gateway and real Mattermost transport completed two account-only reloads with both accounts replying and the sibling's authenticated connection unchanged. The global control restarted both accounts.

- **Bug:** Valid secret references caused an account-only settings change to restart the whole channel.
- **Fix:** Planning uses the existing inspection contract; secret activation, restart execution, and authentication remain strict.

| Before | After |
| --- | --- |
| One account change disconnected both accounts | Only the changed account reconnects; both keep working |

430 focused tests and 13 fresh independent runtime checks passed, including credential failures, rotation, and active-work deferral. The original cold-case harness failure is retained and independently explained in the PR evidence.

[Full CI passed](https://github.com/openclaw/openclaw/actions/runs/34352536160). The required exact-head local review passed with no findings or security concerns. The dependency guard also passed after correcting stale comparison metadata; no dependency override or CI exception was used. All five landed files match the tested candidate bytes.

PR #142194, opened 2026-09-08
fix(crabbox): keep lease renewal after temporary heartbeat failures

steipete:
Related: #134931

## What Problem This Solves

Fixes an issue where a placed cloud worker could lose lease renewal after Crabbox reported a temporary claim conflict. OpenClaw incorrectly reported that heartbeat was unavailable and stopped scheduling renewals, allowing coordinator-idle cleanup to reap the worker.

## Why This Change Was Made

Crabbox uses exit code 2 for claim conflicts as well as unsupported commands. Disable renewal only for explicit unsupported-command or unsupported-provider diagnostics; other failures retain the existing schedule. Teardown and process ownership are unchanged. This is an independent fix extracted from #134931 and needs no schema or configuration change.

## User Impact

Temporary heartbeat failures no longer permanently disable lease renewal. The warning reports the actual failure instead of incorrectly asking the operator to upgrade Crabbox.

## Evidence

- The claim-conflict regression failed on current main with the incorrect upgrade warning; all 255 provider tests pass with the correction, including unsupported-command/provider behavior, retry timing, and teardown.
- Live process/HTTP proof used the actual heartbeat manager, process runner, and installed Crabbox 0.51.0 against a synthetic loopback coordinator. A real CLI configuration failure exited 2; after repairing only the fixture config, two ordinary scheduled heartbeats exited 0. Disposal stopped subsequent calls and all children and the server were cleaned up. This proves the CLI/scheduling boundary, not native AWS or claim-conflict reproduction.
- Independent P2 review: no accepted/actionable findings.


steipete:
Merged via squash.

- Prepared head SHA: [b4ffa4b015adefb38419760545129cfa2aa979bf](https://github.com/openclaw/openclaw/pull/142194/commits/b4ffa4b015adefb38419760545129cfa2aa979bf)
- Landed commit: [e321e32abde7b91f3ec1e83f599ea78e98c53673](https://github.com/openclaw/openclaw/commit/e321e32abde7b91f3ec1e83f599ea78e98c53673)

PR #142196, opened 2026-09-08
fix(terminal): preserve blank lines in multiline table cells

steipete:
## What Problem This Solves

Fixes an issue where users reading multiline CLI tables would see text shift onto the wrong sibling line when a cell contained explicit blank lines. For cells containing `alpha\n\nomega` and `1\n2\n3`, the data rows changed from:

```text
alpha | 1
omega | 2
      | 3
```

To:

```text
alpha | 1
      | 2
omega | 3
```

## Why This Change Was Made

Emit explicit hard line breaks at the table wrapping owner, and suppress only an empty final remainder. Remove the blanket empty-line filters and their one-use wrapper. Clear pending CR state when a character is actually buffered so discarded spacing inside an ANSI sequence cannot split an existing CR/LF pair.

## User Impact

Blank lines in descriptions and other multiline table cells retain their positions. ANSI colors, hyperlinks, CRLF, and neighboring columns stay aligned without adding a phantom trailing row.

## Evidence

- Reproduced the shifted rows through the exported `renderTable` function before changing production code.
- Five new regression cases failed on the original code: LF, CR, CRLF, colored CRLF, and linked LF. They cover leading, internal, and trailing explicit blank lines against a numbered sibling column.
- All 91 table/note tests pass, including existing ANSI/OSC-8, separated CR/CSI-tab/LF, large-row, and wrapping checks. The full suite caught the connected pending-CR issue during implementation; its existing assertion was retained and the owner state transition was repaired.
- The owning package test type graph, full-file typed lint, formatting, and diff checks pass. Independent P2 review found no actionable findings.

Production: 13 added / 16 removed, net -3 lines. Tests: +49 lines. No configuration, protocol, dependency, or persistence changes.

### DIFF
--- a/packages/terminal-core/src/table.test.ts
+++ b/packages/terminal-core/src/table.test.ts
@@ -620,6 +620,55 @@ console.log(JSON.stringify({
   );
 
   it.each([
+    ["LF", "\n", "", ""],
+    ["CR", "\r", "", ""],
+    ["CRLF", "\r\n", "", ""],
+    ["colored CRLF", "\r\n", "\x1b[31m", "\x1b[39m"],
+    ["linked LF", "\n", "\x1b]8;;https://example.com/\x07", "\x1b]8;;\x07"],
+  ])(
+    "preserves blank %s lines at their positions across cells",
+    (_label, separator, open, close) => {
+      const out = renderTable({
+        border: "ascii",
+        width: 30,
+        columns: [
+          { key: "A", header: "A", flex: true },
+          { key: "B", header: "B", flex: true },
+        ],
+        rows: [
+          {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3ec18bf11085c24c7adb78f3210e5a9dea80a5bd](https://github.com/openclaw/openclaw/pull/142196/commits/3ec18bf11085c24c7adb78f3210e5a9dea80a5bd)
- Landed commit: [34dd0c607143425aed6d973440d2ef2656369a8b](https://github.com/openclaw/openclaw/commit/34dd0c607143425aed6d973440d2ef2656369a8b)

PR #142197, opened 2026-09-08
fix(llm): retain prepared runtimes through transport cleanup

steipete:
## What Problem This Solves

A direct plugin completion can return after cancellation while its response callback or response-body cancellation is still running. Releasing the prepared runtime at that point drops its last lease, so an overlapping request rebuilds the same generation while accepted work still uses the previous one. Parent shutdown can also finish without joining that work.

## Why This Change Was Made

Keep the existing lease inside the existing AsyncWorkScope until accepted transport callbacks and cancellation promises settle. A narrow transport-host observer records the actual promises before they can be lost behind a cancellation race. The caller receives its result independently; parent cleanup retains and drains the operation before releasing the lease.

This six-file slice follows #141938 and replaces part of #140674. It reuses the existing work-scope implementation and preserves callback deferral, synchronous cancellation invocation, and the original completion error. Physical registration disposal remains a separate change.

## User Impact

Cancelled calls still return before a held cleanup operation finishes. Overlapping calls reuse prepared facts while keeping independent mutable request stores. Once accepted work and the overlapping call finish, the generation can retire normally.

## Evidence

The two new native cases fail on the original code: parent drainage finishes early and the overlapping request creates a second generation. Both pass with the fix, along with all nine existing native ownership cases. The selected native, package transport, and runtime suites pass 124 tests. Full canonical changed checks, a complete build, and independent Codex review pass.

Separate compiled baseline and candidate processes call createPluginRuntime().llm.complete through a registered synthetic provider and real loopback HTTP. Both callback-after-abort and native cancellation cases demonstrate:

| --- | ---: | ---: |
| Generation builds while cancelled work remains pending | 2 | 1 |
| Independent request-store forks | 2 | 2 |
| Caller result arrives before the held work settles | Yes | Yes |
| Later request rebuilds after drainage | Yes | Yes |

The cancellation probe uses the public guarded-fetch and transport-notification helpers with a native Web Streams cancellation operation. It verifies emitted host/transport ownership, not the private cancellation internals of a particular provider SDK. Source and compiled artifacts remain unchanged through the probes; all owned processes, sockets, and synchronization operations are joined. These are structural reuse observations, not a general inference-latency claim.

### DIFF
--- a/packages/ai/src/host.ts
+++ b/packages/ai/src/host.ts
@@ -129,6 +129,8 @@ type AnthropicInlineContentNormalizer = (
 
 /** Narrow host ports consumed by the built-in provider adapters. */
 export interface AiTransportHost {
+  /** Retains accepted lifecycle work after its caller observes cancellation. */
+  observePendingProviderWork?: (pending: Promise<unknown>) => void;
   /**
    * Builds a policy-guarded fetch for one model request.
    * Returning undefined keeps the provider SDK's default fetch.
--- a/packages/ai/src/transports/transport-stream-shared.acceptance.test.ts
+++ b/packages/ai/src/transports/transport-stream-shared.acceptance.test.ts
@@ -1,5 +1,6 @@
 import type { Model, StreamOptions } from "@openclaw/llm-core";
 import { describe, expect, it, vi } from "vitest";
+import { configureAiTransportHost, getAiTransportHost } from "../host.js";
 import {
   copyProviderAcceptanceObserver,
   notifyProviderHttpMetadata,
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [185194f66d6995dfbcc78d9820972c17d29febd4](https://github.com/openclaw/openclaw/pull/142197/commits/185194f66d6995dfbcc78d9820972c17d29febd4)
- Landed commit: [868ac24cb561b026d8abef4a4e5c7dd6c88716d7](https://github.com/openclaw/openclaw/commit/868ac24cb561b026d8abef4a4e5c7dd6c88716d7)

PR #142198, opened 2026-09-08
fix(worker): allow long prepared workspace paths

steipete:
Related: #134931

## What Problem This Solves

Fixes an issue where a cloud worker rejected a deeply nested project path even though project preparation had accepted it. Launch used the 256-character identifier limit for workspace and containment paths.

## Why This Change Was Made

Use the existing 4096-character workspace limit for both launch paths. Preserve absolute POSIX/Windows/UNC validation, whitespace checks, and the permission-mode/containment-root pairing; reject embedded NUL before filesystem operations. Identifier limits and protocol versions are unchanged.

## User Impact

Prepared projects under long directory prefixes can launch and execute commands. This is an independent correction extracted from #134931, with no schema or configuration changes.

## Evidence

- Original source failed the long-path regression and a real child-worker launch before Gateway admission. The corrected descriptor, process protocol, and command suites pass all 61 focused tests.
- Complete changed-file checks and independent P2 review passed.
- A real Node child worker consumed its IPC/JSONL descriptor, resolved a 405-character workspace and 387-character containment root, connected to a synthetic local Gateway, executed a real command in that workspace, committed its successful result, and exited. All owned processes and state were cleaned up. This is local process/transport proof, not a native Windows or cloud-provider test.

### DIFF
--- a/src/worker/launch-descriptor.test.ts
+++ b/src/worker/launch-descriptor.test.ts
@@ -106,6 +106,43 @@ describe("worker launch descriptor", () => {
     });
   });
 
+  it.each(["workspaceDir", "workerContainmentRoot"] as const)(
+    "accepts absolute %s paths through the project preparation limit",
+    (field) => {
+      for (const root of ["/", "C:\\", "\\\\server\\share\\"]) {
+        for (const length of [257, 4_096]) {
+          const descriptor = launchDescriptor();
+          descriptor.assignment[field] = root + "a".repeat(length - root.length);
+
+          expect(parseWorkerLaunchDescriptor(descriptor)).toEqual(descriptor);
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6f6f7af4d15246a7f58147e1db8270652e1e701c](https://github.com/openclaw/openclaw/pull/142198/commits/6f6f7af4d15246a7f58147e1db8270652e1e701c)
- Landed commit: [670d92d5cabea30de6ff17434de33ccc9bf3ca60](https://github.com/openclaw/openclaw/commit/670d92d5cabea30de6ff17434de33ccc9bf3ca60)

PR #142201, opened 2026-09-08
fix(ui): numeric step buttons skip the first allowed value

steipete:
<details>

This is a same-repository branch; repository maintainers can update it.

</details>

## What Problem This Solves

Fixes an issue where the first step-button click in an empty bounded numeric setting skipped the first permitted value. For example, incrementing an empty positive integer started at 2 while the same field's keyboard arrow started at 1. Negative ranges had the corresponding decrement error.

## Why This Change Was Made

The numeric control now normalizes the requested step once instead of clamping its empty seed before stepping. The same owner also combines two invalid-input states whose validation and commit behavior were identical, while keeping a genuine empty value distinct from partially typed numeric text. The disabled guard, inherited-default behavior, finite-number handling, and parent validation are preserved. Production code shrinks by four lines.

## User Impact

Step buttons and keyboard arrows start empty positive or negative ranges at the same permitted endpoint. Existing values and defaults continue to step normally, and incomplete numeric text remains distinct from an intentional clear.

## Evidence

The complete Settings form with a synthetic Gateway reproduced all four discrepancies before the change: buttons saved 2, 4, -2, and -4 while native arrows saved 1, 2, -1, and -2. The same browser scenario now passes and verifies actual config writes and optional clears. Before/after panel screenshots were inspected and contain only synthetic values.

The 41 existing scalar, integrity, and rejection browser cases pass, including inherited defaults, partial numeric input, true clears, precision checks, composed constraints, and parent rejection. The full owning UI E2E test type graph, full-file typed lint, formatting, and diff sanity pass. Fresh independent P2 review of the exact final three-file candidate found no actionable issues.

| Before: the first button step skips each endpoint | After: buttons match native arrow endpoints |
| --- | --- |
| Before: numeric buttons select 2, 4, -2, -4 | After: numeric buttons select 1, 2, -1, -2 |

Proof uses the real Control UI in Chromium with a synthetic mock Gateway. Config writes were checked at the RPC boundary; no operator Gateway, live provider, or real config storage was used.


steipete:
Merged via squash.

- Prepared head SHA: [706c3e2906d6373110f11619584b5c9ecb522809](https://github.com/openclaw/openclaw/pull/142201/commits/706c3e2906d6373110f11619584b5c9ecb522809)
- Landed commit: [ab24c12337b5484898368cfcda54c74985626027](https://github.com/openclaw/openclaw/commit/ab24c12337b5484898368cfcda54c74985626027)

PR #142202, opened 2026-09-08
feat(providers): keep registered setup connection-only

obviyus:
## What Problem This Solves

Provider setup saved copies of built-in catalogs into configuration, duplicating plugin-owned inventory and obscuring the distinction between a connection and authored model settings.

## Why This Change Was Made

Eleven provider registrations use shared connection-only presets. Shared descriptors remove duplicate decisions while explicit replace mode keeps its existing catalog merge and required-default rules. Published helper APIs retain catalog-seeding behavior.

## User Impact

Ordinary setup saves the connection while preserving authored rows, defaults, fallbacks, and aliases. The new plugin imports require a matching host release; they are not compatible with older hosts.

Consumers: the migrated provider setup registrations stop seeding built-in rows during ordinary setup.

## Evidence

All 128 focused provider and plugin-development checks passed. Real registered setup saved seven generated rows before the change and zero afterward while preserving authored settings. Compiled setup passed fresh, authored-merge, and authored-replace cases. A real Gateway listed the default and delivered a synthetic reply without changing saved configuration. Required continuous integration passed. Two representative providers were exercised; no live provider request or package publication occurred. An unrelated standalone plugin dependency-link failure remains outside these successful checks.

### DIFF
--- a/docs/plugins/sdk-provider-plugins/model-catalogs.md
+++ b/docs/plugins/sdk-provider-plugins/model-catalogs.md
@@ -340,10 +340,33 @@ agent/workspace paths.
 
 If your auth flow also needs to patch `models.providers.*`, aliases, and
 the agent default model during onboarding, use the preset helpers from
-`openclaw/plugin-sdk/provider-onboard`. The narrowest helpers are
-`createDefaultModelPresetAppliers(...)`,
+`openclaw/plugin-sdk/provider-onboard`. For registered connection-only setup,
+use `createProviderConnectionPresetAppliers(...)` with a lazy `catalogModels`
+supplier. Ordinary setup writes connection facts, aliases, and a missing
+primary without copying the built-in catalog into saved configuration.
+Explicit `models.mode: "replace"` evaluates the supplier and merges generated
+rows behind authored rows. Each setup result owns its generated model data.
+
+Use `createDefaultModelsConnectionPresetAppliers(...)` when replace mode must
+retain the existing required-default rule: add the supplied `defaultModels`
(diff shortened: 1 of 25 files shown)


obviyus:
Thanks @obviyus!

**Proof:** Registered setup saves connection/default facts without copying the built-in catalog. The pinned NVIDIA baseline saved seven generated rows; the candidate saved zero. Authored merge and replace settings remain intact. The matching compiled Mistral host/plugin completed fresh and existing setup, then a real Gateway returned the synthetic provider reply for the configured default with matching requested/effective/response model identities. Saved configuration remained unchanged through catalog and inference.

- **Change:** Eleven provider registrations now use the shared connection-only presets. Public helper behavior and explicit replace rules remain intact.
- **Validation:** 128 focused checks, five independent retained-artifact acceptance clauses, exact-head local review, and [required CI](https://github.com/openclaw/openclaw/actions/runs/34234967988).
- **Release boundary:** The new SDK imports require the paired host/plugin release and normal compatibility-range synchronization. No old-host compatibility, real vendor response, or npm-install proof is claimed.

| Before | After |
| --- | --- |
| Ordinary setup copied generated catalog rows into saved configuration. | Ordinary setup preserves authored model preferences and saves the connection without generated rows. |
| Provider registrations used catalog-seeding helpers. | Registrations use one shared connection-only policy while published helper APIs retain their contracts. |

Earlier fixture and standalone archive-validation failures remain recorded. Their successful operations were preserved; only the fully verified Mistral dependency path supplies the compiled runtime claim.

PR #142154, opened 2026-09-08
ci: avoid unrelated upgrade checks on test-only PRs

steipete:
Related: #142140, #141146

## What Problem This Solves

Resolves unnecessary Docker upgrade checks on PRs that only change Doctor, updater or state tests and their private helpers. For example, #142140 changed no production code but selected the packaged `published-upgrade-survivor` lane because both test filenames matched the broad Doctor owner prefix.

That job executed for 9m50s: its Docker body included 315s of package preparation and 194s of meaningful upgrade validation. Those workloads do not consume the changed Doctor test files.

## Why This Change Was Made

Apply the existing test-only classifier to source-tree published-upgrade owner matches. Production Doctor/update/state sources, both schema constants, survivor fixtures, and explicit planner/workflow triggers keep their existing selection. Every admitted canonical main push still runs the upgrade tripwire.

This is a PR scope correction in the existing selector; no new matcher catalog, runner, shard, cache, timeout or gate override. Production/tooling LOC: +4/-1; tests: +37; docs: +1/-1. The small conditional expresses the missing source ownership distinction.

## User Impact

Qualifying test-only PRs avoid an unrelated package build and Docker run. Their selected Node tests and checks still run. This does not remove main integration coverage or claim that main's Docker work is unnecessary.

## Evidence

- Four regression cases failed against the original selector and pass with the fix: Doctor test, private helper, normalized Windows helper path, and state test.
- **60 focused checks passed**, including real production/schema paths, mixed production and test changes, Docker survivor fixtures, and real-selector output passed through the PR/main manifest logic. Existing unconditional-main and scheduler guards remain green.
- Required `check:changed` checks passed, including scripts/test-root types, lint, format and Docker E2E boundary checks. Node 24.20.0, pnpm 12.3.4. Independent P0–P2 review found no actionable issue.
- Source audit traced the two Doctor files exclusively to Vitest/private worker preparation. The package build consumes the separate production runtime-entry set; the installed survivor scenario consumes the built tarball and Docker harness, not these source tests.

This planner-editing PR still selects its own explicit Docker proof lanes. No future wall-time saving is claimed as measured until a qualifying test-only PR runs under the repaired selector.

PR #142215, opened 2026-09-08
test: join native controllers before removing triage fixtures

steipete:
Related: #142154

## What Problem This Solves

Resolves a native test-fixture cleanup race exposed by [CI 34227269292](https://github.com/openclaw/openclaw/actions/runs/34227269292/job/102064585661). Cleanup could remove its temporary directory while a controller process was still starting and had not registered its PID. The run failed with `ENOTEMPTY`; deterministic regressions independently prove the missing join, although the exact writer from that historical failure was not captured.

## Why This Change Was Made

Give the fixture helper its own process group and record nested detached groups at their creator before launch. Cleanup closes further group admission, waits for ownership publication, stops and joins known groups, and removes files only after verified quiescence. Unknown ownership retains the directory and reports failure. Publication and group completion share the existing five-second cleanup budget.

The disconnect scenario now signals its owned synthetic executor after its handler is installed, replacing a filesystem watcher that could miss the trigger. It still tests actual IPC loss, scope shutdown and absence of late restoration. Failure diagnostics include the existing helper log and event history.

## User Impact

More reliable CI and complete cleanup of test-owned native processes. Production behavior, service control, schemas, deadlines and existing assertions remain unchanged. No removal retry or extra sleep. Production LOC: 0; fixture support: +88/-21; tests: +82/-10.

## Evidence

- Two new regressions hold a controller before it can register, covering helper and executor process groups. Both fail on the original fixture because cleanup returns with the exact controller alive; both pass after repair. Probe cleanup verifies PID start identity before rescuing the old-code child.
- The existing cancelled-loss case passes in the broad sibling runs. Those runs passed **54/55** and exposed the separate disconnect fixture issue; the repaired disconnect then passed a targeted run. This is not presented as one final 55/55 run.
- A forced-order probe demonstrated the old watcher could miss a trigger after advertised readiness. The final signal-based fixture removes that incidental filesystem-notification dependency without changing the IPC-loss assertions.
- All required changed-gate components passed; the initial lint pair was corrected and final focused lint passed. Fresh independent P0–P2 review of the final source found no actionable issue. No additional runtime run after the final error-aggregation cleanup is claimed; exact-head Linux CI remains required before landing.

Local compilation was affected by heavy workstation load, so this PR makes no throughput speedup claim. Its benefit is fixing the demonstrated lifecycle gap and avoiding cleanup-driven CI retries. Temporary diagnostics and held processes were cleaned up.

### DIFF
--- a/src/infra/update-managed-service-triage.test-support.ts
+++ b/src/infra/update-managed-service-triage.test-support.ts
@@ -1,11 +1,13 @@
 // Synthetic native boundary only: real Node helpers, IPC, leases and descendants.
 import { spawn } from "node:child_process";
+import { readFileSync, readdirSync } from "node:fs";
 import fs from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { createInterface } from "node:readline";
 import { DatabaseSync } from "node:sqlite";
 import { vi } from "vitest";
+import { inspectManagedProcessGroup } from "../../scripts/lib/managed-child-process.mts";
 import { resolveServiceManagerEnv } from "../daemon/service-process-env.js";
 import { buildCliRespawnPlan } from "../entry.respawn.js";
 import { getFileLockProcessStartTime, isPidAlive } from "../shared/pid-alive.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [dbc16942fa44a1bd3204c2675d83b83c93447cbe](https://github.com/openclaw/openclaw/pull/142215/commits/dbc16942fa44a1bd3204c2675d83b83c93447cbe)
- Landed commit: [76fb286ee76d399e3582e3a5c74c5a344ebb9834](https://github.com/openclaw/openclaw/commit/76fb286ee76d399e3582e3a5c74c5a344ebb9834)

PR #142225, opened 2026-09-08
feat(canvas): report widget runtime script errors back to the agent

steipete:
## Problem

After #141939, `show_widget` rejects scripts that do not parse, but a widget whose script throws at runtime (a missing element id, a null dereference, a handler that throws on click) still renders as a dead card and the agent never learns about it. Follow-up to the roboclaw duplicate-widget case: the model needs a ping it can act on, without the tool waiting on the client.

## Solution

Asynchronous, no tool wait, no new Gateway method:

- `src/canvas/wrap.ts`: a sixth host bridge, injected right after the widget bridge and before widget code, listens on `window` for `error` and `unhandledrejection` (capture phase) and posts `openclaw:widget-runtime-error` to the parent like the size reporter. Bounds: message ≤ 500 UTF-16 units, source basename ≤ 200, integer line/column when present, at most 3 distinct reports per document load, resource-load errors ignored.
- `ui/src/components/canvas-widget-view.ts`: after the existing source/origin verification, renders a one-line `Script error: …` notice above the frame and forwards one report per document and chat session per page load through the existing operator `wake` method (`mode: "now"`, the chat's session key, bounded plain text), under the shared widget-prompt rate limiter. Reports are forwarded only for widgets whose chat message is at most ten minutes old, so restored history shows the notice without waking the agent (the message timestamp is threaded from the bubble through `widget-card` to the view).
- Tool description gains "Runtime script errors surface later as a session event; fix and show again." Docs describe the bridge, bounds, freshness window, delivery, and limits.

Agent-facing text:

```text
Inline widget "<title>" (<docId>) threw a script error after rendering: <message>[, line L[, column C]]. Fix the script and show the widget again; if show_widget is unavailable in this turn, reply with the corrected widget code and show it on the next turn.
```

## Impact

The human sees the failure on the card immediately. The agent receives the error as a session wake event on its next available turn (the immediate wake runs once the session lane is free, per `src/infra/heartbeat-runner-execution.ts` and `src/infra/session-event-wake.ts`). No config, schema, CSP, size, retention, or dashboard behavior changes. Native apps do not report yet.

Known limit, kept out of this PR: the wake turn runs without the originating client's capabilities, so `show_widget` is not in the tool set of that turn. The agent diagnoses and posts the corrected code, and re-shows on the next user turn. Carrying the caller's capabilities through the heartbeat wake queue into the run context is a small authority-boundary change and belongs in its own PR.

## Evidence

- `node scripts/run-vitest.mjs src/canvas/wrap.test.ts src/canvas/widget-tool`: 8 files, 59 tests passed. `node scripts/run-vitest.mjs ui/src/components/canvas-widget-view.test.ts`: 16 passed. The new regressions failed before their fixes.
- `pnpm check:changed`: all lanes green. `pnpm ui:check-performance`: startup JS 338.6 KiB gzip against a 344.0 KiB limit.
- Two Codex autoreview rounds (P0–P2): scoped-clean.
- Live proof on an isolated token-auth Gateway built from this branch, driven through the Control UI with Playwright, `anthropic/claude-sonnet-4-6`:
  1. Agent sent a widget whose script sets `textContent` on a missing element. The card showed `Script error: Cannot set properties of null (setting 'textContent')` within a few seconds and the client sent exactly one `wake` with the text above (docId, line 82, column 55).
  2. The Gateway ran a wake turn for the chat session ten seconds later. The agent replied in the chat with the diagnosis and the corrected script (null guard), and noted `show_widget` was unavailable in that turn, which is the documented limit.
  3. Freshness: reloading a session with three broken widgets aged 6, 16, and 21 minutes sent one wake (the 6-minute widget) and rendered all three notices.

Widget card notice and the agent's wake-turn diagnosis with the corrected script


steipete:
Merged via squash.

- Prepared head SHA: [e093202aecbbe1e2df3d802fdc139a401898da50](https://github.com/openclaw/openclaw/pull/142225/commits/e093202aecbbe1e2df3d802fdc139a401898da50)
- Landed commit: [e572e323471dfea830c24746108a19889fb29d35](https://github.com/openclaw/openclaw/commit/e572e323471dfea830c24746108a19889fb29d35)

Issue #114967, opened 2026-07-28
[Bug]: agent-driven live update left a launchctl submit keepalive validator force-restarting the gateway every ~2 minutes


piotx:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

An agent-driven live-update pass left behind an ad-hoc `launchctl submit` keepalive job (no plist) whose "validate" script began unconditionally with `openclaw gateway restart`, so launchd relaunched the script on every exit and the managed gateway was force-restarted every ~2 minutes indefinitely (11 forced restart cycles in 20 minutes per the gateway lifecycle log); the `openclaw-live-updater` skill has no boundary forbidding ad-hoc persistent launchd jobs and no leftover-job cleanup/verification step.

### Steps to reproduce

Observed incident path (2026-07-27, macOS, live LaunchAgent gateway served from a source checkout):

1. An agent performing the live self-update flow (`.agents/skills/openclaw-live-updater/`) improvised post-update verification instead of relying on the deterministic updater: it created `~/clawd/openclaw-restart-validate-20260727.sh` and registered it via `launchctl submit -l ai.openclaw.update-validator.20260727.v2` with keepalive semantics (no plist file on disk).
2. The script starts with `"$openclaw_bin" gateway restart`, then polls health for up to 60s and exits.
3. launchd keepalive relaunches the script on exit → every cycle begins with another forced `gateway restart` → the gateway is force-restarted every ~2 minutes and never finishes booting; the loop persists across the update finishing, until a human finds and removes the job.

### Expected behavior

The live-updater flow's own contract (SKILL.md + `scripts/update-main.mjs`): restart the managed gateway once through the suspension fence, verify via deep RPC status/health plus a restart-window log audit, and leave no residual launchd state beyond the managed `ai.openclaw.gateway` LaunchAgent. `update-main.mjs` itself never uses `launchctl submit` (its only launchctl usage is print/enable/bootstrap/bootout of the managed LaunchAgent).

### Actual behavior

[...]

PR #142230, opened 2026-09-08
fix: diagnose and repair stray macOS Gateway restart jobs

steipete:
Related: #114967 (kept open)

## What Problem This Solves

Agent-created launchd jobs could repeatedly invoke Gateway lifecycle commands after an update while remaining invisible to normal Gateway status and Doctor. Submitted jobs can have inferred KeepAlive behavior without any plist, so the existing stale-updater label/plist scanner missed the reported validator and upgrade labels.

## Why This Change Was Made

The launchd owner now inspects live OpenClaw-namespace job definitions and reports program, KeepAlive, and verified Gateway lifecycle commands. Normal status includes structured JSON diagnostics. Doctor requires explicit `--fix`, re-inspects the current command, and removes only eligible foreign lifecycle jobs; managed labels, generated services, unrelated labels, and ambiguous scripts are preserved. Plist-backed stray jobs are disabled at login and booted out, with their files retained.

The managed Gateway warns after three external CLI forced restarts in ten minutes, naming likely KeepAlive jobs when available. Its existing lifecycle log supplies the bounded history and warning deduplication, with no new configuration or store. The live-updater skill explicitly forbids `launchctl submit` and auxiliary KeepAlive watchdogs and requires its deterministic suspension-fenced update, status, and health flow.

## User Impact

Operators can discover and repair confirmed stray restart jobs through status and Doctor, with explicit noninteractive removal results. The warning preserves operator restart commands. Complex/unverifiable wrappers and jobs outside `ai.openclaw.*` remain outside automatic repair; updater-only commands are reported but not removed by this lifecycle repair.

## Evidence

Validation for `ac88f920a1d19548dc819c65a6c52c318256d603` (final shell-environment and absolute-path follow-up to `57ca23a2c9a`):

- Shell jobs with the specified execution-altering environment names or any `BASH_*` name remain report-only with `Shell environment alters execution; left unchanged.` Inspection reads names only. Direct non-shell programs and PATH/HOME/OPENCLAW-only environments remain unaffected.
- Scripts and direct argv require absolute OpenClaw executable paths, including helper values and accepted `env`/`node`/`bun` chains. The closed raw-line grammar remains intact. The source comment and macOS guide document the metadata-only boundary; no binary executability, quarantine, or interpreter-availability probes were added.
- P2 review exposed two guard-ordering bypasses involving shell launchers with CLI/runtime basenames and executables selected through `env`; both were reproduced and fixed. A concrete `PS4`/xtrace non-execution case was also reproduced in isolated macOS Bash/sh child processes and added to the same environment guard. The final review's request to block `HOME` for zsh was rejected because the accepted contract explicitly preserves HOME-only environments. No accepted finding remains.
- 119 focused tests passed: 106 classifier/repair cases, two native macOS process cases, and 11 Doctor tests. Regressions demonstrated failures before each accepted repair and passed afterward. Command: `pnpm test src/daemon/launchd-foreign-jobs.test.ts src/daemon/launchd-foreign-jobs.process.test.ts src/commands/doctor-foreign-launchd-jobs.test.ts`.
- Native scratch proof passed for interpreter-launched and directly submitted scripts through the maintenance-owned Doctor runner with isolated policy/history. Detection and the report preserved the jobs; explicit repair removed only the lifecycle scratch job. Managed and observer scratch jobs remained loaded until cleanup. All scratch jobs were removed; the real Gateway and LaunchAgents were untouched.
- `pnpm check:changed`: passed on the final patch, including formatting, typechecks, lint, export scans, and repository guards.
- `git diff --check`: passed. Follow-up diff: three files, 195 insertions and 20 deletions.
- Exact-head CI watch: [run 34249182120](https://github.com/openclaw/openclaw/actions/runs/34249182120) is **queued** on this head. The 180-second bounded watch ended with `TIMEOUT state=PENDING pending=106`; the fresh run read has no conclusion yet. CI is not yet green.

Lead follow-up `bacabd4fb7bcfe324f1cff05919d6c7f1fc9c9e7` (inherited launchd environment):

- A fifth independent P2 review found that the environment guard scanned only the job-specific `environment` block, while `launchctl print` also reports `inherited environment` and `default environment` blocks that reach the shell. The guard now scans all three (names only). Three new classifier cases (inline, script, and direct launches with `BASH_ENV` in an inherited or default block) fail with the guard reverted and pass with it; 106 classifier/repair tests pass, `pnpm check:changed` and `git diff --check` passed. A sixth independent P2 review of this head is scoped-clean under the documented contract.

### DIFF
--- a/.agents/skills/openclaw-live-updater/SKILL.md
+++ b/.agents/skills/openclaw-live-updater/SKILL.md
@@ -14,6 +14,7 @@ Keep one operator-selected canonical live checkout as a read-only-to-the-agent d
 - Escalate only irreversible or materially risky security, privacy, auth, destructive migration/data-loss, protocol-version, credential-rotation, dependency patch/override, or product choices.
 - Diagnose ordinary update, runtime, app, CI, test, and release-validation failures; fix them through a focused PR; prove exact head; land with `scripts/pr`.
 - Never edit, stash, reset, clean, or create a branch in the live mirror.
+- Never use `launchctl submit` or create an ad-hoc KeepAlive job to run updates or Gateway lifecycle commands. A submitted job can relaunch `openclaw gateway restart` after every exit and cause a persistent restart storm, even without a plist ([#114967](https://github.com/openclaw/openclaw/issues/114967)).
 
 ## Fast Update And Maintenance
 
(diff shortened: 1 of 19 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [bacabd4fb7bcfe324f1cff05919d6c7f1fc9c9e7](https://github.com/openclaw/openclaw/pull/142230/commits/bacabd4fb7bcfe324f1cff05919d6c7f1fc9c9e7)
- Landed commit: [cd98a01723689179c2695dcfe626856c3af06186](https://github.com/openclaw/openclaw/commit/cd98a01723689179c2695dcfe626856c3af06186)

PR #142237, opened 2026-09-08
fix(cloud): stop cancelled workspace uploads from hanging

steipete:
Related: #134931

## What Problem This Solves

Fixes an issue where a cancelled cloud workspace upload could leave its client waiting for a response. The request could already be fully read and destroyed while the Gateway was still validating staged content. A second window occurred when ownership ended before the HTTP handler installed its abort listener.

## Why This Change Was Made

The HTTP transfer owner now closes both the response and request on cancellation and handles an already-aborted signal immediately. Existing authorization checks still prevent cancelled content from becoming an accepted upload. This is an independent extraction with no compression, schema, provider, or timeout changes.

## User Impact

Cancelled workspace uploads settle promptly instead of waiting for staging or transfer timeout. Normal uploads continue to finish and publish their accepted manifests.

## Evidence

- Both cancellation cases failed on original main; normal controls passed.
- Real loopback HTTP tests use actual service/token admission and staged-manifest validation. They prove the uploader settles while validation remains paused, then resume validation and verify no cancelled result becomes takeable.
- 37 focused and sibling cases passed; all 14 owner/attachment cases passed after the final assertion refinement. Complete changed-file checks and final independent P2 review passed.
- No native cloud provider was used; real request/response I/O is the changed boundary.


steipete:
Merged via squash.

- Prepared head SHA: [01b41d2103afc903af51674f3a3180fb949696cb](https://github.com/openclaw/openclaw/pull/142237/commits/01b41d2103afc903af51674f3a3180fb949696cb)
- Landed commit: [9956abfbe30c914443fc3467f5472c07ddb68c5a](https://github.com/openclaw/openclaw/commit/9956abfbe30c914443fc3467f5472c07ddb68c5a)

PR #142242, opened 2026-09-08
feat(cloud): compress workspace manifest transfers

steipete:
Related: #134931, #142237, #143133

## What Problem This Solves

Cloud workspace synchronization sends immutable file manifests as uncompressed JSON, even when the node can accept compression. Large workspaces therefore transfer more manifest data than necessary.

## Why This Change Was Made

Negotiate fast gzip for manifest responses and decode it in the node transfer client. Reuse the existing Control UI encoding-selection rules, preserve identity clients, bound both wire bytes and decoded bytes at 64 MiB, and keep manifest digests tied to decoded content. The Gateway rechecks live transfer authority after compression before sending a response.

This is a standalone extraction from #134931: eight files, 446 added and 76 removed lines. It adds no prepared workspaces, pool state, configuration, dependencies, or protocol-version change.

## User Impact

Capable workers transfer fewer manifest bytes. Clients that do not request gzip retain plain responses. Corrupt, unsupported, oversized, or cancelled responses fail without replacing the workspace.

## Evidence

- The current candidate at `a4f4137dd7e404341baa0ff0b90937aa7405e3e4` passed 235 tests across ten files, with one existing Windows-only skip. Coverage includes identity compatibility, encoding quality values and 406 responses, post-compression revocation, digest mismatch, corrupt/truncated/stacked encodings, wire/decoded size limits, cancellation, and the landed upload-cancellation and stream-fixture boundaries.
- The full changed-file gate passed against `73b4086ff658973fb9ae07fa208c49ca8de82a67`, including production and test typechecks, all three dead-export scans, lint, schema/store guards, and runtime import cycles. Formatting and documentation validation also passed.
- Independent P2 review found no actionable defects. Reconciliation onto the current base preserved the seven-file code/test patch byte for byte; only the existing documentation paragraph moved to the new cloud session lifecycle page. All eight committed files match the validated candidate.
- A real authenticated loopback HTTP exchange prepared and transferred a synthetic 12,000-file manifest through the production transfer handler and node HTTP transport, verified identical decoded content, and cleaned up its resources. Identity used 1,755,103 wire bytes; gzip used 548,761, a 68.73% reduction. Median response completion plus decoding was 2.97 ms for identity and 14.04 ms for gzip. This demonstrates reduced transfer size and its local processing cost; it does not establish cloud startup or network latency improvement.

- [CI run 34378045922](https://github.com/openclaw/openclaw/actions/runs/34378045922) passed on exact head `a4f4137dd7e404341baa0ff0b90937aa7405e3e4` after one targeted artifact-job retry. The first job lost most of its 20-minute budget to repeated cache-service HTTP 500s; build/upload/smokes passed, but artifact checks were interrupted. The unchanged-head retry completed those checks and the aggregate gate without source, timeout, or cache-policy changes.


steipete:
Merged via squash.

- Prepared head SHA: [a4f4137dd7e404341baa0ff0b90937aa7405e3e4](https://github.com/openclaw/openclaw/pull/142242/commits/a4f4137dd7e404341baa0ff0b90937aa7405e3e4)
- Landed commit: [e23df626180590bd81dcbe2eccf0f1208e2edd42](https://github.com/openclaw/openclaw/commit/e23df626180590bd81dcbe2eccf0f1208e2edd42)

PR #142253, opened 2026-09-08
feat(pdf-tool): local OCR fallback for scanned PDFs with no vision model

abacha:
## What changed

The `pdf` tool had zero path for a scanned PDF (image-only page, no text
layer) once every configured vision model failed or none was configured —
native providers and the extraction-fallback model call both need a working
vision-capable model, and there was no other option.

Adds a local OCR fallback, gated tightly so it only ever engages for that
exact case:

- Extraction stays lazy (unchanged) so native-provider and normal textual
  PDFs never pay for it.
- When every vision-model candidate fails or none is configured, the tool
  checks whether extraction already showed a near-empty text layer
  (`< 200 chars`) *and* rendered images (the existing scan-detection signal
  reused from the extraction fallback path). If extraction itself never ran
  or also fails, the original vision error is what surfaces — OCR can't help
  there.
- Only for a confirmed scan: rasterize the requested pages with `pdftoppm`
  (poppler) and OCR each page with the local `tesseract` binary (`por` by
  default). No network call, no credentials. The OCR text is returned as-is
  — this is text **extraction**, not comprehension, so it is never fed back
  through a model.
- If the toolchain is missing, the tool raises a clear
  `Local OCR fallback unavailable: ...` instead of degrading silently.
- If OCR runs but finds nothing (chart, skewed table, handwriting), the tool
  still errors, explicitly saying those need a working vision model —
  distinct from "couldn't read the text at all".
- New `src/media/pdf-ocr-local.ts` module; `details.ocrFallback: true` is set
  on a successful OCR-fallback result.

## Validation run

- `node scripts/run-vitest.mjs run src/agents/tools/pdf-tool.test.ts src/agents/tools/pdf-tool.ocr-fallback.test.ts` — 38/38 pass (all pre-existing PDF-tool tests unaffected; 4 new OCR-fallback tests: success, textual-PDF-unaffected, toolchain-missing, OCR-finds-nothing).
- `node scripts/run-vitest.mjs run src/agents/tools/pdf-tool.helpers.test.ts src/agents/tools/pdf-tool.model-catalog.test.ts src/agents/tools/pdf-native-providers.test.ts src/agents/tools/pdf-tool.model-config.test.ts src/agents/tools/pdf-tool.native-providers.test.ts src/agents/tools/pdf-tool.runtime-abort.test.ts src/agents/tools/pdf-tool.static-runtime.test.ts src/media/pdf-extract.test.ts` — 76/76 pass, including "uses native PDF path without eager extraction" (confirms native/textual paths are unaffected).
- `pnpm tsgo:core` — clean.
- `node scripts/run-oxlint.mjs` on the touched files — clean.
- Live end-to-end proof on a host with the toolchain installed: built a fabricated scanned PDF (pure raster page, `pdftotext` confirms empty text layer), ran `runLocalPdfOcr` directly against it — output: `"Prestação de serviço — Soluções Agrícolas Ltda\n\nTotal: R$ 3.947,55\n\nData de emissão: 08/09/2026"`, matching the source image exactly (accentuation included).

## Trade-offs / residual risks

- OCR accuracy/quality is whatever `tesseract` gives at 200 DPI with the
  `por` model — no attempt to tune DPI or language selection beyond the
  existing fleet default.
- Multi-PDF OCR runs pages sequentially per PDF; fine for the occasional
  scan fallback, not optimized for throughput.
- No new tool parameter for OCR language; kept to the minimum surface area
  (default `por`) per the scope of this fix.

### DIFF
--- a/docs/tools/pdf.md
+++ b/docs/tools/pdf.md
@@ -90,6 +90,24 @@ Details:
 - If image rendering fails, OpenClaw drops the images and continues with the extracted text.
 - If the target model is text-only and extraction produced images, OpenClaw drops the images and sends text only.
 
+### Local OCR fallback (scanned PDFs, no vision model)
+
+When extraction finds a scan with no usable text layer (every page's extracted text is under
+`200` characters _and_ the page rendered images), and every configured vision model is either
+unconfigured or fails, the tool rasterizes the requested pages locally with `pdftoppm` (poppler)
+and reads them with the local `tesseract` binary — no network call, no credentials. The result
+is the raw OCR'd text, returned as-is: this is text **extraction**, not comprehension, so it is
+not sent through any model.
+
+- Default OCR language is `por`.
+- Requires `pdftoppm` and `tesseract` on the host; if either binary is missing the tool raises
+  `Local OCR fallback unavailable: ...` instead of silently degrading.
+- If OCR runs but finds no text either (charts, skewed tables, handwriting), the tool still
+  errors — those need a working vision model, and the error says so explicitly.
+- On success, `details.ocrFallback` is `true` and `details.native` is `false`.
+- A normal textual PDF is unaffected: this path only triggers when text extraction is already
+  near-empty, so nothing changes for the extraction-fallback mode described above.
+
 ## Config
 
 ```json5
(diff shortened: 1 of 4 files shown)


abacha:
Closing this unmerged. The review's ownership call is right: OCR belongs to the document-extraction capability, not to a post-model catch inside the pdf tool, and the P1 findings (no cancellation signal into the subprocesses, no pixel budget on the 200 DPI rerender) are defects of that placement rather than details to patch.

The capability gap itself stays on record as #143375 — the tool is gated on a vision model before it opens the document, so the extraction plugin's existing scan signal is unreachable when there is no model to call. We are not carrying a local patch for it.

Not deleting the branch (`pdf-ocr-fallback`) — the helper and its tests are there if the extraction-owner version gets taken up.

Issue #103429, opened 2026-07-10
[Bug]: Release notes verifier omits valid source PRs


vincentkoc:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

Yes

### Summary

The release-notes verifier omits valid source PRs and retains stale contextual PR rows, so its complete contribution record is not an ancestry-complete account of the release.

### Steps to reproduce

1. Check out current `main`.
2. Run `.agents/skills/openclaw-changelog-update/scripts/verify-release-notes.mjs` for version `2026.7.1`, base `v2026.6.11`, source target `306b800ace5398dcfc5eae6e15dcae533db42c95`, and final target `811ddd96180583bae00001f71971419182ae0520`.
3. Compare the rendered verifier ledger with an independent enumeration of every target-side commit from the comparable merge base, GitHub associated PRs, explicit source-PR references, canonical cherry-pick provenance, stable patch identities, reverts, and direct commits.

### Expected behavior

The verifier should enumerate the complete target ancestry, exclude already-shipped or reverted work only with proof, resolve every source PR through auditable provenance, reject contextual issue/PR mentions as ownership, and fail closed when provenance is unresolved.

### Actual behavior

The verifier reports 1,461 PRs. Independent reconciliation finds 1,960 canonical source PRs: 501 valid source PRs are missing and two contextual/stale rows are incorrectly retained. The same history contains 940 manifest-direct audit commits and 929 mutually exclusive direct-source commits; those must remain auditable without being rendered as fake PR rows.

### OpenClaw version

`2026.7.1` release history through `811ddd96180583bae00001f71971419182ae0520`

### Operating system

macOS; git history plus GitHub API verification

### Install method

Repository checkout

### Model

N/A

### Provider / routing chain

N/A

### Logs

```text
reported_prs=1461
missing_source_prs=501
false_context_rows=2
canonical_source_prs=1461-2+501=1960
active_unshipped_nonmerge_commits=3172
source_owned_commits=2243
manifest_direct_commits=940
exclusive_direct_source_commits=929
```

### Screenshots, recordings, and evidence

- Range boundary: `merge-base(v2026.6.11, 306b800ace5398dcfc5eae6e15dcae533db42c95)..306b800ace5398dcfc5eae6e15dcae533db42c95`.
- Provenance split for the 1,960 source PRs: 175 association-only, one explicit-reference-only, and 1,784 supported by both.
[...]

PR #142261, opened 2026-09-08
fix(release): keep changelog credits within verified history

steipete:
Related: #103429

## What Problem This Solves

Resolves a problem where release-note generation rejected explicit CI run references, credited changes that had been reverted together, or promoted contextual PR references from outside the selected release history into the contribution record.

## Why This Change Was Made

Bring the accounting repairs qualified during 2026.9.3 preparation back to main. An unresolved reference is classified as an Actions run only with explicit source context and an exact same-repository run identity; actual issue/PR resolution still wins. Explicit multi-commit reversals require an independently reconstructed inverse tree. Contribution membership follows the frozen source graph and verified backport provenance, while contextual references remain available for inline resolution and explicit seeds retain historical credit.

The current EOF transport handling and newer release preparation policy are preserved. This addresses related accounting boundaries reported in #103429; it does not claim to complete that issue's broader historical reconciliation.

## User Impact

Release maintainers can generate notes for a frozen cut without crediting unrelated later work or genuinely reverted changes. Valid CI references no longer block generation, and unresolved or unproven references still fail strictly. Runtime behavior and published release artifacts are unchanged.

## Evidence

- CLI/fake-GitHub regression cases failed on unchanged main for the intended defects: 19 failures, with 23 controls passing.
- The repaired verifier and ledger suites pass all 122 tests, including existing EOF retry coverage.
- The verifier body matches the implementation already qualified for 2026.9.3.
- Full `check:changed` passed, including all typecheck lanes, repository guards, lint, runtime import cycles, and changed root-test lint.
- Fresh independent P2 review found no actionable findings.

### Review follow-up

Keep Actions IDs above GraphQL's signed 32-bit `Int` range out of issue queries, while retaining strict unresolved-reference errors and exact workflow-run identity checks. A schema-aware CLI fixture reproduced six failures before this follow-up; the repaired full suites pass 129 tests, including the inclusive maximum issue-number boundary. A direct live query currently returned partial data with `NOT_FOUND`, so this is a documented API-contract repair rather than a claim of a reproduced live outage. Independent P2 review is clean.

This changes release-generation accounting only. It does not change runtime database schemas or stored user state; manifest schema identity remains unchanged, with compatibility covered by the existing seed, snapshot, and ledger fixtures.

### DIFF
--- a/.agents/skills/openclaw-changelog-update/SKILL.md
+++ b/.agents/skills/openclaw-changelog-update/SKILL.md
@@ -76,6 +76,13 @@ every human `Thanks @...` attribution.
    `--refresh-github-snapshot` after suspect API data, `--github-snapshot
 <path>` for an explicit artifact, or `--no-github-snapshot` for a live-only
    audit. GitHub release bodies are always read live.
+   Explicit `CI #`, `CI run #`, `Actions run #`, and `workflow run #` references
+   in active source are classified separately only when issue/PR resolution
+   fails and a live same-repository Actions lookup confirms the exact run ID.
+   Any ordinary occurrence of that number in active source, notes, or the
+   contribution record remains a strict issue/PR requirement. Confirmed runs
+   appear as `workflowRuns` in verification output and the manifest, never as
+   PR associations or contributor credit.
    - the manifest is the required input to the rewrite, not an after-the-fact
      audit; it contains every referenced PR, eligible contributor credit,
      inline issue context, every direct commit, and an editorial-eligibility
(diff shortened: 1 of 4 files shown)


steipete:
Addressed the GraphQL integer-boundary finding in 4f27339b08658e11b6444c4f2364c0ccd62177f2. Values above signed 32-bit `Int` no longer enter issue queries; they still require explicit Actions context and exact same-repository run confirmation. Ordinary and mixed references remain strict failures. The inclusive maximum still resolves through the issue path.

The schema-aware CLI fixture failed six cases before the repair; the full verifier/ledger suites now pass 129 tests, with full `check:changed` and fresh P2 review also passing. A direct live GitHub query currently returned partial data plus `NOT_FOUND`, rather than a coercion error, so this is API-contract hardening and not a claim of a reproduced live outage.

The generic data-model warning does not apply to runtime storage: there is no database schema or user-state migration in this patch. Existing manifest identity, seeds, snapshots, and ledger compatibility tests remain intact.

steipete:
@clawsweeper re-review

The reviewed large-ID boundary is addressed in 4f27339b08658e11b6444c4f2364c0ccd62177f2; the preceding disposition records 129 passing tests, the schema-aware regression, and the live API nuance.

steipete:
Merged via squash.

- Prepared head SHA: [4f27339b08658e11b6444c4f2364c0ccd62177f2](https://github.com/openclaw/openclaw/pull/142261/commits/4f27339b08658e11b6444c4f2364c0ccd62177f2)
- Landed commit: [7699b4d3a0350905d483e54bafb9716996e4b762](https://github.com/openclaw/openclaw/commit/7699b4d3a0350905d483e54bafb9716996e4b762)

PR #142210, opened 2026-09-08
fix(tui): preserve long words and addresses while wrapping

steipete:
Related: #69344.

## What Problem This Solves

Fixes an issue where users reading or copying long email addresses, words, and identifiers in the TUI received text with extra spaces inserted into it—even when it fit on one terminal line. A combining accent could also be separated from its letter.

For example, `alexandertheodorewilliamson@example.org` appeared as `alexandertheodorewilliamson@exam ple.org` in both user messages and assistant replies.

## Why This Change Was Made

Make the terminal renderer the sole wrapping owner. The pinned renderer already wraps by display width and grapheme boundary, so the duplicate 32-character pre-wrapping pass, text-classification exceptions, extra Markdown parser, and obsolete wrapper can be deleted. Callers now use the existing terminal sanitizer directly; terminal-control, binary-content and RTL handling remains unchanged, as does literal tool-output rendering.

## User Impact

User messages, assistant replies, and tool output preserve their original long text. It still wraps at the actual terminal width without inventing spaces or separating grapheme sequences.

## Evidence

- Four actual renderer regression cases failed before the change and pass afterward, covering wide/narrow widths and both constructor/update paths.
- The real native PTY reproduction failed before the change with the address corrupted in both message rows. The same email flow passes afterward, exercising the real TUI with its existing synthetic backend.
- All **149 focused tests** pass. The PTY scenario plus the complete evidence-parser suite pass **6/6** (one native PTY scenario and five parser cases).
- The canonical owning test type graph, full-file lint, formatting, and fresh P0–P2 independent review pass. Seven retained terminal-control/binary/RTL functions are byte-identical to the baseline.

The intact email now produces the pinned renderer's intended `mailto:` hyperlink. The test evidence parser was updated to recognize that scheme explicitly, retaining its existing control, whitespace, URL-validation and unsupported-scheme checks. Its new cases accept valid mailto links and reject empty/whitespace/control targets. This is test support for the observed renderer output, not a product URI-policy change.

[...]

PR #142262, opened 2026-09-08
improve(tui): reduce URL preparation work for link-heavy messages

steipete:
## What Problem This Solves

Link-heavy TUI messages repeat URL extraction and membership work when their terminal rendering is rebuilt. This adds avoidable preparation cost to streaming updates and resized messages.

## Why This Change Was Made

Keep the ordered, deduplicated URL set from extraction through OSC8 annotation instead of converting it to an array and repeatedly scanning it. Collect authored Markdown destinations while removing their source spans, preserving exact targets, punctuation, wrapping, ANSI handling and the existing render-cache lifetime. This changes internal preparation only; production LOC is neutral and tests add 36 lines.

## User Impact

Less URL-preparation work for link-heavy messages, with unchanged visible text and clickable destinations. Results are workload-dependent; this is not a claim that all terminal rendering is faster.

## Evidence

Observed warm medians in the two measurement orders:

| Boundary and fixture | First order, before → after | Reverse order, before → after |
|---|---:|---:|
| Extract 128 authored URLs; native hyperlinks off | 38.500 → 33.666 µs | 34.667 → 32.166 µs |
| Annotate prepared lines with those URLs; native hyperlinks off | 280.375 → 102.375 µs | 271.666 → 100.416 µs |
| HyperlinkMarkdown lifecycle, 128 mixed URLs; native hyperlinks on | 14.107 → 13.193 ms | 13.951 → 13.225 ms |

The annotation case improves 63.49%/63.04%; the component lifecycle improves 6.48%/5.20%. Adverses are retained: duplicate-URL extraction plus annotation with native hyperlinks off is 29.56%/32.04% slower, and the duplicate-URL ChatLog lifecycle is 8.03% slower in the first off order (2.28% faster in reverse). Across the complete cohort, 1,224 of 2,564 indexed lifetime comparisons are adverse. Of 300 warm lifecycle median comparisons, 139 were adverse. After forced GC with 641 components/collections retained, whole-process heap was about 0.125–0.172 MiB higher in all four pairs; this is not a per-component allocation estimate or a memory improvement. Lifecycle measurements include frame snapshots and identity bookkeeping; these are descriptive observations, not confidence or tail-latency estimates.

- Four complete suites passed on both variants: **178 baseline + 178 candidate tests**. All **29 normal checks** passed; independent P2 review was clean.
- An independent saved-data audit verified **all 5,128 records and 600 gzip members**, including complete frame strings, ordered URLs, recorded array-identity relations and OSC8 destinations. Collection kind intentionally changes from Array to Set. All timings, adverse cases, resource readings and timer-resolution zeros remain retained.
- The real PTY narrow-Unicode/copy-safe-URL test passed over a **fake backend**: one selected test, 68 skipped. This does not prove live Gateway/provider/network behavior.
- The first B0/off Node target succeeded, but its outer Python 3.9 verifier failed on an unsupported `zip` keyword. Existing data was validated with Python 3.14 **without rerunning the product**. Later parents used Python 3.14; the observer mismatch and interruption gap remain limitations. The ABBA sequence was neither uninterrupted nor randomized.

Attached before/after images present the same inspected synthetic frame from saved terminal output. They are captured-frame proof renderings, not live desktop screenshots or an additional product run. Measurements remain tied to baseline `1e1385e0`; no whole-Gateway speedup is claimed.

Before: synthetic recorded terminal frame

After: synthetic recorded terminal frame

Additional integration after main advanced: upstream #142210 replaces the component sanitizer to preserve long words and email addresses. The clean combined tree retains that fix and the ordered URL Set. All **86 tests in the two whole existing HyperlinkMarkdown and OSC8 suites passed** on the combined owner/caller slice; source images were restored after positive process closure. This checks the composition with the newer sanitizer, not an entire-current-main build or a new timing cohort.

### DIFF
--- a/src/tui/components/hyperlink-markdown.ts
+++ b/src/tui/components/hyperlink-markdown.ts
@@ -23,7 +23,7 @@ function sanitizeMarkdownDisplayText(text: string): string {
  */
 export class HyperlinkMarkdown implements Component {
   private inner: Markdown;
-  private urls: string[];
+  private urls: ReadonlySet<string>;
   private cachedRender?: { width: number; lines: string[] };
 
   constructor(
--- a/src/tui/osc8-hyperlinks.test.ts
+++ b/src/tui/osc8-hyperlinks.test.ts
@@ -6,13 +6,13 @@ import { addOsc8Hyperlinks, extractUrls } from "./osc8-hyperlinks.js";
 describe("extractUrls", () => {
   it("extracts bare URLs", () => {
     const urls = extractUrls("Check out https://example.com for more info");
-    expect(urls).toEqual(["https://example.com"]);
+    expect([...urls]).toEqual(["https://example.com"]);
   });
 
   it.each([".", ",", ";", "!", "?", ":", "*", "_", "~", ".?!"])(
     "excludes trailing GFM punctuation %s from bare URLs",
     (punctuation) => {
-      expect(extractUrls(`Visit https://example.com/path${punctuation}`)).toEqual([
+      expect([...extractUrls(`Visit https://example.com/path${punctuation}`)]).toEqual([
         "https://example.com/path",
       ]);
     },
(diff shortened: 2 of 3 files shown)


steipete:
The workload-dependent performance tradeoff is accepted for this scoped optimization. The patch keeps the already-created URL membership Set within the existing rendering lifecycle and preserves exact destinations; it adds no new cache or public behavior. This is the mitigation described in the review.

The PR already reports the duplicate-heavy regressions, higher retained-process heap readings, interrupted first off-mode comparison, and absence of a universal speedup. Those qualifications remain. The useful evidence is narrower: prepared-line annotation and large mixed-link component lifecycles improve in both orders, while complete captured frames and hyperlink targets match. Independent correctness review found no actionable defect.

No timing samples will be discarded or rerun to make the result look uniform. This disposition accepts the documented tradeoff; it does not treat the reported measurements as statistical proof or a full-Gateway gain. Exact-head CI and native landing gates still apply.

steipete:
Merged via squash.

- Prepared head SHA: [31d53f357f06523ceee2b6083a45131637bcb1d7](https://github.com/openclaw/openclaw/pull/142262/commits/31d53f357f06523ceee2b6083a45131637bcb1d7)
- Landed commit: [6939d37adb2b7d397eddfbebff87665632013348](https://github.com/openclaw/openclaw/commit/6939d37adb2b7d397eddfbebff87665632013348)

PR #138438, opened 2026-09-04
fix(cli): automations runs required --id while sibling commands took a positional id

DasX:
Closes #138437

## What Problem This Solves

- Accept an optional positional job ID for `openclaw automations runs [id]`.
- Keep the shipped leaf-local `--id <id>` form supported for existing scripts.
- Accept equal trimmed positional and flag values in either order.
- Reject blank, missing, or conflicting IDs before `cron.runs` is called, with ID validation taking precedence over `--limit` parsing.
- Update the CLI, automation, FAQ, and troubleshooting documentation to use the positional form.

## Root cause

`automations runs` was the only automation history leaf that required a leaf option for its job selector while sibling commands and operator expectations used positional IDs. The command passed `opts.id` directly to the existing Gateway RPC, so the parser rejected the positional form before the request could be made.

The repair stays at the CLI owner boundary: Commander now accepts `[id]`, the positional and compatibility flag values normalize into one canonical ID, ambiguity is rejected locally, and the existing `cron.runs` request remains unchanged.

## Compatibility

- `openclaw automations runs <job-id>` is the preferred form.
- `openclaw cron runs <job-id>` remains a documented command alias.
- `openclaw automations runs --id <job-id>` remains supported indefinitely as a leaf-local compatibility form.
- Gateway options, `--run-id`, `--limit`, and `--json` retain their existing behavior.
- No Gateway protocol, configuration, persistence, or schema contract changes.

## Evidence

Exact reviewed head: [`70fef77b5ab9020078544c886d7fbb87cd9de2ef`](https://github.com/openclaw/openclaw/commit/70fef77b5ab9020078544c886d7fbb87cd9de2ef)

- `node scripts/run-vitest.mjs src/cli/cron-cli.test.ts src/cli/program/error-output.test.ts`: 257 tests passed.
- `pnpm check:docs`: passed, including Markdown, MDX, glossary, config-fence, and internal-link checks.
- Targeted TypeScript formatting and lint checks passed.
- `git diff --check`: passed.
- Formal P1 autoreview: scoped clean, no P0/P1 findings.
- `pnpm check:changed --` passed its earlier changed-surface gates, then reached the same unrelated `tsgo:core` missing-dependency/type failures reproduced on untouched current main.

### Exact-head CLI-to-Gateway proof

[...]

PR #142264, opened 2026-09-08
fix(cli): reject blank automation job ids before calling the Gateway

steipete:
## What Problem This Solves

Eight automation subcommands (`rm`, `enable`, `disable`, `get`, `show`, `run`, `edit`, `scratch`) accepted a blank positional job id and forwarded it to the Gateway. The operator then saw the Gateway's raw params-schema error instead of the CLI's remedy, for example from `openclaw automations run "$JOB"` with an unset variable (observed live on an isolated Gateway at `1db32b4f4d6`):

```
invalid cron.run params: at /id: must not have fewer than 1 characters; must have required property 'jobId'; at root: unexpected property 'id'; must match a schema in anyOf
```

Only `automations runs` validated the id locally (`Missing job id. Pass it positionally or with --id.`), because #138438 added that check inline for one command.

## Why This Change Was Made

The job-id contract had no owner, so each command cast or passed the raw argument through. `requireCronJobId` in the cron CLI shared module now trims the id, rejects a blank one with `Missing job id.` plus the accepted form, and returns the string every command uses; `runs` delegates to it with its `--id` hint, and the ad-hoc `String(id)` casts go away. The check runs first in each action, before any other option parsing or Gateway round-trip, and surfaces through the existing `handleCronCliError` path in both human and `--json` modes.

## User Impact

A blank job id now fails fast with the CLI's own message on every automation subcommand and never reaches the Gateway. Non-blank ids behave exactly as before, including `show`'s id-or-name lookup and `run --wait`.

## Evidence

- The rejection table in `src/cli/cron-cli.test.ts` gains rows for the eight sibling commands and asserts no Gateway call at all: 8 rows fail on unmodified `main` (the mock Gateway is called with `id: "   "`), all pass with the fix.
- `node scripts/run-vitest.mjs src/cli/cron-cli.test.ts src/cli/cron-cli`: 529 + 17 tests pass.

Production LOC delta: +13 (one shared owner replaces one inline check and five casts).

### DIFF
--- a/src/cli/cron-cli.test.ts
+++ b/src/cli/cron-cli.test.ts
@@ -640,11 +640,21 @@ describe("cron cli", () => {
       args: ["cron", "runs", "job-1", "--id", "job-2"],
       message: 'Conflicting job ids: positional "job-1" and --id "job-2".',
     },
+    ...(["rm", "enable", "disable", "get", "show", "run", "scratch"] as const).map((command) => ({
+      name: `blank ${command} id`,
+      args: ["cron", command, "   "],
+      message: "Missing job id",
+    })),
+    {
+      name: "blank edit id",
+      args: ["cron", "edit", "   ", "--enable"],
+      message: "Missing job id",
+    },
   ])("rejects $name before calling the Gateway", async ({ args, message }) => {
     await expectCronCommandExit(args);
 
     expectRuntimeErrorContaining(message);
-    expect(callGatewayFromCli.mock.calls.some((call) => call[0] === "cron.runs")).toBe(false);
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5a563df55579d18b29b9b142ce781f6cdc30d306](https://github.com/openclaw/openclaw/pull/142264/commits/5a563df55579d18b29b9b142ce781f6cdc30d306)
- Landed commit: [fbc1ea2a8397ee50974e71605c81f9d30e268404](https://github.com/openclaw/openclaw/commit/fbc1ea2a8397ee50974e71605c81f9d30e268404)

PR #142272, opened 2026-09-08
refactor(tests): reuse named OAuth credential fixtures

steipete:
Related: #139428

## What Problem This Solves

Auth-profile tests repeat the same five-field OAuth credential setup, while the existing named-parameter constructor lives in a broader utility module with filesystem and auth-store dependencies.

## Why This Change Was Made

Moves the existing `oauthCred` implementation unchanged into the lightweight credential-fixture owner and migrates its two existing callers. Fifty additional input fixtures reuse it while keeping provider, access, refresh, and expiry labels visible. The old export is removed without a compatibility alias or competing constructor.

## User Impact

No production behavior changes or test cases removed. Explicit expiry expressions, field absence and order, fresh objects, optional account/email semantics, and independent expected values remain intact. The change removes 51 net test/support lines, including three consequential formatter lines from the real import migration.

## Evidence

- All five complete affected suites passed 83 cases before and after, with identical per-file ordered case names.
- Full AST expansion checked all 50 converted inputs, 39 retained clock-expression sites, eight existing calls, and the unchanged helper declaration.
- Runtime controls verified freshness, ordered keys/values, absent versus explicitly undefined optional fields, and erased type-only dependency.
- Full mapped checks, diff-scoped cleanup, and fresh full-scope P2 review passed without findings.

No runtime speedup or aggregate coverage percentage is claimed.

### DIFF
--- a/src/agents/auth-profiles.provider-refresh-lifecycle.test.ts
+++ b/src/agents/auth-profiles.provider-refresh-lifecycle.test.ts
@@ -13,11 +13,8 @@ import {
   writePlugin,
 } from "../plugins/loader.test-fixtures.js";
 import { withOpenClawTestState } from "../test-utils/openclaw-test-state.js";
-import {
-  oauthCred,
-  readAuthProfileStoreForTest,
-  storeWith,
-} from "./auth-profiles/oauth-test-utils.js";
+import { oauthCred } from "./auth-profiles/credential-fixtures.test-support.js";
+import { readAuthProfileStoreForTest, storeWith } from "./auth-profiles/oauth-test-utils.js";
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ca2384a3ece729755b63a1418d72cc4ccc9eb5d6](https://github.com/openclaw/openclaw/pull/142272/commits/ca2384a3ece729755b63a1418d72cc4ccc9eb5d6)
- Landed commit: [8a19cc21d6965e0ebb1a609c6534a40707df0a0d](https://github.com/openclaw/openclaw/commit/8a19cc21d6965e0ebb1a609c6534a40707df0a0d)

PR #142273, opened 2026-09-08
fix(ci): keep Docker upgrade seed baseline older than target

steipete:
## What Problem This Solves

Resolves a problem where Docker seed CI fails its real-upgrade case after npm's latest release reaches the target version. The updater correctly reports already-current, but the test is supposed to prove an actual upgrade before running its separate no-op control.

## Why This Change Was Made

When the published survivor lane is selected, use the existing release baseline resolver to pin a published stable version strictly older than the selected source package version. Pass the release context so frozen extended-stable lines retain their existing predecessor policy. Materialize the resolver and its version helper from the trusted workflow revision through both Git harness checkout paths. Export the exact baseline before the Docker scheduler; unrelated lanes need no registry lookup.

## User Impact

Release publication no longer changes this CI case into a same-version update or downgrade. Current main, future betas, and extended-stable targets keep the same real-upgrade and already-current assertions. No product updater behavior or package version changes.

## Evidence

- Six executed workflow-to-resolver-to-scheduler cases fail on the original floating-latest workflow and pass after the repair. They cover target 9.3, current main 9.2, two beta positions, a frozen extended-stable line, and fail-closed behavior when no predecessor exists.
- 26 existing baseline resolver tests and seven focused CI guard cases passed. Scoped changed checks and workflow validation passed.
- A live npm versions snapshot resolves 9.3 to openclaw@2026.9.2, current main 9.2 to openclaw@2026.9.1, and 9.4-beta.1 to openclaw@2026.9.3.
- Sixteen harness and sibling checks passed, including real Git materialization from both the index and a distinct workflow revision. The materialized resolver executes correctly after candidate policy files are changed; both new Linux cases fail against the original materializer.
- Exact-head CI must pass before landing, including the unchanged Docker seed survivor assertions selected by this workflow change.

### DIFF
--- a/.github/actions/git-owner/owner.py
+++ b/.github/actions/git-owner/owner.py
@@ -391,6 +391,7 @@ def checkout_selected_ref():
 def checkout_harness(sha):
     action = ".github/actions/setup-node-env/action.yml"
     evidence_scripts = ("scripts/ios-screenshot-evidence.mjs", "scripts/lib/direct-run.mjs")
+    upgrade_scripts = ("scripts/lib/release-upgrade-baseline.mjs", "scripts/lib/release-version.mjs")
     if kind == "linux-node" and not os.path.isfile(os.path.join(workspace, action)):
         raise GitFailure(1)
     harness = os.path.join(workspace, ".ci-harness")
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e888cef559f9374a08c1581c5e0c61b36f5d97fb](https://github.com/openclaw/openclaw/pull/142273/commits/e888cef559f9374a08c1581c5e0c61b36f5d97fb)
- Landed commit: [a1c1660b345559b701211ba6f9f1381cd57f360a](https://github.com/openclaw/openclaw/commit/a1c1660b345559b701211ba6f9f1381cd57f360a)

PR #142286, opened 2026-09-08
fix(gateway): retain shared state until cleanup finishes

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where Gateway shutdown or restart could close shared SQLite and reset plugin state while a slow cleanup callback was still using it. The initial timeout allowed shutdown to continue, but did not cancel the callback.

## Why This Change Was Made

The existing shutdown owners retain actual hook and runtime cleanup work through their grace periods, let independent teardown proceed, and join that work before releasing dependent shared state. Active plugin host cleanup now finishes before the shared database closes. Registry retirement still revokes runtime authority immediately.

## User Impact

Slow cleanup can finish using its existing resources without racing a closed database. A callback that never settles can hold graceful in-process shutdown open; existing forced process-exit behavior and initial grace periods are unchanged.

## Evidence

Five native SQLite regressions failed before the repair: shutdown and pre-restart hooks, harness cleanup, and plugin lifecycle callbacks and their tracked child work. All 135 focused tests passed after the repair, along with the full changed-file gate and builds.

An isolated compiled Gateway with a synthetic plugin served a successful health check, then shut down through its actual signal handler. Shutdown, harness, and lifecycle callbacks remained pending beyond their five-second grace periods. Each performed a successful late SQLite read; independent service teardown advanced; the process then exited successfully and the captured connection closed. A read-only reopen preserved the schema version and passed the integrity check. This used the existing private-QA observation build and task-owned state; it does not claim live failed-startup or external MCP/LSP service coverage. Cleanup from a registry retired by an earlier hot reload remains a separate follow-up; this slice joins the active-registry close owner.

The focused suite took 43.449 seconds, the final changed-file gate 298.109 seconds, the ordinary build 258.446 seconds, and the compiled Gateway case 20.258 seconds. Initial test-type and synthetic-fixture failures were corrected without changing assertions, trust rules, or product timeouts.


steipete:
Retaining the documented shutdown tradeoff. A grace timeout permits independent teardown to advance; it cannot make a still-running cleanup callback safe to use after its database closes. Embedded close therefore waits for actual cleanup, and the existing process-owned watchdog remains unchanged. The native and compiled proofs cover continued teardown, late SQLite use, and eventual closure.

The review found no actionable patch defect. Earlier hot-retired registry cleanup remains the separately tracked follow-up described in this PR.

steipete:
Merged via squash.

- Prepared head SHA: [d8e8bca0ec5d2cd9b51cd79d298936599ee8d6ef](https://github.com/openclaw/openclaw/pull/142286/commits/d8e8bca0ec5d2cd9b51cd79d298936599ee8d6ef)
- Landed commit: [dbe985b4e31e0bdf81ab2fecf59a266545f0f93e](https://github.com/openclaw/openclaw/commit/dbe985b4e31e0bdf81ab2fecf59a266545f0f93e)

PR #142240, opened 2026-09-08
chore(release): close out 2026.9.3 on main

steipete:
## What Problem This Solves

OpenClaw 2026.9.3 is published, while main still identifies itself as 2026.9.2 and lacks the published release notes.

## Why This Change Was Made

Align root, native source, bundled plugin and companion package version metadata with the shipped stable release. Regenerate the channel catalog and copy the exact tagged 2026.9.3 changelog section, preserving newer main code, translations, dependencies and existing changelog sections.

The 158-path patch remains identical after rebasing onto main containing the separately landed locale and Docker-baseline prerequisites (#142255, #142147, #142273). It changes no config/default, SQLite/schema/migration, protocol or persistent-store code. Existing installation minimum-host fields are preserved; generated API/peer/build metadata reproduces the published stable values.

## User Impact

Source checkouts and generated catalogs identify the actual shipped stable version. This closeout does not publish a new version or prefill another release train. Native appcast and the larger translation/tooling carry remain with their independent owners.

## Evidence

- [Published stable release](https://github.com/openclaw/openclaw/releases/tag/v2026.9.3), immutable tag commit `1391f7cd2d40ab5bbcf2f5f831d3a64f520e72d7`.
- On head `a2ec674e6af32ab0dd0391658116fa14ddd74b0b`, the real package launcher prints `OpenClaw 2026.9.3 (a2ec674)` and the macOS version metadata checker passes. Exact tagged changelog section and unchanged surrounding raw bytes verified.
- The full canonical Testbox changed check passed on the identical reviewed patch: all types, lint, ownership/architecture guards and import-cycle checks. Wrapper exit 0 after 56m52s; owned lease stopped ([backing run 34242376853](https://github.com/openclaw/openclaw/actions/runs/34242376853)). Normal delegated cleanup may cancel the backing Testbox step; wrapper completion is authoritative. Git range-diff confirms the rebased patch is unchanged.
- Fresh generated-state checking passed the version, native, config, SDK, catalog and 95 npm-lock checks. Its aggregate command reports one proven unchanged-main Control UI baseline drift: `modelProviders.defaults.discoveringMore`, `discoverFailed` and `retryDiscover`, introduced by #142105 before this branch's base. Base and head English hashes are identical; no locale file is changed here. The release owner is recovering the canonical locale refresh separately and directed this head remain pinned. This is not reported as a green aggregate generated check.
[...]

PR #142289, opened 2026-09-08
fix: translate widget script errors in Control UI locales

steipete:
## What Problem This Solves

Fixes an issue where the widget script-error label falls back to English in non-English Control UI locales. The new English key also left the strict generated-locale catalog stale. Related: #142240.

## Why This Change Was Made

Run the canonical targeted locale refresh for `board.widget.runtimeError`, then update the generated catalog metadata. A provider-free planning pass confirmed this was the only pending key in every locale before translation.

## User Impact

Widget script errors display a localized label in all 20 supported non-English locales while preserving the original error message.

## Evidence

Canonical strict locale verification passes for all 20 locales with zero fallbacks; source verification passes with 6645 keys and the unchanged raw-copy baseline. All 115,850 prior translation-memory records remain identical. The 20 new values each preserve `{message}` exactly once, and independent linguistic review found no semantic errors. `git diff --check` passes. Independent P2 review found no actionable issues.


steipete:
The canonical locale refresh in #142147 already fixed this on main at 9bd6b737930679e04e67330b323eac74794f11dd while this PR's CI was running.

Verified that landed source with the strict generated-locale check: all 20 locales are clean with zero fallbacks, and source verification passes with 6645 keys. Nineteen new translations match this PR exactly; Japanese differs only in valid colon spacing. Every locale retains `{message}` and the matching English source hash. Closing this redundant PR and preserving the landed wording.

PR #142290, opened 2026-09-08
fix(gateway): prevent read-only password connection failures

steipete:
## What Problem This Solves

Fixes an issue where users making identity-bearing, read-only Gateway calls with only a password could fail before their request was sent when saved device credentials could not be read. Concurrent SQLite WAL writes can trigger this failure even though the connection does not use a saved device token.

## Why This Change Was Made

The OpenClaw client host adapter now skips that unused lookup for unscoped, read-only, password-only connections. Signed device identity, prepared credentials, token/bootstrap routes, writable clients, and SQLite snapshot consistency and source-preservation checks retain their existing behavior.

## User Impact

Password-only read-only clients can connect without depending on unrelated saved device credentials. Connections that need those credentials still fail closed when they cannot be read, and rejected passwords still prevent method dispatch.

## Evidence

- The new WebSocket-flow regression failed on the original production code with the stored-credential read error before `connect` was sent. All 117 client tests pass with the fix, including normalized credentials and excluded authentication routes.
- A separate synthetic API probe uses the real client adapter, actual SQLite token readers, signed device identity and a real localhost WebSocket mock Gateway. With a 16 MiB database and ordinary WAL commits in another process, the baseline fails before dispatch while the fixed password-only client completes `system.info`. Required-token and denied-password controls retain their failures; persisted auth rows remain unchanged and SQLite integrity passes.
- Independent review found no actionable P0–P2 issue. The synthetic API evidence concerns the client connection boundary, not a production Gateway or every SQLite read path.

- Linux Blacksmith Testbox passed all 117 client tests and all six real-WebSocket/SQLite API cases on Node 24.19.0. In one synthetic churn run, the baseline failed before dispatch after 555 ms; the fixed password-only call completed in 24 ms with zero token-snapshot workers. Required-token and denied-password controls remained enforced. [Delegated proof run](https://github.com/openclaw/openclaw/actions/runs/34247757262): the wrapper returned exit 0 and stopped the one-shot lease after success.

### DIFF
--- a/src/gateway/client.test.ts
+++ b/src/gateway/client.test.ts
@@ -2196,6 +2196,112 @@ describe("GatewayClient connect auth payload", () => {
     client.stop();
   });
 
+  it.each([
+    { label: "default scopes", password: "shared-password", scopes: undefined, token: undefined }, // pragma: allowlist secret
+    {
+      label: "normalized credentials",
+      password: " shared-password ",
+      scopes: ["operator.read"],
+      token: "  ",
+    }, // pragma: allowlist secret
+  ])(
+    "connects read-only password auth without reading unused device credentials ($label)",
+    async ({ password, scopes, token }) => {
+      loadDeviceAuthTokenReadOnlyMock.mockImplementation(() => {
+        throw new Error("SQLite source did not stabilize for read-only inspection");
+      });
+      const onHello = vi.fn();
+      const onConnectError = vi.fn();
+      const client = createClientWithIdentity("device-1", () => {}, {
+        password,
+        token,
+        sharedStateMode: "read-only",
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7dae0cb73a8ac694ac3567fb7959afe938037ba1](https://github.com/openclaw/openclaw/pull/142290/commits/7dae0cb73a8ac694ac3567fb7959afe938037ba1)
- Landed commit: [63ff23f8659986b372b5eedb40f1d1d6bafe5a75](https://github.com/openclaw/openclaw/commit/63ff23f8659986b372b5eedb40f1d1d6bafe5a75)

PR #142305, opened 2026-09-08
fix(terminal): keep wrapped multiline cells aligned

steipete:
## What Problem This Solves

A table cell that fills its width and ends with a space before a newline could insert an extra blank row. For example, `Read \nNext` at four content columns moved `Next` below its neighboring cell's second line.

## Why This Change Was Made

Track whether the current logical line has already emitted output. A soft wrap no longer makes the following newline emit an additional empty row. Buffer handling and ANSI/OSC-8 close/reopen processing still run, preserving the intentional blank-line behavior from #142196.

## User Impact

Multiline CLI tables keep neighboring cells aligned, including colored and linked text. No option, dependency or output format is added.

## Evidence

Strengthened the existing five LF, CR, CRLF, colored and linked blank-line cases without adding test cases. All five failed before the repair and passed afterward. The full owner and sibling suites passed 190 tests; one existing Windows-only test was skipped on macOS.

Selected changed checks passed, including core and test types, lint and import-cycle checks. Independent managed P0–P2 review found no actionable issue. The patch adds five production lines and no net test lines.

Exact-head [CI](https://github.com/openclaw/openclaw/actions/runs/34248459284) passed on attempt 2. The first attempt had a profiler-test timeout and an unrelated managed-handoff restoration assertion failure. Their owner paths were unchanged in the actual tested merge; one failed-job-only retry passed without source, assertion or timeout changes. The original failures and source diagnoses remain recorded; this table repair does not claim to fix their unlocated causes.

### DIFF
--- a/packages/terminal-core/src/table.test.ts
+++ b/packages/terminal-core/src/table.test.ts
@@ -626,18 +626,18 @@ console.log(JSON.stringify({
     ["colored CRLF", "\r\n", "\x1b[31m", "\x1b[39m"],
     ["linked LF", "\n", "\x1b]8;;https://example.com/\x07", "\x1b]8;;\x07"],
   ])(
-    "preserves blank %s lines at their positions across cells",
+    "preserves blank %s lines without adding rows after wrapped spacing",
     (_label, separator, open, close) => {
       const out = renderTable({
         border: "ascii",
-        width: 30,
+        width: 15,
         columns: [
-          { key: "A", header: "A", flex: true },
-          { key: "B", header: "B", flex: true },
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a7f40f84bb637a819bdbff1fad1bc4d263c602bb](https://github.com/openclaw/openclaw/pull/142305/commits/a7f40f84bb637a819bdbff1fad1bc4d263c602bb)
- Landed commit: [f7c9adacdc5682fdce74099a827d5e1f29ec9ec2](https://github.com/openclaw/openclaw/commit/f7c9adacdc5682fdce74099a827d5e1f29ec9ec2)

PR #139929, opened 2026-09-06
fix(auto-reply): queued follow-up messages fail with "no active tool authority snapshot"

ylcn91:
Closes #139847

## What Problem This Solves

Fixes an issue where users who sent a chat message while the agent was still working on their previous message would, on 2026.9.2, get `⚠️ Something went wrong while processing your request.` as soon as the queued message started, and the message was lost rather than requeued. Gateway logs showed:

```
Embedded agent failed before reply: Reply operation has no active tool authority snapshot
```

The failure fired at model-route binding, before any provider request, so every fallback candidate that runs through a CLI backend died the same way. Reported on Telegram direct chats; the queued path is shared by every channel and by `chat.send`.

## Why This Change Was Made

**Root cause.** #134003 (commit e3ce08cd5ec, shipped in v2026.9.2) introduced `bindToolAuthoritySnapshot` / `bindToolAuthorityRoute` on the reply operation and bound the snapshot for foreground runs right after admission (`src/auto-reply/reply/agent-runner-run.ts:543`). Queued follow-up admission (`admitFollowupTurn` in `src/auto-reply/reply/followup-turn-admission.ts`) obtains its own fresh operation from `admitReplyTurn` and was never given the same binding. Two consumers then see an operation with no snapshot:

- `src/auto-reply/reply/agent-runner-cli-candidate.ts:141` calls `bindToolAuthorityRoute` unconditionally, and the registry throws the reported error for every CLI-backed candidate. This is the user-visible drop.
- `src/auto-reply/reply/agent-runner-embedded-candidate.ts:220` passes `replyOperation.toolAuthorityFingerprint`, which is `undefined`, so embedded candidates fall back to direct attempt authority and still run (which matches the reporter's log, where one hosted fallback made a real request).

**Fix.** In `admitFollowupTurn`, bind `prepareReplyToolAuthority(queued)` to the admitted operation immediately after the admitted `run` facts (session id, session file, model lock, fallback probe) are final and before the turn object is built. This is the same call and the same point in the lifecycle as foreground admission. The operation is fresh per admitted queued turn, so the once-only binding contract holds; session rotation after binding (preflight compaction, role-ordering reset) behaves exactly as it already does for foreground runs.

[...]

Issue #139847, opened 2026-09-06
[Bug]: message sent while a reply run is active is dropped — "Reply operation has no active tool authority snapshot" (regression in 2026.9.2)


motochan:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

On 2026.9.2, a chat message that arrives **while a reply run is already active for the
same session key** fails as soon as it starts. The user sees the generic
`⚠️ Something went wrong while processing your request.` and **the message is lost, not
requeued** — it has to be retyped.

The underlying error is:

```
Embedded agent failed before reply: Reply operation has no active tool authority snapshot
```

It is thrown while binding the model route, so it fires *before any provider request*.
That means **every fallback candidate dies the same way in ~25 ms** and the whole chain
collapses.

The guard is new in this release. I extracted the published tarballs for both versions
and grepped `dist/`:

| version | occurrences of `no active tool authority snapshot` in `dist/` |
|---|---|
| 2026.9.1 | **0** (6,729 JS files; `reply-run-registry-*.js` present, string absent) |
| 2026.9.2 | 1 — `dist/reply-run-registry.registry-BMo0d96L.js` |

Gateway logs agree: **63,174 journal lines before upgrading contain zero occurrences;
14 occurrences appeared in the 216 lines after.**

The guard:

```js
bindToolAuthorityRoute(route) {
  if (result || !toolAuthoritySnapshot ||
      replyRunState.activeRunsByKey.get(currentSessionKey) !== operation)
    throw new Error("Reply operation has no active tool authority snapshot");
```

All three conditions share one message, so I cannot isolate from logs alone which one
fired. The perfect correlation with "a second inbound message displaced the active run"
implicates the `activeRunsByKey.get(currentSessionKey) !== operation` clause, but that is
inference, not proof.

### Steps to reproduce

1. Run a Gateway on 2026.9.2 with a chat channel bound to a session (Telegram direct in
   my case, session key shape `agent:main:telegram:direct:<id>`).
2. Send a message that starts a **long** agent turn — long enough to still be running a
   minute later (a multi-tool investigation, a sweep, anything with several exec calls).
3. **While that turn is still running**, send a second message to the same chat.
4. Wait for the first turn to finish and reply normally.
5. The queued second message begins as a new run the instant the first completes — and
   immediately fails.

Reproduced 2/2 in my session. Both failures were messages sent mid-turn; every message
sent while the session was idle succeeded.

### Expected behavior

[...]

PR #142306, opened 2026-09-08
fix: queued follow-up messages fail while an earlier reply is running

zeroaltitude:
Related: #139847, #139929

## What Problem This Solves

Fixes an issue where users send a second message while the first reply is still running and the queued follow-up never receives a reply. The first reply can succeed while the follow-up is rejected with “no active tool authority snapshot.”

## Why This Change Was Made

Capture and bind the queued turn’s tool-authority snapshot during admission, before preflight or execution. The queued turn gets its own immutable policy snapshot; route-binding, terminal-state, and snapshot-replacement guards remain enforced. This does not bypass authority checks or reuse another turn’s authority.

### What remains different from current `main`

`main` now binds the queued snapshot immediately before execution (`followup-turn-execution.ts`, from the merged #143661). That closes the original crash, so the remaining difference is a **window**, not a missing bind: `admitFollowupTurn` registers the operation as the session’s active reply run and then runs preflight — including session compaction, which can be a multi-second, model-driven step — before execution is reached. On `main` the operation is active and *unowned* for that whole window.

Two observable consequences during that window, both reproduced below **under `queue.mode: "steer"`, OpenClaw’s default mode** (`docs/concepts/queue.md`, “Defaults”):

1. **A same-authority follow-up never reaches the steer path.** `agent-runner-run.ts` compares the incoming message’s authority fingerprint against `operation.toolAuthorityFingerprint`. With the active operation unowned, the comparison is `undefined !== <fingerprint>`, so `shouldQueueAuthorityMismatch` is true and the message is never handed to `runActiveReplySteer` (“has different or unknown tool authority; queuing instead of steering”). On this branch the same message *is* handed to that path and parked as a steer candidate, which is where same-turn delivery and its overflow protection become possible at all. **Scope note:** the harnesses below prove this routing difference and what follows from it through execution. Neither carries a steer through to a *completed injection* — see the scope paragraph under the execution evidence for why that is not a difference between the two trees.
2. **The message that arrives in that window loses queue-overflow protection, and can therefore be dropped before it ever executes.** A steer candidate is parked with `steerAnchor`, and `applyQueueDropPolicy`'s `isProtected` predicate skips `steerAnchor` items when choosing overflow victims. On `main` the same message is an ordinary queued item and, being the oldest, is the first victim of `drop: old`. The execution evidence below shows it never reaching the model at all.
3. **The operation is unowned for the window.** `bindToolAuthorityRoute` throws `Reply operation has no active tool authority snapshot` whenever no snapshot is bound. To be precise about what that is and is not: every production caller of that method runs after `setPhase("running")`, so on current `main` this is a property of the unowned window rather than a crash a user hits today — the original crash is closed by #143661. It is stated here as the reason admission-time ownership is the invariant, not as a reproduced failure.

Both consequences are scoped to `steer` mode by construction, because `shouldSteer` is computed only for `resolvedQueue.mode === "steer"` (`get-reply-run-admission.ts:562-570`). Under `followup` or `collect` no message is ever routed into the steer path on either tree, so no `steerAnchor` exists and this overflow difference does not arise there. The admission-time ownership the patch introduces is unconditional; the *measured* benefit below is a `steer`-mode benefit, and it is claimed only as that.

Moving the capture to admission makes the admitted turn its own authority owner for the whole of its life, rather than only from execution onward. Execution keeps that same owner and no longer rebinds.

### Overlapping work — please compare before merging

#139929 independently addresses the same admission-owned invariant. We maintain this implementation in our deployed integration branch and are retaining our own branch until a suitable fix lands on upstream main, rather than cherry-picking another contributor’s unmerged PR.

**Maintainers: please compare this PR with #139929 before merging either one**, including admission/preflight ordering, immutable policy capture, terminal rejection, and attempts to replace an already-bound snapshot. These are alternative fixes, not two changes that need to land together. If the other implementation lands first, we will compare the landed behavior and retire this PR when it covers our fix. This note is not a claim that the other PR is deficient.

## User Impact

Queued messages execute under their own admitted authority instead of being silently lost after the earlier reply. With the default `steer` queue mode, a follow-up that arrives while a queued turn is still preparing is routed into the steering path for that turn rather than being split off as a second turn outright, and it keeps the overflow protection that comes with being a steer candidate. In the measured burst below, that is the difference between the user's message being answered and being discarded unanswered. Existing rejection behavior for invalid or retired operations is preserved.

## Evidence

Validated head: `3b79869af31c4cadc10055d1e92f2254405133c7`.

### Queue mode and admission inputs are production-derived (Revision 9's [P2])

Revision 9 was right that the previous revision of both harnesses configured `resolvedQueue.mode: "followup"` while forcing `shouldSteer: true` and `queueAdmissionState: "steering"` — an input pair the configured mode cannot produce, since production computes `shouldSteer` only for `mode === "steer"` with a queue admission state other than `ready` (`get-reply-run-admission.ts:562-570`). That is fixed at the source of the inputs rather than by relabelling the claim:

- **`resolvedQueue` is no longer written by the harness.** Both harnesses now call production's own `resolveQueueSettingsCore` over their config, the way `get-reply-run-admission.ts:358` calls it. Neither config sets a queue mode, so the resolver supplies OpenClaw's documented default — `steer`. A precondition asserts the resolved mode *is* `steer` and states why it matters, so a future config change cannot silently move the measurement into a mode where the mechanism does not exist.
- **`queueAdmissionState` is derived from the real queue before every inbound message**, by production's own `resolveReplyQueueAdmissionState` applied to the live follow-up queue and the live active reply operation — the same call `get-reply-run-admission.ts:557` makes.
- **`shouldSteer` / `shouldFollowup` are production's conjunctions over that state.** The conjuncts this fixture holds constant are named at the call site and are each a property of the fixture rather than an override: an ordinary CLI user message is not a room event and not a heartbeat, nothing triggers a reset or preempts a heartbeat, no recovery owner is registered for the store, and `resolveActiveRunAcceptsCurrentThread` returns `true` for any route that is not a Slack direct-routed thread turn (`:474-480`).
- **`cap`, `drop` and the debounce come from `messages.queue` config** read through that same resolver. The execution harness sets `cap: 2`, `drop: "old"`; the admission harness sets neither and resolves the real defaults `cap: 20`, `drop: "summarize"`. A small cap only decides how many messages are needed to reach overflow; `drop: "old"` makes the loss observable as a discarded message rather than a summary line. `applyQueueDropPolicy` consults the same `isProtected` predicate under every drop policy (`queue/enqueue.ts:238`).

The derived state is now printed for every inbound message, and it shows FIFO eligibility being **honoured, not bypassed**. On this branch the window follow-up is admitted at state `empty` (steering computed, and taken), the next burst message at `steering`, and the last two at `ready` — where production stops computing steering because other waiting messages hold FIFO priority. On `main` the same derivation yields `ready` one message earlier, because the unsteered window follow-up is sitting in the queue as an ordinary backlog item. Nothing in either run forces a flag.

### Real-runtime proof of fresh queued admission

`scripts/proof-142306-queued-admission-authority.ts` — run with `pnpm tsx scripts/proof-142306-queued-admission-authority.ts`.

The harness drives production code, not mocks. `admitFollowupTurn` runs unmodified, including the real `admitReplyTurn` lane admission, the real reply-run registry, the real send-policy resolution and the real preflight compaction check. The steering-versus-queueing decision is made by the unmodified `runReplyAgent` — the same production function a live inbound message goes through. Fingerprints come from the real `reply-tool-authority` functions, and the observation is real follow-up-queue state. There is no vitest, no module mock and no model call: the admitted turn is deliberately held in the post-admission, pre-execution window that this change moves the binding into.

**18 assertions, EXIT 0, three consecutive runs on this head:**

```
queue settings resolved by production: mode=steer cap=20 drop=summarize debounceMs=0
  ok  production resolved the default queue mode `steer`, the only mode that computes steering

[1] fresh queued admission owns tool authority before execution
  ok  real admitFollowupTurn admitted the queued turn
  ok  the admitted operation is the session's active reply run
  ok  the admitted turn has NOT started executing
  ok  admitted queued turn owns a tool authority fingerprint before execution
  ok  an identical inbound message computes the SAME fingerprint the admitted turn owns
  ok  a different-authority inbound message computes a DIFFERENT fingerprint
  ok  execution's route bind resolves against the admitted snapshot

[2] a same-authority follow-up arriving in that window is ROUTED INTO the steer path
  ok  the real registry reports the session active during the window
  admission inputs for proof-msg-2 (derived from the live queue): state=empty shouldSteer=true isActive=true
  ok  the same-authority inbound turn ran to completion without error
  ok  the same-authority follow-up reached the real follow-up queue
  ok  the same-authority follow-up was ROUTED INTO the steer path for the in-flight turn
  ok  that steer then falls back here, because the held turn has no injectable backend

[3] discrimination control — a different-authority follow-up is NOT steered
  admission inputs for proof-msg-3 (derived from the live queue): state=steering shouldSteer=true isActive=true
  ok  the different-authority inbound turn ran to completion without error
  ok  the different-authority follow-up reached the real follow-up queue
  ok  the different-authority follow-up is enqueued as its own turn, NOT routed into the steer path

[4] the admitted snapshot is immutable for the rest of the turn
  ok  a later bind cannot replace the turn's admitted tool authority
  ok  the admitted fingerprint survived the replacement attempt

All runtime assertions passed. (18 checks)
```

Note scenario 3: production computed `shouldSteer: true` for that message from the real queue state, and it was still not steered — the authority guard, not a harness input, is what routed it to its own turn.

### Failing control for that harness against current `main`

The same harness, byte-identical, run in a pristine worktree of the merge base (`38c3edcc4ca`, which carries `main`’s execution-time binding). **EXIT 1 on three consecutive runs, with the same six failures (failure lists byte-identical across runs):**

```
  FAIL admitted queued turn owns a tool authority fingerprint before execution — toolAuthorityFingerprint=undefined
  FAIL an identical inbound message computes the SAME fingerprint the admitted turn owns
  FAIL execution's route bind resolves against the admitted snapshot — routeBindError=Reply operation has no active tool authority snapshot
  FAIL the same-authority follow-up was ROUTED INTO the steer path for the in-flight turn — steerAnchor is absent, so production skipped runActiveReplySteer and enqueued the message as its own turn
  FAIL a later bind cannot replace the turn's admitted tool authority — rejection=(none)
  FAIL the admitted fingerprint survived the replacement attempt — admitted=undefined

PROOF FAILED after 18 checks: Error: 6 of 18 runtime assertions failed
```

Reading those six against the ask:

- the fourth failure is the **routing difference** described above, observed on real queue state: on `main` the message never reaches `runActiveReplySteer` at all — and it is reached with the *same* derived admission inputs as on this branch (`state=empty shouldSteer=true`), so the difference is the authority comparison, not the inputs;
- the third shows the operation is **unowned**: the harness calls `bindToolAuthorityRoute` directly and gets the literal error from this PR's problem statement. No production caller reaches that method during the window — every one of them runs after `setPhase("running")` — so read this as evidence of the unowned window, not as a crash a user hits on current `main`;
- the fifth and sixth show the window has **no owner** on `main` — the first snapshot to arrive defines the turn’s authority, whereas on this branch admission has already claimed it and a later bind is refused.

Scenario 3 passes on `main`, scenario 1’s first three assertions pass there too, and 12 of the 18 assertions pass overall — so the control fails for the specific reasons claimed rather than failing uniformly.

### After-fix execution evidence

`scripts/proof-142306-queued-execution-benefit.ts` — run with `pnpm tsx scripts/proof-142306-queued-execution-benefit.ts`.

Revision 8 asked for evidence that goes past admission and routing into **queued execution**, plus the remaining benefit over current `main`. This companion harness runs the whole lifecycle for real, and now does so under production-derived admission inputs.

- It drives the real `admitFollowupTurn` + `executeFollowupTurn` pair — the exact two steps, in order, that production's own drain owner `createFollowupRunner` performs for every queued turn. They are called separately for one reason: the window this PR changes lives *between* them, and racing a millisecond-wide window from inside the runner would not be deterministic.
- The queued turn really executes: real model resolution, real prompt build, real attempt dispatch, real `attachBackend`, real streaming transport. It reaches `phase === "running"` **with an injectable backend**, asserted from production registry state while the run is in flight, and settles `ok`.
- The **model endpoint is the only edge stub**: a `node:http` server on `127.0.0.1` speaking the OpenAI Responses streaming format, reached through an ordinary `models.providers` entry — the same shape `src/gateway/test-openai-responses-model.ts` uses for gateway HTTP tests. No credential, no external network. Its request log is the record of which user prompts the agent runtime actually executed.
- After that turn completes, the **real follow-up drain** (`createFollowupRunner`, registered by production as the queue's executor) runs the surviving queued messages.

The scenario is one admitted turn plus a four-message burst arriving during its admission window, with `messages.queue` set to `cap: 2`, `drop: "old"` and resolved through production's settings resolver, which supplies the default mode `steer`.

**17 assertions, EXIT 0, three consecutive runs on this head:**

```
[0] production resolves the queue settings this run is measured under
queue settings resolved by production: mode=steer cap=2 drop=old debounceMs=0
  ok  production resolved the default queue mode `steer`, the only mode that computes steering

[1] a queued turn is admitted and owns its tool authority before execution
  ok  real admitFollowupTurn admitted the queued turn
  ok  the admitted operation is the session's active reply run
  ok  the admitted turn has NOT started executing yet
  ok  the admitted queued turn owns a tool authority fingerprint before execution

[2] a same-authority follow-up arriving in that window reaches the steer path
  admission inputs for proof-exec-2 (derived from the live queue): state=empty shouldSteer=true isActive=true
  ok  the window follow-up's inbound turn ran to completion without error
  ok  the window follow-up reached the real follow-up queue
  ok  the window follow-up was ROUTED INTO the steer path and parked with steerAnchor
  ok  that parked steer then falls back, because the admitted turn is not running yet

[3] a burst overflows the queue during the same window
  admission inputs for proof-exec-3 (derived from the live queue): state=steering shouldSteer=true isActive=true
  admission inputs for proof-exec-4 (derived from the live queue): state=ready shouldSteer=false isActive=true
  admission inputs for proof-exec-5 (derived from the live queue): state=ready shouldSteer=false isActive=true
  ok  every later inbound turn ran to completion without error
  queue settings in effect: cap=2 dropPolicy=old inFlight=1
  queue after the burst: [proof-exec-2, proof-exec-4, proof-exec-5]
  queue dispositions: {"proof-exec-3":"queue-cap-old"}
  ok  the follow-up that arrived during the preflight window SURVIVED queue overflow
  ok  overflow evicted a later, unprotected message instead

[4] the admitted turn EXECUTES for real against the agent runtime
  ok  the queued turn executed and settled successfully
  ok  the provider really received the admitted turn's prompt
  ok  the executing turn reached running WITH an injectable backend attached

[5] the real drain then EXECUTES the surviving queued messages
  ok  the window follow-up was EXECUTED by the real follow-up drain
  ok  a later surviving message was also executed (the drain works on both trees)

provider requests observed: 4
  executed   PROOF142306_ADMITTED_TURN
  executed   PROOF142306_WINDOW_FOLLOWUP
  NOT run    PROOF142306_LATER_SECOND
  executed   PROOF142306_LATER_THIRD
  executed   PROOF142306_LATER_FOURTH

All runtime assertions passed. (17 checks)
```

**The same harness, byte-identical, against pristine `origin/main` (`38c3edcc4ca`). EXIT 1 on three consecutive runs, the same four failures (failure lists byte-identical across runs):**

```
[0] production resolves the queue settings this run is measured under
queue settings resolved by production: mode=steer cap=2 drop=old debounceMs=0

  admission inputs for proof-exec-2 (derived from the live queue): state=empty shouldSteer=true isActive=true
  FAIL the window follow-up was ROUTED INTO the steer path and parked with steerAnchor — steerAnchor is absent, so production skipped runActiveReplySteer and enqueued the message as a plain queued turn

  admission inputs for proof-exec-3 (derived from the live queue): state=ready shouldSteer=false isActive=true
  queue settings in effect: cap=2 dropPolicy=old inFlight=0
  queue after the burst: [proof-exec-4, proof-exec-5]
  queue dispositions: {"proof-exec-2":"queue-cap-old","proof-exec-3":"queue-cap-old"}
  FAIL the follow-up that arrived during the preflight window SURVIVED queue overflow — it was evicted with disposition=queue-cap-old

  FAIL the window follow-up was EXECUTED by the real follow-up drain — the provider never received its prompt, so the user's message was never answered

provider requests observed: 3
  executed   PROOF142306_ADMITTED_TURN
  NOT run    PROOF142306_WINDOW_FOLLOWUP
  NOT run    PROOF142306_LATER_SECOND
  executed   PROOF142306_LATER_THIRD
  executed   PROOF142306_LATER_FOURTH

PROOF FAILED after 17 checks: Error: 4 of 17 runtime assertions failed
Failed assertions:
  - the admitted queued turn owns a tool authority fingerprint before execution
  - the window follow-up was ROUTED INTO the steer path and parked with steerAnchor
  - the follow-up that arrived during the preflight window SURVIVED queue overflow
  - the window follow-up was EXECUTED by the real follow-up drain
```

**The observable benefit, stated exactly.** With `queue.mode: "steer"` (the default) and the identical four-message burst, this branch executes the follow-up that arrived during the preflight window and evicts one later message; `main` evicts that follow-up *and* a later one, and the provider never receives its prompt. On both trees the admitted turn executes and a later message executes, so 13 of the 17 assertions still pass on `main` — the control discriminates rather than failing uniformly.

**Scope, stated plainly.** This harness does **not** show a successful steer *injection* that `main` lacks, and does not claim one. `resolveReplyMessageInjectionRejection` rejects with `not_running` before it ever compares authority, so an injection needs `phase === "running"` with an attached backend — and by the time a queued turn is running, both trees own a snapshot. The difference this PR creates is confined to the admission→running window, and that window is what is measured above. The parked steer falling back is asserted rather than glossed, on both trees.

### Gates on this head

- Admission harness: 18/18, EXIT 0 (×3). Its control: EXIT 1 (×3), 12/18 passing.
- Execution harness: 17/17, EXIT 0 (×3). Its control: EXIT 1 (×3), 13/17 passing. Every run on both sides was taken from the final, post-lint, post-format file; the four files were `md5sum`-compared across the two worktrees before the runs.
- `followup-turn-admission{,.authority}` and `followup-turn-execution{,.steering,.lifecycle}` suites: 5 files requested / 5 executed, **92 passed, EXIT 0** (53 literal cases; the two `.each` files expand to 37 and 45). No test file changed this round, and the count is unchanged from the previous head.
- `pnpm tsgo:scripts` EXIT 0; scripts oxlint lane (`run-oxlint.mjs --tsconfig config/tsconfig/oxlint.scripts.json scripts`) EXIT 0, probe-verified live by appending a deliberate violation and confirming the lane named this exact file before restoring it; `pnpm format:check` EXIT 0 over 38,614 files; `pnpm check:no-conflict-markers` EXIT 0; `git diff --check` clean.
- `pnpm deadcode:unused-files`: 0 entries on both the production and the full-tree scan, with both harnesses registered in `config/knip.config.ts`.
- This round changes **no production file** and adds no file — the diff is harness inputs in the two existing proof scripts.

### Earlier evidence, retained

- 39 focused admission tests passed across two files on the earlier refreshed candidate `afb500b063cd`.
- Original repair: the new admission regressions failed before the fix; 186 affected tests passed afterward.
- Original September 5 live before/after observation used Gateway WebChat RPC to submit a second input while the first reply was active: before deployment only the first reply arrived, after deployment both arrived in order. That was the original repair deployment, not an isolated runtime proof of the current head, and it did not exercise direct Discord/Telegram delivery.

### CI state

**At `3b79869af31`: settled with 0 failures — 227 checks, 176 success, 50 skipped, 1 neutral (CodeQL).** `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. The total creeps upward by one on each body edit because `PR context and evidence` re-triggers on `edited`; every one of those runs has also been `success`. Reads use `--paginate` over `?per_page=100`, because a single-page read on this PR under-reports failures.

The previous head `04cb0574bdc` also settled with 0 failures (225 checks at first settlement: 177 success, 47 skipped, 1 neutral).

No credentials, private configuration, model identifiers, session identifiers, or raw transcripts are included.

*Written by Tank (Claude), an AI agent operated by Eddie Abrams (zeroaltitude).*


DasX:
I ran this branch against a live `claude-cli` backend to see whether the change
is observable at runtime. Reporting both what it showed and what it did not.

**Environment**

- Linux 6.8, Node 24.20.0, commit `afb500b063c` (unchanged, see below).
- Backend `claude-cli`, subscription auth, model `claude-cli/claude-fable-5-1`.
- Harness: `extensions/qa-lab` gateway child on the QA channel transport, driven
  by a new e2e test that sends two overlapping inputs, the second one while the
  first CLI turn is still running.

**Test-environment adjustment, disclosed**

The e2e worker replaces `HOME` with a throwaway directory
(`test/test-env.ts`, `fs.mkdtempSync(..., "openclaw-test-home-")`). The QA
harness then forwards "the host HOME" to the CLI child
(`extensions/qa-lab/src/providers/env.ts`, `HOME: hostHome`), but by that point
the host HOME is already the throwaway one, so `claude` reports
`Not logged in - Please run /login`. Running with `OPENCLAW_LIVE_TEST=1` makes
`stageLiveTestState()` copy the existing Claude credentials into that throwaway
home, which is the path this repo already provides. No product code was touched
and `OPENCLAW_LIVE_USE_REAL_HOME` was deliberately not set.

**Implementation under test**

Raw command output from the run itself, not a hand-written header:

```
$ git rev-parse HEAD
afb500b063cd2fb1d647a4b9ba260738b01eb26f
$ git status --porcelain
?? test/e2e/qa-lab/runtime/pr142306-queued-followup-cli.e2e.test.ts
$ git diff --stat HEAD
```

The only untracked file is my own e2e driver; `git diff --stat HEAD` is empty,
so no tracked file of this branch was modified.

**What the runtime run shows**

On `afb500b063c`, both overlapping inputs are answered. Two separate CLI
executions, not a fallback ladder: the second one resumes the session the first
one created, and it starts only after the first turn has reported its result.
The two inputs differ in length (67 vs 292 characters), which the two prompt
sizes reflect.

```
09:10:55.559 cli exec: provider=claude-cli model=claude-fable-5-1 promptChars=331 trigger=user useResume=false session=none resumeSession=none reuse=none
09:10:59.306 cli turn: provider=claude-cli model=claude-fable-5-1 durationMs=3748 outBytes=17 outHash=880c8c27afea
09:10:59.864 cli exec: provider=claude-cli model=claude-fable-5-1 promptChars=554 trigger=user useResume=true session=present resumeSession=daeb3cdc20e1 reuse=reusable
09:11:02.250 cli turn: provider=claude-cli model=claude-fable-5-1 durationMs=2386 outBytes=18 outHash=a599777dfe1e
```

Both turns are in the log with distinct output hashes and matching sizes:
17 bytes is `PR142306_FIRST_OK`, 18 bytes is `PR142306_SECOND_OK`. The test's own
outbound record agrees:

```
"firstReplyMatched": true, "secondReplyMatched": true, "cliTurnLines": 2, "failure": null
outbound: PR142306_FIRST_OK, PR142306_SECOND_OK
```

The second input was sent at 09:10:57.255, i.e. 2.05 s before the first turn
reported its result, so it was queued while that turn was still running.

**What it does not show, and this is the point**

The same scenario also passes with the fix reverted. I reset
`src/auto-reply/reply/followup-turn-admission.ts` to its `main` state (dropping
`operation.bindToolAuthoritySnapshot(prepareReplyToolAuthority(turn.queued))`),
rebuilt, and re-ran:

```
"firstReplyMatched": true, "secondReplyMatched": true, "failure": null
```

I also tried the variant where the queued input comes from a different sender,
because that is what the unit test pins down. It passes without the fix as well;
`Reply operation has no active tool authority snapshot` never appears. (Measured,
but not captured in a separate attached transcript.)

The reason looks structural: at runtime the queued follow-up inherits the running
reply operation, which already has its snapshot bound through the normal
admission path, so the missing call is not observable.
`bindToolAuthorityRoute` is reached on every CLI candidate
(`agent-runner-cli-candidate.ts:141`), but its guard in
`reply-run-registry.operation.ts:365` does not trip. The failing state needs the
queued turn to land on a *fresh* operation, which
`followup-turn-admission.authority.test.ts` constructs directly
(`expect(operation).not.toBe(first)` after `first.complete()` fires mid-admission).
That is a race between turn completion and admission, and I could not hit it
from outside with fixed delays.

**Where the change is observable**

The unit test that ships with this PR does discriminate, in both directions:

With the fix line reverted to `main`:

```
FAIL  |auto-reply| src/auto-reply/reply/followup-turn-admission.authority.test.ts
Error: Reply operation has no active tool authority snapshot
 at Object.bindToolAuthorityRoute src/auto-reply/reply/reply-run-registry.operation.ts:365:15
Test Files  1 failed (1)
     Tests  2 failed (2)
```

On `afb500b063c` unchanged:

```
✓ initializes the queued turn's own authority after the first reply completes  50ms
✓ freezes settled queued policy before preflight and later callback mutations   6ms
Test Files  1 passed (1)
     Tests  2 passed (2)
```

Command in both cases:
`node scripts/run-vitest.mjs src/auto-reply/reply/followup-turn-admission.authority.test.ts`

**Summary**

The live CLI backend is healthy on this branch and queued follow-ups are answered.
I could not produce a message-channel scenario that separates the patched from the
unpatched build, so I am not offering the passing e2e run as evidence for the fix.
The shipped unit test is currently the only measurement I have that does separate
them.

DasX:
@zeroaltitude In case it helps move this along, a pointer about where the proof needs to sit.

I posted runtime evidence for this branch in [a comment above](https://github.com/openclaw/openclaw/pull/142306#issuecomment-5614655671) on September 10: a live `claude-cli` backend run on `afb500b063c`, two overlapping inputs with visibly different prompt sizes (331 vs 554 `promptChars`), two `cli exec` lines where the second resumes the session the first created, and both `cli turn` lines with distinct output hashes. The PR's own unit test fails on the pre-fix line and passes with it. `git rev-parse HEAD` and `git status --porcelain` are quoted as raw command output so the tested tree is identifiable.

ClawSweeper's review says the proof has to go in the **PR body** for the automatic re-review to pick it up, and that `@clawsweeper re-review` only works for the PR author or someone with repository write access. Both are out of reach for me as an outside contributor, which is presumably why the `needs proof` label is still on despite the evidence being here.

Feel free to lift anything useful from that comment into the description, verbatim or trimmed. I am happy to reshape it into whatever format suits the body better, or to re-run the measurement against a newer head if you push one.

One limitation I want to keep visible rather than bury: the QA channel test passes both with and without the fix, so it is not the discriminator. The PR's own unit test is. I said so in the comment too.

kiranvk-2011:
Runtime data point from a self-hosted fleet: **on current main this change collides with #143661 and the queued message is lost.**

### Environment

- OpenClaw 2026.9.4 built from upstream main `23e6b1b30` (2026-09-12) with this PR's single commit (`218c49aef`, rebased cleanly) cherry-picked on top; Linux x64, Node v24.19.0, Docker, single gateway, multiple agents.
- Embedded runtime, `github-copilot/claude-opus-4.8`. No other change to `src/auto-reply`.

### Reproduction (Gateway WS `chat.send`, one session key, `/queue collect`)

1. `/queue collect`
2. Message A: an `exec` turn that runs `sleep 45` and then replies `DONE-A`
3. Message B, sent 12–25 s into A's run: "reply with exactly DONE-B"

**With this PR:** A completes (`DONE-A`). B never enters the transcript and never gets a reply. The gateway logs, every ~500 ms until a restart:

```
followup queue drain failed for agent:opus46:<session>: Error: Reply operation cannot change tool authority after admission
```

Reproduced twice (two fresh sessions). No `run-failed-before-reply` event, no user-visible error — silent loss.

**Without this PR, same base (`96524b732`), same test:** A completes, B is delivered as `[Queued messages while agent was busy]` and answered (`DONE-B`), zero tool-authority lines.

### Why

`bffda1f94521` (#143661, merged 2026-09-10 23:05Z) added an execution-time bind for queued turns in `src/auto-reply/reply/followup-turn-execution.ts` (`turn.operation.bindToolAuthoritySnapshot(prepareReplyToolAuthority(turn.queued))`, currently line 359). This PR binds a *second* `prepareReplyToolAuthority(turn.queued)` object at admission. `bindToolAuthoritySnapshot` refuses a different object once one is bound, so the execution-time bind throws and the drain retries forever.

So on main ≥ `bffda1f9` this PR needs either a rebase that drops one of the two binds, or closing as superseded by #143661 if that change is confirmed to cover the queued path (it does for the WS collect case above). Happy to re-run the test on a refreshed head.

zeroaltitude:
@kiranvk-2011 thank you — that is a careful report with a clean reproduction and a correct mechanism, and the silent-loss failure mode you describe is exactly the kind of thing worth catching before merge. I chased it down against the current head and the conclusion is that **the collision is real for the artifact you tested, but is not present on this PR's current head**. Details, so you can check my work.

### What you tested vs. what is on the branch now

You cherry-picked this PR's single commit `218c49aef` onto upstream main `23e6b1b30`. That commit predates #143661 (`bffda1f9452`, merged 2026-09-10). At the time it was written there was no execution-time bind, so it only *added* one at admission. Cherry-picking it onto a post-#143661 main therefore produces **two** `bindToolAuthoritySnapshot` calls against the same operation — and, as you correctly identified, `bindToolAuthoritySnapshot` refuses a second, different snapshot object once one is bound. The drain then throws every retry, which is the `Reply operation cannot change tool authority after admission` loop you captured. Your diagnosis of the mechanism is right.

What the branch has since done is merge `origin/main` (which contains #143661) at `21d5d3932f5452c405ca4acf46ca0abf4e1f020a`, and that merge resolved precisely this overlap by **removing** the execution-time bind rather than keeping both:

```diff
 # src/auto-reply/reply/followup-turn-execution.ts
-import { prepareReplyToolAuthority } from "./reply-tool-authority.js";
@@
     try {
-      turn.operation.bindToolAuthoritySnapshot(prepareReplyToolAuthority(turn.queued));
+      // Admission froze authority before preflight; execution keeps that same owner.
       turn.operation.setPhase("running");
```

So on the current head there is exactly one production bind in the follow-up path, at `src/auto-reply/reply/followup-turn-admission.ts:304`:

```ts
operation.bindToolAuthoritySnapshot(prepareReplyToolAuthority(turn.queued));
```

`git grep bindToolAuthoritySnapshot -- src/auto-reply/reply/followup-turn-execution.ts` returns nothing on this branch and returns line 359 on `main`. The net effect of the PR against current main is to **move** the bind from execution time to admission time, not to add a second one.

### What this does and does not settle

- It settles your reproduction: a re-run against `21d5d3932` rather than a cherry-pick of `218c49aef` should not produce the tool-authority loop or the lost `DONE-B`. If you have the fleet capacity, that re-run would be genuinely useful — you offered, and I would rather have your independent result than my source reading.
- It does **not** settle the open review question, which is a fair one and still outstanding: now that main binds at execution time, what is the remaining runtime benefit of freezing authority *before* preflight? That is the gap the current review is holding this PR on, and it is the right thing to hold it on.
- One caveat on the artifact you built: anyone else cherry-picking the single commit onto a post-#143661 base will hit exactly what you hit. That is a property of the cherry-pick, not of the branch, but it is worth knowing.

Thanks again — a runtime report with logs, a bisect-quality attribution to `bffda1f94521`, and an offer to re-test is a lot more than most bug reports carry.

— written by Tank/Claude

zeroaltitude:
## Round: naming the exact window in which this PR's bind can differ from main's

Head unchanged at `21d5d3932f5452c405ca4acf46ca0abf4e1f020a`. No code change in this round — this answers the open **[P1]** directly:

> The runtime benefit of capturing permissions before preflight, beyond main's execution-time binding, remains unverified.

That is a fair hold, and I did not want to answer it with a proof harness whose result would be ambiguous. So I traced the mechanism instead, and the divergence window turns out to be exactly one case, which I can name precisely.

### `prepareReplyToolAuthority` is a pure function of `turn.queued`

Both bind sites call `prepareReplyToolAuthority(turn.queued)` on the same object reference. Nothing in `reply-tool-authority.ts` reads live config, the session store, or any ambient singleton — its input type is a `Pick` over `FollowupRun["run"]` plus `originatingChannel` / `toolsAllow` / `disableTools`. So moving the call earlier changes the computed snapshot **only if `turn.queued` changes in between**.

### `turn.queued` is reassigned in exactly two places between the two bind sites

Both live in `followup-turn-admission.ts`:

1. `refreshTurnSessionState` replaces `currentInboundContext`. That field is **not** in the authority input's pick list, so it cannot affect the snapshot.
2. `synchronizeTurnGeneration` replaces `run.sessionId`, `run.sessionFile`, `run.cliSessionBindingFacts`, `run.autoFallbackPrimaryProbe`, and `run.modelSelectionLocked` — but only when `isSameSessionGeneration(entry, previousEntry)` is false, i.e. when the session generation rotated during preflight.

Of those five, the authority input picks `sessionId` and `sessionFile` (they are in the second `Pick` alongside `workspaceDir`, `provider`, `model`). The other three are not read.

Both mutators are invoked after preflight compaction, at admission lines 430–431 (success path) and 443–444 (failure path) — strictly between this PR's bind and main's.

### So the whole behavioral delta is:

**When the session generation rotates during preflight, `main` binds tool authority computed against the post-rotation `sessionId`/`sessionFile`, and this branch binds it against the values the turn was admitted under. In every other case the two snapshots are identical by construction.**

That is a real difference, and it is narrow enough to state without a harness. What I am deliberately *not* doing is asserting which one is correct, because that is a contract decision rather than a measurement:

- If the snapshot is meant to record **the authority the admission decision was made against** — an audit/approval identity — then freezing at admission is right, and main currently lets a mid-preflight rotation rewrite the recorded identity after the fact.
- If the snapshot is meant to describe **the session the turn will actually execute in**, then main is right, and this PR would pin a stale `sessionId`/`sessionFile` for exactly the rotated case.

I can build the real-runtime proof for whichever reading you want — drive real `admitFollowupTurn` + `executeFollowupTurn` with a compaction that rotates the generation, and read the bound snapshot on both trees — but proving it before the intended contract is settled would just pin whichever answer I guessed.

**Maintainer question, in one line:** should a session-generation rotation during preflight be allowed to change a queued turn's bound tool-authority identity, or should the turn keep the authority it was admitted under?

### Also closed out this round

@kiranvk-2011's collision report is answered above and the mechanism is confirmed fixed on this head: the branch's `origin/main` merge resolved the double-bind by removing the execution-time call rather than keeping both, so there is exactly one production bind. `git grep bindToolAuthoritySnapshot -- src/auto-reply/reply/followup-turn-execution.ts` is empty here and non-empty on `main`.

CI at this head has three red lanes and none of them are branch-owned:

- `check-lint-core-4` — `src/infra/update-managed-service-handoff-boundary.test-support.ts` over the `max-lines` ceiling. Not in this PR's changed-file list.
- `checks-node-compact-large-29` — `src/gateway/server-methods/sessions.send-yield-resume.test.ts` (`residual registry roots: expected 1 to be +0`). Not in this PR's changed-file list.
- `openclaw/ci-gate` — the rollup of those two.

*Written by Tank (Claude), an AI agent operated by Eddie Abrams (zeroaltitude).*

zeroaltitude:
## Round: real-runtime proof of fresh queued admission (head `41ca2c1ce10`)

Revision 6 reported **Findings: None** with patch quality 🐚 platinum hermit, so this round adds no production change. It closes the three remaining "Before merge" items, all of which were the same gap seen from different angles.

**[P2] "Add runtime proof of fresh queued admission through the revised snapshot lifecycle."**
Added `scripts/proof-142306-queued-admission-authority.ts` (commit `41ca2c1ce10`, GPG-signed; the only other change is its one-line Knip registration). It drives production code rather than mocks: the unmodified `admitFollowupTurn` — real `admitReplyTurn` lane admission, real reply-run registry, real send-policy resolution, real preflight compaction check — and then the unmodified `runReplyAgent`, which is the production function that decides whether a message arriving during an active run is steered into it or queued as its own turn. Fingerprints come from the real `reply-tool-authority` functions and the observation is real follow-up-queue state. No vitest, no module mock, no model call: the admitted turn is held in the post-admission, pre-execution window that this change moves the binding into. **14 assertions, EXIT 0 on four consecutive runs**; full terminal output is in the PR body.

**[P1 merge risk] "The runtime benefit of capturing permissions before preflight, beyond main's execution-time binding, remains unverified."**
The benefit is now named and measured. `main` binds immediately before execution, so what remains is a *window*: `admitFollowupTurn` registers the operation as the session's active reply run and then runs preflight — including session compaction — before execution is reached, and for that whole window the active operation is unowned on `main`. The same harness, byte-identical, in a pristine worktree of the merge base `38c3edcc4ca` gives **EXIT 1 on three consecutive runs with six failures**, of which three are the benefit:

- *"the same-authority follow-up was admitted through the STEER path"* fails — `steerAnchor` is absent, i.e. production skipped `runActiveReplySteer` because `undefined !== <incoming fingerprint>` set `shouldQueueAuthorityMismatch`, and the user's clarification became a separate later turn instead of joining the turn about to run.
- *"execution's route bind resolves against the admitted snapshot"* fails with `Reply operation has no active tool authority snapshot` — the literal error from this PR's problem statement, still reachable on `main` for the duration of preflight.
- *"a later bind cannot replace the turn's admitted tool authority"* fails because on `main` nothing owns the window, so the first snapshot to arrive defines the turn's authority; on this branch admission has already claimed it and the later bind is refused.

**Proving the control is a control, not a red run.** Scenario 3 — a *different*-authority follow-up in the same window — passes on both trees, and the first three assertions of scenario 1 pass on `main` too. So the harness discriminates on real authority and fails for the specific reason claimed rather than failing uniformly.

**[Proof] "Real behavior: Needs proof."**
Covered by the above. The PR body now carries the mechanism, the exact terminal output of both the passing run and the failing control, and an explicit `### What remains different from current main` section.

### Gates on `41ca2c1ce10`

Proof 14/14 EXIT 0 (×4); negative control EXIT 1 (×3); `followup-turn-admission{,.authority}` + `followup-turn-execution{,.steering,.lifecycle}` **5 files requested / 5 executed, 92 passed EXIT 0** (53 literal cases; the two `.each` files expand to 37 and 45); `pnpm tsgo:scripts` EXIT 0; the scripts oxlint lane EXIT 0; `oxfmt --check` and `git diff --check` clean; `pnpm deadcode:unused-files` 0 entries on **both** the production and the full-tree scan.

### Not addressed, deliberately

`checks-node-compact-large-29` (`sessions.send-yield-resume.test.ts`) and `check-lint-core-4` (`update-managed-service-handoff-boundary.test-support.ts` `max-lines`) are red at this head. Neither file is in this branch's diff, so I have left them alone rather than widening this PR's scope.

The overlap question with #139929 and #143661 is unchanged and remains a maintainer call; nothing in this round argues for or against either.

*Written by Tank (Claude), an AI agent operated by Eddie Abrams (zeroaltitude).*

zeroaltitude:
## Round 2: Revision 7's [P2] was right — the harness claim is now bounded to what it observes (head `b31294a572`)

**The finding is correct and it was my defect, introduced by the previous commit in this same round.** Paraphrasing it: scenario 2 asserted `steerAnchor` and described that as the message being *steered*, but the scenario deliberately holds the admitted turn non-running with no backend attached, so `resolveCurrentMessageInjectionTarget` returns undefined and `runActiveReplySteer` falls back to a separate queued turn. `settleParkedSteerAcceptance` clears `steerPending` on that fallback but leaves `steerAnchor` in place — so the assertion passed while the completed steer had not happened. The comment also noted that the harness swallowed failures and timeouts, which it did.

**Taking the second option you offered — narrow the scenario and the claim** (commit `b31294a5727a56db05f55f494eab66fe3ea51814`, GPG-signed, harness only, no production file touched). Asserting a *completed* steer requires an executing turn with a real attached backend, which is a materially larger harness than this one and a separate piece of work; claiming it here would have been the same overreach in a new shape.

What changed:

- Scenario 2 is now **"ROUTED INTO the steer path"**, not "steered". That routing decision is exactly what this PR moves, and it is what the queue state can honestly witness.
- **The fallback is asserted explicitly** rather than left implicit: `steerPending` must be absent, with the reason stated inline. A completed steer can no longer be implied, and a future change that silently stopped falling back would now fail the run instead of passing quietly.
- **`runReplyAgent` outcomes are captured, not swallowed.** Both inbound turns now assert they settled with no error, so a rejection or a timeout fails the run. The previous `.catch(() => undefined)` plus a silent race is gone.
- The file header carries the scope limit in writing: proving a completed injection needs an executing turn with an attached backend and is out of this harness's scope.
- The PR body's `## User Impact` and `### What remains different from current main` were corrected to match — they said "is steered into that turn", and they now say the message is routed into the steering path, with an explicit scope note.

**The routing difference the control demonstrates is unchanged by the narrowing.** Against pristine `origin/main` (`38c3edcc4ca`), three consecutive runs, EXIT 1, six failures including:

```
  FAIL the same-authority follow-up was ROUTED INTO the steer path for the in-flight turn
       — steerAnchor is absent, so production skipped runActiveReplySteer and enqueued the
         message as its own turn
  FAIL execution's route bind resolves against the admitted snapshot
       — routeBindError=Reply operation has no active tool authority snapshot
```

On this head: **17 assertions, EXIT 0, three consecutive runs**. Eleven of the seventeen still pass on `main`, including the whole discrimination scenario, so the control remains discriminating rather than uniformly red.

### Gates on `b31294a572`

Proof 17/17 EXIT 0 (×3); negative control EXIT 1 (×3); `pnpm tsgo:scripts` EXIT 0; scripts oxlint lane EXIT 0; `oxfmt --check` and `git diff --check` clean. CI on the previous head `41ca2c1ce10` finished **226 checks: 177 success, 48 skipped, 1 neutral, 0 failure**; this commit changes only the harness.

### Still open, and honestly stated

The "successful execution through the revised lifecycle" request is **not** met by this harness and I am not claiming it is. What is proven is that admission owns the turn's authority before preflight, that the identical-authority comparison an inbound message performs therefore succeeds instead of failing against `undefined`, that this changes which path production takes, and that all of it fails on current `main`. Carrying a steer through to a completed injection needs an executing turn with an attached backend — I would rather flag that as the remaining gap than assert it from a harness that cannot reach it.

*Written by Tank (Claude), an AI agent operated by Eddie Abrams (zeroaltitude).*

zeroaltitude:
Round response to ClawSweeper **Revision 8** (reviewed head `b31294a5727a`, `updated_at` 2026-09-18T15:39:28Z). Findings: None; the two remaining items were both about proof. New head: **`04cb0574bdcbc9541d1f744d5dc416b012e15c19`**. No production file changed this round.

### [P2] "Add after-fix execution evidence showing the remaining benefit over current main"

**Problem as stated:** the previous terminal trace exercises real admission and routing and correctly records fallback, but stops before queued execution or successful injection.

**What I did.** Added `scripts/proof-142306-queued-execution-benefit.ts`, a companion harness that runs the whole lifecycle for real:

- It drives the real `admitFollowupTurn` + `executeFollowupTurn` pair — the exact two steps `createFollowupRunner` performs for every queued turn. They are called separately only because the window this PR changes lives *between* them; racing a millisecond-wide window from inside the runner would not be deterministic.
- The queued turn **really executes**: real model resolution, prompt build, attempt dispatch, `attachBackend`, streaming transport. It reaches `phase === "running"` **with an injectable backend** — asserted from production registry state while the run is in flight, which is the specific thing Revision 8 noted was missing — and settles `ok`.
- The **model endpoint is the only edge stub**: a `node:http` server on `127.0.0.1` speaking the OpenAI Responses streaming format, reached through an ordinary `models.providers` entry, the same shape `src/gateway/test-openai-responses-model.ts` uses for gateway HTTP tests. No credential and no external network are involved, and its request log is the record of which user prompts the agent runtime actually executed.
- After that turn completes, the **real follow-up drain** (`createFollowupRunner`, which production registers as the queue's executor) runs the surviving queued messages.

**The benefit, measured rather than argued.** One admitted turn plus a four-message burst arriving during its admission window, under real `queue` settings `cap: 2`, `drop: old`:

| | this branch | pristine `origin/main` `38c3edcc4ca` |
|---|---|---|
| window follow-up routed to the steer path | yes (`steerAnchor`) | no |
| window follow-up survives overflow | yes | no — evicted `queue-cap-old` |
| window follow-up **executed** (provider received its prompt) | **yes** | **no** |
| admitted turn executes / later message executes | yes / yes | yes / yes |
| result | 16/16, **EXIT 0 ×3** | **EXIT 1 ×3**, same 4 failures, 12/16 passing |

`steerAnchor` is a protected identity in `applyQueueDropPolicy`'s `isProtected` predicate; on `main` the same message is an ordinary queued item and, being the oldest, is the first victim of `drop: old`. The small cap only makes overflow reachable in four messages instead of twenty-one — the mechanism is unconditional production code. Full terminal output for both sides is in the PR body under **After-fix execution evidence**.

### Scope, stated plainly — what this does NOT show

It does not demonstrate a successful steer **injection** that `main` lacks, and it does not claim one. `resolveReplyMessageInjectionRejection` rejects with `not_running` before it ever compares authority, so injection requires `phase === "running"` with an attached backend — and by the time a queued turn is running, both trees own a snapshot. The difference this PR creates is confined to the admission→running window, which is what the harness measures. The parked steer falling back is asserted explicitly on both trees rather than glossed over.

I also corrected a related claim in the PR body while I was there. The earlier harness's route-bind failure is induced — it calls `bindToolAuthorityRoute` directly, and every production caller of that method runs after `setPhase("running")`. The body now says so, and reads that assertion as evidence of the unowned window rather than a crash a user hits on current `main`.

### Gates on `04cb0574bdc`

- Execution harness 16/16 **EXIT 0 ×3**; its control **EXIT 1 ×3** (12/16 passing). Admission harness still 17/17 **EXIT 0**; its control **EXIT 1** (11/17 passing). Both controls re-derived from the final, post-lint file.
- `followup-turn-admission{,.authority}` + `followup-turn-execution{,.steering,.lifecycle}`: **5 files requested / 5 executed, 92 passed, EXIT 0**.
- `pnpm tsgo:scripts` EXIT 0 · scripts oxlint lane EXIT 0 (probe-verified live against the new file) · `pnpm format:check` EXIT 0 over 38,614 files · `pnpm check:no-conflict-markers` EXIT 0 · `git diff --check` clean.
- `pnpm deadcode:unused-files`: **0 entries on both the production and full-tree scans**, with the new harness registered in `config/knip.config.ts`.

CI at this head is still running; I will edit this comment in place with the settled tally rather than posting a second one.

*Written by Tank (Claude), an AI agent operated by Eddie Abrams (zeroaltitude).*

zeroaltitude:
## Round 10 — ClawSweeper Revision 9 `[P2]`: production-consistent queue admission inputs

Head `04cb0574bdc` → **`3b79869af31c4cadc10055d1e92f2254405133c7`** (GPG verified). One commit, `proof(auto-reply): derive queue admission inputs from production (review feedback)`. **No production file changed and no file added** — the diff is the admission inputs of the two existing proof harnesses.

### `[P2]` Use production-consistent queue admission inputs in the proof — `scripts/proof-142306-queued-execution-benefit.ts:295-298`

**The finding, paraphrased.** Both harnesses set `resolvedQueue.mode` to `followup` while forcing `shouldSteer: true` and `queueAdmissionState: "steering"` for every inbound message. Production computes `shouldSteer` only when `resolvedQueue.mode === "steer"` and the queue admission state is not `ready` (`get-reply-run-admission.ts:562-570`), so that input pair is one the advertised mode cannot produce, and the overflow benefit was measured under it. The companion admission harness had the same defect.

**This is correct, and it was a real defect in the harness inputs.** Nothing was relabelled; the inputs are now produced by production code.

- **`resolvedQueue` is no longer written by the harness.** Both harnesses call production's own `resolveQueueSettingsCore` over their config, the way `get-reply-run-admission.ts:358` calls it. Neither config sets a queue mode, so the resolver supplies OpenClaw's documented default — **`steer`** (`docs/concepts/queue.md`, "Defaults"). A precondition asserts the resolved mode *is* `steer` and says why, so a later config change cannot silently move the measurement into a mode where the mechanism does not exist. `cap` / `drop` / debounce now come from `messages.queue` config through that same resolver; the admission harness sets none of them and resolves the real defaults `cap: 20`, `drop: "summarize"`.
- **`queueAdmissionState` is derived from the real queue before every inbound message**, by production's own `resolveReplyQueueAdmissionState` over the live follow-up queue and the live active reply operation — the same call `get-reply-run-admission.ts:557` makes.
- **`shouldSteer` / `shouldFollowup` are production's conjunctions over that state.** The conjuncts the fixture holds constant are named at the call site and are properties of the fixture, not overrides: a CLI user message is not a room event and not a heartbeat, nothing triggers a reset or preempts a heartbeat, no recovery owner is registered, and `resolveActiveRunAcceptsCurrentThread` returns `true` for any route that is not a Slack direct-routed thread turn (`:474-480`).

**The derived inputs are printed per message, and they show FIFO eligibility honoured rather than bypassed.** On this branch the window follow-up is admitted at `state=empty` (steering computed, and taken), the next burst message at `state=steering`, and the last two at `state=ready shouldSteer=false` — production stops computing steering once other waiting messages hold FIFO priority. On `main` the same derivation reaches `ready` one message earlier, because the unsteered window follow-up is sitting in the queue as ordinary backlog.

**The comparison still discriminates, on identical inputs.** The window follow-up is admitted on *both* trees with `state=empty shouldSteer=true`; what differs is the authority comparison, exactly as claimed.

| | this head | pristine `origin/main` `38c3edcc4ca` |
|---|---|---|
| `proof-142306-queued-execution-benefit` | **17/17, EXIT 0 ×3** | **EXIT 1 ×3**, 4 failures, 13/17 still passing |
| `proof-142306-queued-admission-authority` | **18/18, EXIT 0 ×3** | **EXIT 1 ×3**, 6 failures, 12/18 still passing |

Both control failure lists are byte-identical across all three runs. The four execution-side failures are unchanged in substance: no admission-time fingerprint, never routed into the steer path, evicted by overflow as the oldest unprotected item, never executed — the provider never receives that user's prompt.

**Complete next step — "publish a fresh discriminating run with the PR body's claims aligned to the tested mode".** Done, and the alignment is explicit rather than implied: the body now states that both observable consequences are scoped to `steer` mode by construction, because `shouldSteer` exists only there; under `followup` or `collect` no message is routed into the steer path on either tree, so no `steerAnchor` exists and this overflow difference does not arise. The admission-time ownership the patch introduces is unconditional; the *measured* benefit is claimed only as a `steer`-mode benefit — which is the default mode.

### Gates at `3b79869af31`

- Execution harness 17/17 EXIT 0 ×3; control EXIT 1 ×3. Admission harness 18/18 EXIT 0 ×3; control EXIT 1 ×3. All six runs per side taken from the final post-lint, post-format file; the harness files were `md5sum`-compared across the two worktrees before the runs.
- `followup-turn-admission{,.authority}` + `followup-turn-execution{,.steering,.lifecycle}`: 5 files requested / 5 executed, **92 passed, EXIT 0** — unchanged, since no test file changed.
- `pnpm tsgo:scripts` EXIT 0 · scripts oxlint lane EXIT 0, probe-verified by appending a deliberate violation and confirming the lane named this exact file before restoring · `pnpm format:check` EXIT 0 over 38,614 files · `pnpm check:no-conflict-markers` EXIT 0 · `git diff --check` clean · `pnpm deadcode:unused-files` 0 entries on both the production and full-tree scans.
- PR body updated; it was pre-validated against `real-behavior-proof-policy.mjs` (`status: passed`) before the edit, the edit left `head.sha` unchanged, and the body was re-read afterwards.

CI is running on this head; I will edit this comment in place with the settled tally. At the previous head the matrix settled with 0 failures and the PR reported `mergeable_state: clean`.

*Written by Tank (Claude), an AI agent operated by Eddie Abrams (zeroaltitude).*

PR #142318, opened 2026-09-08
perf(plugin-state): cover live-row quota scans

### LINKED ISSUE #142244: Plugin-state live-row counts: extend the existing listing index
## Accepted scope

The maintainer approved extending the existing `idx_plugin_state_listing` index with one trailing column:

```sql
CREATE INDEX IF NOT EXISTS idx_plugin_state_listing
  ON plugin_state_entries(plugin_id, namespace, created_at, entry_key, expires_at);
```

The current definition ends at `entry_key`. Keep the index name and ordered prefix, primary key, partial expiry index, all SQL predicates and call order, scalar APIs, TTL cutoffs, quotas, errors, write authority, transactions, workspace locks, and scheduling yields unchanged. No schema-version or package-version bump is included.

## Problem and owning boundary

Ordinary plugin-state registration performs both namespace and plugin live-row counts. Their expiry predicate currently requires value-bearing table access because `expires_at` is absent from the listing index. Adding it lets SQLite answer those same queries from the index. This remains a scan proportional to the matching index rows; it does not cache counts or make writes constant-time.

The shared-state database owns the canonical definition and repair. Memory workspace writers continue using the scalar store API. The scheduling fairness fix in #141613 and overflow-only recount removal in #142148 remain separate and unchanged.

## Evidence and tradeoffs

A source-bound synthetic experiment on macOS arm64, Node 26.8.1 / SQLite 3.53.4, changed both COUNT plans to use the covering listing index in six dense/sparse/TTL shapes. In the 50,000-row sparse fixture with 1 KiB values, median plugin COUNT time changed from 15.067 to 1.663 ms. These were warm-cache native SQL component measurements on a non-isolated host, not a complete store/Gateway benchmark or production latency claim.

That fixture's seeded database grew by 48 KiB. Its retained post-batch WAL file was 889,920 bytes larger; this is a point-in-time file-size difference, not cumulative write amplification. Mixed TTL payloads grew more. The index therefore trades extra persisted bytes and maintenance work for avoiding table access during count-heavy writes.

A separate count cache, cross-write retention ledger, batched memory transaction, altered count SQL, and skip-on-overwrite shortcut are outside the accepted scope because they would change freshness, visibility, failure, or ownership assumptions.

## Upgrade, failure, and reverse repair

Use the existing canonical index repair and `openclaw doctor --fix` paths. `CREATE INDEX IF NOT EXISTS` alone cannot replace the old same-name definition. The owner builds a temporary candidate before dropping the old index, builds the canonical index, removes the probe, and retains table/full-file integrity plus post-repair validation inside the existing savepoint. Preserve that entire sequence and its failure rollback.

A full-schema read-only open rejects the opposite named-index shape until a writable canonical owner or Doctor repairs it. Lightweight readers that check only the numeric schema version may continue reading either shape. An older compatible build can repair the same name back to the four-column definition, with the same data semantics but another one-time index rebuild. Plan for old/probe/final index overlap and existing full-file validation cost; do not bypass preflight or perform live SQL overlays.

## Required validation

Before landing, establish actual pre-fix and candidate query plans without forced index hints; preserve ordered listing. Prove old-definition upgrade through the real shared-state runtime and Doctor owners, same-version/idempotent reopen, failed-repair rollback, and reverse repair with unchanged logical row bytes and metadata contracts. Reuse existing whole-file integrity and write-authority refusal tests.

Run a representative ordinary store/memory path with source-bound persisted-state checks and bounded timing observations, followed by Linux Crabbox/Testbox proof. Complete focused tests, the changed gate, independent review, current review findings, exact-head CI, and the native PR landing workflow. Report upgrade and steady-state costs separately and retain unsuccessful attempts. Performance acceptance of a deployment requires its own representative runtime evidence.

steipete:
Closes #142244

## What Problem This Solves

Memory workspace updates and other plugin-state writes do avoidable synchronous database work when a plugin has many stored entries. The namespace and plugin quota checks need expiry data that is absent from the existing listing index, forcing SQLite to look up the corresponding table rows.

## Why This Change Was Made

Append `expires_at` to the existing nonunique listing index so the unchanged live-row counts and namespace expiry selection can use covering access. The index name and ordering prefix remain intact. Existing writable startup and Doctor owners repair the old named definition inside their current fences, transactions, and integrity checks; older writable owners can reverse the definition. The accepted storage and upgrade decision is recorded in #142244.

## User Impact

Plugin-state writes avoid table lookups for these checks while keeping the same scalar calls, expiry cutoffs, quotas, eviction/error ordering, transactions, and scheduling yields. The schema version is unchanged. First repair builds indexes and performs the existing integrity checks, requiring temporary disk space and startup work. Full-schema read-only consumers still require the matching index definition before use.

## Evidence

- Three regression cases failed on the original definition and passed after the change: covering COUNT access, runtime upgrade, and Doctor repair. All 334 focused tests passed, with one existing skip, including rollback, corruption, write authority, expiry/caps, import and workspace fairness. The full changed gate passed.
- Eleven separate-process real-owner stages passed locally and in Linux Testbox: canonical predecessor creation, full-schema read-only refusals, writable upgrade, idempotent reopen, old-owner reverse repair, re-upgrade, Doctor repair, and failed runtime/Doctor validation retaining the old index and data. Every child exited naturally and every owned process group closed. The failure proof combines inspected source ordering with retained final state; it does not trace transient DDL.
- The actual workspace helper and store API passed four fixed shapes on both platforms. Each 64-row replacement still issued 64 plugin COUNTs and 64 namespace COUNTs; counts and namespace expiry selection became covering, peer rows stayed unchanged, and the existing yields exposed committed progress.

Linux Node 24.19.0 results, medians of three warm-cache observations per variant:

| Plugin rows / namespace shape | Original | Changed |
| --- | ---: | ---: |
| 512 dense | 29.083 ms | 18.215 ms |
| 8,192 sparse | 134.351 ms | 32.546 ms |
| 50,000 sparse | 801.268 ms | 149.182 ms |
| 50,000 dense | 2,135.172 ms | 535.457 ms |

Dense fixtures place all plugin rows in the updated namespace; sparse fixtures place only the 64 overwritten rows there. Dense final listing still materializes all namespace entries, and changed dense heartbeat gaps remained 179–192 ms. These are synthetic helper/store measurements under observed, uncontrolled host load; they exclude enclosing workspace-lock admission, plugin trust dispatch, and Gateway/agent handling. No production-latency or CPU claim follows.

A separate canonical-copy supplement measured full owner open at 226.774 ms for the already-canonical predecessor and 561.368 ms for four-to-five-column repair on Linux. The baseline schema cookie stayed unchanged; the candidate advanced it while preserving rows and other schema definitions. Repair grew the 83,124,224-byte database to 89,174,016 bytes, with 1,661 free pages and a 12,281,752-byte WAL observed after open. These are one warm-cache pair and point-in-time footprints, not isolated DDL cost or cumulative write amplification.

The original workload’s separate owner-open fixture also triggered SQL-text normalization on its baseline, so that observation is not used as a no-repair comparison. A later local cost-harness preflight stopped before owner execution on a JSON object-prototype comparison; it remains retained, and the corrected supplement used fresh copies. The 64-write measurements were not repeated or relabeled.

The source-bound Linux proof used [Blacksmith Testbox run 34246080546](https://github.com/openclaw/openclaw/actions/runs/34246080546). Independent source and artifact review passed; managed autoreview found no actionable P0–P2 defect. No query, schema-version, configuration, or production TypeScript changes were needed.

### DIFF
--- a/docs/reference/database-schemas.md
+++ b/docs/reference/database-schemas.md
@@ -20,6 +20,26 @@ OpenClaw stores control-plane state in a global SQLite database and agent data i
 
 The task registry uses the global control-plane database. Runtime trajectory events live with their sessions in the per-agent database or a configured shared session SQLite store.
 
+### Plugin state listing index
+
+Plugin keyed stores use the shared `plugin_state_entries` table. Its listing
+index includes `expires_at` after the existing plugin, namespace, creation-time,
+and entry-key columns, so live-row counts can read the index without fetching
+stored values. Quotas, TTL cutoffs, ordering, and row contents are unchanged.
+
+Writable startup and `openclaw doctor --fix` replace the older four-column
+definition through canonical index repair, without a schema-version bump. The
+repair builds temporary indexes and runs the existing table and full-file
+integrity checks; allow for extra disk space and work proportional to stored
+entries during the first repair.
+
+An older build can rebuild the same index back to its expected definition.
+Full-schema read-only validation rejects a mismatched definition until a
+writable owner repairs it; lightweight readers that validate only the numeric
+schema version may read either shape. See the
+[accepted index design](https://github.com/openclaw/openclaw/issues/142244) for
+upgrade, reverse-repair, and performance proof requirements.
+
 ### Mentions Inbox
 
 The [mentions Inbox](/concepts/multi-user#temporary-mentions-inbox) uses existing
--- a/src/state/openclaw-state-db.test.ts
+++ b/src/state/openclaw-state-db.test.ts
@@ -4806,6 +4806,69 @@ INSERT INTO macos_port_guardian_records VALUES (4242, 18789, '/usr/bin/ssh', 're
     );
   });
 
+  it.each(["runtime", "doctor"])(
+    "upgrades the plugin listing index through %s without rewriting entries",
+    (repairPath) => {
+      const stateDir = createTempStateDir();
+      const options = { env: { OPENCLAW_STATE_DIR: stateDir } };
+      const databasePath = materializeCurrentStateDatabase(stateDir);
+      const { DatabaseSync } = requireNodeSqlite();
+      const legacy = new DatabaseSync(databasePath);
+      const entriesSql =
+        "SELECT * FROM plugin_state_entries ORDER BY plugin_id, namespace, entry_key";
+      const metadataSql =
+        "SELECT role, agent_id, schema_version, app_version FROM schema_meta WHERE meta_key = 'primary'";
+      let entries: unknown;
+      let metadata: unknown;
+      try {
+        legacy.exec(`
+          DROP INDEX idx_plugin_state_listing;
+          CREATE INDEX idx_plugin_state_listing
+            ON plugin_state_entries(plugin_id, namespace, created_at, entry_key);
+          INSERT INTO plugin_state_entries VALUES
+            ('plugin', 'written', 'live', '{ "value": 1 }', 10, NULL),
+            ('plugin', 'written', 'at-cutoff', '{ "value": 2 }', 10, 1000),
+            ('plugin', 'sibling', 'expired', '{ "value": 3 }', 20, 999),
+            ('plugin', 'sibling', 'future', '{ "value": 4 }', 20, 1001),
+            ('peer', 'written', 'live', '{ "value": 5 }', 10, NULL);
+        `);
+        entries = legacy.prepare(entriesSql).all();
+        metadata = legacy.prepare(metadataSql).get();
+        expect(metadata).toMatchObject({
+          schema_version: OPENCLAW_STATE_SCHEMA_VERSION,
+          app_version: VERSION,
+        });
+      } finally {
+        legacy.close();
+      }
+
+      if (repairPath === "doctor") {
+        const repaired = repairOpenClawStateDatabaseSchema(options);
+        expect(repaired.warnings).toEqual([]);
+        expect(repaired.changes).toContain("Rebuilt canonical shared-state SQLite indexes (1)");
+      }
+      const upgraded = openOpenClawStateDatabase(options);
+      expect(
+        upgraded.db
+          .prepare("PRAGMA index_info(idx_plugin_state_listing)")
+          .all()
+          .map((row) => row.name),
+      ).toEqual(["plugin_id", "namespace", "created_at", "entry_key", "expires_at"]);
+      expect(upgraded.db.prepare(entriesSql).all()).toEqual(entries);
+      expect(upgraded.db.prepare(metadataSql).get()).toEqual(metadata);
+      expect(readSqliteNumberPragma(upgraded.db, "user_version")).toBe(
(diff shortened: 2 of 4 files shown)


steipete:
Maintainer disposition for the three merge-risk items at `fc32f1a03ab11db5d040fcecd433cfaf18c2cf9d`: retain the accepted repair contract in #142244, as the review recommends. No new correctness or security finding requires a source change.

1. The synchronous repair and temporary disk costs are accepted and documented. A separate canonical-input comparison used byte-identical copies of an actual old-owner database with 50,002 rows. Full owner open was 161.292 ms versus 694.197 ms on macOS, and 226.774 ms versus 561.368 ms on Linux. The baseline retained its exact SQL, schema cookie and file hash; the candidate completed the existing repair. Its main file grew from 83,124,224 to 89,174,016 bytes, retaining 1,661 free pages, with 12,281,752 WAL bytes observed while open. These are single warm full-owner observations and retained sizes, not isolated DDL cost, peak disk demand, latency distributions or a production startup budget. The earlier fixture that caused repair on both sides remains separately qualified.

2. Full-schema read-only refusal and older-owner reverse repair are accepted compatibility behavior. Eleven fresh-process stages passed on macOS and Linux through the real runtime and Doctor owners, including upgrade, unchanged reopen, opposite-definition read-only refusal, actual older-owner reverse repair, re-upgrade and failed-repair preservation. Logical rows, stable metadata and unrelated schema remained intact; all children exited naturally. The existing write authority, savepoint and full integrity checks remain unchanged. Numeric-version-only lightweight readers remain a separate unchanged path.

3. The bot's external-host access limitation is recorded. Maintainer acceptance independently checked the downloaded Linux artifacts from [the hosted proof run](https://github.com/openclaw/openclaw/actions/runs/34246080546): all 74 main-packet files and all 20 supplemental files, source/tree bindings, manifests and natural-exit receipts. The exported small owner databases were also independently read using immutable SQLite. Large Linux workspace/cost databases were not exported, so their preservation evidence remains the reviewed source-bound assertions and receipts. Both runs match published tree `b8bcd80f1d98f6a92c4353b73423dcda8503e0a7`. The lease was closed and its completion verified.

These dispositions close the three maintainer decisions. The improvement remains bounded: the 50,000-row dense helper case still showed 179–192 ms Linux heartbeat gaps, and the proof does not establish production latency or complete Gateway/agent behavior. Exact-head required CI and native landing remain required.

steipete:
Merged via squash.

- Prepared head SHA: [fc32f1a03ab11db5d040fcecd433cfaf18c2cf9d](https://github.com/openclaw/openclaw/pull/142318/commits/fc32f1a03ab11db5d040fcecd433cfaf18c2cf9d)
- Landed commit: [64f7a06ab3992c839355a9d14c40467c5b929b87](https://github.com/openclaw/openclaw/commit/64f7a06ab3992c839355a9d14c40467c5b929b87)

PR #142338, opened 2026-09-08
perf(dev): defer compiler loading on clean launcher paths

### LINKED ISSUE #142337: Development runner loads the checker compiler before AST work is needed
The development runner imports the built-plugin control-plane checker through the postbuild module, which eagerly loads TypeScript even when no AST validation is requested. This retains a compiler module on an ordinary import path.

This is a maintainer-requested improvement: reuse the existing synchronous lazy compiler loader and acquire the same compiler at the checker's AST use sites. Preserve validation results, artifact freshness decisions, process and shutdown ownership, and the dependency contract.

A fresh-process regression fails on the original eager import and passes after the change, with an explicit compiler-load positive control. The 166 checker/postbuild/runner/lifecycle tests, selected gate, independent review, committed build and clean-artifact checks pass.

A separately audited 65-second keyless idle pair observed lower structural-parent RSS and summed launcher-tree RSS, while serving-Gateway RSS increased slightly. That single sequential observation cannot attribute the whole difference to TypeScript or establish physical-memory, model-workload or general performance gains.

steipete:
Closes #142337

<details>

</details>

## What Problem This Solves

Starting an already-built OpenClaw Gateway loads the full TypeScript compiler into the long-lived launcher through tooling imports, even when its clean artifact checks only need file and inventory facts. Both the bundled-plugin checker and the newer updater-compatibility module create this eager import path.

## Why This Change Was Made

Use type-only imports and the existing cached synchronous compiler loader at the actual AST operations. The compatibility recorder acquires the compiler inside its qualifying-file loop, so empty inventory work stays lazy. Compiler arguments and parent pointers, export mapping, generated bridges, inventory bytes, freshness checks, subprocess protocols and shutdown remain unchanged.

This maintainer-requested change touches four tooling/test files, with ten net production lines and 38 net test lines. It adds no dependency, configuration, cache owner, runtime policy or release behavior.

## User Impact

Clean artifact checks and launcher imports can finish without retaining a compiler module they do not need. AST-backed validation and generation still load the same compiler when required. Serving-Gateway memory is measured separately from launcher memory.

## Evidence

- The extended fresh-process regression failed on the original patch with the newer compatibility import path, then passed with this extension. All 198 focused cases and the full selected changed gate passed; independent review found no remaining issue in the final staged and working trees.
- Both normal full builds and real warm checks passed. The required-output inventory and clean build/postbuild decisions matched across 37 paths. The baseline loaded TypeScript; the candidate kept it unloaded until the explicit positive control. Existing bridge/AST behavior remains covered.
- The new comparison uses baseline `2a88b1e0f49f854e9bc0080fa4393ef7b9be8ebf` and candidate `da77ec6cf16cfdcb0dbb3e4199e3a87a2a9490ce` in the same checkout. Both normal keyless macOS Gateway runs passed readiness, process-roster/source binding and graceful zero-exit cleanup, released their ports and removed their private state. Each idle observation lasted 65.173 seconds.

| Observed endpoint RSS | Baseline MiB | Candidate MiB |
| --- | ---: | ---: |
| Structural launcher parent | 238.59 | 143.66 |
| Serving Gateway | 450.47 | 444.92 |
| Sum across the launcher tree | 801.17 | 700.39 |

This is one baseline-first observational pair with a **29m58.731s** interval for Git/source work, checks, review and the candidate build. Host paging/load, cache warmth, disk pressure and build/order differences remain confounders. The entire observed difference is not attributed to TypeScript. Summed RSS is not physical memory or PSS, and this keyless idle proof does not claim a real-provider workload or general latency improvement. Earlier measurements against the prior import graph remain historical and are not used for this current-main result.


steipete:
Merged via squash.

- Prepared head SHA: [da77ec6cf16cfdcb0dbb3e4199e3a87a2a9490ce](https://github.com/openclaw/openclaw/pull/142338/commits/da77ec6cf16cfdcb0dbb3e4199e3a87a2a9490ce)
- Landed commit: [251634acaccd9d0576a915461f7a154cc51b50f4](https://github.com/openclaw/openclaw/commit/251634acaccd9d0576a915461f7a154cc51b50f4)

PR #142165, opened 2026-09-08
feat(models): read session catalogs directly in the Control UI

obviyus:
Related: #136257, #142063.

## What Problem This Solves

Chat and New Session used competing catalog state instead of directly reading choices for the saved session or selected draft account. Refresh failures and account changes could leave misleading choices.

## Why This Change Was Made

Both surfaces use the published inventory owner. Gateway requests preserve saved account pins, validate draft-account ownership around asynchronous work, and return account display facts with the catalog. Shared direct readers replace the UI cache and implicit picker discovery. Explicit refresh waits for authentication publication.

## User Impact

Failed refreshes retain compatible rows with a warning; successful empty results clear them. Same-account preview retry can recover. Request and response additions preserve the protocol version and stored account defaults; the legacy metadata endpoint remains available.

Consumers: Chat, New Session, Agents, slash commands, and diagnostics share direct catalog reads.

## Evidence

The original models.list rejected session selectors. Gateway checks passed direct reads, ownership refusal, failed refresh, and empty results. A real built UI/Gateway preserved saved pins, showed refresh warnings, and made usable personal-account selections without inference. Its final check failed on an existing first-visit preference marker; subsequent readback confirmed no account-default change. Browser regressions and continuous integration passed. Native client adoption and live credentials were not tested.

PR #142105, opened 2026-09-08
fix(ui): discover models when default-model pickers open

xiaoguomeiyitian:
Closes #141813.

## What Problem This Solves

Fixes an issue where users opening a default-model picker on the Models page saw only the prepared choices, with no indication that more models could be discovered.

## Why This Change Was Made

All three pickers use the existing shared catalog discovery owner and show accessible progress. Discovery now retires its page subscription when core data, route, agent, or connection ownership changes. Success clears the catalog error while preserving unrelated errors and saved choices; a rejected request offers Retry.

The discovery and page-shell extraction keeps the page controller within its existing size limit. The contributor's three commits are preserved, including the latest test corrections; current rendered-control tests cover the same fixes against the current picker implementation.

## User Impact

Owners can discover and select additional models without blocking initial navigation or losing their primary, utility, and fallback selections. Older picker replies cannot overwrite newer page data. Provider-level unavailable results keep their existing behavior.

## Evidence

- Real Gateway, Control UI, browser, and deterministic local model provider: pinned main `451a568c2e75` showed three prepared choices and no picker discovery request. The corrected source showed expanded choices with the saved selection intact.
- Original candidate: a delayed old response replaced six newer choices with five; successful discovery also left an earlier catalog error visible. Both corrected paths passed ordered browser checks.
- Independent browser acceptance covered all three picker entrypoints, shared requests, accessible progress, enabled choices, actual Retry, error isolation, saved IDs, and core-refresh/agent/reconnect races.
- 146 focused tests passed across eight page, view, profile, load, catalog-store, model-picker, and select-picker files. Targeted formatting, lint, size/assertion checks, translation verification, and UI build passed. Full typechecks and broad checks run in hosted CI.
- The signed build at `657c878ade54` passed a final real-browser expansion and saved-selection check. The contributor reconciliation preserved that tested source tree. A subsequent type-only ownership correction removes a static import cycle without changing runtime behavior.

| Before: prepared choices only | After: discovered choices, saved selection retained |
| --- | --- |
| Models picker before discovery repair | Models picker after discovery repair |

Image links are verified and expire on September 15, 2026.

Thanks @xiaoguomeiyitian for the original implementation. AI-assisted repair and verification.

PR #142346, opened 2026-09-08
fix(ui): restore default-model picker discovery after the catalog store rewrite

steipete:
Related: #142105, #142165

## What Problem This Solves

Resolves a problem where Control UI typechecks on main fail after the merge race between #142105 and #142165. The Models settings default-model pickers also stopped issuing discovery requests.

#142105 merged at 16:11 UTC on September 8; #142165 replaced the catalog store at 16:24 UTC without reconciling the new picker caller and tests. The resulting errors are:

```text
catalog-discovery.ts(94,9): TS2353: 'refreshIfDue' does not exist in the loadModelCatalog options type.
model-providers-page.catalog.test.ts(7,10): TS2305: model-catalog-store.ts has no exported member 'invalidateModelCatalogCache'.
```

## Why This Change Was Made

Use the supported `refresh: true` request to preserve demand-driven discovery for primary, utility, and fallback pickers. Ordinary catalog reads are passive; the Gateway owns acquisition and publication, while the page retains its cancellation and stale-result guards.

Remove the tests' obsolete cache invalidation setup. Core refresh and route replacement directly retire the page request. Update request counts for direct reads: overlapping opens on one page still share its pending request, reopening after completion requests again, and separate pages own independent requests. No removed cache, cooldown, or invalidation API is restored. Clarify this Models settings behavior in the protocol documentation.

## User Impact

Opening a default-model picker discovers additional choices while preserving saved selections, usable controls, search, progress, retry, and protection against stale replies. Initial page loading remains passive. There are no appearance changes.

## Evidence

Before the repair, `pnpm tsgo:ui` reproduced TS2353 and all 11 picker discovery tests failed because no refresh request was issued.

- `pnpm tsgo:ui` — passed.
- `pnpm tsgo:core:test` — passed.
- `node scripts/run-vitest.mjs ui/src/pages/model-providers/model-providers-page.catalog.test.ts ui/src/pages/model-providers/model-providers-page.test.ts ui/src/lib/model-catalog-store.test.ts ui/src/pages/model-providers/load.test.ts` — 73 tests passed, including all 11 picker discovery cases.
- `node scripts/run-vitest.mjs ui/src/lib/chat/chat-metadata-store.test.ts` — 16 tests passed.
- `node scripts/check-changed.mjs` — passed; complete log inspected, including typechecks, formatting, i18n, ratchets, dead exports, and lint.
- `node scripts/check-docs-mdx.mjs docs/gateway/protocol/operator-methods.md` and `git diff --check` — passed.
[...]

PR #142347, opened 2026-09-08
fix(ui): model settings pickers send a retired catalog option

steipete:
## What Problem This Solves

Resolves a problem where opening a default-model picker in Models settings sends the retired `refreshIfDue` option to the Gateway. The direct catalog reader no longer accepts that UI-cache option, so the request is invalid and the call also fails TypeScript checks.

## Why This Change Was Made

Read the published catalog on picker open using the current typed request contract. The Gateway owns freshness and explicit Refresh owns provider discovery. Update the existing page tests to reflect independent reads while preserving pending-request deduplication, cancellation, retry behavior, and saved selections.

## User Impact

Primary, utility, and fallback model pickers can read published choices without requesting implicit provider discovery. Retiring one page's request does not cancel another page's read.

## Evidence

- The exact request regression fails before the fix because the retired field reaches `models.list`.
- All 33 focused page, catalog-reader, and page-loader tests pass, including the eleven catalog-page cases previously failing in CI.
- Independent P2 review passed. The full local changed gate passed, including its type graphs. Exact-head hosted CI remains required before landing.

Related: #142105, #142165

### DIFF
--- a/ui/src/pages/model-providers/catalog-discovery.ts
+++ b/ui/src/pages/model-providers/catalog-discovery.ts
@@ -1,10 +1,6 @@
-// Demand-driven catalog discovery for the Models settings page.
-//
-// The initial page load uses the fast prepared catalog (configured models only)
-// so full discovery stays out of first navigation. Opening a default-model picker
-// signals interest; this controller fetches the full catalog through the shared
-// model-catalog store (cooldown + concurrency dedupe) and merges it in without
-// disturbing the saved selection.
+// Picker-triggered reads of the Gateway's published model catalog.
+// The Gateway owns freshness and provider discovery; this controller only owns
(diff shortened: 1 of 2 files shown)


steipete:
Superseded by #142346, merged as f74837eec91453eb26ec18de94f907a0698721e7. That repair removes the retired store API usage and documents the intended distinction: conversation pickers read passively, while Models settings default pickers request discovery.

Verified the landed files against its reviewed head, its [successful required CI](https://github.com/openclaw/openclaw/actions/runs/34253166559), and current main with 33 focused catalog/page/loader tests plus `pnpm tsgo:ui`. Closing this PR to preserve that newer contract rather than land a conflicting behavior change.

PR #112811, opened 2026-07-23
feat(msteams): support multiple bot accounts

jimmypuckett:
## Current Exact-Head Status — 2026-09-20

- Current head: `93de7442e55c057caeede580984d127aeb402971`. The canonical worktree is clean, and local, fork, and GitHub PR heads match after normal forward merges and non-force pushes.
- GitHub's mergeability cache is currently `UNKNOWN`. An independent merge-tree check against current upstream `e6b42643e7a` is clean: the PR is 41 commits behind and 161 branch-only commits ahead, with no merge conflict.
- The prior exact-head CI found one branch-owned baseline-ratchet failure: `extensions/msteams/src/channel.test.ts` had grown from 1,000 to 1,004 counted lines. Head `93de7442e55` moves the existing directory-contract coverage into a coherent sibling test file without changing behavior. The affected 46/46 tests, exact line-cap ratchet, formatting, and `git diff --check` pass; fresh Codex Sol/high autoreview found no actionable P0-P2 findings and judged the repair correct at 0.97 confidence.
- The preceding repair cycle reproduced the sole Node failure locally, integrated its upstream-owned correction, and fixed three account-boundary defects found by fresh review: configured named-default outbound sends now carry the selected account through injected senders; omitted `enabled` no longer starts a previously inactive Teams listener; and named-account `appPassword` SecretRefs carry isolated owner attribution so one unavailable sibling does not become a global failure.
- Focused proof on the repaired lineage is green: 135 account/outbound/Graph/proactive/SecretRef tests, 20 listener lifecycle tests, and 4/4 formerly failing outbound-worker tests. Generated config-doc metadata, formatting, diff checks, and the current-base line-cap ratchet are also clean.
- Exact-head CI is complete: 179 checks succeeded, 41 skipped, one CodeQL result was neutral, zero failed, and `openclaw/ci-gate` passed.
- ClawSweeper revision 180 reviewed exact head `93de7442e55`, found no new code defect, and retained one medium authority-proof concern: live final effects still need to demonstrate revocation after reload with sibling access preserved and issuing-listener acceptance versus sibling-listener rejection through actual Gateway settlement.
- Last confirmed operational evidence remains OpenClaw 2026.9.4 with 15 healthy Teams listeners (14 isolated personal agents plus the shared agent) from the reviewed private package lineage. This remains supporting continuity evidence, not a claim that exact public head `93de7442e55` is deployed.
[...]

PR #142349, opened 2026-09-08
fix(diagnostics): stuck-session recovery never runs on hosts with late heartbeat ticks

### LINKED ISSUE #142200: Heartbeat lateness is used as an event-loop-health proxy, silently disabling stuck-session recovery on VMs
**Version:** openclaw 2026.9.2 (logic verified identical in 2026.9.3)
**Environment:** WSL2 (Linux 6.18.33.2-microsoft-standard-WSL2), Node 26.7.0, 16 vCPUs, Windows host on "Balanced" power scheme

## Summary

`startDiagnosticHeartbeat` infers event-loop congestion from how late its own 30s `setInterval` fired. On a VM, timer wakeup latency routinely exceeds the 1000ms tolerance for reasons unrelated to event-loop health. The result is that `shouldDeferRecovery` is true on **100% of ticks**, which permanently disables automatic stuck-session recovery while emitting a WARN every 30 seconds.

## The code

`dist/diagnostic-ryAZMr5N.js` (2026.9.2), lines 1224-1231:

```js
const heartbeatElapsedMs = lastDiagnosticHeartbeatTickAt === void 0 ? 0 : now - lastDiagnosticHeartbeatTickAt;
lastDiagnosticHeartbeatTickAt = now;
const heartbeatOverdueMs = Math.max(0, heartbeatElapsedMs - DIAGNOSTIC_HEARTBEAT_INTERVAL_MS);
const shouldDeferRecovery = heartbeatOverdueMs >= DEFAULT_LIVENESS_EVENT_LOOP_DELAY_WARN_MS;
if (shouldDeferRecovery && !inStartupGrace) diagnosticLogger.warn(`liveness heartbeat delayed: ...`);
```

and the consequence, line ~1288:

```js
if (!recovery || shouldDeferRecovery) continue;   // requestStuckSessionRecovery() never reached
requestStuckSessionRecovery({ ... });
```

With `DIAGNOSTIC_HEARTBEAT_INTERVAL_MS = 3e4` and `DEFAULT_LIVENESS_EVENT_LOOP_DELAY_WARN_MS = 1e3`, the check tolerates only 3.3% timer slop on a 30-second timer.

## Observed behavior

Every heartbeat tick, without exception, over the full log:

```
13:16:26.368Z WARN liveness heartbeat delayed: overdue=1483ms elapsed=31483ms; deferring recovery decisions
13:16:57.914Z WARN liveness heartbeat delayed: overdue=1546ms elapsed=31546ms; deferring recovery decisions
13:17:29.472Z WARN liveness heartbeat delayed: overdue=1558ms elapsed=31558ms; deferring recovery decisions
13:18:00.953Z WARN liveness heartbeat delayed: overdue=1481ms elapsed=31481ms; deferring recovery decisions
```

`overdue` ranged 1120-1681ms across 32 consecutive occurrences. Never once below the 1000ms threshold. The gateway was idle (2.2% CPU) throughout.

## The delay is not event-loop congestion

Reproduced with vanilla Node, no openclaw code involved:

```js
import { monitorEventLoopDelay } from 'node:perf_hooks';
const h = monitorEventLoopDelay({ resolution: 10 }); h.enable();
let last = Date.now();
setInterval(() => {
  const now = Date.now();
  console.log(`elapsed=${now-last}ms overdue=${now-last-30000}ms loopMax=${Math.round(h.max/1e6)}ms`);
  last = now; h.reset();
}, 30000);
```

Output — the same magnitude the gateway reports, while true event-loop delay is ~1% of it:

```
tick=1 elapsed=31483ms overdue=1483ms loopMax=13ms loopP99=10ms
tick=2 elapsed=31483ms overdue=1483ms loopMax=13ms loopP99=11ms
```

Other causes ruled out by measurement on the affected host:

- **Clock skew:** wall-clock vs `hrtime.bigint()` over 20s gave ratio 0.99997 (-1ms per 30s tick).
- **Swap / page faults:** a 30s sleep showed `majflt_delta=0`.
- **Heartbeat callback duration:** the bare reproduction above contains no callback work and shows the same lateness.
- **Event loop going idle:** adding a 1s keepalive timer did not help (still `overdue=1494ms`).

The delay is host timer/vCPU wakeup latency: it affects any process sleeping ≥~1s and is per-process. A 20ms poller running concurrently observed zero stalls >150ms during a window in which the gateway logged `overdue=1565ms`.

The lateness is also quantised in ~1.5s units, suggesting a periodic host-side stall of sleeping processes rather than a scaling error: one-shot `setTimeout` lateness measured 20s → 4ms, 25s → 1510ms, 30s → 1523ms, 35s → 3059ms.

Note this is not something the user can readily configure away: switching the Windows host from the "Balanced" to the "High performance" power scheme (restoring the hidden scheme via `powercfg /duplicatescheme 8c5e7fda-...`) changed the measured lateness only from ~1483ms to ~1456ms, i.e. not at all. Affected hosts stay above the 1000ms threshold on every tick.

## Impact

Automatic stuck-session recovery never runs on affected hosts. `logSessionAttention` still fires, so attention warnings appear, but `requestStuckSessionRecovery` is unreachable — the failure is silent in the sense that the warning text says "deferring recovery decisions" without conveying that deferral is permanent, not transient.

There is also no way to tune this: `DEFAULT_LIVENESS_EVENT_LOOP_DELAY_WARN_MS` is a module constant with no config override, unlike `stuckSessionWarnMs` / `stuckSessionAbortMs`, which are resolved through `resolveStuckSessionWarnMs()`.

## Suggested fixes

1. **Use the real signal.** The module already runs a liveness sampler producing `eventLoopDelayP99Ms` / `eventLoopDelayMaxMs` (via `monitorEventLoopDelay`). Gate `shouldDeferRecovery` on that rather than on timer lateness. In the reproduction above the true metric read 10-13ms, correctly indicating a healthy loop.
2. **Don't let deferral be unbounded.** If N consecutive ticks defer, either proceed anyway or escalate — a permanently-deferred recovery subsystem should not be a steady state.
3. **Make the threshold configurable**, consistent with the neighbouring stuck-session thresholds.
4. **Scale tolerance to the interval.** 1000ms on a 30s timer is 3.3%; typical VM timer slop exceeds that.

ayaangazali:
Fixes #142200.

## What Problem This Solves

Resolves a problem where automatic stuck-session recovery stops running for the lifetime of a Gateway process on virtualized hosts, so a session that hangs is warned about repeatedly but never recovered.

The diagnostic heartbeat treats its own lateness as a sign the event loop is congested. When a tick fires more than one second past its 30 second deadline, recovery decisions are deferred to the next healthy tick:

`src/logging/diagnostic.ts:1165`

```ts
const shouldDeferRecovery = heartbeatOverdueMs >= DEFAULT_LIVENESS_EVENT_LOOP_DELAY_WARN_MS;
```

```ts
if (!recovery || shouldDeferRecovery) {
  continue;
}
requestStuckSessionRecovery({ ... });
```

That tolerates 3.3 percent slop on a 30 second timer. The reporter measured a WSL2 host waking a sleeping timer 1120 to 1681 milliseconds late on 32 consecutive ticks while the true event-loop delay stayed at 10 to 13 milliseconds and the Gateway sat at 2.2 percent CPU. Every tick deferred, so the next healthy heartbeat the deferral waits for never arrived:

```
13:16:26.368Z WARN liveness heartbeat delayed: overdue=1483ms elapsed=31483ms; deferring recovery decisions
13:16:57.914Z WARN liveness heartbeat delayed: overdue=1546ms elapsed=31546ms; deferring recovery decisions
13:17:29.472Z WARN liveness heartbeat delayed: overdue=1558ms elapsed=31558ms; deferring recovery decisions
```

`logSessionAttention` still fires, so the operator sees stuck-session warnings, but `requestStuckSessionRecovery` is unreachable.

## Why This Change Was Made

The deferral protects a real case and is kept. What is wrong is the signal it reads. Wakeup latency is a property of the host scheduler; whether queued progress is stuck behind a blocked loop is a property of the event loop, and the Gateway already runs a sampler that measures the second one. On the reporter's ticks that sampler read 10 to 13 milliseconds while the heartbeat called the same ticks congested.

**The verdict belongs to the monitor, not the heartbeat.** Only the sampler knows whether its own next sample is overdue, and that turns out to be the whole answer:

`src/gateway/server/event-loop-health.ts`

```ts
eventLoopDelayed: () =>
  nowMs() - lastSampleAt >= EVENT_LOOP_DELAY_WARN_MS ||
  (lastSnapshot?.reasons.includes("event_loop_delay") ?? false),
```

`lastSampleAt` advances on every sampler callback, including the early-return path where a quiet window commits no snapshot, so it is a liveness marker for the sampler itself. A stall starves that 20 millisecond callback, which means the retained observation predates the stall and still reads healthy. The interval since the last callback is itself a delay measurement, so the monitor can answer correctly without waiting for a new observation to commit.

The Gateway passes that verdict for recovery and keeps the persistent filter for warnings, which are different questions:

```ts
readEventLoopDelayed: readinessEventLoopHealth.eventLoopDelayed,
```

`persistentDegradationSnapshot()` withholds anything under 60 seconds of continuous degradation (`event-loop-health.ts:14`, `:207-213`), which is right for an operator warning and wrong for a recovery decision, because a shorter genuine stall still holds queued progress.

Session ages continue to be measured against `recoveryObservationNow = now - heartbeatOverdueMs`, which is untouched, so timer jitter still cannot push a run past its threshold early.

Two earlier revisions of this PR got the mechanism wrong and are worth naming, since the current shape is a response to both. A counter capping consecutive deferrals gave chronic lateness a standing exemption. Comparing observation identity across ticks only proved that something had been published since the last heartbeat, which across a 30 second interval is about thirty commits, so a pre-stall observation still passed. Reading the pending sample interval has neither hole, adds no constant, and removes the third "no evidence" state entirely.

Because the callback is now a plain `() => boolean`, no freshness or identity semantics leak to SDK hosts implementing it. There is no longer a wrong way to return the value.

An earlier revision of this PR also carried a standalone fallback that read completed histogram samples when no verdict was supplied. That had the same ordering hole, with no way to detect its own pending sample, so it is removed rather than rebuilt. `startDiagnosticHeartbeat` is exported through `src/plugin-sdk/logging-core.ts`, and callers that omit the verdict now defer on a late tick, which is exactly main's behavior today. Only the Gateway changes.

No configuration option is added.

## User Impact

On hosts with chronic timer lateness, mainly WSL2 and other VMs, stuck sessions are recovered automatically again instead of only being warned about. A genuinely blocked event loop still defers recovery, including a stall shorter than the warning filter's minute and a stall that starved the sampler itself. The repeated warning now says which of the two it observed:

```
liveness heartbeat delayed: overdue=1483ms elapsed=31483ms; event loop stayed responsive, keeping recovery decisions
```

## Evidence

Production change is +40/-5 across three files, of which the Gateway wiring is one line.

The contract is proven at its owner, against the real monitor rather than a fake:

```ts
harness.samples(50);
expect(harness.monitor.snapshot()).toMatchObject({ degraded: false, reasons: [] });
expect(harness.monitor.eventLoopDelayed()).toBe(false);
harness.elapseWithoutSampling(1_200);
expect(harness.monitor.snapshot()).toMatchObject({ degraded: false, reasons: [] });
expect(harness.monitor.eventLoopDelayed()).toBe(true);
```

Healthy sampling, then the heartbeat-before-sampler window, asserting the snapshot still reads healthy while the verdict does not. Reducing the verdict to snapshot-only, the pre-fix semantics, fails it:

```
× reports a delayed loop once its own sample is overdue, before any new observation
  AssertionError: expected false to be true
  Tests  1 failed | 20 passed (21)
```

Reverting `src/logging/diagnostic.ts` to main, so the deferral reads lateness again:

```
× keeps recovery running when ticks are late but the event loop stayed responsive
  Error: missing recoverStuckSession call
```

A companion test asserts ten consecutive keeping-up samples stay `false`, so the guard fails in both directions rather than only one.

```
node scripts/run-vitest.mjs src/logging                              9 files, 110 passed
node scripts/run-vitest.mjs src/gateway/server/event-loop-health.test.ts,
  question.recovery.test.ts                                                    49 passed
node scripts/run-vitest.mjs src/gateway (recovery lane)             36 files, 697 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                                  clean
oxlint / oxfmt on all five touched files                                         clean
```

Proof gap worth stating plainly: this is fake-timer coverage of the heartbeat decision plus real-monitor coverage of the contract it depends on, not a runtime trace on an affected host. I have no WSL2 or VM host and cannot manufacture chronic host wakeup lateness here, so producing that trace would mean fabricating it. For anyone who does have such a host: run the Gateway with diagnostics enabled, leave a session `processing` past `stuckSessionWarnMs`, and grep for `liveness heartbeat delayed`. Before this change every line ends `deferring recovery decisions` and recovery is never requested; after it, lines end `event loop stayed responsive, keeping recovery decisions` and recovery is requested.

### DIFF
--- a/src/gateway/server-lifecycle.ts
+++ b/src/gateway/server-lifecycle.ts
@@ -620,6 +620,10 @@ export async function prepareGatewayLifecycle(params: {
     startDiagnosticHeartbeat(undefined, {
       getConfig: getRuntimeConfig,
       startupGraceMs: 60_000,
+      // Recovery reads the current observation; the warning path below reads the
+      // persistent filter, which withholds under 60s of continuous degradation and so
+      // would report a shorter genuine stall as a responsive loop.
+      readEventLoopHealth: readinessEventLoopHealth.snapshot,
       sampleLiveness: () => {
         const sample = readinessEventLoopHealth.persistentDegradationSnapshot();
         if (!sample || sample.degradedSinceMs == null) {
--- a/src/gateway/server-lifecycle.ts
+++ b/src/gateway/server-lifecycle.ts
@@ -620,10 +620,10 @@ export async function prepareGatewayLifecycle(params: {
     startDiagnosticHeartbeat(undefined, {
       getConfig: getRuntimeConfig,
       startupGraceMs: 60_000,
-      // Recovery reads the current observation; the warning path below reads the
-      // persistent filter, which withholds under 60s of continuous degradation and so
-      // would report a shorter genuine stall as a responsive loop.
-      readEventLoopHealth: readinessEventLoopHealth.snapshot,
+      // Recovery asks the monitor for a verdict that accounts for its own pending sample;
+      // the warning path below reads the persistent filter, which withholds under 60s of
+      // continuous degradation and so would report a shorter genuine stall as healthy.
+      readEventLoopDelayed: readinessEventLoopHealth.eventLoopDelayed,
       sampleLiveness: () => {
         const sample = readinessEventLoopHealth.persistentDegradationSnapshot();
         if (!sample || sample.degradedSinceMs == null) {
--- a/src/gateway/server/event-loop-health.test.ts
+++ b/src/gateway/server/event-loop-health.test.ts
@@ -148,6 +148,33 @@ describe("createGatewayEventLoopHealthMonitor", () => {
     },
   );
 
+  it("reports a delayed loop once its own sample is overdue, before any new observation", () => {
+    // Recovery asks this after a late heartbeat. A stall starves the 20ms sampler, so the
+    // retained observation still reads healthy; the pending interval is the only evidence
+    // that exists at that moment, and it is itself a delay measurement.
+    const harness = createMonitorHarness();
+    harness.samples(50);
+    expect(harness.monitor.snapshot()).toMatchObject({ degraded: false, reasons: [] });
+    expect(harness.monitor.eventLoopDelayed()).toBe(false);
+
+    harness.elapseWithoutSampling(1_200);
+
+    expect(harness.monitor.snapshot()).toMatchObject({ degraded: false, reasons: [] });
+    expect(harness.monitor.eventLoopDelayed()).toBe(true);
+
+    harness.sample();
+    expect(harness.monitor.eventLoopDelayed()).toBe(true);
+  });
+
+  it("reports a responsive loop while the sampler keeps up", () => {
+    const harness = createMonitorHarness();
+    harness.samples(50);
+    for (let index = 0; index < 10; index++) {
+      harness.sample();
+      expect(harness.monitor.eventLoopDelayed()).toBe(false);
+    }
+  });
+
   it("retains completed observations until the next sampling window", () => {
     const harness = createMonitorHarness();
     harness.samples(50);
--- a/src/gateway/server/event-loop-health.ts
+++ b/src/gateway/server/event-loop-health.ts
@@ -34,6 +34,7 @@ export type GatewayEventLoopHealth = {
 type GatewayEventLoopHealthMonitor = {
   snapshot: () => GatewayEventLoopHealth | undefined;
   persistentDegradationSnapshot: () => GatewayEventLoopHealth | undefined;
+  eventLoopDelayed: () => boolean;
   reset: () => void;
   stop: () => void;
 };
@@ -203,6 +204,13 @@ export function createGatewayEventLoopHealthMonitor(
 
   return {
     snapshot: () => lastSnapshot,
+    // A stall starves this sampler, so the retained observation predates it and reads
+    // healthy. The interval since the last callback is itself a delay measurement, so the
+    // verdict comes from here rather than leaving callers to infer freshness from the
+    // observation object.
+    eventLoopDelayed: () =>
+      nowMs() - lastSampleAt >= EVENT_LOOP_DELAY_WARN_MS ||
+      (lastSnapshot?.reasons.includes("event_loop_delay") ?? false),
     // The heartbeat consumes the sampler's snapshot without advancing its window.
     persistentDegradationSnapshot: () => {
       const current = lastSnapshot;
--- a/src/logging/diagnostic.test.ts
+++ b/src/logging/diagnostic.test.ts
@@ -180,6 +180,16 @@ function expectNoLoggerMessageContaining(spy: unknown, text: string): void {
   expect(loggerMessages(spy).join("\n")).not.toContain(text);
 }
 
+/** Sampler result for a genuinely blocked event loop, the only state that defers recovery. */
+function stalledEventLoopSample() {
+  return {
+    reasons: ["event_loop_delay" as const],
+    intervalMs: 30_000,
+    eventLoopDelayP99Ms: 1_200,
+    eventLoopDelayMaxMs: 1_500,
+  };
+}
+
 function expectRecoveryCall(
   recoverStuckSession: unknown,
   fields: Record<string, unknown>,
@@ -506,7 +516,11 @@ describe("stuck session diagnostics threshold", () => {
     const warnSpy = vi.spyOn(diagnosticLogger, "warn").mockImplementation(() => undefined);
 
     vi.setSystemTime(0);
-    startEnabledDiagnosticHeartbeat({ recoverStuckSession });
+    startEnabledDiagnosticHeartbeat({
+      recoverStuckSession,
+      readEventLoopHealth: () => ({ reasons: ["event_loop_delay"] }),
+      sampleLiveness: () => stalledEventLoopSample(),
+    });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
 
     vi.setSystemTime(120_001);
@@ -530,12 +544,101 @@ describe("stuck session diagnostics threshold", () => {
     );
   });
 
+  it("keeps recovery running when ticks are late but the event loop stayed responsive", () => {
+    // Reported on WSL2: host timer wakeup latency put every tick ~1.5s past its 30s
+    // deadline while measured event-loop delay stayed at 10-13ms, so recovery deferred
+    // for the lifetime of the process.
+    const recoverStuckSession = vi.fn();
+    const warnSpy = vi.spyOn(diagnosticLogger, "warn").mockImplementation(() => undefined);
+
+    vi.setSystemTime(0);
+    startEnabledDiagnosticHeartbeat({
+      recoverStuckSession,
+      readEventLoopHealth: () => ({ reasons: [] }),
+      sampleLiveness: () => null,
+    });
+    logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
+
+    const advanceLateTick = () => {
+      vi.setSystemTime(Date.now() + 1_500);
+      vi.advanceTimersByTime(30_000);
+    };
+
+    advanceLateTick();
+    advanceLateTick();
+    advanceLateTick();
+
+    expectRecoveryCall(
+      recoverStuckSession,
+      { sessionId: "s1", sessionKey: "main", queueDepth: 0 },
+      ["ageMs", "stateGeneration"],
+    );
+    expectLoggerMessageContaining(warnSpy, "event loop stayed responsive");
+  });
+
+  it("defers a late tick when no event-loop health evidence is available", () => {
+    // Without a working delay monitor nothing rules out a stall holding queued progress,
+    // so the tick keeps the original conservative behavior rather than authorizing abort.
+    const recoverStuckSession = vi.fn();
+
+    vi.setSystemTime(0);
+    startEnabledDiagnosticHeartbeat({
+      recoverStuckSession,
+      readEventLoopHealth: () => undefined,
+      sampleLiveness: () => null,
+    });
+    logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
+
+    vi.setSystemTime(120_001);
+    vi.advanceTimersByTime(30_000);
+
+    expect(recoverStuckSession).not.toHaveBeenCalled();
+  });
+
+  it("still defers recovery when a late tick follows a real event-loop stall", () => {
+    // Chronic lateness must not buy a standing exemption: a genuine stall can queue
+    // progress that has not run yet, and aborting on that snapshot cancels healthy work.
+    const recoverStuckSession = vi.fn();
+    const warnSpy = vi.spyOn(diagnosticLogger, "warn").mockImplementation(() => undefined);
+    let eventLoopStalled = false;
+
+    vi.setSystemTime(0);
+    startEnabledDiagnosticHeartbeat({
+      recoverStuckSession,
+      readEventLoopHealth: () => ({ reasons: eventLoopStalled ? ["event_loop_delay"] : [] }),
+      // Deliberately still null while stalled: this is the Gateway contract, where
+      // persistentDegradationSnapshot() withholds under 60s of continuous degradation.
+      // Recovery must not read that filter as a responsive loop.
+      sampleLiveness: () => null,
+    });
+    logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
+
+    const advanceLateTick = () => {
+      vi.setSystemTime(Date.now() + 1_500);
+      vi.advanceTimersByTime(30_000);
+    };
+
+    advanceLateTick();
+    advanceLateTick();
+    recoverStuckSession.mockClear();
+
+    eventLoopStalled = true;
+    advanceLateTick();
+
+    expect(recoverStuckSession).not.toHaveBeenCalled();
+    expectLoggerMessageContaining(warnSpy, "deferring recovery decisions");
+  });
+
   it("defers a material heartbeat stall even when elapsed time is below the abort threshold", () => {
     const recoverStuckSession = vi.fn();
     const warnSpy = vi.spyOn(diagnosticLogger, "warn").mockImplementation(() => undefined);
 
     vi.setSystemTime(0);
-    startEnabledDiagnosticHeartbeat({ recoverStuckSession });
+    startEnabledDiagnosticHeartbeat({
+      recoverStuckSession,
+      readEventLoopHealth: () => ({ reasons: ["event_loop_delay"] }),
+      sampleLiveness: () => stalledEventLoopSample(),
+    });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
     markDiagnosticEmbeddedRunStarted({ sessionId: "s1", sessionKey: "main" });
 
--- a/src/logging/diagnostic.test.ts
+++ b/src/logging/diagnostic.test.ts
@@ -181,16 +181,15 @@ function expectNoLoggerMessageContaining(spy: unknown, text: string): void {
 }
 
 /**
- * Mirrors the Gateway monitor's `snapshot()` contract: the same object is returned until a
- * new sampling window commits. Returning a fresh object per read would hide the stale-read
- * ordering that a starved sampler produces.
+ * Stands in for the Gateway monitor's `eventLoopDelayed()` verdict. The monitor owns the
+ * freshness rule, including its own pending sample, and is covered by its own suite.
  */
 function createFakeEventLoopMonitor() {
-  let observation: { reasons: string[] } | undefined;
+  let delayed = false;
   return {
-    read: () => observation,
-    commit: (...reasons: string[]) => {
-      observation = { reasons };
+    delayed: () => delayed,
+    setDelayed: (next: boolean) => {
+      delayed = next;
     },
   };
 }
@@ -533,7 +532,7 @@ describe("stuck session diagnostics threshold", () => {
     vi.setSystemTime(0);
     startEnabledDiagnosticHeartbeat({
       recoverStuckSession,
-      readEventLoopHealth: () => ({ reasons: ["event_loop_delay"] }),
+      readEventLoopDelayed: () => true,
       sampleLiveness: () => stalledEventLoopSample(),
     });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
@@ -568,17 +567,15 @@ describe("stuck session diagnostics threshold", () => {
 
     vi.setSystemTime(0);
     const monitor = createFakeEventLoopMonitor();
-    monitor.commit();
     startEnabledDiagnosticHeartbeat({
       recoverStuckSession,
-      readEventLoopHealth: monitor.read,
+      readEventLoopDelayed: monitor.delayed,
       sampleLiveness: () => null,
     });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
 
     // The sampler keeps up, so every tick sees an observation covering its own interval.
     const advanceLateTick = () => {
-      monitor.commit();
       vi.setSystemTime(Date.now() + 1_500);
       vi.advanceTimersByTime(30_000);
     };
@@ -603,7 +600,7 @@ describe("stuck session diagnostics threshold", () => {
     vi.setSystemTime(0);
     startEnabledDiagnosticHeartbeat({
       recoverStuckSession,
-      readEventLoopHealth: () => undefined,
+      readEventLoopDelayed: () => true,
       sampleLiveness: () => null,
     });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
@@ -620,12 +617,11 @@ describe("stuck session diagnostics threshold", () => {
     const recoverStuckSession = vi.fn();
     const warnSpy = vi.spyOn(diagnosticLogger, "warn").mockImplementation(() => undefined);
     const monitor = createFakeEventLoopMonitor();
-    monitor.commit();
 
     vi.setSystemTime(0);
     startEnabledDiagnosticHeartbeat({
       recoverStuckSession,
-      readEventLoopHealth: monitor.read,
+      readEventLoopDelayed: monitor.delayed,
       // Deliberately still null while stalled: this is the Gateway contract, where
       // persistentDegradationSnapshot() withholds under 60s of continuous degradation.
       // Recovery must not read that filter as a responsive loop.
@@ -638,46 +634,15 @@ describe("stuck session diagnostics threshold", () => {
       vi.advanceTimersByTime(30_000);
     };
 
-    monitor.commit();
     advanceLateTick();
-    monitor.commit();
     advanceLateTick();
     recoverStuckSession.mockClear();
 
-    monitor.commit("event_loop_delay");
+    monitor.setDelayed(true);
     advanceLateTick();
 
     expect(recoverStuckSession).not.toHaveBeenCalled();
     expectLoggerMessageContaining(warnSpy, "deferring recovery decisions");
-  });
-
-  it("refuses recovery when a stall starved the sampler and the observation is stale", () => {
-    // The monitor commits about once a second and returns the same object until it does.
-    // A stall delays both timers; if the heartbeat resumes first, the retained observation
-    // still reads healthy and must not be mistaken for a responsive loop.
-    const recoverStuckSession = vi.fn();
-    const monitor = createFakeEventLoopMonitor();
-    monitor.commit();
-
-    vi.setSystemTime(0);
-    startEnabledDiagnosticHeartbeat({
-      recoverStuckSession,
-      readEventLoopHealth: monitor.read,
-      sampleLiveness: () => null,
-    });
-    logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
-
-    // First late tick sees a fresh healthy window, so recovery is allowed to run.
-    monitor.commit();
-    vi.setSystemTime(Date.now() + 1_500);
-    vi.advanceTimersByTime(30_000);
-    recoverStuckSession.mockClear();
-
-    // Now the sampler is starved: the same object is returned, still reading healthy.
-    vi.setSystemTime(Date.now() + 1_500);
-    vi.advanceTimersByTime(30_000);
-
-    expect(recoverStuckSession).not.toHaveBeenCalled();
   });
 
   it("defers a material heartbeat stall even when elapsed time is below the abort threshold", () => {
@@ -687,7 +652,7 @@ describe("stuck session diagnostics threshold", () => {
     vi.setSystemTime(0);
     startEnabledDiagnosticHeartbeat({
       recoverStuckSession,
-      readEventLoopHealth: () => ({ reasons: ["event_loop_delay"] }),
+      readEventLoopDelayed: () => true,
       sampleLiveness: () => stalledEventLoopSample(),
     });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
--- a/src/logging/diagnostic.test.ts
+++ b/src/logging/diagnostic.test.ts
@@ -592,15 +592,15 @@ describe("stuck session diagnostics threshold", () => {
     expectLoggerMessageContaining(warnSpy, "event loop stayed responsive");
   });
 
-  it("defers a late tick when no event-loop health evidence is available", () => {
-    // Without a working delay monitor nothing rules out a stall holding queued progress,
-    // so the tick keeps the original conservative behavior rather than authorizing abort.
+  it("defers a late tick when the caller supplies no loop-delay verdict", () => {
+    // Standalone callers omit readEventLoopDelayed. Nothing then rules out a stall holding
+    // queued progress, so the tick keeps the original conservative behavior rather than
+    // authorizing abort from a measurement it cannot check for freshness.
     const recoverStuckSession = vi.fn();
 
     vi.setSystemTime(0);
     startEnabledDiagnosticHeartbeat({
       recoverStuckSession,
-      readEventLoopDelayed: () => true,
       sampleLiveness: () => null,
     });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
--- a/src/logging/diagnostic.test.ts
+++ b/src/logging/diagnostic.test.ts
@@ -180,6 +180,21 @@ function expectNoLoggerMessageContaining(spy: unknown, text: string): void {
   expect(loggerMessages(spy).join("\n")).not.toContain(text);
 }
 
+/**
+ * Mirrors the Gateway monitor's `snapshot()` contract: the same object is returned until a
+ * new sampling window commits. Returning a fresh object per read would hide the stale-read
+ * ordering that a starved sampler produces.
+ */
+function createFakeEventLoopMonitor() {
+  let observation: { reasons: string[] } | undefined;
+  return {
+    read: () => observation,
+    commit: (...reasons: string[]) => {
+      observation = { reasons };
+    },
+  };
+}
+
 /** Sampler result for a genuinely blocked event loop, the only state that defers recovery. */
 function stalledEventLoopSample() {
   return {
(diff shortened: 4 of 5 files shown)


ayaangazali:
Two things: the red checks were not mine, and the P1 finding was. Both handled.

**CI.** The five failures on `4cbf27e` (`check-prod-types`, `check-test-types`, `check-test-types-core-1`, `check-test-types-core-2`, `checks-node-compact-large-17`) all came from main, not this branch. This PR touches only `src/logging/`, no `ui/` files at all. The single root cause:

```
ui/src/pages/model-providers/catalog-discovery.ts(94,9): error TS2353: Object literal may only
specify known properties, and 'refreshIfDue' does not exist in type '{ agentId?: ... }'
```

plus the 11 matching runtime failures in `ui/src/pages/model-providers/model-providers-page.catalog.test.ts`. I reproduced both on a clean detached checkout of `upstream/main` at `d9c6554e782` with an empty working tree. `refreshIfDue` was introduced by #142105 (`6e1aba41cc2`) and appeared nowhere else in the repository, so the option was never added to the `loadModelCatalog` request type. #142346 fixed it on main at 17:03 UTC. I rebased rather than pushing anything speculative, and the checks are green.

**The P1 finding was correct and I dropped the approach it was aimed at.** You wrote that once two ticks are late, every later late tick permits recovery even when it follows a genuine stall, and that subtracting the overdue interval only moves the observation timestamp without processing queued progress. That is right, and it is a hole I argued around in the first PR body rather than fixing. A counter turns chronic lateness into a standing exemption, which trades one wrong outcome for another.

The rewrite reads the signal the module already measures instead:

```ts
const shouldDeferRecovery =
  heartbeatDelayed && (rawLivenessSample?.reasons.includes("event_loop_delay") ?? false);
```

The tick is reordered so the sample is taken first. Nothing between the old and new positions depends on the deferral, and the read happens before the histogram resets, so the reported window is unchanged. Because the sampler reports the interval maximum rather than an instantaneous value, a stall that started and ended inside the interval still defers. The `MAX_CONSECUTIVE_DEFERRED_HEARTBEAT_TICKS` constant is gone.

I had listed two objections to this approach in the original body. Both were wrong on inspection. The ordering one is a three-line move. The "null means healthy or unavailable" one does not apply here: on the Gateway path the sampler is supplied by the caller and always present, and on the built-in path `monitorEventLoopDelay` has been core Node since v11, so I am not adding a branch for a state that does not occur.

Coverage now proves the guard in both directions, since neither case alone distinguishes the two failure modes. Reverting `diagnostic.ts` to main:

```
× keeps recovery running when ticks are late but the event loop stayed responsive
  Error: missing recoverStuckSession call
```

Forcing `shouldDeferRecovery = false`:

```
× still defers recovery when a late tick follows a real event-loop stall
× suppresses liveness warnings during startupGraceMs while still sampling
```

That second test is the case you asked to cover: two late ticks on a responsive loop, then a late tick on a stalled one. Recovery runs on the first two and is refused on the third. The two pre-existing deferral tests keep their assertions and now supply a stalled sampler result, which is the state that actually defers under this change.

On the real-runtime proof item, I want to be straight rather than quietly stall. I have no WSL2 or VM host, and I cannot manufacture chronic host wakeup lateness on this machine, so producing a "redacted real-runtime trace" would mean fabricating the one artifact you asked for. What I can offer: if a maintainer has a Gateway on an affected host, the check is `OPENCLAW_DIAGNOSTICS=1` with a session left in `processing` past `stuckSessionWarnMs`, then grep the log for `liveness heartbeat delayed`. Before this change every line ends `deferring recovery decisions` and no recovery is requested. After it, lines end `event loop stayed responsive, keeping recovery decisions` and recovery is requested. I am happy to close this instead if you would rather the fix come from someone who can produce that trace.

ayaangazali:
Checked the finding against the source before touching anything, and it is correct. Fixed.

You pointed at `src/gateway/server-lifecycle.ts:648-651`. `persistentDegradationSnapshot()` is

```ts
persistentDegradationSnapshot: () => {
  const current = lastSnapshot;
  return current?.degradedSinceMs != null &&
    current.degradedSinceMs >= PERSISTENT_DEGRADATION_WARN_AFTER_MS
    ? current
    : undefined;
},
```

with `PERSISTENT_DEGRADATION_WARN_AFTER_MS = 60_000` (`src/gateway/server/event-loop-health.ts:14`, `:207-213`). So on the production Gateway my predicate was reading warning persistence, not loop health: a stall measuring a multi-second delay reports `null` for its first minute, my predicate went false, and recovery could hand `allowActiveAbort` to a snapshot whose progress callback was still queued. My test bypassed exactly that by injecting `event_loop_delay` into the sample. Your read of it is right on every point.

**Recovery now has its own seam, separate from the warning sampler.** The Gateway passes the unfiltered observation for recovery and keeps the persistence filter for warnings:

```ts
readEventLoopHealth: readinessEventLoopHealth.snapshot,
```

and the heartbeat consumes it without touching `sampleLiveness`:

```ts
const eventLoopDelayed = opts?.readEventLoopHealth
  ? suppliedHealth?.reasons.includes("event_loop_delay")
  : readBuiltInEventLoopDelayed();
const shouldDeferRecovery = heartbeatDelayed && eventLoopDelayed !== false;
```

Standalone runtimes keep the built-in monitor, read without consuming it since `sampleDiagnosticLiveness` still owns the reset. An absent observation defers rather than proceeding: only a positive "responsive" reading unblocks recovery, so no health signal cannot authorize aborting active work.

The regression test is now built to the Gateway contract instead of around it. It supplies `sampleLiveness: () => null` while the health reader reports `event_loop_delay`, which is precisely what a sub-minute stall looks like in production. Replacing the seam with the previous revision's `sampleLiveness` read:

```
× still defers recovery when a late tick follows a real event-loop stall
× suppresses liveness warnings during startupGraceMs while still sampling
  AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
```

and reverting `diagnostic.ts` to main:

```
× keeps recovery running when ticks are late but the event loop stayed responsive
  Error: missing recoverStuckSession call
```

A third test covers an absent observation and confirms it defers.

One thing I did not add, and want to say rather than leave implied. You asked to "cover the actual Gateway callback contract". I considered a test in `event-loop-health.test.ts` asserting that `snapshot()` reports a short stall while `persistentDegradationSnapshot()` withholds it. I left it out because it would pass on unfixed code as well as fixed, so it protects nothing this change introduces, and the neighbouring "does not turn reads without samples into healthy observations" test already exercises both readers. If you would rather have that distinction pinned explicitly against future edits to the monitor, say so and I will add it.

On the real-runtime proof item my position is unchanged and I would rather repeat it than pad the PR. I have no WSL2 or VM host and cannot manufacture chronic host wakeup lateness on this machine, so a "redacted real-runtime trace" from me would be fabricated. Everything else in the checklist is addressed. If a maintainer has a Gateway on an affected host: diagnostics enabled, one session left `processing` past `stuckSessionWarnMs`, then grep `liveness heartbeat delayed`. Before, every line ends `deferring recovery decisions` and recovery is never requested. After, lines end `event loop stayed responsive, keeping recovery decisions` and recovery is requested.

Unrelated but worth flagging since it nearly cost a second red run: `server-lifecycle.ts` sits at 699 counted lines on main against the 700 `max-lines` cap, so the one line added here puts it at exactly 700. It lints clean, but the file has no headroom left.

ayaangazali:
`checks-node-compact-small-31` is not this diff. Evidence below, and I rebased rather than pushing a guess at it.

**What failed.** All of it is one file, `src/cli/cron-output.process.test.ts`, and every failure is the same thing:

```
AssertionError: expected Error: spawnSync /home/runner/_work/_temp… { …(5) } to be undefined
  "message": "spawnSync .../node/v24.19.0-linux-x64/bin/node ETIMEDOUT",
  "errno": -110,
  "code": "ETIMEDOUT",
```

Not an assertion about behavior. The test spawns a real Node process with `--import tsx` under `timeout: 30_000` (`src/cli/cron-output.process.test.ts:98-102`) and the spawn never returned. The shard reported `12 failed | 38 passed (50)`; the file itself defines exactly six cases, and the six named in the log are all six of them.

**Why it cannot be mine.** This branch touches `src/logging/diagnostic.ts`, `src/logging/diagnostic.test.ts` and `src/gateway/server-lifecycle.ts`. The failing subprocess loads `src/cli/cron-cli.ts`. I walked the relative-import graph from that entry point: 2203 modules, and neither changed file is in it. `server-lifecycle.ts` is a Gateway module the CLI cron path never starts, and `diagnostic.ts` is not reachable either. There is no path by which this diff runs in that process.

**It passes here.** Same file, same head, on this machine:

```
node scripts/run-vitest.mjs src/cli/cron-output.process.test.ts
 ✓ drains output and preserves the exit code for wait success 4708ms
 ✓ drains output and preserves the exit code for wait failure 4194ms
 ✓ drains output and preserves the exit code for queued run 3946ms
 ✓ drains output and preserves the exit code for not-due run 4381ms
 ✓ drains output and preserves the exit code for run history 4019ms
 ✓ drains output and preserves the exit code for human error 3894ms
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

Around four seconds per spawn locally against a 30 second cap in CI, so the hosted runner was roughly seven times slower than that budget on `agentic-cli-process-hosted-6`. I also checked the fourteen most recently updated open PRs in the repo and none currently show a `compact-small` failure, so this is not an active repo-wide break; it looks like that one runner.

**Staleness checked, and it was not the cause, but I rebased anyway.** The branch was 57 commits behind `main`. `src/cli/cron-output.process.test.ts` is byte-identical between my merge base and `main`, so the newer base would not have changed the outcome. Rebased onto `1d257a9f6d7` regardless, which re-runs CI on a current base and settles it either way. Substance is unchanged at +150/-7 across the same three files, and both key hunks are intact: `readEventLoopHealth: readinessEventLoopHealth.snapshot` at `server-lifecycle.ts:651` and the deferral predicate at `diagnostic.ts:1197`.

At the rebased head `a292ba3`:

```
node scripts/run-vitest.mjs src/logging/diagnostic.test.ts                  95 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                             clean
node scripts/run-oxlint.mjs <all three touched files>                       clean
./node_modules/.bin/oxfmt --check <all three touched files>                 clean
```

`server-lifecycle.ts` is still at exactly 700 counted lines against the 700 `max-lines` cap after the rebase, so that stays as flagged.

No unresolved inline review threads on this PR. I am waiting on a re-review of the Gateway-seam revision, which landed after the last one; the outstanding item there is still the real-runtime trace I cannot produce without an affected host.

ayaangazali:
@clawsweeper re-review

Asking because the last run did not finish. A lease was taken for `a292ba3` at 2026-09-09T04:38 and expired at 04:53 without posting a verdict, so the labels and the "Before merge" list on this PR are still Revision 2's, which reviewed `a8e1dbb`.

Two things changed after that revision:

- The Gateway sampler mismatch you filed as P1 is fixed. Recovery now reads its own seam (`readEventLoopHealth: readinessEventLoopHealth.snapshot` at `src/gateway/server-lifecycle.ts:651`) instead of `sampleLiveness`, so `persistentDegradationSnapshot()`'s 60 second filter no longer decides recovery. Details and the both-directions test evidence are in my 2026-09-08 comment.
- Rebased onto current main. `checks-node-compact-small-31` was `spawnSync ... ETIMEDOUT` in `src/cli/cron-output.process.test.ts`, a file whose entry point does not import either changed module; that is written up in the thread and the PR is green now.

The one item I expect to remain open is the real-runtime trace, which I still cannot produce without a WSL2 or VM host.

ayaangazali:
Checked this one against the monitor before answering, and you are right again. I am not pushing a fix tonight because this machine is out of memory headroom and I will not ship a predicate change I cannot test, but here is the confirmation and the shape I intend.

The test you named demonstrates it directly (`src/gateway/server/event-loop-health.test.ts:134-148`):

```ts
harness.samples(priorWindow ? 50 : 4);
const previous = harness.monitor.snapshot();
harness.elapseWithoutSampling(1_200);
for (let index = 0; index < 100; index++) {
  expect(harness.monitor.snapshot()).toBe(previous);   // stale healthy read
}
harness.sample();
expect(current).toMatchObject({ degraded: true, reasons: ["event_loop_delay"] });
```

So during the exact window where the 20 ms sampler has been starved, `snapshot()` keeps handing back the pre-stall object with `reasons: []`. My predicate reads that as a responsive loop and permits recovery while progress callbacks are still queued. That is the hole, and injecting reasons in my test is what hid it.

The adjacent test `retains completed observations until the next sampling window` makes it worse than a stall-only edge: the monitor only commits a new observation about once per second, so a stale read is ordinary steady state, not an exceptional condition.

What I plan to do about it, without touching the monitor. `GatewayEventLoopHealth` has no observation timestamp, but `snapshot()` returns the same object reference until a new window commits, which both of those tests assert with `toBe`. So the heartbeat can hold the previous tick's reference and require the observation to have actually advanced:

```ts
const health = opts?.readEventLoopHealth?.();
const advanced = health !== undefined && health !== lastObservedEventLoopHealth;
lastObservedEventLoopHealth = health;
// An observation that has not moved since the previous tick cannot describe the window
// this tick is deciding about, so it counts as no evidence rather than a healthy loop.
const eventLoopDelayed = advanced ? health.reasons.includes("event_loop_delay") : undefined;
```

with the existing `eventLoopDelayed !== false` rule unchanged, so no evidence still defers. In steady state the sampler commits every second against a 30 second heartbeat, so the observation has always advanced and the chronic-lateness fix keeps working. The only time it fails to advance is when the sampler itself was starved, which is the case you are protecting.

Coverage would drive the real monitor rather than an injected reason: sample, stall without sampling, fire a late heartbeat, and assert recovery is refused; then sample once and assert the next tick behaves. That also removes the "injecting immediately updated reasons misses it" objection.

Tell me if you would rather see an explicit `observedAtMs` on `GatewayEventLoopHealth` instead of reference identity. Identity is the smaller change and needs no monitor edit, but it is load bearing on `snapshot()` continuing to return a stable reference, and a field would make that contract explicit rather than implied by two tests.

On the runtime-proof item, nothing has changed on my side: no WSL2 or VM host here, so I cannot produce that trace and will not fake one.

ayaangazali:
Fixed, and the coverage now drives a monitor that behaves like the real one rather than injecting reasons.

The predicate requires the observation to have advanced since the previous tick:

```ts
function resolveSuppliedEventLoopDelayed(
  health: { reasons: readonly string[] } | undefined,
): boolean | undefined {
  const advanced = health !== undefined && health !== lastObservedEventLoopHealth;
  lastObservedEventLoopHealth = health;
  return advanced && health ? health.reasons.includes("event_loop_delay") : undefined;
}
```

`undefined` already means no evidence and no evidence already defers, so an unchanged observation now takes the conservative path instead of reading as a responsive loop. Steady state commits about once a second against a 30 second heartbeat, so the observation has always advanced on a healthy host and the chronic-lateness fix is unaffected.

You were right that injecting fresh reasons per read hid this, so the test fixture was the actual problem. Replaced it with a fake that mirrors `snapshot()`:

```ts
function createFakeEventLoopMonitor() {
  let observation: { reasons: string[] } | undefined;
  return {
    read: () => observation,
    commit: (...reasons: string[]) => {
      observation = { reasons };
    },
  };
}
```

Same object on every read until a window commits, which is the property `preserves overdue delay across reads with prior window=%s` and `retains completed observations until the next sampling window` pin on the real monitor. All three deferral tests now drive it, and the new one reproduces your ordering directly: commit a healthy window, let one late tick run and allow recovery, then fire a second late tick with no intervening commit.

On the previous head that new test fails for exactly the reported reason:

```
× refuses recovery when a stall starved the sampler and the observation is stale
  AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
  Tests  1 failed | 95 passed (96)
```

At head `2ce4103`:

```
node scripts/run-vitest.mjs src/logging                                     8 files, 103 passed
node scripts/run-vitest.mjs src/gateway/server/event-loop-health.test.ts,
  question.recovery.test.ts                                                          47 passed
node scripts/run-vitest.mjs src/gateway (recovery lane)                    36 files, 685 passed
oxlint / oxfmt on both touched files                                                    clean
```

I went with reference identity rather than adding `observedAtMs` to `GatewayEventLoopHealth`, since it needs no monitor change and the stable-reference property is already asserted by those two tests. It is still an implied contract rather than a stated one, so if you would rather make it explicit I will add the field and key off it instead. Say the word.

Two gaps, both stated rather than papered over. I did not re-run `tsgo -p tsconfig.core.json` at this head: this machine was over its memory ceiling by the time the suites finished and I would rather leave it to CI than push the box into another out-of-memory event. And the real-runtime trace is unchanged, since I still have no WSL2 or VM host.

ayaangazali:
Right again, and this one tells me the mechanism is wrong rather than the threshold.

Your sequence is exactly reproducible from the code: heartbeat reads snapshot A, the sampler publishes healthy snapshot B somewhere in the next 30 seconds, then a stall delays both timers and the heartbeat resumes first. `snapshot()` returns B, `B !== A`, my check calls that fresh, and recovery proceeds on an observation that predates the stall. Identity only proves something was published since the last heartbeat, which over a 30 second interval is roughly thirty commits, so it says almost nothing about whether the last second is covered. And you are right that my test hid it by committing nothing in between; that is the same fixture mistake as last round, one level up.

The second risk you raised lands on the same root. `readEventLoopHealth` is exposed through logging-core, and identity-based freshness is an unstated requirement on anything that implements it. A standalone SDK host returning a fresh object per call, which is the obvious implementation, silently gets the pre-fix behavior back. That is a bad contract to ship even if I patched the timing.

So I want to stop patching the predicate and move the question to the owner, which is what "sampler-owned freshness" reads like to me. Rather than the heartbeat inferring health from an object it does not own, the monitor answers a question only it can answer, something like whether the loop was delayed over a caller-supplied recent window, with the pending-sample interval accounted for on its side. That gives the callback a plain verdict instead of an object with identity semantics, so the SDK contract becomes explicit, and it puts the freshness rule next to the sampler that knows its own cadence.

Before I build a fourth revision I would rather check the shape with you, since the last three each fixed a real hole and still landed in the wrong place:

1. Extend `GatewayEventLoopHealthMonitor` with a freshness-aware query the heartbeat calls, and reduce `readEventLoopHealth` to that verdict.
2. Keep returning the observation but add an explicit `observedAtMs` to `GatewayEventLoopHealth`, and have the heartbeat require it to cover its own overdue window.

Option 1 is what your wording points at and it fixes the SDK contract as a side effect. Option 2 is the smaller diff and keeps the monitor untouched, but it leaves the heartbeat deciding freshness for a sampler whose cadence it does not know. I lean 1. Tell me which you want and I will build it.

Proof gap unchanged and still honestly stated: no WSL2 or VM host here, so the redacted runtime trace showing recovery under harmless lateness and protection during a genuine stall is not something I can produce. Every revision so far has been fake-timer coverage of the decision, and I have labeled it that way each time.

ayaangazali:
Addendum to my question above, because I went looking for what "sampler-owned freshness" could read and the monitor already tracks it. This makes the choice concrete rather than a design debate.

`sample()` updates `lastSampleAt` on **every** callback (`src/gateway/server/event-loop-health.ts:129-130`), including the early-return path a few lines later where a quiet window commits no snapshot:

```ts
histogram.record(BigInt(Math.max(1, Math.round((now - lastSampleAt) * 1_000_000))));
lastSampleAt = now;
const intervalMs = Math.max(1, now - lastWallAt);
const delayMaxMs = nanosecondsToMilliseconds(histogram.max);
if (delayMaxMs < EVENT_LOOP_DELAY_WARN_MS && intervalMs < SUSTAINED_LOAD_SAMPLE_MIN_INTERVAL_MS) {
  return;
}
```

So `lastSampleAt` is a liveness marker for the sampler itself, independent of snapshot commits: it advances every 20 ms on a healthy loop while `lastSnapshot` only advances about once a second. That is exactly the pending-sample evidence my identity check was a poor substitute for.

Which means the monitor can answer the whole question, and the answer collapses the ambiguity I was carrying:

- If `now - lastSampleAt >= EVENT_LOOP_DELAY_WARN_MS`, the sampler's own 20 ms callback has been blocked for at least a second. That is not merely missing evidence, it is a delay measurement, so recovery defers.
- Otherwise the sampler is keeping up and `lastSnapshot` is at most a sampling window old, so `reasons.includes("event_loop_delay")` is current enough for a 30 second heartbeat.

That removes the third state entirely. There is no "no evidence" case left to reason about, `readEventLoopHealth` reduces to a plain boolean verdict, and the SDK contract stops depending on object identity, which was your second risk.

So I would like to revise my recommendation to option 1 with that specific mechanism: a monitor-owned query keyed off `lastSampleAt`, and `readEventLoopHealth` narrowed to the verdict. It reuses state the monitor already maintains, so the diff is a query on the monitor plus the narrowed callback, and the coverage can drive the real monitor with `elapseWithoutSampling` rather than a fake.

Still your call, and I have not pushed anything. Confirm the shape and I will build it, including the ordering test you asked for, healthy sampling between heartbeats followed by heartbeat-before-sampler.

ayaangazali:
Built option 1. You had not answered which shape you wanted, and after two days I judged this the reversible choice rather than leaving the P1 open: it is a PR revision you can reject, and your own wording pointed at it.

The verdict now belongs to the monitor, because it is the only place that knows whether its own next sample is overdue:

```ts
eventLoopDelayed: () =>
  nowMs() - lastSampleAt >= EVENT_LOOP_DELAY_WARN_MS ||
  (lastSnapshot?.reasons.includes("event_loop_delay") ?? false),
```

That is the sequence you described, answered at the owner. After a stall starves the 20 ms sampler the retained observation still reads healthy, but `now - lastSampleAt` has grown past a second, and that interval is itself a delay measurement. No snapshot needs to have advanced for the answer to be correct, so the B-differs-from-A hole is gone rather than narrowed.

It also closes the second risk you raised. `readEventLoopDelayed` takes a plain `() => boolean`, so nothing about object identity or freshness leaks into the callback. An SDK host implementing it cannot accidentally reproduce the old behavior, because there is no longer a wrong way to return the value. The Gateway side is one line:

```ts
readEventLoopDelayed: readinessEventLoopHealth.eventLoopDelayed,
```

The heartbeat lost the module-level `lastObservedEventLoopHealth`, the identity helper, and the third "no evidence" state. Consumer diff dropped from +58/-5 to +38/-5 and its tests from +156/-2 to +121/-2, against +8 on the monitor and +27 on the monitor's tests.

Coverage moved with the contract, which is the part I got wrong twice. The ordering case is now a monitor test driving the real monitor through its own harness rather than a fake in the heartbeat suite:

```ts
harness.samples(50);
expect(harness.monitor.snapshot()).toMatchObject({ degraded: false, reasons: [] });
expect(harness.monitor.eventLoopDelayed()).toBe(false);
harness.elapseWithoutSampling(1_200);
expect(harness.monitor.snapshot()).toMatchObject({ degraded: false, reasons: [] });  // stale read
expect(harness.monitor.eventLoopDelayed()).toBe(true);
```

Healthy sampling first, then the heartbeat-before-sampler window, asserting the snapshot still reads healthy while the verdict does not. Reducing the verdict to snapshot-only, which is the pre-fix semantics, fails it for the right reason:

```
× reports a delayed loop once its own sample is overdue, before any new observation
  AssertionError: expected false to be true
  Tests  1 failed | 20 passed (21)
```

A companion test asserts ten consecutive keeping-up samples stay `false`, so the guard fails in both directions.

At head `b589a49`, rebased onto current main:

```
node scripts/run-vitest.mjs src/logging                              9 files, 110 passed
node scripts/run-vitest.mjs src/gateway/server/event-loop-health.test.ts,
  question.recovery.test.ts                                                    49 passed
node scripts/run-vitest.mjs src/gateway (recovery lane)             36 files, 697 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                                  clean
oxlint / oxfmt on all five touched files                                         clean
```

Unchanged and still worth stating plainly: no WSL2 or VM host here, so the redacted runtime trace showing recovery under harmless lateness and protection during a genuine stall is not something I can produce. Everything above is fake-timer coverage of the decision plus real-monitor coverage of the contract it now depends on.

ayaangazali:
`control-ui-performance` went red on this head and it is not this diff.

The violation is 29 bytes over an absolute budget, measured on the base:

```
[control-ui-performance-base]
  startup CSS gzip vs base: 46046 B -> 46046 B (+0 B; growth below 1024 B allowed)
  violations:
    - startup JS gzip: 346.8 KiB exceeds 346.8 KiB (355145 B vs 355116 B)
```

The delta-vs-base lines are all `+0 B`, the failing step is the `-base` measurement, and this branch touches `src/logging/` and `src/gateway/` with no `ui/` files at all. The same check is currently FAILURE on #146403, #146534 and #146541, so it is main-side budget drift rather than anything a single PR introduced.

Not pushing a speculative change for it. Re-baselining that budget is a maintainer call and belongs in its own PR.

ayaangazali:
Both red checks are main's, with different fates. Rebased onto current main, which already clears one of them.

**`checks-node-compact-small-40`** is a repository capacity check, not a test of this diff:

```
test/scripts/tsgo-core-test-shards.test.ts > covers the repository test roots exactly once
  with headroom below the hard cap
+   "plugins-platform: 702 test roots exceeds the 700 limit"
```

It counts test roots across the repo. This branch adds no new test file, it only edits two existing ones, so it cannot move that count. I reproduced the identical failure on a clean detached `upstream/main` with an empty working tree:

```
× covers the repository test roots exactly once with headroom below the hard cap
+   "plugins-platform: 702 test roots exceeds the 700 limit"
  Tests  1 failed | 23 passed (24)
```

It is also currently failing on #146191 and #112811, and three PRs are already in flight against it: #146543, #146545, and #146553. Nothing for me to push here.

**`control-ui-performance`** was the 29-byte budget overage I described above, and main fixed it while this sat: #146526, "chore(ui): refresh the Control UI startup budget baseline". The rebase picks that up, so that one should go green on its own.

`openclaw/ci-gate` is the aggregate of those two.

Rebased to `e854b64` with no conflicts. Substance unchanged at +46/-5 production across three files. Revalidated at the new head:

```
node scripts/run-vitest.mjs src/logging/diagnostic.test.ts              95 passed
node scripts/run-vitest.mjs src/gateway/server/event-loop-health.test.ts  21 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                          clean
```

ayaangazali:
Fixed by deleting the fallback rather than teaching it to detect its own pending sample.

You are right about the hole, and it is worth saying that the helper you flagged was mine, added earlier in this same PR. It does not exist on main. So the question was never how to make it safe, it was whether it should exist at all.

It should not. `readBuiltInEventLoopDelayed` read completed histogram samples with no way to know its own next callback was overdue, which is the same ordering you had just made me fix on the Gateway side. Rebuilding pending-awareness for it would mean duplicating the monitor's job in a second place, for a path with no in-repo consumer.

```ts
// Only a positive "responsive" reading unblocks recovery. A caller without a monitor
// cannot rule out a stall holding queued progress, so it keeps deferring on lateness.
const eventLoopDelayed = opts?.readEventLoopDelayed?.();
const shouldDeferRecovery = heartbeatDelayed && eventLoopDelayed !== false;
```

The important consequence for the SDK callers you named: `startDiagnosticHeartbeat` is exported through `src/plugin-sdk/logging-core.ts`, so those callers are real. Omitting `readEventLoopDelayed` now means `undefined`, which fails `!== false` and defers. That is exactly main's behavior today:

```
main: const shouldDeferRecovery = heartbeatOverdueMs >= DEFAULT_LIVENESS_EVENT_LOOP_DELAY_WARN_MS;
```

So no SDK caller regresses, and none of them inherits a stale measurement I introduced. Only the Gateway, which supplies the pending-aware verdict, changes behavior. Source is down to +40/-5 across three files from +45/-5.

One thing I will not overclaim. The renamed test, "defers a late tick when the caller supplies no loop-delay verdict", pins the omitted-reader path, but it does not fail if I restore the fallback: the module monitor is not initialized under the test's fake timers, so the old helper returned `undefined` there and deferred anyway. I checked, and it passed 95/95 with the fallback restored. The justification for removing it is the ordering hole you identified plus the absence of any in-repo consumer, not a failing test, and I would rather say that than dress it up.

At head `8f74e80`:

```
node scripts/run-vitest.mjs src/logging                              9 files, 110 passed
node scripts/run-vitest.mjs src/gateway/server/event-loop-health.test.ts,
  question.recovery.test.ts                                                    49 passed
node scripts/run-vitest.mjs src/gateway (recovery lane)             36 files, 697 passed
node scripts/run-tsgo.mjs -p tsconfig.core.json                                  clean
oxlint / oxfmt on both touched files                                             clean
```

The runtime-proof item is unchanged: no WSL2 or VM host here, so I cannot produce the after-fix trace and will not fabricate one.

PR #142352, opened 2026-09-08
fix(diagnostics): defer recovery on observed loop stalls, not timer lateness

### LINKED ISSUE #142200: Heartbeat lateness is used as an event-loop-health proxy, silently disabling stuck-session recovery on VMs

ylcn91:
Closes #142200

WIP draft: stuck-session recovery deferral keys off measured event-loop delay instead of heartbeat timer lateness. Body will be completed before marking ready.

### DIFF
--- a/src/gateway/server-lifecycle.ts
+++ b/src/gateway/server-lifecycle.ts
@@ -646,14 +646,16 @@ export async function prepareGatewayLifecycle(params: {
       getConfig: getRuntimeConfig,
       startupGraceMs: 60_000,
       sampleLiveness: () => {
-        const sample = readinessEventLoopHealth.persistentDegradationSnapshot();
-        if (!sample || sample.degradedSinceMs == null) {
+        // Loop-delay evidence is fresh every tick; warning reasons need persistent degradation.
+        const sample = readinessEventLoopHealth.snapshot();
+        if (!sample) {
           return null;
         }
+        const persistent = readinessEventLoopHealth.persistentDegradationSnapshot();
         return {
-          reasons: sample.reasons,
+          reasons: persistent?.reasons ?? [],
           intervalMs: sample.intervalMs,
-          degradedSinceMs: sample.degradedSinceMs,
+          degradedSinceMs: persistent?.degradedSinceMs ?? undefined,
           eventLoopDelayP99Ms: sample.delayP99Ms,
           eventLoopDelayMaxMs: sample.delayMaxMs,
           eventLoopUtilization: sample.utilization,
--- a/src/logging/diagnostic.test.ts
+++ b/src/logging/diagnostic.test.ts
@@ -506,7 +506,8 @@ describe("stuck session diagnostics threshold", () => {
     const warnSpy = vi.spyOn(diagnosticLogger, "warn").mockImplementation(() => undefined);
 
     vi.setSystemTime(0);
-    startEnabledDiagnosticHeartbeat({ recoverStuckSession });
+    // No loop monitor reading: heartbeat lateness stays the only stall evidence.
+    startEnabledDiagnosticHeartbeat({ recoverStuckSession, sampleLiveness: () => null });
     logSessionStateChange({ sessionId: "s1", sessionKey: "main", state: "processing" });
 
     vi.setSystemTime(120_001);
@@ -516,7 +517,7 @@ describe("stuck session diagnostics threshold", () => {
     const delayedHeartbeat = loggerMessages(warnSpy).find((message) =>
       message.includes("liveness heartbeat delayed"),
     );
-    expect(delayedHeartbeat).toMatch(/overdue=\d+ms elapsed=\d+ms/u);
+    expect(delayedHeartbeat).toMatch(/overdue=\d+ms elapsed=\d+ms eventLoopDelayMaxMs=unknown/u);
     const timing = delayedHeartbeat?.match(/overdue=(\d+)ms elapsed=(\d+)ms/u);
     expect(Number(timing?.[2]) - Number(timing?.[1])).toBe(30_000);
     expect(recoverStuckSession).not.toHaveBeenCalled();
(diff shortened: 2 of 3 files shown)


ylcn91:
Closing in favor of #142349, which was opened first for #142200. This branch keeps the alternative that keys the deferral off the measured event-loop delay instead of a consecutive-tick cap (ylcn91:fix/heartbeat-loop-delay-deferral), in case that direction is wanted.

PR #142353, opened 2026-09-08
fix(e2e): mount trusted registry module closure

RomneyDa:
## Root cause

The upgrade-survivor container executes the trusted current `npm-registry-server.mjs`, but its relative `scripts/lib/bounded-response.mjs` import can resolve from the frozen target tree. The mixed module graph fails before the upgrade scenarios with a missing named export.

## Fix

Mount the trusted workflow `scripts/` tree as one read-only module root and point the existing registry helper override at that exact server entrypoint in both direct and published-baseline paths. Candidate application data and behavior remain unchanged.

## Proof

- Exact release validation red: `npm-registry-server.mjs` requested `createBoundedResponseTooLargeError` from the frozen helper.
- `test/scripts/docker-build-helper.test.ts`: 284 passed, 2 skipped.
- Bash syntax, Oxfmt, OXLint, diff check pass.
- AutoReview P1 scoped-clean (0.96).

## LOC

- Production/tooling: +4
- Tests: +10
- Application runtime: 0


RomneyDa:
@clawsweeper review

RomneyDa:
@clawsweeper re-review

PR #142366, opened 2026-09-08
refactor(tests): consolidate runtime group policy contracts

steipete:
## What Problem This Solves

Six channel-labelled fallback suites repeat the same core group-policy resolver. The WhatsApp and Zalo test helpers are direct aliases, so these cases do not exercise plugin wiring despite their names.

## Why This Change Was Made

Move the four missing unique cases to the canonical resolver suite, keep its existing configured-provider case, and delete the redundant contract suite and two sole-use helpers. Missing-provider behavior remains covered with absent, open, and disabled global defaults, plus an explicit disabled provider override. Remove the empty top-level hook while retaining the real nested module-reset hook.

## User Impact

No runtime behavior changes. The test suite removes 112 net lines while retaining every distinct argument/result contract represented by the old suites.

## Evidence

- Baseline: both complete suites passed all 24 cases.
- After: the complete canonical suite passed all nine cases, including all five original canonical cases and four migrated cases.
- AST expansion confirms the 19 removed executions match exactly the five retained unique argument/oracle pairs, including absent properties and both result fields.
- Mapped changed gate and fresh independent review cover the full four-file change. No runtime savings are claimed.


steipete:
Merged via squash.

- Prepared head SHA: [1cbbd65c5224325d87e8946a632bb871ce7bbb66](https://github.com/openclaw/openclaw/pull/142366/commits/1cbbd65c5224325d87e8946a632bb871ce7bbb66)
- Landed commit: [39d494b4324ee8e18e2cf187e4ad076bd4bb4988](https://github.com/openclaw/openclaw/commit/39d494b4324ee8e18e2cf187e4ad076bd4bb4988)

PR #142369, opened 2026-09-08
fix(matrix): keep escaped mentions literal after unmatched backticks

steipete:
<details>

This PR uses a branch in the same repository, so maintainers can update it directly.

</details>

## What Problem This Solves

Fixes an issue where Matrix messages containing an unmatched literal backtick could turn a later escaped user or room ID into a notification mention. For example, `` `literal then \@alice:example.org `` produced a mention link and `m.mentions` metadata instead of keeping the escaped ID literal.

## Why This Change Was Made

Preserve Markdown-it's canonical escaped-`@` token before its text-joining step removes that distinction. This lets the parser own code, escape, image, and link boundaries and removes the duplicate backtick scanner, private-marker restoration, and token copies used only by that restoration path.

## User Impact

Escaped user and room IDs remain literal even when surrounding Markdown is incomplete. Normal mentions continue to work, and existing code, link, image, spoiler, and message-body behavior is preserved.

## Evidence

- Four new formatter/send regressions failed before the fix, while 189 existing controls passed. Afterward, all **208 selected tests** passed across formatting, send preparation, automatic replies, and preparation reuse.
- Actual local message-content construction and enrichment failed before and passed after the fix: body bytes stay unchanged, escaped IDs produce literal HTML and empty `m.mentions`, and the normal mention control retains its link and metadata. This used a synthetic identity callback and made no sends or network requests; it is not a live Matrix service test.
- Full `check:changed` passed, including production and owning test typegraphs, full-file lint, formatting, export checks, and zero runtime import cycles. Independent P0–P2 review was scoped-clean.
- The pinned Markdown-it escape, backtick, and text-joining implementations were inspected directly. Existing escaped image/link and private-character controls pass.
- Production code: **17 lines added, 72 removed; net −55**. Tests: net +18 lines. No new dependency, configuration, or payload field.

### DIFF
--- a/extensions/matrix/src/matrix/format.test.ts
+++ b/extensions/matrix/src/matrix/format.test.ts
@@ -543,6 +543,16 @@ describe("markdownToMatrixHtml", () => {
       html: "<p>`literal then @alice:example.org</p>",
     },
     {
+      name: "keeps escaped mentions literal after unmatched backticks",
+      markdown: "`literal then \\@alice:example.org",
+      html: "<p>`literal then @alice:example.org</p>",
+    },
+    {
+      name: "keeps escaped room mentions literal after unmatched double backticks",
+      markdown: "``literal then \\@room",
+      html: "<p>``literal then @room</p>",
+    },
+    {
       name: "restores escaped mentions in markdown link labels without linking them",
       markdown: "[\\@alice:example.org](https://example.com)",
       html: '<p><a href="https://example.com">@alice:example.org</a></p>',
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [266137ce894f45d08c4870624d181300734984f8](https://github.com/openclaw/openclaw/pull/142369/commits/266137ce894f45d08c4870624d181300734984f8)
- Landed commit: [4e01e2fd5e789c06cc57c8de1e7a772ca22f076b](https://github.com/openclaw/openclaw/commit/4e01e2fd5e789c06cc57c8de1e7a772ca22f076b)

PR #142371, opened 2026-09-08
fix: avoid unnecessary diagnostic loading during CLI startup

steipete:
## What Problem This Solves

Fixes unnecessary startup work when users launch OpenClaw or prepare a cloud computer on a supported runtime. The supervising CLI eagerly loaded the diagnostic formatter and its logging/configuration dependencies even when no runtime error needed reporting.

## Why This Change Was Made

Load the existing diagnostic formatter only after runtime validation fails and no caller-provided error sink exists. Await validation in each existing caller so unsupported runtimes still stop before configuration selection, onboarding, or worker startup. Use the existing normalization helper subpath to avoid another unrelated barrel import.

## User Impact

Healthy startup no longer loads this failure-only dependency graph. Node/Bun/SQLite requirements, configured JSON diagnostics and redaction, supervised respawn, and worker behavior remain unchanged. This is a bounded startup improvement; it does not by itself make cloud desktops ready in under ten seconds.

## Evidence

- Regression proof failed on the old eager diagnostic import and on a caller that ignored asynchronous validation; both pass with the repair.
- 598 focused and sibling tests passed, covering runtime predicates and errors, CLI ordering, onboarding, entry/respawn, and sealed worker construction.
- Full `pnpm build` passed, including the single-file worker build contract.
- Independent Codex review found no actionable P0–P2 findings.
- The first changed check hit a pre-existing UI `refreshIfDue` type error; the branch now includes the canonical main repair from #142346. The refreshed check then caught a parameter-reassignment lint error in this patch. That is corrected, with 333 focused tests and targeted lint passing; the final full changed check passed on `c9649042`; the exact-head full build passed as well.

Production code grows by one line. The exact build has a nine-file / 36,248-byte static runtime-guard graph; diagnostic formatting is reachable only through its failure-path dynamic import. No before/after timing claim is inferred from import counts.

Live validation on the exact head:

- The dev Gateway started healthily, and the retained macOS node returned a real native screenshot.
- A fresh AWS session created in 670 ms and became active in 265 seconds. It forked the existing desktop image and installed the changed runtime; this is not an identical-runtime warm-reuse comparison.
- The real OpenAI-backed agent completed Computer screenshots, typing, Return, and marker verification before WebVNC opened. It then completed another fresh-frame Computer input turn while the real Chrome WebVNC viewer remained connected. Both markers were inspected in the Computer results and the rendered viewer; no native VNC client or manual terminal input substituted for agent actions.
- An earlier AWS attempt failed: its desktop stream closed immediately after the first Computer screenshot call, the turn timed out, and normal reclaim hit the admission-drain deadline. A bounded diagnostic found both SSH ports unreachable and never ran remotely. Documented forced teardown completed, the provider confirmed release, and the session was archived. The cause is unresolved; the passing comparison does not erase that failure or prove which component caused it.

The ten-second desktop-readiness target is not met. These timings are absolute observations, not a claimed speedup from the import change. Normal cleanup of the passing comparison completed: the provider confirmed release, the allocation pin was removed, and the session was archived. Both Linux test environments are gone. The prepared candidate image remains under the existing retention policy.

### DIFF
--- a/src/cli/run-main.exit.test.ts
+++ b/src/cli/run-main.exit.test.ts
@@ -55,7 +55,7 @@ const normalizeEnvMock = vi.hoisted(() => vi.fn());
 const pinConfigDirMock = vi.hoisted(() => vi.fn());
 const pinRuntimePathsMock = vi.hoisted(() => vi.fn());
 const ensurePathMock = vi.hoisted(() => vi.fn());
-const assertRuntimeMock = vi.hoisted(() => vi.fn());
+const assertRuntimeMock = vi.hoisted(() => vi.fn(async () => {}));
 const closeActiveMemorySearchManagersMock = vi.hoisted(() => vi.fn(async () => {}));
 const hasMemoryRuntimeMock = vi.hoisted(() => vi.fn(() => false));
 const listRegisteredAgentHarnessesMock = vi.hoisted(() => vi.fn((): unknown[] => []));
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c9649042bbb73c262ab2308e2ae93fa80e8be427](https://github.com/openclaw/openclaw/pull/142371/commits/c9649042bbb73c262ab2308e2ae93fa80e8be427)
- Landed commit: [27da5a4bb3200b22509250a9a9673c704d180b3f](https://github.com/openclaw/openclaw/commit/27da5a4bb3200b22509250a9a9673c704d180b3f)

PR #125969, opened 2026-08-18
fix(gateway): scope image capability discovery

RomneyDa:
## Summary

- scope attachment image-capability fallback to the selected provider instead of rebuilding every provider catalog
- keep runtime-only models by using the existing request-scoped live catalog path
- retry the complete Gateway projection when its published owner is replaced during scoped discovery

## Root cause and owner

Attachment admission first reads the lifecycle-published catalog. When a selected runtime-only model is absent or has provisional text-only metadata, current main falls back to readOnly false, which rebuilds the full cross-provider catalog inside the request. The provider-scoped catalog machinery now exists on main, but this Gateway caller still does not use it.

The canonical owner is the Gateway catalog projection. The second read now carries the selected provider into the existing prepared request-scoped live catalog. The projection checks the published owner before and after the asynchronous scoped read and reacquires the full owner on replacement, so catalog, config, agent identity, and auth projection never mix generations.

## ClawSweeper findings

- Early owner replacement: fixed by retrying at the Gateway projection boundary; the regression replaces the owner during scoped discovery and proves only the replacement catalog returns.
- Final provider I/O fence: the old branch tried to turn generation liveness into a public provider-auth revocation contract, but that was incomplete and added a large cross-plugin surface. This port removes that contract. Published-generation liveness governs whether results may publish; it does not retroactively revoke already-started provider requests. Results from a retired generation are discarded and the owner is reacquired.
- Runtime-only models: retained. An independent P1 review caught and rejected an intermediate static-only rewrite; the final branch uses the existing scoped live mode.

## Scope

- production: +76/-40, net +36
- tests: +67/-13, net +54
- no provider/plugin implementation changes, new config, cache, timeout, retry-on-provider-failure, or broad fallback path

Positive production growth is the internal Gateway load-parameter contract plus the atomic scoped projection. It replaces duplicated loader parameter shapes and reuses existing prepared-catalog implementation.

## Validation

- exact head: 1da26e9410a268368ad13de94fee4ec500105ca7
- base: 1fcfebc11ebb4c638fb385089f3cd58f7dd08884
- git diff --check: clean
- independent P1 autoreview: scoped-clean; no findings
- exact-head CI: pending
- ClawSweeper: fresh exact-head review requested after this rewrite
- local tests/checks not run because this isolated worktree has no installed dependencies; GitHub CI is authoritative

Agent transcript omitted.

PR #142375, opened 2026-09-08
fix(models): stop model checks from rediscovering providers

obviyus:
Related: #136257, #125969, #142100.

## What Problem This Solves

Fixes an issue where an ordinary session model-selection command started provider discovery even though browsing should use published inventory.

## Why This Change Was Made

Turn capability, subagent/hook validation and probe planning read the published catalog. Remove separate live and manifest fallbacks for missing facts, retain catalog ownership through asynchronous projection, and reuse configured rows already captured by visibility policy.

## User Impact

Model selection saves the requested model without unrequested discovery. Explicit Refresh still acquires new inventory. Existing route checks, tool refusals and authentication boundaries remain.

Consumers: model selection, capability validation and public cached catalog reads use published facts.

## Evidence

A real Gateway reproduced two unrequested discovery calls before the fix. The candidate saved the exact selection with no catalog or inference requests; explicit Refresh then published the model. Fifty-nine focused controls passed. Compiled public-interface checks confirmed passive getters; standalone installation and external inference were not tested.

### DIFF
--- a/docs/concepts/models.md
+++ b/docs/concepts/models.md
@@ -91,8 +91,10 @@ configured or completed catalog. They do not start provider discovery, including
 after a restart. If the owner is not yet published, the request reports that the
 catalog is not ready. Retry after startup or an in-progress refresh finishes.
 Use an explicit Refresh action or `openclaw models list --refresh` to acquire
-provider inventory. Startup and turn-path capability preparation keep their
-separate runtime responsibilities.
+provider inventory. Model selection, subagent capability checks, and hook-model
+validation also use the published inventory. Missing capability facts do not
+start another provider discovery. Native runtime observations keep their separate
+owner and authentication requirements.
 For models configured to use a CLI runtime, channel picker availability follows that
 runtime's prepared authentication; a provider API key does not substitute for its
 native login.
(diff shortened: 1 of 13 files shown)


obviyus:
**Proof:** A real authenticated Gateway model command saves `ollama/catalog-probe:latest` without starting provider discovery. Explicit Refresh then acquires and publishes its reasoning and text-input capabilities.

| Before | After |
| --- | --- |
| Model selection sends two unrequested catalog requests. | Model selection sends no catalog or inference requests; explicit Refresh sends the two catalog requests. |

- The shared capability reader uses published facts and retains temporary ownership through asynchronous projection. Connected fallback and duplicate getter paths are removed.
- Validation: 59 distinct focused controls, a compiled public SDK getter/cache-only sequence before and after Refresh, qualified independent public acceptance, exact-head local ClawSweeper review, and [green required CI](https://github.com/openclaw/openclaw/actions/runs/34260312472).
- The public run uses the compiled checkout and read-only source plugin overlays. It does not claim a standalone installation or stable-release upgrade. Earlier failed attempts and missing historical receipts remain in the retained evidence.

PR #120171, opened 2026-08-07
fix(progress): preserve callback acceptance results

joshavant:
Related: #115914

<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers
can help update the branch when needed.

</details>

## What Problem This Solves

Resolves a problem where channel progress callbacks could lose their acceptance result, async legacy callbacks resolving `void` could be misclassified, and rejected compositor renders could become the dedupe baseline instead of remaining retryable.

Extracted from and supersedes the compatibility-contract portion of #119921 by @joshavant; this PR alone does not close #115914.

## Why This Change Was Made

Progress callbacks now use a backward-compatible `boolean | void` contract: `true` means accepted and visible, `false` means not visible or still pending, and both synchronous and asynchronous `void` remain legacy-visible. Core preserves that result through followups, direct-run wrappers, ordered typing/presentation, embedded/CLI callback mirrors, and compositor boundaries. The compositor records a dedupe baseline only after an accepted render.

Maintainer confirmation: the requester's prior `Approved` and current execution approval confirm this public `boolean | void` Plugin SDK contract and authorize applying it consistently across the affected core and plugin surfaces.

This PR intentionally does not add the receipt timer, receipt coordinator, listener API, or late transport-acknowledgement lifecycle. Original #119921 remains open until both replacement PRs land.

## User Impact

Channels gain a truthful, retry-safe foundation for reporting visible progress without breaking existing callback implementations. Operators keep current behavior for legacy plugins, while acceptance-aware channels can distinguish visible work from pending or rejected updates.

## ClawSweeper Rank-up Moves

- Direct embedded and CLI progress callbacks now propagate `false`, legacy `void`, and `true` through ordered presentation; a rejected direct partial no longer suppresses the original terminal failure.
- Discord preamble delivery now returns the combined headline/commentary acceptance result, and an unacknowledged identical preamble remains retryable until Discord acknowledges it.
- Matrix and Discord return `false` until their flushed draft has an acknowledged event/message id.
- Native Slack commits its chunk key, task snapshot, reply-plan state, and delivery facts only after `session.delivered`; buffered identical updates remain retryable.
- Added direct terminal-failure and Discord false-then-ack regressions alongside the existing Matrix, Discord, and Slack fault-injection coverage.

## Evidence

- Head: `99d595aebfcddaab9162bf4dff850dff425536d1`
- Tree: `bf445443da3172c6445d17e88ea8617d15e65355`
[...]

PR #142378, opened 2026-09-08
fix(release): qualify frozen Telegram scenario selection

RomneyDa:
## Summary

- classify the selected package source before running default Telegram package scenarios
- omit only `telegram-partial-failure-recovery` for the exact pre-recovery Telegram source contract
- keep explicit scenario requests strict and keep current, unknown, partial, and unauthorized targets unchanged

## Root cause

The 2026.6.35 candidate predates the partial-failure recovery contract added by #120626 and its earlier cross-channel callback-acceptance prerequisite #120171. Current trusted package validation nevertheless selected that current-only scenario from the default catalog. Backporting the product behavior is not a bounded stable repair: the prerequisite spans 50 files and changes callback acceptance across channels.

The release workflow already has an explicit frozen-target omission authorization. This change resolves applicability from the exact selected checkout, then forwards only the closed scenario omission—not the authorization or SHAs—into the package runner.

## Validation

- exact `b91472789839dccc59f8d5618c8082c4f942e0d3` source resolves only `telegram-partial-failure-recovery` as default-inapplicable
- `test/scripts/npm-telegram-live.test.ts`: 48/48
- Telegram-focused `test/scripts/package-acceptance-workflow.test.ts`: 61/61
- `actionlint` on both workflows
- `tsgo:test:root`
- Oxfmt, OXLint, bash syntax, diff check
- AutoReview: scoped-clean, no P0/P1, confidence 0.94

## LOC

Release tooling/workflows: +90/-1. Tests: +40. Application runtime: +0.

Production growth is the source-identity/applicability boundary and its workflow transport. No product compatibility shim, version/SHA allowlist, or alternate Telegram runtime path is added.


RomneyDa:
@clawsweeper review

PR #142402, opened 2026-09-08
refactor(agents): group equivalent Nodes dispatch cases

### LINKED ISSUE #142397: Group repeated Nodes tool dispatch cases
## Problem

The Nodes tool repeats identical media and command executor argument objects in several switch branches. This maintainer-requested cleanup can group equivalent actions around their existing executor calls.

## Scope

Preserve every action, argument, return value, error path and permission boundary. Use normal switch narrowing instead of the unnecessary command-action cast, and make only the corresponding assertion-baseline decrease.

## Validation

The existing 87 focused Nodes tests pass. Production/test typechecks, the complete changed-file gate and independent P0–P2 review pass. The cast/import follow-up emits the same runtime JavaScript. No new helper, test, configuration option or API is needed.

steipete:
Closes #142397

## What Problem This Solves

The Nodes tool repeats identical executor argument objects across its media and command switch cases. This maintainer-requested cleanup groups equivalent cases around the existing calls.

## Why This Change Was Made

The existing media and command executors retain every action, argument and return path. Switch narrowing also removes an unnecessary command-action cast. The internal assertion-baseline count decreases from 3 to 2 to match that removal; no user configuration changes.

## User Impact

No intended behavior change. The complete change removes 68 production code lines and 68 physical lines, without a new runtime abstraction or test seam.

## Evidence

- 87 focused tests pass across three existing Nodes suites.
- The final changed-file gate passes on the source and exact assertion-baseline decrease, including production/test typing, lint and dead-export checks.
- Fresh independent P0–P2 review is scoped-clean.
- The subsequent cast/import removal emits byte-identical JavaScript with object shorthand downleveled for comparison. Runtime tests were not repeated for that type-only follow-up; this is not a production bundle comparison.

No schema, protocol, permissions or persisted-data change is included. No live device was used.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -2009,7 +2009,7 @@ src/agents/tools/message-tool-visible-content.ts	9
 src/agents/tools/mobile-ui-tool.ts	5
 src/agents/tools/model-config.helpers.ts	1
 src/agents/tools/music-generate-tool.ts	2
-src/agents/tools/nodes-tool.ts	3
+src/agents/tools/nodes-tool.ts	2
 src/agents/tools/openclaw-delegate-tool.ts	1
 src/agents/tools/pdf-native-providers.ts	2
 src/agents/tools/pdf-tool.ts	3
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ce59aad0e714b1dc0f5b58f516ec8957ee1a482d](https://github.com/openclaw/openclaw/pull/142402/commits/ce59aad0e714b1dc0f5b58f516ec8957ee1a482d)
- Landed commit: [ae51fd17a641e82b814bf137d0b6c43de73af59a](https://github.com/openclaw/openclaw/commit/ae51fd17a641e82b814bf137d0b6c43de73af59a)

PR #142411, opened 2026-09-08
refactor(tests): remove dead Telegram edit scaffolding

steipete:
## What Problem This Solves

The Telegram edit table retains fallback setup and assertions that none of its three rows can reach. Every row disables that fallback and expects exactly one edit call. A long-message check also repeats a condition already guaranteed by its exact formatted-span count.

## Why This Change Was Made

Remove the dead fallback branches and uniform row metadata, assert the shared call count directly, and rename the third case to describe its actual HTML path. Keep all three invocations, their inputs, mock behavior, and meaningful keyboard and formatting assertions.

## User Impact

No production changes. Removes 21 net test lines and makes the retained coverage clearer.

## Evidence

- The full send suite passes 276/276 before and after, with no skips. Exactly one title changes from “rich edit preserves cleared keyboard” to “HTML edit preserves cleared keyboard”; the other 275 ordered identities and all statuses remain unchanged.
- Static checks verify all three old fallback predicates are false, all expected call counts are one, and no row has the second-call expectation. The retained mock statement and production-call input expression are identical.
- The mapped changed gate and fresh independent P2 review pass with no findings.

No runtime improvement is claimed.

### DIFF
--- a/extensions/telegram/src/send.test.ts
+++ b/extensions/telegram/src/send.test.ts
@@ -1826,7 +1826,6 @@ describe("sendMessageTelegram", () => {
     const chunks = sendMessageTexts(botApi.sendMessage);
     const joinedChunks = chunks.join("");
     expect(joinedChunks).toContain("Long");
-    expect(joinedChunks).toContain("section");
     expect(joinedChunks.match(/<b>section<\/b>/g)).toHaveLength(200);
     expect(joinedChunks.match(/<i>style<\/i>/g)).toHaveLength(200);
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c7c8fd15e43158ab25644d9e2b91418451288690](https://github.com/openclaw/openclaw/pull/142411/commits/c7c8fd15e43158ab25644d9e2b91418451288690)
- Landed commit: [253f76a882f040d1219612a268dd45bf31873b8f](https://github.com/openclaw/openclaw/commit/253f76a882f040d1219612a268dd45bf31873b8f)

PR #142415, opened 2026-09-08
fix(tts): keep ordinary Markdown prose in spoken replies

steipete:
<details>

Keep **Allow edits from maintainers** enabled for this PR so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Fixes an issue where users receiving speech replies with inline code or prose after a quoted code block would hear no audio, or Talk would replace the prose with its short screen fallback. Code blocks opened directly on a list marker could also be missed by the code-heavy filter.

## Why This Change Was Made

Speech classification used a separate approximate fence scanner that disagreed with the Markdown renderer. Remove it and count actual fenced content through the existing cached Markdown parser. Preserve the inclusive half-code threshold and code whitespace by excluding only the parser's terminal framing line break.

## User Impact

Ordinary prose with inline-code examples remains eligible for speech. Actual fenced code, including nested list forms, follows the existing code-heavy behavior. Explicit speech overrides and normal speech rendering keep their current contracts.

## Evidence

Before the repair, both auto-TTS caller regressions failed because synthesis was never invoked; both Talk regressions received the canned fallback instead of the requested prose. The three shared classifier regressions also failed. After the repair, 110 focused tests pass across the classifier, exact fenced-content accounting, Markdown block metadata, auto-TTS runtime, and Talk runtime. The two changed Talk cases passed again after completing a required field in the fake provider result.

Full changed-scope validation passes, including core production/test typing, full-file lint, dead-export scans and guards. Managed P2 review is scoped-clean. One initial cold Talk import timed out before assertions; the unchanged retry reached the real regression, and the repaired full suite passed without changing timeouts or widening mocks. Proof used synthetic text/audio with fake providers through production callers; no external provider or live channel session was used.

The change deletes 86 net production lines by removing the duplicate parser. No configuration, dependency, storage, protocol or SDK return-shape changes are included.

### DIFF
--- a/docs/nodes/talk.md
+++ b/docs/nodes/talk.md
@@ -18,6 +18,8 @@ Talk mode covers these runtime shapes:
 
 Native Talk is a continuous loop: listen for speech, send the transcript to the model through the active session, wait for the response, then speak it via the configured Talk provider (`talk.speak`).
 
+For replies dominated by fenced code, `talk.speak` uses a short spoken message directing the listener to the screen. Inline code and ordinary prose remain part of the spoken reply.
+
 Apple Watch also retains **Talk to Claw**, the separate [one-turn companion flow](/platforms/ios#talk-to-claw-with-the-iphone): native dictation, text relayed through the iPhone, and system-voice readback. **Talk on Watch** is the realtime path included in normal Watch setup; see [standalone voice setup](/platforms/ios#standalone-voice).
 
 ## Choose a Talk voice from chat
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6a5a303461218b4841f582b4225515380d621772](https://github.com/openclaw/openclaw/pull/142415/commits/6a5a303461218b4841f582b4225515380d621772)
- Landed commit: [884927b88c2f62d769f60fd401600dec1dedde07](https://github.com/openclaw/openclaw/commit/884927b88c2f62d769f60fd401600dec1dedde07)

PR #142417, opened 2026-09-08
fix(memory): avoid repeated local provider initialization across registrations

### LINKED ISSUE #142414: [Bug]: Local memory search reinitializes the same provider across registrations
### Bug type

### Beta release blocker

No

### Summary

Local memory search can replace an otherwise compatible manager because the same host local-service acquirer receives a new function identity at three plugin runtime boundaries.

### Steps to reproduce

1. On `main@b138ae4f1e40ea35519c282efe511fc725b2b67a`, create two plugin runtime facades and compare their `llm.acquireLocalService` functions.
2. Resolve `api.runtime.llm.acquireLocalService` twice for the same plugin and host, then compare the functions.
3. Register `memory-core` twice, resolve each lazy memory runtime, and compare the injected host acquirers.
4. Observe that all three same-owner comparisons return different function identities.
5. Pass those functions to the memory manager registry. Its compatibility key assigns host identities from the acquirer function, so the otherwise equivalent requests receive different keys.

The regression tests in the linked PR reproduce all three identity changes and fail before the fix.

### Expected behavior

The local-service acquirer keeps one identity for the same process host and plugin while remaining distinct for a different host callback or plugin. Repeated requests for the same memory configuration can then reuse the compatible manager selected by the registry.

### Actual behavior

The runtime facade, plugin-scoped runtime accessor, and `memory-core` registration each allocate another forwarding callback. The manager registry treats those callbacks as different local-service hosts and replaces the compatible manager.

### OpenClaw version

`main@b138ae4f1e40ea35519c282efe511fc725b2b67a`

### Operating system

macOS 26.6.2

### Install method

pnpm development checkout

### Model

N/A (the defect occurs in local embedding service setup before a chat-model request)

### Provider / routing chain

OpenClaw `memory-core` -> local embedding provider service

The deterministic reproduction uses synthetic runtime hosts and does not require credentials, a live provider, or private configuration.

### Logs

```shell
same process runtime host: different callback identity
same plugin and host: different callback identity
two memory-core registrations: different callback identity
```

The linked PR contains focused regression coverage at the runtime facade, plugin scope, and `memory-core` registration boundaries. Before the production change, the four new identity assertions fail; after it, the three focused test files pass 77/77 tests.

### Impact and severity

- Affected: users of local memory embeddings when the same logical manager is reached through repeated plugin runtime creation or registration.
- Severity: medium; repeated cold initialization can delay or time out memory search.
- Frequency: the callback mismatch is deterministic at all three reproduced boundaries; user-visible timing depends on local provider startup cost.
- Consequence: a reusable memory manager is replaced and its local embedding service is initialized again.

### Additional information

NOT_ENOUGH_INFO

artdaal:
Closes #142414

## What Problem This Solves

Fixes an issue where users of local memory embeddings could pay repeated provider initialization and hit memory-search timeouts when the same memory configuration crossed repeated runtime or plugin registration paths.

## Why This Change Was Made

The runtime now exposes one process-stable host acquirer, the plugin registry reuses one scoped acquirer per host callback and plugin ID, and `memory-core` reads that reference lazily instead of allocating another forwarding function. Different plugins and genuinely different host callbacks remain isolated. The change does not alter memory configuration, index schema, ranking, or provider fallback behavior.

## User Impact

Repeated native, HTTP, CLI, or plugin registration paths can reuse a compatible local memory manager instead of replacing it solely because a wrapper function received a new identity. This avoids unnecessary local-provider cold starts while preserving plugin ownership scope.

## Evidence

- Before the production change, four new assertions fail at the runtime facade, plugin scope, and `memory-core` registration boundaries because the callback identity changes.
- Focused regression and sibling tests: 77 passed across `src/plugins/runtime/index.test.ts`, `src/plugins/registry.runtime-config.test.ts`, and `extensions/memory-core/index.test.ts`.
- Plugin contract suite: 47 files and 1117 tests passed.
- `pnpm check:changed` passed for the selected core, core-test, extension, and extension-test lanes.
- `pnpm build` passed, including CLI bootstrap imports, plugin SDK exports, runtime postbuild checks, and Control UI budgets.
- Independent autoreview found no actionable P0-P2 findings (`scoped-clean`, confidence 0.96).

### Redacted real-runtime proof

Before opening this PR, the equivalent three-boundary identity repair was exercised on an installed OpenClaw 2026.9.2 runtime using a real local GGUF embedding provider and a populated native SQLite memory index. A content-free profiler observed a cold native search, an HTTP search, and a fresh native search in one process:

```text
route   acquirer_token   manager_token   acquire_ms   search_ms   completed
native  1                2               352          5626        yes
HTTP    1                2               0            851         yes
native  1                2               0            1046        yes
```

All three requests reached `search-end`; the acquirer and manager tokens stayed identical across the registration/route transitions, and the two warm acquisitions returned immediately. The profiler was then removed byte-for-byte. A separate uninstrumented cold native -> HTTP -> native sequence completed without the prior 15-second timeout (8,886 ms, 839 ms, and 1,051 ms respectively), and every search returned without an error.

Queries, result content, paths, hostnames, endpoints, IP addresses, and credentials are omitted. This operational run used the installed 2026.9.2 build with the equivalent guarded fix at the same host-runtime, plugin-scope, and `memory-core` boundaries; the regression suites above validate the source implementation in this PR.

### DIFF
--- a/extensions/memory-core/index.test.ts
+++ b/extensions/memory-core/index.test.ts
@@ -400,6 +400,17 @@ describe("memory-core plugin runtime registration", () => {
     });
   });
 
+  it("keeps the host local-service hook stable across plugin registration", async () => {
+    const firstRuntime = registerMemoryCoreRuntime();
+    await firstRuntime.getMemorySearchManager({ cfg: {}, agentId: "main" });
+    const firstHost = createMemoryRuntimeMock.mock.calls.at(-1)?.[0];
+    const secondRuntime = registerMemoryCoreRuntime();
+    await secondRuntime.getMemorySearchManager({ cfg: {}, agentId: "main" });
+    const secondHost = createMemoryRuntimeMock.mock.calls.at(-1)?.[0];
+
+    expect(firstHost?.acquireLocalService).toBe(secondHost?.acquireLocalService);
+  });
+
   it("defers nested host runtime access until the injected operation runs", async () => {
     const acquireLocalService = vi.fn(async () => undefined);
     const openKeyedStore = vi.fn(() => ({}));
@@ -427,12 +438,13 @@ describe("memory-core plugin runtime registration", () => {
     expect(stateGetter).not.toHaveBeenCalled();
     await runtime?.getMemorySearchManager({ cfg: {}, agentId: "main" });
     const injectedHost = createMemoryRuntimeMock.mock.calls.at(-1)?.[0];
-    if (!injectedHost?.acquireLocalService || !injectedHost.openKeyedStore) {
+    const injectedAcquireLocalService = injectedHost?.acquireLocalService;
+    if (!injectedAcquireLocalService || !injectedHost?.openKeyedStore) {
       throw new Error("expected memory-core host operations");
     }
 
     const target = { providerId: "local", baseUrl: "http://127.0.0.1:11434" };
-    await injectedHost.acquireLocalService(target);
+    await injectedAcquireLocalService(target);
     const storeOptions = { namespace: "lazy-host", maxEntries: 1 };
     injectedHost.openKeyedStore(storeOptions);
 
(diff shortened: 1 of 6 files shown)


artdaal:
@clawsweeper re-review

PR #142433, opened 2026-09-08
fix(opencode-go): give 7.33 live probes stable session identity

RomneyDa:
## Summary
- supply a stable `sessionId` to direct OpenCode Go live-model probes
- keep other provider probes unchanged
- cover the provider-scoped identity contract

## Why
The 7.33 provider wrapper correctly derives `x-opencode-session` from normal agent-turn session IDs. The direct release live-model probe bypassed that normal caller contract and supplied no session ID, so OpenCode rejected all four Go models with HTTP 400.

## Validation
- `pnpm exec vitest run --config test/vitest/vitest.live.config.ts src/agents/models.profiles.live.test.ts --testNamePattern resolveLiveModelSessionId`
- `pnpm exec oxfmt --check src/agents/models.profiles.live.test.ts`

### DIFF
--- a/extensions/opencode-go/index.test.ts
+++ b/extensions/opencode-go/index.test.ts
@@ -432,6 +432,40 @@ describe("opencode-go provider plugin", () => {
     ]);
   });
 
+  it("adds the OpenCode session header to direct completions", async () => {
+    const provider = await registerSingleProviderPlugin(plugin);
+    const capturedOptions: unknown[] = [];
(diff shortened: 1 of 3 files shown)


RomneyDa:
@clawsweeper re-review

Connected the existing OpenCode Go session-header wrapper to `wrapSimpleCompletionStreamFn` and added provider-level coverage that captures the outgoing `x-opencode-session` header.

Validation:
- OpenCode Go extension batch: 6 files, 35 tests passed
- Live-provider unit config: 37 passed, 1 skipped

PR #142434, opened 2026-09-08
fix(whatsapp): avoid duplicate sends after socket failures

### LINKED ISSUE #142398: [Bug]: WhatsApp retries can change message identity after ambiguous failures
### Bug type

Behavior bug (incorrect delivery state without a crash)

### Beta release blocker

No

### Summary

WhatsApp has nested retry owners that can replay one logical outbound send with a newly generated provider message ID after an ambiguous socket failure.

### Steps to reproduce

1. Use the bundled WhatsApp plugin with `baileys@7.0.0-rc14`.
2. Send a normal auto-reply through an attached WhatsApp socket.
3. Make the first `sendMessage` attempt fail after publication may have begun with an error treated as retryable.
4. Observe that the socket session retries `sendMessage`, while the auto-reply delivery layer can independently retry the whole transport call.
5. Observe that neither layer supplies a stable `messageId`; Baileys therefore generates a new ID for each call.

### Expected behavior

One logical outbound send should have one retry owner and one stable provider message ID across every in-process attempt. A later message should receive a different ID.

### Actual behavior

The same logical send can be attempted by both the socket-session retry loop and the outer auto-reply retry wrapper. Each call can generate a different WhatsApp message ID, so an ambiguous first attempt can become a second recipient-visible message instead of the same idempotent attempt.

### OpenClaw version

`main` at `d4889f5c31231b15c0aff17a1bdd2d868221e47b`

### Operating system

macOS 26.6.2; the affected retry logic is platform-independent.

### Install method

Git source checkout

### Model

N/A — the defect is in channel delivery after model output.

### Provider / routing chain

OpenClaw outbound delivery → bundled WhatsApp plugin → `baileys@7.0.0-rc14`

No model or provider setting affects this behavior.

### Logs

```shell
NOT_ENOUGH_INFO
```

Current source has two retry owners:

- `extensions/whatsapp/src/inbound/socket-session.ts` retries `sendMessage` across socket failures.
- `extensions/whatsapp/src/auto-reply/deliver-reply.ts` wraps the transport call in `sendWhatsAppOutboundWithRetry`.

The pinned Baileys source in `lib/Socket/messages-send.js` generates `messageId` inside `sendMessage` unless the caller supplies one. A focused regression test confirms that supplying one ID reuses it across retries while a later logical send receives a different ID.

Related but distinct:

- #116232 narrows which error strings are retried but retains the nested retry topology and does not preserve a message ID.
- #122758 concerns the later provider-ack custody boundary.
- #130024 concerns restart recovery after durable state reaches `send_attempt_started`.

### Impact and severity

Affected: WhatsApp users when a send encounters an ambiguous retryable transport failure.

Severity: High message-delivery correctness risk.

Frequency: Edge condition tied to socket failure timing.

Consequence: A recipient can receive duplicate messages with different provider IDs.

### Additional information

The smallest repair is to keep retry ownership in the attached socket session, assign one message ID before its retry loop, and remove the outer auto-reply retry. Accepted partial-delivery evidence must continue to prevent replay. This does not require a configuration, SQLite, session, or transcript migration.

ashawwal:
Closes #142398

## What Problem This Solves

Fixes an issue where a WhatsApp reply could be sent more than once when an ambiguous socket failure caused nested retry layers to replay one logical send with different provider message IDs.

## Why This Change Was Made

The attached socket session becomes the single in-process retry owner and assigns one WhatsApp message ID before its retry loop. The outer auto-reply layer no longer retries the whole transport call, so accepted or ambiguous work cannot be restarted under a new identity.

This does not change provider-ack custody, durable restart reconciliation, configuration, SQLite schemas, sessions, or transcripts.

## User Impact

WhatsApp users are less likely to receive duplicate replies during reconnects or ambiguous send failures. Separate logical messages still receive distinct IDs, and terminal failures remain visible to the existing delivery owner.

## Evidence

Validated at `163f74da53e` against frozen upstream base `a92c15a5dc3`.

The candidate-head regression now crosses the real pinned Baileys client boundary:

- OpenClaw's production `sendTrackedMessage` owner calls Baileys' real `sendMessage`, message construction, and `relayMessage` paths;
- the injected low-level `sendNode` records the first outbound binary stanza and then throws an ambiguous post-write timeout;
- OpenClaw retries through the real client, and the accepted retry carries the same relay ID as the failed attempt;
- the next logical send reaches the transport with a distinct relay ID.

The existing lifecycle regression separately proves that the same identity survives replacement-socket recovery.

Candidate-head checks:

- real Baileys transport-boundary regression — 1 test passed
- extension production and test type checks — passed
- targeted extension lint, formatting, and `git diff --check` — passed

The preceding production-equivalent head also passed the full WhatsApp extension suite (1,401 tests); hosted CI will validate the new test on the literal candidate head.

This does not use a live WhatsApp account and does not claim direct recipient or server-ack observation. The injected seam is the low-level socket write beneath Baileys' real relay implementation. Provider acknowledgement loss and replay after process restart remain separate contracts.


ashawwal:
Maintainer CI rerun requested for the two failed leaf jobs on exact head d442fbffa39d699acd17af0cea85ee5334ddd88b. Both failures are outside this PR and pass when run directly on the PR head and frozen base; the PR Evidence section now records the exact tests, timings, and failure modes. GitHub rejected the contributor-side failed-job rerun because upstream Actions reruns require repository maintain/admin permission.

PR #142440, opened 2026-09-08
fix: remove per-file delays before cloud turns

steipete:
## What Problem This Solves

Fixes an issue where users running remote-exec cloud turns wait for dozens of small skill-file uploads before the model can start. A live session with 35 available skills spent 13.1 and 12.8 seconds preparing each turn: 56 serialized write requests carried only about 305 KB of encoded input.

## Why This Change Was Made

The existing transfer owner now packs file chunks into batches within the unchanged 128 KiB input limit. The receiver checks each chunk and re-enters the captured directory before writing another file. Fresh per-turn resources, hashes, read-only permissions, owner checks, cancellation, and cleanup keep their existing behavior; no cache, database schema, migration, configuration option, or protocol version is added.

## User Impact

Cloud and SSH remote-exec sessions avoid a network round trip for every small supporting file while still receiving current skill contents for each turn. Skill inputs remain private and excluded from ordinary Git staging and workspace reconciliation.

## Evidence

- Live before trace on main: 35 skills, 56 write requests per turn, 13,135 / 12,831 ms total transfer. Writes accounted for 12,504 / 12,346 ms; workspace recovery and computer preparation were negligible.
- Local SSH-script and node-runtime regression failed on the old implementation's 25 write requests, then passed with at most three writes for the same mixed large-file, small-file, empty-file, nested-path, and Unicode-name fixture.
- 79 targeted and sibling tests passed, including transfer bounds, byte preservation, permissions, cleanup, stale owners, and cancellation after initialization or a batch write.
- Changed-file checks passed, including core/test typechecking, lint, dead exports, and schema/store guards. The candidate build passed. Candidate live verification passed. Exact-head CI passed on attempt 2.

The live candidate transferred the same 35 skills in three writes per turn (maximum request: 131,069 bytes), taking 1.65 and 1.34 seconds instead of 13.14 and 12.83 seconds. Both actual Codex tool turns and file readbacks passed; the reused turn completed in 22.1 seconds. Normal Stop completed, the lease is released with cleanup complete, and all Gateway/tunnel/tail process owners exited. The serving coordinator stayed unchanged during the run. This comparison used hash-bound local timing instrumentation around the existing transfer calls; no source or build artifact was edited by that instrumentation.

The first candidate allocation encountered a coordinator Durable Object storage reset before a host became available. It was cancelled through normal reclamation and independently confirmed released/complete before this successful fresh retry. CI's first attempt failed only an unchanged proxy CLI subprocess test at its 30-second deadline. That case passed unchanged locally; only the failed same-head CI jobs were rerun, without timeout or assertion changes.

The automated stored-data classification is resolved by the command boundary: this JSON is stdin paired with the Node receiver script in the same `runWorkspaceCommand` request. Its `argv`/`input` envelope and 128 KiB bound are unchanged. The resulting file bytes, private directory layout, identity marker, and discovery/cleanup operations are unchanged. There is no independently deployed decoder or stored representation to migrate; SQLite schema/version guards passed. The review's live-proof and rank-up requests are covered by the completed candidate comparison above.

Post-merge verification also passed on an uninstrumented build of main `b7d79b5e08ae2a1e5e4bcb488aa3998d12644dcb`, using Crabbox CLI main `bb17902e8f029fd01e14b847f3e8fd23216db72b`. Both worker-turn and Codex remote-exec sessions completed initial, reused, and restored real Linux tool turns, with file readbacks and no-local-checkout assertions. All four final placements were destroyed, remaining task snapshots were deleted, and Gateway/tunnel/tail process owners exited cleanly. The serving Worker stayed unchanged during the run.


steipete:
Merged via squash.

- Prepared head SHA: [cd02214a5cb9a5dd04f645d1e562520939139a82](https://github.com/openclaw/openclaw/pull/142440/commits/cd02214a5cb9a5dd04f645d1e562520939139a82)
- Landed commit: [b7d79b5e08ae2a1e5e4bcb488aa3998d12644dcb](https://github.com/openclaw/openclaw/commit/b7d79b5e08ae2a1e5e4bcb488aa3998d12644dcb)

PR #142441, opened 2026-09-08
fix(e2e): keep frozen survivor tooling coherent

RomneyDa:
## What Problem This Solves

Full validation of the 2026.6.35 candidate mixes the selected historical upgrade-survivor scenario with current trusted helper modules. The historical runner hardcodes paths under `/app/scripts`, so it loaded the current npm fixture server against the selected target old bounded-response module and failed before testing the candidate. Failure diagnostics were also absent because the historical scenario predates the trusted diagnostics module.

## Why This Change Was Made

Keep the selected release scenario and assertions, but bind the current trusted npm fixture server, its bounded-response dependency, and diagnostics module as one coherent helper closure at the exact hardcoded paths. This does not add a compatibility mode or weaken scenario assertions.

## User Impact

Extended-stable validation can exercise historical upgrade scenarios without module-generation skew in the test harness. Product runtime is unchanged.

## Evidence

- Red: full validation run 34261943173, candidate release-check jobs 102187203871 and 102188432481. The fixture server failed because `createBoundedResponseTooLargeError` was missing from the selected old helper; diagnostics then failed with `MODULE_NOT_FOUND`.
- Green: focused `docker-build-helper` contract test; Bash syntax; Oxfmt; OXLint; diff check.
- Production/runtime LOC: 0. Release harness: +5/-2. Tests: +2/-0.

### DIFF
--- a/scripts/e2e/lib/clawhub-fixture-server.cjs
+++ b/scripts/e2e/lib/clawhub-fixture-server.cjs
@@ -77,6 +77,29 @@ async function assertNoRequests(baseUrl) {
   if (!Array.isArray(payload?.requests)) {
     throw new Error("ClawHub fixture request ledger must contain a requests array");
   }
+  const legacyPackage = process.env.OPENCLAW_FROZEN_UPGRADE_SURVIVOR_CLAWHUB_PACKAGE;
+  if (legacyPackage) {
+    const packagePath = `/api/v1/packages/${encodeURIComponent(legacyPackage)}`;
+    const artifactPrefix = `GET ${packagePath}/versions/`;
+    const artifactSuffix = "/artifact";
+    const artifactRequest = payload.requests[1] ?? "";
+    const version =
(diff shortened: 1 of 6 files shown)


RomneyDa:
@clawsweeper review

RomneyDa:
Extended exact validation exposed the second half of the same mixed-generation defect: the selected 2026.6.35 installer legitimately performs its historical three-request ClawHub npm-pack flow, while the current fixture asserted an idle ledger. The amended head derives that contract only from the authorized selected source, passes only the exact official package, and verifies the complete package/artifact/download sequence. Current/unknown targets remain strict-idle.

Proof: focused ClawHub + upgrade owner tests 14/14; Bash syntax, Oxfmt, OXLint, diff check; exact-head autoreview scoped-clean (P0-P2, 0.91).

@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

PR #142460, opened 2026-09-08
perf(plugins): keep provider discovery metadata lightweight

steipete:
## What Problem This Solves

Resolves unnecessary cold-import work when offline CLI commands prepare the model catalog. Loading metadata descriptors for Anthropic Vertex, DeepInfra and Google also loaded live catalog, transport or token code before that work was needed.

## Why This Change Was Made

Static metadata and live execution now have separate owners within the existing plugins. Vertex defers its catalog runtime and keeps environment/ADC facts separate from endpoint helpers. DeepInfra shares one static manifest/model-definition module between its descriptor and live catalog. Google keeps static Studio/Vertex models and synchronous ADC facts independent of live discovery and token exchange.

Existing public Vertex/DeepInfra API names and canonical model builders are retained. Google token refresh, caching and deadline logic stays with its existing owner. No new dependency, SDK subpath, configuration option, cache infrastructure, or shard change is introduced.

The tests enforce these import boundaries and retain real static output and live delegation coverage. A Vertex ADC fixture now initializes its fake reader explicitly instead of depending on an unrelated import. The moved ADC type boundary is documented, an unnecessary assertion is removed, and the old file's assertion baseline decreases from six to four.

## User Impact

Less startup work for metadata-only catalog reads. Both existing unknown-provider process tests remain intact; they protect separate CLI routes and a specific past JSON error regression.

Production changes are mostly moves, with three net added lines. Test changes add 104 net lines; the assertion-baseline adjustment is neutral in line count and reduces the old allowance.

## Evidence

- A fresh-process original/candidate/original Linux comparison, with each arm built from its corresponding source, measured the representative Commander JSON error case at **40.52s / 30.15s / 32.60s**. Source, loaded descriptors and compiled chunks were verified, and original descriptor bytes returned in the final arm. The variation between originals limits any general percentage claim.
- New import-boundary guards fail against the original eager imports and pass with the separated owners. Models and auth behavior remain covered by existing tests.
- Normal full build passed on Linux, including SDK/export/bootstrap checks and the UI build. **575 tests passed across 36 files:** 530 provider cases, all 28 unchanged help/exit cases, and 17 prepared-catalog CLI cases.
- Independent P2 review found no actionable issue. The final 30-path changed-file gate passed after the assertion annotation/baseline cleanup. Exact-head CI will be verified before landing.

The timing comparison and full runtime tests precede the final non-executable assertion annotation/baseline cleanup. Earlier diagnostic attempts and differing raw-CLI environments are not used as performance evidence. No live provider request was needed for the unchanged token implementations; existing transport/auth tests and the full build verify their retained boundaries.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -520,7 +520,7 @@ extensions/google/onboard.ts	3
 extensions/google/provider-registration.ts	2
 extensions/google/realtime-voice-provider.ts	6
 extensions/google/transport-stream.ts	13
-extensions/google/vertex-adc.ts	6
+extensions/google/vertex-adc.ts	4
 extensions/google/video-generation-provider.ts	13
 extensions/googlechat/src/accounts.ts	1
 extensions/googlechat/src/actions.ts	1
--- a/extensions/anthropic-vertex/api.ts
+++ b/extensions/anthropic-vertex/api.ts
@@ -17,12 +17,14 @@ export {
 export {
   hasAnthropicVertexAvailableAuth,
   hasAnthropicVertexCredentials,
-  resolveAnthropicVertexClientRegion,
   resolveAnthropicVertexConfigApiKey,
   resolveAnthropicVertexProjectId,
   resolveAnthropicVertexRegion,
-  resolveAnthropicVertexRegionFromBaseUrl,
 } from "./region.js";
+export {
+  resolveAnthropicVertexClientRegion,
+  resolveAnthropicVertexRegionFromBaseUrl,
+} from "./region-endpoint.js";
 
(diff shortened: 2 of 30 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [fd9b2e47cd899c999a27250ee6b8a48e5b5667fc](https://github.com/openclaw/openclaw/pull/142460/commits/fd9b2e47cd899c999a27250ee6b8a48e5b5667fc)
- Landed commit: [5fed661a46f62f85b5bf2c0dbf3f88ae9cf231d4](https://github.com/openclaw/openclaw/commit/5fed661a46f62f85b5bf2c0dbf3f88ae9cf231d4)

PR #142465, opened 2026-09-08
fix(ui): prevent model picker lag with large catalogs

steipete:
<details>

This is a same-repository branch, so maintainers retain their normal repository access. The fork-only collaboration setting does not apply.

</details>

## What Problem This Solves

Fixes an issue where users with large model catalogs would experience slow model selection in Chat and sluggish typing in New Session.

## Why This Change Was Made

Prepare the catalog lookup once per render and reuse the original rows for metadata, provider display, and labels. Keeping source order and duplicate rows preserves first-match, canonical-label, raw-ID ambiguity, and default-selection behavior while deleting repeated scans and an unused argument.

## User Impact

Large published catalogs are cheaper to render, with the same model values, aliases, labels, search results, and selections. No configuration, authentication, routing, or API changes are needed.

## Evidence

Actual-render regression tests failed before the repair and pass after it. At 400 models, model-ID reads fell from **407,004 to 4,404** (98.9% fewer); the protected lookup work now grows linearly.

In controlled Chromium runs with a synthetic 1,000-model Gateway catalog, observed main-thread time fell from **1,142 ms to 110 ms for Chat model selection**, and **2,251 ms to 199 ms for New Session typing**. These are observed timing samples, not wall-clock assertions. There is no Chat-per-keystroke performance claim.

All **496 focused unit/sibling tests**, **18 browser tests**, and the full scoped changed-file gate passed, including test/UI type graphs, lint, styles, and export checks. Managed **P0–P2 review is scoped-clean**. Production code is **16 lines smaller**, with 133 added test lines.

The inspected synthetic captures below demonstrate unchanged search results and metadata. They were recaptured with the canonical invitation dismissal; the timing claims above remain bound to the original benchmark runs, not the recapture.

| Chat before | Chat after |
| --- | --- |
| Chat finds Model 0999 before the lookup repair | Chat finds the same Model 0999 after the lookup repair |

| New Session before | New Session after |
| --- | --- |
| New Session finds Model 0999 before the lookup repair | New Session finds the same Model 0999 after the lookup repair |

### DIFF
--- a/ui/src/e2e/chat-model-controls.e2e.test.ts
+++ b/ui/src/e2e/chat-model-controls.e2e.test.ts
@@ -1,3 +1,4 @@
+import { writeFile } from "node:fs/promises";
 import { expect, it } from "vitest";
 import { createControlUiE2eArtifactDir } from "../test-helpers/control-ui-e2e-artifacts.ts";
 import { installMockGateway } from "../test-helpers/control-ui-e2e.ts";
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [87a9c565a90ce4dd7f3af725d5b087aef838be40](https://github.com/openclaw/openclaw/pull/142465/commits/87a9c565a90ce4dd7f3af725d5b087aef838be40)
- Landed commit: [b9f1e86810cc093ab97fc2b04c3fc269ecac95e2](https://github.com/openclaw/openclaw/commit/b9f1e86810cc093ab97fc2b04c3fc269ecac95e2)

PR #142466, opened 2026-09-08
refactor: remove unreachable process poll test branch

steipete:
Related: #139428

## What Problem This Solves

The finished-process output table retained an omission-note branch after all three rows had changed to expect complete unread output. Every row set the same false flag, so that branch could never run.

## Why This Change Was Made

Remove the unused row flags and unreachable branch, executing the existing six assertions directly for each of the three output lengths. Real omission and retention-cap coverage remains in the neighboring cases.

## User Impact

No production change or lost test case. This removes nine net test lines: six lines of dead code/scaffolding and three lines from the resulting callback formatting.

## Evidence

The full owning file and process-registry sibling passed all 66 tests before and after, with identical names and order. Selecting the original table's false branch produces the same complete AST as the revised file, preserving the three row names, lengths, generated output, call arguments/order and six executed assertions per row. No helpers or deadlines changed, and no runtime improvement is claimed.

Required changed-file checks and independent P2 review passed.

### DIFF
--- a/src/agents/bash-tools.process.poll-timeout.test.ts
+++ b/src/agents/bash-tools.process.poll-timeout.test.ts
@@ -681,40 +681,31 @@ test("process poll resets retryInMs when output appears and clears on completion
 });
 
 test.each([
-  { name: "below the retained tail", outputLength: 1_999, expectsOmissionNote: false },
-  { name: "at the retained tail", outputLength: 2_000, expectsOmissionNote: false },
-  { name: "above the retained tail", outputLength: 2_001, expectsOmissionNote: false },
-])(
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0cafadcc0a951e2e4b1f6823e41af3338ee17e40](https://github.com/openclaw/openclaw/pull/142466/commits/0cafadcc0a951e2e4b1f6823e41af3338ee17e40)
- Landed commit: [e5f860243b9ea3d1583112f9788272ae2d34725a](https://github.com/openclaw/openclaw/commit/e5f860243b9ea3d1583112f9788272ae2d34725a)

PR #142470, opened 2026-09-08
fix: queued follow-ups drop the implicit reply target

### LINKED ISSUE #142268: Queued follow-up delivery drops the implicit reply target that the normal path injects
## Summary

Named follow-up from #142008 ("preserve multiline text-directive commands at ingress"). While proving that fix on Telegram's Test Server, an exploratory overlapping-turn control exposed a separate, preexisting defect: **when a second user turn is queued while the first is still running, the queued turn's delivered messages lose their reply target** (`reply_to_message_id` is absent), even though the identical turn on the normal (non-queued) path replies to the triggering message correctly.

A controlled early second turn reproduces the same failure on pinned main, so this is independent of #142008's change.

## Evidence chain

The reply target exists end-to-end on the normal path but is never injected on the queued-delivery path:

1. Normal path: core `buildReplyPayloads` threads `currentMessageId = sessionCtx.MessageSidFull ?? sessionCtx.MessageSid` (`src/auto-reply/reply/agent-runner-result-payloads.ts:241`) into `resolveReplyThreadingPayloads` (`src/auto-reply/reply/agent-runner-payloads.ts:236-245`), whose implicit-injection branch (`src/auto-reply/reply/reply-payloads-base.ts:47-56`) sets `payload.replyToId`. Telegram's outbound adapter then maps it to `reply_parameters` (`extensions/telegram/src/outbound-adapter.ts:119`, `extensions/telegram/src/reply-parameters.ts:52-95`).
2. Queued path: the followup run **already carries the right data** — `FollowupRun.messageId` and `originatingReplyToId` are set at admission (`src/auto-reply/reply/get-reply-run-execute.ts:356-401`) and survive into the execution context (`src/auto-reply/reply/followup-turn-execution.ts:33-69`).
3. Breakpoint: `resolveFollowupDeliveryPayloads` (`src/auto-reply/reply/followup-delivery-payloads.ts:82`) resolves tags with a bare `applyReplyTagsToPayload(payload)` — no `currentMessageId`, no implicit injection — so `payload.replyToId` stays `undefined`. The queued run's `messageId`/`originatingReplyToId` is never consumed anywhere in this delivery data flow.
4. Downstream there is no recovery: `sendFollowupPayloads` routes through `routeReply` (`src/auto-reply/reply/followup-delivery.ts:410-430`), whose params carry no reply-target field (`src/auto-reply/reply/route-reply.ts:78-125`); with Telegram's `resolveReplyTransport` unimplemented, `resolvedReplyToId` falls back to `payload.replyToId ?? undefined` (`src/auto-reply/reply/route-reply.ts:306-321`) and the batch goes out with `replyToId: null` (`route-reply.ts:369`).

Existing tests only assert pass-through of an **explicit** `payload.replyToId` (e.g. `followup-delivery.channel.test.ts`, `route-reply.test.ts:355-376`); nothing covers implicit injection on the queued path, which is why this stayed invisible.

## Reproduction

Telegram DM against a Gateway running pinned main: send turn 1 (a slow task), then immediately send turn 2 while turn 1 is still streaming. Turn 2 is admitted as a queued followup; when its answer is delivered, the Telegram message has no `reply_to_message_id`, while the equivalent non-queued turn replies to its triggering message. Reproduced with a deterministic HTTP provider on Telegram's Test Server during #142008 validation (main `4f84c11f`).

## Behavior-preserving repair

This is a behavior-preserving parity fix: bring the queued path up to the normal path's existing implicit-injection semantics.

- Owner: `src/auto-reply/reply/followup-delivery-payloads.ts` — accept an optional `currentMessageId` and route through the same threading resolution the normal path uses, instead of the bare tag application.
- Call sites in `followup-delivery.ts` resolve the value with the existing admission-side logic (`resolveFollowupCurrentMessageId`, `followup-turn-admission.ts:99-104`), keeping the restart-sentinel special case (`originatingReplyToId`) consistent with `buildFollowupTemplateContext`.
- No new `routeReply` parameter: payload-level injection keeps the existing `replyToMode` filter (`createReplyToModeFilterForChannel`) authoritative — `"off"` still strips implicit targets, `"first"`/single-use semantics unchanged. Only deployments with reply mode enabled see a change, and the change is "restore the reply target", not new behavior.
- Telegram extension needs zero changes; every routable channel benefits.

Regression tests: queued turn with `messageId`/`originatingReplyToId` delivers payloads with the injected `replyToId`; covers `replyToMode` `"all"`/`"first"`/`"off"` and the restart-sentinel case; deletion of the injection must turn the new tests red.

Happy to open the PR.

Related: #142008 (where the defect was observed), #112018 (adjacent rapid-follow-up symptom on Slack, closed).

zackmsa777-a11y:
Closes #142268

<details>

</details>

## What Problem This Solves

Queued follow-up answers on resolver-less channels (Telegram after restart, no live dispatcher) dropped the implicit reply target, so `reply_to_message_id` was missing.

This PR restores that injection for same-channel external/legacy turns. It does **not** claim ordinary overlapping-turn parity through a live Telegram Test Server run.

## Why This Change Was Made

Queued follow-up delivery already stores `messageId` / `originatingReplyToId` at admission. Payload resolution only applied reply tags and never ran the same implicit-injection helper the live path uses.

Queued payloads now go through `resolveReplyThreadingPayloads`. Implicit `currentMessageId` is injected only for resolver-less channels, with the same origin/provenance exclusions as `routeReply` (same-channel external or legacy user turns). Channels that own `resolveReplyTransport` keep correlation at `routeReply`, so empty/`replyToId` opt-outs survive. Restart-sentinel still keeps `originatingReplyToId`. Explicit `replyToId` still wins. `replyToMode` `off` / `first` / `all` stay authoritative.

## User Impact

On resolver-less transports, queued follow-up answers reply to the triggering same-channel message. Cross-channel and internal/inter-session IDs are not forwarded. Channels with a transport resolver still decide correlation at final routing.

## Evidence

Rebased onto current `main` as a single commit. Head SHA: `c6f38a93c8f7f9266efe7d3b39dbfd1aa799494b`. Marker: `PROOF_OK_c6f38a93`.

Focused Vitest after this revision (`test/vitest/vitest.auto-reply-reply.config.ts`): `followup-delivery.test.ts` + `followup-delivery.channel.test.ts` + `followup-delivery.implicit-reply.test.ts` + `followup-delivery-payloads.test.ts` + `followup-turn-admission.test.ts` — **137 passed**.

Coverage:

- same-channel queued `messageId` still injects `replyToId` on resolver-less channels
- Discord-origin / Slack-destination IDs are withheld at decision time
- inter-session child announcements stay unthreaded
- restart-sentinel keeps `originatingReplyToId`
- explicit `replyToId` survives excluded implicit IDs
- empty `replyToId` opt-out survives decision and delivery through the real channel resolver
- decision-to-delivery matrix matches those exclusions through the real channel resolver
- core tests use a fixture channel; they no longer import `extensions/qa-channel`

Sabotage: switching decision injection back to unguarded `resolveFollowupCurrentMessageId` turns the foreign-channel and inter-session tests red.

Runtime observation on `ba68eaeb` used the real Gateway + Telegram transport after restart and recorded the queued answer referencing message 11. That remains the dispatcher-less recovery proof. Telegram Test Server E2E for ordinary overlap was **not** rerun (no Convex QA broker credentials on this fork).

## Maintainer refresh — September 14, 2026

Refreshed onto `82a9bef596011a6145d7f639d9773ff5e7267f7d`, preserving the original patch and contributor authorship. Current candidate: `6d592e236f8d40d33c911391ad795ed59062b439`.

Fresh review exposed an empty-target regression on channels without a transport resolver. The shared threading owner now distinguishes an absent reply target from `replyToId: ""`; the latter remains an explicit opt-out. Inline explicit-reply directives still override that implicit-threading opt-out. This is an owner-level condition change, not a follow-up-specific workaround.

Verification:
- Added regressions failed on the pre-repair candidate: **3 failed, 46 passed**.
- Repaired candidate: **183 passed across 7 suites**, covering queued delivery, channel boundaries, admission/provenance, resolver delegation, shared threading, empty/false opt-outs, inline override, first-reply slot accounting, and deduplication.
- Command: `node scripts/run-vitest.mjs src/auto-reply/reply/followup-delivery.test.ts src/auto-reply/reply/followup-delivery.channel.test.ts src/auto-reply/reply/followup-delivery.implicit-reply.test.ts src/auto-reply/reply/followup-delivery-payloads.test.ts src/auto-reply/reply/followup-turn-admission.test.ts src/auto-reply/reply/reply-plumbing.test.ts src/auto-reply/reply/reply-threading.test.ts`
- `git diff --check` passed. The repair adds one net production line (an explanatory comment) and 61 test lines. Full changed-surface static checks and hosted CI are tracked separately; this section does not claim they have completed.
- This is deterministic owner/channel-boundary proof, not new authenticated live-channel proof.

Review adjudication: the fresh autoreview returned four reports, not a scoped-clean result. The index-only empty-target finding is resolved in the committed candidate. Two resolver-owned issues were verified as unchanged baseline behavior and recorded as follow-ups: **resolver-owned empty/false opt-out consistency** and **separating explicit inline current-message resolution from implicit correlation policy**. The reported QA dedupe mismatch was not established: that resolver returns a native message reference without `threadId: null`, while the dedupe owner treats `replyToId` as a conversation thread only for the explicit-null transport contract. No new unresolved P0–P2 regression was independently confirmed in this candidate.


zackmsa777-a11y:
Addressed the ClawSweeper P2: implicit queued reply IDs now use the same origin/provenance exclusions as `routeReply`.

- same-channel external/legacy user turns still get `replyToId`
- Discord→Slack and inter-session announcement IDs are withheld
- restart-sentinel still uses `originatingReplyToId`
- explicit `replyToId` still wins

Decision-to-delivery coverage is in `followup-delivery.test.ts` + `followup-delivery.channel.test.ts` (100 passed). Sabotage of the unguarded inject turns the new tests red.

Telegram Test Server E2E is still blocked here — no Convex QA lease on this fork. Maintainer/ClawSweeper lease needed for that proof.

zackmsa777-a11y:
@clawsweeper re-review

wangmiao0668000666:
Mechanism analysis + a controlled Telegram loopback proof for this fix (no QA lease needed — stub Bot API + real gateway from `dist` + scripted mock provider).

## Why an overlapping-turn proof cannot validate this fix (measured)

While verifying my own report in #142268 I traced both queued follow-up delivery lanes on current main:

- **Lane A — source-turn dispatcher.** When the follow-up drains while the source turn's session callbacks are still registered (the common overlap case), the answer rides the *source* turn's reply dispatcher into the channel turn plan. For Telegram that is `deliverWithProviderMessageSending` bound to the *first* message's dispatch context, so the queued answer threads to the earlier inbound message, not its own. Verified end-to-end: on post-#141905 main-line code the overlap scenario yields `reply_parameters.message_id = 10` (the first message), both with and without this PR — the decision lane never engages there because the source dispatcher delivers first.
- **Lane B — decision lane.** When no source dispatcher exists (restart-recovery drains), delivery goes through `resolveFollowupDeliveryDecision` → `routeReply` → channel durable send. Telegram has no `threading.resolveReplyTransport`, so the `currentMessageId` handoff from #141905 is dropped. **This PR's injection engages precisely here.**

## Controlled restart-recovery proof (PR base vs PR head)

Harness: stub Bot API behind `channels.telegram.apiRoot`, real gateway from the checkout's `dist`, `scripts/e2e/mock-openai-server.mjs` with scripted responses. Two DM updates (msg id 10, then 11) with turn 1 held so msg 11 queues; SIGKILL the gateway; restart with the same state dir; observe `sendMessage` on the stub. All timestamps UTC 2026-09-09.

**Kill while the follow-up is still queued (turn 1 never settles):**

| build | post-restart deliveries |
|---|---|
| `ba68eaeb` (PR head) | turn-1 recovered answer: no reply target; **queued follow-up answer: `reply_parameters.message_id = 11`** (3 provider requests: turn 1 pre-kill, turn-1 recovery, follow-up re-execution) |
| `083b498` (PR base) | the queued update never delivered within a 10-minute window in my harness — the spool replay kept failing with `restart recovery claim changed before agent adoption` (36 retries, then gave up). I see this on base only; likely recovery-machinery timing, flagging it separately rather than claiming it as the PR's delta |

The head-side `message_id = 11` is the fix working end-to-end on the real Telegram path: the persisted queue drains dispatcher-less after restart, the decision lane runs, and the payload now carries the queued turn's implicit target.

**Kill during the follow-up's first execution (adopted):**

| build | recovered answer |
|---|---|
| `083b498` / `ba68eaeb` | both standalone (no reply target) |

Worth naming: when the follow-up already started executing at kill time, delivery comes from the interrupted-run recovery (`main-session-restart-recovery`), which does not thread the reply target on either build. That lane is a different owner from this PR's decision-lane injection — a possible follow-up, not a gap in this change. (The head run's standalone turn-1 answer above is the same lane.)

## On the "core injection vs channel-owned reply policy" concern

On the decision lane there is no live channel turn plan or dispatcher, so the only alternative carrier would be a per-channel `resolveReplyTransport` like qa-channel's. Resolving the implicit target at payload level covers every resolver-less channel at once, keeps `replyToMode` authoritative, and mirrors what the normal non-queued path already does in core (`buildReplyPayloads`). The origin/provenance guard in the latest revision matches the exclusions `sendFollowupPayloads` already applies for its `routeReply` handoff, which addresses the cross-channel leak risk.

Happy to share the harness scripts if they help clear the telegram-e2e proof gate.

zackmsa777-a11y:
Addressed the remaining ClawSweeper P2 and the hosted CI failures.

- Channel-owned implicit replies: payload injection now consults `resolveReplyTransport` first. QA first-mode stays unthreaded; Telegram and other resolver-less channels still inject.
- CI `max-lines` (1308) — split implicit-reply and route tests out of `followup-delivery.test.ts`.
- CI knip — unexported unused `resolveFollowupCurrentMessageId`.
- Discord `native-command.reset` 120s timeout in bundle-12 is unrelated to this change (same flake on #142497).

Local: follow-up suites 103 passed.

zackmsa777-a11y:
@clawsweeper re-review

zackmsa777-a11y:
P1/P2 addressed on `437760c2`.

- Decision no longer calls `resolveReplyTransport`. Resolver-less channels still inject at decision; channels that own a transport resolver keep correlation on `routeReply`, so `{ replyToId: "" }` survives.
- Core tests use a fixture plugin. The `extensions/qa-channel` import is gone.

Focused Vitest: 105 passed (`followup-delivery.test.ts`, `.channel.test.ts`, `.implicit-reply.test.ts`, `.route.test.ts`).

PR body now matches the dispatcher-less recovery scope plus the empty opt-out contract.

@clawsweeper re-review

zackmsa777-a11y:
Rebased onto current `main` as a single commit to clear merge conflicts.

Head: `5be56898a42901765ccdf22dcb0ac8635c6eac8e`

Behavior is unchanged: implicit queued reply IDs still use origin/provenance exclusions, resolver-less channels inject at decision, channels with `resolveReplyTransport` decide at `routeReply`. Focused auto-reply suite: 137 passed.

@clawsweeper re-review

zackmsa777-a11y:
Assertion-safety ratchet: dropped the grandfathered `followup-delivery-payloads.ts` entry now that the remaining casts have `// SAFETY` comments.

Head: `c6f38a93c8f7f9266efe7d3b39dbfd1aa799494b`

@clawsweeper re-review

jalehman:
Refreshed candidate: `6d592e236f8d40d33c911391ad795ed59062b439`.

The PR body now records the empty-target owner repair, failing/passing regression evidence (183 passing tests), and full review adjudication with two named baseline follow-ups. Hosted CI is running; this is not a claim that all gates have passed.

@clawsweeper re-review

PR #141916, opened 2026-09-08
fix(agents): continue settled empty post-tool turns

RomneyDa:
## Summary

- backport the settled post-tool continuation from #110565 to the 2026.6.35 candidate
- continue exactly once after a side-effecting tool batch is proven complete but the assistant emits no visible answer
- remove built-in and client-provided tools from that continuation so completed effects cannot repeat
- keep replay-safe reads on the existing empty-response path and fail closed for active, failed, asynchronous, delivered, spawned, or approval-bearing work

## Root cause

The package Telegram acceptance scenario `telegram-empty-response-after-write-recovery` completed a write, then received an empty terminal assistant turn. The candidate treated every potentially side-effecting attempt as unrecoverable, so it surfaced the generic incomplete-turn error instead of asking the model for the final visible answer from the already-settled transcript.

Failure: https://github.com/openclaw/openclaw/actions/runs/34189437057/job/101945150686

Canonical main fix: #110565 / `8636bb6981844e4674ee2cdbc0d8d32aa2a8b816`. This backport adapts only that owner invariant to the candidate's pre-split runner.

## Verification

- focused Vitest: OpenClaw settled write continues once with `disableTools: true` and no client tools; native harness does not retry
- exact-head AutoReview: correct, no actionable findings (0.94)
- type-aware OXLint on the three changed files
- Oxfmt and `git diff --check`
- pre-fix release scenario: timeout with two outbound messages and no success marker; gateway logged `payloads=0, tools=1, replaySafe=no`

## Scope

- application production: +71 net lines
- tests: +94 net lines
- no config, protocol, persistence, package, or release-workflow changes

PR #142471, opened 2026-09-08
fix(agents): preserve settled tool continuation boundary

RomneyDa:
## Release blocker

The packaged Telegram recovery scenario in [Full Release Validation 34261943173](https://github.com/openclaw/openclaw/actions/runs/34261943173) reached the settled-tool continuation from #141916, then produced no matching final marker. Its artifact records four model fetches before the only outbound reply and a retry of the same Telegram inbound message.

## Root cause

#141916 disabled tools for its one recovery pass but left the already-persisted inbound message as the next attempt's prompt. That replays the source user turn instead of continuing from the settled assistant/tool transcript. The runner already owns the canonical boundary for this case: internal continuations use a new model prompt and suppress re-persisting the current inbound message.

## Fix

When the current channel message is known to be persisted, use the settled-tool instruction as the next model prompt and mark the next attempt as persistence-suppressed. The recovery still has no tools. If persistence is not proven, retain #141916's existing prompt-addition path rather than dropping user input.

## Proof

- Red: the focused settled-write test observed the old second attempt as `prompt: "hello\\n\\n<continuation>"` with `suppressNextUserMessagePersistence: false`.
- Green: `node scripts/run-vitest.mjs src/agents/embedded-agent-runner/run.incomplete-turn.test.ts` — 113 passed.
- `pnpm exec oxfmt --check src/agents/embedded-agent-runner/run.ts src/agents/embedded-agent-runner/run.incomplete-turn.test.ts` — passed.
- `node scripts/check-changed.mjs --dry-run -- src/agents/embedded-agent-runner/run.ts src/agents/embedded-agent-runner/run.incomplete-turn.test.ts` — `core, coreTests`.
- `git diff --check` — passed.

Production delta: +9 lines; test delta: +13 lines. The production addition is the current-message identity gate plus an ownership comment; it avoids reintroducing a second user leaf into a persisted conversation.

### DIFF
--- a/src/agents/embedded-agent-runner/run.incomplete-turn.test.ts
+++ b/src/agents/embedded-agent-runner/run.incomplete-turn.test.ts
@@ -77,6 +77,7 @@ describe("runEmbeddedAgent incomplete-turn safety", () => {
     prompt?: string;
     disableTools?: boolean;
     clientTools?: unknown[];
+    suppressNextUserMessagePersistence?: boolean;
   } {
     // Continuation prompt assertions read the exact prompt passed to the runner
     // attempt rather than derived result metadata.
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper review

PR #136439, opened 2026-09-02
fix(ui): retain speaker names in conversation exports

steipete:
Closes #136437

## What Problem This Solves

Exporting a shared conversation replaced each person's name with “You” and a named assistant with the default assistant name. Both `/export` downloads and Copy → Conversation as Markdown lost the attribution shown in the transcript.

## Why This Change Was Made

Use the existing transcript sender and role normalizer in the shared Markdown builder. Chat, Usage and Logs also share their existing text-download lifecycle, removing three duplicate implementations while preserving each caller's MIME type, filename, contents and URL cleanup.

## User Impact

Downloaded and copied conversations retain explicit speaker names and durable sender metadata. Unnamed speakers retain the existing You/assistant/Tool labels. Content, timestamps, silent-reply filtering, Usage snapshot fencing, and log filtering remain unchanged.

## Evidence

- Actual Chromium download and clipboard flows on untouched main `65e5b2e19c9`: Alex, Sam and Review assistant become You, You and OpenClaw. The same two browser regressions now preserve all three names.
- 64 focused unit tests across six files pass. Five additional production-build browser cases cover unnamed/empty exports, attachment state, and actual isolated Gateway Logs download, source recovery and reconnect.
- UI typecheck, typed lint, formatting and whitespace checks pass. Independent Codex review through P2 is clean; the committed sources match the reviewed hashes.
- Production: **30 added / 40 removed (net −10)**, including all eight lines of the new shared helper. Tests/support: **+159 / −18 = +141**. Docs: **1 changed line**.
- Export regression proof uses real Chromium with synthetic Gateway replies; the Logs sibling proof uses a real task-owned ephemeral Gateway. No external service or user session was changed.

Before, distinct participants collapse to generic roles:

Before: participant names lost in Markdown

After, the same exported conversation preserves its speakers:

After: participant names retained in Markdown

## Recorded behavior output

The following is the actual synthetic Markdown downloaded by Chromium after invoking `/export`, not an expected-value fixture. Copy → Conversation as Markdown produced identical bytes. The complete before/after images above were also personally inspected; the reviewer's attachment-DNS failure is an access limitation, not a missing capture.

Before, both actions produced headings `You`, `You`, `OpenClaw`. After, both produced:

```markdown
# Chat with OpenClaw

## Alex

I will write the release notes.

## Sam

I will verify the build.

## Review assistant

Alex owns the notes; Sam owns build verification.
```

[...]

PR #142491, opened 2026-09-08
perf(ui): reuse speaker metadata during chat export

### LINKED ISSUE #142490: Avoid discarded display normalization during chat export
## Summary

Resolve chat-export speaker metadata without constructing the full display-normalized message for every row.

## Problem to Solve

Both Markdown download and clipboard export use the shared chat builder. It normalizes each complete message to obtain its role and speaker label, then extracts the exported body separately. Normalized content arrays, media and Canvas processing are discarded by export, including when the existing text cache is warm.

## Proposed Solution

Reuse canonical role classification and a shared sender-label resolver. Keep the existing sender policy in one owner, and let full normalization pass its already prepared sender, including known absence. Export should retain its existing text projection, visibility filtering, timestamps, exact Markdown, filename and MIME behavior.

## Impact and Evidence

The recorded comparison covers six reference cohorts with 26 messages and preserves Markdown/download bytes, complete normalized values and property order, without mutating inputs. Validation records 127 focused tests and the full selected changed gate passing. The unchanged browser download/clipboard flows pass before and after.

Synthetic 128-message transcript probes recorded about 28–47% less sampled allocation across text, mixed-block and imported-text cohorts with cold and warm text caches. Sampling includes collected objects; it does not measure retained heap or RSS. Revisions ran baseline then candidate, with three trials each, so timing is supporting microbenchmark evidence rather than a browser-latency claim.

Related: #136439 restored exported speaker attribution. Preserve that behavior; this is an ownership/allocation improvement, not a new attribution fix.

steipete:
Closes #142490

<details>

</details>

Related: #136439

## What Problem This Solves

Chat Markdown exports normalize each complete message just to obtain its role and speaker label, then extract the exported body separately. The resulting content arrays and media/Canvas preparation are discarded, including when the existing text cache is warm.

## Why This Change Was Made

This is a maintainer-requested performance cleanup. Use canonical role classification and a scalar sender-label resolver in the shared export builder. The existing metadata-to-sender policy moves into one private helper; full normalization reuses its already prepared sender, including explicit `null`. Export resolves a label only for user and assistant rows, retaining fixed Tool headings.

Imported-body projection remains with `extractTextCached`. Visibility, raw timestamp handling, identity/avatar provenance, full normalized output shape, Markdown bytes, download MIME/filename and clipboard behavior remain unchanged. Production grows by 20 lines across two owners, with six cases added to the existing export suite. No cache, configuration, persistence or public API is introduced.

## User Impact

Downloaded and copied conversations retain the speaker attribution established in #136439 and the same content. The export path avoids constructing display content that it does not use.

## Evidence

Validation:

- Six reference cohorts containing 26 messages match exact Markdown and download Blob bytes, MIME/filename, complete normalized values and property order. Inputs remain unchanged. Cases include canonical tool envelopes, explicit and durable sender metadata, imported content shapes, timestamps, filtering, media and Canvas.
- **127 focused tests pass**, including six additional export cases for tool classification and imported attribution/body bytes. Existing sender identity and provenance tests are unchanged. The full selected changed gate and independent review pass.
- The existing full synthetic Control UI export E2E passes both download and clipboard flows before and after. All four Markdown files are byte-identical; visible speaker names are retained and neither action sends `chat.send`. The application runs with a mocked Gateway and isolated Chromium contexts.

Probes exporting 128-message synthetic transcripts six times per trial recorded **27.66–47.02% less cumulative sampled allocation** across text, mixed-block and imported-text cohorts with cold and warm caches. V8 sampling used an 8 KiB interval and includes collected objects. Input creation, cache warming and output hashing/serialization were excluded. Baseline ran before candidate, with three trials per arm; timing is supporting microbenchmark evidence. No retained-heap, RSS, whole-Gateway or full-browser latency improvement is claimed.

Before: synthetic speaker attribution after Markdown download

After: preserved speaker attribution after Markdown download

### DIFF
--- a/ui/src/lib/chat/message-normalizer.ts
+++ b/ui/src/lib/chat/message-normalizer.ts
@@ -28,7 +28,7 @@ import { getMediaFileExtension } from "../media-file-extension.ts";
 import type { NormalizedMessage, MessageContentItem } from "./chat-types.ts";
 import { projectImportedMessageForDisplay } from "./imported-message-display.ts";
 import { normalizeAttachmentContentBlock } from "./message-normalizer-attachments.ts";
-import { formatSenderLabel, normalizeSenderIdentity } from "./sender-label.ts";
+import { formatSenderLabel, normalizeSenderIdentity, type SenderIdentity } from "./sender-label.ts";
 
 // Keep legacy labels readable without treating their UUID suffix as profile evidence.
 const OPAQUE_ID_LABEL_SUFFIX_RE =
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [30e6d87350c333b22756f00ed92a7bb563a5f259](https://github.com/openclaw/openclaw/pull/142491/commits/30e6d87350c333b22756f00ed92a7bb563a5f259)
- Landed commit: [dd5594c19748f2a43f511ac16388973078282e97](https://github.com/openclaw/openclaw/commit/dd5594c19748f2a43f511ac16388973078282e97)

PR #142496, opened 2026-09-08
fix(msteams): decode HTML-only message text correctly

steipete:
## What Problem This Solves

Fixes an issue where users sending HTML-only Teams messages would pass encoded text such as `&lt;`, `&#39;`, and `&copy;` to the agent instead of the intended characters. For example, `Use x &lt; 5 &copy;` arrived as literal entity text.

## Why This Change Was Made

Reuse the Teams plugin's existing quote-text HTML converter for inbound attachment text, replacing the partial entity decoder and duplicate whitespace/tag processing. Native mentions and links are prepared before single-pass decoding; raw activity text still takes precedence, and Adaptive Card value fallback keeps its existing behavior. Production code decreases by 3 lines.

## User Impact

HTML-only messages reach the agent with the intended punctuation, symbols, and Unicode characters. Literally typed entity text and encoded `<at>...</at>` text stay intact, while native bot addressing is removed once.

## Evidence

- Original-code regression: three inbound-body cases and the real Teams handler's `BodyForAgent` assertion failed for encoded entities/literal tags; six preservation controls passed.
- **68 focused tests passed**: inbound facts, the handler/debounce lifecycle, quote extraction, Graph text projection, and shared HTML entity decoding. Coverage retains raw-text precedence, card-value fallback, links, whitespace, double-encoded entities, and literal mention tags.
- The handler proof uses actual preparation, the actual attachment-triggered immediate debounce flush, inbound body processing, and the shared context builder. Provider/reply and persistence boundaries use the existing synthetic test runtime. This is not a live Teams-client rendering test and makes no external Microsoft API claim.
- Scoped `check-changed` passed, including extension production/test typechecks, lint, boundary checks and selected guards. Independent P2 review found no actionable P0–P2 issues. Node 24.19.0, pnpm 12.3.4; unchanged dependency lockfile.
- The reviewed owner/caller/decoder source remains unchanged on main at `b9f1e86810cc093ab97fc2b04c3fc269ecac95e2`; no rebase was needed.

Focused command:

```sh
node scripts/run-vitest.mjs extensions/msteams/src/monitor-handler/inbound-facts.test.ts extensions/msteams/src/monitor-handler/message-handler.ingress-lifecycle.test.ts extensions/msteams/src/inbound.test.ts extensions/msteams/src/graph-thread.test.ts src/shared/html-entities.test.ts
```

### DIFF
--- a/extensions/msteams/src/inbound.ts
+++ b/extensions/msteams/src/inbound.ts
@@ -15,7 +15,7 @@ type MSTeamsQuoteInfo = {
 /**
  * Strip HTML tags, preserving text content.
  */
-function htmlToPlainText(html: string): string {
+export function htmlToPlainText(html: string): string {
   return decodeHtmlEntities(html.replace(/<[^>]*>/g, " "))
     .replaceAll("\u00a0", " ")
     .replace(/\s+/g, " ")
--- a/extensions/msteams/src/monitor-handler/inbound-facts.test.ts
+++ b/extensions/msteams/src/monitor-handler/inbound-facts.test.ts
@@ -11,7 +11,7 @@ describe("msteams inbound facts", () => {
     const textEntry = await prepareMSTeamsDebounceEntry({
       context: context({
         type: "message",
-        text: "  hello  ",
+        text: "  <at>Bot</at> hello &lt;tag&gt;  ",
         attachments: [{ contentType: "text/html", content: "<p>ignored</p>" }],
         value: { action: "ignored" },
       }),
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4f2d4b3f8f7087f1fa656d6a6b8c3f84653b68b3](https://github.com/openclaw/openclaw/pull/142496/commits/4f2d4b3f8f7087f1fa656d6a6b8c3f84653b68b3)
- Landed commit: [63943fffb46c5fc1714d46f7aa5ef53170b492cd](https://github.com/openclaw/openclaw/commit/63943fffb46c5fc1714d46f7aa5ef53170b492cd)

PR #142500, opened 2026-09-08
fix(plugins): reuse prewarmed catalog presentation

steipete:
<details>

</details>

## What Problem This Solves

Fixes an issue where repeated plugin list and inspection requests rebuilt stable catalog presentation metadata even after startup prewarming. The hosted response was cached, but each request repeated bundled matching and package identity preparation.

## Why This Change Was Made

Cache the prepared catalog in the existing plugin-cache generation. Keep the existing identity-checked rejection eviction and lifecycle refresh behavior, and remove unused private overlay parameters and the redundant asynchronous wrapper around the loader. The shared result type lives in the existing catalog types module, keeping cache definitions independent of the runtime catalog implementation.

## User Impact

Warm plugin catalog requests reuse their prepared metadata. Current configuration and public response objects remain request-specific, so refreshes still take effect and callers can modify returned lists without changing later results.

## Evidence

The regression through the real plugin management service failed before the fix: both cold and warm requests performed 98 package preparations while the hosted feed loaded only once. After the baseline refresh, the regression and all 104 focused lifecycle, featured, inspection, list, and refresh tests pass. The extra case comes from the upstream configuration-readonly behavior, which is preserved. The test also modifies a returned public list and verifies the next response remains intact.

The original corrected candidate passed the owning plugins-platform type graph, typed lint, formatting, both import-cycle graphs, and the full architecture gate. The refresh preserves the accepted patch identity and direct type-leaf ownership; all 104 owner tests and the full pinned-base changed checks pass, including core/test types, typed lint, source guards, and the runtime import-cycle check. The accepted independent P2 review remains applicable to the unchanged catalog delta. The tests use synthetic hosted-feed and installed-metadata inputs; no live service or operator state was used.

Initial CI reported a static type-import cycle and an unexpected config-doc check selection. Moving the shared result type into the existing catalog types module resolved both: the full architecture gate and all 31 config-doc selection cases pass locally.

A later CI run reported a planner-growth timeout and subagent announcement failures. The measured planner repair landed separately in #142630, and this branch is refreshed onto main containing that repair. The announcement failure remains unreproduced: a Linux replay of its exact merged-source shard passed 1,306 tests with one skipped, but used different host capacity and fresh cache state. This PR does not claim to fix that failure.

The next integration run passed all test jobs (106 successful jobs, 17 skipped), but lint rejected an inherited oversized Codex protocol file and source-contract setup failed while downloading the Matrix native library (`ECONNRESET`, followed by a dependency installer error). The aggregate gate therefore failed. This branch now includes the separately landed protocol-size repair #142648; the catalog patch and its source hashes are unchanged. The Matrix download failure remains separate from the catalog fix. This is a qualified baseline refresh, not a rerun of the failed integration head.

Production: 27 additions, 30 deletions, net -3. Tests: 23 additions, 5 deletions, net +18.

### DIFF
--- a/src/plugins/management-catalog.ts
+++ b/src/plugins/management-catalog.ts
@@ -148,12 +148,10 @@ export function prepareCatalogEntries(entries: readonly OfficialExternalPluginCa
  */
 function overlayBundledOfficialPluginCatalogMetadata(
   entries: readonly OfficialExternalPluginCatalogEntry[],
-  bundledEntries: readonly OfficialExternalPluginCatalogEntry[] = listOfficialExternalPluginCatalogEntries(),
-  options: { hostedFeaturedAuthoritative: boolean } = {
-    hostedFeaturedAuthoritative: false,
-  },
+  options: { hostedFeaturedAuthoritative: boolean },
 ): OfficialExternalPluginCatalogEntry[] {
-  const bundledFacts = entries.length > 0 ? bundledEntries.map(prepareCatalogEntry) : [];
+  const bundledFacts =
+    entries.length > 0 ? listOfficialExternalPluginCatalogEntries().map(prepareCatalogEntry) : [];
   return entries.map((entry) => {
     const { clawhub, npmPackage } = prepareCatalogEntry(entry);
     const matches = bundledFacts.filter(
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5105c79d57ca122c621c10c26a0f916da7009d67](https://github.com/openclaw/openclaw/pull/142500/commits/5105c79d57ca122c621c10c26a0f916da7009d67)
- Landed commit: [2542a8c2cd2f23b7e4e18434cf212b0de0ef1d96](https://github.com/openclaw/openclaw/commit/2542a8c2cd2f23b7e4e18434cf212b0de0ef1d96)

PR #142508, opened 2026-09-08
fix: load the host SDK for plugins in sibling checkouts

steipete:
Related: #135599

## What Problem This Solves

Fixes an issue where plugins stored inside another OpenClaw checkout could load that checkout's older SDK instead of the SDK belonging to the running Gateway. This can cause incompatible runtime or storage helpers to be selected even when the Gateway itself is up to date.

## Why This Change Was Made

Resolve the SDK from the executing loader's explicit module URL before considering the plugin's directory or launcher. Keep the explicit development-source override and existing no-host fallbacks, trust checks and private SDK ownership rules. The change removes seven production lines.

## User Impact

Plugins can live under a sibling checkout without silently selecting its SDK. Existing standalone and development-source loading continue to work.

## Evidence

Six source/dist cross-checkout cases failed before the fix; all 12 cases now pass, including misleading or missing launchers, development overrides and no-host controls. The full channel-entry, alias and native-resolver suites passed 118 tests with one existing skipped test. The synthetic SDK artifacts are inspected as metadata and never executed by the new regression cases.

A synthetic actor scenario in #135599 exposed the wrong-host selection. After integration, the same isolated Gateway storage probe succeeded using the running host SDK. Broader actor integration work remains in #135599; this is not a claim that the whole actor scenario passed. This repair does not change database schemas.

The first CI run exposed a fast-path fixture that implicitly selected its temporary SDK as the host. The fixture now declares that host through the existing development override, preserving its native-load and zero-transform assertions. The original production change is unchanged. Local changed-file gates passed, including the follow-up test gate, and fresh review of the complete candidate through P2 found no actionable issues.

### DIFF
--- a/src/plugin-sdk/channel-entry-contract.test.ts
+++ b/src/plugin-sdk/channel-entry-contract.test.ts
@@ -349,6 +349,8 @@ function runCompiledEsmSidecarFastPathProbe(): SpawnSyncReturns<string> {
     },
   });
   fs.writeFileSync(path.join(tempRoot, "openclaw.mjs"), "#!/usr/bin/env node\n", "utf8");
+  fs.mkdirSync(path.join(tempRoot, "src"));
+  fs.mkdirSync(path.join(tempRoot, "extensions"));
   fs.mkdirSync(path.join(tempRoot, "dist", "plugin-sdk"), { recursive: true });
   fs.writeFileSync(
     path.join(tempRoot, "dist", "plugin-sdk", "channel-outbound.js"),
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b3811c233dd93fbaf8e0d1543d428333a6a85cf8](https://github.com/openclaw/openclaw/pull/142508/commits/b3811c233dd93fbaf8e0d1543d428333a6a85cf8)
- Landed commit: [f9119d8c76ff19b89d40b3a5723529add2ccb0c6](https://github.com/openclaw/openclaw/commit/f9119d8c76ff19b89d40b3a5723529add2ccb0c6)

PR #142518, opened 2026-09-08
fix(mcp): node tool results copy large text twice

steipete:
<details>

**MUST:** Keep **Allow edits from maintainers** enabled so maintainers can resolve small issues before merging.

</details>

## What Problem This Solves

Fixes an issue where agents using node-hosted MCP tools copy large plain-text results twice into each result snapshot, increasing allocation and serialization work.

## Why This Change Was Made

Remove the auxiliary text mirror left over from the former Code Mode result format. The existing MCP projector already supplies model content and independently retains the native MCP result for Code Mode; the mirror's old caller was migrated to that native contract.

## User Impact

Text-heavy node MCP calls produce smaller result snapshots while preserving model text, resource links, error status and native Code Mode data. Structured results and non-MCP node tools retain their existing behavior.

## Evidence

The new actual node-tool boundary regression failed on both normal and error results before the deletion because snapshot size was about 1.1 MB rather than under 551,141 bytes. All model/raw/error assertions already passed. After the nine-line production deletion, all 21 node-plugin-tool tests pass, including structured results, empty errors and native namespace siblings.

On Node 24.19.0, the isolated actual baseline/edited-owner snapshot benchmark for 525,000 bytes of text shrank serialized results from 1,100,125 to 550,087 bytes. Median mapper+snapshot times across three runs were 1.409–1.426 ms before and 0.685–0.707 ms after. Separate benchmark-process peak RSS was 157,089,792 bytes before and 111,427,584 bytes after. These are isolated boundary measurements, not end-to-end network latency or production RSS claims.

Managed P2 review is scoped-clean. The complete selected changed-file gate passed, including owning production/test types, full-file lint/format and all selected guards. Production removes 9 lines; tests add 40 lines. No live MCP server or provider call was made.

### DIFF
--- a/src/agents/node-plugin-tools.test.ts
+++ b/src/agents/node-plugin-tools.test.ts
@@ -1,5 +1,6 @@
 /** Tests connected node-hosted plugin tool materialization. */
 
+import { CallToolResultSchema } from "@modelcontextprotocol/sdk/types.js";
 import { expectDefined } from "@openclaw/normalization-core";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import type { NodePluginToolDescriptor } from "../../packages/gateway-protocol/src/index.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6fd283d3ad21b238936a250ae75a2f8b00e9fea3](https://github.com/openclaw/openclaw/pull/142518/commits/6fd283d3ad21b238936a250ae75a2f8b00e9fea3)
- Landed commit: [9a209e0a18de9b37576c68468e40d23330e01aa3](https://github.com/openclaw/openclaw/commit/9a209e0a18de9b37576c68468e40d23330e01aa3)

PR #142525, opened 2026-09-08
fix(sessions): expose slow reclamation between writer scopes

steipete:
## What Problem This Solves

Resolves a problem where operators investigating slow session cleanup could see no operation-duration warning when reclamation preparation took several seconds between otherwise short SQLite writer admissions.

## Why This Change Was Made

The archive Worker dispatcher now observes the existing joined completion promise and emits one warning for reclamation operations taking at least one second. It records the typed reclamation kind, actual Node Worker identity, parent process/thread identity, elapsed lifetime, terminal outcome and exit code after Worker exit and parent admission settlement. The original result or rejection stays authoritative even if logging fails.

The timing excludes the outer archive queue, includes startup/preparation/admission/work/cleanup, and does not claim CPU time or isolate validation. Existing parent trace context is retained without logging paths, session identifiers, plan content or raw errors. No configuration, storage, authorization or concurrency behavior changes.

## User Impact

Operators can identify slow whole reclamation operations even when their individual writer sections remain below the warning threshold. This is an observability repair, not a claimed cleanup speedup.

## Evidence

- The real-Worker regression failed before the fix in all four slow cases because the terminal diagnostic was missing; the fast case already passed. All five cases pass after the fix, including successful/rejected outcomes and logging transport failures.
- All 76 focused reclamation, archive Worker, writer-scope, commit and admission-race tests passed locally.
- A real-clock fixture delayed the second admission by 1.25 seconds. Another writer progressed, no slow-writer warning appeared, and the joined terminal record retained the actual Worker identity after exit.
- Independent source review and isolated Codex autoreview through P2 found no actionable findings.
- Full local build and changed-file checks passed. The source-verified Linux build also passed on Blacksmith Testbox with Node 24.19.0 and pnpm 12.3.4.
- Real CLI/Gateway create, rename, reset, delete, repeated-delete and final-health assertions passed locally and on Linux. Linux additionally verified wrapper exit, process-group disappearance and listener closure. CLI timings include client startup and are not RPC latency measurements.

The first Gateway fixture incorrectly expected reset to replace the session ID; the corrected fixture verifies the existing stable-ID/new-lifecycle-revision contract. A later archive helper failed after the Linux runtime assertions and shutdown checks completed; evidence collection was repaired separately. Both failed proof-tool attempts are retained. These synthetic flows do not establish sustained production performance or a CPU cause.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -312,6 +312,18 @@ integrity check; resumed validation and repair can still run on the opener.
 Correlate the process ID with the log timestamp and current process; PIDs can be
 reused after exit.
 
+SQLite reclamation Workers also emit `slow SQLite reclamation Worker operation`
+at `warn` when their joined operation takes at least one second. The record is
+emitted after Worker exit and parent admission settlement. It includes the
+parent's `pid`, `threadId` and `isMainThread`, the actual Node `workerThreadId`,
+`reclamationKind`, `elapsedMs`, terminal `outcome` (`resolved` or `rejected`), and
+`exitCode`. Timing starts after admission to the archive Worker queue and includes
+startup, validation, admission waits, work, and cleanup. It does not measure CPU
+time or isolate a validation phase. Short writer sections can therefore remain
+quiet while this whole-operation warning exposes slow preparation between them.
+The record inherits an existing parent trace when available; it contains no
+database path, session identifier, plan content, or raw error.
+
 ### Slow reply preparation
 
 When a reply spends a long time preparing, inspect the normal Gateway logs:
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [37c25150383f9cc02a45d7d8a734670be49d5cb8](https://github.com/openclaw/openclaw/pull/142525/commits/37c25150383f9cc02a45d7d8a734670be49d5cb8)
- Landed commit: [f2912393bcabab831bff533d2ecdbee82ccba2b5](https://github.com/openclaw/openclaw/commit/f2912393bcabab831bff533d2ecdbee82ccba2b5)

Issue #142393, opened 2026-09-08
[Bug]: Deep-dreaming promotes low-value, zero-recall snippets into MEMORY.md and grows it past the bootstrap char cap


zachisfine:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

  The deep-dreaming promotion writes low-value snippets (code/config fragments) into MEMORY.md at recalls=0, and does so repeatedly, growing MEMORY.md past the
  agent's bootstrapMaxChars cap so its tail is silently truncated on every load.

### Steps to reproduce

  1. Run OpenClaw 2026.7.1-2 with memory-core dreaming enabled and an agent that has bootstrapMaxChars set (e.g. 9000).
  2. Let the nightly promotion run over several days.
  3. Inspect MEMORY.md: it accumulates "## Promoted From Short-Term Memory (DATE)" blocks whose entries are raw code/config snippets tagged `recalls=0`.
  4. Check total size vs the cap: MEMORY.md exceeds bootstrapMaxChars, so the injected copy is truncated (the file's tail silently drops out of the agent's
  context).

### Expected behavior

  Promotion should only durably write snippets that meet a meaningful recall/uniqueness/quality bar (the configured minRecallCount / minUniqueQueries / minScore),
  should not promote entries at recalls=0, and should not grow MEMORY.md unbounded past bootstrapMaxChars. High-signal, recalled content should be preferred over
  incidental code fragments. 

### Actual behavior

  MEMORY.md accumulated three "Promoted From Short-Term Memory" blocks (2026-09-06/07/08) whose entries are code/config scraps sourced from old memory files, each
  tagged `recalls=0`. The file reached 9728 bytes against a 9000-char bootstrap cap, so ~728 bytes at the tail were silently truncated on every load. Separately,
  genuinely relevant recent session content was never promoted, because session-corpus / memory/dreaming/* hits are excluded from the short-term recall counters
  (DREAMING_MEMORY_PATH_RE), so they can never accrue the recalls that would qualify them — while low-value files that already sit in memory/ get promoted at
  recalls=0.

### OpenClaw version

2026.7.1-2 (0790d9f)

### Operating system

macOS 26.5.2 (Darwin 25.5.0)

### Install method

npm global

### Model

ollama/qwythos-9b-16k:latest (configured via dreaming.model for narrative prose); agent session default is ollama/gemma4:12b

### Provider / routing chain

[...]

PR #119624, opened 2026-08-05
fix(memory): require query diversity for promotion

pcpilot-dev:
Related: #142393

## What Problem This Solves

Short-term promotion treated scheduler-generated daily identifiers and grounded-backfill keys as query diversity. Repeated nightly ingestion could therefore satisfy `minUniqueQueries` without any real recall query and promote operational residue into `MEMORY.md`.

## Why This Change Was Made

`minUniqueQueries` is intended to require independent recall contexts. Daily recurrence remains useful frequency and consolidation evidence, but it should not masquerade as distinct queries.

## User Impact

- Daily ingestion and grounded backfill alone no longer qualify content for automatic durable-memory promotion.
- Real recall queries establish the diversity needed for promotion.
- Existing recall-only state keeps its earned diversity; ambiguous mixed legacy state waits for fresh recall evidence.

## Summary

- record recall-path hashes separately as `userQueryHashes`
- keep daily and grounded synthetic keys out of query diversity
- enforce the same qualified diversity count during ranking and apply
- preserve unambiguous recall-only legacy hashes while failing closed for mixed legacy rows
- retain qualified recall hashes when dreaming deduplicates candidates
- document query diversity as distinct recall queries rather than query/day contexts

## Maintainer Decision

Accept the conservative mixed-legacy transition: recall-only rows retain attributable historical diversity, while mixed rows do not guess the origin of old hashes and must earn fresh recall-query evidence. This avoids laundering scheduler evidence while preserving the unambiguous upgrade path.

## Evidence

- isolated fresh-workspace runtime proof through the production SQLite store and built CLI:
  - three daily signals produced `ranked=0`
  - three distinct recall queries produced `ranked=1`, `uniqueQueries=3`, `signalCount=6`, score `0.830`
  - `openclaw memory promote --apply --json` reported `applied=1`, and the synthetic fact was present in `MEMORY.md`
- isolated upgrade runtime proof:
  - created recall-only persisted evidence, removed the new field to reproduce the legacy row shape, then reopened it through the production normalizer
  - the upgraded row retained `uniqueQueries=3`, ranked at `0.780`, and the built CLI reported `applied=1`
  - the synthetic upgraded fact was present in `MEMORY.md`
- `pnpm test:extension memory-core`: 110 files passed, 2 skipped; 1,584 tests passed, 8 skipped
- exact-head focused run over the four changed behavior suites: 4 files and 282 tests passed
- `pnpm check:changed`: passed
[...]

PR #142545, opened 2026-09-08
fix(memory): prevent dreaming promotion from exceeding bootstrap limits

RomneyDa:
Related to #142393

## What Problem This Solves

Memory Core promotion previously used a 10,000-character writer default even when a consuming agent injected less of `MEMORY.md`. Its append fallback could then commit an oversized file and mark candidates promoted.

## Behavior

- Resolve the effective writer budget from the smallest per-file bootstrap cap among agents sharing the canonical workspace.
- Preserve user-authored and ambiguous content unconditionally.
- Bound removal of prior machine-generated promotion entries with the existing `maxPriorEntryLossFraction` policy (default 25%) for both consolidation and append fallback.
- Apply the same configured loss policy in scheduled dreaming and `memory promote --apply`.
- Commit the file and promotion state together only when the complete result fits. Otherwise both remain unchanged so a later sweep can retry after capacity changes.
- Rebuild append text from the final lock-revalidated candidate set, so changed or newly untrusted candidates cannot reach durable memory through a valid sibling.
- Emit dream narrative/report content only for candidates actually applied.

This remains budget-only. Recall eligibility and query diversity were addressed separately by merged #119624.

## Safety Invariants

- No user-authored memory truncation.
- No partial file/state update.
- No unbounded loop inside a sweep.
- A single write cannot evict more than the configured prior-entry loss fraction; if that is insufficient, the write is deferred.
- Repeated scheduled sweeps may retry consolidation, but an unapplied attempt does not append another dream narrative/report.
- Consolidation remains available near capacity because a validated replacement may fit even when a plain append cannot.

## Evidence

- Exact-head focused suite: 252 passed across CLI, promotion writer, and dreaming tests.
- Mixed-batch production-writer regression changes one candidate from trusted to untrusted while consolidation is pending, then proves only its eligible sibling is applied and serialized to `MEMORY.md`.
- CLI regression proves `maxPriorEntryLossFraction: 1` is forwarded and permits compaction that the 25% default would defer.
- Budget regressions prove impossible writes leave `MEMORY.md` and promotion state unchanged, later retry succeeds after space is freed, and the loss ceiling preserves prior generated entries.
- Shared and symlink-aliased workspace coverage proves the smallest canonical consumer cap is selected.
- Full Memory Core suite: 1,603 passed, 8 skipped before the final review-only fixes; the affected exact-head suites were rerun afterward.
- `pnpm check:changed`: all substantive lanes passed on the prior exact head; after the review fixes, types/dead-code/core lint passed and extension lint's sole unused-import finding was removed, then `pnpm lint:extensions` passed.
- No persisted schema or data-format change.


RomneyDa:
@clawsweeper re-review

The prior CLI symlink-alias finding is addressed by resolving the canonical shared-workspace roster through the host SDK before calculating the budget. The added CLI regression proves beta’s symlink spelling still applies alpha’s smaller 9,000-character cap. This patch does not change a persisted schema or require a data migration.

RomneyDa:
@clawsweeper re-review\n\nFinal rebased head: `9283797665c2036788506a9d6640ef3e3c2bd1a7`. The canonical shared-workspace/symlink finding is fixed with end-to-end CLI coverage; exact-head focused suite is green. This changes no persisted schema or data format.

RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

RomneyDa:
CI note: the sole product test failure is `test/scripts/docs-sync-publish.test.ts`, reproduced locally on current main. It expects the pre-#143977 release route list and is missing `gateway/security/dependency-locking`. The dedicated repository-wide fix is #144002; this branch intentionally does not carry that unrelated docs-test change. I will rebase after it lands.

RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

PR #142556, opened 2026-09-08
fix: keep sandbox transports out of Linux OOM adjustment

### LINKED ISSUE #142481: Sandbox backend launchers inherit the Linux exec OOM-score wrapper
## Bug type

Behavior bug (incorrect process policy without a crash)

## Summary

On Linux, sandboxed `exec` routes still apply OpenClaw's child `oom_score_adj=1000` wrapper to the local sandbox backend launcher, even though that process is a control-plane transport and the backend owns the actual workload lifecycle.

## Steps to reproduce

1. Run the Gateway on Linux with `/bin/sh` available and without `OPENCLAW_CHILD_OOM_SCORE_ADJ=0`.
2. Configure any backend sandbox, such as Docker, Podman, SSH, or OpenShell.
3. Execute a sandboxed command through the `exec` tool.
4. Observe the launch path: `runExecProcess()` obtains the backend-owned argv from `buildExecSpec()`, submits it to the generic child supervisor with `cleanupOwnership: "external"`, and does not request exact environment handling.
5. The child adapter therefore wraps the backend launcher with `/bin/sh -c 'echo 1000 > /proc/self/oom_score_adj ...'` before starting Docker, Podman, SSH, or the OpenShell SSH transport.

## Expected behavior

Sandbox backends should own resource policy for their workloads. OpenClaw should launch the local backend transport without the generic workload OOM-score wrapper, while retaining the wrapper for ordinary Gateway-owned command children and PTYs.

The exemption should apply through the shared sandbox backend boundary rather than checking backend names, so it covers Docker, Podman, SSH, OpenShell, and future sandbox backends.

## Actual behavior

Every sandbox backend exec specification is submitted through the same child adapter as an ordinary local command. On Linux, the adapter applies the generic OOM wrapper unless the environment opts out or the spawn requests `exactEnv`.

For daemon-backed or remote sandboxes, this biases the local Docker/SSH transport rather than the actual sandbox workload. Killing that transport can lose the result while externally owned work may still require backend finalization.

## OpenClaw version

Current `main` at 2026.9.3.

## Operating system

Linux.

## Install method

Source checkout; the behavior is in the shared runtime and is independent of installation method.

## Model and provider

Not applicable; this occurs after tool routing at the process-supervisor boundary.

## Impact and severity

- Affected: Linux operators using any sandbox backend for `exec`.
- Severity: moderate reliability defect under memory pressure.
- Frequency: every eligible sandboxed exec launch on Linux; consequences surface only when the OOM killer selects the adjusted transport.
- Consequence: the sandbox transport can be preferred as an OOM victim even though it is not the workload whose memory policy the wrapper was designed to control.

## Proposed implementation

Mark backend-owned sandbox launcher spawns as exact-environment child spawns when `runExecProcess()` submits the backend's prepared argv to the supervisor. This uses the existing supervisor contract that skips environment-mutating wrappers, preserves backend-prepared environment bytes, and avoids adding backend-name checks or a second OOM-specific option.

Add regression coverage at the exec-runtime/supervisor boundary proving that sandbox backend launchers request exact environment handling while ordinary host exec continues to use the OOM wrapper. Update the Linux OOM documentation to distinguish Gateway-owned children from sandbox backend transports.

## Related work

- #70404 introduced the child-first OOM policy for Gateway-owned transient processes.
- #109402 tracks broader optional OS resource caps and explicitly treats sandbox resource controls as a separate boundary.

mrunalp:
Closes #142481

## What Problem This Solves

On Linux, sandboxed exec currently applies OpenClaw's child-first OOM-score wrapper to the local sandbox backend transport. For Docker, Podman, SSH, OpenShell, and plugin-provided backends, that local child is a transport rather than the sandbox workload, so biasing it as the first OOM victim can sever the workload's control path.

## Why This Change Was Made

This marks backend-prepared sandbox launch specifications as exact-environment child launches when they enter the generic process supervisor. That reuses the supervisor's existing wrapper-bypass contract at the ownership boundary, without adding backend-name checks or changing ordinary Gateway-owned child behavior.

The regression test covers the sandbox-to-supervisor handoff. The Linux documentation now distinguishes Gateway-owned children from backend-owned sandbox transports.

## User Impact

Sandbox backend transports no longer receive OpenClaw's `oom_score_adj=1000` wrapper. Workload resource and OOM policy remains owned by the selected sandbox backend. Non-sandboxed Gateway child processes retain the existing child-first OOM behavior.

## Evidence

- `node scripts/run-vitest.mjs src/agents/bash-tools.exec-runtime.test.ts src/process/supervisor/adapters/child.test.ts`
- `pnpm test:changed`
- `node scripts/check-changed.mjs`
- `pnpm docs:check-mdx`
- `git diff --check`

All commands above passed under the repository-supported Node 24.16 runtime. The new regression assertion failed before the production change and passed afterward.

### Linux runtime proof (Podman)

Captured locally from exact head `f92ae10a557d0d462c6a8b116d544788690e6a19` through the real `runExecProcess` and `createPodmanSandboxBackend` paths, using isolated OpenClaw state and a task-owned Alpine container:

```text
platform:                         linux
parent inherited oom_score_adj:  200
ordinary host child (bash):      1000
sandbox transport (podman):      200

ordinary host child adjusted:    PASS
sandbox transport inherited:     PASS
real Podman transport observed:  PASS
```

The task-owned container and isolated state were removed after capture, and the repository remained clean.

A fresh automated review was also attempted after validation; its engine remained active without returning a verdict for 31 minutes, so it was stopped rather than represented as a passing review.

### DIFF
--- a/docs/platforms/linux.md
+++ b/docs/platforms/linux.md
@@ -359,13 +359,18 @@ For eligible Linux child spawns, OpenClaw wraps the command in a short
 `1000`, then `exec`s the real command. This is unprivileged: a process may
 always raise its own OOM score.
 
-Covered child process surfaces:
+Covered Gateway-owned child process surfaces:
 
-- Supervisor-managed command children
+- Supervisor-managed command children, excluding sandbox backend transports
 - PTY shell children
 - MCP stdio server children
 - Managed local model and embedding service children
 - OpenClaw-launched browser/Chrome processes (via the plugin SDK process runtime)
+
+Sandbox backends own the workload lifecycle and resource policy beyond their
+local transport process. OpenClaw therefore launches prepared Docker, Podman,
+SSH, OpenShell, and plugin-provided sandbox transports without the OOM-score
+wrapper.
 
 The wrapper is Linux-only and skipped when `/bin/sh` is unavailable, or when
 the child env sets `OPENCLAW_CHILD_OOM_SCORE_ADJ` to `0`, `false`, `no`, or
--- a/src/agents/bash-tools.exec-runtime.test.ts
+++ b/src/agents/bash-tools.exec-runtime.test.ts
@@ -203,7 +203,50 @@ describe("runExecProcess cursor tracking", () => {
   });
 });
 
-describe("sandbox exec preparation failures", () => {
+describe("sandbox exec preparation", () => {
+  it("preserves the backend transport environment when spawning sandbox exec", async () => {
+    const backendEnv = {
+      BASH_ENV: "/sandbox/transport/bash-env",
+      PATH: "/sandbox/transport/bin",
+    };
+    supervisorMock.spawn.mockImplementationOnce(async (input: SpawnInput) =>
+      runtimeManagedRun(input),
+    );
+
+    const run = await runExecProcess({
+      command: "sandbox-command",
+      workdir: "/tmp",
+      env: {},
+      sandbox: {
+        containerName: "sandbox",
+        workspaceDir: "/workspace",
+        containerWorkdir: "/workspace",
+        buildExecSpec: async () => ({
+          argv: ["sandbox-transport", "exec", "sandbox-command"],
+          env: backendEnv,
+          stdinMode: "pipe-closed",
+        }),
+      },
+      usePty: false,
+      warnings: [],
+      maxOutput: 1_000,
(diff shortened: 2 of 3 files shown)


mrunalp:
@clawsweeper re-review

PR #142570, opened 2026-09-08
perf(ui): share message preparation within each render

### LINKED ISSUE #142569: Avoid repeated preparation during Chat message rerenders
## Summary

Reuse message preparation within a Chat render while preserving display, action and recovery behavior.

## Problem to solve

Actions and message bodies repeatedly normalize the same message and resolve its display text. Completed frames also have an intentionally separate footer evaluation. Text-heavy multi-message rerenders can avoid the duplicated action/body work.

## Proposed solution

Prepare raw, normalized and preview facts once per render-local message and pass them explicitly to action/body consumers. Preserve action-before-body admission and the separate frame-footer evaluation. Keep preparation out of caches, retained actions and callbacks, and leave normalizer policy unchanged.

## Impact

The scoped Lit/JSDOM workload shows 38.2%/30.5%/35.4% lower sampled allocation for 1/8/32-message groups with 32 text blocks per message. This is not a general browser/stream/small-message or latency claim: smaller control cohorts include noisy allocation increases. Prepared records deliberately remain live until the group's body map completes.

## Evidence/examples

409 candidate assertions and two existing Chromium identity cases pass, alongside 456 paired comparisons, full changed validation and scoped P2 review. Two production UI builds run the same synthetic recovery/Copy/Reply/pending/eviction flow with matching declared projections and application payloads. Clipboard transport is stubbed, and object-lifetime evidence is scoped to instrumented JSDOM.

## Do you plan to open a PR for this?

Yes, the implementation and scoped validation are ready.

steipete:
Closes #142569

<details>

</details>

## What Problem This Solves

Rerendering text-heavy Chat message groups repeats message preparation for actions and bodies. The same display and interaction results can be produced with less repeated work.

## Why This Change Was Made

This is a maintainer-requested performance cleanup. Share raw, normalized and preview facts within each synchronous render. Ordinary body paths prepare once instead of twice; completed frames keep their separate footer pass, reducing three passes to two. Four production owners grow by 11 lines, with 24 net lines in tests and fixture support.

Preparation remains local to the render and never enters a cache, retained action field or event callback. Action admission still precedes body rendering, and frame-footer actions reevaluate state afterward. The private frame-content parameter now reflects its existing array-only producer. Normalizer policy, recovery, pending/loaded-empty behavior, media and Copy/Reply semantics remain unchanged.

## User Impact

Text-heavy multi-message rerenders show lower sampled allocation in the scoped Lit/JSDOM workload. No blanket browser, streaming, small-message, latency or retained-memory improvement is claimed.

Preserving action-before-body order holds each group's prepared records until its body map completes. This temporary lifetime extension is intentional; it is not a cross-render cache.

## Evidence

- 361 original assertions passed before editing; the candidate passes 409 assertions across nine suites and both existing Chromium tool-identity cases. The full changed gate and both managed P2 review parts pass.
- 456 paired comparisons pass, including 54 exact HTML/action/footer outputs, in-place mutation, frame admission order/state/labels and scoped normalized-object collection checks.
- Two production UI builds run the same synthetic four-session flow. The declared message projection, recovery counts, Copy/Reply payloads, loaded-empty/pending behavior and eviction behavior match. Clipboard writes use a stub; this does not prove native clipboard transport or whole-UI pixel equality.
- Actual Lit rerenders with 32 text blocks per message show 38.2%/30.5%/35.4% lower sampled allocation for 1/8/32-message groups; the 32-message cohort is about 4.93 → 3.19 MB per render. These are three-trial JSDOM observations.
- Counter-results remain part of the evidence: streaming total allocation changes only −0.8%/−0.4%; short, loaded-empty and pending-user totals rise 1.4%/35.9%/12.5%, with overlapping noisy ranges. First-body forced-GC median differences of +11,688/+20,112/+62,592 bytes are noisy checkpoints, not peak or isolated normalized-object costs. Later collection checks are scoped JSDOM proof, not a universal browser-lifetime guarantee.

The retained-pane fixture exposed an existing move/reload behavior: hidden empty content is restored by a second fetch, rather than retained across movement. Both arms follow that same request sequence; the original fixture failure and separate follow-up remain recorded.

The inspected synthetic before/after pair shows the same recovered Copy/Reply state.

Before: synthetic recovered Copy and Reply state

After: preserved recovered Copy and Reply state

### DIFF
--- a/ui/src/pages/chat/chat-view.plugin-identity.test.ts
+++ b/ui/src/pages/chat/chat-view.plugin-identity.test.ts
@@ -14,8 +14,9 @@ import { createComposerProps, resetComposerFixture } from "./chat-composer.test-
 import { createTestTranscript } from "./chat-view.test-helpers.ts";
 import { renderChat, type ChatProps } from "./chat-view.ts";
 import { renderGroupedMessage } from "./components/chat-message-bubble.ts";
-import { threadProps } from "./components/chat-transcript.test-support.ts";
+import { prepareChatMessageRender } from "./components/chat-message-markdown.ts";
 import "../../plugins/control-ui-view.runtime.ts";
+import { threadProps } from "./components/chat-transcript.test-support.ts";
 
 afterEach(() => resetComposerFixture());
 
@@ -88,12 +89,12 @@ describe("native chat view session identity", () => {
       override render() {
         return html`${renderChat(props)}
         ${renderGroupedMessage(
-          {
+          prepareChatMessageRender({
             role: "toolResult",
             toolCallId: "identity-result",
             toolName: "inspect",
             content: [{ type: "text", text: "Inspection complete" }],
-          },
+          }),
           "identity-message",
           {
             isStreaming: false,
--- a/ui/src/pages/chat/chat-view.test-helpers.ts
+++ b/ui/src/pages/chat/chat-view.test-helpers.ts
@@ -1,6 +1,9 @@
 import type { ReactiveControllerHost } from "lit";
 import { vi } from "vitest";
-import { resolveMessageActionDetails } from "./components/chat-message-markdown.ts";
+import {
+  prepareChatMessageRender,
+  resolveMessageActionDetails,
+} from "./components/chat-message-markdown.ts";
 import { ChatTranscriptController } from "./components/chat-transcript-controller.ts";
 
 export function createTestTranscript(): ChatTranscriptController {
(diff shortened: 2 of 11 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f3eb01233afa49ae896eb1ffd1e4d490f9ed5247](https://github.com/openclaw/openclaw/pull/142570/commits/f3eb01233afa49ae896eb1ffd1e4d490f9ed5247)
- Landed commit: [d0db7d9e61e8a96fef99fffa3c817112a0a8980e](https://github.com/openclaw/openclaw/commit/d0db7d9e61e8a96fef99fffa3c817112a0a8980e)

PR #142572, opened 2026-09-08
fix(channels): preserve 7.33 Telegram workspace media access

RomneyDa:
## Summary

- build the agent-scoped outbound media authority at the Gateway message-action boundary
- preserve that complete `mediaAccess` contract through Telegram action and outbound adapters
- pass `workspaceDir` into the final media loader so relative attachments resolve under the same authority as direct sends
- add regressions across Gateway, adapter, action runtime, outbound adapter, and final loader boundaries

## Root cause

The 7.33 candidate reduced the Gateway-owned media authority to `mediaLocalRoots` and `mediaReadFile` at multiple handoffs. That discarded `workspaceDir`, so direct Gateway sends succeeded while `message.action` rejected the same relative attachment as missing.

## Verification

- Gateway suites: 126/126
- Telegram suites: 275/275
- production TypeScript: core + extensions green
- Oxfmt and `git diff --check`

This is the complete 7.33 port of the already-proven media authority correction; it does not weaken the package acceptance scenario.

### DIFF
--- a/extensions/telegram/src/action-runtime.test.ts
+++ b/extensions/telegram/src/action-runtime.test.ts
@@ -1153,7 +1153,12 @@ describe("handleTelegramAction", () => {
         content: "Hello with local media",
       },
       telegramConfig(),
-      { mediaLocalRoots: ["/tmp/agent-root"] },
+      {
+        mediaAccess: {
+          localRoots: ["/tmp/agent-root"],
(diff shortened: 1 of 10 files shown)


RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

PR #142573, opened 2026-09-08
fix(release): preserve frozen update compatibility

RomneyDa:
## Summary

- capability-bind 7.33 package dry-run reporting in the update-channel assertion
- retain authored non-plugin config while parking restart probes with plugins disabled
- preserve config metadata and size so a frozen baseline cannot auto-restore plugin entries from `.last-good`

## Root causes

1. 2026.7.33 stores and resolves the `dev` channel correctly but reports the dry-run install kind as `package`/`npm` rather than current tooling's expected `git` switch. The compatibility projection is version-scoped and fail-closed unless the runner sets it.
2. The restart-auth probe replaced a full authored config with a tiny config lacking `meta`. Baseline recovery treated that as corruption, restored plugin entries, then Doctor installed 7.33-era fixtures into a 2026.7.1-2 runtime and refused startup. Parking now preserves all authored non-plugin fields while replacing plugins and gateway probe settings.

## Validation

- tooling Docker contracts: 285 passed, 2 skipped
- package compatibility/config parking: 54 passed
- shell syntax, Oxfmt, `git diff --check`

This covers the duplicated update-channel and update-restart-auth failures in the 2026.7.33 Full Validation run.

### DIFF
--- a/scripts/e2e/lib/package-compat.mjs
+++ b/scripts/e2e/lib/package-compat.mjs
@@ -11,6 +11,14 @@ export function legacyPackageAcceptanceCompat(version) {
   );
 }
 
+export function updateChannelDryRunReportsPackageCompat(version) {
+  const match = /^(\d{4})\.(\d{1,2})\.(\d{1,2})(?:[-+].*)?/.exec(version || "");
+  const [year, month, day] = match?.slice(1, 4).map(Number) ?? [];
+  return (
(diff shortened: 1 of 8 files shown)


RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

Exact-head CI is now green, including docker-seed-e2e. The focused compatibility suites also pass 343 checks (plus 2 skips); the config parking change affects only isolated release-test fixtures, not OpenClaw production persisted state.

PR #142577, opened 2026-09-08
fix: show attachment overflow after adding files or resizing chat

steipete:
<details>

This is a same-repository branch; maintainers with repository push access can update it.

</details>

## What Problem This Solves

Fixes an issue where users adding attachments or narrowing Chat or New Session would lose the faded-edge cue while more files were clipped offscreen. Composer menus could also keep stale overflow indicators when their results changed.

## Why This Change Was Made

Mount-only refs left retained elements with stale measurements. One Lit directive now owns overflow measurements after rendering, size observation, scrolling, and disconnect/reconnect cleanup. It replaces the separate attachment, menu, and goal calculations; the queue's intentional height cap remains unchanged.

## User Impact

Faded edges follow the current content and available width, including after adding or removing files and resizing a composer. Existing menu scroll options and mobile goal behavior are preserved. Production code is reduced by nine lines.

## Evidence

Three real-browser regressions failed against the original implementation. All five final browser tests pass, covering attachment changes and scrolling, width changes, retained-host reconnection, menu result changes, and expanded mobile goals. The 36 existing goal and queue tests also pass.

Isolated Chromium exercised the actual file picker in Chat and New Session, then appended files or resized the viewport. All four cases changed from stale indicators before the fix to the correct fade afterward, with no page errors. These runs use source Vite and a mocked Gateway with synthetic files; they do not test live provider delivery.

**Chat after adding files, before fix**

Chat after adding files, before fix

**Chat after adding files, after fix**

Chat after adding files, after fix

**New Session after narrowing, before fix**

New Session after narrowing, before fix

**New Session after narrowing, after fix**

New Session after narrowing, after fix

The explicit seven-path `check-changed` gate and `pnpm ui:build` passed, including the UI asset budgets. Independent managed review is clean through P2. The first check run found two test-helper lint issues, corrected before the final run; an initial browser collection was interrupted by first-use Vite dependency optimization, then passed after optimization completed.

The first CI run also exposed an unrelated race in the self-upgrade test fixture: its parent could see the JSON record before the child finished writing it. The fixture now writes beside the destination and renames the complete record into place. Existing assertions, waits, timeouts, and process cleanup remain unchanged. All five actual manager-fixture tests and four focused publication controls passed locally; the corrective changed gate and independent P2 review also passed. This adds no production code and preserves the original UI proof.


steipete:
Merged via squash.

- Prepared head SHA: [6d1f267ecc4ebf3f6bf3ba96db96713ae1256529](https://github.com/openclaw/openclaw/pull/142577/commits/6d1f267ecc4ebf3f6bf3ba96db96713ae1256529)
- Landed commit: [92aa0bafbe5e88ae7963d5f539ffccc532f042d2](https://github.com/openclaw/openclaw/commit/92aa0bafbe5e88ae7963d5f539ffccc532f042d2)

PR #142589, opened 2026-09-08
fix(security): preserve existing WhatsApp group allowlists

eleqtrizit:
## What changed

`openclaw security audit --fix` changes an open WhatsApp group policy to an allowlist. When it does that, it copies existing pairing approvals into `groupAllowFrom` so the operator is not accidentally locked out.

The command currently changes the policy first. By the time the WhatsApp-specific repair runs, it cannot tell the difference between:

- a policy that was open and was just converted; and
- a policy the user had already configured as an allowlist.

As a result, running the command can add paired users to an existing group allowlist even though that policy was not changed by the command.

This fix runs the WhatsApp-specific step while the original policy is still visible. Pairing approvals are copied only when that policy is actually being converted from `open` to `allowlist`. Existing allowlists, disabled policies, explicit sender lists, and account-default inheritance are left unchanged.

## Tests

- Focused tests cover both paths: converting an open policy still seeds the list, while an existing allowlist is not modified.
- Full WhatsApp extension test suite: 1,411 tests passed.
- Full repository checks passed.

### DIFF
--- a/extensions/whatsapp/src/security-fix.test.ts
+++ b/extensions/whatsapp/src/security-fix.test.ts
@@ -1,0 +1,86 @@
+import type { OpenClawConfig } from "openclaw/plugin-sdk/config-contracts";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { readChannelAllowFromStoreMock } = vi.hoisted(() => ({
+  readChannelAllowFromStoreMock: vi.fn(async () => [] as string[]),
+}));
+
+vi.mock("openclaw/plugin-sdk/channel-pairing", () => ({
+  readChannelAllowFromStore: readChannelAllowFromStoreMock,
+}));
+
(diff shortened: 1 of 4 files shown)


eleqtrizit:
@clawsweeper re-review

eleqtrizit:
## Merge record

**Summary:** This fixes `security audit --fix` so it only copies WhatsApp pairing approvals when converting an open group policy into an allowlist. Existing allowlists, explicit sender lists, disabled policies, and account-default inheritance remain unchanged.

**Decision Reason for Merge:** Merged because the narrow fix preserves existing configuration while preventing `security audit --fix` from expanding existing WhatsApp group allowlists, all 107 checks passed, and the maintainer approved the compatibility-preserving behavior.

**Compatibility Checks Done:** Focused regression tests covered both the conversion and no-mutation paths. The full WhatsApp extension suite, repository checks, compatibility review, and all 107 PR checks passed.

Issue #105188, opened 2026-07-12
[Bug]: Session Reaper lacks backoff on persistence errors causing disk thrashing


aniruddhaadak80:
### Summary
If the session cleanup cron job catches an error, the throttle timestamp is never updated, causing instant retries and disk thrashing.

### Steps to reproduce
1. Trigger an EACCES or disk full error during sweepCronRunSessions.
2. Observe the reaper instantly retrying on the next timer tick.

### Expected behavior
The reaper should back off and wait for the standard cooldown period even if an error occurs.

### Actual behavior
The reaper enters an unbounded retry loop, leading to severe CPU and disk thrashing.

### OpenClaw version
Latest main

### Operating system
Cross-platform

### Impact and severity
Affected: Systems with temporary disk or permission issues.
Severity: High (Resource exhaustion).
Frequency: Edge case.
Consequence: System instability and I-O overload.

PR #142591, opened 2026-09-08
fix(cron): list run sessions read-only in the reaper to stop event-loop stalls (#142476)

### LINKED ISSUE #142476: 2026.9.3: cron session reaper opens every agent database with a synchronous PRAGMA integrity_check, blocking the event loop 14-76s every few minutes on a 632-agent gateway

zachisfine:
## What Problem This Solves

The cron run-session reaper (`sweepCronRunSessions`) opens every agent database on a fixed interval to list session entries and decide which expired cron run sessions to prune. It listed via `listSessionEntriesCore`, whose writable open runs a synchronous `PRAGMA integrity_check` plus a foreign-key check on **every open**.

On a large fleet, one sweep pass opens ~every agent database and runs a full integrity + foreign-key check on each, on the main thread, every `MIN_SWEEP_INTERVAL_MS` (5 min). In #142476 the gateway event loop stalled 14–76s at a time on a 632-agent config, so any liveness supervisor with a ~5s timeout keeps declaring a healthy gateway dead.

Fixes #142476.

## Change

Switch the reaper's listing from `listSessionEntriesCore` to `listSessionEntriesReadOnly`, which uses the read-only agent-database open (`withOpenClawAgentDatabaseReadOnly`) and does not run the integrity gate.

Why this is safe:
- The reaper only **reads** rows to select removals; it does not write through this path.
- The read-only listing's default `"full"` projection returns **owned** entries (`projectSessionEntriesForListing`: "Full snapshots own their nested values"), so they remain safe to hold across the `await` on `applySessionEntryLifecycleMutation` below — identical to the previous `listSessionEntriesCore` full-projection behavior.
- The actual pruning **write** still goes through `applySessionEntryLifecycleMutation`, which keeps its own integrity gate, so corruption is still caught whenever a row is actually mutated — and that write only runs when there is something to prune, not on every sweep.

## Evidence

Validated locally against this change on `main` (2026.9.3), Node 24.18.0:

- **Regression test added** — asserts the sweep lists via the read-only accessor and never the integrity-gated `listSessionEntriesCore`, even when there is nothing to prune (the fleet-wide hot path from the issue).
- **Existing `#105188` test updated** to spy on the read-only accessor the reaper now calls; its intent (the throttle is updated after a listing error so the next tick does not thrash) is unchanged and still verified.
- `pnpm exec vitest run src/cron/session-reaper.test.ts` → **28 passed** (26 pre-existing + updated #105188 + new #142476 regression).
- `pnpm exec tsc --noEmit -p tsconfig.core.json` → **0 errors**.

## Fleet measurement (independent, 2026-09-11)

The reporter of #142476 ran this change on the 632-agent gateway from the report and [posted the numbers](https://github.com/openclaw/openclaw/issues/142476#issuecomment-5622204086): same host, same sanitized state copy per arm, Node 24.20.0, one boot per arm, 10 ms `setInterval` max-delay sampling and `/health` once a second.

Steady state, minutes 5 to 15 after ready:

| arm | loop blocked | `/health` timeouts |
|---|---|---|
| stock 2026.9.3 | 56s | 2.6% |
| **this PR** | **17s** | **0.2%** |
| #142534 (read-only + integrity gating) | 31s | 0.4% |
| stock + `cron.sessionRetention=false` (reaper off) | 30s | 0.4% |
| stock 2026.9.4 | 40s | 0.7% |

Two things worth drawing out:

- **The reaper-off arm is the ceiling for any reaper fix, and this change reaches it.** The reporter reads arms 2, 3 and 4 as one band and calls the 17s vs 30s gap single-boot noise. That is the right reading.
- **The integrity gating in #142534 adds nothing on top.** It is indistinguishable from the arm that never runs the reaper at all, while carrying a persisted-schema edit and a fleet-wide change to when integrity checks run.

## A second cost this also removes

The measurement caught a 5s+ block during a sweep with **zero** file opens, with the main thread in `openOpenClawAgentDatabase` → `evictLruAgentDatabaseHandles` → `closeCachedOpenClawAgentDatabase` → `releaseOpenClawAgentDatabaseLease` → `runOpenClawStateWriteTransaction` → `acquireStateDatabaseCoordinator`.

That cost is inside the **writable** open: `evictLruAgentDatabaseHandles()` is called from `openOpenClawAgentDatabase` and nowhere else (`src/state/openclaw-agent-db.ts:312`, again at `:384`). The read-only open constructs its connection, validates schema version and owner, and closes it directly — it never enters the handle cache, claims a lease, or opens a write transaction on the shared state database.

So the switch is not only skipping `PRAGMA integrity_check`. It also takes the sweep off the lease-and-eviction path that serializes every open through one write transaction on the state coordinator, which is why a warm handle cache did not make the sweep cheap. The code comment records both reasons.

Measured on a 632-agent fleet, 449 of 594 databases are opened twice each per recurring pass, so the 64-handle cache retains none of this fleet between passes.

Not this PR, filed separately by the reporter as #145184: the once-per-boot post-ready session hydration (70–82s) and the Gateway session-list prewarm. Both survive with the reaper disabled entirely, so a reaper fix will not close them.


zachisfine:
CI status note for triage: the change itself looks green. The diff is limited to `src/cron/session-reaper.ts` and its test, and the `src/cron` checks pass. The only red is `checks-node-changed-extensions-bundle-12` (and the `openclaw/ci-gate` that aggregates it).

That shard is failing on an unrelated, flaky Discord test **timing out**, not on anything in this change:

- Run 1: `extensions/discord/src/monitor/native-command-reply.test.ts`
- Runs 2 and 3: `extensions/discord/src/monitor/native-command.reset.test.ts` — `Error: Test timed out in 120000ms.` at `native-command.reset.test.ts:100` (a parameterized `it.each` case)

All three are timeouts on the same shard (`changed-extensions-config-36`: 266 files / 3,696 tests, 3 workers, ~255s). Each of those tests passes locally in milliseconds (13/13 and 16/16 respectively), so this reads as CI shard resource/timing pressure rather than a real failure — this being a core change, it triggers the full changed-extensions test matrix.

Reproduced across 3 runs, each on latest `main` (rebased). Could a maintainer re-run bundle-12 / that shard when convenient? External contributors can't re-run jobs. Happy to dig further if any of this turns out to be attributable to the change.

jalehman:
@clawsweeper re-review

Freshness renewal for the unchanged current head `b92cf1875e6857e7debee4671a20eba42118c59a`: native merge verification requires a completed review under 12 hours old. The previous review completed on September 11 at 22:39 UTC. Review only; no repair or merge requested.

steipete:
Merged via squash.

- Prepared head SHA: [b92cf1875e6857e7debee4671a20eba42118c59a](https://github.com/openclaw/openclaw/pull/142591/commits/b92cf1875e6857e7debee4671a20eba42118c59a)
- Landed commit: [86e43eb402e39fc834d06ff15f8690e2a7335147](https://github.com/openclaw/openclaw/commit/86e43eb402e39fc834d06ff15f8690e2a7335147)

PR #142600, opened 2026-09-08
fix(signal): preserve final reply quotes after failed streamed sends

steipete:
## What Problem This Solves

Fixes an issue where Signal users with `replyToMode: first` received an unquoted final answer when an earlier streamed text or media block failed to send. The successful final answer lost its link to the incoming message even though the earlier reply never reached the recipient.

## Why This Change Was Made

The monitor now uses the same quote plan as durable final delivery: inspect the next quote before sending and consume the slot only after the send succeeds. One shared text/media send path replaces duplicated option and result assembly, removing the consume-on-attempt resolver and 51 production lines.

## User Impact

The first successful reply retains the incoming message's quote after a streamed block fails. Successful first replies still consume their slot; `all`, `batched`, explicit replies, and status notices retain their existing behavior.

## Evidence

- Before the production edit, the actual monitor regression table failed both first-mode cases (failed text and failed media) and passed six original-order controls. After the repair, all eight pass, including a rerun against the final reviewed source.
- 57 tests passed across the monitor reply suite, partial-final-delivery suite, and approval/presentation delivery suite. Coverage includes successful text/media blocks, quote modes, chunking, UUID/group/account overrides, notices, and existing partial-delivery handling.
- The complete changed gate passed for all three changed paths, including extension and test typechecks, lint, dead exports, and repository guards. A mechanical unbound-method lint finding in the first run was corrected before the passing run.
- Independent P2 review is clean. An initial concern about post-send persistence rejection was resolved by inspecting the existing registration owner and executing five synthetic store-failure/success cases. That owner catches opening, update, and lookup failures; no persistence change was needed.

The monitor proof uses the real monitor, isolated ingress, dispatcher, and durable final delivery with synthetic messages and mocked Signal transport/model replies. It verifies outgoing text order and quote metadata. It does not exercise a live Signal account, external provider, or mobile app rendering. Batched single-message behavior is a preservation control, not a separate defect.

The quoted-reply owners, callers, planner, and dependency lock were also compared with main at `63943fffb46c5fc1714d46f7aa5ef53170b492cd`; their contracts are unchanged.

### DIFF
--- a/extensions/signal/src/monitor.tool-result.sends-tool-summaries-responseprefix.test.ts
+++ b/extensions/signal/src/monitor.tool-result.sends-tool-summaries-responseprefix.test.ts
@@ -1,6 +1,8 @@
 // Signal tests cover monitor.tool result.sends tool summaries responseprefix plugin behavior.
 import { expectPairingReplyText } from "openclaw/plugin-sdk/channel-test-helpers";
 import type { OpenClawConfig } from "openclaw/plugin-sdk/config-contracts";
+import { PlatformMessageNotDispatchedError } from "openclaw/plugin-sdk/error-runtime";
+import type { GetReplyOptions } from "openclaw/plugin-sdk/reply-runtime";
 import { resolveAgentRoute } from "openclaw/plugin-sdk/routing";
 import { normalizeE164 } from "openclaw/plugin-sdk/text-utility-runtime";
 import { describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c7d672700feca265fe3612e90029f8a11df7735a](https://github.com/openclaw/openclaw/pull/142600/commits/c7d672700feca265fe3612e90029f8a11df7735a)
- Landed commit: [39fa56fc30a7f772d1f78baebfec0d24d80557a1](https://github.com/openclaw/openclaw/commit/39fa56fc30a7f772d1f78baebfec0d24d80557a1)

PR #142601, opened 2026-09-08
refactor(apple): centralize gateway error overrides

### LINKED ISSUE #142594: Centralize Apple Gateway error presentation overrides
Apple Gateway connection errors repeat presentation-override resolution at 24 case constructions before calling the same existing problem factory. This duplicates title/message/action/documentation/request-ID shaping while the cases already have one owner for fallback facts.

Move override resolution into that existing private factory and keep the cases as defaults. Preserve all public problem fields, unknown-message handling, localized versus verbatim classification, pairing/protocol distinctions and retry/reconnect behavior. The full one-file change removes 93 production code/physical lines without changing public APIs, resources or platform wrappers.

The actual shared package builds, native and Apple localization checks are unchanged, and the required changed gate passes. A sandboxed filtered run passes 24 tests, including unchanged error tests, the real node-session protocol-error caller and 8,775 complete Equatable comparisons of original/candidate outputs across 39 scenarios.

A fresh independent P2 review claimed that ownerRaw newly overrides recognized auth cases. The pinned original factory already used the identical expression for every case; source and executable comparisons disprove the finding. No accepted findings remain; the raw review and rejection evidence are retained separately.

Related #118107 is the merged predecessor that introduced shared construction while retaining these per-case overrides; this continues its existing owner.

steipete:
Closes #142594

## What Problem This Solves

Apple Gateway connection errors repeat the same presentation-override policy in 24 case constructions. Maintaining that policy in each case makes future changes harder to keep consistent.

## Why This Change Was Made

Move override selection and output shaping into the existing private problem factory. Each case retains its own fallback values. Public fields, string literals, localized/verbatim handling, pairing/protocol variants, URL fallback, owner hints and retry/reconnect flags are preserved.

## User Impact

Existing Gateway messages, actions and diagnostics stay the same. The complete one-file change removes 93 production code/physical lines and introduces no helper, adapter, public API, resource or dependency change.

## Evidence

- The actual shared Swift package and tests compile. All 22 unchanged GatewayErrorsTests, the real GatewayNodeSession protocol-error case and the differential characterization pass: 24 tests in exactly three audited suites.
- The characterization executes the original mapper and candidate and compares 8,775 complete Equatable outputs across 39 scenarios and override combinations, including false flags, owner aliases, invalid URLs and equal-copy localization cases.
- Native localization check: 3,783 entries, changed=false. Apple catalogs and the required changed gate pass; no retained generated/test/tooling changes.
- Fresh P2 review reported one owner-override regression. It is rejected against the [original factory](https://github.com/openclaw/openclaw/blob/83c609b0365dc349e580f203879891127b9bff00/apps/shared/OpenClawKit/Sources/OpenClawKit/GatewayConnectionProblem.swift#L962): both old and new code already use `owner: authError.ownerRaw.flatMap(self.owner(from:)) ?? defaults.owner`. The complete-output comparisons include its claimed case. No accepted findings remain; the raw review is preserved with its separate source-backed rejection.
- Native tests ran inside an OS sandbox denying access to operator files, network, credential/preferences/window services and external process/service control. An earlier unfiltered launcher probe aborted inside that sandbox and has zero passing credit. The exact filtered launch was then verified with separate synthetic selector canaries before the successful product run.

Related #118107 is the already-merged shared-construction predecessor; its public contracts and ownership boundaries remain.

### DIFF
--- a/apps/shared/OpenClawKit/Sources/OpenClawKit/GatewayConnectionProblem.swift
+++ b/apps/shared/OpenClawKit/Sources/OpenClawKit/GatewayConnectionProblem.swift
@@ -176,8 +176,7 @@ public enum GatewayConnectionProblemMapper {
         let message: String
         let actionLabel: String?
         let actionCommand: String?
-        let docsURL: URL?
-        let requestId: String?
+        let docsURLString: String?
         let retryable: Bool
         let pauseReconnect: Bool
     }
@@ -263,15 +262,11 @@ public enum GatewayConnectionProblemMapper {
                 .init(
                     kind: .tailscaleIdentityMissing,
                     owner: .network,
-                    title: authError.titleOverride ?? "Tailscale identity check failed",
-                    message: authError.userMessageOverride
-                        ?? "This connection expected Tailscale identity headers, but they were not available.",
-                    actionLabel: authError.actionLabel ?? "Turn on Tailscale",
-                    actionCommand: authError.actionCommand,
-                    docsURL: self.docsURL(
-                        authError.docsURLString,
-                        fallback: "https://docs.openclaw.ai/gateway/tailscale"),
-                    requestId: authError.requestId,
+                    title: "Tailscale identity check failed",
+                    message: "This connection expected Tailscale identity headers, but they were not available.",
+                    actionLabel: "Turn on Tailscale",
+                    actionCommand: nil,
+                    docsURLString: "https://docs.openclaw.ai/gateway/tailscale",
                     retryable: false,
                     pauseReconnect: true),
                 authError: authError)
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [61e39c189118e7b02e0ca7c691b76bc40137ca22](https://github.com/openclaw/openclaw/pull/142601/commits/61e39c189118e7b02e0ca7c691b76bc40137ca22)
- Landed commit: [57f02536f4d027bbe65c47549564b0dd5fe1b129](https://github.com/openclaw/openclaw/commit/57f02536f4d027bbe65c47549564b0dd5fe1b129)

PR #142480, opened 2026-09-08
fix(browser): report blocked and interrupted CLI actions

steipete:
## What Problem This Solves

Fixes an issue where users running browser CLI actions would see `null` or a success message when a modal dialog blocked the action, or when navigation or page closure stopped a batch before its remaining actions ran.

## Why This Change Was Made

One shared CLI action runner now presents the outcomes already recorded by the browser: pending dialog IDs, interrupted batches, and failed child actions. This removes parallel output paths and repeated error handling, with 20 fewer production lines. Browser execution, request deadlines, and protocol behavior are unchanged.

## User Impact

Users can identify a pending dialog and see how many batch actions were skipped before deciding what to do next. Successful action messages, primitive evaluate values, and complete `--json` output are preserved. Failed batch children still exit nonzero; a failure combined with navigation reports both the error and skipped actions.

## Evidence

| Actual scenario | Before | After |
| --- | --- | --- |
| Evaluate opens a confirm dialog | `null` | Reports the blocked action, dialog ID, and accept/dismiss guidance |
| Press opens a confirm dialog | `pressed Enter` | Reports the blocked action and dialog ID |
| First batch action reloads or closes its page | `batch ran 2 action(s)` | Reports the stop after action 1 and one skipped action |
| First batch action fails while reloading | Child error only, exit 1 | Child error plus skipped count, exit 1 |

- Eight regression assertions fail against the original code. All 83 focused tests pass, including CLI registration, JSON output, primitive values, failed-child exit behavior, and lazy loading.
- Ten scenarios pass before and after through isolated Chromium, a local synthetic page, the actual CLI parser, and the real `/act` and `/hooks/dialog` HTTP routes. Returned dialog IDs were answered through the real route. The fixture adapts Gateway transport to loopback HTTP; this is not a full installed-Gateway end-to-end claim. Independent modal cases use fresh pages and do not claim sequential dialog lifecycle behavior.
- Required changed-scope checks pass, including extension production/test typechecks, lint, formatting, boundaries, and import-cycle guards. The assertion baseline shrinks by the exact entry for two removed element casts.
- Independent P0–P2 review found no actionable issues. Current-main qualification found the browser output owners and dependency pins unchanged; newer, disjoint baseline reductions are preserved by the normal merge.

PR #142608, opened 2026-09-08
fix(browser): keep newer dialogs pending after earlier actions settle

steipete:
## What Problem This Solves

Fixes an issue where users answering sequential managed-browser dialogs could receive `Dialog "d2" is not pending` even though the second dialog remained open. This happens when the first action finishes its delayed cleanup after a new dialog has appeared.

## Why This Change Was Made

Scope late dialog reconciliation to the IDs recorded when that action was interrupted. Share closed-dialog record construction and remove the redundant pending-record serializer, deleting ten production lines. Existing external-close reconciliation remains in place; authoritative CDP close-event ownership is a separate follow-up.

## User Impact

A newly opened dialog remains pending and can be answered by its returned ID. Explicit responses, armed responses, prompt text, and externally handled dialogs retain their existing behavior. This complements the CLI presentation fix in #142480.

## Evidence

- Before: real registered `/act` and `/hooks/dialog` HTTP routes with Playwright 1.62.1 and isolated Chromium returned `d2`, then rejected its response with HTTP 500. Independent native CDP events showed only `d1` had closed while bookkeeping incorrectly recorded `d2` as remotely closed.
- After: the same sequential flow retains `d2` as pending and answers it with HTTP 200. External closure, armed dismissal, prompt text, normal evaluation, and two sequential keyboard-triggered dialogs pass.
- Two new regressions failed on the original code; all 101 tests across five focused browser test files pass after the repair.
- Required changed-scope checks passed, including production/test typechecks, lint, dead-export scans, and architecture guards. Independent P2 review found no actionable issues.

The browser proof used task-owned localhost pages and a fresh browser profile. It exercises real browser execution and dialog bookkeeping; it does not claim signed-in operator-browser or extension-relay coverage. Remote closure still uses the existing action-settlement mechanism, now scoped to the recorded dialogs.

### DIFF
--- a/extensions/browser/src/browser/pw-session-dialogs.ts
+++ b/extensions/browser/src/browser/pw-session-dialogs.ts
@@ -19,11 +19,17 @@ export function resolveObservedDialogTimeoutMs(timeoutMs: number | undefined): n
   return Math.max(1, Math.floor(parsed ?? OBSERVED_DIALOG_TIMEOUT_MS));
 }
 
-export function appendRecentDialog(state: PageState, record: BrowserObservedDialogRecord): void {
+export function recordClosedDialog(
+  state: PageState,
+  dialog: BrowserObservedDialogRecord,
+  closedBy: NonNullable<BrowserObservedDialogRecord["closedBy"]>,
+): BrowserObservedDialogRecord {
+  const record = { ...serializeDialogRecord(dialog), closedAt: new Date().toISOString(), closedBy };
   state.recentDialogs.push(record);
   while (state.recentDialogs.length > MAX_RECENT_DIALOGS) {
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ef3f9c74fa21bf70d843b21764df6d3412629260](https://github.com/openclaw/openclaw/pull/142608/commits/ef3f9c74fa21bf70d843b21764df6d3412629260)
- Landed commit: [99d9c43695501edbd1cc19e3db93ec0f318d0897](https://github.com/openclaw/openclaw/commit/99d9c43695501edbd1cc19e3db93ec0f318d0897)

PR #142612, opened 2026-09-08
fix: reduce cold presence work when viewing multiple sessions

steipete:
## What Problem This Solves

Fixes an issue where viewing several sessions could slow presence updates when their session store had no retained database handle. Each distinct watched session could repeat validation of every session metadata row in the store.

## Why This Change Was Made

Presence now prepares exact watched-session reads through the existing grouped reader after the first eligible recipient requests a watched reference. It retains each logical result or error and exposes errors in the original visitor order. Full selected-entry decoding, canonical corruption refusal, recipient visibility checks, and deleted-main fallback rules remain intact. Existing eager batch callers keep their failure ordering.

The change uses the existing grouped admission boundary: read-only work can happen earlier, and an admission or close failure affects its store group. It adds no schema, transaction, configuration, or handle-cache API.

## User Impact

Presence updates avoid repeated cold metadata validation when users view multiple sessions. Pending or unauthorized recipients still trigger no session reads, and each recipient retains the existing visibility filtering.

## Evidence

- The regression failed on the unchanged baseline with eight metadata censuses for eight watched sessions over 64 rows; the candidate requires and performs one. Repeated watcher rows and recipients reuse those results.
- 308 focused tests passed across presence, store lookup, serialization, session utilities, and read-only accessors. After consolidating the two new cases into the existing viewer-presence suite to preserve the test-root budget, all six cases there passed.
- Tests cover full selected-entry projection, independent logical errors, deferred visitor order versus existing eager failure order, and legacy success/error suppression of replacement planning.
- Changed-file checks passed, including typechecks and the existing test-root limits. Independent autoreview is clear through P2.

- Exact-head [CI passed](https://github.com/openclaw/openclaw/actions/runs/34282710781) for `a38d161dcde174b53783b6a2380eb5bffad2b226`. Current ClawSweeper review reports no concrete code defect.
- Linux Node 24.19.0 proof used `blacksmith-testbox`, lease `tbx_01m21gz22cjv2rz64c6apb90gn` ([run](https://github.com/openclaw/openclaw/actions/runs/34284230537)), comparing exact baseline `8cca96e1185994165a4e475679f6d16874392c25` and candidate. Both frozen installs and builds passed. All ten measurement pairs produced identical output. For 1,024 rows and eight watched sessions, instrumented canonical censuses fell from eight to one, rows visited from 8,192 to 1,024, and exact reads remained eight; every observed handle closed. The uninstrumented cold projection observations were 69.97 → 14.84 ms for eight watches and 183.60 → 15.14 ms for 32 watches. These are bounded single-cell observations, not production latency estimates.
- The ordinary package-bin `node openclaw.mjs gateway run` flow passed on both revisions with fresh isolated state, 30 sampled presence RPCs each, matching declaration ACK/event/RPC output at 1/8/32 watches, creator/draft/incognito filtering, fresh visibility changes, and declaration clearing. The serving PID was the directly spawned process with respawn disabled; both Gateways exited zero with no signal, all sockets/process groups closed, and the exact listeners refused connection.

Two initial dispatches were rejected before workload execution because of unsupported delegated transport flags. The first runtime attempt preserved all successful measurements but failed its synthetic Control UI handshake because the client omitted the built ID; the Gateway correctly rejected it. A separately reviewed Gateway-only follow-up supplied each revision's real build ID and verified the server echoed it. The original failure remains retained; no assertion or deadline was relaxed. Full synthetic state evidence is retained privately rather than published. This PR does not attribute historical stalls to this specific path.

### DIFF
--- a/src/gateway/presence-projection.ts
+++ b/src/gateway/presence-projection.ts
@@ -1,3 +1,5 @@
+import { expectDefined } from "@openclaw/normalization-core";
+import { err, ok, type Result } from "@openclaw/normalization-core/result";
 import type { SessionEntry } from "../config/sessions.js";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import type { SystemPresence } from "../infra/system-presence.js";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a38d161dcde174b53783b6a2380eb5bffad2b226](https://github.com/openclaw/openclaw/pull/142612/commits/a38d161dcde174b53783b6a2380eb5bffad2b226)
- Landed commit: [00bd9750be30a8938473e3d738ca75e7bdb9e9ac](https://github.com/openclaw/openclaw/commit/00bd9750be30a8938473e3d738ca75e7bdb9e9ac)

PR #142614, opened 2026-09-08
refactor(tests): retire duplicate deep import memory case

steipete:
## What Problem This Solves

The SQLite import memory suite runs the same deep transcript twice: once through the direct importer and again through public Doctor orchestration.

## Why This Change Was Made

Retire the duplicate direct `deep` row. The public case retains the identical 100,000-event fixture, full digest and selected-ancestry checks, 256 MiB child heap limit and existing deadline through the same importer. The batch case retains 256-session breadth and positive import totals. The direct deep-specific first-import total of 100,002 is no longer asserted; independent direct-import reporting tests remain.

## User Impact

No production behavior changes. One redundant test execution and two net support lines are removed.

## Evidence

- Linux Node 24.19: original memory suite 3/3 passed; candidate 2/2 passed with identical retained cases and fixture sizes. Reporting siblings: 242 passed, two existing platform-alias skips.
- Required mapped gate passed locally after remote validation hit a missing private baseline ref, then a 15-minute budget cap. No memory tests were repeated for that recovery. Independent P2 review passed; the owned Testbox was stopped.
- One measured file pair was 88.96s before and 88.31s after. The removed deep case took 18.76s, but retained cases ran slower. No substantial wall-time or overall test-runtime improvement is claimed.

### DIFF
--- a/src/commands/doctor-session-sqlite.memory.test-support.ts
+++ b/src/commands/doctor-session-sqlite.memory.test-support.ts
@@ -19,10 +19,8 @@ async function main() {
   process.env.OPENCLAW_STATE_DIR = stateDir;
   process.env.OPENCLAW_CONFIG_PATH = path.join(stateDir, "openclaw.json");
   const sessionCount = scenario === "batch" ? 256 : 1;
-  const eventCount =
-    scenario === "deep" || scenario === "public" ? 100_000 : scenario === "batch" ? 64 : 8;
-  const payloadBytes =
-    scenario === "deep" || scenario === "public" ? 4096 : scenario === "batch" ? 32768 : 256;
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0f9f40a9591d3ad44c5d455afaad8a54ef399627](https://github.com/openclaw/openclaw/pull/142614/commits/0f9f40a9591d3ad44c5d455afaad8a54ef399627)
- Landed commit: [62d1ef6a71d3fbbca61cb1af3c26f6e83cb949fa](https://github.com/openclaw/openclaw/commit/62d1ef6a71d3fbbca61cb1af3c26f6e83cb949fa)

PR #130494, opened 2026-08-26
fix(gateway): retain CLI metadata after history dedupe

jalehman:
## What Problem This Solves

Fixes #130456. Claude native-history rows that exactly matched local rows were deduplicated without transferring their import provenance, so `chat.history` lost the bound CLI session ID.

## Why This Change Was Made

The history merge owner now preserves validated native identity metadata on the surviving local row and reports incorporation correctly across paginated history. It does not synthesize IDs downstream, duplicate visible messages, or change persistence schemas.

Repeated-text matching preserves transcript-order identity assignment without assuming timestamps are chronological. A timestamp index retains eligible candidates behind future-dated rows and across timestamp-less fallbacks; consumed rows cannot be reused.

## User Impact

CLI-backed sessions retain native session identity after local/native history deduplication, so resume and history consumers can continue using the authoritative Claude session metadata.

## Evidence

- Reproduced the latest review counterexample against `f1f34f3a6cbf`: two local rows with timestamps `[600000, 0]` and one matching import incorrectly produced three visible rows. The repaired candidate returns two rows and preserves the imported identity on the matching survivor.
- Focused regressions cover nonchronological timestamps, repeated identities, order across candidate pools, timestamp-less fallback, exact-identity reservation, image deduplication, anchored metadata, and offset pagination.
- The 20,000-row repeated-text regression completed in approximately 0.54 seconds locally.
- Validation on `4b17af97b545be1811ae04ff3da6febb2105bfc0`: 231 focused CLI history/Gateway tests passed; required `check:changed` passed; formatting and diff checks passed.
- Repaired the exact-head ClawSweeper P2: an unmatched imported row no longer advances the synthetic role/text floor, so a later timestamp-consistent import can enrich the eligible local survivor instead of producing a duplicate third row.
- Mandatory independent autoreview completed through P2 on the frozen candidate. Its only finding was explicitly `INDEX-only` and confirmed the working-tree repair was correct; staging the reviewed files resolved it before commit.
- No live Claude subscription-backed trace was captured; validation uses the real history merge and Gateway test boundaries with synthetic transcripts.

PR #142618, opened 2026-09-08
fix(ui): settings search loses phrases with middle tag filters

steipete:
<details>

The branch is in `openclaw/openclaw` and is editable by repository maintainers.

</details>

## What Problem This Solves

Fixes an issue where users searching Settings would lose matching configuration fields when a `tag:` filter appeared between words. `Log File tag:storage` finds Logging, but `Log tag:storage File` incorrectly shows “No matching settings.”

## Why This Change Was Made

The query parser retained the separator before each removed tag, leaving two spaces inside the literal phrase. Remove the complete tag span and the redundant preliminary trim. Both Settings navigation search and configuration-field filtering use this owner; tag matching, deduplication, Unicode text, and ordinary literal whitespace retain their existing semantics. Production code decreases by one line.

## User Impact

Tag filters work before, within, or after search phrases. Clicking the result opens the matching field. The Settings guide now includes the example and explains that multiple tags must all match.

## Evidence

- Three regression assertions failed with the original parser. The final focused parser and Settings-search caller suites pass all 58 tests, including tag position, empty/tag-only queries, Unicode, duplicate tags, wrong tags, embedded nonoperator text, and literal double spaces.
- `node scripts/check-changed.mjs -- ui/src/components/config-form.search.ts ui/src/components/config-form.search.node.test.ts ui/src/e2e/chat-cli-history-redaction.e2e.test.ts docs/web/control-ui/settings.md` passed, including typechecks, formatting, i18n, guards, dead-export scans, and targeted lint.
- The first CI run exposed an unrelated stale CLI-history E2E assertion after main #130494 began retaining native identity on deduplicated messages. After refreshing the base, the untouched E2E reproduced the failure. Its assertion now checks the exact local message plus the expected native identity; the complete bundled Chromium E2E passes with all redaction, on-disk preservation, message-count, and rendering assertions retained. No runtime behavior changed for this maintenance correction.
- Independent review through P2 is clean. Current-main qualification found the affected owner, callers, label/tag producer, and dependencies unchanged.
- Real Chromium against the source UI and a synthetic mock Gateway reproduces the missing result before the fix and the matching result/click-through afterward. Deferred schema loading, an unmatched tag, and clearing search also complete. These captures contain synthetic data only; no live operator Gateway or provider traffic was used.

Before: the middle tag hides the matching setting.

Before: Settings search has no match for the middle-tag phrase

After: the same phrase finds Logging.

After: Settings search finds Logging with the middle-tag phrase

Clicking the result opens Log File Path.

The search result opens Log File Path with its storage and observability tags

### DIFF
--- a/docs/web/control-ui/settings.md
+++ b/docs/web/control-ui/settings.md
@@ -10,6 +10,8 @@ sidebarTitle: "Settings"
 
 Everything under Settings, plus the settings-owned pages the sidebar links to.
 
+Use **Search settings** to find pages and configuration fields. Add `tag:storage`, for example, to filter configuration fields by tag. Tags can appear before, within, or after a text phrase: `Log File tag:storage` and `Log tag:storage File` both find **Log File Path**. Multiple tags require a field to match every tag.
+
 Model menus with more than eight choices include search. Filter by model name or provider/model reference, then choose a result to apply it. Typing or dismissing the menu leaves the current selection unchanged. Short menus stay compact, and custom model entry remains available where the setting supports it.
 
 ## Environment identity
--- a/ui/src/components/config-form.search.node.test.ts
+++ b/ui/src/components/config-form.search.node.test.ts
@@ -24,10 +24,17 @@ const schema = {
 };
 
 describe("config form search", () => {
-  it("parses tag-prefixed query terms", () => {
-    const parsed = parseConfigSearchQuery("token tag:security tag:Auth");
-    expect(parsed.text).toBe("token");
-    expect(parsed.tags).toEqual(["security", "auth"]);
+  it.each([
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [192b73b5d73a97ec63e2f43af0e793919ae0d8b9](https://github.com/openclaw/openclaw/pull/142618/commits/192b73b5d73a97ec63e2f43af0e793919ae0d8b9)
- Landed commit: [ddbead2ea5e41c388b9f710b708154709ff817d2](https://github.com/openclaw/openclaw/commit/ddbead2ea5e41c388b9f710b708154709ff817d2)

PR #142621, opened 2026-09-08
fix(codex): tolerate transient native-home auth state on homeScope=user

ComfyChloe:
# PR — fix(codex): tolerate transient native-home auth state under `homeScope=user`

> **Linked issue:** <paste-issue-number-after-filing>
> **Suggested branch:** `fix/codex-native-home-auth-race`
> **Reviewer hint:** the harness already isolates Codex state per agent by default; this only changes the verify call inside `assertNativeCodexAccountMatchesRoute`, which is reached only via `authRequirement: "subscription" | "api-key"` on `homeScope: "user"` connections.

## What Problem This Solves

Fixes an issue where users with `appServer.homeScope="user"` (the explicit opt-in that shares the native `~/.codex/` Codex home between OpenClaw and the standalone `codex` CLI / Codex Desktop) intermittently see `Codex subscription route requires ChatGPT auth in the native Codex home...` on subscription-route turns, even though the native home contains a valid ChatGPT sign-in. The verify is a single `account/read` against a file the native codex CLI / Codex Desktop is allowed to rewrite (token refresh, workspace switch, account rotation, concurrent app-server spawns). A read that catches a transient state looks identical to a misconfiguration, and the error's suggested `codex login` does not address a race condition `codex login` cannot fix.

## Why This Change Was Made

One bounded retry through the codex app-server's own `refreshToken: true` path lets a transient token-refresh window settle before we treat the read as a configuration error. This mirrors what the standalone `codex` CLI already does on the same home. The change is contained to the one verify call site that emits the misleading error. There is no call-site signature change, no behavior change for the "no account after retry" or "wrong account type" cases, and no behavior change for the `homeScope="agent"` path (which never reaches this verifier).

A sharper error message and a docs note are tracked separately so this PR stays focused on the race-condition leg of the bug.

## User Impact

Users with `homeScope="user"` no longer see subscription-route turns fail under normal native-side auth churn. The path keeps returning `status: 401` for genuine misconfigurations (no account after retry / wrong account type) so the observable billing-boundary protection is unchanged. Agent homes (the default) are unaffected because they never enter this function.

## Evidence

- **Repro before the change:** enable `appServer.homeScope="user"` with a valid ChatGPT OAuth sign-in in `~/.codex/auth.json` (or `%USERPROFILE%\.codex\auth.json` on Windows), drive concurrent OpenClaw turns while Codex Desktop is also active, and observe the error appearing non-deterministically. A retry against the unchanged `auth.json` succeeds, confirming the read race.
- **Behavior after the change:** same workload; the verify retries once through the codex app-server's own refresh path before throwing. Genuine misconfig (no account / wrong type after refresh) still throws the original error verbatim with the same `status: 401` shape, so callers (`auth-bridge`, preflight, fallback-API-key resolution) and their tests are unaffected.
- **Tests:** `pnpm test:extension codex` (full codex extension suite) covers this path; the targeted suites (`attempt-startup`, `auth-binding`, `run-attempt.auth-context`) also pass on the same edit.
- **Source pointer:** `extensions/codex/src/app-server/auth-bridge.ts` (`assertNativeCodexAccountMatchesRoute`).

## Code change

Single file: `extensions/codex/src/app-server/auth-bridge.ts`.

The function `assertNativeCodexAccountMatchesRoute` (lines 721-753 on `main` commit `0965053`) changes from:

```ts
async function assertNativeCodexAccountMatchesRoute(
  client: CodexAppServerClient,
  authRequirement: CodexAppServerAuthRequirement | undefined,
  assertCurrent?: () => void,
): Promise<void> {
  if (!authRequirement) {
    return;
  }
  const response = await client.request<CodexGetAccountResponse>(
    "account/read",
    { refreshToken: false },
    { assertCurrent },
  );
  const accountType = isJsonObject(response.account) ? response.account.type : undefined;
  if (authRequirement === "subscription") {
    if (accountType !== "chatgpt") {
      throw createCodexAppServerAuthError(
        'Codex subscription route requires ChatGPT auth in the native Codex home. Run `codex login` for that home, or use appServer.homeScope="agent" with an OpenClaw OAuth profile, then retry.',
      );
    }
    return;
  }
  if (accountType === "chatgpt") {
    throw createCodexAppServerAuthError(
      'Codex Platform route requires an API-key account, but the native Codex home is signed in with a ChatGPT subscription. Sign that home in with `codex login --with-api-key`, or set appServer.homeScope="agent" so OpenClaw can inject its own key.',
    );
  }
}
```

…to:

```ts
async function assertNativeCodexAccountMatchesRoute(
  client: CodexAppServerClient,
  authRequirement: CodexAppServerAuthRequirement | undefined,
  assertCurrent?: () => void,
): Promise<void> {
  if (!authRequirement) {
    return;
  }
  // The native home is shared with the standalone codex CLI / Codex Desktop and
  // can be rewriting auth.json mid-flight (token refresh, workspace switch,
  // account rotation, concurrent app-server spawn). A single stale read should
  // not look like a configuration error; one bounded refresh through the
  // app-server gives a transient token-refresh window a chance to settle
  // before we throw, matching the read path the standalone codex CLI uses on
  // the same home.
  let response = await client.request<CodexGetAccountResponse>(
    "account/read",
    { refreshToken: false },
    { assertCurrent },
  );
  if (!isJsonObject(response.account)) {
    response = await client.request<CodexGetAccountResponse>(
      "account/read",
      { refreshToken: true },
      { assertCurrent },
    );
  }
  const accountType = isJsonObject(response.account) ? response.account.type : undefined;
  if (authRequirement === "subscription") {
    if (accountType !== "chatgpt") {
      throw createCodexAppServerAuthError(
        'Codex subscription route requires ChatGPT auth in the native Codex home. Run `codex login` for that home, or use appServer.homeScope="agent" with an OpenClaw OAuth profile, then retry.',
      );
    }
    return;
  }
  if (accountType === "chatgpt") {
    throw createCodexAppServerAuthError(
      'Codex Platform route requires an API-key account, but the native Codex home is signed in with a ChatGPT subscription. Sign that home in with `codex login --with-api-key`, or set appServer.homeScope="agent" so OpenClaw can inject its own key.',
    );
  }
}
```

Three diff hunks conceptually:

1. `const response` → `let response` on the existing `account/read` call (so the retry can re-assign).
2. Insert a 6-line comment block between the early-return `}` and the first read, explaining why we retry.
3. Insert a 7-line retry block (`if (!isJsonObject(response.account)) { ... }`) between the original read and the `accountType` check.

No other source files are touched. No tests are added because the existing `pnpm test:extension codex` lanes already exercise both the success path (`account` present and correctly typed) and the throw path (no account after refresh). A more thorough test could simulate a missing-then-present account across two `account/read` calls — happy to add that as a follow-up if reviewers want it.

## Out of scope (tracked separately)

- Improving the error message to distinguish "no account" vs "wrong account type" vs "transient refresh". Worth doing, but it changes user-visible strings and is easier to review on its own.
- Adding a `Known limitations under homeScope="user"` section to `docs/plugins/codex-harness-reference.md`. Documentation-only, also a separate change.


ComfyChloe:
## Live repro captured from a real install (OpenClaw 2026.9.2, Windows 11)

The PR was filed with source-level evidence only; here is the actual runtime repro.

### Environment

- **OpenClaw version:** 2026.9.2 (`cd8c748`) — local install on Windows 11
- **Codex plugin version:** bundled `codex` (matching 2026.9.2)
- **Native `~/.codex/auth.json`:** `auth_mode: "chatgpt"`, valid `access_token` (Plus plan, account_id `887c94b6-…`, id_token email `minersparks77@gmail.com`)
- **Codex runtime active:** `account/read` calls observable in gateway logs as recently as 21:55:45 UTC today (request_id `2926377e-…`); the bug surface is reachable

### Repro command

```bash
openclaw agent --agent main --model openai/gpt-5.6-luna \
  --message "Reply with exactly: PROBE_OK and nothing else."
```

### Live output (verbatim)

```
openclaw.cmd : Codex subscription route requires ChatGPT auth in the native Codex home. Run `codex login` for that home, or use appServer.homeScope="agent" with an OpenClaw OAuth profile, then retry.
```

### What this proves

- The harness path **is reached** on a real subscription-route model run (`gpt-5.6-luna` is one of the two ChatGPT-served GPT-5.6 models in `modelPolicy.allow`; the main agent pins `gpt-5.6-sol` to `agentRuntime: "openclaw"` and so doesn't hit this path, but the other two do)
- `assertNativeCodexAccountMatchesRoute` **does fire** on the first call against the existing valid ChatGPT auth.json — confirming the source-level analysis in the PR description (single-shot `account/read { refreshToken: false }` race against a shared native home)
- The error message is **verbatim** what the source code at `extensions/codex/src/app-server/auth-bridge.ts:736` produces — confirming the same code path is still vulnerable

### Reproduce rate

Captured once on the first try today (2026-09-08 23:27 UTC). A second attempt immediately after hit a cooldown rather than the same error (auth_profile_failure_state_updated `cooldown:auth`), which is the harness's own downstream consequence of the bug firing.

### Why the error fires here

The `~/.codex/auth.json` has `last_refresh: "2026-08-31T23:56:51Z"` and the access_token's JWT `exp` is `2026-09-08T23:50:22Z` — the access_token was about to expire when the turn ran, which is exactly the kind of native-side auth state churn that the verifier races against.

### Suggested next steps for review

- Run the PR's `pnpm test:extension codex` to verify all existing `toHaveBeenNthCalledWith(1, "account/read", ...)` and `toHaveBeenCalledOnce()` assertions still pass with the retry path added under the `subscription` case only
- Consider whether the docs section on `homeScope: "user"` should warn about the shared-home race; the existing wording focuses on the cross-client thread sharing motivation but doesn't flag this verifier behavior
- The sharper error-message improvement (`codex login` vs `codex login --with-api-key` based on actual failure mode) is intentionally out of scope here but is worth tracking separately

PR #142626, opened 2026-09-08
fix(imessage): restore feedback after bridge recovery

### LINKED ISSUE #142603: [Bug]: iMessage typing and read receipts stay disabled after bridge recovery
### Bug type

### Beta release blocker

No

### Summary

After the iMessage private bridge recovers, inbound messages can permanently stop sending typing indicators and read receipts because the monitor never refreshes an invalidated or expired capability snapshot.

### Steps to reproduce

1. Start the iMessage monitor with private API capabilities available.
2. Invalidate the cached private API status, matching the path taken when an RPC request detects a bridge stall before automatic `imsg launch` recovery.
3. Allow the bridge to become available again.
4. Deliver another inbound direct message.
5. Observe that the monitor reads only the missing cache entry and skips both `typing` and `read` RPCs.

### Expected behavior

The next inbound message re-probes a missing or expired private API capability snapshot and resumes supported typing and read receipt RPCs after the bridge is available again.

### Actual behavior

The inbound monitor calls `getCachedIMessagePrivateApiStatus()` only. Once the snapshot is invalidated, every later inbound message treats typing and read as unsupported until an unrelated action, explicit status probe, or gateway restart repopulates the cache.

### OpenClaw version

2026.9.1 and current main at 251634acaccd9d0576a915461f7a154cc51b50f4

### Operating system

macOS 26

### Install method

npm global for the live observation; pnpm source checkout for the deterministic regression test

### Model

Not model-dependent

### Provider / routing chain

Not provider-dependent; the failure occurs in iMessage inbound feedback before model dispatch

NOT_ENOUGH_INFO

### Logs

```shell
imsg basic RPC remained healthy while `imsg status --json` reported advanced features unavailable. After `imsg launch` and a gateway restart, `v2_ready`, `typing`, and `read` were available again.
```

Current main invalidates the capability snapshot and automatically runs `imsg launch` for a positively identified bridge stall, but the inbound monitor only reads the cache. A deterministic regression test reproduces the missing-cache state and proves that typing/read are skipped without a lazy re-probe.

### Impact and severity

Affected: iMessage users after a private bridge stall or transient unavailable state. Severity: annoying but persistent loss of native feedback. Frequency: reproduced in one live sleep/wake incident and deterministically from the cache lifecycle. Consequence: typing indicators and read receipts remain absent even after bridge recovery, making the agent appear unresponsive.

### Additional information

This is the remaining capability-refresh gap after #137026 and #137029. It does not retry a possibly completed mutation and does not launch Messages merely because a cold-start probe reports the private API unavailable.

omarshahine:
Closes #142603

## What Problem This Solves

Fixes an issue where iMessage users would permanently lose typing indicators and read receipts after the private bridge recovered from a stall or transient unavailable state.

## Why This Change Was Made

The existing recovery invalidates the stale private API capability snapshot and re-injects the bridge, but the inbound monitor only read the cache afterward. It now resolves capabilities through the existing lazy probe, which preserves valid cached results and only runs the read-only status probe when the snapshot is missing or expired.

This does not retry a possibly completed mutation and does not launch Messages merely because a cold-start status is unavailable.

## User Impact

Typing indicators and read receipts resume on the next inbound message after bridge recovery, without requiring an unrelated private API action, an explicit status probe, or a Gateway restart.

## Evidence

- `pnpm exec vitest run extensions/imessage/src/monitor.last-route.test.ts`: 44 passed
- `pnpm check:changed`: passed, including extension production and test typechecks, lint, formatting, boundaries, and contract guards
- Independent autoreview: scoped clean through P2, with the cache fast path, read-only probe boundary, and regression fidelity verified
- Added a deterministic regression that starts with a missing capability snapshot, resolves the recovered typing/read surface, and proves both RPCs are sent

### DIFF
--- a/extensions/imessage/src/monitor.last-route.test.ts
+++ b/extensions/imessage/src/monitor.last-route.test.ts
@@ -38,6 +38,7 @@ import {
   getCachedIMessagePrivateApiStatus,
   setCachedIMessagePrivateApiStatus,
 } from "./private-api-status.js";
+import type { probeIMessagePrivateApi } from "./probe.js";
 import { installIMessageStateRuntimeForTest } from "./test-support/runtime.js";
 
 const DEFAULT_SENDER = "+15550001111";
@@ -159,6 +160,7 @@ const waitForTransportReadyMock = vi.hoisted(() =>
   vi.fn<typeof waitForTransportReady>(async () => {}),
 );
 const createIMessageRpcClientMock = vi.hoisted(() => vi.fn<typeof createIMessageRpcClient>());
+const probeIMessagePrivateApiMock = vi.hoisted(() => vi.fn<typeof probeIMessagePrivateApi>());
 const readChannelAllowFromStoreMock = vi.hoisted(() => vi.fn(async () => [] as string[]));
 const ensureConfiguredBindingRouteReadyMock = vi.hoisted(() =>
   vi.fn<typeof ensureConfiguredBindingRouteReady>(async () => ({ ok: true })),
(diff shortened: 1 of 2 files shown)


omarshahine:
@clawsweeper review

omarshahine:
@clawsweeper status

omarshahine:
@clawsweeper automerge

omarshahine:
@clawsweeper stop

Issue #138760, opened 2026-09-05
[Bug]: Git update leaves Gateway stopped after loading a removed restart bundle


jason-allen-oneal:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

A Git update installed and verified the new version but left the managed Gateway stopped because the old updater attempted to load a generated JavaScript module that the rebuild had removed.

### Steps to reproduce

Observed sequence on a Linux Git installation with a systemd user Gateway:

1. Run the Git update from the installed checkout. The recorded transition was `c6f6055a527e91c5d17b166b81d0023581318526` (2026.8.1) to `0229a108fee749327b2cc60c4d228299dfe2f889` (2026.9.1).
2. The updater stops the managed Gateway, rebuilds the checkout, and completes Doctor.
3. Finalization preserves the existing service definition and attempts to launch the candidate restart command.
4. The old updater requests a removed hashed namespace module before spawning that command.

### Expected behavior

The updater launches the installed candidate and verifies Gateway readiness after the update, using the execution closure loaded before replacement.

### Actual behavior

The restart command is never launched. The updater reports `restart-unhealthy` and `runtime-verification-failed`; systemd shows the Gateway still stopped after a clean shutdown. These result labels describe the failed finalization stage, not an observed failed candidate health probe.

### OpenClaw version

2026.8.1 `c6f6055a527e91c5d17b166b81d0023581318526` updating to 2026.9.1 `0229a108fee749327b2cc60c4d228299dfe2f889`; the delayed imports also remain at main `dcdbd055c7b4d2c96b7ca96adf6b9ab6fd9ef617`.

### Operating system

Linux x64, kernel 7.0.12+kali-amd64; Node.js 24.19.0.

### Install method

Git checkout with a systemd user Gateway service.

### Model

Not applicable; failure occurs in update/service activation before model execution.

### Provider / routing chain

Not applicable.

### Logs

Sanitized excerpt from the observed update:

```text
Stopping managed gateway service before git update...
✓ Running doctor checks (244.67s)
✓ Verifying update (8ms)
Gateway service definition left unchanged
Restarting service...
Gateway: restart failed: Error: ENOENT: no such file or directory,
open '<install>/dist/shared-VSNmp8jC.js'
Update Result: ERROR
Reason: restart-unhealthy
```

### Screenshots, recordings, and evidence

[...]

PR #142631, opened 2026-09-08
fix(update): restart Git installs upgrading from 2026.9.1

jason-allen-oneal:
Related: #138760

## What Problem This Solves

A Git install running the exact 2026.9.1 source build (`0229a108`) can stop its Gateway, replace `dist/`, and then fail before spawning the restart command. The still-running old updater lazily imports `shared-1Uyqkfns.js`, but newer builds did not preserve that Git-build-only chunk name, so activation ends with `ERR_MODULE_NOT_FOUND` and `restart-unhealthy` even after the candidate passes Doctor.

## Why This Change Was Made

- Preserve `shared-1Uyqkfns.js` with the existing minimal `resolveNodeRunner` compatibility implementation.
- Keep this exact source-build hash in the manual compatibility list rather than the npm release inventory.
- Reproduce the real order in a regression test: load the old updater owner, replace `dist/`, write candidate compatibility output, then invoke the old lazy restart import.

## User Impact

Git installs upgrading from that 2026.9.1 build can activate the new Gateway without requiring a manual service start. Configuration, migrated state, history, and databases are unaffected by this compatibility file.

## Evidence

The observed 2026.9.1 to 2026.9.3 Git update stopped the Gateway, completed the candidate build and Doctor migration, then failed before launching the restart command with `ENOENT` for `dist/shared-1Uyqkfns.js`. Starting that same verified candidate restored Gateway health and authenticated RPC without rollback or database repair.

Before the production change, the lifecycle regression failed with `ERR_MODULE_NOT_FOUND` for that exact chunk. Afterward, `test/scripts/runtime-postbuild.test.ts` passes 72/72 tests. Formatting and `git diff --check` also pass; these checks are supplemental to the native behavior proof below.

## Real behavior proof

- **Claim:** The actual emitted 2026.9.1 Git updater, loaded before `dist/` replacement, can use the patched compatibility bridge to launch the candidate's Gateway restart process afterward.
- **Exercised surface:** Real emitted `runUpdatedInstallGatewayCommand()` from exact historical source commit `0229a108fee749327b2cc60c4d228299dfe2f889`; production `writeLegacyCliExitCompatChunks()` from exact PR head `9ed614258538a27665b57429b414c3a97b592ef6`; native Node ESM resolution, filesystem replacement, and a real child process. No module or loader mocks were used.
- **Scenario / fixture:** A disposable copy of the exact 2026.9.1 generated `dist/` and its installed dependencies was loaded natively. The proof located and imported the real emitted activation owner, deleted the entire copied `dist/`, installed a recording candidate entrypoint, ran the patched production postbuild writer, and invoked the already-loaded old command without injecting a Node runner.
- **Command / environment:** Linux, Node 24.19.0: `OPENCLAW_PROOF_OLD_ROOT=<exact-0229a108-checkout> node --import ./scripts/tsx.mjs /tmp/openclaw-update-bridge-proof.mjs`, run from the exact PR checkout with isolated temporary state and no live service selectors.
- **Observable result:** The patch recreated `dist/shared-1Uyqkfns.js`; the real old emitted owner resolved it after replacement, selected `/usr/bin/node`, spawned the candidate, and delivered `gateway restart --preserve-definition --json`. The candidate receipt matched exactly and the proof exited 0.
- **Artifact / trace:** Sanitized copied terminal output:

```text
OLD_2026_9_1_EMITTED_OWNER_LOADED
DIST_REPLACED_WITH_CANDIDATE
PATCHED_COMPAT_BRIDGE_WRITTEN shared-1Uyqkfns.js
POST_SWAP_REAL_CANDIDATE_PROCESS_OK
NODE_RUNNER /usr/bin/node
ARGV gateway restart --preserve-definition --json
```

- **Limits:** The candidate is a recording Node entrypoint, so this proof does not invoke systemd/launchd or install the unreleased patch into the live operator checkout. It proves the real old emitted owner, patched production bridge, native loader, and candidate subprocess boundary in the original order on Linux. The original machine incident supplies the end-to-end service failure/recovery trace; patched activation stays isolated to preserve the restored Gateway and state.


steipete:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [9ed614258538a27665b57429b414c3a97b592ef6](https://github.com/openclaw/openclaw/pull/142631/commits/9ed614258538a27665b57429b414c3a97b592ef6)
- Landed commit: [352b372c0c32977b70dbbdb2a64407a0ec75dc9d](https://github.com/openclaw/openclaw/commit/352b372c0c32977b70dbbdb2a64407a0ec75dc9d)

PR #142632, opened 2026-09-08
fix(mcp): drain admitted tool work before shutdown cleanup

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where MCP tool-server shutdown could release resources while canceled tool calls or their cooperating cleanup were still using them. Native SQLite queries could then fail with a closed-database error. A failed stdio connection also skipped the shutdown callback.

## Why This Change Was Made

The factory's server now joins admitted handler work after closing the transport. Concurrent close calls share that completion, and the server remains reusable after an awaited close. The stdio connector performs shutdown after connection failure and preserves the original connection error.

## User Impact

Tool bridges finish accepted cleanup before their shutdown owner releases dependencies. Transport cancellation remains prompt. External force-kills can still interrupt cleanup.

## Evidence

- Four native SDK/SQLite cases pass, including both reproduced early-close failures, cancellation, and reconnect controls.
- The exact changed-file gate passed in 344 seconds; the ordinary build passed in 244 seconds.
- Compiled stdio proof ran real JSON-RPC initialize/list/call, sent SIGTERM, and verified the final SQLite read completed before shutdown closed the database. The child exited successfully.
- Compiled connection-failure proof preserved the actual SDK error, joined native cleanup, and restored process listeners. A separate control preserved return-after-connect behavior without a shutdown callback.
- The optional control initially sampled before signal delivery; its fixture was corrected to await the actual signal, and only that control was rerun. All three accepted cases use the same production and compiled artifacts.
- Independent Codex review found no actionable findings. Both source files remain identical after rebasing onto current main.

### DIFF
--- a/src/mcp/plugin-tools-handlers.cancel.test.ts
+++ b/src/mcp/plugin-tools-handlers.cancel.test.ts
@@ -1,9 +1,14 @@
 // Plugin MCP cancellation tests cover cancellation of in-flight plugin tool calls.
+import { DatabaseSync } from "node:sqlite";
+import { setImmediate as nextTurn } from "node:timers/promises";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
+import { Type } from "typebox";
 import { describe, expect, it } from "vitest";
+import { createDeferred } from "../../test/helpers/promise.js";
 import { consumeTrackedToolExecutionStarted } from "../agents/agent-tools.before-tool-call.state.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4058ec385bb2f9281246ae89ce73f73f2257b4a4](https://github.com/openclaw/openclaw/pull/142632/commits/4058ec385bb2f9281246ae89ce73f73f2257b4a4)
- Landed commit: [c5c3748d8bd16022639e65c1d105de1f7cad1bd1](https://github.com/openclaw/openclaw/commit/c5c3748d8bd16022639e65c1d105de1f7cad1bd1)

PR #142636, opened 2026-09-08
fix(plugins): retain registration work through final resource cleanup

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where plugin resource cleanup captured during registration could inherit a caller's already-closed work context. A later resource owner could retain the database but still fail to run its disposer. Registration abort work could also outlive the database it was cleaning up.

## Why This Change Was Made

Owned registration now has its own work context through final disposal. Construction and abort work settle before physical cleanup starts; rollback preserves successful siblings, and the final resource claim joins actual cleanup. The existing work-scope helper handles synchronous registration without inspecting its return value.

## User Impact

Releasing an inspection still retires its authority immediately. Retained resources remain usable by their existing owners, and cleanup can finish in its original registration context before the last claim closes it. Independent cleanup remains concurrent; work explicitly owned by a sibling retains that sibling until it finishes.

## Evidence

- Native regressions reproduced closed captured contexts, cross-registration abort work, and rollback while a successful sibling remained active. All 31 final native lifecycle cases pass; nine shared-work and 59 loader controls also passed on the same production.
- The final six-file changed gate passed in 227 seconds. The ordinary build passed in 205 seconds; the only later source change added native test coverage and consolidated its fixture setup.
- Eight isolated compiled scenarios passed using the actual built loader, real registration claims, native SQLite, and readonly reopen checks. These cover retained context/tracker cleanup, disposal errors, ordinary and raw controls, cross-registration aborts, rollback, and explicit sibling dependencies. All 13,707 build artifacts remained unchanged.
- Independent Codex review completed. Reproduced defects were repaired; follow-up dependency reports were adjudicated against real callers and native/compiled evidence. A captured tracker owns its captured scope, so callers must also await or track work that uses their own resources; the implementation does not infer dependencies from arbitrary closures.
- All six reviewed source files are unchanged after integration with main, including the already-landed CLI work-scope helper.

### DIFF
--- a/src/plugins/loader-module-runtime.ts
+++ b/src/plugins/loader-module-runtime.ts
@@ -114,7 +114,15 @@ export function runPluginRegisterSyncInRegistry(
   withPluginRegistrationContext(
     registry,
     pluginId,
-    () => runPluginRegisterSync(register, api, registry),
+    () => {
+      const inspection = getPluginRegistryInspectionResources(registry);
+      const run = () => runPluginRegisterSync(register, api, registry);
+      if (inspection) {
+        inspection.runRegistration(pluginId, run);
+      } else {
+        run();
+      }
+    },
     {
       registerMemoryCapability: api.registerMemoryCapability,
     },
--- a/src/plugins/registry-inspection-resources.ts
+++ b/src/plugins/registry-inspection-resources.ts
@@ -38,6 +38,10 @@ export class PluginRegistryInspectionResources {
     this.#source.register(pluginId, disposer);
   }
 
+  runRegistration(pluginId: string, run: () => void): void {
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5e62d99f691555368bded1f8b71670b5da5f50ab](https://github.com/openclaw/openclaw/pull/142636/commits/5e62d99f691555368bded1f8b71670b5da5f50ab)
- Landed commit: [d40a54a430bed0eeb5dd8995230805a93ff20c6d](https://github.com/openclaw/openclaw/commit/d40a54a430bed0eeb5dd8995230805a93ff20c6d)

PR #142646, opened 2026-09-08
fix(ui): stop guessing model thinking capabilities

obviyus:
## What Problem This Solves

Fixes an issue where chat status and session editors invented thinking options or advertised a default when the selected model offered no levels. New Session could also replace a published profile with a guessed default.

## Why This Change Was Made

Omit advertised defaults for empty choice lists and read one compatible published profile across status, command arguments, session editing and draft controls. Remove model-name and Medium guesses while preserving saved overrides and unknown-support validation.

## User Impact

Empty choices remain empty. Thinking status shows Unknown for a missing current/default level and none for an explicitly empty list.

Consumers: chat, Sessions and New Session thinking controls use the same published profile.

## Evidence

A real Gateway/browser reproduced the incorrect default. The corrected synthetic empty-profile case showed Unknown/none, clear/inherit-only session editing and No effort without a slider. Public readback preserved the absent override. Thirty-eight UI and nine Gateway controls passed; external/native authentication was not tested.


obviyus:
landpr-approve-admin-merge
pr: 142646
head: ce3fa4b06f8ef7f2823343cd84cd03feff26cc2c
scope: owner-review

PR #139892, opened 2026-09-06
fix: preserve Markdown literals in replies containing tables

steipete:
<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Fixes an issue where users reading replies with Markdown tables would see escaped literals or inline code change meaning when the table was converted to a text fallback. For example, adding a table after ````Keep \*stars\* and ``a`b`` literal.```` could turn the stars into emphasis and break the embedded-backtick code span. Headings and blockquotes elsewhere in the same reply could also lose their formatting.

## Why This Change Was Made

The converter now replaces only the table ranges recognized by the existing MarkdownIt parser, preserving the original source everywhere else and retaining cell Markdown for bullet output. Native IR and Markdown fallbacks share table layout and newline emission, including the existing behavior for empty rows; the whole-message Markdown roundtrip and duplicate layout paths are removed.

## User Impact

Replies containing tables retain their surrounding formatting, escaped literals and inline code. Bullet tables preserve authored cell markup and document reference links, and tables remain inside their original quote or list containers. Code-table alignment and the block-mode code fallback keep their existing behavior.

A separate pre-existing issue remains: unescaped standalone emphasis/backtick delimiters in cells can bind to generated bullet-formatting markers. That requires its own inline delimiter-ownership repair; authored escapes are preserved by this change.

## Evidence

- Captured the original failure before editing: 9 failing converter regressions and 3 passing controls.
- Final focused verification: **138 tests passed** across the converter, native bullet/code/block table renderers, and Discord formatter/chunker. Five row-spacing cases cover ordinary, empty-middle, empty-edge, all-empty and labeled rows; six standalone comparisons confirm that the complete native IR and converted Markdown match the original spacing behavior.
- Scoped typed lint passed for all four changed files. Production code is **175 lines added / 179 removed (net −4)**; regression coverage adds 132 test lines.
- Exercised the actual `renderDiscordMarkdown` caller locally with synthetic Markdown in `code`, `bullets` and `block` modes, then parsed the result with MarkdownIt. Escaped literals, multi-backtick code and reference links remained intact. Additional source-range checks covered nested quotes/lists, tabs, CRLF/bare CR, header-only tables and MarkdownIt's existing list-like-header recognition.
- Independent Codex reviews of the row-spacing repair and the subsequent type-only correction are **scoped-clean through P2**, with no actionable findings.
[...]

PR #142649, opened 2026-09-08
fix(markdown): own generated table delimiters against bare cell markdown

### LINKED ISSUE #142647: Markdown tables: bare cell delimiters pair across generated row-label markers in bullets mode
## Summary

In `bullets` table-rendering mode, the raw cell markdown is concatenated next
to generated `**` row-label markers without ownership. A bare `*` / `_` /
backtick inside a cell can pair across the cell boundary — consuming the
generated marker itself — so the rendered bullets differ from the table
content and can leak raw markdown into chat output.

## Reproduction

With a channel whose `defaultMarkdownTableMode` is `bullets` (WhatsApp and
Signal today; Telegram/Matrix use `block`, Mattermost `off`), a message
containing this table:

```
| tick | more |
| --- | --- |
| `tick | more` | 2 |
```

renders as:

- the row label becomes `<code>tick**` — the cell's opening backtick consumed
  the generated `**` closer, so the rest of the row renders inside a code span
- conversely an isolated `*` in one cell can close against emphasis opened in
  another cell (`| a * | b * |` renders `b *` with a stray `<em>`)

Verified on main @ `ecbad71d09b` with `convertMarkdownTables(input, "bullets")`.

## Link

- `packages/markdown-core/src/tables.ts:32` (main @ `ecbad71d09b`): the write-cell
  callback emits `` `**${cell.markdown}**` `` — `cell.markdown` is the raw cell
  source (`tables.ts:20` / `tables.ts:22` feed `source.headers[column]` and
  `source.rows[index]?.[column]` verbatim).
- The same contract is honored in `code`/`block` modes, which fence the payload
  or emit raw rows and are unaffected.

## Impact

Table output in bullets mode can render wrong text, swallow the row-label
delimiters, or emit stray `<em>`/`<code>` spans into the chat message. It is a
rendering-fidelity defect, not a security issue; the model's table content is
silently misrepresented to the user.

## Suggested fix

Before concatenation, escape bare `*`/`_`/`` ` `` delimiters in each cell's
markdown that are not inside a closed code span (closed spans must keep their
delimiters so inline code keeps rendering). The repository already has the
needed scanner: `buildCodeSpanIndex` in `packages/markdown-core/src/code-spans.ts`.
We have a patch with red/green regression tests ready and will open the PR
referencing this issue.

wangmiao0668000666:
## What Problem This Solves

Fixes an issue where users chatting through channels that render tables with
`convertMarkdownTables` in `bullets` mode (Discord, Feishu, IRC, Mattermost,
and others; resolved per channel config) would see table content silently
misrepresented when a cell contained a stray `*`, `_`, or backtick: the bare
delimiter could pair with the generated `**` row-label markers or with a
delimiter in a neighboring cell, swallowing the row boundary, leaking later
cells into a code span, or emitting stray `<em>` markup.

## Why This Change Was Made

The bullets formatter is the only table surface that mixes raw cell markdown
with generated markup without separating ownership: the generated `**` belongs
to the formatter, while delimiters inside `cell.markdown` belong to the cell
author. Inline markdown-it parsing runs over the joined string, so an unpaired
cell delimiter can pair with (or consume) the generated marker.

The fix lets the inline parser decide which delimiters are literal. Each cell
is parsed with `fromMarkdown` (an existing dependency), and only delimiters
inside parsed text nodes — the parser's literal-text regions — are escaped
with a backslash. Delimiters that belong to authored emphasis, inline code, or
links live in non-text nodes and survive untouched, so valid markup keeps
rendering while strays stay confined to their own cell. Escape-aware scanning
keeps authored backslash pairs (`\*`, `` \` ``, `\\`) verbatim.

Reference links are the one exception, and they are handled by non-
intervention rather than approximation: the renderer matches reference labels
against document definitions by identifier, an escaped delimiter inside a
label breaks that match, and every local approximation of the reference
grammar (scanners, definition slicing, label prechecks) leaks another edge
case — label folding, containers, multiline labels, escaped brackets. So a
cell that could contain a reference — any `[` while the document holds a
definition shape (`]:`, which every reference definition contains) — passes
through byte-identical to main. Cells without `[`, or any cell when the
document has no `]:`, are provably reference-free and get full delimiter
isolation. The stray-delimiter defect stays open in possible-reference cells
by design; that intersection is rare and this PR chooses zero reference
regressions over covering it. Angle-bracket autolinks get the same protection
structurally: their parsed text child *is* the destination, so autolink
subtrees (the only link form starting with `<`) are exempt from escaping.

**Fix classification: behavior-completing** — #139892 (merged as
`5d1a4193ef8`) preserved table literals by routing only parsed table ranges
through this formatter, but the formatter itself still mixes generated and
authored delimiters. This PR completes that boundary at the cell level.

Related: #139892, fixes #142647

## User Impact

Table output in bullets mode now renders as authored: deliberate `**bold**`,
`_italic_`, and `` `code` `` keep their formatting, escaped characters stay
literal, document reference links keep resolving exactly as on main (their
cells are byte-identical), and a stray delimiter in one cell can no longer
corrupt the rendering of later cells or swallow the generated bullet markers.
One disclosure: a cell delimiter that previously paired across the cell
boundary (misrepresenting the table) now renders literally at its own
position, and the underlying bytes gain a `\`; rendered glyphs are otherwise
unchanged, and cells with valid markup are byte-identical.

## Evidence

Reproduction through the real Discord outbound formatter
(`renderDiscordMarkdown`, extensions/discord/src/markdown.ts:324 — the
`convertMarkdownTables` call at outbound-text.ts:19), run under the
extension-discord vitest project so `openclaw/plugin-sdk/*` resolves exactly
like production. BEFORE is main itself (no cell delimiter isolation), so main
already renders valid emphasis, code spans, and reference links correctly;
the defect is the stray-delimiter rows:

| Input | BEFORE (main) | AFTER (this branch) |
|---|---|---|
| `a *` and `b *` in adjacent cells | stray `<em>` spans cells | each `*` stays literal |
| `` `tick `` in row label | code span swallows the row boundary | row label stays bold, backtick literal |
| `**bold** and _italic_` | renders correctly on main | still renders `<strong>`/`<em>` (no regression) |
| `` \` literal and `code` `` | renders correctly on main | still renders literal backtick + `<code>` (no regression) |
| `[manual][api_v1]` + `[api_v1]:` definition | renders correctly on main | byte-identical cell; still resolves `<a>` |
| `[manual][api\]_v1]` + escaped-bracket definition | renders correctly on main | byte-identical cell; still resolves `<a>` |
| `[manual][api_ v1]` + continued block-quote definition | renders correctly on main | byte-identical cell; still resolves `<a>` |
| `<first_last@example.com>` / `<https://example.com/a_b>` | autolinks resolve on main | destinations untouched; autolinks still resolve |

Red/green proof: the reference pass-through cases (continued
block-quote definition, escaped-bracket label) fail on revision
`f9ac6ba0134` exactly on their assertions, and the autolink destination case
fails on `982ad77c54a` (escaped underscore kills the email/URL autolink);
all pass on this head. Full
`tables.test.ts` passes 36/36 and the markdown-core suite passes 27 files /
396 tests; `extensions/discord/src/markdown.test.ts` passes 2/2. All test
runs executed in an isolated Docker environment on a remote test host;
`oxfmt` clean (re-verified 2026-09-09 UTC after the pass-through reslice and
the autolink guard).

<details>
<summary>Details maintainers may want to adjust</summary>

- The issue text, reproduction rows, and classification wording can be edited
  freely to match house style.
- The regression cases live in the existing `convertMarkdownTables`
  `it.each` block neighborhood; maintainers may prefer them split or renamed.
- If maintainers prefer ownership handled further upstream (at table parse
  time), the same escape can move there — this patch keeps it at the formatter
  because `code`/`block` modes must stay byte-faithful.

</details>

<details>
<summary>Alternatives considered / land-ready rationale</summary>

- **Reject and close** as `wontfix` if bullets-mode fidelity is considered
  acceptable drift — but the converter is the bullets fallback for a dozen
  channels (discord, feishu, irc, matrix, mattermost, msteams, nextcloud-talk,
  nostr, zalo, and more), so real chat output hits this.
- **Behavior-altering alternative**: render row labels as plain text instead
  of `**` — rejected, it changes deliberate output for all tables.
- **Escape everything outside closed code spans** (first revision): rejected
  after review — it degraded authored emphasis and broke escaped-backtick +
  code-span cells. The parser-owned text-node rule replaces it.
- **Reconstruct reference context locally** (revisions 2–5): rejected after
  five review rounds — scanning, definition slicing, and label prechecks each
  leaked another renderer-grammar edge case (containers, Unicode folds,
  multiline labels, escaped brackets). Byte-identical pass-through for
  possible-reference cells removes the whole regression class instead of
  approximating one more corner of it.
- **Parse-time ownership**: escape at table parse instead of render — kept out
  of scope; render-time is the only place that knows `code`/`block` modes do
  not need it.
- This patch is land-ready: no config/schema changes, no dependency changes,
  one file plus tests, `oxfmt` clean, and it reuses the existing
  `fromMarkdown` dependency rather than adding a new scanner.

</details>

### DIFF
--- a/packages/markdown-core/src/tables.test.ts
+++ b/packages/markdown-core/src/tables.test.ts
@@ -315,6 +315,21 @@ describe("convertMarkdownTables", () => {
     expect(rendered).not.toContain("\\*");
   });
 
+  it("preserves angle-bracket autolink destinations in bullet cells", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | <first_last@example.com> |\n| Other | <https://example.com/a_b> |",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // The autolink text is the destination; escaping its underscore breaks
+    // markdown-it's autolink grammar and kills the link.
+    expect(rendered).toContain("<first_last@example.com>");
+    expect(rendered).toContain("<https://example.com/a_b>");
+    expect(parsed).toContain('<a href="mailto:first_last@example.com">');
+    expect(parsed).toContain('<a href="https://example.com/a_b">');
+  });
+
   it("preserves CRLF source around a table", () => {
     const before = "Keep \\*literal\\*.\r\n\r\n";
     const after = "\r\n\r\nAfter.\r\n";
--- a/packages/markdown-core/src/tables.test.ts
+++ b/packages/markdown-core/src/tables.test.ts
@@ -189,6 +189,26 @@ describe("convertMarkdownTables", () => {
     expect(rendered.endsWith(references)).toBe(true);
   });
 
+  it("preserves delimiter-bearing reference labels against document definitions in bullets", () => {
+    const references = "\n\n[api_v1]: https://example.com";
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [manual][api_v1] |" + references,
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // The reference label keeps its underscore so the definition still matches.
+    expect(rendered).toContain("[manual][api_v1]");
+    expect(parsed).toContain('<a href="https://example.com">manual</a>');
+    // Bracket text without a matching definition is still delimiter-isolated.
+    const unlabeled = convertMarkdownTables(
+      "| A | B |\n| --- | --- |\n| x | [a_b] stays literal |",
+      "bullets",
+    );
+
+    expect(unlabeled).toContain("[a\\_b]");
+  });
+
   it("preserves CRLF source around a table", () => {
     const before = "Keep \\*literal\\*.\r\n\r\n";
     const after = "\r\n\r\nAfter.\r\n";
--- a/packages/markdown-core/src/tables.test.ts
+++ b/packages/markdown-core/src/tables.test.ts
@@ -235,6 +235,47 @@ describe("convertMarkdownTables", () => {
     expect(parsed).toContain('<a href="https://example.com">manual</a>');
   });
 
+  it("preserves references whose definition label spans lines", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [manual][api_ v1] |\n\n[api_\nv1]: https://example.com",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // The renderer folds the definition's newline to a space, so the reference
+    // resolves; the label must keep its underscore for the link to survive.
+    expect(rendered).toContain("[manual][api_ v1]");
+    expect(parsed).toContain('<a href="https://example.com">manual</a>');
+  });
+
+  it("preserves shortcut references with whitespace-padded labels", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [ api_v1 ] |\n\n[api_v1]: https://example.com",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // The renderer trims labels during lookup, so the padded shortcut is a
+    // valid reference and its underscore must not be escaped.
+    expect(rendered).toContain("[ api_v1 ]");
+    expect(parsed).toContain('<a href="https://example.com">');
+  });
+
+  it("preserves references that differ from definitions by Unicode whitespace", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [manual][api_ v1] |\n| Other | [guide][docs_\u00a0v2] |\n\n[api_\u00a0v1]: https://example.com/api\n\n[docs_ v2]: https://example.com/docs",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // markdown-it folds identifiers with Unicode `\s`, so a nonbreaking space in
+    // either the reference or the definition still matches an ordinary space.
+    expect(rendered).toContain("[manual][api_ v1]");
+    expect(rendered).toContain("[guide][docs_\u00a0v2]");
+    expect(parsed).toContain('<a href="https://example.com/api">manual</a>');
+    expect(parsed).toContain('<a href="https://example.com/docs">guide</a>');
+  });
+
   it("preserves CRLF source around a table", () => {
     const before = "Keep \\*literal\\*.\r\n\r\n";
     const after = "\r\n\r\nAfter.\r\n";
--- a/packages/markdown-core/src/tables.test.ts
+++ b/packages/markdown-core/src/tables.test.ts
@@ -81,6 +81,40 @@ describe("convertMarkdownTables", () => {
     expect(html).toContain(mode === "bullets" ? "<code>x`y</code>" : "x`y");
   });
 
+  it.each([
+    {
+      name: "row label stray delimiter binds through the generated bold markers",
+      input: "| Name | Value |\n| --- | --- |\n| `tick | more` |",
+      // A bare label backtick must not own the generated `**`; otherwise the
+      // code span swallows the bullet line (`**<code>tick** • V: more</code>`).
+      notHtml: ["<code>tick**"],
+      html: ["• Value: more`"],
+    },
+    {
+      name: "value cell stray delimiters bind across the bullet line",
+      input: "| A | B |\n| --- | --- |\n| a * | b * |",
+      // Bare stars must stay literal instead of pairing across the bullet line
+      // boundary (`*<em>a ***\n• V: b</em>` once emphasis misbinds).
+      notHtml: ["<em>"],
+      html: ["• B: b *"],
+    },
+  ])(
+    "keeps unescaped stray cell delimiters literal in bullet output: $name",
+    ({ input, notHtml, html }) => {
+      const rendered = convertMarkdownTables(input, "bullets");
+      const parsed = new MarkdownIt().render(rendered);
+
+      for (const forbidden of notHtml) {
+        expect(parsed).not.toContain(forbidden);
+      }
+      for (const literal of html) {
+        expect(parsed).toContain(literal);
+      }
+      // Row labels stay bold and no delimiter leaks outside its own cell text.
+      expect(parsed).toContain("<strong>");
+    },
+  );
+
   it("chooses a code fence that contains literal backtick runs", () => {
     const rendered = convertMarkdownTables(
       "| A | B |\n| --- | --- |\n| 1 | ````a```b```` |",
--- a/packages/markdown-core/src/tables.test.ts
+++ b/packages/markdown-core/src/tables.test.ts
@@ -209,6 +209,32 @@ describe("convertMarkdownTables", () => {
     expect(unlabeled).toContain("[a\\_b]");
   });
 
+  it("preserves reference labels whose only definition lives in a block quote", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [manual][api_v1] |\n\n> [api_v1]: https://example.com",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // The block-quoted definition still backs the reference, so the label must
+    // keep its underscore for the link to resolve.
+    expect(rendered).toContain("[manual][api_v1]");
+    expect(parsed).toContain('<a href="https://example.com">manual</a>');
+  });
+
+  it("preserves reference labels that differ from definitions by Unicode case folding", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [manual][straße_v1] |\n\n[straße_v1]: https://example.com",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // micromark folds identifiers with lower+uppercase (ß → SS), so matching
+    // must use the same fold or the underscore gets escaped and the link dies.
+    expect(rendered).toContain("[manual][straße_v1]");
+    expect(parsed).toContain('<a href="https://example.com">manual</a>');
+  });
+
   it("preserves CRLF source around a table", () => {
     const before = "Keep \\*literal\\*.\r\n\r\n";
     const after = "\r\n\r\nAfter.\r\n";
--- a/packages/markdown-core/src/tables.test.ts
+++ b/packages/markdown-core/src/tables.test.ts
@@ -276,6 +276,45 @@ describe("convertMarkdownTables", () => {
     expect(parsed).toContain('<a href="https://example.com/docs">guide</a>');
   });
 
+  it("preserves references backed by a continued block-quote definition", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [manual][api_ v1] |\n\n> [api_\n> v1]: https://example.com",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // The definition's continuation lines belong to the block quote; the cell
+    // passes through untouched so the renderer's own label folding resolves it.
+    expect(rendered).toContain("[manual][api_ v1]");
+    expect(parsed).toContain('<a href="https://example.com">manual</a>');
+  });
+
+  it("preserves references whose label contains an escaped closing bracket", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | [manual][api\\]_v1] |\n\n[api\\]_v1]: https://example.com",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    // The renderer skips the escaped `]` while reading the label; the cell must
+    // keep its exact source bytes for the definition match to survive.
+    expect(rendered).toContain("[manual][api\\]_v1]");
+    expect(parsed).toContain('<a href="https://example.com">manual</a>');
+  });
+
+  it("passes bracket-bearing cells through untouched while definitions exist", () => {
+    const rendered = convertMarkdownTables(
+      "| A | B |\n| --- | --- |\n| x | [not a ref * stray] |\n\n[other]: https://example.com",
+      "bullets",
+    );
+
+    // A stray `*` in a bracket cell would normally be delimiter-isolated, but
+    // any `[` cell might be a reference while definitions exist; the cell keeps
+    // main's exact bytes by design instead of approximating reference grammar.
+    expect(rendered).toContain("[not a ref * stray]");
+    expect(rendered).not.toContain("\\*");
+  });
+
   it("preserves CRLF source around a table", () => {
     const before = "Keep \\*literal\\*.\r\n\r\n";
     const after = "\r\n\r\nAfter.\r\n";
--- a/packages/markdown-core/src/tables.test.ts
+++ b/packages/markdown-core/src/tables.test.ts
@@ -115,6 +115,32 @@ describe("convertMarkdownTables", () => {
     },
   );
 
+  it("preserves authored emphasis across bullet cells", () => {
+    const rendered = convertMarkdownTables(
+      "| Name | Value |\n| --- | --- |\n| Entry | **bold** and _italic_ |",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    expect(parsed).toContain("<strong>bold</strong>");
+    expect(parsed).toContain("<em>italic</em>");
+    expect(parsed).not.toContain("\\*");
+    expect(parsed).not.toContain("\\_");
+  });
+
+  it("preserves escaped backticks followed by real code spans in bullet cells", () => {
+    const rendered = convertMarkdownTables(
+      "| A | B |\n| --- | --- |\n| a | \\` literal and `code` |",
+      "bullets",
+    );
+    const parsed = new MarkdownIt().render(rendered);
+
+    expect(parsed).toContain("<code>code</code>");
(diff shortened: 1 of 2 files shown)


wangmiao0668000666:
Thanks for the review — both code findings reproduced locally and are fixed in `857827d215c` (rebased onto current main). Round responses below.

**[P1] Preserve authored emphasis instead of escaping every delimiter — fixed.**
Reproduced: the previous revision turned `**bold** and _italic_` into `\*\*bold\*\* and \_italic\_`. Ownership now follows the inline parser instead of the streaming scanner: each cell is parsed with `fromMarkdown`, and only delimiters inside parsed text nodes (the parser's literal-text regions) are escaped. Emphasis, code-span, and link delimiters live in non-text nodes and survive. The cell body `**bold** and _italic_` now renders `<strong>bold</strong> and <em>italic</em>`.

**[P2] Use Markdown-aware code boundaries when preserving backticks — fixed.**
Reproduced: `` \` literal and `code` `` previously had its real closer escaped (the scanner counted the escaped backtick as an opener). With parser-owned text-node ranges the escaped backtick is literal text and the code span is a non-text node, so the cell renders `` ` literal and <code>code</code> `` unchanged. Escape handling is also explicit now: scanning steps over `\X` pairs verbatim, so `\\*`-style sequences (literal backslash followed by a live delimiter) classify correctly instead of relying on a lookbehind.

**Merge-risk (compatibility) — addressed.** The compat concern came from the same over-broad escaping; with parser-owned ranges, cells containing valid authored markup are byte-identical to main, and only cells whose delimiters previously paired across cell boundaries (the misrendering this PR fixes) change output.

**Real behavior proof — added.** New trace through the real Discord outbound formatter (`renderDiscordMarkdown`, `extensions/discord/src/markdown.ts:324`; the `convertMarkdownTables` call at `outbound-text.ts:19`), run under the extension-discord vitest project so `openclaw/plugin-sdk/*` resolves exactly like production. Before/after table plus red/green stash evidence is in the updated PR body Evidence section. WhatsApp/Signal (the other bullets-default channels) consume the same exported converter.

**check-lint-core-5 failure — pre-existing on our base, unrelated to this PR.** The failing rule is `eslint(max-lines)` in `extensions/codex/src/app-server/protocol.ts` (702 > 700, skipBlankLines/skipComments). Our base `ecbad71d09b` already fails it (file is 776 lines there); main already carries a dedicated fix, #142648 (`f059d3f8fec`, "refactor(codex): keep protocol maps within lint budget", merged Sep 8). Nothing in this PR touches extensions/codex.

Both new regression cases (authored emphasis preserved, escaped backtick + code span preserved) fail on the previous implementation and pass with this one; full markdown-core suite 26 files / 363 tests green after rebase.

wangmiao0668000666:
Thanks — the reference-link finding reproduced locally and is fixed in `4d80ba45fb1`. Responses below.

**[P2] Preserve document reference labels when classifying cell text — fixed.**
Reproduced: `[manual][api_v1]` in a value cell was emitted as `[manual][api\_v1]`, breaking the link against the untouched `[api_v1]: https://example.com` definition (verified against the vendored markdown-it 15.0.1 source — `normalizeReference` trims, folds whitespace, and case-folds but does not strip backslash escapes, so the escaped label no longer matches). The fix collects reference-definition identifiers from the full document before cell ownership runs (`convertMarkdownTables` parses the document already; definitions are rare, so the extra parse is gated on a cheap `[…]: ` presence check) and exempts reference syntax whose identifier matches a definition — full `[label][ref]`, collapsed `[ref][]`, and shortcut `[ref]` forms — from escaping. Bracket text without a backing definition is still delimiter-isolated (`[a_b]` in a cell still renders as literal `[a_b]`). A regression case with a delimiter-bearing label and an external definition now fails on the previous implementation and passes here.

**Proof-baseline corrections — done in the updated PR body.** You are right on both counts: (1) the first two BEFORE rows described the first-revision behavior, not main — main renders valid emphasis, code spans, and delimiter-free reference links correctly, and the BEFORE column now states that with the defect confined to the stray-delimiter rows; (2) WhatsApp and Signal resolve their `bullets` default through the IR renderer (`markdownToIRWithMeta` → `renderTableAsBullets`, which consumes parsed cell text, not raw source), so they are not converter consumers. The impact surface now names the actual consumer set of `convertMarkdownTables` (discord, feishu, irc, matrix, mattermost, msteams, nextcloud-talk, nostr, zalo, …).

**Dependency-version inspection gap** — the contract check you cited (whitespace/case normalization without escape stripping) was re-verified against the vendored markdown-it 15.0.1 `dist/markdown-it.mjs` that ships in this repo (`normalizeReference` at its definition, plus the inline reference rule's raw `state.src.slice` label read), so the matching contract now rests on the declared version, not the 14.1.0 proxy.

Validation: markdown-core suite 27 files / 387 tests green after rebase; `oxlint`/`oxfmt` clean on the touched files; regression red-proof via fix-stash re-confirmed for the new reference case (fails exactly on the `[manual][api_v1]` assertion).

wangmiao0668000666:
Thanks — both reference-link findings reproduced locally and are fixed in `0cfadfdd9aa` (rebased onto current main). Round responses below.

**[P2] Reference collection skips supported containers — fixed.**
Reproduced: with the document's only definition inside a block quote (`> [api_v1]: https://example.com`), the line-anchored precheck (`(^|\n)\s*\[[^\]]+\]:`) never matched, so the identifier set came back empty and `[manual][api_v1]` was emitted as `[manual][api\_v1]`, breaking the link. The precheck now gates on any `[…]:` shape in the document (`tables.ts:188`) — a false positive only costs the definition-collecting parse, and the mdast parse remains the authority on what actually is a definition, so container-nested definitions (block quotes, list items) are collected wherever CommonMark allows them.

**[P2] Reference normalization differs between producer and lookup — fixed.**
Reproduced: micromark's `normalizeIdentifier` (which markdown-it's `normalizeReference` matches) case-folds with `.toLowerCase().toUpperCase()`, so the definition `[straße_v1]: …` collects as `STRASSE_V1`, while the lookup side only lowercased raw source (`straße_v1` → `straße_v1`) and missed. Verified against the vendored `micromark-util-normalize-identifier@2.0.1` source that ships in this repo (`dev/index.js`: collapse Markdown whitespace, trim, then lower- AND uppercase — the double fold exists precisely for characters like ß whose uppercase differs from their lowercase's uppercase). Both sides now share one `normalizeReferenceIdentifier` helper mirroring that algorithm exactly (`tables.ts:57`), applied to collected mdast identifiers (idempotent — they arrive pre-folded) and to scanned cell references alike.

**checks-fast-baseline-ratchets failure — fixed.** The two `fromMarkdown(...) as PositionedNode` structural casts now carry single-line `// SAFETY:` invariants directly above each assertion; `check-assertion-safety-ratchet` passes locally against the base (3949 files, 11706 grandfathered assertions).

Validation: markdown-core suite 27 files / 389 tests green after rebase; both new regression cases (block-quoted definition, Unicode-folded label) fail on the previous implementation and pass with this one — red-proof re-confirmed via fix-stash: the block-quote case fails exactly on the `[manual][api_v1]` assertion (escaped to `api\_v1`), the folding case fails the same way on `straße\_v1`. `oxlint`/`oxfmt` clean on the touched files; core tsgo lane reports no errors in `packages/markdown-core`.

wangmiao0668000666:
@clawsweeper

wangmiao0668000666:
Thanks — all three findings reproduced locally. Rather than patching the scanner a third time, this round replaces it: reference resolution now goes through the parser's own grammar in `f9ac6ba0134` (rebased onto current main). Round responses below.

**Root-cause change: the local reference scanner is gone.**
Each finding so far traced to the same defect — `referenceSpans` approximated CommonMark's reference grammar with a regex, and every approximation leaked (definition containers, case folding, and now multiline labels, padded labels, Unicode whitespace). The rewrite deletes the scanner and the normalization helper entirely: `collectReferenceDefinitions` (`tables.ts:74`) collects the exact source slices of the document's `definition` nodes, and each cell's parse input is the cell source plus those definitions appended after a blank line (`literalTextRanges`, `tables.ts:44`). micromark then resolves `[label][ref]`, `[ref][]`, and `[ref]` into `linkReference` nodes using its own label grammar — the same grammar markdown-it's `normalizeReference` implements — so multiline labels, padded labels, trimmed lookups, and case folding all behave identically by construction. Text inside `linkReference`/`imageReference` nodes is exempt from escaping because reference labels must keep their exact source bytes for the definition match; inline-link labels remain escapable since they render identically either way. Two details keep the classification faithful: whitespace in the parse copy is flattened to ordinary spaces (UTF-16 offsets preserved 1:1, so ranges still index the original cell bytes) to match markdown-it's Unicode `\s` folding, and ranges from the appended definitions are dropped by offset. Net effect on production code: −30 lines and one fewer grammar to maintain.

**[P2] Allow multiline labels in the definition precheck — fixed.**
Reproduced: `[api_\nv1]: https://example.com` backing `[manual][api_ v1]` was rejected by the `[^\]\n]+` precheck, the identifier set came back empty, and the cell was emitted as `[manual][api\_ v1]`. The gate now allows newlines inside the brackets (`/\[[^\]]+\]:/u`, `tables.ts:206`), and the mdast parse — not the precheck — decides what is a definition, so a multiline label is collected and re-resolved like any other.

**[P2] Recognize shortcut references with leading whitespace — fixed.**
Reproduced: `[ api_v1 ]` with `[api_v1]: https://example.com` in the document was escaped to `[ api\_v1 ]`. Verified against both vendored parsers: micromark parses the padded shortcut as a `linkReference` (identifier `api_v1`, label trimmed during normalization) and markdown-it 15.0.1 renders it as `<a href="https://example.com"> api_v1 </a>`. With definitions appended to the parse input, the padded label resolves through the same path as any other reference — no scanner boundary to miss it.

**[P2] Match MarkdownIt's Unicode whitespace normalization — fixed.**
Reproduced: `[manual][api_ v1]` against a definition whose label contains U+00A0 (and the mirror image — U+00A0 in the reference, ordinary space in the definition) was escaped on the ASCII-only normalization. The parse copy now flattens all Unicode whitespace to ordinary spaces before parsing — a 1:1 UTF-16 substitution, so node offsets still index the original cell bytes, emphasis flanking still sees whitespace, and markdown-it's `\s`-based fold is matched in both directions. Classification always applies to the original, untouched source.

Validation: all three new regression cases fail on the previous head for the intended reason (label emitted with an escaped underscore, link dies) and pass with this change — red-proof re-confirmed via patch revert. `tables.test.ts` 32/32, markdown-core suite 27 files / 392 tests green after rebase onto current main; assertion-safety ratchet OK locally (3949 files, 11706 grandfathered); `oxlint`/`oxfmt` clean; core tsgo lane reports no errors in `packages/markdown-core`.

wangmiao0668000666:
@clawsweeper

wangmiao0668000666:
@clawsweeper re-review

Thanks — both findings reproduced locally. This round removes the regression class instead of patching it: all reference-reconstruction machinery is deleted in `982ad77c54a`, and cells that could contain a reference now pass through byte-identical to main. Round responses below.

**Root-cause change: no more reference grammar approximation.**
Five rounds in, every reference finding traced to the same structural mistake — the converter kept re-implementing one more corner of the renderer's reference grammar (identifier scanner → definition slicing → label precheck), and each approximation leaked the next edge case. The grammar that decides whether `[manual][api\]_v1]` resolves belongs to the renderer, so this revision stops interpreting it entirely:

- A cell could contain a reference only if it contains `[` **and** the document contains a definition — and every reference definition contains the `]:` terminator. So `referencesPossible = /\]:/u.test(document)` plus `cell.includes("[")` is a complete, conservative gate: either condition failing *proves* the cell is reference-free (no false negatives; false positives only widen pass-through).
- Possible-reference cells pass through with main's exact bytes. No escaping, no reconstruction, no parse — so no regression is structurally possible for any reference form the renderer supports: containers, multiline labels, Unicode folds, padded labels, escaped brackets, or anything else.
- Deleted: `collectReferenceDefinitions`, the definition precheck, the `definitionSuffix` threading, the whitespace-flattening parse copy, and the `linkReference` exemption (unreachable now — cells are only parsed when no definition exists anywhere). Production code drops ~40 lines net versus the previous revision.

Documented tradeoff: a stray delimiter inside a possible-reference cell is no longer isolated (main's behavior kept there). Covering that intersection would require exactly the reference-grammar approximation this revision deletes, so the PR chooses zero reference regressions over covering it; the pass-through is pinned by a dedicated test.

**[P2] Preserve container-free definition content when rebuilding references — fixed by deletion.**
Reproduced: `[manual][api_ v1]` backed by `> [api_\n> v1]: https://example.com` was escaped to `[manual][api\_ v1]` because the flattened definition slice (`[api_ > v1]: …`) defined a different identifier. There is no definition slicing anymore: the document contains `]:` and the cell contains `[`, so the cell passes through untouched and the renderer resolves the reference with its own container handling. Regression test `preserves references backed by a continued block-quote definition` fails on `f9ac6ba0134` (asserts the unescaped label and the resolved `<a href>`) and passes on this head.

**[P2] Allow escaped closing brackets through the definition precheck — fixed by deletion.**
Reproduced: `[manual][api\]_v1]` backed only by `[api\]_v1]: https://example.com` found no definition through the `[^\]]+` precheck, so the underscore was escaped and the link died. There is no precheck anymore: `[api\]_v1]: …` contains `]:`, so `referencesPossible` is true and the cell passes through byte-identical. Regression test `preserves references whose label contains an escaped closing bracket` fails on `f9ac6ba0134` and passes on this head.

**Merge-risk (compatibility) — the regression surface is gone.** Previous revisions carried the risk because they rewrote reference-bearing cells; this revision's whole point is that those cells are now byte-identical to main. The only cells whose bytes change are ones the parser proves reference-free.

**Proof.** Red/green: both new reference tests fail on `f9ac6ba0134` exactly on their assertions, pass on `982ad77c54a`. Full `tables.test.ts` 35/35, markdown-core suite 27 files / 395 tests, `extensions/discord/src/markdown.test.ts` 2/2 — all run in an isolated Docker test environment. The PR body's evidence table now includes the two new pass-through rows; the earlier formatter trace (stray-delimiter and authored-markup rows through `renderDiscordMarkdown`) is unaffected by this reslice since those cells contain no `[`.

wangmiao0668000666:
@clawsweeper re-review

Thanks — the autolink finding reproduced locally and is fixed in `d2c8acdc2e0`. Round response below.

**[P2] Preserve autolink text before escaping cell delimiters — fixed.**
Reproduced: `<first_last@example.com>` in a value cell was emitted as `<first\_last@example.com>`, which markdown-it's email-autolink grammar rejects (verified against the vendored markdown-it 15.0.1 — `EMAIL_RE` accepts underscores but not the inserted backslash, and URL autolinks pass the altered source to `normalizeLink`). Root cause: an autolink parses to a `link` node whose positioned `text` child *is* the destination, so the traversal classified destination bytes as escapable prose. The fix is structural, not another grammar approximation: autolinks are the only link form whose source starts with `<`, so the visitor now skips `link` subtrees with `markdown[start] === "<"` (`tables.ts:55`) — destination bytes are never classified. Inline and reference links start with `[` and keep prose text, so their escapable behavior is unchanged. This closes the node-type class: across mdast's inline grammar, the autolink is the only node whose `text` child carries destination rather than prose semantics.

**Regression coverage with destination assertions.** New test `preserves angle-bracket autolink destinations in bullet cells` covers both forms — `<first_last@example.com>` and `<https://example.com/a_b>` — asserting the rendered cell keeps its exact bytes and the parsed output still contains `<a href="mailto:first_last@example.com">` and `<a href="https://example.com/a_b">`. Red/green: the test fails on `982ad77c54a` exactly on those assertions (escaped underscore present, autolink dead) and passes on this head.

**Proof.** Full `tables.test.ts` 36/36, markdown-core suite 27 files / 396 tests, `extensions/discord/src/markdown.test.ts` 2/2 — all run in the same isolated Docker environment as before; `oxfmt` clean. The PR body's evidence table gained the autolink row; the accepted formatter trace from the previous round is unaffected (its cells contain no autolinks).

On the partial-coverage merge-risk note (possible-reference cells keep main's stray-delimiter behavior): that tradeoff is deliberate and documented — covering it requires exactly the reference-grammar approximation that five review rounds showed cannot be made faithful locally, so the PR chooses zero reference regressions over covering the rare intersection.

wangmiao0668000666:
Closing this PR after 7 review cycles — with a summary so the next person who picks up #142647 doesn't have to re-derive it.

**Why I'm closing.** The last two findings (CJK emphasis classification, bare-URL destinations) are real, and they reveal that the remaining defects are not approximation bugs but a structural mismatch: there is no single correct delimiter classification for a cell, because each downstream renderer brings its own configuration — `markdownToIR` installs `markdown-it-cjk-friendly` (Teams path), Matrix enables linkification after table conversion, Discord renders plain markdown-it. Classifying delimiters with any one grammar (CommonMark via `fromMarkdown`, as here) provably diverges from at least one consumer, and aligning with all of them from inside `convertMarkdownTables` would mean re-implementing every channel's renderer configuration — exactly the local-grammar-approximation trap the earlier reference-link rounds fell into, one level deeper. Seven cycles produced eight revisions; each fix of one renderer's contract surfaced another's. I don't see this approach converging, so I'm cutting it here.

**What was established (for whoever revisits #142647):**

1. **The defect is real and current** — bullets-mode conversion mixes raw cell markdown with generated `**` markers, and a stray `*`/`_`/`` ` `` in a cell pairs across cells or swallows the row boundary. Confirmed on main and on the latest release tag (v2026.9.3).
2. **Renderer-config compatibility is the hard constraint** — any fix must preserve: authored emphasis under `markdown-it-cjk-friendly` flanking rules, reference links matched by identifier against document definitions (incl. containers, multiline labels, Unicode folds, escaped brackets), angle-bracket autolinks whose text child *is* the destination, and bare URLs later consumed by linkify-enabled renderers.
3. **The promising direction is parse-time ownership, not render-time classification** — escape or tag stray delimiters when the table is first parsed (where the full document context and the actual renderer config are available), rather than in the bullets formatter, which is the one surface that lacks both. The evidence table and regression tests in this branch's history document the full constraint set.

Thanks to ClawSweeper for the thorough reviews — the late findings were each verified as real before this decision.

PR #142657, opened 2026-09-08
fix(ui/chat): persist the background-tasks rail collapse across restarts

### LINKED ISSUE #142611: Control UI show background toggle does not persist across gateway restarts
Every time the gateway restarts, the show background toggle in the session view resets. The user has to manually re-enable it each time.

Steps to reproduce:
1. Open Control UI session view
2. Toggle show background on
3. Restart the gateway (or close/reopen OpenClaw)
4. Toggle is back to off — background activity not shown

Expected: toggle preference persists across restarts
Actual: toggle resets every time

Environment: Windows 10.0.26200, OpenClaw 2026.8.2 (stable)

xiaoguomeiyitian:
## What Problem This Solves

The Control UI background-tasks rail toggle ("show background") reset to collapsed every gateway restart or pane reload, so operators had to re-enable it each session. The choice was kept purely in memory.

Fixes #142611.

## Why This Change Was Made

`BackgroundTasksState.collapsed` defaulted to `true` on every scope reset with no persistence. Persist the preference in `localStorage` through a safe get/set helper (mirroring the existing `chat-observer-display` pattern), restoring the expanded choice on pane init.

## User Impact

The toggle now survives gateway restarts. The default remains collapsed for first-time users; only an explicit expansion is remembered. The in-memory value still wins across thread switches within a session. Privacy mode (unavailable `localStorage`) degrades to the prior in-memory-only behavior.

## Evidence

- New test: toggling open persists the expansion and restores it when the pane remounts with no in-memory state.
- `chat-background-tasks.test.ts` 36/36 pass; `ui/src/pages/chat/components/` 1271 tests pass.
- `oxlint` clean; `git diff --check` clean.

## Approach (non-obvious)

- Storage protocol: default collapsed; only `"0"` means expanded, so any missing/corrupt value (or `localStorage` throwing) falls back to the collapsed default instead of accidentally expanding.

### DIFF
--- a/ui/src/pages/chat/components/chat-background-tasks.test.ts
+++ b/ui/src/pages/chat/components/chat-background-tasks.test.ts
@@ -156,6 +156,7 @@ it("renders task-shaped placeholders while the initial task list loads", async (
 afterEach(() => {
   document.body.replaceChildren();
   vi.restoreAllMocks();
+  localStorage.removeItem("openclaw.chat.backgroundTasks.collapsed");
 });
 
 describe("background tasks rail state", () => {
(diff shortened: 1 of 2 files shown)


xiaoguomeiyitian:
Closing this PR. After deeper review, the implementation adds a second `localStorage`-backed collapse preference in the background-tasks helper, but production wires the visible Tasks toggle through `sidebarSessionLayouts` (`chat-pane-rails.ts` overrides `collapsed` and `onToggleCollapsed`). As a result, this change cannot affect the observed UI behavior. The reported reset needs to be reproduced and fixed through the sidebar-layout owner (chat-state-page.ts save/restore boundaries), not via a parallel preference. Will revisit if the reset reappears on current main.

PR #142661, opened 2026-09-08
fix(gateway): preserve MCP App ticket authority

eleqtrizit:
## What Problem This Solves

Fixes an issue where a paired Gateway client limited to read operations could open a standalone MCP App view and receive a ticket that still allowed App-visible tool calls.

## Why This Change Was Made

Standalone tickets now preserve whether their issuer is authorized for MCP App tool operations, and ticket reuse keeps weaker and stronger grants separate. Redemption intersects that ticket authority with the live view's existing tool allowlist, while channel-created launch actions explicitly retain their established capability.

## User Impact

Read-scoped clients can continue rendering standalone MCP App views and using permitted resources, but they can no longer execute App tools. Write-authorized clients and channel launch actions retain existing interactive behavior.

## Evidence

- Focused MCP App Gateway, standalone-host, and channel-presentation suites: 83 tests passed.
- Real Gateway integration proof on head `dd480d59f5c`: independently paired read-only and read/write device identities connect over WebSocket, receive authority-separated standalone tickets, and redeem them against a real stdio MCP server.
- The proof verifies read-only direct and standalone calls are denied without mutation, the write-authorized standalone call mutates exactly once, and revoking the live App-interaction grant blocks a later call before a second mutation.
- Exact proof command: `node scripts/run-vitest.mjs src/gateway/server.mcp-request-shutdown.test.ts` (2/2 passed).
- Scoped core and core-test changed-file checks passed, including typechecks, lint, formatting, dead-code, import-cycle, schema, and storage guards.
- Independent P1 autoreview found no accepted or actionable findings.

### DIFF
--- a/src/gateway/mcp-app-channel-action.test.ts
+++ b/src/gateway/mcp-app-channel-action.test.ts
@@ -96,6 +96,12 @@ describe("materializeMcpAppChannelPresentation", () => {
     });
 
     expect(mocks.createTicket).toHaveBeenCalledOnce();
+    expect(mocks.createTicket).toHaveBeenCalledWith({
+      sessionKey: "agent:main:do-not-emit-session",
+      view,
+      toolOperationsAuthorized: true,
+      nowMs,
+    });
     expect(presentation).toEqual({
       blocks: [
         {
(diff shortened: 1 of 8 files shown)


eleqtrizit:
@clawsweeper re-review

eleqtrizit:
@clawsweeper re-review

eleqtrizit:
Merging as squash. Merged because the exact PR head passed the repository CI gate and independent security review found no blocking defect after accepting the production-path authorization proof. Compatibility audit found no config, default, schema, protocol, dependency, or public SDK changes: standalone ticket URLs and authorized write/admin behavior are unchanged, tickets remain ephemeral in-process state, and the only removed behavior is the unauthorized tool bridge for read-scoped callers, which is the intended security fix.

PR #142678, opened 2026-09-08
fix: HTML-only email stalls packaged IMAP delivery

### LINKED ISSUE #142574: imap: "sweep failed=Failed to parse HTML" on single-part HTML-only emails (entities@8 vs htmlparser2@10 fromCodePoint mismatch)
## Environment

- `ghcr.io/openclaw/openclaw:2026.9.2-slim` (bundling is arch-independent; reproduced on arm64)
- FastMail account, `watch.mode: auto`, `includeBody: true`

## Symptom

Every 60s:

```
[plugins] imap: account=personal sweep failed=Failed to parse HTML
```

The sweep aborts on the first such message and **never advances the cursor** (plugin state `imap/cursor/<account>` stays frozen), so all subsequent mail silently stops being routed to the agent. Observed backlog: 30 messages.

## Root cause

The `extensions/imap` bundle inlines `htmlparser2@10.1.0` — whose code calls the `entities@^7` API (`fromCodePoint`, `uriDecode`) — but **externalizes** `entities` via `__require("entities/decode")`. At runtime that resolves to the app-level hoisted `entities@8.0.0` (the root `package.json` pins `entities: 8.0.0`), where those exports were removed.

For single-part HTML-only messages, mailparser calls `htmlToText`; html-to-text runs htmlparser2 with `decodeEntities: true`; any HTML entity (e.g. `&apos;` in an `alt` attribute) reaches `onattribentity`/`ontextentity`, which calls `decode_1.fromCodePoint(cp)`. With entities v8 that's `undefined`, so it throws `TypeError: decode_1.fromCodePoint is not a function`, which mailparser catches and re-emits as `Failed to parse HTML`.

The unbundled pnpm-installed mailparser resolves `entities@7` correctly from its own node_modules — only the shipped bundle is broken.

## Self-contained repro (no real message required; run inside the image)

```js
const { createRequire } = require('node:module');
const d = createRequire('/app/dist/extensions/imap/index.js')('entities/decode');
typeof d.fromCodePoint; // => undefined

// Extract the inlined mailparser from the shipped bundle (regions: encoding-japanese,
// libmime, mailsplit, htmlparser2/html-to-text tree, mailparser; plus the externalized
// entities/iconv-lite requires), then run the plugin's exact call on a synthetic message:
simpleParser(Buffer.from(
  'From: s@example.com\r\n' +
  'To: r@example.com\r\n' +
  'Subject: t\r\n' +
  'Content-Type: text/html\r\n' +
  '\r\n' +
  '<p>Sheena&apos;s expert advice.</p>\r\n'
), { skipImageLinks: true, skipTextToHtml: true });
// => rejects: Failed to parse HTML

// Control: the same entity inside multipart/alternative WITH a text part parses fine —
// only the root-level html path (single-part HTML-only) triggers htmlToText.
```

Or minimally, the throw is simply:

```js
htmlToText('<p>Sheena&apos;s expert advice.</p>');
// => TypeError: (0, decode_1.fromCodePoint) is not a function
```

## Suggested fixes

1. Inline `entities@7` into the imap bundle, or pin/alias it so the externalized `entities` resolves to the major the inlined htmlparser2 expects.
2. Make the sweep tolerate per-message parse failures and advance the cursor instead of aborting the whole batch (a single malformed email should not stall all mail forever).
3. Consider exposing `skipHtmlToText` in the imap config as a defensive option.

## Impact

Email routing silently stops (no error escalation beyond the repeated log line).

ericcaiwx-star:
Closes #142574

## What Problem This Solves
HTML-only mail containing `&apos;` can stop packaged IMAP delivery and leave later mail queued.

## Why This Change Was Made
The decision to bundle the parser's decoder has one owner: `shouldAlwaysBundleDependency` in `tsdown.config.ts`. It now retains `entities/decode` from the parser's compatible dependency. The watcher remains the sole cursor writer.

## User Impact
HTML and later mail dispatch in order without resetting mailbox state.

## Evidence
`pnpm build` passes on main and the candidate.
`node scripts/run-vitest.mjs run test/scripts/tsdown-config.test.ts -t 'packaged imap-watch'` fails on main and passes with this change.
The unchanged generated plugin connects to a local synthetic mailbox: main stops at message 3; the fix decodes the apostrophe, dispatches later mail, and reaches message 5. Restart dispatches no duplicates. No external account or recipient delivery is claimed.

## Compatibility
No dependency versions, configuration, sender policy, or stored data formats change.

## Consumers
Packaged IMAP uses its existing parser and sender checks.

## Invalidation
Rebuild the package; no mailbox reset is needed.

## Tests
57 build checks and 53 IMAP checks pass. Plain, multipart, denied-sender, ordering, cursor, and restart controls are retained.
The subprocess fixture uses a module-relative URL so dependency analysis follows the same file that the test executes.

### DIFF
--- a/test/scripts/tsdown-config.test.ts
+++ b/test/scripts/tsdown-config.test.ts
@@ -201,6 +201,75 @@ describe("tsdown config", () => {
     },
   );
 
+  it("parses HTML-only IMAP mail from the bundled runtime", async () => {
+    const entryName = "extensions/imap/index";
+    const selected = configs.find((config) =>
+      hasWorkerEntry(config, entryName, "extensions/imap/index.ts"),
+    );
+    expect(selected).toBeDefined();
+
+    const root = fs.realpathSync(createTempDir("openclaw-tsdown-imap-"));
+    fs.writeFileSync(path.join(root, "package.json"), '{"type":"module"}');
+    // Mirror the packaged layout: the bundled plugin runs beside root production dependencies.
+    fs.symlinkSync(fs.realpathSync("node_modules"), path.join(root, "node_modules"), "dir");
+
+    const bundles = await build({
+      ...selected,
+      config: false,
+      entry: { [entryName]: "extensions/imap/index.ts" },
+      outDir: path.join(root, "dist"),
+      dts: false,
+      logLevel: "silent",
+    });
+    try {
+      const artifact = path.join(root, "dist", `${entryName}.js`);
+      const proofArtifact = path.join(root, "dist", `${entryName}-proof.js`);
+      const artifactSource = fs.readFileSync(artifact, "utf8");
+      const defaultExport = "export { imap_default as default };";
+      expect(artifactSource).toContain(defaultExport);
+      // Expose the parser binding from the actual plugin artifact; its bundled code is unchanged.
+      fs.writeFileSync(
+        proofArtifact,
+        artifactSource.replace(
+          defaultExport,
+          "export { imap_default as default, import_mailparser as __mailparser };",
+        ),
+      );
+      const script = `
+        import assert from "node:assert/strict";
+        import { pathToFileURL } from "node:url";
+        const [entry] = process.argv.slice(1);
+        const { __mailparser } = await import(pathToFileURL(entry).href);
+        const mail = await __mailparser.simpleParser(Buffer.from(
+          "From: sender@example.com\\r\\n" +
+          "To: recipient@example.com\\r\\n" +
+          "Subject: Bundled HTML\\r\\n" +
+          "Content-Type: text/html\\r\\n" +
+          "\\r\\n" +
+          "<p>Sheena&apos;s expert advice.</p>\\r\\n"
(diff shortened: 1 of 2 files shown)


ericcaiwx-star:
@clawsweeper re-review

obviyus:
landpr-approve-admin-merge
pr: 142678
head: f5b472a832cbb71b7bef1eddf52cbbc286b71ad5
scope: owner-review

obviyus:
Thanks @ericcaiwx-star!

**Proof:** The packaged IMAP service decodes HTML-only mail, requests accepted messages in order, advances its cursor, and does not repeat requests after restart. Denied senders remain blocked. This uses a local test mailbox and does not claim external delivery.

- **Bug:** An HTML entity could stop parsing and leave later mail queued.
- **Fix:** The package builder retains the parser's compatible entity decoder.

| Before | After |
| --- | --- |
| HTML-only mail stalls the cursor and later mail stays queued. | HTML mail decodes, later requests proceed, and the cursor advances. |

PR #142685, opened 2026-09-08
test(live): avoid secret-shaped MiniMax tool probes

RomneyDa:
## Summary

- keep the gateway live tool-read proof unique and fail-closed
- present generated values as harmless test markers instead of secret-like nonce material
- retain the exact two-value readback assertion and retry budget

## Root cause

MiniMax M3 used the read tool successfully, then refused all three requests to repeat values described as nonces. Both selected MiniMax models consequently skipped and the stable live profile completed with zero successful models.

## Validation

- gateway live test file: 174 passed, 2 skipped
- Oxfmt and `git diff --check`

This addresses the MiniMax live-profile failure in Full Validation run 34287144233 / child run 34287196590.


RomneyDa:
@clawsweeper re-review

PR #142708, opened 2026-09-09
fix: keep shipped SDK context aliases pending removal

steipete:
## What Problem This Solves

Resolves a problem where the compatibility gate treated the channel prompt-context SDK aliases as ready to remove after September 8, even though published-plugin migration and breaking-release approval are still outstanding. These aliases ship in stable v2026.9.3 and v2026.9.2.

## Why This Change Was Made

Classify the existing record as `removal-pending` and state its outstanding conditions in the compatibility guide and report. Keep the original September 8 review date: the report still lists this item as due for review instead of extending its deadline or removing shipped exports.

## User Impact

Existing plugins retain the same SDK aliases and runtime behavior. Maintainers get an accurate pending-removal queue with the remaining migration and approval work visible. No API, configuration, runtime, or eligibility-guard change; net production LOC is zero.

## Evidence

- Actual boundary-report command on the base exits 1 with this sole date-eligible record; the candidate exits 0 and reports the same record as `removal-pending`, `dueForReview: true`, with its unchanged date and explicit blockers.
- All 16 existing compatibility-registry and boundary-report tests pass, including the updated pending inventory and retained-date/blocker assertion. The initial stale-inventory failures are preserved in local evidence.
- Independent full P2 autoreview: scoped-clean, no actionable findings.
- Full canonical changed-code gate passes, including core/test types, all dead-export scans, lint, docs/config checks, and boundary guards.

PR #142722, opened 2026-09-09
fix(plugins): retain context aliases pending approved removal

steipete:
Related: #139428

## What Problem This Solves

The channel-context SDK alias migration window ended on September 8, but removal still requires verified published-plugin migration and explicit breaking-release approval. Keeping the record classified as eligible deprecated debt made the boundary guard fail at UTC midnight even though removing the public aliases was not authorized.

## Why This Change Was Made

Move only `sdk-untrusted-context-identifier-aliases` into the existing `removal-pending` review queue and state both blockers. Preserve its original `2026-09-08` date so it remains overdue and visible; update the compatibility guide and existing registry/report assertions.

## User Impact

The public aliases, deprecation annotations, canonical-field precedence, and inbound normalization are unchanged. This does not renew the deadline or approve a breaking removal. The boundary guard is unchanged, and the pending record still exposes its blocker and reader references.

## Evidence

- Original guard failed with one eligible record. Afterward the same guard passes with zero eligible records, nine pending records and one due review. The other eight pending records are identical.
- The record retains `2026-09-08`, `dueForReview: true`, its explicit blocker, and nonempty reader references. Those references are existing surface-token triage matches, not verified external-plugin usage.
- All 56 tests across the existing registry, boundary-report, inbound-context and channel-metadata suites passed.
- Public alias, normalization and gate source files are byte-identical to the base.

Required changed-file checks and independent P2 review passed. This maintenance change adds eight test lines; it claims no test removal or runtime improvement.

### DIFF
--- a/docs/plugins/compatibility.md
+++ b/docs/plugins/compatibility.md
@@ -158,13 +158,14 @@ New channel plugins should use `MsgContext.ChannelPromptContext`,
 `SupplementalContextFacts.channelStructuredContext`. The older
 `UntrustedContext`, `UntrustedStructuredContext`,
 `UntrustedStructuredContextEntry`, and supplemental `untrustedContext` names
-remain as deprecated SDK aliases until 2026-09-08 (registry record
-`sdk-untrusted-context-identifier-aliases`). Inbound finalization folds those
-deprecated fields into the channel-named fields and removes the old keys from
-runtime context.
+remain available as deprecated SDK aliases. Their original removal window ended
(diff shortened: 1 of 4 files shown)


steipete:
Superseded by #142708, merged as c7cb45dfd5dbcf2098d7dd10efd4a64054a2662d. That change implements the same removal-pending status, preserves the original date and public aliases, and records both migration and breaking-release approval blockers. Its exact-head CI run [34295018599](https://github.com/openclaw/openclaw/actions/runs/34295018599) passed. Closing this duplicate without another metadata change.

PR #142725, opened 2026-09-09
fix(chat): prevent duplicate messages after CLI history refresh

steipete:
## What Problem This Solves

Fixes an issue where users viewing a CLI-backed chat could see the same persisted message twice after a history refresh overlaps a live message event. The history response adds CLI provenance to a local row; the live event for that same row keeps its native metadata.

## Why This Change Was Made

The shared identity reader now keeps a locally persisted message in the native identity namespace when its reader-owned transcript position is valid. CLI provenance remains available in the message, and genuine imported-only IDs remain isolated. Projection matching, pending messages, and final-message replay all use that same classification.

## User Impact

A native message remains one message as CLI history enriches it. Distinct imported messages still remain separate, even when their provider-local IDs collide with native IDs.

## Evidence

- 179 current-base package/UI owner cases passed, including terminal reconciliation, imported collisions, and canonical user/assistant enrichment. Core production/test types, targeted lint, and diff checks passed. Managed review completed both passes with no actionable findings.
- The real Gateway transcript writer, queued broadcaster, history handler, and UI reducer reproduced the duplicate with one failing case and two passing controls. All ten cases passed after the repair. This integration proof ran on main commit [127e3ffa](https://github.com/openclaw/openclaw/commit/127e3ffa3080841cef31927c57071c7ae9ae52ae); the producer code and repaired identity implementation are unchanged on the current base [306351ac](https://github.com/openclaw/openclaw/commit/306351ac3c9950a89a9873c092f835494bb982d1).
- The same synthetic prebuilt-browser driver rendered two bubbles before and one afterward, with the same session, leaf, native ID, and message data. These captures ran on [127e3ffa](https://github.com/openclaw/openclaw/commit/127e3ffa3080841cef31927c57071c7ae9ae52ae) and use the exact repaired production bytes; current-base owner tests and source review cover the subsequent carry.
- The full changed gate remains failed on an unrelated SDK compatibility record that became due at UTC midnight. That policy decision is tracked separately; no date override, waiver, or skipped-gate pass is claimed.

| Before | After |
| --- | --- |
| Before: the same native message appears twice | After: one native message remains |

### DIFF
--- a/packages/gateway-client/src/session-projection-conformance.test.ts
+++ b/packages/gateway-client/src/session-projection-conformance.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import {
   createSessionProjection,
   readSessionMessageIdentity,
+  readSessionProjectionFinalMessageIdentity,
   reduceSessionProjection,
   type SessionProjectionEvent,
   type SessionProjectionScope,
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [03a3a30a5f075cc42910ca3fc8bfc759da459c98](https://github.com/openclaw/openclaw/pull/142725/commits/03a3a30a5f075cc42910ca3fc8bfc759da459c98)
- Landed commit: [1c597f626cceb4d136d0e6eb047251a65953f417](https://github.com/openclaw/openclaw/commit/1c597f626cceb4d136d0e6eb047251a65953f417)

PR #142727, opened 2026-09-09
fix(release): admit frozen dirty-block exit contract

RomneyDa:
## Summary

- derive the frozen dirty-worktree exit contract from the selected source
- accept exit zero only for that exact frozen contract
- retain the structured rejection and unchanged-checkout assertions

## Root cause

2026.7.33 rejects an untracked-file Git update with a structured blocked result but exits zero. Current tooling required exit 1 for every non-legacy package, so both update-channel diagnostic lanes failed after the actual rejection succeeded.

The compatibility flag is derived from the selected source owner shape, defaults closed, and is separate from the preview-reporting flag. Current and unknown candidates remain strict.

## Validation

- tooling compatibility: 34 passed
- shell syntax
- Oxfmt
- `git diff --check`

This covers both `update-channel-switch` failures in Full Validation 34294189135.

### DIFF
--- a/scripts/e2e/lib/update-channel-switch/assertions.mjs
+++ b/scripts/e2e/lib/update-channel-switch/assertions.mjs
@@ -10,7 +10,7 @@ const controlUiHtml = "<!doctype html><title>fixture</title>\n";
 
 function usage() {
   console.error(
-    "usage: assertions.mjs <prepare-git-fixture|write-control-ui|assert-update|assert-dry-run|assert-config-channel|assert-status-kind|assert-installed-version|assert-runtime-staging-clean|assert-dirty-update> [...]",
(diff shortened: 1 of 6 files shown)


RomneyDa:
@clawsweeper re-review

The exit admission is now explicit: current candidates accept only 1; source-qualified frozen/legacy candidates accept only 0 or 1. Behavioral coverage rejects timeout 124 and abnormal exit 2 while retaining the structured dirty payload and unchanged-checkout assertion. Focused tooling suite: 34 passed.

RomneyDa:
Updated the source-contract test for the stricter `assert-dirty-exit` helper after CI exposed the stale inline-status expectation.

Exact tooling-Docker suite: 286 passed, 2 skipped. Oxfmt and diff-check pass.

@clawsweeper re-review

PR #142733, opened 2026-09-09
refactor(tests): avoid idle wait after profiled fork cleanup

steipete:
## What Problem This Solves

The paused profile-fork cleanup test spends a minute waiting after the real worker has already flushed its teardown acknowledgement and been stopped. That wait slows validation without adding another observation to the cleanup contract.

## Why This Change Was Made

The existing generated preload now observes the selected fork's stopped state after Vitest has consumed its acknowledgement, requires exactly one live 60-second stop timer, and advances that timer's callback. The real fork, awaited CPU/heap profile writes, native 500 ms kill escalation, exit receipt, namespace cleanup, and independent inner test verdicts remain exercised. The production timeout values are unchanged.

## User Impact

This is a test-only change with 40 net lines added. In one local paired run, the focused case fell from 62.529 seconds to 2.411 seconds (native shard wall time: 63.88 to 3.37 seconds). It avoids the controlled 60-second wait in that measured case; this is not a claim of an overall 30% suite improvement.

## Evidence

- Complete cleanup and fork-shutdown suites: 93 passed (76 + 17); all existing cleanup case names remain in order.
- Separate capture-only run retained the exact acknowledgement/SIGKILL/deadline receipt, valid CPU/heap profiles, namespace/config receipts, and the intended one-passed/one-failed inner verdict.
- Removing awaited profile finalization fails the existing profile-file assertion. Suppressing worker termination fails the existing exit-receipt check; the supervisor's original cleanup leaves the recorded group and both PIDs dead. These controls were restored before final validation.
- Fresh independent review and the normal full changed gate passed. The patch and relevant test/dependency inputs remained byte-identical across the upstream metadata-fix refresh.

### DIFF
--- a/test/scripts/run-vitest-state-cleanup.test.ts
+++ b/test/scripts/run-vitest-state-cleanup.test.ts
@@ -390,6 +390,18 @@ export default {
         `
 const { subscribe } = require("node:diagnostics_channel");
 const fs = require("node:fs");
+const { execFileSync } = require("node:child_process");
+const schedule = globalThis.setTimeout;
+const cancel = globalThis.clearTimeout;
+const deadlines = new Map();
+globalThis.setTimeout = (callback, delay, ...args) => {
+  if (delay !== 60000) return schedule(callback, delay, ...args);
+  const invoke = () => callback(...args);
+  const timer = schedule(() => { deadlines.delete(timer); invoke(); }, delay);
+  deadlines.set(timer, invoke);
+  return timer;
+};
+globalThis.clearTimeout = timer => { deadlines.delete(timer); return cancel(timer); };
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d8999b486d0e9c47b114a1d5c5070930c7914c60](https://github.com/openclaw/openclaw/pull/142733/commits/d8999b486d0e9c47b114a1d5c5070930c7914c60)
- Landed commit: [ba068afa8c9a8df1a70a31b7ec8c6177ad8137ec](https://github.com/openclaw/openclaw/commit/ba068afa8c9a8df1a70a31b7ec8c6177ad8137ec)

PR #142738, opened 2026-09-09
fix(migrate): close provider resources after completed operations

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where migration commands and setup flows retained plugin database connections after finishing. Completed imports, optional onboarding results, or projected configuration could also be lost when registration cleanup failed. A Gateway result-encoding failure could allow the same completed import to run again on retry.

## Why This Change Was Made

Each cold migration operation owns its registration through planning, prompts, apply, output, and preparation cleanup. Nested callers carry the selected provider and opaque plans; setup carries plain discovery descriptors so listing choices stays inert. Active Gateway registrations keep their existing owner.

The operation owner can report cleanup failure separately after its callback succeeds. Optional discovery, Memory import, and post-install setup use that boundary to preserve completed outcomes; acquisition, operation, and authority failures still reject. Gateway apply records completion before result projection and retains either its JSON result or a terminal encoding error for the same request.

## User Impact

Migration connections close when their operation finishes. Completed imports and setup results survive registration cleanup errors, with visible cleanup diagnostics. Gateway retries reuse completed outcomes without reapplying. The Memory import documentation explains when to inspect reports and destination files before starting another import.

## Evidence

The full 27-path candidate passed fresh independent Codex review, its canonical changed gate (347 seconds), and the ordinary build (223 seconds). The final optional-flow and affected-owner sweep passed 101 tests, including five reproduced normal-outcome cleanup failures and an authority-failure control. Earlier native lifecycle, wizard, CLI, and RPC regressions also passed.

Current built CLI list/plan/apply recorded three native opens, three explicit closes, and one durable import. Three real isolated Gateway cases—success, cleanup failure after apply, and result-encoding failure after apply—each recorded two opens/two explicit closes, one durable import, and identical retry JSON after physical closure without additional provider execution. The complete live sequence took 26.7 seconds. All three Gateways stopped normally; 67 source/context files and 8,776 compiled modules stayed unchanged.

The review finding about optional onboarding cleanup is fixed and covered by native tests. The review's persisted-table warning was a classification false positive: this diff adds no SQLite schema, migration, index, or stored-data format change; replay uses the existing in-memory dedupe map.

Two earlier CI attempts failed unchanged Signal first-reply tests before the injected reply callback. The same failure was reproduced independently on main and traced to cold provider-policy loading; a separate guarded import reduction is being landed. Those failures remain recorded and are not treated as green checks. All requested tests must pass on the updated PR head before merge.

### DIFF
--- a/docs/web/control-ui/settings.md
+++ b/docs/web/control-ui/settings.md
@@ -321,6 +321,11 @@ or replace conflicts. It reports each source's confirmed copy count and warns
 when a failure may have happened after a partial copy. Use the dedicated Import
 Memory page when you need destination selection, a file preview, or replacement.
 
+If an error says that apply completed but its result could not be returned,
+inspect the migration report and destination files before starting another
+import. Retrying the same pending request reuses its recorded outcome while
+the Gateway retains it. A plugin cleanup warning does not undo completed copies.
+
 Planning and applying require `operator.admin`. Every apply creates a verified
 OpenClaw backup when state exists, writes a redacted migration report, and keeps
 item-level backups before replacing existing destination files. See
--- a/src/commands/migrate.resources.test.ts
+++ b/src/commands/migrate.resources.test.ts
@@ -1,0 +1,405 @@
+import path from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createWizardPrompter } from "../../test/helpers/wizard-prompter.js";
+import { clearRuntimeConfigSnapshot } from "../config/runtime-snapshot.js";
+import { loadAndActivateRootPluginRegistry } from "../plugins/loader.js";
+import { resetPluginLoaderTestStateForTest } from "../plugins/loader.test-fixtures.js";
+import { createMigrationResourceFixture } from "../plugins/migration-provider.test-support.js";
+import { clearPluginMetadataLifecycleCaches } from "../plugins/plugin-metadata-lifecycle.js";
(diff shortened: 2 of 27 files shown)


steipete:
Accepted the optional setup outcome-retention finding. The Memory wizard and advisory discovery can complete their callbacks and then reject on registration cleanup; I am adding native regression coverage and repairing that completion boundary before landing. Genuine acquisition and operation failures will remain failures.

The persisted-table warning is a classification false positive for this diff. These files invoke existing migration providers; the PR adds no SQLite DDL, table, index, schema version, migration, backfill, or stored-data format change. The Gateway outcome cache uses the existing in-memory dedupe map. Native SQLite fixtures exercise resource closure and import replay, without changing the persistent schema contract.

The first CI attempt also failed an unchanged Signal first-message ingress test before any reply callback. Its owners are byte-identical to the parent, while later same-file cases passed. One exact-head failed-job retry is running; that failure is preserved and has not been called a passing check.

steipete:
Merged via squash.

- Prepared head SHA: [3503bcbee5613e8b72abacac64095b55f7fea4e3](https://github.com/openclaw/openclaw/pull/142738/commits/3503bcbee5613e8b72abacac64095b55f7fea4e3)
- Landed commit: [f422bbe2e86508db56eaeec5ab7b34385e4fc8de](https://github.com/openclaw/openclaw/commit/f422bbe2e86508db56eaeec5ab7b34385e4fc8de)

PR #142741, opened 2026-09-09
fix(cron): prevent Gateway hangs on stalled reservations

### LINKED ISSUE #142737: Gateway hangs and retains cron ticks when reservations make no progress
## Summary

A Gateway can become unresponsive and accumulate unfinished cron scheduler invocations when a partial batch obtains execution capacity but cannot reserve any durable jobs, and those same jobs remain due.

## Observed impact

A Windows Gateway running Node 24.18.0 at OpenClaw commit `3e374966f123f8ae3b98693b39d59326daafea7b` experienced an approximately 38-minute event-loop delay. A later full-memory dump contained 595,870 active scheduler ticks despite only seven occupied cron execution slots.

Read-only dump analysis, using matching Node symbols and exact-version V8 layouts, found repeating strong references through child scheduler promises, await contexts, capacity-recheck callbacks, and parent `drain()` promises. The sampled active JavaScript stack was loading and normalizing cron jobs.

The raw dump and operational state are private and are not attached.

## Reproduction

An isolated, file-backed regression reproduces the mechanism without production data:

1. Create two due jobs and leave one execution slot available.
2. Give the first job a live run receipt while its job row remains due.
3. Run one scheduler tick.
4. Durable reservation rejects the first candidate. Releasing the unused slot invokes the capacity listener.
5. The empty batch permits an immediate child tick even though the runnable set has not changed. The parent waits for the child, which repeats the same sequence.

An in-band stop after five nested ticks bounds the failing reproduction because an ordinary timer cannot interrupt a continuously replenished microtask queue.

## Expected behavior

A no-progress reservation attempt should finish and retain normal timer scheduling. A real capacity release or a changed runnable set should still allow pending work to start promptly. Retiring the conflicting receipt should let both jobs complete.

## Relevant owner

`src/cron/service/timer-scheduler.ts` sets `allowEmptyCapacityRecheck` for partially admitted batches. `src/cron/service/timer-capacity-recheck.ts` retains child promises until they finish.

The distinction is between releasing unused capacity and making durable scheduling progress; those are not equivalent.

## Scope

The synthetic reproduction identifies a persistent receipt conflict as one trigger. The dump establishes the runaway retaining chain but does not identify the original reservation predicate that failed.

The incident also had a large SQLite checkpoint backlog. Its reader/connection owner has not been established, so this report does not attribute that entire backlog, or every byte of process memory, to the cron defect.

galiniliev:
Closes #142737

## What Problem This Solves

Fixes an issue where operators running scheduled jobs could lose Gateway responsiveness and see growing memory use when a partially admitted cron batch could not reserve any durable jobs and the same candidates remained due.

A read-only analysis of a Windows full-memory dump found 595,870 unfinished scheduler ticks with only seven occupied execution slots. Strong promise references repeatedly linked child scheduler invocations to parents waiting in capacity-recheck finalization. The raw dump and operational state remain private.

The investigation also reproduced separate SQLite initialization failures that could leave native handles, statement caches, or maintenance timers undisposed. Those failures are not claimed as the cause of the captured cron retaining chain.

## Why This Change Was Made

An unused execution-slot release is no longer treated as sufficient progress to immediately repeat an empty reservation attempt. An empty batch rechecks immediately only when refreshed runnable state shows that a selected candidate is no longer due; unchanged conflicts retire the tick and keep normal timer scheduling. Real capacity releases and concurrent-owner progress retain their immediate wake behavior.

The SQLite change keeps disposal ownership through cache setup, connection pragmas, timeout-wrapper finalization, schema initialization, and final permission hardening. Cleanup attempts independent resources, preserves primary and cleanup errors, and keeps a failed native close available for explicit disposal without publishing it as a healthy cache entry.

Single cleanup failures are rethrown unchanged in both by-path and global retirement. Only multiple failures are aggregated. This preserves actionable top-level messages for QA diagnostics without inspecting or exposing potentially sensitive nested causes.

These are separate commits. No schema, configuration option, dependency, permission-policy, or timeout changes are included.

## User Impact

The Gateway no longer continuously creates and retains scheduler invocations for this no-progress condition. Pending work can recover when its conflicting receipt retires, and existing capacity-release behavior is preserved.

Failed database initialization releases acquired resources instead of silently leaving background maintenance and native handles behind. Failures remain visible rather than being converted into successful opens.

## Evidence

The full-build and isolated-Gateway smoke evidence below predates the rebase. The P2-specific cleanup coverage was rerun on the rebased branch.

- P2 diagnostics follow-up: the existing shared-store QA cleanup case and two global-retirement cases failed before the repair because the actual close reason was hidden. All 22 acquisition, wrapper-retention, and QA artifact-cleanup cases pass afterward, including secret redaction, nested-cause exclusion, and cleanup retry. Focused P0/P1/P2 autoreview found no accepted/actionable issues.
- The file-backed cron regression fails on the original scheduler: an in-band guard observes five nested ticks instead of one. The repaired scheduler retires after one tick, releases Gateway work admission, keeps its timer armed, and completes both jobs after the conflicting receipt retires.
- Cron admission, cross-tick lifecycle, and timer regression suites: 75 passing tests.
- SQLite acquisition, WAL, busy-timeout, and Kysely suites: 103 passing tests. The original acquisition implementation fails 11 owner cases; the original pragma handoff fails three cases.
- Integrated cron, shared/agent permission, closed-wrapper retention, and plugin-state SDK coverage: 93 passing tests across seven files. These overlap the focused groups above; counts are not additive.
- Core production/test typechecking, scoped lint and formatting, database-schema guards, and runtime import-cycle checks passed.
- Independent P0/P1 autoreview of the combined changes: no accepted/actionable findings.
- `pnpm build` passed. The normal CLI also rebuilt the committed candidate before the isolated runtime started.
- An isolated loopback Gateway accepted authenticated RPC, ran a scheduled model-free command, persisted a successful run-history entry, returned ten successful readiness responses, and removed the synthetic job. After startup, readiness reported no failing components and `eventLoop.degraded=false`. The task-owned Gateway was then stopped; no external provider or channel traffic was used.

Local executable coverage uses Linux and Node 24.19.0; the captured Windows process used Node 24.18.0. The dump establishes the runaway promise-retention mechanism, not a complete retained-size accounting of every process allocation.

No live deployment or 48-hour soak is claimed. The incident's separate WAL checkpoint backlog still requires attribution to its reader/connection owner; this PR does not delete WAL files or claim that backlog is independently repaired.

### DIFF
--- a/src/cron/service.cross-tick-admission.test.ts
+++ b/src/cron/service.cross-tick-admission.test.ts
@@ -308,6 +308,77 @@ describe("cron service cross-tick bounded admission", () => {
     stop(state);
   });
 
+  it("retires an empty tick when a receipt conflict leaves the same jobs due", async () => {
+    const store = fixtures.makeStorePath();
+    const t0 = Date.parse("2026-02-06T10:07:00.000Z");
+    const conflicted = createDueIsolatedJob({
+      id: "unchanged-conflict",
+      nowMs: t0,
+      nextRunAtMs: t0,
+    });
+    const pending = createDueIsolatedJob({
+      id: "after-unchanged-conflict",
+      nowMs: t0,
+      nextRunAtMs: t0,
+    });
+    await saveCronStore(store.storePath, { version: 1, jobs: [conflicted, pending] });
+    const prepared = prepareCronRunReceiptClaim({
+      storePath: store.storePath,
+      job: conflicted,
+      agentId: conflicted.agentId ?? "main",
+      startedAtMs: t0,
+    });
+    const receipt = runOpenClawStateWriteTransaction(({ db }) =>
+      claimCronRunReceiptInDatabase({
+        database: db,
+        prepared,
+        resolveAgentId: (job) => job.agentId ?? "main",
+      }),
+    );
+    let peakTicks = 0;
+    const runIsolatedAgentJob = vi.fn(async () => ({ status: "ok" as const }));
+    const state: ReturnType<typeof createCronRegressionState> = createCronRegressionState({
+      storePath: store.storePath,
+      nowMs: () => {
+        peakTicks = Math.max(peakTicks, state.activeTimerTicks);
+        // A timer cannot stop a microtask livelock; bound the pre-fix failure here.
+        if (peakTicks >= 5) {
+          stop(state);
+        }
+        return t0;
+      },
+      runIsolatedAgentJob,
+    });
+    state.runAdmission.active = DEFAULT_CRON_MAX_CONCURRENT_RUNS - 1;
+
+    try {
+      await onTimer(state);
+
+      expect(peakTicks).toBe(1);
+      expect(state.stopped).toBe(false);
+      expect(state.activeTimerTicks).toBe(0);
+      expect(getActiveGatewayRootWorkCount()).toBe(0);
+      expect(state.runAdmission.active).toBe(DEFAULT_CRON_MAX_CONCURRENT_RUNS - 1);
+      expect(state.queuedRunReservationsByJobId.size).toBe(0);
+      expect(state.timer).not.toBeNull();
+      expect(runIsolatedAgentJob).not.toHaveBeenCalled();
+
+      finishCronRunReceipt({ handle: receipt, status: "skipped", finishedAtMs: t0 });
+      await onTimer(state);
+
+      expect(runIsolatedAgentJob).toHaveBeenCalledTimes(2);
+      expect(state.activeTimerTicks).toBe(0);
+      expect(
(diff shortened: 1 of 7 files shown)


galiniliev:
Merged via squash.

- Prepared head SHA: [7be9a01de66d4ef523ab0a588bb422edd2d1954f](https://github.com/openclaw/openclaw/pull/142741/commits/7be9a01de66d4ef523ab0a588bb422edd2d1954f)
- Landed commit: [6812dd0f93e807ed92a5ce87bb47d58786a66ef0](https://github.com/openclaw/openclaw/commit/6812dd0f93e807ed92a5ce87bb47d58786a66ef0)

PR #142746, opened 2026-09-09
fix(agents): preserve HTTP conversations with uppercase keys

### LINKED ISSUE #142616: [Bug] openai-compat /v1/chat/completions: session keys containing ULID-shaped tokens are permanently fenced from the second turn (SessionWorkStartChangedError surfaced as opaque 500)
### Bug type

### Beta release blocker

No

### Summary

On 2026.9.2/2026.9.3, a client that supplies its own session key via the `x-openclaw-session-key` header to the OpenAI-compatible `/v1/chat/completions` surface gets exactly **one** successful turn when the key contains a **ULID-shaped token** (26-char Crockford Base32) in the tail segment. Every subsequent request on the same session key fails with `SessionWorkStartChangedError` ("Session … changed while starting work. **Retry.**") — visible **only in the gateway journal** — while the HTTP response is an opaque `500 {"error":{"message":"internal error","type":"api_error"}}`. The session stays fenced permanently: client retries seconds or minutes later keep failing, so the error's own "Retry." guidance cannot succeed, and clients cannot even detect the condition because the response body is generic.

### Steps to reproduce

Self-hosted gateway, more than one agent configured, any agent id (`$AGENT`), gateway token (`$TOKEN`):

```bash
# Turn 1 — succeeds and creates the session
curl -sS -X POST http://127.0.0.1:18789/v1/chat/completions \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -H "x-openclaw-agent-id: $AGENT" \
  -H "x-openclaw-session-key: agent:$AGENT:assist:01M21F31SCNCCQQ3N4X43AY420" \
  -d '{"messages":[{"role":"user","content":"first question"}],"stream":false}'

# Turn 2 — same key, 2+ seconds (or minutes) later — fails
curl -sS -X POST http://127.0.0.1:18789/v1/chat/completions \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -H "x-openclaw-agent-id: $AGENT" \
  -H "x-openclaw-session-key: agent:$AGENT:assist:01M21F31SCNCCQQ3N4X43AY420" \
  -d '{"messages":[{"role":"user","content":"second question"}],"stream":false}'
```

### Expected behavior

- Client-chosen session keys with arbitrary (non-reserved) tails behave consistently across turns; a session admitted once remains usable for later turns.
- If a recoverable admission race exists, the HTTP response should be **typed** (the journal message itself says "Retry.") instead of a generic `internal error` 500, so clients can implement the suggested retry.

### Actual behavior

Turn 2 (and every later request on that key) returns:

```json
{"error":{"message":"internal error","type":"api_error"}}
```

Gateway journal:

```
[openai-compat] chat completion failed: SessionWorkStartChangedError: Session "agent:$AGENT:assist:01M21F31SCNCCQQ3N4X43AY420" changed while starting work. Retry.
```

The fence never clears for that key. We tested retries at 3/6/9 s backoff and again minutes later — all fenced. Only switching to a different session key recovers (for one turn, if that key also embeds a ULID-shaped token).

### Evidence matrix

All via curl with identical parameters, 2–3 s between turns, on 2026.9.3:

| Session key (after `agent:$AGENT:`) | Turn 1 | Turn 2 |
|---|---|---|
| `assist:repro7` (letters) | pass | pass (6+ rapid turns pass) |
| `assist:ha_02f029d8dba67eb83d1e` (20-char hex hash) | pass | pass |
| `assist:01M21F31SCNCCQQ3N4X43AY420` (bare ULID) | pass | **fenced** |
| `assist:ha-01M21F31SC7GJXGWHZ987KV0KN` (prefixed ULID) | pass | **fenced** |
| `01M21EHQSCQR7QVDVCDG3EBCFC` (bare ULID directly after agent segment) | pass | **fenced** |

So a ULID-shaped token inside the tail segment triggers it — even with a scope segment (`assist:`) and a non-Crockford prefix (`ha-`) around it. We did not map the exact detection boundary (e.g., whether a ULID substring anywhere in the key triggers it); the matrix is what we verified.

Since `agent:{agentId}:{sessionId}` is OpenClaw's internal canonical session key form, this looks like an **id-vs-key collision**: the `assertAllowed` re-load inside `beginSessionWorkAdmission` (`latestEntry.sessionId !== sessionId`) appears to disagree with the initial `loadGatewaySessionEntry` result deterministically whenever the client key embeds a ULID-shaped token.

### Impact

This breaks any client that passes ULID-shaped conversation ids as session keys. Concretely: our Home Assistant voice-satellite setup (assist satellite addon → OpenClaw custom integration → gateway) generates ULID conversation ids, so on 2026.9.2 and 2026.9.3 the voice pipeline answers the **first** question and 500s on **every follow-up**. That is what forced us to roll back 2026.9.2 earlier; to run 2026.9.3 we now deterministically hash ULID-shaped conversation ids client-side before building the session key (workaround, happy to share details). The same integration ran unchanged against 2026.8.1 for weeks without this failure.

### OpenClaw version

2026.9.3 (1391f7c) — same behavior observed on 2026.9.2.

### Operating system

Linux (self-hosted gateway)

### Install method

Official release package (npm)

### Model

Agent primary is a custom `openai-completions` proxy model, but the reproduction is model-independent: the fenced request never reaches a model call.

### Logs

Redacted journal lines included above; the HTTP response body is the generic `internal error` quoted in "Actual behavior". Can provide more if useful.

ruel225:
## What Problem This Solves

Fixes an issue where applications using uppercase conversation keys could send one chat completion, then receive HTTP 500 on every follow-up.

Closes #142616.

## Why This Change Was Made

Command preparation now uses the existing canonical store key for both hidden-session exclusion and exact lookup. The original request key remains available for routing, and existing case-sensitive peer identities remain intact.

## User Impact

Applications can continue the same conversation without replacing or lowercasing their keys. Internal run-owned sessions remain inaccessible through alternate-case keys.

## Evidence

- Real Gateway baseline: first HTTP turn succeeded; the second failed with a session identity mismatch while the original persisted row remained intact.
- Repaired Gateway: 33 successful HTTP turns across 11 conversations, including separate agents and case-sensitive peer keys. The hidden-session probe started no model work and left its row unchanged.
- 120 focused tests passed. The added hidden-session regression failed against the incoming patch and passed after correction.
- Targeted formatting, lint, runtime build, and fresh code review passed.

Original continuation repair by @ruel225. Raw fixture data and credentials remain private.

Observed HTTP results, with fixture identifiers removed:

```text
Uppercase conversation, baseline:  200 -> 500
Uppercase conversation, repaired:  200 -> 200 -> 200
Hidden conversation probe:        rejected; no model request; row unchanged
```

Independent acceptance confirmed complete prior conversation inputs, separate case-sensitive conversations, incognito behavior, command history, and a working reset followed by a fresh turn. The hosted [CI run](https://github.com/openclaw/openclaw/actions/runs/34301715119) checks the published revision.

### DIFF
--- a/src/agents/command/session.exact-target.test.ts
+++ b/src/agents/command/session.exact-target.test.ts
@@ -79,6 +79,32 @@ it.each(["work", "dashboard:incognito-work"])(
   },
 );
 
+it("reuses the persisted session when an explicit key's opaque tail is uppercase", async () => {
+  await withOpenClawTestState({ label: "command-uppercase-tail-session" }, async (state) => {
+    const storePath = state.path("sessions.sqlite");
+    // Client-generated opaque tails (a ULID, for example) carry uppercase, while
+    // persisted store rows always hold the canonical store-folded key.
+    const sessionKey = "agent:main:assist:01M21F31SCNCCQQ3N4X43AY420";
+    const cfg = {
+      agents: { defaults: {} },
+      session: { store: storePath, reset: { mode: "idle", idleMinutes: 60 } },
+    } satisfies OpenClawConfig;
+
+    const first = resolveSession({ cfg, sessionKey });
+    expect(first.sessionEntry).toBeUndefined();
+    expect(first.isNewSession).toBe(true);
+    await sessionAccessor.replaceSessionEntry(
+      { sessionKey: first.sessionKey, storePath },
+      { sessionId: first.sessionId, updatedAt: Date.now(), sessionStartedAt: Date.now() },
+    );
+
+    const second = resolveSession({ cfg, sessionKey });
+    expect(second.sessionId).toBe(first.sessionId);
+    expect(second.isNewSession).toBe(false);
+    expect(second.sessionEntry?.sessionId).toBe(first.sessionId);
+  });
+});
+
 it("does not provision a missing incognito lookup or select a hidden run-owned entry", async () => {
   await withOpenClawTestState({ label: "command-private-session" }, async (state) => {
     const storePath = state.path("sessions.sqlite");
--- a/src/agents/command/session.exact-target.test.ts
+++ b/src/agents/command/session.exact-target.test.ts
@@ -79,12 +79,17 @@ it.each(["work", "dashboard:incognito-work"])(
   },
 );
 
-it("reuses the persisted session when an explicit key's opaque tail is uppercase", async () => {
+it.each([
+  "agent:main:assist:01M21F31SCNCCQQ3N4X43AY420",
+  "agent:main:assist:prefix-01M21F31SCNCCQQ3N4X43AY420",
+  "agent:main:assist:lowercaseletters",
+  "agent:main:assist:abcdef0123456789",
+  "agent:main:assist:ordinary-42",
+  "agent:main:signal:group:AbCdEf123",
+  "agent:main:matrix:channel:!Room:Example.org:thread:$Event",
+])("reuses the persisted session for explicit key %s", async (sessionKey) => {
   await withOpenClawTestState({ label: "command-uppercase-tail-session" }, async (state) => {
     const storePath = state.path("sessions.sqlite");
-    // Client-generated opaque tails (a ULID, for example) carry uppercase, while
-    // persisted store rows always hold the canonical store-folded key.
-    const sessionKey = "agent:main:assist:01M21F31SCNCCQQ3N4X43AY420";
     const cfg = {
       agents: { defaults: {} },
       session: { store: storePath, reset: { mode: "idle", idleMinutes: 60 } },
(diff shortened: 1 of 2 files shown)


obviyus:
Thanks @ruel225!

Fixed HTTP conversation continuation for uppercase application-owned keys. Command preparation now uses the existing canonical store key for both hidden-session exclusion and exact lookup, while preserving the original routing key and case-sensitive peer identities.

- Before: the first HTTP turn succeeded, but its follow-up failed with a session identity mismatch.
- After: 33 HTTP turns across 11 conversations succeeded with stable sessions and complete preceding conversation history.
- Independent acceptance covered hidden-session exclusion, agent isolation, case-sensitive peers, incognito conversations, history, and reset recovery.
- All 120 focused tests passed. Required CI passed on the merged PR head in [run 34331567522](https://github.com/openclaw/openclaw/actions/runs/34331567522).

The final head contains an empty contributor commit over the runtime-tested tree; source and proof equivalence were verified before the fresh final review.

PR #142751, opened 2026-09-09
fix(sessions): show the correct inherited model across views

steipete:
## What Problem This Solves

A child session's inherited model could differ between agent-scoped lists, All-agent lists, detail reads, history metadata, and session change events. Lists could borrow another agent's visible `global` parent or miss a qualified parent in another store; exact reads could miss the parent or reject a valid child in a shared store.

## Why This Change Was Made

Session projection retains the original entry and its owning store separately from display rows. List rendering and model search share that source. Exact reads carry the admitted physical address through selection, batching, and request caches; scoped lists reuse the existing exact reader for qualified parents outside their prepared stores. Raw parents and known misses keep their original ownership.

Model policy remains in the existing resolver, including direct pins, explicit Default, native runtime ownership, and fallback exclusions. Duplicate model-resolution and child-index work is removed. Removing an upward Gateway export from the config facade lets the readers share canonical routing without an import cycle. The net 197 production lines retain previously discarded ownership; no schema, configuration, public SDK, persistent cache, or broad foreign-store scan is added.

## User Impact

Inherited model metadata agrees across session list, detail, history, event, and embedded TUI views, including multi-agent and shared-store layouts. Model execution policy is unchanged.

## Evidence

Regressions use isolated SQLite state and real create, patch, list, describe, and history handlers. Creating an unpinned child before pinning its parent reproduces incorrect All-agent and scoped-list results; the shared-store case reproduces failed exact reads. Direct selection, explicit Default, qualified aliases, missing parents, search, cached/batched reads, sharing, startup and SDK-wrapper controls are covered.

The broader 256-case suite passed. After a local-identifier-only cleanup, 37 affected cases and the full changed-file gate passed on the exact final source; those counts are not additive. A clean Linux checkout also passed frozen installation, the complete `pnpm build`, and `pnpm plugins:assets:check`, with all tracked source bytes unchanged. Fresh independent review completed both passes with no actionable findings.

These are synthetic handler/SQLite tests and compilation checks; no live model-provider or operator Gateway run is claimed. Required [hosted CI](https://github.com/openclaw/openclaw/actions/runs/34313500321) passed on the final commit. Its first attempt failed in an unchanged one-second observation-cap fixture; the same tooling shard passed on one diagnostic rerun. The timing-sensitive fixture remains a separate follow-up, not a fix claimed by this PR.


steipete:
Merged via squash.

- Prepared head SHA: [0e8201fdbef6d2595845741b143161409e28bd11](https://github.com/openclaw/openclaw/pull/142751/commits/0e8201fdbef6d2595845741b143161409e28bd11)
- Landed commit: [89168d55ee1932072c50a1699c6529e5092e11db](https://github.com/openclaw/openclaw/commit/89168d55ee1932072c50a1699c6529e5092e11db)

Issue #142718, opened 2026-09-09
[Bug]: registerGatewayMethod return-value support is still rejected by the exported handler type


aeoess:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

#85785 added return-value completion for plugin gateway methods, but `registerGatewayMethod` still types the handler as `GatewayRequestHandler`, whose return type is `void | Promise<void>`.
So the handler style added by that fix works at runtime and still fails TypeScript.

### Steps to reproduce

1. register a gateway method:

```ts
api.registerGatewayMethod("demo.check", async () => {
  return { ok: true };
});
```

2. compile the plugin against `openclaw@2026.9.3`
3. TypeScript rejects the handler because `Promise<{ ok: boolean }>` isn't assignable to `void | Promise<void>`
4. cast around the type error, load the same plugin and call `demo.check`
5. the Gateway returns:

```json
{ "ok": true }
```

### Expected behavior

The exported plugin type accepts the return-value completion style added in #85785.

### Actual behavior

The runtime turns the returned value into a successful Gateway response, but the exported handler type rejects the same code at compile time.

### OpenClaw version

2026.9.3

### Operating system

macOS 15

### Install method

npm, with `openclaw` pinned as a dependency

### Model

N/A, no model involved

### Provider / routing chain

N/A, no provider request involved

### Logs

```shell
error TS2345: Argument of type '() => Promise<{ ok: boolean; }>' is not assignable to parameter of type 'GatewayRequestHandler'.
  Type 'Promise<{ ok: boolean; }>' is not assignable to type 'void | Promise<void>'.
```

### Impact and severity

Affected: TypeScript plugin authors using return-value completion with `registerGatewayMethod`
Severity: Low, it blocks compilation for that handler shape, and `respond(...)` is a straightforward workaround
Frequency: Always for the reproduced handler shape
Consequence: a plugin gateway handler style supported by the runtime can't be written against the exported type without a cast

### Additional information

Found while moving our plugin off local OpenClaw type mirrors and onto the exported host types.

#85785 intentionally added return-value completion for plugin gateway methods, with regression coverage. A pre-merge ClawSweeper comment on that PR also noted that the public handler type still returned `Promise<void> | void` and recommended aligning the SDK type and docs. The runtime change merged; the exported type is still unchanged at `b077653f6a`.

PR #142756, opened 2026-09-09
fix(plugin-sdk): allow gateway handlers to return responses

zhangguiping-xydt:
## What Problem This Solves

Issue #142718: `registerGatewayMethod` already supports returning a value that becomes the Gateway response, but the exported plugin handler type describes handlers as returning only `void`. TypeScript plugin authors cannot express the supported return-value contract cleanly.

## Why This Change Was Made

The plugin gateway adapter has converted non-`undefined` handler results into successful responses since #85785. The public plugin registration input is the remaining stale contract. This change widens only that plugin-facing input type and keeps core Gateway handlers callback-only.

## User Impact

TypeScript plugins can return a Gateway response from `registerGatewayMethod` without a cast. Existing core handlers, explicit `respond(...)` handlers, and runtime dispatch behavior are unchanged.

## Evidence

- `PluginGatewayRequestHandler` is used at the plugin registration boundary; the adapter converts its returned value and exposes a core `GatewayRequestHandler` afterward.
- Core `GatewayRequestHandler` remains `Promise<void> | void`; the plugin SDK exports both public handler contracts from `gateway-runtime`.
- The Workboard shared result wrapper now explicitly returns `void`, preserving no-return callback inference under the widened plugin input.
- `pnpm tsgo:core` passed.
- `pnpm tsgo:extensions` passed.
- `pnpm tsgo:test:root` passed.
- Exact-head runtime contract test under Node 24.18.0 passed: `src/plugins/contracts/scheduled-turns.contract.test.ts` (26 tests), including real plugin registration lifecycle coverage and returned Gateway method behavior.
- `git diff --check` passed.

## Added exact-head proof (44e7bcdc96b)

- Added `test/e2e/qa-lab/runtime/plugin-gateway-return.e2e.test.ts`.
- The fixture is a TypeScript plugin whose `register(api)` calls `api.registerGatewayMethod("ts-gateway-return.echo", async () => ({ ok: true, source: "typescript" }))` with no cast and no explicit `respond` call.
- The test starts an isolated Gateway with the plugin loaded through `plugins.load.paths`, connects with the real `GatewayClient`, invokes `ts-gateway-return.echo`, and asserts the RPC payload is `{ ok: true, source: "typescript" }`.
- This exercises the public plugin registration type, the plugin loader, registration adapter, authenticated Gateway dispatch, and the client response path together.

## Public handler-union compatibility

The widened `PluginGatewayRequestHandler` is confined to the plugin registration boundary. Existing callback handlers remain valid because the union retains the original `(opts) => void | Promise<void>` shape; the adapter still supplies the same `respond` callback and preserves explicit responses. Core `GatewayRequestHandler` remains callback-only, so core dispatch does not acquire an implicit-return contract. The new test covers the cast-free implicit-return branch, while the existing contract fixtures cover explicit-response and callback-only handlers. Contextual typing can expose a downstream plugin that depended on an inferred callback-only return type, so this is an additive public-input compatibility risk limited to TypeScript plugin authors; runtime dispatch semantics and existing callback registrations are unchanged.

## CI follow-up on exact head

The previous CI run exposed two contract fallout errors from the widened plugin handler type: Workboard callbacks had inconsistent return paths (`extensions/workboard/src/gateway.ts` and `gateway-helpers.ts`). They now return one canonical Promise path while preserving explicit `respond` behavior. The plugin SDK surface budget was refreshed from 4,353 to 4,355 to account for the two intentional public type exports; unrelated tooling shard failures were rechecked against the exact head.

## Follow-up fixes (975638accf2)

- Wrapped Workboard handler invocation in a deferred Promise boundary so synchronous validation throws and rejected promises both reach `respondError`.
- Closed the Gateway server in the RPC proof test before restoring the test environment.
- Focused checks: `oxlint` passed for touched files; `plugin-sdk-surface-report.mts --check` passed; Workboard gateway tests were run with Node 24.15.0 (the repository-required SQLite-safe runtime).

## Follow-up proof and fixes (93b878309f4)

- Workboard handler invocation is deferred before `Promise.resolve`, so synchronous validation throws and rejected promises share the `respondError` boundary.
- The Gateway RPC proof closes the returned server before restoring the test environment.
- The proof plugin is now a committed TypeScript fixture importing `OpenClawPluginApi` from the public `openclaw/plugin-sdk/plugin-entry` entrypoint. `pnpm tsgo:test:root` passes with the fixture, and the E2E test loads that same TypeScript entry through the real plugin loader and invokes it through `GatewayClient`.

### DIFF
--- a/extensions/workboard/src/gateway-helpers.ts
+++ b/extensions/workboard/src/gateway-helpers.ts
@@ -41,10 +41,12 @@ export function registerWorkboardResultMethods(
     api.registerGatewayMethod(
       method,
       (context) =>
-        Promise.resolve(handler(context)).then(
-          (result) => context.respond(true, result),
-          (error: unknown) => respondError(context.respond, error),
-        ),
+        Promise.resolve()
+          .then(() => handler(context))
+          .then(
+            (result) => context.respond(true, result),
+            (error: unknown) => respondError(context.respond, error),
+          ),
       { scope },
     );
   }
--- a/extensions/workboard/src/gateway-helpers.ts
+++ b/extensions/workboard/src/gateway-helpers.ts
@@ -1,7 +1,6 @@
 import { WORKBOARD_STATUSES, type WorkboardCard } from "@openclaw/workboard-contract";
 // Workboard plugin module implements shared gateway request helpers.
 import { formatErrorMessage } from "openclaw/plugin-sdk/error-runtime";
-import type { GatewayRequestHandler } from "openclaw/plugin-sdk/gateway-runtime";
 import { parseStrictPositiveInteger } from "openclaw/plugin-sdk/number-runtime";
 import type { OpenClawPluginApi } from "../api.js";
 import {
@@ -20,9 +19,7 @@ export type GatewayMethodContext = Parameters<
   Parameters<OpenClawPluginApi["registerGatewayMethod"]>[1]
 >[0];
 type GatewayRespond = GatewayMethodContext["respond"];
-type WorkboardGatewayResultHandler =
-  | GatewayRequestHandler
-  | ((context: GatewayMethodContext) => Promise<unknown>);
+type WorkboardGatewayResultHandler = (context: GatewayMethodContext) => Promise<unknown> | void;
 type WorkboardGatewayScope = NonNullable<
   NonNullable<Parameters<OpenClawPluginApi["registerGatewayMethod"]>[2]>["scope"]
 >;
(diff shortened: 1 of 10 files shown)


zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
Closing this contributor PR per request. Future daily-fix candidates will avoid SDK-related topics.

PR #142766, opened 2026-09-09
fix(models): retain captured rows in passive catalogs

obviyus:
## What Problem This Solves

Gateway startup could omit an authored model from passive model lists in merge mode even though its saved row remained present.

## Why This Change Was Made

Retain captured rows after configured rows so explicit settings win. Keep startup retention separate from Refresh membership, and share metadata conversion using the exact donor row. Replace mode still excludes unconfigured rows.

## User Impact

Authored merge-mode models remain visible at startup without restoring removed models during fresh discovery.

## Evidence

Compiled Gateway merge/replace checks reproduced the missing row before the fix and passed afterward, preserving explicit fields and saved data. The captured-registry-to-empty-refresh regression failed before the correction and passed afterward; 14 affected checks and required CI passed. Empty Refresh was verified by the owner regression, not a successful public restart flow. Persistent inventory across restart and inference were not established.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -1848,7 +1848,6 @@ src/agents/memory-write-provenance.ts	1
 src/agents/minimax-vlm.ts	1
 src/agents/model-auth-label.ts	1
 src/agents/model-auth-model.ts	2
-src/agents/model-catalog.ts	1
 src/agents/model-fallback-attempt.ts	1
 src/agents/model-fallback-runner.ts	2
 src/agents/model-provider-auth.ts	1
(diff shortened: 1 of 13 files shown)


obviyus:
Landed in 1c576b0a0fa0acd56eba18bcb4ae15064f27b9f0.

**Proof:** Real Gateway startup and authenticated `models.list` show the retained authored row in merge mode and exclude it in replace mode. Explicit selected fields and authored files are preserved, and all recorded Gateway instances stop cleanly.

| Before | After |
| --- | --- |
| Passive merge omits the existing unconfigured authored row. | Passive merge retains it after configured rows. |
| Replace excludes the unconfigured row. | Replace keeps that exclusion and explicit configured settings. |

The captured-to-empty-refresh owner regression failed before the correction and passes afterward. Fresh independent acceptance passes all five public clauses; 14 affected correction checks, exact-head code review and [CI 34302730148](https://github.com/openclaw/openclaw/actions/runs/34302730148) pass. Valid unchanged evidence from the initial 75 controls is retained.

The additional public restart experiment did not reach empty Refresh because its seed was an in-memory Gateway plan. That failed setup and write-reserve stop remain recorded; no public empty-refresh, cross-restart persistence, inference or native-account proof is claimed.

Issue #138561, opened 2026-09-04
[Bug]: active-memory plugin stops recalling after 2026.8.2 upgrade, with no signal at info log level


Oliverbot26:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

After upgrading to 2026.8.2, the bundled `active-memory` plugin stopped performing recall entirely (0 recall transcripts across two days, with agent turns still running), and the stop produces no output at `info` level, so it is indistinguishable from "no relevant memory found".

### Steps to reproduce

Grounded on direct observation of one installation; the exact trigger is not established.

1. Run 2026.8.x with the bundled `active-memory` plugin enabled and `persistTranscripts: true`.
2. Run agent turns on the configured agent (here: 11 turns on the day of the upgrade, both `user` and `heartbeat` triggered).
3. Count the files written by the plugin per day:
   `find ~/.openclaw/plugins/active-memory/transcripts/agents/<agent>/active-memory -type f -newermt <day>`
4. Observe that the count drops to 0 from the moment the 2026.8.2 gateway starts, and stays 0.

### Expected behavior

Prior observed behavior on the same installation, same config, same paths:

```
2026-09-02 (full day, 8.1)   51 recall transcripts
2026-09-03 up to 11:55:26    105 recall transcripts
```

### Actual behavior

```
2026-09-03 after 15:36 (first 2026.8.2 gateway start)   0     ← 11 agent turns that same day
2026-09-04 (full day)                                   0
```

Confirmed by four independent channels over the same window:

- `0` lines matching `active-memory:` in `gateway.log`
- `0` occurrences of the plugin's context header (`ACTIVE_MEMORY_CONTEXT_HEADER`, `"Context:"`) in the submitted prompts of 11 turns
- `0` sessions named `active-memory-*` in the session store (the recall subagent session id prefix)
- no plugin status line on the session entry

The transcript directory did not move in 8.2 (`resolvePersistentTranscriptBaseDir` → `resolveStateDir()`), so this is an absence of writes, not a relocated artifact.

**The part that is a defect on its own, independently of the cause:** every early-exit path in the hook is `api.logger.debug?.(…)`. At `info`, these two states are indistinguishable:

```
"there was no relevant memory to recall"   ← normal operation
"the plugin did not run at all"            ← what happened here, for two days
```

[...]

PR #142767, opened 2026-09-09
fix(active-memory): show skipped recall when logging is enabled

FabianJun:
Related: #138561

## What Problem This Solves

Fixes an issue where operators who enable Active Memory logging receive no
info-level outcome when an ordinary turn is intentionally skipped because it has
no recall intent. This makes a healthy skip indistinguishable from a missing
plugin invocation at the configured log level.

## Why This Change Was Made

The existing bounded skip result now follows the plugin's established `logging`
setting: it is emitted at info level when logging is enabled and remains a debug
message otherwise. The broader upgrade-specific recall report remains outside
this focused observability repair.

## User Impact

Operators with Active Memory logging enabled can now see that recall ran far
enough to make an intentional `no-recall-intent` decision. Default installations
retain the existing quiet debug-level behavior.

## Evidence

- The focused regression assertion failed against the original production code
  because no info-level skip message was emitted.
- The focused regression passes with the fix (1 passed, 207 skipped).
- The complete Active Memory test file passed 207 of 208 tests. The remaining
  failure is an unrelated existing Windows-only fixture expecting a POSIX `/tmp`
  path while the runtime returns the equivalent `C:\tmp` path.
- Targeted `oxlint`, formatting, extension typechecks, test typechecks, plugin
  boundary checks, doctor contract tests (6 passed), and `git diff --check`
  passed.
- Independent `gpt-6-astra` high-reasoning review found no actionable P0-P3
  findings in commit `b21f1f6ab8`.

### DIFF
--- a/extensions/active-memory/index.test.ts
+++ b/extensions/active-memory/index.test.ts
@@ -1700,8 +1700,8 @@ describe("active-memory plugin", () => {
     },
   );
 
-  it("records why default escalation skips an ordinary turn", async () => {
-    registerPluginConfig({ mode: undefined });
+  it("logs why default escalation skips an ordinary turn when logging is enabled", async () => {
+    registerPluginConfig({ mode: undefined, logging: true });
 
     const result = await runPromptBuild(
       { prompt: "Explain the current configuration" },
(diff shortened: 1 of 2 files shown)


FabianJun:
Withdrawing this PR because the required after-fix proof is an eligible real
Gateway turn, and I cannot produce that evidence from the available isolated
environment without relying on an unapproved live model/service configuration.

The focused source fix, regression test, Astra review, and CI checks completed so
far found no patch defect. However, keeping a PR open without a path to its stated
merge gate would waste maintainer time. The broader issue remains open for someone
with the required isolated real-runtime proof setup.

PR #136713, opened 2026-09-02
fix: gateway restart hangs while followup drain retries a draining error

ericcaiwx-star:
Closes #136684

## What Problem This Solves

Fixes an issue where operators restarting a Gateway with a queued follow-up could wait several minutes for shutdown. The queued turn retried after restart admission rejected it, so the old Gateway stayed active until its shutdown deadline.

## Why This Change Was Made

The follow-up queue now follows the existing one-way restart signal. When restart drain commits, it retires process-local queue work and its old callback before the Gateway inspects active work. A reversible restart-signal fence still waits for settlement and retries after rollback.

Durable accepted-input recovery remains the restart owner. The old in-memory queue and execution authority do not cross restart, as documented in `docs/gateway/restart-recovery.md`.

## User Impact

A Gateway with a queued follow-up can finish a supervised restart on the normal shutdown path instead of emitting a retry burst until forced termination. If restart signal admission rolls back, the queued follow-up resumes normally.

## Evidence

Current-main red at `3100ba535f1d003d67a6a7536deacf8666367b38`:

- Blacksmith child-Gateway run: https://github.com/openclaw/openclaw/actions/runs/33711422936
- One held model turn, one confirmed queued same-session follow-up, then `SIGTERM`.
- Queued case reached the real follow-up admission path and emitted `7` `followup queue drain failed` events.
- No-queue control emitted `0` drain failures.

Exact repaired tree at `2baff4f9ed3f84dc9493149dcc20fed441c9baea`:

- Blacksmith exact-tree run: https://github.com/openclaw/openclaw/actions/runs/33714572460
- The same child-Gateway scenario emitted no retry burst and exited in `3.465s`.
- `src/auto-reply/reply/queue.drain-restart.test.ts`: `29` passed.
- `test/gateway-queued-session-rotation.e2e.test.ts`: `2` passed.
- Materialized source hashes matched the local commit on https://github.com/openclaw/openclaw/actions/runs/33715159753.
- Type-aware Oxlint passed on all touched files.
- Oxfmt and `git diff --check` passed.

The focused tests also cover:

- a typed draining error while restart admission stays open;
- restart-signal rollback and queue resume;
- restart-signal promotion to one-way drain;
- synchronous retirement of queued turn ownership and stale callbacks.

Production LOC: `+33/-2` (net `+31`). The growth binds queue authority to the existing restart signal and closes the user-visible shutdown lifecycle gap. Tests: `+316/-0`.

## AI Assistance

[...]

Issue #136684, opened 2026-09-02
[Bug]: followup queue drain reschedules itself on GatewayDrainingError, prolonging restart shutdown (563 retries / 4m46s observed)


potterdigital:
## Bug type

Behavior bug (shutdown / restart path)

## Beta release blocker

No

## Summary

When the followup queue drain throws a non-deferred error, `scheduleFollowupDrain` reschedules itself unconditionally, so a `GatewayDrainingError` raised during restart drain produces an unbounded self-reschedule loop that keeps the old process alive; one observed restart spun 563 times over 4m46s and took 6m27.9s to complete instead of the ~33.5s measured on the same box the same day.

## Steps to reproduce

1. Run a Gateway with a chat channel attached (observed on Slack, socket mode).
2. Send a message so an item is sitting in that session's followup queue (`FOLLOWUP_QUEUES` entry with `queue.items.length > 0`) and the drain is running.
3. While that item is still queued, trigger a supervised restart (`systemctl --user restart openclaw-gateway`, i.e. SIGTERM → `markGatewayDraining()`).
4. Watch the gateway log. The drain attempt fails with `GatewayDrainingError`, the item stays in `queue.items`, and the drain immediately reschedules itself. This repeats until the process is killed.

## Expected behavior

`GatewayDrainingError` raised by restart drain is terminal for the remaining lifetime of the process, so retrying it cannot succeed. In `src/process/gateway-work-admission.ts` @ `54592465`, `markGatewayRestartDraining()` (`:250`) sets `restartDraining = true` and describes it in-comment as *"the one-way close"* (`:254-255`); it early-returns if already set, and the only thing that clears the flag is `resetGatewayWorkAdmission()`, documented at `:576` as *"Clears restart/suspend admission during SIGUSR1 and isolated tests."* — not the SIGTERM path. `runWithGatewayIndependentRootWorkAdmission` then throws unconditionally while that flag is set (`:399-401`), which is the throw the drain is catching.

A drain that fails for that reason should stop rescheduling and let shutdown proceed, leaving the item queued for whatever persistence/replay policy applies.

Grounded reference: the same gateway, same build, same host, restarted twice earlier the same day with no conversation in flight and completed in **33.5s** and **33.9s** (SIGTERM → `[gateway] ready`).

## Actual behavior

The drain reschedules on every non-deferred error, including `GatewayDrainingError`, and because the failed drain never removes the item, `hasPendingQueueWork` stays true forever.

[...]

PR #142064, opened 2026-09-08
fix(channels): stop ingress retry loops during gateway suspension

goutamadwant:
Related: #142013. This fixes the durable-ingress retry storm; parked-session drain accounting remains separate.

## What Problem This Solves

Queued channel messages could repeatedly enter delivery while Gateway suspension rejected new work.

## Why This Change Was Made

The shared ingress monitor now retains them without starting the pump while suspension admission is closed, then requests a drain on the existing reopen notification. It also checks suspension after an awaited claim so that a racing claim is released without spending an attempt. Existing monitor types remain available through their original exports.

## User Impact

Incoming messages stay queued while accepted work settles, then resume delivery when Gateway admission opens.

## Evidence

Real Discord/Gateway proof used a leased driver and tested bot with a deterministic provider fixture. One already-admitted chat turn stayed active while a real guild mention entered the durable queue during suspension.

- Baseline `21ad735be2f44143bcd2f53ec2ffa8c128e5ede2`: the target row was claimed under two different tokens during the sampled suspension window, and logs show repeated `GatewayDrainingError` for that same message. Attempts stayed at zero, so attempt counters alone did not expose the repeated work.
- Candidate product `5a5f66abac0e33b918bb1bdd541788f3afdd957e`: all 30 suspended row samples stayed pending, unclaimed, and unchanged. No pump or delivery execution appeared in the closed-phase profiler collection; the same modules executed after resume. The prior active turn settled, suspension reached ready, and explicit resume produced one completed row and one visible reply.
- A fresh source-blind run confirmed retention, idle readiness and one reply after resume. Source-only mechanisms and the internal race/failure matrix were kept separate from that live verdict.
- The added real-SQLite claim-race regression passes on the candidate and fails on original production at the intended suspended-delivery assertion. It then verifies one delivery after reopening, with no timer advance or manual drain request.
- Focused ingress-monitor, drain, cancellation and admission checks passed on the Testbox. The full changed regression file and normal commit hooks passed there before the locally signed test commit. The added commit changes only tests, so the product bytes exercised by the live proof are unchanged.

The original fixture startup failure is retained: explicit agent ownership conflicted with a legacy default marker. Correcting the fixture resolved it before the successful baseline/candidate comparison. Profiler counts are scoped to their recorded intervals and are not target-message counts unless separately correlated. The reported production cadence and failure count were not reproduced as numeric claims.

[...]

PR #142768, opened 2026-09-09
fix(channels): hold draining ingress claims before release

### LINKED ISSUE #142013: [Bug]: channel ingress drain re-claims the same message every ~40 ms while the gateway is draining (7,098 failures in three minutes)
### Summary
When the gateway is draining (admission closed, "suspend phase" after a native restart request), every inbound Discord message fails with `GatewayDrainingError`. In `src/channels/message/ingress-drain.ts`, `applyFailureDisposition` handles that error by releasing the claim immediately (`releaseClaim(claim, { recordAttempt: false })`), and the pump re-claims and re-dispatches it at once. Observed on 2026.9.2: 7,098 `message run failed: GatewayDrainingError` lines in three minutes for one message, until the process was restarted. #136684 / #136713 fixed the same pattern for the follow-up queue drain, not for the channel ingress drain.

### Environment
openclaw 2026.9.2, @openclaw/discord 2026.9.2. The drain never completed because a session parked by `sessions_yield` counts as active work, so the storm lasted until an external restart.

### Expected behavior
While the gateway drains, a held claim should wait (a bounded backoff, or until admission reopens) instead of retrying immediately. A parked session (no live turn) should not block the drain indefinitely.

### Local candidate
Hold the claim for five seconds before releasing it when the error is `GatewayDrainingError`.

teddytennant:
Closes #142013.

## What Problem This Solves

Fixes an issue where users sending Discord (or any channel) messages while the Gateway is draining would have the same inbound claim released and immediately re-claimed, producing thousands of `GatewayDrainingError` log lines until an external restart. Observed on 2026.9.2: 7,098 failures in three minutes for one message.

The follow-up queue drain already holds on this error (#136684 / #136713). Channel ingress still released with `recordAttempt: false`, so the pump spun at ~40ms.

## Why This Change Was Made

Hold the claimed row for 5 seconds before release, still without spending the failure budget. The next pump cannot re-dispatch the same event while it is held. Stop/abort skips the release so a successor drain can recover the claim.

No change to retry accounting, dead-letter policy, or non-draining failures.

## User Impact

A draining Gateway no longer storms the same inbound message. The row stays pending for the successor process. Operators do not need to restart to stop the loop.

## Evidence

Baseline: `03497256f95`.

```
node scripts/run-vitest.mjs run src/channels/message/ingress-drain.test.ts
```

29 passed, including `holds a GatewayDrainingError claim before release so the next pump cannot re-claim instantly`. Fake timers: still claimed at 4999ms (`drainOnce` starts 0), released at 5000ms with `attempts: 0`.

### DIFF
--- a/src/channels/message/ingress-drain.draining-hold.test.ts
+++ b/src/channels/message/ingress-drain.draining-hold.test.ts
@@ -1,0 +1,61 @@
+// Gateway draining hold: keep the claim until the backoff elapses.
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { GatewayDrainingError } from "../../process/gateway-work-admission.js";
+import { closeOpenClawStateDatabaseForTest } from "../../state/openclaw-state-db.js";
+import { createChannelIngressDrain } from "./ingress-drain.js";
+import {
+  createTestIngressQueue,
+  type IngressDrainTestPayload as Payload,
+  withTempState,
+} from "./ingress-drain.test-helpers.js";
+
+describe("channel ingress drain draining hold", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+    closeOpenClawStateDatabaseForTest();
+  });
+
+  it("holds a GatewayDrainingError claim before release so the next pump cannot re-claim instantly", async () => {
+    await withTempState(async (stateDir) => {
+      const queue = createTestIngressQueue(stateDir);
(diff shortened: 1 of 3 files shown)


steipete:
Thanks for the fix. Closing this PR as superseded by #142064 (`77e931195bd`), shipped in v2026.9.4: ingress now stays queued during suspension and resumes when admission reopens, with Discord before/after proof; the separate parked-session accounting concern in #142013 remains open. Please reopen with the exact version and sanitized ingress/suspension logs if repeated dispatch persists on v2026.9.4 or later.

PR #141268, opened 2026-09-07
fix(gateway): preserve length in chat completion responses

ooiuuii:
Closes #141264.

A successful partial answer can reach its output budget while the public Chat Completions endpoint reports a natural `stop`. Preserve the agent's authoritative `length` reason in JSON and streaming responses. Streaming keeps that reason until the final result and usage settle; completed tool calls retain precedence.

The production change remains five added and three removed lines. It changes no output budget, retry/fallback policy, authentication, transcript ownership, provider parser, custom header, Responses API or published SDK surface.

Fresh native proof uses a real Gateway, normal configured agent, production HTTP transport and a controlled local OpenAI-compatible provider. No command result is injected, no test-trust runtime flag is used, and no vendor-model token exhaustion is claimed. The provider receives the actual cap of 4 and supplies controlled partial text and usage.

| Public request | Pinned main | Candidate | Preserved behavior |
| --- | --- | --- | --- |
| JSON | `stop` | `length` | HTTP 200, partial text and existing truncation notice |
| Stream without usage | `stop` | `length` | Partial text, one terminal choice, one final `[DONE]`, no usage chunk |
| Stream with usage | `stop` | `length` | Same text/termination, one trailing usage result |

Each length request makes one provider call. Eighteen additional paired controls preserve natural stop, completed client-tool calls, incomplete length-ended tool payload rejection, provider failure/refusal, empty output, context overflow, timeout, client disconnect and a successful request after disconnect. Both versions shut down their owned Gateway and provider cleanly. A fresh independent 20-case public batch also passes its behavioral checks, including different Unicode text, both usage modes, tools, errors, timeout and disconnect. Its four source/delivery joins are tracked separately; completed-tool precedence and unchanged policy/API surfaces are checked against their source owners.

- Baseline: `896fa451da77aa1c72fabe67283e99169efe96cc`.
- Candidate: `a00a19d9dc1e7b2f698e79a653325b9163de0e47`.
- Three focused regressions fail on main for `stop` versus `length`.
- All 221 tests in six owner/sibling suites pass, including all 129 HTTP tests.
- Native formatting, lint, size guards and normal runtime build pass. Fresh complete-source review found no actionable issue.

[...]

Issue #141264, opened 2026-09-07
Chat Completions reports stop when a partial answer hits its output budget


ooiuuii:
# Chat Completions reports stop after a partial answer hits its output budget

## Problem

When an agent returns a usable partial answer after reaching its output-token budget, the OpenAI-compatible `POST /v1/chat/completions` endpoint reports `finish_reason: "stop"`. This happens in both ordinary JSON responses and the final SSE choice.

Clients therefore cannot distinguish budget-truncated output from a natural completion using the standard finish-reason field. The partial answer remains available; this report is not about an empty response or a failed agent run.

## Expected and actual behavior

- Expected: preserve `finish_reason: "length"` when the successful agent result has `meta.stopReason: "length"`.
- Actual: ordinary replies are serialized with `"stop"`; the streaming finalizer only selects `"stop"` or `"tool_calls"`.
- Existing tool calls and failed-run error responses should remain unchanged.

## Reproduction and environment

Source baseline: `00fd5d6a404939da829ceae80e01d941a729cf06`; native Windows with Node 22.22.3. A live model/provider run was not used.

The focused reproduction uses the real embedded terminal resolver to generate a partial-length result, then supplies that result at the existing command test boundary to the real Gateway HTTP server. It tests JSON and SSE, including a lifecycle-end event before the final result and both streaming usage settings.

```text
node scripts/run-vitest.mjs src/gateway/openai-http.test.ts --testNamePattern="preserves the output-budget finish reason"
```

Observed on the original owner: all three selected cases failed at the expected finish-reason assertion (`stop` instead of `length`), while the resolver's deliverable partial-result assertions passed. With the narrowly scoped JSON/SSE mapping repair, the same three regressions passed and the runner exited zero. The other126 test cases were not selected in this focused proof. This is a controlled real-HTTP reproduction, not a live provider measurement.

## Scope

The producer already recognizes and preserves length termination. The missing behavior is its projection to the Chat Completions response. No new retry, header, provider policy, or Responses API behavior is proposed.

PR #142778, opened 2026-09-09
fix(gateway): report truncated Responses replies as incomplete

zhangguiping-xydt:
Related: #141264, #141268

## What Problem This Solves

Fixes an issue where clients calling `POST /v1/responses` would treat an output-token-limited partial reply as a completed answer.

## Why This Change Was Made

The endpoint preserves the agent's authoritative length outcome as an incomplete response and final message, with `incomplete_details.reason: "max_output_tokens"`. Streaming sends `response.incomplete` and the existing `[DONE]` marker. Failure and tool-choice checks retain priority over truncation.

This retains the contributor's repair and aligns the terminal event with the public Open Responses protocol. One final status supplies the message, response, and event; output budgets and provider behavior stay unchanged.

## User Impact

JSON clients can detect a partial answer from its response status. Streaming clients receive the matching incomplete event, including SDK event listeners. Normal completions and failures retain their current behavior.

## Evidence

- Real authenticated Gateway requests through a controlled local provider: main reported completed in both JSON and streaming after authoritative length termination; the candidate reports incomplete, `max_output_tokens`, an incomplete final message, and exactly one `response.incomplete` event plus one `[DONE]` marker.
- Matching before/after controls preserve normal completion, provider failures, content filtering, whole-agent timeout, client cancellation and recovery, required-tool rejection, ordered client calls, usage, response identity, and same-runtime session history. Both Chat Completions length modes remain correct.
- Fresh source-blind acceptance exercised 19 additional public HTTP cases, including identical text with different termination reasons, misleading truncation wording, required-tool precedence, and continuation after an incomplete response. No behavioral findings.
- The actual installed OpenAI SDK receives exactly one incomplete callback for the length case and one completed callback for its normal control; both accumulated final responses retain the correct status and text.
- Focused Responses and Chat Completions suites: 257 passed. Updated regressions fail on main for completed status/event, and on the original contributor implementation for the terminal event and failed-message status.
- Targeted formatting, lint, and diff checks passed. Full hosted checks and the exact-head landing review gate are required before merge.

The runtime proof uses a synthetic provider through normal agent transport. Cancellation proves disconnection and upstream teardown; no stored cancellation result is claimed. History proof covers a same-runtime continuation. Raw prompts, credentials, identifiers, and infrastructure metadata remain private.

### DIFF
--- a/docs/gateway/openresponses-http-api.md
+++ b/docs/gateway/openresponses-http-api.md
@@ -243,11 +243,11 @@ Set `stream: true` to receive Server-Sent Events:
 - Each event line is `event: <type>` and `data: <json>`
 - Stream ends with `data: [DONE]`
 
-Event types currently emitted: `response.created`, `response.in_progress`, `response.output_item.added`, `response.content_part.added`, `response.output_text.delta`, `response.output_text.done`, `response.content_part.done`, `response.output_item.done`, `response.completed`, `response.failed` (on error).
+Event types currently emitted: `response.created`, `response.in_progress`, `response.output_item.added`, `response.content_part.added`, `response.output_text.delta`, `response.output_text.done`, `response.content_part.done`, `response.output_item.done`, `response.completed`, `response.incomplete` (on output-budget truncation), `response.failed` (on error).
 
 Failed agent runs, including whole-agent timeouts, return a failed response. Streaming failures emit `response.failed` followed by `[DONE]`; partial content may already have reached the client. Timeout settings follow the [agent loop](/concepts/agent-loop#timeouts).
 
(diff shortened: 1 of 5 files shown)


obviyus:
Thanks @zhangguiping-xydt!

**Proof:** Authenticated Responses JSON and streaming requests through the real Gateway now report output-budget truncation as incomplete. The installed SDK receives the matching incomplete callback.

- **Bug:** A partial reply was labeled completed.
- **Fix:** Preserve the authoritative length outcome in the response, final message, and terminal event while retaining failure and required-tool precedence.
- **Validation:** 257 focused cases, four affected regressions after the assertion typing correction, 19 fresh public-route acceptance cases, and [green exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34308075794). Final independent review found no remaining issues.

| Before | After |
| --- | --- |
| Length-limited reply reported completed | Incomplete response and message with `max_output_tokens` |
| Stream ended with `response.completed` | Exactly one `response.incomplete` and one `[DONE]` |

Merged as `a2bdd85eace67e12e2a57b94464f46c48136b94a`. Both contributor revisions were preserved in the repair history. Runtime proof used a controlled provider; history coverage was a same-runtime continuation.

PR #142780, opened 2026-09-09
fix(ci): add Android emulator diagnostic

vincentkoc:
## What Problem This Solves

Resolves a release-diagnostic gap where the Android beta screenshot emulator can fail during startup without retaining enough runner, SDK, emulator, ADB, and process evidence to distinguish a hosted-capacity problem from an app or workflow defect.

## Why This Change Was Made

Adds a manual, read-only diagnostic workflow that accepts only an exact lowercase commit SHA and reproduces the phone screenshot emulator prerequisites on `macos-26-intel`. A short Ubuntu validation job checks the selected target out under `candidate/`, verifies its exact HEAD, and requires its committed setup action to be a regular `100644` blob byte-identical to trusted `github.workflow_sha` tooling. It executes no local action. The dependent macOS diagnostic job starts fresh, receives no candidate files or paths, and checks out and executes only trusted workflow tooling. The workflow keeps caches disabled and uses the release screenshot Pixel 2/API 36 x86_64 image and boot arguments, two 180-second readiness windows, owned-process cleanup, and an always-uploaded seven-day diagnostic artifact.

The workflow has no environment, secrets, signing, package install, Gradle, Fastlane, or store access. The frozen mobile candidate is unchanged.

## User Impact

Release operators can capture bounded, reproducible evidence for Android emulator startup failures before changing release automation or retrying a store run. This diagnostic cannot publish an app or modify a release.

## Evidence

- Red regression: the original workflow failed the trusted-checkout ordering assertion because it executed the setup action from the selected target. The first same-job repair then received the exact hosted CodeQL cache-poisoning finding because the query conservatively treats a local action following any input-selected checkout as tainted.
- Isolation regression: before the job split, the focused test failed because the required `validate-target` job did not exist.
- Green owner suite: `test/scripts/mobile-release-authority.test.ts` passed 61/61. Its executable parity fixture allows matching committed bytes and rejects both a modified blob and a symlink before the setup sentinel.
- Green workflow guards: `test/scripts/ci-workflow-guards.test.ts` passed 535 with 2 skipped.
- `actionlint`, embedded Bash syntax, targeted oxlint, oxfmt, and `git diff --check` passed.
- P1 autoreview: scoped clean, 0.97 confidence.
- Production/tooling delta: +255 lines for the bounded diagnostic workflow and isolated trusted setup boundary. Test delta: +268 lines for exact input, permission, runner, setup provenance and parity, job isolation, emulator behavior, cleanup, and artifact contracts.
- Hosted runtime evidence is intentionally pending until this manual workflow exists on the default branch; a separate, explicitly authorized secretless dispatch will target the frozen candidate after landing.

### DIFF
--- a/.github/workflows/android-emulator-diagnostic.yml
+++ b/.github/workflows/android-emulator-diagnostic.yml
@@ -1,0 +1,193 @@
+name: Android Emulator Diagnostic
+
+run-name: Android emulator diagnostic (${{ inputs.target_sha }})
+
+on:
+  workflow_dispatch:
+    inputs:
+      target_sha:
+        description: Exact lowercase 40-character commit SHA to diagnose
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  diagnose:
+    runs-on: macos-26-intel
+    timeout-minutes: 25
+    permissions:
+      contents: read
+    env:
+      ANDROID_SCREENSHOT_EMULATOR_TIMEOUT_SECONDS: "180"
+      AVD_NAME: OpenClaw_Screenshots_API36
+      DEVICE_PROFILE: pixel_2
+      SYSTEM_IMAGE: system-images;android-36;google_apis;x86_64
+    steps:
+      - name: Validate target SHA
+        shell: bash
+        env:
+          TARGET_SHA: ${{ inputs.target_sha }}
+        run: |
+          set -euo pipefail
+          if [[ ! "$TARGET_SHA" =~ ^[0-9a-f]{40}$ ]]; then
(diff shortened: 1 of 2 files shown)


vincentkoc:
Merged via squash.

- Prepared head SHA: [1101bac9b8d15c972d7954d96ebd51c524a6a30f](https://github.com/openclaw/openclaw/pull/142780/commits/1101bac9b8d15c972d7954d96ebd51c524a6a30f)
- Landed commit: [f882bfdc8674e929464b9ec918a213e1cf4c11d9](https://github.com/openclaw/openclaw/commit/f882bfdc8674e929464b9ec918a213e1cf4c11d9)

PR #142795, opened 2026-09-09
refactor(tests): assert live meeting capture drafts

steipete:
## What Problem This Solves

A meeting-capture test could pass after the visible title input lost an edit because it checked a cached DOM element and did not wait for a controlled health-refresh transition.

## Why This Change Was Made

Enter the title through an input event in the existing saved-source validation cases, then re-query the live control after the controlled refresh state. Remove the weaker standalone test while retaining every validation row and its save assertions. Removes 8 net test lines.

## User Impact

Stronger regression coverage for unsaved source titles; no production behavior change.

## Evidence

- Full file passes: 37 cases before, 36 retained cases after, with unchanged retained names and statuses.
- A test-only control confirms the old assertion passes with a detached cached input and reset live value. The strengthened assertion detects that fault in all 8 validation rows.
- Diagnostic edits removed; full changed checks and independent P2 review pass.

No runtime improvement claimed.

### DIFF
--- a/ui/src/pages/config/meeting-capture.test.ts
+++ b/ui/src/pages/config/meeting-capture.test.ts
@@ -166,6 +166,7 @@ describe("curated meeting capture", () => {
       await page.updateComplete;
       const control = page.querySelector<HTMLInputElement>(`input[name="${field}"]`)!;
       const fields = [...page.querySelectorAll("input")].map((item) => item.name);
+      input(page, "title", "Not yet submitted");
       const refresh = startedRefresh ?? (await beginRefresh());
       try {
         if (health === "error") {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a871fa2711d4c4e31da0f0206becae11fdcbcb1e](https://github.com/openclaw/openclaw/pull/142795/commits/a871fa2711d4c4e31da0f0206becae11fdcbcb1e)
- Landed commit: [5fae4269e26999cd8ccc0f1654fc11f8adc21f67](https://github.com/openclaw/openclaw/commit/5fae4269e26999cd8ccc0f1654fc11f8adc21f67)

PR #142775, opened 2026-09-09
test(signal): await canonical ingress idle

vincentkoc:
## What Problem This Solves

Resolves an intermittent Signal extension test failure where the harness could observe `0 pending, 1 claimed` while an accepted ingress delivery was still completing. The race blocked unrelated pull requests, including #142628, on the shared Signal shard.

## Why This Change Was Made

The harness now captures the real `SignalIngressMonitor` returned by `startSignalIngressMonitor()` and awaits its canonical `waitForIdle()` lifecycle boundary before checking durable queue state. The existing bounded queue wait remains because debounced claims can settle later at turn adoption.

This changes test support only. It does not serialize production Signal ingress or increase the existing queue timeout.

## User Impact

No runtime behavior changes. Signal monitor coverage stops racing valid claimed deliveries in CI.

## Evidence

- Hosted pre-fix failure: https://github.com/openclaw/openclaw/actions/runs/34298811260/job/102301282817
- Independent failed-job rerun with the same `0 pending, 1 claimed` state: https://github.com/openclaw/openclaw/actions/runs/34298811260/job/102304080968
- Focused Signal monitor tests: 2 files, 51 tests passed
- Full Signal extension suite: 48 files, 863 tests passed
- Final diff: test support `+26/-4`; production `+0/-0`

PR #142796, opened 2026-09-09
improve: speed up Signal delivery tests

steipete:
Related: #142775

## What Problem This Solves

Signal transport fixtures return mocked final replies and expect automatic delivery, but leave visible-reply policy implicit. Their first dispatch can spend several seconds selecting unused harness defaults before reaching the delivery assertions.

## Why This Change Was Made

Declare the existing automatic-delivery policy in the shared fixture and the three complete configuration overrides that dispatch mocked replies. The real ingress, adoption, dispatcher, queue assertions, and timeouts remain in place. Three generic file-header comments are removed, leaving total LOC unchanged. Main's canonical idle synchronization from #142775 is preserved.

## User Impact

Signal delivery tests reach their intended assertions faster and keep their policy expectations explicit. Core harness-default selection retains its separate regression coverage.

## Evidence

On the original CI merge, an isolated Linux diagnostic reproduced the failure and measured 5,112 ms in visible-reply policy selection. The matched fixture diagnostic measured 0.053 ms with all 863 Signal cases passing; the clean Linux run also passed all 863 cases. These measurements are from Node 24.19.0 with two workers and the observed two-CPU CI budget, not a claim about every host.

After carrying the patch onto main `ee2d54f0dc8`, all 48 Signal test files / 863 cases and the complete changed-file gate passed, including extension production/test types and lint. Four existing core implicit-default/explicit-override guards passed separately. No test name or assertion changed. Independent review completed with no actionable findings.

### DIFF
--- a/extensions/signal/src/monitor.tool-result.pairs-uuid-only-senders-uuid-allowlist-entry.test.ts
+++ b/extensions/signal/src/monitor.tool-result.pairs-uuid-only-senders-uuid-allowlist-entry.test.ts
@@ -1,4 +1,3 @@
-// Signal tests cover monitor.tool result.pairs uuid only senders uuid allowlist entry plugin behavior.
 import { Buffer } from "node:buffer";
 import { describe, expect, it, vi } from "vitest";
 import {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [24b9d26003b7ccba88f08c1c43a334729c52c9d7](https://github.com/openclaw/openclaw/pull/142796/commits/24b9d26003b7ccba88f08c1c43a334729c52c9d7)
- Landed commit: [3f0f98ef7efe3c3fceb180b9bef908de5792c7b4](https://github.com/openclaw/openclaw/commit/3f0f98ef7efe3c3fceb180b9bef908de5792c7b4)

PR #142682, opened 2026-09-08
fix(models): disable Fast choices without a request mapping

obviyus:
Related: #136257, #141852, #107378.

## What Problem This Solves

Fixes an issue where users could enable Fast for a request whose model had no Fast mapping or whose explicit service tier already controlled the request.

## Why This Change Was Made

Provider capability metadata now shares alias and service-tier rules with request construction. The selected-model capability carries that result to controls while preserving explicit-tier precedence, unknown native support and saved-preference clearing.

## User Impact

Known ineffective Fast choices are disabled; supported mappings remain available. Typed fast commands can still set or clear preferences. Capability describes request mapping, not speed, price or entitlement.

Consumers: selected-model Fast controls use request-specific support.

## Evidence

A compiled Gateway and authenticated client verified six supported/unsupported metadata cases. Forty-three focused checks passed, including aliases, selected parameters and tier precedence. These checks establish metadata and request mapping; browser clicks and external inference were not tested.

PR #142804, opened 2026-09-09
refactor(ai): keep model policy parsing independent of transport host

steipete:
Related: #142682, #142738

## What Problem This Solves

Resolves a problem where reading OpenAI model policy pulls in the AI transport host solely to parse a positive integer. The unnecessary dependency was found while investigating cold first-reply Signal test failures.

## Why This Change Was Made

Move the exact existing parser into a private leaf module and have Responses payload policy import that leaf. The old transport-utils export remains available. An import-boundary regression prevents model-policy loading from depending on the transport host again.

## User Impact

Model-policy inspection has a narrower dependency graph. Fast choices, model routes, numeric coercion, and existing public exports retain their behavior. No package exports, dependencies, versions, or configuration change.

## Evidence

The import-boundary test failed on the original graph and passes after the change. The canonical five-path gate, ordinary build, and independent Codex review passed. Sixty-six focused tests cover the import boundary, Fast and route policy, Responses handling, and coercion. Rebuilt AI package exports and four native threshold cases also passed.

The original Signal CI shard passed on Linux with fresh caches and three workers: 48 files and 863 tests, without changing its assertions, timeouts, fixtures, or routing. The previously failing prefix case took 4.228 seconds; the cancellation case took 0.447 seconds. All temporary profiling changes were removed.

A causal comparison confirmed that the earlier Fast metadata import increased cold source-loader work. This patch removes a verified transport-host dependency, but residual source-test startup cost remains; these measurements do not establish a production latency improvement. The remote test runtime was Node 24.19, while the original CI failure used Node 24.20; the failure was also reproduced independently on the remote runtime.

### DIFF
--- a/extensions/openai/model-construction.imports.test.ts
+++ b/extensions/openai/model-construction.imports.test.ts
@@ -2,6 +2,14 @@ import { findSourceImportBackedges } from "openclaw/plugin-sdk/test-fixtures";
 import { describe, expect, it } from "vitest";
 
 describe("OpenAI model construction imports", () => {
+  it("loads model route policy without the transport host", async () => {
+    expect(
+      await findSourceImportBackedges("extensions/openai/provider-policy-api.ts", [
+        "packages/ai/src/host.ts",
+      ]),
+    ).toEqual([]);
+  });
+
   it("constructs auth descriptors without loading credential stores", async () => {
     expect(
       await findSourceImportBackedges("extensions/openai/openai-provider.ts", [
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [197cc3bb4136f9296459747b1dd64fa1d03fb2a6](https://github.com/openclaw/openclaw/pull/142804/commits/197cc3bb4136f9296459747b1dd64fa1d03fb2a6)
- Landed commit: [b74002f906435a8674d86880f8a279aca6a29316](https://github.com/openclaw/openclaw/commit/b74002f906435a8674d86880f8a279aca6a29316)

Issue #134515, opened 2026-08-31
[Bug]: update_goal fails with "goal not found", causing a silent infinite retry loop


aminekhettat:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

During a coding/dev task, the agent gets stuck in a silent infinite loop:
it calls the model, gets a normal 200 streaming response, then calls
`update_goal` with `{"status":"blocked","note":"x"}`, which fails with
"goal not found". It then immediately re-queries the model and repeats,
every ~4-9 seconds, indefinitely, with no user-facing error, no fallback,
and no automatic recovery. Confirmed via two journal captures ~5 minutes
apart showing the identical repeating pattern still active.

### Steps to reproduce

Not yet fully isolated. Occurred while running a coding/dev task
against moonshot/kimi-k3 on OpenClaw 2026.7.1-2 in gateway-only mode.

### Expected behavior

A failed update_goal call (especially when it repeats identically)
should surface an error to the user or break the retry loop / trigger
recovery, rather than loop silently and burn API calls indefinitely.

### Actual behavior

System memory was healthy at the time (6.4GB available of 7.9GB), ruling
out OOM. This may be related to the session/goal state-desync issues
mentioned in #66251, though that issue doesn't mention this specific
error message. Suspect the model believes an active goal exists that
needs to be marked "blocked", while the gateway's goal tracker has no
record of it (possibly after a session reset).

### OpenClaw version

2026.7.1-2

### Operating system

Raspi OS headless june 18th 2026  trixie

### Install method

npm

### Model

moonshot / kimi-k3 (openai-completions API shape)

### Provider / routing chain

moonshot

### Additional provider/model setup details

- open AI (codex), GitHub, Anthropic
- Deployment: gateway-only (no local LLM), cloud model providers only
- Coding backend: bundled Codex sub-agent (@openclaw/codex, codex-code-mode-host)

### Logs

```shell
[provider-transport-fetch] [model-fetch] start provider=moonshot api=openai-completions model=kimi-k3 ...
[provider-transport-fetch] [model-fetch] response provider=moonshot ... status=200 ... contentType=text/event-stream
[tools] update_goal failed: goal not found raw_params={"status":"blocked","note":"x"}
(repeats indefinitely)
```

### Impact and severity

Critical as no task is completed correctely. The agent stucks systematically in all the tasks that needs  coding.

jinngimk-lang:
[...]

PR #142808, opened 2026-09-09
fix(agents): return guidance for rejected goal updates

SunnyShu0925:
## What Problem This Solves

When `update_goal` rejects a missing goal or a change from complete to blocked, the model receives a bare error without a next action. Preserve the error and tell the model not to retry, to continue the turn, and to respond to the user.

## Why This Change Was Made

Handle these expected transition errors in the goal tool, where the existing success response already provides final-response guidance. The goal store and transition rules remain unchanged. Missing-session and unrelated errors retain the generic error path.

## User Impact

Rejected goal updates now give the agent an explicit next action. Active → complete and complete → complete still succeed. Rejection neither creates nor reopens a goal.

## Evidence

- Real QA-channel inbound turns through an isolated Gateway and local deterministic provider captured the actual post-call model input and final channel replies on pinned main and the candidate.
- Main returned bare errors for both rejected transitions. The candidate returned the original errors with actionable guidance. Public goal reads confirmed that rejection neither created nor reopened a goal.
- All 26 focused goal-tool and session-goal tests passed. Both guidance regressions failed on the original code for the expected errors. Formatting, focused lint, and native pre-commit checks passed.

Related to #134515. This proves the delivered tool guidance, not natural model compliance or termination of the issue's historical retry loop. The contributor's six live comparisons did not reproduce that loop. The issue remains open.

AI-assisted.


obviyus:
Thanks @SunnyShu0925!

**Proof:** Real QA-channel turns through an isolated Gateway captured the model-facing tool results before and after the fix. Independent behavior acceptance and exact-head review passed. All 26 focused goal-tool and session-goal tests passed; both guidance regressions failed on the original code for the expected errors.

- **Bug:** Rejected goal updates returned a bare error without a next action.
- **Fix:** Preserve the error and tell the model not to retry, to continue the turn, and to respond to the user.

| Before | After |
| --- | --- |
| Missing-goal and completed-goal rejection returned only an error. | Both retain the error and provide next-action guidance. |
| Active → complete and complete → complete succeeded. | Both still succeed; rejection does not create or reopen a goal. |

This verifies delivered guidance with a deterministic local provider. It does not prove natural model compliance or resolve the historical loop in #134515, which remains open.

Hosted CI passed: https://github.com/openclaw/openclaw/actions/runs/34331613197.

SunnyShu0925:
@obviyus Thanks~

PR #142692, opened 2026-09-08
fix(meeting): own a fresh root admission for agent consults

chelsealong:
## What Problem This Solves

Fixes #142610.

Every spoken question in a Google Meet agent-mode session fails with a misleading
`Gateway is draining; new tasks are not accepted` error, even though the gateway is
healthy. `consultMeetingAgent()` in `src/meeting-bot/agent-consult.ts` calls
`consultRealtimeVoiceAgent()` directly, without acquiring its own root-work
admission. Because the consult runs inside the same async chain that handled the
`googlemeet.create` request, it inherits that request's AsyncLocalStorage
root-work context — a context that is already `released` by the time a
participant actually asks a question minutes later. `enqueueCommandInLane()`
then rejects the consult via `isGatewaySubordinateWorkAdmissionClosed()`,
which returns `true` for a released inherited root.

The browser Talk path (`src/gateway/talk-client-agent-consult.ts`) already
handles this correctly by acquiring an independent root admission with
`runOutsideGatewayRootWorkAdmission(tryBeginGatewayRootWorkAdmission)` before
running the consult, and releasing it in a `finally`. The meeting path omitted
this step.

## Fix

`consultMeetingAgent()` now acquires its own root-work admission outside the
inherited (possibly-released) context before calling
`consultRealtimeVoiceAgent()`, mirroring the Talk path exactly, and releases it
once the consult settles. If admission is refused (a genuine drain/suspension),
it now throws `GatewayDrainingError` from an accurate state instead of from a
stale one.

## Evidence

Added a regression test that reproduces the real bug shape: it opens a root
admission (simulating the `googlemeet.create` request), schedules the consult
call as an async descendant of that root's callback (so it inherits the store
via Node's AsyncLocalStorage propagation, exactly like a transcript-triggered
callback registered during meeting creation), releases the root (simulating the
finished creation request), and only then lets the consult proceed. It asserts
that `isGatewaySubordinateWorkAdmissionClosed()` is `false` at the point the
consult actually runs — i.e. a fresh, open admission was acquired instead of the
stale released one.

Confirmed the test fails without the fix, for the intended reason:

```
$ git checkout HEAD~1 -- src/meeting-bot/agent-consult.ts
$ node scripts/run-vitest.mjs src/meeting-bot/agent-consult.test.ts

 × |unit| ... > consults successfully under a fresh admission after the meeting's creation-request root released
   → expected true to be false // Object.is equality
 Tests  1 failed | 5 passed (6)
```

And passes with the fix restored:

[...]

PR #142815, opened 2026-09-09
fix(line): batch proactive reply parts into as few push requests as possible

### LINKED ISSUE #142734: [Bug]: A LINE reply is pushed one request per part, so a card with its caption and image costs three monthly messages
**Summary**

LINE counts one message against the monthly allowance per request per recipient, and documents that the number of message objects inside a request has no effect on that count. The LINE plugin's proactive send path issues one request per part instead, so a reply carrying a card, its caption, and an image spends three of the account's monthly messages rather than one, and three times as many again for every member of a group. The batching that avoids this already exists a few lines away in the same function for one payload shape, and the inbound reply path has always batched.

**Steps to reproduce**

1. Configure a LINE account and send a reply whose payload carries `channelData.line` — a card, a location, or quick replies — together with text, for example through the `message` tool or `openclaw message send`.
2. Read `GET /v2/bot/message/quota/consumption` before and after.

**Expected behavior**

The reply spends one message from the monthly allowance, which is what the platform charges for one request.

**Actual behavior**

It spends one per part. Measured on a real LINE channel against a single recipient with a payload of a card, a caption, and an image:

```
totalUsage 111 -> 114   (3 monthly messages for one reply)
```

**Impact and severity**

The LINE free plan includes 200 messages per month. Every reply built from more than one part multiplies its cost, and an exhausted allowance makes the bot silent for the rest of the billing cycle — the failure #133445 was written to explain.

**Additional information**

Platform rule (LINE, *Sending messages*): "The number of messages counted as sent is the number of people you send the message to. The number of message objects you specify in a single request has no effect on the number of messages sent." <https://developers.line.biz/en/docs/messaging-api/sending-messages/>

Confirmed on the same account: one request carrying five mixed message objects (two text, one Flex, one image, one text with quick replies) returned five sent-message ids and raised `totalUsage` by 1; the same five sent as five requests raised it by 5.

`origin/main` `b077653f6a6`, `extensions/line/src/outbound.ts`: `sendPayload` awaits a separate provider call for the Flex message, the template, the location, each text chunk, and each media URL. The one shape that already does it correctly is the branch guarded by `shouldSendQuickRepliesInline` (`chunks.length === 0 && hasQuickReplies`), which collects every part into one array and sends it through `pushMessagesLine`; the inbound reply path batches five at a time too (`extensions/line/src/auto-reply-delivery.ts`). The cheap path was reachable only for payloads with no text at all, and nothing measured the request count, which is why this survived.

**Distinct related issue**: this overlaps with #133445, which explains what an operator sees once the allowance is gone, but does not own it — that issue is about reporting exhaustion, this one is about reaching it three times faster than the platform charges for.

pengzh1:
## What Problem This Solves

LINE bills one message per request per recipient, regardless of how many message objects a request carries. The LINE plugin's proactive send path (`sendPayload`) issued one request per part, so a reply carrying a card, its caption, and an image spent **three** of the account's monthly messages instead of one (issue #142734: measured `totalUsage 111 -> 114` on a real channel) — multiplied again for every group member. LINE's free plan includes 200 messages/month, and an exhausted allowance leaves the bot silent for the rest of the billing cycle.

## What This Changes

The non-quick-reply path of `sendPayload` now collects the reply's parts — channelData flex card, template, location, markdown-derived flex/text segments, media — into one ordered message array (same part precedence and quick-reply-on-final-message attachment as before) and sends it through the existing `sendMessageBatch` helper (5 objects per request, receipts recorded per chunk) that the quick-reply path already used. Text chunking and the media-before/after-text ordering rule are preserved.

## Evidence

Regression tests in `extensions/line/src/outbound.batch.test.ts` (mocked LINE runtime, no network):

- **One request for three parts**: a payload of card + caption text + image produces exactly **one** `pushMessagesLine` call carrying `[flex, text, image]` — and neither `pushFlexMessage` nor `pushMessageLine` is invoked. Verified **red** on the per-part path (two+ separate sends).
- **Chunking**: a seven-segment reply produces two push requests of 5 + 2 messages.

Full `extensions/line/src/` suite note: this branch's upstream base moved fast today and several untouched suites (`bot-message-context` etc.) fail identically on unmodified upstream HEAD locally (verified) — CI Linux is authoritative.

Fixes #142734

### DIFF
--- a/extensions/line/src/outbound.batch.test.ts
+++ b/extensions/line/src/outbound.batch.test.ts
@@ -1,0 +1,110 @@
+// Line tests cover outbound payload batching (issue #142734).
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const runtimeStub = vi.hoisted(() => {
+  const sendResult = () => ({ messageId: "m1", messageIds: ["m1"] });
+  return {
+    pushMessageLine: vi.fn(sendResult),
+    pushMessagesLine: vi.fn(sendResult),
+    pushFlexMessage: vi.fn(sendResult),
+    pushTemplateMessage: vi.fn(sendResult),
+    pushLocationMessage: vi.fn(sendResult),
+    pushTextMessageWithQuickReplies: vi.fn(sendResult),
+    createQuickReplyItems: vi.fn(() => []),
+    createFlexMessage: vi.fn((altText: string, contents: unknown) => ({
+      type: "flex",
+      altText,
+      contents,
+    })),
+    createLocationMessage: vi.fn(() => ({ type: "location", title: "loc", address: "addr" })),
+    sendMessageLine: vi.fn(sendResult),
+  };
+});
+
+const outboundRuntimeStub = vi.hoisted(() => ({
+  processLineMessage: vi.fn((text: string) => ({ text, flexMessages: [] })),
+  createFlexMessage: vi.fn((altText: string, contents: unknown) => ({
+    type: "flex",
+    altText,
+    contents,
+  })),
+  createLocationMessage: vi.fn(() => ({ type: "location", title: "loc", address: "addr" })),
+  createQuickReplyItems: vi.fn(() => []),
+  pushMessageLine: vi.fn(() => ({ messageId: "om", messageIds: ["om"] })),
+  pushMessagesLine: vi.fn(() => ({ messageId: "om", messageIds: ["om"] })),
+  pushFlexMessage: vi.fn(() => ({ messageId: "of", messageIds: ["of"] })),
+  pushTemplateMessage: vi.fn(() => ({ messageId: "ot", messageIds: ["ot"] })),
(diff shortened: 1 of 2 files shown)


pengzh1:
Closing as superseded by #142692, which opened ~26 hours earlier for the same #142734 root cause and is the better-positioned candidate: its author has live LINE channel access (the quota measurements in the issue), while this PR cannot produce the required after-fix live proof without LINE credentials.

Two findings from this PR's review that apply to any sendPayload refactor and are worth carrying into #142692:

1. `channel.sendPayload.test.ts` asserts the per-part sends (`pushFlexMessage`/`pushMessageLine`/`sendMessageLine` etc.) — landing a batched send requires migrating those assertions to the batch transport (`pushMessagesLine` batches of ≤5).
2. In a batched implementation, batch message objects must still satisfy the LINE message schema (e.g. the `text` field requirements flagged by the type lane).

For reference, this PR's batched non-inline implementation + batching regression tests are preserved in the branch history (`f9b0b0ccd2`).

PR #129535, opened 2026-08-25
fix(talk): return yielded subagent results to voice

vincentkoc:
## What Problem This Solves

Talk could acknowledge delegated subagent work and remain responsive, but the consolidated result returned only to chat after the requester yielded. Two follow-up defects also allowed transcript-only finals to omit their visible text and retained callbacks to outlive a replaced Talk Gateway owner.

Related: #129186, #129144, #121894, #118499, #118505

## Why This Change Was Made

- Keep durable subagent settlement and the consolidated requester wake as the single source of truth.
- Register one bounded, process-local presentation attachment for the exact Talk requester turn.
- Promote it only after durable yielded settlement is frozen.
- Append the visible consolidated answer once to the currently authorized provider delegation.
- Derive transcript-only visible final text from the committed delivery outcome.
- Revalidate the current Gateway owner synchronously before forwarding and revoke retained registration synchronously on replacement.
- Retain retryable work while revoking presentation authority on cancellation, close, detach, replacement, lifecycle rotation, expiry, stale ownership, steering failure, or terminal delivery failure.
- Add no public SDK, config, environment, protocol, SQLite, catalog, manifest, package, or tool-policy surface.

## User Impact

Talk can delegate work without blocking the live conversation. The consolidated parent answer returns to chat and the active voice delegation exactly once. Retired owners cannot speak a late result.

## Exact-Head Evidence

- Head: `fa2bf20e148536c1c33dcf3bf12712206bc4a6b9`
- Pre-fix regressions reproduced: two transcript-only visible-final failures and one immediate replacement-revocation failure.
- Focused tests: 603/603 passed across settlement, registration, promotion, consumption, duplicate rejection, lifecycle revocation, steering, requester-final attachment, Talk relay, Gateway ownership, and provider controller paths.
- Focused repair suites: delivery 296/296; Gateway 36/36.
- Hosted CI exposed a controller max-lines violation and a stale owner request type that omitted `requesterFinal`. The controller-local types now live in a scoped companion module and the owner runner advertises the real request shape.
- Exact-head CI-repair validation: both affected lint stripes passed; focused Gateway/requester-final tests passed 38/38.
- Build passed.
- Native i18n verification passed with 3,784 entries unchanged.
- Diff formatting, conflict-marker, privacy, and secret scans passed.
- Mandatory exact-head autoreview completed with zero P0/P1 findings.
- `check:changed` passed the changed-file gates and stopped only at one unrelated repository-wide compatibility record that became due on September 9, 2026.

[...]

PR #142830, opened 2026-09-09
fix(talk): deliver final answers after repeated delegation

steipete:
Related: #129535

## What Problem This Solves

Fixes an issue where a Talk consult could lose its final answer after yielding through a second delegation wave. The work could finish in the requester transcript while the active voice call never received the final callback.

## Why This Change Was Made

The settlement dispatcher now transfers the current attachment to the continuation's existing run ID. A subsequent yield can bind its next batch, and completion of the previous batch leaves that attachment intact. Shared matching preserves the session, batch, generation, expiry and replacement checks; final consumption still happens once.

## User Impact

Talk can receive the final result when delegated work needs another round of subagents. Cancellation, replacement and terminal cleanup keep their existing behavior.

## Evidence

The existing serial-continuation fixture reproduces the missing append: one new case fails while two controls pass, with both delegation waves, the final response and no-replay assertions already succeeding. The same case passes after the repair. Final validation passes 73 distinct cases covering serial continuation, attachment lifetime, yield, retry, cleanup and Talk delivery, plus all 29 classified checks.

This is isolated synthetic proof through the real attachment, registry and settlement boundaries. It does not claim a live voice or model-provider run. The source fixture uses its documented focused-test path. Independent Codex review found no actionable issues.

### DIFF
--- a/src/agents/subagents/announce/subagent-announce.requester-settle-wake.ts
+++ b/src/agents/subagents/announce/subagent-announce.requester-settle-wake.ts
@@ -34,6 +34,7 @@ import { hasSubagentRunEnded } from "../registry/subagent-run-liveness.js";
 import {
   consumeRequesterFinalAttachment,
   revokeRequesterFinalAttachment,
+  transferRequesterFinalAttachment,
 } from "../requester-final-attachment.js";
 import { getSubagentDepthFromSessionStore } from "../spawn/subagent-depth.js";
 import {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6639910d8e12ed8657599d73610dd0f82277e987](https://github.com/openclaw/openclaw/pull/142830/commits/6639910d8e12ed8657599d73610dd0f82277e987)
- Landed commit: [8d7e1e9069372734b1d1a5af3b5aff2d13793096](https://github.com/openclaw/openclaw/commit/8d7e1e9069372734b1d1a5af3b5aff2d13793096)

PR #142834, opened 2026-09-09
fix(music): close registration resources after generation

steipete:
Related: #140674

## What Problem This Solves

Music generation could leave a cold plugin's database open after the operation, or lose a managed registration while an accepted callback still used it. Provider projection could also throw after starting work, releasing the registration before that work finished.

## Why This Change Was Made

Capability acquisition and the actual music operation now share one resource owner. It retains managed registrations before reading provider properties, joins admitted work on success or error, then releases the registration. Fresh fallback registrations are owned by that finite operation. Providers supplied by the caller and unmanaged host registrations retain their existing ownership.

The shared selection policy and raw resolver signatures remain unchanged. The acquisition helper is internal; this adds no public SDK export, configuration, schema, dependency, or version change.

## User Impact

The public music runtime closes operation-owned plugin resources after completion, including failed operations. Returned audio remains usable after cleanup. Existing provider selection, aliases, fallback, and caller-owned provider behavior remain intact.

## Evidence

Three initial native SQLite ownership regressions failed before the repair while the raw-owner control passed. Independent review found an additional throwing-provider-getter race; its native regression failed before the callback-owned repair. Nine native cases and 149 existing sibling tests now pass, including registration-stable and Node-bound delayed cleanup controls. Fresh independent Codex review is clean. The canonical changed gate passed in 335.1 seconds and the ordinary build in 210.2 seconds.

Actual compiled public music-generation calls ran success → rejection → success with synthetic plugins and native SQLite. Cold calls registered/disposed three times; managed calls shared one registration until host release; raw calls performed no automatic disposal and closed through explicit caller cleanup. Returned buffers were saved and read back, database disposal markers were checked, and read-only reopen verified the final state. Every owned process joined and source/build seals stayed stable.

Observed call timings were 23.03 / 5.29 / 3.18 ms cold, 13.56 / 2.13 / 0.78 ms managed, and 7.05 / 2.05 / 0.61 ms raw. These are single-fixture measurements, not a before/after performance claim. Initial managed/raw proof fixtures lacked the production host registry scope; the corrected probes use that existing scope and preserve the original failures. No external model calls or vendor credentials were used.

### DIFF
--- a/docs/plugins/sdk-overview/host-hooks.md
+++ b/docs/plugins/sdk-overview/host-hooks.md
@@ -45,6 +45,15 @@ still does not invoke a discovery-only engine factory. `dispose()` must not dele
 durable state or disable another registration. Existing raw loader and Gateway
 lifetimes do not gain automatic disposal: keep their `cleanup(ctx)` behavior.
 
+Music generation also owns fresh registrations acquired by
+`api.runtime.musicGeneration.generate(...)`. It waits for the provider's complete
+audio buffers and tracked operation cleanup, then awaits registration disposal
+before resolving or rejecting. Existing managed registrations are retained for
+the operation; raw host registrations and caller-supplied providers retain their
+existing owner. Provider listing still returns caller-owned callbacks and does
+not acquire a generation lifetime. Return asynchronous provider work and have
+`dispose()` stop and join any additional background work it owns.
+
 Executable CLI command registration also uses an owned, uncached registry. Its
 resources remain available through asynchronous registration, command actions,
 and their tracked cleanup, then `dispose()` runs. Closing command preparation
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [26fe3a4c5a0c1baf4e6b988d9c073927be80a0f0](https://github.com/openclaw/openclaw/pull/142834/commits/26fe3a4c5a0c1baf4e6b988d9c073927be80a0f0)
- Landed commit: [6f2bce0aa92be063a6f4356716ae4ac3fe50b878](https://github.com/openclaw/openclaw/commit/6f2bce0aa92be063a6f4356716ae4ac3fe50b878)

PR #142842, opened 2026-09-09
fix(auto-reply): resolve the CLI execution auth identity at dispatch

### LINKED ISSUE #142820: [Bug]: claude-cli — Telegram inbound turns receive an API-key credential that identical RPC turns do not
### Bug type

### Beta release blocker

No

### Summary

A Telegram-originated turn on the `claude-cli` backend spawns a `claude` child carrying `CLAUDE_CODE_API_KEY_FILE_DESCRIPTOR` and bills the Anthropic API key, while an `openclaw agent` turn on the same session key 28 seconds later spawns a child with no credential and correctly uses the CLI's subscription login.

### Steps to reproduce

1. Configure the `claude-cli` runtime for an Anthropic model on a host with a working local `claude` login (Claude Max): `agents.defaults.models["anthropic/claude-sonnet-5"].agentRuntime.id = "claude-cli"`, `auth.order.claude-cli = ["anthropic:claude-cli"]`. The auth store holds exactly one Anthropic profile, `anthropic:default` (api_key); zero oauth/token profiles.
2. Send the agent a Telegram DM. While the child is alive, inspect `/proc/<pid>/environ` of the gateway-parented `claude` process.
3. Run the same agent with the same session key via `openclaw agent --agent main --session-key agent:main:main --message "..."` and inspect its child the same way.

### Expected behavior

With `auth.order.claude-cli` naming the native-login id and no OAuth/token profile registered, the inbound channel path should resolve credentials the same way the RPC path does — falling through to the CLI's own login — rather than materializing the stored API-key profile.

The RPC path demonstrates this expected behavior on the same host, same config, same session key, seconds apart.

### Actual behavior

Two `claude` children of the same gateway process (ppid 2191126), same session key `agent:main:main`, same model `claude-sonnet-5`, 28 seconds apart:

| pid | path | credential in env |
|---|---|---|
| 2191398 | Telegram inbound | `CLAUDE_CODE_API_KEY_FILE_DESCRIPTOR` present, fd 3 an open pipe |
| 2191783 | `openclaw agent` RPC | none |

The Telegram turn completes successfully and bills the API key. No `authProfileOverride` (session pin) was present on the session during either run.

### OpenClaw version

2026.9.3 (1391f7c)

### Operating system

Debian 13 (LXC container on Proxmox)

### Install method

npm global

### Model

anthropic/claude-sonnet-5

### Provider / routing chain

openclaw -> claude-cli backend -> Claude Code CLI 2.1.261 -> Anthropic

- Claude Code CLI 2.1.261, logged in as root, `authMethod: claude.ai`, `subscriptionType: max`.
- Node 26.8.1.
- Two agents configured (`main`, `voice`), which use separate auth stores.
- `agents.defaults.models["anthropic/claude-sonnet-5"].agentRuntime.id = "claude-cli"`; `anthropic/claude-haiku-4-5` is deliberately left on the `openclaw` runtime.
- `auth.order.claude-cli = ["anthropic:claude-cli"]` (native-login id).
- Anthropic auth store: one profile, `anthropic:default` (api_key). Zero oauth/token profiles. A second, invalid Anthropic key profile that had been sitting in an auth cooldown was removed before these runs, so it is not a factor.
- This configuration was produced by `openclaw models auth login --provider anthropic --agent main`, which selected the Claude CLI reuse method and reported "Claude CLI auth detected; kept Anthropic model refs and selected the local Claude CLI runtime." That command created no auth profile.

No API keys, tokens, or passwords are included above.

### Logs

```shell
20:17:22 [telegram] Inbound message ... (direct, 111 chars)
20:17:27 [agent/cli-backend] cli exec: provider=claude-cli model=claude-sonnet-5 promptChars=5916 trigger=user useResume=true session=present reuse=reusable historyPrompt=none
20:17:55 [agent/cli-backend] cli exec: provider=claude-cli model=claude-sonnet-5 promptChars=54 trigger=user useResume=false session=none reuse=invalidated:auth-profile historyPrompt=none
```

`reuse=invalidated:auth-profile` on the RPC run indicates the resolved auth profile differed between the two turns.

### Impact and severity

- **Affected:** subscription users running the `claude-cli` backend who reach the agent through a channel rather than the CLI. Observed on Telegram.
- **Severity:** High — silent pay-as-you-go API billing while the user believes traffic is on their subscription plan.
- **Frequency:** Always, on every channel-originated turn observed. Reproduced twice on 2026.9.3 and five times on 2026.8.2.
- **Consequence:** Unexpected Console charges. One day of traffic cost $27.96 before a spend cap stopped it. Because the turn succeeds and emits no error, there is no signal short of the billing console.

### Additional information

- Also reproduced five times on 2026.8.2, so this is not new in 2026.9.3. No known-good version has been identified, which is why this is filed as a behavior bug rather than a regression.
- This matches the behavior described in #135548, which was closed with "reopen only with a reproduction on a current release showing the first compatible saved profile is bypassed." This report is that reproduction, on 2026.9.3, with a same-minute RPC control on the identical session.
- An earlier internal investigation attributed this to a session pin (`authProfileOverride`). That explanation does not hold: no pin was present during these runs.
- Workaround attempts that do **not** resolve it: setting `auth.order.claude-cli` to the native-login id, resetting the session, re-running `openclaw models auth login`, and removing the invalid second key profile. Routing the model to the `openclaw` runtime instead avoids the child descriptor but bills the same key directly over HTTP.

ruel225:
## What Problem This Solves

Fixes #142820.

A Telegram-inbound turn on the `claude-cli` backend spawned the `claude` child with the stored `anthropic:default` api_key forwarded as `CLAUDE_CODE_API_KEY_FILE_DESCRIPTOR` — silently billing the API-key credential. An `openclaw agent` turn on the same agent and session key 28 seconds later spawned the child with no credential and correctly fell through to the CLI's own subscription login. No `authProfileOverride` session pin existed during either run.

## Root Cause

The two run paths resolve auth differently:

- **Gateway command path** (`src/agents/command/attempt-execution.ts:884-895`): for CLI backends it converts the session selection through `resolveCliExecutionAuthProfileId` (`src/agents/cli-execution-auth.ts`), which resolves the order under the *backend id* (`claude-cli`). With `auth.order.claude-cli = ["anthropic:claude-cli"]`, the explicit order names a native-login id that the store does not hold, so the resolver returns `undefined` and the run falls through to the CLI's native login.
- **Reply dispatch path** (`src/auto-reply/reply/agent-runner-cli-candidate.ts`): the session layer (`resolveSessionAuthSelection` → `resolveSessionAuthProfileOverride` in `src/agents/auth-profiles/session-override.ts`) resolves the order under the *model provider* (`anthropic`), misses the backend-keyed order, auto-pins the only compatible store profile (`anthropic:default`, api_key), and the dispatch boundary forwarded that id verbatim into the run. `prepare.ts`'s explicit-profile branch then materialized the key and the CLI child inherited it.

The reply path skipped the CLI execution auth owner entirely.

## Solution

Convert the session-layer selection through the same CLI execution owner at the reply dispatch boundary:

- New `resolveCliForwardedAuthProfileId` in `src/auto-reply/reply/agent-runner-auth-profile.ts` composes the two steps the boundary needs: it resolves the candidate run's session-layer pin (`resolveRunAuthProfile`) and converts it through `resolveCliExecutionAuthProfileId` with the backend's agent dir.
- `agent-runner-cli-candidate.ts` (the shared dispatch for the primary and every fallback CLI candidate) consumes that single unit for the forwarded `authProfileId`.

Behavior preserved: profiles owned by the CLI backend are still forwarded, native-login ids forward as none, and a user-locked profile owned by another provider still fails closed (`CliExecutionAuthProfileError`) instead of running as another identity.

## Evidence

- Dispatch-level regression `src/auto-reply/reply/agent-runner-cli-candidate.auth-profile.test.ts` drives `executeAgentTurn` through the real CLI fallback candidate dispatch (`runCliFallbackCandidate` → `runCliAgentWithLifecycle`) and asserts the `authProfileId` that actually reaches the CLI runner (`runCliAgent` params): a model-provider auto pin (`anthropic:default`, api_key) forwards as `undefined`, and a backend-owned profile (`claude-cli:work`) still forwards. Pre-fix red: with the dispatch site restored to `main`'s raw `resolveRunAuthProfile(...).authProfileId` forwarding, the dispatch assertion fails with `expected 'anthropic:default' to be undefined` — the reported silent-billing defect observed at the exact forwarded boundary — and passes with the fix. Helper-level cases pin the conversion contract (native-login id forwards as none, cross-provider user lock fails closed).
- Child-level regression `src/auto-reply/reply/agent-runner-cli-candidate.cli-child-auth.test.ts` spawns a real claude stub child (JSONL on stdout, credential env key names only — never values) through `executeAgentTurn` → the real CLI runner, with a real SQLite auth store seeded under an isolated `OPENCLAW_STATE_DIR`. A model-provider auto pin (`anthropic:default`) completes the turn with the child observing **no** `CLAUDE_CODE_API_KEY_FILE_DESCRIPTOR` (its native login stays intact), and the same real-store dispatch forwards a backend-owned profile (`claude-cli:work`) for the backend's own materialization. Pre-fix red: restoring `main`'s dispatch forwarding fails the real-store drop assertion with `expected 'anthropic:default' to be undefined`. (The plugin-owned last mile — credential → private descriptor env — stays covered by the plugin's own `prepareExecution` suites; this suite observes the credential env the runner forwards.)
- Sibling suites green: `agent-runner-utils`, `agent-runner-execution-probes`, `agent-runner-cli-candidate`, `cli-execution-auth` (42 tests total).
- `tsgo` core and the auto-reply test shard pass on fresh `tsbuildinfo` caches; `oxlint --type-aware` and `oxfmt` clean on the touched files.
### Installed-backend real-behavior proof (after-fix, 2026-09-09)

Covers the three review asks through the installed Anthropic CLI backend. Setup: `dist` built from this head (banner `2026.9.3 (84c3b42)`), the bundled Anthropic plugin's `claude-cli` backend with its direct stdio transport to the installed Claude Code 2.1.266, an isolated `OPENCLAW_STATE_DIR` and workspace, real agent turns (`openclaw agent --local --model claude-cli/sonnet`). The claude child process was observed through `/proc/<pid>/environ` and `/proc/<pid>/fd` while the turn was live — credential env **key names only**, never values; endpoint and credential values never printed.

1. **Explicit managed credential materializes through the real plugin seam.** Config `auth.order.claude-cli = ["claude-cli:work"]` with a seeded store profile (`claude-cli:work`, token type). The run selected that profile (the auth-failure ledger keyed the first turn's model-region 403 to `profile=sha256:ac5dc42b7f3c provider=claude-cli`, not to a fallback), and the live claude child's environment had `CLAUDE_CODE_OAUTH_TOKEN_FILE_DESCRIPTOR` **present** with every host `ANTHROPIC_*` var cleared (`ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL` all absent — the backend's clearEnv list). The turn completed (`MANAGED-OK`).
2. **Existing-session transition across the credential boundary is explicit, not silent.** Same session key, then the store profile was removed so the session's selection dangles. The next turn's claude child environment had **no** descriptor key at all (native CLI login), and OpenClaw refused to carry the CLI-side history across the boundary (`cli session history refused across auth boundary: reason=auth-unknown`, `historyPrompt=none`) while continuing with a fresh CLI session id. The turn completed (`NATIVE-STILL-OK`). This is the P1 merge-risk scenario end to end: the credential identity changes and the run neither silently reuses the previous identity's CLI history nor fails.
3. **Native login on a fresh session.** No profiles configured. The claude child environment again had no descriptor key and no inherited `ANTHROPIC_*` values; the CLI authenticated from its own config (`--setting-sources user`). The turn completed (`NATIVE-FRESH-OK`) and the CLI session id stayed stable across a follow-up turn (`useResume=true`, same resume id).

These observations close the gap between the dispatch-level regression (the forwarded `authProfileId` the runner receives) and the plugin's own `prepareExecution` suites (credential → descriptor): on the installed backend, the forwarded identity is what actually materializes in the child — descriptor present only for an explicit managed selection, never for the auto-pin drop path — and the transition from a managed credential to native CLI login is a deliberate, logged boundary event with a fresh CLI session id.
### Installed-backend channel-reply proof (after-fix, 2026-09-10)

Covers the remaining ask directly: an **inbound channel turn** through the installed backend with the report's store state, observed at the live claude child. Setup: `dist` built from this head (banner `2026.9.3 (4af9225)`), a real gateway process with a real channel plugin — `qa-channel`, the repo's synthetic Slack-class transport, which enters through the same `channelRuntime.inbound.dispatch` → reply-dispatch boundary a Telegram adapter uses (no substitute backend; the bundled Anthropic plugin's `claude-cli` backend spawns the installed Claude Code 2.1.266 child). Isolated `OPENCLAW_STATE_DIR`, `agents.defaults.model = "claude-cli/sonnet"`, and the report's exact store state: exactly one profile `anthropic:default` (api_key) with `auth.order.claude-cli = ["anthropic:claude-cli"]`, no session pin. While each turn was live an observer polled `/proc` under the gateway's task tree for the claude child and recorded credential env **key names** and descriptor-fd presence only — never values.

1. **Reported scenario, after fix — the inbound reply run forwards no credential.** An inbound DM injected through the channel's local bus entered the reply dispatch (`trigger=user`). The spawned claude child's environment had **zero** credential keys (`CLAUDE_CODE_API_KEY_FILE_DESCRIPTOR`, `CLAUDE_CODE_OAUTH_TOKEN_FILE_DESCRIPTOR`, `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL` all absent), the turn completed through the CLI's own login, and the channel delivered the reply back to the bus transcript (`REPLY_OK`):

   ```shell
   [agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
   [agent/cli-backend] cli exec: provider=claude-cli model=sonnet promptChars=287 trigger=user useResume=false session=none resumeSession=none reuse=none historyPrompt=none
   [agent/cli-backend] cli turn: provider=claude-cli model=sonnet durationMs=12221 outBytes=8 outHash=7dbe04c44128
   ```

2. **Same reply path forwards a managed credential when the policy selects one.** With `auth.order.claude-cli = ["claude-cli:work"]` and that backend-owned profile in the store (everything else identical, fresh gateway + fresh session), the same inbound-reply dispatch spawned the claude child with `CLAUDE_CODE_API_KEY_FILE_DESCRIPTOR` **present**, and the turn completed with `MANAGED_OK` delivered back through the channel. The zero-credential observation above is therefore the automatic-selection conversion at the changed dispatch boundary — not a pipeline that never forwards.

Together with the command-path observations above, the credential the channel-originated turn finally uses is now established end to end on the installed backend: the automatic model-provider api_key selection reaches the claude child as no credential (native CLI login), an explicit managed selection still materializes its descriptor, and the transition between identities stays a logged boundary event.

## Impact

Subscription users reaching a `claude-cli` agent through any channel (Telegram observed) stop silently switching from their CLI subscription login to the stored API key; the channel path now matches the documented CLI-backend behavior of the command path.

### DIFF
--- a/src/auto-reply/reply/agent-runner-auth-profile.ts
+++ b/src/auto-reply/reply/agent-runner-auth-profile.ts
@@ -1,11 +1,9 @@
 // Resolves auth profile settings that agent runner forwards to providers.
 import { normalizeProviderId } from "@openclaw/model-catalog-core/provider-id";
-import { resolveCliExecutionAuthProfileId } from "../../agents/cli-execution-auth.js";
 import {
   resolveProviderIdForAuth,
   type ProviderAuthAliasLookupParams,
 } from "../../agents/provider-auth-aliases.js";
-import type { OpenClawConfig } from "../../config/types.openclaw.js";
 import type { FollowupRun } from "./queue.js";
 
 /** Keeps an auth profile only when the current provider shares the primary auth scope. */
@@ -44,29 +42,6 @@ export function resolveRunAuthProfile(
     authProfileIdSource: run.authProfileIdSource,
     config: params?.config ?? run.config,
     workspaceDir: run.workspaceDir,
-  });
-}
-
-/**
- * Converts the session-layer profile selection into the CLI execution identity
- * at the dispatch boundary. A session pin resolved for the model provider (for
- * example "anthropic:default") must not reach a claude-cli child as a forwarded
- * API key when auth.order names the backend's native login; the gateway command
- * path applies the same filter before preparing its runs.
- */
-export function resolveCliExecutionForwardedAuthProfileId(params: {
-  cliExecutionProvider: string;
-  authProfileProvider: string;
-  config: OpenClawConfig;
-  agentDir: string;
-  selected: { authProfileId?: string; authProfileIdSource?: "auto" | "user" };
-}): string | undefined {
-  return resolveCliExecutionAuthProfileId({
-    cliExecutionProvider: params.cliExecutionProvider,
-    authProfileProvider: params.authProfileProvider,
-    config: params.config,
-    agentDir: params.agentDir,
-    selected: params.selected,
   });
 }
 
--- a/src/auto-reply/reply/agent-runner-auth-profile.ts
+++ b/src/auto-reply/reply/agent-runner-auth-profile.ts
@@ -1,9 +1,11 @@
 // Resolves auth profile settings that agent runner forwards to providers.
 import { normalizeProviderId } from "@openclaw/model-catalog-core/provider-id";
+import { resolveCliExecutionAuthProfileId } from "../../agents/cli-execution-auth.js";
 import {
   resolveProviderIdForAuth,
   type ProviderAuthAliasLookupParams,
 } from "../../agents/provider-auth-aliases.js";
+import type { OpenClawConfig } from "../../config/types.openclaw.js";
 import type { FollowupRun } from "./queue.js";
 
 /** Keeps an auth profile only when the current provider shares the primary auth scope. */
@@ -42,6 +44,33 @@ export function resolveRunAuthProfile(
     authProfileIdSource: run.authProfileIdSource,
     config: params?.config ?? run.config,
     workspaceDir: run.workspaceDir,
+  });
+}
+
+/**
+ * Converts a candidate run's session-layer auth selection into the CLI execution
+ * identity forwarded to a backend child. The session layer pins by model provider
+ * (for example "anthropic:default"), while auth.order can name the backend's
+ * native login; forwarding the model-provider pin would silently bill the stored
+ * API key instead of the CLI's own login. The gateway command path applies the
+ * same conversion before preparing its runs.
+ */
+export function resolveCliForwardedAuthProfileId(params: {
+  candidateRun: FollowupRun["run"];
+  cliExecutionProvider: string;
+  authProfileProvider: string;
+  config: OpenClawConfig;
+  agentDir: string;
+}): string | undefined {
+  const selected = resolveRunAuthProfile(params.candidateRun, params.cliExecutionProvider, {
+    config: params.config,
+  });
+  return resolveCliExecutionAuthProfileId({
+    cliExecutionProvider: params.cliExecutionProvider,
+    authProfileProvider: params.authProfileProvider,
+    config: params.config,
+    agentDir: params.agentDir,
+    selected,
   });
 }
 
--- a/src/auto-reply/reply/agent-runner-auth-profile.ts
+++ b/src/auto-reply/reply/agent-runner-auth-profile.ts
@@ -1,9 +1,11 @@
 // Resolves auth profile settings that agent runner forwards to providers.
 import { normalizeProviderId } from "@openclaw/model-catalog-core/provider-id";
+import { resolveCliExecutionAuthProfileId } from "../../agents/cli-execution-auth.js";
 import {
   resolveProviderIdForAuth,
   type ProviderAuthAliasLookupParams,
 } from "../../agents/provider-auth-aliases.js";
+import type { OpenClawConfig } from "../../config/types.openclaw.js";
 import type { FollowupRun } from "./queue.js";
 
 /** Keeps an auth profile only when the current provider shares the primary auth scope. */
@@ -42,6 +44,29 @@ export function resolveRunAuthProfile(
     authProfileIdSource: run.authProfileIdSource,
     config: params?.config ?? run.config,
     workspaceDir: run.workspaceDir,
+  });
+}
+
+/**
+ * Converts the session-layer profile selection into the CLI execution identity
+ * at the dispatch boundary. A session pin resolved for the model provider (for
+ * example "anthropic:default") must not reach a claude-cli child as a forwarded
+ * API key when auth.order names the backend's native login; the gateway command
+ * path applies the same filter before preparing its runs.
+ */
+export function resolveCliExecutionForwardedAuthProfileId(params: {
+  cliExecutionProvider: string;
+  authProfileProvider: string;
+  config: OpenClawConfig;
+  agentDir: string;
+  selected: { authProfileId?: string; authProfileIdSource?: "auto" | "user" };
+}): string | undefined {
+  return resolveCliExecutionAuthProfileId({
+    cliExecutionProvider: params.cliExecutionProvider,
+    authProfileProvider: params.authProfileProvider,
+    config: params.config,
+    agentDir: params.agentDir,
+    selected: params.selected,
   });
 }
 
--- a/src/auto-reply/reply/agent-runner-cli-candidate.auth-profile.test.ts
+++ b/src/auto-reply/reply/agent-runner-cli-candidate.auth-profile.test.ts
@@ -4,6 +4,8 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { AuthProfileCredential } from "../../agents/auth-profiles/types.js";
 import { testing as cliBackendsTesting } from "../../agents/cli-backends.test-support.js";
+import { resolveCliExecutionAuthProfileId } from "../../agents/cli-execution-auth.js";
+import { resolveRunAuthProfile } from "./agent-runner-auth-profile.js";
 import type { FollowupRun } from "./queue.js";
 
 const mocks = vi.hoisted(() => ({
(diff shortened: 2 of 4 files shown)


ruel225:
@clawsweeper re-review

The PR body gained an "Installed-backend channel-reply proof (after-fix, 2026-09-10)" section at the same head `4af92255f6e5430a7bb97e23087634053c929845` (no code change): an inbound channel turn through a real channel plugin and the installed `claude-cli` backend, with the report's exact store state — the spawned claude child observes zero credential env keys and the turn completes through the CLI's own login (`REPLY_OK`), while the paired managed-selection run on the same reply path still materializes its descriptor (`MANAGED_OK`). This covers the P1 channel-originated-credential observation and the P2 after-fix inbound reply run.

PR #142845, opened 2026-09-09
refactor(tests): share independent no-queued-reply expectations

steipete:
Related: #139428

## What Problem This Solves

Fifteen dispatch assertions repeat the same no-queued-reply result shape across related delivery, cancellation and terminal-failure suites.

## Why This Change Was Made

Share a pure expected-result constructor that returns fresh literal objects, independently of production and mock inputs. Preserve every existing matcher and companion no-send, terminal-outcome, receipt and operation-ownership assertion. Queue counts alone are not treated as delivery proof.

## User Impact

No user-visible behavior change. The five-file cleanup removes 35 net test/support lines without deleting scenarios or changing runtime behavior.

## Evidence

- All four complete suites passed before and after: 256 cases, with identical per-file ordered case names and statuses.
- Full AST expansion reconstructs the original consumers; all fifteen constructor calls remain expected-only, with exact values/key order and two fresh objects per call.

No runtime improvement is claimed.

### DIFF
--- a/src/auto-reply/dispatch.delivery-order.test.ts
+++ b/src/auto-reply/dispatch.delivery-order.test.ts
@@ -3,6 +3,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { createDeferred } from "../../test/helpers/promise.js";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import { resetGlobalHookRunner } from "../plugins/hook-runner-global.js";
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [300124b4cef8389800757e6a6575cf660c87a983](https://github.com/openclaw/openclaw/pull/142845/commits/300124b4cef8389800757e6a6575cf660c87a983)
- Landed commit: [d9216e894e89cec406d42607c175aeecf4a13ee9](https://github.com/openclaw/openclaw/commit/d9216e894e89cec406d42607c175aeecf4a13ee9)

PR #142852, opened 2026-09-09
fix(announce): deliver completion media through the direct text fallback

### LINKED ISSUE #132780: Subagent completion announce text-fallback silently drops media attachments (MEDIA: lines stripped, structured media ignored)
### Summary
When a subagent/media completion's wake into the requester session fails (or the route requires message-tool delivery and the agent's reply is delivered via the announce direct-delivery fallback), media attachments are SILENTLY DROPPED: the text is sent, the media never is, and nothing in the reply pipeline ever stages or sends it.

### Repro
1. Telegram DM; fire an `image_generate` in a turn while another run holds the session lane (e.g. a subagent completion announce in flight).
2. Completion wake into the requester session fails → "Active requester session could not be woken ... falling back to requester-agent handoff" (journal).
3. The announce flow runs the completion agent, then `deliverTextCompletionDirect` ships the reply via the generic outbound send.
4. The agent's reply contained a `MEDIA:<path>` directive line and the payload carried structured media — but the photo NEVER ARRIVES. Only the caption text lands.

### Root cause (traced in `src/agents/.../subagent-announce` → shipped dist `subagent-announce-origin-*.js`)
- `extractAgentVisibleReplyText()` reads ONLY `payload.text` from the agent result payloads — `payload.mediaUrl` / `payload.mediaUrls` are ignored entirely.
- It regex-STRIPS `MEDIA:` directive lines (`/^[ \t]*MEDIA:[^\n]*\r?\n?/gm`) — destroying the media reference instead of converting it to an attachment.
- `deliverTextCompletionDirect()` then sends text-only via the generic outbound send — bypassing the channel reply pipeline: no media staging, no sendPhoto. (Generic send converts markdown to plain text too, so the message also arrives formatting-stripped.)
- Net: a compliant agent does everything right (caption + MEDIA line + structured attachment) and the fallback path destroys the attachment.

### Evidence (2026-08-29 09:17–09:33 UTC, Telegram DM)
- 09:17:44 journal: wake failure → requester-agent handoff (`queue_message_failed reason=runtime_rejected`).
- 09:17:45 `[turn/kernel] visible channel turn dispatched with no queued reply payloads` — the reply was queued, then flushed via the fallback.
- 09:19:33 outbound send: `sendMessage` TEXT ONLY. No `sendPhoto`, no media staging line, no `deliver kind=` trace. Delivered text was markdown-stripped with the `MEDIA:` line gone. User-visible symptom: text report, no image.
- Image file existed on disk the whole time (404 KB) — generation succeeded; only delivery dropped.

### Proposed fix (deployed locally as a dist patch, working)
In the announce direct-delivery text fallback:
1. `extractAgentVisibleReplyText` collects `payload.mediaUrl`/`payload.mediaUrls` (deduped) AND rescues paths from `MEDIA:` directive lines *before* stripping them; returns `{ text, mediaUrls }`.
2. `deliverTextCompletionDirect` passes `mediaUrls` into the outbound send when non-empty — the send API already supports it (the generated-media fallback path uses the identical signature), so this reuses a proven capability instead of inventing one.
3. No double-delivery risk: the generated-media fallback early-returns before the text path can run for agent-mediated completions; the paths are mutually exclusive.

### Known residual limitation (disclosed honestly)
Caption-LESS media replies still drop: when the reply text is empty after stripping (or a bare structured `mediaUrls` payload with no text), the extractor returns before media collection. A complete fix should collect media regardless of caption presence.

### Related: reply-payload emitters degrade media to text
Several reply-payload modules emit `Attachment: ${url}` text lines rather than structured media when assembling fallback payloads — the media reference survives as TEXT only. Same family; worth auditing while in the area.

---

---

**Community context:** similar media-drop/duplicate-delivery reports exist (#59956, #48979, #46240), but the outbound completion-delivery lane (wake-failure → announce direct fallback → silent media strip) appears undocumented. Local dist patch deployed and verified; happy to turn the fix into a PR if the direction is acceptable to maintainers.

Finn763:
## What Problem This Solves

Fixes #132780.

When a subagent completion falls back to the announce direct-delivery text path (`deliverCompletionDirect`), the completion's structured media was silently dropped: the fallback read only the child's sanitized result text from the internal events, so `event.mediaUrls` / `attachments` never reached the outbound send. Caption-less completions (empty result, media only) were worse — the fallback returned `undefined` before sending anything, while the completion still settled as delivered.

What changed in `src/agents/subagents/announce/subagent-announce-completion-delivery.ts`:

- Collect the completion's structured media with the existing `collectAgentInternalEventMedia` helper and pass it to the same outbound send the text fallback already uses. The shared outbound payload plan stages and sends it, so no second `MEDIA:` parser is introduced and caption directives keep working.

- A media-only completion now sends `content: ""` plus its media instead of early-returning; a `failed_notice` never carries the failed child's partial media.

- Delivery bookkeeping: with media present, the commit/`onDeliveryResult` callback fires only after the whole send settles, and a post-send `OutboundDeliveryError` (an identified platform send already began) is reported as a terminal `ambiguous` failure with `missingMediaUrls`, so a retry cannot duplicate the visible completion. Text-only behavior is unchanged.

## Evidence

Red → green on `src/agents/subagents/announce/subagent-announce-direct-completion-media.test.ts` — the PR's own regression file on this head (12 cases), run through `scripts/run-vitest.mjs` with `--config test/vitest/vitest.agents-support.config.ts` and `OPENCLAW_VITEST_MAX_WORKERS=1`:

- **RED** (the two PR source files taken from the merge base `b97803a2`): `7 failed | 5 passed (12)`.
- **GREEN** (head `baaf8946`): `12 passed (12)`.

The full suite was not run; verification is scoped to this file plus the transport harness below.

Closes #132780

# Evidence — real attachment transport on the current head (`baaf8946`)

This replaces the earlier evidence blocks, which referenced revisions (`3c43ebb`, `cfe36a7b50b`) that the Sep 17 rebase no longer contains and that the author withdrew for that reason. Everything below was produced on this branch's **current** head.

* **after** = head `baaf8946c16f10562de57016ff4c1c20b76bd6c0` (branch sources as pushed).
* **before** = the same head with only the two PR source files taken from the merge base `b97803a2f3b207b56605b33c7a236d67b9d5c65b` (`git checkout b97803a2 -- src/agents/subagents/announce/subagent-announce-completion-delivery.ts src/agents/subagents/announce/subagent-announce-direct-delivery.ts`), i.e. this branch's behaviour without its fix.
* Harness `ocproof-142852.transport.test.ts` drives the PR entry point `deliverCompletionDirect(...)` with a **real** `sendMessage`, the **real** Telegram channel adapter and a **real** HTTP receiver that records every request and sha256-hashes the uploaded file bytes. It is copied into `src/agents/subagents/announce/` for a run and deleted afterwards; nothing is pushed by it.
* One fresh vitest process per scenario (isolated test HOME/state; never two vitest processes at once on this host):

```bash
OPENCLAW_VITEST_MAX_WORKERS=1 OCPROOF_LABEL=<after|before> OCPROOF_ONLY=<scenario> \
  node scripts/run-vitest.mjs run --config test/vitest/vitest.agents-support.config.ts \
    src/agents/subagents/announce/ocproof-142852.transport.test.ts
```

Node `v26.4.0`, win32. `git status --porcelain` is empty after every run.

## Observed wire traffic — head `baaf8946` vs the same head without the two fixes

| scenario | after (head `baaf8946`) | before (PR sources from `b97803a2`) |
|---|---|---|
| `media-unblocked` | `sendPhoto` multipart @827 ms — `generated-media-unblocked.png` 4194 B sha256 `216b9b893371998d…`; delivered=True, settled=1; harness exit=0 | text-only `sendMessage` @657 ms; delivered=True, settled=1; harness exit=1 |
| `media-blocked-mirror` | `sendPhoto` multipart @828 ms — `generated-media-blocked-mirror.png` 4199 B sha256 `d7f1c9d4a0a15a32…`; transportBeforeMirror=True (media @828 ms, mirror @875 ms); settledWhileBlocked=True; delivered=True, settled=1; harness exit=0 | text-only `sendMessage` @670 ms; transportBeforeMirror=False (media @670 ms, mirror @700 ms); settledWhileBlocked=True; delivered=True, settled=1; harness exit=1 |
| `text-blocked-mirror` | text-only `sendMessage` @667 ms; settledWhileBlocked=True; delivered=True, settled=1; harness exit=0 | text-only `sendMessage` @713 ms; settledWhileBlocked=True; delivered=True, settled=1; harness exit=0 |
| `media-refused-mirror` | `sendPhoto` multipart @945 ms — `generated-media-refused-mirror.png` 4199 B sha256 `124e9e5152db50bf…`; delivered=True, settled=1; harness exit=0 | text-only `sendMessage` @745 ms; delivered=True, settled=1; harness exit=1 |
| `hidden-context-media` | text-only `sendMessage` @726 ms; mediaRequests=[] (hidden file sha256 `b220d5fea89331b8…` never uploaded); delivered=True, settled=1; harness exit=0 | text-only `sendMessage` @752 ms; mediaRequests=[] (hidden file sha256 `b220d5fea89331b8…` never uploaded); delivered=True, settled=1; harness exit=0 |
| `media-partial-failure` | `sendPhoto` multipart @741 ms — `generated-media-partial-failure.png` 4200 B sha256 `4f90bd9ca346114f…`; settledCount=0; delivered=False, settled=0; harness exit=0 | text-only `sendMessage` @764 ms; settledCount=1; delivered=True, settled=1; harness exit=1 |

`harness exit` is the test file's own exit code: the cases assert the after-fix contract, so `exit=1` on the `before` side is the expected failure, not a harness error.

## What this shows

1. **Media reaches the wire.** On the after side every media scenario produces a real `sendPhoto` multipart whose uploaded file bytes sha256-match the file on disk (`generated-media-unblocked.png` 4194 B, `generated-media-blocked-mirror.png` 4199 B, `generated-media-refused-mirror.png` 4199 B, `generated-media-partial-failure.png` 4200 B). On the before side **no attachment is ever uploaded** — the fallback sends `sendMessage` text only, exactly the reported bug.
2. **Settlement follows the media batch, not the mirror.** With the transcript mirror held open, the attachment arrives at 828 ms and the reply settles after that while the mirror is still blocked (`transportBeforeMirror=true`, `settledWhileBlocked=true`); the mirror is only entered at 875 ms. Before, the text-only send settles with no attachment on the wire at all.
3. **The hidden-context guard still holds.** A file referenced only from hidden runtime context (sha256 `b220d5fea89331b8…`) is never uploaded on either side — `mediaRequests=[]`, text-only delivery.
4. **A partially failed media batch stays terminal.** First attachment on the wire, second file missing → `delivered=false`, `settledCount=0`, no settlement. Before, the completion settles regardless (`settledCount=1`) because it never had media to fail.
5. **The text-only path is unchanged.** `text-blocked-mirror` behaves identically on both sides (settle before the mirror is entered).

## Caveats

* **Telegram Test Server** (`proof: telegram-e2e`): not run here. The repository `telegram-e2e-userbot` skill leases SUT/TDLib credentials through the Convex QA broker (`qa/convex-credential-broker`), and this host has neither an authenticated `convex` CLI nor `OPENCLAW_QA_CONVEX_SITE_URL` / `OPENCLAW_QA_CONVEX_SECRET_CI`, so no lease can be taken. The harness above is therefore the strongest evidence available locally: it drives the real channel adapter and a real HTTP endpoint rather than a mocked sender, but it is not the Test Server demonstration the label asks for.
* The `before` label is the current head minus the two source fixes, not the pre-PR baseline.
* The full `subagent-announce-delivery.test.ts` file cannot complete on this Windows host (EPERM cascades on `%TEMP%` transcript dirs; identical signature at both revisions), so unit verification is scoped to the PR's own regression file.
* Raw artifacts (per-scenario JSON reports, summary, vitest logs): `%LOCALAPPDATA%\Temp\ocproof-142852\out-r3\`.


Finn763:
@clawsweeper re-review — real attachment transport evidence is in the PR description (owner re-ran the harness: after side shows a `sendPhoto` multipart whose file bytes sha256-match the disk file, arriving before the blocked mirror; before side is text-only with no attachment on the wire).

Finn763:
@clawsweeper re-review - revision cfe36a7b50b fixes both retained findings: (1) internal runtime context is stripped before MEDIA directive extraction (hidden references never reach transport; the transport harness shows BEFORE uploading the hidden-context file vs AFTER text-only), and (2) media completion now settles at the complete-payload boundary before await mirrorDeliveredPayloads (settledBeforeMirror false -> true, with the first-attachment guard intact). Owner re-ran the six-scenario harness on this revision; full red->green table is in the PR description.

Finn763:
Rebased onto current `main` (`8d6decdeca`) — the branch was 652 commits behind and no longer mergeable. Conflicts came from the subagent delivery refactor (`#148828`); resolution follows main's new structure and inserts our semantics at the equivalent point, so **the behavioural change is unchanged**: completion media still goes through the direct text fallback.

- Scope unchanged: `subagent-announce-completion-delivery.ts`, `subagent-announce-direct-delivery.ts`, `subagent-announce-delivery.test.ts` (**+460/−28**).
- New head: `60af50b4cf`. Both commits keep original authorship.
- **Previous head's evidence is void** (the rebase replaced it) — re-review of the new head is needed.

**Evidence on the new head** (`OPENCLAW_VITEST_MAX_WORKERS=1 node scripts/run-vitest.mjs run src/agents/subagents/announce/subagent-announce-delivery.test.ts --maxWorkers=1 --no-file-parallelism`):

```
Tests  1 failed | 322 passed (323)
```

The one failure is **pre-existing on main, not from this change**. Same file, same case, clean `origin/main` checkout at `8d6decdeca`:

```
FAIL deliverSubagentAnnouncement active requester steering > loads a custom...
AssertionError: expected "vi.fn()" to be called with arguments: [ { agentId: 'research', ...(3) } ]
Tests  1 failed | 312 passed (313)      # origin/main, without this PR
```

Identical assertion text on both sides, and every test this PR adds passes.

Finn763:
`checks-fast-baseline-ratchets` fixed in the new head. The ratchet was rejecting **growth in two already-grandfathered files**, which is exactly what a rebase produces when the new code lands back where it was before the base moved:

```
BEFORE (60af50b4cf)
Line-cap ratchet rejects new violations or growth:
  src/agents/subagents/announce/subagent-announce-delivery.test.ts: 5594 -> 5864 counted lines (cap 1000)
  src/agents/subagents/announce/subagent-announce-direct-delivery.ts: 700 -> 709 counted lines (cap 700)
EXIT=1

AFTER (5c29f6c45376)
Line-cap ratchet OK: 4 changed source files; no new violations or over-cap growth.
EXIT=0
```

Fix follows the ratchet's own instruction (*"Extract a coherent sibling module; never trim coverage or disable max-lines"*) — **no test was deleted and no rule was disabled**:

* the 10 added announce tests moved into a new `subagent-announce-direct-completion-media.test.ts` (428 counted lines, cap 1000); `subagent-announce-delivery.test.ts` is now **byte-identical to the merge base** and absent from the diff (313 tests at base + the 10 moved = the 323 the head had);
* the 9-line production growth became a sibling module `subagent-announce-direct-text-fallback.ts` (19 counted lines), with `subagent-announce-direct-delivery.ts` back to **699 ≤ 700**; the change is purely structural — the in-file closure is now a factory call, same default kind, same `agentResult` plumbing, same three call sites.

Diff is now 4 files, +664/−42. Also ran the sibling ratchets in the same job: `pnpm check:max-lines-ratchet` → *822 grandfathered suppressions, OK*; the assertion-safety ratchet is untouched (the new module contains no type assertions); `oxfmt --check` clean on the directory.

Re-verified locally, one file per invocation (`OPENCLAW_VITEST_MAX_WORKERS=1 node scripts/run-vitest.mjs run <file>`): the new test file **10/10 passed**, `subagent-announce-completion-delivery.test.ts` passed, and the announce requester-settle suite 16/16 passed. `subagent-announce-delivery.test.ts` still has one **pre-existing** Windows-only failure (`deliverSubagentAnnouncement … loads a custom main alias through its canonical requester key`: `D:\stores\shared.sqlite` vs `/stores/shared.sqlite`) — that file is byte-identical to the merge base now, so it cannot come from this change; it passes on Linux, which is where CI runs.

Finn763:
@clawsweeper re-review — the transport evidence block has been refreshed for the current head `baaf8946` (all six scenarios re-run on this branch: real `sendPhoto` multipart bytes sha256-matched to disk, media settling after the batch and before the blocked mirror, hidden-context media still excluded, partial batch terminal). The stale blocks referencing `3c43ebb`/`cfe36a7b50b` are gone; the Telegram Test Server lease is not reachable from this host (no Convex broker credential), which is stated in the block.

PR #138517, opened 2026-09-04
fix: reuse existing bootstrap files in read-only workspaces

steipete:
Closes #138514

<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Fixes an issue where users reusing an established read-only workspace would fail during workspace initialization, even when all bootstrap files already exist. Ordinary reply preparation reaches this initializer with bootstrap enabled by default.

## Why This Change Was Made

The atomic publisher introduced in #131108 stages and flushes content before discovering that the destination already exists. The publisher now checks the existing directory entry first, preserving the earlier exclusive-create no-op for regular files and dangling symlinks. Missing files still use complete atomic publication, including the final `EEXIST` race handling, delegated-authority checks, and owned staging cleanup.

This belongs in the canonical publisher shared by normal, sandbox, and dev workspace initialization. Caller bypasses or requiring a new `skipBootstrap` setting would leave the create-if-missing contract broken. Production is **+11/-1, net +10 lines**; that growth implements the existing-entry no-op before any workspace mutation while propagating errors other than `ENOENT`. Tests are **+46 lines**. No schema, persisted representation, or configuration contract changes are needed.

## User Impact

A fully initialized workspace can be reused when its directory permits reading and traversal but denies file creation. Existing authored files and symlinks retain their contents and identities. This concerns host filesystem permissions; the sandbox's `workspaceAccess: "ro"` mount setting alone is not the reported trigger.

## Evidence

- **Before editing:** both new POSIX regression tests failed on the affected owner with `EACCES` at `mkdtemp`; all eight previous publication tests passed.
- **After the repair:** **97 tests across seven files passed**, including partial-write/retry, existing winners, concurrent creators, safe single-link reads, cleanup errors, workspace provisioning, delegated agent-creation cancellation, sandbox seeding, and real dev configuration initialization/retry.
- **Real filesystem probe:** macOS 27.0 arm64, Node v26.8.1, non-root user, synthetic authored files, mode `0555`, and a real separate temporary SQLite state store. The complete pre-change workspace owner succeeds; the affected owner fails staging creation. The fixed owner succeeds, the actual bootstrap prompt reader returns the authored content, new-file creation is still denied, and both authored bytes and dangling symlinks remain intact. The fixed-owner probe was rerun after the final type-only lint annotation.
[...]

PR #142857, opened 2026-09-09
fix(workspace): treat Windows EPERM as an existing bootstrap target (#135230)

### LINKED ISSUE #135230: [Bug]: writeFileIfMissing throws EPERM (not EEXIST) on Windows NTFS
### Bug type

### Beta release blocker

No

### Summary

On Windows, sending a message fails with EPERM: operation not permitted, open AGENTS.md. Root cause: writeFileIfMissing() in dist/workspace-DkQ7irPD.js uses flag wx and only catches EEXIST, but Windows NTFS reports EPERM instead of EEXIST when the bootstrap file already exists or is momentarily held.

### Steps to reproduce

1. On Windows, run OpenClaw 2026.7.1-2 with a workspace on NTFS (optionally behind a directory junction, e.g. D:////OpenClaw -> E:////OpenClaw).//n2. Start the gateway and open the UI.
3. Send any message to an agent.
4. Observe the gateway throw EPERM during workspace bootstrap, and the UI show a send failure.

### Expected behavior

When the bootstrap file (AGENTS.md / SOUL.md / IDENTITY.md / etc.) already exists or is momentarily held, writeFileIfMissing should treat it as already exists and skip, never crashing the request.

### Actual behavior

The gateway throws EPERM: operation not permitted, open E:////OpenClaw////workspace////AGENTS.md, which bubbles up through ensureAgentWorkspace and reaches the UI as send failed. The AGENTS.md file itself is perfectly writable (verified by read/write/append).

### OpenClaw version

2026.7.1-2

### Operating system

Windows 11 (NTFS, workspace behind a directory junction)

### Install method

npm global (portable install)

### Model

N/A - failure occurs before any model request (workspace bootstrap)

### Provider / routing chain

N/A - failure occurs before any LLM request (ensureAgentWorkspace)

The gateway throws "EPERM: operation not permitted, open 'E:\OpenClaw\workspace\AGENTS.md'", which bubbles up through ensureAgentWorkspace and reaches the UI as "send failed". The AGENTS.md file itself is perfectly writable (verified by read/write/append).

### Logs

```shell

```

_No response_

### Impact and severity

Affected: Windows users on NTFS (especially with a directory junction or AV briefly holding a handle). Severity: blocks workflow (cannot send any message). Frequency: intermittent. Consequence: agent chat is completely unavailable until the file is manually touched.

### Additional information

OpenClaw 2026.7.1-2, Node 22, Windows NTFS. The same EPERM quirk is already handled in the @openclaw/fs-safe dependency (fsync_best_effort swallows EPERM). Proposed fix: treat EEXIST + EPERM + EACCES uniformly in the catch block.

Finn763:
## What Problem This Solves

On Windows NTFS, an exclusive bootstrap publish can fail with `EPERM`/`EACCES` instead of `EEXIST` when the target name already exists (delete-pending, antivirus hold, workspace behind a directory junction). `publishBootstrapFile` only accepted `EEXIST` before falling through to the hard-link fallback classification — and `EPERM` is in that fallback set — so the failure surfaced as the misleading `Workspace filesystem does not support atomic bootstrap publication...` (or a raw `EACCES`), aborting `ensureAgentWorkspace` and blocking the chat turn (#135230).

The fix accepts a Windows-only permission-shaped publish failure **only after a post-error check establishes that the bootstrap target already exists**, preserving true permission and path failures:

- `isWindowsExistingTargetCollision(error, targetPath)` is gated on `process.platform === "win32"`, accepts only `EPERM`/`EACCES`, and requires `syncFs.lstatSync(targetPath)` to succeed (same existence notion as the pre-write `lstat`).
- It is checked before `isHardlinkFallbackError`, because `EPERM` is in that fallback set and would otherwise keep masking the collision as a hard-link failure.
- Non-Windows behavior is unchanged; a genuine permission or path failure with no existing target still throws.

> The mechanism in the opening paragraph is the reported one. The **real-Windows errno matrix under Evidence** shows which shapes actually reproduce on NTFS today and flags the open question about this branch; read the two together.

## Evidence

Environment: Windows 11 (10.0.26200, NTFS on every volume), **Node v26.4.0 (libuv 1.52.1)** — a supported runtime per `engines` (`>=24.16.0 <25 || >=26.1.0`), head `bfb112d49c01`. All output below is real terminal output on a real NTFS volume; the only redaction is the temp-directory prefix (`<TEMP>`).

Command (region shard config, single worker, no file parallelism):

```
node node_modules/vitest/vitest.mjs run src/agents/workspace-bootstrap-publish.test.ts --config test/vitest/vitest.agents-core.config.ts --maxWorkers=1 --no-file-parallelism
```

### RED — parent commit `7069a4772ff` (`git checkout HEAD~1 -- src/agents/workspace.ts`)

```
 Test Files  2 failed (2)
      Tests  3 failed | 11 passed | 3 skipped (17)
FAIL  ... > treats a Windows EPERM from the exclusive publish as an existing winner
FAIL  ... > treats a Windows EACCES from the exclusive publish as an existing winner
FAIL  ... > permission-shaped link failure while the bootstrap target exists
AssertionError: expected 'threw: Workspace filesystem does not …' to be 'resolved'
Received: "threw: Workspace filesystem does not support atomic bootstrap publication. Use a workspace on a filesystem with hard-link support."
Caused by: Error: EPERM: operation not permitted, link      (serialized code: EPERM)
Caused by: Error: EACCES: operation not permitted, link     (serialized code: EACCES)
```

(`src/agents/workspace-bootstrap-publish.test.ts` alone before the fix: `2 failed | 8 passed | 3 skipped (13)`.)

### GREEN — head `bfb112d49c01`

```
 Test Files  2 passed (2)
      Tests  14 passed | 3 skipped (17)
```

(`src/agents/workspace-bootstrap-publish.test.ts` alone after the fix: `10 passed | 3 skipped (13)`. The 3 skipped cases are the pre-existing read-only/symlink scenarios, unchanged.)

### Real-filesystem behavior proof (no module mocks)

The reviewer's P1 item is that the hard-link mapping was only supported by mocked `link` errors. The harness below therefore runs the **real `publishBootstrapFile` import** against a real NTFS workspace, with a real second writer claiming the name inside the publish window (through the real `beforePersistentApply` hook + real `fs.writeFileSync`). Nothing intercepts `fs`; every errno is produced by the OS.

Post-fix run:

```
[evidence] RACE :: hookCalls=3 returned=false targetContent="CONCURRENT-WINNER\n" nativeLinkErrno=EEXIST errno=-4075 syscall=link stagingLeftovers=[]
[evidence] PERMISSION :: nameCollisionErrno=EEXIST errno=-4075 syscall=link permissionErrno=EPERM errno=-4048 syscall=link outcome=resolved returned=false targetContent="CONCURRENT-WINNER\n" stagingLeftovers=[]
[evidence] NEGATIVE-CONTROL :: permissionErrno=EPERM errno=-4048 syscall=link threw=Workspace filesystem does not support atomic bootstrap publication. Use a workspace on a filesystem with hard-link support. targetExists=false stagingLeftovers=[]
[evidence] HIDDEN-TARGET :: lstat=exists writeOpenErrno=EPERM errno=-4048 syscall=open exclusiveCreateErrno=EEXIST errno=-4075 syscall=open
 ✓ race: real concurrent winner during the real publish window
 ✓ permission-shaped link failure while the bootstrap target exists
 ✓ negative control: same real permission failure with no target still propagates
 ✓ exclusive-create collision on a real hidden existing target (retired path form)
```

What each scenario is, and what it shows:

- **RACE** — the target name really is absent at the initial `lstat` and really exists when the publish links; the OS reports `EEXIST errno=-4075 syscall=link` for that pair. Pre-fix and post-fix both return `false` and keep the winner bytes, i.e. the plain name collision was already handled by the pre-existing `EEXIST` branch (`stagingLeftovers=[]`, no partial file).
- **PERMISSION** — same real race, plus a real OS state change that makes the staging entry un-linkable, so the real link call fails with a **permission-shaped errno while the bootstrap target really exists** (`EPERM errno=-4048 syscall=link`). Pre-fix: rejected with the misleading `Workspace filesystem does not support atomic bootstrap publication…` (EPERM is a member of `HARDLINK_FALLBACK_CODES`). Post-fix: resolves `false`, winner bytes preserved, staging cleaned up.
- **NEGATIVE-CONTROL** — the identical real `EPERM` with **no** existing target still throws. The `lstat` gate does not swallow genuine access failures.
- **HIDDEN-TARGET** — the retired exclusive-create form from #135230, reproduced natively: a real *hidden* existing `AGENTS.md` opened with `'w'` reports `EPERM errno=-4048 syscall=open` (byte-identical to the issue's `EPERM: operation not permitted, open …AGENTS.md`) even though `lstat` resolves it, while the exclusive `'wx'` open reports `EEXIST`.

### Native Windows errno matrix (real syscalls, redacted)

```
# native errno matrix -- Windows 10.0.26200, node 26.4.0, libuv 1.52.1
# both volumes NTFS: C: yes D: yes
EEXIST errno=-4075 | target = existing regular file            [link]
EEXIST errno=-4075 | target = existing directory               [link]
EEXIST errno=-4075 | target = existing non-empty directory     [link]
EEXIST errno=-4075 | target = existing directory junction      [link]
EEXIST errno=-4075 | target = existing file symlink            [link]
EEXIST errno=-4075 | target = existing hidden file (attrib +H) [link]
EEXIST errno=-4075 | target = existing hidden dir  (attrib +H) [link]
EEXIST errno=-4075 | target = existing, held with FileShare.Read (AV-style hold) [link]
EEXIST errno=-4075 | target = case-insensitive variant of existing name [link]
EEXIST errno=-4075 | target name = the source itself           [link]
EEXIST errno=-4075 | target behind junction on another volume  [link]
OK         | target = delete-pending (unlinked while open) [link]
EPERM errno=-4048  | target = existing hidden file, write-open 'w'  [open]
EEXIST errno=-4075 | target = existing hidden file, exclusive 'wx'  [open]
EEXIST errno=-4075 | target = existing plain file,  exclusive 'wx'  [open]
EPERM errno=-4048  | SOURCE is a directory, target existing    [link]
EPERM errno=-4048  | SOURCE is a directory, target absent      [link]
ENOENT errno=-4058 | SOURCE missing, target existing           [link]
EEXIST errno=-4075 | SOURCE hidden, target existing            [link]
# lstat on the hidden existing target still resolves: yes
```

### How this differs from the existing-file failure already avoided by v2026.9.3

- The failure v2026.9.3 already avoids is the **exclusive-create collision**: an existing target opened for creation/writing. On real Windows an existing *hidden* target reports `EPERM … syscall=open` (row `write-open 'w'`), while a plain existing target reports `EEXIST` (rows `exclusive 'wx'`). This is the errno and syscall in #135230 and it is gone on current main.
- The mapping added here covers the **hard-link** call, a different syscall (`link`, not `open`). On this OS `fs.linkSync` is `CreateHardLinkW(target, source)` followed by `GetLastError()`, and every target-name collision I could build on NTFS reports `EEXIST`, never `EPERM`/`EACCES`. The only real native `EPERM` from `fs.linkSync` on this runtime comes from a non-linkable *source* (rows `SOURCE is a directory`), independent of whether the target exists.

### Remaining risk (honest scope)

- The claim "NTFS reports `EPERM`/`EACCES` instead of `EEXIST` when the hard-link target already exists" **did not reproduce on this machine** in any of the 11 collision shapes above, all of which report `EEXIST` — and `EEXIST` was already handled before this patch (see RACE). So the accepted mapping is a defensive superset, not the fix for the plain collision race. Maintainer call: keep it as defense-in-depth, or drop the branch and rely on the existing `EEXIST` path.
- What **is** proven above on real Windows, with no mocks: a real permission-shaped `link` failure (`EPERM errno=-4048`) alongside a real existing target now resolves to "existing winner" and preserves the file, while the same failure with no existing target still throws (PERMISSION vs NEGATIVE-CONTROL).
- Not exercised here: deny-ACL directories (the elevated token bypasses deny ACEs: `writeFileSync` into a `DENY(W,D)` directory still succeeds), and non-NTFS volumes (only NTFS `C:`/`D:` exist on this machine, so a FAT/exFAT workspace could not be tested).

Closes #135230

### DIFF
--- a/src/agents/workspace-bootstrap-publish.test.ts
+++ b/src/agents/workspace-bootstrap-publish.test.ts
@@ -163,6 +163,55 @@ describe("bootstrap publication atomicity", () => {
     expect(await listTempSiblings(tempDir)).toEqual([]);
   });
 
+  it.each(["EPERM", "EACCES"])(
+    "treats a Windows %s from the exclusive publish as an existing winner",
+    async (code) => {
+      // Windows NTFS can report EPERM/EACCES instead of EEXIST when the
+      // exclusive link collides with a target that already exists (pending
+      // delete, AV hold, or a workspace behind a directory junction). The
+      // writer must confirm the target exists and accept the collision
+      // instead of failing the workspace turn.
+      const tempDir = await makeTempWorkspace("openclaw-workspace-");
+      const agentsPath = path.join(tempDir, DEFAULT_AGENTS_FILENAME);
+      const platformSpy = vi.spyOn(process, "platform", "get").mockReturnValue("win32");
+      const linkSpy = vi.spyOn(syncFs, "linkSync").mockImplementation((_source, target) => {
+        // A concurrent writer won the target; Windows reports a
+        // permission-shaped error for the loser instead of EEXIST.
+        syncFs.writeFileSync(target, "CONCURRENT-WINNER\n");
+        throw Object.assign(new Error(`${code}: operation not permitted, link`), { code });
+      });
+
+      try {
+        await expect(workspace.publishBootstrapFile(agentsPath, "OURS\n")).resolves.toBe(false);
+        expect(await fs.readFile(agentsPath, "utf8")).toBe("CONCURRENT-WINNER\n");
+        expect(await listTempSiblings(tempDir)).toEqual([]);
+      } finally {
+        linkSpy.mockRestore();
+        platformSpy.mockRestore();
+      }
+    },
+  );
+
+  it("keeps a Windows EPERM failure when the bootstrap target is still absent", async () => {
+    // The Windows collision acceptance must not swallow a genuine permission
+    // failure: without an existing target the error still propagates.
+    const tempDir = await makeTempWorkspace("openclaw-workspace-");
+    const agentsPath = path.join(tempDir, DEFAULT_AGENTS_FILENAME);
+    const platformSpy = vi.spyOn(process, "platform", "get").mockReturnValue("win32");
+    const linkSpy = vi.spyOn(syncFs, "linkSync").mockImplementation(() => {
+      throw Object.assign(new Error("EPERM: operation not permitted, link"), { code: "EPERM" });
+    });
+
+    try {
+      await expect(workspace.publishBootstrapFile(agentsPath, "OURS\n")).rejects.toThrow();
+      await expectPathMissing(agentsPath);
+      expect(await listTempSiblings(tempDir)).toEqual([]);
+    } finally {
+      linkSpy.mockRestore();
+      platformSpy.mockRestore();
+    }
+  });
+
   it("keeps a safe reader on the complete single-link file", async () => {
     const tempDir = await makeTempWorkspace("openclaw-workspace-");
     const agentsPath = path.join(tempDir, DEFAULT_AGENTS_FILENAME);
--- a/src/agents/workspace.ts
+++ b/src/agents/workspace.ts
@@ -298,6 +298,28 @@ export function isExpectedAbsentBootstrapFile(name: string): boolean {
   return OPTIONAL_BOOTSTRAP_FILENAMES.has(name) || name === DEFAULT_MEMORY_FILENAME;
 }
 
+/**
+ * Windows NTFS can report a permission-shaped error (EPERM/EACCES) instead of
+ * EEXIST when an exclusive create or hard link collides with a target that
+ * already exists (pending delete, antivirus hold, directory junction). Accept
+ * that mapping only after a post-error check confirms the bootstrap target
+ * exists; a genuine access or path failure still propagates.
+ */
+function isWindowsExistingTargetCollision(error: unknown, targetPath: string): boolean {
+  if (process.platform !== "win32") {
+    return false;
+  }
+  if (!hasErrnoCode(error, "EPERM") && !hasErrnoCode(error, "EACCES")) {
+    return false;
+  }
+  try {
+    syncFs.lstatSync(targetPath);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
 export async function publishBootstrapFile(
   filePath: string,
   content: string | Buffer,
@@ -339,7 +361,10 @@ export async function publishBootstrapFile(
       syncFs.unlinkSync(staging.path);
       outcome = { kind: "created" };
     } catch (error) {
-      if (!linked && hasErrnoCode(error, "EEXIST")) {
+      if (
+        !linked &&
+        (hasErrnoCode(error, "EEXIST") || isWindowsExistingTargetCollision(error, targetPath))
+      ) {
         outcome = { kind: "exists" };
       } else if (!linked && isHardlinkFallbackError(error)) {
         outcome = {


Finn763:
Closing: the path this widened is already handled on main. #138517 replaced the exclusive `wx` open with `publishBootstrapFile`, which `lstat`s first and no-ops when the name already exists, so the Windows EPERM/EACCES case this PR treated as success no longer occurs on the reported path. ClawSweeper's review reached the same conclusion — declining a wider permission policy with no recorded production failure. Backup branch: `backup/oc-142857`. If a real remaining failure can be shown, I'm glad to reopen with that evidence.

PR #142871, opened 2026-09-09
fix: retain model budgets for durable context-engine turns

### LINKED ISSUE #142870: Durable context-engine commits lose the resolved model window and budget
### Bug type

### Beta release blocker

No

### Summary

The durable context-engine turn handoff drops the resolved provider, model ID, model context window, and token budget before calling `commitTurn`.

### Steps to reproduce

1. Use a context engine that implements durable `commitTurn` and reads its budget from `runtimeContext`.
2. Complete an admitted native turn with a resolved model window and token budget.
3. Inspect the accepted-turn outbox payload and the eventual `commitTurn` arguments. Neither carries those model facts.

The accompanying PR extends existing outbox recovery and retry tests with a 200,000-token window and a 180,000-token budget. Both assertions fail against the unpatched outbox.

### Expected behavior

The accepted turn's resolved model window and token budget should reach post-turn processing, including after an outbox retry or recovery.

### Actual behavior

Observed post-turn diagnostic:

```text
tokenBudget=128000 thresholdTokens=96000 model=none modelContextWindow=none
```

Inference used a 1,000,000-token window. LCM fell back to its default budget because the host dropped the resolved model facts.

### OpenClaw version

2026.9.3; source defect also present at upstream b04a3f83ada.

### Operating system

macOS 26.6.1

### Install method

Source checkout

### Model

Regression uses synthetic model facts and a mock context engine. The host defect is independent of the provider model.

### Provider / routing chain

OpenClaw native turn -> durable turn outbox -> LCM `commitTurn`. Provider dispatch is not needed for the deterministic regression.

### Impact and severity

Affected: durable context engines that need the resolved budget during post-turn processing. A smaller fallback budget can trigger premature compaction and prompt-cache invalidation. Frequency across all engines is not established.

### Sanitized work log

AI-assisted diagnosis, implementation, and independent review. Traced native, plugin-harness, and CLI candidate producers through acceptance, recovery, and commit. Two existing outbox tests failed new context assertions on stock and passed with the patch. The 8-test outbox, 13-test accepted-finalization, and 12-test native lifecycle suites passed. Independent review found no actionable correctness or security issues. The fix persists only four primitive model facts, not credentials or live runtime capabilities.

Marvinthebored:
Closes #142870

## What Problem This Solves

Fixes an issue where users of durable context engines could get premature compaction because accepted turns lost the model's resolved context window and token budget. For example, a context engine used its 128,000-token fallback during post-turn processing despite a 1,000,000-token inference window.

## Why This Change Was Made

Carry the resolved provider, model ID, context window, and token budget through the existing durable turn payload into `commitTurn.runtimeContext`. Native, plugin-harness, and CLI producers provide these facts. Accepted-turn recovery and retries preserve them.

Only these four primitive model facts are persisted, not live capabilities, credentials, or the full runtime configuration. Existing rows without this optional metadata keep their current behavior. No database schema migration is needed.

## User Impact

Context engines that accept `runtimeContext` can use the turn's actual model budget during post-turn processing, including after recovery. This prevents the host from forcing fallback budgets merely by dropping those facts.

## Evidence

- Before the fix, two existing outbox tests failed new assertions because recovered and retried commits lacked `runtimeContext`. With the fix, all 8 outbox tests passed.
- Extended existing native-producer and accepted-finalization fixtures to assert a 200,000-token model window and a distinct 180,000-token budget. The lifecycle and finalization suites passed all 12 and 13 tests respectively.
- Existing CLI-transcript and harness-selection suites also passed. The five selected suites passed all 190 tests on the upstream-based PR branch.
- Real-engine proof used the OpenClaw durable outbox with the actual local `LcmContextEngine`. After reconstructing the engine before draining, an older row without metadata used the documented `tokenBudget=128000` fallback, while a new row carried `tokenBudget=180000`, `modelContextWindow=200000`, `provider=anthropic`, and `model=anthropic/claude-sonnet-4-6`; both rows drained successfully. The redacted terminal output was:

  ```text
  pendingAfterDrain=false retainedRows=[]
  legacy-session: thresholdTokens=96000 tokenBudget=128000 model=none modelContextWindow=none
  new-session: thresholdTokens=135000 tokenBudget=180000 model=anthropic/claude-sonnet-4-6 modelContextWindow=200000
  ```
- Core source and all core-test typecheck shards passed on the combined development patch, as did targeted lint and formatting. The broader changed-files gate stopped on an unrelated compatibility-expiry record, also reproduced on the unchanged checkout.
- Independent adversarial review found no actionable correctness or security issues.

AI-assisted implementation and independent review. [Sanitized work log](https://github.com/openclaw/openclaw/issues/142870#sanitized-work-log); private session transcripts are not published. The optional persisted payload contract remains explicitly submitted for maintainer acceptance.


Marvinthebored:
Real-engine proof is now included in the PR body. I exercised the OpenClaw durable outbox with the actual local LCM engine after reconstructing it: an older row without runtime metadata drained using the documented 128k fallback, while a new row carried 180k budget, 200k model window, provider, and model ID; both drained with no pending rows.

The remaining decision is intentionally for maintainers: accept the optional restart-visible JSON payload extension, with old rows retaining fallback behavior.

@clawsweeper re-review

Yigtwxx (commented):
I measured this at head `b74687746d1`. The plumbing is clean and the recovery/retry coverage is genuinely good, but I think the value being carried is the wrong one, and the three producers do not agree on what they are carrying.

## 1. The persisted `tokenBudget` is the raw override, not the resolved budget

`attempt-finalize.ts:239` and `selection.ts:599` write `attempt.contextTokenBudget` / `internalParams.contextTokenBudget` straight through. That field is optional (`embedded-agent-runner/run/types.ts:166`, `agent-settings.ts:36`) and is only set when an explicit budget override is configured. The number the turn actually ran under is resolved separately:

```ts
// src/agents/embedded-agent-runner/run/attempt-history-prepare.ts:186-194
const contextTokenBudget = Math.max(1, Math.floor(
  attempt.contextTokenBudget ??
    attempt.model.contextWindow ??
    attempt.model.maxTokens ??
    DEFAULT_CONTEXT_TOKENS,
));
```

So in the default configuration - no explicit override - the runtime resolves a budget from the model window, but the durable row records nothing.

Measured on your own fixture. `attempt-phase-lifecycle.test.ts` passes both `model: { contextWindow: 200_000 }` and `contextTokenBudget: 180_000`; I removed only the second line, leaving everything else identical, and reran the suite:

```
node node_modules/vitest/vitest.mjs run \
  --config test/vitest/vitest.agents-embedded-agent-run.config.ts \
  src/agents/embedded-agent-runner/run/attempt-phase-lifecycle.test.ts
```

| Fixture | recorded `runtimeContext` |
|---|---|
| as in the PR (12/12 pass) | `{ provider: "test", modelId: "model", modelContextWindow: 200000, tokenBudget: 180000 }` |
| explicit budget removed | `{ provider: "test", modelId: "model", modelContextWindow: 200000, tokenBudget: undefined }` |

The window survives, the budget does not. For that same attempt the resolved budget is 200,000 by the expression above, so the row records `undefined` where the runtime resolved a real number, and a consumer with a documented 128,000 fallback still falls back.

That is the case in the problem statement. "A context engine used its 128,000-token fallback during post-turn processing despite a 1,000,000-token inference window" describes a model with a large window; unless that deployment also sets an explicit `contextTokenBudget`, this change records `tokenBudget: undefined` for it and the fallback still wins. Every test and the terminal evidence in the body set `contextTokenBudget` explicitly (180,000), so the default configuration is never exercised.

There is also a contract mismatch on the field itself:

```ts
// src/context-engine/types.ts:317-318
/** Runtime-resolved context window budget for the active model call. */
tokenBudget?: number;
```

The slot is documented as runtime-resolved; the PR writes the unresolved override into it. Passing the same expression `attempt-history-prepare.ts` computes (or the already-resolved value, if it is reachable at finalize time) would make the field mean what its doc comment says, and would make the fix work without an override.

## 2. The three producers are mirror images of each other

| Field | `cli-run-transcript.ts:375-380` | `attempt-finalize.ts:235-240` | `selection.ts:595-600` |
|---|---|---|---|
| `provider` | `runParams.modelProvider ?? runParams.provider` | `attempt.provider` | `internalParams.provider` |
| `modelContextWindow` | `runParams.modelContextWindow` | `attempt.model.contextWindow` | `internalParams.model.contextWindow` |
| `tokenBudget` | `context.contextWindowInfo?.tokens` | `attempt.contextTokenBudget` | `internalParams.contextTokenBudget` |

The CLI row is the inverse of the embedded row, not a variant of it:

- CLI `tokenBudget` is `contextWindowInfo.tokens`, which is exactly what the CLI runner already uses as its own budget (`cli-runner/prepare.ts:1665`, `contextTokenBudget: contextWindowInfo.tokens`) - i.e. the **resolved** value, essentially always defined.
- CLI `modelContextWindow` is `runParams.modelContextWindow`, which is optional (`cli-runner/types.ts:162`) and is a caller-supplied raw value, not `contextWindowInfo.tokens` (`prepare.ts:1601` shows the resolved window flowing into a different field).

So for the same model an engine can see `{ modelContextWindow: undefined, tokenBudget: 200000 }` from the CLI route and `{ modelContextWindow: 200000, tokenBudget: undefined }` from the embedded route. `provider` diverges too: only the CLI producer prefers `modelProvider`. Since the whole point is that engines can trust these four facts, it would help to resolve them once in a shared helper and have all three producers call it, rather than three hand-written literals.

The body says "Native, plugin-harness, and CLI producers provide these facts." Measured, in the default configuration the native and harness producers provide three of the four.

## 3. `runtimeContext` on `commitTurn` is a shared slot, and this makes it partial

`commitTurn`'s parameter is typed `ContextEngineRuntimeContext` (`src/context-engine/types.ts:451`), the same type `assemble`, `afterTurn` and the maintenance hooks take. On the non-durable path that slot is filled by `buildAfterTurnRuntimeContextFromUsage` (`attempt-finalize.ts:244`), which produces the full object from `buildAfterTurnRuntimeContext` (`attempt-prompt-helpers.ts:508`): `cwd`, `sessionTarget`, `promptCache`, `transcriptStorage`, `rewriteTranscriptEntries`, `llm`, and more.

Before this PR, `commitTurn` simply received no `runtimeContext` from the durable path, so an engine could branch on `undefined`. After it, `runtimeContext` is defined but carries four keys, and `runtimeContext !== undefined` no longer implies the fields an engine reads elsewhere are present. An engine sharing one `readRuntimeContext(ctx)` helper across `afterTurn` and `commitTurn` - a natural thing to do given the identical type - now silently gets `undefined` for everything except these four.

Related: `ContextEngineRuntimeContext` is `Record<string, unknown> & { ... }` (`types.ts:309`), and of the four fields only `tokenBudget` is declared. `provider`, `modelId` and `modelContextWindow` ride in on the index signature, so a plugin engine written against the public type reads them as `unknown` and has to cast. If they are meant to be part of the contract, declaring them next to `tokenBudget` would make that explicit and would let the type system catch producer drift like point 2.

## Smaller note

`recoverContextEngineTurnOutbox` reads `payload.runtimeContext` back out of the stored row. The outbox tests cover this, and I did not find a schema that would strip unknown keys, but it is worth confirming that a row written by an older build (no `runtimeContext`) and one written by a newer build round-trip through the same parse without the field being dropped on the way in - the retry assertion currently checks the field only after `Object.assign`ing it onto the payload in the test itself, not after a real write/read cycle.

None of this touches the recovery and retry work, which looks right to me.

jalehman:
@clawsweeper re-review

jalehman:
Acting on Josh Lehman’s explicit instruction to land this PR, I’m recording his maintainer acceptance of the additive optional durable metadata contract:

- older readers can parse these rows while ignoring the four model facts;
- legacy rows retain engine fallback behavior; and
- if older code performs accepted→ready recovery, it may permanently drop those facts, after which engine fallback applies.

That downgrade behavior is accepted. The payload remains bounded to four optional primitive facts and does not require a SQLite schema or protocol version change.

jalehman:
Maintainer review complete for exact head `f8cc0ea9d7c18d9d2b1e6dd63b5182be1b78223d`.

The original contributor head had one actionable P2: embedded producers could persist the already-capped effective window as the native model window. Commit [`f8cc0ea9d7c`](https://github.com/openclaw/openclaw/commit/f8cc0ea9d7c18d9d2b1e6dd63b5182be1b78223d) fixes that by snapshotting and forwarding the native context window independently from the effective token budget. I rechecked the prior review concerns against current source: production setup resolves the effective budget, dispatch forwards it, CLI native-window/effective-budget fields are intentionally distinct, and the outbox coverage performs real SQLite enqueue/recovery/drain and retry cycles.

Evidence:

- focused/expanded relevant tests: **235 passed across 3 shards**;
- changed-file gates for all 13 paths: **passed**;
- fresh exact-head Codex autoreview: **clean**, no accepted/actionable P0–P2 findings, correctness confidence **0.96**;
- isolated install/build/type+lint check: **passed**;
- exact-head required `openclaw/ci-gate`: **passed** ([run](https://github.com/openclaw/openclaw/actions/runs/34554673869/job/103127288948));
- repository-native `OPENCLAW_TESTBOX=1 scripts/pr prepare-run 142871`: **passed**, verified remote and prep trees both at `f8cc0ea9d7c`, with hosted exact-head gates and no ancestry replacement.

ClawSweeper’s latest exact-head review reports sufficient real-engine proof and no correctness or security finding. Its remaining checklist items are the same persisted-contract owner decision, now explicitly resolved in the [maintainer acceptance](https://github.com/openclaw/openclaw/pull/142871#issuecomment-5628658571).

jalehman (approved):
Approved at `f8cc0ea9d7c18d9d2b1e6dd63b5182be1b78223d` after fresh source review, clean P0–P2 autoreview, focused and changed-file gates, exact-head required CI, and repository-native prepare. The persisted optional metadata and documented downgrade behavior have explicit maintainer acceptance.

jalehman:
Merged via squash.

- Prepared head SHA: [f8cc0ea9d7c18d9d2b1e6dd63b5182be1b78223d](https://github.com/openclaw/openclaw/pull/142871/commits/f8cc0ea9d7c18d9d2b1e6dd63b5182be1b78223d)
- Landed commit: [d6997b6bbf2e269887f03bddae6a488d74908af2](https://github.com/openclaw/openclaw/commit/d6997b6bbf2e269887f03bddae6a488d74908af2)

PR #142873, opened 2026-09-09
fix(auth): preserve device-code verification link boundaries

obviyus:
## What Problem This Solves

Device-code sign-in messages could produce a broken clickable link when the channel renderer included the next line’s Code label in the URL.

## Why This Change Was Made

The three device-code prompt producers now emit standard autolinks around the complete verification URL. Browser callbacks still receive the raw URL; authentication behavior is unchanged.

## User Impact

Users can open the sign-in URL independently of the code label. This fixes authored login prompts rather than the channel’s general bare-link parser.

## Evidence

Signed channel-event replay through a real Gateway and live delivery reproduced the malformed stored link before the fix and the exact destination afterward. A streamed link with query parameters also retained its destination. All five regressions failed when the production edits were removed; all 101 affected tests passed with the fix. Sign-in completion was not performed.

### DIFF
--- a/extensions/github-copilot/index.test.ts
+++ b/extensions/github-copilot/index.test.ts
@@ -17,6 +17,7 @@ import type {
 } from "openclaw/plugin-sdk/plugin-entry";
 import { createTestPluginApi } from "openclaw/plugin-sdk/plugin-test-api";
 import type { fetchWithSsrFGuard } from "openclaw/plugin-sdk/ssrf-runtime";
+import { markdownToIR } from "openclaw/plugin-sdk/text-chunking";
 import { afterAll, afterEach, describe, expect, it, vi } from "vitest";
 import { runGitHubCopilotDeviceFlow } from "./login.js";
 import manifest from "./openclaw.plugin.json" with { type: "json" };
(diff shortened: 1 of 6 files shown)


obviyus:
Thanks @obviyus! Landed in `64ef483a10033c2c92520f2bb2fecfe392566996`.

**Proof:** The same isolated Gateway login replay reached the real provider and Slack before and after the fix. Slack's saved link destination is now exact while authentication is still pending.

- **Bug:** Slack included the following code label in a bare device-verification URL.
- **Fix:** OpenAI, xAI, and GitHub Copilot prompts now emit explicit autolinks; raw browser URLs and authentication behavior are unchanged.

| Before | After |
| --- | --- |
| `/codex/device\nCode` was the link destination | `/codex/device` is the exact destination; the code label stays separate |

All 101 focused tests, five regression-red cases, type-aware lint, build, and live split-link/query-parameter checks passed. Exact-head local ClawSweeper: tier A, zero findings, security cleared, no required decisions or rank-up moves. [CI passed](https://github.com/openclaw/openclaw/actions/runs/34317805761).

Reviewed head: `ce63cd048c52aee6a249733853078abd82f6ce87`. Required co-author attribution was verified. No production deployment was performed.

PR #142875, opened 2026-09-09
fix(macos): recognize installed asdf Node and pnpm CLI

j0urneyk:
## What Problem This Solves

Fixes an issue where macOS app users could not start their local Gateway with an asdf-managed Node runtime and a pnpm-installed CLI, even though both worked in Terminal. Supported Node 26.1.0 was reported as `(unreadable)`, followed by `Command "openclaw" not found`.

## Why This Change Was Made

Both the Node version probe and the downstream Gateway CLI version probe now preserve the app's inherited environment and override only PATH with the discovered search paths. Previously, passing a PATH-only environment removed HOME and caused asdf to fail. CLI discovery also searches `~/Library/pnpm/bin`, retaining the existing parent-directory layout and path precedence.

## User Impact

Users can reuse supported asdf Node and pnpm CLI installations without adding manual symlinks. The change stays in macOS app discovery and includes updated installation documentation.

### Actual asdf/pnpm installations

Following ClawSweeper's review, I tested the compiled native app's `GatewayEnvironment.check()` (used by Connection Recheck) and `CLIInstaller.status()` (used by onboarding) with **actual installed asdf 0.18.0, Node 26.1.0, pnpm 12.2.1, and OpenClaw CLI 2026.9.3**.

Each run had a fresh isolated Foundation HOME/defaults store and no remembered CLI path. Symlinks exposed the actual installed asdf shim and pnpm CLI at `~/.asdf/shims/node` and `~/Library/pnpm/bin/openclaw`; `ASDF_DATA_DIR` pointed to the existing Node installation. The initial PATH contained system/Homebrew directories. Assertions checked the selected shim, discovered CLI, exact versions, and final native statuses. Actual installed executables and package contents were readable, while operator configuration, Gateway state, Keychain services, and network access remained blocked by an OS sandbox.

**Before this follow-up**, on the initial PR head `19d53cd0`, the real installation run confirmed ClawSweeper's downstream concern:

```text
Node: ~/.asdf/shims/node -> 26.1.0
CLI: ~/Library/pnpm/bin/openclaw
GatewayEnvironment.check: kind=error, node=26.1.0, gateway=nil
CLIInstaller.status: ready=true
```

**After preserving the downstream probe environment** (follow-up commit `a97eb115`), a separate fresh HOME/defaults run passed:

```text
Foundation HOME matches isolated HOME; no cached CLI preference
Node: ~/.asdf/shims/node -> 26.1.0
CLI: ~/Library/pnpm/bin/openclaw
GatewayEnvironment.check: kind=ok, node=26.1.0, gateway=2026.9.3
CLIInstaller.status: ready=true
CLI validated: 2026.9.3; compatible with app 2026.9.3
Test run with 1 test in 1 suite passed (exit 0)
```

The temporary proof harness called the actual compiled app methods and was removed from the submitted sources. This proves native discovery and downstream validation with actual installations; signed-app GUI interaction and starting an operator Gateway were outside this run.

### Regression and repository checks

- The initial HOME-dependent Node and pnpm-layout regression cases failed before their respective fixes. The new downstream CLI-to-Node-shim regression failed on `19d53cd0` and passed after the follow-up.
- All 24 selected Swift tests passed, including runtime compatibility, both pnpm layouts, downstream CLI validation, path precedence, profile isolation, and inherited SSH environment behavior.
- `swift build --package-path apps/macos --build-system native --build-tests` passed with only the submitted test sources after removing the temporary proof harness.
- `node scripts/check-changed.mjs --base d9216e894e89cec406d42607c175aeecf4a13ee9` passed, including 1,141 macOS-related Vitest tests, SwiftLint, formatting, dependency guards, and the native state schema guard.
- Independent autoreview confirmed that the working-tree follow-up repairs the downstream defect. Its sole INDEX-only finding referred to the previous PR revision; the reviewed repair is included unchanged in this update.
- Native i18n verification, the changed document's MDX check, and `git diff --check` passed.


j0urneyk:
@clawsweeper re-review

Addressed the downstream `GatewayEnvironment.readGatewayVersion` environment loss in `a97eb115`, with a regression that failed before the follow-up. The PR body now includes before/after evidence using actual installed asdf 0.18.0, Node 26.1.0, pnpm 12.2.1, and OpenClaw CLI 2026.9.3 through the compiled native app's `GatewayEnvironment.check()` and `CLIInstaller.status()` in fresh isolated homes/defaults. Recheck now returns `ok` with both real versions, and CLI status is `ready`.

All 24 selected native tests and the full changed-check gate (1,141 macOS-related Vitest tests) passed. CI is attached to the new head; the proof scope and GUI smoke-test limitation are stated in the body.

PR #136322, opened 2026-09-02
fix: openai-completions cumulative text frames double live assistant deltas

tzlwn1:
﻿Closes #136262

<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers
can help update the branch when needed.

</details>

## What Problem This Solves

Fixes live assistant text doubling when OpenAI-compatible streams deliver **message-shaped cumulative snapshots** (`choice.message`) as full-text frames that would otherwise be appended again.

## Why This Change Was Made

ClawSweeper review required restoring the shared assistant-event contract:

1. **Bare `choice.delta` / `text_delta` stays additive** — intentional long repeats must not be dropped by text-equality heuristics.
2. **Cumulative normalization only at the openai-completions producer**, and only when the wire frame is an authoritative **`choice.message` snapshot**.
3. **Agent Core stays transport-neutral** — no OpenAI heuristic in the generic accumulator. It only skips re-append when a polluted `text_start` partial already holds the confirming prefix (HTTP/SSE fanout still observes the deltas).

Out of scope for this PR: bare `choice.delta` frames that look cumulative without a snapshot marker cannot be distinguished from legitimate repeats under the shared contract; those need a provider-specific frame discriminator if they remain in the wild.

## User Impact

Live TUI/dashboard/channel renders no longer double on message-shaped snapshot replays/extensions from openai-completions. Legitimate repeated long bare deltas (including refusal repeats) remain intact.

### After-fix (PR head `6faea7ea4`)

```text
node scripts/run-vitest.mjs packages/ai/src/transports/openai-completions-stream.usage.test.ts
# → 73 passed

node scripts/run-vitest.mjs packages/ai/src/providers/openai-completions-structured-content-parity.test.ts
# → identical refusals intentionally repeated across separate chunks → "Safety refusal.Safety refusal."

node scripts/run-vitest.mjs packages/agent-core/src/agent-loop.test.ts -t "agentLoop streaming updates"
# → bare exact repeats additive; polluted text_start prefix still emits confirming deltas without double-append
```

### Contract matrix (synthetic, redacted)

| Wire shape | Expected | Result |
|---|---|---|
| `choice.message` equal to prior text | drop snapshot replay | covered by `suppresses message-shaped snapshot replays` |
| `choice.message` extends prior text | emit growth only | covered by cumulative message-shaped / longer snapshot tests |
[...]

PR #132607, opened 2026-08-29
fix(agents): bypass trailing runtime-context carrier relocation for local providers

shreyanshjain05:
## Description
Bypasses trailing runtime-context carrier relocation for local and self-hosted model backends while preserving prompt-cache tail relocation for cloud providers, the bundled LiteLLM cloud route, cloud models behind local proxies, built-in `ollama-cloud`, and Ollama cloud models (`*:cloud`).

## What Problem This Solves
`relocateCurrentRuntimeContextCarrierToTail` moves the current-turn runtime-context carrier message to the absolute tail of the LLM request array to optimize prompt-caching on cloud providers like Anthropic and OpenAI.

However, local and self-hosted model engines (Ollama, vLLM, SGLang, llama.cpp, LM Studio) parse turns strictly chronologically and do not use Anthropic-style prompt cache billing. Relocating the metadata carrier (`role: "user"`) to the tail makes the local model interpret internal metadata (`openclaw.inbound_meta.v2`) as the active user request. On inbound messaging channels (e.g. Telegram), local models (such as `gpt-oss:20b` or Llama 3) consequently ignore the user prompt, echo private metadata, or repeatedly refuse with `"I am sorry, but I cannot share that."`.

*Note on #141544:* Merged PR #141544 sanitizes outward reasoning previews at the shared dispatch boundary; it does not touch model input or runtime-context carrier placement. This PR resolves the underlying model-request ordering defect.

### Classification Logic & Design Decisions
1. **Shared Self-Hosted Classification**: Uses canonical `isSelfHostedProviderId` exported from `model-provider-local` (covering `ollama`, `vllm`, `sglang`, `llama-cpp`, `lmstudio`, and `local`).
2. **Transport API Gate**: Checks `model.api === "ollama"` to cover custom providers running over the Ollama transport.
3. **Cloud Proxies & Local Gateways**: Preserves prompt-cache tail placement for known cloud providers (`anthropic`, `openai`, `openrouter`, etc.) and the bundled LiteLLM route (`provider: "litellm"`, `claude-opus-4-6`), even when accessed via local reverse proxies (`localhost`, `127.0.0.1`, LAN).
4. **Open-Weights Local Priority**: Excludes locally served open-weights models (e.g. `gpt-oss-20b`, `openai/gpt-oss-20b`) from cloud-name inference so local endpoints retain before-user placement. LiteLLM routes targeting open-weights models intentionally classify as self-hosted because local execution wins on ambiguity and those backends do not bill prompt cache.
[...]

PR #142888, opened 2026-09-09
fix(openai-completions): live assistant text doubles when a provider resends full content in one delta

### LINKED ISSUE #136262: openai-completions stream: occasional bare text_delta replaying the full accumulated text doubles message content (n → 2n → n oscillation)
### Environment

- openclaw 2026.8.2; model via `api=openai-completions` (volcengine ark coding endpoint, deepseek model)
- Reproducible without any chat channel: `openclaw agent --session-id <tmp> -m "..."` with `OPENCLAW_RAW_STREAM=1`

### Observed

With `OPENCLAW_RAW_STREAM=1`, `raw-stream.jsonl` shows the accumulated assistant text oscillating **n → 2n → n** within a single message, e.g. `68 → 136 → 68` or `45 → 90 → 45` (the doubled value = the text concatenated with itself). On the dashboard this renders as transient duplicated messages — F5 self-heals because re-rendering uses the persisted transcript. Cumulative channel UIs (e.g. the Feishu streaming card, whose merge rule keeps `current` when `current.includes(next)`) **solidify the doubled text permanently**.

In one run with 4 assistant segments, every segment received such a replay delta (lengths 29/50/57/191/126/435 — each exactly equal to the full text accumulated so far at that point).

### Mechanism

Normal increments arrive as stream events carrying a `partial` snapshot (`resolveAssistantMessageUpdate` prefers `event.partial`). Occasionally a **bare `text_delta` (no `partial`) arrives whose `delta` equals the entire accumulated text so far** — a replayed snapshot masquerading as a delta. `appendTextDeltaToAssistantMessage` (`dist/agent-core-*.js`) appends it unconditionally:

```js
content[contentIndex] = { ...currentContent, text: currentContent.text + delta }; // delta == full accumulated text → doubles
```

The next `partial` snapshot then corrects the state back, producing the n→2n→n signature. 2026.7.1-2 already had this normalization (`resolveAssistantTextChunk`), but in 2026.8.2 it only guards the `deltaBuffer` path, not the message accumulator.

Note: committed final messages (from `response.result()`) are unaffected — only the live streaming view is corrupted.

### Suggested fix

Normalize bare deltas at accumulation time with the same semantics as `resolveAssistantTextChunk`, e.g. inside `appendTextDeltaToAssistantMessage` (and, for defense in depth, the gateway-side rebuild that does `content.text += event.delta`):

- `delta === accumulated` → skip
- `delta.startsWith(accumulated)` → apply only the growth (skip entirely if the growth itself repeats the accumulated prefix — the doubling signature)
- `accumulated.startsWith(delta)` for `delta.length >= 8` → skip (length floor avoids false positives on short, legitimate repeats)
- otherwise append as-is

Fixing it at the accumulator fixes all downstream consumers at once (dashboard plus every channel emit site).

### Local workaround (verified)

Applied the same normalization inside `appendTextDeltaToAssistantMessage`: oscillation gone in raw-stream, replay deltas intercepted, legitimately repeated output (a phrase repeated 3×) unaffected, committed transcripts clean.

Richard355168:
Closes #136262

## What Problem This Solves

Fixes an issue where users of OpenAI-compatible providers would see live assistant text double (n → 2n) mid-message when the provider intermittently resends the entire accumulated text as a single bare `text_delta` frame. Because downstream consumers append text deltas additively, the replayed full content is appended onto the text that is already there; the next provider snapshot corrects it back, producing a visible n → 2n → n oscillation in live channels and dashboards until the message completes.

## Why This Change Was Made

The `AssistantMessageEvent` contract defines `text_delta` as additive: consumers that need the current text rebuild it by replaying deltas from the latest start/end partial checkpoint. A provider frame whose payload exactly equals the whole accumulated visible text violates that producer-side invariant — it is a cumulative snapshot wearing a delta's clothes — and once it has entered the event stream no consumer can distinguish it from a legitimate delta.

This change enforces the invariant at the producer boundary. When `models.providers.<id>.models[].compat.dropCumulativeTextDeltaReplays` is enabled for a model, the guard classifies each maximal visible-text run of an incoming frame (delta-, message-, or structured-array-shaped) — the run boundaries fall where structured reasoning content interrupts the visible text — and drops a run whose filtered visible text is exactly equal to all the visible text the producer has accepted within the current text block (8+ chars, with a length pre-check so the hot path is untouched) before it reaches the additive accumulation path. The run streams through the reasoning-tag partitioner and the DSML recovery/filter stages into an ordered emit plan, and the guard classifies the plan's total only after every stage has released its buffers, so a run that continues (an incomplete trailing tag, a partial DSML token) never matches early. A run that restates the accepted text exactly carries no characters the consumer has not already accumulated, so suppressing it removes no information while keeping the live stream clean; every run that adds anything flows untouched. The ledger itself advances as filtered visible text is released, before the post-tool-call queue can hold it back. Both visible-text feeders — content deltas and visible reasoning details — route through the same admission seam. Everything else keeps flowing: short exact repeats, long repeats that do not restate the whole block, checkpoint-style replay sequences, and retry streams (which start a fresh stream with a new `text_start` checkpoint) all behave exactly as before.

The flag is wired end to end: it is part of the OpenAI completions compat config whitelist and zod schema, and a schema regression test exercises the config-loaded path. Agent-side accumulation stays a single additive path; the sibling accumulation sites are intentionally unchanged because they are contract-conforming consumers — the violation originates at the producer, which is also where the accumulated text is available to detect it.

On the open question whether OpenClaw should expose this per-model equality heuristic at all: the alternatives considered were a default-on guard and producer snapshot normalization. Default-on changes behavior for every provider, and snapshot normalization is the larger design discussed in #136322 with its own compatibility traps; this PR intentionally takes the narrowest slice that fixes the field-reported defect while leaving default behavior untouched. The discriminator this guard relies on is narrow and observable: the reported replay is a single frame, mid-stream, with no new `text_start`, whose payload exactly equals the accumulated text — unlike checkpoint replays (multi-frame sequences restarting from a boundary) and retry streams (new stream, new checkpoint). If maintainers prefer a different shape (for example a different flag placement or requiring the flag to be set by tooling), happy to adjust.

## User Impact

- Default: no behavior change. The guard is off unless a model's `compat.dropCumulativeTextDeltaReplays` is set to `true` in config. The config surface is purely additive: the key is optional, defaults to false, needs no migration, and configs without it validate exactly as before (covered by the schema regression test).
- Users hitting the doubling bug on affected providers can enable the flag on that model and get stable streaming output without local workarounds. Enabling is per-model and reversible; disabling restores the exact historical append behavior.
- Tradeoff when enabled: content equality cannot distinguish an intentional long repetition (a model deliberately restating its whole answer so far) from a malformed cumulative frame, so such a restatement is suppressed once. When structured reasoning content splits a frame's visible text into runs, each run is judged independently: a run that restates the whole block is suppressed even when a later run adds new text, because the suppressed run itself contributes no new characters. The 8-character floor keeps short natural repeats (`"Ha"`, `"Ha"`) flowing, and partial restatements that add new text are never touched.
- Scope: the guard covers the `openai-completions` producer's visible-text path — the route of the field report and wire capture in #136262. The native DeepSeek plugin route's reasoning-text rendering duplication reported in the issue thread is a separate transport with its own producer boundary (thinking deltas) and is intentionally not addressed here.
- No changes to other providers, other transports, or agent-side accumulation.

## Configuration

Enable per model; the setting is reversible and requires no migration:

```json
{
  "models": {
    "providers": {
      "<provider-id>": {
        "models": [
          {
            "id": "<model-id>",
            "compat": {
              "dropCumulativeTextDeltaReplays": true
            }
          }
        ]
      }
    }
  }
}
```

Supported routes: the `openai-completions` transport in both direct and managed modes, covering its visible-text path — plain content deltas, structured content arrays, message-shaped frames, and visible reasoning details. The native DeepSeek plugin transport has its own producer boundary and is out of scope. Disabling the flag restores the exact historical append behavior.

Because the flag is a `compat` override, the standard route-ownership rules apply: bundled and catalog-known model routes resolve compat from the provider catalog, so an authored `compat` block on such a route does not take effect, and `openclaw doctor --fix` reports matching overrides for cleanup. Enablement is therefore effective on genuinely custom providers, custom models, or a catalog model routed to a different endpoint; extending it to a bundled route would be a catalog change. The option defaults to `false` and needs no migration, so existing configurations keep the current additive behavior.

## Evidence

- Focused tests (`packages/ai/src/transports/openai-completions-stream.replay.test.ts`) cover both the direct and managed transports:
  - a bare frame restating the whole accumulated text is dropped, and subsequent deltas keep appending;
  - the historical append behavior is preserved when the flag is off;
  - short exact repeats still append while enabled;
  - long repeats that do not restate the whole message still append while enabled;
  - a message-shaped frame equal to the accumulated text is dropped while enabled and appended while disabled;
  - a cumulative replay arriving after a native `tool_calls` chunk is dropped on both transports, and a legitimate post-tool-call delta restating the earlier block is kept;
  - a cumulative replay after a visible reasoning detail is dropped (reasoning-detail text feeds the same ledger);
  - a cumulative replay of the filtered visible text after inline reasoning tags is dropped (the ledger tracks the visible domain, not raw frames);
  - a cumulative replay after a queued post-tool-call reasoning detail is dropped (admission covers the queued path);
  - a multi-piece restatement frame (reasoning tags between the repeated prefix and further visible text) is dropped whole, and a frame whose repeated prefix is followed by genuine text is kept (the comparison unit is the whole frame's filtered visible contribution);
  - structured content array parts are classified as one frame: a repeated first part followed by a genuine second part is kept;
  - a frame split by structured reasoning content into a full-restatement run followed by a genuine new-text run keeps the new text while the restated run is suppressed, so the accumulated result matches the provider-intended text;
  - a frame ending in an incomplete tag keeps its released prefix (parser buffering), and a frame ending in a partial DSML tool token keeps its released prefix (recovery-stage buffering);
  - a replay hidden behind DSML tool markup is suppressed when the markup filters away, and the same wrapper followed by a genuine suffix is kept (classification runs on the filtered output).
- Released text followed by structured reasoning while tag syntax stays pending keeps its event order (text, then thinking, then text) with the option enabled and disabled, and both produce identical output (`openai-completions-stream.reasoning-and-buffering.test.ts`).
- The split-frame case is also asserted at the delivery-phase level: the interrupted run is tagged `commentary` and the genuine follow-up `final_answer`, so with the flag off and on the `final_answer` block is byte-identical and only the `commentary` block differs (it holds one copy of the repeated run instead of two);
- The config-loaded path is covered by a schema regression test: a provider model with `compat.dropCumulativeTextDeltaReplays: true` validates and preserves the flag (`src/config/config.schema-regressions.test.ts`).
- All `openai-completions` transport suites pass unchanged.
- Negative controls: each review round's regression cases were written before the fix and failed on the then-current head in exactly the way the reviewer traced (per-piece comparison dropping a repeated prefix, and the DSML-wrapped replay slipping through); the suite now covers 41 cases across the direct/managed matrix.
- Event trace through the real transport with the reported frame shape (synthetic SSE frames: three ordinary deltas of 35/30/26 chars, then a frame restating everything so far):

  Guard off (historical behavior — the replay frame is emitted, and additive consumers double the text):

  ```text
  seq=3 text_delta deltaLen=35 accumulatedBefore=0
  seq=4 text_delta deltaLen=30 accumulatedBefore=35
  seq=5 text_delta deltaLen=65 accumulatedBefore=65   <-- full-text replay frame
  seq=6 text_delta deltaLen=26 accumulatedBefore=130
  final emitted additive text length = 156 (clean length is 91)
  ```

  Guard on (same provider frames, flag enabled):

  ```text
  seq=3 text_delta deltaLen=35 accumulatedBefore=0
  seq=4 text_delta deltaLen=30 accumulatedBefore=35
  seq=5 text_delta deltaLen=26 accumulatedBefore=65
  final emitted additive text length = 91
  ```

- How the field symptom connects to incoming provider chunks: the report in #136262 captured the offending frame in the internal assistant event stream (`OPENCLAW_RAW_STREAM=1`), downstream of the transport. To bridge that gap, the reported frame shape — a single mid-stream `choice.delta` whose payload equals the whole accumulated text, with no new `text_start` — was driven through an unmodified v2026.9.3 build (provider transport and agent core) against a controlled provider endpoint. The frame passed through every layer unfiltered, and the live accumulated text ended at 470 characters versus 309 for the identical stream without that frame (the text accumulated up to the replay point is counted twice). The stock transport neither synthesizes nor filters this frame shape, so the field-captured event must have arrived from the provider side of the transport boundary; the one internal alternative, an SDK-level retry resending the full text, was ruled out separately — retry streams restart with a fresh `text_start` checkpoint, which the reported replay lacks. The trace above reproduces this shape end to end.
- A passive wire capture on the affected provider route (17 requests; provider and model identifiers withheld, frame shapes only) saw exclusively delta-shaped text frames — no message snapshots — and no replay within that window; the quirk is intermittent. Nothing observed on the wire marks a frame as a replay, and the field signature reaches the guard as an ordinary bare `text_delta`, so when a replay occurs it differs from a legitimate delta only by its content. Content equality at the producer is the only available trigger, which is why the guard is opt-in and length-floored rather than metadata-based.

## Config baseline note

`docs/.generated/config-baseline.counts.json` (core 2429 → 2430) and `docs/.generated/config-baseline.sha256` were regenerated mechanically via `pnpm config:docs:gen` on the merged tree. The +1 config-surface growth originates in main (#145724), which landed after the baseline there was last regenerated; this PR adds no configuration surface.


Richard355168:
@clawsweeper re-review

Yigtwxx (commented):
I measured this on the PR head `ab5e770cbe3` and found a defect class that is orthogonal to the opt-in tradeoff already discussed: **the guard's accumulator and its comparison point sit at opposite ends of the same pipeline, so they desynchronize as soon as a native tool call appears in the stream.** On the managed transport this makes the guard a no-op, and on both transports it makes the guard delete legitimate assistant text.

## Where the two ends are

- Comparison happens at the **entry** of the chunk loop, on the raw provider delta: `packages/ai/src/transports/openai-completions-stream.ts:541` (`replayGuard.shouldDrop(contentDelta.text)`).
- Accumulation happens at the **exit** of the append pipeline: `packages/ai/src/transports/openai-completions-stream.ts:226` (`replayGuard.observe(text)` inside `appendTextDeltaInternal`).

Between those two points the text passes through `reasoningTagTextPartitioner.pushVisible()`, `appendFilteredVisibleTextDelta` (DeepSeek recovery/filtering), and `appendVisibleTextDelta`. That last one is the important one:

```ts
// openai-completions-stream.ts:267-276
const appendVisibleTextDelta = (text: string) => {
  if (!text) return;
  if (currentBlock?.type === "toolCall" && !directMode) {
    queuePostToolCallDelta({ kind: "text", text });   // <- never reaches observe() yet
  } else {
    appendTextDelta(text);
  }
};
```

`flushPendingPostToolCallDeltas` returns early while `currentBlock?.type === "toolCall"` (`:238-245`), and `currentBlock` only stops being a tool call when a *different* tool call block opens (`:577-580`) or at end of stream (`:668`). So in non-direct mode, every visible delta after a native tool call sits in `pendingPostToolCallDeltas` until the stream ends, while `shouldDrop` keeps comparing against a frozen `messageVisibleText`.

Separately, `messageVisibleText` is **message-scoped** and never resets, while the contract this guard cites is **block-scoped** ("consumers rebuild text by replaying deltas from the latest start/end checkpoint"). After a tool call a new text block starts with an empty checkpoint, but the guard still holds the previous block's text.

## Measured consequences

I added a probe file next to your tests, same harness, same two arms (`direct` = `streamOpenAICompletions`, `managed` = `createOpenAICompletionsTransportStreamFn()`), with one `tool_calls` chunk inserted after the first delta.

**Probe A — your exact replay shape, but preceded by a tool call, flag on:**

```
chunks: [ text A ] [ tool_calls ] [ text B ] [ text A+B (the replay) ] [ text C ] [ stop ]
```

| Arm | Output | Guard effective? |
|---|---|---|
| direct | `A B C` | yes |
| **managed** | `A B A B C` | **no** |
| managed, flag **off** (baseline) | `A B A B C` | n/a |

On the managed transport the output with the flag on is byte-identical to the output with the flag off. This is the transport the Gateway and agent runtime actually use, and `text -> tool_call -> text` is the ordinary agent turn shape, so the doubling this PR targets survives in the case where it is most visible in live channels.

**Probe B — legitimate post-tool-call text, flag on:**

```
chunks: [ text A ] [ tool_calls ] [ text A ] [ text C ] [ stop ]
```

| Arm | Expected | Actual |
|---|---|---|
| direct | `A A C` | **`A C`** |
| managed | `A A C` | **`A C`** |

Here the second `A` is not a cumulative replay of anything: the current text block is empty when it arrives. The guard drops it purely because it still equals the *previous* block's text. 35 characters of real model output are removed from both the live events and the final assistant message, with no diagnostic. This is the "intentional repetition loss" risk in a form that needs no unusual provider behavior at all, only a tool call.

**Mutation arm** (forcing `shouldDrop` to `return false && ...` in `openai-transport-shared.ts`):

- Your `openai-completions-stream.replay.test.ts`: **6 of 14 fail** -> the harness is sensitive to the guard where you tested it.
- Probe A managed: unchanged (`A B A B C` before and after) -> confirms the guard was already inert there, not merely masked.
- Probe B: both arms become `A A C` -> confirms the guard, not the harness, is what deletes the text.

Control arm before any of this: `openai-completions-stream.replay.test.ts` **14/14 pass** on the unmodified PR head.

```
node node_modules/vitest/vitest.mjs run \
  --config test/vitest/vitest.unit-support.config.ts \
  packages/ai/src/transports/openai-completions-stream.replay.test.ts
```

## Why the current tests do not see it

Every case in `openai-completions-stream.replay.test.ts` is a pure text stream. The post-tool-call queue is the only thing that separates the two arms you already parametrize, so the `direct`/`managed` split is currently proving less than it looks: both arms take the same path until a tool call exists. Adding one `tool_calls` chunk to `replayChunks()` splits them, as above.

## Suggested direction

The fix seems to be making the accumulator mirror the *comparison unit* rather than the append sink:

1. Advance the accumulator at the point where the frame is accepted (right after the `shouldDrop` check in the chunk loop), so buffering downstream cannot desynchronize it; and
2. Reset it on each `text_start`, so "the whole accumulated text" means the same thing to the guard as it does to the consumers rebuilding from the latest checkpoint.

That also narrows the heuristic in a useful way: a frame is only suspicious if it restates the block it belongs to, which removes the Probe B false positive without weakening the case you actually reported. Whether the scope should be the block or the message is your call, but right now it is neither consistently: the write is per-message and the read is per-frame.

Two smaller notes:

- The drop is completely silent. For an opt-in compat guard that removes model output, a debug-level diagnostic naming the dropped length would make field reports diagnosable; the streaming path already has `diagnostic-run-activity` wired in.
- `reasoningTagTextPartitioner.pushVisible()` can also hold text back (`hasPending()` / `hasPendingSyntax()`). I did not measure that path, so I am raising it as a question rather than a finding: for a provider that both emits inline reasoning tags and replays full content, does the accumulator stay in step?

The producer-boundary placement and the off-by-default posture both look right to me; this is about the guard's internal bookkeeping, not about whether the guard should exist.

Yigtwxx (commented):
Re-measured `966894a0a0d` / `0013a244f4f` independently, with the same probes I used for the first review rather than the tests added here. Both findings are closed, and the arm split that exposed them is gone.

| probe (flag on) | before | now, `direct` | now, `managed` |
|---|---|---|---|
| replay frame after a native `tool_calls` chunk | managed byte-identical to flag-off | `A B C` | `A B C` |
| legitimate post-tool-call delta equal to the earlier block's text | `A C` - 35 chars deleted, both arms | `A A C` | `A A C` |
| original no-tool-call replay (regression check) | `A B C` | `A B C` | `A B C` |
| replay of the whole accepted text after a tool call | - | dropped | dropped |
| first delta of a new block equal to all accepted text | - | kept | kept |

Moving the ledger to frame acceptance is the right seam: the comparison and the write now happen on the same value, so no amount of downstream buffering can desync them, and `acceptedText.length > textBlockStartLength` gives the block scope the contract already implied. `openai-completions-stream.replay.test.ts` is 20/20 on the new head. The `log.debug` on the drop path is a good addition - a compat guard that removes model output should leave a trace.

One residual, same class, different feeder.

The ledger is fed only from the `contentDeltas` loop. Visible text that arrives as a reasoning detail takes another path into the same block and never reaches `admitTextDelta`:

```ts
// openai-completions-stream.ts - appendReasoningDeltas
if (reasoningDelta.kind === "text") {
  appendTextDelta(reasoningDelta.text, reasoningDelta.source);   // source: "reasoning_detail"
}
```

Those deltas are produced by `readOpenAICompletionsReasoningBatch` for any `detail.type` in `visibleReasoningDetailTypes` (`openai-transport-shared.ts:157-163`), and they are visible text in the current block - so `acceptedText` is missing part of the block it is meant to describe.

Measured on the managed transport, flag on, `compat.visibleReasoningDetailTypes: ["response.output_text"]`:

```
chunks: [ reasoning_details: [{type:"response.output_text", text:A}] ]
        [ content: B ]
        [ content: A+B ]        <- cumulative replay of the whole visible text
        [ stop ]

result: "A B A B"    (the replay was appended)
```

The guard compares `A+B` against an `acceptedText` of just `B`, so the lengths differ and the frame passes. For a provider that emits visible reasoning details and also replays full content, the guard is inert on exactly the shape it exists for.

This fails safe - it under-triggers, never deletes text - so it is not a blocker, and it is much cheaper to close now than before: the same `admitTextDelta` call in front of the `reasoning_detail` text branch would put both feeders on one ledger. Worth a line in the doc comment either way, since "the comparison unit is the current text block" is not quite true while one feeder bypasses it.

Nothing else from my first review is outstanding.

Richard355168 (commented):
Thank you for the independent re-measurement — running the original probes against the fixed heads is exactly the rigor this PR needed, and your table matches our own regression runs.

**Closing the loop on the fix shape.** The two findings were reproduced on the pre-fix head with a negative control before the fix landed: the probe cases added to `openai-completions-stream.replay.test.ts` fail on the unmodified head in exactly the three ways you measured (the managed-arm replay no-op and the post-tool-call deletion on both arms — 3 of 20 suite cases), and pass on the fixed head. Your ordering note was correct and worth expanding: a literal reset on each emitted `text_start` does not survive the managed transport's deferred flush, because queued post-tool-call text materializes (and its `text_start` fires) only at the next tool-call switch or at end of stream — resetting on emit would either wipe the ledger before the replay arrives or fire after the fact. The boundary is therefore armed when a frame is accepted while the current block is not text (or the visible-text source changed) and cleared when the new block materializes; the observable semantics match your suggestion.

**The residual is closed.** Visible reasoning details now route through the same admission seam as content deltas — an `admitVisible(text, source)` wrapper feeds the guard for both feeders, so the ledger describes the whole block. Your probe is now a regression case in the same suite: with `compat.visibleReasoningDetailTypes: ["response.output_text"]`, a reasoning-detail + content + full-text-replay stream ends at the clean length on both transports (22/22 across the direct/managed matrix). The guard's JSDoc now states that every visible-text feeder must route through the seam.

**On the smaller notes:** the drop path logs a debug diagnostic naming the dropped and accepted lengths, and the partitioner question dissolves by construction — the ledger advances before the reasoning-tag partitioner holds or routes anything, so it cannot desync the comparison.

Nothing else is outstanding on our side either. The remaining work is the opt-in tradeoff decision itself.

Yigtwxx (commented):
I ran the exact scenario from the open finding ("Classify the entire provider frame across reasoning transitions") against this head, `690ded28415`, through the real transport rather than reading the loop, and I read the resulting text blocks' delivery phase instead of only their concatenated text. The result does not reproduce a loss of delivered text.

**Probe.** A loopback SSE server feeds the same three frames the finding describes, through both production entrypoints (`streamOpenAICompletions` and `createOpenAICompletionsTransportStreamFn()`), with `reasoning: false` exactly as the new test does:

```
chunk 1: content = A
chunk 2: content = [ {text: A}, {thinking: "Recheck."}, {text: "Additional detail."} ]
chunk 3: finish_reason = stop
```

I then printed `result.content` with each block's `textSignature`.

| arm | `dropCumulativeTextDeltaReplays: false` | `dropCumulativeTextDeltaReplays: true` |
|---|---|---|
| direct | block 1 `A A` phase `commentary`; block 2 `Additional detail.` phase `final_answer` | block 1 `A` phase `commentary`; block 2 `Additional detail.` phase `final_answer` |
| managed | same | same |

**What that shows.** In this scenario the repeated run is not what carries `A` to the reader in either arm. The structured reasoning part interrupts the open text block, the following genuine text confirms the interruption, and `tagInterruptedTextPhases` tags the preceding block `commentary`; the `final_answer` phase is `Additional detail.` with the flag off and with the flag on. The guard's only observable effect here is that the commentary block holds one copy of `A` instead of two. So the enabled path drops a duplicate, not a delivered prefix.

**Two caveats, so this is not read as more than it is.**

1. This measures the finding's own scenario. It does not settle the design question of whether the comparison unit should be the whole provider frame or the run between reasoning transitions; it only shows that the run-level decision does not remove delivered characters in the case cited as evidence for the frame-level rule.
2. The new test cannot carry this argument by itself. `runStream` concatenates every `type: "text"` block regardless of phase, so a green assertion of `TEXT_A + "Additional detail."` mixes an interim commentary block with the final answer, and it would stay green whether or not the phase split moved. If you want the test to answer the finding rather than sit beside it, assert the phase: with the flag on and off, `final_answer` text should be identical, and only the `commentary` block should differ. That is the invariant your comment in the test is actually claiming.

Reproduction is the harness already in `openai-completions-stream.replay.test.ts`, plus a print of `block.textSignature`; I ran it with `node --import ./scripts/tsx.mjs` against this head, both arms, both entrypoints.

Richard355168:
Thanks for running the scenario through the real transport and reading the delivery phases — that is exactly the invariant the run-level policy was claiming.

The follow-ups are now on the branch (head `a78d3735467`):

1. **Phase-level assertion** — the full-restatement case runs both arms on both entrypoints and asserts block phases: with the flag off the interrupted `commentary` block holds the repeated run twice, with the flag on once, and the `final_answer` block is byte-identical in both arms.
2. **Pending-syntax ordering** — released text followed by structured reasoning while tag syntax stays pending keeps its event order (text → thinking → text) with the option enabled and disabled, producing identical output in both modes.
3. **Configuration guidance** — the PR body now documents that the flag is a `compat` override and follows the route-ownership rules: bundled and catalog-known routes resolve compat from the catalog, so authored compat does not take effect there (`openclaw doctor --fix` reports matching overrides); enablement applies to custom providers, custom models, or catalog models routed to a different endpoint.

The comparison-unit question stays with maintainer judgment — the body documents the run-level unit and its tradeoff, and the new phase test encodes the invariant that motivated it.

Yigtwxx (commented):
Re-measured the head my last review asked about, `67e9cecb00e`, on two questions: what the range since `a78d3735467` actually contains, and whether the new phase test carries the claim I said the old one could not.

**What changed since my last review.** Nothing in the PR's own surface. Comparing `a78d3735467...67e9cecb00e` through the API, no file under `packages/ai/`, `packages/llm-core/`, or `src/config/` appears in the range - every commit in it is a main landing, so the range is a refresh, and `a78d3735467` itself is the only behavior-relevant commit added after my last read. `mergeStateStatus` is `CLEAN` on that head.

**The caveat is closed.** `keeps the delivered final answer identical whether the restated run is suppressed` asserts what my comment claimed the earlier assertion only implied: it reads `textSignature`, splits blocks by phase, and pins all three facts separately - `commentary` is `TEXT_A` with the flag on and `TEXT_A + TEXT_A` with it off, and the `final_answer` text is equal across the two arms *and* equal to `["Additional detail."]`. The last of those matters more than it looks: `replayTextBlockPhase` returns `undefined` for an unparseable signature, so a phase split that silently disappeared would leave both filters empty - and the explicit content assertions turn that into a failure instead of a vacuous pass. The test cannot go green by losing the thing it measures.

**Mutation, to check the test holds the mechanism and not just this head's output.** I removed the per-transition `settlePlan()` in the content loop (`openai-completions-stream.ts`, the `beginReasoning(hasLaterVisible)` branch), which reverts classification to the whole frame - the alternative the open finding argued for. On `67e9cecb00e`, with that single edit:

| suite | clean | mutated |
|---|---|---|
| `openai-completions-stream.replay.test.ts` | 41 pass | 4 fail |
| `openai-completions-stream.reasoning-and-buffering.test.ts` | 12 pass | 3 fail |

7 failures total, and the useful detail is which ones. Besides the two arms of the new phase test and `suppresses a full-restatement run and keeps the frame's new text while enabled`, the mutation fails `orders released text before structured reasoning while tag syntax stays pending with replays disabled` - the flag-*off* arm. So the frame-level variant does not merely relax the guard; it moves block structure on streams where the compat flag is not set at all, which is the property `b5e25732af9` was written to protect and the reason the run-level unit is the correct one here. The first failing assertion in the new test is on the disabled arm's commentary block (`"A A"` becomes `"A"`), which is the same signal from the other side.

Run as `node node_modules/vitest/vitest.mjs run --config test/vitest/vitest.unit-support.config.ts <file>`, clean arm before and after the mutation to confirm the tree was restored (53/53).

Nothing outstanding from my three earlier reviews.

Yigtwxx (commented):
Checked `798a853c298` against my last note on `67e9cecb00`. The PR surface is unchanged (the `merge-base..head` diff is byte-identical apart from `config-baseline.json` hunk offsets and the core count that `main` regenerates), so nothing new to review on the code. Two notes on the red run `34908051821`, since the branch has been retriggering CI:

- `checks-node-compact-large-4`: `src/gateway/worker-environments/live-chat.test.ts` > "does not resurrect a capped completed prefix when the active worker snapshot shrinks". That is the flake steipete fixed in #148628 (`21cd7d8c4d1`, "await worker preview reception before cap assertions"), which landed on `main` at 00:21 UTC on the 15th, about an hour after this head's last merge from `main` (`798a853c298`, 23:15 UTC on the 14th). It is not in this branch yet, so an empty retrigger will keep hitting it; one more merge from `main` picks the fix up.
- `checks-node-compact-small-14`: `test/scripts/test-projects-build-admission.test.ts` > "blocks Gateway selective exclusion CLI readers until successful build and preserves SIGTERM". The shard log shows the vitest fork-shutdown fixtures exiting `143` with `ETIMEDOUT` (`vitest-fork-cancellation-*` under `.artifacts/vitest-fork-shutdown`), i.e. the runner did not tear the fixture down inside the window. Nothing in this PR touches `test/scripts` or the gateway worker path.

Neither red is on the openai-completions surface. Still nothing outstanding from me.

Yigtwxx (commented):
Three reds on `8334416137a` (run started 2026-09-18 03:26Z); one of them is this branch's own and needs a change, the other two are inherited.

**Own: `checks-fast-baseline-ratchets`.** `check-line-cap-ratchet.mts` rejects `src/config/zod-schema.core.ts: 783 -> 784 counted lines (cap 700)`. The single added line is `dropCumulativeTextDeltaReplays: z.boolean().optional()` in the `compat` block (`zod-schema.core.ts:358`). This gate did not exist when the branch was last green: #149622 (`c1f00edb04c`, 2026-09-16) started rejecting any growth of an over-cap file on pull requests, so a main merge will not clear it. The ratchet's own instruction is "Extract a coherent sibling module; never trim coverage or disable max-lines" — the `compat` object schema is a self-contained candidate (`src/config/zod-schema.model-compat.ts` next to the existing `zod-schema.agent-model.ts`, re-imported from core), which takes core well under its current count instead of adding to it. While touching that block: every neighbouring flag carries a JSDoc comment describing the behaviour and default; the new one is the only bare entry.

**Inherited: `checks-node-compact-large-1`.** `test/vitest-projects-config.test.ts > covers each normal full-suite test file exactly once` lists `extensions/codex/src/app-server/run-attempt.native-config.test.ts` and `run-attempt.subscription.test.ts` as uncovered. Both files landed on main in #151001 (`3f0b58e2696`, 02:44Z) and main registered them in `vitest.extension-database-workers-paths.mjs` with #151344 (`757e3c9d3ef`, 04:01Z); this run sits between the two. A main merge clears it.

**Inherited: `checks-node-changed-extensions-bundle-10`.** Shard `changed-extensions-config-74` reports `[vitest-pool-runner]: Timeout waiting for worker to respond` with no failing test in the log (72/18/11/11 files passed on the neighbouring shards). Infra, not the diff.

The production surface is unchanged since my earlier review; the remaining items in the ClawSweeper verdict are the maintainer decision on the default-off heuristic, not code defects.

Richard355168:
Done: `ModelCompatSchema` (with its private routing schemas) moved to `src/config/zod-schema.model-compat.ts`, re-imported from core, and the new flag now carries the JSDoc its neighbours have. `zod-schema.core.ts` is back under the ratchet cap, so `checks-fast-baseline-ratchets` should go green on this head; the fresh `main` merge also picks up the `compact-large-1` registration fix. Verified locally with the ratchet check and the targeted vitest suites (schema + compat + replay); full lanes on CI. No schema behavior change.

Yigtwxx (commented):
Re-checked `3d5a6124ccd` after the compat-schema extraction. The move itself is faithful: diffing the 173 lines removed from `zod-schema.core.ts` against the new `zod-schema.model-compat.ts` leaves only the `export` keyword, the two imports the block needs, and the new JSDoc for `dropCumulativeTextDeltaReplays`; `ModelCompatSchema` has no other consumer, so the surface is unchanged. `checks-fast-baseline-ratchets` is green again since `6353c5ed591`.

The extraction did introduce one red of its own, and since it was my suggestion I should have flagged this with it. `zod-schema.core.ts` is in the eager runtime import closure (`scripts/pr-lib/wrapper-components.txt:467`), and the closure guard requires every relative import reachable from an inventoried file to be inventoried too. The new module is not, so:

- `checks-node-compact-small-28` -> `test/scripts/eager-import-closure.test.ts`: `src/config/zod-schema.core.ts: unresolved ./zod-schema.model-compat.js`
- `checks-node-compact-small-36` -> `test/scripts/pr-worktree-provision.test.ts:37`: `expected [ "src/config/zod-schema.model-compat.ts" ] to deeply equal []`, plus the three provisioning cases that fail on the same `Cannot find module .../canonical/src/config/zod-schema.model-compat.js`

This pair has been red on every run since `955d9ef6ac0` (18.09 11:46Z onward: small-39/32, then 37/27, 40/29, now 28/36 as the shard plan moves); it is just easy to miss behind the changing shard numbers. Fix is one line, in the file's C-sort order:

```
--- a/scripts/pr-lib/wrapper-components.txt
+++ b/scripts/pr-lib/wrapper-components.txt
@@ -476,2 +476,3 @@
 src/config/zod-schema.messages.ts
+src/config/zod-schema.model-compat.ts
 src/config/zod-schema.node-host.ts
```

Measured on a local merge of this head with today's `origin/main` (clean merge; `02d37b5221d` is the CI base of the red run), Node 24.21.0, both files under `vitest.tooling.config.ts`:

- without the entry: 2 failed / 5 passed, and the `eager-import-closure` error text is byte-identical to the CI one above
- with the entry: 2 files passed, 7 passed / 23 skipped (the provisioning cases are POSIX-only, so on Windows only the closure assertion runs; on CI both went red on the same missing entry)

Same pattern #132607 hit last week with `model-provider-local.ts`, for what it's worth.

The third red, `checks-node-compact-large-9`, is `src/sessions/session-lifecycle-diagnostics.test.ts` failing its `afterEach` `expect(vi.getTimerCount()).toBe(0)` (`:111`). This branch does not touch `src/sessions/`, main has not changed that test since 12.09, and the file passes 10/10 on the same local merge, so I read it as a timer-leak flake that the next run should clear. Worth a second look only if it reappears after the inventory commit.

Richard355168:
Pushed `c4e0f6fb8ac` — the entry is in at `wrapper-components.txt:468`, C-sort order as suggested.

Local verification: `eager-import-closure.test.ts` is green with the entry under the tooling config.

One disclosure: two `pr-worktree-provision.test.ts` cases ("cold provisioning", count/parameters transport) fail on this machine with a pre-existing `dubious ownership` error on the `/tmp` fixture repos — reproduced with the change stashed, so it reads as local-environment, not branch-related; deferring to CI for those.

Agreed on `large-9` being a timer-leak flake — watching only, will revisit if it survives the inventory commit.

Yigtwxx (commented):
Confirming `c4e0f6fb8ac` closes the inventory red I raised: the entry sits at `scripts/pr-lib/wrapper-components.txt:468` in C-sort order between `zod-schema.messages.ts` and `zod-schema.node-host.ts`, and on the current head `e8d39e27ce7` the three shards that were red are green — `checks-node-compact-small-28` (eager-import-closure), `checks-node-compact-small-36` (pr-worktree-provision) and `checks-node-compact-large-9` (the timer-count flake) all completed successfully in the 04:40Z run, with `checks-fast-baseline-ratchets` and `openclaw/ci-gate` green as well. That also settles the `dubious ownership` disclosure: it was the local `/tmp` fixture, not the branch. The PR surface is unchanged since my last read apart from that one line, so nothing further from my side.

PR #142896, opened 2026-09-09
fix(agents): retain model context through tool inventory projection

steipete:
Related: #140674

## What Problem This Solves

Dynamic tool inventory released its prepared model lease before provider callbacks projected the tool schemas. The Gateway then cached that executable model after release. MCP projection could also miss the selected provider's normalizer because it ran outside that prepared generation.

## Why This Change Was Made

An internal acquisition now keeps the original model lease and runs projection inside its exact generation. The informational command, session-reset check, and Gateway consumers release it after projection. The Gateway cache stores inventory data; a nonempty MCP projection acquires its workspace-specific model context when needed.

Static resolution, stale/background refresh, coalescing, and read-only MCP behavior remain unchanged. The handle rejects use after its own release without introducing a new snapshot-publication freshness policy. No public SDK, configuration, schema, dependency, or version changes.

## User Impact

Tool inventories use the intended provider's schema normalization while its resources are owned, including sandbox MCP workspaces. Unrelated ambient providers cannot substitute for the selected generation. Empty or filtered MCP catalogs preserve the base inventory without additional acquisition.

## Evidence

A native failing-before test observed zero prepared owners inside the actual provider normalization hook. The repaired path observes one, then releases it after projection. All 87 focused tests pass, covering native same/sandbox MCP scopes, competing ambient providers, errors, cache behavior, and command/reset callers. The canonical changed gate, 207.2-second ordinary build, and fresh full-scope independent Codex review pass. A callback config-narrowing error found by the first gate was repaired by capturing the existing config value, and the final reset cases were rerun.

A real built Gateway and two built CLI tools.effective calls returned identical inventories containing the synthetic plugin tool; the dynamic provider normalizer ran. Readiness took 4.56 seconds and the CLI calls 0.80 / 0.61 seconds, including process startup and connection. The owned Gateway drained and exited zero. The initial fixture was missing its required tools manifest contract; only that synthetic fixture was corrected.

Paired in-process handler measurements on the same base used 25 alternating samples after five warmups. With Node 24.20, uncached base+MCP took 6.35 → 6.62 ms, warm inventory without MCP 0.118 → 0.118 ms, and warm MCP 0.25 → 1.44 ms. The latter pays for one real acquisition and the previously missed selected normalizer. Node 26 showed a similar approximately 1.3 ms warm-MCP cost. These are bounded synthetic measurements, not a general speedup claim.

All runtime proof used isolated synthetic state and no external provider requests. General prepared/cache physical disposal remains separate work.

### DIFF
--- a/src/agents/tools-effective-inventory.cold-provider.test.ts
+++ b/src/agents/tools-effective-inventory.cold-provider.test.ts
@@ -1,6 +1,7 @@
 import fs from "node:fs";
 import path from "node:path";
 import { expectDefined } from "@openclaw/normalization-core";
+import { Type } from "typebox";
 import { describe, expect, it, vi } from "vitest";
 import { setRuntimeConfigSnapshot } from "../config/config.js";
 import { applySessionEntryLifecycleMutation } from "../config/sessions/session-accessor.js";
@@ -9,6 +10,7 @@ import {
   createToolsEffectiveHandlers,
   testing,
 } from "../gateway/server-methods/tools-effective.js";
+import { toolsEffectiveTestDependencies } from "../gateway/server-methods/tools-effective.test-support.js";
 import type { GatewayRequestContext, RespondFn } from "../gateway/server-methods/types.js";
 import { planEffectiveModelCatalogRows } from "../model-catalog/index.js";
 import {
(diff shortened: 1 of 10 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d050608fba764b8a2b5ada1412a1c42c6a51a53d](https://github.com/openclaw/openclaw/pull/142896/commits/d050608fba764b8a2b5ada1412a1c42c6a51a53d)
- Landed commit: [34e0e5d7e53cf1b8119354f1736f1ae067ac9775](https://github.com/openclaw/openclaw/commit/34e0e5d7e53cf1b8119354f1736f1ae067ac9775)

PR #142899, opened 2026-09-09
refactor(tests): consolidate duplicate setup inference cases

steipete:
Related: #139428

## What Problem This Solves

Setup inference tests repeat a populated-partial timeout worker run and an injected owner-error scenario already covered through real credential drift.

## Why This Change Was Made

Move the complete partial-result equality assertion into the existing HTTP-responsiveness case, then remove the identical partial-timeout case. Move disposition-absence and generic owner-error assertions into the real credential-drift case before removing its injected-error duplicate. Preserve all other scenarios, worker deadlines, runtime calls and production behavior.

## User Impact

No user-visible behavior change. Two redundant cases and 52 net test lines are removed while their meaningful observations remain in stronger existing cases.

## Evidence

- Both complete owner suites passed before and after: 174 cases before, 172 after, with exactly the two named cases removed and every retained case in the same per-file order.
- Three focused temporary owner mutations failed at the intended retained assertions: missing partial-result field, unexpected disposition property, and missing generic owner-error context. All production bytes were restored exactly.

The removed timeout case used a separate three-second worker deadline. No measured runtime improvement is claimed from the before/after runs because cache state differed.

### DIFF
--- a/src/system-agent/setup-inference-detection.test.ts
+++ b/src/system-agent/setup-inference-detection.test.ts
@@ -137,16 +137,7 @@ describe("isolated setup inference detection", () => {
     expect(JSON.parse(response.body)).toEqual({ ok: true, status: "live" });
     expect(elapsedMs).toBeLessThan(500);
     const detection = await pending;
-    expect(detection).toMatchObject({
-      candidates: fallback.candidates,
-      unavailableCandidates: fallback.unavailableCandidates,
-      manualProviders: fallback.manualProviders,
-      authOptions: fallback.authOptions,
-      recommendedInstalls: fallback.recommendedInstalls,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [adb87f91adc2657537f59cb3321903c22a18ed0a](https://github.com/openclaw/openclaw/pull/142899/commits/adb87f91adc2657537f59cb3321903c22a18ed0a)
- Landed commit: [6ae13a3e45db3b92f380dec231efa4224d6eed59](https://github.com/openclaw/openclaw/commit/6ae13a3e45db3b92f380dec231efa4224d6eed59)

PR #142903, opened 2026-09-09
refactor(tests): advance the captured ngrok startup deadline

steipete:
## What Problem This Solves

The real ngrok startup-timeout process test spends 30 seconds waiting before it can exercise the child-termination behavior it protects.

## Why This Change Was Made

Capture the actual native startup timer during the synchronous `startTunnel` call and assert its configured 30,000ms budget. Restore the call-through spy immediately, wait for the real signal-resistant child to report readiness, then cancel and invoke that exact startup callback once. The real two-second SIGTERM grace, one-second bounded close wait, process-exit assertions and 40-second test deadline remain unchanged.

## User Impact

No production behavior changes. This test-only change adds 27 lines and avoids the controlled startup wait in one process test.

## Evidence

- Both complete process and unit suites passed before and after: 19 cases, zero pending, identical ordered names and statuses.
- The target process case took 32.014s before and 2.054s after in one local pair. This is a single-case result, not an overall suite-runtime claim. The baseline external PID observer missed Vitest’s nested temporary namespace; its original process assertions passed. The corrected candidate observer recorded the two PID-file fixtures and confirmed they were dead after the run.
- A temporary 30,001ms budget failed the new argument assertion. Disabling forced termination failed the existing process-exit assertion. Both controls’ observed PID fixtures were dead after cleanup, and production source was restored exactly.
- The mapped changed gate, deslop pass and fresh P2 review passed.

The initial revision cancelled the cleanup deadline when PID discovery failed. The follow-up removes that unconditional cancellation: the accelerated success path still cancels its timer, while failed PID discovery retains the original native termination fallback. A real-child control held cleanup for 34 seconds before runner teardown: the initial revision left its child alive and required a safety kill; the repaired revision had already terminated its child through the original deadline, without a safety kill. Both controls retained the intended synthetic failure and left their observed PID fixture dead.

### DIFF
--- a/extensions/voice-call/src/tunnel.process.test.ts
+++ b/extensions/voice-call/src/tunnel.process.test.ts
@@ -221,7 +221,6 @@ describe.skipIf(process.platform === "win32")("voice-call tunnel child process",
         .toBe("SIGTERM");
       expect(await waitForProcessExit(childPid, 1_000)).toBe(true);
     } finally {
-      clearTimeout(startupTimer);
       process.env.PATH = previousPath;
       if (previousPidPath === undefined) {
         delete process.env.OPENCLAW_NGROK_PID_FILE;
--- a/extensions/voice-call/src/tunnel.process.test.ts
+++ b/extensions/voice-call/src/tunnel.process.test.ts
@@ -2,7 +2,7 @@ import fs from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { withEnvAsync, withTempDir } from "openclaw/plugin-sdk/test-env";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { startTunnel } from "./tunnel.js";
 
 function isProcessAlive(pid: number): boolean {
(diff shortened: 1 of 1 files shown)


steipete:
Fixed the PID-discovery cleanup regression in the follow-up commit by removing the unconditional outer-finally timer cancellation. The normal accelerated path already cancels before invoking the callback; failed PID discovery now retains the original native deadline.

Verified with the same real-child failure control before runner teardown: after readiness, PID discovery rejects before assignment, and cleanup waits 34 seconds. The old revision still had a live child with no SIGTERM and needed the safety kill; the repaired revision had already received SIGTERM and exited, without a safety kill. Both runs failed only with the injected discovery error and left the observed child dead. All 19 normal process/unit cases also pass unchanged on the repair.

steipete:
Merged via squash.

- Prepared head SHA: [82f7368a01cab39a5a2ef262ea1ba121427d1e50](https://github.com/openclaw/openclaw/pull/142903/commits/82f7368a01cab39a5a2ef262ea1ba121427d1e50)
- Landed commit: [3d2d68e8783abc7211493b4bbc644b3548d269e4](https://github.com/openclaw/openclaw/commit/3d2d68e8783abc7211493b4bbc644b3548d269e4)

PR #134038, opened 2026-08-31
perf(ci): remove repeated setup and balance complete test workloads

steipete:
## What Problem This Solves

CI repeats checkout, fixture compilation and dependency work, while its planner packs long suites using estimates that omit their prerequisites. The earlier complete green integration run took 19m 22s; the final native run takes 17m 31s. This change removes repeated work and balances the complete workload; whole CI below five minutes remains the target.

## Why This Change Was Made

The canonical planner charges each job's strongest runtime prerequisite once, uses measured file costs, separates expensive tooling work and partitions the complete plugin catch-all. Windows keeps two jobs and two project slots, admits expensive selections first, and moves the cron process-identity test to its actual runtime owner so it no longer triggers full-package E2E preparation. Compiled session workers and native Buffer comparison remove repeated compilation and per-byte JavaScript work without changing SQLite data or schemas.

Checkout reuses the authenticated candidate index when the harness SHA is identical, including preflight. Different pinned revisions retain Actions checkout and its retry contract. Historical acquisition, authentication and cancellation stay with their existing owner. Hybrid macOS Node uses its existing hosted macOS 15 route with the same 19 files, two workers and deadline; macOS 26 siblings are unchanged.

Dependency restores use the actual runner environment and include pnpm registry metadata. The protected warmer remains the sole shared publisher. Standalone UI consumes its transform seed through the existing executor, preserving root Node maxWorkers=3 and Chromium's separate pool policy. Go follows the selected docs-translation consumer, reducing hybrid admissions from 17 to two. Scoped discovery, shared CPU/RAM policy for lint children, cancellation, docs classification and Testbox base preparation remove other repeated work.

The wrapper merge retains main's complete trusted source inventory, adds the planner's shared metadata dependency, and exposes only the four existing third-party tooling packages through native directory links. The obsolete fixture copier is removed. Package timeout proof waits for child readiness before expiring its captured deadline, then joins command and readiness outcomes before cleanup; actual process termination remains real. No production timeout is increased.

## User Impact

Existing test inventories, trust boundaries, cache authority and runner limits are preserved. There is no new runner capacity, paid service, configuration option, persistent-store design, schema or protocol. Fresh protected cache keys remain cold until their producer lands and warms them.

This incorporates #134096, #134192, #134196, #133994, #133931 and #134279; they remain open until integration merge proof. Main's Linux runtime-only pretest (#134294) and complete UI file reporter (#134283) are preserved. The Windows prerequisite repair is part of this PR, not a result attributed to #134294.

[...]

PR #142912, opened 2026-09-09
improve(test): reduce startup work for focused fast tests

steipete:
## What Problem This Solves

Focused runs through an explicit fast Vitest configuration classify unrelated test source files before applying the command-line selection, adding startup work even for a single test file.

## Why This Change Was Made

The ordinary, isolated, and fake-timer fast configurations now pass the existing CLI selection into source discovery. The existing owners still decide final test membership, ENV precedence, isolation, timer ordering, and full-inventory caching.

Related: #134038, which introduced the selected-analysis APIs reused here.

## User Impact

Developers running focused fast tests avoid unnecessary source reads. Full runs and the selected tests' execution settings retain their behavior.

## Evidence

The identical regression fails before the change with 123 unrelated source reads and passes afterward with zero. All 173 owner, sibling, and pattern-selection cases pass, including later full-inventory behavior. Fresh independent review found no actionable findings.

Four fresh-process config-only measurements on macOS with Node 26.8.1, ordered before/after/after/before, retained identical selection and one Git inventory call per process. Unrelated test-body reads fell from 5,735 to zero; measured config import times were 1.11–1.87 seconds before and 0.34–0.35 seconds after. These are descriptive config-import measurements, not end-to-end test timings or a universal speed guarantee.

All 15 canonical changed-file checks passed after a clean full-workspace frozen install on Linux with Node 24.19.0. The initial local typecheck failed because the borrowed dependency layout lacked plugin-local packages; that failure is retained, and the clean-install run passed the full type graph without source changes. Required [hosted CI](https://github.com/openclaw/openclaw/actions/runs/34322160501) passed on this exact commit.

### DIFF
--- a/test/vitest-unit-fast-config.test.ts
+++ b/test/vitest-unit-fast-config.test.ts
@@ -156,11 +156,20 @@ describe("unit-fast vitest lane", () => {
       try {
         const includeFile = path.join(directory, "include.json");
         fs.writeFileSync(includeFile, JSON.stringify(selectedTests));
-        process.env.OPENCLAW_VITEST_INCLUDE_FILE = includeFile;
-        const selections = [];
-        for (const name of ["unit-fast", "unit-fast-isolated", "unit-fast-fake-timers"]) {
-          const { default: config } = await import("./test/vitest/vitest." + name + ".config.ts?io-probe=" + Date.now());
-          selections.push(config.test.include);
+        const selections = {};
+        for (const mode of ["cli", "env"]) {
+          if (mode === "cli") {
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3ab184e8cce536b79375956691d2406805163c19](https://github.com/openclaw/openclaw/pull/142912/commits/3ab184e8cce536b79375956691d2406805163c19)
- Landed commit: [e057f059f9cf3f57f3d2770bcc2577f0eca51a66](https://github.com/openclaw/openclaw/commit/e057f059f9cf3f57f3d2770bcc2577f0eca51a66)

PR #142915, opened 2026-09-09
fix(agents): retain Traditional Chinese policy digest markers

### LINKED ISSUE #142848: [Bug]: AGENTS.md policy digest silently drops non-Latin policy lines — \b cannot match CJK
**Bug type:** Bug — agents / bootstrap
**Beta release blocker:** No
**OpenClaw version:** 2026.9.1 / 2026.9.2; root cause still present on `main` @ `b72b30249f`
**Operating system:** macOS 15 (gateway)
**Model / provider chain:** not model-dependent — this is in prompt assembly

## Summary

When an `AGENTS.md` is larger than `bootstrapMaxChars`, `trimAgentsBootstrapContent()` keeps a head slice (45%), a **policy digest** (35%) and a tail slice (15%). The digest exists to rescue the mandatory rules sitting in the middle of the file, which plain truncation would throw away.

Both of the digest's matchers are English-only **and** anchored with `\b`. CJK characters are not `\w`, so `\b` can never form a boundary next to them. A rule written in Chinese can therefore never be ranked high priority, and if it is not also a bullet or a heading it never enters the candidate pool at all.

Nothing is logged. The rule is in the file, the operator believes it is in force, and the model never sees it.

## What triggers it

1. `AGENTS.md` is larger than `bootstrapMaxChars`. The default is **20,000** (`src/agents/embedded-agent-helpers.ts:89`), so this is easy to hit — 44 of our 53 agent files exceed it.
2. The mandatory rules are written in a non-Latin language.
3. The rules sit outside the head 45% / tail 15% slices, i.e. in the middle of the file — exactly what the digest is supposed to protect.

The more explicitly a rule is marked mandatory in the local language, the more likely it is to be dropped, because the digest cannot see it.

### 1. The matchers, run verbatim from `main`

| line shape | `isPolicyDigestCandidate` | `highPriority` |
|---|---|---|
| Chinese prose line | **false** | **false** |
| Chinese bullet line | true (via the structural check) | **false** |
| Chinese heading line | true (via the structural check) | **false** |
| English bullet line | true | true |

So Chinese bullets and headings do get into the pool, but can never be prioritized; Chinese prose lines never get in at all.

### 2. `\b` is the cause, not the vocabulary list

```js
const withB = /\b(?:禁止|嚴禁|不得)\b/u;
const noB   = /(?:禁止|嚴禁|不得)/u;
withB.test("🔴 禁止使用共用登入帳號");  // false
noB.test("🔴 禁止使用共用登入帳號");    // true
```

Adding CJK terms to the existing alternation would change nothing; the `\b` has to go for that group.

### 3. Corpus scale

53 `AGENTS.md` files, 21,521 non-empty lines, measured 2026-09-09:

| | lines | share |
|---|---|---|
| matched by upstream's `highPriorityPattern` | 293 | 1.4% |
| containing a strict Chinese prohibition (`🔴 禁止 嚴禁 不得 絕不 絕對不 切勿`) | 2,185 | 10.2% |
| matched by both | 5 | — |
| **strict prohibitions invisible to `highPriorityPattern`** | **2,180** | **10.1%** |

## Why it happens

- `src/agents/embedded-agent-helpers/bootstrap.ts:161-163` — the keyword branch of `isPolicyDigestCandidate()`.
- `src/agents/embedded-agent-helpers/bootstrap.ts:221-222` — `highPriorityPattern`.

Both are `/\b(?:…English words…)\b/iu`. `\b` requires a `\w`↔non-`\w` transition, and CJK codepoints are not `\w` under `u` semantics.

## Suggested fix

Join a CJK alternation **without** `\b` to the existing Latin alternation. No behavior change for existing English content.

```ts
// CJK mandatory-language groups carry no \b: CJK codepoints are not \w, so a
// word boundary can never match beside them. The Latin group keeps \b.
const POLICY_DIGEST_CANDIDATE_PATTERN =
  /\b(?:AGENTS\.md|scoped|required|must|never|do not|before subtree|read scoped|owner|security|secret|credential|test|validation|command|commit|push|github|pr)\b|(?:🔴|禁止|嚴禁|不得|絕不|絕對不|切勿|必須|務必|一律|紅線)/iu;

const POLICY_DIGEST_HIGH_PRIORITY_PATTERN =
  /\b(?:AGENTS\.md|scoped|required|must|never|do not|before subtree|read scoped|security|secret|credential)\b|(?:🔴|禁止|嚴禁|不得|絕不|絕對不|切勿)/iu;
```

Why the two groups differ in width, measured on the same corpus:

| CJK term set | lines | share | ≈ per file |
|---|---|---|---|
| strict only — `🔴 禁止 嚴禁 不得 絕不 絕對不 切勿` → **high priority** | 2,185 | 10.2% | ~41 |
| strict + `必須 務必 一律 紅線` → **candidate** | 3,684 | 17.1% | ~70 |
| strict + all soft terms (`不可 只能 唯一 強制 硬規則`) — rejected | 4,591 | 21.3% | ~87 |

High priority stays strict so ranking keeps meaning; the candidate pool is wider because its job is admission, not ranking.

## Result after the fix

Method: the selection logic of `trimAgentsBootstrapContent()` from `main` was ported verbatim (head/digest/tail ratios, `normalizePolicyDigestLine`, `buildAgentsPolicyDigest`'s two-pass high-priority-first selection, and the overflow shrink loop) and run over the real corpus twice — once with upstream's patterns, once with the patterns above. The metric is **how many of a file's strict-mandatory lines actually appear in the final rendered bootstrap text**, which is what the model receives.

At the upstream default `bootstrapMaxChars = 20000` (44 of 53 files over the limit):

| | mandatory lines visible | share |
|---|---|---|
| upstream today | 812 | 46.3% |
| with the fix | 1,661 | **94.6%** |

Worst-affected files at that setting:

| agent file | size (chars) | mandatory lines | visible today | visible after | gained |
|---|---|---|---|---|---|
| A | 54,595 | 60 | 14 | 54 | +40 |
| B | 48,240 | 52 | 15 | 51 | +36 |
| C | 47,376 | 63 | 18 | 52 | +34 |
| D | 38,453 | 54 | 17 | 51 | +34 |

At `bootstrapMaxChars = 30000` (23 of 53 files over the limit): 555 → 1,091 visible, i.e. 50.1% → 98.5%.

Two honest caveats:

- On **our own current configuration** (`bootstrapMaxChars = 60000`, after slimming our instruction files) no file exceeds the limit today, so the digest path is not firing for us right now. The measurements above are at the upstream default and at our previous setting. The bug is real for anyone on the default; for us it is currently latent.
- The term lists are the ones we validated against a Traditional-Chinese corpus. A maintainer with a wider sample may want to adjust them, and Japanese/Korean equivalents are not covered by this draft.

If you would rather not hard-code vocabulary in core, an alternative shape is a configurable `policyDigestKeywords` with Latin and non-Latin groups kept separate so the `\b` distinction survives. Happy to build whichever you prefer.

## Impact

Every non-English deployment whose `AGENTS.md` exceeds the bootstrap budget — which, at a 20,000-character default, is most of them. The failure is silent, and it is inverted from what an operator expects: the digest exists precisely to protect mandatory lines, and for these users it protects the English ones and drops theirs.

sunlit-deng:
Fixes #142848

## What Problem This Solves

Oversized `AGENTS.md` files can lose standalone Traditional Chinese mandatory lines from the middle of the context sent to the model. Chinese bullets enter the existing policy digest but lack mandatory-marker priority.

## Why This Change Was Made

Extend the existing candidate and priority patterns with the demonstrated Traditional Chinese markers, without Latin word boundaries around those alternatives. Keep the high-priority vocabulary narrower than the candidate vocabulary. English matching, source ordering, framing, and configured context limits retain their existing rules.

The priority regression uses a Chinese bullet without a red-circle marker, so it detects lost Chinese priority independently of structural admission.

## User Impact

The bounded policy digest can retain the demonstrated mandatory markers in Traditional Chinese workspace instructions. This does not promise complete Chinese, Japanese, Korean, or universal non-Latin coverage, or retention of every line under an exhausted budget.

## Evidence

The public packaged `agent --local` command read an oversized synthetic workspace and sent its complete request to a local capture endpoint. Both builds used the same authored files, explicit bootstrap budgets, ordinary turn, and fixed endpoint response. The proof checks submitted context, not model compliance.

| Scenario | Baseline `9d89877dd6d9` | Candidate |
| --- | --- | --- |
| Standalone Traditional Chinese prohibition | Omitted from submitted context | Retained once |
| Ordinary Chinese middle prose | Omitted | Omitted |
| English prohibition | Retained | Retained |

- Independent acceptance passed 14 paired scenarios (28 public turns). Each completed one full captured request without retries or fallback.
- A Chinese structural bullet without a red circle gained priority over earlier ordinary bullets. The tested obligation/prohibition pair preserved the narrower high-priority vocabulary.
- English rules, source order, quoted framing, backtick/tilde frame resets, omission notices, repeated frames, UTF-16 boundaries, total budgets, and short-file content passed. The tight shared-budget case stayed at 649 of 650 UTF-16 units on both builds.
- 113 focused cases passed across the context builder, budget, and bootstrap loader suites. The 37-case builder suite was rerun after the test correction. The corrected priority case fails on the baseline because the already-admitted Chinese bullet loses priority. Formatting, syntax lint, and the native commit hook passed.
- Public proof used baseline `9d89877dd6d9` and candidate `540241a808e7`. Final head `0fe758a9ca26` only strengthens the test fixture; production code is unchanged. [Hosted checks](https://github.com/openclaw/openclaw/pull/142915/checks) validate that final head.

Limits: the capture endpoint returns a fixed response, so this proves submitted context and transport rather than model compliance. Fences preserve the existing frame-reset behavior; they do not universally exclude interior policy-like lines. Short files retain existing `trimEnd` behavior. Bootstrap limits apply to injected file content, not the entire request with tool definitions and other prompt text.

### DIFF
--- a/src/agents/embedded-agent-helpers.buildbootstrapcontextfiles.test.ts
+++ b/src/agents/embedded-agent-helpers.buildbootstrapcontextfiles.test.ts
@@ -311,10 +311,10 @@ describe("buildBootstrapContextFiles", () => {
     expect(result?.content).toContain("[...1 more policy lines omitted...]");
     expect(result?.content.length).toBeLessThanOrEqual(600);
   });
-  it("prioritizes non-Latin mandatory policy markers", () => {
+  it("prioritizes Traditional Chinese mandatory bullets", () => {
     const earlyShort = "- Early normal ".padEnd(20, "a");
     const earlyLong = "- Normal payload ".padEnd(70, "b");
-    const urgent = "🔴 禁止共享账号 ".padEnd(140, "字");
+    const urgent = "- 嚴禁共用登入帳號 ".padEnd(140, "字");
     const lateShort = "- Late normal ".padEnd(20, "d");
     const [result] = buildBootstrapContextFiles(
       [makeMiddleBootstrapFile([earlyShort, "", earlyLong, "", urgent, "", lateShort])],
--- a/src/agents/embedded-agent-helpers.buildbootstrapcontextfiles.test.ts
+++ b/src/agents/embedded-agent-helpers.buildbootstrapcontextfiles.test.ts
@@ -173,6 +173,28 @@ describe("buildBootstrapContextFiles", () => {
     expect(result?.content).toContain(requiredScopedInstruction);
     expect(result?.content).toContain("[...truncated, read AGENTS.md for full content...]");
   });
+  it("keeps non-Latin mandatory policy lines from oversized AGENTS.md middle content", () => {
+    const mandatory = "禁止在此子树使用共享账号";
+    const ordinary = "请保持本段内容简洁";
+    const content = [
+      "# Root policy",
+      "A".repeat(900),
+      "",
+      mandatory,
+      "",
+      ordinary,
+      "B".repeat(700),
+      "tail marker",
+    ].join("\n");
+    const [result] = buildBootstrapContextFiles([makeFile({ content })], {
+      maxChars: 600,
+    });
+
+    expect(result?.content).toContain("[Policy digest from AGENTS.md]");
+    expect(result?.content).toContain(mandatory);
+    expect(result?.content).not.toContain(ordinary);
+    expect(result?.content.length).toBeLessThanOrEqual(600);
+  });
   it("keeps the quoted heartbeat example with its framing", () => {
     const frame = "Example heartbeat prompt:";
     const [result] = buildBootstrapContextFiles(
@@ -278,6 +300,21 @@ describe("buildBootstrapContextFiles", () => {
     const earlyShort = "- Early normal ".padEnd(20, "a");
     const earlyLong = "- Normal payload ".padEnd(70, "b");
     const urgent = "Never ".padEnd(140, "c");
+    const lateShort = "- Late normal ".padEnd(20, "d");
+    const [result] = buildBootstrapContextFiles(
+      [makeMiddleBootstrapFile([earlyShort, "", earlyLong, "", urgent, "", lateShort])],
+      { maxChars: 600 },
+    );
+
+    expect(result?.content).toContain([earlyShort, urgent, lateShort].join("\n"));
+    expect(result?.content).not.toContain(earlyLong);
+    expect(result?.content).toContain("[...1 more policy lines omitted...]");
+    expect(result?.content.length).toBeLessThanOrEqual(600);
+  });
+  it("prioritizes non-Latin mandatory policy markers", () => {
+    const earlyShort = "- Early normal ".padEnd(20, "a");
+    const earlyLong = "- Normal payload ".padEnd(70, "b");
+    const urgent = "🔴 禁止共享账号 ".padEnd(140, "字");
     const lateShort = "- Late normal ".padEnd(20, "d");
     const [result] = buildBootstrapContextFiles(
       [makeMiddleBootstrapFile([earlyShort, "", earlyLong, "", urgent, "", lateShort])],
--- a/src/agents/embedded-agent-helpers/bootstrap.ts
+++ b/src/agents/embedded-agent-helpers/bootstrap.ts
@@ -105,6 +105,10 @@ const AGENTS_POLICY_DIGEST_RATIO = 0.35;
 const AGENTS_POLICY_HEAD_RATIO = 0.45;
 const AGENTS_POLICY_TAIL_RATIO = 0.15;
 const AGENTS_POLICY_DIGEST_MAX_LINE_CHARS = 240;
+const AGENTS_POLICY_DIGEST_CANDIDATE_PATTERN =
+  /\b(?:AGENTS\.md|scoped|required|must|never|do not|before subtree|read scoped|owner|security|secret|credential|test|validation|command|commit|push|github|pr)\b|(?:🔴|禁止|嚴禁|不得|絕不|絕對不|切勿|必須|務必|一律|紅線)/iu;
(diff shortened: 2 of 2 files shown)


obviyus:
Thanks @sunlit-deng!

**Proof:** The public packaged agent command submitted the expected Traditional Chinese prohibition to a local capture endpoint. Independent acceptance checked the actual request text across 14 paired scenarios, with 28 completed public turns.

- **Bug:** Oversized workspace instructions could lose a standalone Chinese prohibition, while Chinese bullets lacked mandatory-marker priority.
- **Fix:** Extend the existing digest predicates with the demonstrated Traditional Chinese markers, preserving the existing context limits and a narrower high-priority vocabulary.

| Before | After |
| --- | --- |
| Standalone Chinese prohibition omitted | Retained once in submitted context |
| Chinese prohibition bullet displaced by ordinary bullets | Mandatory priority retained in source order |
| Ordinary Chinese prose omitted under the budget | Still omitted |

English matching, quoted framing, fence resets, omission notices, UTF-16 boundaries, shared budgets, and short-file content passed. All 113 focused cases have passing proof; the 37-case builder suite was refreshed after strengthening the priority regression. [Final-head hosted checks](https://github.com/openclaw/openclaw/actions/runs/34401953375) and the fresh local review passed.

Scope remains the demonstrated markers in the existing bounded digest. This proves submitted context, not model compliance or universal language coverage. Short content follows existing `trimEnd` behavior, and fences retain their existing frame-reset semantics.

Issue #142783, opened 2026-09-09
[Bug]: Plugin typed hooks (before_prompt_build / agent_end) stop dispatching after multi-agent migration — memory plugins dead


140358:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

Memory-plugin typed hooks (before_prompt_build / agent_end) are never dispatched after migrating to explicit multi-agent ownership: the plugin (memos-local-plugin) loads and serves its viewer and retrieval, but no turn hook ever fires, so trace capture is dead (zero invocations across versions 2026.8.1-beta.2, 2026.9.1-beta.1, and 2026.9.3).

### Steps to reproduce

1. Baseline (single-agent implicit ownership): every agent turn logs memos.onTurnStart and memos.agent_end.received in the gateway log (1964 onTurnStart occurrences before the breakpoint).
2. Migrate config to explicit multi-agent ownership: agents.ownership=explicit, 8 agents, isolated cron sessions, agents.defaults.heartbeat.agentId set; restart the gateway.
3. Run any agent turn on the migrated config (heartbeat, cron, or channel session).
4. Observe: the plugin boots cleanly (memos-local: plugin ready, viewer live on :18799) but no memos.onTurnStart / memos.agent_end line is ever emitted and no trace row is written to the memos SQLite DB.
Reproduced identically on 2026.8.1-beta.2, 2026.9.1-beta.1 and 2026.9.3; on node v22.22.3 and v26.7.0; under launchd and under an external supervisor.

### Expected behavior

before_prompt_build, agent_end, session_start and the other typed hooks the plugin registers via api.on(...) fire on every agent turn, matching pre-migration behavior where the same plugin logged onTurnStart hits=N and persisted each turn as a trace.

### Actual behavior

After the multi-agent migration no typed hook of the plugin is invoked at all: zero memos.onTurnStart / memos.agent_end lines in gateway logs across every tested version, node and supervisor.
The last persisted trace is 2026-08-20T15:06 (UTC+8).
Non-hook surfaces of the plugin still work (viewer HTTP server on :18799, memory-capability prompt section, memos_search retrieval), showing register() and the memory-slot path complete while per-request hook dispatch never reaches the plugin.

### OpenClaw version

2026.9.3 (also reproduced on 2026.8.1-beta.2 and 2026.9.1-beta.1)

### Operating system

macOS 26.6.2 (arm64)

### Install method

[...]

PR #142941, opened 2026-09-09
fix(plugins): keep an intentional empty generation selection exclusive

xydigitLybnnnn:
Refs #142783

## Summary

- `resolveHookRegistry()` now falls back to the process-root hook view only for a run that carries **no prepared plugin registry at all**. A run that prepared a registry — including an explicitly empty selection such as `plugins.enabled=false` → `onlyPluginIds: []` — keeps owning hook and policy dispatch exclusively, so root callbacks that the generation excluded stay excluded before they can read the conversation or modify the prompt.
- The earlier revision used "is the registry empty?" as the criterion. That cannot distinguish a deliberate empty selection from a missing inheritance: it let the process-root fail-closed tool policy back into an isolated generation, which is the `merge-risk: 🚨 security-boundary` item from the review.
- `withPluginRuntimeGenerationScope` still installs the placeholder empty registry when `generation.pluginRegistry` is absent, so provider and metadata isolation are unchanged; only the new `selected` flag records whether the generation actually selected plugin content. `provider/metadata` assertions in `src/plugins/current-plugin-metadata-snapshot.test.ts` are untouched apart from the hook-policy expectation this behaviour change implies.
- Regression coverage in `src/plugins/issue-142783-hook-dispatch-repro.test.ts` now has three cases: a registry-less generation keeps dispatching process-root typed hooks, an explicitly empty selection stays exclusive, and a content-bearing generation stays exclusive.

## What Problem This Solves

It removes the isolation regression the previous revision of this PR introduced (an empty generation inheriting process-root hooks and tool policy) while keeping registry-less runs on the process-root view they had before `02272c345f3`.

## What this does **not** do

It does **not** demonstrate a fix for the deployment reported in #142783. See below.

### Real Gateway trace (before/after)

Harness: `.artifacts/evidence-142783/hook-trace.mts` (kept at
`logs/real_behavior/issue-142783-evidence/harness/`) — a real `openclaw gateway`
process from this checkout, driven over the real WebSocket API (`chat.send`,
`cron.add` + `cron.run`), with a real workspace plugin registering the same typed
per-turn hooks the reported plugin registers (`before_prompt_build` / `agent_end`
/ `session_start`). The plugin appends to a marker file when a hook runs and
returns `prependContext`, and a stub OpenAI-Responses provider records every
request body, so prompt mutation is visible in what the model received.

| Scenario | Hook invocations per turn | Injected prompt reached the model |
|---|---|---|
| single agent, implicit ownership (healthy baseline) | `session_start`, `before_prompt_build`, `agent_end` | yes |
| `agents.ownership=explicit`, two agents — **this branch** | same, both agents | yes |
| `agents.ownership=explicit`, two agents — **base (pre-PR)** | same, both agents | yes |
| isolated cron agent turn (`cron.add` + `cron.run`) — base | `before_prompt_build`, `agent_end` | yes |

With a temporary trace at the two boundaries (`withPluginRuntimeGenerationScope`
install and hook-registry resolution):

| Scenario | generation scopes installed | registry-less installs | dispatches with a generation registry | dispatches with no generation scope |
|---|---|---|---|---|
| explicit multi-agent | 76 | **0** | 40 | 18 |
| + `/compact`, `/recap`, `/btw` | 88 | **0** | 65 | 34 |
| + isolated cron agent turn | 82 | **0** | 57 | 21 |

Every installed generation carried a prepared registry, and every dispatched typed
hook resolved to a registry containing the probe plugin. **No hook-dispatching
turn ever ran inside a registry-less generation**, and the reported multi-agent
configuration does not distinguish the base revision from this branch.

`prepareWorkspacePluginRegistries` omits `pluginRegistry` only when the run is
`readOnly` **and** does not `loadRuntimePlugins` **and** carries no
`runtimePluginSelections`. All callers of that combination are read-only
catalog / auth-label / registry-load / synthetic-auth paths; the only producer of
an isolated empty runtime registry (`preparedModelRuntimeMode: "isolated-read-only"`)
is the setup-inference probe turn and the `openclaw models` auth probe — not an
operator agent turn. This agrees with the call-site trace in the review thread
(27/34 scopes carry a registry, 6 omit it in read-only paths, 1 is explicitly empty).

Consequence: the "registry-less generation silences process-root typed hooks"
mechanism cannot be reached through the product entry points exercised here, so
before/after evidence for restored hook capture cannot be produced — base already
dispatches. The isolation boundary itself can only be asserted at the scope
boundary (`issue-142783-hook-dispatch-repro.test.ts`, case 2), because no product
configuration selects "no plugins" while process-root registrations stay live.

Scope limits: two agents rather than eight, and the reporter's full config
(`agents.defaults.heartbeat.agentId`, isolated cron sessions per agent) was
approximated by a real isolated cron agent turn. `logs/real_behavior/issue-142783-evidence/harness/EVIDENCE.md`
records the method and raw results.

### Local validation

- `node scripts/run-vitest.mjs run <all files referencing generation scopes>` — 57 files green (the only two failures are `hosted-setup.runtime.test.ts` prompt-string assertions caused by this machine's `LANG=zh_CN.UTF-8`; that file passes 16/16 under `LANG=C`).
- `pnpm check:assertion-safety`, `pnpm check:max-lines-ratchet`, `pnpm tsgo:core` — clean.
- `pnpm deadcode:dependencies` reports only generated `apps/linux/src-tauri/target/**` build artifacts in this working tree, unrelated to the change.


aeoess:
I traced all 34 `withPluginRuntimeGenerationScope` call sites in `src` at `ad8a648`, following the actual `generation` value back to where it is constructed.

The result was:

- 27 carry a plugin registry
- 6 omit `pluginRegistry`, so the empty default is installed
- 1 is explicitly empty

The six omitted cases I found are read-only catalog, auth-label, registry-load, or synthetic-auth paths. In the traced code they do not enclose `before_tool_call`, `before_prompt_build`, `before_compaction`, or another agent-lifecycle typed hook. The turn, compaction, cron, reply, and tool-inventory paths I traced carry a runtime registry.

I also ran the isolation counter-test against this PR. With current main, a root `before_tool_call` hook stays dark inside an explicitly empty generation. With `4ee2a74b` applied, that same root hook runs. So the fallback cannot distinguish a registry that is empty because none was supplied from one that is deliberately empty, which reproduces the isolation concern in the review.

This does not establish whether the PR fixes the deployment reported in #142783. It only narrows the empty-registry hypothesis.

One path I have not finished tracing is selection inside a non-empty generation. An agent turn can carry a registry while a root-loaded plugin is absent from that registry's `runtimePluginSelections`. If that happens, the root hook would still be hidden without an empty registry being involved. I'm tracing that separately rather than treating the empty-default mechanism as the deployment explanation.

xydigitLybnnnn:
Revision 5 pushed (`bfe97154f3f`) and the body rewritten.

**What changed since the last review**

The criterion no longer asks "is the generation registry empty?". `withPluginRuntimeGenerationScope` records whether the generation prepared a registry at all (`generation-state.ts`), and `resolveHookRegistry()` falls back to the process-root view only when there is no prepared registry. An explicitly empty selection — the case the review called out — therefore stays exclusive again, so root fail-closed tool policy and excluded root callbacks do not re-enter that generation before it can touch the conversation or the prompt. Provider/metadata isolation is unchanged (the placeholder empty registry is still installed).

**Real behaviour proof — and it does not support the recovery claim**

I built a real-Gateway harness (`.artifacts/evidence-142783/hook-trace.mts`, kept under `logs/real_behavior/issue-142783-evidence/harness/`): a real `openclaw gateway` process from this checkout, driven over the real WebSocket API (`chat.send`, `cron.add` + `cron.run`), a real workspace plugin registering `before_prompt_build` / `agent_end` / `session_start` and appending to a marker file, plus a stub provider that records every request body.

| Scenario | base (pre-PR) | this branch |
|---|---|---|
| single agent, implicit ownership | 3 hooks/turn, prompt mutation reaches the model | same |
| `agents.ownership=explicit`, two agents | 3 hooks/turn for both agents | same |
| isolated cron agent turn | `before_prompt_build`, `agent_end` | same |

With a temporary trace at the scope-install and hook-resolution boundaries: **0 registry-less installs out of 246 scopes** across those runs (76 / 88 / 82), and every dispatched typed hook resolved to a registry containing the plugin. `prepareWorkspacePluginRegistries` only omits `pluginRegistry` for `readOnly && !loadRuntimePlugins && !runtimePluginSelections` runs, i.e. read-only catalog/auth/registry-load/synthetic-auth paths — which matches the call-site trace posted earlier in this thread. The only isolated-empty runtime registry comes from the setup-inference probe and `openclaw models` probe, not an operator turn.

So I cannot produce "affected Gateway turn" evidence for restored capture: on the base revision those turns already dispatch. The isolation boundary is asserted at the scope boundary in `issue-142783-hook-dispatch-repro.test.ts` (case 2) instead, because no product configuration selects "no plugins" while root registrations remain live. The PR body now states this explicitly and uses `Refs #142783` rather than `Closes`.

@clawsweeper re-review

PR #142943, opened 2026-09-09
feat(models): add credential-only provider sign-in

obviyus:
## Problem and change

Settings → Models has no credential-only provider sign-in action, and a second authenticated connection can inspect an existing setup wizard. Add manifest-owned sign-in choices for the selected agent and run only the chosen plugin/provider/method through the existing login writer. Saving credentials preserves the default model and model restrictions. Explicit setup activation keeps its existing entry point and result handling.

Wizard prompts, answers and cancellation belong to the starting connection. Disconnect ends input before persistence; an effect already being saved can finish. Models waits for the prior wizard's admission to release and ignores completion from an old agent or connection.

## Verification status

This pull request is an **unverified CI input**. The pinned baseline reproduced the socket ownership defect through the ordinary compiled Gateway and default setup wizard. Both clients and the Gateway stopped, and no prompt answer or credential write occurred.

Candidate tests, the matching compiled Gateway/UI proof, independent acceptance and final code review remain required before landing. The normal required CI includes the new real-Gateway Models sign-in case, with cancellation, peer refusal, credential readback and unchanged-default/restriction controls. Automatic platform jobs remain part of that normal CI.

The change preserves the existing shared refresh interface and personal-account boundary. It does not implement a new credential storage or publication protocol. Mechanical generated-client alignment is included with the core change.


obviyus:
Superseded by #144329, rebuilt from current main for credential-only provider sign-in and the shared /login menu.

PR #142818, opened 2026-09-09
improve(plugins): avoid rebuilding provider ownership metadata

steipete:
Related: #127178

## What Problem This Solves

Provider ownership lookup could rebuild plugin manifests even when the registry loader had already prepared compatible metadata. This repeated work affects the cold lookup used to identify a provider or CLI backend's owning plugin.

## Why This Change Was Made

The ownership lookup now keeps the loader's paired manifest metadata and passes it through the existing installed-plugin adapter. Explicit and current metadata keep their precedence; the adapter still owns installed filtering, disabled-plugin ownership, normalization, and reconstruction when prepared metadata is absent.

## User Impact

Provider and CLI backend selection keep their existing results while avoiding redundant manifest reconstruction. No configuration or plugin API changes are required.

## Evidence

The new regression fails on unchanged production because prepared metadata triggers two unnecessary reconstructions; the absent-metadata control and existing ownership cases pass. After the repair, 137 tests pass across the provider, snapshot, installed-adapter, and current-runtime-metadata suites. All 29 canonical changed-file checks pass, including core and test types, lint, and architecture guards.

The two-file patch and its inspected dependencies carry unchanged to main `3f0f98ef7ef`. Production shrinks by two lines. These are synthetic metadata tests through the real adapter; no live-provider or elapsed-time claim is made. Independent Codex review found no actionable issues.

PR #142975, opened 2026-09-09
improve(plugins): reuse metadata during contribution discovery

steipete:
## What Problem This Solves

Plugin contribution discovery could recheck manifest files even when a matching metadata snapshot was already available. This adds unnecessary work to callers such as channel setup, which asks for installed channel IDs without supplying its own lookup table.

## Why This Change Was Made

Contribution lookup now carries the installed index and its prepared manifest registry together through the existing adapter. One shared selection path replaces a private wrapper and duplicate index loads, removing 16 production lines. Current enablement policy, disabled-owner inspection, raw declarations, indexed aliases, and explicit-index fallback retain their existing behavior.

## User Impact

Metadata-backed contribution discovery reuses the published inventory without checking its manifest files again. Channel and provider selection remain unchanged; this change adds no configuration or cache. No end-to-end timing improvement is claimed.

## Evidence

The regression failed on the previous code because contribution lookups performed two filesystem checks against already-published manifests. The identical test passes after the repair and checks both the returned IDs/owners and the absence of filesystem access. All 93 cases across five owner and sibling suites passed, including explicit-index reconstruction, raw/indexed matching, and enablement-policy controls.

Independent review found no actionable P0–P2 issue.

All 29 selected changed-file checks passed.

Related: #141193, #142818.

### DIFF
--- a/src/plugins/plugin-registry-contributions.current-snapshot.test.ts
+++ b/src/plugins/plugin-registry-contributions.current-snapshot.test.ts
@@ -11,8 +11,10 @@ import { clearPluginMetadataLifecycleCaches } from "./plugin-metadata-lifecycle.
 import type { PluginMetadataSnapshot } from "./plugin-metadata-snapshot.types.js";
 import {
   loadPluginManifestRegistryForPluginRegistry,
+  listPluginContributionIds,
   resolveManifestContractOwnerPluginId,
   resolveManifestContractPluginIds,
+  resolvePluginContributionOwners,
 } from "./plugin-registry-contributions.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e74e946f237634fafa11224226d148c55112f6fc](https://github.com/openclaw/openclaw/pull/142975/commits/e74e946f237634fafa11224226d148c55112f6fc)
- Landed commit: [2cb7337e20771dbbb793cbb235cdde577010ff68](https://github.com/openclaw/openclaw/commit/2cb7337e20771dbbb793cbb235cdde577010ff68)

PR #142978, opened 2026-09-09
fix(tests): close reloaded token fixture database caches

steipete:
Related: #139428

## What Problem This Solves

Gateway token tests reload modules to exercise restart boundaries, but teardown closes only the database cache imported before those resets. Reloaded cache generations can retain live SQLite handles when the temporary fixture directories are removed.

## Why This Change Was Made

Track the state-database closer from every reloaded token-module generation and invoke all of them before module reset and directory removal. Retain the original static closer because that cache is also used. Multi-import cases close every acquired generation, not just the latest one.

## User Impact

No production behavior change. This repairs resource ownership in two test fixtures and adds 16 net test lines. SQLite storage semantics, maintenance timers and the WAL safety tripwire are unchanged.

## Evidence

- Both complete suites passed normally before and after: 20 cases with identical per-file order.
- Original-generation observations found 21 live reloaded handles after the old closer across 19 cases, including two cases with multiple generations.
- Candidate diagnostic copies passed all 20 cases and verified the same observed handles were closed before unlink and before emergency cleanup. Observers did not proactively open databases; all temporary copies and roots were removed afterward.

Local proof used Node 24.20 on macOS. This confirms the fixture cleanup bug; it does not identify the exact database behind the previously captured Linux WAL tripwire event. No runtime improvement is claimed.

### DIFF
--- a/src/gateway/agent-runtime-identity-token.test.ts
+++ b/src/gateway/agent-runtime-identity-token.test.ts
@@ -22,6 +22,7 @@ import {
 const envSnapshot = captureEnv(["HOME", "OPENCLAW_HOME", "OPENCLAW_STATE_DIR"]);
 
 const tempHomes: string[] = [];
+const reloadedStateDatabaseClosers = new Set<() => void>();
 
 function operationalRun(runId = "run-1") {
   const operationalRunInstance = { instanceId: `instance-${runId}`, runId } as const;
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9fe775c1eff45351044b90b69a155617e973df77](https://github.com/openclaw/openclaw/pull/142978/commits/9fe775c1eff45351044b90b69a155617e973df77)
- Landed commit: [36233dacceb7e5764fc333d7c70a75b5ed598fb5](https://github.com/openclaw/openclaw/commit/36233dacceb7e5764fc333d7c70a75b5ed598fb5)

PR #142991, opened 2026-09-09
fix(agents): keep accepted image jobs alive through provider cleanup

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where an accepted `image_generate` job could lose its plugin resources after returning `started`, while generation, image persistence, or rollback was still running. Preparing tools and checking remote image references could also outlive the original provider registration.

## Why This Change Was Made

Carry the original inspection's resource ownership through prepared facts, preflight, and the actual accepted job. A retired source cannot admit a new task, while an already accepted task retains its captured providers until work and cleanup finish. Shared capability acquisition now exposes the same existing ownership operation for this longer lifetime.

Existing execution and image-geometry helpers move into an execution module to keep the tool readable; their behavior is unchanged. The oversized-file exception is removed. Raw prepared registries retain their host-owned lifetime; this does not enable general prepared-cache disposal.

## User Impact

Background image generation keeps working after its originating preparation is released. A source retired during preflight produces a retryable setup failure before task admission. `started` remains an immediate acceptance result, and completed media remains available after provider cleanup.

## Evidence

- Native before/after regressions cover queued generation, saving, rollback, retirement during a held HTTP preflight, and retirement during actual preparation.
- Focused resource, runtime, policy, background-task and Cron suites passed. The full changed-file gate and independent Codex review through P2 passed; ordinary build completed in 219 seconds.
- The built public agent-harness SDK accepted a synthetic SQLite-backed image job in 16.6 ms. After source A retired and B was published, A persisted a real PNG and closed exactly once; B was never invoked. A separate held HTTP preflight rejected admission after retirement. Native SQLite reopening succeeded.
- The probe used source-produced prepared facts with the built SDK, not a fully packaged CLI. With no delivery runtime attached, the durable task reached its existing blocked-delivery state with media retained; real channel delivery was not tested.
- Rebased production files are byte-identical to the reviewed and built candidate. Focused composition verification covers the subsequently landed Video and inventory consumers.

### DIFF
--- a/config/max-lines-baseline.txt
+++ b/config/max-lines-baseline.txt
@@ -387,7 +387,6 @@ src/agents/tool-loop-detection.test.ts
 src/agents/tool-search.test.ts
 src/agents/tools/cron-tool.test.ts
 src/agents/tools/image-generate-tool.test.ts
-src/agents/tools/image-generate-tool.ts
 src/agents/tools/image-tool.test.ts
 src/agents/tools/image-tool.ts
 src/agents/tools/media-generate-background-shared.test.ts
--- a/docs/plugins/sdk-overview/host-hooks.md
+++ b/docs/plugins/sdk-overview/host-hooks.md
@@ -61,6 +61,16 @@ completed video assets and result metadata. Video assets may contain buffers or
 provider-hosted URLs. Downloads after the call returns belong to the caller;
 registration disposal must not invalidate those completed artifacts.
 
+For `image_generate` tools prepared from an owned inspection, resources remain
+held through preflight and, once accepted, through generation, image saving, and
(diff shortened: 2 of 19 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [bda45b23f48aeb20bb7ec12cde185b05e64890ce](https://github.com/openclaw/openclaw/pull/142991/commits/bda45b23f48aeb20bb7ec12cde185b05e64890ce)
- Landed commit: [b07d30ae5e0a1fea55d25f70a18ba07c844f414e](https://github.com/openclaw/openclaw/commit/b07d30ae5e0a1fea55d25f70a18ba07c844f414e)

PR #138157, opened 2026-09-04
docs: keep ClawHub guidance in its canonical source

steipete:
Related: https://github.com/openclaw/openclaw/pull/138121 — restores section links and unique maturity anchors, including the macOS caller that exposed this competing-source problem.

<details>
<summary>Additional instructions</summary>

Maintainer edits are enabled through this branch in the upstream repository.

</details>

## What Problem This Solves

Resolves a problem where contributors editing the checked-in ClawHub CLI and publishing pages would see that guidance silently replaced during docs publication. The local copies also advertised headings absent from the real published input: the macOS skills page linked to `/clawhub/cli#release-trust` until #138121 moved that caller to OpenClaw-owned content.

## Why This Change Was Made

Keep the existing ownership boundary: `openclaw/clawhub/docs` remains the only source for `/clawhub/**`. Remove the two competing OpenClaw pages after comparing every useful section against current ClawHub input and existing OpenClaw references. Preserve OpenClaw-specific install, update, verification, removal, refresh, and trust guidance in its existing CLI references; move the unique owner-and-role primer into the existing skill-publishing guide. Document the mirror boundary and require real mirror input for anchor checks.

The publishing implementation, public routes, navigation, and runtime behavior are unchanged. Keeping or regenerating a second local copy would preserve the source conflict; overriding upstream publication with OpenClaw pages would change ownership and discard the substantially richer canonical references.

## User Impact

Readers keep the canonical ClawHub CLI and publishing pages at the same URLs. OpenClaw-specific workflows remain available and linked, while contributors have an explicit source-of-truth rule instead of editing pages that never reach publication. Unsupported tarball-URL publishing, a scope-invalid package-transfer example, and an obsolete canonical skill URL are not copied forward.

| Material | Preserved location |
| --- | --- |
| OpenClaw skill install/update/verify, workspace/global targeting, trust, uninstall, and watcher refresh | `docs/cli/skills.md` |
| OpenClaw plugin installation/update and release trust | `docs/cli/plugins.md` |
| Personal/organization owners and publisher roles | `docs/tools/creating-skills.md#publishing-to-clawhub` |
| Standalone CLI commands, publishing, transfer, rename, and registry workflows | Canonical ClawHub CLI and publishing pages |

## Evidence

[...]

PR #143002, opened 2026-09-09
fix(docs): keep mirrored ClawHub routes resolvable in anchor audits

vincentkoc:
## Summary

`node scripts/docs-link-audit.mjs --anchors` reports 42 broken links on a clean `main` when the checkout has no ClawHub source beside it, while plain mode reports 0. All 39 `/clawhub/**` findings are false positives. **No docs content is changed by this PR** — the links are correct as authored.

## Diagnosis

`/clawhub/**` pages are authored in [openclaw/clawhub](https://github.com/openclaw/clawhub/tree/main/docs), not here. `scripts/docs-sync-publish.mjs` replaces the whole publish `docs/clawhub/` tree from that source, and `docs/AGENTS.md` (added in #138157, which removed the last local copies) says explicitly not to keep authored copies in this repo. `docs/docs.json` declares the twelve `/clawhub/**` navigation entries and three redirects into `/clawhub`; those are real published routes on `docs.openclaw.ai`.

`buildAuditIndex` already handles this: when no ClawHub checkout is resolvable it sets `allowExternalClawHubRoutes` and adds the navigation-declared routes to `routes`, so `resolveRoute` succeeds. But the link check discarded that allowance in anchors mode:

```ts
if (!page && (options.anchors || !resolved.ok) && !index.relAllFiles.has(...))
```

`options.anchors` short-circuits ahead of `resolved.ok`, so anchors mode demanded a parsed Markdown page and ignored the allowance that plain mode honours. Same for redirect resolution, which built its `pages` list from parsed pages only — hence the three `Redirect has no terminal Markdown page: /clawdhub -> /clawhub` errors.

The intent behind the short-circuit is sound: a route declaration cannot prove a *fragment*. But it can prove the *route*, and 39 of the 39 findings are fragment-less links.

## Fix

- Accept a declared mirrored route when the link carries no fragment.
- Count declared mirrored routes as terminal pages for redirect resolution.
- **In anchors mode only**, report a fragment into a mirrored route as `fragment unverified without the ClawHub source checkout; set OPENCLAW_DOCS_SYNC_CLAWHUB_REPO`, instead of the misleading `route/file not found`. Plain mode has never inspected fragments, so it keeps resolving the declared route and stays silent.
- Report the same for a **redirect destination** that lands on a mirrored route with a fragment, so an unreferenced redirect is not silently unchecked.
- Print `mirrored_clawhub_docs=yes|no` in anchors mode and warn on stderr when the source is absent, so reduced coverage is visible rather than silent.

When the ClawHub source *is* present, `mirroredRoutes` is empty and behaviour is byte-identical to before.

## Results

Run against the repo as authored (no `/clawhub/**` link here carries a fragment today):

| run | before | after |
| --- | --- | --- |
| `--anchors`, no ClawHub source | 42 | **3** |
| `--anchors`, with ClawHub source | 4 | 4 (unchanged) |
| plain, no ClawHub source (what CI runs) | 0 | 0 (unchanged) |
| plain, with ClawHub source | 0 | 0 (unchanged) |

Each of the three unchanged rows is **byte-identical** to `main`, diagnostics included.

The remaining 3 are the pre-existing `maturity/#windows-app-node` failures. The 4th in the with-source run is `clawhub/publishing.md:216 :: /clawhub/cli#package-publish-source` — a real broken anchor **in `openclaw/clawhub`**, not fixable from this repo. I have not filed it upstream; it is reported to the docs-audit orchestrator for routing to `openclaw/clawhub`.

### Fragment-bearing link (the plain-mode edge case)

Temporarily adding a page containing `[publishing](/clawhub/publishing#package-publish-source)`:

| run | `main` | first revision of this PR | current |
| --- | --- | --- | --- |
| plain, no ClawHub source | 0 (exit 0) | **1 (exit 1)** | 0 (exit 0), identical to `main` |
| plain, with ClawHub source | 0 | 0 | 0 |
| `--anchors`, no source | 43 | 4 | 4 |
| `--anchors`, with source | 5 | 5 | 5 |

### Fragment-bearing redirect destination

Temporarily rewriting the `/tools/clawhub` redirect to `/clawhub/publishing#package-publish-source`, `--anchors` with no ClawHub source:

```
main    docs.json:unknown :: /tools/clawhub :: Redirect has no terminal Markdown page: /tools/clawhub -> /clawhub/publishing
first   (no line mentioning /tools/clawhub — silently unchecked)
current docs.json:unknown :: /tools/clawhub :: fragment unverified without the ClawHub source checkout (terminal: /clawhub/publishing#package-publish-source); set OPENCLAW_DOCS_SYNC_CLAWHUB_REPO
```

Plain mode reports 0 in all three.

## Why CI did not catch this

It is not a CI regression — there was nothing to catch. `check:docs` runs plain `docs:check-links`, which correctly reports 0, and both lanes (`.github/workflows/ci.yml`, `.github/workflows/docs.yml`) check out `openclaw/clawhub` into `clawhub-source` and set `OPENCLAW_DOCS_SYNC_CLAWHUB_REPO`, so even anchors mode would be clean there.

Two related gaps worth recording, neither introduced here:

1. **`docs:check-links:anchors` is not run by any workflow.** It appears only in `package.json`, `CHANGELOG.md`, and three docs pages describing it as a manual step. Anchor regressions are ungated, which is why the three `#windows-app-node` failures have survived. Turning it on would need those fixed first, so I have not done it here.
2. **The audit silently auto-discovers `<repo>/../clawhub`** via `DEFAULT_CLAWHUB_REPO_CANDIDATES`. Coverage therefore depends on the directory *beside* the checkout, with no output saying which mode ran. That is what made this look like a regression: the same command reports 4 from a clone with a sibling `clawhub/` and 42 from a git worktree without one. The new `mirrored_clawhub_docs=` line makes that visible.

## Validation

`pnpm check:docs` — the aggregate the CI docs job runs — passes (exit 0), with and without `OPENCLAW_DOCS_SYNC_CLAWHUB_REPO` set.

Tests: `src/scripts/docs-link-audit.test.ts` under `test/vitest/vitest.tooling.config.ts`, 32 passed / 1 pre-existing failure (`preserves Mint component targets…`, `as-s-guide`; reproduced unchanged at `origin/main`, where the file is 24 passed / 1 failed).

Eight hermetic tests cover: mirrored route accepted; fragment reported unverified in anchors mode; **fragment left alone in plain mode, both through `auditDocsLinks` and through the real CLI**; **redirect destination fragment reported unverified in anchors mode and silent in plain mode**; undeclared `/clawhub/*` route still reported missing; allowance-off still reporting missing. Three of the four added in the current revision fail against the previous revision of this PR.

`pnpm check:test-types` reports the same 8 pre-existing `src/infra/*` errors with and without this change (identical error set); none are in the audited files.

### DIFF
--- a/scripts/docs-link-audit.mts
+++ b/scripts/docs-link-audit.mts
@@ -462,6 +462,14 @@ function parseAuditUrl(
 }
 
 /**
+ * Explains that a fragment into a ClawHub-mirrored route could not be checked
+ * because the upstream source is absent from this checkout.
+ */
+function mirroredFragmentReason(terminal: string, hash: string) {
+  return `fragment unverified without the ClawHub source checkout (terminal: ${terminal}${hash}); set ${CLAWHUB_REPO_ENV}`;
+}
+
+/**
  * Audits local docs links against route, file, and redirect indexes.
  */
 export function auditDocsLinks(
@@ -545,6 +553,22 @@ export function auditDocsLinks(
           link: record.source,
           reason: `redirect fragment not found: ${record.destination}`,
         });
+        continue;
+      }
+      // A redirect into a ClawHub-mirrored route now resolves off the navigation
+      // declaration, which proves the route but not the fragment. Say so instead
+      // of letting the destination fragment go silently unchecked.
+      if (!page && destination.hash) {
+        const terminal = normalizeRoute(destination.pathname);
+        if (index.mirroredRoutes.has(terminal)) {
+          unverifiedMirroredFragments++;
+          broken.push({
+            file: "docs.json",
+            line: 0,
+            link: record.source,
+            reason: mirroredFragmentReason(terminal, destination.hash),
+          });
+        }
       }
     }
   }
@@ -586,18 +610,22 @@ export function auditDocsLinks(
       // Markdown page exists locally. Its existence is proven by the declaration;
       // only its fragments need the real source checkout.
       const mirroredOnly = !page && index.mirroredRoutes.has(terminal);
-      if (mirroredOnly && !destination.hash) {
+      if (mirroredOnly) {
+        // Plain mode does not inspect fragments at all, so a declared mirrored
+        // route resolves there whether or not the link carries one. Only anchor
+        // mode reports the fragment it cannot verify without the source.
+        if (options.anchors && destination.hash) {
+          unverifiedMirroredFragments++;
+          broken.push({
+            file: rel,
+            line,
+            link: raw,
+            reason: mirroredFragmentReason(terminal, destination.hash),
+          });
+        }
         continue;
       }
-      if (mirroredOnly) {
-        unverifiedMirroredFragments++;
-        broken.push({
-          file: rel,
-          line,
-          link: raw,
-          reason: `fragment unverified without the ClawHub source checkout (terminal: ${terminal}${destination.hash}); set ${CLAWHUB_REPO_ENV}`,
-        });
-      } else if (
+      if (
         !page &&
         (options.anchors || !resolved.ok) &&
         !index.relAllFiles.has(url.pathname.slice(1))
--- a/scripts/docs-link-audit.mts
+++ b/scripts/docs-link-audit.mts
@@ -10,7 +10,11 @@ import { isRecord } from "@openclaw/normalization-core/record-coerce";
 import { err, ok, type Result } from "@openclaw/normalization-core/result";
 import MarkdownIt from "markdown-it";
 import type { Nodes } from "mdast";
-import { resolveClawHubRepoPath, syncClawHubDocsTree } from "./docs-sync-publish.mjs";
+import {
+  CLAWHUB_REPO_ENV,
+  resolveClawHubRepoPath,
+  syncClawHubDocsTree,
+} from "./docs-sync-publish.mjs";
 import { parseDocsDocument, resolveDocsFragment } from "./lib/docs-markdown.mjs";
 import {
   addRoute,
@@ -346,13 +350,27 @@ function buildAuditIndex(
     routes.add(normalizeRoute(permalink));
   }
 
+  // Without a ClawHub checkout the mirrored tree is absent, so its pages cannot be
+  // parsed. The navigation still declares the published routes, so keep them
+  // resolvable and remember which ones are only backed by that declaration.
+  const mirroredRoutes = new Set<string>();
   if (options.allowExternalClawHubRoutes === true) {
     for (const route of collectMirroredDocsRoutes(docsConfig.navigation)) {
       routes.add(route);
+      mirroredRoutes.add(route);
     }
   }
 
-  return { docsDir, docsConfig, redirects, allFiles, relAllFiles, markdownFiles, routes };
+  return {
+    docsDir,
+    docsConfig,
+    redirects,
+    allFiles,
+    relAllFiles,
+    markdownFiles,
+    routes,
+    mirroredRoutes,
+  };
 }
 
 let defaultAuditIndex: ReturnType<typeof buildAuditIndex> | undefined;
@@ -446,7 +464,7 @@ function parseAuditUrl(
 /**
  * Audits local docs links against route, file, and redirect indexes.
  */
-function auditDocsLinks(
+export function auditDocsLinks(
   options: { docsDir?: string; allowExternalClawHubRoutes?: boolean; anchors?: boolean } = {},
 ) {
   const docsDir = options.docsDir ?? DOCS_DIR;
@@ -455,6 +473,7 @@ function auditDocsLinks(
   });
   const broken: Array<{ file: string; line: number; link: string; reason: string }> = [];
   let checked = 0;
+  let unverifiedMirroredFragments = 0;
 
   // The publisher writes physical/index routes; frontmatter permalinks do not
   // create pages. Only emitted pages and configured redirects can prove fragments.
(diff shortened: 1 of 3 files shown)


vincentkoc:
Both findings fixed in `9999bf149fd` (rebased onto current `main`). The PR body now carries the full before/after tables; below is each finding with the evidence that settles it.

Two notes on how the evidence was produced. This worktree has no sibling `clawhub/`, so it is the *source-absent* shape; `OPENCLAW_DOCS_SYNC_CLAWHUB_REPO` pointed at a real `openclaw/clawhub` checkout gives the *source-present* shape. And neither edge case is reachable from the repo as authored — no `/clawhub/**` link carries a fragment today, and no redirect destination does either — so each was reproduced by temporarily introducing the triggering content, then reverted.

---

### 1. Keep unverified fragment errors behind anchor mode — `scripts/docs-link-audit.mts:588-594`

Confirmed, and the diagnosis was exactly right. The branch is now gated:

```ts
const mirroredOnly = !page && index.mirroredRoutes.has(terminal);
if (mirroredOnly) {
  // Plain mode does not inspect fragments at all, so a declared mirrored
  // route resolves there whether or not the link carries one. Only anchor
  // mode reports the fragment it cannot verify without the source.
  if (options.anchors && destination.hash) {
    unverifiedMirroredFragments++;
    broken.push({ ... });
  }
  continue;
}
```

**Reproduced from the CLI.** With a page containing `[publishing](/clawhub/publishing#package-publish-source)` added to `docs/`, running `node scripts/docs-link-audit.mjs` (plain, no ClawHub source):

| | `broken_links` | exit |
| --- | --- | --- |
| `origin/main` | 0 | 0 |
| previous revision of this PR | **1** | **1** |
| current | 0 | 0 |

The previous revision's failure line was:

```
zz-clawhub-fragment-repro.md:10 :: /clawhub/publishing#package-publish-source :: fragment unverified without the ClawHub source checkout (terminal: /clawhub/publishing#package-publish-source); set OPENCLAW_DOCS_SYNC_CLAWHUB_REPO
```

The current revision's plain-mode output is byte-identical to `main`'s for that same tree. Anchors mode is unaffected by the gate — it still reports the link as unverified (4 findings with the repro page present, vs 43 on `main`).

**Covered through the CLI test boundary.** Two tests added, in the same `describe` block:

- `leaves fragments into mirrored routes alone in plain mode` — direct `auditDocsLinks({ allowExternalClawHubRoutes: true })`, expects `broken` empty and `unverifiedMirroredFragments === 0`.
- `exits clean from the CLI in plain mode for a fragment into a mirrored route` — `spawnSync` on `scripts/docs-link-audit.mjs` with no `--anchors`, expects `broken_links=0`, exit `0`, and no `fragment unverified` in stdout. The fixture root is nested one directory deeper than the existing CLI harness's so that `<root>/../clawhub` cannot accidentally resolve and flip the allowance off — this run has to be the source-absent shape to mean anything.

Both fail against the previous revision.

---

### 2. Report unverified fragments in mirrored redirect destinations — `scripts/docs-link-audit.mts:514-517`

Confirmed, including the subtlety that the link loop does not compensate when nothing references the redirect. The redirect loop now diagnoses it while processing the record, using the same reason string as the link path:

```ts
// A redirect into a ClawHub-mirrored route now resolves off the navigation
// declaration, which proves the route but not the fragment. Say so instead
// of letting the destination fragment go silently unchecked.
if (!page && destination.hash) {
  const terminal = normalizeRoute(destination.pathname);
  if (index.mirroredRoutes.has(terminal)) {
    unverifiedMirroredFragments++;
    broken.push({ file: "docs.json", line: 0, link: record.source, reason: ... });
  }
}
```

**Reproduced from the CLI.** Rewriting the existing `/tools/clawhub` redirect destination to `/clawhub/publishing#package-publish-source` and running `--anchors` with no ClawHub source — nothing in `docs/` links to `/tools/clawhub`, so this is exactly the unreferenced case:

```
main     docs.json:unknown :: /tools/clawhub :: Redirect has no terminal Markdown page: /tools/clawhub -> /clawhub/publishing
previous (no line mentioning /tools/clawhub — resolved silently, and unverifiedMirroredFragments stayed 0)
current  docs.json:unknown :: /tools/clawhub :: fragment unverified without the ClawHub source checkout (terminal: /clawhub/publishing#package-publish-source); set OPENCLAW_DOCS_SYNC_CLAWHUB_REPO
```

Totals for those runs: 42 / 3 / 4. Plain mode reports 0 in all three.

Two tests added: `reports unverified fragments in redirect destinations into mirrored routes` (anchors; asserts the count, `file`, `link`, and reason) and `keeps mirrored redirect destinations silent in plain mode` (the mode guard, so the fix for finding 2 cannot reintroduce finding 1).

---

### 3. Complete next step

Both cases are fixed and validated through the CLI test boundary, per the above.

---

### Verification on `9999bf149fd`

`pnpm check:docs` — the aggregate the CI docs job runs, not a hand-picked subset — **exit 0**, both with and without `OPENCLAW_DOCS_SYNC_CLAWHUB_REPO` set.

Plain mode is unchanged from `main` in every configuration, diagnostics included:

| run | `main` | this PR |
| --- | --- | --- |
| plain, no ClawHub source | 0 | 0 — byte-identical |
| plain, with ClawHub source | 0 | 0 — byte-identical |
| `--anchors`, with ClawHub source | 4 | 4 — byte-identical |
| `--anchors`, no ClawHub source | 42 | 3 — the intended change |

Tests: `src/scripts/docs-link-audit.test.ts` under `test/vitest/vitest.tooling.config.ts` is 32 passed / 1 failed. The failure is `preserves Mint component targets for raw component apostrophes normalized after separators` (`as-s-guide`), which reproduces unchanged at `origin/main`, where the same file is 24 passed / 1 failed. It is untouched by this PR.

`pnpm check:test-types` reports an identical set of 8 pre-existing `src/infra/*` errors with and without this change; none are in the audited files. (Run from a full clone — `tsgo` refuses to start in a git worktree whose `node_modules` is a symlink.)

@clawsweeper re-review

Issue #142922, opened 2026-09-09
Bug: system-agent delegation loses active run authority


0xCyda:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

Delegated `openclaw` calls can reach `openclaw.chat` without an active admitted-run authority and fail before the system agent can process the request.

### Steps to reproduce

1. Start OpenClaw 2026.9.3 in a Discord channel with a Full Access turn.
2. Invoke the built-in `openclaw` tool for a permitted system-control request, such as a configuration change.
3. Observe the tool fail before the system agent acts.

### Expected behavior

`openclaw` should carry the authority of the current admitted run into the system-agent delegation, so permitted requests can reach the system agent.

### Actual behavior

The Gateway rejects the request with: `delegated OpenClaw approval requires an active run authority`.

### OpenClaw version

2026.9.3 (1391f7c)

### Operating system

Linux 6.18.33.2-microsoft-standard-WSL2

### Install method

npm global

### Model

openai/gpt-5.6-terra

### Provider / routing chain

OpenClaw Gateway -> OpenAI OAuth -> Codex

### Additional provider/model setup details

Observed with the built-in `openclaw` delegate tool from Discord. The failing call carried session context and Full Access, but the delegated authority was absent at the system-agent approval boundary.

### Logs

```shell
Error: delegated OpenClaw approval requires an active run authority
```

### Impact and severity

**Affected:** system-control requests delegated through the built-in `openclaw` tool from Discord.
**Severity:** blocks workflow.
**Frequency:** reproduced from this channel until the locally installed runtime guard restored the current run authority.
**Consequence:** configuration and other permitted system-agent actions cannot run through the delegate path.

### Additional information

Regression status and version comparison: NOT_ENOUGH_INFO.

Root-cause investigation found `createOpenClawDelegateToolsForRun` receives `runId` in `OpenClawToolsOptions`, but its picked options omit it and the caller identity does not include `operationalRunInstance` or `approvalAuthority`. A local source patch that reacquires `getAgentRunContext(options.runId)?.delegatedAuthority` and passes that exact authority through the trusted caller context restored the path. A focused regression test passed (12/12 tests in `openclaw-delegate-tool.test.ts`). No upstream code change is included with this issue.

lisheguo:
[...]

PR #143006, opened 2026-09-09
fix(discord): restore OpenClaw delegation from guild messages

Marvinthebored:
Related: #142922

## What Problem This Solves

Discord guild-message turns could expose the `openclaw` tool but reject its requests with `delegated OpenClaw approval requires an active run authority`, including Full Access sessions.

## Why This Change Was Made

Preserve the existing host context builder through preflight and queued processing. Keep its closure in job runtime state so the existing consumer receives the same builder and Gateway binding. The approval gate, run ownership, sender policy, and cancellation rules stay intact.

The regression tests cover both handoff boundaries and materialization. The context test now preserves the builder's overloaded type while observing its calls, resolving the earlier test-type failure.

## User Impact

Allowed owners can again delegate a read-only system request from an ordinary Discord guild message. Allowed non-owners still lack owner-only tools, and disallowed senders are rejected before inference.

This repair covers the bundled Discord message handoff. The broader runtime, OAuth, external-plugin, and cross-process reports in #142922 remain unproved, so that issue should stay open.

## Evidence

- Real guild-message baseline: one delegate call, working inference readiness, the exact missing-authority error, and one visible error reply.
- Real candidate: the same registered builder survives preflight and job materialization; its Gateway resolver reaches the admitted run; in-process delegation and live authority checks lead to one reply containing the development checkout's actual source revision from system status. No release-version claim is made.
- Fresh independent guild validation passed owner, allowed non-owner, and disallowed-sender scenarios. QA leases released after every drive.
- 122 focused Discord tests and 94 authority/queue checks passed. Focused Telegram, Slack, WhatsApp, and iMessage context checks also passed.
- Formatting, targeted lint, line-count and assertion checks, and the normal commit hook passed on the isolated build host. Full type checking remains covered by the PR's hosted checks.

Private raw evidence retains the earlier fixture failures, status detail-probe refusal, source-only validation limits, and build-capture timing limits. The supported status result is used without bypassing the detail probe's device-identity requirement. No credentials or execution metadata are published.


obviyus:
Thanks @Marvinthebored!

**Proof:** Real Discord guild-message before/after drives used a deterministic provider and the actual system-agent tool. Independent owner, allowed non-owner, and disallowed-sender controls passed. All 220 focused cases and [exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34346050171) passed; the final security-aware review had no findings.

- **Bug:** An admitted message turn lost its Gateway binding and delegated requests failed with the active-run-authority error, including in Full Access sessions.
- **Fix:** Preserve the original host builder through preflight and queued runtime state while retaining existing sender, run, cancellation, and approval checks.

| Before | After |
| --- | --- |
| One visible missing-authority error reply. | One reply with the development checkout’s source revision from the real system-status result. |
| Permitted delegation lost its host binding. | The exact builder, admitted run, and in-process authority remain joined. |

This proves the bundled guild-message path. The broader runtime and transport reports in #142922 remain open and unproved. The status detail probe’s device-identity requirement was not bypassed, and no release-version claim is made.

PR #134290, opened 2026-08-31
refactor: reuse prepared plugin facts during Gateway turns

steipete:
Related: #133903

## What Problem This Solves

Concurrent Gateway turns repeatedly rebuild provider auth aliases, plugin ID normalization and discovery fingerprints even when their prepared plugin generation already contains those package facts. This consumes event-loop time that could serve chat history and control-plane requests.

## Why This Change Was Made

The existing immutable metadata snapshot now owns normalized auth-alias candidates. Scalar lookups select the requested alias directly; full-map callers retain declaration order and origin precedence. Current workspace trust is evaluated on every lookup, credentials are not cached, and callers supplying a mutable partial manifest view still get a fresh projection.

Prepared runtime scopes carry their existing metadata and registry without repeating discovery/policy hashing. Memory selection reuses the snapshot's plugin ID normalizer while applying current activation config. The runtime load-context owner already supplies install records, so its duplicate projection is removed. Snapshot projection, replacement and worker restoration share these facts through the existing lifecycle; no global memoization or new configuration, persistence, protocol or dependency surface is added.

Production: **148 added / 161 removed, net −13**. Tests/support: **334 added / 42 removed, net +292**. Documentation: **+5**. This is an internal performance refactor; release changelog is unchanged.

## User Impact

Gateway turns do less repeated metadata work while retaining current trust, activation and runtime authority. This is a measured improvement to one owner boundary, not a claim that every saturation source is fixed. Restart and explicit metadata-management behavior are unchanged.

## Evidence

Validated source was committed unchanged as `295f2adfc1a5757246817e7c7ebe0e3987ca5308`; the build/profile manifests pin its exact source afterimages.

- Three regressions fail on the original owners: repeated auth lookup enumerates an unrelated alias dictionary 24 times, memory selection rebuilds manifest normalization three times, and prepared scope entry recomputes a discovery fingerprint. All pass after the refactor. Behavior coverage includes alias precedence/order, prototype keys, caller mutation isolation, changing workspace trust, enable/disable transitions, nested scopes, projected/replaced/retained snapshots and worker restoration.
[...]

PR #143020, opened 2026-09-09
refactor(agents): use prepared metadata for runtime plugin plans

steipete:
## What Problem This Solves

The internal runtime-plugin planner still allowed callers to omit metadata and discover memory plugins itself, even though both production callers already prepare and supply that metadata. This left an unused discovery path to maintain and let direct tests exercise a different contract from production.

## Why This Change Was Made

Require the existing metadata snapshot on the internal planner and its private memory selector. Remove the fallback registry load, fallback alias normalization, unused workspace argument, and imports. The independently called harness-owner APIs keep their existing optional metadata contract. Production shrinks by 10 lines.

## User Impact

No user-visible behavior change. Memory selection continues to use the prepared generation's aliases and the current configuration's activation policy. The internal type contract now matches the production callers.

## Evidence

The 79 owner and sibling cases passed before and after the production cleanup. After correcting a lint issue in the new fixture, all 45 affected cases passed again with the existing assertions unchanged. All 29 canonical changed-file checks and fresh independent review passed on the final patch.

Related: #134290.

A clean Linux installation, the full `pnpm build`, and the plugin asset check passed on the exact patch. All 16 build stages and 152 SDK subpath checks passed; generated assets matched the committed source.

### DIFF
--- a/src/agents/harness/runtime-plugin-load-plan.ts
+++ b/src/agents/harness/runtime-plugin-load-plan.ts
@@ -21,10 +21,6 @@ import type {
   PluginMetadataSnapshotPluginIdScope,
 } from "../../plugins/plugin-metadata-snapshot.types.js";
 import {
-  loadPluginRegistrySnapshot,
-  normalizePluginsConfigWithRegistry,
-} from "../../plugins/plugin-registry.js";
-import {
   resolveActivatableProviderOwnerPluginIds,
   resolveBundledProviderCompatPluginIds,
   resolveOwningPluginIdsForProviderRef,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [52ef9dbafb739269565faaadbc28dac8a9bfafe5](https://github.com/openclaw/openclaw/pull/143020/commits/52ef9dbafb739269565faaadbc28dac8a9bfafe5)
- Landed commit: [f9781f6e5b25217a5b44e43b93a5dc9fdbd20228](https://github.com/openclaw/openclaw/commit/f9781f6e5b25217a5b44e43b93a5dc9fdbd20228)

PR #142284, opened 2026-09-08
fix(doctor): complete canonical model-reference migration

obviyus:
## What Problem This Solves

Doctor could miss retired references in media settings, fallbacks, model maps, and sessions. It could also misread provider-local names and remove valid cached fields.

## Why This Change Was Made

Complete repair in the existing reference, runtime-policy, and session owners. Canonical entries retain collision precedence; missing nested settings merge. Session repair captures the old route before rewriting it, preserving explicit choices and avoiding new absent fields.

## User Impact

Heartbeat, subagent, and session runtime selections survive migration. Repeated repair leaves config and sessions unchanged; credential-store and native session-binding migrations stay separate.

## Evidence

Real Doctor repair and public session-list commands reproduced missing policies and retired session runtime before the fix, then confirmed preservation and repeat-run stability. Six runtime-preservation regressions and existing Doctor suites passed, as did type checks, lint, and build. A channel test timed out in CI; its transport cause remained unknown. Proof used synthetic state without authentication or inference.

Consumers: repaired media references, heartbeat/subagent policies, and saved session runtime selections retain their intended meaning.

PR #143031, opened 2026-09-09
fix(mcp): close plugin resources after accepted stdio work finishes

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where standalone plugin-tool MCP services could leave plugin registrations open after shutdown, or lose the registration/SDK context needed by a cancelled tool's remaining work. A failed transport close could also be reported as successful serving.

## Why This Change Was Made

The stdio serving owner acquires its registration view before invoking tool factories and keeps it through accepted handlers, cooperating descendants, and harness cleanup. It then releases both the primary registration and nested SDK claims. Tool projection uses the original view and cannot revive a released registry.

Managed request cancellation keeps its original reason and serving context. Its relay remains connected while accepted descendants finish, including the MCP SDK's response-send backpressure window. Returning a cached tool result stays early; it does not cancel valid work. The reusable Server and caller-supplied tools retain their existing ownership/reconnect behavior.

## User Impact

Stopping either standalone plugin-tool service joins its admitted work before closing registration resources. Connection, transport-close and registration-disposal failures remain visible. Per-tool plugin identity in generic abort callbacks is a separate follow-up; this change covers serving registry, SDK host and work ownership.

## Evidence

- Native before/after failures cover plain/nested registration release, stale-view reactivation, cancellation under a foreign host, response-send backpressure, and combined connection/transport/disposal errors.
- The original 57 targeted cases and two additional transport self-close cases are covered by completed runs. Full 13-path checks and fresh independent Codex review through P2 passed. A fresh ordinary build completed in 200.6 seconds.
- Three actual compiled stdio scenarios passed with native SQLite, delayed tool writes and delayed disposal: plain, nested SDK lookups from factory/handler/abort/harness cleanup, and transport-close failure. All owned process trees joined under the existing 120-second deadline; observed runs took 1.41, 1.08 and 1.30 seconds.
- Two earlier fixture runs missed file-watch notifications despite their release files existing. The fixture now observes durable gate state using the repository's polling approach; deadlines and assertions are unchanged. Those failures, and an earlier unverified external process-owner attempt, remain excluded from passing proof.

CI previously hit the independent Telegram loopback failure fixed in #143197. This branch now includes that fix and removes the temporary investigation code. All ownership production files match the compiled proof; 21 post-refresh composition tests passed across native registrations, cancellation, stdio and the real Telegram HTTP fixture.

### DIFF
--- a/docs/plugins/architecture.md
+++ b/docs/plugins/architecture.md
@@ -170,6 +170,8 @@ Gateway shutdown also joins actual harness, MCP, LSP, embedding, and media clean
 
 Executable CLI cleanup reports each disposer that exceeds five seconds and proceeds with later cleanup without canceling the pending work. On macOS with Node's system CA support enabled, automatic exit after command completion waits for this pending cleanup to finish. Explicit command exit requests and the update exit watchdog retain their bounded behavior.
 
+Standalone plugin and Codex supervision MCP stdio services retain their discovered registrations through accepted tool work, harness cleanup, and nested SDK provider lookups. Terminal shutdown cancels and joins handlers before releasing these registrations and awaiting their resource disposers. Transport-close and registration-disposal failures reach the serving caller. Programmatic servers created from supplied tools leave those resources with the caller; closing and reconnecting the same server does not dispose them.
+
 Hot registry publication does not wait for retired host cleanup; terminal shutdown also joins cleanup already started by earlier registry replacements before resetting shared state. Activation and rollback apply only to their captured registry version. An activation superseded by a lifecycle callback reports an error.
 
 The cache rule is documented in [Plugin architecture internals](/plugins/architecture-internals#plugin-cache-boundary): Gateway retains one cache generation, while explicit management operations use isolated generations of the same cache. There are no wall-clock TTLs for Gateway metadata.
--- a/src/mcp/codex-supervision-tools-serve.test.ts
+++ b/src/mcp/codex-supervision-tools-serve.test.ts
@@ -8,37 +8,37 @@ import {
   serveCodexSupervisionToolsMcp,
 } from "./codex-supervision-tools-serve.js";
 
-type EnsureStandalonePluginToolRegistryLoaded =
-  typeof import("../plugins/tools.js").ensureStandalonePluginToolRegistryLoaded;
-type ConnectToolsMcpServerToStdio =
-  typeof import("./tools-stdio-server.js").connectToolsMcpServerToStdio;
-
-const ensureStandalonePluginToolRegistryLoadedMock = vi.hoisted(() =>
-  vi.fn<EnsureStandalonePluginToolRegistryLoaded>(() => undefined),
+const acquireStandalonePluginToolRegistryMock = vi.hoisted(() =>
+  vi.fn<typeof import("../plugins/tools.js").acquireStandalonePluginToolRegistry>(),
 );
 const resolvePluginToolsMock = vi.hoisted(() => vi.fn<() => AnyAgentTool[]>(() => []));
-const connectToolsMcpServerToStdioMock = vi.hoisted(() =>
-  vi.fn<ConnectToolsMcpServerToStdio>(async () => {}),
+const getRuntimeConfigMock = vi.hoisted(() =>
+  vi.fn<() => import("../config/config.js").OpenClawConfig>(() => ({})),
 );
-const disposeRegisteredAgentHarnessesMock = vi.hoisted(() => vi.fn(async () => {}));
 
-vi.mock("../agents/harness/registry.js", async (importOriginal) => {
-  const actual = await importOriginal<typeof import("../agents/harness/registry.js")>();
-  return { ...actual, disposeRegisteredAgentHarnesses: disposeRegisteredAgentHarnessesMock };
-});
-
+vi.mock("../config/config.js", () => ({ getRuntimeConfig: getRuntimeConfigMock }));
 vi.mock("../plugins/tools.js", async (importOriginal) => {
   const actual = await importOriginal<typeof import("../plugins/tools.js")>();
   return {
     ...actual,
-    ensureStandalonePluginToolRegistryLoaded: ensureStandalonePluginToolRegistryLoadedMock,
-    resolvePluginTools: resolvePluginToolsMock,
+    acquireStandalonePluginToolRegistry: acquireStandalonePluginToolRegistryMock,
   };
 });
-
 vi.mock("./tools-stdio-server.js", async (importOriginal) => {
   const actual = await importOriginal<typeof import("./tools-stdio-server.js")>();
-  return { ...actual, connectToolsMcpServerToStdio: connectToolsMcpServerToStdioMock };
+  const serve: typeof actual.serveRegisteredToolsMcpServer = async (params) => {
+    const { LegacyPluginSdkResourceHost } = await import("../plugins/legacy-sdk-resource-host.js");
+    const host = new LegacyPluginSdkResourceHost();
+    const acquisition = await host.run(params.acquireRegistry);
+    try {
+      const server = params.createServer(acquisition.resolveTools(), host);
+      await server.close();
+    } finally {
+      await acquisition.release();
+      await host.close();
+    }
+  };
+  return { ...actual, serveRegisteredToolsMcpServer: serve };
 });
 
 const TOOL_NAMES = [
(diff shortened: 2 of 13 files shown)


Yigtwxx (commented):
I read this at head `13c12d0639` (one commit past the head ClawSweeper reviewed, `4d127e6502e0`) and ran a few measurements against the new ownership path. The lifecycle work itself holds up; two things look like they are outside the scope the description claims, and one of them lands on a file this PR does not touch.

## 1. The third standalone stdio service silently inherits the new helper contract

`connectToolsMcpServerToStdio` changed shape in two ways:

- it lost `options.onShutdown`, and with it the `if (options.onShutdown) { await shutdownComplete; }` gate, so it now **always** blocks until shutdown;
- it no longer writes `MCP stdio shutdown failed: ...` to stderr; shutdown failures are now rethrown (`failures[0]` or an `AggregateError`).

Two of its three callers were rewritten to `serveRegisteredToolsMcpServer`. The third, `src/mcp/openclaw-tools-serve.ts`, is byte-identical to `main` and still calls the helper directly:

```ts
// src/mcp/openclaw-tools-serve.ts:89-94 (unchanged by this PR)
async function serveOpenClawToolsMcp(): Promise<void> {
  const server = createOpenClawToolsMcpServer({ ... });
  await connectToolsMcpServerToStdio(server);
}
// :97-100
serveOpenClawToolsMcp().catch((err: unknown) => {
  process.stderr.write(`openclaw-tools-serve: ${formatErrorMessage(err)}\n`);
  process.exit(1);
});
```

So for the `openclaw-tools` service, before this PR: `serveOpenClawToolsMcp()` resolved as soon as `server.connect()` resolved, and a close failure produced one stderr line and a normal exit. After this PR: the promise stays pending until shutdown, and a `server.close()` or transport-close failure rejects it, which the entry wrapper turns into **`process.exit(1)`**.

This is the MCP server the CLI runner spawns for the system-agent tools (`src/mcp/openclaw-tools-serve-config.ts:104`), so the exit code is observable by the parent session. It may well be the behavior you want, but the PR description scopes itself to "either standalone plugin-tool service" and there are three entrypoints on this helper, and no test covers `openclaw-tools-serve` shutdown. Worth either saying so explicitly or moving it onto `serveRegisteredToolsMcpServer` with the others.

Related: `createToolsMcpServer` now branches on `params.sdkResourceHost`, and `openclaw-tools` is the one caller that does not pass one, so it takes the `if (!params.sdkResourceHost) return await handlers.callTool(...)` path and gets none of the new per-request scoping. It has no plugin registry to release, so the registration half genuinely does not apply, but "requests get their own work scope" is now a property of which caller you are rather than of the server.

## 2. Shutdown joins accepted work with no bound anywhere in production

The description says the process trees "joined under the existing 120-second deadline". The only 120 s I can find is the external test-runner timeout in `src/mcp/plugin-tools-registration.stdio.test.ts:22` (`runNodeScript(..., 120_000, ...)`), which bounds the fixture, not the service. In production the chain is:

`shutdown()` -> `server.close()` -> `super.close()` -> `work.drain()` -> `while (pending.size > 0) await Promise.allSettled(pending)`

and the new tail adds `AsyncWorkScope.runWhenAllIdle(() => [work], () => work.drain())` on top, which loops until the child scope stops registering work. Neither has a deadline, and `connectToolsMcpServerToStdio` now sits on that path unconditionally (point 1).

Measured with `InMemoryTransport`, a tool that returns immediately and leaves one descendant registered via `trackAsyncWork`, then `server.close()` raced against a 1500 ms timer:

| Case | `server.close()` resolved | elapsed |
|---|---|---|
| descendant observes `getAsyncWorkSignal()` (control) | yes | **1 ms** |
| descendant ignores it, `sdkResourceHost` present | **no** | >1500 ms (only after the gate was released) |
| descendant ignores it, no `sdkResourceHost` | **no** | >1500 ms |

The control arm resolving in 1 ms is what makes the other two meaningful: the harness detects a bounded close when there is one.

To be fair about attribution: the unbounded `work.drain()` in `ToolsMcpServer.close()` is already on `main`, so this PR does not introduce it. What this PR does is put it on the exit path of every standalone stdio service unconditionally, and widen what is joined to include descendants that outlive the tool call. `shutdown()` also removes its own `SIGINT`/`SIGTERM` listeners on first entry, so the operator's escape hatch is a second signal hitting Node's default handler.

## 3. Cancellation never signals the descendants that shutdown then waits for

Measured on the same harness: with the client cancelling an in-flight call, the descendant's `getAsyncWorkSignal()` is **not** aborted, on either path.

```
PROBE_CANCEL_HOST   descendantAborted=false
PROBE_CANCEL_NOHOST descendantAborted=false
```

That looks deliberate - `abort` calls `controller.abort(...)` for the tool, while `work.beginClose(...)` is reached only through `closeWork`, i.e. only when the parent scope aborts - and it matches "it does not cancel valid work". Combined with point 2, though, it means the only event that ever signals a request's descendants is server close, and server close then waits for them with no bound. A client that cancels a slow tool leaves its descendants running unsignalled, and the service's exit time becomes whatever the least cooperative one decides.

A bounded join with a diagnostic naming what is still pending would make this diagnosable in the field without changing the cooperative case, which already settles in 1 ms.

## Smaller notes

- `serveRegisteredToolsMcpServer` only releases through `acquisition?.release()`, so if `params.acquireRegistry()` itself throws after `acquirePluginRegistryForInspection` has succeeded internally, `acquisition` stays `undefined` and nothing is released. Is `acquireStandalonePluginToolRegistry` all-or-nothing on its own failure paths?
- `OwnedStdioTransport.close()` calls `shutdown()` in its `finally`, and `shutdown()` calls `server.close()`, which calls `transport.close()` again. `shuttingDown` stops the recursion at the second `shutdown()`, but `super.close()` does run twice on the SDK transport in the self-close ordering. Worth a line saying that is intended and idempotent.
- `closeFailures` being a `Set` to dedupe the rethrown-and-recaught transport error is a nice touch, but it depends on the SDK handing back the same error instance; a wrapped rejection would be reported twice in the `AggregateError`.

Reproduction for the two measurements above is a single file under `src/mcp/` run with `test/vitest/vitest.unit-src.config.ts`; happy to attach it if useful.

Yigtwxx (commented):
Follow-up on the red `ci-gate` here, since this PR is otherwise marked ready: the failure is not from this branch.

`checks-node-changed-extensions-bundle-27` has exactly one failing test in the whole bundle:

```
FAIL extension-telegram extensions/telegram/src/model-callback.loopback.integration.test.ts
  > Telegram model callback loopback > sends, authorizes, resolves, persists, answers, and edits an opaque callback
AssertionError: expected '❌ Failed to change model: HttpError: …' to contain 'Model changed to <b>ollama/…</b>'
  Received: "❌ Failed to change model: HttpError: Network request for 'editMessageText' failed!"
  at extensions/telegram/src/model-callback.loopback.integration.test.ts:245:45
```

This PR touches `docs/plugins`, `src/mcp`, `src/plugins`, and `test/fixtures` - no `extensions/` path at all, and nothing under `extensions/telegram`.

Same bundle, four data points today:

| PR | head | bundle-27 | when |
|---|---|---|---|
| #143031 (this one) | `13c12d06393` | **FAILURE** - the assertion above | 12:56Z |
| #142284 | merged | SUCCESS | 14:15Z |
| #143140 | merged | SUCCESS | 14:26Z |
| #143175 | `a489e2d1e04` | **FAILURE** - identical assertion, identical line | 14:35Z |

#143175 touches `src/auto-reply`, `src/config` and a baseline file, also nothing under `extensions/`, and it fails on the same `:245` assertion with the same `editMessageText` network error on a different runner an hour and a half later. Two unrelated heads failing identically, with two green runs of the same bundle in between, puts this on the shard rather than on either branch. A rerun of `checks-node-changed-extensions-bundle-27` should be enough here.

I could not pin the flake's cause locally: the same test on plain `main` fails on Windows for an unrelated reason (`EBUSY unlink ...openclaw-agent.sqlite-shm` during teardown), which masks the network path, so I am not claiming a root cause for the flake itself - only that it is not this PR.

steipete:
Merged via squash.

- Prepared head SHA: [49c1b40da95738cfa72948890f36cdbc73d3e571](https://github.com/openclaw/openclaw/pull/143031/commits/49c1b40da95738cfa72948890f36cdbc73d3e571)
- Landed commit: [d7c38695127ae9e0d4e22b52011867278b77f594](https://github.com/openclaw/openclaw/commit/d7c38695127ae9e0d4e22b52011867278b77f594)

PR #143009, opened 2026-09-09
improve(test): reduce worker transform test time

steipete:
Related: #142911

## What Problem This Solves

The consolidated worker-transform test exhausted its 120-second Windows case budget during the sixth subprocess launch in [main CI](https://github.com/openclaw/openclaw/actions/runs/34330280215/job/102397192371). The first five launches passed their native execution and cache assertions.

## Why This Change Was Made

Combine the return from source mode with the existing leaf edit. Each layout now performs four complete worker builds and one source invocation, removing one redundant full build. This continues the setup reduction from #142911 by Vincent Koc.

The unchanged parent still proves compiled-cache reuse after source mode while the edited leaf proves content invalidation. Adjacent compiled generations retain unchanged leaf/parent reuse, and the final alias edit retains configuration invalidation. Vitest keys both ordinary modules using their ID, content and shared environment; OpenClaw contributes one uniform compiled-mode key. The removed standalone observation is unchanged-leaf reuse immediately after source mode, covered by those remaining checks under this owner contract.

## User Impact

Less repeated compilation in CI and local tooling tests. Both layouts, native SQLite/archive-worker execution, source/compiled URL checks, distinct generations, disposal checks and the 120-second deadline remain. Production code is unchanged; test delta is +2/-2.

## Evidence

- Matched local Node 24.19.0 / pnpm 12.3.4 full-file runs at the failing source: original 2/2 passed in 46.14s; candidate 2/2 passed in 36.46s (about 21% less wall time). This is one local pair, not a Windows or statistical speedup claim.
- Both layouts retain `C [1,1] -> C [1,1] -> S [2,2] -> C "second" [3,2] -> C "configured" [4,3]`; all launch assertions passed.
- Required changed checks passed, including root-test types and lint. Independent managed P2 review found no actionable findings.
- Native Windows [test-2](https://github.com/openclaw/openclaw/actions/runs/34335395737/job/102413788440) passed both cases at this exact head: `single` 97.365s, `projects` 92.461s, below the unchanged 120-second limit. All ten launch observations and eight distinct compiled-generation checks passed. This is one Windows validation run, not a matched-host speedup estimate.
- [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34335395737) passed, including both Windows jobs and 48 compact Node jobs in the existing PR plan. No workflow or shard settings changed.
- GitHub initially reported a workflow startup failure with zero jobs. One documented close/reopen recovery started CI on the unchanged commit; no test retry was requested.
[...]

PR #143034, opened 2026-09-09
fix(test): avoid competing full-worker builds in transform proofs

steipete:
Related: #142991

## What Problem This Solves

The two worker-transform proof sequences each build and launch several subprocess generations. Running both together on a small CI runner makes full-worker builds compete within each case's 120-second deadline. Both cases timed out after their fifth successful launch in an Image PR's CI run.

## Why This Change Was Made

Run these two sequences serially on every platform, extending their existing Windows behavior. Both layouts, all cache/source/edit transitions, every assertion, and the deadline remain unchanged.

## User Impact

This changes test scheduling only. It gives each case more deadline margin at the cost of a longer total run.

## Evidence

- Canonical single-file checks, formatting, and independent Codex review through P2 passed.
- A controlled Linux/Node 24.19 comparison used the same installed dependencies and source, pinned to two CPUs. Concurrent cases took 85.77/86.17 seconds; sequential cases took 64.51/58.34 seconds. The tightest deadline margin increased from 33.83 to 55.49 seconds. Total runtime increased from 88.79 to 125.18 seconds.
- Both comparison variants passed; this did not reproduce the original CI timeout. The comparison used base `34e0e5d7`, while the original CI used `0f32f846`; the test and runner owners match, but the complete compiled graph was not claimed identical across those two heads. The Testbox's hardware also differs from the original small CI runner.

Current main also removes one redundant build in #143009. This scheduling change preserves that five-launch sequence; the comparison above measured the earlier six-launch version.

### DIFF
--- a/test/scripts/vitest-worker-artifacts.transforms.test.ts
+++ b/test/scripts/vitest-worker-artifacts.transforms.test.ts
@@ -6,10 +6,8 @@ import { createWorkerArtifactTest, workerProbe } from "./vitest-worker-artifacts
 
 const root = process.cwd();
 const it = createWorkerArtifactTest();
-// Each sequence rebuilds all workers; avoid contention on Windows CI runners.
-const concurrent = process.platform !== "win32";
-
-describe("fresh compiled subprocess invocation", { concurrent }, () => {
+// Each sequence rebuilds all workers; avoid competing builds within one runner.
+describe("fresh compiled subprocess invocation", { concurrent: false }, () => {
   it.for((["single", "projects"] as const).map((layout) => ({ layout })))(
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ca59d3c18260f3f752447b73abb32ef5590aaedc](https://github.com/openclaw/openclaw/pull/143034/commits/ca59d3c18260f3f752447b73abb32ef5590aaedc)
- Landed commit: [f8dc8705bf234d8c542cc7a5c4e585a115ef823a](https://github.com/openclaw/openclaw/commit/f8dc8705bf234d8c542cc7a5c4e585a115ef823a)

PR #143039, opened 2026-09-09
fix(sqlite): preserve fatal WAL diagnostics from worker threads

steipete:
Related: #132844

## What Problem This Solves

The SQLite WAL containment tripwire could terminate a process without its structured fatal diagnostic when maintenance ran in a Worker thread. Node Workers do not expose `process.stderr.fd`, so the synchronous write threw before containment continued.

## Why This Change Was Made

Write synchronously to standard error descriptor 2. The diagnostic payload, sink-failure handling, SIGKILL/abort sequence, identity detector, and storage behavior remain unchanged. Extend the existing isolated replacement-WAL fixture to exercise both main-thread and Worker containment and join both owned children before cleanup.

## User Impact

Operators retain the fatal WAL mismatch diagnostic when containment originates in a Worker. This changes observability only; it neither weakens containment nor establishes the cause of any particular database mismatch.

## Evidence

- Actual Linux Node 24.19.0 regression: original main-thread case passes; original Worker case fails specifically because the fatal event array is empty. SIGKILL, absence of the fallback marker, and no parent-watchdog intervention are verified before that assertion.
- Repaired main and Worker cases pass, including structured diagnostic fields, replacement database integrity, and exact retained rows. The complete WAL, transaction, and busy-timeout suites pass 84 tests with no skips.
- Native receipts verify all four recovered runtime stages had joined, dead process groups.
- The remote transport was interrupted after those runtime stages. Their raw reports and receipts were recovered and independently verified, but no completed remote gate or final proof envelope is claimed. The remaining full mapped gate passed locally on the unchanged candidate.
- Fresh independent P2 review passed. Test growth is 37 net lines; production growth is one net line. No runtime savings claim.

### DIFF
--- a/src/infra/sqlite-wal.test.ts
+++ b/src/infra/sqlite-wal.test.ts
@@ -656,14 +656,14 @@ describe("sqlite WAL maintenance", () => {
     expect(close).not.toHaveBeenCalled();
     expect(maintenance.checkpoint()).toBe(false);
     expect(write).toHaveBeenCalledWith(
-      process.stderr.fd,
+      2,
       expect.stringContaining(`"sidecarPath":"${sidecarPath}"`),
     );
   });
 
-  it.runIf(process.platform === "linux")(
-    "preserves the replacement WAL family across fatal containment and reopen",
-    async () => {
+  it.runIf(process.platform === "linux").each(["main", "worker"] as const)(
+    "preserves the replacement WAL family across fatal containment and reopen from a %s thread",
+    async (thread) => {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [999021d7c7fb5b4be54403a0ea643278a74664ff](https://github.com/openclaw/openclaw/pull/143039/commits/999021d7c7fb5b4be54403a0ea643278a74664ff)
- Landed commit: [270b4adc0302b90188d72919c249208a15189390](https://github.com/openclaw/openclaw/commit/270b4adc0302b90188d72919c249208a15189390)

PR #143045, opened 2026-09-09
fix(tests): close acquisition databases after module resets

steipete:
## What Problem This Solves

Gateway acquisition tests can remove their temporary state while a database opened by a reloaded module remains live. The original behavior assertions pass despite the leaked handle.

## Why This Change Was Made

Import the state fixture at its six acquisition sites so cleanup uses the same module generation as the dynamic clients. This preserves the existing resets, assertions, socket deadlines, and exported fork helper without loading the fixture in parent cases that only delegate to a fork.

## User Impact

No production behavior changes. Test fixtures close their database handles before removing temporary state.

## Evidence

- The original ordered two-case diagnostic passed its behavior assertions but observed one handle still open after cleanup; the fixed diagnostic observed zero before safety cleanup and unlink.
- All 31 cases, including the three child-process scenarios, pass with their original names and order.
- Mapped checks and fresh independent P2 review pass.

No database was proactively opened by the diagnostic, and no WAL guard or timer was changed or triggered. No claim is made about the cause of an earlier full-suite process termination.

### DIFF
--- a/src/gateway/test-helpers.acquisition.test.ts
+++ b/src/gateway/test-helpers.acquisition.test.ts
@@ -12,7 +12,6 @@ import { WebSocketServer } from "../../packages/gateway-client/src/websocket.tes
 import { createVitestResourceOwner } from "../../scripts/lib/vitest-resource-ownership.mts";
 import { createDeferred } from "../../test/helpers/promise.js";
 import { runVitestShutdownCommand } from "../../test/helpers/vitest-shutdown-command.js";
-import { withOpenClawTestState } from "../test-utils/openclaw-test-state.js";
 import {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [10d6f24ae23dae6ce9681b2e4b5b7acf6f78225d](https://github.com/openclaw/openclaw/pull/143045/commits/10d6f24ae23dae6ce9681b2e4b5b7acf6f78225d)
- Landed commit: [e1bc2922ac3174b6304bdce29680e0004ba1d7e6](https://github.com/openclaw/openclaw/commit/e1bc2922ac3174b6304bdce29680e0004ba1d7e6)

PR #143048, opened 2026-09-09
refactor: remove duplicate runtime parity image capture

steipete:
Related: #139428

## What Problem This Solves

The runtime-parity suite repeats the same terminal-image capture and assertion in a standalone positive case and a stronger case that also rejects unrelated media and failed-step evidence. Those fixtures also duplicate the same provider request and transcript inputs.

## Why This Change Was Made

Remove the redundant standalone case and construct the three retained inputs with a local helper that requires the step status and evidence text. Preserve the real SQLite transcript and HTTP capture path, fresh nested values, input order, and all remaining assertions. The one-file change removes 60 net test lines, including the full helper overhead.

## User Impact

No product behavior changes. The suite avoids one duplicate capture while retaining the positive, unrelated-media and failed-step boundaries.

## Evidence

The complete runtime-parity suite and outcome sibling passed 31 cases before and 30 after. Exactly the redundant case was removed; all remaining per-file case order is unchanged. Structural expansion preserves all three retained inputs and every outside-edit token, including assertions and capture order.

The mapped changed gate passed, and fresh P2 review found no actionable issues. No runtime speedup is claimed.

### DIFF
--- a/extensions/qa-lab/src/runtime-parity.test.ts
+++ b/extensions/qa-lab/src/runtime-parity.test.ts
@@ -149,6 +149,38 @@ async function captureRuntimeParityWithMockRequests(params: {
   }
 }
 
+function createTerminalImageParityInput(
+  stepStatus: "pass" | "fail",
+  details: string,
+): Parameters<typeof captureRuntimeParityWithMockRequests>[0] {
+  return {
+    requests: [{ plannedToolName: "image_generate", plannedToolArgs: { prompt: "same" } }],
+    messages: [
+      { role: "user", content: "Generate the QA image." },
+      {
+        role: "assistant",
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9a5f55aeafa23e730f79792e31b7d90aec9c68a1](https://github.com/openclaw/openclaw/pull/143048/commits/9a5f55aeafa23e730f79792e31b7d90aec9c68a1)
- Landed commit: [b466c4464afc55aea6ff37c0cb4ca28512ea73ac](https://github.com/openclaw/openclaw/commit/b466c4464afc55aea6ff37c0cb4ca28512ea73ac)

PR #143049, opened 2026-09-09
fix(ui): keep session views consistent across concurrent updates

steipete:
## What Problem This Solves

Fixes an issue where users switching between or updating sessions could see the sidebar, selected session, and chat header disagree when requests finished out of order. Older list, history, or detail responses could overwrite a newer label, thinking setting, pin/read indicator, archive state, or run outcome. A selected child could also retain its previous label or parent after newer metadata arrived.

## Why This Change Was Made

The existing session capability now owns freshness across primary lists, filtered lists, detail reads, history responses, and accepted events. Each response retains when and for which connection it was requested; copied or optimistically decorated rows retain those observations. List membership remains with each list, and selected-session presentation observes the admitted row without inserting it into unrelated results.

Child ancestry follows the current selected session and retires when its identity or parent changes. Terminal metadata stays together for the run it describes, presentation cannot leak between known different agents, and the history Retry button uses the same refresh path as the chat pane. These changes replace competing reconciliation paths and add 2,814 production lines overall. Shared field provenance and connection/session tracking enforce the missing ordering rules while each list retains its own membership.

Devices page operations and attachment poster loading now sit behind their consuming boundaries to keep startup within the existing size budgets. The boot-module inventory was regenerated with the canonical tool. CI fixtures use real session capabilities and canonical mutation responses so they exercise the same ownership rules as the application.

## User Impact

Session views converge on the newest accepted facts even when background refreshes, events, and user actions overlap. Selected children update their label and ancestry, successful history retries recover the transcript, and an older run cannot acquire a newer run's terminal status.

There are no new operator settings or schema migrations. Roster-cache version, scope, retention, size limits, incognito filtering and volatile-field projection are unchanged; corrected row values can flow into that existing cache. Freshness metadata remains in memory. The added UI patch-response timestamps already come from the Gateway; wire producers and stored session types are unchanged.

## Evidence

- Combined UI unit coverage: 980 suites, 16,856 passed cases and one existing opt-in Chromium skip. The original full run hit a zero-case Mermaid collection failure from a borrowed dependency path; a private installation and the same 13-case file resolved that gap. This count combines that run and its focused recovery.
- The Gateway repair passed 42 unique cases on a clean Linux installation. Six diagnostic repeats are excluded from that count; the original local disk-capacity failures remain recorded.
- Synthetic browser proof reproduced the stale selected-child label before the repair. The later 11-case run passed child/breadcrumb refresh, history Retry/loading, pin ordering, placement, terminal attention, video previews and scroll behavior. An additional video case passed with the complete build retained and verified; HTTP byte capture covered 22/54 assets, so exhaustive served-asset verification is not claimed. These are mock-Gateway checks, not live provider tests.
- The responsive lane initially had 188 passes and one animation-frame measurement failure. The independently landed atomic-measurement fix from #143103 is preserved exactly in the reconciled branch.
- The pre-reconciliation build measured 351,929 startup compressed bytes against the unchanged 352,235-byte cap; the largest ordinary chunk measured 184,092 bytes against 220,160. No performance budget was raised.
- The full changed-file gate and Madge passed after the CI repairs. After reconciling main, 388 focused cases across three affected files, UI typechecking and diff checks passed.
- The original full review's nine suggestions were checked against producer and lifecycle contracts and were not accepted. A fresh three-pass review of the complete CI-repair increment found no actionable issues through P2. Exact-head CI passed after one diagnostic rerun of the Matrix bundle job, with source, assertions and deadlines unchanged. The original first-admission timeout did not reproduce locally; no timeout fix is claimed.

Before: the selected child and breadcrumb retain Research in progress after newer metadata arrives

After: the selected child and breadcrumb both show Research completed


steipete:
Reviewed the complete 124-file diff at `4b50db128d8dce18145158eba47d30b6e90babc2` for both compatibility checklist items. The remaining generated flags come from `unknown-truncated-pull-files`; the technical review correctly identifies no migration requirement.

There are no new operator settings. Configuration changes are shrink-only lint inventories and generated/internal UI chunking data. The five additional UI patch-response timestamp fields already come from the unchanged Gateway response producer.

The roster-cache definition, reader, writer and lifecycle preserve version 1, credential/profile scope, retention, size limits, incognito exclusion and volatile-field stripping. Corrected row values can reach the existing cache, while freshness observations remain in memory and are not serialized. No stored schema, storage format or migration changes are required. The PR body records this distinction and the proof limits.

Required CI is now green on the same head after one diagnostic rerun of the failed Matrix job. Its first-admission timeout did not reproduce in the unchanged four-case Node 24.19 test file; source, assertions and deadlines were preserved, and no fix for that timeout is claimed.

steipete:
Merged via squash.

- Prepared head SHA: [4b50db128d8dce18145158eba47d30b6e90babc2](https://github.com/openclaw/openclaw/pull/143049/commits/4b50db128d8dce18145158eba47d30b6e90babc2)
- Landed commit: [1ad9fe2a759d57465e4898eecaa0ffd19af2a020](https://github.com/openclaw/openclaw/commit/1ad9fe2a759d57465e4898eecaa0ffd19af2a020)

PR #143053, opened 2026-09-09
docs(ci): split CI scope and routing into selection, node lanes, budgets, and dispatches

vincentkoc:
## What

`docs/ci/scope-and-routing.md` was **48,194 bytes** on a single page and mixed four
distinct reader jobs. This moves four contiguous blocks into
`docs/ci/scope-and-routing/` and keeps the parent as an index, matching the
`docs/ci.md` split already on `main` and the recent `docs/install/updating/` split.

| Page | Bytes |
| --- | --- |
| `docs/ci/scope-and-routing.md` (index) | 48,194 → 2,143 |
| `docs/ci/scope-and-routing/selection.md` | 10,728 |
| `docs/ci/scope-and-routing/node-test-lanes.md` | 26,109 |
| `docs/ci/scope-and-routing/job-budgets.md` | 7,130 |
| `docs/ci/scope-and-routing/manual-dispatches.md` | 5,679 |

`node-test-lanes.md` stays large because it is a single flat 26-bullet list under one
introductory sentence, with no interior headings. Cutting that list across a page
boundary, or minting headings to justify a cut, would not be content-preserving, so it
moves as one block.

## Losslessness (proven)

- The four children, with frontmatter and the authored lede paragraph removed and the
  index's retained `## Related` section appended, concatenate to a **byte-identical**
  copy of the original body: sha256 `fcb941adb8967620266d59711e76addaff7403050210a7d38208444a372a1a3b` (47,933 bytes), before and after.
- Fences match one-for-one on info string + body sha256 (one `bash` fence,
  `999a9141d8d2f1a8e83d20e0cd5bb36fd08aba36c3bf448e9b1b7b75282b832b`).
- The original body contained no tables (0 rows before, 0 rows after in moved content;
  the index gains a new navigation table).
- Word count grows only (6,707 → 7,136), from the index table, the stub list, and the
  four one-sentence child ledes.

No prose was rewritten or reordered. `python3 .audit/check-orphan-refs.py` reports two
directional references, both in `node-test-lanes.md` and both fine: `"still admitted
separately below 150 seconds"` is numeric, not directional, and `"the hosted Vitest lane
set above"` refers to a bullet that is on the same child page. **The entire prose diff is
the index rewrite plus the four child ledes.**

## Anchors

Ids were enumerated with the repo's own `parseDocsDocument` (`scripts/lib/docs-markdown.mjs`),
not a slug approximation. Four pre-split ids:

| Id | Disposition |
| --- | --- |
| `scope-and-routing` | stubbed on the index → `/ci/scope-and-routing/selection#scope-and-routing` |
| `manual-dispatches` | stubbed on the index → `/ci/scope-and-routing/manual-dispatches#manual-dispatches` |
| `windows-testbox-probe` | stubbed on the index → `/ci/scope-and-routing/manual-dispatches#windows-testbox-probe` |
| `related` | still published by the index itself (no stub — that would collide) |

Collisions are empty on the index and on all four children. This keeps working:
the `/ci/manual-dispatches` → `/ci/scope-and-routing#manual-dispatches` redirect in
`docs/docs.json`, the three `/ci/scope-and-routing#…` entries in `docs/ci.md`'s "Where
each section moved" list, and `docs/ci/pipeline.md`'s page-level link.

## Pins and CI path allowlists

Classified every hit of `git grep -n "scope-and-routing" origin/main` (no path filter),
plus `python3 .audit/docs-test-pins.py`:

- **`test/scripts/docs-sync-publish.test.ts:478`** — the only real pin. It asserts the
  exact flattened page list of the "Release & CI" tab. It reads `docs.json` navigation,
  not the page body, and asserts no literal strings from this file, so this is a
  navigation pin, not a content contract. The four child routes are added to
  `releaseRoutes` in nav order, matching how `reference/test/*`,
  `reference/full-release-validation/*`, and `concepts/qa-e2e-automation/*` children are
  already listed there. **Not a refusal case.**
- **`scripts/changed-lanes.mts`** — `RELEASE_METADATA_PATHS` names `docs/install/updating*`
  and `CONFIG_DOC_BASELINE_PATHS` names only `docs/.generated/config-baseline.*`. Neither
  names this page, so `scripts/check-release-metadata-only.mts` is unaffected and nothing
  needs adding.
- **`.github/workflows/**`, `scripts/**`** — no hard-coded reference to `docs/ci/…` paths;
  the only `docs/ci` strings in workflows are comments pointing at `docs/ci.md#…` anchors.
- **`.github/labeler.yml`** — the `"docs"` label uses the recursive glob `docs/**`, and no
  file-exact glob names this page. The new directory is matched automatically; this split
  creates no labeler gap and none pre-exists for this page.
- `docs/docs.json` — nav updated; the existing redirect needs no change (see Anchors).
- **`test/scripts/package-acceptance-workflow.test.ts:12390`** — a pin that neither
  `.audit/docs-test-pins.py` nor `git grep "scope-and-routing"` could find, because it
  names no page: it reads `docs/ci.md` plus a flat `readdirSync("docs/ci")` filtered to
  `.md` names, then asserts literal strings appear somewhere in that set. The filter drops
  directories, so `docs/ci/scope-and-routing/*` was never read and
  `VALIDATION_SHA="<full-commit-sha>"` — which moves verbatim into
  `manual-dispatches.md` — looked deleted. **CI caught it; local greps did not.** This is a
  gap the split creates, so it is fixed here, in the second commit: the walk becomes
  recursive, which is exactly what the assertion's own comment already asks for ("Read the
  whole set so this assertion follows the content instead of a single file path"). The
  assertion is inclusion-only (`expectTextToIncludeAll`), so reading more files cannot
  weaken it. Verified both ways: the named test passes with the fix and fails without it,
  with the split in place. The sibling `docs/reference/full-release-validation` walk in the
  same test has the identical latent flat-readdir gap, but no subdirectory exists under it
  today, so nothing is broken there and it is left alone.
  **This is not a rule-5 refusal case:** the test does not read this single file, and it
  asserts that content exists somewhere in the CI docs set — a set the split keeps whole.
  The recursive walk names its encoding (`{ encoding: "utf8", recursive: true }`): without
  it, `readdirSync` resolves to the `string[] | Buffer[]` overload and `check-test-types`
  fails with `TS2339: Property 'endsWith' does not exist on type 'string | NonSharedBuffer'`.
  Both directions were checked in isolation under `--strict`.

## CODEOWNERS

`.github/CODEOWNERS` has no rule covering `docs/ci/`. Under last-match-wins, no maintainer
team owns this page or the new directory.

## Validators

Run on the staged tree in a worktree with `node_modules` linked. Baseline measured at the
merge base `38d2f4822be` after a clean rebase onto `main`.

- `pnpm check:docs` — **exit 0** (format-docs, markdownlint x2, check-docs-mdx over 1,134
  files, i18n glossary, link audit, config examples).
- `pnpm format:check` (repo-wide, also run by the `check-docs` job) — **exit 0**, 35,425 files.
- `node scripts/docs-link-audit.mjs --anchors` — **3 errors before, 3 after, identical
  set**; only `checked_internal_links` moves, 11,933 → 11,948. Those 3 are the known
  `maturity/taxonomy#windows-app-node` baseline. The baseline was measured in a separate
  detached worktree at the merge base `38d2f4822be`, not by stashing. The 39 `/clawhub/*`
  false positives that used to sit in this baseline are gone: #143002 fixed the checker
  before this PR was rebased.
- `npx vitest run --config test/vitest/vitest.tooling.config.ts test/scripts/docs-sync-publish.test.ts`
  — 1 file, 11 tests, **pass**.
- `npx vitest run --config test/vitest/vitest.full-core-unit-src.config.ts src/docs/`
  — 8 files, 16 tests, **pass**.
- `npx vitest run --config test/vitest/vitest.tooling.config.ts src/scripts/docs-link-audit.test.ts`
  — 1 failure, `preserves Mint component targets for raw component apostrophes normalized…`.
  **Pre-existing**: reproduced identically at `056d14028ff` with the change stashed.

## Glossary

`scripts/check-docs-i18n-glossary.mts` requires a `source` entry for the **frontmatter
title** of each new English doc. Four entries were added — `Scope selection`,
`Node test lanes`, `Job budgets and platform lanes`, `Manual dispatches` — inserted
next to the existing `CI scope and routing` entry rather than appended at the end, so
concurrent splits touch different regions of the file and merge cleanly. Append-only;
no existing target was rewritten or reordered. Targets follow the file's own
conventions (`lanes` → `通道`, `selection` → `选择`).

## Rebase

Rebased cleanly onto `38d2f4822be` after three concurrent docs splits landed. The zh-CN
glossary did **not** conflict, because the four new entries were inserted next to the
existing `CI scope and routing` entry rather than appended at the end of the array. The
glossary result was verified set-wise: a true superset of `origin/main`, 0 duplicate
sources, 0 removed sources, exactly 4 added. Every validator above was re-run after the
rebase; nothing was carried forward from the pre-rebase run.


vincentkoc:
@clawsweeper re-review

The previous verdict covered `e7e27ce83e4`. The branch has since been rebased cleanly onto `38d2f4822be` (current head `7e50187981d`); the tree is unchanged apart from the rebase, and every validator was re-run after it — `pnpm check:docs` exit 0, repo-wide `pnpm format:check` exit 0, the anchors audit at 3 known `#windows-app-node` errors before and after (measured against a detached worktree at the merge base, not a stash), and the two Vitest suites green. The byte-identity proof still holds at the same sha256.

vincentkoc:
@clawsweeper re-review

Head is now `3d9c217d02b`. CI on the previous head found one real failure that no local grep could reach, and it is fixed in the second commit.

`test/scripts/package-acceptance-workflow.test.ts` reads `docs/ci.md` plus a flat `readdirSync("docs/ci")` filtered to `.md` names and asserts literal strings appear somewhere in that set. The filter drops directories, so `docs/ci/scope-and-routing/*` was never read and `VALIDATION_SHA="<full-commit-sha>"` — which moves verbatim into `manual-dispatches.md` — looked deleted. The walk is now recursive, which is what the assertion's own comment already asks for ("Read the whole set so this assertion follows the content instead of a single file path"). The assertion is inclusion-only, so reading more files cannot weaken it.

Verified both directions with the split in place: the named test passes with the fix and fails without it. The byte-identity proof is unchanged, `pnpm check:docs` and repo-wide `pnpm format:check` are exit 0, and the branch still merges cleanly into `main`.

vincentkoc:
@clawsweeper re-review

Head is now `fac930548d1`. The previous head failed `check-test-types` with `TS2339: Property 'endsWith' does not exist on type 'string | NonSharedBuffer'` — `readdirSync(path, { recursive: true })` resolves to the `string[] | Buffer[]` overload. Fixed by naming the encoding explicitly: `readdirSync("docs/ci", { encoding: "utf8", recursive: true })`, which selects the `string[]` overload.

Verified: the isolated pattern type-checks clean under `--strict` with the encoding and reproduces the exact TS2339 without it, the named package-acceptance test passes, and repo-wide `pnpm format:check` is exit 0. No docs bytes changed since the last review — the split commit is untouched at `7e50187981d`.

vincentkoc:
@clawsweeper re-review

Both before-merge items addressed. The follow-up commit you flagged was real and was **missing from the PR head** — it existed only in the local worktree and had never been pushed. It is now on the branch at head `fac930548d1`.

That commit is `test(ci): follow the docs/ci tree when asserting CI documentation content` — it updates `test/scripts/package-acceptance-workflow.test.ts`, which asserts CI documentation content and would otherwise still be reading the pre-split single file. Verified green on the pushed head: `package-acceptance-workflow.test.ts` passes under `vitest.tooling.config.ts`.

Your wording pointed at a commit that was not in the PR, which read as a review artifact — it wasn't. Thank you for the catch.

PR #143059, opened 2026-09-09
refactor(tests): remove duplicate warm chat metadata scenario

steipete:
Related: #139428

## What Problem This Solves

The chat refresh suite runs the same warm metadata recovery scenario twice. The original unavailable-versus-empty model distinction became obsolete when #142165 moved catalog reads to their direct owner and excluded models from the commands cache.

## Why This Change Was Made

Keep one warm commands-cache case for startup responses that omit metadata. Preserve its catalog, loading-state, and revalidation assertions and subscription cleanup. The separate store regression still proves legacy model fields are discarded; the competing startup/direct catalog regression remains unchanged.

## User Impact

No user-visible behavior changes. This removes one duplicate test invocation and nine net test lines.

## Evidence

The complete chat-send and chat-metadata-store suites passed before (363 tests) and after (362 tests), with zero skips. Raw result comparison confirms exactly the two obsolete rows were replaced by one case; every other test name and order is unchanged. The mapped changed-file gate and fresh independent P2 review passed. No full-suite runtime improvement is claimed.

### DIFF
--- a/ui/src/pages/chat/chat-send.test.ts
+++ b/ui/src/pages/chat/chat-send.test.ts
@@ -691,42 +691,33 @@ describe("refreshChat", () => {
     },
   );
 
-  it.each(["missing", "empty"])(
-    "revalidates a warm %s snapshot when startup omits metadata",
-    async (kind) => {
-      const ready = { id: "model", name: "Model", provider: "test", available: true };
-      const host = makeChatHost({
-        requestHandlers: {
-          "chat.startup": async () => ({ messages: [] }),
-          "chat.metadata": async () => ({ commands: [] }),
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [68e52803fc9bf4fdac3f747fd60c10fdc2984015](https://github.com/openclaw/openclaw/pull/143059/commits/68e52803fc9bf4fdac3f747fd60c10fdc2984015)
- Landed commit: [b13971bdcc4580760029fb474d0e7021360455f2](https://github.com/openclaw/openclaw/commit/b13971bdcc4580760029fb474d0e7021360455f2)

PR #143071, opened 2026-09-09
fix(tests): close skill-upload fixture database handles

steipete:
## What Problem This Solves

The ordinary successful skill-upload test passed while leaving its dynamically loaded shared-state database open after fixture cleanup returned. Its fixture and closer came from the initial module graph, but `makeHarness` reset modules before loading the handlers and their separate database cache.

## Why This Change Was Made

Reset modules before acquiring the fixture, then load the existing fixture helper and handlers from that same generation. Each fixture now owns the cleanup functions for the database its handlers open. Keep the static database opener and closer used by existing SQL assertions, and keep fixture tracking before the awaited handler import.

## User Impact

No production behavior changes. This test-fixture repair closes observed database handles before removing temporary state. All test inputs, assertions and timers remain unchanged; the diff removes one net test line.

## Evidence

- The original ordinary successful-upload diagnostic passed, yet its actual dynamic database handle remained open after normal cleanup returned. The initial-generation handle was already closed.
- The same ordinary case on the repair observed the dynamic handle open before cleanup and closed afterward, before any diagnostic safety-close or unlink. Safety-close changed nothing; WAL/SHM were already absent and the fixture root was removed normally.
- The repaired complete existing owner suite passed all 10 cases. Parser-aware checks verify every case body, assertion, static SQL oracle and original closer is unchanged.
- Mapped changed gate, deslop and fresh P2 review passed. Diagnostics captured only naturally opened, fixture-owned state handles; no proactive database opens or timer changes.

This suite is Gateway-methods. The result does not attribute a historical Gateway-core failure or claim global runtime savings.

### DIFF
--- a/src/gateway/server-methods/skills-upload.test.ts
+++ b/src/gateway/server-methods/skills-upload.test.ts
@@ -10,10 +10,7 @@ import {
   closeOpenClawStateDatabaseForTest,
   openOpenClawStateDatabase,
 } from "../../state/openclaw-state-db.js";
-import {
-  createOpenClawTestState,
-  type OpenClawTestState,
-} from "../../test-utils/openclaw-test-state.js";
+import type { OpenClawTestState } from "../../test-utils/openclaw-test-state.js";
 import type { GatewayRequestHandlers } from "./types.js";
 
 const agentScopeState = vi.hoisted(() => ({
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [53c1d93a3f7b75e3c260d7026e70ea3607c03c92](https://github.com/openclaw/openclaw/pull/143071/commits/53c1d93a3f7b75e3c260d7026e70ea3607c03c92)
- Landed commit: [faaf218fa042d4855ae39b3f9abc507d1fd5f892](https://github.com/openclaw/openclaw/commit/faaf218fa042d4855ae39b3f9abc507d1fd5f892)

PR #143089, opened 2026-09-09
fix: node exec on Windows corrupts quoted arguments before the command runs

### LINKED ISSUE #142847: [Bug]: Windows node host mangles quoted arguments in native exec — runCommand never sets windowsVerbatimArguments
**Bug type:** Bug — node host / exec
**Beta release blocker:** No
**OpenClaw version:** 2026.9.1 and 2026.9.2 (node hosts); root cause still present on `main` @ `b72b30249f`
**Operating system:** Windows 11 (node host), macOS 15 (gateway)
**Model / provider chain:** not model-dependent — this is in the process spawn layer

## Summary

When the gateway runs a native `exec` on a Windows node host, the node host receives the shell-wrapper argv `["cmd.exe", "/d", "/s", "/c", <payload>]`, where `<payload>` is already a complete `cmd` command line.

`src/node-host/invoke.ts` `runCommand()` hands that argv straight to `runCommandWithTimeout()` without `windowsVerbatimArguments`. Node then applies its normal per-argument quoting to the payload, which inserts backslashes in front of every quote the payload contains. The command that actually runs is not the command that was sent.

The same argv shape spawned through the core exec path is fine, because that path does resolve `windowsVerbatimArguments`. The node-host path is the asymmetry.

## What triggers it

All three conditions together:

1. The command runs on a **Windows** node host (not the gateway host, not Linux/macOS nodes).
2. It goes through **native `exec`** (`host=node`), so it travels through `src/node-host/invoke.ts`.
3. The command payload contains **double quotes** — a quoted URL, a quoted path with spaces, a quoted argument value.

It gets worse when the payload both starts and ends with a quote (a quoted executable path followed by a quoted argument): that case fails outright rather than merely being altered.

It does **not** reproduce when you run the same command string manually over SSH or in a `cmd` window on the same machine, because those never traverse the node-host spawn layer. That divergence is what makes the bug hard to attribute — we spent a day looking at DNS and endpoint security before looking at spawn.

## Evidence (measured on a real Windows node)

Run on Windows 11, Node **v24.19.0**, spawning `cmd.exe` directly with `child_process.spawnSync`, comparing three configurations of the identical argv:

- **A** = what upstream does today (`windowsVerbatimArguments` unset)
- **B** = same argv, `windowsVerbatimArguments: true`
- **C** = payload wrapped in one extra pair of quotes **and** `windowsVerbatimArguments: true`

| # | payload sent | A — upstream today | B — verbatim only | C — wrap + verbatim |
|---|---|---|---|---|
| 1 | `echo "https://tenant.example.com/"` | `\"https://tenant.example.com/\"` | `"https://tenant.example.com/"` | `"https://tenant.example.com/"` |
| 2 | `echo "https://x.example.com/?a=1&b=2"` | `\"https://x.example.com/?a=1&b=2\"` | correct | correct |
| 3 | `echo "C:\Program Files\demo\tool.exe" --flag` | `\"C:\Program Files\demo\tool.exe\" --flag` | correct | correct |
| 4 | `echo "C:\dir\" end` | `\"C:\dir\\\" end` | correct | correct |
| 5 | `"C:\Windows\System32\cmd.exe" /c echo "tail"` | **fails:** `'\"C:\Windows\System32\cmd.exe\"' is not recognized as an internal or external command` | **fails:** `The directory name is invalid.` | `"tail"` |

Two things to note in that table:

- In cases 1–4 the corruption is silent. The command still runs; it just runs with different arguments than were sent. The caller sees a plausible downstream error (a bad URL, a missing file) and looks in the wrong place.
- **Case 5 shows that setting `windowsVerbatimArguments: true` alone is not sufficient.** With `/s`, `cmd` strips the first and last quote of the whole command line, so a payload that already begins and ends with a quote loses them. Only the wrap-plus-verbatim combination survives all five cases.

Our original real-world symptom was case 1: an argument `"https://tenant.sharepoint.com/"` reached the target program with a leading backslash and a trailing literal quote, the program built `https://https//…%22`, and the browser reported a DNS failure.

## Why it happens

- `src/node-host/invoke.ts:344` — `runCommand()` calls `runCommandWithTimeout(argv, { baseEnv, cwd, killProcessTree, … })`. `windowsVerbatimArguments` is never passed.
- `src/process/exec-spawn.ts:111-126` — the core path pulls `windowsVerbatimArguments` out of its options, passes it to `resolveSafeChildProcessInvocation()`, and applies `invocation.windowsVerbatimArguments` to `execa`.
- `src/process/windows-command.ts:160` — `windowsVerbatimArguments: useCmdWrapper ? true : params.windowsVerbatimArguments`. `useCmdWrapper` comes from `isWindowsBatchCommand()`, which is true only for `.cmd` / `.bat`. Here the executable is `cmd.exe`, extension `.exe`, so `useCmdWrapper` is `false` and the caller-supplied value is `undefined` — no verbatim, default escaping.

Worth noting: upstream already models this exact argv shape. `src/infra/shell-wrapper-resolution.ts:330-361` recognizes the `/d`, `/s`, `/c` tokens, and `src/node-host/invoke-system-run.ts` uses `isShellWrapperInvocation` / `resolveShellWrapperTransportArgv` / `isCmdExeInvocation` to keep `cmd.exe /c` approval-gated. The policy layer understands the shape; the spawn layer does not honor it.

## Suggested fix

**(a) Narrow — at the node-host boundary.** In `runCommand()`, detect the canonical five-element invocation `[cmd.exe | %SystemRoot%\System32\cmd.exe, /d, /s, /c, payload]` (case-insensitive switches), wrap the payload once for cmd's outer envelope, and pass `windowsVerbatimArguments: true`. Everything else keeps current behavior: PowerShell, direct executables, other `cmd` paths, non-canonical argv, non-Windows hosts. Environment, cwd, cancellation, timeout, process-tree termination, output limits, execution policy and approval handling are untouched.

**(b) Broad — at the shared boundary.** Teach `resolveSafeChildProcessInvocation()` to recognize the `cmd.exe /d /s /c` shape and handle it there, fixing every caller at once. Larger blast radius; needs a sweep of callers that already pass the flag explicitly.

Either way, per case 5 above, the fix must be **wrap + verbatim**, not verbatim alone.

## Result after the fix

We have been running (a) in production since 2026-09-08 on Windows node hosts at 2026.9.1 and 2026.9.2.

- The five cases above: A fails or corrupts in all five; C is correct in all five.
- 41 automated checks covering scope, payload preservation, original run options, installer rejection, backup and idempotence.
- Nine live Windows argument cases against the real patched function and the installed process runner, including a quoted executable path with spaces, an empty value, nested quotes and a trailing backslash.
- End-to-end gateway-to-node native `exec` after a node-host restart: a quoted URL, spaces and an ampersand all arrive with no characters added or lost. The browser flow that originally failed now opens the correct page.
- No regression observed in non-Windows nodes, PowerShell invocations or direct-executable invocations, which the patch does not touch.

Happy to open a PR for (a) if that is the direction you prefer.

## Impact

Any Windows node host running a native `exec` whose command carries quoted arguments. Not tied to a particular tool, model or provider. Because the corruption surfaces inside the invoked program — as a malformed URL, a missing file, a wrong flag value — it is likely to be misattributed to that program rather than to the spawn layer, which is probably why it has not been reported before.

## Search performed before filing

No existing issue found for `windowsVerbatimArguments node exec`, `node exec Windows quoting cmd`, or `node host exec quotes mangled Windows`. The only adjacent hit, #65724 (`[Windows] DeprecationWarning: Passing args to child process with shell option true`), is closed and concerns `shell: true`, a different path.

igs-rogenlo:
Closes #142847

## What Problem This Solves

Fixes an issue where operators running a native `exec` on a **Windows node host** would have their command silently altered when it contained a double-quoted argument — a quoted URL, a quoted path with spaces, a quoted flag value.

The command still ran, and it ran without error, so the corruption surfaced inside the invoked program instead: a malformed URL, a file that could not be found, a flag with the wrong value. In our case an argument `"https://tenant.example.com/"` arrived with an added leading backslash and a trailing literal quote, the target program built `https://https//…%22`, and the browser reported a DNS failure. We spent a day looking at DNS and endpoint security before looking at the spawn layer.

Running the same command string manually over SSH or in a `cmd` window on the same machine does **not** reproduce it, because those never traverse the node-host spawn path. That divergence is what makes the problem hard to attribute to OpenClaw at all.

## Why This Change Was Made

The gateway hands the node host a complete `cmd` command line as a single argv entry (`["cmd.exe", "/d", "/s", "/c", payload]`). `runCommand()` forwarded that argv to `runCommandWithTimeout()` without `windowsVerbatimArguments`, so Node applied its own per-argument quoting to a payload that is already a full command line.

The core exec path already resolves `windowsVerbatimArguments` through `resolveSafeChildProcessInvocation()`; only this node-host caller was missing it. `runCommandWithTimeout()`'s options already accept and forward the flag, so this is a caller-side change in one file plus its tests.

`prepareWindowsCmdInvocation()` recognises the canonical five-element invocation — `cmd.exe` or the normalized `%SystemRoot%\System32\cmd.exe`, case-insensitive switches — wraps the payload in one extra quote pair, and sets `windowsVerbatimArguments`.

**The wrap is required, not belt-and-braces.** With `/s`, `cmd` strips the first and last quote of the whole command line, so a payload that already begins and ends with a quote loses them when only the verbatim flag is set. That case fails both today and with the flag alone; only wrap-plus-verbatim survives it. The Evidence section shows the measurement.

**Non-goals / boundaries.** Everything that is not the canonical five-element `cmd.exe` invocation keeps its current behavior and gets the caller's argv array back untouched: PowerShell, direct executables, other `cmd` paths, non-canonical argv, and non-Windows hosts. Environment, cwd, cancellation, timeout, process-tree termination, output limits, execution policy and approval handling are unchanged. The payload's own content is never rewritten.

One note for reviewers: `src/node-host/invoke.ts` already carries an `oxlint-disable max-lines` for being a grandfathered oversized file, and this adds roughly 50 lines to it. Happy to move the helper to its own module if you would prefer that.

## User Impact

Operators can run a native `exec` on a Windows node host with quoted arguments and have the command arrive exactly as written. Anything that previously worked keeps working — the change is inert for every invocation shape other than the canonical `cmd.exe /d /s /c` transport, and inert on non-Windows hosts.

## Evidence

**Measured on Windows 11, Node v24.19.0**, spawning `cmd.exe` directly with `child_process.spawnSync` and comparing three configurations of the identical argv. **A** = current behavior, **B** = verbatim flag only, **C** = this change (wrap + verbatim):

| # | payload sent | A — today | B — verbatim only | C — this change |
|---|---|---|---|---|
| 1 | `echo "https://tenant.example.com/"` | `\"https://tenant.example.com/\"` | correct | correct |
| 2 | `echo "https://x.example.com/?a=1&b=2"` | backslashes added | correct | correct |
| 3 | `echo "C:\Program Files\demo\tool.exe" --flag` | backslashes added | correct | correct |
| 4 | `echo "C:\dir\" end` | `\"C:\dir\\\" end` | correct | correct |
| 5 | `"C:\Windows\System32\cmd.exe" /c echo "tail"` | **fails:** `'\"C:\Windows\System32\cmd.exe\"' is not recognized as an internal or external command` | **fails:** `The directory name is invalid.` | `"tail"` |

Row 5 is why the payload is wrapped rather than only flagged.

**Regression tests at the runCommand boundary** — `src/node-host/invoke.windows-cmd-quoting.test.ts`, 18 cases driving the real `testing.runCommand` and asserting what reaches the shared runner: the wrapped argv and `windowsVerbatimArguments: true` for the canonical shapes (bare executable, upper-case switches, backslash and forward-slash `SystemRoot` paths, `WINDIR` fallback, unusable Windows root), the untouched **same array instance** and absent option for everything else (a `cmd.exe` outside the Windows root, PowerShell, a direct executable, four- and six-element argv, reordered switches, a missing switch, a non-string entry, `darwin`, `linux`), plus the surrounding run options (`baseEnv`, `cwd`, `killProcessTree`, `timeoutMs`) surviving unchanged.

**These fail if the wiring is reverted.** Restoring the call site to its previous form (`runCommandWithTimeout(argv, { baseEnv: env, …})`) makes **7 of them fail**, which is the point: a helper-only suite would have stayed green while the reported bug returned.

**Real Windows output** — two cases added to the existing `src/node-host/invoke.run-command.test.ts`, gated with `describe.runIf(process.platform === "win32")` in the same style as the file's existing platform-gated tests. They run real `cmd.exe` and assert the child's stdout for a quoted URL and for a payload that itself begins and ends with a quote. They skip on non-Windows hosts.

No helper-only test API is added — `src/node-host/invoke.test-support.ts` is byte-identical to `main`.

```
node scripts/run-vitest.mjs src/node-host/invoke.windows-cmd-quoting.test.ts src/node-host/invoke.run-command.test.ts
  Test Files  2 passed (2)
       Tests  30 passed | 2 skipped (32)     # the 2 skipped are the Windows-only cases

pnpm tsgo:core        -> exit 0
pnpm tsgo:core:test   -> exit 0
npx oxlint <changed files>  -> clean
oxfmt --write         -> applied
```

**Production soak.** We have been running this change on Windows node hosts since 2026-09-08 (OpenClaw 2026.9.1 and 2026.9.2). End-to-end gateway-to-node native `exec` after a node-host restart: quoted URLs, spaces and ampersands arrive with no characters added or lost, and the browser flow that originally failed now opens the correct page. No regression observed in non-Windows nodes, PowerShell invocations or direct-executable invocations.

### DIFF
--- a/src/node-host/invoke.run-command.test.ts
+++ b/src/node-host/invoke.run-command.test.ts
@@ -185,4 +185,33 @@ describe("runCommand", () => {
       expect(testing.clarifyNodeExecCwdSpawnError(enoent(message), "/unreadable")).toBe(message);
     });
   });
+
+  // Runs the real cmd.exe, so it only reports on a Windows host. The
+  // cross-platform wiring assertions live in invoke.windows-cmd-quoting.test.ts.
+  describe.runIf(process.platform === "win32")("Windows cmd payload quoting", () => {
+    it("delivers a quoted argument to the child without added escapes", async () => {
+      const url = "https://tenant.example.com/";
+      const result = await testing.runCommand(
+        ["cmd.exe", "/d", "/s", "/c", `echo "${url}"`],
+        undefined,
+        undefined,
+        undefined,
+      );
+
+      expect(result.success).toBe(true);
+      expect(result.stdout.trim()).toBe(`"${url}"`);
+    });
+
+    it("delivers a payload that itself begins and ends with a quote", async () => {
+      const result = await testing.runCommand(
+        ["cmd.exe", "/d", "/s", "/c", '"C:\\Windows\\System32\\cmd.exe" /c echo "tail"'],
+        undefined,
+        undefined,
+        undefined,
+      );
+
+      expect(result.success).toBe(true);
+      expect(result.stdout.trim()).toBe('"tail"');
+    });
+  });
 });
--- a/src/node-host/invoke.ts
+++ b/src/node-host/invoke.ts
@@ -341,6 +341,56 @@ function clarifyNodeExecCwdSpawnError(
   return `node exec working directory ${reason} on the node host: ${cwd} (os reported: ${message})`;
 }
 
+const WINDOWS_CMD_SWITCHES = ["/d", "/s", "/c"] as const;
+
+/**
+ * The gateway hands the node host a complete `cmd` command line as a single argv
+ * entry (`["cmd.exe", "/d", "/s", "/c", payload]`). Without
+ * `windowsVerbatimArguments` Node applies its own per-argument quoting to that
+ * payload, which inserts backslashes before every quote it contains, so the
+ * command that runs is not the command that was sent.
+ *
+ * The payload is wrapped in one extra pair of quotes as well: with `/s`, cmd
+ * strips the first and last quote of the whole command line, so a payload that
+ * already begins and ends with a quote loses them when only the verbatim flag
+ * is set. Wrapping gives `/s` a pair to consume and leaves the payload intact.
+ *
+ * Anything that is not the canonical five-element `cmd.exe` invocation keeps its
+ * current behavior: PowerShell, direct executables, other `cmd` paths,
+ * non-canonical argv and non-Windows hosts.
+ */
+function prepareWindowsCmdInvocation(argv: string[]): {
+  argv: string[];
+  windowsVerbatimArguments?: boolean;
+} {
+  if (process.platform !== "win32" || argv.length !== WINDOWS_CMD_SWITCHES.length + 2) {
+    return { argv };
+  }
+  if (!argv.every((entry) => typeof entry === "string")) {
+    return { argv };
+  }
+  const switchesMatch = WINDOWS_CMD_SWITCHES.every(
+    (expected, index) => argv[index + 1]?.toLowerCase() === expected,
+  );
+  if (!switchesMatch) {
+    return { argv };
+  }
+  const executable = argv[0] ?? "";
+  const windowsRoot = process.env.SystemRoot || process.env.WINDIR || "C:\\Windows";
+  const trustedAbsoluteCmd =
+    /^[A-Za-z]:[\\/]/.test(windowsRoot) &&
+    path.win32.normalize(executable).toLowerCase() ===
+      path.win32.join(windowsRoot, "System32", "cmd.exe").toLowerCase();
+  if (executable.toLowerCase() !== "cmd.exe" && !trustedAbsoluteCmd) {
+    return { argv };
+  }
+  const payload = argv[WINDOWS_CMD_SWITCHES.length + 1] ?? "";
+  return {
+    argv: [...argv.slice(0, WINDOWS_CMD_SWITCHES.length + 1), `"${payload}"`],
+    windowsVerbatimArguments: true,
+  };
+}
+
 async function runCommand(
   argv: string[],
   cwd: string | undefined,
@@ -349,7 +399,9 @@ async function runCommand(
   signal?: AbortSignal,
 ): Promise<RunResult> {
   try {
-    const result = await runCommandWithTimeout(argv, {
+    const invocation = prepareWindowsCmdInvocation(argv);
+    const result = await runCommandWithTimeout(invocation.argv, {
+      ...(invocation.windowsVerbatimArguments === true ? { windowsVerbatimArguments: true } : {}),
       baseEnv: env,
       cwd,
       killProcessTree: true,
--- a/src/node-host/invoke.windows-cmd-quoting.test.ts
+++ b/src/node-host/invoke.windows-cmd-quoting.test.ts
@@ -1,0 +1,178 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const { runCommandWithTimeoutMock } = vi.hoisted(() => ({
+  runCommandWithTimeoutMock: vi.fn(),
+}));
+
+vi.mock("../process/exec.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../process/exec.js")>()),
+  runCommandWithTimeout: runCommandWithTimeoutMock,
+}));
+
+const { testing } = await import("./invoke.test-support.js");
+
+const PAYLOAD = 'echo "https://tenant.example.com/"';
+const SWITCHES = ["/d", "/s", "/c"] as const;
+
+const originalPlatform = process.platform;
+const originalSystemRoot = process.env.SystemRoot;
+const originalWindir = process.env.WINDIR;
+
+function setPlatform(platform: NodeJS.Platform): void {
+  Object.defineProperty(process, "platform", { value: platform, configurable: true });
+}
+
+/** Drives the real runCommand wiring and returns what reached the shared runner. */
+async function captureRunnerCall(
+  argv: string[],
+): Promise<{ argv: string[]; options: Record<string, unknown> }> {
+  await testing.runCommand(argv, undefined, undefined, undefined);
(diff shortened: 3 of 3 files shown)


igs-rogenlo:
Both review items are addressed in `63ca9e09`:

- The quoting regressions now assert at the `testing.runCommand` boundary instead of on the helper, so reverting the call-site wiring or dropping the forwarded `windowsVerbatimArguments` fails them — with the call site restored to its previous form, 7 of them fail. Real `cmd.exe` output is additionally verified on Windows hosts via `describe.runIf(process.platform === "win32")` in the existing `invoke.run-command.test.ts`.
- The helper-only test API is gone; `src/node-host/invoke.test-support.ts` is byte-identical to `main`.

PR body updated with the new evidence. CI is green (131 passed).

@clawsweeper re-review

Issue #52, opened 2025-12-25
silent restarts


fredheir:
I LOVE working with Clawdis (thinking to invest in more phones lol)
A point of pain in practice ends up being agents restarting the clawdis service, or the service failing silently,  restarting, and losing context. What about having clawdis dm a 'system restarted' message on launch? Would you welcome a pr for this?

steipete:
Thank you! There's a new tool I'm experimentingt that allows Clawd to restart gateway itself.

And when you dev, check the attach only flag in debug on mac and run gateway via pnpm gateway:watch so it auto-restarts on file changes.

Plus whatsapp can now be re-registered using the dashboard website.

Open for more ideas tho!

PR #143095, opened 2026-09-09
feat(vctraderai): render_card — the agent asks for a card, the platform composes it

ebakoachere:
The propfirm_manager BFF has been able to compose an in-chat card since its PR #1834, and had **never once done it in production** — because nothing could call it. On the rolled workspace the founder asked for a chart of XAU_USD and the agent wrote code, executed it, and returned a **PNG image: 8.4 seconds, zero composer hits.** Exactly what "no tool" predicts.

This is the tool.

## One plugin, one `kind` enum

Not one tool per card: one plugin, one allowlist row, one parity entry, **one bake**, and one description to keep honest against twenty card shapes. A single tool with an enum is also what a model calls most reliably, and the platform's slash commands bind to it with the kind fixed.

## The result is ONE LINE, and that is a contract

A tool result is replayed into the model's context on **every later turn of the thread**, exactly like the model's own text. Returning the composed card here would put the whole payload back into the conversation and undo the only reason the composer exists — and the numbers would be re-read by the model rather than rendered from the store.

Every other `vctraderai` tool returns the BFF's verbatim envelope. **This one deliberately does not**, so the deviation is stated in the plugin's own description, where the next editor will read it, and pinned by tests twice over:

- the result is a single line;
- the card does not leak **even if the BFF sends one** (defence in depth — the route is specified to return only `{ok, kind, confirmation}`, but a later change on that side must not silently widen what the model receives).

## The kind list is pinned by name

The source of truth is `web_api/agent_alpha/artifact_schema.py::_TYPES` in the other repository and cannot be imported, so the pin is what makes a drift loud on this side. If the two disagree, the model is offered a kind the platform cannot compose — or denied one it can — and the only symptom is a refusal the user never understands.

**19 kinds. `approval` is excluded deliberately:** the platform renders the governed approval card from the staged action itself, so offering the kind here would invite a call that can only be refused, and a second unauthoritative copy of a governed object is exactly what the persona forbids.

## The client learns to POST

`render_card` is the first `vctraderai` tool that sends a body — its `params` are a nested object (indicators, annotations) that cannot ride in a query string, and a silent fallback to one would truncate the intent and compose a card from less than was asked for. The BFF client is shipped **per plugin** by the extensions boundary, so this touches no other tool.

The egress allowlist tests carry over, plus two new ones: that the params travel as a JSON body with a content type, and that a missing `PFM_WORKSPACE_ID` refuses rather than calling an unscoped path.

## Provenance

Cut from the **deployed tag `vctraderai-v1.42.7-oc84c8a2b4133b`**, verified with `git describe --tags --exact-match` before the first edit — never the fork's main, because the deployed lineage is the only correct bake base.

**16 tests pass** across the two files (run, not merely written — the workspace install was needed to run them).

## Sequencing

The next bake bundles this plugin with fork **#52** and **#54** and the `get_fresh_candles` description fix (the payload's `source`/`age_seconds` are authoritative). Per DEC-67 the order is: bake first, move the tag pins, **then** enable — the propfirm_manager side already declares this tool dark with the reason and the removal condition, so enabling before the bake would be the silent no-op that rule exists to prevent.

pushes: 1


ebakoachere:
Opened in error against the upstream repository by a mistaken --head fork-qualified ref; this belongs on the ebakoachere/openclaw fork only. Closing immediately.

PR #143098, opened 2026-09-09
fix(fleet): prevent removal of a container that reuses a cell name

### LINKED ISSUE #143073: [Bug]: fleet stops and force-removes containers it did not verify
### Bug type

### Beta release blocker

No

### Summary

Fleet proves a cell container's ownership by inspecting it by name, then performs the stop/start/restart/remove that follows on the container **name** rather than on the container it just validated, so `fleet rm --force --purge-data` can destroy a container that never passed the ownership guard.

### Steps to reproduce

Observed directly against `src/fleet` on `main` (`34d7970e82a`) by driving `createFleetService` through the real `FleetContainerRuntime` seam, with the runtime stubbed to resolve a reference the way Docker and Podman do — an id stays pinned to one container, a name follows whatever holds it now.

1. Create a fleet cell (`openclaw fleet create acme`).
2. Let the managed container stop and disappear, and let an unrelated container take the name `openclaw-cell-acme` — the exact case `assertManagedInspection` exists to refuse ("Refusing to manage …: fleet ownership labels do not match tenant").
3. Arrange for that takeover to land after `remove()` inspects and validates the managed generation but before it issues the removal.
4. Run `openclaw fleet rm acme --force`.

The same window exists on `fleet stop/start/restart`, `fleet upgrade`, `fleet restore`, and on the create/upgrade recovery paths.

### Expected behavior

The action targets the container the ownership guard proved, and refuses when that container is no longer there.

This is not a new expectation: `fleet logs` already behaves this way. #106918 pinned it to the inspected container with the comment *"Pin the inspected generation so a concurrent restore cannot redirect the stream."* Five of the affected call sites go further and compare `openclaw.fleet.attempt` — they already throw `"container generation changed during upgrade recovery"` — and then discard the generation they just proved. `src/fleet/backup.runtime.ts` states the expectation in prose, *"Restart the same managed generation"*, five lines above a name-based `start`.

### Actual behavior

`fleet rm --force` **succeeds**, having force-removed the container that currently holds the name, and then deletes the tenant registration as though the removal were legitimate. Observed as an ownership-validated removal reaching an unvalidated container:

```
AssertionError: promise resolved "{ tenant: 'acme', action: 'rm', …(1) }" instead of rejecting
- Expected
+ Received
- Error { "message": "rejected promise",
+ { "action": "rm", "dataPurged": false, "tenant": "acme" }
```

With `--purge-data`, the cell data directory and the auth-profile secrets are deleted alongside it.

### OpenClaw version

2026.9.3 (source checkout at `34d7970e82a`)

### Operating system

Ubuntu 22.04.5 LTS (Linux 6.8.0)

### Install method

pnpm dev (source checkout)

### Model

Not applicable — the defect is in fleet container lifecycle management and is independent of model routing.

### Provider / routing chain

Not applicable — no model request is involved.

### Logs

```shell
# src/fleet/service.runtime.ts, remove()  (main @ 34d7970e82a)
const inspection = await containers.inspect(record.runtime, record.containerName);
...
assertManagedInspection(record, inspection);          # proves generation A
checkpoint();
await containers.remove(record.runtime, record.containerName, params.force === true);
                                        # ^ acts on whatever holds the name now

# the sibling that already does it correctly, since #106918:
await containers.logs(record.runtime, inspection.containerId, { ... });
```

Affected call sites on `34d7970e82a`:
`src/fleet/service.runtime.ts:529, 611, 614, 778`
`src/fleet/service-support.runtime.ts:522, 623, 634`
`src/fleet/backup.runtime.ts:682, 690, 753, 791`

Focused before/after at the `FleetContainerRuntime` seam, using the same mock the #106918 regression test uses. Full fleet suite is green after the fix (215/215). A PR follows.

### Impact and severity

Affected: operators running `openclaw fleet` cells.
Severity: data risk — the worst case is a forced container removal plus `--purge-data` deletion of the cell data directory and auth-profile secrets for a container that failed no check because it was never checked.
Frequency: edge case — it needs the container to be replaced inside the window between the ownership check and the action. `withFleetCellOperation` already serializes fleet's own operations per tenant, so the replacement has to come from outside that lease.
Consequence: a guard that exists specifically to refuse a foreign name-squatting container does not protect the operations that can destroy one.

### Additional information

The same reachability was already accepted upstream for `fleet logs` in #106918, where the consequence was a redirected log stream. Here the consequence is a destroyed container.

masatohoshino:
Most changed lines move existing tests; the container fix adds 14 lines across three modules.

Closes #143073

## What Problem This Solves
Fixes an issue where `fleet rm --force` could delete a foreign container that took a managed cell's name after inspection.

## Why This Change Was Made
Lifecycle actions use the inspected container ID. The other writer of container state is Docker or Podman; immutable IDs keep external name changes from redirecting an authorized action.

Growth: service +4, support +7, backup +3 lines retain the existing ownership checks and explain identity pinning. Growth accepted under the owner's standing acceptance (2026-09-15).

## User Impact
Removal affects only the inspected container. Stop and restore recovery use the same rule.

## Evidence
Real Docker, through the registered commands:
- `openclaw fleet rm <tenant> --force`: before, foreign B was deleted and renamed A survived; after, A was removed and B survived when read by its own ID.
- `openclaw fleet stop <tenant>`: after, A stopped and B stayed running.
- `openclaw fleet backup <tenant> --out <archive>` succeeded. Restore preserved B in running and stopped cases, then reported the expected name conflict.

## Compatibility
Ordinary removal is unchanged. Missing-ID removal fails unforced and is a forced no-op. Initial create/start and explicit data purge are unchanged. Podman was not exercised.

## Consumers
Managed lifecycle, upgrade cleanup, removal, and restore use the existing runtime ID contract.

## Invalidation
IDs last for one operation; no new cache or stored state.

## Tests
219 Fleet tests and 23 CLI tests pass on a fresh main merge. All 279 original assertions remain. Each changed suite is selected once; the test partition uses 649 of 720 roots. Type, lint, architecture, and generated-file checks pass.


PollyBot13:
Two suggestions for the author/maintainers to consider, either here or as focused follow-ups. These are scope/policy choices, not objections to the inspected-ID repair in this PR.

1. **Extend identity pinning to create → start.** The body explicitly excludes this path because it does not yet hold an inspection. Could creation retain the newly created container's immutable identity, validate the expected ownership/attempt, and start that exact container rather than re-resolving its name? Preserve the reservation/lease checks across any added await. This would extend the same protection to initial startup, at the cost of another validation/failure path. A regression should swap the name between creation and startup and verify that the replacement is never started. A separate small follow-up seems reasonable if adding this would delay the current fix.

2. **Make missing-container cleanup versus data purge an explicit contract.** Accepting `rm --force <missing-id>` as success is reasonable for idempotent container removal. Could the `--purge-data` case be documented and tested separately when the validated container disappears between inspection and removal? One option is to preserve data/auth and sufficient registration for recovery on that unexpected disappearance, with an explicit retry/recovery path; another is to explicitly define the existing `--purge-data` request as authorizing cleanup even when removal becomes a no-op. The former favors preservation but adds retry friction; the latter favors predictable cleanup automation. A zero exit from forced removal alone does not distinguish those outcomes. Tests should cover both the post-inspection disappearance and the initially-missing retry case, so the policy is consistent and recoverable.

My preference is to pursue create-start pinning and explicitly separate container-removal success from the destructive purge decision, while keeping this P0 repair focused. Neither suggestion needs a competing implementation of the lifecycle fix.

obviyus:
landpr-approve-admin-merge
pr: 143098
head: 13ab8bf80f770ce5d561bc9b6558963bf00b6fe6
scope: owner-review

obviyus:
landpr-approve-admin-merge
pr: 143098
head: bdfb7e2bf4ee9436b2c11ef7cf50aff09e6ad480
scope: owner-review

obviyus:
Thanks @masatohoshino!

**Proof:** Real Docker commands verified that removal affects the inspected container and preserves a foreign container that later takes its name.

- **Bug:** A managed cell removal could delete an unrelated replacement container.
- **Fix:** Fleet uses the inspected container ID for lifecycle actions and recovery.

| Before | After |
| --- | --- |
| Forced removal deleted foreign B and left inspected A alive. | Forced removal deleted A and left B alive. |

PR #143102, opened 2026-09-09
fix(ci): reduce Gateway startup cost and preserve Cron test failures

vincentkoc:
## What Problem This Solves

Resolves two test-fixture ownership problems investigated during maintainer-requested repair of intermittent main CI failures: the real Gateway fixture spends its readiness budget transforming source, and failed Cron tests can abort the heartbeat they still need to drain, replacing the original failure with a cleanup error.

Related CI evidence:
- Gateway readiness timeout: https://github.com/openclaw/openclaw/actions/runs/34338882869/job/102424948619
- Cron cleanup failure: https://github.com/openclaw/openclaw/actions/runs/34340643015/job/102431987202

## Why This Change Was Made

The Gateway fixture now participates in the existing invocation-owned compiled-worker preparation flow. Standalone source execution retains its original ESM-only loader; prepared JavaScript executes without a TypeScript loader.

Cron fixture teardown stops new admission, drains active jobs and the Cron lane while the heartbeat owner remains available, and always disposes that owner afterward. Body results and errors are captured before cleanup; failures propagate after cleanup rather than from a `finally` block. When both the test body and cleanup fail, both errors are retained in their original order, with the primary failure as the cause. Three controlled regressions cover early teardown, unconditional disposal after a drain failure, and preservation of the primary failure.

## User Impact

No production runtime, configuration, dependency, or public API changes. Maintainers get less Gateway startup work inside the readiness hook and actionable Cron test failures instead of teardown masking. Existing deadlines and assertions remain unchanged; no retries or new global resets were added.

## Evidence

Validated the four-file candidate against frozen base `056d14028ff128218ab98e1558832ba627bad146` on Linux with Node 24.19.0, pinned pnpm 12.3.4, Vitest 5.0.0, and oxfmt 0.65.0. Remote receipts verified the source tree, tracked lockfile, patch, and installed tool versions. The final lint-corrected tree is `7f2bbd6624b0be82f6ad6cdaf469b029fc8b98da`.

- Controlled Cron fault: the new early-teardown regression failed with the original ordering because the real heartbeat signal was aborted during drain. It passed after the repair, with active-job and Cron-lane counts both zero.
- Cron focused: 17 tests passed, including the original 14 cases and three new regressions.
- Cron original-order replay after the lint correction: all 102 service test files and 1,350 tests passed in 110.99 seconds. Actual execution order was checked against frozen baseline file sizes; the affected file ran 31st, after `service.cross-tick-admission.test.ts`, under the normal non-isolated single-worker configuration.
- Gateway standalone compiled fixture: 3 tests passed. Original-order update-lease/Gateway pair: 34 tests passed. Source fallback: 3 tests passed. Runtime resolver coverage: 9 tests passed. Gateway sibling files: 13 tests passed.
- Same-lease source/compiled comparison: child readiness was 57.1 seconds with the original source loader and 4.31 seconds with the prepared JavaScript fixture. Compiled preparation ran before the readiness hook. This is a measured sample, not a universal performance guarantee.
- Ready-state CI found seven lint violations in the added cleanup flow: https://github.com/openclaw/openclaw/actions/runs/34352615901/job/102469568895. The follow-up preserves results and both errors outside `finally`, supplies the original cause, and fixes the shadowed catch parameter. Targeted type-aware oxlint 1.80.0 with oxlint-tsgolint 7.0.2001 passed with zero warnings or errors; the focused and original-order Cron proofs above were repeated on the corrected tree.
- Targeted formatting and `git diff --check` passed. Production runtime LOC delta is zero.
- Fresh scoped autoreview completed with no actionable P0-P2 findings.

The historical intermittent 120-second Gateway and initial 10-second Cron triggers were not reproduced deterministically. This change repairs the demonstrated fixture preparation and teardown defects; it does not claim every historical timeout has been fully explained. Hosted exact-head CI and review remain required before landing.

### DIFF
--- a/scripts/lib/vitest-worker-build-entries.mts
+++ b/scripts/lib/vitest-worker-build-entries.mts
@@ -8,6 +8,7 @@ import { cliCompactionBackendEntrypoints } from "../../src/agents/command/cli-co
 import {
   cliRecoveryEntrypoints,
   gatewayDirectStopEntrypoints,
+  stateDirGatewayFixtureEntrypoint,
 } from "../../src/cli/cli-entrypoint.test-support.ts";
 import { doctorConfigRuntimeEntrypoints } from "../../src/commands/doctor-config-runtime.test-support.ts";
 import { cronOwnerHardeningEntrypoints } from "../../src/cron/owner-hardening-runtime.test-support.ts";
@@ -42,6 +43,7 @@ export const vitestWorkerBuildEntries = {
       ...publishedSdkBridgeEntrypoints,
       ...Object.values(cliRecoveryEntrypoints),
       ...Object.values(gatewayDirectStopEntrypoints),
+      stateDirGatewayFixtureEntrypoint,
       ...Object.values(doctorConfigRuntimeEntrypoints),
       ...Object.values(cronOwnerHardeningEntrypoints),
       ...Object.values(tuiPtyRuntimeEntrypoints),
--- a/src/cli/cli-entrypoint.test-support.ts
+++ b/src/cli/cli-entrypoint.test-support.ts
@@ -17,6 +17,14 @@ export const cliRecoveryEntrypoints = {
   },
 } as const;
 
+// Prepare the real Gateway fixture before its readiness hook starts; source
+// transforms must not consume that hook's startup deadline.
+export const stateDirGatewayFixtureEntrypoint = {
+  currentModuleUrl: import.meta.url,
+  sourceWorkerName: "state-dir-gateway-check.server-fixture.test-support",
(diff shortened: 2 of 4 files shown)


vincentkoc:
@clawsweeper re-review

Updated head: `0a6dfb11cd2b8df866a89e2cd1c0e2e208ad0eda`.

The follow-up fixes the CI lint errors in Cron fixture cleanup without changing deadlines or assertions. Type-aware lint and formatting passed; the 17-case fixture and all 1,350 tests across the original-order 102-file Cron shard passed. The PR body contains the updated proof and preserved timeout limitations.

PR #143103, opened 2026-09-09
fix(ui): prevent animation races in attachment layout checks

steipete:
## What Problem This Solves

Resolves an intermittent CI failure in the mobile attachment layout test. Main run [34348886307](https://github.com/openclaw/openclaw/actions/runs/34348886307/job/102457720621) compared the preview and its remove button at different moments, reporting a 0.134px clipping failure at 393×852.

## Why This Change Was Made

Read the related rectangles, text origin, and top-edge hit targets in one synchronous browser evaluation, following the neighboring composer tests. The shell and card entrance animations move all descendants together; separate browser calls could compare different animation frames. This reduces six measurement calls to one while retaining the original assertions, selector uniqueness, finite rectangles, focus behavior, and scrolled right-edge hit tests.

## User Impact

More reliable layout CI with fewer browser round trips. Product rendering, animation behavior, shard counts, timeouts, and geometry tolerances are unchanged. Production delta is zero; only the existing test changes.

## Evidence

- Controlled Chromium reproduction: pause the existing ancestor animations at 0ms, read the preview, then advance them to 250ms before the later measurements. The original assertion fails with remove y=666.007385 versus earlier preview y=679. At either individual frame the remove button remains 7px inside the preview; ancestor/input/rail scroll positions remain zero. This proves the frame-crossing failure mode, not the precise animation phase of the CI failure.
- Original focused four cases and full 127-case browser file pass locally; the failure is intermittent. The candidate also passes all 127 cases, covering all four original attachment viewports and neighboring composer interactions. Temporary diagnostics were removed before candidate validation.
- Whole-file local durations varied from 28.10s to 38.91s, so no wall-time speedup is claimed from this pair.
- Managed independent review through P2: no actionable findings. Required changed-file checks passed after correcting local variable shadowing.

Command: `node scripts/run-vitest.mjs run ui/src/pages/chat/chat-responsive.browser.test.ts` (Node 24, pnpm 12).

Exact-head CI [34352752812](https://github.com/openclaw/openclaw/actions/runs/34352752812) passed, including all 127 changed-file cases in Linux (31.98s; attachment cases 278–305ms). One failed-job rerun was needed for an unchanged catalog-reopen E2E test. Its retained failure showed the picker closed and only two catalog requests; an unchanged local 11-case run and one instrumented replay passed without establishing the cause. The same-head retry passed all 70 files/322 tests in that shard (707.82s; catalog case 2.064s). That separate toggle race remains a follow-up; this PR does not claim to fix it.

### DIFF
--- a/ui/src/pages/chat/chat-responsive.browser.test.ts
+++ b/ui/src/pages/chat/chat-responsive.browser.test.ts
@@ -4338,27 +4338,49 @@ describeBrowserLayout.concurrent("chat responsive browser layout", () => {
         await page.locator(".agent-chat__input").evaluate((node) => {
           node.scrollTop = 0;
         });
-        const input = await getBoundingBox(
-          page,
-          ".agent-chat__composer-shell > .agent-chat__input",
+        // Ancestor entrance animations move these boxes together; compare one frame.
+        const { input, preview, attachment, remove, topHits, textStart } = await page.evaluate(
+          () => {
+            const elementFor = (selector: string) => {
+              const [element, ...others] = document.querySelectorAll(selector);
+              if (!element || others.length > 0) {
+                throw new Error(`Expected one layout element: ${selector}`);
+              }
+              return element;
+            };
+            const rectFor = (selector: string) => {
+              const {
+                x,
+                y,
+                width: rectWidth,
+                height: rectHeight,
+              } = elementFor(selector).getBoundingClientRect();
+              return { x, y, width: rectWidth, height: rectHeight };
+            };
+            const removeNode = elementFor(".chat-attachment-remove");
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [688fb10d2e3beb18d55227d3997c7681eb4064f7](https://github.com/openclaw/openclaw/pull/143103/commits/688fb10d2e3beb18d55227d3997c7681eb4064f7)
- Landed commit: [8642c68b6b9f6ee42fa91641dd8656fcf0a617a9](https://github.com/openclaw/openclaw/commit/8642c68b6b9f6ee42fa91641dd8656fcf0a617a9)

PR #143109, opened 2026-09-09
fix: close plugin resources after read-only model checks

steipete:
## What Problem This Solves

Resolves a problem where read-only model checks, effective tool inventory, and isolated model probes could leave plugin connections open after finishing.

## Why This Change Was Made

Temporary prepared runtimes now own their registrations through the final borrower and actual callback completion, including cancellation and shutdown. Copied registry views retire together, and SDK borrowers retain the original physical resources without preserving retired authority. Ordinary run/configured owners and raw loader caching retain their existing behavior.

## User Impact

Repeated inspections release temporary plugin resources safely. In-flight preparation keeps its connections until it settles, and metadata-only catalog reads remain runtime-free.

## Evidence

- Native regressions reproduced the leaked registration, copied-view authority, SDK custody, and reentrant retirement failures before their fixes. All 66 affected native controls pass, including both registry test-file orders.
- Post-rebase composition: 102 tests passed across prepared runtime, inventory, SDK, and registry boundaries. Production source is unchanged from the fully checked, independently reviewed, and built candidate.
- Full changed-file gate, isolated review through P2, and ordinary build passed; earlier prepared owner/catalog/worker siblings passed 201 tests.
- Actual compiled CLI/Gateway proof: metadata loaded no plugin runtime; warm inventory acquired nothing new; two cold requests each closed their own SQLite registration once, then reopened the same file read-only with its persisted marker intact. Raw startup registrations stayed outside ephemeral disposal, and strict process-tree shutdown completed cleanly in 163 ms.

The first hosted architecture job found a type-only import cycle missed by the selected local gate. Moving the claim declaration into the existing types module removed it: the original full Madge check failed, the corrected full check passed, all three emitted JavaScript outputs are unchanged, and the follow-up changed-file gate and fresh independent review passed.


steipete:
Merged via squash.

- Prepared head SHA: [b803ce82d8bb589bab03fe4fdee87a06202ade9c](https://github.com/openclaw/openclaw/pull/143109/commits/b803ce82d8bb589bab03fe4fdee87a06202ade9c)
- Landed commit: [73b4086ff658973fb9ae07fa208c49ca8de82a67](https://github.com/openclaw/openclaw/commit/73b4086ff658973fb9ae07fa208c49ca8de82a67)

Issue #83510, opened 2026-05-18
[Bug]: Embedded agent failover treats session-file mutation as model failure and exhausts all fallbacks


jsompis:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

OpenClaw 2026.5.16-beta.6 can fail an embedded agent turn before replying when the active session JSONL changes while the embedded prompt lock is released. The failure is classified as a model/candidate failure, so the same local session-file mutation is retried across unrelated fallback models. This exhausts the fallback chain and surfaces as “All models failed”, even though the first two failures are not provider/model failures.

### Steps to reproduce

1. Use a long-lived embedded agent session with automatic context/maintenance activity enabled.
2. Trigger an embedded agent turn on the active main session.
3. While the embedded prompt lock is released, allow a local process such as context compaction, maintenance, memory sync, or another gateway write path to append/update the same session JSONL file.
4. Observe model fallback behavior for the turn.

### Expected behavior

A local session takeover/session-file mutation should be treated as a runtime coordination failure, not as a model failure.

The system should either:
- abort the current turn with a clear non-model error and avoid consuming fallback models, or
- restart/rebase the turn safely after re-reading the updated session state, if that is supported.

Fallback should be reserved for provider/model failures such as timeout, rate limit, auth, or provider runtime errors.

### Actual behavior

The embedded run fails before reply with All models failed. The same session-file mutation is counted against multiple model candidates:

- openai/gpt-5.5 fails with: session file changed while embedded prompt lock was released
- claude-bridge/claude-opus-4-7 fails with the same local session-file mutation
- google/gemini-3.1-pro-preview then times out

The user sees a generic model-chain failure even though the primary cause is local session state changing underneath the embedded run.

### OpenClaw version

2026.5.16-beta6

### Operating system

macOS 26.5

### Install method

npm global

### Model

openai/gpt-5.5

### Provider / routing chain

openclaw->openai/gpt-5.5

### Logs, screenshots, and evidence

```shell

```

### Impact and severity

Severity: High for interactive reliability.

Impact:
- User-facing turns fail before any assistant reply.
- Fallback models are wasted on a non-model local coordination error.
- The final error message misleads operators toward provider/model diagnosis instead of session lock/session mutation handling.
- Repeated occurrences make long-lived or maintenance-heavy sessions unreliable.
- This can mask the true cause when the last fallback happens to time out, because the final surfaced error is All models failed rather than a deterministic session takeover error.

[...]

PR #143115, opened 2026-09-09
fix(agents): name the last two silent model fallback chain stops

### LINKED ISSUE #141604: [Bug]: model fallback chain aborts on a non-final hop and logs nothing, skipping remaining candidates

von8794:
Closes #141604.

## What Problem This Solves

A configured model fallback chain can stop partway and say nothing about it. The operator sees a chain that did not run, remaining candidates never attempted, and no line anywhere explaining why.

#141658 named the chain stops inside `runFallbackCandidate`'s terminal branch. Two rethrow sites on current `main` are still silent:

- `src/agents/model-fallback-attempt.ts:286` rethrows a local coordination failure (session superseded, gateway admission, stale lifecycle, writer rebound) so that fallback cannot blame a model for something no model did. That behaviour is correct. It is also invisible.
- `src/agents/model-fallback-runner.ts:611` rethrows for the same reason under the #83510 guard, for a non-provider runtime coordination error or a transcript that cannot be continued.

Both skip every remaining configured candidate without emitting a decision or a chain-stop record. Same class of blind spot #141658 closed, at two sites it did not reach.

## What This Changes

Two new reasons, `local_runtime_coordination` and `transcript_not_continuable`, emitted from both sites through a single helper in `model-fallback-observation.ts`.

No control-flow change. Both sites still rethrow exactly as before, the same errors reach the same callers in the same order, and no candidate is newly attempted or newly skipped. The only difference is that the stop is now named.

## Evidence

- `src/agents/model-fallback.chain-stop.test.ts`, written failing first: **2 failed / 29 passed before** the source change, **31 passed after**. The two new cases assert the specific reason on the emitted record, not merely that something was emitted. Re-run on this branch immediately before opening: `31 passed (31)`.
- `pnpm check` exits **0**, including the max-lines ratchet. Both edited source files land one line under the cap, which is why the emit goes through one shared helper rather than being inlined at each site.
- `pnpm build` exits **0**.
- CI on this head: 160 checks passing. The one red test, `src/cli/doctor-output.process.test.ts > omits backup tips for Git-backed nested agent workspaces`, is a ~12s process test in the CLI doctor path, untouched by this diff, and it passes locally on this exact branch. Reads as flake rather than a regression; happy to be told otherwise.

## Note On The Base

This branch is based on `037e0059e` rather than the current tip. It is an ancestor of `main`, and the touched files have no conflicting commits since. Happy to rebase if you would rather review it on the tip.

### DIFF
--- a/src/agents/model-fallback-attempt.ts
+++ b/src/agents/model-fallback-attempt.ts
@@ -25,6 +25,7 @@ import { resolveAgentHarnessPolicy } from "./harness/policy.js";
 import { getRegisteredAgentHarness } from "./harness/registry.js";
 import { LiveSessionModelSwitchError } from "./live-model-switch-error.js";
 import {
+  logLocalChainStop,
   logModelFallbackChainStopped,
   logModelFallbackDecision,
   type ModelFallbackChainStopReason,
@@ -284,6 +285,10 @@ async function runFallbackCandidate<T>(params: {
       lane: params.attribution?.lane,
     });
     if (fallbackError.kind === "coordination") {
+      // A local coordination failure is not a provider failure, so the chain
+      // stops here by design. Name it: the remaining candidates are skipped and
+      // nothing else on this path says so.
+      logLocalChainStop(err, params, params.attribution);
       throw err;
     }
     return { ok: false, error: fallbackError.error };
--- a/src/agents/model-fallback-observation.ts
+++ b/src/agents/model-fallback-observation.ts
@@ -356,7 +356,32 @@ export type ModelFallbackChainStopReason =
   | "caller_signal_aborted"
   | "agent_run_direct_abort"
   | "agent_run_restart_abort"
-  | "terminal_abort_wrapper";
+  | "terminal_abort_wrapper"
+  | "local_runtime_coordination"
+  | "transcript_not_continuable";
+
+/**
+ * Names the local, non-provider conditions that stop a chain at a rethrow site.
+ *
+ * The predicate lives in `model-fallback-attempt.ts`, which already imports this
+ * module, so the caller passes the classification instead of this module
+ * importing it back and closing a cycle.
+ */
+export function logLocalChainStop(
+  error: unknown,
+  candidate: { provider: string; model: string },
+  attribution?: { sessionId?: string; lane?: string },
+  transcriptNotContinuable?: boolean,
+): void {
+  logModelFallbackChainStopped({
+    reason: transcriptNotContinuable ? "transcript_not_continuable" : "local_runtime_coordination",
+    provider: candidate.provider,
+    model: candidate.model,
+    sessionId: attribution?.sessionId,
+    lane: attribution?.lane,
+    error,
+  });
+}
 
 /** Record a local or terminal stop separately from provider-failure decisions. */
 export function logModelFallbackChainStopped(params: {
--- a/src/agents/model-fallback-runner.ts
+++ b/src/agents/model-fallback-runner.ts
@@ -68,6 +68,7 @@ import {
 } from "./model-fallback-cooldown.js";
 import {
   isModelFallbackDecisionLogEnabled,
+  logLocalChainStop,
   logModelFallbackDecision,
   type ModelFallbackDecisionParams,
 } from "./model-fallback-observation.js";
@@ -609,6 +610,7 @@ async function runWithModelFallbackInternal<T>(
     // the same local condition and surfacing a misleading "All models
     // failed" summary. See #83510.
     if (isNonProviderRuntimeCoordinationError(err) || isTranscriptNotContinuableError(err)) {
+      logLocalChainStop(err, candidate, runAttribution, isTranscriptNotContinuableError(err));
       throw err;
     }
     if (transientProbeProviderForAttempt) {
--- a/src/agents/model-fallback.chain-stop.test.ts
+++ b/src/agents/model-fallback.chain-stop.test.ts
@@ -9,6 +9,7 @@ import {
   it,
   vi,
 } from "vitest";
+import { TranscriptNotContinuableError } from "../../packages/agent-core/src/errors.js";
 import { createDeferred } from "../../test/helpers/promise.js";
 import { createSuiteLogPathTracker } from "../logging/log-test-helpers.js";
 import { resetLogger, setLoggerOverride } from "../logging/logger.js";
@@ -21,7 +22,10 @@ import { abortable } from "./embedded-agent-runner/run/abortable.js";
 import { resolveEmbeddedRunAttemptTerminalState } from "./embedded-agent-runner/run/terminal-outcome.js";
 import { resolveEmbeddedRunTerminalTimeout } from "./embedded-agent-runner/run/terminal-timeout.js";
 import { FailoverError } from "./failover-error.js";
-import { AgentHarnessPreflightError } from "./harness/errors.js";
+import {
+  AgentHarnessPreflightError,
+  AgentHarnessSessionSupersededError,
+} from "./harness/errors.js";
 import { type ModelFallbackStepHandler, runFallbackAttempt } from "./model-fallback-attempt.js";
 import { runWithImageModelFallback } from "./model-fallback-image.js";
 import { runWithModelFallback } from "./model-fallback-runner.js";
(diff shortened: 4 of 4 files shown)


von8794:
`ci-gate` is red on one shard only. Its log resolves to a single selected job:

```
checks-node-core-test-nondist-shard: failure (selected=true)
```

which is `checks-node-compact-small-29`, and the single failing case there is:

`src/cli/doctor-output.process.test.ts > Doctor report process output > omits backup tips for Git-backed nested agent workspaces`

That is a ~12s process test in the CLI doctor path. This branch touches four files, all under `src/agents/`, and nothing in `src/cli/` or the doctor path. I ran that case 5 times on this exact head before saying so:

```
run 1: Tests  1 passed | 7 skipped (8)
run 2: Tests  1 passed | 7 skipped (8)
run 3: Tests  1 passed | 7 skipped (8)
run 4: Tests  1 passed | 7 skipped (8)
run 5: Tests  1 passed | 7 skipped (8)
```

Every other check is green: 160 passing, and the `PR context and evidence` gate now passes after I added the missing section heading.

I do not have rerun rights on this repo. If someone with them re-runs the failed job, I expect it to go green. If it fails again I would rather treat that as a real signal than keep calling it flake, so please say so and I will dig into the doctor path properly.

von8794:
Follow-up on the one remaining red check, with the specific cause this time rather than a flake assertion.

`checks-node-compact-small-29` fails in `src/cli/doctor-output.process.test.ts > omits backup tips for Git-backed nested agent workspaces`:

```
AssertionError: expected Error: spawnSync ... to be undefined
  message: "spawnSync .../node ETIMEDOUT"
  errno:   -110
  code:    "ETIMEDOUT"
  spawnargs: [--import, tsx, src/entry.ts, doctor, --lint,
              --only, core/doctor/workspace-suggestions, --severity-min, info, --json, --no-color]
```

It is a subprocess spawn timing out, not an assertion about behaviour. The shard it ran on was starved:

- `[vitest] still running with no output for 30000ms`, then `60000ms`
- the immediately preceding test in the same file, `explains and preserves retained custom agent databases in preview and repair`, **passed in 102658 ms**
- the failing test hit its ceiling at 60313 ms

A sibling case in the same file taking 102 seconds to pass is the tell: at that pace a 60 s spawn budget is not a behavioural limit, it is the runner.

This PR changes four files, all in the model-fallback area:

```
src/agents/model-fallback-attempt.ts
src/agents/model-fallback-observation.ts
src/agents/model-fallback-runner.ts
src/agents/model-fallback.chain-stop.test.ts
```

Nothing imports or influences `core/doctor/workspace-suggestions`, and the change is logging on existing rethrow paths with no behavioural difference.

A rerun of that job should clear it. I cannot trigger one without write access. If it reproduces on a rerun I will treat it as real and investigate properly rather than attributing it to the runner a second time.

PR #143117, opened 2026-09-09
refactor: share provider auth-choice precedence cases

steipete:
Related: #139428

## What Problem This Solves

Two adjacent provider auth-choice tests repeat the same setup and assertions for origin precedence, making their differing inputs and expected winners harder to compare.

## Why This Change Was Made

Use an explicit typed row loop with separate input plugin/provider IDs, origins, and expected winner fields. Each test still allocates its own fixtures and checks the complete choice list, individual lookup, and onboarding flags. Direct `it(testCase.name)` registration preserves both original full names and their order.

## User Impact

No user-visible behavior change. Both bundled-over-workspace and configured-over-bundled precedence cases remain intact, with 23 fewer net test lines. No runtime improvement is claimed.

## Evidence

All 13 owner tests and five preferred-provider sibling tests passed before and after, with identical full case names and per-file order. Parser-aware expansion reproduced both original test bodies; allocation checks confirmed fresh input and expected graphs and a fresh mock resolver callback per case. Mapped changed checks and fresh independent review passed.

An initial `it.each` version passed its tests but truncated long case names in Vitest. It was replaced with direct-name loop registration, then the same suites, structural proof, checks, and review were rerun. No runner settings or production behavior changed.

### DIFF
--- a/src/plugins/provider-auth-choices.test.ts
+++ b/src/plugins/provider-auth-choices.test.ts
@@ -581,125 +581,102 @@ describe("provider auth choice manifest helpers", () => {
     ]);
   });
 
-  it("prefers bundled auth-choice handlers when choice IDs collide across origins", () => {
-    setManifestPlugins([
-      {
-        id: "evil-openai-hijack",
-        origin: "workspace",
-        providers: ["evil-openai"],
-        providerAuthChoices: [
-          {
-            provider: "evil-openai",
-            method: "api-key",
-            choiceId: "openai-api-key",
-            choiceLabel: "OpenAI API key",
-            optionKey: "openaiApiKey",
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [927f477512b593814a27eb1413bd0df82e126f21](https://github.com/openclaw/openclaw/pull/143117/commits/927f477512b593814a27eb1413bd0df82e126f21)
- Landed commit: [e7567735b316fa82f0fc5e16d0d3c732ee2e2b08](https://github.com/openclaw/openclaw/commit/e7567735b316fa82f0fc5e16d0d3c732ee2e2b08)

Issue #142715, opened 2026-09-09
Managed pnpm update leaves Gateway stopped after post-core handoff root mismatch


timetrvlr:
### Summary

A managed update from OpenClaw 2026.9.2 to 2026.9.3 completed the core package replacement but left the user-systemd Gateway stopped. Manual service start restored a healthy Gateway on 2026.9.3; no rollback or configuration change was required.

### Environment

- Debian/Linux
- Global pnpm installation (`pnpm 10.28.2`)
- Gateway managed by user systemd
- Node.js 26.1.0
- OpenClaw 2026.9.2 -> 2026.9.3

### Observed behavior

The updater stopped the Gateway successfully. The post-core update handoff then failed with:

```
Managed update handoff root mismatch
```

The updater classified recovery as unsafe (`runtime-verification-failed`) and intentionally left the Gateway stopped. `Restart=always` on the systemd unit did not restart it because the updater explicitly stopped the unit.

Starting the service manually afterward worked and the Gateway was healthy on the new version.

### Expected behavior

A verified managed update should preserve its installation ownership across pnpm package replacement and restore the Gateway. If recovery cannot complete, the updater should provide a visible recovery action that does not require manual service intervention.

### Actual behavior

The Gateway remained down until manually started.

### Reproducibility and evidence limits

I have not reproduced this on current `main`. The failure occurred during the production upgrade above. The historic journal no longer contains the concrete pre/post replacement path targets, so I cannot provide the canonical targets from the failed transition.

This report is therefore about the observed release outage, not a claim that ordinary symlink resolution is missing. The investigation likely needs to inspect whether the retained handoff `rootIdentity` survives a pnpm generation/package replacement correctly, including the pre/post canonical root transition.

PR #143137, opened 2026-09-09
fix(update): preserve post-core handoff across pnpm generations

steipete:
Refs #142715

## What Problem This Solves

Fixes an issue where managed pnpm updates can reject the post-core CLI after successfully replacing the package. The global `openclaw` link can retain its path while its canonical version directory changes; the next CLI then reports `Managed update handoff root mismatch` against the previous generation.

## Why This Change Was Made

The post-core launcher now gives its child a private copy of the existing handoff metadata bound to the activated package root. The helper's original metadata and recovery/lease identity remain unchanged, and the child still rejects a different installation. No recovery guard, state schema, configuration option, or protocol changes.

## User Impact

Post-core plugin convergence can continue across a legitimate pnpm generation replacement instead of failing the root check. The existing verified-recovery requirements remain in force.

## Evidence

The regression uses real package directories and replaces a global symlink, then runs the actual post-core producer and CLI admission consumer through a substituted spawn boundary. It verifies both acceptance of the replacement and rejection of a foreign installation, plus preservation of the original helper metadata.

With the production fix reversed, the regression fails with `Managed update handoff root mismatch`: the expected root ends in `.pnpm/openclaw@1.0.0/node_modules/openclaw`, while the unchanged global link resolves to the 2.0.0 generation. With the fix restored:

```text
node scripts/run-vitest.mjs src/cli/update-cli/update-command-post-core-root.test.ts src/cli/update-cli/update-command-post-core.test.ts
Test Files 2 passed (2)
Tests 19 passed (19)
```

`node scripts/check-changed.mjs` passed, including core/test typechecking, formatting, lint, and guards. The repository wrapper's required helper runtime build passed.

The broader native-transaction and handoff-lifecycle run produced 186 passes and one failure in the unchanged `preserves update failure after unavailable triage (0, timeout=true, missing=false)` case: its service-restore step returned exit 1 instead of 0. The fixture sets `recoveryTimeoutMs: 1000` when `triageHang` is enabled (`update-managed-service-handoff-boundary.test-support.ts:332`), also constraining the preceding service recovery. This run was under concurrent host load. The native transaction cases and verified previous-generation recovery/unsafe-refusal cases passed. No timeout, assertion, or recovery change was made to conceal this failure. `node scripts/run-vitest.mjs src/cli/update-cli/update-command-lease.test.ts` passed all 31 CLI lifecycle/lease tests.

Known gap: the historical pnpm 10.28.2 pre/post canonical paths from #142715 are unavailable. This proves and fixes the generation-transition failure shape, not a full reproduction of that user's release upgrade. No live Gateway or real package-manager upgrade was exercised. The fix is owned by the launching updater: installing a target containing it cannot retroactively change an already-running older updater during that first hop. The issue is referenced rather than automatically closed. The independent canary and unsafe-recovery reports remain outside this patch. Lead-owned autoreview and landing are pending.

Exact-head CI: [job 102493298024](https://github.com/openclaw/openclaw/actions/runs/34359621941/job/102493298024) failed in `checks-node-changed-extensions-bundle-15`. The failure is `extensions/matrix/src/matrix/monitor/handler.active-turn-steering.test.ts:176`, “lets configured steer mode reach queue policy while the prior Matrix turn is active”: expected the queue-policy spy once, received zero calls. That Matrix shard had 780 passes and one failure. This patch changes no Matrix or inbound-message implementation; the failure is recorded for a separate repair, without retrying or weakening its timeout/assertion. CI is not green.

### DIFF
--- a/docs/cli/update/how-updates-run.md
+++ b/docs/cli/update/how-updates-run.md
@@ -151,6 +151,11 @@ chained after the update. After a handoff result, use the printed follow-up comm
 for the final outcome. Plain terminal updates remain synchronous, and `--no-restart`
 does not authorize stopping the agent's Gateway.
 
+Post-core steps follow the verified replacement package, including pnpm updates
+that change the target of the global package link. The helper retains the
+original installation identity for recovery; a child running from a different
+installation is still rejected.
+
 The Gateway core auto-updater requires a managed service restart path. It hands
 the CLI update to a detached helper before activation. A foreground
 Gateway keeps update hints but leaves installation and activation to the
--- a/src/cli/update-cli/update-command-post-core-root.test.ts
+++ b/src/cli/update-cli/update-command-post-core-root.test.ts
@@ -1,0 +1,120 @@
+import { EventEmitter } from "node:events";
+import fs from "node:fs/promises";
+import path from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { writePackageRoot } from "../../infra/package-update-steps.test-support.js";
+import {
+  CONTROL_PLANE_UPDATE_SENTINEL_META_ENV,
+  readControlPlaneUpdateSentinelMeta,
+} from "../../infra/update-control-plane-sentinel.js";
+import { withEnvAsync } from "../../test-utils/env.js";
+import { createOpenClawTestState } from "../../test-utils/openclaw-test-state.js";
+
+const mocks = vi.hoisted(() => ({ spawn: vi.fn(), root: vi.fn<() => Promise<string>>() }));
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4ecee6b1353f9880aec8d9709280f47df3383b5d](https://github.com/openclaw/openclaw/pull/143137/commits/4ecee6b1353f9880aec8d9709280f47df3383b5d)
- Landed commit: [aaa5d326d65fee18eeb68b718171ed51539597d7](https://github.com/openclaw/openclaw/commit/aaa5d326d65fee18eeb68b718171ed51539597d7)

PR #142112, opened 2026-09-08
fix(cli): --url without credentials was framed as a CLI startup crash

DasX:
Closes #142111.

## What Problem This Solves

When `--url` or `OPENCLAW_GATEWAY_URL` lacks matching explicit credentials, the CLI now prints the existing credential remedy directly. It no longer adds a command/startup failure banner or unrelated Doctor, debug, and help hints.

## Why This Change Was Made

The change extends the shared expected-error predicate without importing the Gateway client. Credential enforcement, producer messages, exit code 1, and the canonical `cli_error` JSON structure remain intact. Human `nodes` commands now use that shared path. Command-owned machine error formats retain their existing handling. This builds on #125007, #125304, and #125556.

## User Impact

Operators get the existing credential remedy without misleading crash advice. Credential rejection, exit status, and machine output remain intact.

## Evidence

Verified full source CLI entry processes on pinned main `6e55c1df` and candidate `e785a1b6`, using isolated synthetic configuration and a loopback connection observer with no Gateway service:

- Nine missing-credential cases made zero connections. Human stdout stayed empty; stderr contained the complete remedy. JSON contained one canonical error object.
- Explicit token, password, and matching environment-token controls each reached the connection observer once, then reported the normal transport failure.
- Devices, nodes, environment URL, debug, JSON, and credential-shaped text masking passed independent behavior acceptance. The masking control preserves the remedy; it does not promise an identical replacement string.
- 22 shared-renderer tests and four entry-boundary controls passed on the current head. Four full-entry regressions passed on the original candidate; unchanged regression/producer inputs and fresh current-head CLI proof justify reuse after the contributor rebase. Three process cases and the added classification unit case failed on pinned main for the intended reason.
- Focused formatting, syntax lint, line/assertion checks, and `git diff --check` passed. Full CI owns broad typechecks and suites. Fresh independent AutoReview found no actionable findings.

Example candidate stderr (config path normalized):

```text
gateway url override requires explicit credentials
Fix: pass --token or --password *** --url (or gatewayToken in tools).
For the default local or SSH-tunneled Gateway, remove --url to use the configured target.
Config: <isolated-config>
```

The `***` in the remedy is existing console redaction. No live Gateway, external credential, or production state was used.

Implementation and regression coverage by @DasX. AI-assisted verification.

[...]

PR #143140, opened 2026-09-09
fix(cli): print immutable-config refusals without crash framing

steipete:
## What Problem This Solves

When an operator runs OpenClaw with an externally managed config (`OPENCLAW_CONFIG_READONLY=1`, added in #140719) or under `OPENCLAW_NIX_MODE=1`, the config write guard refuses mutations with a message that already carries the remedy ("Edit the config in your external deployment source, then redeploy or restart OpenClaw as needed."). `openclaw config set` printed that message plainly, but `openclaw plugins enable telegram` on a fresh isolated state (observed live at `71555bcbcb6`) printed:

```
[openclaw] The CLI command failed.
[openclaw] Reason: Config is externally managed (`OPENCLAW_CONFIG_READONLY=1`), so OpenClaw treats openclaw.json as immutable.
Edit the config in your external deployment source, then redeploy or restart OpenClaw as needed.
[openclaw] Help: openclaw --help
```

Every writer that throws through to the root renderer (`plugins enable|disable|install|uninstall|update`, `openclaw update`, doctor repairs) got the same crash framing. Doctor cannot lift an externally managed config, and there is no stack worth reading; the framing contradicts the guard's own guidance.

## Why This Change Was Made

`isExpectedCliError` in `src/cli/failure-output.ts` is the owner of "this is an operator condition, print the producer message as-is". It already classifies the Gateway credential preflight errors structurally by name (#142112) so the root renderer stays free of the config and transport stacks. This change adds `ConfigReadOnlyError` and `NixModeConfigMutationError` from `src/config/config-write-guard.ts` to that predicate with the same lean name check. The guard messages are already complete, and the `--json` path keeps the canonical `cli_error` envelope through the same predicate. `config set` keeps its own catch; nothing else changes.

## User Impact

Immutable-config refusals now print exactly the guard message on every command, with exit code 1 and no stack-trace or Doctor hint. Non-refusal failures keep their crash framing.

## Evidence

- Before (main `71555bcbcb6`, built dist, isolated state, `OPENCLAW_CONFIG_READONLY=1`): `openclaw plugins enable telegram` printed the six-line crash framing above.
- After (this branch, built dist, same command): prints only the guard message (two lines), exit 1; the same holds for `OPENCLAW_NIX_MODE=1` and for `plugins uninstall`.
- `src/cli/failure-output.test.ts`: two new rows in the existing "routes $label through the shared expected-condition predicate without crash framing" table; both fail on unmodified `main` (`isExpectedCliError` returns false), pass with the fix; the file passes at 24 tests.

Production LOC delta: +11 (one predicate with its contract comment, one line in the shared classifier).

### DIFF
--- a/src/cli/failure-output.test.ts
+++ b/src/cli/failure-output.test.ts
@@ -1,5 +1,6 @@
 // Failure output tests cover CLI error formatting and failure summaries.
 import { describe, expect, it } from "vitest";
+import { ConfigReadOnlyError, NixModeConfigMutationError } from "../config/config-write-guard.js";
 import {
   GatewayCredentialsRequiredError,
   GatewayExplicitAuthRequiredError,
@@ -219,6 +220,14 @@ describe("formatCliFailureLines", () => {
     {
       label: "gateway URL override without explicit credentials",
       createError: () => new GatewayExplicitAuthRequiredError(EXPLICIT_GATEWAY_AUTH_MESSAGE),
+    },
+    {
+      label: "externally managed config",
+      createError: () => new ConfigReadOnlyError({ configPath: "/tmp/openclaw.json" }),
+    },
+    {
+      label: "Nix-managed config",
+      createError: () => new NixModeConfigMutationError({ configPath: "/tmp/openclaw.json" }),
     },
     {
       label: "unreachable gateway",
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [41e6da52a584788c2fb22e65a6d702490846b11a](https://github.com/openclaw/openclaw/pull/143140/commits/41e6da52a584788c2fb22e65a6d702490846b11a)
- Landed commit: [ae65c403d032932e91eabdbf5f0299a3524982dd](https://github.com/openclaw/openclaw/commit/ae65c403d032932e91eabdbf5f0299a3524982dd)

PR #142322, opened 2026-09-08
fix(update): correct misleading Node compatibility guidance

SunnyShu0925:
## What Problem This Solves

Fixes an issue where `openclaw update` recommends Node versions that contradict the requested package's engine range. It also calls an incompatible version “too old” when it exceeds an exclusive upper bound.

## Why This Change Was Made

Refer to the complete target range already displayed and use direction-neutral wording throughout the message. Managed Gateway advice still identifies the runtime that owns the service. Four production lines replace four lines; the canonical parser, runtime selection and update behavior remain unchanged. The latest bot migration-proof checkbox is inapplicable: this four-string diagnostic rewrite changes no configuration, schema or persisted data.

## User Impact

Operators receive consistent guidance for choosing a compatible Node runtime and retrying the update. Thanks to @SunnyShu0925.

Closes #142208
Related: #142321 (the same underlying diagnostic issue).

## Evidence

[Baseline proof](https://github.com/steipete/openclaw/actions/runs/34289891442) reproduced both incorrect messages, with compatible and absent-target controls passing. The [refreshed candidate proof](https://github.com/steipete/openclaw/actions/runs/34310080906) passed all four owner cases and 16 checks, then hit its 15-minute deadline during core-test typechecking. That run remains failed; its retained overlay was not restored. The earlier candidate run also remains failed at a subsequently resolved upstream compatibility-record check.

[CI for the published commit](https://github.com/openclaw/openclaw/actions/runs/34316006896) passed. Its actual checkout and jobs independently cover the complete core-test graph, all five core-lint partitions and the remaining mapped guards. The three repair files match the focused proof, and both committed-head source reviews are clean.

The [two remaining canonical guards](https://github.com/steipete/openclaw/actions/runs/34348780290) passed on the published commit: `check:media-download-helpers` and `check:runtime-sidecar-loaders`. Both completed with exact output, unchanged inputs, joined processes, removed temporary state and clean final source audits. Final native landing gates still apply. This proof does not execute an update, installation or managed-service operation; the broad managed-update suite is not claimed as executed.

Issue #107930, opened 2026-07-15
[Feature]: Improve OpenClaw upgrade experience when Node.js version requirement changes


jialele:
### Summary

Improve the OpenClaw upgrade experience when the Node.js runtime requirement changes.

Currently, upgrading OpenClaw to a version with a newer Node.js requirement may require users to manually upgrade Node.js, reinstall packages, and fix the managed Gateway service runtime path.

The upgrade process should handle this migration more seamlessly.

### Problem to solve

When upgrading to OpenClaw v2026.7.1, the required Node.js version changed.

In my case, the previous environment was:

- Node.js v24.14.0
- OpenClaw installed globally with nvm

OpenClaw v2026.7.1 requires a newer Node.js version, so I had to manually:

1. Install a newer Node.js version using nvm (v24.18.0)
2. Switch the active Node.js version
3. Reinstall the global OpenClaw package
4. Update the Gateway service installation

After the Node.js upgrade, the shell CLI and the managed Gateway service were using different OpenClaw installations:

Shell OpenClaw:
/data/nvm/versions/node/v24.18.0/lib/node_modules/openclaw
Managed Gateway service:
/data/nvm/versions/node/v24.14.0/lib/node_modules/openclaw

This requires users to understand internal implementation details such as:
- Node.js version compatibility
- nvm environments
- npm global installation paths
- Gateway service runtime binding

The current upgrade experience is not smooth for end users.

### Proposed solution

Improve the upgrade workflow to automatically handle Node.js runtime changes.

Possible improvements:

- Detect incompatible Node.js versions during upgrade.
- Provide a clear migration flow instead of only showing an error.
- Automatically update/reinstall the managed Gateway service to use the current OpenClaw installation.
- Verify that the CLI and Gateway service are running from the same OpenClaw installation after upgrade.

The ideal user experience:
”openclaw update“
should complete the migration automatically, without requiring users to manually fix Node.js versions or service paths.

### Impact

This improvement would:
- Reduce upgrade failures caused by Node.js version changes.
- Avoid users needing to debug nvm/npm/systemd internals.
- Provide a smoother upgrade experience for existing OpenClaw users.
- Reduce support issues caused by mismatched CLI and Gateway service environments.

TheAngryPit:
Adding a narrower runtime-selection observation to this discussion, separate from the managed-service repair already shipped in v2026.9.1.

[...]

PR #143159, opened 2026-09-09
fix(update): replace unsupported Gateway service Node runtimes

steipete:
Refs #107930

## What Problem This Solves

Fixes an issue where users running `openclaw gateway install` after upgrading Node could be told the service was already installed while its recorded Node executable was unsupported. A macOS field report described Homebrew Node 26.8.1 and OpenClaw 2026.9.3 installed successfully while the LaunchAgent still referenced another tool's Node 22.23.1; forcing a reinstall recovered the Gateway.

The contradictory update-preflight guidance is already repaired on this base by #142322 (thanks @SunnyShu0925). This change adds explicit coverage for Node 22.23.2, 24.15.0, 25.9.0, and 26.0.0 against the target package's engine range.

## Why This Change Was Made

The shared Gateway install command now probes the recorded Node with the existing runtime classifier and treats an unsupported executable as a required refresh. Repair prefers the supported running CLI Node, retaining stable Homebrew paths, then uses the existing supported-system-node selection. The chosen runtime goes through the normal service install plan. Supported services, explicit wrappers, Bun selection, and service-definition ownership rules retain their existing behavior.

POSIX installer refreshes retain a fixed confirmation that Node was replaced after installation succeeds, while continuing to suppress incidental child output and runtime paths. Failed refreshes cannot report a successful runtime replacement.

## User Impact

Missing or non-executable recorded Node binaries now trigger the same supported-runtime replacement, while genuine probe failures on an existing executable retain an actionable `openclaw gateway install --force` hint.

Users can repair a stale unsupported service Node by running `openclaw gateway install` from a supported Node without adding `--force`. Installer and update refreshes share that repair. Genuine probe failures and missing supported replacements produce actionable errors instead of an already-installed result.

## Evidence

- Regressions fail against the original code: both non-force platform cases skip installation; forced cases omit the replacement notice; the resolver selects system Node instead of the requested supported CLI; both POSIX installers suppress the confirmation. The four version-guidance cases also fail with the diagnostic hunk from #142322 reverted.
- Source owner suites pass: 71 runtime-selector tests, 94 install/unit/integration tests on the final repair, and 8 target-package preflight tests.
- POSIX installer suites pass: 192 `install.sh` tests and 106 `install-cli.sh` tests, including success, failed refresh, output filtering, and no duplicate restart.
- `node scripts/check-changed.mjs`: passed, including production/test typechecks, lint, ratchets, unused-export scans, and runtime/security boundary guards. `git diff --check` and Bash syntax checks pass.

The definition fixtures use the real install plan and serializers, temporary files, a simulated obsolete Node, and the real supported CLI Node. Native service managers were not started. Windows-specific installer forwarding is outside this patch; PowerShell execution was unavailable locally. The broader preference question for an already-supported private Node in #107930 remains open. Maintainer review and landing are pending.

Compatibility evidence: this PR changes no SQLite schema, persisted user-data format, or store retention/recovery rules and introduces no service-definition format. Existing readers decode the recorded launchd/systemd definitions and the normal install plan writes their replacements. Upgrade fixtures cover unsupported, missing, and non-executable recorded Node paths with and without force, verify the Gateway entrypoint is retained, and keep supported-runtime, explicit-wrapper, genuine-probe-error, and definition-ownership controls covered. No database migration or backfill is needed.

Follow-up validation: eight missing/non-executable definition cases failed before repair and pass afterward; malformed-version and missing-notice regressions also fail before/pass after. All 173 requested owner tests, six focused installer-notice cases, and `node scripts/check-changed.mjs` pass. Independent follow-up autoreview is clean at P0/P1. Hosted CI is green for `f8066edee26fd9208464b8863ed0e956c187ca88`: https://github.com/openclaw/openclaw/actions/runs/34366200361. No rerun was needed on this head.

### DIFF
--- a/docs/cli/gateway/service.md
+++ b/docs/cli/gateway/service.md
@@ -137,6 +137,7 @@ openclaw gateway restart
   </Accordion>
   <Accordion title="Service runtime">
     - Node is the primary, default, and recommended managed Gateway runtime.
+    - `gateway install` checks the Node executable recorded in the managed service. If it is unsupported, installation refreshes the service without requiring `--force` and reports the replacement path. This also applies when an installer or update refreshes the service. Repair prefers the current CLI's supported Node, retaining stable Homebrew paths, then checks supported system installations. Custom wrappers keep control of their runtime; protected service definitions still require their deployment owner to repair them.
     - Bun 1.4+ with WAL-reset-safe `node:sqlite` is available as an explicit opt-in with `gateway install --runtime bun`.
 
   </Accordion>
--- a/docs/cli/gateway/service.md
+++ b/docs/cli/gateway/service.md
@@ -137,7 +137,7 @@ openclaw gateway restart
   </Accordion>
   <Accordion title="Service runtime">
     - Node is the primary, default, and recommended managed Gateway runtime.
-    - `gateway install` checks the Node executable recorded in the managed service. If it is unsupported, installation refreshes the service without requiring `--force` and reports the replacement path. This also applies when an installer or update refreshes the service. Repair prefers the current CLI's supported Node, retaining stable Homebrew paths, then checks supported system installations. Custom wrappers keep control of their runtime; protected service definitions still require their deployment owner to repair them.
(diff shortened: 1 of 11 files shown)


steipete:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [f8066edee26fd9208464b8863ed0e956c187ca88](https://github.com/openclaw/openclaw/pull/143159/commits/f8066edee26fd9208464b8863ed0e956c187ca88)
- Landed commit: [fe79d36c0ddef3d1d1d0baf0a65c30973dd68766](https://github.com/openclaw/openclaw/commit/fe79d36c0ddef3d1d1d0baf0a65c30973dd68766)

PR #143162, opened 2026-09-09
fix(android): wait for picked attachments before sending captions

steipete:
## What Problem This Solves

Android could send a caption without the file the user had just selected while attachment loading was still in progress. The picker import was tracked internally, but Send did not observe or check that pending work. Failed document or video imports could also be described as image failures.

## Why This Change Was Made

Use the existing import-owner map as observable state and check it at the composer's send-admission boundary. Publish completed attachments before releasing that gate. Cancellation and owner migration use the same existing map; there is no second counter, queue, or background loader.

The gate applies to the manual draft, not assistant auto-send's separate explicit payload. This restores the documented rule that sending waits while attachment staging owns the draft. The common import warning now says attachment rather than image. No configuration, dependency, protocol, permission, or persistence changes. Production grows by nine lines to connect the existing import state to UI and admission; the old unobservable map is replaced, not duplicated.

## User Impact

A caption waits for selected attachments to finish loading. Cancelling a picker or removing one attachment preserves the rest of the draft. Completed imports become part of the admitted message, and a document-loading failure has an appropriate attachment warning.

## Evidence

- Three focused pre-fix regressions failed for the intended reasons: Send admitted while a valid document loader was held, the rendered Send control stayed enabled, and a document failure used an image warning. The same cases passed after the repair.
- 139 integrated Android tests passed, including draft/import completion and cancellation, real Compose send state, native strings, and assistant auto-send ownership siblings. The integrated head also passed ktlint, native-string verification, Play/ThirdParty compilation and independent P0-P2 review. Both Android lint variants passed before the mechanical main integration; hosted CI rechecks the current head.
- Installed the signed Play debug APK built from exact candidate `776e3e9ced4bfe73a30b6ab15028d9cf0e9098a0`. Through ordinary Android DocumentsUI and Photo Picker, entered a caption, cancelled each picker, selected a PDF plus an image, removed only the image, and pressed Send once. The caption and PDF survived; the fixture received exactly one attachment with the PDF's original 1,639 bytes and SHA-256 `389800772dae521626b21961de5cff8b45623195ea3bb0ca5fcbd8baa152e4cd`.
- The normal ACK, streamed completion and history settled with caption/PDF visible, composer cleared and no Stop control. The synthetic fixture and emulator stopped cleanly; task-owned temporary files and credentials were removed.

The screenshots show the candidate before removing the image and after the single PDF send completes. They are normal app screens containing only synthetic data. The installed flow proves ordinary picker/cancellation/removal/submission behavior; the delayed-loader interval and failure wording are covered by the rendered regression tests, not claimed as installed observations. This fixture does not execute a model or prove provider-side PDF processing. Its pre-seeded Search text and reply prefixes are inherited synthetic fixture context, not additional user sends.

Related: the Android improvement work from #134939. Existing behavior documentation: [Android chat and history](https://docs.openclaw.ai/platforms/android).

Candidate: caption with PDF and image before removing only the image

Candidate: one completed send with the caption and surviving PDF

## Current main integration

Current source: `93c006b580179eede7ffdbb2897e4fa1446d274b`. This update includes the landed CI timestamp repair #143207. The previous two Play jobs reached their 20-minute job budget after passing unit tests; neither cancellation is counted as a pass, and no timeout or coverage was relaxed.

The main-integrated Android candidate `36d0be5ac25552811e94b6f3a92b010b048767f7` passed 173 tests, both-flavor compilation, native-string verification, ktlint and independent review. Its signed APK also passed ordinary onboarding, New/rename/parent-child navigation/inactive-child archive, and Dark→Light→process restart→Dark restore→second restart, with the original theme family and accent retained. Those scenarios used a synthetic Gateway; unsupported background Settings reads were rejected, not represented as working services.

The final integration preserves the complete Android source tree, all seven repair afterimages, dependency lockfile and Android build-runner inputs from that tested candidate. Refreshed changed-file/native checks and independent P0–P2 review are clean. Artifact evidence keeps its original commit identity rather than being relabelled as this head. The final signed APK from exact `93c006b580179eede7ffdbb2897e4fa1446d274b` (SHA-256 `ca7260239ebd02d178d097bd019db7ed8d47ffaa1116402f91c2f058c5427c33`) also passed a fresh ordinary caption-plus-PDF send through main's current dispatch path. One Send carried the exact caption and original PDF bytes; ACK, both streamed updates and final reply settled with Ready status, an empty composer and no Stop control. The fixture, emulator, task app state and temporary credentials were cleaned up and independently checked. This remains synthetic-Gateway app/wire proof, not provider-side processing. All four Android jobs in CI run 34390233702 attempt 1 passed. Play and ThirdParty completed unit tests plus lint in 17m19s and 17m53s within the unchanged 20-minute budget; the lint invocations reused generated BuildConfig and Kotlin compilation as UP-TO-DATE. The whole run failed only on the existing MCP registration close/reconnect case in src/mcp/plugin-tools-registration.test.ts:133 (120-second timeout, measured 267 seconds). Its unchanged isolated local run passed in 46.6 seconds. Source and SDK review found no blocking reconnect cycle; expensive first-use fixture setup is a lead, not a proven root cause. The single unchanged failed-job retry passed, including the same MCP case in 98.6 seconds. The full exact-head workflow (attempt 2) and required openclaw/ci-gate are now successful, with 95 passing checks, 27 skips and none nonpassing. No source, timeout, assertion or mock changed. MCP first-load performance remains a separate follow-up, not a defect claimed repaired by this rerun.

Final integrated APK: caption and PDF sent once, reply complete


steipete:
Merged via squash.

- Prepared head SHA: [93c006b580179eede7ffdbb2897e4fa1446d274b](https://github.com/openclaw/openclaw/pull/143162/commits/93c006b580179eede7ffdbb2897e4fa1446d274b)
- Landed commit: [36de76610c257b5e58b1407b20df5bdafab1049e](https://github.com/openclaw/openclaw/commit/36de76610c257b5e58b1407b20df5bdafab1049e)

PR #143168, opened 2026-09-09
fix(tests): close paired-device databases before cleanup

steipete:
Related: #139428

## What Problem This Solves

Control UI paired-device test fixtures removed their temporary state directories while cached SQLite handles were still open. A full test run reported a WAL sidecar identity mismatch for this fixture's directory family; a separate ordinary-flow diagnostic confirmed the open handle immediately before cleanup.

## Why This Change Was Made

Give both paired-device helpers one local lifetime owner. It captures the database path while the temporary home is selected, awaits the complete callback, closes that exact cached database, and then removes the directory. If closing fails, the files remain intact. The scoped-device fixture now retires at its awaited callback boundary instead of a later temp-directory hook.

## User Impact

Test-only repair; production authentication, permissions, SQLite behavior, and the WAL guard are unchanged. All pairing inputs and HTTP assertions remain intact. This adds 13 test-support lines.

## Evidence

- The original ordinary case observed one open database and an existing WAL immediately before directory removal, with no prior close event. The diagnostic then safely closed the handle before unlinking.
- Both repaired ordinary and scoped cases observed natural closure before removal; the temporary roots were verified absent afterward.
- All 159 existing tests passed. Case identities/order, pairing callbacks, and unrelated source were preserved.
- Isolated controls verified callback settlement, original callback-error propagation, and that a close error prevents removal without creating live database resources.
- Mapped changed-file gate and fresh independent P2 review passed.

These results establish this fixture's cleanup defect and repair; they do not identify an individual historical invocation or claim an overall test-runtime improvement.

### DIFF
--- a/src/gateway/control-ui.http.test.ts
+++ b/src/gateway/control-ui.http.test.ts
@@ -22,6 +22,8 @@ import { createEmptyPluginRegistry } from "../plugins/registry-empty.js";
 import { resetPluginRuntimeStateForTest, setActivePluginRegistry } from "../plugins/runtime.js";
 import { AVATAR_MAX_DATA_URL_CHARS } from "../shared/avatar-limits.js";
 import { AVATAR_MAX_BYTES } from "../shared/avatar-policy.js";
+import { closeOpenClawStateDatabaseByPath } from "../state/openclaw-state-db.js";
+import { resolveOpenClawStateSqlitePath } from "../state/openclaw-state-db.paths.js";
 import { withEnvAsync } from "../test-utils/env.js";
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0b80d1f710a423344f58cea193b31c1ceefaca62](https://github.com/openclaw/openclaw/pull/143168/commits/0b80d1f710a423344f58cea193b31c1ceefaca62)
- Landed commit: [99abf6f285d71ae13daa57ffd634fbaae3c69772](https://github.com/openclaw/openclaw/commit/99abf6f285d71ae13daa57ffd634fbaae3c69772)

PR #143175, opened 2026-09-09
fix: reduce memory spikes when checking long sessions

steipete:
## What Problem This Solves

Resolves a problem where checking long conversations for preflight compaction or memory flushing causes excessive temporary memory use while looking for an older provider-usage anchor. The reader retained both the selected raw JSON window and all parsed events, including historical message content that accounting had already discarded.

## Why This Change Was Made

Account for usage and current-turn taint during synchronous newest-first visits inside one SQLite snapshot. Retain the usage facts and every later message needed for prompt projection, while releasing older message bodies. The existing 512-event scan, full-context retry, usage barriers, taint rules, and SQL/JSON failure precedence remain intact.

First-row and legacy flat projections can still contain leaf controls. Those exceptional tails retain the existing selector through a second traversal of the same snapshot. The ordinary full-event reader shares the streaming parser, and the assertion baseline shrinks to match a removed assertion.

## User Impact

Long conversations need less temporary heap for transcript accounting without changing compaction decisions or truncating context. Required trailing messages still determine memory use; exceptional leaf-controlled tails still materialize their selected output.

## Evidence

- 150 focused tests passed across memory-runner accounting, context eligibility, reset races, and SQLite context accounting. Added coverage includes repeatable snapshot reads, expired visitors, cursor cleanup, oldest-malformed/SQL error precedence, independent usage and taint boundaries, and exceptional leaf controls.
- A matched synthetic run through the real preflight owner used 640 active events totaling 16,861,580 serialized bytes, with 512 complete trailing messages. Both versions parsed 1,152 events and made the same below-threshold decision. Sampled peak live-heap growth fell from 76,474,344 to 43,138,920 bytes (31.8 MiB less). This forced-GC experiment includes cold runtime work; it is not a throughput benchmark or attribution of a production OOM.
- Final validation passed `node scripts/check-changed.mjs`, a 77-test focused accounting/storage/bootstrap pass after lint cleanup, and independent review with no actionable P0–P2 findings.
- [Blacksmith Testbox CLI proof](https://github.com/openclaw/openclaw/actions/runs/34361532126) built the candidate and ran real local CLI turns against an isolated mock provider. It charged all 637 trailing messages: 105,494 tokens stayed below the 108,000 threshold, while 109,494 committed compaction and carried the saved summary into the foreground request. Both raw JSON outputs contained the expected successful reply payload. Captured source hashes matched the local candidate, native process trees and ports closed, and the Testbox was stopped.

Exact-head hosted CI on the original commit failed in the Telegram loopback and Windows media tests, and both failures repeated on one unchanged retry. Investigation reproduced stale loopback socket reuse and outdated Windows provider-fixture wiring; the original Windows timeout did not expose its internal stage. Existing repairs #143197 and #143213 landed for those boundaries while our separate investigations were in progress. The redundant fixture PRs #143217 and #143218 were closed unmerged with their evidence retained. No assertions or timeouts were weakened.

This PR is refreshed onto main containing both existing repairs. The accounting patch is identical (`git range-diff`), and all five accounting source/test files are byte-for-byte unchanged. The allocation and CLI receipts above remain bound to original commit `a489e2d1e045081866e00ec604ed84633bdb7f08`; [Fresh exact-head hosted CI passed](https://github.com/openclaw/openclaw/actions/runs/34374037878) for rebased commit `bbf8977f43aae5f98da44851ea69e7adbbc03687`, including both Windows jobs. The completed ClawSweeper review for that exact head has no actionable findings.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -2063,7 +2063,7 @@ src/auto-reply/reply/agent-lifecycle-terminal.ts	2
 src/auto-reply/reply/agent-runner-event-handler.ts	1
 src/auto-reply/reply/agent-runner-execution.ts	1
 src/auto-reply/reply/agent-runner-fallback-candidate.ts	1
-src/auto-reply/reply/agent-runner-memory.ts	6
+src/auto-reply/reply/agent-runner-memory.ts	5
 src/auto-reply/reply/agent-runner-run.ts	1
 src/auto-reply/reply/agent-runner-session-reset.ts	1
 src/auto-reply/reply/agent-runner-trace.ts	1
--- a/src/auto-reply/reply/agent-runner-memory.test.ts
+++ b/src/auto-reply/reply/agent-runner-memory.test.ts
@@ -23,11 +23,13 @@ import { makeAssistantMessageFixture } from "../../agents/test-helpers/assistant
 import type { InternalSessionEntry as SessionEntry } from "../../config/sessions.js";
 import { isInternalSessionEffectsKey } from "../../config/sessions/internal-session-key.js";
 import {
+  appendTranscriptEvent,
   loadSessionEntry,
   readSessionTranscriptMessageEvents,
   readSessionTranscriptActiveStats,
   readTranscriptStatsSync,
   upsertSessionEntryCore,
+  waitForSessionTranscriptProjection,
 } from "../../config/sessions/session-accessor.js";
 import { replaceTranscriptEvents } from "../../config/sessions/session-accessor.sqlite-transcript-write.js";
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [bbf8977f43aae5f98da44851ea69e7adbbc03687](https://github.com/openclaw/openclaw/pull/143175/commits/bbf8977f43aae5f98da44851ea69e7adbbc03687)
- Landed commit: [4d37a868b7e89ec73172ee011ca98b614dc26ef1](https://github.com/openclaw/openclaw/commit/4d37a868b7e89ec73172ee011ca98b614dc26ef1)

PR #143181, opened 2026-09-09
refactor: share Memory Wiki visibility test callbacks

steipete:
Related: #139428

## What Problem This Solves

Two adjacent Memory Wiki visibility tests repeat the same full callback while differing only in whether the caller is sandboxed.

## Why This Change Was Made

Register the two original named cases from a typed table of primitive names and sandbox flags. Keep all vault, manager, search-hit, expected-result and nested callback allocations inside each test. The independent expected paths, assertions, cleanup and registration order remain unchanged. This removes 45 net test lines; the earlier session-input cleanup is not part of this change.

## User Impact

No product behavior changes. Both sandboxed and session-bound non-sandboxed visibility paths retain their existing coverage with less duplicated test code.

## Evidence

The full query and query-read suites passed all 74 cases before and after, with no skips and identical ordered names within each file. Substituting each row into the shared callback reproduces its original body; all outside-table tokens and per-test mutable allocations are preserved.

The mapped changed gate passed, and fresh P2 review found no actionable issues. No runtime speedup is claimed.

### DIFF
--- a/extensions/memory-wiki/src/query.test.ts
+++ b/extensions/memory-wiki/src/query.test.ts
@@ -1075,111 +1075,66 @@ describe("searchMemoryWiki", () => {
     expect(manager.search).toHaveBeenCalledWith("alpha", { maxResults: 5 });
   });
 
-  it("filters session memory hits outside the caller visibility policy", async () => {
-    const { config } = await createQueryVault({
-      initialize: true,
-      config: {
-        search: { backend: "shared", corpus: "memory" },
-      },
-    });
-    mockSessionTranscriptStore();
-    const manager = createMemoryManager({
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [517687bc76ea10e801cf54afae97c6d94933e061](https://github.com/openclaw/openclaw/pull/143181/commits/517687bc76ea10e801cf54afae97c6d94933e061)
- Landed commit: [cb684c2838f3450f5b61f5b7d4df09f63e1b27cb](https://github.com/openclaw/openclaw/commit/cb684c2838f3450f5b61f5b7d4df09f63e1b27cb)

PR #142869, opened 2026-09-09
fix: preserve native Git attribution across prompt replay

Marvinthebored:
Closes #142868

## What Problem This Solves

Fixes an issue where native agent turns with Git attribution send user text that changes when replayed on a later turn. The first request adds host Git instructions to the user message, while the transcript stores the original text. Later requests therefore lose that suffix from the historical message.

## Why This Change Was Made

Put native Git attribution in the existing runtime-context carrier so it follows the established persistence and replay policy. Keep the user-authored text separate. Preserve CLI, plugin-harness, and raw-model prompt behavior, existing fragments, and the current profile, consent, and credit-ordering rules.

## User Impact

Native agents retain their Git instructions without this avoidable change to historical user text. Append-only context survives later turns and cold session resume; transient context retains its existing rules. This fixes a demonstrated prefix mismatch, without claiming measured cache-hit or billing gains.

## Evidence

Tested baseline `71555bcbcb639a0d912f0133b0c14c438cab57f4` and candidate `05f676cfa18edfde324d6974ce58dcd21c812b6c` through a real authenticated Gateway-backed TUI seed and consecutive public `openclaw agent` commands. The actual native runtime called a deterministic local provider. Complete requests and canonical transcript/participant records were captured privately.

| Observed boundary | Before | After |
| --- | --- | --- |
| First measured user message on replay | Loses its Git-attribution suffix | Keeps the same user text |
| Git instructions | Model-only user suffix is absent from history | Existing hidden carrier remains after its user message |
| Gateway restart and same-session continuation | Baseline mismatch established before repair | Prior user text and carriers replay correctly |
| Transient context | Existing policy retained | Current instructions present; older transient context omitted as designed |
| CLI-only session without a profile participant | No attribution expected | No attribution added |

- The native regression fails on original code for the unwanted prompt suffix.
- All 185 focused tests passed: 156 command-attempt cases, 25 carrier lifecycle cases, and four attribution-producer cases. Coverage includes both raw-mode switches, CLI/plugin handoffs, existing fragments, retry, reopen, fallback, compaction, and profile-backed credit rules.
- Native formatting, focused lint, and commit hooks passed. Independent source review found no actionable defect. A fresh source-blind validator independently ran both public flows and passed replay, cold resume, transient context, and no-profile controls.

[...]

Issue #142868, opened 2026-09-09
Native Git attribution disappears from replayed user prompts


Marvinthebored:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

Native Git attribution is appended to a provider-bound user message but omitted from transcript replay, changing historical prompt text on later turns.

### Steps to reproduce

1. Run a native agent attempt with `gitCoauthorAttribution` set and append-only runtime-context replay.
2. Capture the provider-bound user message and replay the accepted turn on the next request.
3. Compare the historical user message: the attribution suffix is no longer present.

The accompanying PR adds a deterministic regression to the existing command-attempt suite. No paid provider is required to reproduce the model-only suffix.

### Expected behavior

Host Git instructions should not introduce a difference between the original user message and its transcript replay. The existing persisted runtime-context carrier provides a place for those instructions.

### Actual behavior

The first 116 messages, system text, and tools were identical across two captured requests. The next historical user message lost its Git attribution suffix; its adjacent runtime-context carrier stayed identical.

### OpenClaw version

2026.9.3; source defect also present at upstream b04a3f83ada.

### Operating system

macOS 26.6.1

### Install method

Source checkout

### Model

Deterministic regression uses the existing mocked native-agent fixture. This occurs before provider dispatch and is not model-specific.

### Provider / routing chain

Native OpenClaw agent -> provider-bound prompt. Original capture used an Anthropic-compatible proxy; private route details are omitted.

### Impact and severity

Affected: native agent sessions using Git attribution and transcript replay. Historical prompt mutation can invalidate a cache prefix and increase token costs. The captured pair demonstrates one such mutation; it does not establish every cause of cache misses.

### Sanitized work log

AI-assisted diagnosis, implementation, and independent review. Reproduced the model-only attribution suffix with a failing assertion, then passed the 151-test command-attempt suite and 20-test runtime-carrier suite with the fix. Independent review found no actionable correctness or security issues. Private transcripts are not published.

PR #143198, opened 2026-09-09
fix: stop injecting Git co-author attribution into every turn

steipete:
Refs #142868
Refs #142869

## What Problem This Solves

Fixes an issue where every agent turn on a Gateway session received Git co-author attribution text appended to the user prompt, whether or not the turn had anything to do with Git. A personal install logging office temperature through Home Assistant was told on every message that "Git commit attribution for this turn has no additional exact Co-authored-by trailer" plus a notice about omitted participants. Because that suffix was model-only and never persisted, the previous user turn changed shape on replay and invalidated prompt-prefix caches (#142868, reported by @Marvinthebored).

## Why This Change Was Made

Who to credit is session state, not per-turn conversation content. This deletes the per-turn injection on both the CLI-harness and native runner paths, the fallback and notice text, and the `gitCoauthorAttribution` plumbing through the command options. The resolver becomes a pure data function that returns trailers and logins, or nothing. The only model-facing form is now one block in the system prompt's Runtime section, rendered only when a shared session has verified co-authors beyond the configured primary Git author. Personal installs render nothing. The Gateway publication broker keeps applying credit directly in its own commits and pull requests, unchanged. The github skill's per-turn attribution paragraph is removed; the agent commits with plain `git` and `gh`. The Codex runtime, which builds its own developer instructions, gets the same conditional block there. #142869 landed first and moved the native-path suffix into the persisted runtime-context carrier, which fixed the replay mismatch but persisted the block in every historical turn on prefix-binding models; this PR removes that carrier routing together with the rest of the per-turn injection.

## User Impact

- Personal installs see no attribution text in any turn. Incognito sessions never carry credit.
- Shared Gateways still get exact `Co-authored-by` trailers on commits made from a session, via one stable system-prompt block that only changes when a new eligible person prompts, instead of a per-turn instruction set.
- Prompt-prefix caches are no longer broken by a model-only suffix on the previous user turn.
- Direct `git commit` by the agent relies on that system-prompt line; broker-published commits and PRs are unchanged.

## Evidence

- Regression coverage: `resolveGitCoauthorAttribution` returns `undefined` for sessions whose only participant is unresolved, opted out, or the primary author. On `main` these three cases fail with the old fallback prompt text; they pass here.
- New system-prompt tests: the block renders exactly once immediately after the `Runtime:` line; the prompt is byte-identical when there are no trailers; the lookup is skipped without config, agent, or session key and for isolated cron runs; a failing lookup logs a warning and never blocks prompt building.
- Focused suites via `node scripts/run-vitest.mjs`: eight agent and auto-reply files (586 tests) and three publication-broker files (66 tests), all passing.
- `node scripts/check-changed.mjs` (core, coreTests, docs lanes) and `git diff --check`: clean. No baseline or ratchet changes.
- Codex autoreview at P2 on the first cut: scoped-clean, zero findings, verdict "patch is correct". Two ClawSweeper findings were then accepted and fixed: the resolver now routes through the canonical session store helper so a configured `session.store` (and incognito keys) is honored, with a test that fails on the default-store read; and the Codex runtime receives the same conditional block in its developer instructions: the block is prepared once per attempt and carried on the attempt params like `extraSystemPrompt`, so no plugin-sdk export was added, covered by the plugin prompt test. A third finding was accepted as well: incognito sessions never render credit, because the Codex runtime freezes generic instructions for a live incognito thread and a recomputed block would refuse the next turn; incognito sessions also keep no durable participant history, so publishing identities from them was never right.
- Net change: 15 files, +263 / −343.

Thanks @Marvinthebored for the diagnosis in #142868 and the replay fix in #142869; this PR removes the carrier routing that fix introduced along with the rest of the per-turn injection.


steipete:
Both findings were accepted and are fixed in the latest push.

- Session store routing: `resolveGitCoauthorAttribution` now resolves the participant store through `resolveSessionStorePathForScope`, so a configured `session.store` and incognito keys are honored again. A regression test records participants under a templated custom store with an empty default database and fails on the previous head.
- Codex runtime: the rendered block is prepared once per attempt in `run-attempt-dispatch.ts` and carried on the attempt params as `gitCoauthorPrompt`, next to `extraSystemPrompt`. The native system prompt consumes it through a prepared-value path in `buildSystemPromptParams`, and the Codex plugin appends it to its developer instructions immediately before `extraSystemPrompt`. No plugin-sdk export was added; the SDK surface check is back within budget. The plugin prompt test covers exact placement and mode independence.

steipete:
Accepted and fixed at the producer. `resolveSessionGitCoauthorPrompt` now returns nothing for incognito session keys, so incognito developer instructions never contain credit and a credit-preference change cannot alter the frozen generic policy of a live incognito thread. This is deliberate product behavior rather than a delivery workaround: incognito sessions keep no durable participant history, and they should not publish participant identities into commit trailers either. The unit test covers the incognito skip alongside the isolated-cron skip, and the existing live incognito reuse tests already prove continuation with unchanged instructions. Non-incognito Codex threads keep the existing config-fingerprint resume path when instructions change, which is not a conversation-context injection.

steipete:
Merged via squash.

- Prepared head SHA: [90bcda97be8e0d688978877a97970ff4e3a5a670](https://github.com/openclaw/openclaw/pull/143198/commits/90bcda97be8e0d688978877a97970ff4e3a5a670)
- Landed commit: [31d03f782230799ea6ceea8b41b6602b29d0da50](https://github.com/openclaw/openclaw/commit/31d03f782230799ea6ceea8b41b6602b29d0da50)

PR #143128, opened 2026-09-09
fix(android): distinguish connection status from chat readiness

steipete:
## What Problem This Solves

Fixes an issue where Android users were told chat needed a connection even though the Gateway was connected. The chat header also announced “Offline” after a failed health check. Meanwhile, Settings could claim “All systems operational,” and Overview could claim “Healthy,” without having established overall health.

## Why This Change Was Made

Keep connection, chat readiness, and highlighted issues distinct. Health and the chat header now say “Not ready” when chat health is unconfirmed or fails. Overview reports “Online” and scopes its quiet state to “No highlighted items.” The redundant Settings status footer is removed; the version remains.

This changes presentation, not connection policy, health checks, retry timing, provider availability, permissions, or persistence. Unknown provider availability still does not manufacture an attention warning. Production code shrinks by 13 lines; existing regression coverage is extended rather than adding a new status owner.

## User Impact

Users can tell a failed chat-health check from an offline Gateway and use Refresh chat to check again. Refresh Logs only reloads logs. Detailed connection and node status remain available, and actual disconnects still show Gateway offline with recovery actions.

## Evidence

- 140 focused Android tests passed, including connected/failed-health accessibility and Health-screen cases, the Overview unknown-provider regression, and existing chat/composer coverage. The new Overview assertion failed on the previous wording: expected Online, actual Healthy. The earlier Health/header cases also have verified failing-before/passing-after results.
- Changed-file checks, ktlint, native-string verification, Play and ThirdParty compilation, and independent code review passed. Only the source string inventory is changed; generated translations follow the normal workflow.
- Installed the signed Play debug APK built from `110c524dd429806148502cb4e5b3dedcbc07ace7` in an isolated emulator. Ordinary onboarding and app navigation exercised healthy chat, a failed health response while both connections stayed open, Refresh Logs without a chat-health refresh, successful health recovery, Refresh chat, and the real Disconnect action. The fixture observed four health requests, two log reads, and both connections closing on Disconnect; it stopped cleanly.
- The installed Gateway is a synthetic local fixture, not a real provider. Overview's no-attention/unknown-provider branch is covered by the owner regression; the installed Overview run exercised its attention branch. No provider execution is claimed.

[...]

PR #143207, opened 2026-09-09
fix(ci): keep Android test and lint build timestamps consistent

steipete:
## What Problem This Solves

Fixes unnecessary Android build-metadata changes between CI's separate unit-test and lint processes. Each invocation otherwise generates a new build timestamp, so lint sees changed BuildConfig source even though the checkout is unchanged.

This was observed while investigating Android jobs crossing their existing 20-minute limit. The repair removes demonstrated redundant work; it does not claim that timestamp regeneration accounts for the entire job duration or guarantee a particular speedup.

## Why This Change Was Made

Capture one actual UTC instant for a normal Play or ThirdParty unit/lint pair and pass it to both commands through the existing `openclawBuildTimestamp` Gradle property. Keep the separate processes, full task lists, cache and Robolectric arguments, failure ordering, runner/memory settings, and timeouts.

Unit-only compatibility tasks, Wear, build, and ktlint rows retain their existing metadata behavior. There is no new public configuration option or change to application build-timestamp defaults outside these CI pairs. The runner documentation is updated at its existing canonical page.

## User Impact

Android validation no longer regenerates build metadata solely because lint starts later. Checks and coverage remain intact; failed unit tests still stop the paired lint invocation.

## Evidence

- A real four-invocation Gradle experiment ran `:app:generatePlayDebugBuildConfig`. Two default invocations produced different source hashes, differing only in BUILD_TIMESTAMP. Two invocations with the same existing timestamp property produced identical bytes, with the second task reported UP-TO-DATE. This is generated-input evidence, not a full compiler/lint benchmark.
- The existing executable CI-shell fixture was extended with an advancing clock. Both new cases failed before the workflow change because the property was absent, then passed afterward. They verify one clock read and the same property in both child commands.
- 32 focused Android CI contract tests passed, including the new cases, retained task/compatibility coverage, and failure propagation. Actionlint, canonical changed-file checks, formatting, and independent P0-P2 review passed.
- The investigation's original Health Play job completed unit tests in 13m4s and lint in 6m1s but exceeded its overall job budget. Its unchanged-job retry passed. The picker job completed unit tests but was cancelled during lint at the same limit. Neither failed run is relabelled as passing evidence, and this PR changes no timeout or runner capacity.

Source: `82337c79417c5614749057bbf2cbd3ce50156198`. Related: #143128 and #143162.

## Main integration and CI recovery

Integrated main to consume the canonical Windows test-fixture repair in #143213. The original head failed the same 120-second Windows media test twice. That fixture still mocked the old provider-list function after image generation moved to provider acquisition; the canonical repair supplies the same synthetic provider through the current acquisition boundary without dropping file handling or reference-image assertions. Its exact-head [Windows case passed in 804 ms](https://github.com/openclaw/openclaw/actions/runs/34369296327/job/102526310237), and its complete CI run passed. The earlier timeout logs do not identify the precise inner phase, so no phase-level timing claim is made.

The Android timestamp patch is unchanged by this integration. The refreshed 32-case workflow proof, actionlint, canonical three-path changed checks, and independent P0–P2 review passed on the integrated tree. This head still requires its own green CI; upstream results are not a substitute. No timeout, runner, or assertion was relaxed. A supplementary raw-Windows diagnostic stopped before the test because the image lacked tooling; it supplies no behavioral proof and its lease was released.


steipete:
Merged via squash.

- Prepared head SHA: [82337c79417c5614749057bbf2cbd3ce50156198](https://github.com/openclaw/openclaw/pull/143207/commits/82337c79417c5614749057bbf2cbd3ce50156198)
- Landed commit: [99fef046a4f77ae2d2733ee12b58d766f79182a0](https://github.com/openclaw/openclaw/commit/99fef046a4f77ae2d2733ee12b58d766f79182a0)

PR #143213, opened 2026-09-09
fix(test): keep Windows media file URL coverage isolated

steipete:
Related: #142991

## What Problem This Solves

Resolves a problem where Windows media file-URL coverage could enter real plugin discovery despite configuring a synthetic image generator. The fixture still mocked the old provider-list function after image generation moved to provider acquisition.

## Why This Change Was Made

Supply the same synthetic provider through `acquireImageGenerationToolProviders`, with the fixture's resource callbacks, following the current image-generation test boundary. Real Unicode and space paths, registered image/PDF tools, reference-image bytes, existing assertions, and the Windows guard remain unchanged.

## User Impact

No product behavior changes. Windows CI exercises media file handling without unrelated plugin-registry acquisition.

## Evidence

- Local macOS sibling proof: 68 image-generation and PDF-provider tests passed. The Windows-only case was skipped and is not counted as Windows proof.
- The exact-base changed gate, formatting, and `git diff --check` passed.
- Independent P2 review is clean with no actionable findings.
- The earlier Windows case timed out in [run 34362692276](https://github.com/openclaw/openclaw/actions/runs/34362692276/job/102503681660). That log has no phase-level timing, so it does not establish the precise stalled step.

Native Windows proof passed on the exact PR head: the original Unicode/space file-URL case executed with all existing assertions in **804 ms** in [Windows test part 1](https://github.com/openclaw/openclaw/actions/runs/34369296327/job/102526310237). The job passed. The complete exact-head CI run also passed.

### DIFF
--- a/src/agents/tools/media-tool-file-url.windows.test.ts
+++ b/src/agents/tools/media-tool-file-url.windows.test.ts
@@ -9,6 +9,7 @@ import * as imageGenerationRuntime from "../../image-generation/runtime.js";
 import * as mediaStore from "../../media/store.js";
 import { createOpenClawTools } from "../openclaw-tools.js";
 import { createImageGenerateTool } from "./image-generate-tool.js";
+import * as mediaGenerationToolProviders from "./media-generation-tool-providers.js";
 import * as pdfNativeProviders from "./pdf-native-providers.js";
 import {
   createPdfToolInfraStub,
@@ -108,7 +109,7 @@ describe.runIf(process.platform === "win32")("host-local media tool file URLs", 
         expect(pdfResult.content).toEqual([{ type: "text", text: "native summary" }]);
         expect(pdfResult.details).toMatchObject({ pdf: pdfPath, native: true });
 
-        vi.spyOn(imageGenerationRuntime, "listRuntimeImageGenerationProviders").mockReturnValue([
+        const providers = [
           {
             id: "fixture",
             defaultModel: "edit",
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d1c7fdc814463b192c3347bb075e364e75b8038c](https://github.com/openclaw/openclaw/pull/143213/commits/d1c7fdc814463b192c3347bb075e364e75b8038c)
- Landed commit: [b1db51436d8f623bd24d1ad1e04b29e31eb89740](https://github.com/openclaw/openclaw/commit/b1db51436d8f623bd24d1ad1e04b29e31eb89740)

steipete:
Phase-level attribution for the stalled step this PR body said the Windows log could not establish, measured on a build-less macOS worktree at the pre-fix tree with the platform guard flipped:

- The test body reached `pdf-done` at +5.1 s, then the worker spun at 100% CPU with ~2.3 GB RSS and printed nothing for the rest of the 120 s no-output window until the wrapper killed the process group (exit 143). Markers placed inside `loadImageGenerationReferences` never fired, so the stall is before reference loading: `acquireImageGenerationToolProviders` with a prepared snapshot that carries no `mediaCapabilityProviders` falls through to `acquirePluginCapabilityProviders`, i.e. cold bundled-plugin acquisition that transpiles extension sources synchronously through jiti in a checkout without `dist`. That also explains the CI duration overshoot (179 s reported against a 120 s timeout): the timeout timer could not fire while the event loop was blocked.
- With `OPENCLAW_DISABLE_BUNDLED_PLUGINS=1` the same body fails in 169 ms at the PDF step (`enable the document-extract plugin`), confirming both tool steps were entering bundled plugin discovery.
- At the landed head the same body completes in ~14 s on the same loaded host.

PR #143190, opened 2026-09-09
fix(update): keep bundled plugin aliases bundled in candidate rehearsal

steipete:
Refs #142633

<details>
<summary>Additional instructions</summary>

Maintainers can update this same-repository branch.

</details>

## What Problem This Solves

Fixes an issue where users running `openclaw update` with path-install aliases of bundled plugins can have the candidate Gateway reject a plugin that the serving Gateway loads as bundled. This can prevent candidate readiness and block the update.

## Why This Change Was Made

Copying a bundled alias into independent storage changed discovery to `origin=global` and `trust.reason=origin-path`. Candidate projection now redirects only physically verified bundled plugin roots and entrypoints to the staged bundled plugin with the same manifest ID, using runtime directory selection for source-checkout and split-runtime layouts.

Source metadata inspection stays read-only, and candidate containment is checked before discovery. External paths, escaping aliases, and absent or mismatched staged plugin IDs retain the existing independent-copy path and trust policy. Install-record source values are unchanged.

## User Impact

Candidate rehearsal preserves bundled ownership for these aliases while keeping external path installations subject to their existing trust restrictions. Live plugin files, permissions, install records, and host links remain untouched. The update workflow documentation describes this behavior.

## Evidence

The regression runs the real rehearsal snapshot worker and then real manifest discovery against synthetic plugin packages and SQLite state. With the production fix temporarily reversed against the branch base, the directory-alias regression fails with:

```text
- origin: bundled
+ origin: global
  installSource: path
- trust.reason: bundled
+ trust.reason: origin-path
Test Files 1 failed; Tests 1 failed, 11 skipped
```

The repaired regression matrix covers direct and symlink aliases, entrypoints, renamed staged directories, split runtime and source-checkout layouts, external paths, source/candidate symlink escapes, absent/mismatched IDs, and preservation of live permissions and database bytes. A separate before/after regression proves a candidate bundled-root symlink cannot change permissions outside the staged package during projection.

Focused validation: **205 tests passed across 9 files**, including a real built Gateway reaching started/ready and verified process-group/state cleanup. The test runner rebuilt the final runtime successfully. The final provenance-only rerun passed all **13 cases**; changed-file checks, including production/test typechecks and lint, also passed.

```sh
node scripts/run-vitest.mjs \
  src/infra/update-candidate-bundled-provenance.test.ts \
  src/infra/update-candidate-state.test.ts \
  src/infra/update-candidate-canary.test.ts \
  src/infra/update-candidate-canary.integration.test.ts \
[...]

PR #143174, opened 2026-09-09
fix(update): converge plugin cohort when core is already current

steipete:
Refs #135776 #133884

## What Problem This Solves

Fixes an issue where users who upgraded core with `npm i -g openclaw@latest` and then ran `openclaw update` got an `already-current` result without the normal plugin update pass. Official plugins could remain on the previous release, and retained exact pins did not receive the existing aggregate advisories.

## Why This Change Was Made

Run the existing post-update convergence/finalization flow when core is already current. Reuse plugin target admission, managed-service ownership and handoff checks, the managed environment, capability/integrity policy, and the existing retained-pin warnings. A changed plugin result parks an owned running Gateway only for fresh Doctor validation, then uses the existing restart and health verification; unchanged plugins do not restart it. `--no-restart` retains manual activation with explicit guidance.

Core files are not replaced. Git no-ops retain their current version from the before-only receipt. Exact-pin intent and plugin cohort-selection policy are unchanged; this does not infer whether a legacy pin was intentional. Finalization parameter types live in a separate contract, and existing downgrade-confirmation code moves into shared update primitives to keep the orchestration files within the existing line limits without dependency cycles.

## User Impact

One `openclaw update` invocation now performs the normal plugin maintenance even after a manual core upgrade. JSON uses the existing `postUpdate.plugins` fields. Retained pins are reported, unchanged runs retain `skipped` / `already-current`, and stopped or absent services are not started by this plugin-only path.

## Evidence

The final regression tests ran against the original production source and then the restored fix using the same command:

```sh
node scripts/run-vitest.mjs src/cli/update-cli.test.ts -t 'converges plugins on an already-current core.*failure=undefined' --reporter=dot
```

Before: **3 failed**, each with `expected "vi.fn()" to be called once, but got 0 times` at the plugin-update assertion. After: **3 passed**. Cases cover a running Gateway, `--no-restart`, and an absent service.

The full update CLI, finalization/recovery, execution, and shared-command suites passed **569 tests**. Subsequent focused checks passed **15 tests**, including retained-pin text/JSON advisories, unchanged/channel-only results, plugin-target refusal, service ownership changes, Git receipts, and Doctor/partial-stop failures. The final result-builder relocation was followed by the three-case regression rerun above. The test harness also completed its required runtime build and bootstrap/plugin-loading guards.

[...]

PR #143141, opened 2026-09-09
fix(doctor): preserve ambiguous Workshop migrations for review

steipete:
## What Problem This Solves

Fixes an issue where users updating OpenClaw would have Doctor stop when valid, inert legacy Skill Workshop proposals or collection backups could not be assigned to exactly one configured agent. It also makes remaining external proposal targets distinguishable from proposals successfully retargeted during the same pass.

Fixes #143094
Refs #142583
Refs #142770
Refs #142582

## Why This Change Was Made

Workshop already preserved ambiguous artifacts, but its warnings became fatal migration receipts. Classify ownership-only deferrals as recoverable, name the retained path and candidate agents, discover backup roots before Workshop mutations, and preflight all sidecar artifacts and owners before importing proposals.

Read-only diagnostics identify up to 20 remaining proposal IDs, target paths, and owners. After Doctor has attempted migration, explain that ownership/recovery blockers must be resolved before retrying. Successful retargeting uses the existing canonical target resolver.

## User Impact

Doctor can finish eligible work while preserving validated, inert ambiguous proposals and backups for manual review. Corrupt artifacts and unfinished applies still stop Doctor, with their recovery evidence retained. Operators can identify remaining targets without confusing their count with successful retargets.

This does not fix 2026.9.2 package rollback after state migration. The v16 schema transaction commits before later Doctor work; an already-running 9.2 updater can restore old package files without restoring the database or relocated skills. Full recovery needs a separately approved design. Multi-agent config recovery in #142582 is also outside this change.

## Review notes

The accepted P1 is repaired at the legacy bundle validator. Before an ownership deferral, Doctor validates rollback metadata, reads PROPOSAL.md with the existing symlink/hardlink and size guards, and verifies its hash. An ownerless bundle with rollback evidence remains fatal. A missing draft with rollback evidence also stays fatal instead of entering incomplete-bundle quarantine. Owned bundles are checked again immediately before import through the same validator. Valid ownerless bundles without rollback evidence remain advisory.

For #142583, the reporter's original records were not preserved. This PR proves convergence and diagnostic ambiguity, not the exact reported eight-record failure; the issue is referenced rather than marked fixed.

No persisted SQLite tables, columns, indexes, or stored record formats change. The changes are limited to Doctor classification, preflight ordering, validation, and diagnostics; the existing v16 schema migration and package rollback implementation are unchanged.

[...]

PR #143139, opened 2026-09-09
fix(update): confirm before automatic triage launches a coding agent

steipete:
<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Fixes an issue where users running `openclaw update` or `openclaw update repair` could unexpectedly launch a coding agent after an interactive update failure, spending their own account quota or configured-model tokens. Community feedback after 2026.9.3 described surprise Claude launches, substantial subscription quota usage, and expensive inference, while other users valued the automatic repair.

## Why This Change Was Made

Before an automatic interactive update-failure launch, triage names the selected binary (or the embedded OpenClaw agent using the configured model), prints the saved prompt path when available, explains that it uses the operator's own account/tokens, and asks `Open <agent> to diagnose and repair the installation now? [Y/n]`.

Enter or `y` continues. `n` launches no agent and prints the existing ready-to-run handoffs plus an installation-specific manual `openclaw triage` command. Diagnostics and the updater's original failure exit status remain intact. If this confirmation receives no answer for 30 seconds, it aborts and settles the existing Clack prompt, announces `No answer; continuing with <agent>`, and proceeds as Yes. This default preserves unattended operation at this prompt. Cancellation and current update ownership are checked before launch.

## User Impact

Operators can decline unexpected account/token usage while retaining a usable repair handoff. Explicit `openclaw triage` does not gain a confirmation. Agent selection, JSON/non-interactive behavior, and existing owned managed recovery remain unchanged. The update/triage documentation describes the confirmation and timeout; no configuration, dependencies, or persistent-store changes are added.

## Evidence

Regression command:

```sh
node scripts/run-vitest.mjs src/commands/triage.test.ts -t 'declining automatic update repair'
```

Before the production fix, both external and embedded decline cases failed because their agent execution mock was called once. After the fix, both pass without agent execution and retain saved failure diagnostics and manual commands.

```text
Before: Tests 2 failed | 35 skipped (37)
  expected "vi.fn()" to not be called at all, but actually been called 1 times
After:  Tests 2 passed | 35 skipped (37)
```

[...]

PR #143230, opened 2026-09-09
fix(test): Telegram model-callback loopback test fails with editMessageText ECONNRESET on unrelated PRs

steipete:
## What Problem This Solves

Fixes a main-side flake where `extensions/telegram/src/model-callback.loopback.integration.test.ts` failed on PRs that did not touch Telegram, then passed on rerun. On 2026-09-09 it hit at least #143139, #143159, #143174, #143190, and #143141 (for example [bundle-27](https://github.com/openclaw/openclaw/actions/runs/34362151740/job/102501745976) and [bundle-10](https://github.com/openclaw/openclaw/actions/runs/34365714887/job/102514685679)) with:

```
AssertionError: expected '❌ Failed to change model: HttpError: …' to contain 'Model changed to <b>ollama/…'
Received: "❌ Failed to change model: HttpError: Network request for 'editMessageText' failed!"
```

## Why This Change Was Made

The test's loopback Bot API stub is a plain `node:http` server, whose default `keepAliveTimeout` is 5 s. grammY's Node client (node-fetch with a keep-alive `http.Agent`) answers the callback query and then edits the message over the same idle socket. Between those two requests the router awaits `applySessionModelSelection`, which loads the provider model-policy surface synchronously (`addAllowedCatalogRef` → `resolveProviderModelPolicySurface` → jiti source transform). In a build-less checkout such as CI that blocks the worker event loop for 6–9 s, so neither the client's 4 s free-socket timer (Node subtracts a 1 s buffer from the server's `Keep-Alive: timeout=5` hint) nor the server's 5 s idle timer can run first. When the stall ends, the continuation writes the edit onto the idle socket before the timers phase, then the server's keep-alive timer destroys that socket under the in-flight POST → `read ECONNRESET` → the router's error path edits the message with "Failed to change model". Whether a run fails depends only on whether the stall crosses 5 s, which is why it flaked per runner speed.

The stub now sets `server.keepAliveTimeout = 0`, so it never idle-closes a connection on its own timer; connections are retired by the existing `server.closeAllConnections()` in the `finally` block. The client's own free-socket handling is unchanged, and the test still drives both API calls over one persistent connection as with the real Bot API. No test timeout, retry, or assertion was changed.

The synchronous jiti load itself is the known build-less-checkout cost of plugin public-surface loading (production installs load the built artifact); it is not changed here.

## User Impact

No user-facing change. Contributors stop seeing this unrelated Telegram test fail on their PRs.

## Evidence

Reproduction (fresh worktree from `origin/main`, `node scripts/run-vitest.mjs extensions/telegram/src/model-callback.loopback.integration.test.ts`): failed 2/2 plain runs before the fix (test 9.5 s, same `HttpError`).

Instrumented timeline from a scratch copy (socket events, grammY API transformer, event-loop stall monitor):

```
[+    65ms] client: answerCallbackQuery done          (idle keep-alive socket :56918)
[+    67ms] mock: catalog
[+  6161ms] client: editMessageText start             (written to :56918)
[+  6165ms] EVENT LOOP STALL 6108ms
[+  6165ms] server: socket :56918 TIMEOUT (keepAliveTimeout=5000)
[+  6166ms] client: editMessageText ERROR HttpError: Network request for 'editMessageText' failed!
            cause=FetchError: … reason: read ECONNRESET
[+  6167ms] server: request editMessageText on socket :56973   (error-path edit, new connection)
```

Worker CPU profile (`NODE_OPTIONS=--cpu-prof`) for the stall, inclusive time:

```
7842ms  applySessionModelSelection        src/model-picker/apply-session-model-selection.ts
7254ms  addAllowedCatalogRef              src/agents/model-selection-shared.ts
7550ms  resolveProviderModelPolicySurface src/plugins/provider-model-routes.ts
7549ms  loadPluginPublicSurfaceModuleSync src/plugins/plugin-module-loader-cache.ts
7436ms  jitiRequire                       jiti/dist/jiti.cjs
```

Deterministic check with an injected synchronous stall in the catalog mock:

| Variant | Result |
| --- | --- |
| 5.5 s sync stall, default `keepAliveTimeout` | fails, same `ECONNRESET` on `editMessageText` |
| 5.5 s sync stall, `keepAliveTimeout = 0` | passes (11.4 s total stall) |

After the fix, five consecutive runs of the real test in the same build-less worktree (stall still present, test 6–9 s each):

| Run | Result | Test time |
| --- | --- | --- |
| 1 | pass | 6.7 s |
| 2 | pass | 7.2 s |
| 3 | pass | 7.0 s |
| 4 | pass | 6.5 s |
| 5 | pass | 5.7 s |

`node scripts/check-changed.mjs -- extensions/telegram/src/model-callback.loopback.integration.test.ts`: all lanes pass except `lint extension changed file`, which fails before linting in this nested worktree (the tsdown declaration-boundary guard refuses ancestor `node_modules`); direct `oxlint --tsconfig extensions/tsconfig.json` on the file passes, and CI runs the real lane.

Codex autoreview (`--mode local`, P2): scoped-clean, no P0–P2 findings.

Note: this file runs under `vitest.extension-telegram.config.ts` with `isolate: true` and `fileParallelism: false`, so the shared-worker (`isolate: false`) mode does not apply to it; the single-file shard in CI matches the local invocation above.

### DIFF
--- a/extensions/telegram/src/model-callback.loopback.integration.test.ts
+++ b/extensions/telegram/src/model-callback.loopback.integration.test.ts
@@ -96,6 +96,14 @@ describe("Telegram model callback loopback", () => {
         response.destroy(error instanceof Error ? error : new Error(String(error)));
       });
     });
+    // grammY reuses one keep-alive socket for answerCallbackQuery and
+    // editMessageText. Between them applySessionModelSelection loads the
+    // provider model-policy surface synchronously (several seconds through
+    // jiti in build-less checkouts such as CI), so neither side's idle timer
+    // can run until the edit is already on the wire. Node's default 5 s idle
+    // close would then reset that in-flight request; connections are retired
+    // by closeAllConnections() in the finally block instead.
+    server.keepAliveTimeout = 0;
     await new Promise<void>((resolve) => {
       server.listen(0, "127.0.0.1", resolve);
     });


steipete:
Closing as superseded: #143197 landed `Connection: close` on this fixture's JSON responses (4d475f153ae) before this PR was opened, which already removes the idle-socket reuse this change targeted; my sibling-fix preflight missed it. No second fix is needed on main.

One detail from this investigation worth keeping for the loader-side follow-up: a worker CPU profile pins the multi-second synchronous stall to `applySessionModelSelection` → `addAllowedCatalogRef` → `resolveProviderModelPolicySurface` → `loadPluginPublicSurfaceModuleSync` (jiti source transform, about 7.5 s in a build-less checkout), which is why neither keep-alive timer could run before the edit was written. Full evidence stays in the PR body above.

PR #143237, opened 2026-09-09
refactor(test): reuse shared budget flag parsing

steipete:
## What Problem This Solves

The performance-budget script copied the shared numeric flag parser during its TypeScript migration because two destination fields start as `null`. The copy had identical parsing and assignment behavior, leaving two implementations of the same validation rules.

## Why This Change Was Made

Let the shared helper accept that existing nullable initial state. It still assigns only a validated number. Reuse it for both wall-time flags and remove the private copy, its sole-use option type and unused imports: 42 tooling lines removed. Keep the numeric parser private now that only the shared helpers call it.

## User Impact

Flags, defaults, error messages, budget enforcement and runner behavior are preserved.

## Evidence

- The same six existing cases passed before and after: four performance-parser cases and two startup CLI/environment argument controls. Thirteen unrelated cases were filtered; tests were unchanged.
- All 16 selected changed-file checks passed, including full scripts typechecking.
- Independent review of the consolidation and the subsequent export-only cleanup found no actionable issues through P2.
- Initial CI caught the numeric parser’s now-unused export. After removing that export, both `deadcode:full` and the canonical dead-export check passed, followed by the same six cases and all 16 changed-file checks. The original failure is retained; [final-head CI](https://github.com/openclaw/openclaw/actions/runs/34376265694) is green.
- No benchmark speedup or changed budget is claimed; this consolidates identical parsing behavior.

### DIFF
--- a/scripts/lib/budget-number-args.mts
+++ b/scripts/lib/budget-number-args.mts
@@ -32,7 +32,7 @@ export function budgetFloatFlag<Key extends string>(flag: string, key: Key) {
         flag,
         nextIndex: index + 1,
         repeatable: false,
-        apply(target: Record<Key, number>) {
+        apply(target: Record<Key, number | null>) {
           const parsed = parseBudgetNumber(value, flag);
           if (parsed === null) {
             throw new Error(`${flag} requires a value`);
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7e1d8bec6642a79bb832eb24f0995bf225310441](https://github.com/openclaw/openclaw/pull/143237/commits/7e1d8bec6642a79bb832eb24f0995bf225310441)
- Landed commit: [d10ca466cd2b09230c735430de682ec1fbdfe060](https://github.com/openclaw/openclaw/commit/d10ca466cd2b09230c735430de682ec1fbdfe060)

PR #143240, opened 2026-09-09
fix(process): prevent false cleanup failures after successful commands

steipete:
## What Problem This Solves

Fixes an issue where users running a successful agent command or managed-service repair could receive a cleanup failure instead of the successful result. The POSIX process-group anchor could exit after writing its closing receipt but before the host consumed it. A crossing cancellation then failed with `EPIPE`, destroyed the host's unread control stream, and left cleanup correctly reported as uncertain.

## Why This Change Was Made

The anchor now stays alive until the host acknowledges the exact closing receipt on the existing private control channel. The host retires cancellation before acknowledging; process reaping and kernel-confirmed group disappearance remain independently required. Missing, stale, premature or lost acknowledgements use the original cleanup deadline and escalation path.

The private start envelope negotiates this behavior because documented `update --no-restart` can retain an older Gateway while worker files are replaced. A retained older host keeps its existing protocol until restart; newly started hosts request acknowledgements. Windows Job ownership is unchanged. No new configuration, persistent state, public RPC or protocol-version change.

Also corrects the handoff fixture's unrelated coupling between a hung diagnostic and the healthy recovery timeout, without increasing the diagnostic cap or weakening assertions. Production delta is +72 lines, tests +204 and test support +1: the growth implements the missing two-party retirement boundary using existing channels, deferreds and deadlines; no test-only production seam was added.

## User Impact

Successful POSIX commands and repairs retain their result when cleanup races with receipt delivery. Genuine missing ownership/closure evidence still fails cleanup. Gateways intentionally retained by `update --no-restart` gain the new behavior on their next restart.

## Evidence

- Captured the original race in the original handoff execution order: the anchor wrote its closing receipt, then the host cancelled before consuming it and recorded `service child cleanup identity lost: write EPIPE`. Nine regression scenarios fail against the original anchor and pass after the repair.
- Final source: all 29 normal `check-changed --base f2a629ea05ce3f46c3d89a6d45866c50ad2c04cf` checks passed; normal worker build and 18 targeted host/anchor/repair cases passed; independent Codex review clean through P2.
- Prior behavior-equivalent revision: original four-file ordered handoff run 226 passed/1 platform skip; five-file supervisor sibling run 102 passed/1 skip; isolated Linux Node 24.19 run 51 passed/1 Windows-only skip. The final revision only moves capability validation to raw IPC input for type-aware lint and removes unused test defaults; the final build and focused cases cover that delta.
- Actual Bun 1.4.2 worker coverage passed eight distinct retirement/compatibility scenarios. An initial observer incorrectly awaited `close` after parent IPC disconnect; its failed attempt is retained, and only the two unfinished scenarios were rerun with exit, pipe and IPC joins. The final rebuilt worker additionally passed matching-ACK and unsupported-capability cases, including import resolution. Owned groups, pipes, processes and remote leases were verified closed.

These are real built child-process and synthetic service/provider flows, not an operator Gateway deployment or a new live Windows run. Existing Windows host assertions verify that this POSIX negotiation and ACK are not sent to the Job worker.

### DIFF
--- a/src/infra/update-managed-service-handoff-boundary.test-support.ts
+++ b/src/infra/update-managed-service-handoff-boundary.test-support.ts
@@ -329,7 +329,8 @@ export function createManagedServiceManagerBoundary({
             ? {}
             : { parentExitDeadlineAt: Date.now() + options.systemdHandoffDeadlineMs }),
           ...commandFixture,
-          ...(options?.recoveryHang || options?.triageHang ? { recoveryTimeoutMs: 1000 } : {}),
+          // Triage hangs must reach the diagnostic cap without timing out healthy recovery.
+          ...(options?.recoveryHang ? { recoveryTimeoutMs: 1000 } : {}),
           recovery: { serviceRestartSafe: true, version: "1.0.0" },
           recoveryModulePath,
           commandArgv: [process.execPath, "-e", updaterScript],
--- a/src/process/supervisor/service-child-group-anchor.retirement.test.ts
+++ b/src/process/supervisor/service-child-group-anchor.retirement.test.ts
@@ -1,0 +1,164 @@
+import { spawn } from "node:child_process";
+import { randomUUID } from "node:crypto";
+import { Duplex } from "node:stream";
+import { describe, expect, it } from "vitest";
+import { runtimeProcessEntrypoints } from "../../infra/runtime-process-entrypoints.js";
+import {
+  resolveRuntimeWorkerArgv,
+  resolveRuntimeWorkerUrl,
+} from "../../infra/runtime-worker-url.js";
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9f474414d8ef0808afb23bb53eb25909ce3fb6a8](https://github.com/openclaw/openclaw/pull/143240/commits/9f474414d8ef0808afb23bb53eb25909ce3fb6a8)
- Landed commit: [ef37c4f3755d4a424c12530d9339d48f2fbcdc8d](https://github.com/openclaw/openclaw/commit/ef37c4f3755d4a424c12530d9339d48f2fbcdc8d)

PR #143256, opened 2026-09-09
feat(voice-call): add fixed realtime tool bindings

johnemau:
## What Problem This Solves

Routine read-only voice requests currently require a full supervisor consultation, adding avoidable latency.

## Why This Change Was Made

Add configured realtime tools that bind to fixed Gateway methods, carry trusted call context, and speak marked results exactly once. Bound calls are cancelled on barge-in, continuity reset, bridge replacement, or close; arbitrary Gateway dispatch is not exposed.

## User Impact

Plugins can serve bounded, low-latency voice reads while unsupported providers and unbound tools keep the completed-result fallback.

## Evidence

- 97 focused voice-call tests pass.
- Extension production typecheck and targeted lint pass.
- Lifecycle tests cover parallel calls, exact-response deduplication, barge-in, replacement, and close.

Generated with AI under the guidance of John.

### DIFF
--- a/docs/plugins/voice-call/realtime-and-streaming.md
+++ b/docs/plugins/voice-call/realtime-and-streaming.md
@@ -91,6 +91,33 @@ When a host tool run reports cancellation, the realtime model receives a
 cancelled result and the phone call stays open. Timeouts and other tool failures
 remain errors; ending the phone session suppresses pending consult results.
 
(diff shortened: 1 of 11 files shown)


johnemau:
Thank you for your consideration and the detailed review feedback. I'm closing this proposal while we consolidate the work in a downstream fork and address the remaining correctness and validation gaps.

Generated with AI under the guidance of John.

PR #143258, opened 2026-09-09
fix: Doctor source-isolation tests fail on tmpfs

steipete:
Related: #139428

<details>

Maintainers can edit this same-repository branch.

</details>

## What Problem This Solves

Resolves a problem where four Doctor source-isolation tests fail on Linux hosts whose temporary directory uses tmpfs. The source paths are correct, but a legitimate volatile-storage warning adds unrelated findings and repair effects to the fixtures' exact expected results.

## Why This Change Was Made

Supply stable ext4 mount metadata only for the exact `/proc/self/mountinfo` UTF-8 read during these four cases. All other reads retain their original arguments, and the spy restores in `finally`. The real detector, source isolation, repair mapping and every existing assertion remain intact. Dedicated volatile-filesystem tests are unchanged.

## User Impact

No production behavior changes. Developers can run these source-isolation tests on tmpfs-backed test environments while retaining coverage of actual volatile-storage warnings. This correctness repair adds 36 test lines; it makes no overall runtime reduction claim.

## Evidence

- Unchanged Linux/Node 24.19 contrast: four cases passed on ext4 and the same four failed on tmpfs, with only the anticipated extra findings and correct source paths.
- Final candidate on actual Linux tmpfs: all four repaired cases passed; all nine unchanged volatile-filesystem cases passed separately. Native process groups were joined and the isolated lease was cleaned up.
- Complete local owner and sibling suites: 52 cases passed before and after with identical ordered names and statuses.
- Exact inverse checks preserve all original test code. Focused forwarding checks retain path/options identity, propagated errors and restoration on success/failure.
- Full changed gate and fresh independent review passed. The initial implicit spy callback failed TypeScript's overload check; the final callback explicitly models Node's actual options types without casts or argument normalization.

### DIFF
--- a/src/commands/doctor-lint.state-isolation.test.ts
+++ b/src/commands/doctor-lint.state-isolation.test.ts
@@ -501,7 +501,24 @@ describe("doctor lint state isolation", () => {
             },
           ]);
           const stdout = vi.spyOn(process.stdout, "write").mockImplementation(() => true);
+          const readFileSync = fs.readFileSync;
+          const mountInfo = vi.spyOn(fs, "readFileSync");
           try {
+            // Source isolation is independent of the temporary directory's backing filesystem.
+            mountInfo.mockImplementation(
+              (target, options?: fs.ReadFileSyncOptions | BufferEncoding | null) => {
+                if (typeof options === "string") {
+                  if (target === "/proc/self/mountinfo" && options === "utf8") {
+                    return "22 1 0:21 / / rw,relatime - ext4 /dev/sda1 rw";
+                  }
+                  return readFileSync(target, options);
+                }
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [00fbdcd3c4abd54a09e7ba4b93d575b9ceef2535](https://github.com/openclaw/openclaw/pull/143258/commits/00fbdcd3c4abd54a09e7ba4b93d575b9ceef2535)
- Landed commit: [577492132500d5c5f66022c25023643e0554422b](https://github.com/openclaw/openclaw/commit/577492132500d5c5f66022c25023643e0554422b)

PR #143266, opened 2026-09-09
fix(cli): treat agent selection refusals as expected CLI conditions

steipete:
## What Problem This Solves

On a config with more than one agent and no explicit owner, most CLI commands report the missing `--agent` selection as a crash. Observed live on an isolated two-agent state:

```
$ openclaw skills list --json
{ "ok": false, "error": { "type": "cli_error", "message": "Multiple agents are configured, but the skills command has no explicit owner. Pass --agent <id>." } }
[openclaw] The CLI command failed.
[openclaw] Reason: Multiple agents are configured, but the skills command has no explicit owner. Pass --agent <id>.
[openclaw] Help: openclaw --help
```

`openclaw sessions list --json` printed the same class of refusal as a plain message, because the sessions command family wrapped it into an expected CLI error on its own. Every other surface that resolves an agent (skills, system, memory, and the rest of the `resolveAgentScope` callers) let the error reach the root handler unclassified, so operators were sent to a stack trace and `openclaw doctor` for a missing flag.

## Why This Change Was Made

Expected-condition classification for CLI errors lives in one owner, `isExpectedCliError` in `src/cli/failure-output.ts`; the root handlers in `src/entry.ts` and `src/index.ts` and every `rethrowExpectedCliError` / `runCommandWithRuntime` call site read it. The agent-selection refusal was the one operator-input error classified in a consumer instead of that owner.

- `src/cli/failure-output.ts`: `isAgentSelectionCliError` (lean name-based predicate, same shape as the gateway-auth and immutable-config predicates) joins `isExpectedCliError`. The producer message already names the surface and the `--agent` remedy.
- `src/commands/session-store-targets.ts`: the sessions family now rethrows the typed error with its session-store wording instead of flattening it into `ExpectedCliError`; other resolution errors keep the existing handoff.
- Commands that call `rethrowExpectedCliError` (`system`) or `runCommandWithRuntime` (`skills`) now hand the error to the root owner like every other expected condition, so their tests moved from "prints its own copy" to "rethrows"; the rendered output is covered once, in the failure-output table.

Production LOC: +8 / −5 across the two files (one predicate plus its invariant comment; the sessions wrapper lost the message flattening).

## User Impact

Every CLI command reports a missing agent selection the same way: the message with its `--agent` remedy, no stack-trace hint and no `openclaw doctor` suggestion. JSON mode still emits the same `cli_error` envelope. The sessions commands are unchanged in wording.

## Evidence

- Regression tests fail on pre-fix code and pass with the fix: `src/cli/failure-output.test.ts` (new "missing agent selection" row in the shared expected-condition table) and `src/commands/session-store-targets.test.ts` (typed rethrow keeps the session-store surface wording).
- Updated to the rethrow contract: `src/cli/skills-cli.commands.test.ts`, `src/cli/system-cli.test.ts`, and after the first full CI run also `src/cli/capability-cli.test.ts` (3 cases), `src/cli/channel-auth.owner.integration.test.ts` (2), `src/commands/sessions.default-agent-store.test.ts` (1) — every remaining test that encoded the old per-command rendering; sibling suites green: `directory-cli`, `export-trajectory`, `sessions-cleanup` (25 + 196 + 10 + 48 tests).
- `node scripts/check-changed.mjs`: all lanes ok (conflict markers, ratchets, format, plugin boundaries, wrapper shadowing, dead-export scans, core tsgo graph boundary, typecheck core, typecheck core tests).
- Codex autoreview (`--mode local`, P2, with the classifier, root handlers, command wrappers and producer supplied as context): scoped-clean, no accepted or actionable findings ("patch is correct (0.95)").
- Live proof (isolated two-agent state, built dist): `openclaw skills list --json` and `openclaw skills list` now print only the message with its `--agent` remedy (JSON mode keeps the same `cli_error` envelope); `openclaw sessions list --json` is unchanged; a gateway-unreachable error on `system event` still renders its own expected copy.

### DIFF
--- a/src/cli/capability-cli.test.ts
+++ b/src/cli/capability-cli.test.ts
@@ -1102,10 +1102,15 @@ describe("capability cli", () => {
       },
     });
 
-    await expect(runCapability("audio", "providers", "--json")).rejects.toThrow("exit 1");
-
-    expectRuntimeErrorContains("inference provider inspection has no explicit owner");
-    expectRuntimeErrorContains("Pass --agent <id> or set agents.defaults.systemAgent.agentId");
+    // Agent selection is an expected CLI condition rendered by the root failure
+    // owner; the command rethrows instead of printing its own copy.
+    await expect(runCapability("audio", "providers", "--json")).rejects.toMatchObject({
+      name: "AgentSelectionRequiredError",
+      message: expect.stringMatching(
+        /inference provider inspection has no explicit owner[\s\S]*Pass --agent <id> or set agents\.defaults\.systemAgent\.agentId/,
+      ),
+    });
+    expect(runtimeErrorMessages()).toEqual([]);
     expect(mocks.loadAuthProfileStoreForRuntime).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ed2d0c9d0976f576f9bef96f8b790c003cc14765](https://github.com/openclaw/openclaw/pull/143266/commits/ed2d0c9d0976f576f9bef96f8b790c003cc14765)
- Landed commit: [8a1710361efa5397d9c4138e290b1f5574262e9c](https://github.com/openclaw/openclaw/commit/8a1710361efa5397d9c4138e290b1f5574262e9c)

PR #143269, opened 2026-09-09
refactor(restart): keep continuation policy with update payloads

steipete:
## What Problem This Solves

Restart acknowledgement handling retained a helper whose session-routing argument did nothing. Its three scalar tests obscured the actual contract: routing a restart notice must not request another agent turn, while a successful update may carry an explicitly requested continuation.

## Why This Change Was Made

Remove the internal helper and unused input, keep `/restart`'s existing explicit `null` continuation, and put message trimming beside the update payload's existing normalized-success gate. Session, delivery and thread routing remain intact. The two negative subjects now exercise the payload owner; the existing pending-to-success case also verifies trimming of the explicit message.

The current stable package, `2026.9.3`, has 340 explicit export keys. Its actual 492 exported files contain no named export of this helper and no wildcard re-export, so this does not retire a public SDK contract. Sentinel types and persisted representations are unchanged.

Production delta: -21 lines (-17 excluding import reformatting). Test delta: -9 lines and one fewer case. Persistence, malformed-row, revision, lifecycle and restart-command coverage is retained.

## User Impact

Behavior is preserved: `/restart` reports its routed acknowledgement without starting an implicit agent turn; successful updates retain trimmed explicit continuation messages. Pending or unsuccessful updates remain continuation-free. There is no new configuration, database change, compatibility shim or test-only production seam.

## Evidence

- Untouched baseline: 61 tests passed across `restart-sentinel.test.ts`, `restart-sentinel.update-result.test.ts` and `commands-session-restart.test.ts`.
- Candidate: 60 tests passed on the same three files, with both native shards completing on their first attempt. The exact case comparison maps three removed scalar cases to two payload cases and the strengthened existing explicit-message case; 58 case identities remain.
- Deliberate owner-level controls confirm coverage: reintroducing an implicit session-based turn fails the session-route payload assertion, and removing trimming fails the existing success assertion. Both temporary faults were removed and the final source hashes verified.
- Normal changed checks and independent review are recorded with the frozen candidate before publication. No operator Gateway was restarted: this is a behavior-preserving owner/test consolidation proved through the real payload, typed persistence and command-handler suites.

### DIFF
--- a/src/auto-reply/reply/commands-session.ts
+++ b/src/auto-reply/reply/commands-session.ts
@@ -23,7 +23,6 @@ import { logVerbose } from "../../globals.js";
 import { getSessionBindingService } from "../../infra/outbound/session-binding-service.js";
 import type { SessionBindingRecord } from "../../infra/outbound/session-binding-service.js";
 import {
-  buildRestartSuccessContinuation,
   clearRestartSentinel,
   formatDoctorNonInteractiveHint,
   type RestartSentinelPayload,
@@ -74,7 +73,7 @@ function buildRestartCommandSentinel(params: HandleCommandsParams): RestartSenti
     deliveryContext,
     threadId,
     message: "/restart",
-    continuation: buildRestartSuccessContinuation({ sessionKey }),
+    continuation: null,
     doctorHint: formatDoctorNonInteractiveHint(),
     stats: {
       mode: "gateway.restart",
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0bea20fdfe0a0c1a44d2b4af45230eb1bb01672e](https://github.com/openclaw/openclaw/pull/143269/commits/0bea20fdfe0a0c1a44d2b4af45230eb1bb01672e)
- Landed commit: [7438ed26f4329d882e985dcb052c4647a6e7d901](https://github.com/openclaw/openclaw/commit/7438ed26f4329d882e985dcb052c4647a6e7d901)

PR #143270, opened 2026-09-09
fix(plugin-sdk): retain prepared models through host cleanup

steipete:
Related: #140674. Internal caller migrations landed in #143302, #143307, and #143308.

## What Problem This Solves

Resolves a problem where plugins receive an executable prepared model after its preparation lease has already been released. Completion callbacks and nested preparation can then outlive the model's owner, particularly during cancellation or host shutdown.

## Why This Change Was Made

Keep the existing bare-result SDK API under its SDK host. Capture the host before lazy preparation, retain the successful model, and track actual completion and callback work through host closure. Re-enter that host for nested preparation without replacing Gateway request authority.

With all three internal callers using acquisition, delete the obsolete core bare adapter. Shared parameter and result types preserve the public fingerprint, optional selection fields, and deprecated parameter; the SDK runtime import remains lazy. Core preparation tests now release acquired results. The distinct direct preparation API remains unchanged.

## User Impact

Plugins keep their existing prepared-model API and gain reliable host-owned cleanup. Host closure joins pending work and refuses new requests afterward. Provider selection, authentication checks, and public result shapes are preserved.

## Evidence

The original native SDK regressions and a separately reproduced nested-callback host-context failure pass after repair. Final validation passed 166 focused tests, the unchanged export-collision guard, all selected changed-file gates, fresh isolated Codex review through P2, and a full build in 248.17 seconds.

A plain Node run of the built public SDK used actual ModelRegistry instances, a synthetic provider, and loopback HTTP. It exercised nested preparation in a real response callback, prompt cancellation while that callback remained held, host drainage, independent request stores, base-catalog reuse, post-close refusal, and fresh preparation after the last host closed. It also verified that the built core exposes acquisition and no longer exports the obsolete bare adapter.

That strict, joined run completed in 2.495 seconds, observing two base catalogs, four request forks, and three HTTP calls. All 21,093 build artifacts remained unchanged; 2,880 resolved modules had no application TypeScript/source fallback. Preparation samples were 145.059 / 3.162 / 5.768 ms for initial, retained, and fresh-after-closure preparations; these are local observations, not a comparative speedup benchmark.

After rebasing onto the three landed caller migrations, all 14 source files match the reviewed proof. The collision guard and 42 SDK composition tests passed again. This does not activate general run-registration physical disposal or claim a paid-provider or operator-Gateway test.

### DIFF
--- a/docs/plugins/sdk-runtime/models.md
+++ b/docs/plugins/sdk-runtime/models.md
@@ -138,3 +138,16 @@ Call a model, resolve model-selection policy, and resolve provider auth without 
 
   </Accordion>
 </AccordionGroup>
+
+## Prepared completion SDK compatibility
+
+Prefer `api.runtime.llm.complete` for new plugin code. Existing callers of
+`openclaw/plugin-sdk/simple-completion-runtime` can continue to prepare a model
+with `prepareSimpleCompletionModelForAgent` and execute it with
+`completeWithPreparedSimpleCompletionModel`.
+
+These prepared results have no release method. Their original Gateway or CLI
+host retains the model resources until shutdown; standalone callers retain them
+for the process lifetime. A closed host rejects new preparation and execution.
+Shutdown waits for accepted provider callbacks and cancellation work before
+releasing the prepared resources, even when the completion has already returned.
--- a/src/agents/simple-completion-execution.ts
+++ b/src/agents/simple-completion-execution.ts
@@ -13,6 +13,7 @@ import type { ThinkLevel } from "../auto-reply/thinking.js";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import {
   bindModelLlmRuntime,
+  getModelCompletionOwner,
   getModelCompletionTransport,
   getModelLlmRuntime,
 } from "../llm/model-runtime-binding.js";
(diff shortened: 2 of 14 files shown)


steipete:
Maintainer decision: retain the documented host boundary. A prepared model owned by a closed Gateway or CLI host must reject new execution; callers prepare again under the current open host. Keeping the old model usable would retain retired resources or silently change its model/authentication ownership.

The public preparation/result types remain unchanged, and standalone callers retain process-lifetime ownership. The host-lifetime behavior is documented in `docs/plugins/sdk-runtime/models.md`; the compiled public-SDK proof covers zero requests after closure and successful preparation under a fresh host. This accepts the compatibility tradeoff identified in the review; no compatibility shim or code change is requested.

steipete:
Merged via squash.

- Prepared head SHA: [731e23f098ce5e402ed4da2d5db4ab6917054a95](https://github.com/openclaw/openclaw/pull/143270/commits/731e23f098ce5e402ed4da2d5db4ab6917054a95)
- Landed commit: [51b311ba2c5656c9a8a90b8446720d910f0eadae](https://github.com/openclaw/openclaw/commit/51b311ba2c5656c9a8a90b8446720d910f0eadae)

PR #143238, opened 2026-09-09
fix(agents): drop the blanket credential prompt that blocked owner-authorized logins

steipete:
Closes #128076
Related: #120728 #120656 #128077 #142010

## What Problem This Solves

Fixes an issue where agents would refuse owner-authorized work that needs a credential the owner already supplied: typing a password into a login form through the browser tool, running a command with a workspace credential file, or completing setup for a remote operator who has no reachable Control UI. The refusal came from a blanket prompt block added in #120728, which told every agent to keep credentials out of "commands, arguments, URLs, tool arguments, and shell text". Operators running unattended browser logins reported their previously working flows breaking after upgrading; #128076/#128077 and #142010 report the same class.

## Why This Change Was Made

The block was a prompt-level patch for a tool-routing miss. The original report (#120656) came from the local embedded TUI, which strips the `openclaw` and `gateway` tools, so the model had no masked handoff to call and improvised "paste the token here". The masked-entry surfaces already carry their own guidance in their tool descriptions (`secrets`, `connect_channel`, `open_setup`), and the block duplicated that guidance while adding only the harmful "never in tool arguments" rule.

This change deletes the blanket contract from the core system prompt, Codex developer instructions, Copilot guidance, and the system-agent prompts, and keeps credential routing where the surfaces live: one clause on the `openclaw` setup tool and delegate tool descriptions. What remains shared is deliberately narrow and lives in `src/agents/credential-safety-prompt.ts` under the shipped SDK name `buildCredentialSafetyPrompt`: the private login-code handoff line for group conversations (the owner-requested behavior from #139196), plus a terminal setup route (`openclaw channels add <channel>` / `openclaw configure`) emitted only when neither the `openclaw` nor the `gateway` tool is callable, which is the embedded/local TUI case. The helper takes `{ controlToolsAvailable }`; the legacy string argument is accepted and ignored under a dated compat record through 2026-11-30, so plugins built against 2026.9.3 keep loading. The egress usage line stays with the `secrets` tool description. Out of scope: transcript redaction (#104992), the browser tool, configuration, and in-process delegation for `--local` mode (follow-up).

## User Impact

- Agents complete owner-authorized tasks with owner-supplied credentials again, including browser form fills and commands that read workspace credential files, without echoing the values into chat.
- Setup flows still collect channel tokens, API keys, and passwords through masked entry, and the model is told so where it matters: on the setup tools themselves.
[...]

Issue #120656, opened 2026-08-08
Local TUI onboarding can ask users to paste a Telegram bot token into the chat transcript


Sedrak-Hovhannisyan:
## Summary

A fresh local embedded TUI onboarding conversation can instruct a user to paste a Telegram bot token directly into the chat, then only retract that advice after the user explicitly questions it. This is unsafe guidance for a long-lived secret and contradicts the assistant’s preceding promise to handle system-related work through OpenClaw.

## Environment

- OpenClaw `2026.8.1 (7f9364c)`
- Linux x64, Node `24.18.0`
- local embedded TUI, agent `main`, session `main`
- model displayed by the TUI: `gpt-5.6-sol`, thinking `medium`

## Reproduction

1. Start a fresh local embedded TUI session.
2. Complete the basic identity/personality prompt.
3. Ask: `let's connect telegram`.
4. The agent instructs the user to use BotFather and then says to paste the bot token into the chat.
5. Ask whether pasting the token in chat is safe. Only then does it say not to do that, and tells the user to run the guided CLI setup.

## Actual

The first answer invites disclosure of a credential in an agent transcript. The later correction does not undo exposure if the user already complied.

## Expected

Any built-in onboarding/system guidance should have a non-bypassable rule: never request a bot token, API key, password, or OAuth secret in a chat transcript. For Telegram, explain that `openclaw channels add telegram` opens guided setup and that the token must be entered only into its masked prompt (or another dedicated secret input), never pasted into chat or a command line.

## Why this matters

A non-technical user is likely to regard the agent as the trusted setup path. The system needs to be safe before the user knows enough to challenge a bad instruction.

No token, password, server address, account name, or private log is included in this report.

PR #143273, opened 2026-09-09
feat(tui): answer agent questions in gateway and local modes

steipete:
Related: #120656 #143238

## What Problem This Solves

Fixes an issue where an agent calling `ask_user` in the terminal UI left the user with nothing to answer: the TUI never rendered the question, only the raw tool call, and in local mode (`openclaw chat`, `openclaw tui --local`) the tool's question RPCs went to a Gateway that usually is not running, so the request failed outright. The `secrets` tool had the same gap, which is part of why the local TUI could only fall back to asking for credentials in chat (#120656). Ordinary answers typed while a question was pending in collect or followup queue mode also waited behind the blocked run.

## Why This Change Was Made

Both TUI backends now share one session-scoped question controller. Gateway mode uses the existing `question.*` RPCs and events under the scopes the TUI already holds. Local mode serves the same five methods in-process through an embedded question broker wrapping the Gateway's `QuestionManager`, registered like the existing embedded plugin-approval broker, and the question dispatcher routes there only when embedded mode has a broker. A typed reply is claimed by the pending question through the active run's guarded input owner before queue policy runs, which is what removed the collect/followup deadlock. Masked answers keep their exact bytes.

Store-bound `secrets` requests in local mode return a clear blocker to the model instead of collecting a value the process cannot commit, because the store write and runtime refresh are Gateway-owned. No protocol, schema, config, or Control UI changes.

## User Impact

- `ask_user` prompts render in the TUI in both modes: one question at a time with a stepper, arrow or number selection, Other… free text, multi-select toggles, Skip, and the time remaining.
- Esc collapses the prompt and returns the composer with a pending-question status line; `/question` reopens it; a plain reply still answers the pending question.
- Pending questions are restored after reconnects and session switches; questions answered elsewhere or expired close automatically.
- Gateway-connected `secrets` requests use a masked input that never touches the chat log or input history, showing the entry name, reason, and proposed allowed hosts (accepted as proposed; edit hosts in the Control UI).
- Local mode: store-bound `secrets` requests explain that they need a running Gateway; plain masked questions work.

## Evidence

Real-model live run on an isolated state directory and free port (never the operator's Gateway), model `anthropic/claude-fable-5-1`, driven through tmux; captures are text only and contain no credentials.

Local mode, prompt rendered from a real `ask_user` call:

```text
 openclaw tui - local embedded - agent main - session main
 ⠧ running • 7s | local ready
 Question pending · 898s · /question to open
───────────────Question 1/1 · Color                                                            ───────────────
               Which color do you prefer?
───────────────Expires in 898s                                                                 ───────────────
               › 1. Blue
                 2. Green
                 3. Other…
                 Confirm answer
                 Skip
               ↑/↓ or numbers select · Enter confirm · Esc collapse
```

After `1` + Enter: `Question: answered.` followed by the model's `You chose Blue.` A second question was collapsed with Esc (status line `Question pending · 899s · /question to open`), answered by typing `Large` into the composer, and the model replied `You chose Large.`

Gateway mode (`tui --url ws://127.0.0.1:<port>`): same prompt, answered `2` + Enter, model replied `You chose Green.` Then a real `secrets` request rendered the masked prompt:

```text
               Question 1/1 · API key
               Provide the secret for LIVE_TEST_TOKEN.
               Secret store entry: LIVE_TEST_TOKEN
               Reason: TUI live test
               Allowed hosts (accept as proposed): example.com
                 1. Other…
               > ••••••••••••••••••••
                 Confirm answer
                 Skip
               Enter next/submit · Tab choices · Esc collapse
               Masked entry · Answer through this prompt, not the composer.
```

After Enter: `Question: answered.` and the model replied `The credential is stored as LIVE_TEST_TOKEN with allowed host example.com.` The typed value did not appear in any capture (checked by grep across all frames).

Automated proof: `node --import ./scripts/tsx.mjs scripts/dev/tui-questions-proof.ts` runs an isolated Gateway plus TUI and a local TUI with a scripted OpenAI-compatible model, verifying option resolution and the Esc/plain-reply continuation from the model's request bodies.

Tests: 1,893 tests across 62 files in the TUI, infra, harness, tools, runner, and Gateway question suites, plus 102 PTY tests. The exact-secret-byte cases fail on the original `QuestionManager` and pass with the fix. `node scripts/check-changed.mjs` passed all selected gates; `pnpm build` clean; docs mdx, link audit, glossary, and markdownlint pass. Codex autoreview (P0–P2): scoped-clean.

Known limits: the overlay host composites prompts over the chat log, so long background lines show at the overlay margins, the same as the existing plugin-approval overlay; secret input is single-line (use the CLI `--value-file` route for multiline values); local-mode questions do not survive exiting the TUI.

Overlay update: question prompts now use Pi TUI's supported `width: "100%"` option, resolving the background-chat margin limitation noted above.

## Data-model compatibility evidence

The serialized writes in `scripts/dev/tui-questions-proof.ts` are isolated proof fixtures and requested evidence artifacts, not a runtime data-model change. The script [creates its own temporary state](https://github.com/openclaw/openclaw/blob/50e0096b033cd526b041544a3b679cbe8146c7d2/scripts/dev/tui-questions-proof.ts#L214) and [removes it after verified child cleanup](https://github.com/openclaw/openclaw/blob/50e0096b033cd526b041544a3b679cbe8146c7d2/scripts/dev/tui-questions-proof.ts#L519), retaining only terminal frames and answer witnesses in the requested output directory. The repeatable proof passed Gateway selection, local selection, and local Esc/plain-reply continuation.

Runtime questions remain in [QuestionManager's in-memory map](https://github.com/openclaw/openclaw/blob/50e0096b033cd526b041544a3b679cbe8146c7d2/src/gateway/question-manager.ts#L109). The embedded broker reuses that manager and [rejects store-bound requests](https://github.com/openclaw/openclaw/blob/50e0096b033cd526b041544a3b679cbe8146c7d2/src/infra/embedded-question-broker.ts#L68). The diff from the rebased parent leaves `packages/gateway-protocol`, `src/config`, `src/secrets`, and `src/gateway/server-methods/question.ts` unchanged. Existing persisted formats and the Gateway secret-store write path are preserved; this change requires no data migration or upgrade conversion.

### DIFF
--- a/docs/cli/tui.md
+++ b/docs/cli/tui.md
@@ -90,6 +90,11 @@ Aliases: `openclaw chat` and `openclaw terminal` invoke this command with
 - Local mode adds `/auth [provider]` to the TUI command surface.
 - Plugin approval gates still apply in local mode: tools that require approval
   prompt for a decision in the terminal, nothing is silently auto-approved.
+- [`ask_user` questions](/web/tui#questions) appear as interactive prompts in
+  both modes. Esc collapses a prompt without answering; `/question` reopens it.
+- Gateway mode accepts [`secrets`](/tools/secrets) requests in a masked input
+  with a read-only allowed-host list. Local mode cannot fulfill store-bound
+  requests; use `openclaw secrets store` or the Control UI with a running Gateway.
 - Session [goals](/tools/goal) appear in the footer and can be managed with
   `/goal`.
 
--- a/docs/tools/ask-user.md
+++ b/docs/tools/ask-user.md
@@ -23,6 +23,11 @@ You can answer from any supported conversation surface:
   multi-question prompts, the panel shows one question at a time and advances
   through a short stepper. After resolution, the panel closes and the chat
   keeps only a compact answer summary.
+- The TUI shows a question prompt in both Gateway and local modes. Use arrow
+  keys or number keys to choose an option, **Other…** to type an answer, or
+  **Skip**. Multi-select prompts let you toggle choices before confirming;
+  multi-question prompts advance one question at a time. Press Esc to return
+  to the composer without answering, then `/question` to reopen the prompt.
 - Telegram renders each choice as a full-width native button for one
   single-select question. **Other…** switches to Telegram's reply input without
   resolving the question.
@@ -35,7 +40,7 @@ You can answer from any supported conversation surface:
 
 Questions from a standalone [attached MCP client](/cli/attach) do not carry an
 OpenClaw run's creator binding. Answer those using the question controls in the
-Control UI or native app, not an ordinary channel message.
+Control UI, TUI, or native app, not an ordinary channel message.
 
 OpenClaw always enables a free-text **Other** answer. The agent must not add an
 `Other` option to the authored option list.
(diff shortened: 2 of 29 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [50e0096b033cd526b041544a3b679cbe8146c7d2](https://github.com/openclaw/openclaw/pull/143273/commits/50e0096b033cd526b041544a3b679cbe8146c7d2)
- Landed commit: [62ce5547234b731f0f4404339e73b0791fa2afea](https://github.com/openclaw/openclaw/commit/62ce5547234b731f0f4404339e73b0791fa2afea)

PR #133843, opened 2026-08-31
fix(secrets): warn on shared auth-profile plaintext in configure

SunnyShu0925:
Refs https://github.com/openclaw/openclaw/issues/133296

## What Problem This Solves

Resolves a problem where operators running `secrets configure` see no indication that shared auth profiles still contain plaintext credentials outside the selected agent's migration scope.

The warning-only revision is accepted independently of the earlier shared-target proposal. That proposal remains outside this PR; the underlying migration issue stays open.

## Why This Change Was Made

The diagnostic reads the existing raw shared-store row and uses the existing credential iterator and `coerceSecretRef` classification. This includes plaintext left beside a sibling reference while excluding supported inline environment references.

The warning goes to stderr, preserving parseable stdout for `secrets configure --skip-provider-setup --json`, including the existing no-candidate failure. It appears before that failure without introducing any credential writes or migration.

There are no schema, protocol-version, target-contract, configuration-option, or persistent-store semantic changes. Existing v1 `agentId` targets still mean agent-local writes or creation; plan/apply behavior is unchanged. This PR does not approve or implement a shared-target design.

The remaining automated data-model flag does not apply to this revision: production schema and persistence writers are unchanged, and test fixture inserts only seed disposable SQLite state. The diagnostic uses the existing read-only accessor; it adds no migration to validate.

## User Impact

Operators are directed to `openclaw secrets audit` when shared plaintext remains. Existing agent-local migration behavior is preserved, and JSON consumers do not receive the new human-readable warning on stdout.

## Evidence

- Integrated current main with a merge retaining the contributor's commit ancestry.
- Added an isolated actual-entrypoint test using real Clack output, the real CLI JSON failure wrapper, and real SQLite. It separately captures stdout/stderr, restores TTY state, verifies unchanged credentials/config, and checks that an absent shared auth row stays absent. The existing test helper isolates HOME and ambient OpenClaw state.
- Local bounded red/green proof executed the exact original and repaired warning statements against Clack 1.7.0. Before: warning text contaminated stdout and JSON parsing failed. After: stdout parsed and the warning appeared on stderr. This producer proof used a synthetic JSON envelope, not the configure entrypoint or persistence.
- Targeted formatting and `git diff --check` passed. Changed-lane classification selects core production and core tests.
- Full-candidate independent P2 review completed with no actionable findings across all three review passes. The two-line test-contract correction also received a clean focused P2 review.
[...]

Issue #133296, opened 2026-08-30
[Bug]: Secrets apply routes shared auth-profile refs to the agent store


nasacontrol-hq:
## Bug type

Behavior bug (incorrect output/state without crash)

## Beta release blocker

No

## Summary

`secrets configure` does not preserve whether an auth-profile candidate came from the per-agent or canonical shared store, so `secrets apply` can route a selected shared profile to the agent database instead of the shared state database.

## Steps to reproduce

1. Check out current `main` at `d4f6966dbf3b7448017dd14621fe3cc873269d78`.
2. Under a disposable `OPENCLAW_STATE_DIR`, seed canonical shared ownership and an API-key profile in the schema-v13 `config_machine_state` row `authProfiles.store`; leave the selected agent database absent.
3. Build a SecretRef plan target for that shared profile with `agentId: "main"` and `authProfileStore: "shared"`.
4. Run `secrets apply --from <plan> --json` with an isolated environment provider.
5. Observe that `changedFiles` names `agents/main/agent/openclaw-agent.sqlite`; the canonical shared state database is not the target.

A focused regression on unmodified current main fails with the same path mismatch: expected the canonical state database, received the per-agent database.

## Expected behavior

A shared auth-profile selection should serialize explicit shared ownership, apply the SecretRef to the canonical shared state database, remove the plaintext field there, leave the per-agent database untouched, and report the shared database in `changedFiles`.

Per-agent selections should continue targeting the agent database, and plans that predate the ownership field should retain their existing agent-local behavior.

## Actual behavior

Configure candidates and plans do not carry an auth-store owner. Apply resolves every auth-profile target from `agentId`, so a profile selected from the shared store can be written to the per-agent database. In the isolated regression, `changedFiles` reports the agent database rather than the canonical shared state database.

## OpenClaw version

Current main package version `2026.8.1`, commit `d4f6966dbf3b7448017dd14621fe3cc873269d78`.

## Operating system

macOS 26.6.2 (25G83), arm64.

## Install method

Source checkout with pnpm; built CLI invoked against a disposable state root.

## Model

Not applicable (local `secrets` CLI workflow; no model inference).

## Provider / routing chain

Not applicable (local environment SecretRef fixture; no provider network request).

## Additional provider/model setup details

The reproduction uses only disposable fixture databases and placeholder values. No live credentials or production state are involved.

## Logs

```shell
FAIL  secrets apply > writes a SecretRef to the existing shared auth-profile owner
[...]

PR #143275, opened 2026-09-09
fix(secrets): route auth-profile targets to the shared store owning the profile

### LINKED ISSUE #143173: Legacy auth-profiles.json presence blocks ALL provider auth for an agent, not just the affected provider
## Summary

When a legacy-format credential file (`~/.openclaw/agents/<agent>/agent/auth-profiles.json`) exists next to an agent's SQLite auth-profile store (`auth_profile_store` table in `openclaw-agent.sqlite`), and that SQLite store's `profiles` object happens to be empty, `AuthProfileMigrationRequiredError` (`AUTH_PROFILE_MIGRATION_REQUIRED`) is thrown and blocks **every** provider's auth for that agent — not just the provider the legacy file is actually about. This can take down all chat functionality (in our case, `litellm`) even though `litellm` itself has nothing to do with the stale legacy file.

Version: openclaw 2026.9.3

## Root cause (from reading the installed package source)

In `legacy-source-diagnostic.ts` (`assertAuthProfileMigrationCandidates`):

```js
function assertAuthProfileMigrationCandidates(params) {
  assertAuthProfileMigrationStateAtDatabasePath(params.databasePath);
  const sources = params.candidates.filter((source) => isCredentialSource(source) && fs.existsSync(source.path));
  if (sources.length === 0) return;
  if (params.hasCredentials()) {
    warnLegacyAuthProfileSourcesIgnored({ databasePath: params.databasePath, sources });
    return; // only warns
  }
  const migrationError = new AuthProfileMigrationRequiredError({ databasePath: params.databasePath, sources });
  migrationRequiredByDatabase.set(params.databasePath, migrationError);
  throw migrationError; // blocks the WHOLE agent's auth
}
```

`hasCredentials()` checks whether the SQLite store has *any* profile at all for that agent — it is not scoped to the specific provider(s) referenced by the legacy file. So:

- If the store already has at least one profile (for any provider), a stale legacy file for a *different* provider is harmless (warn-only).
- If the store happens to be empty at that exact moment (e.g. right after `openclaw models auth logout` on the only profile that existed, or right after a `doctor --fix` migration that hasn't repopulated it yet), the presence of *any* legacy file for *any* provider throws and blocks auth for every other provider too.

## How we hit this in practice

We had a real, dormant `anthropic-vertex` provider defined in `models.json` (Claude via Vertex AI), authenticated via a local `gcloud` ADC/service-account credential. `vertex-adc.ts`'s token-refresh path (`refreshGoogleVertexAuthorizedUserAccessToken` / `resolveGoogleVertexAccessTokenViaGoogleAuth`) still persists refreshed tokens through the legacy JSON-file format instead of the SQLite `auth_profile_store`, even though the rest of the credential system has moved to SQLite. Every time that token refreshed (no fixed interval — anywhere from under a minute to several hours), it recreated `auth-profiles.json`, and if the SQLite store was empty at that instant, our unrelated `litellm` provider's auth broke completely until `openclaw doctor --fix` was run again to re-migrate it.

`openclaw models auth logout google-vertex:default/:models-json` did not fix this — the ADC-driven writer just recreated the profile. The only way we found to stop it for good was removing the `anthropic-vertex` provider entirely from `models.json` so the vertex-adc refresh path never fires again.

## Suggested fixes (two independent issues, either would help)

1. Scope the migration-required check to the specific provider(s) referenced by the legacy source, not the whole agent's auth. A stale/regenerating legacy file for provider A should never be able to block provider B's auth.
2. Have the Vertex ADC token-refresh path (`vertex-adc.ts`) persist through the SQLite `auth_profile_store` like other providers, instead of the legacy JSON file, so it stops re-triggering the legacy-migration-required code path at all.

## Environment

- openclaw 2026.9.3, installed via npm, running as a systemd `--user` service on Linux.
- Provider triggering the legacy writes: Vertex AI via `authorized_user`/service-account ADC (`gcloud auth application-default login` style credential), referenced only via a dormant `anthropic-vertex` entry in `models.json` (Claude Opus/Sonnet), not in the active agent's default/fallback model list.

soroush5:
Closes #143173

## What Problem This Solves

Fixes an issue where users running `openclaw secrets apply` for an auth profile would silently keep the live plaintext key: with `state-db` shared-store ownership (post-upgrade migration), the profile lives in the shared store while the per-agent store is empty, but apply wrote the SecretRef into the per-agent store — so `secrets audit` kept flagging `PLAINTEXT_FOUND` and the runtime kept reading the old key.

## Why This Change Was Made

The plan-target router always resolved an `agentId` to the per-agent store and never consulted store ownership, while the audit side already reads whichever store owns the profile — so apply and audit disagreed. Now, when the target profile is absent from the agent store but present in the shared store under `state-db` ownership, the write is routed to the owning shared store. Agent-owned profiles route exactly as before.

## Validation

- New regression test replicating the issue shape (empty agent store + state-db shared profile): fails before the fix, passes after.
- Full `src/secrets/apply.test.ts`: 39 passed. `tsgo --noEmit` clean for the touched files, `oxfmt`/`oxlint` clean, `git diff --check` clean.

## Evidence

- New regression test `routes auth-profiles targets to the shared store owning the profile` in `src/secrets/apply.test.ts` replicates the issue shape (empty per-agent store + `state-db` shared profile): it **fails** on `main` and **passes** with this fix.
- Full `src/secrets/apply.test.ts` shard: **39 passed** (`vitest --config test/vitest/vitest.secrets.config.ts`).
- `tsgo --noEmit` clean for the touched files; `oxfmt`/`oxlint` clean; `git diff --check` clean.

### DIFF
--- a/src/secrets/apply.test.ts
+++ b/src/secrets/apply.test.ts
@@ -692,6 +692,84 @@ describe("secrets apply", () => {
     });
   });
 
+  it("routes auth-profiles targets to the shared store owning the profile", async () => {
+    // Issue #143173: with state-db shared ownership after the upgrade
+    // migration, the live profile sits in the shared store while the
+    // per-agent store is empty. Applying a SecretRef must replace the
+    // plaintext where it lives instead of writing the ref to the agent
+    // store and leaving the live key behind.
+    await writeJsonFile(fixture.authStorePath, { version: 1, profiles: {} });
+    const stateDatabase = openOpenClawStateDatabase({ env: fixture.env }).db;
+    stateDatabase
+      .prepare(
+        `INSERT INTO config_machine_state (state_key, value_json, updated_at_ms)
+         VALUES ('auth.sharedStore', ?, 1)`,
+      )
+      .run(JSON.stringify({ location: "state-db" }));
+    stateDatabase
+      .prepare(
+        "INSERT INTO config_machine_state (state_key, value_json, updated_at_ms) VALUES (?, ?, 1)",
+      )
+      .run(
+        "authProfiles.store",
+        JSON.stringify({
+          version: 1,
+          profiles: {
+            "openai:default": {
+              type: "api_key",
+              provider: "openai",
+              key: "sk-ope...text", // pragma: allowlist secret
+            },
+          },
+        }),
+      );
+    noteCommittedSharedAuthStoreOwnership({ location: "state-db" }, fixture.env);
+    const plan: SecretsApplyPlan = {
+      version: 1,
+      protocolVersion: 1,
+      generatedAt: new Date().toISOString(),
+      generatedBy: "manual",
+      targets: [
+        {
+          type: "auth-profiles.api_key.key",
+          path: "profiles.openai:default.key",
+          pathSegments: ["profiles", "openai:default", "key"],
+          agentId: "main",
+          ref: OPENAI_API_KEY_ENV_REF,
+        },
+      ],
+      options: {
+        scrubEnv: false,
+        scrubAuthProfilesForProviderTargets: false,
+        scrubLegacyAuthJson: false,
+      },
+    };
+
+    const result = await runSecretsApply({ plan, env: fixture.env, write: true });
+    expect(result.changed).toBe(true);
+
+    expect(readPersistedSharedAuthProfileStoreRaw(fixture.env)).toMatchObject({
+      profiles: {
+        "openai:default": {
+          type: "api_key",
+          provider: "openai",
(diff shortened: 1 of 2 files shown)


soroush5:
Thanks for the thorough review — you're right on both counts. Inferring shared intent from profile presence can't distinguish migration from an intentional agent-local override, and the v1 local-creation contract from #133843 must stay intact. The changedFiles misreport alone disqualifies this approach either way.

Closing this in favor of the explicit shared-target design under #133296 — happy to rework or test that direction once the target contract is agreed.

PR #95757, opened 2026-06-22
fix(memory): respect QMD timeout for memory_search

bdjben:
## What Problem This Solves

Fixes #91947.

QMD-backed `memory_search` can legitimately take longer than the historical fixed 15-second whole-tool deadline, even though QMD already exposes `memory.qmd.limits.timeoutMs`. As a result, configured longer QMD searches were aborted before their QMD-owned budget elapsed.

## Why This Change Was Made

The final implementation assigns deadlines at their owner boundaries:

- `memory.qmd.limits.timeoutMs` governs only the actual QMD `query` / `search` / `vsearch` subprocess or mcporter call.
- The existing 15-second default remains active for manager setup, QMD pre/post-command maintenance, builtin search, wiki/supplement work, zero-hit sync wait, visibility filtering, and transient cleanup.
- QMD-to-builtin fallback receives a fresh bounded 15-second setup/search budget, and failed QMD retirement no longer delays it.
- Caller cancellation propagates through the tool wrapper and active search subprocess while preserving the original reason and avoiding cooldown.
- Deadline timers/listeners and late one-shot managers are cleaned up.

This PR began with @bdjben's focused fix and local 23.9-second QMD proof. The maintainer repair preserves that credit while removing the broader whole-manager exemption and QMD close override from the earlier patch. The final test mock forwards the private symbol-keyed deadline option without importing the production deadline module, keeping current-main debug typing at its owner and avoiding a recurring merge conflict.

## User Impact

Users who intentionally raise `memory.qmd.limits.timeoutMs` can now complete slow QMD-backed agent memory searches beyond 15 seconds. Non-QMD work and QMD-to-builtin fallback remain bounded by the existing default deadline, so raising the QMD timeout does not extend setup, wiki, supplement, builtin, or cleanup work.

No config, schema, storage, protocol, plugin SDK, or dependency surface is added.

## Evidence

- Verified current-main PR head: `dcd0e5040012cc67fee63204a287c56edb00ad8a`, including base `ad1b4d9731709430757c8776ff8e99602589d8ec`.
- Exact-head Testbox `tbx_01kxd3jvggptc47g9yanc6772p`, Actions run [29229938011](https://github.com/openclaw/openclaw/actions/runs/29229938011): 230/230 passed across deadline, tools, search manager, QMD manager, and host-process tests.
[...]

PR #143276, opened 2026-09-09
fix(memory): preserve managed local service startup budget

### LINKED ISSUE #143169: memory_search aborts managed local embedding cold start before ready timeout
## Bug type

## Beta release blocker

No

## Summary

`memory_search` aborts a configured managed local embedding service during a cold start when service readiness takes longer than the fixed 15-second whole-tool deadline.

## Steps to reproduce

1. Configure an OpenAI-compatible embedding provider under `models.providers.<id>` with `localService` and a `readyTimeoutMs` greater than 15 seconds.
2. Configure memory search to use that provider.
3. Let the managed service stop according to `idleStopMs`.
4. Invoke `memory_search` while the service is cold.
5. Observe that the tool aborts after 15 seconds even though the service is still within its configured readiness budget.

## Expected behavior

Managed local-service startup should use the existing `localService.readyTimeoutMs` budget. Once the service is ready, the normal bounded memory-search deadline should continue to govern search work. Caller cancellation should remain immediate.

## Actual behavior

The outer `memory_search` deadline aborts the signal passed into local-service acquisition after 15 seconds, before the managed service can become ready. The tool returns `memory_search timed out after 15s` and pauses retries even though a later warm search succeeds.

## OpenClaw version

2026.9.2; current `main` remains affected at `9956abfbe30c914443fc3467f5472c07ddb68c5a`.

## Operating system

Debian 12 arm64 on Raspberry Pi.

## Install method

npm global for the observed deployment; pnpm source checkout for the proposed regression test.

## Model

EmbeddingGemma 300M GGUF served by a local `llama-server`.

## Provider / routing chain

`memory_search` -> Memory Core -> OpenAI-compatible embedding provider -> configured managed `localService` -> local `llama-server`.

The provider uses the existing managed-local-service configuration. No remote embedding API or credentials are involved. `readyTimeoutMs` allows the measured startup, but the separate Memory Core deadline currently aborts first.

## Logs

```shell
```

With a temporary local build patch that changed only the outer deadline to 60 seconds, the same cold search completed successfully in 38.9 seconds. After the service was warm, the search completed in 6.7 seconds. The index was valid and unchanged in both cases.

Source path on current `main`:

- `extensions/memory-core/src/memory/search-deadline.ts` sets `DEFAULT_MEMORY_SEARCH_TIMEOUT_MS = 15_000`.
- `src/plugins/openai-compatible-embedding-provider.ts` passes the search-owned abort signal to `acquireLocalService` before issuing the embedding request.
- `src/agents/provider-local-service.ts` already owns and enforces `localService.readyTimeoutMs`, but the outer abort signal can terminate that wait first.

Related context: #91947 and #95757 fixed the analogous QMD-owned search phase without removing the default bound from unrelated work. The managed-local-service cold-start phase remains affected.

## Impact and severity

Affected: Operators using managed local embedding services whose cold start exceeds 15 seconds.

Severity: Medium; semantic memory recall fails after the configured idle stop or a Gateway restart.

Frequency: Reproducible on every observed cold start above 15 seconds; warm searches succeed.

Consequence: The first recall attempt fails and enters the Memory Core cooldown despite a healthy provider that becomes ready within its configured service budget.

## Additional information

A focused fix should exclude only managed local-service acquisition from the whole-search budget while preserving the existing 15-second deadline for manager setup, retrieval, supplements, cleanup, and other providers. The existing `readyTimeoutMs` remains the startup owner, so no new configuration surface is required.

RileyJJY:
## What Problem This Solves

Resolves two deadline regressions in `memory_search`: cold managed local embedding startup could be cancelled by the search budget, and `corpus=all` could let a paused primary budget also extend the wiki supplement budget.

## Why This Change Was Made

The tool and runtime now resolve one stable adapter per host acquisition callback, so manager cache identity is shared across runtime → tool → runtime transitions while different hosts remain isolated. Readiness-only pause/resume is carried by the provider target boundary; post-readiness reconciliation, ordinary manager acquisition, and registry waits remain inside each search's 15-second deadline. Primary memory and wiki searches keep independent budgets, both still linked to immediate caller cancellation.

## User Impact

Cold managed local embedding services can finish startup within their configured readiness window. A slow primary startup no longer extends wiki retrieval, reconciliation cannot consume an unbounded readiness exemption, and runtime/tool consumers do not replace a healthy manager merely because they crossed a plugin boundary.

## Evidence

- Exact-head (`e7ba92a51eddde86722077565357bd91c1d5f575`) production-module-boundary proof using the real `createMemoryRuntime` and `createMemorySearchTool` consumers, with the same host callback across runtime → tool → runtime:

```text
runtime → tool → runtime: one adapter identity was preserved
different host callbacks: adapter identities remained isolated
negative-control: caller cancellation preserved
```

- Memory extension validation: `node scripts/run-vitest.mjs run --config test/vitest/vitest.extension-memory.config.ts extensions/memory-core/src/runtime-provider.test.ts extensions/memory-core/src/tools.test.ts extensions/memory-core/src/memory-corpus.test.ts extensions/memory-core/src/tools.budget.test.ts extensions/memory-core/src/tools.real-manager.test.ts` (5 files, 84 tests passed), including the real manager cold-start path and the runtime/tool identity sequence.
- Provider readiness boundary: `node scripts/run-vitest.mjs run --config test/vitest/vitest.plugins.config.ts src/plugins/openai-compatible-embedding-provider.test.ts` (35 tests passed), including pause/resume around readiness and slow reconciliation outside the exemption.
- Targeted preflight: format, changed-file lint, owning-project types, focused tests, identity, merge compatibility, dependency, and PR-body checks passed.
- `git diff --check`.
- Canonical precedents: merged PR #95757 keeps timeout ownership at the backend phase boundary while preserving immediate caller cancellation; PR #142417 preserves callback identity across manager reuse.

<details>
<summary>Production boundary proof source</summary>

```ts
const acquireLocalService = async () => ({ release() {} });
const runtime = createMemoryRuntime({ acquireLocalService });
await runtime.getMemorySearchManager({ cfg, agentId: "main" });
await createMemorySearchTool({ config: cfg, acquireLocalService })?.execute("identity", {
  query: "runtime tool runtime",
});
await runtime.getMemorySearchManager({ cfg, agentId: "main" });
// The saved exact-head test asserts all three manager callbacks share identity.
console.log({ adapterIdentity: "preserved", callerAbort: "preserved" });
```

</details>

AI-assisted: built with Codex

### DIFF
--- a/extensions/memory-core/src/memory-corpus.ts
+++ b/extensions/memory-core/src/memory-corpus.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from "node:async_hooks";
 import { runTasksWithConcurrency } from "openclaw/plugin-sdk/concurrency-runtime";
 import { extractErrorCode, formatErrorMessage } from "openclaw/plugin-sdk/error-runtime";
 import {
@@ -17,6 +18,19 @@ import {
 
 type MemoryCorpus = "memory" | "wiki";
 const memoryCorpusDeadlineChecks = new WeakMap<AbortSignal, () => void>();
+const activeMemoryCorpusDeadlineControl = new AsyncLocalStorage<MemorySearchDeadlineControl>();
+
+export function runWithMemoryCorpusDeadlineControl<T>(
+  controlDeadline: MemorySearchDeadlineControl,
+  run: () => T,
+): T {
+  return activeMemoryCorpusDeadlineControl.run(controlDeadline, run);
+}
+
+export function getMemoryCorpusDeadlineControl(): MemorySearchDeadlineControl | undefined {
+  return activeMemoryCorpusDeadlineControl.getStore();
+}
+
 type MemorySupplement = ReturnType<typeof listMemoryCorpusSupplements>[number];
 type MemorySupplementGetResult = NonNullable<
   Awaited<ReturnType<MemorySupplement["supplement"]["get"]>>
--- a/extensions/memory-core/src/memory-corpus.ts
+++ b/extensions/memory-core/src/memory-corpus.ts
@@ -1,4 +1,3 @@
-import { AsyncLocalStorage } from "node:async_hooks";
 import { runTasksWithConcurrency } from "openclaw/plugin-sdk/concurrency-runtime";
 import { extractErrorCode, formatErrorMessage } from "openclaw/plugin-sdk/error-runtime";
 import {
@@ -18,18 +17,6 @@ import {
 
 type MemoryCorpus = "memory" | "wiki";
 const memoryCorpusDeadlineChecks = new WeakMap<AbortSignal, () => void>();
-const activeMemoryCorpusDeadlineControl = new AsyncLocalStorage<MemorySearchDeadlineControl>();
-
-export function runWithMemoryCorpusDeadlineControl<T>(
-  controlDeadline: MemorySearchDeadlineControl,
-  run: () => T,
-): T {
-  return activeMemoryCorpusDeadlineControl.run(controlDeadline, run);
-}
-
-export function getMemoryCorpusDeadlineControl(): MemorySearchDeadlineControl | undefined {
-  return activeMemoryCorpusDeadlineControl.getStore();
-}
 
 type MemorySupplement = ReturnType<typeof listMemoryCorpusSupplements>[number];
 type MemorySupplementGetResult = NonNullable<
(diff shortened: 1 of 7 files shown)


RileyJJY:
@clawsweeper re-review

PR #143189, opened 2026-09-09
refactor: share sandbox filesystem parameter contracts

steipete:
Closes #143186

## What Problem This Solves

Sandbox filesystem implementations repeat parameter records already defined by their shared public contract. Independent copies increase the maintenance cost of keeping implementations aligned. This is part of the maintainer-requested code reduction campaign.

## Why This Change Was Made

Twenty-four exactly matching multiline annotations now derive their parameters from the corresponding `SandboxFsBridge` method. Optional methods use `NonNullable` before extracting parameters; each concrete class retains its required methods and return types. The seven differing mxc signatures remain unchanged. No helpers, aliases, imports, runtime expressions, or SDK exports are added.

## User Impact

No user-visible behavior or public API change. This removes 116 production code/physical lines across four files while keeping the existing filesystem contract as the single owner of those parameter fields.

## Evidence

- The patch is byte-identical after refreshing onto main with the separately landed Telegram fixture repair #143197 and standalone reset repair #143281.
- All four complete files emit byte-identical JavaScript after TypeScript erasure.
- Canonical OpenShell and mxc builds produce identical runtime/static output: three files, 126,518 bytes.
- All 380 regenerated SDK declaration files are byte-identical: 3,504,296 bytes per side.
- Canonical public API rendering is byte-identical across 152 entrypoints, 4,447 exports and 9,504 reachable declaration sections; the complete 14,623,526-byte JSON reports have no changes.
- The full changed-file gate passed, including all four typecheck lanes, six existing doctor cases, formatting, lint, dependency, cycle and boundary guards. SDK export-map verification also passed.
- Independent P2 review has no findings and remains applicable to the identical patch; refreshed source-contract checks retain the differing mxc signatures and OpenShell's concrete method presence.

No live sandbox or Gateway run is claimed for this erased-annotation change. The affected plugin builds and complete SDK/API output were verified; a full root runtime package build was not run. API rendering used the canonical command and its documented memory budget; declaration generation ran after the plugin builds completed.

## Stored-data compatibility and CI

No serialized fields, filesystem operations, storage formats, schemas, migration rules or version markers change. The automated stored-data flag does not apply to these erased annotations, as confirmed by complete runtime and public API equivalence. The [current source review](https://github.com/openclaw/openclaw/pull/143189#issuecomment-5603887596) independently disproves that concern; its remaining generated migration checklist is therefore not applicable.

The earlier CI failures were repaired at their owners in #143197 and #143281. This refreshed head still requires green hosted CI before landing.

PR #143281, opened 2026-09-09
fix: standalone session resets stall during thinking lookup

### LINKED ISSUE #143252: Standalone session resets wait for thinking catalog discovery
## Problem

A standalone `/new` or `/reset` resolves default thinking and reasoning before the existing reset handler can acknowledge or reject the command. On a cold source checkout, that lookup loads provider-discovery modules even though no model turn will run. A slow or unavailable catalog therefore delays or prevents a session-control operation.

## Reproduction and cause

The unchanged Discord native reset suite reproduces this at CI merge `f10de943e9b7638a1279e9df1d8593feefc96a10`: its first `/new` case exceeds the existing 120-second deadline while the remaining 15 cases pass. Standalone local runs reproduce it without any preceding test file. Phase and CPU profiling locate the delay in default-thinking catalog discovery during reply directive resolution, including more than 113 seconds loading provider modules. Registry preparation and reply-resolver loading each take less than half a second in the phase probe.

These are cold source-test measurements, not a claim about packaged production latency. A deterministic catalog-failure regression also demonstrates the dependency: baseline standalone and denied reset cases fail when catalog loading throws, while a `/new` with follow-up text correctly reaches model-capability resolution.

## Expected behavior and repair

The existing reset owner should acknowledge or reject standalone commands before any model-derived settings are needed. A continuing model turn, status output, explicit thinking controls, and compaction must retain the same selected-model defaults, explicitness, authorization and reasoning policy.

Keep thinking and reasoning resolution together in one lazy owner, because treating unresolved thinking as active would change model-derived reasoning behavior. Carry that resolver through internal command dispatch and consume it only when required; retain the existing reset, ACP, follow-up-text and run-admission contracts. This requires no schema, storage, protocol or configuration change.

Related: #143189 exposed the failure during validation; its sandbox annotation change does not modify this path. Historical startup-discovery issue #120222 is adjacent, with a different measured cause and scope.

steipete:
Closes #143252
Related: #143189

## What Problem This Solves

Fixes an issue where users sending a standalone `/new` or `/reset` could wait for thinking-catalog discovery before receiving an acknowledgement or authorization denial. The existing Discord native reset flow exposed this on a cold source checkout: its first case exceeded the unchanged 120-second deadline.

## Why This Change Was Made

Thinking and reasoning defaults now resolve together through one lazy owner, consumed by command handlers or the continuing model turn. The existing reset handler runs before that resolution and before loading the general command handlers. Ordinary and native-slash callers share the resolver; selected-model, explicit-setting, authorization, ACP and fallback behavior remain at their existing owners.

Earlier model-selection validation is unchanged and can still load a catalog for some wildcard allowlists. This repair removes the measured default-thinking dependency; it does not promise that all reset configurations are catalog-independent.

## User Impact

Standalone reset acknowledgements and denials no longer require model-derived thinking and reasoning defaults. Follow-up text, status and thinking controls, compaction, and ordinary turns retain their configured settings. No configuration, schema, storage, protocol or public SDK change is introduced.

## Evidence

- The original 16-case Discord reset suite passed with unchanged assertions and timeouts; the final cold run's first `/new` case completed in 8.1 seconds. These are source-test observations, not a packaged-production latency benchmark.
- Regression-only changes on unchanged main fail three standalone/denied-reset cases with `thinking catalog unavailable`. The `/new take notes` control still reaches capability resolution; the repaired trigger suite passes all 22 cases.
- The producer/consumer group passes 276 cases across 10 suites, covering selected-model defaults, explicit and unauthorized reasoning settings, native slash dispatch and affected command paths.

- All 12 fallback cases pass after the final selected-model correction. The wildcard-policy preservation case also passes on original main and detects using the filtered primary's defaults instead of the selected fallback's defaults.
- The final complete changed-file gate, production build and fresh independent P0–P2 review pass. Earlier targeted proof also includes 34 fixture/catalog-policy cases.

- Final E2E passes all 23 selected cases: 22 trigger cases and the existing isolated-Gateway replacement turn. The latter verifies cancellation, a second preparation hook, a second model request and preserved `/new` follow-up text. One unrelated shutdown sibling was excluded. The harness uses a local synthetic model server; no external provider live proof is claimed.


steipete:
Merged via squash.

- Prepared head SHA: [a674b7acc8de4736cc3a2310d49820e5a73a9d37](https://github.com/openclaw/openclaw/pull/143281/commits/a674b7acc8de4736cc3a2310d49820e5a73a9d37)
- Landed commit: [0720f31dfc81df7b8f4344c402117faeb0c9f02b](https://github.com/openclaw/openclaw/commit/0720f31dfc81df7b8f4344c402117faeb0c9f02b)

PR #136241, opened 2026-09-02
fix(cli): preserve literal completion cache paths

steipete:
<details>
<summary>Additional instructions</summary>

Keep **Allow edits from maintainers** enabled.

</details>

## What Problem This Solves

Installing shell completion can report success while a fresh shell cannot load a valid generated cache. Raw double quotes let dollar signs, quotes, doubled backslashes and command substitutions change the resolved source path. Fixes #136237.

## Why This Change Was Made

Installed cache hooks now share the existing reload formatter's shell-specific literal quoting policy. Bash/zsh, Fish and PowerShell retain their distinct rules; PowerShell output is unchanged. Reinstall recognizes current literal hooks and exact stable-release hooks, including malformed guarded hooks generated when an old directory name contained a quote.

Previous-directory cleanup still requires the OpenClaw marker, complete matching guard/source operands, the completions directory and the current cache filename. Unmarked sources, compound user commands, mismatched guards and other shell files are preserved. Profile selection, symlink handling, atomic writes, missing-cache status, Doctor behavior and cache generation retain their contracts.

The owner cleanup removes repeated supported-shell branches, obsolete private nullable states, redundant invocation branches, the manual dynamic-line scan and a one-use extension wrapper. Production changes are **+69/-77, net -8 LOC**; tests +90/-47, net +43; docs +2. Pure moves receive no deletion credit. No dependency, configuration, protocol or storage change.

## Evidence

The original runtime is byte-identical in stable v2026.8.2 and pre-fix main. Actual Commander registration generates real completion scripts and installs them into synthetic temporary profiles. In the unchanged before/after harness, nine variants through both routes improve from 16/36 to 36/36 successful Bash/zsh loads. Independent Fish 4.8.1 checks pass all nine installed/reload cases after repair; four failed before.

Observed literal-dollar results from the real-shell receipts:

```json
[
  {
    "shell": "bash",
    "pathVariant": "literal$qa24unset",
    "before": {
      "reportedInstalled": true,
      "routes": [
        {
          "route": "installed",
          "exit": 1,
          "loaded": false
        },
        {
          "route": "reload",
          "exit": 1,
          "loaded": false
        }
      ]
    },
    "after": {
      "reportedInstalled": true,
      "routes": [
        {
          "route": "installed",
          "exit": 0,
          "loaded": true
        },
        {
          "route": "reload",
          "exit": 0,
          "loaded": true
        }
      ]
    }
  },
  {
    "shell": "zsh",
[...]

PR #143282, opened 2026-09-09
fix(cli): recognize managed portable Bash completion hooks without rewriting them

### LINKED ISSUE #143267: [Bug]: completion installer duplicates portable managed Bash source lines
### Bug type

### Beta release blocker

No

### Summary

The completion installer does not recognize an existing portable, guarded source line for its cached Bash completion script. It appends a second block containing a literal absolute home path, creating recurring conflicts when `.bashrc` is managed by chezmoi.

### Steps to reproduce

Reproduced with the installed 2026.9.2 CLI against a temporary home, without running a Gateway. Resolve the actual package entry point before changing the child process's environment; do not invoke a home-relative wrapper.

1. Create a temporary home containing `.openclaw/completions/openclaw.bash`. A minimal fixture containing `complete -W "completion doctor" openclaw` is sufficient because installation only needs the cached file to exist.
2. Create `.bashrc` containing exactly:

   ```bash
   [[ -f "${HOME}/.openclaw/completions/openclaw.bash" ]] && source "${HOME}/.openclaw/completions/openclaw.bash"
   ```

3. Run the real package entry point in a child process with `HOME` set to that temporary home, `OPENCLAW_STATE_DIR` set to its `.openclaw` directory, and `SHELL=/bin/bash`:

   ```text
   node <absolute-package-path>/openclaw.mjs completion --shell bash --install --yes
   ```

4. Compare `.bashrc` before and after. In a fresh Bash process, sourcing the profile and running `complete -p openclaw` confirms completion still works.

### Expected behavior

An existing source of the same cached completion should be recognized as configured. Users who manage their shell profile should be able to retain ownership of that profile while allowing OpenClaw to generate and refresh its completion cache.

The documented `--write-state` flow already separates cache generation from profile installation. This request concerns preserving that separation during detection and repair as well.

### Actual behavior

The CLI exits successfully and preserves the portable line, but appends:

```bash
# OpenClaw Completion
[ -f '<temporary-home>/.openclaw/completions/openclaw.bash' ] && source '<temporary-home>/.openclaw/completions/openclaw.bash'
```

The installed completion runtime reports `cacheExists=true`, `completionRecognized=false`, and `slowPattern=false` for the portable setup. Thus a working setup is classified as missing.

This is **not** a claim that every doctor/update invocation writes the profile: current doctor skips missing installation in non-interactive mode. Interactive doctor offers installation with a default-yes prompt; accepting it reaches the profile installer. Explicit `completion --install` deterministically reproduces the write above.

### OpenClaw version

2026.9.2 (installed package, actual CLI reproduction). Current main was also inspected at `ef37c4f3755d4a424c12530d9339d48f2fbcdc8d`; main was not built or executed for this report.

### Operating system

Ubuntu 26.04.1 LTS, Bash

### Install method

npm global

### Model

Not applicable; shell completion only.

### Provider / routing chain

Not applicable; no model or Gateway request.

### Logs

```json
{
  "version": "2026.9.2",
  "cliExit": 0,
  "profileChanged": true,
  "portableLinePreserved": true,
  "completionWorks": true
}
```

- [Completion runtime](https://github.com/openclaw/openclaw/blob/ef37c4f3755d4a424c12530d9339d48f2fbcdc8d/src/cli/completion-runtime.ts): `isCompletionProfileLine` recognizes literal cache paths; `updateCompletionProfile` appends the literal-path block.
- [Doctor completion flow](https://github.com/openclaw/openclaw/blob/ef37c4f3755d4a424c12530d9339d48f2fbcdc8d/src/commands/doctor-completion.ts): the missing-profile path prompts unless non-interactive.
- [Documented cache/profile split](https://docs.openclaw.ai/cli/completion).

### Impact and severity

This blocks unattended dotfile reconciliation when the manager refuses destination edits. In the affected setup, chezmoi owns `.bashrc`, its source intentionally contains portable paths, and an application-appended block makes the file locally modified. Unrelated targets can still apply, but the account reconciliation remains nonzero until manual conflict resolution. Capturing the absolute path into a portable dotfile repository does not resolve the ownership conflict.

The isolated installer reproduction was run once and reproduced the duplicate write. The recurrence risk is a subsequent installation/interactive repair, not every CLI invocation.

### Additional information

Related prior work: #136241 hardens literal completion paths and profile-line ownership. This report concerns portable user-managed source lines, rather than escaping or deleting an old literal-path hook.

I'd like to contribute the repair. Which ownership contract would maintainers prefer?

- An explicit way to suppress shell-profile writes while retaining completion-cache generation/refresh, honored by direct installation and repair flows; or
- Recognition and preservation of an existing portable source line for the same cached script.

The first option would add configuration, so I am requesting direction before implementing it, per AGENTS.md. The second must recognize only supported safe forms without evaluating arbitrary shell code or consuming unrelated profile commands. Converting `.bashrc` into a dotfile-manager merge script would sacrifice ordinary capture in this setup and does not establish a clear owner for the completion block.

LiuwqGit:
## What Problem This Solves

Closes #143267.

Completion installation can append a duplicate absolute-path hook to a working Bash profile managed by a dotfile tool. The installer does not recognize guarded `HOME`-relative hooks for the same completion cache, so a repeated install changes a profile the user owns.

## Why This Change Was Made

Recognition and rewrite ownership are separate. Supported portable hooks remain byte-identical, including an existing OpenClaw marker. Old generated literal and dynamic hooks still migrate. Previous literal-hook detection requires an absolute path, so it cannot consume a user-managed portable line. Shell token boundaries use spaces and tabs, preventing malformed Unicode-whitespace guards from suppressing a needed installation.

## User Impact

Running `openclaw completion --shell bash --install --yes` twice preserves a working portable profile. When only the cache is missing, `openclaw doctor --fix --non-interactive` regenerates it without rewriting the profile. Mismatched, compound, single-quoted, unsafe, and malformed hooks remain user-owned.

## Evidence

- Built and deployed the pinned main baseline and candidate with the installed package entrypoint resolved before changing the child home.
- The real baseline CLI appended a duplicate literal block. The corrected installed CLI preserved the profile across both installs, including marked single- and double-bracket hooks.
- Fresh Bash processes registered the generated completion. After deleting only the cache, the public Doctor command regenerated it and left the profile unchanged.
- Real malformed-whitespace controls failed in Bash before installation. The corrected installer preserved those user lines and added the existing literal hook instead of falsely reporting the malformed hook as sufficient.
- Focused checks passed: 109 CLI tests, 20 Doctor tests, 17 setup tests, formatting, syntax lint, and `git diff --check`. The corrected regression suite failed eight cases with the incoming runtime for the intended errors.
- Bash is the required live-shell proof. Zsh and Fish recognition have supplemental test coverage; this does not claim arbitrary shell syntax, portable PowerShell hooks, or universal dotfile-manager compatibility.

Related: #136241 covers earlier literal-path quoting and migration. #143306 is a later implementation of the same issue.

Independent installed CLI validation also passed portable variants, owned-hook migrations, and twelve negative forms. Doctor cache repair used an explicit `SHELL=/bin/bash` environment; the raw evidence retains the initial omitted-shell setup failure separately. Zsh and Fish executables were unavailable.

### DIFF
--- a/docs/cli/completion.md
+++ b/docs/cli/completion.md
@@ -44,6 +44,8 @@ Profile changes are staged beside the destination and atomically replace it only
 
 Installed source lines preserve literal cache paths, including spaces, quotes, dollar signs, and backslashes. Reinstalling replaces OpenClaw's previous source line after the state directory changes.
 
+A user-managed portable hook for the same cached script (for example `[[ -f "${HOME}/.openclaw/completions/openclaw.bash" ]] && source "${HOME}/.openclaw/completions/openclaw.bash"`, as a dotfile manager would write) is recognized as completion being configured. Doctor and `completion --install` leave such managed lines byte-for-byte untouched instead of appending a duplicate literal-path block.
+
 ## Permission failures
 
 If Doctor or onboarding cannot update your shell profile, completion remains
--- a/src/cli/completion-runtime.test.ts
+++ b/src/cli/completion-runtime.test.ts
@@ -849,4 +849,256 @@ describe("completion-runtime", () => {
       );
     });
   });
+
+  it
+    .skipIf(process.platform === "win32")
+    .each([
+      '[[ -f "${HOME}/.openclaw/completions/openclaw.bash" ]] && source "${HOME}/.openclaw/completions/openclaw.bash"',
+      '[ -f "$HOME/.openclaw/completions/openclaw.bash" ] && source "$HOME/.openclaw/completions/openclaw.bash"',
+    ])(
+    "preserves a managed portable Bash hook byte-for-byte across installs: %s",
+    async (portableHook) => {
+      const homeDir = tempDirs.make("openclaw-bash-portable-home-");
+      const stateDir = path.join(homeDir, ".openclaw");
+      await withEnvAsync(
+        {
+          HOME: homeDir,
+          USERPROFILE: homeDir,
+          OPENCLAW_STATE_DIR: stateDir,
+          XDG_CONFIG_HOME: undefined,
+          ZDOTDIR: undefined,
+        },
+        async () => {
+          const profilePath = path.join(homeDir, ".bashrc");
+          const cachePath = resolveCompletionCachePath("bash", "openclaw");
+          expect(cachePath).toBe(path.join(stateDir, "completions", "openclaw.bash"));
+          await fs.mkdir(path.dirname(cachePath), { recursive: true });
+          await fs.writeFile(cachePath, "complete -W 'completion doctor' openclaw\n", "utf-8");
+          await fs.writeFile(profilePath, `${portableHook}\n`, "utf-8");
+
+          await expect(isCompletionInstalled("bash", "openclaw")).resolves.toBe(true);
+
+          await installCompletion("bash", true, "openclaw");
+          await expect(fs.readFile(profilePath, "utf8")).resolves.toBe(`${portableHook}\n`);
+
+          await installCompletion("bash", true, "openclaw");
+          await expect(fs.readFile(profilePath, "utf8")).resolves.toBe(`${portableHook}\n`);
+
+          const freshBash = spawnSync(
+            "bash",
+            ["-c", `source "${profilePath}" && complete -p openclaw`],
+            { encoding: "utf8" },
+          );
+          expect(freshBash.stderr).toBe("");
+          expect(freshBash.status).toBe(0);
+          expect(freshBash.stdout).toContain("openclaw");
+        },
+      );
+    },
+  );
+
+  it.skipIf(process.platform === "win32")(
+    "preserves a managed portable Zsh hook without appending a literal block",
+    async () => {
+      const homeDir = tempDirs.make("openclaw-zsh-portable-home-");
+      const stateDir = path.join(homeDir, ".openclaw");
+      await withEnvAsync(
+        {
+          HOME: homeDir,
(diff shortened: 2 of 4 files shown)


obviyus:
Thanks @LiuwqGit!

**Proof:** installed CLI runs reproduced the duplicate on main and verified profile preservation after the fix. Fresh Bash loaded completion, and Doctor regenerated a deleted cache without rewriting the profile when `SHELL=/bin/bash` was set.

- **Bug:** working portable hooks were missed, so installation appended a duplicate absolute-path block.
- **Fix:** recognition stays separate from rewrite ownership; supported portable hooks and their markers remain intact, while malformed shell blanks are rejected and owned literal/dynamic hooks still migrate.

| Before | After |
| --- | --- |
| Reinstall dirties a managed profile with a duplicate hook. | Two installs leave a supported portable profile byte-identical. |
| A missing cache can be mistaken for missing installation. | Doctor restores the cache without changing the managed profile. |

146 focused CLI, Doctor, and setup tests passed. Independent installed checks covered portable variants, migration, and twelve negative forms. Final review had no findings, and [exact-head hosted checks passed](https://github.com/openclaw/openclaw/actions/runs/34397976161) without an exception.

Live-shell proof is Bash; Zsh/Fish have supplemental test coverage. No arbitrary shell or portable PowerShell compatibility is claimed.

Merged as `a91a987c89202b335671164409bf237d842f3e86`.

PR #142505, opened 2026-09-08
fix(sessions): skip cleanup for newly protected history

steipete:
Related: #112423.

<details>
<summary>Maintainer edits</summary>

This same-repository branch remains editable by maintainers.

</details>

## What Problem This Solves

Fixes an issue where disk-budget cleanup continued expensive reclamation after a historical session became protected while its archive was being prepared. Cleanup could also continue to another candidate using an outdated pressure estimate after another actor freed physical space.

## Why This Change Was Made

The parent uses its existing fresh protection set to cancel an excluded candidate before reclamation Worker admission. A distinct cancellation outcome triggers physical-usage measurement after lifecycle and writer holds are released, before another candidate is considered. Other no-op outcomes retain their existing behavior.

Every admitted Worker retains the full integrity and foreign-key checks, current-owner and commit fences, and actual-exit ownership. Canceled work is not credited for another actor's removal and does not open a Worker solely to validate discarded work; earlier preparation failures and subsequent measurement failures still propagate.

## User Impact

Newly protected history avoids unnecessary Worker startup and database validation. Cleanup checks actual remaining pressure before deleting more history. Retention settings, archive durability, schema, and the handling of work protected after dispatch remain unchanged.

## Evidence

- The reviewed pre-fix test on `a92c15a5dc3710c0799320d07c35df862b074b0d` reached its intended failure after a real reclamation Worker received an already-protected historical ID. History remained intact and no removal occurred.
- Both permanent regression cases failed against unchanged production code: after real peer artifact bytes were removed, cleanup still deleted the next candidate. They pass with this change and observe actual Worker admission messages, retained transcript bytes, post-hold physical measurement, and exact measurement-error propagation.
- Final local focused proof: 81 tests passed across history eviction, protected cancellation, reclamation, commit, and admission-race suites. This includes the unchanged protection-after-dispatch coverage. The runner exited naturally and verified Worker-artifact cleanup.
- Changed-scope formatting, typechecking, lint, schema, boundary, and import-cycle checks passed. Independent Codex review through P2 found no actionable findings.
- [Blacksmith Testbox Linux proof](https://github.com/openclaw/openclaw/actions/runs/34273524347): normal build, 36 history/cancellation tests, API seeding, the real explicit-offline `sessions cleanup --enforce --store ... --json` command, and verification all exited naturally with code 0 and closed their process groups. Cleanup removed one eligible historical generation and preserved both current generations plus protected history; integrity and foreign-key assertions passed.
[...]

PR #143285, opened 2026-09-09
fix(sessions): prevent excess cleanup after disk pressure clears

steipete:
Related: #142505

## What Problem This Solves

Resolves a problem where session cleanup could delete additional history or cap-archived sessions after concurrent work had already reduced disk usage below the cleanup target. A skipped candidate could leave the sweep using an older physical-byte total for the next deletion.

## Why This Change Was Made

Remeasure physical usage after a historical planning/reclamation no-op or a declined cap-archived entry deletion, once the existing lifecycle and writer holds have returned. Historical no-op outcomes now share one path; measurement failures still stop the sweep.

## User Impact

Cleanup preserves the next eligible history or archived session when disk pressure has cleared. Configured targets, retention and protection rules, archive ordering, and worker/integrity behavior remain unchanged.

## Evidence

- Before the historical repair, four synthetic cases—planning and actual Worker no-op, each with successful or failed remeasurement—deleted the next history. The two existing materialization-cancellation controls passed. Separately, both cap-entry cases failed with the historical repair already applied and the cap-entry branch unchanged.
- The tests use real SQLite, a real physical peer file, actual reclamation Workers and durable transcript assertions. Post-dispatch cases observe an actual `deleted: false` result before verifying that the next history remains intact. Measurement-error cases verify that cleanup stops without deleting that next history.
- All 56 focused tests pass across cancellation, eviction, ownership and parity suites, including the eight no-op cases. `node scripts/check-changed.mjs` passes, and independent P0–P2 review found no actionable issues.
- Exact-head CI [34382247348](https://github.com/openclaw/openclaw/actions/runs/34382247348) is green for `b6355f4ba64e50af63a28a16650fbbe9ae431e14`.
- The real cleanup CLI proof passed all four cases on Linux through Crabbox's Blacksmith Testbox provider ([run 34387658809](https://github.com/openclaw/openclaw/actions/runs/34387658809)). The same-run build, four seed commands and four cleanup commands all exited 0 with confirmed process cleanup. Each CLI finished in 17–19 seconds within its original 180-second deadline.

```text
pnpm --silent openclaw sessions cleanup --agent main --store <isolated-store> --enforce --json
```

| Real CLI case | Actual reclamation results | Removed entries | Physical bytes after | Durable result |
| --- | --- | ---: | ---: | --- |
| History, peer frees 8 MiB | A: deleted=false | 0 | 929,792 | A and B transcript hashes unchanged |
| History, pressure remains | A: deleted=false; B: deleted=true | 1 | 9,220,096 | A preserved; B history reclaimed |
| Cap-archived entry, peer frees 8 MiB | A: deleted=false, expectedEntryMismatch=true | 0 | 929,792 | A and B transcript hashes unchanged |
| Cap-archived entry, pressure remains | A: deleted=false; B: deleted=true | 1 | 9,220,096 | A preserved; B's reported session ID and durable deletion match |

Every case began at 9,318,400 physical bytes, with a 3,026,944-byte cleanup target. Relieved-pressure cases observed a fresh real measurement after the declined reclamation Worker exited. Controls kept the peer file and still reclaimed the next eligible candidate. The native artifact export preserved the raw CLI output, Worker results, before/after snapshots and cleanup receipts; all 80 manifest entries were hash-verified.

The first proof attempt produced the expected cleanup JSON but timed out. Investigation reproduced a harness liveness defect: its observer could keep an otherwise idle Worker referenced. A standalone old/new preload control confirmed that mechanism. The corrected observer only attaches where an owner already has a message listener; the fresh four-case run then passed without extending deadlines or changing production code. The failed attempt remains recorded as failed.

This is synthetic controlled-interleaving CLI evidence. It does not attribute a live incident, change retention or integrity policy, or demonstrate Worker-pool performance. `CHANGELOG.md` remains release-owned.

### DIFF
--- a/docs/concepts/session.md
+++ b/docs/concepts/session.md
@@ -309,6 +309,10 @@ until physical usage exceeds `maxDiskBytes`; disk-budget cleanup may then delete
 the oldest cap archives after cheaper artifacts and unreferenced history are
 exhausted. Sessions without a recorded archive reason remain protected.
 
+After skipping a history generation or archived session, disk-budget cleanup
+rechecks physical usage before considering another deletion. A measurement
+failure stops the sweep.
+
 If you previously used DM isolation and later returned `session.dmScope` to
 `main`, preview stale peer-keyed DM rows with
 `openclaw sessions cleanup --dry-run --fix-dm-scope`. Applying the same flag
--- a/src/config/sessions/session-history-eviction.protected-cancellation.test.ts
+++ b/src/config/sessions/session-history-eviction.protected-cancellation.test.ts
@@ -4,7 +4,7 @@ import type { Worker } from "node:worker_threads";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { resetAgentRunRegistryForTest } from "../../infra/agent-run-registry.js";
 import { executeSqliteQuerySync } from "../../infra/kysely-sync.js";
-import { isSessionLifecycleMutationActive } from "../../sessions/session-lifecycle-admission.js";
+import * as lifecycle from "../../sessions/session-lifecycle-admission.js";
 import {
   closeOpenClawAgentDatabasesForTest,
   openOpenClawAgentDatabase,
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b6355f4ba64e50af63a28a16650fbbe9ae431e14](https://github.com/openclaw/openclaw/pull/143285/commits/b6355f4ba64e50af63a28a16650fbbe9ae431e14)
- Landed commit: [057e357034401a668a88e25fcb7895f2b73e36e1](https://github.com/openclaw/openclaw/commit/057e357034401a668a88e25fcb7895f2b73e36e1)

PR #143289, opened 2026-09-09
fix(voice-call): stream consult results before completion

johnemau:
## What Problem This Solves

Resolves a problem where callers using GPT-Realtime heard no work-supervisor result until the embedded agent completed.

## Why This Change Was Made

Forward sanitized assistant partial replies through an awaited, cancellable speech queue that preserves order and deduplicates cumulative snapshots and true deltas. Append one bounded final tool result with response suppression after successful streaming, while retaining the existing non-streaming path and clear failure fallback.

## User Impact

Callers can hear completed supervisor segments while work continues, exactly once and in order, and can interrupt them immediately.

## Evidence

- Focused suite: 185 tests passed across the consult runtime, voice-call runtime/handler, stream reducer, and OpenAI Realtime bridge.
- Changed-file gate: formatting, core and extension type-checks, dead-export scan, lint, import-cycle checks, and repository guards passed.
- Autoreview: addressed cumulative/delta, whitespace, paragraph, bounded long-answer, coalesced-call, guarded-barge-in, reconnect-fallback, and non-streaming compatibility findings.

### Manual Test

Expected: With a 100 ms pause between two supervisor partial replies, the first segment reaches voice output before completion; both segments are heard once; one consolidated final result is appended without another response.

Actual: Before the change, first-spoken text did not occur before the final result. After the change, the controlled harness recorded first-spoken text at 1 ms and the final response at 104 ms, with both segments delivered once.

Generated with AI under the guidance of John.

### DIFF
--- a/extensions/openai/realtime-voice-bridge.ts
+++ b/extensions/openai/realtime-voice-bridge.ts
@@ -128,7 +128,7 @@ export class OpenAIRealtimeBridge extends OpenAIRealtimeEvents implements Realti
     if (this.pendingToolCallIds.size > 0) {
       // Control/status speech must not wait behind the long-running consult whose
       // function output owns the default conversation response.
-      this.standaloneSpeechQueue.push(text);
+      this.standaloneSpeechQueue.push({ text });
       this.flushStandaloneSpeech();
       return;
     }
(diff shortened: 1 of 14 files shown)


johnemau:
Thank you for your consideration and the detailed review feedback. I'm closing this proposal while we consolidate the work in a downstream fork and address the remaining correctness and validation gaps.

Generated with AI under the guidance of John.

PR #132103, opened 2026-08-28
feat(sandbox): discover environment skills and authorized workspace MCP servers

sallyom:
## What Problem This Solves

The built-in runtime can execute shell and filesystem tools in a sandbox, but it could not discover skills or explicitly authorized MCP servers that exist only in that execution environment.

This PR extends the native agent runtime's environment-capability contract so sandboxed agents can discover skills installed inside the execution environment and connect to explicitly authorized stdio MCP servers there—similar to the capability discovery already supported by Codex remote-harness. This enables capabilities that exist only in the sandbox without requiring them to also be installed on the Gateway host. The contract is versioned and implemented for OpenShell remote mode.

## Design

One bounded backend discovery snapshot is shared by skill and MCP consumers. Core retains parsing and policy ownership; the executor owns bounded filesystem I/O, environment requirement probes, and process launch.

Workspace `.mcp.json` is executable configuration and is denied by default. A server may start only when `agents.*.sandbox.environment.capabilityRoots` explicitly selects its root and exactly matches its normalized server name, command, complete arguments, cwd, and environment. Startup and reconnect consult the current Gateway-owned sandbox configuration, require the same live backend/runtime/root, and revalidate after each awaited preparation step immediately before process launch. Ordinary per-tool policy remains a separate post-discovery visibility and invocation gate.

MCP cwd is resolved against the discovery root and backend-validated before launch. Cancellation cannot revive a closed transport, late preparation still finalizes its lease exactly once, stderr uses the existing bounded logger, and executor-side descendants are reaped through the shared sandbox cleanup owner.

Environment skills use the canonical parser, policy, and prompt bounds. OS, binary, and environment-presence requirements are checked in the executor without copying Gateway credentials. Native skill names win; host-projected/generated copies are excluded; command dispatch is disabled.
The final merged native/environment catalog is compacted against the resolved model context budget without changing admitted skill identities or full instruction reads.

## User Impact and Compatibility

- Existing Gateway-configured MCP servers and sandbox configuration remain unchanged.
- Existing configurations remain valid. Omitted `capabilityRoots` means workspace MCP startup is denied.
- Workspace MCP execution is new and unreleased in this PR, so requiring explicit authorization is not a stable-release break.
- Existing OpenShell `remote` workspaces can begin contributing eligible remote skills without a new setting. This is an intentional prompt/catalog behavior change; ordinary skill policy, collision rules, and prompt bounds still apply.
- OpenShell mirror mode is unchanged and does not advertise environment capabilities.
[...]

PR #143291, opened 2026-09-09
fix: sandboxed plugin agents can read listed skills

### LINKED ISSUE #124555: Sandboxed Codex agents get a skills catalog whose locations do not exist inside the container
### Bug type

### Beta release blocker

No

### Summary

For agents running through the Codex app-server **with a sandbox enabled**, the `## OpenClaw Skills` developer-prompt catalog advertises every skill with a **host** path. Those paths do not exist inside the container, so the model receives a catalog it cannot act on. Unsandboxed runs are unaffected.

### Steps to reproduce

1. Configure an agent with `agents.defaults.sandbox.mode` set to a sandboxing mode.
2. Give that agent the Codex runtime with the app-server native tool surface and the sandbox exec-server enabled, so the turn runs on the remapped execution path (thread cwd is the container workdir).
3. Publish at least one skill so it is materialized into the sandbox skills workspace.
4. Start a turn and inspect the developer prompt the model receives, or ask the agent to open one of the advertised `SKILL.md` locations.

### Expected behavior

The `<location>` values in the skills catalog are container-visible paths, so the model can read the advertised `SKILL.md` files.

### Actual behavior

The advertised locations are host paths. Every read of an advertised location fails from inside the container.

The cause is a mapped/unmapped split. The workspace bootstrap computes a mapped prompt for the sandbox and keeps the unmapped snapshot beside it, then the Codex plane consumes the unmapped one: the catalog renderer receives `params.skillsSnapshot?.prompt` raw. The sandbox skill path mapper is never imported by the Codex `run-attempt` module at all, so the mapped prompt only reaches the built-in/embedded agent path.

Two recent fixes in this train do not address it: one repairs `exec.workdir` validation for sandbox-skills paths, the other filters conflicting bind mounts. Both are mount/validation plumbing; neither rewrites the `<location>` values rendered into the developer prompt.

A narrow seam looks like `preparePluginHarnessParams`, which already runs only for non-built-in harnesses -- the same shape as the sandboxed AGENTS.md fix in #120600 / #120618. I am happy to open a PR if that direction is acceptable, or to follow a different shape if the split is intentional.

### OpenClaw version

2026.8.1-beta.2 (mechanism also present in 2026.7.2-beta.7)

### Operating system

Linux

### Install method

npm

### Model

gpt-5.6 family via the Codex app-server harness

### Provider / routing chain

openai -> Codex app-server (native tool surface + sandbox exec-server enabled)

### Impact and severity

Skills are effectively unusable for sandboxed Codex agents: the model is handed a catalog whose every entry 404s inside the container. It fails silently -- the catalog looks well-formed, so the failure surfaces only as the agent being unable to follow its own published skills.

---
Filed with AI assistance; investigated and validated by me.

sallyom:
Closes #124555

Related: #132103

## What Problem This Solves

Fixes an issue where sandboxed agents using a plugin-owned harness received a skills catalog containing host-only file paths, so listed skills could not be opened inside the sandbox.

## Why This Change Was Made

Plugin-harness dispatch now uses the shared embedded skill-preparation path to project skill prompts and explicit skill references into the sandbox filesystem namespace. Prompt-only preparation skips process-wide skill environment mutation, and existing remote resource-mount handling remains unchanged.

The native embedded runtime already used this shared preparation path and was not affected. PR #132103 is adjacent because it expands the same preparation pipeline with environment-discovered skills; after rebasing, its async call-site change should compose with this projection.

## User Impact

Sandboxed plugin-backed agents can open the skills they are told are available. The fix applies to every plugin-owned harness, not only Codex.

## Evidence

- Regression test fails on unpatched `main` because the dispatched catalog contains the host path, then passes with the sandbox path after this fix.
- `node scripts/run-vitest.mjs src/agents/embedded-agent-runner/run/run-attempt-dispatch.owner.test.ts`: 4 tests passed.
- `git diff --check`: passed.

### DIFF
--- a/src/agents/embedded-agent-runner/run/run-attempt-dispatch.owner.test.ts
+++ b/src/agents/embedded-agent-runner/run/run-attempt-dispatch.owner.test.ts
@@ -1,7 +1,10 @@
+import fs from "node:fs/promises";
+import path from "node:path";
 import { afterEach, beforeEach, expect, it, vi } from "vitest";
 import { getAgentEventLifecycleGeneration } from "../../../infra/agent-events.js";
 import { createEmptyPluginRegistry } from "../../../plugins/registry-empty.js";
 import { setActivePluginRegistry } from "../../../plugins/runtime.js";
+import { buildSkillSnapshot } from "../../../skills/loading/workspace-skill-prompt.js";
 import { withOpenClawTestState } from "../../../test-utils/openclaw-test-state.js";
 import {
   createOperationalRunInstanceRef,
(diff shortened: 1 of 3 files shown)


sallyom:
Maintainer status: **merge-ready** at exact head `6409b763ea365add12c543cf23738071e0fdb9a9`.

- ClawSweeper's P1 was the stale-base conflict. The branch is rebased onto current `main`; the conflict resolution preserved both main's newer test coverage and this regression.
- This is the best narrow fix for #124555: plugin-harness dispatch reuses the existing shared sandbox skill-preparation owner. It does not add Codex-specific behavior or broaden capability.
- No breaking or backward-compatibility concerns: no public API, config, storage, protocol, or Plugin SDK contract changes. Native embedded agents, unsandboxed agents, and existing remote read-only skill mounts retain their prior behavior.
- Local proof: focused regression 4/4 passed; `check:changed` passed.
- Remote proof: AWS Crabbox Linux run [`run_80fa15039d69`](https://crabbox.openclaw.ai/portal/runs/run_80fa15039d69), focused regression 4/4 passed.

No remaining issue-scoped blockers.

sallyom:
Merged via squash.

- Prepared head SHA: [6409b763ea365add12c543cf23738071e0fdb9a9](https://github.com/openclaw/openclaw/pull/143291/commits/6409b763ea365add12c543cf23738071e0fdb9a9)
- Landed commit: [8880d708d562d6dc2e3f474a7ea98a0b1b70e5a8](https://github.com/openclaw/openclaw/commit/8880d708d562d6dc2e3f474a7ea98a0b1b70e5a8)

PR #143300, opened 2026-09-09
refactor(tests): share bound-turn client fixtures

steipete:
Related: #139428

## What Problem This Solves

Five conversation-binding tests repeat the same mock client while exercising different prompt and permission policies.

## Why This Change Was Made

Use a local factory that receives each test's existing request array and owns a separate notification-handler closure. It preserves fresh mocks, request-object identity, late handler lookup inside the existing setImmediate callback, and the existing no-op disposers. All scenarios and independent assertions remain unchanged. This removes 94 net test lines.

## User Impact

No production or conversation behavior changes. The tests retain coverage for fallback prompt text, sandbox resolution, network-proxy permissions, and custom-provider policy handling.

## Evidence

- Complete conversation-binding and both turn-router suites before and after: the same 129 ordered cases passed, with no skipped cases.
- Source expansion and capture-binding checks reproduce the original source bytes and syntax tree.
- Two temporary controls using real Vitest mocks and real setImmediate verify original/candidate identities, handler replacement and late registration, callback ordering, fresh notification objects, and unknown-method rejection.
- Mapped changed-file gate and fresh independent P2 review passed.

The scheduling checks preserve the existing test fixture's behavior; they do not assert a native-server timing guarantee.

### DIFF
--- a/extensions/codex/src/conversation-binding.test.ts
+++ b/extensions/codex/src/conversation-binding.test.ts
@@ -456,6 +456,37 @@ function conversationThreadStartResult(threadId: string, canAcceptDirectInput?: 
   };
 }
 
+function createCompletingBoundTurnClient(turnStartParams: Record<string, unknown>[]) {
+  let notificationHandler: ((notification: unknown) => void) | undefined;
+  return {
+    request: vi.fn(async (method: string, requestParams: Record<string, unknown>) => {
+      if (method === "turn/start") {
+        turnStartParams.push(requestParams);
+        setImmediate(() =>
+          notificationHandler?.({
+            method: "turn/completed",
+            params: {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1f0f7cadeb4fbac7d71ba65ecef27fcd8a3d66b3](https://github.com/openclaw/openclaw/pull/143300/commits/1f0f7cadeb4fbac7d71ba65ecef27fcd8a3d66b3)
- Landed commit: [9865df7610c858975899e9abc3699f9bdcb7d858](https://github.com/openclaw/openclaw/commit/9865df7610c858975899e9abc3699f9bdcb7d858)

PR #143302, opened 2026-09-09
fix(cli): retain local model resources through completion cleanup

steipete:
Related: #140674

## What Problem This Solves

Local model commands release their prepared model generation before sending the request. Overlapping commands can rebuild the same catalog, and callback or response-cancellation work can outlive the command result without retaining that generation.

## Why This Change Was Made

The local command uses the existing acquired model contract and admits its cleanup owner before preparation. It retains the lease through completion interpretation and actual callback/cancellation drainage. The existing command runtime owns that drainage independently of when the result becomes available.

## User Impact

Local model commands reuse their active prepared generation and finish cleanup before its final release. Existing output, errors, model/profile selection, unsupported-runtime handling and Gateway behavior remain unchanged.

## Evidence

- Three native regressions reproduced premature generation release during overlap, response callback and response-body cancellation; the provider-error control passed. All four pass with the repair, including fresh generation creation after final release.
- 277 relevant tests, the full changed-file gate and isolated Codex review through P2 passed. Build completed in 333.9 seconds.
- The actual built CLI completed synthetic loopback success/error/success commands in 1.365/1.598/1.622 seconds. Source and 12,109 built artifacts stayed unchanged, and all managed process trees terminated.
- Refreshed directly onto main, independently of the SDK PR: all three changed files are byte-identical to the reviewed/build-tested candidate, and 252 composition tests passed. General run-generation physical disposal remains a separate producer change.

### DIFF
--- a/src/cli/capability-cli.test.ts
+++ b/src/cli/capability-cli.test.ts
@@ -108,7 +108,9 @@ const mocks = vi.hoisted(() => ({
   loadModelCatalog: vi.fn<
     typeof import("../agents/prepared-model-catalog.js").readPreparedModelCatalog
   >(async () => []),
-  prepareSimpleCompletionModelForAgent: vi.fn(async () => ({
+  releaseSimpleCompletion: vi.fn(),
+  acquireSimpleCompletionModelForAgent: vi.fn(async () => ({
+    release: () => mocks.releaseSimpleCompletion(),
     selection: {
       provider: "openai",
       modelId: "gpt-5.4",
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0b5e973c7d0cb600647f84601a50a749b0d07a3f](https://github.com/openclaw/openclaw/pull/143302/commits/0b5e973c7d0cb600647f84601a50a749b0d07a3f)
- Landed commit: [0fadfb493a2afec97bc0d82802c612bcfd4dfaac](https://github.com/openclaw/openclaw/commit/0fadfb493a2afec97bc0d82802c612bcfd4dfaac)

PR #143303, opened 2026-09-09
feat(hooks): accept Standard Webhooks signatures on mapped hook paths

Potrock:
## What Problem This Solves

Resolves a problem where operators wiring a SaaS webhook producer into an OpenClaw hook could not do it directly: inbound hooks authenticate only with the shared bearer token, and most webhook producers cannot attach a custom `Authorization` header. The workaround today is a reverse proxy that injects the token, which leaves the Gateway trusting anyone who reaches the URL, or a transform module that re-implements signature checking by hand. Many producers already sign every delivery with the [Standard Webhooks](https://www.standardwebhooks.com/) scheme (`webhook-id`, `webhook-timestamp`, `webhook-signature`, the format Svix and Svix-compatible senders use), and the Gateway ignored it.

Concrete case: an Ambush news stream delivering signed events to `POST /hooks/ambush` to wake a Telegram agent. Nothing in `hooks.*` could accept it without a shim.

## Why This Change Was Made

A mapping can now declare the sender's signing secret:

```json5
hooks: {
  mappings: [
    {
      match: { path: "ambush" },
      action: "agent",
      messageTemplate: "News alert: {{payload.data.headline}}",
      deliver: true, channel: "telegram", to: "<chat-id>",
      signature: { scheme: "standard-webhooks", secret: "whsec_..." },
    },
  ],
}
```

On that path the Gateway verifies HMAC-SHA256 over the exact request bytes, rejects timestamps outside `toleranceSeconds` (default 300) as replays, and admits the request without the hook token. The token alone is not sufficient there: the signature is the sender's identity. `secret` accepts an array so a sender's rotation overlap keeps verifying. Paths without a `signature` mapping keep the existing token contract byte for byte.

Authentication boundary, as tightened across the four ClawSweeper rounds (`4049e548`, `d1f192d5`, `77e712f1`, `9dff2b35`):

- **Built-in endpoints are never signature-authenticated.** `/hooks/agent` and `/hooks/wake` always require the token; a signature that is valid for a mapped path buys nothing there, and bearer callers to those endpoints are unaffected.
- **A signed mapping owns its path exclusively.** It must declare an explicit custom `match.path` (not `agent`/`wake`), and no other mapping may match that path, by the same `match.path` or by omitting one. Config validation (`hooks.mappings[i].signature` issue) and mapping resolution both reject overlaps, so the mapping whose secret authenticated the request is the mapping that dispatches it.
- **Signing authority is re-resolved after the body arrives.** Verification uses the mapping and secret the live configuration holds once the upload completes, not the ones captured before the body-read await. An upload that outlives a secret rotation or a mapping removal is rejected, and the handler continues with that current configuration.
- **Signed replay identity comes only from signed bytes, scoped to the mapping.** For a signed delivery the replay key is the verified `webhook-id` under `signature:<mapping id>`; an added `Idempotency-Key`, a payload `idempotencyKey`, a bearer token the signed path never validates, or a URL alias (`news` vs `news/`) cannot mint a new identity.
- **Signing authority is fenced again before dispatch.** Transforms are asynchronous, so after `applyHookMappings` resolves the request re-checks the live mapping and re-verifies the signed bytes; a rotation or removal that lands while a transform runs answers 401 instead of dispatching under a revoked key.
- **Signed replay identity uses signed facts only.** For a signed agent dispatch the replay scope is the mapping id, its path, the verified `webhook-id`, and the item's original position in the signed fan-out array (carried on the action, so a transform that drops earlier items does not shift later identities). Rendered action values, which can embed unsigned inputs such as `{{headers.*}}` or `{{query.*}}`, no longer take part, so a captured request cannot re-run by varying an unsigned header.
- **Replay records outlive the signature window.** Signed replay entries are retained until the signed timestamp itself can no longer verify (`webhook-timestamp` + `toleranceSeconds`, which for a future-dated timestamp exceeds one window from receipt), never less than the generic five-minute floor, so a redelivery that is still cryptographically valid still replays.
- **Signed paths never share replay state.** Replay and wake-ledger keys include the mapping id and its path, and config validation rejects a signed mapping whose explicit `id` another mapping reuses.
- **Signed wake actions deduplicate per item.** Each fan-out item's verified `webhook-id` plus position is remembered for the signature's lifetime; a redelivery after the queue consumed a wake answers `eventOutcome: "duplicate"` for that item without enqueueing again, while items that never dispatched (an error mid-batch) still run on retry. Agent actions keep the existing replay cache.
- Request admission (token or signature, body read, throttle, body errors) lives in `src/gateway/server/hooks-request-auth.ts`; verification, secret decoding, path ownership, and the conflict check in `src/gateway/hooks-signature.ts`. Signature failures count against the same auth throttle as bad tokens. Unsigned paths still read the body through `readJsonBody`.

## User Impact

Operators can point any Standard Webhooks or Svix-compatible sender straight at a mapped hook path with real sender authentication and replay protection, no proxy header injection and no hand-written transform. Nothing changes for existing configs: the new `signature` field is optional, built-in endpoints and unsigned paths behave exactly as before, and a config that would make signed-path ownership ambiguous is rejected at validation and at hot reload rather than silently accepted. Config labels and help text cover the new keys. New response value: mapped wake actions may answer `eventOutcome: "duplicate"` for a redelivered signed event.

### Live behavior through the changed Gateway admission path (head `9dff2b35`)

Gateway built from this branch (`pnpm build`) and started with a throwaway config on a spare port: a signed `wake` mapping on `/hooks/ambush`, a second signed `wake` mapping on `/hooks/slow` whose transform module sleeps four seconds, a third signed `forEach` wake mapping on `/hooks/batch`, hook token set, Node 26.8.1, Ubuntu 22.04. Cases 1–12, 14, and 16–18 are curl with Python-computed signatures; case 13 opens a raw socket, sends the headers and the first 20 body bytes signed with secret B, rotates the mapping secret to A in the config file, waits for the Gateway's hot reload, then sends the rest; case 15 sends a fully valid request to `/hooks/slow` signed with C and rotates that mapping's secret to D while its transform is still running. Output verbatim (secrets redacted by the script):

```
1 signed (secret A), no token                              -> 200  {"ok":true,"mode":"now","eventOutcome":"queued"}
2 forged signature                                         -> 401  Unauthorized
3 no signature, no token                                   -> 401  Unauthorized
4 hook token only on the signed path                       -> 401  Unauthorized
5 valid signature, stale timestamp (-3600s)                -> 401  Unauthorized
6 tampered body (signed bytes differ)                      -> 401  Unauthorized
7 built-in /hooks/wake with the mapped signature           -> 401  Unauthorized
8 built-in /hooks/wake with hook token (unchanged)         -> 200  {"ok":true,"mode":"now","eventOutcome":"queued"}
--- revoke: rotate mapping secret A -> B in config, wait for hot reload
reload after 2s
9 signed with revoked secret A                             -> 401  Unauthorized
10 signed with current secret B                            -> 200  {"ok":true,"mode":"now","eventOutcome":"queued"}
11 redeliver case 10 (same signed bytes) after consumption -> 200  {"ok":true,"mode":"now","eventOutcome":"duplicate"}
12 same signed request via path alias /hooks/ambush/       -> 200  {"ok":true,"mode":"now","eventOutcome":"duplicate"}
--- in-flight revocation: start an upload signed with B, rotate B -> A during the upload, then finish
13 upload signed with B, secret rotated to A mid-upload    -> HTTP/1.1 401 Unauthorized
14 fresh request signed with the now-current secret A      -> 200  {"ok":true,"mode":"now","eventOutcome":"coalesced"}
--- delayed transform fence: request to /hooks/slow signed with C (4s transform), rotate C -> D while it runs
15 /hooks/slow signed with C, secret rotated to D mid-transform -> 401  Unauthorized
16 /hooks/slow signed with the current secret D            -> 200  {"ok":true,"mode":"now","eventOutcome":"queued"}
--- signed fan-out wake: two items, then the same signed delivery again
17 /hooks/batch signed with E (2 items)                    -> 200  {"ok":true,"mode":"now","eventOutcome":"queued"}
18 same signed batch redelivered                           -> 200  {"ok":true,"mode":"now","eventOutcome":"duplicate"}
--- misconfiguration: second mapping on the signed path is rejected
```

Gateway log for the same run (`[reload]` lines are the secret rotations and the rejected overlapping-mapping edit; the `signature-mismatch` after the second reload is case 13, and the `signing authority changed before dispatch` line is case 15, logged after the transform finished):

```
[hooks] hook ambush rejected: standard-webhooks signature signature-mismatch
[hooks] hook ambush rejected: standard-webhooks signature missing-headers
[hooks] hook ambush rejected: standard-webhooks signature missing-headers
[hooks] hook ambush rejected: standard-webhooks signature timestamp-out-of-tolerance
[hooks] hook ambush rejected: standard-webhooks signature signature-mismatch
[reload] config change detected; evaluating reload (hooks.mappings)
[reload] config hot reload applied (hooks.mappings)
[hooks] hook ambush rejected: standard-webhooks signature signature-mismatch
[reload] config change detected; evaluating reload (hooks.mappings)
[reload] config hot reload applied (hooks.mappings)
[hooks] hook ambush rejected: standard-webhooks signature signature-mismatch
[reload] config change detected; evaluating reload (hooks.mappings)
[reload] config hot reload applied (hooks.mappings)
[hooks] hook slow rejected: signing authority changed before dispatch
[reload] config reload skipped (invalid config): hooks.mappings.0.signature: hook mapping 2 can also match this signed path; a signed path must belong to one mapping
```

Case 14 reports `coalesced` because its rendered wake text matches the still-pending wake from case 10; admission and dispatch succeeded.

### Tests (repo vitest lanes, Node 26.8.1)

- `vitest.gateway-server.config.ts` → `server.hooks.test.ts`, `server.hooks-signature.test.ts`, `server-http.hooks-delivery.test.ts`, `server.hooks-admission.test.ts`, `server.hooks-session-mode.test.ts`, `server.hooks-lane.test.ts`, `server-http.hooks-request-timeout.test.ts`: **61 passed**. HTTP-to-dispatch coverage: signed without token → 200 and one dispatch; redelivery with the same `webhook-id` → 200 with no second run, also with a fresh `Idempotency-Key`, a different bearer token, and the `/hooks/ambush/` alias; token-only on the signed path → 401; wrong secret, tampered body, stale timestamp → 401 with no dispatch; a valid mapped signature on `/hooks/wake` → 401; token on `/hooks/wake` and `/hooks/agent` → 200; signed wake redelivery and alias → `eventOutcome: "duplicate"`, a new delivery id → `queued`; the same signed bytes re-sent with a different `x-variant` header feeding a `{{headers.x-variant}}` template → no second run; with `toleranceSeconds: 900`, a redelivery six minutes later (past the five-minute floor, with `Date.now` advanced) → still replayed, one run; a signed `forEach` wake batch → `queued`, its redelivery → `duplicate`.
- `vitest.gateway.config.ts` → `hooks-signature.test.ts`, `hooks-mapping.test.ts`, `hooks.test.ts`, `server/hooks-request-auth.test.ts`, new `hooks-mapping.fanout-index.test.ts`: **97 passed**. Includes the admission helper driven directly with a fake request: rotated secret between the initial config and the live config → 401 with a throttle failure recorded; signed mapping removed from the live config → 401; the post-transform fence answering 401 for a rotated or removed mapping while passing for the unchanged one, retention following a future-dated signed timestamp, a fan-out whose transform drops the middle item keeping indices 0 and 2, the conflict finder rejecting a shared id, the signed dispatch scope carrying only signed facts, the wake ledger's expiry and size bound, and the verifier, decoder, normalizer, path-ownership, and conflict-finder cases.
- `vitest.runtime-config.config.ts` → `config.hooks-signature.test.ts`, `config.hooks-module-paths.test.ts`, `schema.help.quality.test.ts`: **38 passed**. Config validation accepts one or several secrets, rejects unknown schemes, empty secret lists, signed mappings without `match.path` or on `agent`/`wake`, any second mapping that can match a signed path, and a signed mapping whose id another mapping reuses.
- `oxlint` and `oxfmt --check` clean on all changed files; `tsgo --noEmit` reports no errors under `src/gateway` or `src/config`.

Docs updated: `docs/automation/cron-jobs/webhooks.md` (new "Signed senders" section, troubleshooting row) and `docs/gateway/config-hooks.md` (mapping fields; ownership, re-validation, replay, and wake dedupe paragraph).

https://claude.ai/code/session_01CZ15jNgPUHdPsMw4wWon6v


Potrock:
@clawsweeper re-review

Addressed all three findings in `4049e548` (built-in endpoints excluded from signature auth, signed paths must be exclusively owned by one mapping, signed replay identity derived only from the verified `webhook-id` under the authenticating mapping) and replaced the evidence section with a live run through the changed admission path on a gateway built from this branch, plus the new HTTP-boundary tests.

Potrock:
@clawsweeper re-review

Round two addressed in `d1f192d5`: signing authority is re-resolved from the live config after the body arrives (in-flight rotation now rejects), signed replay identity is scoped to the mapping so URL aliases and unsigned headers cannot mint a new identity, and signed wake actions deduplicate on the verified `webhook-id` for at least the replay window. Evidence section rerun at this head, including the mid-upload rotation via a raw socket (case 13), alias replay (12), and consumed-wake redelivery (11).

Potrock:
@clawsweeper re-review

Round three addressed in `77e712f1`: signing authority is re-verified against the live config after asynchronous mapping work, so a rotation during a transform now rejects (live case 15 with a 4-second transform); signed agent replay identity is built only from signed facts (mapping, verified webhook-id, item index), so unsigned template inputs cannot mint a new identity; and signed replay records are retained for at least `toleranceSeconds` instead of the five-minute floor. Evidence section rerun at this head; lanes 90/60/37.

Potrock:
@clawsweeper re-review

Round four addressed in `9dff2b35`: fan-out actions carry their original item index so a transform dropping earlier items cannot shift later replay identities; signed wake dedupe is tracked per item; signed replay records live until the signed timestamp itself stops verifying (timestamp + tolerance), not just one window from receipt; and replay state is keyed by mapping id and path with unique ids enforced for signed mappings. Evidence rerun at this head with a signed fan-out case (17–18); lanes 97/61/38.

Potrock:
Update from our side: following the Discord discussion, we moved the Ambush → OpenClaw integration to sender-side bearer auth instead. Ambush webhook destinations can now carry custom request headers (psychobot-team/reflex#5860), so `Authorization: Bearer <hooks.token>` is sent to `/hooks/<mapping>` directly and the existing mapping + transform path covers the rest without any OpenClaw changes.

Leaving this PR open as a proposal for native Standard Webhooks signature verification, since that is still the gap for senders that only sign. Happy to rework it in whatever direction maintainers prefer, or close it if you'd rather track the idea as an issue.

PR #242, opened 2026-01-05
fix(telegram): respect routing.groupChat.mentionPatterns

regenrek:
Telegram group mention gating previously only recognized @botUsername. This updates Telegram inbound handling to also treat routing.groupChat.mentionPatterns as a mention trigger (consistent with identity.name-derived patterns), while preserving the existing behavior when bot username is unknown and no patterns exist.

Tests: vitest src/telegram/bot.test.ts

** IMPORTANT**
Long-term recommendation (not in this PR): move mention gating into a shared single-source module used by all providers, combining native mentions and routing.groupChat.mentionPatterns under one consistent rule set (including command bypass), with a cross-provider test matrix to prevent drift.

iScreen Shoter - Telegram - 260105230254

PR #143304, opened 2026-09-09
fix(exec): let commands omit unneeded secret proxy credentials

fuller-stack-dev:
## Landing paused: scope reassessment

This PR implements an explicit per-call escape hatch, not an automatic repair of the proxy/autoreview integration. The requester has challenged the per-call flag as the right product solution. Do not land this as the canonical fix while that concern is unresolved.

The affected-server proof required `secretEgress: false` and independently available inference authentication. It does not prove the unchanged default invocation or protected-store-only inference. Available bindings are the run's protected store snapshot, not evidence that a particular command uses those secrets.

Current published head `314b1afcc09dae9711ffdbb291e8d0cb6d0bcdba` also has an actionable CI failure: three existing schema guidance assertions require "omit to inherit". A local description-only correction passes those three tests plus the 69 schema tests, but is not yet published. Required CI is not green.

Related: #143231

## What Problem This Solves

Fixes an issue where Gateway commands that need no protected store secrets still receive authenticated secret-egress proxy URLs when the shared store contains protected secrets. Hardened tools such as the unchanged autoreview helper reject those URLs before their reviewer starts.

The zero-binding guard in #143231 does not cover this case: the store can have bindings even when this particular command needs none.

## Why This Change Was Made

Adds one optional Gateway exec argument, `secretEgress: false`. It omits protected store sentinels and managed proxy injection for that invocation. It never re-registers or empties the run's existing bindings, so concurrent/background secret-using commands keep their authorization.

The running proxy captures whether its traffic allowlist (including `[]`) or bypass-host routing requires it. The new option refuses under that policy instead of silently taking a direct route. Defaults, proxy authentication, destination checks, revocation, approvals, and autoreview remain unchanged. No persistent configuration or additional proxy is introduced.

This is standalone on main. It shares the captured-policy fact proposed in #143231, but does not depend on or modify agent-skills #242.

## User Impact

A Gateway command that needs no protected secrets can request:

```json
{
  "command": ".agents/skills/autoreview/scripts/autoreview --mode local --max-priority P2",
  "secretEgress": false
}
```

This is not a promise of an entirely credential-free process: ordinary store environment, inherited proxy/CA settings, managed GitHub identity, and other authentication remain unchanged. The option is Gateway-only and does not bypass required routing policy.

Autoreview still needs inference authentication. This option works when that authentication is already available independently, such as an inherited API key or native Codex login. It is not sufficient when the only inference key is a protected-store entry requiring proxy substitution, and it does not automatically change existing exec calls. The caller must request the opt-out.

## Evidence

- Before production edits, three new regressions failed for the intended reasons; the 24 existing store-environment tests passed.
- After the patch: 78 focused checks passed across store-env, pinned tool lifecycle, proxy server, proxy lifecycle, and real-exec cancellation/closure suites. The same five-file command passed on the affected Linux server (Node 26.7.0).
- Real local exec → unchanged autoreview, using an isolated store with a synthetic protected secret and a real authenticated proxy:
  - Default: exit 1 with the exact `unsafe credentialed or malformed proxy URL in HTTP_PROXY` error.
  - `secretEgress: false`: completed the full independent P2 review, exit 0, scoped-clean. This was not a dry run.
- Real TLS lifecycle coverage proves an existing registration still substitutes its secret after a no-secret invocation, and revocation still returns 407. Explicit empty/nonempty allowlists and bypass-host policy refuse the option.
- Final independent P2 review of the full local candidate: scoped-clean, exit 0. Core and core-test typechecks passed; changed-file lint/format, architecture guards, dead-export and import-cycle checks passed after correcting two lint findings.

### Affected-server reproduction

Ran in a task-owned, isolated checkout on the affected server, without restarting or changing the live Gateway, its approval policy, credentials, or proxy configuration. Its configured secret-egress policy was enabled with no traffic allowlist or bypass hosts.

The harness used real run admission, `createExecTool`, store projection, process launch, and a real authenticated secret-egress proxy. It put one synthetic protected secret in isolated state and invoked the exact unchanged helper command shown above:

| Invocation | Result |
| --- | --- |
| Default secret egress | Exit 1: exact `unsafe credentialed or malformed proxy URL in HTTP_PROXY` rejection |
| `secretEgress: false` | Exit 0: full Codex P2 review completed, scoped-clean, no actionable findings |

The server's existing source base was `4a2917f033d6e42c8d46788072c13c7789ecf759`. All nine candidate files were made byte-identical to PR head `1c74d9720c22a75b2ee61fee0087a9a332377127`; their ordered path/content SHA-256 digest was verified on both machines: `23abf116302c554cdfdc1db194b52b7eca82027dc9fca194791700a6adf2a298`. This was an isolated source run on the actual server, not an in-place deployment of the live Roboclaw agent.

The unchanged helper's SHA-256 also matched on both machines: `7602619edff7e1c567f5c17ff0d206b698c93805d9f2a9c2ced4bf3a90fe80ca`. No proxy variable stripping, TLS relaxation, autoreview patch, or production configuration change was used.

### Landing review follow-up

CI caught a schema-description budget regression (773 characters against the existing <550 constraint). The descriptions were shortened without changing that constraint. Schema and store-environment suites then passed 100/100, including preservation of an inherited inference API key. The follow-up reviewer identified ambiguous routing wording; it now explicitly says the call fails when routing is required. Runtime behavior remains identical to the affected-server-tested head above; only schema descriptions, one regression case, and clarification text changed afterward.

The affected-server review used existing API-key authentication independent of the synthetic protected store. No new key, login, proxy policy, or credential-store change was made. This proves the opt-out for that setup, not proxy-backed inference with no alternative credential.

### Maintainer decision

@fuller-stack-dev explicitly confirmed security ownership and conditionally directed landing if the fix is appropriate and validated. The subsequent question challenges the per-call opt-out as the correct product solution. That conditional landing direction is not being treated as acceptance of this narrower workaround. Required CI and product-fit review remain unresolved.

### DIFF
--- a/docs/tools/exec.md
+++ b/docs/tools/exec.md
@@ -24,6 +24,14 @@ Working directory for the command.
 Key/value environment overrides merged on top of the inherited environment.
 </ParamField>
 
+<ParamField path="secretEgress" type="boolean">
+Gateway only. Set `false` when this command needs no protected shared-store secrets:
+OpenClaw omits their sentinels and its secret-egress proxy credentials for this call.
+The call is refused if the running proxy has an explicit traffic allowlist (including
+`[]`) or bypass-host routing. Omitted/`true` keeps the default behavior. Ordinary
+store environment, inherited proxy/CA settings, and other authentication are unchanged.
+</ParamField>
+
 <ParamField path="yieldMs" type="number" default="10000">
 Auto-background the command after this delay (ms).
 </ParamField>
--- a/docs/tools/exec.md
+++ b/docs/tools/exec.md
@@ -30,6 +30,8 @@ OpenClaw omits their sentinels and its secret-egress proxy credentials for this 
 The call is refused if the running proxy has an explicit traffic allowlist (including
 `[]`) or bypass-host routing. Omitted/`true` keeps the default behavior. Ordinary
 store environment, inherited proxy/CA settings, and other authentication are unchanged.
+Keep the default if the command needs a protected-store API key; this option omits
+that key's sentinel too.
 </ParamField>
 
 <ParamField path="yieldMs" type="number" default="10000">
--- a/src/agents/agent-tools.scheduled-exec-target.test.ts
+++ b/src/agents/agent-tools.scheduled-exec-target.test.ts
@@ -104,17 +104,40 @@ describe("createOpenClawCodingTools scheduled exec target", () => {
 
     const pinned = pinExecToolTarget(source, { host: "gateway", ask: "always" });
     await pinned.prepareBeforeToolCallParams?.(
-      { command: "echo hi", host: "node", node: "remote", security: "full", ask: "off" },
+      {
+        command: "echo hi",
+        secretEgress: false,
+        host: "node",
+        node: "remote",
+        security: "full",
+        ask: "off",
+      },
       { hookContext: undefined },
     );
-    pinned.finalizeBeforeToolCallParams?.({ command: "echo hi", host: "node", ask: "off" }, {});
+    pinned.finalizeBeforeToolCallParams?.(
+      { command: "echo hi", secretEgress: false, host: "node", ask: "off" },
+      {},
+    );
+
+    await pinned.execute("scope", {
+      command: "echo hi",
+      secretEgress: false,
+      host: "node",
+      ask: "off",
+    });
+    expect(execute).toHaveBeenCalledWith(
+      "scope",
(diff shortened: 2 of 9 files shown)


fuller-stack-dev:
Superseded by the automatic repair merged in #143354 and openclaw/agent-skills#243. The replacement supports authenticated proxy inference and WebSocket forwarding without a per-call secretEgress opt-out. Protected-store-only inference was verified in an isolated checkout on the affected host. This escape-hatch proposal was not merged.

PR #143306, opened 2026-09-09
fix(cli): recognize portable managed completion source lines as installed

### LINKED ISSUE #143267: [Bug]: completion installer duplicates portable managed Bash source lines

soroush5:
Fixes #143267.

The installer didn't recognize portable `[[ -f ... ]]` source lines that use `${HOME}`, so it kept appending an absolute-path copy next to the user's managed line.

This change treats a portable guarded source line that resolves to the exact cached completion file as installed, and leaves it untouched instead of replacing it. Lines pointing anywhere else still get the normal install behavior.

Verified with two new tests in `completion-runtime.test.ts` (recognized + preserved, and a line pointing at a different file is still treated as missing). Full `completion-runtime` + `doctor-completion` suites pass (85 tests), oxfmt/oxlint clean.

## What Problem This Solves

`openclaw completion bash` duplicates the completion setup for anyone whose shell profile is managed by a dotfile manager (chezmoi, stow, Nix): the installer doesn't recognize a portable guarded source line like `[[ -f "${HOME}/.openclaw/completions/openclaw.bash" ]] && source ...`, so every run appends an absolute-path copy next to the user's line. Worse, the absolute path bakes in a temp deploy dir (`/tmp/openclaw-chezmoi*/home/...`), so the duplicate is broken on the next shell.

## Evidence

Real patched-CLI run against a temp HOME (mixed profile: portable line + stale slow `source <(...)` line, cache file present):

```
$ isCompletionInstalled → true
$ installCompletion → Updating completion...

--- .bashrc after ---
[[ -f "${HOME}/.openclaw/completions/openclaw.bash" ]] && source "${HOME}/.openclaw/completions/openclaw.bash"

$ bash --noprofile --norc -c 'source ~/.bashrc; complete -p openclaw'
complete -W 'status' openclaw
```

The portable line is preserved byte-for-byte, the stale slow hook is removed, no absolute-path copy is appended, and a fresh Bash registers the completion from the cache file. (Paths use a temp HOME; no private data involved.)

Tests: `completion-runtime.test.ts` + `doctor-completion.test.ts` — 87/87 pass, including new cases for single-quoted operands, quoted tilde, `..` segments, guard/source mismatch, and mixed-profile cleanup. `oxfmt`/`oxlint` clean, no new `as` assertions (baseline-ratchets safe). No other open PR touches these files.

Addresses the ClawSweeper review findings: quoting-aware operand resolution, `..` rejection, owned-hook cleanup preserved alongside portable lines, and a platform-independent fixture.

### DIFF
--- a/src/cli/completion-runtime.test.ts
+++ b/src/cli/completion-runtime.test.ts
@@ -321,6 +321,68 @@ describe("completion-runtime", () => {
     },
   );
 
+  it("recognizes a portable [[ -f ... ]] home-relative source line as installed and preserves it", async () => {
+    const homeDir = tempDirs.make("openclaw-portable-completion-home-");
+    const stateDir = path.join(homeDir, ".openclaw");
+
+    await withEnvAsync(
+      {
+        HOME: homeDir,
+        USERPROFILE: homeDir,
+        OPENCLAW_STATE_DIR: stateDir,
+        XDG_CONFIG_HOME: undefined,
+        ZDOTDIR: undefined,
+      },
+      async () => {
+        const cachePath = resolveCompletionCachePath("bash", "openclaw");
+        expect(cachePath.startsWith(homeDir)).toBe(true);
+        await fs.mkdir(path.dirname(cachePath), { recursive: true });
+        await fs.writeFile(cachePath, "complete -W 'status' openclaw\n", "utf-8");
+
+        const portablePath = `\${HOME}${cachePath.slice(homeDir.length)}`;
+        const portableLine = `[[ -f "${portablePath}" ]] && source "${portablePath}"`;
+        const before = `# dotfiles-managed\n${portableLine}\n`;
+        const profilePath = path.join(homeDir, ".bashrc");
+        await fs.writeFile(profilePath, before, "utf-8");
+
+        await expect(isCompletionInstalled("bash", "openclaw")).resolves.toBe(true);
+
+        await installCompletion("bash", true, "openclaw");
+
+        await expect(fs.readFile(profilePath, "utf-8")).resolves.toBe(before);
+      },
+    );
+  });
+
+  it("rejects portable source lines that point at a different completion file", async () => {
+    const homeDir = tempDirs.make("openclaw-portable-completion-other-");
+    const stateDir = path.join(homeDir, ".openclaw");
+
+    await withEnvAsync(
+      {
+        HOME: homeDir,
+        USERPROFILE: homeDir,
+        OPENCLAW_STATE_DIR: stateDir,
+        XDG_CONFIG_HOME: undefined,
+        ZDOTDIR: undefined,
+      },
+      async () => {
+        const cachePath = resolveCompletionCachePath("bash", "openclaw");
+        await fs.mkdir(path.dirname(cachePath), { recursive: true });
+        await fs.writeFile(cachePath, "complete -W 'status' openclaw\n", "utf-8");
+
+        const profilePath = path.join(homeDir, ".bashrc");
+        await fs.writeFile(
+          profilePath,
+          '[[ -f "${HOME}/.openclaw/completions/other.bash" ]] && source "${HOME}/.openclaw/completions/other.bash"\n',
+          "utf-8",
+        );
+
+        await expect(isCompletionInstalled("bash", "openclaw")).resolves.toBe(false);
+      },
+    );
+  });
+
   it("prints the same canonical reload hint used by Doctor and onboarding", async () => {
     await withBashCompletionHome(async ({ homeDir }) => {
       const cachePath = resolveCompletionCachePath("zsh", "openclaw");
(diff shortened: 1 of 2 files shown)


soroush5:
Closing this one — upstream has since landed its own portable-hook recognition (`isPortableCompletionSourceLine`, shell-gated with quoting/`..`/tilde handling), which covers what these two commits were doing and more. Rebasing on top would just be a duplicate, so I'm withdrawing. The `doctor` repair path keeping missing-cache source lines was the main thing I wanted preserved, and that's in there too.

PR #143307, opened 2026-09-09
fix: retain approval classification resources through cleanup

steipete:
Related: #140674

## What Problem This Solves

Resolves a problem where ambiguous approval classification releases its prepared model runtime before inference and transport cleanup finish. Concurrent classifications can rebuild a generation that should still be retained, and a timeout can leave callbacks running beyond their owner.

## Why This Change Was Made

Acquire the existing finite model contract and admit cleanup to the parent before preparation. Preserve the caller's result deadline while draining actual completion, response callbacks, and cancellation work before releasing the lease. This is the approval-classifier slice of #140674.

The closed-list shortcut, verified embedded route, exact provider/model/profile/auth matching, final route revalidation, and fail-closed policy remain intact. The ten-second timer still starts after route matching and clears after completion and revalidation, independently of drainage.

## User Impact

Approval classification keeps its model and authentication runtime available until its real work finishes, including after a timed-out verdict. Approval decisions, input limits, and timeout behavior are unchanged.

## Evidence

Three native regression cases fail against the original code and pass with the repair. They cover overlapping preparation, response-body cancellation, and credential rotation during pending verdicts. All 158 focused tests passed; changed-file gate coverage, isolated Codex review through P2, and the full build passed.

A plain Node run of the emitted classifier used real SQLite authentication state, the production verified-binding machinery, a synthetic plugin, and loopback HTTP. A real response callback remained held after the unchanged deadline returned `other` at 10,021 ms. Parent drainage, generation reuse, fresh preparation after release, and credential-rotation rejection passed. Strict process cleanup completed in 12.987 seconds; all 12,111 built artifacts remained unchanged. No pending mutation was executed.

These are local observations, not a comparative performance benchmark or paid-provider proof. Native cancellation coverage uses the declared hermetic test transport; the compiled proof uses the actual plugin callback and HTTP path. Broader run-registration physical disposal is a separate slice.

### DIFF
--- a/src/system-agent/approval-intent.resources.test.ts
+++ b/src/system-agent/approval-intent.resources.test.ts
@@ -1,0 +1,343 @@
+import fs from "node:fs";
+import { createServer, type ServerResponse } from "node:http";
+import { getAiTransportHost } from "@openclaw/ai";
+import { expect, it, vi } from "vitest";
+import { createDeferred } from "../../test/helpers/promise.js";
+import { resetPreparedModelRuntimeSnapshotsForTest } from "../agents/prepared-model-runtime.test-support.js";
+import { ModelRegistry } from "../agents/sessions/model-registry.js";
+import {
+  acquireSimpleCompletionModelForAgent,
+  completeWithPreparedSimpleCompletionModel,
+} from "../agents/simple-completion-runtime.js";
+import { readConfigFileSnapshot } from "../config/config.js";
+import type { OpenClawConfig } from "../config/types.openclaw.js";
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6080906cb375bdc77567f885fc8331148572ba67](https://github.com/openclaw/openclaw/pull/143307/commits/6080906cb375bdc77567f885fc8331148572ba67)
- Landed commit: [4693a96225abf4cfefd3b2922f3d5492959d7045](https://github.com/openclaw/openclaw/commit/4693a96225abf4cfefd3b2922f3d5492959d7045)

PR #143308, opened 2026-09-09
fix: retain exec review resources after timeout

steipete:
Related: #140674

## What Problem This Solves

Resolves a problem where exec auto review releases a prepared model before inference and transport cleanup finish. A preparation that completes after the timeout can also escape the caller's lifetime.

## Why This Change Was Made

Retain the existing acquired model contract through actual preparation, completion, response callbacks, and cancellation work. Admit the cleanup owner before preparation; separate the caller's result from drainage so late preparation releases without starting inference and existing timeout results stay prompt. This is the exec-reviewer slice of #140674.

The separate preparation and completion budgets, policy checks, prompt construction, parser, and input limits are preserved. Internal test injection follows the acquired contract; no public adapter or compatibility shim is added.

## User Impact

Exec review keeps its model and authentication runtime alive until cleanup completes, while preserving existing approval decisions and timeout fallback behavior.

## Evidence

Four final native regression cases fail against the original code and pass with the repair: overlap, late preparation, held response callback, and response cancellation. All 293 focused tests passed, including 4,096 seeded malformed responses, concurrency controls, and the fresh completion budget. Changed-file gate coverage, isolated Codex review through P2, and the full build passed. The SDK-removal audit also found and reproduced a stale security-floor test mock. Its acquired-contract migration preserves every assertion; all 26 owning tests, targeted types/lint, and fresh independent review passed.

A plain Node run of the built public exec-review SDK used genuine async provider authentication, actual loopback HTTP, response callbacks, and the configured guarded response body's cancellation. The five-second fallback returned in 5,004 ms for late preparation and 5,008 ms for a held callback. Callback failure returned in 9 ms while cancellation still retained its owner. Strict process cleanup passed in 12.450 seconds; all 12,111 built artifacts remained unchanged. No reviewed command was executed.

These are illustrative local samples, not a speedup benchmark, paid-provider proof, or activation of general registration disposal. Earlier proof-only module-path and ineffective global-fetch instrumentation failures were corrected without weakening production guards; the successful proof delegates through the actual guarded transport.


steipete:
Merged via squash.

- Prepared head SHA: [f029c43bbff60944bb88c778bc86a1f993009edc](https://github.com/openclaw/openclaw/pull/143308/commits/f029c43bbff60944bb88c778bc86a1f993009edc)
- Landed commit: [cc81bfade702bc14ad913704a22f9a5e337320a9](https://github.com/openclaw/openclaw/commit/cc81bfade702bc14ad913704a22f9a5e337320a9)

PR #143309, opened 2026-09-09
refactor: share Docker harness timeout fixture text

steipete:
Related: #139428

## What Problem This Solves

Four Docker harness tests repeat the same pass-through timeout executable, making identical fixture behavior harder to maintain.

## Why This Change Was Made

Share only the immutable script body. Each test still writes its executable through the same quoted heredoc, at the same point in setup, with its original permissions, PATH, shell scope, and assertions. This removes 34 net test lines.

## User Impact

No user-visible behavior or runtime improvement. Coverage for direct helper loading, stdin forwarding, termination cleanup, and caller-owned file descriptors remains unchanged.

## Evidence

All four selected cases passed before and after; 293 unrelated cases were filtered out in each run. All four generated shell scripts are byte-identical, including both cooked-template and String.raw consumers. Expanding the shared literal restores the original full source exactly, preserving full case identities and assertions. Existing Vitest display truncation affects three names in both logs.

Mapped changed checks and fresh independent review passed. Production helpers are unchanged.

### DIFF
--- a/test/scripts/docker-build-helper.test.ts
+++ b/test/scripts/docker-build-helper.test.ts
@@ -24,6 +24,20 @@ import { useAutoCleanupTempDirTracker } from "../helpers/temp-dir.js";
 
 const tempDirs = useAutoCleanupTempDirTracker(afterEach);
 
+const PASSTHROUGH_TIMEOUT_SCRIPT = `#!/usr/bin/env bash
+case "$1" in
+  --kill-after=1s)
+    exit 0
+    ;;
+  --kill-after=30s)
+    shift 2
+    ;;
+  *)
+    shift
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [eb70a11839e761e387870b368038a0a9eb635ff8](https://github.com/openclaw/openclaw/pull/143309/commits/eb70a11839e761e387870b368038a0a9eb635ff8)
- Landed commit: [b527661779cac96ef58d743b4da5ec5d83fa7913](https://github.com/openclaw/openclaw/commit/b527661779cac96ef58d743b4da5ec5d83fa7913)

PR #143065, opened 2026-09-09
fix(tts): release telephony provider resources after synthesis

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where telephony speech requests could leave newly loaded plugin resources open, or lose managed provider resources while configuration, preparation, or synthesis was still running.

## Why This Change Was Made

Own speech registrations for the complete finite telephony request, reusing the existing capability acquisition machinery. Share the canonical singular lookup policy while keeping configured catalogs separate from direct preference and override lookups; direct-only providers must not become automatic fallbacks.

Each attempt prepares its applicable configuration view. If runtime configuration refreshes while a primary provider is pending, fallback resolves the updated provider configuration while preserving the initial fallback order. Existing raw hosts, synchronous public SDK signatures, buffered synthesis and streaming retain their current contracts.

## User Impact

Telephony PCM and result metadata remain usable after cleanup. Normal provider failures remain structured results. Persona selection, voice overrides, timeouts, aliases and fallback ordering are preserved.

## Evidence

- Original production failed six native resource cases; the raw-host control passed. Eight final native cases now pass, including a configuration-refresh comparison that first exposed and then repaired stale fallback selection in the intermediate candidate.
- 265 speech, VoiceCall, capability and media siblings passed. Full eight-path checks, independent Codex review through P2 and ordinary build passed; build time was 263 seconds.
- The actual compiled public telephony runtime passed success/failure/success in cold, managed and raw modes. Genuine 8 kHz PCM was saved, decoded and round-tripped after cleanup. SQLite close markers, read-only reopening and every managed process-tree exit were verified.
- Illustrative call timings were cold 34.81/28.98/12.44 ms; managed 7.43/5.05/1.31 ms; raw 7.44/4.10/2.02 ms. These are three synthetic samples per mode, not a statistical speedup claim.
- No carrier/vendor request or real credentials were used. Upstream VoiceCall request preparation and escaping streams are separate ownership boundaries. All eight candidate files are byte-identical after rebasing onto current main.

The branch is refreshed onto the independent Telegram fixture repair in #143197. Post-refresh composition passed 111 tests covering native telephony resources/personas, capability resolution and installed-provider compatibility. The incoming provider-enablement predicate is scoped to each synchronous resolution operation; the ownership change continues to refresh configuration between attempts.

PR #143313, opened 2026-09-09
fix: retain buffered speech providers through synthesis cleanup

steipete:
Related: #140674. Extends the request owner landed in #143065.

## What Problem This Solves

Resolves a problem where buffered speech and host Talk synthesis use provider callbacks without retaining their registrations through synthesis, result metadata, and asynchronous cleanup.

## Why This Change Was Made

Use the existing telephony request owner for the shared buffered synthesis function. Refresh the applicable provider view for each attempt while keeping fallback catalogs separate from preference and override lookups. Completed audio buffers no longer need the provider: transcoding and file persistence continue after its release.

## User Impact

Buffered speech and host Talk retain fresh or managed provider resources until their actual work finishes. Existing raw hosts keep their ownership. Provider selection, persona behavior, secret boundaries, fallback policy, and audio output remain unchanged. Returned streams have a separate lifetime and are outside this PR.

## Evidence

The shared native SQLite fixture covers telephony, buffered speech, and Talk: 16 intended failures on original buffered code, 11 existing controls, then all 27 cases passing. Another 108 sibling tests passed. Changed-file gate coverage, isolated Codex review through P2, and the full build passed.

The built public TTS runtime completed success/error/success in cold, managed, and raw-host modes. It saved real WAV files and verified them with ffprobe and PCM roundtrips; managed resources disposed once, fresh resources disposed per operation, and raw resources remained with their caller. Actual SQLite connections closed and reopened correctly. All 11,380 built artifacts and 3,934 resolved files remained unchanged; every process group was joined.

Observed synthesis calls took 5–52 ms across these local samples. These are not comparative benchmark results. The proof uses synthetic providers and isolated state; it does not claim an external provider, carrier call, or live Gateway Talk session.

### DIFF
--- a/docs/plugins/sdk-overview/host-hooks.md
+++ b/docs/plugins/sdk-overview/host-hooks.md
@@ -61,11 +61,13 @@ completed video assets and result metadata. Video assets may contain buffers or
 provider-hosted URLs. Downloads after the call returns belong to the caller;
 registration disposal must not invalidate those completed artifacts.
 
-`api.runtime.tts.textToSpeechTelephony(...)` also owns fresh provider registrations
-through configuration, persona preparation, synthesis, and PCM result metadata.
-It retains managed registrations during the operation and preserves raw host
-ownership. Configured fallback catalogs stay separate from direct preference and
-override lookups. Returned audio buffers outlive registration disposal. The
+`api.runtime.tts.textToSpeechTelephony(...)`, buffered TTS, and host Talk speech
+also own fresh provider registrations through configuration, persona preparation,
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c297001fb046244c8f3c96b353c9463c21512156](https://github.com/openclaw/openclaw/pull/143313/commits/c297001fb046244c8f3c96b353c9463c21512156)
- Landed commit: [9b99c6113fe03bde15af4539fd6a9692c9fa252c](https://github.com/openclaw/openclaw/commit/9b99c6113fe03bde15af4539fd6a9692c9fa252c)

PR #143316, opened 2026-09-09
refactor(testing): inspect authoritative handoff state

steipete:
## What Problem This Solves

Managed-service handoff tests could validate the restart sentinel's JSON shadow even when the typed fields consumed by the Gateway were wrong. The updater fixture also hand-built only part of the sentinel row, keeping its setup coupled to that shadow representation.

## Why This Change Was Made

Read the authoritative typed row in the shared boundary fixture and the existing fallback ownership assertion. Missing state remains missing; invalid typed state fails the fixture instead of being treated as absence. Construct updater notifications with the existing parser and complete row builder, preserving the child's publication position and original revision timestamp.

Raw readers that prove exact preservation, malformed foreign state and revision-floor behavior remain. The already-landed timeout and process-lifecycle corrections from #143240 are not duplicated. Application production delta is zero; test support is +3 lines and the ownership test +4. No test case, platform condition, timeout or assertion site is removed.

## User Impact

No application behavior or persisted representation changes. The suite now detects routing/state defects through the same representation the Gateway reads, while preserving existing lifecycle, cancellation, recovery and foreign-state coverage.

## Evidence

- Fresh untouched baseline and candidate each passed 226 tests with the same one existing Windows-only skip. All 227 case names/statuses match across the four-file batch, in the original invocation order: ownership, canonical sentinel, update-result sentinel and lifecycle.
- The real triage-timeout case and every repair-authority row remain in that full run; the candidate triage case passed in 61.3 seconds under the existing cap.
- Separate current `update-run-abandonment` and `update-triage` suites passed all 55 cases. All 20 normally selected changed checks passed; source, context hashes and process cleanup were verified across the serial phases.
- A retained historical fault control removed the generated writer's typed session-key binding while preserving its JSON shadow. The old assertion passed; the typed assertion failed on the missing route. The unfaulted writer, store and reader are unchanged, but the current composed green proof above was run separately after the process fix landed.
- Independent Codex review is clean through P2. This is test-only consolidation using real child processes and typed persistence; no operator Gateway or live provider was invoked.

### DIFF
--- a/src/infra/update-managed-service-handoff-boundary.test-support.ts
+++ b/src/infra/update-managed-service-handoff-boundary.test-support.ts
@@ -7,10 +7,9 @@ import { waitForFile } from "../../test/helpers/process-wait.js";
 import { DEFAULT_VITEST_TEST_TIMEOUT_MS } from "../../test/vitest/vitest.timeouts.js";
 import { writeTriageUpdateFailure } from "../commands/triage-update.js";
 import { getFileLockProcessStartTime } from "../shared/pid-alive.js";
-import type { DB as OpenClawStateKyselyDatabase } from "../state/openclaw-state-db.generated.js";
 import { openOpenClawStateDatabase } from "../state/openclaw-state-db.js";
 import { resolveOpenClawStateSqlitePath } from "../state/openclaw-state-db.paths.js";
-import { executeSqliteQueryTakeFirstSync, getNodeSqliteKysely } from "./kysely-sync.js";
+import { readRestartSentinelRowSync } from "./restart-sentinel-store.js";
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b86849981bf61aa7af8c5595a3f6484b97b56b9e](https://github.com/openclaw/openclaw/pull/143316/commits/b86849981bf61aa7af8c5595a3f6484b97b56b9e)
- Landed commit: [75b169bf2e198bf6f74c8cee3016b96c9851a680](https://github.com/openclaw/openclaw/commit/75b169bf2e198bf6f74c8cee3016b96c9851a680)

PR #143324, opened 2026-09-09
fix: preserve staged path replacements during worktree cleanup

steipete:
Related: #139428

## What Problem This Solves

Fixes an issue where users removing a managed worktree can encounter Git index conflicts after staging a file-to-directory or directory-to-file replacement. Large deletion-heavy checkouts also spend substantial cleanup time updating the snapshot index.

## Why This Change Was Made

Order snapshot updates using membership read from the actual private index: missing indexed paths in reverse byte order, other indexed paths in their original order, then additions. This processes indexed file/directory conflicts before adding replacement paths. Missing observations only influence ordering; Git still reads current file contents. Snapshot publication, lease and capacity checks, and restore ancestry retain their existing behavior.

## User Impact

Managed worktrees with staged path replacements can be snapshotted and restored. Deletion-heavy snapshots avoid repeatedly moving index entries that will also be removed.

## Evidence

- Linux proof covered five complete owner suites with the same 96 case identities: original code passed 94 and failed the two new staged-transition regressions; the candidate passed all 96. The refreshed candidate passed all 96 again, preserving the upstream branch-inventory changes.
- Six real-Git boundary cases cover staged and unstaged transitions in both directions, tracked-file reappearance, and an untracked child disappearing and reappearing during a file-to-directory transition. Existing tests and the 180001-entry GC workload retain their assertions and deadlines.
- Global path reversal, forced removal of a reappearing file, and the previous all-missing ordering each produced their intended failures. The initial run stopped when its strict parser rejected Git's missing-blob wording and stack; that raw force-removal failure was independently reviewed. The first matrix remains recorded as incomplete, and a separate follow-up completed the outstanding all-missing control and verified candidate restoration. Native process groups were joined.
- With Git 2.53.0, the large-index GC case took 46.0 seconds on the original and 11.2 seconds on the candidate in one paired observation. The refreshed candidate took 12.7 seconds in the separate follow-up, without a fresh baseline. These are focused observations, not a whole-suite speedup measurement.
- The refreshed source passed the complete required changed gate and fresh P2 review.

### DIFF
--- a/src/agents/worktrees/service.provisioned.test.ts
+++ b/src/agents/worktrees/service.provisioned.test.ts
@@ -4,7 +4,8 @@ import fs from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { promisify } from "node:util";
-import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import * as commandRunner from "../../process/exec-runner.js";
 import { closeOpenClawStateDatabaseForTest } from "../../state/openclaw-state-db.js";
 import { insertRegistryWorktree } from "./registry.js";
 import { ManagedWorktreeService } from "./service.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e1bdd389b3f2c7f7fbdfd857395b8a1fbb32b343](https://github.com/openclaw/openclaw/pull/143324/commits/e1bdd389b3f2c7f7fbdfd857395b8a1fbb32b343)
- Landed commit: [a5351b8b2ade865cf16afbb7c93777ff76919c1d](https://github.com/openclaw/openclaw/commit/a5351b8b2ade865cf16afbb7c93777ff76919c1d)

PR #143226, opened 2026-09-09
fix(sessions): move cold cleanup integrity checks off the main thread

steipete:
## What Problem This Solves

Deleting sessions or cleaning lifecycle artifacts could temporarily block other Gateway work when the parent database connection was cold.

## Why This Change Was Made

Cold preparation now uses the existing asynchronous database owner inside the same writer queue. The initiating caller's live guard is checked before native-open work resumes, including index repair, schema work and registration; coalesced callers retain their own operation guards. Refusal preserves the original thrown value while joining owned cleanup, and distinct cleanup failures remain visible.

## User Impact

Full integrity and foreign-key checks for these cold preparation paths run off the main thread. Writer ordering, all required checks, archive/deletion order, and reclamation Worker settlement remain intact.

## Evidence

- The focused and surrounding SQLite owner/lifecycle suites passed (85 tests). The final inventory and owner follow-up passed 23 tests, required changed-file checks, and independent review through P2. Source-preserving rebases retain the reviewed production and regression-test bytes. ClawSweeper also reviewed the current head without actionable findings.
- A retained 35 MB synthetic cold-deletion fixture preserved all four full agent checks: one standalone integrity child and three identified reclamation Workers, with zero parent-main-thread full checks. All three archives and the unrelated filler payload were preserved. The integrity child and identified reclamation Workers closed, and managed process-group extinction was verified.
- A source-bound Linux Testbox run completed a fresh build in 352 seconds and the built-CLI Gateway flow in 19 seconds: health, create without a model turn, reset, delete, repeat missing delete, and health. Reset preserved the session ID and changed its lifecycle revision; delete returned true, the repeat returned false, and final session count was zero. All seven Gateway/CLI owners exited cleanly and proved descendant extinction and release. The downloaded evidence, complete selected source tree, fresh build, dist fingerprint, and transport carrier were independently audited.
- Linux proof selected source: `9d6e72d6c1112e500585205b469dbe5f0d679853`. [Testbox lifecycle workflow](https://github.com/openclaw/openclaw/actions/runs/34377128924): the native proof command exited zero; the workflow is canceled during lease cleanup. A nonfatal portal-sync warning and subsequent cleanup API timeout are retained separately from the successful workload.
- [CI for the reviewed head](https://github.com/openclaw/openclaw/actions/runs/34375441556) is green.

[...]

PR #143332, opened 2026-09-09
fix: prevent cold history planning from blocking the Gateway

steipete:
Related: #143226, #143285.

## What Problem This Solves

Resolves a problem where history disk-budget eviction planning can pause the Gateway's main thread when its cached parent database handle becomes cold while waiting for lifecycle ownership or archive materialization.

## Why This Change Was Made

Both history-planning callbacks now use the existing guarded database acquisition flow used by explicit session deletion. The helper is shared through the session database scope module. Warm acquisition stays synchronous; cold acquisition runs full integrity and foreign-key checks in the existing read-only child while retaining the same writer FIFO and lifecycle ownership. Candidate protection is read from the admitted database after validation.

## User Impact

These two planning stages no longer run cold full-file validation on the Gateway's main thread. Retention, deletion, full validation, current authority checks, and reclamation Worker policy are unchanged. Candidate-discovery and archive-pruning cold opens remain separate follow-ups; this change does not remove Worker validation costs or establish a quantified production speedup.

## Evidence

- Independent P0–P2 autoreview of the final rebased candidate found no actionable findings, including publication-borrow lifetime, current authority, FIFO ordering, and fresh protection reads.
- The final changed-file gate passes, including core and test typechecks, lint, dependency/boundary guards, and runtime import-cycle checks.
- On the original source, both warm controls passed and both cold boundaries completed real cleanup but failed the off-main-thread admission assertion. The regression preserves the original queued-writer order.
- All 27 focused tests pass after rebasing onto #143285, covering both cold boundaries, warm acquisition, protection changes during validation, pending-open revocation, and existing lifecycle/database-authority contracts.
- A frozen Linux Blacksmith Testbox run passed 70 tests across five suites and the full `pnpm build`. The wrapper verified the source/tree and all five changed-file hashes; the lease stopped successfully. This proof preceded a patch-identical rebase, plus two mechanical test lint fixes. [Run context](https://github.com/openclaw/openclaw/actions/runs/34389789780).
- The Linux tests exercised real synthetic SQLite history, archive materialization, reclamation Workers, publication, and remeasurement through the maintenance owner. The cold-boundary wrappers call the real implementations. No CLI parser, managed Gateway, or production data was exercised.

### DIFF
--- a/docs/reference/database-schemas/storage-changes.md
+++ b/docs/reference/database-schemas/storage-changes.md
@@ -38,13 +38,15 @@ admission. Publish live session changes and other dependent effects only after
 the durable write succeeds. A future network-backed owner must preserve that
 ordering while awaiting its driver.
 
-Explicit session deletion and lifecycle-artifact cleanup prepare their plans
-inside the session writer queue. When the parent database handle is cold, its
+Explicit session deletion, lifecycle-artifact cleanup, and history disk-budget
+eviction prepare their plans inside the session writer queue. When the parent database handle is cold, its
 existing asynchronous admission owner runs the full integrity and foreign-key
 checks in a read-only child, moving those full checks off the main thread while
 retaining that queue position. A supplied caller guard is rechecked before the open
 resumes into index repair, schema work, or registration, and before that caller
-uses the admitted handle. Coalesced callers retain their own guards.
+uses the admitted handle. Coalesced callers retain their own guards. History
+eviction also uses this admission when reopening after archive materialization,
+then rereads candidate protection before preparing reclamation.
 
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0167832aa05cd46fa214190ced4bc790cf2888f7](https://github.com/openclaw/openclaw/pull/143332/commits/0167832aa05cd46fa214190ced4bc790cf2888f7)
- Landed commit: [561d3199fd5f02027d89e388815381afec012c26](https://github.com/openclaw/openclaw/commit/561d3199fd5f02027d89e388815381afec012c26)

PR #143337, opened 2026-09-09
fix: keep multi-agent session lists working after model fallback

steipete:
Related: #141396

## What Problem This Solves

Fixes an issue where users could not load session lists or Workboard when a completed session had a model fallback and the list combined multiple agent stores. Writable installations could instead silently omit the completed fallback model and create an unintended empty database.

## Why This Change Was Made

Each listed row already carries its resolved physical store target. Pass that target to row projection for every session, keeping the aggregate path label solely as response metadata. This repairs the producer of the transcript-read scope and preserves the existing completed-model projection.

## User Impact

Multi-agent session lists load successfully and show the completed model from the correct agent transcript. The same shared listing path serves Workboard, embedded session tools, and the TUI.

## Evidence

- A live `sessions.list` request reproduced SQLite `CANTOPEN` on a sealed installation. Failed-open tracing showed the aggregate display label being interpreted as a relative database filename.
- The extended two-store regression writes actual completed transcript messages through the SQLite accessor. It fails before the fix because the completed model disappears, then passes with correct per-agent model results and no aggregate-label database creation.
- `node scripts/run-vitest.mjs src/gateway/server.sessions.store-paths.test.ts src/gateway/session-utils.test.ts`: 261 passed.
- `node scripts/check-changed.mjs -- src/gateway/session-utils-list.ts src/gateway/server.sessions.store-paths.test.ts`: passed, including core/core-test types, targeted lint and storage/import guards.
- Independent review: no actionable P0-P2 findings on the final staged candidate.
- Production change: +1/-1 line. No schema, configuration, or protocol change.

## Post-merge live verification

Deployed main `10e57bf8ab38e3e52ebd4bbb5c1dae5465c04191`, which includes merge `401f390d58b79022ad4eaa71633b025fda899c48`, through the canonical release owner. The deployment completed successfully.

Both the default `sessions.list` request and Workboard's exact lifecycle query returned populated session lists, including active-model metadata. `workboard.cards.list` also succeeded. The serving process and build identity remained stable throughout verification. Schema compatibility, enabled channel connectivity, and served Control UI asset parity passed.

The initial strict health sample reported CPU/event-loop utilization pressure after the session reads. A bounded follow-up on the same process passed strict channel and readiness checks with no degradation and no performance waiver. This is backend/RPC and static-asset proof; it does not claim authenticated rendered-browser coverage.

A further bounded log observation spanning more than three Workboard polling intervals found no database-open, session-list, or Workboard lifecycle failure on the same serving process.

### DIFF
--- a/src/gateway/server.sessions.store-paths.test.ts
+++ b/src/gateway/server.sessions.store-paths.test.ts
@@ -4,6 +4,7 @@ import { syncBuiltinESMExports } from "node:module";
 import path from "node:path";
 import { expect, test, vi } from "vitest";
 import * as sessionDirs from "../agents/session-dirs.js";
+import type { InternalSessionEntry } from "../config/sessions.js";
 import {
   appendTranscriptEvent,
   appendTranscriptMessage,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [126c3a40ed7cb1f46ac13339dbe8d1a5a40df35a](https://github.com/openclaw/openclaw/pull/143337/commits/126c3a40ed7cb1f46ac13339dbe8d1a5a40df35a)
- Landed commit: [401f390d58b79022ad4eaa71633b025fda899c48](https://github.com/openclaw/openclaw/commit/401f390d58b79022ad4eaa71633b025fda899c48)

PR #143338, opened 2026-09-09
refactor: share Docker package-builder fixture text

steipete:
Related: #139428

## What Problem This Solves

Two Docker harness tests repeat the same package-builder Node stub, making identical setup harder to maintain.

## Why This Change Was Made

Share only the immutable shell text. Each script still defines and exports its own Node function at the original point, preserving argument parsing, real-Node fallback, newline escaping, and late shell-variable lookup. This removes 28 net test lines.

## User Impact

No user-visible behavior or runtime improvement. Tests retain generated-package cleanup, caller-owned package preservation, failure status, diagnostics, and container cleanup coverage.

## Evidence

Both complete selected cases passed before and after; 295 unrelated cases were filtered out in each run. The two generated shell scripts are byte-identical, and expanding the shared literal restores the original full source exactly. Existing Vitest display truncation affects both names in both logs; full source identities and order remain unchanged.

Mapped changed checks and fresh independent review passed. Production helpers are unchanged; no live Docker or package build was needed.

### DIFF
--- a/test/scripts/docker-build-helper.test.ts
+++ b/test/scripts/docker-build-helper.test.ts
@@ -24,6 +24,38 @@ import { useAutoCleanupTempDirTracker } from "../helpers/temp-dir.js";
 
 const tempDirs = useAutoCleanupTempDirTracker(afterEach);
 
+const PACKAGE_BUILDER_NODE_SCRIPT = `node() {
+  local script="$1"
+  shift
+  if [[ "$script" != "$DOCKER_E2E_PACKAGE_LIB_DIR/../package-openclaw-for-docker.mjs" ]]; then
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e23bbe930953ce34140c88124b22176c9912cd84](https://github.com/openclaw/openclaw/pull/143338/commits/e23bbe930953ce34140c88124b22176c9912cd84)
- Landed commit: [85704ff8eb305a224378307fd26da42f29c2961e](https://github.com/openclaw/openclaw/commit/85704ff8eb305a224378307fd26da42f29c2961e)

PR #143339, opened 2026-09-09
test(infra): exercise real install workspaces

steipete:
The install-flow tests replaced the temporary-workspace owner with a shared fixed path and checked archive helper call shapes. They now use tiny real archives and the normal private workspace lifecycle, removing both test helpers and every archive/workspace replacement.

All six cases remain. The marker-root fixture includes a distracting child directory so it proves marker forwarding, readable callback input, result propagation, the requested workspace prefix, and cleanup after return. Unsupported input and an invalid flat archive retain the two distinct error paths and verify that neither invokes the callback. The existing extraction-limit/default case is unchanged.

Validation: untouched baseline 59/59; candidate plus archive-helper and workspace siblings 74/74; all 20 changed checks passed; independent Codex review clean. A missing-marker control failed at the wrong selected root, and the production file was restored exactly. No production change; tests net -11 lines.

### DIFF
--- a/src/infra/install-flow.test.ts
+++ b/src/infra/install-flow.test.ts
@@ -2,28 +2,9 @@
 import fs from "node:fs/promises";
 import path from "node:path";
 import JSZip from "jszip";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { withTestDir } from "../test-helpers/temp-dir.js";
-import * as archive from "./archive.js";
 import { resolveExistingInstallPath, withExtractedArchiveRoot } from "./install-flow.js";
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b4d13fef82b451f5cf357a63f3af8ed1a517a3a2](https://github.com/openclaw/openclaw/pull/143339/commits/b4d13fef82b451f5cf357a63f3af8ed1a517a3a2)
- Landed commit: [10e57bf8ab38e3e52ebd4bbb5c1dae5465c04191](https://github.com/openclaw/openclaw/commit/10e57bf8ab38e3e52ebd4bbb5c1dae5465c04191)

Issue #115367, opened 2026-07-28
Provider-owned read gate requires `origin: bundled`, but every privileged chat surface (slack/discord/matrix/msteams/feishu) now ships as an external plugin → reads locked to current conversation


josh-cornelius:
## Summary

The host-side conversation read gate introduced in the `2026.7.2-beta` line
(`enforceMessageActionConversationReadGate` — *"the sole host chokepoint before any
read-capable plugin callback runs"*) grants unrestricted **provider-owned** reads only to
channel plugins whose `origin === "bundled"`. But **every channel in the gate's own
privileged set now ships as an external plugin**, so the provider-owned path is
unsatisfiable in a stock image: any first-party chat surface installed the documented way
(`openclaw plugins install @openclaw/<channel>`) is locked to the *exact current
conversation* for all read-capable actions.

Net effect: a normally-installed Slack/Discord/Matrix/MS Teams/Feishu plugin cannot read
channel history, download files, search, list pins, etc. — even in the conversation it is
actively replying in — and throws:

```
Delegated slack:read requires the exact current conversation and account for this plugin.
```

## Environment

- Gateway: `2026.7.2-beta.5` (behavior is new in the `2026.7.2-beta` line — the gate,
  its enforcement resolver, and the privileged-channel set are **all absent in
  `2026.7.1`**).
- Channel plugin: `@openclaw/slack` installed via `openclaw plugins install @openclaw/slack`
  (reproduces identically for `@openclaw/mattermost`; by inspection the same trap applies
  to any member of the privileged set below).

## Reproduce

1. Run a `2026.7.2-beta` gateway from the published image.
2. Install a first-party chat plugin the documented way: `openclaw plugins install @openclaw/slack`.
   `openclaw plugins inspect slack` reports `Origin: global`.
3. From an agent bound to that channel, invoke `message action=read` (or `download-file`,
   `search`, `list-pins`, `thread-list`, …) targeting the current channel/thread.
4. The call throws `Delegated slack:read requires the exact current conversation and
   account for this plugin.` Reads of any other conversation, and most reads of the
   current one, fail the same way.

## Root cause

The gate is a trust boundary that uses **install location as the proxy for trust**, and
that proxy broke when the first-party chat plugins were moved out of the image.

Source pointers (symbol names as they appear in the shipped `dist`):

- `resolveMessageActionReadEnforcement()` returns `{ kind: "provider-owned" }` **only when**
[...]

PR #143341, opened 2026-09-09
fix: restore context reads for installed official Discord and Slack plugins

anyech:
Related: #115367

## What Problem This Solves

An agent in one conversation can need messages, pins, or reactions from another conversation that its configured Discord or Slack account is allowed to read. The same provider implementation could serve that request when bundled, but installing the official plugin through npm or ClawHub caused the host to reject it before the provider's destination checks could run.

The existing provider-owned read gate admitted bundled registrations only. Extending admission also requires keeping the operation attached to the caller and plugin instance that authorized it: target lookup, queued requests, retries, and result delivery must not continue with authority that has expired or been replaced.

## User Impact

With matching updated core and plugin versions, verified official installations can perform the following provider-permitted context reads:

| Plugin | Actions covered |
| --- | --- |
| Discord | `read`, `search`, `reactions`, `list-pins`, `thread-list`, `channel-info` |
| Slack | `read`, `reactions`, `list-pins` |

Account matching, current-provider context, requester context, and the provider's destination policy still apply. Bundled plugins retain their existing admission rules and receive the same lifetime checks for these opted-in reads. Other actions, unverified installations, and older adapters keep their existing authorization behavior.

Delegated Discord reaction reads also stop using the resolver that can create a DM. A user alias in the current trusted DM resolves to its existing native channel; an unmatched or unattested user target fails before provider traffic. Direct operator reads and reaction writes retain their existing resolution path.

## Why This Change Was Made

The host now combines the existing loader-verified official installation record with an explicit `readAuthorityActions` list and the adapter's `providerOwnedReadGates` declaration. The host intersects that list with the six actions above. This keeps lifetime participation and provider admission explicit, and prevents a later host action addition from silently expanding an existing adapter's opt-in. Plugin-supplied fields cannot substitute for the selected registration's authority.

Local message tools capture the existing operational caller assertion before asynchronous preparation and retain the original turn capability. That assertion is carried through routing and dispatch; Gateway requests retain their server-owned caller assertion as well. Opted-in delegated execution needs a live registered instance, including for bundled plugins. A bundled artifact fallback or an omitted scoped registration cannot supply that lifecycle.

The shared `captureChannelReadAuthority()` SDK callback is captured when each transport request is submitted, retained through queues and retries, and checked immediately before provider I/O after asynchronous request preparation. Host and installed SDK copies share the invocation scope. The host also checks completion, covering stale results and errors, and closes captured callbacks when the action finishes. These checks stop subsequent requests and stale delivery; they cannot undo a request already sent.

The complete target-alias owner repair in [#147492](https://github.com/openclaw/openclaw/pull/147492) is included in the integrated proof and remains a separate change.

The new SDK export is absent from the published 2026.9.3 and 2026.9.4 cores. Core and plugin releases therefore need coordinated version and plugin-API alignment. Package proof uses the canonical release tools in a disposable 2026.9.5 fixture, with matching build metadata and no version change in this PR.

## Evidence

- Current integration `6f7be4b8`, based on main `8c776697`, passed all 1,581 affected tests across 44 selected files. Native reports verify every case and file, successful exits, unchanged source, and joined process cleanup. This includes all 15 settings-storage cases and the current Slack probe cleanup, profile authorization, and real-child state preflight paths.
- The 35 real HTTP read-authority cases cover installed and bundled success, destination denial, account mismatch, plugin and caller revocation, local message-tool preparation, queued retries, stale results, and Discord endpoint lookup boundaries. All 11 Discord reaction cases pass, including current-DM aliases and mismatched or unattested targets with zero provider traffic. Registration and dispatch coverage exercises all six host action classifications and the selected registration's lifetime. The added caller-lifetime cases fail on the original PR source.
- Fresh canonical core, Discord, and Slack packages built from `6f7be4b8` in a disposable 2026.9.5 fixture. All 41 campaign stages passed. Generated build metadata, plugin-API versions, the SDK export, package contents, compiled bindings, and the six published baseline tarball identities were verified. No release was published.
- Each actual installed 2026.9.3 and 2026.9.4 CLI completed `update --tag <candidate-core-tarball> --yes --no-restart` against those fresh packages. Channel configuration, workspace, plugin entries, and native install records are preserved. Both provider reads pass after each upgrade: four current installed reads. The three fixture processes and provider servers stopped; the source stayed unchanged and the final task-process inventory was empty.
- The separately retained `e407b697` package campaign passed 83 stages and 31 installed cases. Its 24 ordinary npm/official-ClawHub cases cover allowed reads, destination denial, account mismatch, plugin revocation before I/O and before delivery, and real caller-claim revocation. Three installed Slack HTTP/1 proxy cases cover an allowed read and caller/plugin revocation. Four post-update reads complete that matrix. Actual installation, integrity validation, native records, loading, dispatch, proxy transport, and updater commands ran against synthetic registry/provider servers.
- Published 2026.9.3 and 2026.9.4 cores lack the SDK export and reject both fixture plugins at the API compatibility gate. All four retained rejection checks preserve the old plugin records and channel/plugin configuration. These results keep their original source identities; the current build and upgrade results are recorded separately.
- All five parent patches are range-diff equivalent to the accepted `e407b697` review. The integration is conflict-free and matches its expected tree. Current-main caller, probe-dispatcher, package-metadata, update, and plugin-runtime changes were inspected for compatibility. [ClawSweeper's current-head review](https://github.com/openclaw/openclaw/pull/143341#issuecomment-5607361304) rates the patch Platinum with no actionable code findings.
- [Current CI](https://github.com/openclaw/openclaw/actions/runs/34815826893) passed 147 jobs, including the Gateway startup/reload regression. Its lint stripe and aggregate gate fail only on two unrelated UI test line-count violations. Direct native lint on the unchanged CI baseline reproduces the exact 702/700 and 1,010/1,000 failures; the parent changes neither file nor the rule or checker. All other current-head checks have completed successfully or were skipped/neutral.


anyech:
Updated in `4741ea7c`: the existing integration test now uses real HTTP for both Discord and Slack (12/12). The PR body includes actual CLI installation/loader evidence and the installed-provider authorization/revocation results: 10/10 in fresh processes, plus 10/10 in one process with matching host runtime snapshots. No additional product compatibility shim or runtime switch was added.

One hosted CI job currently needs a maintainer rerun: [checks-node-compact-large-18](https://github.com/openclaw/openclaw/actions/runs/34403259665/job/102640277722). The unchanged `code-mode-headless.test.ts` deadline test received `internal_error` instead of `timeout`; the whole file passes locally (61/61), and [the same job on the preceding head passed](https://github.com/openclaw/openclaw/actions/runs/34393360724/job/102608427525). This looks timing-sensitive, not a proven fix: no assertion or timeout was relaxed. The rerun API rejected the author's request with HTTP 403 (repository admin rights required).

@clawsweeper re-review

shakkernerd:
Fixed: provider-permitted context reads now work with verified official Discord and Slack installations while preserving caller and plugin lifetime checks.

With thanks to @anyech.

PR #143347, opened 2026-09-09
fix(doctor): preserve legacy Claude conversations during migration

obviyus:
Related: #136257.

## What Problem This Solves

Doctor left legacy conversation IDs outside canonical bindings while runtime reads and saves still used the retired field. Repair could also restore a deliberately cleared stale conversation or disrupt a locked session.

## Why This Change Was Made

Doctor copies safe legacy IDs before removing the old field, and runtime reads and saves change together. Canonical bindings and resume metadata retain precedence. Locked rows change only their binding representation. Stale-route cleanup also clears the legacy source so later migration cannot restore it; ambiguous records retain warnings.

## User Impact

Run openclaw doctor --fix after upgrading and before resuming sessions that have only the old field. Configured and locked conversations retain their IDs and metadata. Stale conversations remain cleared, and normal saves no longer recreate the retired field.

Consumers: Doctor migration, runtime conversation lookup, normal saves, and stale-route repair now use canonical bindings consistently.

## Evidence

The original regression missed the canonical binding after repair. Public Doctor repair, session listing, and repeated repair passed on ten synthetic sessions across two agents. Reopened databases preserved IDs, metadata, and locked ownership; stale state stayed cleared on repeat. Native checks passed; continuous integration covers migration, save, fork, and Doctor-to-history cases. No live conversation or credential operation was tested.


obviyus:
Landed as 227e2f4fe1c279f429138bce53d5015a8a398f7d.

**Proof:** The compiled public Doctor flow repaired 10 synthetic sessions across two agents, listed them, reopened both databases, and repeated repair without changing the resulting binding state. Independent acceptance passed all six stored-state clauses. Exact-head local review passed with proof A, overall B, no findings, and no security concerns. [Required CI passed](https://github.com/openclaw/openclaw/actions/runs/34397549548), including ordinary/locked Doctor-to-history, binding save/fork, and stale-cleanup cases.

| Before | After |
| --- | --- |
| Old Claude IDs depended on a single-provider fallback. | Doctor migrates safe IDs; runtime lookup and normal saves use provider-keyed bindings. |
| Locked rows could retain bytes but lose resume/history lookup. | Their bindings migrate while identity, ownership, route, and recency stay intact. |
| Stale cleanup could be followed by obsolete-ID resurrection. | Cleanup removes the old migration source too; stale conversations stay cleared. |

Run `openclaw doctor --fix` after upgrading before resuming legacy-only sessions. Ambiguous records remain saved with reconciliation warnings. No live provider conversation or credential operation is claimed. The earlier failed proof and CI results remain retained; all 13 landed file blobs match the reviewed head.

PR #143349, opened 2026-09-09
fix(crabbox): report completed source cleanup after capture cancellation

steipete:
## What Problem This Solves

Fixes an issue where stopping a cloud session during project image capture reported worker cleanup as failed even after Crabbox confirmed the source VM was stopped.

## Why This Change Was Made

After successful source cleanup, report the optional capture error as a warning. Failed source cleanup still rejects. The uncertain capture reservation and the fence preventing node enrollment remain intact, so successful VM cleanup never claims that an interrupted snapshot was resolved.

## User Impact

Stop and reclaim can finish when the worker is actually gone. Operators still receive the existing guidance for reconciling an uncertain image capture.

## Evidence

Observed a real canceled project capture where the source lease reached released/cleanup-complete while the Gateway reported an unresolved-capture teardown error. The task-owned snapshot was subsequently identified and deleted; capture recovery state was retained.

The regression failed before the fix. Its successful-stop and failed-stop cases verify allocation retirement, retained capture uncertainty, no repeated capture, and no node enrollment. All 132 focused provider tests passed, the complete changed-file gate passed, and independent isolated P0–P2 review was clean. Existing warm-image documentation already specifies this behavior.

### DIFF
--- a/extensions/crabbox/src/crabbox-worker-project.test.ts
+++ b/extensions/crabbox/src/crabbox-worker-project.test.ts
@@ -504,6 +504,52 @@ describe("Crabbox project snapshot provisioning", () => {
     },
   );
 
+  it.each([false, true])(
+    "preserves capture uncertainty but reports the source cleanup result after cancellation (stopFails=%s)",
+    async (stopFails) => {
+      const controller = new AbortController();
+      const { options } = projectOptions([], controller);
+      const leaseId = operationLeaseId("cancelled-project-capture");
+      const { provider, calls, warn } = createWarmProvider(({ argv }) => {
+        if (argv[2] === "create") {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [745c6a56869840a46432e9a37de7b81e13956323](https://github.com/openclaw/openclaw/pull/143349/commits/745c6a56869840a46432e9a37de7b81e13956323)
- Landed commit: [d48d7c2a8d02ceebeca3ec0d35210c837eb7e2ba](https://github.com/openclaw/openclaw/commit/d48d7c2a8d02ceebeca3ec0d35210c837eb7e2ba)

PR #143365, opened 2026-09-09
fix(browser): prevent premature existing-session timeouts

steipete:
## What Problem This Solves

Fixes an issue where existing-session browser waits and evaluations could lose their transport before their advertised execution time elapsed. Regression cases demonstrate a default wait being cut off at 25 seconds and a requested 180-second evaluation being cut off at 125.25 seconds.

Nested action calls and navigation probes also renewed their individual timeouts, while cancellation could stop at a DNS lookup or disappear during final URL handling.

## Why This Change Was Made

The existing-session route now owns one overall deadline spanning tab preparation, action execution and response completion. Action steps share an execution window; navigation verification shares a separate bounded allowance. Cancellation reaches the existing MCP and DNS mechanisms, and late completion cannot dispatch another action step or produce success.

The shared transport calculation covers both browser backends while retaining Playwright's sequential action and batch accounting. Canonical timeout-capability metadata avoids misclassifying valid inputs before route normalization. Existing-session wait and navigation helpers now have a separate owning module.

## User Impact

Existing-session waits and evaluations can use their supported execution budgets. Filling and submitting with `type` share one action budget instead of receiving a fresh timeout for each internal step. Navigation probes share their verification allowance, and cancellation stops waiting on DNS through the existing resolver. The CLI reference documents these bounds.

## Evidence

The frozen runtime candidate passed 578 owner and sibling tests across 14 files on Linux with supported Node 24.19 and pnpm 12.3.4. Before/after cases cover transport truncation, action and verification cancellation, DNS and final URL cancellation, slow tab preparation, and normalized click inputs. Existing Playwright, CLI, client and tool cases are included.

Independent review completed with no accepted findings. A whitespace-input concern was rejected after real client/route checks confirmed that normalization removes blank text before execution. All 27 final changed-file check stages passed after removing an obsolete lint suppression and its baseline entry. Runtime and test bytes are unchanged from the 578-test candidate. These are synthetic client/route tests with mocked Chrome MCP operations, not a live browser session.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -90,7 +90,7 @@ extensions/browser/src/browser/pw-tools-core.snapshot.ts	1
 extensions/browser/src/browser/pw-tools-core.state.ts	2
 extensions/browser/src/browser/routes/agent.act.normalize.ts	1
 extensions/browser/src/browser/routes/agent.act.shared.ts	3
-extensions/browser/src/browser/routes/agent.act.ts	2
+extensions/browser/src/browser/routes/agent.act.ts	1
 extensions/browser/src/browser/routes/agent.shared.ts	2
 extensions/browser/src/browser/routes/agent.snapshot.ts	8
 extensions/browser/src/browser/routes/agent.storage.ts	2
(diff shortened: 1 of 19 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e4f6915d33d1c93170c960182b7ddb8a4ae6cf4a](https://github.com/openclaw/openclaw/pull/143365/commits/e4f6915d33d1c93170c960182b7ddb8a4ae6cf4a)
- Landed commit: [edfcd0ae70cb807e108eac8a4693f8e0864ce1ed](https://github.com/openclaw/openclaw/commit/edfcd0ae70cb807e108eac8a4693f8e0864ce1ed)

PR #143366, opened 2026-09-09
fix: retain side-question resources through runtime cleanup

steipete:
Related: #140674

## What Problem This Solves

Resolves a lifetime gap where `/btw` keeps using a published runtime without retaining its resources through the side question and cleanup. That prevents safe retirement of managed generations while admitted side work is still running.

## Why This Change Was Made

Acquire the exact generation at the synchronous published-owner read, then retain it through side-question output, harness cleanup, and tracked descendants. Existing publication, standalone activation, configuration replacement, model selection, and catalog policy remain unchanged. The shared read policy moves into one internal module.

Parent cancellation and normal cleanup run in the captured generation context. The caller can receive its result before physical cleanup finishes, while the parent work owner still joins that cleanup.

## User Impact

Side questions retain their prepared resources until admitted work finishes, including during replacement and cancellation. This prepares the path for managed resource disposal; ordinary configured/raw producer disposal is not enabled by this PR. No public option, backend policy, or database schema changes.

## Evidence

Four lifetime cases fail against both original production files with premature SQLite closure. The combined case uses a real acquired plugin registration, the actual published loader and BTW harness path, delayed cleanup, replacement, exactly-once disposal, and database reopening. A test-only input adapter selects the existing owned read-only source; it does not enable configured disposal.

Final validation passed 81 BTW/native tests, 156 affected sibling tests, source-affected type/lint/Knip and remaining guard coverage, fresh isolated Codex review through P2, and a full build in 243.30 seconds. Counts overlap. Regression coverage includes parent MCP cancellation and normal cleanup context.

A strict plain-Node compiled run passed in 2.886 seconds with 12,120 unchanged artifacts and no application TypeScript/source fallback. It exercises actual built BTW on normal standalone publication, cancellation and cleanup context, plus actual built owned read-only published borrowing with SQLite close/reopen. Combined managed-source BTW composition is covered by the maintained native test. Cold BTW admission measured 241.82 ms; a warm call plus cleanup took 3.69 ms. These are local samples, not a comparative performance benchmark.

After rebasing onto current main, all five source files match the reviewed proof and all 81 composition cases pass again. No real external provider, packaged CLI, or native Codex execution is claimed; existing Codex interrupt/unsubscribe contracts were inspected and backend behavior is unchanged.

### DIFF
--- a/src/agents/btw.test.ts
+++ b/src/agents/btw.test.ts
@@ -1,5 +1,6 @@
 /** Tests BTW side-question execution, session context, auth, and harness routing. */
 
+import { DatabaseSync } from "node:sqlite";
 import { expectDefined } from "@openclaw/normalization-core";
 import { MAX_TIMER_TIMEOUT_MS } from "@openclaw/normalization-core/number-coercion";
 import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e794c8ded20d52e5ac105479981cb83283121fb2](https://github.com/openclaw/openclaw/pull/143366/commits/e794c8ded20d52e5ac105479981cb83283121fb2)
- Landed commit: [c3f242fe0a79062b7417b9f93c5ac10653ee06a0](https://github.com/openclaw/openclaw/commit/c3f242fe0a79062b7417b9f93c5ac10653ee06a0)

PR #143369, opened 2026-09-09
fix: retain speech providers until stream cleanup finishes

steipete:
Related: #140674. Builds on telephony #143065 and buffered speech #143313.

## What Problem This Solves

Resolves a problem where speech streams outlive the provider registrations they use, or leave resources behind when metadata fails, cancellation is delayed, or an unopened stream has already ended.

## Why This Change Was Made

Transfer the existing owned speech acquisition to the returned stream. Retain it through metadata, reads, source cancellation, provider release, and tracked producer work. Start cancellation before waiting for reads it must unblock, preserve the last queued chunk, and finalize outside the work being drained.

Providers with an explicit release callback keep their existing contract: callers await release after EOF or read failure. Providers without one clean up automatically, including empty or failed unopened streams and a consumed final queued chunk. Failed metadata projection cleans up before fallback; errors after bytes have escaped do not start another provider.

All three synthesis paths now require an owned prepared registry. The unused raw setup function and its type-only coupling are removed; public synchronous provider lookup contracts remain unchanged.

## User Impact

Streaming speech retains its provider resources until actual cleanup finishes. It preserves explicit release, cancellation, provider selection and raw-host ownership, without adding a release obligation for providers that previously had none. Completed audio remains usable after disposal.

## Evidence

Native regressions cover retirement, cancellation, blocked producer work, metadata/getter failures, cleanup-error precedence, unopened termination, and final-chunk delivery. Final combined testing passes 48 distinct cases across streaming, telephony, buffered speech and Talk. The preceding ownership candidate also passed 141 sibling/E2E cases, including Discord voice and provider binary-stream controls.

Affected checks, fresh full-candidate Codex review through P2, and the final ordinary build passed. The final build took 298.41 seconds and emitted a non-fatal UI startup gzip warning of 29 bytes over its threshold; no UI source, baseline or guard was changed.

Both actual compiled public APIs passed on one sealed build: 21 streaming requests and nine buffered calls across cold, managed and raw ownership. They exercised real WAV/PCM roundtrips and SQLite reads through metadata, cleanup and delayed work, followed by close/reopen. All six process groups terminated; all 11,390 artifact hashes remained unchanged, with no application source fallback. These use synthetic local providers, not an external speech vendor, carrier or operator Gateway.

An additional 32 Web-reader cases covered pending reads, timed/gated cleanup and cleanup failures, preserving final chunks and errors. Timings are local observations, not a comparative speedup claim.

### DIFF
--- a/docs/plugins/sdk-overview/host-hooks.md
+++ b/docs/plugins/sdk-overview/host-hooks.md
@@ -71,6 +71,15 @@ transcodes and saves those completed buffers after releasing the provider. The
 separate synchronous speech lookup and request-preparation APIs keep their
 existing caller lifetime; streaming speech is not part of this finite operation.
 
+Streaming speech owns its provider registrations until stream cleanup finishes.
+`api.runtime.tts.textToSpeechStream(...)` preserves an explicit provider
+`release()` callback's lifetime after EOF or a read error: callers must still
+invoke and await the returned `release()`. Without a provider release callback,
+EOF or a read error releases registrations automatically. Stream cancellation
+and explicit release start source cancellation and provider cleanup before
+joining both, including tracked producer work. Release also handles an unopened
+stream. Existing raw host registrations keep their host lifetime.
+
 For `image_generate`, `music_generate`, and `video_generate` tools prepared from an owned inspection,
 resources remain held through preflight and, once accepted, through generation, media saving, and
 any rollback. A `started` result acknowledges acceptance; it does not mean the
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [82c780cf6b17f3303c59aea95c976809b9ac1ed5](https://github.com/openclaw/openclaw/pull/143369/commits/82c780cf6b17f3303c59aea95c976809b9ac1ed5)
- Landed commit: [d77cb0a6e8d37aef6caafbb0c6f474aa70c2aa13](https://github.com/openclaw/openclaw/commit/d77cb0a6e8d37aef6caafbb0c6f474aa70c2aa13)

PR #143377, opened 2026-09-09
fix: retain Anthropic requests through cancellation cleanup

steipete:
Related: #140674.

## What Problem This Solves

Resolves a problem where an aborted Anthropic request can release its prepared runtime while the response body is still cancelling. Cancellation can also run in the aborting caller's context instead of the request's original context.

## Why This Change Was Made

Start body cancellation once, in the parser's original continuation, and give that actual promise to the existing host work observer before reporting the abort. Previously the abort listener started cancellation and discarded its promise; the parser's second cancellation was already resolved even though the first cleanup was still running.

The production repair is seven added and two removed lines. Normal non-abort cleanup remains awaited, and aborts preserve their original reason without waiting for cancellation to finish. No public API, dependency, or database policy changes.

## User Impact

Aborted Anthropic calls still return promptly, while their runtime remains alive until accepted cancellation work finishes. Cleanup failures do not replace the original abort reason.

## Evidence

Original code fails three transport regressions and the actual core ownership case: the parent drains early, cancellation runs in a foreign context, and a new ModelRegistry is built during held cleanup. The candidate passes all 148 transport and core cases. Affected checks, fresh isolated Codex review through P2, and the ordinary build (317.9 seconds) pass.

Compiled proof uses the actual exec reviewer and managed Anthropic transport against a local synthetic HTTP server. Abort reporting took 4.357 and 0.534 milliseconds for resolving and rejecting cancellation. The original owner stayed alive, overlapping preparation reused its generation, and final drainage followed the held cancellation's settlement. Non-abort cleanup and successful completion also passed.

All child processes exited; 12,119 build artifact hashes stayed unchanged and 2,872 resolved files used no application source fallback. No reviewed command was executed, and no real vendor credentials or production Gateway were used. These are individual correctness timings, not a comparative benchmark. The proof establishes ModelRegistry retention; general RUN SQLite physical disposal remains a separate prerequisite chain.

The final rebase combines the newly landed SDK cases with this cancellation regression. Both transport source files remain byte-identical to the reviewed build; all 157 transport/core composition cases and formatting checks pass on the refreshed head.

### DIFF
--- a/packages/ai/src/transports/anthropic-transport-stream.test.ts
+++ b/packages/ai/src/transports/anthropic-transport-stream.test.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from "node:async_hooks";
 import type { AssistantMessage, Model } from "@openclaw/llm-core";
 /**
  * Tests Anthropic Messages transport streaming.
@@ -6,6 +7,7 @@ import type { AssistantMessage, Model } from "@openclaw/llm-core";
  */
 import { expectDefined } from "@openclaw/normalization-core";
 import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { createDeferred } from "../../../../test/helpers/promise.js";
 import { makeUserMessage } from "../../../../test/helpers/user-message.js";
 import {
   configureAiTransportHost,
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [fd9a80b2d2b14a743e688d9f5b0497c7ba448dc0](https://github.com/openclaw/openclaw/pull/143377/commits/fd9a80b2d2b14a743e688d9f5b0497c7ba448dc0)
- Landed commit: [f30b1e9d97d40f82f8c072a868b526a49ff815dd](https://github.com/openclaw/openclaw/commit/f30b1e9d97d40f82f8c072a868b526a49ff815dd)

PR #143379, opened 2026-09-09
fix: prevent cold archive pruning from blocking the Gateway

steipete:
Related: #143332, #143285.

## What Problem This Solves

Resolves a problem where disk-budget archive pruning can pause the Gateway's main thread when its cached database connection closes between free-page drain steps, measurements, or archive file removal.

## Why This Change Was Made

Queued pruning now uses the existing guarded asynchronous acquisition owner for its cold database phases. Each free-page pass keeps its checkpoints, freelist reads, and bounded vacuum together on one admitted connection. Archive queries read current rows after validation, and the complete canonical transaction runs after reacquisition following file removal.

The existing writer sections, full integrity and foreign-key checks, initial freelist bound, 512-page steps, no-progress predicate, archive ordering, unpublished exclusions, and file-before-row recovery behavior are preserved. The synchronous archive-inspection export and the two advisory readers outside the writer queue remain unchanged.

## User Impact

Cold full-file validation at these five queued pruning boundaries no longer runs on the Gateway's main thread. Pruning continues to retain the canonical recovery row if acquisition fails after the derived file was removed. No retention policy, schema, retry behavior, Worker reuse, or new writer section is introduced.

## Evidence

- On original main, five warm controls pass and all five cold cases complete real maintenance but fail the off-parent validation assertion. The independent replay uses the exact final regression file.
- The final targeted proof passes all 13 new cases and the existing large-freelist regression. The new cases cover warm/cold free-page, canonical-presence, ordered-row, unpublished-name, and post-file-removal boundaries; FIFO exclusion; revocation during drain and after file removal; and fresh unpublished-row protection. The post-removal refusal preserves the exact canonical row and blob.
- Full-context independent P0–P2 autoreview found no actionable defects.
- The final changed-file gate passes, including core/test typechecks, lint, dependency/boundary guards, and runtime import-cycle checks.
- Linux Testbox verified the exact committed source and all three changed-file hashes, then passed all 99 tests across five suites and the full build; the workload wrapper exited successfully. [Run context](https://github.com/openclaw/openclaw/actions/runs/34401131913).
- This proof exercises real synthetic SQLite archives, filesystem removal, and the maintenance owner with real integrity checks. It does not establish production LRU frequency or Gateway latency; no CLI parser, production data, or managed Gateway was exercised.

### DIFF
--- a/docs/reference/database-schemas/storage-changes.md
+++ b/docs/reference/database-schemas/storage-changes.md
@@ -89,6 +89,14 @@ do not cause unnecessary archive deletion. Full logical deletion with resumable
 physical cleanup remains a separate design; existing deletion visibility and rollback
 semantics are unchanged.
 
+Queued archive pruning prepares cold connections through the same asynchronous
+admission owner while retaining its existing writer section. Each page-drain
+pass keeps its checkpoints, freelist reads, and bounded vacuum in one synchronous
+phase on the admitted connection. Archive-row and unpublished-name reads follow
+validation. After removing a derived archive file, pruning reacquires before the
+canonical row-deletion transaction; an acquisition failure propagates without
+deleting that recovery row.
+
 ### Preserve the data and concurrency contracts
 
 An adapter must make these contracts explicit and verify them against a real
--- a/src/config/sessions/session-history-archive-pruning.admission.test.ts
+++ b/src/config/sessions/session-history-archive-pruning.admission.test.ts
@@ -1,0 +1,348 @@
+import assert from "node:assert/strict";
+import fs from "node:fs";
+import path from "node:path";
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c52dee0012d8351b6e5751c635946ccfebc785d7](https://github.com/openclaw/openclaw/pull/143379/commits/c52dee0012d8351b6e5751c635946ccfebc785d7)
- Landed commit: [8ba562ec51654cd7031b92d33908484f4f3dd08e](https://github.com/openclaw/openclaw/commit/8ba562ec51654cd7031b92d33908484f4f3dd08e)

PR #143382, opened 2026-09-09
test(chutes): use native JSON responses

steipete:
The Chutes model and implicit-provider tests each wrapped native response construction in the same private `jsonResponse` helper. Use `Response.json` directly at their nine existing object-payload call sites and delete both wrappers and their unused response-init parameter.

The fixtures remain real responses. Status 200, JSON content type and the actual payload serialization are preserved, including nested undefined and nonfinite values. Direct 401/503 responses, pricing/default checks, token-scoped caches, authenticated request order and proxy cleanup remain unchanged. This does not claim equivalence for arbitrary top-level undefined payloads, which these tests never use.

Validation: baseline and candidate 26/26 with identical cases; all 19 normal changed checks passed; direct installed Node/Undici source inspection and independent Codex review clean. All 16 static definitions and 56 assertion sites remain. Tests net -16 lines; production delta 0.

### DIFF
--- a/extensions/chutes/implicit-provider.test.ts
+++ b/extensions/chutes/implicit-provider.test.ts
@@ -8,14 +8,6 @@ import { refreshChutesOAuthCredential } from "./oauth.js";
 
 const CHUTES_OAUTH_MARKER = resolveOAuthApiKeyMarker("chutes");
 
-function jsonResponse(payload: unknown, init: ResponseInit = {}): Response {
-  return new Response(JSON.stringify(payload), {
-    status: 200,
-    headers: { "Content-Type": "application/json" },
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c299be9cc99ff17cc523e41774eef2108add525b](https://github.com/openclaw/openclaw/pull/143382/commits/c299be9cc99ff17cc523e41774eef2108add525b)
- Landed commit: [fc1d90e0cfae1e9b22a20576b7e31d12db587e6d](https://github.com/openclaw/openclaw/commit/fc1d90e0cfae1e9b22a20576b7e31d12db587e6d)

PR #143319, opened 2026-09-09
fix(routing): show implicit default bindings in health diagnostics

qingminglong:
## What Problem This Solves

Health diagnostics omit an agent binding when its account ID is omitted or blank. With a binding to the literal `default` account and Telegram's command preference set to `work`, the health summary selects the unrelated `work` account.

## Why This Change Was Made

Diagnostic binding reads, including the exported `listBoundAccountIds` helper, now include implicit `default` accounts. This observable listing change restores the documented meaning of omitted and blank stored bindings. Command and outbound binding helpers still ignore omitted accounts, so configured command preferences and explicit-binding precedence remain unchanged. Wildcards remain excluded from concrete diagnostic accounts, and malformed account values remain rejected.

## User Impact

Operators can see default-account bindings in health diagnostics without changing the account used by commands or sends.

## Evidence

Installed CLI and matching Gateway comparison, using isolated configuration and local TLS probe stand-ins:

| Observation | Baseline | Corrected candidate |
| --- | --- | --- |
| `health` with the existing `health` diagnostic flag | Binding map empty | `telegram: main=[default]` |
| `health --json` selected Telegram account | `work` | `default` |
| Normal `message send --channel telegram` without an account option | `work` | `work` |
| Configured accounts retained | `default`, `work` | `default`, `work` |

- Baseline: `fd490cf304fea7d4fd83c3a8643078ebe2884ff7`.
- Runtime proof build: `a4cf9886887594953f54d112770f1d6e74ae9412`. The following `425980e85bca9d51332a1b6d99ed524e22be28ab` commit changes only test fixtures, so it reuses the same package and runtime proof.
- Baseline evidence shows four failing diagnostic cases. The current five-case suite protects implicit diagnostics, outbound preferences, ordering, and agent scope; all five pass.
- Four malformed account values are independently checked through the installed JavaScript SDK on both builds. Each preserves the valid neighboring work binding without inventing a default account.
- The type-suppression inventory check passes unchanged; the correction introduces no TypeScript suppression.
- Focused sibling coverage passes: 16 outbound binding cases, 72 route cases, and 40 Telegram account cases.
- Independent behavior validation covered 22 installed baseline/candidate fixtures and 27 published routing API requests per build. It confirmed explicit-account precedence, wildcard handling, agent and peer/group separation, ordering, successful explicitly owned multi-agent sends, and preserved invalid-config/owner rejections.
- Formatting, syntax lint, maximum-line and assertion checks pass.
[...]

PR #137614, opened 2026-09-03
perf(gateway): resolve heartbeat enrollment once per health and status snapshot

ylcn91:
Closes #137570

## What Problem This Solves

After `ready`, a Gateway with a large agent roster freezes its event loop for tens of seconds on every automatic health refresh: the reporter's 632-agent deployment blocked 45-57 s roughly every 6 minutes, `/health` timed out for the supervisor, and the gateway was marked down while it was in fact up. The V8 stack in the report is the health summary resolving heartbeat enrollment once per agent, and each of those resolutions walking the whole roster again.

## Why This Change Was Made

`buildHealthAgentSummaries` called `resolveHeartbeatSummaryForAgent(cfg, agentId)` for every agent. That function decides enrollment through `isHeartbeatEnabledForAgent`, which calls `resolveHeartbeatAgents(cfg)`: a full pass over `agents.entries` that resolves each candidate through `resolveAgentConfig` -> `resolveAgentEntry`. Outside a roster-facts batch `resolveAgentEntry` is itself a linear scan, so one summary cost O(agents²) config reads and O(agents²) entry clones. `getStatusSummary` in `src/status/summary.ts` has the same per-agent shape, which the ClawSweeper triage on the issue called out as the sibling that a collector-only cache would miss.

The repair lives at the heartbeat-summary owner:

- New internal module `src/infra/heartbeat-summary-projection.ts` owns enrollment (`enrolledHeartbeatAgentIds`), the per-agent projection (`buildHeartbeatSummary`), and `resolveHeartbeatSummariesForAgents(cfg, agentIds)`, which runs inside `withAgentRosterFactsBatch` (the memoization #135743 added for startup discovery), resolves the enrolled set once, and projects every agent from that set. `src/infra/heartbeat-summary.ts` keeps exactly its previous public exports (`HeartbeatSummary`, `resolveHeartbeatIntervalMs`, `isHeartbeatEnabledForAgent`, `resolveHeartbeatSummaryForAgent`) and implements the two functions through the shared projection, so single-agent callers (prompt build, run history, capability changes, spawn) are unchanged. The split matters because `src/plugin-sdk/infra-runtime.ts` re-exports `heartbeat-summary.ts` with `export *`: putting the fleet resolver there widened the pinned plugin SDK surface by one export and failed `test/scripts/plugin-sdk-surface-report.test.ts` in CI; the internal module keeps the SDK surface unchanged.
[...]

PR #134488, opened 2026-08-31
perf(sessions): bound health and status session reads

steipete:
Related: #130324. Split from #130706; the broader CPU/stall investigation remains separate.

## What Problem This Solves

Health and status loaded every stored session payload to report counts and a small recent window. Large stores therefore performed work proportional to all historical payloads even when the response needed only five or ten entries.

## Why This Change Was Made

One read-only SQLite summary owner counts canonical visible rows and hydrates only the required recent windows. Each physical store supplies one collection snapshot, including separate windows for configured agents. The existing canonical-key, readable pending-row, participant projection, and read-only database contracts are retained. Cold canonical validation remains complete.

The old full-entry listing/grouping/projection path is removed from health/status summaries. ACP metadata projection uses the already captured entry instead of rereading a possibly replaced generation. Disabled heartbeats and targets that do not consume a session route avoid that unnecessary read.

This preserves main's listing-metadata cache/data-version and pending-input receipt work. It adds no schema, index, retention, write transaction, configuration option, migration, or persistent-state design.

## Evidence

Current head: `42721f8ba7bd0c911788d88a63b4a13af8e063b5`, integrated on main `e27a72435258a05d09ff5e98b2c2ac2fe18d49f4`.

- Fresh unchanged-main reproduction failed both bounded-payload cases: old session payloads were parsed two/four times instead of zero. The identical focused command passed after the repair. These are work assertions, not timing thresholds.
- The ordered complete owner/sibling cohort passed **432 tests across 23 files**, with one existing platform-specific SQLite lock skip. Main's cache, deletion, default-agent listing, and pending-input suites are included. No retries or deadline changes.
- The complete staged changed gate passed. QA/test audit and deslop are complete. Managed Codex P0 review returned scoped-clean with no findings; no all-priority review claim. Source/index hashes remained unchanged through validation and match the committed tree.
- The fresh ordinary declaration-enabled full build passed on this exact head. All **15,523 emitted entries** stayed unchanged through both subsequent runs; source remained clean and identical. Actual emitted owner/export and protocol bindings were inspected before execution.
- Both first-attempt built Gateway runs passed: **four separate agent stores** and **one shared physical store**, each with four logical agents and **1,000 sessions per agent**, using 8 KiB payloads. First and warm status returned **4,000 total / 10 global recent**, with **1,000 / 10** per agent. First and warm health returned **1,000 / 5** per agent. Every recorded key, timestamp, and ordering matched the seeded rows plus normal label updates.
[...]

PR #143384, opened 2026-09-09
fix(gateway): avoid repeated health refreshes on reconnect

steipete:
Related: #134488, #137614

## What Problem This Solves

Fixes unnecessary Gateway work when operators repeatedly run diagnostics or clients reconnect. Each accepted connection started another background health collection, even when the previous collection had just completed. Health collection still walks session keys and prepares agent and plugin summaries, so repeated connections amplified that work.

## Why This Change Was Made

Share the existing one-minute, Gateway-owned background admission rule between post-connect refreshes and cached health requests. Keep explicit probes, missing/stale/runtime-mismatched health refreshes, lifecycle refreshes, audience separation and generation ordering unchanged. The collector, session queries, canonical validation, event-loop sampler and CPU/readiness acceptance are unchanged. This builds on current main including #143319 and preserves its implicit-default account bindings.

## User Impact

Repeated diagnostic connections and cached health reads share the permitted background refresh instead of each rebuilding the health snapshot. Operators can still request immediate live health probes.

## Evidence

- The connection-first and cached-health-first regressions both failed before the fix: three background refreshes instead of one. Both pass after the fix, including explicit/missing/stale refreshes, cadence expiry and independent Gateway ownership.
- 83 focused connection, health-cache, refresh-strength, audience and generation tests passed.
- Independent review found no actionable P0–P2 issues.
- An ordinary synthetic macOS/Node 26.8.1 baseline with 10,000 valid sessions measured median full passive refreshes of 15.5 ms on a held handle and 54.6 ms on a fresh handle, including acquisition and canonical validation. This establishes repeatable collection cost; it does not attribute a production CPU event or promise a latency improvement of that size.
- Changed-file checks passed, including core/core-test types, lint, dead-export, import-cycle and repository guards.
- [Installed Linux Gateway/CLI comparison](https://github.com/openclaw/openclaw/actions/runs/34402364298) passed on Blacksmith Testbox with Node 24.19.0. Canonical full builds and tarball-integrity checks passed for both variants; the fixed materialized tree matches `7416444663d` exactly. After observer setup, `system.info` → connect-only `gateway status` → `channels.status` produced 3 public health publications before the fix and 0 after it; a cached health request changed from 1 to 0; an explicit public probe still produced 1 for each variant. Both task-owned foreground Gateways exited cleanly. The fixture had zero sessions and no configured channels, so this proves installed admission/publication behavior without claiming channel networking or production CPU causality.
- [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34402840578) passed for `7416444663d`.

### DIFF
--- a/docs/gateway/health.md
+++ b/docs/gateway/health.md
@@ -115,6 +115,9 @@ When no `x-openclaw-session-key` header or `user` field is provided, `/v1/chat/c
 `openclaw health` asks the running gateway for its health snapshot (no direct channel
 sockets from the CLI). By default it returns a fresh cached gateway snapshot and the
 gateway refreshes that cache in the background; `--verbose` forces a live probe instead.
+Connections and cached health reads share a one-minute background refresh cadence, so
+repeated diagnostic connections do not each rebuild the health snapshot. Explicit live
+probes and refreshes for missing or stale health still run immediately.
 Snapshots describe loaded and configured channels. Stored credentials alone do not
 activate a channel or add it to Gateway health; use channel setup to enable it.
 The command reports linked creds/auth age when available, per-channel probe summaries,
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7416444663d21e32f170f8866c9adbd613c958c9](https://github.com/openclaw/openclaw/pull/143384/commits/7416444663d21e32f170f8866c9adbd613c958c9)
- Landed commit: [ce51d6fc1c2a3d5730efed0730505761d334ea39](https://github.com/openclaw/openclaw/commit/ce51d6fc1c2a3d5730efed0730505761d334ea39)

PR #143391, opened 2026-09-09
fix(agents): resolve the CLI execution auth identity for native compaction

### LINKED ISSUE #143390: [Bug]: manual /compact of a claude-cli session bills the stored model-provider API key instead of the CLI's native login
### Bug type
Behavior bug (incorrect credential use without crash) — agents / compaction

### Beta release blocker
No

### OpenClaw version
Root cause present on `main` @ `9753ab2522e` (compaction dispatch predates the reply-boundary conversion).

### Operating system
Not OS-dependent (gateway-side compaction dispatch).

### Install method
Not install-dependent — prompt/compaction dispatch path.

### Model / provider chain
Any CLI backend that owns native compaction (`claude-cli`, `google-gemini-cli`) with a session pin that names a stored model-provider credential.

## Summary

Manual compaction of a CLI-backed session forwards the raw session pin into the compaction child. `compactNativeCliSession()` (`src/agents/embedded-agent-runner/compact.ts`, the `runCliAgent` call) falls back to `compactParams.authProfileId` — the session pin — without converting it to the CLI execution identity. The pin names the **model provider's** credential (for example the auto-selected `anthropic:default` API key), so the claude CLI child runs its compaction turn under the stored OpenClaw API key instead of the backend's own login.

The reply dispatch and the gateway command path already apply the conversion:

- `src/auto-reply/reply/agent-runner-cli-candidate.ts` composes the pin through `resolveCliForwardedAuthProfileId` → `resolveCliExecutionAuthProfileId`.
- `src/agents/command/attempt-execution.ts` applies `resolveCliExecutionAuthProfileId` before preparing CLI runs.
- `src/agents/cli-execution-auth.ts` documents the contract: automatic selection must not replace the CLI's native identity; a user-locked profile must fail closed rather than run as another user.

The native-compaction dispatch boundary is the remaining CLI child entry point that skips it.

## What triggers it

1. A session pinned to a CLI backend that owns native compaction (`claude-cli`, `google-gemini-cli`).
2. The CLI session binding records no credential pin — the normal state for a native-login session, because `prepareCliRunContext` clears native profiles and the binding records only what a run actually consumed.
3. The session pin (`authProfileOverride`) names a stored model-provider credential, e.g. an auto-selected `anthropic:default` API key — exactly the pin state the reply-boundary conversion exists to neutralize.
4. The operator runs `/compact` (manual trigger).

### The unconverted fallback

`src/agents/embedded-agent-runner/compact.ts`, `compactNativeCliSession()`:

```ts
...(cliSessionBinding?.authProfileId
  ? { authProfileId: cliSessionBinding.authProfileId }
  : params.compactParams.authProfileId
    ? { authProfileId: params.compactParams.authProfileId }
    : {}),
```

The session pin is forwarded raw into `runCliAgent`. `prepareCliRunContext` (`src/agents/cli-runner/prepare.ts`) then resolves that profile's stored API key and forwards it to the child: the pin is not one of the backend's `nativeAuthProfileIds`, so `usesNativeAuthProfile` does not clear it, and the credential is resolved for execution.

### The design intent it contradicts

The same call site runs before generic model preparation with this comment: "Native control operations reuse the backend's existing authenticated session. Run them before generic model preparation so subscription-only CLI sessions do not incorrectly require an OpenClaw model API credential." Forwarding the session's model-provider pin reproduces exactly that incorrect requirement, on the compaction turn.

### Dispatch-boundary comparison

| CLI child boundary | Converts the session pin? |
|---|---|
| Reply dispatch (`agent-runner-cli-candidate.ts`) | Yes — `resolveCliForwardedAuthProfileId` |
| Gateway command runs (`attempt-execution.ts`) | Yes — `resolveCliExecutionAuthProfileId` |
| System inference route (`inference-route.ts`) | Yes — `resolveCliExecutionAuthProfileId` |
| Native compaction dispatch (`compactNativeCliSession`) | **No** |

## Why it happens

The reply-boundary conversion fixed the credential mismatch where the reply path forwarded the session pin to the CLI child. Manual compaction reaches the same child through a separate dispatch boundary that predates that conversion, so the same session state that the reply path neutralizes silently re-appears on the compaction turn.

## Impact

A subscription-only claude-cli session (no OpenClaw-side Anthropic credential intended for use) whose session pin auto-selected a stored API key compacts under that API key: the compaction turn bills the stored key, breaks credential-boundary consistency with the surrounding turns, and can invalidate the session's resume contract across the credential boundary. Nothing is logged; the operator sees a successful compaction.

## Suggested fix

Compose the session-pin fallback through `resolveCliExecutionAuthProfileId` before dispatch, mirroring the reply boundary:

- keep the binding pin as-is (it records the credential the session's CLI runs last consumed — the identity that produced the native session);
- an automatic pin naming a foreign credential drops, so the CLI child keeps its native login;
- an explicitly selected compatible credential still forwards;
- an explicit selection with no stored credential fails closed (`CliExecutionAuthProfileError`), per the resolver's contract.

Regression coverage can observe the credential that reaches `runCliAgent` through `compactNativeCliSession`, so restoring the raw session pin at the dispatch site fails the test.

ruel225:
Closes #143390

## What Problem This Solves

Manual compaction of a CLI-backed session forwarded the raw session pin into `runCliAgent` (`compactNativeCliSession`, `src/agents/embedded-agent-runner/compact.ts`). The pin names the model provider's credential — for example the auto-selected `anthropic:default` API key — so a claude CLI child ran its compaction turn under the stored OpenClaw API key instead of the backend's own login.

The reply dispatch (`agent-runner-cli-candidate.ts`), the gateway command path (`attempt-execution.ts`), and the system inference route (`inference-route.ts`) all convert the session pin through `resolveCliExecutionAuthProfileId` before reaching a CLI child; the native-compaction dispatch boundary was the remaining entry point that skipped the conversion. The defect contradicts the boundary's own stated intent: native control operations "reuse the backend's existing authenticated session" so subscription-only CLI sessions do not require an OpenClaw model API credential.

The review added a second, deeper boundary defect at the same call: with a native-login binding, an automatic `anthropic:default` pin, and an eligible stored `claude-cli:work` credential, the composed resolver call fell through to automatic fallback discovery and returned `claude-cli:work` — introducing a *different stored account* into compaction. Control operations reuse the recorded native session without the auth-profile/epoch compatibility re-check, so the discovered credential would receive a session produced under another identity.

## Solution

Compose the session-pin fallback through `resolveCliExecutionAuthProfileId` before dispatch, mirroring the reply boundary, and decline discovery for this caller:

- the binding pin still wins unchanged — it records the credential the session's CLI runs last consumed, so compaction resumes the native session under the identity that produced it;
- an automatic pin naming a foreign (model-provider) credential drops, and the resolver now accepts an explicit `discoverFallbackAccount: false` so automatic selection cannot discover a replacement account for a session it does not revalidate — the CLI child keeps its native login;
- an explicitly selected compatible credential still forwards (including the canonical-provider explicit case);
- an explicit selection with no stored credential fails closed via `CliExecutionAuthProfileError` — a user-locked profile must not run the compaction as another user.

Production delta is one composed call reusing the existing resolver at the dispatch boundary plus one documented opt-out parameter on that resolver; no new policy is invented. The existing dispatch (`attempt-execution.ts`) and inference-route callers keep the default discovery behavior unchanged.

## Evidence

**Competing-profile regression, pre-fix red.** With a native-login binding (no binding pin), an automatic `anthropic:default` session pin, and an eligible stored `claude-cli:work` credential:

- *mocked-store dispatch level* (`compact.native-cli.test.ts`, "keeps the native identity when dropping an auto pin despite an eligible CLI credential"): post-fix the dispatch forwards no `authProfileId`. On pre-fix code the same run forwarded `claude-cli:work` — the review's P1 scenario, red for the intended reason.
- *real-store dispatch level* (`compact.native-cli.cli-child-auth.test.ts`, "keeps the native login when a competing stored CLI account is eligible"): the auth store is a real on-disk SQLite store seeded through `updateAuthProfileStoreWithLock` with both `anthropic:default` (api_key) and the competing `claude-cli:work` (oauth); the resolver inside `compactNativeCliSession` reads that real store. Post-fix the run reaches `runCliAgent` with no forwarded `authProfileId`; on pre-fix code the same seeded store produced the forwarded `claude-cli:work` (`expected 'claude-cli:work' to be undefined`).

**Fail-closed before child I/O.** A `user`-sourced pin naming a missing profile rejects with `No credentials found for profile "anthropic:missing"` and never dispatches (`runCliAgent` not called) — observed at the dispatch seam in both suites, so the rejection happens before any child launch.

**Existing conversion behaviors stay green.** `compact.native-cli.test.ts`: 7 passed (binding-pin precedence, auto model-provider pin drop, explicit forwarding, explicit fail-closed). `cli-execution-auth.test.ts`: 16 passed — the resolver's default discovery behavior is unchanged for the existing callers. Sibling compaction suites from the earlier head: `compact.hooks.test.ts`, `compact.delegate.test.ts`, `cli-compaction.test.ts`: 32 passed.

**Child-level spawn proof boundary, stated honestly.** The desired child-level observation (a real spawned CLI child recording its credential env during a compaction turn) was attempted and does not run in this sandbox: driving the real `runCliAgent` with `controlOperation: "compact"` fails at process spawn (`spawn /bin/sh ENOENT` — also with the OOM wrapper disabled, `spawn <node> ENOENT` for the very interpreter running the suite), while the identical runner path through the reply dispatch (`executeAgentTurn` → `runCliAgent`, head 4af92255f6e) spawns the stub child successfully in the same environment. The spawn failure is environmental, not asserted by these tests; what the real-store suite proves is the credential *selection* — the value forwarded into the runner — under a real on-disk competing store, and that an unavailable explicit selection never reaches a launch.

**Types, format, lint.** `tsgo` core sources and the full core test-shard matrix (fresh cache names): green. `oxfmt` clean, `oxlint` clean on the changed files.

## Scope

- `src/agents/embedded-agent-runner/compact.ts` — compose the session pin through the existing resolver before the `runCliAgent` dispatch, with discovery declined for this control-operation caller.
- `src/agents/cli-execution-auth.ts` — new documented `discoverFallbackAccount` opt-out; callers that revalidate the resulting identity against the session keep the default.
- `src/agents/embedded-agent-runner/compact.native-cli.test.ts` — dispatch-level regression including the competing-profile case.
- `src/agents/embedded-agent-runner/compact.native-cli.cli-child-auth.test.ts` — real-store credential-selection regression at the same boundary.

The queued (budget) compaction path is unaffected: auto-reply maintenance skips CLI and native-compaction runtimes entirely, and the harness-owned native compaction bridge is Codex-only. The embedded context-engine fallback runs on OpenClaw's own runtime, where a stored model-provider API key is a legitimate credential.

### DIFF
--- a/src/agents/cli-execution-auth.ts
+++ b/src/agents/cli-execution-auth.ts
@@ -41,6 +41,14 @@ export function resolveCliExecutionAuthProfileId(params: {
   config: OpenClawConfig;
   agentDir: string;
   selected?: CliExecutionAuthProfileSelection;
+  /**
+   * Callers that reuse an existing native session must not let automatic
+   * selection introduce a different stored account: the session keeps its
+   * original identity without a compatibility re-check, so discovery would
+   * hand that session to another credential. Such callers pass `false` and
+   * fall back to the CLI child's own login instead.
+   */
+  discoverFallbackAccount?: boolean;
   loadAuthProfileStoreForRuntime?: typeof loadAuthProfileStoreForRuntime;
 }): string | undefined {
   const loadStore = params.loadAuthProfileStoreForRuntime ?? loadAuthProfileStoreForRuntime;
@@ -89,6 +97,13 @@ export function resolveCliExecutionAuthProfileId(params: {
         `CLI backend "${params.cliExecutionProvider}" cannot use auth profile "${selectedAuthProfileId}" owned by "${credential.provider}".`,
       );
     }
+  }
+
+  // Fallback discovery may only introduce a stored account when the caller
+  // revalidates the resulting identity against the session it resumes.
+  // Callers that decline discovery keep the CLI child's own login.
+  if (params.discoverFallbackAccount === false) {
+    return undefined;
   }
 
   const cliProfileId = resolveAuthProfileOrder({
--- a/src/agents/embedded-agent-runner/compact.native-cli.cli-child-auth.test.ts
+++ b/src/agents/embedded-agent-runner/compact.native-cli.cli-child-auth.test.ts
@@ -10,6 +10,7 @@ import { afterEach, beforeEach, describe, expect, it, onTestFinished, vi } from 
 import { useAutoCleanupTempDirTracker } from "../../../test/helpers/temp-dir.js";
 import { closeOpenClawAgentDatabasesForTest } from "../../state/openclaw-agent-db.js";
 import { closeOpenClawStateDatabaseForTest } from "../../state/openclaw-state-db.js";
+import type { PreparedAgentRunAdmission } from "../admitted-run-context.js";
 import { resolveAgentDir } from "../agent-scope-config.js";
 import { clearAuthProfileMigrationDiagnostics } from "../auth-profiles/legacy-source-diagnostic.js";
 import { updateAuthProfileStoreWithLock } from "../auth-profiles/store-runtime.js";
@@ -17,7 +18,7 @@ import { testing as cliBackendsTesting } from "../cli-backends.test-support.js";
 import { compactNativeCliSession } from "./compact.js";
 
 const { runCliAgentMock } = vi.hoisted(() => ({
-  runCliAgentMock: vi.fn(async () => ({
+  runCliAgentMock: vi.fn(async (_params: { preparedRunAdmission?: PreparedAgentRunAdmission }) => ({
     meta: {
       durationMs: 1,
       agentMeta: { sessionId: "native-session", provider: "claude-cli", model: "opus" },
@@ -58,30 +59,31 @@ describe("native CLI manual compaction: CLI child credential handoff", () => {
   }
 
   function registerScriptedClaudeCliBackend() {
-    const backend = {
-      id: "claude-cli",
-      modelProvider: "anthropic",
-      pluginId: "anthropic",
-      ownsNativeCompaction: true,
-      bundleMcp: false,
-      config: {
-        command: "claude",
-        args: ["-p"],
-        input: "stdin" as const,
-        output: "jsonl" as const,
-        sessionMode: "existing" as const,
-      },
-      manualCompaction: {
-        buildPrompt: (instructions?: string) =>
-          instructions ? `/compact ${instructions}` : "/compact",
-        input: "arg" as const,
-        validateOutput: () => ({ ok: true }),
-      },
-    };
     cliBackendsTesting.setDepsForTest({
-      resolvePluginSetupCliBackend: ({ backend: id }) =>
-        id === backend.id ? { pluginId: backend.pluginId, backend } : undefined,
-      resolveRuntimeCliBackends: () => [backend],
+      resolveRuntimeCliBackends: () =>
+        [
+          {
+            id: "claude-cli",
+            modelProvider: "anthropic",
+            pluginId: "anthropic",
+            ownsNativeCompaction: true,
+            bundleMcp: false,
+            config: {
+              command: "claude",
+              args: ["-p"],
+              input: "stdin",
+              output: "jsonl",
+              sessionMode: "existing",
+            },
+            manualCompaction: {
+              buildPrompt: (instructions?: string) =>
+                instructions ? `/compact ${instructions}` : "/compact",
+              input: "arg",
+              validateOutput: () => ({ ok: true }),
+            },
+          },
+        ] as never,
+      resolvePluginSetupCliBackend: () => undefined,
     });
   }
 
--- a/src/agents/embedded-agent-runner/compact.native-cli.cli-child-auth.test.ts
+++ b/src/agents/embedded-agent-runner/compact.native-cli.cli-child-auth.test.ts
@@ -1,0 +1,160 @@
+// Proves the native compaction credential handoff at the real store boundary:
+// the compaction resolver reads a real on-disk auth-profile store seeded with a
+// model-provider key and a competing CLI-owned account. Dropping an automatic
+// model-provider pin must keep the native login (no discovered replacement
+// account reaches the CLI run), and an unavailable explicit selection must be
+// rejected before any child runs.
+import { mkdirSync } from "node:fs";
+import nodePath from "node:path";
+import { afterEach, beforeEach, describe, expect, it, onTestFinished, vi } from "vitest";
+import { useAutoCleanupTempDirTracker } from "../../../test/helpers/temp-dir.js";
+import { closeOpenClawAgentDatabasesForTest } from "../../state/openclaw-agent-db.js";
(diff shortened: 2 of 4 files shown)


ruel225:
@clawsweeper re-review

The P1 finding is repaired at head `5b87232b511e5430a7bb97e23087634053c929845` (one commit on top of the reviewed `c2521d0b8c7393c5abb0acf433454a9f16189a4a`):

- The resolver gained a documented `discoverFallbackAccount` opt-out, and native compaction passes `false`: automatic selection can no longer discover a replacement stored account for a control-operation session it never revalidates. Dropping the automatic model-provider pin now leaves the compaction under the CLI child's own login. The dispatch (`attempt-execution.ts`) and inference-route callers keep the default discovery behavior — their suites stay green.
- Competing-profile regression added at two levels, both pre-fix red with `expected 'claude-cli:work' to be undefined` and post-fix green: the mocked-store dispatch level (`compact.native-cli.test.ts`) and a real on-disk SQLite store seeded through `updateAuthProfileStoreWithLock` with both `anthropic:default` (api_key) and the competing `claude-cli:work` (oauth) (`compact.native-cli.cli-child-auth.test.ts`), where the resolver inside `compactNativeCliSession` reads the real store.
- Fail-closed before child I/O is observed at the dispatch seam in both suites: an unavailable explicit selection rejects and `runCliAgent` is never called.

One proof boundary is stated honestly in the PR body: a child-level spawn observation for the compaction turn does not run in this sandbox (the real `runCliAgent` path with `controlOperation: "compact"` fails at process spawn with `spawn ENOENT` in this environment, while the identical runner path through the reply dispatch spawns the stub child successfully — same environment, head 4af92255f6e). The real-store suite therefore proves credential selection — the identity forwarded into the runner — under a real competing store, plus the pre-launch rejection.

ruel225:
Follow-up at head `6b2071eb59b58b8b22e0a3f83a4bbbd06c7a1c993`: CI's test-types stripe rejected the new store-boundary suite's untyped mock and inline backend literal. The fix types the runCliAgent mock with the prepared-admission parameter shape and registers the backend through the existing test seam — no behavior, assertion, or evidence change; both competing-profile regressions and the fail-closed case are unchanged from `5b87232b511e` and stay green locally. Typecheck stripes 1/5 and 2/5 (the failed `check-test-types-core-1` workload) now pass.

ruel225:
Closing to stay under the PR budget after the ClawSweeper re-estimation (5 items, 2/6). The remaining gates are final-process evidence through the real runner that a competing account cannot receive the resumed session, plus the listed security owner's involvement — both need real native-compaction credentials and a maintainer security review we cannot self-serve. The dispatch-path sibling #142842 remains open at a stronger rating and covers the auth-identity family. Branch preserved on the fork. Revival: security-owner involvement or maintainer invitation. Issue #143390 stays open.

PR #143398, opened 2026-09-09
fix(tests): close Control UI fixture databases before cleanup

steipete:
Related: #139428

## What Problem This Solves

Fixes a Control UI test-fixture lifetime bug: rejected and rate-limited token reads open a cached SQLite database, but their temporary HOME directories were removed without closing it. A full-suite diagnostic reported a WAL-sidecar identity failure under the invalid-token fixture after that HTTP case had passed.

## Why This Change Was Made

Reuse the existing HOME lifecycle helper for both failure cases and give it the general name `withControlUiHome`. It captures the selected database path inside the HOME scope, closes that exact owner before deleting the directory, and preserves the files if closing throws. Existing paired-device callers only change helper name.

## User Impact

Test cleanup now retires the fixture-owned database before unlinking its WAL files. Production authentication, rate limiting and HTTP assertions are unchanged.

## Evidence

- All 159 Control UI HTTP cases pass, with the same ordered case identities.
- Original observation of the two failure cases found one open database and WAL immediately before tracker cleanup. That diagnostic safely closed each exact owner before allowing deletion; it did not induce a guard failure or run an unmodified full-suite baseline.
- Candidate observation found both databases already closed and WAL files absent before unlink, using only the existing helper's close call. Both fixture roots were removed.
- The helper body is unchanged apart from its name, preserving the prior success, callback-error and close-error checks; a close failure still prevents directory removal.
- The required changed gate and fresh P2 review passed. No deadline, global tracker or production changes.

### DIFF
--- a/src/gateway/control-ui.http.test.ts
+++ b/src/gateway/control-ui.http.test.ts
@@ -476,7 +476,7 @@ describe("handleControlUiHttpRequest", () => {
     }
   }
 
-  async function withPairedDeviceHome<T>(prefix: string, fn: () => Promise<T>): Promise<T> {
+  async function withControlUiHome<T>(prefix: string, fn: () => Promise<T>): Promise<T> {
     const tempHome = await fs.mkdtemp(path.join(os.tmpdir(), prefix));
     let databasePath: string | undefined;
     try {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [44faabb4a3c642adf7809d82b7bfeb95eb2d2b53](https://github.com/openclaw/openclaw/pull/143398/commits/44faabb4a3c642adf7809d82b7bfeb95eb2d2b53)
- Landed commit: [30ce57e5e3f7533b2531d0cb3fd0a25f96c2b93a](https://github.com/openclaw/openclaw/commit/30ce57e5e3f7533b2531d0cb3fd0a25f96c2b93a)

PR #143400, opened 2026-09-09
fix(chat): clear previews from stopped typing connections

steipete:
## What Problem This Solves

Fixes an issue where another viewer could see an outdated draft after a user cleared, sent or blurred their composer in one tab while another tab remained typing. The stopped tab's preview could remain visible or even be broadcast afterward.

## Why This Change Was Made

The connection tracker already removes the stopped tab and selects the remaining draft. The handler now passes that aggregate state through the existing throttle instead of returning early whenever another tab is still active. This removes four production lines without adding state or timers.

## User Impact

Typing presence now reflects the user's remaining active tabs at the existing broadcast cadence. Session replacement, actor attribution and recipient scope remain unchanged.

## Evidence

Three real-handler regressions fail before the fix and pass afterward, alongside 18 controls. Six UI/protocol cases also pass, for 27 distinct tests. All 29 changed-file checks and independent review through P2 pass. Tests use isolated synthetic sessions; no live Gateway or browser session was used.

### DIFF
--- a/src/gateway/server-methods/sessions-suggestions.ts
+++ b/src/gateway/server-methods/sessions-suggestions.ts
@@ -582,10 +582,6 @@ export const sessionSuggestionHandlers: GatewayRequestHandlers = {
       ...(params.typing && params.preview ? { preview: params.preview } : {}),
       now,
     });
-    if (!params.typing && effectiveTyping) {
-      respond(true, { ok: true, broadcast: false });
-      return;
-    }
     const broadcast = broadcastTypingThrottled({
       key: typingKey,
       typing: effectiveTyping,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7bb16cf85cee011eec63ec1a1ebaa315e553e166](https://github.com/openclaw/openclaw/pull/143400/commits/7bb16cf85cee011eec63ec1a1ebaa315e553e166)
- Landed commit: [4617cfcacac2d24455360bc62e1decb6c4cf712e](https://github.com/openclaw/openclaw/commit/4617cfcacac2d24455360bc62e1decb6c4cf712e)

PR #143405, opened 2026-09-09
refactor(ui): keep one Cron editor job snapshot

steipete:
## What Problem This Solves

The Automations editor kept separate ID and revision fields that duplicated its retained job. Every editor transition had to maintain all three values together.

## Why This Change Was Made

Use the retained job as the sole editor owner and read its revision through the existing save guard. Remove redundant test assertions where whole-job or awaited-state checks already cover the same values.

## User Impact

Behavior is unchanged. An open editor keeps its own job snapshot across filtered list refreshes, preserves its revision after failed conflict recovery, and advances it after accepted saves or toggles. The change removes eight production lines and sixteen test lines.

## Evidence

All 297 existing cases across six Cron suites and all 20 changed-file checks pass. The final edit only removes an already-awaited duplicate assertion; the same 31 page cases then pass again, with formatting and whitespace checks clean. Independent review through P2 found no issues. No visual or live Gateway change is claimed.

### DIFF
--- a/ui/src/lib/cron/index.test.ts
+++ b/ui/src/lib/cron/index.test.ts
@@ -696,9 +696,9 @@ describe("cron controller", () => {
       delivery: { mode: "none" },
     });
     expect(requestPatch(call)).not.toHaveProperty("deleteAfterRun");
-    expect(state.cronEditingJobId).toBe("job-1");
+    expect(state.cronEditingJob?.id).toBe("job-1");
     expect(state.cronEditingJob?.name).toBe("Existing job");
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c0d764d81d0c8dcd54164e0de1e4282fbfb28347](https://github.com/openclaw/openclaw/pull/143405/commits/c0d764d81d0c8dcd54164e0de1e4282fbfb28347)
- Landed commit: [fda2de606028f552d3bc4326fac08f56601207a8](https://github.com/openclaw/openclaw/commit/fda2de606028f552d3bc4326fac08f56601207a8)

PR #143412, opened 2026-09-09
refactor(signal): exercise real inbound orchestration in tests

steipete:
Related: #139428

## What Problem This Solves

Four Signal test suites replayed inbound orchestration in their mocks, so they could miss behavior owned by the real runner, including finalization when dispatch fails.

## Why This Change Was Made

Use the real inbound runner through one local test adapter that preserves each suite's existing session-recording and dispatch sinks. The adapter retains the captured adoption lifecycle; Signal's outer flush continues to settle skipped claims. This removes 75 net test/support lines without changing production code.

## User Impact

No user-visible behavior change. Signal tests now exercise the actual orchestration path while retaining their control-lane, retry, mention and ingress-ownership assertions.

## Evidence

All 65 existing cases passed before and after, with identical ordered case results and unchanged assertions/hooks outside the four mock definitions. Another 50 core orchestration cases and five Signal partial-delivery cases passed.

Temporary sensitivity controls verified that the real runner finalizes a failed dispatch once and preserves its original error even if finalization also rejects. The old replay failed that same control because it never finalized. A separate control verified skipped dispatch, exact mapped lifecycle binding and settlement by the real outer owner. Temporary controls were removed after verification.

Changed checks, including extension typechecks and lint, passed. Independent review found no actionable P2-or-higher issues. No runtime improvement is claimed.

### DIFF
--- a/extensions/signal/src/monitor/event-handler.control-lane.test.ts
+++ b/extensions/signal/src/monitor/event-handler.control-lane.test.ts
@@ -41,44 +41,14 @@ vi.mock("openclaw/plugin-sdk/channel-inbound", async () => {
   const actual = await vi.importActual<typeof import("openclaw/plugin-sdk/channel-inbound")>(
     "openclaw/plugin-sdk/channel-inbound",
   );
-  type RunParams = Parameters<typeof actual.runChannelInboundEvent>[0];
+  const { createSignalPreparedDispatchRunner } = await import("./event-handler.test-harness.js");
   return {
     ...actual,
-    runChannelInboundEvent: async (params: RunParams) => {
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d257ee6211b64c455f577425c40104426c3b7e0c](https://github.com/openclaw/openclaw/pull/143412/commits/d257ee6211b64c455f577425c40104426c3b7e0c)
- Landed commit: [7818177bf0bedc8a35e3cf666824c20e643dad66](https://github.com/openclaw/openclaw/commit/7818177bf0bedc8a35e3cf666824c20e643dad66)

PR #143414, opened 2026-09-09
fix: retain isolated completion state through cleanup

steipete:
Related: #140674.

## What Problem This Solves

Resolves a problem where an isolated model completion can release its prepared runtime after reporting a result or cancellation while accepted auth or transport cleanup is still running.

## Why This Change Was Made

Start work ownership before runtime acquisition and setup, retain the selected lease through actual cooperating work, and release it after drainage. Parent cancellation and normal drainage restore the captured runtime context. Normal completion lets accepted auth work finish before closing its signal.

A small ownership helper keeps this lifecycle separate from the existing policy code. Input snapshots, authority checks, model/profile/harness/CLI selection, and result/error behavior remain unchanged. No public API, configuration, dependency or producer-activation change.

## User Impact

Isolated completions keep their current result and cancellation behavior while retaining the original runtime through cleanup. This also prevents overlapping calls from rebuilding a generation that still has an active user.

## Evidence

Original production fails seven ownership/context cases while 31 controls pass. An intermediate implementation fails four cases, showing why normal completion must not abort accepted auth work and why cancellation must restore the admitted context. The final candidate passes all 89 owner and caller tests across six files.

The complete affected checks, fresh full-candidate Codex review through P2, and ordinary build pass. The new helper preserves the original policy body and keeps the existing file limit without a baseline exception. The build took 249.73 seconds.

Compiled plain-Node proof uses the actual isolated-completion entry, a synthetic provider and real loopback HTTP. Standalone auth-tail, callback-abort and foreign-context parent-cancellation scenarios pass in 2.17 seconds overall. Each makes three HTTP requests and observes two actual base generations: overlapping work reuses the retained generation, and a later call rebuilds after cleanup. All 12,119 artifact hashes and five source hashes remain unchanged, with no application source fallback; all processes exit and synthetic state is removed.

One probe initially assumed the second call's physical release coincided with its logical result. Joining that control call's actual cleanup fixed the assertion; the tested first calls remain standalone, and no product change or delay was added. Timings are correctness observations, not a speedup benchmark. General RUN SQLite disposal and real vendor execution are separate from this proof.

The final rebase includes the landed Anthropic cancellation repair and preserves all five validated code/test files byte-for-byte.

### DIFF
--- a/src/agents/harness/isolated-completion.native-auth.test.ts
+++ b/src/agents/harness/isolated-completion.native-auth.test.ts
@@ -15,6 +15,8 @@ import {
 const { createPluginMetadataSnapshot, makeRegistry } =
   await import("../../config/plugin-auto-enable.test-helpers.js");
 
+const { AsyncWorkScope } = await import("../../shared/async-work-scope.js");
+
 beforeEach(resetIsolatedCompletionTestState);
 
 describe("runIsolatedCompletion native authorization", () => {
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [25abe41cd2947153f1d4e01e91720c81e95a678f](https://github.com/openclaw/openclaw/pull/143414/commits/25abe41cd2947153f1d4e01e91720c81e95a678f)
- Landed commit: [065dc15433d3561ead0a13d643de990d72aaa769](https://github.com/openclaw/openclaw/commit/065dc15433d3561ead0a13d643de990d72aaa769)

PR #143415, opened 2026-09-09
test: avoid cold source SDK loading in MCP coverage

steipete:
## What Problem This Solves

The MCP registration-ownership test spends roughly 52 seconds loading the source SDK through Jiti before it exercises its native resources. The complete seven-case command took 65–72 seconds locally; main CI reported 50 seconds in its first case alone.

## Why This Change Was Made

Reuse the test runner's existing verified build of current source for the real provider-catalog SDK entry. Extract the existing compiled-SDK host fixture so MCP and the published-SDK loader test share its setup. Each MCP case still imports the real SDK during registration and exercises its original transport, cancellation, SQLite lifetime, and disposal assertions. Additional loader counters require native loading without source-transform fallback.

The host is tracked and removed after the file; loader state and the temporary environment setting are restored between cases. Standalone/watch execution retains the existing source path. This changes test preparation only, with no production loader, API, dependency, configuration option, timeout, or shard change.

## User Impact

Faster native MCP coverage without replacing the SDK or lifecycle behavior with mocks. The private compiler gains only one input and one output. Peak RSS remains approximately 2.28 GB; no memory improvement is claimed. Production LOC is unchanged; tests/support are +68/-18 and the build-entry list is +5/-1.

## Evidence

| Source-only checkout | Complete seven-case command | Test result |
| --- | ---: | --- |
| Original | 64.89s | 7 passed |
| Compiled SDK fixture | 14.21s | 7 passed |
| Restored original | 71.95s | 7 passed |

These timings include compilation, host creation/copy, verification, and cleanup, with identical physical dependencies and Node 24.19.0/pnpm 12.3.4. The candidate's first case took 556 ms and its cancellation case 12 ms. Compiler samples were 4.449/3.493/3.649 seconds; their variance is not a separate compiler-speed claim.

- Root validation after the final method-binding correction: all 10 MCP and published native-loader cases passed in 23.79 seconds.
- Standalone source-declaration branch: the existing published-SDK bridge case passed; two other cases were deliberately excluded by its focused filter.
- CPU profiling attributed the original cost to Jiti resolution/stack work and Babel in the native SDK graph. A simpler conditional import was tested and rejected because it merely moved 52 seconds to another case.
- All changed checks passed, including core/script typechecking, lint, and build-entry guards. Managed independent P2 review is clean. The broader shared build graph runs in the PR's required CI. Existing path filters did not select Windows; no Windows result is claimed for this patch.

### DIFF
--- a/scripts/lib/vitest-worker-build-entries.mts
+++ b/scripts/lib/vitest-worker-build-entries.mts
@@ -20,7 +20,10 @@ import {
   triageMaintenanceRuntimeEntrypoints,
 } from "../../src/infra/triage-runtime.test-support.ts";
 import { nodeHostConfigRuntimeEntrypoint } from "../../src/node-host/config-runtime.test-support.ts";
-import { publishedSdkBridgeEntrypoints } from "../../src/plugins/loader-sdk-bridge-artifacts.test-support.ts";
+import {
+  mcpProviderCatalogEntrypoint,
+  publishedSdkBridgeEntrypoints,
+} from "../../src/plugins/loader-sdk-bridge-artifacts.test-support.ts";
 import { persistenceRuntimeEntrypoint } from "../../src/skills/library/persistence-runtime.test-support.ts";
 import {
   agentDatabaseHeldRuntimeEntrypoint,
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ff8eca3110883e9b2a8d0f5a89e5a330caadbc65](https://github.com/openclaw/openclaw/pull/143415/commits/ff8eca3110883e9b2a8d0f5a89e5a330caadbc65)
- Landed commit: [7e3a90bf5ce8959843a0758a9151fdef5d1e756e](https://github.com/openclaw/openclaw/commit/7e3a90bf5ce8959843a0758a9151fdef5d1e756e)

PR #143416, opened 2026-09-09
fix(channels): preserve labels for unloaded plugins

steipete:
## What Problem This Solves

Fixes an issue where configured plugins could lose their declared channel subtitle before their runtime loaded. The Channels page could show “Open details” instead of the plugin's description. Static catalog output also dropped selection labels and icon metadata.

## Why This Change Was Made

Project the three existing presentation fields from normalized package metadata into the read-only adapter, only for the package's declared channel. Keep primary label/description overrides and loaded-plugin precedence intact. The broader metadata builder is not used because it also projects policy fields.

## User Impact

Configured channels retain their declared presentation without loading plugin runtime or setup code. The change adds eleven production lines and changes no channel policy or configuration contract. Icon metadata is restored in the catalog; no current Control UI icon-rendering defect is claimed.

## Evidence

Four real adapter/catalog regressions fail before the fix and pass afterward, with all 67 tests and 33 changed-file checks passing. Independent review through P2 is clean.

The synthetic screenshots replay actual recorded before/after catalog output through the unchanged Channels view: “Open details” becomes “Package Chat Bot.” The corrected capture has no page errors; its first attempt stopped on an incomplete mock pairing response, fixed only in the capture script. No live Gateway or real operator data was used.

Before: synthetic catalog output loses the plugin detail label

After: synthetic catalog output preserves Package Chat Bot

### DIFF
--- a/src/channels/plugins/manifest-channel-plugin.types.ts
+++ b/src/channels/plugins/manifest-channel-plugin.types.ts
@@ -14,6 +14,8 @@ export type ManifestChannelPlugin = {
     id: string;
     label: string;
     selectionLabel: string;
+    detailLabel?: string;
+    systemImage?: string;
     docsPath: string;
     blurb: string;
     preferOver?: readonly string[];
--- a/src/channels/plugins/read-only.test.ts
+++ b/src/channels/plugins/read-only.test.ts
@@ -14,6 +14,7 @@ import {
   resetPluginLoaderTestStateForTest,
   useNoBundledPlugins,
 } from "../../plugins/loader.test-fixtures.js";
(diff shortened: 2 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2ab2943fc8654472ef70a948e82543fb8b795400](https://github.com/openclaw/openclaw/pull/143416/commits/2ab2943fc8654472ef70a948e82543fb8b795400)
- Landed commit: [3415f22f1022eda6d94f9d5bb9291ea0bebc4930](https://github.com/openclaw/openclaw/commit/3415f22f1022eda6d94f9d5bb9291ea0bebc4930)

PR #143426, opened 2026-09-09
fix: retain background compaction state through disposal

steipete:
Related: #140674.

## What Problem This Solves

Resolves a problem where accepted background compaction can release its prepared runtime before maintenance, reruns or engine disposal finish. Cancellation during disposal can also lose its signal handler or wait on the operation's own write owner.

## Why This Change Was Made

Transfer the existing prepared lease to the accepted maintenance completion promise. Include worker and disposer descendants, superseded queued-engine disposal and coalesced reruns in that completion. Capture source and maintenance-writer context at the operation boundary, and keep cancellation attached until final disposal.

Accepted background work follows its existing maintenance lifetime. Normal foreground completion does not cancel it. The existing task-ID marker release point and write-release ordering stay intact so a parent cannot clear or block its rerun's owner. No schema, configuration, public API, timeout-policy or RUN producer-activation changes.

## User Impact

Background compaction can finish and clean up using its original resources after its foreground acknowledgement. Shutdown cancellation continues to reach a disposer that is still running.

## Evidence

Original maintenance fails three actual-tail regressions. The original queued-compaction source also closes SQLite before the real managed primary/runtime-donor path can finish maintenance and disposal. The final native composition passes 40 cases, covering retained sources, reruns, normal foreground completion, cancellation, transcript ownership and disposal. Earlier broader sibling coverage overlaps and is not added to that count.

Affected checks and final proportional checks pass. Fresh full-candidate Codex review through P2 is clean. The ordinary build passed in 322.72 seconds; the existing line limit was preserved by moving the cohesive disposer operation into its work helper.

Compiled runtime proof passed in 3.00 seconds across normal operation, cancellation during disposal and restart. Foreground acknowledgements took 131.42, 7.40 and 6.68 milliseconds. All 11,390 artifacts and six source files stayed unchanged; 3,896 resolved files used no application source fallback, and all processes exited.

Compiled proof exercises actual queued compaction and maintenance with its normal raw registration owner, plus separate managed read-only SQLite close/reopen. The combined managed primary/donor path is maintained native proof using the existing acquisition input seam. SIGTERM was a synthetic event inside the test child, never a signal to an operator process. These are correctness timings, not a speedup benchmark or a claim of general RUN disposal activation.

### DIFF
--- a/docs/concepts/context-engine.md
+++ b/docs/concepts/context-engine.md
@@ -86,6 +86,11 @@ Every time OpenClaw runs a model prompt, the context engine participates at four
 
 Engines can also implement an optional `maintain()` method for transcript maintenance (safe rewrites via `runtimeContext.rewriteTranscriptEntries()`) after bootstrap, a successful turn, or compaction. Set `info.turnMaintenanceMode: "background"` to run it as deferred work instead of blocking the reply.
 
+When queued budget compaction accepts background maintenance, it keeps the prepared
+runtime alive through maintenance, coalesced reruns, and engine disposal. Acceptance
+does not mean cleanup has finished. Return asynchronous work from engine methods
+and `dispose()` so the host can join it before releasing their resources.
+
 For the bundled non-ACP Codex harness, OpenClaw applies the same lifecycle by projecting assembled context into Codex developer instructions and the current turn prompt. Codex still owns its native thread history and native compactor.
 
 ### Subagent lifecycle (optional)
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1891fb443c568ca8db9041f7f61aeb3b4f9cc0b8](https://github.com/openclaw/openclaw/pull/143426/commits/1891fb443c568ca8db9041f7f61aeb3b4f9cc0b8)
- Landed commit: [46af4a9582b34444d11cc98310838060e1aae5ca](https://github.com/openclaw/openclaw/commit/46af4a9582b34444d11cc98310838060e1aae5ca)

PR #143437, opened 2026-09-09
fix(ui): restore Send when Stop finds a finished run

### LINKED ISSUE #143356: [Bug]: Stop button remains after turn finishes; clicking it does nothing until page refresh
### Bug type

### Beta release blocker

No

### Summary

The Control UI stop button stays visible after a turn completes; clicking it is a no-op and only a page refresh restores the normal send icon, although the backend has already finished.

### Steps to reproduce

1. Start a turn that involves tool calls or a longer dialog.
2. Wait until the turn finishes (backend shows no active exec/subagent).
3. Observe the stop button (red square) still present in the compose box.
4. Click it — nothing happens.
5. Refresh the page — the button returns to the normal blue send arrow.

### Expected behavior

The stop button should disappear automatically once the turn ends, without requiring a manual refresh.

### Actual behavior

The stop button remains and clicking it has no effect. Only refreshing the page clears it. It does not consume tokens (no active process), which suggests the frontend never received/processed the turn-close event.

### OpenClaw version

2026.9.3

### Operating system

Windows 11

### Install method

npm global

### Model

opencode-zen/deepseek-v4-flash-vision-exp

### Provider / routing chain

opencode-zen -> deepseek-v4-flash-vision-exp

_No response_

### Logs

```shell

```

_No response_

### Impact and severity

_No response_

### Additional information

_No response_

ylcn91:
Closes #143356.

## What Problem This Solves

Fixes an issue where clicking Stop leaves the Control UI stuck on a completed run when its terminal notifications did not reach the browser. The Gateway answers that nothing is active, but the composer keeps Stop until the page is refreshed.

## Why This Change Was Made

The shared Stop handler now consumes explicit no-active-run answers from `chat.abort` and `sessions.abort` and invokes the existing authoritative chat refresh. The session owner decides whether the run is finished. Recovery checks the captured connection, client, session, run (including no local run), and selected agent before refreshing, so a delayed answer cannot refresh replacement work.

This preserves @ylcn91's contribution and all nine original commits. The regression drives an authenticated Control UI and real Gateway, with a deterministic fixture provider and narrowly withheld terminal notifications. The test owns its server and runs in the serial browser lane.

## User Impact

Stop can restore Send and the completed reply without a page reload. Finalizing work retains its activity and tool progress. Accepted live aborts still settle through lifecycle events. Drafts, queued exact-run stops, descendant activity, error recovery, and Gateway abort authority retain their existing behavior.

## Evidence

- On pinned main `71b21c88574752cd9acc917970ec2850279cdfa2`, one real turn reaches an authoritative inactive/done state. Four exact-run terminal notifications are withheld on one connected WebSocket; ordinary replies and chat deltas continue. Stop receives `{ok:true, aborted:false, runIds:[]}`, yet the browser keeps the same run and Stop for ten seconds without refreshing history.
- On candidate runtime `727574f45fca7a0cf0e15604d7e209cb987d844b`, the same real flow refreshes history after that answer, reads the matching terminal session and run, clears local ownership, shows the finished reply, and restores Send. The full private capture and the corrected committed regression pass. Later test-only corrections read session identity from the Gateway's existing history delta response and omit a redundant default type argument, and include the new private-server test in the existing CI ownership guard; runtime inputs are unchanged.
- 53 focused lifecycle/abort tests, 43 test-routing checks, and both affected serial-ownership controls pass. The two added replacement-run/selected-agent tests first fail on contributor code for the intended ownership violations. Scoped formatting, lint, canonical commit hooks, and normal Gateway/UI builds pass. UI startup size remains within the existing limit.
- All 17 mock-browser scenarios have applicable passing evidence across retained runs, including finalizing tools, accepted abort, session-only direct/descendant recovery, failed history followed by another Stop, and existing reconnect/queued-stop controls. Earlier fixture failures and skipped selections are retained; this is not one new whole-file 17/17 run. Ownership races also have focused tests and source review, rather than an exhaustive browser permutation claim.

Terminal-event withholding is deliberate fault injection. The provider is a deterministic fixture, not live-model proof. This repairs recovery after a no-active-run answer; it does not establish or fix the reporter's natural notification-loss cause. The new real-test draft is entered after recovery; draft survival across Stop is covered separately by focused tests.

The first committed regression expected full-history-only identity fields from a valid delta response and failed after the composer recovered. Its corrected assertion checks the exact identity in `sessionInfo`; the original failure and successful private capture remain in the evidence. Full raw traces and state remain private. Public images are inspected crops that remove session headers and the fixture prompt.

| Before: Stop remains after the no-active-run answer | After: completed reply and Send restored |
| --- | --- |
| Stale Stop after the Gateway completed | Send restored after authoritative refresh |


ylcn91:
Revision 4 (`714510cf750`) addresses both findings; the PR body is updated with the design and evidence.

- **Finalizing runs**: a no-op abort answer no longer clears anything locally. `handleAbortChat` now awaits the page's existing `refreshCurrentChat` (history plus session row), so the run is cleared only when the refreshed Gateway row reports it terminal; a finalizing run keeps its ownership, indicators, and tool activity, and a failed refresh leaves everything untouched.
- **Typed Stop / `/stop`**: the recovery lives in the shared abort owner, so `chat-send-submit.ts` and `chat-commands.ts` callers get the same refresh; the composer callback no longer carries its own.

Re-verified on the same real-Gateway lane (`chat.abort` → `aborted:false` → `chat.history`/`sessions.list` → composer restored, single connection), the mocked lifecycle file (12/12), and the unit file (49/49). Autoreview on this revision: scoped-clean.

ylcn91:
Revision 6 (`9f3cdf863cd`) addresses the session-only Stop finding; the PR body is updated.

- A Stop with no browser-owned run ID (Stop kept visible by cached direct or descendant activity) goes through `sessions.abort`; its explicit `no-active-run` answer now settles through the same authoritative refresh. The settlement is fenced to the client and session key the Stop was captured for, and run-scoped Stops keep their exact-run-ID fence.
- Unit coverage adds the session-only path and the fence (a selection change while the abort is in flight never refreshes the new session); 51/51 in `run-lifecycle.test.ts`.
- Also since the last review: the new real-Gateway proof file is registered as a shared-bundle real-Gateway file (ownership test 43/43) and the `consistent-return` lint stripe is green. `check:changed` and autoreview are green at this head.

obviyus:
The current contribution is nine files. GitHub initially compared the branch to an old base and included 510 files of carried main history; refreshing the same `main` base restored the complete nine-file comparison. The candidate source was not rebased or changed by that metadata correction.

The two compatibility checklist items name `unknown-truncated-pull-files`. The complete current diff changes the Control UI Stop handler, its tests/helper, the UI test inventory and one existing CI test command. It changes no persistent schema, stored data model, product configuration/default, dependency, release workflow or security-owned path. There is no migration or upgrade behavior to waive. The earlier truncated-input review remains recorded; a fresh exact-head review will assess the complete contribution.

The actual CI lint finding is corrected by omitting the existing deferred helper's default `void` type argument. The native check now uses the same core type-aware lint configuration and passes. The real product capture retains its original tested runtime identity because this correction changes no runtime behavior.

obviyus:
Thanks @ylcn91! Merged as 8d96c4fba4d5f412a8b7156773e92f7042552a42.

**Bug:** Stop could remain stuck after the Gateway finished when terminal notifications were missed.

**Fix:** Explicit no-active-run replies now use the existing scoped history/session refresh. Authoritative state decides when to restore Send while preserving live/finalizing work and replacement-context boundaries.

**Proof:** The real authenticated Control UI/Gateway flow fails on pinned main and recovers on the candidate. The provider and narrowly withheld terminal notifications are controlled fixtures; the natural notification-loss cause is not established.

| Before | After |
| --- | --- |
| The Gateway replies `aborted:false`, but Stop remains without a history refresh. | The matching terminal history is loaded, the completed reply remains, and Send returns. |

Focused tests, independent source/behavior acceptance, fresh exact-head review, and [full CI](https://github.com/openclaw/openclaw/actions/runs/34487671845) passed. Earlier failed attempts remain in the retained evidence; no CI exception was used. Contributor history is preserved in the PR branch, with human credit verified at landing.

PR #143440, opened 2026-09-09
fix(cli): reject blank directory group IDs before setup

steipete:
## What Problem This Solves

Fixes an issue where `directory groups members --group-id ""` or a whitespace-only ID could run plugin setup and write configuration before reporting `Missing --group-id`.

## Why This Change Was Made

Move the existing group-ID normalization and check immediately after limit validation, before directory resolution. Valid-ID trimming, setup, lookup and error handling remain unchanged. The CLI reference documents the early rejection.

## User Impact

Invalid group-member requests fail before directory plugin installation, enablement or configuration writes. Valid commands retain their existing behavior.

## Evidence

Two regressions fail on the original code because the configuration-write boundary is reached; the same 43 tests pass after the guard move. Controls cover a valid trimmed ID with setup writes and invalid-limit precedence. All 29 changed-file checks, docs listing and independent review through P2 pass. Tests isolate configuration/provider operations; no real settings or credentials were used.

### DIFF
--- a/docs/cli/directory.md
+++ b/docs/cli/directory.md
@@ -106,6 +106,8 @@ openclaw directory groups list --channel zalouser --query "work"
 openclaw directory groups members --channel zalouser --group-id <id>
 ```
 
+`groups members` requires a non-blank `--group-id`; empty or whitespace-only IDs fail before plugin setup or lookup.
+
 ## Related
 
 - [CLI reference](/cli)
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [247f5140afc78a4fb9dbf86f1292a7bf67091189](https://github.com/openclaw/openclaw/pull/143440/commits/247f5140afc78a4fb9dbf86f1292a7bf67091189)
- Landed commit: [d2e676214b5b3fcdcc3042eb70e18f6a47c63bdb](https://github.com/openclaw/openclaw/commit/d2e676214b5b3fcdcc3042eb70e18f6a47c63bdb)

PR #143455, opened 2026-09-09
fix(docs): detect overdue mirror changes during steady edits

steipete:
## What Problem This Solves

Resolves a problem where continuous documentation edits could keep the docs mirror monitor reporting that synchronization was within its grace period even while older changes remained unpublished. The monitor could also report a change as overdue just before the configured threshold because it rounded the age first.

## Why This Change Was Made

When the latest watched commit is still within grace, the existing commit lookup also checks watched history at the grace cutoff. Grace now applies only when the mirror contains that overdue history. Threshold decisions use the exact timestamp; rounded minutes remain diagnostic output. Existing recovery dispatch and active-run suppression are preserved.

## User Impact

Maintainers receive the intended stale-mirror failure and recovery attempt when older documentation changes remain missing during steady source activity. Genuinely recent changes retain the existing grace period.

## Evidence

Eight real CLI scenarios pass with fetch intercepted and network sockets blocked; three fail for the intended reasons on the original code. Controls cover a current mirror, genuine grace, quiet staleness, active recovery, missing older history, a longer configured grace, and the minute boundary. All 18 selected validation stages pass.

The GitHub List commits `until` contract was checked against its primary documentation and three bounded read-only source-repository requests, including the exact cutoff format and second boundary. No live mirror recovery or workflow dispatch was executed, and workflow settings are unchanged.

Hosted CI exposed an outdated target-list expectation: the new mirror test is correctly selected for docs-sync workflow edits. That expectation now includes it only for docs-sync, with docs-agent and strict ordering preserved. Five focused routing cases and the same eight CLI cases pass, as do all 16 checks selected for this test-only repair. Independent review found no actionable issues in either the production patch or the CI repair.

### DIFF
--- a/scripts/docs-mirror-freshness.mjs
+++ b/scripts/docs-mirror-freshness.mjs
@@ -83,11 +83,12 @@ function toCommitApiPath(pattern) {
   return bare;
 }
 
-async function newestWatchedCommit(watchPaths) {
+async function newestWatchedCommit(watchPaths, until = "") {
   let newest = null;
+  const untilQuery = until ? `&until=${encodeURIComponent(until)}` : "";
   for (const watchPath of watchPaths) {
     const commits = await githubJson(
-      `/repos/${SOURCE_REPO}/commits?sha=main&path=${encodeURIComponent(watchPath)}&per_page=1`,
+      `/repos/${SOURCE_REPO}/commits?sha=main&path=${encodeURIComponent(watchPath)}&per_page=1${untilQuery}`,
     );
     const head = commits[0];
     if (!head) {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [35a01fdcbc05f8e340e3165d09f59214c158e743](https://github.com/openclaw/openclaw/pull/143455/commits/35a01fdcbc05f8e340e3165d09f59214c158e743)
- Landed commit: [ac5ba7d7381e36ca91ba8e5b6ab77e2df9304955](https://github.com/openclaw/openclaw/commit/ac5ba7d7381e36ca91ba8e5b6ab77e2df9304955)

PR #143457, opened 2026-09-09
fix(usage): preserve the selected session owner in details

steipete:
## What Problem This Solves

Fixes an issue where users opening a session in the All-agent Usage view would see missing-owner errors instead of its timeline and conversation when the session used a raw key such as `global`. An overview refresh could also keep details from the previous agent when another agent acquired that same raw key.

## Why This Change Was Made

The detail controller now carries the selected session's key and agent through requests, retained results, and recovery. Raw keys send their owner to the existing Gateway resolver. Qualified keys keep their existing key-only routing, including system sessions outside the configured roster. Automatic refresh retains matching details and replaces them when ownership changes; manual refresh still refreshes all details.

Both Gateway detail methods share key validation and owner resolution, and a single-use context-weight forwarding helper is removed. Existing authorization, key bytes, selection behavior, and storage semantics are preserved.

## User Impact

Users can inspect the correct timeline, conversation, and available context weight for the selected agent's session, including system sessions. Stale responses cannot replace a successor owner's details.

## Evidence

The original Gateway regression and three mounted UI regressions fail on the original production code. A separate system-session control proves why qualified requests must remain key-only. The final owner and sibling suites pass 94 distinct cases across six files, and all 32 selected validation stages pass. Independent review found no actionable issues through P2.

The before/after images use the same synthetic session and mock Gateway response policy in a source-built Control UI. Before, actual key-only detail requests produce missing-owner errors; after, the emitted owner-aware requests render the timeline and conversation. This is isolated browser and Gateway fixture proof, without live operator data. All full images were inspected before upload.

Before: selected session details fail with a missing owner

After: the selected session timeline loads

After: the selected session conversation loads

### DIFF
--- a/docs/gateway/protocol/rpc-methods.md
+++ b/docs/gateway/protocol/rpc-methods.md
@@ -45,6 +45,7 @@ methods. Treat this as feature discovery, not a full enumeration of
   Both usage methods accept `mode: "specific"` with an IANA `timeZone` for DST-aware calendar-day boundaries and buckets. `utcOffset` remains supported for older clients and as a fallback when the Gateway runtime does not recognize the requested zone.
 - `sessions.usage.timeseries` returns timeseries usage for one session.
 - `sessions.usage.logs` returns usage log entries for one session.
+  Both detail methods accept the selected row's `key` and optional `agentId`. Preserve both fields when opening details for an unqualified key such as `global`.
 
 ### Channels and login helpers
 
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [495de05d3b2aa02108ed5b14b182ae91e34e47ad](https://github.com/openclaw/openclaw/pull/143457/commits/495de05d3b2aa02108ed5b14b182ae91e34e47ad)
- Landed commit: [46cfa0e410272510aa5bdb682932cc2614acb0ac](https://github.com/openclaw/openclaw/commit/46cfa0e410272510aa5bdb682932cc2614acb0ac)

PR #143403, opened 2026-09-09
fix: retain image provider state through request cleanup

steipete:
Related: #140674.

## What Problem This Solves

Resolves a problem where image requests can close plugin databases while timed-out setup, provider calls, or cleanup are still running. Supplied prepared runtimes also need to retain the original provider and any adopted donor registrations through the request.

## Why This Change Was Made

Transfer resource ownership immediately after acquisition, before asynchronous model and auth setup. Keep the existing caller deadlines while tracking the actual setup and provider work separately, then release after cooperating cleanup settles. Supplied managed snapshots retain their exact prepared owner, including donors; raw snapshots retain their existing caller ownership.

Closed managed sources refuse new generic and MiniMax calls, including retries and each image in a MiniMax batch. The private early handoff removes competing failed-setup and late-result release paths. No public SDK fields, configuration, schema, dependency, producer-activation or timeout-policy changes.

## User Impact

Image requests keep their existing cancellation and timeout behavior while allowing already accepted work to finish using the correct retained resources. Cleanup closes each owned database once after the final user.

## Evidence

Sixteen native SQLite cases cover timeout, cancellation, late rejection, supplied and raw owners, adopted donors, setup descendants, cleanup, and closed-source admission. Original code reproduces five ownership failures with two controls. Review found two further defects in the intermediate implementation—failed setup releasing before its descendants and MiniMax missing per-image closure checks—and both have failing-before, passing-after regressions.

Focused Image, timeout, auth-profile and provider/tool suites pass. Complete affected checks, proportional checks after the final narrow fence change, and fresh full-candidate Codex review through P2 pass. The ordinary build passed in 254.76 seconds; its nonfatal UI gzip warning of 20 bytes over the threshold is recorded, with no UI or budget changes.

The actual compiled public SDK and compiled runtime producer passed together in 6.42 seconds. A successful request took 83.88 milliseconds; the unchanged 1-second timeout reported at 1.004 seconds. After replacement, the original provider and donor databases stayed usable through late work. Closed-source admission was refused, all three fixture databases closed exactly once and reopened with their expected values, and the successor received no calls.

All seven source hashes and 11,683 build artifact hashes stayed unchanged; the plain-Node process and descendants exited. The proof uses synthetic SQLite stores and a valid PNG, without application source fallback, external credentials, a real provider or a production Gateway. Timings are individual correctness observations, not a comparative performance claim.

PR #143460, opened 2026-09-09
fix: retain PDF provider state through cancelled work

steipete:
Related: #140674. Uses the prepared-owner helper from #143403.

## What Problem This Solves

Resolves a problem where PDF requests can release plugin databases while a cancelled provider call or setup cleanup is still running. A supplied prepared runtime also needs to remain alive while the tool downloads its documents.

## Why This Change Was Made

Own supplied resources before the first download, and transfer default acquisition into that same execution owner before asynchronous setup. Track actual stream creation, completion and cooperating cleanup beneath the existing abort race, then release after drainage.

The normal PDF operation stays separate from its small lifetime wrapper. Default acquisition still happens once after loading and before fallback. Native provider routing, extraction, filesystem checks, model/auth selection and existing timeouts are unchanged. Raw supplied registries retain their caller lifetime; no general RUN/configured disposal is enabled.

## User Impact

Cancelled PDF requests preserve their current response behavior while accepted work can finish using its original state. Supplied managed runtimes retain their original provider and adopted donor resources through the operation.

## Evidence

Original production reproduces eight ownership failures with two passing controls. Final native and sibling validation covers 120 distinct PDF cases. Two existing release assertions now join actual cleanup before checking release; their model, routing and output assertions remain intact.

Affected types, lint, all three Knip scans and remaining guards pass, with fresh proportional checks after fixture-only lint corrections. Full-candidate Codex review through P2 is clean. The ordinary build passed in 269.29 seconds, including 152 public SDK subpaths.

Compiled proof uses the actual public agent-harness tool factory and PDF tool, a real generated PDF, bundled document extraction and a native SQLite-backed provider. After cancellation and replacement, the old source remains usable for late work; both sources close once and reopen successfully. The proof passed in 2.45 seconds with all 11,688 artifact hashes unchanged, 3,799 resolved files, no application source fallback and all processes joined.

Single synthetic samples: cold PDF execution 75.00 milliseconds and cancellation response 13.83 milliseconds. Five failed fixture configurations are preserved; correcting the fixture's agent entry, syntax and provider/extractor scope required no application change. This is a built-checkout proof, not installed-package or real-vendor coverage, and the timings are not a comparative benchmark.

Only the PDF commit was rebased after its Image dependency landed; all four validated source files remain byte-identical.

### DIFF
--- a/src/agents/tools/pdf-tool.native-providers.test.ts
+++ b/src/agents/tools/pdf-tool.native-providers.test.ts
@@ -5,6 +5,7 @@ import path from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import type { OpenClawConfig } from "../../config/config.js";
 import * as pdfExtractModule from "../../media/pdf-extract.js";
+import { AsyncWorkScope } from "../../shared/async-work-scope.js";
 import * as preparedModelRuntime from "../prepared-model-runtime.js";
 import * as pdfNativeProviders from "./pdf-native-providers.js";
 import {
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d6bebc93e91e4e4c856d8dea024be6defc208f44](https://github.com/openclaw/openclaw/pull/143460/commits/d6bebc93e91e4e4c856d8dea024be6defc208f44)
- Landed commit: [62c3fb73f352f56075b77d53d7be97db8bd4a6e1](https://github.com/openclaw/openclaw/commit/62c3fb73f352f56075b77d53d7be97db8bd4a6e1)

PR #143470, opened 2026-09-09
refactor(channels): reuse normalized setup catalog metadata

steipete:
## What Problem This Solves

Channel setup normalized catalog metadata while building its installed and installable buckets, then repeated that same work in two separate display-list loops. The second pass could never supply an existing fallback because it ran only for missing display entries.

## Why This Change Was Made

One ordered loop now reuses the already-normalized catalog records. Built-in metadata, loaded-plugin overrides, catalog filtering and the separate installed/installable maps retain their existing owners and precedence. This removes 19 production lines and two unnecessary type assertions; the assertion baseline shrinks by exactly those two assertions.

## User Impact

No user-visible behavior changes. Setup keeps the same labels, fallbacks, display order and installation choices with less duplicate processing.

## Evidence

The extended existing fixture passes before and after the refactor. It covers partial loaded metadata, built-in/catalog collisions, installed and installable entries, optional metadata, display order, and both lookup maps. All 76 owner and sibling cases pass, along with all 30 selected validation stages. The first full check required the expected assertion-baseline reduction from 10 to 8; the final full check passes with that exact shrink. No runtime import, configuration or storage changes.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -2356,7 +2356,7 @@ src/commands/backup-restore.ts	1
 src/commands/backup-schedule.ts	1
 src/commands/backup-verify.ts	2
 src/commands/channel-setup/config-compatibility.ts	2
-src/commands/channel-setup/discovery.ts	10
+src/commands/channel-setup/discovery.ts	8
 src/commands/channel-test-registry.ts	2
 src/commands/channels/add-wizard.ts	1
 src/commands/channels/add.ts	2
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [39092ccfb84ae19db808296df62baf5d6097ef3c](https://github.com/openclaw/openclaw/pull/143470/commits/39092ccfb84ae19db808296df62baf5d6097ef3c)
- Landed commit: [f1ff9d895fc81fe3db7a732f4aa58749b4a820d7](https://github.com/openclaw/openclaw/commit/f1ff9d895fc81fe3db7a732f4aa58749b4a820d7)

PR #143474, opened 2026-09-09
fix(triage): prevent Claude startup hooks from blocking repair

RomneyDa:
## What Problem This Solves

Fixes an issue where `openclaw triage`, including failed-update recovery, launches Claude Code but appears not to submit its repair prompt when the current project's startup hooks block. In the reported case, a dependency-install hook delayed the first prompt for ten minutes before timing out; the prompt itself was passed correctly.

## Why This Change Was Made

Interactive Claude handoffs now use `--safe-mode`, matching automatic triage. Recovery starts without project hooks, plugins, or project instructions while retaining Claude's authentication, built-in tools, and permission settings. The session remains interactive and receives the original positional prompt. Documentation explains this behavior.

## User Impact

Operators can reach Claude's repair turn without waiting for unrelated project setup. Other supported agents keep their existing launch behavior.

## Evidence

- Original session evidence confirmed a 600-second startup-hook timeout followed by the unchanged repair prompt.
- Regression proof: original launch arguments fail eight assertions; the fix passes all 60 focused triage and recovery tests, including Windows shim and sibling-agent coverage.
- Real Claude probes verified that safe mode skips a synthetic startup hook and that an interactive Clack-to-Claude handoff displays and answers its positional prompt without manual submission.
- Production and test typechecks, focused lint, documentation checks, formatting, and diff checks pass. Independent autoreview is scoped-clean at P0.
- `check:changed` stops on an unrelated, unchanged compatibility record whose removal deadline has expired (`sdk-untrusted-context-identifier-aliases`).

### DIFF
--- a/docs/cli/triage.md
+++ b/docs/cli/triage.md
@@ -17,6 +17,8 @@ openclaw triage
 
 In an interactive terminal, triage starts the first directly launchable agent on `PATH` in this detection order: Claude Code (`claude`), Codex (`codex`), OpenCode (`opencode`), then Pi (`pi`). An explicit `openclaw triage` invocation prints the selected agent and passes a bounded repair prompt directly, without a picker or launch confirmation. The agent uses its existing authentication, sandbox, and approval settings.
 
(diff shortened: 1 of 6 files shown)


RomneyDa:
@clawsweeper re-review

PR #143484, opened 2026-09-09
fix(feishu): reject replayed webhook callbacks with stale signed timestamps

eleqtrizit:
## Problem

Feishu webhook signature verification authenticated requests purely by their
signature (`sha256(timestamp + nonce + encryptKey + body)`) with no check on
the signed timestamp's age. Feishu signs each delivery at send time, so a
captured signed callback stays validly signed forever.

Card-action callbacks are excluded from durable event admission and replay
protection is handled downstream: synthetic card commands are deduplicated by
a stable `card-action-<token>` key with a 24h persistent TTL. That gate
already blocks replays inside the TTL window (including across restarts),
but once the entry expires, replaying a captured callback re-authenticated and
re-dispatched the same command — for quick-button and unrecognized-envelope
card actions, which carry no expiry of their own.

## Fix

Reject callbacks whose signed `x-lark-request-timestamp` is more than one
hour from the local clock (symmetric: stale or far-future), inside
`isFeishuWebhookSignatureValid`, before any JSON parsing:

- hardcoded `FEISHU_WEBHOOK_TIMESTAMP_MAX_SKEW_MS` (1h) — generous enough for
  provider retries and host clock skew, far shorter than the 24h dedup TTL, so
  auth + dedup together close replay with no gap;
- second-level Feishu timestamps normalized by magnitude (values < 1e12 treated
  as seconds, otherwise milliseconds).

No config keys, no default changes, no schema changes, no public API changes.

## Tests

- regression: correctly signed callbacks with a stale (−2h) or far-future (+2h)
  timestamp are rejected with 401 before parsing;
- acceptance: a correctly signed callback near the window edge (−55 min) still
  completes the `url_verification` challenge end-to-end;
- webhook test helpers now sign with fresh timestamps instead of a hardcoded
  stale one;
- full extension suite passes (102 files / 2051 tests).

### DIFF
--- a/docs/channels/feishu/setup.md
+++ b/docs/channels/feishu/setup.md
@@ -40,3 +40,14 @@ The wizard also asks for the API domain (Feishu vs Lark) and the group policy. I
 OpenClaw durably queues authenticated `im.message.receive_v1` and `drive.notice.comment_add_v1` envelopes before agent dispatch. In webhook mode, the durable `200` carries `x-openclaw-delivery-accepted: durable`; verification challenges, non-durable event types, and error responses omit the marker, so reverse proxies can require it to distinguish durable acceptance from a generic `200`. Pending or retryable events survive a Gateway restart, remain serialized per chat or document, and use Feishu's event ID to suppress duplicate queue entries while the active or retained completion record exists.
 
 If a WebSocket event cannot be persisted after bounded retries, OpenClaw closes that socket and forces a fresh authenticated connection instead of continuing past an uncommitted turn. Other Feishu event types, including reactions and VC meeting invitations, use their normal event paths and do not receive this durable-queue guarantee.
+
+## Webhook delivery window
+
+Feishu signs each webhook delivery at send time, so a captured signed callback stays validly signed forever. Webhook mode therefore rejects any signed callback whose `x-lark-request-timestamp` is more than one hour before or after the Gateway host's clock, before the body is parsed. This is a replay defense: it works together with the per-message replay guard (24-hour window) so a re-delivered signed callback cannot re-trigger the same action.
+
+Practical consequences:
+
+- Keep the Gateway host clock synchronized (NTP). A host clock drifting more than one hour from Feishu's servers rejects fresh deliveries.
+- Feishu deliveries are signed when they are sent, so ordinary redeliveries carry fresh timestamps and are unaffected.
+- WebSocket mode is not affected by this window.
+- There is no configuration key for this window; it is intentionally fixed.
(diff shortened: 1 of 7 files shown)


eleqtrizit:
Addressing the review's before-merge items:

**1. Owner decision on the cutoff (accepted).** The repository owner reviewed and explicitly accepts the documented delivery limit: signed webhook deliveries whose timestamp is more than one hour from the host clock are rejected by design. This is a replay defense — a captured signed callback stays validly signed forever, and the cutoff is what makes replay rejection independent of downstream replay-claim state. Operators are required to keep the host clock synchronized.

**2. Provider redelivery compatibility.** Feishu signs each delivery at send time, so redeliveries carry fresh timestamps and ordinary retry traffic is unaffected; only deliveries retaining an original signature older than the window (the replay case) are dropped. Upgrade impact is nil: there are no config keys, no defaults, and no migration steps — the window is fixed in code and applies automatically.

**3. Documentation added (9e20acb66eb):**
- `docs/channels/feishu/setup.md`: new *Webhook delivery window* section — what the window is, why it exists, NTP requirement, retry behavior, WebSocket-mode exemption, and that the window is intentionally not configurable.
- `docs/channels/feishu/troubleshooting.md`: new *Webhook callbacks rejected with 401 Invalid signature* entry — clock-skew diagnosis and recovery steps, plus encryptKey/path checks.

@clawsweeper re-review

eleqtrizit:
## Merge record

**Summary:** Feishu webhook handling now rejects signed callbacks whose signed timestamp is more than one hour from the Gateway clock, closing a replay window where captured signatures stayed valid indefinitely and message deduplication (24h TTL) was the only replay barrier. Adds a symmetric freshness check at the existing signature boundary, extends boundary tests (±2h rejections, near-window acceptance), and documents the delivery window plus 401 troubleshooting.

**Decision Reason for Merge:** Merged because the independent merge review returned MERGE AS-IS, the Clawsweeper security review loop converged with zero findings at ready-for-maintainer-look status, CI is fully green (118 passing checks, 0 failing), and the repository owner explicitly instructed the merge with the documented one-hour webhook window accepted.

**Compatibility Checks Done:** Full extension suite 2051 tests passing; CI on the merged head fully green (118 checks, 0 failing); security compatibility review SAFE (no config keys, defaults, schema, or public API changes; behavior change declared and documented); independent merge review MERGE AS-IS; Clawsweeper loop round 2: zero findings, ready-for-maintainer-look, remaining P1 merge-risk item (deliveries older than one hour now return 401) explicitly accepted by the repository owner.

PR #143485, opened 2026-09-09
fix(cli): allow short diagnostic flags on unsupported Node

steipete:
## What Problem This Solves

Fixes an issue where users running `openclaw -h` or `openclaw -v` on a capable but unsupported Node runtime would be refused or offered runtime installation even though the equivalent long help/version flags remain available for diagnostics.

## Why This Change Was Made

The shared diagnostic classifier now recognizes the two aliases already supported by the launcher. Runtime capability checks, the minimum diagnostic Node version, command-tail validation, and recovery behavior remain unchanged.

## User Impact

Short and long help/version flags behave consistently when users need installation or runtime recovery information. The installer reference now lists the supported aliases.

## Evidence

Both aliases fail on the original source-guard and packaged-launcher paths: four intended failures across 27 admission cases. The same 27 pass after the one-line repair; the complete source-guard suite passes 78 cases and the selected packaged table passes 12. All 30 canonical changed-check stages pass, including all typecheck lanes, export checks and lint.

Packaged proof runs a copied launcher with isolated home/state, a synthetic Node 22 version label and the current host SQLite. It exercises the actual diagnostic admission branch without installing an obsolete Node binary or operating a live updater. No performance or native Windows recovery claim is made.


steipete:
Merged via squash.

- Prepared head SHA: [2278fbc437950fdbcf49e5d65d94aa4261b55c0b](https://github.com/openclaw/openclaw/pull/143485/commits/2278fbc437950fdbcf49e5d65d94aa4261b55c0b)
- Landed commit: [e1a5d8041d4983cd4c2e7badd3d38fc0cf2f003e](https://github.com/openclaw/openclaw/commit/e1a5d8041d4983cd4c2e7badd3d38fc0cf2f003e)

PR #143488, opened 2026-09-09
fix: keep sandboxed worktree creation isolated

eleqtrizit:
## What Problem This Solves

Fixes an issue where a sandboxed agent creating a visible worktree could cause the Gateway to execute repository-selected transport commands while resolving the default worktree base.

## Why This Change Was Made

For same-agent spawns whose source workspace is both sandboxed and writable, an omitted base now resolves and pins the already accepted local `HEAD` before managed-worktree preparation. Trusted operator, registered-project, cross-agent, reuse, and explicit-base paths keep their existing behavior.

This is a narrow containment for implicit remote refresh. It avoids disabling legitimate repository-local Git settings or requiring a sandbox backend solely to create a worktree.

## User Impact

Sandboxed read-write agents can continue creating child worktrees normally. Those spawns no longer refresh `origin` automatically when no base is selected; they start from local `HEAD`. Users can still select an already available local or remote ref explicitly.

Other worktree creation flows are unchanged.

## Evidence

- Added a Gateway-level agent-spawn regression using a disposable SSH origin and benign marker-writing repository command.
- Before the fix, the regression observed the Gateway-host marker.
- After the fix, worktree creation succeeds, records `HEAD`, and the marker remains absent.
- `src/gateway/server.sessions.create-worktree-spawn.test.ts`: 14 passed.
- `src/gateway/server-methods/sessions-create-worktree-base.test.ts`: 6 passed.
- Core TypeScript check passed.
- Targeted formatting and diff checks passed.

### DIFF
--- a/src/gateway/server-methods/sessions-create.ts
+++ b/src/gateway/server-methods/sessions-create.ts
@@ -10,6 +10,7 @@ import {
   validateSessionsCreateParams,
 } from "../../../packages/gateway-protocol/src/index.js";
 import { resolveAgentWorkspaceDir } from "../../agents/agent-scope.js";
+import { resolveSandboxWorkspaceAuthority } from "../../agents/sandbox/workspace-authority.js";
 import { insideGitCheckout } from "../../agents/worktrees/git.js";
 import { resolveAgentMainSessionKey } from "../../config/sessions/main-session.js";
 import { sessionEntryForkedFromParent } from "../../config/sessions/session-entry-lineage.js";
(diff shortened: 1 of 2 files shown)


eleqtrizit:
My agent implemented the wrong plan entirely.  Redoing....

eleqtrizit:
Closing this prototype because it implements only the narrow no-fetch containment. The replacement will enforce the complete sandboxed worktree-provisioning boundary before a new PR is opened.

PR #143492, opened 2026-09-09
fix: retain model catalog resources through projections

steipete:
Related: #140674

## What Problem This Solves

Resolves a resource-lifetime gap in finite catalog projections used by local CLI and TUI model lists. Synthetic SQLite tests demonstrate a managed publication’s connection closing while its asynchronous projection is still held.

This is preparation for broader managed-runtime ownership. It does not enable general physical disposal of configured registrations.

## Why This Change Was Made

Finite reads acquire the exact existing publication before its snapshot crosses an await. They retain resources through materialization and the awaited callback, releasing them on success, failure, or config rejection. Acquisition reuses existing generation-retention logic shared with BTW.

Passive lookup, exact config matching, explicit refresh intent, fallback selection, and stale-result rejection remain unchanged. Bare snapshot aliases keep their existing contract; fresh standalone activation and pending-publication handoff remain separate work.

## User Impact

Finite projections can safely consume managed publications. The actual model-list projector still rejects replaced publications, and ordinary reads retain their passive discovery and activation policy.

## Evidence

- Three native SQLite regressions fail originally for early closure and pass with the fix, including release before the callback’s first microtask. Final release closes once and persisted data reopens successfully.
- 90 focused and sibling tests pass, covering catalog policy, BTW cleanup, CLI projection, and model-list freshness fences. All required changed-file checks pass; independent Codex review found no actionable issues.
- Normal build passes in 261.29 seconds. It reports a nonfatal UI startup gzip warning of 42 bytes over budget; no UI or guard changes are included.
- Two compiled probes pass: managed passive acquisition with SQLite replacement, and the actual finite model-list flow with stale-result rejection. All 11,833 sealed artifacts remain unchanged. Combined managed-source callback coverage uses the maintained native input adapter; the compiled model-list flow uses normal standalone publication.
- Observed passive acquisition: 0.148 ms; warm model-list projections: 1.84/1.35 ms. These are individual timings, not a statistical speedup claim.

### DIFF
--- a/src/agents/prepared-model-catalog.resources.test.ts
+++ b/src/agents/prepared-model-catalog.resources.test.ts
@@ -1,0 +1,188 @@
+import fs from "node:fs";
+import path from "node:path";
+import { DatabaseSync } from "node:sqlite";
+import { expectDefined } from "@openclaw/normalization-core";
+import { expect, it, vi } from "vitest";
+import type { OpenClawConfig } from "../config/types.openclaw.js";
+import {
+  cleanupPluginLoaderFixturesForTest,
+  resetPluginLoaderTestStateForTest,
+} from "../plugins/loader.test-fixtures.js";
+import { clearPluginMetadataLifecycleCaches } from "../plugins/plugin-metadata-lifecycle.js";
+import { createColdPluginFixture } from "../plugins/test-helpers/cold-plugin-fixtures.js";
+import { createDeferredCore } from "../shared/deferred.js";
+import { withEnvAsync } from "../test-utils/env.js";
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [51339cc6506a72790f0fc4724c3c4f8857cf20e7](https://github.com/openclaw/openclaw/pull/143492/commits/51339cc6506a72790f0fc4724c3c4f8857cf20e7)
- Landed commit: [60fabc708423eb36577fafbce9470828e162eb31](https://github.com/openclaw/openclaw/commit/60fabc708423eb36577fafbce9470828e162eb31)

PR #143493, opened 2026-09-09
fix: retain prepared speech overrides until host cleanup

steipete:
## What Problem This Solves

Fixes an issue where plugins preparing speech through `api.runtime.tts.prepareTtsRequest(...)` could receive overrides that stopped working after the original plugin inspection was released. VoiceCall and Discord prepare requests separately from synthesis, so an override can still be needed after preparation returns. A preparation that failed while projecting directive metadata could also leave background work accessing SQLite after its registration resources had closed.

## Why This Change Was Made

The public async preparation facade now retains selected managed registration sources in the existing SDK resource host before provider callbacks run, preserving opaque override objects and allowing host shutdown to join tracked work before disposal, including work started by failed projections. Claims use existing per-source deduplication and preserve canonical raw loading, cache reuse, and the public Promise signature without creating a fresh registration per utterance.

## User Impact

Plugins can use prepared overrides in subsequent synthesis after releasing the originating inspection. Provider-owned nested values retain their identity and stay usable for the SDK host's lifetime. Repeated preparation reuses its registration source rather than accumulating registrations for each utterance.

Failed preparations may also retain a selected managed source until the SDK host closes. Raw registrations retain their existing caller or process lifetime, and public synchronous directive and lookup helpers remain unchanged.

Accounting for registrations copied from a managed donor remains a separate prerequisite; this PR does not claim to repair donor provenance.

## Evidence

- Original production bytes reproduced six ownership failures while two raw lifetime/cache controls passed. A combined projection-failure and pending-work regression also reproduced premature SQLite closure before the final repair.
- **264 unique tests passed**, including native preparation, buffered/Talk/Telephony resource tests, provider selection, SDK ownership, directives, and VoiceCall consumers. Required changed-file checks and affected follow-up checks passed.
- Fresh full-candidate **Codex review** reported no actionable P0–P2 findings. `pnpm build` passed in **304.75 seconds**; its non-fatal UI gzip report was seven bytes over budget, with no UI or budget changes.
- **Four compiled public-API probes passed:** 19 preparation attempts and three Telephony calls, three real PCM/WAV roundtrips, and six SQLite stores closed and reopened. All processes exited; 11,392 artifact hashes remained unchanged, with **no TypeScript source fallback**.

### DIFF
--- a/docs/plugins/sdk-overview/host-hooks.md
+++ b/docs/plugins/sdk-overview/host-hooks.md
@@ -68,7 +68,7 @@ These operations retain managed registrations and preserve raw host ownership.
 Configured fallback catalogs stay separate from direct preference and override
 lookups. Returned audio buffers outlive registration disposal; standard TTS
 transcodes and saves those completed buffers after releasing the provider. The
-separate synchronous speech lookup and request-preparation APIs keep their
+separate synchronous speech lookup and directive-parsing APIs keep their
 existing caller lifetime; streaming speech is not part of this finite operation.
 
 Streaming speech owns its provider registrations until stream cleanup finishes.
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b33b5aa810c82cdae6046dc522845c0324a8fb2c](https://github.com/openclaw/openclaw/pull/143493/commits/b33b5aa810c82cdae6046dc522845c0324a8fb2c)
- Landed commit: [51b7e2e9ec35dffcd1a9b26d11224bfe8ce9aac4](https://github.com/openclaw/openclaw/commit/51b7e2e9ec35dffcd1a9b26d11224bfe8ce9aac4)

PR #143498, opened 2026-09-09
refactor(gateway): remove unused no-op heartbeat config parameter

vincentkoc:
## What Problem This Solves

The no-op heartbeat runner declares a configuration parameter it never uses, making the inert implementation look configuration-dependent.

## Why This Change Was Made

This requested cleanup removes that unused parameter and its type-only import. The `HeartbeatRunner` annotation, real runner interface, callers, both no-op methods, and existing test assertions remain unchanged.

## User Impact

No intended user-visible change. The no-op method's JavaScript arity changes from one to zero; callers can still pass configuration through the unchanged typed interface.

## Evidence

- 45 tests passed across the same two whole Gateway suites before and after the change, with no failures or skips.
- 28 changed checks passed, including three Knip scans with zero findings.
- P3 autoreview completed with no actionable findings.
- Validation was local on Linux; no CI or performance result is claimed.

### DIFF
--- a/src/gateway/server-runtime-service-shared.ts
+++ b/src/gateway/server-runtime-service-shared.ts
@@ -1,6 +1,5 @@
 // Shared Gateway runtime service helpers.
 // Supplies minimal service handles for tests and reduced startup paths.
-import type { OpenClawConfig } from "../config/types.openclaw.js";
 import type { HeartbeatRunner } from "../infra/heartbeat-runner.js";
 
 export type GatewayRuntimeServiceLogger = {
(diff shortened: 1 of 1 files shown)


vincentkoc:
Closing this as a maintainer product decision, not as fixed or superseded.

The branch removes one unused parameter from the no-op heartbeat implementation while preserving the existing interface and behavior. It has no user-visible defect or operational outcome to justify a standalone landing cycle, and its exact-head CI is not green.

If a future heartbeat repair naturally removes this parameter, that cleanup can land inside the concrete deliverable.

PR #143519, opened 2026-09-09
fix(mac-gateway): stop discarding launchd stderr

### LINKED ISSUE #90711: [Bug] launchd plist StandardErrorPath hardcoded to /dev/null, hides all gateway stderr (5.28 regression)
## Summary

After 5.28, OpenClaw's generated `~/Library/LaunchAgents/ai.openclaw.gateway.plist` contains `<key>StandardErrorPath</key><string>/dev/null</string>`, discarding all gateway stderr (including `[prompt-cache] cache read dropped` diagnostics, model fallback warnings, liveness warnings). This is a regression from earlier versions where stderr was written to `~/.openclaw/logs/gateway.err.log` or `~/Library/Logs/openclaw/gateway.err.log`.

## Steps to reproduce

1. Install or upgrade OpenClaw (currently 2026.6.1)
2. Run `openclaw gateway install` (or rely on `rewriteLaunchAgentPlistForRestart`)
3. Inspect `~/Library/LaunchAgents/ai.openclaw.gateway.plist`:

   ```xml
   <key>StandardErrorPath</key>
   <string>/dev/null</string>   <!-- bug -->
   <key>StandardOutPath</key>
   <string>.../Library/Logs/openclaw/gateway.log</string>
   ```

4. Trigger any error condition in gateway (e.g., force a cache miss / fallback)
5. Run `ls -la ~/Library/Logs/openclaw/gateway.err.log` → file does not exist or stays at 0 bytes
6. Run `launchctl print ai.openclaw.gateway | grep -E "OutPath|ErrorPath"` → confirms `/dev/null`

## Expected

stderr should be written to `~/Library/Logs/openclaw/gateway.err.log` (alongside `gateway.log`), or at minimum to a real file (not `/dev/null`). Users should be able to diagnose cache / fallback / liveness issues from stderr.

## Actual

stderr is silently discarded. Diagnostics impossible for any post-5.28 incident unless the user runs gateway manually with `2> some-file` redirect.

## Root cause

`/opt/homebrew/lib/node_modules/openclaw/dist/launchd-DirYWoav.js` contains a hardcoded constant near the top of the file:

```js
const LAUNCH_AGENT_STDERR_PATH = "/dev/null";
```

This constant is passed as `stderrPath` to `buildLaunchAgentPlist` at two call sites in the same file (one in `writeLaunchAgentPlist`, one in `rewriteLaunchAgentPlistForRestart`), so:

- A fresh install writes `/dev/null` to plist
- A `rewriteLaunchAgentPlistForRestart` after a restart also writes `/dev/null`
- Any user-side fix to the plist gets overwritten on the next gateway restart

A `resolveMacLaunchAgentLogPaths(env)` helper already exists in `/opt/homebrew/lib/node_modules/openclaw/dist/restart-logs-*.js` and correctly computes `stderrPath: <logDir>/<prefix>.err.log`, but it is **not used** by the launchd plist generator.

## Impact

- All `[prompt-cache] cache read dropped` events lost → cannot diagnose cache hit rate regressions
- All `[model-fallback/decision]` and `[diagnostic]` warnings lost
- Affects every macOS user running gateway via launchd (the default install path)
- Related user reports: #56543 (Strongly +1 on `doctor --fix` detecting stale plist), #65734 (referenced the now-discontinued stderr path)

## Suggested fix

1. Remove the hardcoded `LAUNCH_AGENT_STDERR_PATH = "/dev/null"` constant
2. Pass `stderrPath: resolveMacLaunchAgentLogPaths(env).stderrPath` (or equivalent) to `buildLaunchAgentPlist`
3. Verify the resulting plist contains a real file path (not `/dev/null`)
4. (Optional) Add a `doctor --fix` check that detects plists with `StandardErrorPath = /dev/null` and repairs them

## Environment

- OpenClaw 2026.6.1 (regression introduced in 5.28)
- macOS launchd (user LaunchAgent)
- plist path: `~/Library/LaunchAgents/ai.openclaw.gateway.plist`

## Related

- #56543 (stale plist detection)
- #65734 (referenced previous stderr path)
- Local workaround: `plutil -replace StandardErrorPath -string "~/Library/Logs/openclaw/gateway.err.log" ~/Library/LaunchAgents/ai.openclaw.gateway.plist` + `launchctl bootout/unload/load`

RomneyDa:
## What Problem This Solves

Fixes #90711. On macOS the LaunchAgent plist hard-coded `StandardErrorPath` to `/dev/null` (`src/daemon/launchd-service-files.ts`), so any gateway that died *before* its logger initialized left no trace anywhere. The operator saw a `launchctl` service in a crash loop with an empty `gateway.log`.

The failure is worse than "one log is missing", because a consumer already depends on the opposite contract. `readLastGatewayErrorLine` (`src/daemon/diagnostics.ts`) deliberately skips stderr on darwin:

```
const readStderr = platform !== "darwin";
// launchd supervisor mode combines child stderr into stdout; other platforms
// keep stderr as the strongest failure signal.
```

Nothing combined the streams. The producer discarded stderr and the consumer skipped it because it believed the producer had merged it, so darwin startup crashes fell through the gap entirely. `openclaw gateway status` also advertised `Launchd stderr (if installed): suppressed`, stating the discard as if it were intended behavior.

Real-world cost, hit on a live install today: the gateway refused to start on a SQLite schema check and crash-looped under launchd with `last exit code = 78 (EX_CONFIG)`. `gateway.log` contained nothing after the SIGTERM of the previous process, and `openclaw doctor` reported "Doctor complete". The reason was only recoverable by re-running the plist's exact `ProgramArguments` by hand with `2>&1`.

## Why This Change Was Made

The fix makes the already-documented merge real instead of adding a second log file: both plist handles now point at the resolved stdout path, so launchd itself combines the streams.

This is deliberately **not** the approach of the earlier attempt (#90828, closed unmerged). That PR routed stderr to a separate `gateway.err.log`, which created a new uncapped file and therefore a new retention-ownership question — "which restart seams own the cap" — and stalled waiting on that maintainer decision. Reusing the existing stdout log leaves supervisor-log retention byte-for-byte as it is today, so this change carries no retention decision at all. If a retention policy is chosen later it covers both streams uniformly, for free.

Because the merge is now genuine, `diagnostics.ts` needs no logic change — its darwin branch becomes correct as written. Its comment is updated to name the coupling so the plist side cannot be silently reverted.

## User Impact

macOS operators whose gateway fails to start now get the reason in `~/Library/Logs/openclaw/gateway.log` instead of silence. `openclaw gateway status` reports the real location rather than claiming stderr is suppressed. No config, no new file, no migration, no behavior change on Linux or Windows.

## Upgrade compatibility

The plist is generated content, not user data, so there is nothing to migrate. Existing installs still carry a `/dev/null` plist until a definition-writing path runs: `rewriteLaunchAgentPlistForRestart` regenerates and reloads the definition on any ordinary `openclaw gateway restart` (`src/daemon/launchd-lifecycle.ts:290-298`), and a fresh `openclaw gateway install` writes it directly. Restarts that pass `preserveDefinition` deliberately skip the rewrite, so those keep the old plist until the next definition-writing restart — which is existing behavior for every plist field, not something this change introduces.

## Evidence

Executable production change is net **-2 lines** (deleted the `LAUNCH_AGENT_STDERR_PATH` discard constant and the false status hint); the remaining additions are the two coupling comments and the regression test.

Regression test `points launchd stderr at the stdout log so startup crashes survive` asserts the `key`→`value` pairing rather than bare containment, because `StandardInPath` is also `/dev/null` and a `toContain("/dev/null")` assertion cannot distinguish the two — the pre-existing assertions had exactly that weakness.

Verified it fails on pre-fix code for the intended reason, by restoring `src/daemon/launchd-service-files.ts` from `HEAD` and re-running:

```
× launchd install > points launchd stderr at the stdout log so startup crashes survive
  → expected '<?xml version="1.0" ...' to contain '<key>StandardErrorPath</key>\n    <st…'
Tests  1 failed | 184 skipped (185)
```

With the fix restored:

```
node scripts/run-vitest.mjs src/daemon
Test Files  46 passed (46)
Tests  1500 passed | 10 skipped (1510)
```

`node scripts/check-changed.mjs` passes every lane except `plugin boundaries`, which fails on a clean checkout of `main` with zero changes:

```
compat deprecated=19 eligibleForRemoval=1 removalPending=8 removalPendingDue=0
  2026-09-08 sdk-untrusted-context-identifier-aliases owner=sdk codeRefs=6927 docRefs=62
```

That record came due 2026-09-08 and clearing it is a 6927-reference SDK compat removal needing owner judgment, so it is out of scope here and reported separately rather than bundled into this PR.

Follow-up commit `0fd6f4d991` corrects a second stale message found in review: the darwin not-listening branch of `gateway status` still printed `Errors: suppressed` (`src/cli/daemon-cli/status.print.ts`), which is precisely the path an operator reaches while investigating these startup failures. It now reports `Logs (stdout and stderr): <path>`, and the regression test asserts the word `suppressed` no longer appears. `src/commands/status-all/diagnosis.ts` needed no change — it already reads the stdout path on darwin, so it inherits the merged stream.

Re-verified after that commit: `node scripts/run-vitest.mjs src/daemon src/cli/daemon-cli` → `Test Files 74 passed (74)`, `Tests 2121 passed | 11 skipped (2132)`; `pnpm tsgo` exit 0.

Not live-verified on a fresh `openclaw gateway install`: the plist rewrite is covered at the unit boundary, and reinstalling the LaunchAgent on the operator's running gateway was not appropriate for this change.

### DIFF
--- a/docs/help/faq/logging-and-debugging.md
+++ b/docs/help/faq/logging-and-debugging.md
@@ -20,7 +20,7 @@ read_when:
 
     Service/supervisor logs (when the gateway runs via launchd/systemd):
 
-    - macOS launchd stdout: `~/Library/Logs/openclaw/gateway.log` (profiles use `gateway-<profile>.log`; stderr is suppressed).
+    - macOS launchd stdout and stderr: `~/Library/Logs/openclaw/gateway.log` (profiles use `gateway-<profile>.log`; both streams share this file, so startup failures that happen before the logger starts are recorded here too).
     - Linux: `journalctl --user -u openclaw-gateway[-<profile>].service -n 200 --no-pager`.
     - Windows: `schtasks /Query /TN "OpenClaw Gateway (<profile>)" /V /FO LIST`.
 
--- a/docs/platforms/mac/bundled-gateway.md
+++ b/docs/platforms/mac/bundled-gateway.md
@@ -163,7 +163,8 @@ Logging:
 
 - launchd stdout: `~/Library/Logs/openclaw/gateway.log` (profiles use
   `gateway-<profile>.log`)
-- launchd stderr: suppressed
+- launchd stderr: merged into the same `gateway.log` file, so startup failures
+  that happen before the logger starts are still recorded
 - If the host loops with repeated `EADDRINUSE` or fast restarts, check for
   duplicate `ai.openclaw.gateway` / `ai.openclaw.node` LaunchAgents and the
   launchd-marker workaround in
--- a/src/cli/daemon-cli/status.print.test.ts
+++ b/src/cli/daemon-cli/status.print.test.ts
@@ -624,7 +624,7 @@ describe("printDaemonStatus", () => {
     expectMockLineContains(runtime.error, formatCliCommand("openclaw gateway restart"));
   });
 
-  it("prints macOS launchd stdout and suppressed stderr when gateway is not listening", () => {
+  it("points macOS launchd stdout and stderr at one log when gateway is not listening", () => {
     const originalPlatform = process.platform;
     Object.defineProperty(process, "platform", { value: "darwin" });
     try {
@@ -669,7 +669,9 @@ describe("printDaemonStatus", () => {
 
     expectMockLineContains(runtime.error, "Gateway port 18789 is not listening");
     expectMockLineContains(runtime.error, "/Users/test/Library/Logs/openclaw/gateway.log");
-    expectMockLineContains(runtime.error, "Errors: suppressed");
+    expectMockLineContains(runtime.error, "Logs (stdout and stderr):");
+    const printed = runtime.error.mock.calls.map(([line]) => line).join("\n");
+    expect(printed).not.toContain("suppressed");
     const errors = runtime.error.mock.calls.map(([line]) => line).join("\n");
     expect(errors.match(/Last gateway error:/g)).toHaveLength(1);
   });
--- a/src/cli/daemon-cli/status.print.ts
+++ b/src/cli/daemon-cli/status.print.ts
@@ -526,8 +526,11 @@ export function printDaemonStatus(status: DaemonStatus, opts: { json: boolean; d
       );
     } else if (process.platform === "darwin") {
       const logs = resolveGatewaySupervisorLogPaths(serviceEnv, { platform: "darwin" });
-      defaultRuntime.error(`${errorText("Logs:")} ${shortenHomePath(logs.stdoutPath)}`);
-      defaultRuntime.error(`${errorText("Errors:")} suppressed`);
+      // The plist points both launchd handles at this file, so startup crashes that
+      // never reached the logger land here too; do not advertise a separate stderr.
+      defaultRuntime.error(
+        `${errorText("Logs (stdout and stderr):")} ${shortenHomePath(logs.stdoutPath)}`,
+      );
     }
     defaultRuntime.error(
       `${errorText("Restart log:")} ${shortenHomePath(resolveGatewayRestartLogPath(serviceEnv))}`,
--- a/src/daemon/diagnostics.test.ts
+++ b/src/daemon/diagnostics.test.ts
@@ -30,7 +30,7 @@ function makeTempStateDir(): string {
 }
 
 describe("readLastGatewayErrorLine", () => {
-  it("ignores stale launchd stderr when stderr is suppressed", async () => {
+  it("reads the launchd supervisor log instead of a state-dir stderr file on darwin", async () => {
     const stateDir = makeTempStateDir();
     const homeDir = makeTempStateDir();
     const env = { HOME: homeDir, OPENCLAW_STATE_DIR: stateDir };
--- a/src/daemon/diagnostics.ts
+++ b/src/daemon/diagnostics.ts
@@ -89,8 +89,9 @@ export async function readLastGatewayErrorLine(
 ): Promise<string | null> {
   const platform = options?.platform ?? process.platform;
   const readStderr = platform !== "darwin";
-  // launchd supervisor mode combines child stderr into stdout; other platforms
-  // keep stderr as the strongest failure signal.
+  // launchd combines child stderr into stdout only because the plist points both
+  // handles at one file (buildLaunchAgentPlist); break that and darwin startup
+  // crashes stop reaching this reader. Other platforms keep stderr separate.
   const { stdoutPath, stderrPath } =
     platform === "darwin"
       ? resolveGatewaySupervisorLogPaths(env, { platform })
--- a/src/daemon/launchd-service-files.ts
+++ b/src/daemon/launchd-service-files.ts
@@ -25,7 +25,6 @@ const LAUNCH_AGENT_PRIVATE_DIR_MODE = 0o700;
 export const LAUNCH_AGENT_ENV_FILE_MODE = 0o600;
 export const LAUNCH_AGENT_ENV_WRAPPER_MODE = 0o700;
 const LAUNCH_AGENT_ENV_DIR_NAME = "service-env";
-const LAUNCH_AGENT_STDERR_PATH = "/dev/null";
 export function resolveLaunchAgentPlistPathForLabel(
   env: Record<string, string | undefined>,
   label: string,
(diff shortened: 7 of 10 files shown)


RomneyDa:
Addressed in `0fd6f4d991`.

**[P2] `src/cli/daemon-cli/status.print.ts:545` — confirmed and fixed.** The finding is correct: the darwin not-listening branch still printed `Errors: suppressed`, which is exactly the path an operator lands on while investigating the startup failures this PR preserves. It now prints `Logs (stdout and stderr): <path>`, and the regression expectation in `status.print.test.ts` was updated to assert the word `suppressed` no longer appears rather than just swapping one literal for another.

Worth recording why it was missed: my original sweep for stale stderr messaging was `grep -rn "suppress" src/daemon src/cli/daemon-cli ... | head -6`, and the `head -6` truncated the results before the `src/cli/daemon-cli` hits. I caught the `runtime-hints.ts` copy and lost this one to the truncation. Re-swept without the limit; `runtime-hints.ts` and `status.print.ts` were the only two, and both are now correct.

`src/commands/status-all/diagnosis.ts` needed no change — it already gates on `readStderr = process.platform !== "darwin"` and reads the stdout path on darwin, so it inherits the merged stream automatically.

**Stored data model flag.** No migration applies: the plist is generated content, not user data. Existing installs keep their `/dev/null` plist until a definition-writing path runs — `rewriteLaunchAgentPlistForRestart` regenerates and reloads on any ordinary `openclaw gateway restart` (`src/daemon/launchd-lifecycle.ts:290-298`), and a fresh install writes it directly. `preserveDefinition` restarts skip the rewrite, but that is pre-existing behavior for every plist field rather than something this change introduces. Added to the PR body as an explicit upgrade-compatibility section.

**On the live-capture gap.** Correct that actual launchd capture after the fix is unverified, and I have not claimed otherwise. Reinstalling or restarting the LaunchAgent to capture it would rewrite the operator's running managed Gateway definition, which was not appropriate for this change; the plist writer is covered at the unit boundary for both the install and restart-rewrite paths.

Re-verified after the follow-up commit: `node scripts/run-vitest.mjs src/daemon src/cli/daemon-cli` → `Test Files 74 passed (74)`, `Tests 2121 passed | 11 skipped (2132)`. `pnpm tsgo` exits 0.

RomneyDa:
@clawsweeper re-review

The P2 finding (`src/cli/daemon-cli/status.print.ts` still printing `Errors: suppressed`) is fixed in `0fd6f4d991`, and the stored-data-model flag is answered with the plist-rewrite path in the PR body. Head has moved from `326ca4deaa` to `3bd47dd59f`.

RomneyDa:
Addressed in `adb56f197c`.

**[P3] `docs/help/faq.md:1288` — confirmed and fixed.** It read `stderr is suppressed`; it now documents that both streams share `gateway.log` and that startup failures which never reach the logger are recorded there.

That was the third instance of the same stale claim, and I want to be straight about why I kept missing them: both of my verification sweeps were piped through `head`, which truncated the results before reaching `src/cli/daemon-cli` the first time and `docs/` the second. Re-swept with no limit across `src` and `docs`. The only remaining matches are unrelated (`nodes-cli.coverage.test.ts` suppressing test output, `docs/cli/cron.md` on `NO_REPLY`).

The same commit also renames `src/daemon/diagnostics.test.ts` "ignores stale launchd stderr when stderr is suppressed" → "reads the launchd supervisor log instead of a state-dir stderr file on darwin". The assertion is unchanged and still correct — darwin ignores the state-dir stderr path, which is a different file from the launchd one — but the old name asserted the suppression this PR removes.

`node scripts/run-vitest.mjs src/daemon/diagnostics.test.ts` → 7 passed.

steipete:
Maintainer review of `a2616fa8dd86`: there is no schema or persistent-store change. Both plist writers reuse the existing supervisor log, preserving its retention policy; ordinary restart rewrites and reloads the generated definition. [Exact-head hosted CI passed](https://github.com/openclaw/openclaw/actions/runs/34438914271), including the install/restart boundary coverage. Actual after-fix launchd stderr capture remains a documented proof gap, accepted for this landing.

steipete:
Merged via squash.

- Prepared head SHA: [a2616fa8dd86c1b76fc0f4f83e61b6ef640b9c60](https://github.com/openclaw/openclaw/pull/143519/commits/a2616fa8dd86c1b76fc0f4f83e61b6ef640b9c60)
- Landed commit: [0b92bfeda7ea00e01cc09efb4bd21181a784a0c3](https://github.com/openclaw/openclaw/commit/0b92bfeda7ea00e01cc09efb4bd21181a784a0c3)

PR #143520, opened 2026-09-09
docs: split gateway/secrets into five child pages

vincentkoc:
Splits `docs/gateway/secrets.md` (53,408 chars, 24 H2 sections) into five child pages under `docs/gateway/secrets/`, keeping the parent as an index. Ledger finding `r3-0344`.

## Owned files in this PR

This PR touches files owned under `.github/CODEOWNERS`:

- `docs/gateway/secrets.md` — `@openclaw/openclaw-secops` (rule L57, file-exact)
- `docs/gateway/secrets/*.md` (5 new) — `@openclaw/openclaw-secops` via the rule this PR adds
- `.github/CODEOWNERS` — `@steipete` (rule L2)

### The split would have silently dropped secops ownership

The secops docs rules here are file-exact (`/docs/gateway/secrets.md`), and the
neighbouring directory rule is `/docs/gateway/security/` — `security`, not
`secrets`, and root-anchored. Nothing else matches `docs/gateway/secrets/**`.

Simulated last-match-wins over the unmodified `CODEOWNERS`, positive and
negative controls in the same run:

```
OWNED    docs/gateway/secrets.md                    <- L57 "/docs/gateway/secrets.md" @openclaw/openclaw-secops
OWNED    docs/gateway/security/threat-model.md      <- L58 "/docs/gateway/security/"  @openclaw/openclaw-secops
OWNED    .github/CODEOWNERS                         <- L2  "/.github/CODEOWNERS"      @steipete
OWNED    src/gateway/server-methods/secrets-list.ts <- L40 "/src/gateway/server-methods/secrets*.ts"
UNOWNED  docs/gateway/heartbeat.md                  <- no matching rule
UNOWNED  README.md                                  <- no matching rule
UNOWNED  docs/gateway/opentelemetry/setup.md        <- no matching rule
UNOWNED  docs/gateway/secrets/runtime-model.md      <- no matching rule   <-- the regression
```

So a naive split moves this content **out from under its ownership rule** — a
real governance regression, not a technicality. This PR therefore adds one line
beside the existing file-exact rule, inside the secops block (so the file's own
last-match-wins warning is satisfied — no later rule overrides it), matching the
directory-rule shape already used by `/docs/gateway/security/`:

```
 /docs/gateway/secrets.md @openclaw/openclaw-secops
+/docs/gateway/secrets/ @openclaw/openclaw-secops
 /docs/gateway/security/ @openclaw/openclaw-secops
```

Re-simulated against the edited file: all five children resolve to
`@openclaw/openclaw-secops`; `docs/gateway/heartbeat.md` and `README.md` still
resolve to no owner.

## Children

| Page | Sections | Size |
| --- | --- | --- |
| `secrets/runtime-model.md` | Runtime model → Onboarding reference preflight (6 H2) | 9,063 |
| `secrets/secretref-contract.md` | SecretRef contract, Provider config (2 H2) | 7,290 |
| `secrets/secret-store-and-egress.md` | Shared secret store → File-backed API keys (3 H2 + 1 H3) | 14,051 |
| `secrets/integration-examples.md` | Exec integration examples → Sandbox SSH auth material (3 H2) | 9,214 |
| `secrets/operations.md` | Supported credential surface → Control UI (9 H2) | 11,917 |

Index is 8,892 chars: original frontmatter and lede unchanged, a page table, the
anchor stub list, and the original `## Related` section unchanged.

Sibling precedent verified on `main` with `test -d`: `docs/gateway/doctor/`,
`opentelemetry/`, `config-tools/`, `config-agents/`, `cloud-workers/`,
`troubleshooting/`, `configuration/`. The index shape follows `docs/gateway/doctor.md`.

## Losslessness

The five children, frontmatter and ledes removed, concatenate to a byte-identical
copy of the original body:

```
ORIGINAL BODY  sha256: 33c4527c5bd76a24f6481026b69351c492b2f75a148029cc71e6ed5c37b6459c 51539
RECONSTRUCTED  sha256: 33c4527c5bd76a24f6481026b69351c492b2f75a148029cc71e6ed5c37b6459c 51539
BYTE-IDENTICAL: True
fences orig/recon: 15 15 | identical: True   (info string + body sha256, one-for-one)
words orig/recon: 6628 6628
words WHOLE TREE vs original: 7618 vs 6849 | grew: True
```

**No link rewrites were needed.** The page's only two intra-page links (both
`](#secret-egress-proxy)`, at original L289 and L294) sit in the same section
group as their target, so they stayed on one child. Byte-identity therefore holds
for the final tree, not just the raw extraction.

**The prose diff is empty.** `check-orphan-refs.py` found one directional
reference — `"the protocol limits below"` in the egress-proxy Warning (original
L384). Its referent (`bypassHosts` / blind CONNECT limits) was below it on the
same page before and is below it on the same page now, because the whole
`Secret egress proxy` section moved as one unit. Nothing was orphaned, so
nothing was reworded.

## Anchors

Enumerated with the repo's own `parseDocsDocument`, not a slug approximation.

- 48 pre-split ids; **all 48 resolve on the index**; 0 collisions on the index and on each child.
- 47 authored `<a id>` stubs. `related` is excluded because the index still publishes it (an authored stub would be a duplicate authored/canonical id).
- Two heading forms for one heading: `## Egress-time injection (sentinels)` mints **both** `egress-time-injection-(sentinels)` and `egress-time-injection-sentinels`. Both are stubbed. The parenthesised stub links to the *cleaned* child id, because an unbalanced `(` inside a markdown link URL does not parse — same treatment as `what-it-does-(summary)` on `docs/gateway/doctor.md`.
- Two component ids carry document-wide verbatim-marker indices that are recomputed per file. Stubs keep the pre-split id and target the child's new id:
  - `bitwarden-secrets-manager-openclawverbatim568end` → `integration-examples#bitwarden-secrets-manager-openclawverbatim229end`
  - `password-store-openclawverbatim579end` → `integration-examples#password-store-openclawverbatim240end`

  These indices shift whenever anything above them changes, so they were never stable anchors. Any inbound link using one is fragile and worth treating as such.
- `python3 .audit/stranded-stubs.py docs` → **0 suspects**. The page had no pre-existing authored `<a id>` stubs, so nothing could be swept into the wrong child.

## Pins

`docs-test-pins.py` reports 11 "tests"; the label is loose. Classified individually:

- `test/scripts/ci-node-test-plan.test.ts:104` — `STORE_ALIAS_CHANGED_PATHS`, a **synthetic `changedPaths` fixture**, not a path allowlist. It asserts a docs path produces no test shards (`toBeNull()`). The index still exists at the same path and children behave identically, so no gap. Test run: 54 passed.
- `taxonomy.yaml` ×9, `qa/scenarios/*.yaml` ×3 (`docsRefs`), `AGENTS.md`, `extensions/visitor-access/README.md`, `src/config/schema.hints.ts` — plain file-path or route references with **no `#anchor`**. The parent still exists; all remain valid.
- `ui/src/pages/secrets/view.ts:56` — `…/gateway/secrets#shared-secret-store`. This is a real anchor pin. It still resolves, via the stub on the index, and now costs one extra hop to reach the content. Left as-is rather than editing product code in a docs split; flagged below as follow-up.
- `.github/CODEOWNERS` — handled above.

`scripts/github/security-sensitive-guard.mjs` also carries a hard-coded path list
(`securitySensitiveFiles`), checked because it matches on exact paths — it contains
only `.gitignore`, so this page was never in it and no gap is created.

## Labeler

No change needed, and no gap created: `.github/labeler.yml` has **no file-exact glob**
for this page. The `gateway` label matches via recursive `docs/gateway/**`, which
covers `docs/gateway/secrets/**` too. Labeler tests run: 2 files, 15 passed.

Pre-existing and **not fixed here**: the `security` label globs are
`docs/cli/security.md` and `docs/gateway/security/**` — `docs/gateway/secrets.md`
was never matched by it. Reporting only, per program rules.

## Glossary

`check-docs-i18n-glossary.mts` required exactly the five new child titles. Added
as five sources, composed from vocabulary already in the file (密钥 / 运行时 /
契约 / 提供商配置 / 集成 / 存储) — no coined terms. Inserted **beside the parent's
`Secrets management` entry (index 12), not appended at the end**, so concurrent
splits touch different regions of the file and merge cleanly. Diff is 20
insertions, 0 deletions, no reformatting; 0 duplicate sources.

## Validation

Baseline measured in a detached worktree at the merge-base `df76050`, not at a
moving `origin/main`, with `node_modules` symlinked in.

- `pnpm check:docs` (the `check-docs` job's own aggregate) — **exit 0**. Covers format check (1,256 files), `markdownlint-cli2` with the config's own globs (1,255 files incl. `.mdx`), `lint:templates`, MDX check, i18n glossary, link check, config examples.
- `docs-link-audit.mjs --anchors` — 3 broken, **identical to baseline** (the known pre-existing `maturity/#windows-app-node` ones). Internal links checked grew 13,134 → 13,191. No new breakage.
- `vitest.full-core-unit-src.config.ts src/docs/` — 8 files, 16 passed.
- `vitest.tooling.config.ts src/scripts/docs-link-audit.test.ts` — 1 file, 33 tests, **1 failed**: a slugify expectation (`ass-guide` vs `as-s-guide`) unrelated to this change. **Reproduced identically at base sha `df76050`** (same 1 failed / 32 passed) — pre-existing, not caused here.

## Follow-ups (not in this PR)

- `ui/src/pages/secrets/view.ts:56` could point directly at
  `/gateway/secrets/secret-store-and-egress#shared-secret-store` to skip the index hop.
- The `security` labeler label does not match this page (pre-existing).
- Ledger link findings `r3-1509`, `r3-2250`, `r3-2270`, `r5-0110` concern *other*
  pages linking back to this one; out of scope for a content-preserving split.


vincentkoc:
Merging on the org admin's direct approval (@vincentkoc), with no `@openclaw/openclaw-secops` review recorded — and flagging that this PR also touches `.github/CODEOWNERS`, which `@steipete` owns.

**The reason it touches CODEOWNERS is the point of the PR.** The secops docs rules here are file-exact (`/docs/gateway/secrets.md`), and `/docs/gateway/security/` is *security*, not *secrets* — so nothing would have matched `docs/gateway/secrets/**`. A split with no CODEOWNERS change would have silently dropped secops ownership of this content while looking like pure reorganisation. The added `/docs/gateway/secrets/ @openclaw/openclaw-secops` line sits beside the file-exact rule inside the secops block, and re-simulation against the edited file confirms all five children are owned with the negative controls still unowned.

Findings: None. Byte-identical reassembly of the five children (`33c4527c…`, 51,539 chars), 15 fences matched one-for-one, all 48 pre-split ids resolving with 0 collisions, empty prose diff, `pnpm check:docs` exit 0, CI 88 pass / 0 fail.

Two things from the author worth recording. It found and fixed a stub whose parenthesised link target had an unbalanced `(` — it would not parse as a markdown URL, so the anchor would have pointed at nothing; `docs/gateway/doctor.md` already solves this by targeting the cleaned id, and it matched that. And `ui/src/pages/secrets/view.ts:56` links `#shared-secret-store`, which still resolves through the index stub at one extra hop — left alone rather than editing product code inside a docs split, and worth a follow-up.

PR #143531, opened 2026-09-09
fix(sqlite): name the repair path in schema drift failures

RomneyDa:
## What Problem This Solves

A missing or drifted canonical SQLite index aborts Gateway startup, and the failure text told operators nothing about how to fix it:

```
SQLite schema is incomplete or noncanonical for
/Users/…/.openclaw/agents/codex/agent/openclaw-agent.sqlite:
missing or drifted index idx_agent_session_nodes_active
```

That is the entire message. The Gateway then exits `78 (EX_CONFIG)` and launchd restarts it into a crash loop.

Hit on a live install today. The affected database was stamped `user_version = 19` (current), passed `pragma quick_check`, and had every other canonical index — exactly one index was absent, on a *secondary* agent that was not even in use. `openclaw doctor` ran to completion and reported "Doctor complete" without mentioning it. The repair turned out to be a single idempotent `CREATE INDEX IF NOT EXISTS`, and `openclaw doctor --fix` applied it in milliseconds ("Rebuilt canonical shared-state SQLite indexes (1)") — but nothing in the failure path said so.

The two sibling throws in `assertOpenClawAgentCurrentRuntimeSchema` (`src/state/openclaw-agent-db-schema-helpers.ts`) already end with *"run openclaw doctor --fix before using it."* The index-drift throw, which is the one that actually fires on a healthy-looking database, was the only one that dead-ended.

## Why This Change Was Made

`throwSqliteSchemaMismatches` is the single owner of this message, so naming the repair path there fixes both call sites in `sqlite-schema-contract.ts` at once and keeps the wording consistent with its siblings.

This intentionally does **not** make runtime open self-repair the index. That would be a material change to recovery semantics: both the shared state database (`repairStateSchema`) and the agent migration path (`repairAndAssertAgentSchemaGroup`) call `repairCanonicalSqliteIndexes` only inside explicit repair/migration contexts, never on a normal runtime open. Per `AGENTS.md`, same-version changes to indexing or recovery need maintainer acceptance before implementation, so that option is raised for a maintainer decision rather than taken here. See "Open question" below.

## User Impact

An operator whose Gateway will not start now gets a message that names the fix instead of only the symptom. No schema change, no persistence-semantics change, no config surface.

## Evidence

Production change is one message string plus a two-line comment.

Regression test `names the doctor repair path when a canonical index is missing` drops a canonical index and asserts both the drift detail and the remediation. Verified it fails on pre-fix code by restoring `src/infra/sqlite-schema-issues.ts` from `HEAD`:

```
× assertSqliteSchemaContains (statement cache: false) > names the doctor repair path when a canonical index is missing
× assertSqliteSchemaContains (statement cache: true)  > names the doctor repair path when a canonical index is missing
Tests  2 failed | 95 skipped (97)
```

With the fix restored:

```
node scripts/run-vitest.mjs src/infra/sqlite-schema-contract.test.ts src/infra/sqlite-index-schema.test.ts
Test Files  2 passed (2)      Tests  110 passed | 2 skipped (112)

node scripts/run-vitest.mjs src/state
Test Files  65 passed | 1 skipped (66)      Tests  956 passed | 6 skipped (962)
```

`pnpm tsgo` exits 0. `node scripts/check-changed.mjs` passes every lane except `plugin boundaries`, which fails identically on a clean checkout of `main` with zero changes (`2026-09-08 sdk-untrusted-context-identifier-aliases`, 6927 code refs) and is reported separately.

## Open question for a maintainer

Should a canonical **index** — as opposed to a table or column — be repairable during normal runtime open rather than only under `doctor --fix`? An index is a performance artifact; refusing to boot on a missing one, when the repair is idempotent and already implemented, is arguably disproportionate. It is deliberately left out of this PR because it changes recovery semantics and needs your call.

### DIFF
--- a/src/infra/sqlite-schema-contract.test.ts
+++ b/src/infra/sqlite-schema-contract.test.ts
@@ -77,6 +77,21 @@ describe.each([false, true])("assertSqliteSchemaContains (statement cache: %s)",
     }
   });
 
+  it("names the doctor repair path when a canonical index is missing", () => {
+    const database = createDatabase(CANONICAL_SCHEMA);
+    try {
+      database.exec("DROP INDEX idx_children_parent;");
+
+      // Operators hit this throw as gateway startup failure text, so it must
+      // name the repair owner instead of dead-ending on the drift detail.
+      expect(() => assertSqliteSchemaContains(database, "test database", CANONICAL_SCHEMA)).toThrow(
+        /missing or drifted index idx_children_parent; run openclaw doctor --fix to repair it\./,
+      );
+    } finally {
+      database.close();
+    }
+  });
+
   it("accepts an extra non-unique index on a canonical table", () => {
     const database = createDatabase(CANONICAL_SCHEMA);
     try {
--- a/src/infra/sqlite-schema-issues.ts
+++ b/src/infra/sqlite-schema-issues.ts
@@ -121,7 +121,9 @@ export function throwSqliteSchemaMismatches(
   if (mismatches.length > shown.length) {
     shown.push(`${mismatches.length - shown.length} additional mismatch(es)`);
   }
+  // Drift is repairable by the doctor migration owner, so the throw must name it:
+  // callers surface this straight to operators, and the gateway refuses to start.
   throw new Error(
-    `SQLite schema is incomplete or noncanonical for ${databaseLabel}: ${shown.join("; ")}`,
+    `SQLite schema is incomplete or noncanonical for ${databaseLabel}: ${shown.join("; ")}; run openclaw doctor --fix to repair it.`,
   );
 }


RomneyDa:
Thanks — noting the review records no contributor-facing blocker and endorses keeping the explicit Doctor repair boundary rather than repairing on runtime open. That boundary is deliberate and is called out as an open question in the PR body: both `repairStateSchema` and `repairAndAssertAgentSchemaGroup` call `repairCanonicalSqliteIndexes` only inside explicit repair/migration contexts, so self-healing a canonical index during a normal runtime open would be a new recovery semantic and needs a maintainer decision rather than an agent one.

On the proof gap: correct that the before/after regression results and the live incident narrative are not an observed after-fix Gateway diagnostic run, and the body does not claim otherwise. I am adding an observed run of the new failure text against an isolated state directory and will post it here.

RomneyDa:
Observed proof, as promised — this closes the "no observed after-fix diagnostic run" gap.

Rather than a synthetic fixture, I ran the real production assertion (`assertOpenClawAgentCurrentRuntimeSchema`) against a **copy of a live agent database** from this machine — the same `codex` agent database whose missing `idx_agent_session_nodes_active` crash-looped the Gateway and prompted this PR. The index was dropped on the copy only; the original was verified untouched before and after (`select count(*) ... where name='idx_agent_session_nodes_active'` → `1`).

**Before (`origin/main`):**

```
SQLite schema is incomplete or noncanonical for …/agent-copy.sqlite: missing or drifted index idx_agent_session_nodes_active
```

**After (this PR):**

```
SQLite schema is incomplete or noncanonical for …/agent-copy.sqlite: missing or drifted index idx_agent_session_nodes_active; run openclaw doctor --fix to repair it.
```

Same database, same code path, same run — only `src/infra/sqlite-schema-issues.ts` swapped between the two invocations. This is the exact text an operator sees on the failure that takes the Gateway down with `EX_CONFIG`, and it now names the command that fixes it. For what it is worth, `openclaw doctor --fix` did repair precisely this on the live install ("Rebuilt canonical shared-state SQLite indexes (1)").

Scope note: this observes the failure text, not a full Gateway boot. Reproducing the crash loop end-to-end would mean corrupting a real agent database under a running managed Gateway, which was not appropriate here.

steipete:
Maintainer review of `e92f72296866`: this changes the schema-drift diagnostic and its regression test only—no schema, stored data, migration, or recovery semantics change. The existing Doctor repair owner is preserved. [Exact-head hosted CI passed](https://github.com/openclaw/openclaw/actions/runs/34419292533), and the missing-index regression covers statement caching both enabled and disabled. The generic data-model compatibility checklist does not require a migration for this patch.

steipete:
Merged via squash.

- Prepared head SHA: [e92f722968663b85b91e30aecc2f1b6286e5c024](https://github.com/openclaw/openclaw/pull/143531/commits/e92f722968663b85b91e30aecc2f1b6286e5c024)
- Landed commit: [d9d12d462590258ad55a3334985cc71f06e3ae18](https://github.com/openclaw/openclaw/commit/d9d12d462590258ad55a3334985cc71f06e3ae18)

PR #143534, opened 2026-09-10
refactor(plugin-sdk): compact reply payload case matrices

RomneyDa:
## What Problem This Solves

The Plugin SDK reply-payload suite repeats object property names across homogeneous helper matrices, obscuring the actual outbound edge cases.

## Why This Change Was Made

Compacts four matrices into positional tuples, retaining explicit optional-media typing and preserving every reasoning, media, text, and rich-content case.

## User Impact

No user-visible behavior changes. Maintainers get unchanged outbound payload coverage with 63 fewer test lines.

## Evidence

- `node scripts/run-vitest.mjs src/plugin-sdk/reply-payload.test.ts` — 80 passed
- `node scripts/check-changed.mjs` — passed, including Plugin SDK surface and native services graphs
- P3 autoreview — clean, 0.99

### DIFF
--- a/src/plugin-sdk/reply-payload.test.ts
+++ b/src/plugin-sdk/reply-payload.test.ts
@@ -56,37 +56,17 @@ describe("ask_user question option indices", () => {
 
 describe("isReasoningReplyPayload", () => {
   it.each([
-    { name: "flagged", payload: { text: "Visible", isReasoning: true }, expected: true },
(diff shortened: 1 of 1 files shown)


RomneyDa:
@clawsweeper review

RomneyDa:
@clawsweeper re-review

PR #143537, opened 2026-09-10
refactor(tests): share Discord REST runtime spies

steipete:
Related: #139428

## What Problem This Solves

Eleven Discord REST proxy tests repeat the same three runtime spies, adding boilerplate around their proxy and logging assertions.

## Why This Change Was Made

Create fresh log, error, and exit spies through one private local helper at the original allocation sites. The helper preserves the sparse runtime shape and nonthrowing exit behavior; every proxy scenario and independent assertion remains unchanged.

## User Impact

No user-visible behavior changes. This removes 40 net test lines without removing test cases or assertions.

## Evidence

The full REST proxy suite and proxy request-client sibling passed before and after: 16 tests, zero skips, with identical ordered names and results. Inlining all 11 helper calls reconstructs the original file's parser tokens, accounting only for formatter-removed trailing commas. Separate controls confirm fresh spies, independent call histories, the same three own keys, and unchanged exit behavior. The mapped changed-file gate and fresh independent P2 review passed. No full-suite runtime improvement is claimed.

### DIFF
--- a/extensions/discord/src/monitor/provider.rest-proxy.test.ts
+++ b/extensions/discord/src/monitor/provider.rest-proxy.test.ts
@@ -123,6 +123,10 @@ function dispatchRequest(dispatcher: unknown, origin: string): void {
   target.dispatch({ origin, path: "/", method: "GET" }, {});
 }
 
+function createRuntimeSpies() {
+  return { log: vi.fn(), error: vi.fn(), exit: vi.fn() } as const;
+}
+
 function installUndiciRuntimeDeps(): void {
   const runtime = createMockUndiciRuntime();
   class Pool {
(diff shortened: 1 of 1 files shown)


steipete:
Merged as f98cd2d29b5a75b0c15d4580eb729e01a2e97485. The 11 REST proxy scenarios retain their assertions and fresh three-spy runtime shape; 40 net test lines were removed.

The complete target and proxy-client sibling passed before and after (16 tests, zero skips), and exact-head required CI passed. Native review/prepare completed. Native merge stopped before dispatch solely because the advisory ClawSweeper verdict was not complete; the authorized maintainer exception used an ordinary non-admin squash pinned to the reviewed head. No enforced check was bypassed.

Merged source matches the reviewed candidate. The PR worktree and remote branch are absent, and the task checkout is clean on main.

Issue #115008, opened 2026-07-28
[Bug]: Schema-version error blames a downgrade that never happened; version string cannot identify the stale build


steipete:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

`SqliteSchemaVersionError` tells the user "Do not downgrade OpenClaw" and "Upgrade OpenClaw" when no downgrade happened and the version string cannot distinguish the two builds involved — on one machine three different state-schema levels (4, 5, 6) were all produced by builds that identify themselves as `2026.7.2`, leaving the error's prescribed action unactionable.

### Steps to reproduce

1. Have two OpenClaw builds from different `main` commits that share a version string (here both print `2026.7.2`): one whose bundled dist supports state schema 4, and a newer one that writes schema 5/6.
2. Let the newer build create/upgrade a state database (`~/.openclaw/state/openclaw.sqlite` reached `user_version=6`; a second state dir `~/.openclaw-clawmac-node` reached `user_version=5`).
3. Run the older build against either database, e.g. `OPENCLAW_STATE_DIR=~/.openclaw-clawmac-node openclaw doctor --fix`.
4. Observe the error below, then run `openclaw --version` on the failing build and compare it to the version recorded in `schema_meta` of the database it refused to open.

### Expected behavior

The error should identify *which* install is stale in a way the user can act on. The build already knows its own commit — `openclaw --version` prints `2026.7.2 (6b80b01)` — and `createNewerSqliteSchemaVersionError` already has both schema numbers. The message should state what is actually known (this build, its commit/install path, the schema it supports vs. the schema found) instead of asserting a downgrade that did not occur.

### Actual behavior

The message asserts a user action that never happened and gives no way to locate the stale install:

- "Do not downgrade OpenClaw or modify the database" — nothing was downgraded; two builds that both call themselves `2026.7.2` simply diverged.
- "Upgrade OpenClaw before opening this database" — unactionable, because the failing build reports the *same* version string (`2026.7.2`) as the build that wrote the newer database. `schema_meta` in the refused database also records the writer as `2026.7.2`.
- "To run this older build" — the build is only "older" by commit, which the error never prints, even though the CLI already surfaces it in `--version`.

[...]

PR #143538, opened 2026-09-10
fix(update): report the built runtime instead of the unbuilt source version

RomneyDa:
## What Problem This Solves

A git-mode install that pulls but never rebuilds keeps executing the previous `dist/`. Every surface an operator would check to notice that reported the opposite.

`openclaw --version` resolved the version from the **source** `package.json` while the commit came from the **build**:

```
resolveVersionFromModuleUrl = readVersionFromPackageJson(...) || readVersionFromBuildInfo(...)
```

Observed on a live install: the machine printed `OpenClaw 2026.9.2 (1623683)` while actually running a build stamped `{"version":"2026.8.1","commit":"1623683f…"}`. The version half came from source, the commit half from the build, and `2026.9.2` had never been compiled. `openclaw status` reported `git main @ <HEAD>` with no indication that `dist/` was six days behind it.

This is not a cosmetic mismatch. The repo already detects the exact condition — `collectGitRuntimeErrors` compares `build-info.commit`, `.buildstamp.head` and `.runtime-postbuildstamp.head` against the expected SHA, and `verifyGitUpdateRecovery` returns `runtime-verification-failed` — but only inside the update path. When an update aborts at its dirty-tree precondition it writes that verdict to a support file and stops; nothing tells the operator, and the version string then actively contradicts it. On the install that prompted this, that combination hid a stale runtime for six days until a plugin failed to load against a module its build predated.

## Why This Change Was Made

Two minimal changes, both reusing what already exists:

**Version precedence.** `build-info.json` records the version the running artifact was actually built from, so it now wins. Both readers were already present and exported; only the `||` order changed. Source runs (`tsx`, `pnpm dev`) have no reachable `build-info.json` and still fall back to `package.json`, and published packages ship `dist/build-info.json` alongside a matching `package.json`, so neither case changes. No test pinned the old precedence — the existing one covers "falls back to build-info when package metadata is unavailable", which still holds.

**Launcher precedence.** The package executable `openclaw.mjs` answers bare `--version`/`-V`/`-v` and exits *before* the runtime entry loads, so it never reaches the resolver above. Its `resolveLauncherVersion()` read `package.json` first while `resolveLauncherCommit()` already preferred `dist/build-info.json` — which is exactly how a stale build printed a source version paired with a built commit (`OpenClaw 2026.9.2 (1623683)`). Both halves now agree.

**Status visibility.** `GitUpdateStatus` already carried `root` and `sha`, so detecting staleness needed only a built-commit read (`readBuiltRuntimeCommit`, added next to the existing `readBuiltGatewayBuildId`) and a comparison.

The status check deliberately compares the built commit to HEAD rather than reusing `collectGitRuntimeErrors` wholesale. That helper also fails on a missing `dist/entry.js` and on Control UI asset health, which would flag every developer running from source as "stale". Comparing commits reports only the real condition, and stays silent when no build exists.

## The version string is not only cosmetic

`VERSION` is derived from this resolver and is stamped into persisted state and compared at two decision points, so the wrong value had consequences beyond display:

- `src/state/openclaw-state-db-fast-path.ts:86` returns `metadata?.app_version === VERSION` to decide whether startup repairs can be skipped ("app_version commits only after this release's repairs; same-build writes are canonical"). Under the old precedence a dist built at `2026.8.1`, sitting in a checkout whose `package.json` read `2026.9.2`, reported `2026.9.2` — so against a database already stamped `2026.9.2` by a genuine build, the stale runtime **matched and skipped startup repairs it had never run**. Reporting the built version makes that comparison fail, which falls through to the full repair path; a false negative here only does more work.
- `src/commands/doctor-gateway-health.ts:68` compares the gateway version against the CLI's `VERSION`; a stale CLI now correctly reports the mismatch instead of hiding it.
- `src/infra/sqlite-user-version.ts:39` builds the diagnostic identity `OpenClaw ${VERSION} (${commit})`. That is the exact string #115008 reported as unable to identify the stale build; it is now truthful.

Nothing gates a migration on the version *string* — the schema decisions key off `user_version` — so no migration behavior changes.

## User Impact

`openclaw --version` now reports the version actually executing. `openclaw status` gains a `stale build (running <commit>, run pnpm build)` marker on the update row when `dist/` was built from a different commit than HEAD. Nothing changes for npm installs, source runs, or checkouts whose build is current. No config surface, no new files.

## Evidence

```
node scripts/run-vitest.mjs src/version.test.ts                  Tests  17 passed (17)
node scripts/run-vitest.mjs src/commands/status.update.test.ts   Tests  17 passed (17)
node scripts/run-vitest.mjs src/commands/status.update-channel.test.ts \
  src/commands/status.update-ledger.test.ts src/infra/update-check.test.ts
                                                                 Tests  30 passed (30)
pnpm tsgo                                                        exit 0
```

New tests:

- `reports the built version when dist lags the source package version` — pins the real-world shape (source `2026.9.2`, build `2026.8.1`) and asserts the built version wins while `readVersionFromPackageJsonForModuleUrl` still sees the source value.
- `reports a stale build when dist was built from a different commit`
- `stays quiet when the built commit matches HEAD` — guards against the false-positive mode described above.
- `test/openclaw-launcher-version.e2e.test.ts` gains `reports the built version when the source package version moved ahead`, with deliberately differing package and build versions. Verified failing on pre-fix code: `expected 'OpenClaw 1.2.3-test (1234567)' to be 'OpenClaw 2026.8.1 (1234567)'`.
- `src/infra/update-git-runtime.test.ts` covers `readBuiltRuntimeCommit`, pinning the null results (no `dist/`, provenance without a commit) that keep source-only checkouts from being reported stale.

`node scripts/check-changed.mjs` passes every lane except `plugin boundaries`, which fails identically on a clean checkout of `main` with zero changes (`2026-09-08 sdk-untrusted-context-identifier-aliases`, 6927 code refs) and is reported separately.

## Follow-up not taken here

`openclaw doctor` still does not run the stale-runtime check at all; this PR only reaches `status` and `--version`. Wiring `collectGitRuntimeErrors` into a doctor check (with the source-run false positives handled) is the natural next step and is left for a separate change.

### DIFF
--- a/openclaw.mjs
+++ b/openclaw.mjs
@@ -617,15 +617,18 @@ function readLauncherJson(relativePath) {
 }
 
 function resolveLauncherVersion() {
-  const packageJson = readLauncherJson("./package.json");
-  const packageVersion = normalizeLauncherMetadataValue(packageJson?.version);
-  if (packageVersion) {
-    return packageVersion;
-  }
+  // Report what is built, not what the source says: resolveLauncherCommit already
+  // prefers dist provenance, so reading package.json first pairs a source version
+  // with a built commit and hides a checkout that pulled without rebuilding.
   const buildInfo = readLauncherJson("./dist/build-info.json");
   const buildVersion = normalizeLauncherMetadataValue(buildInfo?.version);
   if (buildVersion) {
     return buildVersion;
+  }
+  const packageJson = readLauncherJson("./package.json");
+  const packageVersion = normalizeLauncherMetadataValue(packageJson?.version);
+  if (packageVersion) {
+    return packageVersion;
   }
   return normalizeLauncherMetadataValue(process.env.OPENCLAW_BUNDLED_VERSION) ?? "0.0.0";
 }
--- a/src/commands/status.update.test.ts
+++ b/src/commands/status.update.test.ts
@@ -55,6 +55,50 @@ describe("resolveUpdateAvailability", () => {
     });
   });
 
+  it("reports a stale build when dist was built from a different commit", () => {
+    const update = buildUpdate({
+      installKind: "git",
+      git: {
+        root: "/tmp/repo",
+        sha: "abc123456789",
+        tag: null,
+        branch: "main",
+        upstream: "origin/main",
+        dirty: false,
+        ahead: 0,
+        behind: 0,
+        fetchOk: true,
+        builtSha: "def987654321",
+      },
+    });
+
+    // Pulling without rebuilding keeps the old dist running, which is invisible
+    // from HEAD alone and is exactly what a failed update leaves behind.
+    expect(formatUpdateOneLiner(update)).toContain(
+      "stale build (running def98765, run pnpm build)",
+    );
+  });
+
+  it("stays quiet when the built commit matches HEAD", () => {
+    const update = buildUpdate({
+      installKind: "git",
+      git: {
+        root: "/tmp/repo",
+        sha: "abc123456789",
+        tag: null,
+        branch: "main",
+        upstream: "origin/main",
+        dirty: false,
+        ahead: 0,
+        behind: 0,
+        fetchOk: true,
+        builtSha: "abc123456789",
+      },
+    });
+
+    expect(formatUpdateOneLiner(update)).not.toContain("stale build");
+  });
+
   it("flags registry update when latest version is newer", () => {
     const latestVersion = nextMajorVersion(VERSION);
     const update = buildUpdate({
--- a/src/commands/status.update.ts
+++ b/src/commands/status.update.ts
@@ -199,6 +199,11 @@ export function formatUpdateOneLiner(update: UpdateCheckResult): string {
     if (update.git.fetchOk === false) {
       parts.push("fetch failed");
     }
+    // A checkout that pulled but never rebuilt keeps executing the previous dist,
+    // so report the built commit rather than letting HEAD imply what is running.
+    if (update.git.builtSha && update.git.sha && update.git.builtSha !== update.git.sha) {
+      parts.push(`stale build (running ${update.git.builtSha.slice(0, 8)}, run pnpm build)`);
+    }
     appendRegistryUpdateSummary();
   } else {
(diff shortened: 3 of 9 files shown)


RomneyDa:
Addressed in `853d0c0258`.

**[P2] launcher fast path — confirmed, and this was a real miss on my part, not a false positive.** `openclaw.mjs` answers bare `--version`/`-V`/`-v` and exits before the runtime entry loads, so changing `src/version.ts` alone left the packaged CLI still reporting the source version. `resolveLauncherVersion()` now prefers `dist/build-info.json`, matching `resolveLauncherCommit()`, which already did.

This also explains the original symptom precisely: the commit half already came from dist while the version half came from `package.json`, which is how the reporting install printed `OpenClaw 2026.9.2 (1623683)` — a source version paired with a build stamped `2026.8.1`. The two halves disagreed by construction.

`test/openclaw-launcher-version.e2e.test.ts` gains `reports the built version when the source package version moved ahead` with deliberately differing package and build versions. Verified it fails on pre-fix code for the intended reason:

```
AssertionError: expected 'OpenClaw 1.2.3-test (1234567)' to be 'OpenClaw 2026.8.1 (1234567)'
Tests  1 failed | 14 skipped (15)
```

With the fix: `Test Files 1 passed (1)`, `Tests 15 passed (15)`. `pnpm tsgo` exits 0.

Separately, the PR body now records that this resolver is not display-only: `VERSION` feeds `openclaw-state-db-fast-path.ts:86`, where a stale build previously matched `schema_meta.app_version` and **skipped startup repairs it had never run**.

steipete:
The remaining review gap is plugin compatibility proof: `VERSION` also drives external-plugin admission, so preferring built metadata can change which plugins load on a stale checkout. Please link a hosted CI run and identify the tests demonstrating (1) admission with matched source/build metadata, (2) rejection when the actual stale build is below the plugin's declared host/API floor, and (3) successful admission after rebuilding to a compatible version. The current launcher/resolver/status tests do not demonstrate that admission-and-recovery sequence.

PR #143542, opened 2026-09-10
fix: retain model setup resources through asynchronous work

steipete:
## What Problem This Solves

Resolves a problem where provider authentication setup can leave asynchronous work using a prepared runtime after model preparation fails or its caller releases the result. With a managed plugin registration, that can close its database before the remaining work finishes.

## Why This Change Was Made

Model acquisition now owns setup and its admitted descendants from the moment it acquires the runtime. It releases the runtime only after setup finishes and any successful caller releases its claim. SDK host work covers setup rather than the returned lease's lifetime, preserving shutdown order without making host close wait on itself.

## User Impact

Provider setup work keeps access to its runtime through completion or failure. Model selection, authentication policy, public SDK contracts, and the current managed-runtime activation rules remain unchanged.

## Evidence

- Original-source regression: three managed SQLite cases fail because disposal occurs before a held authentication descendant reads its database. All now pass; the SDK shutdown success control remains green. These tests use an input adapter returning a real already-managed runtime, without enabling managed RUN acquisition.
- 440 focused and sibling tests pass, including public SDK, model execution, capability CLI, approval classification, and exec review controls.
- Changed-scope guards, types, dead-code scans, lint, and import-cycle checks pass. Fresh independent Codex review found no actionable P0–P2 findings.
- Full build passes in 245 seconds. A plain Node child exercises the compiled public SDK with a cold provider, held authentication HTTP work, and four real loopback model completions after setup drainage. It verifies generation reuse, host-close ordering, and eventual fresh acquisition; the child exits in 1.77 seconds with source and 12,129 build artifacts unchanged.
- Compiled proof keeps current raw RUN acquisition unchanged; it does not claim managed registration disposal. The build emitted a separate 21-byte UI startup gzip budget warning without failing.

### DIFF
--- a/src/agents/simple-completion-runtime.ts
+++ b/src/agents/simple-completion-runtime.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from "node:async_hooks";
 import { prepareModelForSimpleCompletion } from "@openclaw/ai/transports";
 /**
  * Simple completion runtime preparation.
@@ -16,6 +17,12 @@ import {
 } from "../plugins/provider-hook-runtime.js";
 import { prepareProviderRuntimeAuth } from "../plugins/provider-runtime.runtime.js";
 import { withPluginRuntimeGenerationScope } from "../plugins/runtime/generation-scope.js";
+import {
+  AsyncWorkScope,
+  captureAsyncWorkTracker,
+  getAsyncWorkSignal,
+} from "../shared/async-work-scope.js";
+import { createDeferredCore } from "../shared/deferred.js";
 import {
   resolveAgentDir,
   resolveAgentEffectiveModelPrimary,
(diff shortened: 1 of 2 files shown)


steipete:
@clawsweeper re-review

The review checkout metadata fix has landed in https://github.com/openclaw/clawsweeper/pull/1519. Please review the current head using the updated preparation path; the source and validation evidence in this PR are current.

steipete:
Merged via squash.

- Prepared head SHA: [f8f4ef145eaabaea512b59114f170f1e98726981](https://github.com/openclaw/openclaw/pull/143542/commits/f8f4ef145eaabaea512b59114f170f1e98726981)
- Landed commit: [b158c5214f2a1eef5cd59aa8d1bd35b163e3dc4f](https://github.com/openclaw/openclaw/commit/b158c5214f2a1eef5cd59aa8d1bd35b163e3dc4f)

PR #143557, opened 2026-09-10
fix(cli): preserve update diagnostics when history is unavailable

steipete:
## What Problem This Solves

Fixes an issue where users running `openclaw update status` could lose runtime repair findings when SQLite was rejected by the runtime guard or saved update history was unreadable. An eager schema fingerprint calculation could fail during module loading, before the command could report those findings.

## Why This Change Was Made

Retired-schema fingerprints are now derived when schema recognition needs them, using the same canonical SQL, cache and compatibility checks. The CLI treats reading and classifying run history as one fallible diagnostic section. If that section fails, it reports `runStatusError` and omits partial run fields while retaining update availability, runtime findings and the final update hint.

SQLite safety checks, accepted schema layouts, migration decisions and stored records are unchanged. An unavailable history report is distinct from an empty history, and the status command does not repair state.

## User Impact

Users can still see useful runtime and update guidance when optional run history is unavailable. Human output labels the failure; JSON includes `runStatusError` rather than presenting an incomplete active/latest-run snapshot.

## Evidence

The final regression fixture has six intended failures before the repair and 19 passing cases afterward. It covers cold imports, missing and healthy state, unsafe-reader rejection, and a malformed latest row after a valid active-row read. Across the final status suite, SQLite/update siblings and selected retirement controls, 725 distinct cases pass in 11 files.

Tests use real config and ledger readers with isolated state and a healthy SQLite engine whose version query is intercepted. Runtime capability facts and external observations are synthetic. This proves the guarded reader and command boundary without operating an unsafe runtime or live updater.

The 29-stage changed check passed before a final test-only receiver annotation; production and documentation bytes are identical. The final annotation passed lint, affected test types, formatting and all 19 status cases. Independent review found no actionable issues through P2. No casts, lint suppressions or weakened checks were added.

### DIFF
--- a/docs/cli/update/status-and-history.md
+++ b/docs/cli/update/status-and-history.md
@@ -47,6 +47,11 @@ gets a separate `runId`.
 
 `openclaw update --json` includes `runId` and the `run` record. `openclaw update status --json`
 includes `activeRun` when a run is active and `lastRun` when history exists.
+If history cannot be read or classified, status still shows update availability
+and runtime findings. Human output explains that run status is unavailable;
+JSON includes `runStatusError` and omits the run fields. This does not mean
+there are no active or past runs, and status does not repair the underlying state.
+
 When the active row has been inactive for more than 30 minutes and its recorded
 driver is verifiably dead, status also reports `abandonedRun` with its `runId`
 and reconciliation `rule`. Status remains read-only: the stored row stays in
--- a/src/cli/update-cli/status.test.ts
+++ b/src/cli/update-cli/status.test.ts
@@ -1,14 +1,20 @@
+import path from "node:path";
+import { DatabaseSync } from "node:sqlite";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { createTempDirTracker } from "../../../test/helpers/temp-dir.js";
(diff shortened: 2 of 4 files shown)


steipete:
@clawsweeper re-review

The prior review stopped before producing findings because remote blob size metadata was unavailable (run 34423162142). The PR head remains `06f2c068434ef80c416b6de3966ac5a14d959536`, and its CI is green.

steipete:
@clawsweeper re-review

The checkout-metadata failure in the prior attempts is now addressed by https://github.com/openclaw/clawsweeper/pull/1519 (`1d41377123e9d796222f3d8014059202ab00ce21`). I verified that the landed request preserves all required tree metadata while fitting the existing subprocess limit. The OpenClaw head remains `06f2c068434ef80c416b6de3966ac5a14d959536`.

steipete:
Merged via squash.

- Prepared head SHA: [06f2c068434ef80c416b6de3966ac5a14d959536](https://github.com/openclaw/openclaw/pull/143557/commits/06f2c068434ef80c416b6de3966ac5a14d959536)
- Landed commit: [d39a83958787a2a735447635da257384ff1fe1bf](https://github.com/openclaw/openclaw/commit/d39a83958787a2a735447635da257384ff1fe1bf)

PR #133661, opened 2026-08-31
fix(cron): preserve complete JSON when run output is piped

steipete:
Related: #133259

## What Problem This Solves

Operators piping `openclaw cron run <id> --wait` to another process could receive truncated, invalid JSON despite the correct success or failure exit code. A valid 128 KiB run summary reproduced a 65,536-byte cutoff for successful and failed waited runs.

## Why This Change Was Made

Cron now uses the existing one-shot output finalizer before terminating, including non-waited outcomes and human-error exits. Its error handler preserves `ExitError` so cleanup and stdout/stderr draining finish without replacing the outcome's diagnostic or exit code. This follows the same output-drain ownership used by Doctor in #133259.

The canonical run-log type replaces a redundant partial type. Scheduling, Gateway responses, summary generation, stores, configuration and shared finalizer deadlines remain unchanged. Production is **+12/−12 (net zero)**; tests/support are **+283/−148 (net +135)** and docs **+1/−1**.

## User Impact

Scripts receive complete waited-run JSON with the original exit status: success remains `0`, failure remains `1`. Queued/not-due outcomes, run history and missing-job diagnostics retain their behavior.

## Evidence

- Maintained OS-pipe regression: original production fails two waited JSON cases while four controls pass; the repair passes all six. Only Gateway RPC results are synthetic; Commander, runtime, finalizer and OS pipes are real.
- Actual source-entry after proof: all six cases pass. Waited outputs contain 133,845/133,885 bytes, the exact 131,072-byte summary and final sentinel, valid JSON, correct 0/1 exits and empty stderr. Proof used Node 26.8.1 source entry, not a packaged binary or live Gateway/job/provider.
- **517 maintained Cron/owner/sibling tests pass**: six process cases, 378 across seven owner/sibling files, and 133 across four corrected validation files. Initial hosted CI exposed 68 obsolete exit-mock expectations; the corrected assertions retain diagnostic and Gateway/no-write checks.
- A later hosted run exposed two tooling-test failures. The heap fixture now retains a named allocation through sampler stop and asserts its positive sampled size; the missing-allocation counterfactual fails, and the full 32-case profiler file passes. The benchmark uses the existing canonical zombie-aware PID helper; its 16 cases and 10 focused PID/wait controls pass. No sampling interval, production cleanup or test deadline was changed.
[...]

PR #143559, opened 2026-09-10
test: reuse source transforms in cron output cases

steipete:
## What Problem This Solves

Six cron-output process cases explicitly disabled TSX caching and gave each child a new temporary root. They repeatedly transformed the same source graph, taking about 35–36 seconds for the complete file.

## Why This Change Was Made

Use the existing cleanup tracker to keep a temporary transform cache for the file, then remove it after every child has finished. Each case keeps separate state/configuration and a fresh Node process. TMPDIR/TEMP/TMP bind the temporary directory on both platform families; Node's separate compile cache remains disabled.

Retain all six cases in their original order: successful/failed waits, queued/not-due admission, run history and human lookup failure. Commander, CLI actions, runtime output and finalization remain real; the existing synthetic Gateway RPC responses, 128-KiB summary/sentinel, 256-KiB buffer, deadlines and all assertions are unchanged. No production, dependency, worker or sharding change. One test file, +6/−3 lines.

## User Impact

Faster CLI output coverage while retaining the regression detector for truncated piped JSON and truthful exit codes.

## Evidence

| Complete six-case invocation | Time |
| --- | ---: |
| Original | 36.18s |
| Candidate with temporary scratch audit | 11.96s |
| Restored original | 34.78s |
| Final candidate without audit | 10.34s |

All positive runs passed in the original order with the same physical dependencies, Node 24.19.0/pnpm 12.3.4, and a newly empty benchmark temporary base. Timings include the normal wrapper, preparation and cleanup. Final peak command RSS was approximately 369 MB.

Twelve boundary snapshots found an empty shared root initially, then 550 TSX cache files retaining their sizes/mtimes across later cases. Only TSX paths persisted at those observed boundaries; the root was removed after cleanup. The audited candidate includes 56ms of observer work. No diagnostic instrumentation is committed.

Regression sensitivity was also checked by temporarily restoring the two premature exits that #133661 replaced. Both cache-disabled and cache-enabled fixtures still failed on truncated JSON at position 65,536, after the original exit-code and signal guards passed. An additional diagnostic run moved the existing history case first to populate the cache: the subsequent warm wait cases ran in 0.691/0.705s against unchanged cache files and still detected the same truncation. Other four cases passed. Production bytes and the original test order were restored before final validation; no counterfactual code is retained.

Required changed-file checks and independent P2 review passed. These are controlled POSIX results; no Windows performance gain or full-main CI saving is claimed.

### DIFF
--- a/src/cli/cron-output.process.test.ts
+++ b/src/cli/cron-output.process.test.ts
@@ -1,9 +1,11 @@
 import path from "node:path";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterAll, afterEach, describe, expect, it } from "vitest";
 import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
 import { spawnNodeEvalSync } from "../test-utils/node-process.js";
 
 const tempDirs = useAutoCleanupTempDirTracker(afterEach);
+// Reuse temporary source transforms while each child keeps its own scenario state.
+const childTempDir = useAutoCleanupTempDirTracker(afterAll).make("openclaw-cron-child-tmp-");
 const summaryBytes = 128 * 1024;
 const sentinel = "\nEND-CRON-SUMMARY\n";
 const summary = "x".repeat(summaryBytes - sentinel.length) + sentinel;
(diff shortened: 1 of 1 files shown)


steipete:
@clawsweeper re-review

The [previous review run](https://github.com/openclaw/clawsweeper/actions/runs/34422153639) failed during source preparation with retryable `review_checkout_unavailable`. The head remains `fe9d231f70eedb4b4e131c9e20b6c1944722455b`; [exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34421433088).

steipete:
Merged via squash.

- Prepared head SHA: [fe9d231f70eedb4b4e131c9e20b6c1944722455b](https://github.com/openclaw/openclaw/pull/143559/commits/fe9d231f70eedb4b4e131c9e20b6c1944722455b)
- Landed commit: [04b265be7fd86d8337a035e7fa344cbc9e7556e3](https://github.com/openclaw/openclaw/commit/04b265be7fd86d8337a035e7fa344cbc9e7556e3)

PR #143576, opened 2026-09-10
test(gateway): isolate CLI announce continuity probe

RomneyDa:
## Summary

- run the CLI-backed announce ordering probe after continuity, cache, image, and MCP assertions
- prevent its separate session from changing process-wide tool topology between continuity turns
- wait for both parent and completion turns to become readable before asserting canonical transcript order

## Release-validation evidence

The original Full Validation failure came from interleaving the independent announce session between consecutive continuity/cache turns. Authenticated isolated Docker validation confirmed the cache lane passes after relocation. The ordinary lane then exposed that delivery metadata can become durable before the CLI transcript flushes the completion turn; the proof now waits for both entries and still requires parent-before-completion order.

Failing parent: https://github.com/openclaw/openclaw/actions/runs/34418806238

## Validation

- `pnpm test:docker:live-cli-backend:claude:cache` (passed; 1 live test, 50.8s)
- `pnpm test:docker:live-cli-backend:claude` (passed after transcript barrier; 1 live test, 35.7s)
- focused oxfmt and oxlint checks passed

### DIFF
--- a/src/gateway/gateway-cli-backend.live-cache.test-helpers.ts
+++ b/src/gateway/gateway-cli-backend.live-cache.test-helpers.ts
@@ -249,9 +249,9 @@ export function prepareClaudeCacheProbeBackend(params: {
   };
 }
 
-async function waitFor<T>(resolve: () => T | undefined): Promise<T> {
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper re-review

PR #143583, opened 2026-09-10
fix(telegram): keep suppressed sends from failing tools and jobs

### LINKED ISSUE #143461: [Bug]: message_sending hook cancellation surfaces as a generic UNAVAILABLE transport error, discarding cancelReason (agents retry a deliberate suppression as if the channel were down)
### What happens

When a `message_sending` hook cancels an outbound Telegram message, the calling agent receives a generic transport error:

```
errorCode=UNAVAILABLE
errorMessage=Error: Telegram sendMessage was suppressed before delivery.
```

The `cancelReason` string returned by the hook is discarded. From the agent's point of view, "a guard intentionally suppressed this message" is indistinguishable from "the Telegram channel is down".

### Why it matters

An agent that cannot tell those apart will do the wrong thing, and it is the *reasonable* thing: treat it as an outage and retry.

We hit this in production. A dedupe hook (`message_sending`, `cancel: true` with a `cancelReason`) started cancelling messages from a scheduled end-of-day run. The agent read the `UNAVAILABLE` errors as a dead channel, sent nine "connectivity test" messages to the user (each of them also cancelled, since they were near-identical to each other), and ran out of its time budget before sending the actual report. The cron then retried the whole run four times. Measured that day: 14 cancellations, 11 of them from that one job.

The hook was doing exactly what it was designed to do, and it had already put the reason in `cancelReason`. Nothing in the error surfaced it.

### Where

In the compiled action runtime (`dist/action-runtime-*.js`):

```js
if (durableResult.status === "failed" || durableResult.status === "partial_failed") throw durableResult.error;
if (durableResult.status === "suppressed") throw new Error("Telegram sendMessage was suppressed before delivery.");
```

The `suppressed` branch constructs a new `Error` with a fixed string. Whatever the hook returned as `cancelReason` is not read.

### Versions

Reproduced on **2026.8.2**. The same two lines are byte-identical in **2026.9.2**, so upgrading does not change this.

The docs for the hook contract (`docs/plugins/hooks.md`) do say that cancellations return a `cancelReason`, and the newer lifecycle APIs surface `cancelled_by_message_sending_hook`. The gap is specifically the legacy direct-delivery path, which returns an empty result array and then throws this generic error.

### Suggested fix

Make an intentional cancellation distinguishable from a transport failure on this path too. Any of these would be enough for a caller to branch correctly:

1. Include `cancelReason` in the thrown error's message and/or attach it as a property.
2. Use a distinct `errorCode` (something like `CANCELLED_BY_HOOK`) instead of `UNAVAILABLE`, which currently reads as "the channel is unavailable".
3. Return a structured non-error result for `suppressed`, the way the newer lifecycle APIs do.

Happy to test a patch against 2026.9.2 on our side.

sunlit-deng:
Closes #143461

## What Problem This Solves

Fixes an issue where an intentional Telegram `message_sending` hook cancellation appeared as a message-tool outage. Announced scheduled jobs could also finish with “Message failed” even though the hook correctly prevented delivery.

## Why This Change Was Made

Return a structured, non-error suppression result with the durable sender's bounded reason. Keep possible delivery, partial sends, and genuine failures distinct. Suppressed sends do not save a delivery route, mirror a delivered message, or clear an earlier genuine message error. Free-form hook diagnostics remain private.

## User Impact

Plugins can intentionally omit a message without causing a false tool failure or failed scheduled job. Normal sends and hook rewrites keep their existing delivery behavior.

## Evidence

- Pinned-main reproduction used a registered hook, the real Gateway-backed agent message tool, Telegram Test Server, a deterministic model endpoint, and an announced isolated scheduled job.
- Baseline: the durable owner recorded intentional suppression with zero platform dispatch; the model still received an error, and the scheduled job saved an error status.
- Candidate `ac3e7a9`: ordinary, scheduled, reasonless, and empty-content suppression return bounded non-error results with zero platform sends. The scheduled job finishes `ok` and remains undelivered. Ordinary commands finish `ok/completed` with one message call and no replay.
- Normal sends and hook rewrites have exact Test Server receipts. Fresh independent live acceptance passes; awkward failure/recovery and unchanged lifecycle preservation are covered by separate native owner checks and source review.
- New regressions failed on pinned main at the affected owners. Focused owner and lifecycle/queue checks passed after repair; corrected Telegram suite: 144 passing tests. SDK surface limits are unchanged and pass.

Preserves @sunlit-deng's contribution and addresses the scheduled-job follow-up from @jerabinovich.

### DIFF
--- a/docs/plugins/sdk-channel-outbound.md
+++ b/docs/plugins/sdk-channel-outbound.md
@@ -243,6 +243,14 @@ Use `payloadOutcomes` when a batch mixes sent, suppressed, and failed
 payloads. Do not infer hook cancellation from an empty legacy
 direct-delivery result.
 
+A `suppressed` result is ambiguous when its reason is
+`adapter_returned_no_identity`, or any payload outcome records a send, an
+identityless send, or a failure with `sentBeforeError`. Check the whole batch
+before treating suppression as intentional non-delivery. For a known omission,
+an action can return `{ status: "suppressed", reason: result.reason }` without
+a tool error or a fabricated receipt. Keep free-form hook diagnostics private.
+Suppression does not recover an earlier failed send.
+
 Failure is not permission to send the same payload through another path.
 Once admitted, the queue owns retry or reconciliation until its exact owner
 acknowledges or terminally retires the intent. Pending custody is not a
--- a/extensions/telegram/src/action-runtime.test.ts
+++ b/extensions/telegram/src/action-runtime.test.ts
@@ -1,6 +1,7 @@
 import os from "node:os";
 import path from "node:path";
 import type { ChannelMessageActionContext } from "openclaw/plugin-sdk/channel-contract";
+import type { DurableMessageBatchSendResult } from "openclaw/plugin-sdk/channel-outbound";
 import type { OpenClawConfig } from "openclaw/plugin-sdk/config-contracts";
 import { resolveStorePath } from "openclaw/plugin-sdk/session-store-runtime";
 import { captureEnv } from "openclaw/plugin-sdk/test-env";
@@ -82,7 +83,7 @@ const sendDurableMessageBatch = vi.fn(
       readFile?: (filePath: string) => Promise<Buffer>;
       workspaceDir?: string;
     };
-  }) => {
+  }): Promise<DurableMessageBatchSendResult> => {
     const payload = params.payloads[0] ?? {};
     const mediaUrls = payload.mediaUrls?.length
       ? payload.mediaUrls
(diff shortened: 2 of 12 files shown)


obviyus:
Thanks @sunlit-deng! Landed in 14f30a55449f47fe13a2e8609c91c19cb8164dd6.

**Proof:** A registered hook, the real Gateway-backed message tool, and Telegram Test Server exercised the repair. Independent live acceptance and the source/owner checks passed.

- **Bug:** Intentional suppression became a tool error and could fail an otherwise healthy scheduled job.
- **Fix:** Preserve bounded non-error suppression without inventing delivery, changing routes/transcripts, or clearing an earlier genuine error. Real and uncertain delivery failures stay distinct.

| Before | After |
| --- | --- |
| Cancelled message returned `UNAVAILABLE` | Structured `suppressed` result; no platform send |
| Announced job saved “Message failed” | Job finishes `ok` and remains undelivered |

Normal and rewritten sends have exact recipient receipts. Reasonless cancellation and emptied content retain their distinct bounded outcomes. Native owner/recovery tests, the final 144-test Telegram suite, and [exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34548819910) passed. The final review found no blocking issues. Issue #143461 is closed.

PR #143590, opened 2026-09-10
fix(memory): keep managed local-service readiness on its own budget during memory_search

### LINKED ISSUE #143169: memory_search aborts managed local embedding cold start before ready timeout

wangmiao0668000666:
Fixes #143169.

Managed embedding cold starts could exceed the 15-second memory-search limit even while inside the configured readiness timeout. The search then aborted startup and lost semantic recall.

This change pauses the corpus and query-embedding clocks only while the service owner waits for readiness. Each clock resumes with its remaining budget before router reconciliation or embedding HTTP work. Concurrent wiki searches keep a separate 15-second deadline. A wiki timeout preserves completed memory results and leaves memory retries available. Caller cancellation, shared service leases, readiness failure, and idle shutdown retain their existing owners.

The repair also handles a query watchdog that subscribes during an existing pause and removes the corpus listener when work ends. No new configuration, database schema, persistence policy, or dependency is introduced.

Validation uses an isolated built Gateway, the registered `memory_search` tool, real Memory Core SQLite storage, a synthetic loopback chat endpoint, and a managed child embedding service. The baseline aborts a 70-second cold start at the search deadline. The corrected candidate returns the known semantic hit after readiness; its warm retry works. Slow wiki results are excluded and fast wiki results are retained. Persistent public-client controls cover shared and solo cancellation, the configured readiness failure, and an embedding HTTP request that exceeds the remaining search budget.

A separate registered-adapter control confirms that 100 ms reconciliation after a 70-second startup preserves recall, while 25-second reconciliation is aborted at about 15 seconds before any query embedding request.

After the final CI correction, 120 focused memory tests passed, including the stalled-wiki result and retry regression. A rebuilt public flow preserved memory at the wiki timeout and completed the next memory search in 107 ms. Earlier provider, service, and deadline-control tests remain applicable to their unchanged owners. Original-code probes fail at the repaired boundaries. Formatting, ordinary lint, source ratchets, the native commit hook, and the runtime build passed. Full type checking and broader suites run in GitHub CI. The loopback fixtures use deterministic embeddings; they do not certify an external provider or arbitrary model quality.


obviyus:
Thanks @wangmiao0668000666!

**Proof:** A built Gateway used the registered `memory_search` tool, real Memory Core storage, and a managed child service with synthetic loopback providers.

- **Bug:** The search deadline stopped valid cold starts and lost semantic recall.
- **Fix:** Readiness uses its own budget; wiki, reconciliation, and embedding requests retain their bounds. A wiki timeout preserves completed memory results and permits an immediate retry.

| Before | After |
| --- | --- |
| A 70-second startup was stopped at 15 seconds. | Startup completes and returns the semantic memory hit. |
| A stalled wiki could hide available memory and trigger cooldown. | Memory remains available; the next search completed in 107 ms. |

Validation: 120 focused memory tests, retained provider/lifecycle controls, fresh exact-head review, and the [complete GitHub CI run](https://github.com/openclaw/openclaw/actions/runs/34441359178) passed. Both previously failing CI jobs are green.

The automated migration reminder does not apply: this change adds only per-call timing state and changes no stored schema or data format. The deadline-control export is a private host interface, so the earlier third-party SDK suggestion is not a public compatibility requirement.

wangmiao0668000666:
Thank you @obviyus — both for the thorough review and for carrying this over the finish line.

Your merge-commit rework taught me a couple of things I'll keep for future contributions: the way you ran the reconciliation-boundary control (100ms vs 25s reconcile after a 70s cold start) to pin the budget ownership semantics, and validating against a real built Gateway rather than stopping at focused tests. Seeing the findings land with that level of proof set a much higher bar than my original submission.

Appreciate you preserving the original commit in the history as well. Glad the cold-start path is fixed.

Issue #142310, opened 2026-09-08
WebChat: text before tool calls rendered twice (session observer invalid JSON on every turn)


djesse4me-afk:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

In WebChat, any text written before a tool call is displayed twice in the chat. The duplication happens on every agent turn that includes a tool call. Text-only responses (no tool calls) appear correctly.

The gateway logs a session observer returned invalid JSON twice → session observer disabled error on every single agent turn, across all sessions.

### Steps to reproduce

Open WebChat
Send a message that causes the agent to make a tool call
Observe the text before the tool call appears twice

### Expected behavior

Text before a tool call appears once.

### Actual behavior

Text before a tool call appears twice. Gateway logs show:

subsystem: gateway/session-observer
error: "session observer returned invalid JSON twice"
message: "session observer disabled after consecutive failures"

This fires on every turn, gets disabled, re-arms for the next turn, and repeats indefinitely.

### OpenClaw version

2026.9.3

### Operating system

macOS 26.6.2 Chrome 152 (also reproduced in Incognito) nstall method: npm / LaunchAgent

### Install method

Install method: npm / LaunchAgent

### Model

claude-sonnet-4.6

### Provider / routing chain

That's asking how your models are configured. Write:  **anthropic/claude-sonnet-4-6 (primary) · ollama/llama3.2:3b (heartbeat only) · no fallbacks configured**

### Additional provider/model setup details

Began September 7 after openclaw doctor --fix auto-enabled the browser plugin
Persists with browser plugin disabled, memory-core dreaming disabled, iMessage disabled
Persists across multiple gateway restarts, fresh sessions, and openclaw update repair
openclaw doctor --fix cleared it for exactly one agent turn, then returned

### Logs

```shell

```

djesse4me-afk:
Clarification: The doubling seems to be session-specific. Fresh sessions appear to work correctly across multiple tool calls. The affected session had been running since the previous day and survived numerous gateway restarts and config changes. The bug may involve accumulated session state in the WebChat or SQLite that causes the rendering issue.

djesse4me-afk:
UPDATE: Behavior is recurring on new chats. disregard previous comment.

djesse4me-afk:
RESOLVED — Disabling the session observer fixed this issue.

What I did:

Added gateway.controlUi.sessionObserver.enabled = false to openclaw.json
Reloaded the gateway
[...]

PR #143591, opened 2026-09-10
fix(ui): hide history-owned pre-tool stream text

teddytennant:
## What Problem This Solves

Fixes an issue where WebChat rendered pre-tool assistant text twice when the same wording was already in history and still arriving as an unkeyed live stream segment.

## Why This Change Was Made

`buildChatItems` appended unkeyed stream segments whenever `persisted !== true`, even if the current turn already had that text as a history assistant message. Hide only the overlapping prefix on those segments, scoped to the current turn (reset on the next user row). The cumulative live `stream` path is left alone so a repeated same-run answer still renders.

Related to #142310.

## Evidence

Baseline: `4947087c`.

```
node scripts/run-vitest.mjs run ui/src/pages/chat/chat-thread-build.test.ts ui/src/pages/chat/chat-state.test.ts
```

120 passed, including `hides a pre-tool stream segment already persisted in history`, `still renders a later same-text stream after a new user turn`, `still renders a novel tail after a persisted prefix`, and `preserves repeated commentary and distinct answers within the same active run`.


VACInc:
Related merge: #146554 prevents tool-triggered splitting of an unfinished live assistant body. It does not add this PR’s current-turn history-prefix overlap handling for unkeyed stream segments. Complete supersession is not established, so this PR should remain open for its history/live overlap scope.

teddytennant:
yeah, still needed. 146554 stopped tool events from rolling the stream over; this one hides the unkeyed pre-tool copy that buildChatItems still paints on top of history.

VACInc:
Related to #155336, but not superseded by it. #155336 fixes fallback-notice contamination of terminal answer matching. This PR fixes a separate `buildChatItems` overlap: unkeyed pre-tool stream text already owned by current-turn history. Its scoped current-turn reset and preservation of later same-text answers remain independently necessary. Please keep this PR open.

Action: KEEP OPEN. DO NOT CLOSE.

teddytennant:
yeah, leaving it open.

Issue #112572, opened 2026-07-22
[Feature]: Inspect native sub-agent transcripts in Background tasks


steipete:
## Summary

Show provider-native sub-agents in the Control UI Background tasks rail and let operators inspect their detailed transcript without leaving the parent conversation.

## Problem to solve

The Control UI task rail already renders OpenClaw task-registry rows and can open ordinary OpenClaw child sessions. Provider-native child work has a different ownership model:

- Codex native sub-agents are mirrored into the task ledger but have no OpenClaw `childSessionKey`, so the existing transcript action falls back to the parent session instead of the child thread.
- Claude Code exposes native background Agent/workflow lifecycle events, but OpenClaw currently uses them only to keep the parent turn alive; it does not mirror those children into the task ledger.
- The current transcript action replaces the parent pane. Operators lose the Background tasks context instead of getting the focused side inspector used by modern coding-agent UIs.

The practical result is that long native child work can be visible only as parent activity, with no reliable way to inspect the child's tool calls, outputs, progress, or final result from the task rail.

## Proposed solution

- Project Claude Code native background Agent/workflow lifecycle onto the existing generic task ledger while keeping `task_id`, spawning `tool_use_id`, and sub-agent `agentId` as separate identifiers.
- Give native task rows a typed transcript target owned by the provider's existing session catalog:
  - ordinary OpenClaw sub-agents keep using their child session;
  - Codex native sub-agents target the Codex app-server child thread;
  - Claude Code native sub-agents target the public Agent SDK sub-agent transcript reader.
- Keep transcript bodies out of task records. The task stores only the small locator; provider catalogs continue to own paginated history and normalization.
- Change **View transcript** into an in-rail inspector with Back, Open full, and Close controls. Reuse the canonical chat transcript and tool-card renderer so command, read, edit, write, tool-result, reasoning, loading, pagination, and error presentation do not fork.
- Keep native task cancellation honest. A childless native row must not claim that a live external worker was stopped unless its harness has a real cancellation handle.

No new config, storage file, or SQLite schema version is needed.

### Keep navigating to the child session

This is correct for OpenClaw-owned child sessions, but it replaces the parent pane and cannot identify childless native harness threads.

### Copy native transcripts into task `detail_json`

This would duplicate provider-owned history, grow the shared task ledger, create another retention boundary, and require a second transcript parser. A typed locator keeps one owner for each transcript.

### Parse Claude Code's private `isSidechain` JSONL layout

[...]

PR #143594, opened 2026-09-10
feat: inspect native child activity in Background tasks

anyech:
Related: #112572

## What Problem This Solves

Users cannot inspect a Codex-native sub-agent's tool activity and public output from Background tasks because it has no OpenClaw child session. The task panel only shows the task summary, leaving users unable to tell what the child is doing without asking the parent.

## Why This Change Was Made

Add a task-scoped, paginated native transcript reader and reuse the existing read-only transcript renderer in the task sidebar. The open inspector refreshes independently of parent events, including while the parent has yielded. Active tasks retry transient read failures; stable source IDs keep pagination and live refresh from duplicating output. Native children remain excluded from the ordinary session catalog.

Access is checked against the requesting session and the actual parent binding, provider connection, and child lineage; changes during a read revoke its result. Raw provider payloads and reasoning are excluded.

This is an implementation proposal for the existing feature request, not a claim that its product-decision labels have been resolved. Discord progress publishing and lifecycle notifications are outside this PR.

## User Impact

Users can open a native child in Background tasks, see tool calls/results and public messages, load older activity, and keep watching while the parent is idle. The inspector updates its heading when the task registry reports completion.

## Evidence

- Behavior addressed: native child activity is unavailable in the task sidebar.
- Real setup tested: built Control UI in Chromium with a deterministic mock Gateway; no production sessions or credentials.
- Exact steps or command run after this patch: `node scripts/run-vitest.mjs run --config test/vitest/vitest.ui-e2e.config.ts --configLoader runner ui/src/e2e/chat-native-task-history.e2e.test.ts`.
- Evidence after fix: browser assertions verify child tool output, a second history read without parent events, and the completed heading. Captures were inspected locally.
- Observed result after fix: the child transcript updates independently of the parent, and no native thread identifier is accepted from the browser.
- What was not tested: Discord publishing, production Gateway deployment, and platform-native app builds.

## Validation

- Built Control UI Chromium E2E: 1 passed using a deterministic mock Gateway.
- Task sidebar state/rendering regression tests: 20 passed, including transient `tasks.get` and `tasks.history` recovery.
- Gateway task-history and method-scope tests: 197 passed; Codex task-history reader: 14 passed.
- UI, core and extension production typechecks passed; all 17 core-test shards and extension-test typechecks passed.
- Core lint, protocol registry/since checks, contributor i18n verification and plugin-boundary checks passed.
- Two independent review findings were fixed with red/green regressions: missing stable record IDs and polling that stopped permanently after transient errors.

- Codex plugin lint passed.

Validation is from the isolated source checkout. Native Android/iOS/macOS app builds and a production Gateway rollout were not performed.


anyech:
Superseded by #143747 and its follow-ups #143869 and #144166. Current main provides task-scoped native Codex transcripts and pagination through the harness-owned history path. Closing this alternative implementation to avoid maintaining a duplicate session-catalog hook. Thanks!

PR #137783, opened 2026-09-04
fix(firecrawl): surface DNS resolution failures instead of SSRF policy errors

gaoanze888:
## What Problem This Solves

In the Firecrawl plugin, `firecrawlEndpointTargetsPrivateNetwork()` swallows hostname-resolution errors with `catch { return false }`. A `baseUrl` whose host simply **fails to resolve** is therefore reported as a *privateness policy* violation:

```
Firecrawl HTTP baseUrl must target a private or internal self-hosted endpoint. Use https:// for public hosts.
Firecrawl custom baseUrl must target a private or internal self-hosted endpoint. Host: <host>
```

Both messages point at SSRF policy / scheme, so users (and AI assistants asked to debug it) chase a config or HTTPS problem when the actual cause is DNS. This is a follow-up to #63877 — self-hosted HTTP endpoints work now, but the failure mode for unresolvable hosts is misleading.

Closes #135333.

## Approach

`firecrawlEndpointTargetsPrivateNetwork` resolves the host via `resolvePinnedHostnameWithPolicy`. That call fails in two distinct ways that the blanket `catch { return false }` conflated:

1. **The host cannot be resolved at all** — `resolvePinnedHostnameWithPolicy` throws `Error("Unable to resolve hostname: …")` (empty DNS answer) or propagates the raw lookup rejection (e.g. `getaddrinfo ENOTFOUND`). This is a DNS failure, not an SSRF result.
2. **The host resolves to a blocked IP** — `resolvePinnedHostnameWithPolicy` throws `SsrFBlockedError` (metadata / link-local rebinding). This *is* an SSRF result and must keep falling through to the policy error the caller raises for non-private targets.

The fix narrows the catch to the SSRF case:

```ts
} catch (error) {
  if (!(error instanceof SsrFBlockedError)) {
    throw new Error(
      `Firecrawl baseUrl host could not be resolved: ${url.hostname}. Verify the hostname or DNS configuration.`,
      { cause: error },
    );
  }
  return false;
}
```

`SsrFBlockedError` is already imported in this module (used by `assertFirecrawlScrapeTargetAllowed`), so the fix adds no new dependency. A blocked-IP resolution still returns `false` → the caller raises the existing privateness policy error; only the genuine DNS failures now surface with a message that names the host and points at DNS.

## Evidence

- `npx vitest run extensions/firecrawl/src/firecrawl-client.test.ts` → 65 passed (62 baseline + 3 new).
- New `describe("resolveEndpoint baseUrl resolution")`:
[...]

PR #143597, opened 2026-09-10
fix(firecrawl): unresolvable baseUrl reported as private-endpoint policy

### LINKED ISSUE #135333: [Bug]: Unresolvable self-hosted Firecrawl baseUrl reported as "not private" instead of a DNS failure
### Bug type

### Beta release blocker

No

### Summary

In the Firecrawl plugin, `firecrawlEndpointTargetsPrivateNetwork()` swallows hostname-resolution errors with `catch { return false }`. A `baseUrl` whose host simply **fails to resolve** is therefore reported as a *privateness policy* violation:

```
Firecrawl HTTP baseUrl must target a private or internal self-hosted endpoint. Use https:// for public hosts.
Firecrawl custom baseUrl must target a private or internal self-hosted endpoint. Host: <host>
```

Both messages point at SSRF policy / scheme, so users (and AI assistants asked to debug it) chase a config or HTTPS problem when the actual cause is DNS. This is a follow-up to #63877 — self-hosted HTTP endpoints work now, but the failure mode for unresolvable hosts is misleading.

### Steps to reproduce

Environment: OpenClaw `2026.8.1`, `@openclaw/firecrawl-plugin` `2026.8.1`, Linux (aarch64), self-hosted Firecrawl reachable on a private IP.

Calling the plugin's own validation (`testing.resolveEndpoint`) with various `baseUrl` values:

| baseUrl | result |
| --- | --- |
| `http://192.168.8.19:3002` | OK — `selfHosted` |
| `http://firecrawl.local:3002` | OK — `selfHosted` |
| `http://nichtexistent.local:3002` | **OK — `selfHosted`, although the host does not resolve at all** |
| `http://firecrawl.internal:3002` | OK — `selfHosted` |
| `http://firecrawl.lan:3002` | **ERROR — "must target a private or internal self-hosted endpoint. Use https:// for public hosts."** |
| `http://raspberrypi:3002` | **ERROR — same message** |
| `https://firecrawl.example.com` | ERROR — correct, resolves publicly |

`.local` and `.internal` short-circuit through `isBlockedHostnameOrIp()` and are accepted without any resolution — including names that do not exist. Meanwhile `.lan` and bare hostnames (both perfectly ordinary on a LAN) hit the `resolvePinnedHostnameWithPolicy()` path, fail to resolve, and are then reported as if they were public.

### Expected behavior

Distinguish the two cases and say which one occurred:

- host could not be resolved -> "Unable to resolve Firecrawl baseUrl host: `<host>`"
- host resolved to a public address -> the current "must target a private or internal self-hosted endpoint" message

Ideally the resolution failure should not be silently converted into `false` at all.

### Secondary issue: keyless self-hosted search is only reachable via the "free hosted tier" provider

`runFirecrawlSearch()` throws

```
web_search (firecrawl) needs a Firecrawl API key. Set FIRECRAWL_API_KEY in the Gateway environment,
or configure plugins.entries.firecrawl.config.webSearch.apiKey.
```

for provider `firecrawl` whenever no API key is set — even when `baseUrl` points at a self-hosted instance that requires none. The only keyless path is provider `firecrawl-free`, which passes `access: "keyless"`.

That provider is described as:

> "Free web search via Firecrawl's hosted starter tier — no API key required"

which reads as though it ignores the self-hosted endpoint. It does not: `resolveFirecrawlBaseUrl()` is shared and prefers the configured `webSearch.baseUrl`. So the correct configuration for "self-hosted Firecrawl, no API key" is a provider whose label says it uses Firecrawl's hosted tier. That is easy to get wrong, and the resulting error message suggests obtaining an API key rather than switching providers.

Suggestion: let provider `firecrawl` run keyless when the resolved `baseUrl` is a self-hosted/private endpoint, or make the label and the error message mention the self-hosted case.

DonnieFi:
Closes #135333.
Related: #137783.

## What Problem This Solves

Fixes an issue where operators configuring a self-hosted Firecrawl `baseUrl` would see a private-endpoint / SSRF policy error when the hostname simply failed DNS (for example `http://raspberrypi:3002` or `http://firecrawl.lan:3002`). The message pointed them at HTTPS or privateness policy when the real cause was resolution failure.

## Why This Change Was Made

`firecrawlEndpointTargetsPrivateNetwork` was swallowing every resolver failure with `catch { return false }`, so DNS misses took the same path as “resolved to a public host.” Resolution failures (empty answers and lookup rejects such as `ENOTFOUND`) now throw a host-specific DNS error; `SsrFBlockedError` still falls through to the existing private-endpoint policy messages. Successful public resolves keep those policy errors. The same classification was previously attempted in #137783 and self-closed over live channel proof this error-path change does not need. The searxng client still has the same swallow-all catch and is left as a follow-up.

## User Impact

Operators and plugin authors who mistype or cannot resolve a Firecrawl `baseUrl` host now see:

`Unable to resolve Firecrawl baseUrl host: <host>. Verify the hostname or DNS configuration.`

instead of the HTTP/HTTPS “must target a private or internal self-hosted endpoint” messages. Hosted `api.firecrawl.dev` and correctly resolving private self-hosted endpoints behave as before.

### Real behavior (system DNS, no mocked lookupFn)

Same plugin invocation path on current `main` vs this branch. `baseUrl=http://unresolvable.invalid:3002`, keyless `runFirecrawlSearch`, plus `testing.resolveEndpoint`. Host deliberately uses the reserved `.invalid` TLD so resolution fails via real DNS (`ENOTFOUND`).

**Before (origin/main @ cdda5825f8bb):**

```text
=== Before-fix real-DNS Firecrawl proof (current main) ===
ref: origin/main
sha: cdda5825f8bb50e4112f13a9405856cad87bfa79
baseUrl: http://unresolvable.invalid:3002
lookupFn: (none — system DNS)

--- 1) testing.resolveEndpoint ---
error.message: Firecrawl HTTP baseUrl must target a private or internal self-hosted endpoint. Use https:// for public hosts.

--- 2) runFirecrawlSearch (keyless) ---
error.message: Firecrawl HTTP baseUrl must target a private or internal self-hosted endpoint. Use https:// for public hosts.
```

**After (fix/firecrawl-unresolvable-baseurl @ c01c38b263a0):**

```text
=== After-fix real-DNS Firecrawl proof ===
branch: fix/firecrawl-unresolvable-baseurl
head: c01c38b263a071d15542606060c6017cac21cdec
baseUrl: http://unresolvable.invalid:3002
lookupFn: (none — system DNS)

--- 1) testing.resolveEndpoint (plugin validation path) ---
error.message: Unable to resolve Firecrawl baseUrl host: unresolvable.invalid. Verify the hostname or DNS configuration.

--- 2) runFirecrawlSearch (plugin invocation; fails at endpoint validation before HTTP) ---
error.message: Unable to resolve Firecrawl baseUrl host: unresolvable.invalid. Verify the hostname or DNS configuration.

--- 3) https sibling ---
error.message: Unable to resolve Firecrawl baseUrl host: unresolvable.invalid. Verify the hostname or DNS configuration.

--- 4) DNS sanity (system resolver) ---
getaddrinfo ENOTFOUND: getaddrinfo ENOTFOUND unresolvable.invalid
```

### Regression tests (mocked DNS)

```text
pnpm exec vitest run extensions/firecrawl/src/firecrawl-client.test.ts
# 34 passed — including DNS vs privateness classification cases

pnpm test:extension firecrawl
# Test Files  3 passed (3)
# Tests       103 passed (103)
```

### DIFF
--- a/extensions/firecrawl/src/firecrawl-client.test.ts
+++ b/extensions/firecrawl/src/firecrawl-client.test.ts
@@ -1,5 +1,6 @@
+import type { LookupFn } from "openclaw/plugin-sdk/ssrf-runtime";
 // Focused parser contracts; public Firecrawl execution lives in firecrawl-tools.test.ts.
-import { beforeAll, describe, expect, it } from "vitest";
+import { beforeAll, describe, expect, it, vi } from "vitest";
 
 let firecrawlClient: typeof import("./firecrawl-client.js").testing;
 const hostileControlToken = ["<|im_", "start|>system"].join("");
@@ -212,5 +213,44 @@ describe("Firecrawl scrape payloads", () => {
         maxChars: 1_000,
       }),
     ).toThrow("Firecrawl scrape returned no content.");
+  });
+});
+
+// ---------------------------------------------------------------------------
+// resolveEndpoint — baseUrl SSRF + DNS failure classification (#135333)
+// ---------------------------------------------------------------------------
+describe("resolveEndpoint baseUrl resolution", () => {
+  it("surfaces a DNS resolution failure instead of a privateness policy violation", async () => {
+    // Unresolvable hosts fail inside the SSRF resolver. Previously the catch
+    // swallowed that and reported a privateness policy violation (#135333).
+    const lookupFn: LookupFn = vi.fn(async () => {
+      throw new Error("getaddrinfo ENOTFOUND unresolvable.invalid");
+    });
+    await expect(
+      firecrawlClient.resolveEndpoint("http://unresolvable.invalid", "/v2/search", lookupFn),
+    ).rejects.toThrow(/Unable to resolve Firecrawl baseUrl host: unresolvable\.invalid/i);
+  });
+
+  it("surfaces a DNS resolution failure when lookup returns no addresses", async () => {
+    const lookupFn: LookupFn = vi.fn(async () => []);
+    await expect(
+      firecrawlClient.resolveEndpoint("http://empty.invalid", "/v2/search", lookupFn),
+    ).rejects.toThrow(/Unable to resolve Firecrawl baseUrl host: empty\.invalid/i);
+  });
+
+  it("surfaces a DNS resolution failure for https unresolvable hosts", async () => {
+    const lookupFn: LookupFn = vi.fn(async () => {
+      throw new Error("getaddrinfo ENOTFOUND unresolvable.invalid");
+    });
+    await expect(
+      firecrawlClient.resolveEndpoint("https://unresolvable.invalid", "/v2/search", lookupFn),
+    ).rejects.toThrow(/Unable to resolve Firecrawl baseUrl host: unresolvable\.invalid/i);
+  });
+
+  it("still rejects a host resolving to a public address as a privateness policy violation", async () => {
+    const lookupFn: LookupFn = vi.fn(async () => [{ address: "8.8.8.8", family: 4 }]);
+    await expect(
+      firecrawlClient.resolveEndpoint("http://public.example", "/v2/search", lookupFn),
+    ).rejects.toThrow(/must target a private or internal self-hosted endpoint/i);
   });
 });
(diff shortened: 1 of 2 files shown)


DonnieFi:
Added real-DNS after-fix proof to the PR body (before/after against `unresolvable.invalid`, no mocked `lookupFn`, including keyless `runFirecrawlSearch`).

@clawsweeper re-review

DonnieFi:
PR body now includes before/after real-DNS terminal proof (system resolver, keyless `runFirecrawlSearch` + `testing.resolveEndpoint` against `http://unresolvable.invalid:3002`). Prior re-review queue entry completed as a no-op.

@clawsweeper re-review

PR #143599, opened 2026-09-10
fix(doctor): recover legacy node tokens with invalid scopes

### LINKED ISSUE #143263: [Bug]: doctor recommends 'devices rotate --role node' for an operator-only-baseline device, but that rotation is always denied (scope-outside-approved-baseline)
### Bug type
Behavior bug (doctor recommends a command that cannot succeed).

### Summary
For a paired device whose approved scope baseline contains only `operator.*` scopes but which also holds a `node` role token issued with those operator scopes, `openclaw doctor` reports the node token as outside the baseline and recommends `openclaw devices rotate --device <id> --role node`, but that exact command is always denied with `scope-outside-approved-baseline`.

### Steps to reproduce
1. Paired device `cli` (the local CLI, paired March 2026) with roles `operator, node`, approved scopes `operator.read, operator.admin, operator.write, operator.approvals, operator.pairing`, and a `node` token whose scopes are those same five `operator.*` scopes (`openclaw devices list --json` → `tokens[role=node].scopes`).
2. `openclaw doctor` → "Paired device cli (…) has a node token outside the approved scope baseline [operator.admin, operator.approvals, operator.pairing, operator.read, operator.write]. Rotate it with `openclaw devices rotate --device <id> --role node`."
3. Run exactly that command from the same CLI.

### Expected behavior
Rotation succeeds and issues a node token that satisfies the baseline check (e.g. with `operator.*` scopes filtered out, as `resolveRoleTokenScopes` does when a node token is first created), or doctor recommends a reachable fix.

### Actual behavior
```
{"ok":false,"error":{"type":"cli_error","message":"device token rotation denied"}}
```
Gateway log: `device token rotation denied device=<id> role=node reason=scope-outside-approved-baseline`.

Why it can't succeed (from the bundle): `rotateDeviceToken` uses `params.scopes ?? existing.scopes ?? device.scopes`, so without `--scope` the operator scopes are reused; `roleScopesAllow(role="node")` requires every scope to start with `node.` and be in the baseline. `--scope` with an empty value is rejected by `DeviceTokenRotateParamsSchema` (`NonEmptyString[]`), and any `node.*` scope fails because the baseline holds only operator scopes. So every input is rejected.

Workaround used: `openclaw nodes remove --node <id>` (removes only the node role on a multi-role device; operator pairing intact). Doctor then reports "Local cached node device auth … no longer has a matching active gateway token, and that role is no longer approved … remove the stale cached node auth entry", and nothing in the CLI removes that `device_auth_tokens` row; cleared it by calling the bundle's `clearDeviceAuthToken`.

### OpenClaw version
2026.9.1 (ad6fe23)

### Operating system
macOS 26.6.2 (Apple Silicon, M4)

### Install method
npm global install under Homebrew `node@22` (22.23.2); gateway as LaunchAgent

obviyus:
Fixes #143263. Thanks @matthewmoroz for the report.

## What Problem This Solves

Doctor identified legacy node tokens with inherited operator scopes, but its suggested rotation kept those scopes and was denied.

## Why This Change Was Made

The devices rotate command adds --no-scopes, and Doctor uses it in node recovery guidance. Rotation removes only the matching stale local node credential in the same transaction while preserving operator pairing, approved scopes, authorization, and token-delivery rules.

## User Impact

Operators can rotate a legacy node token without retaining invalid scopes. Omitted scopes keep their existing meaning; --scope and --no-scopes conflict. Operator access survives node recovery.

Consumers: node recovery guidance and token rotation now support explicitly empty scopes.

## Evidence

Public pairing on release 2026.3.11 followed by an upgrade reproduced the failure. The corrected command succeeded, removed the matching cached credential, and preserved operator access through restart and reconnect. All 203 focused command-line, Doctor, pairing, credential-store, and Gateway authorization tests passed. Public command and signed-protocol checks covered recovery, rejected changes, token delivery, revocation, interrupted responses, and restart. The real node fixture used Linux; native macOS executable discovery was not run.

### DIFF
--- a/docs/cli/devices.md
+++ b/docs/cli/devices.md
@@ -137,9 +137,18 @@ openclaw devices rotate --device <deviceId> --role operator --scope operator.rea
 
 - The target role must already exist in that device's approved pairing contract; rotation cannot mint a new unapproved role.
 - Omitting `--scope` retains the target token's current scopes. Passing explicit `--scope` values replaces that scope set, within the device's approved baseline, for future cached-token reconnects.
+- Pass `--no-scopes` to request an empty scope set. It cannot be combined with `--scope`.
 - A non-admin paired-device caller can rotate only its **own** device token, and the target scope set must stay within the caller's own operator scopes; rotation cannot mint or preserve a broader token than the caller already has.
 
 Returns rotation metadata as JSON. If the caller rotates its own token while authenticated with that device token, the response includes the replacement token so the client can persist it before reconnecting. Shared-secret callers and callers rotating another device never receive the bearer token.
+
+When Doctor reports a legacy node token carrying operator scopes, use its explicit recovery command:
+
+```bash
+openclaw devices rotate --device <deviceId> --role node --no-scopes
+```
+
+This recovery requires `operator.admin` and preserves the device's operator pairing and approved scopes. The Gateway removes only a local cached node token that matches the retired legacy token, in the same commit as rotation. For a node host using a separate state directory, provide valid shared Gateway authentication and restart the node to refresh its cache. A retired device token alone cannot authenticate the reconnect.
 
 ### `openclaw devices revoke --device <id> --role <role>`
 
--- a/src/cli/devices-cli.runtime.ts
+++ b/src/cli/devices-cli.runtime.ts
@@ -58,6 +58,7 @@ type DevicesRpcOpts = {
   device?: string;
   role?: string;
   scope?: string[];
+  scopes?: boolean;
   name?: string;
 };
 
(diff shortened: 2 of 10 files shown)


obviyus:
Thanks @matthewmoroz!

**Proof:** Public pairing on v2026.3.11 reproduced the failure after upgrade. Independent CLI and signed-protocol acceptance verified recovery, preserved operator access, node reconnect, authorization denials, token delivery, interruption, and hosting lifecycle behavior.

| Before | After |
| --- | --- |
| Doctor's node-token rotation retained invalid operator scopes and was denied. | Doctor recommends `--no-scopes`; the authorized rotation succeeds and removes only the matching stale node cache. |

203 focused tests passed. Exact-head security review cleared the patch, and [CI passed](https://github.com/openclaw/openclaw/actions/runs/34425830886). Landed in e75ebdebdfc6338b5d2125e94fd707258aaec89a.

PR #143601, opened 2026-09-10
fix(system-agent): read quoted configuration keys

steipete:
## What Problem This Solves

Fixes an issue where users reading configuration through the OpenClaw setup agent or its tool would receive “not set” for existing record keys containing dots or brackets. For example, `config get channels.modelByChannel.telegram["team.ops[west]"]` was accepted as a command but could not read the matching value.

## Why This Change Was Made

Use the existing configuration-path parser for reads as well as command validation, removing the reader's independent dot/bracket splitting. Keep whole-config redaction before selection and retain existing array lookup behavior, including length, zero-padded indices, and indices above the CLI writer's sparse-write limit. Malformed paths supplied directly through the tool now receive the canonical path error through its existing error handler.

## User Impact

Quoted and escaped record keys can be read correctly, and sensitive values remain redacted. The setup-agent documentation includes a quoted-key example. No configuration writes, schema, or inference behavior changes.

## Evidence

- Before the fix, four new operation-boundary cases failed with “not set”; the ordinary indexed-array control passed. The cases cover double quotes, single quotes, escaped punctuation, and a redacted credential field.
- Final operation, parser, and redaction suites: 129 tests passed, including preserved zero-padded indices and array length.
- A comparison of the original and final reader functions with the real shared parser preserves 523 existing results, including missing/null/undefined values and indices above 100,000, and verifies four corrected quoted-key results. This is bounded function proof; no live Gateway or real configuration was used.
- Fresh managed review: scoped clean through P2. The full changed-file gate and `pnpm build` pass. The build reports a non-enforcing UI startup gzip budget overrun of 8 bytes (352,243 versus 352,235 bytes); no UI source is changed.

This is a correctness repair found during the reduction audit. Sharing the existing config-read test setup keeps every assertion and makes room for the new regressions without raising the file limit. Complete local cost is −3 production / −2 test = −5 maintained code lines (−10 physical); documentation adds 3 code / 4 physical lines separately.

### DIFF
--- a/docs/cli/openclaw.md
+++ b/docs/cli/openclaw.md
@@ -102,6 +102,10 @@ quit
 
 OpenClaw uses typed operations instead of editing config ad hoc.
 
+For `config get`, quote record keys that contain dots or brackets, such as
+`config get channels.modelByChannel.telegram["team.ops[west]"]`.
+Config reads redact sensitive values before selecting the requested path.
+
 Read-only operations run immediately: show overview, list agents, list installed plugins, search ClawHub plugins, show model/backend status, run status/health checks, check Gateway reachability, run doctor without interactive fixes, validate config, show the audit-log path.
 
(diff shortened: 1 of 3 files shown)


steipete:
@clawsweeper re-review

PR #143605, opened 2026-09-10
fix(gateway): keep update history reads responsive

steipete:
## What Problem This Solves

Fixes an issue where requesting update-run history could pause other Gateway work when the shared-state database was not already open.

## Why This Change Was Made

`update.runs.get` and `update.runs.list` now await an artifact-preserving history snapshot. The existing raw-copy worker and readonly query logic remain authoritative; warm callbacks still run directly, and fresh admission is rechecked after snapshot preparation. Existing synchronous ledger consumers and `update.status` keep their behavior.

## User Impact

Opening update history or a run's details can wait for a cold snapshot without blocking the Gateway event loop on that preparation. Database contents, sidecars, committed-view isolation, and error behavior are preserved.

## Evidence

- The authenticated CLI regression fails on the original code because the Gateway isolate cannot progress before snapshot preparation returns, and passes with the repair. It lives in the existing update integration suite.
- 64 focused local tests passed, covering Gateway requests, warm callback timing, missing stores/history tables, committed rows with retained WAL, admission changes during preparation, callback failures, and snapshot cleanup.
- Required changed-file checks passed, including core/test types, lint, import/dead-export checks, and native state-schema guards. The existing Gateway test-root cap remains unchanged.
- Independent P0–P2 review found no actionable findings.
- The configured [Blacksmith Linux proof](https://github.com/openclaw/openclaw/actions/runs/34426184910) also passed all 64 tests through the repository wrapper (exit 0); the one-shot lease completed after cleanup.

### DIFF
--- a/docs/cli/update/status-and-history.md
+++ b/docs/cli/update/status-and-history.md
@@ -87,6 +87,10 @@ fields and adds optional `activeRun` and `lastRun` records. While a run is activ
 the Gateway broadcasts `update.run.changed` with `runId`, `phase`, `status`, and
 `updatedAtMs`. Reconnect and read the row to recover changes missed during restart.
 
+When a history request needs a read-only snapshot, the Gateway prepares it
+asynchronously so other requests can continue. The snapshot preserves the source
+database and its sidecar files.
+
 Native service-stop observations do not advance the update's recorded phase.
 If the Control UI cannot read fresh progress, it shows the read error alongside
 the last recorded run; use **Check status** to retry without starting another update.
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [961468de2fbfac969454b3d186dced9da592b5ac](https://github.com/openclaw/openclaw/pull/143605/commits/961468de2fbfac969454b3d186dced9da592b5ac)
- Landed commit: [f2ccf21b43288810a2a5f2dae5cb64c3709e3e9a](https://github.com/openclaw/openclaw/commit/f2ccf21b43288810a2a5f2dae5cb64c3709e3e9a)

PR #143486, opened 2026-09-09
feat(ui): show sender identities in shared chat rail previews

vyctorbrzezowski:
Closes #143473

## What Problem This Solves

Today, chat position rail previews show the role label “User message,” while shared conversation transcripts can also identify the sender by name and avatar. Including that existing identity in the preview lets participants recognize who sent a turn while scanning the rail.

## Why This Change Was Made

The open preview now uses the transcript's normalized sender and existing compact author avatar, including authenticated images and initials. Messages without sender attribution keep the existing role label. Avatar initials are excluded from the marker's accessible description.

Related: #142227 and #143296 by @vyctorbrzezowski. The separate change to which turns appear on the rail is not included here.

## User Impact

Participants can recognize who sent a turn directly from its preview. Standard installations without per-message identity continue to show “User message.”

## Evidence

The sender-attribution boundary test failed on the original code and passed after the change. The focused rail and shared avatar lifecycle suites passed all 16 tests. Browser checks covered switching between two senders, an uploaded image and initials, Escape dismissal, clean accessible descriptions, opening the actual attachment, and navigation to an unattributed session.

UI production and selected test typechecks, targeted TypeScript/CSS lint, i18n verification, formatting, coercion guards, and unused-export scans passed. The combined changed check expanded into unrelated core test graphs and was stopped; the remaining UI checks were run directly.

These are inspected captures of the real Control UI with a mocked Gateway, synthetic participants, and the app's Instrument Sans font. Each desktop matrix reads left to right, top to bottom: own short/long turns without attachments, own short/long turns with attachments, the same four cases from another participant, then an unattributed session. The 390px matrices show sequential scroll samples of the existing mobile transcript, followed by the unattributed session: the rail stays hidden because that layout has insufficient side gutter.

<table>
<thead><tr><th>Scenario</th><th>Before</th><th>After</th></tr></thead>
<tbody>
<tr><td>1440px, dark</td><td><a href="https://github.com/user-attachments/assets/024e16f3-da30-4091-97dc-384e85b352ba">Desktop dark before</a></td><td><a href="https://github.com/user-attachments/assets/2c974c3c-2d00-4bea-82bd-d3495bdde479">Desktop dark after</a></td></tr>
[...]

PR #143296, opened 2026-09-09
fix(ui): group chat position markers by assistant run

vyctorbrzezowski:
Related: #143290

## What Problem This Solves

Fixes an issue where users navigating the chat position rail would see several markers for one assistant run when the loaded history starts after that run's user message. A run with three responses separated by tool calls could appear three times. The current marker could also disappear while reading a later response in the same run.

## Why This Change Was Made

The rail now keeps one marker per visible user message and one per assistant run. It uses the first displayed response as the jump target and the latest displayed response for the preview. Tool calls, results, thinking, and status alone create no marker. A streaming marker keeps its identity and focus when the response is saved; an empty failed run creates no marker.

The marker index reuses visibility prepared by transcript grouping, avoiding repeated normalization of loaded history. All visible responses in a run contribute to its current-position highlight. Message jumps replace the virtualizer's row target with the selected bubble's position, so returning from a long continuation reaches the first response reliably. This builds on the compact rail in #142227 by @vyctorbrzezowski.

## User Impact

Users can navigate questions and assistant runs without duplicate tool-related entries, keep their place through long responses, and retain user/assistant previews, assistant image previews, and shared-participant attribution.

## Evidence

- Rebased onto main at `2d87daf04b78992382299d552f15ac84a86a23ed`, reconciling #146883 by @vincentkoc with assistant-run grouping. The native mutation observer ignores streamed text edits and still reconciles mounted, removed, and renamed bubble targets, including provisional stream messages.
- The original regression from #146883 is unchanged. The streaming handoff regression additionally covers provisional bubble observation, removal/remount, identity changes, and zero bubble-target rescans while text changes.
- Current validation: 6,002 unit tests passed across 267 chat files, and all four rail E2E scenarios passed in dark/light Chromium, including returning from a long continuation. The production UI bundle built successfully.
- Additional verification on this head reported 9,519 passing cases with no test assertion failures. The broad run's parent launcher exited 143; its surviving Vitest child completed 9,489 passing cases, but an exit 0 for that entire launcher chain was not confirmed. The other three invocations exited 0.
[...]

PR #143611, opened 2026-09-10
feat(ui): complete chat rail keyboard navigation

### LINKED ISSUE #143604: [Feature]: Complete keyboard navigation for the chat position rail
### Summary

Complete keyboard navigation for the chat position rail: a single Tab stop, current-message reentry, and Escape back to the transcript.

### Problem to solve

A rail with N sequential Tab stops makes keyboard users cross every message marker before reaching the next control. The desired interaction is one composite Tab stop with arrow navigation inside it, while preserving direct keyboard jumps to messages.

Current `main` already provides one roving Tab stop, arrow/Home/End navigation, and native Enter/Space activation. Browser verification at `c98e00b4f94` found the remaining gaps: leaving and reentering the rail keeps the last focused marker instead of the current reader position, and Escape only dismisses the preview without returning focus to the transcript. This is an accessibility/navigation improvement, not a report that current main has N Tab stops.

### Proposed solution

Keep one Tab stop at the current message, falling back to the first marker when there is no current message. Inside the rail, ArrowUp/ArrowDown move focus, Home/End move to the endpoints, and Enter/Space jump to the focused message. Tab and Shift+Tab leave the rail normally. Escape returns focus to the transcript, the destination of rail activation. Preserve native buttons, labeled list semantics, aria-current, visible keyboard focus, and focused previews.

Reuse the existing rail interaction owner and preserve pointer, touch, and mobile behavior.

### Alternatives considered

Removing the rail from keyboard navigation would prevent keyboard-only users from jumping directly to messages. Keeping a separate tab stop per marker would make long conversations tedious to traverse. A second roving implementation would duplicate the existing owner.

### Impact

Affected: keyboard users of desktop Control UI chat.
Severity: navigation friction.
Frequency: when returning to the rail after moving through the conversation, or leaving it with Escape.
Consequence: extra navigation and focus left away from the reading surface.

### Evidence/examples

The [WAI-ARIA APG keyboard interface guidance](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex) describes one Tab stop per composite with internal arrow navigation.

An isolated mock-Gateway Chrome/CDP flow with 40 synthetic messages confirmed one tab stop and normal Tab exit in light and dark themes, but stale reentry and no Escape focus return. Existing tests cover arrows and activation, but not the full Tab/reentry/Escape sequence.

### Do you plan to open a PR for this?

Yes, I plan to implement this myself

### Additional information

Related: #143296 changes which messages appear in the rail; this improvement starts from main and keeps that decision separate. Both changes touch the rail component and its E2E test, so conflict review will be needed if either lands first. #143486 also changes preview content. This work is issue and PR only; no merge is requested.

vyctorbrzezowski:
Closes #143604

## What Problem This Solves

Keyboard users returning to the chat position rail land on the last focused marker instead of the current reading position, and Escape leaves focus in the rail after dismissing its preview. Current main already has a single roving Tab stop; this completes that navigation contract rather than claiming every marker currently requires a separate Tab press.

## Why This Change Was Made

Reuse the existing rail interaction owner so focus determines the Tab stop only while inside the rail. On exit, the current message (or first marker) becomes the entry point. Escape returns focus to the transcript that rail activation navigates. A labeled list retains native buttons, and the existing hover selectors follow the list wrappers without changing geometry or pointer behavior.

Based on main, including the sender previews from #143486. Related: #143296 changes which messages appear in the rail and remains separate from this improvement. Both PRs touch the rail component and E2E file, so review those hunks when reconciling whichever change lands later.

## User Impact

Tab enters once and Tab/Shift+Tab leaves once. Arrow keys and Home/End move focus and show previews; Enter/Space activate the chosen message. Escape returns to the transcript. Mouse/touch behavior and mobile visibility remain unchanged.

## Evidence

- The original browser regression failed on current-message reentry, then passed on the previous head in light and dark. Those production-UI E2E runs covered one Tab entry/exit, arrows, Home/End, Enter/Space, Escape focus return, previews, nested scrolling and mobile hiding. The attached captures and recordings come from that head and remain useful keyboard-flow evidence.
- Rebased onto current main, preserving both adjacent documentation paragraphs and the existing roving-focus implementation. The maintainer confirmed the single-Tab-stop direction.
- Investigated the original [checks-ui failure](https://github.com/openclaw/openclaw/actions/runs/34427299983/job/102715515796). The terminal replacement-client test began checking the exited tab before awaiting the lazy module that releases reconnect restoration. With a temporary 1500 ms delay in the real module import, its original assertion failed with the same undefined status; waiting for dynamic imports made it pass. The delay was removed. Assertions, mocks and timeouts are unchanged. This demonstrates the preparation race; the historical CI log does not measure import duration.
- All 51 terminal, reconnect and restore tests pass locally with the final repair. Formatting and git diff --check pass. Local rail validation could not load main's newly added markdown-it-cjk-friendly dependency, and broad local test typechecking encountered installed proxyline 0.3.11 where main requires 0.3.12. No dependencies were installed. Targeted TypeScript/style lint, i18n and the initial changed-check guards passed; the full local check:changed command did not pass.
- Fresh independent autoreview through P2 completed scoped-clean, with no findings. The [complete CI run](https://github.com/openclaw/openclaw/actions/runs/34569206182) for head afb835d2f39900249ddac0b7ef1c0d71708ceb99 passed: all three checks-ui shards, all seven UI E2E shards, real-Gateway E2E, typechecks and the required CI gate. Existing browser proof uses synthetic data and a mock Gateway; screen-reader-specific behavior and live-Gateway rail behavior were not exercised.

Before - light theme focused preview

After - light theme ArrowDown focus and preview

Before - dark theme focused preview

After - dark theme ArrowDown focus and preview

https://github.com/user-attachments/assets/acc3014b-58a7-4374-bb8f-076f2b6fc27e

https://github.com/user-attachments/assets/ba341241-f095-4e0e-a8e9-05a0e8c74a94

### DIFF
--- a/docs/web/control-ui/chat.md
+++ b/docs/web/control-ui/chat.md
@@ -311,6 +311,13 @@ change to stored history. Live turns, search results, and turns without an answe
 stay expanded. User messages,
 forwarded inputs, and structural markers remain boundaries for grouping.
 
+On wide desktop panes, the conversation position rail provides keyboard shortcuts
+to messages. Tab enters at the current message, or the first marker if no message
+is current. ArrowUp and ArrowDown move focus; Home and End go to the first and last
+markers. Enter or Space jumps to the focused message. Tab or Shift+Tab leaves the
+rail in one step, and Escape closes the preview and returns focus to the transcript.
+Focusing a marker also shows its preview without jumping to the message.
+
 The chat transcript uses a centered readable frame aligned with the composer. Assistant and tool output stay left-aligned while your own messages stay right-aligned inside that frame. In multi-user sessions (for example a group chat relayed from a channel plugin), messages from other attributed participants render left-aligned with the author's avatar, name, and a stable per-identity color, so only the signed-in viewer's messages read as "mine". When two or more attributed participants are present, assistant replies carry a small "Replying to name" marker naming the participant whose message triggered the turn. System entries such as local slash-command output render as centered notice rows without an avatar.
 
 Images and video previews in your own messages appear above any accompanying text, without a surrounding bubble background. Videos use a still frame with a play icon; select the preview to open the video in the Files panel. If a preview cannot load, the attachment card remains available. Hovering media leaves that layout unchanged, and the text keeps its normal bubble color, including any per-identity tint. Assistant videos retain their inline player.
--- a/ui/src/components/terminal/terminal-panel.test.ts
+++ b/ui/src/components/terminal/terminal-panel.test.ts
@@ -977,6 +977,8 @@ describe("OpenClawTerminalPanel", () => {
     });
     panel.client = newClient;
     await panel.updateComplete;
+    // Reconnect restoration waits for the lazy service-worker refresh module.
+    await vi.dynamicImportSettled();
 
     await waitForFast(() => {
       expect(panel.renderRoot.querySelector(".tabstrip-tab__status")?.textContent).toBe("exited");
--- a/ui/src/e2e/chat-position-rail.e2e.test.ts
+++ b/ui/src/e2e/chat-position-rail.e2e.test.ts
@@ -111,6 +111,48 @@ suite.define(() => {
           expect(await rail.locator('[role="status"]').count()).toBe(0);
           await captureUiProof(suite, page, "chat-position-rail", "idle.png");
 
+          // Exercise the composite through real Tab navigation, including reentry.
+          const focusedMarkerId = () =>
+            page.evaluate(() => document.activeElement?.getAttribute("data-position-marker-id"));
+          const currentMarkerId = () =>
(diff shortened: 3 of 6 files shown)


vyctorbrzezowski:
@clawsweeper re-review

vyctorbrzezowski:
Merged via squash.

- Prepared head SHA: [afb835d2f39900249ddac0b7ef1c0d71708ceb99](https://github.com/openclaw/openclaw/pull/143611/commits/afb835d2f39900249ddac0b7ef1c0d71708ceb99)
- Landed commit: [664094283827a8e7b67501f9533e99f3dfd6ee12](https://github.com/openclaw/openclaw/commit/664094283827a8e7b67501f9533e99f3dfd6ee12)

PR #143620, opened 2026-09-10
fix(agents): avoid repeated subagent tree refreshes

### LINKED ISSUE #143389: [Bug]: WhatsApp /new stalls ~82s on retained subagent tree; per-node scope.refresh repeats full traversal
### Bug type

Crash (process/app exits or hangs)

### Beta release blocker

No

### Summary

WhatsApp `/new` repeatedly stalled the gateway for approximately 80–103 seconds; profiling traced the hot path to repeated full-tree refreshes in `killSubagentRunTree` while traversing retained, already-ended subagent runs, and removing the per-node refresh reduced the observed reset to seconds.

### Steps to reproduce

Observed production reproduction (not a clean-install fixture):

1. Run OpenClaw 2026.9.3 with a WhatsApp direct session that has accumulated retained subagent runs. The affected session had 184 root runs selected by requester/controller association, all with `execution.endedAt` set; the running task list was empty during the profiling test.
2. Send `/new` through WhatsApp to that session.
3. Observe delayed reset acknowledgement, gateway main-thread CPU activity, delayed heartbeat, and a subsequent WhatsApp/WebSocket timeout.

The exact steps to create equivalent retained registry state from an empty installation are not established. An isolated reset with copied SQLite stores but without the equivalent live process-local registry did **not** reproduce the stall.

### Expected behavior

Reset should not require a full-tree scan for every already-ended node. After the one-line local change below and a gateway restart, a real WhatsApp `/new` completed in seconds according to the operator, instead of the previously measured approximately 82 seconds. The post-patch result is qualitative user confirmation, not a precisely instrumented latency measurement.

### Actual behavior

Repeated WhatsApp `/new` attempts took about two minutes. A profiled attempt measured 82.020 seconds total, including 81.231 seconds in `init_session_state`. Another recorded attempt measured approximately 102.6 seconds. Normal messages could respond without this reset-specific delay.

The Control UI reset on the same logical session was reported to return immediately; that observation was not separately profiled and does not establish identical internal paths.

### OpenClaw version

2026.9.3 (1391f7c)

### Operating system

Ubuntu 24.04.4 LTS on WSL2, Linux 6.18.33.2-microsoft-standard-WSL2, x86_64. Node v26.8.1.

### Install method

Global npm installation, gateway managed by a systemd user service.

### Model

openai/gpt-5.6-terra. The delayed `/new` completed as an inline command without a model call.

### Provider / routing chain

OpenClaw -> OpenAI, OAuth/native Codex WebSocket transport for subsequent ordinary messages. The reset delay occurred before inference.

### Logs

Derived, redacted timing summary (not verbatim raw log lines):

```text
Profiled reset:
  resolver total                    82020 ms
  init_session_state               81231 ms
  heartbeat overdue                78254 ms
  freeze/gap warning               72793 ms

Earlier reset:
  resolver total                  102568 ms
  reply.init_session_state        101954 ms
  channel ingress open            102663 ms
```

WSL-side samplers continued while the gateway main thread consumed roughly one CPU core. The captured V8 main-thread profile identified `refreshTree`, `ownerCurrent`, `canTraverse`, registry snapshot matching, and exact-session reads in the hot call stacks. Some native ticks were unresolved; this is not a complete native-stack attribution.

Raw profiles, databases and conversation logs are not attached because they contain private runtime data.

Relevant source: `src/agents/subagents/registry/subagent-control-kill.ts`, function `killSubagentRunTree`. The installed bundle was `subagent-control-kill-B6wJbHXK.mjs`.

The same per-node refresh is present in `main` as inspected when filing (source blob SHA `b43efcaeb3cc353ab84064a2b4adcfa9f6e6f292`). This is source inspection, not a reproduction against a new build of main.

Candidate change applied locally:

```diff
 if (result.descendants && tree.canTraverse()) {
-  params.scope.refresh();
   await Promise.all(tree.children.map(visit));
 }
```

`scope.refresh()` traverses the entire selected tree. Calling it inside each node visit repeats ownership/registry/session work across the same nodes. Before/after-pass refreshes and the selected-count convergence loop remain intact, as do kill-time/retarget refreshes and owner, generation, session-incarnation, retirement and reservation checks. No retained history was deleted.

An isolated Node VM harness executed the actual extracted original/candidate `killSubagentRunTree` functions with explicit doubles for `scope`, tree nodes and `killLatestSubagentRun`. Eight paired cases produced equal kill calls and result objects:

| Case | Original full-scope refresh calls | Candidate |
| --- | ---: | ---: |
| 184 terminal root nodes | 186 | 2 |
| Nested active and yielded runs | 5 | 2 |
| Late descendant discovered after an await | 8 | 4 |
| Ownership revoked during kill | 2 | 2 |
| Superseded run excludes descendants | 2 | 2 |
| Kill exception excludes descendants | 2 | 2 |
| Mocked successor-retarget refresh discovers children | 9 | 5 |
| Traversal ownership initially invalid | 2 | 2 |

These are traversal tests with mocked boundaries, **not** full concurrency/SQLite integration coverage. Syntax checks passed. After deploying the candidate and restarting, WhatsApp was connected/healthy and the operator confirmed `/new` returned in seconds. This supports the hot-path diagnosis, but the patch still needs maintainer review and upstream lifecycle/concurrency tests.

### Impact and severity

- Affected: the observed WhatsApp direct session and gateway responsiveness during its reset.
- Severity: temporarily blocks interaction and causes heartbeat/transport timeouts.
- Frequency: repeated across multiple `/new` attempts and gateway restarts on the affected installation; not established across other installations.
- Consequence: delayed reset acknowledgement and transient WhatsApp disconnection. No demonstrated data loss.

### Additional information

No last-known-good version or introducing commit has been isolated; this report does not claim a version bisect. Restarting/relinking alone did not resolve the repeated delay before the code change. There is no evidence here of SQLite corruption or an actively running editorial subagent causing the stall.

Please review whether the per-node full-scope refresh can be removed as above, or whether a narrower refresh is needed to preserve a lifecycle invariant not covered by the boundary-double tests.

zhangguiping-xydt:
## What Problem This Solves

Fixes #143389. Resetting a session with many retained subagent runs repeatedly refreshed the entire selected tree once per traversable node, stalling the gateway during `/new`.

## Why This Change Was Made

The cancellation owner already refreshes before each convergence pass and when a successor is rebound. The per-node refresh duplicated that whole-tree discovery without changing the traversal contract, so it was removed at the producer of the redundant work.

## User Impact

Session resets avoid repeated registry/session traversal while retaining late-child discovery, ownership checks, and convergence behavior.

## Evidence

- Current head: `52a038620381c5f8b46e07d9f4818fe809615817`.
- Production diff: one deletion in `src/agents/subagents/registry/subagent-control-kill.ts` (`0 additions, 1 deletion`); `git diff --check` passes.
- GitHub CI for this exact head is green, including `openclaw/ci-gate`, `check-prod-types`, lint, test types, dependency guards, security-fast, and PR context/evidence checks: [CI run](https://github.com/openclaw/openclaw/actions/runs/34428923937).
- The change preserves the existing refresh boundaries: initial selection refresh, successor-retarget refresh, and the before/after convergence refresh loop remain unchanged. The per-node refresh was the only removed operation.
- The linked production report recorded 184 retained terminal roots and an 82.020 s reset, with 81.231 s in session initialization; after deploying this identical deletion and restarting, the operator reported a reset completing in seconds. The report also records eight extracted-function paired cases covering terminal roots, nested active/yielded runs, late descendants, ownership revocation, retargeting, exceptions, and invalid traversal ownership.

## Validation

- GitHub CI passed on the exact PR head.
- Source-level lifecycle boundaries and the linked real-operator after-fix observation were reviewed by ClawSweeper.
- Local focused Vitest could not be executed because this environment has Node 24.13.1 while the checkout requires Node >=24.16 or >=26.1; a retry with shared dependencies hit a Vitest `defineCacheKeyGenerator` version mismatch.

## Validation limits

A fresh WhatsApp reset transcript and before/after RSS capture are not available from this environment. The linked issue contains the real operator result and retained-run conditions; no private runtime data is included here.

### DIFF
--- a/src/agents/subagents/registry/subagent-control-kill.ts
+++ b/src/agents/subagents/registry/subagent-control-kill.ts
@@ -383,7 +383,6 @@ async function killSubagentRunTree(
         result.descendants = true;
       }
       if (result.descendants && tree.canTraverse()) {
-        params.scope.refresh();
         await Promise.all(tree.children.map(visit));
       }
     } catch (error) {


zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
## Closing as fixed upstream — this exact change is already on `main`

This PR is blocked on runtime proof of a performance win, so before building the harness I checked whether the targeted call still exists on `main`. It does not — and the diff that removed it is byte-identical to this PR.

### The PR's entire diff

```
src/agents/subagents/registry/subagent-control-kill.ts
index b43efcaeb3c..1c85f05ddfd 100644
@@ -383,7 +383,6 @@ async function killSubagentRunTree(
         result.descendants = true;
       }
       if (result.descendants && tree.canTraverse()) {
-        params.scope.refresh();
         await Promise.all(tree.children.map(visit));
       }
```

### Where it landed

`git log origin/main -- src/agents/subagents/registry/subagent-control-kill.ts` points at `03e52642753` — *"fix: keep session reset responsive and stop yielded child work"* (PR #143916, merged 2026-09-10T10:16:42Z). Its hunk for this file is the same removal, with the **same blob pair**:

```
index b43efcaeb3c..1c85f05ddfd 100644   <- #143916
index b43efcaeb3c..1c85f05ddfd 100644   <- this PR
```

Identical pre- and post-image blobs means the two patches are byte-equal, not merely similar.

### Current state of `main`

`killSubagentRunTree` on `main` (638 lines) contains no `params.scope.refresh()` in the traversal branch:

```ts
      if (result.descendants && tree.canTraverse()) {
        await Promise.all(tree.children.map(visit));
      }
```

The only surviving `refresh()` calls are the two convergence checks of the `do/while`, which this PR never touched:

```
408:    selected = params.scope.refresh();
413:  } while (params.scope.refresh() !== selected);
```

### Why closing rather than rebasing

- The diff against `main` is now **empty** — there is nothing left to apply, so there is no runtime behaviour left to prove. The `needs proof` gate is moot, not unmet.
- #143916 opened ~8h after this PR, landed the same one-line removal, and covers the broader intent (session-reset responsiveness) plus the yielded-child-work path, so the motivation is fully addressed upstream.
- Rebased, this PR would be a no-op commit.

Happy to reopen and rebase if there is a reason to keep it as the tracked fix.

PR #120854, opened 2026-08-09
feat(mattermost): isolate progress posts and deliver finals separately

aolea:
Related context: #109471
Related context: #114589
Builds on: #109555

## Current-head refresh — September 5, 2026

- Rebased feature head: [0270094e6c25](https://github.com/openclaw/openclaw/commit/0270094e6c250664df62ea88446d360ef3723d37), on upstream integration base [9dd869ec98eb](https://github.com/openclaw/openclaw/commit/9dd869ec98ebd7ac08295f549856ab168d483261).
- Both review findings are repaired at their owner boundaries: default in-place source cleanup remains installed, and separate finals bypass progress-only receipt settlement so the durable sender still runs.
- Local verification on the current integration base: focused real-stream regression passed; boundary suites 123/123; Mattermost extension suite 892/892; canonical changed-file repository gate passed, including production/test/plugin type graphs, lint, formatting, generated contracts, import cycles, and policy guards.
- GitHub exact-head CI is fully green with zero failures and zero pending; GitHub reports the PR mergeable with a clean merge state.
- Prior canary package: [canary-mattermost-separate-progress-bd6d8288-r1](https://github.com/aolea/openclaw/releases/tag/canary-mattermost-separate-progress-bd6d8288-r1). Clean Node 22 Linux installation and runtime-entry smoke passed.
- Prior isolated single-agent canary: the package remained Ready with zero restarts; peer runtimes were left unchanged.
- Prior canary runtime identity: OpenClaw 2026.9.1 (bd6d828), bundled Mattermost 2026.9.1, and trusted official Codex harness. A no-delivery exact-model smoke completed in 7.1 seconds with provider openai, model gpt-6-astra, and payload ASTRA_CANARY_OK.
- [ClawSweeper reviewed the exact current head](https://github.com/openclaw/openclaw/pull/120854#issuecomment-5229807960), accepted the live proof, confirmed both delivery findings repaired, and found no actionable correctness or security issue. The PR is ready for maintainer review; sponsorship of the opt-in configuration/API contract remains the product decision.

## Summary

This PR adds an opt-in Mattermost lifecycle that separates transient agent progress from authoritative final replies.

While an agent works, OpenClaw creates and edits one typed `custom_openclaw_progress` post. When the turn completes, OpenClaw sends the final answer as a fresh normal Mattermost post and deletes the progress post only after final delivery is confirmed.

[...]

PR #143621, opened 2026-09-10
fix(mattermost): keep replacement previews through retraction

steipete:
## What Problem This Solves

Fixes an issue where Mattermost users could lose a replacement progress preview when a plan was cleared and replaced before its old post finished deleting. Repeating the same plan text could also suppress the replacement entirely.

## Why This Change Was Made

Retire the old preview generation before yielding, keep its deletion bound to its captured post owner, and release the compositor's deduplication state before accepting a replacement. The existing cleanup queue also retries retired posts at turn completion while preserving the current final post; even a failed flush must wait for earlier deletions before releasing that queue.

## User Impact

Replacement progress remains visible through overlapping retraction and publication, including identical text. Final answers stay intact during cleanup. If a retired post still cannot be deleted, the existing verbose warning records that failure. The Mattermost documentation describes this behavior.

This adds no configuration or exported SDK symbol. The deletion loop is shared rather than copied; production grows 42 lines to represent generation ownership, stale-publication fencing, and cleanup ordering. Related: https://github.com/openclaw/openclaw/pull/120854 proposes opt-in separate final delivery; this repair applies to the existing preview lifecycle.

## Evidence

- **51 focused cases pass:** 13 shared lifecycle cases and 38 Mattermost cases. The latter exercise the actual compositor, stream, client-request boundary, provider-post map, and in-place final-delivery adapter with synthetic data.
- Against the original source, four retraction/publication cases failed while 31 controls passed. Separate tests then exposed and fixed two cleanup cases and one queue-order case introduced during the repair; those results are not counted as original-source failures or additional unique coverage.
- The queue-order regression observes two overlapping physical deletions before its correction, then one, while preserving the original flush error and synchronous final marking. Final-delivery cases cover successful and failed retired-post cleanup without deleting the accepted final.
- Fresh independent Codex review found no actionable findings through P2 on the completed six-file patch.
- All 34 canonical changed-file checks now pass on the corrected source, including the complete core, test, plugin, and plugin-test type graphs and both type-aware lint lanes. The earlier sibling attempt executed zero tests because of `ENOSPC`; that failure is retained.
- Initial CI found three redundant default type arguments and a Promise returned through a callback typed as `void`. The correction uses a typed settlement result to preserve the proved ordering and original error; all 51 tests and fresh incremental review pass. A separate npm-lock job failed during the Matrix native dependency download before lock validation. The corrected head `e6f0a26636c85ff3827855f403871e254213f11a` passed [CI run 34431828115](https://github.com/openclaw/openclaw/actions/runs/34431828115), including the required CI gate.

No live Mattermost account, model, or Gateway was used. All six changed paths remain unchanged on the refreshed main at `fe5d90ba6ff9fe085e4ceafabde36031f6468c3d` before landing.

## Review Disposition

Accepted the shared cleanup tradeoff identified by ClawSweeper on the current head: `stop()` joins preceding cleanup and retries previously failed retired-post deletions, so it may take longer and invoke those deletion callbacks again. This follows the lifecycle's existing ownership of retained failed deletions; the SDK migration guide already describes that ownership. Mattermost is the only in-repository production caller. Existing signatures, synchronous final marking, the current final post, and the original flush error remain protected. The shared lifecycle tests exercise deletion serialization and callback failure, and the Mattermost boundary cases exercise retired cleanup with final delivery. External plugins and live Mattermost were not exercised. No actionable source finding remains from that review.


steipete:
Merged via squash.

- Prepared head SHA: [e6f0a26636c85ff3827855f403871e254213f11a](https://github.com/openclaw/openclaw/pull/143621/commits/e6f0a26636c85ff3827855f403871e254213f11a)
- Landed commit: [84366178223e918748173fba477a1e92c20733c1](https://github.com/openclaw/openclaw/commit/84366178223e918748173fba477a1e92c20733c1)

PR #143626, opened 2026-09-10
fix: retain copied engine registries through cleanup

steipete:
## What Problem This Solves

Resolves a problem where an agent using a copied plugin context engine could fall back unexpectedly after the original registry retired, even though its prepared registry still kept the engine's resources alive. Cleanup could also close the prepared registry while a copied engine's asynchronous disposer still needed one of its provider callbacks.

## Why This Change Was Made

Logical engine construction now captures the supplying registry once and retains its managed resource source for each factory and engine lifetime. Successful donor acquisition records weak one-hop source coverage, so a copied engine can use its live covering source without reopening the retired donor's authority. An uncovered donor still receives its own claim, and partial acquisition failures use the existing cleanup path.

The supplying registry is retained independently for each factory: finishing fallback cleanup cannot release resources that the selected engine still needs. Cleanup releases the primary before an uncovered donor, including when primary disposal fails. Raw supplying registries preserve their existing direct-donor ownership behavior. This adds no SDK contract, cache, schema change, or transitive dependency graph.

## User Impact

Prepared agent runs keep their selected copied context engine and can finish asynchronous cleanup safely. Retiring the supplying registry still prevents new logical turns; this change extends physical cleanup custody without reviving execution authority. Raw registrations retain their documented caller-owned lifetime.

## Evidence

Native SQLite regressions use the real prepared-registry acquisition path and a full-registration donor. Original source fails three cases: retired-donor selection and supplying-registry lifetime with live and raw donors. The candidate passes all three, plus uncovered and failed-dependency controls. A raw supplying-registry control also passes on the original source.

The final candidate passes 62 distinct tests across context-engine ownership, prepared-registry integration, donor resources, and inspection resources. The full changed gate passed in 371.80 seconds, followed by focused checks for the final test-only control. After the one-line weak-membership refinement, all 48 affected lifecycle controls and core type, targeted lint, and formatting checks passed again.

Independent Codex review found no actionable P0–P2 issues. The normal build passed in 235.76 seconds. Three compiled probes then passed against the same 11,400 sealed build artifacts, using the actual prepared-registry acquisition path and public speech-provider callback. All six SQLite files reopened with the expected values; all managed child processes exited, and 3,948 resolved modules contained no TypeScript source fallback. Engine resolution took 1.12–1.24 ms and disposal 0.97–1.54 ms in these one-shot synthetic probes; these are smoke timings, not a throughput comparison.

### DIFF
--- a/docs/concepts/context-engine.md
+++ b/docs/concepts/context-engine.md
@@ -91,6 +91,13 @@ runtime alive through maintenance, coalesced reruns, and engine disposal. Accept
 does not mean cleanup has finished. Return asynchronous work from engine methods
 and `dispose()` so the host can join it before releasing their resources.
 
+A logical turn also retains its managed supplying registry through engine disposal.
+When that registry copied a runtime engine from another inspection, its recorded
+donor dependency can keep the engine usable after the donor inspection retires.
+Retiring the supplying registry still refuses new logical turns; existing engine
+work keeps its physical resources until cleanup finishes. Raw registrations keep
+their caller-owned lifetime.
+
 For the bundled non-ACP Codex harness, OpenClaw applies the same lifecycle by projecting assembled context into Codex developer instructions and the current turn prompt. Codex still owns its native thread history and native compactor.
 
 ### Subagent lifecycle (optional)
--- a/src/context-engine/registry.copied-view-resources.test.ts
+++ b/src/context-engine/registry.copied-view-resources.test.ts
@@ -1,0 +1,328 @@
+import fs from "node:fs";
+import { createRequire } from "node:module";
+import path from "node:path";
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ebfb5910edda7eeb79b202022c48c1208f1c0f69](https://github.com/openclaw/openclaw/pull/143626/commits/ebfb5910edda7eeb79b202022c48c1208f1c0f69)
- Landed commit: [b8054674868f0a389b3b1ae638a79921786b8235](https://github.com/openclaw/openclaw/commit/b8054674868f0a389b3b1ae638a79921786b8235)

PR #143627, opened 2026-09-10
fix: retain summary models through completion cleanup

steipete:
## What Problem This Solves

Resolves a problem where TTS summarization could use a model after its plugin resources had been released, causing completion or late cleanup to fail when the source has a managed lifetime. Ordinary configured sources remain raw today; this closes the summary ownership gap before enabling their physical disposal.

## Why This Change Was Made

The default summary path now acquires the already-selected provider/model and retains it through preparation, actual completion, and cooperating cleanup, reusing the shared prepared-model acquisition lifecycle. The shipped optional injected-dependencies contract remains caller-owned, and model selection, discovery-before-timeout, deadline/result behavior, auth-error classification, and truncate fallback remain unchanged.

## User Impact

TTS summaries can safely finish and clean up when their model source has managed resources, including work that finishes after cancellation. This is lifecycle preparation: it does not enable automatic disposal for ordinary configured sources or change TTS configuration and provider routing.

## Evidence

- With the original production files restored byte-for-byte, six managed-source regressions failed while four raw/injected controls passed. The real SQLite connection was already closed at provider entry. All ten passed with the fix; 130 affected sibling cases also passed, for 140 distinct tests covering selection, setup failures, cancellation, cleanup, and the public injected contract.
- The native tests pass a real acquired read-only SQLite source through the existing acquisition input adapter into the actual public default summary path. They prove late reads, automatic exactly-once closure, and reopening. This adapter supplies managed ownership; it does not prove that ordinary configured producers already acquire managed sources.
- The full changed-file gate passed, including types, Knip, lint, formatting, database guards, and import-cycle checks. Additional static Madge found zero cycles. Fresh isolated Codex review found no actionable P0–P2 findings. Ordinary `pnpm build` passed in 261.66 seconds, including all 152 public SDK subpaths.
- A compiled synthetic probe passed in 1.99 seconds through public speech-core, TTS config resolution, and prepared-completion SDK entry points. It verified explicit summary-model selection on all four provider calls, cleanup before parent shutdown, late completion after the request signal expired, and repeated injected calls using one SDK-prepared model. All 11,694 built artifacts remained unchanged; a guard rejected TypeScript source fallback. No external inference was used.
- The compiled fixture passed the actual built config schema before execution. A final audit verified canonical config fields, matching provider registration, sealed probe/schema/log hashes, and no invalid-config or fallback diagnostics. No checkpoint or session-store behavior is claimed.
- Compiled default sources remained raw under current production policy: SQLite stayed open after summary/SDK shutdown, and the fixture owner then closed and reopened it. Automatic default-source disposal still needs compiled integration proof when that producer is activated. Observed cold-summary and warm deadline-signaled completion times were 200.47 ms and 32.26 ms respectively; these are individual synthetic observations with different work, not a benchmark or speedup claim.

### DIFF
--- a/src/agents/simple-completion-runtime.ts
+++ b/src/agents/simple-completion-runtime.ts
@@ -522,6 +522,31 @@ async function withPreparedSimpleCompletionRuntime<T>(
   }
 }
 
+type AcquiredSimpleCompletionModel =
+  | (Extract<PreparedSimpleCompletionModel, { model: Model }> & { release: () => void })
+  | Extract<PreparedSimpleCompletionModel, { error: string }>;
+
+/** Acquire the exact provider/model already selected by a finite internal caller. */
+export async function acquireSimpleCompletionModel(
+  params: Omit<Parameters<typeof prepareSimpleCompletionModel>[0], "preparedModelRuntime">,
+): Promise<AcquiredSimpleCompletionModel> {
+  return await acquirePreparedSimpleCompletionModel(
+    params,
+    [
+      {
+        provider: params.provider,
+        modelId: params.modelId,
+        ...(params.agentRuntimeId ? { runtime: params.agentRuntimeId } : {}),
+      },
+    ],
+    (context) =>
+      prepareSimpleCompletionModelCore(
+        { ...params, agentDir: context.preparedModelRuntime.agentDir },
+        context,
+      ),
+  );
+}
+
 type AcquiredSimpleCompletionModelForAgent =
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [93a830a0206a6ceb589a03418b3d032e25979a63](https://github.com/openclaw/openclaw/pull/143627/commits/93a830a0206a6ceb589a03418b3d032e25979a63)
- Landed commit: [fe958097d424125d93cbfc47b42a561ab308e877](https://github.com/openclaw/openclaw/commit/fe958097d424125d93cbfc47b42a561ab308e877)

PR #143628, opened 2026-09-10
test: reuse transforms across native Gateway fixtures

steipete:
## What Problem This Solves

Gateway lifecycle tests launch four fresh native Vitest processes. Each received a new project root/configuration and no filesystem module cache, so startup scenarios repeatedly transformed the same Gateway source graph. Those imports took roughly nine seconds per fork in a phase trace.

## Why This Change Was Made

Give each test file a disposable code/config/cache owner through the existing fixture-lifetime helper. Keep that project identity stable, and use Vitest's supported `--dir` option to select each unchanged private case root. This permits transformed-code reuse while preserving four separate native processes, fresh evaluation, state/configuration, resource owners, journals and PID checks.

The shared owner joins complete borrowers before cleanup. If a child has no verified join receipt, shared code is retained along with its case root. A lightweight native regression verifies that boundary. The session test registers cleanup through its actual afterAll hook, outside the fixture hooks it intentionally intercepts.

All 11 original cases and assertions remain; one ownership regression is appended within an existing file. The existing cache toggle and Windows default-off policy are honored; the outer cache path is never inherited. No production, dependency, timeout, worker, shard, root-limit or cache-key-policy change.

## User Impact

Faster native Gateway startup/failure coverage without combining processes or resetting intentionally retained failed owners. Warm startup forks two and three fall from roughly 15–18 seconds each to 4–5 seconds.

## Evidence

| Same 11-case whole invocation | Time |
| --- | ---: |
| Original | 97.29s |
| Candidate | 71.60s |
| Restored original | 77.54s |

All cases passed in the same order. These include wrapper preparation, the first cold transforms, native execution and cleanup. Local load varied: the conservative candidate improvement over the faster original control is 7.7%. The first cache population adds work; no speedup is promised for the session file's single native fork.

Final validation passed all **12 cases** in **68.88s**, including the added ownership regression. Its file ordering differed from the comparison above, so it is correctness evidence rather than another matched timing sample. Peak command RSS remained around 2.27 GB. The startup cache held about 107 MB of temporary transformed artifacts.

- An early-start observer captured separate original native CLI/worker pairs. After the first startup row, 4,436 cache artifacts retained identical sizes/mtimes across the next row; one new fixture entry appeared, and configuration bytes stayed identical. All original process/journal/liveness assertions passed, and code/cache/case roots were removed after verified completion.
- The ownership regression uses a real lightweight child that omits only its required join receipt. Shared code and case state must remain. Removing only the new failed-cleanup obligation made the test fail because cleanup incorrectly resolved; restoring it passed. A separate probe PID permits test cleanup without repairing the helper's missing receipt.
- A native probe caught an initially lost cache-disable toggle. Forwarding the existing toggle made the same probe pass with no module-cache directory; the temporary diagnostic assertion was removed.
- All 13 existing cache-policy tests passed, including source/config/alias/lock invalidation and fresh evaluation. Required changed-file guards, typechecks, dead-export checks and lint passed. Independent P2 review is clean.

Proof used pinned Node 24.19.0/pnpm 12.3.4, unchanged physical dependencies/build stamps and freshly emptied private benchmark paths. Diagnostic traces and observers are not committed. No Windows performance or full-main CI saving is claimed.

### DIFF
--- a/src/gateway/server.fixture-lifetime.test-support.ts
+++ b/src/gateway/server.fixture-lifetime.test-support.ts
@@ -5,34 +5,28 @@ import os from "node:os";
 import path from "node:path";
 import { expect, type TestContext } from "vitest";
 import { createVitestResourceOwner } from "../../scripts/lib/vitest-resource-ownership.mts";
+import { createFixtureLifetime } from "../../test/helpers/fixture-lifetime.js";
 import { runVitestShutdownCommand } from "../../test/helpers/vitest-shutdown-command.js";
 import { hasErrnoCode } from "../infra/errno.js";
 
-export function runGatewayFixtureFork(
-  context: Pick<TestContext, "signal" | "onTestFinished">,
-  source: (repoRoot: string, root: string) => string,
-  assertJournal: (journal: unknown, text: string) => void,
-): Promise<void> {
-  const run = Promise.resolve().then(async () => {
-    context.signal.throwIfAborted();
-    const repoRoot = path.resolve(import.meta.dirname, "../..");
-    const root = await fs.realpath(await fs.mkdtemp(path.join(os.tmpdir(), "gateway-close-fork-")));
-    let joined = false;
-    try {
-      // Failed child claims belong here, never to the outer worker's resource namespace.
-      createVitestResourceOwner(root);
-      const require = createRequire(import.meta.url);
-      const vitestPackageDir = path.dirname(require.resolve("vitest/package.json"));
+export function createGatewayFixtureFork(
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ced99541bfdb76f91bc32993698ce6c456bd5984](https://github.com/openclaw/openclaw/pull/143628/commits/ced99541bfdb76f91bc32993698ce6c456bd5984)
- Landed commit: [dc060722b83632e2ae5dc031568ca182326e875c](https://github.com/openclaw/openclaw/commit/dc060722b83632e2ae5dc031568ca182326e875c)

PR #143633, opened 2026-09-10
fix: retain compaction services through issued work

steipete:
## What Problem This Solves

Fixes an issue where cancelling or timing out compaction could close MCP/LSP resources while issued compaction work or plugin hooks were still using them. A timed-out hook could also escape the work owner that retained its prepared database resources.

## Why This Change Was Made

Compaction now separates the caller's result from physical cleanup. Its owner waits for actual preparation, compaction, provider, and hook work before disposing tool runtimes, then joins cleanup tails before releasing the prepared generation. Partial setup failures transfer cleanup ownership before allocating resources.

Timed hook dispatch observes the already-started handler promise while preserving the original timeout race and error policy. Session disposal still retires authority at the original boundary; diagnostics, transcript fences, accounting, checkpoints, and skill-environment restoration keep their existing ordering.

## User Impact

Compaction can report cancellation or timeout promptly while allowing already-issued work to finish safely. Normal compaction, native CLI routing, fallbacks, and the default 180-second watchdog remain unchanged. Hook deadlines and fail-open behavior are preserved.

Normal-run registry disposal is not enabled by this change. The database disposal regression uses the existing managed read-only producer at the native delegate input boundary. The compiled smoke uses the public delegate with the current normal-run producer; managed default-flow integration remains required before activating its disposal.

## Evidence

- Validated fixture configuration reproduces ten premature-close failures on the original code; success and raw-source controls pass. All twelve cases pass with the repair. Separate hook tests reproduce early drainage while preserving raw/closed-scope reporting controls.
- 321 focused and sibling tests passed. The twelve delegate cases were rerun after correcting fixture configuration. Required changed checks and the affected fixture checks passed.
- Final independent Codex review found no actionable P0–P2 issues.
- Full build passed in 241 seconds. Compiled public-delegate proof used four loopback model completions, real MCP HTTP sessions, real LSP processes, and both timed compaction hooks. Each service stayed live through the reported result, then closed once; the LSP SQLite files reopened successfully.
- With a declared 25 ms hook timeout, reports arrived 27–57 ms after hook entry and physical cleanup took 48–50 ms after releasing the held work. These are synthetic smoke timings, not throughput benchmarks.

### DIFF
--- a/src/agents/embedded-agent-runner/compact.delegate-resources.test.ts
+++ b/src/agents/embedded-agent-runner/compact.delegate-resources.test.ts
@@ -1,0 +1,607 @@
+import fs from "node:fs";
+import path from "node:path";
+import { DatabaseSync } from "node:sqlite";
+import { getAiTransportHost, type AssistantMessage, type Model } from "@openclaw/ai";
+import { createAssistantMessageEventStream } from "@openclaw/ai/event-stream";
+import { expectDefined } from "@openclaw/normalization-core";
+import { describe, expect, it, vi } from "vitest";
+import { upsertSessionEntryCore } from "../../config/sessions/session-accessor.js";
+import type { ModelDefinitionConfig } from "../../config/types.models.js";
+import type { OpenClawConfig } from "../../config/types.openclaw.js";
+import { validateConfigObjectRaw } from "../../config/validation-core.js";
+import { delegateCompactionToRuntime } from "../../context-engine/delegate.js";
+import { initializeGlobalHookRunner } from "../../plugins/hook-runner-global.js";
+import {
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [34744c330cef945d2a0b571c18d0367fcd2d2fb6](https://github.com/openclaw/openclaw/pull/143633/commits/34744c330cef945d2a0b571c18d0367fcd2d2fb6)
- Landed commit: [3b627adf6c6a9ef7028bb3561e33dea87368dbba](https://github.com/openclaw/openclaw/commit/3b627adf6c6a9ef7028bb3561e33dea87368dbba)

PR #143639, opened 2026-09-10
fix(test): preserve expanded shard timing identities

steipete:
Related: #139428

## What Problem This Solves

Expanded full-suite test chunks lose their selected-file information before scheduling and timing history are recorded. Different chunks are priced as whole projects and overwrite the same timing history, which can leave long projects late in the queue.

## Why This Change Was Made

Carry the planner's known file targets separately as scheduling metadata. The existing cost estimator and timing-key owner use those targets, while execution arguments, include-file filtering, worker budgets, concurrency, and build prerequisites retain their existing owners. Expanded chunks get distinct target-derived keys even under a shared parent label; existing include-file callers keep their named-key behavior.

## User Impact

Cold scheduling and learned timing histories now describe the actual chunks. The CLI weight is unchanged. No runtime speedup or reduction in total test time is claimed yet.

## Evidence

Four focused regressions failed on original code for the intended missing-target, wrong-cost-order, and colliding-key conditions. The candidate passed all 608 tests across the three complete planner, timing, and build-admission owner suites. A subsequent two-line lint correction replaced conditional metadata spreads with direct assignments; all four focused regressions passed again, followed by the complete mapped changed-file gate and fresh independent P2 review.

Read-only native metadata analysis on a coherent source compared 664 specs: execution descriptors and membership were identical, with unchanged outer concurrency 8, requested workers 1, and build prerequisite. Distinct timing keys increased from 96 to 664, and the CLI process project moved from position 581 to 143. These are planning results, not benchmark measurements. The change adds 101 test lines and six tooling lines to protect the repaired contracts.

### DIFF
--- a/scripts/lib/vitest-shard-metadata.mts
+++ b/scripts/lib/vitest-shard-metadata.mts
@@ -13,6 +13,8 @@ export type VitestShardTimingSpec = {
   config: string;
   env?: NodeJS.ProcessEnv;
   includePatterns?: readonly string[] | null;
+  /** Exact chunk files for scheduling; does not configure execution filtering. */
+  timingTargets?: readonly string[];
   watchMode?: boolean;
 };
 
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [04eeb7be68590bc43aca3bddbbe85a4526507c86](https://github.com/openclaw/openclaw/pull/143639/commits/04eeb7be68590bc43aca3bddbbe85a4526507c86)
- Landed commit: [7bd659c308b571e6440bdc196d5d87cdf21ee812](https://github.com/openclaw/openclaw/commit/7bd659c308b571e6440bdc196d5d87cdf21ee812)

PR #143643, opened 2026-09-10
fix(models): preserve captured routes in configured catalog views

obviyus:
Related: #136257.

## What Problem This Solves

Configured model listings could show provider-default routing even when the captured catalog retained a different route, such as reporting a local endpoint as remote.

## Why This Change Was Made

Configured composition now prefers an explicit model route, then the captured route for the exact model identity, then provider defaults. Startup and visibility use that same decision. A per-call identity index removes repeated scans without introducing a persistent cache.

## User Impact

Listings retain captured routes while respecting explicit settings and case-distinct identities. Configuration-only callers retain provider defaults when no captured catalog is supplied. No credential, schema, protocol, or operator-option change is introduced.

Consumers: configured catalog composition, startup publication, and visibility now share captured-route selection.

## Evidence

The original Gateway returned incorrect locality in all three catalog views. Twelve authenticated candidate responses across four route scenarios retained the expected locality and metadata. Source regressions passed, including the case-distinct identity control. A producer benchmark improved from 1,963 to 35 ms at 400 rows; this is not end-to-end latency. Continuous integration passed its 400-model, 11-agent case with no upstream requests. Passive-read checks preserved saved configuration and catalog bytes. Inference and live account sign-in were not tested.

### DIFF
--- a/src/agents/model-catalog-visibility.ts
+++ b/src/agents/model-catalog-visibility.ts
@@ -17,10 +17,7 @@ import {
   resolveConfiguredModelCatalogOverrides,
 } from "./model-catalog-route.js";
 import type { ModelCatalogEntry } from "./model-catalog.js";
-import {
-  buildConfiguredModelCatalog,
-  dedupeModelCatalogEntries,
-} from "./model-selection-shared.js";
+import { dedupeModelCatalogEntries } from "./model-selection-shared.js";
 import {
   RUNTIME_MODEL_VISIBILITY_NORMALIZATION,
   createModelVisibilityPolicy,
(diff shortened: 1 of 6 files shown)


obviyus:
Thanks @obviyus!

**Proof:** Ordinary Gateway startup and authenticated `models.list` reproduce the baseline locality error and verify captured-route, explicit-pin, absent-route and case-distinct controls across all/configured/default views. A separate accepted normal/error supplement records raw state boundaries, actual connection closes and clean Gateway shutdown. Evidence keeps its CI merge-build identity and stated limits.

- **Bug:** Configured catalog views replaced an exact captured route with provider defaults.
- **Fix:** One exact-identity composition rule gives model pins priority, then captured routes, then provider defaults. Visibility reuses that prepared result.
- **Validation:** 91 focused checks, scoped independent acceptance, clean exact-head review with no findings, and successful CI including the 400-model/11-agent case. No rank-up moves remain.

| Before | After |
| --- | --- |
| Captured loopback model reported as remote | Captured locality preserved |
| Missing exact donor could borrow a case sibling | Case-distinct routes and metadata stay separate |

This lands the configured-route contribution to #136257, not the remaining registry/native/auth work or whole original-area closure. No live-provider inference claim is made.

PR #143644, opened 2026-09-10
fix(plugins): avoid installed matches across ClawHub registries

steipete:
## What Problem This Solves

Fixes an issue where users browsing plugins would see a package marked installed when only a same-name package from another ClawHub registry was installed. When ClawHub detail lookup failed, an unrelated npm package name or runtime plugin ID could also select the wrong local details.

## Why This Change Was Made

Installed discovery identities now require a recorded registry matching the effective ClawHub registry. Verified official/npm counterparts retain their public-registry mapping. Browse and offline detail fallback share the same ClawHub identity index, removing the broader alias lookup that made their decisions diverge.

## User Impact

Plugins from other registries remain locally manageable without claiming the current catalog's installed status. Custom registry selection through either supported environment variable, exact local plugin routes, public official counterparts, and valid same-source offline fallback retain their behavior. Records without a known registry remain local rather than acquiring an origin from the current environment.

## Evidence

Six regression cases fail against unchanged production code for the intended reasons: four incorrect registry matches and two unrelated-alias detail fallbacks. The repaired candidate passes all 123 focused tests across discovery, management, Gateway handlers, and Featured/source-linkage coverage, including positive custom-registry and legitimate offline-fallback cases.

Proof uses the real management projection, discovery join and RPC handlers with synthetic metadata and stubbed network boundaries. All five-file changed checks, docs checks and the isolated P0–P2 managed review passed. The change keeps existing external request shapes and UI layout; no live ClawHub request was used for proof.

### DIFF
--- a/docs/plugins/manage-plugins.md
+++ b/docs/plugins/manage-plugins.md
@@ -32,6 +32,10 @@ the full inventory or choose **Show all** to browse every installed plugin. Each
 card shows the plugin description. Choose a card to open its settings, where
 administrators can enable or disable it and read-only operators can inspect it.
 
+A package installed from another ClawHub registry stays manageable under
+**Installed plugins**. It does not mark a same-name package in the current
+catalog as installed.
+
 Choose a card to open `/settings/plugins/<plugin-id>`.
 That routed page uses the plugin's declared schema for configuration, explains
 effective access in plain language, and keeps raw capability declarations and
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [215bd3536002efd9a5843a70acf02cf6df922984](https://github.com/openclaw/openclaw/pull/143644/commits/215bd3536002efd9a5843a70acf02cf6df922984)
- Landed commit: [80a2b36ce445a70d853b1d69bf119102394592af](https://github.com/openclaw/openclaw/commit/80a2b36ce445a70d853b1d69bf119102394592af)

PR #143647, opened 2026-09-10
fix(models): purge plugin catalog credentials on logout

### LINKED ISSUE #142421: [Bug]: models auth logout leaves plaintext credentials in plugin-model-catalog cache; doctor --fix resurrects them
## Summary

Logging out of an `api_key` (or `token`) provider does not invalidate the plaintext `apiKey`/`token` value that was previously written into the per-agent `cache_entries` table under scope `plugin-model-catalog-v1`. The credential silently survives the logout, `openclaw doctor` correctly detects it as an orphan plaintext credential, and `openclaw doctor --fix` then copies that orphan credential **back into** the SQLite auth store — effectively resurrecting the credential the user just deleted. There is also no user-facing command to prune stale cached catalogs.

## Bug type

Behavior bug (incorrect output/state without crash).

## Steps to reproduce

1. On a fresh gateway, configure an `api_key` provider that has a plugin model catalog (e.g. `openrouter` or `github-copilot`) via `openclaw models auth login` / `paste-api-key` / `login-github-copilot`.
2. Trigger a plugin model catalog fetch — either wait for scheduled refresh or run `openclaw models refresh`.
3. Confirm the cached catalog row now contains the plaintext key:

    ```
    sqlite3 <state-dir>/agents/<agent-id>/agent/openclaw-agent.sqlite \
      "SELECT scope, key, LENGTH(value_json), \
       CASE WHEN value_json LIKE '%\"apiKey\"%' THEN 'has-apiKey' ELSE 'no' END \
       FROM cache_entries WHERE scope='plugin-model-catalog-v1' AND key='<provider>'"
    ```

4. Confirm `openclaw doctor` reports 0 plaintext model credentials.
5. Log out of the provider: `openclaw models auth logout --agent <agent-id> <provider>:<profile>`.
6. Confirm the profile is gone from `openclaw models auth list --agent <agent-id>`.
7. Run `openclaw doctor` — it now reports:

    ```
    Found 1 plaintext model credentials. Run openclaw doctor --fix to copy
    and verify them in agent SQLite before plaintext catalog authentication
    is retired.
    ```

8. Run `openclaw doctor --fix` — the deleted profile is re-created in the auth store from the cached copy, verifiable via `openclaw models auth list --agent <agent-id>`. The credential the user intentionally removed is now live again.

## Expected behavior

Any of:

- Logout invalidates cached catalog entries for that provider (removes the row from `cache_entries` with `scope='plugin-model-catalog-v1'` and `key=<provider>`, or rewrites it without the `apiKey` field), so the credential doesn't outlive the auth profile that owned it.
- Or: `openclaw doctor` distinguishes "orphan we should migrate into the store" from "orphan we should evict from the cache," and `--fix` prunes the latter (or at minimum prompts). Silently resurrecting a deleted credential is the wrong default.
- Or: a `openclaw models cache prune` command is documented so users have a supported way to remove stale cached catalog rows without hand-editing SQLite. (`openclaw models refresh` today only refreshes the hosted model catalog and does not accept an agent scope or a prune option.)

## Actual behavior

- Logout leaves the cached catalog row untouched. The `apiKey`/`token` string is still readable via `sqlite3` after the profile is gone.
- `openclaw models refresh` does not evict stale rows for providers that no longer have live auth — the cached row survives indefinitely.
- `openclaw doctor` correctly detects it as a plaintext credential (`collectCredentials` in `src/commands/doctor-model-catalog-credentials.ts:82-109`).
- `openclaw doctor --fix` copies the orphan cached credential into the SQLite auth store via `maybeMigrateModelCatalogCredentials` (same file, lines 285-350). Post-fix, `openclaw models auth list` shows the previously-deleted profile as active again.

## OpenClaw version

`main` (source install), verified against `origin/main` `9982628` (2026-09-08). Config `meta.lastTouchedVersion: 2026.9.2`.

## Operating system

macOS 12.7.6 (build 21H1320, Darwin 21.6.0, x86_64).

## Install method

Source install (`pnpm dev` / bin symlink).

## Model

Not model-specific. Reproduces with any provider that has a plugin model catalog and uses `api_key` or `token` auth (verified with `github-copilot`; the same code path applies to `openrouter`, `nvidia`, `google`, etc.).

## Provider / routing chain

Not involved. This is entirely a local storage + doctor code path.

Cache location: `<state-dir>/agents/<agent-id>/agent/openclaw-agent.sqlite`, table `cache_entries`, scope `plugin-model-catalog-v1`, one row per plugin. Value is a JSON blob with `providers[<provider>].apiKey` (or `.token`) populated from whatever credential was in effect at fetch time.

Auth store location(s): shared `<state-dir>/openclaw.sqlite` under `config_machine_state.value_json` where `state_key='authProfiles.store'`, and/or per-agent `<state-dir>/agents/<id>/agent/openclaw-agent.sqlite` under `auth_profile_store.store_json` where `store_key='primary'`.

## Root cause

Two connected issues in `src/commands/doctor-model-catalog-credentials.ts` (verified against `origin/main` `9982628`):

1. **`collectCredentials` at lines 82-109** walks each plugin catalog's `providers[*].apiKey` and treats every entry not matched by a current profile as a "plaintext credential to migrate":

    ```ts
    return Object.entries(providers).flatMap(([provider, entry]) => {
      if (!isRecord(entry) || typeof entry.apiKey !== "string") return [];
      const key = entry.apiKey;
      const credential = { provider, key };
      if (
        !key.trim() ||
        (cfg && resolveProviderConfigSecretInput(cfg, provider).ref) ||
        isNonSecretApiKeyMarker(key) ||
        store.profiles[key] !== undefined ||
        findMatchingProfileId(store, credential, blockedStores) !== undefined
      ) return [];
      return [credential];
    });
    ```

    An orphan from "user deleted the profile" is indistinguishable from an orphan from "a valid legacy install with credentials only in the catalog." The check has no signal for the former.

2. **`maybeMigrateModelCatalogCredentials` at lines 285-350** treats every detected orphan as migration input and calls `updateAuthProfileStoreWithLock` to write it into the auth store. There is no consideration of "was this profile explicitly removed" — the fix assumes forward migration is always desired.

Meanwhile, the underlying persistence in `src/agents/plugin-model-catalog.ts` (`readPersistedPluginModelCatalogs` at line 86 and its callers) has no invalidation hook wired to auth-profile removal, so `openclaw models auth logout` completes without touching `cache_entries`.

## Impact and severity

- **Affected:** any operator who has logged out of an `api_key`/`token` provider after that provider's plugin model catalog was fetched (which is essentially all real-world usage — refresh runs on gateway lifecycle events).
- **Severity:** Medium-to-high, security-adjacent. Two concrete problems:
  1. Users who log out to remove a credential (a natural "revoke locally" gesture) still have that credential sitting in plaintext in a local SQLite file, invisible unless they run `openclaw doctor` or query SQLite directly. It also survives across gateway restarts.
  2. The single documented remediation for the doctor's warning (`--fix`) actively works against user intent by re-creating deleted profiles from cached copies.
- **Frequency:** Every logout-of-catalog-provider after any catalog fetch. Deterministic.
- **Consequence:** Credential retention beyond intended lifetime; user confusion when `doctor --fix` resurrects revoked profiles; no clean self-service remediation short of SQL.

## Additional information

Concrete real-world evidence: in a personal install, `openclaw doctor` continued to report N stale plaintext credentials from removed `api_key` provider profiles across multiple per-agent SQLites. `openclaw models refresh` did not clear the stale rows. The only viable cleanup was:

```
sqlite3 <state-dir>/agents/<id>/agent/openclaw-agent.sqlite \
  "DELETE FROM cache_entries WHERE scope='plugin-model-catalog-v1' AND key='<provider>'"
```

A minimal fix would be one of:

- Hook `openclaw models auth logout` (and related profile-removal paths) to also invalidate matching `cache_entries` rows for that provider under the `plugin-model-catalog-v1` scope, in the same transaction.
- Change `doctor --fix` to prune orphaned cached credentials rather than migrating them into the auth store, or to prompt distinguishing "migrate" vs "prune."
- Provide a documented `openclaw models cache prune` command so users have a supported cleanup surface without touching SQLite.

Any one of these would resolve the credential-retention half of the bug; both (1) and either (2) or (3) would additionally fix the "fix resurrects deleted credentials" trap.

---

## Edit (2026-09-13)

Confirmed independently on a separate live multi-agent install (source checkout, `main` @ `5ee87a5`), and consolidating what's come in from other reporters below since this seems to be broader than the original logout-only framing.

**Scope is bigger than logout.** Between this issue and the comments, there are now four independent ways this plaintext residue gets stuck with no supported removal path:

1. **Logout** (original report above).
2. **A completed SecretRef migration** (@matthewmoroz): after `secrets configure`/`secrets apply` moves a credential from plaintext `key` to a store-backed `keyRef`, the stale plaintext copy stays in `cache_entries` — and `doctor --fix` then creates a *second* profile from it, undoing the migration.
3. **No logout or migration involved at all** (@sunhsiao, and confirmed again independently on the install checked for this edit): a provider that's SecretRef-managed from the start still gets a plaintext-resolved copy written into the cache at catalog-generation time. `doctor` is blind to this case specifically because the SecretRef guard in `collectCredentials` (`resolveProviderConfigSecretInput(cfg, provider).ref` check) suppresses detection entirely — so this variant isn't even flagged, let alone fixable via `--fix`.
4. **Backup leakage** (@CHAESH): `cache_entries` is absent from the redaction set in `src/state/secret-state-tables.ts`, so `openclaw backup git create --exclude-secrets` still ships the plaintext key, and with `--push` it lands permanently in remote git history.

On the reconfirmation pass: doctor reported "Found 4 plaintext model credentials" on an install with 2 configured agents; inspecting both agents' `cache_entries` (`scope='plugin-model-catalog-v1'`) showed the same 2 stale provider ids present in **both** agent databases with no corresponding `auth_profile_store` rows in either — i.e. the count is `(stale provider catalogs) × (agent dirs scanned)`, confirming `collectAgentCatalogs`/`collectCredentials` sums unmatched entries per-agent independently. One of the two stale entries had **no config entry at all** for that provider anywhere in `openclaw.json` — a fully orphaned residue with no traceable source, reproducing @sunhsiao's point independently.

**New, distinct bug found while digging into this: the CLI logout path has its own config-level gap, separate from the `cache_entries` issue.**

`src/commands/models/auth-logout.ts`'s `modelsAuthLogoutCommand` calls:

```ts
const warning = await removeModelAuthCredentials({ cfg, agentDir, profileIds: [profileId] });
```

— never passing `apiKeyProvider`. Compare the Gateway RPC handler for the same operation, `models.authLogout` in `src/gateway/server-methods/models-auth-status.ts`:

```ts
const configWarning = await removeModelAuthCredentials({
  cfg, agentDir, profileIds: removedProfiles,
  ...(apiKeyOnly ? { apiKeyProvider: provider } : {}),
  ...
});
```

Inside `removeCredentialConfigReferences` (same file as the CLI command), the branch that strips a **literal** (non-profile-id-shaped) `apiKey` out of `models.providers.<provider>.apiKey` is gated entirely on `apiKeyProvider` being passed:

```ts
if (params.apiKeyProvider === undefined || !next.models?.providers) {
  return next;
}
```

Net effect: logging out through the Gateway/Control UI correctly clears a literal plaintext key from config; logging out the exact same profile through the terminal CLI does not — the key stays in `models.providers.<provider>.apiKey` in plaintext indefinitely. Existing test coverage doesn't catch this: `auth-logout.test.ts`'s "clears a provider binding before removing its key" test uses `apiKey: "openai:manual"` — shaped like a profile id, so it's actually cleared by the unrelated, unconditional `removeAuthProfileConfig(next, id)` call earlier in the function, not by the `apiKeyProvider`-gated branch. A genuine literal secret unrelated to any profile id isn't covered and isn't cleared by the CLI path. Verified locally with a fixture using a real literal key + an `api_key`-typed stored credential (fails today, passes once `apiKeyProvider: credential.provider` is passed when `credential.type === "api_key"`, mirroring the Gateway handler).

**Confirmed workaround for the "no prune surface" gap:** `replacePersistedPluginModelCatalogs()` (`src/agents/plugin-model-catalog.ts`) already supports pruning by omission — its default `deleteMissing: true` behavior deletes any existing `cache_entries` row under `scope='plugin-model-catalog-v1'` whose plugin id isn't in the `pluginCatalogWrites` map passed in. Calling it directly with the current catalog set minus the stale plugin ids removes exactly those rows and leaves the rest untouched. So a supported `openclaw models cache prune [--agent] [--provider]` command wouldn't need new deletion logic, just a CLI surface onto this existing capability (plus whatever confirmation/dry-run UX fits, given it's destructive on data that's currently plaintext-secret-bearing).

This probably changes the right fix shape from "invalidate cache on logout" to something like: (a) add a prune surface, (b) fix `secrets audit`/backup redaction to actually cover this table, and (c) close the SecretRef-blind-spot in doctor's detection — with the CLI-vs-Gateway logout asymmetry above as a separate, smaller fix alongside it.

jason-allen-oneal:
Fixes #142421

## What Problem This Solves

Generated model-catalog credentials can survive local logout, be recovered by Doctor, and appear in Git backups created with `--exclude-secrets`. The earlier provider-wide suppression draft also disabled valid same-provider accounts and could let an older sign-in completion undo logout.

## Why This Change Was Made

This revision removes that unshipped provider-wide fence. It uses the existing credential owners and catalog serialization to retire exact generated credential copies while preserving other accounts and model inventory. It retains the contained backup-cache exclusion and the reviewed fingerprint/write-queue repair.

- Logout pins the selected effective credential and its physical owner separately, revalidates after confirmation and before config mutation, and removes only matching authentication from generated root/plugin catalogs, SQLite catalog rows, migration recovery rows, and interrupted legacy claims.
- Other profiles for the same provider and independent credentials in other agents remain usable. Existing OAuth account/generation peer cleanup remains in force; caller-owned targets are removed by their own exact-row checks rather than an earlier peer sweep.
- Existing catalog generations reject work already in flight. Login, setup-token, paste-token, API-key entry, and credential import use a common admission/completion path and revalidate the persisted owner after asynchronous effects. A fresh explicit sign-in remains allowed.
- Doctor does not create missing authentication from generated caches, even if an old generated cache is restored. Explicit configured and user-authored catalog migration retains its repair path and is checked again after confirmation.
- Synchronous catalog readers attempt migration without waiting on an async writer. On contention they read the last committed SQLite inventory and defer all migration/repair writes; normal legacy migration still runs after the lock becomes available.
- Secret-redacted Git backups exclude agent `cache_entries`, covering both normal generated catalogs and migration-recovery payloads. Restore recreates empty credential/cache tables without a separate logout ledger.

No new schema, table, credential-retirement ledger, or competing auth owner is introduced. Canonical runtime catalog authentication is coordinated with https://github.com/openclaw/openclaw/pull/143781.

## User Impact

Logging out profile A no longer suppresses profile B or an independently owned account. Catalog metadata survives cleanup. Auth commands interrupted by a credential change fail with a retry instruction rather than reporting stale success. Missing generated credentials require a fresh sign-in or explicit import; authored configuration remains an explicit recovery source.

This is local credential removal, not provider token revocation. An old full backup containing canonical credentials can restore them. The redacted-backup contract does not depend on a tombstone in the omitted cache. Existing backup history is not rewritten.

## Evidence

Corrective head: `25c18c955cf9bda49768dcc3a2e296a586040871`; incorporates current upstream `1237e647809c37f7f9aa61ca502f8df317420ad1`.

Validation of the latest integration and the preceding reader correction:

- **148 integration tests passed at `8d29c13d904955e98eed503743936e0dca771ed0` (runtime and test sources unchanged by the final shard-only correction)** across login/config preservation, auth refresh, logout, actual-store admission/completion lifecycle, provider-resolution, Gateway login, and SDK suites after integrating upstream credential-only login. Upstream's concurrent-config protections, explicit login/logout refresh reasons, returned refresh outcomes, and shared saved-credential config error are retained. The final one-file resolution also preserves upstream credential-only login, selected plugin ownership, caller authorization at credential/config commits, hosted refresh callbacks, and partial-save reporting. Exact-profile checks remain after refresh and notes; imports still promote once inside admission. The 99-test earlier merge result is historical, not counted twice.
- **86 catalog/registry/serialization tests passed** after the reader correction at the preceding head; these files are unchanged by the latest main merge. The new real-lock/SQLite test creates and refreshes disk-backed ModelRegistry objects while a catalog writer is suspended; both retain committed inventory, leave a pending legacy catalog untouched, then observe the published inventory and safely migrate the legacy file after release.
- Production core and agent-session test typechecks passed through the repository wrapper. Targeted core type-aware lint, formatting, and diff checks passed. The earlier complete 44-test physical-owner suite and its 17 command/lifecycle tests passed before this integration; those historical runs are not silently attributed to this new head.

Commands used the production repository runners: `node scripts/run-vitest.mjs` with the named auth and catalog suites, `pnpm tsgo:core`, `node scripts/run-tsgo.mjs -p test/tsconfig/tsconfig.core.test.agents-other.json --incremental`, and `node scripts/run-oxlint.mjs --tsconfig config/tsconfig/oxlint.core.json --type-aware` on the touched reader files.

Removal snapshots compare physical-store rows with the physical-store updater, preserving scoped/env-only admission and migration refusals first. Inherited credentials are removed through their separate owner target. The nonblocking reader fallback catches only acquisition contention, not errors from migration/repair; existing writer and logout retry/stale-lock policy is unchanged.

Fresh independent scoped source reviews found no actionable P0–P2 issue in the supplied merge and reader-boundary code, including the final credential-only integration. Their filesystem relay was unavailable, so this is excerpt review, not an assertion of independent file-hash verification or executed tests. All reported execution results were collected separately. No assertions, timeouts, or security gates were weakened.

### Final CI shard correction

On the preceding integration head, `check-additional-boundaries` correctly rejected 721 roots in the agents-root typecheck shard, exceeding the unchanged 720-root cap. The final commit transfers the three existing exec-approval test files to agents-tools alongside exec-runtime coverage. Only two shard configuration files change; no runtime, tests, assertions, deadlines, dependencies, or guard limits change.

The production boundary command now passes with exactly-once canonical coverage, no bundled-extension leakage, distinct build-info ownership, and compiler-expanded counts of 718 agents-root / 711 agents-other / 209 agents-tools. Both affected graph typechecks returned exit 0. A fresh independent read-only review inspected the actual four-line diff and found no P0–P2 issue. The real-store production proof below was rerun on the final clean committed head and passed. Hosted CI on that head remains separately tracked; the earlier run is not counted as green.

Commands: `node --import ./scripts/tsx.mjs scripts/check-tsgo-core-boundary.mts`; `node scripts/run-tsgo.mjs -p test/tsconfig/tsconfig.core.test.agents-root.json --incremental`; `node scripts/run-tsgo.mjs -p test/tsconfig/tsconfig.core.test.agents-tools.json --incremental`; `git diff --check`. JSON files are excluded by the repository formatter; compiler parsing and existing layout were preserved, not misreported as a formatter pass.

### Real committed-head logout / Doctor / backup proof

Node v24.19.0; disposable SQLite stores and generated files; actual `modelsAuthLogoutCommand`, canonical `AuthStorage`, Doctor repair, `dumpGitBackupDatabase`, and `restoreGitBackupDirectory`. No authentication or storage mocks. The fixture owns a loopback refusal endpoint for the expected Gateway refresh attempt; it does not contact a real Gateway or provider.

Sanitized raw result (exit 0):

```text
HEAD 25c18c955cf9bda49768dcc3a2e296a586040871
TREE clean
SOURCE_SHA256 4e118619ef24ed9f1557969f3059d887528e4624e4d5136401880d34275ba203
PASS actual logout: selected A removed; B and independent other-agent A retained; current/migration/legacy/root copies sanitized; stale-generation write refused
PASS same-process cold SQLite-owner reopen: real AuthStorage resolves B and independent other-agent A
PASS production Doctor does not promote restored generated stale authentication
PASS production redacted dump/restore: no normal/migration/auth markers in exported files; empty cache/auth credential tables; schema_meta and logical session entry preserved; integrity_check=ok
PASS disk-backed ModelRegistry create/refresh keep committed plugin inventory during an awaited catalog writer; legacy migration deferred, then admitted after release
LIMITS: synthetic provider; 1 local refusal connections; no provider calls or full Gateway restart; non-secret preservation covers schema_meta and one synthetic logical session entry
RESULT PASS
```

The standalone harness is retained as `logout-real-proof.mts`. From the isolated checkout it runs with a secretless environment: `env -i PATH=/usr/bin:/bin /usr/bin/node --import ./scripts/tsx.mjs <harness-path> <label>`. It creates its own private fixture root. The committed regression suites provide reproducible command/store coverage in the repository.

Cold reopen here means closing SQLite handles and runtime snapshots and constructing new readers in the same process. It is not a full Gateway-process restart. Backup preservation covers canonical schema metadata and one synthetic logical-session entry, not every optional plugin's state. The loopback endpoint deliberately refuses Gateway refresh; no successful live refresh or hosted-provider request is claimed.

### Previous review findings

1. **Surviving credentials:** exact credential cleanup replaces provider-wide suppression; same-provider and independent-agent command/store/regeneration tests retain valid authentication.
2. **Successful credential imports:** imports use the same generation admission and exact persisted-owner completion checks as login; fresh imports remain usable.
3. **Stale authentication completion:** stale login/setup/paste/import work is rejected, including a credential removed while Gateway refresh is pending. A fresh login after logout remains admitted.
4. **Writer ordering:** the queued fingerprint migration fix remains; its serialization suite passed in the 86-test final reader group.
5. **Concurrent synchronous readers:** the late availability finding is repaired by nonblocking migration admission plus committed-SQLite reads. The new regression and committed-head production trace cover registry creation, refresh, continued inventory, deferred legacy migration, and post-write publication. No filesystem-existence shortcut or unlocked repair was introduced.

## Remaining Maintainer Review

The implemented recovery contract is explicit: generated caches are not credential authorities, and missing credentials require explicit sign-in/import; configured or user-authored sources retain their authorized repair flow. Auth owners should review that upgrade/recovery behavior before merge. This is an acceptance question about the concrete implementation, not an unresolved contributor code defect or an assertion of existing approval.

Repository-issued Platinum/readiness labels are present; a fresh review of the final shard-only head and its hosted CI are tracked separately. The main integration resolved the merge conflicts and started full CI; this final correction addresses that run's concrete test-graph overflow. Earlier update-helper failures and rerun denials belong to previous heads. No all-green claim is made before the final CI result.


jason-allen-oneal:
@clawsweeper re-review

jason-allen-oneal:
@clawsweeper re-review

Updated head: `f388bfb93c38dd3bd060bca11c7600f58be8fda7` (current-main integration plus the late synchronous-reader correction).

The remaining reader finding is fixed: registry create/refresh now uses committed SQLite inventory when an async catalog writer holds the lock; it does not block that event loop or perform unlocked migration. Pending legacy migration runs only after lock admission. The new real-lock/SQLite regression and the clean committed-head production trace in the PR body cover both phases.

The latest integration passes 99 auth/refresh/logout/lifecycle tests, core typechecking and type-aware lint. Prior 86 catalog/registry tests remain valid on unchanged reader files. The actual logout/Doctor/redacted-backup/registry proof was rerun on this exact clean head and passed. Upstream saved-credential/refresh outcomes and the existing stale-completion guards are both preserved. Hosted CI is being checked separately; no advance rating or all-green claim.

jason-allen-oneal:
@clawsweeper re-review

Current head: `25c18c955cf9bda49768dcc3a2e296a586040871`. The PR body contains the updated clean committed-head production trace.

The current-main integration is now mergeable and preserves upstream credential-only login, caller authorization, saved-credential outcomes, and our credential-scoped lifecycle checks. Its 148 integration tests passed; runtime and test sources are unchanged by the final four-line test-shard correction.

The final correction fixes the hosted 721/720 test-root overflow by assigning the three exec-approval tests to the existing tools shard. The production boundary guard and both affected typechecks pass, with canonical roots covered exactly once (718 agents-root, 209 agents-tools). No tests, assertions, timeout limits, or guard caps changed. Independent source review found no P0–P2 issue.

The actual logout/Doctor/redacted-backup/registry proof was rerun on this final clean head and passed. Full CI is running at https://github.com/openclaw/openclaw/actions/runs/34541510955; no all-green result is claimed in advance. The documented recovery-contract acceptance remains a separate maintainer merge decision.

PR #143651, opened 2026-09-10
fix(daemon): persist Bun SQLite library selection into managed services

steipete:
Related: #142186

## What Problem This Solves

Fixes an issue where a managed Gateway or node-host service installed with `--runtime bun` on macOS could select a different SQLite library than the shell that installed it, because the service's generated environment never carried `OPENCLAW_SQLITE_LIBRARY` or `HOMEBREW_PREFIX`. An operator setting `OPENCLAW_SQLITE_LIBRARY` before install saw it work in that shell and then silently vanish once the service ran under launchd or systemd.

## Why This Change Was Made

`src/daemon/service-env.ts` now forwards both inputs into the generated environment for the Gateway and node-host services, gated to macOS and to a Bun runtime (or an `OPENCLAW_WRAPPER`, which hides the runtime the same way the heap `NODE_OPTIONS` logic already treats it) so Node-runtime services never carry them and never log Gateway startup's "override ignored" warning. `OPENCLAW_SQLITE_LIBRARY` is trimmed; `HOMEBREW_PREFIX` must be an absolute POSIX path or it is omitted. Both values are inline and non-secret, and regenerate from the invoking shell on every install or repair exactly like every other `OPENCLAW_*` service key, so `mergeInstallInvocationEnv` needed no change. This closes the loop with #142186: the daemon probe, the installing shell, and the running service now all judge the same two inputs.

## User Impact

`openclaw gateway install --runtime bun` and `openclaw node install --runtime bun` on macOS now install a service that opens the same SQLite library the operator configured, instead of silently falling back to discovery or Apple's system library. Re-running install or `openclaw gateway start` repair from a shell with the desired values changes them.

## Evidence

New test file `src/daemon/service-env.sqlite-library.test.ts` covers: macOS+Bun forwards both keys (trimmed); macOS+Node forwards neither; Linux/Windows+Bun forwards neither; a wrapper on macOS+Node forwards both; blank overrides and a relative `HOMEBREW_PREFIX` are omitted; node-host service environment forwards only when Bun is selected.

135 tests green across `src/daemon/service-env.sqlite-library.test.ts`, `src/daemon/service-env.test.ts`, `src/commands/node-daemon-install-helpers.test.ts`, `src/daemon/launchd-plist.test.ts`, `src/daemon/systemd-unit.test.ts`. `oxfmt`/`oxlint` clean on all touched files. `check:changed` green (typecheck core + core tests, both Knip dead-export scans, import cycles). Codex autoreview at P1 scoped-clean (0.94).

### DIFF
--- a/docs/install/bun-compatibility.md
+++ b/docs/install/bun-compatibility.md
@@ -46,6 +46,8 @@ Set `OPENCLAW_SQLITE_LIBRARY` in the process environment before starting OpenCla
 OPENCLAW_SQLITE_LIBRARY=/path/to/libsqlite3.dylib bun openclaw.mjs gateway
 ```
 
+On macOS, `openclaw gateway install --runtime bun`, `openclaw node install --runtime bun`, and wrapper-based installs persist `OPENCLAW_SQLITE_LIBRARY` and `HOMEBREW_PREFIX` from the installing shell into the managed service definition, so the service selects the same library. To change these values for an already-installed service, reinstall with `openclaw gateway install --runtime bun --force` (or `openclaw node install --runtime bun --force` for a managed node host) from a shell with the desired values; a bare reinstall of an already-loaded service is a no-op. Direct Node-runtime services never persist them.
+
 An invalid override fails with:
 
 ```text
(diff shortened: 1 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9ac2f61c3aee7f61e278133fc3d0eb6e81d0b285](https://github.com/openclaw/openclaw/pull/143651/commits/9ac2f61c3aee7f61e278133fc3d0eb6e81d0b285)
- Landed commit: [4fd011dedaa5d4d137f0c2d8829994a556043b5f](https://github.com/openclaw/openclaw/commit/4fd011dedaa5d4d137f0c2d8829994a556043b5f)

PR #143658, opened 2026-09-10
fix: keep model probe state until cleanup finishes

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where model probes could lose temporary authentication state or release exclusive state ownership while accepted engine cleanup was still running after the reported result.

## Why This Change Was Made

Probe targets now retain their files, database, internal session, and state ownership through actual work and cleanup. Each target registers with its parent before setup, so an early concurrent-batch failure cannot release state underneath a live sibling. Independent cleanup steps continue after a failure, and the signal bridge is released even if lock release fails.

## User Impact

Probe results and admission closure remain prompt. Slow cleanup keeps the resources it still needs, including after interruption. Calls without pending work retain their existing cleanup and error timing. Late cleanup failures are recorded and reported; isolated read-only authentication behavior is unchanged.

## Evidence

The original source fails a native public-runner regression when the staged directory disappears under a blocked engine disposer. The repair passed 78 focused/sibling tests, an 11-case final fixture replay, required changed checks, documentation validation, and independent Codex review. Coverage includes exclusive locking, signal handling, hidden-session lifetime, concurrent sibling failure, discovery-only exclusion, and independent cleanup after database-disposal errors.

The normal build passed in 245.75 seconds on the rebased head; all six files remain identical to the reviewed source. A compiled proof then used a synthetic plugin loaded by the normal full composition-root loader, the public probe and embedded runner, and three real loopback SSE requests. It verified staged auth-store and directory access during late work, exclusive lock lifetime, and actual OS SIGTERM delivery to the isolated child's process bridge. Each engine instance disposed once; the full root retained its existing raw lifetime.

The strict process tree joined in 6.436 seconds, with all 11,402 captured runtime artifacts unchanged and no application-source fallback among 3,933 resolved modules. The fixture sets the existing cleanup-timeout override to 25 ms; the production default is unchanged. This is synthetic programmatic-host proof, not a packaged CLI or Gateway RPC carrier test. Earlier programmatic-registry/emitter fixtures are retained separately and are superseded for the full-loader and OS-signal observations.

### DIFF
--- a/docs/cli/models.md
+++ b/docs/cli/models.md
@@ -113,7 +113,7 @@ Options:
 
 Probe rows can come from auth profiles, env credentials, or `models.json`. Probe status buckets: `ok`, `auth`, `rate_limit`, `billing`, `timeout`, `format`, `unknown`, `no_model`.
 
-Direct `models status --probe` runs create temporary internal sessions in the selected agent's canonical database, so the command requires exclusive ownership of the configured state directory. Stop a running Gateway with `openclaw gateway stop` before probing; the command removes its internal sessions and releases the state lock when it finishes or is interrupted.
+Direct `models status --probe` runs create temporary internal sessions in the selected agent's canonical database, so the command requires exclusive ownership of the configured state directory. Stop a running Gateway with `openclaw gateway stop` before probing. Probe results can be reported before slow cleanup finishes. Temporary auth directories, internal sessions, and the state lock remain held until accepted work and cleanup settle, including after interruption. Cleanup failures are reported; a timeout does not certify that resources have closed.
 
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e1c2c12372c8caa491785e566f66636e6b0ec3b1](https://github.com/openclaw/openclaw/pull/143658/commits/e1c2c12372c8caa491785e566f66636e6b0ec3b1)
- Landed commit: [15548547d0785013c96a87246b06ed89e64d4861](https://github.com/openclaw/openclaw/commit/15548547d0785013c96a87246b06ed89e64d4861)

PR #143660, opened 2026-09-10
test: reuse transforms for Gateway acquisition fixtures

steipete:
Related: #143628

## What Problem This Solves

The Gateway acquisition ownership suite rebuilt the same TypeScript dependency graph in each of three isolated failed-shutdown workers, adding repeated startup work to CI.

## Why This Change Was Made

Reuse the file-owned native Gateway fixture runner introduced in #143628 and remove the duplicate runner. Each failed acquisition still gets a fresh Vitest process, private home/state/temporary directories, and its own resource owner. Only transformed code is reused. The shared runner now accepts the acquisition suite's existing 2 MiB output limit and permits callers whose assertions live entirely inside the native test to omit a journal callback. Worker death, launcher exit, the one-test JSON report and cleanup remain mandatory.

All 31 acquisition cases, socket/lifecycle assertions, dynamic database-cleanup imports after module resets, original deadlines and three rejected-owner forks remain. The runner is created inside the parent describe block, which the native child already suppresses. The child writes a real worker PID receipt before loading its assertions.

## User Impact

CI-only improvement; no production behavior change. Production delta is zero, and test/support code is reduced by 91 lines. Runner counts, sharding and runtime state isolation are unchanged.

## Evidence

The complete 31-case file passed on both versions using the repository wrapper and empty task-owned transform/temp roots: 86.37 seconds before and 77.87 seconds after. The three native cases changed from 7.98/8.03/8.07 seconds to 8.96/2.75/3.28 seconds. The first case pays cache population; later workers reuse it. Whole-command timing includes worker preparation, cache writes, process joins and cleanup. This is one matched local A/B sample, not a CI speed guarantee.

The observer saw three distinct launchers and three workers, all absent after completion. The initial 2,381 cache files (55.1 MB) remained unchanged across the later forks, which added only their distinct fixture entries. Config bytes stayed identical; all private roots were removed. Maximum command RSS was 2.19 GB before and 2.14 GB after.

A repeat run of the unchanged baseline encountered a host-load spike: worker preparation rose from about five seconds to 72 seconds and the first original case timed out at its unchanged 120-second deadline (30 other cases passed). That failed run is retained as a limitation and is excluded from the successful timing comparison. Cold-import sensitivity under contention remains a follow-up; no timeout was raised.

A later combined run passed 42 of 43 cases, with the existing session fixture reaching its unchanged 90-second deadline. The checkout dependency directory was subsequently removed by a background cache cleanup. The removal occurred after the timed-out run and is not evidence of its direct cause. After restoring the frozen install, all 12 startup/session lifecycle cases passed, the full changed-file gate passed, and the source remained identical to the independently reviewed patch (no actionable P2 findings). These later runs are correctness evidence, not matched timing samples. No Windows speedup is claimed; existing platform cache defaults remain.

### DIFF
--- a/src/gateway/server.fixture-lifetime.test-support.ts
+++ b/src/gateway/server.fixture-lifetime.test-support.ts
@@ -11,6 +11,7 @@ import { hasErrnoCode } from "../infra/errno.js";
 
 export function createGatewayFixtureFork(
   registerCleanup: (cleanup: () => Promise<void>) => unknown,
+  maxBytes = 4 * 1024 * 1024,
 ) {
   const lifetime = createFixtureLifetime();
   registerCleanup(() => lifetime.cleanup());
@@ -54,7 +55,7 @@ export default defineConfig({
   return function runGatewayFixtureFork(
     context: Pick<TestContext, "signal" | "onTestFinished">,
     source: (repoRoot: string, root: string) => string,
-    assertJournal: (journal: unknown, text: string) => void,
+    assertJournal?: (journal: unknown, text: string) => void,
   ): Promise<void> {
     const run = lifetime.run(async () => {
       context.signal.throwIfAborted();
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f112d58e5c8ed0ab9d717a1b8e372018d9c0c499](https://github.com/openclaw/openclaw/pull/143660/commits/f112d58e5c8ed0ab9d717a1b8e372018d9c0c499)
- Landed commit: [421245a6068d2c5164c55291918c1d86eedc9115](https://github.com/openclaw/openclaw/commit/421245a6068d2c5164c55291918c1d86eedc9115)

PR #143663, opened 2026-09-10
fix(workers): retain bundles during prepared provisioning

steipete:
Related: #134931.

## What Problem This Solves

A prepared worker can lose the bundle it still needs while provisioning. When a successor producer prepares another bundle, cache cleanup previously retained only bootstrap receipts and active placements, overlooking the admitted archive recorded before readiness.

## Why This Change Was Made

The shared retention selector now keeps the preparation's bundle hash while the environment is provisioning and has no bootstrap receipt. Both Gateway cache pruning and node retention use that selector. Existing receipt priority, terminal-state cleanup and placement retention remain unchanged.

## User Impact

Interrupted prepared-worker provisioning keeps its admitted archive across store reopen and successor bundle creation. Cleanup can reclaim the archive after the environment becomes terminal, provided no other owner still needs it.

## Evidence

- Reproduced on current main with real bundle archives, a persisted SQLite environment, store close/reopen and the actual exclusive cache pruner: the old code deleted bundle A while its request remained provisioning, producing `ENOENT`.
- The regression now verifies byte-for-byte preservation before terminalization and removal after terminal failure, while retaining the current producer's bundle.
- All 37 focused tests across the bundle, retention-selector and node-retention suites passed. The full changed-file gate passed, including types, lint, dead exports and runtime cycles. Final documentation formatting and whitespace checks passed.
- Fresh independent review through P2 completed with no findings.

This is a four-file extraction of the retention repair from #134931. The original combined PR remains draft and unmerged. Exact-head CI is required before landing.


steipete:
Merged via squash.

- Prepared head SHA: [97d461944a5cd1f1e1069f0aa75e204ef50eb0f2](https://github.com/openclaw/openclaw/pull/143663/commits/97d461944a5cd1f1e1069f0aa75e204ef50eb0f2)
- Landed commit: [7c728e0fc41e912e5a94328f78ee4ca856403c13](https://github.com/openclaw/openclaw/commit/7c728e0fc41e912e5a94328f78ee4ca856403c13)

PR #143664, opened 2026-09-10
fix(sdk): preserve authored requests when model catalogs overlap

obviyus:
Related: #136257. AI-assisted.

## What Problem This Solves

Session API callers could select an authored model but receive credentials, headers, or request defaults from a generated catalog. Cached inventory could also revive removed authentication.

## Why This Change Was Made

The registry captures each source's routes and compatibility settings before composition. Authored request settings apply only to their authored endpoints; generated rows supply model metadata without credential authority. Encoded provider/model identity pairs prevent header collisions. Shared configuration callers retain their existing route precedence.

## User Impact

Authored keys, headers, routes, sparse metadata, and custom APIs remain usable. Generated-only inventory requires current credentials or authored configuration. No operator field or persistent schema is added.

Consumers: session model lookup, availability, and compaction preserve each source's routing, authentication, headers, and token settings.

## Evidence

Public API compaction requests reproduced the original crossover. Compiled transport checks verified endpoint isolation, removed-auth rejection, literal-identity header isolation, and source-specific token fields. The final correction passed 66 focused checks and continuous integration. Fresh checks repeated compatibility requests and targeted API controls on the compiled build. Responses and credentials were synthetic; persistent credential writing, live activation, and provider success were not tested.

### DIFF
--- a/docs/concepts/models.md
+++ b/docs/concepts/models.md
@@ -392,6 +392,12 @@ values. A self-hosted mirror can be selected with an HTTPS
 
 Custom providers configured under `models.providers` are written into `models.json` under the agent directory (default `~/.openclaw/agents/<agentId>/agent/models.json`). Provider-plugin catalogs are stored separately as generated plugin-owned catalog shards and load automatically. This file is merged with config by default; set `models.mode: "replace"` to use only your configured providers.
 
+Generated plugin catalogs supply model inventory, not request credentials. Their
+cached API keys, authentication modes, and request headers do not authorize model
+requests. Use a current auth profile or authored request configuration instead.
+The session SDK preserves authored `models.json` keys and headers while merging
+generated model metadata below authored rows.
+
 <AccordionGroup>
   <Accordion title="Merge mode precedence">
     For matching provider IDs:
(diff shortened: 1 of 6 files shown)


obviyus:
Landed as `5815dd0379c5295039b8d9efe02275dcfe922438` through the guarded exact-head merge.

**Proof:** SDK hardening through the compiled public session SDK, conversation compaction and synthetic loopback HTTP. All 66 affected tests and native hooks/lint pass. Fresh independent acceptance covers the targeted compatibility, header-isolation and SDK gaps. Final local review passes with no findings; [exact-head CI passes](https://github.com/openclaw/openclaw/actions/runs/34450565453).

| Before | After |
| --- | --- |
| Generated catalogs could replace authored request credentials or headers. | Authored/current request authority stays separate from generated inventory. |
| Different literal provider/model pairs could share model headers. | Each pair retains its own headers. |
| A generated endpoint could receive the authored provider's token parameter. | Each selected source retains its effective compatibility and correct HTTP parameter. |

Cached-only authentication intentionally stops; current credentials or authored settings provide recovery. Proof remains scoped: no live-provider, native-login or canonical credential-writer success is claimed.

PR #143670, opened 2026-09-10
fix(cron): keep legacy jobs on maintenance timer

qingminglong:
<details>

</details>

## What Problem This Solves

Fixes an issue where legacy persisted cron jobs that omit the optional `enabled` field could stop receiving scheduler maintenance wakes when their next-run timestamp was missing.

## Why This Change Was Made

Cron scheduling already treats a missing `enabled` value as enabled through `isJobEnabled`, and next-wake selection follows that rule. The maintenance-timer summary used a different explicit-true check, so it could report no enabled jobs and leave the scheduler without a recovery timer. The summary now uses the canonical enabled predicate, with a focused regression covering the legacy shape.

## User Impact

Older cron records continue receiving the 60-second maintenance wake that repairs a missing 
extRunAtMs`, instead of remaining silent until another mutation or restart.

## Evidence

- **Persisted production path:** The focused regression first writes a cron job to the real SQLite-backed cron store, removes only `job_json.enabled` from the stored row (leaving the supported current-schema record with the legacy omission), then calls production `ensureLoaded(..., forceReload)` before `armTimer`. This is distinct from the merged schema-v13 migration repair, which reconstructs `job_json.enabled` during an upgrade from an older wide row.
- **Inspectable result on exact head:**
  ```text
  persisted job after production load: enabled=undefined
  timer diagnostic: jobCount=1 enabledCount=1 withNextRun=0 delayMs=60000
  maintenance wake: armed (60000 ms)
  maintenance callback: nextRunAtMs restored and persisted (> now)
  ```
- **Focused proof:** 
ode scripts/run-vitest.mjs src/cron/service.armtimer-tight-loop.test.ts --reporter=verbose` — 1 file, 6 tests passed, including the persisted-row case.
- **Real service timer proof on exact head:** A real `CronService.start()` loaded the persisted SQLite row with `enabled=undefined`; after `resumeScheduling()`, the actual 60-second maintenance callback ran and persisted 
extRunAtMs=1789066128525` (> proof start `1789062593261`).
- **Checks:** Targeted `oxfmt --check` and `git diff --check` passed.
- **Proof boundary:** No external channel, provider, or user account was used; the proof covers the local production SQLite loader and scheduler timer path only.
- **Exact candidate head:** `5f5d662a101c86c2615c6eb677e26c51ccbed040`.

### DIFF
--- a/src/cron/service.armtimer-tight-loop.test.ts
+++ b/src/cron/service.armtimer-tight-loop.test.ts
@@ -6,7 +6,7 @@ import { createCronServiceState } from "./service/state.js";
 import { ensureLoaded } from "./service/store.js";
 import { armTimer } from "./service/timer.js";
 import { onTimer } from "./service/timer.test-support.js";
-import { saveCronStore } from "./store.js";
+import { loadCronStore, saveCronStore } from "./store.js";
 import type { CronJob } from "./types.js";
 
 const noopLogger = createNoopLogger();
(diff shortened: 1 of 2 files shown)


qingminglong:
@clawsweeper re-review

qingminglong:
@clawsweeper re-review

qingminglong:
@clawsweeper re-review

PR #143673, opened 2026-09-10
fix: preserve attachments with prototype-named content types

xialonglee:
## What Problem This Solves
Remote attachments with `Content-Type: __proto__` or `constructor` crash media handling before upload.

## Why This Change Was Made
Both MIME lookups now accept only their own entries. Unknown types remain strings, with no preferred extension. No shared state is written. The MIME module changes by +6/-2 to enforce these contracts.
Growth accepted under the owner's standing acceptance (2026-09-15).
Alternative: #144458 applies the same guards.

## User Impact
Malformed headers no longer prevent attachment delivery. Ordinary types keep their behavior.

## Evidence
`openclaw message send --channel discord --target channel:<qa-channel> --media <fixture-url>`
Before: both headers fail with `mime.split is not a function`, before upload.
After: recipient downloads contain the exact 31 fixture bytes for both headers and a normal unknown type. An ordinary PNG also arrives byte-for-byte.

## Compatibility
No configuration, schema, protocol, dependency or public signature changes.

## Consumers
Shared remote readers and channel attachment sends use the repaired MIME results.

## Invalidation
No cache or persistent state changes.

## Tests
290 tests passed on a fresh merge: `packages/media-core/src/mime.test.ts`, `src/media/fetch.test.ts`, and `src/media/fetch.prototype-headers.test.ts`. The hosted runner selects the new HTTP test once; all three cases pass.

### DIFF
--- a/packages/media-core/src/mime.test.ts
+++ b/packages/media-core/src/mime.test.ts
@@ -570,6 +570,28 @@ describe("normalizeMimeType", () => {
   });
 });
 
+describe("prototype-named mime keys", () => {
+  // Remote senders control Content-Type headers; object-literal lookups must
+  // not resolve inherited Object.prototype members or downstream string ops throw.
+  it.each([
+    { input: "__proto__", expected: "__proto__" },
+    { input: "constructor", expected: "constructor" },
+  ] as const)("normalizeMimeType($input) stays a plain string", ({ input, expected }) => {
+    expect(normalizeMimeType(input)).toBe(expected);
+  });
+
+  it.each(["__proto__", "constructor"])(
+    "kindFromMime(%s) returns undefined, not a throw",
+    (input) => {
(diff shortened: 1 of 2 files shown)


xialonglee:
@clawsweeper re-review

obviyus:
landpr-approve-admin-merge
pr: 143673
head: a6ab075053de373cc4b359cd5cda3542a1975a56
scope: owner-review

obviyus:
Thanks @xialonglee!

**Proof:** Remote attachments with `Content-Type: __proto__` or `constructor` now reach Discord with their exact original bytes. An ordinary unknown type and a PNG also pass.

- **Bug:** These headers made media handling throw before upload.
- **Fix:** Both MIME tables now return only their own entries and preserve unknown types.

| Before | After |
| --- | --- |
| Attachment send failed before upload. | Recipient downloads match the original bytes. |

PR #143676, opened 2026-09-10
fix: distinguish interrupted dev runners from completed shutdowns

shakkernerd:
<details>

</details>

## What Problem This Solves

Fixes an issue where developers running from source could see an ordinary error or an automatic doctor/restart after a native runner was killed, even while detached workers survived. The UI wrapper also reported cooperative shutdown as a signal, making completed and interrupted stops indistinguishable to callers.

## Why This Change Was Made

On Unix, preserve raw signal termination through the runner, watcher, doctor, and startup-error paths. The UI wrapper acknowledges a stop only after its child returns normally and a complete captured process tree is quiescent; child signal loss, forced cleanup, and failed enumeration retain a signal outcome. Ordinary returned errors, acknowledged rebuilds, and Windows restart behavior retain their existing handling.

## User Impact

Callers can distinguish acknowledged shutdown from losing a native runner and avoid treating interrupted cleanup as a successful restart or stop.

## Evidence

- 178 affected runner, watcher, native lifecycle, and UI wrapper tests passed remotely, with changed-file formatting/lint, line and assertion ratchets, and the relevant scripts/root/infra type graphs.
- Actual native entrypoints reproduce the old runner exit 1 and watch exit 0 after SIGKILL, then preserve `{code: null, signal: "SIGKILL"}` with this change.
- Native UI controls distinguish acknowledged TERM/HUP, raw child termination, forced descendant cleanup, and a failed process scan with a surviving worker. Each fixture independently releases and verifies its owned processes.

### DIFF
--- a/docs/help/debugging.md
+++ b/docs/help/debugging.md
@@ -159,7 +159,20 @@ Benchmark mode suppresses sync-I/O trace spam by default. Set `OPENCLAW_TRACE_SY
 
 The tmux wrapper carries common non-secret runtime selectors into the pane, including `OPENCLAW_PROFILE`, `OPENCLAW_CONFIG_PATH`, `OPENCLAW_STATE_DIR`, `OPENCLAW_GATEWAY_PORT`, and `OPENCLAW_SKIP_CHANNELS`. Put provider credentials in your normal profile/config, or use raw foreground mode for one-off ephemeral secrets.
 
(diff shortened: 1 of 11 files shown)


shakkernerd:
Fixed: native source commands now distinguish acknowledged shutdowns from interrupted cleanup.

PR #143677, opened 2026-09-10
fix: render scalar unions with literal branches

### LINKED ISSUE #143646: Control UI: cron.sessionRetention renders 'Unsupported schema node' — normalizeUnion cannot merge a literal branch with a non-boolean scalar branch
### Version

OpenClaw `2026.9.3` (`1391f7c`) gateway + Control UI served by that gateway. Verified the same logic is present on current `main` in `ui/src/components/config-form.analyze.ts`.

### Summary

Under **Settings → Automation → Automations → Advanced Settings**, the **"Automations Session Retention"** row renders the red error `Unsupported schema node. Use Raw mode.` instead of a control. Every neighbouring row (`cron.skipMissedJobs`, `cron.webhookSsrfPolicy`, `cron.failureAlert.*`, `cron.triggers.enabled`) renders normally.

### Root cause

`cron.sessionRetention` is authored as zod `union([string(), literal(false)])` (`dist/zod-schema-*.mjs`), which the gateway serves as:

```json
{
  "anyOf": [
    { "type": "string" },
    { "type": "boolean", "const": false }
  ]
}
```

In `normalizeUnion()` (`ui/src/components/config-form.analyze.ts`), the mixed *literals + remaining branches* path only merges when the single remaining branch is a **plain `{ type: "boolean" }`** (no other keys). For any other scalar remaining branch it `return null`s, the path lands in `unsupportedPaths`, and the renderer bails with `configForm.unsupportedNode`:

```ts
if (literals.length > 0 && remaining.length > 0) {
  const booleanBranch = remaining.length === 1 ? remaining[0] : undefined;
  const plainBooleanBranch =
    booleanBranch?.type === "boolean" && Object.keys(booleanBranch).length === 1;
  if (
    !plainBooleanBranch ||        // ← { type: "string" } branch fails here
    literals.includes("true") ||
    literals.includes("false") ||
    (schema.anyOf === undefined && literals.some((literal) => typeof literal === "boolean"))
  ) {
    return null;                  // → "Unsupported schema node. Use Raw mode."
  }
  ...
}
```

Notably, the **renderer** (`config-form.render.ts` → `renderNode`) already handles mixed primitive unions (`string|number|boolean`) by falling back to a text input — the analyzer is stricter than the renderer, so this node could render without any renderer change.

### Programmatic reproduction

Against the *shipped* bundle of `2026.9.3` (`dist/control-ui/assets/config-form-*.js`), running the analyzer entrypoint on the exact schema served by `openclaw config schema`:

```js
// imports the shipped bundle; runs its lazy init, then Dr(schema)
analyze({
  type: "object",
  properties: {
    cron: { type: "object", properties: {
      sessionRetention: {
        anyOf: [{ type: "string" }, { type: "boolean", const: false }],
        title: "Automations Session Retention",
      },
    } },
  },
})
```

Output:

```
sessionRetention unsupportedPaths: ["cron.sessionRetention"]   // ← the bug
neighbours unsupportedPaths: []                                // skipMissedJobs, webhookSsrfPolicy — render fine
plain string|number union unsupportedPaths: []                 // all-scalar unions were fixed by #128275
string|literal(true) unsupportedPaths: ["test2"]               // same failure class
```

UI steps: open the gateway Control UI → **Settings → Automation → Automations → Advanced Settings** → the Session Retention row shows the red error instead of a text/select control.

### Scope

The failure class is **any `union([<scalar>, literal(...)])`** where the scalar branch is not a plain `{ type: "boolean" }`. Currently affected config keys:

- `cron.sessionRetention` — `union([string(), literal(false)])`
- `gateway.securityHeaders.strictTransportSecurity` — `union([string(), literal(false)])`

(All-literal unions render as enum selects; `union([boolean(), literal("auto")])` merges into an enum via the existing plain-boolean path — both fine.)

### Suggested fix

Extend the mixed literals+remaining path in `normalizeUnion()`: when `remaining.length === 1` and that branch is a plain scalar (`{ type: "string" }` / number / integer, i.e. in `RENDERABLE_UNION_TYPES`) — not just plain boolean — either

1. pass the schema through unchanged (the renderer's mixed-primitive path already renders a text input, and zod still validates on save), or
2. merge it like the boolean case, treating the branch as the dominant input type with the literal as an accepted value.

Option 1 is the minimal change and matches the existing renderer behaviour for `string|number` unions.

### Related

Same recurring family as #128275 (closed 2026-08-23), which fixed the all-scalar union case; this is the residual literal-branch case.

glenn-agent:
## What Problem This Solves

`cron.sessionRetention` and similar settings schemas can be expressed as a scalar branch plus a literal branch, for example `anyOf: [{ type: "string" }, { const: false }]`. The Control UI analyzer treated that shape as unsupported, so otherwise editable Settings rows fell back to Raw mode even though the scalar editor can represent the valid string value and the explicit `false` literal.

Fixes #143646.

## Summary

- allow config-form analysis to keep safe renderable `anyOf` unions that mix scalar branches with literal branches
- narrow that support to non-nullable string/number/integer scalar branches with supported constraint-only keywords
- preserve original literal branch schemas instead of rebuilding them as const-only entries
- keep unsafe guarded, nullable, boolean, overlapping, and `oneOf` cases in Raw mode
- update integrity fixtures so Raw-only coverage still uses genuinely unsupported schemas

## Evidence

- `git diff --check`
- `pnpm exec oxfmt --write ui/src/components/config-form.analyze.ts ui/src/components/config-form-composition-integrity.browser.test.ts ui/src/components/config-form-map-integrity.browser.test.ts`
- `node --import ./scripts/tsx.mjs /tmp/check-config-analyze.mts`
  - confirmed `anyOf: [{ type: "string" }, { const: false }]` is no longer listed in `unsupportedPaths`
  - confirmed guarded boolean, nullable boolean, ambiguous boolean-label, and overlapping `oneOf` fixtures remain unsupported
  - confirmed the map integrity fixture still reports `values.*.retained` as unsupported for the guarded boolean union
- Earlier temporary jsdom/unit copy of the two new regression cases: `2 passed`

## Local Validation Notes

I attempted the repository's targeted browser Vitest command, but local validation is partially blocked by this host environment:

- `pnpm --dir ui vitest run src/components/config-form.browser.test.ts src/components/config-form-composition-integrity.browser.test.ts src/components/config-form-map-integrity.browser.test.ts` fails when Playwright Chromium launches because `libatk-1.0.so.0` is missing on the host.
- `pnpm --dir ui vitest run src/components/config-form.browser.test.ts --project unit` finds no matching files because `.browser.test.ts` files are excluded from the unit project.
- Earlier `pnpm install --frozen-lockfile` linked packages, then failed the preinstall Node gate because this host has Node `24.15.0` while current OpenClaw requires `>=24.16.0 <25 || >=26.1.0`.

I have not yet produced Gateway-served screenshot/recording proof; CI and reviewer feedback may still require that before merge.

### DIFF
--- a/ui/src/components/config-form-composition-integrity.browser.test.ts
+++ b/ui/src/components/config-form-composition-integrity.browser.test.ts
@@ -124,7 +124,6 @@ describe("config form composition integrity", () => {
     });
 
     expect(analysis.unsupportedPaths).toEqual([
-      "retention",
       "guarded",
       "nullableBoolean",
       "ambiguousBooleanLabel",
--- a/ui/src/components/config-form-map-integrity.browser.test.ts
+++ b/ui/src/components/config-form-map-integrity.browser.test.ts
@@ -337,7 +337,7 @@ describe("config form map integrity", () => {
         type: "object",
         properties: {
           name: { type: "string" },
-          retained: { anyOf: [{ type: "string" }, { const: false }] },
+          retained: { anyOf: [{ type: "boolean", not: { const: true } }, { const: "auto" }] },
         },
       } satisfies JsonSchema;
       const collection = (name: string) =>
--- a/ui/src/components/config-form.analyze.ts
+++ b/ui/src/components/config-form.analyze.ts
@@ -77,6 +77,8 @@ const RENDERABLE_UNION_TYPES = new Set([
   "object",
   "array",
 ]);
+const MIXED_LITERAL_RENDERABLE_UNION_TYPES = new Set(["string", "number", "integer"]);
+const MIXED_LITERAL_BRANCH_KEYS = new Set([...SUPPORTED_CONSTRAINT_ONLY_KEYS, "type"]);
 
 function isAnySchema(schema: JsonSchema): boolean {
   const keys = Object.keys(schema ?? {}).filter((key) => !META_KEYS.has(key));
@@ -625,9 +627,14 @@ function normalizeUnion(
       literals.unshift(true, false);
     } else if (
       schema.anyOf === undefined ||
+      nullable ||
       !remaining.every((entry) => {
         const type = schemaType(entry);
-        return Boolean(type) && RENDERABLE_UNION_TYPES.has(String(type));
+        return (
+          Boolean(type) &&
+          MIXED_LITERAL_RENDERABLE_UNION_TYPES.has(String(type)) &&
+          hasOnlySupportedKeywords(entry, MIXED_LITERAL_BRANCH_KEYS)
+        );
       })
     ) {
       return null;
@@ -653,7 +660,10 @@ function normalizeUnion(
     return {
       schema: {
         ...schema,
-        anyOf: [...remaining, ...literals.map((literal) => ({ const: literal }))],
+        anyOf: [
+          ...remaining,
+          ...union.filter((entry) => "const" in entry || Array.isArray(entry.enum)),
+        ],
         oneOf: undefined,
         nullable,
       },
--- a/ui/src/components/config-form.analyze.ts
+++ b/ui/src/components/config-form.analyze.ts
@@ -613,16 +613,25 @@ function normalizeUnion(
     const booleanBranch = remaining.length === 1 ? remaining[0] : undefined;
     const plainBooleanBranch =
       booleanBranch?.type === "boolean" && Object.keys(booleanBranch).length === 1;
-    if (
-      !plainBooleanBranch ||
-      literals.includes("true") ||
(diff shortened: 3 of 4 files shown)


glenn-agent:
@clawsweeper re-review

glenn-agent:
Closing this duplicate PR because #143700 has already merged the canonical fix for #143646. ClawSweeper's review confirmed the reported Settings analyzer failure is resolved on current main, and keeping this broader branch open would add noise rather than a clearer repair. Thanks for the review context.

PR #143679, opened 2026-09-10
fix(release): enforce trailing-completed-month extended-stable rule

steipete:
## Problem

`docs/reference/RELEASING.md` has defined the extended-stable channel as **"the trailing completed month's `.33+` maintenance line"** since #99352 (2026-07-04), and the public launch post promised one line, released monthly, supported until the next one.

The guard did not enforce that. `validateExtendedStableNpmReleaseRequest` only required protected `main` to be in *any* later calendar month:

```js
if (mainCalendarMonth <= releaseCalendarMonth) { throw ... }
```

So N-minus-anything passed. That is how a `2026.6.35` publish reached the publish step with `main` on `2026.9.3` — three months stale, on a line policy says had already been retired — while `2026.7.33` and `2026.8.33` were in flight at the same time. The npm `extended-stable` tag has meanwhile pointed at `2026.6.34` since 2026-08-04.

## Solution

Require the gap to be exactly one calendar month, so an older `.33+` line retires when `main` advances another month. Year rollover keeps working through the existing `year * 12 + month` arithmetic.

No new workflow input, config key, or env var. `bypassExtendedStableGuard` already short-circuits above this check, so the existing `BYPASS_EXTENDED_STABLE_GUARD` dispatch remains the sanctioned exception for an explicitly approved emergency or final retirement release.

The rejection now names the version that would be legal and the exception:

```text
Extended-stable publishes only the trailing completed month: protected main 2026.9.3
allows 2026.8.PATCH, not 2026.6.35. Retire the older line or dispatch with
BYPASS_EXTENDED_STABLE_GUARD for an explicitly approved exception.
```

Docs, the workflow input description, and root `AGENTS.md` now state the rule the code enforces.

## Impact

User-visible release-channel behavior. Publishing a retired extended-stable line now fails at the boundary instead of proceeding, and requires an explicit maintainer decision via the existing bypass. Maintenance patches on the current line (`.34`, `.35`, …) stay eligible while it remains the trailing completed month.

## Evidence

Guard exercised against the real current state (`main` at `2026.9.3`):

```text
REFUSE  2026.6.35   -> allows 2026.8.PATCH, not 2026.6.35
REFUSE  2026.7.33   -> allows 2026.8.PATCH, not 2026.7.33
ALLOW   2026.8.33
ALLOW   2026.8.34   (maintenance patch on the current line)
ALLOW   2026.12.33  vs main 2027.1.1 (year rollover)
ALLOW   2026.6.35   with explicit BYPASS_EXTENDED_STABLE_GUARD
```

- `node scripts/run-vitest.mjs test/scripts/openclaw-npm-extended-stable-release.test.ts test/scripts/openclaw-npm-extended-stable-workflow.test.ts test/scripts/openclaw-npm-publish.test.ts` — 121 passed
- `node scripts/check-changed.mjs` — pass
- autoreview (codex, P1) — `scoped-clean`

Note: the previous test named "accepts .33, later patches, and any later protected-main calendar month" asserted the old N-minus-anything behavior, so it is inverted rather than extended; the new rejections assert the error text so a regression cannot pass by throwing for an unrelated reason.

## Release notes

Extended-stable publication now enforces the documented single trailing-completed-month line. Releasing a retired `.33+` line requires an explicitly approved `BYPASS_EXTENDED_STABLE_GUARD` dispatch.


steipete:
Merged via squash.

- Prepared head SHA: [15ee4b257cdc9bf686f613cd4ce4454a174dd332](https://github.com/openclaw/openclaw/pull/143679/commits/15ee4b257cdc9bf686f613cd4ce4454a174dd332)
- Landed commit: [bbeedd9a0dc28e06f6b43d215a204e2eb53be03e](https://github.com/openclaw/openclaw/commit/bbeedd9a0dc28e06f6b43d215a204e2eb53be03e)

PR #143686, opened 2026-09-10
fix: keep exact model metadata when aliases coexist

steipete:
Related: #130706, #142100.

## What Problem This Solves

Fixes an issue where catalog lookup returns another model row's capabilities when a provider has both a literal model ID and an equivalent ID in its catalog. For Arcee, changing row order could swap the 32K and 64K context metadata.

## Why This Change Was Made

Prefer the exact provider-scoped literal ID before applying provider equivalence. Keep provider-owned aliases and the unique case-insensitive SDK fallback on a literal miss. The production change is four lines.

## User Impact

Model capability lookup returns the requested row consistently regardless of catalog ordering.

## Evidence

Both catalog orders failed on the baseline; all 15 focused tests pass with the repair. Existing alias, case, ambiguity, and thinking-capability cases remain intact. Independent P2 review and the complete changed-file gate passed. One fresh full build passed at `bb00b5c147906e709c3e2802dac651d3caebb3e9`; [exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34437863073) is green.

The compiled public SDK passed both missing-literal alias-fallback controls and all four literal-row selections across both catalog orders. Each result retained exact object identity, context limits, and input capabilities. The supplemental isolated CLI inventory reported the expected 64K logical model; it canonicalizes equivalent rows upstream, so the SDK matrix is the decisive two-row proof. Both processes exited and their isolated state was removed. No provider request was needed for this catalog lookup boundary.

### DIFF
--- a/src/agents/model-catalog-lookup.test.ts
+++ b/src/agents/model-catalog-lookup.test.ts
@@ -23,6 +23,25 @@ describe("catalog model identity", () => {
     expect(findModelCatalogEntry(catalog, { modelId: "READER" })).toBeUndefined();
   });
 
+  it.each([false, true])(
+    "prefers a literal catalog row over provider equivalence (reversed=%s)",
+    (reverse) => {
+      const rows = [
+        {
+          provider: "arcee",
+          id: "arcee-ai/trinity-large-thinking",
+          name: "Wire",
+          contextWindow: 32_000,
+        },
+        { provider: "arcee", id: "trinity-large-thinking", name: "Logical", contextWindow: 64_000 },
+      ];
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [bb00b5c147906e709c3e2802dac651d3caebb3e9](https://github.com/openclaw/openclaw/pull/143686/commits/bb00b5c147906e709c3e2802dac651d3caebb3e9)
- Landed commit: [b6dfee471c0d10e01856ceab743c5f5a5d2e031c](https://github.com/openclaw/openclaw/commit/b6dfee471c0d10e01856ceab743c5f5a5d2e031c)

PR #143687, opened 2026-09-10
fix(plugins): report supported bundle hook layouts

steipete:
## What Problem This Solves

Fixes plugin inspection reporting detected Claude JSON automation as a supported hook. A JSON-only bundle has no OpenClaw hook pack to load, while a mixed bundle should keep its valid hook collection without advertising the JSON file as a hook.

## Why This Change Was Made

Share the existing metadata reader between workspace hook discovery and installed-component inspection. The projector now inspects manifest-owned roots with the runtime's collection semantics. Raw declarations, native plugin projection, workspace selection, reload retention, and handler execution remain unchanged.

The reader accepts the existing runtime warning callback so cold inspection avoids importing logging/config machinery. Real object narrowing replaces an old unchecked manifest cast, and its one obsolete assertion-baseline entry is removed. The change adds 23 production lines; the bulk of the diff moves the existing reader.

## User Impact

Supported hook lists match the layouts OpenClaw can discover. JSON-only automation remains declared but is classified unavailable. Mixed bundles retain their OpenClaw hook packs. Documentation now explains collection roots and the difference between inspection and runtime loading.

## Evidence

- The final 10-case hook file has three intended projection failures and seven passing controls against original production; all 10 pass after the repair. Cases cover JSON-only, mixed, ordinary collections, directly declared leaves, and existing reload behavior.
- The direct-leaf control verifies that the original runtime loads zero entries from that shape. Inspection preserves those runtime semantics instead of enabling another layout.
- 95 additional owner/sibling cases passed on identical production code, covering metadata parsing, loading, component projection, and management inspection. Counts are not summed across repeated runs.
- On the private dependency checkout, all 10 final cases, all 30 canonical changed-file checks, and the full 16-phase build pass. The build verifies all 152 public SDK subpaths.
- Fresh independent Codex review found no actionable findings through P2. The newer base retains all independent baseline edits and removes only this patch's obsolete assertion entry.

Earlier compiler containment failures came from a borrowed dependency installation; the completed proof uses an owned graph. Nonfatal build warnings are retained. No live Gateway or provider was used, and component inspection does not execute hook handlers.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -2920,7 +2920,6 @@ src/hooks/loader.ts	1
 src/hooks/message-hook-mappers.ts	2
 src/hooks/module-loader.ts	3
 src/hooks/policy.ts	1
-src/hooks/workspace.ts	1
 src/infra/advertised-lan-host.ts	1
 src/infra/agent-event-lifecycle.ts	1
 src/infra/agent-events.ts	2
--- a/docs/plugins/bundles.md
+++ b/docs/plugins/bundles.md
@@ -78,7 +78,7 @@ is detected but not yet wired.
 | ------------- | ------------------------------------------------------------------------------------------------- | -------------- |
 | Skill content | Bundle skill roots load as normal OpenClaw skills                                                 | All formats    |
 | Commands      | `commands/` and `.cursor/commands/` treated as skill roots                                        | Claude, Cursor |
-| Hook packs    | OpenClaw-style `HOOK.md` + `handler.ts` layouts                                                   | Codex          |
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b66733e32885426994cd3ddfe27f9f2228f47cba](https://github.com/openclaw/openclaw/pull/143687/commits/b66733e32885426994cd3ddfe27f9f2228f47cba)
- Landed commit: [a883f52dc4de10a6782717b9b70f15d4d45ff11e](https://github.com/openclaw/openclaw/commit/a883f52dc4de10a6782717b9b70f15d4d45ff11e)

PR #143702, opened 2026-09-10
fix(agents): land forced block-chunk breaks on real boundaries

### LINKED ISSUE #143641: EmbeddedBlockChunker: forced breaks cut mid-sentence and leave code fences unbalanced
## Summary

`EmbeddedBlockChunker` (`src/agents/embedded-agent-block-chunker.ts`) has two
independent defects that corrupt streamed block replies:

1. A forced break ignores a line boundary that is only a few characters away,
   cutting mid-sentence — for every `breakPreference` except `"newline"`.
2. A break landing exactly at the end of a **still-streaming** (unterminated)
   fence is misclassified as being outside the fence, so the chunk is emitted
   with an unbalanced ``` ``` `` and renders as broken markdown.

Both reproduce on `main` today.

## Defect 1 — forced break ignores a nearby line boundary

`drain()` emits as soon as the buffer crosses `minChars`. At that instant only a
few characters past `minChars` are buffered, so a paragraph/newline/sentence
boundary at index ≥ `minChars` usually cannot exist yet. `#pickBreakIndex` then
falls through to the whitespace-scan fallback and cuts inside a sentence.

The guard that defers exactly this case is scoped to a single preference:

```ts
if (preference === "newline" && buffer.length < maxChars) {
  return { index: -1 };
}
```

The guard sits *after* all three preference-specific searches, so it can only be
reached once paragraph, newline, and sentence boundaries have all failed. In
that state, deferring is correct for every preference, not just `"newline"` —
more buffered text can only add candidate boundaries, and once
`buffer.length >= maxChars` the whitespace fallback and `maxChars` clamp still
run. The `"newline"`-only scope looks like an oversight.

Because the emit trigger depends on how many characters arrived in the delta
that crossed `minChars`, the resulting break position is **delta-dependent**,
which makes the bug look intermittent.

## Defect 2 — unbalanced fence when the break lands at the buffer end

A still-streaming fence has no closing marker, so `parseFenceSpans` gives it a
span ending at the buffer end. `findFenceSpanAt` requires
`span.start < index < span.end`.

On the first `maxChars`-triggered break, `maxChars === buffer.length`, so:

```ts
isSafeFenceBreak(fenceSpans, offset + maxChars)
```

probes an index equal to `span.end` and reports the cut as *outside* the fence.
`isSafeFenceBreak` returns `true`, the `fenceSplit` close/reopen path is skipped
entirely, and the chunk is emitted raw with an odd number of fence markers.

## Reproduction

Against `main`, with the chunker driven at its documented API and a config of
`{ minChars: 800, maxChars: 1200, breakPreference: "paragraph", flushOnParagraph: false }`:

```
=== Bug 1: forced break ignores a nearby line boundary ===
delta=1  len=800  MID-LINE  ends " be set to a verified value\n- Item 14:"
delta=3  len=800  MID-LINE  ends " be set to a verified value\n- Item 14:"
delta=10 len=804  MID-LINE  ends "set to a verified value\n- Item 14: the"

=== Bug 2: unbalanced fence across a forced break ===
delta=1  chunks=3  unbalanced=2
  first unbalanced chunk ends: " = compute(1);\nconst x = compu"
delta=50 chunks=3  unbalanced=2
  first unbalanced chunk ends: " = compute(1);\nconst x = compu"
```

Repro harness (synthetic input, no external deps):

```ts
import { EmbeddedBlockChunker } from "./src/agents/embedded-agent-block-chunker.js";

const CONFIG = { minChars: 800, maxChars: 1200, breakPreference: "paragraph", flushOnParagraph: false };

function stream(text, delta) {
  const c = new EmbeddedBlockChunker(CONFIG);
  const out = [];
  const emit = (chunk) => out.push(chunk);
  for (let i = 0; i < text.length; i += delta) {
    c.append(text.slice(i, i + delta));
    c.drain({ force: false, emit });
  }
  c.drain({ force: true, emit });
  return out;
}

// Bug 1 — line boundaries every ~55 chars; the break still lands mid-line.
const bullets =
  "Here is a summary of the proposed changes to the schema.\n\n" +
  Array.from({ length: 40 }, (_, i) => `- Item ${i}: the field must be set to a verified value\n`).join("");
for (const delta of [1, 3, 10]) {
  const first = stream(bullets, delta)[0];
  console.log(delta, first.length, bullets[first.length] === "\n" ? "line-boundary" : "MID-LINE");
}

// Bug 2 — a fence still being streamed when the forced break happens.
const code =
  "Intro paragraph before the block.\n\n```ts\n" +
  "const x = compute(1);\n".repeat(120) +
  "```\n\nTrailing text.\n";
for (const delta of [1, 50]) {
  const chunks = stream(code, delta);
  const bad = chunks.filter((c) => (c.match(/```/g) ?? []).length % 2 !== 0);
  console.log(delta, chunks.length, "unbalanced:", bad.length);
}
```

## Suggested fix

Two small changes in `#pickBreakIndex`:

```diff
-    if (preference === "newline" && buffer.length < maxChars) {
+    // Every preference benefits from waiting: this point is only reached once
+    // paragraph, newline, and sentence boundaries have all failed, and more
+    // buffered text can only add candidates. Breaking now would fall through
+    // to the whitespace scan and cut mid-sentence.
+    if (buffer.length < maxChars) {
       return { index: -1 };
     }
```

```diff
     if (buffer.length >= maxChars) {
-      if (isSafeFenceBreak(fenceSpans, offset + maxChars)) {
+      // An unterminated fence's span ends at the buffer end, and findFenceSpanAt
+      // is exclusive of span.end — so when the cut lands exactly at the end of
+      // the buffer we must probe the last *included* character instead, or a
+      // still-streaming fence is misreported as already closed.
+      const fenceProbe = maxChars === buffer.length ? offset + maxChars - 1 : offset + maxChars;
+      if (isSafeFenceBreak(fenceSpans, fenceProbe)) {
         return { index: maxChars };
       }
-      const fence = findFenceSpanAt(fenceSpans, offset + maxChars);
+      const fence = findFenceSpanAt(fenceSpans, fenceProbe);
```

(`maxChars === buffer.length` is equivalent to "the cut lands at the end of the
source", since in the drain loop `buffer` is `source.slice(start)` and
`offset === start`.)

## Validation

With both changes:

- Defect 1 becomes delta-invariant and the break lands on a line boundary at
  every delta tested (1/3/5/10/20/64).
- Defect 2: a matrix of 4k/8k/21k-character code blocks ×
  `flushOnParagraph` false/true × one-shot plus deltas 1/7/50/400 (24
  configurations) yields **zero** unbalanced chunks, with content byte-identical
  after stripping fence markers.
- One-shot (non-streaming) output is **identical before and after**, so the
  already-correct non-streaming path is undisturbed.
- The existing `embedded-agent-block-chunker.test.ts` suite passes unchanged.

Happy to open a PR with these two hunks plus regression tests if that's welcome.

ruel225:
Fixes #143641.

## What Problem This Solves

`EmbeddedBlockChunker` corrupted streamed block replies in two independent ways, both reproduced on `main` (`15a781e3df2`'s parent `92888bf9fdc`):

1. **Mid-sentence forced cuts.** The drain that crosses `minChars` holds only a few characters past it, so paragraph/newline/sentence boundary searches at index ≥ `minChars` cannot succeed yet. The defer guard that waits for real boundaries was scoped to `breakPreference: "newline"` only; every other preference fell through to the whitespace scan and cut at whatever whitespace was already buffered. The cut was delta-dependent, which made the bug look intermittent.
2. **Unbalanced fences at the buffer boundary.** A still-streaming fence has no closing marker, so its span ends at the buffer end. The forced-break path probed the cut directly, and the exclusive fence lookup (`span.start < index < span.end`) reported a cut at `span.end` as outside the fence. `isSafeFenceBreak` returned true, the close/reopen path was skipped, and the chunk was emitted with an odd number of ``` markers.

## Solution

`#pickBreakIndex`, two changes:

1. The defer guard now applies to **every** preference. It is only reachable once paragraph, newline, and sentence searches have all failed; more buffered text can only add candidates, and the whitespace fallback plus `maxChars` clamp still run once the buffer reaches `maxChars`. Force drains are unaffected (they reach this helper only when `remainingLength > maxChars`).
2. When the forced cut lands at the end of the source, the fence lookup probes the **last included character** instead of the exclusive boundary. A span whose final line is a genuine closing marker (checked with the existing `findFenceCloseLineStart`) is still recognized as balanced, so a fence that ends exactly at the cut is emitted as-is without a synthetic close/reopen pair.

## Why not the issue's suggested diff verbatim

The issue proposes probing `maxChars - 1` whenever `maxChars === buffer.length`. That variant also treats a **balanced fence whose closing marker is the buffer's final bytes** as still-streaming and wraps it in a synthetic close/reopen pair, so the balanced fence is emitted twice-closed and the content after it is pulled into the reopened code block. The added test `cuts cleanly after a balanced fence that ends exactly at the source boundary` is red under that variant and green under this patch.

## Evidence

All runs against this head (`15a781e3df2`), pre-fix runs against the same source before the two `#pickBreakIndex` hunks:

- **Pre-fix red (defect 1):** `waits for maxChars before a forced paragraph break so the cut lands on a line boundary (delta 1/3/10)` — failed with the first chunk ending mid-line (`bullets.charAt(firstChunk.length)` was `" "`, not `"\n"`).
- **Pre-fix red (defect 2):** `closes and reopens a still-streaming fence cut exactly at the buffer boundary (delta 1/50)` — failed with chunks carrying an odd ``` count.
- **Boundary guard red under the issue's variant:** `cuts cleanly after a balanced fence that ends exactly at the source boundary` — passes on `main` (43-test baseline), stays green with this patch, fails with the `maxChars - 1` probe alone.
- **Post-fix:** `embedded-agent-block-chunker.test.ts` 49/49 (43 pre-existing unchanged), `packages/markdown-core` 27 files / 382 tests, chunking-adjacent `embedded-agent-subscribe.code-span-awareness` + `block-reply-rejections` suites.
- `tsgo -p tsconfig.core.json --noEmit` exit 0; `oxlint --type-aware` and `oxfmt --check` clean on both changed files.

## Impact

Streamed embedded-agent replies (block reply chunking is user-configurable per channel via `blockReplyChunking`, default `breakPreference: "paragraph"`) no longer split mid-sentence and no longer publish broken markdown when a forced chunk boundary lands on a streaming code fence. Non-streaming and force-flush output is unchanged apart from the two repaired cut positions.

### DIFF
--- a/src/agents/embedded-agent-block-chunker.test.ts
+++ b/src/agents/embedded-agent-block-chunker.test.ts
@@ -447,4 +447,90 @@ describe("EmbeddedBlockChunker", () => {
       expect(chunks.every((chunk) => chunk.trimEnd() !== `${marker}\n${marker}`)).toBe(true);
     },
   );
+
+  it.each([1, 3, 10])(
+    "waits for maxChars before a forced paragraph break so the cut lands on a line boundary (delta %i)",
+    (delta) => {
+      const chunker = new EmbeddedBlockChunker({
+        minChars: 800,
+        maxChars: 1200,
+        breakPreference: "paragraph",
+        flushOnParagraph: false,
+      });
+      const bullets =
+        "Here is a summary of the proposed changes to the schema.\n\n" +
+        Array.from(
+          { length: 40 },
+          (_, i) => `- Item ${i}: the field must be set to a verified value\n`,
+        ).join("");
+
+      const chunks: string[] = [];
+      for (let i = 0; i < bullets.length; i += delta) {
+        chunker.append(bullets.slice(i, i + delta));
+        chunker.drain({ force: false, emit: (chunk) => chunks.push(chunk) });
+      }
+      chunker.drain({ force: true, emit: (chunk) => chunks.push(chunk) });
+
+      expect(chunks.length).toBeGreaterThan(1);
+      // The drain that crosses minChars holds only a few characters past it,
+      // so the cut must wait for real boundaries instead of scanning the
+      // whitespace inside the current sentence.
+      expect(chunks[0].length).toBeGreaterThanOrEqual(800);
+      expect(bullets.charAt(chunks[0].length)).toBe("\n");
+    },
+  );
+
+  it.each([1, 50])(
+    "closes and reopens a still-streaming fence cut exactly at the buffer boundary (delta %i)",
+    (delta) => {
+      const chunker = new EmbeddedBlockChunker({
+        minChars: 800,
+        maxChars: 1200,
+        breakPreference: "paragraph",
+        flushOnParagraph: false,
+      });
+      const code =
+        "Intro paragraph before the block.\n\n```ts\n" +
+        "const x = compute(1);\n".repeat(120) +
+        "```\n\nTrailing text.\n";
+
+      const chunks: string[] = [];
+      for (let i = 0; i < code.length; i += delta) {
+        chunker.append(code.slice(i, i + delta));
+        chunker.drain({ force: false, emit: (chunk) => chunks.push(chunk) });
+      }
+      chunker.drain({ force: true, emit: (chunk) => chunks.push(chunk) });
+
+      expect(chunks.length).toBeGreaterThan(1);
+      // An unterminated fence's span ends at the buffer boundary, so the cut
+      // at that boundary must carry the synthetic close/reopen pair instead of
+      // emitting an odd number of fence markers.
+      for (const chunk of chunks) {
+        expect((chunk.match(/```/g) ?? []).length % 2).toBe(0);
+      }
+    },
+  );
+
+  it("cuts cleanly after a balanced fence that ends exactly at the source boundary", () => {
+    // The closing marker is the buffer's final bytes, so the fence span ends
+    // exactly at the cut; probing inside that span must still recognize the
+    // balanced fence and emit it without a synthetic close/reopen pair.
+    const chunker = new EmbeddedBlockChunker({
+      minChars: 100,
+      maxChars: 200,
+      breakPreference: "paragraph",
+    });
+    const body = "const x = 1;\n".repeat(14);
+    const text = `Intro.\n\n\`\`\`ts\n${body}\n\`\`\``;
+    expect(text.length).toBe(200);
+
+    const chunks: string[] = [];
+    for (let i = 0; i < text.length; i += 1) {
+      chunker.append(text.slice(i, i + 1));
+      chunker.drain({ force: false, emit: (chunk) => chunks.push(chunk) });
+    }
+    chunker.drain({ force: true, emit: (chunk) => chunks.push(chunk) });
+
+    expect(chunks).toEqual([text]);
+  });
 });
--- a/src/agents/embedded-agent-block-chunker.ts
+++ b/src/agents/embedded-agent-block-chunker.ts
@@ -486,7 +486,11 @@ export class EmbeddedBlockChunker {
       }
     }
 
-    if (preference === "newline" && buffer.length < maxChars) {
+    // Only reachable once paragraph, newline, and sentence boundaries have all
+    // failed. More buffered text can only add candidates, so every preference
+    // benefits from waiting until maxChars instead of falling through to the
+    // whitespace scan and cutting mid-sentence.
+    if (buffer.length < maxChars) {
       return { index: -1 };
     }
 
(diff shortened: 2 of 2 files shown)


ruel225:
Superseded by #143722 (merged as 14d272df728), which closed #143641 with superset coverage of this branch's fix — forced breaks land on real block boundaries, and I verified its repair surface covers the cases this PR added. Closing in favor of the landed work.

PR #143704, opened 2026-09-10
test(telegram): align prepared catalog recovery proof

RomneyDa:
## Summary

- explicitly refresh the prepared model catalog before asserting refreshed Telegram provider buttons
- make ordinary startup reads remain passive instead of relying on an implicit refresh
- synchronize catalog replacement on a recorded durable stale-catalog retry
- require both callbacks, offset advancement, and complete ingress-queue drain

## Release-validation evidence

Full Release Validation Gateway shard 4 reproduced two prepared-catalog races: startup provider browsing could run before an explicit refresh, and catalog replacement could race callback recovery. The recovery proof now waits for a pending row with a positive attempt, last-attempt timestamp, and the recorded `Model catalog is not ready` error before replacing the catalog.

## Test plan

- `telegram-model-picker-prepared-gateway.e2e.test.ts` (3 passed; full real-Gateway file)
- focused oxfmt and oxlint checks passed


RomneyDa:
@clawsweeper re-review

PR #143708, opened 2026-09-10
fix(models): prefer exact-case provider inference

steipete:
Related: #130706, #142100.

## What Problem This Solves

Fixes an issue where choosing a bare model ID is treated as ambiguous even when one provider has an exact match. A differently cased model on another provider could force a fallback, and an early ambiguity result could miss an exact row later in the catalog.

## Why This Change Was Made

Share an exact-first matcher between configured and discovered catalog inference. Within each existing precedence scope, use exact matches before case-insensitive matches and require a unique provider. Preserve the agent-specific scope and the existing unique case-insensitive fallback. The refactor removes 22 production lines.

## User Impact

A bare model selection consistently resolves to its exact matching provider regardless of row order. Ambiguous selections remain ambiguous.

## Evidence

Eight regression cases failed on main and pass with the repair. All 159 focused model-selection and session-model cases pass, including agent precedence and unique case-insensitive fallback controls. Independent P2 review and the complete changed-file gate passed. The model-selection documentation describes the precedence.

A fresh full build passed at `ae7a5347c7d1b35b381ee8408b2d428b1ece0cac`; [exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34439134317) is green. Six isolated compiled CLI `models status --json` cases covered defaults, per-agent settings, and configured provider catalogs in both row orders. All resolved `third/Model`, exited successfully, and left no process groups.

The strict import guard observed no source imports or rejections. All 1,034 loaded files projected into the distribution fixture matched the completed build byte-for-byte; remaining imports were built modules or ready dependencies. This is compiled distribution-layout proof, not npm installation proof. Two earlier development-checkout attempts rejected optional source probes and remain recorded; the final fixture retained the guard and assertions.


steipete:
Merged via squash.

- Prepared head SHA: [ae7a5347c7d1b35b381ee8408b2d428b1ece0cac](https://github.com/openclaw/openclaw/pull/143708/commits/ae7a5347c7d1b35b381ee8408b2d428b1ece0cac)
- Landed commit: [2dca96e7c7f5b4e85d8e24c133ab1c8c3cb2efee](https://github.com/openclaw/openclaw/commit/2dca96e7c7f5b4e85d8e24c133ab1c8c3cb2efee)

PR #143717, opened 2026-09-10
fix(media): preserve native image hooks in provider registry

steipete:
Related: #130706, #142100.

## What Problem This Solves

Fixes an issue where sending multiple images to a plugin with a native single-image hook silently bypasses that hook and uses generic model dispatch instead.

## Why This Change Was Made

Keep absent native hooks absent in the provider registry. The existing image dispatch boundary can then call a native batch hook, call the native single-image hook for each image in order, or supply model-backed dispatch when no native hook exists. Preserve capability-only image providers in automatic selection. Audio and video requirements are unchanged. The repair removes 18 production lines across two owners.

## User Impact

Image plugins receive the image requests their native hooks support. Manifest-only and configured image-capable providers remain available through generic dispatch.

## Evidence

The two native two-image regressions fail on the original main code. Removing registry hydration alone also exposed a generic auto-selection regression; the same cases pass on original main and with the complete repair. All 172 focused tests now pass, including native dispatch, manifest/config automatic selection, overrides, and resets. The complete changed-file gate and independent P2 review passed.

A full build and six compiled image-tool cases passed at `5678943431a41b2bb07d1ac3fc59638eebec9878`. Native single-image hooks received both images in order; the batch hook ran once; both avoided generic HTTP dispatch. Generic automatic selection made one loopback POST containing two distinct images. Prepared and unprepared paths, overrides and reset controls passed with no checkout-source imports and complete cleanup.

CI exposed 15 stale registry diagnostic expectations for synthesized hooks. The failure reproduced locally; the test-only follow-up preserves all 27 ownership combinations and passes all 31 diagnostics tests, fresh P2 review and the complete six-file changed gate. Final head `2dd4f2a0366ff29fa297539f4f5739fb4566a162` changes only that test file from the built revision. Source, runtime/dependency inputs and executed-module hashes were reverified unchanged; build metadata retains its original revision. [Updated-head CI](https://github.com/openclaw/openclaw/actions/runs/34444646481) passed.

### DIFF
--- a/packages/media-understanding-common/src/provider-supports.ts
+++ b/packages/media-understanding-common/src/provider-supports.ts
@@ -1,6 +1,7 @@
 import type { MediaUnderstandingCapability } from "./types.js";
 
 type MediaCapabilityProvider = {
+  capabilities?: readonly MediaUnderstandingCapability[];
   transcribeAudio?: unknown;
   transcribeAudioWithContext?: unknown;
   describeImage?: unknown;
@@ -9,7 +10,7 @@ type MediaCapabilityProvider = {
 
 // Capability checks for media-understanding provider objects.
 
-/** Return true when a provider exposes the method for a media capability. */
+/** Image providers can use shared model dispatch; audio/video require registered methods. */
 export function providerSupportsCapability(
   provider: MediaCapabilityProvider | undefined,
   capability: MediaUnderstandingCapability,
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2dd4f2a0366ff29fa297539f4f5739fb4566a162](https://github.com/openclaw/openclaw/pull/143717/commits/2dd4f2a0366ff29fa297539f4f5739fb4566a162)
- Landed commit: [a2137ccd0af58d4af8ff15360b98d3511b1fa8ae](https://github.com/openclaw/openclaw/commit/a2137ccd0af58d4af8ff15360b98d3511b1fa8ae)

PR #143721, opened 2026-09-10
fix(workers): skip unused bootstrap files during startup

steipete:
Related: #134931.

## What Problem This Solves

Embedded worker turns use local `AGENTS.md` plus the prepared context supplied by the Gateway, but startup opened all six workspace bootstrap files before selecting `AGENTS.md`. This caused unnecessary filesystem reads and could report unreadable files that the worker would never use.

## Why This Change Was Made

The bootstrap loader accepts a typed optional filename selection and filters the existing canonical list before any per-file lookup or guarded read. The embedded worker requests only `AGENTS.md`. The default call retains the same six filenames and order, guarded reads, source-identity checks, and cache behavior.

## User Impact

Worker startup reads only the local bootstrap file it uses. Existing default bootstrap loading and prepared prompt content retain their behavior.

## Evidence

Both worker prompt variants reproduced all six guarded opens on unchanged code. The repaired cases observe only `AGENTS.md` while retaining the existing prompt assertions. A real-filesystem selected-then-default read verifies that a narrow call does not poison later default loading.

All 85 unique tests passed: 83 bootstrap/cache cases and both worker prompt variants. The full changed-file gate passed, and fresh independent review through P2 found no actionable issue. The deterministic improvement is six guarded opens to one; single-run timings are not a performance guarantee. [Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34440573991). No native provider timing claim is made for this filesystem change.

### DIFF
--- a/src/agents/workspace.bootstrap-cache.test.ts
+++ b/src/agents/workspace.bootstrap-cache.test.ts
@@ -6,6 +6,7 @@ import fsSync from "node:fs";
 import fs from "node:fs/promises";
 import path from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import * as boundaryFileRead from "../infra/boundary-file-read.js";
 import { makeTempWorkspace, writeWorkspaceFile } from "../test-helpers/workspace.js";
 import { getOrLoadBootstrapFiles } from "./bootstrap-cache.js";
 import * as workspaceBootstrapRead from "./workspace-bootstrap-read.js";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1075927cc5c618040c8ca8c198fb06852411184c](https://github.com/openclaw/openclaw/pull/143721/commits/1075927cc5c618040c8ca8c198fb06852411184c)
- Landed commit: [e17752535e22e62268252661e51de335906cc01a](https://github.com/openclaw/openclaw/commit/e17752535e22e62268252661e51de335906cc01a)

PR #143722, opened 2026-09-10
fix: preserve streamed reply boundaries without replaying code

### LINKED ISSUE #143641: EmbeddedBlockChunker: forced breaks cut mid-sentence and leave code fences unbalanced

obviyus:
Closes #143641. Related: #143702. Thanks @rgregg for the report.

## What Problem This Solves

Small streamed updates could split a bullet line too early or display code outside its fence. The final answer could then repeat code already delivered in continuation blocks.

## Why This Change Was Made

The chunker waits for a natural boundary before splitting at whitespace and preserves open-fence state across forced breaks. Private reply metadata carries each block's original source through coalescing; final accounting credits it after successful delivery. Rejected or rewritten text retains final-answer recovery.

## User Impact

Prose keeps complete lines when they fit. Code continuations stay fenced within message limits, and delivered code is not repeated in the final answer.

Consumers: streamed chunks, coalesced blocks, and final-reply accounting now preserve source coverage across continuation wrappers.

## Evidence

Real guild messages reproduced premature prose splitting and an unclosed fence. Corrected messages reconstructed the full code exactly once for six update sizes. All 263 focused tests and 24 final-tag streaming tests passed. Draft edits, handoff, deletion, cancellation, and replacement checks passed. Existing platform normalization of line endings and trailing spaces remains; general byte preservation is not claimed. A misleading cancellation timeout notice also reproduced before the change.

### DIFF
--- a/src/agents/embedded-agent-block-chunker.test.ts
+++ b/src/agents/embedded-agent-block-chunker.test.ts
@@ -28,6 +28,82 @@ function expectChunksWithinLength(chunks: string[], maxLength: number) {
 
 describe("EmbeddedBlockChunker", () => {
   it.each([
+    { breakPreference: "paragraph", suffix: "ready\n\nTail", expected: "First line is ready" },
+    { breakPreference: "newline", suffix: "ready\nTail", expected: "First line is ready" },
+    { breakPreference: "sentence", suffix: "ready. Tail", expected: "First line is ready." },
+  ] as const)(
+    "waits for a $breakPreference boundary before falling back to whitespace",
+    ({ breakPreference, suffix, expected }) => {
+      const chunker = new EmbeddedBlockChunker({
+        minChars: 8,
+        maxChars: 30,
+        breakPreference,
+        flushOnParagraph: false,
+      });
+
+      for (const character of "First line is ") {
+        chunker.append(character);
+        expect(drainChunks(chunker)).toEqual([]);
+      }
+      chunker.append(suffix);
+
+      expect(drainChunks(chunker)).toEqual([expected]);
+      expect(drainChunks(chunker, true)).toEqual(["Tail"]);
+      expect(chunker.bufferedText).toBe("");
+    },
+  );
+
+  it("balances an unfinished fence at the exact cap and retains its later continuation", () => {
+    const chunker = new EmbeddedBlockChunker({
+      minChars: 8,
+      maxChars: 20,
+      breakPreference: "paragraph",
+    });
+    chunker.append("```ts\nabcdefghijklm");
+    expect(drainChunks(chunker)).toEqual([]);
+
+    chunker.append("n");
+    expect(drainChunks(chunker)).toEqual(["```ts\nabcdefghij\n```"]);
+    expect(chunker.bufferedText).toBe("```ts\nklmn");
+
+    chunker.append("op\n```");
+    expect(drainChunks(chunker)).toEqual([]);
+    expect(drainChunks(chunker, true)).toEqual(["```ts\nklmnop\n```"]);
+    expect(chunker.bufferedText).toBe("");
+  });
+
+  it.each([
+    { text: "```ts\nabcdefg.", expected: [] },
+    { text: "```ts\nabcdefghijklm.", expected: ["```ts\nabcdefghij\n```"] },
+  ])("keeps terminal code punctuation inside its unfinished fence: $text", ({ text, expected }) => {
+    const chunker = new EmbeddedBlockChunker({
+      minChars: 8,
+      maxChars: 20,
+      breakPreference: "paragraph",
+    });
+    chunker.append(text);
+
+    expect(drainChunks(chunker)).toEqual(expected);
+  });
+
+  it("keeps a genuinely closed fence at the exact cap without reopening it", () => {
+    const chunker = new EmbeddedBlockChunker({
+      minChars: 8,
+      maxChars: 20,
+      breakPreference: "paragraph",
+    });
+    chunker.append("```ts\nabcdefghij\n");
+    expect(drainChunks(chunker)).toEqual([]);
+
+    chunker.append("```");
+    expect(drainChunks(chunker)).toEqual(["```ts\nabcdefghij\n```"]);
+    expect(drainChunks(chunker, true)).toEqual([]);
+
+    chunker.append("Tail");
+    expect(drainChunks(chunker, true)).toEqual(["Tail"]);
+  });
+
+  it.each([
     { tail: "Tail", changed: false, expected: ["Tail"] },
     { tail: "", changed: true, expected: [] },
     { tail: "Fixed tail", changed: true, expected: ["Fixed tail"] },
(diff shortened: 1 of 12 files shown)


obviyus:
Thanks @rgregg for the report.

**Proof:** Real Discord guild requests through an isolated Gateway pass prose and fenced-code update sizes 1, 3, 5, 10, 20 and 64. The principal code source reconstructs exactly once after removing only synthetic fence wrappers. Independent acceptance also covers closed-fence/punctuation boundaries, draft edits and finalization, and matched cancellation/replacement behavior. Focused validation passed 263 tests plus 24 final-tag tests.

| Before | After |
| --- | --- |
| The first prose block stopped inside a bullet line. | Nearby natural boundaries keep complete lines together. |
| A forced code block reached the buffer end without a closing fence. | The scanner's open-fence state preserves balanced continuations. |
| Wrapped code could be followed by a repeated full final answer. | Only successfully delivered source counts toward final suppression; rejected delivery retains its fallback. |

The final revision also fixes the payload's declaration type and keeps the exact literal final-tag test aligned with the intended boundary behavior. Its runtime matches the live-tested revision.

Existing Discord whitespace losses and the misleading cancellation timeout notice were reproduced on the pinned baseline and remain separate limitations. The late-write cancellation probe used an already-closed connection; it does not establish an open-transport internal race result.

[Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34442034745) · Related contributor work: #143702, left unchanged.

Exact-head CI and the required independent landing review passed.

PR #143724, opened 2026-09-10
fix(gateway): stop pipelined MCP calls after rejected uploads

steipete:
## What Problem This Solves

An oversized request to the loopback MCP server could return HTTP 413 while a valid tool call pipelined behind it still executed. The private MCP body reader did not mark the connection as closing when it rejected the body.

## Why This Change Was Made

Use the existing bounded HTTP body reader and rejection transport. They mark the connection as closing synchronously, prevent later body reads from reaching tool execution, and flush the caller's error response before retiring the socket. MCP capture finalization remains independent of the transport's close grace period. The duplicate reader and its error predicates are removed.

## User Impact

Rejected oversized uploads cannot execute a following pipelined tool call. The 1 MiB limit, default 30-second deadline, HTTP 413/408 JSON responses, and MCP diagnostics remain. An excessive declared Content-Length is now rejected immediately with 413, rather than potentially waiting for 408. The shared timer caps extreme overrides at 2,147,000,000 ms, up to 483,647 ms earlier than the old cap for deadlines of roughly 25 days.

## Evidence

The original implementation passed 162 existing tests. Real loopback TCP probes reproduced the pipelined tool execution and the declared-size error difference. The candidate passed all 167 tests across MCP, shared body ingestion, and shared rejection, including all five new probe cases. Final probe validation also asserts the exact UTF-8 tool argument after splitting its bytes across writes.

The probes cover exact-size input, declared oversize, stalled upload, disconnect, capture completion before peer half-close, and suppression of the pipelined tool call. Capture may still register and finalize the following request; the verified guarantee is zero tool execution. This uses synthetic local HTTP and mocked tool discovery, not an external service or operator Gateway.

The complete change removes 80 production code lines and adds 119 test code lines: net +39 maintained code lines (+33 physical). The obsolete type-assertion baseline row is removed separately.

All required commands selected by `check:changed` completed successfully. The initial run caught two Promise-executor lint issues in the new test; after the mechanical block-body repair, lint and every remaining required command passed. The 167-test suite and final five-case probe preceded that repair; no production code or runtime assertion changed. Independent review found no P0–P2 issues, and the final block-body repair was reviewed separately.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -2717,7 +2717,6 @@ src/gateway/mcp-app-operations.ts	3
 src/gateway/mcp-app-reconstruction.ts	3
 src/gateway/mcp-grant-store.ts	2
 src/gateway/mcp-http.handlers.ts	4
-src/gateway/mcp-http.request.ts	2
 src/gateway/mcp-http.schema.ts	3
 src/gateway/method-scopes.ts	10
 src/gateway/node-browser-proxy-policy.ts	1
--- a/src/gateway/mcp-http.request.ts
+++ b/src/gateway/mcp-http.request.ts
@@ -12,7 +12,6 @@ import type { OpenClawConfig } from "../config/types.openclaw.js";
 import { isTruthyEnvValue } from "../infra/env.js";
 import { safeEqualSecret } from "../security/secret-equal.js";
 import { normalizeMessageChannel } from "../utils/message-channel.js";
-import { resolveSafeTimeoutDelayMs } from "../utils/timer-delay.js";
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f897a0c746f429172489722d01f5430b4c4a8562](https://github.com/openclaw/openclaw/pull/143724/commits/f897a0c746f429172489722d01f5430b4c4a8562)
- Landed commit: [1a6d46b02abaa373087ba511d2b1f0ce972f7c6f](https://github.com/openclaw/openclaw/commit/1a6d46b02abaa373087ba511d2b1f0ce972f7c6f)

PR #143725, opened 2026-09-10
fix: release model-selected plugin resources after agent runs

steipete:
Related: #140674

## What Problem This Solves

Fixes an issue where repeated agent runs could retain model-selected plugin resources after their last user finished.

## Why This Change Was Made

Fresh RUN generations now use the existing registration-resource owner. Resources stay available through admitted work and cleanup, then dispose after the final claim releases, including existing bounded idle retention. Warm callers continue sharing their prepared generation.

The internal borrowed-model helper now requires its caller's prepared snapshot, removing an unused implicit-acquisition path. Public SDK and injected TTS request contracts remain unchanged. Production code shrinks by 21 lines.

## User Impact

Model-selected resources gain a defined cleanup lifetime across cancellation, replacement, and shutdown. Fresh registrations bypass the global raw registry cache; configured, standalone, root, and raw SDK sources retain their existing host contracts. Borrowing an already managed generation preserves its ownership. Shared SDK state remains process-owned, and its durable rows survive registration disposal.

## Evidence

Required changed checks passed, along with 42 lifecycle/helper tests and 28 Cron cases. Independent Codex reviews found no actionable P0–P2 findings. Initial CI exposed a shared Cron mock missing the acquisition export; this PR fixes that fixture without changing its assertions or production behavior.

The final normal build passed in 208 seconds. Ten compiled runtime groups passed on the final head: public SDK setup, TTS, foreground and delegated compaction, copied engines, Image, PDF, isolated completion, terminal disposal failure/shared KV, and public probe cleanup. These exercise real SQLite reads during late work, exactly-once disposal and reopening, replacement/refusal, cancellation, and process cleanup. The public probe includes actual OS SIGTERM. Setup, isolated completion and compaction use loopback model transports; delegate proof also runs MCP HTTP and LSP processes.

Each group checked the complete source and artifact inventories before and after execution. A final independent audit verified all 40,582 tracked files, index/diffs, and 14,588 built artifacts unchanged; all owned processes joined. Image/PDF/isolated cover 14 scenarios, 30 provider calls, and 24 SQLite registrations with automatic or explicitly caller-owned cleanup.

Proof uses private synthetic state. Copied-engine managed-donor controls use an explicitly constructed host; raw-root proof uses normal full loading. The terminal-failure fixture copies byte-identical production artifacts and adds three sealed synthetic plugin files. Existing timeout settings are shortened only in fixtures. These runs do not claim production-vendor or standalone CLI process-liveness coverage.

Earlier matched five-sample timings measured warm acquisition around 0.18 ms and fresh-after-release acquisition increasing from 2.79 to 5.73 ms as registrations are created anew. That is a measured fresh-owner cost, not a speedup claim or a benchmark of the final base. Final smoke runs recorded 41 ms preparation and 97 ms close/reopen for the terminal case, and 2.91 seconds for all three public probe modes.

Hosted CI is green on the final head: [run 34442074608](https://github.com/openclaw/openclaw/actions/runs/34442074608). The prompt-snapshot job initially failed during a Matrix native dependency download (HTTP 500), before tests. One unchanged retry and the dependent CI gate passed; no snapshots were changed.

### DIFF
--- a/docs/plugins/sdk-overview/host-hooks.md
+++ b/docs/plugins/sdk-overview/host-hooks.md
@@ -45,6 +45,20 @@ still does not invoke a discovery-only engine factory. `dispose()` must not dele
 durable state or disable another registration. Existing raw loader and Gateway
 lifetimes do not gain automatic disposal: keep their `cleanup(ctx)` behavior.
 
+Prepared model runtimes for agent runs also own fresh model-selected registrations.
+They use the same discovery registration mode and plugin selection, but fresh
+registrations bypass the global registry cache. Warm callers share their prepared
+generation. Registration resources remain held through admitted work and cleanup;
+`dispose()` runs after the final claim releases, including any bounded idle
+retention between runs.
+
+Creating a configured or standalone publication does not enable this ownership.
+Existing root registries and raw SDK host registrations keep their original owner;
+borrowing an already managed generation preserves that source's ownership.
+The shared database behind `api.runtime.state` keyed and blob stores remains
+process-owned. A registration's disposer must not close that database or delete
+its durable rows.
+
 Image and music generation also own fresh registrations acquired by
 `api.runtime.imageGeneration.generate(...)` and
 `api.runtime.musicGeneration.generate(...)`. They wait for the provider's complete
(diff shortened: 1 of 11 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0d52f9d4354dead08983a54fb9a65084dba5149f](https://github.com/openclaw/openclaw/pull/143725/commits/0d52f9d4354dead08983a54fb9a65084dba5149f)
- Landed commit: [4cac00f89940bac08510360ee92c2194451c00c9](https://github.com/openclaw/openclaw/commit/4cac00f89940bac08510360ee92c2194451c00c9)

PR #143726, opened 2026-09-10
fix(sessions): keep old metadata from breaking reset history

steipete:
## What Problem This Solves

Fixes an issue where users could not load the current conversation history after resetting a session that contained an older plugin message with deeply nested metadata. That older message was outside the current history window, but reads could still fail with `malformed JSON`.

## Why This Change Was Made

Reuse the existing reset boundary to limit non-message markers before their JSON is evaluated. Keep the active-position predicate available to the query planner and guard only excluded JSON arguments, while retaining the existing message-prefix selection, snapshot, and cache ownership.

## User Impact

Current history remains readable when excluded older marker metadata exceeds SQLite's JSON nesting limit. Retained user and assistant messages, the reset marker, later compactions, pagination, and selected historical/current errors keep their existing behavior. No stored data, schema, index, or retention policy changes.

## Evidence

- The new regression fails on original `989abf687037` at the first current-history read with `malformed JSON`, after canonical transcript replacement succeeds.
- All 12 tests in the complete history-events suite pass with the repair, including retained messages separated by an old custom marker, inclusive reset with no later message, later compaction, paging/anchor ordinals, and preserved selected-row errors.
- Full changed-file validation passed with all gates enabled; independent P0–P2 review found no actionable issues. The first check run hit its outer supervision limit during typechecking; the unchanged checks passed with a longer outer budget. Test and assertion timeouts were unchanged.
- [Linux Testbox proof](https://github.com/openclaw/openclaw/actions/runs/34440214812) tested the same tree as `f3b7476c90f1` and exercised the real Gateway `readSessionMessagesWithSourceAsync` entrypoint on isolated synthetic SQLite state. Full and recent reads returned the expected retained messages, markers, and sequence numbers, and a selected deep marker still raised an error. Native artifacts confirm source and child/pipe closure; the lease is stopped.

The Linux workload and artifact transfer succeeded. Its local controller retained a nonzero result because a final local Git verification read timed out; a separate exact source, index, tool, bundle, and program verification passed afterward. This is a correctness proof for the history boundary, not an attribution of earlier production latency observations.

### DIFF
--- a/src/config/sessions/session-accessor.sqlite-history-events.test.ts
+++ b/src/config/sessions/session-accessor.sqlite-history-events.test.ts
@@ -20,6 +20,7 @@ import {
   readSessionTranscriptHistoryAnchorPage,
   readSessionTranscriptHistoryEvents,
   readSessionTranscriptHistoryEventById,
+  readSessionTranscriptHistoryEventCount,
   readSessionTranscriptHistoryEventPage,
 } from "./session-accessor.sqlite-history-events.js";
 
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f3b7476c90f18cbd7fb15c1a1a55d3f702c81e83](https://github.com/openclaw/openclaw/pull/143726/commits/f3b7476c90f18cbd7fb15c1a1a55d3f702c81e83)
- Landed commit: [1c0d32c78f248a5f328eadc751457097e265d147](https://github.com/openclaw/openclaw/commit/1c0d32c78f248a5f328eadc751457097e265d147)

Issue #143180, opened 2026-09-09
[Bug]: Tool Search Catalog silently strips tool-call arguments for local Ollama models (2026.9.3)


nkdmike:
### Bug type

Regression (worked before, now fails)

### Beta release blocker

No

### Summary

Area: Gateway / agent runtime — tool dispatch (tools.toolSearch default-on behavior)
Severity: High — breaks ALL tool calls with required parameters; session becomes inoperable except for argument-less calls.

Environment: OpenClaw 2026.9.3 (commit 1391f7c), Linux x64, Node v24.20.0, Ollama local provider, tools.toolSearch unset (default).

Workaround (verified): "tools": { "toolSearch": false } in openclaw.json, restart gateway — full-argument tool calls work again.

Notes for maintainers: suspect clause drops non-empty dottedInput when nestedInput is an empty record; also consider gating default-on deferral behind provider capability since local models never had the schema visibility two-stage deferral assumes.

### Steps to reproduce

Summary
In 2026.9.3, Tool Search Catalog is enabled by default when tools.toolSearch is unset. For local Ollama models this routes invocations through the two-stage tool_search/tool_call meta-tools. The dispatcher's argument reader (readToolSearchCallArgs, in local-model-lean-BgtBla-n.mjs) treats an empty nested args object as "no arguments at all" and falls through to id-only resolution — discarding any flattened/dotted parameters the model actually supplied. Local models (observed: Qwen3.8-class) frequently emit calls like { "id": "openclaw:core:message", "args": {} } with real params flattened as args.*. Result: inner tool receives {} and fails validation (action: must have required property 'action'), looping indefinitely until mitigated.

### Expected behavior

Expected behavior: either merge dotted + nested inputs regardless of whether the nested object is empty, or default deferral off for local models unless explicitly opted in.

### Actual behavior

Summary
[...]

PR #143729, opened 2026-09-10
fix(agents): preserve flat arguments beside empty Tool Search wrappers

xydigitLybnnnn:
Related: #143180. That issue's reported dotted-argument Ollama failure remains unresolved. This PR addresses the separate loss of ordinary top-level arguments beside an empty wrapper.

## What Problem This Solves

Tool Search already accepts flattened target arguments. However, an empty `args: {}` or `input: {}` wrapper returned early and discarded those ordinary fields. A call containing a valid top-level `path` therefore reached the real read tool with `{}` and failed required-input validation.

Treat an empty record wrapper like a missing wrapper and reuse the existing selector-aware fallback. Non-empty nested arguments keep their precedence. Target validation, tool policy, result handling, retry limits, and explicit `tools.toolSearch: false` remain unchanged.

## Evidence

Public proof uses the built `openclaw agent --local` command, a scripted loopback Ollama-compatible `/api/chat` provider, and a task-owned marker file. The primary turns discover the real core read tool, describe it, and emit the captured `tool_call`. Harmless plugin tools supply the selector-named and argument-less controls. The provider's final response echoes the actual returned tool result. This is transport and agent-path proof; it does not claim natural model inference or reproduction of the linked issue.

| Raw call shape | Baseline `874b2ce292f4` | Candidate `fdfdddba6fe0` |
| --- | --- | --- |
| `{id, args: {}, path}` | Real read target received `{}`; required-path error | Real read target received `path`; exact file marker returned |
| `{id, input: {}, path}` | Real read target received `{}`; required-path error | Real read target received `path`; exact file marker returned |

Twenty-one public candidate cases cover both wrappers, ordinary and dotted arguments, non-empty nested precedence, matching and conflicting selectors, selector-named target arguments, argument-less calls, implicit local defaults, explicit disable, denied tools, invalid target input, unknown targets, and earlier outer-schema rejection of alias-only calls. All providers, agent commands, and listening ports were closed after capture. Raw wire data and public trajectory exports are retained, including existing diagnostic redactions.

Both decoder/dispatcher regressions fail on baseline for the intended reasons. All 272 focused tests, formatting, syntax lint, and native commit-hook checks pass. Full type-aware checks and the final landing gates run through hosted CI and the required review.

Independent source and public-call validation passed within the stated scope.

Preserves @xydigitLybnnnn's contributor commit and repair intent. AI-assisted.

### DIFF
--- a/src/agents/tool-search-runtime.test.ts
+++ b/src/agents/tool-search-runtime.test.ts
@@ -141,6 +141,31 @@ describe("Tool Search flattened call arguments", () => {
       expected: { command: "flattened" },
     },
     {
+      label: "empty args wrapper with flattened top-level params",
+      arguments: {
+        id: "inspect_resource",
+        args: {},
+        command: "list",
+        timeout_ms: 5_000,
+      },
+      expected: { command: "list", timeout_ms: 5_000 },
+    },
+    {
+      label: "empty args wrapper with dotted params",
+      arguments: {
+        id: "inspect_resource",
+        args: {},
+        "args.path": "projects/example.md",
+        "args.limit": 20,
+      },
+      expected: { path: "projects/example.md", limit: 20 },
+    },
+    {
+      label: "empty args wrapper without other params",
+      arguments: { id: "inspect_resource", args: {} },
+      expected: {},
+    },
+    {
       label: "bare selector",
       arguments: { id: "inspect_resource" },
       expected: {},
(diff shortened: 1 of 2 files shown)


obviyus:
Thanks @xydigitLybnnnn! Landed in acf8497f7bca3c76b302628f127fe7438480467a.

**Proof:** The built public `agent --local` command used a scripted loopback Ollama-compatible provider to discover and call the real read tool. Both empty-wrapper payloads failed on baseline and returned the exact file marker with this fix.

| Before | After |
| --- | --- |
| `{id, args: {}, path}` reached read with `{}` and failed required-path validation | The supplied path reaches read and returns the marker |
| `{id, input: {}, path}` lost the same required field | The supplied path reaches read and returns the marker |

- All 272 focused tests and 21 public cases passed, including nested/dotted precedence, reachable selectors, target fields, default/disabled settings, validation, policy, and closure controls.
- [Exact-head hosted CI](https://github.com/openclaw/openclaw/actions/runs/34445147481) and the required final review passed. No CI exception was used.
- Related issue #143180 remains open. Its reported dotted-argument failure and natural model inference are outside this proof's scope.

PR #143734, opened 2026-09-10
refactor: remove redundant terminal media trust evidence

steipete:
Related: #139428

<details>

This is a same-repository branch; repository maintainers can push updates.

</details>

## What Problem This Solves

The embedded-agent terminal path maintains a second media-evidence calculation whose result cannot change the outcome for media produced by the runtime. Four private-helper tests keep that redundant path and a test-only export in place.

## Why This Change Was Made

Remove the subscriber-specific all-trusted-media predicate and its direct tests. The same call already supplies every pending and deferred media URL, and the real producers normalize or derive nonblank URLs: a true result from the removed predicate therefore already implies terminal URL evidence. The general terminal-evidence helper and outbound media trust classification remain unchanged.

## User Impact

No user-visible behavior change. This removes 21 production lines and 52 net test lines while retaining the real lifecycle, subscription, delivery, and media-producer coverage.

## Evidence

- Five complete owner/producer suites: 236 passing cases before, 232 after; exactly the four deleted private-predicate cases disappear, with every retained case and status in the same order.
- Repeated the final five suites under the corrected frozen dependency install: 232 passed, zero pending, identical ordered cases. The initial changed gate exposed an installed proxyline version older than the manifest; normal installation resolved it without source or lockfile changes.
- Eight temporary real-subscription controls pass on both original and candidate. Removing only the retained URL-evidence argument from the candidate makes the two nonempty pending-media controls fail with abandoned instead of working terminal state.
- Deferred MEDIA-only controls preserve queue consumption, withheld callbacks, ordered delivery, and trust. They do not isolate deferred liveness: raw MEDIA directives also count as visible text. The source producer/aliasing proof establishes the deferred implication separately.
- Full changed gate and independent P2 review passed. No runtime-savings claim.

### DIFF
--- a/src/agents/embedded-agent-subscribe.handlers.lifecycle.test.ts
+++ b/src/agents/embedded-agent-subscribe.handlers.lifecycle.test.ts
@@ -4,11 +4,7 @@ import { describe, expect, it, vi } from "vitest";
 import { createInlineCodeState } from "../../packages/markdown-core/src/code-spans.js";
 import { createHookRunner } from "../plugins/hooks.js";
 import { createMockPluginRegistry, TEST_PLUGIN_AGENT_CTX } from "../plugins/hooks.test-fixtures.js";
-import {
-  __testing,
-  handleAgentEnd,
-  handleAgentStart,
-} from "./embedded-agent-subscribe.handlers.lifecycle.js";
+import { handleAgentEnd, handleAgentStart } from "./embedded-agent-subscribe.handlers.lifecycle.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [499096b08807a1401fe95d95e645c85e1c07814d](https://github.com/openclaw/openclaw/pull/143734/commits/499096b08807a1401fe95d95e645c85e1c07814d)
- Landed commit: [aae6c11904ca48f47ab862c110685f8912ad8bc0](https://github.com/openclaw/openclaw/commit/aae6c11904ca48f47ab862c110685f8912ad8bc0)

PR #143746, opened 2026-09-10
fix(workers): stop workspace scans after cancellation

steipete:
Related: #134931.

## What Problem This Solves

Canceling worker preparation or attachment transfer could leave the Gateway scanning and hashing the workspace. The caller's cancellation signal reached Git discovery but was dropped before the filesystem inventory scan.

## Why This Change Was Made

Snapshot and attachment preparation forward the caller's signal to the inventory owner, which combines it with the existing internal failure signal. Cancellation stops new file and metadata work, joins already-admitted operations, and closes their handles before returning. The owner records the first observed cause explicitly, preserving an earlier I/O failure or caller cancellation through later failures. This avoids relying on lazy composite-signal reason evaluation.

## User Impact

Canceled preparation stops reading additional workspace files, and the operation returns only after its admitted file and directory work has settled.

## Evidence

Real-filesystem regressions reproduce continued reads after cancellation through the public snapshot and attachment APIs, metadata and directory traversal, and empty inventories. Additional regressions reproduce both cancellation/error orderings and real directory removal during initial root setup. The final 27-case focused run passes, including exact handle closure and preservation of already-observed failures.

Fresh full-slice independent review through P2 passed. The full changed-file gate passed. [Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34442915964). The existing inventory limits, canonical manifests, and storage formats remain unchanged. This is local filesystem and API-boundary proof, without a cloud-provider timing claim.

### DIFF
--- a/src/gateway/worker-environments/node-workspace-transfer-service.ts
+++ b/src/gateway/worker-environments/node-workspace-transfer-service.ts
@@ -269,7 +269,11 @@ export function createNodeWorkspaceTransferService(options: {
       }
       const root = await fsp.realpath(params.localPath);
       params.signal.throwIfAborted();
-      const actual = await readActualWorkspaceManifest({ root, baseCommit: null });
+      const actual = await readActualWorkspaceManifest({
+        root,
+        baseCommit: null,
+        signal: params.signal,
+      });
       params.signal.throwIfAborted();
       if (!isCurrentContext(context) || !params.isAuthorized()) {
         throw new Error("Worker attachment transfer authority closed");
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [54040b9f934ab73b89d17b3f9835d826d4b43fa2](https://github.com/openclaw/openclaw/pull/143746/commits/54040b9f934ab73b89d17b3f9835d826d4b43fa2)
- Landed commit: [41b336a3031e78216d1b66726509b833b99ab7e3](https://github.com/openclaw/openclaw/commit/41b336a3031e78216d1b66726509b833b99ab7e3)

PR #143747, opened 2026-09-10
fix: show full subagent transcripts in the task sidebar

steipete:
## What Problem This Solves

Fixes an issue where users selecting a native subagent in the Control UI Tasks sidebar would see progress metadata instead of the child's transcript. OpenClaw-managed children already had transcript viewing, while native children exposed a different experience and runtime-specific fallback labels.

## Why This Change Was Made

Both paths now use one paginated `tasks.history` API and the existing read-only transcript renderer. The Gateway authorizes the requester session and routes reads to the OpenClaw child session or the native harness's verified parent/descendant store; native tasks keep their existing lifecycle ownership.

The optional harness history capability is the intended plugin contract for this unification; existing plugins are not required to implement it.

The viewer preserves complete projected message groups during refreshes, supports older history and retries, and keeps the prompt/output inspector for tasks belonging to the current conversation. Native history reads retain connection, ancestry, disposal, and response-size guards.

## User Impact

Subagents show messages, thinking, and expandable tool calls/results in the same view, including nested native subagents. Generated labels and progress text say **Subagent**. Native history remains dependent on an available parent binding; unavailable history fails visibly instead of reading another conversation.

## Evidence

- The original renderer failed the native transcript regression; the repaired renderer passes. Additional regressions cover nested ancestry, requester access, asynchronous ownership changes, pagination gaps, and shrinking/growing sibling display groups.
- Focused Gateway, Codex, Copilot, and UI tests passed; all four background-task browser E2E scenarios passed with synthetic Gateway data.
- `pnpm build`, source/test/UI/plugin typechecks, lint and architecture guards passed. Android ktlint passed with the installed Android Studio JDK; generated `OpenClawProtocol` compiled and Swift lint passed. The full macOS app suite was not run on the operator desktop.
- Independent reviews were completed and accepted findings repaired; the final follow-up found no actionable P0–P2 issue.
- Review follow-up: native projection output now includes supported, thread-scoped `messageId` values. Tests use actual production projection output to verify all row identities, full/paged identity equality, and identity stability when assistant or tool content changes. The missing-identity regression failed before the fix.
- CI follow-up: the registration import boundary is covered by its existing guard, and the dependent Workboard manifest is regenerated and committed. All 77 focused Codex tests and `pnpm plugins:assets:check` pass locally.
- All six file-link browser E2Es pass after migrating the task-result fixture to the shared history endpoint.
- Post-merge live verification found a parent-thread replacement case, repaired in #143869. Final real-provider, pagination, Gateway-restart, and actual sidebar checks passed on main `5ae901326b85b094d58339587157900093e0b48d`. The before/after images below are synthetic; #143869 also includes inspected real-Gateway captures with synthetic test data.

### Before

Before: native subagent opens only the prompt/output inspector

### After

After: shared subagent transcript with earlier messages and an expanded tool result

### DIFF
--- a/apps/android/app/src/main/java/ai/openclaw/app/gateway/GatewayProtocol.kt
+++ b/apps/android/app/src/main/java/ai/openclaw/app/gateway/GatewayProtocol.kt
@@ -931,6 +931,7 @@ enum class GatewayMethod(
   PluginsCatalogBrowse("plugins.catalog.browse"),
   PluginsCatalogCategories("plugins.catalog.categories"),
   PluginsCatalogGet("plugins.catalog.get"),
+  TasksHistory("tasks.history"),
 }
 
 enum class GatewayEvent(
--- a/apps/shared/OpenClawKit/Sources/OpenClawProtocol/GatewayModels.swift
+++ b/apps/shared/OpenClawKit/Sources/OpenClawProtocol/GatewayModels.swift
@@ -13011,6 +13011,7 @@ public struct TaskSummary: Codable, Sendable {
     public let agentid: String?
     public let sessionkey: String?
     public let childsessionkey: String?
+    public let hastranscript: Bool?
     public let ownerkey: String?
     public let runid: String?
     public let taskid: String?
@@ -13042,6 +13043,7 @@ public struct TaskSummary: Codable, Sendable {
         agentid: String? = nil,
         sessionkey: String? = nil,
         childsessionkey: String? = nil,
+        hastranscript: Bool? = nil,
         ownerkey: String? = nil,
         runid: String? = nil,
         taskid: String? = nil,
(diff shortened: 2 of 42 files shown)


steipete:
Addressed the native identity finding in 1bef0f8152264a51b320189854030c9bc0feba3f. Every native display row now exposes a thread-scoped `messageId` before response-size checks. The producer tests use actual projection output to verify nonempty/unique identities, paged/full identity equality, and stable identities after assistant and tool output updates; the missing-identity regression failed before the fix.

The SDK docs and task-history contract now name the supported identity fields. The lightweight registration boundary and generated asset drift were also repaired in abb85c8c2a7113ced25560db0da4278ee06c9d28. Current-head CI is running; the requested real-provider and browser probe will run against merged main afterward.

steipete:
Merged via squash.

- Prepared head SHA: [a6fa40cf3e221a822a488a5d4dc46ef2fd63ba80](https://github.com/openclaw/openclaw/pull/143747/commits/a6fa40cf3e221a822a488a5d4dc46ef2fd63ba80)
- Landed commit: [5da05f85e49740d48dc984a39e8b4b65e8881a28](https://github.com/openclaw/openclaw/commit/5da05f85e49740d48dc984a39e8b4b65e8881a28)

Issue #103198, opened 2026-07-10
[Bug]: WebChat image attachments not mapped to media store path — image tool receives "image_0" instead of real file path


woohahahaaa:
### Beta release blocker: No

---

### Summary

When a user sends an image attachment via WebChat, the assistant's `image` tool receives the reference `image_0` which does not resolve to the actual file path in the media store (`~/.openclaw/media/inbound/`). The message says `[User sent media without caption]` but the image data is not accessible to the `image` tool.

### Steps to reproduce

1. Open a WebChat session
2. Send an image as attachment (no caption)
3. The assistant receives `image_0` as the image reference
4. Calling `image` tool with `image_0` fails with `Local media file not found`

### Expected behavior

The `image` tool should receive the real file path (e.g., `/Users/Agent/.openclaw/media/inbound/<uuid>.jpg`) so it can read and analyze the image.

### Actual behavior

The `image` tool receives `image_0` which resolves to `<workspace>/image_0` — a file that does not exist. The actual image is stored at `~/.openclaw/media/inbound/<uuid>.jpg` but there is no mapping from `image_0` to that path.

### Workaround

Manually discovering the file by timestamp: `ls -lt ~/.openclaw/media/inbound/` and passing the full path to the `image` tool.

### Version

OpenClaw 2026.6.11 (e085fa1)

### Operating system

macOS 24.6.0 (Darwin, x64)

### Install method

npm global

### Model

wooh-anthropic/deepseek-v4-pro

PR #143753, opened 2026-09-10
fix(webchat): expose inline uploads to file tools

harjothkhara:
Refs #103198

## What Problem This Solves

Small images uploaded through WebChat reach vision, but the agent cannot use the same upload with file tools because the active turn omits its persisted media entry.

This change passes that entry to the existing validated staging owner and selects the session's execution workspace. It suppresses image hydration from the managed copy, so the model keeps one original user-image input. Configured workspaces for skills and model preparation remain separate from session and subagent execution workspaces.

## Evidence

The built Gateway and bundled Control UI reproduced the missing-path failure on pinned main `94a2d79bf934`. Real browser validation also found that the initial fix staged a selected-workspace upload in the default workspace, where the file guard correctly rejected it. The completed repair stages the upload in its selected workspace before the actual file read and digest command.

Verified runtime outcomes:

- Ordinary small-image upload, actual file read, and actual SHA-256 command agree on the uploaded bytes.
- Mixed image/document uploads remain readable with one image on the current user message.
- A second workspace receives its own files; its file guard rejects the first workspace.
- Larger offloaded images retain their original bytes and single vision input.

The small-image digest was `860f57b09f6d8d3e4264ec14ea3ca2157878804e0cd7ae0cd375ea3da62c502e`. The distinct mixed/selected-workspace image digest was `86c4823ac6fb1fc20bb1e0cad44a384c1bbbf60e580761caf82baa8c40abd296`.

Validation: 125 distinct focused attachment, Gateway, staging, lifecycle, and workspace cases passed across the retained proof; affected staging and skills/cwd cases were rerun after the workspace correction. Durable image claims remain path-free while active media provides the files for staging. New regressions fail on baseline production for the intended reasons. Runtime builds, changed-file formatting, syntax lint, and line/assertion checks pass. Independent runtime validation uses synthetic uploads and fresh judgments for each local provider request.

Scope: this fixes ordinary inline upload access. The existing vision-capable all-offloaded route still supplies no file-tool path, so the broader issue remains linked. File-policy proof is not an OS-container test. AI-assisted.

### DIFF
--- a/src/auto-reply/reply/current-turn-images.test.ts
+++ b/src/auto-reply/reply/current-turn-images.test.ts
@@ -393,6 +393,34 @@ describe("resolveCurrentTurnImages", () => {
     expect(resolveAgentTurnAttachments).not.toHaveBeenCalled();
   });
 
+  it("does not rehydrate a managed copy of an already-provided inline image", async () => {
+    vi.mocked(resolveAgentTurnAttachments).mockClear();
+    const inlineImage = {
+      type: "image" as const,
+      data: Buffer.from("inline").toString("base64"),
+      mimeType: "image/png",
+    };
+
+    const result = await resolveCurrentTurnImages({
+      ctx: {
+        Body: "inspect",
+        media: [
+          {
+            path: "/state/media/inbound/photo.png",
+            contentType: "image/png",
+            hydrationSuppressed: true,
+          },
+        ],
+      } satisfies MsgContext,
+      cfg: {} as OpenClawConfig,
+      images: [inlineImage],
+      imageOrder: ["inline"],
+    });
+
+    expect(result).toEqual({ images: [inlineImage], imageOrder: ["inline"] });
+    expect(resolveAgentTurnAttachments).not.toHaveBeenCalled();
+  });
+
   it("hydrates only current image facts missing prompt descriptions", async () => {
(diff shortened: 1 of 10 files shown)


obviyus:
Thanks @harjothkhara!

**Proof:** real bundled WebChat uploads reached file tools in the selected workspace. Actual image reads and SHA-256 output matched the uploaded bytes, with one current user-image input and correct outside-workspace rejection.

- **Bug:** ordinary inline uploads had no staged file path; a selected-workspace upload could also be staged into the default workspace.
- **Fix:** pass persisted inline media through the existing staging owner and use the session's execution workspace.
- **Validation:** 125 distinct focused cases across the retained proof, affected workspace/skills checks, runtime builds, and independent browser/file-tool acceptance.

| Before | After |
| --- | --- |
| Vision received the image, but file tools had no usable path or rejected the wrong-workspace path. | The selected workspace receives the file before actual read and digest tools run. |

The broader issue remains open for the existing all-offloaded image path; this landing covers ordinary inline uploads.

[Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34500283115/attempts/2) after one diagnostic rerun. The first attempt's deadline assertion failure remains recorded; its exact cause was not captured. No red-check exception was used.

PR #143761, opened 2026-09-10
feat(codex): retry cyber-refused turns on Daybreak and surface cyber notices

steipete:
## What Problem This Solves

OpenAI declines some defensive-cyber work on its general models and directs approved workspaces to a Daybreak model instead. Until now a cyber-policy refusal simply ended the turn: the Control UI had no explanation next to the composer, and the refused work had nowhere to go.

## Why This Change Was Made

This PR does two things.

It projects Codex's structured cyber signals into bounded, session-scoped composer notices, distinguishing an in-progress review, a terminal block, and a provider-performed reroute. Stale events are rejected and transient buffering clears when output begins or the turn ends.

It then retries a cyber-refused turn once on the configured Daybreak model, so refused work reaches the tier allowed to answer it. Daybreak trails the general models in capability, so routing stays scoped to work that was actually refused rather than becoming the session's model.

## User Impact

Codex-backed OpenAI chats show an accessible notice immediately above the input for every cyber outcome, including whether an automatic Daybreak retry happened and whether it succeeded.

Escalation behavior, on by default:

- At most one escalated attempt per turn. A second refusal under Daybreak keeps the block and stops.
- Escalation is turn-local. It moves `runtimeModelId` for one attempt and never changes the session's stored model selection.
- A turn already running on the configured Daybreak model is never escalated.
- After an attempt the session gets a bounded in-memory window. Only a Daybreak reply pre-routes later turns in that window; an unauthorized or still-refusing target only suppresses further attempts. The window expires back to the selected model.
- Bio and misalignment refusals are untouched.

Configured under `plugins.entries.codex.config.appServer.cyberFailover` with `mode` (`auto` default, `off`), `model` (default `gpt-daybreak-blue-latest`), and `cooloffMs` (default `600000`).

## Design Note: Why A Model Switch, Not An Access-Program Flag

Codex also exposes a per-turn `cyberAccessProgram` field. Live probing against a real `codex app-server` showed it is not a usable standalone switch and is more likely to fail than a plain model switch, so it is deliberately not implemented here. See Evidence.

## Evidence

Live probes against a real local `codex app-server`, driving JSON-RPC over stdio with `capabilities.experimentalApi: true`:

1. `turn/start` carrying only `cyberAccessProgram: "daybreakBlue"` is accepted by app-server and reaches the backend, which answers `403 Forbidden: "The requested Cyber access program is not authorized for this model."` The access program is scoped to a model, so it is not a standalone switch.
2. `model/list` returns Daybreak as a first-class catalog model, `gpt-daybreak-blue-latest` / `Daybreak Blue`. Catalog entries carry no cyber-program metadata, so program support is not discoverable from the catalog.
3. `turn/start` with `model: "gpt-daybreak-blue-latest"` answers `401 Unauthorized: "You are not authorized to access this model."` on this account, both with and without the access-program field. Catalog presence therefore does not prove entitlement, which is why the retry itself is treated as the only evidence and an unauthorized target degrades to an explicit notice.
4. Each unauthorized attempt drives Codex's own reconnect ladder (`Reconnecting... 1/5` through `5/5`) before the terminal error. That is why a suppressed window blocks repeat attempts rather than retrying.

Automated checks:

- 14 new focused tests for config resolution, refusal-category matching, unauthorized-target detection, replay safety, answered-vs-suppressed outcome, single-attempt planning, session isolation, window expiry, and window bounding.
- Extended Control UI notice tests, including that an escalation supersedes the block it followed and that the escalation outcome is then terminal for the turn. Writing these caught and fixed a real defect where a later buffering event could overwrite the escalation outcome.
- Existing projector, Control UI, and Chromium end-to-end notice tests pass unchanged.
- `node scripts/check-changed.mjs` passes every lane except `lint extensions`, which fails only with `Declaration input escapes checkout` because this worktree is nested inside another install. The changed extension and UI files were linted directly with `oxlint` and are clean; CI is authoritative for that lane.

The screenshots below contain synthetic data only. This is browser boundary proof, not a live upstream classifier reproduction.

| Before | Safety buffering | Policy block |
| --- | --- | --- |
| Composer before provider notice | Composer showing cyber safety review | Composer showing cyber policy block |

## Review Rounds

Six independent review rounds ran against this branch. They changed the design, not just the details, so the summary is worth reading before the diff.

The largest change was a removal. An earlier revision kept using Daybreak for the rest of the cooloff after a successful retry, pre-routing later turns in that session. That sends work the provider never refused to a weaker tier, which is the opposite of the requirement, and it was the source of most of the defects found: what to do when entitlement lapsed mid-window, when a pre-routed turn was itself refused, and how to classify an attempt that neither answered nor failed cleanly. Removing it deleted 192 lines and left one behavior instead of two interacting ones.

Defects found and fixed along the way:

- The retry ignored the attempt's own `replayMetadata.replaySafe` verdict, so a turn refused after it had already sent a message, added a cron entry, or spawned a session would have repeated that action.
- The retry re-ran the full attempt with the same parameters, mirroring the user's prompt into the transcript twice on a successful failover.
- A later buffering event could overwrite a settled escalation outcome in the composer.
- Every authorized retry emitted the escalated notice, including one Daybreak also refused, replacing the terminal block with a reassuring card.
- An attempt counted as answered whenever it lacked an authorization error, so a transport failure or cancellation looked like success.
- The cooloff damper applied to successful escalations too, so for ten minutes after Daybreak worked, later refusals in that session got a bare refusal.
- Authorization was stored per session, so it could be evicted by session churn and was never shared with sessions that had not yet hit it. It is now an account-level record keyed by workspace and target.
- Sibling sessions in one workspace could each start a Daybreak probe and each pay the 401 reconnect ladder.
- Refusal detection matched the diagnostic category but not its provider, and could inherit an earlier turn's refusal when an attempt produced no row of its own.
- Both module-global maps could grow without bound.

## Known Limitation

The refused attempt completes before the wrapper decides to retry, so its terminal assistant row is discarded with its result rather than staged. The prompt is not duplicated, and the discarded result is not committed, but the refused turn still exists upstream in the native Codex thread, which OpenClaw does not rewrite. Isolating that would mean staging the first attempt's projection, which is a larger change to the attempt runner than this PR should carry.

## Config Surface Budget

`docs/.generated/config-baseline.counts.json` carries this PR's three `cyberFailover` keys, regenerated rather than hand-merged after rebasing onto current main.

No Control UI startup-budget exception is carried. An earlier revision raised that baseline; rebasing onto current main moved the measured startup JS from 353195 B to 353060 B, which fits under the existing baseline plus its allowance, so the raise was reverted and the file is identical to main. Headroom is 47 bytes against the 353107 B enforcement limit, so this sits inside the build-variance allowance rather than comfortably under it.

## Size

Production is +356 lines across 31 files. The escalation owner is 230 lines and the notice surface about 240. Two earlier revisions were larger: sticky pre-routing was removed (-192) and the owner was then collapsed from two state maps and four result predicates to one of each (-145). This still does not meet the repository's LOC-neutral preference, and no offsetting cleanup is included.

## Limitations

Daybreak entitlement is server-owned and was not authorized on the account used for probing, so a successful end-to-end escalation against a live refusal is not demonstrated here. The unavailable path is the one exercised live. Operators whose workspace is enrolled get the answered path; everyone else sees the explicit unavailable notice instead of a silent block.


VACInc:
Related follow-up: #145291 (https://github.com/openclaw/openclaw/pull/145291) merged on September 13, 2026. It adds a one-shot Daybreak retry for replay-safe, structured OpenAI cyber refusals in the embedded OpenClaw harness. This PR remains the separate Codex-plugin implementation, including its notices and native-runtime behavior; it is not superseded by #145291. The embedded addition preserves strict model selections and does not change the selected session model.

VACInc:
Related embedded-runtime follow-up PR #147146 is merged on current main. It adds a successful-retry notice to the ordinary embedded OpenClaw source-reply path. It does not change the separate Codex-plugin retry, composer-notice, or native-runtime behavior implemented here.

PR #143762, opened 2026-09-10
refactor: consolidate config test environment ownership

steipete:
<details>

</details>

## What Problem This Solves

The config test environment helper captured and changed each key in sequence. On the Windows main thread, deleting differently cased aliases could lose the value that later restoration needed. Several Doctor fixtures also inherited host provider state and could activate unrelated official providers.

## Why This Change Was Made

Use the canonical environment helper, which captures every requested value before making changes, at the same 59 call sites. Use the existing Doctor isolation helper for 21 fixture homes. Remove the duplicate helper while preserving callbacks, assertion ordering, refusal cases and real persistence checks across all 247 cases.

## User Impact

No production behavior or public API change. The 15 test and test-support files add 110 lines and remove 123, for a net reduction of 13 lines; shared runtime code and database schemas are unchanged.

## Evidence

The original full run passed 245 cases and failed two fixture-isolation cases. Changing only Doctor fixture isolation then passed all 247 cases with the original environment helper still in place. The final candidate also passed all 247 cases, retaining the same test identities. Independent source review is clean through P2.

All 30 normal local changed checks passed, including core and test typechecks, type-aware lint, dead-export checks, import-cycle checks and the required guards. Source remained unchanged through the tests and checks.

A separate native Windows main-thread probe compared eight helper/scenario combinations. The original helper failed restoration in two case-alias scenarios, while the canonical helper passed all four scenarios. This is a narrow native helper contract result, not a Windows Vitest or full-suite result.

### DIFF
--- a/src/cli/gateway-cli.coverage.test.ts
+++ b/src/cli/gateway-cli.coverage.test.ts
@@ -5,8 +5,8 @@ import path from "node:path";
 import { Command } from "commander";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
-import { withEnvOverride } from "../config/test-helpers.js";
 import type { CostUsageSummary } from "../infra/session-cost-usage.js";
+import { withEnvAsync } from "../test-utils/env.js";
 import { ExpectedCliError } from "./failure-output.js";
 import { registerGatewayCli } from "./gateway-cli.js";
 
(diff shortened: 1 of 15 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [40be3cc2d03c6cb23977a2bbf13b6d402295edb2](https://github.com/openclaw/openclaw/pull/143762/commits/40be3cc2d03c6cb23977a2bbf13b6d402295edb2)
- Landed commit: [87261acd0557b2bc7f94a65e90afe8d5af2f6fad](https://github.com/openclaw/openclaw/commit/87261acd0557b2bc7f94a65e90afe8d5af2f6fad)

PR #143766, opened 2026-09-10
refactor(gateway,tui): one owner for question validation and controller state

steipete:
Related: #143273 #143238

## What Problem This Solves

Cleanup pass over the question subsystem after the TUI question support in #143273. Nothing here changes behavior; it removes duplication that had already started to drift and collapses state that had to be kept consistent by hand.

Three things were wrong structurally. Question shape validation existed twice, once in the Gateway request handler and once reimplemented in the embedded broker, and the two copies had already diverged in their error text. The TUI question controller kept six separate containers keyed by question id, so every mutation had to update several of them in step. And the three pending-question claim guards looked like accidental duplication, which invites someone to merge them.

## Why This Change Was Made

**Shared validation with the policy difference made explicit.** `src/gateway/question-validation.ts` now owns the shape rules both callers share: duplicate ids, option counts, duplicate option labels, secret-store pairing and metadata rules, and URL shape. The one genuine behavior difference is now a named option instead of two divergent copies: the Gateway rejects a masked question that is not bound to the secret store, the local broker accepts one, because plain masked prompts are how local mode works. Each caller keeps its own error type and its own extra rules; the broker still rejects store-bound requests in local mode at its own call site.

**The claim guards are documented, not merged.** Tracing showed three guards with genuinely distinct contracts and other inbound routes that reach the steer path with no TUI pre-claim, so nothing is redundant. Each now carries a one-line comment naming what it owns, so the next reader does not unify them. No logic changed in that commit.

**One map for controller state.** The six per-question containers in `src/tui/tui-questions.ts` collapse into one map of a small record, with a single writer that keeps an entry only while it has a record, is resolving, or has a recovery in flight. The separate `mutations` map stays, because it solves a different problem: reconciling events that arrive during an in-flight list refresh.

## User Impact

No user-visible behavior change. One model-facing string changes: a malformed question rejected by the local broker for duplicate option labels now gets the Gateway's existing wording, which names the offending label. That was the point of sharing the rule, and no test asserted the old text.

## Evidence

- The safety property for the whole pass was that existing tests pass **unmodified**. They do: 1,901 tests across 62 files, plus 102 PTY tests, with no test file touched in the diff.
- `pnpm build` clean; `check:changed` passed (the nested-worktree extension-lint preparation step is a known local guard, and the changed extension files were linted directly).
- The repository's own end-to-end proof, `scripts/dev/tui-questions-proof.ts`, drives a real isolated Gateway plus TUI and a scripted model; it passes.
- Live re-test with a real model on isolated state and ports, both TUI modes: option selection, Esc collapse followed by a plain-text reply answering the pending question, and a masked secret-store request. All flows behave as before the refactor: the prompt renders and resolves in both modes, a collapsed question is still answered by an ordinary composer reply, the masked secret request completes its store write, and the typed value appears in no terminal capture.
- Production line count: +161/−163, net −2. The win is structural rather than in raw lines: six containers become one, and one validator replaces two.
- Codex autoreview at P0–P2 returned a single finding, the broker's duplicate-option-label message text. Rejected deliberately: unifying that message with the Gateway's shipped wording is the intent of the change, it is model-facing error text rather than a contract, and no test or doc asserts the old string.

### DIFF
--- a/src/agents/embedded-agent-runner/run/attempt-queue-message.ts
+++ b/src/agents/embedded-agent-runner/run/attempt-queue-message.ts
@@ -397,6 +397,7 @@ export async function steerActiveSessionWithOptionalDeliveryWait(
   }
 }
 
+// Attempt claims allow legacy steering and preserve supplied run or source-bound authority.
 export async function claimEmbeddedPendingUserInputAnswer(
   text: string,
   options: EmbeddedAgentQueueMessageOptions | undefined,
--- a/src/agents/embedded-agent-runner/runs.ts
+++ b/src/agents/embedded-agent-runner/runs.ts
@@ -620,7 +620,7 @@ export async function queueEmbeddedAgentMessageWithOutcomeAsync(
   return queueEmbeddedAgentMessageAsync(sessionId, text, options);
 }
 
-/** Answers pending input without making ordinary messages bypass local queue policy. */
+/** TUI preflight requires V2 ownership; failure leaves ordinary input to local queue policy. */
 export async function claimPendingEmbeddedAgentQuestionAnswer(
   sessionId: string,
   text: string,
--- a/src/agents/harness/gateway-question.ts
+++ b/src/agents/harness/gateway-question.ts
@@ -288,7 +288,7 @@ export async function claimPendingAgentQuestionAnswerFromCaller(params: {
   });
 }
 
-/** Claims eligible question input; unmatched input remains owned by ordinary admission. */
+/** Owns reservation and persistence; absent questions return before checking caller authority. */
 export async function claimPendingAgentQuestionAnswer(params: {
   sessionKey?: string;
   text: string;
--- a/src/gateway/question-validation.ts
+++ b/src/gateway/question-validation.ts
@@ -1,0 +1,57 @@
+import { hasHttpUrlPrefix } from "@openclaw/net-policy/url-protocol";
+import type { QuestionRequestParams } from "../../packages/gateway-protocol/src/index.js";
+import { ENV_SECRET_REF_ID_RE } from "../config/types.secrets.js";
+import { SECRET_STORE_ALLOWED_HOSTS_MAX } from "../secrets/store/secret-store-validation-error.js";
+
+/** Shape rules after protocol validation; callers retain admission policy and error types. */
+export function questionShapeError(
+  questions: QuestionRequestParams["questions"],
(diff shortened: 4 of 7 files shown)


steipete:
Landing verification for head `81ca1371df0dc154e62ee49b4699d4d68c8451dc`:

- Rebased the three reviewed commits cleanly onto main. All three entries in `git range-diff` are unchanged; the aggregate binary patch is byte-identical before and after the rebase. No code or test edits were made during publication.
- Re-ran the focused suite below: **1,901 tests across 62 files passed**, seven Vitest shards, no failures. `git diff --check` passed.
- Exact-head CI [34444877263](https://github.com/openclaw/openclaw/actions/runs/34444877263) completed successfully on attempt 1; no CI reruns were needed.
- [ClawSweeper's exact-head review](https://github.com/openclaw/openclaw/pull/143766#issuecomment-5614190106) reports no findings, no security findings, and no before-merge requests. It explicitly recognizes the duplicate-label wording change as intentional.
- The owner-completed review, fresh Codex autoreview, build, changed checks, 102 PTY tests, repository proof script, and real-model tests in both TUI modes remain recorded in the PR body. Those earlier runs were not repeated during this publication task. The new run above is the post-rebase focused proof.

```sh
node scripts/run-vitest.mjs src/tui src/infra/embedded-*.test.ts src/agents/harness/gateway-question*.test.ts src/agents/tools/ask-user-tool*.test.ts src/agents/tools/secrets-tool*.test.ts src/agents/embedded-agent-runner/runs.steering.test.ts src/gateway/question-manager.test.ts src/gateway/server-methods/question*.test.ts
git diff --check
```

Production LOC: +161/-163 (net -2). Tests: +0/-0. Shared shape validation retains each caller's admission policy; the controller retains resolving/recovery lifetimes and the separate refresh mutation journal. The three claim guards remain logically unchanged.

steipete:
Merged via squash.

- Prepared head SHA: [81ca1371df0dc154e62ee49b4699d4d68c8451dc](https://github.com/openclaw/openclaw/pull/143766/commits/81ca1371df0dc154e62ee49b4699d4d68c8451dc)
- Landed commit: [61ae3f1b0aad3b0b07b2517819abecd09011ef7d](https://github.com/openclaw/openclaw/commit/61ae3f1b0aad3b0b07b2517819abecd09011ef7d)

PR #143700, opened 2026-09-10
fix(ui): keep string-or-false Settings editable

wangyan2026:
Closes #143646

## What Problem This Solves

Fixes an issue where Settings shows “Unsupported schema node. Use Raw mode.” for Automations Session Retention and Strict Transport Security Header, even though their string-or-false values can use the existing text editor.

## Why This Change Was Made

The existing form analyzer now admits string branches with string or boolean literals and retains the complete original union for validation. Numeric, structured, and nullable sentinels remain unsupported when a text editor cannot preserve their types. The renderer, automatic save, Gateway validation, and configuration schema stay with their existing owners.

This revision preserves @wangyan2026's repair and adds the unsupported-sentinel controls found during source review. The analyzer-to-form test covers both string-to-false and false-to-string edits; redundant rendering-only coverage was removed.

## User Impact

Users can edit duration strings or disable retention with boolean false through the normal Settings form. The same editor supports `gateway.http.securityHeaders.strictTransportSecurity`. Invalid retention durations still fail validation, retain a useful error and draft, and allow recovery with valid input. Unsupported union shapes continue to show Raw-mode guidance.

## Evidence

An independent validator drove real Chromium against a built Gateway and its matching served Control UI, using normal dashboard authorization, public Form autosave, independent `config.get`, and same-state reloads.

- Baseline: `c89f3681428a93b6bc023b4053a5f0aca12ad0bc`. Both affected rows showed the unsupported message with no editor; centered screenshots were inspected.
- Candidate: `071d2d3fb095db918de5837e000bf891d868cc9a`. Public acceptance used [CI integration artifact](https://github.com/openclaw/openclaw/actions/runs/34544385250/artifacts/10178555319) `f3a574c5696b41ccac91f0c7e7ab456b06cc994e`, which contains that candidate. Its unchanged Settings owners and differing build inputs were separately qualified; the integration build identity is retained.

| Real form flow | Observed result |
| --- | --- |
| Retention: `7d` → `false` → `48h` | Every save and reload preserved the exact string or boolean type |
| Invalid retention → valid `72h` | Gateway rejected the invalid duration, kept the stored `48h`, showed Save failed with the retained draft, then saved and reloaded `72h` |
| Header: `max-age=120` → `false` → `max-age=300` | Public save/readback/reload preserved strings and boolean false |
| Existing boolean/auto enum, literal enum, scalar union, object, array, and map | Representative controls matched baseline behavior and restored their original values |

[...]

PR #143668, opened 2026-09-10
fix(ui): Automations Session Retention setting shows unsupported schema error in Control UI

LiuwqGit:
> **AI-assisted: yes**

Closes #143646

## What Problem This Solves

- Fixes an issue where users opening **Settings → Automation → Automations → Advanced Settings** in the Control UI would see the red error `Unsupported schema node. Use Raw mode.` on the **Automations Session Retention** row instead of an editable control, forcing a trip to Raw mode to change a plain string value.
- Same failure class affects `gateway.securityHeaders.strictTransportSecurity` and any zod `union([<string>, literal(false)])` setting whose open branch is a plain `{ type: "string" }`.
- **Related:** #128275 fixed the all-scalar union case; this is the residual literal-branch case. Issue carries clawsweeper `fix-shape-clear` / `source-repro` / `queueable-fix` labels.

## Why This Change Was Made

- **Intended outcome:** `normalizeUnion()` now passes a union through unchanged when its single remaining branch is a plain `{ type: "string" }` alongside literal branches, mirroring the existing all-scalar passthrough. The renderer's mixed-primitive path renders such unions as a text input, and every edited candidate is coerced and validated against the original union schema (typing `false` commits boolean `false`; `7d` commits a string) — the analyzer was strictly stricter than the renderer for no benefit.
- **Numeric sentinels stay in Raw mode (review follow-up):** the passthrough is deliberately restricted to string scalars. The renderer picks `inputType: "number"` for number/integer unions without a string branch, and a numeric input cannot display or commit the boolean `false` sentinel, so `union([number(), literal(false)])` and its integer equivalent keep the existing Raw-mode fallback. This boundary is locked by regression cases in `config-form-composition-integrity.browser.test.ts`.
- **Existing integrity tests reconciled (review follow-up):** the finite-boolean composition case now expects the string-sentinel union to render (text input, `"false"` edit commits boolean `false`), and the map/array Raw-only sibling cases now use a genuinely unsupported fixture (`anyOf: [{type:"string", minLength:1}, {const:false}]` — a constrained branch) so their original integrity coverage remains meaningful.
- **Out of scope:** No renderer, schema (zod), or UI layout changes. Constrained string branches (extra keys like `pattern`/`minLength`) and string literals `"true"`/`"false"` keep the existing Raw-mode behavior.
- **Highest-risk area:** `ui/src/components/config-form.analyze.ts` union normalization.
[...]

PR #143771, opened 2026-09-10
fix(ui): render mixed literal + scalar schema unions instead of Raw mode

### LINKED ISSUE #143646: Control UI: cron.sessionRetention renders 'Unsupported schema node' — normalizeUnion cannot merge a literal branch with a non-boolean scalar branch

pengzh1:
## What Problem This Solves

Under **Settings → Automation → Automations → Advanced Settings**, the **Automations Session Retention** row renders the red error `Unsupported schema node. Use Raw mode.` instead of a control. `cron.sessionRetention` is authored as `union([string(), literal(false)])`, and `normalizeUnion()`'s mixed literals+remaining path only merged when the single remaining branch was a plain `{ type: "boolean" }` — `{ type: "string" }` returned null and the node landed in `unsupportedPaths` (#143646). `gateway.securityHeaders.strictTransportSecurity` (`union([string(), literal(false)])`) hits the same class.

The renderer (`config-form.render.ts` → `renderNode`) already handles mixed primitive unions by falling back to a text input — the analyzer was stricter than the renderer.

## What This Changes

Option 1 from the issue: when the union has literals plus exactly one remaining plain scalar branch (`string`/`number`/`integer`), the analyzer passes the schema through to the existing single-remaining-branch normalization (renderer renders a text input; zod still accepts the literal on save). Literal + non-scalar branches keep routing to Raw mode, and the plain-boolean merge path is unchanged.

## Evidence

New unit tests in `ui/src/components/config-form.analyze.test.ts` (3/3 passing):

- `anyOf: [{ type: "string" }, { type: "boolean", const: false }]` (the exact `cron.sessionRetention` schema) → `unsupportedPaths` no longer contains the path. **Red before the fix** (reproduced the reported `Unsupported schema node` routing).
- `anyOf: [{ type: "string" }, { type: "boolean", const: true }]` — same failure class → renderable.
- `anyOf: [{ const: "preset" }, { type: "array", ... }]` — literal + non-scalar still routes to Raw mode (scope guard).

Fixes #143646

### DIFF
--- a/ui/src/components/config-form-composition-integrity.browser.test.ts
+++ b/ui/src/components/config-form-composition-integrity.browser.test.ts
@@ -123,8 +123,9 @@ describe("config form composition integrity", () => {
       },
     });
 
+    // retention (string | literal(false)) renders as a text input since the
+    // scalar pass-through fix (issue #143646); it is no longer Raw-only.
     expect(analysis.unsupportedPaths).toEqual([
-      "retention",
       "guarded",
       "nullableBoolean",
       "ambiguousBooleanLabel",
--- a/ui/src/components/config-form-map-integrity.browser.test.ts
+++ b/ui/src/components/config-form-map-integrity.browser.test.ts
@@ -372,8 +372,14 @@ describe("config form map integrity", () => {
         }),
         container,
       );
-      expect(container.textContent).toContain("Unsupported schema node. Use Raw mode.");
-      expect(container.querySelector('[aria-label="Retained"]')).toBeNull();
+      // Since the scalar pass-through (issue #143646), retained renders as an
+      // editable text input beside its siblings instead of a Raw-mode error.
+      expect(container.textContent).not.toContain("Unsupported schema node. Use Raw mode.");
+      const retained = expectElement(
+        container.querySelector<HTMLInputElement>('[aria-label="Retained"]'),
+        "supported retained field",
+      );
+      expect(retained.value).toBe("false");
       const name = expectElement(
         container.querySelector<HTMLInputElement>('[aria-label="Name"]'),
         "supported name field",
--- a/ui/src/components/config-form-map-integrity.browser.test.ts
+++ b/ui/src/components/config-form-map-integrity.browser.test.ts
@@ -353,7 +353,9 @@ describe("config form map integrity", () => {
               : { type: "array", items: rowSchema },
         },
       });
-      expect(analysis.unsupportedPaths).toEqual(["values.*.retained"]);
+      // retained: literal(false) renders as a text input beside its siblings
+      // since the scalar pass-through fix (issue #143646).
+      expect(analysis.unsupportedPaths).toEqual([]);
       const state = createInitialConfigState();
       state.configSchema = analysis.schema;
       state.configForm = { values: collection("before") };
--- a/ui/src/components/config-form.analyze.test.ts
+++ b/ui/src/components/config-form.analyze.test.ts
@@ -1,0 +1,52 @@
+// UI tests cover config-form schema analysis for union nodes (issue #143646).
+import { describe, expect, it } from "vitest";
+import { analyzeConfigSchema } from "./config-form.analyze.ts";
+
+describe("analyzeConfigSchema mixed literal unions", () => {
+  it("keeps a string + literal(false) union renderable instead of unsupported", () => {
+    // cron.sessionRetention is authored as union([string(), literal(false)]).
+    const analysis = analyzeConfigSchema({
+      type: "object",
+      properties: {
+        cron: {
+          type: "object",
+          properties: {
+            sessionRetention: {
+              anyOf: [{ type: "string" }, { type: "boolean", const: false }],
+              title: "Automations Session Retention",
+            },
+          },
+        },
+      },
+    });
+
+    expect(analysis.unsupportedPaths).not.toContain("cron.sessionRetention");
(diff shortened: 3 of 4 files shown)


pengzh1:
The P2 is fixed at `e7aecffdc4`: the plain-scalar exception now returns the union schema **untouched** instead of falling through to the single-remaining-branch replacement. The literal branch survives analysis, so `coerceTextInputValue` keeps recognizing boolean sentinels through the original branches — typing `false` stays a boolean and passes the duration refinement. The regression test now also asserts the union's `anyOf` (both branches) survives analysis intact.

pengzh1:
The P2 is fixed at `2baeada6cd`: the scalar pass-through is narrowed to **string branches only** — numeric/integer unions with boolean sentinels keep routing to Raw mode, since a text input cannot distinguish a typed number from a boolean sentinel. Regression test asserts the Raw-mode routing for `anyOf: [{ type: "number" }, { type: "boolean", const: false }]` (verified red when the narrowing is reverted). The string path (the actual `cron.sessionRetention` shape) keeps the unchanged-union pass-through with the literal-branch preservation and `anyOf`-survival assertions from the previous revision.

pengzh1:
Closing as superseded: two earlier PRs (#143668 by @LiuwqGit and #143700 by @wangyan2026) already address the same issue. Our PR added integration test coverage and a typo fix but is redundant given the earlier submissions.

PR #143376, opened 2026-09-09
fix(ui): keep translated config hints out of startup preloads

vincentkoc:
## What Problem This Solves

Fixes an issue where the Control UI startup bundle exceeded the hosted validation budget because every translated config-hint fragment was named in the boot-time module-preload dependency table.

## Why This Change Was Made

Split translated config hints from each lazy locale catalog, then filter only hint-bearing JavaScript preload lists so the hint fragment and locale wrapper remain deferred until that locale is selected. Other preload lists, HTML preloads, runtime dependencies, and locale behavior remain unchanged.

## User Impact

The Control UI stays within its startup JavaScript budget while preserving all translated config hints and lazy locale loading.

## Evidence

- Focused UI tests: 101 passed across Vite config, chunking, locale registry, and translation behavior.
- Production Control UI build: 351,896 B startup JavaScript gzip versus the 352,235 B enforcement limit.
- Built `__vite__mapDeps` table contains only the shared runtime, with zero locale or `locale-config-hints-*` entries.
- `oxfmt`, `oxlint`, `git diff --check`, Control UI i18n verification, and independent P0-P2 autoreview are clean.
- `pnpm check:changed` completed all preceding checks but the local typecheck lane rejected the required Codex-worktree `node_modules` symlink before compilation. Exact-head CI remains the authoritative typecheck proof.

PR #143776, opened 2026-09-10
fix(ui): recover locale imports after source errors

steipete:
Related: #143376

## What Problem This Solves

A failed locale source-catalog import stays cached. If it fails before its source files are watched, correcting the source and requesting the locale module again still returns the old error until an unrelated watched-file change clears the cache.

## Why This Change Was Made

Clear a rejected source promise once at its producer and rethrow the original error. A later request can then load corrected source. Successful imports still share one promise, and an obsolete generation only clears its own cache. Partition caching and watcher behavior stay unchanged.

The repair adds four production lines and extends an existing regression table with one case.

## User Impact

Locale module requests can recover after source corrections without an unrelated file edit. The original failure still reaches the caller; the change adds no automatic retry loop or fallback catalog.

## Evidence

- The new permanent case fails on original production while both stale-generation controls pass. The same test source passes after the repair.
- 108 focused Vite/build/chunk/locale cases across six files and all 20 canonical changed-file checks pass.
- A real Vite 8.2.2/tsx 4.23.13 fixture reproduces old-version recovery versus current-version cached failure after repairing an outside-UI source file. The unchanged driver succeeds on the candidate before any watch event, and its genuine filesystem-watch control also succeeds.
- Fresh independent Codex review found no actionable findings through P2.

The real-plugin fixture uses synthetic catalog input, middleware mode, and no network listener. A paired callback control confirmed that its initial missing source watches came from the fixture’s outer TypeScript loader; removing that loader restored source watches without a product change. Full default ui:dev/HMR behavior remains outside this proof.

### DIFF
--- a/ui/config/control-ui-locales.ts
+++ b/ui/config/control-ui-locales.ts
@@ -127,7 +127,11 @@ export function controlUiLocaleModulesPlugin(): Plugin {
       const memoryPath = path.join(i18nAssetsDir, `${request.locale}.tm.jsonl`);
       while (true) {
         const activeCache = catalogCache;
-        activeCache.sourceCatalogLoad ??= loadCurrentSourceCatalog();
+        activeCache.sourceCatalogLoad ??= loadCurrentSourceCatalog().catch((error: unknown) => {
+          // A later request can retry a corrected source without a watched-file change.
+          activeCache.sourceCatalogLoad = null;
+          throw error;
+        });
         let sourceCatalogResult: Awaited<ReturnType<typeof loadCurrentSourceCatalog>>;
         try {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7afa62f93a5b16a857fd856f22e15ddcfb4b169f](https://github.com/openclaw/openclaw/pull/143776/commits/7afa62f93a5b16a857fd856f22e15ddcfb4b169f)
- Landed commit: [50ff4b7d526ac1addcd243c7cb5ff645876b3a4f](https://github.com/openclaw/openclaw/commit/50ff4b7d526ac1addcd243c7cb5ff645876b3a4f)

PR #143780, opened 2026-09-10
fix: keep model auth facts case-sensitive

steipete:
Related: #130706, #142100.

## What Problem This Solves

Fixes an issue where runtime authentication evidence for a case-sensitive model ID can satisfy a differently cased model. The recorder and availability reader both lowercase IDs, collapsing distinct custom models such as `Model` and `model`.

## Why This Change Was Made

Preserve model-ID case when recording and matching runtime auth facts. Keep trimming and existing provider normalization, profile restrictions, route compatibility, revocation and the 64-fact bound. The two production changes remove two lines; there is no persistent storage or configuration change.

## User Impact

A successful authentication result applies to the exact model that established it. Differently cased custom models remain separate.

## Evidence

Four intended baseline failures were reproduced. All 72 focused producer, reader and publication cases pass, along with the complete changed-file gate and independent P2 review.

A fresh canonical `sourcePerformance` runtime build passed at `2fb44fd507797fdbc19200adbde22c28347992db`. Compiled proof used the real bundled custom-route policy and two authenticated synthetic loopback requests preserving `Model` and `model`. After each request, the production recorder and availability reader accepted only its exact model case, rejected the opposite case, retained trimming and suppressed duplicate facts. Both fact scopes were revoked and verified empty; the server and isolated state were removed. All 2,476 imports were compiled, with no source imports.

This is compiled loopback plus recorder/reader boundary proof; it does not exercise the native harness lifecycle. The runtime build profile covers runtime assets, not UI or declaration generation.

[Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34446322907) passed.

### DIFF
--- a/src/agents/auth-profiles/runtime-materializations.ts
+++ b/src/agents/auth-profiles/runtime-materializations.ts
@@ -60,7 +60,7 @@ export function recordRuntimeAuthMaterialization(params: {
   const provider = normalizeProviderId(params.provider);
   const fact: RuntimeAuthMaterialization = {
     provider,
-    modelId: params.modelId.trim().toLowerCase(),
+    modelId: params.modelId.trim(),
     modelApi: params.modelApi.trim().toLowerCase(),
     modelBaseUrl: params.modelBaseUrl.trim(),
     requestTransportOverrides: params.requestTransportOverrides,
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2fb44fd507797fdbc19200adbde22c28347992db](https://github.com/openclaw/openclaw/pull/143780/commits/2fb44fd507797fdbc19200adbde22c28347992db)
- Landed commit: [4bf86a306c4c71e9b9cbe613a261bcba29417668](https://github.com/openclaw/openclaw/commit/4bf86a306c4c71e9b9cbe613a261bcba29417668)

PR #143782, opened 2026-09-10
fix(test): close CLI fixture databases before deleting state

steipete:
Related: #139428

## What Problem This Solves

Long Linux test runs can terminate when a CLI preparation fixture deletes its state directory while the shared-state SQLite connection is still open. The WAL identity guard later sees a deleted sidecar and correctly terminates the process.

## Why This Change Was Made

Close each fixture-owned shared-state database after agent database closure releases its lease, and before removing the directory. Resolve those paths from recorded fixture roots so cleanup preserves unrelated open databases. Remove a redundant per-case directory deletion that bypassed the fixture's teardown owner.

## User Impact

CLI preparation tests release their SQLite resources safely. Runtime SQLite protection and production behavior remain unchanged.

## Evidence

A deterministic regression creates two fixture roots and an unrelated writable state owner. Its removal observer checks real handle closure before forwarding any unlink. Against the original helper it fails at the live-handle assertion without deleting the live database files; test-owned final cleanup then closes them safely. No uninstrumented original full-suite pass is claimed.

All five complete consumer suites pass after the repair: 202 tests, zero failures or skips. The non-isolated single-worker run completes prepare.test.ts before prepare.rooted-execution.test.ts, preserving the observed failure-run order. All existing case identities and assertions remain intact.

The mapped changed gate and independent review validate the final change. The first gate caught unchecked indexes in the new regression; pairing each directory with its database removed those accesses, and the final original-red/candidate-green proof was repeated. This adds 52 net test/support lines and makes no runtime speedup claim.

Exact-head CI passed on attempt 2 of run 34446810813. Attempt 1 failed in the unrelated package-boundary compiler test when tsgo execve aborted with ETXTBSY before compilation. The single retry reran only compact-small-19 and its dependent gate and passed without source changes. The busy inode's underlying cause remains unconfirmed; the original failure is retained.

### DIFF
--- a/src/agents/cli-runner.test-helpers.ts
+++ b/src/agents/cli-runner.test-helpers.ts
@@ -18,6 +18,8 @@ import {
 import type { CliBackendPlugin } from "../plugins/cli-backend.types.js";
 import { parseAgentSessionKey } from "../routing/session-key.js";
 import { closeOpenClawAgentDatabaseByPath } from "../state/openclaw-agent-db.js";
+import { closeOpenClawStateDatabaseByPath } from "../state/openclaw-state-db.js";
+import { resolveOpenClawStateSqlitePath } from "../state/openclaw-state-db.paths.js";
 import {
   prepareSystemAgentRunAdmission,
   type PreparedAgentRunAdmission,
@@ -429,6 +431,9 @@ export function createCliRunnerPrepareFixture(prepareCliRun: PrepareCliRun) {
       }
       databasePaths.clear();
       for (const dir of tempDirs) {
+        closeOpenClawStateDatabaseByPath(
+          resolveOpenClawStateSqlitePath({ OPENCLAW_STATE_DIR: dir }),
+        );
         fs.rmSync(dir, { recursive: true, force: true });
       }
       tempDirs.clear();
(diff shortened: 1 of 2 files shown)


steipete:
The compact-small-19 concern is resolved for this head. The failed log identifies `test/scripts/check-extension-package-tsc-boundary.test.ts` during its first real compiler launch: tsgo `execve` aborted with `ETXTBSY` before compilation. It did not fail the CLI fixture regression.

After the original run completed, one retry of that exact job and its dependent gate passed without source changes. [CI run 34446810813, attempt 2](https://github.com/openclaw/openclaw/actions/runs/34446810813/attempts/2) is successful at `fb6b19b2963dc5a60674281d1624bc09b1973991`. The underlying busy-inode cause remains unconfirmed; no check or assertion was weakened.

steipete:
Merged via squash.

- Prepared head SHA: [fb6b19b2963dc5a60674281d1624bc09b1973991](https://github.com/openclaw/openclaw/pull/143782/commits/fb6b19b2963dc5a60674281d1624bc09b1973991)
- Landed commit: [918c3d212590478ca9cc11669f3b5ccd3ca2936f](https://github.com/openclaw/openclaw/commit/918c3d212590478ca9cc11669f3b5ccd3ca2936f)

PR #143795, opened 2026-09-10
fix: stop oversized worker manifest reads early

steipete:
Related: #134931

## What Problem This Solves

Fixes excessive worker filesystem reads when files grow after workspace inventory or while a completed turn's manifest is being captured. An oversized result could be rejected only after hashing its contents.

## Why This Change Was Made

The existing sequential capture now accounts for each opened file's current size, including memo hits, alongside symlink target bytes before reading. A streaming file that exceeds its opened size fails before the extra chunk is hashed. The existing 4 GiB limit, manifest format, memo identity, and cleanup owners are preserved.

## User Impact

Growing or oversized workspaces fail promptly with the existing diagnostic, without publishing a partial manifest. Stable files that grow or shrink before opening remain supported within the total limit.

## Evidence

- Two regressions fail on the original code; 37 focused and sibling cases pass after the repair, covering memo hits, aggregate boundaries, pre-open growth/shrinkage, streaming growth, and stream/descriptor cleanup.
- A real Node 24.20.0 run of the generated script at its normal limit used a sparse file grown to 4,294,967,297 bytes (4,096 bytes allocated). The original reached its first content-read attempt and was stopped by the fixture; the fix rejected the byte limit with zero content-read attempts and a closed handle. No multi-gigabyte read was performed.
- Fresh independent review is clean through P2, and the full changed-file gate passed, including production/test typechecks and selected repository guards.

This is a two-file extraction with 13 added production lines. Runtime evidence uses the actual generated remote script and local filesystem; it is not a cloud-provider experiment.

### DIFF
--- a/src/gateway/worker-environments/workspace-sync-inventory-limits.test.ts
+++ b/src/gateway/worker-environments/workspace-sync-inventory-limits.test.ts
@@ -4,6 +4,7 @@ import path from "node:path";
 import { afterEach, expect, it } from "vitest";
 import { useAutoCleanupTempDirTracker } from "../../../test/helpers/temp-dir.js";
 import { runCommandWithTimeout } from "../../process/exec.js";
+import { workspaceStatIdentity } from "./workspace-hash-memo.js";
 import {
   MAX_WORKSPACE_GIT_CANDIDATES,
   MAX_WORKSPACE_INVENTORY_ENTRIES,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [dbfbb6bac9c4a2994ee0a9fc7a7649e54f5a8b97](https://github.com/openclaw/openclaw/pull/143795/commits/dbfbb6bac9c4a2994ee0a9fc7a7649e54f5a8b97)
- Landed commit: [d491797e6b7975bde6a9d337337a68480449a27d](https://github.com/openclaw/openclaw/commit/d491797e6b7975bde6a9d337337a68480449a27d)

PR #143805, opened 2026-09-10
fix(gateway): restore CPU diagnostics on Linux Bun

### LINKED ISSUE #143800: Bug: CPU and event-loop delay are missing on Linux Bun
### Bug

The Control UI's **System busyness** overlay permanently shows dashes for CPU and event-loop delay on Linux with Bun 1.4.2. Memory and disk readings remain available.

### Reproduction and cause

Start a Gateway with Bun 1.4.2 on Linux, wait past the first sampling window, and open System busyness. The `status` response has no `eventLoop` sample.

The event-loop sampler requests a histogram upper bound of `2^63 - 1`. On the observed Linux x64 runtime, `createHistogram` throws `TypeError: Failed to initialize histogram`; the existing catch disables its timer. The same options work with Bun 1.4.2 on macOS, which masks the platform defect in local tests. Using the standard histogram range succeeds on Linux.

### Expected behavior

CPU and delay readings should appear after the first completed sampling window. CPU should reflect process CPU time, including load and recovery, and the Bun runtime smoke test should reject missing telemetry.

steipete:
Closes #143800.

## What Problem This Solves

Fixes an issue where users opening **System busyness** saw permanent dashes for CPU and event-loop delay when their Gateway ran on Linux Bun 1.4.2. Memory and disk readings continued to work.

## Why This Change Was Made

The sampler requested a histogram upper bound of `2^63 - 1`, which fails initialization on the observed Linux Bun runtime and silently disables sampling. Use the standard histogram range (about 104 days) and a numeric lower bound compatible with Node. Sampling cadence, CPU units, and degradation classification stay unchanged. No persistent data model or schema changes: the smoke assertion only reads a task-owned JSON capture of the existing RPC.

## User Impact

CPU and event-loop delay appear after the first completed sampling window. The Bun runtime smoke flow now requires finite sampled telemetry through the installed Gateway's `status` RPC.

## Evidence

- Reproduced on Linux x64 / Bun 1.4.2: original histogram initialization throws `TypeError: Failed to initialize histogram`; the original sampler returns no snapshot.
- Final sampler on the same Linux runtime: CPU 0.7% idle → 93.4% during a controlled 1.25-second CPU load → 0.4% after recovery; delay reports the stall and clears its degraded state after recovery. The new diagnostics assertion rejects the original sample and accepts the repaired sample.
- Real isolated Gateway with the rebuilt Control UI, verified through Chrome: CPU and delay values render and their sparklines accumulate samples. The production Gateway was inspected without restart.
- 19 sampler tests and 122 installer-tooling tests pass. The numeric lower bound also preserves Node compatibility with the numeric default upper bound.
- Independent review through P2: no actionable findings.
- Full build, rebuilt Control UI, and `pnpm check:changed` pass on the committed source. [Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34448659262). Full packaged Bun install smoke has not been rerun in this task.

### DIFF
--- a/docs/gateway/health.md
+++ b/docs/gateway/health.md
@@ -83,6 +83,11 @@ core equivalents: `1` means one CPU core fully occupied over the interval, and
 parallel work can produce values above `1`. It is not a percentage of the host's
 total CPU capacity.
 
+The Control UI's **System busyness** overlay reads the same sampler through
+`status.eventLoop` on both Node and Bun. Its CPU percentage uses `100%` for one
+fully occupied core. CPU and delay show a dash until the first sample completes;
+a persistent dash means the telemetry is unavailable, not zero CPU usage.
+
 Event-loop delay and utilization describe the main thread separately. A `cpu`
 degradation reason reports process CPU pressure with delay co-evidence; it does
 not identify the thread consuming CPU or prove a main-thread hang. Inspect the
--- a/scripts/e2e/bun-global-install-smoke.sh
+++ b/scripts/e2e/bun-global-install-smoke.sh
@@ -46,6 +46,7 @@ MOCK_REQUEST_LOG=""
 LOCAL_AGENT_LOG=""
 GATEWAY_LOG=""
 GATEWAY_HEALTH_LOG=""
+GATEWAY_STATUS_LOG=""
 GATEWAY_AGENT_LOG=""
 DIRECT_BUN_LOG=""
 
@@ -74,6 +75,7 @@ dump_debug_logs() {
     "$LOCAL_AGENT_LOG" \
     "$GATEWAY_LOG" \
     "$GATEWAY_HEALTH_LOG" \
+    "$GATEWAY_STATUS_LOG" \
     "$GATEWAY_AGENT_LOG" \
     "$DIRECT_BUN_LOG" >&2 || true
 }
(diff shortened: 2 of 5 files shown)


steipete:
The data-model compatibility checklist item is a false positive. `assert-gateway-diagnostics` only reads a task-owned JSON capture of the existing `status` RPC and checks numeric measurements. It does not write or change persistent runtime state, a SQLite schema, a migration, a serialized-state contract, or retained fields. The production change only replaces histogram construction options; the response shape and sampling lifecycle are unchanged. No migration or upgrade-data conversion is needed for this patch.

The fresh full build passed on the committed head. The 19 sampler tests, 122 installer-tooling tests, Linux Bun idle/load/recovery probe, and real Gateway/Chrome diagnostics check also passed. Remaining local guards and exact-head CI are still being completed before native landing.

steipete:
Merged via squash.

- Prepared head SHA: [be9156af6bedb2dd7dd09302cd90c64b34085ca0](https://github.com/openclaw/openclaw/pull/143805/commits/be9156af6bedb2dd7dd09302cd90c64b34085ca0)
- Landed commit: [84b9cf4657547dd49ed0f6c55a85b329198b98db](https://github.com/openclaw/openclaw/commit/84b9cf4657547dd49ed0f6c55a85b329198b98db)

PR #143399, opened 2026-09-09
improve(android): make reply notifications session-aware across devices

pestelence2k:
Closes #143387

## What Problem This Solves

Improves the existing Android reply-notification system: foreground suppression currently hides completions from other sessions, generic titles obscure the originating agent/chat, and laptop readers receive duplicate phone alerts for replies they are already viewing. Internal Dreaming narration should not produce ordinary conversation alerts.

## Why This Change Was Made

Reuses the existing session-viewer owner with a 30-second server-owned lease, renewed every 10 seconds while Control UI is visible, focused, and within two minutes of user interaction. Blur/hidden/unwatch clears viewing; disconnect or missed renewal expires it.

For final chat events, the Gateway derives an optional `suppressNotification` hint separately for each Android operator recipient after normal delivery gates. It requires an exact canonical session and matching authenticated profile, ignores producer-supplied hints, and defaults to notifying when presence or identity is uncertain. Android consumes the hint only at notification publication, preserving terminal handling and history synchronization. Swift protocol models are regenerated from the schema.

This does not add suppression support to iOS/macOS notifications.

### Existing notification system and scope

Upstream already owns Android conversation-reply notifications (`ConversationReplyNotifier`), including permission checks, private lockscreen content, and routed notification actions. This PR improves that existing system rather than introducing another notifier:

- Replace app-wide foreground suppression with exact visible Gateway/agent/session matching, with chat-screen disposal cleanup.
- Label notifications and shortcuts with the configured agent name and originating session title, using agent ID / Chat fallbacks.
- Suppress canonical internal Dreaming narration without blanket filtering scheduled work or background sessions.
- Add same-user cross-device suppression through the viewing lease and notification-only hint described above.

No personal agent aliases, unrelated custom app features, configuration, persistent schema, version changes, APKs, or deployment overlays are included. Notification taps/replies retain original routing, and the public lockscreen presentation remains generic.

## User Impact

- Reading a session in Control UI silences redundant phone alerts for that session.
- Other sessions notify even while Android is open on a different chat or non-chat page.
- Agent/session labels make the source recognizable without exposing it in the public lockscreen version.
- Internal Dreaming narration stays silent; normal scheduled/background replies remain eligible.
- Leaving or becoming idle in Control UI restores phone notifications after the declaration clears or expires.
- Messages continue to synchronize regardless of whether an alert is published.

## Evidence

[...]

PR #142651, opened 2026-09-08
fix(mobile): show failures and recovery for unreachable gateways

vincentkoc:
## What Problem This Solves

Fixes an issue where users pairing the Android or iOS app with an unreachable gateway could remain in Connecting or receive an error without a useful recovery path. This is particularly confusing when a Tailscale gateway is offline or the phone cannot reach its tailnet.

## Why This Change Was Made

Android connection attempts now own their preflight, transport and cleanup deadlines, and publish a failure for the current target without allowing canceled work to replace a newer attempt. iOS preflight failures use the existing typed problem presentation, and Retry retains the failed target and unconsumed setup intent instead of silently returning to a previously saved gateway.

Tailscale endpoint guidance links to official setup help without claiming that the VPN is off or required. Certificate trust, setup fingerprints, credential handoff and pairing approval remain separate; this does not change VPN settings, relax TLS/authentication, add configuration, or increase existing iOS timeout budgets.

## User Impact

- Unreachable gateway attempts have an actionable failure and a separate Retry action.
- Retrying setup preserves its target, context path and unconsumed security constraints.
- Canceling a replacement restores the previous route when appropriate; stale callbacks cannot restore an older target or overwrite a newer attempt.

## Evidence

**Exact-head CI is green on `1dd803e24844cc9cdf49e0a2f33da5d4821c9715`: 88 successful jobs, 11 intentional skips, zero failures or cancellations. Focused Android/iOS validation, the full macOS test lane, targeted native UI captures, and 35/35 localization tests passed.**

**CI follow-up:** the [previous ready-state run](https://github.com/openclaw/openclaw/actions/runs/34298150549) exposed three stale localization-test literals, now corrected without weakening the multiline-kind, full-message, or indexed-placeholder assertions. Its Android JVM also crashed in Temurin 21.0.10+7 HotSpot (`SIGSEGV`, `Node::uncast(bool)`, exit 134). The new-head run then encountered a null-decode assertion in an unchanged image-preview test. An approved 27-test class diagnostic passed, followed by one successful Play-job retry. Two type-check jobs that never acquired runners were retried individually; passed native jobs were preserved. No Android source, assertions, JVM flags, or workflow settings changed for these failures. The diagnostic used macOS/Microsoft JDK 21.0.11 rather than CI's Linux/Temurin environment, so it is not a claim of branch-independent flakiness.

| Surface | Observed result |
| --- | --- |
[...]

PR #143812, opened 2026-09-10
fix(android): keep notification actions on their gateway

steipete:
Related: #134939, #142651

Draft #143399 addresses notification eligibility, labels, and cross-device suppression. This PR leaves those product choices unchanged and repairs notification action/lifecycle ownership.

## What Problem This Solves

Fixes an issue where opening or replying to an Android conversation notification could interrupt an accepted Gateway connection or TLS approval, leave reply status unclear, or let an older reply result replace a newer notification.

## Why This Change Was Made

Queued UI requests and accepted Gateway connections have different lifetimes. Target selection now preserves an accepted connection while superseding obsolete UI work; an explicit replacement or Disconnect still retires the previous attempt immediately. Readiness comes from the selected operator connection, with the existing queue deadlines, single TLS worker, and transport cleanup retained.

Reply feedback belongs to the notification publication that produced the action. Queue admission is distinguished from delivery and from an unknown outcome. Already queued input keeps its existing recovery behavior. No configuration, protocol, schema, or dependency change is introduced.

## User Impact

- Opening a saved conversation reconnects when needed without restarting its accepted attempt. An unavailable target opens Gateway settings with a visible explanation and leaves another Gateway alone.
- Inline replies show **Reply queued**, **Chat failed**, or an unknown-status message with **Open conversation**. Stale results cannot overwrite a newer notice; retained notices from older versions can still send without rewriting current feedback.
- Dismissed node-approval feedback stays closed during background refreshes. An explicit check can show the waiting result again.

## Evidence

- 575 tests across 20 Android suites passed, including notification routing, foreground lifetime, Gateway readiness, TLS-worker/socket cleanup, and onboarding.
- Play and ThirdParty Kotlin compilation and Android lint passed. The actual changed checks passed, including app, benchmark, Wear, and shared Kotlin style checks.
- An integration regression reproduced an older TLS result replacing a newer queued request's progress; the owner-aware status update and its sibling cases pass.
- The onboarding test now uses a real client connection to a loopback fixture and controlled node-list responses. It passes with the fix and fails when the original dialog-reset behavior is restored.

- Independent P0–P2 review completed with no actionable findings.
- Installed API36 proof used the verified Play debug APK from `283e47d03de6e09949c9cc85c3ce5f362fa05216` (SHA-256 `989fdf7ab2407125c252dedd91b640df1dde281cc8077f384fecd95531634c75`). The real SystemUI Reply action showed queued feedback, the submitted preview, and Open conversation before the synthetic fixture accepted or acknowledged its one received reply. The unchanged timing limit was preserved.
- The foreground notification's Disconnect action closed both roles. Tapping the retained conversation notice reconnected and opened the same session with all four canonical messages. The seed and reply each reached the fixture once. App data, reverse mapping, private fixture materials, and emulator were cleaned up.

CI follow-up: the first hosted run completed 3,171 Play tests and 3,302 ThirdParty tests, with one assertion failure in each. The cleanup test could reuse the earlier Disconnect cancellation signal before its later switch entered cleanup. Its existing fixture now waits for that switch’s distinct completed operator cleanup tail, retaining every deadline, no-probe, and final-result assertion and the same wait bound. Both complete owner suites pass in both variants (372 test executions), and incremental review is clean. The unrelated composer failure was at the final tap-to-branch-completion wait after its scope assertions passed; that test remains unchanged and passes locally. The correction changes only test synchronization, so the installed APK evidence above remains exact for the unchanged production code. [Hosted CI 34452173834](https://github.com/openclaw/openclaw/actions/runs/34452173834) is now successful on `294aad6692cae3e1065ebc10fa1b2524c0deef27` (attempt 2), including the required gate. Play’s initial unit and lint invocations both succeeded, but the job crossed its 20-minute deadline at completion; one unchanged Play-job retry passed in 18m55s without changing coverage or the deadline. The full local Play suite also passed 3,171 tests across 231 suites with zero failures, errors, or skips. The original failures and cancellation remain recorded; no further source correction was needed.

The screenshots show the initial notice, feedback after Reply, and recovered conversation—not a before/after software-version comparison. Installed proof uses a synthetic loopback Gateway, not a real provider or production pairing verifier. Stale-publication ordering is covered by unit regressions and source review; this run does not claim an installed ordering or crash-durability proof.

Initial conversation notification before Reply

Queued reply feedback before fixture acknowledgement

Same conversation recovered after Disconnect and notification Open


steipete:
Merged via squash.

- Prepared head SHA: [294aad6692cae3e1065ebc10fa1b2524c0deef27](https://github.com/openclaw/openclaw/pull/143812/commits/294aad6692cae3e1065ebc10fa1b2524c0deef27)
- Landed commit: [025d572de58f0895afd7ffcee00ab53797e190ff](https://github.com/openclaw/openclaw/commit/025d572de58f0895afd7ffcee00ab53797e190ff)

PR #143777, opened 2026-09-10
fix: retain executable model identities in catalogs

steipete:
Related: #130706, #142100.

## What Problem This Solves

Fixes an issue where building a model catalog normalizes an executable registry ID again as authored input. With aliases `latest → middle → final`, an existing `middle` model can become `final`, carrying its capabilities with it. Repeating that normalization in the account-visible set also prevents valid metadata enrichment of the selected model.

## Why This Change Was Made

Keep emitted registry IDs literal and derive enrichment eligibility directly from emitted identities. Authored supplemental IDs still normalize at their existing boundary. Provider equivalence, suppression, ordering, and exclusion of unobserved models remain intact. Two production hunks remove four lines.

## User Impact

The selected registry model retains its identity and capabilities. Runtime catalog enrichment applies to that model without introducing models the account did not expose.

## Evidence

Both original regressions failed on the baseline; applying only the registry-ID change left the enrichment failure intact. The reviewed patch was replayed without conflicts or changes onto main containing the shared test-fixture repairs in #143813 and #143875 and the UI lint repair in #143889. The earlier failures were in those baseline checks; the catalog patch remains unchanged.

Fresh validation at `1c32db0c8fb4f5d3fe1c6e1073c944fe794bb39c` passed 56 catalog/route tests, the full changed-file gate, and one canonical `sourcePerformance` runtime build. The independent P2 review remains applicable to the identical patch.

The freshly compiled catalog used a real registered provider augmentation hook: `middle` stayed literal and enriched from 32K to 64K on the same API route; vision remained intact; the unobserved model stayed excluded; borrowed inputs stayed unchanged; registrations were removed by rollback. The refreshed proof passed on its first attempt. It made no external provider request. This runtime build profile excludes UI and declaration generation. Original source, builds, and failure evidence were retained with their original revision identities.

PR #143813, opened 2026-09-10
test: serialize database probe teardown

steipete:
Related: #143777.

## What Problem This Solves

The concurrent database-initialization test can fail during worker cleanup: one process retires its handle while another still owns the state-lifecycle lock. Explicit retirement intentionally rejects contention, so this cleanup race obscures the initialization behavior being tested.

## Why This Change Was Made

Keep all worker opens concurrent and retain the held-coordinator assertion. Wait for every worker to report a successful open and integrity check, then permit retirement one worker at a time and join its actual close before proceeding. Preserve both the original failure and a cleanup failure when both occur. This changes one test file only; production locking and timeout values are unchanged.

## User Impact

CI tests concurrent initialization without racing unrelated teardown. There is no product behavior change.

## Evidence

The original failure is recorded in [CI job 102769936669](https://github.com/openclaw/openclaw/actions/runs/34445629228/job/102769936669). The original helper is byte-identical at that run and the repair base. One unchanged local baseline passed, consistent with a scheduling-dependent race.

The corrected fresh-database and additive-upgrade probes both ran real child processes successfully. Retained markers show all successful opens precede retirement permission, and each worker exits before the next receives permission. The full file passed 190 tests; its one Linux-only case remains skipped on macOS. Independent P2 review and the full changed-file gate passed. No production schema, concurrency, retry, or timeout change is included.

### DIFF
--- a/src/state/openclaw-state-db.test.ts
+++ b/src/state/openclaw-state-db.test.ts
@@ -1199,14 +1199,17 @@ function runConcurrentSchemaProbe(params: {
     const readyPath = process.env.OPENCLAW_SCHEMA_TEST_READY_PATH;
     const startPath = process.env.OPENCLAW_SCHEMA_TEST_START_PATH;
     const workerIndex = process.env.OPENCLAW_SCHEMA_TEST_WORKER_INDEX;
-    fs.writeFileSync(readyPath, "ready");
-    const deadline = Date.now() + 15_000;
-    while (!fs.existsSync(startPath)) {
-      if (Date.now() >= deadline) {
-        throw new Error("timed out waiting for concurrent schema upgrade start");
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4beaf2718b2b368491cc800bd48a51f9dcd468d5](https://github.com/openclaw/openclaw/pull/143813/commits/4beaf2718b2b368491cc800bd48a51f9dcd468d5)
- Landed commit: [6f0c85eb9a6e6ffdacff421e34dce19ba83f5d73](https://github.com/openclaw/openclaw/commit/6f0c85eb9a6e6ffdacff421e34dce19ba83f5d73)

PR #143819, opened 2026-09-10
fix(deepseek): preserve object-union tool arguments

### LINKED ISSUE #143790: DeepSeek tool-schema normalisation drops all but the first `anyOf` variant for object unions, making MCP tools with union parameters uncallable
## Summary

`normalizeDeepSeekSchema` rewrites an `anyOf` whose variants are **objects** (rather than string `const` literals) down to the **first variant**. The const-union case was fixed in #86468, and that branch is present and correct; the object-variant case still falls through to `nonNullVariants[0]`.

The emitted schema is therefore narrower than anything the MCP server published, so calls are rejected **client-side** and never reach the server. With the Notion MCP server this makes `notion-create-pages` unable to create a row in a database at all, because `parent` is a three-way object union and only the `page_id` branch survives.

## Environment

- OpenClaw `2026.9.3` (1391f7c)
- Provider/model: `deepseek/deepseek-v4-flash`
- macOS 26.6.2 (arm64), Node `v26.7.0`
- MCP server: `notion`, remote streamable-http at `https://mcp.notion.com/mcp`

## What the MCP server publishes

Verbatim from `tools/list`, tool `notion-create-pages`. `parent` is a three-way union of objects:

```json
"parent": {
  "description": "The parent under which the new pages will be created. This can be a page (page_id), a database page (database_id), or a data source/collection under a database (data_source_id).",
  "anyOf": [
    { "type": "object",
      "properties": { "page_id": { "type": "string" }, "type": { "type": "string", "enum": ["page_id"] } },
      "required": ["page_id"], "additionalProperties": {} },
    { "type": "object",
      "properties": { "database_id": { "type": "string" }, "type": { "type": "string", "enum": ["database_id"] } },
      "required": ["database_id"], "additionalProperties": {} },
    { "type": "object",
      "properties": { "data_source_id": { "type": "string" }, "type": { "type": "string", "enum": ["data_source_id"] } },
      "required": ["data_source_id"], "additionalProperties": {} }
  ]
}
```

## What OpenClaw hands the model

The tool definition the model receives has `parent` as:

```json
"parent": {
  "description": "The parent under which the new pages will be created. This can be a page (page_id), a database page (database_id), or a data source/collection under a database (data_source_id).",
  "properties": {
    "page_id": { "type": "string", "description": "The ID of the parent page (with or without dashes), for example, 195de9221179449fab8075a27c979105" },
    "type": { "type": "string", "enum": ["page_id"] }
  },
  "required": ["page_id"],
  "type": "object"
}
```

`database_id` and `data_source_id` are gone, `type` is narrowed to a single allowed value, and `required` still demands `page_id`. Calling with the data-source branch fails locally:

```
Validation failed for tool "notion__notion-create-pages":
  - parent.page_id: must have required properties page_id
  - parent.type: must be equal to one of the allowed values
```

Sibling tools show the same collapse. `notion-move-pages` publishes `new_parent` as a four-way object union (`page_id`, `database_id`, `data_source_id`, `workspace`) and the model only ever sees the `page_id` branch.

## Root cause

`normalizeDeepSeekSchema`, in the bundled `provider-tools` chunk and wired through `buildProviderToolCompatFamilyHooks("deepseek")`:

```javascript
const normalizedVariants = record[unionKey].map((entry) => normalizeDeepSeekSchema(entry));
const nonNullVariants = normalizedVariants.filter((entry) => !isNullSchemaVariant(entry));
const hasNullVariant = nonNullVariants.length < normalizedVariants.length;
if (nonNullVariants.length > 1 && nonNullVariants.every((entry) => isStringConstVariant(entry))) {
  const enumValues = nonNullVariants.map((entry) => entry.const);
  const merged = { ...normalized, type: "string", enum: enumValues };
  if (hasNullVariant) merged.nullable = true;
  return merged;
}
const selected = nonNullVariants[0] ?? normalizedVariants[0];
if (!selected || typeof selected !== "object" || Array.isArray(selected)) return normalized;
const merged = { ...selected, ...normalized };
if (hasNullVariant) merged.nullable = true;
return merged;
```

The `isStringConstVariant` branch, which is the fix from #86468, covers unions of string literals. Object variants skip it and reach `selected = nonNullVariants[0]`. Because `normalized` holds only the keys that sit beside the union keyword (here just `description`), the result is the first variant plus the union's shared annotations. Every later variant is silently discarded, and the union keyword has to go because DeepSeek rejects `anyOf` and `oneOf`.

## Suggested fix

When the variants are objects that cannot be collapsed into a single accepted shape, emitting the first variant is worse than emitting a permissive schema: the result is narrower than the server's contract, so it rejects calls the server would accept. Two options, in preference order:

1. If the object variants differ only in which key is required, merge the properties and set `required` to the intersection across variants, which is empty in the `parent` case, so the model can express any branch. This mirrors what the MCP loopback path already does in `flattenUnionSchema`.
2. Otherwise drop the union and emit a permissive schema for that property, either `{ "type": "object" }` or the merged superset of the variants, and let the server validate. That is what #83361 asked for at the gateway level.

## Why this is not a duplicate

- #86468 fixed the **const-union** case only, and that branch is present and working.
- #128743 covered a **root-level** `anyOf` sitting beside `properties`.
- #78737 reported the same Notion `parent` symptom and was closed as completed, but the failure is reproducible on `2026.9.3`, so it was not fixed for this shape.
- #83361 asked for union stripping at the gateway level rather than in one provider family.
- #113130 is the same class of defect on the Moonshot/Kimi family, so the fix is worth making in a shared place rather than per provider.

This case is a **nested** `anyOf` of **object** variants, where the const branch does not apply and no root-level flattening runs.

## Impact

Any MCP or plugin tool whose parameter is a discriminated union of objects becomes partially or wholly uncallable on the DeepSeek path. For the Notion MCP server that is creating a page inside a database, which is the standard "capture a row" operation, and there is no other MCP tool that can do it. The failure is silent at the server side, because the request never leaves the client.

Workaround, for anyone hitting the same thing: call the API directly with the same credential instead of going through the tool, which works, and confirms this is schema normalisation rather than a server capability.

azuretek:
## What Problem This Solves

DeepSeek kept only the first object alternative in MCP tool schemas. Valid Notion-style database and data-source parent arguments then failed local validation before reaching the tool.

Merge supported object properties, keep common required keys, and pool compatible discriminator literals. Preserve permissive keys, descriptions, and own prototype-shaped keys. Keep object/null alternatives usable by the local validator, including literal and outer constraints. Shared type-array coercion now preserves genuine null without turning invalid non-null values into null; valid conversions and null-only behavior remain supported.

The flattened schema remains an approximation. Conflicting property shapes retain the first definition, shared annotations retain the first annotation, and the tool owner still enforces original whole-object constraints that the flattened schema cannot express. Primitive and mixed-union selection remains unchanged.

## Evidence

- 86 focused owner and sibling tests pass; selecting the production normalizer runs all 40 provider cases.
- Independent public Gateway validation covers 125 cases with bundled DeepSeek/OpenAI routing, captured model requests and actual MCP receipts. Baseline rejects the valid later parent branches; candidate delivers all three exactly once. Negative cases remain local rejections or explicit tool-owner errors, as applicable.
- Runtime proof uses commit `a243f9793987b9091a356ea984b43107dbdd180c`. The later assertion-baseline count reduction changes no runtime, dependency, test, or fixture input.
- The hosted candidate build and artifact checks passed: https://github.com/openclaw/openclaw/actions/runs/34500899684/job/102951656517. Full PR CI and exact-head review remain separate gates.

The public validation uses synthetic loopback providers and a local MCP server. The contributor's earlier live Notion report recorded successful data-source page creation and archival; that remains attributed contributor evidence. All six contributor commits are preserved.

Fixes #143790.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3239,7 +3239,7 @@ src/plugin-sdk/provider-entry.ts	3
 src/plugin-sdk/provider-onboard.ts	5
 src/plugin-sdk/provider-stream-shared.ts	14
 src/plugin-sdk/provider-stream.ts	1
-src/plugin-sdk/provider-tools.ts	10
+src/plugin-sdk/provider-tools.ts	9
 src/plugin-sdk/qa-channel-protocol.ts	1
 src/plugin-sdk/qa-channel.ts	15
 src/plugin-sdk/qa-lab.ts	2
--- a/packages/llm-core/src/validation.test.ts
+++ b/packages/llm-core/src/validation.test.ts
@@ -30,7 +30,7 @@ const decimalTools = [
 ];
 
 describe("validateToolArguments", () => {
-  it.each(["anyOf", "oneOf", "TypeBox"])(
+  it.each(["anyOf", "oneOf", "TypeBox", "type-array integer/null", "type-array null/integer"])(
     "keeps invalid non-null values out of a nullable integer %s",
     (union) => {
       const tool: Tool = {
@@ -44,7 +44,12 @@ describe("validateToolArguments", () => {
             : {
                 type: "object",
                 properties: {
-                  limit: { [union]: [{ type: "integer", minimum: 1 }, { type: "null" }] },
+                  limit:
+                    union === "type-array integer/null"
+                      ? { type: ["integer", "null"], minimum: 1 }
+                      : union === "type-array null/integer"
+                        ? { type: ["null", "integer"], minimum: 1 }
+                        : { [union]: [{ type: "integer", minimum: 1 }, { type: "null" }] },
                 },
               },
       };
@@ -70,6 +75,115 @@ describe("validateToolArguments", () => {
       }
     },
   );
+
+  it.each([
+    { name: "object/null", types: ["object", "null"] },
+    { name: "null/object", types: ["null", "object"] },
+  ])("keeps invalid non-null values out of a $name type array", ({ types }) => {
+    const tool: Tool = {
+      name: "nullable-parent",
+      description: "Accept an object or null without replacing rejected scalar input",
+      parameters: {
+        type: "object",
+        properties: {
+          parent: {
+            type: types,
+            properties: {
+              kind: { type: "string", enum: ["page"] },
+              count: { type: "integer", minimum: 1 },
+            },
+            required: ["kind", "count"],
+            additionalProperties: false,
+          },
+        },
+        required: ["parent"],
+        additionalProperties: false,
+      },
+    };
+    const validate = (parent: unknown) =>
+      validateToolArguments(tool, {
+        type: "toolCall",
+        id: "nullable-parent-call",
+        name: tool.name,
+        arguments: { parent },
+      });
+
+    expect(validate(null)).toEqual({ parent: null });
+    const parent = { kind: "page", count: "2" };
+    for (const input of [parent, JSON.stringify(parent)]) {
+      expect(validate(input)).toEqual({ parent: { kind: "page", count: 2 } });
+    }
+    for (const input of [false, 0, "", [], { kind: "wrong", count: 2 }]) {
+      expect(() => validate(input)).toThrow(/Validation failed for tool "nullable-parent"/);
+    }
+  });
+
+  const numericConversions = [
+    { input: "2.5", output: 2.5 },
+    { input: false, output: 0 },
+    { input: 0, output: 0 },
+  ];
+  const stringConversions = [
+    { input: false, output: "false" },
+    { input: 0, output: "0" },
+    { input: "", output: "" },
+    { input: "existing", output: "existing" },
+  ];
+
+  it.each([
+    { name: "number/null", types: ["number", "null"], conversions: numericConversions },
+    { name: "null/number", types: ["null", "number"], conversions: numericConversions },
+    { name: "string/null", types: ["string", "null"], conversions: stringConversions },
+    { name: "null/string", types: ["null", "string"], conversions: stringConversions },
+  ])("retains valid non-null coercions in a $name type array", ({ types, conversions }) => {
+    const tool: Tool = {
+      name: "nullable-value",
+      description: "Keep conversions into a supported non-null alternative",
+      parameters: {
+        type: "object",
+        properties: { value: { type: types } },
+        required: ["value"],
+      },
+    };
+    const validate = (value: unknown) =>
+      validateToolArguments(tool, {
+        type: "toolCall",
+        id: "nullable-value-call",
+        name: tool.name,
+        arguments: { value },
+      });
+
+    expect(validate(null)).toEqual({ value: null });
+    for (const { input, output } of conversions) {
+      expect(validate(input)).toEqual({ value: output });
+    }
+  });
+
+  it.each([
+    { name: "a null type", type: "null" },
+    { name: "a single-member null type array", type: ["null"] },
+  ])("preserves existing coercion for $name", ({ type }) => {
+    const tool: Tool = {
+      name: "null-only",
+      description: "Retain the existing null-only conversion contract",
+      parameters: {
+        type: "object",
+        properties: { value: { type } },
+        required: ["value"],
+      },
+    };
+
+    for (const value of [null, false, 0, ""]) {
+      expect(
+        validateToolArguments(tool, {
+          type: "toolCall",
+          id: "null-only-call",
+          name: tool.name,
+          arguments: { value },
+        }),
+      ).toEqual({ value: null });
+    }
+  });
 
   it.each(decimalTools)("coerces strict decimal numeric strings for $label", ({ tool }) => {
     expect(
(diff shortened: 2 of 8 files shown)


azuretek:
**Revision 2 pushed.** Both review findings are fixed and the branch is rebased onto current `main`.

SHA mapping, since the rebase replayed the original commit and the tracking ref had gone stale:

| Was | Now | What |
|---|---|---|
| `96e298e10` | `2854777f5` | the original commit, rebased onto `main` |
| | `99f60a43a` | review fixes: open-union keys and own-property lookups |

The reviewed revision stays reachable at `backup/deepseek-object-union-variants-20260910` (`96e298e10`).

**Finding 1, permissive branches.** A key that a variant accepts through `additionalProperties` without declaring it is now left unconstrained, keeping only its documentation and dropping what would constrain it. The regression asserts through the real `validateToolArguments` that `{a: "x", b: {nested: true}}`, which only the open variant accepted, still validates, and that the previously unreachable branch is callable too.

**Finding 2, own-property accumulator.** The accumulator now uses `Object.hasOwn`, matching the MCP loopback merger. The regression declares `constructor` and `toString`; against the flagged code it reported `expected [Function Object] to deeply equal { type: 'string' }`, which is exactly the dropped definition you described, while the name stayed in `required`.

Post-rebase state:

- `src/plugin-sdk/provider-tools.test.ts`: 21 passed with the change, 3 failed without the fix in `provider-tools.ts`.
- `tsgo --noEmit -p packages/plugin-sdk/tsconfig.json`: exit 0.
- `oxlint` on both changed files: 0 warnings, 0 errors.

The PR body now carries the normalization output, the before and after argument validation, and the live Notion call, including the production error reproduced from `main`'s schema for this tool.

azuretek:
**Revision 3 pushed.** Both new findings are fixed in one commit on top, no rebase, so every SHA from the previous comment still resolves: `e16307190` addresses pattern-matched keys and prototype-safe accumulation.

**Finding 1, pattern-matched keys.** A variant now counts as accepting a key it does not declare when its `additionalProperties` is not `false`, or when one of its `patternProperties` patterns matches the key. An uncompilable pattern widens nothing rather than making the merge throw.

The regression uses your exact shape: branch A declares and requires `a`, sets `additionalProperties: false`, and permits an object-valued `b` through `patternProperties: {"^b$": {"type": "object"}}`; branch B declares `b` as a string. The flattened `b` stays unconstrained, `{a: "x", b: {nested: true}}` still validates through the real `validateToolArguments`, and branch B stays callable.

**Finding 2, prototype-safe accumulation.** The accumulator is `Object.create(null)`, as the MCP loopback merger does, so writing a literal `__proto__` key no longer runs the inherited setter. The regression builds two alternatives with an own JSON-parsed `__proto__` property and asserts the own key survives, that the first definition is kept, and that it is present in the serialized schema.

Both regressions fail without their fix, confirmed by reverting each half separately: the pattern case fails at `props["b"]`, the prototype case at `Object.hasOwn(props, "__proto__")`.

Gates on this head:

- `src/plugin-sdk/provider-tools.test.ts`: 23 passed. The three other files that consume the normalizer: 146 passed.
- `oxlint` type-aware with the CI flags on both changed files: 0 warnings, 0 errors.
- `tsgo --noEmit -p packages/plugin-sdk/tsconfig.json`: exit 0.
- `check:assertion-safety` and `check:max-lines-ratchet`: exit 0.

One CI note on the previous head: `checks-fast-bundled-protocol` appears as failed, but its job conclusion is `cancelled` and its log blob does not exist, and that is what flips `openclaw/ci-gate`. The three failures from revision 2 now pass (`PR context and evidence`, `check-lint-core-1`, `checks-fast-baseline-ratchets`), as does `checks-node-compact-large-11`.

obviyus:
Thanks @azuretek! All six contributor commits are preserved.

**Proof:** 125 public Gateway cases with bundled DeepSeek/OpenAI routing, captured provider requests, and a local MCP server; 86 focused owner and sibling tests. These use synthetic loopback providers, not paid provider calls.

- **Bug:** DeepSeek dropped later object alternatives and rejected valid database and data-source parent arguments before the MCP tool received them.
- **Fix:** The existing schema adapter preserves supported alternatives and discriminator values. The shared validator preserves genuine null without converting invalid non-null values to null.

| Before | After |
| --- | --- |
| Valid database and data-source parent calls fail locally. | Page, database, and data-source calls each reach the same MCP tool once. |
| Invalid scalar values can become null in multi-type schemas. | Invalid values reject locally; genuine null and valid conversions remain supported. |

The flattened schema remains an approximation. The tool owner continues to reject original whole-object constraints that the flattened schema cannot express. Original failed observations and subsequent clarified checks are retained.

Merged as `b47f1848acf3b774ab27239ed8a299adbbc65eb0`. Exact-head ClawSweeper review passed with no findings. [Hosted CI](https://github.com/openclaw/openclaw/actions/runs/34506951511) passed, including its aggregate gate. The landed runtime, tests, and test-selection files match the reviewed candidate. Owned test resources are closed.

PR #143822, opened 2026-09-10
fix(models): use exact configured rows before aliases

steipete:
Related: #130706, #142100.

## What Problem This Solves

A configured model can receive another row's endpoint or request overrides because a legacy-prefixed or equivalent row appears first. Applying provider defaults before duplicate rows merge can also make transport disagree with early auth selection. The catalog adapter additionally passed a provider-qualified key where the resolver expects a model ID.

## Why This Change Was Made

Use one explicit configured-row resolver. Within the selected provider configuration, exact spelling wins and only same-spelling duplicates fill omitted fields; duplicates merge before provider defaults. Preserve explicit empty overrides and existing preferred provider-key selection. The catalog adapter passes the provider-owned logical model ID to the same owner.

Keep the existing normalized-index API separate: its callers still receive only normalized keys and their established first-row merge behavior. Literal priority is private to the explicit resolver. This eight-file repair adds 27 production lines and enables the separate context/inventory cleanups.

## User Impact

Configured requests use the selected row's endpoint, API, and overrides. A neighboring alias cannot donate omitted fields to an exact row. Public catalog overrides follow the same logical selection instead of accidentally choosing a prefixed legacy row.

## Evidence

At `b0d341e5d123bc5077ea1c3b55d44f487dac2bf6`, 312 focused cases, fresh independent P2 review, and the full changed-file gate passed. Coverage includes row precedence, catalog adapters, provider routes, compaction thresholds, harness projections, and an inbound-image flow. The assertion baseline contains only its exact shrink for removed casts.

Review of the previous revision exposed two introduced normalized-index regressions: an image-capable `Model` could select the text-only `model` row, and a 200k model could inherit a 1m row's compaction floor. Both were reproduced against that head, compared with the baseline owner, and repaired by restoring the normalized-index contract. The new two-case map regression table fails on the previous implementation. Existing case-distinct behavior of normalized consumers is preserved; migrating those consumers to literal selection is a separate follow-up.

One canonical `sourcePerformance` build passed at the exact head. Ten compiled loopback HTTP requests verified paths, wire IDs, and headers across exact/legacy rows, duplicate-before-default behavior, and Arcee ordering. Ten compiled logical-override cases verified precedence, omissions, duplicate merging, and legacy fallback; two compiled normalized-index controls verified the retained API. All 9,009 build files were unchanged, no source imports occurred, the listener closed, and child process groups exited. This runtime build excludes UI and declaration generation.

Earlier failures and evidence remain associated with their original heads. Requests to change cross-provider-key selection were checked against the existing contract and rejected; the final review included that contract and found no actionable issue.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -1713,7 +1713,6 @@ src/agents/embedded-agent-runner/model-context-tokens.ts	1
 src/agents/embedded-agent-runner/model.compat.ts	1
 src/agents/embedded-agent-runner/model.configured-fallback.ts	1
 src/agents/embedded-agent-runner/model.configured-overrides.ts	2
-src/agents/embedded-agent-runner/model.inline-provider.ts	2
 src/agents/embedded-agent-runner/model.provider-hooks.ts	10
 src/agents/embedded-agent-runner/model.registry-resolution.ts	6
 src/agents/embedded-agent-runner/model.ts	5
--- a/src/agents/embedded-agent-runner/model.configured-overrides.ts
+++ b/src/agents/embedded-agent-runner/model.configured-overrides.ts
@@ -117,23 +117,32 @@ export function findInlineModelMatch(params: {
 }) {
   const inlineModels = params.preparedModels ?? buildInlineProviderModels(params.providers);
   const normalizedProvider = normalizeProviderId(params.provider);
-  const find = (normalizeModelId?: (modelId: string) => string) => {
-    const matchesModelId = (entry: { provider: string; id?: string }) =>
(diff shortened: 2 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b0d341e5d123bc5077ea1c3b55d44f487dac2bf6](https://github.com/openclaw/openclaw/pull/143822/commits/b0d341e5d123bc5077ea1c3b55d44f487dac2bf6)
- Landed commit: [341932a8ef40c4228742a158387110d6f394da11](https://github.com/openclaw/openclaw/commit/341932a8ef40c4228742a158387110d6f394da11)

PR #143831, opened 2026-09-10
fix: open chat desktops on their assigned machines

steipete:
## What Problem This Solves

Fixes an issue where opening Desktop from a chat could show the Gateway and unrelated workers instead of connecting to that session's machine. Desktop startup also waited for unrelated provider catalogs, some worker connections stalled during VNC authentication, and moving keyboard focus between visible split panes could disconnect the desktop.

## Why This Change Was Made

Chat Desktop now follows the session's assigned machine through a targeted environment lookup and shows one connection status or a retry action. Worker-supplied VNC passwords reach the connection client even when the response has no auth discriminator, and stream lifetime follows actual panel visibility. Standalone machine selection, explicit source links, permission checks, and session placement changes retain their existing ownership boundaries.

## User Impact

Users can open their chat's desktop without choosing between unrelated machines. Connections avoid unnecessary catalog loading, authenticate correctly with worker-provided credentials, and stay visible when focus moves to another split pane. A missing or stopped session machine never falls back to another desktop.

## Evidence

- Regression failures reproduced before the fix: dropped worker password, global inventory lookup, and disappearing desktop canvas after a split-focus change.
- 62 focused component tests and all 47 desktop/sidebar/pop-out browser scenarios passed on the final source tree (109 tests). The complete browser set was rerun on the published head.
- Hosted CI caught a missed fixture in the existing rail-navigation suite. Its gateway desktop response now uses the targeted status RPC; both light and dark navigation variants pass with their connection assertions unchanged.
- `node scripts/check-changed.mjs` passed for the implementation, including core/UI type checks, lint, i18n, and source guards. The subsequent fixture-only correction passed focused browser tests, lint, formatting, and independent review.
- Independent review found no actionable P0–P2 findings.
- Screenshots below use inspected synthetic mocked Gateway data. The before capture reproduces the original three-source picker; the after capture shows automatic connection progress with no machine choices.

The live Gateway has not been restarted or deployed as part of this PR.

### Before

Chat Desktop offers three unrelated sources

### After

Chat Desktop automatically connects to its assigned machine

### DIFF
--- a/docs/gateway/cloud-sessions.md
+++ b/docs/gateway/cloud-sessions.md
@@ -67,11 +67,11 @@ See [Cloud Workers](/gateway/cloud-workers) for profiles, requirements, dispatch
 
 ## Viewing the session desktop
 
-Open **Desktop** from a session to view its execution machine. Cloud sessions select their worker desktop; sessions on paired devices select that device. By default, the pop-out window keeps the session in its link, so both viewers follow placement changes and disconnect from the previous machine when the session moves or stops. A stopped cloud session does not switch either viewer to the Gateway desktop.
+Open **Desktop** from a session to connect directly to its execution machine. Cloud sessions select their worker desktop; sessions on paired devices select that device. The chat panel shows connection progress or a retry action instead of offering unrelated machines. The pop-out window keeps the session in its link, so both viewers follow placement changes and disconnect from the previous machine when the session moves or stops. A stopped cloud session does not switch either viewer to the Gateway desktop.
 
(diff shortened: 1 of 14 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3cd40a17b285792afe851a956ae2a7c13ca02b2e](https://github.com/openclaw/openclaw/pull/143831/commits/3cd40a17b285792afe851a956ae2a7c13ca02b2e)
- Landed commit: [e3729e2b498530e509ae6c2ecca99756a9a84d96](https://github.com/openclaw/openclaw/commit/e3729e2b498530e509ae6c2ecca99756a9a84d96)

PR #143837, opened 2026-09-10
refactor(models): compose prepared catalog sources before lookup

obviyus:
Related: #136257, #143664.

AI-assisted.

## What Problem This Solves

Model lookup needed consistent composition of static, generated, authored and current configuration sources while keeping request credentials separate from inventory.

## Why This Change Was Made

Compose these sources in the registry before applying defaults. Current declarations own request settings, and replace mode admits only current declarations. Move catalog-row normalization to its existing responsibility boundary to remove a credential-normalization import cycle.

## User Impact

Explicit routes, sparse metadata and default precedence are preserved. Removing a current key cannot revive an older file key; developer-interface callers without a replacing declaration keep authored-file settings. No operator option or schema change.

## Evidence

The import-cycle check passed on the base, failed before the correction and passed afterward. Four hundred sixty-two tests passed across 33 affected suites. Retained Gateway and recording-service checks covered refresh retention, empty withdrawal, source routes and request isolation; they retain their original build scope.

### DIFF
--- a/docs/concepts/models.md
+++ b/docs/concepts/models.md
@@ -396,8 +396,15 @@ Custom providers configured under `models.providers` are written into `models.js
 Generated plugin catalogs supply model inventory, not request credentials. Their
 cached API keys, authentication modes, and request headers do not authorize model
 requests. Use a current auth profile or authored request configuration instead.
-The session SDK preserves authored `models.json` keys and headers while merging
-generated model metadata below authored rows.
+Without a current configuration snapshot, the session SDK preserves authored
+`models.json` keys and headers while merging generated metadata below authored rows.
+With a current snapshot, the provider's current declaration owns request settings;
+keys and headers left only in an older file do not regain authority.
+
+Prepared catalogs compose static, generated, authored-file, and current configured
+rows before resolving models. Explicit model routes win over captured routes, which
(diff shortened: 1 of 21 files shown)


obviyus:
Thanks @obviyus!

**Proof:** 462 tests passed across 33 catalog, prepared-runtime, registry, provider-normalization, and PDF caller files. Retained Gateway catalog operations and published-SDK HTTP observations cover source composition, replace membership, current-key removal, and recovery at their recorded build identities. Exact-head review passed with zero findings, and [CI is green](https://github.com/openclaw/openclaw/actions/runs/34496614447).

- **Bug:** Prepared lookup assembled model sources inconsistently, and importing credential normalization into the registry introduced an import cycle.
- **Fix:** The registry composes admitted sources before lookup; catalog row normalization has one owner separate from credential resolution.

| Before | After |
| --- | --- |
| Static, generated, authored, and current rows could disagree at lookup boundaries. | Prepared views consume the registry composition with current request settings and replace membership. |
| The import-cycle check failed. | Both import-cycle checks pass without an allowlist exception. |

Merged as `f7a419aeaf720d655e47d177a5e0de2322d5be24`. No unrelated-CI exception was needed.

PR #143844, opened 2026-09-10
fix(update): report temporary state snapshot cleanup failures

fuller-stack-dev:
## What Problem This Solves

Fixes an issue where updates could accept state inspection or a rehearsal snapshot even when deletion of the temporary raw database copy failed. The worker returned successful JSON and exited zero while a source-sized private copy could remain on disk.

## Why This Change Was Made

The existing snapshot consumer now checks its cleanup result before returning success. A failure names the owned temporary directory and suggests checking storage and permissions. If reading also failed, that original diagnostic and error cause are preserved.

The existing process-exit retry is unchanged. This adds no sweeper, cleanup store, background work, retention policy, or deletion of other captures.

## User Impact

Update state inspection now reports cleanup failure instead of silently accepting it. Healthy cleanup behaves as before. A transient exit retry can still recover the copy, but it does not retroactively certify an inspection that already failed cleanup.

## Evidence

- Reproduced on canonical main `924381f9f8b3fa7a325b5c03aca4acd0565cd8ff` through both real worker modes. Deletion failure was injected only for each child's synthetic raw snapshot. Persistent failure produced exit 0 and successful JSON while the private copy remained after process exit.
- Added 10 table-driven worker controls covering healthy cleanup, transient exit-retry recovery, persistent failure, and preservation of a real registry-read error with and without cleanup failure.
- Fixed behavior: cleanup failures return exit 1 with empty stdout and the owned path. Read failures remain visible. Source bytes, directory entries, and unrelated files remain unchanged; parent test cleanup runs after child settlement.
- `node scripts/run-vitest.mjs src/infra/update-candidate-state.cleanup.test.ts src/infra/update-candidate-state.test.ts src/infra/sqlite-readonly-location.test.ts`: 92 passed, 1 platform skip.
- Full `node scripts/check-changed.mjs -- src/infra/update-candidate-state.ts src/infra/update-candidate-state.cleanup.test.ts` passed on committed head `01c3c6129c3587051004ba61889be2b0c4cb4b19`.
- Final 10-case worker regression and changed-file lint: passed. Independent P2 autoreview on the final control flow: scoped-clean, no actionable findings.

Production delta: +19/-3 in one existing consumer. No runtime flags, database schema, snapshot primitive, package cleanup, or export-path changes. Persistent filesystem failure can still leave the owned copy; this correction makes that failure visible and does not promise durable reclamation.

### Post-activation recovery consequence

Cleanup failure uses the existing inspection-error policy: post-activation inspection returns `rollback-state-unverified`, keeps the candidate installed, and refuses automatic package rollback. It does not assert that schemas changed. A later successful process-exit retry removes the copy but does not change the failed worker result.

A separate bounded composed probe on this exact head exercised the real source worker, `inspectActivatedUpdateState`, and `rollbackFailedUpdate`. Both transient and persistent failures were checked through the forwarded-block and direct-reinspection paths: all refused rollback, retained the candidate root, preserved source bytes/family entries, and never reached the package-transaction tripwire. Transient exit retries removed both owned copies; persistent failures were measured after child settlement and then removed with the disposable parent fixture. The fixture used no service and a private package-shaped source entry; this is not installed-package or service-recovery proof. Healthy inspection still passed. No production recovery policy or package-restoration code was changed.

### DIFF
--- a/src/infra/update-candidate-state.cleanup.test.ts
+++ b/src/infra/update-candidate-state.cleanup.test.ts
@@ -1,0 +1,160 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { pathToFileURL } from "node:url";
+import { afterEach, beforeEach, expect, it } from "vitest";
+import { runCommandBuffered } from "../process/exec.js";
+import { openNodeSqliteDatabase } from "./node-sqlite.js";
+import { runtimeProcessEntrypoints } from "./runtime-process-entrypoints.js";
+import { resolveRuntimeWorkerArgv, resolveRuntimeWorkerUrl } from "./runtime-worker-url.js";
+
+let root: string;
+beforeEach(async () => {
+  root = await fs.realpath(await fs.mkdtemp(path.join(os.tmpdir(), "candidate-cleanup-")));
+});
+afterEach(async () => {
+  await fs.rm(root, { recursive: true, force: true });
+});
+
+const cases = [
+  { cleanup: "healthy", readError: false },
+  { cleanup: "transient", readError: false },
+  { cleanup: "persistent", readError: false },
+  { cleanup: "healthy", readError: true },
+  { cleanup: "persistent", readError: true },
+] as const;
+
+it.each(
+  (["versions", "snapshot"] as const).flatMap((mode) =>
+    cases.map(({ cleanup, readError }) => ({ mode, cleanup, readError })),
+  ),
+)(
+  "$mode: $cleanup cleanup with readError=$readError",
+  async (scenario) => {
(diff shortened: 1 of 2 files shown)


fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [01c3c6129c3587051004ba61889be2b0c4cb4b19](https://github.com/openclaw/openclaw/pull/143844/commits/01c3c6129c3587051004ba61889be2b0c4cb4b19)
- Landed commit: [2d3dd28da4569ecc8066a418a55055a34b4c91b9](https://github.com/openclaw/openclaw/commit/2d3dd28da4569ecc8066a418a55055a34b4c91b9)

PR #143858, opened 2026-09-10
fix: preserve cloud machines across Gateway updates

steipete:
## What Problem This Solves

Fixes an issue where an already-running cloud session would return to provisioning and lose its machine after a Gateway update. Exact build mismatch recovery destroyed the old lease, and restart recovery also reclaimed surviving nodes after accepting pending workspace results. Workspace checkpoints did not preserve host-installed packages or the existing desktop.

## Why This Change Was Made

Refresh the worker bundle on the existing machine after stopping the old runtime and revoking its credential. Verify the new bundle and atomically update both environment and placement build receipts under the original ownership checks. Keep the environment epoch because it also identifies the retained node workspace; resumed turns receive fresh claims and credentials.

Recover interrupted node turns and pending/staged workspace results without reclaiming their surviving machine. Failed refreshes retain the lease for retry and report the installer cause; startup placement recovery preserves the typed pending-update outcome instead of destroying an SSH machine after a failed refresh. Stop, Move, incompatible provider transports, and actual provider loss retain their explicit cleanup paths. Node and SSH installers revalidate authority before dispatching installation work.

## User Impact

Ordinary Gateway updates retain cloud machines, their workspace files, host-installed packages, and desktop. Interrupted node-backed sessions continue through the existing restart-recovery flow without replaying an interrupted tool call automatically. Nodes must reconnect with a supported installer before their runtime can be refreshed.

## Evidence

- Four machine-reuse regression cases fail against the original provider lifecycle, which tears down instead of installing the current runtime.
- Store, placement, recovery, and runtime suites: 222 tests passed after splitting the coverage into focused files. Provider suites: 45 tests passed. The startup SSH retry follow-up passes 77 runtime/recovery tests and all 10 explicit-teardown lifecycle cases; its new regression failed first because startup called provider destroy on the original lease.
- Additional SSH-bootstrap, RPC, credential, and node-supervisor checks passed; coverage includes stale ownership, failed installation and retry, pending/staged results, and fresh-turn admission.
- Independent autoreview of the final diff found no actionable P0/P1 findings. Targeted lint passes.
- `pnpm build` and `node scripts/check-changed.mjs` passed on the complete machine-retention change. The SSH-startup follow-up also passes the complete changed-code checks, lint, and regression suites; [final-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34461320479). Production Gateway restart verification is outside this PR's local proof.
- Initial CI exposed an obsolete active-recovery teardown fixture, now corrected to explicitly request drain while preserving its failure assertions. The separate model-selection shared-database collision was fixed independently in [#143875](https://github.com/openclaw/openclaw/pull/143875). This branch includes that repair and the separately landed desktop lint fix [#143889](https://github.com/openclaw/openclaw/pull/143889); both cloud-worker commits remain patch-identical after rebasing onto the repaired main.


steipete:
Addressed the SSH startup-recovery finding in c6ea02e7bcd27ba82d28129360cfd4e5ca29b38d. Tunnel admission now reports a typed pending-runtime-refresh outcome, and startup placement recovery retains that machine for retry. The new regression runs the real environment service through startup placement recovery: it first reproduced destruction of the original SSH lease, then passed with the fix and verified successful retry on the same lease. The affected runtime/recovery suites pass 77 tests.

The CI teardown fixture now explicitly enters workspace-result drain before exercising provider failure; all 10 lifecycle cases pass with their assertions intact. No SQLite schema, DDL, or serialized fields were added: the existing receipt fields change together, the workspace epoch stays stable, the v17 schema guard passes, and store tests verify persistence across reopen. Final-head CI is running.

steipete:
Merged via squash.

- Prepared head SHA: [c6cf93c53e3c6a8baf67ae0cabb9778e77bc5b49](https://github.com/openclaw/openclaw/pull/143858/commits/c6cf93c53e3c6a8baf67ae0cabb9778e77bc5b49)
- Landed commit: [20fb81f2489dd56d7cfef86a4eabc5164403a20b](https://github.com/openclaw/openclaw/commit/20fb81f2489dd56d7cfef86a4eabc5164403a20b)

PR #143861, opened 2026-09-10
fix(plugins): skip repaired canonical owners

steipete:
## What Problem This Solves

Resolves a problem where a managed plugin can be processed again during a core update after missing-payload repair has already handled it under a new canonical ID.

## Why This Change Was Made

Carry the repair's existing owner-migration records into the following update pass's skip set. This keeps both the original and repaired owner identities accounted for while preserving normal updates for unrelated plugins.

## User Impact

An already repaired plugin is skipped consistently after its ID changes. Package reconciliation, retained configuration, and unrelated plugin updates keep their existing behavior.

## Evidence

- The regression fails on the original code because the repaired config contains the canonical owner but the following pass skips only the old ID; five control cases pass.
- All 28 focused owner, migration, duplicate, package-reconciliation, and channel-sync cases pass with the repair.
- All 29 canonical changed-file checks pass.
- Fresh independent Codex review found no actionable findings through P2.

The proof establishes repeated processing eligibility and the repaired handoff. It does not claim a measured performance improvement, a second actual reinstall, or a live operator update.

### DIFF
--- a/src/plugins/update-cohort.test.ts
+++ b/src/plugins/update-cohort.test.ts
@@ -244,10 +244,15 @@ describe("plugin release cohort package reconciliation", () => {
     },
   );
 
-  it("removes the legacy load path after a successful post-core owner migration", async () => {
+  it.each(["update", "repair"])("reconciles the %s owner migration", async (phase) => {
     const legacyRoot = "/plugins/qqbot-legacy";
     const canonicalRoot = "/plugins/openclaw-qqbot";
     const legacyRecords = {
+      unrelated: {
+        source: "npm",
+        spec: "@example/unrelated",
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [11003e603cd1037b84465a3f619af7be976d0317](https://github.com/openclaw/openclaw/pull/143861/commits/11003e603cd1037b84465a3f619af7be976d0317)
- Landed commit: [f3c230c208d9f48a102ab457c0669e01341d91d1](https://github.com/openclaw/openclaw/commit/f3c230c208d9f48a102ab457c0669e01341d91d1)

PR #143866, opened 2026-09-10
fix: keep background completion wakes alive after replies

steipete:
## What Problem This Solves

Fixes an issue where a completed chat could later show “Async work scope is closed” when a background completion woke it. A shared wake timer could carry a finished foreground turn's work scope into several sessions. Those event-driven turns were also recorded as generic heartbeat polls, making ordinary completion handling look like a scheduled heartbeat in an unrelated chat.

## Why This Change Was Made

Delayed event dispatch and skill review explicitly acquire a fresh async-work scope, with a separate admission reference retaining cleanup without delaying their logical result. Ordinary independent admissions keep their caller's resource owner, preserving server shutdown cancellation for setup and plugin reload. Already idle scopes finish cleanup immediately, so completed work is no longer counted as active. Wake producers retain their event provenance through compact transcript markers for exec completions, cron wakes, and session events.

## User Impact

Background completions can resume their originating chats after the foreground answer has finished. Scheduled heartbeats still use the agent's main session by default, existing explicit session overrides remain supported, and silent acknowledgments stay hidden.

## Evidence

- A real-timer regression reproduced the original closed-scope error in both the originating session and another queued session before the fix; it passes with the fix.
- The new producer-provenance regression produced 13 intended failures on the pre-fix source; existing routing assertions passed.
- Focused tests cover scope drainage, prompt-result timing, suspension/replacement, event routing and provenance, and silent transcript filtering.
- `pnpm build` passed, including the Control UI build.
- Live OpenAI/OpenClaw-runtime proof passed with a real isolated Gateway and QuickJS Code Mode. An external gate held a background process until the foreground final reply; its subsequent exec completion produced a successful provider turn in the originating session. A forced scheduler-owned heartbeat ran in main without adding another turn to that session.
- The final build and live test were repeated successfully after CI exposed and the patch fixed an extra idle-queue microtask; the original cron root-count assertion passes unchanged.
- Clean Linux Testbox validation passed all 36 Gateway setup and plugin lifecycle tests; the one-shot lease was stopped after command success.
- Deslop completed; independent review found no actionable P0–P2 findings.
- Final `node scripts/check-changed.mjs` passed all selected gates after the explicit detached-boundary correction.
- [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34462128274) passed. An unrelated desktop lint regression was separately fixed by #143889; a zero-job GitHub startup failure was recovered by reopening, and a conflict-free rebase brought in the UI fix without changing this PR's aggregate patch.

### DIFF
--- a/docs/gateway/heartbeat.md
+++ b/docs/gateway/heartbeat.md
@@ -26,6 +26,8 @@ Setting `heartbeat.every: "0m"` also disables only the recurring cadence. A targ
 
 Targeted event wakes retain the same per-agent rate limits when recurring cadence is disabled: a 30-second minimum between event turns, and a flood guard after five starts within 60 seconds. Deferred work resumes when its guard expires. Config reloads preserve this accounting without enrolling the agent in recurring or broadcast heartbeats.
 
+Transcript markers distinguish `[OpenClaw heartbeat poll]` from an exec completion, cron wake, or session event. Scheduled polls use the configured heartbeat session (the agent's main session by default); targeted completion events return to the session that owns the work. Event markers retain their source provenance without copying internal instructions into chat history. Silent acknowledgment pairs remain hidden.
+
 Troubleshooting: [Automations](/automation/cron-jobs#troubleshooting)
 
 ## Quick start (beginner)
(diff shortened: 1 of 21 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7733747474c4e23df818d8fa8bd9c2c266a404d5](https://github.com/openclaw/openclaw/pull/143866/commits/7733747474c4e23df818d8fa8bd9c2c266a404d5)
- Landed commit: [cdda5825f8bb50e4112f13a9405856cad87bfa79](https://github.com/openclaw/openclaw/commit/cdda5825f8bb50e4112f13a9405856cad87bfa79)

PR #143869, opened 2026-09-10
fix: preserve subagent history across parent thread changes

steipete:
## What Problem This Solves

Follow-up to #143747. Live testing on merged main found that delivery of a native subagent's completion can start a replacement parent thread. The history reader then checked the completed child against the replacement and rejected its transcript as unrelated.

## Why This Change Was Made

Capture the original parent thread, session lifecycle, and a hashed connection identity in the existing task detail when the native task is created. History continues to use that original source while rechecking current requester access, session lifecycle, native connection, and bounded child ancestry. Progress, completion, and recovery preserve the first recorded owner.

Compaction transfers within the same lifecycle remain readable. Resets, account/store changes, malformed ownership data, and changes during an awaited read are rejected. Existing tasks without a recorded owner keep the prior binding-based path. The Gateway also includes native task detail in its cursor and asynchronous-read identity fence.

## User Impact

Completed subagent transcripts remain available after later parent turns replace the native thread. The shared sidebar, generic Subagent labels, cancellation, and task delivery continue to use their existing paths.

## Evidence

- Live main reproduction: the child completed, but `tasks.history` returned `Unable to load this task's transcript`; bounded diagnostics identified the parent-lineage rejection after the completion announcement replaced the parent thread.
- The new parent-replacement and asynchronous-owner-change regressions failed on the original readers for the expected reasons.
- All 346 focused Gateway, native history, mirror, monitor, runtime execution, and registration-import tests passed. Coverage includes repeated compaction transfers, immutable recovery metadata, and session/account/store/lifecycle changes.
- Independent review found no actionable P0–P2 issue in the final repair.
- Full build, production/test type checks, scoped lint, and the remaining changed-scope architecture checks passed.
- Real-provider proof on `bef1cb07d92bd2150f29b77524134b1e9278fa0f` passed: managed history traversed 2 pages and native history 4 pages, with four stable message IDs each, exact paginated/full equality, and paired real tool calls/results. A real browser opened both task transcripts through `tasks.history` and expanded their tool output. The fixture uses an isolated Gateway and synthetic profile; unrelated live scenarios remained disabled.

- Reopening proof on the same head passed: the isolated Gateway was stopped and restarted over its existing SQLite state. The actual native task retained its original history metadata; a seeded legacy-shaped native record retained the absence of that metadata. Both real managed/native transcripts were identical before and after restart, passed paginated checks again, and rendered their real tool output in the browser after reopening. This is restart/record-compatibility proof, not an old-binary upgrade campaign.

- Final post-merge replay passed on main `5ae901326b85b094d58339587157900093e0b48d`, including real managed/native history, identical results after Gateway restart, legacy-shaped record preservation, and both browser transcript views. Temporary instrumentation was removed and the task checkout is clean.

### Live managed subagent

Live managed subagent transcript and process result

### Live native subagent

Live native subagent transcript and native command result

### DIFF
--- a/docs/plugins/sdk-agent-harness/sessions-and-results.md
+++ b/docs/plugins/sdk-agent-harness/sessions-and-results.md
@@ -78,6 +78,15 @@ It accepts a task ID, an optional opaque cursor, and a limit from 1 to 200
 bound connection, and call `assertCurrent()` after awaited work. The Gateway
 rechecks access before returning a page and caps the response at 4 MiB.
 
+Record immutable native history routing facts in task detail when creating the
+task. A later parent turn can replace its current native thread without changing
+the child's source. Preserve the original parent and connection identity across
+progress, completion, and recovery; never reconstruct them from a replacement
+binding. Preserve authorized compaction transfers within the same session
+lifecycle, while rejecting resets and account or connection changes.
+Runtime task detail participates in the Gateway's cursor and
+after-await identity checks.
+
 Keep `childSessionKey` absent for native children: it describes an OpenClaw
 session and also determines lifecycle ownership. Reading history must not adopt
 the child, create another transcript store, or change cancellation and recovery.
--- a/docs/web/control-ui/chat.md
+++ b/docs/web/control-ui/chat.md
@@ -315,8 +315,11 @@ Codex harness. Select a task to read its messages, thinking, and tool calls;
 select **Show earlier** to load older history. Task activity refreshes the view
 while the subagent runs. The generic fallback label is **Subagent**.
 
-The viewer reads history from the runtime that owns it. If that runtime or its
-parent binding is unavailable, the panel shows an error with a retry action.
+The viewer reads history from the runtime that owns it. New native subagent
+tasks retain their original history source when later turns replace the parent's
+native thread. Changing the parent session or account can make that history
+unavailable. If the runtime or its parent binding is unavailable, the panel shows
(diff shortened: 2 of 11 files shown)


steipete:
Completed the requested database-backed reopening proof on bef1cb07d92bd2150f29b77524134b1e9278fa0f. The isolated real Gateway was stopped and restarted over the existing SQLite state. The actual native task's original history locator remained byte-equivalent, and a seeded legacy-shaped native record still had no locator. Both real managed/native histories were identical before and after restart, passed paginated reads again, and displayed their actual tool calls/results in the real browser afterward.

The prior live parent-rotation failure is repaired. The PR body contains the focused test, build, and live evidence plus inspected synthetic screenshots. This validates current readers reopening compatible stored records; it does not claim an old-binary upgrade campaign. The existing legacy reader and recovery regressions continue to cover the unchanged fallback path. No actionable code finding remains in the independent review.

steipete:
Merged via squash.

- Prepared head SHA: [bef1cb07d92bd2150f29b77524134b1e9278fa0f](https://github.com/openclaw/openclaw/pull/143869/commits/bef1cb07d92bd2150f29b77524134b1e9278fa0f)
- Landed commit: [e53fab421a53038aa3294c06855bc8e9c393f830](https://github.com/openclaw/openclaw/commit/e53fab421a53038aa3294c06855bc8e9c393f830)

PR #143872, opened 2026-09-10
fix: preserve literal model catalog metadata rows

steipete:
Related: #130706, #142100.

## What Problem This Solves

Catalog metadata indexes use display keys that can collapse distinct provider/model pairs. For example, a literal model ID that repeats its provider prefix can receive metadata from its sibling. The same key also drops distinct rows when catalogs merge.

## Why This Change Was Made

Use a literal provider/model tuple for private catalog merge and configured-metadata indexes. Keep public display keys and alias matching unchanged. Remove the synthetic metadata copier: the existing metadata application path already owns those fields. The focused two-file change removes 20 production lines.

## User Impact

Distinct configured rows keep their own names, limits, capabilities, and compatibility metadata. Existing provider normalization and aliases continue to work.

## Evidence

Three intended baseline failures and two passing controls establish the defect. All 245 focused tests, the full changed-file gate, and independent P2 review pass.

A canonical `sourcePerformance` runtime build passed at `cdc96c1cca6fbd60ed511a8a41e3c4000567d3f6`. Eleven compiled catalog scenarios cover provider-prefix and slash-boundary collisions in both orders, case distinctions, sparse synthetic metadata, normalization, and exact-row aliases. The probe made no network requests and imported no source modules; source and built artifacts remained unchanged. This is synthetic compiled catalog-boundary proof, not external-provider or live-Gateway traffic. The runtime profile omits UI and SDK declaration lanes. Global public-key behavior is outside this repair.

The identical patch was refreshed onto main containing the independently landed fixture and UI lint repairs in #143875 and #143889. All 245 focused tests, the full changed gate, and the runtime build/proof were rerun successfully; the prior P2 review remains applicable because the patch and relevant context are unchanged. Earlier failed CI evidence remains preserved.

### DIFF
--- a/src/agents/model-selection-shared.ts
+++ b/src/agents/model-selection-shared.ts
@@ -257,14 +257,19 @@ function sanitizeModelWarningValue(value: string): string {
   return sanitizeForLog(stripped.slice(0, controlBoundary));
 }
 
+// Metadata follows literal rows, even when display keys collapse provider-prefixed IDs.
+function modelCatalogEntryKey(entry: Pick<ModelCatalogEntry, "provider" | "id">): string {
+  return JSON.stringify([entry.provider.trim(), entry.id.trim()]);
+}
+
 function mergeModelCatalogEntries(params: {
   primary: readonly ModelCatalogEntry[];
   secondary: readonly ModelCatalogEntry[];
 }): ModelCatalogEntry[] {
   const merged = [...params.primary];
-  const seen = new Set(merged.map((entry) => modelKey(entry.provider, entry.id)));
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [cdc96c1cca6fbd60ed511a8a41e3c4000567d3f6](https://github.com/openclaw/openclaw/pull/143872/commits/cdc96c1cca6fbd60ed511a8a41e3c4000567d3f6)
- Landed commit: [1028efe3acbb3ae0d89a132f9c59480448618669](https://github.com/openclaw/openclaw/commit/1028efe3acbb3ae0d89a132f9c59480448618669)

PR #143832, opened 2026-09-10
fix: canceled automations stop blocking session updates

steipete:
## What Problem This Solves

Fixes an issue where users could have session updates remain blocked after a current-session automation was canceled while waiting for an active conversation turn.

## Why This Change Was Made

The background result's wait for existing work now observes its abort signal. Cancellation exits through the existing lifecycle finalizer, releasing the mutation fence before media preparation or transcript persistence. Noncanceled results keep their current ordering. The broader noncanceled wait cycle remains open; this change repairs cancellation during that wait.

## User Impact

Canceled automation results no longer hold up pinning or other metadata updates while the current turn continues. A result canceled during this wait is not added to chat history.

## Evidence

- The regression failed before the fix with cancellation and a real `sessions.patch` handler both pending while the old work lease stayed active. Explicit fixture release settled every promise and left no transcript message.
- After the fix, the cancellation regression, seven existing background-result cases, and sixteen abort-helper cases pass (24 tests): `node scripts/run-vitest.mjs src/sessions/background-session-result.test.ts src/infra/abort-signal.test.ts`.
- `check:changed` passed for the changed source, regression, and documentation, including core/test typechecks, lint, dead-code and schema guards.
- [Blacksmith Testbox Gateway proof](https://github.com/openclaw/openclaw/actions/runs/34449903707): after cancellation, authenticated `sessions.patch` completed in 32 ms while the old work lease remained active; public `chat.history` stayed empty and no media preparation ran. All owned processes and pipes settled, and the Testbox completed successfully. One-shot Testbox cleanup canceled the backing Actions session; the Crabbox command itself exited 0 with a successful proof and completed lease.
- The Gateway fixture creates the old-work and background-result owners internally and drives the cancellation signal there. Metadata/history requests use the real authenticated Gateway; upstream scheduler cancellation propagation is outside this proof.
- Initial managed autoreview completed with no actionable P0 findings; the subsequent review through P2 also completed scoped-clean.
- Baseline CI issues were handled separately: #143875 isolates a handoff fixture, #143889 restores the desktop line-limit check, and #143908 restores the existing SQLite read timeout before schema inspection. The cancellation patch remains unchanged.
- A profiler case timed out in the second CI run. A bounded, instrumented single-file Linux run on Node 24.19.0 passed all 35 cases with all 26 recorded children closed; the failing case completed in 360 ms. That 32-CPU run did not reproduce the full CI load, so no profiler fix or root cause is claimed.
[...]

PR #143875, opened 2026-09-10
fix(ci): isolate legacy handoff fixtures from concurrent config tests

steipete:
Related: #143832

## What Problem This Solves

Resolves a problem where concurrent CI model-command tests fail with `existing managed handoff lease is incompatible` while a managed-handoff regression test is running. This maintainer-requested fixture repair is required to validate #143832 independently of an existing baseline defect. [The observed CI job](https://github.com/openclaw/openclaw/actions/runs/34450754996/job/102786053112) failed eight model-selection cases.

## Why This Change Was Made

The legacy-reclamation case briefly writes a version-1 lease into the default shared handoff database. Unrelated config writers scan that database for retained custody and reject the unsupported row. Redirect this case's coordinator into its existing temporary fixture before creating the store. The parent passes the resolved database path explicitly to its helper, preserving real cross-process reclamation, cancellation, and cleanup.

## User Impact

CI can run these tests concurrently without sharing the synthetic legacy lease. This changes only test isolation; production behavior and its fail-closed lease checks are unchanged.

## Evidence

- A deterministic Linux probe paused the existing case after writing its legacy row and invoked a separate process using the real config-write lock. Both baseline `5a7eb757` and cancellation candidate `b725cedb` failed with the same incompatibility error; cleanup completed.
- With the isolated coordinator, the identical overlap probe passed. The full retarget suite (28 tests) and model-selection suite (36 tests) then ran concurrently through the CI shard runner: **64 passed**.
- [Blacksmith Testbox execution](https://github.com/openclaw/openclaw/actions/runs/34454278404): command exit 0. Its backing Actions session may show canceled after normal lease cleanup; that session conclusion is distinct from the successful command result.
- Changed-file checks and independent P2 review passed with no actionable findings.

The deterministic probe exercises the initial config guard; the observed CI stack also reached the same retained-custody guard during lock release. This is boundary proof for test contamination, not evidence about production latency or the broader noncanceled wait cycle tracked separately from #143832.

### DIFF
--- a/src/infra/update-managed-service-handoff-retarget.test.ts
+++ b/src/infra/update-managed-service-handoff-retarget.test.ts
@@ -610,6 +610,11 @@ unix(
     const { createManagedHandoffLeaseStore, resolveManagedUpdateLeaseDatabasePath } =
       await import("./update-managed-service-handoff-lease.js");
     const { to } = fixture();
+    const tmpDirOwner = await import("./tmp-openclaw-dir.js");
+    // Keep the legacy fixture row away from concurrent config writers.
+    vi.spyOn(tmpDirOwner, "resolvePreferredOpenClawTmpDir").mockReturnValue(
+      path.join(path.dirname(to), "coordinator"),
+    );
     const store = createManagedHandoffLeaseStore();
     const reserved = store.acquire(to, "legacy-owner", { kind: "update" });
     expect(reserved.kind).toBe("acquired");


steipete:
Merged via squash.

- Prepared head SHA: [fbfad9a43f91a804bd8758533ed45744c3c38bee](https://github.com/openclaw/openclaw/pull/143875/commits/fbfad9a43f91a804bd8758533ed45744c3c38bee)
- Landed commit: [7207976bd2fc72631f1622f71ff4363b8e383b94](https://github.com/openclaw/openclaw/commit/7207976bd2fc72631f1622f71ff4363b8e383b94)

PR #143877, opened 2026-09-10
chore: cover update owner revocation before acknowledgment

fuller-stack-dev:
## What Problem This Solves

The update owner suite covers revocation before dispatch and during acknowledgment, but not while asynchronous update discovery is still running. A regression at that earlier boundary could acknowledge an update for a revoked chat owner even if a later check still prevents mutation.

## Why This Change Was Made

Add two cases to the existing owner suite, covering managed and direct execution. Discovery changes the configured external owner before returning. Both cases require a durable `owner_required` failure with no acknowledgment, updater, helper, restart or sentinel effect.

This tests the existing policy. It does not change internal/channel-less requester authority or make socket disconnect cancel admitted updates.

## User Impact

No behavior change. Regression coverage protects the refusal-before-acknowledgment contract for externally configured update owners.

## Evidence

- Existing focused suite before the change: 8/8 passing.
- Focused suite with the change: 10/10 passing.
- P0/P1/P2 scoped review: no actionable findings.
- The handler and temporary SQLite run ledger are real; discovery, lifecycle notification, update execution and handoff effects use the existing mocks. This is not an installed/native update proof.

### DIFF
--- a/src/gateway/server-methods/update-owner.test.ts
+++ b/src/gateway/server-methods/update-owner.test.ts
@@ -8,6 +8,7 @@ import type { GatewayRequestContext } from "./types.js";
 import {
   adoptUpdateCampaignMock,
   detectRespawnSupervisorMock,
+  initializeGatewayUpdateStatusMock,
   runGatewayUpdateMock,
   scheduleGatewaySigusr1RestartMock,
   sendGatewayLifecycleNoticeMock,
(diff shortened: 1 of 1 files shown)


fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [d26eb0d2646f51a6abe6fc8c5cd64b81228de46f](https://github.com/openclaw/openclaw/pull/143877/commits/d26eb0d2646f51a6abe6fc8c5cd64b81228de46f)
- Landed commit: [d303532fd5f68da7a8502863d77b65ad06046776](https://github.com/openclaw/openclaw/commit/d303532fd5f68da7a8502863d77b65ad06046776)

PR #143881, opened 2026-09-10
fix(slack): honor relay proxy settings without direct fallback

n1hility:
Slack relay mode dialed secure WebSockets directly even when the Gateway host configured an HTTP proxy. The WebSocket client supplies its own connection function, so Node's global proxy agent did not apply.

Pass the relay URL to the existing shared proxy agent. Preserve direct connections for matching bypass rules and local plaintext relays. Report proxy setup errors through the existing recovery path, keep terminal socket listeners during connecting aborts, and sanitize malformed proxy diagnostics at their shared producer so credentials cannot reach error output.

Validation uses the actual Gateway and bundled Slack plugin, a local TLS relay, an HTTP CONNECT proxy, and synthetic Slack events. The baseline connected directly with zero relay CONNECT requests. The repaired Gateway tunnels the authenticated connection, acknowledges two delivery IDs on the same socket, and retains one durable message record. This proves transport and durable admission; the fixture does not request a model reply.

Focused validation: 47 passing tests across relay, durable ingress, shared proxy, and provider HTTP behavior; ordinary syntax lint and the changed documentation check pass.

Gateway controls cover proxy precedence, bypass rules, local plaintext, rejected CONNECT without direct fallback, reconnect backoff, stalled-connect abort, handshake timeout, payload rejection, disabled compression, and relay identity clearing after channel stop. Independent runtime acceptance passed 15 executions, including socket closure and a full post-stop retry interval.

Additional relay-monitor tests cover Basic proxy authentication, rejected credentials without direct fallback, authenticated bypass, and HTTPS proxy certificate trust.

### DIFF
--- a/docs/channels/slack/transports.md
+++ b/docs/channels/slack/transports.md
@@ -58,6 +58,8 @@ Relay mode separates Slack ingress from the OpenClaw gateway. A trusted router o
 
 The relay URL must use `wss://` unless it targets localhost. Treat the bearer token and router route table as part of the Slack authorization boundary: routed events enter the normal Slack message handler as authorized activations. A router-provided `slack_identity` in the websocket `hello` frame can set the default outbound username and icon; an explicit identity supplied by the caller still wins. The relay connection reconnects with the same bounded backoff timing as Socket Mode and clears the router-provided identity whenever it disconnects.
 
+Relay WebSocket connections honor the Gateway host's proxy environment (`HTTPS_PROXY`, `HTTP_PROXY`, `ALL_PROXY`, lowercase variants, and `NO_PROXY`). A `wss://` relay uses an HTTP `CONNECT` tunnel when a proxy is configured. A matching `NO_PROXY` entry or a localhost `ws://` URL connects directly. Proxy errors are reported instead of silently falling back to a direct relay connection.
+
 ## Socket Mode transport tuning
 
 OpenClaw sets the Slack SDK client pong timeout to 15 seconds for Socket Mode. This is a fixed internal default and is not operator-configurable.
--- a/docs/channels/slack/transports.md
+++ b/docs/channels/slack/transports.md
@@ -56,7 +56,7 @@ Relay mode separates Slack ingress from the OpenClaw gateway. A trusted router o
 }
 ```
 
-The relay URL must use `wss://` unless it targets localhost. Treat the bearer token and router route table as part of the Slack authorization boundary: routed events enter the normal Slack message handler as authorized activations. A router-provided `slack_identity` in the websocket `hello` frame can set the default outbound username and icon; an explicit identity supplied by the caller still wins. The relay connection reconnects with the same bounded backoff timing as Socket Mode and clears the router-provided identity whenever it disconnects.
+The relay URL must use `wss://` unless it targets localhost. Treat the bearer token and router route table as part of the Slack authorization boundary: routed events enter the normal Slack message handler as authorized activations. A router-provided `slack_identity` in the websocket `hello` frame can set the default outbound username and icon; an explicit identity supplied by the caller still wins. The relay connection reconnects with the same bounded backoff timing as Socket Mode and clears the router-provided identity whenever it disconnects. The relay websocket honors the gateway host's standard proxy environment (`HTTPS_PROXY`, `HTTP_PROXY`, lowercase variants, `NO_PROXY`): a `wss://` relay URL is dialed through the configured proxy with an HTTP `CONNECT` tunnel, and a `NO_PROXY` match or a localhost `ws://` URL connects directly.
 
 ## Socket Mode transport tuning
 
--- a/extensions/slack/src/monitor/relay-source.test.ts
+++ b/extensions/slack/src/monitor/relay-source.test.ts
@@ -1,6 +1,9 @@
+import fs from "node:fs";
 import http from "node:http";
 import https from "node:https";
 import net, { type AddressInfo } from "node:net";
+import os from "node:os";
+import path from "node:path";
 import type { Duplex } from "node:stream";
 import tls from "node:tls";
 import { rawDataToString } from "openclaw/plugin-sdk/webhook-ingress";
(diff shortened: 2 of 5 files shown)


obviyus:
The accepted scope for this repair requires configured-proxy enforcement with no silent direct fallback. This confirms the compatibility choice raised in review: invalid, unreachable, or rejecting proxy settings must produce a connection failure. Operators can correct the proxy or use the existing `NO_PROXY` setting for an intentional direct connection.

The transport docs state this behavior. Gateway proof covers both proxy rejection and working direct/bypass paths. The follow-up commit only replaces the test's WebSocket byte decoder; it does not change that production behavior.

n1hility:
Added `cde218393` on top of your changes: test-only, additive, covering the cases the earlier review listed as not tested (authenticated proxy, HTTPS proxy, and NO_PROXY with credentials present). Details below; happy to drop or fold it if you would rather keep the file as is.

Pushed one test-only commit (`cde218393`, rebased onto `4f747cc45`): `test(slack): cover authenticated and https relay proxies`. It closes the "authenticated proxies" and "HTTPS proxies" items from the earlier not-tested list. No source or docs changes; the existing tests and fixtures are untouched.

## What it adds

A sibling fixture beside `createRelayProxyFixture`, `createRelayGatedProxyFixture`, with the same TLS relay side and a CONNECT proxy that can require Basic credentials (answers 407 otherwise) and can terminate TLS itself. It records, per CONNECT, whether the destination `Authorization` header reached the proxy (`authorizationPresent`) and the `Proxy-Authorization` outcome (`missing` / `valid` / `invalid`), plus raw TCP connection counts on the relay and the proxy and completed TLS handshakes on the proxy. The proxy uses a second, distinct self-signed loopback certificate that is never added to the default CA store, so proxy trust can only come from the managed-proxy CA file. Two small helpers (`expectRelayAck`, `startRelayMonitor`) drive the real `monitorSlackRelaySource` the way the existing monitor tests do. Credentials are the fixed fixture strings `relay-user` / `relay-pass`; all hosts are loopback, ports are ephemeral, servers and sockets are closed in each test's `finally`.

| Case | Environment | Expected route | CONNECT count | Proxy-Authorization observed | Ack observed | Result |
|---|---|---|---|---|---|---|
| Proxy requiring Basic, correct credentials | `HTTPS_PROXY=http://relay-user:relay-pass@127.0.0.1:P` | tunnel | 1 | valid; destination bearer absent on the proxy hop | `authenticated-delivery` acked over the tunnel | pass |
| Same proxy, wrong password | `HTTPS_PROXY=http://relay-user:wrong-pass@127.0.0.1:P` | none (407, reported, retried) | 2 (attempt 1 and the backoff retry) | invalid on both; destination bearer absent | none; 0 TCP connections to the relay | pass |
| Credentials present, `NO_PROXY` match | `HTTPS_PROXY=http://relay-user:relay-pass@127.0.0.1:P`, `NO_PROXY=127.0.0.1` | direct | 0 (0 TCP connections to the proxy) | n/a | `bypassed-delivery` acked on the direct connection | pass |
| HTTPS proxy, managed-proxy CA file | `HTTPS_PROXY=https://127.0.0.1:P`, `OPENCLAW_PROXY_ACTIVE=1`, `OPENCLAW_PROXY_CA_FILE=<proxy cert>` | TLS to the proxy, then tunnel | 1 (after 1 completed TLS handshake) | missing | `tls-proxied-delivery` acked over the tunnel | pass |
| HTTPS proxy, certificate not trusted | `HTTPS_PROXY=https://127.0.0.1:P` | none (TLS failure at the proxy hop, reported) | 0 (1 TCP connection, 0 completed handshakes) | n/a | none; 0 TCP connections to the relay | pass |

## Observed output

Per-case observations printed from the tests during one run (trimmed; these are the values the assertions check):

```
http proxy, Basic credentials
  connects=[{"target":"127.0.0.1:59945","authorizationPresent":false,"proxyAuthorization":"valid"}]
  relay_connections=1 upgrades=[{"via":"proxy","authorization":"Bearer relay-secret","url":"/gateway/ws?gateway_id=pash"}]
  ack={"type":"ack","delivery_id":"authenticated-delivery"} accepted=1
http proxy, wrong password
  connects=[{"target":"127.0.0.1:59949","authorizationPresent":false,"proxyAuthorization":"invalid"},
            {"target":"127.0.0.1:59949","authorizationPresent":false,"proxyAuthorization":"invalid"}]
  relay_connections=0 upgrades=[] accepted=0
  disconnects=[{"lifecycle":"recovering","lastError":"HTTP/1.1 407 Proxy Authentication Required; code: CONNECT_FAILED"},
               {"lifecycle":"recovering","lastError":"HTTP/1.1 407 Proxy Authentication Required; code: CONNECT_FAILED"}]
http proxy, Basic credentials + NO_PROXY=127.0.0.1
  proxy_connections=0 connects=[]
  upgrades=[{"via":"direct","authorization":"Bearer relay-secret","url":"/gateway/ws?gateway_id=pash"}]
  ack={"type":"ack","delivery_id":"bypassed-delivery"}
https proxy, managed CA file
  proxy_secure_connections=1 connects=[{"target":"127.0.0.1:59956","authorizationPresent":false,"proxyAuthorization":"missing"}]
  upgrades=[{"via":"proxy","authorization":"Bearer relay-secret","url":"/gateway/ws?gateway_id=pash"}]
  ack={"type":"ack","delivery_id":"tls-proxied-delivery"}
https proxy, CA not trusted
  proxy_connections=1 proxy_secure_connections=0 connects=[] relay_connections=0 upgrades=[]
  status={"connected":false,"lifecycle":"recovering","lastError":"self-signed certificate; if the root CA is installed locally, try running Node.js with --use-system-ca; code: DEPTH_ZERO_SELF_SIGNED_CERT"}
```

Results on this head: `extensions/slack/src/monitor/relay-source.test.ts` 16 passed (16; 5 new); `node scripts/run-vitest.mjs extensions/slack` 156 files, 2930 tests passed; `oxfmt --check` clean; `git diff --check` clean; `oxlint --type-aware` reports nothing in the added code. Local Node v26.7.0. The wrong-password case waits through one real reconnect backoff (about 2.3 s) and carries a 15 s test timeout.

## Owner support facts

Authenticated proxies: the env proxy URL keeps its userinfo end to end. `src/infra/net/node-proxy-agent.ts` `resolveEnvNodeProxyTarget` parses the `HTTPS_PROXY`/`HTTP_PROXY` value with `new URL` and `fixedProxyEnv` pins that `href` for proxyline; `@openclaw/proxyline` `src/node-http.ts` (`ProxylineConnectAgent`, `onConnected`) builds `Proxy-Authorization: Basic base64(decodeURIComponent(username):decodeURIComponent(password))` from the URL and sends it proactively on every `CONNECT`; there is no challenge-response, and a non-2xx status line becomes a `ProxylineError` with code `CONNECT_FAILED` and the status line as message, which is what the relay records as `lastError`. HTTPS proxies: `@openclaw/proxyline` `src/proxy-socket.ts` `connectToProxy` opens `tls.connect` to an `https://` proxy (default port 443, `ALPNProtocols: ["http/1.1"]`, SNI for non-IP hosts) and writes the `CONNECT` after `secureConnect`; the CA for that hop is `proxyConnect.ca ?? proxyTls.ca`, and the relay passes neither, so it comes only from `src/infra/net/node-proxy-agent.ts` `createEnvNodeProxyAgentForTarget`, which calls `resolveActiveManagedProxyTlsOptions({ proxyUrl })` (`src/infra/net/proxy/active-managed-proxy-tls.ts`). That returns a CA only when the proxy URL equals OpenClaw's active managed proxy, either registered in-process (`src/infra/net/proxy/active-proxy-state.ts` `registerActiveManagedProxyUrl` with `proxyTls`) or inherited through `OPENCLAW_PROXY_ACTIVE=1` plus `OPENCLAW_PROXY_CA_FILE` (`src/infra/net/proxy/proxy-tls.ts` `resolveManagedProxyCaFileForUrl`, which requires an `https:` proxy URL, and `loadManagedProxyTlsOptionsSync`). Any other `https://` proxy is verified against Node's default CA store (`NODE_EXTRA_CA_CERTS`, `--use-system-ca`), which is the last row above.

## Not tested

- A deployed Slack relay router and a real corporate proxy; every server here is on `127.0.0.1`.
- Proxy authentication schemes other than Basic (Digest, NTLM, Negotiate): the SDK agent only emits Basic and does not answer `Proxy-Authenticate` challenges.
- An `https://` proxy trusted through the process CA store rather than the managed-proxy CA file, and the in-process `registerActiveManagedProxyUrl` registration path; the tests exercise the env-inherited managed-proxy path.
- Proxy client certificates (`proxyConnect` `cert`/`key`); the relay does not pass `proxyConnect`.
- Node 24; the suites ran on Node 26.7.0 only, CI covers the Node 24 lane.

Observation, unchanged here: `monitorSlackRelaySource` still rejects with the backoff sleep's `AbortError` when aborted during a reconnect delay; the two failure-path tests assert that outcome the same way the existing invalid-proxy test does.

obviyus:
Thanks for the additional proxy coverage. I independently ran all 16 relay tests on Node 24.19.0; they pass. I retained the additions and adjusted cleanup so a failed stop assertion still closes the fixture, with the CA directory owned by the shared test cleanup helper.

The automated review is currently blocked by its input-safety scan: four unverified URI findings in the generated review prompt. Your report contains credential-bearing example URLs. Please redact the userinfo from those prose examples and refer readers to the synthetic fixture credentials in the tests. Keep the observed authentication results and other evidence intact. The scan must pass normally before the final review; its rules will remain unchanged.

obviyus:
Thanks @n1hility!

**Proof:** The actual Gateway and bundled Slack plugin connected through a local TLS relay and HTTP CONNECT proxy. Two delivery IDs received acknowledgements on the same socket and retained one durable message record. Independent runtime acceptance passed 15 executions; 47 focused tests passed. This covers transport and durable admission, without claiming a live Slack service or assistant reply.

- **Bug:** Secure relay connections bypassed the configured proxy. Proxy setup errors could also silently select a direct connection.
- **Fix:** Reuse the shared proxy owner, report connection failures, retain socket error handling during abort, and keep malformed proxy credentials out of diagnostics.

| Before | After |
| --- | --- |
| Relay connected directly despite `HTTPS_PROXY`. | Relay uses CONNECT; explicit bypass rules and local plaintext still connect directly. |
| Unusable proxy settings could permit direct fallback. | Failures enter the existing recovery path, with no direct relay attempt. |
| Connecting abort could emit an unhandled socket error. | Public channel stop closes pending sockets and prevents later connection attempts. |

Exact-head review and CI passed. The configured-proxy failure behavior is intentional and was [accepted in review](https://github.com/openclaw/openclaw/pull/143881#issuecomment-5622938108).

PR #143880, opened 2026-09-10
fix(channels): Slack task cards repeat the command text when a command finishes

n1hility:
## What Problem This Solves

Fixes an issue where users watching Slack native progress with command text enabled (`streaming.progress.commandText: raw`) would see a finished command's task card show its command twice, as "run tests · command run tests", the second copy prefixed with the word "command", and the plan header switch to "Exec — command run tests" at the same moment. It happened for every command a run executed. In channels that render the progress draft as text, the same event changed the finished line's detail from "run tests" to "command run tests".

## Why This Change Was Made

The command output event describes its command through the agent's item title (`command <meta>`), which the shared progress line builder uses as the output line's detail when command text is on. That is a second description of the command whose detail the progress line already shows from the tool start. The compositor's merge (`mergeProgressDraftLineUpdate` in `src/channels/streaming.ts`) kept the shown detail only when the output line had none, so an output line with its own copy replaced "run tests" with "command run tests". Slack's native task rows append `details` text per `task_update` instead of replacing it, so the Slack card joined the two.

Command output now keeps the detail the reader has been watching and carries the result in the status; an output line whose detail already equals the shown one is used as is. The shown detail is kept only when the incoming detail is a restatement of it: empty, the line's own status, the shown detail itself, or the shown detail behind a `command` prefix (whitespace collapsed, prefix case-insensitive). Any other incoming detail replaces the shown one as before, so a command-output line that ever carries real output or a different description is not discarded by this merge. The existing guard is unchanged: a line that already ended (a failed nested command, the Matrix recovery case in #89920) is still replaced whole so a recovered run does not keep the stale failure text, and an output line arriving for a line that showed no detail keeps its own. The change is in the shared compositor rather than the Slack plugin because the compositor owns what a command line says (it already kept the shown detail for the no-detail case), every text-rendering channel sees the same flip, and Slack's reconciler already treats a restated detail as no-op once the strings agree.

Alternative considered: teaching the Slack reconciler to hold a row's details once shown. Not chosen because it would leave the text channels showing the relabeled command and would need a per-row policy crossing the chunk boundary; fixing the producer removes the divergence for every channel.

## User Impact

A command's task card in Slack keeps the command text it showed at start and only changes state when the command finishes; the plan header no longer relabels the command. Text-rendering channels show the same command text before and after completion. Nothing is configurable and no defaults change.

## Evidence

[...]

PR #143879, opened 2026-09-10
fix(channels): Slack native progress shows two task cards for one exec call

n1hility:
## What Problem This Solves

Fixes an issue where users watching Slack native progress (`streaming.mode: progress` with the native transport) would see two task cards for every `exec` call: a first card that stayed at the tool's start title ("🛠️ Exec") and was marked complete early, and a second card that carried the command's real completion. The pair appeared for every command a run executed, so a turn with five commands showed ten cards. Nothing in the Slack plugin's settings changes it.

## Why This Change Was Made

The agent keys one exec call's items separately in its item stream: the tool item is `tool:<call>` and the command item is `command:<call>`. The shared progress compositor already merges the tool line, the command item and the command output into one progress line for the call through the command correlation key, but the merged line took the id of whichever event replaced it, so the line's id changed from `tool:<call>` to `command:<call>` while the call was running. Slack keys its native task rows on the line id, so the id change opened a second row and the reconciler completed the first as "dropped out".

A progress line now keeps the id it was created with: `mergeChannelProgressDraftLine` restores the existing line's id on a correlation-matched replacement (`keepProgressDraftLineId` in `src/channels/streaming.ts`). The replacement still carries the newest content and status, later events for the same call still find the line through the correlation key, and the ids the builders put on individual lines are unchanged, so callers that address lines by their public item id keep working. The one existing expectation that pinned the merged line to the newest id (`keeps public command progress ids while replacing by command correlation`) now pins it to the line's first id; the ids on the built lines it asserts before merging are untouched.

Alternative considered: making the builders emit `tool:<call>` for command items and command output. Not chosen because the item ids on built lines are the agent's public item identities (the existing lifecycle test guards them), and Matrix's recovered-command flow (#89920) relies on replacement by correlation regardless of id.

## User Impact

In Slack native progress, one exec call renders as one task card that moves from in progress to complete or error in place. Other channels render the merged line text and see no change. Nothing is configurable and no defaults change.

## Evidence

Failing before, passing after. The new tests were run against the unpatched `upstream/main` source (`9865df761`) and then with the change:

```
### BEFORE (upstream/main source, new tests) ###
 × channels  src/channels/streaming.lifecycle.test.ts > channel-streaming > keeps public command progress ids while replacing by command correlation
 × channels  src/channels/streaming.lifecycle.test.ts > channel-streaming > keeps one line id while a tool call's item families replace the line
[...]

PR #141061, opened 2026-09-07
docs(channels): split the Slack page by reader job

vincentkoc:
`docs/channels/slack.md` was 110,670 characters and 42 headings in one file, mixing setup how-to, deployment explanation, runtime reference, and troubleshooting. It becomes a 12,721-character index plus eleven pages, one per reader job. No child exceeds 17,638 characters or 10 headings.

## Children

| Page | Sections it holds | Size |
| --- | --- | --- |
| `channels/slack/setup` | Install, Quick setup, User identity, Token model | 17,638 |
| `channels/slack/transports` | Choosing a transport (incl. Relay mode), Socket Mode transport tuning | 7,051 |
| `channels/slack/enterprise-grid` | Enterprise Grid org-wide installs | 11,733 |
| `channels/slack/manifest-and-scopes` | Manifest and scope checklist, Additional manifest settings | 11,475 |
| `channels/slack/access-control` | Actions and gates, Access control and routing, Group DMs | 11,232 |
| `channels/slack/threads-and-sessions` | Threading, sessions, and reply tags, Agent View DMs | 6,864 |
| `channels/slack/messaging` | Ack reactions, Text streaming, Typing reaction fallback, Commands and slash behavior | 11,641 |
| `channels/slack/media` | Voice input, Media/chunking/delivery, Attachment media reference | 10,763 |
| `channels/slack/rich-messages` | Native charts, Native tables, Plugin-owned modal submissions, Native approvals | 10,614 |
| `channels/slack/events` | Events and operational behavior, Presence events | 6,152 |
| `channels/slack/troubleshooting` | Troubleshooting | 7,232 |

The index keeps `## Configuration reference` (a pointer plus its field accordion) and `## Related`, and gains "What each page covers" and "Where each section moved".

## Anchor strategy

`redirectSource()` throws on any source containing `[?#]`, so **no redirect can ever carry a fragment**, and a per-anchor *route* would not help an external link written as `/channels/slack#text-streaming`. Instead, every one of the **75** anchor IDs the single page published stays alive on the index:

- **3** are still published by the index itself (`configuration-reference`, `related`, `high-signal-slack-fields`) and are deliberately **not** stubbed, so no `duplicate authored/canonical ID` is raised.
- **72** are authored `<a id="..." />` stubs in the "Where each section moved" list, each linking to `/channels/slack/<child>#<anchor>`.

[...]

PR #140837, opened 2026-09-07
perf(slack): prepare structured replies once per delivery attempt

steipete:
Closes #140836

<details>
<summary>Additional instructions</summary>

Maintainer edits are enabled on this repository-owned branch.

</details>

## What Problem This Solves

One Slack structured delivery repeatedly compiles the same authored Markdown during duplicate checks, actual send and success marking. Preview projection repeats that work again to identify matching authored sections.

## Why This Change Was Made

Prepare authored chunks and projections once per attempt, and pass that preparation through the private delivery flow. Record the exact key computed before the send await after confirmed success. Preview/media delivery retain separate policies; overrides create their own preparation and no global/payload-identity cache is added.

## User Impact

Twelve actual dispatch-to-send synthetic wire/error cases are identical, preserving media captions, raw blocks, text overrides, thread keys, split/native-table output, failed-send retry and duplicate suppression. A roughly 6 KiB structured send plus duplicate attempt drops from 11 to 4 formatter calls and 70,213 to 25,532 formatter input bytes.

Three allocation-instrumented samples measured median 2,317→870 ms and 6.665→2.437 GB cumulative JavaScript allocation. These include collected objects under instrumentation and synthetic I/O; they are not production latency, retained heap or RSS claims. The remaining render-aware prefix search is unchanged.

## Evidence

- 272 focused tests, complete changed-file checks and fresh isolated P 2 review passed on all seven final files; existing wire goldens and the original trace fixture are unchanged.
- Actual dispatchPreparedSlackMessage through real projections/delivery/send reaches a recording WebClient. Only core turn/network boundaries and the media receipt control are synthetic; no live Slack messages were sent.
- Preview-before-media-delivery preserves the caption and chart/post order. Raw matching sections keep cloned verbatim text objects; retries mark only successful keys, and resets remain intact.
- Production grows 56 lines for the attempt owner and two policy-specific resolutions; tests net+5. The only ratchet change shrinks dispatch-streaming.ts from 6 to 5. No SDK/configuration/persistence change or full build requirement.

Related #133135/#133136 cover distinct visible output behavior and remain separate. Preserve raw-upload #140787's independent Blob path/ratchet row. Active #133136 rendering-policy changes and #119737 confirmed upload receipts require fresh integration proof if they land first.

PR #143882, opened 2026-09-10
feat(slack): render streamed reasoning as task cards, with continuation messages for long streams

n1hility:
## Decision requested

This PR adds one optional setting, `channels.slack.streaming.progress.reasoning: "narration" | "cards"` (channel and per-account scope), and one additive compositor option (`reasoningLine`) used only by the Slack plugin. Default is unchanged (`narration`, today's behavior). `cards` is opt-in, there is no migration, and existing configs are untouched. It needs a maintainer's yes or no on the setting; the rendering code is only reachable once it is set.

Rebased onto current main, which already contains the first companion fix (#143879, merged); the second (#143880) is being landed by a maintainer. Two commits for review: the first renders reasoning as task cards on a single message; the second adds continuation messages, the byte ledger, and recovery when Slack rejects an oversized append. They ship together: the rendering half is not safe on its own for long turns, which is why an attempted split (#144256) was folded back.

## What Problem This Solves

With `agents.defaults.reasoningDefault: "stream"` (an existing operator opt-in), OpenClaw streams the model's reasoning while the reply is generated. On Slack's native progress card (`streaming.mode: "progress"`, native task cards, which is the default) the plugin keeps one compacted reasoning line, filters it out of the task list (`resolveNativeProgressLines`) and appends it to the stream's narration text (`resolveNativeProgressNarration`). The first sentence of the reasoning therefore lands at the top of the reply as plain prose, later reasoning is appended to that prose as further snapshots arrive, and a long think shows as an unlabeled paragraph above the card rather than as the thought it is. There is no setting that renders reasoning in the card itself.

## Why This Change Was Made

This adds one opt-in, `channels.slack.streaming.progress.reasoning: "narration" | "cards"` (default `narration`, today's behavior unchanged). In `cards` mode the Slack plugin renders streamed reasoning as task rows in the plan block:

- The plugin accumulates the reasoning text through the compositor's existing merge (`mergeReasoningProgress`, core semantics: an explicit or detected snapshot or a prefix extension replaces, otherwise the delta is appended) and splits it into segments of about 240 characters at word boundaries. Cuts depend only on preceding text, so a segment already shown never changes as the text grows. Each segment is one task row with a stable id (`reasoning:<n>`), label `Reasoning`, a 🧠 prefix, and `status: completed` for every segment but the newest. Row titles are capped at 250 characters; Slack documents 256 characters per `task_update` field ([chat.appendStream](https://docs.slack.dev/reference/methods/chat.appendStream/)).
- One streamed message holds only so much. Slack's plan block renders at most 50 tasks ([plan block](https://docs.slack.dev/reference/block-kit/blocks/plan-block/)), and `chat.appendStream` rejects further content with `msg_too_long` once a message that carries text holds about 10,600-11,300 UTF-8 bytes of titles, `details` (about double) and text (undocumented; measured, see Evidence). A long think therefore continues in a new message: the plugin keeps a per-message ledger of everything it has put on the current stream message (task titles, `details` and `output` weighted 2x, the plan title, streamed text, all in UTF-8 bytes; `extensions/slack/src/stream-size.ts`) and, before a send, splits the update into what still fits and what does not. New rows are admitted while the message stays under 6,000 weighted bytes of rows and 50 rows; streamed text while the total stays under 9,000. Segments are cut in the same unit (at most 240 bytes, never splitting a code point), so a card with its 5-byte prefix is at most 245 bytes and an emoji- or CJK-heavy segment can no longer be truncated by the title cap. When something does not fit, the current message is finished with `chat.stopStream` carrying the plan title `Thinking, continued below` and every open row marked complete, and a new native stream starts in the same thread for the same recipient with the title `Thinking, continued`, a fresh reconcile snapshot and the deferred rows; no task id carries over. Card numbering continues (`reasoning:<n>`), later tool rows land on the newest message, and the answer plus the `Thought for …` title land on the message that is current when the turn ends. Before the answer is appended the same check runs for closeout plus answer; if they would not fit, the think closes on its message with `Thought for …` and the answer streams as its own message in the thread. The reasoning runtime seals the cards that landed on a finished message at their current cut (`rolloverSlackReasoningCards`) while the compositor keeps merging the same phase, so cumulative snapshots neither repeat nor lose text.
- `msg_too_long` is now classified (`SlackStreamMessageTooLongError`, a `SlackStreamNotDeliveredError`): the stream is healthy, nothing from the rejected call landed, and the SDK's retained text is discarded before the message is stopped. In `cards` mode a rejected card update or streamed text rolls to a continuation and is retried once; a rejected answer finishes the think and retries the answer as its own stream; if the retry is rejected as well, the stream is stopped and the answer goes through the normal reply path. Every mode gains the last part: a `msg_too_long` on the answer no longer leaves the message `in_progress` without it.
- A tool call (`onToolStart`, phase `start`) and the end of a reasoning phase (`onReasoningEnd`) seal the open segment, so thinking after a tool result starts a new row below the tool row instead of extending the one above it. The seal runs before the tool row is admitted, which fixes the row order in the same append.
- Reasoning stays out of the narration text in this mode; commentary still streams as text.
- At completion the plan title becomes `Thought for <n>s, <k> tool call(s)` (part of the same mode rather than a second setting; easy to split if preferred). An explicit `progress.label` still wins, and while running the headline never borrows a reasoning row's text: it keeps the newest tool row's title or falls back to `Thinking`.
- Cards are ordinary compositor lines, so they share the existing one-second coalescing loop (`createDraftStreamLoop({ throttleMs: 1_000, coalesceInFlight: true })`), the delayed start gate and the native reconciler with tool rows: one `chat.appendStream` call per batch carrying every changed row, and no throttle of its own. Because each card occupies one line of the compositor's rolling window, `cards` mode raises the effective window to at least 64 lines (`progress.maxLines` raises it further); with the default of 8, a burst of more than eight new segments between two sends would evict a card before it was ever sent. Rows that leave the window later are already complete in Slack and stay there.
- Per-turn state (segments, the rolled prefix of the open phase, tool count, start time, the message chain and its ledger) resets with the rest of the progress turn state.
- `cards` applies to the detailed native card (`progress.toolProgress: true`); the quiet card keeps its single summary row and shows reasoning as narration, as today.

Layout: the pure segmentation/planning helpers live in `extensions/slack/src/progress-reasoning.ts`, the per-turn runtime in `extensions/slack/src/monitor/message-handler/dispatch-progress-reasoning.ts`, the per-message size ledger and fit planner in `extensions/slack/src/stream-size.ts`, the stream chain (snapshot, narration state, rollover, `msg_too_long` retry) in `dispatch-progress-stream.ts`, and `progress-blocks.ts` gains the 250-character title for reasoning rows plus the headline rule. `dispatch-streaming.ts` records streamed text in the ledger, accepts a thread override for continuations and a `msg_too_long` retry hook; the transport accepts a thread override. Core changes are limited to the config type, the UI-hint label and the regenerated bundled channel metadata. `narration` mode never enters the rollover paths (`rollover: false` skips every fit check and hook); the regression proof below shows its chunk sequence is byte-identical to `upstream/main`, and the dispatch tests assert one start, two appends and one stop for both the narration fixture and a short `cards` think.

Config surface: this adds one channel config key (`channels.slack.streaming.progress.reasoning`, also under `accounts.*`), so `docs/.generated/config-baseline.counts.json` `channel` moves from 3710 to 3712. Per AGENTS.md this is a conscious budget raise for an owner decision: the alternative of overloading `nativeTaskCards` or `style` would change the meaning of existing values.

Alternative considered: streaming the reasoning into the reply text instead of the card. Not chosen because a native stream is append-only, so every reply would carry its thinking permanently in the message body; task rows can be marked complete and collapse with the card, and the final answer stays the message text.

## User Impact

Operators who set `reasoningDefault: "stream"` can add `channels.slack.streaming.progress.reasoning: "cards"` (with `progress.toolProgress: true`) and see the model's reasoning as a sequence of 🧠 rows in the plan block, interleaved in order with tool rows, with the card titled `Thought for <n>s, <k> tool calls` when the turn finishes. A think longer than one message holds continues in a second (third, …) message in the same thread titled `Thinking, continued`, the previous one ending as `Thinking, continued below`; the answer follows the last one. In any mode, a `msg_too_long` on the answer now ends with the answer posted normally instead of a message stuck in streaming state. Everyone else sees no change: the default is `narration`, and with `reasoningDefault` off (the default) nothing is rendered either way. The setting validates as an enum; other values are rejected by the strict Slack schema.

### 1. Regression proof: `narration` (default) is byte-identical to `upstream/main`

Same probe test, same fixture (two cumulative reasoning snapshots, one `bash` tool call, a post-tool reasoning delta, `reasoning_end`, final reply), run through the real `dispatchPreparedSlackMessage` in the Slack dispatch test harness on a clean `upstream/main` (`6e9ee8af4`) worktree and on this branch with `progress.reasoning` unset. The probe dumps every `startSlackStream` / `appendSlackStream` / `stopSlackStream` call to JSON.

```
$ /tmp/fsrc-probe/run-probe.sh <upstream-main worktree> /tmp/fsrc-probe/out-main
$ /tmp/fsrc-probe/run-probe.sh <this branch worktree>   /tmp/fsrc-probe/out-branch
$ diff /tmp/fsrc-probe/out-main/narration.json /tmp/fsrc-probe/out-branch/narration.json
(empty diff)
```

On `upstream/main` the same probe with `progress.reasoning: "cards"` in the account config produces the same output as without it (the key is unknown there), so the "before" capture below is also what `cards` did before this change.

### 2. Before / after on the fixture

Before (`upstream/main`, and this branch in `narration` mode): the first reasoning snapshot becomes the stream's opening narration text, the rest of the reasoning is appended as prose, and no reasoning reaches the task list.

```json
"start": [{ "channel": "C123", "threadTs": "thread-1", "taskDisplayMode": "plan", "text": "Checking" }],
"append": [
  { "chunks": [
    { "type": "markdown_text", "text": " the Slack handler before the fix.\n\nTests pass; now summarize the change." },
    { "type": "plan_update", "title": "🛠️ Bash" },
    { "type": "task_update", "id": "tool_1_7998d275", "title": "🛠️ Bash", "status": "complete" } ] },
  { "text": "\nfinal answer", "chunks": [] } ]
```

After (this branch, `progress.reasoning: "cards"`): the reasoning is task rows, the pre-tool row is sealed complete before the tool row, the post-tool reasoning is a new row below it, no reasoning text is streamed as narration, and the completion retitles the plan.

```json
"start": [{ "channel": "C123", "threadTs": "thread-1", "taskDisplayMode": "plan", "chunks": [
  { "type": "plan_update", "title": "Thinking" },
  { "type": "task_update", "id": "reasoning_1_96e82800", "title": "🧠 Checking", "status": "in_progress" } ] }],
"append": [
  { "chunks": [
    { "type": "plan_update", "title": "Thought for 1s, 1 tool call" },
    { "type": "task_update", "id": "reasoning_1_96e82800", "title": "🧠 Checking the Slack handler before the fix.", "status": "complete" },
    { "type": "task_update", "id": "tool_1_7998d275", "title": "🛠️ Bash", "status": "complete" },
    { "type": "task_update", "id": "reasoning_2_f5df4246", "title": "🧠 Tests pass; now summarize the change.", "status": "complete" } ] },
  { "text": "\nfinal answer", "chunks": [] } ]
```

### 3. Live runs and the size probes (real Slack, 2026-09-09/10)

Harness: scripted agent events through the real `dispatchPreparedSlackMessage` with a real `@slack/web-api` client (8.1.1) against a test workspace (`streaming: { mode: "progress", nativeTransport: true, progress: { toolProgress: true, nativeTaskCards: true, reasoning: "cards" } }`; real compositor, start gate, coalescing loop and stream helpers; only the agent run is stubbed; every Web API call logged).

Motivating failure, cards without rollover (2026-09-09): scripted 12,010-character reasoning with one `bash` call, 45 cumulative snapshots. Slack accepted 1 `chat.startStream` + 29 `chat.appendStream` (76 `task_update` chunks over 48 ids: 46 cards, the tool row and the tail row of the earlier 47-row budget) and rendered all 48 rows with `Thought for 36s, 1 tool call`. The final `markdown_text` append (the 92-character answer) was rejected with `msg_too_long`; 11,215 UTF-8 bytes of task titles were on the message. The plugin did not classify the code, rethrew, never called `chat.stopStream`, and the message stayed `in_progress` with the reasoning but no answer. The same fixture cut to 6,949 characters (30 cards) completed normally.

Size probes (standalone script over the raw SDK, one call per second; artifacts `size-probe-{A,B,C,D,E}.calls.jsonl` / `.summary.json`, read-back `size-probe.replies.json`):

- A, 250-character ASCII titles, a 1-character text probe after each row: 42 rows accepted (10,571 bytes on the message), the 43rd rejected.
- B, 125-character titles: 56 row ids accepted with `ok: true`; Slack rendered 50 of them.
- C, 250-character titles with 100 characters of `details` each: 24 rows accepted (8,471 bytes), the 25th rejected; `details` render as rich text and cost about twice their length.
- D, 50 emoji-only rows of 250 UTF-16 units, no body text: all 50 accepted and rendered (22,900 bytes of titles).
- E, text-only stream: appends of 3,048 + 2,990 + 2,990 accepted (9,064 bytes on the message), the next 2,990 rejected (would be 12,054).

Rollover build live (2026-09-10; first on the code-point sizing build of this branch plus the two separately upstreamed core fixes, then again on the UTF-16 sizing build plus the same fixes, final set under seed `1789019616.230469`): the 12,010-character think rendered as three streamed messages in the thread, `Thinking, continued below` (26 rows: 25 cards + `🛠️ Bash`, all complete) → `Thinking, continued below` (24-25 rows) → `Thought for 36s, 1 tool call` (last cards + answer); 3 starts, 28 appends, 3 stops, 81 `task_update`s over 52 reasoning ids + `tool_1`, no errors, each rollover's stop/start pair about 300 ms, all messages `streaming_state: completed`. Plan titles in send order: `Thinking` → `🛠️ Bash` → `Thinking, continued below` → `Thinking, continued` → `Thinking, continued below` → `Thinking, continued` → `Thought for 36s, 1 tool call`. Short think, control (`narration`) and one-exec scenarios unchanged (1 start / 4, 0, 2 appends / 1 stop).

`msg_too_long` fallback live (scenario S5, emoji-only reasoning, both rollover builds): on the code-point ledger, 41 cards valued at 5,863 (10,209 UTF-16 units, 18,983 bytes) rejected the 3,048-character answer; on the UTF-16 ledger the cards rolled at 25/24/24 rows and the third card message, 24 rows the ledger valued at 5,606 units (10,418 bytes), still rejected the 3,045-character answer at a ledger total of 8,666, below the 9,064 characters a text-only stream had accepted. In both runs the plugin stopped the card message cleanly (`Thought for 30s` / `Thought for 31s`, all rows complete, `completed`), logged `finished at … weighted chars (msg_too_long); streaming the answer as a new message`, and streamed the answer as its own message in the thread. The card roll-and-retry path and the normal-reply fallback were exercised by the unit tests only.

Real-model run (2026-09-10, OpenClaw gateway from source, GLM 5.3 over Socket Mode, the UTF-8 sizing build of this branch plus the two core fixes; every stream call tapped): cards rendered correctly, and two defects surfaced that the scripted driver could not: a short think (9 rows, 483-character answer) closed the card with `Thought for 16s`, stopped it, and started a second stream for the answer; a long think that rolled correctly ended its continuation under the model's preamble headline instead of `Thought for …`. Root cause, one mechanism: core fires `onReasoningEnd` and `onAssistantMessageStart` best effort (`runBestEffortCallback`, not awaited). With a live model the `onReasoningEnd` handler is still waiting on the seal's in-flight append when the run ends and the final reply is dispatched; `deliverNativeFinal` cancels the pacing loop, which was the only guard on `beginNewProgressTurn`, and opens the compositor's `beginNewTurn` via `markFinalReplyStarted`, so the late boundary ran a turn rotation under the final: it stopped the stream (short turn: the answer then opened its own stream) or, in the long turn, hit the window between the roll's stop and the continuation's start, clearing the compositor and the reasoning runtime (no `Thought for`, headline kept). Fix: a non-forced `beginNewProgressTurn` now waits for the in-flight final delivery (`finalDelivery`) before it acts, so the answer and the summary title land on the card message and a genuine next assistant message still rotates afterwards. Two dispatch tests reproduce the live order (best-effort boundaries fired un-awaited while the seal append is in flight; a preamble headline showing before the answer).

Reading every run together (bytes of titles + details x2 + body text at the failing or last call): probe E 9,064 accepted / 12,054 rejected; probe A 10,571 accepted / 10,821 rejected; old S2 11,215 accepted titles-only / 11,279 rejected with the answer; S5 on the UTF-16 build 13,463 rejected (8,666 units); S5 on the code-point build 22,046 rejected; probe D 22,900 accepted with no body text. The check applies once the message carries body text and is consistent only in UTF-8 bytes, at roughly 10,600-11,300; code points and UTF-16 units each fit only part of the runs. Budgets in `extensions/slack/src/stream-size.ts`: 9,000 weighted bytes per streamed message (titles, plan title and text at 1x, `details` and `output` at 2x), rows roll at 6,000 so a 2,000-3,000 character answer still fits on the last message, 50 rows per message; about 19% under the ceiling for any script (Latin one byte per character, CJK three, emoji four; with bytes the S5 emoji messages would have rolled at about 13 rows and the answer would have fit), bounded by measurements rather than a documented limit, which is why `msg_too_long` still rolls and retries.

Deterministic rollover run (the dispatch test, fake timers, 50 numbered 240-character segments delivered as cumulative snapshots, `bash` after segment 20, one second advanced after each event): three stream messages. Message 1: `Thinking`, cards 1-20, `🛠️ Bash`, cards 21-24, stopped with `plan_update "Thinking, continued below"` and the tool row plus cards 21-24 `complete`. Message 2: `chat.startStream` in the same thread for the same `userId`/`teamId` with `plan_update "Thinking, continued"` and card 25, cards 26-48 appended, stopped the same way. Message 3: `Thinking, continued`, cards 49-50, `plan_update "Thought for 7s, 1 tool call"`, `"\nfinal answer"`, `chat.stopStream`. 24 cards of 245 bytes per message (5,880 plus the title, under 6,000); every segment appears once, in order, no id on two messages, no title over 250.

### 4. Source of the limits

- 50 tasks per plan block: https://docs.slack.dev/reference/block-kit/blocks/plan-block/ (hence the 50-row rollover bound; the API accepts more ids and renders 50, see the probe).
- 256 characters per `task_update` / `plan_update` text field: https://docs.slack.dev/reference/methods/chat.appendStream/ (hence 240-character segments and the 250-character title cap).
- Per-message size: not documented for `chat.appendStream`; `msg_too_long` measured 2026-09-09/10 against Slack as above: the check runs once the message carries text and counts titles, `details` (about 2x) and text in UTF-8 bytes, ceiling about 10,600-11,300 bytes (text-only 9,064 accepted / 12,054 rejected; 48 rows of 11,215 title bytes rejected a 92-byte answer; 24 emoji rows of 10,418 bytes rejected a 3,045-byte answer; 50 emoji rows of 22,900 bytes with no text accepted; with 100-character details 8,471 accepted / 8,821 rejected; the API accepts task ids beyond 50 but renders 50).
- Corroboration only: on a downstream build of this behavior against `@openclaw/slack@2026.8.1-beta.3`, a 62-second turn with one tool call rendered 30 cards in Slack (observed 2026-09-09).

### 5. Tests and checks

Test coverage added:

- `extensions/slack/src/stream-size.test.ts` (new, 10 tests): UTF-8 byte counting (emoji 4, CJK 3); titles and plan title replace while details, output and text accumulate, details/output at 2x; clone isolation; the measured budget constants; new rows admitted in order up to the row budget (24 cards of 242 under 6,000) and deferred past it; once one row is deferred all later rows are, so the plan keeps its order; updates to rows already on the message always apply; the 50-row cap; text that would pass 9,000 moves to the next message and counts before new rows; an empty message takes the text and a first row regardless.
- `extensions/slack/src/progress-reasoning.test.ts` (new, 13 tests): word-boundary segmentation within 240 UTF-8 bytes and round-trip; earlier segments byte-identical as the text grows through 40/80/160/240/300 words; hard cut of a spaceless run; whitespace collapse; byte sizing that never splits a code point (61 emoji → 240 + 4 bytes, 81 CJK characters → 240 + 3) and keeps emoji-heavy prose within a card; planner statuses (open vs sealed), post-tool text never in a pre-tool card; 60 segments number 1-60 with only the last open; rollover through card n keeps the same texts and cuts, later text continues as card n+1; cumulative snapshots after a rollover do not repeat sealed text; a partial card on the rolled message is sealed as is and the next card starts after it; rollover is a no-op when no open card was on the rolled message; a seal after a rollover closes the phase; summary title rounding and pluralization.
- `extensions/slack/src/streaming.test.ts` (+3): `msg_too_long` on an append throws `SlackStreamMessageTooLongError` (also a `SlackStreamNotDeliveredError`) with the pending text and leaves the session open; a chunk-only rejection the same with no pending text; `discardSlackStreamPendingText` clears the session and the SDK buffer so the following `stop()` sends no text.
- `extensions/slack/src/progress-blocks.test.ts` (+2): reasoning rows use their text as the title capped at 250 with an ellipsis (regular rows keep 120); the plan headline never borrows a reasoning row (`Thinking` fallback, newest tool row kept).
- `extensions/slack/src/monitor/message-handler/dispatch.preview-fallback.test.ts` (+12, through the real dispatcher and stream mocks): `reasoning: "narration"` reproduces the existing narration expectations with exactly one start, two appends and one stop; `cards` renders task rows and streams no reasoning text, with the same call counts; delta append / prefix-extension replace / flagged `<think>` snapshot replace merge into one row; sealing at a tool boundary yields `reasoning_1` complete, `🛠️ Bash`, `reasoning_2`, with the `Thought for …, 1 tool call` title; an explicit `label` survives completion; the quiet card (`toolProgress: false`) keeps narration even with `cards` set; the rollover run above (three messages, rolled stops with `Thinking, continued below` and all rows complete, continuations started in the same thread for the same recipient with `Thinking, continued`, numbering continuing, answer and `Thought for 7s, 1 tool call` on the last message, all titles ≤ 250); `msg_too_long` on a card append rolls and retries once (rejected append, rolled stop, continuation carrying the rows, turn finished there); `msg_too_long` on the append and again on the continuation start stops the stream and delivers the answer through `deliverReplies`; `msg_too_long` on the answer finishes the think and streams the answer as a text-only message in the same thread; `msg_too_long` on the answer and on its retry delivers it normally; a 24-card think with a 3,300-character answer closes with `Thought for …` and streams the answer separately; reasoning-end and message-start fired un-awaited during final delivery (real-model order) keep the answer on the card message with one start and one stop; a preamble headline showing mid-turn still yields the `Thought for …` completion title.
- `extensions/slack/src/config-schema.test.ts` (+1): accepts `cards`/`narration`, rejects other values at `streaming.progress.reasoning`.
- `src/config/schema.test.ts`: `reasoning` present under `channels.slack.streaming.progress` and absent for Discord; UI hint label `Slack Progress Reasoning Presentation`.

Results (Node 26.7.0, pnpm 12.3.4). The full lane set was run on the rollover tree; where the cards-only first commit gave different numbers they are in parentheses. The later sizing revisions (UTF-16 units, then UTF-8 bytes after the live S5 rerun refuted UTF-16) and the late-draft-boundary fix, all squashed into the second commit, re-ran the focused files, `oxfmt`, `oxlint`, both ratchets, `tsgo:extensions` and the Slack extension suite with the results shown:

- `node scripts/run-vitest.mjs extensions/slack/src/stream-size.test.ts extensions/slack/src/progress-reasoning.test.ts extensions/slack/src/progress-blocks.test.ts extensions/slack/src/streaming.test.ts extensions/slack/src/monitor/message-handler/dispatch.preview-fallback.test.ts`: 5 files, 311 passed (`stream-size` 10, `progress-reasoning` 13, `dispatch.preview-fallback` 197).
- Failing-before: the same `dispatch.preview-fallback.test.ts` run against the first commit's sources (the cards commit without rollover, streaming/progress files checked out underneath the new test file): `rolls a long think into continuation messages` (1 message instead of 3), `rolls to a continuation and retries once … msg_too_long` (no roll), `streams the answer as its own message when Slack rejects it` (sequence ends after the rejected append: `start, append, append, stop`), `finishes the think and streams a long answer separately` (1 message) fail; the two regression tests (`narration` and the short `cards` think, exact call counts) pass on both.
- `node scripts/run-vitest.mjs src/config/schema.test.ts src/config/types.slack.test.ts`: 2 files, 89 passed (first commit; the second touches no config).
- `node scripts/run-vitest.mjs src/config/ src/plugin-sdk/channel-config-ui-hints` (every core config test, first commit): 304 files, 4453 passed, 2 failed in `src/config/logging.test.ts` (ANSI color codes in the logged path under this terminal); the same 2 fail identically on the clean `upstream/main` worktree.
- `pnpm test:extension slack`: 155 files, 2929 passed, 0 failed (earlier revisions of the second commit: 155 files, 2927 passed, 0 failed; first commit: 154 files, 2903 passed, 5 environmental failures in `send.upload.test.ts` and `provider.transport-credentials.test.ts` that fail identically on clean `upstream/main` with this machine's network; they passed on the re-run).
- Re-run after rebasing onto `upstream/main` `d09401295` (2026-09-10, two commits): the six focused Slack files above plus `config-schema.test.ts` 363 passed; `src/config/schema.test.ts` + `src/config/types.slack.test.ts` 89 passed; `pnpm tsgo:extensions` exit 0; `pnpm config:docs:check` exit 0 (channel 3710 → 3712 against the current baseline); `pnpm test:extension slack` 158 files, 2964 tests passed. The rebase moved the doc additions from `docs/channels/slack.md` to `docs/channels/slack/messaging.md` (the page split in #141061) and kept `prepareSlackReply` from #140837 next to the ledger import in `dispatch-streaming.ts`. Three follow-ups from the first CI run of this PR, each folded into the commit that introduced the code: the segmentation and ledger helpers that only their tests imported (`segmentReasoningText`, `resolveOpenReasoningText`, the card and budget constants, `countSlackStreamChars`) are no longer exported and the tests exercise them through `planSlackReasoningCards`, `SlackStreamMessageLedger` and `planSlackStreamUpdateFit` (`pnpm deadcode:full` clean); the docs example no longer carries an `agents.defaults` stub, which the current schema rejects without `agents.entries` (`src/docs/channel-config-examples.test.ts` 3 passed); and the scripted checkpoint type in `dispatch.preview-fallback.test.ts` accepts the `Promise<VitestUtils>` that `vi.advanceTimersByTimeAsync` now returns (`pnpm tsgo:extensions:test` exit 0). After those: `progress-reasoning.test.ts` 13 and `stream-size.test.ts` 10 passed, `dispatch.preview-fallback.test.ts` 197 passed, `pnpm tsgo:extensions` exit 0, `pnpm test:extension slack` 158 files, 2964 tests passed.
- `pnpm check:changed --base upstream/main` (from a detached worktree under `/private/tmp`, lanes core, coreTests, extensions, extensionTests, docs, bundledChannelConfigMetadata): 16 lanes pass (conflict markers, max-lines and assertion SAFETY ratchets, changelog attributions, doctor deprecation registry, wildcard re-export guards, extension test core imports, duplicate scan coverage, dependency pin guard, format, doctor contract tests, bundled channel config metadata, config docs baseline, Plugin SDK package exports, Plugin SDK surface budget). The `plugin boundaries` lane stops the run with `eligibleForRemoval=1` (a date-driven compat record unrelated to this change; the identical command fails the same way on the clean `upstream/main` worktree). Lanes behind it were run directly from the same worktree: `pnpm tsgo:core`, `pnpm tsgo:core:test` and `pnpm tsgo:extensions` all exit 0, plus the Slack extension suite above.
- `oxfmt --check` on every changed file: clean. `oxlint -c .oxlintrc.json --tsconfig extensions/tsconfig.json` on every changed file: clean; `dispatch-progress.ts` is 643 lines after the stream chain moved to `dispatch-progress-stream.ts` (547 lines), both under the 700-line cap (the `scripts/run-oxlint.mjs` wrapper cannot run on this machine because module resolution escapes into an unrelated `~/node_modules` install, the same limitation noted on #139868; CI runs it).
- `pnpm tsgo:extensions`: exit 0.
- `pnpm config:channels:gen` and `pnpm config:docs:gen` regenerated `src/config/bundled-channel-config-metadata.generated.ts` and `docs/.generated/config-baseline.sha256`; `config-baseline.counts.json` channel 3710 → 3712 (first commit; see Config surface above).

### 6. Not tested / dependencies

Not tested:

- Live: the card roll-and-retry path after a `msg_too_long` on a card update, and the normal-reply fallback after a second rejection (unit tests only). The rollover chain itself and the answer-rejected path were exercised live (Evidence 3).
- The budgets rest on the measured points above rather than a documented limit; the `msg_too_long` roll-and-retry path covers the gap. The 2x weight for `details` comes from one probe (100-character details), and no live run has yet exercised the byte-based ledger on emoji or CJK reasoning (S5 ran on the code-point and UTF-16 builds).
- The Block Kit session-card fallback (`nativeTaskCards: false`) and `compact` style are untouched by design; `cards` has no effect there and this is documented, but there is no test that asserts it beyond the quiet-card case.
- Multi-message turns where a second reasoning phase starts after the final reply of the previous message: behavior matches `narration` (the compositor refuses lines after the final reply), asserted by reading the code, not by a test.

Dependencies on other branches: none. Pacing is already on `main` (#133516, the one-second coalescing loop), and this change routes cards through it. The progress-line id fix (#143879) and the command-detail fix (#143880), upstreamed separately, are independent of this branch; the only overlap is the plan-headline rule, which here is scoped to reasoning rows (`tasks.findLast((task) => !task.reasoning)`) and rebases cleanly either way.

Docs: `docs/channels/slack/messaging.md` (the streaming section of the Slack docs after the page split in #141061) gains a `streaming.progress.reasoning` bullet and a "Reasoning cards" section with the limits, the doc links, the `reasoningDefault: "stream"` prerequisite, the `toolProgress: true` requirement, the window note and a rollover paragraph (measured ceiling, budgets, what the user sees, the `msg_too_long` fallback). `CHANGELOG.md` is release-owned per AGENTS.md, so no entry.

### Live rendering (2026-09-10, this branch + the id and details fixes)

Scripted agent events were fed through the real `dispatchPreparedSlackMessage` with a real
`@slack/web-api` `WebClient` (8.1.1) against a test workspace; the harness keeps the real
compositor, start gate, one-second coalescing loop and the real
`startSlackStream`/`appendSlackStream`/`stopSlackStream`, and stubs only the agent run.
Every Web API request and response was logged (seed `1789019616.230469`).
No call in this run returned an error.

Short think, no tool call (two cumulative snapshots, 300 characters, `reasoning_end`,
answer; stream `1789021888.003149`): 1 `chat.startStream`, 4 `chat.appendStream`,
1 `chat.stopStream`. Slack renders one plan block titled `Thought for 5s` with two
completed 🧠 rows and the answer below it; no reasoning text in the body. The same input
with `progress.reasoning` unset (`1789021910.051989`) renders no plan block and the
compacted reasoning line (ending in "…") as the opening text of the message.

Long think with one tool call (scripted first-person prose, 12,010 characters, 45
cumulative snapshots growing 150–400 characters every 500–900 ms, `bash` at 40 %;
streams `1789021944.318419` → `1789021961.017469` → `1789021975.809329`): 3 starts,
28 appends, 3 stops over 36 s, 81 `task_update` chunks over 52 reasoning ids and one tool
id, longest title 242 code points. The think rolled twice on the byte budget
(`message 1 rolled (budget; 5711 weighted chars, 26 rows)`, `message 2 rolled (budget;
5713 weighted chars, 24 rows)`): message 1 ends as `Thinking, continued below` with 26
rows complete (25 cards and the `🛠️ Bash` row between cards 21 and 22, 5,870 bytes of
titles), message 2 as `Thinking, continued below` with 24 rows (5,796 bytes), message 3
carries `Thought for 35s, 1 tool call`, the last three cards and the answer. Each
stop/start pair took about 300 ms and the appends stayed on the one-second loop.

Non-Latin text (emoji-only reasoning, `1789021993.428589` … `1789022019.235869`): the
byte segmenter cuts cards at 240 bytes (73 code points), messages roll at 25 rows /
5,950 bytes, six messages in all, and the 3,045-character answer lands on the last one.
The same input on the code-point and UTF-16 builds had reached Slack's limit
(`msg_too_long` on the answer, answer streamed as its own message), which is what the
byte accounting fixes.

Why bytes: before rollover (2026-09-09, `1789013356.584549`) the 12,010-character think
rendered 48 rows and `Thought for 36s, 1 tool call`, then the 92-character answer was
rejected with `msg_too_long` and the message stayed in its streaming state. Probes
against the live API located the ceiling: one 250-character row per `chat.appendStream`
was rejected at the 43rd row; a text-only stream accepted 9,064 characters and rejected
the next 2,990; 50 emoji-only rows with no body text (22,900 bytes) were accepted; and
24 emoji rows of 10,418 bytes plus a 3,045-character answer were rejected although they
measured 8,666 UTF-16 units. The rejection threshold is consistent across all runs only
in UTF-8 bytes (about 10,600–11,300 of titles, details ×2 and body text, applied when
the message carries body text), hence the 6,000-byte row and 9,000-byte message budgets.
### Real model, real Slack (2026-09-10, gateway run from source on this branch + the id and details fixes)

OpenClaw gateway started from this tree on a workstation (`openclaw gateway run`), Slack in Socket Mode against a test workspace, model GLM 5.3 through an OpenAI-compatible endpoint with `reasoning_effort: high` and `agents.defaults.reasoningDefault: "stream"`, `streaming: { mode: "progress", nativeTransport: true, progress: { toolProgress: true, commandText: "raw", reasoning: "cards" } }`. A tap on the built plugin recorded every `chat.startStream` / `chat.appendStream` / `chat.stopStream` with its chunks; `conversations.replies` was read back afterwards.

- Short question: one stream, 3 appends, 1 stop; 1 reasoning row; plan title `Thought for 6s`; answer (475 characters) appended to the same message. Read-back: one bot reply with `plan` + `rich_text` blocks.
- Riddle: one stream, 8 appends, 1 stop; 6 reasoning rows; `Thought for 10s`; answer (553 characters) on the same message. Read-back: one reply, `plan` + `rich_text`.
- Task with a shell command (25 reasoning segments before and after one `exec`): first message 31 appends, 26 rows (25 reasoning + `Exec`, complete), rolled with `Thinking, continued below` when the next row would not fit the row budget; continuation 7 appends, 3 rows, `Thought for 59s, 1 tool call`, answer (995 characters) on it. Read-back: two replies, the first `plan` only, the second `plan` + `rich_text`. Elapsed 58 s, no Slack errors, no rate-limit answers.

The two earlier real-model turns, before the late-boundary fix that is now part of the second commit, showed the defects it addresses: the answer to a short question streamed as a separate text-only message after the card message was stopped (`Thought for 16s`, then a new `chat.startStream` carrying 483 characters), and a continuation message that ended under the model's preamble headline instead of the summary title. Both traced to OpenClaw firing the reasoning-end callback un-awaited while the final reply was being delivered; the fix makes a late turn boundary wait for the in-flight final. The stream tap for all four turns and the model-side event log are kept with the evidence.

### Screenshots

Captured from the Slack web client during the live runs above (a test workspace, cropped to one thread reply, plan block expanded). Full sets, with a manifest per run, are on the fork's [evidence branch](https://github.com/n1hility/openclaw/tree/f8b16d429ab4cea8a7d0a0cb44be1d773594ff71/screenshots).

Short think, no tool call (synthetic driver): two 🧠 rows under `Thought for 5s`, answer below the card.

Short think, no tool call (synthetic driver): two 🧠 rows under `Thought for 5s`, answer below the card

Long think with one tool call, first of three messages: `Thinking, continued below` with the cards and the `🛠️ Bash` row all complete (top of the message).

Long think with one tool call, first of three messages: `Thinking, continued below` with the cards and the `🛠️ Bash` row all complete (top of the message)

Same turn, third message: the last cards under `Thought for 35s, 1 tool call`, followed by the answer.

Same turn, third message: the last cards under `Thought for 35s, 1 tool call`, followed by the answer

Real model over Socket Mode, turn with one `exec` call: the continuation message ends with `Thought for 59s, 1 tool call` and carries the answer.

Real model over Socket Mode, turn with one `exec` call: the continuation message ends with `Thought for 59s, 1 tool call` and carries the answer

## Review findings: fixes and evidence

Each finding below came from review; each fix carries a test that fails on the reviewed head and passes after. Later rounds replaced per-site patches with a structural rule, bounded admission, enforced by a guard over the whole dispatcher suite; carried plan rows are now budgeted per continuation under the same rule.

## Review findings: fixes and evidence

Each finding has a regression test that fails on the previous head (`9af65ae5`) and passes on this revision. Test commands: `node scripts/run-vitest.mjs extensions/slack/src/monitor/message-handler/dispatch.preview-fallback.test.ts -t "<name>"` and `node scripts/run-vitest.mjs src/channels/progress-draft-compositor.test.ts`.

### 1. Rejected closeout content lost on rollover

Scenario: a budget rollover's `chat.stopStream` carries four newly admitted cards and comes back `msg_too_long`; the message is stopped without them, and the continuation opened with cards 25-30 only, so cards 21-24 never reached Slack.

Fix: `finishStream` returns `{ chunksRejected }`. `rollMessage` records as delivered only the rows that were already on the message when the closeout was rejected (`roll.previous`) instead of the reconciled closeout snapshot, seals the reasoning prefix through that card only, and the budget path commits the narration delta only when `closeoutDelivered` is true, so the admitted rows and the narration stay pending and the next round sends them on the continuation.

Test: `carries closeout rows Slack rejected with msg_too_long to the continuation` (`dispatch.preview-fallback.test.ts`).

Before:
```
× carries closeout rows Slack rejected with msg_too_long to the continuation
  → expected [ 25, 26, 27, 28, 29, 30 ] to deeply equal [ Array(10) ]
  -   21,
  -   22,
  -   23,
  -   24,
```

After: passes; the continuation's start carries cards 21-30, the rejected stop is followed by a stop without chunks, the answer lands on the continuation.

### 2. Unsent reasoning rows evicted beyond the compositor window

Scenario: one cumulative snapshot admits 70 cards before the paced loop's next send; with a 64-line rolling window the compositor evicted the earliest unsent cards, `pushedKeys` still counted them as pushed, and they never reached the stream.

Fix: cards mode no longer uses a rolling window. `withSlackReasoningCardsWindow` sets the compositor's line limit to `Number.MAX_SAFE_INTEGER`, so every admitted row stays in the compositor until the turn ends and each paced send reads the complete set; rows already delivered on earlier messages of the chain are filtered by `rolledTaskIds` as before. Message size stays bounded by the rollover budget (6,000 weighted bytes of rows, 50 rows), which is where the per-message limit belongs.

Test: `delivers every card of a burst wider than a rolling window, once and in order` (70 segments in one snapshot; asserts every card id 1-70 appears exactly once across the three messages, each card's title matches its segment, at most 24 cards per message).

Before:
```
× delivers every card of a burst wider than a rolling window, once and in order
  → expected [ Array(66) ] to deeply equal [ Array(70) ]
  -   3,
  -   4,
  -   5,
  -   6,
```

After: passes; 70 cards over 3 messages (24, 24, 22), none repeated.

### 3. Raw reasoning corrupted by restoring normalized text

Scenario: after admitting a card the runtime restored the normalized display text into the compositor's reasoning accumulator, so `"Reading "` + `"the handler"` rendered as `Readingthe handler`, and `"<think>Fresh"` + `" snapshot</think>"` lost the second delta entirely.

Fix: the compositor's `pushToolProgress` accepts a `reasoningLine` option; admitting a line flagged this way does not reset the reasoning accumulator (admitted work still does). The card runtime passes it for every card and no longer writes display text back, so the compositor keeps the raw phase text and only the rendered segments are normalized. No new SDK export; the option is additive on the existing compositor method.

Tests: `keeps the raw reasoning across deltas split at a trailing space` and `... at a wrapper tag closed by a later delta` (dispatcher), and `keeps the reasoning burst open when its own projection is admitted as a line` (`src/channels/progress-draft-compositor.test.ts`, also checks that a tool line still delimits the burst).

Before (dispatcher):
```
× keeps the raw reasoning across deltas split at a trailing space
  -   "title": "🧠 Reading the handler",
  +   "title": "🧠 Readingthe handler",
× keeps the raw reasoning across deltas split at a wrapper tag closed by a later delta
  -   "title": "🧠 Fresh snapshot",
  +   "title": "🧠 Fresh",
```
Before (compositor unit test, run with the compositor change stashed):
```
× keeps the reasoning burst open when its own projection is admitted as a line
  AssertionError: expected 'the handler' to be 'Reading the handler'
  Tests  1 failed | 42 skipped (43)
```

After: both dispatcher cases pass; compositor suite `Tests 43 passed (43)`.

### 4. Slack Stop lost across continuation messages

Scenario: a Slack Stop is admitted while the rollover's stop request is in flight; `finishStream` returned normally with `stoppedBySlack` set on the session, `rollMessage` forgot that session and opened a continuation, and the answer was posted on it.

Fix: the delivery runtime keeps a turn-level `stoppedBySlack` flag that `finishStream` sets from the session before the chain drops the session, and `isStoppedBySlack()` (flag or current session) now guards `deliverNormally`, `deliverWithStreaming`, `send`, `rollMessage`, `finishBeforeAnswer`, `prepareForAnswer` and the completion append. A Stop during a rollover therefore ends the turn with no continuation, completion or answer write; the flag is cleared when a new turn starts (queued follow-up).

Test: `does not open a continuation or post the answer after a Slack Stop lands during the rollover`.

Before:
```
× does not open a continuation or post the answer after a Slack Stop lands during the rollover
  → expected "vi.fn()" to be called 1 times, but got 2 times   (startSlackStream)
```

After: passes; one start, one stop, the last stream call is the stop, no answer text on any stream, no normal-reply fallback.

### Results

- Before (head `9af65ae5`), the five dispatcher cases: `Tests 5 failed | 197 skipped (202)`; compositor unit test: `1 failed | 42 skipped (43)`.
- After: the five dispatcher cases `5 passed | 197 skipped (202)`; `extensions/slack` whole suite `Test Files 158 passed, Tests 2969 passed`; `src/channels/progress-draft-compositor*` `Test Files 4 passed, Tests 103 passed`.
- `oxfmt --check`, `oxlint --type-aware` (extensions and core shards), `tsgo` for `tsconfig.extensions.json` and `tsconfig.core.json`, `git diff --check`, and the plugin SDK surface budget check all pass.

### Not tested

- Real Slack API behaviour for these fault paths: `msg_too_long` on `chat.stopStream` and a Stop racing a stop request are simulated through the mocked stream client, as in the other rejection tests.
- Narration text inside a rejected rollover closeout: the same `closeoutDelivered` outcome gates the narration commit, but the regression test covers rows only.
- A Stop admitted during `finishBeforeAnswer` (answer streamed as its own message): guarded by the same recheck, no dedicated test.
- A Stop followed by a queued follow-up in the same dispatch: the flag is cleared when the new turn starts, by design, not exercised by a test.
- Other channels passing `reasoningLine`: none do; the option is off by default and the existing compositor suites pass unchanged.

### 5. Rejected updates to already-visible reasoning rows

Scenario: cards 1-19 complete and card 20 partial are on the message; card 20 is extended and ten more cards arrive, the rollover closeout carrying the extension plus four admitted cards comes back `msg_too_long`, and the roll still sealed card 20 at its extended text, so the part Slack never took was missing from every continuation.

Fix: in a rejected closeout, `rollMessage` compares each previously delivered row's acknowledged title (`roll.previous`) with the title the closeout carried (`roll.full`); a row whose retitle Slack refused is not recorded in `rolledTaskIds` and does not advance the sealed card index. The reasoning runtime therefore seals only through card 19, and the next round sends card 20 whole, followed by the deferred cards, on the continuation; the rolled message keeps the shorter title Slack acknowledged. No re-segmentation is involved, so the compositor's rows and the runtime's plan stay identical.

Test: `carries the rejected extension of a partial card to the continuation` (`dispatch.preview-fallback.test.ts`; card 20 = 100 characters, extension adds 125, the rejected stop is asserted to carry the extension, the continuation must start with the full card 20 and hold cards 20-30, and the rolled message must never receive the extension outside the rejected stop).

Before:
```
× carries the rejected extension of a partial card to the continuation
  → expected [ 'reasoning_21', …(1) ] to deeply equal [ 'reasoning_20', …(1) ]
  -   "reasoning_20",
  -   "🧠 hhhh…(100) tttt…(125)",
  +   "reasoning_21",
  +   "🧠 s20xxxx…",
```

After: passes; continuation cards 20-30, card 20 title is the full extended text.

### 6. Rejected SDK text flushed by the rollover stop

Scenario: a paced update carrying commentary text plus new task rows is rejected with `msg_too_long`; the SDK keeps the text in its buffer, the rollover's smaller stop request flushes it onto the old message, and the retry sends the same text again on the continuation.

Fix: the `msg_too_long` branch in `send()` calls `discardSlackStreamPendingText` on the rejected session before `rollMessage`, as `rollForNarration` and `finishBeforeAnswer` already did for their retries. The local `text` stays pending and is sent once, with the continuation's opening update.

Test: `sends commentary rejected together with a card update once, on the continuation` (the mocked append stores the rejected text on the session like the real streamer, the stop mock records the session's pending text at the rollover stop; asserts it is empty, the rolled message's stop carries no `markdown_text`, and apart from the rejected append the commentary appears in exactly one call, the continuation's start).

Before:
```
× sends commentary rejected together with a card update once, on the continuation
  → expected '_Let me check the handler._' to be '' // Object.is equality
```

After: passes.

### Results (this revision)

- Before (head `c17c0a2f`), the two new cases: `Tests 2 failed | 202 skipped (204)`.
- After: the two new cases and the three earlier rollover cases together `5 passed | 199 skipped (204)`; `extensions/slack` whole suite `Test Files 158 passed, Tests 2971 passed`.
- `oxfmt --check`, `oxlint --type-aware` (extensions shard), `tsgo` for `tsconfig.extensions.json`, and `git diff --check` pass. Part 1 (`e3fa7957`) is unchanged.

### Not tested (this revision)

- Real Slack behaviour of a stopped message whose last row update was refused: the row keeps its acknowledged title and `in_progress` status on the finalized message; the continuation carries the complete card.
- A rejected closeout that retitles a non-reasoning row (tool `details` delta): handled by the same acknowledged-title rule, no dedicated test.
- The SDK buffer itself is simulated through the mocked session's `pendingText`; the real `ChatStreamer` buffer discard is covered by the existing `streaming.test.ts` case for `discardSlackStreamPendingText`.

### 7. Rejected same-title tool-row updates

Scenario: a tool row delivered as `Bash` (running) finishes while ten more cards arrive; the rollover closeout carrying its terminal status plus the admitted cards comes back `msg_too_long`. The previous guard kept a delivered row out of `rolledTaskIds` only when its title changed, so the same-title status change was recorded as delivered and the continuation and the final completion never showed the tool as finished.

Fix: in a rejected closeout, `rollMessage` now runs `reconcileSlackNativeTaskChunks` with the acknowledged rows (`roll.previous`) as the baseline and the intended update from the compositor (`roll.full`, restricted to already delivered ids) as the target; every row the reconciler would emit (title, status, `details` or `output` delta, sources) is treated as rejected, left out of `rolledTaskIds`, and re-sent whole on the continuation. The same field-level comparison the transport uses for deltas therefore decides what Slack still owes, instead of a title check. A consequence covered by the updated earlier test: a card whose only rejected change was its completion status is re-sent too.

Test: `carries a rejected same-title tool-row update to the continuation` (`dispatch.preview-fallback.test.ts`; `item` command event `running` then `command_output` end with the same `Bash` title; asserts the rejected stop carried `Bash complete`, the rolled message never received it outside that stop, the continuation's opening update carries `Bash complete` as its only tool update, cards 20-30 follow, and the turn finishes on the continuation).

Before:
```
× carries a rejected same-title tool-row update to the continuation
  → expected [ Array(11) ] to deep equally contain { type: 'task_update', …(3) }
  - Expected: { id: /^tool_call_1_…/, status: "complete", title: "Bash" }
  + Received: continuation start without the tool row
× carries closeout rows Slack rejected with msg_too_long to the continuation
  → expected [ Array(10) ] to deeply equal [ Array(11) ]   (card 20, completion rejected, not re-sent)
Tests  2 failed | 203 skipped (205)
```

After: `6 passed | 199 skipped (205)` for the six rollover cases; `extensions/slack` whole suite `Test Files 158 passed, Tests 2972 passed`.

Also in this revision: the scenario fixture type gains `commentary?: boolean`, which the `check-test-types` lane (`tsgo -p test/tsconfig/tsconfig.extensions.test.json`) flagged on the previous head; that project and `tsconfig.extensions.json` both typecheck clean, and `oxfmt --check`, `oxlint --type-aware`, `git diff --check` pass. Part 1 (`e3fa7957`) is unchanged.

Not tested: a rejected closeout whose only change is a `details`/`output` delta on a row (the same reconciler path decides it; the regression exercises the status field), and how Slack renders the stale `in_progress` row on the finalized message.

### 8. Queued cards bypassing the rollover owner at terminal paths

Scenario: in cards mode a 70-segment snapshot is admitted inside the throttle window and the turn then ends with an error final, a media final, or no final at all; the pending send was cancelled, `prepareForAnswer` was skipped for those finals, and the completion append (or the silent closeout's stop) carried every remaining row in one unbudgeted batch that Slack can reject or truncate to 50 rows.

Fix: a rollover-owned drain runs before each terminal path: inside the final-delivery chain before a final that does not stream (`drainNativeProgressBeforeCloseNow`, ahead of the normal reply and the completion append), and joined to the chain before the silent closeout in `dispatch.ts` and before turn rotation (`beginNewProgressTurn`, `onQueuedFollowupSettled`). The drain is `nativeStream.send()`, so the rows go out under the same byte and row budgets and roll into continuations as needed; it is skipped once the turn's final was delivered or the completion sent, so a final that streamed (which already drained through `prepareForAnswer`) is never followed by a second diff. The error status still lands on the message current at the end.

Test: `drains queued cards through budgeted messages before a $terminal closeout` (`it.each` over error, media, silent): asserts every card 1-70 reaches Slack once across three messages of at most 24 cards, the last message ends with a stop and carries card 70 with the turn's status (`error` for the error final), and the non-streamed final is delivered once through the normal path.

Before:
```
× drains queued cards through budgeted messages before a error closeout
  → expected [ [ { kind: 'start', …(1) }, …(2) ] ] to have a length of 3 but got 1
× drains queued cards through budgeted messages before a media closeout
  → expected [ [ { kind: 'start', …(1) }, …(2) ] ] to have a length of 3 but got 1
× drains queued cards through budgeted messages before a silent closeout
  → expected [ [ { kind: 'start', …(1) }, …(1) ] ] to have a length of 3 but got 1
```

After: all three pass (three messages, 24/24/22 cards, status preserved).

### 9. Completion chunks rejected before a separate answer

Scenario: a long answer needs `finishBeforeAnswer`; Slack rejects the think's completion (summary title, terminal rows, receipt) with `msg_too_long`; `finishStream` stopped without it, the caller set `completionSent` and forgot the stream, and the answer streamed as a bare text message.

Fix: `finishBeforeAnswer` reads the `finishStream` outcome. On `chunksRejected` it retires the finished message through the shared `retireMessage` bookkeeping (acknowledged rows retired, rows the rejected completion would have changed carried), then opens a plan-mode continuation whose first update is the completion rebuilt for a fresh message: the summary title, carried and turn-scoped rows, session receipt and diff output, plus any pending narration. Only then is `completionSent` set, and the answer streams into that message; if the continuation cannot be opened the caller falls back as before.

Test: `carries a rejected completion to the message that streams the separate answer`: the rejected stop carried `Thought for Ns`, the retry stop carried no chunks, and the reply message is `start` (plan mode, opening with `Thought for Ns`), `append` (the answer), `stop`, with the summary title appearing exactly once on it.

Before:
```
× carries a rejected completion to the message that streams the separate answer
  → expected [ 'start', 'stop' ] to deeply equal [ 'start', 'append', 'stop' ]
```

After: passes.

### 10. Running tool rows retired at rollover

Scenario: a tool is still running when a later update rolls the message; the closeout painted the row complete and its id was retired, so the tool's later failure and output disappeared from both running and terminal rendering.

Fix: `retireMessage` (shared by `rollMessage` and `finishBeforeAnswer`) distinguishes closing a message from finishing its tools: a non-reasoning row whose intended state is still `in_progress` when the message closes is painted complete on that message (it is finalized) but is not retired, so it continues on the next message and its real result lands there and in the turn's completion. The newest reasoning card is deliberately not carried: its cut is final and the rest of the text continues as the next card, which keeps card numbering unique across messages.

Test: `keeps a tool that is still running across a rollover until its result arrives`: the rolled message's stop paints the tool row complete; the continuation opens with the row `in_progress`, receives its `error`, and the completion settles it as `Recovered: …` complete; cards 25-30 follow it. The earlier rollover test was updated accordingly: a tool that never reports a result now follows the chain (and is the running headline on each continuation) instead of being frozen complete on the first rolled message.

Before:
```
× keeps a tool that is still running across a rollover until its result arrives
  → expected undefined to be 'in_progress' // Object.is equality
```

After: passes.

### Results (this revision)

- Before (head `7911e570`), the five new cases: `Tests 5 failed | 205 skipped (210)`.
- After: dispatcher suite `210 passed (210)`; `extensions/slack` whole suite `Test Files 158 passed, Tests 2977 passed`.
- `oxfmt --check`, `oxlint --type-aware`, `tsgo` for `tsconfig.extensions.json` and `test/tsconfig/tsconfig.extensions.test.json`, `git diff --check` pass. Part 1 (`e3fa7957`) is unchanged.

### Not tested (this revision)

- The oversized-text final (over the edit limit) takes the same non-streamed branch as the media final and is not exercised separately.
- Turn rotation drains (`beginNewProgressTurn`, `onQueuedFollowupSettled`) run the same helper as the silent closeout but have no dedicated regression.
- A rejected completion that also carries unfinished tool rows or a session receipt on the continuation; the regression covers the summary title.
- How Slack renders a row painted complete on a rolled message whose real result arrives on the continuation.

### Design: bounded admission

The reasoning runtime no longer hands every planned card to the compositor. It admits a card only while the message being built can still take it, rows and bytes, on top of everything already admitted for that message (the chain's `admitsRow` runs the production fit planner over the current draft plus the candidate against the current ledger, or a fresh one before the first start; tool and plan rows count the same way). Cards the message refuses queue in order inside the runtime and are released when the chain rolls to a new message. A card counts as queued only when the message refused it for room, not while a render is still waiting behind the send in flight, and a release is deferred while a compositor flush holds the gate's start promise on that send (releasing there would wait on itself). Nothing is evicted: the compositor's rolling window stays disabled in cards mode as a safety net, because the default window (eight lines) is smaller than one message's row budget; with admission bounded, at most one message's rows are unsent at any time.

Every message of the chain after the first is opened by one function (`openMessage`): it fits the opening under a fresh message's budget, text first (or room reserved for the text the caller appends next), then rows in order, defers the rest to the compositor and records only what Slack accepted. The send loop, the narration rollover and the rejected completion before a separate answer all open through it; `transport.start` has no other call site in the chain.

`send()` now reports `{ sent, settled }` (settled: nothing queued, every row of the draft on a message) and keeps going past a full message on its own, bounded by the work in hand (queued cards plus unsent rows, two rounds each) instead of a fixed round count. The terminal paths drain until settled: a final that does not stream, the silent closeout, turn rotation, and the queued cards before an in-stream answer (previously skipped when the answer fit). If the compositor refuses a release after the passes, the remainder is placed straight from the queue on budgeted messages, so no completion append or stop ever carries an unbudgeted batch; the turn's status (error/complete) still lands on the last message.

Guard: every dispatcher test charges each accepted start, append and stop to its message with the production `SlackStreamMessageLedger` and fails in `afterEach` if any message passes 50 rows or the 9,000-byte budget; the thresholds are read back from the exported fit planner at load time, so no production export exists for them (knip production scan: 0). It is on for all dispatcher cases and no scenario needed it off.

### 11. Narration continuation opened without a budget

Scenario: 60 reasoning rows queued inside one throttle window, then a text payload the current message cannot take (8,800 bytes), or one Slack rejects with `msg_too_long`; the narration rollover opened the continuation with every remaining row through `transport.start` and recorded them all as sent.

Fix: the narration rollover opens through `openMessage` with the narration reserved, so the continuation takes only the rows that fit beside that text (none, for the 8,800-byte case) and the rest follow on later messages through the paced sends and the pre-answer drain.

Test: `opens the narration continuation within budget and delivers the queued cards later` (`it.each`: by size, after msg_too_long). Asserts every card 1-60 arrives once in order, no message holds more than 24 cards, the narration is appended exactly once on the message opened for it, and the budget guard stays clean.

Before:
```
× opens the narration continuation within budget and delivers the queued cards later (by size)
  → expected 58 to be less than or equal to 24
  → "start: message holds 58 rows and 14229 weighted bytes (budget 50 rows, 9000 bytes)"
× ... (after msg_too_long)
  → expected 58 to be less than or equal to 24
```

After: both pass.

### 12. Terminal drain bounded by a fixed round count

Scenario: a 400-segment snapshot queued behind the pacing loop, then an immediate error or media final; the single `send()` in the terminal drain stopped after 16 rounds and the completion append carried the remaining rows in one batch (208 rows, 50,974 weighted bytes on one message).

Fix: with admission bounded the queue lives in the runtime; `send()` reports settled state and is bounded by the queued work, the terminal helpers loop until settled and place any remainder through `openMessage`, and the final status is applied on the last message.

Test: `drains a 400-card queue through budgeted messages before a $terminal closeout` (error, media). Asserts all 400 cards arrive once in order across at least 17 messages of at most 24 cards, the last message ends with a stop carrying card 400 in the turn's status (`error` for the error final), the final is delivered once through the normal path, and the guard stays clean.

Before:
```
× drains a 400-card queue through budgeted messages before a error closeout
  → expected 9 to be greater than or equal to 17
  → "append: message holds 208 rows and 50974 weighted bytes (budget 50 rows, 9000 bytes)"
× ... media closeout: same
```

After: both pass.

### Results (this revision)

- Before (head `d3bd14f5`), the four new cases: `Tests 4 failed | 210 skipped (214)`.
- After: dispatcher suite `214 passed (214)` with the budget guard on for every case; `extensions/slack` whole suite `Test Files 158 passed, Tests 2981 passed`.
- `oxfmt --check`, `oxlint --type-aware`, `tsgo` for `tsconfig.extensions.json` and `test/tsconfig/tsconfig.extensions.test.json`, `git diff --check` pass. Part 1 (`e3fa7957`) is unchanged; `src/channels` is untouched in this revision.
- Three earlier rejected-closeout tests were re-based: with bounded admission a budget roll's closeout no longer carries new rows or row updates (they are appended within budget first), so those scenarios now reject the append with `msg_too_long` and then the closeout; their assertions are otherwise unchanged.

### Not tested (this revision)

- The last-resort direct placement (`placePendingRemainder`) has no dedicated regression; with the final mark deferred until after the drain, no test reached it.
- Tool rows admitted by the compositor past the budget still go through the send loop's roll-with-admitted path; that path is exercised by the existing rollover tests only through reasoning cards.
- Real Slack behaviour of the reserved-text opening (title plus 8,800 bytes of text on one message) is inferred from the measured ceiling, not re-probed.

### 13. Deferred compositor rows not counted as unsettled

Scenario: more distinct tool rows inside one throttle window than a message can hold (120 in the test; 60 do not reach the case, because the first message and its roll absorb a budget's worth before any fresh opening has to defer), then an immediate error final. `openMessage` admitted only the rows that fit, but `send()` declared the result settled because only queued reasoning cards were counted, so `drain()` returned and the terminal completion appended the remaining rows unbudgeted.

Fix: `openMessage` reports whether the opening deferred rows, and `send()` settles a fresh opening only when no card is queued and nothing was deferred; the deferred rows go out on further messages through the existing roll path. A second defect surfaced by the regression: a message that overflowed with rows it could not take must not carry its running rows along (round 3's rule), or every continuation re-admits the same running rows ahead of the deferred ones and the chain never converges; such a message freezes its rows as before round 3, while a roll for queued cards, narration or the answer still carries running tool rows.

Test: `drains tool rows the compositor admitted past the budget before an error closeout` (120 distinct tool starts, error final): every row reaches Slack once across at least three messages, the last message ends with a stop carrying the error status, the final is delivered once, and the budget guard stays clean.

Before:
```
× drains tool rows the compositor admitted past the budget before an error closeout
  → expected 2 to be greater than or equal to 3
  → "append: message holds 120 rows and 1452 weighted bytes (budget 50 rows, 9000 bytes)"
```

After: passes.

### 14. Tool rows overtaking queued reasoning

Scenario: a 60-segment snapshot queues reasoning beyond the current message, and a tool starts before the next paced send; `seal()` returned with cards still queued and the tool row went straight into the compositor, so Slack showed the tool above the reasoning that preceded it.

Fix: `noteToolCall` flushes the queued cards through the chain (`drain`, joined to the transport order) before the tool row is admitted, so admission order is shared: reasoning that preceded the tool reaches its messages first. The chain's `releasePending` now waits for a pending compositor flush to settle unless the caller is the pacing loop's own send (which defers to the next paced send, avoiding the self-wait), so a tool-boundary drain queued behind the startup send never falls through to direct placement.

Test: `keeps a tool row behind the reasoning cards queued before it` (60 cards, tool start before the next send): the first appearance of every row in transport order is reasoning 1-60 then the tool, across the messages the cards need.

Before:
```
× keeps a tool row behind the reasoning cards queued before it
  → expected [ 'reasoning_1', 'reasoning_2', …(59) ] to deeply equal [ 'reasoning_1', 'reasoning_2', …(59) ]
    (tool_1 after reasoning_24, before reasoning_25)
```

After: passes.

### Design note (knip)

The guard no longer needs a production export. `SLACK_STREAM_MESSAGE_BUDGET` is private again; the dispatcher test derives the thresholds from the exported planner at load time (the row cap is how many tiny rows a fresh message admits, the byte budget the longest text a message already holding text still takes: 50 and 9,000). `scripts/check-deadcode-exports.mts` reports 0 entries for the production, script and full-tree scans.

### Results (this revision)

- Before (head `484b37b4`), the two new cases: `Tests 2 failed | 214 skipped (216)`.
- After: dispatcher `216 passed (216)` with the guard on; `extensions/slack` whole suite `Test Files 158 passed, Tests 2983 passed`.
- Knip unused-export scans (production, script, full-tree): 0 entries. `oxfmt --check`, `oxlint --type-aware`, `tsgo` for `tsconfig.extensions.json` and `test/tsconfig/tsconfig.extensions.test.json`, `git diff --check` pass. Part 1 (`e3fa7957`) unchanged; branch base `25a6148d1`.

### Not tested (this revision)

- A tool started while cards are queued and a compositor flush is pending (the startup send): the wait path in `releasePending` is exercised only indirectly.
- Running tool rows on an overflowing message whose result arrives later: by design they are frozen on that message; no test asserts the frozen state.

### 15. Later results for running rows frozen by an overflow roll

Scenario: a burst of tool starts larger than one message rolls the chain; a tool on a rolled message then fails. The overflow rule from the previous revision kept that running row out of `carried`, so its id entered `rolledTaskIds`, running and completion rendering both filtered it out, Slack kept the synthetic completed row, and the real result was never delivered.

Fix: running rows on an overflowed message are now frozen rather than plainly retired: `planSlackMessageRetirement` (extracted to `dispatch-progress-stream-rows.ts`) records each such row with its rendered state (title, status, details, output) alongside retiring it, so continuations still do not fill up with unchanged running rows. Before each render, `thawChangedRows` compares the frozen rows' current rendered state with the recorded one; a row whose state changed (a failure, output, a terminal status) leaves both the frozen register and `rolledTaskIds`, renders whole on the current message through the normal budgeted path, and is settled by the turn's completion there. Unchanged frozen rows stay frozen, so the non-convergence case remains fixed.

Test: `delivers a later result for a running row frozen by an overflow roll` (120 tool starts, paced sends roll the burst, then `command_output` with exit 1 for the first tool): the first message closed with the row painted complete, the current message receives the row with `error` and the completion settles it as `Recovered: …` complete, the answer lands there, no normal-reply fallback. The earlier non-convergence test (`drains tool rows the compositor admitted past the budget before an error closeout`) still passes.

Before:
```
× delivers a later result for a running row frozen by an overflow roll
  → expected [] to include 'error'
```

After: passes.

### Results (this revision)

- Before (head `c807766e`), the new case: `1 failed | 216 skipped (217)`.
- After: dispatcher `217 passed (217)` with the budget guard on; `extensions/slack` whole suite `Test Files 158 passed, Tests 2984 passed`.
- Knip unused-export scans (production, script, full-tree): 0 entries. `oxfmt --check`, `oxlint --type-aware` (including the 700-line cap on the stream chain, which motivated the extraction), `tsgo` for `tsconfig.extensions.json` and `test/tsconfig/tsconfig.extensions.test.json`, `git diff --check` pass. Part 1 (`e3fa7957`) unchanged; branch base `25a6148d1`.

### 16. Continuation capacity behind a full plan

Scenario: a plan update with about 50 short steps fills the native row budget by itself; once reasoning queues, every rollover kept all `plan_step_*` rows eligible and every continuation opened with those plan rows first, leaving no slot for a card, so the send and drain loops produced plan-only messages until their bounds and direct placement returned with the cards still queued.

Fix (in `dispatch-progress-stream-rows.ts`, wired from the stream chain):

- Every message of the chain after the first is opened through `planSlackMessageOpening`, which budgets plan rows: it fits the rows that must carry and the queued cards first (the cards the runtime will release next are reserved by fitting their chunks without sending them, and only the reserved cards that actually fit count, since the planner defers everything after the first row that does not), then gives the remaining slots to plan rows in priority order: running steps, then steps not started, then completed steps (already visible on an earlier message), and within each group steps no message has shown yet before steps shown before. The plan rows a message shows are fixed for that message (`messagePlanStepIds`); later sends update their statuses in place and do not add plan rows, so they cannot compete with cards. Plan rows left out are never counted as deferred work; they wait for the next continuation.
- Steps shown on any message are tracked (`shownPlanStepIds`, recorded when a message retires), so the remainder paginates forward across continuations while running steps stay visible on each.
- At the closeout, `admitPlanForCompletion` lets plan rows the current message left out onto it, in the same priority order, as far as they fit beside what the message holds and the answer to come (`prepareForAnswer` reserves the answer text; the non-streamed and silent closeouts reserve nothing). Completed steps are therefore elided on a message only when the plan does not fit beside the rows already on it (for the final message: beside the cards it carries and the answer). This is broader than "only when the plan alone exceeds the row budget": on the last message the plan competes with the cards that landed there, and Slack's 50-row render cap cannot be exceeded either way. The first message still shows the whole plan.
- Cards refused for room now schedule the paced send themselves (`noteQueued`); previously nothing reached the compositor in that case, so the roll waited for an unrelated later event.

Tests: `leaves continuation capacity for queued cards behind a plan that fills the row budget` (50 pending steps, 30 queued cards: every continuation carries cards, 2-4 messages, each card once, answer on the last message, guard clean) and `keeps plan progress visible across budgeted continuations and settles it at completion` (50 steps progressing from step 1 to step 2 while the cards go out: the running step opens each continuation, its completion and the next step's start land on the message current at that time, and the closeout settles the running step). The existing rollover and plan tests pass unchanged.

Before:
```
× leaves continuation capacity for queued cards behind a plan that fills the row budget
  → expected [] to deeply equal [ Array(30) ]
× keeps plan progress visible across budgeted continuations and settles it at completion
  → expected 'complete' to be 'in_progress'
```

After: both pass.

### Results (this revision)

- Before (head `7498b866`), the two new cases: `2 failed | 217 skipped (219)`.
- After: dispatcher `219 passed (219)` with the budget guard on; `extensions/slack` whole suite `Test Files 158 passed, Tests 2986 passed`.
- Knip unused-export scans (production, script, full-tree): 0 entries. `oxfmt --check`, `oxlint --type-aware` (the stream chain stays under the 700-line cap; opening, retirement, thaw and completion planning live in `dispatch-progress-stream-rows.ts`), `tsgo` for `tsconfig.extensions.json` and `test/tsconfig/tsconfig.extensions.test.json`, `git diff --check` pass. Part 1 (`e3fa7957`) unchanged; branch base `25a6148d1`.

### Not tested (this revision)

- A plan larger than one message (more than 50 steps) on the first message: the first message still shows the whole plan and relies on Slack rendering the first 50 rows; only continuations budget plan rows.
- Plan rows added by a later plan update while a continuation is open appear on the next continuation or the closeout, not on the open message; no dedicated test.


n1hility:
Patch updates for the 2 findings coming soon

n1hility:
@clawsweeper re-review

n1hility:
@clawsweeper re-review

n1hility:
Status for whoever picks this up: review findings are cleared (16 fixed across the rounds, each with a failing-before test; the last rounds made bounded admission a rule enforced by a guard over the whole dispatcher suite). The branch is rebased onto current main, which already contains the first companion fix (#143879); the second (#143880) is being landed separately. The one lane still red is a memory-core extension test unrelated to this change.

The only open item is the product decision at the top of the description: the optional `channels.slack.streaming.progress.reasoning` setting (`"narration"` default, unchanged behavior; `"cards"` opt-in) and the additive compositor `reasoningLine` option it uses. Happy to adjust the setting name, scope, or default, or to split differently, if that helps the decision.

PR #143558, opened 2026-09-10
fix(opencode-go): attach stable session routing per turn

RomneyDa:
## Summary

- attach `x-opencode-session` through provider turn state for native OpenCode Go endpoints
- route the fallback through both Chat Completions and Responses transports
- preserve explicit model and request session headers ahead of generated conversation/turn identity

## Release-validation evidence

Full Release Validation run 34418806238 failed all six OpenCode Go live probes with HTTP 400 because sessionless requests lacked `x-opencode-session`. The shared provider turn-state hook now reaches both affected transports while explicit configured identity retains precedence.

Failing job: https://github.com/openclaw/openclaw/actions/runs/34418865241/job/102691864264

## Test plan

- `provider-transport-session-headers.test.ts` (14 passed; real fetch-egress coverage for both transports and explicit header precedence)
- `extensions/opencode-go/index.test.ts` (39 passed)
- `pnpm check` (passed)

PR #143890, opened 2026-09-10
fix(ai): preserve provider turn headers with explicit sessions

vincentkoc:
Related: #143558

## What Problem This Solves

Fixes an issue where OpenCode-backed turns with an explicit session header could lose other provider turn-state headers, including routing metadata, when using managed completions, Responses HTTP or WebSocket, or the simple completion path.

## Why This Change Was Made

Provider turn-state resolution now removes only the conflicting `x-opencode-session` header when the caller supplies an explicit session. All non-conflicting provider headers remain attached, while caller-provided session precedence is unchanged across the affected transports.

## User Impact

Explicit OpenCode sessions retain provider routing and other turn-state metadata instead of silently dropping the entire provider header set.

## Evidence

- Added regression coverage for managed completions, simple completions, Responses HTTP, and Responses WebSocket.
- `oxfmt --check` passes for all five changed files.
- `git diff --check` passes.
- Independent autoreview of the complete corrective patch found no P0-P2 issues.
- Local Vitest and oxlint execution could not start because this task worktree uses a borrowed dependency symlink: the donor install is missing markdown dependencies and the declaration-boundary guard rejects compiler inputs outside the checkout. Exact-head hosted CI is the executable proof path.

### DIFF
--- a/packages/ai/src/transports/openai-completions-transport.ts
+++ b/packages/ai/src/transports/openai-completions-transport.ts
@@ -37,8 +37,11 @@ import {
   type MutableAssistantOutput,
   type OpenAIModeModel,
 } from "./openai-transport-shared.js";
-import { resolveProviderTransportTurnState } from "./provider-transport-turn-state.js";
-import { hasOpencodeSessionHeader, resolveOpencodeSessionHeaders } from "./session-affinity.js";
+import {
+  filterProviderTurnHeadersForExplicitOpencodeSession,
+  resolveProviderTransportTurnState,
+} from "./provider-transport-turn-state.js";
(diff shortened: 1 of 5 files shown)


vincentkoc:
Merged via squash.

- Prepared head SHA: [76e83e480248a8f6391869b453b0d76096e0239b](https://github.com/openclaw/openclaw/pull/143890/commits/76e83e480248a8f6391869b453b0d76096e0239b)
- Landed commit: [1e81ab7f143d806e7ed99efafed831397636e982](https://github.com/openclaw/openclaw/commit/1e81ab7f143d806e7ed99efafed831397636e982)

PR #143552, opened 2026-09-10
refactor(config): derive install record types from schemas

RomneyDa:
## Summary

- derive generic and plugin install-record types from their canonical Zod schemas
- keep internal schema shapes private
- preserve the existing passthrough persisted-record codec

## Why

Install source, trust, artifact, marketplace, and accepted-capability fields were declared independently in runtime schemas and TypeScript contracts. This makes the validated persisted schema the single owner and removes 57 production lines.

## Compatibility and upgrade proof

Persisted field keys, validators, defaults, normalization, serialization, and exported type names are unchanged. The existing `.passthrough()` parser preserves unknown fields across cold reload and upgrade, and the installed-index reader continues consuming that same parser.

## Validation

- 66 focused normalization, persistence, capability, and commit tests
- core and test typechecks
- Plugin SDK declaration generation
- export-name collision guard
- full `pnpm check:changed`
- P0–P2 autoreview: scoped clean

Production diff: **+13 / −70 (net −57 LOC)**

PR #143551, opened 2026-09-10
refactor(config): derive sandbox settings from schemas

RomneyDa:
## Summary

- move Docker, Browser, and Prune sandbox schemas into a leaf ownership module
- derive their authoring contracts from that canonical leaf
- retain SSH as an explicit SecretInput authoring contract and preserve doctor-only `enableNoVnc`

## Why

The sandbox validator repeated large handwritten TypeScript contracts. Leaf schema ownership removes 94 production lines without introducing a config-runtime import cycle.

## Compatibility

Docker keeps its normalized string `setupCommand`, Browser retains legacy `enableNoVnc`, SSH keeps SecretInput authoring values, and resolved runtime sandbox types remain separate.

## Validation

- 126 focused config, doctor, Docker, Browser, and SSH tests
- core and all test-type shards
- Plugin SDK declaration generation
- runtime and static import-cycle guards
- full `pnpm check:changed`
- P0–P2 autoreview: scoped clean

Production diff: **+179 / −273 (net −94 LOC)**

PR #143895, opened 2026-09-10
docs(plugins): correct stale discovery alias and Claude bundle capability claims

vincentkoc:
Two lower-confidence observations from an earlier audit round, tested against source before touching anything. Both held; both were doc defects, not code defects.

## 1. `docs/plugins/architecture-internals/provider-hooks.md` — the `discovery` alias

The page claimed:

> - `discovery` still works as a legacy alias, but emits a deprecation warning
> - if both `catalog` and `discovery` are registered, OpenClaw uses `catalog` and emits a warning

Both statements were true once and are false now.

The alias existed on `ProviderPlugin` as `discovery?: ProviderPluginDiscovery`, and `src/plugins/provider-validation.ts` really did emit both warnings — `provider "<id>" uses deprecated discovery; use catalog` and `provider "<id>" registered both catalog and discovery; using catalog`. All of it was deleted in **c7e7ac2728917f085aff0ad50755fd39aa57f2b3** — *refactor: remove expired plugin compatibility surfaces (#111451)*, 2026-07-19 — which also dropped `resolveProviderCatalogHook`'s `?? provider.discovery` fallback and the `prefers catalog over discovery when both exist` test.

On `main` today:

- `src/plugins/provider-discovery.ts:24-26` — `resolveProviderCatalogHook` returns `provider.catalog`, full stop
- `grep -rn "discovery?:" src/plugins/provider-plugin.types.ts` — no matches
- neither warning string exists anywhere in `src/`, `packages/`, or `extensions/`

So a provider plugin that still registers `discovery` publishes no catalog rows **and** is told nothing about why. The page now says that, which is the part a plugin author actually needs. The `augmentModelCatalog` bullet is untouched — `provider-plugin.types.ts:454-456` still marks it `@deprecated` in favour of `registerModelCatalogProvider`.

## 2. `docs/plugins/bundles.md` — Claude `agents` and `outputStyles`

The page listed Claude `agents` and `outputStyles` under **"Detected but not executed"**, described as "recognized and shown in diagnostics, but OpenClaw does not run them".

The observation that prompted this was that `isBundleCapabilitySupported` returns `true` for both. That alone would not have settled it, so here is what `true` means: its two callers are `inspectBundlePluginArtifact`, which sorts capabilities into `mapped` vs `unavailable`, and `recordBundleDiagnostics`, which emits `bundle capability detected but not wired into OpenClaw yet: <capability>` for exactly the `false` ones. `true` *is* the "wired" side — it is the negation of the doc's category, not a weaker sense of "supported".

And the loading path agrees. `src/plugins/bundle-manifest.ts:427-433`:

```ts
const skillRoots = resolveClaudeComponentPaths(raw, "skills", params.rootDir, ["skills"]);
const commands = resolveClaudeComponentPaths(raw, "commands", params.rootDir, ["commands"]);
const agents = resolveClaudeComponentPaths(raw, "agents", params.rootDir, ["agents"]);
const outputStyles = resolveClaudeComponentPaths(raw, "outputStyles", params.rootDir, ["output-styles"]);
const skills = mergeBundlePathLists(skillRoots, commands, agents, outputStyles);
```

`agents/` and `output-styles/` go into the manifest's `skills` list by the same call that puts `commands/` there — and the page already documented `commands/` as a supported skill root. `src/plugins/bundle-claude-inspect.test.ts:170` asserts `skills: ["skill-packs", "extra-commands", "agents", "output-styles"]`, and `bundle-capability-support.ts:16-17` states the split in a comment: Claude merges agent directories into the runtime skill roots, Cursor detects `.cursor/agents` and never loads it.

So the Cursor row in that list is correct and stays. The Claude row loses `agents` and `outputStyles`, and they move to the supported table, the skill-content bullets, and the Claude bundle accordion. `output-styles/` is also added to the manifestless default layout, where it was missing.

**Claude `hooks/hooks.json` stays in "detected but not executed"** even though `isBundleCapabilitySupported("claude", "hooks")` is also `true`. That `true` is about OpenClaw-style hook packs (`HOOK.md` + `handler.ts`) declared by a Claude bundle, which the supported table already credits to Claude and Codex. The `hooks/hooks.json` default path is a file, not a hook-pack collection directory, and the page's own Hook packs section says it "remains in the declared capabilities, but does not appear as a supported hook". Consistent — nothing to change.

## Nothing was changed in code

Neither finding is a code defect. In both cases the source is the intended behaviour and the docs had drifted behind it.

## Validation

- `pnpm check:docs` — exit 0 (formatting 1280 files clean, markdownlint, lint:templates, mdx, i18n glossary, links, config examples)
- `node scripts/docs-link-audit.mjs --anchors` — 3 broken links, all the pre-existing `maturity/#windows-app-node` ones; `checked_internal_links=13410`
- `npx vitest run --config test/vitest/vitest.tooling.config.ts src/scripts/docs-link-audit.test.ts` — 1 file / 33 tests passed
- `npx vitest run --config test/vitest/vitest.full-core-unit-src.config.ts src/docs/` — 8 files / 16 tests passed
- `npx vitest run --config test/vitest/vitest.plugins.config.ts src/plugins/bundle-claude-inspect.test.ts src/plugins/bundle-manifest.test.ts` — 2 files / 39 tests passed (evidence for §2, not a change of mine)

No glossary entries needed: no `title:` frontmatter changed and no new `/`-route list-item link labels were introduced. No headings renamed, so no anchors moved. Neither page is generated — `scripts/generate-plugin-inventory-doc.mts` owns only `docs/plugins/plugin-inventory.md`, `docs/plugins/reference.md`, and `docs/plugins/reference/`. No CODEOWNERS team matches either path.

### DIFF
--- a/docs/plugins/architecture-internals/provider-hooks.md
+++ b/docs/plugins/architecture-internals/provider-hooks.md
@@ -284,8 +284,9 @@ static catalog rows automatically from `defaultModel`, `models`, and
 
 Compatibility:
 
-- `discovery` still works as a legacy alias, but emits a deprecation warning
-- if both `catalog` and `discovery` are registered, OpenClaw uses `catalog`
-  and emits a warning
+- `discovery` was a legacy alias for `catalog`. OpenClaw removed the alias and
+  the deprecation warnings it emitted
+- rename `discovery` to `catalog`. A provider plugin that still registers
+  `discovery` publishes no catalog rows
 - `augmentModelCatalog` is deprecated; bundled providers should publish
   supplemental rows through `registerModelCatalogProvider`
--- a/docs/plugins/bundles.md
+++ b/docs/plugins/bundles.md
@@ -340,9 +340,10 @@ bundles as trusted content for the features they do expose.
     not wired, that is a product limit, not a broken install.
   </Accordion>
 
-  <Accordion title="Claude command files do not appear">
+  <Accordion title="Claude command, agent, or output-style files do not appear">
     Make sure the bundle is enabled and the markdown files are inside a detected
-    `commands/` or `skills/` root.
+    `skills/`, `commands/`, `agents/`, or `output-styles/` root. All four load
+    through the same skill loader.
   </Accordion>
 
   <Accordion title="Claude settings do not apply">
--- a/docs/plugins/bundles.md
+++ b/docs/plugins/bundles.md
@@ -74,20 +74,22 @@ detected but not wired.
 
 ### Supported now
 
-| Feature       | How it maps                                                                                       | Applies to     |
-| ------------- | ------------------------------------------------------------------------------------------------- | -------------- |
-| Skill content | Bundle skill roots load as normal OpenClaw skills                                                 | All formats    |
-| Commands      | `commands/` and `.cursor/commands/` treated as skill roots                                        | Claude, Cursor |
-| Hook packs    | OpenClaw-style `HOOK.md` + `handler.ts` layouts                                                   | Claude, Codex  |
-| MCP tools     | Bundle MCP config merged into embedded OpenClaw settings; supported stdio and HTTP servers loaded | All formats    |
-| Env contract  | `PLUGIN_ROOT` and `PLUGIN_DATA` env vars plus placeholder expansion for stdio MCP servers         | Agent Plugins  |
-| LSP servers   | Claude `.lsp.json` and manifest-declared `lspServers` merged into embedded OpenClaw LSP defaults  | Claude         |
-| Settings      | Claude `settings.json` imported as embedded OpenClaw defaults                                     | Claude         |
+| Feature                  | How it maps                                                                                       | Applies to     |
+| ------------------------ | ------------------------------------------------------------------------------------------------- | -------------- |
+| Skill content            | Bundle skill roots load as normal OpenClaw skills                                                 | All formats    |
+| Commands                 | `commands/` and `.cursor/commands/` treated as skill roots                                        | Claude, Cursor |
+| Agents and output styles | Claude `agents/` and `output-styles/` treated as skill roots                                      | Claude         |
+| Hook packs               | OpenClaw-style `HOOK.md` + `handler.ts` layouts                                                   | Claude, Codex  |
+| MCP tools                | Bundle MCP config merged into embedded OpenClaw settings; supported stdio and HTTP servers loaded | All formats    |
+| Env contract             | `PLUGIN_ROOT` and `PLUGIN_DATA` env vars plus placeholder expansion for stdio MCP servers         | Agent Plugins  |
+| LSP servers              | Claude `.lsp.json` and manifest-declared `lspServers` merged into embedded OpenClaw LSP defaults  | Claude         |
+| Settings                 | Claude `settings.json` imported as embedded OpenClaw defaults                                     | Claude         |
 
 #### Skill content
 
 - Bundle skill roots load as normal OpenClaw skill roots.
-- Claude `commands/` roots are treated as additional skill roots.
+- Claude `commands/`, `agents/`, and `output-styles/` roots are treated as
+  additional skill roots.
 - Cursor `.cursor/commands/` roots are treated as additional skill roots.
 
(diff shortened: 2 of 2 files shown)


vincentkoc:
Thanks — the P2 is correct and is now fixed at `55d44154ff4`. Responses to all three "Before merge" items:

### [P2] Remove output-styles from standalone manifestless detection guidance — accepted, fixed

You are right, and I conflated two different lists. `output-styles` is a **default component path** (`resolveClaudeComponentPaths(raw, "outputStyles", rootDir, ["output-styles"])`, `bundle-manifest.ts:430`) but it is **not** in `manifestlessClaudeMarkers`, which is `skills`, `commands`, `agents`, `hooks/hooks.json`, `.mcp.json`, `.lsp.json`, `settings.json`. A directory containing only `output-styles/` returns `null` from `detectBundleManifestFormat` and never reaches the manifest reader. Adding it to the manifestless list implied autodetection that does not exist.

`output-styles/` is removed from the manifestless bullet, restoring it to its previous content, and I took your suggested shape: an explicit paragraph saying it is not a detection marker, that an `output-styles`-only bundle needs `.claude-plugin/plugin.json` or another recognized marker, and that detection runs before manifest loading. It stays in the supported-content sections, where it is accurate.

### [P2] "Complete next step" — same item, same fix

Covered by the change above.

### "Review config compatibility" — no config or default surface is changed by this branch

This one is a base-drift artifact rather than a real surface. The review reports "Source -237 across 11 runtime files" and reviewed against `a57399b58aba`, but this branch touches only two files:

```
$ git diff origin/main --stat
 docs/plugins/architecture-internals/provider-hooks.md |  7 ++--
 docs/plugins/bundles.md                               | 37 ++++++++++++++--------
 2 files changed, 27 insertions(+), 17 deletions(-)
```

The eleven runtime files are the schema/type refactors from #143551, #143552 and #143553, which your own evidence confirms merged separately and match fetched `main` byte for byte. They appeared in the delta because the review base predated those merges. The branch has since been rebased onto current `main`, so they should drop out of the next diff. No config schema, default value, or upgrade path is touched by this PR.

### Validation on `55d44154ff4`

- `pnpm check:docs` — exit 0
- `node scripts/docs-link-audit.mjs --anchors` — `checked_internal_links=13415`, `broken_links=3`, all three the pre-existing `maturity/#windows-app-node` failures
- `vitest … vitest.tooling.config.ts src/scripts/docs-link-audit.test.ts` — 1 file / 33 tests passed
- `vitest … vitest.full-core-unit-src.config.ts src/docs/` — 8 files / 16 tests passed

@clawsweeper re-review

vincentkoc:
Merging. Findings: None, and both observations this PR was sent to test turned out to hold — one of them more precisely than it was reported.

**The `discovery` alias claim was stale, not imaginary.** The field existed and did emit both documented warnings (`provider "<id>" uses deprecated discovery; use catalog`, and the both-registered variant), introduced in `311e4608d17`. All of it — field, warnings, the `?? provider.discovery` fallback, and the test — was deleted in `c7e7ac272891` (#111451, 2026-07-19). So today a plugin registering `discovery` publishes nothing **and is told nothing**, which is what the page now says. That distinction is only visible because the clone is full-history again.

**`bundles.md` was wrong in the direction the observation guessed**, but the reasoning matters: `true` in `bundle-capability-support.ts` is not a weaker sense of "supported", it is the exact negation of "detected but not executed" — the `false` branch is what emits `bundle capability detected but not wired into OpenClaw yet`. Claude `hooks/hooks.json` correctly **stays** in that category despite also returning `true`, because that `true` is about OpenClaw-style hook packs, not the default file. A third `true` the original observation did not mention, and the one that would have been wrong to "fix".

**Neither is a code defect** — in both cases the source is the intended behaviour and the docs drifted behind it.

The author's own first pass added `output-styles/` to the manifestless *detection* list, which was wrong (`manifestlessClaudeMarkers` is a fixed set, and detection runs before manifest loading). ClawSweeper flagged it P2 at 0.99 and was right; it is fixed, with a paragraph added on the distinction.

One loose end for a maintainer: the PR body still describes that reverted addition. Editing the body would supersede the completed review and invalidate this verdict, so the correction is in a follow-up comment instead.

PR #143759, opened 2026-09-10
refactor(matrix): unify message action parsing

steipete:
Closes #143756

Related: #130129

## What Problem This Solves

Matrix message actions maintained two parsing layers for the same public requests. The adapter converted public action names and fields into a private dialect, then the runtime parsed them again, making aliases, accounts and validation order harder to keep consistent.

## Why This Change Was Made

The existing lazy Matrix runtime now accepts the existing public action context and owns parsing, action gating and provider dispatch together. The lightweight adapter keeps discovery and send preparation. Provider operations, read authorization, account resolution, profile/verification backends, result projections and client cleanup stay with their existing owners.

Media/voice precedence, own-undefined versus snake-case aliases, ignored private parameters, trusted read context and the placement of validation errors are preserved. Unsupported inherited operation names now receive the normal unsupported-operation diagnostic instead of the accidental private `action required` error; no operation is newly admitted.

No public SDK subpath, configuration option, dependency, protocol, database schema or persistence behavior is added or changed. This does not implement the separate emote feature in #133934. No measured performance improvement is claimed.

## User Impact

Supported Matrix message actions retain their current behavior. Invalid permissions operation names receive a clearer diagnostic. Maintainers have one execution flow to update instead of two private parsing dialects.

LOC classification: Matrix product code +364/-534, net **170 removed**. Tests, shared test fixtures and private QA scenario/data grow by 269 lines. Two assertion-baseline rows shrink to match removed casts. QA Lab is private test infrastructure, counted separately rather than hidden in the product reduction.

## Evidence

- Unchanged original: 59 cases across the original three action suites passed. Migrated public-entry coverage then passed 71 cases on unchanged production.
- Original diagnostic regression: the two inherited-name cases failed with `action required`; the ordinary unsupported-name control passed. Candidate public suite: all 73 cases passed.
- Broader candidate: 149 Matrix action/read-policy cases and 23 QA registration/module-flow cases passed. Useful forwarding coverage is retained at the public boundary; no private compatibility shim was added for tests.
- Normal full build passed in 3m22.9s, including 91 external plugin builds, CLI bootstrap checks, native plugin control-plane checks, public SDK declarations and Control UI build.
[...]

PR #143897, opened 2026-09-10
fix: prevent handoff fixtures from breaking model-command tests

### LINKED ISSUE #143894: Model command tests collide with the managed-handoff legacy fixture
## Summary

Model-command CI cases can fail when the final managed-handoff retarget fixture runs concurrently. This is test-state interference, not a Matrix action regression or a reason to relax the production lease guard.

## Observed failure

[CI run 34444039096, small-3 job](https://github.com/openclaw/openclaw/actions/runs/34444039096/job/102765084654) executed merge `c419c8ab706b01d919a3a52bedfd2838634bf65c` for #143759. Eight `src/commands/models/model-selection.runtime.test.ts` cases failed with:

```text
existing managed handoff lease is incompatible; retain diagnostics and run openclaw triage manually
```

The final `src/infra/update-managed-service-handoff-retarget.test.ts` case deliberately writes a legacy-v1 lease through the default store. That coordinator is independent of HOME, OPENCLAW_STATE_DIR and OPENCLAW_CONFIG_PATH. The config-write owner correctly refuses an undecodable retained row before deciding whether its resource is unrelated.

## Confirmed mechanism

On fresh disposable Linux Testboxes with Node24.19.0, a controlled original-order barrier retained that fixture row while a separate unmocked process attempted an ordinary write to its own synthetic config:

- Original: same coordinator; config callback refused with the exact diagnostic. Real handoff admission, cancellation and cleanup still finished before the final noninterference assertion failed.
- Fixture-isolated candidate: distinct coordinators; config callback succeeded while the actor row stayed unchanged. A direct negative control against the incompatible actor row still rejected it.

The complete native logs were checked through process shutdown, including generated-worker disposal and unhandled-error reporting. The offending row was not extracted from the historical CI runner, so this proves the mechanism rather than claiming that unavailable forensic artifact. Do not run the original unisolated fixture against a maintainer's shared temporary coordinator.

## Repair scope

Use the existing temporary-directory resolver mock only in that final fixture, keeping the actual default store, serialized child database path and real helper lifecycle. No production code, schema, dependency, timeout or guard change is needed. The six-line candidate passes all 28 retarget and 36 model-selection cases, the normal 20-step changed-file gate, and precommit review. A separate fix will link this issue and the affected Matrix validation.

steipete:
Related: #143759
Closes #143894

## What Problem This Solves

Resolves a problem where unrelated model-command tests fail during concurrent CI shards because a managed-handoff test briefly leaves an intentionally incompatible legacy lease in the shared temporary coordinator. This maintainer-requested validation repair is needed to finish the new Matrix action-owner change in #143759.

The observed failure is [CI run 34444039096, small-3 job](https://github.com/openclaw/openclaw/actions/runs/34444039096/job/102765084654): eight model-selection cases report an incompatible handoff lease while the final retarget admission case is running.

## Why This Change Was Made

Scope that final test's existing temporary-directory resolver to its own fixture. HOME and OPENCLAW_STATE_DIR do not select the ordinary handoff coordinator. The default store, real spawned helper, serialized database path, admission, cancellation, sensitive-file cleanup and lease-row assertions all stay real and unchanged.

The production retained-row guard still rejects incompatible leases. No production code, SQLite schema, runtime configuration, dependency, timeout or assertion is weakened or changed. Production LOC: +0/-0. Tests: +6/-0.

## User Impact

No operator-visible runtime change. Independent test shards can write their own synthetic config without interference from this fixture's legacy-row window.

## Evidence

- Controlled original-order reproduction on exact failing CI merge `c419c8ab706b01d919a3a52bedfd2838634bf65c`, using separate real Node processes in a fresh disposable Linux Testbox: the independent config callback is refused while the fixture's legacy row exists. The only failing assertion occurs after the original real handoff cancellation and cleanup. Native actor exit 1, observer exit 0.
- Identical proof with the six-line fix in a second fresh Testbox: distinct coordinators, config callback succeeds, legacy row remains unchanged during observation, and direct access to that incompatible row still rejects. Both native processes exit 0. Both roles use the same Node24.19.0 executable bytes; root read all complete logs through native close, including explicit generated-worker disposal and zero-unhandled-error receipts.
- All non-shipping observation code was removed afterward. The normal runner passes all 28 retarget and 36 model-selection tests. The same 64 cases also pass after moving this separate repair to current-main snapshot `58080e69543eacffdbbc00323ecd3eb2fb06db71` ([fresh Testbox run](https://github.com/openclaw/openclaw/actions/runs/34456216077)); both shards finish in 18.93 seconds. All completed proof leases are stopped.
- Precommit managed Codex review: scoped-clean through P2, zero findings.
- Separate fresh review of exact signed commit `36022ec4d17836b45765ee54cf0044401e3f951f`: scoped-clean through P2, zero findings.
- All 20 normal changed-file checks passed through the configured Testbox route ([run 34456969549](https://github.com/openclaw/openclaw/actions/runs/34456969549)), including the infra test type graph, all three unused-export scans, and targeted type-aware lint. No baseline, timeout, provider, or check was weakened. The one-shot lease is stopped.

The reproduction is source-level synthetic cross-process proof, not an extracted offending database row from CI or a live-provider test. Original/candidate raw archive SHA256 values are `b07a408bb524f91617acc6f50fed83ad4ca6bdd0d11166d4b626c9961ac2251c` and `71063c30c66773d58d04e4ea4124cd44055daa69cc620f9eb864fbc5bf940dcf`. Do not run the original unisolated fixture against a maintainer's shared coordinator.

### DIFF
--- a/src/infra/update-managed-service-handoff-retarget.test.ts
+++ b/src/infra/update-managed-service-handoff-retarget.test.ts
@@ -7,6 +7,7 @@ import { afterEach, expect, it, vi } from "vitest";
 import { resolveServiceManagerEnv } from "../daemon/service-process-env.js";
 import * as pidIdentity from "../shared/pid-alive.js";
 import * as nodeSqlite from "./node-sqlite.js";
+import * as tmpDir from "./tmp-openclaw-dir.js";
 import { createManagedHandoffLeaseStore as createStore } from "./update-managed-service-handoff-lease.js";
 import type {
   createManagedHandoffLeaseStore,
@@ -610,6 +611,11 @@ unix(
     const { createManagedHandoffLeaseStore, resolveManagedUpdateLeaseDatabasePath } =
       await import("./update-managed-service-handoff-lease.js");
     const { to } = fixture();
+    // The ordinary parent/helper coordinator is independent of OPENCLAW_STATE_DIR.
+    // Keep its legacy-row admission window private to this fixture.
+    vi.spyOn(tmpDir, "resolvePreferredOpenClawTmpDir").mockReturnValue(
+      path.join(path.dirname(to), "common-coordinator"),
+    );
     const store = createManagedHandoffLeaseStore();
     const reserved = store.acquire(to, "legacy-owner", { kind: "update" });
     expect(reserved.kind).toBe("acquired");


steipete:
Superseded by #143875, already merged as [7207976bd2fc72631f1622f71ff4363b8e383b94](https://github.com/openclaw/openclaw/commit/7207976bd2fc72631f1622f71ff4363b8e383b94).

Verified that commit's raw parent (`025d572de58f0895afd7ffcee00ab53797e190ff`) and its five-line fixture-only patch, and confirmed it is an ancestor of current-main snapshot `3e6b53c492d9bd6019983efea1c890b35a930146`. The target test blob on that main snapshot is exactly the landed canonical blob `1f458c688706fb39d6034436aa2d4567074639af`.

Its exact source head `fbfad9a43f91a804bd8758533ed45744c3c38bee` also has [successful CI run 34455590003](https://github.com/openclaw/openclaw/actions/runs/34455590003).

Both implementations redirect the same existing resolver before constructing the real default store, then serialize that private path to the real helper. Comparing the complete files leaves only static versus awaited namespace import, comments, and the fixture-private directory name. The lease guard, case inventory, helper admission/cancellation and cleanup are unchanged. Our separate deterministic original/candidate probe and 64-case replay support the same boundary; they are not claimed as executions of the canonical commit.

Closing this duplicate without merging or replaying its patch. Canonical repair credit and history remain with #143875. This also resolves #143894 and provides the fixed main baseline needed by #143759.

PR #143916, opened 2026-09-10
fix: keep session reset responsive and stop yielded child work

### LINKED ISSUE #143389: [Bug]: WhatsApp /new stalls ~82s on retained subagent tree; per-node scope.refresh repeats full traversal

obviyus:
Closes #143389. Related: #143620. Thanks @aleps001 for profiling and @zhangguiping-xydt for the related traversal repair.

## What Problem This Solves

/new resets slowed down when history contained many completed child runs. A yielded child could also start another turn after reset acknowledged cancellation.

## Why This Change Was Made

Traversal no longer refreshes the whole tree at each node. Existing ownership checks remain. Completion retains actual yield and accepted-cancellation facts, and reconciliation defers to an active session cancellation before retiring obsolete wake requests.

## User Impact

Resets stay responsive while active and yielded child work remains cancelled. Completed results, task records, and unrelated conversations remain available. Configuration, schemas, and the existing reconciliation grace are unchanged.

## Evidence

Real Gateway chat, child spawning, yielding, and /new checks used a recording provider. A matched 64-node reset improved from 2.180 to 0.220 seconds with all records retained and zero requests during reset. Active and yielded cancellation passed 390-second observation windows without new provider requests. The production repair passed 173 focused tests; the test follow-up passed 37. One staged continuous-integration run observed an unexplained request that smaller reruns did not reproduce; its strict assertion remains. The reported 82-second workload was not reproduced, and a single memory sample was higher after the repair.

### DIFF
--- a/src/agents/subagents/announce/subagent-announce.requester-settle-cancel.test.ts
+++ b/src/agents/subagents/announce/subagent-announce.requester-settle-cancel.test.ts
@@ -24,6 +24,7 @@ import {
   setSubagentAnnounceDeliveryDepsForTest,
   type SubagentAnnounceDeliveryDeps,
 } from "./subagent-announce-delivery.runtime.js";
+import { dispatchGatewayMethodInProcess } from "./subagent-announce.runtime.js";
 
 const fixture = useSubagentControlFixture();
 afterEach(() => setSubagentAnnounceDeliveryDepsForTest());
@@ -91,7 +92,8 @@ it.each([
     const waitBeforeExecution =
       phase === "admitted" || phase === "requester reset" || phase === "requester replacement";
     type Dispatch = SubagentAnnounceDeliveryDeps["dispatchGatewayMethodInProcess"];
-    const completed = vi.fn<Dispatch>().mockResolvedValue({
+    const completion = { dispatch: dispatchGatewayMethodInProcess };
+    vi.spyOn(completion, "dispatch").mockResolvedValue({
       status: "ok",
       result: { payloads: [{ text: "The child has settled." }], meta: {} },
     });
@@ -109,7 +111,7 @@ it.each([
       }
       options?.onExecutionStarted?.();
       startedTurns.push(String(params?.sessionKey));
-      return await completed<T>(...args);
+      return await completion.dispatch<T>(...args);
     };
     setSubagentAnnounceDeliveryDepsForTest({ dispatchGatewayMethodInProcess: dispatch });
 
--- a/src/agents/subagents/announce/subagent-announce.requester-settle-cancel.test.ts
+++ b/src/agents/subagents/announce/subagent-announce.requester-settle-cancel.test.ts
@@ -1,0 +1,180 @@
+import { afterEach, expect, it, vi } from "vitest";
+import { getRuntimeConfig } from "../../../config/config.js";
+import { patchSessionEntryCore } from "../../../config/sessions/session-accessor.js";
+import { createDeferredCore } from "../../../shared/deferred.js";
+import { findTaskByRunId } from "../../../tasks/task-registry.js";
+import { killSessionSubagentRuns } from "../registry/subagent-control-kill.js";
+import { useSubagentControlFixture } from "../registry/subagent-control.test-support.js";
+import { subagentRuns } from "../registry/subagent-registry-memory.js";
+import { markSubagentRunPausedAfterYield } from "../registry/subagent-registry-run-manager.js";
+import { persistSubagentRunsToDiskOrThrow } from "../registry/subagent-registry-state.js";
+import {
+  adoptPausedSubagentRunForFollowUp,
+  markRequesterTurnYielded,
+  markSubagentRunTerminated,
+  registerSubagentRun,
+  settleRequesterAfterSessionSpawns,
+} from "../registry/subagent-registry.js";
+import {
+  settleSubagentRegistryPersistenceWork,
+  writeSubagentSessionEntry,
+} from "../registry/subagent-registry.persistence.test-support.js";
+import { testing as registryTesting } from "../registry/subagent-registry.test-helpers.js";
+import {
+  setSubagentAnnounceDeliveryDepsForTest,
+  type SubagentAnnounceDeliveryDeps,
+} from "./subagent-announce-delivery.runtime.js";
+
+const fixture = useSubagentControlFixture();
+afterEach(() => setSubagentAnnounceDeliveryDepsForTest());
+
+it.each([
+  "pending",
+  "admitted",
+  "unsuppressed",
+  "failed kill",
+  "requester reset",
+  "requester replacement",
+] as const)(
+  "settles a yielded requester's child after %s cancellation without reviving cancelled work",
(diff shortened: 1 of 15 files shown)


obviyus:
Thanks @aleps001 for the profiling report and @zhangguiping-xydt for the related traversal repair in #143620.

Landed as 03e52642753252de66492e60ac90779218f5521b.

**Proof:** Authenticated WebChat spawn/yield/reset flows passed, including 390-second active and yielded cancellation observations with no renewed provider inference and retained task results and histories. The matched 64-node reset measured 2.180 seconds before and 0.220 seconds after.

- **Bug:** Retained child trees caused repeated full-tree discovery during reset, and yielded work could resume after cancellation.
- **Fix:** Remove redundant discovery and preserve cancellation ownership through interruption, completion, and delayed requester continuation.
- **Validation:** Final focused tests and native commit checks passed; exact-head ClawSweeper review reported no findings; [CI run 34462914716](https://github.com/openclaw/openclaw/actions/runs/34462914716) passed without a CI exception.
- **Limit:** The earlier staged unmarked provider request remains unexplained. Its failure evidence and strict recurrence diagnostics are retained; later passing checks do not establish its cause. The single candidate memory sample was higher, and no memory improvement is claimed.

| Before | After |
| --- | --- |
| Repeated selected-tree refresh during each node visit | Refresh remains with the existing convergence and retargeting owners |
| Yielded children could resume after reset | Active and yielded work stays cancelled through the observed natural reconciliation window |
| Retained history makes reset slow | Task results and histories remain retained with faster measured reset |

PR #143791, opened 2026-09-10
fix(update): fence retained rollback and simplify recovery lifecycle

vincentkoc:
Fixes #143750

Issue: https://github.com/openclaw/openclaw/issues/143750
Related: https://github.com/openclaw/openclaw/issues/143752 (package activation recovery remains separate).

Builds on Jason (@fuller-stack-dev)'s rollback work in https://github.com/openclaw/openclaw/pull/140339. His contribution on this branch is preserved in the signed merge history; the connected lifecycle cleanup is now included in this PR, not a separate stack.

## What Problem This Solves

Fixes an issue where a managed update could continue replacing package or launcher entries after its original executor lost ownership during rollback. The outer check detected ownership loss only after those filesystem effects.

## Why This Change Was Made

Carries the original executor assertion through retained rollback and completion, including awaited planning, package restoration, launcher compensation, cross-device copy-source cleanup, and backup retirement. Repeated calls remain bound to that first owner.

Removes dormant package-recovery execution/writers and the no-op replay prepass, leaving one active lifecycle. Persisted record schemas, readers, admission checks, and refusal behavior remain intact. Native EXDEV and relative-launcher behavior are preserved.

## User Impact

Ownership loss stops subsequent protected filesystem operations and preserves recovery material. Already-dispatched filesystem operations are not cancellable.

This does not provide crash-atomic activation, interrupted-package repair, a new recovery journal, or full-state/migration-crossing rollback. The activation gap in https://github.com/openclaw/openclaw/issues/143752 is not fixed by this PR.

## Evidence

Head: `a0fe5c66827dc6730aa4768e83615f25f1fa0a41`; tree: `2551da6a71639ae2e7f59f8f398c974e90b6fc45`.

- Real retained-transaction regressions fail on the original implementation with unauthorized package/launcher mutations.
- Final-source remote proof: all three canonical dead-export scans reported zero; `update-command-run.test.ts` passed 35 tests with zero skips; full `pnpm build` passed in 4m20.5s; isolated dist-backed `node openclaw.mjs update --help` passed. The command wrapper exited 0. Proof run: https://github.com/openclaw/openclaw/actions/runs/34453609190 (selected source is the head above, not the workflow-control branch).
[...]

PR #143767, opened 2026-09-10
fix(update): finish the update with recorded warnings on recoverable hiccups

steipete:
<details>
<summary>Additional instructions</summary>

This is a same-repository branch; maintainers can update it through repository write access. Landing follows the native maintainer workflow.

</details>

## What Problem This Solves

Fixes an issue where `openclaw update` aborts an otherwise viable update because disposable validation copies, retired derived caches, or Git upstream tracking could not be cleaned up or refreshed. Successful Doctor warnings could also disappear between the child process, update ledger, and human-readable report.

Approval for the severity changes is Peter's update contract (2026-09-10): **“update should always work. It's best effort. If there are minor hiccups we add that in the warning log.”** Ownership, data compatibility, consent, cancellation, and Gateway boot/readiness failures remain blocking. No backup/rollback path, user configuration option, SQLite schema, dependency, protocol version, or retention rule changes.

Refs #143138, #143141, #143153, #143325, #143310, #143344. These merged campaign repairs establish the `warningDisposition: "recoverable"` precedent. This PR contains independently authored follow-up fixes; no contributor code was copied from another PR.

## Why This Change Was Made

Classify a hiccup at its owner, where successful import, disposable state, or a validated runtime can be distinguished from uncertain live state. Keep original failure diagnostics, record actionable advisories, and continue the remaining validation and activation steps. Doctor still stops at unclassified or explicitly refused migrations.

Warning records use the existing bounded update-step ledger. They do not infer warnings from arbitrary success messages or convert a hard-failing Doctor into success. Historical warnings appear in update status and Doctor as recorded history, since a later repair may already have resolved them.

## User Impact

Recoverable housekeeping failures no longer roll back or abort a verified update. The report names the skipped work, explains the failure, and supplies `openclaw doctor --fix` or the exact scoped cleanup/Git repair command. The Gateway must still pass its existing activation checks.

Release-note context: Updates finish with recorded warnings when disposable validation cleanup, retired derived-cache cleanup, or optional Git tracking setup fails; data, ownership, and Gateway readiness refusals remain enforced.

## Evidence

[...]

PR #143921, opened 2026-09-10
fix(update): preserve cleanup outcomes and settle terminal results

fuller-stack-dev:
Related: #143767, #143791

## What Problem This Solves

Fixes an issue where `openclaw update` could print success and mark history succeeded before the original executor finished its final ownership checks and release. A late release failure then produced a nonzero exit inconsistent with the published result.

The updater also discarded ordinary package-completion outcomes. A denied backup removal or fallback rename could leave files behind without showing their actual retained path, and repeated completion could lose the original failure.

## Why This Change Was Made

Complete verified package cleanup before preparing the report, then publish the terminal result only after the existing outer executor promise settles. Reuse the warning contract landed in #143767. Only the package owner's verified obsolete-directory cleanup can produce a maintenance advisory; unverified recovery, link identity/read failures, authority loss, and unknown exceptions remain errors.

The package transaction retains one completion outcome, including link retirement and the last post-await authority check. Repeated calls preserve that outcome, while the original caller assertion remains mandatory. Existing terminal history remains immutable.

## User Impact

Successful activation with obsolete-backup cleanup pending now reports the actual retained path and a manual follow-up. Late authority or release failures produce one error result instead of an earlier success. Ordinary healthy updates remain successful.

This changes diagnostics and ordinary package cleanup only. It does not introduce recovery storage, change child-custody APIs, or grant full-state retirement authority.

## Evidence

- 21 composed cases passed using the real package swap/transaction, finalizer, executor, SQLite lease, history, and JSON/human report consumers. Faults include a final lease DELETE failure, authority loss during completion, retained and renamed backups, repeated completion, foreign terminal history, unverified restoration, and native project/link refusal.
- The final two review findings were reproduced by three targeted failing cases before repair. All 21 cases then passed.
- `check:changed` passed at frozen source tree `7bb508f9b9b96d56aef70263964bc04a3814c5ca`, including production/test types, dead exports, lint, and repository guards. The published seven-file delta is byte-identical to that tested source on main cbc245824eb821a4f7ad1bc68cad5d143674f84c. That qualification is preserved; the follow-up qualification below covers the new five-file delta.
- P0/P1/P2 independent review on the prior tree found two P2s, both resolved. This is findings-resolved review, not a second scoped-clean review claim.
- All ten canonical warning-contract files and the executor/filesystem/npm-root/native implementation files are unchanged by this residual patch. #143767 is already merged; its changes are not copied into this PR.

The local composed tests use disposable state. They do not claim installed-service, surviving-child, Doctor/plugin effect closure, or whole-journey rollback proof. Those remain separate work.

## Follow-up corrections

- Fixed two existing tests that read history inside cleanup. They now expect `running` there and still require `rolled-back` after completion. The affected routed run passed 211 tests and the changed-file gate.
- Addressed the review finding where an already-published pending recovery failure could enter managed diagnostic triage after executor revocation. The wrapper now retains the existing finalized, exit-only marker. Ordinary authorized failure triage keeps its original metadata and exit code.
- Real helper/borrowed-executor regressions first reproduced unsafe diagnostic creation and replacement in two revoked cases. After the fix, both leave artifacts absent or unchanged, perform zero atomic opens/renames, report once, and exit 1. The authorized control still writes its diagnostic and exits 7.
- The focused follow-up passed 80 tests. After a test-only lint correction, all three new cases passed on final tree `5434dc900c919dadc9b955fab64a563510d12c48`; remaining production and test bytes were unchanged. Required changed-file checks passed on that tree, and a fresh exact-final five-file P0/P1/P2 review was scoped-clean.
- The original CI failure and prior review/test evidence are preserved. Exact-head hosted CI is required for the updated PR before landing. No installed or full-state recovery qualification is implied.

### DIFF
--- a/docs/cli/update/status-and-history.md
+++ b/docs/cli/update/status-and-history.md
@@ -86,6 +86,18 @@ Doctor also shows warnings from the latest run as historical observations: a lat
 repair may already have resolved them. The existing report and history size limits
 still apply.
 
+A foreground updater publishes its final result after required finalization work
+and its local executor have settled. A late ownership or release failure returns
+an error instead of publishing an earlier success. Existing terminal history is
+not overwritten.
+
+Successful installation verification does not imply that obsolete package backups
+were deleted. If the package owner confirms that only obsolete-backup cleanup is
+pending, JSON, history, and human reports include a warning with the retained path
+and follow-up guidance. Unverified recovery, unreadable backup state, and unknown
+completion failures remain errors. Inspect retained paths before manually removing
+obsolete backups; unresolved recovery material is not eligible for this cleanup.
+
 Gateway clients with `operator.admin` can inspect history:
 
 ```bash
--- a/src/cli/update-cli/update-command-generation.test-support.ts
+++ b/src/cli/update-cli/update-command-generation.test-support.ts
@@ -243,7 +243,8 @@ export function registerGenerationRecoveryTests(
           after: { version: VERSION },
           verification: { serviceRunning: true, runningVersion: VERSION },
         });
-        expect(completedStatus).toBe("rolled-back");
+        // Cleanup is pre-terminal; rollback is recorded only after completion settles.
+        expect(completedStatus).toBe("running");
         expect(record.downtimeMs).toBeGreaterThanOrEqual(0);
         expect(record.confirmedAtMs).toBeGreaterThanOrEqual(before.stoppedAtMs!);
         expect(renderUpdateRunReport(record).headline).toBe(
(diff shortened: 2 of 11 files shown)


fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [6ef04835ccc51587abbdf2c620b5ca6f49ff1694](https://github.com/openclaw/openclaw/pull/143921/commits/6ef04835ccc51587abbdf2c620b5ca6f49ff1694)
- Landed commit: [1dfcddb5c98eb2f5922f9966153886e04a7f2878](https://github.com/openclaw/openclaw/commit/1dfcddb5c98eb2f5922f9966153886e04a7f2878)

PR #143930, opened 2026-09-10
fix(trajectory): preserve paginated read evidence

### LINKED ISSUE #143830: Paginated read results are falsely redacted during trajectory export
# OpenClaw 2026.9.1: paginated read results are falsely redacted during trajectory export

## Summary

OpenClaw 2026.9.1's diagnostic projection replaces a successful `read` tool
result with `[Malformed diagnostic JSON redacted]` when the result contains the
normal numeric pagination notice. The result remains marked `isError: false`.

This corrupts trajectory evidence and can make a model's subsequent answer
appear unsupported, even though the underlying read operation succeeded.

## Minimal reproduction

The following isolated probe imports the installed `read` implementation and
diagnostic projection, injects a four-line text fixture, and performs no
Gateway, Docker, Ollama, configuration, or file-write operations:

```bash
node scripts/diagnostics/probe-openclaw-read.mjs /absolute/path/to/openclaw
```

Environment tested:

- OpenClaw: `2026.9.1`
- Node: `v24.19.0`
- Fixture:

  ```text
  Project: Orion
  Owner: Vega
  Build: 4172
  Status: ready
  ```

The smallest failing input to the sanitizer is:

```text
[2 more lines in file. Use offset=3 to continue.]
```

The installed read implementation returns that notice unchanged, but
diagnostic projection returns:

```text
[Malformed diagnostic JSON redacted]
```

The isolated matrix produced:

| Read arguments | Read result | Diagnostic projection |
|---|---|---|
| path only | preserved | preserved |
| `optional: true` | preserved | preserved |
| `limit: 2` | first two lines + notice | malformed sentinel |
| `offset: 1, limit: 2, optional: true` | first two lines + notice | malformed sentinel |
| `limit: 4` | complete file | preserved |
| `offset: 3, limit: 2` | final two lines | preserved |

`offset: 0` continues to produce a genuine tool error, so unsuccessful tool
errors should not be treated as this issue.

## Evidence and source boundary

The read implementation constructs the notice in `createBoundedReadTextPage`.
The trajectory exporter passes transcript messages through
`sanitizeDiagnosticPayload`, which delegates to
`@openclaw/ai/diagnostics`.

The diagnostic dependency uses a heuristic equivalent to:

```js
const JSON_ARRAY_START_RE = /\[\s*(?:[{"\d\]-]|true\b|false\b|null\b)/u;
```

That heuristic classifies the leading `[2` in the human-readable notice as a
JSON array. JSON parsing then fails, and the whole string is replaced by the
malformed-JSON sentinel. The model-visible tool-payload redaction path
preserves the notice, so the corruption is specific to the diagnostic/export
projection boundary.

The same successful-result shape was captured twice in Llama 3.1 8B
benchmark sessions: valid `read` arguments, `isError: false`, and the exact
sentinel in the exported trajectory. The raw/wrapped read probe reproduces the
same transformation deterministically.

## Expected behavior

Harmless human-readable tool text should be preserved byte-for-byte by
diagnostic projection. Credential and media redaction should continue to work,
and genuinely malformed diagnostic JSON should retain its existing protection.

Potential fixes include tightening the JSON-array detection heuristic or
excluding known tool-result pagination notices from diagnostic JSON detection.
Please add a regression test covering numeric bracketed text with ordinary
prose, including the OpenClaw pagination notice.

## Benchmark-side mitigation

The benchmark harness fails closed when a successful result contains the
sentinel. Its temporary repair path reads the exact session's read-only SQLite
transcript, verifies session/run/call/tool identity and success state, retains
the original export, and records repair provenance. It does not reconstruct
file contents or weaken the check for unsuccessful tool errors.

linhongyu510:
Closes #143830

## What Problem This Solves

Fixes an issue where users exporting an agent trajectory after a paginated `read` would lose successful read text. The export replaced the text and continuation notice with `[Malformed diagnostic JSON redacted]`, including pages containing harmless JSON.

## Why This Change Was Made

The shared sanitizer preserves complete plain-bracket fragments while retaining broad JSON detection and bounded parsing. It checks the complete outer fragment, so an inner notice or a partial tag cannot hide malformed sensitive content. Prose exemptions use the existing credential and media rules, including typed media context.

## User Impact

Exported trajectories retain successful read evidence and continuation instructions. Ordinary directives, complete reads, and final pages remain readable. Read errors stay errors, and credentials and media stay redacted.

## Evidence

- Pinned main `da7e3b6` reproduced the lost text through a real isolated Gateway agent turn, the shipped `read` tool, session persistence, and public `openclaw sessions export-trajectory`.
- Final build `eddbec63` preserved the complete 77-byte paginated result in both runtime and transcript exports, with `isError: false`.
- Fresh source-blind acceptance passed all 23 additional public controls, including the 21,053-byte paginated transcript, safe JSON/directives, error identity, and the new malformed-fragment/assignment cases. Runtime capture retains its existing long-text limit; the complete transcript export remains byte-equal.
- Focused source-mode tests: 3 files, 337 passed. Formatting, syntax lint, file-size and assertion-safety checks passed. Fresh independent source review found no remaining actionable issue.
- Regression probes cover mixed JSON/prose, malformed numeric and nonnumeric fragments, nested notices, short credentials, typed media, chained assignments, incomplete private keys, and the structured-data size bound. The earlier test-type error was corrected at its caller; no typecheck exception is used.

Captured public-route text:

```text
Read / final export:
Project: Orion
Owner: Vega

[2 more lines in file. Use offset=3 to continue.]

Baseline export:
[Malformed diagnostic JSON redacted]
```

The existing projection may replace detailed read errors and credential-bearing JSON strings with the redaction marker. Baseline-owner checks preserve that limitation rather than claiming harmless fields survive every sensitive projection.

The local endpoint generated deterministic traffic; it did not supply independent model reasoning. Both original commits by @linhongyu510 are retained, together with the subsequent review corrections and failure history.

### DIFF
--- a/packages/ai/src/utils/credential-redaction.ts
+++ b/packages/ai/src/utils/credential-redaction.ts
@@ -1,5 +1,9 @@
 import { estimateBase64DecodedBytes } from "@openclaw/media-core/base64";
-import { extractBalancedJsonFragments, stableStringify } from "@openclaw/normalization-core";
+import {
+  expectDefined,
+  extractBalancedJsonFragments,
+  stableStringify,
+} from "@openclaw/normalization-core";
 import { parseRetryAfterHeadersSeconds } from "../internal/retry-after.js";
 
 const NON_CREDENTIAL_FIELD_NAMES = new Set([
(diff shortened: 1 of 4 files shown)


linhongyu510:
Addressed the review false positive and strengthened the behavior proof in `0b91e9bbec`:

- ordinary double-bracket directives and hyphenated `[true-*]` / `[false-*]` / `[null-*]` tags are preserved;
- malformed structured arrays, including sensitive media payloads, still fail closed;
- the trajectory regression now calls the production `read` tool to produce `[12 more lines in file. Use offset=4 to continue.]` before exporting it;
- an additional production-path proof persisted that real tool result through `SessionManager.open(...).appendMessage(...)`, then exported it byte-for-byte with `isError: false`;
- focused suite: 3 files / 254 tests; `pnpm check:changed` passed; final independent P0-P2 review was scoped-clean.

linhongyu510:
@clawsweeper re-review

I added the requested inspectable, sanitized production-path output directly to the PR body. It contains only the read pagination notice and boolean/count verification fields; no private path, endpoint, credential, or user data.

obviyus:
Thanks @linhongyu510! Merged as 949beec095c22fac9db7246e4c714396a599b54c.

- **Bug:** Successful paginated file reads became `[Malformed diagnostic JSON redacted]` in trajectory exports.
- **Fix:** The shared sanitizer preserves complete safe bracketed prose, including pagination notices, while retaining credential, media, malformed-sensitive-JSON, and size safeguards.
- **Proof:** A real Gateway agent turn used the shipped `read` tool, persisted its result, and exported the same session through `sessions export-trajectory`. The baseline lost the text; the final build preserved it byte-for-byte. Independent acceptance passed all 24 public cases, including JSON-led pages and a 21,053-byte transcript result. The existing runtime-capture length limit is unchanged.
- **Validation:** 337 focused tests passed, fresh independent source and ClawSweeper reviews cleared the final change, and [final-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34482792158/job/102897277414). The merged files match the tested head.

PR #143933, opened 2026-09-10
refactor(tui): compact command handler test cases

RomneyDa:
## What Problem This Solves

The TUI command-handler suite repeats object property labels across table-driven cases, making the test data longer and harder to scan than the behavior it covers.

## Why This Change Was Made

Convert homogeneous Vitest object tables to positional tuples while preserving every case, value, title, and assertion. Production code is unchanged.

## User Impact

No user-visible behavior changes. Maintainers get a smaller, easier-to-review command-handler suite.

## Evidence

- 236 focused tests pass
- `pnpm check:changed` passes on the exact head
- P3 autoreview: scoped clean (0.99)
- Net change: 66 LOC removed

### DIFF
--- a/src/tui/tui-command-handlers.test.ts
+++ b/src/tui/tui-command-handlers.test.ts
@@ -2393,37 +2393,34 @@ describe("tui command handlers", () => {
   );
 
   it.each([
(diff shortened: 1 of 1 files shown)


RomneyDa:
@clawsweeper review

PR #143935, opened 2026-09-10
refactor(tui): compact session action test cases

RomneyDa:
## What Problem This Solves

The TUI session-action suite repeats property names in table-driven state cases, adding noise around the routing conditions under test.

## Why This Change Was Made

Convert homogeneous Vitest object tables to positional tuples while preserving every row, optional value, generated title, and assertion. Production behavior is unchanged.

## User Impact

No user-visible behavior changes. Maintainers get a smaller, easier-to-scan session-action suite.

## Evidence

- 103 focused tests pass
- `pnpm check:changed` passes on the exact head
- P3 autoreview: scoped clean (0.99)
- Net change: 54 LOC removed

### DIFF
--- a/src/tui/tui-session-actions.test.ts
+++ b/src/tui/tui-session-actions.test.ts
@@ -1474,45 +1474,15 @@ describe("tui session actions", () => {
   });
 
   it.each([
-    {
-      description: "a different session",
-      sameSession: false,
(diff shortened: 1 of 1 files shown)


RomneyDa:
@clawsweeper review

PR #143937, opened 2026-09-10
fix(redaction): replay only provenance-marked persisted masks (#142821)

### LINKED ISSUE #142821: [Bug]: Default-on model-visible transcript redaction poisons replayed agent    context (masks propagate into commands, files, replies)
### Bug type

### Beta release blocker

No

## Summary

OpenClaw 9.2 applies a "model-visible" redaction battery to session transcript messages **at persist time**, on by default. The battery matches token **shapes** and field **names** — not registered secret values — so it masks far beyond actual secrets. The live in-turn tool result is delivered to the model unredacted, but everything later rebuilt from the transcript store (gateway restart, session resume, reconcile, compaction) returns the masked form. After any replay, the model's context contains masks where real values used to be, and the model composes new commands, files, and replies from that masked data.

Consequences observed on the affected production install:

- Credential-bearing shell commands composed from replayed context sent the mask string as the credential → upstream API rejected with 403. The same key worked from a host running an older OpenClaw.
- Files authored by the agent permanently embedded masks — including a skill file created specifically to instruct the agent to stop emitting masks; the mask propagated into the very artifact meant to prevent it.
- Replies contained masks, and operator instructions to stop could not be followed — the model does not possess the original strings after replay, so compliance is impossible by construction.

## Mechanism (code refs, 2026.9.2, commit 3928bad)

1. **Write-side application.** Every persisted message passes `redactTranscriptMessageForStorage` before the sqlite transcript write (`dist/session-accessor-YsytfDtG.js` lines 153, 212, 1887, 4404 — e.g. `JSON.stringify(redactTranscriptMessageForStorage(message, …))`). The store never holds the plaintext for matched spans.

2. **Model-visible routing for tool results.** In `dist/session-accessor.sqlite-transcript-store-B0zw3fAU.js` (lines ~411–416, 688, 710): when a message has `role === "toolResult"` and the field is `content`, redaction switches to the model-visible variants (`redactModelVisibleToolPayloadTextWithConfig`, `redactModelVisibleSensitiveFieldValueWithConfig`). Assistant tool-call `arguments` and assistant reasoning text are also walked by the same structured redaction.

3. **The battery is shape- and name-based.** `dist/redact-BtvPPfTi.js`:
   - `DEFAULT_REDACT_PATTERNS`: ~100 literal token-shape regexes (`sk-[A-Za-z0-9_-]{8,}`, `ghp_…`, `AIza…`, JWT shapes, AWS/GitLab/Slack/Stripe token shapes, …) plus `Authorization: Bearer/Basic`, HTTP auth-header, env-assignment, and config-assignment patterns.
   - `TOOL_PAYLOAD_REDACT_PATTERNS`: the same minus ambiguous assignment patterns, used for model-visible tool payload text.
   - `STRUCTURED_SECRET_FIELD_RE` / `STRUCTURED_SECRET_ENV_FIELD_RE`: field names including `key`, `token`, `secret`, `session`, `jwt`, `signature`, `cookie`, `authorization`, and any `*_KEY` / `*_TOKEN` / `*_SECRET` / `*_PASSWORD` shape (case-insensitive) force the **value** to be masked regardless of content.
   - Mask forms: `***` for header/assignment/field-name hits; partial `first6…last4` for bare token-shape hits (`maskToken`, `DEFAULT_REDACT_KEEP_START = 6`).

4. **No registration required.** The exact-value registry (`dist/secret-redaction-registry-BOLC6DkF.js`) only contains values registered through SecretRef sentinel minting (`dist/sentinel-ZgmfESkJ.js` — `mintSecretSentinel`/`resolveSecretSentinel`). Plaintext config values and environment credentials are never registered, so for them the pattern battery is the only layer in play, and it fires on shape alone. (Docs, `docs/gateway/secrets.md`: plain env credentials are "outside this mechanism" — meaning sentinels; the transcript battery is not.)

5. **Live/replay divergence.** The in-turn tool result returned to the model is *not* passed through this battery (verified: a `cat` of a dummy key returned full plaintext live). Only the persisted copy is redacted. Every later read — session resume, context rebuild after restart/reconcile, compaction windows, `sessions_history` — serves the masked form.

## Repro (clean 9.2 box; synthetic dummy values only — no real credentials involved)

1. Generate a throwaway dummy key **inside the shell** so it is never seen outside tool results:
   `umask 177; printf 'sk-bugreport-dummy-%s' "$(openssl rand -hex 8)" > /tmp/keyfile; chmod 600 /tmp/keyfile`
2. Ask the agent to `cat /tmp/keyfile` via the exec tool.
   - **Live tool result (what the model sees in-turn): the full dummy value.**
3. Read the session's own history back (`sessions_history`, `includeTools: true`):
   - The same toolResult now presents as `sk-bug…<last4>` (partial mask).
   - Assistant toolCall arguments that mentioned the value are masked in storage too.
   - Assistant reasoning text that quoted the value is masked as well.
   - Result: after a gateway restart / session resume / compaction, the model's context contains only masked forms. A subsequent `curl -H "Authorization: Bearer …"` composed from that context carries the mask as the credential and the provider rejects it (401/403). On the affected production install this was observed with the literal `***` form as the rejected credential.
4. Library-level over-match (importing the shipped `dist/` modules directly on the same build):
   - An **unregistered** lookalike string (`sk-` + ≥8 chars, never a secret, never registered) is masked by the model-visible tool-payload redactor.
   - A structured result `{"key":"leftshift","token":"146","session":"abc","note":"keep me"}` redacts to `{"key":"***","token":"***","session":"***","note":"keep me"}` — benign values masked purely by field name.
   - A benign JWT-shaped string is partially masked.

## Expected

- Secret redaction belongs on surfaces humans read (logs, UI, exports). The model's working context and its persisted source of truth must carry ground truth — or a recoverable handle, exactly as the existing `oc-sent-v2…` sentinel mechanism already does for SecretRef-backed provider credentials at egress.
- At minimum: shape/name-based masking should not be default-on for persisted transcripts; unregistered lookalikes and benign field names (`key`, `token`, `session`, …) should not match; and if transcripts are redacted at rest, the unmasked ground truth must remain recoverable for context rebuild.
- Agent-authored artifacts (commands, files) must never silently embed redaction masks.

## Actual

- Default-on, shape- and name-based masking applied at persist time to tool results, tool-call arguments, and reasoning text; live/replay divergence with no recovery path for non-SecretRef credentials; downstream credential rejection (403), permanent artifact poisoning, and an agent that cannot comply with instructions to stop emitting masks.

## Version comparison

- **2026.7.1 does not exhibit** (reporter's production data point: downgrade eliminated the symptom).
- Source at tag `v2026.7.1`, `src/agents/transcript-redact.ts`: transcript redaction is **opt-in** — `redactTranscriptOptions()` returns `undefined` when neither `logging.redactSensitive` nor `logging.redactPatterns` is configured — and the module imports only log-grade redactors (`redactSensitiveText`, `redactSensitiveFieldValue`). The model-visible toolResult-content routing (`redactModelVisibleToolPayloadText*` / `redactModelVisibleSensitiveFieldValue*`) does not exist at that tag.
- 9.2 applies the model-visible battery to `role === "toolResult" → content` at storage with defaults on and no configuration.

## Relation to #14586

Different root cause. #14586 is the config-redaction sentinel `__OPENCLAW_REDACTED__` leaking into an outbound API request (restore miss on the native-app path, 2026.2.9). This issue is model-context poisoning via shape/name-based transcript redaction at persist/replay (mask forms `***` and `first6…last4`, 2026.9.2, main gateway path). Same broad family — a redaction layer applied beyond the boundary where its output is safe to consume — but different subsystem, different sentinel/mask, different code path, different versions. Filed separately and cross-referenced.

## Reporter notes

- All credential-like strings in this report are synthetic dummies generated for the reproduction (`sk-bugreport-dummy-<random hex>`); no real keys or fragments appear.
- Reproduction scripts remain on the reporting host under `/tmp/bughunt/` (`repro1.mjs`, `repro2-structured.mjs`).

### OpenClaw version

2026.9.2 (3928bad)

### Operating system

Ubuntu 24.04.4 LTS (headless LXC container)

### Install method

npm global (/usr/lib/node_modules/openclaw) with systemd user service

### Model

zai/glm-5.3 (repro); issue is provider-independent — redaction happens at transcript persist, before the model call path

### Provider / routing chain

zai/glm-5.3 (repro); issue is provider-independent — redaction happens at transcript persist, before the model call path

_No response_

### Logs

```shell
receipts embedded in repro steps above
```

_No response_

### Impact and severity

_No response_

### Additional information

_No response_

Finn763:
## What Problem This Solves

Default-on model-visible transcript redaction persists masks (`***`, `first6…last4`)

into the transcript store, and every replay path (restart, resume, reconcile,

compaction) feeds those masks back to the model as if they were real values. The model

then copies masks into new tool calls, regenerated scripts, and replies — and it cannot

comply with instructions to stop, because the ground truth is gone from its context.

The first revision of this patch tried to recognise masks from their shape at replay

time (an ellipsis regex plus a `***` regex). Review proved that is unsound: the stored

bytes cannot distinguish redaction output from ordinary text, so it rewrote prose

(`the file is here…world of pain`), markdown rules (`***`), and consumed the surrounding

syntax of embedded masks (`value=sk-bug…9f3a` → the placeholder alone, losing `value=`).

`…` alone appears in 38 places in this repository for plain truncation, so no shape rule

can be correct.

This revision replaces guessing with **explicit provenance**, exactly as the review

recommended.

## Design

**1. Persistence marks every mask it produces.** New shared marker grammar in

`@openclaw/normalization-core/redaction-provenance`; `src/logging/redact.ts` gains a

scoped opt-in (`withRedactionProvenance`) that wraps each freshly produced mask:

```


sk-abc…0xyz      →  U+001F⟦openclaw:redacted:1⟧sk-abc…0xyzU+001F⟦/openclaw:redacted:1⟧


(the U+001F byte is written literally; it is not whitespace, so every redaction


path treats the mark as one contiguous token)


```

`src/agents/transcript-redact.ts` (the transcript write-side) opts in; logs, UI, and

every other caller keep the bare mask, so the default output of `redact.ts` is

byte-identical (`src/logging/redact.test.ts` unchanged, 214 tests still pass). The hint

bytes (`sk-abc…0xyz`) are preserved, so the mask stays diagnosable and replay can hand

the model a visible instruction instead of a fake value.

**2. Replay rewrites only marked spans.** `packages/agent-core/src/harness/session/session.ts`

now replaces each `U+001F⟦openclaw:redacted:1⟧…U+001F⟦/openclaw:redacted:1⟧` span with

`[redacted: re-derive this value, do not reuse]` and leaves every other byte alone. There

is no ellipsis heuristic, no `***` heuristic, and no key denylist — surrounding text is

preserved by construction. An unterminated opener is left verbatim rather than guessed at.

**3. Nested payload lookalikes are processed (P2).** The previous revision skipped whole

subtrees by key name at any depth, so a lookalike under `arguments.id` was presented as a

usable argument even though `src/agents/transcript-redact.ts:561` records that persistence

still redacts those positions. Replay no longer needs a key list at all: persistence only

marks masks it produced in the same pass, and it never redacts the correlation fields that

attribute redaction relies on (`idempotencyKey`; root `toolCallId` for tool results; `id`

on `type: "toolCall"` content blocks; the tool-detail ids), so a marked span cannot land

where the exact bytes must stay identical. Replay therefore processes unrelated nested

values and leaves correlation fields byte-identical.

**4. Unmarked history is never touched.** A transcript written before this change carries

no provenance, and its bytes cannot be reinterpreted without guessing — the failure mode

review rejected. Such values now replay exactly as persisted. Changing that (a visible

relookup/recovery command, or a migration that marks historical masks) is a compatibility

decision the maintainers own; this PR deliberately does not pick one.

**Compatibility.** Stored values remain JSON strings, so older readers keep parsing both

old and new transcripts; they simply see the mask with an explicit boundary. The marker is

greppable (`⟦openclaw:redacted⟧`) for both audit and future migration work.

## Evidence

Real isolated run (`OPENCLAW_STATE_DIR` pointing at a fresh temp state dir), driving the

production persistence and resume paths rather than fixtures — the harness appended a

tool-call message with a synthetic secret through `SessionManager.appendMessageToTranscript`,

then read the durable transcript back with `loadTranscriptEvents` and rebuilt model context

with `SessionManager.readSessionContext` (the same `iterateSessionContextMessages`

projection resume and compaction use):

```


state dir: C:\Users\A~1\AppData\Local\Temp\openclaw-resume-evidence-<rand>





persisted bytes (as the store returns them):


  "arguments":{"apiKey":"⟦openclaw:redacted⟧plains…e123⟦/openclaw:redacted⟧",


               "command":"OPENAI_API_KEY=⟦openclaw:redacted⟧sk-abc…0xyz⟦/openclaw:redacted⟧ ok"}





rebuilt model context after resume:


  "arguments":{"apiKey":"[redacted: re-derive this value, do not reuse]",


               "command":"OPENAI_API_KEY=[redacted: re-derive this value, do not reuse] ok"}





legacy transcript (no provenance, same harness):


  persisted: "arguments":{"apiKey":"***","command":"value=sk-bug…9f3a","note":"***"}


  resumed:   "arguments":{"apiKey":"***","command":"value=sk-bug…9f3a","note":"***"}


```

The secret value never appears in either the stored bytes or the rebuilt context; the

assignment prefix and trailing words around embedded masks survive; and the unmarked

legacy row is returned byte-identical, which is the policy that review asked for.

Boundary: this is the real store + real resume read on disk, not a live model turn — no

provider call was made, so it does not attempt to show a model declining to reuse the

placeholder. It does establish that the resolved model context no longer contains a

copyable secret or a copyable mask.

<details>

<summary>Harness that produced the fragments above (reproduce with <code>vitest run &lt;file&gt;</code>)</summary>

```ts


const dir = fs.mkdtempSync(path.join(os.tmpdir(), "openclaw-resume-evidence-"));


process.env.OPENCLAW_STATE_DIR = dir;


const storePath = path.join(dir, "agents", "main", "sessions", "sessions.json");


const target = { agentId: "main", sessionId: "ev1", sessionKey: "agent:main:ev1", storePath };


ensureSessionEntrySync(target, { sessionId: "ev1", updatedAt: Date.now() });





// real persist path (redacts + writes the durable transcript)


SessionManager.appendMessageToTranscript(target, assistantToolCall(SECRET), { config: {} });


const persisted = JSON.stringify(await loadTranscriptEvents(target));





// real resume read: same projection resume/compaction consume


const resumed = JSON.stringify(


  SessionManager.readSessionContext(target, (messages) => [...messages]),


);


```

</details>

Regression tests (`npx vitest run --maxWorkers=1 --no-file-parallelism`, single run,

9 files / 404 tests, exit 0) — the full log is available on request:

| Suite | Result |

|---|---|

| `packages/normalization-core/src/redaction-provenance.test.ts` (new) | 7 / 7 |

| `packages/normalization-core/src/package-exports.test.ts` | 1 / 1 |

| `packages/agent-core/.../session-replay-redaction.test.ts` | 11 / 11 |

| `src/agents/transcript-redact-provenance.test.ts` (new) | 5 / 5 |

| `src/agents/transcript-redact.test.ts` | 86 / 86 |

| `src/config/sessions/transcript-append-redact.test.ts` | 15 / 15 |

| `src/logging/redact.test.ts` | 220 / 220 |

| `src/agents/session-tool-result-guard.transcript-events.test.ts` | 29 / 29 |

| `src/agents/exec-auto-review-transcript.test.ts` | 30 / 30 |

The three review counterexamples are pinned as tests: ordinary ellipsis prose, `***` as a

markdown rule/emphasis, and `value=sk-bug…9f3a` (prefix preserved). Provenance is pinned by

running the real persist-side redaction and then the real replay projection against each

other.

RED verification: reverting the replay half fails 7 of the 11 session-replay tests;

reverting the persistence half fails 3 of the 5 persist→replay tests.

`tsgo -p tsconfig.core.json` reports no error in any changed file. The only diagnostic is

the pre-existing `src/infra/net/node-proxy-agent.ts` missing-dist error for

`@openclaw/proxyline`, which also reproduces with this branch stashed.

`src/agents/embedded-agent-runner/transcript-rewrite.test.ts` fails on this machine with

`EPERM ... rm` while deleting a Windows temp dir; it fails identically with this branch

stashed.

Closes #142821

**Scope note (write-side coverage).** The transcript writer is not the only path that persists

tool-result values: `installSessionToolResultGuard` redacts tool-result details on its own way

to the store. Those masks now opt in as well (`redactPersistedDetailValue` /

`redactPersistedDetailString` wrap their redaction in `withRedactionProvenance`), so a short

detail value persists as `⟦openclaw:redacted⟧***⟦/openclaw:redacted⟧` instead of a bare `***`

that replay would have handed back as a usable value. Two deliberate boundaries: **values** are

marked (they are what a model can copy into later calls); **keys** stay bare, because replay must

never rewrite an identifier. Longer values keep their hint bytes (`sk-abc…0xyz`) — that is

existing persistence behaviour, now visibly bounded.

---

**Marker encoding and upgrade behaviour (revision 6, head `b37d754f268`).** Replay recognizes only the

versioned, escape-prefixed mark (`^_⟦openclaw:redacted:1⟧ … ^_⟦/openclaw:redacted:1⟧`), and

persistence escapes every escape byte it did not emit as part of a mark, so user text cannot spell one by

itself. Pre-upgrade rows (bare masks) and pre-versioning delimiter text (`⟦openclaw:redacted⟧…`) carry no

escape byte at all and stay literal history: nothing is migrated, rewritten, or reinterpreted.

Append identity survives the encoding change — `messagesMatchForIdempotentReplay` compares the canonical

form, so a retry that now produces marked masks still deduplicates against a row written with bare masks,

while conflicting payloads still conflict. Marker examples in the notes above are from the earlier

revision and are superseded by this encoding.

## Review round 5 — repairs at `b37d754f268`

**1. Input can no longer exempt itself from masking (P1).** `maskToken` used to return early whenever a value *contained* an opener

(`hasRedactionProvenance`), so `redactSensitiveFieldValueWithConfig("password", "hunter2⟦openclaw:redacted⟧", {})` persisted the secret

unmasked — reachable from structured sensitive fields and registered-secret replacement, in and outside provenance mode. The pass-through

condition is now `isRedactionProvenanceMask`: only a value that is *nothing but* one whole mask this encoder produced (delimiters at both

ends, mask-shaped body, no other byte) survives; any value carrying reserved marker bytes is masked whole, so no marker byte can survive

into a hint and forge a mark. Independently probed on the pushed head: a legacy-marker spoof, a versioned-marker spoof, and a full literal

pair all come back as `***`, while a genuine mark still passes through unchanged.

**2. The grammar is versioned and unambiguous (P2).** The mark is now `U+001F⟦openclaw:redacted:1⟧ … U+001F⟦/openclaw:redacted:1⟧` — a

version suffix plus a leading non-printable byte, which the tokenizer/trim/serializer never splits off. Persistence escapes every reserved

byte it did not emit as part of a mark, and replay restores those bytes, so literal text round-trips byte-for-byte. Replay recognises only

this form: bare-mask rows and pre-versioning delimiter text stay literal history — not migrated, not rewritten, not re-interpreted.

**3. Append retries stay idempotent across the format change (P1).** `messagesMatchForIdempotentReplay` compared raw payloads, so a retry

that produced a marked mask never matched a row persisted before the upgrade with a bare mask, and raised

`TranscriptTurnAdmissionConflictError` instead of returning the existing append. Comparison now runs on the canonical form

(`stripRedactionProvenance`) in all three branches; identical content in either encoding dedupes, genuinely conflicting payloads still

conflict.

Each of the three has a regression test that fails on the previous head (`5 failed | 7 passed`, `4 failed | 1 passed`, `2 failed | 5 passed`,

`1 failed | 1 passed` before the fix). No assertion was loosened to make a finding pass; the few assertions that compared serialized JSON

against marker constants now compare the serialized form the encoder actually emits.

**Residual boundary, stated plainly:** a value that is byte-for-byte one complete mask is indistinguishable from generated output — but that

requires the exact bytes the encoder emits, not delimiter text a user can type. The mark stays inline rather than moving to out-of-band

metadata because replay must preserve the bytes around the mask (assignment prefixes, quoting) verbatim.

---

# ocproof-143937 — refreshed boundary evidence for openclaw PR #143937 (#142821)

Run context (nothing was pushed; the PR was not modified):

- Worktree `D:\code\wt-oc-142821`, branch `fix/oc-142821-descr`, HEAD `a97624eecca503c0d4842cbd75c705e11ad66793`

  — identical to the PR's `headRefOid` at run time.

- PR body hash: `sha256(body) = e73775189d08b93b1f8e94a36b8d0c9f725bb4cfd44adc48c92db814abf1b7d7` (12839 chars, fetched with

  `gh pr view 143937 -R openclaw/openclaw --json body` at run time).

- Why this refresh: ClawSweeper asked to refresh the cited boundary evidence ("Windows trace through real transcript append,

  durable read, and session-context reconstruction") for the current versioned encoding, literal markers and escape runs,

  registered-secret masking, and legacy append retries. Lanes 2 and 3 below reproduce the exact bytes the body cites

  (`⟦openclaw:redacted⟧plains…e123⟦/openclaw:redacted⟧`) and show they are the superseded unversioned codec.

## Harness — same layer as the cited evidence, real vs synthetic

Command (from the worktree root; vitest 5.0.0):

```

node scripts/run-vitest.mjs run --config test/vitest/vitest.runtime-config.config.ts src/config/sessions/zz-ocproof-143937.probe.test.ts

```

Probe file: `src/config/sessions/zz-ocproof-143937.probe.test.ts`, sha256 `879cd350f0097fdfafc5a0f47b7973e23a9772d4896174af2985ce71f0f6c443` (13675 bytes).

Archived copy: `%LOCALAPPDATA%\Temp\ocproof-143937.probe.test.ts` (same hash); removed from the worktree after the runs.

- **Real (not fixture):** the production persist path `SessionManager.appendMessageToTranscript`; the durable read

  `loadTranscriptEvents` against the real SQLite transcript store in an isolated state dir (`withOpenClawTestState`,

  `OPENCLAW_STATE_DIR`); the production resume projection `SessionManager.readSessionContext` →

  `iterateSessionContextMessages` (the projection resume/compaction use); `runOpenClawAgentWriteTransaction` +

  `appendTranscriptMessageInTransaction` for the pre-upgrade row; exact-value secret registry

  `registerSecretValueForRedaction`.

- **Synthetic (fixture):** the message payloads (plain fixture objects cast to the message union) and the secret values

  (fake, e.g. `sk-abcTESTONLY0xyz`). No provider/model call was made — there is no live-model turn, exactly the boundary

  the PR body declared.

Raw logs (verifiable copies): `%LOCALAPPDATA%\Temp\ocproof-143937.after.raw.log`

(sha256 `14a3764f740cec867171cb0d9a5932dc4d80139f5ba77ec96a4ff49fa3032e3e`), `…before.raw.log`

(`ac9bda169a2642f8e462613fb421863f564a226384091d7d67bfe73e328116fb`), `….prefix-head.raw.log`

(`8fc6587b556278bed04ec10a2be1600b1a0dbe4f3b2f11864055dc31566f72f2`).

## Lane 1 — after: HEAD `a97624eecca` (4/4 pass, exit 0)

```

OCPROOF A.persisted-store-bytes {"apiKey":"\^_⟦openclaw:redacted:1⟧plains…e123\^_⟦/openclaw:redacted:1⟧","command":"OPENAI_API_KEY=\^_⟦openclaw:redacted:1⟧sk-abc…0xyz\^_⟦/openclaw:redacted:1⟧ openclaw health"}

OCPROOF A.resumed-model-context [{"role":"assistant","api":"openai-responses","provider":"test-provider","model":"test-model","stopReason":"toolUse","timestamp":0,"usage":{"input":1,"output":1,"cacheRead":0,"cacheWrite":0,"totalTokens":2,"cost":{"input":0,"cache":0,"cacheRead":0,"cacheWrite":0,"total":0}},"content":[{"type":"toolCall","id":"call_1","name":"shell","arguments":{"apiKey":"[redacted: re-derive this value, do not reuse]","command":"OPENAI_API_KEY=[redacted: re-derive this value, do not reuse] openclaw health"}}]}]

OCPROOF sha256(A.persisted-store-bytes) 80239fa145785b70c60896137fb9cc10cef00d646e281b0b42bf544a6a2425ed

OCPROOF sha256(A.resumed-model-context) 22809269503a74e5f0645997e8061b88844000151cbc0b931d907c7081285acb



OCPROOF B.v1-literals-input "the doc quotes \^_⟦openclaw:redacted:1⟧example\^_⟦/openclaw:redacted:1⟧ verbatim"

OCPROOF B.v1-literals-persisted "the doc quotes \^_\^_⟦openclaw:redacted:1⟧example\^_\^_⟦/openclaw:redacted:1⟧ verbatim"

OCPROOF B.v0-literals-persisted "legacy delimiter text ⟦openclaw:redacted⟧example⟦/openclaw:redacted⟧ stays literal"

OCPROOF B.resumed-model-context [{"role":"user","content":"the doc quotes \^_⟦openclaw:redacted:1⟧example\^_⟦/openclaw:redacted:1⟧ verbatim","timestamp":0},{"role":"user","content":"legacy delimiter text ⟦openclaw:redacted⟧example⟦/openclaw:redacted⟧ stays literal","timestamp":0}]

OCPROOF sha256(B.v1-literals-persisted) efac2114b1437a0e81816fb01a28a1aa1492d43f82343c92a34b9f728d3dcaae

OCPROOF sha256(B.v0-literals-persisted) 73873dcd6a2fbb025001c05f8421d304d7899981d2b174f91277326d749589dc

OCPROOF sha256(B.resumed-model-context) 10692ed57e5d8b6c948ee668d2817231abd6e960fb3008eceb450e63b63f877d



OCPROOF C.persisted-store-bytes {"registered":"\^_⟦openclaw:redacted:1⟧ocproo…4f21\^_⟦/openclaw:redacted:1⟧","password":"\^_⟦openclaw:redacted:1⟧***\^_⟦/openclaw:redacted:1⟧"}

OCPROOF C.resumed-model-context [{"role":"assistant","api":"openai-responses","provider":"test-provider","model":"test-model","stopReason":"toolUse","timestamp":0,"usage":{"input":1,"output":1,"cacheRead":0,"cacheWrite":0,"totalTokens":2,"cost":{"input":0,"cache":0,"cacheRead":0,"cacheWrite":0,"total":0}},"content":[{"type":"toolCall","id":"call_1","name":"shell","arguments":{"registered":"[redacted: re-derive this value, do not reuse]","password":"[redacted: re-derive this value, do not reuse]"}}]}]

OCPROOF sha256(C.persisted-store-bytes) 6e6c497c3b7f62d4e82a885ee2bb3b9801a346bd82f32675a676bd15df594bb0

OCPROOF sha256(C.resumed-model-context) f35cf78e91a50e8cc204261ddfd87f360bba20ec9fbd9ee49df4227393b28586



OCPROOF D.legacy-row-resumed [{"role":"assistant","content":"Here is your key: sk-pro…wwww","idempotencyKey":"ocproof-legacy-retry"}]

OCPROOF D.retry-appended false

OCPROOF D.retry-returned-message {"role":"assistant","content":"Here is your key: sk-pro…wwww","idempotencyKey":"ocproof-legacy-retry"}

OCPROOF D.conflicting-retry-error "TranscriptTurnAdmissionConflictError: Transcript idempotency key \"ocproof-legacy-retry\" conflicts with the admitted message."

OCPROOF sha256(D.legacy-row-resumed) 76edc75198a54fa1f32434575a0fcf9bac364e187c22fc0a36b33005d95d6d3d

OCPROOF sha256(D.retry-returned-message) b736995bf4af39f71e9b5fa452a384ea61dea4d8cf2e7de8362d10124e04ae24

```

Notes on lane 1 (checked by the probe's assertions): the two masks are `first6…last4` hints; the escape byte `U+001F` sits

literally on both sides of every mark (shown as `\^_` because the store returns JSON strings); assignment prefixes

(`OPENAI_API_KEY=`) and trailing words survive; the resumed context carries the re-derive placeholder and no marker byte,

and no secret substring (`TESTONLY`, `hunter2`, the registered value) survives anywhere. Literal text that spells the

current encoding gains exactly one escaped copy of each escape byte (B, 76 → 78 chars) and replays byte-identical; the

pre-versioning delimiter text is left alone. A registered exact value with no surrounding syntax (`ocproof-registered-secret-4f21`)

persists as a marked `ocproo…4f21` mask; a sensitive field whose own bytes carry a marker opener (`hunter2\^_⟦openclaw:redacted:1⟧`)

persists as a marked `***` — the input cannot exempt itself from masking. The pre-upgrade row (bare `sk-pro…wwww`, idempotency

key `ocproof-legacy-retry`) replays byte-identical (no placeholder, not rewritten) and a same-turn retry that now encodes

marked masks is deduplicated (`appended:false`, returns the stored bare bytes) while a conflicting payload still throws

`TranscriptTurnAdmissionConflictError`.

## Lane 2 — before: PR parent `4ed6af1ea45` sources (old shipped behavior)

Revert command (restored afterwards with `git checkout HEAD -- …`):

```

git checkout 4ed6af1ea45 -- packages/agent-core/src/harness/session/session.ts packages/normalization-core/package.json \

  src/agents/session-tool-result-guard.ts src/agents/transcript-redact.ts \

  src/config/sessions/session-accessor.sqlite-transcript-message-append.ts src/logging/redact.ts tsconfig.json

# plus: PR-added sources moved aside (packages/normalization-core/src/redaction-provenance.ts,

# src/agents/transcript-redact-values.ts) — the parent tree has neither file

```

```

OCPROOF A.persisted-store-bytes {"apiKey":"plains…e123","command":"OPENAI_API_KEY=sk-abc…0xyz openclaw health"}

OCPROOF A.resumed-model-context [{"role":"assistant","api":"openai-responses","provider":"test-provider","model":"test-model","stopReason":"toolUse","timestamp":0,"usage":{"input":1,"output":1,"cacheRead":0,"cacheWrite":0,"totalTokens":2,"cost":{"input":0,"cache":0,"cacheRead":0,"cacheWrite":0,"total":0}},"content":[{"type":"toolCall","id":"call_1","name":"shell","arguments":{"apiKey":"plains…e123","command":"OPENAI_API_KEY=sk-abc…0xyz openclaw health"}}]}]

OCPROOF sha256(A.persisted-store-bytes) 01adf5fd6e01cf599a4eccac600854e1e9522d30761029a1566c1c6823bef69a

OCPROOF sha256(A.resumed-model-context) eb8105394fdda8364cf1ad85fa095ce7108b04272e9271d021da1356dc11c870



OCPROOF B.v1-literals-persisted "the doc quotes \^_⟦openclaw:redacted:1⟧example\^_⟦/openclaw:redacted:1⟧ verbatim"

OCPROOF sha256(B.v1-literals-persisted) 04b0ee4c775451243fa197073b0d2c58e45e6520f3d2f8d2267aa2f3e61bfc19

OCPROOF sha256(B.resumed-model-context) 10692ed57e5d8b6c948ee668d2817231abd6e960fb3008eceb450e63b63f877d



OCPROOF C.persisted-store-bytes {"registered":"ocproo…4f21","password":"hunter…d:1⟧"}

OCPROOF sha256(C.persisted-store-bytes) 50ff114a2031f29a486bc323169f983a5aa795a8173f41a1494e4934ab209b16

OCPROOF sha256(C.resumed-model-context) 12ee075fe4d52ddd5b188a3ecdfae5b5d7bb987c4059bb0ffcf7f3f9c65fcdad



OCPROOF D.legacy-row-resumed [{"role":"assistant","content":"Here is your key: sk-pro…wwww","idempotencyKey":"ocproof-legacy-retry"}]

OCPROOF D.retry-appended false

OCPROOF D.retry-returned-message {"role":"assistant","content":"Here is your key: sk-pro…wwww","idempotencyKey":"ocproof-legacy-retry"}

OCPROOF sha256(D.legacy-row-resumed) 76edc75198a54fa1f32434575a0fcf9bac364e187c22fc0a36b33005d95d6d3d

OCPROOF sha256(D.retry-returned-message) b736995bf4af39f71e9b5fa452a384ea61dea4d8cf2e7de8362d10124e04ae24

```

Lane-2 deltas: no marker byte anywhere, bare masks are handed back to the model on resume, literal mark bytes are not

escaped, and a marker-bearing sensitive field leaks delimiter characters into a hint mask (`hunter…d:1⟧`) instead of being

masked whole. Because no marked masks exist in this lane, scenario D can only compare bare vs bare — its observable result

is the same as HEAD. 3 of 4 probe tests fail here; only D passes, by construction.

## Lane 3 — previous PR head `e9e74816083` (pre-fix version of this PR)

Revert command: `git checkout a97624eecca^ -- packages/normalization-core/src/redaction-provenance.ts src/agents/session-tool-result-guard.ts src/agents/transcript-redact-values.ts src/config/sessions/session-accessor.sqlite-transcript-message-append.ts src/logging/redact.ts` (restored afterwards).

```

OCPROOF A.persisted-store-bytes {"apiKey":"⟦openclaw:redacted⟧plains…e123⟦/openclaw:redacted⟧","command":"OPENAI_API_KEY=⟦openclaw:redacted⟧sk-abc…0xyz⟦/openclaw:redacted⟧ openclaw health"}

OCPROOF sha256(A.persisted-store-bytes) 65022e615719d4a463767411fea3e4958a1bfb85514ec584df4be5e0461e0e2b

OCPROOF sha256(A.resumed-model-context) 22809269503a74e5f0645997e8061b88844000151cbc0b931d907c7081285acb



OCPROOF B.resumed-model-context [{"role":"user","content":"the doc quotes \^_⟦openclaw:redacted:1⟧example\^_⟦/openclaw:redacted:1⟧ verbatim","timestamp":0},{"role":"user","content":"legacy delimiter text [redacted: re-derive this value, do not reuse] stays literal","timestamp":0}]

OCPROOF sha256(B.resumed-model-context) 71a51681f7adc0387ae039b4698c7ea6e06c952297d7871d703b2314a4d8096d



OCPROOF C.persisted-store-bytes {"registered":"⟦openclaw:redacted⟧ocproo…4f21⟦/openclaw:redacted⟧","password":"⟦openclaw:redacted⟧hunter…d:1⟧⟦/openclaw:redacted⟧"}

OCPROOF sha256(C.persisted-store-bytes) 9b831eaaec0face73e277ecd73fc3cef9dcec38de7f6060cf42819ab3e47781e



FAIL  … > legacy rows stay literal and legacy append retries dedupe across the encoding change

TranscriptTurnAdmissionConflictError: Transcript idempotency key "ocproof-legacy-retry" conflicts with the admitted message.

```

Lane-3 deltas, all fixed at HEAD: (a) persisted masks are the unversioned delimiter text — byte-for-byte what the PR body

quotes as evidence; (b) replay consumed the pre-versioning delimiter text from literal history

(`⟦openclaw:redacted⟧example⟦/openclaw:redacted⟧` → placeholder), which HEAD leaves verbatim; (c) a marker-bearing

sensitive field produced a partial-hint mask instead of a whole `***`; (d) the legacy retry **threw** a

`TranscriptTurnAdmissionConflictError` on the mixed case (stored bare row + retry that now encodes marked masks), which

HEAD deduplicates — the legacy-retry fix is only observable here, not in lane 2.

## Difference vs the superseded codec, in one line

The old output was the unversioned, escape-less `⟦openclaw:redacted⟧mask⟦/openclaw:redacted⟧` (lane 3, byte-identical to

the body's citation) — the current output is `U+001F⟦openclaw:redacted:1⟧maskU+001F⟦/openclaw:redacted:1⟧` (lane 1), with

literal escape bytes doubled in storage and restored on replay.

## Honest limits

- Synthetic messages and fake secrets; no live provider/model run (not required by the review, and not attempted).

- `withOpenClawTestState` isolates the state dir, but the store is the real SQLite transcript store, and the write/read/resume

  functions are the production ones (same functions the earlier evidence used).

- The worktree was restored to HEAD after lanes 2 and 3 (`git status --short` clean except the untracked probe, which was

  then removed); nothing was pushed and the PR was not edited.

### Re-ran by the author

The probe was re-run by hand against this head (`a97624eecca`): `Test Files 1 passed (1) / Tests 4 passed (4)`, emitting the same OCPROOF lanes — versioned `\^_⟦openclaw:redacted:1⟧…⟦/openclaw:redacted:1⟧` markers in the persisted store, placeholder on resume, registered-secret masking (`***` inside the marker), and the legacy append retry deduping (`D.retry-appended false`). The probe file was removed from the tree afterwards; `git status --porcelain` clean.

---

### Revision `8230e9be72f` - findings addressed (code, not just proof)

1. **Provenance now lives in the existing owner.** `src/agents/transcript-redact-text.ts` is the single implementation; `redactTranscriptText` / `redactTranscriptStructuredFieldValue` signatures unchanged, and both `transcript-redact.ts` call sites route through the new `encodePersistedTranscriptText()` wrapper. The unused competing module (`transcript-redact-values.ts`) is deleted - zero references left in the tree.

2. **No-marker branch keeps sanitized output** (`encodePersistedDetailText()` in `session-tool-result-guard.ts`): truncation / partial-secret omission / custom-pattern results persist in sanitized form; only a byte-identical sanitizer result falls back to the escaped raw. An unterminated private-key fragment now persists only as "partial secret span omitted".

3. **Literal marks stay distinguishable on replay**: when redaction produced no fresh mark, the encoder returns the escaped form, so a user-authored `START + *** + END` literal never reads back as generated provenance. New tests cover the no-fresh-mark literal case (the previous helper test missed it by adding a genuine mark).

Knock-on changes the encoding requires (covered by tests): canonical compare for the delivery-mirror dedupe (`config/sessions/transcript.ts:801`), decode-before-redact on CLI history imports (`gateway/cli-session-history.claude.ts`), canonicalized compare keys in the Claude history merge.

**Red -> green (owner re-run):** sources reverted with new tests kept -> `transcript-redact-provenance` 4 failed, `transcript-redact` 12 failed, persist-hook 9, guard 1, append-redact 2, embedded-runner.guard 2 (relevant set all red). After -> six focused files **190 passed** across 3 shards, exit 0. `tsgo` core exit 0, `oxfmt --check` clean, `oxlint` clean on changed files.

**Known semantic trade-off (pinned in tests + commit message):** re-sanitizing already-encoded bytes is not a byte-wise fixed point (a hint-shaped body itself masks to `***`). "Re-redacting stored rows preserves hints" and finding 3's literal-mark protection cannot both hold for the same byte sequence; this revision follows the review and prioritizes finding 3.

### Revision `e014c5e8efe` — all five ClawSweeper findings fixed (code, not just proof)

Review Revision 13 listed five findings (plus a security duplicate of the first). All five are fixed in code now, each pinned by a test whose failure was re-produced with the source reverted:

- **[P1] Match registered secrets before changing their bytes** (`transcript-redact-text.ts:53-56`): the write path escapes raw bytes before exact-value matching, so a secret containing the encoding discriminator (U+001F) could be persisted unmasked. `src/logging/secret-redaction-registry.ts` now also registers the transcript-escaped surface form of every registered secret (before the raw value, so eviction drops derived forms first) — the credential is matched and masked on the exact bytes the matcher sees, so replay cannot restore it, and the recorded security repro ("encoding can bypass registered-secret matching") no longer reproduces.
- **[P1] Preserve producer provenance across guarded append passes** (`session-tool-result-guard.ts:206-212`): both persistence encoders now detect their own prepared output via the storage mark and reuse that body instead of re-escaping it as raw text, so a second detail pass keeps the marker alive while policy revalidation still re-runs.
- **[P2] Decode escaped literals even when no secret was masked** (`redaction-provenance.ts:247-250`): an explicit string-level storage mark (`REDACTION_PROVENANCE_STORAGE_MARK`, written only for strings that carry a reserved byte) — `resolveStoredRedactionProvenance` decodes a marked string even when it holds no mask; unmarked legacy bytes (no marker, or a lone escape run in raw history) stay byte-identical. Round trip and idempotence pinned.
- **[P2] Canonicalize the re-redacted legacy mirror candidate** (`transcript.ts:828-833`): both sides of the mirror comparison now go through `canonicalStoredText`/`stripRedactionProvenance`, so a fresh marker added while re-redacting a legacy row cannot make an equivalent delivery append a second mirror.
- **[P2] Keep raw Claude history out of the stored-text decoder** (`cli-session-history.claude.ts:117-119`): the importer decodes only strings that carry the storage mark, sending raw external rows straight through the raw-input writer; the merge's drift-note view decodes stored bytes before matching.

**Evidence (owner re-verified):** with the fixed sources reverted one at a time, each pin fails with its own assertion (registry: the discriminator-bearing secret survives to storage; guard second pass loses the marker; mask-free escaped write stays doubled; second mirror appended; literal marker collapses). Post-fix: 6 shards / **687 tests, exit 0**; my own re-run of 4 of these files (167 tests, exit 0) plus a fresh red→green of the security pin (1 failed | 9 passed → 10 passed). `oxfmt --check` clean on all 16 changed files; `oxlint` clean; `tsgo -p tsconfig.core.json` exit 0.

**One honest note for the storage owner:** finding 3 is a durable-format change (a minimal per-string mark). A user-typed complete mark is byte-identical to a produced one, so findings 2 and 3 cannot both be settled by content sniffing alone. Mask-free writes now decode, unencoded bytes stay byte-identical, and repeated passes are a fixed point — but storage acceptance / older-reader behavior remains the maintainer question the review flags under "Decision needed".


Finn763:
**CI note — one own fix, one main-side failure**

**Fixed (this PR, head 870f935cf):** `checks-node-compact-large-12` failed on `src/agents/embedded-agent-runner.guard.test.ts` (2 tests) because the guard now provenance-marks persisted masks and this file still asserted bare `peter@d***.io`. Actual persisted form keeps hint bytes outside the markers (`peter@d` + marked `***` + `.io`). Aligned the 5 assertions with `markRedactionProvenance`, same pattern as the sibling guard tests; file passes 24/24 locally plus 178/178 across the related redaction suites.

**Not this PR (`checks-node-compact-small-39`):** `test/official-channel-catalog.test.ts > keeps the generated channel docs index and navigation synchronized` fails with `expected [ 'channels/qa-channel' ] to deeply equal []`. This PR touches only redaction files (no `docs/docs.json`, `docs/channels/index.md`, catalog builder/seed, or extension manifests). Same main-side docs/catalog mismatch as before: a main docs commit added `channels/qa-channel` to nav while the catalog deliberately excludes it. No code change appropriate here.

Finn763:
Repaired all three repeated findings at `b37d754f268` (rebased onto current main). Each fix has a regression test that fails on the previous head — no assertion was relaxed to make a finding pass.

## [P1] Marker bytes in input no longer exempt a value from masking — `src/logging/redact.ts`

Root cause: `maskToken` returned early whenever the token *contained* an opener (`hasRedactionProvenance`), so `redactSensitiveFieldValueWithConfig("password", "hunter2⟦openclaw:redacted⟧", {})` took the "already redacted" path and the value was persisted as-is. Structured sensitive fields and registered-secret replacement both call `maskToken`, inside and outside provenance mode, which is why the bypass reached beyond transcript persistence.

Fix: pass-through now requires the *whole* value to be exactly one mark this encoder produced — both delimiters present, nothing outside them, body mask-shaped (`isRedactionProvenanceMask`). Everything else is masked, and a value carrying any reserved marker byte is masked whole to `***` so no marker byte can survive into a diagnostic hint.

Red first (`src/logging/redact-provenance-input.test.ts` against the previous head's sources): `Tests 4 failed | 1 passed (5)` — the marker-bearing password, the legacy-delimiter password, and a complete literal mark all came back unmasked; only "a real marker mask passes through unchanged" passed. Green now: `5 passed (5)`.

## [P2] Literal delimiter text is no longer readable as provenance — `packages/normalization-core/src/redaction-provenance.ts`

Root cause: `⟦openclaw:redacted⟧example⟦/openclaw:redacted⟧` is text a user can type, and the parser replaced every complete delimiter pair, including in rows this encoder never wrote.

Fix:
1. The mark is versioned and opened by a non-printable byte: `\u001F⟦openclaw:redacted:1⟧ … \u001F⟦/openclaw:redacted:1⟧`. The byte is not whitespace (so redaction patterns keep the mark one contiguous token) and is not trimmed or rewritten by storage.
2. Persistence escapes every escape byte it did not emit as part of a mark (`escapeRedactionProvenanceLiterals`, idempotent so repeated passes cannot re-encode), and replay restores those bytes. A mark counts only when the delimiter pair is complete and the body is mask-shaped (`***` or one `prefix…suffix` hint).

Replay recognizes only this form: pre-versioning delimiter text and pre-upgrade rows with bare masks carry no escape byte at all and stay literal history — nothing is migrated or reinterpreted. Stated bound: bytes that are *exactly* one mask-shaped mark (control byte included) remain indistinguishable from encoder output, which requires copying the encoder's own bytes rather than typing delimiter text. In-band marks were kept over out-of-band metadata because replay must preserve surrounding text (an assignment prefix, a quoted value) byte for byte.

Red first (`packages/normalization-core/src/redaction-provenance.test.ts`): `Tests 6 failed | 7 passed (13)`; (`src/agents/transcript-redact-provenance.test.ts`): `Tests 2 failed | 5 passed (7)` — the persisted user text came back from replay as `[redacted: re-derive this value, do not reuse]`. Green now: `13 passed (13)` and `7 passed (7)`; the round-trip tests go through the real transcript writer and the real session-context replay, not fixtures.

## [P1] Append retries stay idempotent across the encoding change — `src/config/sessions/session-accessor.sqlite-transcript-message-append.ts`

Root cause: `messagesMatchForIdempotentReplay` compared raw payloads, so a row persisted before the upgrade with bare masks (`sk-pro…wwww`) never matched the marked candidate for the same secret, and the retry threw instead of returning the existing append.

Fix: the comparison canonicalizes each string (`stripRedactionProvenance`: marked spans collapse to their mask body, escaped literals are restored) before `isDeepStrictEqual`. Both encodings of the same payload now match; genuinely different payloads still conflict.

Red first (`src/config/sessions/transcript-append-redact-idempotency.test.ts` seeds a pre-upgrade row through the store and retries with the same idempotency key): `TranscriptTurnAdmissionConflictError: Transcript idempotency key "assistant-retry-after-upgrade" conflicts with the admitted message.` at `session-accessor.sqlite-transcript-message-append.ts:119`. Green now: `2 passed (2)` — the identical retry deduplicates against the pre-upgrade row, and a conflicting payload under the same key still throws.

## Test command and result

```
node scripts/run-vitest.mjs run packages/normalization-core/src/redaction-provenance.test.ts \
  src/logging/redact-provenance-input.test.ts src/agents/transcript-redact-provenance.test.ts \
  src/config/sessions/transcript-append-redact-idempotency.test.ts --maxWorkers=1 --no-file-parallelism
→ Test Files  4 passed (4) | Tests  27 passed (27) | [test] passed 4 Vitest shards
```

Wider surface (same command plus `transcript-redact`, `embedded-agent-runner.guard`, both tool-result-guard tests, `transcript-append-redact`, `session-replay-redaction`, and the two memory-core redaction tests): 245 tests, 241 pass. Assertions that compared serialized JSON text against marker bytes were updated to the serialized form (`serializeRedactionMarker`) because JSON escapes a control byte — the same bytes, compared where they actually appear.

The remaining 4 failures are not from this change: `src/agents/sessions/agent-session-runtime-projection.test.ts` fails in `cleanupTempDirs` with `EPERM` on this Windows host and fails identically with the previous head's sources checked out.

Finn763:
@clawsweeper re-review

The PR body's boundary evidence has been refreshed for the current (versioned) encoding: real transcript append -> durable read -> session-context reconstruction, showing literal marker/escape handling, registered-secret masking, the marker-input fallback, and legacy append retry dedupe (`appended:false`), with the superseded unversioned run kept alongside for comparison (the parent leaks bare masks; the previous head raises the admission conflict on the legacy retry). Probe re-run by the author: 4/4.

Finn763:
Addressed the 4 outstanding code findings in aa11a934 (rebased onto 0fcaa5c):

1. P1 maskToken bypass (`src/logging/redact.ts`): removed the `isRedactionProvenanceMask` shape-based passthrough entirely — shape never proves provenance, so raw/registered complete marker-shaped values are now masked like any other value. Bare `***` stays idempotent.
2. P2 literal marks (`redaction-provenance.ts`): write paths pre-escape raw input (`escapeRaw` doubles every raw U+001F), so literal complete markers can no longer be mistaken for producer spans; readers fall back to raw bytes when no genuine mark exists.
3. P2 escape pairs: every raw escape byte is encoded on write; legacy EE pairs round-trip byte-identical instead of collapsing.
4. P2 prefilter/scanner split: replaced the negative-lookbehind prefilter with scanner-based `containsGenuineMark` — detection and replacement now use the same parsing rule.

Tests: new `src/logging/redact-review-143937.test.ts` (6 tests) failed 5/6 pre-fix (RED) and passes post-fix; `redaction-provenance` 14/14 + `redact-provenance-input` 5/5 green (vitest, 25/25 total, verified on the pushed head). Broader agent/session suites fail identically on unmodified base (missing import-meta-resolve, sqlite subprocess build) — pre-existing infra, unrelated.

Left for owner/maintainer (cannot resolve from fork): storage-owner acceptance of the inline provenance encoding, older-reader/rollback compatibility contract, and the inline-vs-separate provenance decision.

Finn763:
@clawsweeper re-review - revision 8230e9be72f wires the provenance encoding into the existing transcript writer (single owner, competing module deleted), keeps sanitized output in the no-marker branch, and preserves the encoded/raw distinction on replay; red->green re-run by the owner (190 passed).

Finn763:
@clawsweeper re-review

PR #143954, opened 2026-09-10
fix(models): preserve literal rows in model policy catalogs

steipete:
Related: #130706, #142100. Follows #143872.

## What Problem This Solves

Model-policy catalog construction can drop distinct literal provider/model rows when their display keys coincide. An explicit allowlist or wildcard can therefore lose a catalog row even though the selection policy already permits it. Synthetic rows can also collide during deduplication.

## Why This Change Was Made

Use the existing literal catalog-row key for private deduplication and synthetic entries, and retain all logical identities before filtering. Reuse the shared deduplication helper and one lazy allow-all result builder. Public display/authorization keys, provider equivalence, aliases, and unique case-insensitive lookup keep their existing contracts. The change removes 25 production lines from one owner.

## User Impact

Model browsers and selection catalogs retain distinct allowed rows, including provider-prefixed and slash-containing model IDs. Genuine missing rows remain visible as sparse synthetic entries without copying another row's capabilities.

## Evidence

Eleven intended missing-row regressions failed before the repair. The final focused cohort passed 225 tests across four suites, including partial catalogs, both policy orders, case-insensitive compatibility controls, and Arcee wire/direct identities. Fresh managed P2 review and the full changed-file gate passed. The replay onto main containing #143872 is byte-identical with unchanged reviewed owner/dependency and gate-tooling context; a further 189 selection/visibility tests passed on that main replay.

At exact commit `4eba0cd1084fab2510d1a63c7147dd4b2d13828b`, one canonical `sourcePerformance` build and all 22 compiled catalog/policy scenarios passed. The probe made no network requests and imported no checkout source. All 9,255 captured build files remained unchanged; the process and isolated state were cleaned up. This is synthetic compiled boundary proof, not an external-provider or live Gateway test. The runtime profile excludes UI and SDK declarations.

The first compiled probe stopped before scenarios because its copied build inventory omitted workspace package outputs. Those outputs were admitted only after verifying their canonical build ownership, current source, and generation within the same build interval. The original failure is retained; no product files, stamps, or source-import guard were changed.

### DIFF
--- a/src/agents/model-catalog-visibility.test.ts
+++ b/src/agents/model-catalog-visibility.test.ts
@@ -14,11 +14,12 @@ import { openAIModelCatalogRoutePolicy } from "./openai-model-routes.js";
 
 describe("resolveLogicalVisibleModelCatalog", () => {
   it.each(["all", "configured", "default"] as const)(
-    "keeps case-distinct configured identities in the %s view",
+    "keeps case-distinct and literal provider-prefixed identities in the %s view",
     async (view) => {
       const catalog: ModelCatalogEntry[] = [
         { provider: "fixture", id: "MixedCase", name: "Large", contextWindow: 64_000 },
         { provider: "fixture", id: "mixedcase", name: "Small", contextWindow: 16_000 },
+        { provider: "fixture", id: "fixture/MixedCase", name: "Namespaced", contextWindow: 32_000 },
       ];
       const result = await resolveLogicalVisibleModelCatalog({
         cfg: { agents: { defaults: { modelPolicy: { allow: ["fixture/*"] } } } },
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4eba0cd1084fab2510d1a63c7147dd4b2d13828b](https://github.com/openclaw/openclaw/pull/143954/commits/4eba0cd1084fab2510d1a63c7147dd4b2d13828b)
- Landed commit: [e2c1483bfe4188510ee0713278786895c11ea632](https://github.com/openclaw/openclaw/commit/e2c1483bfe4188510ee0713278786895c11ea632)

PR #143960, opened 2026-09-10
fix(gateway): avoid duplicate user turns after Claude CLI resume

Marvinthebored:
## What Problem This Solves

Fixes an issue where users reloading a Claude CLI conversation see the current user turn twice after the session resumes with changed prompt content.

## Why This Change Was Made

The runner and history importer share the complete context-note format. History first matches the full user text, then tries comparison without one exact generated note. Literal quotes keep their normal match. The existing merge retains identity, timestamp, image and ordering rules, and failed comparison attempts preserve candidates for later turns.

## User Impact

A resumed user turn appears once. Literal note text, distinct native messages, images and unmatched imports remain intact. Stored transcripts, session reuse, authentication and configuration are unchanged.

## Evidence

- Real Gateway and served Control UI used the unmodified official Claude Code 2.1.267 client. The client handled stdio, resumed its own native session, and wrote its own project history. A local Anthropic Messages API supplied synthetic assistant responses; it did not write transcripts. All proof processes shared a private network with no external route, so no vendor credentials or paid inference were used.
- Two real public chat turns, with normal bootstrap completion between them, selected actual system-prompt drift in the same native session. Baseline history returned five rows and the reloaded UI showed the second user turn twice. The candidate returned four rows and showed that turn once.
- Independent official-client acceptance passed all 14 real turns and eight copied-history controls. Literal notes, images, mixed turns, distinct native identities, timestamps, roles, sender metadata and reply directives were preserved. Eighteen browser captures were inspected. The submitted image bytes also match the image block sent by the actual client.
- The runtime proof executed `d2475ee4b0ca468123a634ae676f47ab05228b42`. The final follow-up only renames internal bindings and changes test property-access syntax for lint. Independent source review confirmed that these edits preserve the exercised behavior; the original build identity stays attached to the proof.
- 135 focused history/handler tests passed on the final corrected source. The unchanged CLI session/preparation paths retain 214 passing sibling tests. Regression cases failed on the baseline, original parser and rejected intermediate candidates for their intended reasons.
- Focused non-type guards, formatting, docs checks and the normal native hook passed. Full typechecks and type-aware lint are assigned to hosted CI on the final head.

The API responses are synthetic. This proves the supported client's persistence and the real Gateway/history/UI path, without claiming vendor inference or model image understanding.

AI-assisted.

### DIFF
--- a/docs/gateway/cli-backends.md
+++ b/docs/gateway/cli-backends.md
@@ -247,9 +247,11 @@ register a small wrapper backend plugin.
 - Helper runs with a caller-owned in-memory transcript use that history for hooks, bounded session notes, and fresh-session reseeding, including meaningful history before compaction. Empty memory stays empty even when the run carries another session's storage identity. Context-engine maintenance rewrites that same memory before the helper returns, even when the engine requests background maintenance. Durable transcripts retain their background maintenance path. An explicitly owned native CLI binding can still resume. Resumed turns send the current prompt and bounded session notes without replaying the conversation history.
 
 When prompt content changes, a compatible CLI session can resume with an OpenClaw
-context note before the current user prompt. Chat history ignores that exact note
-when matching imported Claude user turns to local turns, so the same turn appears
-once. Stored transcript text and unmatched imported turns remain intact.
+context note before the current user prompt. Chat history first matches imported
+Claude user turns against the full local text, including any literal quote of the
+note. If that does not match, it ignores one exact context note for comparison, so
+the same turn appears once. Stored transcript text and unmatched imported turns
+remain intact.
 
 ### History account boundaries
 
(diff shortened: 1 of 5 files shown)


obviyus:
Thanks @Marvinthebored and @Peetiegonzalez!

**Proof:** The unmodified official Claude Code 2.1.267 client, real Gateway routes and reloaded Control UI confirm the repair. The client wrote its own native history through a local test API; no vendor inference was used. Independent acceptance passed 14 real turns and eight preservation controls.

- **Bug:** A generated resume note made one user turn appear twice after prompt drift.
- **Fix:** Match literal user text first, then compare without one exact generated note while preserving identity, timestamps, images and ordering.

| Before | After |
| --- | --- |
| Two bubbles for the resumed user turn | One bubble for that turn |
| One intact literal-note bubble | One intact literal-note bubble |
| One image-bearing user turn | One image-bearing user turn |

Final corrected source passed 135 focused history/handler tests and ordinary lint. The unchanged CLI session/preparation paths retain 214 passing sibling tests. The runtime proof keeps its original build identity; the final lint-only changes received fresh source review confirming unchanged behavior.

PR #134496, opened 2026-08-31
fix: show exact fallback models in notices and status

steipete:
Related: #130706. This is a focused display repair; it does not establish the CPU/stall cause in #130324.

## What Problem This Solves

Fixes an issue where `/model` and `/status` could omit the model that answered after fallback, or show its context usage against the selected model's larger window. Fallback notices could also collapse distinct references such as `custom/model` and `custom/custom/model` into the same label.

## Why This Change Was Made

Share the existing bounded terminal-transcript reader across Gateway rows and both commands. It accepts an answering model only when the settled run, visible successful assistant message, selected model, and recorded fallback notice agree. The status renderer consumes the prepared model pair and retains the original session entry for auth and context provenance.

Duplicate selection and prefix-stripping helpers are removed. Production code is **30 lines smaller**; the obsolete assertion baseline shrinks by two lines. No new configuration, persisted fields, schema, or migration is introduced.

## User Impact

Users can distinguish the selected model from the fallback that answered and see the fallback's correct context limit. Repeated fallback notices stay suppressed, recovery clears the notice, and querying status does not change selection or issue a provider request. Previously persisted lossy labels cannot reveal their original identity and are not guessed or reconstructed.

## Evidence

Validated commit `666dddf10b6ff4b5a496edbf948551258c1a60b1` on main `96b5a339f3d9a165bf5c2931b4259a50bf9dafbb`, including the separate config repair #144275.

- **808 focused/parser tests passed.** The affected status and directive suites also passed after final fixture typing and equivalent catalog-guard cleanup.
- The complete changed-file gate, fresh independent P2 review, one normal full build including SDK declarations and UI, and built status-module smoke passed.
- A real isolated compiled Gateway and its public reply dispatcher passed seven stages: fallback activation, `/model`, `/status`, repeat fallback, recovery, and both recovered commands. Five HTTP requests exercised primary 404/fallback 200 and primary recovery; command queries made zero provider requests.
- Actual persisted terminal run IDs matched the successful assistant messages. `/model` showed selected `custom/model` and active `custom/custom/model`; `/status` used the fallback's **128k** context window instead of the primary's **1M**. Recovery returned status to the primary window. Selection and config stayed unchanged.
- Normal Gateway shutdown, worker retirement, and process exit completed. All **14,637 build outputs** stayed unchanged, and **5,456 executed files** were hash-verified.

[...]

PR #143961, opened 2026-09-10
fix(agents): preserve selected models through fallback and auth refresh

steipete:
Related: #130706, #142100, #134496. Builds on #143994.

## What Problem This Solves

Selected model IDs could be parsed again as authored input during fallback planning or credential refresh. A chain such as `entry → middle → final` could advance an already selected model, while display-key deduplication could discard distinct provider/model pairs.

## Why This Change Was Made

Resolve authored input at its owner, return the logical reference from successful model resolution, and carry that reference through embedded setup and auth refresh. Canonical internal producers mark completed selections; actual plugin-hook redirects return to raw resolution. Executable wire IDs remain separate from logical model IDs.

The appended configured primary still needs its permitted runtime-only normalization. Apply that refinement once through the existing captured registry helper, without repeating manifest normalization. Keep disabled-plugin and exact configured-route guards. Existing official-plugin resolver callbacks retain their shipped shape without supplying core-owned logical-reference facts.

This repair adds 19 production lines overall. The candidate-marker cleanup, initial completion/media selection, compaction, and fallback display remain separate cuts. Configuration, persistence, and fallback policy are unchanged.

## User Impact

Fallback execution and credential rotation retain the intended model. Runtime-only aliases in an appended configured primary continue working, and distinct literal candidates remain available.

## Evidence

At `72ec42297ef10cef59520ec2f85d86241f2d4de8`:

- 294 focused tests passed, together with the full changed gate and fresh complete-PR P2 review.
- The previous `41ebecf` head reproduced five intended failures. Two runtime-only appended-primary controls passed on the original working `5f161b6` source. The corrected fifteen-case generation suite passes with the same configuration-bound fixture, including manifest/runtime chains, A/B and empty generations, disabled normalization, exact rows, and deduplication.
- One normal full build passed, including SDK declarations. Nine compiled public-runner scenarios made 24 actual loopback HTTP requests through synthetic credential rejection/rotation, configured fallback, both appended-primary variants, hook redirects, and locked-selection controls. Two additional compiled contracts use the real Arcee plugin and materializer.
- All 11,931 built files remained unchanged. No checkout TypeScript was imported, all owned runtime processes exited, and source stayed clean.

The earlier 31-case auth-rotation E2E passed at `41ebecf`; it is retained as evidence for that revision, not claimed as a rerun on the new head. Current auth regression coverage comes from focused owner tests and the compiled HTTP scenarios. An initial request-scope fixture lacked its normal config-compatibility binding; that failed attempt is preserved, and baseline/control/candidate were rerun with the same corrected fixture.

The HTTP proof uses isolated loopback endpoints and synthetic credentials. Arcee proof is offline. Neither is an authenticated hosted-provider test or a fresh npm installation test. Exact-head CI remains required before landing.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -1719,7 +1719,6 @@ src/agents/embedded-agent-runner/prompt-cache-observability.ts	2
 src/agents/embedded-agent-runner/replay-history.ts	53
 src/agents/embedded-agent-runner/result-fallback-classifier.ts	2
 src/agents/embedded-agent-runner/run-entry.ts	3
-src/agents/embedded-agent-runner/run-orchestrator.ts	1
 src/agents/embedded-agent-runner/run.overflow-compaction.harness.ts	4
 src/agents/embedded-agent-runner/run/abortable.ts	2
 src/agents/embedded-agent-runner/run/attempt-async-tasks.ts	1
--- a/src/agents/command/attempt-execution.ts
+++ b/src/agents/command/attempt-execution.ts
@@ -1341,6 +1341,7 @@ export function runAgentAttempt(params: {
     provider: embeddedAgentProvider,
     model: params.modelOverride,
     modelRoutingProvenance: params.modelRoutingProvenance,
+    requestedRouteResolution: "resolved",
     modelHasVision: params.modelHasVision,
     modelThinkingCapability: params.modelThinkingCapability,
     modelFallbacksOverride: params.modelFallbacksOverride,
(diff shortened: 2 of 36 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [72ec42297ef10cef59520ec2f85d86241f2d4de8](https://github.com/openclaw/openclaw/pull/143961/commits/72ec42297ef10cef59520ec2f85d86241f2d4de8)
- Landed commit: [f49e3bb72e349c992f2632a95fc69ec2f409631c](https://github.com/openclaw/openclaw/commit/f49e3bb72e349c992f2632a95fc69ec2f409631c)

PR #143965, opened 2026-09-10
fix(update): keep progress consistent without redundant ledger reads

steipete:
## What Problem This Solves

Fixes an issue where CLI update progress would reopen the run ledger immediately after each committed step and could announce a newer phase beside an older final report. In the measured 481-case update CLI suite, synchronous SQLite snapshot processes consumed about 85% of elapsed time.

## Why This Change Was Made

Preserve all step and warning-row writes, then pass the existing step writer's final committed record directly to the private CLI display callback, and use one observation for the final phase flush and report. No record is cached for later use. Initial observation, polling, resume, disposal, and deferred-step fallback still read through the existing guarded owner. Shared runner callbacks, ledger writes, database schema, integrity checks, admission, and operational authority are unchanged.

Publish the progress observer only after its initial observation succeeds, so a failed construction cannot leave an observer whose disposer was never returned. Doctor's unbound step display, JSON output and suspended presentation keep their existing behavior.

## User Impact

Update progress avoids redundant synchronous database snapshots and keeps phase output consistent with the final report. No new flags or configuration are needed. Production changes are +20 lines for the private callback type and read/render ownership split; tests add 184 lines. No existing tests, assertions, deadlines or shards are removed or weakened.

## Evidence

- Four regression failures on current main prove the mixed-observation, committed-step reread and observer-publication defects. All 83 focused progress, real-ledger command-run and Doctor tests pass on the final candidate.
- Before integrating main’s new per-step warning rows, full CLI runs passed the same 481 cases in the original order. The candidate reduced synchronous snapshot launches from 5,226 to 4,843 (383 fewer), with all 5,598 ledger writes retained. Whole-run time was 809.82 seconds before and 846.34 seconds after: this pair does **not** establish a full-suite speedup.
- The same nine existing package-spec cases passed an A/B/A comparison in 61.986 / 27.704 / 38.469 seconds. Callback-heavy cases benefit, but baseline variability prevents a precise general CI speedup claim. This comparison also predates the warning-row integration; the final integration is covered by the 83-case run. Same test source, dependencies, Node 24.19.0, pnpm 12.3.4 and source-only root layout; no diagnostic instrumentation or altered cache flags in this comparison.
- Focused command: `node scripts/run-vitest.mjs src/cli/update-cli/progress.test.ts src/cli/update-cli/update-command-run.test.ts src/commands/doctor-update.test.ts`.
- Timing command: `node scripts/run-vitest.mjs src/cli/update-cli.test.ts -t 'resolves package install specs from tags and env overrides' --reporter=verbose`.

These are local CLI and real isolated ledger tests, with external update/service actions mocked by the existing harness; they do not claim a live installed-package update or an eight-minute full CI result. Independent review is clean through P2; `pnpm check:changed` passes, including core/test type checks, formatting, lint, dead exports, import cycles and storage guards. Exact-head CI will be verified before landing.

### DIFF
--- a/src/cli/update-cli/progress.test.ts
+++ b/src/cli/update-cli/progress.test.ts
@@ -81,6 +81,100 @@ describe("update progress", () => {
     expect(lines.join("\n")).toContain("Build type error");
   });
 
+  it("does not leave a phase observer after initial observation fails", () => {
+    const log = vi.spyOn(defaultRuntime, "log").mockImplementation(() => {});
+    vi.mocked(getUpdateRun).mockImplementationOnce(() => {
+      throw new Error("initial ledger read failed");
+    });
+    try {
+      expect(() => createUpdateProgress(true, context)).toThrow("initial ledger read failed");
+      run.phase = "verifying";
+      run.steps = [
+        { step: "requested", status: "completed" },
+        { step: "verifying", status: "in_progress" },
+      ];
+      printResult(result, { run: context });
+      const lines = log.mock.calls.flat();
+      expect(lines.join("\n")).toContain("OpenClaw update in progress: verifying.");
+      expect(lines.filter((line) => typeof line === "string" && line.startsWith("Phase:"))).toEqual(
+        [],
+      );
+    } finally {
+      // Replace and dispose a leaked observer when this regression runs on old code.
+      vi.mocked(getUpdateRun).mockImplementation(() => run);
+      presentation = createUpdateProgress(true, context);
+      presentation.dispose();
+      presentation = undefined;
+    }
+  });
+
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4481c8dcd349f3401780786ad7c318cca3bf329e](https://github.com/openclaw/openclaw/pull/143965/commits/4481c8dcd349f3401780786ad7c318cca3bf329e)
- Landed commit: [f13d6746e6106a0d5ebb4631818f51f6704d9c2b](https://github.com/openclaw/openclaw/commit/f13d6746e6106a0d5ebb4631818f51f6704d9c2b)

PR #143974, opened 2026-09-10
docs(update): clarify automatic legacy status expiry

steipete:
## What Problem This Solves

The update status and installation guides still said status is always read-only and identityless update records always need manual repair. Current code already expires a narrow class of untouched legacy admissions after 24 hours, so those blanket statements gave operators conflicting guidance.

## Why This Change Was Made

Describe the existing exception and link the installation guide to its detailed conditions: an untouched identityless admission at its initial requested step, older than 24 hours, with no retained recovery descriptor. Preserve manual recovery guidance for other stale records.

## User Impact

Operators can distinguish records that can expire automatically from those requiring explicit recovery. This changes documentation only; update behavior and storage policy remain unchanged.

## Evidence

The 10-stage docs-only changed gate, canonical formatting, changed-page MDX, glossary and whitespace checks passed. Independent review found no actionable findings through P2. The canonical publishing parser resolves the new link.

The whole-doc anchor audit still reports three pre-existing `windows-app-node` references in generated maturity pages. Their source and parser are unchanged from the base, and the same parser confirms the baseline failures. External ClawHub fragments remain unverified without that source checkout. No runtime tests were needed for this wording correction.

### DIFF
--- a/docs/cli/update/status-and-history.md
+++ b/docs/cli/update/status-and-history.md
@@ -50,14 +50,20 @@ includes `activeRun` when a run is active and `lastRun` when history exists.
 If history cannot be read or classified, status still shows update availability
 and runtime findings. Human output explains that run status is unavailable;
 JSON includes `runStatusError` and omits the run fields. This does not mean
-there are no active or past runs, and status does not repair the underlying state.
+there are no active or past runs, and status does not repair unreadable history.
+
+Status can reconcile an untouched, identityless legacy admission after more than
+24 hours if it remains at its initial `requested/in_progress` step and has no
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [523f9d9eb92da5cdff1a5a78773019deea099217](https://github.com/openclaw/openclaw/pull/143974/commits/523f9d9eb92da5cdff1a5a78773019deea099217)
- Landed commit: [88c98a90f5d0241076315cb37b80364414392399](https://github.com/openclaw/openclaw/commit/88c98a90f5d0241076315cb37b80364414392399)

PR #143975, opened 2026-09-10
fix(sessions): preserve resolved model override identities

steipete:
Related: #130706, #142100.

## What Problem This Solves

The session model-override writer records an exact selected model ID and marks it resolved. Readers then strip a provider prefix and parse aliases again before checking that provenance, turning a stored `custom/model` into `model` or advancing a selected alias target. Gateway row caches also conflate raw and resolved overrides, making the result depend on row order.

## Why This Change Was Made

Pass existing route provenance through the persisted model readers and preserve IDs already selected by the writer. Reuse the existing provenance owner, including legacy automatic-fallback provenance. Key the in-memory Gateway projection cache by the original override fields and resolution state. This changes five production files by six net lines and adds no persistent field or schema.

## User Impact

Resolved session overrides retain the selected model through direct reads, agent-scoped selection, and Gateway row projection. Raw legacy parsing, default resets, parent inheritance, and automatic-fallback exclusion retain their existing behavior. Reply-selection integration remains a separate small cut.

## Evidence

Four intended writer-to-reader/cache regressions failed against unchanged production. Applying only the reader repair left both cache-order cases failing, proving the cache distinction is necessary. The complete repair passed 192 tests across five suites, fresh managed P2 review, and the full changed-file gate.

At exact commit `c6bf573fbcf0b5ed7485a3718f99d43e2a222caf`, one canonical `sourcePerformance` build and all eight compiled behavior cases passed. The real writer, direct/session readers, and exported Gateway row builder preserved resolved prefix and alias-like IDs, legacy provenance, raw controls, both cache orders with repeat hits, and runtime-versus-agent selection. All 12,179 inventoried build files remained unchanged; source and other-checkout import denial controls passed; owned processes and state were cleaned up.

This is synthetic compiled component-boundary proof, without external-provider or live Gateway traffic. The runtime build excludes UI and SDK declarations. Initial artifact discovery found the projection helper was internal, so the probe used its real exported row-builder caller; no production seam or rebuild was added.

### DIFF
--- a/src/agents/model-selection-persisted.ts
+++ b/src/agents/model-selection-persisted.ts
@@ -1,6 +1,7 @@
 // Persisted model metadata normalization without loading the broader selection runtime.
 import { normalizeOptionalString } from "@openclaw/normalization-core/string-coerce";
 import { DEFAULT_PROVIDER } from "./defaults.js";
+import type { ModelFallbackRouteResolution } from "./model-fallback.types.js";
 import type { ModelRef } from "./model-ref-shared.js";
 import { parseModelRef } from "./model-selection-normalize.js";
 
@@ -12,6 +13,7 @@ export function resolvePersistedOverrideModelRef(params: {
   defaultProvider?: unknown;
   overrideProvider?: unknown;
   overrideModel?: unknown;
+  routeResolution?: ModelFallbackRouteResolution;
   allowManifestNormalization?: boolean;
   allowPluginNormalization?: boolean;
 }): ModelRef | null {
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c6bf573fbcf0b5ed7485a3718f99d43e2a222caf](https://github.com/openclaw/openclaw/pull/143975/commits/c6bf573fbcf0b5ed7485a3718f99d43e2a222caf)
- Landed commit: [8295be88168a3fb56f0b9ea7f62c739736acb2f4](https://github.com/openclaw/openclaw/commit/8295be88168a3fb56f0b9ea7f62c739736acb2f4)

PR #143978, opened 2026-09-10
refactor(memory-wiki): share deferred test fixtures

steipete:
## What Problem This Solves

Memory Wiki's test suites maintained six copies of the same deferred-promise fixture used to coordinate concurrent work.

## Why This Change Was Made

Replace those copies with one pure, private test helper that supports both void and value-bearing promises. All six suites import it directly, so source-sync tests do not load the larger shared harness or its cleanup hooks. Assertions, ordering and cleanup remain unchanged; two redundant explicit void type arguments are omitted.

## User Impact

No runtime or user-visible behavior changes. This removes 35 lines of duplicated test scaffolding without removing coverage.

## Evidence

- The same six complete suites passed before and after: 85 cases on each side, with identical case identities and outcomes.
- All 29 deferred calls and 212 assertion sites are retained, including queue ordering, service retirement, rollback recovery, background responses and source-sync coalescing.
- Formatting, the complete 20-check changed-file gate (including extension-test types and lint), and independent source review passed. The first gate caught two redundant default type arguments; the corrected candidate then passed the complete gate and all 85 cases.

### DIFF
--- a/extensions/memory-wiki/index.test.ts
+++ b/extensions/memory-wiki/index.test.ts
@@ -13,6 +13,7 @@ import {
   type MemoryWikiCompiledCacheSnapshot,
 } from "./src/compiled-cache.js";
 import { resolveMemoryWikiConfig } from "./src/config.js";
+import { deferred } from "./src/deferred.test-helpers.js";
 import {
   appendMemoryWikiLog,
   loadMemoryWikiValidatedVaultIdentity,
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f2eb2da2cbfc58ceaa7ad9a4b92ec40ff9771a13](https://github.com/openclaw/openclaw/pull/143978/commits/f2eb2da2cbfc58ceaa7ad9a4b92ec40ff9771a13)
- Landed commit: [2f8cd215e92320703237b8def415b691f15f3b56](https://github.com/openclaw/openclaw/commit/2f8cd215e92320703237b8def415b691f15f3b56)

PR #143989, opened 2026-09-10
fix(qa): close Crabline after gateway shutdown

RomneyDa:
Backports the QA lifecycle correction for Crabline-backed channel runs. Crabline local servers now close only after the gateway has stopped, so active long-poll connections cannot pin server.close during post-scenario cleanup. Includes adapter-phase assertions and updates transport cleanup coverage. Validation: 22 focused tests; stable test typecheck; repository lint; Linux before-fix loop reproduced on iteration 6 and identical after-fix loop passed 20/20.

### DIFF
--- a/extensions/qa-lab/src/crabline-transport.test.ts
+++ b/extensions/qa-lab/src/crabline-transport.test.ts
@@ -31,6 +31,8 @@ describe("crabline transport", () => {
 
       try {
         expect(transport.id).toBe("crabline");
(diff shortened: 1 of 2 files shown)


RomneyDa:
Additional exact-head validation: full QA Lab suite passed (102 files, 1,367 tests). Linux reproduction passed 20 consecutive group-visible-reply-tool runs after the fix; the unchanged build had timed out on iteration 6 inside Crabline adapter close.

PR #143992, opened 2026-09-10
fix(qa): close Crabline after gateway shutdown

RomneyDa:
Moves Crabline local-provider server close into the post-gateway cleanup phase. Closing it first can wait forever on a gateway-owned long-poll connection, pinning an otherwise successful QA run. Includes adapter-phase assertions and updates direct transport tests. Validation: 32 focused tests; full test typecheck; repository lint and formatting; Linux before-fix loop reproduced on iteration 6 and identical after-fix loop passed 20/20.


RomneyDa:
CI note: this head is temporarily stacked on #144037 so the synthetic merge includes the current-main docs navigation repair. Merge #144037 first; the resulting #143992 diff remains the focused Crabline lifecycle change.

PR #143993, opened 2026-09-10
refactor(cli): keep root option forwarding in scanner

steipete:
## What Problem This Solves

Root CLI option handling had a forwarding adapter used only by the shared scanner. Keeping the token-copy loop in a separate owner added an unnecessary layer to the profile and container parsing paths.

## Why This Change Was Made

The shared scanner now owns forwarding directly and the one-caller adapter is removed. The existing root-option policy still decides how many tokens to consume. Visitor order, undefined filtering, literal `--` handling, argument identity and profile/container parsing remain unchanged; Commander is not substituted for the intentionally conservative root parser.

## User Impact

No user-visible behavior change is intended. Existing flags, aliases, precedence, diagnostics, profile selection and container routing are preserved. No public API, configuration, dependency, protocol or persistence surface changes.

Production LOC: **+8/-26, net 18 removed**. Tests are unchanged.

## Evidence

- Original and candidate each passed all 148 cases across the four existing profile, container-target, root-option-value and root-options suites.
- Separate original/candidate processes exercised 66 representative cases through the real profile and container parsers. Both asserted independent expected results, and their recorded observations matched. This covers representative public string-array inputs, not every possible scanner callback.
- The normal candidate build completed in 3m17.3s. Built-CLI smoke verified help with profile flags before and after `--help`, plus a missing-profile diagnostic: exits 0, 0 and 2. All three processes completed normally and the private fixture home was removed. This is CLI startup/help proof, not real container or Gateway execution.
- Those tests, parser captures, build and smoke ran at base `3e6b53c492d9bd6019983efea1c890b35a930146` plus the identical two-file patch; they are not claimed as executions of the newer whole tree.
- The [normal changed gate](https://github.com/openclaw/openclaw/actions/runs/34468205150) passed all 28 planned phases on base `7b6287b2db9b6867293e5e5f0bad13642a97f01c` plus the patch, tree `06c8ee5c70a279009bf02447e1fa4383ef9e1173`. Core types, test types, all three unused-export scans and core lint passed; core lint checked 22,045 files with zero warnings or errors. The source-bound wrapper exited 0, the one-shot lease completed, and its temporary checkout/registration and recorded processes were absent.
- Managed precommit review through P2 completed without actionable findings using 19 approved context files. Two unchanged background-context files, `src/infra/inline-option-token.ts` and `src/cli/container-target.test.ts`, were explicitly omitted. The first two preparations stopped before model execution; their refusals remain retained, and no omitted contents or scanner-policy overrides were introduced. This is a scoped review, not a claim that the omitted contexts were reviewed by that model.

The initial saved commit was `45c4dbbad8f6aa0c5fb0c2af964ee6f26b6546ae`, with the same gate-tested tree. Its separate exact-commit Codex review completed through P2 with no actionable findings, using the same 19 approved contexts and retaining both omissions above. The current refreshed head and its proof are described below.

## CI-driven refresh

The first exact-head CI run [34471698491](https://github.com/openclaw/openclaw/actions/runs/34471698491) failed only the unrelated docs-navigation fixture after an intentional route move. Canonical #144002 fixed that route expectation; #144037 then aligned the fixture and Chinese overlay with the newer Releases/Contributing split. Both fixes are upstream, with #144037 landed as `1670c5a1cc375f0121d4f59b17d531fa151c7434`. Our separate tested alternative was retained unpublished to avoid a competing repair.

Failure-driven refresh onto that landed base produced signed `0e8b0d010480cf52abb7ae8810cfb20ffe4177fb` (tree `8da4eadbbe5805000cb35c8fe4e7a33839a56389`). The two-file patch and scanner afterimage are byte-identical to the earlier candidate. All reviewed CLI owner contexts, package/lock and Commander contracts are unchanged; the docs index only has prose and worktree-list additions. Fresh native proof on this exact head passed all 148 CLI cases plus all 11 docs-publisher cases. Earlier parser captures, build/smoke and full 28-stage gate retain their original source bindings above; they are not relabeled as fresh whole-tree executions.

Separate exact-head Codex review passed through P2 with no findings using the same 19 contexts and two omissions. Its first request ended with a service-capacity error before any verdict; one later same-settings retry passed. Neither model nor review policy was changed. New exact-head hosted CI and ordinary native hosted preparation are still required before landing.


steipete:
Merged via squash.

- Prepared head SHA: [0e8b0d010480cf52abb7ae8810cfb20ffe4177fb](https://github.com/openclaw/openclaw/pull/143993/commits/0e8b0d010480cf52abb7ae8810cfb20ffe4177fb)
- Landed commit: [71965b81e0c44c7001d6fd97cab9aa3bdae91120](https://github.com/openclaw/openclaw/commit/71965b81e0c44c7001d6fd97cab9aa3bdae91120)

PR #143792, opened 2026-09-10
refactor: share prepared completion cleanup

steipete:
Related: #130706, #142100. Follows #143716.

## What Problem This Solves

Prepared completion acquisition duplicates the tracked cleanup lifecycle already shared by isolated completions and TTS. That duplication makes it harder to maintain the requirement that setup work and the acquired caller both finish before model resources are released.

## Why This Change Was Made

Use the existing shared async-work resource owner. Keep the release state, successful caller callback, captured runtime context, and both release conditions unchanged. This maintainer-requested cleanup changes one production file and removes 29 lines.

## User Impact

Prepared completions retain their existing result timing and resource lifetime with one cleanup implementation.

## Evidence

The reviewed patch was replayed without conflicts or changes onto main containing #143813 and the separate test-fixture and UI-lint repairs #143875 and #143889. Fresh validation at `859c1d72b3f13e50d822829cce4e400a777703f5` passed 37 focused cases, the full changed-file gate, and one canonical `sourcePerformance` runtime build. All seven prior review source inputs remain identical, so the independent P2 review remains applicable.

Two freshly compiled ordinary acquisition cases passed with the caller releasing before cleanup drained and with cleanup draining before caller release. Both completed a loopback request, kept the resource open while either condition remained outstanding, and observed exactly one disposal afterward. Each loaded 2,881 compiled imports and no checkout-source files. The build uses its natural new commit metadata; no artifact was restamped. This runtime profile excludes UI and declarations.

Run generations own selected plugin resources, following #143725. An older external proof initially asserted caller-owned raw-resource behavior; that failure and the correction to the current owner contract remain preserved in the original evidence. The refreshed proof uses the corrected contract unchanged. Ephemeral ownership remains covered by existing source tests.

Issue #130324, opened 2026-08-26
Gateway remains at ~100% CPU with 32–54s event-loop stalls on current main 99a02bf1


elikampf:
## Summary

The gateway remains at approximately 100% CPU on current `main` after updating and rebuilding with the supported source update flow. The gateway is responsive to a deep connectivity probe, but repeatedly stalls the Node.js event loop for 32–54 seconds.

This is a current-main reproduction related to:

- #118846
- #125595

## Version and installation

- Running source version: `OpenClaw 2026.8.1 (99a02bf)`
- Full source commit: `99a02bf115cfc4a594885bc170c877ae182ff352`
- Source update method: `scripts/update-gateway.sh`
- Repository: `https://github.com/openclaw/openclaw.git`
- Service command: `/usr/bin/node /home/ubuntu/openclaw-src/dist/index.js gateway --port 18789`
- Node.js: `v24.19.0`
- OS: Ubuntu 24.04.4 LTS, Linux 6.17.0-1020-oracle, ARM64 (`aarch64`)

There is also an older global CLI on this host:

```text
$ openclaw --version
OpenClaw 2026.8.1-beta.3 (5831b80)
```

It is **not** the gateway executable. The systemd `ExecStart` and deep status both confirm that the running gateway uses `/home/ubuntu/openclaw-src/dist/index.js` at `99a02bf1`.

## Reproduction

1. From the source checkout, run `scripts/update-gateway.sh`.
2. Clean dependency installation and full source/UI build complete successfully.
3. Restart the systemd user gateway service, which points at the newly built source `dist/index.js`.
4. Run the source CLI deep check and inspect the gateway process CPU.

## Actual behavior

The deep status reports:

```text
CLI version: 2026.8.1 (~/openclaw-src/dist/index.js)
Gateway version: 2026.8.1
Runtime: running
Connectivity probe: ok
Listening: 127.0.0.1:18789
```

After approximately 7 minutes, the gateway process was still at 96.5% CPU. A later sample was 99.1%:

```text
PID      ELAPSED  %CPU  %MEM  RSS      STAT  COMMAND
1122168  729      99.1  6.4   1591384  Rsl   /usr/bin/node /home/ubuntu/openclaw-src/dist/index.js gateway --port 18789
```

The first three post-update stalls were:

```text
liveness heartbeat delayed: overdue=2030ms elapsed=32030ms
liveness heartbeat delayed: overdue=23597ms elapsed=53597ms
[...]

PR #143994, opened 2026-09-10
fix(models): preserve selected IDs during rematerialization

steipete:
Related: #130706, #142100. Follows #143792.

## What Problem This Solves

Refreshing model metadata can reinterpret an already selected model ID as another input alias. With `entry → middle → final`, refreshing `middle` can select `final`. Captured metadata can also choose an equivalent row instead of the exact logical model, supplying the wrong image limit.

## Why This Change Was Made

Give the existing async resolver an explicit input/selected distinction. Four verified callers preserve the IDs returned by model resolution: simple completion, isolated completion, BTW, and image credential refresh. Captured metadata prefers the exact logical row before the existing static-equivalent fallback. This adds 21 production lines.

## User Impact

These completion and refresh paths keep the selected executable ID and image limits. Auth-plan and compaction callbacks still accept mixed logical inputs and retain their existing input-normalization behavior. A real auth-boundary regression prevents a raw alias from being fabricated as an executable model. Existing validation can still reject a raw logical ID that differs from the resolved ID; the canonical producer-to-auth handoff belongs to #143961.

## Evidence

The original resolver regressions and the new real auth regression failed before their respective repairs. The narrowed cohort passed 55 model/auth/compaction tests; four auth tests were repeated after correcting the new registry fixture's declared input shape. Fresh P2 review covers current neighboring row/registry changes. The full changed gate passed after that test-only type correction.

One sourcePerformance build at exact commit `4e923e0129bec3d316a7de2b6269418ef56157f1` and six real compiled loopback HTTP requests passed, covering raw aliases, selected IDs, exact 1,100-pixel limits, and credential/route rematerialization. Compiled raw-logical-ID rejection also passed. The private auth callback's direct proof is source-level (six real setup/auth cases), not represented as a compiled export.

All 9,549 root and canonical workspace outputs remained unchanged; zero checkout source imports, and the isolated listener/state were cleaned up. This is synthetic loopback proof, not an upstream-provider request; the runtime profile omits UI and declarations. The original build and failed CI remain preserved. This PR does not establish a fix for #130324.

CI refresh: the unchanged task patch is replayed onto main `3943149f8ac269fc138ca21b5786ff2a9db3dd23`, including the navigation repair in #144037. Every changed-file postimage matches the previous reviewed head `4e923e0129bec3d316a7de2b6269418ef56157f1`. Original exact-commit runtime proof remains preserved and is not restamped; fresh CI is running for `b55528b79eea857e6f908e72f8daaa4fe99bda03`.

### DIFF
--- a/src/agents/btw.ts
+++ b/src/agents/btw.ts
@@ -439,6 +439,7 @@ async function materializeBtwRuntimeModel(
       ...(params.forceResolve !== undefined ? { forceResolve: params.forceResolve } : {}),
       resolveModel: ({ config, authProfileId, authProfileMode }) =>
         resolveModelAsync(params.provider, params.modelId, agentDir, config, {
+          modelIdSource: "selected",
           authStorage: params.authStorage,
           modelRegistry: params.modelRegistry,
           skipAgentDiscovery: true,
--- a/src/agents/embedded-agent-runner/model.registry-resolution.ts
+++ b/src/agents/embedded-agent-runner/model.registry-resolution.ts
@@ -377,6 +377,7 @@ export function shouldCompareProviderRuntimeResolvedModel(params: {
 export function normalizeProviderModelRef(params: {
   provider: string;
   modelId: string;
+  modelIdSource?: "input" | "selected";
   cfg?: OpenClawConfig;
   workspaceDir?: string;
 }): {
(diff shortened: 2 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b55528b79eea857e6f908e72f8daaa4fe99bda03](https://github.com/openclaw/openclaw/pull/143994/commits/b55528b79eea857e6f908e72f8daaa4fe99bda03)
- Landed commit: [5f161b648bc9127a84ec9eadff1cfa5a86b21a13](https://github.com/openclaw/openclaw/commit/5f161b648bc9127a84ec9eadff1cfa5a86b21a13)

PR #143899, opened 2026-09-10
fix: keep moved capture directories out of exports

fuller-stack-dev:
Related: #143799

## What Problem This Solves

Fixes an issue where backups or support exports could include private update-capture files after the original state directory was removed or a retained capture directory was relocated. The paired-root exclusion introduced in #143799 cannot identify those paths on its own.

## Why This Change Was Made

The existing shared export guard now recognizes a fixed, versioned, exclusion-only marker in a selected directory or its ancestors. It checks lexical and canonical paths, reads only the bounded regular marker file without following marker symlinks, and rejects malformed, unreadable or changed markers. It never parses workspace manifests or scans other state roots. Archive planning checks original lexical selections and their canonical targets before deduplication, including managed-skill target promotion. Archive traversal errors are carried through to a publication refusal after tar settles.

The producer contract is `.openclaw-private-update-capture` containing exactly `openclaw-private-update-capture-v1\n`. Producers must durably write it before raw data in each independently movable container. Producer creation, capture/reopen, lifecycle eligibility and deletion are separate work. The marker grants exclusion only, not ownership or authority. This patch creates no captures and changes no retention policy.

## User Impact

Marked directories remain excluded after owner removal, relocation or copying with their markers intact. Explicit SQLite/Git inputs and support-export aliases get the same protection. Healthy siblings, similarly named directories and ordinary manifest files remain usable. Unmarked files copied out of protected directories are outside this classification rule.

## Evidence

- Six new real-caller regressions failed on the original guard: enclosing/nested archive selection, malformed-marker publication refusal, explicit SQLite/Git inputs, and support config/log/stability inputs.
- Nine new reader controls failed on the original guard: malformed/type/version/size cases, unreadable marker, FIFO refusal without opening it, and identical-byte replacement during a bounded read.
- Initial marker candidate: 32 affected archive/snapshot/support tests plus ten reader controls passed. The subsequent lexical-admission correction passed all 22 archive/privacy cases plus three healthy managed-skill archive controls. The unchanged reader/support results remain applicable and were not replayed. Original failure logs are preserved.
- Initial P0/P1/P2 review found one short-circuit inconsistency. It was reproduced in three additional controls and fixed by validating full marker ancestry before legacy exclusion. The original finding is preserved. The unused-export gate also required moving the shared constants into a format-only module with a real production consumer. Final required changed-source checks passed, including core/test typechecks, unused-export scans, lint and repository guards. Exact-head CI is required before native landing.
[...]

PR #143799, opened 2026-09-10
fix(backup): exclude private update captures from ordinary exports

fuller-stack-dev:
Related: #142770. Follow-up to #140339.

## What Problem This Solves

Resolves a problem where selecting a workspace containing the managed private capture root includes raw capture files in an ordinary backup. Explicit capture-file inputs can also enter support exports, SQLite snapshots and Git backups. The accepted capture design requires this boundary before enabling capture creation.

## Why This Change Was Made

Exclude the selected state's exact managed `<stateDir>.update-captures/` root and another state's exact paired root through the existing backup inventory, archive traversal, named support inputs, and shared snapshot producer. Explicit nested workspace selection cannot override this exclusion. Config or database backup sources inside those roots are refused. Cross-state recognition checks only the selected path's lexical and canonical ancestors for an exact `<owner>.update-captures` directory paired with an existing `<owner>` directory. Similarly named and unpaired unrelated workspace directories stay included. This directory-layout rule is not writer authority and cannot identify moved, renamed or orphaned captures belonging to another state.

Accepted storage/privacy decision: https://github.com/openclaw/openclaw/pull/140339#issuecomment-5614409997

Accepted prospective transaction-scoped lifetime revision: https://github.com/openclaw/openclaw/pull/140339#issuecomment-5614508856

This is the independent privacy prerequisite. It does not create retained captures, change ordinary snapshot sanitization, add cleanup/pruning, implement replay, or touch #143750/#143752 package mutation ownership. Capture/reopen will follow separately on this boundary.

## User Impact

Managed raw update artifacts cannot be included accidentally in ordinary backup and support exports through these paths. Existing ordinary config/workspace files and backup sanitization remain supported. This is not a guarantee for raw data manually copied outside the managed root.

## Evidence

- Seven new privacy controls fail on unchanged production source for the intended reason: parent/nested workspace archives, full/config-only backup source selection, SQLite/Git snapshot source selection, and support export inputs.
- Final validation on `3d665e6a` passes eleven archive/SQLite/Git controls and all thirteen support-export tests, including ordinary sanitization and healthy-error cases. Six cross-state regressions failed on the previous guard for the intended reasons and pass with this correction. The archive case reads real tar output, verifies the archive, preserves both healthy neighbors, and checks source bytes.
- The existing transient-row sanitization/global-owner control also passes (one selected test; 65 unrelated cases skipped).
[...]

PR #144005, opened 2026-09-10
fix(update): back up state before migrations and restore it on rollback

steipete:
## Current integration status (2026-09-17)

Draft head `156282f60d28e722bd339eb17f34b7abfcf58a1e` incorporates pinned main `9ca2d1dfcd161ed862a30749f40486e95d6459f1`. This aligns the base for stacked #145169 so its CI does not treat upstream locale refreshes as task source changes. The eight merge conflicts follow current source owners; normal commit hooks passed. The original unpublished recovery-history contribution is preserved.

**Not ready for main landing alone.** Independent canonical review confirmed the pre-existing raw-baseline restore can discard live update-ledger history. The reviewed generation/publication correction in #145169 is a required dependency, not waived by this base refresh. Combined exact-source review/CI, full-built unmodified-published-driver first hop, separate built Doctor plus a second protected update, and native qualification remain required.

Older checkpoints below are historical evidence, not current-head PASS claims. Task-generated containment/admission cycles and historical cleanup slot holds no longer apply to ordinary maintainer work; real product ownership, data preservation, review and CI requirements remain.

---

**Integration prerequisite; not ready for main.** Current-source composition is being completed with stacked #145169. That stack must merge here before main landing; its forward-recovery and final-effect retirement fixes must not be separated from this capture prerequisite. Separate full-built first-hop and Doctor acceptance remain pending.

Related to #142770; original-driver H1/H2 acceptance remains open.
Refs #144422

**Blocked; not land-ready.** The remaining-copy capacity handoff still needs its reserved-owner exception. Fresh published-9.3 proof now identifies an additional first-hop gap: the installed 9.3 Matrix migration has no recovery-resource declaration, so candidate rehearsal refuses before live migration. A newer-main bootstrap contract also needs reconciliation: the target Doctor can write authored config before capture/run registration, and main's named test preserves those writes on failure. No weakening, guessed inventory, timeout increase, or silent unprotected exception has been applied.

## What Problem This Solves

Fixes an issue where updating from 2026.9.2 could leave the Gateway unusable after rollback: target Doctor migrated persistent state, package rollback restored the old binary, and that binary failed with `no such column: workspace_dir`. It also fixes no-op updates stopping healthy Gateways, completed legacy captures blocking later updates, and repair capture overhead growing with every inventory entry.

## Why This Change Was Made

This implements the [accepted lossless-capture contract](https://github.com/openclaw/openclaw/issues/140339#issuecomment-5614409997) and [transaction-scoped lifecycle decision](https://github.com/openclaw/openclaw/issues/140339#issuecomment-5614508856). Peter's explicit 2026-09-10 approval: “update should always work. It's best effort. If there are minor hiccups we add that in the warning log. We should backup things before, and rollback restores the backup.”

### Capture, restore and retirement

Before protected mutation, existing backup machinery captures config/include files, shared and per-agent SQLite, Skill Workshop and inventoried plugin resources. SQLite uses online backup; recovery captures remain unsanitized, without delivery-row or TTL pruning. One immutable manifest binds numbered resources by size and SHA-256; the terminal outcome is separately write-once.

Sets live at `<stateDir>.update-captures/<captureId>/` with private ancestors and owner-only permissions. The shared privacy marker excludes them from ordinary backup/support exports, including explicit containing/nested selections. User archives retain their existing behavior. Capacity admission covers capture, package/recovery staging, growth and reserve; insufficient space preserves live data and earlier sets.

Generic JSON/plan adapters and custom bundled migration owners enumerate inputs through existing collectors. The optional read-only `requireLocalResources` collector argument rejects remote resources when complete local capture is required; it is not configuration. Missing declarations refuse protection.

Rollback restores package files and captured state, then verifies the previous managed Gateway. State-only finalization restores its capture without inventing a package transaction. Database-directory restore removes migration-created files only within the declared boundary; reopened LanceDB verifies captured schema, rows and version. The shared copy helper clears restored `state_leases` and `agent_database_leases`: process leases cannot survive copying. Captures remain lossless.

Both finalizer Doctor phases share one verified capture and config-write receipts. Mutating children settle before restoration. Manual recovery retains protection through subsequent Doctor/bootstrap writes, restores again if continuation fails, and closes reopened stores. Standalone repair never starts the Gateway. Data failures are hard errors; reporting failures after verified recovery are advisory.

There is no count, age or capacity-pressure pruning. Existing terminal owners retire captures only after durable, identity-bound success and settled recovery dependencies. Admission reconciles completed prior transactions through Doctor's resolution owner, requiring confirmed matching runtime/version/build and successful Doctor/Gateway verification. Explicit false readiness fails; missing newer fields remain compatible with 9.2. Nonterminal, restored, failed-restoration and unverified sets are not automatically adopted.

### No-op and tagged-upgrade contracts

Convergence requests protection only for actual plugin/config mutation, using facts already available inside its existing lease. Unchanged baseline publication and repeated failure of an already-disabled plugin preserve their no-change result. Real enablement, alias and slot changes remain mutations. Exact target identity is always validated; an already-current core uses its installed schema/runtime metadata. Lease ordering, publication fencing and `--no-restart` semantics are unchanged.

Admission checks capacity and independent writers before shutdown. A late pre-mutation refusal restores the original service, including with `--no-restart`. Native parent/launcher identity and Gateway-lock checks distinguish a supervised Gateway child from unrelated writers. Capture callbacks now check live executor/maintenance fences; full recovery admission runs before and after capture, avoiding repeated SQLite snapshots per resource.

The named **tagged 2026.9.3/2026.9.4 rehearsal contract** accepts their disposable copies only with the existing update-in-progress marker, exact private environment/config/data layout, observed live parent chain and matching copied-ledger identity. It verifies local resources, OAuth paths, SQLite companions, ownership and links under existing maintenance authority. Identity-bearing driver/capture calls remain strict. Admission emits a typed `legacy-driver-rehearsal` fact and warning. Remove this compatibility path when those tags cease being supported upgrade sources.

Inline Git RPC remains explicitly unprotected, recorded in its parent run and reflected in result/status: “This Gateway-initiated update is not protected by a recovery capture. Run `openclaw update` from a terminal for a protected update.” The external drain/stop/capture continuation is [#144422](https://github.com/openclaw/openclaw/issues/144422).

This builds on @fuller-stack-dev's #143693, #143799 and #143899, and reuses @vincentkoc's cleanup fencing (#143791) and terminal settlement (#143921). Package activation stays with #143750/#143752. Please keep maintainer edits enabled.

### Data-model compatibility

No SQLite schema/column, protocol, configuration option or dependency change. One recovery-set kind and existing run metadata hold config receipts, restoration facts and unprotected RPC identity. No ratchet increased; the upstream environment-variable budget passes at **492/492**.

### Owner flow map

Pre-fix references are pinned to `298493fa54b8d2f9091e504c178f850591a1757c`:

| Boundary | Previous behavior |
| --- | --- |
| `src/infra/package-update-swap.ts:118`; `package-update-steps.ts:1215` | Package/launcher backups only. |
| `src/cli/update-cli/update-command-package.ts:84,110,278,306`; `update-command-config-snapshot.ts:25` | Root-config rotation followed by target Doctor; no database/include capture. |
| `src/commands/doctor.ts:164`; `src/flows/doctor-health.ts:121,160,237` | Migration maintenance without state restoration on failure. |
| `src/cli/update-cli/update-command-rollback.ts:285,341,436`; `src/infra/update-managed-service-handoff.ts:1042,1821,1898` | Schema refusal or package/root-config rollback, then Gateway health/version/build checks. |

Published 9.3's previously healthy rollback preserved schema compatibility; it did not restore SQLite.

| Current responsibility | Owner |
| --- | --- |
| Admission/capture | `src/cli/update-cli/update-command-backup-lifecycle.ts:48,180,214` |
| Capture inventory/capacity | `src/infra/update-recovery-backup-create.ts:423,427`; `update-recovery-capacity.ts:44` |
| Plugin resource collection | `src/plugins/doctor-migration-resources.ts:12` |
| Tagged rehearsal | `src/commands/doctor-update-rehearsal.ts:127,489` |
| Doctor recovery/reconciliation | `src/commands/doctor-update-recovery.ts:186,347,630,654` |
| Restore/lease clearing | `src/cli/update-cli/update-command-rollback.ts:387,477`; `src/infra/update-recovery-backup-restore.ts:22,220` |
| Terminal publication/retirement | `src/cli/update-cli/update-command-terminal.ts:264`; `update-command-backup-lifecycle.ts:375` |

## User Impact

Protected failures recover state before restarting the previous Gateway. True no-ops keep healthy Gateways serving. Retained sets appear in `update status --json` with `npx openclaw@latest doctor --fix` guidance. With 9.2, a failing target Doctor restores state before package rollback; because that updater may leave the Gateway stopped, output also names `openclaw gateway start`.

Release-note context: persistent-state backup before migration, complete rollback, safe unresolved recovery and healthy no-op behavior.

### Latest checkpoint (2026-09-11)

Published branch checkpoint: `ee09337d115b4a721b46572443f042f7c3ce650e`; local rebase: `f118e3396f15dfb40a901fcd323efc4a57554dd8` onto `24109cc97c2`. The installed-schema/terminal-registration corrective diff remains uncommitted because its fresh P1 review identifies the bootstrap ownership gap. Earlier evidence below retains its original source identity; it is not current-head green proof.

- Fresh-profile current-package regression failed before with `ExitError: exit 1` instead of reaching the installed initialization owner. After sharing the existing bounded manifest reader: fresh-preview plus backup-lifecycle suites **62/62 passed**.
- Native restart argv correction: failing-before expected argv omitted `--update-executor run`; focused after **1/1 passed**, then upstream independently incorporated the same assertion.
- Root-idle Windows probe: one of two cases timed out at **120.131 s**. Instrumentation in the actual nested test worker counted **153,196 installation-tree filesystem operations / 38.669 s**, with no SQLite read-only child invocation. This is an owned capacity-walk cost, not an attribution to host load.
- The metadata six-row table and refresh five-row table passed; those focused probes do not dismiss the previous original-order failures. Full CLI rerun remains blocked on the capacity repair.
- Fresh P1 review of the corrective diff: **one finding (0.99)** at `update-command-target.ts:306`, because allowing already-current initialization exposes config writes before recovery ownership. The full-branch review was interrupted after local typechecking found a duplicate `updateRun` property; that duplicate is removed. The rerun passed all typechecks, then was intentionally stopped during type-aware core lint after this P1 confirmed the ownership STOP.

| Live control | Exact candidate | Result | Captures / bytes | Observed downtime | Previous Gateway |
| --- | --- | --- | --- | --- | --- |
| Published 2026.9.3 driver, populated baseline | `a3493020bf9c` | Refused in candidate rehearsal: `matrix/matrix-account-sqlite-schema does not declare its migration data resources` | 0 / 0 B | 0 ms, 261 samples | Status, RPC and mock chat passed on 2026.9.3 |

The command took **258.720 s**; health was verified at **262.491 s**. Packaging passed the canonical tarball and dist import-graph checks: **10,936 files, 209,542,875 unpacked bytes**, no task artifacts/captures; SHA-256 `0c17ca456bb12d5539d60c5a6138805205316a729e80fc6209736db57edf801b`. This was not a fingerprint timeout and is not a successful update control. Official survivor and populated repair reruns are not completed for this checkpoint.

The bootstrap conflict is explicit in main's `update-command-fresh-preview.test.ts`, “accepts target Doctor config changes that preserve the selected stored channel”: it throws after Doctor writes and expects the migrated config to remain, with no shared database. Capture before that initialization needs an owner design that does not create a newer-schema run ledger first. The pending copy-footprint proposal carries read-only source/destination facts through existing callbacks; it does not alter package activation ordering or fallbacks.

Earlier proof was taken after rebasing onto `b7693bf50aa`, which contains the required upstream planner fix `87504f7ec9e`. Workshop continues to use main's single artifact-preserving source reader and retains main's mutable migration behavior.

Latest focused proof: **317 passing tests across the latest delta's 19 test files**, run individually—231 plugin tests in 15 files, 37 Doctor, 29 registry, four SDK and 16 backup-lifecycle tests—plus **24 sibling tests**. All 37 changed extension files pass lint. Latest scoped review (`review-final-export.log`) reports no actionable P0/P1 findings (**0.94**), including the routing, native handoff, symlink and unused-export fixes. Built-registry census, under synthetic default and enabled-channel configurations, finds **46/46 collectors across 21 plugins** without executing migrations.

| Regression | Failing-before evidence |
| --- | --- |
| Stale recovery | Newer session disappeared; expected `after-update`, received undefined. |
| Logical directory restore | LanceDB reopened migrated schema/extra row/version 3 instead of captured version 1. |
| Retention | Unresolved set disappeared after three later backups. |
| Restored leases | Four failures; lease timeout with 297299 ms remaining. After: 160 passing, one existing skip; acquisition 29 ms. |
| No-op/admission | Six failures: two false mutation requests, two premature stops, missing late restart and retained completed set. Additional Git refusal and three false-readiness cases failed. |
| Metadata/plugin changes | Installed schema 17 rejected using registry schema 16; unreadable override accepted; two unchanged disabled-plugin cases reported mutation. |
| Manual continuation | Two second partial migrations survived failure; one reopened database remained open. |
| Rehearsal paths | External OAuth directory and substituted WAL/SHM/journal were accepted. Final companion proof: three intended failures, two controls passed; afterward Doctor 37/37. |
| Capture overhead | Expected four recovery reads, received 28; afterward lifecycle 16/16. |
| Native recovery run handoff | Two cases returned undefined before the fix; both pass afterward. |
| Config-symlink restore | Three regressions failed while two controls passed; the complete restore suite now passes 21 tests. Scoped P1 review is clean (0.96). |

Relevant complete suites include `node scripts/run-vitest.mjs src/commands/doctor.update-recovery.test.ts` (37), `src/plugins/doctor-contract-registry.state-migrations.test.ts` (29), and `src/cli/update-cli/update-command-backup-lifecycle.test.ts` (16). Historical full CLI and plugin-update proof passed **482/482** and **219/219** before this rebase/fix round; these are not claims that the current full CLI run passes. The JSON review concern was disproved by the registered read-only lint contract and **127 passing registration/bootstrap cases**.

The typecheck-routing guard exposed 721 roots against the unchanged cap of 720. Moving 69 update-CLI roots from the config-CLI shard to the infra shard yields **652 / 704**, with every root assigned exactly once. The local routing guard exits **0**; the cap is unchanged, and the canonical guard verifies every test remains covered exactly once.

Local gate history: the latest rerun removed a rebase-introduced duplicate `updateRun` field; all core/plugin production and test typechecks passed. The gate was stopped during type-aware core lint at the ownership STOP (exit 130), so it is not a full pass. Capacity inventory of the whole Git checkout remains an owned blocker for full CLI proof. The proposed remaining-copy handoff awaits the reserved package owner's narrow exception; no guessed ignores or larger budgets are used.

### Fresh artifact and installed proof

Earlier published checkpoint: **`f956f3063467cb2477fb51cb89a5570ef839b67b`**, rebased onto `b7693bf50aa` (which contains `87504f7ec9e`). Native recovery handoff, config-symlink restoration and exact-once typecheck routing fixes are committed.

Fresh candidate packaging and the installed matrix for this head are not yet complete. Capture sizes, downtime, Gateway outcomes and tagged-driver results below belong only to their explicitly identified historical source. No current-head live success is claimed.

Historical context only: all four cells passed on `ecbcdc5e681d`; integrated `49b181b20b8c` passed 9.2/no-op, healthy update and injected rollback, but repair timed out before capture and published-9.3 rehearsal refused. Those failures led to the fixes above. Earlier fingerprint timeouts were load-induced only where quiet scans (15.121 s candidate; 13.837 s baseline) and unchanged-candidate controls confirmed it. Versioned fixtures regenerate the canonical dist-content seal; executable bytes remain unchanged.

### CI attribution

Prior pull-request CI: [run 34643012782](https://github.com/openclaw/openclaw/actions/runs/34643012782). Merge commit `536eb4b1cd62eaeccadc14499bdde85593cebd79` contains this head and required upstream `87504f7ec9e00ca080e6c6bf17899a9213f32e67`. The bounded watcher exited with owned lint failures in core-1/core-2 (execution, post-update and plugin-install files exceeded max-lines). Existing-owner extractions fixed them and passed 373 focused tests before the next rebase. No green current-head CI result is claimed.

| Previous failure | Attribution/proof |
| --- | --- |
| `docker-seed-e2e` | Owned tagged-driver rehearsal admission, reproduced locally. Latest compatibility fix awaits fresh installed/CI proof. |
| compact-small-18 | Eight owned stale fixtures; corrected Doctor mock and managed unsafe-exit expectations. Both files pass 41 tests without weakening retained-state/no-write assertions. |
| large-4/large-5 | Owned Node-only import reaching browser update reporting; before five failures, after both unchanged browser files pass 129 tests. Helpers moved to the existing driver owner. |
| compact-small-30 | Unrelated main planner: 81 jobs against cap 80. Exact reproduction on main `d1c9e4690adb`; independently #144941, run 34603439927/job 103283250640. Passed in run 34616163753. No cap change here. |

### Known limits

9.2 failures after Doctor succeeds remain outside the target's rollback contract. Doctor completion, standalone repair and no-service/no-restart finalization do not prove whole-update runtime health. Remote/uninventoried stores are not covered. Changed-core rollback without a retained package transaction remains refused, including ordinary Git updates after migration; captures remain for explicit recovery. Process exit never proves success. Inline Git RPC remains unprotected pending #144422.

Evidence retention note: the task's `finalization-delta` directory was deleted by another operation while the gate held its log open. The live-control evidence and all 24 timeout diagnostic files survive; the latter were recovered from the task container and hash-verified. The original full gate/test/review logs from that directory are unavailable; displayed results above were observed before deletion.


steipete:
Maintainer note: this PR is being re-scoped onto the lossless-capture contract accepted on #140339 (https://github.com/openclaw/openclaw/issues/140339#issuecomment-5614409997 and https://github.com/openclaw/openclaw/issues/140339#issuecomment-5614508856) so it lands as the first capture-and-restore slice rather than a parallel design.

Concretely: recovery sets move under the private sibling root `<stateDir>.update-captures/<captureId>/` with the privacy marker from #143899, so the export/backup exclusion from #143799 covers them; count-based retention goes away in favor of transaction-scoped retirement through the lifecycle owners from #143791 and #143921 (unresolved sets are never pruned, one unresolved set blocks another protected mutation); capacity admission runs before the protected mutation. Package mutation files stay reserved to #143750/#143752.

@fuller-stack-dev @vincentkoc: your prerequisites are reused, not replaced. If either of you has an in-flight capture/restore slice, say so here so we do not race; otherwise this PR is the vehicle and review is welcome once the re-scoped head is pushed.

Glucksberg:
Independent Linux evidence for one pre-Doctor race that may be worth adding to this PR's Git matrix.

On a source checkout driven by 2026.9.1, the updater stopped the managed Gateway at `2026-09-10T01:07:51Z` and then mutated/built the checkout. A host-local watchdog restarted `openclaw-gateway.service` at `01:08:17Z` while that build was in progress. The intermediate Gateway eventually started from the target checkout and logged:

```text
2026-09-10T01:09:49Z [state/db] state database schema migration pending; verifying integrity first
```

A retained SQLite backup confirms `schema_meta.primary = schema 17, app 2026.9.3, updated_at=1789002589993` (`01:09:49.993Z`). The updater's own Doctor had not begun, so its local `stateMigrationStarted` flag was still false. Its later build failure/rollback restored the 2026.9.1 source/runtime, whose Gateway then repeatedly failed with:

```text
OpenClaw state database ... uses newer schema version 17; this build supports 15.
```

The important variant is that the persistent migration was performed by an out-of-band Gateway start, not by the updater's Doctor. The Git rollback guard in both the shipped 2026.9.1 code and current main is based on the updater-local `stateMigrationStarted` boolean, so it cannot observe this race.

PR #144005 creates the recovery set before mutation, which is the right recovery material. However, I could only find that reference passed into the Git Doctor path; the pre-Doctor `build-failed` path still appears to perform its direct source/runtime rollback without consuming the recovery set.

Suggested regression/fence:

1. create the verified recovery set;
2. simulate an external process migrating live state before Git Doctor starts;
3. fail the Git build or another pre-Doctor step;
4. require either package/source + state restoration from that set, or a conservative refusal to restore the older code.

This is not a request to support arbitrary watchdog behavior. It is evidence that process-local migration bookkeeping alone is insufficient to prove a code-only rollback safe. We have fixed the local watchdog to suppress restarts while the Nurse or a live update-ledger owner is active.

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
Lead ruling on the remaining CI blocker (`docker-seed-e2e`, lane `published-upgrade-survivor-legacy-operator-state`), recorded here so the contract owner can object before it lands.

**Finding.** The shipped 2026.9.3 driver invokes the candidate's rehearsal Doctor with `OPENCLAW_UPDATE_IN_PROGRESS=1` (`v2026.9.3:src/infra/update-candidate-canary.ts:308`) but strips run identifiers (`update-candidate-rehearsal.ts:200`) and supplies no disposable-state receipt. The new target admission in `src/commands/doctor-update-recovery.ts` (around line 522) therefore refuses every rehearsal started by a 2026.9.3 driver, which would make every 2026.9.3 → next update fail, the same class as #143859.

**Ruling (first-hop rule: shipped drivers cannot be fixed, so the candidate must key on markers they already set).** Add an identity-bound legacy rehearsal admission as a named tagged-upgrade-path compatibility contract for the shipped 2026.9.3/2026.9.4 drivers only: admit when the in-progress marker is set, no identity/receipt is supplied (exactly the legacy shape; identity-bearing invocations keep the strict path), and the target state directory is provably the rehearsal copy that 2026.9.3's rehearsal produces (its temp-root env overrides, copied config path, no live coordinator/lease ownership). The live state directory stays refused. The admission is recorded as a typed fact in the run ledger and as a warning. Regression tests cover admitted-copy, refused-live-dir, and identity-bearing paths; the Docker survivor lane and the published-9.3-driver control cell must pass.

@jalehman this sits next to your #145044 ("older target-only repair workers fail closed on ambiguous authority"); I read the two as different boundaries (yours is the orchestrator↔repair-child protocol, this is the target Doctor admitting a rehearsal from an old driver whose disposable copy is verifiable), but if you see the legacy admission as widening the #140339 contract in a way you did not intend, say so here and the lane will hold before landing.

fuller-stack-dev:
## Independent no-loss rollback review and ownership boundary

@steipete: this PR remains the canonical capture/restore vehicle. I have fresh authority to complete the stricter migration-crossing journey (preserve candidate data and all newer user writes; refuse unknown changes/effects). I have not changed this PR's branch or your checkout. The existing lossless-capture/privacy and transaction-retirement decisions are reused, not reopened. Vincent's package-only recovery ownership remains separate.

Reviewed production head: `2d1d919ce8ae5c40baddbd62ae0498a8b6c918ce`, against refreshed main `7a92d3633b9dd48e8c4f7e8287b3ba22aecf4bb3`.

### Reproduced preservation gaps

Using the existing `update-recovery-backup.test.ts` fixture and real `restoreUpdateRecoveryBackup` API, with no production changes:

1. After capture, commit a newer WAL row and delete a pending delivery (acknowledgement surrogate). Restore **returns success**, removes the new row, and resurrects the deleted delivery.
2. After capturing a declared plugin directory, edit an existing record and add a new record. Restore **returns success**, reverts the edit, and deletes the new record.
3. Inject EIO on the second of two resource renames. Restore fails, but the first resource's candidate bytes no longer exist anywhere in the isolated fixture tree; the second survives and the baseline still verifies.

The first invocation produced **2 failures / 3 passing original WAL controls / 13 name-filtered tests**, exit 1. The partial-publication invocation produced **1 failure / 18 name-filtered tests**, exit 1. Native macOS, Node 24.20.0, pnpm 12.3.4. The 125-line test-only evidence patch has SHA-256 `aaf80f6444501ef1dbdf1dfbbaf26de18e41935d06d73a971a63e117dc3a40f2`; formatting and diff checks pass. These are deliberately red restoration-boundary tests, **not installed-driver, live-writer-custody, actual external-delivery, process-kill or full-journey proof**. Existing authority fixture is unchanged.

Sources: [whole-database overwrite](https://github.com/openclaw/openclaw/blob/2d1d919ce8ae5c40baddbd62ae0498a8b6c918ce/src/infra/update-recovery-backup-restore.ts#L155), [created-resource removal and directory pruning](https://github.com/openclaw/openclaw/blob/2d1d919ce8ae5c40baddbd62ae0498a8b6c918ce/src/infra/update-recovery-backup-restore.ts#L96), [capture/run/config validation, without database/plugin write provenance](https://github.com/openclaw/openclaw/blob/2d1d919ce8ae5c40baddbd62ae0498a8b6c918ce/src/infra/update-recovery-backup.ts#L198). Current CLI rollback skips `stateUnchanged` when a capture is supplied; manual Doctor uses the same restore owner after stopped-owner admission. Closing current writers does not classify already-committed newer writes. Real Workshop v16 migration also drops information and rewrites rows; it is not a reversible column rename.

### Accepted retained-generation extension, 2026-09-11

The task implementation owner will extend this PR from a separate dependent branch, preserving @steipete's capture/restore ownership and credit. No publisher branch or checkout changes; no package-only journal takeover. Current source head remains `2d1d919ce8ae5c40baddbd62ae0498a8b6c918ce`. This acceptance does not permit dropping newer user data or weakening refusal/readiness requirements.

Under the [material storage checkpoint](https://github.com/openclaw/openclaw/blob/7a92d3633b9dd48e8c4f7e8287b3ba22aecf4bb3/docs/reference/database-schemas/storage-changes.md#L164), extend this same transaction-scoped set as follows:

- Keep the original captured generation **B** immutable. Before any package/state reversal, seal the stopped current generation **C** under `<captureId>/candidate/`, using the existing lossless inventory/payload representation, complete config includes and shared/configured/registered/external-agent/plugin closure. Recheck current maintenance ownership and join all relevant children.
- Prepare an exact older-compatible generation **T** under `<captureId>/prepared/`, using the existing migration owners. Preserve newer records/acknowledgements; only proven migration-owned changes may be reversed. Already-compatible agent databases may remain current after exact old-runtime preflight. Unknown database/plugin changes, ambiguous Workshop ownership, undeclared resources, or unresolved external effects refuse the entire publication. Keeping user writes only in C is not permission to silently remove them from active state.
- Before the first canonical replacement, seal `<captureId>/publication/manifest.json`, an immutable recovery-artifact plan binding B/C/T digests, exact package/build identities, resource identities/absence facts and idempotent local operations. Reuse the existing terminal outcome. No new mutable JSON progress journal, global checkpoint manager, daemon, or extension of the package-only journal. A fresh live maintenance owner must reconcile current resources with the sealed candidate/prepared states; unknown states refuse. The plan is never authority. Preserve both generations through every interruption, and block startup into partial publication.
- Keep the existing private root/exclusion and transaction-scoped retention. Capacity includes B/C/T plus package staging, growth and reserve. No age/count/pressure pruning. Only the existing terminal owner may retire artifacts after exact compatibility, readiness, settled children, complete preservation of current user writes, and no unresolved effects. Refusal/interruption retains them.
- Initial supported migrations must name the Workshop v15→v16 and prepared-worker v16→v17 owner adapters needed by the current target, plus any plugin migration actually claimed; unknown transitions refuse. Actual released-driver proof is required. Published 9.2 failure after Doctor has returned cannot be advertised as handled without a proven continuation/bridge. macOS/Linux native real-directory support needs native proof; unproven Windows/layout/filesystem paths stay explicit.

The user explicitly accepted this **retained-generation and restartable-publication contract** on 2026-09-11 with "yes do it" in reply to the concrete approval request. This clears the storage-design checkpoint for the contract below; implementation and full-journey qualification remain in progress. I am not opening a competing capture framework or claiming the feature complete. The finite full acceptance remains actual old installed driver → target → real migration → failure → exact older-runtime compatibility, preserved newer writes, native restart/readiness, plus real interruption/resume or safe refusal.

The published-9.3 rehearsal blocker already has [your ruling](https://github.com/openclaw/openclaw/pull/144005#issuecomment-5637090464); that decision is not being requested again. Exact-head CI still reports failed `docker-seed-e2e` on run 34616163753. Source-only red tests above are a separate preservation boundary, not an explanation for that CI failure.

steipete:
@fuller-stack-dev thanks, the three reproductions are accepted as findings against head `2d1d919ce8`, and they split cleanly:

- **Findings 1 and 2 (restore returns success while discarding writes made after the capture):** restoring the pre-update snapshot is the contract, but doing it silently is not. This PR will make restore honest before it lands: every store whose post-capture writes are discarded is recorded in the run ledger and the update summary as a warning naming the store and the capture time, and the restore outcome says "restored to capture; newer writes discarded" instead of a bare success. Preserving those newer writes across the rollback (the stricter migration-crossing journey) is yours on the dependent branch, as you describe; please link it here when it exists so the landing order is explicit: this PR first, your extension on top.
- **Finding 3 (EIO on the second rename loses the first resource's candidate bytes):** that is a plain bug in the publication step and will be fixed in this PR: publication keeps every candidate resource until the whole set is placed, and a partial failure leaves both the candidate and the baseline recoverable.

The lane working this branch is also adding the identity-bound legacy rehearsal admission for the shipped 2026.9.3/2026.9.4 drivers (ruling above). Nothing about ownership or credit changes; capture/restore stays here.

steipete:
Maintainer note: this PR is being split so the proven part lands today.

After three rounds it now spans 206 files (+16,940/−1,881), conflicts with `main`, and carries three separate ownership gaps (installed 2026.9.3 plugin migrations without resource declarations, bootstrap running Doctor before capture registration, and the capacity inventory walking the whole installation). Each of those deserves its own reviewable change.

- Slice 1 (new PR, opening shortly): the current-driver capture-and-restore core on the #140339 contract — private sibling capture dir, capture before migrations, state-only restore on rollback, transaction-scoped retirement, capacity admission scoped to the capture set, honest restore warnings for discarded post-capture writes, and all-or-nothing publication.
- Slice 2: legacy first-hop admission (identity-bound rehearsal) plus shipped-plugin inventory compatibility — plugins without resource declarations are treated as having no private resources, with a recorded warning rather than a refusal.
- Slice 3: capture registration before any Doctor write during bootstrap, and the capacity inventory scoped to declared resources and state stores.

This PR stays open as the extraction source until the slices land, then closes in favor of them. Thanks to everyone who reviewed the rounds here; the findings carry over into the slices.

steipete:
Status: the capture-and-restore series is parked until the corrective release ships. Two rounds of extraction could not get a reviewable slice under a 4,000-line / 40-file cap while still passing the installed-driver cells (the reduced tree is 66 files / 6,274 lines with 57 production files), and the remaining owners (capture, restoration, candidate preservation, retirement) need a redesign to fit rather than more trimming. The reduced tree, exact counts and the pre-cut source are preserved on `steipete/upd-capture-slice1`; this PR stays open as the design record.

Two things from the last round do land separately: the missing-resource policy (a plugin migration that declares no data resources is treated as owning no private resources, with a typed warning instead of a refusal) will go into the next capture slice as its first commit, and the cell work uncovered a real update bug independent of capture: on a supported npm/systemd-user install the 2026.9.3 driver's candidate capability probe starts the candidate CLI, which attempts a shared-state schema migration before reaching the capability handler and is refused because the live Gateway owns the state directory, so the update ends as `target-native-unsupported`. That gets its own fix now.

PR #144007, opened 2026-09-10
refactor(terminal): verify closed output through writes

steipete:
## What Problem This Solves

The internal terminal writer exposed a state getter only for tests, while the before-write error test did not check whether later writes stayed suppressed. Two tests also used incomplete objects cast as Node streams.

## Why This Change Was Made

Remove the getter and verify closure through actual write behavior. The EPIPE and before-write EIO cases share a table; after the first failure, their mocks become harmless while call history remains intact. A later write must still be rejected without another hook, stream write or notification. Real stdout spies replace the two fake-stream casts, with restoration in `finally`.

The reentrant and throwing-callback cases retain their error identity, notification ordering and stream-call checks. No supported public SDK getter contract or known production consumer was found. The helper has shipped through internal package aliases, so this does not establish zero external use.

## User Impact

No intended user-visible behavior change. The internal writer exposes only the operations used by Logs, and its four retained tests protect observable closure. Removes two production lines and eight test lines.

## Evidence

All four stronger cases passed against unchanged production. The same test bytes then passed after getter removal alongside all 35 Logs CLI cases: 39 distinct cases, no skips. All 28 canonical changed-file checks, formatting, whitespace and independent review passed.

No live Gateway or external plugin execution was used for this internal cleanup.

### DIFF
--- a/packages/terminal-core/src/stream-writer.test.ts
+++ b/packages/terminal-core/src/stream-writer.test.ts
@@ -2,49 +2,41 @@ import { describe, expect, it, vi } from "vitest";
 import { createSafeStreamWriter } from "./stream-writer.js";
 
 describe("createSafeStreamWriter", () => {
-  it("signals broken pipes and closes the writer", () => {
-    let brokenPipeCount = 0;
-    const writer = createSafeStreamWriter({
-      onBrokenPipe: () => {
-        brokenPipeCount += 1;
-      },
+  it.each([
+    { code: "EPIPE", method: "writeLine", beforeWriteFails: false },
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [03657e68a7ae0284bec582dd9b825d3fd402346b](https://github.com/openclaw/openclaw/pull/144007/commits/03657e68a7ae0284bec582dd9b825d3fd402346b)
- Landed commit: [a024092e2a88754508e9993303504487488f0349](https://github.com/openclaw/openclaw/commit/a024092e2a88754508e9993303504487488f0349)

PR #144014, opened 2026-09-10
fix(cron): avoid pauses in automation delivery previews

steipete:
## What Problem This Solves

Fixes an issue where users listing automations could experience long pauses while delivery previews were generated for jobs without a directly stored route, particularly when the agent had many sessions.

## Why This Change Was Made

Delivery lookup now shares metadata reads and lazy alias indexes within one synchronous preview batch, then passes detached delivery facts to asynchronous target resolution. The existing read owners and alias-selection rules preserve target casing, thread/base fallbacks, store ownership, per-target errors, and fresh reads on the next request.

## User Impact

Full automation lists can resolve delivery previews without repeatedly loading the same session inventory for every job. No-delivery, webhook, and unresolved-owner results retain their lazy paths, and a routable earlier store prevents opening unused fallback stores.

## Evidence

- The real SQLite regression fails on the original source: 53 jobs with 65 small session rows return 7,220 session rows, exceeding a 260-row bound. It passes with the repair and confirms a same-timestamp route update is visible on the next request. The fixture uses synchronous seeding so queued maintenance cannot alter its measured workload.
- 130 focused alias, preview, context, target, and cross-room inheritance safety tests pass. The final fixture correction also passes all 21 current-origin cases; its original-source regression remains red. A separate regression proves unused fallback stores are not opened.
- In the fixed local synthetic 53-job/2,048-small-row case, warm preview time fell from 5,795 ms to 56 ms with fresh read-only connections; native returned session rows fell from 1,088,938 to 6,237 and instrumented close calls from 424 to 3. These single synthetic measurements are not a production latency guarantee or physical disk-I/O measurement.
- The full changed-file gate and subsequent fixture gates pass, including types and lint. Independent reviews are scoped-clean through P2. CI-driven fixture corrections retain the original assertions and row bound.
- [Ordinary Linux Gateway proof](https://github.com/openclaw/openclaw/actions/runs/34476282660) passed on the unchanged cron/session production code: a real build and Gateway startup, 65 sessions and 53 disabled jobs created through public CLI calls, full/compact/no-preview lists, and a revision-guarded update yielding 52 current-session previews plus one webhook preview. Scheduler-disabled checks, natural SIGTERM/CLI shutdown, process/stream/listener cleanup, and canonical Testbox lease completion all passed. Independent artifact review verified the source binding and raw CLI results. No job execution, external channel send, or model call was involved.


steipete:
@clawsweeper review

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [b608260da762dd05d46a904588eb4d97a6b801ee](https://github.com/openclaw/openclaw/pull/144014/commits/b608260da762dd05d46a904588eb4d97a6b801ee)
- Landed commit: [93e0d9ad0d3058357474e69b9312684f8a5b5971](https://github.com/openclaw/openclaw/commit/93e0d9ad0d3058357474e69b9312684f8a5b5971)

PR #144022, opened 2026-09-10
improve: reduce SQLite snapshot process startup cost

steipete:
## What Problem This Solves

SQLite snapshot processes repeatedly load parent-side launch and logging code that they do not use. In the measured update CLI suite, 5,226 synchronous snapshot processes accounted for about 85% of elapsed time, making their cold imports a significant test and command cost.

## Why This Change Was Made

Move the existing public snapshot-routing functions into the existing snapshot-source orchestration module. Keep copying, recovery, staging, adoption and the single cleanup registry in their current low-level owner. The child argument stays unchanged and moves to the shared process-entrypoint module; the pure SQLite error predicates move out of transaction logging into the existing diagnostics module.

All callers follow the owning modules. The plugin SDK façade keeps the same export names and types, and the test-only compiled library entry follows its implementation. The actual worker entry, arguments, budgets, copy algorithms, cancellation, process isolation, source preservation and ownership checks are unchanged. Two moved field reads use the existing equivalent object coercer, retiring two assertions without adding baseline allowance.

## User Impact

The measured native snapshot child starts faster without changing update, Doctor, database or plugin behavior. No flags, caches, worker pools, schema or persistence changes are introduced. Most of the 47-file diff updates imports; production is +19 net lines for the internal owner boundary, tests/support +4, and assertion baseline -1 row. No test case, assertion or timeout is removed or weakened.

## Evidence

- Real managed compiled-child A/B/A: median **110.752 / 87.944 / 114.255 ms** across three native children per leg. The original ledger case passes in every leg. The verified static dependency closure is **67 / 47 / 67 modules**, **462,705 / 307,340 / 462,705 bytes**.
- Whole wrapper time is **8.15 / 6.77 / 6.63 seconds**, including ordinary build, inspection and cleanup; this does **not** establish a whole-wrapper or full-CI speedup. Reported whole-command maximum RSS is 2,296,840,192 / 2,306,850,816 / 2,299,084,800 bytes; no memory reduction is claimed.
- Same Node 24.19.0, pnpm 12.3.4, frozen dependencies, original test source and source-only root layout. The temporary observer forwarded native arguments/options/results unchanged, stayed outside the child graph, and captured files after the child timer. It was removed before correctness checks. Every recorded process exited, generated worker directories were removed by their normal owner, and source/dependency seals matched.
- **339 focused tests passed**: snapshot copy/cancellation/worker behavior, actual SQLite transaction/coordinator and read-only callers, Doctor/Gateway/state ownership, the two source/build artifact cases, and all 16 real subprocess repair/finalize output cases. The Linux-specific live-WAL POSIX lock case passed in CI (194 ms); current-main fresh/expired/retained history rows also passed. The remaining 32 artifact cases were deliberately filtered; no other broad success is implied.

Full `pnpm build` passed in 240.97 seconds, the correctly scoped `pnpm check:changed --base a265abfecad458603578d72395a8dde6f5d3eadc` passed, and the built iMessage entrypoint profile passed (98.44 MB peak RSS; no comparative memory claim). The initial CI run exposed an omitted export in the subprocess fixture’s prepared-entrypoint substitute; it now forwards the canonical child marker while preserving prepared worker locations. The existing JSON case failed before that fix, then all 16 process cases passed. A separate documentation expectation failure was repaired upstream in #144037. Independent review has no accepted actionable P0–P2 findings; the fixture follow-up is scoped-clean through P2. Fresh full CI must pass before landing. This is not a live installed-package update or a claim that the eight-minute CI objective is complete.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3061,7 +3061,6 @@ src/infra/sqlite-readonly-location.ts	3
 src/infra/sqlite-schema-contract.ts	9
 src/infra/sqlite-snapshot.ts	2
 src/infra/sqlite-strict.ts	7
-src/infra/sqlite-transaction.ts	2
 src/infra/sqlite-user-version.ts	1
 src/infra/sqlite-wal.ts	5
 src/infra/startup-migration-checkpoint.ts	1
--- a/scripts/lib/vitest-worker-build-entries.mts
+++ b/scripts/lib/vitest-worker-build-entries.mts
@@ -68,7 +68,7 @@ export const vitestWorkerBuildEntries = {
   // The retention fixture executes the real nested QuickJS worker.
   "agents/code-mode.worker": "src/agents/code-mode.worker.ts",
   // The real ulimit fixture must import its parent before imposing a file-size limit.
-  "infra/sqlite-readonly-location": "src/infra/sqlite-readonly-location.ts",
+  "infra/sqlite-snapshot-source": "src/infra/sqlite-snapshot-source.ts",
   // Keep provider preparation in the same compiled graph as payload rendering;
   // a source-injected plugin would miss duplicated registry scope state.
   "plugins/provider-hook-runtime": "src/plugins/provider-hook-runtime.ts",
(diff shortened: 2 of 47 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [100178478fd3d63ddcfc280169586d94b4e7616b](https://github.com/openclaw/openclaw/pull/144022/commits/100178478fd3d63ddcfc280169586d94b4e7616b)
- Landed commit: [8de89effa7c2526a28f49f5fe34a42fb4728a25e](https://github.com/openclaw/openclaw/commit/8de89effa7c2526a28f49f5fe34a42fb4728a25e)

PR #144024, opened 2026-09-10
refactor: test provider HTTP policy through shared owners

steipete:
<details>

</details>

## What Problem This Solves

Provider media tests did not consistently exercise the response limits, retry rules, and polling deadlines used in production. This made regressions in shared HTTP behavior harder to catch during provider development.

## Why This Change Was Made

Run the existing provider tests through the shared JSON, retry, deadline, polling, and download-response owners, with native response fixtures and mocked network transport. Bind transport spies to the actual module namespaces at each operation so earlier SDK imports and per-test `restoreAllMocks()` calls still use the fixture transport when an operation starts; fresh response factories and explicit polling-clock advances preserve body consumption and timing behavior.

## User Impact

No user-visible runtime change. Developers get provider contract tests that exercise the shared HTTP policies, with all existing cases retained. The complete change removes 90 code lines and 97 physical lines across 13 test/support files; production LOC is unchanged.

## Evidence

- The refreshed checkout passed **397 tests in 18 suites**: all 365 provider/shared-owner cases plus 32 docs integration cases. These runs blocked unexpected external global fetches while allowing exact loopback/localhost for existing TCP checks. The rebase preserved the complete HTTP patch and all 13 changed file contents.
- Reproduced the preloaded-namespace transport bypass as a local failure under that fetch guard, then passed all 16 Runway and 17 PixVerse cases. The final full run retained the real-time Runway tee-cancellation and OpenAI socket-close checks.
- Four polling cases exercise five 5,000ms intervals and assert the request count at 4,999ms before advancing to the next poll. Three exact polling-budget cases freeze only `Date.now()` and restore it in `finally`; an advancing-clock probe reproduced all three failures before the fix and passed afterward, with clock and timer identity cleanup verified. No test cases were removed.
- The required changed-file gate and managed autoreview passed **before the rebase**, on this unchanged HTTP patch. The gate covered formatting, SDK boundaries, selected production/test type checks, lint, dead-export scans, and structural guards; review was clean through P2. After rebasing, `git diff --check` passed and all 13 source hashes still matched the reviewed patch. The 397-test integration run above is separate, fresh proof on the new base; hosted CI must independently validate the refreshed head.
- cloc 2.10 reported the same net reduction with default parsing and `--strip-str-comments`, including every fixture, timer, and transport-binding change.

Integration proof ran at `adb3c9ba909b12b92159410d9d1bb20463b9de70` on base `1670c5a1cc375f0121d4f59b17d531fa151c7434`, which includes the upstream docs fix. Shared owners and package/toolchain files are unchanged from the original review base. Hosted CI must validate the refreshed head; previous failures and their attribution remain recorded.

The facade retry adapter still supplies an instant default sleep through the existing option. These provider suites therefore do not prove default-backoff timing or cancellation during that substituted sleep; caller-supplied sleep and signals are preserved. Authentication, request-policy/header, sanitizer, and exported assertion fixtures remain mocked.

### DIFF
--- a/extensions/deepinfra/image-generation-provider.test.ts
+++ b/extensions/deepinfra/image-generation-provider.test.ts
@@ -44,12 +44,12 @@ describe("deepinfra image generation provider", () => {
   it("sends OpenAI-compatible image generation requests and sniffs JPEG output", async () => {
     const release = vi.fn(async () => {});
     const jpegBytes = Buffer.from([0xff, 0xd8, 0xff, 0x00]);
-    postJsonRequestMock.mockResolvedValue({
+    postJsonRequestMock.mockImplementation(async () => ({
       response: Response.json({
         data: [{ b64_json: jpegBytes.toString("base64"), revised_prompt: "red square" }],
       }),
       release,
-    });
+    }));
 
     const provider = buildDeepInfraImageGenerationProvider();
     const result = await provider.generateImage({
@@ -127,7 +127,7 @@ describe("deepinfra image generation provider", () => {
   });
 
   it("sends image edits as multipart OpenAI-compatible requests", async () => {
-    postMultipartRequestMock.mockResolvedValue({
+    postMultipartRequestMock.mockImplementation(async () => ({
       response: Response.json({
         data: [
           {
@@ -138,7 +138,7 @@ describe("deepinfra image generation provider", () => {
         ],
       }),
       release: vi.fn(async () => {}),
-    });
+    }));
 
     const provider = buildDeepInfraImageGenerationProvider();
     const result = await provider.generateImage({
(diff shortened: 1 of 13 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [adb3c9ba909b12b92159410d9d1bb20463b9de70](https://github.com/openclaw/openclaw/pull/144024/commits/adb3c9ba909b12b92159410d9d1bb20463b9de70)
- Landed commit: [a5d1cfeb1d4d0b5fa663691b67de13df7aeaa25a](https://github.com/openclaw/openclaw/commit/a5d1cfeb1d4d0b5fa663691b67de13df7aeaa25a)

PR #143977, opened 2026-09-10
docs(gateway,concepts,install,help): fix information-architecture findings

vincentkoc:
Structure-only pass over the 50 open `ia` (information-architecture) audit rows filed against `docs/gateway/`, `docs/concepts/`, `docs/install/` and `docs/help/`.

**No page splits, no file moves, no URL changes, no new pages.** Everything here is either an on-page structural edit or a `docs.json` navigation move.

## CODEOWNERS

Last-match-wins simulated over the full changed-file list **before** editing: **zero owned files**. `docs/gateway/security/dependency-locking.md` is the only `secops`-owned page any row touched, and its row (`r3-1679`) is satisfied by a `docs.json` nav move alone — the "leave a one-line link in the security index" half was already present at `docs/gateway/security/index.md:47`, so that file is untouched. `docs/docs.json`, `docs/vps.md` and every `docs/{gateway,concepts,install,help}` file changed here match no CODEOWNERS rule.

## Anchor proof

All 15 changed pages enumerated with the repo's own `parseDocsDocument`, before and after:

| Page | before | after | lost | added |
| --- | --- | --- | --- | --- |
| `docs/concepts/multi-user.md` | 16 | 22 | 0 | 6 |
| `docs/concepts/model-failover.md` | 37 | 37 | 0 | 0 |
| `docs/concepts/session-tool.md` | 11 | 11 | 0 | 0 |
| `docs/concepts/session.md` | 13 | 13 | 0 | 0 |
| `docs/concepts/compaction.md` | 16 | 17 | 0 | 1 |
| `docs/concepts/memory-builtin.md` | 11 | 11 | 0 | 0 |
| `docs/concepts/queue.md` | 14 | 16 | 0 | 2 |
| `docs/gateway/logging.md` | 15 | 15 | 0 | 0 |
| `docs/gateway/multi-tenant-hosting.md` | 8 | 8 | 0 | 0 |
| `docs/install/backups.md` | 13 | 13 | 0 | 0 |
| `docs/help/environment.md` | 25 | 25 | 0 | 0 |
| `docs/help/debugging.md` | 14 | 14 | 0 | 0 |
| `docs/help/testing-updates-plugins.md` | 9 | 10 | 0 | 1 |
| `docs/install/upstash.md` | 9 | 10 | 0 | 1 |
| `docs/vps.md` | 9 | 9 | 0 | 0 |

**Zero ids lost, zero collisions, 11 added.** The after-set is a strict superset on every page.

Measured, not assumed:
[...]

PR #144002, opened 2026-09-10
test(docs): expect dependency locking under release navigation

steipete:
## Problem

`test/scripts/docs-sync-publish.test.ts` pins the Release & CI route list. #143977 moved `gateway/security/dependency-locking` from Gateway & Ops into the Release process group in `docs/docs.json`, but the pinned list was not updated, so `keeps generated locale navigation aligned with English routes` fails on main and blocks every landing (for example https://github.com/openclaw/openclaw/actions/runs/34470840974/job/102850320085).

## Solution

Add the moved route to the expected release routes at its navigation position (after `reference/release-performance-sweep`).

## Impact

Main CI is green again; no navigation or docs behavior changes.

## Evidence

- `node scripts/run-vitest.mjs test/scripts/docs-sync-publish.test.ts` — 11 passed.

PR #144026, opened 2026-09-10
fix(ci): align release navigation regression coverage

steipete:
Follows #143977. Related: #130706.

## What Problem This Solves

CI fails the generated locale-navigation test after the documentation reorganization moved the existing dependency-locking page into Release & CI. The generated navigation includes the route, while the test's expected list omits it.

## Why This Change Was Made

Add that one route at its intended position in the test expectation. The navigation, generator, exact ordering checks, and localized-route assertions remain unchanged. Production code is unchanged.

## User Impact

Restores the navigation regression check to match the intended documentation structure and unblocks unrelated PR checks.

## Evidence

The exact failure was reproduced on main before this one-line repair. All 11 tests in the file, independent P2 review, and the full proportional scripts gate passed. The original CI logs from #143994 and #143999 show the same missing route; their model changes cannot cause it.

Tested commit: `61f16130ab0c45345dd39d6e00cb924113a3617f`. This is a test-only correction; no product runtime build is needed.

### DIFF
--- a/test/scripts/docs-sync-publish.test.ts
+++ b/test/scripts/docs-sync-publish.test.ts
@@ -464,6 +464,7 @@ fs.writeFileSync('package-lock.json', JSON.stringify(lock));
       "reference/full-release-validation/profiles",
       "reference/full-release-validation/evidence",
       "reference/release-performance-sweep",
+      "gateway/security/dependency-locking",
       "reference/test",
       "reference/test/local",
       "reference/test/lanes",


steipete:
Closing in favor of #144002, merged as a23560202def802a7bb2e3620f0dbba3684a39b4. I verified its main ancestry, identical one-line patch, and identical complete test-file contents against this branch. Our independent reproduction, 11 passing tests, P2 review, and scripts gate corroborate that repair.

This branch needs no second merge. The model PRs blocked by the old expectation will pick up the canonical main fix and run fresh CI.

PR #143857, opened 2026-09-10
docs(plugins): fix accuracy findings across the plugin docs

vincentkoc:
Closes the open `accuracy` audit findings scoped to `docs/plugins/` (41 rows). Every row was checked against source before a word changed; 12 are refuted with evidence rather than edited.

## Corrections where the docs understated or misstated the code

| Page | Was | Is |
| --- | --- | --- |
| `architecture-internals/runtime-helpers.md` | "OpenAI and ElevenLabs support telephony today." | 9 bundled providers implement `synthesizeTelephony` (`src/tts/tts-synthesis-support.ts:165`); the doc named 2. |
| `manifest/package-json.md` | 2 recovery cases | `isAllowedPluginRecoveryIssue` (`src/cli/plugins-install-config.ts:76-97`) permits 5. |
| `architecture-internals/load-pipeline.md` | `manifest-command-alias` only | `manifest-cli-command-owner` is live (`src/plugins/activation-planner.ts:198`) and appeared in no doc. |
| `codex-harness-reference/app-server-transport.md` | ` ```js ` + `export default` | Not a real config surface — config is JSON5 (`src/config/io.load.ts:57`) — and `js` fences are *skipped* by `check-docs-config-examples` (`src/config/docs-config-examples.ts:96`), which is why it survived. Now `json5` and validated. |
| `meeting-plugins.md` + 3 google-meet pages | `npm:@openclaw/google-meet` | The `npm:` prefix skips the official-catalog install plan and bundled fallback (`src/cli/plugin-install-plan.ts:105-178`); the package *is* in the catalog, so the bare spec is correct. |
| `manifest/surfaces.md` | "Agent Plugins 1.1" | That is an unmerged spec proposal; `src/plugins/bundle-manifest.ts:32` pins schema `1.0.0`. |

## Version scope

Named only where a release or a dated compat record exists, each cited: **2026.4.22** (guardian env removal), **2026.5.2** (Meet access-type control), **2026.8.1** (Teams/Zoom live validation — via the documented 2026.7.2→2026.8.1 mapping in `docs/releases/index.md`), **2026.8.2** (Beam named links), **2026.9.2** (legacy doctor selector), plus the `removeAfter` dates in `src/plugins/compat/deprecation-marking.ts` and `media-legacy-projection.ts`.

Everywhere else no release note exists, so the time-relative wording is deleted and present behaviour stated. No version number is invented anywhere in this diff.

## Refuted (12)

[...]

PR #144032, opened 2026-09-10
docs: scope version-locked claims across plugins, platforms, providers, and gateway

vincentkoc:
Resolves 78 rows from the accuracy pass of the documentation audit, worked as five directory batches (plugins, tools, gateway/CI, platforms/providers/web, and a small automation/diagnostics remainder).

## What changed

Most edits replace time-relative wording — "currently", "not yet", "previously", "older app" — with one of two things:

- **A release number**, where one is establishable. Every version cited here was fixed by tag containment (`git merge-base --is-ancestor <sha> v<X>`), not by walking version bumps on `main`. Two examples where that mattered: the provider `discovery` alias drop is `2026.4.26` (commit `6a5ecb955c1`, first contained by `v2026.4.26`), and the codex catalog UI is `2026.8.1`, not the `2026.7.1` the audit row suggested.
- **A present-tense restatement**, where no release record exists. `not yet placement-bound` becomes `not placement-bound`; `Older Gateways may acknowledge …` becomes the observable condition (`a Gateway that acknowledges … with no handled field`). This is deliberate: inventing a version would be worse than either, and leaving "not yet" promises a future the repository has not committed to.

Two deprecation notes are dated from the compatibility registry rather than from releases, because that registry is date-keyed (`deprecated:` / `removeAfter:` in `src/plugins/compat/`). For the same reason, two rows asking for release numbers on registry-backed notes are refuted below — converting them would contradict the source of truth.

The rest: placeholder identifiers in SDK samples that read as exported APIs are now marked as your own (`receiptImageBuffer`, `fetchAcme*`), one `clientPresentation` sample that was a bare object fragment is now a valid `registerCommand` call verified against `src/plugins/plugin-command.types.ts`, and one maintainer smoke command moved from a user page to the contributor page.

`scripts/docs-sync-publish.mjs` is a comment-only change: two "Mintlify does not currently accept" notes now carry the date the override was added.

## Outcomes by finding id

**Fixed (27):** r3-1513, r3-1523, r3-1843, r3-1847, r3-1849, r3-1856, r3-1862, r3-1924, r3-1926, r3-1928, r3-1931, r3-1957, r3-1961, r3-1987, r3-2053, r3-2082, r3-2140, r3-2299, r3-2379, r3-2537, r3-2538, r5-0027, r5-0073, r5-0105, r5-0118, r5-0144, r5-0282

**Fixed, verdict line missing from its batch report (4):** r3-1017, r3-1022, r3-1504, r3-2434 — each verified by hand against the diff at the line the row cites.

**Already satisfied on `main` (40):** r3-1593, r3-1900, r3-1902, r3-1905, r3-1909, r3-1917, r3-1919, r3-1921, r3-1937, r3-1994, r3-1998, r3-2003, r3-2036, r3-2041, r3-2045, r3-2056, r3-2060, r3-2061, r3-2360, r3-2368, r3-2383, r3-2394, r3-2402, r3-2409, r3-2415, r3-2428, r3-2430, r3-2439, r3-2443, r3-2444, r5-0005, r5-0029, r5-0036, r5-0039, r5-0049, r5-0072, r5-0077, r5-0079, r5-0115, r5-0116

Two earlier PRs (#143982, #143857) closed a large share of these. Staleness in this batch ran ~54%, well above the ~18% seen elsewhere.

**Refuted with evidence (6):**

- **r3-1781** — renaming the Railway template slug back to `openclaw-railway-template` would restore the dead link that commit `2f0592dbc65` fixed in January.
- **r3-1933** — asks for release numbers on two notes whose source of truth is date-keyed (`MARKING_DATE`, `DEFAULT_REMOVE_AFTER` in `src/plugins/compat/deprecation-marking.ts`).
- **r3-1528** — see batch notes.
- **r5-0026** — `invokeCommand` / `invokeParamsJson` are the real schema names (`src/agents/tools/nodes-tool-commands.ts:179`); there is no parameter reference page to link.
- **r5-0031** — both keys named are retired migration-only paths; the live key `host.enabled` is already documented. Adding them to the config reference would document keys validation rejects.
- **r5-0141** — the `<YYYY.M.PATCH>` angle-bracket form appears nowhere in the repository; the bare form is the convention across 11 files.

**Deferred (1):** r5-0145.

**Not attempted (12):** r3-1009, r3-1020, r3-1021, r3-1025, r3-1144, r3-1148, r3-2356, r3-2370, r3-2425, r3-2497, r3-2520, r3-2522. These were queued for this batch but not assigned to any of the five sub-batches; they stay open for a follow-up.

## Ownership

CODEOWNERS simulated last-match-wins over all 36 changed files: **no rule matches any of them**. No owned paths in this PR.

## Anchors

No heading text changed anywhere in the diff (`git diff -U0 | grep -E '^[+-]#{1,6} '` is empty), so no published ids moved and no stubs were needed. No `docs.json` or glossary change.

## Validators (post-rebase, staged tree)

`pnpm check:docs` exit 0 — format-docs 1280 files, markdownlint 1279 files / 0 issues, mdx, i18n glossary, links, config fences. `docs-link-audit --anchors`: 13488 internal links, 3 broken, all the known pre-existing `/maturity/taxonomy#windows-app-node` baseline.


vincentkoc:
Fixed in `f4acdfd8e34`. The finding is correct and the reasoning behind the error is worth recording.

I dated the SQLite cutover by tag containment — `git merge-base --is-ancestor d115fb4cf9f v<X>` — and took the first containing tag, `v2026.5.30-beta.2`, then dropped the prerelease suffix on the assumption that a stable `v2026.5.30` existed. It does not: only `v2026.5.30-beta.1` and `-beta.2` were ever published under that prefix. A version prefix can have betas and no stable counterpart, so stripping the suffix invents a release.

The first tag containing that commit with no prerelease suffix at all is `v2026.6.1` (my first pass at this also missed `v2026.5.31-alpha.1`, because I filtered for `-beta` rather than for any suffix).

Rather than cite `2026.6.1` alone, both sentences now use the form this repository already uses for a beta-only version, at `docs/reference/database-schemas/agent-schema-history.md:13` — the beta tag and the stable release named separately:

> The shared database replaced the `tasks/runs.sqlite` and `flows/registry.sqlite` sidecar stores in `v2026.5.30-beta.2`, stable from `v2026.6.1`.

`docs/automation/taskflow.md:72` carries the same correction. `pnpm check:docs` exit 0.

vincentkoc:
Correct again, and the cause is worth naming.

`v2026.5.30-beta.1` does contain `d115fb4cf9f`. My first containment query returned beta.2 as the earliest containing tag because **the local tag set was incomplete when I ran it** — I only ran `git fetch --tags` later, while checking whether a stable `v2026.5.30` existed. So the query was correct and the input was not, which is the worst combination: it produces a confident, wrong, verifiable-looking answer.

Both pages now read `v2026.5.30-beta.1`, stable from `v2026.6.1`. Fixed in `c0428ef6376`, rebased onto current main. `pnpm check:docs` exit 0.

The rule I am taking from this, for the program's brief: **fetch tags before any `merge-base --is-ancestor` dating query, and confirm the tag you land on is the earliest containing one rather than merely the first your list happened to include.**

vincentkoc:
Rebased onto current `main` to pick up #144037, which repaired the `docs-sync-publish.test.ts` break that #144021 left on `main`. That break was the cause of the two red shards here; it had nothing to do with this PR's contents. Test passes 11/11 on the rebased branch, `pnpm check:docs` exit 0.

PR #144041, opened 2026-09-10
fix(models): preserve session selections and enforce exact allowlists

steipete:
Related: #130706, #142100. Follows #143975.

## What Problem This Solves

A resolved session model could be normalized again before a reply. Model-policy checks could also authorize a different provider/model tuple through a flattened display key or provider-owned catalog equivalence. For example, `custom/*` must not grant the distinct provider `custom/team`, and an exact logical Arcee grant must not grant its wire spelling merely because both rows describe the same catalog model.

## Why This Change Was Made

Preserve writer-resolved session IDs and route typed selection, status, fallback, command, live-switch, HTTP, tool, and directive admission through the existing policy predicate for provider/model pairs. Exact grants use literal identities; wildcard checks keep the provider boundary and unchanged model bytes. Remove the obsolete string admission method. Catalog visibility remains separate from selection authority, and fallback rows must independently pass admission.

Existing raw/default normalization, model locks, CLI mapping, heartbeat/turn precedence, and permitted fuzzy fallback behavior remain intact. Config-write tests now use their existing owned temporary root for coordinator state, preserving real SQLite and fail-closed checks while avoiding unrelated shared leases.

## User Impact

Replies keep the authorized model selected for the session. Exact and wildcard grants cannot select a different tuple through display-key or catalog-equivalence collisions. When no authorized fallback exists, the existing caller error explains that no allowed model is available.

Compatibility decision: selections admitted only through a display-key collision or catalog-equivalent spelling intentionally stop qualifying. Operators who want that distinct provider/model pair must add its explicit entry to the existing model allowlist. The repair keeps normal raw-input resolution and authorized fuzzy fallback; it does not expand selection authority to preserve collision-dependent behavior. Deployment prevalence of those accidental grants is unknown.

## Evidence

- Eleven additional owner/caller regression cases fail on the frozen literal-only predecessor, covering policy/status, Arcee fallback, reply discovery, HTTP, session status, and raw/alias/partial-alias directives. Earlier resolved-selection and exact-Arcee admission baselines remain preserved separately.
- **1,033 focused tests across 18 files passed**, including the affected consumers, legacy migration/write behavior, browse, picker, reset, and worker-inference paths.
- The normal complete-PR changed gate passed across all 22 paths. Fresh staged P2 review found no actionable P0–P2 findings.
- One canonical full build of `77e043dfcefc55b86df2c23b6ef64216e3890b35` passed, including UI, SDK declarations, and SDK export checks.
- **37 compiled cases passed**: 21 reply/policy cases including the original 20 plus a real runtime-normalization hook; 12 actual-Arcee/status/fallback/wildcard/directive cases; and four HTTP/session-tool cases. The latter use real loopback HTTP requests through the emitted Gateway header helper and the real emitted tool factory with an owned SQLite session store.
- Four intentional source/cross-checkout imports were rejected; actual source imports were zero. All 11,932 files across 17 build-output roots remained unchanged. The prior 9,014-file build and frozen predecessor source/index also remained unchanged.

The compiled HTTP fixture exercises the Gateway header-validation helper, not a full Gateway startup or provider inference. No real credentials, external provider calls, CLI backend execution, or operator Gateway connections were used. Exact-head CI and native prepare must pass before landing.

Complete PR delta: **+13 net production lines, +601 test lines**. This admission follow-up removes 12 production lines relative to `9a9d2535`.

### DIFF
--- a/src/agents/agent-command.live-model-switch.test-helpers.ts
+++ b/src/agents/agent-command.live-model-switch.test-helpers.ts
@@ -195,7 +195,6 @@ export function createTestModelVisibilityPolicy(params: ModelSelectionParams) {
     providerWildcards: new Set<string>(),
     hasConfiguredEntries: !allowed.allowAny,
     hasProviderWildcards: wildcardModelKeys.size > 0,
-    allowsKey,
     allows: ({ provider, model }: { provider: string; model: string }) =>
       allowsKey(`${provider}/${model}`),
     allowsByWildcard: ({ provider, model }: { provider: string; model: string }) =>
--- a/src/agents/command/model-selection.ts
+++ b/src/agents/command/model-selection.ts
@@ -218,9 +218,7 @@ export async function resolveEmbeddedModelSelection(params: {
         overrideModel,
         params.modelManifestContext,
       );
-      if (
-        !visibilityPolicy.allowsKey(modelKey(normalizedOverride.provider, normalizedOverride.model))
-      ) {
+      if (!visibilityPolicy.allows(normalizedOverride)) {
         const { updated } = applyModelOverrideToSessionEntry({
           entry,
           selection: { provider: defaultProvider, model: defaultModel, isDefault: true },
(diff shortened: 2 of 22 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [77e043dfcefc55b86df2c23b6ef64216e3890b35](https://github.com/openclaw/openclaw/pull/144041/commits/77e043dfcefc55b86df2c23b6ef64216e3890b35)
- Landed commit: [2b5c0abe361a0d89e62d8f8a2b64eb78fbc1c7cd](https://github.com/openclaw/openclaw/commit/2b5c0abe361a0d89e62d8f8a2b64eb78fbc1c7cd)

PR #144037, opened 2026-09-10
fix(docs): align locale navigation with split release tabs

RomneyDa:
Related: #144021. Restores the documentation checks blocking #144017 and other PRs.

## What Problem This Solves

The English navigation was split into Releases and Contributing, but the Simplified Chinese overlay and publisher test still expected the combined tab. The reorganization also exposed the private QA channel in navigation despite its explicit hidden-docs setting.

## Why This Change Was Made

Split the Chinese labels to match the English layout, and validate the release and contributor route sets independently. Remove only the accidental QA navigation entry, preserving its page and existing links and keeping the catalog exposure guard intact.

## User Impact

Chinese navigation uses the matching release/contributor labels, and the documentation regression checks cover the current layout. Internal QA transport docs stay outside advertised navigation.

## Evidence

Author validation:

- docs-sync-publish and official-channel-catalog tooling suites: 32/32
- pnpm check:test-types
- pnpm lint
- pnpm format:check
- git diff --check

Maintainer verification: a fresh independent P2 review found no actionable issue. A direct JSON comparison confirmed all 294 Chinese route occurrences are preserved; the sole English navigation removal is `channels/qa-channel`, whose unchanged manifest declares `channel.exposure.docs: false`. The page and inbound links are unchanged.

Current head: `0d90f952ad90403eb1886499164b8a9c49573abc`. Its [replacement CI run](https://github.com/openclaw/openclaw/actions/runs/34477767029) passed. The actual tested merge tree and the formerly failing locale-navigation test were independently verified; the prior head's catalog failure is preserved in the review evidence. No checks or allowlists were weakened.

PR #144045, opened 2026-09-10
fix(docs): restore the localized release navigation after the tab split

vincentkoc:
`main` is red. This restores it and repairs a navigation regression that the red test was masking.

## What broke

[#144021](https://github.com/openclaw/openclaw/pull/144021) renamed the English `Release & CI` tab to `Releases` and moved the `Maturity` and `Testing and CI` groups into a new `Contributing` tab. It did not touch `docs/.i18n/zh-Hans-navigation.json`, which still described one four-group `发布与 CI` tab.

`scripts/docs-sync-publish.mjs` derives each locale's navigation from the English structure and takes labels from that file. With no localized tab for the two moved groups to land in, **they were dropped**: 57 pages — `maturity/scorecard`, `maturity/taxonomy`, and the 55 routes under `Testing and CI` — fell out of the zh navigation entirely. That is the actual regression; the failing assertion was the symptom.

## Why it landed on `main` instead of on the PR

`test/scripts/docs-sync-publish.test.ts` pinned the release navigation three ways, and all three are the shape that fails late:

- `english!.tabs.slice(-4)` pinned tab **labels**, so a rename fails the assertion rather than the thing the rename could actually break.
- `collectPages(releaseTab)` collected pages from a tab found **by label**, so moving a group between tabs silently changed what was being compared.
- `releaseRoutes` is a hard-coded list of 50 routes. It had already drifted by 18 — `#141968` split the testing guide into `help/testing/*` and `help/testing-live/*` and added them to `Testing and CI` without updating the pin. Nothing was missing from the nav; the pin was just stale, and the tab assertion failing first hid it.

The test's own comment, a few lines below, already names this failure mode for the release-notes list: *"Pinning the list makes this assertion fail on release PRs whose change classification never selects this lane, so the break first lands on main."* The same reasoning applies to the tab pins directly above it.

## What this changes

**`docs/.i18n/zh-Hans-navigation.json`** — splits the localized tab to mirror English: `发布` keeps `发布说明` and `发布流程`; a new `贡献` tab takes `成熟度` and `测试与 CI`.

The two labels are the part worth a native reviewer's eye, so to be explicit about where they come from: `贡献` is the rendering already used for "contribut*" in `docs/.i18n/glossary.zh-CN.json` (`贡献者指南`, `参与威胁模型贡献`), and `发布` is the stem shared by both groups the tab now holds. Neither is coined here, but neither is an exact attested tab label either — if a native reviewer prefers something else, that is a label-only change with no structural effect.

**`test/scripts/docs-sync-publish.test.ts`** — the assertions now follow the groups rather than the tabs:

- Tab-label pins updated to the current four, and the `Releases` / `Contributing` group memberships asserted explicitly, so a future move fails on the PR that makes it.
- The release routes are collected by group name in the order `releaseRoutes` uses, so the groups can live in any tab.
- The localized release tab is found by the group it carries, not by its label — a hand-translated label lagging an English rename should not turn into a red build.
- The per-locale loop compares route **sets** across the two tabs. Ordering within each group is still pinned by the English assertions above it.
- `releaseRoutes` synced to the 68 routes actually published. The 18 additions are all from `#141968`: `help/testing`, `help/testing/{suites,live-workflows,docker,qa-runners,contracts,writing-tests}`, `help/testing-updates-plugins`, `help/testing-live`, `help/testing-live/{quick-smokes,model-smoke,cli-backends,acp-and-codex,long-context-and-matrix,media-providers}`, `channels/qa-channel`, `concepts/mantis`, `concepts/mantis-slack-desktop-runbook`. No route was removed.

## Verification

`vitest tooling test/scripts/docs-sync-publish.test.ts` — 11/11 pass (1 failed before this change). `pnpm check:docs` exit 0. `pnpm format:check` clean over 36360 files.

### DIFF
--- a/docs/.i18n/zh-Hans-navigation.json
+++ b/docs/.i18n/zh-Hans-navigation.json
@@ -520,7 +520,7 @@
       ]
     },
     {
-      "tab": "发布与 CI",
+      "tab": "发布",
       "groups": [
         {
           "group": "发布说明",
@@ -531,16 +531,21 @@
           ]
         },
         {
-          "group": "成熟度",
-          "pages": ["zh-CN/maturity/scorecard", "zh-CN/maturity/taxonomy"]
-        },
-        {
           "group": "发布流程",
           "pages": [
             "zh-CN/reference/RELEASING",
             "zh-CN/reference/full-release-validation",
             "zh-CN/reference/release-performance-sweep"
           ]
+        }
+      ]
+    },
+    {
+      "tab": "贡献",
+      "groups": [
+        {
+          "group": "成熟度",
+          "pages": ["zh-CN/maturity/scorecard", "zh-CN/maturity/taxonomy"]
         },
         {
           "group": "测试与 CI",
--- a/test/scripts/docs-sync-publish.test.ts
+++ b/test/scripts/docs-sync-publish.test.ts
@@ -424,20 +424,26 @@ fs.writeFileSync('package-lock.json', JSON.stringify(lock));
     expect(simplifiedChinese).toBeDefined();
     expect(german).toBeDefined();
     expect(english!.tabs.slice(-4).map((tab) => tab.tab)).toEqual([
-      "Gateway & Ops",
       "Reference",
-      "Release & CI",
+      "Releases",
+      "Contributing",
       "Help",
     ]);
 
-    const releaseTab = english!.tabs.find((tab) => tab.tab === "Release & CI");
-    const releaseNotes = collectPages(releaseTab?.groups?.[0]);
-    expect(releaseTab?.groups?.map((group) => group.group)).toEqual([
-      "Release notes",
-      "Maturity",
-      "Release process",
-      "Testing and CI",
-    ]);
+    // Find the release-notes group by its own name rather than by the tab that happens to
+    // hold it. Tabs get reorganised — "Release & CI" was split into "Releases" and
+    // "Contributing" — and pinning the tab name made this assertion fail on main rather than
+    // on the PR that moved the group.
+    const releaseNotesGroup = english!.tabs
+      .flatMap((tab) => tab.groups ?? [])
+      .find((group) => group.group === "Release notes");
+    expect(releaseNotesGroup).toBeDefined();
+    const releaseNotes = collectPages(releaseNotesGroup);
+    const groupsByTab = new Map(
+      english!.tabs.map((tab) => [tab.tab, tab.groups?.map((group) => group.group)]),
+    );
+    expect(groupsByTab.get("Releases")).toEqual(["Release notes", "Release process"]);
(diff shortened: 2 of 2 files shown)


vincentkoc:
One thing this PR deliberately does **not** fix, because it is pre-existing and wants a maintainer's call on where the test belongs.

`test/scripts/docs-sync-publish.test.ts` asserts on the contents of `docs/docs.json` and `docs/.i18n/zh-Hans-navigation.json`, but it lives in a lane that docs-only pull requests skip.

`.github/workflows/ci.yml:1114` runs `detect-docs-changes`, and when it reports `docs_only`, the "Detect changed scopes" step at line 1121 is skipped along with every test lane. On [#144021](https://github.com/openclaw/openclaw/pull/144021) the rollup shows `check-docs` SUCCESS and the tooling and core test lanes SKIPPED. So a change to `docs/docs.json` cannot fail the one test that pins `docs/docs.json`, and the break necessarily lands on `main`.

That is why the failure mode the test's own comment describes — *"the break first lands on main"* — keeps happening. It has now happened at least twice in a week: once for `gateway/security/dependency-locking` (fixed by [#144002](https://github.com/openclaw/openclaw/pull/144002), also after landing) and once here.

Two ways to close it, both outside the scope of a red-`main` repair:

1. Run this test from the `check-docs` job, so it follows the files it asserts on.
2. Have `detect-docs-changes` stop treating `docs/docs.json` and `docs/.i18n/**` as inert, so a nav change selects the tooling lane.

The first is cheaper and keeps the docs-only fast path intact. Happy to open it as a follow-up if you tell me which you prefer.

vincentkoc:
Closing: superseded by #144037, which landed the same repair while this was in review. `main` is green — `docs-sync-publish.test.ts` passes 11/11 there, and the overlay carries the same `发布` / `贡献` tabs this branch added.

Two corrections to this PR's body, since the record should not stand as written.

**The 57-page loss claim was wrong.** ClawSweeper caught it and it is right. `applyLocaleNavLabelOverlay` (`scripts/docs-sync-publish.mjs:475`) clones every English tab and then only *assigns labels*; an overlay group that matches nothing is skipped, and no route is ever removed. The actual regression was that the two moved groups and the new tab fell back to **English labels** in the localized sidebar. A visible i18n regression, not content loss. I inferred the loss from a failing assertion about group names without reading the function that produces them, which is the same mistake as trusting a row's suggested fix without checking its premise.

**The route count was ambiguous.** "68 routes" was the pinned tail; the full list including release notes is 76.

The one thing here that main does not have is the assertion style: #144037 re-pinned the tab labels, English and Chinese, so the next rename fails on `main` again rather than on the PR that makes it. I am not reopening that as a competing change to a file that was just repaired — and it is secondary to the real gap, which is that this test cannot run on the PRs that change the files it guards (detail in the comment above). That is the fix worth making, and it wants a maintainer's decision on where the test belongs.

PR #144046, opened 2026-09-10
fix(gateway): report busctl-incompatible when busctl rejects --json

### LINKED ISSUE #142940: [Bug]: openclaw gateway install` fails with `SERVICE_DEFINITION_UNKNOWN` on systemd 239 (busctl `--json=short` unsupported)
### Bug type

### Beta release blocker

No

### Summary

`openclaw gateway install` (and `--force`) consistently fails with:
```
SERVICE_DEFINITION_UNKNOWN: Service definition cannot be safely inspected.
```
even after ruling out every documented cause (directory permissions, ACLs, umask, systemd user session availability, SELinux).

### Steps to reproduce

`openclaw gateway install` (and `--force`) consistently fails with:
```
SERVICE_DEFINITION_UNKNOWN: Service definition cannot be safely inspected.
```
even after ruling out every documented cause (directory permissions, ACLs, umask, systemd user session availability, SELinux).

### Expected behavior

- Detect `busctl`/systemd version and fall back to a compatible invocation (e.g. drop `--json=short` and parse the legacy text/table output, or use `systemctl --user` subprocess calls instead of raw D-Bus/busctl) on systemd < 240.
- At minimum, when the `busctl` subprocess itself fails due to an unrecognized-option/parse error (as opposed to a legitimate "unit not found" or permission error), surface a clearer error distinguishing "inspection tool incompatible with this systemd version" from the generic `SERVICE_DEFINITION_UNKNOWN`.

## Root cause found via `strace`
Tracing the install command (`strace -f -e trace=execve`) shows OpenClaw does **not** shell out to `systemctl` for the service-definition inspection step — it calls `busctl` directly against the systemd D-Bus API:

```
execve("/usr/bin/busctl", ["busctl", "--user", "--json=short", "call",
  "org.freedesktop.systemd1", "/org/freedesktop/systemd1",
  "org.freedesktop.systemd1.Manager", "LoadUnit", "s",
  "openclaw-gateway.service"]) = 0
... exited with 1
```

Running that command manually reproduces the failure directly:
```
$ busctl --user --json=short call org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager LoadUnit s openclaw-gateway.service
busctl: unrecognized option '--json=short'
```

`--json=short` was added to `busctl` in **systemd 240**. This host ships **systemd 239** (still fairly common on CentOS 8 / Alibaba Cloud Linux 8 / RHEL 8-family distros, which are widely used on cloud VPS providers). Because the `busctl` call itself fails to even parse its arguments, OpenClaw's install-time inspection can't distinguish "unit not found" (expected, normal on fresh install) from "tool invocation failed," and conservatively refuses with `SERVICE_DEFINITION_UNKNOWN` rather than proceeding.

## What I ruled out first (in case it's useful for triage)
- `~/.config`, `~/.config/systemd`, `~/.config/systemd/user` — permissions fixed to `755`, no group-write, no unexpected ACLs (`getfacl` clean)
- `umask` was `0002` (caused the original group-writable dirs) — fixed to `022`, confirmed new dirs are `755`
- `systemctl --user status` — user systemd instance running fine, `XDG_RUNTIME_DIR` correctly set
- `getenforce` → `Disabled` (no SELinux interference)
- `lsattr` → no immutable flags
- `systemctl --user cat/is-enabled/show openclaw-gateway.service` all behave normally (unit simply not found, as expected pre-install)
- Manually creating the unit file and running `systemctl --user daemon-reload && systemctl --user enable --now openclaw-gateway.service` works fine — the actual systemd user session and unit mechanics are healthy. Only OpenClaw's own `busctl` inspection call is broken.

### OpenClaw version

2026.9.3 (1391f7c)

### Operating system

Alibaba Cloud Linux 8 (CentOS 8 family)

### Install method

npm global (`/usr/lib/node_modules/openclaw`, root-owned)

### Model

openai/gpt-5.6-sol

### Provider / routing chain

openclaw --> openai/gpt-5.6-sol

_No response_

### Logs

```shell

```

_No response_

### Impact and severity

_No response_

### Additional information

_No response_

JamieJ5926:
Fixes #142940 (minimum remediation path from the issue).

## What Problem This Solves

On hosts with systemd older than 240, busctl has no --json flag, so every manager inspection subprocess fails at argument parsing. Install conflated that tool-invocation failure with an uninspectable definition and refused with generic SERVICE_DEFINITION_UNKNOWN, leaving systemd 239 users (RHEL 8 family, Alibaba Cloud Linux 8) with no actionable error.

## Change

- Adds throwIfBusctlJsonUnsupported guard plus BUSCTL_JSON_UNSUPPORTED marker (src/daemon/systemd-unavailable.ts). Native stderr never leaves the boundary; only the static classification maps to the reason.
- The manager query path throws the branded marker instead of generic unavailable when busctl rejects --json (src/daemon/systemd-service-files.ts).
- New busctl-incompatible service-definition reason with remediation: upgrade systemd to 240+, or manage the user unit with systemctl from the service account (src/daemon/service-types.ts).
- Definition-mutation capability maps the marker to the new reason (src/daemon/systemd-definition-mutation.ts).
- Initial install read preserves the allowlisted reason instead of the generic refusal, covering both ordinary install and --force before any write (src/cli/daemon-cli/install.ts).

No persisted data model changes: the marker and reason exist only in memory during inspection.

## Evidence

- src/daemon/systemd-definition-mutation.test.ts: busctl unrecognized-option stderr yields {kind unknown, reason busctl-incompatible}, redacts diagnostics, and stage() fails with SERVICE_DEFINITION_UNKNOWN: [busctl-incompatible].
- src/daemon/systemd-unavailable.test.ts: guard-level cases (branded throw, unit-absence passthrough, non-exit passthrough).
- src/cli/daemon-cli/install.test.ts: initial readCommand rejection with the marker fails with SERVICE_DEFINITION_UNKNOWN: [busctl-incompatible] before config reads or credential generation.
- src/daemon/systemd.test.ts (242 passed), systemd-definition-mutation (85 passed), install (42 passed) suites green locally.
- Real Linux Gateway install proof (ordinary install and --force on systemd 239 with redacted terminal output) is not available from this machine (macOS, no systemd) and is parked for a Linux host run.

## Scope note

This is the issue's sanctioned minimum: install still refuses on systemd < 240, but now with an actionable error instead of a generic one. A full legacy-output fallback (parsing pre-JSON busctl text) needs a systemd 239 fixture host and is left for a follow-up.

### DIFF
--- a/src/cli/daemon-cli/install.test.ts
+++ b/src/cli/daemon-cli/install.test.ts
@@ -2,6 +2,7 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { GatewayServiceCommandConfig } from "../../daemon/service.js";
 import { mockSystemAccountHome } from "../../daemon/service.test-helpers.js";
+import { BUSCTL_JSON_UNSUPPORTED_CODE } from "../../daemon/systemd-unavailable.js";
 import type { ResolvedGatewayAuth } from "../../gateway/auth.js";
 import { captureFullEnv } from "../../test-utils/env.js";
 import { createCliRuntimeCapture } from "../test-runtime-capture.js";
@@ -352,6 +353,20 @@ describe("runDaemonInstall", () => {
     expect(readConfigFileSnapshotMock).not.toHaveBeenCalled();
     expect(randomTokenMock).not.toHaveBeenCalled();
     expect(service.readCommand).toHaveBeenCalledOnce();
+  });
+  it("reports busctl-incompatible when the initial read hits old busctl", async () => {
+    const unsupported: NodeJS.ErrnoException = new Error(
+      "busctl does not support --json output on this host.",
+    );
+    unsupported.code = BUSCTL_JSON_UNSUPPORTED_CODE;
+    service.readCommand.mockRejectedValueOnce(unsupported);
+    await runDaemonInstall({ json: true, force: true });
+    expect(actionState.failed[0]?.message).toContain(
+      "SERVICE_DEFINITION_UNKNOWN: [busctl-incompatible]",
+    );
+    expect(actionState.failed[0]?.message).toContain("systemd to 240 or newer");
+    expect(readConfigFileSnapshotMock).not.toHaveBeenCalled();
+    expect(randomTokenMock).not.toHaveBeenCalled();
   });
 
   it("blocks non-default install identities before inspecting host services", async () => {
--- a/src/cli/daemon-cli/install.ts
+++ b/src/cli/daemon-cli/install.ts
@@ -28,6 +28,7 @@ import {
 } from "../../daemon/service-types.js";
 import { resolveGatewayService, type GatewayServiceCommandConfig } from "../../daemon/service.js";
 import { isNonFatalSystemdInstallProbeError } from "../../daemon/systemd-exec.js";
+import { BUSCTL_JSON_UNSUPPORTED_CODE } from "../../daemon/systemd-unavailable.js";
 import { resolveGatewayAuth } from "../../gateway/auth.js";
 import {
   defaultGatewayBindMode,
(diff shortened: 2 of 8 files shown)


JamieJ5926:
Maintainer note on the remaining red (no code action requested from me): head 7fda6f73 shows 163 pass / 1 fail. The single fail is checks-node-changed-extensions-bundle-4 in run 34491603625 (job 102919819301): a timing-flake assertion (vi.waitFor release-called-once) in extensions/openrouter/music-generation-provider.test.ts, 1284/1285 passing in that shard. This PR's full diff is 8 files, all under src/daemon/* and src/cli/daemon-cli/* (install.ts, install.test.ts, service-types.ts, systemd-definition-mutation.ts, systemd-definition-mutation.test.ts, systemd-service-files.ts, systemd-unavailable.ts, systemd-unavailable.test.ts) — none under extensions/*. The consequential openclaw/ci-gate fail follows the shard fail. I cannot rerun it (needs admin: gh run rerun --failed was denied). A shard rerun should clear it.

PR #144054, opened 2026-09-10
fix(test): avoid update ledger worker shutdown races

steipete:
## What Problem This Solves

Resolves an intermittent CI failure in the concurrent CLI/Gateway update-ledger test when a worker exits during shared-state cleanup. The failure was observed in [CI job 102866134156](https://github.com/openclaw/openclaw/actions/runs/34475740237/job/102866134156).

## Why This Change Was Made

One worker could close its database while the other was still writing or retiring its handle, so the lifecycle coordinator correctly refused cleanup. Each worker now acknowledges completed writes before the parent requests and joins their closes sequentially. Writes still run concurrently. The readiness wait also observes write failures immediately, preserving early child errors and joining both workers during failure cleanup.

## User Impact

More reliable update-ledger CI coverage. All 32 recorded-step assertions and the final update outcome assertions remain. Production database behavior and timeouts are unchanged.

## Evidence

- After the patch-identical rebase onto the landed docs fix #144037, 79 integration cases passed: 37 update-ledger, 10 lifecycle coordinator, and 32 official-channel catalog/docs-sync cases. The existing retirement-refusal cases remain intact.
- A real-child early-exit probe reproduced an unhandled rejection before the readiness correction; afterward the same child error propagated with zero unhandled rejections and joined cleanup.
- Managed P0–P2 review of the unchanged patch: no actionable findings.
- Before the patch-identical rebase, the full changed-file gate passed, including core test typechecks, dead-export scans, and targeted lint. Both changed source files and their owner/toolchain inputs are unchanged on the new base.

### DIFF
--- a/src/infra/update-run-ledger.process.test-support.ts
+++ b/src/infra/update-run-ledger.process.test-support.ts
@@ -35,7 +35,10 @@ process.once("message", () => {
       options,
     );
   }
-  closeOpenClawStateDatabase();
-  process.disconnect?.();
+  process.once("message", () => {
+    closeOpenClawStateDatabase();
+    process.disconnect?.();
+  });
+  process.send?.("written");
 });
 process.send?.("ready");
--- a/src/infra/update-run-ledger.test.ts
+++ b/src/infra/update-run-ledger.test.ts
@@ -737,7 +737,21 @@ describe("update run ledger", () => {
           code === 0 ? resolve() : reject(new Error(`${role} exited ${code}: ${output}`)),
         );
       });
-      return { child, ready, exited };
+      const written = ready.then(() =>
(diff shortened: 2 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [25c4a14958fb5d9541887d188ab4b4ca45d3efa8](https://github.com/openclaw/openclaw/pull/144054/commits/25c4a14958fb5d9541887d188ab4b4ca45d3efa8)
- Landed commit: [d9f1f50e074bf3d963b5b100cd7da8c01ba4cc88](https://github.com/openclaw/openclaw/commit/d9f1f50e074bf3d963b5b100cd7da8c01ba4cc88)

PR #144055, opened 2026-09-10
fix(auto-reply): suppress a text-only final already delivered as a caption (#144006)

### LINKED ISSUE #144006: [Bug]: explicit tts tool causes double delivery on Telegram
### Bug type

### Beta release blocker

No

### Summary

When the tts tool is invoked during a conversation turn on Telegram, the system delivers two duplicate messages simultaneously:

A Voice Message with the synthesized audio AND the full turn reply text attached as its caption.
A Standalone Text Message (sendMessage) containing the exact same text content duplicated.

This leads to redundant text delivery where the user receives the same response twice (once as the voice note caption, once as a separate text bubble).

### Steps to reproduce

Configure Telegram channel and Microsoft / Edge TTS provider in openclaw.json.
Disable Auto-TTS (tts.auto: off).
Send a user prompt requesting voice output (e.g. "Please respond with voice" or "Generate a voice response").
The agent triggers the tts tool with speech text and completes the turn with closing text in <final>.
Observe the Telegram chat output.

### Expected behavior

The agent should either:

Deliver only the Voice Note (with or without caption), suppressing the subsequent redundant text sendMessage; OR
Deliver the Voice Note as audio and only send supplementary text if explicitly required, without repeating the exact same text twice.

### Actual behavior

The user receives two consecutive bubbles in Telegram at the exact same timestamp:

Bubble 1: Voice audio player + full text response as caption.
Bubble 2: Standalone text message with identical full text response.

### OpenClaw version

2026.9.3 (1391f7c) (upgraded from 2026.9.1)

### Operating system

Rocky Linux 10.1 x86_64

### Install method

_No response_

### Model

google/gemini-3.7-flash

### Provider / routing chain

Primary: google/gemini-3.7-flash (direct, no proxy) Fallbacks: disabled (user model override active for this session) TTS Provider: Microsoft Edge TTS (via talk-voice plugin)

_No response_

### Logs

```shell
Physical Evidence & Diagnostic Traces
1. Gateway Runtime Logs (~/.openclaw/logs/gateway.log)
log
2026-09-10T19:14:32.898+08:00 [telegram] Inbound message telegram:chatId=<redacted> -> @<bot> (messageId: 48084)
2026-09-10T19:14:37.190+08:00 [auto-reply/agent-turn-timing] milestone=tool_execution_started (calling tts)
2026-09-10T19:14:51.014+08:00 [telegram] outbound send ok accountId=default chatId=chatId=<redacted> messageId=48086 operation=sendMessage deliveryKind=text chunkCount=1
2. Media Store Artifacts (~/.openclaw/media/tool-speech-synthesis/)
-rw-r--r--. 1 user user 74736 2026-09-10 19:14:42 voice---31d8048c-c55e-4338-bb8c-b7a12634ecbe.mp3
3. Session History Trace

The turn transcript confirms that tts tool auto-delivered the audio payload with caption, but the pipeline subsequently dispatched the assistant's <final> text block through the standard text reply dispatcher without deduplication
```

_No response_

### Impact and severity

- **Affected**: Any Telegram user invoking the explicit `tts` tool with Auto-TTS disabled on OpenClaw 2026.9.3.
- **Severity**: Annoying but does not block workflow. No data loss or security risk.
- **Frequency**: Always reproducible — every explicit `tts` tool call results in double delivery.
- **Consequence**: User receives duplicate messages on every voice response — one voice note with caption, one redundant standalone text bubble. Increases message noise and may confuse users expecting a single reply. No missed messages or extra provider cost observed.

### Additional information

- Initially suspected as a 2026.9.3 regression, but confirmed present in 2026.9.1 as well.
- This appears to be a long-standing issue rather than a regression.

JamieJ5926:
Closes #144006.

## What Problem This Solves

With an explicit `tts` tool call on Telegram (`tts.auto: off`), the reporter sees the reply text twice: once as the voice note's caption and once as a standalone `sendMessage`.

An explicit `tts` tool result queues audio media (`src/agents/tools/tts-tool.ts:80-94` → `queuePendingToolMedia`, `src/agents/embedded-agent-subscribe.handlers.tools.results.ts:436-475`). The next non-reasoning assistant reply absorbs that media (`consumePendingToolMediaIntoReply`, `src/agents/embedded-agent-subscribe.handlers.messages.replies.ts:89-147`), so the block that goes to the channel carries **text + audio**, and a channel that renders media text as a caption (Telegram sets `captionedFinalText: true`, `extensions/telegram/src/setup-plugin.ts:40-42`) delivers that reply text once as the caption.

The same turn later produces a text-only final payload. `buildReplyPayloads` (`src/auto-reply/reply/agent-runner-payloads.ts`) is supposed to drop finals whose content a block already delivered, but only **media-bearing** finals consulted the recorded direct-send evidence:

- `preserveUnsentMediaAfterBlockSend` handled media-less payloads with `blockReplyPipeline.hasSentPayload` only, never `hasDirectlySentText`.
- `preserveDirectlyUnsentPayload` returned media-less payloads untouched (`if (!reply.hasMedia || !reply.trimmedText) return payload`), so the branches that drop already-sent payloads never reached the text comparison.

Because `createBlockReplyContentKey` includes the media list, a text-only final can never match the content key of the media-bearing block, so the exact-key path cannot cover it either. The result is the reported second bubble.

The fix makes media-less finals consult the same recorded delivery evidence as media-bearing ones, gated on terminal content so reasoning, commentary, status, error and fallback payloads keep their existing behavior. Text that was never delivered still ships; the audio is untouched.

## Evidence

`src/auto-reply/reply/agent-runner-payloads.test.ts` gains a table modelled on the reported shape: a block that delivered `Spoken answer.` as a captioned voice block, followed by a text-only final with identical text.

Failing first (before the fix), same file:

```
FAIL  src/auto-reply/reply/agent-runner-payloads.test.ts > buildReplyPayloads media filter integration > suppresses a text-only final already delivered as a captioned voice block
FAIL  src/auto-reply/reply/agent-runner-payloads.test.ts > buildReplyPayloads media filter integration > suppresses a text-only final already delivered as a captioned voice block while streaming
AssertionError: expected [ { text: 'Spoken answer.', …(6) } ] to have a length of +0 but got 1
 Tests  2 failed | 89 passed (91)
```

Passing after the fix, same file:

```
Test Files  1 passed (1)
      Tests  91 passed (91)
```

Command used for both runs:

```
PATH=/opt/homebrew/bin:$PATH node scripts/run-vitest.mjs run --config test/vitest/vitest.auto-reply-reply.config.ts src/auto-reply/reply/agent-runner-payloads.test.ts
```

The same table keeps a final whose text was never delivered and keeps an error final. Neighbouring dedupe and pipeline suites stay green (`reply-delivery`, `reply-payloads-transport-dedupe`, `block-reply-pipeline`: 69 passed), and the whole `src/auto-reply/reply` project ran at `5651 passed, 1 failed` where the single failure is `stage-sandbox-media.scp.test.ts` timing out at 120s — that test spawns a real `scp` process tree, never imports this module, and is unrelated to this change.

`node_modules/.bin/oxlint` on both changed files reports nothing.

## Not Verified

Which runtime entry produced the reporter's turn (streaming on/off, whether the captioning block came through the block pipeline or a direct send) is **not** confirmed by a live Telegram reproduction — that needs Telegram credentials and is out of scope here. The fix closes the shared defect that lets already-delivered caption text be sent again as a text-only final; the dedupe branch it exercises is exactly the one that let a media-less final survive.

### DIFF
--- a/docs/concepts/model-providers/official-provider-plugins.md
+++ b/docs/concepts/model-providers/official-provider-plugins.md
@@ -151,7 +151,14 @@ Claude CLI reuse (`claude -p`) is a sanctioned OpenClaw integration path. Anthro
 ### Google Vertex and Gemini CLI runtime
 
 - `google-vertex`: managed Google Cloud access through gcloud Application
-  Default Credentials.
+  Default Credentials. Set `GOOGLE_CLOUD_PROJECT` (or `GCLOUD_PROJECT`) and
+  `GOOGLE_CLOUD_LOCATION` in the Gateway service environment; without a project
+  the run fails with `Vertex AI requires a project ID`. The ADC path requires
+  the literal credential value `gcp-vertex-credentials` — store it with
+  `openclaw models auth paste-api-key --provider google-vertex`. Any other
+  value, including a token from `gcloud auth print-access-token`, is sent as
+  `x-goog-api-key` and rejected by Vertex with `401 UNAUTHENTICATED`. See
+  [Google (Gemini)](/providers/google).
 - `google-gemini-cli`: optional local runtime for an explicitly configured
   canonical `google/*` model.
 
--- a/docs/providers/google.md
+++ b/docs/providers/google.md
@@ -11,7 +11,7 @@ The Google plugin provides access to Gemini models through Google AI Studio, plu
 - Provider: `google`
 - Auth: `GEMINI_API_KEY` or `GOOGLE_API_KEY`
 - API: Google Gemini API
-- Managed-cloud provider: `google-vertex` with Google Cloud Application Default Credentials
+- Managed-cloud provider: `google-vertex` with Google Cloud Application Default Credentials; the stored credential must be the literal value `gcp-vertex-credentials`
 - Optional runtime: `agentRuntime.id: "google-gemini-cli"` runs an explicitly configured model through the local Gemini CLI
 
 ## Getting started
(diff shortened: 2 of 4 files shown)


JamieJ5926:
Superseded by #144079. This branch's head picked up two unrelated docs commits from a shared checkout, so the diff against main no longer contained only the fix. The replacement branch `oc-fix-144006-tts-dedup-v2` is cut from main e6031c1f and carries only the fix commit (851702c4). Closing this PR; no force-push was used.

PR #144056, opened 2026-09-10
fix(gateway): re-drive self-admitted channel accounts after a terminal handshake instead of pinning them blocked

### LINKED ISSUE #138947: [Bug]: openclaw doctor [--fix] silently blocks QQ and WeCom channel accounts; recovery requires manual channels.start RPCs
## Summary

Running `openclaw doctor` (or `openclaw doctor --fix`) silently breaks the QQ (`@tencent-connect/openclaw-qqbot`) and WeCom (`@wecom/wecom-openclaw-plugin`) channel accounts. After the command exits, both channels report a `blocked` runtime status ("Channel runtime is blocked and needs operator action.") and stop admitting inbound messages until the operator manually runs:

```
openclaw gateway call channels.start --params '{"channel":"wecom"}'
openclaw gateway call channels.start --params '{"channel":"qqbot"}'
```

This affects every external plugin channel where the plugin owns a `gateway.startAccount` hook (not just QQ/WeCom, but at minimum these two are the noisy cases). The same pattern does not happen on bundled channels (Discord/Slack/WhatsApp), which auto-recover from hot reload.

## Repro

1. OpenClaw `2026.9.1 (ad6fe23)`, macOS 26.6.2 arm64, Node v24.20.0, Homebrew global npm install.
2. `@tencent-connect/openclaw-qqbot@2.0.1` and `@wecom/wecom-openclaw-plugin@2026.7.2` installed as managed projects under `~/.openclaw/npm/projects/`. Both channels report healthy and connected.
3. Run `openclaw doctor --yes` (or interactively accept the recommended repairs).
4. Observe: `openclaw channels status` shows `qqbot` and `wecom` accounts in a `blocked` health state. `openclaw channels status --probe` surfaces the issue:
   > Channel runtime is blocked and needs operator action.
   > fix: resolve the reported channel error, then restart the channel
5. Inbound messages are dropped silently — no inbound events admitted, no session created, no reply. `openclaw status` still reports both channels `OK · configured`.
6. Run the two `channels.start` calls above. Both channels return to running and admit traffic again.

This reproduces 100% on my host on every `openclaw doctor --yes` invocation. No secrets, no hostnames, no account ids included.

## Expected vs Actual

**Expected:** `openclaw doctor [--fix]` is a config-only operation. Writing the candidate config and triggering hot reload should leave channel runtime state equivalent to a normal `openclaw gateway restart` — every previously-healthy channel should re-enter a running state, or doctor should refuse to write and tell the operator which channels will need explicit restart.

**Actual:** Doctor writes the config, hot reload stops and internally re-spawns the QQ/WeCom channel processes, but the channel runtime stays in `blocked` state with no operator-visible cause. Inbound messages are silently dropped. Recovery requires the operator to know and run two `channels.start` RPCs that doctor does not mention.

## Root cause (from source review)

There are three layers that compound:

### 1. `doctor` mutates the `channels.*` subtree even with no obvious repair

`doctor-config-flow-1ei9na0g.js` runs these on every invocation, not gated by `--fix`:

- `seedMissingDefaultAccountsFromSingleAccountBase` (lines ~234-289) — promotes flat `channels.<id>.dmPolicy / allowFrom / groupPolicy / groupAllowFrom` into `channels.<id>.accounts.default.*`.
- `applyChannelDoctorCompatibilityMigrations` — imported from `io.runtime-CKu7Rw0j.js`.
- `migrateLegacySecretRefEnvMarkers` — may rewrite SecretRefs anywhere under `channels.*`.
- `normalizeLegacyBrowserConfig`, `normalizeLegacyTalkConfig`, etc.

These are reasonable normalizers, but combined with `finalizeDoctorConfigFlow` (lines 67-90) which writes the candidate whenever `pendingChanges > 0`, the operator's "yes" applies a config diff even on a "just lint" run. This is the trigger that flips the channel reload plan.

### 2. Hot reload stops but does not invoke plugin `startAccount`

`server-reload-handlers-DblQx0j4.js` lines 320-333 (function around `restartChannelAccount`):

```js
params.logChannels.info(`restarting ${channel} channel`);
if (!channelsStoppedBeforePluginReload.has(channel))
  await params.stopChannel(channel, void 0, { manual: false });
// … then the reload planner restarts the channel account via the
// gateway's internal restart path, which for external plugins is just
// re-spawning the plugin process — NOT invoking plugin.gateway.startAccount
```

For bundled channels this is sufficient — the SDK reconnects itself. For external plugins where the plugin owns the lifecycle (QQ's `qqbot-connector` websocket handshake, WeCom's webhook URL re-registration in `wecom-openclaw-plugin/dist/src/webhook/gateway.js`), the plugin's `startAccount` is the only path that completes the auth/registration handshake. The internal restart path stops short of calling it.

### 3. The resulting status is `blocked`, not `stopped`

`channels-status-issues-Ci7xtXFL.js` lines 41-46 define the `blocked` health reason:

```js
case "blocked":
  message = "Channel runtime is blocked and needs operator action.";
  fix = "resolve the reported channel error, then restart the channel";
```

This is a deliberately operator-actionable state, but doctor does not surface it at the end of its run. The operator only discovers it via a separate `channels status` invocation, if at all.

## Why bundled channels (Discord/Slack/WhatsApp) are unaffected

Their plugins either auto-reconnect in-process or are restarted by the same internal path that already drives startup. They do not need an external `startAccount` callback because the SDK handshake is internal to the plugin and re-runs on every spawn. The external plugin contract (where the plugin owns a `gateway.startAccount` hook that does external registration) is the missing case in the reload path.

## Suggested fixes (in priority order)

1. **Make `server-reload-handlers` call `plugin.gateway?.startAccount` after stopping/restarting any external plugin channel.** This is the surgical fix — it brings external plugin channel recovery in line with bundled channel behavior. Channel-specific tests should cover both bundled and external plugin channels on hot reload.
2. **Have `doctor` emit a final "channels that may need explicit restart" warning** when the reload plan touches any external plugin channel. Even before (1) ships, surfacing the issue at the end of doctor closes the "silent breakage" gap.
3. **Auto-recover via `channels.start` after doctor finishes**, scoped to channels that landed in `blocked` health within a short grace window after the doctor's config write. Should be opt-out-able for operators who want to do it manually.

(2) is the smallest change with the highest operator-impact win. (1) is the principled fix.

## Workaround

```bash
openclaw doctor --yes && \
openclaw gateway call channels.start --params '{"channel":"wecom"}' && \
openclaw gateway call channels.start --params '{"channel":"qqbot"}'
```

## Environment

- OpenClaw `2026.9.1 (ad6fe23)`, macOS 26.6.2 arm64, Node v24.20.0, Homebrew global npm install.
- Affected plugins: `@tencent-connect/openclaw-qqbot` (cohort-pinned to `2.0.1`), `@wecom/wecom-openclaw-plugin` (cohort-pinned to `2026.7.2`).
- Cross-references: this is unrelated to the cohort-version skew tracked in #135776; the plugin versions in question here are exactly what OpenClaw wants installed, so the channel reload path should handle them.

Finn763:
## What Problem This Solves

`openclaw doctor --fix` (and any hot reload) can silently leave externally-provided channel accounts — QQ, WeCom, and every other plugin-owned channel — pinned in

`lifecycle: "blocked"` / `terminalDisconnect: true`:

```


Channel runtime is blocked and needs operator action


auto-restart skipped, terminal disconnect


```

The account is not running, nothing retries it, and the health monitor reports `skipping restart`. Plain `doctor --fix` never recovers it — only explicit

`channels.start` RPCs (`manual: true`) bring the account back. Bundled channels hide the bug because the plugin reconnects inside the gateway process, so the same

gateway path is never exercised.

**Root cause.** Hot reload *does* stop and re-admit the affected accounts, and `startChannel` *does* invoke the plugin's own `gateway.startAccount`. The defect is in

how the result is judged: on re-handshake, an external plugin channel collides with its own not-yet-released previous session, so the replacement instance produces a

terminal verdict (`channelBlockedPatch` → `lifecycle: "blocked"` + `terminalDisconnect: true`) **without ever having become ready**. `createChannelManager`'s task

finalizer treats *any* terminal verdict as "needs operator action" and pins the account blocked; the health monitor then refuses to restart blocked /

terminal-disconnect accounts.

**Fix.** A replacement instance that was admitted by the gateway itself, whose predecessor was recently healthy (running, or connected within the last 5 minutes), and

that terminated before its first `ready`, now falls through to the existing bounded crash-restart supervisor (backoff retries, gives up after `MAX_RESTARTS`) instead

of being pinned blocked. Accounts with no healthy predecessor — and accounts that had already reached `ready` — keep the existing pin-blocked semantics, so a

genuinely broken channel still asks for an operator.

## Evidence

- `src/gateway/server-reload-channel-restart.test.ts` (extended, +178): `reload replacement was not re-driven after a terminal handshake` is red before the

  fix, green after; `7 passed` (5 existing + 2 new). The two cases live in the existing reload/restart suite rather than a new file because the core test

  graph is capped at 720 roots and a standalone file pushed `gateway-root` to 721 — `scripts/check-tsgo-core-boundary.mts` exits 0 with the cases folded in.

- `src/gateway/server-channels.test.ts`: `146 passed`.

- `src/gateway/channel-health-monitor.test.ts` + `src/gateway/channel-health-monitor.clock-rollback.test.ts`: `58 passed`.

- All run with `--maxWorkers=1 --no-file-parallelism`.

- Diff surface: `src/gateway/server-channels.ts`, +58/−13.

Closes #138947

---

# Proof: openclaw PR #144056 — real Gateway + plugin transport, replacement collision → recovery

Worktree: `D:/code/wt-oc-138947` (branch `fix/oc-138947-descr`)

After side: `HEAD = 1d0f100cbb8` ("fix(gateway): re-drive channel replacements that terminally disconnect before ready")

Before side: `src/gateway/server-channels.ts` reverted to parent `e9ab9c9997b` (single-file checkout, everything else unchanged)

PR source file blob shas: before `272c8e89387`, after `42bce9fa235`

Nothing was pushed and the PR was not modified; the worktree is restored to `HEAD` and `git status --porcelain` is empty at the end.

## 1. What the PR changes

`src/gateway/server-channels.ts` only (plus its unit test). A channel account task that reports a **terminal** verdict (`terminalDisconnect`) *before its first `ready`*, admitted as a **replacement for an account that was recently healthy**, no longer wedges:

```diff

+ const CHANNEL_TERMINAL_RETRY_HEALTHY_WINDOW_MS = 5 * 60_000;   // new constant

+ let taskReachedReady = false;                                  // per-task

+ let previousRuntimeWasHealthy = false;                         // per-attempt

...

+     previousRuntime = getRuntime(channelId, id);

+     previousRuntimeWasHealthy = previousRuntime.running === true ||

+       (lastConnectedAt within 5 min)

...

  if (getRuntime(channelId, id).terminalDisconnect) {

-   recoveryStopTimedOut.delete(rKey); recoveryStartRequested.delete(rKey);

-   restarts.delete(rKey); ... log.info(`[${id}] auto-restart skipped, terminal disconnect`);

-   return;

+   if (previousRuntimeWasHealthy && !taskReachedReady) {

+     log.info(`[${id}] re-driving terminal disconnect before first ready; last connection was recent`);

+     // falls through to the bounded crash supervisor (real backoff, MAX_RESTARTS=10)

+   } else { /* previous wedging behaviour kept */ }

  }

```

The ClawSweeper objection was that the added test only exercises this through a **synthetic in-process registry and mocked retry timing** (`vi.mock("../../packages/retry/src/index.js")` collapses the ladder to 10 ms, `vi.mock("../infra/backoff.js")` replaces `sleepWithAbort`). The evidence below is the requested real-plugin transport trace instead.

## 2. Harness (re-runnable, outside the repo)

All artifacts live in `%LOCALAPPDATA%/Temp/ocproof-144056/` (nothing is written into the repo):

| File | Role |

| --- | --- |

| `plugin-collision-probe/` (`package.json`, `openclaw.plugin.json`, `index.cjs`) | Minimal **external channel plugin**, loaded from disk by the repo's real plugin loader. Channel id `collisionprobe`, account `default`. Its `gateway.startAccount` opens a **real TCP listener** on `$OCPROOF_PORT` = its "transport session"; on abort it keeps the socket bound for `OCPROOF_LINGER_MS` (2500 ms) and then releases it — the still-live predecessor session. A re-admission that tries to bind the same port gets a real `EADDRINUSE` and reports the terminal verdict `blocked` + `terminalDisconnect` with `lastError = "session collision with the previous connection (EADDRINUSE)"`. It appends a JSONL trace (event, ISO timestamp, pid, attempt, address) to `$OCPROOF_TRACE`. |

| `before-run.mjs` | Boots the real Gateway (dev/source entry), waits for the account's baseline session, then writes the config change that owns the channel and records collision → recovery-or-wedge, the last `channels.status` account row, and the host log lines. |

| `watch-reload.mjs` | Same reload step against an already-running Gateway. |

| `runs/dbg/` (after side, host log `gateway-after.log`), `runs/before/` (before side, host log `gateway.log`) | Per-run state dir, config, `plugin-trace.jsonl`, `gateway.log`, `summary.json`. |

Config (both sides identical except the marker/ports): `plugins.load.paths=[plugin dir]`, `plugins.allow=["collision-probe"]`,

`plugins.entries.collision-probe.enabled=true`, `channels.collisionprobe.enabled=true`, `gateway.reload.mode="hybrid"`,

`logging.consoleLevel="debug"` (so channel host logs land on stdout), and `agents.defaults.model.primary="localprobe/probe-model"`

pointing at a config-declared, unreachable local provider so **no model backend / bundled provider plugin** is needed.

Reproduce:

```bash

# boot (after side)

OPENCLAW_STATE_DIR=<run>/state OPENCLAW_CONFIG_PATH=<run>/state/openclaw.json \

OPENCLAW_NO_RESPAWN=1 OPENCLAW_SKIP_CRON=1 OPENCLAW_SKIP_CANVAS_HOST=1 \

OPENCLAW_SKIP_GMAIL_WATCHER=1 OPENCLAW_SKIP_STARTUP_MODEL_PREWARM=1 \

OCPROOF_TRACE=<run>/plugin-trace.jsonl OCPROOF_PORT=<probePort> OCPROOF_LINGER_MS=2500 \

  node --import file:///D:/code/wt-oc-138947/scripts/tsx.mjs D:/code/wt-oc-138947/src/entry.ts \

       gateway --port <gwPort> --bind loopback



# then (from any shell) write the config change that owns the channel:

#   plugins.entries.collision-probe.config.marker  1 -> 2

# host-side account state:

  node --import file:///D:/code/wt-oc-138947/scripts/tsx.mjs D:/code/wt-oc-138947/src/entry.ts \

       gateway call channels.status --json --url ws://127.0.0.1:<gwPort> --token ocproof-144056-token



# one-shot driver (boot + reload + verdict), used for the "before" side:

  node "C:/Users/Administrator/AppData/Local/Temp/ocproof-144056/before-run.mjs" <label> <gwPort> <probePort>

```

## 3. AFTER — HEAD `1d0f100cbb8`

Gateway accepted the plugin (`[plugins] loading collision-probe from ...plugin-collision-probe/index.cjs`), the channel account was admitted by the real manager, and the config write produced a real hot reload with a route handoff.

Plugin transport trace (`runs/dbg/plugin-trace.jsonl`, real socket events):

```json

{"at":"2026-09-13T17:11:30.994Z","pid":31700,"event":"start-attempt","accountId":"default","channel":"collisionprobe","attempt":1,"port":15502}

{"at":"2026-09-13T17:11:31.984Z","pid":31700,"event":"transport-listening","accountId":"default","attempt":1,"address":{"address":"127.0.0.1","family":"IPv4","port":15502}}

{"at":"2026-09-13T17:14:54.144Z","pid":31700,"event":"abort-observed","accountId":"default","attempt":1,"releaseInMs":2500}

{"at":"2026-09-13T17:14:54.570Z","pid":31700,"event":"start-attempt","accountId":"default","channel":"collisionprobe","attempt":2,"port":15502}

{"at":"2026-09-13T17:14:54.571Z","pid":31700,"event":"transport-collision","accountId":"default","attempt":2,"port":15502,"code":"EADDRINUSE","message":"session collision with the previous connection (listen EADDRINUSE: address already in use 127.0.0.1:15502)"}

{"at":"2026-09-13T17:14:56.651Z","pid":31700,"event":"predecessor-port-released","port":15502}

{"at":"2026-09-13T17:14:59.888Z","pid":31700,"event":"start-attempt","accountId":"default","channel":"collisionprobe","attempt":3,"port":15502}

{"at":"2026-09-13T17:14:59.889Z","pid":31700,"event":"transport-listening","accountId":"default","attempt":3,"address":{"address":"127.0.0.1","family":"IPv4","port":15502}}

```

Host log (`runs/dbg/gateway.log`, redacted ports, local time +08:00):

```

01:14:54 [reload] config change detected; evaluating reload (plugins.entries.collision-probe.config.marker)

01:14:54 [gateway/channels] stopping collisionprobe channel before plugin reload

01:14:54 [gateway/channels] restarting collisionprobe channel

01:14:54 [collisionprobe] [default] re-driving terminal disconnect before first ready; last connection was recent

01:14:54 [collisionprobe] [default] auto-restart attempt 1/10 in 5s

01:14:54 [reload] config hot reload applied (plugins.entries.collision-probe.config.marker)

```

`gateway call channels.status` account row while the gateway was live (`runs/dbg/channels-status-after.json`):

```json

{"accountId":"default","restartPending":false,"reconnectAttempts":1,"lastConnectedAt":1789319699890,

 "lifecycle":"ready","running":true,"connected":true,"lastError":null,"enabled":true}

```

`terminalDisconnect` is absent and the lifecycle is `ready`: the replacement that lost the handshake against the still-live predecessor was re-driven by the bounded crash supervisor (real 5 s backoff, `reconnectAttempts=1`) and reached `ready` on attempt 3.

## 4. BEFORE — parent `e9ab9c9997b` (`git checkout e9ab9c9997b -- src/gateway/server-channels.ts`, then restored)

Same harness, same plugin, same config write, separate Gateway process (`runs/before/`, gw port 15503, probe port 15504).

The file was reverted for the run and restored afterwards (`git status --porcelain` empty, blob `42bce9fa235` = HEAD).

Plugin transport trace (`runs/before/plugin-trace.jsonl`):

```json

{"at":"2026-09-13T17:29:23.557Z","pid":23992,"event":"start-attempt","accountId":"default","channel":"collisionprobe","attempt":1,"port":15504}

{"at":"2026-09-13T17:29:24.616Z","pid":23992,"event":"transport-listening","accountId":"default","attempt":1,"address":{"address":"127.0.0.1","family":"IPv4","port":15504}}

{"at":"2026-09-13T17:29:27.124Z","pid":23992,"event":"abort-observed","accountId":"default","attempt":1,"releaseInMs":2500}

{"at":"2026-09-13T17:29:27.375Z","pid":23992,"event":"start-attempt","accountId":"default","channel":"collisionprobe","attempt":2,"port":15504}

{"at":"2026-09-13T17:29:27.376Z","pid":23992,"event":"transport-collision","accountId":"default","attempt":2,"port":15504,"code":"EADDRINUSE","message":"session collision with the previous connection (listen EADDRINUSE: address already in use 127.0.0.1:15504)"}

{"at":"2026-09-13T17:29:29.626Z","pid":23992,"event":"predecessor-port-released","port":15504}

```

Host log (`runs/before/gateway.log`):

```

01:29:26 [reload] config change detected; evaluating reload (plugins.entries.collision-probe.config.marker)

01:29:26 [reload] config change requires channel reload (collisionprobe) — deferring until 1 gateway request(s) complete

01:29:27 [gateway/channels] stopping collisionprobe channel before plugin reload

01:29:27 [gateway/channels] restarting collisionprobe channel

01:29:27 [collisionprobe] [default] auto-restart skipped, terminal disconnect

01:29:27 [reload] config hot reload applied (plugins.entries.collision-probe.config.marker)

```

`gateway call channels.status` account row ~26 s after the config write, while the gateway was live (`runs/before/summary.json`):

```json

{"accountId":"default","restartPending":false,"reconnectAttempts":0,"lastConnectedAt":1789320564616,

 "lifecycle":"blocked","terminalDisconnect":true,"configured":true,"running":false,"connected":false,

 "lastError":null,"enabled":true,"healthState":"terminal-disconnect"}

```

No third attempt and no re-drive ever happened — the predecessor socket was released at `17:29:29.626Z`, yet the account stayed `blocked` with `terminalDisconnect: true` and `reconnectAttempts: 0` (the wedged behaviour the PR fixes; recovery needs a manual `channels.start`).

| | AFTER (`1d0f100cbb8`) | BEFORE (`e9ab9c9997b`) |

| --- | --- | --- |

| attempts after the config write | 2 (collision) → 3 (`listening`) | 2 (collision), none after |

| finalizer line | `re-driving terminal disconnect before first ready; last connection was recent` + `auto-restart attempt 1/10 in 5s` | `auto-restart skipped, terminal disconnect` |

| elapsed collision → verdict | 5.3 s (real backoff) | never |

| account lifecycle | `ready`, `connected:true`, `terminalDisconnect` absent, `reconnectAttempts:1` | `blocked`, `terminalDisconnect:true`, `healthState:terminal-disconnect`, `reconnectAttempts:0` |

## 5. Real vs minimized

Real: the Gateway process (repo dev entry `src/entry.ts gateway`, isolated state dir); plugin loading from an on-disk plugin directory by the repo's own loader; the channel manager, account lifetime and the changed finalizer (`src/gateway/server-channels.ts`) under test; the config hot-reload watcher (chokidar, `reload.mode=hybrid`) driven by a real config file write; the route-handoff stop + re-admission sequence the reload planner runs; the transport collision (a real OS `EADDRINUSE` against a real listener still bound by the predecessor task); retry/backoff timing (`RetrySupervisor` + `sleepWithAbort`, no mocks — the 5 s delay is `RESTART_POLICY.initialMs`); the `channels.status` RPC over the gateway websocket.

Minimized: the channel plugin is a purpose-built fixture rather than a third-party channel; it emulates "the re-admitted handshake loses against the still-live predecessor session" with a TCP port the predecessor still holds (a live Discord/IM session cannot be reproduced offline). No provider or agent traffic is involved; the default model points at an unreachable config-declared local provider, and cron/canvas/gmail sidecars are skipped via the documented `OPENCLAW_SKIP_*` env flags. Sources are executed through the repo's `scripts/tsx.mjs` (dev mode), not a built bundle.

Timing note (environment, not the PR): in a fresh process the first channel admission pays a very slow lazy load of the plugin-runtime module graph through jiti in source mode. Measured: `http server listening` at +20 s, first `start-attempt` at +17 min. Verified with the V8 inspector mid-load: `resolveRegisteredChannelRuntime → resolveRecordChannelRuntime → getRuntimeProperty → resolveRuntime → resolveRuntimeModule → withPluginCache/withProfile → jitiResolve`. It does not change the lifecycle semantics observed below; it is why the boot waits in this document are minutes, not seconds.

## 6. Conclusion

Exercised through a **real Gateway process with a real on-disk plugin** (not a synthetic registry), a **real config-write hot reload** (route-handoff stop + re-admission, the planner path that owns the changed `plugins.entries.<plugin>` path) and a **real transport collision** (OS `EADDRINUSE` against the predecessor's still-live socket), the changed finalizer behaves as the PR claims:

- **AFTER `1d0f100cbb8`**: `re-driving terminal disconnect before first ready; last connection was recent` → `auto-restart attempt 1/10 in 5s` → attempt 3 binds and reports ready ⇒ account `lifecycle: "ready"`, `connected: true`, no `terminalDisconnect`, `reconnectAttempts: 1` (single bounded re-drive, `MAX_RESTARTS` still applies).

- **BEFORE `e9ab9c9997b`**: `auto-restart skipped, terminal disconnect` → the account stays `blocked` / `terminalDisconnect: true` / `reconnectAttempts: 0` even after the predecessor socket is released; recovery requires the manual `channels.start` RPC.

The difference is a single-line-class fix in `server-channels.ts` and it is observable end-to-end, without any mocked retry timing (`RetrySupervisor`/`sleepWithAbort` real; the ~5 s gap is `RESTART_POLICY.initialMs = 5_000`).

Known limitations of this proof: the plugin is a purpose-built fixture (a live third-party IM session cannot be recreated offline) and sources run in dev/source mode, where the first channel admission costs ~13–17 min of lazy plugin-runtime compilation — see the timing note in section 5.

### Re-ran by the author

Artifacts re-inspected (event logs, host log, RPC snapshots). AFTER: config write triggers a real hot reload; the plugin transport's second attempt collides with the still-listening predecessor (`EADDRINUSE`), the host logs `re-driving terminal disconnect before first ready; last connection was recent` + `auto-restart attempt 1/10 in 5s`, and attempt 3 binds (`transport-listening`); RPC account: `lifecycle=ready, connected=true, reconnectAttempts=1`. BEFORE (finalizer source at the parent): `auto-restart skipped, terminal disconnect` — the account stays `lifecycle=blocked, terminalDisconnect=true, reconnectAttempts=0` with no third attempt. Tree restored, clean.

---

### Revision `3b4f82d8eff` - findings addressed + rebased onto current main

- **P1 terminal failures are no longer inferred as collisions.** The host-side health/age predicate is deleted. The transport classifies a retryable session collision explicitly via the new opt-in `ChannelAccountSnapshot.retryableCollision`, captured per task report (never read from the merged runtime row, so a stale/aborted task cannot authorize a retry; cleared at lifecycle start and in `channelReadyPatch`). Unclassified terminal reports (auth rejection, fatal closes, relink-required) keep their **exact previous behavior**: `blocked`, diagnosis preserved, no re-drive. **(Superseded by revision `637c98b2d73` below: this contract was removed per ruling option 1.)**

- **P2 predecessor health captured pre-stop.** `store.preStopHealthy` is recorded at the top of the stop attempt, before teardown clears `running`/`connected`; the recovery decision reads that fact instead of `lastConnectedAt` (which records connection establishment, so accounts connected >5 minutes were previously invisible).

- Recovery rule: `retryableCollision && !reachedReady && predecessorWasHealthyAtStop` -> the existing bounded crash supervisor; everything else wedges as before. **(Superseded — see revision `637c98b2d73`.)**

- **Merge conflicts resolved.** Rebased onto current main `03c70bad04e`; the single import-level conflict in `server-reload-channel-restart.test.ts` resolved onto main's idioms (`reloadChannels`/`restartGatewayChannels`). `git merge-tree` against the live tip is clean.

**Red -> green (owner re-run):** sources reverted + new tests kept -> **3 failed / 11 passed** (P1 reproduced: the old code re-drove a 401 terminal report; P2 reproduced: a 10-minute-connected predecessor wedged). After -> reload suite **14 passed** (incl. 5 recovery cases), `server-channels.test.ts` **167 passed**, health/status suites **117 passed**, SDK/channels **36 passed**. `tsgo` core + gateway-server/gateway-root shards exit 0; `oxlint`/`oxfmt` clean.

**Maintainer notes:**

1. The classification is transport-owned by design. The listed alternative (no host exception at all - transports report collisions as recoverable failures by rejecting `startAccount`) needs no contract change and would shrink this diff to a revert + tests; either is acceptable to us.

2. ~~New optional additive SDK surface `ChannelAccountSnapshot.retryableCollision`~~ **(Superseded by revision `637c98b2d73`: per ruling option 1 the surface was removed; transports use the existing recoverable-start path.)**

3. The previously accepted runtime proof's AFTER log line (`re-driving terminal disconnect before first ready; last connection was recent`) no longer exists on this head; the new line is `re-driving transport-classified session collision before first ready`, and the fixture additionally sets `retryableCollision`. Flagged for the record - we can re-run the runtime proof against this head on request.

### Revision `637c98b2d73` — ruling option 1 executed: the parallel retry contract is removed

ClawSweeper's current findings asked to reuse the existing recoverable-start contract instead of adding a second one (`types.core.ts:195-206`) and to fix the write-once collision boolean (`server-channels.ts:1001-1003`). Both are resolved by deletion, per ruling option 1:

- Removed `ChannelAccountSnapshot.retryableCollision` and all of its producers/consumers: `src/channels/plugins/types.core.ts`, `src/gateway/server-channels.ts` (including the classification block and the write-once boolean), the plugin-SDK snapshot round-trip (`src/channels/plugins/sdk.ts`), and the docs section in `status-and-media.md`.
- A transport signals a recoverable collision exactly the way it already signals any non-terminal start failure: it rejects `startAccount` without emitting `terminalDisconnect`. The existing finalizer handles the rejected start with bounded backoff — no new field, contract, or boolean is introduced.
- Kept intact: the narrowed predicate (a terminal verdict is no longer retried as a handshake collision) and the pre-stop health snapshot.
- The original scenario stays covered by construction: a collision is a rejected start, and the pre-existing recoverable-start re-drive path is what makes the hot-reload replacement recovery tests pass both before and after the contract deletion.

**Evidence (owner re-verified):** reload shard **13/13** (`server-reload-channel-restart.test.ts`, including the four rewritten hot-reload replacement cases: recoverable pre-ready start failure re-driven; a long-running predecessor with a 10-minute timer advance; a genuine terminal verdict stays operator-actionable; a terminal disconnect without reload still wedges); `server-channels.test.ts` 167/167; `channel-status-patches.test.ts` 5/5; `oxfmt --check` clean on all 6 touched files; `tsgo -p tsconfig.core.json` exit 0.


Finn763:
CI fix at `27fb51b2dd0`: `check-additional-boundaries` failed with `gateway-root: 721 test roots exceeds the 720 limit` \u2014 the new test file pushed the core\ntest graph over its cap. The two cases now live in the existing `src/gateway/server-reload-channel-restart.test.ts` (same reload/restart subject) and the\nstandalone file is gone: `scripts/check-tsgo-core-boundary.mts` exits 0, `server-reload-channel-restart.test.ts` \u2192 `7 passed`, `server-channels.test.ts` \u2192\n`146 passed`.

Finn763:
@clawsweeper re-review

PR body updated with the requested transport trace: a real Gateway process with a real on-disk plugin, triggered through a real config write (hot reload). After the fix the replacement collision (a genuine OS `EADDRINUSE` against the still-listening predecessor session) is followed by a re-drive and a successful third bind, ending `lifecycle=ready, connected=true, reconnectAttempts=1`; on the parent source the same collision ends `auto-restart skipped` with the account permanently `blocked / terminalDisconnect=true`.

Finn763:
@clawsweeper re-review - revision 3b4f82d8eff: collision classification moved to the transport owner (`retryableCollision`, per-task), predecessor health captured pre-stop, unclassified terminal reports keep their exact previous behavior; branch rebased onto main. Owner re-ran red->green (3 failed -> suites green).

Finn763:
@clawsweeper re-review

Finn763:
CI red on the previous head (`637c98b2d73`) is **not from this change**, and the branch is now rebased.

**What failed:** `check-lint-core-4` → `oxlint`/`eslint(max-lines)` on

```
ui/src/test-helpers/app-sidebar-cases/session-ownership.ts
  702  error  File has too many lines (702).
```

That file is untouched by this PR — the commit changes 6 files, none of them under `ui/`. The tree at that point carried a **702-line** version (limit 700); `main` has since trimmed it to **576 lines**. So the failure was a main-side condition, and the merge-time tree is where it surfaced.

**Fix:** rebased onto current `main` (`782bb245cd0`). The offending file is back to main's 576 lines, so the `max-lines` rule no longer trips.

**One thing worth flagging honestly, since the rebase itself revealed it:** after rebasing, this branch's source files are now **byte-identical to `main`** — `src/gateway/server-channels.ts`, `src/channels/plugins/types.core.ts` and `channel-status-patches.ts` all match, and the `retryableCollision` contract this PR set out to drop is already absent on main (`0` occurrences). The remaining delta on top of `main` is therefore:

- `docs/plugins/sdk-channel-plugins/status-and-media.md` (+9), and
- `src/gateway/server-reload-channel-restart.test.ts` (+240/−1) — the regression cases that pin the recoverable-start behaviour.

If those regression tests are worth keeping on their own, this PR is now effectively a docs + tests PR and the title/body should be narrowed to say so. If maintainers consider the behaviour already covered by main's implementation and its existing tests, close it — no objection from me either way.

**On evidence:** I could not run the local `vitest` for these tests in this checkout (the available `node_modules` here is missing `mdast-util-from-markdown` and `@vitest/browser-playwright`), so CI on this revision is the test evidence rather than a local run. What I did verify locally is the lint-blocking condition itself — the line count of the file that failed (576 after rebase, was 702).

Finn763:
@clawsweeper re-review — rebased onto current main (782bb245cd0); the lint failure came from an untouched file that main has since trimmed (702 → 576 lines).

PR #144060, opened 2026-09-10
fix(models): align vision and compaction with selected rows

steipete:
Related: #130706, #143822, #142100.

## What Problem This Solves

Vision detection and the host-side Responses compaction threshold can read a differently cased configured model row from the one used for transport. With separate `Model` and `model` rows, capabilities and thresholds can depend on row order instead of the selected model.

## Why This Change Was Made

Use the existing configured-row selector in these two consumers, retaining their existing normalization callbacks for fallback. This aligns exact row choice with request transport and adds four production lines.

## User Impact

Each literal model gets its own vision capability and compaction threshold in either row order. When no literal `MODEL` row exists but both case variants do, the shared selector consistently prefers the canonical `model` destination; this intentionally replaces the previous first-row ambiguity. A single equivalent `Model` row remains a valid fallback.

## Evidence

The two baseline mismatches were reproduced. All 50 focused tests, refreshed P2 review, and the full changed gate passed. An initial test type error was corrected using the existing typed assertion for the actual resolved context window; production was unchanged.

One sourcePerformance build at exact commit `02329d8423ad63fd1cc81d3231f14fda6efd0405` and seven compiled consumer scenarios passed. Vision and the public heartbeat compaction preflight were compared with the real resolver and transport payload policy across literal rows, both orders, single-equivalent fallback, and missing-literal cases. The preflight recorded the intended threshold and returned before compaction; no compaction execution is claimed.

All 9,256 root/workspace build files stayed unchanged. There were no source imports, network calls, or compaction starts, and state/process cleanup completed. This is compiled boundary proof; the runtime profile excludes UI and SDK declarations.

### DIFF
--- a/src/auto-reply/reply/agent-runner-run-params.ts
+++ b/src/auto-reply/reply/agent-runner-run-params.ts
@@ -7,8 +7,8 @@ import {
 import { findModelInCatalog, modelSupportsInput } from "../../agents/model-catalog-lookup.js";
 import { modelTransportRoutesMatch } from "../../agents/model-compat-catalog.js";
 import {
+  findConfiguredProviderModel,
   resolveMergedModelProviderConfig,
-  resolveMergedModelProviderModels,
 } from "../../config/model-provider-config.js";
 import type { resolveProviderScopedAuthProfile } from "./agent-runner-auth-profile.js";
 import type { FollowupRun } from "./queue.js";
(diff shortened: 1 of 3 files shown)


steipete:
The data-model compatibility checklist does not require a migration for this change. The patch changes two read-only selectors; it does not modify config types or validation, serialized formats, database schemas, or any persistence writer. Existing configurations remain valid and are read directly.

The compatibility cases retain single-equivalent and provider-prefix fallback. Literal `Model` and `model` rows now select their own metadata in either order, matching the already-landed transport selector from #143822. The unmatched `MODEL` ambiguity deliberately follows that selector's canonical-destination priority and is documented in the PR body. All 50 focused tests and seven compiled consumer/transport comparisons passed. No binary-upgrade or storage migration is being claimed or needed.

steipete:
Merged via squash.

- Prepared head SHA: [02329d8423ad63fd1cc81d3231f14fda6efd0405](https://github.com/openclaw/openclaw/pull/144060/commits/02329d8423ad63fd1cc81d3231f14fda6efd0405)
- Landed commit: [59c4852ad6f571922c9be59643965d833598bd0d](https://github.com/openclaw/openclaw/commit/59c4852ad6f571922c9be59643965d833598bd0d)

PR #119937, opened 2026-08-06
Improve sessions_search limit diagnostics

BSG2000:
## Summary
- narrow scope to `sessions_search` limit diagnostics only
- move max-bound guidance to the schema boundary for `limit` (positive integer, maximum 25)
- keep execute-time max enforcement and explicit bounded error text
- update regression test expectation for the new runtime message

## Why
Keeps this change independent from the separate empty-reply lifecycle fix track while still improving model/user-facing limit diagnostics.

## Validation
- `node scripts/run-vitest.mjs run src/agents/tools/sessions-search-tool.test.ts` ✅

PR #144061, opened 2026-09-10
fix(agents): stop reporting over-limit tool params as invalid integers

sla55er:
> **Corrected 2026-09-10.** The original version of this description claimed this defect affects `sessions_search` on current builds. That was wrong, and chasing the after-fix proof ClawSweeper asked for is what surfaced it. The corrected analysis is below; the code change is unchanged. See the review-response comment for the delta.

## What Problem This Solves

`readPositiveIntegerParam` and `readNonNegativeIntegerParam` reuse the "must be a positive/non-negative integer" text for their exceeds-max branch:

```ts
if (value !== undefined && options.max !== undefined && value > options.max) {
  throw new ToolInputError(options.message ?? `${key} must be a positive integer`);
}
```

A caller passing a valid integer above the cap is told its value is not an integer, and the message never names the bound. There is nothing to correct toward.

**Reachability matters here, and I got it wrong the first time.** A caller is shielded from this branch if its declared input schema already bounds the value, because schema validation runs before the handler. `sessions_search` declares `limit: optionalPositiveIntegerSchema({ maximum: SESSIONS_SEARCH_MAX_LIMIT })`, so on a real gateway `limit: 50` is rejected earlier with an accurate message:

```
Invalid arguments for tool "sessions_search": limit: must be <= 25.
```

Verified on a live 2026.9.3 deployment with `limit: 50` and `limit: 26`. The unit test reaches the reader message only because it calls `tool.execute` directly, bypassing schema validation.

**The reachable caller today is the ollama node-inference `run` action.** `extensions/ollama/src/node-inference.ts:486,488` read `maxTokens` and `timeoutMs` imperatively with no bounded schema, so an over-limit value there does produce the contradictory message. The same file passes explicit messages for the same two params in its chat handler (lines 327-336) and omits them in the `run` path — the inconsistency this default invites.

So this is a latent-defect fix plus defense in depth for the two shared readers, not a user-visible `sessions_search` fix.

Related: #119937. Note for that PR: the same reachability finding applies to it, since the schema already rejects over-limit `limit` before the handler runs.

## Why This Change Was Made

Fixed in the shared readers rather than at each call site, so every caller that omits an explicit `message` gets a correct default instead of each one having to remember. The wording copies what call sites already pass by hand (`cron-tool.ts:279`), keeping messages consistent. Callers supplying their own `message` are unaffected, and no schema, tool description, or prompt snapshot is touched.

## User Impact

Narrow and mostly preventive. A caller that overshoots a bounded parameter with no schema bound — the ollama `run` action today — receives a message naming the limit rather than one contradicting its input. Nothing changes for callers whose schema already bounds the value, `sessions_search` included.

## Evidence

- **Baseline `c55297caa246aedabd1fe349cc59ee3558f5e07e`**, tests applied without the source change: `pnpm test src/agents/tools/common.params.test.ts src/agents/tools/sessions-search-tool.test.ts` → **2 failed | 36 passed**, `Expected: "limit must be a positive integer no greater than 25"` / `Received: "limit must be a positive integer"`.
- **Candidate `49666092ecb`**, same command → **38 passed**.
- `common.params.test.ts` gains coverage for the max branch *without* an explicit message; neither reader had any.
- Regression sweep: `pnpm test src/agents/tools/` → **120 files, 2693 passed, 1 skipped**. `pnpm test:extension ollama` → **30 files, 821 passed**. `pnpm check` → exit 0.
- **Live-deployment check (2026.9.3), which is what corrected the framing:** a real agent turn calling `sessions_search` with `limit: 50` and `limit: 26` returns the schema message above, not the reader message — so the branch is not reachable through that tool.

**On after-fix real-behavior proof:** I cannot supply it. The only caller that reaches this branch today is the ollama `run` action, and I have no Ollama node to drive it. What I have is the unit coverage above plus the live check that establishes where the branch is and is not reachable. If maintainers would rather not carry a latent fix without runtime proof, say so and I will close this.

AI-assisted.

### DIFF
--- a/src/agents/tools/common.params.test.ts
+++ b/src/agents/tools/common.params.test.ts
@@ -132,6 +132,18 @@ describe("readNumberParam", () => {
     ).toThrow("deleteDays must be an integer from 0 to 7");
   });
 
+  it("names the bound when a max-constrained param exceeds it", () => {
+    // Without an explicit message the over-max branch used to reuse the
+    // "not an integer" text, so a caller sending a valid integer above the cap
+    // was told its value was not an integer and never learned the bound.
+    expect(() => readPositiveIntegerParam({ limit: 26 }, "limit", { max: 25 })).toThrow(
+      "limit must be a positive integer no greater than 25",
+    );
+    expect(() => readNonNegativeIntegerParam({ offset: 8 }, "offset", { max: 7 })).toThrow(
+      "offset must be a non-negative integer no greater than 7",
+    );
+  });
+
   it("treats empty or whitespace-only strings as unset for optional positive integer params", () => {
     // Tool-calling models routinely emit empty-string defaults for optional
     // params (e.g. Telegram replyTo/threadId) they are not actually setting.
--- a/src/agents/tools/common.ts
+++ b/src/agents/tools/common.ts
@@ -243,7 +243,9 @@ export function readPositiveIntegerParam(
     }
   }
   if (value !== undefined && options.max !== undefined && value > options.max) {
-    throw new ToolInputError(options.message ?? `${key} must be a positive integer`);
+    throw new ToolInputError(
+      options.message ?? `${key} must be a positive integer no greater than ${options.max}`,
+    );
   }
   return value;
 }
@@ -267,7 +269,9 @@ export function readNonNegativeIntegerParam(
     }
   }
   if (value !== undefined && options.max !== undefined && value > options.max) {
-    throw new ToolInputError(options.message ?? `${key} must be a non-negative integer`);
+    throw new ToolInputError(
+      options.message ?? `${key} must be a non-negative integer no greater than ${options.max}`,
+    );
   }
   return value;
 }
--- a/src/agents/tools/sessions-search-tool.test.ts
+++ b/src/agents/tools/sessions-search-tool.test.ts
@@ -172,7 +172,7 @@ describe("sessions_search tool", () => {
       "query must not be empty",
     );
     await expect(tool.execute("call-2", { query: "ok", limit: 26 })).rejects.toThrow(
-      "limit must be a positive integer",
+      "limit must be a positive integer no greater than 25",
     );
     await expect(tool.execute("call-3", { query: "x".repeat(4097) })).rejects.toThrow(
       "query must not exceed 4096 characters",


sla55er:
Chasing the after-fix proof you asked for turned up something better: my original framing was wrong, and I have corrected the PR body.

**What I got wrong.** I claimed this defect affects `sessions_search`. It does not, on any current build. `sessions_search` declares `limit: optionalPositiveIntegerSchema({ maximum: 25 })`, and schema validation runs before the handler, so the reader's over-max branch is never reached through that tool. A live 2026.9.3 deployment returns:

```
Invalid arguments for tool "sessions_search": limit: must be <= 25.
```

for both `limit: 50` and `limit: 26`. That message is accurate and names the bound. The unit test reaches the reader message only because it calls `tool.execute` directly and bypasses schema validation — which is exactly what misled me.

**What was wrong with my production numbers.** I cited 87 rejected `sessions_search` calls. Splitting them by error text and date shows two distinct populations:

| Path | Count | Window |
|---|---|---|
| Reader — "limit must be a positive integer" | 64 | 2026-08-12 → 2026-08-31 |
| Schema — "limit: must be <= 25" | 23 | 2026-08-31 → 2026-09-10 |

The switchover is the day that deployment migrated to 2.0. So the misleading message was real on the older build and has been shadowed by schema validation since. Presenting the combined 87 as evidence for current behavior was my error.

**What survives.** The defect in the two shared readers is still real on main, and it is still reachable — just not where I said. `extensions/ollama/src/node-inference.ts:486,488` read `maxTokens` and `timeoutMs` with no bounded schema, so an over-limit value there does get told its valid integer is not an integer. That same file passes explicit messages for the same two params in its chat handler at lines 327-336, which is the inconsistency the current default invites.

So this is a latent-defect fix plus defense in depth, not the user-visible fix I described.

**On the proof gate.** I cannot give you after-fix runtime proof. The one reachable caller needs a connected Ollama node and I do not have one. What I have is the unit coverage and the live check that establishes reachability. That may not clear the bar, and I would rather say so than dress up test output as something it is not — if you would rather not carry an unproven latent fix, I will close this.

sla55er:
Closing this. The defect it fixes is unreachable, and the problem I opened it for does not exist. Both conclusions are mine to own — the review gate asking for real behavior proof is what forced me to actually check, and the check went against me twice.

**The branch is unreachable.** Every in-repo caller that passes `max` to these readers already bounds the same value in its declared input schema, and schema validation runs before the handler:

| Caller | Schema bound |
|---|---|
| `sessions-search-tool.ts:368` | `limit: optionalPositiveIntegerSchema({ maximum: 25 })` |
| `node-inference.ts:486` | `maxTokens: Type.Integer({ minimum: 1, maximum: MAX_TOKENS })` |
| `node-inference.ts:488` | `timeoutMs: Type.Integer({ minimum: 1, maximum: MAX_INFERENCE_TIMEOUT_MS })` |

I claimed `sessions_search` was affected, corrected that to "ollama is the reachable case", then found ollama is bounded too. Confirmed by adding an over-limit assertion to `node-inference.paired-node.e2e.test.ts` and running it against a real Gateway and paired node socket. Unit tests reach the reader message only because `tool.execute` is called directly and bypasses validation — that is what misled me both times.

**The behaviour I reported as a failure is the tool working.** I cited 87 rejected `sessions_search` calls as evidence a model could not recover. Tracing an actual episode:

```
12:38:16  sessions_search {"query":"СУПТО","limit":50}          → must be <= 25
12:38:16  ... four more parallel calls, all limit 50            → must be <= 25
12:38:22  sessions_search {"query":"СУПТО","limit":25}          → ok
12:38:23  ... the same four queries retried at limit 25         → ok
```

Six seconds. The rejections come in one parallel batch issued before any response returns, which is why they all carry `limit: 50` — the model had not yet seen the first error. It then reads the message, corrects the argument, and every query succeeds. My 87 were roughly 40 such pairs: I counted the first half of successful retries and called it an outage.

So the current message does its job, and there is nothing here worth changing.

**What might still be worth a maintainer's minute**, offered as an observation rather than a PR: since all three `max`-passing call sites are shielded by their schemas, the `max` option on `readPositiveIntegerParam` / `readNonNegativeIntegerParam` is currently dead in-repo. Either it is redundant and could go, or it is a real safety net for `plugin-sdk/param-readers` consumers who have no schema — in which case its fallback message is wrong for the over-limit case, which is what this PR changed. That is a judgement call I am not in a position to make.

Sorry for the noise. #144070 is unaffected by any of this.

sla55er:
Reposting my closing note without the raw trace I originally pasted — it contained operational data from a private deployment that I should not have published. The technical conclusion is unchanged.

Closing this because the branch it fixes is unreachable and the problem I opened it for does not exist.

**Unreachable.** Every in-repo caller that passes `max` to these readers already bounds the same value in its declared input schema, and schema validation runs before the handler:

| Caller | Schema bound |
|---|---|
| `sessions-search-tool.ts:368` | `limit: optionalPositiveIntegerSchema({ maximum: 25 })` |
| `node-inference.ts:486` | `maxTokens: Type.Integer({ minimum: 1, maximum: MAX_TOKENS })` |
| `node-inference.ts:488` | `timeoutMs: Type.Integer({ minimum: 1, maximum: MAX_INFERENCE_TIMEOUT_MS })` |

I claimed `sessions_search` was affected, corrected that to "ollama is the reachable case", then found ollama is bounded too. Verified by adding an over-limit assertion to `node-inference.paired-node.e2e.test.ts` and running it against a real Gateway and paired node socket. Unit tests reach the reader message only because `tool.execute` is called directly and bypasses validation — that is what misled me both times.

**The behaviour I reported as a failure is the tool working.** I cited a large count of rejected `sessions_search` calls as evidence that a model could not recover. Tracing an actual episode shows the rejections arrive as a single parallel batch issued before any response returns, which is why they all carry the same over-limit value — the model had not yet seen the first error. It then reads the message, drops to a valid limit, reissues the same queries, and they all succeed, about six seconds end to end. My count was roughly forty such pairs: I counted the first half of successful retries and called it an outage.

So the current message does its job and there is nothing here worth changing.

**One observation for a maintainer**, not a proposal: since all three `max`-passing call sites are shielded by their schemas, the `max` option on `readPositiveIntegerParam` / `readNonNegativeIntegerParam` is dead in-repo today. Either it is redundant, or it is a genuine safety net for `plugin-sdk/param-readers` consumers that have no schema — in which case its fallback message is wrong for the over-limit case. That is a judgement call I am not in a position to make.

Sorry for the noise.

Issue #143980, opened 2026-09-10
[Bug]: taskSuggestions.accept fails with "sessions.create cwd is unavailable: lstat '/workspace'" for Docker-sandboxed agents


shang-go:
**Bug type:** Behavior bug (incorrect output/state without crash)
**Beta release blocker:** No

## Summary
On 2026.9.3 with a Docker-sandboxed agent (`sandbox.mode=all`), accepting a task suggestion fails with `INVALID_REQUEST "sessions.create cwd is unavailable: ENOENT: no such file or directory, lstat '/workspace'"` because the container-internal cwd `/workspace` is persisted by `taskSuggestions.create` and later `lstat`'ed on the gateway host, where `/workspace` does not exist.

## Steps to reproduce
1. Configure `agents.defaults.sandbox = { mode: "all", scope: "agent", backend: "docker" }`. The sandbox container runs with `WorkingDir=/workspace`, bind-mounted from the host agent workspace (e.g. `/home/<sandbox-user>/.openclaw/workspace-agent-rag -> /workspace`).
2. Start a session with an agent that inherits the sandbox (no per-agent override) and ask the agent to create task suggestions. The `suggest_task` tool either receives explicit `cwd "/workspace"` from the model or falls back to its default cwd, which is the container workdir `/workspace`.
3. Observe the cards are created successfully: `taskSuggestions.create` only validates `path.isAbsolute(cwd)`, which `/workspace` passes.
4. In Control UI, accept one of those cards.
5. Observe `taskSuggestions.accept -> sessions.create` fail with `"sessions.create cwd is unavailable: ENOENT: no such file or directory, lstat '/workspace'"`.

## Expected behavior
Task suggestions created from inside a sandboxed session should be accepted without error: the gateway should either map the container path `/workspace` back to the sandboxed agent's host workspace (as it does for other sandbox/host path translations), or reject non-host paths at `taskSuggestions.create` time instead of persisting a cwd that can never resolve on the host. Cards created with the host workspace path accept successfully, demonstrating the intended flow works.

## Actual behavior
[...]

PR #144068, opened 2026-09-10
fix(gateway): reject a sandbox container cwd before probing the Gateway host

xydt-juyaohui:
## What Problem This Solves

With a Docker-sandboxed agent (`sandbox.mode=all`), accepting a task suggestion from the Control UI fails:

```
INVALID_REQUEST "sessions.create cwd is unavailable: ENOENT: no such file or directory, lstat '/workspace'"
```

The agent suggests a task without an explicit `cwd`, so `suggest_task` persists its **container** workdir `/workspace`. `taskSuggestions.create` only checks `path.isAbsolute`, which `/workspace` passes, so the card is stored. On accept, the value is handed to `sessions.create`, where `prepareSessionCreateFilesystemRoot` calls `fs.realpathSync` on it **on the Gateway host**. `/workspace` cannot exist there, the call throws, and the raw host `ENOENT` is reshaped into the message above — which reads like a broken installation rather than a path that belongs to a different filesystem. A card created with a host path accepts fine, so only the container-path case is broken.

## Why This Change Was Made

The container-to-host path translation is not sound here, so the fix classifies instead of translating. A suggestion record keeps only a bare `cwd` plus `sessionKey` and an optional `agentId`; it does not retain the mount table, the container root, or the sandbox generation that was live when the card was created. Suggestions also outlive restarts, while the sandbox may be recreated against a different mount — so mapping `/workspace` back to "the agent workspace" could silently point an old card at a different directory than the user saw.

The change moves sandbox runtime resolution **before** the host probe and rejects a container-only path with a distinct, actionable message. The ordering is the substance of the fix: today `realpathSync` runs first, so the raw `ENOENT` always wins.

`isContainerOnlyPath` judges shape only, deliberately. Host-owned paths that are genuinely missing must keep failing through the existing probe (`session-create-root.test.ts` requires missing cwd, dangling inward/outward links, and a missing workspace to keep reporting `cwd is unavailable:`), so "reject anything that does not exist" would have broken that contract. `C:\...` is not `path.isAbsolute` under POSIX rules, hence the Windows-drive clause; `//server/share` stays a host path.

## User Impact

A sandboxed agent's task card is now accepted or rejected with a message that names the actual cause, instead of a host-filesystem `ENOENT` that looks like a broken install. Non-sandboxed agents, the worktree exemption, and remote-node (`requestedExecNode`) paths are untouched.

### Real behavior proof: standalone script against the real code path

`proof-task-cwd-sandbox.mts` calls the real exported `prepareSessionCreateFilesystemRoot` with real filesystem access (no mocks), first with a host-resolvable cwd and then with the literal container workdir `/workspace`.

**Before the source change** (source file reverted, test harness unchanged):

```
sandboxed agent resolved: true
host cwd -> ok
container cwd -> ERR sessions.create cwd is unavailable: ENOENT: no such file or directory, lstat 'E:\workspace'
RESULT: host path preserved = true; container path rejected with a domain error (no host lstat) = false -> FAIL (defect present)
```

**After the source change** (same harness, same input):

```
sandboxed agent resolved: true
host cwd -> ok
container cwd -> ERR sessions.create cwd is a container path that does not exist on the Gateway host: /workspace
RESULT: host path preserved = true; container path rejected with a domain error (no host lstat) = true -> PASS (fixed)
```

Note the pre-fix line: the host resolved `/workspace` to the nonsense path `E:\workspace` before failing. The fix prevents that host lookup entirely.

### Regression test (defect-sensitive, verified both directions)

`src/gateway/server-methods/session-create-root.test.ts`, run via `node scripts/run-vitest.mjs`:

- **With the fix:** `Test Files 1 passed (1)` / `Tests 22 passed (22)`
- **With only the source file reverted:** `Tests 1 failed | 21 passed (22)`, and the single failure is exactly the new test:

```
FAIL  src/gateway/server-methods/session-create-root.test.ts > session create filesystem root >
      rejects a container workdir instead of probing it on the Gateway host
-  "message": "sessions.create cwd is a container path that does not exist on the Gateway host: /workspace",
```

All 21 pre-existing tests pass in both directions, including the sandbox-containment, missing-path fail-closed, and worktree-exemption cases.

### Static checks

- `node_modules/.bin/oxlint -c .oxlintrc.json <both files>` — `Found 0 warnings and 0 errors.`
- `node_modules/.bin/oxfmt --check <both files>` — `All matched files use the correct format.`

### Environment

Windows, Node v22.22.3, branch based on `upstream/main` @ `16d11819a9c`. The reproduction needs no Docker because the defect is a pure path decision made before any container is contacted.

### What was not tested

No live Docker sandbox end-to-end run: the Gateway host here has no Docker sandbox to accept a real card through the Control UI. The proof exercises the exact function that produces the reported error with the exact container path the issue names. The `suggest_task` tool call itself was read, not executed, as it requires a running sandboxed session.

## Tests and validation

- `node scripts/run-vitest.mjs src/gateway/server-methods/session-create-root.test.ts --reporter=dot` — 22 passed
- `node_modules/.bin/oxlint -c .oxlintrc.json src/gateway/server-methods/session-create-root.ts src/gateway/server-methods/session-create-root.test.ts` — 0 warnings, 0 errors
- `node_modules/.bin/oxfmt --check` on both files — clean
- Standalone proof script — FAIL before, PASS after

## Risk checklist

- Did user-visible behavior change? Yes — a sandboxed agent's container-path task card now reports an actionable error instead of a host `ENOENT`.
- Did config, environment, or migration behavior change? No.
- Did security, auth, secrets, network, or tool execution behavior change? No. The change only makes a rejection happen earlier, before a host filesystem probe, so it narrows rather than widens what reaches the filesystem.
- Highest-risk area: the reordering of `resolveSandboxRuntimeStatus` relative to `realpathSync` in `session-create-root.ts`.
- Mitigation: the sandbox runtime is resolved only when a `sessionCwd` was requested and containment is enforced (the same condition as the pre-existing branch), the early `requestedExecNode` return is untouched, and the full existing suite passes unchanged in both directions.

<details>
<summary>🤖 AI-assisted PR</summary>
This PR was created with AI assistance. Human-run real behavior proof is included above.
Prompts used: scan open OpenClaw issues for a Diamond-rated, uncontested, self-provable defect; verify it against live `main`; reproduce it locally before changing code.
</details>

### DIFF
--- a/src/gateway/server-methods/session-create-root.test.ts
+++ b/src/gateway/server-methods/session-create-root.test.ts
@@ -168,4 +168,62 @@ describe("session create filesystem root", () => {
       }),
     ).toEqual({ ok: true, value: { sessionCwd } });
   });
+
+  it("rejects a container workdir instead of probing it on the Gateway host", () => {
+    // A Docker-sandboxed agent reports its container workdir; `suggest_task` persists
+    // that value, so the Host must classify it instead of failing on a raw lstat.
+    expect(
+      prepareSessionCreateFilesystemRoot({
+        cfg,
+        targetAgentId: "main",
+        enforceSandboxContainment: true,
+        sessionCwd: "/workspace",
+      }),
+    ).toMatchObject({
+      ok: false,
+      error: {
+        code: "INVALID_REQUEST",
+        message:
+          "sessions.create cwd is a container path that does not exist on the Gateway host: /workspace",
+      },
+    });
+  });
+
+  it("still probes host-owned POSIX-style paths so missing ones stay unavailable", () => {
+    // Inside the workspace, a missing entry is a Host problem and must keep reporting
+    // as unavailable rather than being misread as a container path.
+    expect(
+      prepareSessionCreateFilesystemRoot({
+        cfg,
+        targetAgentId: "main",
+        enforceSandboxContainment: true,
+        sessionCwd: path.join(workspace, "missing"),
+      }),
+    ).toMatchObject({
+      ok: false,
+      error: { code: "INVALID_REQUEST", message: expect.stringContaining("cwd is unavailable:") },
+    });
+  });
+
+  it.each([
+    ["sandbox is off", "off"],
+    ["containment is not enforced", undefined],
+  ] as const)("keeps the container-path shape unrejected when %s", (_name, mode) => {
+    if (mode) {
+      cfg.agents!.defaults!.sandbox = { mode };
+    }
+    // Without an active sandbox there is no container namespace to classify against,
+    // so the pre-existing host probe behavior must be preserved unchanged.
+    const result = prepareSessionCreateFilesystemRoot({
+      cfg,
+      targetAgentId: "main",
+      enforceSandboxContainment: mode !== undefined,
+      sessionCwd: "/workspace",
+    });
+    expect(result.ok).toBe(false);
+    if (!result.ok) {
+      expect(result.error.message).toContain("cwd is unavailable:");
+      expect(result.error.message).not.toContain("container path");
+    }
+  });
 });
--- a/src/gateway/server-methods/session-create-root.ts
+++ b/src/gateway/server-methods/session-create-root.ts
@@ -8,6 +8,7 @@ import {
 import { resolveAgentWorkspaceDir } from "../../agents/agent-scope.js";
 import { resolveSandboxRuntimeStatus } from "../../agents/sandbox/runtime-status.js";
 import type { OpenClawConfig } from "../../config/types.openclaw.js";
+import { isWindowsDrivePath } from "../../infra/archive-path.js";
 import { formatErrorMessage } from "../../infra/errors.js";
 import { isPathInside } from "../../infra/path-guards.js";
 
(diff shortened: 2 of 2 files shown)


xydt-juyaohui:
Closing this PR: the approach is unsound, and CI proved it.

## What went wrong

The fix classified a "container path" by **shape** — a POSIX-absolute path that is not a Windows
drive path. That is wrong on Linux, where every legitimate host path is also POSIX-absolute.

`checks-node-compact-small-7` failed with 9 tests, all of them existing tests that pass on `main`:
`src/gateway/server.sessions.create.test.ts` and `src/gateway/server.sessions.create.projects.test.ts`.
Every failure carries my new message against a perfectly valid host path:

```
"message": "sessions.create cwd is a container path that does not exist on the Gateway host: /tmp/oc-vt-ymGpmi/openclaw-session-workspace-project-f51wqC/workspace"
```

That path is a real temporary host directory. The change rejected it, and it also changed the
message for genuine escape cases (`cwd is outside the sandboxed agent workspace` became the
container-path message), which is a behavior regression on top of the false rejection.

Two further problems with the shape heuristic:

- The container workdir is configurable (`agentDocker?.workdir ?? globalDocker?.workdir ?? DEFAULT_SANDBOX_WORKDIR`) and differs per backend — the SSH backend mounts at `/agent`. A name-based guess cannot track that.
- On Linux a host workspace can legitimately *be* `/workspace`, so the guess is not even well-defined there.

My local proof passed only because it ran on Windows, where `/workspace` is not a valid host path.
The reproduction was therefore locally true and generally false — the worst kind of green.

## What is actually established

The underlying defect is real and still present on `main`: `task-suggestion-tools.ts:99` persists
`params.cwd`, which under `sandbox.mode=all` is the **container** workdir
(`openclaw-tools.ts:421` passes `resolveWorkspaceRoot(options?.cwd ?? ...)`), and
`session-create-root.ts:37` then `fs.realpathSync`es that value on the Gateway host, producing
`cwd is unavailable: ENOENT ... lstat '/workspace'`.

The correct fix belongs at the producer, not in the Gateway's path classifier: `suggest_task` should
persist a **host-resolvable** cwd. `src/agents/embedded-agent-runner/sandbox-info.ts:119-120` already
exposes both `workspaceDir` (host) and `containerWorkspaceDir` (container), so the mapping data
exists. That change is larger than a one-line classification and needs the sandboxed tool-build path
threaded through, so it should not be rushed into this PR.

Apologies for the noise — the branch and PR can be ignored/closed. I'll re-approach #143980 at the
producer with the sandboxed run path in view.

PR #143958, opened 2026-09-10
fix(ui): show failed file opens in Review

Marvinthebored:
Fixes #143971.

## Problem

Clicking a missing file, directory, or out-of-workspace file link in chat silently showed unrelated session changes in Review. The Gateway had rejected the file read, but clearing Review's pending selection caused the panel to load the session diff.

## Change

Keep a failed file or artifact open as an unavailable Review selection and show “Unable to open” with the Gateway error. Retain the existing request and session guards so stale failures cannot replace newer selections or reopen a closed panel. Use a compact error callout across the panel width.

Gateway access checks, successful text/image/binary previews, Files errors, and editor permissions remain unchanged. The original contributor commits are preserved. Thanks to @Marvinthebored and @Peetiegonzalez for the original repair and tests.

## Validation

- Real baseline chat reply → file link → Gateway file read → unrelated session diff, with visible tracked changes.
- Rebuilt candidate: missing, outside-workspace, and directory links show the actual error without requesting a diff; valid UTF-8 and unsupported-binary previews work.
- Both new regressions fail on baseline for the intended reasons; 186 focused UI tests pass.
- Formatting, translation verification, CSS lint, source ratchets, and the normal commit hook pass.
- Fresh independent browser acceptance covers delayed failures after selecting another file, the real Files artifact, an actual background task, and Diff; closing Review and switching sessions also retain the newer intent.
- The final error layout was inspected at two viewport sizes. Files errors/retry and editor controls were checked. HTML-shaped error text and denied-editor scopes have focused unit coverage.
- The bounded native type-aware lint attempt exceeded its memory limit. Hosted CI owns that remaining gate; no failed check is counted as passed.

Some cold-start browser control attempts stopped before a file read reached the Gateway. Their timing cause remains unassigned; successful acceptance cases retain the actual request, response, and visible result.

All browser fixtures use synthetic data and a deterministic provider through the normal agent path.

## Before and after

Before: the rejected file link displays unrelated session changes.

Missing file opens the unrelated session diff

After: Review retains the failed selection and shows the error.

Review shows Unable to open and the Gateway error

Valid file links still open the code preview.

Valid file retains its code preview

The hosted image links expire on September 17, 2026. Their downloads were verified against the captured image hashes.

PR #144073, opened 2026-09-10
fix(ui): stop stale task reads after Review changes

steipete:
Related: #143958

## What Problem This Solves

Fixes an issue where leaving a task transcript for a pending or failed file preview could keep fetching the old task's history. Opening an attachment in Files could also clear and reload a still-selected Review transcript.

## Why This Change Was Made

Reconcile task history from Review's current selection and layout before either panel renders. The generic content renderer no longer controls that lifetime, so loading/error states and independent Files rendering follow the same ownership rule. Existing task, connection, paging, and pane-retention owners keep their responsibilities.

## User Impact

Leaving a task's Review selection stops its transcript reads. Switching to Files or hiding the side panel preserves a still-selected task and its loaded transcript.

## Evidence

Mounted Review/Files tests use the real task-history loader and page event dispatcher. Before the fix, five cases fail and two controls pass; the same seven pass afterward. The selected suites passed 132 distinct cases across seven files. After correcting a test-cleanup callback's return type, the final seven cases and all 20 canonical changed checks passed. Independent review found no actionable findings.

Two tests coupled to the old renderer's internal state were replaced by tests of rendered content and history requests. Executable production line count is unchanged, with one ownership comment added; tests grow by 24 lines net. This is synthetic UI boundary proof; no real Gateway, provider, native app, or appearance change is claimed.

### Browser comparison

The same Chromium fixture mounts the production Review, Files, task transcript and attachment components with synthetic content and production CSS. Open the attachment in Files, then click Review. Before the fix, Review reloads and issues three history requests; after the fix, its loaded transcript remains visible with one request. Later mock replies are held to make the reset visible. This is a component composition proof, not a full app or live Gateway run.

The first image shows the original behavior; the second shows the repaired behavior.

Before: returning from Files reloads the Review transcript

After: returning from Files preserves the Review transcript

### DIFF
--- a/ui/src/pages/chat/chat-pane-embedded-panels.test.ts
+++ b/ui/src/pages/chat/chat-pane-embedded-panels.test.ts
@@ -3,7 +3,7 @@
 import { expectDefined } from "@openclaw/normalization-core";
 import { html, render, type LitElement } from "lit";
 import "./components/chat-detail-panel.ts";
-import { afterEach, describe, expect, it, onTestFinished, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, onTestFinished, vi } from "vitest";
 import { createDeferred } from "../../../../test/helpers/promise.js";
 import type { SessionWorkspaceGetResult, SessionWorkspaceListResult } from "../../api/types.ts";
 import type { TaskSummary } from "../../lib/tasks/task-summary.ts";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b3055442d11af696fc5ba17ca2c5617d940648af](https://github.com/openclaw/openclaw/pull/144073/commits/b3055442d11af696fc5ba17ca2c5617d940648af)
- Landed commit: [ee7a1d33c7ce6c777d2a5e845bc48ef684163979](https://github.com/openclaw/openclaw/commit/ee7a1d33c7ce6c777d2a5e845bc48ef684163979)

PR #144080, opened 2026-09-10
improve: simplify source-reply send test inputs

steipete:
Related: #139428

## What Problem This Solves

Source-reply send tests repeat the same Telegram request structure across failure handling, asynchronous mirroring, ordering, and different-target checks. The repeated routing fields obscure the message, target, and idempotency key that distinguish each request.

## Why This Change Was Made

Use one same-file input factory with three required string arguments at the five existing allocation points. Each call creates a fresh request, params object, and tool context. Keep all four affected cases, direct-handler calls, deferred gates, response spies, explicit negative target, and independent expected values unchanged.

## User Impact

No product behavior changes. This removes 31 net test lines while preserving the routing and transcript-mirroring coverage; no production helper or test-only API is added.

## Evidence

- Full send owner and source-reply-mirror sibling before and after: 145 passed, zero failed or skipped (133 send, 12 mirror), with identical per-file ordered case identities and statuses.
- Expanding the actual factory at all five call sites reproduces the complete original file AST, including all assertions and call order; each invocation allocates three fresh objects.
- Required changed-file gate passed; deslop and fresh independent review found no actionable findings.

No tests were deleted or weakened, and no runtime improvement is claimed.

### DIFF
--- a/src/gateway/server-methods/send.test.ts
+++ b/src/gateway/server-methods/send.test.ts
@@ -324,6 +324,21 @@ async function runPollWithClient(
   return { respond };
 }
 
+function createTelegramSourceSendRequest(to: string, message: string, idempotencyKey: string) {
+  return {
+    channel: "telegram",
+    action: "send",
+    params: { to, message },
+    sessionKey: "agent:main:telegram:direct:chat-123",
+    agentId: "main",
+    toolContext: {
+      currentChannelProvider: "telegram",
+      currentChannelId: "chat-123",
+    },
+    idempotencyKey,
+  };
+}
+
 async function runMessageActionRequest(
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9f2aaf18b81f1347127badcbf4a069d2bee15d14](https://github.com/openclaw/openclaw/pull/144080/commits/9f2aaf18b81f1347127badcbf4a069d2bee15d14)
- Landed commit: [03628385d33207df1df0a60cf50548e2ec8f2bdf](https://github.com/openclaw/openclaw/commit/03628385d33207df1df0a60cf50548e2ec8f2bdf)

PR #144082, opened 2026-09-10
fix(xai): stop OAuth refresh loops caused by a stale token expiry

### LINKED ISSUE #127372: xAI refresh can persist a new access token with the previous expired expiry
## Summary

xAI accepts a successful token response whose `expires_in` is missing/malformed/unsafe and whose access-token JWT has no safe `exp`. During refresh, the new credential is built by spreading the old credential and conditionally overwriting `expires` only when the new response produced one. A new access token can therefore inherit the prior token's already-expired timestamp, be persisted, and immediately require another refresh on the next materialization.

Frozen deterministic tests at `a434545620127a3105699513d662395a43b43eb1` refresh at synthetic `now=1000` and prove three malformed/overflow/JWT-fallback responses return a new token with old `expires=100` (3/3 passed). Current `main` was rechecked and retains the conditional spread behavior.

## Owner / call chain

- `src/agents/auth-profiles/oauth-manager.ts` sees the stale credential and enters refresh.
- `extensions/xai/xai-oauth.ts` parses a nonempty `access_token`, but makes `expires` optional when neither a valid `expires_in` nor safe JWT `exp` exists.
- `refreshXaiOAuthCredential()` spreads `...credential`, replaces `access`, and adds `expires` only when `tokens.expires` is truthy.
- The generic OAuth manager persists and returns that refreshed credential without post-validating lifetime.
- Next resolution rejects the retained expired timestamp and refreshes again.

The login sibling shares the parser and can publish a new xAI credential with no expiry at all, which core immediately treats as unusable.

#86583 fixed the earlier absence of xAI auto-refresh. This issue is the later refresh-publication lifetime invariant, not missing refresh support.

## Impact

The triggering request may use the new token once, but every later materialization repeats refresh/token rotation and provider work. Rate limits or refresh-token rotation policy can turn the loop into complete auth failure. xAI OAuth and Grok OAuth consumers are affected.

## Suggested repair

Define one token-lifetime policy in `parseXaiOAuthTokenResponse`: an accepted token response must yield a fresh Date-safe expiry (using a provider-defined conservative fallback only if legitimate) or fail explicitly. A refreshed access token must never inherit the previous access token's expiry.

## Regression

Add login and refresh cases for missing, malformed, overflowed `expires_in`, unsafe JWT `exp`, and no fallback. Assert each response either rejects or produces a credential that passes `hasUsableOAuthCredential` at response time. Explicitly assert old expiry is removed/replaced when access changes.

## Risk

Review required: auth/provider contract and availability. Medium xAI/Grok blast radius; no schema change required.

jensonhirst:
Closes #127372

## What Problem This Solves

Fixes an issue where users signed in to xAI (Grok) through OAuth would refresh their token on every single request, rotating and discarding a refresh token each time, when xAI returned a token response without a usable lifetime.

Refresh built the new credential by spreading the old one and overwriting `expires` only when the response produced a new value. A response carrying a valid `access_token` but a missing, malformed, or out-of-range `expires_in` (and no safe `exp` claim on the access token) therefore published a **new** access token stamped with the **previous** token's expiry.

The device-code login path shares the same parser, so it could also complete "successfully" and store a credential with no expiry at all, which the runtime immediately treats as unusable.

## Why This Change Was Made

The repair is in the producer of the bad credential, `parseXaiOAuthTokenResponse`, rather than in a downstream core guard, so login and refresh get one lifetime policy:

- An accepted token response must yield its own Date-safe expiry, from `expires_in` or, as the existing legitimate fallback, a safe `exp` claim on the access token. When neither is available the response is rejected with an explicit message instead of being accepted.
- `expires` is now a required field on the parsed token response, and refresh publishes it unconditionally. A refreshed access token can no longer inherit the previous token's expiry.

Failing loudly matches how sibling provider plugins already treat this contract (`extensions/chutes/oauth.ts` throws on invalid `expires_in`; `extensions/openai/openai-chatgpt-oauth-token.runtime.ts` reports `expires_in` as a missing field). No conservative default lifetime is invented, since xAI publishes no documented fallback. Non-goals: no change to retry policy, endpoint discovery, refresh-token rotation handling, or the `id_token` identity fields.

## User Impact

xAI OAuth users no longer end up with a stored credential that is expired or near-expired the moment it is written, so the refresh-on-every-request loop and the refresh-token churn behind it are gone. In the rare case where xAI returns a token with no usable lifetime, sign-in or refresh now fails with a message that names the cause and says to re-run the login, instead of silently producing a credential that never works.

### Scope correction, and why the loop does reach persistence on current main

The earlier version of this description claimed the stale credential is persisted, and review pushed back: the shared OAuth manager checks the refreshed credential with `hasUsableOAuthCredential(rotated, { refreshMarginMs: 0 })` at `src/agents/auth-profiles/oauth-manager.ts:1203` before settlement, so an already-expired inherited timestamp is caught there.

That check is real, and it does catch the fully-expired case. It does **not** catch the common one. Refresh normally fires *proactively*, while the credential is still valid but inside the 5 minute `DEFAULT_OAUTH_REFRESH_MARGIN_MS`. In that window the inherited timestamp is still in the future, so a zero-margin check passes and the stale credential settles and persists. The next request is still inside the margin, so it refreshes again.

I verified this end to end rather than by reading, using the **real** `createOAuthManager` settlement and persistence path, the **real** SQLite auth profile store in a temp state dir, the **real** `refreshXaiOAuthCredential`, and a **real** HTTP round trip to a local stand-in for the token endpoint (the plugin's trusted-endpoint check still sees `https://auth.x.ai/oauth2/token`). Seeded credential expires in 60s; the endpoint answers `200` with a rotated `access_token` and `refresh_token` and no `expires_in`; three consecutive `resolveOAuthAccess` calls:

**Before (`origin/main`)** — every request refreshes, every refresh persists the original expiry, every refresh burns a rotation:

```json
[{"attempt":1,"tokenEndpointHits":1,"outcome":"resolved apiKey=access-1","persistedAccess":"access-1","persistedRefresh":"refresh-1","persistedExpiresEqualsOriginal":true},
 {"attempt":2,"tokenEndpointHits":2,"outcome":"resolved apiKey=access-2","persistedAccess":"access-2","persistedRefresh":"refresh-2","persistedExpiresEqualsOriginal":true},
 {"attempt":3,"tokenEndpointHits":3,"outcome":"resolved apiKey=access-3","persistedAccess":"access-3","persistedRefresh":"refresh-3","persistedExpiresEqualsOriginal":true}]
```

**After (this PR)** — one refresh, rejected at the plugin boundary with a message that names the cause; the profile is fenced and no further token traffic is generated:

```json
[{"attempt":1,"tokenEndpointHits":1,"outcome":"threw: OAuth token refresh failed for xai: xAI OAuth token response is missing a usable access-token lifetime (no valid expires…","persistedAccess":"openclaw-oauth-refresh-fence:v1:…:failed:access:…","persistedExpiresEqualsOriginal":false},
 {"attempt":2,"tokenEndpointHits":1,"outcome":"resolved apiKey=none","persistedAccess":"openclaw-oauth-refresh-fence:v1:…:failed:access:…","persistedExpiresEqualsOriginal":false},
 {"attempt":3,"tokenEndpointHits":1,"outcome":"resolved apiKey=none","persistedAccess":"openclaw-oauth-refresh-fence:v1:…:failed:access:…","persistedExpiresEqualsOriginal":false}]
```

### On the merge risk: this reduces refresh-token loss, it does not add it

The flagged risk is that rejecting a successful response discards a rotated refresh token and can force re-authentication. The comparison above is the answer: on `origin/main` this same response **already** rotates and discards a refresh token on *every* request, indefinitely. With this change it happens once, then the profile is fenced and the user is told to sign in again. Same recovery, one burned rotation instead of one per request.

For responses that do carry a usable lifetime, nothing changes. Same harness, same real manager and store, healthy `expires_in: 3600`, three resolves — one refresh, fresh expiry persisted, next two resolves reuse it with no further token traffic:

```json
[{"attempt":1,"tokenEndpointHits":1,"outcome":"resolved apiKey=access-1","persistedAccess":"access-1","persistedExpiresEqualsOriginal":false},
 {"attempt":2,"tokenEndpointHits":1,"outcome":"resolved apiKey=access-1","persistedAccess":"access-1","persistedExpiresEqualsOriginal":false},
 {"attempt":3,"tokenEndpointHits":1,"outcome":"resolved apiKey=access-1","persistedAccess":"access-1","persistedExpiresEqualsOriginal":false}]
```

**Limitation, stated plainly:** I do not have an xAI OAuth account, so none of this is a live `auth.x.ai` observation. The manager, settlement, persistence, plugin parser, and HTTP transport are all real; the token endpoint is a local stand-in returning the response shapes at issue. A maintainer with a live xAI account can confirm the healthy path in one sign-in and one forced refresh.

### The availability tradeoff, stated for the maintainer decision

Review is right that this changes a temporarily usable state into an immediate terminal one, and that is the call to make. Concretely: on `main`, an account that receives one expiry-less successful refresh keeps working while the inherited expiry is still in the future. With this change that refresh fails, which fences the profile and requires signing in again.

I chose to fail closed for three reasons:

1. **The retained expiry is factually wrong.** The stored credential describes a lifetime that belongs to a token that no longer exists, and the runtime budgets refreshes against it.
2. **The "still usable" state is bounded and self-terminating.** It lasts only while the inherited expiry is in the future, and it spends a refresh-token rotation on every request to stay there. Once that timestamp passes, the manager's zero-margin check rejects it and the account is fenced anyway. `main` does not avoid the fence; it defers it and pays rotations until then.
3. **The alternatives are worse here.** Retrying is unsafe on this path by existing design: xAI rotates refresh tokens, and `exchangeXaiOAuthToken` deliberately does not resend one that may already be consumed. Inventing a conservative fallback lifetime has no documented xAI contract to justify it.

**What would change the answer:** if xAI is known to intermittently omit lifetime information on otherwise-valid responses, preserving availability may be worth more than the correct invariant, and this should instead surface a warning while leaving the old credential in place. I can't produce that provider-contract evidence, and I'd rather flag the question than guess at it. Both open merge-risk items come down to this one decision, and I'm happy to rework the change if the call goes the other way.

### Regression coverage

The three tests that encoded the retention behavior are replaced by rejection cases, and new cases cover the omitted-expiry response, publication usability, the JWT `exp` fallback, and device-code login. Against the pre-fix implementation the new assertions fail 5/5; with the fix they pass:

```
$ node scripts/run-vitest.mjs extensions/xai/xai-oauth.test.ts   # pre-fix source
 Tests  5 failed | 32 passed (37)
   AssertionError: promise resolved "{ type: 'oauth', provider: 'xai', …(5) }" instead of rejecting

$ node scripts/run-vitest.mjs extensions/xai/xai-oauth.test.ts   # with this change
 Tests  37 passed (37)
```

### Suite runs

On Node 24.21.0:

```
$ pnpm test:extension xai
 Test Files  37 passed (37)
      Tests  619 passed (619)

$ pnpm test:contracts:plugins
 Test Files  47 passed (47)
      Tests  1117 passed (1117)

$ pnpm check:changed
 (passed: format, lint, typecheck, deadcode, boundary, import-cycle guards)
```

### DIFF
--- a/extensions/xai/xai-oauth.test.ts
+++ b/extensions/xai/xai-oauth.test.ts
@@ -5,7 +5,7 @@ import {
   createRuntimeEnv,
   createTestWizardPrompter,
 } from "openclaw/plugin-sdk/plugin-test-runtime";
-import type { OAuthCredential } from "openclaw/plugin-sdk/provider-auth";
+import { hasUsableOAuthCredential, type OAuthCredential } from "openclaw/plugin-sdk/provider-auth";
 import { clearLiveCatalogCacheForTests } from "openclaw/plugin-sdk/provider-catalog-live-runtime";
 import { withProxyFixture } from "openclaw/plugin-sdk/test-env";
 import { markdownToIR } from "openclaw/plugin-sdk/text-chunking";
@@ -557,45 +557,108 @@ describe("xAI OAuth", () => {
     expect(fetchImpl).toHaveBeenCalledTimes(1);
   });
 
-  it("does not coerce partial xAI expires_in values", async () => {
-    const fetchImpl = vi.fn<typeof fetch>(async () =>
-      jsonResponse({
-        access_token: "access-2",
-        expires_in: "120s",
-      }),
-    );
-    const credential = createXaiOAuthCredential();
-
-    const refreshed = await refreshXaiOAuthCredential(credential, { fetchImpl, now: () => 1_000 });
-
-    expect(refreshed.expires).toBe(100);
-  });
-
-  it("preserves the cached xAI expiry when token lifetimes overflow safe milliseconds", async () => {
-    const fetchImpl = vi.fn<typeof fetch>(async () =>
-      jsonResponse({
+  it.each([
+    {
+      label: "does not coerce partial xAI expires_in values",
+      body: { access_token: "access-2", expires_in: "120s" },
+    },
+    {
+      label: "rejects xAI token lifetimes that overflow safe milliseconds",
+      body: {
         access_token: createJwt({ exp: Number.MAX_SAFE_INTEGER }),
         expires_in: Number.MAX_SAFE_INTEGER,
-      }),
-    );
+      },
+    },
+    {
+      label: "rejects unsafe JWT expiry fallbacks from xAI access tokens",
+      body: { access_token: createJwt({ exp: Number.MAX_SAFE_INTEGER }) },
+    },
+    {
+      label: "rejects xAI token responses that omit any expiry",
+      body: { access_token: "access-2" },
+    },
+  ])("$label instead of retaining the previous expiry", async ({ body }) => {
+    const fetchImpl = vi.fn<typeof fetch>(async () => jsonResponse(body));
     const credential = createXaiOAuthCredential();
 
-    const refreshed = await refreshXaiOAuthCredential(credential, { fetchImpl, now: () => 1_000 });
-
-    expect(refreshed.expires).toBe(100);
+    await expect(
+      refreshXaiOAuthCredential(credential, { fetchImpl, now: () => 1_000 }),
+    ).rejects.toThrow("missing a usable access-token lifetime");
   });
 
-  it("ignores unsafe JWT expiry fallbacks from xAI access tokens", async () => {
+  it("publishes a refreshed xAI credential that is usable at response time", async () => {
+    const now = 1_000;
     const fetchImpl = vi.fn<typeof fetch>(async () =>
-      jsonResponse({
-        access_token: createJwt({ exp: Number.MAX_SAFE_INTEGER }),
-      }),
+      jsonResponse({ access_token: "access-2", expires_in: 3_600 }),
     );
     const credential = createXaiOAuthCredential();
 
-    const refreshed = await refreshXaiOAuthCredential(credential, { fetchImpl, now: () => 1_000 });
+    const refreshed = await refreshXaiOAuthCredential(credential, { fetchImpl, now: () => now });
 
-    expect(refreshed.expires).toBe(100);
+    expect(refreshed.access).toBe("access-2");
+    expect(refreshed.expires).not.toBe(credential.expires);
+    expect(hasUsableOAuthCredential(refreshed, { now })).toBe(true);
+  });
+
+  it("derives the refreshed xAI expiry from the access-token exp claim", async () => {
+    const now = 1_000;
+    const fetchImpl = vi.fn<typeof fetch>(async () =>
+      jsonResponse({ access_token: createJwt({ exp: 7_200 }) }),
+    );
+    const credential = createXaiOAuthCredential();
+
+    const refreshed = await refreshXaiOAuthCredential(credential, { fetchImpl, now: () => now });
+
+    expect(refreshed.expires).toBe(7_200_000);
+    expect(hasUsableOAuthCredential(refreshed, { now })).toBe(true);
+  });
+
+  it("rejects a device-code login whose token response has no usable lifetime", async () => {
+    const fetchImpl = vi
+      .fn<typeof fetch>()
+      .mockResolvedValueOnce(
+        jsonResponse({
+          authorization_endpoint: "https://auth.x.ai/oauth2/authorize",
+          device_authorization_endpoint: "https://auth.x.ai/oauth2/device/code",
+          token_endpoint: "https://auth.x.ai/oauth2/token",
+        }),
+      )
+      .mockResolvedValueOnce(
+        jsonResponse({
+          device_code: "device-code-1",
+          user_code: "ABCD-1234",
+          verification_uri: "https://accounts.x.ai/oauth2/device",
+          expires_in: 900,
+          interval: 5,
+        }),
+      )
+      .mockResolvedValueOnce(
+        jsonResponse({
+          access_token: "access-token",
+          refresh_token: "refresh-1",
+        }),
+      );
+    vi.stubGlobal("fetch", fetchImpl);
+    const ctx: ProviderAuthContext = {
+      config: {},
+      isRemote: false,
+      assertCurrent: vi.fn(),
+      openUrl: vi.fn(async () => {}),
+      prompter: createTestWizardPrompter({
+        progress: vi.fn(() => ({ update: vi.fn(), stop: vi.fn() })),
+        deviceCode: vi.fn(async () => {}),
+      }),
+      runtime: { ...createRuntimeEnv(), log: vi.fn() },
+      oauth: {
+        createVpsAwareHandlers: () => {
+          throw new Error("unexpected VPS OAuth handler request");
+        },
+      },
+    } as unknown as ProviderAuthContext;
+
+    await expect(createXaiOAuthAuthMethod().run(ctx)).rejects.toThrow(
+      "missing a usable access-token lifetime",
+    );
   });
 
   it.each(["fresh", "subscription", "api"] as const)(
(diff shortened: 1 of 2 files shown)


jensonhirst:
Thanks — the review pushed on the right thing, and I've updated the PR body with real end-to-end proof rather than only stubbed-fetch runs. Two substantive responses:

**1. The "current runtime already rejects unusable refresh results" finding is half right, and it changes the severity rather than removing it.**

The zero-margin check at `src/agents/auth-profiles/oauth-manager.ts:1203` is real and does catch a fully-expired inherited timestamp. It does not catch the common case. Refresh normally fires *proactively*, while the credential is still valid but inside the 5 minute `DEFAULT_OAUTH_REFRESH_MARGIN_MS`. There the inherited timestamp is still in the future, the zero-margin check passes, and the stale credential settles and persists — then the next request is still inside the margin and refreshes again.

I verified that through the real `createOAuthManager` settlement path, the real SQLite auth profile store, the real `refreshXaiOAuthCredential`, and a real HTTP round trip to a local stand-in for the token endpoint. Seeded credential expiring in 60s, endpoint answering 200 with rotated tokens and no `expires_in`, three consecutive `resolveOAuthAccess` calls on `origin/main`:

```
attempt 1  tokenEndpointHits=1  persistedAccess=access-1  persistedRefresh=refresh-1  persistedExpiresEqualsOriginal=true
attempt 2  tokenEndpointHits=2  persistedAccess=access-2  persistedRefresh=refresh-2  persistedExpiresEqualsOriginal=true
attempt 3  tokenEndpointHits=3  persistedAccess=access-3  persistedRefresh=refresh-3  persistedExpiresEqualsOriginal=true
```

Three requests, three refreshes, three burned refresh-token rotations, each persisting a new access token stamped with the original expiry. So the persistence loop does reach current main; it just needs the proactive-refresh window rather than an already-expired credential. I've narrowed the claim in the PR body accordingly.

**2. On the P1 merge risk, the comparison runs the other way.**

The concern is that rejecting a successful response discards a rotated refresh token and can force re-authentication. On `origin/main` that same response already rotates and discards a refresh token on *every* request, indefinitely. With this change it happens once, the profile is fenced, and the user is told to sign in again. Same recovery path, one burned rotation instead of one per request.

Responses that carry a usable lifetime are unaffected. Same harness, healthy `expires_in: 3600`, three resolves: one refresh, fresh expiry persisted, next two reuse it with no further token traffic.

**Limitation, stated plainly:** I don't have an xAI OAuth account, so none of this is a live `auth.x.ai` observation. The manager, settlement, persistence, parser, and HTTP transport are all real; the token endpoint is a local stand-in returning the response shapes at issue. A maintainer with a live account can confirm the healthy path with one sign-in and one forced refresh, and that's the one gap I can't close myself.

The maintainer decision I'd still like a call on is the intended one: whether failing closed on an expiry-less response is the behavior you want, given the evidence above that the alternative is unbounded rotation churn.

@clawsweeper re-review

jensonhirst:
Thanks for the re-review, and for taking the zero-margin correction on board.

Both remaining merge-risk items reduce to the one maintainer decision, so I've written the tradeoff into the PR body rather than leaving it implicit. Short version:

The state this replaces is bounded and self-terminating. An account that takes one expiry-less refresh keeps working only while the inherited timestamp is still in the future, and it spends a refresh-token rotation on every request to stay there. When that timestamp passes, the manager's zero-margin check rejects it and the account is fenced anyway. `main` doesn't avoid the fence, it defers it and pays rotations until then. That's why I picked the correct invariant over the deferral.

The alternatives both look worse on this path. Retrying is unsafe by existing design, since xAI rotates refresh tokens and `exchangeXaiOAuthToken` deliberately avoids resending one that may already be consumed. A fallback lifetime has no documented xAI contract behind it, and the first review was right to warn against inventing one.

What would flip my answer: if xAI intermittently omits lifetime information on otherwise-valid responses, availability is probably worth more than the invariant, and this should warn and leave the old credential in place instead. That's the provider-contract evidence I can't produce from here, which is exactly why it's your call. Happy to rework it that way if you'd prefer.

No further re-review needed from me — this is waiting on a human decision, not on more proof.

PR #144106, opened 2026-09-10
refactor(replies): use the inline directive parser directly

steipete:
## What Problem This Solves

Audio and reply-tag processing passed through two private modules that only renamed fields from the existing inline-directive parser. The media and reply-payload owners then unpacked those projections again. This cleanup removes that redundant layer without changing directive policy or delivery behavior.

## Why This Change Was Made

Call `parseInlineDirectives` directly at the existing consumers, preserving their different options: media keeps reply tags while extracting audio directives; reply processing keeps audio tags while extracting reply targets. Preserve text, target precedence, optional fields, payload cloning, hidden metadata and callback identity. Remove both obsolete modules and retain their assertions at the canonical parser.

The Control UI boot inventory is regenerated by its canonical ready-route capture, not hand-edited. Against refreshed base `28589e76ead3`, production is **+10/-54: net 44 lines removed**. Tests are +33/-18; generated inventory is +24/-3 and excluded from production accounting. The complete diff is +67/-75. No public SDK, configuration, dependency declaration, protocol, persistence or authorization change is introduced by this branch.

## User Impact

No behavior change is intended. Audio/reply directives, code-literal preservation, whitespace, media/text ordering and implicit/explicit threading remain intact. This does not retire the directives or combine the differently configured parsing passes.

## Evidence

- Fresh validation on the refreshed candidate passed **420 tests across six files** through the normal wrapper. The complete case inventory retains duplicate and multiline names; within-file names/order match the retained earlier run. This is a new execution, not a claim that the older 420/475 results cover current main.
- Original-main and candidate `pnpm ui:boot-manifest:gen` completed with independently captured generator and logging exits of 0. Complete ordered arrays and serialized bytes differ only by removal of `src/media/audio-tags.ts` from the shared group. The 24 additions and 2 removals in original regeneration are common inherited inventory updates, not attributed to this refactor. An initial original-side log capture failed; its separate exits remain unknown. One authorized observability rerun passed with complete logs and observed process/listener cleanup; the failed observation was not relabeled.
- Normal `pnpm ui:build` and `pnpm build` passed. The full build completed 16 phases, verified 152 SDK subpaths, 53 control-plane imports and 870 UI sidecars. A **separate enforced** `pnpm ui:check-performance --json` on the final full-build output passed with no violations: 353,036 startup-gzip bytes against the unchanged 353,107-byte limit. The build's embedded report-only check is not used as the enforcement claim.
- The [normal configured changed gate](https://github.com/openclaw/openclaw/actions/runs/34542670036) passed all 30 checks on Linux Node 24.19.0 / pnpm 12.3.4. Core lint checked 22,124 files with zero warnings/errors. Native targeted sync could not resolve its `main` reference and used its ordinary full-sync fallback, then verified the exact materialized tree. No manual base/provider/resource override or gate retry was used. The one-shot lease completed, and its capsule, registration and observed processes were absent. The backing Actions holder can show cancellation after successful delegated cleanup; the native command result is retained separately.
- The first refreshed P2 review reported an INDEX-only staging finding. After the gate and cleanup completed, only the canonical manifest was staged; working-source bytes did not change. A fresh complete P2 review then completed four passes with no findings. The earlier findings and exit 1 remain retained.
- A separate exact-commit review of signed follow-up `0e49c3884be70adff2272daf81753714e79e488a` completed four passes with no accepted P0–P2 findings. Its scope is that manifest follow-up against raw parent `2f88bc9b0271`; the complete seven-path candidate was covered by the preceding fresh P2 review.

These refreshed executions used source `2f88bc9b0271` plus the canonical generated manifest. Signed follow-up `0e49c3884be7`, raw parent `2f88bc9b0271`, preserves their exact tree `189798d3802d1f3604dffb48dea6ac7bbfd4d887`; no amend or source rewrite was used to save it.

Historical original/candidate projection proof on base `5366f8db96f4` matched 44 ordered observations across the real media/reply entry points, including own undefined fields, hidden metadata and zero callback invocation, with normal child/home cleanup. That evidence remains bound to its older pair, not a new Gateway or live-channel delivery run.

## CI-driven refresh

The first [failed exact-head run 34490433326](https://github.com/openclaw/openclaw/actions/runs/34490433326) exposed the OpenRouter capture fixture's too-small acquisition budget after shared HTTP deadline delegation. Its separately proved test-only repair landed in [#144178](https://github.com/openclaw/openclaw/pull/144178); the subsequent f883 refresh passed 475 tests, retained as that earlier execution.

The later [f883 run 34502529051](https://github.com/openclaw/openclaw/actions/runs/34502529051) failed the model-picker reopen E2E case. A separate investigation reproduced missing close-side pane invalidation under controlled native-toggle/render ordering; the owner repair landed in [#144399](https://github.com/openclaw/openclaw/pull/144399). This refreshed base includes it. The controlled reproduction is not a claim about the exact historical CI event schedule.

Both old CI runs remain failed and were not replayed. New exact-head hosted CI, the current completed ClawSweeper review and normal native preparation remain required before landing. No whole-program test, hidden-page/pointer, continuous-video or live Gateway/channel-delivery claim is made.


steipete:
Merged via squash.

- Prepared head SHA: [0e49c3884be70adff2272daf81753714e79e488a](https://github.com/openclaw/openclaw/pull/144106/commits/0e49c3884be70adff2272daf81753714e79e488a)
- Landed commit: [5b8492058e73c10a33b5f649771f6dd137535f63](https://github.com/openclaw/openclaw/commit/5b8492058e73c10a33b5f649771f6dd137535f63)

PR #144109, opened 2026-09-10
fix(prometheus): avoid unused private diagnostic copies

steipete:
## What Problem This Solves

Resolves unnecessary private-content copying when operators use Prometheus alongside diagnostic model or tool content capture. Each interested Prometheus listener copied and froze the private payload even though its metrics only consume the event and trust metadata.

## Why This Change Was Made

Trusted exporter subscriptions can opt out of private data with an optional registration argument. Prometheus uses that option and receives a frozen empty third argument while retaining the existing trusted audience, filters, and service lifetime. Default private-data delivery and OTel attribution are unchanged. Older hosts ignore the optional argument and retain their existing behavior.

## User Impact

Prometheus avoids private-payload traversal and cloning while preserving model, tool, GC, and skill metrics, including trusted skill usage when diagnostics are disabled. This does not change content capture settings or diagnostic queue behavior.

## Evidence

The same small synthetic workload through the actual host service bridge and Prometheus exporter measured two private-payload native clone calls and two private-field reads before the repair, and zero of each afterward. Required metric values, rejection of untrusted events, disabled-diagnostics skill accounting, and stop cleanup passed in both runs. These counts are not allocation-byte, RSS, or CPU measurements.

The committed exporter regression failed on the original code because it read the private payload. Focused dispatcher, plugin service, and Prometheus tests passed: 87 tests across five files. Independent review found no actionable P0–P2 findings. The same before/after workload also passed on Linux through [Blacksmith Testbox](https://github.com/openclaw/openclaw/actions/runs/34488671538); both probe processes returned and the lease was independently confirmed completed. This is source behavior proof with matching shared dependencies, not two independent package installations. The final production files are byte-identical to that remote proof. All changed-scope checks passed.

Compatibility decision: retain the documented additive subscription argument. The inspected pre-change host bridge accepts the listener and filter and forwards only those arguments; an extra optional argument leaves its existing copy behavior intact. Existing plugin calls omit the option and retain the new host's default private-data delivery. The optimization therefore requires a supporting host. Independently installed host/plugin combinations remain unexercised, and these results make no production latency or CPU attribution.

### DIFF
--- a/docs/plugins/sdk-runtime/gateway-and-nodes.md
+++ b/docs/plugins/sdk-runtime/gateway-and-nodes.md
@@ -216,6 +216,14 @@ throws after the service lease is revoked. Hosts that do not provide this
 optional capability leave runtime identity unavailable. This diagnostic fact
 does not grant authority or identify a service-reload epoch.
 
+Their `ctx.internalDiagnostics.onEvent(listener, filter?, options?)` subscription
+delivers events, trust metadata, and a frozen private-data object. Exporters that
+only need events and metadata can pass `{ includePrivateData: false }` as the
+third registration argument to skip private payload copies and receive a frozen
+empty object instead. This preserves event filters, trusted-only delivery, and
+service lease cleanup. Private data remains enabled by default; older hosts
+ignore the optional argument and retain their existing copying behavior.
+
 Long-lived services registered with `api.registerService(...)` receive a process-local
 `ctx.gatewayEvents` facade when the process runs a Gateway broadcaster; in runtimes without one the
 field is absent, so feature-detect it and keep a fallback (for example a coarse poll). Use
--- a/extensions/diagnostics-prometheus/src/service.test.ts
+++ b/extensions/diagnostics-prometheus/src/service.test.ts
@@ -1,7 +1,12 @@
 import { createServer } from "node:http";
 import { expectDefined } from "@openclaw/normalization-core";
 // Diagnostics Prometheus tests cover service plugin behavior.
-import type { DiagnosticEventPrivateData } from "openclaw/plugin-sdk/diagnostic-runtime";
+import {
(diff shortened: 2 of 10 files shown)


steipete:
Maintainer decision for `76be4cebb62afb2afc5fd05385e8afd4594edd67`: retain the documented compatibility behavior recommended in the completed review.

The [pre-change service bridge](https://github.com/openclaw/openclaw/blob/2a2f70180ac5d3d1d13cc0e9c238d53c97ca0660/src/plugins/services.ts#L88) accepts the listener and filter and forwards only those two arguments. The new optional third argument is consequently ignored there, preserving existing metrics and copy behavior. The [pre-change Prometheus subscription](https://github.com/openclaw/openclaw/blob/2a2f70180ac5d3d1d13cc0e9c238d53c97ca0660/extensions/diagnostics-prometheus/src/service.ts#L1103) omits the option, so an older plugin retains default private-data delivery on the updated host. Supporting hosts enable the optimization explicitly; there is no version gate or fallback branch.

The PR body now records this decision. Independently installed host/plugin combinations remain unexercised; the source-based compatibility reasoning and shared-dependency Linux proof do not claim that package matrix. No production latency or CPU attribution is claimed. The review found no actionable code or security defect, and the implementation remains unchanged.

steipete:
Merged via squash.

- Prepared head SHA: [76be4cebb62afb2afc5fd05385e8afd4594edd67](https://github.com/openclaw/openclaw/pull/144109/commits/76be4cebb62afb2afc5fd05385e8afd4594edd67)
- Landed commit: [4d08adb468c60d0b555c86be5699b3c84a41c9af](https://github.com/openclaw/openclaw/commit/4d08adb468c60d0b555c86be5699b3c84a41c9af)

PR #144110, opened 2026-09-10
test: reuse source transforms across gateway startup fixtures

steipete:
## What Problem This Solves

Gateway startup process tests repeatedly transform the same application source because every scenario gives its private package fixture a different source path. The startup-allowance cases also discard the transform cache between child processes.

## Why This Change Was Made

Reuse one fixture pathname within the test file, recreating its mutable package assets before every sequential case. Give child temporary files a separate file-owned directory so TSX can reuse its source transforms. Each scenario still gets fresh home, configuration, workspace and state directories and a fresh native Node process.

All eleven cases and their assertions remain: future-version recovery refusal, read-only state preservation, environment selection, and real fast/Commander parsing of both startup flags. The existing final Gateway action hook is unchanged. This is test-only: +21/-5 lines, zero production or dependency changes.

## User Impact

No runtime behavior changes. Developers get faster startup contract tests without adding CI shards.

## Evidence

The same two real startup-mode cases passed in a local original/candidate/original comparison: 19.98s / 14.55s / 16.07s including wrapper setup and the first cold child. The second child took 2.88s with reuse, versus 5.62s and 5.09s on the original. These are local measurements, not a whole-CI speedup claim.

The final candidate also maps Windows TEMP/TMP to the owned temporary directory. All eleven existing cases pass in original order, taking 25.00s locally including wrapper cleanup. The scoped changed-file gate passes, and independent review found no actionable issues. The installed TSX implementation keys transformations by source bytes, source path, options and compiler version; package assets are recreated rather than retained between cases. The separate built Gateway rig continues to own real listener-startup proof.

### DIFF
--- a/src/cli/gateway-cli/pre-bootstrap.process.test.ts
+++ b/src/cli/gateway-cli/pre-bootstrap.process.test.ts
@@ -1,7 +1,7 @@
 import { createHash } from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
-import { afterAll, describe, expect, it } from "vitest";
+import { afterAll, beforeEach, describe, expect, it } from "vitest";
 import { useAutoCleanupTempDirTracker } from "../../../test/helpers/temp-dir.js";
 import {
   createSourceRuntime,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5aa35ca6c93c6b7ff3c754e34e76cd1c5d477141](https://github.com/openclaw/openclaw/pull/144110/commits/5aa35ca6c93c6b7ff3c754e34e76cd1c5d477141)
- Landed commit: [ecf8371c34026c0547fe268cb5cd1be151461908](https://github.com/openclaw/openclaw/commit/ecf8371c34026c0547fe268cb5cd1be151461908)

PR #111964, opened 2026-07-20
fix(google): keep Gemini web search available for new API keys

dillona:
Fixes https://github.com/openclaw/openclaw/issues/96974

## What Problem This Solves

Fixes an issue where new Gemini API-key users receive a model-not-found error from `web_search` when no search model is configured. Google restricts Gemini 2.5 Flash to previously active users; it has not been retired for everyone.

## Why This Change Was Made

Use the stable `gemini-3.6-flash` default in the Google plugin. Preserve explicit `plugins.entries.google.config.webSearch.model` selections, including existing 2.5 pins. Maintainer review selected a stable model instead of the original rolling alias so preview changes cannot silently change the default.

This updates the search guides and tests the public provider's request model, result model, grounding payload, and credential fallback. Thanks @dillona for the original fix and @anguslogan01 for the report.

## User Impact

Unconfigured Gemini searches use a supported grounding model. Explicit pins remain unchanged and must be available to the API key. Gemini 3 grounding bills per search query; Gemini 2.5 bills per prompt. See [Google's pricing contract](https://ai.google.dev/gemini-api/docs/google-search#pricing) and [Gemini 3.6 Flash capabilities](https://ai.google.dev/gemini-api/docs/models/gemini-3.6-flash).

## Evidence

- Blacksmith Testbox [34462536723](https://github.com/openclaw/openclaw/actions/runs/34462536723), Linux Node 24.19.0: `node scripts/run-vitest.mjs extensions/google/web-search-provider.test.ts -t 'selects model'` fails the three unset/empty/blank cases with the former 2.5 default while both explicit pins pass. Restoring the candidate then running the complete provider file passes **77/77 tests**.
- Independent source review: no findings. Fresh autoreview: no actionable P0-P2 findings. Targeted formatting and `git diff --check` pass.
[...]

Issue #96974, opened 2026-06-26
[Bug]: web_search (gemini) default model `gemini-2.5-flash` hardcoded


anguslogan01:
### What happened

The built-in `web_search` tool's gemini provider hardcodes a dated default model (gemini-2.5-flash) instead of a rolling alias.

The provider's default model is hardcoded:

```ts
// extensions/google/src/gemini-web-search-provider.shared.ts
const DEFAULT_GEMINI_WEB_SEARCH_MODEL = "gemini-2.5-flash";          // line 8
export function resolveGeminiModel(gemini?: GeminiConfig): string {
  return trimToUndefined(gemini?.model) ?? DEFAULT_GEMINI_WEB_SEARCH_MODEL; // line 35
}
```

### Confirmed on

- Installed: `OpenClaw 2026.6.8`
- Latest release `v2026.6.10` — same hardcoded `gemini-2.5-flash`
- `main` (today) — `extensions/google/src/gemini-web-search-provider.shared.ts:8` unchanged

This is the same class as #87116 (Generative AI 404 after a model id was retired). Hardcoding a concrete dated/short-lived model id as a default means the default rots every time Google sunsets that id.

### Suggested fix

A bare version bump (`gemini-2.5-flash` → `gemini-3.5-flash`) fixes it today but rots again at the next sunset. More durable options, ideally combined:

1. **Default to a rolling alias** that Google keeps pointed at a live model, e.g. `gemini-flash-latest` (verified: resolves today), instead of a dated id.
2. **Inherit the configured provider model** — fall back to `models.providers.google.models[0].id` (the model the user already trusts for this provider) before any hardcoded constant.

```ts
export function resolveGeminiModel(gemini?: GeminiConfig): string {
  return trimToUndefined(gemini?.model)
    ?? trimToUndefined(gemini?.providerModel)   // inherit configured provider model
    ?? "gemini-flash-latest";                   // rolling alias, not a dated id
}
```

Happy to open a PR for whichever direction you prefer.

### Environment

- OpenClaw 2026.6.8 / v2026.6.10 / main
- Provider: Gemini Developer API (`google-generative-ai`), `web_search` provider `gemini`

vincentkoc:
[...]

PR #144113, opened 2026-09-10
improve(google): default web search to Gemini 3.8 Flash

dillona:
Related: #96974. Follow-up to #111964.

## What Problem This Solves

Gemini web search still defaults to 3.6 Flash. Advance it to the current stable [`gemini-3.8-flash`](https://ai.google.dev/gemini-api/docs/models/gemini-3.8-flash), which supports Google Search grounding. 3.6 remains supported.

## Why This Change Was Made

Update the pinned default, its existing tests, and the two Google search guides. The production change is one line; the request format is unchanged.

## User Impact

Searches without a model override use 3.8. Explicit model selections remain pinned; set `plugins.entries.google.config.webSearch.model` to `gemini-3.6-flash` to retain the previous default. Search-query counts can vary by model, so grounding costs may change.

## Evidence

Node 24.16.0 / pnpm 12.3.4:

- Provider unit tests: **77/77 passed**, covering default selection, explicit pins, credential/base-URL fallback, and time filters.
- Authenticated production-provider live tests: **3/3 passed** for default 3.8, explicit 3.6, and provider-key fallback, each with answer text and citations.
- Temporary live time-filter checks: **2/2 passed** for `week` and an explicit date range; harness removed afterward.
- Formatting, production/test extension typechecks, changed-file lint, and repository guards passed. The optional dead-export scan was stopped under host memory pressure; the full default `check:changed` command therefore did not complete successfully.
- Independent API and test/docs reviews, a fresh review round, and repository `autoreview`: clean.

### DIFF
--- a/docs/providers/google.md
+++ b/docs/providers/google.md
@@ -178,7 +178,7 @@ or let it reuse `models.providers.google.apiKey` after `GEMINI_API_KEY`:
           webSearch: {
             apiKey: "AIza...", // optional if GEMINI_API_KEY or models.providers.google.apiKey is set
             baseUrl: "https://generativelanguage.googleapis.com/v1beta", // falls back to models.providers.google.baseUrl
-            model: "gemini-3.6-flash",
+            model: "gemini-3.8-flash",
           },
         },
       },
--- a/docs/tools/gemini-search.md
+++ b/docs/tools/gemini-search.md
@@ -50,7 +50,7 @@ citations.
                 id: "GEMINI_GATEWAY_TOKEN",
               },
             },
-            model: "gemini-3.6-flash", // default
+            model: "gemini-3.8-flash", // default
           },
         },
       },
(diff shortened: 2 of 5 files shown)


Yigtwxx (commented):
Checked this against `origin/main` — one gap that is worth deciding together with the default itself, plus two measurements that should make the maintainer decision cheaper.

### The new default exists nowhere in the bundled catalogs

`git grep gemini-3.8 origin/main` returns exactly two hits, both fixture strings in `src/flows/doctor-core-checks.test.ts`. The Google plugin catalog stops at 3.7: `extensions/google/openclaw.plugin.json` ships entries for `gemini-3.5-flash`, `gemini-3.6-flash` and `gemini-3.7-flash`, each with `cost`, `contextWindow`, `maxTokens` and `compat.codeMode: "preferred"` (3.7 additionally carries `thinkingLevelMap: { minimal: null }`). This PR makes `gemini-3.8-flash` the shipped default for one feature without adding it to that catalog.

Measured, so the scope of that gap is clear: it does **not** break web search. `resolveGeminiModel` in `extensions/google/src/gemini-web-search-provider.shared.ts` is `trimToUndefined(gemini?.model) ?? DEFAULT_GEMINI_WEB_SEARCH_MODEL` — no catalog lookup, and the request URL is built from that string directly, which is consistent with the live runs in the body. What is affected is everything that reads the catalog rather than the constant: 3.8 will not appear in model listings/pickers for the Google provider, it carries no cost or context-window metadata, and it is not `codeMode: "preferred"` while 3.6 and 3.7 both are. Also worth noting that `docs/tools/code-mode/configuration.md:82` documents the preferred set per provider and would need the same follow-up if a catalog entry is added.

Copying the 3.7 entry with the 3.8 id looks like the whole fix, if maintainers want the catalog and the default to agree.

### Data for the "should the default advance now" question

The review comment frames the decision as 3.6 → 3.8, but main already ships `gemini-3.7-flash` as a catalog entry flagged `codeMode: "preferred"`. So the change skips a model the repo already knows about, and the docs update ("Set this field to `gemini-3.6-flash` to keep the previous default", `docs/tools/gemini-search.md`) does not mention 3.7 as an option either. If the jump past 3.7 is deliberate, saying so in the body would answer the question the reviewer raised.

### Checked and clean

The default really is single-sourced: the only non-test occurrence of the old value outside docs is the constant this PR changes. The other `gemini-3.6-flash` hits in main are unrelated — catalog entries in `extensions/github-copilot/openclaw.plugin.json` and `extensions/opencode/openclaw.plugin.json` (each with their own `replacedBy` chains) and explicit ids in `src/plugin-sdk/test-helpers/provider-runtime-contract.ts`. The config schema does not encode a default either (`webSearch.model` is a free `string` in `configSchema`, with only a label/help entry), so no manifest default is left stale by this change. Both docs pages that state the default are updated, and the test table keeps the previous default as an explicit-pin case.

PR #144115, opened 2026-09-10
refactor(tasks): omit runtime detail from status notifications

### LINKED ISSUE #144114: Task status notifications copy runtime-owned detail they do not use
## Problem

Background-task status notifications repeat work proportional to retained runtime detail even though the Gateway only consumes task metadata. A dirty activity flush deep-clones `TaskRecord.detail`, and mutation notifications clone both current and previous detail. Registry restore also clones the full task inventory into an observer payload that the Gateway ignores.

## Reproduction

On main at `eb9f4ff537c01eed0596bf0949f40be0d5b16e9f`, create a task with ordinary nested JSON detail and register the normal task observer. Send assistant activity and allow the existing one-second activity timer to flush. The observer receives a deep copy of detail, although the Gateway summary, requester routing and terminal transition paths read no detail.

A real Gateway regression test observes one `structuredClone` call for the nested detail during that flush. A registry test also confirms that create/update/publication/deletion observer records contain detail. Full task inspection and persistence do need to retain that detail; the unnecessary copy is at the internal notification boundary.

## Expected behavior

Observer notifications should carry detached metadata sufficient for unchanged Gateway behavior. Full task records, SDK returns, queries and SQLite persistence must retain their existing deep-copy behavior. Restore should keep its reload signal, and activity ordering, terminal-session ownership and stale-observer disposal must remain intact.

## Evidence and scope

An isolated source-bound probe with one synthetic task and 11,088 bytes of JSON detail measured approximately 21.95 µs per synchronous flush/copy/callback before removing the unused copy. This is a small boundary measurement, not end-to-end latency, an allocated-byte measurement, or a claim about a deployed incident. The source audit found the Gateway is the sole production observer; the public `AgentHarnessTaskRecord` full-record contract is separate.

Environment: Node 24.20.0 on macOS. The repair passes 77 affected tests, 173 sibling tests, changed checks and independent review. [Ordinary Linux source-module/SQLite proof](https://github.com/openclaw/openclaw/actions/runs/34490444319) also passes; the PR will link this issue.

steipete:
Closes #144114

<details>

Maintainers can update this same-repository branch.

</details>

## What Problem This Solves

Resolves unnecessary processing in background-task status updates when tasks retain nested runtime details. Registry restores also construct a full task payload that the status subscriber does not use.

## Why This Change Was Made

Task observer notifications now carry detached metadata without runtime-owned detail, and restore notifications retain only their existing reload signal. The Gateway consumes those fields for task summaries, session routing and terminal transitions.

Full task records, plugin SDK returns, queries and SQLite persistence retain their existing deep-copy behavior. Notification ordering, activity throttling, terminal-session ownership and observer disposal are unchanged.

## User Impact

Background-task status updates avoid copying retained details. Task cards, restore refreshes and full task inspection keep the same behavior; no configuration or database change is required.

## Evidence

- The original code failed the registry regression because observer records still contained nested detail, and the Gateway regression because one activity flush cloned that detail.
- All 77 tests in the affected registry and Gateway files pass, including metadata mutation isolation, retained full records, deferred atomic publication, restore refreshes, activity-before-terminal ordering and stale observer disposal.
- All 173 sibling tests pass, covering full-record SDK/runtime consumers and the observer-triggered parent-abort boundary.
- Fresh independent P2 review and the full changed gate pass. The initial gate caught the test-file size limit; restore coverage was consolidated into the existing dedup case without changing production or weakening its assertions.
- An isolated source-bound probe with one synthetic task and 11,088 bytes of JSON detail measured median synchronous flush/copy/callback time of 21.95 → 0.79 µs across three fixed processes. This is a small boundary measurement, not an end-to-end Gateway latency or allocated-byte measurement; all 36 timing batches were retained.
- [Ordinary Linux proof](https://github.com/openclaw/openclaw/actions/runs/34490444319) passed on Node 24.19.0 / SQLite 3.53.3 using actual source modules and isolated SQLite. It verifies retained full detail, detached observer metadata, the real activity timer, terminal/deferred-publication order, restore, deletion and fresh-process readback. All processes and streams closed normally, artifact hashes matched, and the Testbox lease is completed. The committed tree exactly matches this proof source. This is source-module proof; no live Gateway or package-build claim is made.

### DIFF
--- a/src/gateway/server-runtime-subscriptions.test.ts
+++ b/src/gateway/server-runtime-subscriptions.test.ts
@@ -37,6 +37,7 @@ import {
   markTaskLostById,
   markTaskTerminalById,
   recordTaskProgressByRunId,
+  reloadTaskRegistryFromStore,
 } from "../tasks/task-registry.js";
 import { getTaskRegistryObservers } from "../tasks/task-registry.store.js";
 import { resetTaskRegistryForTests } from "../tasks/task-runtime.test-helpers.js";
@@ -780,6 +781,7 @@ describe("startGatewayEventSubscriptions", () => {
       runId: "run-throttle-primary",
       task: "Implement live progress",
       status: "running",
+      detail: { notes: [["runtime-owned task detail"]] },
     });
     const secondary = createTaskRecord({
       runtime: "subagent",
@@ -807,9 +809,15 @@ describe("startGatewayEventSubscriptions", () => {
       data: { text: "parallel" },
     });
 
-    await vi.advanceTimersByTimeAsync(999);
-    expect(broadcast).not.toHaveBeenCalled();
-    await vi.advanceTimersByTimeAsync(1);
+    const clone = vi.spyOn(globalThis, "structuredClone");
+    try {
+      await vi.advanceTimersByTimeAsync(999);
+      expect(broadcast).not.toHaveBeenCalled();
+      await vi.advanceTimersByTimeAsync(1);
+      expect(clone).not.toHaveBeenCalled();
+    } finally {
+      clone.mockRestore();
+    }
     const firstFlush = readTaskUpserts(broadcast);
     expect(firstFlush).toHaveLength(2);
     expect(firstFlush.find((event) => event.task.id === primary.taskId)?.task.lastActivity).toBe(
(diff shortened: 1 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2bd02f2e3bdbe2e5e30b9fca075efdf8b4be3456](https://github.com/openclaw/openclaw/pull/144115/commits/2bd02f2e3bdbe2e5e30b9fca075efdf8b4be3456)
- Landed commit: [7c64f3130236194b11574cc327c9b6df77697101](https://github.com/openclaw/openclaw/commit/7c64f3130236194b11574cc327c9b6df77697101)

PR #143929, opened 2026-09-10
feat(ui): build and rebuild cloud worker snapshots

steipete:
## What Problem This Solves

Cloud worker snapshots could be inspected and recovered in Settings, but operators could not start a build, rebuild a project image, or cancel an active build from that view.

## Why This Change Was Made

Add Build snapshot with eligible-profile and Gateway-local repository selection, Rebuild using a recorded local project root, and confirmed cancellation through the existing environment methods. Group build environments with their snapshots, refresh worker inventory before image inventory, and poll every 10 seconds while a worker build or capture remains active. Preserve the observed worker state when image listing fails.

Failed and orphaned builds remain visible with their existing worker error and count toward Needs attention; they do not keep polling or offer cancellation. Capture counts are deduplicated by lease ID. Pending build dialogs retain their errors, and profile-save notices direct operators to Snapshots after restart.

Carry optional local project-root metadata through preparation and Crabbox allocation/listing, and include profile identity in the existing worker summary. Preserve main's computed project label for repository-backed sources; only local snapshots expose Gateway-local root metadata. No new RPCs, configuration options, database schema, or image reuse policy. The rebase retains #143893's pin, previous-generation, rollback, delete, and retention controls.

## User Impact

Operators can prepare committed project snapshots without starting a session. Inline errors explain invalid selections and prepared-pool capacity. Rebuild uses ordinary preparation/reuse policy and does not force a capture when the existing image remains reusable. Images without a recorded local root omit Rebuild; the repository picker remains available for local checkouts.

Cancellation explicitly asks to destroy the selected worker, including an unused ready reserve. It uses non-force destruction: the existing lifecycle owner checks attachment under the environment lock and retains failed/orphaned records. The earlier stale-cancellation review concern was rejected on that source evidence, and the review accepted the rejection.

## Evidence

- Rebased onto #144002, which repairs the unrelated docs-navigation CI expectation. The previously failing navigation test now passes locally.
- Expanded focused tests passed: 50 project-preparation/provider tests, 117 cloud-worker UI tests, four snapshot browser flows, and 23 Crabbox Gateway tests (194 total).
- Changed-file `pnpm exec oxlint` passed. Core typechecking passed after narrowing the existing local/repository project union without casts or suppressions.
- Independent review of the complete feature was scoped-clean at P0–P2. The new one-line project-variant integration repair also received a clean scoped review.
- Asynchronous failure, capture polling, inventory ordering, and image-list failure regressions were each demonstrated failing before their repair and passing afterward.
[...]

PR #144127, opened 2026-09-10
fix(ui): keep snapshot builds visible when image loading fails

steipete:
Related: #143929

## What Problem This Solves

The Snapshots page hides active and failed builds when its first image-inventory request fails. Operators lose the visible Cancel action and recorded build errors even though the worker list loaded successfully.

## Why This Change Was Made

Render known builds through the existing grouped snapshot view independently of image-inventory availability. Keep image-derived totals absent until inventory succeeds, while retaining the existing polling, confirmation, cancellation and connection ownership.

## User Impact

Operators can see build progress, cancel an active build and read a failed build's error while image inventory is unavailable. Refresh restores the complete image summary after recovery.

## Evidence

The candidate is refreshed onto main ea41203f6c6d, preserving the newer confirmation owner and typed fixtures. The same two first-load regression cases fail on that baseline because the build row is missing and pass with the repair. All 38 cases across the Snapshots page, Cloud Workers page and Gateway page controller pass, along with all 20 canonical changed checks. Fresh independent review found no actionable findings.

A fresh production UI bundle passes the four existing snapshot browser flows: preparing a build, cancelling through the actual confirmation dialog, acknowledging recovery, and pinning/unpinning/deleting an image. The unchanged stale-archive browser case also passes. The earlier CI run 34494388814 failed that archive case with five requests instead of three; the local passing replay does not explain that failure, and its assertions and timeouts remain unchanged. The separate roster scheduling fix in #144239 is not included here.

The attached before/after captures use the same Chromium driver with the actual page and styles and synthetic Gateway responses. Before, only the inventory error appears. After, both known builds, the enabled Cancel button and the worker failure remain visible; image totals stay unknown. These are component browser captures, complemented by the bundled app tests above. No live Gateway, real provider or cloud lease was used.

Before: the initial image inventory failure hides known builds and Cancel

After: known builds, Cancel and build errors stay visible with unknown image totals

### DIFF
--- a/ui/src/pages/cloud-workers/cloud-worker-snapshots.test.ts
+++ b/ui/src/pages/cloud-workers/cloud-worker-snapshots.test.ts
@@ -711,6 +711,88 @@ describe("Snapshot builds", () => {
     }
   });
 
+  it.each(["provisioning", "failed"] as const)(
+    "shows an admitted %s build when the first image inventory fails",
+    async (state) => {
+      let environments = [
+        buildFixture(state, state === "failed" ? "Setup recipe failed" : undefined),
+      ];
+      let failImages = true;
+      const result = { ...snapshotListFixture(), images: [], legacyLeases: [] };
+      const fixture = mountPage(buildMethods, {
+        response: (method) => {
+          if (method === "environments.list") {
+            return { environments };
+          }
+          if (method === "crabbox.images.list") {
+            if (failImages) {
+              throw new Error("Image inventory is unavailable");
+            }
+            return result;
+          }
+          if (method === "environments.destroy") {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d36b8a68b2362b2b258ffa5038f41c339a8fedbf](https://github.com/openclaw/openclaw/pull/144127/commits/d36b8a68b2362b2b258ffa5038f41c339a8fedbf)
- Landed commit: [4742db8697207f7137338bff4b2430c08fdd2bff](https://github.com/openclaw/openclaw/commit/4742db8697207f7137338bff4b2430c08fdd2bff)

PR #144133, opened 2026-09-10
docs: fix 20 link defects confirmed live in the verified backlog

vincentkoc:
Fixes 20 `link` findings, all confirmed live against the current tree. 43 files, +149/−44, adding 57 internal links with none broken.

## Why these were missed

An earlier round reported this category closed on the basis of zero rows in the ledger's `open` bucket. These 30 sat in `verified` — a status meaning a second auditor re-confirmed the finding is real, which is still-open work. A census of all 30 found 20 live, 6 already fixed, 3 refutable by class, 1 out-of-repo. This is those 20.

## What was wrong

**Mis-pointed targets where a correct target exists.** `docs/gateway/external-apps.md` sent readers to `docs/reference/rpc.md` for the Gateway RPC reference — a 43-line page titled "RPC adapters" that names zero Gateway methods (r5-0254, three call sites). `docs/help/troubleshooting.md` and `docs/gateway/index.md` both pointed at `/gateway/background-process` for launchd/systemd supervision; that page is titled "Background exec and process tool" and mentions neither (r5-0255, r5-0256). A card in `sdk-channel-plugins.md` promised reply modes but targeted a section documenting `api.registrationMode` (r3-0560). `manifest.md` linked `/gateway/configuration`, which has no Plugins heading, where `configuration-reference.md:92` has `## Plugins` (r5-0109). `dns.md` pointed wide-area config at a page with zero `wideArea` matches (r3-0208).

**Documented surfaces named repeatedly and never linked.** `chrome-extension.md` had **one** outbound link across 522 lines, with `browser.profiles`, `extensionRelay.allowLegacyAuth` and `browser extension` all unlinked (r3-0829). `manifest.md` named six documented surfaces many times each — none of the six routes appeared anywhere in the page or its seven children (r5-0110). Six `docs/concepts/` pages discussed memory with zero links to `/concepts/memory-architecture` (r3-0240). `agent-send.md` had no reactions link and 8 of the 9 channel pages it lists had none either (r3-0858).

**Pages with no Related section at all**: `mantis.md` (zero root-relative links across the page), and the three `sdk-channel-*` pages (r3-0256, r3-0628, r3-0634, r3-0636).

## One suggested fix that would have broken the build

**r5-0002** proposed the anchor `#agentsdefaultscontextpruning`. The repo's canonical form, at `docs/gateway/config-agents.md:77`, is `#agents.defaults.contextpruning` — dots preserved. Applying the row verbatim would have produced `fragment not found`. The link now uses the working form, and that reference section also gained the `ttl`, `tools.{allow,deny}` and `hardClear.{enabled,placeholder}` keys it was missing, sourced from `src/config/types.agent-defaults.ts:70-87`.

**r3-0208**'s target was half right: after the split, `/gateway/bonjour` owns wide-area *behaviour* but `/gateway/config-runtime#wide-area-dns-sd` owns the *config* the link label promises. Both are linked rather than the one the row named.

## Ownership — and one deliberate omission

Last-match-wins simulation over all 73 rules with positive and negative controls in the same pass: 10 positive controls fired on the correct rule (`docs/gateway/secrets.md`, `docs/gateway/authentication.md`, `docs/security/incident-response.md`, `docs/cli/sandbox.md`, `docs/reference/secretref-credential-surface.md`, `SECURITY.md`, `docs/gateway/sandboxing/index.md`, `docs/reference/RELEASING.md`, `.gitignore`, `pnpm-lock.yaml`); every path in the change set returned unowned.

**No owned paths in this PR** — and one thing was left out to keep it that way. r5-0110's suggested fix ends "add a manifest link from `gateway/secrets`", but that file is file-exact secops-owned. The row's actual finding is about `manifest.md`'s outbound links and is fully closed; putting a one-line back-link through the ownership gate seemed the wrong trade. Flagging it in case a maintainer wants it as a separate owned-file change.

## Verification

Baseline measured in this worktree at the merge-base before starting, not assumed:

- `docs-link-audit --anchors`: **before** 13532 links / **0** broken; **after** 13589 / **0** broken.
- **Negative control on the checker itself**: temporarily rewriting one new Card `href` to `#zzz-negative-control` produced `broken_links=1 … fragment not found`, proving the auditor validates fragments inside JSX `href` attributes and that the new anchor genuinely resolves. File restored byte-for-byte.
- `pnpm check:docs` exit 0 · `pnpm format:check` exit 0 (36376 files) · `pnpm channels:catalog:check` exit 0 (run because 8 channel pages were touched; no `summary` frontmatter changed) · `docs-link-audit.test.ts` 1 file / 33 tests.
- `docs/docs.json` untouched, so the nav-pinning test is not implicated.

Glossary: six sources added, each **inserted beside a related existing entry** rather than appended, so sibling branches do not conflict on the tail of the file. 1205 → 1211 entries, 0 duplicate sources.

Worth noting for future link work: the fixes had to be authored in two syntaxes, because `agent-send.md` and four channel pages use JSX Cards while others use markdown lists — and the i18n glossary gate only inspects markdown list-item links, so the Card-form additions needed no glossary entry while the list-form ones did.


vincentkoc:
Addressing the one open item — **Review config compatibility**.

This PR changes no config surface and no default. It documents keys that already exist, in a reference section that was missing them, and every value it states was checked against source rather than copied from the row that asked for it:

| documented | source | matches |
|---|---|---|
| `ttl` default `5m`, bare numbers are minutes | `src/agents/embedded-agent-runner/tool-result-truncation.ts:63` (`ttlMs = 5 * 60_000`) and `:65` (`parseDurationMs(config.ttl, { defaultUnit: "m" })`) | yes |
| `hardClear.placeholder` default `[Old tool result content cleared]` | `:46` `CACHE_TTL_DEFAULT_PLACEHOLDER`, applied at `:75` | yes |
| `ttl`, `tools.allow`, `tools.deny`, `hardClear.enabled`, `hardClear.placeholder` exist under `contextPruning` | `src/config/types.agent-defaults.ts:70-87` | yes |

No key is added, renamed, defaulted differently, or deprecated. The upgrade impact is nil: a reader who follows the page today gets behaviour that already ships.

For context on why the section was incomplete — the row behind this (`r5-0002`) reported that `/gateway/configuration` documents `contextPruning` **zero times**, and the reference section that does own it listed only `mode`. The link now points at the real owner, and the owner now lists all five keys.

One thing worth flagging since it is adjacent to compatibility: the row's *suggested* fix specified the anchor `#agentsdefaultscontextpruning`. The repo's canonical form at `docs/gateway/config-agents.md:77` preserves the dots — `#agents.defaults.contextpruning` — so applying the row verbatim would have shipped a `fragment not found`. The working form is what landed.

Verification, measured in this worktree at the merge-base and again after: `docs-link-audit --anchors` 13532 → 13589 internal links, **0** broken both times. I also ran a negative control on the checker itself — temporarily rewriting one new `href` to a bogus fragment produced `broken_links=1 … fragment not found`, then restored the file byte-for-byte — so "the check passed" here means "the check would have caught it".

PR #144140, opened 2026-09-10
fix(sessions): rebind CLI session state on compaction checkpoint branch and restore

yetval:
## Summary
On a session that has a compaction checkpoint and a live native CLI conversation (for example a `claude-cli` backed session), `sessions.compaction.restore` reports success and rewinds the local transcript, but the model keeps answering from the post-compaction native conversation. `sessions.compaction.branch` has the mirror problem: the new session key inherits the parent's native session id, so two OpenClaw sessions interleave turns into one native conversation.

Trigger: a session with at least one stored compaction checkpoint and a stored CLI session binding, then branch or restore that checkpoint. Restoring the same retained checkpoint a second time has the same visible outcome for a different reason, covered below.

## Root cause
`src/config/sessions/session-accessor.sqlite-checkpoint.ts`

Branch and restore both mint a brand new `sessionId` and fork a fresh transcript at the checkpoint boundary via `forkSqliteCheckpointTranscriptInTransaction`, then project the successor entry through `cloneSqliteCheckpointSessionEntry`. That clone spreads `...currentEntry` and clears the run, usage and token fields, but it never touched any CLI session state. Two session-scoped fields therefore survived a session rotation that invalidates both:

1. `cliSessionBindings` / `cliSessionIds` / `claudeCliSessionId` still name the source's native conversation, so the next turn resumes the conversation the checkpoint was meant to undo.
2. `cliHistoryBoundary` still names the source `sessionId` and the source transcript `generation`, so it is a proof about a transcript the successor does not have.

The second one is why clearing the binding alone is not a complete repair. With the binding gone and a stale boundary present, `prepareCliHistoryBoundary` rejects the boundary on the `sessionId` mismatch (`src/agents/cli-runner/history-boundary.ts`), and it cannot fall through to the proven-empty path because the forked conversation is not empty. `prepare.ts` then selects `auth-unknown` and `loadCliSessionPromptContext` returns no history. The successor starts a fresh native conversation with nothing in it.

Repeated restore, the P1 raised in the last review: a restore keeps the stored checkpoint positions, and those positions keep naming the transcript the checkpoint was captured on. After one restore the live session owns only its own copy of those rows, so the next restore of the same retained checkpoint forked a foreign session. Ownership could not be proven for rows read out of a session the boundary says nothing about, the successor was left unowned, and the operation still reported `created`. The next CLI turn selected `auth-unknown` and opened an empty conversation.

## Fix
Three changes in the one transaction that both operations share.

Clear the native binding with the same `clearAllCliSessions()` helper every other rotation path already uses, and stop inheriting the source's history proof.

Re-stamp that proof onto the forked transcript when, and only when, the source boundary actually covers every row the fork copied. `resolveCheckpointSuccessorCliHistoryBoundary` requires all of:

- the stored boundary is a `known` proof naming the source `sessionId`
- the fork read from the source session itself, not a legacy blob or a cross-session checkpoint position
- the boundary `generation` still equals the source transcript's live generation
- the boundary `maxSeq` is at or beyond the highest seq the fork copied

It then writes a boundary carrying the successor `sessionId` and the forked transcript's own `generation` and `maxSeq`, keeping the source `authFingerprint` and `writerRunId`. The fingerprint is the account check, so a different account on the next turn is still refused. If any condition fails, the successor is left unowned rather than inheriting a stale proof.

For the repeated restore, resolve the fork source against the transcript the live session owns before falling back to the position the checkpoint recorded:

```ts
const addForkSourceWithCurrentCopy = (source: SqliteCheckpointTranscriptForkSource) => {
  if (source.leafId && source.sessionId !== currentSessionId) {
    sources.push({ ...source, sessionId: currentSessionId });
  }
  sources.push(source);
};
```

A retained checkpoint position is only re-pointed at the live session when an exact leaf id bounds it, and `readSqliteTranscriptRowsForFork` still has to find that leaf in that session, so the copied rows are the same rows. A position without a leaf keeps naming the recorded session, because reading the live session unbounded would copy the rows the checkpoint exists to drop. This picks the first of the two directions review offered, preserving verifiable ownership rather than refusing the operation, and it needs no new failure status, no relaxed check and no schema change. The account, generation, coverage and live-run safeguards are untouched: the successor still has to satisfy all of them, and a source that cannot satisfy them is still left unowned.

Branch after a restore takes the same path, so the two operations stay symmetric.

`cloneSqliteCheckpointSessionEntry` is the single place where both branch and restore project the successor entry, so one repair covers both operations and cannot drift apart later. The near duplicate rewind clone at `src/config/sessions/session-accessor.sqlite-message-cut.ts` already carries the comment "A rotated transcript cannot resume provider/runtime identity from the old tail" and clears the binding there; the checkpoint clone was the one rotation that was missed.

Re-ownership is done at the checkpoint, not in `prepareCliHistoryBoundary`, because the checkpoint transaction is the only place that knows the copy happened and can see both transcripts at once. Nothing in the CLI runner's admission rules is relaxed.

Deliberate scope boundary: the message-cut sibling also clears `agentHarnessId`, `contextTokens`, `contextBudgetStatus` and `compactionCount`. This patch does not touch those. Whether a checkpoint successor should also drop harness identity and compaction accounting is a policy question with its own user-visible consequences, and it is not needed here. Happy to extend if a maintainer wants that alignment.

Known residual, stated rather than hidden: a checkpoint whose recorded position cannot be located in the live session, for example an imported or legacy-sourced position with no exact leaf, still produces an unowned successor. That is the existing accepted contract from #140294, which deliberately refuses unverifiable history, and this patch does not widen it.

## Regression coverage
`src/config/sessions/session-accessor.conformance.test.ts`

- the existing checkpoint case now also asserts the successor does not inherit the source history proof
- both the branch and the restore successor carry a re-owned boundary whose `sessionId`, `generation` and `maxSeq` match the forked transcript while the `authFingerprint` is preserved
- a source boundary that stops short of the copied rows leaves the successor unowned
- restoring the same retained checkpoint a second time still produces an owned successor whose boundary matches its own transcript, and the copied conversation rows are present
- a new case drives the real preparation path on the twice-restored successor: the owning account receives the checkpoint conversation, a revoked successor run is rejected by the history writer's `assertReadable` before execution, and a second account is refused and receives no checkpoint history

## Verification
- `node scripts/run-vitest.mjs src/config/sessions/session-accessor.conformance.test.ts` passes 52 cases on this head; the two repeated-restore cases fail on the previously reviewed head `8de35315747`
- `node scripts/run-vitest.mjs src/gateway/server.sessions.compaction.test.ts src/agents/cli-runner/history-boundary.test.ts` passes, covering the gateway checkpoint handlers and the account boundary contract this patch transfers
- `oxfmt` clean on both changed files
- full typecheck and full build not run locally on this box; CI owns them

## Real behavior proof
Behavior addressed: branching or restoring a compaction checkpoint on a session with a live native CLI binding rotates the local transcript but leaves the successor pointing at the old native conversation, so the next turn resumes the conversation the checkpoint was meant to undo; clearing only that binding replaces the wrong context with no context, and restoring the same retained checkpoint again lands in the same empty state while still reporting success.

Real environment tested: the real production runtime driven end to end on pristine `origin/main` `550353573e3` and on this patched head `fad01fa8d2c`, identical inputs, plus one middle run on the previously reviewed head `8de35315747` to isolate the repeated-restore finding. Real `node:sqlite` store on disk under a scratch state directory, real `compactionCheckpointStore` from `src/agents/embedded-agent-runner/compaction-checkpoint.ts` for capture and persistence, real `persistCompactionCheckpoint`, real `branchCheckpointSession` / `restoreCheckpointSession`, and then the real next-turn path the CLI runner uses: `prepareSystemAgentRunAdmission`, `prepareCliHistoryBoundary`, `loadCliSessionPromptContext` and `buildCliSessionHistoryPrompt`, with the binding read through the real `getCliSessionBinding` and the execution-boundary authority check `executePreparedCliRun` installs as its pre-send assertion. No stubs and no doubles; the only thing not driven is the `claude` subprocess itself. Account fingerprints, transcript generations and leaf ids are redacted; the accounts are synthetic.

Exact steps or command run after this patch: seed a SQLite backed session, take a real first CLI turn that establishes a genuine `cliHistoryBoundary` under a `claude-cli` account credential and writes two conversation messages through the granted history writer, take a second turn so the proof covers every row, persist the native binding the backend would return (`native-pre-checkpoint`), capture and persist one manual compaction checkpoint at the live leaf, then branch the checkpoint, restore it, restore the same retained checkpoint again, and drive each successor's first CLI turn, printing the native conversation the successor would resume, whether the history writer was granted, the reseed reason `prepare.ts` would select, the messages handed to the model, the size of the prompt actually sent to the backend and whether it carries the checkpoint conversation, then revoke the successor's live run and re-run the execution authority check, then repeat the successor turn under a second account.

Evidence after fix:
```
# BEFORE (pristine main 550353573e3)
SOURCE   boundary={"version":1,"sessionId":"checkpoint-source","state":"known","authFingerprint":"<account-fingerprint redacted>","generation":"<generation redacted>","maxSeq":2,"writerRunId":"cli-run-2"} binding={"claude-cli":{"sessionId":"native-pre-checkpoint"}}
CHECKPOINT preCompaction={"sessionId":"checkpoint-source","leafId":"<leaf redacted>","entryId":"<leaf redacted>"}
BRANCH   result=created
[agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
BRANCH   nativeResume=native-pre-checkpoint historyWriter=refused reseedReason=auth-unknown restoredMessages=0
BRANCH   restoredContext=[]
BRANCH   promptSentToBackend chars=0 carriesCheckpointHistory=false
RESTORE  result=created
[agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
RESTORE  nativeResume=native-pre-checkpoint historyWriter=refused reseedReason=auth-unknown restoredMessages=0
RESTORE  restoredContext=[]
RESTORE  promptSentToBackend chars=0 carriesCheckpointHistory=false
RESTORE2 result=created
[agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
RESTORE2 nativeResume=native-pre-checkpoint historyWriter=refused reseedReason=auth-unknown restoredMessages=0
RESTORE2 restoredContext=[]
RESTORE2 promptSentToBackend chars=0 carriesCheckpointHistory=false
[agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
OTHERACCT nativeResume=native-pre-checkpoint historyWriter=refused reseedReason=auth-unknown restoredMessages=0
OTHERACCT restoredContext=[]
OTHERACCT promptSentToBackend chars=0 carriesCheckpointHistory=false

# MIDDLE (previously reviewed head 8de35315747, identical inputs)
SOURCE   boundary={"version":1,"sessionId":"checkpoint-source","state":"known","authFingerprint":"<account-fingerprint redacted>","generation":"<generation redacted>","maxSeq":2,"writerRunId":"cli-run-2"} binding={"claude-cli":{"sessionId":"native-pre-checkpoint"}}
CHECKPOINT preCompaction={"sessionId":"checkpoint-source","leafId":"<leaf redacted>","entryId":"<leaf redacted>"}
BRANCH   result=created
BRANCH   nativeResume=none historyWriter=granted reseedReason=missing-transcript restoredMessages=2
BRANCH   restoredContext=["user:deploy the canary build","assistant:canary build deployed"]
BRANCH   promptSentToBackend chars=466 carriesCheckpointHistory=true
RESTORE  result=created
RESTORE  nativeResume=none historyWriter=granted reseedReason=missing-transcript restoredMessages=2
RESTORE  restoredContext=["user:deploy the canary build","assistant:canary build deployed"]
RESTORE  promptSentToBackend chars=466 carriesCheckpointHistory=true
RESTORE2 result=created
[agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
RESTORE2 nativeResume=none historyWriter=refused reseedReason=auth-unknown restoredMessages=0
RESTORE2 restoredContext=[]
RESTORE2 promptSentToBackend chars=0 carriesCheckpointHistory=false
[agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
OTHERACCT nativeResume=none historyWriter=refused reseedReason=auth-unknown restoredMessages=0
OTHERACCT restoredContext=[]
OTHERACCT promptSentToBackend chars=0 carriesCheckpointHistory=false

# AFTER (this patch, head fad01fa8d2c, identical inputs)
SOURCE   boundary={"version":1,"sessionId":"checkpoint-source","state":"known","authFingerprint":"<account-fingerprint redacted>","generation":"<generation redacted>","maxSeq":2,"writerRunId":"cli-run-2"} binding={"claude-cli":{"sessionId":"native-pre-checkpoint"}}
CHECKPOINT preCompaction={"sessionId":"checkpoint-source","leafId":"<leaf redacted>","entryId":"<leaf redacted>"}
BRANCH   result=created
BRANCH   nativeResume=none historyWriter=granted reseedReason=missing-transcript restoredMessages=2
BRANCH   restoredContext=["user:deploy the canary build","assistant:canary build deployed"]
BRANCH   promptSentToBackend chars=466 carriesCheckpointHistory=true
RESTORE  result=created
RESTORE  nativeResume=none historyWriter=granted reseedReason=missing-transcript restoredMessages=2
RESTORE  restoredContext=["user:deploy the canary build","assistant:canary build deployed"]
RESTORE  promptSentToBackend chars=466 carriesCheckpointHistory=true
RESTORE2 result=created
RESTORE2 nativeResume=none historyWriter=granted reseedReason=missing-transcript restoredMessages=2
RESTORE2 restoredContext=["user:deploy the canary build","assistant:canary build deployed"]
RESTORE2 promptSentToBackend chars=466 carriesCheckpointHistory=true
RESTORE2 revokedSuccessorRun=REJECTED before any backend send: CLI history authority changed before execution
[agent/cli-backend] cli session history refused across auth boundary: reason=auth-unknown
OTHERACCT nativeResume=none historyWriter=refused reseedReason=auth-unknown restoredMessages=0
OTHERACCT restoredContext=[]
OTHERACCT promptSentToBackend chars=0 carriesCheckpointHistory=false
```
Observed result after fix: on pristine main both successors resume `native-pre-checkpoint`, the conversation the checkpoint was meant to undo, and the runtime logs its refusal to hand over local history. On the previously reviewed shape the first branch and the first restore recover the two pre-checkpoint messages, but restoring the same retained checkpoint again drops back to `auth-unknown` with an empty prompt while still reporting `created`, which is the P1 raised in review. With this patch all three operations open a fresh native conversation and hand the backend the same 466 character prompt carrying the checkpoint conversation. Revoking the successor's live run between preparation and execution makes the run authority check reject it before any backend send, and driving the same successor under a second account is refused and receives an empty prompt, so the account boundary and the live-run boundary both still hold on the transferred history.
What was not tested: this drives the store, the real preparation path, and the execution-boundary authority check that `executePreparedCliRun` installs before it sends anything; it does not spawn a real `claude` binary to watch the provider open the new conversation, and it does not exercise the gateway RPC transport in front of `sessions.compaction.restore` and `.branch`. Cold-archived source transcripts are not covered; the boundary re-own reads the hot transcript tip, and a source whose generation no longer matches leaves the successor unowned, which is the safe side.

## ClawSweeper verdict items
- **Preserve usable history when restoring a retained checkpoint again (P1)**: fixed in `src/config/sessions/session-accessor.sqlite-checkpoint.ts` by resolving the fork source against the transcript the live session owns when an exact leaf bounds it, so the retained position stays provable instead of pointing at a session the boundary cannot speak for. Review offered two directions and this is the narrower one: no safeguard was removed, no new failure status was added, and the account, generation, coverage and live-run checks all still apply. The evidence block above reproduces the finding on the reviewed head and shows it repaired on this head.
- **Add real behavior proof (authority chain)**: the proof now runs past preparation to the execution boundary. It shows the reseeded prompt the backend receives for branch, restore and repeated restore, the rejection of a revoked successor run before any send, and the empty prompt a second account receives.
- **Resolve merge risk: transferred history must stay unavailable to a different account or a revoked run (P1)**: answered by the last two lines of the after block, driven on the twice-restored successor, and covered by a regression case.
- **Resolve merge risk: the durable history-transfer contract lacks recorded maintainer acceptance (P1, owner decision)**: still open and not author-actionable. This PR extends the account-bound history contract accepted in #140294 to a copied transcript. The rule it adds is narrow: a successor inherits ownership only when the source proof is `known`, names the source session, matches its live generation, and covers every copied row, and the copied rows are read from a transcript the live session owns. Both maintainer options remain real. Option 1 is to accept this bounded transfer, which is what this PR implements. Option 2 is to pause the ownership transfer entirely, keep only the binding clear, and accept that every checkpoint successor starts with no CLI history; if that is the preferred contract, say so and I will cut the transfer back to the clear-only shape rather than argue for the transfer.
- **Resolve maintainer decision**: the question in the verdict, whether restores should preserve verifiable ownership across retained checkpoint sources or refuse CLI restoration when ownership cannot be carried, is answered here as preserve-when-provable, refuse-nothing, and leave the unprovable case unowned as today. Accepting or rejecting that shape is the maintainer call.

### DIFF
--- a/src/config/sessions/session-accessor.conformance.test.ts
+++ b/src/config/sessions/session-accessor.conformance.test.ts
@@ -2,10 +2,18 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { afterEach, beforeEach, describe, expect, it, onTestFinished, vi } from "vitest";
+import { prepareSystemAgentRunAdmission } from "../../agents/admitted-run-context.js";
 import {
   readPersistedAuthProfileStateRaw,
   writePersistedAuthProfileStateRaw,
 } from "../../agents/auth-profiles/sqlite.js";
+import { prepareCliHistoryBoundary } from "../../agents/cli-runner/history-boundary.js";
+import {
+  buildCliSessionHistoryPrompt,
+  loadCliSessionPromptContext,
+} from "../../agents/cli-runner/session-history.js";
+import type { PreparedCliRunContext } from "../../agents/cli-runner/types.js";
+import { CURRENT_SESSION_VERSION, SessionManager } from "../../agents/sessions/session-manager.js";
 import { executeSqliteQueryTakeFirstSync, getNodeSqliteKysely } from "../../infra/kysely-sync.js";
 import { resetPluginRuntimeStateForTest, setActivePluginRegistry } from "../../plugins/runtime.js";
 import { beginSessionWorkAdmission } from "../../sessions/session-lifecycle-admission.js";
@@ -24,8 +32,11 @@ import {
   readSessionArchiveContentSync,
 } from "./archive-compression.js";
 import { isSessionArchiveArtifactName } from "./artifacts.js";
+import { runWithCliHistoryWriter, type CliHistoryWriter } from "./cli-history-boundary.js";
+import { getCliSessionBinding } from "./cli-session-binding.js";
 import {
   appendTranscriptEvent,
+  appendTranscriptEventSync,
   appendTranscriptMessage,
   cleanupSessionLifecycleArtifactsCore,
   listSessionEntriesCore,
@@ -62,6 +73,7 @@ import {
 } from "./session-accessor.sqlite-entry.js";
 import { forkSessionEntryFromParentTarget } from "./session-accessor.sqlite-parent-session.js";
 import { loadTranscriptEventsSync } from "./session-accessor.sqlite-read.js";
+import { readSessionTranscriptWatermark } from "./session-accessor.sqlite-transcript-watermark.js";
 import { replaceTranscriptEvents } from "./session-accessor.sqlite-transcript-write.js";
 import { setCanonicalSqliteSessionMainKey } from "./session-canonical-key.js";
 import type { InternalSessionEntry, SessionCompactionCheckpoint, SessionEntry } from "./types.js";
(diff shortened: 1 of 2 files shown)


yetval:
Rebased onto current main (`db2666090f1`) and addressed the verdict. New head `8de35315747`.

**[P1] Preserve authorized checkpoint history before clearing native continuity** - fixed, and confirmed as a real defect first. Driving the real runtime through the reviewed shape reproduces exactly what the finding predicted: with the binding cleared and the copied `cliHistoryBoundary` still naming the source `sessionId` and generation, `prepareCliHistoryBoundary` refuses, `prepare.ts` selects `auth-unknown`, and the successor's first turn opens with `restoredMessages=0`.

The repair stays in the same clone. It stops inheriting the source's history proof, and re-stamps it onto the forked transcript only when the source boundary actually covers every copied row: known proof naming the source session, fork read from that same session rather than a legacy or cross-session position, generation still live, and `maxSeq` at or beyond the highest seq the fork read. The successor keeps the source `authFingerprint`, so the account check is untouched. Anything short of that leaves the successor unowned instead of carrying a stale proof. Nothing in the CLI runner's admission rules was relaxed.

**[P1] Resolve merge risk** - answered in the body's proof block. Both successors now open a fresh native conversation and receive the checkpoint conversation through the normal `missing-transcript` reseed route. The same successor turn driven under a different account credential is still refused, so the boundary still does the job it exists for.

**Add real behavior proof** - the proof now drives the next CLI turn, not just the store mutation. Four runs on the same inputs: pristine main, the binding-clear-only shape you reviewed, this patch, and this patch under a second account. It uses the real checkpoint store, real `prepareSystemAgentRunAdmission`, real `prepareCliHistoryBoundary`, real `loadCliSessionPromptContext`, and the real `getCliSessionBinding` lookup for what the successor would resume.

**Complete next step (P2)** - covered by the two items above.

Still deliberately out of scope, unchanged from before: the message-cut sibling also clears `agentHarnessId`, `contextTokens`, `contextBudgetStatus` and `compactionCount`. Whether a checkpoint successor should drop those too is a policy call with its own user-visible consequences, and it is not needed to close this defect. Happy to extend on request.

@clawsweeper re-review

yetval:
Rebased onto current main and pushed. Head is now `fad01fa8d2c`. Item by item:

**[P1] Preserve usable history when restoring a retained checkpoint again** - fixed, `src/config/sessions/session-accessor.sqlite-checkpoint.ts`. A restore retains the checkpoint positions, which keep naming the transcript the checkpoint was captured on, while the successor owns only its own copy of those rows. The fork source is now resolved against the transcript the live session owns before the recorded position, and only when an exact leaf id bounds it to the same rows. That took the first of the two directions you offered: verifiable ownership is preserved, nothing is refused, and no safeguard was touched. The account fingerprint, generation, full-coverage and live-run checks all still gate the transfer, and a source that cannot satisfy them is still left unowned. A position with no exact leaf keeps naming the recorded session, because reading the live session unbounded would copy the rows the checkpoint exists to drop.

**Add real behavior proof (authority chain)** - the proof in the body now runs past preparation to the execution boundary, on pristine main, on the previously reviewed head, and on this head with identical inputs. It shows branch, restore and repeated restore, the reseeded prompt the backend actually receives in each case, the rejection of a revoked successor run by the history writer's `assertReadable` before any send, and the empty prompt a second account receives. The middle run reproduces this finding on `8de35315747` exactly as described.

**Resolve merge risk: transferred history unavailable to another account or a revoked run (P1)** - answered by the last two lines of the after block, both driven on the twice-restored successor, and covered by a new regression case that drives the same real preparation path.

**Resolve merge risk: the durable history-transfer contract lacks recorded maintainer acceptance (P1)** - still open, and left to you rather than worked around. Both options in the verdict remain real. Option 1, accept the bounded transfer, is what this PR implements: ownership moves only when the source proof is known, names the source session, matches its live generation and covers every copied row, and the copied rows come from a transcript the live session owns. Option 2, pause the ownership transfer, is a straight cut back to clearing the binding only and accepting that every checkpoint successor starts with no CLI history. If option 2 is the contract you want, say so and I will reduce the patch to that shape.

**Complete next step (P2)** - repeated-restore continuity is fixed, the final-execution evidence is in the body, and the owner acceptance above is the only remaining gate.

**Resolve maintainer decision** - answered as preserve-when-provable, refuse nothing, and leave the unprovable source unowned as today. That last case, an imported or legacy position with no locatable leaf, stays inside the contract accepted in #140294 and is named in the body as a known residual rather than widened here.

Regression coverage added: a repeated restore keeps an owned boundary matching its own transcript with the copied rows present, and a real preparation-path case on the twice-restored successor covering the owning account, the revoked run and the second account. Both new cases fail on `8de35315747` and pass on this head.

PR #144143, opened 2026-09-10
fix(secrets): keep store entry kind when rotating a value

yetval:
## Summary

`openclaw secrets store set <NAME> --value-file <path>` silently converts a write-only `secret` entry into a readable `env` entry when only the value is rotated.

The trigger is the ordinary credential-rotation path: an entry that was explicitly created with `--kind secret` under a name the automatic suffix heuristic does not classify as sensitive (`OPENAI_KEY`, `ANTHROPIC_ADMIN_KEY`, `GH_PAT`, `AWS_ACCESS_KEY_ID`), rotated with no `--kind` and no `--allow-host` / `--clear-allowed-hosts`.

After the rotation the entry is no longer write-only. `openclaw secrets store list` prints a preview of the rotated credential, `openclaw secrets store get` returns it instead of refusing, every agent subprocess receives it as plaintext in its environment rather than a sealed sentinel, and the `--allow-host` egress destination allowlist that bound the secret is deleted. Nothing warns the operator; `set` reports `Stored OPENAI_KEY (env).` and exits 0.

`openclaw secrets store import` has the same root cause and reclassifies every existing entry it rewrites.

## Root cause

`src/cli/secrets-store-cli.ts` looks the existing entry up only when a host-policy flag was passed, so the resolved kind falls back to the name heuristic on a plain value rotation:

```ts
// before
const requestedHosts = options.allowHost ?? [];
const hostPolicyRequested = requestedHosts.length > 0 || options.clearAllowedHosts === true;
const existingEntry = hostPolicyRequested
  ? storeModule.listSecretStoreEntries({ scope }).find((entry) => entry.name === name)
  : undefined;
const kind = options.kind
  ? storeKind(options.kind, name)
  : (existingEntry?.kind ?? storeKind(undefined, name));
```

With no host flags `existingEntry` is `undefined`, so `storeKind(undefined, name)` applies `isSensitiveEnvName` (`src/secrets/secret-env-name.ts`, `/_?(API_KEY|TOKEN|PASSWORD|PRIVATE_KEY|SECRET)$/i`). `OPENAI_KEY` does not match, so the operator's explicit `--kind secret` is replaced by `"env"`. `import` never looks an existing entry up at all.

The upsert in `writeSecretStoreEntryInternal` then rewrites the stored kind and, for `env`, nulls the host allowlist. Every consumer branches on the stored `row.kind`, so one wrong resolution breaks four boundaries at once: `readSecretStoreExecEnvironment` hands `env` rows to agent subprocesses as plaintext instead of sealing a sentinel and registering an egress binding, `store get` drops its write-only refusal, `renderList` prints `valuePreview`, and the refusal of `--value` on argv stops applying.

The conditional lookup shows the intent was already to inherit the stored kind; it was scoped to the host-policy branch by accident.

## Fix

Both CLI paths now inherit the stored kind, and that inheritance is resolved inside the authoritative store write transaction rather than captured in the CLI before it awaits input.

`SecretStoreWriteParams` gains an optional `inheritExistingKind`. When set, the store reads the live row inside the transaction it already opens, resolves the kind from that row, and drives value validation, the allowed-host refusal, and the `allowed_hosts` clearing from the resolved kind rather than the requested one:

```ts
// after
const kind =
  inheritExistingKind && previous ? (previous.kind as SecretStoreKind) : entry.kind;
if (inheritExistingKind) {
  assertSecretStoreWriteShape(entry.value, kind, entry.allowedHosts);
}
```

`set` and `import` pass `inheritExistingKind: options.kind === undefined`, and `writeSecretStoreEntry` returns the kind it committed so `set` reports what actually landed.

This is correct because the name heuristic is a classifier for names that have never been seen before. Once an entry exists, its stored kind is a recorded operator choice and outranks a guess made from the name. An explicit `--kind` still wins over both, in either direction, so the operator keeps full control.

Resolving at the write boundary rather than at the start of the command closes the replay window. `set` awaits `readSecretStoreInput` and `import` awaits `confirmMutation`, and both waits are human-scale. A kind captured before those waits could be replayed over a protection change committed during them, which would downgrade a freshly protected entry and delete the host allowlist the operator had just attached. Because the resolution now happens inside the same transaction as the insert, no store state can change between reading the kind and writing the row.

### Import writes its batch under one transaction

Resolving the kind at write time means a value the CLI already accepted can become invalid by the time it is written: an empty value is legal for an `env` entry and refused for a `secret` one, so an entry that a concurrent protection change turns into a secret must fail. `import` previously wrote one entry per transaction, so that failure could arrive after earlier entries of the same batch had already committed, which breaks the no-write-on-validation-failure contract the existing empty-secret refusal in `src/cli/secrets-store-cli.test.ts` relies on.

`import` now hands its whole batch to the store as one call. `writeSecretStoreEntries` resolves every live kind and validates every resulting entry before it writes any of them, inside the single `runOpenClawStateWriteTransaction` it already opened:

```ts
const resolved = params.entries.map((entry) => {
  /* live row lookup, kind resolution and assertSecretStoreWriteShape */
});
for (const { entry, kind, allowedHosts } of resolved) {
  /* upsert */
}
```

`writeSecretStoreEntry` is now a single-entry call into the same code, so the two paths cannot drift, and `writeSecretStoreEntryWithRollback` keeps its previous-value snapshot unchanged.

The CLI is the only kind resolver in the codebase. The other writer of the store, the Gateway RPC `secrets.store.set`, takes `kind` as a required field of `SecretsStoreSetParamsSchema` and passes it straight through, so it never infers a kind. It does not set `inheritExistingKind`, so its behavior is unchanged and `writeSecretStoreEntry` still receives a kind its caller decided.

I considered instead refusing any `secret -> env` transition inside the store. That is wrong: an explicit `--kind env` downgrade is a legitimate operator action, and the store cannot tell an explicit downgrade from an inferred one without being told. `inheritExistingKind` is exactly that missing signal, and it keeps the policy choice with the CLI while giving the store the atomicity only the store can provide. The same reasoning puts the batch in the store rather than in the CLI: only the store owns the transaction, and a CLI-side pre-check would just recreate the window it is meant to close.

`import` is fixed in the same commit rather than left as a one-sided fix; it shares the root cause and reclassifies every existing entry it rewrites. `docs/cli/secrets.md` states that automatic detection applies to new names, that an existing entry keeps its kind, and that the inherited kind is resolved as the write commits.

## Verification

- `src/cli/secrets-store-cli.kind-inheritance.test.ts` drives the real commander program through `registerSecretsCli` against a real SQLite store in a temp directory. Eight cases pass: value-only rotation keeps a secret write-only and host-bound; an explicit `--kind env` still downgrades; a brand-new name is still classified from its name; `import` keeps an existing secret while classifying new names; a protection change made while `set` waits for its value is applied; the same while `import` waits for confirmation; a protection change that invalidates a later entry of an import batch leaves every entry of that batch unwritten; and an explicit `--kind` is still honored across that interleaving.
- The three interleaving cases fail on earlier heads of this PR and pass here, so they are regressions against the exact defects reported rather than restatements of the fix. Reverting only the batch call site to the previous per-entry loop fails the new case with `expected { name: 'SERVICE_MODE', ... } to be undefined`.
- `src/secrets/store/secret-store.test.ts` passes against the changed store, covering the existing empty-secret refusal, the allowed-host validation, the exec projection, and the hidden GitHub record paths.
- `src/cli/secrets-store-cli.test.ts` passes 15 of 15. Its import assertion now reads the single batch call instead of three separate writes, and its empty-secret refusal additionally asserts that the batch write was never reached.
- No existing assertion depended on the old fallback, so nothing had to be relaxed or rewritten.
- `tsgo` and `oxlint` were not run on this host: this box cannot hold them without exhausting memory. Only the repository formatter and the two targeted suites above were run here, so a type error confined to new code would first surface in the repository's automated checks rather than locally.
- Values used throughout are synthetic (`sk-protected-credential`, `sk-rotated-credential`, `plain-original-value`, `production-next`).

## Real behavior proof

Behavior addressed: a value-only `openclaw secrets store set` downgraded a write-only `secret` entry to a readable `env` entry, handed the credential to agent subprocesses as plaintext, and deleted its `--allow-host` egress allowlist. Inheriting the kind at write time then introduced a second failure, reported as the remaining review finding: `openclaw secrets store import` could commit an earlier entry of its batch and then fail on a later one whose kind had just been changed by another process.
Real environment tested: the real `openclaw secrets store` command line entry point (`src/entry.ts`) run as separate processes against a real SQLite secret store in a scratch `OPENCLAW_STATE_DIR`, plus a real child process spawned from the environment `readSecretStoreExecEnvironment` resolves, which is the same projection `createExecTool` gives agent commands in `src/agents/bash-tools.exec-run.ts`. The interfering process is a second real `openclaw secrets store set` run while the first command holds its confirmation prompt on a pseudo terminal. Captured on pristine main `d23730ee7c54975df5f50294dca34a06592b3590`, on the previous head of this PR `13c7014552c3e43dd8f76a235c1aee6264e8c8d2`, and on this head `66a8c01c8db76bf63ab79c506880f462856b2d1e`, with identical inputs and identical scripted timing. Only the credential values are synthetic.
Exact steps or command run after this patch: scenario A creates `OPENAI_KEY` with `--kind secret --allow-host api.openai.com`, rotates it with `openclaw secrets store set OPENAI_KEY --value-file <rotated>`, then spawns a real child from the resolved agent environment. Scenario B creates `SERVICE_API_KEY` as `--kind env`, starts `openclaw secrets store import --from values.env` where `values.env` holds `SERVICE_MODE=production-next` and an empty `SERVICE_API_KEY=`, leaves it waiting at its confirmation prompt for 12 seconds, protects `SERVICE_API_KEY` from a second process with `--kind secret --allow-host api.example.com` during that wait, answers the prompt, and then lists the store. Scenario C imports the same file with `--yes` and no interference. Each listing prints name, kind, allowed hosts and whether the value is withheld.
Evidence after fix:
```
######## pristine main d23730ee7c54
### tree: base (d23730ee7c54975df5f50294dca34a06592b3590)
--- A. protect OPENAI_KEY, then rotate only its value
Stored OPENAI_KEY (secret).
  after create:
   OPENAI_KEY kind=secret allowedHosts=["api.openai.com"] valuePreview=(withheld)
Stored OPENAI_KEY (env).
  after value-only rotation:
   OPENAI_KEY kind=env allowedHosts=[] valuePreview="sk-rotated-credential"
  what a real child process receives:
   child process sees OPENAI_KEY: PLAINTEXT sk-rotated-credential
   sealed sentinel offered: false
   egress bindings: []
--- B. import while a second process protects one of its names
Stored SERVICE_API_KEY (env).
  before import:
   SERVICE_API_KEY kind=env allowedHosts=[] valuePreview="plain-original-value"
  second process protects SERVICE_API_KEY while the import prompt is pending:
   Stored SERVICE_API_KEY (secret).
  store while the prompt is still pending:
   SERVICE_API_KEY kind=secret allowedHosts=["api.example.com"] valuePreview=(withheld)
  import command exit code: 2
  import command output: Secret store value is empty. Secret entries require a value; check the command that produced it.
  store after the import command returned:
   SERVICE_API_KEY kind=secret allowedHosts=["api.example.com"] valuePreview=(withheld)

######## previous head of this PR 13c7014552c3
### tree: prev (13c7014552c3e43dd8f76a235c1aee6264e8c8d2)
--- A. protect OPENAI_KEY, then rotate only its value
Stored OPENAI_KEY (secret).
  after create:
   OPENAI_KEY kind=secret allowedHosts=["api.openai.com"] valuePreview=(withheld)
Stored OPENAI_KEY (secret).
  after value-only rotation:
   OPENAI_KEY kind=secret allowedHosts=["api.openai.com"] valuePreview=(withheld)
  what a real child process receives:
   child process sees OPENAI_KEY: absent
   sealed sentinel offered: true
   egress bindings: [{"name":"OPENAI_KEY","allowedHosts":["api.openai.com"]}]
--- B. import while a second process protects one of its names
Stored SERVICE_API_KEY (env).
  before import:
   SERVICE_API_KEY kind=env allowedHosts=[] valuePreview="plain-original-value"
  second process protects SERVICE_API_KEY while the import prompt is pending:
   Stored SERVICE_API_KEY (secret).
  store while the prompt is still pending:
   SERVICE_API_KEY kind=secret allowedHosts=["api.example.com"] valuePreview=(withheld)
  import command exit code: 2
  import command output: Secret store value is empty. Secret entries require a value; check the command that produced it.
  store after the import command returned:
   SERVICE_API_KEY kind=secret allowedHosts=["api.example.com"] valuePreview=(withheld)
   SERVICE_MODE kind=env allowedHosts=[] valuePreview="production-next"

######## this head 66a8c01c8db76bf63ab79c506880f462856b2d1e
### tree: new (66a8c01c8db76bf63ab79c506880f462856b2d1e)
--- A. protect OPENAI_KEY, then rotate only its value
Stored OPENAI_KEY (secret).
  after create:
   OPENAI_KEY kind=secret allowedHosts=["api.openai.com"] valuePreview=(withheld)
Stored OPENAI_KEY (secret).
  after value-only rotation:
   OPENAI_KEY kind=secret allowedHosts=["api.openai.com"] valuePreview=(withheld)
  what a real child process receives:
   child process sees OPENAI_KEY: absent
   sealed sentinel offered: true
   egress bindings: [{"name":"OPENAI_KEY","allowedHosts":["api.openai.com"]}]
--- B. import while a second process protects one of its names
Stored SERVICE_API_KEY (env).
  before import:
   SERVICE_API_KEY kind=env allowedHosts=[] valuePreview="plain-original-value"
  second process protects SERVICE_API_KEY while the import prompt is pending:
   Stored SERVICE_API_KEY (secret).
  store while the prompt is still pending:
   SERVICE_API_KEY kind=secret allowedHosts=["api.example.com"] valuePreview=(withheld)
  import command exit code: 2
  import command output: Secret store value is empty. Secret entries require a value; check the command that produced it.
  store after the import command returned:
   SERVICE_API_KEY kind=secret allowedHosts=["api.example.com"] valuePreview=(withheld)

######## this head 66a8c01c8db76bf63ab79c506880f462856b2d1e, scenario C: the same import with no interference
Stored OPENAI_KEY (secret).
Imported 2 team store entries.
   OPENAI_KEY kind=secret allowedHosts=["api.openai.com"] valuePreview=(withheld)
   SERVICE_MODE kind=env allowedHosts=[] valuePreview="production-next"
```
Observed result after fix: in scenario A the rotated entry stays `secret` on this head instead of flipping to `env`, keeps its `api.openai.com` egress binding instead of losing it, withholds the value from the listing, and hands the spawned child nothing rather than `sk-rotated-credential`. In scenario B all three runs exit 2 with the same message, but the store afterwards differs: pristine main and this head contain only the protected `SERVICE_API_KEY`, while the previous head of this PR also contains `SERVICE_MODE=production-next`, a durable write from a command that failed. This head restores the all-or-nothing behavior pristine main had, now with the kind resolved from the live row. Scenario C shows the batch still writes every entry of an uninterfered import, so the repair did not trade partial writes for no writes.
What was not tested: only the `team` scope was exercised, because it is the only scope the command line accepts today. The egress proxy was not driven end to end against a live upstream host, so the proof shows the restored binding and its allowed hosts rather than an intercepted outbound request. The Gateway RPC path was read but not driven, since it carries an explicit kind and does not request inheritance.

### DIFF
--- a/docs/cli/secrets.md
+++ b/docs/cli/secrets.md
@@ -61,7 +61,7 @@ Naming and value rules:
 - Names must match `^[A-Z][A-Z0-9_]{0,127}$`.
 - Values are limited to 64 KiB (65,536 UTF-8 bytes). An oversized value exits `2` whether it arrives from stdin, `--value`, or `--value-file`.
 - A `secret` entry may not be empty, because an empty credential cannot be diagnosed later. `get` refuses secret kinds, and listings mask them. `env` entries may be empty.
-- `--kind secret|env` overrides automatic kind detection. Otherwise names ending in a common credential suffix such as `_API_KEY`, `_TOKEN`, `_PASSWORD`, `_PRIVATE_KEY`, or `_SECRET` become `secret`, and other names become `env`.
+- `--kind secret|env` overrides automatic kind detection. Otherwise `set` and `import` keep the kind an existing entry already has, so rotating a value never changes it. That inherited kind is resolved as the write commits, so protecting an entry while a value or confirmation is still pending is never undone by the pending write. A new name ending in a common credential suffix such as `_API_KEY`, `_TOKEN`, `_PASSWORD`, `_PRIVATE_KEY`, or `_SECRET` becomes `secret`, and any other new name becomes `env`.
 
 ### Set values safely
 
--- a/src/cli/secrets-store-cli.kind-inheritance.test.ts
+++ b/src/cli/secrets-store-cli.kind-inheritance.test.ts
@@ -1,0 +1,413 @@
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { Command } from "commander";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { closeOpenClawStateDatabaseForTest } from "../state/openclaw-state-db.js";
+import { registerSecretsCli } from "./secrets-cli.js";
+
+const mocks = await vi.hoisted(async () => {
+  const { createCliRuntimeMock } = await import("./test-runtime-mock.js");
+  return {
+    ...createCliRuntimeMock(vi),
+    database: { path: "" } as { path: string },
+    interleave: { run: undefined as (() => Promise<void>) | undefined },
+  };
+});
+
+vi.mock("../runtime.js", () => ({ defaultRuntime: mocks.defaultRuntime }));
+vi.mock("./one-shot-exit.js", () => ({
+  exitCliAfterOutput: (runtime: typeof mocks.defaultRuntime, exitCode: number) =>
+    runtime.exit(exitCode),
+}));
+vi.mock("../infra/gateway-lock.js", () => ({
+  readActiveGatewayLockIdentity: () => Promise.resolve(undefined),
+}));
+vi.mock("./secrets-store-input.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("./secrets-store-input.js")>();
+  return {
+    ...actual,
+    readSecretStoreInput: async (params: Parameters<typeof actual.readSecretStoreInput>[0]) => {
+      const value = await actual.readSecretStoreInput(params);
+      await runPendingInterleave();
+      return value;
+    },
+  };
+});
+vi.mock("@clack/prompts", () => ({
+  confirm: async () => {
+    await runPendingInterleave();
+    return true;
+  },
+  isCancel: () => false,
+}));
+
+async function runPendingInterleave(): Promise<void> {
+  const pending = mocks.interleave.run;
+  mocks.interleave.run = undefined;
+  await pending?.();
+}
+vi.mock("../secrets/store/secret-store.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../secrets/store/secret-store.js")>();
+  const withDb = <T extends { database?: unknown }>(params: T) => ({
+    ...params,
+    database: mocks.database,
+  });
+  return {
+    ...actual,
+    listSecretStoreEntries: (p: Parameters<typeof actual.listSecretStoreEntries>[0]) =>
+      actual.listSecretStoreEntries(withDb(p)),
+    writeSecretStoreEntry: (p: Parameters<typeof actual.writeSecretStoreEntry>[0]) =>
+      actual.writeSecretStoreEntry(withDb(p)),
+    writeSecretStoreEntries: (p: Parameters<typeof actual.writeSecretStoreEntries>[0]) =>
+      actual.writeSecretStoreEntries(withDb(p)),
+    updateSecretStoreAllowedHosts: (
+      p: Parameters<typeof actual.updateSecretStoreAllowedHosts>[0],
+    ) => actual.updateSecretStoreAllowedHosts(withDb(p)),
+    readSecretStoreValue: (p: Parameters<typeof actual.readSecretStoreValue>[0]) =>
+      actual.readSecretStoreValue(withDb(p)),
+    deleteSecretStoreEntry: (p: Parameters<typeof actual.deleteSecretStoreEntry>[0]) =>
+      actual.deleteSecretStoreEntry(withDb(p)),
+    purgeExpiredSecretStoreEntries: () =>
+      actual.purgeExpiredSecretStoreEntries({ database: mocks.database }),
+  };
+});
+
+const { listSecretStoreEntries, readSecretStoreExecEnvironment } =
+  await import("../secrets/store/secret-store.js");
+
+const scope = { kind: "team" } as const;
+const roots: string[] = [];
+
+function createProgram(): Command {
+  const program = new Command();
+  program.exitOverride();
+  registerSecretsCli(program);
+  return program;
+}
+
+function createStoreRoot(): string {
+  const root = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), "openclaw-store-kind-")));
+  roots.push(root);
+  mocks.database.path = path.join(root, "state.sqlite");
+  return root;
+}
+
+function writeValueFile(root: string, fileName: string, value: string): string {
+  const filePath = path.join(root, fileName);
+  fs.writeFileSync(filePath, value);
+  return filePath;
+}
+
+function entryFor(name: string) {
+  return listSecretStoreEntries({ scope, database: mocks.database }).find(
+    (entry) => entry.name === name,
+  );
+}
+
+async function protectEntry(name: string, valueFile: string, host: string): Promise<void> {
+  await createProgram().parseAsync(
+    [
+      "secrets",
+      "store",
+      "set",
+      name,
+      "--kind",
+      "secret",
+      "--value-file",
+      valueFile,
+      "--allow-host",
+      host,
+    ],
+    { from: "user" },
+  );
+}
+
+function exposureFor(name: string) {
+  const execEnvironment = readSecretStoreExecEnvironment({
+    includeSecretSentinels: true,
+    database: mocks.database,
+  });
+  return {
+    kind: entryFor(name)?.kind,
+    allowedHosts: entryFor(name)?.allowedHosts,
+    valuePreview: entryFor(name)?.valuePreview,
+    plaintextInSubprocessEnv: execEnvironment.env?.[name],
+    sealedSentinel: execEnvironment.secretSentinels?.[name] !== undefined,
+    egressBindings: execEnvironment.secretEgressBindings?.length ?? 0,
+  };
+}
+
+afterEach(() => {
+  mocks.interleave.run = undefined;
+  closeOpenClawStateDatabaseForTest();
+  mocks.runtimeLogs.length = 0;
+  mocks.runtimeErrors.length = 0;
+  for (const root of roots.splice(0)) {
+    fs.rmSync(root, { recursive: true, force: true });
+  }
+});
+
+describe("secrets store kind inheritance", () => {
+  it("keeps a stored secret write-only and host-bound when only its value is rotated", async () => {
+    const root = createStoreRoot();
+    const original = writeValueFile(root, "original.txt", "sk-original-credential");
+    const rotated = writeValueFile(root, "rotated.txt", "sk-rotated-credential");
+
+    await createProgram().parseAsync(
+      [
+        "secrets",
+        "store",
+        "set",
+        "OPENAI_KEY",
+        "--kind",
+        "secret",
+        "--value-file",
+        original,
+        "--allow-host",
+        "api.openai.com",
+      ],
+      { from: "user" },
+    );
+    expect(entryFor("OPENAI_KEY")).toMatchObject({
+      kind: "secret",
+      allowedHosts: ["api.openai.com"],
+    });
+
+    await createProgram().parseAsync(
+      ["secrets", "store", "set", "OPENAI_KEY", "--value-file", rotated],
+      { from: "user" },
+    );
+
+    const execEnvironment = readSecretStoreExecEnvironment({
+      includeSecretSentinels: true,
+      database: mocks.database,
+    });
+    expect({
+      kind: entryFor("OPENAI_KEY")?.kind,
+      allowedHosts: entryFor("OPENAI_KEY")?.allowedHosts,
+      valuePreview: entryFor("OPENAI_KEY")?.valuePreview,
+      plaintextInSubprocessEnv: execEnvironment.env?.OPENAI_KEY,
+      sealedSentinel: execEnvironment.secretSentinels?.OPENAI_KEY !== undefined,
+      egressBindings: execEnvironment.secretEgressBindings?.length ?? 0,
+    }).toEqual({
+      kind: "secret",
+      allowedHosts: ["api.openai.com"],
+      valuePreview: undefined,
+      plaintextInSubprocessEnv: undefined,
+      sealedSentinel: true,
+      egressBindings: 1,
+    });
+  });
+
+  it("still downgrades a stored secret when --kind env is explicit", async () => {
+    const root = createStoreRoot();
+    const original = writeValueFile(root, "original.txt", "sk-original-credential");
+    const rotated = writeValueFile(root, "rotated.txt", "sk-rotated-credential");
+
+    await createProgram().parseAsync(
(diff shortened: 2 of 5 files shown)


yetval:
Rebased onto current main and pushed `13c7014552c3e43dd8f76a235c1aee6264e8c8d2`. The body now carries proof pinned to that head. Item by item against the last verdict:

**[P1] Revalidate inherited kinds inside the authoritative write** - fixed, and thank you, the finding was correct. `set` awaits `readSecretStoreInput` and `import` awaits `confirmMutation`, and the previous head captured the inherited kind before both waits, so a protection change committed during one of them was replayed away. `SecretStoreWriteParams` now takes an optional `inheritExistingKind`, and `writeSecretStoreEntryInternal` resolves the kind from the live row inside the transaction it already opens, then drives value validation, the allowed-host refusal, and the `allowed_hosts` clearing from the resolved kind. No store state can change between reading the kind and writing the row. `set` and `import` request inheritance only when the operator passed no `--kind`, so an explicit override still wins in either direction.

**Security: concurrent protection changes can be overwritten** - resolved by the same change, and proved as a final effect rather than by reading the source. Scenario B in the body starts a real `openclaw secrets store set SERVICE_API_KEY --value-file -` process, leaves it blocked on stdin, protects the entry from a second real process with `--kind secret --allow-host api.example.com`, then releases the pending value. On the previous head the pending write restored `env`, deleted the host allowlist, and a real child process spawned from the resolved agent environment received `sk-rotated-credential`. On this head the same interleaving lands the value under the newer `secret` policy, keeps `api.example.com`, and the child receives nothing but a sealed sentinel.

**Add real behavior proof, authority chain** - the body now carries five traces across three trees: pristine main `db2666090f19`, the previous head `e1f3ee8f1efa`, and this head. Each ends with `store list --plain`, `store get`, and a real child process spawned from `readSecretStoreExecEnvironment`, which is the same projection `createExecTool` gives agent commands in `src/agents/bash-tools.exec-run.ts`. That is the plaintext-reaching-a-child-process effect you asked for, on both the sequential rotation and the stale-policy interleaving. The original sequential trace is preserved and re-captured against current main.

**Merge risk: runtime evidence does not cover a protection change during pending input or confirmation** - covered by scenario B above. The interleaving is real rather than simulated: two separate CLI processes against one SQLite store, with the protecting write confirmed committed before the pending value is released.

**Merge risk: incomplete broader CLI and store validation** - re-run and clean on this head. `src/secrets/store/secret-store.test.ts` passes 25 of 25 against the changed store, and `src/cli/secrets-store-cli.test.ts` passes 15 of 15. The two cases I previously reported as timing out were load on my host; both exercise `store list` and `store get` and are unaffected by this change.

**Explicit permitted overrides** - scenario C shows an operator-supplied `--kind env` still downgrades a protected entry and still yields a readable value, so inheritance never traps an entry in its current kind. There is also a regression covering an explicit `--kind` applied across the same interleaving.

Regression coverage: the two interleaving cases in `src/cli/secrets-store-cli.kind-inheritance.test.ts` fail on the previous head and pass here, so they pin the exact defect you identified rather than restating the fix.

**Left to maintainers:** the request for openclaw/openclaw-secops involvement is yours to route; I cannot add reviewers. One intended behavior change is also worth an explicit look, since it is a policy call rather than a bug fix: `openclaw secrets store set <NAME> --value <literal>` against an existing secret entry now hits the pre-existing argv refusal with exit 2 instead of silently downgrading the entry and accepting the credential on the command line.

yetval:
Rebased onto `main` at `d23730ee7c54975df5f50294dca34a06592b3590` and pushed as `66a8c01c8db76bf63ab79c506880f462856b2d1e`. The body carries refreshed real behavior proof pinned to that head. Item by item against the review of `13c7014552c3`:

**[P2] Validate and commit inherited import kinds as one batch (`src/cli/secrets-store-cli.ts:419-424`).** Fixed, and the interleaving you described reproduces exactly as written. `import` no longer loops over `writeSecretStoreEntry`; it hands the whole batch to a new `writeSecretStoreEntries`, which resolves every live kind and runs `assertSecretStoreWriteShape` on every resulting entry before it writes any of them, inside the single `runOpenClawStateWriteTransaction` the store already opened. `writeSecretStoreEntry` is now a one-entry call into the same code, so the single and batch paths cannot drift, and `writeSecretStoreEntryWithRollback` keeps its previous-value snapshot.

Real-process evidence is in the body. With an existing env-kind `SERVICE_API_KEY`, importing `SERVICE_MODE=production-next` plus an empty `SERVICE_API_KEY=` while a second process protects `SERVICE_API_KEY` during the confirmation prompt: on `13c7014552c3` the command exits 2 and `SERVICE_MODE` is durably written; on this head the command exits 2 with the same message and nothing from the batch is written, which matches what pristine main does. An uninterfered `import --yes` still writes every entry.

**Rank-up move, the regression you asked for.** `src/cli/secrets-store-cli.kind-inheritance.test.ts` gains a case that drives the real commander program against a real SQLite store, protects the entry during the confirmation wait, and asserts both that the command fails and that no earlier entry of the batch exists afterwards. Reverting only the batch call site to the previous per-entry loop fails it with `expected { name: 'SERVICE_MODE', ... } to be undefined`. `src/cli/secrets-store-cli.test.ts` is updated in the same commit: its import assertion now reads the single batch call, and its existing empty-secret refusal additionally asserts the batch write was never reached.

**Merge risk (P1), argv value rotations of protected entries.** Unchanged and still intended. `secrets store set <NAME> --value <literal>` against an existing secret entry now hits the pre-existing argv refusal and exits 2 rather than silently downgrading the entry, and the documented alternatives are stdin and `--value-file`. I have not softened this, because softening it is the defect. It needs listed-owner acceptance rather than an author change.

**Decision needed, secops sponsorship.** Still open and still not author-actionable. I cannot add reviewers to this PR, so routing it to `openclaw/openclaw-secops` is a maintainer step. The scope to accept is the two behavior changes above: an existing entry's kind is inherited when no `--kind` is given, and an argv value against a protected entry is refused instead of downgrading it.

**Reviewed head.** The verdict above was written against `13c7014552c3`, which is no longer the head. The prior P1 about resolving inheritance inside the authoritative write stays fixed; this push only adds the batch transaction, its regression, and the proof refresh.

Not run on this host: `tsgo` and `oxlint`, which this box cannot hold without exhausting memory. Only the repository formatter and the two suites named above were run locally, so a type error confined to new code would first surface in the repository's automated checks.

PR #144146, opened 2026-09-10
fix(hooks): claim session memory filenames exclusively

yetval:
## Summary
When one agent serves several chats and they roll over in the same minute, the `session-memory` bundled hook writes every rollover into the same `<workspace>/memory/<YYYY-MM-DD>-<HHMM>.md` file. Each rollover truncates the previous one, so only the last session's context survives, while `Session context saved to ...` is logged for every rollover with the same path. The user loses a memory artifact they can see in their workspace and gets a success message saying it was saved.

Trigger: two or more sessions of one agent cross the daily or idle reset boundary within the same minute, or `/new` is issued in two chats at once. All sessions of an agent resolve the same `memoryDir`, and the default non-LLM slug is minute-resolution, so the candidate filename is identical.

## Root cause
`src/hooks/bundled/session-memory/handler.ts` resolved a free filename with `fs.access` and then committed with the overwriting primitive several `await`s later.

```ts
// before
const filename = await resolveAvailableMemoryFilename({ memoryDir, dateStr, slug });
const memoryFilePath = path.join(memoryDir, filename);

// ... entry construction, root(), provenance observer ...

const commit = () => memoryRoot.write(filename, entry, { encoding: "utf-8" });
await provenanceObserver.write({
  absolutePath: memoryFilePath,
  contentBefore: "",
  contentAfter: entry,
  commit,
});
```

The name is checked at one point and used at another, with real `await` gaps in between (`root(memoryDir)`, provenance reservation, and for the LLM-slug configuration the model call itself). `Root.write` is the overwriting primitive; `Root.create` is the exclusive one. Two rollovers that both observe the name as free both call `write`, and the second truncates the first.

Auto-reset dispatch is detached and unserialized, so the handler runs genuinely overlap: `src/auto-reply/reply/session.ts` calls `emitSessionAutoResetHook`, which fires the internal hook under a bare `void` with no per-sessionKey queue. `src/gateway/session-reset-service.ts` uses the same entry point. The handler itself takes no lock.

## Fix
Publication is one operation: the candidate filename is created exclusively with the entry content already in it, and provenance is reserved before that filename exists.

```ts
// after
for (const candidate of iterateMemoryFilenameCandidates(params.basename)) {
  const candidatePath = path.join(params.memoryDir, candidate);
  if (await memoryCandidateExists(candidatePath)) {
    continue;
  }
  try {
    const outcome = await params.provenanceObserver.createExclusive({
      absolutePath: candidatePath,
      contentAfter: params.entry,
      commit: async () =>
        await params.memoryRoot.create(candidate, params.entry, { encoding: "utf-8" }),
    });
    if (outcome === "occupied") {
      continue;
    }
    return { filename: candidate, filePath: candidatePath };
  } catch (err) {
    if (!(await isOccupiedMemoryCandidate(err, candidatePath))) {
      throw err;
    }
    lastOccupiedError = err;
  }
}
```

`Root.create` passes `overwrite: false`, which reaches `fs.open` with `O_CREAT | O_EXCL` and writes the entry into the file descriptor it just created, so the name appears to readers only with its content in place and its provenance already on record.

The reservation that precedes it is now exclusive too. `claimMemoryArtifactCreateProvenance` writes the record for a candidate path inside the store's write transaction, and only when that path has no record at all:

```ts
const claimed = await store.update(address.storeKey, (current) =>
  current === undefined
    ? { version: 1, /* ... */ originClass: params.originClass, reservationId }
    : undefined,
);
if (!claimed) {
  return { status: "occupied" };
}
```

A capture that loses the reservation never touches the filesystem and has no record to undo; a capture that wins and then loses the exclusive create deletes only its own reservation, because by construction there is no displaced predecessor to restore.

## Why the diff reaches the provenance owner
The previous revision kept the whole change inside the hook and used `MemoryWriteProvenanceObserver.write` for publication. That entry point is built for overwriting an artifact the caller already owns: it reserves unconditionally, keeps the record it displaced, and restores that record if the commit fails. Applied to a brand new path it turns several captures into several simultaneous reservation holders for one absent name, and the `reservationId` guard in `recordMemoryArtifactWriteProvenance` only proves that the rollback owns the current record, not that the predecessor it restores is still valid. With contenders A, B and C reserving in that order, A creates the file, B's rollback is a no-op because C's reservation is current, and C's rollback then restores B's record onto A's file, which is the defect the last review identified.

Locking around that sequence inside the hook would not fix it, because the record that ends up wrong is written by the provenance owner, not by the hook. Making the reservation itself the point of mutual exclusion does fix it, and it belongs with the owner of that record. `src/memory/memory-artifact-provenance.ts` gains one exported claim function, `src/agents/memory-write-provenance.ts` gains the matching `createExclusive` observer method, and `recordMemoryArtifactWriteProvenance`, its rollback semantics and every existing caller of `write` are untouched. The store's `update` runs the comparison and the write in one SQLite write transaction, so the claim holds between processes sharing a workspace as well as inside one.

Exclusivity then lives in one place per fact: the filesystem decides who owns the name, the provenance store decides who may reserve it, and both are asked before the artifact is visible. A stale record for a path whose file was deleted outside the observer makes that one candidate name unavailable and the capture moves to the next candidate, which is the conservative direction.

## Relationship to the accepted sequential fix
This is the concurrent half of a defect whose sequential half is already fixed and accepted. #77749 (`61383aff4b8`, "fix(hooks): avoid session memory filename collisions") introduced the suffix loop precisely because two same-minute rollovers must not share a file. That fix works when the rollovers are ordered, because the first one has already landed on disk by the time the second scans. It does nothing when they overlap. The change here keeps that behavior and its regression coverage intact and closes the overlapping case with the same filename scheme.

The LLM-slug path is covered by construction. The claim is keyed on `memoryBasename`, which is `${dateStr}-${slug}` regardless of whether `slug` came from `generateSlugViaLLM` or from the minute fallback, and the claim sits after slug resolution. The LLM path is in fact the more exposed one on main, since the provider call is an additional and much longer await between the old check and the old use.

Single-session behavior is unchanged: the first candidate is `${basename}.md` with no suffix, and the file the user ends up with is the same one main would have produced.

## Review items from the previous cycle

**Failed claims no longer restore another loser's provenance (the P2 finding, and the session-state merge risk).** This is the change described above. The reservation is exclusive, so one path has one reservation holder at a time, losers return before any filesystem work, and a rollback can only ever delete its own record. The three-contender interleaving the review described is unreachable because B and C never obtain a reservation for A's path. The runtime evidence below drives six overlapping captures per round for fifteen rounds: on the previously reviewed head fifteen of fifteen rounds ended with a provenance record attached to the wrong capture; on this head none do.

**Provenance is still recorded before the artifact exists.** The claim writes the record while the path is absent and the exclusive create publishes the content, so the ordering the previous cycle asked for is preserved. `resolveMemoryPathClassification` answers `untrusted` for a non-owner capture at the first moment the file can be seen, with its content already readable, in every scenario below.

**Regression coverage now forces the interleaving.** `src/hooks/bundled/session-memory/handler-auto-reset.test.ts` gains a case that holds three captures at the claim for one candidate until all three have arrived, then asserts each published artifact carries its own session id, its own content hash and its own origin class. `src/memory/memory-artifact-provenance.test.ts` gains a case for the claim primitive itself: one claim per path, later claims occupied, and a rollback that touches only its own record.

**Sibling handler suite.** `src/hooks/bundled/session-memory/handler.test.ts` changes in one place: its module mock for `src/memory/memory-artifact-provenance.js` now also exports the claim function the handler calls, and the two provenance cases observe that function instead of `recordMemoryArtifactWriteProvenance`. The assertions are otherwise unchanged, including that the destination is absent when provenance is recorded and that no file is written when provenance recording fails. Every other case in that file is untouched.

**Captures survive past the numbered names.** Numbered suffixes run to 64 and are then followed by unique-token candidates, so a date and slug whose numbered names are all occupied keeps its capture instead of throwing it away.

**Occupied candidates are skipped whatever shape they have.** `Root.create` reports an occupied directory as `not-file` and an occupied hard link or symlink as `path-alias`, not `already-exists`. A candidate is treated as occupied when the error carries one of those three codes and an `lstat` of the candidate path confirms something is actually there. A path-security failure with nothing at the candidate path still propagates, so an aliased `memoryDir` is not silently retried 64 times.

**Still a maintainer call, unchanged from the last cycle.** Memory Core resolves an artifact's origin class before it reads the content and never revalidates it at publication (`extensions/memory-core/src/memory/manager-embedding-ops.ts`, with the untracked default at `extensions/memory-core/src/memory/memory-path-provenance.ts:63`). Nothing in this hook can close that; it is a Memory Core ordering decision. The same applies to whether a capture that loses a race and lands on a later filename should ever be able to inherit a conservative classification: with this change it cannot, because each claim starts from an empty record, but the general overwrite path still applies the sticky-untrusted rule by design.

## Verification
- `node scripts/run-vitest.mjs src/hooks/bundled/session-memory/handler-auto-reset.test.ts` on this tree: `Test Files 1 passed (1)`, `Tests 10 passed (10)`, including the new three-contender case.
- `node scripts/run-vitest.mjs src/hooks/bundled/session-memory/handler.test.ts src/memory/memory-artifact-provenance.test.ts src/agents/memory-write-provenance.test.ts`: `26 passed`, `14 passed` and `7 passed`. That covers the sibling handler suite the last review flagged, the provenance owner including its existing rollback cases, and the observer.
- `node_modules/.bin/oxfmt` was run on all six changed files.
- Branch rebased onto `2f97cdbacace6e5db2bea3043d64b8797e338d62`.
- Not run on this host: `tsgo` and `oxlint`. This box cannot hold the typecheck lane in memory, so a type error in the new code or in the new test code would surface only in CI, not locally. No local gate result is claimed beyond the three commands above.

## Real behavior proof
Behavior addressed: sessions of one agent rolling over in the same minute collapsed into one `<workspace>/memory/2026-03-04-0506.md`, losing every capture but the last, and the exclusive claim that repairs that must leave each surviving artifact with its own session identity, its own content hash and its own origin class, and must never let a Memory Core index pass read an untracked memory path and treat a non-owner transcript as trusted.
Real environment tested: the real `session-memory` hook handler from `src/hooks/bundled/session-memory/handler.ts` driven through the real `createInternalHookEvent` dispatch over real temporary workspace directories, interleaved in one process with the real Memory Core path classifier `resolveMemoryPathClassification` from `extensions/memory-core/src/memory/memory-path-provenance.ts` called in the order `prepareIndexEntry` calls it; run on pristine main `2f97cdbacace6e5db2bea3043d64b8797e338d62`, then on the head reviewed at revision 4 `428274b9b8577cfa200fd5f704c3b8fe558d4331`, then on this patched tree `c0c0a0a9fdbd10187c69b0512becefd54be4b44c`, with identical inputs and a pinned rollover timestamp. Transcripts were written through the real `replaceTranscriptEvents` session store, the real fs-safe `Root` performed every filesystem mutation, the real memory provenance observer wrapped every commit, and provenance was read back through the real `listMemoryArtifactProvenance` against a real state directory. The only substitution is module resolution: the Memory Core plugin's `openclaw/plugin-sdk/*` specifiers were resolved to the same in-repo host SDK source modules rather than to `dist`, because this checkout has no build output.
Exact steps or command run after this patch: `OPENCLAW_STATE_DIR=<scratch>/state HOME=<scratch>/home TZ=UTC node --import tsx --import ./register.mjs ./probe.mts` and `... PROOF_CONTENDERS=6 PROOF_ROUNDS=15 node --import tsx --import ./register.mjs ./sweep.mts`. The first driver runs four scenarios against separate workspaces at the pinned timestamp `2026-03-04T05:06:07.000Z`: an owner rollover alone, a non-owner rollover alone, an owner rollover concurrent with a non-owner rollover, and three overlapping rollovers in one workspace; in each scenario an index pass opens on every `.md` path as it becomes visible and resolves its origin the way `prepareIndexEntry` does before reading content. The second driver raises the overlap to six concurrent rollovers per workspace and repeats it fifteen times, then reports every published artifact whose recorded session id or recorded content hash does not belong to the capture actually stored in that file.
Evidence after fix:
```
# BEFORE (pristine main 2f97cdbacace6e5db2bea3043d64b8797e338d62)
[three overlapping rollovers, same minute]
  artifacts kept: 1 -> 2026-03-04-0506.md
  2026-03-04-0506.md: carries=GAMMA-ONLY originClassAtFirstSight=untrusted bytesAtFirstSight=219
    recorded provenance: originClass=untrusted sessionId=gamma-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  ALPHA-ONLY retained copies: 0
  BETA-ONLY retained copies: 0
  GAMMA-ONLY retained copies: 1
overlapping rollovers per round: 6, rounds: 15, rounds with a provenance record attached to the wrong capture: 0, captures lost overall: 75

# BEFORE (head reviewed at revision 4, 428274b9b8577cfa200fd5f704c3b8fe558d4331, identical inputs)
[three overlapping rollovers, same minute]
  artifacts kept: 3 -> 2026-03-04-0506-2.md, 2026-03-04-0506-3.md, 2026-03-04-0506.md
  2026-03-04-0506-2.md: carries=BETA-ONLY originClassAtFirstSight=untrusted bytesAtFirstSight=216
    recorded provenance: originClass=untrusted sessionId=beta-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  2026-03-04-0506-3.md: carries=GAMMA-ONLY originClassAtFirstSight=agent bytesAtFirstSight=219
    recorded provenance: originClass=agent sessionId=gamma-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  2026-03-04-0506.md: carries=ALPHA-ONLY originClassAtFirstSight=agent bytesAtFirstSight=219
    recorded provenance: originClass=agent sessionId=alpha-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  ALPHA-ONLY retained copies: 1
  BETA-ONLY retained copies: 1
  GAMMA-ONLY retained copies: 1
overlapping rollovers per round: 6, rounds: 15, rounds with a provenance record attached to the wrong capture: 15, captures lost overall: 0
  round 1 2026-03-04-0506-2.md: carries=R1C1 recordedSessionId=round-1-contender-4-session recordDescribesThisFile=false originClass=untrusted
  round 1 2026-03-04-0506-3.md: carries=R1C2 recordedSessionId=round-1-contender-4-session recordDescribesThisFile=false originClass=untrusted
  round 1 2026-03-04-0506-4.md: carries=R1C3 recordedSessionId=round-1-contender-4-session recordDescribesThisFile=false originClass=untrusted
  round 2 2026-03-04-0506-2.md: carries=R2C1 recordedSessionId=round-2-contender-4-session recordDescribesThisFile=false originClass=untrusted

# AFTER (this patch at c0c0a0a9fdbd10187c69b0512becefd54be4b44c, identical inputs)
[owner rollover] owner=true
  index pass opened on first visible artifact: name=2026-03-04-0506.md originClass=agent bytesAtClassify=219
  artifact after hook completed: name=2026-03-04-0506.md bytes=219 carries_OWNER-ONLY=true
  recorded provenance: path=memory/2026-03-04-0506.md originClass=agent
  index pass would publish: originClass=agent withTranscriptContent=true automaticContextEligible=true
[non-owner rollover] owner=false
  index pass opened on first visible artifact: name=2026-03-04-0506.md originClass=untrusted bytesAtClassify=228
  artifact after hook completed: name=2026-03-04-0506.md bytes=228 carries_INTRUDER-ONLY=true
  recorded provenance: path=memory/2026-03-04-0506.md originClass=untrusted
  index pass would publish: originClass=untrusted withTranscriptContent=true automaticContextEligible=false
[concurrent owner and non-owner rollover, same minute]
  artifacts kept: 2 -> 2026-03-04-0506-2.md, 2026-03-04-0506.md
  2026-03-04-0506-2.md: carries=INTRUDER-ONLY originClassAtFirstSight=untrusted bytesAtFirstSight=238
    recorded provenance: originClass=untrusted sessionId=race-intruder-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  2026-03-04-0506.md: carries=OWNER-ONLY originClassAtFirstSight=agent bytesAtFirstSight=229
    recorded provenance: originClass=agent sessionId=race-owner-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  OWNER-ONLY retained copies: 1
  INTRUDER-ONLY retained copies: 1
[three overlapping rollovers, same minute]
  artifacts kept: 3 -> 2026-03-04-0506-2.md, 2026-03-04-0506-3.md, 2026-03-04-0506.md
  2026-03-04-0506-2.md: carries=BETA-ONLY originClassAtFirstSight=untrusted bytesAtFirstSight=216
    recorded provenance: originClass=untrusted sessionId=beta-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  2026-03-04-0506-3.md: carries=GAMMA-ONLY originClassAtFirstSight=agent bytesAtFirstSight=219
    recorded provenance: originClass=agent sessionId=gamma-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  2026-03-04-0506.md: carries=ALPHA-ONLY originClassAtFirstSight=agent bytesAtFirstSight=219
    recorded provenance: originClass=agent sessionId=alpha-session recordDescribesThisFile=true recordBelongsToCaptureInThisFile=true
  ALPHA-ONLY retained copies: 1
  BETA-ONLY retained copies: 1
  GAMMA-ONLY retained copies: 1
overlapping rollovers per round: 6, rounds: 15, rounds with a provenance record attached to the wrong capture: 0, captures lost overall: 0
```
Observed result after fix: on pristine main six overlapping rollovers per round kept one artifact per round and 75 of 90 captures were gone. On the head reviewed at revision 4 every capture survived, but in fifteen of fifteen rounds at least one surviving artifact carried a provenance record belonging to a different capture, with `recordDescribesThisFile=false` and one losing session id spread across several artifacts, exactly the rollback interleaving the review described. On this head all 90 captures survive and no artifact carries another capture's record: `rounds with a provenance record attached to the wrong capture: 0, captures lost overall: 0`. The four scenario runs show the same at the artifact level: three overlapping rollovers publish three files, each with its own marker, its own session id, `recordDescribesThisFile=true`, and the origin class its own sender earned, and the non-owner capture answers `untrusted` from the first moment it can be seen, with its content already readable, `bytesAtFirstSight=216`.
What was not tested: only same-process overlap was driven. Two OpenClaw processes sharing one workspace directory are not covered here; `Root.create` reaches `O_CREAT | O_EXCL` and the claim runs inside one SQLite write transaction, so both should hold across processes on a local filesystem, but that was not driven, and network filesystems where `O_EXCL` is weak were not driven either. The full Memory Core chunk and embedding publication was not driven; the classifier call that decides origin and the hook that produces the artifact were driven for real, and the persisted chunk origin is inferred from the classification `prepareIndexEntry` carries. The LLM-slug configuration was exercised only through the shared claim path, not with a live model call. The unique-token candidates past suffix 64 were driven as a group, not to their own exhaustion. No build output was produced, so the Memory Core plugin ran from its source modules with the host SDK specifiers resolved to the in-repo sources.

https://claude.ai/code/session_0141MfnPJYgyybnqkv98rhfr

### DIFF
--- a/src/agents/memory-write-provenance.ts
+++ b/src/agents/memory-write-provenance.ts
@@ -3,6 +3,7 @@ import { isMissingPathError } from "../infra/errors.js";
 import { canonicalPathFromExistingAncestor } from "../infra/fs-safe.js";
 import { logWarn } from "../logger.js";
 import {
+  claimMemoryArtifactCreateProvenance,
   clearMemoryArtifactProvenance,
   normalizeMemoryArtifactRelativePath,
   recordMemoryArtifactWriteProvenance,
@@ -18,6 +19,14 @@ export type MemoryWriteProvenanceObserver = {
     commit: () => Promise<void>;
   }) => Promise<void>;
   clearAfterDelete: (absolutePath: string, contentBefore: string) => Promise<void>;
+};
+
+export type MemoryExclusiveCreateObserver = {
+  createExclusive: (params: {
+    absolutePath: string;
+    contentAfter: string;
+    commit: () => Promise<void>;
+  }) => Promise<"created" | "occupied">;
 };
 
 type ProvenanceWriteOperations = {
@@ -108,7 +117,7 @@ export function createMemoryWriteProvenanceObserver(params: {
   sessionId?: string;
   sessionKey?: string;
   now?: () => number;
-}): MemoryWriteProvenanceObserver {
+}): MemoryWriteProvenanceObserver & MemoryExclusiveCreateObserver {
   const now = params.now ?? Date.now;
   const resolvePath = params.resolvePath ?? canonicalPathFromExistingAncestor;
   const resolveRelativePath = async (absolutePath: string) => {
@@ -151,6 +160,43 @@ export function createMemoryWriteProvenanceObserver(params: {
         }
         throw error;
       }
+    },
+    createExclusive: async ({ absolutePath, contentAfter, commit }) => {
+      const relativePath = await resolveRelativePath(absolutePath);
+      if (!relativePath) {
+        await commit();
+        return "created";
+      }
+      const claim = await claimMemoryArtifactCreateProvenance({
+        workspaceDir: params.workspaceDir,
+        relativePath,
+        contentAfter,
+        originClass: params.resolveOriginClass(),
+        observedAt: now(),
+        sessionId: params.sessionId,
+        sessionKey: params.sessionKey,
+      });
+      if (claim.status === "occupied") {
+        return "occupied";
+      }
+      if (claim.status === "unclassified") {
+        await commit();
+        return "created";
+      }
+      try {
+        await commit();
+      } catch (error) {
+        try {
+          await claim.rollback();
+        } catch (rollbackError) {
+          throw new Error(
+            `File create failed and memory provenance rollback also failed: ${String(error)}`,
+            { cause: rollbackError },
+          );
+        }
+        throw error;
+      }
+      return "created";
     },
     clearAfterDelete: async (absolutePath, contentBefore) => {
       const relativePath = await resolveRelativePath(absolutePath);
--- a/src/hooks/bundled/session-memory/handler-auto-reset.test.ts
+++ b/src/hooks/bundled/session-memory/handler-auto-reset.test.ts
@@ -1,12 +1,86 @@
+import { createHash } from "node:crypto";
 import fs from "node:fs/promises";
 import path from "node:path";
 import { expectDefined } from "@openclaw/normalization-core";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { useAutoCleanupTempDirTracker } from "../../../../test/helpers/temp-dir.js";
 import type { OpenClawConfig } from "../../../config/config.js";
 import { replaceTranscriptEvents } from "../../../config/sessions/session-accessor.js";
+import { listMemoryArtifactProvenance } from "../../../memory/memory-artifact-provenance.js";
+import { resetPluginStateStoreForTests } from "../../../plugin-state/plugin-state-store.js";
+import { withStateDirEnv } from "../../../test-helpers/state-dir-env.js";
 import { createInternalHookEvent } from "../../internal-hooks.js";
 import handler, { flushSessionMemoryWritesForTest } from "./handler.js";
+
+type ProvenanceRecordObservation = {
+  relativePath: string;
+  originClass: "agent" | "untrusted";
+  destinationExists: boolean;
+  markdownFilesPresent: string[];
+};
+
+const provenanceObservations = vi.hoisted(() => ({
+  records: [] as ProvenanceRecordObservation[],
+}));
+
+const claimBarrier = vi.hoisted(() => ({
+  contenders: 0,
+  relativePath: undefined as string | undefined,
+  arrived: 0,
+  pending: undefined as Promise<void> | undefined,
+  release: undefined as (() => void) | undefined,
+}));
+
+const awaitClaimBarrier = vi.hoisted(() => async (relativePath: string): Promise<void> => {
+  if (claimBarrier.contenders <= 0) {
+    return;
+  }
+  claimBarrier.relativePath ??= relativePath;
+  if (claimBarrier.relativePath !== relativePath) {
+    return;
+  }
+  claimBarrier.pending ??= new Promise<void>((resolve) => {
+    claimBarrier.release = resolve;
+  });
+  claimBarrier.arrived += 1;
+  if (claimBarrier.arrived >= claimBarrier.contenders) {
+    claimBarrier.release?.();
+  }
+  await claimBarrier.pending;
+});
+
+vi.mock("../../../memory/memory-artifact-provenance.js", async (importOriginal) => {
+  const actual =
+    await importOriginal<typeof import("../../../memory/memory-artifact-provenance.js")>();
+  const nodeFs = await import("node:fs/promises");
+  const nodePath = await import("node:path");
+  return {
+    ...actual,
+    claimMemoryArtifactCreateProvenance: async (
+      params: Parameters<typeof actual.claimMemoryArtifactCreateProvenance>[0],
+    ) => {
+      const destination = nodePath.default.join(params.workspaceDir, params.relativePath);
+      await awaitClaimBarrier(params.relativePath);
+      const claim = await actual.claimMemoryArtifactCreateProvenance(params);
+      if (claim.status !== "claimed") {
+        return claim;
+      }
+      provenanceObservations.records.push({
+        relativePath: params.relativePath,
+        originClass: params.originClass,
+        destinationExists: await nodeFs.default
+          .lstat(destination)
+          .then(() => true)
+          .catch(() => false),
+        markdownFilesPresent: await nodeFs.default
+          .readdir(nodePath.default.dirname(destination))
+          .then((entries) => entries.filter((entry) => entry.endsWith(".md")).sort())
+          .catch(() => []),
+      });
+      return claim;
+    },
+  };
+});
 
 const tempDirs = useAutoCleanupTempDirTracker(afterEach);
 
@@ -15,11 +89,104 @@ describe("session-memory automatic reset", () => {
 
   beforeEach(() => {
     tempDir = tempDirs.make("openclaw-session-memory-auto-");
+    provenanceObservations.records.length = 0;
+    claimBarrier.contenders = 0;
+    claimBarrier.relativePath = undefined;
+    claimBarrier.arrived = 0;
+    claimBarrier.pending = undefined;
+    claimBarrier.release = undefined;
   });
 
   afterEach(async () => {
     await flushSessionMemoryWritesForTest();
+    resetPluginStateStoreForTests();
   });
+
+  const rolloverAt = new Date("2026-03-04T05:06:07.000Z");
+
+  function makeConfig(storePath: string): OpenClawConfig {
+    return {
+      agents: { defaults: { workspace: tempDir } },
+      session: { store: storePath },
+    } satisfies OpenClawConfig;
+  }
+
+  async function storeRolloverTranscript(params: {
+    storePath: string;
+    sessionId: string;
+    sessionKey: string;
+    marker: string;
+    senderIsOwner?: boolean;
+  }): Promise<void> {
+    await replaceTranscriptEvents(
+      {
+        agentId: "main",
+        sessionId: params.sessionId,
+        sessionKey: params.sessionKey,
+        storePath: params.storePath,
+      },
+      [
+        {
+          type: "message",
+          id: `${params.sessionId}-user`,
+          parentId: null,
+          message: {
+            role: "user",
+            content: `Remember ${params.marker}`,
+            __openclaw: { senderIsOwner: params.senderIsOwner ?? true },
+          },
+        },
+        {
+          type: "message",
+          id: `${params.sessionId}-assistant`,
+          parentId: `${params.sessionId}-user`,
+          message: { role: "assistant", content: "Captured automatically" },
+        },
+      ],
+    );
+  }
+
+  function buildRolloverEvent(params: {
+    cfg: OpenClawConfig;
+    storePath: string;
+    sessionId: string;
+    sessionKey: string;
+  }) {
+    return {
+      ...createInternalHookEvent("session", "auto-reset", params.sessionKey, {
+        cfg: params.cfg,
+        agentId: "main",
+        workspaceDir: tempDir,
+        storePath: params.storePath,
+        sessionEntry: { sessionId: params.sessionId },
+        reason: "daily",
+      }),
+      timestamp: rolloverAt,
+    };
+  }
+
+  async function runRollover(params: {
+    storePath: string;
+    sessionId: string;
+    sessionKey: string;
+    marker: string;
+    senderIsOwner?: boolean;
+  }): Promise<void> {
+    await storeRolloverTranscript(params);
+    await handler(
+      buildRolloverEvent({
+        cfg: makeConfig(params.storePath),
+        storePath: params.storePath,
+        sessionId: params.sessionId,
+        sessionKey: params.sessionKey,
+      }),
+    );
+  }
+
+  async function resolveCapturedBasename(memoryDir: string): Promise<string> {
+    const files = await fs.readdir(memoryDir);
+    return expectDefined(files[0], "files[0] test invariant").replace(/\.md$/u, "");
+  }
 
   it.each(["daily", "idle"] as const)(
     "creates memory from the ended session on %s reset",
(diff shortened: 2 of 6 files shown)


yetval:
Rebased onto current main and addressed the three findings. New head `74b499d47d0d8fcc62d617ec819a86ee38d6cc54`, proof in the body re-pinned to it and to base main `c38897f5544081974bc13632c2f1a17337b52ba7`.

**Preserve existing provenance across competing failed claims (P2)** - Fixed by removing the path rather than repairing rollback. The handler now claims a filename with an exclusive `Root.create` of an empty file before anything else, and only then reserves provenance and writes the entry into the name it already owns. A claimant that loses a candidate performs no provenance operation at all, so competing failed claims cannot reserve against, downgrade, or misattribute another session's artifact under any interleaving, and rollback ordering stops mattering. The ordering cost is named in the body: the claimed file is a zero-byte artifact between the claim and the reservation, which carries no content that could read as trusted, and the claim is released with `memoryRoot.remove` if the provenance-wrapped write fails.

**Preserve captures after the first 64 filenames are occupied (P2)** - Fixed. Numbered suffixes run to 64 and are then followed by unique-token candidates, so an exhausted date and slug keeps its capture instead of throwing it away. Scenario B in the body drives a workspace with 64 occupied names: the capture lands on both main and this patch, and was discarded on the previously reviewed head.

**Skip occupied candidates that fail filesystem type checks (P2)** - Fixed, and the codes were confirmed against the pinned fs-safe 0.8.5 rather than assumed. `Root.create` reports an occupied directory as `not-file` and an occupied hard link or symlink as `path-alias`. A candidate is treated as occupied when the error carries `already-exists`, `not-file`, or `path-alias` and an `lstat` of the candidate path confirms something is actually there; a path-security failure with nothing at that path still propagates, so an aliased `memoryDir` is not silently retried. Scenario C drives a directory and a hard link on the first two candidates and lands on `-4.md`, identical to main.

**Merge risk, existing workspaces (P1)** - Answered with runtime evidence. Scenarios B and C are now unchanged against main in the before/after block, so occupied workspaces with exhausted or non-regular candidate names behave exactly as the released implementation does. Scenario D shows the occupied artifact's provenance record byte-identical before and after two competing failed claims.

**Merge risk, trace coverage (P1)** - The refreshed proof covers four scenarios instead of one, three of them on populated workspaces: concurrent rollover, 64 occupied names, occupied directory plus hard link, and provenance under competing claims. Scenario D also shows a populated workspace going from 2 surviving files on main to 3 here, so the original defect is demonstrated outside a fresh workspace.

On the sibling suite: `handler.test.ts` is untouched by this diff, and its sequential case from #77749 describes the numbering this change preserves. Scenario C reproduces that same `-2` then `-4` numbering on the real runtime.

Regression coverage is three new cases in `handler-auto-reset.test.ts`. Two of them fail on the previously reviewed head `ae719b0341eb` on the exact defects above. The provenance case passes on that head, because the corruption there needs a specific interleaving of two failed reservations; it is kept because it fails on pristine main, and the structural change is what removes the race rather than a timing-dependent case.

yetval:
Rebased onto `bbc599d39212e0d4b9874b84a0b331ca3a45a370` and pushed `428274b9b8577cfa200fd5f704c3b8fe558d4331`. Still a draft. Item by item against revision 3.

**[P1] Keep untracked claims out of indexed Markdown paths - fixed** in `src/hooks/bundled/session-memory/handler.ts`. `claimMemoryFilename` no longer creates a zero-byte `.md` and reserves provenance afterwards. It now skips candidates that already exist, then reserves provenance for a candidate that does not exist yet and commits with `memoryRoot.create(candidate, entry)`, so the exclusive `O_CREAT | O_EXCL` open publishes the name with its content already in it. The earliest moment any reader can resolve the path, the origin is already recorded, so there is no ordering in which `resolveMemoryPathClassification` can fall back to `originClass: agent` for a provenance-free artifact and no ordering in which `prepareIndexEntry` can carry that stale class into `writeChunks`. The exclusive create is still the only authority on name ownership, so the collision protection is unchanged, and no competing per-candidate provenance reservation was restored: a losing claimant rolls its own reservation back under its `reservationId`, which restores the previous record or deletes only its own.

**Security: untracked placeholder permits stale trusted classification - resolved** by the same change. The workspace lock is still not taken and is not needed, because the reader no longer has a window in which a memory Markdown path exists without a recorded origin.

**Sibling provenance cases at `handler.test.ts:428` - verified, not edited.** All three table-driven cases pass unchanged; their absence-before-reservation assertion is exactly the contract the previous head broke and the contract the handler now honors again. `node scripts/run-vitest.mjs src/hooks/bundled/session-memory/handler.test.ts src/hooks/bundled/session-memory/handler-auto-reset.test.ts` reports `Tests 35 passed (35)` on the rebased tree.

**Add real behavior proof - refreshed** in the body and pinned to the new head. The trace drives the real hook handler interleaved in one process with the real Memory Core classifier `resolveMemoryPathClassification`, called in the order `prepareIndexEntry` calls it, on pristine main, on the reviewed head `74b499d47d0`, and on this patch with identical inputs and a pinned rollover timestamp. It covers owner and non-owner captures alone and concurrently in one workspace. On the reviewed head the non-owner artifact was first visible at zero bytes and classified `agent`; on this patch it is first visible with its full content and classified `untrusted`, while the owner artifact stays `agent` and readable. Pristine main keeps only one artifact and loses the owner capture entirely.

**Merge risk (P1): concurrent indexer persists untrusted content as trusted - addressed and exercised.** That interleaving is the concurrent scenario in the refreshed trace, and three new cases in `handler-auto-reset.test.ts` assert `destinationExists` is false at provenance-record time for both an owner and a non-owner capture. Those three cases fail on `74b499d47d0` and pass here.

Two things remain maintainer calls rather than author-actionable items.

- Memory Core still classifies before it reads and does not revalidate origin at publication. This PR removes the window the hook was creating, but it does not change that ordering in `manager-embedding-ops.ts`. Any other producer that creates a workspace memory Markdown path before recording provenance reopens the same hazard, and the untracked default of `originClass: agent` in `memory-path-provenance.ts:63` is what makes it exploitable. Whether to revalidate provenance at `writeChunks`, or to default untracked workspace memory closed, is a Memory Core ownership decision outside this handler.
- Under a genuine lost race the winner's record can be the conservative `untrusted` rather than `agent`, because `recordMemoryArtifactWriteProvenance` downgrades when it finds a previous record whose hash does not match `contentBefore`. That is fail-closed and matches main's existing semantics for overlapping writes to one path, so it was left alone rather than widened into the provenance store.

yetval:
Rebased onto `2f97cdbacace` and pushed as `c0c0a0a9fdbd`. Body and proof are refreshed against that head. Item by item against the revision 4 verdict.

**[P2] Prevent failed claims from restoring another loser's provenance (`handler.ts:128-133`).** Fixed, and the finding was correct. The problem was not the rollback guard but the entry point: `MemoryWriteProvenanceObserver.write` is built for overwriting an artifact the caller already owns, so it reserves unconditionally, keeps the record it displaced and restores it on failure. Applied to an absent path that made several captures simultaneous reservation holders for one name, and `reservationId` only proves the rollback owns the current record, not that the predecessor it restores is still valid. The reservation is now the point of mutual exclusion. `claimMemoryArtifactCreateProvenance` in `src/memory/memory-artifact-provenance.ts` writes the record for a candidate path inside the store's write transaction and only when that path has no record, `src/agents/memory-write-provenance.ts` exposes it as `createExclusive`, and the hook advances to the next candidate when the claim comes back occupied. A capture that loses the reservation never reaches the filesystem, and a capture that wins the reservation but loses the exclusive create deletes only its own record, because by construction there is no predecessor to restore. `recordMemoryArtifactWriteProvenance`, its rollback semantics and every existing caller are unchanged.

**Merge risk (P1): failed concurrent claims can leave a surviving capture associated with another session's hash and identity.** Resolved by the same change and now driven, not argued. Six overlapping rollovers per round, fifteen rounds, on the reviewed head `428274b9b857`: fifteen of fifteen rounds ended with at least one artifact carrying a provenance record belonging to a different capture, `recordDescribesThisFile=false`, with one losing session id spread across several artifacts. On `c0c0a0a9fdbd` with identical inputs: `rounds with a provenance record attached to the wrong capture: 0, captures lost overall: 0`. Pristine main lost 75 of those 90 captures outright.

**Merge risk (P1): the runtime evidence does not exercise competing rollback ordering.** It does now. The proof block has the three-tree run at the artifact level plus the sixfold-overlap sweep described above. Memory Core chunk and embedding publication is still not driven, and the body still says so; the classifier call that decides origin and the hook that produces the artifact are both driven for real.

**Next step (P2): demonstrate that three overlapping captures retain their own provenance without exposing untracked Markdown.** `src/hooks/bundled/session-memory/handler-auto-reset.test.ts` gains a case that holds three captures at the claim for one candidate until all three have arrived, then asserts each published artifact carries its own session id, its own content hash and the origin class its own sender earned, and that every reservation was taken while the destination did not exist. `src/memory/memory-artifact-provenance.test.ts` gains a case for the primitive itself: one claim per path, later claims occupied, and a rollback that touches only its own record.

**Regression coverage misses competing reservations (evidence row).** Correct about the old occupied-artifact case, which finishes the original capture first and so skips the original path without reserving. That case stays as it is, since it covers the populated-workspace path, and the new three-contender case covers the competing-reservation path it could not reach.

On the sibling suite: `src/hooks/bundled/session-memory/handler.test.ts` changes in one place. Its module mock for `src/memory/memory-artifact-provenance.js` now also exports the claim function the handler calls, and the two provenance cases observe that function instead of `recordMemoryArtifactWriteProvenance`. The assertions are unchanged, including that the destination is absent when provenance is recorded and that no file is written when provenance recording fails.

Still a maintainer call, unchanged from the last cycle: Memory Core resolves an artifact's origin class before it reads the content and never revalidates it at publication (`manager-embedding-ops.ts`, with the untracked default at `memory-path-provenance.ts:63`). Nothing in this hook can close that.

Local gates: `handler-auto-reset.test.ts` 10 passed, `handler.test.ts` 26 passed, `memory-artifact-provenance.test.ts` 14 passed, `memory-write-provenance.test.ts` 7 passed, and `oxfmt` on all six changed files. `tsgo` and `oxlint` were not run on this host, so a type error in the new code would surface only in CI.

@clawsweeper re-review

PR #144147, opened 2026-09-10
fix(channels): resolve DM allowlist wildcards after channel normalization

yetval:
## Summary
An operator who writes the provider-prefixed wildcard in a channel DM allowlist (`feishu:*`, `lark:*` or `user:*` on Feishu, `mattermost:*` or `user:*` on Mattermost) gets a channel that is fully open at admission but that every literal-`"*"` consumer reads as a one-entry allowlist. Two things break for that operator.

Routing. `resolvePinnedMainDmOwnerFromAllowlist` returns `"*"` as the pinned main-DM owner. The channel then installs `mainDmOwnerPin: { ownerRecipient: "*", senderRecipient: <sender> }` and `shouldSkipPinnedMainDmRouteUpdate` skips whenever owner and sender differ. No real sender id equals `"*"`, so the main DM session never records a delivery target, and a session that already had one keeps the stale target forever. Anything that resolves `channel: "last"` for that session (cron `announce` jobs, heartbeat and proactive messages, `openclaw message` to the last route) is silently undeliverable or goes to the wrong peer.

Security audit. `resolveDmAllowAuditState` reports `hasWildcard: false` and lists a bogus admitted sender literally named `*`, so `openclaw security audit` presents a wide-open DM channel as an ordinary narrow allowlist: no enumeration of bound DM peers, no evaluation of the "anyone" sender, and no shared-session exposure finding.

Admission itself was already correct on main. What was wrong is every downstream consumer that decided "is this a wildcard?" on the raw config string instead of asking the channel.

## Root cause
The channel adapter decides what a wildcard is through its own normalizer, but three consumers tested the raw entry for a literal `"*"`.

`src/security/dm-policy-shared.ts:22` runs its literal guard before `normalizeEntry`. `normalizeFeishuAllowEntry` maps `feishu:*`, `lark:*` and `user:*` to `"*"` (`extensions/feishu/src/policy.ts:47-62`), and the Feishu ingress descriptor uses that same function for `isWildcardEntry` (`policy.ts:44`), so admission really is open to everyone. The guard misses, `normalizedOwners` becomes `["*"]` with length 1, and the helper returns `"*"` as a pinned owner. `normalizeMattermostAllowEntry` has the same shape (`extensions/mattermost/src/mattermost/ingress-identity.ts:29-45`).

`src/channels/message-access/dm-allow-state.ts:17` derived `hasWildcard` from the raw config. `mattermost:*` survived the `!== "*"` filter, normalized to `"*"`, and landed in `admittedPrincipals` while `hasWildcard` stayed `false`, which skipped the wildcard path in `src/security/audit-channel.ts:398-412` so the audit never called `listExactDirectMessageBindingPeerIds` and never evaluated the "anyone" sender.

`src/channels/message-access/state.ts:427` repeated the literal test for `NormalizedIngressAllowlist.hasWildcard`, which feeds `participantOutcomeAffecting` (`src/channels/message-access/runtime.ts:723-729`) and the `dmPolicy: "open"` branch in `src/channels/message-access/sender-gates.ts:93`.

## Fix
Ask the normalizer first, then test for the wildcard.

- `src/security/dm-policy-shared.ts`: test `normalizedOwners.includes("*")` after normalization, so a normalizer-resolved wildcard stays unpinned.
- `src/channels/message-access/dm-allow-state.ts`: normalize first, derive `hasWildcard` from the normalized list, and drop `"*"` from `admittedPrincipals`.
- `src/channels/message-access/state.ts`: read the wildcard flag off the entry the adapter already stamped through the descriptor's `isWildcardEntry` (`src/channels/message-access/runtime-identity.ts:162-178`), keeping the literal test as an additive fallback for adapters that do not stamp it, and record the raw literal `"*"` declaration separately as `hasLiteralWildcard`.
- `src/channels/message-access/allowlist.ts`: recompute `hasWildcard` inside `applyIdentifierAuthenticationPolicy`, so a normalized wildcard only keeps admitting while at least one wildcard entry survives the authentication result. A literal `"*"` keeps admitting without an identifier match, which is the shipped meaning of an open allowlist.
- `src/security/audit-channel.ts`: require `!hasWildcard` in the locked-DM predicate, and take the `dm.open_invalid` validity warning from `evaluateDmPolicyAllowFromDependency` instead of from the audit wildcard state.

The raw-`"*"` short-circuit in `dm-policy-shared.ts` stays, because Slack (`normalizeSlackAllowOwnerEntry`) and Signal (`normalizeSignalAllowRecipient`) return `undefined` for a wildcard: normalizing first would drop the entry entirely and a bare `["*"]` would then look like an empty list. In `dm-allow-state.ts` a literal `"*"` is likewise never handed to the normalizer, so channels that reject it keep behaving exactly as before.

### Review round 2: the two findings from the previous verdict
Both were real and both are fixed in this revision.

[P1 and the security item] Preserve authentication rejection when enabling normalized wildcards. Correct: on the previous head, `hasWildcard` was set before authentication filtering and then passed through `applyIdentifierAuthenticationPolicy` untouched, so `senderGateForDirect` took its open-policy shortcut even when every matched entry had been rejected for an insufficient identifier claim. With `dmPolicy: "open"`, `allowFrom: ["mattermost:*"]` and `minIdentifierAuthentication: "verified"`, an unverified sender reached dispatch. The fix carries the raw literal declaration as `hasLiteralWildcard` and recomputes `hasWildcard` from the surviving wildcard entries inside the policy function, which is where rejection is resolved. A literal `"*"` is unchanged, because it admits without an identifier match and has no claim to compare. The proof below shows both an accepted and a rejected sender claim at the ingress boundary, and the new cases in `message-access.test.ts` cover the accepted claim, the rejected claim and the preserved literal wildcard.

[P2] Keep open-DM validity warnings aligned with the configuration schema. Correct: moving the flag also suppressed `channels.<id>.dm.open_invalid` for `dmPolicy: "open"` with only `mattermost:*`, while the Mattermost schema still calls `requireOpenAllowFrom` and `evaluateDmPolicyAllowFromDependency` still requires a literal trimmed `"*"` (`src/config/zod-schema.core.ts:714-744`). The audit now asks that canonical dependency check for the validity warning, so the audit and the configuration validator cannot drift, and the normalized wildcard state stays in charge of exposure and routing analysis only. The proof shows `dm.open_invalid` present both before and after this patch for `mattermost:*`, alongside the new shared-session exposure finding.

### The locked-DM guard, and what it fixes beyond this branch
Once a normalized wildcard is no longer an admitted sender, `admittedPrincipals` is empty for a wildcard-only allowlist. The consumer at `src/security/audit-channel.ts:286` treated an empty list under `allowlist` or `pairing` as "DMs are locked; unknown senders are blocked or receive a pairing code". Under those policies admission does not go through `hasWildcard` at all: it goes through `match.matched`, and `createIdentityAdapter` collapses a wildcard entry to `{ value: "*", wildcard: true }` which `matchSubject` pairs with any sender (`runtime-identity.ts:166-176`, `:211-237`). A wildcard-only allowlist therefore admits everyone, and reporting it as locked is false reassurance. This was already true on main for a bare `"*"`. `allowFrom: []` still reports locked.

`resolvePinnedMainDmOwnerFromAllowlist` is the single owner of this policy for roughly twelve channels (Feishu, Mattermost, Slack, Signal, Discord, Telegram, WhatsApp, iMessage, Matrix, LINE, plus `src/infra/event-session-routing.ts`). Fixing it per channel would re-duplicate the policy in every caller, which is exactly what the shared helper exists to prevent. The same reasoning applies to `resolveDmAllowAuditState` and `resolveIngressAllowlist`: they own the audit view and the normalized ingress state.

Sibling surfaces. The authentication repair lives in `applyIdentifierAuthenticationPolicy` and `resolveIngressAllowlist`, the shared ingress kernel every channel adapter runs through, so Feishu, Mattermost, Slack, Signal, Discord, msteams, IRC, Google Chat and the rest are fixed together rather than one at a time. No per-channel wildcard admission code exists to fix separately.

Slack and Signal are proven unaffected by the routing half. Both normalizers return `undefined` for a wildcard, so `normalizedOwners` never contains `"*"` and the raw short-circuit remains the only path that fires. Feishu registers only `collectWarnings` and `collectAuditFindings` (`extensions/feishu/src/channel.ts:2002`) and no `resolveDmPolicy`, so `resolveDmAllowAuditState` is never reached for Feishu through the registered adapter: Feishu is affected by the routing half only. Mattermost does reach it, because `createRestrictSendersChannelSecurity` passes `normalizeDmEntry` through as `normalizeEntry` (`extensions/mattermost/src/channel.ts:141`, `src/plugin-sdk/channel-policy.ts:433`), which is why the proof is driven through the registered Mattermost channel plugin.

## Owner decisions still open
Neither is author-actionable, and neither is worked around in this diff.

1. Existing prefixed-wildcard installations will resume updating the shared main session's delivery target. The proof block shows that directly: a main session seeded with `channel:dm-operator` keeps that recipient on current main and moves to the newest inbound DM peer with this patch. That is the intended repair, but it changes the recipient of subsequent last-route sends on upgrade and needs owner acceptance.
2. Sponsorship for the restricted security path. This touches the shared DM policy and ingress authentication surface, so it needs a security owner to sponsor it.

## Scope note
A wildcard-only allowlist under `dmPolicy: "allowlist"` still emits no DM finding of its own after this patch, exactly as before it. Adding a positive "this allowlist admits everyone" finding is a real gap but a separate policy decision about audit vocabulary, so it is deliberately not in this diff.

## Verification
Rebased onto `e14b3ddac2e5ea9dbb450057b7436c22a012a89b`. Scoped runs with `node scripts/run-vitest.mjs`:

- `src/channels/message-access/message-access.test.ts`, `src/security/audit-channel-dm-policy.test.ts`, `src/channels/message-access/dm-allow-state.test.ts`, `src/security/dm-policy-shared.test.ts`: 4 files, 61 tests passed.
- `src/channels/message-access/identifier-authentication.test.ts`, `src/channels/message-access/runtime-identity.test.ts`, `src/security/audit-channel-entry-authentication.test.ts`: passed, so the existing authentication contract is unchanged.
- `extensions/feishu/src/policy.test.ts`, `extensions/mattermost/src/mattermost/ingress-identity.test.ts`, `extensions/slack/src/monitor/allow-list.test.ts`: passed.
- `src/channels/message-access/admission-evidence.test.ts`, `src/channels/message-access/context-builder-callers.test.ts`, `src/channels/message-access/participant-input.test.ts`, `src/plugins/registry-runtime.channel-ingress.test.ts`, `extensions/mattermost/src/mattermost/monitor.authz.test.ts`: passed.
- `oxfmt --check` clean on all six changed files.
- Reverting only the production files and rerunning the two edited suites fails the new wildcard assertions and keeps the preserved-behavior assertions green, so the new coverage fails on the original defect.
- Typechecking is delegated to GitHub checks on this host.

## Real behavior proof
Behavior addressed: A prefixed DM allowlist wildcard (`feishu:*`, `lark:*`, `user:*`, `mattermost:*`) admits every sender but is resolved as a single pinned main-DM owner named `*`, which freezes the main DM session delivery target, and the registered channel security audit misreports the same allowlist as narrow; while normalized wildcard admission must still respect `minIdentifierAuthentication`, so a below-threshold sender is refused before the turn handoff.
Real environment tested: the real registered Mattermost channel plugin `mattermostPlugin` driven through the real `collectChannelSecurityFindingsCore`, the same collector `openclaw security audit` uses; the real shared ingress resolver `resolveStableChannelMessageIngress` driven with the real `mattermostIngressIdentity` descriptor and the real Mattermost drop-log formatter `formatMattermostDirectMessageDropLog`; and the real main-DM route recorder `recordInboundSession` writing to a real on-disk session store resolved by `resolveSessionStorePathCore`, read back through `loadSessionEntryReadOnly`, with the pin built by the real `resolvePinnedMainDmOwnerFromAllowlist` and `normalizeMattermostAllowEntry` exactly as `extensions/mattermost/src/mattermost/monitor-turn.ts:432-450` builds it. Nothing on these paths was stubbed and no network boundary was replaced; the only hand-built value is the inbound message context record. Run on pristine `main` e14b3ddac2e5ea9dbb450057b7436c22a012a89b with the six production files at their base contents, and on head 1bf20e4e402db9998348716c47dc5380959268f1, with identical inputs and a wiped state directory for each run.
Exact steps or command run after this patch: `OPENCLAW_STATE_DIR=<scratch dir> HOME=<scratch dir> ./node_modules/.bin/tsx ./proof-144147.mts` from the worktree root, driving the registered Mattermost security audit for `allowFrom: ["mattermost:*"]` under `dmPolicy: "open"` and `dmPolicy: "allowlist"`, then the main-DM delivery route for a fresh session and for a session already holding another recipient, then inbound DM admission for a server-authenticated sender and for a below-threshold sender under `minIdentifierAuthentication: "verified"`.
Evidence after fix:
```
# BEFORE (pristine main e14b3ddac2e5ea9dbb450057b7436c22a012a89b)
# registered mattermost audit, dmPolicy=open allowFrom=["mattermost:*"]
  critical channels.mattermost.dm.open
  warn     channels.mattermost.dm.open_invalid
# registered mattermost audit, dmPolicy=allowlist allowFrom=["mattermost:*"]
  (no dm or routing findings)
# main-dm delivery route, allowFrom=["mattermost:*"] dmScope=main
  pinned main-dm owner = "*"
  mattermost: skip main-session last route for user-alpha (pinned owner *)
  fresh session, target after inbound dm = (none recorded)
  seeded session, target before inbound dm = mattermost channel:dm-operator
  mattermost: skip main-session last route for user-alpha (pinned owner *)
  seeded session, target after inbound dm = mattermost channel:dm-operator
# inbound dm admission, dmPolicy=open allowFrom=["mattermost:*"] minIdentifierAuthentication=verified
  server-authenticated sender verdict=allow reasonCode=dm_policy_allowlisted gate=allow thresholdRemovedMatch=false
  turn handoff reached = yes
  below-threshold sender      verdict=block reasonCode=dm_policy_not_allowlisted gate=block-dispatch thresholdRemovedMatch=true
  mattermost: drop dm sender=user-alpha (dmPolicy=open reason=dm_policy_not_allowlisted hint=add-allowFrom-wildcard)
  turn handoff reached = no

# AFTER (this patch, identical inputs, head 1bf20e4e402db9998348716c47dc5380959268f1)
# registered mattermost audit, dmPolicy=open allowFrom=["mattermost:*"]
  critical channels.mattermost.dm.open
  warn     channels.mattermost.dm.open_invalid
  warn     channels.dm.session_collision.mattermost-default.1
# registered mattermost audit, dmPolicy=allowlist allowFrom=["mattermost:*"]
  warn     channels.dm.session_collision.mattermost-default.1
# main-dm delivery route, allowFrom=["mattermost:*"] dmScope=main
  pinned main-dm owner = null
  fresh session, target after inbound dm = mattermost channel:dm-alpha
  seeded session, target before inbound dm = mattermost channel:dm-operator
  seeded session, target after inbound dm = mattermost channel:dm-alpha
# inbound dm admission, dmPolicy=open allowFrom=["mattermost:*"] minIdentifierAuthentication=verified
  server-authenticated sender verdict=allow reasonCode=dm_policy_open gate=allow thresholdRemovedMatch=false
  turn handoff reached = yes
  below-threshold sender      verdict=block reasonCode=dm_policy_not_allowlisted gate=block-dispatch thresholdRemovedMatch=true
  mattermost: drop dm sender=user-alpha (dmPolicy=open reason=dm_policy_not_allowlisted hint=add-allowFrom-wildcard)
  turn handoff reached = no
```
Observed result after fix: On pristine main the registered Mattermost audit reports no shared-session exposure for `mattermost:*` and the main-DM route resolver pins owner `"*"`, so the recorder skips every update and the main session holds no recipient at all on a fresh session and a stale `channel:dm-operator` on a seeded one; with the patch the owner resolves to `null`, both sessions record the real inbound peer `channel:dm-alpha`, the audit adds the shared-session exposure warning while the schema validity warning `dm.open_invalid` is still emitted for the prefixed wildcard, and at the ingress boundary the server-authenticated sender is admitted and reaches the turn handoff while the below-threshold sender is refused with `gate=block-dispatch` and the channel's own drop line, never reaching it.
What was not tested: no live Mattermost server, WebSocket session or HTTP call was made, so the transport layer is exercised only through the plugin's resolved account config; no Feishu, Slack, Signal or Discord runtime was driven, and those channels rest on the shared helper coverage plus their own normalizer suites; the full repository build and typecheck were not run on this host.

### DIFF
--- a/extensions/feishu/src/policy.test.ts
+++ b/extensions/feishu/src/policy.test.ts
@@ -1,9 +1,12 @@
 // Feishu tests cover policy plugin behavior.
 import type { OpenClawConfig } from "openclaw/plugin-sdk/core";
+import { resolvePinnedMainDmOwnerFromAllowlist } from "openclaw/plugin-sdk/security-runtime";
 import { describe, expect, it } from "vitest";
 import { FeishuConfigSchema } from "./config-schema.js";
 import {
   hasExplicitFeishuGroupConfig,
+  normalizeFeishuAllowEntry,
+  resolveFeishuDmIngressAccess,
   resolveFeishuGroupConfig,
   resolveFeishuGroupSenderActivationIngressAccess,
   resolveFeishuGroupToolPolicy,
@@ -274,5 +277,41 @@ describe("resolveFeishuGroupSenderActivationIngressAccess", () => {
         senderUserId: "on_user_123",
       }),
     ).resolves.toBe("allow");
+  });
+});
+
+describe("feishu pinned main-DM owner", () => {
+  async function dmAdmission(allowFrom: string[]): Promise<string> {
+    return (
+      await resolveFeishuDmIngressAccess({
+        cfg: createCfg({}),
+        accountId: "default",
+        dmPolicy: "allowlist",
+        allowFrom,
+        senderOpenId: "ou_stranger_0001",
+        conversationId: "oc_chat_0001",
+        mayPair: false,
+      })
+    ).ingress.admission;
+  }
+
+  function pinnedOwner(allowFrom: string[]): string | null {
+    return resolvePinnedMainDmOwnerFromAllowlist({
+      dmScope: "main",
+      allowFrom,
+      normalizeEntry: normalizeFeishuAllowEntry,
+    });
+  }
+
+  it.each(["*", "feishu:*", "lark:*", "user:*"])(
+    "admits any sender for %s and leaves the main DM owner unpinned",
+    async (entry) => {
+      await expect(dmAdmission([entry])).resolves.toBe("dispatch");
+      expect(pinnedOwner([entry])).toBeNull();
+    },
+  );
+
+  it("still pins a single configured owner", () => {
+    expect(pinnedOwner(["feishu:ou_owner_0001"])).toBe("user:ou_owner_0001");
   });
 });
--- a/extensions/mattermost/src/mattermost/ingress-identity.test.ts
+++ b/extensions/mattermost/src/mattermost/ingress-identity.test.ts
@@ -1,0 +1,22 @@
+import { resolvePinnedMainDmOwnerFromAllowlist } from "openclaw/plugin-sdk/security-runtime";
+import { describe, expect, it } from "vitest";
+import { normalizeMattermostAllowEntry } from "./ingress-identity.js";
+
+function pinnedOwner(allowFrom: string[]): string | null {
+  return resolvePinnedMainDmOwnerFromAllowlist({
+    dmScope: "main",
+    allowFrom,
+    normalizeEntry: normalizeMattermostAllowEntry,
+  });
+}
+
+describe("mattermost pinned main-DM owner", () => {
+  it.each(["*", "mattermost:*", "user:*"])("leaves the main DM owner unpinned for %s", (entry) => {
+    expect(normalizeMattermostAllowEntry(entry)).toBe("*");
+    expect(pinnedOwner([entry])).toBeNull();
+  });
+
+  it("still pins a single configured owner", () => {
+    expect(pinnedOwner(["mattermost:@Owner"])).toBe("owner");
+  });
+});
--- a/extensions/slack/src/monitor/allow-list.test.ts
+++ b/extensions/slack/src/monitor/allow-list.test.ts
@@ -1,8 +1,10 @@
 // Slack tests cover allow list plugin behavior.
+import { resolvePinnedMainDmOwnerFromAllowlist } from "openclaw/plugin-sdk/security-runtime";
 import { describe, expect, it } from "vitest";
 import {
   normalizeAllowList,
   normalizeAllowListLower,
+  normalizeSlackAllowOwnerEntry,
   normalizeSlackSlug,
   resolveSlackAllowListMatch,
   resolveSlackUserAllowListForTeam,
@@ -125,5 +127,24 @@ describe("slack/allow-list", () => {
         teamId: "T11111111",
       }),
     ).toEqual(["w01234567", "team:t11111111:user:u01234567"]);
+  });
+});
+
+describe("slack pinned main-DM owner", () => {
+  function pinnedOwner(allowFrom: string[]): string | null {
+    return resolvePinnedMainDmOwnerFromAllowlist({
+      dmScope: "main",
+      allowFrom,
+      normalizeEntry: normalizeSlackAllowOwnerEntry,
+    });
+  }
+
+  it.each(["*", "slack:*", "user:*"])("leaves the main DM owner unpinned for %s", (entry) => {
+    expect(normalizeSlackAllowOwnerEntry(entry)).toBeUndefined();
+    expect(pinnedOwner([entry])).toBeNull();
+  });
+
+  it("still pins a single configured owner", () => {
+    expect(pinnedOwner(["slack:U01234567"])).toBe("u01234567");
   });
 });
--- a/src/channels/message-access/allowlist.ts
+++ b/src/channels/message-access/allowlist.ts
@@ -93,6 +93,7 @@ function mergeResolvedAllowlists(
     matchedEntryIds,
     hasConfiguredEntries: scopedAllowlists.some((allowlist) => allowlist.hasConfiguredEntries),
     hasMatchableEntries: scopedAllowlists.some((allowlist) => allowlist.hasMatchableEntries),
+    hasLiteralWildcard: scopedAllowlists.some((allowlist) => allowlist.hasLiteralWildcard),
     hasWildcard: scopedAllowlists.some((allowlist) => allowlist.hasWildcard),
     accessGroups: {
       referenced: uniqueStrings(
@@ -169,6 +170,11 @@ export function applyIdentifierAuthenticationPolicy(
     hasMatchableEntries: allowlist.normalizedEntries.some(
       (entry) => !rejectedEntryIds.has(entry.opaqueEntryId),
     ),
+    hasWildcard:
+      allowlist.hasLiteralWildcard ||
+      allowlist.normalizedEntries.some(
+        (entry) => entry.wildcard === true && !rejectedEntryIds.has(entry.opaqueEntryId),
+      ),
     match: {
       matched: matchedEntryIds.length > 0,
       matchedEntryIds,
--- a/src/channels/message-access/dm-allow-state.test.ts
+++ b/src/channels/message-access/dm-allow-state.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from "vitest";
+import { resolveDmAllowAuditState } from "./dm-allow-state.js";
+
+function normalizeEntry(raw: string): string {
+  const trimmed = raw
+    .trim()
+    .replace(/^(demo|user):/i, "")
+    .trim();
+  return trimmed.toLowerCase();
+}
+
+async function auditState(allowFrom: string[]) {
+  return await resolveDmAllowAuditState({
+    provider: "demo",
+    accountId: "default",
+    allowFrom,
+    dmPolicy: "open",
+    normalizeEntry,
+  });
+}
+
+describe("resolveDmAllowAuditState", () => {
+  it.each(["*", "demo:*", "user:*"])("reports %s as an open DM allowlist", async (entry) => {
+    await expect(auditState([entry])).resolves.toEqual({
+      hasWildcard: true,
+      admittedPrincipals: [],
+    });
+  });
+
+  it("keeps narrow allowlists unchanged", async () => {
+    await expect(auditState(["demo:Owner-1", "owner-2"])).resolves.toEqual({
+      hasWildcard: false,
+      admittedPrincipals: ["owner-1", "owner-2"],
+    });
+  });
+
+  it("keeps configured owners alongside a prefixed wildcard", async () => {
+    await expect(auditState(["demo:*", "demo:Owner-1"])).resolves.toEqual({
+      hasWildcard: true,
+      admittedPrincipals: ["owner-1"],
+    });
+  });
+
+  it("merges persisted principals", async () => {
+    await expect(
+      resolveDmAllowAuditState({
+        provider: "demo",
+        accountId: "default",
+        allowFrom: ["demo:*"],
+        dmPolicy: "pairing",
+        normalizeEntry,
+        readStore: async () => ["demo:Paired-1"],
+      }),
+    ).resolves.toEqual({ hasWildcard: true, admittedPrincipals: ["paired-1"] });
+  });
+});
--- a/src/channels/message-access/dm-allow-state.ts
+++ b/src/channels/message-access/dm-allow-state.ts
@@ -14,7 +14,6 @@ export async function resolveDmAllowAuditState(params: {
   const configAllowFrom = normalizeStringEntries(
     Array.isArray(params.allowFrom) ? params.allowFrom : undefined,
   );
-  const hasWildcard = configAllowFrom.includes("*");
   const storeAllowFrom = await readChannelIngressStoreAllowFromForDmPolicy({
     provider: params.provider,
     accountId: params.accountId,
@@ -23,12 +22,15 @@ export async function resolveDmAllowAuditState(params: {
   });
   const normalizeEntry = params.normalizeEntry ?? ((value: string) => value);
   const normalizedCfg = normalizeStringEntries(
-    configAllowFrom.filter((value) => value !== "*").map((value) => normalizeEntry(value)),
+    configAllowFrom.map((value) => (value === "*" ? "*" : normalizeEntry(value))),
   );
+  const hasWildcard = normalizedCfg.includes("*");
   const normalizedStore = normalizeStringEntries(
     storeAllowFrom.map((value) => normalizeEntry(value)),
   );
-  const admittedPrincipals = Array.from(new Set([...normalizedCfg, ...normalizedStore]));
+  const admittedPrincipals = Array.from(
+    new Set([...normalizedCfg.filter((value) => value !== "*"), ...normalizedStore]),
+  );
   return {
     hasWildcard,
     admittedPrincipals,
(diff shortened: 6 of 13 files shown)


yetval:
Rebased onto `c38897f5544` and pushed as `62fee8b9a9d`. Body and proof are regenerated against that head. Answering the verdict item by item.

**[P2] Exclude wildcard access from the locked-DM audit finding. Fixed.** `src/security/audit-channel.ts:286` now requires `!hasWildcard` in the locked-state predicate. You were right about the mechanism and right that the filter itself should stay.

Worth recording: this finding is not only introduced by the branch. On main today, `dmPolicy: "allowlist"` with `allowFrom: ["*"]` already emits `channels.<id>.dm.locked`, because the bare wildcard was filtered out of `admittedPrincipals` there too. Under `allowlist` and `pairing`, admission never consults `hasWildcard`; it goes through `match.matched`, and `createIdentityAdapter` collapses a wildcard entry to `{ value: "*", wildcard: true }` which `matchSubject` pairs with any sender (`runtime-identity.ts:166-176`, `:211-237`). So the locked finding was already false for a bare wildcard. The guard fixes that pre-existing case and prevents the prefixed wildcard from joining it. `allowFrom: []` still reports locked.

**[medium] Public DM admission can be reported as locked. Resolved** by the same change. Covered by seven new assertions in `src/security/audit-channel-dm-policy.test.ts` across `allowlist` and `pairing`, prefixed and bare wildcards, plus the two preserved empty-allowlist cases. Reverting the four production files fails 5 of the 7 and keeps the 2 preserved ones green.

**[P1] The audit transcript exercises the helper directly, leaving the registered Feishu CLI audit claim unverified. You were right, and the claim was wrong.** Feishu registers only `collectWarnings` and `collectAuditFindings` (`extensions/feishu/src/channel.ts:2002`) and no `resolveDmPolicy`, so `resolveDmAllowAuditState` is never reached for Feishu through the registered adapter. That claim is withdrawn from the body. Mattermost is the channel that actually reaches it, exactly as your own evidence note said. The proof now drives the real `mattermostPlugin` through `collectChannelSecurityFindingsCore`, the same collector `openclaw security audit` uses, with a real account config, and prints the emitted finding ids. Feishu is affected by the routing half only.

That run also shows two false findings disappearing on the registered adapter: `dm.open_invalid` for `mattermost:*` under `dmPolicy: "open"`, and `dm.locked` for a bare `*` under `dmPolicy: "allowlist"`. The genuine `dm.open` critical and the genuine `dm.locked` for an empty allowlist are unchanged.

**[P1] Compatibility, prefixed-wildcard installs resume advancing their delivery target. Left to the maintainer.** This is the intended repair, but it does change behavior for live installs, so I have not tried to soften it in code. It is named explicitly under "Maintainer decisions this PR does not make" in the body.

**Decision needed, secops sponsorship. Still open, and correctly so.** `src/security/` is a restricted path and the sponsorship call is not mine to make.

**Scope note.** A wildcard-only allowlist under `dmPolicy: "allowlist"` still emits no DM finding of its own, exactly as before this branch. Adding a positive "this allowlist admits everyone" finding would be a genuine improvement but it is a separate decision about audit vocabulary, so it is deliberately out of this diff.

@clawsweeper re-review

yetval:
Takeover update. Rebased onto `e14b3ddac2e5ea9dbb450057b7436c22a012a89b`, both findings fixed, proof regenerated. Answering the verdict item by item.

**[P1] Preserve authentication rejection when enabling normalized wildcards** and **Security: normalized wildcard bypasses rejected sender authentication**: correct, fixed. The flag is now split. `resolveIngressAllowlist` records the raw literal `"*"` declaration as `hasLiteralWildcard`, and `applyIdentifierAuthenticationPolicy` recomputes `hasWildcard` from the wildcard entries that survived the authentication result, which is where rejection is resolved. A literal `"*"` keeps admitting without an identifier match, since it has no claim to compare; a normalized wildcard must survive the threshold. `dmPolicy: "open"` with `allowFrom: ["mattermost:*"]` and `minIdentifierAuthentication: "verified"` now refuses an unverified sender with `gate=block-dispatch` before the turn handoff, as it did before this branch, and still admits a server-authenticated one. New cases in `message-access.test.ts` cover the accepted claim, the rejected claim and the preserved literal wildcard.

**[P2] Keep open-DM validity warnings aligned with the configuration schema**: correct, fixed. `channels.<id>.dm.open_invalid` now comes from `evaluateDmPolicyAllowFromDependency`, the same canonical check `requireOpenAllowFrom` uses, so the audit and the configuration validator cannot drift. The normalized wildcard state drives exposure and routing analysis only. The proof shows `dm.open_invalid` present for `mattermost:*` both before and after the patch, alongside the new shared-session exposure finding.

**Add real behavior proof** and **merge risk: supplied transcript leaves audit behavior and delivery-route recovery unverified**: you were right that the attached transcript was the wrong artifact. It was another change's session-file output, pasted into the body by mistake. It has been replaced wholesale. The new transcript is one run of the real runtime on pristine main and on the new head with identical inputs, covering the registered Mattermost audit through `collectChannelSecurityFindingsCore`, main-DM delivery-route recovery for a fresh session and for a session already holding another recipient through the real on-disk recorder, and the ingress boundary with an accepted and a below-threshold sender.

**Merge risk: existing prefixed-wildcard installations resume changing the shared main session's delivery target**: unchanged and left to the owners. It is not worked around in the diff. The proof block shows the effect directly: a main session seeded with `channel:dm-operator` keeps that recipient on current main and moves to the newest inbound DM peer with this patch.

**Maintainer decision and security-owner sponsorship**: still open. This touches the shared DM policy and ingress authentication surface, so it needs a security owner to sponsor it. Both open items are named in the body under "Owner decisions still open".

Scoped suites re-run after the rebase: `message-access.test.ts`, `audit-channel-dm-policy.test.ts`, `dm-allow-state.test.ts`, `dm-policy-shared.test.ts` (61 tests), plus `identifier-authentication.test.ts`, `runtime-identity.test.ts`, `audit-channel-entry-authentication.test.ts`, the three extension suites, and the ingress consumer suites including `monitor.authz.test.ts`.

PR #144153, opened 2026-09-10
fix(configure): show action menu for stopped disabled gateway units

### LINKED ISSUE #83354: Helper commands can silently resurrect stopped user-level openclaw-gateway units
## Summary
OpenClaw helper commands can silently resurrect a previously stopped user-level `openclaw-gateway` systemd unit. On systems where the installer created `~/.config/systemd/user/openclaw-gateway.service` and linger is enabled, commands like `openclaw status`, `openclaw doctor`, or `openclaw configure` can cause the user service to start listening again even after `systemctl --user stop` and `disable`.

## Environment
- OpenClaw version: 2026.5.7
- Host OS: Ubuntu 24.04
- Install style: native install with user-level systemd unit

## Reproduction
```bash
# Fresh install as a user
curl -fsSL https://openclaw.dev/install.sh | bash
systemctl --user is-enabled openclaw-gateway
loginctl show-user "$USER" --property=Linger

# Stop and disable the user unit
systemctl --user stop openclaw-gateway
systemctl --user disable openclaw-gateway
ss -tlnp | grep 18789   # port appears free

# Run a helper command
openclaw status

# Gateway is back
ss -tlnp | grep 18789
```

## Expected
If an operator explicitly stops and disables the user-level gateway unit, helper commands should not silently re-enable or restart it.

## Actual
The gateway process can reappear under the user's UID after helper invocations or other triggers, which makes the port unexpectedly busy again.

## Why this matters
This is a painful operational failure mode during cutovers or handoffs. In our case it caused a multi-hour NUC migration/debug session because the wrong user-level gateway kept reclaiming the port and conflicting with the intended service.

## Likely cause
The helper path appears to do something equivalent to:

```bash
systemctl --user is-active openclaw-gateway || systemctl --user start openclaw-gateway
```

That behavior ignores operator intent unless the unit is masked.

## Suggested fix
- Do not auto-start the user gateway from helper commands by default.
- If auto-start is necessary, require an explicit flag such as `--ensure-running`.
- Respect `masked` units and document the difference between `disable` and `mask` in installer/uninstall guidance.

## Notes
Primary AoAOS doc:
- `docs/upstream-bugs/user-systemd-autoresurrect-on-helper-invocation.md`

JamieJ5926:
Fixes #83354.

## What Problem This Solves
Operator-stopped user-level gateway units get silently resurrected. MaybeInstallDaemon gated its existing-service menu on isLoaded, which resolves to enabled on systemd. A stopped and disabled unit fell through to the fresh-install path, which runs daemon-reload, enable, restart with no prompt, reclaiming the port after the operator freed it.

## Fix
When the unit is not loaded, probe hasInstalledDefinition (unit file present regardless of enablement). A present but unenrolled unit now gets the Restart/Reinstall/Skip menu instead of the auto enable+restart install path. Explicit Restart still starts the unit but leaves it disabled. Reinstall still re-enables with consent.

## Evidence
Scoped: src/commands/configure.daemon.test.ts, 13/13 pass, including 3 new tests. Failing-first verified: the 2 new stopped-unit tests fail on the unpatched source (2 failed, 11 passed) and pass with the fix (13 passed). oxfmt/oxlint clean on both touched files.

## Parked
Live systemd user-manager repro parked: no Linux host with a user manager in this environment. Proven at unit level with mocked service probes.

### DIFF
--- a/src/commands/configure.daemon.test.ts
+++ b/src/commands/configure.daemon.test.ts
@@ -15,6 +15,7 @@ const resolveGatewayInstallToken = vi.hoisted(() => vi.fn());
 const buildGatewayInstallPlan = vi.hoisted(() => vi.fn());
 const note = vi.hoisted(() => vi.fn());
 const serviceIsLoaded = vi.hoisted(() => vi.fn(async () => false));
+const serviceHasInstalledDefinition = vi.hoisted(() => vi.fn(async () => false));
 const serviceReadCommand = vi.hoisted(() => vi.fn());
 const serviceInstall = vi.hoisted(() => vi.fn(async () => {}));
 const serviceUninstall = vi.hoisted(() => vi.fn(async () => {}));
@@ -60,6 +61,7 @@ vi.mock("../daemon/service.js", async () => {
     ...actual,
     resolveGatewayService: vi.fn(() => ({
       isLoaded: serviceIsLoaded,
+      hasInstalledDefinition: serviceHasInstalledDefinition,
       readCommand: serviceReadCommand,
       install: serviceInstall,
       uninstall: serviceUninstall,
@@ -75,8 +77,8 @@ vi.mock("./systemd-linger.js", () => ({
 describe("maybeInstallDaemon", () => {
   beforeEach(() => {
     vi.clearAllMocks();
-    progressSetLabel.mockReset();
     serviceIsLoaded.mockResolvedValue(false);
+    serviceHasInstalledDefinition.mockResolvedValue(false);
     serviceReadCommand.mockResolvedValue(null);
     serviceInstall.mockResolvedValue(undefined);
     serviceUninstall.mockReset();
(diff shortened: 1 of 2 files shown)


JamieJ5926:
Thanks for the review. On the real-behavior-proof blocker: this environment has no Linux host with a systemd user manager (macOS arm64 worker), so a live stop/disable/configure cycle cannot be produced from here. What the PR does establish at source level:

- On Linux, isLoaded resolves to enabled (src/daemon/systemd-runtime.ts), so a stopped+disabled unit previously skipped the action menu.
- The fresh-install path it fell into runs daemon-reload, enable, restart unconditionally (src/daemon/systemd-install.ts activateSystemdService).
- The patched gate routes a present-but-unloaded definition to the Restart/Reinstall/Skip menu; Restart invokes systemctl restart without enable, preserving the disabled state; only explicit Reinstall re-enables.
- Unit proof is failing-first: the 2 new stopped-unit tests fail on unpatched source and pass with the fix; 13/13 scoped tests green; oxfmt/oxlint clean.

Request: a maintainer with a Linux systemd host to run the live Skip/Restart/Reinstall cycle, or a merge judgment call on the unit-level proof. Parked live repro is documented in the PR body.

PR #144156, opened 2026-09-10
fix(doctor): preserve reserved system agent ownership

### LINKED ISSUE #144125: [Bug]: doctor reports the reserved system agent dir (agents/openclaw) as an orphan
### Bug type

Incorrect behavior without a crash

### Summary

`openclaw doctor` reports the system agent's own state directory as an orphaned agent directory:

```text
- Found 1 agent directory on disk without a matching agents.list entry.
  These agents can still have sessions/auth state on disk, but config-driven routing,
  identity, and model selection will ignore them.
  Examples: openclaw
  Restore the missing agents.list entries or remove stale dirs after confirming they are
  no longer needed: ~/.openclaw/agents
```

`openclaw` is `SYSTEM_AGENT_ID` (`src/system-agent/agent-id.ts:3`). The id is reserved: `agent-create.ts:131` rejects it through `isReservedSystemAgentId`, and the system agent's own prompt says "The ids `openclaw` and `crestodian` are reserved for the system agent and cannot be created as normal agents" (`src/system-agent/assistant-prompts.ts:80`). So `agents/openclaw` can never have an `agents.list` entry, and the first remediation the warning offers ("restore the missing agents.list entries") is impossible.

### Steps to reproduce

1. Use the system agent at least once, so `~/.openclaw/agents/openclaw/agent/openclaw-agent.sqlite` exists.
2. Run `openclaw doctor`.
3. Read the state-integrity section.

### Expected behavior

`doctor` does not report reserved system agent ids as orphaned agent directories. That directory is the system agent's own store, not a stale roster leftover.

### Actual behavior

`openclaw` is reported as an orphan, with cleanup guidance pointing at `~/.openclaw/agents`. The finding is permanent for any install whose system agent has run.

### OpenClaw version

2026.9.3 (c973fe7)

### Operating system

Ubuntu 26.04.1 LTS

### Install method

Source checkout (git); gateway started from the local build

### Model

n/a — the finding is independent of routing

### Provider / routing chain

n/a

Six agents under `agents.entries`, with `agents.ownership: "explicit"` and `systemAgent.agentId` pointing at one of them. None of that matters: `openclaw` is flagged because a reserved id is absent from the roster by design.

### Logs, screenshots, and evidence

`listOrphanAgentDirs` (`src/commands/doctor-state-integrity.ts:198-247`) decides orphan status by comparing on-disk directory names against configured ids, exempting only the shared-auth-store owner and the live default agent dir. There is no exemption for reserved system agent ids.

The directory is live gateway-owned state, not a leftover. With the gateway running:

```console
$ for p in $(pgrep -f openclaw); do for f in /proc/$p/fd/*; do readlink "$f"; done; done | grep agents/openclaw
/home/ms/.openclaw/agents/openclaw/agent/openclaw-agent.sqlite
/home/ms/.openclaw/agents/openclaw/agent/openclaw-agent.sqlite-wal
/home/ms/.openclaw/agents/openclaw/agent/openclaw-agent.sqlite-shm

$ sqlite3 ~/.openclaw/state/openclaw.sqlite 'select * from agent_databases;'
...
openclaw|agents/openclaw/agent/openclaw-agent.sqlite|19|1789047849908|606208
```

Following the cleanup advice deletes that store. On this host I moved the directory aside and restarted the gateway; the warning cleared, and the gateway logged the store as gone rather than rebuilding it at boot:

```text
- Removed missing agent database registry entry
  /home/ms/.openclaw/agents/openclaw/agent/openclaw-agent.sqlite.
- Skipped missing registered agent database /home/ms/.openclaw/agents/openclaw/agent/openclaw-agent.sqlite.
```

So the suggested cleanup does not fail loudly — it silently discards the system agent's own session/transcript state, and the store is created again under the same reserved id on the next system agent turn (turns run with `agentId: SYSTEM_AGENT_ID`, `src/system-agent/agent-turn.ts:316-328`), which brings the warning back.

This is the same shape as #74313 (doctor flagging the live `agents/main` store), fixed then by exempting the live default agent dir. Reserved system agent ids need the same treatment.

### Impact and severity

Low severity, permanent false positive on any install whose system agent has run. Two concerns: it is unsafe framing (it invites deleting live system-agent state, and nothing warns that this is what is being removed), and it puts a finding that can never be resolved into every `doctor` run, which makes real orphans easier to miss.

### Additional information

The fix is a small guard in `listOrphanAgentDirs` reusing the existing `isReservedSystemAgentId` helper. PR follows.

emes:
Closes #144125.

Doctor reported the system agent's own directory as lacking an `agents.list` entry and suggested restoring that entry or removing the directory. Reserved system agents cannot enter the normal roster, so that advice cannot resolve the finding safely.

The orphan scan now uses the existing reserved-ID owner to exempt `openclaw` and retained `crestodian` state. Genuine orphan warnings, shared-auth/default-directory exemptions, and filesystem handling remain unchanged.

Validation on the pinned main and signed candidate:

- A real authenticated `openclaw.chat` turn through an isolated Gateway and deterministic local provider created `agents/openclaw/agent` naturally. Public history confirmed the conversation.
- The same `openclaw doctor --no-color` command and state reported three orphans on main. The candidate reports only the ordinary orphan; every other output line remains unchanged.
- The candidate's first Doctor invocation preserved all captured file bytes and modes, both database schemas, every exported table row, the agent registry, and the three persisted conversation records.
- Nine public controls cover ordinary/configured ownership, casing, shared authentication, default-directory realpaths, incomplete/moved/symlinked directories, missing/unreadable state, and reserved-name creation rejection.
- The new regression fails on main while the other 28 tests pass. The candidate passes all 35 focused tests, formatting, syntax lint, unchanged ratchets, and the normal commit hook.

The first main Doctor run performed automatic metadata initialization and changed SQLite bytes; its missing raw before-images remain an explicit evidence limit. A separately observed main run preserved all logical data, with shared-file changes explained exactly by checkpointing already committed journal pages. The candidate then preserved the entire captured state without restoring files, changing stamps, or repeating the original turn.

The unreadable-root control preserves the existing permission error before the orphan scan; it does not claim that the scan itself ran. Doctor's warning does not automatically delete state. This repair changes its ownership advice and preserves contributor credit.

### DIFF
--- a/src/commands/doctor-state-integrity.test.ts
+++ b/src/commands/doctor-state-integrity.test.ts
@@ -459,6 +459,20 @@ describe("doctor state integrity oauth dir checks", () => {
     expect(text).not.toContain("Examples:");
   });
 
+  it("ignores reserved system agent dirs that can never appear in agents.list", async () => {
+    createAgentDir("openclaw");
+    createAgentDir("crestodian");
+
+    const text = await runStateIntegrityText({
+      agents: {
+        list: [{ id: "main", default: true }],
+      },
+    });
+
+    expect(text).not.toContain("without a matching agents.list entry");
+    expect(text).not.toContain("Examples:");
+  });
+
   it("protects the shared legacy main auth-store dir for an ops-only roster", async () => {
     createAgentDir("main");
 
--- a/src/commands/doctor-state-integrity.ts
+++ b/src/commands/doctor-state-integrity.ts
@@ -54,6 +54,7 @@ import {
 } from "../infra/state-migrations.legacy-session-store.js";
 import { listConfiguredChannelIdsForReadOnlyScope } from "../plugins/channel-plugin-ids.js";
 import { normalizeAgentId } from "../routing/session-key.js";
+import { isReservedSystemAgentId } from "../system-agent/agent-id.js";
 import { shortenHomePath } from "../utils.js";
 import { repairHeartbeatPoisonedMainSession } from "./doctor-heartbeat-main-session-repair.js";
 import { describeHeartbeatSessionTargetIssues } from "./doctor-heartbeat-session-target.js";
@@ -215,6 +216,11 @@ function listOrphanAgentDirs(cfg: OpenClawConfig, stateDir: string): OrphanAgent
         const nestedAgentDir = path.join(agentsRoot, dirName, "agent");
         const hasNestedAgentDir = existsDir(nestedAgentDir);
         if (!hasNestedAgentDir) {
+          return false;
+        }
+        // Reserved system agent ids own a state dir but can never appear in
+        // agents.list, so their directories are never orphans.
+        if (isReservedSystemAgentId(agentId)) {
           return false;
         }
         if (


obviyus:
Thanks @emes!

**Fix:** Doctor now uses the canonical reserved-agent identity check, so live `openclaw` and retained `crestodian` directories are not reported as missing roster entries. Real orphan warnings and existing ownership/filesystem behavior remain intact.

**Proof:** A real public system-agent turn created the reserved store. The same public Doctor command reported three orphans on main and only the ordinary orphan on the candidate. The first candidate invocation preserved all captured state bytes, modes, schemas and rows; public history retained the original conversation. All nine public controls, 35 focused tests and the exact-head review passed.

| Before | After |
| --- | --- |
| Reserved system-agent directories received impossible roster/removal advice | Both reserved identities are recognized as owned state |
| Real unconfigured directory appeared alongside false positives | The genuine orphan remains visible |

The first main Doctor's automatic initialization and missing raw before-images remain a recorded limitation. No state was restored or rewritten to obtain candidate equality. The unreadable-root control preserves its existing preflight permission error; the warning itself does not automatically delete state.

The PR comparison initially included already-merged changes from an old base. Its corrected two-file comparison cleared the dependency guard on one normal rerun, with no source change or override.


Landed as e3db9654277b; both changed files match the reviewed candidate. [Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34535020246), and #144125 is closed. Michael Smith’s authorship and Ayaan’s co-author credit are preserved.

PR #144161, opened 2026-09-10
fix(tts): hold a trailing lone bracket across stream chunks

### LINKED ISSUE #144151: TTS stream cleaner mishandles a chunk split between the two opening brackets, leaking hidden speech text or swallowing the rest of the reply
## Summary
The streamed TTS directive cleaner (`createTtsDirectiveTextStreamCleaner`) only parks a partial tag in `pending` once it has seen both characters of `[[`. When a stream chunk ends on a single `[`, the lone bracket is emitted (or dropped) and the opener is never recognized, so the reply either leaks the raw `[[tts:text]]` markup plus the speech-only body to the channel, or silently swallows the rest of the reply.

## Environment
- Surface: TTS directive cleaning on the streamed block lane, channel and ACP

## Steps to reproduce
1. With TTS auto mode on (so `shouldCleanTtsDirectiveText` holds), have the assistant emit `Intro [[tts:text]]hidden speech[[/tts:text]] visible` and let the stream split the text between the two opening brackets. Driving the exported cleaner directly, that is:
   - `push("Intro [")` then `push("[tts:text]]hidden speech[[/tts:text]] visible")`, then `flush()`.
2. Same reply, split inside the closer instead:
   - `push("Intro [[tts:text]]hidden speech[")` then `push("[/tts:text]] visible")`, then `flush()`.

## Expected
Both splits should behave like the unsplit input and like a split that lands after `[[`: the directive markup and the speech-only body are hidden, and the visible text `Intro  visible` is delivered.

## Actual
- Split inside the opening `[[tts:text]]`: the visible text becomes `Intro [[tts:text]]hidden speech visible`. The raw markup and the entire speech-only body, which was explicitly marked hidden, are published to the channel.
- Split inside the closing `[[/tts:text]]`: the close is never recognized, `insideHiddenTextBlock` stays `true` forever, and this block plus every later block plus `flush()` all return `""`. The visible answer is truncated at the split point and the remainder is never delivered.

## Root cause
`src/tts/directives.ts:145-151`, the `tagStart === -1` branch of `push`:

```ts
const tagStart = input.indexOf("[[", index);
if (tagStart === -1) {
  if (!insideHiddenTextBlock) {
    output += input.slice(index);
  }
  break;
}
```

The cross-chunk guard lives further down, at `src/tts/directives.ts:157-162`, and only arms after `[[` has been found:

```ts
const tagEnd = input.indexOf("]]", tagStart + 2);
if (tagEnd === -1) {
  // Directive delimiters can cross chunk boundaries; buffer from the
  // opener so a partial tag never leaks to a streamed client.
  pending = input.slice(tagStart);
  break;
}
```

A chunk ending on a single `[` makes `indexOf("[[")` return `-1`, so the lone bracket is emitted (outside a hidden block) or discarded (inside one) and `pending` is left empty. The next chunk then starts at `[tts:text]]...` and no opener is ever recognized for that tag.

Suggested fix, holding a trailing lone `[` in `pending` rather than emitting or dropping it:

```ts
if (tagStart === -1) {
  const tail = input.endsWith("[") ? input.length - 1 : input.length;
  if (!insideHiddenTextBlock) { output += input.slice(index, tail); }
  pending = input.slice(tail);
  break;
}
```

`flush()` already emits `pending` when not inside a hidden block, so a genuine trailing `[` at end of stream still surfaces.

## Sibling surfaces
Both streamed lanes share this cleaner instance shape and are affected.

- Channel block lane: `src/auto-reply/reply/dispatch-from-config.prepare-execution.ts:102-109` creates the cleaner whenever `shouldCleanTtsDirectiveText` holds (TTS auto mode on, `tts.modelOverrides.enabled !== false`); `src/auto-reply/reply/dispatch-from-config.execute.ts:520` calls `cleanBlockTtsDirectiveText.push(payload.text)` and the result replaces the block's visible text sent by `dispatcher.sendBlockReply`.
- ACP lane: `src/auto-reply/reply/acp-projector.ts:456-458` hard-flushes `liveBufferText` at `ACP_LIVE_HARD_FLUSH_CHARS = 480` (`src/auto-reply/reply/acp-projector.ts:105`), at an arbitrary character offset, and `src/auto-reply/reply/dispatch-acp-delivery.ts:363` pushes that raw block payload text through the cleaner. A block that cleans to empty is recorded with `needsFinalDelivery: false` (`src/auto-reply/reply/dispatch-acp-delivery.ts:365-370`), so text swallowed by the stuck-open branch is never re-sent by the final delivery path.

There is no downstream rescue on this lane: the block payload text produced here is what the dispatcher sends, and there is no later directive strip on the block lane. The final reply does run `parseTtsDirectives` over complete text (`src/tts/tts-payload.ts:255`), but on the ACP streaming lane those blocks are already marked `needsFinalDelivery: false`.

`EmbeddedBlockChunker` normally breaks at whitespace, which is why this is intermittent rather than universal, but `src/auto-reply/reply/acp-projector.ts:451` (`truncateUtf16Safe` at `maxOutputChars`) and the 480-char hard flush both split at arbitrary offsets, and any transport that re-chunks can land the same way.

Adjacent but distinct prior art: PR #137329 (open) also edits `createTtsDirectiveTextStreamCleaner`, but only the classification path after a complete `[[...]]` has been located (adding `resolveFreeTextTtsBody` for free-text bodies). It does not touch the `tagStart === -1` branch, and its added cases split only after `[[`. PR #83988 (merged) is final-mode text settlement, not the streaming cleaner. Existing coverage at `src/tts/directives.test.ts:375-389` likewise splits only after `[[` (`"Hello [[tts:voice=al"`, `"Shown [[tts:text]]hid"`), never between the two brackets, which is why this has gone unnoticed.

Severity: P2. One branch is silent user-visible text loss, the other is disclosure of content the model explicitly marked speech-only. It requires a chunk split at a specific offset, so it fires intermittently rather than on every reply.

## Real behavior proof
Behavior addressed: a streamed chunk boundary falling between the two opening brackets of a `[[tts:...]]` directive.
Real environment tested: the real exported `createTtsDirectiveTextStreamCleaner` from `src/tts/directives.ts`, driven directly at commit c973fe7e721905ba2318994f1f47ae97677cf9ac.
Exact steps or command run after this patch: `node --import ./scripts/tsx.mjs probe-tts.mts`, which pushes the four chunk sequences below through one cleaner each and prints `push(...)` concatenated with `flush()`.
Evidence after fix:
```
A whole             : "Intro  visible"
B split in tag name : "Intro  visible"
C split between [ [ : "Intro [[tts:text]]hidden speech visible"
D split in closer   : "Intro "

A whole             : "Intro  visible"
B split in tag name : "Intro  visible"
C split between [ [ : "Intro  visible"
D split in closer   : "Intro  visible"
```
Observed result after fix: A and B are controls and are already correct, which shows the existing cross-chunk guard works once `[[` is seen; only C and D, where the chunk boundary lands between the two brackets, leak the hidden body plus markup (C) or swallow the remainder of the reply (D).
What was not tested: I drove the cleaner directly rather than end to end through a live channel or a live ACP client, so the exact offsets at which a real provider stream or the 480-char hard flush lands between the brackets are shown by code reading, not captured from a live session. I also did not measure how often real streams split at that offset, and I did not apply or verify the suggested patch.

Bartok9:
This repair includes registered tests for both delivery lifecycles and moves existing helpers to keep the coordinator within its size limit.

## What Problem This Solves

`createTtsDirectiveTextStreamCleaner` only parks a partial tag after it has seen both characters of `[[`. When a stream chunk ends on a single `[`, that bracket is emitted (or dropped) and the opener is never recognized.

Fixes #144151. Split openers leak speech-only text; split closers swallow the visible suffix.

## Fix

Retain split delimiters and drain literal carry once at completion. Normal dispatch and ACP each use their existing lifecycle adapter over the same source receipt. That receipt owns source completion and ordering.

ACP recovery reuses the prepared payload and send operation instead of rebuilding a string. It preserves reply metadata, avoids repeated prefixes, transforms, media and speech, and respects cancellation and queue custody. The stateful stream tests now run in their existing auto-reply project.

## Growth record

Ayaan, 2026-09-15: "accept the growth, land #144161".

Production is **+221 net lines against the merge base**, accepted by the owner for the ACP terminal drain, the prepared-payload recovery that replaced the string fallback, and the helper move into the existing payload module that keeps the coordinator under the 700-line rule. This adds no state owner.

## Evidence

[Gateway/Discord captures and correction evidence](https://gist.github.com/obviyus/f14fbf2109f3bb30d68e2af5e54b2d61): split tags and literal tails reach the recipient correctly; audio stays separate from visible speech-only text. Registered recovery tests verify exact bytes, one transform/media delivery, metadata, cancellation and released custody. Real packaged Discord faults verify retained queue custody on both builds.

The 384 owner/sibling tests passed on head and fresh merge. After correcting test routing, all 12 stream cases pass under auto-reply, 136 routing tests pass, and the hosted unit-fast selector passes 9,235 tests with two skips. Local preflight passed.

Record correction: tested merge `eaf170ab` included main `2e7aa1b7` through an intermediate merge; the earlier reported `f1cdcd51` was only an ancestor. Later merge `904f4d9e` has exact parents main `78520903` and candidate `7614c08e`. Earlier captures retain their actual identities.

The ledger records removal of the newline between same-source fragments, leaving `"Intro  visible"` with two spaces; independent sources keep their separator. Duplicate helper-only bracket rows were removed while registered normal/ACP regressions and contributor history remain.

## Consumers

Normal channel dispatch, ACP direct/routed delivery, terminal-only surfaces and retained followups retain their owners. Source receipts serialize continuations; actual delivery controls recovery and transcript coverage. Deferred captions retain complete-text fallback.

## Compatibility

No new configuration, schema or public API.

### DIFF
--- a/src/tts/directives.test.ts
+++ b/src/tts/directives.test.ts
@@ -401,4 +401,20 @@ describe("createTtsDirectiveTextStreamCleaner", () => {
     expect(cleaner.push("See [[note")).toBe("See ");
     expect(cleaner.flush()).toBe("[[note");
   });
+
+  it("hides markup when a chunk split lands between the two opening brackets", () => {
+    const cleaner = createTtsDirectiveTextStreamCleaner();
+
+    expect(cleaner.push("Intro [")).toBe("Intro ");
+    expect(cleaner.push("[tts:text]]hidden speech[[/tts:text]] visible")).toBe(" visible");
+    expect(cleaner.flush()).toBe("");
+  });
+
+  it("hides markup when a chunk split lands between the two closing brackets", () => {
+    const cleaner = createTtsDirectiveTextStreamCleaner();
+
+    expect(cleaner.push("Intro [[tts:text]]hidden speech[")).toBe("Intro ");
+    expect(cleaner.push("[/tts:text]] visible")).toBe(" visible");
+    expect(cleaner.flush()).toBe("");
+  });
 });
--- a/src/tts/directives.ts
+++ b/src/tts/directives.ts
@@ -144,9 +144,13 @@ export function createTtsDirectiveTextStreamCleaner(): TtsDirectiveTextStreamCle
       while (index < input.length) {
         const tagStart = input.indexOf("[[", index);
         if (tagStart === -1) {
+          // A chunk can end on a single "["; hold it so the next chunk can
+          // complete "[[" instead of leaking markup or swallowing the rest.
+          const tail = input.endsWith("[") ? input.length - 1 : input.length;
           if (!insideHiddenTextBlock) {
-            output += input.slice(index);
+            output += input.slice(index, tail);
           }
+          pending = input.slice(tail);
           break;
         }
 


obviyus:
landpr-approve-admin-merge
pr: 144161
head: 691ce254d87023a51e0d26bb3051db867a5f5a74
scope: owner-review

obviyus:
Thanks @Bartok9!

**Proof:** Discord streaming checks preserve split speech tags and the literal `See [` exactly once. Recovery tests preserve reply metadata without repeated prefixes, media or speech.

- **Bug:** Split speech tags could expose hidden speech or lose visible reply text.
- **Fix:** Streaming delivery settles the visible tail once and recovers only an undelivered prepared continuation.

| Before | After |
| --- | --- |
| Split tags could leak hidden speech or swallow the visible suffix. | Speech stays hidden and the visible suffix arrives. |
| A failed literal tail could disappear or repeat a prefix during recovery. | Only the missing tail is recovered, with its reply identity preserved. |

PR #144163, opened 2026-09-10
fix(plugin-sdk): accept return-value gateway handlers in plugin registration type

### LINKED ISSUE #142718: [Bug]: registerGatewayMethod return-value support is still rejected by the exported handler type

JamieJ5926:
## What Problem This Solves

Fixes #142718: `registerGatewayMethod` turns returned handler values into Gateway responses at runtime (since #85785), but the exported handler type still required `void | Promise<void>`, so the supported return-value style failed TypeScript with TS2345 and plugin authors needed a cast.

## Change

- Adds `PluginGatewayRequestHandler` (`GatewayRequestHandler | ((opts) => unknown)`, covering async and sync returns) and uses it for the plugin-facing `registerGatewayMethod` input only (`plugin-api.types.ts`, `registry-registrars-network.ts`). Core gateway handlers stay callback-only; the adapter's respond/return precedence is untouched.
- Exports the new type (plus `GatewayRequestHandler`) from `openclaw/plugin-sdk/gateway-runtime`; bumps the public-exports surface budget 4447 -> 4449.
- Adds a type-level regression assertion in `scheduled-turns.contract.test.ts` so a future narrowing breaks compilation.
- Follow-up: explicit `return undefined` on the workboard gateway error path, which the widened union pushed under `noImplicitReturns` (TS7030). Runtime behavior unchanged.

## Evidence

Failing-first compiler repro: a minimal consumer registering `api.registerGatewayMethod("demo.check", async () => ({ ok: true }))` against current main fails with:

```
error TS2345: Argument of type '() => Promise<{ ok: boolean; }>' is not assignable to parameter of type 'GatewayRequestHandler'.
```

(same for the sync-return shape). After this change:

- `tsgo -p tsconfig.core.json`: exit 0, 0 errors (was 2x TS2345 on the repro before the fix)
- `tsgo -p tsconfig.ui.json`, `-p tsconfig.extensions.json`: exit 0 (extensions caught one TS7030, fixed with the explicit return above)
- `tsgo -p test/tsconfig/tsconfig.core.test.plugins-platform.json`: exit 0
- `pnpm tsgo:test` (all core-test shards + extensions:test + test:root) and `tsgo:scripts`: exit 0
- `vitest run src/plugins/loader.test.ts`: 198 passed (existing #85785 return-value runtime coverage)
- `vitest run src/plugins/contracts/scheduled-turns.contract.test.ts src/plugins/contracts/plugin-sdk-runtime-api-guardrails.test.ts`: 33 passed
- `plugin-sdk-surface-report.mts --check`: passes
- `oxfmt --check` + `oxlint` on all touched files: clean

### DIFF
--- a/extensions/workboard/src/gateway.ts
+++ b/extensions/workboard/src/gateway.ts
@@ -55,6 +55,7 @@ export function registerWorkboardGatewayMethods(params: {
             return await store.runOperation(() => handler(request));
           } catch (error) {
             respondError(request.respond, error);
+            return undefined;
           }
         },
         options,
--- a/scripts/plugin-sdk-surface-report.mts
+++ b/scripts/plugin-sdk-surface-report.mts
@@ -363,7 +363,8 @@ export function readPluginSdkSurfaceBudgets(env: NodeJS.ProcessEnv = process.env
       // -1: one exec policy object replaces two deprecated comparator exports.
       // +1: approved bounded TAR inspection through the archive admission owner.
       // +1: canonical runtime-context classifier for native history projection.
-      4447,
+      // +2: plugin-facing gateway handler contract and its core handler counterpart.
+      4449,
       env,
     ),
     publicFunctionExports: readPluginSdkSurfaceBudgetEnv(
--- a/src/gateway/server-methods/shared-types.ts
+++ b/src/gateway/server-methods/shared-types.ts
@@ -461,8 +461,13 @@ export type GatewayRequestHandlerOptions = {
   hasCurrentClientAuthority?: () => boolean;
 };
 
-/** Single gateway method implementation. */
+/** Single core gateway method implementation. */
 export type GatewayRequestHandler = (opts: GatewayRequestHandlerOptions) => Promise<void> | void;
+
+/** Plugin-facing gateway handler; the registration adapter turns returned values into responses. */
+export type PluginGatewayRequestHandler =
+  | GatewayRequestHandler
+  | ((opts: GatewayRequestHandlerOptions) => unknown);
 
 /** Registry fragment keyed by gateway protocol method name. */
 export type GatewayRequestHandlers = Record<string, GatewayRequestHandler>;
--- a/src/plugin-sdk/gateway-runtime.ts
+++ b/src/plugin-sdk/gateway-runtime.ts
@@ -31,7 +31,11 @@ export {
   parseGatewayPayload as safeParseJson,
   respondUnavailableOnNodeInvokeError,
 } from "../gateway/server-methods/nodes.helpers.js";
-export type { GatewayRequestHandlers } from "../gateway/server-methods/types.js";
+export type {
+  GatewayRequestHandler,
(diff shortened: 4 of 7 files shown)


JamieJ5926:
Maintainer note on the remaining red (no code action outstanding): head 39e89b4b shows prod-types, test-types, lint-core-2, and context checks all green; the sole fail is checks-node-compact-small-18 in run 34500829865 (job 102951234157): src/infra/install-flow.test.ts hit its 1000ms zip-extraction wall-clock timeout under the loaded CI runner. That test imports only node builtins, jszip, vitest, test-helpers, and ./install-flow.js — zero import edge to this PR's 7 touched files (gateway/plugin-sdk types + workboard explicit return + contract test), and it passes 6/6 locally. The ci-gate fail is the knock-on aggregate. I cannot rerun it (gh run rerun --failed: needs admin). A shard rerun should clear it.

PR #116302, opened 2026-07-30
fix: Ask OpenClaw reports no API key when the provider uses a file SecretRef

harjothkhara:
Closes #115062

## What Problem This Solves

Fixes an issue where users whose model provider key is a file `SecretRef` could not start the Ask OpenClaw system assistant in the Control UI. The setup check failed with `No API key found for provider "<id>"`, even though `openclaw secrets audit` reported the ref as clean and normal agent sessions reached the same provider and model successfully.

Because the error told users to sign in again or re-run `openclaw configure`, it pushed people toward reconfiguring credentials that were in fact valid.

## Why This Change Was Made

The Gateway resolves non-env `SecretRef` values once at startup into its runtime config snapshot. `resolveManagedSecretRefRuntimeProviderAuth` will only hand back that resolved credential when the caller's provider entry matches the applied snapshot — a deliberate check, so a config the Gateway cannot prove is the applied one never receives a resolved secret.

Several system-agent surfaces re-read the config file from disk before acting. A fresh read leaves `SecretRef` values unresolved, so its provider entry can never match, and the credential is refused. Normal agent runs are unaffected because they pass the Gateway's live runtime config.

This adds one shared selector, `resolveAppliedSnapshotConfig`, which returns the live runtime config while the file on disk is still exactly what the Gateway applied, and the freshly materialized config once they diverge. It routes the decision through the existing `selectApplicableRuntimeConfig` seam, matching how `server-methods/send.ts`, `cron/isolated-agent/run-config.ts`, `agents/tool-runtime-config.ts` and `web-search/runtime.ts` already resolve the same question. Credential resolution rules are untouched.

It is applied at every system-agent surface that re-reads config and then depends on provider credentials:

- `verifySetupInference` and `completeSetupInference` — the probe itself.
- The probe's post-run drift check, so an unresolved `SecretRef` is not misread as route drift.
- `resolveSystemAgentVerifiedInferenceRoute` and `hasCurrentSystemAgentOwnerPluginArtifacts` — without these the probe would succeed and the session that follows would still fail. A verified binding is fingerprinted from the config that was probed, and `projectInferenceRoute` embeds the whole provider entry, `apiKey` included. Leaving revalidation on a fresh read would reject every later system-agent turn, replacing a clear credential error with a vaguer stale-route one.
- `verifyCurrentSetupInference`, which gates the `openclaw.setup` persistent operation, and the `inference-fallback` ladder, which otherwise filters `SecretRef`-keyed providers out of the candidate list.

[...]

PR #144165, opened 2026-09-10
fix(system-agent): surface the exact harness mismatch from bound verification

### LINKED ISSUE #134471: Managed control fails owner validation on healthy Codex route after restart (v2026.8.1)
## Summary

On OpenClaw **v2026.8.1**, the managed-control path (`functions.openclaw`) fails before dispatch on a healthy OpenAI/Codex route with:

```text
OpenClaw requires working inference: The verified inference owner changed before validation completed. Retry the inference check.
```

Normal Gateway operation and the configured model route are healthy. This blocks managed configuration writes and managed Gateway restart.

## Environment

- OpenClaw v2026.8.1
- Linux x86_64, systemd user Gateway service
- System agent: `main`
- Primary: `openai/gpt-5.6-terra@openai:pavlov.mm@gmail.com`
- Fallback: `openai/gpt-5.6-luna@openai:default`
- Runtime status reports: `openai via codex ... status=usable`

## Reproduction

1. Confirm the Gateway is healthy and the OpenAI/Codex runtime is usable:

```text
Gateway connectivity probe: ok
Runtime auth: openai via codex ... status=usable
```

2. Invoke a read-only managed-control request (no configuration or state change requested).
3. It fails before dispatch with the error above.

## Checks already performed

- Restarted the Gateway through the systemd user service; the Gateway returned healthy and the Codex runtime still reported usable.
- Retried the read-only managed-control request immediately after the clean restart; the same error recurred.
- Temporarily removed the only fallback, validated the config, and retried. The error still recurred. The fallback was restored.
- `openclaw config validate` succeeds.

Relevant Gateway log evidence from an earlier occurrence:

```text
[agents/harness] Codex agent harness failed; not falling back to embedded OpenClaw backend
```

The surfaced managed-control error hides the underlying owner/harness/registry failure, so there is no actionable local diagnosis.

## Expected

Either:

1. managed control runs successfully using the verified Codex route; or
2. it returns the specific underlying cause (for example, Codex runtime registry/harness loading failure), rather than generic owner drift.

## Related

- #115704, #123141, and #124829 appear relevant: their release notes describe Codex runtime reload/registry fixes around verification and activation.
- #134376 has a similar symptom but was closed as not reproducible. This report adds a primary-only reproduction and a post-restart result.

No credentials, tokens, or private paths are included.

JamieJ5926:
Fixes #134471.

## What Problem This Solves

On a healthy Codex route, managed control (Gateway functions.openclaw path) fails before dispatch with a generic owner-drift error that hides the real cause. When the successful probe run reports a different embedded agent harness than the configured one (for example configured codex vs reported openclaw), bound verification surfaced 'The verified inference owner changed before activation completed. Retry the inference check. (Could not load the openclaw runtime plugin.)' instead of naming the harness mismatch. Root cause: revalidateSetupInferenceOwner attempted plugin generation for the reported harness before comparing it with the configured embedded harness, so the misleading plugin-load failure replaced the true mismatch cause wholesale.

## Evidence

- New regression test 'surfaces the exact harness mismatch to bound verification' in src/system-agent/setup-inference.test.ts drives the exact Gateway caller shape (verifySetupInference with bindSession:true, configured codex harness, reported openclaw harness). Failing output before the fix: 'The verified inference owner changed before activation completed. Retry the inference check. (Could not load the openclaw runtime plugin.)'. Passes after.
- Scoped suites green, 276 passed: setup-inference.test.ts, inference-fallback.test.ts, verified-inference.test.ts, setup-inference-error-reporting.test.ts, setup-inference-test.test.ts, gateway/server-methods/system-agent.test.ts.
- oxfmt --check and oxlint --deny-warnings clean on all touched files.

## Change

This keeps owner validation fail-closed and propagates the specific cause instead of replacing it wholesale:

- verified-inference.ts: new shared assertEmbeddedHarnessMatch helper, single-sourcing the 'used agent harness X instead of Y' message the binding check already threw.
- revalidate-inference-owner.ts: fail fast with that specific error when a concrete configured embedded harness differs from the reported one, before plugin generation. The redaction path is unchanged.

Commits are stacked red-then-green: failing regression test first, narrow fix on top.

Conflict watch: open PR #116302 also touches verified-inference.ts (SecretRef setup). This PR does not absorb it; the helper hunk is small and should rebase cleanly.

Parked: live Codex runtime diagnosis (needs Linux/systemd env, not available here); proved with unit-level tests instead.

### DIFF
--- a/src/system-agent/revalidate-inference-owner.ts
+++ b/src/system-agent/revalidate-inference-owner.ts
@@ -15,6 +15,7 @@ import { getPluginRuntimeLoadContext } from "../plugins/runtime/load-context.js"
 import type { SystemAgentConfiguredRoute } from "./inference-route.js";
 import { parseRef } from "./setup-inference-plan-helpers.js";
 import {
+  assertEmbeddedHarnessMatch,
   createSystemAgentVerifiedInferenceBinding,
   type SystemAgentVerifiedInferenceBinding,
   type SystemAgentVerifiedInferenceDeps,
@@ -91,6 +92,12 @@ export async function revalidateSetupInferenceOwner(params: {
   const successfulHarnessId =
     params.auth.agentHarnessId?.trim() ||
     (configuredHarnessId && configuredHarnessId !== "auto" ? configuredHarnessId : undefined);
+  if (params.route.runner === "embedded") {
+    assertEmbeddedHarnessMatch({
+      configuredHarnessId,
+      reportedHarnessId: params.auth.agentHarnessId?.trim() || undefined,
+    });
+  }
   const createBinding = () =>
     (
       params.deps.createSystemAgentVerifiedInferenceBinding ??
--- a/src/system-agent/setup-inference.test.ts
+++ b/src/system-agent/setup-inference.test.ts
@@ -6778,6 +6778,96 @@ describe("verifySetupInference", () => {
       await removeOAuthTestTempRoot(stateDir);
     }
   });
+  it("surfaces the exact harness mismatch to bound verification", async () => {
+    const stateDir = await suiteTempRootTracker.make("case");
+    vi.stubEnv("OPENCLAW_STATE_DIR", stateDir);
+    const profileId = "openai:default";
+    const credential = {
+      type: "oauth" as const,
+      provider: "openai",
+      access: "test-access",
+      refresh: "test-refresh",
+      expires: Date.now() + 3_600_000,
+    };
+    const authFingerprint = fingerprintAuthProfileCredential({ profileId, credential });
+    if (!authFingerprint) {
+      throw new Error("missing external Codex auth fingerprint");
+    }
+    const config = {
+      agents: {
+        defaults: {
+          model: { primary: "openai/gpt-5.6-sol" },
+          models: {
+            "openai/gpt-5.6-sol": { agentRuntime: { id: "codex" } },
+          },
+        },
+      },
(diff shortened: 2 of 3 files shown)


JamieJ5926:
Review response on real-behavior proof (contributor, automated lane).

What is proved at this head: the new regression test drives the exact Gateway caller shape (verifySetupInference with bindSession:true, configured codex harness, reported openclaw harness). Before the fix the Gateway-visible error was 'The verified inference owner changed before activation completed. Retry the inference check. (Could not load the openclaw runtime plugin.)'. After the fix it carries 'used agent harness "openclaw" instead of "codex"' with status auth, fail-closed, redaction path unchanged. The Gateway wrapper interpolates inference.error verbatim (unchanged line, covered by the existing gateway suite). CI is fully green.

What is parked and why: a live Gateway plus installed Codex runtime observation needs a Linux/systemd host with Codex credentials. This workstation is darwin and the human is asleep, so no credentialed live run was attempted. Suggested live proof for later: on the systemd host, configure the embedded codex route, force the probe to report the built-in harness, invoke a read-only managed-control request, and confirm the Gateway-visible error names the harness pair. Terminal output or redacted logs count.

On the P1 merge-risk item: agreed this PR is diagnostic, not a runtime cure. It does not claim the reporter's post-restart Codex failure is fixed. It makes the next occurrence identify itself instead of hiding behind generic owner drift, which is what issue #134471 option 2 asks for.

PR #144009, opened 2026-09-10
fix(agents): use the selected configured context limit

steipete:
Related: #130706, #142100.

## What Problem This Solves

A prefixed or equivalent model row appearing earlier in configuration can supply the context limit for a different selected model. Context lookup and the context-window guard also implement separate matching rules.

## Why This Change Was Made

Share the existing exact configured-row owner between context lookup and the window guard. Exact selected rows take priority while shipped self-prefix and OpenRouter compatibility remain supported. This removes 41 production lines and two obsolete assertion allowances.

## User Impact

The selected model receives its own configured context cap, so an unrelated sibling cannot cause premature context warnings or change its limit.

## Evidence

129 focused tests, a 20-case fixture retest, independent P2 review, and the full changed-file gate passed. Earlier gate failures required only an exact shrink-only assertion-baseline update and a fresh-array test-fixture correction.

One exact-commit sourcePerformance build and 16 compiled context/guard scenarios passed. All 9,258 captured root/workspace output files remained unchanged; no checkout source imports or network requests occurred, and isolated state/processes were cleaned up. This is compiled boundary proof, without a live Gateway or external provider. The runtime build excludes UI and declarations.

Tested commit: `08383a80d9e55bc299e92e6db022e2e98f8b3eab`.

CI refresh: the unchanged task patch is replayed onto main `3943149f8ac269fc138ca21b5786ff2a9db3dd23`, including the navigation repair in #144037. Every changed-file postimage matches the previous reviewed head `c3c72733b6b5902e97c47a657feb886c9151ff51`. Original exact-commit runtime proof remains preserved and is not restamped; fresh CI is running for `aca25caeeb842394b10c83e2551e69052c9ed8bc`.

PR #124303, opened 2026-08-15
fix(sessions): track context window provenance

fuller-stack-dev:
## Summary

- record whether a persisted session context window came from runtime telemetry, a runtime configured with an authored cap, or ordinary config/model resolution
- persist the producing provider/model/harness ownership tuple and invalidate the grouped context state across model, runtime, reset, transcript rotation, command, auto-reply, cron, and settled-turn lifecycle changes
- share the typed trust rule across Gateway rows, interactive `/status`, and CLI/Gateway JSON status
- keep current authored caps authoritative, clamp effective context to a lower native window, and preserve locked native-session ownership

## Root cause

`SessionEntry.contextTokens` was an unqualified last-run cache. After a same-provider/model OpenClaw-to-Codex transition, the current runtime could resolve to Codex while the prior OpenClaw 272k window still won. Fresh/upgraded entries without producing-harness provenance had the same problem, and transcript/budget fallback or status readers could reintroduce the stale value after the Gateway row itself was corrected.

The producer already has typed provider/model/harness and context-window facts. This change records those facts where they occur, clears them at lifecycle ownership changes, and consumes them only when the persisted producer tuple still matches the current selection. It does not infer provenance from UI state, tool text, or provider-specific strings.

## Observable behavior

- a fresh-install/legacy 272k entry with no producing harness/source resolves the current same-model Codex/default 1M window
- a persisted 272k OpenClaw run resolves 1M after the same model moves to Codex
- stale transcript and budget telemetry remain historical after restart and cannot retake current-window ownership
- matching runtime-observed telemetry remains authoritative when no current authored cap supersedes it
- an authored effective cap is clamped by a lower non-fixed native window without inflating lower discovered budgets or weakening fixed-provider contracts
- locked sessions preserve their session-owned native window in Gateway, `/status`, and JSON status
- successful settled-turn recovery retains the authoritative context value and source instead of falling back to an outer stale budget
- `/new`, reset, message-cut rotation, model/runtime/harness changes, and repair paths clear associated provenance atomically

The effective Codex window remains 1M while the catalog-native maximum remains 1.05M. The configured 700k native auto-compaction threshold is unchanged, and the configured runtime remains Codex.

## Verification

- Unmodified-main red: same provider/model moved from OpenClaw272k to Codex while `sessions.list` retained 272k ([run 31906395919](https://github.com/openclaw/openclaw/actions/runs/31906395919)).
[...]

PR #144168, opened 2026-09-10
fix(agents): clamp configured input caps to model windows

steipete:
Related: #144009, #124303

## What Problem This Solves

Fixes an issue where a configured input cap larger than a model's native context window could bypass the startup minimum and overstate its compaction budget. A custom model with a 3,000-token native window and a 16,000-token input cap was admitted with a 16,000-token budget.

## Why This Change Was Made

Share the existing configured-limit calculation between cache reporting and the runtime guard. Apply the configured or fixed provider ceiling before checking the minimum, while preserving exact model-row selection, omitted-cap behavior and explicit overrides above lower discovered metadata.

The shared owner retains each caller's numeric rules: the guard uses finite whole tokens, while cache reporting keeps its existing positive-number projection. The configuration schema permits positive fractional native windows.

## User Impact

Startup and compaction respect configured model ceilings. A native window below the supported minimum produces the existing actionable error even when its input cap is larger.

## Evidence

Eleven regression/control cases were executed once on the baseline: four failed and seven passed. The same cases then all passed with the production repair, alongside 20 existing controls for row selection, fixed windows, explicit overrides and caller budgets. Baseline routes were completed separately because the wrapper stopped after each failing route.

An initial static pass found a test-only map-spread lint issue. Replacing it with an equivalent fresh-object Object.assign mapping preserved the cases, assertions and all production bytes; all 29 final checks then passed. Runtime tests were not repeated solely for that syntax change. Independent review found no actionable findings.

Production adds 35 lines for the shared result and existing numeric rules; the row selector becomes private. Tests add 147 lines. The focused tests cover ordinary setup/compaction and metadata helpers. The additional CLI proof below exercises persisted configuration and real agent startup with a synthetic loopback provider.

### Config-loaded startup and saved-state continuity

Six runs used the supported `pnpm openclaw agent exec` entry point. Both pinned source versions built through the normal wrapper and loaded actual config files.

| Source and state | Native window / input cap | Result | Provider requests |
| --- | --- | --- | --- |
| Baseline, fresh | 3000 / 16000 | Completes incorrectly | 1 |
| Baseline, fresh control | 5000 / 16000 | Completes | 1 |
| Candidate, baseline-created state | 3000 / 16000 | Existing minimum-4000 error | 0 |
| Candidate, fresh | 3000 / 16000 | Same error | 0 |
| Candidate, baseline-created control state | 5000 / 16000 | Completes with effective `ctx=5000` | 1 |
| Candidate, fresh control | 5000 / 16000 | Completes with effective `ctx=5000` | 1 |

Config bytes stayed identical. The reused-state cases began with byte-identical baseline-created state, including SQLite databases. Successful runs returned the fixed mock response; rejected runs emitted the existing actionable floor error before any provider dispatch. All processes closed cleanly.

This verifies source/config startup and saved-state continuity, not npm installation, a released-package upgrade, Gateway service startup, or a live provider. Each process used isolated state and an environment allowlist, with outbound traffic restricted to the task-owned loopback provider.

The intentional behavior change is accepted on this evidence: an oversized input cap no longer masks an unsupported native window. The config format and saved config bytes are unchanged; operators receive the existing guidance to correct the model limit or select a larger model.

### DIFF
--- a/src/agents/context-resolution.ts
+++ b/src/agents/context-resolution.ts
@@ -72,7 +72,7 @@ function resolveProviderModelRef(params: {
 }
 
 /** Preserve shipped self-prefixed context refs after exact configured-row selection. */
-export function resolveConfiguredProviderModel(
+function resolveConfiguredProviderModel(
   cfg: OpenClawConfig | null | undefined,
   provider: string,
   model: string,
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f84848cf43a065c891fe1dc4c564c68b9e2a1a81](https://github.com/openclaw/openclaw/pull/144168/commits/f84848cf43a065c891fe1dc4c564c68b9e2a1a81)
- Landed commit: [acef89b91754f2da7a56552ed8d455290062c51d](https://github.com/openclaw/openclaw/commit/acef89b91754f2da7a56552ed8d455290062c51d)

PR #136900, opened 2026-09-03
fix(gateway): hide foreign drafts from session describe

vincentkoc:
## What Problem This Solves

Operator-role users who knew a foreign draft session key could inspect the
draft through exact-key reads. `sessions.describe` disclosed identity-bearing
metadata, while `sessions.get` could return the draft transcript.

## Why This Change Was Made

Exact-key session reads now share one canonical visibility owner.
`sessions.describe` and `sessions.get` apply the existing operator-boundary
draft policy before disclosure:

- foreign `view`, `suggest`, and `write` callers receive non-disclosing
  successful results;
- draft creators and administrators retain access;
- deployments without configured operator roles retain the shipped
  proof-of-knowledge behavior.

`sessions.get` also revalidates configuration, canonical target, store path,
session identity, and visibility after the asynchronous transcript read. A
draft visibility change or same-key session replacement during that read
returns an empty result instead of stale transcript data.

## User Impact

Foreign drafts are no longer visible through either exact-key metadata or
transcript reads at an operator role boundary. Existing privileged and no-role
behavior is unchanged.

## Evidence

- Exact head: `46b6461b1025df093fdc26e404b29a9b317f621b`
- Current `origin/main` checked: `f0765dff583c78918be4b6cd672a03bfc707f307`;
  the verified head merges without conflicts.
- Pre-fix regression: a foreign `view` caller received the recognizable draft
  transcript instead of an empty result.
- `node scripts/run-vitest.mjs
  src/gateway/server-methods/sessions-read.test.ts
  src/gateway/server.auth.presence-audience.test.ts`: 23 tests passed.
- Real Gateway WebSocket RPC coverage verifies `sessions.describe` and
  `sessions.get` for creator, reader, writer, restricted-agent, and
  administrator callers.
- Direct-handler coverage verifies `view`, `suggest`, `write`, explicit member,
  creator, administrator, and no-role behavior.
- Race coverage pauses transcript reading, then changes visibility to draft or
  replaces the same-key session; both cases return no messages.
- Core production and test typechecks, targeted oxlint, formatting,
  `git diff --check`, and changed-lane classification passed.
- The exact assertion-safety ratchet now passes: the moved `sessions.get`
  assertion carries its dispatch invariant, and the old `sessions-read.ts`
  grandfathered count shrinks from 2 to 1.
- Independent Sol autoreview through P1: scoped-clean, no actionable findings,
  with the exact CI repair rated correct at 0.99.

[...]

PR #144169, opened 2026-09-10
fix(gateway): apply the session sharing predicate to session read-by-key RPCs

yetval:
## Summary

On a multi-identity Gateway, `sessions.files.list`, `sessions.files.get` and `sessions.diff` serve another person's `draft` session to any identified operator holding `operator.read`, as long as that caller can name the session key. `sessions.list` hides the same session from the same caller, and every sibling read-by-key surface refuses it.

Trigger condition: `gateway.roles` configured, a session whose `visibility` is `draft` and whose `createdActor` profile differs from the caller's `authenticatedUserProfile.profileId`, and a non-incognito session key.

What the caller receives today:

- `sessions.files.list` returns the touched-file inventory plus `root`, the absolute filesystem path of the other person's checkout, and with `path`/`search` an arbitrary directory listing under it.
- `sessions.files.get` returns the file body as a string plus its SHA-256, for any path under that root, up to the 256 KiB preview cap.
- `sessions.diff` returns the full checkout diff including uncommitted working-tree changes.

Scope note: keys are not handed out, since drafts are excluded from `sessions.list`. But keys are guessable (`agent:main:main` and channel-derived forms), and a session that was visible before its owner switched it to `draft` leaves its key already known.

Single-user Gateways are unaffected: their caller resolves to `GATEWAY_OWNER_PROFILE_ID`, `sharingIdentity` returns `undefined`, and the predicate is never built.

## Root cause

OpenClaw has exactly one predicate that decides whether an identified operator may see a given session by key, `createSessionListEntryFilter` / `createProfileSessionEntryFilter` in `src/gateway/session-sharing.ts`. Its final clause is the only place drafts are excluded for non-creators anywhere in the tree:

```ts
// the shared predicate, unchanged by this PR
  return (sessionKey, entry) =>
    entry.incognito !== true &&
    !isIncognitoSessionKey(sessionKey) &&
    (creatorMatches(entry.createdActor) ||
      (params.sessionCap !== "none" && resolveSessionVisibility(entry) !== "draft"));
```

The three handlers never bind `client`, so they can never call it. They are also absent from `SESSION_TARGET_FIELDS_BY_METHOD` (`src/gateway/session-method-policy.ts`), so the router's session fence resolves `targetRefs` to `undefined` and returns `{ error: null }` without ever reaching `authorizeSessionSharingTarget`.

The pre-dispatch fence still scrapes `sessionKey` through `resolveDirectSessionTargets`, which is why incognito-keyed sessions and roles with `sessions.others: "none"` stay protected. A `draft` session owned by a different profile is caught by neither path.

Note that the write sibling in the same file, `sessions.files.set`, does take `sessionMutationAuthorization` and use it. Only the reads are open.

## Fix

Apply the same predicate at the same boundary the merged exact-key read repair #136900 already established, so these three RPCs match `sessions.get` and `sessions.describe` exactly rather than inventing a stricter rule.

That boundary was a private helper in `sessions-read-by-key.ts`. This PR promotes it to the module that owns the predicate and routes all five read RPCs through the single definition, so the two cannot drift apart later:

```ts
// after: src/gateway/session-sharing.ts
export function createSessionReadVisibilityFilter(
  client: GatewayClient | null,
  cfg: OpenClawConfig,
): ReturnType<typeof createSessionListEntryFilter> {
  return hasOperatorBoundary(client, cfg)
    ? createSessionListEntryFilter({ client, cfg })
    : undefined;
}
```

`sessions-read-by-key.ts` drops its local copy and imports this one. The behavior there is unchanged, expression for expression.

The read admission for these three RPCs, plus the post-read recheck described in the next section, now live in one owner, `src/gateway/server-methods/session-read-visibility-boundary.ts`. Incognito is role independent, matching what `sessions.search` already does in `sessions-read.ts`; draft filtering is gated on the operator-role boundary:

```ts
// after: src/gateway/server-methods/session-read-visibility-boundary.ts
export function authorizeSessionReadTarget(params: {
  canonicalKey: string;
  cfg: OpenClawConfig;
  client: GatewayClient | null;
  entry: SessionEntry | undefined;
  sessionKey: string;
}): ErrorShape | null {
  const incognitoError = authorizeIncognitoSessionTarget({
    client: params.client,
    sessionKey: params.sessionKey,
    target: params.entry ? { canonicalKey: params.canonicalKey, entry: params.entry } : null,
  });
  if (incognitoError) {
    return incognitoError;
  }
  const entryFilter = createSessionReadVisibilityFilter(params.client, params.cfg);
  return params.entry && entryFilter && !entryFilter(params.canonicalKey, params.entry)
    ? hiddenSessionNotFound(params.sessionKey)
    : null;
}
```

`sessions.diff` and the single point both `sessions.files.*` handlers funnel through each call this once before any awaited work.

Why this is correct:

- `hiddenSessionNotFound` is the existing refusal for exactly this case and is already used by `chat-history-handler.ts`. The caller cannot distinguish a hidden session from an absent one, which is the intended behavior.
- The draft guard is conditional on `loaded.entry`, so a genuinely unknown session keeps its current response byte for byte: `sessions.files.list` still answers `ok: true` with an empty inventory, `sessions.files.get` still answers `session_file_not_found`, and `sessions.diff` still answers `unknown_session`.
- `createSessionReadVisibilityFilter` returns `undefined` for a Gateway admin, for an unidentified single-user caller, and for any identified caller on a Gateway with no `gateway.roles`, so all three keep their current access with no extra work.
- Admission runs before the transcript fold and before any workspace read, so a refused caller costs one store read rather than a full touched-files pass.

## Authority lifetime across the asynchronous read

Admission alone was not sufficient, and the previous revision stopped there. All three RPCs decided visibility, then awaited real work before answering: `sessions.files.*` await the SQLite touched-files fold and then a workspace listing or file read, and `sessions.diff` awaits `loadSessionDiff`, which shells out to git. A caller admitted while the session was `shared` still received the payload if, during that window, the session became a `draft`, its role was narrowed, or the same key was replaced by a different session. `AGENTS.md` states the rule this violated: revalidate after awaited work and immediately before the side effect.

The sibling already had the answer. `sessions.get` rechecks after `readRecentSessionMessagesWithStatsAsync` by reloading the runtime config and the canonical target, comparing identity, and re-applying the current predicate. This PR reuses that shape rather than inventing a second mechanism, in the same owner as the admission check:

```ts
// after: src/gateway/server-methods/session-read-visibility-boundary.ts
export function revalidateSessionReadTarget(params: {
  admitted: AdmittedSessionReadTarget;
  client: GatewayClient | null;
  context: GatewayRequestContext;
  requestedAgentId?: string;
  sessionKey: string;
}): ErrorShape | null {
  const cfg = params.context.getRuntimeConfig();
  const requestedAgent = resolveRequestedSessionAgentId(
    cfg,
    params.sessionKey,
    params.requestedAgentId,
  );
  if (!requestedAgent.ok || requestedAgent.agentId !== params.admitted.agentId) {
    return hiddenSessionNotFound(params.sessionKey);
  }
  const current = loadGatewaySessionEntryReadOnly(params.sessionKey, {
    ...(requestedAgent.agentId ? { agentId: requestedAgent.agentId } : {}),
  });
  if (
    current.canonicalKey !== params.admitted.canonicalKey ||
    current.storePath !== params.admitted.storePath ||
    current.entry?.sessionId !== params.admitted.sessionId
  ) {
    return hiddenSessionNotFound(params.sessionKey);
  }
  return authorizeSessionReadTarget({
    canonicalKey: current.canonicalKey,
    cfg,
    client: params.client,
    entry: current.entry,
    sessionKey: params.sessionKey,
  });
}
```

Placement: `sessions.files.list` and `sessions.files.get` call it after the workspace read resolves and before any response is built, so the too-large and not-found branches are behind it too. `sessions.diff` calls it after `loadSessionDiff` returns and before `respond`. The admitted snapshot carries the resolved agent id, the canonical key, the store path and the session id, so a re-resolved agent, a relocated store, or a replaced session id all read as stale.

A read whose authority is unchanged is unaffected: the recheck costs one more store read on the success path, and the payload is returned exactly as before.

## Compatibility

This keeps the resolution of the earlier blocking review finding: the predicate is still constructed only behind `hasOperatorBoundary`, so identified callers on Gateways with no `gateway.roles` are not newly restricted, at admission or at the recheck.

- **Identified caller, no `gateway.roles`.** Unchanged from current main and from v2026.9.3. `hasOperatorBoundary` is false, no filter is built, and the caller keeps listings, file bodies and diffs for a foreign draft. This is the mode #136900 deliberately preserved for `sessions.get` and `sessions.describe`, with an identified `no roles` row in its regression matrix. These three RPCs now honor the same row.
- **Identified caller, `gateway.roles` configured.** Foreign drafts are refused, which is the defect being fixed and matches the sibling exact-key reads.
- **Gateway admin and unidentified single-user caller.** Unchanged, the predicate returns `undefined` for both.
- **Incognito sessions.** Refused for identified non-admin callers regardless of roles, through `authorizeIncognitoSessionTarget`. This is the one deliberate tightening, and it is not new policy: it is the same role-independent check `sessions.search` already applies in `sessions-read.ts`. Retaining it was requested in review.
- **Same-key replacement and store relocation.** Previously these produced a payload derived from one session identity and labeled with the requested key. They now refuse rather than answer with a mixed-identity result.

The handler owns the read visibility check. That is not an opinion about layering, it is the arrangement the rest of the codebase already uses. Every other session read-by-key surface applies this predicate in its own handler:

1. `sessions.get` and `sessions.describe` at `src/gateway/server-methods/sessions-read-by-key.ts`, through the helper this PR promotes, including the post-read recheck this PR mirrors.
2. `chat.history` at `src/gateway/server-methods/chat-history-handler.ts`, responding `hiddenSessionNotFound(canonicalKey)`.
3. `controlUi.sessionPreview` at `src/gateway/server-methods/control-ui.ts`, whose comment states the invariant outright: "Hover previews must not reveal more than sessions.list: apply the same incognito/draft sharing predicate so a member cannot preview-by-key a session the sidebar hides from them."
4. `artifacts.*` at `src/gateway/server-methods/artifacts-session-resolution.ts`, same `operator.read`, same session-keyed shape.
5. `sessions.usage*` at `src/gateway/server-methods/usage.ts`.
6. `cron.*` at `src/gateway/server-methods/cron.ts`.
7. `projects.list` at `src/gateway/server-methods/projects.ts`.
8. HTTP `/sessions/:key/history` at `src/gateway/sessions-history-http.ts`, which synthesizes a client from request auth rather than skipping the check.

The policy table is the wrong lever. Adding these methods to `SESSION_TARGET_FIELDS_BY_METHOD` drives the mutation gate `authorizeSessionSharingTarget`, which returns participation-required errors. Applying a mutation-participation error to a read is semantically wrong and would break legitimate read access to `read-only` and `suggest` sessions. The codebase says so itself: `READ_ONLY_SESSION_TARGET_METHODS` exists precisely to exempt read methods from that gate. Reads carry their own visibility check in the handler, which is what this patch adds.

Deliberate follow-ups, not fixed here: `board.get` and `session.discussion.info` share the same defect but have different owners and different risk profiles, and each deserves its own change.

On framing: `docs/concepts/multi-user.md` says draft visibility "is a coordination feature, not a security boundary." I agree, and this is not filed as a security fix. The argument that survives that sentence is consistency. The same page states that "Catalog listings and progress updates recheck current session visibility for each recipient," and the `control-ui.ts` comment states the same rule for by-key reads. Eight surfaces honor it. These three do not, and they happen to carry the highest-value payload of the set: a checkout path, file contents, and an uncommitted diff. Severity P2.

## Verification

- `node scripts/run-vitest.mjs src/gateway/server-methods/sessions-files-visibility.test.ts` on the patched tree: 1 file passed, 12 tests passed.
- The same file with only the three revalidation call sites reverted to the previous revision: 3 failed, 9 passed. The three failures are exactly the new transitions, so they fail on the original defect.
- Regression coverage in that file: role-enabled refusal across all three methods, identified no-role compatibility, incognito refusal without roles, the creator, a Gateway admin, an unidentified single-user caller, the shared predicate itself, the unchanged response shape for a genuinely missing session, a control case where the session stays visible for the whole read, and the three concurrent transitions (visibility revoked to draft, caller role narrowed, same key replaced) applied to each of the three RPCs.
- `oxfmt --check` on the four changed files: clean. `git diff --check`: clean.
- Remaining lint, typecheck and full-suite gates are delegated to GitHub CI on this head.

## Real behavior proof
Behavior addressed: an identified `operator.read` caller who is not the session creator can read another person's `draft` session checkout path, file contents and uncommitted diff by naming its session key on a Gateway with `gateway.roles` configured, and a caller admitted while the session was shared can still receive that payload after the session becomes a draft, after its role is narrowed, or after the same key is replaced during the asynchronous read.
Real environment tested: the real `sessionsFilesHandlers` and `sessionsDiffHandlers` exports, the real `createSessionReadVisibilityFilter` / `createSessionListEntryFilter` / `createProfileSessionEntryFilter` predicate, the real `hasOperatorBoundary` role policy reading real `gateway.roles` config, the real `loadGatewaySessionEntryReadOnly` store reader, the real `replaceSessionEntry` writer, the real `persistSessionTranscriptTurn` transcript writer and the real `ensureProfileForEmail` profile store, driven on pristine main 1a2e8cddfcdc7b810b26a19776d66c826fb0dcdf, then on the previous revision 99eb58a76b766fd87e8a94ee184139ec1c64ed34, then on this head 3f2b6a2be4ef7ee28fc4c62c95ec312a23dabebc, with identical inputs and the same state database across all three runs. The caller was a real `GatewayClient` carrying a real profile id, the session checkout was a real git repository with one committed file and one uncommitted edit, and the concurrent change was applied to the real session store and the real runtime config while the handler promise was in flight, after the handler had already admitted the caller and yielded on its first await. Nothing on the authorization path was substituted; only the surrounding process was isolated, with `HOME` and `OPENCLAW_STATE_DIR` pointed at a scratch directory.
Exact steps or command run after this patch: create a git checkout holding `src/readme.md`, commit it, append an uncommitted line, write a `visibility: "shared"` session entry for `agent:main:proof-alpha` owned by another profile through `replaceSessionEntry`, seed one assistant tool-call turn through `persistSessionTranscriptTurn`, then for each of `sessions.files.list`, `sessions.files.get` and `sessions.diff` start the handler, apply one change while it is in flight (nothing, the session made a draft, the role narrowed to `sessions.others: "none"`, or the same key replaced by a different session id), await the response, and print it; repeat the whole matrix against pristine main, against the previous revision, and against this head.
Evidence after fix:
```
# BEFORE (pristine main 1a2e8cddfcdc)
[base 1a2e8cddfcdc] shared session, no change during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[base 1a2e8cddfcdc] shared session, no change during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[base 1a2e8cddfcdc] shared session, no change during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[base 1a2e8cddfcdc] foreign draft session, no change during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[base 1a2e8cddfcdc] foreign draft session, no change during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[base 1a2e8cddfcdc] foreign draft session, no change during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[base 1a2e8cddfcdc] shared session made a draft during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[base 1a2e8cddfcdc] shared session made a draft during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[base 1a2e8cddfcdc] shared session made a draft during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[base 1a2e8cddfcdc] caller role narrowed to none during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[base 1a2e8cddfcdc] caller role narrowed to none during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[base 1a2e8cddfcdc] caller role narrowed to none during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[base 1a2e8cddfcdc] same key replaced during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[base 1a2e8cddfcdc] same key replaced during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[base 1a2e8cddfcdc] same key replaced during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0

# MIDDLE (previous revision 99eb58a76b76, admission-only guard, identical inputs)
[admission-only 99eb58a76b76] shared session, no change during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[admission-only 99eb58a76b76] shared session, no change during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[admission-only 99eb58a76b76] shared session, no change during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[admission-only 99eb58a76b76] foreign draft session, no change during the read | sessions.files.list -> refused: Session "agent:main:proof-alpha" was not found.
[admission-only 99eb58a76b76] foreign draft session, no change during the read | sessions.files.get -> refused: Session "agent:main:proof-alpha" was not found.
[admission-only 99eb58a76b76] foreign draft session, no change during the read | sessions.diff -> refused: Session "agent:main:proof-alpha" was not found.
[admission-only 99eb58a76b76] shared session made a draft during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[admission-only 99eb58a76b76] shared session made a draft during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[admission-only 99eb58a76b76] shared session made a draft during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[admission-only 99eb58a76b76] caller role narrowed to none during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[admission-only 99eb58a76b76] caller role narrowed to none during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[admission-only 99eb58a76b76] caller role narrowed to none during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[admission-only 99eb58a76b76] same key replaced during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[admission-only 99eb58a76b76] same key replaced during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[admission-only 99eb58a76b76] same key replaced during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0

# AFTER (this patch, head 3f2b6a2be4ef, identical inputs)
[head 3f2b6a2be4ef] shared session, no change during the read | sessions.files.list -> served: root=/tmp/oc-144169-workspace files=[src/readme.md]
[head 3f2b6a2be4ef] shared session, no change during the read | sessions.files.get -> served: body="# alpha\nsecret roadmap line\n"
[head 3f2b6a2be4ef] shared session, no change during the read | sessions.diff -> served: changed=[src/readme.md] additions=1 deletions=0
[head 3f2b6a2be4ef] foreign draft session, no change during the read | sessions.files.list -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] foreign draft session, no change during the read | sessions.files.get -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] foreign draft session, no change during the read | sessions.diff -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] shared session made a draft during the read | sessions.files.list -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] shared session made a draft during the read | sessions.files.get -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] shared session made a draft during the read | sessions.diff -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] caller role narrowed to none during the read | sessions.files.list -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] caller role narrowed to none during the read | sessions.files.get -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] caller role narrowed to none during the read | sessions.diff -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] same key replaced during the read | sessions.files.list -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] same key replaced during the read | sessions.files.get -> refused: Session "agent:main:proof-alpha" was not found.
[head 3f2b6a2be4ef] same key replaced during the read | sessions.diff -> refused: Session "agent:main:proof-alpha" was not found.
```
Observed result after fix: on pristine main all fifteen runs hand the foreign caller the other person's absolute checkout path, her file body and her uncommitted working-tree diff. On the previous revision the static foreign draft is refused but the three changes applied while the read was in flight still hand over all three payloads. On this head the same fifteen runs, same inputs, same state database, answer `Session "agent:main:proof-alpha" was not found.` for the static foreign draft and for all nine in-flight changes, while the one shared session whose authority never moved is still served in full with `root`, the file body and the real git working-tree counts.
What was not tested: the proof drove the handler exports directly rather than a live WebSocket Gateway, so the router's pre-dispatch fence and scope gate were not exercised end to end. It covers only the local-workspace branch, not the cloud repository-workspace branch of `sessions.files.*`; that branch reaches the same recheck, but its remote inspection path was not driven. The incognito tightening is covered by regression coverage rather than by this transcript, which uses a non-incognito session. The concurrent change was applied from the same process rather than from a second Gateway client, so it lands deterministically between the handler's awaits rather than at an arbitrary interleaving. `board.get` and `session.discussion.info` are named above as follow-ups and are not covered. Full build and typecheck were not run locally on this head.

### DIFF
--- a/src/gateway/server-methods/session-read-visibility-boundary.ts
+++ b/src/gateway/server-methods/session-read-visibility-boundary.ts
@@ -1,0 +1,74 @@
+import type { ErrorShape } from "../../../packages/gateway-protocol/src/index.js";
+import type { SessionEntry } from "../../config/sessions.js";
+import type { OpenClawConfig } from "../../config/types.openclaw.js";
+import { resolveRequestedSessionAgentId } from "../session-request-agent.js";
+import {
+  authorizeIncognitoSessionTarget,
+  hiddenSessionNotFound,
+} from "../session-sharing-policy.js";
+import { createSessionReadVisibilityFilter } from "../session-sharing.js";
+import { loadGatewaySessionEntryReadOnly } from "../session-utils.js";
+import type { GatewayClient, GatewayRequestContext } from "./types.js";
+
+export type AdmittedSessionReadTarget = {
+  agentId: string;
+  canonicalKey: string;
+  sessionId: string | undefined;
+  storePath: string | undefined;
+};
+
+export function authorizeSessionReadTarget(params: {
+  canonicalKey: string;
+  cfg: OpenClawConfig;
+  client: GatewayClient | null;
+  entry: SessionEntry | undefined;
+  sessionKey: string;
+}): ErrorShape | null {
+  const incognitoError = authorizeIncognitoSessionTarget({
+    client: params.client,
+    sessionKey: params.sessionKey,
+    target: params.entry ? { canonicalKey: params.canonicalKey, entry: params.entry } : null,
+  });
+  if (incognitoError) {
+    return incognitoError;
+  }
+  const entryFilter = createSessionReadVisibilityFilter(params.client, params.cfg);
+  return params.entry && entryFilter && !entryFilter(params.canonicalKey, params.entry)
+    ? hiddenSessionNotFound(params.sessionKey)
+    : null;
+}
+
+export function revalidateSessionReadTarget(params: {
+  admitted: AdmittedSessionReadTarget;
+  client: GatewayClient | null;
+  context: GatewayRequestContext;
+  requestedAgentId?: string;
+  sessionKey: string;
+}): ErrorShape | null {
+  const cfg = params.context.getRuntimeConfig();
+  const requestedAgent = resolveRequestedSessionAgentId(
+    cfg,
+    params.sessionKey,
+    params.requestedAgentId,
+  );
+  if (!requestedAgent.ok || requestedAgent.agentId !== params.admitted.agentId) {
+    return hiddenSessionNotFound(params.sessionKey);
+  }
+  const current = loadGatewaySessionEntryReadOnly(params.sessionKey, {
+    ...(requestedAgent.agentId ? { agentId: requestedAgent.agentId } : {}),
+  });
+  if (
+    current.canonicalKey !== params.admitted.canonicalKey ||
+    current.storePath !== params.admitted.storePath ||
+    current.entry?.sessionId !== params.admitted.sessionId
+  ) {
+    return hiddenSessionNotFound(params.sessionKey);
+  }
+  return authorizeSessionReadTarget({
+    canonicalKey: current.canonicalKey,
+    cfg,
+    client: params.client,
+    entry: current.entry,
+    sessionKey: params.sessionKey,
+  });
+}
--- a/src/gateway/server-methods/sessions-diff.ts
+++ b/src/gateway/server-methods/sessions-diff.ts
@@ -13,6 +13,11 @@ import { normalizeAgentId, parseAgentSessionKey } from "../../routing/session-ke
 import { applySessionDiffBaseline, loadCheckoutDiff } from "../../sessions/session-diff.js";
 import { resolveRequestedSessionAgentId } from "../session-request-agent.js";
 import { loadGatewaySessionEntryReadOnly } from "../session-utils.js";
+import {
+  authorizeSessionReadTarget,
+  revalidateSessionReadTarget,
+  type AdmittedSessionReadTarget,
+} from "./session-read-visibility-boundary.js";
 import { loadRepositoryArtifactDiff } from "./session-repository-artifacts.js";
 import { resolveRepositoryWorkspaceAccess } from "./session-repository-workspace-access.js";
 import type { GatewayRequestContext, GatewayRequestHandlers } from "./types.js";
@@ -95,7 +100,7 @@ export async function loadSessionDiff(
 }
 
 export const sessionsDiffHandlers: GatewayRequestHandlers = {
-  "sessions.diff": async ({ params, respond, context }) => {
+  "sessions.diff": async ({ params, respond, context, client }) => {
     if (!assertValidParams(params, validateSessionsDiffParams, "sessions.diff", respond)) {
       return;
     }
@@ -120,15 +125,44 @@ export const sessionsDiffHandlers: GatewayRequestHandlers = {
       respond(false, undefined, requestedAgent.error);
       return;
     }
-    respond(
-      true,
-      await loadSessionDiff(
-        {
-          ...params,
-          ...(requestedAgent.agentId ? { agentId: requestedAgent.agentId } : {}),
-        },
-        context,
-      ),
+    const target = loadGatewaySessionEntryReadOnly(params.sessionKey, {
+      ...(requestedAgent.agentId ? { agentId: requestedAgent.agentId } : {}),
+    });
+    const admissionError = authorizeSessionReadTarget({
+      canonicalKey: target.canonicalKey,
+      cfg: context.getRuntimeConfig(),
+      client,
+      entry: target.entry,
+      sessionKey: params.sessionKey,
+    });
+    if (admissionError) {
+      respond(false, undefined, admissionError);
+      return;
+    }
+    const admitted: AdmittedSessionReadTarget = {
+      agentId: requestedAgent.agentId,
+      canonicalKey: target.canonicalKey,
+      sessionId: target.entry?.sessionId,
+      storePath: target.storePath,
+    };
+    const diff = await loadSessionDiff(
+      {
+        ...params,
+        ...(requestedAgent.agentId ? { agentId: requestedAgent.agentId } : {}),
+      },
+      context,
     );
+    const staleError = revalidateSessionReadTarget({
+      admitted,
+      client,
+      context,
+      ...(params.agentId ? { requestedAgentId: params.agentId } : {}),
+      sessionKey: params.sessionKey,
+    });
+    if (staleError) {
+      respond(false, undefined, staleError);
+      return;
+    }
+    respond(true, diff);
   },
 };
--- a/src/gateway/server-methods/sessions-files-visibility.test.ts
+++ b/src/gateway/server-methods/sessions-files-visibility.test.ts
@@ -1,0 +1,413 @@
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { closeOpenClawStateDatabaseForTest } from "../../state/openclaw-state-db.js";
+import { ensureProfileForEmail } from "../../state/user-profiles.js";
+import { createTempHomeEnv, type TempHomeEnv } from "../../test-utils/temp-home.js";
+import { prepareSessionCreatorProfile } from "../session-creator.js";
+import { createProfileSessionEntryFilter } from "../session-sharing.js";
+import { sessionsDiffHandlers } from "./sessions-diff.js";
+import { sessionsFilesHandlers } from "./sessions-files.js";
+import {
+  assistantToolCall,
+  createVisibleMessagesMock,
+  prepareSessionFilesTest,
+  removeWorkspaceFixture,
+  visibleMessageEvent,
+} from "./sessions-files.test-support.js";
+import type { GatewayRequestHandlers, RespondFn } from "./types.js";
+
+const hoisted = vi.hoisted(() => ({
+  execOpenPath: vi.fn(),
+  loadSessionEntry: vi.fn(),
+  resolveAgentWorkspaceDir: vi.fn(),
+  resolveDefaultAgentId: vi.fn(),
+  readSessionTranscriptVisibleMessageDeltaCore: vi.fn(),
+}));
+
+vi.mock("./open-path.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("./open-path.js")>()),
+  execOpenPath: hoisted.execOpenPath,
+}));
+
+vi.mock("../../agents/agent-scope.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../../agents/agent-scope.js")>()),
+  resolveAgentWorkspaceDir: hoisted.resolveAgentWorkspaceDir,
+  resolveDefaultAgentId: hoisted.resolveDefaultAgentId,
+}));
+
+vi.mock("../session-utils.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../session-utils.js")>()),
+  loadSessionEntry: hoisted.loadSessionEntry,
+  loadGatewaySessionEntryReadOnly: hoisted.loadSessionEntry,
+}));
+
+vi.mock("../session-transcript-readers.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../session-transcript-readers.js")>()),
+  readSessionTranscriptVisibleMessageDeltaCore:
+    hoisted.readSessionTranscriptVisibleMessageDeltaCore,
+}));
+
+const mockVisibleMessages = createVisibleMessagesMock(
+  hoisted.readSessionTranscriptVisibleMessageDeltaCore,
+);
+
+const SESSION_KEY = "agent:main:main";
+
+const cfg = { agents: { list: [{ id: "main", default: true }] } } as never;
+const rolesCfg = {
+  agents: { list: [{ id: "main", default: true }] },
+  gateway: {
+    roles: {
+      default: "limited",
+      definitions: {
+        limited: { sessions: { others: "view" }, agents: "*", scopes: ["operator.read"] },
+      },
+    },
+  },
+} as never;
+const restrictedRolesCfg = {
+  agents: { list: [{ id: "main", default: true }] },
+  gateway: {
+    roles: {
+      default: "limited",
+      definitions: {
+        limited: { sessions: { others: "none" }, agents: "*", scopes: ["operator.read"] },
+      },
+    },
+  },
+} as never;
+const INCOGNITO_SESSION_KEY = "agent:main:dashboard:incognito-probe";
+
+function operatorClient(profileId: string | undefined, scopes: string[]) {
+  return {
+    connId: `conn-${profileId ?? "anonymous"}`,
+    connect: {
+      role: "operator",
+      scopes,
+      client: { id: "openclaw-control-ui", mode: "operator", version: "1", platform: "linux" },
+    },
+    ...(profileId ? { authenticatedUserProfile: { profileId } } : {}),
+  } as never;
+}
+
+let ALICE: string;
+let BOB: string;
+let bobClient: unknown;
+let aliceClient: unknown;
+let adminClient: unknown;
+const unidentifiedClient = operatorClient(undefined, ["operator.read"]);
+
+async function invoke(
+  handlers: GatewayRequestHandlers,
+  method: string,
+  params: Record<string, unknown>,
+  client: unknown,
+  getRuntimeConfig: () => unknown = () => cfg,
+) {
+  const calls: { ok: boolean; payload?: unknown; error?: unknown }[] = [];
+  const respond: RespondFn = (ok, payload, error) => {
+    calls.push({ ok, payload, error });
+  };
+  await handlers[method]?.({
+    req: { type: "req", id: method, method, params: {} },
+    params,
+    client,
+    isWebchatConnect: () => false,
+    respond,
+    context: { getRuntimeConfig } as never,
+  } as never);
+  expect(calls).toHaveLength(1);
+  return calls[0] as { ok: boolean; payload?: any; error?: any };
+}
+
+function readSessionAs(
+  client: unknown,
+  getRuntimeConfig: () => unknown = () => cfg,
+  sessionKey: string = SESSION_KEY,
+) {
+  return {
+    list: () =>
+      invoke(
+        sessionsFilesHandlers,
+        "sessions.files.list",
+        { sessionKey },
+        client,
+        getRuntimeConfig,
+      ),
+    get: () =>
+      invoke(
+        sessionsFilesHandlers,
+        "sessions.files.get",
+        { sessionKey, path: "src/readme.md" },
+        client,
+        getRuntimeConfig,
+      ),
+    diff: () =>
+      invoke(sessionsDiffHandlers, "sessions.diff", { sessionKey }, client, getRuntimeConfig),
+  };
+}
+
+describe("session read-by-key visibility for sessions.files.* and sessions.diff", () => {
+  let workspaceRoot: string;
+  let home: TempHomeEnv;
+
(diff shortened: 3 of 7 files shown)


yetval:
Rebased onto current main and addressed the blocking finding. New head pinned in the refreshed body.

**[P1] Preserve identified no-role callers' existing read access.** Fixed. The guard no longer calls the predicate unconditionally. All three RPCs now go through `createSessionReadVisibilityFilter(client, cfg)`, which returns `undefined` unless `hasOperatorBoundary(client, cfg)` is true. That is the same expression #136900 established for the exact-key reads, so an identified non-admin caller with no `gateway.roles` keeps foreign-draft listings, file bodies and diffs exactly as on current main and v2026.9.3.

Rather than copy the condition, I promoted it out of `sessions-read-by-key.ts` into `session-sharing.ts`, next to the predicate it wraps, and pointed `sessions.describe`, `sessions.get`, `sessions.files.list`, `sessions.files.get` and `sessions.diff` at the one definition. `sessions-read-by-key.ts` behavior is unchanged, expression for expression, and the two paths can no longer drift.

**[P1] Resolve merge risk (compatibility).** Answered by the above. The no-role mode is preserved, so the upgrade no longer removes access from existing identified callers. The one deliberate tightening that remains is incognito, which you asked to retain: it is applied through `canAccessIncognitoSession`, the same role-independent helper `sessions.search` already uses in `sessions-read.ts`, not a new rule invented here. Incognito behavior for admins and for the single-user owner is unchanged.

**[P2] Complete next step.** Done in all three RPCs, with coverage distinguishing the two modes:

- `refuses sessions.files.list, sessions.files.get and sessions.diff for a non-creator under operator roles`
- `preserves foreign draft reads for an identified caller when no operator roles exist`
- `still refuses an incognito session for an identified caller without operator roles`
- plus the existing creator, admin, unidentified-caller and missing-session cases

The role-enabled cases use real profile rows from `ensureProfileForEmail` against an isolated state database and the shipped `gateway.roles` shape, so the boundary resolves through the real role policy rather than a stubbed cap.

**Rank-up move: show both modes in the real-handler evidence.** Done. The refreshed proof drives one run over the same session data and prints Mode A (`gateway.roles` configured) and Mode B (no roles) side by side, on pristine main and on the patched tree. Bob loses all three payloads only in Mode A; in Mode B he keeps them, which is the compatibility claim above shown rather than asserted.

**Incognito, on the cost of the guard.** It is applied through `authorizeIncognitoSessionTarget` against the session entry each handler has already loaded, rather than re-resolving the target through the store. That keeps the refusal shapes identical to the shipped helper, including the pending-profile case, and adds no extra store read per request.

New head `99eb58a76b766fd87e8a94ee184139ec1c64ed34`, rebased onto `c38897f5544081974bc13632c2f1a17337b52ba7`. The body's proof block is regenerated against that exact head. Remaining lint, typecheck and full-suite gates are left to CI on this head.

yetval:
Rebased onto current main and addressed the three remaining items. New head `3f2b6a2be4ef7ee28fc4c62c95ec312a23dabebc`, base `1a2e8cddfcdc7b810b26a19776d66c826fb0dcdf`. The rebase absorbed 549 commits; none of them touched the five production files, and main still carries none of the three read guards.

**Merge risk (P1), authority lifetime: fixed.** You were right that admission was the whole guard and that the handlers answered without rechecking. All three RPCs now revalidate at the final response boundary, using the `sessions.get` pattern you pointed at rather than a second mechanism: reload the runtime config, re-resolve the requested agent, reload the canonical target, compare agent id, canonical key, store path and session id, then re-apply the current visibility predicate. Admission and the recheck live in one owner, `src/gateway/server-methods/session-read-visibility-boundary.ts`. Call sites: `sessions.files.list` and `sessions.files.get` after the workspace read resolves and before any response is built, so the not-found and too-large branches are behind it as well; `sessions.diff` after `loadSessionDiff` returns and before `respond`.

**Real behavior proof: added, three-way.** The refreshed body carries a real-handler transcript driven on pristine main `1a2e8cddfcdc`, then on the previous revision `99eb58a76b76`, then on this head, with identical inputs and the same state database. Real `replaceSessionEntry`, real `persistSessionTranscriptTurn`, real profile store, real `gateway.roles` policy, real git checkout. The change is applied to the real store or the real runtime config while the handler promise is in flight, after admission has already passed. Result: main serves all fifteen runs; the previous revision refuses the static foreign draft but still serves all nine in-flight changes; this head refuses all twelve and still serves the one shared session whose authority never moved.

**No-role compatibility: preserved.** `hasOperatorBoundary` still gates before the sharing filter is constructed, at admission and at the recheck, so an identified caller on a Gateway with no `gateway.roles` keeps its existing access. That row stays in the regression matrix.

Regression coverage in `sessions-files-visibility.test.ts` now mirrors the sibling matrix: visibility revoked to draft, caller role narrowed, and same-key replacement, each applied to all three RPCs, plus a control case where the session stays visible for the whole read. Reverting only the three revalidation call sites fails exactly those three cases and nothing else.

Still an owner decision, unchanged from the previous cycle: whether draft visibility on these payloads is worth enforcing at all, given that `docs/concepts/multi-user.md` calls it a coordination feature rather than a security boundary. The PR argues consistency with the eight surfaces that already recheck, not a security claim. Also unchanged: `board.get` and `session.discussion.info` share the defect and are left as separate follow-ups under their own owners.

Not covered, stated in the proof: no live WebSocket Gateway, so the router pre-dispatch fence was not driven end to end; the cloud repository-workspace branch of `sessions.files.*` reaches the same recheck but its remote inspection path was not driven; the in-flight change is applied from the same process, so it lands between the handler's awaits deterministically rather than at an arbitrary interleaving.

PR #144172, opened 2026-09-10
fix(mattermost): redact active credentials in errors

### LINKED ISSUE #144173: [Bug]: Mattermost API error text reflects the bot token when a server echoes request headers
### Bug type

Security issue (credential exposure in diagnostics)

### Summary

Mattermost REST requests send `Authorization: Bearer <botToken>`, but `readMattermostError` returns the untrusted error-response text verbatim. A self-hosted or misbehaving Mattermost server that echoes request headers in its error body gets the bot token reflected into the surfaced error message, which reaches the session transcript, logs, and channel error surfaces.

This is the same reflected-credential class already fixed for sibling channels — most recently Discord non-JSON errors (#119536) — but Mattermost was missed in that pass.

### Affected code

- `extensions/mattermost/src/mattermost/client.ts:149` — `readMattermostError` reads the error body with `readResponseTextLimited` and returns it raw (non-JSON text), as `data.message` (JSON), or `JSON.stringify(data)` without redaction.
- `extensions/mattermost/src/mattermost/client.ts:251` — the request error path interpolates the detail into `throw new Error(\`Mattermost API ${res.status} ${res.statusText}: ${detail}\`)`.
- `extensions/mattermost/src/mattermost/client.ts:715` — the upload error path does the same.
- `extensions/mattermost/src/mattermost/probe.ts:57` — the preflight path returns the detail as `error`.

### Steps to reproduce

1. Point a Mattermost account at a server that echoes the request `Authorization` header back in its error body (self-hosted or misbehaving proxy).
2. Trigger any failing Mattermost API call (message send, upload, or probe).
3. Observe the bot token present in the surfaced error text.

Deterministic, no live Mattermost needed: feeding such a response to `readMattermostError` returns the token unchanged.

### Expected behavior

Reflected credential material in error text should be redacted before it enters any surfaced error message, matching the sibling channel/proxy fixes.

### Actual behavior

The token is returned verbatim by `readMattermostError` and interpolated into the thrown error message.

### Suggested fix

Redact the response text once at the read boundary with the existing shared `redactToolPayloadText` helper (`openclaw/plugin-sdk/logging-core`), the same helper already applied in Google Chat (`extensions/googlechat/src/api.ts`), Twilio SMS (`extensions/sms/src/twilio.ts`), GitHub Copilot, LM Studio, WhatsApp, and Slack. Fixing the boundary covers all three call sites without touching each caller.

### Evidence

Repro at head `e98a5c15f3b`:

```text
non-JSON body  -> "Request failed\nAuthorization: Bearer mm-bot-token-…"   (token present)
JSON message   -> "auth failed for Bearer mm-bot-token-…"                  (token present)
```

Proposed fix (PR #144172) redacts both to `mm-bot…WXYZ` and adds two regression tests; full `client.test.ts` passes (56/56).

wangmiao0668000666:
Closes #144173. Related historical work: #123223.

## What Problem This Solves

Mattermost error responses can reflect the active bot credential into Gateway probe results, send diagnostics, logs, and persisted delivery errors. Current main also lets an object or array in a JSON `message` escape as a non-string probe error.

## Why This Change Was Made

Keep one redaction boundary at `readMattermostError`. Decode the bounded JSON first, select a string message or serialize the remaining value, then use the existing request-header-aware SDK redactor. Request, upload, and probe callers supply the actual headers they sent. This masks opaque bare credentials as well as labeled bearer values without relying on ambient secret registration.

Preserve the bounded reader's truncation fact so a credential cut by the 8 KiB limit is also masked. The existing idle timeout, network guard, redirect handling, request deadlines, status context, successful responses, and delivery policy stay intact. The assertion ratchet decreases one entry from 9 to 8 after removing the unchecked JSON cast.

This retains the contributor's parse-before-redaction correction and ancestry. It also closes the bare-token and malformed-message gaps found during the independent security review. No public configuration, schema, dependency, or secret registry is added.

## Evidence

Exact baseline: `ce405b8d252e`. Exact candidate: `349443da96d4`.

Controlled loopback HTTP/WebSocket proof uses a synthetic Mattermost account, the normal private-network opt-in, the built Gateway, and authenticated public `channels.status` and `send` calls. The server reflects the Authorization value it actually receives. This exercises production Gateway/plugin/client code; it is not a third-party Mattermost deployment test.

| Observed boundary | Baseline | Candidate |
| --- | --- | --- |
| Probe error containing a reflected header | Full credential returned | Credential replaced with `***`; status and context retained |
| Bare credential in send errors | Full credential in public error, Gateway log, and delivery diagnostics | Credential masked at the shared reader |
| Object/array JSON message | Non-string probe error containing the credential | Useful serialized string with masked credential |
| Credential cut at the body limit | Retained credential prefix | Retained prefix masked; body remains bounded |
| Successful text and file sends | Valid receipts | Valid receipts |

The public matrix covers plain text, JSON messages, Unicode-escaped JSON, mismatched content type, serialized-object fallback, object/array messages, bare values, and truncated bodies. Upload errors use the same reader. Separate real transport controls confirm private-network denial makes no server request, cross-origin redirects withhold Authorization, and a stalled probe returns its configured timeout.

A complete scan of captured candidate Gateway logs and five SQLite files found no full synthetic credential in diagnostics. Baseline evidence identifies four leaking log lines and eight delivery-error cells. No model turn was exercised; the observed downstream surfaces are Gateway results/logs and persisted delivery diagnostics. This repair prevents newly produced diagnostics from leaking; it does not rewrite historical records.

Native validation: 76 focused client/probe/redaction tests passed, with formatting, scoped lint, ratchets, and the repository hook. Ten new cases fail on the incoming implementation for credential, output-type, and clipped-prefix defects. The initial zero-case transfer and other setup failures remain in the private history and are not counted as proof.

Fresh independent source/security review is clean. Fresh source-blind runtime acceptance completed all four public and network-control runs. It also found no full credential or clipped prefix in captured candidate diagnostics, logs, or stored delivery errors. Full private captures, fixture sources, build identities, source custody and failure history are retained; public text omits tokens and private execution metadata.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -698,7 +698,7 @@ extensions/matrix/src/test-runtime.ts	10
 extensions/matrix/src/tool-actions.ts	1
 extensions/mattermost/src/channel.ts	8
 extensions/mattermost/src/gateway-auth-bypass.ts	3
-extensions/mattermost/src/mattermost/client.ts	9
+extensions/mattermost/src/mattermost/client.ts	8
 extensions/mattermost/src/mattermost/directory.ts	2
 extensions/mattermost/src/mattermost/interactions.ts	4
 extensions/mattermost/src/mattermost/model-picker.ts	1
--- a/extensions/mattermost/src/mattermost/client.error-redaction.test.ts
+++ b/extensions/mattermost/src/mattermost/client.error-redaction.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it } from "vitest";
+import { createMattermostClient, readMattermostError, uploadMattermostFile } from "./client.js";
+
+describe("Mattermost reflected credential diagnostics", () => {
+  it.each([
+    {
+      name: "bare text",
+      contentType: "text/plain",
+      body: "upstream rejected abcdefghijklmnopqrstuvwxyz",
+      expected: "upstream rejected ***",
+    },
+    {
+      name: "bare JSON message",
+      contentType: "application/json",
+      body: '{"message":"upstream rejected abcdefghijklmnopqrstuvwxyz"}',
+      expected: "upstream rejected ***",
+    },
+    {
+      name: "serialized object fallback",
+      contentType: "application/json",
+      body: '{"context":"retry later","echoed":"abcdefghijklmnopqrstuvwxyz"}',
+      expected: '{"context":"retry later","echoed":"***"}',
+    },
+    {
+      name: "object-valued message",
+      contentType: "application/json",
+      body: '{"message":{"context":"retry later","echoed":"abcdefghijklmnopqrstuvwxyz"}}',
+      expected: '{"message":{"context":"retry later","echoed":"***"}}',
+    },
+    {
+      name: "array-valued message",
+      contentType: "application/json",
+      body: '{"message":["retry later","abcdefghijklmnopqrstuvwxyz"]}',
+      expected: '{"message":["retry later","***"]}',
+    },
+    {
+      name: "non-string message without credentials",
+      contentType: "application/json",
+      body: '{"message":503,"context":"retry later"}',
+      expected: '{"message":503,"context":"retry later"}',
+    },
+  ])("redacts the active credential and preserves $name diagnostics", async (scenario) => {
+    const response = new Response(scenario.body, {
+      status: 503,
+      headers: { "content-type": scenario.contentType },
+    });
+
+    const detail = await readMattermostError(response, {
+      Authorization: "Bearer abcdefghijklmnopqrstuvwxyz",
+    });
+
+    expect(detail).toBe(scenario.expected);
+  });
+
+  it("redacts a credential cut by the error-body limit and cancels unread data", async () => {
+    const safePrefix = "upstream diagnostic " + ".".repeat(8192 - 20 - 12);
+    let canceled = false;
+    const response = new Response(
+      new ReadableStream<Uint8Array>({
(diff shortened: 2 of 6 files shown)


obviyus:
Thanks @wangmiao0668000666!

**Proof:** Built Gateway probe, send and upload flows against a synthetic loopback Mattermost transport. Independent runtime acceptance and source/security review passed.

- **Bug:** Reflected active credentials could reach probe results, selected send errors, logs and saved delivery diagnostics. Malformed JSON messages could also return a non-string error.
- **Fix:** Decode the bounded body, normalize the diagnostic to a string, then redact once using the actual request headers. This retains the contributor's decode-order correction and covers bare values and credentials clipped by the body limit.

| Before | After |
| --- | --- |
| Public probes and bare-token send diagnostics expose reflected credentials | Credentials and clipped prefixes are masked; status and context remain |
| Object/array messages can escape as non-string errors | Useful serialized, redacted strings |
| Successful text/upload delivery and existing network controls | Preserved in real Gateway controls |

Validation: 76 focused native tests passed; ten regression cases failed on the incoming implementation for the intended defects. The proof covers Gateway output, logs and stored delivery errors. It does not claim a third-party deployment or model-turn test.

wangmiao0668000666:
Thanks @obviyus for the review, the decode-order tightening, and the independent verification — and for taking it over the line. Appreciate you confirming the after-fix boundary behavior (Gateway output, logs, stored delivery errors) and keeping the network controls intact. Happy to help with any follow-ups in this plugin.

PR #143908, opened 2026-09-10
fix(state): tolerate transient locks during read-only access

steipete:
Related: #143832, #142522, #143661

## What Problem This Solves

Fixes an issue where users reading shared state, including plugin-discovery state used by Doctor, could receive `database is locked` immediately when another connection briefly held an exclusive database lock.

## Why This Change Was Made

Legacy catalog admission performs a schema query before the caller reads any rows. Install the existing 5,000 ms SQLite busy timeout before that admission query so the entire fresh read follows the established timeout. Read-only access, malformed legacy-index handling, source-handle exclusion, and cleanup retain their existing behavior.

## User Impact

Fresh shared-state readers can wait for transient lock contention within the existing timeout instead of failing before that timeout is active.

## Evidence

- The new cross-process regression failed on unchanged main production code with `SQLITE_BUSY` at the same legacy catalog query as the CI failure. Its writer deliberately uses a rollback-journal exclusive transaction; the original CI writer remains unidentified.
- The complete read-only and exclusion files passed all 25 cases after the one-line reorder, including malformed-index admission and source preservation. After a test-only nullable-pipe type correction, the focused contention regression passed again.
- Independent P2 review of the final source found no actionable findings.
- Final changed checks passed, including core/test types, lint, schema and architecture guards.
- Standalone Linux API proof passed on Blacksmith Testbox ([run 34459489877](https://github.com/openclaw/openclaw/actions/runs/34459489877), Node 24.19.0 / SQLite 3.53.3): `readConfigMachineState` returned committed state after the synthetic exclusive lock released; integrity, read-only enforcement, unchanged source bytes, natural child exit, and temporary-state cleanup passed. The exact production source hash was verified. The lease is independently confirmed completed; the backing Actions session may show canceled because one-shot Testbox cleanup owns its lifecycle. A post-success portal-sync warning did not change the successful command or completed lease.

[...]

PR #143838, opened 2026-09-10
feat(gateway): prepare cloud workers on demand

steipete:
## Problem

Operators cannot prepare a cloud worker for a profile and local repository without first dispatching a session. Ready-worker demand currently comes from session activity.

## Solution

Add the admin-only `environments.prepare({ profileId, projectPath })` RPC. It validates the local Git checkout and pins HEAD, authorizes the committed setup recipe, atomically reuses matching inventory or admits a build, and schedules the existing preparation pool. List and status summaries expose preparation purpose and key; `environments.destroy` cancels an in-flight build while retaining provider settlement and cleanup ownership.

Persist purpose through nullable `worker_environments.preparation_purpose TEXT`, using the existing compatible additive migration. Legacy rows remain reserves. Unfinished builds bypass the per-project ready target, including reused reserves promoted to builds, while retaining their original expiry and obeying the global cap. Ready builds return to ordinary reserve retention. A zero global cap rejects admission before reuse.

Pool retention uses the inventory behind each pass's source snapshot while rereading existing rows live. A build admitted during an awaited retention operation is handled by the already-requested next pass, so stale eligibility cannot cancel successful admission. Current store queries still enforce capacity; expiry, cancellation, and provider settlement remain live.

## Impact

An operator can build an eligible project's prepared worker without creating a session. Admission returns its environment ID, preparation key, and whether it reused inventory. Validation and capacity errors remain typed. Existing reserve targets, idle timeouts, provider capture behavior, and global capacity accounting remain intact.

This adds no UI, CLI command, provider changes, version bump, or changelog edit. Build admission has a separate module to keep lifecycle ownership and file-size limits clear; the existing profile validator is shared, with its obsolete assertion-baseline allowance removed.

## Evidence

The [initial exact-head CI attempt](https://github.com/openclaw/openclaw/actions/runs/34455365820/job/102800854096) hit the pre-existing Doctor/plugin-index SQLite lock flake in `channel-plugin-blockers.test.ts`. The failing path is unchanged by this PR and is recorded for separate repair. The first authorized failed-job rerun passed. [Exact-head CI attempt 2](https://github.com/openclaw/openclaw/actions/runs/34455365820/attempts/2) completed successfully on `8ceaeb8f9e60a241badc1c283503c7783d68bc15`, including the required CI gate. No test or locking behavior was changed to bypass the failure.

[...]

PR #144175, opened 2026-09-10
fix(state): install read-only SQLite lock policy at open

steipete:
## Problem

CI shard `agentic-command-support` failed in
[run 34455365820](https://github.com/openclaw/openclaw/actions/runs/34455365820)
on a test that touches no SQLite code:

```
FAIL src/commands/doctor/shared/channel-plugin-blockers.test.ts > accepts an available co-owner for the same package env trigger
Error: database is locked
 ❯ src/state/openclaw-state-db-dangling-workshop-index.ts:36:8
 ❯ read src/state/openclaw-state-db-readonly.ts:124:34
Serialized Error: { code: 'ERR_SQLITE_ERROR', errcode: 5, errstr: 'database is locked' }
```

It failed in 3 ms, not after the established 5000 ms busy timeout, because the
connection had no busy handler at all when that statement ran.

`node:sqlite` opens a connection with busy timeout `0`:

```
new DatabaseSync(':memory:').prepare('PRAGMA busy_timeout').get()  // { timeout: 0 }
```

A connection therefore has no busy handler until one is installed, and both
read-only openers of the shared state and agent databases installed it with a
`PRAGMA busy_timeout` *after* opening. Every statement issued before that PRAGMA
ran unprotected: the state reader's very first statement is the legacy-catalog
admission's `sqlite_schema` read, so any transient lock failed it outright.

#143908 fixed the failing ordering by moving the PRAGMA ahead of that admission
call. The ordering requirement itself remained: lock policy still depended on
which statement happened to come first inside the opener body, and inserting a
statement above the PRAGMA silently disables the busy handler again.

## Solution

Pass `timeout` to the `DatabaseSync` constructor at both read-only openers, so
the connection is born with its lock policy, and delete the now-redundant
post-open PRAGMA. `readSqliteBusyTimeout` already reads both the `busy_timeout`
and `timeout` result column names, so `runWithSqliteBusyTimeout` keeps observing
and restoring the value exactly as before.

## Impact

A fresh read-only state or agent read no longer depends on statement order for
its lock policy, and cannot regress into failing immediately on a transient
lock. No configuration, schema, or public surface changes.

## Evidence

The reported failure reproduces deterministically from the ordering alone.
Restoring the pre-#143908 order (`openDanglingWorkshopIndexReadAdmission` before
the PRAGMA) and running the regression test that #143908 added:

```
× src/state/openclaw-state-db-readonly.test.ts > waits for a transient database lock
    before a fresh read-only schema inspection 71ms
  → database is locked
Serialized Error: { code: 'ERR_SQLITE_ERROR', errcode: 5, errstr: 'database is locked' }
Test Files  1 failed (1)
```

Same signature and same immediate failure as CI. With this change:

- `src/state/openclaw-state-db-readonly.test.ts` — 21 passed
- `src/state/openclaw-agent-db.test.ts`, `src/state/openclaw-agent-db-open-authority.test.ts` — 144 passed, 4 skipped
- `node scripts/run-vitest.mjs src/commands/doctor/shared/channel-plugin-blockers.test.ts src/state/` — 1137 passed, 6 skipped, 3 failed
- `node scripts/check-changed.mjs` — exit 0
- Independent Codex review (P2) — `scoped-clean`

The three failures are `src/state/local-onboarding-state.test.ts` (`existing
managed handoff lease is incompatible`). They reproduce identically on a clean
checkout of `main` with no local modifications and are unrelated to this change;
tracked separately.

The competing writer that held the lock in that CI process is still unidentified
— the shard's Vitest worker threads each get their own temp `HOME` and state
database, so the contention is process-local. It no longer matters for
correctness here: a transient lock is now waited out at either opener.

Fixes the flake seen in #143838's CI.

### DIFF
--- a/src/state/openclaw-agent-db-readonly.ts
+++ b/src/state/openclaw-agent-db-readonly.ts
@@ -81,8 +81,11 @@ export function openOpenClawAgentDatabaseReadOnly(
   if (!fs.existsSync(pathname)) {
     return { found: false, reason: "database-missing" };
   }
+  // Lock policy belongs to the open: node:sqlite has no busy handler until one
+  // is set, so a later PRAGMA leaves every earlier statement unprotected.
   const db = openNodeSqliteDatabase(pathname, {
     readOnly: true,
+    timeout: OPENCLAW_SQLITE_BUSY_TIMEOUT_MS,
     ...(behavior.allowExtension ? { allowExtension: true } : {}),
   });
   let closed = false;
@@ -96,7 +99,6 @@ export function openOpenClawAgentDatabaseReadOnly(
   };
   try {
     registerOpenClawAgentDatabaseIdentity(db);
-    db.exec(`PRAGMA busy_timeout = ${OPENCLAW_SQLITE_BUSY_TIMEOUT_MS};`);
     const userVersion = assertSupportedAgentSchemaVersion(db, pathname);
     assertCanonicalAgentPersistenceVersion(db, pathname, userVersion);
     const schemaMeta = readExistingAgentSchemaMeta(db);
--- a/src/state/openclaw-state-db-readonly.ts
+++ b/src/state/openclaw-state-db-readonly.ts
@@ -118,10 +118,15 @@ function withOpenClawStateReadOnlyLocation<T>(
   location: string,
 ): T {
   const read = () => {
-    const db = openNodeSqliteDatabase(location, { readOnly: true });
+    // node:sqlite opens without a busy handler, so a timeout installed by a
+    // later PRAGMA leaves the first statement — legacy catalog admission's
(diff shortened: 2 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a62ca8dc5ce0d7ace881d503cc5b7b631fab6355](https://github.com/openclaw/openclaw/pull/144175/commits/a62ca8dc5ce0d7ace881d503cc5b7b631fab6355)
- Landed commit: [a0f2309c3adb89bc2bec2f64696eb320318bb2cd](https://github.com/openclaw/openclaw/commit/a0f2309c3adb89bc2bec2f64696eb320318bb2cd)

PR #144180, opened 2026-09-10
fix(models): keep catalog capabilities with their routes

steipete:
Related: #130706, #142100.

## What Problem This Solves

The model-selection catalog duplicated the shared metadata merger. When a configured API or endpoint changed, that copy could retain capabilities and a prepared thinking policy from the previous route. It also failed to apply an explicit thinking-level map.

## Why This Change Was Made

Use the existing catalog overlay owner, including its route-change checks, and clear the old thinking-policy fields with the other route-bound metadata. Keep the captured catalog row's original transport until final route projection. That preserves rejection of incompatible donors, including rows missing an API or endpoint.

This removes eight production lines across two files. The documentation describes the resulting route and thinking-map behavior. Configuration, storage, and plugin APIs are unchanged.

## User Impact

Model selection no longer carries the previous route's capabilities and thinking policy across a configured route change. Explicit thinking maps are retained. Compatible routes keep their existing capability and compatibility precedence.

## Evidence

- The original source failed five intended cases; seven controls passed, including missing-route fields, explicit pins, and provider defaults.
- The candidate passed all 355 focused tests across eight files, including the complete twelve-case route/selection/thinking table.
- The full changed-file gate, full documentation checks, and independent P2 review passed on the committed source.
- One normal full build of `f1bc414c6ab7bcd8da7546fe75a22ce2be3d7a98` passed, including SDK declarations. All twelve cases passed through the emitted production exports. The source-import rejection control also passed.
- No checkout TypeScript or fetch calls occurred. All 11,934 build files across seventeen roots remained unchanged, and all owned proof processes exited. This is compiled selection/thinking-boundary proof, not a hosted-provider request.
- Exact-head CI remains required before landing.

An earlier attempt copied configured transport onto the captured row and allowed missing-field rows to become false capability donors. That candidate was rejected. The current change preserves the captured transport and tests the final selected projection, including property omissions. It does not add a source-binding layer or claim to repair retained-source attribution across configuration generations.

### DIFF
--- a/docs/concepts/model-providers/custom-providers.md
+++ b/docs/concepts/model-providers/custom-providers.md
@@ -381,6 +381,8 @@ Example (OpenAI-compatible):
 
     Recommended: set explicit values that match your proxy/model limits.
 
+    Model-selection metadata keeps capabilities tied to the API and endpoint that supplied them. A configured route change discards metadata from the previous route, while an explicit `thinkingLevelMap` is applied to the configured model.
+
   </Accordion>
   <Accordion title="Proxy-route shaping rules">
     - For `api: "openai-completions"` on non-native endpoints (any non-empty `baseUrl` whose host is not `api.openai.com`), OpenClaw forces `compat.supportsDeveloperRole: false` to avoid provider 400 errors for unsupported `developer` roles.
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f1bc414c6ab7bcd8da7546fe75a22ce2be3d7a98](https://github.com/openclaw/openclaw/pull/144180/commits/f1bc414c6ab7bcd8da7546fe75a22ce2be3d7a98)
- Landed commit: [9ec6795b3aaa1c0092267c7b29870d2fd638f556](https://github.com/openclaw/openclaw/commit/9ec6795b3aaa1c0092267c7b29870d2fd638f556)

PR #144181, opened 2026-09-10
fix: preserve model login settings and refresh gateway auth

obviyus:
Related: #136257, #144195.

## What Problem This Solves

Fixes an issue where provider login overwrote settings changed during sign-in or hid saved credentials when applying settings failed. Authentication changes also lacked acknowledged Gateway publication.

## Why This Change Was Made

Login merges its changes against current configuration, including model-ID upserts, while reporting overlapping conflicts separately from credential persistence. The administrator-only authentication refresh method waits for prepared state. Login, logout, profile order and local notifications share that publication operation.

## User Impact

Concurrent settings and defaults survive login. Full-provider logout cancels matching runs after credential removal even if refresh fails. Older Gateways receive one guarded status refresh plus restart guidance; this is not reported as completed publication.

Consumers: command-line login, logout and ordering use acknowledged authentication refresh.

## Evidence

Focused tests and real command-line/Gateway checks covered merges, publication, permissions, compatibility and logout during refresh failure. Shared-main logout still failed at credential removal on both compared versions; agent-local logout passed. Synthetic credentials were used.


Yigtwxx (commented):
Reviewed this revision (`735f661`) independently against `origin/main` — I read the changed paths and their callers rather than re-running the CLI evidence in the body. Four notes, the first one a compatibility question I would want answered before this lands.

### 1. `refreshRunningGatewayAuthState` loses its cross-version path

`src/commands/models/auth-refresh.ts` switches the CLI from `models.authStatus { refresh: true }` to the new `models.authRefresh`. The descriptor table shows what that costs:

- main: `["models.authStatus", "models-auth-status", "operator.read", "<=2026.7"]`
- here: `["models.authRefresh", "models-auth-status", "operator.admin", "2026.9", CONTROL_PLANE_WRITE]`

The body notes the protocol artifacts are regenerated "without a protocol version bump", so a client cannot separate the two by version either. When the CLI is newer than the running Gateway — the ordinary state right after `openclaw update` and before a Gateway restart, and the normal case for a remote/SSH-tunnelled Gateway — the request comes back as `unknown method: models.authRefresh`, `errorShape(ErrorCodes.INVALID_REQUEST, ...)` at `src/gateway/server-methods.ts:522`. Neither catch arm in `refreshRunningGatewayAuthState` matches it (it is a client request error, not `GatewayLocalBackendSharedAuthUnavailableError`, and not a transport `closed`/ECONNREFUSED), so every `models auth login`, `logout`, `paste-token`, `paste-api-key` and `order` against such a Gateway ends in `Warning: Model auth changes were saved, but the running Gateway could not refresh them. Run openclaw gateway restart` and the live Gateway is never refreshed. On main that same call worked against any Gateway back to `<=2026.7`.

The repo already has the idiom for this: `canFallbackToImplicitLocalGateway` treats an `INVALID_REQUEST` whose message is `unknown method: <legacyMethod>` as a legacy-server signal (`src/cli/gateway-rpc.ts:103`), and the same shape appears at `src/commands/audit.ts:152`, `src/cli/cron-cli/list-jobs.ts:31`, `src/agents/tools/sessions-resolution.ts:108` and `src/node-host/connection.ts:40`. `models.authStatus { refresh: true }` still refreshes on this branch (`refreshModelAuthStatusRuntimeState` is untouched), so one retry on that specific error would preserve the old behaviour for older Gateways while new ones keep the acknowledged path.

Worth noting the required scope also narrows from `operator.read` to `operator.admin`, so a read-scoped operator token that used to get a refresh now gets the restart warning instead.

### 2. "Concurrent unrelated settings survive sign-in" holds for object keys, not array keys

`persistProviderAuthResult` now derives `createMergePatch(params.config, applyProviderAuthConfigPatch(params.config, ...))` and replays it with `applyMergePatch(next, configPatch)`. That call passes no options, so `mergeObjectArraysById` is off and plain RFC-7396 array semantics apply: any array-valued key the derived patch touches is replaced wholesale with the value computed from the pre-login snapshot. For a user on an array-shaped roster — `agents.list`, which `applyProviderAuthConfigPatch` can rewrite through the `roster?.kind === "list"` branch in `src/plugins/provider-auth-choice-helpers.ts` — an entry added during a long OAuth is still dropped, exactly as it was by the old whole-config write.

The repo's own config writer opts in for this reason: `src/gateway/server-methods/config.ts:1056` passes `mergeObjectArraysById: true`. Either passing that option here, or narrowing the guarantee in the body to object-valued settings, would make the claim match the code.

### 3. `models.authOrderSet` quietly changes how it locates the owner, and when it responds

main called `prepareModelRuntimeSnapshot({ agentId, agentDir: preparedSnapshot.agentDir, workspaceDir: preparedSnapshot.workspaceDir, config: preparedSnapshot.config })`. The shared `refreshModelAuthStateAfterMutation` passes no `workspaceDir` and takes config from `context.getRuntimeConfig()`. `ownerKey` includes `workspaceDir`, so with it undefined the exact key no longer matches a configured owner published with a workspace, and resolution falls through to the configured-workspace fallback in `resolvePublishedOwner`, which returns a candidate only when exactly one matches (`src/agents/prepared-model-runtime.owner.ts:295-331`, selected by `projectPublishedModelRuntimeOwner`, `src/agents/prepared-model-runtime.published-owner.ts:49-54`). That is likely fine for the single-configured-owner case, but it is a silent change in owner lookup that the body does not mention; if it is deliberate, a line of comment would help the next reader.

Separately, the handler now awaits `refreshActiveProviderAuthRuntimeSnapshot()` inside the helper *before* `respond`, where main ran it fire-and-forget *after* responding. `authOrderSet` latency now includes a full runtime auth snapshot refresh.

### 4. Follow-up (not blocking this PR): the Gateway setup path still has both defects this PR fixes

`openclaw.setup.prepare.start` does what the CLI used to do: `readSetupConfigFileSnapshot()` at `src/gateway/server-methods/system-agent.ts:260`, then a provider login that can run for minutes, then `persistAuthProfiles()` at `:300`, then `writeWizardConfigFile(prepared.config, { baseSnapshot, baseHash })` at `:301`. That call passes no `mergeBase`, and `writeWizardConfigFile` replays a delta only when `mergeBase` is set — otherwise `transform: (current) => ({ nextConfig: config })` (`src/wizard/setup.shared.ts:112-115`), a whole-file write built from the pre-login snapshot.

The `baseHash`/`baseSnapshot` CAS with `maxAttempts: 1` (`setup.shared.ts:102`) turns the overwrite into a hard failure rather than silent data loss, but the user-visible result is the second defect from your own body: the credential is already persisted, and the caller surfaces a raw `config changed since last load` conflict with no equivalent of your `Credentials saved, but provider settings could not be applied`. Passing `mergeBase: baseConfig` there is what the other wizard writers already do (`src/commands/configure.wizard.ts:645`, `src/wizard/setup.ts:94`).

### Checked and clean

To save the next reader the walk: `src/gateway/model-account-connect.ts` applies no config patch at all (the comment above `connectUserModelAccount` is accurate), and both the `src/flows/model-picker.ts` wizard chain and the `src/system-agent/setup-inference-plan.ts` chain persist through merge patches applied to a freshly-read config under the mutation lock — the latter even conflict-checks with `mergePatchConflicts` (`src/system-agent/setup-inference-persist.ts:196`). Item 4 is the one remaining caller with the old shape.

obviyus:
Addressed the config-array gap in 79550acd6a6110cb41f0c58f87d985720e6be4a7. The regression reproduced a successful login that removed a concurrently added model. Login now checks the captured source snapshot against the writer's current source and preserves conflicting edits. Credentials remain saved, and the settings failure is reported separately. Fresh source/runtime inputs and ordinary model-list replacements still work. The conflict predicate is shared with the existing setup persistence owner. Targeted tests and a fresh six-scenario real CLI acceptance pass cover this behavior.

Older Gateways receive restart guidance. The read-only status refresh creates a transient read catalog and cannot provide the required acknowledgement that serving auth state was published. The new mutation remains administrator-only; read-only status access stays unchanged.

The prepared-owner lookup supports an omitted workspace and resolves the configured owner for the agent. The refresh uses current config so it does not bind publication to stale request facts. Order acknowledgement now includes auth publication; a refresh failure still returns the committed order with a warning. These behavior changes are documented in the PR body.

The separate Gateway setup path has a pre-existing partial-save reporting gap. Its stale base hash is checked before transformation, so adding mergeBase alone does not resolve that conflict. That setup lifecycle remains outside this extraction PR.

obviyus:
landpr-approve-admin-merge
pr: 144181
head: 17c8b548ba4392fb00029d93caaa95607b3294a2
scope: owner-review

obviyus:
landpr-approve-admin-merge
pr: 144181
head: f6044cbdd60a6235b29d9e671cfa6a12f0f0b34d
scope: owner-review

obviyus:
landpr-approve-admin-merge
pr: 144181
head: 117f47d41b4af9c9e806e411f471d0ea13594bb4
scope: owner-review

obviyus:
landpr-approve-admin-merge
pr: 144181
head: 2215907e8c28dc4f8e121141a93e85d9e468cf23
scope: owner-review

obviyus:
Owner-approved admin merge completed at `54c3655e12a66f24cf0189b32323ec21d17b3635` from head `2215907e8c28dc4f8e121141a93e85d9e468cf23`.

- The procedural security-owner involvement hold was waived under the [explicit owner approval](https://github.com/openclaw/openclaw/pull/144181#issuecomment-5624900359). Fresh exact-head ClawSweeper review found no remaining code or security defect and rated the proof sufficient.
- Proof: 398 targeted tests; real CLI settings/default preservation and older-Gateway fallback; real Gateway publication, scope admission, and revoked-request cancellation despite publication failure. The absent-parent regression failed before the correction and passed afterward.
- All [exact-head checks](https://github.com/openclaw/openclaw/actions/runs/34525682660) passed, including Android. No unrelated-red exception was needed.
- The merge retains `Co-authored-by: Ayaan Zaidi <hi@obviy.us>`.

Known limits remain disclosed in the PR body: shared-main removal locking and the unsuccessful targeted-profile fault attempt. Shared login results and chat presentation follow separately.

<details>
<summary>Exact-head review artifact (privacy-filtered)</summary>

```json
{
  "engine": "clawsweeper-local-review",
  "version": 4,
  "head": "2215907e8c28dc4f8e121141a93e85d9e468cf23",
  "status": "complete",
  "correctness": "patch is correct",
  "findings": [],
  "security": {
    "status": "cleared",
    "summary": "Administrator admission, local-state matching, credential isolation, and logout cancellation remain intact; no introduced security or supply-chain defect was found.",
    "concerns": []
  },
  "proof": {
    "status": "sufficient",
    "summary": "Retained compiled CLI and Gateway proof exercises settings preservation, acknowledged auth publication, read-only denial, guarded legacy fallback, and provider-request cancellation despite publication failure. Build identities remain f6044cbd and 117f47d4; the latest absent-parent correction adds focused failing-before/passing-after coverage on the same traced CLI writer path.",
    "evidenceKind": "logs",
    "needsContributorAction": false
  },
  "rating": {
    "proofTier": "A",
    "patchTier": "B",
    "overallTier": "B",
    "summary": "A useful owner-level repair with strong product evidence, explicit coverage limits, and no remaining blocking finding.",
    "nextSteps": []
  },
  "ownerReviewApproval": "https://github.com/openclaw/openclaw/pull/144181#issuecomment-5624900359",
  "retainedGateSha256": "33e43d771479bd2b08eef255c8c010a3ea870e14a16ee11a8183ba067ef2157a"
}
```

</details>

PR #144034, opened 2026-09-10
fix(models): select authored rows before applying defaults

steipete:
Related: #130706, #142100. Follows #143967.

## What Problem This Solves

Config defaults can collapse an alias into an exact destination after filling its omitted capabilities, allowing the alias row to contaminate the selected row. Later planning cannot recover which fields were authored.

## Why This Change Was Made

Reuse the planner's existing row-selection algorithm in the config owner, before capability defaults. Remove the later repeated ID normalization and let the planner delegate to the shared pure helper. Exact destinations retain their omissions and same-spelling duplicate costs retain their existing merge semantics. This adds 16 production lines; sparse capability inheritance and runtime pricing remain separate follow-ups.

## User Impact

The intended authored model row determines which fields are explicit before defaults or prepared runtime metadata are added.

## Evidence

Five baseline failures and one control established the producer defect. The fresh-main replay passed 132 focused tests, six final fixture rechecks, independent P2 review with unchanged production context, and the full changed gate. A test-table type correction produced byte-identical JavaScript and unchanged assertions.

One normal full build at the tested commit and 18 compiled cases passed: six config/defaults-to-inline/runtime cases and 12 planner/registry controls. All 11,929 root/workspace output files remained unchanged; the proof loaded no checkout TypeScript and made no HTTP or catalog rediscovery calls. All owned processes joined. This is compiled producer/planner proof, not an upstream-provider request.

Tested commit: `0ffd387049b04c07dda65f61f68684548105ca36`.

CI refresh: the unchanged authored-row patch is replayed onto main `eb9f4ff537c01eed0596bf0949f40be0d5b16e9f`, including the separate session-fixture cleanup in #144087 identified during the TaskFlow CI investigation. Every changed-file postimage and the stable task patch match `5b6b4ffc04f538feb334186e54aac03eab375fa6`. Original exact-commit runtime proof is preserved without restamping; fresh CI is required for `eb90eba9303c5eae8e7d1972bbb21a600077934a`. #144087 documents the historical causal limit.

PR #144182, opened 2026-09-10
fix(models): preserve exact authored pricing

steipete:
Related: #130706, #142100. Builds on #144034.

## What Problem This Solves

Runtime pricing could merge prices from an alias or a similarly spelled row into an exact configured model. It also normalized selected model IDs again while recovering authored prices, so alias chains could select a different row.

## Why This Change Was Made

Use the existing configured-row materializer to recover authored costs, preserving exact-row precedence and resolving source aliases once. Pass the captured normalization policies through both runtime overrides and configured fallback. Remove the obsolete scoped matcher.

This removes 33 production lines across three files. Same-spelling partial costs, explicit zero, and tiered pricing retain their existing merge rules. No storage or configuration surface changes.

## User Impact

An exact configured model keeps its own authored price fields. Omitted fields retain the catalog schedule instead of receiving another row's prices or materialized defaults.

## Evidence

- The original source produced eight intended regression failures, with seventeen controls passing.
- All 177 focused cases passed, including selected-model rematerialization. After extracting test fixtures to meet the file-size guard, all 57 affected cases passed again; the public-resolver pricing table and assertion body are unchanged.
- The complete changed gate and fresh P2 review passed on the final six-file commit. The extraction adds no production test seam or lint exception.
- One normal full build of `0c54207e6d045042e7973fc8e3eb6f58e68165d8` passed, including declarations. Through emitted `resolveModelAsync` and real registries, all 32 pricing cases passed, with 64 observations of the intended caller paths. V8 coverage confirmed execution of both override and configured-fallback owners.
- All 11,931 build files remained unchanged; 943 inventoried compiled modules loaded with zero checkout TypeScript imports or HTTP requests. Owned processes exited and synthetic state was removed. Eight additional compiled A/B controls reproduced the explicitly inherited limitation below.
- Exact-head CI remains required before landing.

A known limitation remains: retained configurations still consult the current global runtime/source pair. A separate baseline/candidate comparison reproduced identical behavior in both directions of an authorship switch, including distinct byte-identical runtimes, same-object rebinds, and retained originals/clones through both cost callers. This cut fixes row selection within the supplied source; it does not establish retained-source isolation.

### DIFF
--- a/src/agents/embedded-agent-runner/model.configured-fallback.ts
+++ b/src/agents/embedded-agent-runner/model.configured-fallback.ts
@@ -196,6 +196,7 @@ export function buildConfiguredFallbackModel(params: {
               cfg,
               configuredModel,
               catalogCost: staticCatalogModel?.cost,
+              providerMetadataOwners: params.providerMetadataOwners,
             }),
             contextWindow: resolvedFallbackContextWindow,
             contextTokens: configuredModel?.contextTokens ?? staticCatalogModel?.contextTokens,
--- a/src/agents/embedded-agent-runner/model.configured-overrides.ts
+++ b/src/agents/embedded-agent-runner/model.configured-overrides.ts
@@ -3,10 +3,8 @@ import { normalizeConfiguredProviderCatalogModelId } from "@openclaw/model-catal
 import { asOptionalRecord as readModelParams } from "@openclaw/normalization-core/record-coerce";
 import { normalizeLowercaseStringOrEmpty } from "@openclaw/normalization-core/string-coerce";
 import { mergeModelCost } from "../../config/model-cost.js";
-import {
-  findConfiguredProviderModel,
-  matchesProviderScopedModelId,
-} from "../../config/model-provider-config.js";
+import { findConfiguredProviderModel } from "../../config/model-provider-config.js";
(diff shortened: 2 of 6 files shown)


steipete:
The outstanding compiled-validation item is complete at `0c54207e6d045042e7973fc8e3eb6f58e68165d8`, and the PR body now records it. One normal full build passed. The emitted resolver with real registries passed all 32 pricing cases (64 caller observations); V8 coverage confirmed both runtime override and configured-fallback paths. All 11,931 build files remained unchanged, with zero checkout TypeScript imports or HTTP requests, and all owned processes/state were cleaned up.

The eight additional A/B controls reproduce the explicitly documented inherited source-projection limitation; they are not a claim that retained-source isolation was repaired. Exact-head CI is still required before landing.

steipete:
Merged via squash.

- Prepared head SHA: [0c54207e6d045042e7973fc8e3eb6f58e68165d8](https://github.com/openclaw/openclaw/pull/144182/commits/0c54207e6d045042e7973fc8e3eb6f58e68165d8)
- Landed commit: [05d86ff3299d3ad56e96b68c840bee6ddd97537a](https://github.com/openclaw/openclaw/commit/05d86ff3299d3ad56e96b68c840bee6ddd97537a)

PR #144185, opened 2026-09-10
test(update): keep staged Doctor checks at the package owner

steipete:
## What Problem This Solves

A staged-package Doctor test runs the full update CLI even though its distinct assertions concern the package installation owner. A separate compatibility-table variant changes a mock property that production no longer reads.

## Why This Change Was Made

Move the Doctor case into the existing package-owner tests, preserving real staging and swapping, exact Doctor argv and cwd, external service-repair policy, and installed-version propagation. The test also checks the live package version at Doctor entry and restores and retires its retained transaction through the owner API.

Remove only the stale legacy-target variant and its unused checkpointContinuation wrapper. Keep the three real update modes and the foreign-service, advisory, killed-Doctor, repair, and platform integrations.

## User Impact

No runtime behavior changes. Two full CLI invocations are removed, with the Doctor contract retained at its owning layer. Production and support code are unchanged; tests add 79 and remove 52 lines, with one fewer test overall.

## Evidence

The selected original/candidate/original runs pass 10/9/10 cases. The Doctor case takes 2.219s / 45ms / 1.724s. Whole-wrapper times are 25.73s / 23.38s / 22.84s, so this comparison does not establish a whole-wrapper speedup.

All 12 final targeted cases pass: all six package-owner cases, three compatibility modes, and the three neighboring Doctor integrations. The scoped changed-file gate passes, and independent review found no actionable issues. The existing command-boundary fixtures remain synthetic; real npm/network behavior is unchanged and is not claimed as new proof.

### DIFF
--- a/src/cli/update-cli.test.ts
+++ b/src/cli/update-cli.test.ts
@@ -8120,49 +8120,6 @@ describe("update-cli", () => {
     );
   });
 
-  it("runs package post-update doctor from the verified package root after a staged swap", async () => {
-    const tempDir = tempDirs.make("openclaw-update-staged-doctor-");
-    const { nodeModules, pkgRoot, entryPath } = await setupInstalledPackageAtNodeModules(
-      path.join(tempDir, "lib", "node_modules"),
-    );
-    primeNpmChannelTag("latest", "2026.5.14");
-    mockFileBackedPathExists();
-    mockNpmGlobalCommands(nodeModules, async (argv) => {
-      if (argv[0] === "npm" && argv[1] === "i" && argv.includes("--prefix")) {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6e00014c6d6344c118b2d6eef3e55e5e6bc067bf](https://github.com/openclaw/openclaw/pull/144185/commits/6e00014c6d6344c118b2d6eef3e55e5e6bc067bf)
- Landed commit: [71c388b52072113d5ed9790ad430d2761b71cbf6](https://github.com/openclaw/openclaw/commit/71c388b52072113d5ed9790ad430d2761b71cbf6)

PR #144189, opened 2026-09-10
refactor(cron): derive schedule and state from wire contracts

steipete:
## What Problem This Solves

Cron's internal type module repeats schedule and runtime-state fields already owned by the leaf Gateway wire type. Maintaining both declarations invites drift even when runtime behavior is unchanged.

## Why This Change Was Made

Derive `CronSchedule` and the shared portion of `CronJobState` from the existing wire owner. Keep the five wire-absent runtime markers locally, keep `deliverySuppressionReason` narrowed to the existing internal reason type, and retain the deprecated `lastStatus` annotation. Whole-job and delivery envelopes deliberately remain separate because their public and internal contracts differ.

The scheduler activation and stream-identity comments remain at the alias boundary. This changes no runtime expressions, protocol schema, configuration, migration, storage behavior, or authority decision. **Production: +20/-102, net −82. Tests: 0.** The public projection continues stripping only the first four local markers; `scheduleErrorCount` remains visible exactly as before.

## Evidence

- Separate original/candidate runs passed the same 18 protocol/projection/persisted-shape cases and two actual service lifecycle/restart cases, followed by the normal canonical core typecheck. The E2E wrapper performed its standard source-bound build prerequisite.
- A sealed compiler comparison completed all 32 fresh programs: 42 exported types, the sole value export, independent shape/optionality/mutability/literal witnesses, and 14 deliberate controls under each optional-property mode. The entire emitted module is 88 identical bytes; a deliberate value change produces different output.
- The broad diagnostic graph is **not claimed green**: it retains one diagnostic in the normal optional-property comparison and 16,648 in the stricter comparison. Twelve positive-comparison messages differ only in reviewed alias expansion/order/elision. Their full records are individually bound; all other diagnostic records, codes, source spans and related information must match. Control runs receive no message allowance. Ten comparator sanity cases reject unrelated, missing, duplicate and altered records. An independent audit verified all 32 outputs and 11,388 pinned physical compiler reads.
- Earlier failed proof designs remain preserved and are not stitched into the successful comparison. Managed precommit review completed three clean passes through P2. Actual Git review scope was correct; a base-SHA typo in explanatory prose was separately preserved and corrected.
- The [normal configured changed gate](https://github.com/openclaw/openclaw/actions/runs/34498709915) passed all 28 stages on Linux Node 24.19.0 / pnpm 12.3.4 in 26m45s command time. It verified source `195e3c700eb79ce95c9236531c2b587c987a9b84` plus candidate tree `b52f3d133ac1b665b08ded7dc7bb9f14feafc27b`; its lease stopped and temporary capsule/registration were removed. The nonfatal external portal-sync warning remains recorded.
- Signed commit `7a10702d901dd62d56083015b45f34fb78c89a3d` has exactly that gate-tested tree. A separate exact-commit review passed all three complete passes through P2 with no findings. Hosted PR CI and normal native preparation remain required before landing.

Current-main qualification through `802410cd1883` found the target and 28 other recorded contexts unchanged. The remaining schedule-doc context only replaces “V1 does not include” with “There is no” for the same absent WebSocket-source capability; its contract is unchanged.

## User Impact

No intended behavior change. Existing schedules, optional fields, internal scheduling markers, public projections and mutable state remain intact. No live provider or current-main execution is claimed by the compiler comparison.

### DIFF
--- a/src/cron/types.ts
+++ b/src/cron/types.ts
@@ -1,4 +1,7 @@
-import type { CronRunLogEntry as CronRunLogWireEntry } from "../../packages/gateway-protocol/src/schema/cron.types.js";
+import type {
+  CronJob as CronJobWire,
+  CronRunLogEntry as CronRunLogWireEntry,
+} from "../../packages/gateway-protocol/src/schema/cron.types.js";
 import type { EmbeddedAgentExecutionPhase } from "../agents/embedded-agent-runner/execution-phase.js";
 /** Cron scheduling, delivery, diagnostics, and store data contracts. */
 import type { FailoverReason } from "../agents/failover/signal.js";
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7a10702d901dd62d56083015b45f34fb78c89a3d](https://github.com/openclaw/openclaw/pull/144189/commits/7a10702d901dd62d56083015b45f34fb78c89a3d)
- Landed commit: [8033a8c6951e8705c08af875ce1ee482cfa2e21a](https://github.com/openclaw/openclaw/commit/8033a8c6951e8705c08af875ce1ee482cfa2e21a)

PR #144192, opened 2026-09-10
fix(update): stop retired v1 handoff rows from blocking config writes

steipete:
## Problem

A single retired v1 handoff row in the machine-shared lease database made **every config
write and every native service operation on the machine fail**, with:

```
existing managed handoff lease is incompatible; retain diagnostics and run openclaw triage manually
```

`withConfigWriteLock` and `withGatewayServiceOperationLock` call
`createManagedHandoffLeaseStore().assertSourceUnborrowed(path)` before admitting a writer.
That reads **every** row of `<tmp>/openclaw/managed-update-handoffs.sqlite` through `handle()`,
which throws for any payload the strict v2/v3 schema rejects.

Managed-handoff v1 rows (`{"version":1,pid,startIdentity}`) shipped in **2026.8.2**. They are
keyed by install roots that no longer exist, so nothing re-admits them and nothing retires them —
they simply sit in the shared database forever. Harmless until the retained-custody check
started reading them, at which point one survivor bricked config writes for the whole machine.

Observed on a clean checkout of `main`: `src/state/local-onboarding-state.test.ts` failed 3 of 13
(the three tests that reach a config write through `completeLocalSetupRecovery`), one of them by
timing out at 120s waiting on a completion lock whose competitor had already thrown. Proven at the
owner independently of any test — against the real machine database, which held 14 v1 rows next to
13 valid v2 rows:

```
store.assertSourceUnborrowed("~/.openclaw/openclaw.json")
→ THROWS: existing managed handoff lease is incompatible…
```

Deleting only the v1 rows from a copy made the same call return clean, isolating the cause exactly.

## Fix

`readRetainedSources()` now skips rows that decode as the **complete** retired v1 shape. That shape
records no `nativeBorrower`, so it can never have reserved a config path — whatever its process
state. The strictness that mattered is kept: unknown, damaged, or partially-v1 payloads still refuse
the caller through `handle()`, because unreadable prospective data cannot prove the source is
unrelated. Retained v3 custody still refuses config writers exactly as before.

`canCleanupLegacyManagedHandoff` is split into `readLegacyManagedHandoff` (shape only) plus a
liveness wrapper, so the v1 shape has one owner instead of two copies. The admission/cleanup
predicate is unchanged — cleanup eligibility still requires a positively dead process; only the
custody read, which makes no authority decision, ignores liveness.

This is +18 net production lines rather than neutral. The growth is the named split of one
predicate into two; collapsing it back would duplicate the v1 shape check in the lease store.

## Evidence

- `node scripts/run-vitest.mjs src/state/local-onboarding-state.test.ts` — **13 passed** (was 3 failed | 10 passed)
- `node scripts/run-vitest.mjs src/infra/update-managed-service-handoff-retarget.test.ts src/infra/update-retained-custody.unicode.test.ts src/infra/update-retained-custody.legacy.test.ts` — **45 passed**
- `node scripts/check-changed.mjs` — passed
- New `src/infra/update-retained-custody.legacy.test.ts` proves the user-visible flow through
  `withConfigWriteLock`, not a helper. Verified it fails on the original defect: with the fix
  reverted, 2 of its 3 tests fail with the exact production error.

## Follow-up (not in this PR)

Retired v1 rows are now harmless but still never retired, since only admission at their exact
install root can delete one. A bounded sweep of released v1 rows would be a reasonable separate change.

### DIFF
--- a/src/infra/update-managed-service-handoff-cleanup.ts
+++ b/src/infra/update-managed-service-handoff-cleanup.ts
@@ -4,19 +4,20 @@ import os from "node:os";
 import path from "node:path";
 import { isRecord } from "@openclaw/normalization-core/record-coerce";
 
-/** v1 shipped in 2026.8.2. This proves cleanup eligibility, never v2 authority. */
-export function canCleanupLegacyManagedHandoff(
+/**
+ * v1 shipped in 2026.8.2. Decoding the complete retired shape proves the row is
+ * a known released claim, never v2/v3 authority and never a native borrower.
+ */
+export function readLegacyManagedHandoff(
   payload: string,
-  processState: (identity: { pid: number; startIdentity: string }) => "live" | "dead" | "unknown",
-): boolean {
+): { pid: number; startIdentity: string } | null {
   let value: unknown;
   try {
     value = JSON.parse(payload);
   } catch {
-    return false;
+    return null;
   }
-  return (
-    isRecord(value) &&
+  return isRecord(value) &&
     Object.keys(value).length === 3 &&
     value.version === 1 &&
     typeof value.pid === "number" &&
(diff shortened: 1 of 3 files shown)


steipete:
Closing as superseded by #144208, which landed the same fix from a parallel session (merge commit 00edac804478).

Same root cause, same approach: `readRetainedSources()` now skips rows that decode exactly as the retired v1 shape, while undecodable rows still refuse. On rebase the two changes conflicted on the identical line of the identical function, and taking main's side left this branch with nothing but its test file.

#144208 puts the shape recognizer in the schema owner as `isRetiredManagedHandoffLeasePayload`, which is a better home than the cleanup module used here.

Verified against main at 068318110aec: `src/state/local-onboarding-state.test.ts` is 13/13, and this branch's own stricter test file passes unmodified against main's implementation — including the case asserting that retained v3 native custody still refuses config writers beside a retired row.

PR #144195, opened 2026-09-10
fix: retain setup credentials after a failed connection check

obviyus:
Related: #136257.

## What Problem This Solves

Fixes an issue where setup discarded credentials after a failed connection check and forced another sign-in. Managed local models could also receive a second verification turn with tools enabled.

## Why This Change Was Made

One setup operation saves the selected credential, confirms configuration in memory with one tool-free provider turn, and commits configuration only after success. Distinct profile identities preserve working credentials when a replacement fails. Saved-sign-in retry reuses the saved profile.

## User Impact

Failed confirmation retains credentials and leaves configuration unchanged. Local models receive one tool-free check. No schema change. Custom endpoint settings may need re-entry after restart; synchronous plugin imports still share the Gateway event loop.

Consumers: setup, wizard, completion and update-repair callers use the shared confirmation operation.

## Evidence

The public setup request reproduced credential loss. Real-provider checks covered success, rejection, retry and replacement with exact credential/configuration readback. Focused tests and builds passed. Browser checks verified saved-profile retry using a mocked Gateway; provider behavior was checked separately.


obviyus:
The credential replacement finding is resolved. New submissions use separate saved profiles, and confirmation disables credential fallback. Real-provider proof shows one request each for initial success, rejected replacement, and successful verification of the original route. Both credentials remain saved, and rejection leaves the working configuration unchanged.

The accepted extraction scope from #136257 explicitly removes the detection worker. This PR keeps that cutover and documents the remaining synchronous plugin-loading risk. The timer does not claim to restore worker isolation.

The detector now retains the shipped **30-second** allowance, with cancellation and completed partial results. The earlier 10-second value was taken from historical source and has been corrected. A regression covers a valid choice that takes 20 seconds to discover. All four detector cases and 79 LM Studio cases pass. The separate local review accepted the credential implementation and live proof; its only remaining discovery timing choice is addressed by retaining the existing allowance.

No schema, table, column, or migration is added. Saved credentials continue to use the existing auth-profile store and its existing record format.

obviyus:
Landed as `496433f5ae0cd4e1f0323cbf9a142fb4cd8f7a77` after a clean exact-head review and [successful CI](https://github.com/openclaw/openclaw/actions/runs/34516542701).

- **Bug:** failed first-run confirmation discarded the submitted credential and forced another sign-in.
- **Fix:** save a distinct credential through the existing auth store, confirm the in-memory candidate once without tools or credential fallback, and commit configuration only after success.
- **Proof:** real-provider success, HTTP 401 rejection, saved-sign-in retry, and rejected replacement with the original connection still working. The merged source passed 78 setup-owner tests and 44 Gateway tests. Independent browser proof covers saved-row selection with a mocked Gateway.
- **CI recovery:** the initial UI shard failed ten cases. Focused base and candidate probes each passed 64 tests. One retry through the existing hosted-runner route passed all 355 UI files and 5,880 tests. No source, assertion, timeout, or CI policy changed; no unrelated-red exception was used.
- **Retained limit:** synchronous provider imports can still block the Gateway event loop after the accepted worker removal. The 30-second asynchronous deadline does not provide worker isolation.

No schema, table, column, or migration was added. The complete test-deletion ledger and raw proof remain in the task result.

Issue #137525, opened 2026-09-03
2026.8.2: memory search forces a full reindex on every call + O(n) plugin-state COUNT per write → in-session memory_search hits 15s deadline


tedbarnett:
## Summary

After upgrading from 2026.6.10 to 2026.8.2, `memory_search` inside agent turns times out at its hardcoded 15 s deadline on every call, and `openclaw memory search` from the CLI takes ~30 s wall (was 1–2 s on 2026.6.x). Profiling shows the search itself is negligible; the cost is (A) a **forced full in-place reindex on every search** and (B) an **O(n) `COUNT(*)` on plugin-state per write**.

## Environment

- OpenClaw 2026.8.2 (0965053), macOS 26.6, Node 22.23.1 (x64 via nvm), sqlite-vec 0.1.9 loaded
- memory provider `openai` / `text-embedding-3-small`, ~500 files, ~3,100 chunks
- Agent store `openclaw-agent.sqlite` ≈ 578 MB (`memory_embedding_cache` 203 MB / 7.1k rows; `transcript_events` 181 MB; `memory_index_chunks` 96 MB)
- `memory status --deep`: index identity valid, `dirty: false`, vector index `complete`

## Measurements

- `openclaw memory search --agent main "test query"` ×3: 29.6 s / 30.0 s / 30.2 s wall, ~24 s user CPU each. No warm/cold difference.
- `openclaw memory status --agent main`: 4.25 s → CLI startup is ~4 s, the remaining ~26 s is the search path.
- `node --cpu-prof` on the CLI search, self-time top frames:
  - ~8.5 s in `openclaw-state-db` statement execution, all under `pluginStateRegisterIfAbsent → enforcePostRegisterLimits → countLivePluginStateEntries`
  - ~12 s across `memory-schema ensure`, `execTimedTransactionStep`, `replaceMemoryVectorRow`, `replaceVirtualTable`, `runSqliteCheck` (integrity_check), `seedEmbeddingCache`, `deleteMemoryFtsRows`, `deleteVectorRowsForSource`, `upsertMemoryEmbeddingCache` — i.e. a full shadow-DB reindex
  - `searchVector` 0.01 s, `searchKeyword` 0.02 s
[...]

PR #144198, opened 2026-09-10
fix(plugin-state): reduce redundant quota work during large updates

steipete:
Related: #137525

## What Problem This Solves

Resolves avoidable synchronous database work during large plugin-state updates, including memory workspace replacements, when the plugin-wide retention limit is already equal to or stricter than the namespace limit.

## Why This Change Was Made

Ordinary `evict-oldest` writes now use the existing plugin quota check to enforce both limits when the namespace cap is at least the plugin cap. The plugin check selects the same oldest eligible rows from the current namespace, preserves the written key and sibling namespaces, and retains rollback when the store starts over capacity.

Reject-new admission, namespace-only journal cursor enforcement, Doctor retention, expiry cleanup, timestamp/TTL handling, and per-call transactions retain their existing paths. This bounded query improvement changes neither the schema nor the public API.

## User Impact

Eligible plugin-state writes perform one fewer quota scan. Whole-plugin counts and complete listings remain, so this does not claim to resolve every delay reported in #137525.

## Evidence

- Both new work-count cases fail on the original implementation with one namespace COUNT instead of zero, after their retained-row and sibling/foreign preservation assertions pass.
- All 66 focused tests pass across five suites covering stores, retention, expiry, Doctor imports, and errors. They include initially overfull state, protected-key ties, insufficient-victim rollback, and deferred journal cursor enforcement. Changed-code checks pass; independent review found no actionable findings.
- One synthetic Linux x64 source/API workload on Node 24.19.0 / SQLite 3.53.3 observed zero namespace counts for 64 updates, versus 64 in the retained baseline. It retained all 64 plugin counts, point reads, upserts, and zero-deletion expiry sweeps; one listing returned 50,000 live rows. All 49,938 untouched rows matched before and after. Both resulting databases passed independent full SQLite integrity and data checks. This proves deterministic work reduction, not a cross-host timing speedup.
- [Blacksmith Testbox run](https://github.com/openclaw/openclaw/actions/runs/34500217524): the workload passed, but the original artifact export exited 7 at its 30-second deadline. A separately authorized read-only native download recovered the finalized archive; its exact size, original partial prefix, original manifest, all listed file hashes, and source identity were verified. The original whole-archive control digest was unavailable. The lease is stopped; the export failure remains a recorded qualification.

### DIFF
--- a/src/plugin-state/plugin-state-store.retention.test.ts
+++ b/src/plugin-state/plugin-state-store.retention.test.ts
@@ -1,0 +1,409 @@
+// Plugin state retention preserves quotas, eviction order, and rollback.
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { trackSqliteStatementExecutions } from "../../test/helpers/sqlite-statement-execution-counter.js";
+import { openOpenClawStateDatabase } from "../state/openclaw-state-db.js";
+import {
+  createOpenClawTestState,
+  type OpenClawTestState,
+} from "../test-utils/openclaw-test-state.js";
+import {
+  createPluginStateKeyedStore,
+  registerPluginStateSyncSequencedJournalEntry,
+  resetPluginStateStoreForTests,
+  sweepExpiredPluginStateEntries,
+} from "./plugin-state-store.js";
+import {
+  clearPluginStateStoreForTests,
+  seedPluginStateEntriesForTests,
+  setMaxPluginStateEntriesPerPluginForTests,
+} from "./plugin-state-store.test-helpers.js";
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8a856bf2c92d6a8368bf9dd4d0924d96452ea1e2](https://github.com/openclaw/openclaw/pull/144198/commits/8a856bf2c92d6a8368bf9dd4d0924d96452ea1e2)
- Landed commit: [ea41203f6c6de6fee678acee8a7385484577eb84](https://github.com/openclaw/openclaw/commit/ea41203f6c6de6fee678acee8a7385484577eb84)

PR #144199, opened 2026-09-10
fix(doctor): report schema guard snapshot cleanup failures

SunnyShu0925:
## What Problem This Solves

During an update, the schema safety guard creates a temporary read-only copy of the state database to check whether a driving updater is active; when removal of that temporary copy failed, the failure was silently ignored and the copy accumulated on disk with no signal. This fix checks the cleanup result and surfaces a diagnostic warning on failure, while preserving the guard's existing schema decision and upgrade-flow behavior.

- Problem: `readDrivingUpdater` in `src/commands/doctor-update-schema-guard.ts` called `snapshot.cleanup()` in a `finally` block without checking its boolean return value. When `removeTempDirectory` fails (disk full, permission denied, file-handle contention), `cleanup()` returns `false` but the caller ignored it, leaving the temporary SQLite copy behind with no error signal.
- Solution: Capture the read outcome first (value or caught cause), then check `snapshot.cleanup()`. On cleanup failure, build a diagnostic warning string and return it alongside the updater; the guard emits the warning and proceeds with the existing schema decision unchanged. Snapshot preparation failures are caught and tolerated (no updater), matching the pre-fix boundary. In JSON mode the warning is written directly to `process.stderr` so it does not corrupt stdout JSON output (the guard runs before console capture is installed).
- What changed: `src/commands/doctor-update-schema-guard.ts` — `readDrivingUpdater` returns `{ updater, cleanupWarning }`; `prepareSqliteReadOnlyLocation` is wrapped in a try/catch for preparation-failure tolerance; `guardUpdateDoctorSchemaUpgrade` routes the warning to stderr in JSON mode or `runtime.log` otherwise.
- What did NOT change: The inner `try/finally` that closes the database handle is preserved. The read logic, schema validation, blocker detection, refusal policy, and call sites are unchanged. Normal-path behavior (cleanup succeeds) is identical to before.

## Why This Change Was Made

This is the same anti-pattern that #143844 fixed in `update-candidate-state.ts`: a `prepareSqliteReadOnlyLocation*` call site that ignores the `cleanup()` return value. Two sibling call sites already check the return value and surface cleanup failure. This call site was missed.

The fix follows the #143844 outcome-capture + cleanup-check template with adaptations for this call site: (1) cleanup failure is a non-fatal warning because the guard's empty catch tolerates read failures and a fatal throw would be swallowed; (2) snapshot preparation failures are caught inside `readDrivingUpdater` to preserve main's tolerating boundary; (3) in JSON mode the warning goes to `process.stderr` because the guard runs before `installConsoleCapture` and `runtime.log` (console.log → stdout) would prefix the JSON output.

## Evidence

Real behavior proof using real SQLite fixtures + `fs.rmSync` fault injection (EACCES on `openclaw-sqlite-readonly-*` staging directories). Test 4 uses a missing source DB to trigger `prepareSqliteReadOnlyLocation` rejection. Test 5 verifies JSON-mode stderr routing.

```console
$ node --import tsx /home/code/github/contributions/BUG/BUG-073-evidence/proof.mjs

Test 1: cleanup fail + read ok   → threw=false, warning via runtime.log ✅
Test 2: cleanup fail + read fail → threw=false, warning with read cause  ✅
Test 3: normal path              → threw=false, logCalls=[] (no warning)  ✅
Test 4: preparation fail         → threw=false, logCalls=[] (tolerated)   ✅
Test 5: JSON mode + cleanup fail → warning on stderr, stdout clean        ✅

All 5 tests passed.
```

Pre-fix verification (PR head `8e612eb`, same proof): Test 5 FAIL — warning appeared on stdout (via `runtime.log` → `console.log`), stderr empty. This confirms the JSON-output corruption ClawSweeper identified.

Behavior matrix:

```text
scenario                        => pre-fix (main)         => post-fix result
cleanup fails, read succeeds    => silent (temp left)     => warning via runtime.log, upgrade proceeds
cleanup fails, read fails       => silent (temp left)     => warning with read cause, upgrade proceeds
cleanup succeeds, read ok       => guard proceeds         => guard proceeds (unchanged, no warning)
preparation fails (no source)   => tolerated (main)       => tolerated (no throw, no warning)
cleanup fails, JSON mode        => warning on stdout      => warning on stderr (stdout clean for JSON)
```

## AI Assistance

- AI-assisted: Yes
- Tools: Claude Code (Claude Sonnet 4.6)
- Round 3: ClawSweeper identified that `runtime.log` (console.log → stdout) in JSON mode runs before `installConsoleCapture`, corrupting JSON output. Fixed by routing JSON-mode warnings to `process.stderr.write` directly. Added Test 5 for stderr routing with pre-fix/post-fix双向验证.
- Confirmed understanding of code changes: Yes.
- autoreview: N/A (worktree; oxfmt + oxlint + tsgo core + vitest 25 tests passed)

### DIFF
--- a/src/commands/doctor-update-schema-guard.ts
+++ b/src/commands/doctor-update-schema-guard.ts
@@ -1,3 +1,4 @@
+import path from "node:path";
 import { formatCliJsonFailure } from "../cli/failure-output.js";
 import { exitCliAfterOutput } from "../cli/one-shot-exit.js";
 import { clearNodeSqliteKyselyCacheForDatabase } from "../infra/kysely-sync.js";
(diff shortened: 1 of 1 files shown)


SunnyShu0925:
Addressed ClawSweeper's compatibility decision: cleanup failure is now **non-fatal**.

**Change in this revision:** `readDrivingUpdater` no longer throws on cleanup failure. It returns `{ updater, cleanupWarning }`; the guard emits the warning via `runtime.log` (the established Doctor output path, routed to stderr under JSON machine output) and proceeds with the schema decision already determined from the read. Read failures remain tolerated.

This adopts ClawSweeper's recommended option 1 (keep cleanup reporting nonfatal) and resolves the three Before-merge items:

- **Merge risk (P1):** a transient snapshot-removal failure no longer aborts an otherwise permitted upgrade. The schema decision (refuse or permit) is unaffected by cleanup outcome; the exit retry still runs best-effort.
- **Next step (P2):** behavior selected = report cleanup failure as a warning while preserving the schema decision. Boundary-level evidence below.
- **Maintainer decision:** the warning variant is chosen for this call site because its tolerating `catch {}` makes a fatal throw semantically worse than the pre-fix behavior (it would discard the successfully-read updater and silently pass the upgrade). Sibling sites (#143844, BUG-069~072) throw because their callers do not tolerate errors; this site differs.

**Behavior matrix:**

| scenario | pre-fix | post-fix |
|---|---|---|
| cleanup fails, read succeeds | silent (temp left) | warning via `runtime.log`, upgrade proceeds (decision unchanged) |
| cleanup fails, read fails | silent (temp left) | warning with read cause, upgrade proceeds |
| cleanup succeeds, read ok | guard proceeds | guard proceeds (unchanged, no warning) |

**Evidence** (real SQLite fixture + `fs.rmSync` fault injection, 3 tests pass):

```
Test 1: cleanup failure + read ok    → threw=false, warning via runtime.log ✅
Test 2: cleanup failure + read fail  → threw=false, warning with read cause ✅
Test 3: normal path                  → threw=false, logCalls=[] (no warning) ✅
```

Pre-fix on `origin/main`: Test 1 returned `logCalls=[]` (fully silent). No code-level findings were reported; this revision keeps the patch minimal (+52/-14, one file) and aligned with the `note*` Doctor-warning precedent (`noteSourceInstallIssues`, `noteStalePluginRuntimeSymlinks`) and the `runtime.log` path used for deferred schema-publication messages in `doctor-health.ts`.

SunnyShu0925:
@clawsweeper re-review

SunnyShu0925:

PR #138758, opened 2026-09-05
perf(sqlite): reuse compiled plugin state point reads

steipete:
## What Problem This Solves

Plugin state point reads rebuild and compile the same Kysely query for every lookup. Reconstructing a large Reef audit chain or Buzz reconnect cursor set repeats that JavaScript work thousands of times even though the physical connection and query shape are unchanged.

## Why This Change Was Made

Reuse the existing `prepareSqliteQuerySync` helper once per physical SQLite connection at the shared point-read owner. Every invocation still binds its current plugin, namespace, key, and expiry time and executes the same fresh read. Native statement invalidation remains owned by the existing helper, and the compiled runner has weak connection ownership.

This preserves exact point-read timing. Replacing these callers with namespace snapshots would change concurrent reconnect observations or expose unrelated malformed audit rows. No result rows, authorization decisions, or expiry facts are cached.

## User Impact

Existing plugin-state consumers perform less repeated query-building work without changing their results or setup. Schema, stored rows, transactions, error handling, quotas, and public APIs are unchanged.

## Evidence

- The regression fails before the change with seven compilations instead of one. It passes across two independently opened native connections with different plugins, namespaces, keys, absent rows, and exact expiry boundaries.
- All 190 tests in 12 owner/helper/Buzz/Reef files pass, as does the full changed-file gate. Fresh independent Codex review found no actionable P0–P2 findings.
- Five native phases exercise the actual candidate source with synthetic state: a 30,000-entry committed Reef chain, 1,020 Buzz reconnect rooms, malformed uncommitted overflow, an interleaved reconnect update, and physical close/reopen. Complete result hashes and all persisted row bytes/timestamps match baseline. The committed chain still performs 30,001 fresh SELECTs but no per-read Kysely compilations after its prepared reader is initialized; Buzz retains 1,021 SELECTs and only its separate list-query compilation.
- The retained public store reads correctly after its old physical handle closes and a new connection opens. The native proof verifies candidate source hashes before and after. These are structural compilation/query measurements, not a latency benchmark or external channel test.

Production delta is +16 lines for one connection-owned compiled reader; tests add 47. There is no new cache framework, dependency, configuration, or test-only production hook.

PR #144206, opened 2026-09-10
perf(sqlite): reduce repeated plugin-state write compilation

steipete:
Related: #138758, #144198

## What Problem This Solves

Repeated plugin-state writes and Doctor imports rebuild the same database queries even when their SQLite statements are already cached.

## Why This Change Was Made

Cache fixed upsert and insert-if-absent query compilation with each physical connection, using the existing point-read pattern and fresh bindings for every call. The binder now expresses the complete row it already produces.

## User Impact

Less query-construction work during plugin-state writes and imports. SQL, conditional insertion, TTLs, quotas, transaction boundaries, and error handling are unchanged. This composes with the newly landed namespace quota-scan optimization.

## Evidence

The new public-store cases fail on the original code only at the compilation budget (9 upserts or 6 conditional inserts instead of 1), then pass with fresh plugin, namespace, key, value, timestamp, and expiry bindings across reopened connections. The existing prepared-read case moved unchanged into the focused prepared-query suite.

Before rebase, 101 focused store/helper tests and required checks passed, with a clean independent Codex review. On the final rebased code, all 56 combined store/prepared/retention/import tests, core/test types, typed lint, formatting, and real on-disk public-store smokes passed. The smokes exercise two simultaneous databases, retained-store writes after close/reopen, TTL replacement, conditional conflicts, and fresh 500-row imports, at both 1,000 and 50,000-entry namespace caps.

Seven local samples per version and cap compare the new main baseline against this candidate, with the quota optimization present in both. Median 500-row Doctor import replay time fell from 13.88 ms to 6.45 ms at the 1,000-entry cap and 15.35 ms to 6.20 ms at 50,000. Separate counters show 3,000 warmed write compilations eliminated; native INSERT preparation stayed at one. Individual-write timings were inconsistent across scenarios, and memory results varied; these synthetic measurements do not establish general throughput or memory improvements. Runtime proof used actual production APIs through the repository tooling loader and isolated synthetic state, not an operator Gateway or published package.

### DIFF
--- a/src/plugin-state/plugin-state-store.prepared.test.ts
+++ b/src/plugin-state/plugin-state-store.prepared.test.ts
@@ -1,0 +1,136 @@
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { getNodeSqliteKysely } from "../infra/kysely-sync.js";
+import { openOpenClawStateDatabase } from "../state/openclaw-state-db.js";
+import {
+  createOpenClawTestState,
+  type OpenClawTestState,
+} from "../test-utils/openclaw-test-state.js";
+import {
+  closePluginStateDatabase,
+  createPluginStateSyncKeyedStore,
+  resetPluginStateStoreForTests,
+} from "./plugin-state-store.js";
+import {
+  clearPluginStateStoreForTests,
+  seedPluginStateEntriesForTests,
+} from "./plugin-state-store.test-helpers.js";
+
+let testState: OpenClawTestState;
+beforeAll(async () => {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [51b705fa9b1ce45a6238817e95536624cbac26c9](https://github.com/openclaw/openclaw/pull/144206/commits/51b705fa9b1ce45a6238817e95536624cbac26c9)
- Landed commit: [d1e952f991722fab51ff35f2badde602136dbcdb](https://github.com/openclaw/openclaw/commit/d1e952f991722fab51ff35f2badde602136dbcdb)

PR #144210, opened 2026-09-10
fix(web-fetch): keep visible text after hidden HTML

### LINKED ISSUE #144150: web_fetch hidden-content sanitizer discards visible page content: hidden matched inside unrelated attribute values, and drop region runs to end of document on omitted end tags
## Summary
The `web_fetch` hidden-content sanitizer matches the `hidden` attribute against the raw attribute text of a tag, values included, so any element whose ordinary attribute value happens to contain a whitespace-preceded word starting with `hidden` (for example `title="The hidden cost of cloud"`, `alt="a hidden gem"`, `aria-label="Show hidden replies"`) is dropped along with its entire subtree before extraction. A second defect in the same pass lets the drop region run to end of document when the removed element's own end tag is omitted, which is legal for `li`, `td`, `tr`, `p`, `option` and `dd`.

## Environment
- Surface: web_fetch

## Steps to reproduce
1. Call `web_fetch` on a page whose content wrapper carries an innocuous attribute mentioning the word "hidden", for example:
   `<html><body><article title="The hidden cost of cloud"><h1>Cloud Costs</h1><p>Real body text of the article.</p></article></body></html>`
2. Or on a page with a `d-none` list item whose `</li>` is omitted, as minified nav markup commonly emits:
   `<html><body><ul><li class="d-none">Nav item<li>Visible item</ul><p>Article body follows here.</p></body></html>`
3. Equivalently, drive the production sanitizer and basic extractor directly: `sanitizeHtml(html)` from `src/agents/tools/web-fetch-visibility.ts` and `extractBasicHtmlContent({ html, extractMode: "text" })` from `src/agents/tools/web-fetch-utils.ts`.

## Expected
The article text is retained. `hidden` should be recognized only as a real attribute name, and an attribute value that merely contains the word "hidden" should have no effect on visibility. A removed element's drop region should end at that element, not at end of document.

## Actual
The whole subtree is discarded. Case 1 sanitizes to `<html><body></body></html>` and extracts nothing. Case 2 sanitizes to `<html><body><ul>` and extracts nothing, so everything after the list item is lost. An `aria-label` mentioning "hidden" on `<body>` or `<html>` blanks the entire page. `web_fetch` then reports `Web fetch extraction failed: Readability, <provider>, and basic HTML cleanup returned no content.` (`src/agents/tools/web-fetch.ts:898`), or silently returns only chrome when some of the page survives.

## Root cause
`src/agents/tools/web-fetch-visibility.ts:123-133` builds each visibility reader from one grammar and runs it over the raw attribute string, values included:

```ts
function createAttributeReader(attribute: "aria-hidden" | "class" | "hidden" | "style" | "type") {
  const pattern = new RegExp(
    `(?:^|\\s)${attribute}(?:\\s*=\\s*(?:"([^"]*)"|'([^']*)'|([^\\s"'=<>\`]+)))?`,
    "i",
  );
  return (attrs: string): string | undefined => {
    const match = attrs.match(pattern);
    return match ? (match[1] ?? match[2] ?? match[3] ?? "") : undefined;
  };
}
```

For `hidden` this compiles to `/(?:^|\s)hidden(?:\s*=\s*(?:"([^"]*)"|'([^']*)'|([^\s"'=<>`]+)))?/i`. The value group is optional and nothing anchors the end of the name, so the substring `hidden` inside a quoted value matches and the reader returns `""`. `shouldRemoveElement` at `:151-153` treats presence as a boolean:

```ts
  if (readHidden(attrs) !== undefined) {
    return true;
  }
```

so `""` is "the hidden attribute is present" and the element is removed. A word boundary alone does not fix this: `title="Show hidden files"` has whitespace on both sides of the token.

The second defect is `removeMarkedElements` (`:184-232`) together with `popDroppedElement` (`:176-181`):

```ts
function popDroppedElement(dropStack: string[], tagName: string): void {
  const index = dropStack.lastIndexOf(tagName);
  if (index >= 0) {
    dropStack.length = index;
  }
}
```

The drop region is closed only by an end tag whose name is on the stack. When the removed element's own end tag is absent, nothing ever pops it and every remaining byte of the document is suppressed.

Suggested fix, offered as a suggestion since this is an issue and not a PR: have `shouldRemoveElement` read real attribute names rather than substrings of the raw attribute text, reusing the name/value walker that already exists at `src/agents/tools/web-fetch-utils.ts:53-108` (`readAttributeValue`) instead of `createAttributeReader`'s regex. For the drop stack, cap the drop region to the removed element with a depth counter rather than a name stack, or close it at the first ancestor end tag.

## Sibling surfaces
Both extraction paths share the same `sanitizeHtml`, so neither rescues the other. The primary path is `web_fetch` execute (`src/agents/tools/web-fetch.ts:1034`) into `runWebFetch`, the `text/html` branch at `web-fetch.ts:860-871`, `extractReadableContent` (`src/web-fetch/content-extractors.runtime.ts:29-36`), then `extensions/web-readability/web-content-extractor.ts:113` `const cleanHtml = await sanitizeHtml(request.html)`. The basic fallback at `src/agents/tools/web-fetch-utils.ts:421` calls the identical function. Readability cannot restore the dropped nodes because `sanitizeHtml` runs on the raw HTML before `parseHTML`. The provider fallback at `web-fetch.ts:873-901` only fires when the extractor returns no text, and most deployments have no fetch provider configured, so it does not cover the partial-loss case at all. `readAriaHidden` is built from the same flawed grammar but compares `=== "true"`, so a value-substring match yields `""` and is unaffected.

Severity P2: silent content loss on a realistic page shape, on the common `web_fetch` path.

## Real behavior proof
Behavior addressed: `web_fetch` HTML sanitization drops visible article content when an unrelated attribute value contains the word "hidden", and truncates the document at a removed element whose end tag is omitted.
Real environment tested: production `sanitizeHtml` and `extractBasicHtmlContent` from the shipped `src/agents/tools` modules, driven in-process on Node 24 in a read-only checkout, at commit c973fe7e721905ba2318994f1f47ae97677cf9ac.
Exact steps or command run after this patch: `npx tsx drive-webfetch.mts`, calling `sanitizeHtml(html)` and `extractBasicHtmlContent({ html, extractMode: "text" })` on each input below, then the same four pages again with one word changed as a control.
Evidence after fix:
```
--- A: wrapper title mentions the word hidden
input html      : <html><body><article title="The hidden cost of cloud"><h1>Cloud Costs</h1><p>Real body text of the article.</p></article></body></html>
sanitizeHtml    : "<html><body></body></html>"
extracted text  : null

--- B: figure alt mentions the word hidden
input html      : <html><body><figure alt="a hidden gem"><figcaption>Real caption text.</figcaption></figure></body></html>
sanitizeHtml    : "<html><body></body></html>"
extracted text  : null

--- C: body aria-label mentions the word hidden
input html      : <html><body aria-label="Show hidden replies"><h1>Thread</h1><p>Every reply in this thread.</p></body></html>
sanitizeHtml    : "<html></html>"
extracted text  : null

--- D: d-none list item with omitted end tag
input html      : <html><body><ul><li class="d-none">Nav item<li>Visible item</ul><p>Article body follows here.</p></body></html>
sanitizeHtml    : "<html><body><ul>"
extracted text  : null

--- A control: same page, title says 'rising' instead of 'hidden'
sanitizeHtml    : "<html><body><article title=\"The rising cost of cloud\"><h1>Cloud Costs</h1><p>Real body text of the article.</p></article></body></html>"
extracted text  : "Cloud Costs\nReal body text of the article."

--- B control: alt says 'a rare gem'
sanitizeHtml    : "<html><body><figure alt=\"a rare gem\"><figcaption>Real caption text.</figcaption></figure></body></html>"
extracted text  : "Real caption text."

--- C control: aria-label says 'Show more replies'
sanitizeHtml    : "<html><body aria-label=\"Show more replies\"><h1>Thread</h1><p>Every reply in this thread.</p></body></html>"
extracted text  : "Thread\nEvery reply in this thread."

--- D control: same list, hidden item carries an explicit </li>
sanitizeHtml    : "<html><body><ul><li>Visible item</li></ul><p>Article body follows here.</p></body></html>"
extracted text  : "Visible item\nArticle body follows here."
```
Observed result after fix: changing one word inside an attribute value, or omitting a legal-to-omit end tag, flips the same page from fully extracted to nothing at all, which is the silent content loss this issue reports.
What was not tested: no live network fetch was performed and no fetch provider was configured, so the provider-fallback branch at `web-fetch.ts:873-901` was reached only by reading the code, not by execution. The Readability extractor plugin itself was not driven end to end; the shared `sanitizeHtml` call it makes at `extensions/web-readability/web-content-extractor.ts:113` was confirmed by reading. No fix was written and no patch was applied, so there is no post-fix run: the control column is the same production code on inputs the defect does not trigger on. Real-world frequency of the triggering attribute values was not measured.

ruel225:
Closes #144150.

## What Problem This Solves

`web_fetch` can discard an article when an ordinary attribute value contains “hidden”, or when a hidden list item, paragraph, table cell, definition item, or option omits its legal ending. Related malformed comment and script boundaries can also swallow later visible text.

## Why This Change Was Made

Read complete attribute names and values, and keep hidden state with the actual open-element scope. Nested descendants and unmatched closing tags cannot release a hidden ancestor. The shared scanner keeps script text opaque, preserves complete tag names and closing tokens, and handles comment endings consistently before both extraction routes.

The indexed element stack avoids repeated full-depth searches on malformed pages. No parser dependency or network-access policy is added.

## User Impact

Visible articles and siblings survive harmless attribute values, legal omitted endings, and the tested malformed boundaries. Real hidden attributes, classes, styles, script data and comment contents remain excluded. Readability, basic extraction and JSON-LD titles retain their applicable behavior.

## Evidence

- Paired public agent commands invoked production `web_fetch` against hosted HTML and captured the subsequent provider requests. Main loses the original article/list inputs; corrected results retain their visible text and exclude hidden controls.
- 253 focused sanitizer, renderer, Readability, real HTTP tool, output-contract and provider-fallback tests pass. Ordinary lint and repository guards pass. Full types remain with hosted CI.
- Affected public runs pass the reported boundaries across basic and Readability extraction, text/Markdown, script delimiters, quoted closing tags, recovered comments and title metadata. Further checks preserve complete tag identities across scope, Unicode and void-element decisions. The latest seven checks retain visible content inside ordinary non-ASCII names while excluding hidden descendants; the prior 38-check matrix covers the complete-name scope cutover. All earlier failed inputs remain in the retained review evidence.
- Independent source/security review has no actionable findings. Source-blind acceptance confirms the original behaviors and repaired comment/tag boundaries, including actual basic fallback, inner visible content, raw-text controls, both modes and structured titles. The latest independent passes add 15 cases for complete names and visible children, including six actual basic-fallback results.
- Measured malformed-input comparison: a 700 KB unmatched-closer input exceeded 15 seconds in an earlier candidate; the complete-name candidate completed in 36–67 ms (main: 30–45 ms). A 640 KB deep-list input fell from about 2 seconds to 25–48 ms (main: 12–27 ms). Limits and assertions were not raised.

The separate `maxChars: 180` closing-wrapper truncation also occurs on main with plain visible HTML. Its output owner is unchanged; this PR does not claim to fix it.

Contributor ancestry and credit from @ruel225 are preserved.

### DIFF
--- a/extensions/web-readability/web-content-extractor.test.ts
+++ b/extensions/web-readability/web-content-extractor.test.ts
@@ -64,6 +64,69 @@ describe("web readability extractor", () => {
     expect(extracted.title).toBe("Example Article");
   });
 
+  it.each(["text", "markdown"] as const)(
+    "excludes double-escaped script prose in %s mode",
+    async (extractMode) => {
+      const extractor = createReadabilityWebContentExtractor();
+      const result = await extractor.extract({
+        html: `<!doctype html><html><head><title>Cloud Costs</title></head><body><article>
+<script><!--<script></script><p>Secret script paragraph. These words belong only to script data, and they must never become the article text returned to the reader. The paragraph has ordinary prose and no links or unusual attributes.</p>--></script>
+<p>Visible sibling. Cloud costs depend on the services a team uses, the amount of data it stores, and the traffic its customers generate. A useful review starts with the current bill and compares each service with the work it supports. Teams can then remove unused resources, select suitable instance sizes, and set budgets for expected growth. Regular reviews make these decisions easier because each change has a clear purpose and a measured result. This visible article paragraph must remain available after extraction.</p>
+</article></body></html>`,
+        url: "https://example.com/article",
+        extractMode,
+      });
+      const extracted = requireReadabilityResult(result);
+      expect(extracted.text).toContain("Visible sibling");
+      expect(extracted.text).not.toContain("Secret script paragraph");
+      expect(extracted.title).toBe("Cloud Costs");
+    },
+  );
+
+  it.each(["text", "markdown"] as const)(
+    "does not join script delimiters across comments in %s mode",
+    async (extractMode) => {
+      const extractor = createReadabilityWebContentExtractor();
+      const result = await extractor.extract({
+        html: SAMPLE_HTML.replace(
+          "<article>",
+          "<article><script></scr<!-- -->ipt><p>Secret script paragraph. This prose belongs to script data and must remain excluded from extracted article text.</p></script>",
+        ),
+        url: "https://example.com/article",
+        extractMode,
+      });
+      const extracted = requireReadabilityResult(result);
+      expect(extracted.text).toContain("Main content starts here");
+      expect(extracted.text).not.toContain("Secret script paragraph");
+      expect(extracted.title).toBe("Example Article");
+    },
+  );
+
+  it("preserves visible content after a quoted script closer", async () => {
+    const extractor = createReadabilityWebContentExtractor();
+    const result = await extractor.extract({
+      html: SAMPLE_HTML.replace("<article>", '<article><script>x</script data-note="<!--">'),
+      url: "https://example.com/article",
+      extractMode: "text",
+    });
+    const extracted = requireReadabilityResult(result);
+    expect(extracted.text).toContain("Main content starts here");
+    expect(extracted.title).toBe("Example Article");
+  });
+
+  it("preserves JSON-LD article titles while filtering script text", async () => {
+    const extractor = createReadabilityWebContentExtractor();
+    const result = await extractor.extract({
+      html: `<!doctype html><html><head><title>Site home</title><script type="application/ld+json">{"@context": "https://schema.org", "@type": "Article", "headline": "Cloud Spending Review"}</script></head><body><article><p>Visible sibling. Cloud costs depend on the services a team uses, the amount of data it stores, and the traffic its customers generate. A useful review starts with the current bill and compares each service with the work it supports. Teams can then remove unused resources, select suitable instance sizes, and set budgets for expected growth. Regular reviews make these decisions easier because each change has a clear purpose and a measured result. This visible article paragraph must remain available after extraction.</p>
+</article></body></html>`,
+      url: "https://example.com/article",
+      extractMode: "text",
+    });
+    const extracted = requireReadabilityResult(result);
+    expect(extracted.text).toContain("Visible sibling");
+    expect(extracted.title).toBe("Cloud Spending Review");
+  });
+
   it("does not count void tags toward the nesting limit", async () => {
     const extractor = createReadabilityWebContentExtractor();
     const html = SAMPLE_HTML.replace("<article>", `<article>${"<BR>".repeat(3100)}`);
--- a/src/agents/tools/web-fetch-html-tag.ts
+++ b/src/agents/tools/web-fetch-html-tag.ts
@@ -63,15 +63,11 @@ function startsWithClosingTag(html: string, start: number, tagName: string): boo
   return isTagBoundary(html[start + 2 + tagName.length]);
 }
 
-export function readRawTextOpenTagName(
-  html: string,
-  start: number,
-  tagNames: ReadonlySet<string> = RAW_TEXT_TAGS,
-): string | undefined {
+export function readRawTextOpenTagName(html: string, start: number): string | undefined {
   if (html[start] !== "<" || html[start + 1] === "/") {
     return undefined;
   }
-  for (const tagName of tagNames) {
+  for (const tagName of RAW_TEXT_TAGS) {
     let matches = true;
     for (let offset = 0; offset < tagName.length; offset += 1) {
       if (asciiLower(html[start + 1 + offset] ?? "") !== tagName[offset]) {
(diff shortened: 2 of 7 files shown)


obviyus:
Thanks @ruel225!

**Proof:** Public agent calls through production `web_fetch` now retain the original article and later list content while excluding hidden descendants. The latest independent checks also preserve visible children inside ordinary metadata-like names through Readability and basic extraction.

- **Bug:** Harmless attribute text and omitted endings could discard visible content; malformed tag identities could release hidden content.
- **Fix:** The shared scanner preserves complete names, and visibility filtering tracks the actual open-element scope.
- **Validation:** 253 focused tests, ordinary lint and repository guards, independent source/security and behavior checks, exact-head ClawSweeper, and [hosted CI](https://github.com/openclaw/openclaw/actions/runs/34544401391) passed.

| Before | After |
| --- | --- |
| The article or later list item disappears. | Visible article and sibling text survive. |
| A shortened tag name can release a hidden owner. | Complete names preserve hidden ownership. |

Merged as b8e2dbc7ca4237c18a82b274562ba6447d0d88a0. Contributor credit is preserved. The separate small-limit closing-wrapper issue remains outside this repair.

PR #144211, opened 2026-09-10
refactor: reuse account setup in Twilio send tests

steipete:
Related: #139428

## What Problem This Solves

Three Twilio send tests repeat complete account fixtures despite an existing local account helper, obscuring the few sender and callback settings that distinguish each case.

## Why This Change Was Made

Reuse the existing helper at those three sites. Keep both sender fields explicit in sender-choice cases and retain the callback URL's existing retry-policy fragment. Send options, mock fetch functions, response timing, and all expected outputs remain unchanged.

## User Impact

No production behavior change. This removes 40 net test lines while retaining all existing SMS test cases and assertions.

## Evidence

The complete existing Twilio test file passed before and after: 68 cases in identical order, zero failures or skips. Reconstructing the original three literals yields the exact original file. All 13 ordered account fields match, and the unchanged helper produces distinct accounts and allowlist arrays for each call.

Only the existing mocked transport suite ran; no live SMS or Twilio API calls were made. This cleanup does not claim a measured runtime improvement.

The mapped changed-file gate passed, including extension-test typechecking, lint, formatting and guards. Fresh independent review found no actionable P0–P2 findings.

### DIFF
--- a/extensions/sms/src/twilio.test.ts
+++ b/extensions/sms/src/twilio.test.ts
@@ -289,21 +289,9 @@ describe("Twilio SMS helpers", () => {
 
     await expect(
       sendSmsViaTwilio({
-        account: {
-          accountId: "default",
-          enabled: true,
-          accountSid: "AC123",
-          authToken: "secret",
-          fromNumber: "+15557654321",
-          messagingServiceSid: "",
-          defaultTo: "",
-          webhookPath: "/webhooks/sms",
+        account: createAccount({
           publicWebhookUrl: "https://gateway.example.com/webhooks/sms#rp=4xx",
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a9fadfa62692b91f4844cebf3e85286d586343c1](https://github.com/openclaw/openclaw/pull/144211/commits/a9fadfa62692b91f4844cebf3e85286d586343c1)
- Landed commit: [4e46815d68d4df27b136c4b5afee2eae6de2d659](https://github.com/openclaw/openclaw/commit/4e46815d68d4df27b136c4b5afee2eae6de2d659)

PR #144215, opened 2026-09-10
refactor(imessage): remove unused persisted echo mirror

### LINKED ISSUE #144213: Remove redundant persisted iMessage echo state
The iMessage echo module keeps an in-memory projection of persisted entries even though every lookup reloads the persisted store. Registering an outgoing echo first lists and rebuilds that unused projection, and deletion filters it again.

This maintainer-requested cleanup removes the redundant projection while keeping the store as the owner of registration, matching, TTL and bounded history. The same 166 existing tests pass before and after, and the selected changed gate and independent review pass. Removing an incidental preread can alter which operation first logs a persistence failure; no live iMessage, RSS or latency improvement is claimed.

steipete:
Closes #144213

## What Problem This Solves

Recording an outgoing iMessage echo currently reads and rebuilds an in-memory copy of persisted echo entries, even though every echo lookup immediately reloads the store. This maintainer-requested cleanup removes that redundant state and work.

## Why This Change Was Made

The persisted store remains the owner. Each lookup reads the existing live, ordered, bounded entries; remember and forget use their existing store operations without rebuilding an unused mirror. Matching, TTL, pending-entry reconciliation, persistence schema and retention policy are unchanged.

## User Impact

Echo registration avoids an unused store listing and array rebuild, and the module no longer retains a duplicate entry array. Removing the incidental preread can change which operation first emits a persistence warning, or remove that unused read warning; the retained operations keep their existing failure handling. No RSS, latency, or live iMessage performance claim is made.

## Evidence

The same 166 tests across three existing suites passed before and after. The stale-read regression retains its assertions with a behavior-focused title, and the managed P2 review is clean. The full selected changed gate passed. The proof uses isolated fixtures rather than a live iMessage account; test-runtime differences are not a performance benchmark.

### DIFF
--- a/extensions/imessage/src/monitor/monitor-provider.echo-cache.test-support.ts
+++ b/extensions/imessage/src/monitor/monitor-provider.echo-cache.test-support.ts
@@ -259,7 +259,7 @@ describe("iMessage sent-message echo cache", () => {
     expect(cache.has(scope, { messageId: "guid-late" })).toBe(true);
   });
 
-  it("drops the in-memory mirror on persisted read failure so expired echoes do not match", () => {
+  it("does not match stale echoes after persisted read failure", () => {
     vi.useFakeTimers();
     vi.setSystemTime(new Date("2026-02-25T00:00:00Z"));
     const scope = "acct:imessage:+1555";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6e7e8bcb999cb28c8a7664f8b09a2911da893597](https://github.com/openclaw/openclaw/pull/144215/commits/6e7e8bcb999cb28c8a7664f8b09a2911da893597)
- Landed commit: [2a780e2371d539a69845c407f65ab84c3871c041](https://github.com/openclaw/openclaw/commit/2a780e2371d539a69845c407f65ab84c3871c041)

PR #144217, opened 2026-09-10
test: cover real run ownership in queued compaction

steipete:
Related: #143426, #143725

## What Problem This Solves

The queued-compaction resource regression replaced RUN acquisition with a read-only acquisition adapter. That bypassed the production selection callback and RUN owner even after ordinary RUN registration ownership landed.

## Why This Change Was Made

Remove the obsolete 23-line module override so the existing test exercises actual RUN acquisition and its selection options. Keep the fixture and all lifetime assertions unchanged.

## User Impact

No runtime behavior change. The regression now covers the production path that must retain plugin resources through deferred engine maintenance and disposal.

## Evidence

The original and updated native integration each passed. The test executes actual queued compaction with a synthetic managed donor engine and on-disk SQLite, verifies reads during maintenance and disposal, and checks once-only closure and reopening. It uses no replacement for the RUN acquisition function. This is a native source integration, not a compiled distribution or external-provider test.

Required changed-file checks passed, including types, all three dead-export scans, lint, database guards, and runtime import-cycle checks. Independent Codex review found no actionable P0–P2 findings. No production files or assertions changed.

Canonical test runs took 27.41 seconds before and 28.66 seconds after, using separate cold transform caches. These are observations of the test harness, not a performance improvement claim.

### DIFF
--- a/src/agents/embedded-agent-runner/compact.queued-resources.test.ts
+++ b/src/agents/embedded-agent-runner/compact.queued-resources.test.ts
@@ -16,29 +16,6 @@ import { closePreparedModelRuntimeSnapshots } from "../prepared-model-runtime.li
 import { compactEmbeddedAgentSession } from "./compact.queued.js";
 import { waitForDeferredTurnMaintenanceForSession } from "./context-engine-maintenance.js";
 
-// Exercise the managed producer at the existing input boundary without enabling RUN disposal.
-vi.mock("../prepared-model-runtime.js", async (importOriginal) => {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [62efb554d099cfea1cac0753b171e8694f1e17bc](https://github.com/openclaw/openclaw/pull/144217/commits/62efb554d099cfea1cac0753b171e8694f1e17bc)
- Landed commit: [744f7a3067972d45b1694ceee620029880cba179](https://github.com/openclaw/openclaw/commit/744f7a3067972d45b1694ceee620029880cba179)

PR #143716, opened 2026-09-10
refactor: share isolated and TTS cleanup ownership

steipete:
Related: #130706, #142100.

## What Problem This Solves

Isolated completions and TTS summaries separately implement the same resource-retention lifecycle. Keeping those copies aligned makes cleanup repairs harder, particularly when a result is available before owned transport work has drained.

## Why This Change Was Made

Move the existing isolated-completion resource owner into the shared async-work module and reuse it for TTS summaries. Preserve logical result timing, parent cancellation, context capture, and release after owned work drains. Model selection and the injected TTS path are unchanged. This maintainer-requested cleanup removes 29 production lines ahead of the remaining model-runtime repairs.

## User Impact

Completion and TTS behavior stays the same, with one owner for resource cleanup.

## Evidence

All 59 existing focused cases pass with no test edits. Independent P2 review and a fresh full build passed. Six compiled loopback cases passed: isolated completion and TTS each cover success, authentication error, and cancellation. Tracked cleanup reads the live synthetic SQLite resource before the fixture owner closes it, and the import guard observed no checkout-source imports. Actual ephemeral registry disposal remains covered by the source suite. The complete changed-file gate passed.

The build was produced from the frozen staged candidate before commit; its natural build metadata retains the base revision. Commit `eca5e03bf0aa30597b04b57b2ccfb33d54eedbb7` has the identical built source tree; build metadata was not restamped.

[Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34439962212) passed.

PR #144220, opened 2026-09-10
refactor: share resource cleanup for side questions

steipete:
Related: #143716, #143366

## What Problem This Solves

BTW side questions maintained their own resource-cleanup loop alongside the shared owner already used by TTS and isolated completion. Keeping those implementations aligned required duplicating cancellation, generation-context and late-work handling.

## Why This Change Was Made

Use the existing shared resource helper while acquiring the same published generation and capturing its context before execution. Remove 24 production lines without changing model/harness selection, response behavior, or standalone/configured publication policy.

## User Impact

No intended behavior change. Side questions continue to retain their selected resources through actual provider and harness cleanup, including when the parent closes during acquisition.

## Evidence

- All 82 focused BTW and native resource cases passed on both the original production code and the refactor. One added table case verifies cancellation during acquisition handoff, exact generation, late SQLite reads, replacement and once-close/reopen. The existing managed-source input adapter is retained; the fixture now validates canonical config.
- Full changed-file gate passed, including types, all three dead-export scans, lint/format, database guards and zero runtime import cycles. Independent Codex review found no actionable P0–P2 issues. Ordinary build passed in 264.79s.
- Compiled synthetic proof passed in 2.51s through the real BTW entry point, with parent cancellation, held harness cleanup and normal cleanup context. Ordinary standalone publication remains raw. A separate compiled owned-publication control verifies replacement/disposal; it is not presented as default managed BTW disposal. The native fixture covers that managed composition.
- All 40,717 tracked source files and 14,628 artifacts matched their seals after proof; no TypeScript fallback or external inference was used. The successful configuration passed the built schema, and its full log contained no errors or fallback diagnostics. An earlier probe stopped before application execution because its driver mistook the schema object for a function; that preflight check was corrected and the failure retained.
- Observed test runs were 41.74s/2.26 GB peak RSS before and 31.32s/2.28 GB after. Compiled cold admission was 273.84ms and warm result plus cleanup signal was 3.34ms. These are individual synthetic observations, not a benchmark or speedup claim.

Hosted CI passed on the final head: [run 34506348766](https://github.com/openclaw/openclaw/actions/runs/34506348766). Two unrelated fixture failures cleared in targeted unchanged recoveries: iMessage dispatch polling and frozen-target subprocess admission checks. A bounded diagnostic replay preserved the latter's 20-second deadline and verified the expected missing-blob refusal; no production code, test assertions, or deadlines were changed for CI recovery.

### DIFF
--- a/src/agents/btw.ts
+++ b/src/agents/btw.ts
@@ -1,4 +1,3 @@
-import { AsyncLocalStorage } from "node:async_hooks";
 import { randomUUID } from "node:crypto";
 /**
  * Runs `/btw` side questions against the active conversation without resuming
@@ -28,12 +27,7 @@ import type {
 import { prepareProviderRuntimeAuth } from "../plugins/provider-runtime.js";
 import { withPluginRuntimeGenerationScope } from "../plugins/runtime/generation-scope.js";
 import { isModelSelectionLocked } from "../sessions/model-overrides.js";
-import {
-  AsyncWorkScope,
-  captureAsyncWorkTracker,
-  getAsyncWorkSignal,
-} from "../shared/async-work-scope.js";
-import { createDeferredCore } from "../shared/deferred.js";
+import { runWithAsyncWorkResources } from "../shared/async-work-resources.js";
 import { prepareSystemAgentRunAdmission } from "./admitted-run-context.js";
 import { resolveAgentWorkspaceDir } from "./agent-scope.js";
 import { resolveExternalCliAuthOverlayScopeFromSelection } from "./auth-profiles/external-cli-auth-selection.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f550823bb3846785be7196b81f55b12b4737296f](https://github.com/openclaw/openclaw/pull/144220/commits/f550823bb3846785be7196b81f55b12b4737296f)
- Landed commit: [0a4447ecc53e24449cbc2b8c2f9725faeee1213a](https://github.com/openclaw/openclaw/commit/0a4447ecc53e24449cbc2b8c2f9725faeee1213a)

PR #143451, opened 2026-09-09
fix(sessions): identify slow SQLite writer operations

steipete:
Related: #143434

## What Problem This Solves

Resolves a problem where operators investigating slow SQLite session writes could see timing but could not identify which operation owned the writer lane. Failed writes had the same gap, especially when background work had no request trace.

## Why This Change Was Made

Require a closed semantic operation category at every internal writer call and include it in the existing slow/failure warning. Prepared writes carry the same whole-operation category through inline and split commits; existing reclamation fields remain intact. Queue labels, errors, timing boundaries, thresholds, checkpoint modes, persistence, and audit policy are unchanged.

The required argument touches 51 production call sites and their test fixtures. Most diff volume is formatter reindentation; a token comparison confirmed that all 51 callback bodies, scopes, and existing reclamation arguments are unchanged. This internal helper has no Plugin SDK consumers.

## User Impact

Operators can locate the owning write path using the warning’s `operation` field. The category does not identify a particular SQL statement, CPU use, lock contention, or a causal relationship with a nearby RPC. Older records remain unattributed.

## Evidence

- On the original source, five existing warning/file-log cases and all eight real-owner cases fail specifically because operation labels are missing; the cleanup control passes. On the candidate, all 14 warning/owner cases pass.
- Selected SQLite admission/reclamation controls: 77 passed. Rewind storage and authority controls: 25 passed. Total: 116 unique focused cases.
- Full `node scripts/check-changed.mjs` passed. Independent P0–P2 review found no actionable findings.
- [Linux Testbox proof](https://github.com/openclaw/openclaw/actions/runs/34414798580): the native command exited 0 after `pnpm build`, all 14 warning/owner tests, and the built CLI/Gateway health → create → patch → describe → delete → repeat-delete → health flow. Exact source/tree/build identities and all 27 producer artifact hashes were verified; the build owner and nine fixture owners joined and released. The lease was stopped after audit. This is a model-free functional and diagnostic proof.

PR #138669, opened 2026-09-04
perf(sessions): move disk accounting off the Gateway thread

steipete:
Builds on the merged queue-admission repair #138604 (`e270ccbb2f279d63cda71f3d4fb946b0b8a8bf1e`; final CI 108 success/17 skipped). This PR now targets `main`.

## What Problem This Solves

Fixes an issue where users would experience Gateway stalls while automatic session disk accounting scanned large artifact directories. Each eligible file performed synchronous canonical-path work on the caller thread; bounding task admission reduced allocation pressure but left the filesystem scan blocking that thread.

## Why This Change Was Made

Run physical measurement in one reusable worker and return its four byte counters. The shared inventory owner retains the original sequence: directory inventory → prompt blobs → durable SQLite ownership resolution → database/WAL metadata → canonical-path deduplication and totals. Pruning callers reuse the same inventory functions. No configuration, schema, retention, path-canonicalization, or deletion policy changes.

Maintainer disposition: accept an omitted `timeoutMs` for disk measurement, whose previous implementation had no deadline. Supplied deadlines retain their existing normalization and shared queue/preparation/execution budget; abort, pool-close and cleanup behavior remain unchanged. Stable `v2026.9.1` (`ad6fe23a`) contains the pool internally but exposes no public SDK/package export for it; the SDK export arrived later in #138094 (`ea3e77de`). This extends an unshipped SDK surface, without treating undocumented deep imports as a supported stable API. Worker failures remain visible, and the packaged worker is registered in the existing build and release inventory.

Production and tooling LOC: **+207/-151, net +56**, including one Knip entry; tests **+242/-41, net +201**. Most code moves into the shared inventory owner. The remaining growth provides the worker/runtime boundary and optional deadline, replacing caller-thread measurement. The compaction fixture repair removes four net test lines.

## User Impact

Large physical-usage scans can finish without monopolizing the Gateway event loop. Cleanup still measures real bytes, preserves canonical aliases, and remeasures after reclamation. Existing maintenance commands and settings remain applicable.

## Evidence

[...]

PR #144224, opened 2026-09-10
fix(sessions): diagnose slow archive-pruning stages

steipete:
Related: #143379, #143451, #138669.

## What Problem This Solves

Resolves a problem where operators investigating slow or failed session-history archive pruning could see only the writer's total duration. That duration includes database admission, synchronous checkpoint/vacuum work, file operations, and asynchronous physical-disk measurement, so it could not identify which stage required investigation.

## Why This Change Was Made

The existing slow/failure writer warning now includes one bounded `archivePruning` object, with a fixed `initial`, `after-eviction`, or `final` trigger and aggregate stage durations and outcome counters. Timings survive failures and use a fixed serialization whitelist. Existing operations supply every observation; no additional store reads or per-file records are introduced.

Checkpoint mode and timeout, the initial-freelist/512-page vacuum bound, writer and lifecycle ownership, archive selection, file-before-row deletion, errors, and retention behavior remain unchanged. This improves diagnosis; it does not fix checkpoint blocking or assign CPU time from wall-clock durations.

## User Impact

Operators can distinguish checkpoint stalls from awaited measurement or admission delays in the existing warnings. Documentation explains incomplete checkpoint results, requested vacuum pages, partial observations, and why normal completion does not guarantee the high-water target was reached. No configuration is required.

## Evidence

- The pre-change owner/log regression run failed five new assertions because diagnostic records were absent; the other 25 tests passed. The implemented change passes all 31 focused tests, including real maintenance callsites, cold-admission revocation, incomplete checkpoint returns, and file-log whitelisting. A real warm-cache wrong-owner refusal additionally reproduced a fabricated async-admission count on the first revision; the correction preserves refusal timing while counting only explicitly observed admission modes.
- A bounded synthetic full-owner run with an independent 1.5-second reader reported 1,587 ms of checkpoint time; independent native measurements totaled 1,587.300 ms. The queued real session mutation waited 1,604 ms. Both cases drained 64 free pages, persisted the mutation, and passed integrity/FK checks. The control separately exposed 2,619 ms awaiting physical measurement with only 21 ms in checkpoints; that measurement is not claimed to be CPU or scan time.
- Additional legacy-removal/history-eviction tests: 40 passed; six lifecycle setup cases failed in unchanged config-write admission because this host has an incompatible shared managed-handoff lease. The external lease was preserved. These failures occurred before the lifecycle attempts. All six unchanged assertions subsequently passed on the isolated Linux lease at the first candidate (28 unrelated cases skipped); the local lease remains untouched.
- Isolated P2 review: no actionable findings.
- Required `node scripts/check-changed.mjs` passed, including formatting, core/core-test typechecks, lint, unused exports, storage guards, and runtime import-cycle checks.
- Final candidate `ee1973de0bb8f278017c682f8637324b13bdcef6` passed the actual full-owner replay on isolated Linux Blacksmith Testbox (Ubuntu 24.04.3, Node 24.19.0): precise checkpoint total 1,507.370 ms versus independent native total 1,507.338 ms, successful queued mutation, integrity/FK checks, and 64-to-zero free-page reclamation. Tree `d6312ac8ae3f6413f6888617ca94f0199a000ab3` and all 22 source fingerprints matched. The new warm-cache refusal regression also passed on Linux (1 selected, 13 unselected). The task-owned leases are stopped and independently confirmed complete. Lease provisioning: https://github.com/openclaw/openclaw/actions/runs/34507630564.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -458,6 +458,7 @@ The object aggregates observations across the pruning pass:
 - `admissionMs`, `cachedAdmissions`, and `asyncAdmissions` measure database
   acquisition and count its observed modes. Admission time ends at callback entry
   or acquisition failure and can include shared admission and integrity-check waits.
+  A refusal before mode selection adds admission time without incrementing either mode count.
 - `checkpointMs`, `checkpointMaxMs`, and `checkpointCalls` report total time,
   longest call, and calls entered. `checkpointIncomplete` counts calls returning
   false, which can mean a busy checkpoint or an error; it does not identify a lock
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -448,6 +448,50 @@ the warning threshold or prove that a nearby request caused the work. Rounding
 and work outside the measured subphases can leave a difference from
 `writerExecutionMs`; do not assign that remainder to a specific phase.
 
+For `session.history.archive-prune`, the same slow or failure warning can include
+one bounded `archivePruning` object. Its `trigger` is recorded at the call site:
+`initial`, `after-eviction`, or `final`. It distinguishes pruning passes within
+the maintenance flow; it does not identify the request that caused maintenance.
+
+The object aggregates observations across the pruning pass:
+
+- `admissionMs`, `cachedAdmissions`, and `asyncAdmissions` measure database
+  acquisition and count its observed modes. Admission time ends at callback entry
+  or acquisition failure and can include shared admission and integrity-check waits.
+- `checkpointMs`, `checkpointMaxMs`, and `checkpointCalls` report total time,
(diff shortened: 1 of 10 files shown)


steipete:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [ee1973de0bb8f278017c682f8637324b13bdcef6](https://github.com/openclaw/openclaw/pull/144224/commits/ee1973de0bb8f278017c682f8637324b13bdcef6)
- Landed commit: [dbf9d706b58f0d1bc8cf3e800cf00b1085a32eef](https://github.com/openclaw/openclaw/commit/dbf9d706b58f0d1bc8cf3e800cf00b1085a32eef)

PR #144227, opened 2026-09-10
feat(ui): show reported Gateway isolation

ChazGo:
> [!IMPORTANT]
> **Draft dependency:** This PR remains draft until Agent Session support is checked into `openclaw/openclaw-windows-packaging` and the corresponding Windows packaging integration is publicly available.

## What Problem This Solves

The Control UI does not currently show the Gateway isolation posture reported by the running Gateway. Operators therefore cannot confirm from Settings whether Gateway isolation is enabled or identify the command needed to change it.

## Why This Change Was Made

This adds an optional `gatewayIsolation` field to `system.info`, reports the launch-time `OPENCLAW_GATEWAY_ISOLATION` posture when it is valid, and renders that state in Connection settings with a copyable `clawctl gateway-isolation` command. The field is optional: when it is absent, the new section is not rendered, preserving existing behavior on all current platforms and Gateway versions.

## User Impact

Operators using a Gateway that reports isolation can see whether it is enabled and copy the appropriate command to change it. Platforms and Gateway versions that do not report the optional field remain unchanged.

## Evidence

Focused tests were added, but local Vitest, tsgo/changed checks, UI i18n verification, and browser proof could not run because the worktree package-feed proxy returned 404s and public npm TLS failed. Oxfmt and `git diff --check` passed. Independent review found no material issue.

PR CI and screenshot are pending.

### DIFF
--- a/packages/gateway-protocol/src/schema/system-info.test.ts
+++ b/packages/gateway-protocol/src/schema/system-info.test.ts
@@ -34,6 +34,24 @@ describe("SystemInfoResultSchema", () => {
     expect(Value.Check(SystemInfoResultSchema, validSystemInfo)).toBe(true);
   });
 
+  it.each(["enabled", "disabled"] as const)(
+    "accepts the optional Gateway isolation state %s",
+    (gatewayIsolation) => {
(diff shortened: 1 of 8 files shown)


ChazGo:
Superseded by #144243, which recreates this change as one clean commit directly on current official main.

PR #144232, opened 2026-09-10
fix(sessions): speed up database cleanup and find transcript archives consistently

### LINKED ISSUE #144212: Session cleanup repeatedly validates large databases and miscounts archive storage
## What Problem This Solves

Running `openclaw sessions cleanup --enforce` against a large session database can spend most of its time repeatedly validating the entire database for individual historical transcripts. Using `--store` with an explicit SQLite path can also report incorrect disk usage and miss archive artifacts.

## Reproduction and Expected Behavior

1. Prepare an isolated agent database with many unreferenced historical transcript windows and configure a disk budget below its current physical size.
2. Run local `sessions cleanup --store <state>/agents/main/agent/openclaw-agent.sqlite --enforce`.
3. Compare disk usage and artifact cleanup with the equivalent `<state>/agents/main/sessions/sessions.json` selector.

Local cleanup should retain its integrity, protection, and transaction checks without reopening and fully validating the entire database for each victim. Equivalent selectors must measure and prune the same archive directory.

The current path can create a fresh writable reclamation Worker for every victim. The explicit SQLite selector also scans `agent/` even though archive publication uses sibling `sessions/`, counting SHM or unrelated files while missing actual archives and prompt blobs.

## Evidence

Focused regressions reproduce the repeated Worker execution and selector mismatch. Filesystem fixtures also demonstrate the mismatch for a legacy `agent/sessions.json` selector resolving to the same physical database. A scoped repair and live Gateway proof are being prepared.

steipete:
Closes #144212

## What Problem This Solves

Fixes an issue where users cleaning large session databases would wait through repeated whole-database validation for individual historical transcripts. Explicit SQLite store selectors could also measure or prune the wrong artifact directory, missing archives and prompt blobs while counting SQLite shared-memory files.

## Why This Change Was Made

Local CLI cleanup now reuses its admitted database connection through the existing in-process reclamation path. Gateway and background cleanup retain Workers, and the existing transaction, snapshot, admission, retention, and per-candidate disk checks remain. Archive staging batches small UTF-8 JSONL writes, and successful archive INSERTs skip redundant blob rereads while preserving conflict verification.

A shared directory resolver aligns measurement, pruning, preview exclusions, and archive publication for equivalent legacy and SQLite selectors. Selected SQLite SHM files are excluded from budget totals. Documentation clarifies archive storage, copied-state validation, and Doctor's explicit agent selector.

## User Impact

Local cleanup avoids a full integrity scan for every historical victim and consistently finds the owning session artifacts. Existing session protection and retention behavior is preserved. This PR does not add file-only cold storage, new configuration, or a schema migration.

## Evidence

- 92 focused archive, historical cleanup, CLI, and artifact tests pass on the rebased branch. The final 21-test archive-worker run also passes, including the added native custom-directory publication regression. Regressions were demonstrated failing before their fixes.
- Native Gateway live proof using isolated synthetic state: `sessions.cleanup` reclaimed exactly one historical window, retained the current session, and reduced measured storage from 2,732,032 to 667,648 bytes. The current transcript remained byte-identical and visible through `sessions.preview`; SQLite integrity and foreign-key checks passed.
- All 50 HTTP health probes during the live exercise succeeded (local p95 6.9 ms, maximum 72.2 ms). The task-owned Gateway was shut down afterward.
- Offline cleanup and Doctor compaction were also verified on an isolated large SQLite snapshot. Retained transcript hashes and logical-session metadata matched the untouched baseline; protected history correctly prevented further cleanup when no eligible victims remained.
- A synthetic 50,000-row staging benchmark produced identical output while reducing file writes from 100,000 to 190. Median staging-only time fell from 287.7 ms to 17.1 ms; this is not an end-to-end speed claim.
- Full changed-file checks, targeted final lint/formatting, and the documentation link/anchor audit pass (13,535 internal links; zero broken links). Final independent review is clean at P0–P2.

## Stored-data compatibility

No migration is required: this PR leaves the agent/state DDL and schema versions (19/17), archive names, encoding, decoder, and reclamation transaction format unchanged. The serialized-state flag reflects a writer implementation change, not a stored-format change.

Fresh cross-version proof compared the exact pre-change and PR-head staging, zstd encoding, persistence, and decoder implementations using 1,803 synthetic rows (675,176 JSONL bytes), including Unicode, escaping, preserved whitespace, and a row larger than 64 KiB. Both writers produced identical plaintext and compressed hashes (5,154 compressed bytes). All four old/new writer-reader combinations passed plain-file, compressed-file, and SQLite-blob round trips; duplicate writes were accepted and conflicting same-generation blobs rejected in both versions.

The existing exact-byte archive-worker and conflict rollback tests remain regression coverage. Earlier native Gateway and copied-database validation also preserved retained transcript bytes and session metadata. The separate file-only archival/schema proposal is not included in this PR.


steipete:
Merged via squash.

- Prepared head SHA: [e49a09117964fe53f057b6f7f9492b193aa8f212](https://github.com/openclaw/openclaw/pull/144232/commits/e49a09117964fe53f057b6f7f9492b193aa8f212)
- Landed commit: [d748e5d93d8d63df9d90dbb4150c003f48d7e450](https://github.com/openclaw/openclaw/commit/d748e5d93d8d63df9d90dbb4150c003f48d7e450)

PR #144234, opened 2026-09-10
refactor(agents): remove unused pending message state

### LINKED ISSUE #144233: Remove unused pending messaging projections from agent subscriptions
Agent subscriptions still allocate and write pending messaging text, target and media maps even though completion derives confirmed delivery evidence from hook-adjusted arguments. The maps have no remaining consumer.

Remove the redundant projections while retaining captured start context, successful/partial delivery facts, source replies and failure handling. Baseline 259 tests and candidate 258 tests pass after removing the pending-only case; the full selected gate and independent review pass, including exact shrink-only assertion-baseline maintenance. No measured allocator/RSS/latency gain is claimed. Related: #93228.

steipete:
Closes #144233

## What Problem This Solves

Agent tool subscriptions still allocate and populate three pending messaging maps whose data is no longer consumed. Completion now derives delivery evidence from the hook-adjusted arguments, leaving these start-time projections as unused work and retained state.

## Why This Change Was Made

Remove the write-only maps, their duplicate start-time extraction, and matching reset/delete plumbing. Keep the captured start context, including hasRepliedRef, and the existing completion owner for confirmed text, target, media and source-reply evidence. Tests now assert the retained completion behavior instead of the retired map representation.

## User Impact

Subscriptions carry less unused state while preserving successful/partial delivery facts, failed-send behavior, duplicate suppression and source replies. This is a maintainer-requested cleanup; no measured allocation, RSS or latency improvement is claimed.

## Evidence

The baseline passed 259 tests; the candidate passed 258 after removing the obsolete pending-only case. Retained completion, media and source-reply tests passed, including the normal runner's required source builds. The full selected changed gate and fresh independent P2 review passed. The initial gate required the exact assertion allowance to shrink from 8 to 7 after deleting an unused cast; only that row changed. Whole-run timing/RSS includes source builds and is not a performance comparison. Related: #93228, which moved completion evidence to the adjusted arguments.


steipete:
Merged via squash.

- Prepared head SHA: [b4b2851e7fa0af9b7ae3fd9a75bf2dea14e6fcd7](https://github.com/openclaw/openclaw/pull/144234/commits/b4b2851e7fa0af9b7ae3fd9a75bf2dea14e6fcd7)
- Landed commit: [f4a98cee0a72ce2f987d8947ad4eb7676680c393](https://github.com/openclaw/openclaw/commit/f4a98cee0a72ce2f987d8947ad4eb7676680c393)

PR #144238, opened 2026-09-10
fix: keep selected models in lightweight and worker inference

steipete:
Related: #130706

## What Problem This Solves

Fixes an issue where users running lightweight agent completions or worker inference could receive output from a different model when provider aliases form a chain. With `entry → middle → final`, selection chose `middle` but preparation sent the request to `final`. Worker results could still report `middle`.

## Why This Change Was Made

Initial preparation now preserves the model identity already chosen by the agent selector or worker approval path. Raw model inputs continue to normalize normally. The change forwards the existing model-source contract through simple completion preparation without changing worker admission or runtime ownership.

## User Impact

Lightweight completions and worker inference use the model that was selected, so the reported model and the actual request agree.

## Evidence

- The exact regression fails on the base with four wrong-model failures while both raw-input controls pass. The candidate passes all six cases and 18 actual loopback HTTP requests across provider API present/absent configurations. Worker models are explicitly retained configured fallbacks; this does not assert arbitrary dynamic-model admission.
- 110 focused tests passed, including agent defaults/utility selection, auth routing, raw calls, worker inference, and plugin-generation ownership. The final fixture was rerun on both base and candidate after its HTTP-handler lint correction.
- Six additional source scenarios ran on both versions through real acquisition and HTTP. They preserve existing scoped runtime normalization and exact API-owner suppression, and confirm that the cold runtime-only hook limitation is inherited.
- The complete `pnpm check:changed` gate, fresh P2 review, and one normal `pnpm build` passed.
- The emitted package facade passed six further HTTP requests on the exact committed build. Both processes exited naturally, source-import guards passed, and all 12,230 inventoried build files remained unchanged. Worker execution is covered by the real source harness; it has no package-exported compiled executor entry.

Production delta: +4 lines.

### DIFF
--- a/src/agents/simple-completion-runtime.selected-model.test.ts
+++ b/src/agents/simple-completion-runtime.selected-model.test.ts
@@ -1,0 +1,255 @@
+import { createServer } from "node:http";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { OpenClawConfig } from "../config/types.openclaw.js";
+import {
+  createWorkerInferenceExecutor,
+  type WorkerInferenceExecutionParams,
+} from "../gateway/worker-environments/inference-runtime.js";
+import { resetPluginLoaderTestStateForTest } from "../plugins/loader.test-fixtures.js";
+import { clearPluginMetadataLifecycleCaches } from "../plugins/plugin-metadata-lifecycle.js";
+import { withOpenClawTestState } from "../test-utils/openclaw-test-state.js";
+import { resetPreparedModelRuntimeSnapshotsForTest } from "./prepared-model-runtime.test-support.js";
+import {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [76097fb74f738512b807f8e9d4cfaf3ccc5b1bd6](https://github.com/openclaw/openclaw/pull/144238/commits/76097fb74f738512b807f8e9d4cfaf3ccc5b1bd6)
- Landed commit: [6a011b56dfd1911cd53e72deb238c5c6a0e6e275](https://github.com/openclaw/openclaw/commit/6a011b56dfd1911cd53e72deb238c5c6a0e6e275)

PR #144239, opened 2026-09-10
fix(ui): avoid extra session reads after a queued refresh

steipete:
Related: #144127

## What Problem This Solves

Filtered session lists could make an unnecessary extra request when a full refresh was queued behind an existing read. A delayed event already covered by the queued refresh could remain scheduled and cause another fetch.

## Why This Change Was Made

Consume earlier event invalidations at the start of each full replacement in the existing request drain. Initial and queued refreshes now use the same boundary. Events arriving after a read starts still request a trailing refresh, and pagination retains its pending work.

## User Impact

Session lists avoid redundant background reads while preserving later updates, page ownership and normal restoration behavior.

## Evidence

The regression varies an event immediately before versus after a queued replacement starts. Before the fix, the earlier-event case makes three requests instead of two; the later-event control correctly makes three. After the fix, all 77 owner/controller cases pass. Production moves three existing lines, with no net growth; tests add 76 lines.

The unchanged archived-restore browser scenario passes on fresh original and candidate bundles. The paired images compare the same completed restore state; they establish preserved appearance. The timing regression establishes the removed request. The separate five-versus-three CI failure encountered in #144127 remains unexplained and is not claimed fixed here.

The patch moved onto a main-based branch with all 22 relevant source inputs unchanged. Those tests and captures retain their original source binding; all 20 canonical checks were rerun successfully on the new base. Independent review found no actionable findings. Initial test-placement and receiver-binding lint failures were repaired without suppressions.

Baseline: archived restore completes with the expected appearance

Candidate: the same restored view is preserved

### DIFF
--- a/ui/src/lib/sessions/index-list.test.ts
+++ b/ui/src/lib/sessions/index-list.test.ts
@@ -1,7 +1,12 @@
+import { asNullableRecord } from "@openclaw/normalization-core/record-coerce";
 import { describe, expect, it, vi } from "vitest";
 import { SIDEBAR_SESSION_ROSTER_LIMIT } from "../../../../src/shared/session-list-limits.ts";
 import { createDeferred } from "../../../../test/helpers/promise.js";
-import { GatewayRequestError, type GatewayBrowserClient } from "../../api/gateway.ts";
+import {
+  GatewayRequestError,
+  type GatewayBrowserClient,
+  type GatewayEventFrame,
+} from "../../api/gateway.ts";
 import type { SessionsListResult } from "../../api/types.ts";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4607093ffcb240d0f71981bc44b14729b57b631a](https://github.com/openclaw/openclaw/pull/144239/commits/4607093ffcb240d0f71981bc44b14729b57b631a)
- Landed commit: [7a2f281b7ada16628945142f08b7d7b819f03048](https://github.com/openclaw/openclaw/commit/7a2f281b7ada16628945142f08b7d7b819f03048)

PR #144242, opened 2026-09-10
perf(build): skip hashes before mandatory cache restores

### LINKED ISSUE #144241: Skip destination validation for mandatory build cache restores
Mandatory build-cache restores validate destination contents that will be replaced regardless of the result. The cached source and input signature already own restore validity.

Skip only the destination read/hash on the always-restore path, retaining cached-source integrity validation, ordinary warm lookup checks and obsolete output pruning. The original targeted destination-read test fails as intended; all 96 final owner tests, selected gate and independent review pass. This maintainer-requested cleanup makes no aggregate build-time or RSS claim.

steipete:
Closes #144241

## What Problem This Solves

Build steps configured to always restore cached outputs still read and hash the destination files before overwriting them. That work cannot make the existing destination eligible for reuse.

## Why This Change Was Made

Skip destination validation for mandatory restores and keep the existing cached-source validation and input-signature checks. Ordinary warm lookups still validate their outputs, and restore still removes obsolete generated files.

## User Impact

Mandatory cache restores avoid unnecessary destination reads without changing cache integrity checks, restore eligibility or generated output cleanup. No broad build-time or memory improvement is claimed.

## Evidence

The extended existing restore test fails on the original code because the destination is read, then passes on the change. All 96 build/cache owner tests, the full selected changed gate and independent P2 review pass, retaining stale-cache and replaced-record integrity coverage. No new helper export or cache-source read-count assertion was added.

### DIFF
--- a/scripts/lib/build-artifact-cache.mts
+++ b/scripts/lib/build-artifact-cache.mts
@@ -403,7 +403,8 @@ export function resolveBuildStepCacheState(
     const relativeOutputFiles = outputFiles.map((file) => portableRelativePath(artifactRoot, file));
     const stampedOutputs = Object.keys(stamp?.outputs ?? {});
     const requiredOutputs = resolveCacheRequiredOutputs(step.cache, params.env ?? process.env);
-    const actualOutputsPresent =
+    const actualOutputsAcceptable =
+      step.cache.restore !== "always" &&
       artifactRecordMismatch(artifactRoot, stamp, signature, requiredOutputs) === undefined;
     const cachedOutputMismatch = artifactRecordMismatch(
       outputRoot,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4b82dc65d09606a196095e23788b47e1fd553bb2](https://github.com/openclaw/openclaw/pull/144242/commits/4b82dc65d09606a196095e23788b47e1fd553bb2)
- Landed commit: [9edbe0b152265843f48c0727e1f615434b55911d](https://github.com/openclaw/openclaw/commit/9edbe0b152265843f48c0727e1f615434b55911d)

PR #144243, opened 2026-09-10
feat(ui): show reported Gateway isolation

ChazGo:
> [!IMPORTANT]
> **Draft dependency:** This PR intentionally remains draft until Agent Session support is checked into `openclaw/openclaw-windows-packaging` and the corresponding Windows packaging integration is publicly available. The current public [packaging producer commit](https://github.com/ChazGo/openclaw-windows-packaging/commit/613cf63bf9edf2cf6b62bfa78e5ac2d159e39168) proves the strict `disabled` projection for its signed-in-user launch path with 55/55 solution tests, but `clawctl gateway-isolation` and the fail-closed `enabled` Agent Session launch are not public yet.

## What Problem This Solves

The Control UI does not currently show the **Gateway Isolation** posture reported by the running Gateway. Operators therefore cannot confirm from Settings whether **Reported Gateway Isolation** is enabled or identify the command needed to change it.

## Why This Change Was Made

This adds an optional `gatewayIsolation` field to `system.info`, reports the launch-time `OPENCLAW_GATEWAY_ISOLATION` posture only when it is exactly `enabled` or `disabled`, and renders a read-only **Gateway Isolation** section in Connection settings with the reported state and inverse `clawctl gateway-isolation` command. When the optional field is absent, the new section is not rendered, so existing platforms and Gateways remain unchanged.

## User Impact

Operators using a Gateway that reports isolation can see the reported posture and copy the appropriate host command. The UI adds no write RPC, process control, platform inference, session implementation, or behavior for unsupported Gateways.

## Evidence

The screenshot below was captured from the real Control UI built at exact head `7949dcbc451693f10986253931e04ebbe0edcdb9` by [`dist-runtime-build` artifact 10171223596](https://github.com/ChazGo/openclaw/actions/runs/34523610745/artifacts/10171223596) in [CI run 34523610745](https://github.com/ChazGo/openclaw/actions/runs/34523610745). The repository's mocked-Gateway browser harness advertised `system.info`, returned `gatewayIsolation: "enabled"`, observed the UI request, rendered the inverse command, and reported no browser errors.

Gateway Isolation enabled in Connection settings

### Exact-head real Gateway proof

The same exact head was hydrated as a complete Windows x64 runtime, installed on Windows with Node 26.7.0, and launched against isolated disposable state. Its runtime ZIP SHA-256 is `04973F8AE49A598CD50A6FDA262E5119396E6F06AD6AF04AA812397E10CCC009`. A real Control UI browser connected to each real Gateway over WebSocket and observed `system.info`.

| Launch environment | Observed UI | Clipboard result |
|---|---|---|
| `OPENCLAW_GATEWAY_ISOLATION=enabled` | **Enabled** with inverse disable command | `clawctl gateway-isolation disable` |
| `OPENCLAW_GATEWAY_ISOLATION=disabled` | **Disabled** with inverse enable command | `clawctl gateway-isolation enable` |
| Variable omitted | Gateway Isolation section absent | Not applicable |

No browser console or page errors were observed in any of the three runs.

Reported Gateway Isolation enabled from a real exact-head Gateway

Reported Gateway Isolation disabled from a real exact-head Gateway

This proves the exact-head environment-reader → `system.info` → Control UI → clipboard path. The environment values were injected directly for these branch-level runs; this evidence does **not** claim that the unfinished public Windows packaging command or Agent Session isolation is implemented or proven.

The exact-head run completed with 176 successful jobs and 6 skipped jobs. Six underlying jobs failed: the genuine environment-name ratchet at 494 names against the 493 budget, the strict generated-locale release check that current contributor policy intentionally excludes from source PRs while the deterministic source verifier passes, a Control UI performance comparison against the fork's unrelated stale `main`, and three unrelated full-matrix tests in release validation, session reclamation timing, and chat reset fixtures. [Explicit owner approval has been requested](https://github.com/openclaw/openclaw/pull/144243#issuecomment-5624775331) before changing the protected environment budget.

[ClawSweeper's exact-head review](https://github.com/openclaw/openclaw/pull/144243#issuecomment-5622957766) found no actionable correctness defect in the focused patch and kept the draft blocked on the public packaging integration, environment-contract approval, and real end-to-end host proof.

### DIFF
--- a/apps/shared/OpenClawKit/Sources/OpenClawProtocol/GatewayModels.swift
+++ b/apps/shared/OpenClawKit/Sources/OpenClawProtocol/GatewayModels.swift
@@ -18974,6 +18974,7 @@ public struct SystemInfoResult: Codable, Sendable {
     public let diskavailablebytes: Int?
     public let diskpath: String?
     public let disks: [[String: AnyCodable]]?
+    public let gatewayisolation: String?
     public let defaultagentutilitymodel: AnyCodable?
 
     public init(
@@ -18998,6 +18999,7 @@ public struct SystemInfoResult: Codable, Sendable {
         diskavailablebytes: Int? = nil,
         diskpath: String? = nil,
         disks: [[String: AnyCodable]]? = nil,
+        gatewayisolation: String? = nil,
         defaultagentutilitymodel: AnyCodable? = nil)
     {
         self.machinename = machinename
@@ -19021,6 +19023,7 @@ public struct SystemInfoResult: Codable, Sendable {
         self.diskavailablebytes = diskavailablebytes
         self.diskpath = diskpath
         self.disks = disks
+        self.gatewayisolation = gatewayisolation
         self.defaultagentutilitymodel = defaultagentutilitymodel
     }
 
@@ -19046,6 +19049,7 @@ public struct SystemInfoResult: Codable, Sendable {
         case diskavailablebytes = "diskAvailableBytes"
         case diskpath = "diskPath"
         case disks
+        case gatewayisolation = "gatewayIsolation"
         case defaultagentutilitymodel = "defaultAgentUtilityModel"
     }
 }
--- a/packages/gateway-protocol/src/schema/system-info.test.ts
+++ b/packages/gateway-protocol/src/schema/system-info.test.ts
@@ -34,6 +34,24 @@ describe("SystemInfoResultSchema", () => {
     expect(Value.Check(SystemInfoResultSchema, validSystemInfo)).toBe(true);
   });
 
+  it.each(["enabled", "disabled"] as const)(
+    "accepts the optional Gateway isolation state %s",
+    (gatewayIsolation) => {
+      expect(Value.Check(SystemInfoResultSchema, { ...validSystemInfo, gatewayIsolation })).toBe(
+        true,
+      );
+    },
+  );
+
+  it("rejects an invalid Gateway isolation state", () => {
+    expect(
+      Value.Check(SystemInfoResultSchema, {
+        ...validSystemInfo,
+        gatewayIsolation: "unknown",
+      }),
+    ).toBe(false);
+  });
+
   it("rejects extra properties", () => {
     expect(Value.Check(SystemInfoResultSchema, { ...validSystemInfo, extra: true })).toBe(false);
   });
--- a/packages/gateway-protocol/src/schema/system-info.ts
+++ b/packages/gateway-protocol/src/schema/system-info.ts
@@ -45,6 +45,7 @@ export const SystemInfoResultSchema = closedObject({
       }),
     ),
   ),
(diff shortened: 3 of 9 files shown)


ChazGo:
@steipete, this exact head adds one required public production environment contract: `OPENCLAW_GATEWAY_ISOLATION`, projected strictly as `enabled` or `disabled` into the optional authenticated `system.info.gatewayIsolation` field. The shrink-only ratchet correctly reports 494 distinct names against the current 493 budget.

May I update `config/env-var-count-budget.txt` from 493 to 494 for this contract? I will not raise the protected budget without explicit owner approval. The PR remains draft while the matching public Windows packaging producer and its `clawctl gateway-isolation` flow are completed and proven.

ChazGo:
> May I update `config/env-var-count-budget.txt` from 493 to 494 for this contract? I will not raise the protected budget without explicit owner approval. The PR remains draft while the matching public Windows packaging producer and its `clawctl gateway-isolation` flow are completed and proven.

This comment was automatically added because of this comment:
`# Ratchet: lower this number when cleanup removes names; never raise it without owner approval.` in [config/env-var-count-budget.txt](https://github.com/openclaw/openclaw/blob/main/config/env-var-count-budget.txt)

MythiliMur:
I don't think we should do a core change for windows specific code. This setting should be part of windows launcher / bootstrapper.

@RomneyDa, FYI, in case you are asked about this PR.

ChazGo:
> I don't think we should do a core change for windows specific code. This setting should be part of windows launcher / bootstrapper.
> 
> @RomneyDa, FYI, in case you are asked about this PR.

When I was talking with @samanthasong, the guidelines more implied that gateway settings should go here:
>Avoid showing gateway settings in the windows companion app, even if it's a local gateway that was started through the companion app setup script / msix. It is less clean, has little value for now, and will have unpredictable consequences in node networks
>The default ways of editing gateway policies are the Control UI (web dashboard) or directly editing the file"

So, we could decide we won't have any UI reflect this gateway isolation status and it can only be obtained from the CLI.

MythiliMur:
https://github.com/openclaw/openclaw/pull/144452 -- adding this as external supervisor (which is what our windows launcher is）might be better than direct core changes. Can you please explore this option?

Corresponding PR in packaging repo - https://github.com/openclaw/openclaw-windows-packaging/pull/26 though I think you may already have something like this.

ChazGo:
> nal supervisor (which is what our windows launcher is）might be better than direct core changes. Can you please explore this option?

I see, I'm exploring adding this UI with a plugin in the packaging repo which will get automatically installed with the MSIX. I think the UI might be less visible, but it won't touch the openclaw core and we can expose more "clawctl" things in the plugin going forward if needed

ChazGo:
Abandoning in favor of putting the UI into a plugin thats Windows only for now [Add read-only Gateway isolation status plugin](https://github.com/openclaw/openclaw-windows-packaging/pull/28)

PR #144260, opened 2026-09-10
refactor: simplify stored goal fixtures in command tests

steipete:
## What Problem This Solves

Three Goal command tests repeat the same stored-goal setup, making their distinct resume and edit scenarios harder to scan.

## Why This Change Was Made

Use one pure, same-file fixture builder with explicit store path, session key, and status. Each call creates fresh objects and preserves the existing awaited write, all stored fields, and independently authored assertions. This removes 30 net test lines.

## User Impact

No user-visible behavior changes. Existing goal persistence, resume, and edit coverage remains intact; production code and storage semantics are unchanged.

## Evidence

- Complete `commands-goal.test.ts` before/after: all nine cases passed in identical order on Node 26.8.1/macOS.
- Structural expansion restores the complete original source token stream; fresh-object and mutation-isolation checks cover all three calls.
- The changed-file gate and independent review passed. The initial gate caught a shadowed helper parameter; the final naming-only correction passed repeated structural proof and the full gate.

No runtime savings are claimed.

### DIFF
--- a/src/auto-reply/reply/commands-goal.test.ts
+++ b/src/auto-reply/reply/commands-goal.test.ts
@@ -38,6 +38,33 @@ async function upsertSessionEntry(params: {
   );
 }
 
+function buildGoalSessionFixture(
+  storePath: string,
+  targetSessionKey: string,
+  status: "active" | "paused",
+): Parameters<typeof upsertSessionEntry>[0] {
+  return {
+    storePath,
+    sessionKey: targetSessionKey,
+    entry: {
+      sessionId: "sess-main",
+      updatedAt: 1,
+      goal: {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [80ec6931bff5a5d9734c313c47ed99179205027a](https://github.com/openclaw/openclaw/pull/144260/commits/80ec6931bff5a5d9734c313c47ed99179205027a)
- Landed commit: [99cd8f569d2591237eb33a04379f5b228e2330fe](https://github.com/openclaw/openclaw/commit/99cd8f569d2591237eb33a04379f5b228e2330fe)

PR #144273, opened 2026-09-10
refactor(tests): share UI bootstrap deferred fixtures

steipete:
## What Problem This Solves

The Control UI bootstrap tests maintained three local deferred-promise factories and a separate eager resolver capture even though the shared test helper already owns this setup. Maintaining those copies adds no behavioral coverage.

## Why This Change Was Made

Use the existing `createDeferred` helper for startup routing, pending bootstrap responses and connection coordination. All existing assertions, response fixtures, waits and ordering checks remain. Concrete response/location payload types and both `ReturnType<typeof createDeferred<void>>` annotations stay explicit; the ordinary void call uses the helper's default. The shared helper also returns an unused rejection capability; these tests consume only the promise and resolver.

## User Impact

No user-visible or production change. The three test files lose 23 lines overall (+23/-46), with no new test helper or runtime seam.

## Evidence

The original and final candidate each passed all 65 existing cases through `node scripts/run-vitest.mjs ui/src/app/bootstrap.test.ts ui/src/app/config.test.ts ui/src/app/connection-bootstrap.test.ts` on Node 26.5.0/macOS. Both normal native report sets retain 24 shared UI cases and 41 isolated UI cases. Startup/stop ordering, request supersession, connection-generation isolation and credential handling assertions remain covered.

These are local jsdom tests with existing mocks, not a live browser or Gateway exercise. An intermediate scoped lint run caught one unnecessary default type argument; the final candidate removes it while preserving the required type captures. All 18 required changed-file checks passed, including the affected test typecheck and lint. Independent Codex review is clean through P2.

### DIFF
--- a/ui/src/app/bootstrap.test.ts
+++ b/ui/src/app/bootstrap.test.ts
@@ -1,6 +1,7 @@
 import type { RouteLocation } from "@openclaw/uirouter";
 import { describe, expect, it, vi } from "vitest";
 import { CONTROL_UI_BASE_PATH_ATTRIBUTE } from "../../../src/gateway/control-ui-contract.js";
+import { createDeferred } from "../../../test/helpers/promise.js";
 import type { GatewayBrowserClient } from "../api/gateway.ts";
 import { routeIdFromPath, type RouteId } from "../app-routes.ts";
 import {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [58b02c93a9f9d2beb24545a89a8e3e2caf13aca2](https://github.com/openclaw/openclaw/pull/144273/commits/58b02c93a9f9d2beb24545a89a8e3e2caf13aca2)
- Landed commit: [46d4daf83b6d6342dddba2234bec390e1034fad3](https://github.com/openclaw/openclaw/commit/46d4daf83b6d6342dddba2234bec390e1034fad3)

PR #144275, opened 2026-09-10
fix(config): preserve literal provider-prefixed model IDs

steipete:
Related: #130706, #142100, #134496

## What Problem This Solves

Fixes an issue where users configuring a model ID that begins with its provider name would have that namespace removed when config loaded or changed. For example, `custom/custom/model` became `custom/model`, collapsing a fallback into the primary and allowing model settings to move onto the wrong row.

## Why This Change Was Made

The config parser already separates the provider from the literal model ID. Rejoin those parsed segments directly instead of interpreting the model prefix again. Existing Google and Together migrations remain unchanged.

## User Impact

Primary models, fallbacks, and per-model settings keep distinct provider-local namespaces across config reads and writes. This restores the documented first-slash parsing contract.

## Evidence

- Two regression cases failed before the repair; six migration controls passed. The existing tables cover distinct primary/fallback references, model-setting keys, and preserved parameters during writes.
- All 463 tests across five focused config suites passed, including the complete config-write suite. Its eight source-rename variants now use the maintained Gemini retirement and retain every root/include, transform/replacement, canonical-present, and write assertion. All eight also passed in a targeted run.
- The full changed gate and a fresh P2 review passed for the complete four-file PR candidate against its original base.
- A full `pnpm build`, including declarations and UI assets, and the compiled runtime proof passed at `b0502dcfb17ddc21321e6a3895367e84c7d0c497`. This follow-up changes two test constants only; all 40,745 other tracked files match that built source byte for byte. No runtime rebuild was needed for the fixture change.
- Compiled runtime proof read config, changed the fallback alias through the production writer, and reread both distinct model rows with their original parameters. An unpinned reply made the expected HTTP `model` → 404 and `custom/model` → 200 requests, delivered the response, and persisted the exact answering model in its transcript.
- The compiled process exited normally after closing its production services; all source files and 14,639 runtime output entries remained unchanged; 5,146 loaded modules were also checked. Earlier fixture cleanup failures are preserved in the local evidence packet.

Fallback notice formatting and `/model`/`/status` projection remain separate follow-up work in #134496.

Production delta: **−1 line**; test delta: **+17 lines**. Exact-head CI and native prepare must pass before landing.

### DIFF
--- a/src/config/io.write-config.test.ts
+++ b/src/config/io.write-config.test.ts
@@ -3026,8 +3026,8 @@ describe("config io write", () => {
       await withSuiteHome(async (home) => {
         const entry = { alias: "friendly", params: { temperature: 0.2 } };
         const canonicalEntry = canonicalPresent ? { params: { temperature: 0.7 } } : entry;
-        const legacy = "openrouter/openrouter/hunter-alpha";
-        const canonical = "openrouter/hunter-alpha";
+        const legacy = "google/gemini-3-pro-preview";
+        const canonical = "google/gemini-3.1-pro-preview";
         const agents = {
           entries: { main: {} },
           defaults: {
--- a/src/config/io.write-prepare.test.ts
+++ b/src/config/io.write-prepare.test.ts
@@ -1317,13 +1317,20 @@ describe("config io write prepare", () => {
     ).toEqual({ agents: { defaults, entries: { main: {}, ops: {} } }, gateway: { mode: "local" } });
   });
 
-  it("preserves authored Google model params under normalized config keys", () => {
+  it.each([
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c5570b47cfca1e3e781eb833a4b6d349e11f5c54](https://github.com/openclaw/openclaw/pull/144275/commits/c5570b47cfca1e3e781eb833a4b6d349e11f5c54)
- Landed commit: [96b5a339f3d9a165bf5c2931b4259a50bf9dafbb](https://github.com/openclaw/openclaw/commit/96b5a339f3d9a165bf5c2931b4259a50bf9dafbb)

PR #144278, opened 2026-09-10
perf(agents): reuse bounded CLI prefix capture

### LINKED ISSUE #144277: Avoid repeated copying of retained CLI output prefixes
The CLI runner repeatedly concatenates its entire retained stdout/stderr prefix while receiving fragmented output, duplicating the existing bounded head-capture implementation.

Reuse the canonical capture owner while preserving the byte cap, direct replacement decoding, error-prefix precedence and JSONL behavior. Both versions pass 80 tests; 12 comparisons preserve exact bytes/cap decisions. Measured concat output for a one-MiB prefix falls from about 512.5/128.5 MiB to one MiB for one/four-KiB chunks, with more retained Buffer objects; no total-memory or provider-latency claim. Selected gate and independent review pass.

steipete:
Closes #144277

## What Problem This Solves

Fragmented CLI output repeatedly copies the retained prefix as each chunk arrives. A one-MiB parse limit still permits hundreds of MiB of cumulative concatenation work before the run completes.

## Why This Change Was Made

Reuse the existing bounded head-capture owner for stdout and stderr, and concatenate lazily when the retained text is needed. Keep the current byte cap, oversized-output rejection, JSONL streaming, diagnostic hashes/tails and UTF-8 replacement decoding. The generic finalizer is intentionally not used because trimming clipped UTF-8 would change error prefixes.

## User Impact

CLI prefix capture avoids repeated whole-prefix copying and removes the duplicate local append helper. The capture retains one Buffer object per admitted chunk, so this is not a universal retained-heap or RSS improvement.

## Evidence

Both versions pass the same 80 executor/shared-retention tests. All 12 bounded benchmark comparisons preserve exact materialized bytes and cap decisions. For a one-MiB prefix, measured cumulative Buffer.concat output falls from about 512.5 MiB with one-KiB chunks and 128.5 MiB with four-KiB chunks to one MiB; retained capture objects rise from one to 1,025/257. These counters exclude other allocation paths and do not measure provider latency or total memory.

The full selected changed gate and independent P2 review pass. Suite timing/RSS includes imports and preparation and is not used as a speed or memory claim. No Gateway or provider activity was involved.

The paired 80-test executions recorded 27.69 seconds and 2,067,251,200 bytes maximum RSS before, versus 23.13 seconds and 2,058,207,232 bytes afterward. These whole-invocation observations include imports, transform caches and worker preparation; they do not isolate execution speed or retained-memory savings. The copy-volume result above is the measured optimization claim.

### DIFF
--- a/src/agents/cli-runner/execute-process.ts
+++ b/src/agents/cli-runner/execute-process.ts
@@ -8,6 +8,7 @@ import {
 import { createModelCallStreamProgressReporter } from "../../logging/diagnostic-model-stream-progress.js";
 import { beginDiagnosticBackendActivity } from "../../logging/diagnostic-run-activity.js";
 import type { CliBackendConfig } from "../../plugins/cli-backend.types.js";
+import { appendCapturedOutput, createCapturedOutputBuffers } from "../../process/exec-output.js";
 import type { RunExit } from "../../process/supervisor/types.js";
 import type { CliOutput, CliTerminalInterruption } from "../cli-output-contracts.js";
 import { createCliJsonlStreamingParser } from "../cli-output-stream.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5dbed969bdae0ecf6f3439e77d3e087178dbb775](https://github.com/openclaw/openclaw/pull/144278/commits/5dbed969bdae0ecf6f3439e77d3e087178dbb775)
- Landed commit: [a18d4e4962f5a6f5133301c6b4a7c96cb478e4a6](https://github.com/openclaw/openclaw/commit/a18d4e4962f5a6f5133301c6b4a7c96cb478e4a6)

PR #137157, opened 2026-09-03
fix(docs): match source anchor checks to published sections

steipete:
> **Resolved and published (2026-09-03).** All three rollout PRs are merged, normal source sync and R2 publication succeeded, and post-publication browser/live smoke passed. [Final rollout and remaining content findings](https://github.com/openclaw/openclaw/issues/137032#issuecomment-5525222650).

Related: https://github.com/openclaw/openclaw/issues/137032
Support: https://github.com/openclaw/openclaw/pull/137038
Renderer: https://github.com/openclaw/docs/pull/157
Production proof: https://github.com/openclaw/docs/pull/157#issuecomment-5522338636

## What Problem This Solves

Source anchor checks could accept nonexistent published sections or reject links that work on the live docs site. The independent Mintlify checker does not implement the publisher's heading, component, duplicate-ID, and HTML-alias redirect behavior. The replacement also needs reliable diagnostics: repeated links must identify their own source occurrences, and malformed URLs must not abort reporting for the rest of the document.

## Why This Change Was Made

Use the same source-owned document/ID and redirect contracts consumed by the deployed renderer. This follows the approved support → renderer → audit rollout: support was normally synced, the renderer merged, and real production navigation passed before audit activation was published.

Delete the Mint subprocess, Node-version workaround, temporary anchor projection and obsolete tests rather than adding another guessed slugger. Keep external-link projection and canonical ClawHub mirroring intact. Literal IDs take precedence over one fragment decode; only real emitted aliases/components count as targets. Relative links use final published page URLs, and redirects follow the publisher's actual HTML-alias behavior.

Source-line provenance travels through shared preprocessing and existing MarkdownIt/htmlparser2 positions, before href normalization loses occurrence identity. An opt-in `mapLink` projection preserves the publisher's default string-link API and avoids diagnostic tracking during ordinary publishing. URL construction/decoding has one `Result` boundary: malformed internal paths become normal diagnostics, later links are still checked, and external redirect fragments are not checked against local pages.

## User Impact

Maintainers can trust `pnpm docs:check-links:anchors` to check actual published targets while preserving old heading URLs. Repeated/reference/component links report their own locations; unmapped snippet/embedded expansions report `unknown`, not a guessed line. Author/testing guides describe the shared parser accurately. No product configuration, schema, protocol, dependency pin, runtime gateway behavior, or page-specific anchor rewrite changes are included.

## Evidence

[...]

PR #144279, opened 2026-09-10
fix(docs): reject unpublished permalink routes

steipete:
Related: #137157

## What Problem This Solves

Fixes an issue where docs link checks accepted frontmatter permalink URLs that the publisher does not create. A page with `permalink: /unpublished` could make plain links and navigation to `/unpublished` pass even though no page or configured redirect served that route.

## Why This Change Was Made

Remove the legacy permalink inventory and use the existing physical-page, index-route and configured-redirect owners consistently. This also removes the preliminary Markdown read pass. The declared ClawHub projection and shared publishing parser retain their existing contracts.

## User Impact

Docs checks now report these missing destinations in both plain and anchor modes. Published page routes and configured redirects continue to resolve.

## Evidence

The same two actual-CLI regression cases fail before the change and pass afterward. The fixture contains two unpublished links and an unpublished navigation entry: plain mode previously reported none, while anchor mode missed the navigation entry. Both modes now report all three and retain a valid physical-page control.

All 35 owner tests and 25 canonical changed checks pass. Existing coverage retains physical/index routes, redirects, assets and ClawHub behavior. Independent review found no actionable findings.

A transparent probe of the actual CLI fixture observed Markdown `readFileSync` calls fall from two to one per file. This measures calls, not disk I/O or elapsed-time speedup. No publisher deployment or separate publisher build was performed. The change removes 17 tooling lines and adds 21 test lines.

Compatibility: this patch removes entries from a process-local route index. It changes no stored-state schema or serialization and does not rewrite Markdown, frontmatter or docs.json. The existing physical/index and configured-redirect publishing contract is unchanged. The only authored permalink in this checkout is `/security/formal-verification/`, which already matches its physical page. A frontmatter-only alias has no published destination; an intended alias should use a configured redirect to an existing page. No data migration is required.

### DIFF
--- a/scripts/docs-link-audit.mts
+++ b/scripts/docs-link-audit.mts
@@ -329,25 +329,8 @@ function buildAuditIndex(
 
   for (const abs of markdownFiles) {
     const rel = normalizeSlashes(path.relative(docsDir, abs));
-    const text = fs.readFileSync(abs, "utf8");
     const slug = rel.replace(/\.(md|mdx)$/i, "");
     addRoute(routes, slug);
-
-    if (!text.startsWith("---")) {
-      continue;
-    }
-
-    const end = text.indexOf("\n---", 3);
-    if (end === -1) {
-      continue;
-    }
-    const frontMatter = text.slice(3, end);
-    const match = frontMatter.match(/^permalink:\s*(.+)\s*$/m);
-    if (!match) {
-      continue;
-    }
-    const permalink = (match[1] ?? "").trim().replace(/^['"]|['"]$/g, "");
-    routes.add(normalizeRoute(permalink));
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6926a9dfb5b09369ae1b66936fed1e5445021144](https://github.com/openclaw/openclaw/pull/144279/commits/6926a9dfb5b09369ae1b66936fed1e5445021144)
- Landed commit: [f7c4102de5091b6890ff20a21a4f60a6208eac83](https://github.com/openclaw/openclaw/commit/f7c4102de5091b6890ff20a21a4f60a6208eac83)

PR #144284, opened 2026-09-10
fix(ui): keep retained conversation panes in stable slots

### LINKED ISSUE #144283: Retained conversations reload when another pane is evicted
Opening a fourth recent conversation can evict an older pane while moving a surviving hidden pane. That DOM move triggers disconnect cleanup and repeats the survivor's startup and full-message loading, even though it remains within the retained set.

Keep the surviving panes in stable positions owned by the existing three-pane retainer. The actual original production-bundle flow performs two startup/full-message reads; the corrected flow performs one each while preserving recovered text, eviction bounds and existing owner/sibling behavior. Full selected gate and independent review pass. Captures use synthetic conversations; no memory or latency claim is made.

steipete:
Closes #144283

## What Problem This Solves

Fixes an issue where opening another recent conversation could reload a conversation that should remain retained. Evicting an older pane moved the surviving pane in the DOM, triggering disconnect cleanup and duplicate startup/full-message requests.

## Why This Change Was Made

The existing retainer assigns stable positions within its three-pane bound. Rendering keeps each surviving pane in place and replaces only the evicted conversation. Empty positions do not become pane or MCP owner keys; existing preview, alias, draft and close behavior stays with the retainer.

## User Impact

A retained conversation keeps its loaded reply and pane state when another pane is evicted, instead of unnecessarily loading the same data again. The change adds 21 production lines to make DOM position part of the existing lifecycle owner; no memory or latency improvement is claimed.

## Evidence

An independently built original bundle reproduces two startup and two full-message reads for the surviving pane; the final rebuilt candidate performs one of each and preserves the recovered text and three-pane limit. The 23 retainer tests, two unchanged reconnect/Escape sibling E2Es, full selected changed gate and fresh independent P2 review pass.

The initial gate found a no-new-array lint violation; the equivalent typed array construction was corrected and the affected owner/candidate proof plus full gate rerun passed. Both full captures are synthetic and inspected. The before picture shows the reply after its unnecessary reload completed; request counters, rather than the screenshots alone, prove the lifecycle repair.

The first hosted UI run exposed five existing checks that read the first retained DOM pane. Reproduction on the exact CI tree confirmed that the active parent/archive title was correct while the earlier pane remained hidden. Those two test files now target the active pane, and their shared search shortcut uses Playwright’s platform-aware ControlOrMeta. Every expectation and timeout is preserved; the original production patch is unchanged. Current header coverage passes9 cases and the unchanged archive file passes11 from separate invocations, followed by a full seven-path changed gate and fresh P2 review. The earlier failed run remains recorded; the revised head requires fresh hosted CI.


steipete:
Merged via squash.

- Prepared head SHA: [75c1f5ac2ea1b1978d88a30c42d0a19c3de7119a](https://github.com/openclaw/openclaw/pull/144284/commits/75c1f5ac2ea1b1978d88a30c42d0a19c3de7119a)
- Landed commit: [94fd88c0cb05e709bc33bdd09f8f596669cfbee8](https://github.com/openclaw/openclaw/commit/94fd88c0cb05e709bc33bdd09f8f596669cfbee8)

PR #144290, opened 2026-09-10
perf(plugins): avoid temporary collections in provider policy lookup

### LINKED ISSUE #144288: Provider policy lookup allocates temporary ownership sets and registry sorts
Provider policy owner selection allocates a normalized Set for each plugin and sorts all manifest records on each lookup. The ownership declarations and stable lexical winner can be checked directly without constructing those intermediate collections.

The scoped repair must preserve provider normalization, simple versus scoped alias handling, trusted external filtering, equal-ID stable order and the original registry. No new cache or public/configuration/storage contract is needed. A source-bound synthetic lookup comparison and existing owner tests cover this allocation reduction; no whole-Gateway memory claim is implied.

steipete:
Closes #144288

## What Problem This Solves

Provider policy lookup repeatedly constructs a normalized ownership Set for each plugin and sorts the whole registry before selecting matching owners. This duplicates allocation on a path used to resolve lightweight provider artifacts.

## Why This Change Was Made

Check declared provider, CLI backend and embedding IDs directly. Resolve simple aliases only against those same declarations, and return early when aliases are absent. Bundled lookup tracks the first matching owner in the existing stable lexical order; external lookup sorts only the filtered matches. The registry remains unchanged.

## User Impact

Provider selection, normalization, alias scope, trust filtering and equal-ID ordering are preserved. No public API, configuration, schema, persistence or migration behavior changes. The change adds 10 production lines to replace repeated temporary collections with direct selection.

## Evidence

Both source versions pass the same 43 owner tests. Twelve fresh interleaved processes verify identical ordered outputs for two synthetic workloads with 256 records, seven queries and 8,000 measured calls per run. Median sampled allocation falls from 1,944,606,240 to 980,339,352 bytes for external lookup (49.6%) and from 1,786,972,640 to 978,775,496 bytes for bundled lookup (45.2%). The measurement includes all sampled nodes in the isolated interval, including moved helper work and collected objects; it does not establish retained heap, whole-Gateway RSS or latency savings.

The full selected changed gate, normal build and fresh independent P2 review pass. An initial lint failure required changing filtered sort to toSorted; the final source was reviewed and remeasured, and all required checks passed afterward.

### DIFF
--- a/src/plugins/provider-public-artifacts.test.ts
+++ b/src/plugins/provider-public-artifacts.test.ts
@@ -8,6 +8,7 @@ import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
 import type { ModelProviderConfig } from "../config/types.models.js";
 import { createPluginCache, withPluginCache } from "./plugin-cache.js";
 import { clearPluginMetadataLifecycleCaches } from "./plugin-metadata-lifecycle.js";
+import { createPluginManifestRecordFixture } from "./plugin-metadata.test-support.js";
 import { resolveDirectBundledProviderPolicySurface } from "./provider-policy-surface.js";
 import {
   listTrustedExternalProviderPolicyOwners,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [bd02e374af4128ff6242d9ed2dff5817913f4eed](https://github.com/openclaw/openclaw/pull/144290/commits/bd02e374af4128ff6242d9ed2dff5817913f4eed)
- Landed commit: [a0a3bd0b6302b7f25f847cb9e9b356bec489fa36](https://github.com/openclaw/openclaw/commit/a0a3bd0b6302b7f25f847cb9e9b356bec489fa36)

PR #132858, opened 2026-08-29
fix(testing): reject invalid profiler CLI options

steipete:
Closes https://github.com/openclaw/openclaw/issues/132855 — profiler CLI validation runs too late.

Related: https://github.com/openclaw/openclaw/pull/132224 — awaited profiling completion ownership. The separate native-shutdown fixture correction in https://github.com/openclaw/openclaw/pull/132820 is already merged and is not duplicated here.

<details>
<summary>Additional instructions</summary>

This same-repository branch is takeover-ready for maintainers. The exact-version Vitest parser patch was explicitly approved for this repair before implementation.

</details>

## What Problem This Solves

Fixes an issue where developers profiling tests would silently accept an unknown Vitest option, or evaluate their config before receiving a missing-value error, after the profiler moved to the public programmatic Vitest API. Both main and runner modes were affected. Native Vitest rejects the same inputs before loading config.

## Why This Change Was Made

Keep the profiler's awaited capture lifecycle and restore validation at its parser owner. The approved `vitest@4.1.11` patch retains the CAC instance after nonexecuting parsing and calls the same unknown-option, required-value, and required-argument checks used by native command dispatch. Help/version bypasses and repeated-option transforms remain native; there is no copied option catalog, private import, command execution, or second config pass.

Strict parsing also removes reliance on unsupported CLI-only spellings: custom runners remain supported through Vitest configuration, and the unreachable direct-CLI `globalSetup` append path is deleted. The normal Vite config merge still preserves configured setup, reporters, pools, isolation, environments, and runners.

## User Impact

Invalid profiler arguments now fail before config evaluation. Valid profiling commands still await CPU/heap capture completion, preserve native help and option errors, and retain their selected execution settings. This changes developer tooling only; no Gateway, channel, provider, configuration schema, or persistence behavior changes.

### Before and after at the real CLI boundary

An isolated config wrote a marker and threw if evaluated. Before the fix, both profiler modes created that marker for an unknown option and a missing `--pool` value; native Vitest rejected both without evaluating config. Four initial boundary regression cases failed for that exact reason. After the repair, the profiler matches the native reference. Coverage also rejects config-only `--runner` flags, retains explicit `allowUnknownOptions` behavior in the public parser, and keeps required-value/argument checks active.

[...]

PR #144292, opened 2026-09-10
fix(agents): avoid sibling output limits for unlisted models

### LINKED ISSUE #144160: [Bug]: configured-fallback model inherits maxTokens from an unrelated models[0] row and labels it user-configured
## Summary

When a provider is configured with a `baseUrl` and a request names a model id that matches no `models[]` row, no registry row and no static-catalog row, the synthesized fallback model inherits its `maxTokens` output cap from `providerConfig.models[0]`, an unrelated model, and stamps it `maxTokensSource: "configured"` as if the user had authored a cap for the requested model. That cap is then emitted verbatim as the wire-level `max_tokens` / `max_completion_tokens`.

## Environment

- Surface: custom and proxy OpenAI-compatible providers (LiteLLM, vLLM, LM Studio, ds4, any OpenAI-compatible gateway), Anthropic equivalent noted below

## Steps to reproduce

1. Configure a provider exactly as the shipped ds4 setup doc does (`docs/providers/ds4.md:100-125`): a `baseUrl` plus a single `models[]` row for `deepseek-v4-flash` carrying `contextWindow: 32768` and `maxTokens: 128`.

```jsonc
{
  "models": {
    "mode": "merge",
    "providers": {
      "ds4": {
        "baseUrl": "http://127.0.0.1:18000/v1",
        "api": "openai-completions",
        "models": [
          { "id": "deepseek-v4-flash", "contextWindow": 32768, "maxTokens": 128 }
        ]
      }
    }
  }
}
```

2. Run a second model served by that same endpoint without adding a row for it, for example `ds4/deepseek-v4-chat`. Resolution falls through `resolveTieredModel` (`src/agents/embedded-agent-runner/model-resolution.ts:45-73`) to `buildConfiguredFallbackModel`, reached via `model.registry-resolution.ts:456` and via `model.ts:356-367`. `hasConfiguredFallbackSurface` (`model.configured-overrides.ts:238-249`) admits the model from the bare `baseUrl` alone.

3. Inspect the resolved model's `maxTokens` and `maxTokensSource`.

## Expected

No cap is known for `deepseek-v4-chat`, so the field should be omitted, per the documented default at `docs/concepts/model-providers/custom-providers.md:378`: "an unknown output limit omits both `max_tokens` and `max_completion_tokens`, letting the provider apply its default". `maxTokensSource` should not be `"configured"`, since the user authored no cap for this model.

## Actual

`maxTokens: 128` is inherited from the unrelated `deepseek-v4-flash` row and stamped `maxTokensSource: "configured"`. Every reply from `deepseek-v4-chat` is silently truncated at 128 output tokens, with no warning and no diagnostic naming the row the number came from. In the opposite direction, an oversized first-row cap is sent to a model that rejects it, producing an HTTP 400 on strict proxies.

This also contradicts `docs/concepts/model-providers/quick-rules.md:17`, where a per-model `maxTokens` overrides output capacity "for that model".

## Root cause

`src/agents/embedded-agent-runner/model.configured-fallback.ts:159-162`, consumed at `:204-209`:

```ts
const configuredFallbackMaxTokens =
  configuredModel?.maxTokens ??
  providerConfig?.maxTokens ??
  providerConfig?.models?.[0]?.maxTokens;   // a DIFFERENT model's output cap
...
maxTokensSource: configuredFallbackMaxTokens !== undefined ? "configured" : "discovered",
```

In this branch `configuredModel` is `undefined` by definition, since the whole point of the branch is that the requested id matched no configured row. So the chain falls through the documented provider-level default (`providerConfig.maxTokens`) into the per-model cap of whichever row happens to be listed first.

The `"configured"` stamp then makes the value authoritative downstream: `src/agents/models-config.merge.ts:173-176` and `model.configured-overrides.ts:552-553` both treat a `"configured"` source as a user-authored cap that must not be overridden by discovery.

Commit archaeology, which is the core of the argument:

- #114710 (`3d516a609ae6fa1b2b990960bb236c5993eac1ef`, "recognize loopback endpoints and clamp fallback output") introduced `clampModelMaxTokensToContextWindow` and left both positional `models[0]` fallbacks in place, for `maxTokens` and for `contextWindow`.
- #124665 (`715c379fd949a9a77224a590e8d9016f88b62d0c`, "consolidate context budget to one per-model knob") removed `providerConfig?.contextWindow ?? providerConfig?.models?.[0]?.contextWindow` from the `contextWindow` chain and the matching pair from `contextTokens`, but left `models?.[0]?.maxTokens` behind.

That removal made the symptom worse rather than better. Before it, an inherited cap was at least clamped against the same row's own `contextWindow`. Now the cap is inherited from the row while the context window is not, so the clamp runs against `DEFAULT_CONTEXT_TOKENS` (200000) and cannot soften a 128 cap. The captured output below shows exactly this: `maxTokens=128` alongside `contextWindow=200000`.

`clampModelMaxTokensToContextWindow` only clamps downward (`model.configured-overrides.ts:166`), so it can never repair an inherited value. `packages/ai/src/transports/openai-completions-params.ts:92` returns the cap with `clampToModelMaxTokens: false`, and `:523-527` writes it straight into `max_tokens` or `max_completion_tokens`.

A repo-wide grep confirms `model.configured-fallback.ts:162` is now the only positional `models[0]` value fallback left in non-suite code under `src/agents/` and `src/config/`. The neighbouring `configured-provider-fallback.ts:48` use is a deliberate "first listed model is the provider default model" selection, which is a different and intentional thing.

The file's own comment two lines below already states the correct rule, which the `models[0]` leg violates:

```ts
// maxTokens is a wire-level output cap, not a context-budget fallback.
// Omit an unknown cap so strict providers can apply their own limit.
```

Suggested fix, offered as a suggestion since this is an issue and not a pull request: drop the `?? providerConfig?.models?.[0]?.maxTokens` leg, matching what #124665 already did for `contextWindow` and `contextTokens`. The remaining chain is exactly the documented precedence, and omitting an unknown cap is the documented default.

## Sibling surfaces

- The `contextWindow` sibling on the very same expression was already fixed by #124665; only the `maxTokens` leg survived.
- Anthropic emits the same field from the same resolved model at `packages/ai/src/providers/anthropic.ts:647` (`max_tokens: options?.maxTokens ?? model.maxTokens`), so a configured Anthropic-compatible proxy carries the inherited cap onto the wire too.
- The path is already exercised by the shipped case "includes provider baseUrl in fallback model" (`model.test.ts:1920-1929`), which never adds a `models[]` row, so this leg has no coverage today.

## Real behavior proof

Behavior addressed: an unlisted model id on a `baseUrl` provider inherits the first listed row's output cap and is labelled user-configured.
Real environment tested: real runtime driven, at commit c973fe7e721905ba2318994f1f47ae97677cf9ac
Exact steps or command run after this patch: called the real exported `buildConfiguredFallbackModel` from `src/agents/embedded-agent-runner/model.configured-fallback.ts` with the ds4-shaped config above, once for the listed id `deepseek-v4-flash` and once for the unlisted id `deepseek-v4-chat`, via `node --import ./scripts/tsx.mjs`. A control run repeated the unlisted id with the `models` array emptied so no foreign row exists.
Evidence after fix:
```
modelId=deepseek-v4-flash listed=true -> maxTokens=128 maxTokensSource=configured contextWindow=32768 baseUrl=http://127.0.0.1:18000/v1
modelId=deepseek-v4-chat listed=false -> maxTokens=128 maxTokensSource=configured contextWindow=200000 baseUrl=http://127.0.0.1:18000/v1

modelId=deepseek-v4-chat listed=false -> maxTokens=undefined maxTokensSource=undefined contextWindow=200000 baseUrl=http://127.0.0.1:18000/v1

# CONTROL (same unlisted id, foreign row removed, showing the value's true origin)
modelId=deepseek-v4-chat listed=false -> maxTokens=undefined maxTokensSource=undefined contextWindow=200000 baseUrl=http://127.0.0.1:18000/v1
```
Observed result after fix: the unlisted id `deepseek-v4-chat` resolves with a 128 output cap that belongs to `deepseek-v4-flash` and is labelled `configured`, while its own `contextWindow` resolves to the 200000 default rather than the row's 32768, so the downward clamp cannot soften the inherited number; removing the foreign row drops the cap to `undefined`, which pins the row as the sole origin of the value.
What was not tested: no live HTTP request was issued to a provider endpoint, so the wire-level emission of `max_tokens` / `max_completion_tokens` rests on reading `packages/ai/src/transports/openai-completions-params.ts:92,523-527` and `packages/ai/src/providers/anthropic.ts:647` rather than on a captured request body. The resolver was called directly rather than end-to-end through `resolveTieredModel`; the two call edges into it are cited by file and line above but were not driven live. The Anthropic path was not run at all.

ylcn91:
Closes #144160

## What Problem This Solves

An unlisted model on a configured custom provider could inherit the first listed model's output limit. A sibling row capped at 128 therefore caused requests for another model to send that same cap without an authored limit for the selected model.

## Why This Change Was Made

Remove the positional sibling-row fallback from the existing model resolver. Matching model limits, explicit provider defaults, and matching catalog metadata keep their existing precedence. An unknown limit stays unset; context-window fallback and transport guards are unchanged.

## User Impact

Unlisted models no longer receive another model's output cap. Explicit limits, known model ceilings, and the remaining context budget still constrain requests normally.

## Evidence

- Packaged baseline `bc63d63bcb7c` and candidate `80d367abe52e` ran the public `openclaw agent --local --agent main --model ... --json` command against a real loopback OpenAI-compatible HTTP endpoint, with synthetic credentials and the OpenClaw runtime.
- Baseline sent `max_completion_tokens: 128` for an unlisted model whose only sibling row had that cap. Candidate sent neither `max_tokens` nor `max_completion_tokens`. Both selected the requested model and completed with the fixture's `OK` response.
- Nine matched public controls preserved listed limits, provider defaults, row precedence, compatibility field choice and headers, a genuine Mistral catalog limit, native-context clamping, public parameter precedence, and known model ceilings. Both comparison packages included the same genuine Mistral plugin; no catalog sizing was inserted into the fixture config.
- The new regression fails on the original owner with `expected 128 to be undefined`. With the fix, 167 model-resolution tests and 13 managed-transport tests pass. Formatting, targeted lint, line-count and assertion-safety checks pass; the normal commit hook ran on the isolated runner.
- Independent acceptance checked 11 fresh public requests, including unknown limits, provider defaults, real catalog metadata, headers, alias selection, and bounded public overrides.
- Additional same-input baseline controls confirm that larger public output overrides remain bounded by a listed model's known cap, and that shaped prompt/tool input can reduce the final output allowance below the native ceiling.

The HTTP endpoint supplies deterministic protocol responses. This proves actual request construction and response handling, not natural-model truncation behavior. Internal provenance labels are separate from HTTP fields; the configured catalog control preserves the catalog-derived numeric limit without claiming that the wire exposes its source label.

### DIFF
--- a/src/agents/embedded-agent-runner/model.configured-fallback.ts
+++ b/src/agents/embedded-agent-runner/model.configured-fallback.ts
@@ -156,10 +156,7 @@ export function buildConfiguredFallbackModel(params: {
     compat: fallbackCompat,
     reasoning: metadataModel?.reasoning,
   });
-  const configuredFallbackMaxTokens =
-    configuredModel?.maxTokens ??
-    providerConfig?.maxTokens ??
-    providerConfig?.models?.[0]?.maxTokens;
+  const configuredFallbackMaxTokens = configuredModel?.maxTokens ?? providerConfig?.maxTokens;
   const resolvedFallbackMaxTokens = configuredFallbackMaxTokens ?? staticCatalogModel?.maxTokens;
   const resolvedFallbackContextWindow =
     configuredModel?.contextWindow ?? staticCatalogModel?.contextWindow ?? DEFAULT_CONTEXT_TOKENS;
--- a/src/agents/embedded-agent-runner/model.test.ts
+++ b/src/agents/embedded-agent-runner/model.test.ts
@@ -1929,6 +1929,20 @@ describe("resolveModel", () => {
     expect(model.api).toBe("openai-completions");
   });
 
+  it("does not inherit an unrelated configured row's maxTokens for an unlisted fallback model", async () => {
+    const cfg = makeProviderConfig("custom", {
+      baseUrl: "http://localhost:9000",
+      models: [{ id: "listed-model", name: "listed-model", contextWindow: 32_768, maxTokens: 128 }],
+    });
+
+    const result = await resolveModelForTest("custom", "missing-model", state.agentDir(), cfg);
+    const model = expectResolvedModel(result);
+
+    expect(model.id).toBe("missing-model");
+    expect(model.maxTokens).toBeUndefined();
+    expect(model).not.toHaveProperty("maxTokensSource");
+  });
+
   it("defaults baseUrl-only Google fallback models to native Gemini transport", async () => {
     const cfg = makeProviderConfig("google", {
       baseUrl: "https://generativelanguage.googleapis.com",


obviyus:
landpr-approve-unrelated-red
pr: 144292
head: 80d367abe52ef710a108a81c04e1f37ace877774
run: 34539311511 attempt: 1
failed: checks-node-compact-small-55 job=103078413705 conclusion=failure
check: openclaw/ci-gate

obviyus:
Thanks @ylcn91!

**Proof:** The packaged public `openclaw agent --local --agent main --model ... --json` command reached a real loopback HTTP endpoint on both baseline and candidate. The unlisted model completed successfully after the unrelated output cap was removed. All 180 focused tests and 11 independent public cases passed. Final exact-head review found no actionable issues.

- **Bug:** An unlisted custom-provider model inherited another configured model's output limit.
- **Fix:** Resolve limits from the selected model, explicit provider defaults, or matching catalog metadata; leave unknown limits unset.

| Before | After |
| --- | --- |
| An unrelated row caused `max_completion_tokens: 128` on the unlisted model's request. | Both output-limit fields are absent unless an applicable owner supplies a limit. |
| Explicit limits and known model ceilings constrain requests. | Those constraints remain intact. |

The endpoint returned deterministic protocol responses. This proves request construction and response handling, not natural-model truncation.

**CI exception:** Used the guarded standing-approval path for the sole unrelated failed child, [`checks-node-compact-small-55`](https://github.com/openclaw/openclaw/actions/runs/34539311511/job/103078413705), in completed run [34539311511, attempt 1](https://github.com/openclaw/openclaw/actions/runs/34539311511). Its `rejects main --runner=custom-runner.ts before evaluating config` case timed out after 120 seconds; [the aggregate gate](https://github.com/openclaw/openclaw/actions/runs/34539311511/job/103081669316) failed because of that child.

The tested merge was `67803d52203249bd221a4ca16bbd8f562b0bea24`, with main parent `0d6322c2fab39aaa8ae4c805d7b8b6047a123f9c` and candidate `80d367abe52ef710a108a81c04e1f37ace877774`. Its only changes from that main parent are the two agent files in this PR. The failing case launches the profiler in a fresh temporary directory with a marker-only config. Its Node inspector, Vitest parser, profiling, and process cleanup path never enters the model resolver; the changed model test is not selected by the tooling shard. The test, helpers, configuration, manifest, and lockfile are identical to that main parent and the pinned baseline.

The [test case at lines 401–440](https://github.com/openclaw/openclaw/blob/67803d52203249bd221a4ca16bbd8f562b0bea24/test/scripts/run-vitest-profile.test.ts#L401-L440) was introduced by [145b07172959](https://github.com/openclaw/openclaw/commit/145b071729596a7012c50156fc0b1a84dbad2856) in [#132858](https://github.com/openclaw/openclaw/pull/132858), verified against its parent. This identifies test provenance, not the cause of the timeout. The exact stalled operation remains unknown; no matching main failure or counterfactual runtime reproduction is claimed. No retry or timeout increase was used. The retained profiler failure is a separate follow-up.

[Exact-run approval](https://github.com/openclaw/openclaw/pull/144292#issuecomment-5626832929).

PR #144293, opened 2026-09-10
fix(devices): report available memory instead of raw free pages on macOS

### LINKED ISSUE #144176: Device inventory memory bar always red on macOS due to os.freemem() misunderstanding
## Summary

The memory bar in **Control UI → Devices → Paired devices** shows ~100% used (red) for any macOS host, regardless of actual memory pressure.

## Why

`os.freemem()` only counts `Pages free` on macOS, ignoring `Pages inactive` and `Pages purgeable` — both are reclaimable file cache. So `total - free` always lands near 100% even when the system is idle.

The widget computes:

```js
let t = e.memoryTotalBytes - e.memoryFreeBytes;
r.push(ar(`memory`, t/e.memoryTotalBytes*100, ...))
```

with default thresholds `80% warn / 90% danger`. On macOS this is always red.

## Suggested fix

Use macOS-aware "App memory" instead (matches Activity Monitor):

```bash
vm_stat | awk '
  /Pages active/   { a=$3 }
  /Pages inactive/ { i=$3 }
  /Pages wired/    { w=$3 }
  /Pages purgeable/{ p=$3 }
  /Pages free/     { f=$3 }
  END {
    app = (a + w - p) * 100 / (a + i + w + p + f)
    print app
  }'
```

Show **App memory / Physical memory**, with warn at ~75% and danger at ~90%.

Same fix applies to `debug.overlay.memoryMb`, which uses Node `process.memoryUsage().heapUsedBytes` and is equally misleading on macOS.

## Environment

- OpenClaw 2026.9.3
- macOS arm64

ylcn91:
Closes #144176

## What Problem This Solves

Fixes an issue where the memory meter in Control UI → Devices, on both paired node rows and the Gateway row, showed close to 100% used and rendered red on any macOS host, even when the machine was idle.

## Why This Change Was Made

Both host-stat producers read `os.freemem()`, which on Darwin counts only free pages and ignores inactive and purgeable file cache, so `total - free` always lands near the total. Node's `process.availableMemory()`, available on the supported Node 24 and 26 baselines, counts reclaimable pages as well. One `readHostFreeMemoryBytes()` helper in `src/infra/host-memory.ts` now serves the node-host sampler and the Gateway `system.info` handler, and falls back to `os.freemem()` where the runtime reports no available-memory figure. The `memoryFreeBytes` wire field and the meter's 80/90% thresholds are unchanged; the field's meaning is stated in the presence protocol doc.

Not changed: the issue's `debug.overlay.memoryMb` claim concerns process heap, a different diagnostic that matches its current implementation.

## User Impact

macOS operators see a memory meter that reflects memory actually under pressure instead of a permanent danger-level shortage. Linux and Windows readings are unchanged, because libuv already reports available memory there (verified on this Linux host: `os.freemem()` and `process.availableMemory()` both return the `MemAvailable` figure).

## Evidence

- `src/node-host/host-stats.test.ts` drives the sampler through `process.availableMemory` (fails on main, where the sampler still reads `os.freemem`) and covers the free-memory fallback; `system-info.test.ts` asserts the Gateway row uses the same reader.
- `node scripts/run-vitest.mjs src/node-host/host-stats.test.ts src/gateway/server-methods/system-info.test.ts packages/gateway-protocol/src/schema/nodes.test.ts`: all passed.
- Not captured: a live macOS Control UI screenshot; this devbox is Linux. The Darwin difference is libuv's `uv_get_free_memory` (free_count only) versus `uv_get_available_memory` (free + inactive + purgeable) in `deps/uv/src/unix/darwin.c`.
- `pnpm check:changed`: pending, will update.

### DIFF
--- a/docs/gateway/protocol/presence.md
+++ b/docs/gateway/protocol/presence.md
@@ -55,7 +55,9 @@ encoding in `payloadJSON`):
 1-, 5-, and 15-minute averages, each finite and between 0 and 100000. Windows
 has no load average; hosts omit the field when all three readings are zero.
 Memory and disk values are non-negative integer bytes, with free or available
-bytes no greater than their total. Disk fields appear together only when the
+bytes no greater than their total. `memoryFreeBytes` counts free plus reclaimable
+memory (Node's available-memory reading), so idle file cache on macOS is not
+reported as used. Disk fields appear together only when the
 host can read capacity for the volume containing its home directory, independent
 of the worker's current directory.
 
--- a/src/gateway/server-methods/system-info.test.ts
+++ b/src/gateway/server-methods/system-info.test.ts
@@ -42,6 +42,7 @@ describe("system.info", () => {
     sampleTime += 10_001;
     vi.spyOn(Date, "now").mockReturnValue(sampleTime);
     vi.spyOn(os, "platform").mockReturnValue("darwin");
+    vi.spyOn(process, "availableMemory").mockReturnValue(4096);
     mocks.runCommandWithTimeout.mockReset().mockImplementation(mountedVolumeOutput);
   });
   afterEach(() => vi.restoreAllMocks());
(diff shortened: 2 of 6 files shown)


ylcn91:
Closing this one: the user-visible proof for this change is a before/after Devices page capture on macOS, and I cannot produce that on my Linux host. The unit-level evidence in the description is not enough for a Control UI change. The branch stays available at ylcn91:fix/issue-144176 if someone with a Mac wants to pick it up with the missing capture.

PR #144296, opened 2026-09-10
refactor(msteams): decode inline images during materialization

### LINKED ISSUE #144295: Teams attachment discovery decodes inline images before materialization
Teams advertised-media discovery currently decodes inline image bytes even though its consumer only needs descriptors, and download planning repeats this work before materialization. Carry data-URL source references through discovery and leave decoding and byte-budget accounting with the downloader.

Preserve attachment/source ordering, malformed-image placeholders, per-item and total size limits, MIME/save failure budget accounting, and missing HTML attachment IDs. This is a private ownership cleanup with isolated channel-flow proof, not a live Teams or measured RSS claim.

steipete:
Closes #144295

## What Problem This Solves

Teams attachment discovery decodes inline images while describing advertised media, then repeats discovery and retains decoded buffers before materializing the download queue. The descriptor path needs source references and media facts, not image bytes.

## Why This Change Was Made

Carry inline data-URL references through discovery and decode each image at its existing materialization step. Reuse the unrepresented HTML attachment IDs directly instead of rebuilding all advertised media to obtain that suffix. The download owner retains canonicalization, individual and total byte limits, MIME validation, saving and unavailable-media outcomes.

## User Impact

Attachment ordering, source IDs, file-only messages, malformed data and size-limit handling remain covered. Accepted inline bytes still consume the message budget when MIME detection or saving fails. This removes 33 production lines and the eager decoded-buffer descriptors without changing a public API, SDK export, configuration or store contract.

## Evidence

Seven focused suites pass before (168 tests) and after (173 tests), including inbound facts, media and message handling. The full selected gate and fresh independent P2 review pass. The gate identified test formatting and a test-library typing mismatch; the latter now uses the existing SDK deferred helper, with all 41 affected tests passing before the final full gate.

The proof uses isolated attachment/runtime fixtures, not a live Teams service. No allocation, RSS or latency measurement is claimed. Exceptional parser/string-allocation failures still propagate but can occur after preceding remote downloads; ordinary malformed input, decoding and MIME/save failure behavior is preserved. Import edges, lazy entrypoints, package/SDK exports and dependencies are unchanged, so the scoped build rule did not require a normal build.

### DIFF
--- a/extensions/msteams/src/attachments.helpers.test.ts
+++ b/extensions/msteams/src/attachments.helpers.test.ts
@@ -121,17 +121,10 @@ describe("msteams attachment helpers", () => {
       expect(resolveMSTeamsAdvertisedMedia(attachments)).toEqual(expected);
     });
 
-    it("preserves an inline image fact when materialization limits reject it", () => {
-      const attachments = [
-        createHtmlAttachment(`<img src="data:image/png;base64,${"A".repeat(16)}" />`),
-      ];
+    it("preserves an inline image fact for malformed base64", () => {
+      const attachments = [createHtmlAttachment('<img src="data:image/png;base64,A!AA" />')];
 
-      expect(
-        resolveMSTeamsAdvertisedMedia(attachments, {
-          maxInlineBytes: 4,
-          maxInlineTotalBytes: 4,
-        }),
-      ).toEqual([{ kind: "image" }]);
+      expect(resolveMSTeamsAdvertisedMedia(attachments)).toEqual([{ kind: "image" }]);
     });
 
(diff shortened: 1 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8a5ca45554eaddb1ccc66f9b941b0bdd42edc178](https://github.com/openclaw/openclaw/pull/144296/commits/8a5ca45554eaddb1ccc66f9b941b0bdd42edc178)
- Landed commit: [5b097c8b41ae15cae25a88664e2eef7df67c89a5](https://github.com/openclaw/openclaw/commit/5b097c8b41ae15cae25a88664e2eef7df67c89a5)

PR #144298, opened 2026-09-10
refactor(scripts): avoid duplicate SDK source scans

steipete:
## What Problem This Solves

The SDK subpath checker walks `src` twice on every run: once to collect core runtime files and again while collecting its complete source inventory.

## Why This Change Was Made

Classify each already-collected file using the existing test-file predicate and normalized source path. Remove the redundant traversal and set, and reuse that path for diagnostics. Public/private SDK boundaries, erased type references, test suffixes and generated-file exclusions retain the same rules.

## User Impact

The checker does less filesystem enumeration while producing the same diagnostics. This removes five tooling lines without adding an API or dependency.

## Evidence

Actual complete-repository CLI output is byte-identical before and after. The same CLI/helper closure in a 24-file synthetic workspace also produces exactly the same 26 independently expected diagnostics, covering core runtime, package, public/private, erased/inline type, test suffix and generated-path distinctions.

A transparent counter of actual directory reads observes 474 to 237 reads under `src`; all 237 directories are read once instead of twice. Other directory counts are unchanged. Total reads fall from 1,546 to 1,309 with the same unique directories. Individual runs remained approximately 15.7 seconds, so no wall-time speedup is claimed.

All 17 canonical changed checks pass, and independent review found no actionable findings. The isolated CLI fixture is retained as review evidence.

### DIFF
--- a/scripts/check-plugin-sdk-subpath-exports.mts
+++ b/scripts/check-plugin-sdk-subpath-exports.mts
@@ -8,6 +8,7 @@ import { normalizeRepoPath, visitModuleSpecifiers } from "./lib/guard-inventory-
 import { resolveRepoRoot } from "./lib/repo-root.mjs";
 import {
   collectTypeScriptFilesFromRoots,
+  isTestLikeTypeScriptFile,
   resolveSourceRoots,
   toLine,
 } from "./lib/ts-guard-utils.mts";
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [a1d9816bfe93d64ab27aa5c2b826116cfa4b5b38](https://github.com/openclaw/openclaw/pull/144298/commits/a1d9816bfe93d64ab27aa5c2b826116cfa4b5b38)
- Landed commit: [decc0d9e05fba052a8a4bb3473923736e4b98e30](https://github.com/openclaw/openclaw/commit/decc0d9e05fba052a8a4bb3473923736e4b98e30)

PR #144305, opened 2026-09-10
perf(models): reuse the first-wins catalog identity index

### LINKED ISSUE #144304: Configured model catalogs rebuild the first-wins identity index
Configured model catalog construction currently discards the first-wins Map produced by deduplication, then recomputes each retained row's identity while rebuilding a second Map. Expose the existing index from the shared dedupe owner and consume it directly.

Preserve the array-returning helper contract, callback and insertion order, first-route precedence, identity normalization, fallbacks and current route-metadata/allowlist behavior. This needs no global cache or configuration/storage contract change.

steipete:
Closes #144304

## What Problem This Solves

Configured model catalog construction deduplicates captured rows into a Map, converts it to an array, then evaluates every retained row's identity again to build another Map. The same first-wins index can serve the catalog directly.

## Why This Change Was Made

Extract the existing Map loop as indexFirstByKey and keep dedupeByKey as its array-returning consumer. The configured catalog reuses the index while preserving input order, first-row precedence, key normalization and fallback behavior. This avoids the repeated identity pass and intermediate array/tuple/Map reconstruction without adding a cache.

## User Impact

Selected models and their route metadata remain unchanged. The four-file change adds two production lines and 38 net test lines to preserve the shared array contract and configured first-route behavior. Existing exact allowlist and route-metadata fixes on main are retained; no configuration, schema, persistence or migration contract changes.

## Evidence

The composed current-main change passes 205 focused tests across shared deduplication, model selection, manifest workspace and catalog-metadata overlay owners, plus the full selected changed gate and fresh independent P2 review. The composition was checked against current main to preserve all surrounding route and allowlist changes.

A bounded before/after run over those four focused files passed 205 tests in each arm and recorded 12.09 seconds / 2,122,252,288 bytes maximum RSS before and 11.00 seconds / 2,115,305,472 bytes afterward. These sequential correctness-run observations include imports, caches and worker setup; they do not establish a causal speed or RSS reduction. The supported code-level improvement is removal of repeated identity work.

### DIFF
--- a/src/agents/model-selection-shared.ts
+++ b/src/agents/model-selection-shared.ts
@@ -17,7 +17,7 @@ import { createSubsystemLogger } from "../logging/subsystem.js";
 import { getCurrentPluginMetadataSnapshot } from "../plugins/current-plugin-metadata-snapshot.js";
 import { loadManifestMetadataSnapshot } from "../plugins/manifest-contract-eligibility.js";
 import { getActivePluginRegistryWorkspaceDirFromState } from "../plugins/runtime-state.js";
-import { dedupeByKey } from "../shared/dedupe-by-key.js";
+import { dedupeByKey, indexFirstByKey } from "../shared/dedupe-by-key.js";
 import { resolveAgentConfig } from "./agent-scope-config.js";
 import { resolveConfiguredProviderFallback } from "./configured-provider-fallback.js";
 import { DEFAULT_PROVIDER } from "./defaults.js";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d3e4d73780fc321eaf40e1dfd8c32f73845f050d](https://github.com/openclaw/openclaw/pull/144305/commits/d3e4d73780fc321eaf40e1dfd8c32f73845f050d)
- Landed commit: [ec3ea375f2fcbada138772136113f6f0fe3c928c](https://github.com/openclaw/openclaw/commit/ec3ea375f2fcbada138772136113f6f0fe3c928c)

PR #144309, opened 2026-09-10
test(auth): share stale fallback profile fixtures

steipete:
Related: #139428

## What Problem This Solves

Three OAuth fallback tests repeat the same stale/healthy profile store, obscuring their distinct assertions about stale selection, locked profiles, and alternate-profile reporting.

## Why This Change Was Made

Extract the shared input into a local fixture builder that creates a fresh nested store for each call. Keep persistence, refresh-error injection, and every independent assertion at the original call sites, including the order of both expiry clock reads.

## User Impact

No runtime or authentication behavior changes. This removes 42 net test lines while retaining all existing cases and their selection and persistence contracts.

## Evidence

- The full affected file and two fallback/error siblings passed before and after: 50 identical ordered cases, no failures or skips, on Node 24.19.0.
- Expanding the helper recovers the original whole-file AST tokens, accounting explicitly for three trailing argument commas removed by formatting. Fresh nested objects and expiry clock order were verified.
- The mapped changed-file gate and fresh independent review passed. The initial gate stopped on formatting only; native formatting was applied before the final checks.

### DIFF
--- a/src/agents/auth-profiles/oauth.openai-codex-refresh-fallback.test.ts
+++ b/src/agents/auth-profiles/oauth.openai-codex-refresh-fallback.test.ts
@@ -113,6 +113,33 @@ async function readPersistedStore(agentDir: string): Promise<AuthProfileStore> {
   return readAuthProfileStoreForTest(agentDir);
 }
 
+function createStaleLastGoodStore(
+  staleProfileId: string,
+  healthyProfileId: string,
+): AuthProfileStore {
+  return {
+    version: 1,
+    profiles: {
+      [staleProfileId]: {
+        type: "oauth",
+        provider: "openai",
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [aede7366f5e45e985dab1f1cbad32a4313ce2247](https://github.com/openclaw/openclaw/pull/144309/commits/aede7366f5e45e985dab1f1cbad32a4313ce2247)
- Landed commit: [98efaa0bc14e4fb3ffd870d5d03f3c0e1d703e80](https://github.com/openclaw/openclaw/commit/98efaa0bc14e4fb3ffd870d5d03f3c0e1d703e80)

PR #144311, opened 2026-09-10
perf(cli): prepare ordered completion groups once

### LINKED ISSUE #144310: Shell completion repeatedly resolves and registers adjacent command groups
Shell-completion construction resolves subcommand registration groups repeatedly and invokes the same registrar for adjacent descriptors. Prepare the ordered group sequence once, coalescing only adjacent equivalent visits while retaining separated visits and warning names.

Preserve command order/replacement, awaited asynchronous plugin registration, aliases and exact generated shell scripts. This is internal registration work with no CLI/configuration/store contract change.

steipete:
Closes #144310

## What Problem This Solves

Building shell completions repeatedly reconstructs the subcommand registration groups and visits adjacent names owned by the same registrar. This repeats registration work while producing the same completion tree.

## Why This Change Was Made

Prepare the ordered completion groups once at their existing owner. Coalesce only adjacent visits to the same group; preserve separated visits, descriptor names used in warnings, command replacement and awaited asynchronous registration. The old descriptor-only accessor has no remaining callers and is removed.

## User Impact

Generated completions and command ordering remain unchanged, including asynchronous plugin commands and aliases. There are no new CLI options, configuration, schema, persistence or migration changes. The change adds nine production lines and 88 net test lines to protect grouping, ordering and async completion behavior; no overall startup, latency or RSS improvement is claimed.

## Evidence

The new original-source regression fails because the capability registrar runs twice instead of once; the final two test files pass all 47 tests. The supported CLI launcher then generated Bash, Fish, PowerShell and Zsh completion files before and after using the same isolated configuration and synthetic async plugin. All four files match byte-for-byte and by SHA-256, retain the plugin command/alias, leave the configuration unchanged and create no shell profiles.

The normal launcher performed its required runtime rebuilds. The full selected changed gate and fresh independent P2 review pass; the selected scope did not require a separate build. An initial test-local shadowing lint error was renamed and the affected tests, review and complete gate rerun passed. Production bytes remain identical to the successful CLI parity proof.

### DIFF
--- a/src/cli/completion-cli.ts
+++ b/src/cli/completion-cli.ts
@@ -31,7 +31,7 @@ import { publishOutputFileAtomically } from "./output-file.runtime.js";
 import { getCoreCliCompletionGroups } from "./program/command-registry-core.js";
 import { getProgramContext } from "./program/program-context.js";
 import { removeCommandGroupNames } from "./program/register-command-groups.js";
-import { getSubCliEntries, registerSubCliByNameCore } from "./program/register.subclis-core.js";
+import { getSubCliCompletionGroups } from "./program/register.subclis-core.js";
 
 export function getCompletionScript(shell: CompletionShell, program: Command): string {
   if (shell === "zsh") {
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7dd3624a34cfdc69a52da4b93dab6cfcf7f07b55](https://github.com/openclaw/openclaw/pull/144311/commits/7dd3624a34cfdc69a52da4b93dab6cfcf7f07b55)
- Landed commit: [990de0e2d8fe0b8fc09df88dbced6096b930c454](https://github.com/openclaw/openclaw/commit/990de0e2d8fe0b8fc09df88dbced6096b930c454)

PR #144312, opened 2026-09-10
fix(docs): validate examples against source plugin schemas

steipete:
## What Problem This Solves

Fixes an issue where docs configuration checks could use stale built plugin schemas when launched from a checkout subdirectory. The checker read current source docs, but its working directory could change which plugin metadata validated them, producing false errors or missed retired settings.

## Why This Change Was Made

Anchor bundled source manifests beside the checker's source schema. Keep the caller's docs-input root separate, so existing docs-only fixtures still use the current schema. The shared metadata resolver, separate validators and operator-state isolation retain their existing contracts.

## User Impact

The registered checker now gives the same configuration findings from the checkout root and `docs/`, even when stale built plugin metadata is present. It changes no stored data or schema and requires no migration.

## Evidence

A fixture supplies contrasting current-source and stale-built schemas plus one docs example. Before, the actual registered command correctly rejects the retired setting from the root but incorrectly accepts it from `docs/`. After, both runs reject it with byte-identical findings. The same seven fixture files are used throughout; plugin runtime sentinels never execute.

The permanent regression also proves that an alternate working-directory plugin tree cannot reject a supported source setting. All 17 docs audit/baseline tests pass, including operator-state isolation. A clean registered command from `docs/` passes 1,289 files and 1,224 validated fences. All 30 canonical changed checks pass, and independent review found no actionable findings.

This repair covers metadata selection within a source checkout. Separate TypeScript-loader behavior from a completely external working directory, and discovery after a working directory is deleted, remain outside this change. No live provider or Gateway was used.

### DIFF
--- a/src/config/docs-config-examples.test.ts
+++ b/src/config/docs-config-examples.test.ts
@@ -3,7 +3,8 @@ import os from "node:os";
 import path from "node:path";
 import { DatabaseSync } from "node:sqlite";
 import { fileURLToPath } from "node:url";
-import { describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
 import { OPENCLAW_STATE_SCHEMA_VERSION } from "../state/openclaw-state-db-contract.js";
 import { resolveOpenClawStateSqlitePath } from "../state/openclaw-state-db.paths.js";
 import {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [92da4877aebb1d61bc550edef6a29958e23ff47c](https://github.com/openclaw/openclaw/pull/144312/commits/92da4877aebb1d61bc550edef6a29958e23ff47c)
- Landed commit: [7018913aee8351543931f5e8a6c8d5e699d696da](https://github.com/openclaw/openclaw/commit/7018913aee8351543931f5e8a6c8d5e699d696da)

PR #137779, opened 2026-09-04
fix(agents): pass durable session key as visible-spawn parent lineage

SunnyShu0925:
Fixes #137690

## What Problem This Solves

When a user asks the bot to spawn a sub-agent from a Telegram DM, `sessions_spawn` fails with `unknown parent session` because the visible-spawn path sends the wrong session key to Gateway. After this fix, Telegram-originated sessions can spawn visible sub-agents the same way web dashboard sessions do.

- Problem: `createOpenClawTools` built the `sessions_spawn` tool with `agentSessionKey: options?.agentSessionKey` — the sandbox/policy key. For Telegram DMs this key includes an account-id segment (`agent:main:telegram:default:direct:<id>`), but the durable store entry is persisted under a different key, so Gateway's parent-session lookup fails.
- Solution: Pass `options?.runSessionKey ?? options?.agentSessionKey` (durable key preferred) for parent lineage, matching existing durable-session consumers, and separate lineage from sandbox classification, controller identity, and retained-task ownership so the substitution does not weaken hidden-spawn sandbox admission, lose run visibility, or drop retained task rows.
- What changed: (1) `createSessionsSpawnTool` uses the durable key for parent lineage. (2) Hidden native spawn forwards the explicit `opts.sandboxed` flag through `SpawnSubagentContext` to `resolveSubagentChildPlan`, which prefers it over key-derived status, mirroring the visible/ACP paths. (3) `createSubagentsTool` receives the same durable controller key the spawn tool registers runs under. (4) Image/video/music generators preserve existing ownership: the isolated run key for cron, otherwise the policy session key. (5) `createSubagentsTool` also receives `callerPolicySessionKey`, so both retained and new policy-owned tasks remain visible and cancellable for split-key callers. Plus tests.
- What did NOT change: the sandbox/policy key is still used for sandbox and tool-policy decisions; visible-spawn and ACP paths already preserved explicit sandbox state; the worker `createSessionsSpawnTool` call already used the durable key.

## Why This Change Was Made

The durable `runSessionKey` was already available in `createOpenClawTools` (passed only as `completionOwnerKey`). Using it for parent lineage aligns `sessions_spawn` with the existing `runSessionKey ?? agentSessionKey` pattern. The fix separates concerns: durable key for lineage, explicit `sandboxed` flag for admission, durable controller key for listing, and dual-key lookup that preserves policy-owned media tasks without migrating their ownership.

[...]

PR #144320, opened 2026-09-10
fix(sessions): sessions_send treats visible dashboard spawn children as foreign sessions and starts a stale announce run

### LINKED ISSUE #144265: [Bug]: sessions_send misclassifies visible dashboard spawn-child sessions and triggers stale runtime announce failures
### Bug type

Regression / session lifecycle bug

### Summary

On OpenClaw 2026.9.3, a visible native child created through `sessions_spawn` uses a persistent `dashboard:` session key.

When its parent later calls `sessions_send` to that child, `sessions_send` fails to recognize the dashboard session as a native child and starts the generic A2A announce flow.

That extra announce run re-enters the same child session and can fail with:

```text
prepared model runtime plugin generation was superseded for /home/node/.openclaw/agents/penny/agent
```

The original child run itself completes successfully. The failure occurs in the unnecessary follow-up `sessions_send` announce flow.

### Environment

- OpenClaw: `2026.9.3 (1391f7c)`
- Container image: `ghcr.io/openclaw/openclaw:2026.9.3-browser`
- Docker on TrueNAS / Linux
- Native OpenClaw agents, not ACP

### Child session metadata

The affected target session is clearly a native spawned child:

```json
{
  "key": "agent:penny:dashboard:<uuid>",
  "spawnedBy": "agent:main:dashboard:<uuid>",
  "parentSessionKey": "agent:main:dashboard:<uuid>",
  "spawnDepth": 1,
  "kind": "spawn-child",
  "acpRuntime": false,
  "createdActor": {
    "type": "agent",
    "id": "main"
  }
}
```

So the persisted session lineage correctly identifies this as a child of the requester.

### Observed behavior

The first target run completes normally:

```text
[agent:nested] session=agent:penny:dashboard:<uuid> run=<run1> ...
[agents/agent-command] [agent] run <run1> ended with stopReason=stop
```

Later, `sessions_send` runs an additional announce step against the same child session.

The lane then shows approximately a 30-second wait:

```text
[diagnostic] lane wait exceeded:
lane=session:agent:penny:dashboard:<uuid>
waitedMs=29001
queueAhead=0
activeAhead=1
```

followed by:

```text
[diagnostic] lane task error:
lane=nested:agent:penny:dashboard:<uuid>
error="prepared model runtime plugin generation was superseded for /home/node/.openclaw/agents/penny/agent"

[model-fallback/decision] ...
detail=prepared model runtime plugin generation was superseded ...

[agents/sessions-send] sessions_send announce flow failed
```

The fallback models also fail because the failure happens during prepared-runtime admission rather than provider inference.

### Probable root cause

In `sessions-send-tool.ts`, native parent/child suppression currently depends on the session key:

```ts
function isRequesterParentOfNativeSubagentSession(params) {
  if (
    !params.entry ||
    params.acpMeta ||
    params.entry.acp ||
    !isSubagentSessionKey(params.targetSessionKey)
  ) {
    return false;
  }

  const requester = normalizeOptionalString(params.requesterSessionKey);
  if (!requester) return false;

  const spawnedBy = normalizeOptionalString(params.entry.spawnedBy);
  const parentSessionKey = normalizeOptionalString(params.entry.parentSessionKey);

  return requester === spawnedBy || requester === parentSessionKey;
}
```

But `isSubagentSessionKey()` only recognizes keys whose session rest begins with `subagent:`:

```ts
function isSubagentSessionKey(sessionKey) {
  const raw = normalizeOptionalString(sessionKey);
  if (!raw) return false;
  if (normalizeOptionalLowercaseString(raw)?.startsWith("subagent:")) return true;
  const parsed = parseAgentSessionKey(raw);
  return normalizeOptionalLowercaseString(parsed?.rest)?.startsWith("subagent:") === true;
}
```

A visible native child has a key such as:

```text
agent:penny:dashboard:<uuid>
```

Therefore:

```text
isSubagentSessionKey(targetSessionKey) === false
```

even though its persisted metadata says:

```text
kind = spawn-child
spawnDepth = 1
spawnedBy = requester
parentSessionKey = requester
acpRuntime = false
```

As a result:

```text
skipNativeParentA2AFlow = false
```

and `sessions_send` incorrectly starts `runSessionsSendA2AFlow()`.

That flow subsequently executes:

```ts
runAgentStep({
  agentId: params.targetAgentId,
  sessionKey: params.targetSessionKey,
  message: "Agent-to-agent announce step.",
  lane: resolveNestedAgentLaneForSession(params.targetSessionKey),
  sourceTool: "sessions_send"
});
```

This creates an unnecessary second run in the visible native child session.

### Expected behavior

A visible dashboard session whose persisted metadata identifies it as a native `spawn-child` of the requester should receive the same native-parent handling as an `agent:<id>:subagent:<uuid>` session.

`sessions_send` should not start the generic A2A announce/ping-pong flow for this parent → native-child relationship.

The original child result should be sufficient and the child should not be re-entered by an unnecessary announce run.

### Suggested fix

Do not rely solely on `isSubagentSessionKey()` to determine whether the target is a native child.

Use the persisted session envelope / lineage metadata, for example conceptually:

```ts
const isNativeChild =
  isSubagentSessionKey(params.targetSessionKey) ||
  params.entry.kind === "spawn-child";

if (!isNativeChild) return false;

return requester === spawnedBy || requester === parentSessionKey;
```

while continuing to exclude ACP sessions.

Ideally this should use the same shared native-child/session classifier used elsewhere rather than introducing another classification rule.

### Why this matters

Visible sub-agents are intentionally persistent dashboard sessions, so `dashboard:` keys are a normal supported representation of native spawned children.

Using key shape rather than persisted lineage causes parent/child semantics to diverge between hidden `subagent:` sessions and visible `dashboard:` sessions.

It also turns an otherwise successful delegated run into a failed task and can cause duplicate model work/fallback attempts.

### Related issues

There are existing prepared-runtime supersession issues involving config publication / cron admission, but this reproduction has a different trigger: the unnecessary `sessions_send` A2A announce run caused by visible child session classification.

ylcn91:
Closes #144265

## What Problem This Solves

Fixes an issue where an agent that spawned a visible child with `sessions_spawn` (`visible: true`) and later sent it a waited `sessions_send` would trigger an extra agent-to-agent announce run against that child. The child's reply already came back inline, so the follow-up run was pure overhead, and on 2026.9.3 it could fail with `prepared model runtime plugin generation was superseded` after a roughly 30 second lane wait, logged as `sessions_send announce flow failed`.

## Why This Change Was Made

Two things kept `sessions_send` from recognizing its own visible child:

1. The native-parent guard first required the target key to be a `subagent:` key. Visible children live under persistent `dashboard:` keys, so their persisted lineage (`spawnedBy` equal to the requester) was never consulted. The guard now treats a `spawnedBy` match as authoritative regardless of key shape and keeps the `parentSessionKey` match limited to subagent keys, because that field also records ordinary UI threading and forks.
2. `createOpenClawTools` handed `sessions_send` only the channel route key (`agentSessionKey`), while `sessions_spawn` and every other session tool already receive the durable run session key (`runSessionKey ?? agentSessionKey`, see #137779). The Gateway canonicalizes the spawn parent into `spawnedBy` under that durable key, so for channel-originated parents (a DM mapped to the main session) the requester and `spawnedBy` never matched even with the guard fixed. `sessions_send` now receives the same key as its siblings.

3. Suppressing the flow is only right when the child's reply actually returned inline. A send to a spawned child is not a registered spawn run, so when the wait expires before the child finishes, the detached continuation is the only path that brings the late reply back to the parent. The native-parent skip now applies to inline replies only; a timed-out send reports `delivery.status: "pending"` and keeps its continuation (raised by ClawSweeper as a P1 on this PR).

ACP sessions, fire-and-forget sends, and unrelated senders are unchanged. Dashboard threads and forks that were not spawned by the requester still go through the normal A2A path.

## User Impact

Parents that spawn visible children and wait on `sessions_send` get the inline reply without a second announce run against the child: no stale runtime re-entry, no 30 second lane wait, no spurious `announce flow failed` warning. This also holds for parents running from a channel DM, not only for dashboard-to-dashboard spawns.

## Evidence

- Regression `sessions_send skips duplicate A2A delivery for waited visible spawn children on dashboard keys` (`src/agents/openclaw-tools.sessions.test.ts`) fails on main with `expected 'pending' to be 'skipped'` and passes with the fix.
- Boundary test `sessions_send keeps the A2A flow for dashboard threads that were not spawned by the requester` passes on main and with the fix.
- Regression `sessions_send keeps the delayed continuation when a visible spawn child outlives the wait` fails before the third commit with `expected 'skipped' to be 'pending'` and passes with it; the terminal-timeout and inline-reply tests are unchanged.
- Wiring regression `passes the durable runSessionKey to createSessionsSendTool when keys differ` (`src/agents/tools/openclaw-tools.spawn-session-key.test.ts`) fails on main (`expected "vi.fn()" to be called with arguments`) and passes with the fix.
- `node scripts/run-vitest.mjs src/agents/openclaw-tools.sessions.test.ts src/agents/tools/sessions-send-tool.a2a.test.ts src/agents/tools/openclaw-tools.spawn-session-key.test.ts`: 47 + 28 + 7 passed.
- `pnpm check:changed` on the final head (typecheck all lanes, type-aware lint, dead-export scan, changed-lane Vitest shards): passed.
- Real Gateway proof (built `dist/index.js` at this branch head, `createQaGatewayChild` with a loopback mock Responses provider, QA channel DM as the parent): turn 1 spawns a visible child through the real `sessions_spawn` tool call, turn 2 issues a waited `sessions_send` to the persisted `agent:qa:dashboard:<uuid>` key. Recorded run, redacted to loopback values only:

```text
$ OPENCLAW_PROOF_VISIBLE_CHILD_SEND=1 OPENCLAW_E2E_VERBOSE=1 node scripts/run-vitest.mjs run \
    --config test/vitest/vitest.e2e.config.ts \
    test/e2e/qa-lab/runtime/sessions-send-visible-child.product-proof.e2e.test.ts

{"phase":"sessions-send-visible-child",
 "childKey":"agent:qa:dashboard:5e20d14c-2b95-4ff0-b54b-ca54f841dd60",
 "childSpawnedBy":"agent:qa:main","childParentSessionKey":"agent:qa:main",
 "childRequestsAfterSpawn":1,"childRequestsTotal":2,"childReplyStepRequests":0,
 "parentRequests":5,"parentReplyStepRequests":0,
 "sendDeliveryStatus":"skipped","sendInlineReply":"QA-VISIBLE-CHILD-OK",
 "announceFlowLogLines":0}
 ✓ returns the waited reply inline without an announce run against the child 154955ms

sessions_send tool result seen by the parent model (excerpt):
  "status": "ok",
  "delivery": { "status": "skipped", "mode": "announce" },
  "reply": "QA-VISIBLE-CHILD-OK"
```

  The same lane run against the guard-only change (before adding the factory wiring) showed `parentReplyStepRequests: 1` and one `sessions_send announce flow failed` line, which is how the second cause was found: the guard received the raw route key `agent:qa:qa-channel:default:direct:<id>` while the child was persisted with `spawnedBy: agent:qa:main`. The lane is committed as `test/e2e/qa-lab/runtime/sessions-send-visible-child.product-proof.e2e.test.ts`, gated behind `OPENCLAW_PROOF_VISIBLE_CHILD_SEND=1`, following the existing product-proof lanes in that directory.
- `autoreview` (Codex, P2 threshold) on the final four-commit branch: scoped-clean, no findings.
- CI: the hosted shard `checks-node-compact-large-18` failed once per push on two unrelated timing tests, `chat.abort-cascade.test.ts` (`residual registry roots`, `vi.waitFor`) and then `code-mode-headless.test.ts` (`expected 'internal_error' to be 'timeout'`, wall-clock deadline). Neither file is touched here; both pass locally on this branch and on main. The sessions test file resets Gateway work admission after each test, so the new tests do not leak root work into later files.

### DIFF
--- a/src/agents/openclaw-tools.sessions.test.ts
+++ b/src/agents/openclaw-tools.sessions.test.ts
@@ -2438,6 +2438,45 @@ describe("sessions tools", () => {
     expect(countMatching(calls, (call) => call.method === "agent")).toBe(1);
   });
 
+  it("sessions_send keeps the delayed continuation when a visible spawn child outlives the wait", async () => {
+    const calls: Array<{ method?: string; params?: unknown }> = [];
+    const requesterKey = "agent:main:dashboard:parent-uuid";
+    const targetKey = "agent:penny:dashboard:child-uuid";
+    loadSessionEntryByKeyMock.mockImplementation((sessionKey: string) =>
+      sessionKey === targetKey
+        ? { sessionId: "child-session", updatedAt: 1, spawnedBy: requesterKey, spawnDepth: 1 }
+        : undefined,
+    );
+    callGatewayMock.mockImplementation(async (opts: unknown) => {
+      const request = opts as { method?: string; params?: unknown };
+      calls.push(request);
+      if (request.method === "agent") {
+        return { runId: "run-slow-child", status: "accepted", acceptedAt: 2000 };
+      }
+      if (request.method === "agent.wait") {
+        // The child is still running: a nonterminal timeout, not a failed run.
+        return { runId: "run-slow-child", status: "timeout" };
+      }
+      return {};
+    });
+
+    const tool = getSessionTool("sessions_send", { agentSessionKey: requesterKey });
+    const result = await tool.execute("call-slow-visible-child", {
+      sessionKey: targetKey,
+      message: "ping",
+      timeoutSeconds: 1,
+    });
+
+    const details = sessionsSendDetails(result.details);
+    expect(details.status).toBe("accepted");
+    // The late reply still has a delivery path; only inline replies skip it.
+    expect(details.delivery?.status).toBe("pending");
+    await new Promise<void>((resolve) => {
+      setImmediate(resolve);
+    });
+    expect(countMatching(calls, (call) => call.method === "agent.wait")).toBeGreaterThan(1);
+  });
+
   it("sessions_send keeps the A2A flow for dashboard threads that were not spawned by the requester", async () => {
     const requesterKey = "agent:main:dashboard:parent-uuid";
     const targetKey = "agent:main:dashboard:thread-uuid";
--- a/src/agents/openclaw-tools.sessions.test.ts
+++ b/src/agents/openclaw-tools.sessions.test.ts
@@ -2392,6 +2392,87 @@ describe("sessions tools", () => {
     expect(calls.some((call) => call.method === "send")).toBe(false);
   });
 
+  it("sessions_send skips duplicate A2A delivery for waited visible spawn children on dashboard keys", async () => {
+    const calls: Array<{ method?: string; params?: unknown }> = [];
+    const requesterKey = "agent:main:dashboard:parent-uuid";
+    const targetKey = "agent:penny:dashboard:child-uuid";
+    loadSessionEntryByKeyMock.mockImplementation((sessionKey: string) =>
+      sessionKey === targetKey
+        ? {
+            sessionId: "child-session",
+            updatedAt: 1,
+            spawnedBy: requesterKey,
+            parentSessionKey: requesterKey,
+            spawnDepth: 1,
+          }
+        : undefined,
+    );
+    callGatewayMock.mockImplementation(async (opts: unknown) => {
+      const request = opts as { method?: string; params?: unknown };
+      calls.push(request);
+      if (request.method === "agent") {
+        return { runId: "run-child", status: "accepted", acceptedAt: 2000 };
+      }
+      if (request.method === "agent.wait") {
+        return {
+          runId: "run-child",
+          status: "ok",
+          terminalReply: { disposition: "visible", text: "child reply" },
+        };
+      }
+      return {};
+    });
+
+    const tool = getSessionTool("sessions_send", { agentSessionKey: requesterKey });
+
+    const waited = await tool.execute("call-parent-owned-dashboard-child", {
+      sessionKey: targetKey,
+      message: "ping",
+      timeoutSeconds: 1,
+    });
+
+    const waitedDetails = sessionsSendDetails(waited.details);
+    expect(waitedDetails.status).toBe("ok");
+    expect(waitedDetails.reply).toBe("child reply");
+    expect(waitedDetails.delivery?.status).toBe("skipped");
+    expect(countMatching(calls, (call) => call.method === "agent")).toBe(1);
+  });
+
+  it("sessions_send keeps the A2A flow for dashboard threads that were not spawned by the requester", async () => {
+    const requesterKey = "agent:main:dashboard:parent-uuid";
+    const targetKey = "agent:main:dashboard:thread-uuid";
+    loadSessionEntryByKeyMock.mockImplementation((sessionKey: string) =>
+      sessionKey === targetKey
+        ? { sessionId: "thread-session", updatedAt: 1, parentSessionKey: requesterKey }
+        : undefined,
+    );
+    callGatewayMock.mockImplementation(async (opts: unknown) => {
+      const request = opts as { method?: string };
+      if (request.method === "agent") {
+        return { runId: "run-thread", status: "accepted", acceptedAt: 2000 };
+      }
+      if (request.method === "agent.wait") {
+        return {
+          runId: "run-thread",
+          status: "ok",
+          terminalReply: { disposition: "visible", text: "thread reply" },
+        };
+      }
+      return {};
+    });
+
+    const tool = getSessionTool("sessions_send", { agentSessionKey: requesterKey });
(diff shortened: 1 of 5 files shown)


obviyus:
Thanks @ylcn91!

**Proof:** Merged #147394 supersedes this PR and keeps your four commits under your authorship in its PR history. Its two-turn direct-message test confirms one inline reply without another parent reply step.

- **Bug:** A waited send to an owned visible child started extra delivery work after its reply had returned.
- **Fix:** #147394 uses durable spawn lineage and preserves late replies when the wait expires.

| Before | After |
| --- | --- |
| Inline reply followed by pending delivery and extra work | Inline reply once, with late replies retained when needed |

PR #144323, opened 2026-09-10
refactor(google): share image provider metadata

steipete:
## What Problem This Solves

Google image-provider registration and runtime construction maintained separate copies of the same model and capability metadata. That duplication could let the advertised image limits or supported geometry drift between the two paths.

## Why This Change Was Made

Reuse the existing Google-owned generation metadata module for both constructions. Each call still creates fresh nested objects and arrays. The lazy registration still has no own `isConfigured` property; the runtime retains its existing callback and property order before `capabilities`. Lazy loading, generation, authentication, request options, and fallback behavior are unchanged.

Production: **+47/-72, net 25 lines removed**. Tests: **+98/-0**. No core/provider-policy, configuration, dependency, storage, protocol, or public API change.

## User Impact

No intended operator-visible behavior change. The same two models, four-result limit, five-reference edit limit, and geometry metadata remain available through the registered provider and image-provider CLI listing.

## Evidence

- Original, candidate, and restored-candidate selections each passed all 99 Google image/registration/index/music/video cases, with the same per-file case inventory and order. Separate limit and shared-mutable-metadata fault controls failed the intended retained assertions; exact restoration passed.
- Normal original and candidate full builds passed all 16 phases. Four direct built-CLI cases through the canonical isolated test instance passed, with byte-identical original/candidate stdout and stderr for unconfigured and explicitly selected Google image-provider listings. This exercised `node dist/index.js`, not the development wrapper or live image generation.
- All 25 configured changed checks passed on Linux Node 24.19.0, including both plugin type graphs, three dead-export scans, lint, six Doctor-contract tests, and import-cycle checks. [Configured Testbox run](https://github.com/openclaw/openclaw/actions/runs/34516686158). Its native result was successful and its lease and temporary checkout were cleaned up; the backing Actions projection is not a substitute for that delegated result.
- Managed P2 precommit review and a separate exact-commit P2 review completed without actionable findings. The signed commit's full tree matches the tested four-file candidate.

Proof boundaries are retained: an earlier run that cleared inherited Node settings is excluded from the accepted comparison. The paired CLI fixtures emitted the same warning while normalizing a retired `default` marker; they were not changed between variants. No live Google availability, output-quality, Gateway-session, or development-wrapper claim is made. Exact-head hosted PR CI is separate and will attach after publication.

### DIFF
--- a/extensions/google/generation-provider-metadata.ts
+++ b/extensions/google/generation-provider-metadata.ts
@@ -1,4 +1,5 @@
 // Google provider module implements model/runtime integration.
+import type { ImageGenerationProvider } from "openclaw/plugin-sdk/image-generation";
 import type { MusicGenerationProvider } from "openclaw/plugin-sdk/music-generation";
 import { isProviderApiKeyConfigured } from "openclaw/plugin-sdk/provider-auth";
 import type {
@@ -6,6 +7,9 @@ import type {
   VideoGenerationProvider,
   VideoGenerationProviderConfiguredContext,
 } from "openclaw/plugin-sdk/video-generation";
+
+export const DEFAULT_GOOGLE_IMAGE_MODEL = "gemini-3.1-flash-image";
+export const GOOGLE_MAX_IMAGE_RESULTS = 4;
 
 export const DEFAULT_GOOGLE_MUSIC_MODEL = "lyria-3-clip-preview";
 export const GOOGLE_PRO_MUSIC_MODEL = "lyria-3-pro-preview";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b295a5133749672e2aaf7819bbf27c6e0d9e69f3](https://github.com/openclaw/openclaw/pull/144323/commits/b295a5133749672e2aaf7819bbf27c6e0d9e69f3)
- Landed commit: [47e174b73eda683d2b7751e2ea614d2224c4f9dc](https://github.com/openclaw/openclaw/commit/47e174b73eda683d2b7751e2ea614d2224c4f9dc)

PR #144328, opened 2026-09-10
refactor(update): retire obsolete post-activation stop path

steipete:
Related: #140339

## What Problem This Solves

Completes the managed-update lifecycle cutover in #140339. Updates now keep plugin convergence inside the original stopped interval, or park an already-current core before Doctor; obsolete post-activation stop handling still remained alongside that flow.

## Why This Change Was Made

Remove the unused private `activatedInstall` input and its alternate stop path, then retire the updated-install helper's now-unused `stop` action. Native stop keeps the same revalidated environment, authority callback, mutation reporting, and Windows compensation. The public `gateway stop` command and migrated-finalizer input remain unchanged.

The associated tests retain Git/package activation after a newer config stamp, fresh Doctor, stale-parent version guards, and install/restart process ownership. The complete change removes 27 production and 59 test/support code lines.

## User Impact

No intended user-visible behavior change. Managed updates retain their existing stop, activation, recovery, and verification behavior with one fewer private execution path to maintain.

## Evidence

- Original source: **520/520 passed**; candidate: **512/512 passed** across the same 18 selected files. Eight mode/stop-outcome rows become two retained activation/Doctor cases, and two private-stop process cases retire. All 510 other case identities passed in both runs, including 21 selected public-stop cases.
- Production/test typechecks, targeted lint, all three dead-export scans, and the complete 29-step changed gate passed. The test runner's original and candidate runtime builds also passed. Independent P2 review found no actionable findings.
- On Linux/Node 24.19.0, the integrity-pinned published OpenClaw 2026.9.3 updater modules were loaded before replacing their task-owned files. Retained old code accepted synthetic replacement public `gateway stop --force --json` responses and a matching migrated-worker run receipt, rejected a wrong terminal run ID, and preserved data-only recovery metadata. This proves the historical module/process boundary; it is not a full upgrade or live Gateway test.

### DIFF
--- a/src/cli/update-cli/update-command-service-command.process.test.ts
+++ b/src/cli/update-cli/update-command-service-command.process.test.ts
@@ -5,12 +5,10 @@ import { formatCliProcessFailure, runCliProcessChild } from "../cli-process-chil
 it.each([
   "restart",
   "install",
-  "stop",
   "missing candidate",
   "executor revoked",
   "restart revoked",
   "install revoked",
-  "stop revoked",
 ] as const)("handles %s after replacing the updater's module files", async (scenario) => {
   await withOpenClawTestState(
     { prefix: "openclaw-update-command-replacement-", scenario: "minimal", applyEnv: false },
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d074e79a3a70ac17e49d2adb738592a6f9917671](https://github.com/openclaw/openclaw/pull/144328/commits/d074e79a3a70ac17e49d2adb738592a6f9917671)
- Landed commit: [2f0a11096e1f68352ca70ff114efa942b6bebf0b](https://github.com/openclaw/openclaw/commit/2f0a11096e1f68352ca70ff114efa942b6bebf0b)

PR #144332, opened 2026-09-10
fix(doctor): preserve selected model context in tool diagnostics

steipete:
Related: #130706

## What Problem This Solves

Doctor could check active tools using a different model's context window when
the configured primary used a manifest alias. With entry → middle → final,
the diagnostic labeled tools as middle but used final's 4,096-token context
instead of middle's 32,000-token context.

## Why This Change Was Made

Doctor already selects the primary under its captured plugin metadata scope.
Its runtime metadata lookup now preserves that selected model ID, avoiding a
second alias pass while retaining provider ownership and transport resolution.

## User Impact

Active tool diagnostics use the metadata belonging to the selected primary
model. The fix adds one production line and a regression at the tool
construction boundary.

## Evidence

- The regression failed on the baseline with context 4,096 instead of 32,000;
  all 56 focused Doctor and preview tests passed after the fix.
- The complete changed gate passed against the pinned base. Independent P2
  review was clean, and a normal full build passed with SDK declarations and
  Control UI assets.
- Actual compiled `doctor --non-interactive --no-workspace-suggestions` ran
  in fresh owned state with a plugin installed through the real CLI. The
  public schema hook observed the expected selected/materialized IDs and
  context sizes in all four cases below, with 39 tools each. Every Doctor
  command completed and exited successfully.

| Authored primary | Selected / materialized model | Context |
| --- | --- | ---: |
| Alias entry | middle / middle | 32,000 |
| Raw middle alias | final / final | 4,096 |
| Exact ID | exact / exact | 16,384 |
| Custom provider with native API owner | exact / exact; custom provider preserved | 24,576 |

The compiled-only source-import rejection control passed. All 12,239 outputs
across the build root and 15 workspace-package output roots remained unchanged,
and every owned child process exited. This is real CLI/Doctor diagnostic proof
with synthetic plugins, not authenticated provider inference.

### DIFF
--- a/src/commands/doctor/shared/active-tool-schema-warnings.test.ts
+++ b/src/commands/doctor/shared/active-tool-schema-warnings.test.ts
@@ -2,6 +2,9 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import type { createOpenClawCodingTools } from "../../../agents/agent-tools.js";
 import type { AnyAgentTool } from "../../../agents/tools/common.js";
+import type { OpenClawConfig } from "../../../config/types.openclaw.js";
+import { createPluginMetadataSnapshotFixture } from "../../../plugins/plugin-metadata.test-support.js";
+import { withPluginRuntimeGenerationScope } from "../../../plugins/runtime/generation-scope.js";
 import { setPluginToolMeta } from "../../../plugins/tool-metadata.js";
 
 const toolState = vi.hoisted(() => ({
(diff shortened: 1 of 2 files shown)


steipete:
@clawsweeper review

Please review the current head. The one-line Doctor fix has passed the full changed gate, independent review, full build, four compiled Doctor command cases, and exact-head CI.

steipete:
The data-model compatibility checklist does not apply to this change. The entire production diff is the existing `modelIdSource: "selected"` option at Doctor's diagnostic resolver call. It adds no schema, migration, config format, persistence operation, or storage policy. The selected model already exists before this lookup; the fix keeps its identity while reading diagnostic metadata.

All four compiled Doctor cases completed successfully, and each case's config SHA-256 before and after the command is identical. The existing resolver's provider/API ownership and raw-input controls also passed. No migration or backfill is needed. The review reports no actionable correctness or security findings, and exact-head CI is green.

steipete:
Merged via squash.

- Prepared head SHA: [d28af47dde0596a3f4b63aaa29b9b148ee0d9c08](https://github.com/openclaw/openclaw/pull/144332/commits/d28af47dde0596a3f4b63aaa29b9b148ee0d9c08)
- Landed commit: [93b360883619e8c94bdb5cba502dbeee863a2d61](https://github.com/openclaw/openclaw/commit/93b360883619e8c94bdb5cba502dbeee863a2d61)

PR #144343, opened 2026-09-10
perf(ui): adopt completed transcript indexes without copying

### LINKED ISSUE #144341: Transcript projections copy already completed lookup indexes
The chat projection finishes two message-to-row indexes before handing them to the transcript virtualizer. The virtualizer then copies both maps and creates a membership set from one of them on every projection.

The projection never writes either index after the handoff. Reusing those completed maps can remove the repeated collection construction while preserving the candidate/committed snapshot boundary, touch-held history prepends, and MCP teardown gating.

A focused run through the real renderer with 800 persisted messages observed two map constructions and one set construction at this boundary on each of five projections. The proposed cleanup should remove those copies while preserving message lookup and reader position.

steipete:
Closes #144341

## What Problem This Solves

The transcript virtualizer copies message lookup maps and builds a membership set even though each projection has already completed those indexes. This repeats collection work while presenting the same rows.

## Why This Change Was Made

Adopt the completed projection indexes and use map membership for prepend anchoring. Rows and lookup maps still commit together through the existing MCP unmount gate; a touch-held prepend keeps the previous presentation snapshot visible until commit.

## User Impact

Transcript content, message lookup, prepend behavior and reader position remain unchanged. The change adds two production lines and five test lines without a new cache, API, configuration or persistence contract.

## Evidence

Five actual renderer invocations over 800 synthetic persisted messages each avoid two map constructions and one set construction per projection: 2,400 copied collection entries at this boundary. All 800 rendered message IDs remain identical. This is an explicit constructor/entry count, not total allocation bytes, retained heap, RSS or timing evidence.

The same 51 existing tests plus one disposable measurement case passed before and after. The held-prepend test also checks that a candidate-only message ID stays out of the active lookup until commit. Two existing browser cases passed on both versions, covering missing persisted IDs and touch momentum, with zero reader-adjusted anchor drift and byte-identical inspected screenshots. All 20 selected checks and independent P2 review pass; no package, runtime-import or lazy-loading boundary changed.

### DIFF
--- a/ui/src/pages/chat/components/chat-transcript-controller.test.ts
+++ b/ui/src/pages/chat/components/chat-transcript-controller.test.ts
@@ -204,7 +204,10 @@ describe("chat transcript controller", () => {
           await vi.advanceTimersByTimeAsync(200);
         }
         session.syncMessageRows(
-          new Map([["retained", "expanded-row"]]),
+          new Map([
+            ["older", "expanded-row"],
+            ["retained", "expanded-row"],
+          ]),
           new Map([
             ["older", "expanded-row"],
             ["retained", "expanded-row"],
@@ -216,6 +219,7 @@ describe("chat transcript controller", () => {
         ]);
         expect(container.textContent).not.toContain("older");
         expect(session.activeMessageId(["retained"])).toBe("retained");
+        expect(session.activeMessageId(["older"])).toBeNull();
         container.dispatchEvent(new Event("touchend"));
         renderRows(next);
         expect(transcriptRows(container).map((row) => row.dataset.virtualRowKey)).toEqual([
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0de0852168977ff6e554dc46d16e5a2a55a12fc2](https://github.com/openclaw/openclaw/pull/144343/commits/0de0852168977ff6e554dc46d16e5a2a55a12fc2)
- Landed commit: [5d3cc203b67d4a8314a4f7df119538b7ac9128f8](https://github.com/openclaw/openclaw/commit/5d3cc203b67d4a8314a4f7df119538b7ac9128f8)

PR #144348, opened 2026-09-10
docs: correct accuracy defects across the verified backlog

vincentkoc:
Corrects accuracy defects across the ledger's `verified` backlog. 49 files, +364/−182.

**This PR is a salvage, and I want that on the record.** The agent that produced these edits stopped before writing its report, so there is no verdict table mapping each edit to a finding. I validated the tree it left, reconstructed attribution from the diff, verified every release number it added, and dropped the work I could not stand behind. Treat the id list below as "these rows' cited files are in this diff", not as "each of these rows is closed".

## Why this backlog exists at all

These rows sat in a status meaning a second auditor independently re-confirmed the finding was real — still-open work that no round had in scope. A stratified sample put `accuracy` at a **77% live rate**, the highest of any kind, and that held up: these are defects where the documentation contradicts shipped code.

## What I dropped, and why

**`docs/plugins/architecture.md` is reverted out of this PR.** The agent added two release citations there — "since `2026.3.22`" and "retired in `2026.4.27`" — and left no evidence for either. Both tags exist, but existing is not the same as being the right release, and I could not reconstruct the commits behind the claims. That page is row `r3-0610`, which the triage flagged as *the weakest live row in the set*, and it sits next to a standing hold recorded earlier in this programme: **naming a wrong version is worse than naming none.** This programme has already shipped two wrong version numbers and been corrected by review both times. Not a third.

**One rebase conflict resolved against this branch.** `docs/concepts/typebox.md` — the branch added "Do not sort fragment keys or move existing entries: native code generation follows registry insertion order", while `main` had just replaced that whole mechanism ("This replaces the former manually ordered registration fragments"). Landing it would have re-documented something the default branch removed. Taken from `main`.

## The release citation that did verify

`docs/channels/matrix-migration.md` — the removed rust-crypto-store migration is `7a456e362d0` (#104650). Tag containment gives first containing tag `v2026.7.2-beta.1` and first suffix-free tag `v2026.8.1`; both are named, per the form at `docs/reference/database-schemas/agent-schema-history.md:13`. Tags were fetched before the query — an earlier PR in this programme cited a wrong beta precisely because the local tag set was incomplete when the containment loop ran.

## One anchor fixed en route

`docs/tools/lobster.md` gained a cross-reference to `### Important limitation: embedded Lobster vs \`openclaw.invoke\``. The docs link auditor accepted the generated fragment; `markdownlint`'s `MD051` did not, because the two slugify a backticked, dotted heading differently. Rather than pick a spelling and hope, the section now carries an explicit `<a id="embedded-lobster-vs-openclaw-invoke" />` stub and the link targets that. It mints a new id and moves none, so no published anchor changes, and both checkers agree.

## Rows whose cited file appears in this diff (29 of 91)

r3-0002, r3-0058, r3-0066, r3-0075, r3-0085, r3-0087, r3-0092, r3-0093, r3-0165, r3-0174, r3-0228, r3-0237, r3-0252, r3-0266, r3-0273, r3-0274, r3-0298, r3-0352, r3-0410, r3-0437, r3-0453, r3-0462, r3-0483, r3-0497, r3-0539, r3-0559, r3-0844, r6-0013, r6-0016

Four were individually confirmed live by the triage and are worth naming: **r3-0165** (`docs/cli/path.md` — `bash` fences interleaving `$ ` prompts with output, so copying does not run), **r3-0174** (`docs/cli/claws.md` — examples consume a `.claw.json` no earlier step creates), **r3-0844** (`docs/tools/lobster.md` — the first `openclaw.invoke` example precedes the "not reliable in the embedded runner" caveat that qualifies it), **r6-0016** (`docs/gateway/security/secrets-and-storage.md` — one sqlite path listed twice with contradictory Contents cells).

The remaining 62 accuracy rows in the backlog are untouched by this PR.

## Validators

`pnpm check:docs` exit 0, re-run after the rebase — format-docs clean (1281 files), markdownlint 1280 files / 0 issues, MDX, i18n glossary, link audit, config fences. `pnpm format:check` exit 0 over 36398 files. `docs-link-audit --anchors`: **0** broken links, measured in this worktree before and after rather than quoted.

No heading renamed, removed or re-levelled. `docs/docs.json` and the zh-CN glossary untouched.


vincentkoc:
Correct on all three counts, and I have removed the sample rather than rewritten it.

The real contract is at `src/plugin-sdk/provider-catalog-live-acquisition.internal.ts:65-67`:

```ts
export type LiveModelRowProjection<T extends ModelDefinitionConfig = ModelDefinitionConfig> = (
  rows: readonly unknown[],
  fallback: ModelProviderConfig,
) => ...
```

The added helper took `(row: { id?: string; display_name?: string }, fallback: (typeof STATIC_MODELS)[number])` — wrong on both parameters. It received one row where the projection is handed the whole array, and it typed `fallback` as a static model entry where the runtime passes a provider-level `ModelProviderConfig`. A reader following it would have written a projection that does not compile against the type it is assigned to. Your point about the extracted provider losing its `ProviderPlugin` contextual typing is the same defect one level up.

**Why I am reverting rather than fixing.** This pull request is a salvage: the agent that produced these edits stopped before writing its report, so nothing here carries the evidence its author would have supplied. That is stated at the top of the body. Two edits in this batch have now failed verification — two release citations on `docs/plugins/architecture.md`, which I dropped before opening, and this sample. Authoring a replacement would mean adding a fresh unverified sample to a change set that is already one unverified sample too many. The page is back to its state on the default branch.

If the projection example is worth having, it should be written against the type with its own verification, in a change that is about that.

The remaining 48 files stand on their own evidence: claims that contradict shipped code, samples consuming files nothing creates, a caveat printed after the example it qualifies, and one release citation established by tag containment (`7a456e362d0` / #104650, first containing tag `v2026.7.2-beta.1`, first suffix-free tag `v2026.8.1`).

Rebased onto current `main`; `pnpm check:docs` exit 0.

PR #144350, opened 2026-09-10
perf(agents): append ordered CLI reasoning deltas incrementally

### LINKED ISSUE #144349: Ordered CLI reasoning deltas repeatedly rebuild the full block order
Ordered CLI reasoning deltas currently rebuild the complete sorted block text on every append. When the delta extends the greatest content index, the existing lifecycle can append it directly.

Preserve canonical reconstruction for earlier-index changes and full snapshots, and reset the maximum together with its Map. Protect message-ID transitions, duplicate snapshots, event routing and the already-landed bounded-prefix capture tests. The intended claim is less repeated ordering work, not a universal memory reduction.

steipete:
Closes #144349

## What Problem This Solves

Ordered CLI thinking deltas repeatedly sort and reconstruct the accumulated reasoning blocks, even when the new text only extends the greatest content index.

## Why This Change Was Made

Track the current greatest index with the existing reasoning lifecycle and append those deltas directly. Earlier-index changes and complete snapshots retain the canonical sorted reconstruction. The maximum resets with its Map; late message-ID adoption, duplicate suppression and other event routing remain unchanged.

## User Impact

Reasoning output and callback behavior remain covered, while the common ordered stream avoids repeated ordering work. The change adds eight production lines and 109 test lines and preserves the landed bounded-prefix capture tests. No protocol, configuration, persistence or migration contract changes.

## Evidence

Both versions pass the same 129 tests across reasoning, JSONL, snapshots and the combined executor suite. The actual-parser benchmark matches callback counts, total characters and final text hashes. Ordering calls fall from 257 to one for a single block and from 2,056 to eight for eight blocks. The eight-block median was 16.58 versus 3.52 ms over three instrumented trials; this is a bounded observation, not a throughput guarantee.

Paired test invocations recorded 20.37 seconds / 2,048,655,360 bytes maximum RSS before and 18.65 seconds / 2,070,970,368 bytes afterward. Benchmark process RSS also rose slightly (155,058,176 to 155,582,464 bytes). Imports, caches, instrumentation and worker setup limit these observations; no retained-memory or whole-Gateway improvement is claimed. No provider or persisted-session activity was used.

The composed current-main source preserves the earlier CLI capture change. Its full selected changed gate and fresh independent P2 review pass, with no source fixes or gate reruns needed.

### DIFF
--- a/src/agents/cli-output-events.ts
+++ b/src/agents/cli-output-events.ts
@@ -394,6 +394,7 @@ type ThinkingTracker = {
   // is deduped against its own index; a single global concatenation misfires
   // once a message carries more than one thinking block (re-emits or reorders).
   streamedByIndex: Map<number, string>;
+  highestStreamedIndex: number;
   // Full thinking already emitted for the message in block order. The callback
   // contract exposes this as the running snapshot text for downstream coalescing,
   // so it stays a message-level concatenation, not a per-index value.
@@ -406,6 +407,7 @@ type ThinkingTracker = {
 export function createThinkingTracker(): ThinkingTracker {
   return {
     streamedByIndex: new Map(),
+    highestStreamedIndex: Number.NEGATIVE_INFINITY,
     emittedText: "",
     nextSyntheticBlockIndex: 0,
     progressTokens: 0,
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4013a231cfa6c620ee661f0724191eef99186219](https://github.com/openclaw/openclaw/pull/144350/commits/4013a231cfa6c620ee661f0724191eef99186219)
- Landed commit: [d350476f4734f7de3ec65df5a88bd2b9861a60ed](https://github.com/openclaw/openclaw/commit/d350476f4734f7de3ec65df5a88bd2b9861a60ed)

PR #144354, opened 2026-09-10
fix(tts): preserve selected summary models

steipete:
Related: #130706

## What Problem This Solves

Fixes an issue where TTS summaries could use the wrong model when their override or global default resolves through a provider alias. With `entry → middle → final`, selecting `entry` must use `middle`; preparing that selection must not advance it to `final`.

## Why This Change Was Made

TTS now shares the lightweight completion owner's metadata capture and model acquisition while retaining global defaults and aliases. This builds on the captured bare-default repair in #144429. Agent-specific model overrides remain scoped to agent completions, and the shipped injected preparation callback keeps its three-field argument and caller-owned returned model.

The shared parser also rejects malformed slash forms before exact-default-provider lookup. An invalid summary override such as `/` can therefore fall back to the global default, while configured slash aliases and literal qualified model IDs keep their existing behavior. Production delta: **+23 lines**; test delta: **+394 lines**.

## User Impact

Summaries preserve the selected override or global default through preparation. Bare defaults without configured aliases or model rows work with the captured provider policy. Existing visible-text cleanup, timeout behavior, profile-suffix stripping, and model-resource ownership are preserved.

## Evidence

Validation targets commit `68e88ebddf11460c4270411e28518ddad0f73fc2` on the actual #144429 merge, `cc748884201771eeac74dc6d8cf0c5b758bade76`.

- Matching current-base baseline: **16 TTS failures and 20 passing controls**, plus three malformed-input failures and seven passing parser controls. Candidate: **416 tests passed across 14 files**, including 36 TTS selection cases making 37 real loopback HTTP requests, captured-default owner tests, agent completions, worker inference, runtime-hook siblings, output, and resource cleanup.
- Full `pnpm check:changed --base cc748884201771eeac74dc6d8cf0c5b758bade76 --timed` passed without skipped checks. Fresh independent P2 review found no actionable findings. The committed source matches the reviewed and tested hashes.
- One normal full `pnpm build` passed on the committed source. Expanded proof through exported production SDK entry points passed in **41 fresh processes making 46 real loopback HTTP requests**, covering overrides, aliases, global bare defaults, invalid override fallback, exact IDs, injected callback reuse, and agent-completion siblings. Every process exited naturally. All 41 deliberate source-import probes were rejected, there were zero unexpected source imports, the listener closed with zero sockets, isolated state was removed, and all **12,245 build outputs** remained unchanged.

The initial supplemental fixture registered `openai` under a different plugin ID and did not supply the intended policy owner. Its failed runs are retained as fixture diagnostics; both baseline and candidate were rerun with the enabled synthetic `openai` replacement used for the results above.

The exact API-owner suppression case under a caller-owned generation is source-tested because the package exposes no generation acquisition/scope API for an equivalent exported-package probe. The public completion SDK retains bare-result ownership for process lifetime; compiled callback proof checks model reuse and natural process exit. Cold runtime-only `normalizeModelId` behavior remains inherited. HTTP proof uses an isolated synthetic loopback provider, not a hosted vendor API.

### DIFF
--- a/src/agents/model-selection-shared.ts
+++ b/src/agents/model-selection-shared.ts
@@ -489,7 +489,7 @@ function parseModelRefWithCompatAlias(
   } & ModelManifestNormalizationContext,
 ): ModelRef | null {
   const exactConfiguredProviderRef = resolveExactConfiguredProviderRef(params);
-  const exactDefaultProviderRef = hasSlashFormModelRef(params.raw)
+  const exactDefaultProviderRef = params.raw.includes("/")
     ? null
     : resolveExactConfiguredProviderRef({
         ...params,
--- a/src/agents/model-selection.test.ts
+++ b/src/agents/model-selection.test.ts
@@ -2074,6 +2074,56 @@ describe("model-selection", () => {
   });
 
   describe("resolveModelRefFromString", () => {
+    it.each<{
+      raw: string;
+      alias?: string;
+      expected: { provider: string; model: string } | null;
+    }>([
+      { raw: "/", expected: null },
+      { raw: "provider/", expected: null },
+      { raw: "//model", expected: null },
+      { raw: "literal", expected: { provider: "provider", model: "literal" } },
+      { raw: "provider/literal", expected: { provider: "provider", model: "literal" } },
+      { raw: "provider//literal", expected: { provider: "provider", model: "/literal" } },
+      {
+        raw: "provider/provider/literal",
+        expected: { provider: "provider", model: "provider/literal" },
+      },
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [68e88ebddf11460c4270411e28518ddad0f73fc2](https://github.com/openclaw/openclaw/pull/144354/commits/68e88ebddf11460c4270411e28518ddad0f73fc2)
- Landed commit: [bdb61168c086a66ad7a8718adeb3dc715dd80360](https://github.com/openclaw/openclaw/commit/bdb61168c086a66ad7a8718adeb3dc715dd80360)

PR #144360, opened 2026-09-10
fix(logs): clear retired records when the tailed file disappears

steipete:
## What Problem This Solves

Fixes an issue where the Control UI Logs page displayed and exported retired records alongside a replacement file after observing that its configured log file had disappeared. The reader returned cursor zero without telling clients to discard the previous tail.

## Why This Change Was Made

Treat an absent file as size zero in the shared reader and reuse its existing shrink/reset path. The unchanged UI then clears the retired records. This removes 11 production lines and preserves cursorless reads, rolling selection, byte limits, filtering, redaction and operational error handling.

## User Impact

After an observed disappearance, the Logs page and its export follow the recreated file. This covers a missing file observed between polls; it does not add inode tracking for replacements that occur entirely between reads.

## Evidence

The reader regression fails before the repair. The existing real-Gateway Logs lifecycle also reproduces four visible/exported records instead of the single replacement after the missing-file response. The same assertions pass after the repair, and the saved response changes from `reset: false` to `reset: true`.

All 74 distinct owner, sibling and browser cases and all 30 canonical changed checks pass. Independent review found no actionable findings. Proof uses a fresh production UI bundle, a task-owned loopback Gateway and synthetic logs under generic temporary paths; no operator Gateway or real logs are involved.

The before/after captures show the actual display after recreation. Their version footer and “Server updated” badge are fixed fixture metadata, not release-version proof. The export assertions independently verify the same correction.

Before: retired records remain alongside the replacement

After: only replacement log records remain

CI run 34525345506 initially timed out in an unchanged profiler repeated-help test. An isolated replay of the exact CI merge passed all 35 profiler cases on Linux x64 Node 24.19.0; it used x64 emulation and ran the file separately, so the original timeout remains unexplained. The original failure is preserved, and the single failed-job retry on the same PR head passed. No deadline, assertion or profiler code was changed.

### DIFF
--- a/docs/gateway/logging.md
+++ b/docs/gateway/logging.md
@@ -45,6 +45,8 @@ The Control UI Logs tab tails this file via the gateway (`logs.tail`). The CLI d
 openclaw logs --follow
 ```
 
+If a tail read observes that the active file has disappeared, the Control UI clears its previous records and follows the recreated file. Missing files still return an empty tail; filesystem read errors remain visible.
+
 ### Verbose vs. log levels
 
 - **File logs** are controlled exclusively by `logging.level`.
--- a/src/logging/log-tail.test.ts
+++ b/src/logging/log-tail.test.ts
@@ -187,6 +187,38 @@ describe("readConfiguredLogTail", () => {
     expect(fileShrink.skippedBytes).toBeUndefined();
   });
 
+  it.each(["missing", "empty"])(
+    "resets a positive cursor when its file becomes %s",
+    async (state) => {
+      const { readConfiguredLogTail } = await import("./log-tail.js");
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [569f42008da19e43bd59e67b73a644c4262c734d](https://github.com/openclaw/openclaw/pull/144360/commits/569f42008da19e43bd59e67b73a644c4262c734d)
- Landed commit: [d2cf02d0bf2b158e4c71ae8bbd8a9cf1aa3decc9](https://github.com/openclaw/openclaw/commit/d2cf02d0bf2b158e4c71ae8bbd8a9cf1aa3decc9)

PR #144362, opened 2026-09-10
perf(plugins): reuse warm source-cache records before path normalization

### LINKED ISSUE #144361: Warm plugin source-cache hits repeat filename normalization
Repeated warm shared plugin-surface loads normalize a filename before checking an already available exact source-cache record. Reuse the existing alias-first raw lookup, retaining alias precedence, explicit cache generations and normalization for relative/file-URL misses.

Protect the actual checked-open alias writer and shared export identity. This requires no new cache, retention policy or configuration/store change; allocation claims stay scoped to the measured warm loader workload.

steipete:
Closes #144361

## What Problem This Solves

Warm plugin source-cache reads normalize already-cached filenames before looking up the existing record. Repeated shared public-surface loads therefore allocate path strings even when the cache has an exact hit.

## Why This Change Was Made

Check the existing source-alias mapping and exact source key first, mirroring the adjacent cache-root lookup. Aliases keep precedence over stale lexical records, the explicitly supplied cache remains authoritative, and relative/file-URL misses keep the existing normalization path. No new cache or retention policy is introduced.

## User Impact

Plugin source and export identity remain unchanged. Four production lines avoid normalization work on warm hits; the existing cache generation, alias writer and fallback semantics are preserved. No public configuration, schema, persistence or migration contract changes.

## Evidence

Both versions pass 58 cache-owner, module-loader and public-surface tests. Six fresh interleaved processes exercise 100,000 warm loads through the actual shared public-artifact loader, checking the same export identity on every call and identical checksums. Median sampled allocation falls from 221,040,616 to 60,083,016 bytes (72.8%) across the complete isolated sampling interval, including helper work and collected objects.

The full selected changed gate, normal build, isolated keyless canonical Discord entrypoint profile and independent P2 review pass. The warm result does not establish whole-Gateway memory, retained RSS or latency savings; the entrypoint profile is diagnostic only. The unchanged lockfile is a proof input, not part of this patch.

### DIFF
--- a/src/plugins/plugin-cache.test.ts
+++ b/src/plugins/plugin-cache.test.ts
@@ -1,6 +1,7 @@
 import crypto from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
+import { pathToFileURL } from "node:url";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
 import { detectBundleManifestFormat, loadBundleManifest } from "./bundle-manifest.js";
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ad659b71beb44b8f6a6dab2cf56cd21784dec75d](https://github.com/openclaw/openclaw/pull/144362/commits/ad659b71beb44b8f6a6dab2cf56cd21784dec75d)
- Landed commit: [dc0d0b09be873ebdced3054d30bc29b5aa9324fa](https://github.com/openclaw/openclaw/commit/dc0d0b09be873ebdced3054d30bc29b5aa9324fa)

PR #144365, opened 2026-09-10
perf(imessage): bound content work while coalescing messages

### LINKED ISSUE #144364: iMessage coalescing collects content beyond its output bounds
iMessage coalescing builds full intermediate text, attachment and timestamp collections before enforcing existing output caps. Bound the content work during the input walk while retaining every row's GUIDs, recovery maxima, latest lexical timestamp and first qualifying reply-context source.

Preserve first-plus-last content selection, case-insensitive text deduplication, surrogate-safe truncation, attachment order and all metadata from dropped content. This is a private coalescer cleanup with no strict physical-memory or live-service claim.

steipete:
Closes #144364

## What Problem This Solves

iMessage coalescing first builds complete text, attachment and timestamp collections and trims them afterward. Content beyond the established limits therefore does unnecessary normalization and collection work.

## Why This Change Was Made

Bound content while walking the input once. Retain the first permitted entries plus the last, stop collecting attachments at the existing cap, and keep one text lookahead code unit so the final surrogate-safe truncation and marker remain unchanged. Every input row still contributes GUID deduplication, recovery maxima, lexical timestamp selection and reply-context choice.

## User Impact

Merged text, caps, attachment order, reply fields and dropped-content metadata keep their contracts. The implementation removes seven production lines and adds 71 net test lines without changing public APIs, configuration or persistence.

## Evidence

The existing monitor/debounce suites pass before (279 tests) and after (285 tests), including six additional boundary cases. The full selected changed gate and source-bound P2 review pass; runtime imports, hot/lazy entrypoints, SDK exports and dependencies are unchanged, so a normal build was not required.

Before/after correctness commands recorded 17.71 seconds / 2,605,514,752 bytes maximum RSS and 16.75 seconds / 2,113,372,160 bytes. They are single sequential runs with different test counts, not isolated performance measurements. No allocation or RSS gain is claimed, and bounded string length does not guarantee a strict V8 substring backing-storage bound. Proof uses isolated monitor/debounce fixtures, not a live iMessage service.

### DIFF
--- a/extensions/imessage/src/monitor/coalesce.test-support.ts
+++ b/extensions/imessage/src/monitor/coalesce.test-support.ts
@@ -66,62 +66,133 @@ describe("combineIMessagePayloads", () => {
     expect(merged.attachments).toEqual([{ original_path: "/tmp/a.jpg", mime_type: "image/jpeg" }]);
   });
 
-  it("dedupes identical text appearing in both rows (URL in text and balloon)", () => {
-    const a = makePayload({ text: "https://example.com", guid: "row-1" });
-    const b = makePayload({ text: "https://example.com", guid: "row-2" });
-    const merged = combineIMessagePayloads([a, b]);
+  it.each([
+    {
+      label: "identical URL text and balloon",
+      texts: ["https://example.com", "https://example.com"],
+      expected: "https://example.com",
+    },
+    {
+      label: "trimmed case variants and distinct suffixes",
+      texts: ["  MIXED ", "mixed", "Mixed suffix", "mIxEd"],
+      expected: "MIXED Mixed suffix",
+    },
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5ac027f2ca9e856ec02ddeadac762069e9e69844](https://github.com/openclaw/openclaw/pull/144365/commits/5ac027f2ca9e856ec02ddeadac762069e9e69844)
- Landed commit: [dc5c29ed7d6f9cf1f5ea9729881718f3f4096ecd](https://github.com/openclaw/openclaw/commit/dc5c29ed7d6f9cf1f5ea9729881718f3f4096ecd)

PR #144381, opened 2026-09-10
fix(macos): extract app bundle identifiers without full plist JSON

### LINKED ISSUE #144380: Valid Mac bundle identifiers are lost when unrelated plist data cannot be serialized
Mac app discovery serializes all Info.plist metadata to JSON to retrieve one bundle identifier. Unrelated Date values can reject JSON conversion, and unrelated large values can exceed the bounded subprocess output even when the identifier is short.

Use native string extraction for CFBundleIdentifier, preserving strict string/trim behavior, error omission, existing timeout/capture/concurrency limits and scanner ordering/filter/symlink behavior. Prove it with synthetic native plists; do not inventory real applications.

steipete:
Closes #144380

## What Problem This Solves

Mac app discovery converts an entire Info.plist to JSON just to read CFBundleIdentifier. An unrelated Date value can make that conversion fail, and a large unrelated value can exceed the subprocess output limit, leaving a valid identifier missing.

## Why This Change Was Made

Ask native plutil to extract the identifier as a string, then retain the existing trim/empty handling. Remove whole-document JSON parsing and the test-only reader injection; tests use isolated native plist fixtures instead.

## User Impact

Valid identifiers are recovered despite unrelated Date or large metadata values. App filtering, ordering, symlinks, error omission, the 2-second timeout, 1 MiB capture limit and eight-reader concurrency remain unchanged. Device-app sharing and onboarding callers retain their contracts; no configuration or storage migration is introduced.

## Evidence

The actual owner matches 18 synthetic native outcomes before and after; only the intended Date and large-unrelated-value cases improve. Owner/device-app suites pass (five tests before, six afterward), and the new native regression fails against the original reader at the missing Date identifier. An evidence guard initially held restoration because JSON abbreviated the assertion; the full text and process cleanup were checked, then exact candidate bytes restored without rerunning the test.

For one synthetic 512 KiB unrelated value, subprocess stdout shrinks from 524,346 to 23 bytes with the same identifier. This measures representation size, not allocation, RSS, latency or timeout cancellation. No real application inventory was read. The full selected changed gate and independent P2 review pass.

### DIFF
--- a/src/infra/installed-apps.test.ts
+++ b/src/infra/installed-apps.test.ts
@@ -1,10 +1,11 @@
+import { execFileSync } from "node:child_process";
 import fs from "node:fs/promises";
-import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
+import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
 import { scanInstalledApps } from "./installed-apps.js";
 
-const tempRoots: string[] = [];
+const tempDirs = useAutoCleanupTempDirTracker(afterEach);
 
 async function makeFixtureRoot(): Promise<{
   root: string;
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [454242f7ea35bb542eb0225b78da25f8ba700971](https://github.com/openclaw/openclaw/pull/144381/commits/454242f7ea35bb542eb0225b78da25f8ba700971)
- Landed commit: [83129f42703ff36ee68708f8fc0fa88eb7562a25](https://github.com/openclaw/openclaw/commit/83129f42703ff36ee68708f8fc0fa88eb7562a25)

PR #144385, opened 2026-09-10
perf(ui): keep transcript history cached during worker setup

### LINKED ISSUE #144384: Unchanged worker setup invalidates cached transcript history
Repeated unchanged worker-setup projections rebuild cached transcript history because a freshly derived setup-message array is treated as a structural dependency. Carry the stable placement-derived boolean instead, leaving run-ID/state eligibility and workspace-sync matching with their current owner.

Preserve queued messages, setup-to-active invalidation, interruption/cancellation behavior, plugin-icon dependencies and work-group semantics. Prove the actual projector/cache path rather than a directly invoked builder helper.

steipete:
Closes #144384

## What Problem This Solves

An unchanged worker-setup projection creates a fresh setup-message array that invalidates the structural transcript cache. Repeated rendering therefore rebuilds history even when the relevant state has not changed.

## Why This Change Was Made

Carry the placement-derived setup boolean through the actual projector and cache. The queued-input owner retains run-ID/state eligibility, notice precedence and separate workspace-sync matching. Transitioning to active placement still invalidates the view, clears the notice and preserves the queued message.

## User Impact

Queued content and setup/status behavior remain unchanged while redundant historical rebuilding is avoided. Current plugin-icon propagation/dependencies and work-group behavior are preserved. The change removes seven production lines and adds 86 net test lines without a new public API, configuration or persistence contract.

## Evidence

The actual projector regression fails on the composed current-main baseline because unchanged setup calls the history builder twice instead of once. The candidate passes that regression and all 324 owner tests, including active transition, icon, grouping and custody controls. The full selected changed gate and fresh independent P2 review pass.

Both synthetic mock-Gateway browser arms show one queued message and the setup notice, then remove only the notice on active placement, with no page errors. The inspected stage-matched screenshots are byte-identical; the attached pair shows the same setup stage. No live worker provisioning, latency or heap-byte improvement is claimed.

### DIFF
--- a/ui/src/pages/chat/chat-pending-inputs.test.ts
+++ b/ui/src/pages/chat/chat-pending-inputs.test.ts
@@ -123,6 +123,32 @@ describe("server-owned pending input display", () => {
     );
   });
 
+  it.each([
+    { state: "queued", runId: undefined, notice: undefined },
+    {
+      state: "interrupted",
+      runId: "run-queued",
+      notice:
+        "Interrupted before the agent started it. It will not run automatically; copy it and send again.",
+    },
+    {
+      state: "cancelled",
+      runId: "run-queued",
+      notice:
+        "Cancelled before the agent started it. It will not run automatically; copy it and send again.",
+    },
+  ] as const)(
+    "keeps $state custody out of the worker-setup notice without eligible execution",
+    ({ state, runId, notice }) => {
+      const items = buildPendingInputItems([{ ...input, state, runId }], undefined, [], [], true);
+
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9d511185990fa9c5913fe5b0f9b2bf94d2664ff3](https://github.com/openclaw/openclaw/pull/144385/commits/9d511185990fa9c5913fe5b0f9b2bf94d2664ff3)
- Landed commit: [662498c9eb7da6b28b3c1051e8b66ab804cddb79](https://github.com/openclaw/openclaw/commit/662498c9eb7da6b28b3c1051e8b66ab804cddb79)

PR #144393, opened 2026-09-10
fix(docs): restore the complete Show widget reference

vyctorbrzezowski:
## What Problem This Solves

Fixes an issue where readers of [Show widget](https://docs.openclaw.ai/tools/show-widget#use-the-tool) lose the rest of the reference halfway through “Use the tool.” An inline `<script>` example is treated as raw HTML, hiding the remaining content in the browser.

## Why This Change Was Made

Keep backtick code spans out of the initial raw-HTML capture in the shared documentation parser. Respect Markdown block boundaries so an unmatched backtick cannot swallow an actual script block. The publishing repository receives the parser through docs sync.

## User Impact

The widget parameters, examples, related links, and source link remain readable through the end of the page. Authored raw HTML remains intact.

## Evidence

- Five regression cases fail on the original parser; all nine cases pass after the fix, including genuine raw-HTML siblings, unmatched backticks, and CRLF paragraph boundaries.
- Compared rendered HTML, anchors, and links for all 1,289 English Markdown sources. Only `docs/tools/show-widget.md` changes: its raw script tag disappears and the missing parameter anchors and link return.
- Focused syntax, formatting, lint, and `git diff --check` passed.
- The real publisher preview restores the complete reference and its Edit source link.

The installed Vitest is older than the current default configuration, so the test file ran with an external minimal configuration. The default runner fails before collection (`defineCacheKeyGenerator is not a function`); that gate is not claimed as passing.

Before and after below use the real `openclaw/docs` publisher in a local four-page preview. This is focused rendering proof, not a full site build or a deployed fix.

Before:

Before

After:

After

### DIFF
--- a/scripts/lib/docs-markdown.mjs
+++ b/scripts/lib/docs-markdown.mjs
@@ -241,8 +241,13 @@ function prepareDocument(input, { sourceFile, root, seen = new Set() }, firstLin
   };
   const jsxComments = new Set();
   let text = new DocsSource(input, firstLine).replace(
-    /<!--[^]*?-->|\{\/\*[^]*?\*\/\}|<(pre|code|script|style|textarea)\b[^>]*>[^]*?<\/\1\s*>/g,
+    /(`+)(?:(?!\n(?:[ \t]*\r?\n|[ \t]*<(?:pre|script|style|textarea)\b))[^])*?\1|<!--[^]*?-->|\{\/\*[^]*?\*\/\}|<(pre|code|script|style|textarea)\b[^>]*>[^]*?<\/\2\s*>/g,
     (value) => {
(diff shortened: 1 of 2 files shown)


vyctorbrzezowski:
Merged via squash.

- Prepared head SHA: [dd5e15c30c6a803f555b50e1039d7a92baf1a49a](https://github.com/openclaw/openclaw/pull/144393/commits/dd5e15c30c6a803f555b50e1039d7a92baf1a49a)
- Landed commit: [0203193f9d4947d7a548d3126f7755c508c93c56](https://github.com/openclaw/openclaw/commit/0203193f9d4947d7a548d3126f7755c508c93c56)

PR #144396, opened 2026-09-10
fix(codex): permit session deletion of predecessor retired tombstones (#144179)

### LINKED ISSUE #144179: sessions delete permanently vetoed by a retired Codex binding generation on non-Codex sessions
**Summary**

`sessions delete` is permanently vetoed for any session whose `app-server-thread-bindings` row still carries a **retired generation with an older `sessionId`** — and on sessions that never run the Codex harness, nothing ever reclaims that row, so the veto is forever.

```
Error: Codex binding generation changed before session deletion
```

This is the same row family as #135778 (closed as fixed in 2026.8.2). That fix covers the *turn* path: `reclaimCurrentCodexSessionGeneration()` rewrites the stale `sessionId` — but it only runs inside a Codex run. A session on a different harness (here: `claude-cli`) never reaches it, so the tombstone keeps the old generation and `withSessionDeletion()` throws on every attempt.

Observed on **2026.8.2 (0965053)**.

**Repro**

1. Take a session key that once had a Codex app-server binding; let it end up as `{"version":1,"state":"cleared","sessionId":"<OLD>","retired":true}` in `plugin_state_entries` (plugin `codex`, namespace `app-server-thread-bindings`, key `session-key:<agent>:<sha256url(sessionKey)>`).
2. Rotate the session outside Codex, so the session store holds a new `sessionId` for the same key.
3. Run the session on a non-Codex harness (claude-cli) so no Codex run ever happens.
4. `openclaw sessions delete "<sessionKey>" --agent <agent> --yes` → fails as above. Deterministic, identical on every retry.

**Evidence**

Two watchdog runs, 10 minutes apart, identical failure on the same key:

```
15:51   sessions delete rc=1: Error: Codex binding generation changed before session deletion
16:01   sessions delete rc=1: Error: Codex binding generation changed before session deletion
```

Stored row (read-only query, ids shortened):

```
plugin_id=codex namespace=app-server-thread-bindings
entry_key=session-key:main:xjnY1ejR…
value={"version":1,"state":"cleared","sessionId":"cde23540-…","retired":true}
live session store for the same key: sessionId=de02f135-…
```

Not a single stuck session: hashing all live session keys against the binding table found **14 sessions** with a mismatched stored generation on this install, 10 of them chat lanes that will each hit the same wall the first time deletion is attempted.

The relevant guard is `ownsStoredSessionGeneration(identity, current)` in `withSessionDeletion()` — it compares the stored generation to the live one and throws, with no reclaim attempt on the deletion path (unlike `prepareSessionGenerationReclaim` / `mutate({kind:"reclaim-generation"})` on the run path).

**Expected**

Deleting a session should not be blockable by a `cleared`/`retired` tombstone of a *previous* generation of that same session key. A retired row for an older generation carries no live thread to protect: it should be reclaimed (as the run path already does), superseded, or simply deleted along with the session.

**Workaround**

`sessions.reset` does not go through the Codex veto and empties the lane:

```
openclaw gateway call sessions.reset \
  --params '{"key":"<sessionKey>","agentId":"<agent>","reason":"reset"}' --json
```

Verify on `entry.totalTokens == 0` rather than on the exit code. The transcript is not archived on this path, so it is a fallback, not a substitute for delete. Deleting the row straight out of SQLite works too but is not something an operator should have to do.

**Environment**

OpenClaw 2026.8.2 (0965053), Linux 7.0.0-29-generic, Node v26.8.1, npm install, 2 configured agents. All queries above read-only (`mode=ro`); no secrets, ids shortened.

0xalydev:
## What Problem This Solves

Fixes #144179

A session key whose `app-server-thread-bindings` row in `plugin_state_entries` carries a cleared, retired tombstone (`{"version":1,"state":"cleared","sessionId":"<OLD>","retired":true}`) permanently vetoes `openclaw sessions delete` when the session has rotated or moved to a non-Codex runtime:

```text
Error: Codex binding generation changed before session deletion
```

On non-Codex sessions, nothing ever reaches the turn-path reclaim (`reclaimCurrentCodexSessionGeneration`), leaving the stale tombstone in place indefinitely and permanently blocking deletion of the session.

In `createCodexAppServerBindingStore.withSessionDeletion`:
- Line 1171 already acquires the binding lease with `{ allowRetired: true }`.
- However, line 1112 previously rejected any `sessionId` mismatch unconditionally via `ownsStoredSessionGeneration(identity, stored)`.
- When the stored row is a cleared, retired tombstone (`stored.state === "cleared" && stored.retired === true`), it contains no live app-server thread or active execution authority. It must not block deletion of a successor generation for that session key.
- Active bindings (`stored.state === "active"`) continue to strictly enforce `ownsStoredSessionGeneration` so active threads belonging to other generations remain protected.

### Unit & Regression Suite Execution
Ran focused regression in `extensions/codex/src/app-server/session-binding.test.ts`:
```text
✓ extension-codex extensions/codex/src/app-server/session-binding.test.ts > Codex app-server binding store > removes retired fences from predecessor generations during session deletion (#144179) 6ms
```

### Static Analysis
```text
npx oxlint extensions/codex/src/app-server/session-binding.ts extensions/codex/src/app-server/session-binding.test.ts
Found 0 warnings and 0 errors.
Finished in 1.0s on 2 files with 230 rules using 12 threads.
```

### Whitespace & Git Check
```text
git diff --check
(clean, 0 warnings)
```

### DIFF
--- a/extensions/codex/src/app-server/session-binding.test.ts
+++ b/extensions/codex/src/app-server/session-binding.test.ts
@@ -178,6 +178,86 @@ describe("Codex app-server binding store", () => {
     }
   });
 
+  it("removes retired fences from predecessor generations during session deletion (#144179)", async () => {
+    const { state, values } = createStateStore();
+    const store = createCodexAppServerBindingStore(state);
+    const predecessor = {
+      kind: "session" as const,
+      agentId: "main",
+      sessionId: "predecessor-uuid-1",
+      sessionKey: "agent:main:cron:rotated",
+    };
+    await store.mutate(predecessor, {
+      kind: "set",
+      binding: { threadId: "thread-1", cwd: "/repo" },
+    });
+    await store.retireSessionGeneration(predecessor);
+    expect(values.get(bindingStoreKey(predecessor))).toMatchObject({
+      state: "cleared",
+      retired: true,
+      sessionId: "predecessor-uuid-1",
+    });
+
+    const successor = {
+      kind: "session" as const,
+      agentId: "main",
+      sessionId: "successor-uuid-2",
+      sessionKey: "agent:main:cron:rotated",
+    };
+
+    // Deletion under successor identity successfully removes the retired tombstone of predecessor
+    await store.withSessionDeletion(
+      successor,
+      () => {},
+      async (binding, mutation) => {
+        expect(binding).toBeUndefined();
+        mutation.commit();
+      },
+    );
+    expect(values.size).toBe(0);
+
+    // Rollback properly restores the retired tombstone
+    const tombstone = {
+      version: 1 as const,
+      state: "cleared" as const,
+      retired: true as const,
+      sessionId: "predecessor-uuid-1",
+    };
+    state.register(bindingStoreKey(successor), tombstone);
+    await store.withSessionDeletion(
+      successor,
+      () => {},
+      async (_binding, mutation) => {
+        mutation.commit();
+        expect(values.size).toBe(0);
+        mutation.rollback();
+      },
+    );
+    expect(values.get(bindingStoreKey(successor))).toMatchObject({
+      state: "cleared",
+      retired: true,
+      sessionId: "predecessor-uuid-1",
+    });
+
+    // An active successor belonging to a different session is still protected and rejected
+    const activeDifferentGeneration = {
+      version: 1 as const,
+      state: "active" as const,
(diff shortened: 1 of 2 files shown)


0xalydev:
@clawsweeper review

0xalydev:
@clawsweeper re-review

PR #134959, opened 2026-09-01
fix(memory): honor bounded embedding provider cooldowns

shojikumaru:
Closes #108893

## What Problem This Solves

Memory indexing gives up before an embedding provider's cooldown expires. Both Google and generic OpenAI-compatible embedding paths exhaust three short attempts, leaving the index incomplete despite a valid `Retry-After` response.

## Change

Carry validated cooldown and quota metadata through the existing shared HTTP error owner, including the generic embedding request path. Keep Google `RetryInfo` decoding at the Google boundary, with a bounded, redacted response clone that closes when the caller cancels.

Keep one memory retry loop. Index rate limits get at most five attempts and 60 seconds per wait; other transient errors retain three attempts. Interactive search keeps three attempts, at most eight seconds of total retry sleep, and the existing tool deadline. Structured permanent quota errors without a retry hint stop without another memory retry.

The generic error path retains its 8 KiB input bound and existing batch retry behavior. It now uses the shared diagnostic format and 500-character error-body preview. No configuration surface, persistent schema, vector metadata, authentication contract, or endpoint changes.

## Evidence

- Built public `memory index --force --verbose`: pinned main fails before an enforced ten-second server cooldown, with zero of three files indexed. This candidate waits until the cooldown ends and indexes all three files through both Google and OpenAI-compatible providers.
- Google public controls honor a six-second `RetryInfo` over a two-second header. An oversized body with a trailing hint falls back to the valid two-second header. Both index every file.
- Public `agent exec` calls the real `memory_search` tool and delivers its actual result to the continuation. A persistent long-hint query makes three attempts, sleeps 8,000 ms total, and returns unavailable after 8,117 ms. Permanent quota index/query controls do not retry. Caller SIGINT interrupts a pending wait and exits through normal cleanup in 66 ms.
- 364 focused tests pass. Header propagation fails on main and passes with the fix; real response-clone cancellation tests fail before the corrective change and pass afterward. Syntax lint, formatting, safety/size checks, docs links, and the complete native build with CLI/SDK/doctor-closure checks pass. Independent behavior and source judgments agree within the recorded proof boundaries.

## Limits

[...]

Issue #132708, opened 2026-08-29
[memory] openai-compatible embeddings need configurable throttle + honor Retry-After + enable Batch API


gaodaabao:
[memory] openai-compatible embeddings need configurable throttle + honor Retry-After + enable Batch API

## Problem

When using an `openai-compatible` provider (千帆 bge-large-zh, 阿里百炼, 火山方舟, etc.) for `memory.search`, OpenClaw fires embedding requests in a tight loop with no throttle, no concurrency limit, and no `Retry-After` honoring. The provider returns 429 (TPM rate limit), the indexer retries immediately, and the daily quota burns out within seconds — leaving the index dirty and `vector search` permanently disabled.

Real log:
```
openai-compatible embeddings failed: HTTP 429:
{"error":{"code":"tpm_rate_limit_exceeded",
"message":"Rate limit reached for TPM",
"type":"rate_limit_exceeded"},
"id":"as-1k2xjjsf5e"}
```

## Environment

- OpenClaw: v2026.9.1-beta.1 (1d96e5a)
- `memory.search.provider`: `openai-compatible`
- `memory.search.remote.baseUrl`: `https://qianfan.baidubce.com/v2/`
- `memory.search.model`: `bge-large-zh`
- `memory.search.sources`: `["memory", "sessions"]` (~500 files reindexed)
- `memory.search.remote.batch.enabled`: `true` (silently ignored, see below)

## Repro

1. Configure `memory.search.provider = "openai-compatible"` against a TPM-limited endpoint
2. Trigger any full reindex — e.g. by changing `model`, changing `sources`, or upgrading the binary
3. Observe 429 storm during reindex; `Indexing session files…` stalls at e.g. `552/556`
4. `vector search` stays `paused`; memory.search effectively broken until config rolled back

### 1. No throttle / concurrency control on non-batch embedding path
- `extensions/memory-core/manager-runtime.js` has no configurable concurrency for embedding requests
- When `remote.batch.enabled = true` but no provider implements batch (see #2), code falls back to per-file synchronous embedding → instant TPM cliff

### 2. `remote.batch.enabled` is dead code for openai-compatible
- The only batch implementation is Gemini-specific:
  `dist/embedding-batch-BJScf-Sm.js` imports `parseGeminiAuth`, `sanitizeGeminiEmbedding`, `getGeminiVersionedRouteBase`, etc.
- The gate in `manager-runtime.js:1921` is:
  ```js
  Boolean(
    this.batch.enabled
[...]

Issue #91354, opened 2026-06-08
memory: Gemini embeddings still use inline batch requests after remote.batch.enabled=false, causing 429 on low quota


Louisone1:
## Summary

After upgrading to OpenClaw `2026.6.1`, memory reindexing with `gemini-embedding-2-preview` started failing with `429 RESOURCE_EXHAUSTED` even when `remote.batch.enabled=false` and `remote.nonBatchConcurrency=1` are configured.

The same Google/Gemini key and model can successfully generate embeddings when requests are sent one at a time with a delay (for example, 1 request every 10 seconds). This suggests the problem is not that the key/model is unusable, but that OpenClaw's memory indexing path still sends grouped/batch embedding requests too aggressively for low-quota Gemini embedding projects.

## Environment

- OpenClaw: `2026.6.1`
- Memory embedding provider: `gemini`
- Model: `gemini-embedding-2-preview`
- Output dimensionality: `3072`
- Production fallback used during mitigation: local embeddings

## Relevant config

```json
{
  "memorySearch": {
    "enabled": true,
    "provider": "gemini",
    "fallback": "local",
    "model": "gemini-embedding-2-preview",
    "outputDimensionality": 3072,
    "remote": {
      "nonBatchConcurrency": 1,
      "batch": {
        "enabled": false,
        "concurrency": 1,
        "wait": true,
        "pollIntervalMs": 10000,
        "timeoutMinutes": 180
      }
    },
    "sync": {
      "embeddingBatchTimeoutSeconds": 300
    }
  }
}
```

## Reproduction

1. Configure Gemini memory search as above.
2. Run a force memory reindex for one agent.
3. Observe retries followed by fallback/local or failure with Gemini quota errors.

Observed error:

```text
gemini embeddings failed (429): Quota exceeded for aiplatform.googleapis.com/online_prediction_requests_per_base_model with base model: gemini-embedding-2. Please submit a quota increase request. [code=RESOURCE_EXHAUSTED]
```

## What I expected

With:

```json
"remote": {
  "batch": { "enabled": false },
  "nonBatchConcurrency": 1
}
```

I expected indexing to use a very conservative request pattern, ideally single-request `embedContent` calls or at least a configurable small batch size and delay.

## What seems to happen

From inspecting the installed `2026.6.1` bundle:

- Gemini remote batch support uses `asyncBatchEmbedContent` with a high max group size (`GEMINI_BATCH_MAX_REQUESTS = 50000`).
[...]

PR #144404, opened 2026-09-10
fix(memory): reuse completed embeddings after failed rebuilds

jalehman:
Related: #91354, #132708

## What Problem This Solves

Fixes an issue where users retrying a failed memory-index rebuild would send successful embedding work to the provider again, even after a manager restart. On a large, quota-limited rebuild, discarding those completed results can repeatedly consume the same allowance without finishing the index.

The generic OpenAI-compatible embedding adapter also discarded structured HTTP status, quota, and retry metadata. Its errors could not fully participate in the bounded retry policy already landed in #134959.

## Why This Change Was Made

Keep the published agent database as the canonical embedding-cache owner, separate from atomic shadow-index publication. Persist validated successful inline batches, successful split leaves, and completed per-file/source-wide runtime batches immediately. A later indexing failure can discard the incomplete shadow without discarding reusable embeddings. Remove the full-cache shadow seeding and cache replacement at publication.

Cache writes validate source snapshots and respect provider generations, database lifetime/revisions, workspace locking, and session tombstones. Reserve cache capacity before insertion, enforce the canonical cache cap after successful full rebuilds (including all-cache-hit rebuilds), and reject malformed vectors before retaining them. If separate requests disagree about vector dimensions, invalidate only the ambiguous provider identities so a subsequent retry can recover; retain other providers' caches and the published index. Reuse remains bounded by configured cache retention and compatible provider identity/content hashes.

A nonempty Forget selection now clears the selected agent's entire embedding cache, including unpublished rebuild results that cannot be selected through published chunk provenance. This is the accepted no-schema tradeoff: unrelated published index entries remain searchable, but future indexing may need to regenerate their cached embeddings. Dry runs report the whole-cache count without mutation; empty selections do neither.

The remaining generic adapter fix uses the existing shared HTTP-error owner and redacts reflected request credentials. This branch was rebased after #134959; it does **not** replace that PR's cooldown profiles, Google RetryInfo parsing, permanent-quota classification, or interactive search budget. The only retry-policy addition here is a completed-batch callback used by the cache owner.

## User Impact

- With embedding caching enabled, retries can reuse compatible completed work instead of requesting it again after a failed rebuild or manager restart.
- The previous published index stays available until the replacement is successfully published.
- Rebuild preparation no longer copies the entire embedding cache into a temporary database.
- Generic OpenAI-compatible failures retain retry/status metadata and redact reflected request credentials.
- Forget removes cached embeddings from unfinished rebuilds as well as the published-index-derived data covered by the existing purge.

This is a partial repair for the linked reports, not a complete quota-management solution. It does not add shared cross-agent quota admission, persist or resume unfinished remote Batch jobs, add Batch support to another provider, explain every rebuild phase, or suppress legitimate upgrade-triggered rebuilds. No new configuration, database schema, or dependency is introduced.

## Evidence

- Follow-up final validation: the complete changed-file gate passed (exit 0), including all TypeScript lanes, lint, dead-export scans, and storage/plugin guards. The exact staged follow-up passed P0–P2 autoreview with no actionable findings.
- Review follow-up: a new real-SQLite regression reproduced the cache-cap defect (three retained rows with `maxEntries: 1` after a successful cache-only rebuild). Pruning after successful publication fixes it with zero provider requests; the failed-rebuild preservation control remains green.
- Follow-up regression suite: **71 tests passed across five files**, covering recovery, cache capacity, source races, Forget, and chunk publication. The three CI publication failures were old assertions that treated the embedding cache as part of the atomic index. The updated tests still require every published index table to roll back, and separately require completed embeddings to survive and be reused without additional provider calls.
- Existing-database compatibility: recovery fixtures create a published index/cache, fail a subsequent rebuild, close and reopen the manager, and verify reuse plus index preservation. The legacy-schema test still requires doctor rather than silently accepting an incompatible schema. This change introduces no schema migration; whole-cache Forget remains the documented, accepted behavior tradeoff.
- Rebased focused validation: **177 tests passed across ten files** (36 generic adapter tests across two focused files; 141 memory tests across eight files), including recovery/restart, late failure, per-file/source-wide batch retention, malformed vectors, source races, Forget, cache capacity, publication, and the existing cooldown/query-budget regressions.
- A cross-batch dimension regression was observed failing before the fix: after restart, otherwise-correct provider responses were still rejected because the earlier wrong-dimension batch remained cached. It now passes and also verifies that unrelated provider caches survive invalidation.
- Recovery regressions demonstrate the original failure order: successful embeddings, later shadow failure, manager restart, then retry. The repaired retry sends no new requests for retained inputs while the old published index survives the failed attempt.
- Generic adapter proof uses a real loopback HTTP server returning a synthetic 429/Retry-After response. Memory tests use isolated SQLite fixtures and deterministic fake providers, never a user's live memory database.
- Initial P0–P2 autoreview: passed, no actionable findings. Both accepted findings from earlier candidates were addressed: the superseded unbounded retry implementation was removed during rebase, and cross-batch dimension poisoning was reproduced and repaired.
- Initial changed-file gate passed for all 19 changed paths, including four TypeScript graph lanes, formatting, core and extension lint, ratchets, plugin/database boundaries, import-cycle checks, dead-export scans, and the selected contract tests.
- No paid-provider reproduction, full repository test suite, or complete application build was run. These fixtures do not establish external provider billing behavior or the cause of the reporter's apparent Gemini Batch stall.

<details>
<summary>Changes walkthrough</summary>

| File | Change |
| --- | --- |
| `extensions/memory-core/src/memory/manager-embedding-ops.ts` | Persist successful results through the canonical cache owner with lifetime/source/revision fencing. |
| `extensions/memory-core/src/memory/manager-embedding-cache.ts` | Validate cached vectors, deduplicate writes, and reserve bounded capacity. |
| `extensions/memory-core/src/memory/manager-embedding-policy.ts` | Report successful split leaves before later work can fail. |
| `extensions/memory-core/src/memory/manager-sync-ops.ts` | Retain canonical cache ownership and enforce retention after successful publication. |
| `extensions/memory-core/src/memory/manager-sync-base.ts` | Remove obsolete cache-copying implementation. |
| `extensions/memory-core/src/memory/manager-db.ts` | Publish index tables without replacing the live cache. |
| `extensions/memory-core/src/memory-forget.ts` | Clear the entire cache for nonempty selections and invalidate pending work. |
| `src/plugins/openai-compatible-embedding-provider.ts` | Preserve typed HTTP metadata and redact reflected credentials. |
| `extensions/memory-core/src/memory/manager.reindex-recovery.test.ts` | Cover successful-work retention, restart, malformed output, and failure boundaries. |
| `extensions/memory-core/src/memory/manager-memory-source-race.test.ts` | Assert stale source embeddings are not retained. |
| `extensions/memory-core/src/memory/manager-embedding-cache.test.ts` | Prove replacement at capacity never transiently overflows. |
| `extensions/memory-core/src/memory/manager-embedding-policy.test.ts` | Cover successful split-leaf callbacks. |
| `extensions/memory-core/src/memory/manager-chunk-writer.test.ts` | Prove atomic index rollback while completed embeddings survive and are reused. |
| `extensions/memory-core/src/memory/manager-db.test.ts` | Prove publication preserves the canonical cache. |
| `extensions/memory-core/src/memory/manager-sync-yield.test.ts` | Remove tests for the deleted shadow-cache seed path. |
| `extensions/memory-core/src/memory-forget.test.ts` | Cover whole-cache counts, repeat purges, and empty selections. |
| `src/plugins/openai-compatible-embedding-provider.http-error.test.ts` | Exercise metadata and credential redaction through loopback HTTP. |
| `docs/cli/memory.md` | Document the Forget cache-clearing tradeoff. |
| `docs/concepts/memory-provenance.md` | Document retention and purge behavior for unfinished rebuilds. |
| `config/assertion-safety-baseline.txt` | Shrink the ratchet after removing cache-seeding code. |

</details>

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -727,7 +727,7 @@ extensions/memory-core/src/memory/manager-embedding-ops.ts	1
 extensions/memory-core/src/memory/manager-keyword-retrieval.ts	1
 extensions/memory-core/src/memory/manager-search-orchestration.ts	1
 extensions/memory-core/src/memory/manager-search.ts	6
-extensions/memory-core/src/memory/manager-sync-base.ts	4
+extensions/memory-core/src/memory/manager-sync-base.ts	3
 extensions/memory-core/src/memory/manager-vector-rebuild-state.ts	1
 extensions/memory-core/src/memory/manager-watch-ops.ts	4
 extensions/memory-core/src/memory/manager.ts	2
--- a/docs/cli/memory.md
+++ b/docs/cli/memory.md
@@ -282,9 +282,16 @@ review what remains.
 The purge removes matching promotion-marker entries and session-reference
 sections from scanned memory files, selected session-corpus lines, and
 selected-session transcript index chunks. It clears associated full-text and
-vector rows, cached embeddings, matching short-term state, ingestion seen-hash
-scopes, and origin rows. Matching content is scrubbed from dreaming rewrite
-backups, rather than deleting every backup.
+vector rows, matching short-term state, ingestion seen-hash scopes, and origin
+rows. Matching content is scrubbed from dreaming rewrite backups, rather than
+deleting every backup.
+
+For a nonempty session selection, Forget also clears the selected agent's entire
+embedding cache, including results retained from unfinished index rebuilds. Those
+results may not yet be linked to published index chunks. Unrelated published
+index entries remain usable, but later indexing may need to regenerate their
+cached embeddings. The dry-run report includes the whole-cache removal count;
+a dry run or an empty session selection does not clear the cache.
 
 Consolidation preserves origins for replaced promotion markers while retained
 rewrite preimages reference them. Those origins are pruned only after live
--- a/docs/concepts/memory-provenance.md
+++ b/docs/concepts/memory-provenance.md
@@ -192,9 +192,14 @@ merged prose. Surviving sources may support a new entry later, but automatic
 reconstruction is not guaranteed.
 
 The cleanup covers matching promoted entries, session-corpus lines, memory
-index chunks and their full-text/vector rows, cached embeddings, short-term
-state, ingestion deduplication state, and dreaming rewrite preimages. It also
-removes whole lines containing exact selected corpus snippets from scanned
+index chunks and their full-text/vector rows, short-term state, ingestion
+deduplication state, and dreaming rewrite preimages. A nonempty session selection
+also clears the selected agent's entire embedding cache, including results from
+unfinished rebuilds that are not yet linked to published chunks. Unrelated
+published index entries remain usable; later indexing may need to regenerate
+cached embeddings. Dry runs report this removal without applying it.
+
+The purge also removes whole lines containing exact selected corpus snippets from scanned
 memory files and dream diaries. The
 [command reference](/cli/memory#memory-forget) describes the counters and
 selection limits.
--- a/extensions/memory-core/src/memory-forget.test.ts
+++ b/extensions/memory-core/src/memory-forget.test.ts
@@ -48,6 +48,27 @@ describe("memory forget", () => {
     }),
   );
 
+  it.each([true, false])(
+    "reports no cache deletion for an empty selection (dryRun=%s)",
+    async (dryRun) => {
+      const db = openOpenClawAgentDatabase({ agentId: "main" }).db;
+      db.prepare(`INSERT INTO memory_embedding_cache
+      (provider, model, provider_key, hash, embedding, dims, updated_at)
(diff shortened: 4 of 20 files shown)


jalehman:
@clawsweeper re-review

Addressed the successful all-cache-hit rebuild retention finding in b1d2cdcd5cd59d3e272067381275823d4b47aca4. The new SQLite regression failed with three cache rows at a cap of one, then passed after pruning the canonical cache only after successful publication. Failed-rebuild preservation remains covered.

Also corrected the three CI chunk-publication tests to verify both intended contracts: published index tables roll back on failure, while completed embeddings remain cached and are reused without additional provider calls. All 71 focused tests, the complete changed-file gate, and final staged P0–P2 autoreview passed. The PR description now includes these results and existing-database/legacy-schema compatibility boundaries.

PR #144413, opened 2026-09-10
fix(ui): task panel squeezes subagent transcripts beside Show earlier

steipete:
## What Problem This Solves

Fixes an issue where users opening a subagent (or other background task) from the Tasks rail would see its transcript squeezed into roughly half of the right-hand detail panel whenever the task had earlier history: the **Show earlier** control sat in a left column beside the transcript instead of above it. Even without that bug, the panel reused the full chat thread for monitoring a worker in a ~450 px dock: chat bubbles, author labels, tool-card groups, and body-size text, under a header that often just said "Subagent · Subagent".

## Why This Change Was Made

The detail panel's content container was `display: flex` with no direction, so the history boundary and the transcript became side-by-side flex children. Rather than only fixing the direction, the panel now owns a compact activity feed built for monitoring: a prompt-derived title, a metadata line (status, elapsed or finished duration, tool-call count, diff stat, last tool while active), a **Now** strip with the current progress summary (or the outcome / error when finished), and one full-width row per activity item with a small kind icon, a mono one-liner per tool group that expands to every call, formatted assistant text, and right-aligned timestamps. The panel owns scrolling: it follows the bottom only while the reader is already there and keeps the visible entry in place when earlier history loads. The task-only read-only transcript adapter, its transcript controller plumbing, and the panel-only avatar override are removed; tool classification, grouping summaries, Markdown, and task history stay with their existing owners. Non-goals: the Tasks rail rows, the dock tab label, and the left sidebar are unchanged.

## User Impact

Opening a running subagent shows what it is doing right now, how long it has been running, how many tool calls it made, and a scannable feed of commands, paths, and replies that uses the whole panel width. Finished tasks show their outcome and duration in the same place. Expanding a tool row reveals the complete command, including multi-line scripts.

### Before / after (mock Gateway, same fixture)

| Before | After | After, tool group expanded |
| --- | --- | --- |
| Before: Show earlier beside a squeezed transcript | After: compact activity feed | After: expanded tool group |

### Live Gateway

Isolated local Gateway (token auth, real Anthropic model, `env -i` launch) built at the PR head; the "before" instance served the Control UI bundle built from `main` via `gateway.controlUi.root`. The main agent spawned one subagent via `sessions_spawn`; the Tasks rail row was opened on each instance. The subagent ran `ls -la`, `cat package.json | head -20`, and a `read`, then replied.

| Before (main bundle) | After (this branch) |
| --- | --- |
| Live before: chat bubbles and author labels in the panel | Live after: title, status meta, compact feed |

The header on the branch shows the agent-supplied title, `Completed · 16s · 3 tool calls`; the feed shows the prompt, one grouped tool row (`ls -la`, "Ran 2 commands, read a file"), and the formatted reply. This short transcript has no earlier history, so the side-by-side squeeze is only visible in the mock capture above; the live run proves the feed, header, grouping, and completion state against real Gateway task history.

### Checks

- `node scripts/run-vitest.mjs` on the task detail, activity feed, sidebar scroll (browser), transcript controller, embedded panels, transcript render, and tool-card tests; `ui/src/e2e/task-panel-layout.e2e.test.ts` and `ui/src/e2e/subagent-task-notice.e2e.test.ts`.
- `pnpm lint:ui:styles`, `pnpm ui:i18n:verify`, `pnpm tsgo:ui`, `node scripts/check-changed.mjs` (all lanes passed), `pnpm ui:check-performance` (startup JS within the enforced budget).
- Capped replies: session-backed task history caps assistant replies at 8,000 characters and marks them `__openclaw.truncated`; the feed recovers the full text through the same `chat.message.get` loader and bounded-retry contract the chat thread uses (state scoped to the selected task and connection, preview kept while loading, Retry after three failed attempts). Covered by feed/panel unit tests and `ui/src/e2e/task-panel-full-message.e2e.test.ts` through the Tasks rail. This addresses the ClawSweeper P2 on the first revision.
- Autoreview (Codex) on the branch: one P2 finding (expanded tool rows truncated multi-line commands) and one P2 (a second render before the animation frame dropped the pending scroll correction) fixed with regression tests; final rerun scoped-clean.


steipete:
Confirmed the capped-reply recovery finding: session-backed task history still uses the bounded chat-history projection, while the compact feed no longer invokes the existing full-message loader. This remains open, and merge is on hold until recovery is restored with coverage through the task-panel entry point.

Separately, the two CI E2E failures were stale expectations for the old header/tool-card UI and the former two-message mock transcript. Fixed in fb6a613f2ea2b0430f83c4520a313e160c4bd53c. The updated tests retain task lifecycle, paging, command expansion, prompt-before-reply ordering, and history/startup parity. Both repaired scenarios passed locally, and the test-only diff passed scoped autoreview. The broader local run passed 20 tests but hit an unrelated native-plugin mock-server startup timeout; that untouched scenario passed in the original CI run.

steipete:
The capped-reply recovery finding is addressed by 7e3177f714dd23de436b10cd754a9e647a42fc15 and 0fd401b382f08e0c7173777fe8487f96e8fa4fe8. The activity feed now uses the existing full-message eligibility, loader, and bounded-retry contract, with recovery scoped to the selected task and connection and performed once per message.

`ui/src/e2e/task-panel-full-message.e2e.test.ts` covers full-message recovery through the Tasks rail; component tests cover recovery, retry, and stale results. The repair has completed maintainer review and scoped-clean autoreview. Resuming exact-head CI and native landing on 0fd401b382f08e0c7173777fe8487f96e8fa4fe8.

steipete:
Merged via squash.

- Prepared head SHA: [0fd401b382f08e0c7173777fe8487f96e8fa4fe8](https://github.com/openclaw/openclaw/pull/144413/commits/0fd401b382f08e0c7173777fe8487f96e8fa4fe8)
- Landed commit: [238f099b0a8b7bfba29ea3dec9bf54a2aaf8153d](https://github.com/openclaw/openclaw/commit/238f099b0a8b7bfba29ea3dec9bf54a2aaf8153d)

PR #144416, opened 2026-09-10
fix(gateway): cap cloud workspace staging re-read

SebTardif:
## What Problem This Solves

Cloud workspace staging already hashes each matched file under a 64 MiB cap. After that match it used to call uncapped `fs.readFile` to load the same path for the git fast-import payload. A worker that grows or replaces the file in that window can make the gateway allocate the new size before the later size/hash check throws. The gateway should refuse the over-cap file without loading it.

## Evidence

Live `tsx` on this branch compared the old uncapped `fs.readFile` with the new capped reader on a sparse file grown to 64 MiB + 1 after a valid small hash.

```console
$ pnpm exec tsx /tmp/proof-f075.mjs
node v26.7.0
capBytes 67108864
small.type file
small.size 7
huge.snapshot unsupported
uncapped.byteLength 67108865
uncapped.ms 10.7
capped.error Cloud workspace staged result exceeds its byte limit
capped.ms 0.3
```

Uncapped `fs.readFile` returned 67,108,865 bytes. The capped reader returned `unsupported` and threw `Cloud workspace staged result exceeds its byte limit` in 0.3ms.

## Real behavior proof

- **Behavior or issue addressed:** Cloud workspace staging no longer re-reads a just-hashed worker file with uncapped `fs.readFile`. An over-cap replacement is rejected as exceeding the byte limit instead of being loaded into the gateway process.
- **Real environment tested:** macOS, Node v26.7.0, worktree `/tmp/oc-f075` at `origin/main` `f9976d6de6a`, invoked with `pnpm exec tsx /tmp/proof-f075.mjs`.
- **Exact steps or command run after this patch:** `pnpm exec tsx /tmp/proof-f075.mjs` after writing a 7-byte file, hashing it, then truncating a sibling path to 64 MiB + 1.
- **Evidence after fix:** terminal output copied below.

```console
$ pnpm exec tsx /tmp/proof-f075.mjs
node v26.7.0
capBytes 67108864
small.type file
small.size 7
huge.snapshot unsupported
uncapped.byteLength 67108865
uncapped.ms 10.7
capped.error Cloud workspace staged result exceeds its byte limit
capped.ms 0.3
```

- **Observed result after fix:** The capped staging read fails closed on the grown file. The same path still loads 67,108,865 bytes when `fs.readFile` is called directly, which is the allocation the patch removes from `stageWorkerWorkspaceResult`.
- **What was not tested:** A live cloud worker process growing a file during an in-flight gateway stage. The proof uses the production readers and a truncate in the hash-then-read window.

## Summary

`absoluteEntryMatches` already uses `readWorkspaceFileSnapshotWithLimit`. Staging now loads payload bytes through the same cap (`readCappedStagedFile` / `readWorkspaceFileBytesWithLimit`) and throws `Cloud workspace staged result exceeds its byte limit` when the snapshot is unsupported. Under-cap size or hash drift still throws `changed while reading`.

This bug landed in [7a7d6bb51f42](https://github.com/openclaw/openclaw/commit/7a7d6bb51f42) / [#110952](https://github.com/openclaw/openclaw/pull/110952) (2026-07-18). No competing PR owns `workspace-result-staging.ts` for this re-read.

No changelog entry. The repo generates those.

### DIFF
--- a/src/config/types.secrets.ts
+++ b/src/config/types.secrets.ts
@@ -7,7 +7,6 @@ import {
   DEFAULT_SECRET_PROVIDER_ALIAS,
   ENV_SECRET_REF_ID_RE,
   isSecretRef,
-  isValidEnvSecretRefId,
   type SecretRef,
   type SecretRefSource,
 } from "../secrets/ref-contract.js";
--- a/src/gateway/worker-environments/workspace-actual-manifest.ts
+++ b/src/gateway/worker-environments/workspace-actual-manifest.ts
@@ -25,7 +25,7 @@ import {
 import { isDerivedWorkspacePath } from "./workspace-path-exclusions.js";
 
 type WorkspaceFileSnapshot =
-  | { type: "file"; mode: number; size: number; sha256: string }
+  | { type: "file"; mode: number; size: number; sha256: string; content?: Buffer }
   | { type: "unsupported" };
 
 function localPath(root: string, relative: string): string {
@@ -54,6 +54,7 @@ export async function readWorkspaceFileSnapshotWithLimit(
   maxBytes: number | ((openedSize: number) => number),
   root?: string,
   signal?: AbortSignal,
+  options?: { includeContent?: boolean },
 ): Promise<WorkspaceFileSnapshot> {
   signal?.throwIfAborted();
   const handle = await fs.open(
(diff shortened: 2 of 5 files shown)


SebTardif:
@clawsweeper re-review

SebTardif:
@clawsweeper re-review

PR #144417, opened 2026-09-10
fix(pdf): keep model selection and reporting consistent

steipete:
Related: #130706.

## What Problem This Solves

Fixes an issue where PDF analysis could plan a bare model reference under the wrong provider when captured and ambient metadata differed. Results could also report a raw alias while completion used its selected model: `entry` dispatched `middle`, but details still said `entry`.

## Why This Change Was Made

The candidate planner now owns default-provider derivation and candidate parsing under one metadata view. PDF supplies its acquired metadata and preserves the selected logical ID during materialization. Both existing alias-index phases remain; runtime provider hooks still own wire-model normalization.

The obsolete default-provider helper and redundant internal argument are removed. Production code is **12 lines smaller**. Image compression only receives the mechanical caller cleanup; its separate model handoff behavior is outside this fix.

## User Impact

PDF results identify the selected logical model consistently. Arcee still reports `arcee/trinity-large-thinking` while OpenRouter receives `arcee-ai/trinity-large-thinking`. Native PDF behavior and resource/cancellation ownership retain their existing contracts. No configuration, schema, or persisted fields are added.

## Evidence

Validated commit `563b907724483e606e30eb77660238bf426cbfde` against main `a41bd428af45d2cd8538fbfee860572a47cf0850`.

- Original production reproduced four wrong-provider planning failures and two PDF reporting failures; both preservation controls passed.
- **197 focused and sibling tests passed**, including seven new cases plus existing image, native PDF, supplied-runtime, and cancellation coverage.
- The full changed-file gate and fresh P2 review passed with no findings. One normal full build passed.
- **Eleven final compiled cases passed** through the real tool builder and document extractor: six cold API/alias cases, matching/empty/foreign supplied snapshots, and routed/direct Arcee controls. Source imports were rejected, payload bytes stayed unchanged, and selected provider ownership/policies were verified.
- All **14,104 build files** stayed unchanged. Every runtime child joined; listeners, sockets, and owned runtime state were cleaned up.

The transport fixture retains canonical provider URLs for classification and sends unchanged payload bytes to an isolated loopback responder. It retains hostname policy checks but does not prove DNS pinning or external vendor transport. Source/built metadata differ because packaging adds channel schemas; that enrichment is checked separately from selected PDF ownership, without claiming global owner-map equivalence. Earlier failed harness records remain unchanged, and all eleven compiled cases passed in a fresh run after the metadata assertions were corrected.

### DIFF
--- a/src/agents/model-fallback-candidates.test.ts
+++ b/src/agents/model-fallback-candidates.test.ts
@@ -1,11 +1,14 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import type { ModelProviderConfig } from "../config/types.models.js";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
 import { createWarnLogCapture } from "../logging/test-helpers/warn-log-capture.js";
+import { createPluginMetadataSnapshotFixture } from "../plugins/plugin-metadata.test-support.js";
+import { withPluginRuntimeGenerationScope } from "../plugins/runtime/generation-scope.js";
 import {
   resolveImageFallbackCandidates,
   resolveModelCandidateChain,
 } from "./model-fallback-candidates.js";
+import { runWithImageModelFallback } from "./model-fallback-image.js";
 import { makeProviderModelFixture } from "./test-helpers/provider-model-fixture.js";
 
 const customProvider: ModelProviderConfig = {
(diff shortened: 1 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [563b907724483e606e30eb77660238bf426cbfde](https://github.com/openclaw/openclaw/pull/144417/commits/563b907724483e606e30eb77660238bf426cbfde)
- Landed commit: [00a859634fbf259b8de773a8f116aeb5094070b9](https://github.com/openclaw/openclaw/commit/00a859634fbf259b8de773a8f116aeb5094070b9)

PR #144359, opened 2026-09-10
improve: keep baseline maintenance out of unrelated CI tests

steipete:
Related: #139428

## What Problem This Solves

Max-lines baseline maintenance can add tooling tests and their runtime prerequisites to an otherwise focused test change, even when CI already emits the required baseline-ratchets guard. For example, a Matrix test plus baseline cleanup should retain the Matrix and boundary coverage without selecting extra tests solely for the baseline data.

## Why This Change Was Made

The CI manifest tells the changed-test planner when it emits the existing required baseline-ratchets task. Only that task's exact `config/max-lines-baseline.txt` input receives credit. The ratchet still checks the tested merge tree against its prepared base and rejects expansion or stale entries.

The integration retains current main's canonical precise-tooling routes. Without credit, baseline owner tests remain selected. Mixed changes retain explicit owner-test coverage and runtime prerequisites; planner changes and unresolved paths retain broad fallback. The conflict resolution preserves both sets of regression tests and places ratchet cases beside related dedicated-coverage tests.

## User Impact

Maintainers can include required baseline cleanup without adding unrelated tests solely for that data file. Companion tests and boundary coverage remain selected. Workflow permissions, concurrency, required gates and full `pnpm test` behavior are unchanged. No overall runtime improvement is claimed.

## Evidence

- The conflict-only integration exposed five obsolete full-fallback assertions: 699 tests passed, five failed, and two existing cases were skipped. Current main now selects precise tooling for those inputs.
- Updated independent literal oracles verify the three baseline owner tests without credit, Matrix companion coverage, precise ratchet/workflow owner tests and boundary prerequisites, planner fallback, and missing/unknown input rejection. The workflow fixture verifies that a failed or skipped required ratchet guard rejects the aggregate gate.
- Complete current planner and workflow-guard suites: **704 passed, two existing skips**. After reusing the existing group-normalization helper, all **164 planner cases** passed again, with every case name and literal oracle preserved.
- Changed-file gates, workflow sanity and fresh independent review passed. Local proof used the existing fixed Bash 5.3/Python 3.14 environment and the private frozen dependency graph. Required exact-head CI remains the landing gate.

Earlier UI-budget and Windows timeout failures are retained separately. This PR neither changes their thresholds nor claims the original Windows timeout cause is fixed.

PR #144419, opened 2026-09-10
fix: stabilize paginated history eviction coverage

steipete:
Related: #144359

## What Problem This Solves

Fixes an intermittent failure in the paginated-history eviction E2E test: its second upward scroll could arrive before the newly loaded history reached the rendered viewport, leaving the test at the retained message rather than the newly loaded start.

## Why This Change Was Made

The test's 140-message model check can pass while the virtualizer deliberately retains the previous rows during scrolling. Require observable scroll-extent growth and a restored anchor, then reuse the existing scroll-idle helper before the same second gesture. The normal poll settings, timeouts, two gestures, message-count check, visibility check, and eviction assertions remain unchanged. No diagnostic instrumentation is included.

## User Impact

More reliable UI regression coverage without changing production behavior. This is separate from the CI-routing change where the failure surfaced.

## Evidence

- Original Linux CI failed waiting for the older first message before reaching eviction assertions; 341 other cases passed.
- A bounded local timing control changed only the model poll interval to 1 ms on both versions, retaining the configured 15-second poll timeout. The original failed at the same visibility assertion with the second wheel captured before the prepend committed; the candidate passed the visibility and eviction contract.
- Uninstrumented history-recovery and prepend-anchor E2E suites: all 15 cases passed before and after at normal polling settings, with identical ordered cases per file.
- Changed-file gate and independent review passed. Removing the 11 added lines restores the original test byte for byte.

Local proof used Node 26.8.1 and Chromium via Playwright 1.62.1 on macOS. It reproduces the timing mechanism, not the exact Linux CI environment. No full-suite runtime improvement is claimed.

### DIFF
--- a/ui/src/e2e/chat-flow.history-recovery.e2e.test.ts
+++ b/ui/src/e2e/chat-flow.history-recovery.e2e.test.ts
@@ -585,6 +585,7 @@ suite.define(() => {
       const activePane = page.locator('openclaw-chat-pane[aria-hidden="false"]');
       const thread = activePane.locator(".chat-thread");
       await thread.hover();
+      const previousScrollHeight = await thread.evaluate((element) => element.scrollHeight);
       await page.mouse.wheel(0, -1_000_000);
       await expect
         .poll(() =>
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8659b541d12474e4b36cd11be80adf22ba700fb5](https://github.com/openclaw/openclaw/pull/144419/commits/8659b541d12474e4b36cd11be80adf22ba700fb5)
- Landed commit: [b7fa461705596c529f5384a3837139e4ee0b1a0e](https://github.com/openclaw/openclaw/commit/b7fa461705596c529f5384a3837139e4ee0b1a0e)

PR #144429, opened 2026-09-10
fix(agents): normalize bare model defaults with captured metadata

steipete:
Related: #130706. Prerequisite for the bare-default cases held in #144354 and CLI model validation.

## What Problem This Solves

Fixes an issue where a simple completion falling back to a bare configured primary could reject the model or use the wrong metadata, despite already having the provider's normalization policy. For `entry → middle → final`, the fallback could retain raw `entry` while later preparation treated it as fully selected.

## Why This Change Was Made

Captured bare defaults now continue through the existing model-reference parser. It reuses cached metadata, normalization flags, auth-profile parsing and configured API-owner handling. Callers without captured metadata keep the existing cheap raw path; already resolved branches return before this point. The production change is **two added lines**.

## User Impact

Default fallback selection uses the provider's intended model without resolving an already-selected alias twice. No configuration, schema, persisted fields, imports, or public API options are added, and this change introduces no new metadata discovery.

## Evidence

Validated `62675418c10dc22d7f1b673c3b20145543c3acf3` against main `f9976d6de6a086c3b19edbfaeb7412db47b50bb7`.

- Unmodified-main source baseline: nine intended failures and 35 passing controls.
- **265 focused tests across seven files passed**, followed by the full pinned changed-file gate and fresh independent P2 review.
- One normal full build passed with natural metadata bound to the candidate commit.
- **Five fresh public SDK processes made five real loopback HTTP requests.** Invalid-override fallback, valid override, inherited default, raw next-hop and exact-ID cases verified selection, materialized context, response text and request model. Nine public model-config checks covered metadata/normalization modes.
- All **14,636 build outputs** and source stayed unchanged; **3,983 executed files** were verified. Source-import rejection controls passed, all five processes exited naturally, and owned runtime state was removed.

Two earlier compiled fixture attempts remain failed and preserved: an explicit configured API mapping treated `/` as a literal ID, and a disabled canonical provider owner prevented the synthetic plugin from loading. The final fixture verifies the invalid-override precondition and enables its isolated synthetic provider replacement. Source and build were unchanged between attempts. This is compiled SDK/loopback proof, not external vendor inference; it does not borrow the unlanded TTS patch.

### DIFF
--- a/src/agents/model-selection-manifest-workspace.test.ts
+++ b/src/agents/model-selection-manifest-workspace.test.ts
@@ -31,6 +31,17 @@ vi.mock("./provider-model-normalization.runtime.js", () => ({
   normalizeProviderModelIdWithRuntime: normalizeProviderModelIdWithRuntimeMock,
 }));
 
+const defaultNormalizationSnapshot = createPluginMetadataSnapshotFixture({
+  plugins: [
+    {
+      id: "default-normalizer",
+      modelIdNormalization: {
+        providers: { openai: { aliases: { entry: "middle", middle: "final" } } },
+      },
+    },
+  ],
+});
+
 describe("configured model manifest workspace scope", () => {
   beforeEach(() => {
     loadManifestMetadataSnapshotMock.mockReset();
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [62675418c10dc22d7f1b673c3b20145543c3acf3](https://github.com/openclaw/openclaw/pull/144429/commits/62675418c10dc22d7f1b673c3b20145543c3acf3)
- Landed commit: [cc748884201771eeac74dc6d8cf0c5b758bade76](https://github.com/openclaw/openclaw/commit/cc748884201771eeac74dc6d8cf0c5b758bade76)

PR #144430, opened 2026-09-10
fix(usage): refresh details after a session is recreated

steipete:
## What Problem This Solves

Fixes an issue where the Control UI Usage page showed the previous session's timeline and conversation after the selected session was deleted and recreated under the same key and agent. The overview could show the replacement's 600 tokens while its details still showed the old 200-token conversation.

## Why This Change Was Made

Bind the existing detail cache and pending recovery to the overview row's optional session ID. Reject a context response that explicitly belongs to a different instance through the existing localized error/recovery flow. Logical selection, Gateway request parameters, missing-ID compatibility and server cache timing retain their current behavior.

## User Impact

Once the overview adopts the replacement session, its details refresh without clearing selection. A mismatched context response shows an error with a Refresh instruction instead of displaying another instance's details.

## Evidence

Three owner-level failures reproduce before the fix, with nine passing controls. All 83 final owner/sibling/browser cases and 20 selected changed-file checks pass; independent review found no actionable issue. The keyless i18n check leaves generated and baseline files unchanged.

The real Gateway/browser flow creates a synthetic session, deletes it, recreates the same key with a different session ID, and refreshes Usage on the same WebSocket connection. It verifies the replacement overview, retained selection, and corrected timeline/conversation. The existing server cache is allowed to refresh; this does not claim immediate backend freshness. Stale context response rejection and recovery are covered at the mounted page owner boundary.

The four inspected synthetic captures below show the selected replacement's overview before/after, followed by its timeline and conversation before/after. The first pair keeps the 600-token replacement selected; the second shows the stale 200-token Original replies becoming the replacement's 600-token Replacement replies. Dates and version labels are fixed fixture metadata. Initial fixture-preparation failures are retained separately and excluded from behavioral evidence; final proof uses an explicit private log target and silent startup logging.

Before: the replacement session is selected with 600 tokens but retains the old timeline

After: the same selected replacement shows its matching timeline

Before: stale 200-token timeline and Original replies remain

After: replacement 600-token timeline and Replacement replies appear

### DIFF
--- a/docs/web/control-ui/feature-reference.md
+++ b/docs/web/control-ui/feature-reference.md
@@ -93,6 +93,7 @@ Control UI capabilities grouped by area, each with the Gateway RPC methods behin
     - A provider usage failure does not block the session/cost dashboard; unavailable provider cards show their own error state.
     - Incomplete session/cost totals stay readable while the visible, focused page checks for updates. Automatic checks are bounded; if they pause, select **Refresh** to check again.
     - **Refresh** also reloads the selected session's timeline, conversation, and system-prompt breakdown.
+    - If a selected session is deleted and recreated, its new details replace the old ones without clearing your selection. Context details from a different session instance show an error and can be retried with **Refresh**.
     - The overview loads session summaries first. Full system-prompt breakdowns load when you select a session; the `has:context` filter still works before opening details.
     - JSON exports keep the displayed usage snapshot and load prompt details for the same session instance. If that session has been replaced, refresh Usage and export again.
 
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8160669c87a89f4d1acfb09e019690161aa336c0](https://github.com/openclaw/openclaw/pull/144430/commits/8160669c87a89f4d1acfb09e019690161aa336c0)
- Landed commit: [96a0e72498bdb1f0b71d8ec8a9748f0b8b43fd34](https://github.com/openclaw/openclaw/commit/96a0e72498bdb1f0b71d8ec8a9748f0b8b43fd34)

PR #144432, opened 2026-09-10
fix(compaction): preserve configured fallback model selection

steipete:
Related: #130706

## What Problem This Solves

When the primary compaction model fails, a configured fallback alias can be normalized twice. An alias chain such as `entry → middle → final` can therefore summarize with `final`, or fail while refreshing model metadata, even though fallback selection chose `middle`.

## Why This Change Was Made

Carry the selected identity of private non-primary fallback candidates through initial model resolution and credential refresh. The existing candidate builder owns alias selection; these consumers now preserve its choice. Primary/raw inputs and provider-auth separation retain their existing behavior.

The production change is four lines across two files. It stands on current main without a separate candidate-marker change.

## User Impact

Compaction uses the configured fallback model consistently after a primary failure, including when account-specific model metadata must be refreshed.

## Evidence

- The same regression tests on unmodified main fail in the two expected ways: ordinary compaction reaches `final`; credential refresh rejects rematerialization for selected `middle`. Four primary-plan and same-/cross-provider auth controls pass on that baseline.
- All 228 focused tests across seven files pass on the candidate. The complete changed gate is pinned to the main base, and fresh P2 review is clean.
- One normal full `pnpm build` passed, including SDK declarations and export validation.
- Five compiled direct-compaction scenarios pass: ordinary fallback, credential refresh, same-provider profile, cross-provider profile, and raw-primary preservation. The final run records four primary 429 responses, five successful completions, correct auth labels, and five persisted summaries read back through the canonical session API. All 11,943 build outputs and the committed source remain unchanged.

The first compiled attempt met its routing assertions but used an unsupported `reserveTokens` fixture key and a noncanonical store address. It is retained as superseded diagnostics, not final proof. After correcting only the fixture, all five scenarios were rerun against the same unchanged build. Provider responses came from a synthetic loopback endpoint; no external vendor inference is claimed.

### DIFF
--- a/src/agents/embedded-agent-runner/compact.hooks.harness.ts
+++ b/src/agents/embedded-agent-runner/compact.hooks.harness.ts
@@ -20,6 +20,7 @@ import {
   agentSessionSetContextReplacementHook,
 } from "../sessions/agent-session-compaction.js";
 import type { SessionManager } from "../sessions/session-manager.js";
+import type { resolveModelAsync } from "./model.js";
 import type { attemptServerEndpointCompaction } from "./server-endpoint-compaction.js";
 import type { buildEmbeddedSystemPrompt } from "./system-prompt.js";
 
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e5c7a514e3492d2da34c760a7a0ce090789b6cf7](https://github.com/openclaw/openclaw/pull/144432/commits/e5c7a514e3492d2da34c760a7a0ce090789b6cf7)
- Landed commit: [c9947af2b3da6fe60757758df1bc0a8dfcc9b5e6](https://github.com/openclaw/openclaw/commit/c9947af2b3da6fe60757758df1bc0a8dfcc9b5e6)

PR #144441, opened 2026-09-10
refactor(discord): simplify action runtime dependencies

steipete:
## What Problem This Solves

Maintaining Discord actions currently means keeping private copies of the same send-function dependency wiring in sync. This refactor removes that duplication without changing the public action adapter.

## Why This Change Was Made

Replace the private guild and moderation dependency bags with existing module bindings, and replace the messaging bag with named re-exports from its existing owners. Policy and dispatch bodies stay intact; the public adapter and action vocabulary remain. Tests intercept those owners through module mocks while retaining their original implementation-reset behavior.

Production changes are **+12/−99 (87 fewer lines)**; tests are **+188/−118 (+70 lines)**. No new SDK, configuration, schema, or dependency surface is introduced.

## User Impact

No intended user-visible behavior change. Account selection, read gates, moderation authorization, message content and attachment handling, and the public action entry points are preserved. The benefit is one fewer layer of private dependency wiring to maintain.

## Evidence

- The same six-file selection passes all **232 cases** on the original, candidate, restored candidate, and final lint-corrected source, with identical case names and per-file order. A deliberate edit-message trimming fault produced **four expected HTTP-body assertion failures** across both public entry routes, with 46 other cases passing; restoring the candidate returned all 232 to green. Assertions were not weakened. Silent-delivery propagation is covered by mocked runtime tests, not the loopback body suite.
- The first full changed gate genuinely failed on six lint errors: four shadowed hoist-local names and two import-order errors. The correction renames only inner mock bindings and orders the messaging imports after checking the eager dependency graph: the first two eager owners remain first, and the moved helpers were already traversed. Targeted lint and fresh managed P2 precommit review pass; a separate exact-commit P2 review reports no actionable findings.
- Normal extension production/test type checks, `pnpm build`, and the canonical standalone Discord package builder pass. All **98 package files** retain identical bytes and modes between the pre-lint-correction and final candidate builds; this is not original-versus-refactor artifact parity.
- The final candidate's built public exports pass **10 synthetic loopback cases** through plain Node, producing 25 HTTP requests under isolated HOME/state/config/agent fixtures. Source, runtime and artifact pins remain exact; fixtures, child process group, listener and artifact ownership are released.
- The complete configured [25-stage changed gate R2](https://github.com/openclaw/openclaw/actions/runs/34530953670) exits 0, including lint, type checks and runtime-cycle checks. Its lease, capsule, claims and owned processes are closed. The linked one-shot Actions holder may show cancelled after normal lease cleanup; the saved native command result is the gate verdict. A post-command portal-sync warning is retained separately from the successful checks and independently verified cleanup. The tested tree is exactly the tree in signed commit `44a4c19737af7491b717317d2c2cb93970751631`.

These results cover that exact candidate on its recorded base, not a runtime test of every later main change. They are not live Discord, full Gateway, production-only install, or timing evidence.


steipete:
Merged via squash.

- Prepared head SHA: [44a4c19737af7491b717317d2c2cb93970751631](https://github.com/openclaw/openclaw/pull/144441/commits/44a4c19737af7491b717317d2c2cb93970751631)
- Landed commit: [bfac43ad2e78a21676eb36079111e5efb528e4dd](https://github.com/openclaw/openclaw/commit/bfac43ad2e78a21676eb36079111e5efb528e4dd)

PR #144445, opened 2026-09-10
refactor(reply): remove duplicate follow-up context

steipete:
## What Problem This Solves

Follow-up admission keeps a second copy of inbound context even though execution reads it from the queued turn. Payload preparation also returns a heartbeat logging flag that its completion consumer never reads.

## Why This Change Was Made

Remove the unused context mirror and its writes, leaving the queued turn as the single context source. Remove only the unused final returned logging field; payload construction keeps its live state for avoiding duplicate heartbeat-strip logs. Existing admission and compaction assertions now check the actual execution input.

## User Impact

No intended user-visible behavior change. This removes five production lines and a duplicate assertion while preserving context refresh, compaction and heartbeat presentation.

## Evidence

Two existing context-preservation controls pass before the cleanup. Afterward, all 85 selected admission, execution, presentation and ordinary-finalizer test rows pass, along with 29 selected changed-file checks. The final comment correction has identical emitted JavaScript and is included in the passing checks.

The initial combined JSON test invocation stopped at the runner's project/reporter guard after 84 valid passes, before its finalizer case ran. That one unexecuted case then passed in a separate supported invocation. Both required runtime prebuilds and the initial guard failure are retained; no assertion or deadline changed.

### DIFF
--- a/src/auto-reply/reply/agent-runner-result-payloads.ts
+++ b/src/auto-reply/reply/agent-runner-result-payloads.ts
@@ -644,7 +644,6 @@ export async function prepareReplyAgentPayloads(state: {
     kind: "continue" as const,
     activeSessionEntry,
     completedSourceReplyDelivery,
-    didLogHeartbeatStrip,
     guardedReplyPayloads,
     responseUsageLine,
   };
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [fce192f0986ac78e6f1d0f935d68c744cf29be2e](https://github.com/openclaw/openclaw/pull/144445/commits/fce192f0986ac78e6f1d0f935d68c744cf29be2e)
- Landed commit: [cfab7b6a640e9495dcd16ec5021905fc020459d7](https://github.com/openclaw/openclaw/commit/cfab7b6a640e9495dcd16ec5021905fc020459d7)

PR #144448, opened 2026-09-10
fix(sqlite): speed up read-only snapshot startup

steipete:
## What Problem This Solves

Resolves a problem where each short-lived read-only SQLite snapshot process loads shared runtime chunks containing work it does not need. Update tests launch these processes repeatedly, making cold startup a material part of their runtime.

## Why This Change Was Made

Emit the existing snapshot executable as one standalone bundle in both production and the invocation-owned test compiler. One entry classification removes it from the shared executable graph, while parent orchestration and SDK code retain shared runtime identity. The stable worker path, native process protocol, dependency policy, schema inlining, lazy package loading and artifact verification remain intact.

## User Impact

Less cold loading for existing read-only snapshots. Runtime implementation, storage semantics and worker/shard counts are unchanged. The standalone artifact duplicates some helpers used by other outputs; it is 155.8 KB (39.4 KB gzip), which is not a measurement of incremental package size.

## Evidence

The unchanged complete update-CLI file plus native sync/async sibling passed all 503 cases in each local arm:

| Whole command, including preparation and cleanup | Time |
| --- | ---: |
| Original | 620.00s |
| Candidate | 501.08s |
| Restored original | 542.54s |

The candidate saved 41.46s (7.6%) against the nearer restored control. The first two arms had a 19m36s pause, and the original controls differed by 77.46s; this is not a guaranteed whole-CI or cross-platform saving. Test inventory/order, assertions, dependencies and execution settings were unchanged. This comparison is pinned to its earlier source base; the rebased candidate was subsequently validated against current source and its frozen dependencies.

The earlier native-child comparison reduced the local import closure from 47 files to one. The final production build also emits a single-file worker; real built sync/async SQLite smoke checks passed. All 479 focused owner/native/config cases passed with one existing platform skip, along with five selected source-freshness/refusal/artifact checks, the normal production build, complete changed-file gates and independent P2 review.

Build-tooling delta is +36 lines for the dedicated executable boundary and its two consumers; existing config guards gain 32 net lines. Runtime implementation delta is zero. No new settings, process pool, cache or test-only runtime seam. Exact-head CI remains required before landing.

### DIFF
--- a/scripts/lib/runtime-process-core-build-entries.mts
+++ b/scripts/lib/runtime-process-core-build-entries.mts
@@ -19,3 +19,16 @@ export function createRuntimeProcessBuildEntries(
 export const runtimeProcessCoreBuildEntries = createRuntimeProcessBuildEntries(
   Object.values(runtimeProcessEntrypoints),
 );
+
+// Short-lived snapshot children own a separate bundle; parents retain shared runtime identity.
+export const standaloneRuntimeProcessBuildEntries = createRuntimeProcessBuildEntries([
+  runtimeProcessEntrypoints.sqliteReadOnly,
+]);
+
+export function sharedRuntimeProcessBuildEntries(entries: Record<string, string>) {
+  return Object.fromEntries(
+    Object.entries(entries).filter(
+      ([name]) => !Object.hasOwn(standaloneRuntimeProcessBuildEntries, name),
+    ),
+  );
+}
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [fbad8a5e3a6c0e08eab5e4f3039e4ecc36bb54ae](https://github.com/openclaw/openclaw/pull/144448/commits/fbad8a5e3a6c0e08eab5e4f3039e4ecc36bb54ae)
- Landed commit: [7b230efeac324ed70d24fafa3d1f2beeb69bfce6](https://github.com/openclaw/openclaw/commit/7b230efeac324ed70d24fafa3d1f2beeb69bfce6)

PR #121735, opened 2026-08-10
fix: prevent database verifier process CI hangs

steipete:
## What Problem This Solves

Fixes an issue where the database-verifier IPC regression could fail main CI when its real Node/tsx child was starved by the shared compact unit shard.

The failure appeared in [main CI run 31425170184](https://github.com/openclaw/openclaw/actions/runs/31425170184), where `openclaw-database-verify.process.test.ts` exhausted its child-exit deadline.

## Why This Change Was Made

The verifier process regression was missing from the canonical source-child process list, so full CI ran it in the shared unit graph. The existing isolated process project already owns this lifecycle: it runs source children serially in isolated forks and disables esbuild worker threads to prevent synchronous loader IPC hangs.

This adds the verifier regression to that owner list and derives the shared-unit exclusions from the same list, keeping focused and full-suite routing aligned. The production verifier, timeout, and assertion are unchanged.

## User Impact

No product behavior changes. Main CI no longer exposes this process-level regression to unrelated compact-shard contention.

Production verifier/runtime LOC: +0/-0. Current routing, config, and test/support delta: +55/-7 (six files).

## Historical evidence from the original submission

- Pre-fix routing proof: [Testbox run 31431131090](https://github.com/openclaw/openclaw/actions/runs/31431131090) failed because the verifier process test was still classified into the shared unit project.
- Focused post-fix proof: [Testbox run 31433240250](https://github.com/openclaw/openclaw/actions/runs/31433240250) passed 361 routing/config tests and the verifier IPC regression in the isolated process project.
- Stress proof: the focused verifier process test passed 8 consecutive runs on Testbox `tbx_01kzprnb5xnb426eapv3s3m4jt`.
- Changed-file gate: formatting, test-root typecheck, script lint, conflict-marker, dependency-pin, patch, and repository guard checks passed on the same Testbox lease.
- Autoreview: clean, with no accepted/actionable findings.

Provenance: the verifier process regression was added in `c15ab158b06b` / #114016. The isolated source-child loader contract was added later in `57c6f7622133` / #117830, but the state verifier test was not enrolled in that lane.

## Current conflict integration and qualification

The current-main integration preserves every expanded source-child process entry, the Gateway/watch/E2E/database-worker owners, and Peter Steinberger's original contributor ancestry. It uses the existing router and serial isolated-process config; no new runner, package, verifier behavior, or timeout was introduced.

[...]

PR #144451, opened 2026-09-10
fix: avoid shutdown hangs after database verifier IPC errors

steipete:
Related: #121735 (test-runner isolation; this PR repairs production worker custody).

## What Problem This Solves

Fixes an issue where operators stopping the Gateway could hang after the background database verifier lost its IPC connection. Failed signal delivery could also complete the verifier's shutdown step while its child was still alive.

## Why This Change Was Made

Verification and stop now share one child lifecycle, recorded before the child is exposed to callers. A spawned child completes after native exit and IPC disconnect; a failed native launch completes after close. Cleanup removes only the verifier's listeners, preserves the original failure when termination also errors, and reports refused or thrown signals while retaining ownership.

Full integrity and foreign-key checks, generation-bound quarantine, schemas, and daily scheduling are unchanged. No new timeout or signal escalation is introduced.

## User Impact

Gateway shutdown remains joined to the database verifier, and diagnostics retain the failure that triggered cleanup. A termination request that the operating system refuses remains visible and pending until the child exits.

## Evidence

- Four regressions failed on the original production source: premature release and a lost stop waiter, undelivered termination, failed-spawn closure, and masking the original IPC error. Two additional review regressions reproduced a thrown signal error and completion before IPC disconnect.
- All six lifetime cases and the existing verifier coverage pass: **49/49 locally on Node 26.8.1 and 49/49 on Linux Node 24.19.0**. Tests use real child processes with explicit IPC, signal-failure, and notification-order fault injection.
- The scoped `check:changed` gate passes, including core/test typechecks, lint, formatting, schema guards, and import-cycle checks.
- Independent autoreview is clean through P2 after addressing both findings.
- [Linux Testbox proof](https://github.com/openclaw/openclaw/actions/runs/34536114016): the focused command exited zero and the lease was closed. The hosting workflow was cancelled during external lease cleanup; that is separate from the successful test command.

### DIFF
--- a/docs/reference/database-schemas/integrity-and-recovery.md
+++ b/docs/reference/database-schemas/integrity-and-recovery.md
@@ -32,6 +32,13 @@ Quarantine decisions live only in a dedicated `openclaw-quarantine.sqlite` store
 
 Background verification errors retain the original name and message and append bounded Node `code` and SQLite `errcode` values from up to eight cause-chain nodes. These diagnostics do not change the verdict: I/O failures remain inconclusive, while proven corruption is reconfirmed by the database owner before quarantine. A generic `disk I/O error` (`errcode=10`) does not establish disk exhaustion.
 
+The background verifier retains its child through native exit and IPC disconnect,
+including when sending work fails. Failed native launches settle after closure
+without requiring an exit event. If the operating system refuses a termination
+request, the verifier logs the failure and keeps waiting for native exit; an
+undelivered signal does not mean the child has stopped. The original worker or
+IPC error remains the reported failure even if termination also fails.
+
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [e76ea1de36d0f40fae647e75a47a6f22ebd16b0f](https://github.com/openclaw/openclaw/pull/144451/commits/e76ea1de36d0f40fae647e75a47a6f22ebd16b0f)
- Landed commit: [569750c79d049ac5deb9cc85db5e86b6be168eeb](https://github.com/openclaw/openclaw/commit/569750c79d049ac5deb9cc85db5e86b6be168eeb)

PR #144452, opened 2026-09-10
docs: draft generic external supervisor controls contract

MythiliMur:
## Summary

Documents a **draft-only**, generic contract for optional Control UI settings owned by an external Gateway supervisor.

- Keeps platform-specific semantics, persistence, enforcement, and verification out of OpenClaw core.
- Requires a local authenticated allowlisted bridge before a future implementation can make settings mutable from Control UI.
- Requires UI to distinguish a requested posture from an actually effective security boundary.
- Explains why a feature-specific `system.info` field is the wrong integration surface.

## Deliberately documentation-only

There are no protocol, Gateway handler, or Control UI changes in this PR. The Windows packaging side must first provide a real isolation boundary and a documented supervisor transport. This PR is a reviewable contract/proposal for the optional subsequent core implementation.

## Validation

- Reviewed against `docs/cli/gateway/restart-and-supervision.md` and the current Gateway `system.info` protocol schema.

### DIFF
--- a/docs/gateway/external-supervisor-controls.md
+++ b/docs/gateway/external-supervisor-controls.md
@@ -1,0 +1,90 @@
+---
+summary: "Draft contract for optional, launcher-owned Control UI settings"
+title: "External supervisor controls (draft)"
+---
+
+# External supervisor controls (draft)
+
+> **Draft proposal — no Gateway or Control UI implementation is included.**
+>
+> This document defines the narrow extension point proposed for external Gateway
+> supervisors. It must not be treated as an existing API or security boundary.
+
+Some deployments own Gateway lifecycle outside OpenClaw: a desktop launcher,
+a container orchestrator, or an enterprise service manager. Those supervisors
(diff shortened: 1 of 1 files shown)


MythiliMur:
For recommended action in this PR, here's the sequence to follow:

The current core draft is premature: it proposes a new bridge without a demonstrated gap in the existing plugin extension
model. The right sequence is:

1. Keep the Windows packaging PR as the concrete owner of the setting and eventual isolation boundary.
2. Prototype the integration as a plugin—backend RPC plus any supported Control UI extension surface.
3. Identify a specific missing capability only if the plugin cannot safely:
    - expose requested vs. effective state,
    - mutate the package-owned setting under appropriate authorization,
    - render the Control UI setting,
    - communicate “takes effect on next manual Gateway restart.”
4. Only then propose the smallest core extension necessary, backed by that implementation.

MythiliMur:
Like this solution better - https://github.com/openclaw/openclaw-windows-packaging/pull/28

PR #1700, opened 2026-01-25
feat(telegram): add linkPreview config option

zerone0x:
## Summary
Fixes #1675

Adds a `channels.telegram.linkPreview` config option to control whether link previews are shown in outbound messages. Setting it to `false` disables URL previews using Telegram's `link_preview_options.is_disabled` API parameter.

## Changes
- Add `linkPreview?: boolean` to `TelegramAccountConfig` type
- Add Zod schema validation for the new config option
- Pass `link_preview_options` to `sendMessage` API calls when `linkPreview: false`
- Propagate `linkPreview` config through `deliverReplies` and all callers
- Add unit tests for link preview behavior

## Usage
```json
{
  "channels": {
    "telegram": {
      "linkPreview": false
    }
  }
}
```

## Test plan
- [x] Build passes (`pnpm build`)
- [x] Lint passes (`pnpm lint`)
- [x] Unit tests pass with new test cases for linkPreview behavior
- [ ] Manual testing: verify link previews are suppressed with `linkPreview: false`

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #144454, opened 2026-09-10
feat(crabbox): add a lease-backed sandbox backend for tool-call isolation

steipete:
## What Problem This Solves

OpenClaw users can select Crabbox for cloud-hosted tool-call sandboxes. A sandbox must keep the same remote workspace across concurrent creation and session reuse, survive interrupted provisioning, and remain manageable after a restart. Provider execution and cleanup must also respect repository ownership when an operator transfers a lease.

## Why This Change Was Made

The Crabbox plugin implements the existing sandbox backend contract. Core reserves a durable runtime ID before provider allocation, serializes lifecycle changes, and retains uncertain or failed operations for recovery. Shared reservations always replay from the original provider workspace while each caller keeps its own workspace context.

Static SSH and Crabbox now share one remote-shell workspace, skills, execution-staging, and filesystem implementation. Initial uploads publish the complete primary/agent workspace atomically; partial transfers cannot be mistaken for completed workspaces, and concurrent publishers preserve the winning workspace and its later edits. Crabbox supplies a local `exec` command for each operation, retaining provider claim authority through execution and transport cleanup. OpenClaw never caches or executes raw provider SSH credentials. Management uses repository-scoped stop. Core independently fences retired runtime generations through actual process admission.

## User Impact

Set `agents.defaults.sandbox.backend` to `crabbox` and enable/configure the bundled Crabbox plugin. Remote edits survive lease reuse and registry/plugin reopen; recreate releases the old generation and creates a new one. Unsupported command/provider capabilities fail before allocation.

The initial supported route is direct Daytona. This requires a Crabbox build containing claim-owned `exec`, offline `exec --check`, and `stop --current-repo`; the released 0.56.0 binary does not provide the complete contract. The prerequisite cleanup/TTL fix landed in https://github.com/openclaw/crabbox/pull/2111. The command-ownership follow-up is merged in https://github.com/openclaw/crabbox/pull/2119 (`5c3aec438fa8368263b3465d0376003ff00402b4`).

## Evidence

The final bootstrap uses the existing native no-replace rename helper, preserving even an emptied winner. Its exact-stage cleanup restores owner directory permissions without following symlinks. The seven-case product Python proof passed on real non-root Daytona Linux (UID 1001) and macOS (UID 501), covering empty winners, readonly cleanup, symlink preservation, unsupported primitives, and uploaded modes. New SSH workspaces require native atomic directory publication; existing workspaces remain adopted without reseeding.

- Real Daytona registered-plugin test passed in 207.980 seconds: failed pre-submission recovery, concurrent creation, a shared caller from another repository, actual exec runtime, remote file operations, access after the configured 60-second token lifetime, registry/plugin reopen, recreate, and complete cleanup.
- A deliberate transfer between two synthetic repositories occurred after execution preparation and before actual spawn. The old repository's prepared command, file access, and scoped cleanup were rejected; its registry record remained recoverable. The new owner executed successfully and released the lease, after which the old registry entry could be removed.
- The exact final e23770c3 candidate with the merged Daytona-only Crabbox command contract passed in 136.474 seconds, including the same ownership-transfer checks and full cleanup.
- Real filesystem regressions fail on incomplete initial uploads before the fix and pass for primary/agent retry, concurrent publication preserving user edits, and existing SSH workspace adoption. Normal failures clean only their own staging directory; abrupt process loss can leave an unreferenced named stage on static SSH, documented for manual recovery. Crabbox lease deletion removes such artifacts.
- Regression coverage includes reservation replay from the original workspace, terminal-ID retirement, delayed execution after core removal, and upload admission after revocation. The final focused suite passed 190 tests across 13 files, covering core, plugin, SSH, remote-shell, and process-runtime behavior.
- The final cleanup-token correction uses private owner-held records, rejects foreign/duplicate finalization, and removes the inherited assertion. Its 95 focused tests, lint, actual assertion ratchet, and fresh independent P0–P2 review pass.
- Final independent review of the complete candidate is clean at P0–P2. Selected local verification covers types, lint, formatting, SDK/import boundaries, documentation, and dead-code checks; [exact-head GitHub CI passed](https://github.com/openclaw/openclaw/actions/runs/34651306176).

Live coverage is Daytona on Linux from macOS. The test reopens the real registry and plugin lifecycle; it does not claim a full Gateway process restart. This PR does not publish Crabbox or change a live operator Gateway.


steipete:
Reviewed head `0a852dd8bd223179ee4ea1fb523d581748cfb4fc`. This needs repair before landing. The plugin boundary and reuse of the existing SSH/filesystem backend are appropriate, but the current adapter does not preserve the promised workspace lifecycle.

1. **P1 — Resume retained workspaces.** `selectLease` skips every non-ready lease at `crabbox-sandbox-backend.ts:226–229`. The dependency's fixed warmup explicitly resumes native stopped/archived machines. Those states differ from a Crabbox-released lease ID. Skipping them allocates another remote-canonical workspace and strands edits on the existing machine. Preserve the ID and let Crabbox replay it; replace only after a confirmed terminal/released outcome.
2. **P2 — Select and publish one runtime per scope.** At `crabbox-sandbox-backend.ts:232–234`, simultaneous first use can mint different IDs from the same empty registry snapshot. The context owner reads IDs before awaiting the factory and publishes only afterward (`src/agents/sandbox/context.ts:291–320`). Registry uniqueness is by runtime name, so both leases can be registered for one scope.
3. **P2 — Preserve uncertain provisioning for recovery.** Warmup and its subsequent inspect run before the factory's cleanup block (`crabbox-sandbox-backend.ts:354–364`). A timeout after creation leaves an unregistered ID, and the next attempt creates a replacement. Cleanup failures are also discarded. Crabbox deliberately persists uncertain fixed-create attempts for replay under the same ID. A broad match on nested text such as “not found” is not proof that the original allocation is terminal.
4. **P2 — Respect the credential owner's lifetime.** The ten-minute endpoint cache (`crabbox-sandbox-backend.ts:46, 370–377`) assumes a default token lifetime. Daytona supports shorter configured lifetimes, and its current SSH command output does not expose authoritative expiry metadata. Obtain current access through Crabbox for the operation rather than treating that default as an expiry contract.

The retained SSH handle also bypasses renewed repository admission during its cache window. Inspection reports native readiness, not current repository authorization. Preserve Crabbox's admission at remote I/O; the existing positive live proof does not establish this boundary. No live revocation or cross-repository probe was performed for this review.

Dependency evidence was inspected at the merged Crabbox change [#1700](https://github.com/openclaw/crabbox/pull/1700), commit `376ca43032c1bbbc9f879f9394b9789688a35f58`: [fixed replay and resume](https://github.com/openclaw/crabbox/blob/376ca43032c1bbbc9f879f9394b9789688a35f58/internal/providers/daytona/fixed.go#L224), [released versus native status](https://github.com/openclaw/crabbox/blob/376ca43032c1bbbc9f879f9394b9789688a35f58/internal/providers/daytona/backend_run.go#L249), [SSH admission](https://github.com/openclaw/crabbox/blob/376ca43032c1bbbc9f879f9394b9789688a35f58/internal/providers/daytona/backend_ssh.go#L175), and [configurable SSH lifetime](https://github.com/openclaw/crabbox/blob/376ca43032c1bbbc9f879f9394b9789688a35f58/internal/providers/daytona/backend_ssh.go#L676).

The exact-head `openclaw/ci-gate` is successful ([run](https://github.com/openclaw/openclaw/actions/runs/34538326960)). That evidence does not cover these lifecycle failures. The existing tests explicitly expect new IDs for repeated unregistered calls, omit concurrent first use and post-create failure recovery, and swallow execution errors in the endpoint-refresh cases. The PR's earlier live proof remains useful for basic provisioning and remote file I/O, but it is not a fresh execution result from this review.

### Proposed storage decision — awaiting maintainer acceptance

Extend the existing sandbox registry owner to reserve one runtime generation for a backend/scope before provider provisioning, retain that ID through uncertain failures and restart, and publish readiness only for the still-current generation. Crabbox remains the sole provider/claim owner; the plugin consumes the generic registry lifecycle rather than adding a separate state store. Use the existing SQLite table and synchronous transaction owner; do not hold a transaction across provider calls. Coordinate create, retry, recreate, and prune so deletion cannot race a pending generation into publication.

The retained reservation is canonical recovery state, not a cache. Keep it visible to recovery and cleanup until the matching provider release is confirmed. This is a material persistence/publication change even if no schema-version bump is needed, so implementation needs acceptance under the [storage review checkpoint](https://github.com/openclaw/openclaw/blob/0a852dd8bd223179ee4ea1fb523d581748cfb4fc/docs/reference/database-schemas/storage-changes.md#review-checkpoint-for-material-changes).

A process-local map would not survive a Gateway restart. Deterministic single-use IDs alone cannot handle repeated recreation without retaining generation state or walking accumulated tombstones. The registry already owns runtime identity and cleanup, making it the appropriate place for this invariant.

Keep the change bounded to reservation/recovery for this backend and its generic host contract, with no new configuration, dedicated database, retention policy, or data migration. Preserve existing registered runtimes during adoption. Before rollback, disable the new backend and reconcile its recorded pending leases; do not erase uncertain recovery records. Validation must cover concurrent first use, failure after provider creation followed by restart, resume with remote edits preserved, recreation/pruning during provisioning, and successful use with a shortened credential lifetime. Normal review and exact-head CI remain required before merge.

steipete:
Maintainer assessment for head `e23770c3ff0479c0233cf6c73f3c54621c82c178`: the current review reports no code/security findings and sufficient live proof. The exact-head CI run is green: https://github.com/openclaw/openclaw/actions/runs/34651306176.

I am accepting the documented compatibility tradeoffs for this requested landing:

- New SSH workspace initialization requires a native no-replace directory operation. This prevents a concurrent initializer from overwriting even an emptied canonical workspace. The exact product helpers passed on non-root Linux and macOS; unsupported capability fails without publishing, and existing unmarked workspaces are adopted without reseeding. This prerequisite is explicit in the SSH and backend documentation.
- Crabbox is opt-in and requires the command contract merged in https://github.com/openclaw/crabbox/pull/2119. Published v0.56.0 lacks it and is rejected by an offline check before allocation. This does not change the default backend or claim that a new Crabbox release was published.
- The accepted storage decision uses the existing table and preserves uncertain leases. Real management proof covers pending recovery, preserved cleanup failure, successful reconciliation, registry/plugin reopen, and terminal cleanup; all synthetic leases were released. A full published-updater upgrade and disable/reconcile/rollback campaign was not run and is not claimed. That broader release-validation evidence is deferred, rather than inventing an additional merge gate without a demonstrated defect. The recorded rollback contract remains: disable the backend and reconcile its leases before downgrading; never erase pending recovery records.

The completed proof is proportional to this opt-in backend and its shared owner changes: 190 focused tests, 95 tests for the final cleanup-owner correction, selected type/lint/SDK/docs/dead-code gates, independent P0–P2 reviews, exact-head CI, and the final registered-backend Daytona run (136.474 seconds). Proceeding through the repository's native prepare/merge gates under the requested maintainer scope.

steipete:
Merged via squash.

- Prepared head SHA: [e23770c3ff0479c0233cf6c73f3c54621c82c178](https://github.com/openclaw/openclaw/pull/144454/commits/e23770c3ff0479c0233cf6c73f3c54621c82c178)
- Landed commit: [8a32b5a4b54396daf72e3006ebed581785a2ed85](https://github.com/openclaw/openclaw/commit/8a32b5a4b54396daf72e3006ebed581785a2ed85)

PR #144457, opened 2026-09-10
refactor(google): share media understanding metadata

steipete:
## What Problem This Solves

Google's lazy media-understanding registration and loaded provider repeat the same capability, priority, document-input and model-default facts. Audio/video request fallbacks separately repeat their advertised defaults.

## Why This Change Was Made

Move those runtime facts into the existing lightweight Google metadata owner and migrate both registrations and request defaults together. The factory gives each registration independent nested metadata while preserving the explicit empty image hooks used to clear earlier handlers. The static manifest continues to serve its separate discovery contract before runtime loading.

## User Impact

No intended behavior change: advertised capabilities, model defaults, request construction and deferred handler loading stay consistent. The typed shared factory and explicit per-capability defaults add two production lines overall while removing the duplicated descriptors and private constants.

## Evidence

The same five focused cases pass before and after: actual media registration, independent metadata, deferred handler evaluation, audio/video default requests, video payload construction and the image-registration sibling. Five other cases in the existing video file were deliberately outside this selection. All 25 selected changed-file checks pass, including types, lint, boundaries and import-cycle checks. Independent review found no actionable issue.

The registration test calls the real plugin entry point and retains the real loaded provider's metadata; its outbound handlers are mocked. This is preservation coverage, with no live-provider or full-suite claim. The local five-case runs and 25 selected checks passed without retries.

CI run 34539632419 initially failed during dependency setup: Matrix's native-library download returned HTTP 500, followed by an error in that dependency's postinstall handler. The plugin contract tests never started in that job. The exact asset now returns HTTP 200 to an anonymous HEAD check, and the single failed-job retry passed on the unchanged PR head, including the plugin-contract test step. No dependency or source change was made.

### DIFF
--- a/extensions/google/generation-provider-metadata.ts
+++ b/extensions/google/generation-provider-metadata.ts
@@ -1,5 +1,5 @@
-// Google provider module implements model/runtime integration.
 import type { ImageGenerationProvider } from "openclaw/plugin-sdk/image-generation";
+import type { MediaUnderstandingProvider } from "openclaw/plugin-sdk/media-understanding";
 import type { MusicGenerationProvider } from "openclaw/plugin-sdk/music-generation";
 import { isProviderApiKeyConfigured } from "openclaw/plugin-sdk/provider-auth";
 import type {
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [022dc37d8c21246be9d9c8832ed916b01ef3a0f8](https://github.com/openclaw/openclaw/pull/144457/commits/022dc37d8c21246be9d9c8832ed916b01ef3a0f8)
- Landed commit: [c89f3681428a93b6bc023b4053a5f0aca12ad0bc](https://github.com/openclaw/openclaw/commit/c89f3681428a93b6bc023b4053a5f0aca12ad0bc)

PR #144466, opened 2026-09-10
fix(ci): keep catalog UI recordings opt-in

steipete:
## What Problem This Solves

Resolves a problem where the Agents and automation catalog scenarios recorded videos and success screenshots during ordinary CI even when proof capture was disabled.

## Why This Change Was Made

Apply the existing proof-capture opt-in to both catalog scenarios. Their automatic assertion-failure screenshot/JSON capture and unconditional publication/Gateway logs retain their existing ownership and cleanup order. After releasing delayed catalog replies, use a read on the same Gateway connection and a rendered frame before the stale-reply assertions, so optional screenshot latency is not the synchronization mechanism.

## User Impact

Ordinary runs avoid two recordings and nine success screenshots. Explicit proof runs retain their stage filenames, artifact owners, viewports and recording options. This changes test work only; all existing UI assertions, real Gateway operations, persisted CLI readback and test cases remain.

## Evidence

After a fresh canonical CI artifact build at the source base:

- All seven existing real-Gateway cases passed with capture disabled. Their retained directories contained publication/Gateway logs and no success PNGs or videos.
- Both catalog cases passed with capture enabled, producing all nine valid PNGs at the original viewport sizes and two finalized VP8 videos.
- All seven existing failure-capture/readiness helper tests passed with capture disabled.
- Selected type, boundary, format, i18n, dead-export and lint checks passed. The only initial gate failure was the Promise executor's ignored return; a block-bodied executor fixed lint without changing its callback or behavior.
- Independent P2 review of the final diff found no actionable issues. Exact-head CI remains required before landing.

No whole-job timing improvement is claimed from these functional runs. Runtime implementation delta is zero; the test changes remove optional artifact work and make reply ordering explicit.

### DIFF
--- a/ui/src/e2e/agent-file-lifecycle.real-gateway.e2e.test.ts
+++ b/ui/src/e2e/agent-file-lifecycle.real-gateway.e2e.test.ts
@@ -14,6 +14,7 @@ import {
   createOpenClawTestInstance,
   type OpenClawTestInstance,
 } from "../../../test/helpers/openclaw-test-instance.ts";
+import type { GatewayBrowserClient } from "../api/gateway.ts";
 import { waitForControlUiGatewayReady } from "../test-helpers/control-ui-e2e-readiness.ts";
 import { pickerValue } from "../test-helpers/select-picker-e2e.ts";
 import {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [689dac912a9a7603a5374bc39117c5ffeb916ce8](https://github.com/openclaw/openclaw/pull/144466/commits/689dac912a9a7603a5374bc39117c5ffeb916ce8)
- Landed commit: [51203095e644ab319ce80f9d984f2c1e21b86382](https://github.com/openclaw/openclaw/commit/51203095e644ab319ce80f9d984f2c1e21b86382)

PR #144472, opened 2026-09-10
refactor: share pending plugin claim test setup

steipete:
Related: #139428

## What Problem This Solves

Three plugin-claim lifecycle tests repeat the same pending-claim setup, obscuring the distinct admission and mutation behavior each case verifies.

## Why This Change Was Made

Share the setup in one local helper with three required binding values. Keep promise creation inside the original async mock callback and resolve the current closure value only when the returned function is called. Dispatch, admissions, mutations, waits, assertions and cleanup remain in each case.

## User Impact

No user-visible behavior change. This removes 35 test lines without changing production code or lifecycle coverage.

## Evidence

The full lifecycle entrypoint and owner-settlement sibling passed the same 89 cases before and after, with identical full names and per-file order. Parser-based expansion of all three helper calls matches the complete original test structure. The required changed gate and fresh independent review passed.

### DIFF
--- a/src/auto-reply/reply/dispatch-from-config.lifecycle-and-bindings.test-utils.ts
+++ b/src/auto-reply/reply/dispatch-from-config.lifecycle-and-bindings.test-utils.ts
@@ -54,6 +54,43 @@ import { withDispatchProcessedOutcomeSink } from "./dispatch-processed-outcome.j
 import { finalizeInboundContextForSdk } from "./inbound-context.js";
 import { buildTestCtx } from "./test-ctx.js";
 
+function mockPendingPluginClaim(params: {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [220cf983bc4d75f87b2e5d9290cc751f3acddaa3](https://github.com/openclaw/openclaw/pull/144472/commits/220cf983bc4d75f87b2e5d9290cc751f3acddaa3)
- Landed commit: [2f194fdb6b31b0ce7b07df698f1bd755d7da0745](https://github.com/openclaw/openclaw/commit/2f194fdb6b31b0ce7b07df698f1bd755d7da0745)

PR #144474, opened 2026-09-10
fix(cli): validate the selected model during config writes

steipete:
Related: #130706. Builds on #144429.

## What Problem This Solves

Fixes an issue where CLI config validation can resolve an authored model alias twice and validate a different model. With `entry → middle → final`, a cold primary can materialize `final` at 4,096 tokens instead of `middle` at 32,000. A matching warm fallback can call its runtime normalizer before applying the manifest alias.

## Why This Change Was Made

Derive primary and fallback references from metadata captured during read-only runtime admission, then materialize the selected model without another input-alias pass. The existing helpers receive that snapshot explicitly, including bare defaults repaired by #144429. The read-only acquisition facade reuses the existing lease-options contract while preserving ephemeral ownership, live defaults, cancellation and release. Production delta: **+19 lines**.

## User Impact

Config writes validate the intended model while preserving the authored reference. Cold and warm validation use the same static-before-runtime normalization order, and rejected writes leave config bytes unchanged.

## Evidence

Production/runtime validation is bound to `8216e47759774f63abb1f8ff284bce3cbb12417c` against the actual #144429 merge, `cc748884201771eeac74dc6d8cf0c5b758bade76`.

- Matching baseline: three intended failures and 20 passing controls. Candidate: 201 tests across 10 files, full pinned changed-file gate and clean independent P2 review.
- One normal full build passed with natural metadata bound to the committed source.
- All 21 compiled CLI cases passed using 21 ordinary validation commands and 10 fixture installs. Eighteen accepted writes preserved authored refs; three rejected writes left config bytes identical.
- Covered cold aliases/exact IDs, bare defaults, batch enable/disable, native API ownership, runtime-only and foreign controls, and actual warm primary/fallback commands. The matching warm fallback receives `middle` before its runtime hook and materializes `middle-runtime` at 64,000 tokens. Native cases retain configured `pin-customer` while the admitted `ollama` hook runs.
- All 31 children joined without timeout. All 14,639 build outputs and source remained unchanged; 3,387 executed files verified. Successful-case state was removed after exit.

The first compiled matrix passed 19/21 cases but lacked native hook observations because its fixture did not declare the configured logical namespace. That failed attempt remains recorded. The final matrix corrected only external manifest ownership and strengthened namespace assertions; source and build stayed unchanged. This proves compiled CLI/config-write and provider-hook behavior, not hosted vendor inference.

Current test-only follow-up: `7a4d91c8a9a990777de0c1dfa6f8c6efe0d58420`. CI found a missed adapter in an existing isolated-probe fixture: it still treated the new options object as a positional AbortSignal. The fixture now reads `options.abortSignal` and expects the exact `{ abortSignal, catalogMode: "static" }` call, preserving all cancellation, isolation, release and queue-cleanup assertions. The local baseline reproduced all three failures; the correction passed 61 affected/sibling tests, the full changed-file gate and fresh P2 review. Production and package inputs are identical; the original 8216 build and 21-case/31-child compiled proof are explicitly carried after byte verification. No build was repeated or restamped. Fresh CI is required for this updated head.

### DIFF
--- a/src/agents/embedded-agent-runner/compact.delegate-resources.test.ts
+++ b/src/agents/embedded-agent-runner/compact.delegate-resources.test.ts
@@ -120,8 +120,7 @@ vi.mock("../prepared-model-runtime.js", async (importOriginal) => {
                   { provider: providerId, modelId: "model", agentId: "main" },
                 ],
               },
-              options?.abortSignal,
-              "static",
+              { abortSignal: options?.abortSignal, catalogMode: "static" },
             );
       try {
         expect(current.registrations.length).toBe(1);
--- a/src/agents/embedded-agent-runner/compact.foreground-resources.test.ts
+++ b/src/agents/embedded-agent-runner/compact.foreground-resources.test.ts
@@ -47,8 +47,7 @@ vi.mock("../prepared-model-runtime.js", async (importOriginal) => {
           loadRuntimePlugins: true,
           runtimePluginSelections: [{ provider, modelId: "model", agentId: "main" }],
         },
-        options?.abortSignal,
-        "static",
+        { abortSignal: options?.abortSignal, catalogMode: "static" },
       );
     },
   };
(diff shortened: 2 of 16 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7a4d91c8a9a990777de0c1dfa6f8c6efe0d58420](https://github.com/openclaw/openclaw/pull/144474/commits/7a4d91c8a9a990777de0c1dfa6f8c6efe0d58420)
- Landed commit: [773b9dff1ade17d7d303ce05a710d9b5e12c14dc](https://github.com/openclaw/openclaw/commit/773b9dff1ade17d7d303ce05a710d9b5e12c14dc)

PR #144492, opened 2026-09-10
test(auth): share catalog fixture provider runtime

steipete:
Related: #139428

## What Problem This Solves

Model-discovery auth tests supplied synthetic providers to discovery while runtime hooks could still load a separate plugin registry. That mismatch made the Chutes case take over a minute and hit its existing timeout under full-suite load; stubbing individual hooks merely moved the cost into another case.

## Why This Change Was Made

Use the same fixture provider collection for discovery and the existing real hook-runtime constructor. The fixture adapter filters registered providers and preserves owner eligibility; capability checks, formatting, failure-policy dispatch, OAuth fencing, persistence, and catalog behavior remain real. No prepared-runtime ownership is manufactured, and all existing test bodies and hooks remain unchanged.

## User Impact

No production behavior changes. This makes the auth-order fixture consistent and adds 46 test lines while retaining all 24 cases, including synthetic providers without refresh capability and late catalog finalization.

## Evidence

- A controlled local full-file comparison on Node 24.19.0 passed the same 24 ordered cases: 75.80 seconds before and 18.85 seconds after, with no shifted slow case. This is a file-level observation, not a full-suite speedup claim.
- Temporary identity guards confirmed the intended synthetic OAuth/fetch callbacks, with no denials. Those guards were removed and production source restored exactly.
- Normal guard-free validation passed 124 cases: the auth-order file, provider-runtime integration suite, and Chutes implicit-provider suite.
- The final mapped changed-file gate and fresh independent review passed. Lint-required braces and a fresh Object.assign copy were applied before those final checks.

### DIFF
--- a/test/openai-model-discovery-auth-order.test.ts
+++ b/test/openai-model-discovery-auth-order.test.ts
@@ -40,15 +40,61 @@ vi.mock("../src/plugins/provider-discovery.runtime.js", () => ({
   resolvePluginDiscoveryProvidersRuntime: () => discovery.providers,
 }));
 
-function withCatalogProviders<T>(run: () => T): T {
+vi.mock("../src/plugins/provider-hook-runtime.js", async () => {
+  const { createProviderHookRuntime } =
+    await import("../src/plugins/provider-hook-runtime-core.js");
+  const { matchesProviderPluginRef } = await import("../src/plugins/provider-registry-shared.js");
+  const selectProviders = (params: {
+    onlyPluginIds?: string[];
+    providerRefs?: readonly string[];
+  }) =>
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ec19fc18e478bd3918a90490a5743c1d302cdd32](https://github.com/openclaw/openclaw/pull/144492/commits/ec19fc18e478bd3918a90490a5743c1d302cdd32)
- Landed commit: [de750f081de1aeb4c258b714d9fac224c01ec101](https://github.com/openclaw/openclaw/commit/de750f081de1aeb4c258b714d9fac224c01ec101)

PR #144497, opened 2026-09-10
fix(usage): clear timeline ranges after session replacement

steipete:
## What Problem This Solves

Fixes an issue where Usage hid a recreated session's messages when the previous session had a selected timeline range. The old timestamps continued filtering the replacement's later replies. A one-point timeline offered no Reset button; a two-point timeline could show “Turns 1–0.” An unfinished old drag could also restore those bounds after replacement.

## Why This Change Was Made

Clear the range through the existing timeline detail owner's invalidation and reject range writes tied to superseded rendered data. Explicit clearing uses that same owner, removing the redundant page wrapper. Logical session selection, other filters and same-instance range retention remain intact.

## User Impact

Replacement messages stay visible, and users can select a fresh range without closing and reopening the session. Refreshing the same session keeps its selected range. The Usage guide describes the behavior.

## Evidence

Before the fix, four mounted cases and two real Gateway/browser cases fail for the intended stale-range behavior. Afterward, all 86 owner/sibling tests and four complete browser cases pass, along with 20 selected changed-file checks and both independent review passes.

The browser flow completes an actual drag before a one-point replacement and keeps a drag active across a two-point replacement. It creates later replacement timestamps through the existing transcript writer, waits for the real overview to adopt that instance, and verifies the current conversation without reconnecting or clearing session selection. The two-point case also selects and resets a new range successfully. Mounted controls retain ranges through manual/automatic refresh with stable or absent session IDs.

The four inspected synthetic captures below show before/after for the one-point session (300 tokens, one reply), then the two-point session (600 tokens, two replies). The before images show empty filtered conversations; the after images show the replacement replies. This uses a task-owned loopback Gateway with isolated state and silent private logging. Other captures and diagnostics are excluded, and the final complete browser run disables capture of unrelated legacy scenarios. Fixture dates/version labels are not release-version proof. No live provider or operator Gateway was used.

Before one point: the replacement is selected but its only message is hidden by the old range

After one point: the old range is cleared and the replacement message is visible

Before two points: an old drag leaves an empty turn range and hides both replacement messages

After two points: old drag writes are rejected and both replacement messages are visible

### DIFF
--- a/docs/web/control-ui/feature-reference.md
+++ b/docs/web/control-ui/feature-reference.md
@@ -93,7 +93,7 @@ Control UI capabilities grouped by area, each with the Gateway RPC methods behin
     - A provider usage failure does not block the session/cost dashboard; unavailable provider cards show their own error state.
     - Incomplete session/cost totals stay readable while the visible, focused page checks for updates. Automatic checks are bounded; if they pause, select **Refresh** to check again.
     - **Refresh** also reloads the selected session's timeline, conversation, and system-prompt breakdown.
-    - If a selected session is deleted and recreated, its new details replace the old ones without clearing your selection. Context details from a different session instance show an error and can be retried with **Refresh**.
+    - If a selected session is deleted and recreated, its new details replace the old ones and clear the previous timeline interval without clearing your session selection. An unfinished drag on the old timeline cannot change the new interval. Refreshing the same instance retains its selected interval. Context details from a different session instance show an error and can be retried with **Refresh**.
     - The overview loads session summaries first. Full system-prompt breakdowns load when you select a session; the `has:context` filter still works before opening details.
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [fc4ff68a47ff43ea48f6ba5a2ed1dd0c1ba5d052](https://github.com/openclaw/openclaw/pull/144497/commits/fc4ff68a47ff43ea48f6ba5a2ed1dd0c1ba5d052)
- Landed commit: [1e89c7fd30d2e922c68fd8e296e16b4a3ec46c9e](https://github.com/openclaw/openclaw/commit/1e89c7fd30d2e922c68fd8e296e16b4a3ec46c9e)

PR #144499, opened 2026-09-10
feat(doctor): report a runtime built from a different commit than HEAD

RomneyDa:
## What Problem This Solves

`openclaw doctor` did not report when a source checkout and its built `dist/` came from different commits. That is exactly the state behind stale-runtime failures after source moves without a matching rebuild/restart.

## Why This Change Was Made

The check is deliberately advisory and conservative:

1. it is silent while `OPENCLAW_UPDATE_IN_PROGRESS` is set
2. it emits a warning, so it cannot change Doctor's exit code
3. it is silent without both checkout and build provenance
4. packaged installs and current builds do not produce phantom drift

It reads `.git/HEAD` and `dist/build-info.json` only. There is no database, config, schema, migration, repair, or persistent-data write; the ClawSweeper data-model compatibility prompt is therefore not applicable.

## User Impact

Doctor reports that the loaded runtime is older than its source and recommends rebuilding/restarting. This improves diagnosis; it does not prevent a live Gateway from retaining an old module graph.

## Evidence

- `src/commands/doctor-stale-runtime-build.test.ts`: 7 passed
- `src/flows/doctor-stale-runtime-build-contribution.test.ts`: 1 passed
- `node scripts/check-changed.mjs`: passed on the rebased patch
- migrated test fixtures to the repository temp-directory tracker; no new cleanup warnings
- rebased cleanly onto current main

AI-assisted with Claude Code and Codex.

### DIFF
--- a/src/commands/doctor-stale-runtime-build.test.ts
+++ b/src/commands/doctor-stale-runtime-build.test.ts
@@ -1,0 +1,80 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import { collectStaleRuntimeBuildFindings } from "./doctor-stale-runtime-build.js";
+
+const BUILT = "1623683f478b1e4a2e3d5632585f006ea142e08c";
+const HEAD = "5034f2ab174b5c76b0f2a7703ddb332572916e02";
+const roots: string[] = [];
+
+async function makeCheckout(options: { built?: string; head?: string }): Promise<string> {
+  // macOS reports /var while prod resolvers return /private/var; canonicalize first.
+  const root = await fs.realpath(await fs.mkdtemp(path.join(os.tmpdir(), "openclaw-stale-build-")));
+  roots.push(root);
+  if (options.built) {
+    await fs.mkdir(path.join(root, "dist"), { recursive: true });
+    await fs.writeFile(
+      path.join(root, "dist", "build-info.json"),
+      JSON.stringify({ version: "2026.8.1", commit: options.built }),
+    );
+  }
+  if (options.head) {
+    await fs.mkdir(path.join(root, ".git"), { recursive: true });
+    await fs.writeFile(path.join(root, ".git", "HEAD"), `${options.head}\n`);
+  }
+  return root;
+}
+
+afterEach(async () => {
+  while (roots.length > 0) {
+    const root = roots.pop();
+    if (root) {
+      await fs.rm(root, { recursive: true, force: true });
+    }
+  }
+});
+
+describe("collectStaleRuntimeBuildFindings", () => {
+  it("warns when the built commit differs from the checkout commit", async () => {
+    const root = await makeCheckout({ built: BUILT, head: HEAD });
+
+    const findings = await collectStaleRuntimeBuildFindings({ root, env: {} });
+
+    expect(findings).toHaveLength(1);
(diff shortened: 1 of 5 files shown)


RomneyDa:
Both P2s were correct and are fixed in `88dcca04e4`. The first one meant the check never fired for the case it exists to report, so thank you for catching it.

**[P2] Execute the check during ordinary Doctor runs — confirmed.** Verified the path myself rather than taking it on trust: `runStructuredHealthRepairs` returns at `doctor-health-contribution-core.ts:49` when `shouldRepair` is false, and structured selection otherwise only runs via `runDoctorLintCli` under `--lint`. So `healthCheckIds` plus an empty `run` meant a plain `openclaw doctor` executed nothing.

That is precisely the invocation this change exists for. The install that motivated it ran plain `openclaw doctor`, got "Doctor complete", and kept running a six-day-old build — with the empty callback this PR would have left that unchanged.

The contribution now calls `runCoreContributionHealth(ctx, ["core/doctor/stale-runtime-build"])`, matching `runTelegramGeneralTopicConversationHealth`. That runner sets `dryRun = !ctx.prompter.shouldRepair` and reports `result.findings` in dry-run, so the warning appears in ordinary runs while `--lint` and `--fix` keep owning it through `healthCheckIds`.

Covered by `src/flows/doctor-stale-runtime-build-contribution.test.ts`, which invokes the contribution with `shouldRepair: false` and asserts the core runner is called with the check id. Verified it fails against the previous empty callback:

```
× stale runtime build contribution > runs the core check when doctor is not repairing
  → expected "vi.fn()" to be called 1 times, but got 0 times
```

**[P2] Exclude build-identity overrides — confirmed, and it cuts both ways.** `resolveCommitHash` prefers `GIT_COMMIT`/`GIT_SHA`, which declare the *built* identity for packaged installs. Honouring them for the checkout side compares the build against itself (hiding real drift) or against an unrelated commit (inventing it). The checkout identity now comes from a copy of the env with both keys removed, so it resolves from `.git/HEAD`. The `OPENCLAW_UPDATE_IN_PROGRESS` guard still reads the original env and is unaffected.

Both directions are covered, as requested:

- `ignores a GIT_COMMIT override that still names the built commit` — drift is present and must still be reported.
- `ignores an unrelated GIT_SHA override on a current build` — a matching build must stay silent.

**On the proof note.** Correct that the earlier evidence did not demonstrate the warning through an ordinary Doctor invocation — which is exactly what the first finding was about. The non-repair path is now covered at the contribution boundary by the test above. A full end-to-end `openclaw doctor` capture would need a built dist on a deliberately stale checkout; the contribution-level test pins the wiring that was actually broken.

```
node scripts/run-vitest.mjs src/commands/doctor-stale-runtime-build.test.ts   Tests  7 passed (7)
node scripts/run-vitest.mjs src/flows                                        Tests  528 passed (528)
pnpm tsgo / pnpm check:test-types                                            exit 0
node scripts/check-changed.mjs                                               exit 0 (all lanes green)
```

The `check:test-types` lane also caught a typing slip in the new mock that `pnpm tsgo` does not cover; fixed before push.

RomneyDa:
@clawsweeper re-review

The requested data-model compatibility proof is not applicable here. This PR adds a read-only Doctor warning that reads `.git/HEAD` and `dist/build-info.json`; it does not touch a database, config, schema, migration, repair path, serialization format, or persistent write. The branch is rebased, focused tests and changed-file gates pass, and test fixtures now use the repository cleanup tracker.

PR #144510, opened 2026-09-10
test(ui): reuse Full Access fixture provisioning

steipete:
## What Problem This Solves

The default and explicit Full Access delegation tests each repeated private fixture provisioning and several bootstrap CLI commands. Measured acquisition took about 24 seconds per case, including 17–18 seconds of bootstrap work; the browser interaction and delegated turn took about 2.4 seconds.

## Why This Change Was Made

Give those two cases a shared suite-owned prepared fixture. Between cases, the existing replacement owner stops and joins the Gateway, restores logging through the real config CLI with an expected-current-value check, and starts a fresh Gateway process. Each case retains its own browser context and session, verifies the initial config, performs a real delegated change, and checks both persisted state and rendered results.

The approval-policy case keeps its separate fixture. Failed or cancelled cases close shared admission before another case can use it. Suite logs retain the shared lifetime; verdicts and optional media keep their per-case directories. Selecting the explicit Full Access case alone also works.

## User Impact

Test-only setup reuse. All three cases and their original permission, approval, event, persistence and UI assertions remain. Production code and CI sharding are unchanged.

## Evidence

- The full three-case file passed in **88.58 seconds**, versus **111.01 seconds** before the change. This whole-wrapper measurement includes setup, replacement, reset and teardown; it does not compare shortened case bodies with setup moved into hooks.
- The two Full Access lifetimes took 37.15 seconds combined, versus 59.55 seconds before. The replacement/reset cost of 8.41 seconds is included.
- Explicit Full Access selected alone passed, with a fresh fixture and no dependency on the default case.
- Native teardown reported confirmed process settlement. Temporary timing instrumentation was removed and all three instrumented runtime files restored byte-for-byte.
- Final changed-file checks and independent P2 review passed after relocating the one-file patch onto current main.

The local comparison used the same built runtime at its recorded source base. The relevant fixture and command owners remained unchanged when relocating the patch; exact-head CI will build the current runtime and run the complete real-Gateway lane before landing. No whole-CI or cross-platform speedup is claimed from this comparison.

### DIFF
--- a/extensions/qa-lab/src/control-ui-openclaw-delegation.real-gateway.e2e.test.ts
+++ b/extensions/qa-lab/src/control-ui-openclaw-delegation.real-gateway.e2e.test.ts
@@ -351,173 +351,235 @@ suite.define(() => {
       throw new AggregateError(errors, "native OpenClaw approval proof failed");
     }
   }, 180_000);
+});
 
-  it.each(["default", "full"] as const)(
-    "%s Full Access saves a delegated config change without a system-agent approval",
-    { timeout: 180_000 },
-    async (mode) => {
-      const owner = createQaLiveLaneGateway();
-      const proofDir = suite.artifactDir;
-      const errors: unknown[] = [];
-      try {
-        const repoRoot = process.cwd();
-        const runtime = await owner.start({
-          repoRoot,
-          // The isolated HOME must not make the development launcher rebuild the checkout.
-          command: {
-            executablePath: process.execPath,
-            argsPrefix: [path.join(repoRoot, "openclaw.mjs")],
-            cwd: repoRoot,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [bfafe58cb144d452377ee9222a5e2c04920e62b9](https://github.com/openclaw/openclaw/pull/144510/commits/bfafe58cb144d452377ee9222a5e2c04920e62b9)
- Landed commit: [7686246e2f55a0e7922eb2716cbd59f3f365b4b3](https://github.com/openclaw/openclaw/commit/7686246e2f55a0e7922eb2716cbd59f3f365b4b3)

PR #144517, opened 2026-09-11
refactor(tests): simplify keyed FIFO fixture bookkeeping

steipete:
## What Problem This Solves

The keyed-FIFO tests use one fixed global registry key but track it in a Set through a forwarding factory.

## Why This Change Was Made

Call the existing registry factory directly with that key and remove the redundant tracker and same-symbol alias. Teardown still awaits the full-close drain before deleting the owned key.

## User Impact

No runtime or public API change. All seven keyed-FIFO cases and 19 assertions remain unchanged. Production LOC: 0; tests: +9/-20 (net -11). This does not add stronger failure cleanup.

## Evidence

- Fresh baseline and candidate: 18/18 passed, with 11 singleton cases followed by seven keyed-FIFO cases in both runs.
- All 59 frozen source/context pins verified; exact assertion text and order retained.
- Independent review found no actionable findings through P2; all 20 required local changed checks passed.

### DIFF
--- a/src/shared/keyed-fifo-lease.test.ts
+++ b/src/shared/keyed-fifo-lease.test.ts
@@ -3,25 +3,16 @@ import { afterEach, describe, expect, it } from "vitest";
 import { drainGlobalSingletonLifecycleState } from "./global-singleton.js";
 import { createKeyedFifoLeaseRegistry } from "./keyed-fifo-lease.js";
 
-const keysToDelete = new Set<symbol>();
 const TEST_KEY = Symbol.for("openclaw.test.keyedFifoLease");
-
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [420b6b60797eb4a9e68240634c720bf24f0e40cb](https://github.com/openclaw/openclaw/pull/144517/commits/420b6b60797eb4a9e68240634c720bf24f0e40cb)
- Landed commit: [782a994014549592446f281ecb0bd9c32242ef77](https://github.com/openclaw/openclaw/commit/782a994014549592446f281ecb0bd9c32242ef77)

PR #144520, opened 2026-09-11
fix(web-search): enforce the sandboxed provider trust filter for web_search

### LINKED ISSUE #144164: Bug: web_search never forwards sandboxed to provider resolution, so the sandboxed trust filter is a no-op
## Summary
`web_search` provider resolution never forwards the `sandboxed` flag it is given, so the sandboxed-provider trust filter is a no-op for search. A sandboxed agent calling `web_search` can be served by an arbitrary workspace or external plugin provider, while the same agent calling `web_fetch` is correctly restricted to bundled or verified-official providers.

## Environment
- Commit: 67e99bc838172adace381edf3def23919814a436 (origin/main at time of filing)
- Surface: web_search provider resolution

## Steps to reproduce
1. Put a workspace-origin plugin at `<workspaceDir>/.openclaw/extensions/untrusted-web/` whose `openclaw.plugin.json` declares both contracts, so the same plugin is a candidate for search and for fetch:

```json
{
  "id": "untrusted-web",
  "contracts": {
    "webSearchProviders": ["untrusted"],
    "webFetchProviders": ["untrusted"]
  }
}
```

2. Drive the two real exported resolvers with the exact argument objects the two runtimes build for a sandboxed agent, on identical config and identical workspace:

```ts
resolvePluginWebSearchProviders({ config, workspaceDir });                    // what web-search builds
resolvePluginWebFetchProviders({ config, workspaceDir, sandboxed: true });    // what web-fetch builds
```

3. Compare the returned provider lists.

## Expected
For a sandboxed agent the trust filter excludes workspace and external providers from `web_search`, exactly as it does for `web_fetch`. The stated policy lives in the filter's own comment: sandboxed web tools may run bundled providers or a verified official install, never an arbitrary workspace or external plugin with the same contract.

## Actual
The filter never engages for `web_search`. The workspace-origin provider is returned to the sandboxed agent.

## Root cause
`src/web-search/runtime.ts:271-279` builds its own argument object for provider resolution and omits `sandboxed`:

```ts
  const resolveProviders = params.preferRuntimeProviders
    ? resolveRuntimeWebSearchProviders
    : resolvePluginWebSearchProviders;
  return sortPluginEntriesForAutoDetect(
    resolveProviders({
      config: params.config,
      ...(pluginId ? { onlyPluginIds: [pluginId] } : {}),
    }),
  );
```

The flag is available at that point. `ResolveWebSearchDefinitionParams` declares `sandboxed?: boolean` (`src/web-search/runtime-types.ts:17`), `src/agents/openclaw-tools.ts:222-229` passes `sandboxed: options?.sandboxed` into `createWebSearchTool`, and `src/agents/tools/web-search.ts:127` forwards it into `runWebSearch`. It is dropped at the resolution call above and nowhere else.

The two resolver wrappers do not accept the field either: `resolvePluginWebSearchProviders` (`src/plugins/web-search-providers.runtime.ts:52-62`) and `resolveRuntimeWebSearchProviders` (`:69-76`) have no `sandboxed` in their parameter types, unlike `src/plugins/web-fetch-providers.runtime.ts:42`.

So the filter at `src/plugins/web-provider-resolution-shared.ts:96-101` is only ever reached with `params.sandboxed` undefined on the search path:

```ts
        // Sandboxed web tools may run bundled providers or a verified official install,
        // never an arbitrary workspace or external plugin with the same contract.
        (!params.sandboxed ||
          plugin.origin === "bundled" ||
          plugin.trustedOfficialInstall === true) &&
```

The shared machinery underneath already honors the flag end to end (`src/plugins/web-provider-runtime-shared.ts:26`, `:93`, `:107`), so nothing else is missing. Only the search wrapper never passes it.

Scope note, deliberately narrow: this is a trust-boundary policy that does not take effect for one tool. It is not a demonstrated sandbox escape or code-execution bypass. The provider still has to be installed and enabled in config to be resolved at all. What is lost is the origin restriction the policy states.

## Sibling surfaces
`web_fetch` threads the flag correctly at `src/web-fetch/runtime.ts:212-219`, and that call shape is pinned by an assertion at `src/web-fetch/runtime.test.ts:491`. `src/web-search/runtime.test.ts` contains zero occurrences of `sandboxed`. This looks like a one-sided fix: the sibling got the guard and the search path did not.

Suggested fix, since this is an issue and not a patch: thread `sandboxed` from `src/web-search/runtime.ts:271-279` into both resolver signatures the same way `src/web-fetch/runtime.ts:212-219` does, and add the mirror of the `src/web-fetch/runtime.test.ts:491` assertion.

## Real behavior proof
Behavior addressed: for a sandboxed agent, a workspace-origin plugin provider is filtered out of web_fetch resolution but survives web_search resolution.
Real environment tested: real exported provider resolvers driven in-process over a real on-disk workspace plugin, at commit c973fe7e721905ba2318994f1f47ae97677cf9ac (`src/web-search/runtime.ts` is byte-identical to 67e99bc838172adace381edf3def23919814a436)
Exact steps or command run after this patch: `node --import ./scripts/tsx.mjs probe.ts`, calling `resolvePluginWebSearchProviders({config, workspaceDir})` and `resolvePluginWebFetchProviders({config, workspaceDir, sandboxed: true})` on the same config and the same workspace containing the `untrusted-web` plugin above
Evidence after fix:
```
# OBSERVED (buggy, origin/main c973fe7e72)
web_search shape  resolvePluginWebSearchProviders({config})           : ["brave/brave","codex/codex","duckduckgo/duckduckgo","exa/exa","firecrawl/firecrawl","firecrawl/firecrawl-free","google/gemini","xai/grok","moonshot/kimi","minimax/minimax","ollama/ollama","parallel/parallel","parallel/parallel-free","perplexity/perplexity","qa-lab/qa-lab-search","searxng/searxng","tavily/tavily","untrusted-web/untrusted"]

web_fetch  shape  resolvePluginWebFetchProviders({config,sandboxed})  : ["firecrawl/firecrawl"]

# counterfactual, same search resolver once sandboxed is threaded through
web_search shape + sandboxed threaded                                 : ["brave/brave","codex/codex","duckduckgo/duckduckgo","exa/exa","firecrawl/firecrawl","firecrawl/firecrawl-free","google/gemini","xai/grok","moonshot/kimi","minimax/minimax","ollama/ollama","parallel/parallel","parallel/parallel-free","perplexity/perplexity","qa-lab/qa-lab-search","searxng/searxng","tavily/tavily"]

# control, the same plugin is visible to web_fetch when sandboxed is absent
web_fetch  shape, no sandboxed (control)                              : ["firecrawl/firecrawl","untrusted-web/untrusted"]
```
Observed result after fix: `untrusted-web/untrusted` is dropped by the fetch shape and by the counterfactual search shape, and retained only by the search shape as main builds it, so the single differing input is whether `sandboxed` reaches the resolver.
What was not tested: no live sandboxed agent session was run and no network search was issued, so the downstream consequence of the untrusted provider actually answering a query is inferred from resolution output rather than observed; the `preferRuntimeProviders` branch through `resolveRuntimeWebSearchProviders` was read but not driven; the workspace plugin used here is a minimal local one rather than a published external install, so the `trustedOfficialInstall` arm of the filter was not exercised.

Finn763:
## What Problem This Solves

`web_search` provider resolution dropped the `sandboxed` flag, so the sandboxed provider trust filter was a no-op for search: a sandboxed agent calling `web_search` could be served by a workspace-origin or external plugin provider, while the same agent calling `web_fetch` was correctly restricted.

Root cause, at the shared resolution throat:

- `src/web-search/runtime.ts` built its own resolver argument object and omitted `sandboxed`, (`preferRuntimeProviders` defaults to `true` in `runWebSearch`, so the live lazy runtime path ran unfiltered and the flag never reached the shared origin filter on either resolver path).

- `src/plugins/web-search-providers.runtime.ts` re-declared both resolver parameter types by hand and neither declared `sandboxed`, so the flag could not reach `resolvePluginWebProviders` (which already honored it end to end) or the origin filter in `web-provider-resolution-shared.ts`. The sibling `resolveRuntimeWebFetchProviders` had the same hand-written-parameter hole.

Fix:

- `resolvePluginWebSearchProviders` / `resolveRuntimeWebSearchProviders` / `resolvePluginWebFetchProviders` / `resolveRuntimeWebFetchProviders` now take the shared `ResolvePluginWebProvidersParams` / `ResolveRuntimeWebProvidersParams` types declared once in `web-provider-runtime-shared.ts`. `sandboxed` (and every other field) is owned by one handoff type instead of four hand-written copies, so it can no longer be silently dropped from one contract's path.

- `web_search` resolution now fails closed for sandboxed agents: `sandboxed` is forwarded on both the runtime and plugin resolver paths (lazy runtime resolution preserved), so the shared trust filter applies whichever path resolves, and untrusted providers are never eligible. This mirrors `web_fetch`. A sandboxed agent with no eligible provider now gets an actionable error naming the trust restriction and the recovery (select a bundled or explicitly trusted search provider).

- The flag is now forwarded into both bundled-artifact handoffs (`resolveBundledPublicArtifactProviders`, `resolveBundledRuntimeArtifactProviders`) so it survives every handoff, including the lazy bundled re-resolution.

Behavior for non-sandboxed agents is unchanged. The trust filter in `web-provider-resolution-shared.ts` is unchanged; it now receives the flag on the search path.

`Fixes #144164`. The closing keyword is intentionally not used anywhere else in this body.

## Evidence

Regression test with the issue's fixture: a real on-disk workspace plugin at `<workspaceDir>/.openclaw/extensions/untrusted-web/` whose `openclaw.plugin.json` declares both contracts:

```json

{

  "id": "untrusted-web",

  "contracts": { "webSearchProviders": ["untrusted"], "webFetchProviders": ["untrusted"] }

}

```

The plugin registers a provider on both contracts, and the same fixture is driven through both real exported resolvers:

- `resolvePluginWebSearchProviders({ config, workspaceDir, sandboxed: true })` and `resolvePluginWebFetchProviders({ config, workspaceDir, sandboxed: true })` now both return no provider (`[]`); before the fix search returned `untrusted-web:untrusted`.

- With `sandboxed: false` / omitted, both resolvers still return `untrusted-web:untrusted`, which is also the control proving the fixture really loads and registers (the assertion cannot pass vacuously).

- With the untrusted provider configured and `sandboxed: true`, resolution returns `[]` rather than falling back to some other provider: fail-closed, no silent re-selection.

`src/web-search/runtime.test.ts` pins the wiring: for `sandboxed: true` the flag is threaded through the lazy runtime resolver (`resolveRuntimeWebSearchProviders` called with `{ sandboxed: true }`, plugin resolver not called), so the shared trust filter applies without eager plugin activation and lazy resolution is preserved; for `sandboxed: false` the runtime resolver is used as before.

Red/green and sabotage: the two new wiring assertions were red before the fix (`expected "vi.fn()" to be called 1 times, but got 0 times`). Two sabotages were run and both were caught: (a) dropping `sandboxed` from the sandboxed search branch fails the runtime mirror test; (b) making the search wrapper swallow `sandboxed` fails the fixture test with the exact leak `search: ["untrusted-web:untrusted"]` while the non-sandboxed control still passes.

Sibling surfaces swept: `git grep` over every caller of the four resolvers. `src/web-fetch/runtime.ts` already threaded the flag; `resolveRuntimeWebFetchProviders` was missing it and is fixed by the shared parameter type. `src/cli/command-secret-targets.ts`, `src/flows/search-setup.ts`, `src/secrets/runtime-web-tools-fallback.runtime.ts` pass only fields that remain in the shared type. `src/cron/isolated-agent/run-delivery-trace.ts` calls `hasUsableWebSearchProvider` without a sandbox flag; that path only produces a diagnostics message, the call site has no sandbox status available, so it is left alone and reported here rather than guessed at.

## Verification

Ran in the patch worktree (Windows 11):

- `vitest run --config test/vitest/vitest.plugins.config.ts src/plugins/web-provider-sandboxed-resolution.test.ts` -> 3 passed.

- `vitest run --config test/vitest/vitest.unit-src.config.ts src/web-search/runtime.test.ts src/web-fetch/runtime.test.ts` -> 54 passed.

- `vitest run --config test/vitest/vitest.plugins.config.ts src/plugins/web-search-providers.runtime.test.ts` -> 13 passed.

- `vitest run --config test/vitest/vitest.plugins.config.ts src/plugins/web-provider-runtime-shared.test.ts src/plugins/web-provider-resolution-candidates.test.ts` -> passed.

- `src/plugins/web-fetch-providers.runtime.test.ts` fails one assertion on this Windows host before and after this change (it expects `/tmp/openclaw-home/...` while Windows resolves `D:\tmp\...`) — pre-existing, verified by stashing the source changes and re-running the same file on unmodified source.

## Real behavior proof

`src/web-search/runtime-sandboxed-execution.test.ts` drives the registered `runWebSearch` entrypoint end to end over the default lazy path (`preferRuntimeProviders: true`, the path this fix threads the flag through) against a real on-disk workspace plugin (`untrusted-web`, the issue fixture shape) whose tool `execute` appends to a sentinel file, using the existing-configuration shape (`tools.web.search.provider: "untrusted"`):

- Sandboxed: `runWebSearch({ ..., sandboxed: true })` rejects with `web_search is disabled or no provider is available.` and the sentinel file is never created -- the configured workspace provider is rejected before provider I/O, fail-closed with no silent re-selection.
- Non-sandboxed control: the same call without `sandboxed` returns `{ provider: "untrusted" }` and the sentinel records the query -- the fixture really loads, registers, and executes, so the sandboxed assertion cannot pass vacuously.
- RED: with the four source files reverted to the pre-fix revision, the sandboxed case fails (the search executes instead of rejecting) while the control still passes.

Ran in the patch worktree (Windows 11):

- `vitest run --config test/vitest/vitest.unit-src.config.ts src/web-search/runtime-sandboxed-execution.test.ts src/web-search/runtime.test.ts` -> 37 passed.
- `vitest run --config test/vitest/vitest.plugins.config.ts src/plugins/web-provider-sandboxed-resolution.test.ts` -> 3 passed.

Upgrade note: an existing sandboxed configuration that selects a third-party search provider now fails closed with `web_search is disabled or no provider is available.` instead of executing that provider. Recovery is to select a bundled or verified-official provider (the same rule sandboxed `web_fetch` already enforces per `docs/tools/web.md`); nothing migrates because the previous behavior was the policy gap itself. `trustedOfficialInstall` filtering is covered by the existing shared candidate-filter test.

---

**Resubmission note:** the earlier PR for this branch was auto-closed by `openclaw-barnacle` under the repository's 20-active-PR limit (`r: too-many-prs`) — a queue-depth rule, not a review outcome. The active queue has been pruned, and this is the identical commit resubmitted verbatim; the previous PR's comments are still there for reference.


Finn763:
@clawsweeper re-review

Real behavior proof added (head 2503c23): new `src/web-search/runtime-sandboxed-execution.test.ts` drives the registered `runWebSearch` end to end over the default lazy path against a real on-disk `untrusted-web` workspace plugin with an execute sentinel. Sandboxed rejects with `no provider available` and the sentinel is never created (rejection before provider I/O); the non-sandboxed control returns `{provider:"untrusted"}` with the sentinel written. RED verified on the pre-fix revision. PR body updated with the evidence, the corrected lazy-resolution description, and the upgrade/recovery note.

Finn763:
CI re-triggered on `db11053bc37` (empty commit, no code change). The failures on `d86914313` were `src/config/sessions/goals-operations.test.ts`, in files this diff does not touch, against a base from before main was green again.

Finn763:
Addressed the P2 finding on head `12e6a2974c2`.

**Trust-specific recovery message** — `src/web-search/runtime.ts`, the `candidates.length === 0` branch of `runWebSearch` (the diagnostic flagged at `runtime.ts:406`). When the sandbox trust filter empties the candidate set *and* a provider was configured — `providerId` argument, `tools.web.search.provider`, or configured runtime metadata — the error now names it and the recovery path:

> web_search is disabled or no provider is available. The configured search provider "untrusted" is not available to sandboxed web_search: sandboxed agents only run bundled or verified-official providers. Select a bundled or verified-official search provider to restore web_search.

A sandboxed run with nothing configured keeps the previous message verbatim, so this does not turn every empty selection into a trust claim, and the existing `"web_search is disabled or no provider is available."` assertions are unchanged.

Evidence — local Windows run, `scripts/test-projects.mts`, `--maxWorkers=1 --no-file-parallelism`:

| Check | Result |
|---|---|
| `src/web-search/runtime-sandboxed-execution.test.ts` | 3 passed. Rejection now asserted to name `"untrusted"` and say "bundled or verified-official"; new control case (sandboxed, nothing configured) asserts the plain message and an empty provider-I/O sentinel |
| `src/web-search/runtime.test.ts` | 35 passed (regression, expectations untouched) |
| RED check | reverting the `runtime.ts` change while keeping the new assertions → `AssertionError: expected 'web_search is disabled or no provider…' to contain '"untrusted"'` (1 failed \| 2 passed); restored → green |
| `node scripts/run-tsgo.mjs -p tsconfig.core.json` | clean; oxfmt/oxlint clean on both files |

The other three items are the owner decision, not code: this cutover stops sandboxed installs that configured a workspace or unverified provider from running managed web search. Maintainer acceptance of that fail-closed behaviour is still required before merge.

Finn763:
CI on head `12e6a2974c2` — both red jobs are the shared node-shard flake window currently hitting unrelated PRs, not this diff.

| job | failing test | same failure on an unrelated PR |
|---|---|---|
| `checks-node-changed-compact-large-15` | `src/agents/embedded-agent-runner/compact.foreground-resources.test.ts:410` — `AssertionError: expected 78.1214689999997 to be less than 50` (wall-clock budget inside the resource-retention cases) | `checks-node-changed-compact-large-15` on #156989 (`xydigitLybnnnn`) — `expected 67.02821899999981 to be less than 50` at the same line |
| `checks-node-changed-compact-large-31` | `src/agents/harness/context-engine-turn-maintenance.test.ts:288` — `AssertionError: expected "vi.fn()" to be called once, but got 0 times` (inside `vi.waitFor`), 4 cases, all in `durable accepted-turn maintenance handoff` | `checks-node-changed-compact-large-31` on #156989 — identical annotation, same line |

Notes:

- Neither failing file is in this diff (7 files: `src/web-search/{runtime,runtime.test,runtime-sandboxed-execution.test}.ts`, `src/plugins/web-{fetch,search}-providers.runtime.ts`, `web-provider-runtime-shared.ts`, `web-provider-sandboxed-resolution.test.ts`).
- Shard totals on the failing jobs: `1 failed | 1040 passed | 1 skipped (1042)` and `4 failed | 2229 passed (2233)` — a single measurement window each, both poll/time-budget assertions rather than behaviour assertions.
- A local isolated re-run is not diagnostic for these two files on this box: every case in `compact.foreground-resources.test.ts` hits the runner's own 120s test timeout here (`Test timed out in 120000ms`), so the evidence above is the cross-PR red + assertion fingerprint, not a local red/green pair.

No code change made. A rerun of the two failed jobs once the load window clears would be appreciated — fork contributors have no `rerun` permission.

PR #144521, opened 2026-09-11
fix(gateway): keep the display cap from re-materializing retired live-buffer bytes

### LINKED ISSUE #144401: Codex app-server display cap can emit foreign bytes into user-visible response
## Summary

A long Codex-app-server final response was persisted and delivered with unrelated control-plane-looking bytes replacing the tail of an otherwise normal `git diff` report.

## Observed

- OpenClaw 2026.9.2; Telegram direct chat; Codex app-server runtime.
- The original command stdout was clean (25,837 bytes; SHA-256 `eba820e46b8e6f8f35ab5cfb39ebd41b62e26f0e1722e6aa15b37ead1fc9137b`).
- The persisted final `AgentMessage` was 26,782 bytes and ended mid-source line, then contained unrelated Chinese text, the literal `to=collaboration.send_message`, and a JSON-like token-budget status line.
- The local session-history representation marked the message `truncated: true`, `reason: "display-cap"`.
- The corruption is already in the persisted `AgentMessage`, before the Telegram transport. Telegram chunking could not be isolated because no outbound payload spool was retained.

## Why this matters

This is not ordinary truncation: foreign bytes that resemble internal control-plane content became user-visible. It risks leaking internal runtime data and makes a user-facing report look like prompt injection.

## Evidence boundary

The archive sequence contains a clean `CommandExecution` result, then normal reasoning/file-change events, then the corrupted `AgentMessage`; there is no recorded neighbouring collaboration event. The raw streaming buffer was not retained, so the exact producer cannot be proven from the host.

## Expected behavior

On a display/output cap, truncate explicitly and safely (or emit a file/reference), without appending bytes from another stream or internal event.

Finn763:
## What Problem This Solves

A display-cap writer that truncates or replaces the shared live buffer without maintaining the snapshot scope re-materializes retired bytes into the visible tail — the shape reported in #144401, where the persisted row carries `__openclaw: { truncated: true, reason: "display-cap" }` while the visible stream still shows the dropped prefix.

Scope, stated honestly: this is **defensive hardening at a boundary**, not a reproduced runtime failure, and it does not close #144401. A real streaming run crosses the display cap, retires the head of the shared live buffer and keeps streaming without ever reaching a stale offset — the corrupted bytes in the report live only in the reporter's private session store. What the change pins is the invariant the cap depends on: the character offsets recorded in the snapshot scope are only read while the retained buffer still carries the prefix they index; otherwise the retained buffer *is* the current item's text, so the stale prefix and its padding are dropped instead of being re-emitted.

## Evidence

- Real streaming run (real Gateway + real WebSocket client + real provider SSE): 620,000 authored characters across two assistant items with the cap at 500,000 → `staleOffsets: 0` at every step; the client-visible stream and the persisted transcript are identical with and without this change.
- Regression test drives the boundary state *by construction* (not from a stream), and it is the only thing that distinguishes the guard: control runs confirmed no test could tell the earlier revision's sibling repair branch inside `capLiveAssistantText` apart from the main path, so that branch was removed. **Exactly one guard is added.**
- No in-tree producer can desync the cap the way the removed branch assumed, which is why the remaining guard sits at the input point rather than in the cap math.

---

# Keep the display cap from re-materializing retired live-buffer bytes

## Scope of this PR: defensive hardening (updated after the real-streaming proof)

This does not claim to fix a reproduced runtime failure, and it does not close #144401. A real streaming run (real Gateway + WebSocket client + provider SSE - 620,000 authored chars across two assistant items, cap 500,000) crosses the display cap, retires the head of the shared live buffer and keeps streaming, and never reaches a stale offset: `staleOffsets: 0` at every step, with the client-visible stream and the persisted transcript identical with and without this change. What the change pins is the boundary invariant the cap depends on: the character offsets recorded in the snapshot scope are only read while the retained buffer still carries the prefix they index - otherwise the retained buffer *is* the current item's text and the stale prefix and padding are dropped instead of being re-emitted.

That guard is a **defensive input point, not an identified production writer**, and the regression test drives that state by construction rather than from a stream. It is still worth pinning: any future producer that truncates or replaces the live buffer without maintaining the scope re-materializes retired bytes into the visible tail, which is the shape #144401 reported (`__openclaw: { truncated: true, reason: "display-cap" }` on the persisted row). Reachability of that exact event remains unproven - the reporter's corrupted bytes live only in their private session store.

**Exactly one guard is added.** The earlier revision's sibling repair branch inside `capLiveAssistantText` has been removed: control runs showed no test could distinguish it, and no in-tree producer can desync the cap that way (`text` and `prefix` are cut at the same absolute offset with the same surrogate adjustment) - it was an injected-only condition, exactly as flagged. The remaining guard is the one with demonstrated red -> green power.

**Reconciled with current `main`:** rebased onto `a941bb47519`; the merger was reworked in #145683 and now has an append fast path that reads no offsets. The guard sits after that fast path, so the fast path keeps its cost and every offset read (the slice and the prefix re-emission) stays validated.

## Evidence

The incident could not be replayed from a host artifact: the report states the raw streaming buffer

and outbound spool were not retained, and the corrupted row only exists in the reporter's private

session store. What is reproducible is the boundary itself, with locally constructed long

multi-byte (CJK + emoji) text:

- Before the fix, the new regression test shows the corrupted shape directly. The merged live text

  becomes content from the retired region (`审查报告与修复建议🚀🐈‍⬛🧪...`) instead of the retained

  window plus the newest delta (`代码审查报告与修复建议尾巴🙂新增结论`).

- After the fix the same test yields exactly the retained window plus the newest delta, and the

  merged length is bounded by the retained window.

## Verification

- New regression file `src/gateway/live-display-cap-boundary.test.ts` (6 tests - the five boundary cases plus a fast-path interaction case pinning main's new append path):

  - offset read after the display cap retired the head, with CJK + emoji + a split surrogate pair;

  - the retained buffer always carries the prefix its scope records, over a multi-byte boundary

    corpus, with no lone surrogate and no cap overshoot;

  - the history display cap truncates multi-byte text on a character boundary, keeps

    `__openclaw.truncated: true` with `reason: "display-cap"`, and never introduces U+FFFD;

  - an under-limit message is returned untouched.

- Red then green: with the source file reverted (new tests kept), the offset test fails with the corrupted

  string; with the fix it passes.

- Sabotage checks (applied, observed, reverted):

  1. delete the retained-prefix guard in `mergeAssistantText` -> the offset test fails again with

     the corrupted string;

  2. replace the cap's surrogate-safe slice with a raw unit slice -> the boundary test fails with

     `lone surrogate at 0`.

- Existing suite for the same path still passes: `src/gateway/server-chat.stream-text-merge.test.ts`

  plus the new file = 31 tests passed.

- Run: `node scripts/run-vitest.mjs run --config test/vitest/vitest.gateway.config.ts src/gateway/live-display-cap-boundary.test.ts src/gateway/server-chat.stream-text-merge.test.ts`.

- Lint/format: `scripts/run-oxlint.mjs` cannot run in this checkout because its declaration inputs

  resolve into a sibling worktree's `node_modules` (`Declaration input escapes checkout:

  D:\code\wt-oc-143262\node_modules\.pnpm\...`), so the packed `oxlint`/`oxfmt` binaries were run

  directly against the changed files: no diagnostics, and `oxfmt --check` reports the files already

  formatted.

## Evidence boundary

The exact producer of the reported incident is not proven from here: it requires either a stream

that trims the live buffer behind the recorded offsets, or a cut that is not character-safe in the

same truncation path. Both are closed by this change, and the regression test pins the observable

requirement: truncation stays on character boundaries, no bytes outside the retained message enter

the visible tail, and the `display-cap` metadata is preserved. If private triage produces the raw

native stream, this boundary is the first place to compare against.

---

**Resubmission note:** the earlier PR for this branch was auto-closed by `openclaw-barnacle` under the repository's 20-active-PR limit (`r: too-many-prs`) — a queue-depth rule, not a review outcome. The active queue has been pruned, and this is the identical commit resubmitted verbatim; the previous PR's comments are still there for reference.

# Real streaming entrypoint proof — openclaw PR #144521 (display cap vs. stale live-buffer offsets)

**Worktree:** `D:/code/wt-oc-144401` @ `d1ae23ef9807f0c7a0aa8a80a111b713c76bbabc` (PR #144521 head, branch `fix/oc-144401-descr`)

**"Before" revision:** `28589e76ead3876ef73c5675d0fce19f1bbf6e16` (PR parent) for the two changed source files

`src/gateway/agent-event-assistant-text.ts`, `src/gateway/live-chat-projector.ts`

**Discipline:** nothing pushed, PR untouched; the worktree was restored after the "before" run (`git status --porcelain` empty, `git diff HEAD` empty). All scripts/artifacts live outside the repo in `%LOCALAPPDATA%/Temp/ocproof-144521/`.

---

## 1. Verdict (read this first)

ClawSweeper asked for *"real streaming entrypoint reaching stale offsets and showing after-fix recovery."*

Executed result of a real stream (real gateway + real websocket client + real OpenAI-Responses provider SSE, 620,000 authored chars, two assistant items, cap 500,000):

| Question | Result |

|---|---|

| Does a real streaming run reach the display-cap boundary and keep appending (item switch, multi-byte CJK/emoji, retired head)? | **Yes** — retained live buffer saturates at 500,000 chars, published scope carries a retired prefix of 439,998 chars |

| Does the real streaming entrypoint reach the **stale-offset** condition (buffer no longer carrying the scope's recorded prefix)? | **No** — `staleOffsets: 0` at every step of a replay that reproduces the gateway's buffer byte-for-byte (hash match) |

| Are the PR's two guards exercised by the real stream? | **No** — before/after client buffers, scope bookkeeping and persisted rows are identical |

| Is the raised "second cap shortens retained text without updating offsets" producer present in a real stream? | **Not found** (consistent with the reviewer's P1 open item) |

| Does anything corrupt the persisted/visible reply at the boundary? | **No** — live chat stream and stored transcript are the correct suffix/exact authored text in both revisions |

| Is the fix still *dangerous*? | No; it is dead-so-far defensive hardening. Its only demonstrated effect needs an externally shortened retained buffer (§5) |

The honest summary: **the requested positive proof cannot be produced — the production live-buffer loop keeps text and offsets synchronized at the cap boundary.** This file records that with real execution so the PR decision can be made on facts instead of the helper-test injection.

---

## 2. How to re-run (all commands runnable as-is)

```bash

bash "$LOCALAPPDATA/Temp/ocproof-144521/run-proof.sh"

```

or step by step (this is exactly what produced the artifacts below):

```bash

cd D:/code/wt-oc-144401

OX=file:///D:/code/wt-oc-144401/scripts/tsx.mjs

OUT="$LOCALAPPDATA/Temp/ocproof-144521"



# AFTER (PR head)

node --import "$OX" "$OUT/gateway-stream-proof.mts" --tag after-head  --big-a 560000 --big-b 60000 --out "$OUT/after.json"

node --import "$OX" "$OUT/mirror.mts" "$OUT/after.json" "$OUT/after-mirror.json"



# BEFORE (PR parent for the two changed files)

git checkout 28589e76ead3876ef73c5675d0fce19f1bbf6e16 -- src/gateway/agent-event-assistant-text.ts src/gateway/live-chat-projector.ts

node --import "$OX" "$OUT/gateway-stream-proof.mts" --tag before-parent --big-a 560000 --big-b 60000 --out "$OUT/before.json"

node --import "$OX" "$OUT/mirror.mts" "$OUT/before.json" "$OUT/before-mirror.json"

git checkout HEAD -- src/gateway/agent-event-assistant-text.ts src/gateway/live-chat-projector.ts   # restore

git status --porcelain                                                                              # must print nothing



node "$OUT/compare.mjs" "$OUT/after.json" "$OUT/before.json" "$OUT/after-mirror.json" "$OUT/before-mirror.json" --out "$OUT/compare.json"

node "$OUT/inspect-store.mjs" "$LOCALAPPDATA/Temp/ocproof-144521-yrVt52" "$OUT/after.json"   # persisted transcript

```

Runtime: gateway startup 29–39 s (source mode, `--import scripts/tsx.mjs`), full run 58–85 s per revision.

---

## 3. What the real entrypoint is

`gateway-stream-proof.mts` uses the repository's own real-gateway harness and nothing else:

* **Gateway:** `startGatewayWithClient()` from `src/gateway/test-helpers.e2e.ts` → real `startGatewayServer()` on loopback, isolated state (`HOME`/`OPENCLAW_STATE_DIR` = fresh `mkdtemp`), bundled plugins/channels/cron/providers disabled.

* **Streaming entrypoint:** real websocket client (`src/gateway/client.ts`) → `sessions.messages.subscribe` → `chat.send` → `agent.wait` → `chat.history`. The client-visible stream is the real `chat` event stream (`state: delta|final`, `deltaText`, `message`).

* **Provider stream:** a local SSE server implementing the OpenAI Responses streaming protocol (same shape as `scripts/e2e/mock-openai-server.mjs`) wired in via `buildMockOpenAiResponsesProvider`; the real `openai-responses` transport parses it (`[provider-transport-fetch] ... status=200 contentType=text/event-stream` in the run log).

* **Scripted reply (identical in both revisions):**

  * provider request 1: message item `msg_proof_a` = **560,000 chars** of `代码审查报告与修复建议🚀🐈‍⬛🧪…` beginning with `HEADBYTES-144521-代码审查报告🚀`, then a `session_status` function call (`stopReason: toolUse`);

  * provider request 2 (after the tool result): message item `msg_proof_b` = **60,000 chars** ending with `PROOF-TAIL-144521-END`.

  * total authored 620,000 chars > cap 500,000; item ids observed on the `agent` stream: `<session>:1`, `<session>:2`.

`mirror.mts` replays the captured agent events (`itemId`, `delta`, cumulative `text`, `replace`/`replaceable`) through the real `mergeAssistantText` + `capLiveAssistantText`, i.e. the same two functions the gateway calls in `src/gateway/server-chat.ts:1032-1044`. It reports a **hash match against the buffer the real gateway broadcast** (`mirrorMatchesClient: true` in both revisions), so the per-step offsets it prints describe the production state.

---

### 4.1 Live chat stream crossed the cap and never resurrected retired bytes (AFTER)

`after.json → frameHeadMarkers` (client-visible `message.content[0].text`):

| seq | state | chars | contains `HEADBYTES-…` | contains `PROOF-TAIL-…` |

|---|---|---|---|---|

| 5 | delta | 16,000 | true | false |

| 29 | delta | 400,000 | true | false |

| 39 | delta | 500,000 | **false** (cap retired the authored head) | false |

| 48 | delta | 500,000 | false | **true** (second item arrived) |

| 51 | final | 500,000 | false | true |

`finalClientMessage`: 500,000 chars, hash `1496d979`, head `🚀🐈‍⬛🧪代码审查报告…`, tail `…🚀🐈PROOF-TAIL-144521-END`, `suffixOfAuthored: true`, `containsAuthoredHead: false`.

### 4.2 The offsets stayed valid across the retirement and the item switch (both revisions)

`after-mirror.json → steps` (`[seq, itemSwitch, retainedAfterCap, retiredThisStep, scopePrefixChars, prefixStillCarried]`):

```

[5,  false, 16000,  0,     0,      true]

[28, false, 384000, 0,     0,      true]

[39, false, 500000, 60000, 0,      true]   <- cap engaged on item 1

[48, true,  500000, 60002, 439998, true]   <- item switch at the cap; prefix published

mirrorRetainedChars 500000 / mirrorHash 1496d979 / mirrorMatchesClient true / staleOffsets 0

```

`before-mirror.json → steps` (same scripted reply; the runtime coalesced deltas slightly differently, 4 vs 5 `agent` assistant events — timing, not behavior):

```

[5,  false, 16000,  0,     0,      true]

[17, false, 208000, 0,     0,      true]

[37, false, 500000, 28000, 0,      true]

[39, false, 500000, 32000, 0,      true]

[48, true,  500000, 60002, 439998, true]   <- identical retirement/prefix arithmetic

mirrorRetainedChars 500000 / mirrorHash 1496d979 / mirrorMatchesClient true / staleOffsets 0

```

The step that the PR's guard targets is the last one: at the item switch the merged text was 500,000 (retained) + 2 (paragraph padding) + 60,000 (item 2) = 560,002 chars, the cap retired 60,002 and published `scope.prefix = 439,998` chars. The next appends kept slicing at `prefix.length + separatorLength` **and the retained buffer still carried that prefix** — the stale-offset condition never occurred.

### 4.3 Nothing is corrupted in the persisted rows (both revisions, identical)

`transcript_events` in the isolated agent store (node:sqlite, `inspect-store.mjs`):

| revision | assistant row | chars | hash | authored hash |

|---|---|---|---|---|

| after | item 1 (`toolUse`) | 560,000 | `8decdc9a` | `8decdc9a` |

| after | item 2 (`stop`) | 60,000 | `df96c18c` | `df96c18c` |

| before | item 1 (`toolUse`) | 560,000 | `8decdc9a` | `8decdc9a` |

| before | item 2 (`stop`) | 60,000 | `df96c18c` | `df96c18c` |

`chat.history` row (the session-history preview the incident mentioned): 8,018 chars + `...(truncated)...`, `__openclaw: { truncated: true, reason: "display-cap" }`, hash `0f4d1355` — **identical in both revisions**; the stored transcript above it is the full authored text, so the display cap bounds the live/display buffers only.

### 4.4 AFTER vs BEFORE on the real stream: identical

`compare.json → afterEqualsBefore`: `finalClientHash: true`, `finalClientChars: true`, `mirrorHash: true`. The final client buffer (500,000 chars, `1496d979`), the history row and the persisted transcript match. `mirrorStepTraces: false` only because the two runs coalesced provider deltas into 4 vs 5 assistant events (wall-clock batching); the cap/item-switch arithmetic in the traces is the same.

### 4.5 The guards' only demonstrated input is an externally shortened buffer (injected, labelled as such)

`probe2.mts` reproduces exactly the snapshot the new test file injects (retained text cut down to the visible window, scope kept), via the real functions:

| revision | merged text | resurrected retired bytes | keeps visible window + delta |

|---|---|---|---|

| before | 499,989 chars ending `…🚀🐈‍⬛🧪\n\n新增结论` | **499,970 chars** | no (`mergedTextEqualsVisiblePlusDelta: false`) |

| after | 19 chars = `代码审查报告与修复建议尾巴🙂新增结论` | 0 | yes |

This is the *injected* state; §4.2/§4.3 show the production stream does not produce it.

### 4.6 Randomized loop fuzz agrees with §4.4

`probe3.mts` (5 seeds × 40 steps; item churn across 3 items, cumulative snapshots, `replace`/`replaceable`, text that repeatedly crosses the 500,000 cap, CJK + emoji + lone-surrogate-prone alphabet): before and after outputs are byte-identical (`probe3-before.json` == `probe3-after.json`, trace + final hashes included).

---

## 5. Claims

**Proven true (real execution):**

1. A real gateway streaming entrypoint (`chat.send` → ws `chat` frames; real provider SSE) reaches the 500,000 display cap with multi-byte text, retires the head and keeps streaming (two assistant items, 620,000 authored chars).

2. After the cap retires the head, the published scope carries a real retired prefix (439,998 chars) and the next appends still slice at valid offsets — the client-visible text is the correct suffix of the authored reply and the stored transcript is byte-exact.

3. Before/after produce identical output on that real stream; the guards never fire (mirror per-step `prefixStillCarried: true`, `staleOffsets: 0`).

4. The guards do change behavior for an externally shortened retained buffer (499,970 resurrected bytes before → 0 after).

**Not observed / false so far:**

5. "The real streaming entrypoint reaches stale offsets" — not reproduced here, and no in-tree caller that hands a shortened buffer (with a live scope) to the merger was found: `run.rawBuffer` and `run.assistantScope` are written only together (`src/gateway/server-chat.ts:1032-1044`), and `capLiveAssistantText` retires prefix/boundary/separator together (`src/gateway/live-chat-projector.ts:30-39`).

6. The reported incident shape (persisted reply carrying foreign bytes) is not reproduced by this path: transcript rows equal the authored text.

**Closest untested candidate for a real producer** (code reading only, *not* executed): `src/tui/embedded-backend.ts:1552` sets `run.buffer = finalText` (the authoritative final reply, an independent string) while `run.assistantScope` survives — a late assistant delta for that run would then hit the merge with a stale scope; the same file relies on late "text_end" flushes elsewhere. Worth its own streaming proof before claiming the fix is reachable.

**Effect on ClawSweeper's checklist:** the P1 "second cap that changes retained text without updating offsets has not been identified in a production caller" is *confirmed* by execution here, and the requested "after-fix recovery" is only demonstrable for injected snapshots (§4.5). The PR is defensive hardening, not a fix for a reproduced stream transition.

---

## 6. Artifacts

| Path (all under `%LOCALAPPDATA%/Temp/ocproof-144521/`) | Content |

|---|---|

| `run-proof.sh` | after → before → restore → compare, one command |

| `gateway-stream-proof.mts` | real gateway + ws client + scripted Responses SSE; writes one artifact JSON + `.final-message.txt` (the exact client buffer) |

| `mirror.mts` | replays captured agent events through the real merge/cap; asserts hash match with the client buffer; reports per-step offsets and `staleOffsets` |

| `compare.mjs` | after/before/mirror comparison table (`compare.json`) |

| `probe2.mts` + `probe2-after.json`/`probe2-before.json` | injected shortened-buffer boundary case (the PR test's input) |

| `probe3.mts` + `probe3-after.json`/`probe3-before.json` | randomized cap-crossing loop fuzz (identical both revisions) |

| `inspect-store.mjs` | node:sqlite dump of the persisted transcript rows |

| `after.json`, `before.json` (+ `.final-message.txt`, `.log`) | full run artifacts: frames, agent events, mirrors inputs, history rows |

| `after-mirror.json`, `before-mirror.json`, `compare.json` | loop replays and the comparison |

| scratch dirs `ocproof-144521-yrVt52` (after), `ocproof-144521-WkC3Qe` (before) | isolated `HOME`/state/store of each run |

Worktree state at finish: `git status --porcelain` empty, `git diff HEAD` empty, no push, PR unchanged.

---

### Verifier note (owner re-run, 2026-09-14)

Re-ran the full chain myself after the subagent finished (both legs, fresh state): `after exit=0`, `before exit=0`; `compare` reproduced `staleOffsets:0` on both revisions, mirror hash `1496d979` matches the client buffer, `afterEqualsBefore` true for hashes; injected `probe2` reproduced `resurrectedHeadChars: 499970` (before) vs `0` and `mergedTextLength: 19` (after). Worktree restored clean. Note: the re-run script had an argument-quoting bug (`node "$NODE_IMPORT"` passed `--import file://...` as a single argv element, so the node process printed `bad option` and exits 9 while the script kept going); fixed the script to `node --import "$TSX"` before this re-run so it is genuinely re-runnable.

---

### Rework summary (revision `8d6d1eba31e`, 2026-09-14)

- Rebased onto current main (`a941bb47519`); the merger conflict is resolved by placing the guard after main's new append fast path (#145683). Diff is now 3 files, +154/-1.
- The cap-side repair branch is removed (see opening): this PR adds exactly one guard, backed by six tests.
- Evidence re-run on this revision: `live-display-cap-boundary` 6 passed; `server-chat.stream-text-merge` 26 passed; main-touched suites `openai-http` 132 passed and `openresponses-http` 142 passed; `sessions-history-http`'s 49 failures are Windows `%TEMP%` sqlite `EBUSY` environment failures, identical with main's own sources restored. `tsgo` cannot run in this worktree (borrowed `node_modules` junction -> `declaration input escapes checkout`); that is an environment defect of this checkout, unrelated to the change.


Finn763:
CI note on `control-ui-performance` — the job fails on its **baseline** run, not on this branch's measurement.

From the job log on this head (run `34546416958`, job `control-ui-performance`, step "Check Control UI performance against base"):

```
node failed (1)
[control-ui-performance-base] FAILED (exit 1)
[ELIFECYCLE] Command failed with exit code 1.
##[error]Process completed with exit code 1.
```

`control-ui-performance-base` is the run against the base ref used for comparison, so there is no measurement on this branch to compare against — the gate cannot be evaluating a regression here, it is failing to produce a baseline at all. Support for that reading: the same lane is failing across the repo in the same window, including on `main` (`1b2b06ed2fc7` — `checks-ui (3/3)`, `openclaw/ci-gate`) and on maintainer branches (`steipete/pr-view-scroll-obscured-c…` — `checks-ui-e2e (2/7)(4/7)(5/7)`; `codex/perf91-doctor-artifacts` — `checks-node-compact-small-24`).

This diff does not touch the control UI or any rendering path, so I am not changing anything for it. Happy to re-run once the lane is green if you want a fresh datapoint.

(Also for the record: the `PR context and evidence` check that failed on the earlier PR for this branch is green now — that one was the literal heading, and the body has been corrected to `## What Problem This Solves`.)

Finn763:
@clawsweeper re-review — the real streaming entrypoint was executed end-to-end (real gateway + real WS client + real provider SSE, 620k chars, two items, cap 500k). Result, recorded honestly in the PR description: the production live-buffer loop keeps text and offsets synchronized at the cap (staleOffsets:0 on both revisions, identical hashes), and the guards' effect is demonstrated under the documented injected-shortening boundary (before resurrects 499,970 chars vs after correct 19). Full before/after traces + raw artifacts attached; owner re-ran both legs fresh.

Finn763:
@clawsweeper re-review - revision 8d6d1eba31e: rebased onto current main (merger reworked in #145683 - the guard now sits after the new append fast path, so every offset read stays validated); the un-demonstrable cap-side repair branch is removed, leaving exactly one guard with red->green power; opening claims rewritten to match the negative real-stream result (defensive hardening, #144401 not closed). Owner re-ran the suites on this revision.

Finn763:
CI re-triggered on `d799e4f7f9d` (empty commit, no code change). The failures on `8d6d1eba3` were `checks-node-compact-large-15 shard`, in files this diff does not touch, against a base from before main was green again.

PR #144531, opened 2026-09-11
fix: attribute slow session list requests

steipete:
## What Problem This Solves

Operators investigating a slow session list could see its RPC duration without knowing whether the request projected a list, waited for another request, or reused a completed page. The row-projection span also combined synchronous work with time awaiting the event loop, making the next investigation target ambiguous.

## Why This Change Was Made

Add a request-owned `slow session list` warning for handlers taking at least one second when diagnostics and warning logging are enabled. The cache owner supplies the actual hit/follower/owner role and returned page size; an active shared operation carries only its producer trace binding. The handler records fixed phase totals and the existing yield sites supply aggregate synchronous/wait timings. Existing cache keys, fences, deadlines, retained results, authority checks, and yield positions are preserved.

## User Impact

Operators can distinguish their request's waiting and response work from the producer's projection. Hits and followers omit producer timing counters. Logs include process/thread and request trace identity, without query text or session contents. The logging docs explain that these are elapsed durations, including a response callback phase, rather than CPU time or client-receipt proof.

## Evidence

- Complete changed-file checks and independent autoreview passed.
- Fifty focused cache, activity, and diagnostic tests passed. New cases cover overlapping trace ownership, completed-hit laziness, controlled native yield resumption, bounded visibility repair, disabled diagnostics, and throwing sinks. The two diagnostic contract cases fail against the original production source because the records are absent while their normal list results still pass.
- A real isolated Linux Gateway passed build, session creation, repeated and filtered list reads, update readback, offline CLI persistence, restart, and cleanup through [Testbox](https://github.com/openclaw/openclaw/actions/runs/34546163934). Native and wrapper exits were zero and the lease completed. No slow warning was forced; controlled tests cover the one-second diagnostic condition.
- One fixed standalone baseline–candidate–candidate–baseline check retained 12 samples per variant, enablement, and page size. Disabled medians were 2.43→2.53 ms for four rows and 44.80→43.04 ms for 256 rows; enabled candidate medians were 2.61/45.22 ms. The larger page exercised one or two yields and emitted no warnings. This bounded check resolves an earlier comparison with unequal competing test work; it is not a general latency-improvement or zero-overhead claim.


steipete:
Merged via squash.

- Prepared head SHA: [948808b8ebaab12383fc9bc56c26ac660f02e22e](https://github.com/openclaw/openclaw/pull/144531/commits/948808b8ebaab12383fc9bc56c26ac660f02e22e)
- Landed commit: [47d039cb7045ca05c93f0fe8eee99771300daffa](https://github.com/openclaw/openclaw/commit/47d039cb7045ca05c93f0fe8eee99771300daffa)

Issue #135655, opened 2026-09-01
[Bug]: agent-initiated update stops its own managed Gateway process tree


steipete:
### Bug type
Crash (process/app exits or hangs)

### Beta release blocker
No

### Summary
Agent-initiated `openclaw update` on systemd can stop the same service cgroup that owns the updater, leaving the Gateway down; current main also has refusal paths instead of completing the update out-of-band.

### Steps to reproduce
Reported flow: start a systemd-managed Gateway, ask its agent to execute `openclaw update`, and observe the updater disappear at the managed-service stop step. The linked public reports describe two independent occurrences. A deterministic current-source regression is being prepared.

### Expected behavior
The CLI should transfer the entire update to the existing managed-service handoff before stopping its parent service, and return observable progress/recovery guidance. The Gateway update RPC already uses that owner.

### Actual behavior
The reports describe update termination between stopping the managed Gateway and updating via the package manager; the Gateway remains down. Current main's ancestry checks refuse some such invocations, which avoids self-stop but still requires an external shell.

### OpenClaw version
Reported: 2026.8.1 updating to 2026.8.2. Current-source investigation: 9c0ef3b41844ed836625060981e47a1880f38d25.

### Operating system
Linux/systemd; exact distribution NOT_ENOUGH_INFO. macOS launchd has the related parent-stop boundary.

### Install method
Package-manager update with managed Gateway service.

### Model
NOT_ENOUGH_INFO

### Provider / routing chain
NOT_ENOUGH_INFO

### Additional provider/model setup details
Not required for reproducing the CLI/service ownership boundary.

### Logs
Reported last phases: "Stopping managed gateway service before package update" then "Updating via package manager".

### Screenshots, recordings, and evidence
- https://x.com/MisterLeeHODL/status/2094867377653473789
- https://x.com/allinmaxi/status/2094883854767042831
- Related historical reports: #84068 (RPC helper cgroup escape), #85120 (in-band macOS CLI update).
- Source: `src/cli/update-cli/update-command-execution.ts`, `src/cli/update-cli/update-command-service-maintenance.ts`, `src/infra/update-managed-service-handoff.ts`.

### Impact and severity
Blocks agent-initiated updates on the reported systemd installations and can leave the Gateway unavailable. Frequency beyond these two reports: NOT_ENOUGH_INFO.

### Additional information
[...]

PR #144532, opened 2026-09-11
fix: managed Linux updates fail during service activation

jalehman:
Related: #135655

## What Problem This Solves

Fixes an issue where Linux operators running an agent-initiated managed update could pass handoff inspection but have activation refused because the updater still appeared beneath the Gateway in the Unix process tree.

## Why This Change Was Made

Linux `systemd-run --user --scope` transfers cgroup ownership without changing Unix ancestry. This change consistently accepts only the existing, fully validated managed-handoff identity at every ancestry-sensitive stop boundary, including the final pre-stop revalidation; ordinary, stale, or mismatched callers remain blocked, and non-Linux behavior is unchanged.

## User Impact

Managed Linux updates can proceed from the authorized handoff into service activation instead of stopping at a false ancestry refusal. Existing Gateway self-stop protections remain in place for every caller that cannot prove the current handoff lease and process identity.

## Evidence

- `src/cli/update-cli/update-command-service-maintenance.test.ts`: 45/45 tests passed.
- Focused Linux lifecycle, PID identity, and cgroup selection: 6 tests passed (197 filtered).
- Isolated user-systemd proof confirmed that a helper remains in its separate scope when the parent `KillMode=control-group` service stops, retains the same PID/start identity, and terminates when its own scope stops.
- `oxfmt`, `oxlint`, core typecheck, and core-test typecheck passed on the committed head.
- `node scripts/check-changed.mjs -- <changed files>` passed before the final Linux-only refinement; final exact-head formatting, lint, and type checks passed afterward.
- Autoreview completed; its platform-widening finding was resolved by limiting activation authorization to Linux and adding regression coverage.
- ClawSweeper’s macOS portability finding was resolved on head `a457dd052847`: the foreign-PID lease fixture now runs only on a native Linux host while Linux activation and macOS refusal policy cases remain covered. The focused Linux suite passes 45/45; macOS CI is pending.

A real self-update and live Gateway activation were intentionally not run; validation used isolated service fixtures and a disposable user-systemd lifecycle proof.

### DIFF
--- a/src/cli/update-cli/update-command-service-maintenance.test.ts
+++ b/src/cli/update-cli/update-command-service-maintenance.test.ts
@@ -285,7 +285,7 @@ it.each([
   }),
 );
 
-it.runIf(process.platform === "linux" || process.platform === "darwin").each([
+const servingAncestorMaintenanceCases = [
   { platform: "linux", identity: "current updater", phase: "inspect", authorized: true },
   { platform: "linux", identity: "current updater", phase: "prepare", authorized: true },
   { platform: "darwin", identity: "current updater", phase: "inspect", authorized: true },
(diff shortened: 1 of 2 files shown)


jalehman:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [4130e1c7e9a3412bfadf93df10478c6f76756ccf](https://github.com/openclaw/openclaw/pull/144532/commits/4130e1c7e9a3412bfadf93df10478c6f76756ccf)
- Landed commit: [e03cfd2aa9296339a3c85ea7da407add5178fcd2](https://github.com/openclaw/openclaw/commit/e03cfd2aa9296339a3c85ea7da407add5178fcd2)

PR #144540, opened 2026-09-11
docs(web-search): explain Perplexity time filter limits

### LINKED ISSUE #144500: web_search: document and preflight Perplexity freshness/date-range conflict
## Summary

`web_search` calls routed to the native Perplexity provider reject requests that include both `freshness` and `date_after`/`date_before`:

```text
conflicting_time_filters: freshness and date_after/date_before cannot be used together.
Use either freshness (day/week/month/year) or a date range (date_after/date_before), not both.
```

The generic docs list all three parameters, but do not clearly state this mutual exclusion for native Perplexity. A scheduled digest interpreted the resulting all-lane provider errors as a "quiet week," producing a false zero-result report.

## Reproduction

1. Configure `web_search` to use Perplexity.
2. Call with `freshness: "week"` and `date_after: "2026-09-03"` (optionally date_before).
3. Observe the provider error above.
4. Omit the explicit dates and retain `freshness: "week"`; the same query returns normal results.

## Requested improvements

- Document the mutual exclusion in `docs/tools/web.md`, ideally in the parameter caveat section and/or a provider capability table.
- Validate/reject the incompatible combination locally before issuing the provider call, with an actionable error that explicitly advises the caller to select one mode.
- Consider exposing this constraint in the generic tool schema/provider capabilities so agents can avoid the invalid call.

## Why it matters

Search automations can otherwise receive zero candidates without a tool-level failure. For scheduled editorial workflows, that creates the risk of silently reporting content scarcity when the actual problem is request validation.

zhangguiping-xydt:
Closes #144500

## What Problem This Solves

Resolves a documentation and tool-guidance gap where `web_search` users cannot see that Perplexity rejects `freshness` together with `date_after` or `date_before`.

## Why This Change Was Made

The existing Perplexity validator already returns an actionable conflict error before any network request. This change documents that mutual exclusion in the shared tool schema, the Perplexity provider schema, and both public web-search references so callers can choose either a relative freshness filter or an explicit date range.

## User Impact

Agents and operators receive clear parameter guidance before making a rejected Perplexity search request. Existing validation, error behavior, and provider support remain unchanged.

## Evidence

- `extensions/perplexity/src/perplexity-web-search-provider.runtime.ts` and `src/agents/tools/web-search-provider-common.ts` were inspected: the existing `conflicting_time_filters` validation rejects the combination before transport execution.
- `node scripts/run-vitest.mjs extensions/perplexity/src/perplexity-web-search-provider.test.ts src/agents/tools/web-search.test.ts` passed: 48 Perplexity tests and 55 shared web-search tests.
- `pnpm docs:check-mdx` passed (1,294 files).
- `pnpm docs:check-links` passed (`broken_links=0`, 13,297 internal links checked).
- `oxfmt --check` and `git diff --check` passed.

The exact-head production tool constructors were executed without mocks via `node --import ./scripts/tsx.mjs proof-144500.ts`. The captured output showed the shared tool description says `Perplexity time filters are mutually exclusive`, its freshness field says to choose freshness or a date range, and the native Perplexity freshness field says to choose it or a date range. The date fields remain their existing published-date descriptions.

Prompt snapshots were regenerated and `pnpm prompt:snapshots:check` passed. This change does not alter validation, transport, or any external API contract, so no external search service call is required.

### DIFF
--- a/docs/tools/perplexity-search.md
+++ b/docs/tools/perplexity-search.md
@@ -117,15 +117,15 @@ ISO 639-1 language code (e.g. `en`, `de`, `fr`).
 </ParamField>
 
 <ParamField path="freshness" type="'day' | 'week' | 'month' | 'year'">
-Time filter - `day` is 24 hours.
+Time filter - `day` is 24 hours. Cannot be combined with `date_after` or `date_before`.
 </ParamField>
 
 <ParamField path="date_after" type="string">
-Only results published after this date (`YYYY-MM-DD`).
+Only results published after this date (`YYYY-MM-DD`). Use a date range instead of `freshness`.
 </ParamField>
 
 <ParamField path="date_before" type="string">
-Only results published before this date (`YYYY-MM-DD`).
+Only results published before this date (`YYYY-MM-DD`). Use a date range instead of `freshness`.
 </ParamField>
 
 <ParamField path="domain_filter" type="string[]">
--- a/docs/tools/web.md
+++ b/docs/tools/web.md
@@ -471,20 +471,20 @@ provider, OpenClaw does not show the `x_search` prompt.
 
 ## Tool parameters
 
-| Parameter             | Description                                                        |
-| --------------------- | ------------------------------------------------------------------ |
-| `query`               | Search query (required)                                            |
-| `count`               | Results to return (1-10, default: 5)                               |
-| `country`             | 2-letter ISO country code (e.g. "US", "DE")                        |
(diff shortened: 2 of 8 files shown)


zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

zhangguiping-xydt:
@clawsweeper re-review

PR #144543, opened 2026-09-11
fix(agents): honor provider backoff during stale recovery

### LINKED ISSUE #144424: Concurrent heartbeat lanes on unrelated dashboard sessions collide, trip real Anthropic 429s, and the retry backoff isn't honored (self-sustaining storm)
### Observed Behavior
Over two days (2026-09-09, 2026-09-10) our `heartbeat-main` automation repeatedly burned several consecutive 10-minute execution timeouts in a short window (02:00 on 9/9; 11:03-12:04 and 20:13-22:24 on 9/10, six consecutive failed runs in the last one). The 20:13-22:24 window had a pinnable mechanism:

1. Two sessions were each running their own recurring `:heartbeat` lane concurrently: the normal `agent:<agent>:main:heartbeat` lane, and an unrelated dashboard (visible child) session's own `agent:<agent>:dashboard:<uuid>:heartbeat` lane, tied to a background task that session had been running earlier. This part is expected behavior: `resolveHeartbeatSessionSelection` legitimately derives an isolated heartbeat key for any targeted session so a completion follow-up routes back to the session that owns it.
2. When both lanes' model calls landed close together, we got real Anthropic 429s (`rate_limit_error: This request would exceed your account's rate limit`).
3. The recovery path turned that single 429 into a repeating storm for the full 2h11m window instead of one clean backoff.

### Root cause, confirmed against source
Traced against commit `1391f7cd2d40ab5bbcf2f5f831d3a64f520e72d7` (matches our installed v2026.9.3).

After a 429, `maybeRetryTransient` in `failover-retry-controller.ts` correctly computes a multi-hour backoff and sleeps through it in chunks via `sleepWithAbort`. Nothing records that this sleep is an intentional, bounded wait, though. The "last progress" clock (`churnProgress.lastProgressAt`, frozen at the "model_call:ended" event when the 429 landed) just keeps aging while the sleep runs.

`resolveRunStaleThresholdMs` in `diagnostic-run-activity-snapshot.ts` gives a quiet tool call its own protected deadline (`activeToolDeadlineAtMs`), but there's no equivalent for a quiet retry-backoff sleep. With no active tool or model call in flight, `activeWorkKind` is `embedded_run`, which falls through to the flat 5-minute default (`STUCK_SESSION_PROGRESS_STALE_MS` / `RUN_STALE_TAKEOVER_MS`). Five minutes of "no progress" and `diagnostic-stuck-session-recovery.runtime.ts` concludes the run is stuck and reclaims it via `abortAndDrainEmbeddedAgentRun`, which cancels the backend and trips the same abort signal `sleepWithAbort` is waiting on, ending the backoff early.

That matches our logs exactly: 429 at 20:09:15, reclaimed at 20:15:41 (about 6.5 minutes later, consistent with the 5-minute floor plus the roughly 30-second diagnostic check cadence), fresh 429 two seconds after that.

The reclaim also retires the pending followup/wake that started the turn (the stale-takeover path in `reply-run-registry.operation.ts`), which is what lets a fresh attempt fire almost immediately instead of the session just going quiet. That fresh attempt hits the same still-active rate limit, computes a fresh multi-hour backoff, sleeps, and gets reclaimed again five minutes later. Repeat for two hours.

So the actual bug is narrower than a rate-limit or retry-logic problem: the stale-progress watchdog can't tell "genuinely wedged" apart from "deliberately waiting out a known backoff," because the retry sleep never tells it which one is happening.

### Recommended fix
1. Minimal: inside `maybeRetryTransient`'s sleep loop, call the existing progress-recording function (`markReplyRunDiagnosticProgress`, already used elsewhere in the reply-run registry) once per chunk, tagged with something like `reason: "transient_retry_backoff"`. That resets the staleness clock the same way ordinary activity already does, without touching any timeout constants or watchdog logic.
2. More thorough: give the retry sleep its own protected deadline the way tool calls already get `activeToolDeadlineAtMs`, so `resolveRunStaleThresholdMs` covers the whole backoff instead of falling back to the flat 5-minute default. More invasive, since it means the retry controller and the diagnostic snapshot need a way to communicate that deadline.

Either one lets the sleep run out undisturbed, so a rate-limited heartbeat gets one real retry after its actual computed delay instead of colliding every six minutes for hours.

### Evidence (journalctl excerpts, redacted paths)
```
20:09:11 [diagnostic] reply run stale takeover retained: sessionKey=agent:main:main:heartbeat reason=stuck_recovery owner=awaiting_terminal_completion backend=pending
20:09:11 [diagnostic] lane task error: lane=session:agent:main:main:heartbeat durationMs=370142 error="aborted | Reply operation expired as stale"
20:09:13 [diagnostic] stuck session recovery: sessionId=2e497d3c-... sessionKey=agent:main:main:heartbeat age=370s action=abort_embedded_run aborted=true drained=true released=0
20:09:15 [provider-transport-fetch] [model-fetch] response provider=anthropic api=anthropic-messages model=claude-sonnet-5 status=429 elapsedMs=1262 dispatcher=new
20:09:15 [provider-transport-fetch] [model-fetch] response provider=anthropic api=anthropic-messages model=claude-sonnet-5 status=429 elapsedMs=1017 dispatcher=reused
20:09:15 [agent/embedded] embedded run agent end: ... error=rate_limit_error ...
20:09:15 [agent/embedded] transient same-model retry 1/9 for anthropic/claude-sonnet-5 reason=rate_limit: delayMs=9856000

# ~6 minutes later, on the SAME lane, without waiting anywhere near delayMs=9856000:
20:15:41 [diagnostic] lane task error: lane=session:agent:main:main:heartbeat durationMs=387903 error="aborted | Reply operation expired as stale"
20:15:42 [diagnostic] stuck session recovery: sessionId=... sessionKey=agent:main:main:heartbeat age=387s action=abort_embedded_run
20:15:43 [provider-transport-fetch] [model-fetch] response ... status=429 ...
20:15:43 [agent/embedded] transient same-model retry 1/9 ... delayMs=9856000

# Meanwhile a separate dashboard session runs its own heartbeat lane concurrently:
20:11:41 [diagnostic] stalled session: sessionId=613f5280-... sessionKey=agent:main:dashboard:2b79f545-...:heartbeat state=processing age=147s reason=active_work_without_progress activeWorkKind=embedded_run
```
(request_ids of the 429s available on request, omitted here since they're account-identifying)

We don't have a direct synthetic reproduction, this is the log trail from a live incident plus a source-level trace of the mechanism. Happy to help verify against a synthetic 429 if that's useful.

### Related
Possibly related to (or an adjacent bug to) #14440 ("Heartbeat polls firing every ~1 minute instead of configured interval"), closed as stale. That report's proposed fix does appear to be implemented in our installed `heartbeat-runner-*.mjs` (`shouldDeferWake`'s `event` intent now gates on `now < agent.nextDueMs`), so this is a different, narrower issue: the retry-backoff-sleep-vs-stale-recovery interaction described above.

### Environment
- OpenClaw 2026.9.3 (1391f7c)
- Node v26.8.1
- Linux (CachyOS/x86_64)
- `agents.defaults.heartbeat`: `{ every: "1h", target: "none", isolatedSession: true }`
- Model: `anthropic/claude-sonnet-5`

obviyus:
Closes #144424.

## What Problem This Solves

Fixes an issue where heartbeats waiting through a provider's retry minimum were cancelled as stale. A new run could then contact the provider too early and repeat the failure.

## Why This Change Was Made

Keep the bounded retry wait with its current run owner and recheck that deadline before stale recovery or takeover. Release it on cancellation or completion, and record actual resumed activity before preparing the next attempt. Normal retry budgets, timeout rules and shutdown behavior stay intact.

## User Impact

Long provider backoff can finish under the original run. Manual cancellation, a new visible turn, shutdown and execution deadlines can still stop it. Genuine stalled work still recovers.

## Evidence

Public `system event` → enabled heartbeat → real HTTP 429 with `Retry-After: 660`, using an isolated Gateway and a synthetic compatible endpoint:

| Observation | Baseline | Repaired |
| --- | --- | --- |
| Next provider request after the 429 | 387.742 seconds | 660.126 seconds |
| Original run | Cancelled as stale | Retried and completed once |
| Separately queued heartbeat | Ran after premature cancellation | Ran once after original completion |

- Independent behavior acceptance covers the public run, clean transcripts, one transient retry indicator, manual abort, visible replacement, normal shutdown and the configured timeout backstop.
- Real elapsed-time controls preserve genuine stalled-run recovery and ordinary ten-minute takeover. A matched baseline comparison preserves deferred sibling heartbeat delivery without changing scheduling policy.
- Successful configured model fallback, retry budgets, profile rotation and jitter boundaries are covered. Native validation passed 270 focused tests, lint, both ratchets and the normal commit hook.
- The long-wait fixture uses a 20-minute heartbeat timeout to isolate stale recovery. Existing execution deadlines and shutdown grace still apply. No real provider account or channel credential was used.

Thanks @corvid-reads for the report and source trace.

### DIFF
--- a/src/agents/agent-command.compaction-rotation.test.ts
+++ b/src/agents/agent-command.compaction-rotation.test.ts
@@ -362,6 +362,7 @@ describe("agentCommand compaction transcript rotation", () => {
           activeWriterRunId: params.runId,
         }));
         params.deferredLifecycle?.adopt({
+          beginRetryWait: () => undefined,
           discard: () => {},
           complete: async () => {
             expect(loadSessionEntry({ sessionKey, storePath })).toMatchObject({
--- a/src/agents/command/attempt-execution.cli.test.ts
+++ b/src/agents/command/attempt-execution.cli.test.ts
@@ -3598,6 +3598,7 @@ describe("CLI attempt execution", () => {
     const handoffToCli = vi.fn();
     const deferredLifecycle: NonNullable<RunAgentAttemptParams["deferredLifecycle"]> = {
       signal: controller.signal,
+      beginRetryWait: () => undefined,
       abort: vi.fn(),
       adopt: vi.fn(),
       handoffToCli,
--- a/src/agents/command/attempt-execution.ts
+++ b/src/agents/command/attempt-execution.ts
@@ -1401,6 +1401,7 @@ export function runAgentAttempt(params: {
     deferTerminalLifecycle: params.deferTerminalLifecycle,
     onDeferredLifecycleOwner: params.deferredLifecycle?.adopt,
     onDeferredLifecycleAbort: params.deferredLifecycle?.abort,
+    onRetryWait: params.deferredLifecycle?.beginRetryWait,
     suppressNextUserMessagePersistence: params.suppressPromptPersistenceOnRetry === true,
     userTurnTranscriptRecorder: params.userTurnTranscriptRecorder,
     assistantErrorTranscript: params.assistantErrorTranscript,
--- a/src/agents/embedded-agent-runner/run/attempt-stream-prepare.ts
+++ b/src/agents/embedded-agent-runner/run/attempt-stream-prepare.ts
@@ -7,6 +7,7 @@ import { normalizeOptionalString } from "@openclaw/normalization-core/string-coe
 import { isSilentReplyText, SILENT_REPLY_TOKEN } from "../../../auto-reply/tokens.js";
 import { runWithOwnedSessionTranscriptWrite } from "../../../config/sessions/transcript-write-context.js";
 import { captureAgentRunLifecycleGeneration } from "../../../infra/agent-events.js";
+import { validateAgentRunDelegatedAuthority } from "../../../infra/agent-run-registry.js";
 import {
   freezeDiagnosticTraceContext,
   type DiagnosticTraceContext,
@@ -688,6 +689,13 @@ export function prepareEmbeddedAttemptStream(input: {
     deferredLifecycleOwner = createEmbeddedAttemptDeferredLifecycleOwner({
       runId: attempt.runId,
       sessionId: attempt.sessionId,
+      diagnosticOwner: input.diagnosticOwner,
+      onRetryWaitCompleted: () => attempt.replyOperation?.recordActivity(),
+      isCurrent: () =>
+        registration?.delegatedAuthority !== undefined &&
+        validateAgentRunDelegatedAuthority(registration.delegatedAuthority) &&
+        ACTIVE_EMBEDDED_RUN_REGISTRATIONS.get(queueHandle) === registration &&
+        ACTIVE_EMBEDDED_RUNS.get(attempt.sessionId) === queueHandle,
       trajectoryRecorder: input.trajectoryRecorder ?? null,
       clearActiveRun: () =>
         clearActiveEmbeddedRun(
--- a/src/agents/embedded-agent-runner/run/deferred-lifecycle-owner.test.ts
+++ b/src/agents/embedded-agent-runner/run/deferred-lifecycle-owner.test.ts
@@ -1,6 +1,12 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { resolveReplyOperationAbortReason } from "../../../auto-reply/reply/reply-operation-abort.js";
 import {
+  createDiagnosticEmbeddedRunOwner,
+  getDiagnosticSessionActivitySnapshot,
+  markDiagnosticEmbeddedRunStarted,
+  closeDiagnosticEmbeddedRunOwner,
+} from "../../../logging/diagnostic-run-activity.js";
+import {
   isAgentRunRestartAbortReason,
   isAgentRunSupersededAbortReason,
   resolveAgentRunErrorLifecycleFields,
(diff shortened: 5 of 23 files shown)


obviyus:
Landed in fa5f8b197264c5563b30e9061278defbbd67cba5. Thanks @corvid-reads for the report and source trace!

**Proof:** A public system event triggered an enabled heartbeat against a compatible HTTP endpoint returning `429` with `Retry-After: 660`.

- **Bug:** Stale recovery cancelled intentional backoff and allowed a new request before the provider minimum.
- **Fix:** The current run owns a bounded retry deadline that recovery rechecks. Cancellation and genuine-stall recovery remain effective.

| Before | After |
| --- | --- |
| New provider request after 387.742 seconds | Original run retries after 660.126 seconds |
| Original heartbeat cancelled as stale | Original heartbeat completes once, then its queued follow-up completes once |

Manual abort, visible replacement, normal shutdown, configured timeout, fallback, and genuine-stall controls passed. The test-only fixture correction passed 184 focused tests and the normal commit hook. Fresh exact-head review had no findings, and [the full CI run passed](https://github.com/openclaw/openclaw/actions/runs/34551442864) without an exception. All 23 landed file blobs match the reviewed head. Issue #144424 is closed.

PR #144547, opened 2026-09-11
fix(ui): let canvas surfaces keep their own horizontal drag

clawd-ops:
## What Problem This Solves

On a touch device, dragging left to right inside the terminal opens the navigation drawer instead of selecting text. Selection is effectively unusable there, because the gesture that performs it is intercepted by the app shell.

The edge swipe already opts out of elements that own their own horizontal drag, but that selector lists only form and text elements:

```
a, button, input, textarea, select, pre, [role='slider'], [contenteditable]:not([contenteditable='false'])
```

A `<canvas>` matches none of them. It renders its own selection rather than using DOM text, and critically it has no *ancestor* that matches either. The hidden `<textarea>` a terminal keeps for IME is a sibling of the canvas, not a parent, so walking the composed path finds nothing to block on and the drawer claims the gesture.

The same class of bug affects any canvas-based surface where a horizontal drag is meaningful. The VNC view drags the remote pointer and is hit identically.

## Approach

Add `canvas` to the opt-out selector. The selector moves to a named module constant so it can carry a comment explaining why canvas belongs there, since that is the part a future reader is most likely to question and delete.

This deliberately means an edge swipe started on a canvas will not open the drawer. That is the intended trade: the canvas owns the gesture, and the drawer remains reachable from any other part of the content area as well as from its button.

## Evidence

New unit test, `ui/src/app/nav-drawer-swipe.runtime.test.ts`, covering both directions of the behavior so the fix cannot regress into either failure mode:

```
✓ opens the drawer for a horizontal drag over ordinary content  22ms
✓ leaves a horizontal drag over a canvas to the canvas           3ms

Test Files  1 passed (1)
     Tests  2 passed (2)
```

jsdom provides no `TouchEvent` constructor, so the test attaches the touch list to a plain `Event` and dispatches it through a real DOM tree, which keeps `composedPath()` authentic. That path is the part of `handleStart` under test.

Typecheck and lint pass on the changed files.

Originally diagnosed by reading the deployed `nav-drawer-swipe.runtime` bundle to confirm the opt-out selector was the actual cause, rather than inferring it from the symptom.

### DIFF
--- a/ui/src/app/nav-drawer-swipe.runtime.test.ts
+++ b/ui/src/app/nav-drawer-swipe.runtime.test.ts
@@ -1,0 +1,99 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { NavDrawerSwipeOwner } from "./nav-drawer-swipe.runtime.ts";
+
+type TestHost = HTMLElement & {
+  onboardingMode: boolean;
+  updateComplete: Promise<boolean>;
+  navDrawerOpen: boolean;
+};
+
+function createHost(): TestHost {
+  const host = document.createElement("div") as TestHost;
+  host.onboardingMode = false;
+  host.navDrawerOpen = false;
+  host.updateComplete = Promise.resolve(true);
+  host.innerHTML = `
+    <div class="shell-nav"></div>
+    <div class="shell-nav-backdrop"></div>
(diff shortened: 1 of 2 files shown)


clawd-ops:

PR #144548, opened 2026-09-11
fix(ui): deliver Android IME text to the terminal

clawd-ops:
## What Problem This Solves

Typing into the terminal from an Android soft keyboard does nothing. No characters appear, and there is no error. The terminal looks focused and behaves as though the keystrokes were discarded, which is exactly what happens.

`ghostty-web` 0.4.0 installs this listener:

```js
A.addEventListener("beforeinput", (E) => E.preventDefault());
```

and separately returns early for `keyCode === 229`. Android soft keyboards deliver committed text on `beforeinput` with `keyCode 229` on the preceding `keydown`. The listener cancels the browser's own DOM editing but never forwards `event.data` to a PTY write path, and the tested keyboard emitted no later data-bearing composition completion. So the text exists in the event and is then dropped.

Confirmed by live remote debugging on a physical device:

```
keydown     {keyCode: 229, target: "OPENCLAW-TERMINAL-PANEL"}
beforeinput {data: "k", inputType: "insertText", defaultPrevented: false}
composedPath: TEXTAREA > DIV > SLOT > ... > OPENCLAW-TERMINAL-PANEL
```

Upstream bug: [coder/ghostty-web#120](https://github.com/coder/ghostty-web/issues/120), "fix: Korean/CJK IME composition events not captured", open at time of writing.

## Approach

This follows the existing convention in `terminal-runtime.ts`, which already carries workarounds for three other `ghostty-web` 0.4.0 defects, each stating what breaks and when it can be removed. This is a fourth entry in that pattern rather than a new mechanism.

Two choices worth calling out:

**It does not reimplement a write path.** The committed text is handed to ghostty's own `handleCompositionEnd`, which already does `if (data && data.length > 0) onDataCallback(data)`. Ghostty's encoder keeps ownership of the spelling, including live DECCKM state.

**It is scoped to coarse-pointer devices.** On desktop the existing keydown route works and a native `compositionend` still fires, so forwarding unconditionally would deliver CJK input twice. The guard confines the workaround to the path where the defect was observed.

The terminal lives inside an open shadow root, so document-level listeners see events retargeted to the host element. The container is located via `composedPath()`, which does expose the internals, and is identified structurally as the element owning both a canvas and a textarea rather than by any incidental attribute.

Removal is gated on `coder/ghostty-web#120` being fixed and the minimum version moving past it.

## Evidence

New unit test, `ui/src/components/terminal/terminal-runtime-ime.test.ts`:

```
✓ beforeinput insertText during IME active dispatches compositionend on ghostty container
✓ beforeinput during non-IME state does nothing
✓ on non-touch device: fix is not installed
✓ insertLineBreak maps to carriage return
✓ insertParagraph also maps to carriage return
✓ deleteContentBackward maps to DEL byte (0x7f)
✓ cleanup removes listeners so subsequent beforeinput no longer forwards

Test Files  1 passed (1)
     Tests  7 passed (7)
```

The non-touch case is the one that matters most for review: it is the guard against double-delivery of CJK input on desktop, so it is asserted directly rather than assumed.

Typecheck passes on the changed files.

### DIFF
--- a/ui/src/components/terminal/terminal-runtime-ime.test.ts
+++ b/ui/src/components/terminal/terminal-runtime-ime.test.ts
@@ -1,0 +1,230 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { installTerminalImeWorkaround } from "./terminal-runtime.ts";
+
+/**
+ * Creates a ghostty-like container in the real DOM: a div holding both a
+ * canvas and a textarea (the structure ghostty-web creates inside its
+ * container). The caller provides the container element, so tests can
+ * pre-install the ghostty beforeinput bug simulation before calling
+ * installTerminalImeWorkaround.
+ */
+function createGhosttyContainer(): {
+  container: HTMLDivElement;
+  canvas: HTMLCanvasElement;
+  textarea: HTMLTextAreaElement;
+} {
+  const container = document.createElement("div");
+  const canvas = document.createElement("canvas");
+  const textarea = document.createElement("textarea");
+  container.appendChild(canvas);
+  container.appendChild(textarea);
+  document.body.appendChild(container);
+  return { container, canvas, textarea };
+}
+
+/**
+ * Installs the ghostty beforeinput bug simulation: it calls e.preventDefault()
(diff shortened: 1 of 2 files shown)


clawd-ops:

PR #144549, opened 2026-09-11
fix(ui): reserve space for the Android soft keyboard in the terminal

clawd-ops:
## What Problem This Solves

On Android, opening the soft keyboard in the terminal paints the canvas behind the keyboard. The prompt and the cursor end up underneath it, so you cannot see what you are typing.

Ghostty's `FitAddon` derives its row count from its container's `clientHeight` minus padding, and registers no window resize listener of its own. Nothing in the current code informs it that the usable area shrank.

## Anticipating the first objection

The Control UI declares:

```
interactive-widget=resizes-content
```

Under that policy the layout viewport is supposed to shrink with the keyboard, which would make `innerHeight - visualViewport.height` approximately zero and this entire change pointless. That was my own first conclusion, and it was wrong.

Measured on a physical Galaxy Z Fold6 in Chrome:

| State | innerHeight | visualViewport.height | gap | applied | canvas height |
|---|---|---|---|---|---|
| Idle, no keyboard | 676 | 676 | 0 | 0 | 611 |
| **Keyboard open** | **676** | **363** | **313** | **313px** | **247** |

`innerHeight` does not move. The layout viewport is not resized despite the declared policy, the gap is a real 313px, and the reserve is load-bearing.

This is one handset, and I would not want that overstated. Ruled out on the way: both routes serve a byte-identical document with the same viewport meta, and the bundle contains no `navigator.virtualKeyboard.overlaysContent` override, which is the one API that can legitimately shadow the meta policy. That leaves a Chrome, device, or browsing-mode discrepancy. The comment in the source records the measurement explicitly so the next reader does not delete the code on the strength of the declared policy, which is what I nearly did.

## Approach

Apply `padding-bottom` matching the keyboard inset to the ghostty container. Because `FitAddon` already subtracts padding from `clientHeight`, this is sufficient on its own and needs no synthetic resize events.

**A pinch-zoom guard is required, not optional.** At page scale 2 with no keyboard, the naive expression returns a false reserve of roughly 458px, which collapses the terminal. Keyboard avoidance is therefore suppressed at any non-unit scale. That is a deliberate trade: a scale-corrected estimate risks recreating the very collapse the guard prevents, so the conservative behavior is to do nothing while zoomed.

The implementation follows the `visualViewport` observation pattern already used in `chat-composer-dom.ts`, with one documented divergence. It reads `window.innerHeight` rather than that file's `documentElement.clientHeight || innerHeight` pairing, because the quantity needed here is specifically the one the keyboard does not move. Substituting `clientHeight` would reintroduce a dependency on the layout-viewport resize whose absence this workaround exists to handle. The reasoning is in the source comment.

Scoped to coarse-pointer devices, and fully reverted by the returned cleanup function, which is wired into the existing dispose path.

## Evidence

New unit test, `ui/src/components/terminal/terminal-keyboard-reserve.ts` covered by `terminal-keyboard-reserve.test.ts`, using the measured device numbers as fixtures rather than invented ones:

```
✓ at scale 1 with keyboard open (innerHeight 676, vv.height 363, offsetTop 0): applies 313 px padding
✓ at scale 1 with keyboard closed (vv.height = innerHeight): applies no padding
✓ at scale 2 (pinch-zoom) with open-looking gap: applies no padding (zoom guard)
✓ on non-touch device: no padding applied even with a large gap
✓ cleanup removes listeners so subsequent viewport resize has no effect

Test Files  1 passed (1)
     Tests  5 passed (5)
```

Typecheck passes on the changed files.

## Open question for reviewers

This sets `container.style.paddingBottom` directly. If a CSS custom property applied through `terminal-panel-styles.ts` is the preferred layer for this project, say so and I will move it.

### DIFF
--- a/ui/src/components/terminal/terminal-keyboard-reserve.test.ts
+++ b/ui/src/components/terminal/terminal-keyboard-reserve.test.ts
@@ -1,0 +1,182 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { observeTerminalKeyboardReserve } from "./terminal-keyboard-reserve.ts";
+
+/**
+ * Stubs window.visualViewport with controllable height, offsetTop, and scale.
+ * vitest's vi.stubGlobal replaces the property on the window object; we use
+ * Object.defineProperty so the stub is configurable and restorable.
+ */
+function stubVisualViewport(opts: { height: number; offsetTop?: number; scale?: number }): {
+  trigger: (event: "resize" | "scroll") => void;
+} {
+  const listeners: Record<string, Array<() => void>> = {};
+  const vv = {
+    height: opts.height,
+    offsetTop: opts.offsetTop ?? 0,
+    scale: opts.scale ?? 1,
+    addEventListener(type: string, fn: () => void) {
+      (listeners[type] ??= []).push(fn);
+    },
+    removeEventListener(type: string, fn: () => void) {
+      listeners[type] = (listeners[type] ?? []).filter((f) => f !== fn);
+    },
+  };
+  vi.stubGlobal("visualViewport", vv);
+  return {
+    trigger(event: "resize" | "scroll") {
+      for (const fn of listeners[event] ?? []) {
+        fn();
+      }
+    },
+  };
+}
+
+/**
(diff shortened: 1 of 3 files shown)


clawd-ops:

PR #144567, opened 2026-09-11
fix: avoid replaying older Telegram photos in Codex image requests

### LINKED ISSUE #144556: Codex Telegram photo follow-ups replay older image attachments
### Problem

In a Telegram conversation using the Codex harness, discussing a new photo can produce an answer about earlier photos. During history projection, previously saved photos and the current photo can be supplied as images in the same native user input.

### Reproduction

1. Send photo A to a Telegram bot using the Codex harness.
2. Send an unrelated photo B and ask about that photo.
3. Exercise a turn where the harness reconstructs continuity or context-engine history.
4. Inspect the native `turn/start` input: older photo pixels are attached together with the current photo.

Expected: a new photo is not automatically accompanied by earlier photo pixels. A text-only follow-up should retain the latest photo turn or album, with older attachment references available for explicit retrieval.

Observed with OpenClaw 2026.9.3 on macOS. The projection path remains present at main 5a6d0619ab11fb389d96b7ce977c7c987a97ce0a. A focused regression inspects native input images through the real media preparation host. This report is about reconstructed history, not deletion of images already stored in a resumed native thread.

yelloduxx:
Closes #144556.

## What Problem This Solves

Fixes an issue where users discussing a new Telegram photo would get answers about older photos when the Codex harness reconstructed history. Earlier and current photo pixels could become attachments in the same native user input.

## Why This Change Was Made

For Telegram projections, omit historical photo pixels when the current turn supplies images. For text-only follow-ups, retain the latest photo turn, including a complete album. Use the same canonical media classification as hydration, including path-derived image types and stickers. Retain captions and file references; mixed document/image messages and other channels keep their existing behavior.

## User Impact

New Telegram photos no longer automatically reattach unrelated earlier photos. Older references remain available for explicit retrieval. This does not erase images already present in a resumed native Codex thread, and attachment preparation still occurs before the projection filter.

## Evidence

- `pnpm check:changed` passed on the combined source checkout: production/test typechecks, changed-file formatting and lint, SDK surface guards, dead-export scans and runtime import-cycle checks. Each PR contains only its own concern.

- 57 context projection tests and 13 continuity/media tests passed on Node 24.19.0. Regressions cover albums, path-only image metadata, retained document pages, unchanged transcripts and no fallback to an older photo when the latest cannot be hydrated.
- The continuity test loads real PNG bytes through the production media host and inspects native `turn/start` image input.
- Personally inspected Codex source at commit 9e22e74e8dcab53f8bf1799c0eed1f9834c32f1a: `codex-rs/protocol/src/models.rs` combines supplied inputs into one user message. This is not a claim that Codex generates a literal collage.
- Independent review identified a filename-classification gap; it was corrected by carrying the core media classifier’s image-only fact through the existing preparation capability and covered by a real-host regression test. Final independent review found no blocking issues. The Python-based autoreview launcher was not run because this contribution's workflow excludes Python.
- No full release build or live Telegram end-to-end verification of this source implementation was performed.

### DIFF
--- a/docs/plugins/codex-harness.md
+++ b/docs/plugins/codex-harness.md
@@ -19,6 +19,13 @@ the current turn as untrusted external content, subject to the existing file
 extraction limits. This also applies to adopted and forked Codex sessions with
 locked model selection. Images continue through Codex's native image input.
 
+For Telegram, a new photo or album takes priority over saved photos when OpenClaw
+rebuilds context for Codex. A text-only follow-up includes the most recent saved
+photo or album. Older captions and attachment references remain available; the
+agent can retrieve an older image when you explicitly ask about it. Saved
+documents and mixed document/image messages keep their existing behavior. This
+does not erase images already present in a resumed native Codex thread.
+
 Remote Codex app-servers can run on a different machine from the Gateway. Set
 `remoteWorkspaceRoot` to validate remote workspace attachment paths. OpenClaw
 transfers authoritative attachment bytes over the existing app-server connection
--- a/extensions/codex/src/app-server/context-engine-projection.test.ts
+++ b/extensions/codex/src/app-server/context-engine-projection.test.ts
@@ -107,6 +107,81 @@ describe("projectContextEngineAssemblyForCodex", () => {
     );
   });
 
+  it.each(["all", "latest-turn", "none"] as const)(
+    "selects photo history without changing transcripts or document pages (%s)",
+    async (imageHistory) => {
+      const image = (data: string) => ({ type: "image" as const, mimeType: "image/png", data });
+      const older = { role: "user" as const, content: [image("old")], timestamp: 1 };
+      const album = {
+        role: "user" as const,
+        content: "New album caption",
+        timestamp: 2,
+        __openclaw: {
+          media: [
+            { kind: "image", path: "/tmp/first.png" },
+            { kind: "image", path: "/tmp/second.png" },
+          ],
+        },
+      };
+      const document = textMessage("user", "Document caption");
+      const messages = [older, album, document];
+      const original = structuredClone(messages);
+      const result = await projectContextEngineAssemblyForCodex({
+        assembledMessages: messages,
+        originalHistoryMessages: messages,
+        prompt: "What is in the latest picture?",
+        imageHistory,
+        maxRenderedContextChars: 4 * IMAGE_BLOCK_TOKENS * 4 + 2_000,
+        prepareFileContext: async (message) => ({
+          imageOnly: message !== document,
+          text:
+            message === older
+              ? "old.png"
+              : message === album
+                ? "first.png second.png"
+                : "Document text",
+          images:
+            message === older
+              ? [image("old")]
+              : message === album
+                ? [image("first"), image("second")]
+                : [image("document-page")],
+        }),
+      });
+      expect(result.images).toEqual([
+        ...(imageHistory === "all" ? [image("old")] : []),
+        ...(imageHistory !== "none" ? [image("first"), image("second")] : []),
+        image("document-page"),
+      ]);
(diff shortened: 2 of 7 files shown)


pbanuru:
Matrix reproduces the same selection defect on OpenClaw 2026.9.4 with `@openclaw/codex` 2026.9.4.

Observed in a direct Matrix session:

1. Two images were attached in separate, earlier user events.
2. Later Matrix events were text-only. Their canonical transcript rows and session-history results contained no media.
3. On each later turn, both historical image files were hydrated again and supplied at the Codex turn-input boundary.
4. The model received them as if they accompanied the new text-only request and interpreted them as current context.

This is downstream of Matrix ingestion. The current inbound event is clean; history projection reintroduces the pixels.

The proposed channel check in `run-attempt-prompt.ts` would not affect this case because it applies only when `messageChannel/messageProvider === "telegram"`. Even if extended to Matrix unchanged, `latest-turn` would still inject one unrelated historical image into a text-only turn.

Recommended invariant:

- A turn with current images supplies exactly those current images.
- A text-only turn supplies zero historical pixels by default.
- Historical attachment notes and references remain in projected text so an explicit comparison or retrieval can load the requested image deliberately.
- Any automatic latest-image follow-up behavior should be an explicit cross-channel policy, not inferred from the transport, because one session can be reached from multiple clients.

The merged #147913 fixes image-to-message adjacency, which is valuable, but does not resolve this selection problem.

Suggested synthetic regression:

- historical Matrix image A
- historical Matrix image B
- current text-only event => zero image inputs
- current event with image C => C exactly once
- restart/replay preserves both outcomes

No private image bytes, filenames, message text, account identifiers, or local paths are included here.

pbanuru:
Correction to my recommendation above after inspecting the exact OpenAI Codex source used here, `@openai/codex` 0.155.0 at `rust-v0.155.0` (`f0a1b8f0849d90960bc406b848f32e5a129b0457`).

Native Codex does retain historical image pixels. The critical invariant is not zero historical pixels. It is message ownership:

- each historical image remains content of its original historical user `ResponseItem::Message`;
- a new `turn/start` input becomes one new user message containing only that turn input;
- restored history is added as distinct response items, not flattened into the new user message.

Evidence:

- `ResponseInputItem::from_user_input` converts the full `turn/start` input vector into one user message: https://github.com/openai/codex/blob/f0a1b8f0849d90960bc406b848f32e5a129b0457/codex-rs/protocol/src/models.rs#L1982-L2009
- `thread/inject_items` records raw Responses items without starting a turn: https://github.com/openai/codex/blob/f0a1b8f0849d90960bc406b848f32e5a129b0457/codex-rs/core/src/codex_thread.rs#L636-L672
- Codex regression coverage for resumed image history explicitly asserts two separate image-bearing message indices, one historical and one current: https://github.com/openai/codex/blob/f0a1b8f0849d90960bc406b848f32e5a129b0457/codex-rs/core/tests/suite/image_rollout.rs#L385-L405

This means the merged adjacency work in #147913 is a useful improvement, but it still projects retained history and the current request through one `turn/start` vector. Codex therefore persists them as one new user message. Ordered text/image parts preserve local adjacency, but not native historical message boundaries.

Revised recommended invariant:

1. Import or restore retained historical messages as distinct Responses API items, using `thread/inject_items` on a fresh or rotated native thread.
2. Send only the current inbound event through `turn/start`.
3. Preserve every retained image inside the content array of its owning historical user message.
4. On warm threads, rely on Codex native history instead of re-projecting the same history.
5. Test the provider request shape, not only item ordering: historical A and current text must be separate user messages; historical A and current image C must also be separate user messages.

So I withdraw the earlier default recommendation of zero historical pixels. The Telegram selection policy in this PR may still be a deliberate product option, but it is not how native Codex models conversation history. The shared projection boundary is the root cause of the Matrix behavior I reported.

PR #144524, opened 2026-09-11
fix(auth): keep setup replacements inactive until activation

obviyus:
## What Problem This Solves

A rejected replacement credential could enter normal rotation. Successful testing activated replacements without a separate choice, and retry details were lost after restart.

## Why This Change Was Made

The shared credential owner persists retry state in the existing payload, excludes inactive replacements from use and copying, and activates only after testing, confirmation, and config application. Setup and Model Setup use this same flow.

## User Impact

Working connections survive rejection or cancellation. Saved replacements can be retried after restart; ordinary login remains immediate. Before downgrading to older binaries, remove inactive replacements or restore pre-setup state because older versions do not enforce inactivity.

## Evidence

Real Gateway and browser checks cover rejected replacements, continued use of the old credential, decline, cancellation, restart retry, explicit activation, and stale activation after deletion. Targeted setup, credential-persistence, noninteractive, and rendering tests passed. Live proof used one credential with distinct profiles; external browser login and refresh were not covered.

Related: #136257, #144195. [Downgrade boundary](https://github.com/openclaw/openclaw/pull/144524#issuecomment-5627683210).

PR #144436, opened 2026-09-10
fix: report saved credentials and auth refresh failures

obviyus:
Related: #136257, #144181.

## What Problem This Solves

Fixes an issue where private-chat login reported completion after saving credentials even when the Gateway rejected authentication refresh or could not be reached. Settings-write failures also hid that credentials were already saved.

## Why This Change Was Made

The login result carries refresh outcomes and saved-credential state. The shared login runtime owns completion and failure wording for generic chat and native Telegram while retaining cancellation, session switching and delivery behavior.

## User Impact

Users receive accurate saved-credential and recovery guidance after partial success. Successful refresh keeps existing retry guidance. No configuration or permission change.

Consumers: generic chat and native Telegram render the shared login outcome.

## Evidence

Producer and presentation regressions failed before the fix; 106 focused tests passed. Five real Telegram recordings verified acknowledged, rejected and unreachable refresh, rejected settings, and failure before save. Credentials were checked before and after each result. Synthetic login credentials were used; external authentication was not proved.

PR #144568, opened 2026-09-11
fix(setup): preserve config edits made during provider preparation

obviyus:
## What Problem This Solves

Fixes an issue where users preparing a provider through Gateway setup received a raw config-changed error after saving credentials if another config edit occurred during login. The unrelated edit survived, but the provider settings were not applied and the error did not explain that the credential was already saved.

## Why This Change Was Made

CLI login and Gateway preparation now share one config writer. It applies only the provider's changes to the latest authored config, merges model rows by ID, rejects conflicting or invalid edits, and uses the existing saved-credential/settings-failure error. Pending plugin-install records remain with their existing commit owner.

The setup path no longer binds publication to the snapshot taken before login. Strict snapshot checks remain for migration promotion and verified setup operations that require them. Credential persistence, cancellation, and default-model policy retain their existing owners.

## User Impact

Unrelated edits made while provider setup waits for input survive, and setup can complete. If settings cannot be applied, setup explains that credentials were saved and retains the config owner's recovery guidance.

## Evidence

- Reproduced on main `a321595f496e5a95d2bd0ffe81f7a080b4fcdfc8` through the shipped Gateway client: start preparation, edit an unrelated setting while the provider key prompt waits, then complete. Before: raw config-changed error with credential saved. After: setup completes with the edit and credential retained.
- Real config-directory permission failure: before, config-only error despite a saved credential; after, explicit saved-credential/settings-failure message. Used isolated state and synthetic credentials.
- 45 targeted Gateway setup tests and 55 CLI auth tests pass. The three original new regressions fail on main. Added protection for a config that becomes invalid during login.
- Focused lint, formatting, line-count and assertion checks pass. Full type checks run in GitHub CI.

Related: #144181, #144436. No changes to the credential replacement paths under #144524.

### DIFF
--- a/src/commands/models/auth.test.ts
+++ b/src/commands/models/auth.test.ts
@@ -178,6 +178,24 @@ vi.mock("./shared.js", async (importOriginal) => {
   };
 });
 
+vi.mock("../../plugins/install-record-commit.js", () => ({
+  transformConfigWithPendingPluginInstalls: async (params: {
+    transform: (
+      current: OpenClawConfig,
+      context: { snapshot: { valid: boolean } },
+    ) => { nextConfig: OpenClawConfig };
+    writeOptions?: { beforeCommit?: () => void };
+  }) => {
+    const next = await mocks.updateConfig(
+      (current: OpenClawConfig) =>
+        params.transform(current, { snapshot: { valid: true } }).nextConfig,
+      undefined,
+      params.writeOptions?.beforeCommit,
+    );
+    return { nextConfig: next, result: next };
+  },
+}));
+
 vi.mock("../../config/logging.js", () => ({
   logConfigUpdated: mocks.logConfigUpdated,
 }));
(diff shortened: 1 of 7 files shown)


obviyus:
**Proof:** Gateway provider preparation completes after an unrelated concurrent config edit. A real config write failure reports that credentials were saved.

- **Bug:** setup saved the credential, then failed on the snapshot taken before login and returned a config-only error.
- **Fix:** setup and CLI login share a writer that replays provider changes against current settings and reports partial persistence clearly.

| Before | After |
| --- | --- |
| Concurrent edit caused a raw config-changed error after saving credentials. | Setup completes and preserves the concurrent edit. |
| Config permission failure did not explain that credentials were saved. | The wizard explains that credentials were saved but provider settings could not be applied. |

Validated with real Gateway calls in isolated state, independent behavior checks, 45 Gateway setup tests, 55 CLI auth tests, and focused lint.

[Automated checks passed on attempt 2](https://github.com/openclaw/openclaw/actions/runs/34551714214/attempts/2). The first attempt hit a Matrix import-hook timeout; the failed jobs passed on a fresh runner with no source or timeout changes. No failing-check exception was used.

PR #144572, opened 2026-09-11
fix(memory): serve lexical results during OpenClaw chunking upgrade

### LINKED ISSUE #144493: [Bug]: Failed chunking upgrade hides available lexical memory
### Bug type

### Beta release blocker

No

### Summary

Memory search returns no results from an available published lexical index when an automatic OpenClaw chunking-version upgrade cannot rebuild the index.

### Steps to reproduce

1. Publish a memory index with FTS enabled.
2. Mark its metadata as using the preceding OpenClaw chunking version.
3. Change an indexed memory file so the upgrade rebuild requires a new embedding.
4. Make the embedding provider return a permanent quota error.
5. Search for a term contained in the last published index.

### Expected behavior

Search returns matching keyword results from the existing published FTS generation while status continues to report the OpenClaw-owned chunking upgrade as pending.

### Actual behavior

Search returns an empty result set even though the prior FTS generation remains available and queryable.

### OpenClaw version

`main` at `8609f24ece1b2b3dff073544df7e12467a16a2b1`

### Operating system

macOS 26.5.1 (25F80)

### Install method

pnpm source checkout

### Model

Memory embedding provider (deterministic test mock)

### Provider / routing chain

OpenClaw memory-core -> embedding provider; SQLite FTS remains local

The deterministic reproduction uses the bundled memory-core fixture, disables vector search, and injects a structured permanent quota failure for the rebuild.

### Logs

```shell
Expected: lexical hit for memory/2026-01-12.md
Actual: []
Index identity: mismatched, owner=openclaw, code=chunking_version
FTS: enabled and available
```

A focused regression test reproduces the empty result before the orchestration change and returns the prior lexical hit after it.

### Impact and severity

- Affected: memory search after an OpenClaw chunking upgrade when re-embedding is temporarily or permanently unavailable
- Severity: workflow-blocking for memory recall
- Frequency: deterministic under the reproduction conditions
- Consequence: usable lexical memory appears unavailable until the embedding rebuild succeeds

### Additional information

The proposed boundary applies only to OpenClaw-owned `chunking_version` mismatches with a working FTS index. Configuration-owned mismatches continue to fail closed, and status remains mismatched rather than presenting stale content as current.

pengzh1:
## What Problem This Solves

When an OpenClaw-owned `chunking_version` mismatch cannot be repaired (e.g. embedding provider quota exhaustion), search returned an empty result set even though the prior published FTS generation remained available and queryable (#144493).

## What This Changes

Search now degrades to keyword-only retrieval from the last published generation while an OpenClaw chunking upgrade rebuild keeps failing — with corpus compatibility enforced:

- `manager-search-orchestration.ts`: one eligibility helper (`canServeLexicalDuringChunkingUpgrade`) guards both the pre-lease identity check **and** the leased-generation revalidation loop, and an explicit `keywordOnlyUpgrade` decision is carried into retrieval, skipping the semantic lane against the stale generation. Other mismatches keep failing closed.
- `manager-reindex-state.ts`: a stale chunking version no longer masks corpus drift — sources, scope (extra paths + multimodal), chunk settings, or FTS tokenizer changes under an old-chunking index classify as configuration-owned, so a scope narrowed during an upgrade cannot keep exposing excluded paths.
- `memory-search-tool-query.ts`: the `memory_search` tool accepts the narrowly validated lexical generation instead of reporting search unavailable; the existing staleness reporting surfaces the pending upgrade (`stale` + warning + rebuild guidance) on the tool payload.
- The previously included Control UI changes were removed: string/boolean literal union editing is already handled by #143700 on main.

## Evidence

Regression tests (all red against unpatched files, green with the fixes):

- `returns lexical results while a failing upgrade keeps the rebuild pending` — `batch-test` provider runtime so document embeddings flow through `runtime.batchEmbed`; permanently fails **both** embedding lanes (`providerRuntimeBatchFailuresRemaining` + a new `embedBatchFailuresRemaining` fixture control for the non-batch fallback), matching production quota exhaustion; asserts an ordinary `manager.search("alpha")` still serves lexical hits while identity stays `mismatched/chunking_version/openclaw` and published metadata keeps the old `chunkingVersion`.
- `keeps removed extra paths excluded while a chunking upgrade stays pending` — seeds an index with an extra path, removes it from config, fails the rebuild, and asserts search returns `[]` with identity `scope/configuration` (no excluded-snippet leak).
- Resolver corpus-precedence cases: stale chunking + sources/scope/chunk-settings/tokenizer drift → configuration-owned codes.
- `memory_search` tool entry point: serves keyword hits with `mode: keyword-only` plus `stale`/`warning`/`action` diagnostics during the pending upgrade.

Local verification: four suites 80/80 green; `tsgo:extensions:test` clean; `oxfmt --check` + `oxlint` clean. (69 pre-existing Windows-environment failures in the full local directory run reproduce identically on unmodified upstream main and are unrelated to this branch.)

Fixes #144493


pengzh1:
All four findings from the 2026-09-11 review are addressed on the new head `ff8bd2102a5` (rebased onto current main, UI changes removed):

**[P1] Duplicate string-union admission path** — the seven UI commits are gone from this branch. `ui/src/components/config-form.analyze.ts` is no longer touched; main's guarded implementation from #143700 stands. The P1 merge-risk (Settings admission bypassing sentinel-type protections) is eliminated by removal, not re-patching.

**[P2] Carry the lexical fallback through generation revalidation** — the eligibility check is now applied to the leased generation inside the revalidation loop (`canServeLexicalDuringChunkingUpgrade(leasedIdentity)` → break with `keywordOnlyUpgrade = true`), and that decision flows into retrieval (`keywordOnly = embeddingBootstrapKeywordOnly || keywordOnlyUpgrade || !this.provider || opts?.lexicalOnly`). Other mismatches still release the lease and fail closed. This was verified as a real defect: the regression below is red against unpatched orchestration precisely because the old fallback only passed the first gate.

**[P2] Restore database-path bindings** — both `const dbPath = await seedIndex(cfg)` bindings are back, so the metadata assertions at lines 78/92 resolve again (`check-test-types` failure source).

**[P2] Inject the rebuild failure into the provider actually selected** — the test now selects the `batch-test` provider with batching enabled so document embeddings run through `runtime.batchEmbed`. Because that lane's failure is absorbed by the non-batch fallback (`runBatchWithFallback` → `provider.embedBatch`, which the mock never failed), the fixture gained an `embedBatchFailuresRemaining` control and the test fails **both** lanes — the same shape as production quota exhaustion, where batch is disabled after two failures and per-item calls fail too. The test asserts lexical hits from an ordinary search (no `lexicalOnly`), plus that the upgrade stays pending: identity remains `mismatched/chunking_version/openclaw` and published `chunkingVersion` is unchanged.

Local verification: red/green cycle on the new test (red on upstream orchestration, green with fix), `manager-search-upgrade.test.ts` 6/6, fixture consumers `index.test.ts` + `manager-memory-source-race.test.ts` + `manager-session-update-race.test.ts` 102/102, `tsgo:extensions:test` clean, `oxfmt --check` + `oxlint` clean on all three touched files.

@clawsweeper re-review

pengzh1:
Revision 3 pushed as `c74d42f0ce7` — both new findings addressed:

**[P1] Validate corpus compatibility before accepting the old generation** — the identity resolver now classifies a stale chunking version together with any lexical-corpus change as configuration-owned before reporting `chunking_version`: sources drift, `scopeHash` drift (extra paths + multimodal), chunk-setting drift, and FTS tokenizer drift each win precedence with a reason that names the upgrade-pending context. The eligibility helper used by the pre-lease gate and the leased-generation revalidation therefore rejects the old index in exactly the reported scenario — an operator removes an `extraPaths` directory during an upgrade, re-embedding keeps failing, and the excluded snippets can no longer reach manager or CLI output. The regression `keeps removed extra paths excluded while a chunking upgrade stays pending` reproduces that scenario end to end: with the resolver unfixed it fails because the excluded path stays searchable (red run captured); with the fix, search returns `[]`, identity reports `scope/configuration`, and the published metadata keeps the old chunking version.

**[P2] Carry the fallback through the memory tool's diagnostic gate** — `finalizeMemorySearchToolQuery` now treats only the narrowly validated `chunking_version` diagnostic as servable; every other identity diagnostic (configuration-owner, provenance, metadata_missing) still returns the paused result. In the servable case the results flow through unchanged and the existing staleness pipeline reports the pending upgrade on the tool payload (`stale: true` + "Memory index is stale: …" + rebuild guidance), so diagnostics are retained rather than turned into an unavailable error. The new `memory_search` entry-point test asserts hits, `mode: keyword-only`, and the staleness/action diagnostics on the tool output.

Verification this round: both new tests are red against the unpatched files and green with the fixes; `manager-search-upgrade.test.ts` (8), `manager-reindex-state.test.ts` (+4 corpus-precedence cases), `memory-search-tool-query.test.ts`, and `tools.test.ts` (43) all pass — 80/80 across the four suites; `tsgo:extensions:test` clean; `oxfmt --check` + `oxlint` clean on all touched files. (Full-directory local runs show 69 pre-existing Windows-environment failures — darwin watch routing, timezone, and unicode-normalization cases — reproduced identically on unmodified upstream main, so they are unrelated to this branch.)

@clawsweeper re-review

pengzh1:
@clawsweeper re-review

pengzh1:
Status note on the re-reviews: both re-review runs on head `c74d42f0ce7` went silent after the worker started (17:46 UTC and again after the 20:04 UTC trigger), with no verdict posted, so I can't tell whether findings remain. If a maintainer or anyone with repository write access can comment `@clawsweeper re-review` (or rerun the worker), it would be appreciated — the head is CI-green on its own lanes and the branch diff is memory-core only.

pengzh1:
Yielding to #144576 as the canonical replacement for #144493: it covers the same failed-upgrade lexical recovery (including the corpus-compatibility and FTS-diagnostic boundaries this branch hardened), and its real-transport proof is stronger than what I can produce without workspace-level embeddings access.

Closing this branch in its favor — thanks to @clawsweeper for the detailed review cycles that sharpened the corpus-compatibility precedence fix, and to @ruel225 for the broader treatment. If #144576 stalls, the branch history here (fork: `fix/issue-144493`) documents the remaining pieces.

PR #144576, opened 2026-09-11
fix(memory-core): keep lexical recall readable during a pending chunking upgrade

### LINKED ISSUE #144493: [Bug]: Failed chunking upgrade hides available lexical memory

ruel225:
## What Problem This Solves

Fixes lost lexical memory after an OpenClaw software upgrade when rebuilding an older chunking-version index cannot obtain embeddings.

Closes #144493.

## User Impact

Users can still search compatible, already-indexed memory by keyword while the upgrade rebuild is pending. Results remain marked stale with the repair failure and provider-cost guidance. Newer-than-runtime indexes and incompatible corpora remain unreadable; this does not expand recovery for ordinary current-version embedding outages.

## Why This Change Was Made

The existing index identity owner now distinguishes an older chunker with matching provenance, sources, scope, chunk settings, and tokenizer. Search permits lexical retrieval only with usable FTS, revalidates eligibility after acquiring the read-generation lease, and preserves asynchronous release. Embedding identity is not used to authorize keyword reads because no stored vectors are consumed.

There is no stored schema or version change. Automatic upgrade repair remains enabled; failed repair does not replace the published index. Original contribution by @ruel225, adapted to current main while retaining contributor ancestry.

## Evidence

- Focused identity, upgrade, real-manager, SDK diagnostics, and asynchronous generation coverage: 96 relevant tests passed across the completed runs. Real local HTTP 401 and 429 cases verify both available keyword recall and fail-closed scope/FTS behavior.
- Genuine prior-producer qualification: the immutable version-4 chunker from `70a949a7dd2949edb366dae0810c999d2e44279a` created an isolated index. Unpatched version-5 search then lost recall after an actual HTTP 401; this candidate returned the cited stored keyword hit against the same corpus and failure. Metadata, chunks, source rows, and FTS rows were unchanged in both consuming runs.
- This qualification uses the historical chunking producer in a compatible isolated test host, not a full published-release installation or a live Gateway. The committed transport regressions use synthetic version metadata and do not claim otherwise.
- All planned changed checks passed, including four typecheck stages, dead-export scans, and core/extension lint. Extension declaration/lint ran on a separate physical copy of the exact candidate tree because the native nested worktree failed the ancestor-install boundary guard.
- Independent API-only review of the exact integrated candidate completed scoped-clean at P0–P2 (three passes, no actionable findings).
- [Exact-head hosted CI](https://github.com/openclaw/openclaw/actions/runs/35439838166) passed on `a06cd91162872d5ecd1415818441af68f7d94577`. The initial run had two timer-cleanup assertions fail in unchanged session-diagnostics tests; the narrow 10-test diagnostic passed locally and one accepted same-head failed-shard retry passed hosted CI. No unrelated source change or gate bypass was used.

### DIFF
--- a/extensions/memory-core/src/chunking-upgrade-real-transport.test.ts
+++ b/extensions/memory-core/src/chunking-upgrade-real-transport.test.ts
@@ -1,0 +1,390 @@
+// Memory Core tests prove the chunking upgrade fallback against a real HTTP
+// embedding server: the outage reaches the wire through the real
+// openai-compatible provider, the real embedding retry policy, and the real
+// SQLite store instead of a test-double provider.
+import { mkdirSync, rmSync } from "node:fs";
+import fs from "node:fs/promises";
+import { createServer, type IncomingMessage, type ServerResponse } from "node:http";
+import type { AddressInfo } from "node:net";
+import path from "node:path";
+import { DatabaseSync } from "node:sqlite";
+import { MEMORY_CHUNKING_VERSION } from "openclaw/plugin-sdk/memory-core-host-engine-storage";
+import type { OpenClawConfig } from "openclaw/plugin-sdk/memory-core-host-runtime-core";
+import {
+  closeOpenClawAgentDatabasesForTest,
+  closeOpenClawStateDatabaseForTest,
+} from "openclaw/plugin-sdk/sqlite-runtime-testing";
+import { resolvePreferredOpenClawTmpDir } from "openclaw/plugin-sdk/temp-path";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import type { MemoryIndexManager } from "./memory/manager.js";
+import { isolateMemoryManagerTestConfig } from "./memory/test-config-helpers.js";
+import "./memory/test-runtime-mocks.js";
+import {
+  configureMemoryCoreDreamingStateForTests,
+  resetMemoryCoreDreamingStateForTests,
+} from "./test-helpers.js";
+import { createMemorySearchTool, testing } from "./tools.js";
+
+const { closeAllMemorySearchManagers, getMemorySearchManager } = await import("./memory/index.js");
+
+type ServerMode = "ok" | "unauthorized" | "quota";
+
+type CapturedRequest = {
+  method: string | undefined;
+  url: string | undefined;
+  status: number;
+  body: Record<string, unknown>;
+};
+
+type EmbeddingServer = {
+  baseUrl: string;
+  requests: CapturedRequest[];
+  setMode: (mode: ServerMode) => void;
+  close: () => Promise<void>;
+};
+
+const servers: EmbeddingServer[] = [];
+
+async function readJsonBody(req: IncomingMessage): Promise<Record<string, unknown>> {
+  const chunks: Buffer[] = [];
+  for await (const chunk of req) {
+    chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk));
+  }
+  return JSON.parse(Buffer.concat(chunks).toString("utf8")) as Record<string, unknown>;
+}
+
+function errorBody(mode: Exclude<ServerMode, "ok">): Record<string, unknown> {
+  return mode === "quota"
+    ? {
+        error: {
+          message: "You exceeded your current quota, please check your plan and billing details.",
+          type: "insufficient_quota",
+          code: "insufficient_quota",
+        },
+      }
+    : {
+        error: {
+          message: "Invalid API key provided.",
+          type: "invalid_request_error",
+          code: "invalid_api_key",
+        },
+      };
+}
+
+async function startEmbeddingServer(): Promise<EmbeddingServer> {
+  const requests: CapturedRequest[] = [];
+  let mode: ServerMode = "ok";
+  const server = createServer((req: IncomingMessage, res: ServerResponse) => {
+    void (async () => {
+      try {
+        const body = await readJsonBody(req);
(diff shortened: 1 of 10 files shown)


ruel225:
@clawsweeper re-review

The [P2] FTS-eligibility finding and the [P1] SDK-marker wording are repaired at head `a8bffde3f1f` (one commit on top of the reviewed `77521913e3`); the "What Problem This Solves" section gate is also fixed (body updated).

- Tool boundary: `finalizeMemorySearchToolQuery` now requires the published FTS state (`status.fts.enabled && status.fts.available`) alongside the `chunkingVersionOnly` marker — the same compound condition the manager's keyword fallback applies — so a pending upgrade with no usable FTS keeps `pausedIndexIdentity` and `memory_search` reports `unavailable` instead of a successful empty search with only a staleness warning.
- SDK contract: `chunkingVersionOnly`'s public doc now states it is a corpus-compatibility marker (sources, scope hash, chunk settings, FTS tokenizer), excludes embedding identity (provider/model/settings/dims may differ), and does not establish keyword retrieval availability; consumers must check usable FTS separately.
- New regression `keeps memory_search paused when a pending upgrade has no usable FTS index` (tools.real-manager.test.ts): the published database's FTS table name is occupied by a non-FTS object so every schema ensure — including the upgrade rebuild's republish — fails to restore it; pre-fix (tool-side gate reverted) the tool returned `results: []` with only the stale-index warning, post-fix it keeps the paused diagnostic. The successful-fallback and changed-scope scenarios are unchanged and green.
- The CI type failure on the previous head (`types.test.ts` excess `backend`/`provider` properties) is fixed; `ci-gate` is green on `a8bffde3f1f` (132 checks pass, 0 fail).

ruel225:
@clawsweeper re-review

Added the real-transport proof requested in the review (`chunking-upgrade-real-transport.test.ts`, head `869f4663447`): the pending-upgrade serving and fail-closed scenarios now run against a real local HTTP embeddings server through the production `openai-compatible` adapter — including a real OpenAI-style 429 `insufficient_quota` body that the index retry policy exhausts its five-attempt budget on over the wire, and real 401 rejections driving the tool-level scenarios. PR body Evidence and proof-scope note updated accordingly.

ruel225:
@clawsweeper re-review

Added the requested data-model compatibility proof on head `3bc1d2de701`:

- **Storage surface unchanged** — no SQLite table, column, index, migration, or backfill; `chunkingVersionOnly` is derived at read time from the already-stored `memory_index_meta` row and is never persisted. A new "Data-Model Compatibility" section in the PR body walks the three directions (older database → new code, proven by `seedPriorChunkingVersionIndex` rewriting a real stored meta row to the previous release's `chunkingVersion` and serving the fallback over the real transport; additive-optional diagnostic → older readers, pinned both ways by strict `toEqual` in `types.test.ts`; fail-closed scope/provenance invariants pinned in `manager-reindex-state.test.ts`).
- **CI repair** — the prior run's single failure was my 429 scenario pinning a retry count (`expected 2 to be greater than or equal to 5`): how many rate-limit retries fit into a run is clock-dependent, so the run now asserts at least one wire-visible 429 while the five-attempt budget stays owned by the embedding-policy unit tests. The scenario passed locally after the change (175s, real backoff). All other checks in that shard passed (2429 tests green).

fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [a06cd91162872d5ecd1415818441af68f7d94577](https://github.com/openclaw/openclaw/pull/144576/commits/a06cd91162872d5ecd1415818441af68f7d94577)
- Landed commit: [abbda386ef9f5b9221b402bf131eed0db8630260](https://github.com/openclaw/openclaw/commit/abbda386ef9f5b9221b402bf131eed0db8630260)

PR #144583, opened 2026-09-11
fix(agents): recover tasks when Responses streams end mid-tool-call

steipete:
## What Problem This Solves

Fixes an issue where users running agent tasks would see a generic "Agent run failed" error when an OpenAI Responses stream ended while a tool call was still unfinished, even though the task could continue from its saved transcript.

## Why This Change Was Made

Recognize the transport's exact premature-end error in the shared incomplete-stream matcher, so existing transient continuation and settled-result finalization can handle it. Completed responses with inconsistent tool calls remain excluded; partial tool arguments are never admitted for execution.

## User Impact

Tasks can recover from this interrupted stream without resubmitting the original request or automatically repeating completed tool actions. The existing retry budget and cancellation rules continue to apply.

## Evidence

- The synthetic regression feeds an unfinished call through the real Responses stream parser and AgentSession. Before the fix the session ended without an answer; afterward it recovers, the earlier tool executes exactly once, its result survives, and the unfinished call never executes.
- Embedded-runner coverage verifies continuation from the existing transcript and preservation of settled-result finalization context. Classification coverage distinguishes premature EOF from an inconsistent completed response.
- 1,342 focused tests passed across the session regression, retry/classification corpus, embedded recovery, and finalization-context suites.
- The complete changed-file check passed, including core and test typechecking, lint, formatting, and repository guards. Independent blocking-issue review passed.
- The original stream cutoff was observed in runtime logs, but its upstream cause remains unknown because raw SSE events were not recorded. No live deployment is claimed by this PR.

### DIFF
--- a/docs/concepts/retry.md
+++ b/docs/concepts/retry.md
@@ -37,6 +37,8 @@ Agent runs automatically recover from temporary rate limits, overloads, and prov
 
 Recovery continues the existing transcript with an instruction to preserve completed work and inspect interrupted actions before deciding whether to repeat them. It can recover a throttle after tool activity or partial output without resubmitting the original user request. The run shows one transient retry indicator while waiting and remains cancellable. Recovered attempts do not leave persisted assistant errors; only terminal failure retains one error. Billing failures, authentication errors, and provider refusals do not use this transient retry budget.
 
+A Responses stream that ends before its terminal event also qualifies for transient recovery, including when a tool call is still unfinished. Partial tool arguments are never executed. A completed response with inconsistent tool-call identities does not qualify as a disconnected stream.
+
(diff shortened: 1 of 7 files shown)


steipete:
@clawsweeper re-review

The native merge guard requires a fresh review because the previous review expired overnight. The head remains bc09908c208b27871ce0950e37e41175560f8fbd; exact-head CI and native preparation have passed, and no code changed after the previous clean review.

steipete:
Merged via squash.

- Prepared head SHA: [bc09908c208b27871ce0950e37e41175560f8fbd](https://github.com/openclaw/openclaw/pull/144583/commits/bc09908c208b27871ce0950e37e41175560f8fbd)
- Landed commit: [2bfa3fe7e5de127976a06a4e823e4cf44e327b49](https://github.com/openclaw/openclaw/commit/2bfa3fe7e5de127976a06a4e823e4cf44e327b49)

PR #144563, opened 2026-09-11
refactor(models): share prepared catalog views across model pickers

obviyus:
## What Problem This Solves

Model browsing maintained separate implementations for native readiness, explicit refresh, configured model inclusion, and provider-scoped acquisition.

## Why This Change Was Made

The prepared catalog view now owns those operations. Gateway lists, catalog decisions, and both pickers consume that view; four superseded helpers are removed.

## User Impact

Model rows, availability, refusal guidance, and ordering stay the same. Ordinary reads remain passive, and explicit refresh keeps its acquisition behavior. No configuration or migration change.

## Evidence

Focused catalog, Gateway, picker, and native-plugin tests passed. A mutation removing configured-model inclusion failed its owner test. Before/after CLI model lists and authenticated Gateway refresh preserved rows and availability. Real browser picker tests passed before and after, covering desktop/mobile choices, search, order, and catalog replacement. The final rebase retained the tested patch.

Related: #136257, #144342.

PR #144342, opened 2026-09-10
refactor(models): remove superseded catalog paths

obviyus:
Related: #136257.

## What Problem This Solves

Removes unused catalog adapters and discarded onboarding work left after catalog and native sign-in changes.

## Why This Change Was Made

Doctor calls the existing configured-model projection and alias resolver directly. Remove the discarded credential-file candidate and unused document-capability wrapper. Move meaningful projection and alias tests to their existing owners.

## User Impact

Model listing and selection retain their behavior while production code shrinks by 50 lines. No configuration, schema, protocol or dependency change.

## Evidence

Focused validation passed 182 tests across nine suites, plus a first-wins catalog check; final command-line suites passed 20 tests. An intentionally removed alias transfer made its regression fail. Before/after command-line listing and browser selection returned the same configured model and tags. Runtime/UI builds and focused static checks passed.

PR #144589, opened 2026-09-11
refactor(models): retire duplicate auth and catalog test seams

obviyus:
Related: #136257, #144342, #144563.

## What Problem This Solves

Auth and prepared-catalog changes left duplicate test setup and two obsolete file boundaries, making their remaining behavior harder to maintain.

## Why This Change Was Made

Named-login assertions move to the command boundary, and the profile resolver becomes private. Duplicate authentication, account-preview, and catalog cases are consolidated. The unchanged output-flag guard moves into the shared command module. Credential readers and explicit imports remain where their consumers still need them.

## User Impact

No behavior changes. Model listing, conflicting output flags, named profiles, account-preview recovery, and native sign-in status retain their contracts. No configuration, schema, protocol, or dependency change is introduced.

## Evidence

All 241 focused tests passed across command, login, persistence, reply, and UI suites. After rebase, the six affected suites passed 143 tests. Before/after command-line and authenticated Gateway model lists returned the same fixture model; conflicting output flags returned the same error. Isolated native sign-in status remained signed out. Formatting, lint, structural, and whitespace checks passed. Interactive sign-in and inference were not tested.


obviyus:
Landed as `82c306d924c0ea00fac9e04a7b66e446e60fc5b3`.

**Proof:** 241 focused tests passed across 11 suites. After the final rebase, 143 tests passed across the six affected suites. Formatting, targeted lint, ratchets, and full [CI](https://github.com/openclaw/openclaw/actions/runs/34554687427) passed. The exact-head local review and server-side review found no blocking issue.

**Change:** Retired two internal file boundaries and duplicate test setup. Named-profile assertions now run through the login command; the output-flag guard stays unchanged in the shared command owner.

| Before | After |
| --- | --- |
| CLI and authenticated Gateway return the available fixture model | Same output |
| Conflicting output flags fail with the documented error | Same error and exit code |
| Isolated native status reports signed out | Same status |
| Helper-only profile tests and duplicate cases | Command-boundary assertions and retained distinct coverage |

F16 remains a named follow-up: core setup, status, and identity consumers still need facts beyond the plugin's native login-status probe. Explicit Codex auth/memory import remains plugin-owned. The separate prepared-auth endpoint lane was untouched.

PR #144599, opened 2026-09-11
fix(gateway): start with providers without models after #143837

obviyus:
Related: #143837.

## What Problem This Solves

Gateway startup could crash after plugin auto-enable when a bundled provider had credentials or settings but no custom models array.

## Why This Change Was Made

The catalog composition owner maps custom rows only when present. Existing merge logic preserves inherited rows and routes while applying current request credentials and headers.

## User Impact

Provider settings can omit custom models, including when startup enables a configured channel. Bundled models remain available. Provider-level enabled keys remain invalid; no schema or configuration keys change.

## Evidence

Real isolated startup and both new regressions failed before the fix at the same missing-array read. All 61 focused catalog, startup, and connection tests passed afterward. The Gateway reached readiness; explicit catalog refresh exposed ten bundled models, and a passive list retained them. Synthetic credentials left the models unavailable, so upstream authentication was not proved. Channel connections were disabled. Formatting, lint, and structural checks passed.


obviyus:
Thanks @obviyus!

**Proof:** Real isolated Gateway startup, refreshed and cached catalog reads with 10 bundled provider models, 61 focused test cases, and exact-head review and CI.

- **Bug:** Valid provider settings without custom model rows could crash Gateway startup when a configured channel was automatically enabled.
- **Fix:** Catalog composition accepts omitted custom rows while preserving inherited models and current request settings.

| Before | After |
| --- | --- |
| Gateway exits with a model-list iteration error. | Gateway reaches readiness and retains the provider catalog. |

Provider-level `enabled` remains invalid configuration. Synthetic credentials and disabled channel connections limit this proof to startup and catalog behavior.

PR #144616, opened 2026-09-11
fix: reuse the bound Codex thread when compacting chats

### LINKED ISSUE #144612: Compaction can open a competing writer for an already-loaded Codex chat
## Problem

Compacting an already-loaded Codex-backed chat can fail with "thread already has an active writer" when the ordinary turn and the compaction request select different startup cache keys. The chat's native client remains live and is recorded in the private thread binding, but default compaction obtains another client and attempts to load the same thread there.

## Reproduction and expected behavior

A source regression uses the real app-server client, shared registry, private thread binding, thread leases and compactor with an in-memory native transport. Starting and retaining the thread first, then compacting through an explicit factory returning that owner, succeeds. The default factory path instead attempts a second writer and fails. Repeated default compaction should use the recorded live owner and release only the lease acquired for that request.

## Proposed repair

Reuse the existing retain-by-instance-ID operation for the binding's live client on default compaction. Preserve explicit client-factory precedence and the existing fallback when the owner is unavailable. Retain the account, sandbox, binding and thread-ownership checks. No schema or configuration change is needed.

The regression fails before the fix, passes afterward, and fails again when only the production change is reverted. The pinned native Codex dependency is 0.153.4; its compaction processor loads an existing thread and its resume tests explicitly reject competing writers.

JamesY023:
Closes #144612

## What Problem This Solves

Fixes an issue where users running `/compact` on a Codex-backed chat could receive “thread already has an active writer” even though the chat's existing app-server was still available. Compaction could select a second app-server when its startup cache key differed from the one used by the ordinary turn.

## Why This Change Was Made

Default compaction now retains the live app-server identified by the private thread binding, using the existing instance-ID lease lookup. Its own lease is released afterward. Explicit client factories retain precedence, and the existing startup path remains available when the recorded client is unavailable. Binding, sandbox, and account checks remain in place.

## User Impact

`/compact` can complete on the already-loaded native thread, including repeated requests, without opening a competing writer or consuming another caller's lease.

## Evidence

- The regression uses the real app-server client, shared registry, binding store, thread leases, and compactor with an in-memory native transport. An explicit-owner control succeeds before the fix; default compaction fails with the active-writer error.
- `pnpm test extensions/codex/src/app-server/compact.test.ts extensions/codex/src/app-server/shared-client.test.ts extensions/codex/src/app-server/thread-ownership.test.ts`: all 180 tests pass.
- Reverting only the production fix reproduces the active-writer failure. Restoring it passes all 60 compaction tests. The regression also verifies repeated compaction, explicit-factory precedence, no duplicate startup/resume, and preservation of the original caller's lease.
- Exact native dependency inspected: `openai/codex` tag `rust-v0.153.4`, commit `3d2ee51ca2d5db578f328aa75e20aa22c0197c9a`. Its [compaction processor](https://github.com/openai/codex/blob/3d2ee51ca2d5db578f328aa75e20aa22c0197c9a/codex-rs/app-server/src/request_processors/thread_processor.rs#L2350) loads the existing thread; its [resume regression](https://github.com/openai/codex/blob/3d2ee51ca2d5db578f328aa75e20aa22c0197c9a/codex-rs/app-server/tests/suite/v2/thread_resume.rs#L375) checks rejection of a competing writer.
- Production and test extension TypeScript checks, focused oxlint, formatter checks, and `git diff --check` pass.
- Independent upstream autoreview: scoped-clean at P1, with the shared-client lease implementation and pinned native contract supplied.
- Validation ran on Node 26.8.1 and pnpm 12.3.4. No live model/API compaction or full repository build/test run was performed.

### DIFF
--- a/extensions/codex/src/app-server/compact.test.ts
+++ b/extensions/codex/src/app-server/compact.test.ts
@@ -14,9 +14,12 @@ import {
   ensureCodexAppServerClientRuntime,
   retainCodexAppServerLiveThread,
 } from "./client-runtime.js";
-import { CodexAppServerRpcError, type CodexAppServerClient } from "./client.js";
+import { CodexAppServerClient, CodexAppServerRpcError } from "./client.js";
 import { maybeCompactCodexAppServerSession as maybeCompactCodexAppServerSessionImpl } from "./compact.js";
-import { resolveCodexSupervisionAppServerRuntimeOptions } from "./config.js";
+import {
+  resolveCodexAppServerRuntimeOptions,
+  resolveCodexSupervisionAppServerRuntimeOptions,
+} from "./config.js";
 import { buildCodexAppServerConnectionFingerprint } from "./plugin-app-cache-key.js";
 import type { CodexServerNotification } from "./protocol.js";
 import { createSandboxContext } from "./sandbox-exec-server.test-helpers.js";
@@ -31,7 +34,11 @@ import {
   testCodexAppServerBindingStore,
   writeCodexAppServerBinding,
 } from "./session-binding.test-helpers.js";
-import type { CodexAppServerClientFactory } from "./shared-client.js";
+import {
+  getLeasedSharedCodexAppServerClient,
+  releaseLeasedSharedCodexAppServerClient,
+  type CodexAppServerClientFactory,
+} from "./shared-client.js";
 import { createClientHarness } from "./test-support.js";
 import { withCodexAppServerThreadMutation } from "./thread-ownership.js";
 import { CODEX_APP_SERVER_VERSION } from "./version.js";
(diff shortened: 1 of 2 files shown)


compoodment:
This report and PR already own the exact client-selection bug, so I did not open a duplicate issue or PR.

I reproduced the failure on a current multi-client deployment: a fresh native thread completed its turn on app-server client A; manual compaction selected client B; `thread/resume` was rejected with `thread-store conflict: thread already has an active writer`; `thread/compact/start` was never reached. The persisted binding already contained client A's physical `clientId`.

I prepared an updated implementation on `compoodment/openclaw`, commit `e19ce101085f6c5bc9bc5d580e5614435aedf726`:

- native compaction first retains the persisted physical owner when it is live;
- generic pooled acquisition remains the fallback when that owner is gone;
- explicit injected client factories retain precedence;
- the request-scoped owner/fallback lease is released exactly once.

Validation on the current-main-based checkout:

- existing native compaction suite: **59/59**;
- owner-selection helper regression: **3/3**;
- production extension types: passed;
- extension-test types: passed;
- line-cap ratchet: passed;
- independent clean-checkout extension lint: passed (only two pre-existing max-lines warnings in untouched files).

The implementation is intentionally separate from the sender-provenance work in #150490; this lifecycle failure occurs before native compaction receives the message text.

compoodment:
@clawsweeper re-review

compoodment:
Independent live runtime corroboration for the bound-client compaction behavior:

A fresh Codex-native session completed a real compaction without the competing-writer failure. Redacted gateway receipt:

```text
2026-09-17T08:13:01.519+02:00 started codex app-server compaction
2026-09-17T08:13:08.128+02:00 completed codex app-server compaction
2026-09-17T08:13:08.217+02:00 native_slash_command_fast_path outcome=completed
2026-09-17T08:14:35.378+02:00 context-engine projection decision
2026-09-17T08:14:36.091+02:00 context-engine thread binding written
2026-09-17T08:14:48.399+02:00 assistant_output_started
```

No active-writer error occurred. Session, thread, client, channel, host, and account identifiers are omitted.

This is one live compaction receipt, not a repeated-compaction soak, and is offered as runtime corroboration of the behavior—not as a claim that the current PR head itself was deployed.

PR #140989, opened 2026-09-07
fix(backup): handle AppleDouble metadata without losing SQLite validation

ericcaiwx-star:
## What Problem This Solves

Fixes an issue where `openclaw backup create` failed before producing an archive when state contained AppleDouble metadata named `._cron.sqlite`.

## Why This Change Was Made

Exclude signature-confirmed `._*.sqlite` metadata after the existing state/agent inventory checks. The probe uses the existing safe file opener and closes its descriptor after every read outcome. Real databases and hardlink aliases retain verified snapshots, compaction, and canonical sanitization.

## User Impact

Backups can include state copied from macOS without treating AppleDouble metadata as a database. Ordinary workspace files remain included, and corrupt SQLite databases still stop backup creation.

## Evidence

- Pinned-main public `backup create --verify --json` failed actual SQLite validation for a 163-byte metadata fixture and produced no archive.
- Independent candidate public create and separate verify passed. Extracted archives exclude state/agent metadata, retain workspace metadata and ordinary files, and preserve real same-name databases and canonical aliases with their ownership and sanitization.
- 63 distinct focused backup and verifier cases passed, including real hardlinks, canonical WAL data, corruption, inventory scope, replacement symlinks, and descriptor cleanup.
- Native formatting, lint, source guards, and the normal commit hook passed. Independent source review is clean.

The exploratory generic hardlink/live-WAL case failed identically on pinned main and the candidate; it remains tracked separately in #144585. The passing canonical and closed-generic controls do not turn that existing failure into a pass.

Proof uses synthetic files on Linux; it does not claim a macOS filesystem test. All four contributor commits are retained.

Closes #140193.

PR #144624, opened 2026-09-11
fix(backup): preserve committed data in SQLite hardlink aliases

### LINKED ISSUE #144585: Backup hardlink aliases of generic SQLite databases can omit uncheckpointed WAL data
## Problem

`openclaw backup create --verify` can publish a successful archive that omits uncheckpointed SQLite data from a hardlink alias of a noncanonical database. Separate `openclaw backup verify` also accepts that archive.

This was observed independently while validating #140989. It reproduces on pinned main `1c1f85474f5b11b281ab446d568b86a329dee22c` and on candidate `04cb0a6e9aea132be35e437308809491c04aae23`, with identical extracted database bytes.

## Reproduction

1. Use isolated synthetic state, workspace, home, and backup output directories. Keep the output outside the state and workspace. Point `OPENCLAW_STATE_DIR` and `OPENCLAW_CONFIG_PATH` at that state and its minimal config.
2. Open `<state>/data/._cron.sqlite` with Node's `DatabaseSync`. Execute the SQL below and keep that connection open.
3. Create `<state>/data/._alias.sqlite` with `fs.linkSync`, pointing to the same database inode.
4. While the original connection remains open, run `openclaw backup create --verify --output <output> --json`.
5. Run `openclaw backup verify <archive> --json`, extract the archive, and query both database entries.

```sql
PRAGMA journal_mode = WAL;
PRAGMA wal_autocheckpoint = 0;
CREATE TABLE acceptance_probe (id INTEGER PRIMARY KEY, value TEXT NOT NULL);
INSERT INTO acceptance_probe VALUES (1, 'committed-in-wal');
INSERT INTO acceptance_probe VALUES (2, 'deleted-secret-marker');
DELETE FROM acceptance_probe WHERE id = 2;
```

Both live source names have the same device/inode and link count 2. The original WAL contains 20,632 bytes during the backup.

## Expected and actual

Expected: both archive entries preserve the committed row, or creation explicitly refuses a snapshot whose WAL ownership cannot be resolved safely.

Actual: create and separate verify both exit 0. The original archive entry is 8,192 bytes and contains the committed row. The alias entry is 4,096 bytes with an empty SQLite schema; querying the table fails with `no such table: acceptance_probe`.

Canonical global/agent aliases retain their expected WAL data and sanitization in the same public tests. Generic hardlinks whose data is checkpointed also pass. The remaining defect is the generic alias's live-WAL path: the existing planner only matches aliases to canonical global/agent source identities, so generic aliases are opened under their own pathname and WAL basename. The AppleDouble repair leaves that existing source-binding behavior unchanged.

Observed with Node 24.19.0 on Linux using synthetic state. No production data was involved.

obviyus:
Closes #144585. Related: #140989.

## What Problem This Solves

A verified backup of a live database with hardlink aliases could omit committed write-ahead-log data from the alias entries.

## Why This Change Was Made

Generic database aliases now share one verified snapshot from the unambiguous journal-owning name, with a separate regular archive entry for each name. Capture preserves live writer locks and refuses competing journals, incomplete hardlink coverage, active rollback journals, and changed source identities.

## User Impact

Both names retain the same committed content. Every hardlink must be included as a database file under the state directory or a configured agent root, including aliases of closed databases. Uncertain ownership fails explicitly without publishing an archive. Canonical database sanitization remains intact.

## Evidence

Before the fix, backup creation and verification succeeded but an extracted alias lacked the committed table. Corrected public commands preserved the data in both filename orders, and each extracted database passed integrity checks. Eighteen new regressions and 25 existing backup cases passed. Nine public scenarios and ten combined backup scenarios passed with synthetic Linux files, including refusal controls and canonical database sanitization. Writer-lock regressions rejected the earlier capture implementation.

### DIFF
--- a/docs/cli/backup.md
+++ b/docs/cli/backup.md
@@ -342,6 +342,17 @@ SQLite capabilities also fails closed rather than falling back to a direct file
 copy. Other workspace SQLite files outside configured agent roots remain raw
 workspace files and do not receive the SQLite snapshot or compaction guarantee.
 
+Hardlinks to a generic SQLite database share one captured image, stored as a
+separate regular archive entry for each name. Every hardlink must be an included
+`.sqlite` file under the state directory or a configured agent root. If exactly
+one name has a nonempty write-ahead log (WAL), that
+name supplies the committed data. Closed databases without a nonempty WAL remain
+supported. Multiple nonempty WALs, a nonempty rollback journal, or hardlinks
+outside the backup inventory cause an explicit refusal with no archive. Close
+the database writers cleanly and include every hardlink in those roots before retrying.
+Canonical OpenClaw database aliases retain their existing owner validation and
+sanitization.
+
 Installed plugin source and manifest files under the state directory's `extensions/` tree are included, but their nested `node_modules/` dependency trees are skipped as rebuildable install artifacts. After restoring an archive, use `openclaw plugins update <id>` or reinstall with `openclaw plugins install <spec> --force` if a restored plugin reports missing dependencies.
 
 The state directory's `plugin-skills/` root is a generated, OpenClaw-owned symlink index, not authoritative state. Backup creation reports and omits that root because its absolute targets are specific to the source installation. After activating restored state, run `openclaw skills list` or start an agent session to rebuild the links from current plugin metadata.
--- a/src/infra/backup-create.sqlite-hardlinks.test.ts
+++ b/src/infra/backup-create.sqlite-hardlinks.test.ts
@@ -1,0 +1,396 @@
+import fs from "node:fs/promises";
+import path from "node:path";
+import type { DatabaseSync } from "node:sqlite";
+import * as tar from "tar";
+import { describe, expect, it, vi } from "vitest";
+import { verifyBackupArchive } from "../commands/backup-verify.js";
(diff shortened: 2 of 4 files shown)


obviyus:
Landed as 0b48b3e78c91576a2ad9d02654f474d694d3fae6.

**Bug:** A live SQLite hardlink alias could omit committed WAL data even when backup creation and verification reported success.

**Fix:** Capture one verified image from the unambiguous journal owner and store each alias as a regular archive entry. Refuse incomplete or ambiguous ownership without publishing an archive. Preserve canonical validation and sanitization.

| Before | After |
| --- | --- |
| Original entry has the committed row; alias reports `no such table` | Both regular entries retain the committed schema and row |
| Archive verification can miss the omitted source data | Public create, separate verify, and extracted SQL checks agree |

**Proof:** 18 new regressions and 25 focused existing cases passed. Nine independent public cases passed on the PR head; ten combined cases passed with the landed AppleDouble repair. The landed backup source matches that tested combination. Exact-head local review reported no findings, and [required CI passed](https://github.com/openclaw/openclaw/actions/runs/34559457161).

Public proof uses synthetic files on Linux. Complete hardlink inventory remains required, including closed databases with outside aliases, as documented in the accepted compatibility boundary.

PR #144629, opened 2026-09-11
fix(discord): honor persona text limits and preserve reply receipts

steipete:
## What Problem This Solves

Fixes an issue where thread-bound Discord persona replies could exceed a caller's character limit after mention expansion, even though ordinary bot replies honored that limit. It also fixes missing reply-target metadata in the direct webhook delivery receipts and progress callbacks for chunked replies.

## Why This Change Was Made

Both Discord transports now consume the character limit normalized by their existing shared text-preparation owner. The persona adapter forwards the caller's limit through to final webhook chunking, after formatting and mention expansion.

Webhook results now pass each physical chunk's reply reference to the canonical receipt constructor using its current contract. This replaces an ignored legacy property and preserves first-only, all-chunk, and absent reply scope in returned receipts and progress callbacks.

## User Impact

Persona replies honor the caller's requested character cap after rendering, with the same normalization as bot replies. Their delivery receipts accurately retain the reply target selected for each chunk. Existing defaults, line limits, and requested webhook reply fields are preserved.

## Evidence

- The character-cap regression failed on the original webhook path while the bot control and 19 sibling cases passed. The repaired bot and webhook paths reconstruct the exact expanded message and keep every wire chunk within the 500-character cap.
- Direct webhook receipt tests failed before the receipt repair for first-only and all-chunk replies, then passed with matching aggregate and progress-callback metadata. The no-reply control remains covered.
- All 297 selected tests passed across six files, including durable delivery through the real local HTTP adapters, webhook activity, chunking, and ordinary bot sends.
- All 25 checks selected by the changed-file gate passed, including extension and test type checks, lint, formatting, and plugin boundaries. A later documentation-only wording correction received its own formatting and diff checks.

Independent review completed two clean passes through P2 on the final source.

Transport proof uses synthetic loopback HTTP servers. It verifies requested webhook fields and receipt metadata; it does not establish whether Discord applies the existing webhook reply-reference field. No live Discord or Gateway send was run.

### DIFF
--- a/docs/channels/discord/messaging.md
+++ b/docs/channels/discord/messaging.md
@@ -14,6 +14,7 @@ How inbound and outbound Discord messages are routed, formatted, acknowledged, a
 - Gateway owns the Discord connection.
 - Reply routing is deterministic: Discord inbound replies back to Discord.
 - Bot replies and thread-bound persona replies share Markdown formatting, including CommonMark bold and configured table conversion.
+- Both delivery paths reapply caller-supplied character limits after formatting and mention expansion.
 - Forwarded message snapshots reach the agent together with any accompanying caption. Forwarded text is not treated as a typed command; command classification uses only the sender’s own message text.
 - Discord guild/channel metadata is added to the model prompt as untrusted context, not as a user-visible reply prefix. If a model copies that envelope back, OpenClaw strips the copied metadata from outbound replies and from future replay context.
 - By default (`session.dmScope=main`), direct chats share the agent main session (`agent:main:main`).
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b1010579aa5712041aa353558943fed7f25bddd9](https://github.com/openclaw/openclaw/pull/144629/commits/b1010579aa5712041aa353558943fed7f25bddd9)
- Landed commit: [7890ce92cf1b2167420f8af8c046cae41dc5b3da](https://github.com/openclaw/openclaw/commit/7890ce92cf1b2167420f8af8c046cae41dc5b3da)

PR #144632, opened 2026-09-11
fix: keep streamed code examples from redirecting replies

steipete:
## What Problem This Solves

Fixes an issue where a streamed inline-code example containing a reply tag could attach that literal tag as a real reply target before the message finished. A later chunk could also lose the literal marker when the code opener had already been drained. Terminal-only delivery had related inconsistencies between prepared visible text and directive metadata.

## Why This Change Was Made

The subscriber now takes reply-target metadata from the same cumulative parse that owns visible text. Prepared chunks keep their literal bytes instead of parsing them again without the preceding code context. Terminal fallback producers prepare their text before passing it to that path, while final media stays with its existing owner.

Standalone message-tool JSON can introduce control syntax during decoding. Its decoded body now passes through the existing phase-aware visibility extractor before directive parsing, preserving code examples, selected final-answer content, and empty sanitized results.

## User Impact

Reply tags shown in streamed code examples remain literal and do not become reply targets merely because a provider or delivery chunk splits the example. Decoded standalone message text receives the same visibility preparation as other delivered text.

## Evidence

- The original subscriber failed two code-example regressions across provider-delta and physical-block boundaries; unaffected controls passed.
- The decoded-text cases reproduced three failures with five passing controls on both the original main source and an intermediate candidate. This was an existing defect, despite an initial review attributing it to the candidate.
- All 334 selected tests passed across 12 files, with no filtered cases. Coverage includes terminal fallback, literal code, phase selection, chunk boundaries, final media, and nine decoded-body cases.
- All 29 selected changed-file checks passed. Final independent review through P2 found no remaining actionable findings.

Proof uses the registered subscriber with synthetic provider events; no live provider or channel send was run. Per-chunk audio metadata and the separate raw-block accumulator retain their existing behavior and remain separate follow-ups. No transcript, config, persistence, or dependency format changes.

### DIFF
--- a/src/agents/embedded-agent-subscribe.handlers.messages.lifecycle.ts
+++ b/src/agents/embedded-agent-subscribe.handlers.messages.lifecycle.ts
@@ -34,6 +34,7 @@ import { warnIfAssistantEmittedSuspiciousText } from "./embedded-agent-subscribe
 import {
   createThinkingTagStreamState,
   extractAssistantThinking,
+  extractAssistantVisibleText,
   extractEmbeddedAssistantText,
   extractThinkingFromTaggedText,
   promoteThinkingTagsToBlocks,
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [ff074f4a1b2d745b4109a4443972e0b882b53ec1](https://github.com/openclaw/openclaw/pull/144632/commits/ff074f4a1b2d745b4109a4443972e0b882b53ec1)
- Landed commit: [9dda26a44b8cd3fe723340a5d3dfd1f69130fd03](https://github.com/openclaw/openclaw/commit/9dda26a44b8cd3fe723340a5d3dfd1f69130fd03)

PR #144640, opened 2026-09-11
fix(gateway): worker workspace downloads continue after credential revocation

eleqtrizit:
## What Problem This Solves

Fixes an issue where an attached worker that legitimately holds a node workspace-transfer bearer could keep receiving an already-admitted workspace blob, manifest, or pack response after its durable environment credential was permanently revoked. The gateway correctly denied *new* admissions for the revoked credential, but the in-flight response kept streaming: the stream's abort signal observed only the owner/tunnel and capability aborts, never the durable credential state, so any revocation path that did not immediately stop the tunnel left the response unfenced until it ended on its own.

## Why This Change Was Made

The node workspace transfer service now enforces one invariant centrally: a permanent durable credential revocation fences every in-flight transfer belonging to that owner, immediately and regardless of which lifecycle path performed the revocation.

- `src/gateway/worker-environments/store.ts` — `revokeEnvironmentCredential` accepts an explicit `{ fenceWorkspaceTransfers }` opt and exposes `onCredentialRevoked` listener registration. Only fencing (permanent) revocations notify; rotation-style revocations followed by a re-mint (device reconcile) deliberately do not, so routine reconcile never tears down healthy transfers.
- `src/gateway/worker-environments/node-workspace-transfer-service.ts` — new `fenceEnvironment(environmentId, reason?)` synchronously aborts the owner's transfer context (aborting every in-flight response and capability via the existing `authorizationSignal` join), then settles cleanup asynchronously through the existing close path.
- `src/gateway/worker-environments/provider-owner-lifecycle.ts` — teardown revocations (`stopOwner`) opt into the fence, so the stream is cut at revocation time instead of whenever the tunnel stop completes.
- `src/gateway/server-worker-environment-startup.ts` — wires the transfer service to fencing revocations from the store.

The pre-header dynamic authorization check is retained as defense in depth. No config surface, no default flips, no persistence-format changes.

## User Impact

When an operator destroys, fails, or otherwise permanently revokes a worker environment's credential, an in-flight workspace download now ends immediately instead of delivering workspace bytes after the credential is gone. Ordinary (non-revoked) transfers behave byte-identically to before, and device-reconcile credential rotation no longer risks tearing down healthy in-flight transfers because it never fences.

## Evidence

| Path | Mode | Outcome | Bytes after fence |
|---|---|---|---|
| Blob download (`/worker-transfer/v1/.../blobs/:sha256`) | Permanent revocation mid-stream (8 MiB body) | Response aborted before full delivery; same bearer denied for new admission | 0 further authorized delivery (stream destroyed; client received < payload) |
| Blob/manifest/pack download | No revocation (control) | Completes normally, byte-identical behavior | n/a |
| Device-reconcile bulk revoke (rotation) | Revoke then re-mint under same owner epoch | No fence notification; in-flight transfers unaffected | n/a |
| Destroy/failed teardown | `stopOwner` revoke | Fence fires synchronously at revoke time, before tunnel stop returns | 0 |

- `pnpm run tsgo:core` — clean.
- `src/gateway/worker-environments/node-workspace-transfer-revocation.test.ts` — 3 new tests: fence aborts capability signals and denies new admissions; a fencing revocation stops an in-flight blob response; existing upload/revocation matrix unchanged-green.
- `src/gateway/worker-environments/store.test.ts` — new test: revocation listeners fire only when transfers must fence.
- Full `src/gateway/worker-environments/` suite: 176 passed / 1 skipped; one timing test (`workspace-result-finalize.test.ts`) flaked under full-suite parallel load and passes in isolation on this exact head.

## Testing

- New: store listener semantics, service `fenceEnvironment` (signal + currentness + admission denial), and an HTTP-level mid-stream fence regression with a large blob.
- Unchanged-green: `node-workspace-transfer-revocation.test.ts` upload cancellation and attachment revocation matrices, `store.test.ts`, `server-worker-environment-startup.test.ts`, `store-runtime-refresh.test.ts`, `credential-broker.test.ts`.

### DIFF
--- a/src/gateway/server-worker-environment-startup.ts
+++ b/src/gateway/server-worker-environment-startup.ts
@@ -285,6 +285,12 @@ export async function createGatewayWorkerEnvironmentRuntime(params: {
     getOwner: (environmentId) => params.startup.store.getTransferOwner(environmentId),
   });
   await nodeWorkspaceTransfer.initialize();
+  // Permanent credential revocation fences every in-flight workspace transfer for the
+  // owner immediately. Rotation-style revocations (device reconcile re-mints) do not
+  // notify, so routine reconcile never tears down healthy transfer contexts.
+  params.startup.store.onCredentialRevoked((environmentId) => {
+    nodeWorkspaceTransfer.fenceEnvironment(environmentId);
+  });
   const gatewayDeviceId = loadOrCreateProcessDeviceIdentity().deviceId;
   const nodeWorkerGatewayNamespace = resolveNodeWorkerGatewayNamespace(gatewayDeviceId);
   const nodeWorkerTunnelManager = createNodeWorkerTunnelManager({
--- a/src/gateway/worker-environments/node-workspace-transfer-revocation.test.ts
+++ b/src/gateway/worker-environments/node-workspace-transfer-revocation.test.ts
@@ -1,6 +1,7 @@
 import fsSync from "node:fs";
 import fs from "node:fs/promises";
 import { createServer, type IncomingMessage } from "node:http";
+import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { createDeferred, withTestTimeout } from "../../../test/helpers/promise.js";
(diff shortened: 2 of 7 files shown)


eleqtrizit:
@clawsweeper re-review

eleqtrizit:
@clawsweeper re-review

eleqtrizit:
@clawsweeper re-review

eleqtrizit:
## Merge record

**Summary:** Fences in-flight worker workspace transfers at the owning boundary
when a worker credential is durably revoked: `revokeEnvironmentCredential` gains an
explicit `fenceWorkspaceTransfers` option, the store notifies listeners, and
`fenceEnvironment` aborts every in-flight transfer for the owner instantly — closing
the window where blob/manifest/pack streams could continue after durable revocation.
Rotation-style re-mints stay unfenced; teardown opts in explicitly. Adds
store-driven regression tests (~449 lines across 6 files).

**Decision Reason for Merge:** "Merged because all verification gates passed — the
independent merge review returned MERGE AS-IS and every CI check is green on the
final head — and the operator explicitly requested the merge."

**Compatibility Checks Done:** `security-compat-check` returned SAFE both pre-PR and
pre-merge (additive APIs, no config/schema/default changes); the independent merge
review returned MERGE AS-IS; the review bot completed 3 rounds (proof: sufficient, no
contributor-facing blockers); full CI green on the final head `7b3e32e5cc8`.

PR #144641, opened 2026-09-11
fix(memory): recover missing full-text index rows

jinliyl:
Related: #115775

## What Problem This Solves

Fixes an issue where users with a populated but incomplete memory full-text index could receive silently missing search results. Ordinary schema reconciliation only rebuilt a completely empty index, so one or more missing derived rows survived startup.

## Why This Change Was Made

The existing Memory Host SDK FTS owner now compares canonical and derived row counts inside its existing savepoint and invokes the existing canonical rebuild operation only when they diverge. This is routine maintenance of the derived-index contract accepted in #115775; it adds no schema, version, configuration, migration, or public API.

## User Impact

Memory search recovers missing FTS rows during normal schema reconciliation while leaving complete indexes and canonical memory rows unchanged.

## Evidence

- Regression proof before the fix: after deleting one row from a populated FTS table, schema reconciliation left only `kept` and failed to restore/search `missing`.
- `node scripts/run-vitest.mjs packages/memory-host-sdk/src/host/memory-schema-fts-reconcile.test.ts packages/memory-host-sdk/src/host/memory-schema.test.ts --maxWorkers=1` (33 passed across both selected projects)
- Changed-path core gates passed: formatting, assertion ratchet, package/plugin boundaries, core and core-test typechecks, dead-export scan, lint, database-first guards, and runtime import-cycle checks.
- Initial fresh local autoreview: scoped clean, confidence 0.94. Follow-up review after the CI-driven sibling assertion update: scoped clean, confidence 0.98.
- CI's first revision exposed one attributable stale assertion in `memory-schema.test.ts`: it expected a newly canonical `chunk-2` to remain absent from FTS. The assertion now requires both canonical chunks after reconciliation, and the full sibling test file passes.

### Isolated real CLI proof

Used a fresh temporary state/config/workspace with two synthetic memory files and the normal supported CLI. `pnpm openclaw memory index --agent main --force` indexed both files. I then removed only the derived FTS row for `memory/beta.md`, producing this persisted precondition:

```json
{
  "canonical": { "n": 2 },
  "derived": { "n": 1 },
  "missingMatch": []
}
```

Running normal initialization through `pnpm openclaw memory search --agent main --query cobalt --json` returned `memory/beta.md` with its synthetic `cobalt-lantern-482` snippet. Read-only inspection after the command showed:

```json
{
  "canonical": { "n": 2 },
  "derived": { "n": 2 },
  "match": [{ "path": "memory/beta.md" }]
}
```

### Data-model and upgrade compatibility

The proof starts from an already persisted current-format agent database, removes a reconstructible derived row to model the affected existing state, and reopens it through the patched normal CLI path. The repair changes no table definition, columns, schema version, canonical rows, configuration, migration, downgrade behavior, or public API. It only rebuilds the existing derived FTS table from the existing canonical chunks when their cardinalities prove divergence. Equal-count databases retain their existing index.

### DIFF
--- a/packages/memory-host-sdk/src/host/memory-schema-fts-reconcile.test.ts
+++ b/packages/memory-host-sdk/src/host/memory-schema-fts-reconcile.test.ts
@@ -4,6 +4,44 @@ import { describe, expect, it } from "vitest";
 import { ensureMemoryIndexSchema } from "./memory-schema.js";
 
 describe("memory FTS schema reconciliation", () => {
+  it("rebuilds a populated body index when its row count diverges", () => {
+    const db = new DatabaseSync(":memory:");
+    try {
+      ensureMemoryIndexSchema({ db, cacheEnabled: false, ftsEnabled: false });
+      db.exec(`
+        INSERT INTO memory_index_chunks
+          (id, path, source, start_line, end_line, hash, model, text, embedding, updated_at)
+        VALUES
+          ('kept', 'memory/kept.md', 'memory', 1, 1, 'kept-hash', 'model', 'kept body', '[]', 1),
+          ('missing', 'memory/missing.md', 'memory', 1, 1, 'missing-hash', 'model', 'missing body', '[]', 1);
+      `);
+      expect(
+        ensureMemoryIndexSchema({ db, cacheEnabled: false, ftsEnabled: true }).ftsAvailable,
+      ).toBe(true);
+      db.exec("DELETE FROM memory_index_chunks_fts WHERE id = 'missing'");
+
+      expect(
+        ensureMemoryIndexSchema({ db, cacheEnabled: false, ftsEnabled: true }).ftsAvailable,
+      ).toBe(true);
+
+      expect(db.prepare("SELECT id FROM memory_index_chunks_fts ORDER BY id").all()).toEqual([
+        { id: "kept" },
+        { id: "missing" },
+      ]);
+      expect(
+        db
+          .prepare("SELECT id FROM memory_index_chunks_fts WHERE memory_index_chunks_fts MATCH ?")
+          .all("missing"),
+      ).toEqual([{ id: "missing" }]);
+      expect(db.prepare("SELECT id FROM memory_index_chunks ORDER BY id").all()).toEqual([
+        { id: "kept" },
+        { id: "missing" },
+      ]);
+    } finally {
+      db.close();
+    }
+  });
+
   it("rebuilds body and path indexes when their tokenizer changes", () => {
     const db = new DatabaseSync(":memory:");
     try {
--- a/packages/memory-host-sdk/src/host/memory-schema-fts.ts
+++ b/packages/memory-host-sdk/src/host/memory-schema-fts.ts
@@ -175,16 +175,19 @@ export function ensureMemoryChunkFtsSchema(params: {
         `  end_line UNINDEXED\n` +
         `${params.tokenizeClause});`,
     );
-    // Empty derived tables are rebuilt from canonical chunks; populated,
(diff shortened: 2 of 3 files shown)


jinliyl:
Thanks for the review. The requested compatibility evidence is already included in the PR body under **Data-model and upgrade compatibility** and **Isolated real CLI proof**.

This patch does not change the persisted data model: it adds no table, column, index option, schema version, migration, configuration, or public API. It only changes reconciliation of the existing fully derived `memory_index_chunks_fts` rows, using the existing rebuild operation and savepoint accepted in #115775.

The persisted-state proof starts with an existing current-format agent database containing 2 canonical rows and 1 derived FTS row, reopens it through the supported `pnpm openclaw memory search` CLI path, and ends with 2 canonical rows and 2 derived rows plus the recovered `MATCH` result. Canonical data remains unchanged. That exercises the upgrade/reopen path relevant to this repair without requiring a schema migration.

I also rechecked against current `main` (`2e47020ebf87`): the affected owner and tests are unchanged from the revision ClawSweeper reviewed, and the PR merges cleanly in a local merge-tree check.

Could the generic data-model compatibility checkbox be cleared based on that evidence?

@clawsweeper re-review

steipete:
Maintainer disposition of the SQLite-table warning at head 1f7a83f: this is not a persisted-format or schema change. The patch keeps the existing reconciliation owner in `memory-schema-fts.ts` and widens its trigger from "derived table completely empty" to "derived row count differs from canonical chunks", rebuilding the full-text index solely from canonical chunks inside the same savepoint. Stored formats, schema version, and user data are untouched; the index is derived and reconstructible, which is the design accepted in #115775. Upgrade compatibility is the ordinary reopen path: an incomplete index on an existing database is exactly the state this repair targets, and the supplied persisted-database CLI evidence exercises it. Warning accepted as non-material; no migration is required.

@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [1f7a83f3288f7743864d9e989a9b1a5ce04ee7c8](https://github.com/openclaw/openclaw/pull/144641/commits/1f7a83f3288f7743864d9e989a9b1a5ce04ee7c8)
- Landed commit: [2575d0d6356535a58d3a9798f08000caac856d0e](https://github.com/openclaw/openclaw/commit/2575d0d6356535a58d3a9798f08000caac856d0e)

fuller-stack-dev:
@jinliyl Narrow post-merge regression notice for the memory/FTS owner.

`extensions/memory-core/src/memory/manager-temporal-ranking.test.ts`, case `through FTS unavailable`, now fails at line 156: expected `fts.available=false`, received `true`.

- #144715's owner reproduced this on unmodified main `1b2b06ed2fc7680648a1a554c100b887a1de39dd`, without the lease patch. I checked the saved baseline receipt and failure log; I did not rerun it.
- The same failure now blocks exact head `a86c71d9ba6172b753bb0557fc6f57fee45cf7d2`: [failed CI job](https://github.com/openclaw/openclaw/actions/runs/34568807852/job/103166460763).
- #144717 independently documents reproduction on this PR's merge commit `2575d0d6356535a58d3a9798f08000caac856d0e`, with five neighboring cases passing, and all six passing on `ada8b5f65083`. Its diagnosis is that the fixture's injected view remains readable under the new row-count reconciliation, so it no longer simulates FTS unavailability.

Baseline reproducer:
```sh
node scripts/run-vitest.mjs extensions/memory-core/src/memory/manager-temporal-ranking.test.ts -t "through FTS unavailable"
```

Is an existing canonical follow-up already owned? Relaying here to avoid competing memory fixes in the lease or Wear PRs. This is not a request to waive their exact-head CI, restart automation, or weaken the fallback assertion.

PR #144642, opened 2026-09-11
improve: simplify sidebar session tree projection

steipete:
## What Problem This Solves

The Control UI rebuilds nested session summaries as its sidebar updates. This is a maintenance cleanup of that work, preserving the existing hierarchy, running indicators, attention and conflict badges. It does not address a reproduced user-visible failure or promise a noticeable speedup.

## Why This Change Was Made

Reuse one path-local ancestor Set per root instead of copying it at every recursive frame, and compute the five child summaries in one traversal. Each frame releases only membership it introduced, preserving cycle handling and shared descendants. Keep callback order, attention tie precedence, archived/category placement, running-subagent fallback and conflict arithmetic unchanged.

The existing projection owner remains responsible for the result. No cache, persistent state, configuration, public API or rendering behavior is added. Production grows by one formatted line; preservation tests add 136 lines.

## User Impact

No intended visible change. The sidebar continues to show the same rows, links and indicators, while the projection has fewer repeated traversals and ancestor copies. Synthetic before/after captures are byte-identical.

## Evidence

- The final navigation suite passed 40 tests on both the original and modified projector, including the same added preservation cases. Three complete candidate UI suites passed 426 tests across navigation, projection and mounted sidebar behavior. That broader run preceded two test-only corrections; the final navigation suite was rerun on both projectors afterward.
- All 20 selected changed checks passed, including core-test and UI typechecks, type-aware lint, formatting, i18n and dead-export checks. The existing Chromium row-identity test passed. Independent Codex review found no actionable P0–P2 issues.
- A fixed baseline/candidate/candidate/baseline Chromium cohort made 364 direct projector calls and 252 mounted publications. Complete DTO/property/reference, callback-order, machine identity, input, root membership, DOM/link/badge/toggle, row reuse and update-completion checks matched between versions. The filtered All-status scenario refreshed its actual managed-list owner before awaiting the UI update; its timing includes that synthetic list request.

The browser clock advanced in 0.1 ms steps, and 159 of 364 direct samples recorded zero. These observations do **not** support a defensible percentage speedup claim. Mounted steady medians were also small and order-sensitive:

| Mounted scenario | First order: baseline → candidate | Reverse order: baseline → candidate |
| --- | --- | --- |
| Empty | 0.3 → 0.4 ms | 0.3 → 0.3 ms |
| Single | 0.3 → 0.3 ms | 0.3 → 0.4 ms |
| Ordinary shallow tree | 0.5 → 0.6 ms | 0.5 → 0.5 ms |
| Broad tree | 0.9 → 0.9 ms | 0.9 → 0.8 ms |
| Bounded deep stress case | 0.8 → 0.8 ms | 0.9 → 0.7 ms |
| Attention and machine facts | 0.4 → 0.3 ms | 0.4 → 0.3 ms |
| Categories and archived rows, including filtered refresh | 0.8 → 0.8 ms | 1.0 → 0.9 ms |

All first-call, warmup and adverse observations were retained: 38 of 98 steady mounted sample pairs and 26 of 140 steady direct pairs were slower. No Gateway, inference, startup or allocation measurement claim follows from this cohort. Numerical mounted cases kept child branches collapsed; the existing mounted tests cover expansion.

Measurements are bound to baseline `4492b7502d1b3d3986e6f606290f694206503c15`. An earlier baseline-only attempt failed because its All-status fixture published to the active-list owner; that failed attempt remains separate and was not pooled with the corrected full cohort. A nonnumerical preflight verified all three expected All-status roots and updated labels before the corrected cohort ran. All owned browser/process families closed and the candidate source was restored.

Before: synthetic sidebar attention state

After: identical synthetic sidebar attention state

### DIFF
--- a/ui/src/components/app-sidebar-session-navigation-logic.test.ts
+++ b/ui/src/components/app-sidebar-session-navigation-logic.test.ts
@@ -508,6 +508,142 @@ describe("sidebar navigation lineage ownership", () => {
     },
   );
 
+  it("keeps alias cycles path-local and projects shared descendants in postorder", () => {
+    const root: GatewaySessionRow = {
+      key: "root",
+      kind: "direct",
+      childSessions: ["alias", "root", "left", "right"],
+    };
+    const alias = { key: "root", kind: "direct", archived: true } satisfies GatewaySessionRow;
+    const left = {
+      key: "left",
+      kind: "direct",
+      childSessions: ["shared"],
+    } satisfies GatewaySessionRow;
+    const right = { ...left, key: "right" };
+    const shared = { key: "shared", kind: "direct" } satisfies GatewaySessionRow;
+    const rowsByKey = new Map<string, GatewaySessionRow>([
+      ["root", root],
+      ["alias", alias],
+      ["left", left],
+      ["right", right],
+      ["shared", shared],
+    ]);
+    const calls: string[] = [];
+    for (let iteration = 0; iteration < 2; iteration += 1) {
+      calls.length = 0;
+      const trees = projectSessionTree({
+        roots: [root],
+        rowsByKey,
+        loadingChildKeys: new Set(["root"]),
+        knownSessionAttention: [],
+        toSidebarSession: (row, isChild) => {
+          calls.push(`${row.key}:${isChild}`);
+          return { ...projectSidebarSession(row), isChild: isChild === true };
+        },
+      });
+      expect(calls).toEqual([
+        "root:true",
+        "shared:true",
+        "left:true",
+        "shared:true",
+        "right:true",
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [353cd6e04652231a02a12d84906eeed700e8acc6](https://github.com/openclaw/openclaw/pull/144642/commits/353cd6e04652231a02a12d84906eeed700e8acc6)
- Landed commit: [178c98be25bcbfda433b8ce5ef6a28fc909b1aa7](https://github.com/openclaw/openclaw/commit/178c98be25bcbfda433b8ce5ef6a28fc909b1aa7)

PR #144661, opened 2026-09-11
fix(ui): preserve plugin category navigation while searching

steipete:
## What Problem This Solves

Fixes an issue where the Plugins catalog could lose its category navigation when a category response arrived while the user was starting a search. Search results still appeared, but the category chips and saved home shelves stayed empty until the overview was loaded again.

## Why This Change Was Made

Typing a search clears the selected category immediately, while the query is committed after a 250 ms debounce. Completion previously used that mutable UI state to decide whether a response was an overview. A pending category response could therefore overwrite home categories and shelves with values it never contained.

The existing task result now carries its overview classification from the captured request arguments. Completion consumes that fact, using the same existing predicate as the current view. The obsolete hydration-invalidation comment was removed. This adds seven production lines without changing the request protocol or debounce behavior.

## User Impact

Previously loaded catalog categories and home shelves remain available while searching, even when responses arrive during the debounce window. Home loading, filtered reconnects, and pagination keep their existing behavior.

## Evidence

- The original controller failed the new request-order regression while ten existing controls passed.
- The real Plugins page reproduces the same order through the mock Gateway: select Channels, type a search, resolve the category request before debounce, then resolve a distinct search result. Before the fix, all 12 category chips disappear; afterward all 15 original navigation labels remain.
- Final proof passed 42 distinct tests: 24 controller/catalog-renderer cases and all 18 Plugins browser cases. All 20 selected changed-file checks passed.
- The two full-page captures below use identical final browser-test bytes and synthetic catalog/profile data. They show the completed search before and after the fix.
- Independent review completed two clean passes through P2 on this final source.

Local proof and captures are bound to `ae94766f288b54ec7ac840cc4e1479232405949b`. Subsequent main changes to category icons, category fixtures, a sibling browser case, and a Boolean-reader alias were inspected separately. The controller and Gateway overview predicate are unchanged; the new regression derives its expected labels from the actual loaded categories. No live Gateway, ClawHub, provider, or channel request was used.

Before: completed search loses category navigation

After: completed search retains all category navigation


steipete:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [ea82b79c5776dfa17df9e6c473e108c5c6abdd20](https://github.com/openclaw/openclaw/pull/144661/commits/ea82b79c5776dfa17df9e6c473e108c5c6abdd20)
- Landed commit: [1b9331f096c6f729c8413ac855a9924b60365d63](https://github.com/openclaw/openclaw/commit/1b9331f096c6f729c8413ac855a9924b60365d63)

Issue #144592, opened 2026-09-11
Complete asynchronous runtime storage boundaries while retaining SQLite


steipete:
Continue SQLite-only storage cleanup until runtime database operations and their actual callers use asynchronous contracts. This extends the initial progress-card conversion in #144529 to the remaining store owners, including task/flow and SDK contracts.

Complete domain operations should own connection acquisition, authoritative reads, transactions, decoding, and results. Callers must await persistence and revalidate current authority after waits before publication or external effects. SQLite transaction callbacks and exact-connection schema/mutation helpers stay synchronous inside the SQLite implementation. Synchronous dependency getters should consume prepared facts rather than perform hidden database reads.

Ship small coherent PRs with delayed-operation/rejection/retirement coverage, actual SQLite proof, and live behavior tests where appropriate. Track prerequisites separately from completed owner conversions; do not count an async wrapper over an existing async facade as progress. No PostgreSQL driver/configuration, schema migration, or broader persistence policy change is part of this work.

Initial coordinated lanes are heartbeat outcomes, workspace state, board storage and authority, and task/flow publication/mutation ownership. The finishing audit must cover all remaining core/plugin owners and find hidden database access in purported snapshots and callbacks, not just count SQLite helper names.

PR #144663, opened 2026-09-11
refactor(memory): await core KV provenance operations

steipete:
Related: #144592

## What Problem This Solves

Memory artifact provenance exposes asynchronous operations but uses synchronous core KV access internally. Its write and rollback callers need to retain persistence completion when the store becomes asynchronous.

## Why This Change Was Made

Add the missing core-owner factory through the existing async KV facade and namespace policy. Migrate provenance reads, enumeration, writes, conditional clearing and both rollback paths to await their operations. Reservation checks and atomic update/delete callbacks remain inside the existing SQLite owner.

## User Impact

Memory-file operations wait for their provenance write or rollback to settle. Storage failures remain attached to the operation that owns them. No schema, namespace, quota or trust policy changes.

## Evidence

- 53 focused store and caller tests passed. Eight delayed completion/rejection cases fail before the provenance cutover and pass afterward.
- Full selected changed checks and fresh independent Codex review through P2 passed.
- Two-process production-source smoke wrote a synthetic memory file and provenance to SQLite, reopened it, and verified conditional deletion. Phase timings were 612.64 ms for write/setup and 125.90 ms for reopen/read/delete; these are not comparative latency benchmarks.

### DIFF
--- a/src/memory/memory-artifact-provenance.test.ts
+++ b/src/memory/memory-artifact-provenance.test.ts
@@ -1,7 +1,10 @@
 import { mkdir, symlink } from "node:fs/promises";
 import path from "node:path";
-import { afterEach, describe, expect, it } from "vitest";
+import { expectDefined } from "@openclaw/normalization-core";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import * as pluginState from "../plugin-state/plugin-state-store.js";
 import { resetPluginStateStoreForTests } from "../plugin-state/plugin-state-store.js";
+import { createDeferredCore } from "../shared/deferred.js";
 import { withStateDirEnv } from "../test-helpers/state-dir-env.js";
 import {
   clearMemoryArtifactProvenance,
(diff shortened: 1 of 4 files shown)


steipete:
@clawsweeper review

steipete:
Compatibility confirmed: this PR changes the factory selection and awaited completion only. It does not change the schema, record version, encoded fields, namespace, quotas, workspace identity or reservation comparison, so no data migration is required.

The existing [core-store compatibility test](https://github.com/openclaw/openclaw/blob/b67856402f410b978d9b04a91e30fc094e18c75c/src/plugin-state/plugin-state-store.test.ts#L614) writes through the synchronous factory, closes the database, reads through the async factory, writes back, and reads the same row through the synchronous factory. That test and the provenance reopen/conditional-rollback tests passed in the 53-test cohort, alongside the two-process SQLite smoke. Both factories delegate to the same validated native owner.

The automatic unknown-data-model-change flag is not an actual model change in this diff. No correctness findings remain after maintainer inspection and the independent review. The worker-execution migration remains a separate follow-up under #144592.

steipete:
Merged via squash.

- Prepared head SHA: [b67856402f410b978d9b04a91e30fc094e18c75c](https://github.com/openclaw/openclaw/pull/144663/commits/b67856402f410b978d9b04a91e30fc094e18c75c)
- Landed commit: [32423f00e31e7e882bc79856d88a99b55539c18b](https://github.com/openclaw/openclaw/commit/32423f00e31e7e882bc79856d88a99b55539c18b)

PR #144664, opened 2026-09-11
feat(codex): answer async questions while agents keep working

### LINKED ISSUE #144643: Codex async questions lose structured answer controls
Codex can ask structured questions without pausing its work through `request_user_input_async`. Its app-server emits an async assistant message containing question titles and optional suggested answers, and replies arrive later as ordinary user messages.

OpenClaw currently preserves the readable message text but drops the structured questions when projecting and mirroring the message. Control UI users therefore cannot select a suggestion or answer each question through the existing question controls.

Preserve the native question metadata, reuse the Control UI question controls, and send explicitly submitted answers through normal chat/steering. Keep suggested answers inert until submission, preserve readable channel fallbacks, and leave failed-send recovery with the existing outbox. Questions must remain answerable after the original turn finishes.

Reproduced with the existing Codex async-delivery test: the question metadata is absent from the projected assistant message while ordinary async messages pass. Implementation and live validation are in progress under maintainer direction.

steipete:
Closes #144643

## What Problem This Solves

Fixes an issue where users received plain text instead of answer controls when Codex asked structured async questions. The questions could arrive while the agent continued working, but the Control UI lost their suggested answers and free-text form.

## Why This Change Was Made

Codex emits these questions as async assistant messages and accepts answers as ordinary user messages. Preserve that metadata through projection, transcript mirroring, and native history, then reuse the existing Control UI question panel and normal chat/steering path. No waiting question RPC or additional answer store is introduced.

The existing outbox owns retries after admission, including failed sends. Answering leaves unrelated composer text, attachments, and Reply drafts intact. Restricted native message delivery remains enforced; unsupported metadata, read-only views, and message text rewritten by hooks retain the readable text fallback.

## User Impact

Users can choose a suggestion or enter their own answer while the agent works, or after its turn finishes. The first suggestion is selected for convenience but never submitted automatically. Multi-question messages remain readable offline. Channel users continue to receive the complete question text and reply normally.

## Evidence

- Researched the native [async-question handler](https://github.com/openai/codex/blob/7c88f037d935b4e78311027c32da4f046fd84058/codex-rs/core/src/tools/handlers/request_user_input_async.rs) and [bounded answer framing](https://github.com/openai/codex/blob/7c88f037d935b4e78311027c32da4f046fd84058/codex-rs/context-fragments/src/answered_question.rs) directly.
- Demonstrated the metadata-loss regression before the fix; the repaired projection, persistence, replay, restriction, and hook-redaction tests pass.
- Live managed Codex app-server 0.153.4: verified the exact requested model binding, two native questions, independent shell work before any answer, durable async-message persistence, and a later ordinary answer used on the same native thread. No legacy input-request RPC was emitted.
- Five Chromium E2E cases with a mocked Gateway cover active-turn steering, answers after turn completion, composer/Reply preservation, one outbox owner across failed-send retry, offline readability, and malformed-metadata fallback. Inspected synthetic before/after captures are attached.
- 347 existing chat-send tests, 64 Codex projection/mirror tests, seven turn-parameter tests, and 28 question action/panel tests passed. The final in-place hook-redaction regression also passed.
- Full `node scripts/check-changed.mjs` gate passed, including all affected type checks, full extension lint, UI lint/styles, export scans, and architecture guards. Production Control UI build, i18n verification, and diff checks passed.
- Initial CI caught a type-only UI import cycle. Removed the transcript-state back-edge with a component-local contract; `check:madge-import-cycles` reports zero cycles, and UI types plus all seven scope tests pass.
- The next CI run exposed a session-link fixture defect newly merged into `main` by #144251. Reproduced and corrected the layout expectation; main subsequently landed the same correction in #144690, so the duplicate test-only commit was dropped during conflict resolution.
- Repeated the live model/app-server proof on pre-rebase head `84d93317c44`: passed. Its [CI run 34564274043](https://github.com/openclaw/openclaw/actions/runs/34564274043) also passed, including the real-Gateway UI suite. The current head rebases this implementation onto main, retaining main’s shared reply-quote helper and removing our duplicate. All 79 focused sender/readiness/question cases and a fresh isolated P0–P2 integration review pass. That rebase exposed the sender’s line limit after combining main’s additions; simplified reply-target defaulting without changing null/undefined behavior. Focused tests, independent review, and all local changed-file checks pass on the final candidate. Final-head [CI run 34614673597](https://github.com/openclaw/openclaw/actions/runs/34614673597) is green at `e7c3a700c94b`, including the real-Gateway suite and required aggregate gate.
- Independent Codex review at P0–P2: no remaining actionable findings after fixing retry custody and read-only navigation.

Before: async questions appear only as plain text (synthetic fixture)

After: suggested answers while the agent keeps working (synthetic fixture)

After: a free-text answer remains available after the turn ends (synthetic fixture)

### DIFF
--- a/docs/plugins/codex-harness-runtime/permissions.md
+++ b/docs/plugins/codex-harness-runtime/permissions.md
@@ -70,5 +70,25 @@ Only an explicit field `isSecret: true` or Codex question
 at a time through the warned ephemeral text-reply path and never create durable
 Gateway question records. OpenClaw does not infer secrecy from field names.
 
+## Async questions
+
+When the selected Codex model advertises `request_user_input_async`, Codex can
+ask structured questions while continuing work. The tool returns immediately.
+OpenClaw preserves the questions on the async assistant message, and the Control
+UI shows suggested answers and a free-text field. Selecting a suggestion does
+not send it; the user must explicitly submit an answer.
+
+Answers use normal chat delivery: they steer an active turn or start a new turn
+in the same conversation after it finishes. The question is quoted alongside
+the answer so Codex can identify which question it addresses. Async questions
+do not create a waiting tool call or use the blocking question queue above.
+They follow the same tool and message-delivery restrictions as other native
+async messages. Silence and preselected answers never grant approval.
+
+Channels receive Codex's readable question text and choices and accept normal
+replies. The Control UI supports up to 12 questions per message and four choices
+per question, with titles up to 4,096 characters and choices up to 256 characters.
+Unsupported structured payloads retain the complete plain-text question.
+
 For the general plugin approval flow that carries these prompts, see
 [Plugin permission requests](/plugins/plugin-permission-requests).
--- a/extensions/codex/src/app-server/async-questions.real-binary.live.test.ts
+++ b/extensions/codex/src/app-server/async-questions.real-binary.live.test.ts
@@ -1,0 +1,251 @@
+import fs from "node:fs/promises";
+import path from "node:path";
+import type { EmbeddedRunAttemptParamsV2 as EmbeddedRunAttemptParams } from "openclaw/plugin-sdk/agent-harness-runtime";
+import { createAgentHarnessHostCapabilitiesForTest } from "openclaw/plugin-sdk/plugin-test-runtime";
+import { upsertSessionEntry } from "openclaw/plugin-sdk/session-store-runtime";
+import { readSessionTranscriptEvents } from "openclaw/plugin-sdk/session-transcript-runtime";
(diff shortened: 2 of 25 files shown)


steipete:
@clawsweeper review

steipete:
Merged via squash.

- Prepared head SHA: [e7c3a700c94b2f199b8cf279f7345d0eafd3c562](https://github.com/openclaw/openclaw/pull/144664/commits/e7c3a700c94b2f199b8cf279f7345d0eafd3c562)
- Landed commit: [764ea2bffa1c0741f3b9dfba890522b362be8cab](https://github.com/openclaw/openclaw/commit/764ea2bffa1c0741f3b9dfba890522b362be8cab)

PR #144671, opened 2026-09-11
feat(ui): dismiss failed cloud worker snapshot builds

steipete:
## What Problem This Solves

Resolves a problem where a cloud worker snapshot build that failed stayed in
**Settings → Connections → Cloud workers → Snapshots** indefinitely. The failed row
showed its provisioning error (for example `Gateway is only bound to loopback`) with
no action of any kind, kept counting toward **Needs attention**, and crowded the
profile group alongside live builds. Operators had no way to clear it.

## Why This Change Was Made

Failed build rows now offer **Dismiss**: it confirms through the shared dialog,
requests cleanup with the existing `environments.destroy`, and clears the row.

The dismissal is deliberately view-local, and the confirm dialog says so. The Gateway
keeps a terminal build record until its retention window ends: `environments.destroy`
resolves to `service.destroyUnattached`, whose `providerLifecycle.destroy` short-circuits
for records already in `destroyed`/`failed`/`orphaned` and returns them unchanged. So the
record legitimately survives the call and can reappear after a page reload. Purging it
early would change `worker_environments` retention semantics, which is a material
persistence change owed its own storage decision, so this PR does not attempt it. Failed
records are otherwise inert — the prepared pool excludes them from build accounting and
`reconcileEnvironmentCore` returns early for terminal states — so hiding the row loses
nothing operationally.

**Orphaned** builds deliberately keep no dismissal: terminal retention refuses to prune
orphaned rows that still carry a preparation key, because their provider artifacts may
still exist, and hiding one would hide a real leak.

Cancel and dismiss are the same confirm → `environments.destroy` → refresh flow, so they
now share one `destroyBuild` path and one busy guard. The build dialog moved to its own
module to keep the page under the file line cap.

No new config keys, no protocol change.

## User Impact

Operators can clear a failed snapshot build from the Snapshots view instead of living
with a permanent error row. The error text stays visible until they dismiss it,
**Needs attention** drops accordingly, and an orphaned build still stays listed so a
possible provider leak is not hidden.

## Evidence

Before — failed row with error text and no action:

before

After — the same row offers **Dismiss**:

after

Confirmation states what dismissal does and does not do:

confirm

After dismissal — row cleared, **Needs attention** back to 0:

dismissed

Captures come from the mocked-Gateway Control UI E2E harness with synthetic fixtures.

- `ui/src/pages/cloud-workers/cloud-worker-snapshots.test.ts` — dismissal against a
  Gateway that keeps returning the failed record (asserts confirm payload, exact
  `environments.destroy` params, row removal, attention count, and that the row stays
  hidden across an explicit **Refresh**); refusal and destroy-failure both keep the row;
  the failed/orphaned case asserts **Dismiss** appears only for `failed`.
- `ui/src/e2e/cloud-workers-snapshots.e2e.test.ts` — mocked-Gateway click-through:
  dialog cancel sends no request, confirm sends exactly one, notice appears, row and
  error text clear, and a refresh against the still-recorded environment leaves it hidden.
- 34/34 cloud-worker unit tests and 5/5 snapshot E2E tests pass; `check-changed` clean on
  this head.

The no-op behavior of `destroyUnattached` on a terminal record was confirmed with a
throwaway test through the real service fixture, which reported the record still `failed`
and still listed after the call.


steipete:
@clawsweeper review

steipete:
Merged via squash.

- Prepared head SHA: [c5dd85028c2cac6d6f3bb962a4d7b38a32fbecd7](https://github.com/openclaw/openclaw/pull/144671/commits/c5dd85028c2cac6d6f3bb962a4d7b38a32fbecd7)
- Landed commit: [3c8a784720130c4a3c66af83a825dca6c6cc9a02](https://github.com/openclaw/openclaw/commit/3c8a784720130c4a3c66af83a825dca6c6cc9a02)

PR #136399, opened 2026-09-02
fix(plugins): speed up cold Doctor in built checkouts

steipete:
Closes #136385

**Ready-to-prepare checkpoint — `756ef5f86493242c119e2de3e2c1d614af940e79`.** The maintainer lead explicitly accepts the compiled built-checkout default and +133 production LOC. The stored-data flags refer to existing process-local caches and synthetic fixtures, not a persistence/schema change. The [required CI gate](https://github.com/openclaw/openclaw/actions/runs/33676216914/job/100409949500) is green; full clean-build, Doctor/fault, and independent review proof is recorded below. Preparation and merge have **not** run and await the lead’s next explicit checkpoint approval.

<details>
<summary>Additional instructions</summary>

Maintainers can edit this same-repository branch.

</details>

## What Problem This Solves

Fixes an issue where operators running Doctor from a built source checkout pay unnecessary cold TypeScript transformation costs even when the matching compiled plugin artifacts are present. The work preserves full provider-catalog and runtime-tool diagnostic coverage rather than reducing checks or increasing timeouts.

## Why This Change Was Made

The shared runtime artifact owner now distinguishes source execution, implicit built-host preference for checkout plugins, and explicit preference that also permits installed package-local artifacts. Registry cache identity uses the same policy. Discovery keeps source-owned metadata without forcing TypeScript execution; explicit source paths, physical aliases, and mounted overlays carry source intent through installed-index rehydration.

Emitted package metadata owns JS/CJS/MJS selection, so stale neighboring formats cannot win. Native/Jiti fallback semantics, installed-package boundaries, and first-selection registry pinning remain intact. The manifest deduplication map now owns accepted records directly instead of maintaining a parallel record-index structure.

## User Impact

Ordinary cold Doctor runs in two interleaved comparisons took 22–41% less elapsed time. Built-host provider and tool loads reuse compiled checkout output by default; source-only development, explicit source selections, and overlays continue to work. No new configuration/environment surface, schema migration, dependency policy, or timeout change is introduced.

Production LOC: +161/-28 (net +133). Tests/test support: +349/-25. Docs: +6/-4. The growth carries source intent, emitted-format correctness, and the shared cache-safe preference policy; a Doctor-only shortcut or a global native-package exemption would lose those contracts.

### Cold before/after proof

[...]

PR #144676, opened 2026-09-11
improve: use compiled Doctor contracts in built checkouts

steipete:
Related: #136399

## What Problem This Solves

Resolves a problem where built-checkout Gateway startup and Doctor could still load source contracts despite an available compiled copy. Partial builds could also pair an available staging entry with a canonical directory that did not contain it, leaving the selected Doctor contract without its expected hash.

## Why This Change Was Made

One Doctor artifact descriptor now supplies the selected file and its root to loading, installed-index hashing, and freshness checks. It reuses the existing compiled-artifact selector and keeps execution files and roots together across partial builds and filesystem symlinks. Installed metadata reconstruction shares the existing source-preference preparation rather than introducing another cache.

Explicit source selections, external-plugin ownership, package-local extension precedence, migration declarations, and trust policy remain intact. No configuration option, database schema, retention policy, or migration action changes. The additional production code keeps all three consumers aligned instead of optimizing loading alone.

## User Impact

Built checkouts can reuse compiled Doctor contracts consistently with the rest of plugin startup. Operators retain the same migration coverage, including disabled-plugin configuration repair and existing external-plugin compatibility. Selected contract bytes remain correctly tracked when canonical output is incomplete or reached through symlinks.

## Evidence

- Original code fails the built-contract hash selection and partial-build file/root regressions; 20 controls pass. The candidate passes the focused artifact, registry, and runtime-resolution checks, including 18 source/built layouts, real selected detection/migration results, and replacement loading in a fresh Node process.
- The 29 checks selected by `check:changed` are complete across the initial successful stages and the scoped lint/tail recovery. The fixture-only lint repair also passed its behavioral test, mapped test typecheck, and formatting. No checks or assertions were disabled.
- Additional Doctor declaration, closure, installed-index, registry freshness/checkpoint, and ACP ownership suites were exercised. An existing mocked-Windows fixture failure was reproduced on baseline and traced to the validation temp directory inheriting the checkout's ESM package scope; baseline and candidate pass in private system temp without changing that fixture's assertions.
- Fresh independent Codex review of the complete staged repair is clean at P2 and above. Source-only filesystem-boundary findings were fixed and covered before review.
- The final candidate's normal build with source maps passed in 172.791 seconds. The corrected built-CLI persistence proof also passed against that build. Build manifests and selected embedded source-map contents bind the baseline and candidate outputs to their respective commits.
- Twelve unprofiled, fresh-state Gateway starts passed in fixed B0/C0/C1/B1 order, three starts per cohort. Median HTTP readiness improved from 4095.71 to 3331.46 ms (**18.66%**) and from 4026.88 to 3662.26 ms (**9.05%**) in the two orders. All six indexed readiness comparisons improved. Whole-process wall, CPU, and peak RSS median/mean and indexed comparisons also improved in this sample. Each start reached HTTP 200, exited gracefully, and released its port. These are one-host, warm-filesystem, polling-based startup measurements, not model or agent-turn latency, and three samples per cohort do not establish statistical confidence.
- Separate baseline/candidate Node profiles completed successfully. The observed module cache contained 21 Doctor contracts in each: baseline used source TypeScript for Discord, iMessage, Matrix, and WhatsApp; candidate used compiled contracts for all 21. An earlier diagnostic observer failed because Node contexts shared its output path. That failed profile remains separate; the corrected observer is scoped to the Gateway main process and uses its PID. None of the 12 primary timing samples was rerun or mixed with profiles.
- Exact-head CI is green on the unchanged candidate after one targeted retry: 111 jobs passed and 16 skipped. Native maintainer landing remains pending.

### CI follow-up

The first published head exposed a type-dependency cycle through the runtime-record type and a built-CLI test that reread compiled contract hashes through source-mode rules. The repair separates the unchanged runtime binding functions into their own owner, migrates every caller, and retains the real record type without a shim or cycle exception. Both cycle checks now pass, along with 149 existing runtime tests and 71 canonical-descendant integration tests.

The persistence test still requires a persisted registry with no diagnostics. It now checks the repaired raw index before invoking the existing built `plugins list --json`, then verifies that the read-only command leaves the index unchanged. The corrected observer passes against the baseline and first built candidate; source-mode diagnostics showed that only bundled contract hashes differed, while the external legacy record was already correctly repaired.

The complete repair passed its selected local checks and a fresh Codex review. On the current head, all three original CI failures pass, including build-artifacts, architecture, and the CUA extension bundle. The earlier MCP cleanup failure is retained without claiming an established cause.

The initial current-head CI run has one remaining workload failure: both layouts in the worker-transformation test exceed their unchanged 120-second deadline while repeatedly preparing complete worker generations. The test and compiler owners are unchanged from the PR base.

A subsequent baseline/candidate comparison ran those same two cases in separate Linux x64 Node 24.19 containers, each limited to two CPUs and 8 GiB. Baseline passed in 40.486/39.944 seconds; candidate passed in 36.962/45.810 seconds. All 20 transport observations retained the required modes, values, transform counts, and 16 distinct compiled generations. Both installs/tests joined their processes and released resource claims; both containers exited normally and were removed. Candidate timings are mixed, and these containers are not identical to the hosted CI machines, so this does not establish the original timeout's cause or prove flakiness.

The local supervisor later hit its disk floor and exited 143. The already-completed remote evidence was recovered read-only: 44 files and their hashes were independently checked, and the owned lease was stopped. That outer failure remains recorded. One targeted retry of the failed CI workload and aggregate gate passed on unchanged `c6099189d4fec17b378e80d58a0f2993ac2be479`: [CI attempt 2](https://github.com/openclaw/openclaw/actions/runs/34567614441/attempts/2), 111 jobs passed and 16 skipped. No code, checks, timeouts, or assertions were changed.

### Compatibility review disposition

The requested data-model compatibility confirmation is covered by the existing derived-index contract and its recorded proof. The [index types](https://github.com/openclaw/openclaw/blob/c6099189d4fec17b378e80d58a0f2993ac2be479/src/plugins/installed-plugin-index-types.ts), [parser](https://github.com/openclaw/openclaw/blob/c6099189d4fec17b378e80d58a0f2993ac2be479/src/plugins/installed-plugin-index-store.ts), writer, and startup persistence owner are unchanged from the PR base. Index and migration versions remain 1; Doctor hash/signature fields remain optional. This repairs artifact identity within the existing reconstructible index, without a new stored representation or migration policy.

The exact-head [built-CLI case](https://github.com/openclaw/openclaw/blob/c6099189d4fec17b378e80d58a0f2993ac2be479/test/scripts/doctor-config-preflight-plugin-index.built-cli.e2e.test.ts) removes both optional Doctor fields from persisted records, starts the built Gateway, verifies restored hashes/signatures, then confirms that built `plugins list --json` reads a persisted registry with no diagnostics and leaves the stored index unchanged. That targeted case passed; the separate empty legacy-directory case was skipped. The [artifact fixture](https://github.com/openclaw/openclaw/blob/c6099189d4fec17b378e80d58a0f2993ac2be479/src/plugins/doctor-contract-artifact.test.ts) separately covers source/built selection and refreshing a populated stale compiled hash after restart, while preserving manifest/package hashes and executing the selected migration callbacks.

These are compatibility checks at the affected owner boundaries, not a release-binary upgrade or downgrade campaign. The synthetic legacy plugin's version stamp is fixture data, and the artifact fixture does not persist its source-preferred index before switching to compiled mode. Maintainer and independent source review found the existing evidence sufficient for this unchanged derived-index contract; no new persistent-state design or schema migration is introduced.

### DIFF
--- a/docs/help/debugging.md
+++ b/docs/help/debugging.md
@@ -102,9 +102,9 @@ Add gateway CLI flags after `gateway:watch` and they pass through on each restar
 When you run `pnpm openclaw`, `pnpm dev`, or a Gateway development runner from
 a checkout, the runner selects that checkout's plugins ahead of tracked global
 copies with the same id. Built plugin output remains preferred when available,
-including for separately published checkout plugins and Doctor's provider/tool
-checks. Source-only plugins still load from the checkout. Rebuild to pick up
-source changes when using built output. Intentional source-entry selections and
+including for separately published checkout plugins, Doctor contracts, and Doctor's
+provider/tool checks. Source-only plugins still load from the checkout. Rebuild to
+pick up source changes when using built output. Intentional source-entry selections and
 mounted source overlays keep using source instead of their compiled peers.
 
 This selection is separate from the `--dev` profile. It does not grant trusted
--- a/src/acp/runtime/session-meta-doctor.test.ts
+++ b/src/acp/runtime/session-meta-doctor.test.ts
@@ -5,7 +5,7 @@ import { AcpxRuntime, createAgentRegistry, createFileSessionStore } from "acpx/r
 import { expect, it } from "vitest";
 import { updateSessionEntry } from "../../config/sessions/session-accessor.js";
 import { createPluginDoctorStateMigrationContext } from "../../infra/state-migrations.plugin-doctor-context.js";
-import { resolvePluginDoctorContractArtifactPath } from "../../plugins/doctor-contract-artifact.js";
+import { resolvePluginDoctorContractArtifact } from "../../plugins/doctor-contract-artifact.js";
 import {
   coercePluginDoctorContractModule,
   type PluginDoctorContractModule,
@@ -176,7 +176,11 @@ it.each(["global", "shared-project"])(
       expect(before.incomplete).toEqual([]);
       expect(before.claims).toHaveLength(2);
       const rootDir = path.resolve("extensions/acpx");
-      const modulePath = resolvePluginDoctorContractArtifactPath(rootDir)!;
+      const modulePath = resolvePluginDoctorContractArtifact({
+        rootDir,
+        origin: "bundled",
+        sourcePreferred: true,
+      })!.modulePath;
       const load = getCachedPluginModuleLoader({
         modulePath,
         rootDir,
--- a/src/agents/runtime-plugins.test.ts
+++ b/src/agents/runtime-plugins.test.ts
@@ -55,10 +55,8 @@ import {
   makeRegistry,
 } from "../config/plugin-auto-enable.test-helpers.js";
 import type { OpenClawConfig } from "../config/types.openclaw.js";
-import {
-  bindPluginRuntimeArtifactSelection,
-  resolvePluginRuntimeArtifactSelection,
-} from "../plugins/plugin-runtime-artifact-selection.js";
+import { bindPluginRuntimeArtifactSelection } from "../plugins/plugin-runtime-artifact-binding.js";
+import { resolvePluginRuntimeArtifactSelection } from "../plugins/plugin-runtime-artifact-selection.js";
 import { createEmptyPluginRegistry } from "../plugins/registry-empty.js";
 import {
   getPluginRuntimeGatewayRequestScope,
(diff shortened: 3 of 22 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c6099189d4fec17b378e80d58a0f2993ac2be479](https://github.com/openclaw/openclaw/pull/144676/commits/c6099189d4fec17b378e80d58a0f2993ac2be479)
- Landed commit: [a58c71a8f9e6ca0dae6592a4ac192b3e1dfea53b](https://github.com/openclaw/openclaw/commit/a58c71a8f9e6ca0dae6592a4ac192b3e1dfea53b)

PR #144686, opened 2026-09-11
fix(tests): restore heartbeat fixtures when setup fails

steipete:
## What Problem This Solves

The heartbeat test sandbox seeded scratch storage before entering its cleanup scope. If real storage setup failed, the callback never ran and the fixture left its state-directory override, cleared token value, and temporary directory behind, contaminating later tests.

## Why This Change Was Made

Compose the existing asynchronous environment and temporary-directory helpers so setup and callback execution share the same cleanup lifetime. Reset the reply spy and close shared state before restoring the environment and removing the directory. Use the existing temporary-directory tracker in the heartbeat outcome-store tests while retaining database-before-directory cleanup ordering.

All existing tests remain. Add one real-storage regression that obstructs the sandbox's database directory, proves setup fails before the callback, and checks environment restoration and directory removal before independent rescue. It also covers duplicate unset names and the fixture-owned state-directory precedence.

## User Impact

This repairs test fixtures only; production code and user-facing runtime behavior are unchanged. Existing test/support edits are +25/-35 lines; the 55-line regression makes the complete test/support change +80/-35 (net +45). The added test protects a setup-failure lifecycle that generic helper tests and successful heartbeat runs do not cover.

## Evidence

The same 27 existing test files passed before and after: 376 passed and one expected Windows-only skip, with the same per-file test inventory and order. The exact permanent regression failed on the original helper at its cleanup assertion and passed on the repaired helper; its owned filesystem rejection and callback non-entry assertions passed on both. All recorded native children joined, and source/dependency integrity was verified.

Independent source review found no actionable P0-P2 issue. All 29 checks in the normal changed-file gate passed, including both selected type graphs, formatting, lint, and the complete unused-export audit.

### DIFF
--- a/src/infra/heartbeat-outcome-store.test.ts
+++ b/src/infra/heartbeat-outcome-store.test.ts
@@ -1,7 +1,5 @@
-import fs from "node:fs";
-import os from "node:os";
-import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
+import { createTempDirTracker } from "../../test/helpers/temp-dir.js";
 import {
   prepareSystemAgentRunAdmission,
   resolveAdmittedRunActiveAssertion,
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9281b8a1837ce02c8dd63650989cf391823a1313](https://github.com/openclaw/openclaw/pull/144686/commits/9281b8a1837ce02c8dd63650989cf391823a1313)
- Landed commit: [830e105508cec245ecc1aaa2e41f46eaaf66df55](https://github.com/openclaw/openclaw/commit/830e105508cec245ecc1aaa2e41f46eaaf66df55)

PR #130352, opened 2026-08-26
fix(cli): validate node options before contacting the Gateway

steipete:
Related: #116652

## What Problem This Solves

Fixes an issue where malformed `openclaw nodes` options contacted the Gateway before reporting their local validation error. An offline Gateway or unknown node could hide the useful error for invoke, notify, push, location, screen, and camera commands.

## Why This Change Was Made

Run the existing pure option validators before resolving a node. Checks requiring a node's platform still happen after lookup. The push command directly uses the canonical normalized environment value, removing a one-use wrapper and duplicated option fields. Flags, defaults, error text, mixed-case environment names, and successful APNs output are preserved.

Production LOC: +18/-37 (net -19). Tests: +85/-30 (net +55). No dependency, configuration, schema, protocol, or API changes.

## User Impact

Invalid local arguments now produce the intended diagnostic without needing a reachable Gateway. Valid invocations still contact the Gateway normally.

## Evidence

- Candidate `ebd6139f28d116e9de96081608a3eb7bc1fe9f7e`.
- Before: 21 regression cases detected an unwanted `node.list` call before rejection. After: 97 owner and CLI E2E tests passed.
- Actual built CLI: 20 invalid-input scenarios each exited with the expected diagnostic and made zero connections to an isolated TCP observer. A valid invocation reached that observer once, proving the check could detect traffic. No real APNs request was sent.
- All 11 runtime/declaration build stages and native require validation passed. The complete changed-file gate passed with matching dependencies and no skipped gates.
- Fresh structured Codex P1 review and independent source/runtime evidence review passed on this commit.
- Maintainer-requested, AI-assisted repair and simplification. CI on the published candidate is required before landing.

PR #144687, opened 2026-09-11
fix(cli): reject blank APNs environment before node lookup

Alix-007:
Related: #130352 (the earlier validation-order repair).

## What Problem This Solves

An operator can explicitly pass an empty `--environment` value to `openclaw nodes push --node <id>`. The command currently treats it as omission, reaches node lookup, and can use the node's default Apple push environment.

## Why This Change Was Made

The existing validator now checks whether the option was supplied. Empty, whitespace-only, and invalid values fail before lookup with the existing diagnostic. Regression coverage uses the existing full command-registration suite and replaces its contradictory empty-success expectation.

## User Impact

Explicitly blank input fails locally. Omit the flag to retain the existing default selection. Valid normalized values, node resolution, title/body defaults, and complete failed-push output keep their behavior.

## Evidence

- Real compiled CLI through `openclaw.mjs` on pinned main: empty input in human and JSON modes reached `node.list` and the controlled `push.test` boundary, then exited successfully.
- The regression failed on the original validator for exactly those two cases; the other 12 cases passed.
- All 28 focused command-registration, push-handler, dispatch, and payload tests passed with the repair.
- Formatting, syntax lint, size/assertion checks, and the runtime build passed.
- Independent source-blind acceptance passed 31 public cases, including `--environment=` spellings, varied node resolution, omitted environment forwarding, and complete failed-push output. Blank input created no connection.
- Independent source review found no actionable issues.

The public proof uses synthetic Linux state and a loopback Gateway peer. It verifies compiled CLI validation, forwarding and output. It does not send a real Apple notification or claim live server authorization or device delivery. Unchanged downstream payload, priority, expiration, registration and authorization contracts are covered by source review and focused owner tests.

AI-assisted implementation and validation.


obviyus:
Thanks @Alix-007 for the fix.

**Proof:** The real compiled CLI on pinned main accepted an empty environment in both output modes and reached node lookup and the simulated push boundary. The repaired CLI rejects it locally with zero connections. Independent acceptance passed 31 public cases, including equals-form empty input, normalized and omitted values, varied node resolution, title/body defaults, and complete failed-push output.

- **Bug:** An explicitly empty override was treated as omission and could select the node's default push environment.
- **Fix:** Validate option presence before node lookup; retain the existing valid-value and output behavior.

| Before | After |
| --- | --- |
| Empty environment reached lookup and push submission. | Empty environment fails locally before any connection. |

All 28 focused tests passed, including the consolidated full-registration suite and unchanged push-owner controls. Independent source review and the exact-head ClawSweeper review found no actionable issues. [Required CI passed](https://github.com/openclaw/openclaw/actions/runs/34569118057).

Merged as 969840671abdccdbf35ca23b2666f98cf5d10d74. The public proof used a synthetic loopback peer; it does not claim real Apple notification delivery or live server authorization enforcement.

PR #144697, opened 2026-09-11
fix: attribute slow worktree cleanup

steipete:
## What Problem This Solves

Fixes an issue where operators investigating slow task archives could see one `worktreeCleanup` duration but could not distinguish allocation admission, removal work, lease finalization, or waits in the shared Git ref queue.

## Why This Change Was Made

Add two settled-only, payload-free diagnostic records at the existing removal and Git ref queue owners. They use the existing diagnostics gate, logger, trace context, and fixed-window rate limiter, with at most 60 records per owner per minute. The records retain only fixed scalar timings, outcomes, process/thread facts, and omission counts; they add no repository paths, refs, arguments, output, or error text.

Lease ownership, snapshot safety, queue ordering, deadlines, cancellation, and completion behavior stay intact. Admission includes setup and backoff, and queued operation includes promise settlement; the fields are elapsed intervals rather than CPU measurements.

## User Impact

Operators can correlate a slow archive with its inner admission and queue durations using an existing request trace. These diagnostics improve attribution; they do not make cleanup faster or establish the cause of a historical slow operation. No new configuration or storage schema is introduced.

## Evidence

- Both attribution regressions fail on original production at `068318110aec933f9c6c5ea1981489d6c39ca283` because the diagnostic record is absent, after preserving the original behavior assertions.
- Local focused owners and existing Git/ref-queue siblings: 123 passed, 2 existing skipped cases.
- [Linux Testbox proof](https://github.com/openclaw/openclaw/actions/runs/34564871555): all 20 new diagnostic cases, an ordinary canonical archive/restore preserving dirty and untracked files and the transcript, and the real capacity-wait control passed. The command reached its completion marker. Provider cleanup subsequently timed out; an independent exact lease status read confirmed completion.
- A disposable same-process probe retained all 64 blocks and 1,024 calls, using exact original source copies beside the candidate. Each comparison below contains eight alternating pairs of 16 calls per mode, with no timing assertions, warmup, or discarded samples.

| Boundary / comparison | Total baseline → candidate, ms | Median paired delta per 16 calls, ms |
| --- | ---: | ---: |
| Git queue: original disabled → candidate disabled | 1.433193 → 1.431550 | +0.008230 |
| Git queue: candidate disabled → enabled | 0.952418 → 1.237565 | +0.024151 |
| Removal guard: original disabled → candidate disabled | 122.596211 → 128.931687 | −0.923215 |
| Removal guard: candidate disabled → enabled | 156.973103 → 154.127688 | +0.002515 |

The probe ran on Linux x64 / Node 24.19.0. Git calls used real path resolution and the idle queue; removal calls acquired the real isolated allocation lease and rejected at the existing caller guard before worktree effects. Removal measurements were noisy. Enabled mode also enables existing diagnostics. The probe used no request trace and no call emitted a slow record, so log-write overhead, cold module startup, and successful archive overhead are unmeasured. The measurements support no zero-overhead or runtime-speed claim.

The final independent Codex review is clean through P2. The full docs anchor audit reports the same 53 unrelated failures on the original base and candidate; it adds no failures and reports none in the edited pages. The full `node scripts/check-changed.mjs` gate passed, including core/core-test typechecks, lint, and runtime import-cycle checks.

Current-head [CI passed on attempt 2](https://github.com/openclaw/openclaw/actions/runs/34566136413/attempts/2). The first attempt timed out in two unchanged Discord WAV-backlog tests; one rerun of only that failed job passed without source or deadline changes. Their capture-only path does not call either modified timing owner, but the timeout cause remains unlocalized.

### DIFF
--- a/docs/gateway/diagnostics.md
+++ b/docs/gateway/diagnostics.md
@@ -148,6 +148,18 @@ keys. Repeated stage visits contribute to the counts and totals. Parallel and
 nested stages can overlap, so their totals are neither an exclusive breakdown
 of request time nor CPU measurements.
 
+Two related info-level records help attribute slow worktree cleanup:
+`slow managed worktree removal` separates allocation admission, callback work,
+and final settlement; `slow Git ref mutation` separates directory resolution,
+queue waiting, and queued work. Both require diagnostics and info-level logging,
+emit only after an operation lasting at least one second settles, and have
+separate fixed budgets of 60 records per minute per runtime isolate with
+`omittedObservations` counts. They retain fixed scalar fields and existing traces,
+without adding private paths or new identities. Their elapsed intervals can nest
+inside `worktreeCleanup` and include asynchronous waits; they are not CPU or
+individual child-command timings. See [Slow worktree cleanup](/logging#slow-worktree-cleanup)
+for fields and missing-record limits.
+
 SQLite session-write warnings also separate `queueWaitMs`, `writerExecutionMs`,
 and `completionDelayMs`. These measure time until the writer starts, work and
 awaits inside the writer lane, and time until its caller resumes after execution.
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -332,6 +332,48 @@ for the entire wait or which work consumed CPU. These are ordinary performance
 logs. They do not use or change [audit identity](/gateway/audit), decisions,
 retention, principal attribution or admission authority.
 
+### Slow worktree cleanup
+
+With process diagnostics and info-level logging enabled, two subsystems log
+operations lasting at least one second after they return or throw:
+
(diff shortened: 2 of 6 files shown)


steipete:
@clawsweeper status

steipete:
@clawsweeper review

steipete:
Merged via squash.

- Prepared head SHA: [01389c4be46f736db5d97a866077f2fd5a4a8b90](https://github.com/openclaw/openclaw/pull/144697/commits/01389c4be46f736db5d97a866077f2fd5a4a8b90)
- Landed commit: [1c9e8ca4b6ab9bc3d2b16a91dd3c8558eee08f0d](https://github.com/openclaw/openclaw/commit/1c9e8ca4b6ab9bc3d2b16a91dd3c8558eee08f0d)

PR #144703, opened 2026-09-11
fix(ui): transcript end hides under the PR chip and Side chat questions cannot wrap

steipete:
## What Problem This Solves

Fixes an issue where users following a long-running coding session in the Control UI chat would see the newest transcript rows clipped behind the pull-request chip and task-progress dock at the bottom, even though the view believed it was scrolled to the end and hid the scroll-to-latest button. Opening or closing a side panel (any width change) suddenly "fixed" it. Observed live on a stable team instance: the thread rested 48px short of its end, the last row's bottom sat 25px under the PR chip, and the scroll-to-latest affordance stayed hidden.

Also fixes the Side chat question box, which was a single-line input: long questions scrolled horizontally instead of wrapping, and Shift+Enter could not add a line.

## Why This Change Was Made

Transcript end-follow runs in a requestAnimationFrame after the Lit commit that grows a row. When that commit comes from a task (tool-stream sync, session events) rather than the rAF-throttled stream frame, the follow's frame runs before the row's ResizeObserver enlarges the virtual sizer, so it scrolls to a stale maximum and leaves the reader short by the growth. Nothing re-followed afterwards, and the outstanding "end" command never settled, so scroll events were treated as programmatic and the new-messages state was never published.

The transcript host now keeps an explicit end anchor: the maximum a reader is known to be resting at (set by an instant end-follow or observed after any commit). After each commit, if the end moved while the reader still sits at the previous end, it re-pins them; a reader who left the end by any means (input, find-in-page, focus, programmatic scroll) clears the anchor and is never pulled back. Reader takeover still cancels commands; the existing settle logic remains the terminal state. Expanding a disclosure (tool group, details) is an interaction, not the end moving: it clears the anchor so the clicked row stays put.

A second gap compounded it for Codex-harness sessions: their transient commentary/preamble deltas update the stream segments but were not part of the render-lifecycle change capture, so those growths scheduled no follow at all. The capture now tracks stream-segment identity alongside tool messages. Scroll policy in `scroll.ts` is unchanged. Two owner files crossed the 700-line cap and each hands one cohesive unit to a sibling module (pending-offset restoration; the Side chat composer textarea).

The Side chat composer becomes an auto-growing textarea that reuses the main composer's sizing owner and send-key rule (Enter or the configured modifier shortcut asks, Shift+Enter adds a line), keeps its 400-character limit, and shrinks back after asking.

## User Impact

- The newest transcript content stays above the PR chip and task-progress dock while a session streams tool activity and commentary; the scroll-to-latest button reflects real position again.
- Side chat questions wrap and grow like the main composer; multi-line questions are sent verbatim.

## Evidence

- New browser-mode regression `chat-transcript-end-follow.browser.test.ts` drives the real transcript controller, grows the last row from a task while an end command is outstanding, and asserts the settled position. On the previous code it fails with `distance: 48, overhang: 26, programmatic: true`; with the fix `distance: 0`, overhang ≤ 0, `programmatic: false` (re-verified by reverting only the host change). Two further cases guard the anchor: a reader who leaves the end programmatically is not pulled back when the row grows (a command-based retarget displaced them by 348 px), and a reader observed resting at the end stays pinned when a row grows without any scheduled follow (48 px short without the anchor).
- New mocked-Gateway E2E `chat-scroll-dock.e2e.test.ts`: PR chip and task-progress card appearing after settle, scroll-to-latest, and a growing `agent-run` frame row through preamble commentary and tool events all keep `distance ≤ 8px` and the last row clear of the dock. Without the stream-segment change capture the growing-run scenario rests 181, 321, 461, and 601 px short of the end with the scroll-to-latest button hidden (re-verified by reverting only that change).
- New E2E `chat-session-companion-composer.e2e.test.ts`: wrapping and growth, Shift+Enter newline without sending, verbatim multi-line submission and shrink-back, and Control/Meta+Enter under the `modifier-enter` shortcut.
- Existing suites: `chat-scroll`, `chat-flow.streaming`, `chat-pull-requests`, `chat-session-companion-focus`, `chat-selection-to-main`, `chat-rail-columns`, `chat-responsive.browser` (127), rail/transcript unit and isolated lanes; `check-changed` clean.

Side chat composer, synthetic mock captures (before / after):

Side chat before: single-line input scrolled horizontally
Side chat after: wrapped, auto-growing question box

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,6 +186,7 @@ ui/src/ui/theme-variants.browser.test.ts
 ui/src/ui/__screenshots__
 ui/src/ui/views/__screenshots__
 ui/.vitest-attachments
+ui/.vitest/
 
 # Generated docs baseline artifacts (locally generated, only drift detectors tracked)
 docs/.generated/*.json
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3832,7 +3832,7 @@ ui/src/pages/chat/components/chat-model-picker-options.ts	2
 ui/src/pages/chat/components/chat-model-picker.ts	5
 ui/src/pages/chat/components/chat-pane-header.ts	2
 ui/src/pages/chat/components/chat-selection-popup.ts	1
-ui/src/pages/chat/components/chat-session-rail.ts	4
+ui/src/pages/chat/components/chat-session-rail.ts	3
 ui/src/pages/chat/components/chat-session-sharing.ts	1
 ui/src/pages/chat/components/chat-session-workspace.ts	1
 ui/src/pages/chat/components/chat-sidebar-editor-menu.ts	2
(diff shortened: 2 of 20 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [52e2e60b81f27a8f9118d9e95b505eb96f73bf91](https://github.com/openclaw/openclaw/pull/144703/commits/52e2e60b81f27a8f9118d9e95b505eb96f73bf91)
- Landed commit: [de57e394084e08c8d60832c5cfebf5703a378ff5](https://github.com/openclaw/openclaw/commit/de57e394084e08c8d60832c5cfebf5703a378ff5)

PR #144704, opened 2026-09-11
fix: keep subagent completions on the selected conversation

steipete:
## What Problem This Solves

Fixes an issue where a subagent completion retargeted to a different conversation could retain the requester's old thread. Later fallback merges could restore that stale thread even after a binding or hook selected another destination, including the generated-media route.

## Why This Change Was Made

The announcement owner now applies one fallback merge consistently to bound, hook, direct, and final delivery origins. It inherits a thread only when the target identity still matches, and preserves an explicit thread from the binding or hook.

Conversation identity comes from the already-loaded channel's existing contract. Topic parsing stays channel-owned, normalizer-added prefixes are handled consistently, and opaque Matrix IDs remain intact. Declared case folding applies to raw target fallbacks; canonical IDs returned by channel hooks remain exact. Bound canonical IDs also retain their case. The announcement docs describe this thread-inheritance behavior.

The origin integration suite lives in the permitted root test layer, where it can compose real public plugins independently of the large delivery suite's mutable mocks. Its generic Discord routing cases now register a synthetic Discord plugin explicitly. This five-line fixture correction prevents those cases from unnecessarily loading the real bundled plugin; the real Slack and Telegram normalization controls remain intact.

## User Impact

Completion text and generated media use the selected conversation without carrying a stale thread into another room or topic. Same-conversation fallback threads, explicit overrides, partial origins, and channel/account pairing retain their intended behavior.

## Evidence

Before/after regressions cover retargeted bindings, final fallback merges, topic-qualified overrides, opaque IDs, case-sensitive bindings, and metadata-only case-folding channels. Tests compose the actual public Telegram plugin's normalizer and conversation hook. A real Slack control and scoped canonical-ID controls distinguish normalization from raw case folding.

The first CI run correctly rejected the suite's original core-test placement. Relocating its 32 cases preserved their bodies and resolved imports, without an allowlist or boundary exception. The subsequent tooling failure was reproduced on Linux with Node 24.19, source-only artifacts, the recorded nine-file prefix, and one worker: **258 passed / 18 failed before**, then **276 passed / zero failed** after only the five-line Discord registration. Both runs used the same 120-second test timeout, recorded 300-second no-output watchdog, and zero retries. All original assertions, including the 17 LINE identity assertions, remain unchanged. This establishes the fixture trigger; exact runtime loader-cache identity sharing was not instrumented. [Linux proof run](https://github.com/openclaw/openclaw/actions/runs/34574120177).

Final candidate validation passed **74 selected tests and 16 changed-file stages**, including all 32 origin cases and the complete 10-case architecture boundary suite. The earlier relocation passed 78 selected tests and 23 checks; its four hook cases were not rerun for the final fixture-only change. The earlier complete repair passed 29 checks, and its production and documentation bytes remain unchanged through these test repairs. These counts overlap and belong to their respective source states; they are not additive.

Fresh independent review of the complete staged candidate finished six clean passes through P2. CI for the new head, `d871e0f82f76d80818db19a04da8a96f325f68ce`, is pending. A separate extension-job failure in the previous CI run remains unattributed because its captured log was truncated; the Linux routing proof does not resolve that failure.

Proof uses local routing, public plugin functions, and existing hook/Gateway fixtures. The macOS runs retained generated artifacts and did not reproduce the source-only Linux failure, so they do not establish Linux behavior. No live Telegram, Matrix, Slack, Discord, native client, or provider execution is claimed. The change adds no parser, plugin/SDK contract, configuration, or persistence format, and the fixture repair changes no production loader, global cache, runner setting, or assertion.

### DIFF
--- a/docs/tools/subagents/announce.md
+++ b/docs/tools/subagents/announce.md
@@ -27,7 +27,9 @@ For top-level requester sessions, completion-mode direct delivery first
 resolves any bound conversation/thread route and hook override, then fills
 missing channel-target fields from the requester session's stored route.
 That keeps completions on the right chat/topic even when the completion
-origin only identifies the channel.
+origin only identifies the channel. When an override selects a different
+chat or topic, it does not inherit the previous route's thread. An explicit
+thread from the binding or hook is preserved.
 
 Child completion aggregation is scoped to the current requester run when
 building nested completion findings, preventing stale prior-run child
--- a/src/agents/subagents/announce/subagent-announce-delivery.test.ts
+++ b/src/agents/subagents/announce/subagent-announce-delivery.test.ts
@@ -23,11 +23,6 @@ import {
   PlatformMessageNotDispatchedError,
 } from "../../../infra/outbound/deliver-types.js";
 import { sendMessage as runtimeSendMessage } from "../../../infra/outbound/message.js";
-import {
-  testing as sessionBindingServiceTesting,
-  registerSessionBindingAdapter,
-} from "../../../infra/outbound/session-binding-service.js";
-import { normalizeLegacySessionEntryDelivery } from "../../../infra/state-migrations.legacy-session-store.js";
 import { setActivePluginRegistry } from "../../../plugins/runtime.js";
 import { createDeferredCore } from "../../../shared/deferred.js";
 import {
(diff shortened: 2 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d871e0f82f76d80818db19a04da8a96f325f68ce](https://github.com/openclaw/openclaw/pull/144704/commits/d871e0f82f76d80818db19a04da8a96f325f68ce)
- Landed commit: [32427a18568f4ba95b432a2da794200aa37cce97](https://github.com/openclaw/openclaw/commit/32427a18568f4ba95b432a2da794200aa37cce97)

PR #144710, opened 2026-09-11
fix(macos): avoid stale dashboard command assertions

steipete:
## What Problem This Solves

Resolves a problem where the macOS reconnect test can report missing dashboard commands after a delayed polling task resumes, even when the page has already received them.

## Why This Change Was Made

Read the current WebKit events before deciding that the existing five-second deadline has expired. Keep the ordered three-command assertion and 10 ms polling interval unchanged, and include safe loading, failure, queue, and evaluation diagnostics if delivery still fails.

## User Impact

More reliable native CI evidence. Product behavior is unchanged.

## Evidence

Repository SwiftFormat and `git diff --check` pass. The final candidate and both disposable scheduling probes separately passed managed P2 review with no actionable findings.

The [original controlled native run](https://github.com/openclaw/openclaw/actions/runs/34567216977/job/103161802306) proved the stale-sample failure: a fresh WebKit read saw all three commands after the deadline, while the loop still asserted its earlier empty sample. That exact assertion was the sole failure across 2,139 native tests at width three.

The [fixed controlled native run](https://github.com/openclaw/openclaw/actions/runs/34568681317/job/103166070446), at head `83c4e439befceac691e7ff869455efa78a40428e`, confirmed the same stale first sample and verified delivery after the unchanged deadline, then passed the exact-event test. All 2,139 default-profile tests and both named-profile tests passed at width three.

The instrumentation-free [final native run](https://github.com/openclaw/openclaw/actions/runs/34571256379/job/103173807798), at head `0e9b0844289a8a03518120c7855c967def3444f8`, passed all 2,146 default-profile tests and both named-profile tests at width three. The reconnect test passed in 3.633 seconds, and the completed log contains no controlled-probe markers. All response gating, deliberate consumer delay, independent probe reads, and marker output have been removed from the final patch. The controlled experiment establishes the polling defect; the original unrelated CI log did not record enough WebKit state to establish that incident's precise failure ordering.

### DIFF
--- a/apps/macos/Tests/OpenClawIPCTests/DashboardReconnectTests.swift
+++ b/apps/macos/Tests/OpenClawIPCTests/DashboardReconnectTests.swift
@@ -20,6 +20,7 @@ private actor DashboardReconnectAuthGate {
 @MainActor
 struct DashboardReconnectTests {
     @Test func `primary discovery failure preserves commands owned by a pending picker`() async throws {
+        let responseGate = DashboardWindowOwnershipPresentationGate()
         let discoveryGate = DashboardWindowOwnershipPresentationGate()
         let profileGate = DashboardWindowOwnershipPresentationGate()
         let failDiscovery = LockIsolated(false)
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0e9b0844289a8a03518120c7855c967def3444f8](https://github.com/openclaw/openclaw/pull/144710/commits/0e9b0844289a8a03518120c7855c967def3444f8)
- Landed commit: [c1a7a205daa122c601238e01994150fd219258d1](https://github.com/openclaw/openclaw/commit/c1a7a205daa122c601238e01994150fd219258d1)

PR #144716, opened 2026-09-11
refactor(tui): observe picker selection through callbacks

steipete:
## What Problem This Solves

The TUI picker tests inspected selection through two methods that no production caller uses. That kept test-only API surface on both OpenClaw picker components and allowed the tests to observe an item without proving that Enter actually emits the selection used by the overlay.

## Why This Change Was Made

Remove only the two OpenClaw `getSelectedItem()` wrappers and make the existing component assertions observe `onSelect` after Enter, including an exact single-callback assertion. The existing sanitizer test keeps its meaningful callback and raw-value assertions. Pi's real `SelectList.getSelectedItem()` call in Filterable's Enter handler remains unchanged.

The pickers deliberately stay separate: Filterable clears a filter before cancelling, wraps navigation and emits sanitized row copies; Searchable cancels immediately, clamps navigation and emits original items. Their ranking, layout, highlighting and terminal-sanitization behavior are preserved. Production **+0/−8**; tests **+48/−26**. No dependency, configuration, storage, protocol or public plugin API change.

## User Impact

No intended user-visible behavior or appearance change. TUI picker behavior remains the same, while tests now observe the callback contract consumed by production overlays rather than a test-only accessor.

## Evidence

- Original unchanged component suites: **72 passed**. Migrated assertions on original production: **72 passed**. Candidate and final candidate: **72 passed** each, with identical leaf inventory and per-file order.
- An omitted Enter callback in each owner makes its selected regression assertion fail for the intended `expected 1 callback, got 0` reason. Exact inverses restore both selected cases to green.
- The unchanged native PTY suite passed **9 original + 9 candidate cases** on Linux Node **24.19.0**, pnpm **12.3.4**, Pi **0.84.4** and node-pty **1.2.0-beta.15**, with matching actual runtime/dependency bindings. All **21 recorded selection/cancellation frame and call records** match between original and candidate. This covers editing while a model update is pending, model cancellation and the session picker's clear-first behavior across four cancellation encodings.
- The ordinary changed gate passed all **29 selected checks**, including both core type graphs, exact-file lint with zero warnings/errors, dead exports and the selected ownership guards. [Changed-gate run](https://github.com/openclaw/openclaw/actions/runs/34565480833).
- Complete independent precommit review and separate authenticated exact-commit review of `6034c746f956292a1f01f0c03439a14187b3250f`: both scoped-clean through **P2**, with all four changed files and complete owner/caller/dependency context. No prior review verdict was supplied as input to the exact-commit review.

The signed commit has raw parent `bfac43ad2e78a21676eb36079111e5efb528e4dd` and the exact tested tree `a5b6df0c9238efd0eb049c893cb5f0e55d411938`. The complete 40,762-entry physical/committed source map was verified, not only the four changed files. The proof leases and their recorded processes, keys, claims and temporary sync checkouts are closed. [Original PTY environment](https://github.com/openclaw/openclaw/actions/runs/34553284921) · [Candidate PTY environment](https://github.com/openclaw/openclaw/actions/runs/34564004795).

<details>
<summary>Proof boundaries and retained recovery limitations</summary>

The PTY suite runs the real TUI loop and terminal process with its existing synthetic backend. It is not live Gateway, provider, authentication, session-persistence or screenshot proof. Recorded frames are terminal observations, not image captures. Test duration differences are not a performance claim.

The original run's native completion and provider timing report exit 0, but its old local session handle was unavailable during recovery; its local wrapper/tee exit codes remain unknown. The original worktree setup likewise has verified complete materialization but unavailable separate Git/tee exits. Neither operation was replayed to manufacture cleaner history.

Original-run raw index bytes changed without retained earlier bytes to establish why. Logical index/flags, source contents and the complete physical map were independently unchanged; the cause is not labeled a stat-cache refresh. Candidate PTY and changed-gate raw index checks remained equal.

Local component proof used the ordinary Node 26.8.1 binding. A later metadata-only Bash capture resolved Node 26.5.0; it ran no test or lease, and the normal shell binding was separately recorded before wrapper work. No PATH, runtime, heap, worker, provider, timeout or offline override was used. Optional `fd` was absent in the Linux observations; the selected picker flows do not require it. The TUI's ordinary asynchronous helper provisioning path remains possible, so the invocation is not claimed network-free or proof of successful provisioning.

Process evidence covers the recorded identities and fixture namespaces, not an exhaustive descendant trace. The existing first-case disposal path can synthesize exit 137 during forced cleanup; this is not relabeled as an observed native exit. One task-owned idle SSH mux left after the changed gate was positively attributed and closed separately after its control socket had disappeared. Initial guard self-parent bookkeeping and an archive-order observer mismatch are retained as caller/observer corrections; no native test was replayed or weakened.

The linked Actions runs provision the Testbox environments. Native wrapper/completion results, not a backing-run color after lease cleanup, establish the reported outcomes. Post-completion portal-sync warnings are retained.

</details>

### DIFF
--- a/src/tui/components/filterable-select-list.test.ts
+++ b/src/tui/components/filterable-select-list.test.ts
@@ -1,6 +1,6 @@
 // Filterable select list tests cover keyboard filtering and cursor behavior.
-import { CURSOR_MARKER, visibleWidth } from "@earendil-works/pi-tui";
-import { describe, expect, it } from "vitest";
+import { CURSOR_MARKER, type SelectItem, visibleWidth } from "@earendil-works/pi-tui";
+import { describe, expect, it, vi } from "vitest";
 import { stripAnsi } from "../../../packages/terminal-core/src/ansi.js";
 import { FilterableSelectList, type FilterableSelectItem } from "./filterable-select-list.js";
 
@@ -31,6 +31,14 @@ const testItems: FilterableSelectItem[] = [
 ];
 
 describe("FilterableSelectList", () => {
+  function selectByEnter(list: FilterableSelectList) {
+    const onSelect = vi.fn<(item: SelectItem) => void>();
+    list.onSelect = onSelect;
+    list.handleInput("\r");
+    expect(onSelect).toHaveBeenCalledTimes(1);
+    return onSelect.mock.calls[0]?.[0];
+  }
+
   function typeInput(list: FilterableSelectList, text: string) {
     for (const ch of text) {
       list.handleInput(ch);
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6034c746f956292a1f01f0c03439a14187b3250f](https://github.com/openclaw/openclaw/pull/144716/commits/6034c746f956292a1f01f0c03439a14187b3250f)
- Landed commit: [1531482dc3bccafda4260a11b94d7911a572a8b6](https://github.com/openclaw/openclaw/commit/1531482dc3bccafda4260a11b94d7911a572a8b6)

Issue #35277, opened 2026-03-05
openclaw-tui: Messages displaying twice in terminal UI


ziscross-gh:
**Description:** Messages appear duplicated in openclaw-tui window, but display correctly in Telegram (single message).\n\n**Steps to reproduce:**\n1. Use openclaw-tui interface\n2. Send/receive messages\n\n**Expected behavior:** Single message displayed\n\n**Actual behavior:** Messages rendered twice in TUI\n\n**Environment:**\n- OS: macOS (Darwin 24.6.0 arm64)\n- Channel: TUI\n- Version: See \n\n**Additional notes:**\n- Issue persists after restarting TUI\n- Telegram shows messages correctly (not duplicated)\n- Seems to be a rendering/display bug in the terminal UI only

PR #144730, opened 2026-09-11
fix(ui): duplicate assistant replies in the terminal UI

mengyuyuan:
Related: #35277

## What Problem This Solves

Fixes an issue where users answering an interactive question (`ask_user`) in the terminal UI could see the assistant's reply rendered twice. A live replay of the reply could match more than one row of the same agent run (for example an earlier tool-call row), and the projection then inserted the replay as a second copy instead of recognizing it as the reply already on screen. The same duplication was reported in #35277, which was closed for inactivity while the behavior remained.

## Why This Change Was Made

When a live replay matches several rows of one run, the projection now resolves the ambiguous case with the same rules the snapshot path already applies:

- **Full display content** decides the candidate: the replay is reconciled only with a row whose content identity (text, non-text blocks, and `__openclaw.media`) matches exactly, so a same-caption reply with a different attachment is no longer suppressed.
- **Terminal evidence** is required before adopting a row as the replay target: the row must carry a `runTerminal` marker, a terminal stop reason, or a completed-run position in the snapshot. An identical earlier row followed by more run activity no longer swallows a later final.
- **Recoverable tentative matches**: when a match rests only on a position inference (no explicit terminal marker), live reconciliation now records the same tentative recovery the snapshot path keeps - one shared rule for both arrival orders. If a later history refresh contradicts the match, the suppressed replay is restored; the record retires on explicit confirmation, removal, or filtering.
- **Later distinct finals stay visible**: when a run has accepted a later distinct final (its first delivered reply stays the immutable final message), tentative recovery cannot represent it; the live reply is now kept visible instead of being suppressed unrecoverably, so a later contradictory refresh cannot lose it.

Exact identity matches and unambiguous single matches keep their existing behavior, and both arrival orders share one extracted evidence/content rule instead of duplicating it.

## User Impact

Terminal UI sessions no longer show a duplicated copy of an assistant reply after question/tool flows, no longer drop a distinct final (or a distinct attachment) that shares its text with an earlier row, and no longer lose a replay when partial history first looks terminal and later contradicts it. Sessions that already rendered correctly are unaffected.

## Evidence

Before / after captures from the affected session:

- Before - the reply pair rendered twice (second copy highlighted in red):

  before

- After - the reply renders once (highlighted in green; one line redacted):

  after

Additional evidence:

- Regression coverage in `packages/gateway-client/src/session-projection-terminal-reconciliation.test.ts`: history-first ordering (an unsequenced live terminal is retained after a later tool boundary), distinct attachments (a same-caption reply with different content survives while an exact replay still reconciles), ambiguous multi-row replay acceptance (only with terminal evidence), the partial-history -> live-final -> contradictory-history recovery, and the later distinct final retained through contradictory same-run tool history.
- On the previous revision the ambiguous-match, recovery, and later-final regressions fail; on this revision all session-projection suites pass locally (170/170), and the touched files stay within the repository `max-lines` budget.
- The strict type-check (`noUncheckedIndexedAccess`) passes on this revision and reproduces the previous narrowing error on the prior revision.
- Reproduction: in a terminal UI session, trigger an `ask_user` question and answer it; when the run finished, the reply pair could appear twice. An instrumented trace showed the replay matching two same-run rows (a tool-call row and the reply row), which made the resolver fall back to inserting a new row.


mengyuyuan:
Both review findings are addressed in the latest push:

1. **Same-caption, different attachment**: ambiguous matches now compare full display content (text, non-text blocks, and `__openclaw.media`) through the existing final-content identity, so a distinct attachment is no longer discarded, while an exact replay still reconciles into its durable row. Regression added.
2. **Terminal evidence required**: a row is adopted as the replay target only when it carries terminal evidence (its own `runTerminal` marker, a terminal stop reason, or a completed-run position), by extracting the snapshot path's safeguards into one shared rule. The identical-prose-then-later-tool-boundary case is now retained in live arrival order, the counterpart of the reverse-order test already in `session-projection-terminal-reconciliation.test.ts`. Regression added.

Verification on this revision:

- The two ambiguous-match regressions fail against the previous revision of this branch and pass now; the full session-projection suites pass locally (168/168).
- The `max-lines` violations are resolved as well: the regressions now live next to their reverse-order sibling in `session-projection-terminal-reconciliation.test.ts`, and the shared rule lives in `session-projection-final-identity.ts`, leaving both previously-flagged files under budget (706 to 695 and 1017 to 997).

mengyuyuan:
@clawsweeper re-review

mengyuyuan:
Revision 3 pushed (`41d81c7e`).

- **Recovery for position-inferred live matches**: live reconciliation now records the same tentative recovery the snapshot path keeps, through one shared helper (`withTentativeRecovery`) used by both arrival orders. A suppressed replay whose match only rested on position is restored when later history contradicts it; the record retires on explicit confirmation, removal, or filtering.
- **Regression added**: partial-history to live-final to contradictory-history, right after its snapshot-order sibling in `session-projection-terminal-reconciliation.test.ts`. It fails on the previous revision (the replay is lost) and passes now.
- **CI fix**: the type-check and boundary-dts failures were a `noUncheckedIndexedAccess` narrowing error in `findUniqueTerminalContentMatch` (`candidates[0]` vs `T | null`). Fixed; the local strict type-check reproduces the old failure on the previous revision. Full session-projection suites: 169/169 locally; touched files within the `max-lines` budget.

Re-requesting review.

mengyuyuan:
Follow-up (`7d47acac`) for the two failing checks: `oxfmt --check` wanted one wrap in `session-projection-final-identity.ts` (a call over the print width), and `knip` flagged `hasExplicitTerminalEvidence` as an unused export - it is only called inside its own module, so it is now module-private. No behavior change; the local suite remains 169/169 and the touched files stay inside the `max-lines` budget.

mengyuyuan:
Addressed in `ba2287a79e03`:

- Live reconciliation now retains a later distinct final when tentative recovery cannot represent it: a position-inferred match with no explicit terminal evidence is kept visible instead of being suppressed without a recovery record, so a later contradictory refresh cannot lose the reply.
- Explicit-terminal deduplication and the immutable first-final state are unchanged.
- Added the two-distinct-finals followed by contradictory same-run tool history regression (`keeps a later distinct final visible when tentative recovery cannot represent it`).
- Local matrix on this revision: 170/170 session-projection tests pass; the previous revision fails the new regression (169/170). Strict type-check, formatting, and the max-lines budget stay green.

mengyuyuan:
@clawsweeper re-review

mengyuyuan:
Follow-up fix pushed in `ab721582` — the previous head's CI surfaced a webchat recovery regression in `ui/src/pages/chat/chat-state.test.ts` ("keeps reconnected commentary and answer single with persistence before-final").

Root cause: a boundary-split terminal tail carries a sequence fence so it reconciles against its later durable row. After that fenced row matched, the new retention path could still keep the tail as a later distinct final, so it was published on top of the matched row and the reply rendered twice.

Sequence-fenced replays are now excluded from later-distinct-final retention and keep their fence semantics (suppressed once matched, resumed via the snapshot path otherwise); unfenced later finals are unchanged. A regression test covering the fenced tail was added (fails on the prior revision); session-projection suites pass locally (171/171).

mengyuyuan:
@clawsweeper re-review

mengyuyuan:
Type fix pushed in `bd8afad9` — CI's plugin-sdk boundary dts step (and the tsgo lanes behind it) caught a type error introduced by the previous revision: the new fence guard reads `entry.afterSequence`, but the local `TerminalProjectionEntry` helper type didn't declare the field, so `TS2339` took down the lint / test-types / prod-types / package-boundary lanes together.

Fix: declared `afterSequence?: number | null` on the helper type (one line). Runtime behavior is unchanged — the session-projection suites still pass locally (171/171).

Note: `checks-node-compact-small-51` is the compact planner fixture (`compact github node test plan exceeds 80 jobs (81 planned)`); the same planner test is currently failing on other open PRs, so it looks unrelated to this diff.

mengyuyuan:
@clawsweeper re-review

mengyuyuan:
Merge follow-up: `main` moved ahead with the terminal snapshot-context refactor (#145618) and the optimistic-identity projection change (#145663), which touch the same files as this fix.

Conflicts resolved by extracting the completed-run snapshot-context check into a shared factory (`createCompletedRunSnapshotContextCheck`): snapshot promotion and live replay now reuse one evidence rule across candidate rows while keeping #145618's single-sweep indexing.

Runtime behavior is unchanged: session-projection suites pass locally (171/171), typecheck and formatting are clean.

mengyuyuan:
@clawsweeper re-review

VACInc:
Related to #155336, but not superseded by it. #155336 fixes WebChat streamed/final identity divergence caused specifically by a prepended model-fallback status notice. This PR fixes a separate shared-projection case: ambiguous same-run terminal replay matching across multiple rows, including terminal evidence, tentative recovery, and later distinct finals. Please keep this PR open for its independent reconciliation fix.

Action: KEEP OPEN. DO NOT CLOSE.

PR #132552, opened 2026-08-29
fix(auto-reply): stop attachment staging after cancellation

Alix-007:
Related: #98468. This fixes explicit lifecycle cancellation. A transfer that hangs without a lifecycle abort remains separate, so that issue stays open.

## What Problem This Solves

Cancelling a reply could leave attachment copying alive or publish attachment paths after the run ended. Gateway attachment preparation could also finish the send before its temporary uploads were cleaned up, or report cancellation as an attachment error. Plugin-bound preparation did not consistently join its work to the reply lifecycle.

## Why This Change Was Made

The existing reply, dispatch, and Gateway cancellation signal now reaches the shared staging owner and subprocess runner. The staging owner stops retries, waits for the SCP process tree before removing its download directory, and fences publication after awaited path and alias work, including empty results. Cancellation preserves the original abort reason; ordinary transfer failures retain their bounded diagnostic, including when an exit or spawn error coincides with cancellation.

Gateway preparation owns one awaited cleanup path for parsed uploads, including sandbox-off and pass-through results. Established errors retire cancellation before cleanup yields while retaining work admission, so a later abort cannot replace the original outcome. Ordinary managed-PDF fallback remains supported. Hook preparation uses the existing dispatch abort/drain helper.

The repair folds the single-caller SCP wrapper into its temporary-directory owner and removes the duplicate source-path shape, assignment-only helper, separator conversion, and private test export. Production is **+122/−127, net −5**. One obsolete assertion-baseline allowance is removed. The three-attempt transfer policy, strict SSH options, bounded diagnostics, and 50 MiB staging cap remain unchanged. No configuration, dependency, timeout policy, schema, protocol, or public plugin API is added.

Provenance: reproduced on refreshed main `06759eab`; the exact introducing commit is not established.

Credit: this completes and rewrites @Alix-007's cancellation repair and preserves the contributor's commits.

## User Impact

Cancelling a reply stops attachment staging and prevents cancelled attachment facts from reaching later processing. Gateway sends retain their temporary uploads until cleanup settles, with cancellation and ordinary failures reported distinctly.

## Evidence

The exact signed commit passed both independent postcommit reviews and [hosted CI](https://github.com/openclaw/openclaw/actions/runs/33935887661). The [fresh ClawSweeper review](https://github.com/openclaw/openclaw/pull/132552#issuecomment-5461954809) covers this exact head, with no findings or remaining pre-merge work. Native preparation verified hosted proof without changing the reviewed commit.

[...]

Issue #98468, opened 2026-07-01
sandbox: scp spawn has no timeout, leaks hung processes


zhangLei99586:
## Summary

The `scpFile()` function in `src/auto-reply/reply/stage-sandbox-media.ts` spawns an scp child process with no timeout. If the SCP connection hangs (network partition, unresponsive host), the child process and Promise remain alive indefinitely, accumulating across multiple staging operations.

## Reproduction

Source inspection: lines 322-351 create `spawn("scp", ...)` that only settles on `error` or `exit`. No timeout mechanism exists.

## Expected

Add a 30-second timeout with `SIGKILL` and Promise rejection. Use a `settled` guard to prevent double-resolution.

## Scope

`src/auto-reply/reply/stage-sandbox-media.ts:322-351` — 1 file, ~22 LoC change.

PR #98141, opened 2026-06-30
fix: add timeout to scp spawn to prevent hanging in sandbox media staging

zhangLei99586:
Related: #98468

## What Problem This Solves

`scpFile()` spawns scp with no timeout — hung connection leaks process + Promise.

## Why This Change Was Made

Add 30s `SCP_TIMEOUT_MS` with `SIGKILL` + `settled` guard.

## User Impact

Prevents hung SCP operations from leaking resources.

## Evidence

**Runtime proof — source verification + tests (Node 24):**

proof

- `SCP_TIMEOUT_MS=30000 + settled guard + SIGKILL` — 3 source lines confirmed
- 5/5 tests passed ✅

PR #144736, opened 2026-09-11
fix(auto-reply): bound the sandbox scp transfer with a deadline

Finn763:
## What Problem This Solves

Sandbox media staging shells out to `scp`, and that spawn had **no deadline**: a hung transfer (dead peer, wedged ssh, a dropped TCP session that never resets) left the child process alive indefinitely and the caller awaiting a promise that never settles — one leaked `scp` per stalled staging attempt, with no bound on how long the turn could hang.

This gives the transfer a bounded deadline **through the shared process-lifecycle runner rather than by owning the spawn directly**, which is the direction #98141 was closed toward: that PR was closed as "superseded by the current process-lifecycle architecture, not because the linked defect is fixed" — so the timeout belongs on the existing runner call, not on a locally-owned `child_process.spawn`.

## Evidence

The change is two files:

```
src/auto-reply/reply/stage-sandbox-media.ts          |  8 +-
src/auto-reply/reply/stage-sandbox-media.scp.test.ts | 101 ++++++++++++++++++-
```

- `stage-sandbox-media.ts` derives the per-transfer budget (`resolveScpTransferTimeoutMs()`: size cap / supported throughput, 30s floor, `OPENCLAW_SANDBOX_SCP_TIMEOUT_MS` override) and passes it as `timeoutMs` on the existing shared-runner invocation — the runner already owns deadlines and descendant-tree termination.
- The test suite asserts the deadline reaches the runner (`expect.objectContaining({ timeoutMs: SCP_TRANSFER_TIMEOUT_MS })`) so the wiring cannot silently regress, and adds a real-process hang case: a fake `scp` on `PATH` that spawns a descendant, ignores `SIGTERM`, and never exits. The assertion is that the transfer is rejected by the deadline and **both the parent and the descendant are gone afterwards** (via the repo's existing `test-utils/process-tree` helpers), plus a negative control that a normal transfer still succeeds unchanged.

```
node node_modules/vitest/vitest.mjs run src/auto-reply/reply/stage-sandbox-media.scp.test.ts --maxWorkers=1 --no-file-parallelism
Test Files  1 passed (1)
Tests  13 passed | 2 skipped (15)
```

## Verification

The deadline wiring and the normal-transfer assertions run and pass here; the two skipped cases are the POSIX-only hang cases (`it.runIf(process.platform !== "win32")`) — this machine is Windows, so the process-tree termination assertion itself is **not exercised locally**. I am stating that rather than implying it: it needs a POSIX CI lane to run, and that is the one part of this change I could not observe directly.

Partial fix for #98468: the deadline wiring, the revised budget, and real timeout recovery (process tree + temporary cleanup) are verified here; the remaining unobserved part is a real multi-hour suspend/transfer scenario on a POSIX lane, which CI's Linux lane runs via the committed cases.

## Context worth knowing before review

- Three prior attempts exist on this issue: #98141 (closed as superseded by the lifecycle architecture — see above), #101473 (long review loop, closed), #132552 (merged, but that was the *cancellation* half of staging, not the transfer deadline).
- The issue currently also carries `clawsweeper:queueable-fix`. If the automated fix lane produces a PR for the same deadline, take that one — the intent here is to have the fix available in the architecture the maintainer pointed at, not to race a bot.

## Revision 2 — revised transfer budget + real transport proof (head `8707115c920`)

**Decision answer (the question the review asked): revise the deadline instead of an unconditional 30s cutoff.**
`resolveScpTransferTimeoutMs()` derives the per-transfer budget from the size cap and the slowest throughput still called supported, keeps it bounded, and is overridable:

| Case | Budget |
|---|---|
| 50 MiB cap at the 256 KiB/s supported floor | **200s** |
| review example: 40 MiB at 1 MiB/s (~40s) | **160s** (the old fixed 30s could never allow it) |
| small transfers | 30s floor |
| `OPENCLAW_SANDBOX_SCP_TIMEOUT_MS` | operator override (tests use it to reach the deadline in seconds) |

The deadline remains an absolute, bounded value owned by the shared runner, so an unattended hang still dies; it just no longer kills a slow-but-healthy copy. A deadline kill is also no longer retried (one bounded wait instead of three): retrying a copy that stopped progressing only re-waits the whole budget. If maintainers want a different number, or a config field instead of the env override, name it and I'll switch it.

### Real behavior proof (no mocks: real spawned transport through staging, Windows)

A compiled stand-in transport is spawned through the same argv/PATH path the runner uses, so the real spawn, deadline, tree kill, and temp cleanup all execute:

```
[proof] slow transfer staged=1 elapsed=7965ms
[proof] staged bytes=real transport bytes
[proof] hung transport pid=22892 alive=true
[proof] descendant pid=54128 alive=true
[proof] reaped by deadline: parent=true descendant=true
[proof] surface={"value":{"staged":{}}}
[proof] attempts="attempt\n"
[proof] temp dir removed=true
 ✓ windows real transport proof > publishes a real slow transport that finishes inside the budget 8044ms
 ✓ windows real transport proof > kills a real hung transport tree on the deadline without retrying 8507ms
```

Both Windows-native cases pass (17 passed | 3 skipped with the scratch cases in place) and the same file without them is **15 passed | 3 skipped, exit 0**; `oxfmt --check` clean on both files. The three skips are the POSIX-gated cases (they spawn a shebang script), which is why the Windows run above uses a compiled stand-in. `oxlint` produced no verdict for explicit paths in this junction worktree, so I am not claiming a lint verdict — CI's lint lane judges.

The new committed cases: the derived-budget unit math (including the 40 MiB / 160s case), a real slow transfer that publishes, and the hung tree + descendant reaped on the deadline with one attempt and the temp dir removed.

### DIFF
--- a/src/auto-reply/reply/stage-sandbox-media.scp.test.ts
+++ b/src/auto-reply/reply/stage-sandbox-media.scp.test.ts
@@ -28,6 +28,7 @@ import type { RuntimeMsgContext, TemplateContext } from "../templating.js";
 import { stageSandboxMedia } from "./stage-sandbox-media.js";
 
 const SCP_STDERR_TAIL_CHARS = 16_384;
+const SCP_TRANSFER_TIMEOUT_MS = 30_000;
 const REMOTE_PATH = "/synthetic/attachments/report with spaces.txt";
 const SUCCESS = {
   code: 0,
@@ -207,6 +208,7 @@ describe("stageSandboxMedia SCP", () => {
           download,
         ],
         expect.objectContaining({
+          timeoutMs: SCP_TRANSFER_TIMEOUT_MS,
           maxOutputBytes: { stdout: 1, stderr: SCP_STDERR_TAIL_CHARS * 4 },
         }),
       );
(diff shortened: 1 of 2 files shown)


Finn763:
CI note — the failing lane here is unrelated to this diff.

**Failing job:** `checks-node-changed-extensions-config-49` (plus `openclaw/ci-gate`, the downstream gate that reports "Verify selected CI lanes" whenever any lane fails):

```
Failed Tests 1
AssertionError: expected { Object (enabled, available, ...) } to match object { enabled: true, available: false }
  ❯ extensions/memory-core/src/memory/manager-temporal-ranking.test.ts:156:38
Test Files  1 failed | 168 passed | 1 skipped (170)
```

**Why it cannot come from this PR:** the diff is two files in `src/auto-reply/reply/` (`stage-sandbox-media.ts` and its test) — it adds a deadline constant and passes `timeoutMs` to the existing shared runner call. `extensions/memory-core`'s temporal-ranking availability assertion is a different package on a different code path; nothing in this change reaches it. The remaining failing job in the run is `openclaw/ci-gate`, which only reports that a lane failed.

Local evidence for the change itself (run on the current head, `1ecf353c9aa7`):

```
node node_modules/vitest/vitest.mjs run src/auto-reply/reply/stage-sandbox-media.scp.test.ts --maxWorkers=1 --no-file-parallelism
Test Files  1 passed (1)
Tests  13 passed | 2 skipped (15)
```

(the 2 skipped are the POSIX-gated hang cases, as noted in the PR body — that is the part I cannot observe on Windows and is why this is marked a partial fix.)

I am not touching the memory-core lane for this diff; happy to re-run once that lane is green if you want a clean datapoint.

Finn763:
Root cause of the red `CI` on the previous head — it was a repo-side breakage that has since been repaired, so I rebased instead of touching the failing lane.

The failing job was `checks-node-changed-extensions-config-49` (plus the downstream `openclaw/ci-gate`), and the single failing test was:

```
extensions/memory-core/src/memory/manager-temporal-ranking.test.ts:156
memory source temporal ranking > decays SQLite session hits by recorded source activity through FTS unavailable
AssertionError: expected { Object (enabled, available, ...) } to match object { enabled: true, available: false }
```

Two independent reasons that is not this diff:

1. **Wrong package entirely.** This PR changes `src/auto-reply/reply/stage-sandbox-media.ts` (a deadline constant passed as `timeoutMs` to the existing shared runner) and its test. It cannot reach `extensions/memory-core`'s temporal-ranking FTS availability decision.
2. **It was already being repaired upstream, an hour after my run.** steipete's `#144761 test(memory-core): repair temporal-ranking CI after FTS reconciliation` — which touches exactly this test file — **merged at 2026-09-11T08:12:25Z** (`1bf78261b73e`), while my failed run finished at **07:05:19Z**.

So the correct action was to move this branch onto a base that contains the repair rather than to "fix" anything. Done: rebased onto `b43dfe93743` (the repair commit is an ancestor of it), which is why the head here changed to `7826babba66` — the reviewable content is unchanged, and the local run on the rebased head is green:

```
node node_modules/vitest/vitest.mjs run --config test/vitest/vitest.auto-reply-reply.config.ts src/auto-reply/reply/stage-sandbox-media.scp.test.ts
Test Files  1 passed (1)
Tests  13 passed | 2 skipped (15)
```

CI on the new head is queued. The two skipped cases remain the POSIX-gated hang cases described in the PR body — still the one thing I cannot observe on Windows, and still why this is marked a partial fix.

Finn763:
Addressed all five items from the review above; head is `8707115c920` and the PR body now carries both the revised policy and the real-transport evidence.

1. **Real behavior proof** — supplied, no mocks: a compiled stand-in transport is spawned through the same argv/PATH path the runner uses, so the real spawn, deadline, process-tree kill and temp cleanup all execute (Windows). Raw output: `slow transfer staged=1 elapsed=7965ms`, `reaped by deadline: parent=true descendant=true`, `attempts="attempt\n"` (one attempt), `temp dir removed=true`. Full run is in the PR body.
2. **P1 preserve slow successful transfers** — fixed by revising the budget: derived from the size cap and the slowest supported throughput (256 KiB/s) → **200s at the 50 MiB cap**, 160s for the review's 40 MiB at 1 MiB/s example, 30s floor for small transfers, `OPENCLAW_SANDBOX_SCP_TIMEOUT_MS` as the operator override. Still an absolute bounded deadline owned by the shared runner, so an unattended hang still dies.
3. **P1 merge risk** — the concern was that every attempt received an unconditional 30s wall-clock limit. There is no unconditional cutoff now, and a deadline kill is no longer retried (one bounded wait instead of three), so a stalled copy costs one budget rather than three attempts.
4. **P2 next step** — done, see (1) and the PR body evidence section.
5. **Maintainer decision** — answered as "revise the transfer budget" with the numbers above as the conservative default. A different budget, or a config field instead of the env override, is a two-line change: name the value and I'll switch it.

Finn763:
CI attribution on head `8707115c920` — two of the three red lanes are this diff's, and one of them cannot go green without a decision from the ratchet owner.

**1. `checks-fast-baseline-ratchets` — the new operator env name.** The diff adds `OPENCLAW_SANDBOX_SCP_TIMEOUT_MS`, which is one new `OPENCLAW_*` name in production source. Locally on this head:

```
$ node --import ./scripts/tsx.mjs scripts/check-env-var-count.mts --base HEAD
OPENCLAW_* count 494 exceeds budget 493; update config/env-var-count-budget.txt
```

and if the budget file is bumped instead of the count lowered:

```
OPENCLAW_* budget grew from 493 to 494
```

`enforceRatchetScalar` fails on both sides, and `config/env-var-count-budget.txt` says the value is only lowered without owner approval (its last three changes on `main` are all decreases). So this lane stays red while the env override is in the diff. Two ways out, both one-liners:

- **drop the override** and keep the size-derived deadline (`50 MiB / 256 KiB/s → 200 s`, 30 s floor) — no new name, lane green;
- **the ratchet owner raises the budget** (or approves a config field instead of an env name) — I rebase and the lane is green.

That is the same question the review's "Owner decision" item is waiting on; naming the value settles both.

**2. `check-dependencies` — `deadcode:exports` / `deadcode:dependencies`.** The only new exports in this diff are `SCP_MIN_SUPPORTED_THROUGHPUT_BYTES_PER_SEC`, `SCP_TRANSFER_TIMEOUT_FLOOR_MS` and `resolveScpTransferTimeoutMs`, all consumed by the focused test rather than by another production module. Caveat: I could not reproduce knip locally — its `oxc-parser` scan dies with `RangeError: Array buffer allocation failed` on this box (171 MB page file) — so the exact symbols are from the diff, not from a local run.

**3. `checks-node-compact-small-48`** carries the runner annotation `The self-hosted runner lost communication with the server` — infrastructure, not this change.

`openclaw/ci-gate` is the rollup of the three. No code pushed with this note.

PR #144740, opened 2026-09-11
fix: refreshing the plugin catalog with a trailing slash returns 404

### LINKED ISSUE #144721: [Bug]: Refreshing Plugins page returns 404 Not Found due to UI/plugin HTTP route collision (2026.9.4)
### Bug type

### Beta release blocker

No

### Summary

On OpenClaw 2026.9.4, navigating internally to the Control UI **Plugins** catalogue works, but refreshing the browser at `/plugins` replaces the application with plain-text **Not Found** (HTTP 404). Direct requests to the Gateway reproduce the failure without the external reverse proxy.

The client registers `/plugins` as a UI route, while the root-mounted Gateway UI classifier explicitly excludes `/plugins` and all descendants as plugin-owned HTTP routes.

### Steps to reproduce

1. Run OpenClaw 2026.9.4 with the Control UI mounted at the origin root (no custom `basePath`).
2. Open the root Control UI and navigate internally to **Plugins**; the browser URL becomes `/plugins` and the catalogue page renders.
3. Refresh the browser at that URL.
4. Observe that the application is replaced by **Not Found**.

A browser-independent transport reproduction, run on the Gateway host with its default port:

```sh
curl -i -H 'Accept: text/html' http://127.0.0.1:18789/plugins
curl -i -H 'Accept: text/html' http://127.0.0.1:18789/plugins/
```

Both return HTTP 404 with `Content-Type: text/plain; charset=utf-8` and body `Not Found`.

For comparison:

```sh
curl -i -H 'Accept: text/html' http://127.0.0.1:18789/skills
curl -i -H 'Accept: text/html' http://127.0.0.1:18789/settings/plugins
```

These return HTTP 200 and the Control UI HTML document on the same running Gateway.

### Expected behavior

Refreshing a registered Control UI catalogue route should serve the application shell and allow the client router to restore that page, consistent with `/skills` and `/sessions` on this deployment.

Plugin-owned HTTP/API endpoints must retain their routing and authentication semantics; serving the SPA indiscriminately under the entire `/plugins/*` namespace would not be a safe general fix.

### Actual behavior

The initial client-side navigation works, but a full document request to the same URL returns plain text instead of the Control UI shell.

Direct Gateway checks with `Accept: text/html`, repeated during investigation:

| Path | HTTP status | Content-Type | Result |
| --- | ---: | --- | --- |
| `/` | 200 | `text/html; charset=utf-8` | Control UI HTML |
| `/plugins` | 404 | `text/plain; charset=utf-8` | `Not Found` |
| `/plugins/` | 404 | `text/plain; charset=utf-8` | `Not Found` |
| `/skills` | 200 | `text/html; charset=utf-8` | Control UI HTML |
| `/sessions` | 200 | `text/html; charset=utf-8` | Control UI HTML |
| `/settings/plugins` | 200 | `text/html; charset=utf-8` | Control UI HTML |

The failing document request does not reach UI rendering. This reproduction does not depend on catalogue latency, a missing JavaScript asset, or the external ingress.

### OpenClaw version

2026.9.4 (`3a9d69d`), confirmed against the running Gateway and matching release source.

### Operating system

Gateway container: Debian GNU/Linux 12 (bookworm), hosted on OpenShift. Browser version was not captured; the HTTP failure was independently reproduced on localhost.

### Install method

Container deployment on OpenShift; manually upgraded to 2026.9.4. Root-mounted Control UI, no configured custom UI base path.

### Model

N/A — Control UI document routing, no model inference involved.

### Provider / routing chain

Browser → external ingress → Gateway in normal use. The same 404 is reproduced with a direct localhost → Gateway request, bypassing ingress.

#### Conflicting release-source routes

1. [`ui/src/app-route-paths.ts`, lines 84–88](https://github.com/openclaw/openclaw/blob/3a9d69d/ui/src/app-route-paths.ts#L84-L88), declares two distinct UI routes:

   ```ts
   plugins: { path: "/plugins" },
   "plugin-settings": { path: "/settings/plugins" },
   ```

2. [`ui/src/pages/plugins/route.ts`, lines 78–80](https://github.com/openclaw/openclaw/blob/3a9d69d/ui/src/pages/plugins/route.ts#L78-L80), binds these to the `discovery` and `settings` surfaces respectively.

3. [`src/gateway/control-ui-routing.ts`, lines 64–70 and 94–98](https://github.com/openclaw/openclaw/blob/3a9d69d/src/gateway/control-ui-routing.ts#L64-L98), computes HTML SPA eligibility but, in the `!basePath` branch, unconditionally declines this namespace:

   ```ts
   if (pathname === "/plugins" || pathname.startsWith("/plugins/")) {
     return { kind: "not-control-ui" };
   }
   ```

   Consequently, the registered catalogue document path is never served by this root SPA fallback, even with an explicit HTML Accept header. The live requests above end in Gateway 404s.

4. [`src/gateway/control-ui-routing.ts`, lines 17–29](https://github.com/openclaw/openclaw/blob/3a9d69d/src/gateway/control-ui-routing.ts#L17-L29), explicitly recognizes `/settings/plugins` as the plugin recovery/management surface. Its successful response does not resolve the separate `/plugins` discovery-route conflict.

The classifier was also checked in the installed `control-ui-share-DC4YX1OI.mjs` bundle; this is not solely a finding against an unrelated source checkout.

### Impact and severity

Operators refreshing the Plugins catalogue lose the application page and see a bare 404. Direct opening of the same catalogue URL also fails at the HTTP layer. The tested root-mounted deployment reproduces the failure on repeated direct requests; other base-path configurations have not been runtime-tested. No data loss or plugin installation failure was observed.

**Workaround:** open the root Control UI, then navigate to Plugins internally. `/settings/plugins` remains reachable, but it is the separate management surface, not an equivalent catalogue URL.

### Additional information

Possible repair direction: reconcile the public catalogue URL and Gateway route ownership—either use a non-conflicting catalogue namespace or explicitly distinguish valid UI document routes from plugin-owned HTTP endpoints. Keep existing plugin endpoint routing, authorization, non-HTML responses, and malformed-path behavior intact.

Suggested regression coverage:

- Full document GET and browser refresh for the catalogue root, with and without a trailing slash.
- Catalogue detail document routes, which are also constructed below `/plugins/` in the client; their live behavior was not separately tested in this report.
- Root-mounted and configured-base-path variants.
- Plugin-owned HTTP routes and non-HTML/API requests are not swallowed by an SPA fallback.
- `/settings/plugins` remains available as the distinct recovery/management surface.

Observed after the 9.4 upgrade; no last-known-good refresh baseline was captured, so no verified regression range is claimed. Separate report #144719 covers slow plugin-category population; this issue is only the document-route 404.

gokay-ai:
## What Problem This Solves

Opening or refreshing the Plugins catalog with a trailing slash returned a plain-text 404 when the Control UI was served at the website root.

Fixes #144721

Original fix and commit authorship: @gokay-ai. Original report: @Conan-Scott.

## Why This Change Was Made

The Gateway document classifier treated the catalog's trailing-slash root as a plugin request. It now recognizes that root as a catalog page before resolving individual plugin pages.

## User Impact

Direct navigation and refresh at `/plugins/` load the catalog. Plugin settings and plugin HTTP endpoints keep their existing behavior.

## Evidence

- In a real browser against an isolated Gateway, catalog navigation and reload returned HTTP 404 before the fix and HTTP 200 with the visible catalog after it.
- The slashless catalog and separate plugin settings page continued to load.
- The Diffs plugin's JavaScript endpoint continued to return JavaScript.
- [Sanitized before/after browser captures](https://gist.github.com/obviyus/dd01be4e2f59b5996026da9c3df85f76) show the failed refresh and restored catalog.

## Consumers

- Browser visitors to `/plugins/`: direct navigation and refresh now open the Plugins catalog.

### DIFF
--- a/src/gateway/control-ui-routing.test.ts
+++ b/src/gateway/control-ui-routing.test.ts
@@ -195,6 +195,12 @@ describe("classifyControlUiRequest", () => {
         { kind: "serve" as const, spaFallback: true },
       ],
       [
+        "serves the marketplace document at the plugin root slash",
+        "/plugins/",
+        "GET",
+        { kind: "serve" as const, spaFallback: true },
+      ],
+      [
         "keeps API routes outside the SPA catch-all",
         "/api/sessions",
         "GET",
@@ -356,6 +362,42 @@ describe("classifyControlUiRequest", () => {
     });
   });
 
+  describe("plugin catalogue documents vs plugin HTTP", () => {
+    it.each([
+      {
+        name: "serves HTML GET for the catalogue root slash",
+        pathname: "/plugins/",
+        method: "GET",
+        accept: "text/html",
+        expected: { kind: "serve" as const, spaFallback: true },
+      },
+      {
+        name: "keeps JSON catalogue reads outside the SPA",
+        pathname: "/plugins",
+        method: "GET",
+        accept: "application/json",
+        expected: { kind: "not-control-ui" as const },
+      },
+      {
+        name: "keeps JSON catalogue slash reads outside the SPA",
+        pathname: "/plugins/",
+        method: "GET",
+        accept: "application/json",
+        expected: { kind: "not-control-ui" as const },
+      },
+    ])("$name", ({ pathname, method, accept, expected }) => {
+      expect(
+        classifyControlUiRequest({
+          basePath: "",
+          pathname,
+          search: "",
+          method,
+          accept,
+        }),
+      ).toEqual(expected);
+    });
+  });
+
   describe("basePath-mounted control ui", () => {
     it.each<[string, string, string, string, ReturnType<typeof classifyControlUiRequest>]>([
       [
@@ -370,6 +412,27 @@ describe("classifyControlUiRequest", () => {
         "/openclaw/chat",
         "",
         "HEAD",
+        { kind: "serve" as const, spaFallback: true },
+      ],
+      [
+        "serves the plugin catalogue under a configured basePath",
+        "/openclaw/plugins",
+        "",
+        "GET",
+        { kind: "serve" as const, spaFallback: true },
+      ],
+      [
+        "serves the plugin catalogue slash under a configured basePath",
+        "/openclaw/plugins/",
+        "",
+        "GET",
+        { kind: "serve" as const, spaFallback: true },
+      ],
+      [
+        "serves the plugin manager under a configured basePath",
+        "/openclaw/settings/plugins",
+        "",
+        "GET",
         { kind: "serve" as const, spaFallback: true },
       ],
       [
--- a/src/gateway/control-ui-routing.ts
+++ b/src/gateway/control-ui-routing.ts
@@ -92,11 +92,12 @@ export function classifyControlUiRequest(params: {
     if (classifyNodeWorkspaceTransferPath(pathname) !== "outside") {
       return { kind: "not-control-ui" };
     }
-    // Marketplace documents own the root and canonical generated catalog IDs.
+    // Marketplace documents own the catalogue root and canonical generated catalog IDs.
     // Other descendants and non-document requests remain plugin HTTP routes.
     if (pathname === "/plugins" || pathname.startsWith("/plugins/")) {
       const marketplaceDocument =
         pathname === "/plugins" ||
+        pathname === "/plugins/" ||
         resolvePluginDiscoveryIdentity(pathname.slice("/plugins/".length)) !== undefined;
       if (!marketplaceDocument || !isReadHttpMethod(method) || !spaFallback) {
         return { kind: "not-control-ui" };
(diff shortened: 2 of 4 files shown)


obviyus:
Thanks @gokay-ai for the fix and @Conan-Scott for the report!

**Proof:** Direct browser navigation and reload against an isolated root-mounted Gateway now show the Plugins catalog. The slashless catalog, separate plugin settings page, and real Diffs JavaScript endpoint retain their behavior.

- **Bug:** `/plugins/` returned a plain-text 404 on refresh.
- **Fix:** The shared document classifier recognizes the trailing-slash catalog root before resolving descendant identities.

| Before | After |
| --- | --- |
| HTTP 404, `Not Found` | HTTP 200 HTML, visible catalog after reload |

263 targeted tests passed. Exact-head local review found no code or security issues. Contributor authorship is retained.

[Sanitized before/after captures](https://gist.github.com/obviyus/dd01be4e2f59b5996026da9c3df85f76)

PR #144741, opened 2026-09-11
refactor(tests): reuse Vydra speech response fixtures

steipete:
## What Problem This Solves

Two Vydra speech tests duplicated response queues already covered by the plugin's shared JSON, binary-response, and fetch fixtures.

## Why This Change Was Made

Reuse those existing helpers and remove the duplicated setup. All source outside the two queues and their import is unchanged, including every assertion and the separate oversized-JSON fixture. The change adds nine test lines and removes 40: **31 fewer lines**, with no production change.

The helper import moves existing auth bindings and test-global initialization earlier. Existing media setup initializes the registry before collection, and speech already loads that auth graph before resolving synthesis credentials. The speech tests do not call the helper's optional auth-stubbing function.

## User Impact

No provider runtime or API behavior changes. This is test-fixture consolidation.

## Evidence

All **34 tests passed before and after** under the same normal media configuration: six speech, six image, seven video, and 15 shared cases, with identical per-file inventory and order. The 19-check changed gate passed, and independent source review found no actionable P0-P2 issue. Source/dependency hashes stayed unchanged during proof; the launcher was joined and its group checked at terminal. These are local fixture tests, not live Vydra API proof.

### DIFF
--- a/extensions/vydra/speech-provider.test.ts
+++ b/extensions/vydra/speech-provider.test.ts
@@ -2,6 +2,7 @@ import { bufferedOversizedJsonResponse as oversizedJsonResponse } from "openclaw
 // Vydra tests cover speech provider plugin behavior.
 import { installPinnedHostnameTestHooks } from "openclaw/plugin-sdk/test-media-understanding";
 import { afterEach, describe, expect, it, vi } from "vitest";
+import { binaryResponse, jsonResponse, stubFetch } from "./provider-test-helpers.js";
 import { buildVydraSpeechProvider } from "./speech-provider.js";
 
 describe("vydra speech provider", () => {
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3cf2069a0b0cc5f5a2e24be09eb5e611344b9c76](https://github.com/openclaw/openclaw/pull/144741/commits/3cf2069a0b0cc5f5a2e24be09eb5e611344b9c76)
- Landed commit: [bb221a85658ce9d8798f1c2f9e8f6de4391f6512](https://github.com/openclaw/openclaw/commit/bb221a85658ce9d8798f1c2f9e8f6de4391f6512)

PR #144770, opened 2026-09-11
fix(launcher): report external child kills as 128+signum, not exit 1

### LINKED ISSUE #144200: [Bug]: Source launcher converts child signals into normal exit codes
### Bug type

### Beta release blocker

No

### Summary

The Unix source launcher converts a respawned runtime child's actual termination signal into a normal numeric exit, so its caller loses the distinction between interrupted execution and an acknowledged shutdown.

### Steps to reproduce

The following is the direct-child SIGKILL scenario from a completed six-case launcher probe:

1. Use OpenClaw commit `d70f83517220507b6719f1286bcba08462134e24` and Node 24.18.1 on Linux. Copy its unmodified `openclaw.mjs`, `node-runtime-recovery.mjs`, `node-runtime-update.mjs`, `node-sqlite.mjs`, and `node-version.mjs` into a temporary directory.
2. Add `package.json` containing `{"type":"module"}`, `src/entry.ts` containing `export {};`, and this controlled `dist/entry.js`:

   ```js
   process.stdout.write(JSON.stringify({
     pid: process.pid,
     respawn: process.env.OPENCLAW_COMPILE_CACHE_DISABLED_RESPAWNED,
   }) + "\n");
   setInterval(() => {}, 1000);
   ```

3. From a Node observer, spawn `node <fixture>/openclaw.mjs` with piped stdout and `NODE_COMPILE_CACHE=<fixture>/cache`. Wait for the JSON line. Confirm `respawn` is `"1"` and the reported runtime PID differs from the launcher PID.
4. Send `SIGKILL` only to that runtime PID with `process.kill(runtimePid, "SIGKILL")`. Leave the launcher alive to handle its child's exit.
5. Observe the launcher's `child_process` `exit` event: it reports `(code: 1, signal: null)`.

The probe used isolated temporary fixtures and confirmed both processes exited before removing them. This exercises the real launcher modules with a controlled entry, rather than a full Gateway.

### Expected behavior

Preserve the runtime child's actual Unix signal as `(code: null, signal: "SIGKILL")`. Preserve an explicitly returned numeric exit code as numeric, including `143`, and preserve an acknowledged shutdown returning `0`.

The distinction is already required by the [source runner's interruption handling](https://github.com/openclaw/openclaw/blob/d70f83517220507b6719f1286bcba08462134e24/scripts/run-node.mts#L1196-L1208), which treats a signal as unacknowledged termination.

### Actual behavior

The completed baseline probe observed:

| Runtime outcome | Launcher exit code | Launcher exit signal |
| --- | ---: | --- |
| Explicitly returned 23 | 23 | null |
| Explicitly returned 143 | 143 | null |
| Handled SIGTERM and returned 0 | 0 | null |
| Terminated by forwarded SIGTERM | 143 | null |
| Killed directly with SIGKILL | 1 | null |
| Ignored SIGTERM, then killed by shutdown escalation | 1 | null |

### OpenClaw version

`d70f83517220507b6719f1286bcba08462134e24`

### Operating system

Linux; Node 24.18.1. Linux distribution/version: NOT_ENOUGH_INFO.

### Install method

Source launcher modules from the exact commit, executed with Node and controlled entry fixtures.

### Model

None; this occurs before model execution.

### Provider / routing chain

None.

The [source compile-cache branch](https://github.com/openclaw/openclaw/blob/d70f83517220507b6719f1286bcba08462134e24/openclaw.mjs#L126-L146) reaches `runRespawnedChild` when `NODE_COMPILE_CACHE` is inherited, including on the tested supported Node version. Its [exit handler](https://github.com/openclaw/openclaw/blob/d70f83517220507b6719f1286bcba08462134e24/node-runtime-recovery.mjs#L151-L164) explicitly maps actual signals to numeric codes. Existing [launcher tests](https://github.com/openclaw/openclaw/blob/d70f83517220507b6719f1286bcba08462134e24/test/openclaw-launcher.e2e.test.ts#L1330-L1427) expect these numeric outcomes.

### Impact and severity

Affected: callers supervising the Unix source launcher when this respawn branch is active. The observed consequence is loss of termination identity: SIGTERM becomes indistinguishable from an explicit exit 143, while SIGKILL becomes an ordinary exit 1. Each listed scenario was executed once in the completed baseline probe. Full Gateway/watch-loop consequences and production frequency were not measured.

### Additional information

The built-entry sibling is `src/process/respawn-child-runner.ts`, used by compile-cache and warning respawns. A fix should preserve genuine numeric completions and acknowledged shutdowns while retaining the existing Windows termination behavior. This report does not establish a last known good version.

he-yufeng:
## Evidence

Fixes #144200. Verified against current main with a fixture reproducing the issue's probe: a source-checkout launcher (`openclaw.mjs` + `node-runtime-recovery.mjs` + `node-sqlite.mjs` + `node-version.mjs` copied unmodified, a controlled `dist/entry.js` that prints its pid and idles). Sending SIGKILL to the respawned child's pid makes the launcher report `(code: 1, signal: null)` on current main and `code: 137` with this change.

## What changed

`runRespawnedChild`'s exit handler (`node-runtime-recovery.mjs`) fell through to `process.exit(1)` whenever the child died from a signal the launcher never forwarded. Unforwarded signal deaths now exit with the shell's 128+signum convention (SIGKILL -> 137); forwarded-signal exits keep their existing 130/143 codes, and the hard-kill backstop path is unchanged.

## Tests

An E2E case in `test/openclaw-launcher.e2e.test.ts` pins the 137 contract on the same fixture shape as the neighboring signal-forwarding cases. The repo's e2e vitest config could not run on this checkout (a pre-existing TUI build failure in this sparse tree blocks its global setup), so the contract was verified by running the fixture directly against current main and this change; both runs are quoted in Evidence.

## Notes

- macOS arm64, Node 26.5.0.
- The launcher comment asking to keep `src/entry.compile-cache.ts` in sync no longer applies to this handler: on current main the forwarding logic lives only here.

[AI-assisted]

### DIFF
--- a/node-runtime-recovery.mjs
+++ b/node-runtime-recovery.mjs
@@ -159,7 +159,15 @@ export const runRespawnedChild = (command, args, env) => {
               ? 143
               : undefined
           : undefined;
-      process.exit(forwardedSignalExitCode ?? 1);
+      if (forwardedSignalExitCode !== undefined) {
+        process.exit(forwardedSignalExitCode);
+      }
+      // The child died by a signal we never forwarded (e.g. an external
+      // SIGKILL): report it with the shell's 128+signum convention so the
+      // caller can still tell an interrupted run from an ordinary exit 1
+      // (#144200).
+      const signalNumber = os.constants.signals[signal];
+      process.exit(signalNumber ? 128 + signalNumber : 1);
     }
     process.exit(code ?? 1);
   });
--- a/test/openclaw-launcher.e2e.test.ts
+++ b/test/openclaw-launcher.e2e.test.ts
@@ -1377,6 +1377,56 @@ describe("openclaw launcher", () => {
   });
 
   it.runIf(process.platform !== "win32")(
+    "reports 128+signum when an external signal kills the respawn child (#144200)",
+    async () => {
+      const fixtureRoot = await makeLauncherFixture(fixtureRoots);
+      await addGitMarker(fixtureRoot);
+      const childInfoPath = path.join(fixtureRoot, "child-info.json");
+      await fs.writeFile(
+        path.join(fixtureRoot, "dist", "entry.js"),
+        [
+          'import { writeFileSync } from "node:fs";',
+          `writeFileSync(${JSON.stringify(childInfoPath)}, JSON.stringify({ pid: process.pid }) + "\\n");`,
+          'process.title = "openclaw-launcher-external-kill-test-child";',
+          "setInterval(() => {}, 1000);",
+          "",
+        ].join("\n"),
+        "utf8",
+      );
+
+      const launcher = spawn(process.execPath, [path.join(fixtureRoot, "openclaw.mjs")], {
+        cwd: fixtureRoot,
+        env: launcherEnv({
+          NODE_COMPILE_CACHE: path.join(fixtureRoot, ".node-compile-cache"),
+        }),
+        stdio: "ignore",
+      });
+      let respawnChildPid: number | undefined;
+
+      try {
+        const childInfo = await waitForJsonFile<{ pid: number }>(childInfoPath, 5000);
+        respawnChildPid = childInfo.pid;
+
+        // Kill the child from outside: the launcher must not report this as a
+        // plain exit 1, which reads as an ordinary failure instead of a kill.
+        process.kill(respawnChildPid, "SIGKILL");
+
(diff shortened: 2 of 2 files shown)


shakkernerd:
Thanks for taking the time to work on this and put together a PR! #144200 is now fixed by #144838, so I've closed this PR as superseded. I appreciate the contribution.

PR #144713, opened 2026-09-11
fix(update): unblock pnpm 12 global updates

steipete:
Fixes #144667

## What Problem This Solves

Fixes an issue where users updating a global pnpm installation would stop at `validating` / `pnpm staging preflight` with pnpm 12.1.0, even though OpenClaw had requested a temporary bin directory. Thanks to @xilopaint for the report and package-manager comparison.

## Why This Change Was Made

The staging owner now sets both existing pnpm environment spellings of `global_bin_dir` to the staged bin. It retains the CLI override required by pnpm 10/11. The same environment and arguments reach destination probes and installation; the guard still refuses a manager that reports a live destination.

pnpm 12.1.0's [config resolver](https://github.com/pnpm/pnpm/blob/v12.1.0/pnpm/crates/config/src/lib.rs#L3740-L3758) reads the pnpm environment override before falling back to `PNPM_HOME/bin`; [lowercase takes precedence](https://github.com/pnpm/pnpm/blob/v12.1.0/pnpm/crates/config/src/lib.rs#L3941-L3949). Its [CLI override parser](https://github.com/pnpm/pnpm/blob/v12.1.0/pnpm/crates/cli/src/config_overrides.rs#L84-L98) silently drops unimplemented keys, including `global-bin-dir`. pnpm 11 [derives the bin before its environment overlay](https://github.com/pnpm/pnpm/blob/v11.24.0/pnpm11/config/reader/src/index.ts#L422-L437), so its CLI configuration remains necessary.

## User Impact

Global pnpm 12 updates can stage launchers without changing the live bin directory before activation. No new configuration option, environment-variable name, dependency, schema, or recovery policy is introduced. The environment-variable budget remains 493/493.

Release-note context: fix pnpm 12 global updates stopping at the isolated staging-bin preflight; preserve pnpm 10/11 behavior and the live-destination refusal. Credit: @xilopaint.

## Evidence

Read-only `corepack pnpm@<version> bin -g --config.global-bin-dir=<stage-bin>` probes, with both bin directories on PATH:

| pnpm | CLI override only | CLI plus staged pnpm environment |
| --- | --- | --- |
| 11.24.0 | stage bin, exit 0 | stage bin, exit 0 |
| 12.1.0 | live bin, exit 0 | stage bin, exit 0 |

The transaction regression fixtures pnpm 12's manager response at the existing command boundary. Before the production fix:

```text
FAIL preserves pnpm11 native project ownership (pnpm12=true, ...)
name: pnpm staging preflight
[...]

Issue #144667, opened 2026-09-11
[Bug]: pnpm 12.1.0 global update fails staging preflight because global-bin-dir override is ignored


xilopaint:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

A global pnpm installation of OpenClaw 2026.9.3 cannot update to 2026.9.4 with pnpm 12.1.0: the native staging preflight requests a temporary global bin directory through `--config.global-bin-dir`, but `pnpm bin -g` reports the live bin directory, so OpenClaw correctly refuses to proceed.

### Steps to reproduce

Observed update path:

1. Use OpenClaw 2026.9.3 installed globally in pnpm's isolated `global/v11` layout, with pnpm 12.1.0 available through Corepack and a user-systemd Gateway.
2. Run `openclaw update` on the stable channel (target resolved to 2026.9.4).
3. Observe failure during `validating`, at `pnpm staging preflight`, before live package activation.

Minimal non-mutating package-manager probe, reproduced separately:

```sh
pnpm --version
# 12.1.0
pnpm bin -g --config.global-bin-dir=/tmp/openclaw-diagnostic-bin
# Returns the live PNPM_HOME/bin, not /tmp/openclaw-diagnostic-bin
```

The same flag was also tested with the diagnostic bin directory prepended to PATH. A directly invoked cached pnpm 11.24.0 launcher returned the requested directory. No global pnpm version switch was performed.

### Expected behavior

OpenClaw should select and verify an isolated staging bin directory with the owning package manager before installation. The read-only probe with pnpm 11.24.0 honors the requested directory. The fail-closed guard must remain in place when a manager reports a live destination.

### Actual behavior

Two observed update attempts failed at the staging preflight. The updater's sanitized diagnostic identifies a mismatch between the live bin directory and the expected temporary destination. The new package was not activated; the existing Gateway remained healthy on 2026.9.3.

### OpenClaw version

2026.9.3, build `2026.9.3-release-1391f7cd2d40-2026-09-08T07-46-00.264Z`; attempted target 2026.9.4.

### Operating system

Ubuntu 24.04.4 LTS, Linux aarch64.

### Install method

Global pnpm isolated `global/v11` installation; user-systemd Gateway; pnpm 12.1.0 via Corepack.

### Model

Not applicable: CLI/package-manager update preflight.

### Provider / routing chain

Not applicable: reproduced without an inference request.

### Additional provider/model setup details

[...]

PR #144761, opened 2026-09-11
test(memory-core): repair temporal-ranking CI after FTS reconciliation

steipete:
Refs #144641, #144714, #144715.

## What Problem This Solves

Resolves a problem where PRs routed through the changed-extensions CI lane fail the memory temporal-ranking test after #144641. The failure blocks unrelated update repairs, including #144714 and #144715. This is a maintainer-requested repair of that CI regression.

## Why This Change Was Made

The FTS-unavailable fixture replaces the derived index with a SQLite view. That view mirrored every canonical chunk, so the row-count comparison introduced by #144641 skipped the write that previously exposed the view as unusable. Making the view empty forces reconciliation to attempt a write, which SQLite rejects because the target is a view. The test also verifies that specific initialization error.

The case remains valuable: semantic-only fallback must rank session hits by recorded source activity through both the manager and the memory_search tool. All existing ranking and provenance assertions remain intact. Production FTS reconciliation is unchanged.

## User Impact

Restores meaningful CI coverage for update-repair PRs. There is no runtime behavior change, new configuration, production schema change, or persistence-policy change.

## Evidence

On the unchanged base, `node scripts/run-vitest.mjs extensions/memory-core/src/memory/manager-temporal-ranking.test.ts` exited 1:

```text
FAIL ... decays SQLite session hits by recorded source activity through FTS unavailable
- "available": false,
+ "available": true,
Tests  1 failed | 5 passed (6)
```

With the fixture repair, the same command exited 0:

```text
PASS ... decays SQLite session hits by recorded source activity through FTS unavailable
Test Files  1 passed (1)
Tests  6 passed (6)
```

`node scripts/run-vitest.mjs extensions/memory-core` exited 0:

```text
Test Files  112 passed | 2 skipped (114)
Tests  1639 passed | 8 skipped (1647)
```

`git diff --check` passes. `node scripts/check-changed.mjs` passes formatting, extension-test typechecking, boundary guards, ratchets, and dead-export scans, then exits 1 before lint: declaration preparation rejects ancestor-install dependency manifests in this nested worktree (`Declaration input escapes checkout`). A separate physical checkout is outside the assigned worker scope, so no isolation guard was bypassed.

[...]

PR #144778, opened 2026-09-11
fix(update): retain ownership across pnpm package generations

obviyus:
Related: #144667, #144713.

Thanks @xilopaint for the report and @steipete for the staging fix and finalizer reproduction.

## What Problem This Solves

Fixes an issue where package-generation changes during an update could fail finalizer binding, then fail rollback ownership checks or restart without revalidating the restored service.

## Why This Change Was Made

Bind finalization to the activated package generation while preserving recovery ownership for the original installation. Revalidate service ownership before rollback shutdown and after restoring service installation. The staging fix from #144713 remains unchanged.

## User Impact

An updater containing this repair can complete the update or restore the previous healthy Gateway after a compatible failure. Older updaters still require this repair; changed package or service ownership remains a refusal.

## Evidence

Public command-line tests with isolated Linux services confirmed updates to 2026.9.4 and rollback to a healthy 2026.9.3 after a controlled failure. The tested package predates final integration; source comparison justified reuse, and 130 focused integration cases passed. Default-linker service-root discovery remains a separate refusal.


steipete:
Coordination from the update-reliability watch: #144713 carries the same staging-preflight fix (both pnpm bin environment spellings set for the staging child, CLI config kept for pnpm 10/11; pnpm 12.1.0 resolves the bin dir from its environment overlay per `pnpm/crates/config/src/lib.rs` at v12.1.0, pnpm 11.24.0 derives it before the overlay). It is prepared and autoreviewed with live proof on a macOS arm64 pnpm 12 rig (stock 9.3 → 9.4 reproduces the refusal; patched driver completes a tarball update), but its merge has been stalled by proxy timeouts. This PR is the superset because it also binds the finalizer to the activated package generation, which our matrix lane hit as `Candidate executor binding does not match its parent` after pnpm relocated the project directory on a schema-crossing update. So: if this PR lands first I will close #144713 as superseded; if #144713 happens to land first the overlap is the one hunk in `update-native-package-stage.ts` and should rebase trivially.

Two things that may help here:

- An independent reproduction of the finalizer binding refusal is being finished on a pnpm 12 rig with a 2026.9.3 state at shared-schema 16 (the delegated finalizer only runs when the update crosses a schema version); I will post the scenario and roots so it can be run against this head.
- Current CI failures on this head: `checks-node-changed-extensions-config-49` is the memory-core fixture broken by #144641 and repaired by #144761 (landing now), so a rerun after that merge should clear it; `checks-node-compact-small-16` I have not attributed yet.

steipete:
Update: #144713 landed as c8b96cb26c9 (the proxy finally let the merge through), so the staging-preflight half is on main. The overlap with this PR is the single hunk in `update-native-package-stage.ts` (both pnpm bin environment spellings set for the staging child); a rebase should either apply cleanly or drop that hunk, leaving this PR as the finalizer-generation binding plus its recovery-custody handling. #144761 also landed, so a CI rerun here should clear the memory-core lane failure.

steipete:
Independent reproduction of the finalizer binding failure, read against head d44e58137ea (nothing from this PR executed locally):

Rig: macOS arm64, Corepack pnpm 12.1.0 with the default isolated linker, no service manager. Stock 2026.9.3 installed, its Gateway started and stopped once so the shared state sits at schema 16. Then a driver carrying the staging fix from #144713 updated to a candidate built from main (schema 17): swap and post-activation Doctor exited 0, the schema became 17, and the delegated finalizer (child of the updater) failed with `Candidate executor binding does not match its parent.` The parent lease still named the pre-swap pnpm project directory (`global/v11/8b4d-…-0`) while the activated candidate lived in `global/v11/9ec6-…-0`; the child lease stayed under the old package root. Without a schema crossing the finalizer is never delegated, which is why same-schema runs pass.

How this PR relates: the producer change (`update-command-executor.ts` around 224–256) acquires the parent lease at the activated root and sends a matching root/parent/child key, and the cleanup ordering keeps the original recovery exclusion through candidate settlement, so a driver that contains this PR handles the relocation. The receiver (around 94–110) is unchanged and still requires the grant root to equal the candidate's resolved root, so it still rejects a grant produced by an older driver. That leaves one shipped path: a pnpm 11 install on 2026.9.4 (whose staging preflight works) updating to the next release across a schema bump will still fail at the finalizer, because the installed 2026.9.4 driver produces the old-root grant and the candidate, even with this PR, refuses it. pnpm 12 installs on 2026.9.4 die earlier at the staging preflight (#144667) and need a bootstrap driver anyway, which then carries this fix. If you want the candidate side to accept the shipped 2026.9.4 driver's grant when the parent root is the pre-swap project directory of the same installation (same install key/package identity), that is the remaining gap; otherwise the release notes should tell pnpm 11 users to bootstrap with the new CLI.

The scenario is scripted and rerunnable against any driver/candidate tarball pair; happy to run it against a build of this head if useful.

obviyus:
Thanks @xilopaint for the report and @steipete for the staging fix and independent finalizer reproduction.

Landed as 68b7893b3ca977a238af165717f7bc2a40e1a8f0 after exact-head CI and independent local review passed. The merged files match reviewed head `461415fc93c9c`.

**Proof:** Real public pnpm 12.1.0 and 11.24.0 updates reached healthy published 2026.9.4 Gateways. An injected post-activation Doctor exit 73 preserved the failure and restored the package, launcher and healthy 2026.9.3 Gateway; the next normal update succeeded.

- **Bug:** The finalizer received the previous package generation's ownership grant; rollback also lacked complete service-owner revalidation.
- **Fix:** Admit the activated generation through the existing owner while retaining recovery custody, then revalidate the restored service before restart.

| Before | After |
| --- | --- |
| Finalizer binding fails after package relocation | Corrected updater completes the generation transition |
| Compatible failure can leave recovery blocked | Previous healthy package and service are restored under current ownership checks |

The real tested updater remains `f8d27b2053ab`; independent source review qualified its proof for the integrated terminal path. The final integration passed 130 overlapping focused cases, and [CI run 34579907799](https://github.com/openclaw/openclaw/actions/runs/34579907799) passed without an exception. Earlier failures remain retained.

Scope limits remain explicit: this fixes the updater producing the grant, not legacy grants from an unmodified 2026.9.4 updater. The older-updater/future-schema path in [Peter's report](https://github.com/openclaw/openclaw/pull/144778#issuecomment-5631699280) needs separate receiver compatibility proof or a supported corrected-updater bootstrap route. Default-linker service-root discovery is also separate; these live cases use the hoisted linker within pnpm's isolated `global/v11` layout.

PR #144783, opened 2026-09-11
fix(memory): restore FTS fallback for same-name view collisions

steipete:
Related: #144641, #144761

## What Problem This Solves

Fixes an issue where memory searches could report full-text search as available when a SQLite view occupied the body index name, bypassing the intended unavailable-FTS fallback. The fixture repair in #144761 restored CI coverage but left this production detection bug unchanged.

## Why This Change Was Made

The row-count optimization in #144641 exposed a table-only schema lookup: a same-name view could remain in place with matching row counts. The existing FTS schema owner now detects views and rejects the collision before reconciliation, preserving the view and canonical memory rows. It retains SQLite's existing view-collision diagnostic, including the expectation added in #144761. Ordinary derived-table repairs remain unchanged.

## User Impact

Memory search can use its existing fallback when a view shadows the canonical or custom body FTS index. No schema migration or configuration change is required.

## Evidence

- Reproduced the original body-view detection failures before the guard, then reproduced the diagnostic mismatch against the exact merged #144761 fixture before preserving SQLite's wording.
- Passed 53 focused tests across five files on Node 24.20.0, including view preservation, ordinary-table repair, 31 canonical/legacy-schema cases, and all six temporal-ranking modes with the exact merged #144761 fixture. Its assertions were not changed.
- Compatibility proof: schema definitions, versions, configuration, transactions, and recovery rules are unchanged. The canonical/legacy-schema tests exercise existing upgrade paths, and the exact-main fixture passes with the guard. The change introduces no migration.
- The full changed gate passed for the original guard, including production/test typechecks, dead-export scans, lint, and architecture/storage guards. The subsequent message-only adjustment passed the focused suites, formatting, and whitespace checks.
- The original P2 review identified a pre-existing path-trigger recovery limitation; paired original/candidate SQLite proof confirmed it is unchanged by this patch. A separate P2 review of the one-line diagnostic adjustment found no actionable issues. Recovery after externally replacing a live path index remains a separate follow-up.

### DIFF
--- a/packages/memory-host-sdk/src/host/memory-schema-fts-reconcile.test.ts
+++ b/packages/memory-host-sdk/src/host/memory-schema-fts-reconcile.test.ts
@@ -99,10 +99,12 @@ describe("memory FTS schema reconciliation", () => {
     }
   });
 
-  it("repairs malformed derived tables without changing canonical rows", () => {
+  it.each(["fts5", "ordinary"])("repairs malformed %s derived tables", (kind) => {
     const db = new DatabaseSync(":memory:");
     try {
       ensureMemoryIndexSchema({ db, cacheEnabled: false, ftsEnabled: false });
+      const tableDefinition = kind === "fts5" ? "VIRTUAL TABLE" : "TABLE";
+      const columns = kind === "fts5" ? "USING fts5(wrong)" : "(wrong TEXT)";
       db.exec(`
         INSERT INTO memory_index_sources (path, source, hash, mtime, size)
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [341190f5335b220debfc9f6959c946c0c039667d](https://github.com/openclaw/openclaw/pull/144783/commits/341190f5335b220debfc9f6959c946c0c039667d)
- Landed commit: [fdd808c7d5daa89790f16f45c3356f8549474596](https://github.com/openclaw/openclaw/commit/fdd808c7d5daa89790f16f45c3356f8549474596)

PR #144787, opened 2026-09-11
fix(doctor): preserve legacy registry files when quarantine fails

### LINKED ISSUE #144773: [Bug]: preserve malformed legacy sandbox registry when quarantine fails
## Summary

Doctor can delete a malformed legacy sandbox registry after quarantine rename fails, then return an uncreated quarantine path.

## Source and reproduction

Inspected main: `068318110aec933f9c6c5ea1981489d6c39ca283`.
Owner: `src/commands/doctor-sandbox-legacy-registry.ts`, `quarantineLegacyRegistry` and its monolithic migration caller.

In an isolated temporary directory, write malformed registry content and make the intended timestamped quarantine target a non-empty directory. Hold only `Date.now()` at that chosen timestamp; do not fake filesystem operations. Run `migrateLegacySandboxRegistryFiles` with registry paths isolated to the temporary directory. The rename fails because a file cannot replace that directory.

Expected: original registry bytes remain; the rename error is visible, with no successful quarantine result.
Actual on the inspected source: the catch removes the original file and returns the quarantine path. If the source instead disappears immediately before rename, ENOENT is also reported as a successful quarantine.

## Scope and impact

This is potential loss of malformed legacy source data during an explicit repair. It does not establish SQLite corruption, a remotely triggerable vulnerability, broad installation incidence or host-power-loss behavior. Both container and browser monolithic registries share the affected helper.

## Proposed repair and evidence

Propagate non-ENOENT rename errors without deleting the source. Map source disappearance to the existing missing result. Preserve successful migration, sharded processing, locking and empty-file removal.

The source-bound reproduction used Linux temporary files with controlled schema, lock and registry-write dependencies. A repository-native regression with real lock/parser and isolated paths is prepared; native repository validation is not yet claimed. AI-assisted.

## Batch provenance

The implementation was prepared from main at 1b2b06ed2fc7680648a1a554c100b887a1de39dd. Upstream main is now 2579303af1fba3b8d4bdbb6c38abf7c90dcb07b0; a target-footprint recheck found no changes to the affected production or test paths. The issue is filed to make the failure mode and bounded repair reviewable before the PR is opened.

Alix-007:
Closes #144773

## What Problem This Solves

Doctor could delete malformed legacy sandbox registry data after its quarantine rename failed, then report a quarantine location that had not received the bytes. Both container and browser registry repair used the same failure path.

## Why This Change Was Made

The existing quarantine owner now reports success only after rename succeeds. It preserves the source and propagates other rename errors; a source that disappears maps to the existing missing result. Doctor retains its normal warning handling for this optional repair.

The regression cases now share the existing migration fixture and real SQLite writers. The separate mocked-writer fixture is removed, and the successful-quarantine test verifies the actual destination bytes.

## User Impact

Failed quarantine leaves the malformed bytes available for manual repair and reports the failure. Valid migration, existing SQLite rows, sharded precedence, read-only inspection, empty/missing inputs, and maintenance ownership retain their contracts. This changes no schema, configuration, retention policy, dependency, or overall Doctor exit policy.

## Evidence

Pinned main `1d4afe39664428308a2c71a64c94dea7b607bc0a` was exercised through the built public `openclaw doctor --fix --non-interactive` command in isolated offline state. For each registry kind, a real nonempty destination directory caused `rename` to return `EISDIR`; the source was then unlinked and Doctor reported successful quarantine. Original synthetic input and destination bytes were preserved independently.

The corrected source was independently exercised through the same public repair for both registry kinds. The native filesystem trace still records `EISDIR`, but the source and existing destination remain byte-for-byte intact, and Doctor reports:

```text
doctor:sandbox run failed: EISDIR: illegal operation on a directory,
```

| Before | After |
| --- | --- |
| Failed rename was followed by source deletion and a successful-quarantine claim. | Failed rename preserves both inputs and reports failure without claiming quarantine. |
| The claimed destination had not received the malformed bytes. | After resolving the owned collision, repair moves the identical bytes to the reported file; a repeat leaves the result unchanged. |

Independent acceptance also verified real-clock quarantine, registered read-only lint, empty/missing inputs, and real SQLite migration with existing-row conflicts and mixed valid/invalid shards. Whole canonical rows, including JSON and update timestamps, remained unchanged; canonical rows take precedence over shards, and shards over monolithic input. No database writer was mocked in this public proof.

An initial fixture reused an old fixed timestamp and was correctly refused by maintenance before reaching quarantine. Its failed and prematurely batched dependent steps are retained. The corrected fixture uses a contemporaneous timestamp and stops dependent actions on nonzero repair exit. Product leases, workers, permission checks and deadlines were unchanged. The original baseline proof remains valid and was not replayed.

The consolidated regressions produced the intended four failures on baseline. All 34 focused tests across three files pass with the repair. Native formatting, syntax lint, line-size and assertion-safety checks, and the complete runtime/UI artifact build passed. Full type checks and broad suites remain hosted. The tested source tree matches published head `84dffef7a099a9918fd08ae63993dfef4df788d9`; signing changed no runtime input. The retained build keeps its original pre-signing stamp.

The source-disappearance test is a deterministic filesystem-spy primitive, not a public concurrent-process claim. No host-power-loss or elapsed lease-expiry behavior is claimed. Complete private proof and failure history are retained; publication omits account and infrastructure details.

Thanks @Alix-007 for the original repair and regression coverage.


obviyus:
Thanks @Alix-007!

**Proof:** The built public `openclaw doctor --fix --non-interactive` command exercised real failed quarantine for both container and browser registries in isolated offline state.

- **Bug:** A real `EISDIR` rename failure deleted the malformed source and produced a false successful-quarantine claim.
- **Fix:** The existing owner preserves source bytes, reports the rename failure, and reports success only after the bytes actually move.

| Before | After |
| --- | --- |
| Failed rename removed the input and claimed quarantine succeeded. | Failed rename preserves the input and occupied destination, with a visible repair warning. |
| The claimed destination had not received the input. | Resolving the collision permits one correct quarantine; a repeat leaves it unchanged. |

All 34 focused tests passed; four regressions failed for the intended reasons on baseline. Independent native acceptance verified both failures and recoveries, read-only inspection, real SQLite conflict preservation, shard precedence and mixed malformed/valid migration, plus empty/missing inputs. The original stale-clock fixture refusals and dependent-step mistake remain recorded; the corrected fixture changed no product guard, lease or deadline. Source disappearance remains a separately labeled primitive test. Contributor credit is preserved.

Final review found no actionable findings and accepted the real SQLite compatibility proof. [CI attempt 2 passed](https://github.com/openclaw/openclaw/actions/runs/34601591318/attempts/2). Attempt 1 had a 120-second tooling-help test timeout; its logs remain retained. The debug rerun passed, but the original timeout mechanism is not established. No CI exception was used.

PR #144799, opened 2026-09-11
fix(secrets): screenshot and video uploads fail through protected egress with Bad Content-Length

### LINKED ISSUE #144662: [Bug]: screenshot and video uploads fail through protected egress with Bad Content-Length
# Summary

Screenshot and video attachment uploads through OpenClaw protected egress are rejected by GitHub with HTTP400 and `Bad Content-Length`, despite a correct client-side byte length and verified repository access.

Related: #141426. That report concerns a different provider and HTTP401; this report does not establish the same cause for it.

## Bug type

Behavior bug (incorrect output/state without crash).

## Beta release blocker

No.

## Steps to reproduce

1. Run an HTTP uploader in Gateway-hosted exec with protected egress enabled, retaining the supplied proxy environment and TLS verification.
2. Verify the managed GitHub account and push access to the intended repository using the same execution context.
3. POST a valid PNG or H.264 MP4 to `https://uploads.github.com/user-attachments/assets`, supplying `name`, `content_type`, and the authorized `repository_id` query parameters. Use `application/octet-stream`, the existing managed authentication, and an explicit `Content-Length` equal to the exact file byte count. Do not print credentials or disable the proxy.
4. Observe the response. Two tested files,2236-byte PNG and11067-byte MP4, both received HTTP400 with `Bad Content-Length`; neither returned an asset URL.

## Expected behavior

GitHub documents local image/video attachments in its [CLI attachment guide](https://docs.github.com/en/github-cli/github-cli/attaching-files-with-github-cli). Known-length uploads should reach a length-requiring destination with an accurate final byte length after any proxy substitution. A controlled local HTTPS origin accepts the same fixed-length bytes directly and rejects the baseline proxied framing. That local control is not a claim that a previous OpenClaw release successfully uploaded these files to GitHub.

## Actual behavior

Actual GitHub responses captured on September11,2026 at02:58UTC:

| Input | Bytes | Client Content-Length | GitHub status | Message |
|---|---:|---:|---:|---|
| PNG |2236|2236|400|Bad Content-Length|
| H.264 MP4 |11067|11067|400|Bad Content-Length|

The same-process account and target repository permission checks succeeded. A separate secretless loopback test using the actual proxy, CONNECT and verified TLS showed an8192-byte body arriving byte-identically but without Content-Length and with chunked transfer encoding. Its fixed-length-checking origin returned fixture-defined411; that411 is not a GitHub response.

## OpenClaw version

2026.9.3, build9b12fd3 in the live before test. Public source baseline used for the local reproduction: `97301530b3174fb9f62cda3418854770cb4e9ad0`.

## Operating system

Linux x86_64; Node.js26.7.0.

## Install method

Managed built Gateway deployment.

## Model

Not applicable: the reproduction is an HTTP upload, with no model invocation.

## Provider / routing chain

Gateway-hosted HTTP client → authenticated protected-egress proxy → GitHub user-attachment API.

## Logs

```json
{"file":"upload-fixture.png","bytes":2236,"status":400,"message":"Bad Content-Length"}
{"file":"upload-fixture.mp4","bytes":11067,"status":400,"message":"Bad Content-Length"}
```

The inputs are neutral synthetic media, not private UI captures. Both fully decode; the PNG and every frame of the one-second MP4 were inspected. Input SHA-256:

```text
png d747014a23a4aa3162a1a7ffa15e46f7454de0a37a9a5820ab13f5a18208e321
mp4 cbcbb557d277a5b637ce01b0ebc287600f3c9513c751c1f055bf6f07775a1404
```

Source inspection confirms that the baseline removes both framing headers and streams the transformed body. The reported behavior is fixed-length-origin interoperability, not a claim that chunked transfer encoding is invalid HTTP. No successful real GitHub after-upload or production deployment is claimed.

## Impact and severity

Affected: the tested protected-egress attachment workflow. Both observed GitHub attempts failed, preventing the proof image and video from being attached. The scope beyond this workflow is not established.

## Additional information and owner review

Last known good version: `NOT_ENOUGH_INFO`.

@openclaw/openclaw-secops: the affected owner is `src/secrets/egress-proxy/`, covered by security CODEOWNERS. Please take ownership of, explicitly authorize, or review the narrowly scoped fixed-length forwarding correction before it proceeds further. The intended boundaries are bounded in-memory preparation, current authorization checks, unchanged secret scanning, and retained streaming for above-bound/unknown-length bodies. This is a request for the owner involvement required by CONTRIBUTING, not a request to weaken a guard.

zhangguiping-xydt:
Closes #144662

## What Problem This Solves

Fixes an issue where users uploading a screenshot or a video from a Gateway-hosted agent run would have the upload rejected with `400 Bad Content-Length` whenever the request travelled through the secret egress proxy. The same upload succeeded when protected egress was not in the path, so the failure looked like a broken attachment rather than a proxy behaviour.

The reported surface is media upload, but the defect is not media-specific: any request that declares its body length and reaches an origin that requires a final length fails the same way, because the proxy re-framed every forwarded body as chunked.

## Why This Change Was Made

The proxy drops the client's `Content-Length` because sentinel substitution can resize a body, so the declared length cannot be trusted after substitution. Dropping it without a replacement left the upstream HTTP client with no length at all, and it fell back to `Transfer-Encoding: chunked`. An origin that requires a final length then rejects the request.

The forwarding owner now prepares a body that the client declared at or below 1 MiB, substitutes it with the same carry-window transform the streaming path uses, and forwards it with the substituted byte length as `Content-Length`. The client's declared length is never reused, because substitution can change the byte count. Bodies above the bound and bodies with no declared length keep streaming with chunked framing, so large uploads are unaffected. Preparation also finishes before the upstream request is opened, so a body that fails substitution cannot emit a partial upload.

## User Impact

Uploads that declare a length now succeed through protected egress against origins that require a final length, including GitHub's attachment endpoint, and the failure above disappears without the user disabling protected egress or re-encoding the upload.

Operators get the same secret substitution, destination binding, run revocation, and fail-closed refusals as before. The only change is the framing of the upstream request, and the documented framing behaviour is updated to match.

## Evidence

Focused tests (`src/secrets/egress-proxy/proxy-server.test.ts`) cover a declared-length body with and without a sentinel, a declared-length body whose sentinel cannot be resolved, and a body above the prepared bound. All three egress-proxy suites pass together:

    node scripts/run-vitest.mjs run --config vitest.config.ts \
      src/secrets/egress-proxy/proxy-server.test.ts \
      src/secrets/egress-proxy/proxy-server.lifecycle.test.ts \
      src/secrets/egress-proxy/proxy-server.websocket.test.ts

     Test Files  3 passed (3)
          Tests  59 passed (59)

A recorded live run drives the real proxy over real TCP, a real CONNECT tunnel and real TLS interception against a real HTTPS origin that rejects uploads arriving without a final length, the way the reported endpoint does. The same proof, run against the unmodified forwarding owner first:

    [FAIL] strict origin accepted the upload :: status=400 body="Bad Content-Length"
    [FAIL] origin received a Content-Length instead of chunked framing :: content-length=undefined transfer-encoding=chunked

and against the change:

    [PASS] strict origin accepted the upload :: status=200 body="ok"
    [PASS] origin received a Content-Length instead of chunked framing :: content-length=1849 transfer-encoding=undefined
    [PASS] declared Content-Length equals the substituted byte length :: sent=1849 expected=1849
    [PASS] the client's stale declared length was not reused :: declared=1922 sent=1849
    [PASS] large upload still uses chunked framing :: transfer-encoding=chunked content-length=undefined
    [PASS] no byte reached the origin for the refused body :: origin_requests=0

The stale-length assertion is the one that matters most: the client declared 1922 bytes, the substituted body is 1849 bytes, and the origin was told 1849. Re-emitting the client's declaration would have produced a truncated or rejected upload.

The repository changed-file gate (`node scripts/check-changed.mjs`) passes over the touched files, including typecheck, scoped lint, the dead-export scan, and the import and boundary guards.


VACInc:
Superseded by merged PR #144902, the canonical fix for #144662. It implements measured-length forwarding in the existing protected-egress owner, using a 100 MiB preparation limit instead of this proposal's 1 MiB limit, with shared capacity/deadline controls and current authorization checks. Larger and unknown-length bodies retain streaming. This is the replacement for this alternative implementation, not a claim of deployment or successful GitHub attachment readback.

PR #144810, opened 2026-09-11
refactor(auth): await Copilot cache persistence

steipete:
Related: #144592

## What Problem This Solves

Copilot token resolution still reads and writes its SQLite cache through synchronous persistence contracts. It needs to retain the cache operation before returning a token when persistence becomes asynchronous.

## Why This Change Was Made

Use the existing asynchronous core keyed store and await both cache lookup and save. Preserve the shipped explicit cache-file and injected JSON callback options, credential scoping, expiry checks, and token exchange behavior.

## User Impact

Authentication behavior and stored data stay compatible. A cache save finishes before token resolution returns. This is API groundwork; moving native SQLite execution into workers remains tracked by the umbrella issue.

## Evidence

- 30 focused tests and the full selected changed checks passed.
- Fresh independent Codex review through P2 passed. The rebase onto the landed core-store prerequisite is patch-identical.
- Synthetic loopback token exchange with real SQLite persistence and a second process verified cache reuse: the first exchange took 146.46 ms with one HTTP request; reopening took 5.02 ms with zero HTTP requests. These measure different paths and are not a speedup comparison.
- Existing synchronous injected JSON callbacks remain supported. No real provider credentials were used.

### DIFF
--- a/src/plugin-sdk/provider-auth-copilot-cache.test.ts
+++ b/src/plugin-sdk/provider-auth-copilot-cache.test.ts
@@ -1,0 +1,106 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { COPILOT_INTEGRATION_ID } from "../agents/copilot-dynamic-headers.js";
+import * as pluginState from "../plugin-state/plugin-state-store.js";
+import { createDeferredCore } from "../shared/deferred.js";
+import { withStateDirEnv } from "../test-helpers/state-dir-env.js";
+import {
+  fingerprintCopilotSourceCredential,
+  resolveCopilotTokenCache,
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2a2df53ec28de23bcb14a601743db9f8f67c4c76](https://github.com/openclaw/openclaw/pull/144810/commits/2a2df53ec28de23bcb14a601743db9f8f67c4c76)
- Landed commit: [1d4afe39664428308a2c71a64c94dea7b607bc0a](https://github.com/openclaw/openclaw/commit/1d4afe39664428308a2c71a64c94dea7b607bc0a)

PR #138673, opened 2026-09-04
feat(update): repair installations with configured inference

steipete:
## What Problem This Solves

Installation repair needs a bounded execution path that uses the operator's configured inference and checks whether the target actually improved. Previously, embedded `triage --run` ran an agent turn without a tool-call budget or a post-repair validation result.

Related: #124396 (transactional update validation and recovery) and #137580 (extended Gateway downtime during update finalization). This is the repair-loop slice of that work; unattended updater integration follows separately.

## Why This Change Was Made

Share one repair loop with caller-owned validation, cancellation, time limits, tool-call limits, progress events, and structured outcomes. It tries the system-agent owner's primary and configured fallback models before other agents' authenticated routes. A fresh isolated native turn runs against the captured installation; host-side admission counts tool calls after approval checks and drains owned processes before removing temporary state. The validation oracle determines success and stops the loop on regression or lack of improvement.

The recut also adapts the moved inference probe and current auth-preference API, preserves primary-model precedence when a CLI runtime alias differs from the model provider, preserves configured auth rotation, and redacts complete summaries before truncating them. Removed the no-op unattended wrapper and duplicate wrapper test, narrowed exports to current consumers, and shared repeated tool-policy construction. Activation, service lifecycle, snapshots, and rollback remain with the caller.

## User Impact

`openclaw triage --run` still requires an interactive terminal and now allows one repair turn, with a ten-minute overall limit, five-minute turn limit, and 40 tool calls. Doctor lint runs before and after repair; a clean initial check returns without inference. Incomplete repairs return nonzero even if the model claims success. External coding-agent handoffs retain their existing behavior.

The operator-owned update or explicit `triage --run` request permits prompt-free host exec in a temporary configuration. Explicit exec or repair-tool denies return `unavailable` with `exec-denied-by-policy` before the repair agent runs; the CLI explains the policy and points to `openclaw triage` for an external handoff. The workspace is pinned to the installation or staged candidate root, filesystem tools enforce `workspaceOnly: true`, and safe-bin, allow/deny, and provider policies are preserved while capabilities are capped to `exec`, `process`, `read`, `write`, `edit`, and `apply_patch`. Every fallback model must pass the same repair-tool gate before execution. Configured sandbox and non-gateway execution remain refused.

[...]

PR #144314, opened 2026-09-10
fix(update): preserve original authority during recovery admission

vincentkoc:
Related: https://github.com/openclaw/openclaw/issues/143752

## Problem

Interrupted package recovery must reacquire its recorded original authority. It must refuse an authority store that is missing, replaced, linked, or still owned by a live process instead of recreating that store or bypassing its owner.

## Changes

- Preserve read-only original-domain admission and existing SQLite ownership checks, including the retained Bun snapshot-lock handling.
- Capture executor authority before update work, preserve pending-recovery refusal before writable preparation, and retain strict retired-record filtering.
- Reconcile with landed #144313 and main's service-environment owner.
- Retain the reviewed journal BLOB byte bounds and service-plan extraction. Preserve complete test-shard ownership without raising limits.

Vincent Koc's original authority-admission head is retained as the first merge parent. Jason Sy's integration fixes are source-reviewed separately. Layer 2 of eight; #144313 is merged at `7905e684051c206847d649093e077455a87549c0`. Later layers remain separate.

## Evidence

- Signed published head: `d085b8f78bfeccd5bc47208445842a47e749f42b`.
- Exact candidate tree: `df3afed5ccd17b2c77ed0ad4e265dd876241a772` on landed main `7905e684051c206847d649093e077455a87549c0`.
- Completed independent changed-base source review found no actionable P0/P1/P2 issue. Nineteen of twenty layer files are byte-identical to the previously reviewed repair; the additional mainline shard exclusion retains coverage.
- Owner execution on this exact candidate: **21 integration tests passed**, covering pending admission and strict retired-record behavior; **30-stage changed-code gate passed**, including types, export scans, lint, storage and runtime guards.
- Prior exact-head and repair results remain separately source-bound. They do not qualify this candidate by transfer. The independent reviewers ran no tests.
- Successful hosted CI on the exact published head remains required before merge. Draft-skipped runs and predecessor checks do not count.

## Hosted component execution

[...]

PR #144819, opened 2026-09-11
fix(update): report configuration failures during repair

RomneyDa:
Small follow-ups: #147547 separates repair turns; #147584 adds the delegated worker receiver; #147608 isolates local route setup; #147645 keeps delegated repair turns under updater ownership.

## What Problem This Solves

Update repair incorrectly tells users to check model setup when loading the copied configuration fails.

## User Impact

Repair reports the actual configuration or database error. Genuine model-selection failures keep their existing guidance.

## Why This Change Was Made

The blanket catch arrived with the original repair loop in [#138673](https://github.com/openclaw/openclaw/pull/138673), which introduced structured repair outcomes; no catch-specific rationale was recorded. The inner inference selector still handles model/auth failures. Config errors now reach the existing outer handler, which redacts and bounds the reason and returns `aborted`. Both `aborted` and `unavailable` prevent activation; cancellation and environment cleanup remain intact. Manual triage still fails, but shows the config error instead of suggesting onboarding. The migrated-runtime fix is already on main via #145044.

## Evidence

The future-schema regression fails on main and passes with this fix. All 27 focused runtime, repair-loop, and inference tests passed again during the catch-provenance audit on `d86aeec3c572`; independent P0–P2 review is clean. Changed checks pass. The updater patch is unchanged after rebasing onto main, which now includes the independently merged Reef clock-fixture fix #147551; exact-head CI passes.


steipete:
Maintainer review at head 7d31253 (source review plus hosted CI; nothing executed locally). Running the staged repair in the updated runtime is the right selection, but moving it into its own process has to carry the existing executor's child custody as well as the live run/requester checks. The new launch in `update-repair-worker.ts:69-83` bypasses `withUpdateCommandExecutorChild`, so the worker never acquires a delegated executor: if the updater exits while the repair is still draining, the installation lease owner has no registered repair child to consult before admitting another updater (that registration is exactly what `update-command-executor.ts` records to prevent reclamation while children drain, see #144314 and #144778). Please wire the process through the existing child-delegation owner, retain original and staged-root custody through settlement, and prove that parent death leaves acquisition blocked until the repair child exits. Also verify that revoked executor authority prevents the final repair tool effect; the new test only makes the ledger terminal. ClawSweeper's exact-head review is blocked on the related maintainer-acceptance question for refusing older staged workers.

RomneyDa:
Addressed the executor-custody feedback. Each inference turn now uses `withUpdateCommandExecutorChild`; the worker runs under the delegated executor and the parent retains validation and service orchestration. Both roots stay reserved until the child exits and the canonical owner verifies cleanup. Final tool effects recheck the live executor as well as the original requester/update record.

The original failure was reproduced by killing the updater while its repair child remained alive. A separate built regression reproduced an actual tool effect after executor revocation while the ledger remained running. Both are fixed. Final proof passed 33 focused tests, all four managed-service repair cases, and 13 built isolation E2Es, including requester cancellation and immediate refusal of malformed delegated turns. Full build/changed checks and final runtime rebuild/types/lint passed; fresh review completed with its sole claim disproved by the explicit exit handler and built exit-1 tests.

Rebased onto main and removed overlapping staged-runtime implementation already present there. The PR now describes the remaining custody repair and preserves released updater compatibility at the worker receiver.

@clawsweeper re-review

RomneyDa:
The completed exact-head re-review confirms the custody feedback is addressed and reports no remaining code or security findings.

The published-updater proof is now available: [Docker job](https://github.com/openclaw/openclaw/actions/runs/34677339716/job/103509546945) installed `openclaw@2026.9.3` and upgraded seeded operator state to the package built from CI merge `5fdb14d60064`, containing head `bbd5074500e4`. Package integrity and `legacy-operator-state` with automatic authenticated restart passed. Worker custody and revocation are covered separately by the built failure-boundary tests.

The older-target behavior is intentional in both repair phases. An older worker cannot recheck the live executor before a tool effect or yield custody for parent validation, so this updater stops that automatic repair attempt with triage guidance, including after activation or partial rollback. This implements the requested executor guarantee without weakening it. Successful updates and the existing rollback decision are unchanged; the released v2026.9.4 complete-loop receiver remains supported for older updaters installing this version.

No SQLite schema or persistent-store semantics changed. Native Windows descendant-cleanup proof remains unrun: general Windows CI passed but skipped the four real repair-effect cases and did not select the worker E2Es. The PR body records that limit rather than treating those general checks as repair-specific evidence.

fuller-stack-dev:
Maintainer qualification at `d86aeec3c5725d59bd8a979335d899930f765745` is complete. No contributor code changes were needed.

- All 27 existing runtime, repair-loop, and inference tests pass. Scoped type-aware lint and formatting pass. Fresh independent P0–P2 review is clean.
- A separate real-config loop fixture reproduced the bug on pristine main `dd338334572`: a newer-schema SQLite refusal became “Check model setup.” With this head, the same fixture returns `aborted`, zero attempts, and the actual schema reason through the existing bounded/redacted handler. The disposable database version and environment are preserved.
- The same fixture keeps an actual unresolvable model `unavailable` with the existing model guidance, and preserves explicit caller cancellation. Runtime module setup occurs before the fixture’s unchanged 10-second diagnostic budget. No production timeout or schema changed.
- The change removes eight net production lines. `selectUpdateRepairInference` still owns model/auth classification, and `withUpdateRepairEnvironment` still restores its scope in `finally`. Manual triage retains failure exit behavior and now prints the config reason.

Focused command:
```sh
node scripts/run-vitest.mjs src/infra/update-repair-agent.runtime.test.ts src/infra/update-repair-agent.test.ts src/infra/update-repair-inference.test.ts --maxWorkers=1
```

The prior exact-head CI is green but older than the native 24-hour proof window. One unchanged-head refresh is running: https://github.com/openclaw/openclaw/actions/runs/34922800622. Separate delegated-worker changes and their published-driver/Windows proof are not part of this two-file diagnostic fix. RomneyDa’s implementation and attribution are preserved.

fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [ccd812c86fc06498d8bb30e5d91d40a7917c4886](https://github.com/openclaw/openclaw/pull/144819/commits/ccd812c86fc06498d8bb30e5d91d40a7917c4886)
- Landed commit: [060a3eddac975b60f344f9c63eb51cc0d01f9753](https://github.com/openclaw/openclaw/commit/060a3eddac975b60f344f9c63eb51cc0d01f9753)

PR #144820, opened 2026-09-11
refactor(web-fetch): remove private runtime cache reset

steipete:
## What Problem This Solves

The web-fetch provider cache exposes a private reset used only by its own tests, even though those tests already create independent config keys.

## Why This Change Was Made

Remove that reset, make the existing WeakMap binding const, and remove the test-only reset calls. All case bodies and runtime cache invalidation rules remain unchanged. Production and test code each shrink by four lines (+1/-5 each).

The reset shipped internally in v2026.8.2; this changes the internal module export set. Contract inspection found no supported package/SDK route or production caller for it. The change preserves the legacy alias source and the three production exports: isWebFetchProviderConfigured, listWebFetchProviders, and resolveWebFetchDefinition. Unsupported external dist imports were not exhaustively audited.

## User Impact

No intended change to provider selection, cache behavior, config activation, or supported public APIs. The current config-writer authority checks remain untouched.

## Evidence

All 41 cases passed before and after with identical inventories and statuses: 19 owner-cache cases, ten cold-import contracts, and 12 public-tool fallback cases. All 29 selected changed checks passed, and independent review found no actionable P0-P2 issue. The fallback suite mocks provider-definition resolution and includes an existing local loopback cleanup check; it is not external-provider proof.

The normal full build passed all 16 phases, including declarations, SDK export checks, runtime postbuild, and UI budgets. A credential-free process with an isolated HOME imported both built runtime paths: all three executable exports remain functions with identical alias bindings, and the private reset is absent. No provider was invoked. Source and dependency fingerprints stayed unchanged through proof.

### DIFF
--- a/src/web-fetch/runtime.test.ts
+++ b/src/web-fetch/runtime.test.ts
@@ -103,17 +103,14 @@ function requireResolvedWebFetch(
 
 describe("web fetch runtime", () => {
   let resolveWebFetchDefinition: typeof import("./runtime.js").resolveWebFetchDefinition;
-  let clearWebFetchRuntimeCachesForTest: typeof import("./runtime.js").clearWebFetchRuntimeCachesForTest;
   let clearSecretsRuntimeSnapshot: typeof import("../secrets/runtime.js").clearSecretsRuntimeSnapshot;
 
   beforeAll(async () => {
-    ({ clearWebFetchRuntimeCachesForTest, resolveWebFetchDefinition } =
-      await import("./runtime.js"));
+    ({ resolveWebFetchDefinition } = await import("./runtime.js"));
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [03c6604bf3610dd9e19baa74199510d9523d2fa6](https://github.com/openclaw/openclaw/pull/144820/commits/03c6604bf3610dd9e19baa74199510d9523d2fa6)
- Landed commit: [441435b5066cfe4d5fc7c05ddd697119cc12f4fd](https://github.com/openclaw/openclaw/commit/441435b5066cfe4d5fc7c05ddd697119cc12f4fd)

PR #144824, opened 2026-09-11
refactor(memory): await journal operations and allocate sequences atomically

steipete:
Related: #144592

## What Problem This Solves

Memory-host event operations still expose synchronous storage contracts. Introducing awaited append admission also exposes a race if a cursor is evicted after an earlier tail read: a later append can reuse a retained sequence.

## Why This Change Was Made

Await journal registration and reads throughout the memory-host API. Move retained-tail selection into the existing append transaction so sequence allocation considers both the cursor and authoritative retained events at commit. Carry the journal key prefix and range as data, preserving the existing fixed-width storage keys.

## User Impact

Event ordering and export formats remain compatible. Appends cannot overwrite a retained event because a cursor disappears during an asynchronous handoff. Existing schema, retention limits and synchronous native transaction callbacks are preserved.

## Evidence

- 65 focused tests and full selected changed checks passed.
- Fresh independent Codex review through P2 passed with full append-owner and caller context.
- Fingerprinted real-SQLite smoke appended 32 records, removed the derived cursor, reopened in a fresh process and allocated/exported sequence 33 successfully.
- Delayed append coverage exercises the cursor/tail interleaving, and export coverage preserves ordering and limits.
- This prepares the complete journal operation for worker execution; native SQLite still runs on the calling thread in this prerequisite.

### DIFF
--- a/src/memory-host-sdk/event-store.ts
+++ b/src/memory-host-sdk/event-store.ts
@@ -3,7 +3,7 @@ import { z } from "zod";
 import { resolveWorkspaceStateIdentity } from "../agents/workspace-state-identity.js";
 import {
   pluginStateEntriesInKeyRange,
-  registerPluginStateSyncSequencedJournalEntry,
+  registerPluginStateSequencedJournalEntry,
 } from "../plugin-state/plugin-state-store.js";
 import type { MemoryHostEventRecord } from "./event-types.js";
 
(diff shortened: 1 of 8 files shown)


steipete:
Compatibility confirmation: the existing cursor/event namespaces, JSON records, 16-digit sequence suffix, retention limits and SQLite schema stay unchanged. Allocation now rereads the retained tail inside the existing write transaction, preserving the invariant that eviction of a derived cursor cannot overwrite a retained event.

The 65-case native suite and real-SQLite append/reopen probe preserve existing data and export ordering: after 32 appends and cursor removal, fresh reopening allocates sequence 33 and exports all 33 records. Delayed tests cover an intervening append plus cursor removal before admission. No data migration is needed for this unchanged storage format.

steipete:
Merged via squash.

- Prepared head SHA: [453ee7c86d59ea37600ab51744070beff0fc2c15](https://github.com/openclaw/openclaw/pull/144824/commits/453ee7c86d59ea37600ab51744070beff0fc2c15)
- Landed commit: [3ae14efdb0a26f400180d7343ee73bce70bfbf18](https://github.com/openclaw/openclaw/commit/3ae14efdb0a26f400180d7343ee73bce70bfbf18)

PR #144829, opened 2026-09-11
fix(ui): show Error instead of custom on failure notices

RomneyDa:
## What Problem This Solves

Failed-turn notices in the Control UI currently show the internal sender label `custom`, which is confusing and leaks an implementation detail into the transcript.

## Why This Change Was Made

The shared sender-label renderer now uses the recorded notice subtype:

- failed-before-reply and cloud-workspace recovery failures display **Error**
- other custom notices display **System**
- structured workspace conflicts retain **Cloud workspace**
- mixed informational/error groups remain **System** so informational messages are not mislabeled

This is presentation-only. It does not change transcript roles, persistence, grouping, or the runtime failure itself.

## User Impact

Failure footers read **Error** followed by the timestamp, including after history reload. This fixes the confusing label shown alongside the stale-install failure; it does not prevent that failure.

## Evidence

- `node scripts/run-vitest.mjs ui/src/pages/chat/components/chat-message.test.ts`: 236 passed
- focused Chromium history/reload coverage: 2 passed
- `node scripts/check-changed.mjs`: passed on the rebased patch
- rebased onto current main and removed an unrelated historical CI-cleanup commit from this branch

AI-assisted with Codex.

### DIFF
--- a/ui/src/e2e/cloud-workspace-conflict.e2e.test.ts
+++ b/ui/src/e2e/cloud-workspace-conflict.e2e.test.ts
@@ -270,7 +270,9 @@ suite.define(() => {
           await page.reload();
           await notice.waitFor({ timeout: 10_000 });
           expect(await notice.count()).toBe(1);
+          expect(await notice.locator(".chat-sender-name").textContent()).toBe("Error");
           await gateway.waitForRequest("chat.startup");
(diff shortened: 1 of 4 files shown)


RomneyDa:
@clawsweeper re-review

PR #144832, opened 2026-09-11
refactor(memory): await standing intent operations

steipete:
Related: #144592

## What Problem This Solves

Standing-intent tools and hooks assume database work has completed when their synchronous store calls return. They need to retain the operation and propagate its result when database admission becomes asynchronous.

## Why This Change Was Made

Await create, list, sweep, cancel and match through the existing asynchronous agent-database owner, including actual tool and registered hook callers. Keep native schema, lifecycle, FTS selection and claim transactions synchronous and intact; pure scope and context helpers remain synchronous.

## User Impact

Tool results and prompt injection wait for the corresponding store operation. Rejections reach the existing error policy, and an operation retains its originally selected database across ambient path changes. Durable formats, eligibility and firing budgets are unchanged.

## Evidence

- 53 focused cases passed using real isolated SQLite, registered hooks, tool calls and fresh reopen checks.
- Twelve delayed/rejected admission regressions fail against original source for the intended missing-await boundaries, then pass on the candidate.
- Full selected checks, SDK export/type checks and fresh independent Codex P2 review passed.
- No full build was run; this exports an existing owner from an already-imported private-local SDK module.
- This is API groundwork. Native SQLite still executes on the caller thread until the shared worker cutover.


steipete:
Compatibility confirmation: this changes awaited admission and callers, with no changes to the SQLite schema, stored scope encoding, serialization, eligibility predicates, firing limits or native claim transaction. The existing agent-database owner captures the database identity and retains it through settlement.

The 53-case suite uses real isolated SQLite and covers registered tools/hooks, concurrent firing claims and fresh reopen. Twelve delayed/rejected admission cases fail against the prior source at the missing-await boundary and pass after conversion. No data migration is required. The review narrative also confirms the original generic migration concern is unsupported by the actual diff.

steipete:
Merged via squash.

- Prepared head SHA: [05eff8441d67b985ab963191b4438fd3b13de356](https://github.com/openclaw/openclaw/pull/144832/commits/05eff8441d67b985ab963191b4438fd3b13de356)
- Landed commit: [4ca592809742472860992fac6c2202c9e87f5db8](https://github.com/openclaw/openclaw/commit/4ca592809742472860992fac6c2202c9e87f5db8)

PR #144838, opened 2026-09-11
fix: preserve child signal termination during CLI respawn

### LINKED ISSUE #144200: [Bug]: Source launcher converts child signals into normal exit codes

shakkernerd:
Fixes #144200

## What Problem This Solves

Fixes an issue where supervisors receive an ordinary exit code after a respawned OpenClaw runtime is terminated by a Unix signal. The lost signal can send the development watcher into restart or Doctor recovery instead of stopping.

## Why This Change Was Made

Both startup respawn runners now terminate with the child's actual Unix signal after their existing shutdown cleanup. Explicit numeric exits and acknowledged stops stay numeric, and Windows keeps its current handling.

## User Impact

Supervisors can distinguish an interrupted runtime from a command that explicitly returns an exit code, including 143. Forced shutdown preserves the actual SIGKILL outcome.

## Evidence

- Linux/Node 24.19.0 on Blacksmith Testbox: **323 tests passed, 1 skipped across 10 affected files**. All **38 real-process cases** passed across source, packaged, built, and warning respawn routes.
- Killing only the respawned runtime produced `(code: null, signal: "SIGKILL")` at the outer watcher command. Its invocation log contained only `gateway`, with no Doctor pass or restart.
- Restoring the pre-fix launcher owner makes that regression fail: the watcher runs Doctor and restarts instead of preserving SIGKILL.
- Explicit exit 143, acknowledged shutdown returning 0, forwarded SIGINT/SIGTERM, forced SIGKILL, and mocked Linux/Windows termination controls passed.
- Changed-file formatting, syntax checks, targeted typed lint, and repository ratchets passed. Independent review reported no actionable P0–P2 findings.


shakkernerd:
Fixed: startup respawns preserve Unix child signals so supervisors stop interrupted runtimes without entering watcher recovery.

PR #144843, opened 2026-09-11
fix(plugins): stop relocated install records from shadowing bundled plugins

RomneyDa:
## What Problem This Solves

Fixes bundled plugins losing trusted access, and update Doctor failing, when a stored plugin path points into a previous OpenClaw installation.

## User Impact

The current bundled plugin is selected automatically. Existing plugin settings, channel settings, slot choices, install records, and previous files remain intact; no uninstall or re-enable step is needed.

Intentional external overrides, source links, explicit configuration, development sources, active bundled-directory overrides, and unknown package roots retain their existing behavior. No configuration or persistence migration is introduced.

## Why This Change Was Made

Recognize stale records only inside another verified OpenClaw compiled bundle. Require the exact inferred package root to identify itself as OpenClaw, so an unrelated compiled-looking directory cannot lose precedence. The diagnostic names the stale path and explains the safe fallback instead of recommending uninstall, which also removes configuration.

Original implementation and diagnosis by @RomneyDa (Dallin Romney). This follow-up preserves contributor ancestry and resolves the package-identity and destructive-recovery review findings.

## Evidence

- The two stale-record regression cases fail with the former warning and pass with the corrected warning, in both candidate orders, including configuration and file-preservation assertions.
- 19 targeted native tests passed against the integrated cache execution-frame changes. Prior 162-test evidence is retained without rerunning unchanged proof.
- Independent source-only API review found no actionable P0–P2 issues.
- Native changed checks passed, including production/test typechecking, all three zero-finding dead-export scans, changed-file lint, and runtime/storage/pairing guards.
- Maintained compiled package and real fresh npm installation passed; bundled Codex loaded as trusted and the fresh config was unchanged.
- The unmodified published 2026.9.4 updater successfully ran the real 2026.9.5 candidate migration rehearsal, Doctor lint, configuration validation, plugin resolution, and Gateway canary. Actual candidate Doctor child exits were recorded.
- A separate real installed candidate passed Doctor fix/lint and runtime inspection with nonempty plugin, channel-default, and slot settings. Those settings, the stale install record, and the previous compiled bundle remained unchanged.
- The existing main CI timing repair was integrated with its matching runtime cache prerequisite; the previously failing fixed-boundary test passed. Exact-head [hosted CI passed](https://github.com/openclaw/openclaw/actions/runs/35439485069) for 379b7eef58b39d171ab7529467493f342c288587. Native required-gate verification still applies before landing.

### Proof limitation

The published 2026.9.4 driver did **not** complete an end-to-end update: its immutable 30-second baseline fingerprint scan timed out after all candidate checks passed and before package activation. The newer source already contains the separate full-budget fingerprint fix. A same-artifact candidate-driver attempt correctly skipped as already current; it is not swap proof. Neither result is represented as a successful published-driver upgrade.

This PR's relocation acceptance is supported by the actual published-driver candidate checks plus installed-candidate configuration, record, file, and trust preservation. All proof used task-owned synthetic state and prefixes; no operator runtime was changed.


fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [379b7eef58b39d171ab7529467493f342c288587](https://github.com/openclaw/openclaw/pull/144843/commits/379b7eef58b39d171ab7529467493f342c288587)
- Landed commit: [5ec3614976b5badd2008c5a793a8767de984f389](https://github.com/openclaw/openclaw/commit/5ec3614976b5badd2008c5a793a8767de984f389)

PR #144852, opened 2026-09-11
fix(qa): isolate Matrix progress scenario events

RomneyDa:
## Summary

- exclude Matrix events already observed before the current tool-progress mention scenario
- apply the same scenario-local identity boundary to preview, final, and cleanup predicates
- retain current-scenario events even when sync replay duplicates an older event body

## Validation

- 7 focused Matrix progress tests
- `pnpm check:test-types`
- `pnpm lint`
- scoped `oxfmt --check`
- `git diff --check`

The Full Validation retry showed the mention-safety assertion consuming an event ID from the immediately preceding scenario even though the current progress and final events were present. This fixes the cursor identity boundary without weakening the mention-safety assertions.

### DIFF
--- a/extensions/qa-lab/src/live-transports/matrix/scenarios/scenario-runtime-event-scope.ts
+++ b/extensions/qa-lab/src/live-transports/matrix/scenarios/scenario-runtime-event-scope.ts
@@ -1,0 +1,11 @@
+import type { MatrixQaObservedEvent } from "../substrate/events.js";
+
+export function createCurrentScenarioEventPredicate(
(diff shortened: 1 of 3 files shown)


RomneyDa:
@clawsweeper re-review

The replacement head removes the test-only production export that the dependency/dead-code shard rejected. The event-scope predicate now lives in a small module consumed by both runtime and its regression. Seven focused tests, the exact deadcode:dependencies command, full test types, formatting, and diff checks pass.

PR #144863, opened 2026-09-11
fix(qa): stabilize Matrix voice preflight

RomneyDa:
## Summary

- run the frozen Matrix voice maturity scenario against deterministic mock OpenAI transcription
- configure the exact transcription model, prompt, plugin, and mention pattern required by the scenario
- add mock-server and scenario-contract regressions

## Validation

- 197 focused tests passed
- stable core and extension test type graphs passed
- repository lint passed
- exact changed-file formatting and git diff checks passed

## Release validation root

Full Validation 34540886759 and unchanged retry 34541072222 timed out matrix-voice-preflight-mention because the 7.33 catalog still pinned live-frontier audio. This backports the reviewed deterministic transcription contract from later release tooling without changing product behavior.

### DIFF
--- a/extensions/qa-lab/src/providers/mock-openai/server.test.ts
+++ b/extensions/qa-lab/src/providers/mock-openai/server.test.ts
@@ -4667,6 +4667,32 @@ describe("qa mock openai server", () => {
     });
   });
 
+  it("serves deterministic Matrix voice preflight transcription for the request prompt", async () => {
+    const server = await startQaMockOpenAiServer({
+      host: "127.0.0.1",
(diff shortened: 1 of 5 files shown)


RomneyDa:
@clawsweeper re-review

Addressed the matcher finding on replacement head 7df4c4f696a: both observed reply and expected marker now receive symmetric punctuation removal, and the scenario regression passes the exact hyphenated mock transcript echo through the real predicate. The two focused suites pass 197 tests; stable extension test types, repository lint, exact formatting, and diff checks pass.

Issue #141559, opened 2026-09-07
[Bug]: Unbindable-shape exec denial returns three conflicting error classes with no approval request and no remedy, so agents report false loss of exec access


cipp-ashe:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

A `system.run.prepare` command whose shape cannot be bound to an approval returns a single response carrying three different error classifications (`errorCode=UNAVAILABLE`, message prefix `INVALID_REQUEST:`, inner text `SYSTEM_RUN_DENIED:`), emits no approval request, and names no corrective action — so the agent concludes its exec access was revoked and reports that to end users, while the operator is actively approving other commands in the same session.

### Steps to reproduce

Observed on a lane where approval binding is active (any lane that is not `security: full` + `ask: off`).

1. In a session where the operator is present and answering exec approvals, run a single absolute-path command (e.g. `rg -c -F "it(" <abs-path>`). An approval card is presented, the operator answers, the command runs.
2. In the same session, run a compound command through the exec tool — a `cd <dir> && <command>` chain or a pipeline.
3. Observe the response: `errorCode=UNAVAILABLE errorMessage=INVALID_REQUEST: SYSTEM_RUN_DENIED: approval cannot safely bind this interpreter/runtime command`, returned in 5–8 ms.
4. Observe that no approval request accompanies step 3 — the 5–8 ms turnaround excludes a human decision, and the surrounding operator-answered approvals took 3.3–11.3 s (see Logs).
5. Reshape to a single absolute-path command and resend. An approval card appears again and is answered normally.

Counter-check on the same gateway and the same `node.invoke` transport: a `cd … && …` compound command issued from a `security: full` + `ask: off` lane returns `res ✓ node.invoke` and executes normally. On that policy `bindApproval` is false, so the binding step that produces this message is never reached — which places the trigger in the binding path rather than in the command text or the host.

Note on provenance: the gateway log records the denial responses but not the request payloads, so the exact denied command strings are taken from the affected agent's own transcript rather than from the gateway log.

### Expected behavior

The message should identify the command **shape** as the cause and name the corrective action the agent can take unilaterally (resend as a single command), consistent with the repo's own stated principle in root `AGENTS.md` → Product Judgment:

> Every action ends with a visible outcome or a recorded intentional non-outcome. Failure messages explain the next useful step.

and

> Prioritize broken existing behavior, especially silent failure.

A single response should also carry one coherent error classification rather than three.

[...]

PR #130972, opened 2026-08-27
fix(node-host): exec denials tell the agent the sanctioned escalation path

elharonyclaw:
## What Problem This Solves

Fixes an issue where users whose agent hits a `system.run` denial would see the agent burn turns retrying the identical denied request or improvising workarounds. The denial message ended at the denial — it named no sanctioned next step, so the model's next move was unguided.

## Why This Change Was Made

`sendSystemRunDenied` is the single choke point every `SYSTEM_RUN_DENIED` denial flows through before reaching the agent (`src/node-host/invoke-system-run.ts`). The model-facing message now ends with a hint naming the sanctioned next step — **scoped to what actually corrects each denial class**:

- **Policy-owned denials** (`security=deny`, `approval-required`, `allowlist-miss`, `execution-plan-miss`): the policy hint — "ask the operator to adjust the agent's exec policy; an identical retry will be denied again".
- **Approval-binding integrity failures** (cwd drift, script-operand drift/binding, path hardening, cwd capture — all fail closed with `approval-required` but owned by exact identity/content validation, not exec policy): the renewal hint — "the approved binding no longer matches this execution; request approval again for the current command". Policy-relaxation guidance here would send the model and the operator after the wrong lever.
- **Exemptions unchanged**: `approval-state-write-failed` (transient), the self-instructing `exec approval policy changed` denial, mac freeform host failures (raw); structured mac policy verdicts (`security=deny`, `allowlist-miss`) keep the policy hint.

The `exec.denied` telemetry event keeps its structured `reason` untouched; only the model-facing message changes.

## User Impact

Agents that hit exec denials stop retry-looping the same denied command and instead surface the blocker to the operator (or request approval again for binding drift) — saving turns and reducing noise. Operators see correct escalation guidance per denial class on locked-down exec policies.

## Evidence

- AI-assisted PR (Claude, agent-authored; happy to iterate on feedback).
- **Real behavior trace at head `613c5003`** — running Node through the real `handleSystemRunInvoke` dispatcher (no test doubles on the denial path; result sinks only, isolated `HOME`), verbatim output:

  ```text
  [trace A: local policy denial (security=deny)]
    SYSTEM_RUN_DISABLED: security=deny — to change this outcome, ask the operator to adjust the agent's exec policy; an identical retry will be denied again
  [trace B: mac freeform cancellation (non-policy, APPROVAL_CANCELLED)]
    approval prompt cancelled
  [trace C: mac structured policy verdict (reason: "security=deny")]
[...]

PR #144869, opened 2026-09-11
fix(exec): distinguish preparation refusals from user denials

Alix-007:
Related: #141559. Adjacent execution-stage guidance: #130972.

## What Problem This Solves

Fixes approval-preparation refusals that look like a human denied execution, even though the command could not be prepared and no approval request was created.

## User Impact

The error now distinguishes an unsupported command shape from an operator decision and gives recovery guidance specific to the execution path. Node callers are directed toward supported single-executable commands, without promising that nested interpreter scripts can run. Native CLI callers can retry supported direct executable/script forms with explicit paths.

Unsupported commands remain denied. Binding checks, approval decisions, revalidation, error codes, and public protocol fields are unchanged. This does not enable currently unsupported node interpreter execution or resolve the broader error-classification concerns in the linked issue.

## Why This Change Was Made

The binding owner now carries an internal failure classification instead of making callers identify that failure from diagnostic prose. The two existing preparation presentation sites explain the phase and their different capabilities. The internal classification is removed at the opaque harness boundary; it is not a new SDK or wire field.

## Evidence

Baseline `acbeb3a07197` and candidate `d83eff8ac3b5` were exercised through an isolated real Gateway:

| Entry and control | Observed result |
| --- | --- |
| `chat.send → exec(host=node) → paired headless node`, unsupported command | Baseline returned only the binding refusal. Candidate explains that no approval request was created; both create zero approval records and leave the workload marker absent. |
| Node shell chain rewritten as one supported absolute-path executable | An ordinary exec approval is created; `allow-once` executes it. An explicit `deny` does not. |
| Registered Anthropic CLI backend and permission callback, unsupported Perl lookup form | Candidate explains the preparation refusal without claiming a user denial; no approval or command execution occurs. |
| Same Perl runtime using an absolute script path on the native CLI path | An ordinary plugin approval is created; `allow-once` executes the script. Explicit denial and script changes after preparation remain rejected without execution. |

The native transport peer and loopback model responses were synthetic; Gateway routing, the registered plugin/callback, approval records, CLI resolution, binder, filesystem, and permitted local commands were real. No vendor authentication or paid model request is claimed.

A node absolute-script retry remains unsupported, so the contributor's blanket script-file advice is not included. Gateway and node state were task-owned, and OS sandboxing restricted network traffic to loopback.

After colocating the existing Perl option and node preparation helpers with their owners, candidate `1e5fee5d0ccf` repeated both registered preparation-refusal paths with the same diagnostics, zero approval records, and no execution. All 356 focused binding, node-host, native-approval, and Codex-bridge tests passed again, and the changed-source line-cap ratchet passed. Existing rejection tests now assert structured outcomes instead of pinning prose. Full static and matrix validation remain in hosted CI.

### DIFF
--- a/src/agents/cli-runner/cli-native-tool-approval.preparation.test.ts
+++ b/src/agents/cli-runner/cli-native-tool-approval.preparation.test.ts
@@ -1,0 +1,108 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { requestCliNativeToolApproval } from "./cli-native-tool-approval.js";
+
+const mocks = vi.hoisted(() => ({ prepare: vi.fn(), revalidate: vi.fn(), gateway: vi.fn() }));
+vi.mock("../../infra/system-run-approval-binding.js", () => ({
+  prepareSystemRunMutableFileBinding: mocks.prepare,
+  revalidateSystemRunMutableFileBinding: mocks.revalidate,
+}));
+vi.mock("../tools/gateway.js", () => ({ callGatewayTool: mocks.gateway }));
+
+const shapeError =
+  "SYSTEM_RUN_DENIED: approval cannot safely bind this interpreter/runtime command";
+const request = {
+  toolName: "Bash",
+  toolInput: { command: "printf approval-probe" },
+  pluginId: "claude-cli",
+  ask: "always" as const,
+};
+
+describe("native approval preparation guidance", () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mocks.prepare.mockResolvedValue({ ok: true, binding: { operands: [] } });
+    mocks.revalidate.mockResolvedValue({ ok: true });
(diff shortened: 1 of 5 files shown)


obviyus:
Merged as d88e10a752f395af49d95ad5d222eb3041dedece. Thanks @Alix-007 for identifying the misleading preparation failures. Node and native CLI errors now distinguish a failure before approval from a user's denial, with recovery advice limited to forms each path actually supports. Real Gateway/node and registered CLI-callback proofs verified zero approvals or execution on preparation refusal, ordinary approved retries, explicit denial, and native script-drift rejection. All 356 focused tests and the exact-head hosted CI passed; the final review found no blocking issues. Approval and binding policy are unchanged. The broader issue #141559 and adjacent execution-stage work #130972 remain open.

PR #144646, opened 2026-09-11
refactor(team-reports): await persistence and drain shutdown

steipete:
Related: #144592

## What Problem This Solves

Team Reports owns synchronous database operations across startup, scheduling, HTTP handlers and shutdown. Awaiting these operations requires shutdown to account for work admitted before it stops, including a run record that is still being created.

## Why This Change Was Made

Convert the store factory, queries, mutations and close operation to asynchronous contracts and migrate all callers. The scheduler reserves a run before awaiting persistence and drains admitted database work before closing the store. Cancellation interrupts external collection and model work; final database outcomes are allowed to settle. Startup publishes one scheduler/store owner only after initialization succeeds and closes an instance retired during startup.

Remove the duplicate store-closing owner. Prepared in-memory organization and people accessors stay synchronous. The SQLite schema, transaction callbacks and retention policy are unchanged.

## User Impact

Reports keep their existing API and scheduling behavior. Startup and shutdown consistently own pending persistence, preventing store closure underneath an admitted report run.

## Evidence

- 238 focused store, scheduler, lifecycle and HTTP tests passed, including delayed startup, cancellation and drainage.
- Full selected changed checks, the full build (348.8 seconds) and fresh independent Codex review passed.
- Standalone production-source smoke exercised synthetic collection, real SQLite persistence and fresh-process reopening, plus four real loopback HTTP routes returning 200.
- Thirty warm samples measured read median 0.102 ms and write median 0.216 ms; no comparative speedup is claimed.
- This proof uses isolated source runtime and synthetic inputs, not a deployed Gateway or external model service.

CI initially timed out in the unchanged managed-update refusal process test. The exact candidate case passed locally in 1.60 seconds with its original 60-second deadline and journal-immutability assertions; the same-head failed-job rerun is green. No Team Reports or test changes were needed.

PR #144871, opened 2026-09-11
feat(sqlite): run Team Reports storage in owned workers

steipete:
Related: #144592

## What Problem This Solves

Awaiting Node SQLite calls does not move their work off the application thread. Team Reports can still block timers and requests during database operations or lock contention even though its store methods return promises.

## Why This Change Was Made

Move Team Reports opening, schema work, prepared queries, complete transactions, WAL maintenance and close into a long-lived worker owned by a bounded shared broker. The plugin sends typed domain commands to a static private backend and locates it from the loader-selected runtime entry. Its existing service lifetime retains admitted work through shutdown.

The broker allows four lazy workers, 64 opening/live store clients, 128 requests and a bounded input queue. Physical aliases share an owner; admitted pathnames remain protected through client drainage. Queued cancellation removes unstarted work; dispatched operations retain their settlement. Worker loss or an undecodable committed result reports an uncertain outcome without automatically replaying writes. Local overload before dispatch does not retire a healthy shared worker.

This first consumer supports file-backed stores from the application main thread. Incognito, shared canonical databases and callers hosted in other application workers remain separate campaign work. Native synchronous SQLite primitives stay entirely inside the worker; no main-thread blocking RPC bridge is introduced.

## User Impact

Team Reports preserves its schema, report formats, UTF-8 limits, retention and HTTP behavior while SQLite work stops blocking its host event loop. Existing connection and prepared-statement ownership stays with the database worker.

Release context: publish this Team Reports change with the first host release containing the worker SDK and selected-runtime-entry capability. The regular release workflow must synchronize its plugin API compatibility floor; this PR does not publish or guess a future version.

## Evidence

- 296 focused tests passed on current main: 27 broker cases, 243 Team Reports cases and 26 test-routing checks. Full selected checks and both root-bundled and standalone builds passed.
- Fresh independent Codex review through P2 passed. The final 29-file remainder exactly matches that reviewed source after the two prerequisite PRs landed.
- Strict native instrumentation observed zero SQLite calls on the main thread for source execution and both plain-Node 24 compiled package layouts, including opening, HTTP reads/writes, drainage, close and fresh reopening. Compiled proofs imported no TypeScript/source fallback.
- A 751.386 ms external writer lock allowed 128 main-thread timer ticks; the largest gap was 6.665 ms and event-loop p99 was 6.480 ms. All 25 queued writes drained before close. Four actual HTTP routes returned 200, and report/Unicode data survived service restart and fresh-process reopening.
- Thirty warm source samples observed read median/p95 of 0.312/0.580 ms and write median/p95 of 0.408/0.610 ms. These are descriptive timings, not a paired throughput improvement claim.
- Regressions reproduce alias replacement, premature close, uncertain result decoding, invalid async backend execution, leaked client capacity and overload that previously disrupted unrelated writes.
- Proof uses synthetic reports and an isolated registered-service/HTTP harness; it does not claim a production Gateway deployment or completion of all SQLite migrations.

The exact CI wrapper failure is repaired by including the worker-path module in the canonical extracted dependency closure; all 70 wrapper regressions pass. The scheduler fixture waits for actual collector entry while retaining its immediate concurrent-run rejection and no-overlap assertions; all 31 scheduler cases and the original delayed-read reproduction pass. Full candidate checks and fresh P2 review passed for these two fixes.

### DIFF
--- a/docs/plugins/sdk-overview/infrastructure.md
+++ b/docs/plugins/sdk-overview/infrastructure.md
@@ -68,6 +68,63 @@ Admission retries use the connection's existing `busy_timeout`; this is not a
 total deadline for preparation or transaction execution. `options` supplies the
 same transaction diagnostics as `runSqliteImmediateTransactionSync`. Keep the
 database handle and its owning operation alive until the returned promise settles.
+The callback and SQLite calls still run synchronously on the caller's thread.
+
+### SQLite worker stores
+
+Use `openSqliteWorkerStore<Operations>` from
+`openclaw/plugin-sdk/sqlite-runtime` to move a feature's SQLite lifecycle off the
+application event loop. Call it from the main application thread with
+`{ moduleUrl, databasePath, input }`. `Operations` maps each domain operation to
+its `{ input, output }` types; `store.execute({ type, input }, { signal }?)`
+returns the corresponding output promise. The host currently rejects calls from
+other application workers, which need a shared host-broker connection.
+
+This first host supports filesystem-backed databases only. Empty paths, SQLite
+URIs, `:memory:`, and OpenClaw's reserved incognito database basename are refused
+before normalization or worker admission. In-memory and incognito ownership
+remain pending; these locators must never become disk filenames.
+
+The static local `moduleUrl` identifies a trusted feature module exporting
+`createSqliteWorkerBackend(input, { databasePath })`. Its factory owns database
+opening and schema setup; its synchronous `execute(command)` owns queries and
+transactions. `close()` may await cleanup outside transactions; the host retains
+the actor until that cleanup settles. Keep native handles, WAL maintenance, and
+prepared statements inside that backend. Send serializable domain commands and
+results across the boundary, never SQL strings, callbacks, or Kysely builders.
+Declare private build entries with
+[`openclaw.build.workerEntries`](/plugins/dependency-resolution#native-imports-from-a-standalone-source-build)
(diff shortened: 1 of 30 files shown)


steipete:
Compatibility confirmation: Team Reports uses the same SQLite schema, tables, report encodings, UTF-8 limits and retention policy. The existing SQL implementation moved into its private worker; the frontend preserves the awaited domain API introduced in #144646. This requires no stored-data migration.

The final source and both compiled package layouts write synthetic reports, close every store/worker, reopen the same database in a fresh process and verify exact values. The 296-case suite includes existing store behavior and worker failure/drainage controls; all checks and both builds pass. Native instrumentation records zero SQLite calls on the application main thread for those flows.

This plugin must publish with a host containing the new worker/runtime-entry capabilities via the existing release-time plugin API floor synchronization. The PR does not publish a plugin separately or guess a future host version. File-backed application-main-thread callers are this first scope; incognito and other application-worker callers remain explicitly outside the supported first slice.

steipete:
Merged via squash.

- Prepared head SHA: [57ec881ea053b5448045613038198aa90ce45f64](https://github.com/openclaw/openclaw/pull/144871/commits/57ec881ea053b5448045613038198aa90ce45f64)
- Landed commit: [979c0fc29e196000294a9ab5b5b5fc908a879d70](https://github.com/openclaw/openclaw/commit/979c0fc29e196000294a9ab5b5b5fc908a879d70)

PR #144880, opened 2026-09-11
fix(channels): honor pinned accounts in Mattermost model menus

obviyus:
## What Problem This Solves

Channel model menus could show learned choices from a shared account after the session pinned a different, expired account. Slash and thread-picker requests omitted session context.

## Why This Change Was Made

Both affected channel entry points read the latest session and pass it to the shared menu builder; thread interactions use the thread session. Other thinking-menu callers now preserve an explicitly empty catalog.

## User Impact

Model choices honor the session account pin. Existing native thinking controls remain available. No configuration or protocol change is required.

## Evidence

A real Gateway with a controlled channel transport exercised registered slash and thread callbacks. Healthy → expired → healthy → expired pins correctly removed and restored account-only choices after the fix; the baseline retained them. No inference was sent. Telegram Test Server /think retained all seven buttons with an empty provider list. Focused channel suites passed 13, 21, and 71 cases.

Related: #136257.

### DIFF
--- a/extensions/discord/src/monitor/native-command.options.ts
+++ b/extensions/discord/src/monitor/native-command.options.ts
@@ -144,7 +144,7 @@ export function buildDiscordCommandOptions(params: {
             provider: context?.provider,
             model: context?.model,
             agentRuntime: context?.agentRuntime,
-            ...(choiceCatalog?.length ? { catalog: choiceCatalog } : {}),
+            catalog: choiceCatalog,
           });
           const filtered = focusValue
             ? choices.filter((choice) =>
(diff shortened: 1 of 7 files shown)


obviyus:
Landed after clean exact-head review and [green CI](https://github.com/openclaw/openclaw/actions/runs/34593857128).

**Proof:** Compiled baseline and candidate Gateways registered real Mattermost slash and interaction routes against an external HTTP/WebSocket fixture. Actual posts created the parent and thread sessions; default authorization and signed callbacks remained active.

| Current session pin | Baseline, both menus | Candidate, both menus |
| --- | --- | --- |
| Healthy | account-only present | account-only present |
| Expired | account-only present | account-only absent |
| Healthy again | account-only present | account-only present |
| Expired again | account-only present | account-only absent |

The parent stayed pinned to the healthy account. The same thread reply and session survived all four steps. Captured outbound writes and stored post records agree; all callbacks returned HTTP 200.

Telegram Test Server `/think` still delivered all seven controls. The four empty-catalog edits are contract cleanup, not a claim of a distinct reproduced Telegram defect. Focused checks passed: 13 Mattermost, 21 Discord, 71 Slack.

The first local review rejected the earlier route-injected harness. The later review inspected the full registered-Gateway records and resolved that finding. No product source changed during this proof correction.

PR #144891, opened 2026-09-11
fix(ollama): preserve hosted subdomain model context

obviyus:
## What Problem This Solves

Onboarding a bare model name through a hosted service subdomain could save the local 32,768-token cap even when discovery reported a larger context.

## Why This Change Was Made

Discovery and context preparation now share the hosted-hostname classifier. Transport policy keeps its separate HTTPS-origin checks. Obsolete test-routing references are removed.

## User Impact

Hosted aliases preserve their advertised context; local models retain their existing cap. No configuration keys or migration change.

## Evidence

Real noninteractive onboarding against a controlled endpoint reported 262,144 tokens. Before the fix, the hosted alias saved a 32,768-token cap; afterward it saved no local cap. The local control retained the cap. Focused provider and routing tests passed 400 and 480 cases. All 17 sanitized configs reached Gateway readiness with documented repairs and an exposure override. Live hosted inference and graceful shutdown were not established.

Related: #136257.

### DIFF
--- a/config/knip.config.ts
+++ b/config/knip.config.ts
@@ -245,8 +245,6 @@ const rootEntries = [
   "scripts/e2e/*.{js,mjs,ts}!",
   "scripts/e2e/lib/**/{assertions,probe,mock-server}.{js,mjs,ts}!",
   "src/agents/prepared-model-catalog.worker.ts!",
-  // Loaded by URL from setup-inference-detection.ts; no static import edge exists.
-  "src/system-agent/setup-inference-detection.worker.ts!",
   // Split runtime loaded through a path assembled in subagent-registry.ts.
   "src/agents/subagents/registry/subagent-registry.runtime.ts!",
   // Loaded lazily by the sweeper only when a receipt-bearing or interrupted row is found.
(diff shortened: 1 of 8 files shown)


obviyus:
Landed after clean exact-head review and [green CI](https://github.com/openclaw/openclaw/actions/runs/34595780375).

**Proof:** Real onboarding queried a controlled Ollama HTTP fixture and saved the resulting configuration. Independent acceptance inspected the complete CLI, HTTP, and saved-config records.

| Endpoint | Baseline saved context | Candidate saved context |
| --- | --- | --- |
| Hosted Ollama subdomain | contextWindow 262144, contextTokens 32768 | contextWindow 262144, no local cap |
| Local loopback control | Existing local behavior | contextWindow 262144, contextTokens 32768 |

The canonical HTTPS-origin transport policy remains unchanged. The stale setup worker/test and two deleted list-test routes are removed; the live auth route and assertion baseline remain.

Validation passed: 400 Ollama checks, 480 routing checks, and readiness for all 17 real-Gateway config fixtures. Three corpus fixtures needed forced teardown after the fixed grace period, so graceful shutdown and sustained performance are not claimed. This is controlled HTTP-fixture proof, not live Ollama Cloud inference.

PR #144896, opened 2026-09-11
fix(qa): enable Telegram compaction progress evidence

RomneyDa:
## Summary

- enable Telegram tool-progress previews in the Claude compaction scenario
- retain commentary progress alongside tool progress
- guard the exact scenario streaming configuration in the catalog

## Failure evidence

Maturity run 34465387094 observed commentary previews but no compaction notice because the fixture enabled commentary only; compaction notices use the rolling tool-progress lane.

## Test plan

- focused Telegram and QA catalog tests: 30 passed on the rebased head
- pnpm check:changed
- local P0-P2 autoreview: clean

## Compatibility

Not applicable: this PR changes only one ephemeral QA scenario configuration and its catalog assertion. It does not change persisted state, a data model, the configuration schema, migrations, or production serialization. The toolProgress field is an existing optional production setting; this fixture now enables it for the scenario that asserts compaction progress.

### DIFF
--- a/extensions/qa-lab/src/scenario-catalog-channels.test.ts
+++ b/extensions/qa-lab/src/scenario-catalog-channels.test.ts
@@ -288,6 +288,7 @@ describe("qa scenario catalog channel contracts", () => {
     );
     expect(semanticFlow).not.toContain("received.at(-1)?.botApiMessageId");
     expect(compactionFlow).toContain('"minimumPreviewEvents":2');
+    expect(compactionFlow).toContain("progress: { commentary: true, toolProgress: true }");
     expect(compactionFlow).toContain("config.commentaryOne");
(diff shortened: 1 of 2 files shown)


RomneyDa:
Compatibility assessment for reviewed head 60c23b8244c6:

- This PR changes only an ephemeral QA scenario configuration and its catalog assertion.
- It does not change a persisted data model, state schema, config schema, migration, or production serialization contract.
- toolProgress is an existing optional setting already read by the production resolver; the fixture merely enables it for this one scenario.
- Therefore migration and upgrade compatibility proof is not applicable.

@clawsweeper please refresh the exact-head review with this scope evidence.

RomneyDa:
@clawsweeper review

PR #144916, opened 2026-09-11
fix(secrets): trust configured CAs for protected egress

### LINKED ISSUE #144830: [Bug]: Secret egress proxy cannot connect to privately trusted/self-signed HTTPS upstreams
### Bug type

### Beta release blocker

No

### Summary

The Secret Egress Proxy fails to connect to HTTPS upstreams that use an internally trusted or self-signed certificate, even when that certificate is trusted by the host OS and available through NODE_EXTRA_CA_CERTS.

The request reaches the egress proxy successfully, but the proxy then fails its own upstream TLS connection with:

Secret egress proxy could not reach the upstream host.

### Steps to reproduce

IIS Proxy for a local app.
NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt
Protected secret stored with the proxy set as allowed host
OpenClaw: 2026.9.x
Node: 24.19.0
Ubuntu host
Internal HTTPS upstream:
https://internal.example.local
resolves to a private RFC1918 address
Self-signed certificate trusted by Ubuntu
NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt
secrets.egressProxy.enabled=true
internal.example.local present in allowedHosts
bypassHosts=[]

### Expected behavior

Ability to communicate with service using the stored protected secret.

### Actual behavior

OpenClaw reports The curl command returned a 502 Bad Gateway error. The secret egress proxy could not reach the upstream host. This happens whether it's an authenticated endpoint or not.
Putting the host into the bypass list allows unprotected endpoints to work.

### OpenClaw version

2026.9.4

### Operating system

Ubuntu 26.04

### Install method

npm global

### Model

qwen 3.5 122B

### Provider / routing chain

openclaw

_No response_

### Logs

```shell

```

_No response_

### Impact and severity

_No response_

### Additional information

Root cause

The Secret Egress Proxy currently constructs its own upstream TLS agent approximately like this:

const upstreamTlsAgent = new Agent({
  ca: [...rootCertificates, caPem]
});

Because the proxy supplies an explicit ca list, privately trusted/self-signed upstream certificates are not automatically available to that TLS connection.

In testing, adding the required upstream certificate explicitly to this CA list resolves the issue.

Temporary workaround

The following local workaround resolved the issue.

The generated runtime bundle contained:

const upstreamTlsAgent = new Agent({
  ca: [...rootCertificates, caPem]
});

It was changed locally to:

const upstreamTlsAgent = new Agent({
  ca: [
    ...rootCertificates,
    fs.readFileSync(
      "/usr/local/share/ca-certificates/internal-service.crt",
      "utf8"
    ),
    caPem
  ]
});

After restarting the Gateway, Secret Egress requests to the internal HTTPS service succeed, including requests using protected-secret substitution.

This is only a temporary workaround because:

the certificate path is environment-specific;
generated dist files are replaced by OpenClaw upgrades;
it requires modifying compiled output directly.

Ayushdevo:
Closes #144830

## What Problem This Solves

Fixes an issue where protected secret-egress requests could not reach HTTPS services whose certificate was trusted through Node's configured CA sources, including `NODE_EXTRA_CA_CERTS`.

## Why This Change Was Made

Build both egress trust paths from a deduplicated union of Node's bundled roots and configured default certificates, then append the proxy's private CA. Bundled roots remain available when OpenSSL CA mode does not enumerate its lookup store.

## User Impact

Operators can use protected-secret substitution with internally trusted HTTPS services while retaining normal public HTTPS trust.

## Evidence

- Regression reproduced before the fix: the generated bundle omitted a configured default CA.
- Focused secret-egress suite: 21 tests passed after the fix.
- Formatting, type-aware lint, `git diff --check`, and automated review passed.
- Automated review identified the OpenSSL-CA fallback risk; the final union preserves bundled roots.

### DIFF
--- a/src/secrets/egress-proxy/proxy-server.test.ts
+++ b/src/secrets/egress-proxy/proxy-server.test.ts
@@ -4,7 +4,7 @@ import { createServer as createHttpsServer } from "node:https";
 import net, { type Socket } from "node:net";
 import os from "node:os";
 import path from "node:path";
-import tls from "node:tls";
+import tls, { getCACertificates, rootCertificates } from "node:tls";
 import { afterAll, afterEach, beforeEach, describe, expect, it } from "vitest";
 import { createTempDirTracker } from "../../../test/helpers/temp-dir.js";
 import { generateLocalProxyLeaf } from "../../proxy-capture/ca.js";
@@ -306,6 +306,13 @@ describe("secret egress proxy", () => {
 
   it("activates Node environment proxy support for registered Gateway runs", () => {
     expect(proxyEnv.NODE_USE_ENV_PROXY).toBe("1");
+  });
+
+  it("preserves Node's configured default CAs in the generated trust bundle", () => {
+    const trustBundle = fs.readFileSync(path.join(caDir, "trust-bundle.pem"), "utf8");
+    for (const certificate of [...rootCertificates, ...getCACertificates("default")]) {
+      expect(trustBundle).toContain(certificate);
+    }
   });
 
   it("survives a client that resets a refused tunnel instead of crashing the Gateway", async () => {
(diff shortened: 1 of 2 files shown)


VACInc:
Related, but not resolved by #144902. That merge restores fixed-length request framing; it does not change configured/private CA trust or TLS certificate validation. The CA-trust work tracked by #144830 and #144916 is separate, not a duplicate of the framing fix.

PR #144936, opened 2026-09-11
fix: claude-cli backend fails with spawn ENAMETOOLONG on Windows multi-MCP installs

### LINKED ISSUE #144792: bug: claude-cli backend still hits spawn ENAMETOOLONG on Windows — the --allowedTools list is not relocated by the #71600 fix
### Summary
On Windows, the `claude-cli` backend still fails with `spawn ENAMETOOLONG` on installations with many MCP servers, **even after the #71600 fix**. The reason: #71600 relocated only the *system/user prompt* off the command line, but the **allowed-tools list is still passed inline** as `--allowedTools tool_1,tool_2,…,tool_N`, which on its own exceeds the Windows command-line limit.

### Environment
- OS: Windows 11 (command-line length limit).
- Backend: `claude-cli` (subscription / OAuth), with many MCP servers registered.
- Affected: the 8.1+ / 9.x line, where tool-availability enforcement expands **every** tool name into `--allowedTools`. Version 6.34 does **not** do this and works fine.
- **Confirmed still present in the latest release, 2026.9.4 (published 11-Sep-2026)** — verified by inspecting `prepareClaudeCliTransportArgs` in `dist/cli-runtime-args-*.mjs`: it still ends with `args.push("--allowedTools", approvedTools.join(","))`, with no argv length check and no `--settings`/temp-file relocation for the tool list.

### Symptom
Agent turns fail immediately with `spawn ENAMETOOLONG`. In the gateway log it is often disguised as `reason=auth-profile` / `FailoverError`, which misleads toward an auth problem — it is not auth.

### Root cause (complements #71600)
`#71600` moved the prompt to a file (`--append-system-prompt-file`). But the allowed-tools list is still built inline:
`--allowedTools mcp__server__tool_1,mcp__server__tool_2, … ,mcp__server__tool_N`
On a multi-MCP install this argument alone is huge.

### Evidence (measured on a real install)
- **559 tools ≈ 27,128 characters** in the `--allowedTools` argument alone.
- A smaller instance with **139 tools ≈ 6,463 characters works fine** (same version, same OS) → this is purely a command-line-length problem driven by the tool count, not auth or packaging.
- Reducing tools below ~140 avoids it; adding MCPs reintroduces it.

### Why the #71600 fix is not enough
It covers the prompt, not the tools. For multi-MCP setups, `--allowedTools` is the **largest** contributor to the overflow, so the bug persists after #71600.

### Repro
1. Windows, `claude-cli` backend.
2. Register enough MCP servers to exceed ~500 tools.
3. Start an agent turn → `spawn ENAMETOOLONG`.

### Proposed fix
Relocate the tools off the command line, the same way #71600 did for the prompt (and the same way the Codex backend already uses a file via `systemPromptFileConfigKey`).

The `claude` CLI already accepts `--settings <file>`, a JSON file whose `permissions.allow` / `permissions.deny` are the exact equivalent of `--allowedTools` / `--disallowedTools` (verified with `claude --help`).

So, in the claude-cli argument builder, when the joined tool list is large — or unconditionally on Windows — write:
```json
{ "permissions": { "allow": ["mcp__server__tool_1", "…"], "deny": ["…"] } }
```
to a temp file and pass `--settings <path>` **instead of** `--allowedTools …` / `--disallowedTools …`.

Sketch:
```ts
const inlineLen = allowed.join(",").length + disallowed.join(",").length;
if (process.platform === "win32" || inlineLen > TOOLS_INLINE_THRESHOLD /* e.g. 6000 */) {
  const settingsPath = writeTempJson({ permissions: { allow: allowed, deny: disallowed } });
  args.push("--settings", settingsPath);   // reuse the temp-file + cleanup helper from #71600
} else {
  args.push("--allowedTools", allowed.join(","));
  if (disallowed.length) args.push("--disallowedTools", disallowed.join(","));
}
```
Notes:
- Reuse the temp-file lifecycle/cleanup helper introduced in #71600.
- No behavior change on non-Windows with short tool lists (still inline).
- Enforcement is unchanged — only the transport moves off the command line.

### Validation (proof of concept)
I reproduced both paths by spawning `claude.exe` directly (same shell path the gateway uses on Windows):
- **A — current behavior:** `claude --allowedTools <559 tools> --version` → command line ≈ 24,766 chars → **fails: "The command line is too long"** (claude never starts). This is the ENAMETOOLONG.
- **B — proposed fix:** the same 559 tools written to a 25 KB JSON file as `permissions.allow`, passed as `claude --settings <file> --version` → command line ≈ 313 chars → **works** (`2.1.263 (Claude Code)` prints normally).

Same tools, same machine — only the transport changed. The `--settings` file path is well under the limit and Claude runs.

### Test
- Windows + `claude-cli` + ≥559 tools → the turn starts and responds (before: ENAMETOOLONG).
- Verify disallowed tools are still blocked (enforcement preserved — `permissions.allow`/`deny` is the native settings mechanism).

### Impact
This is the last blocker keeping heavy-MCP Windows users on 6.34. Fixing it lets us run the latest version with the Claude subscription backend again.

fszcd:
Closes #144792

## What Problem This Solves

Fixes an issue where users on Windows running the `claude-cli` backend with many MCP servers would have every agent turn fail immediately with `spawn ENAMETOOLONG` (often surfaced misleadingly as `reason=auth-profile` / `FailoverError`), because the full tool-availability allow list is passed inline as `--allowedTools tool_1,tool_2,…` and overflows the Windows command-line limit. The earlier prompt relocation moved the system prompt off the command line, but the allowed-tools list — the largest remaining argument on multi-MCP installs — was still inline.

## Why This Change Was Made

When the joined allow list exceeds an 8 KiB inline budget (well below the ~32,767-character CreateProcess limit; a known-good 139-tool install measures ~6.5K characters, a failing 559-tool install ~27K), the transport arg builder now writes `{ "permissions": { "allow": [...] } }` to a private temp file and passes `--settings <path>` instead of `--allowedTools ...`. `permissions.allow` is the native settings equivalent of `--allowedTools`, so enforcement is unchanged — only the transport moves off the command line.

Key boundaries:

- Small allow lists stay inline; nothing changes for installs that work today.
- Claude Code honors a single `--settings` value, so when the restricted execution projection already passed inline settings JSON, the allow list is merged into that JSON (both `--settings` and `--settings=<json>` forms) rather than shadowing it with a second flag.
- An existing `--settings` file path (or non-JSON value) stays authoritative and the allow list falls back to inline, as does any failure to write the temp file — the fix never introduces a new failure mode.
- The temp file lives for the life of the long-lived stdio session and is removed when the session closes or if process spawn fails.

## User Impact

Heavy-MCP Windows installs can run current releases with the Claude subscription backend again instead of being pinned to 6.34: agent turns start and respond instead of dying at spawn. Non-Windows hosts and small tool lists see no behavior change.

## Evidence

**Real behavior proof through the actual `executeClaudeCli` agent path** (new tests in `extensions/anthropic/cli-runtime.test.ts`, driving the real stdio transport against a protocol-speaking Claude fixture subprocess):

- *Fresh turn with 600 tools (~29K characters of allowlist):* the spawned process argv contains no `--allowedTools`; it carries a single `--settings <path>` whose JSON holds every `mcp__openclaw__*` entry verbatim under `permissions.allow`. The full turn completes through the real protocol: initialize, UserPromptSubmit hook, PreToolUse hook, and `can_use_tool` authorization all still function with the file-based settings.
- *Reused turn:* a second turn on the warm session reuses the same process and argv (`turn: 2`, identical argv) — the settings file stays valid for the life of the session.
- *Session cleanup:* closing the live session removes the temporary settings file (`access()` rejects after `waitForExit()`).
- *Preserved settings:* when the restricted projection already passed inline `--settings` JSON, the relocated file keeps `disableAllHooks`, `enabledPlugins`, and the existing `permissions.deny` alongside the merged `permissions.allow`, and exactly one `--settings` flag survives.

**Real Windows run with actual Claude Code (2.1.233) through this branch's `executeClaudeCli` agent path**, 600 synthetic OpenClaw tools (inline `--allowedTools` would be 35,289 characters — over the Windows limit):

```
PROOF inline --allowedTools length would be: 35289 chars
PROOF spawned argv total length: 316 chars
PROOF argv contains --allowedTools: false
PROOF --settings file bytes: 40737
PROOF permissions.allow entries: 600
PROOF settings file removed after cleanup: true
PROOF real fresh turn result: "PROOF_TURN_1_OK"
PROOF settings workspace alive during session: true openclaw-claude-cli-settings-gLeWpq
PROOF real reused turn result: "PROOF_TURN_2_OK"
PROOF settings workspace removed after session close: true
```

The fresh and reused turns are real Claude Code agent turns interpreting the relocated settings file; the session holds the temp workspace while warm and releases it on close. (Synthetic tool names; no private endpoints, keys, or paths beyond the OS temp dir.)
Focused unit tests in `extensions/anthropic/cli-runtime-args.test.ts` (6 tests) additionally cover the inline small-list path, the `--settings=<json>` form, the file-path fallback, and the write-failure fallback.

Existing suites pass: `extensions/anthropic/cli-runtime.test.ts` + `extensions/anthropic/cli-shared.test.ts` + `cli-runtime-args.test.ts` (77 tests), full `pnpm test:extension anthropic` (592 passed; the only 3 failures are pre-existing `session-catalog.test.ts` symlink tests that require Windows developer mode and fail identically on clean `main`).
`scripts/check-src-extension-import-boundary.mts --json` clean; `node scripts/check-changed.mjs` lanes pass (typecheck, lint, format, boundary guards).

### DIFF
--- a/extensions/anthropic/cli-runtime-args.test.ts
+++ b/extensions/anthropic/cli-runtime-args.test.ts
@@ -1,0 +1,174 @@
+import { existsSync, readFileSync } from "node:fs";
+import path from "node:path";
+import type { CliBackendExecuteContext } from "openclaw/plugin-sdk/cli-backend";
+import { describe, expect, it } from "vitest";
+import { prepareClaudeCliTransportArgs } from "./cli-runtime-args.js";
+
+const BASE_ARGS = [
+  "-p",
+  "--output-format",
+  "stream-json",
+  "--include-partial-messages",
+  "--verbose",
+  "--setting-sources",
+  "user",
+  "--allowedTools",
+  "mcp__openclaw__*",
+  "--disallowedTools",
+  "ScheduleWakeup,CronCreate",
+] as const;
+
+function createContext(
+  overrides: Partial<CliBackendExecuteContext> = {},
+): CliBackendExecuteContext {
+  return {
+    command: "claude",
+    args: [...BASE_ARGS],
+    cwd: process.cwd(),
+    env: {},
+    prompt: "hi",
+    modelId: "claude-sonnet-4-5",
+    systemPrompt: "system",
+    useResume: false,
+    timeoutMs: 1000,
+    requestToolPermission: async () => ({ behavior: "deny", message: "nope" }),
+    requestUserInput: async () => ({ status: "cancelled", message: "nope" }),
+    ...overrides,
+  };
+}
+
+function manyToolNames(count: number): string[] {
+  // ~48 characters per name keeps the joined list well past the inline budget.
+  return Array.from(
+    { length: count },
+    (_, index) => `server-with-a-long-name__tool_with_a_long_name_${index}`,
+  );
+}
+
+function settingsArgValue(args: string[]): string | undefined {
+  const index = args.indexOf("--settings");
+  return index < 0 ? undefined : args[index + 1];
+}
+
+describe("prepareClaudeCliTransportArgs", () => {
+  it("keeps a small allow list inline without a settings file", () => {
+    const result = prepareClaudeCliTransportArgs(createContext());
+    const allowedIndex = result.args.indexOf("--allowedTools");
+    expect(allowedIndex).toBeGreaterThanOrEqual(0);
+    expect(result.args[allowedIndex + 1]).toBe("mcp__openclaw__*");
+    expect(result.args).not.toContain("--settings");
+    expect(result.cleanup).toBeUndefined();
+  });
+
+  it("relocates an oversized allow list into a temporary settings file", () => {
+    const toolNames = manyToolNames(400);
+    const result = prepareClaudeCliTransportArgs(
+      createContext({
+        toolAvailability: { native: ["read", "bash"], openClaw: toolNames },
+      }),
+    );
+    try {
+      expect(result.args).not.toContain("--allowedTools");
+      const settingsPath = settingsArgValue(result.args);
+      expect(settingsPath).toBeDefined();
+      expect(existsSync(settingsPath!)).toBe(true);
+      const settings = JSON.parse(readFileSync(settingsPath!, "utf8")) as {
+        permissions?: { allow?: string[] };
+      };
+      expect(settings.permissions?.allow).toEqual(
+        toolNames.map((name) => `mcp__openclaw__${name}`),
+      );
+      expect(result.cleanup).toBeDefined();
+    } finally {
+      result.cleanup?.();
+    }
+    const settingsPath = settingsArgValue(result.args)!;
+    expect(existsSync(settingsPath)).toBe(false);
+    expect(existsSync(path.dirname(settingsPath))).toBe(false);
+  });
+
+  it("merges the relocated allow list into existing inline settings JSON", () => {
+    const inlineSettings = JSON.stringify({
+      disableAllHooks: true,
+      enabledPlugins: {},
+      permissions: { deny: ["mcp__*"] },
+    });
+    const toolNames = manyToolNames(400);
+    const result = prepareClaudeCliTransportArgs(
+      createContext({
+        args: [...BASE_ARGS, "--settings", inlineSettings],
+        toolAvailability: { native: ["read"], openClaw: toolNames },
+      }),
+    );
+    try {
+      // Exactly one --settings flag survives; the inline JSON moved into the file.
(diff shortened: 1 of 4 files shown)


fszcd:
Closing this for now as the contribution workflow is being consolidated. If the change remains useful, it can be reopened or resubmitted later. Thank you for the review time.

PR #144940, opened 2026-09-11
fix(secrets): avoid unrelated reads during GitHub maintenance

steipete:
## What Problem This Solves

Resolves a problem where GitHub authorization maintenance repeatedly loads unrelated shared-secret values into memory as the secret store grows. Both device and OAuth reconciliation performed this work even without a configured GitHub identity.

## Why This Change Was Made

Narrow the existing indexed query to the requested GitHub name range before projecting values. Keep the exact JavaScript classifier, expiration and scope checks, matched-value redaction, name ordering, and fresh per-record reads. No schema or index change is needed.

## User Impact

GitHub maintenance allocates less memory and spends less synchronous time reading unrelated secrets. Authorization behavior stays the same.

## Evidence

- Real SQLite regression fails on the original code because unrelated value payloads cross the query boundary; it passes after the fix. Separate parity cases cover names, expiration, scope, and redaction.
- Focused coverage: 25 secret-store, 46 GitHub record, and 41 lifecycle tests passed.
- Local synthetic benchmark with 2,048 unrelated 256-byte values: one list materializes 2 records / 50 value bytes instead of 2,052 records / 524,386 bytes. A device+OAuth pair falls from median 3.83 ms to 0.48 ms across 15 samples after 5 warmups. With 512 unrelated 64 KiB values, the pair falls from 17.55 ms to 0.46 ms. These fixtures do not establish production stall attribution.
- Native query plans use the existing live index with a name range. Independent parity checks cover UTF-8 and both UTF-16 encodings; exact classification remains in JavaScript.
- Independent P2 autoreview found no actionable findings.
- Full `check:changed` passed.
- Linux Blacksmith owner-flow exercised actual SQLite and GitHub lifecycle maintenance: invalid device/OAuth rows were removed, a valid pending device and 64 unrelated secrets were retained, and a later exact read observed an updated record. The selected-source receipt and all assertions were retained. The wrapper exited 1 after a cleanup timeout; a separate native check confirmed the lease completed and its workflow ended. This is qualified owner-flow evidence, not a clean overall wrapper run or live GitHub authorization.

### DIFF
--- a/src/secrets/store/secret-store-hidden-github.ts
+++ b/src/secrets/store/secret-store-hidden-github.ts
@@ -283,6 +283,9 @@ export function listHiddenGitHubSecretRecordNames(params: {
             .select(["name", "value", "created_at_ms", "updated_at_ms"])
             .where("scope_kind", "=", "team")
             .where("scope_id", "=", "")
+            // Bound the existing name index before materializing values; the classifier stays exact.
+            .where("name", ">=", `${params.prefix}-`)
+            .where("name", "<", `${params.prefix}.`)
             .where("kind", "=", "secret")
             .where("allowed_hosts", "is", null)
             .where("deleted_at_ms", "is", null)
--- a/src/secrets/store/secret-store.test.ts
+++ b/src/secrets/store/secret-store.test.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
+import * as kyselySync from "../../infra/kysely-sync.js";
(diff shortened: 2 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [df58d27be38e44f740811c19647b7ff7dce90be8](https://github.com/openclaw/openclaw/pull/144940/commits/df58d27be38e44f740811c19647b7ff7dce90be8)
- Landed commit: [06985f991101c07efdcbb462028b6b8989c3f1da](https://github.com/openclaw/openclaw/commit/06985f991101c07efdcbb462028b6b8989c3f1da)

PR #144949, opened 2026-09-11
fix(tasks): preserve ACP completion during cancellation

Gabrielnkl:
## What Problem This Solves

Cancelling an ACP background task could replace a completed result with cancellation if normal completion arrived while the backend cancellation call was still pending.

Original fix and regression: @Gabrielnkl, with co-author credit retained.

## Why This Change Was Made

The task cancellation owner used the earlier task state after waiting for the backend. It now rereads the task and returns any recorded final outcome before writing cancellation, preserving the result, error, and end time.

## User Impact

Completed work keeps its result when completion wins the race. Cancellation that wins still reports success. Existing task storage and Gateway response shapes remain compatible.

## Evidence

- Real Gateway `tasks.cancel` and `tasks.get` calls used the ACP session manager and stdio transport, with a backend storage read held while the task completed.
- Before the fix, releasing that read changed the completed task to cancelled; afterward, the task stayed completed with its original result and end time, and cancellation returned `cancelled: false`.
- When cancellation won, it returned `cancelled: true` and preserved the recorded cancellation end time.

## Consumers

- Background-task cancellation callers: receive the recorded final outcome without overwriting completed work.
- Task status readers: retain the original result and end time after a late cancellation request.

### DIFF
--- a/src/tasks/task-registry-cancel.ts
+++ b/src/tasks/task-registry-cancel.ts
@@ -174,6 +174,15 @@ export async function cancelTaskById(params: {
             ? { expectedInstanceId: managedBacking.instanceId, expectedOwnerKey: task.ownerKey }
             : {}),
         });
+        // The run owns terminal outcomes published while backend cancellation waits.
+        const current = tasks.get(task.taskId);
+        if (current && isTerminalTaskStatus(current.status)) {
+          return {
+            found: true,
+            cancelled: current.status === "cancelled",
+            task: cloneTaskRecord(current),
+          };
+        }
       } else if (task.runtime === "subagent") {
         const { killSubagentRunAdmin } = await loadTaskRegistryControlRuntime();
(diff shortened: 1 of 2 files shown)


obviyus:
Thanks @Gabrielnkl!

**Proof:** Real Gateway `tasks.cancel` and `tasks.get` calls through the ACP session manager and ACPX stdio transport, with one backend store read held during cancellation.

- **Bug:** Normal completion became cancelled after the pending cancellation call returned.
- **Fix:** The cancellation owner preserves the recorded terminal result and correctly reports cancellation when cancellation wins.

| Before | After |
| --- | --- |
| Completed task changed to cancelled; `cancelled=true` | Completed task stays completed; `cancelled=false` |
| Cancellation could rewrite the recorded end time | A winning cancellation returns `cancelled=true` and retains the original end time |

Both race orderings passed independent behavior checks. The targeted task-registry and ACP cancellation suites passed all 164 tests. Your original fix and regression were reworked with contributor credit retained.

PR #140680, opened 2026-09-07
fix(ci): reuse verified runner toolchain archives

vincentkoc:
<details>
<summary>Additional instructions</summary>

This is a same-repository branch.

</details>

## What Problem This Solves

Resolves a problem where Crabbox validation repeatedly downloads Node and pnpm even when its runner image already contains the pinned public artifacts. Trusted hydration also discards prepared runtime state, adding avoidable work before dependency reconciliation.

During integration, the added consumer regression test exposed hosted CI inventory-growth cases that exceeded the existing 80-job cap. The maintainer repair preserves that cap and coverage rather than renaming the test or weakening assertions.

## Why This Change Was Made

Reuse original public archives from `/opt/crabbox/toolchain-archives`: copy them into private temporary storage, authenticate those exact bytes against source-owned digests, then extract fresh runtime state. Node remains pinned to 24.19.0 and pnpm to the existing 12.3.4 SHA512 package-manager pin. The native pnpm payload is independently authenticated and bound to its version and architecture.

The untrusted bootstrap preserves the reviewed-head check, IMDS no-role requirement, candidate-pin rejection, fresh Node extraction, isolated HOME, and frozen dependency install. Trusted Linux hydration retains the upstream Node compatibility selector and seeds a job-private Corepack home. `COREPACK_HOME` survives the existing saved-environment export. Existing executables, image markers, adjacent checksums, image Corepack metadata, and pack bundles are not integrity authority.

Missing or invalid archives retain authenticated download fallback. Dependency reconciliation, runtime pins, configuration, and protocols are unchanged. Image production remains separate.

Original packing stays first. The existing overflow-only anchors-first, duration-ordered alternate retains the same objects, estimator, admission predicate and fewer-bins adoption requirement. If the existing constructions still exceed the cap, one balanced reconstruction targets only selected eligible multi-file hosted-tooling tails without runtime partitions, prebuilds or dist requirements. A supplied parent set prevents repeated balanced reconstruction. Caps, weights, runner strength, admission, affinity, timing-generation rules, serial behavior and group limits are preserved.

This last bounded repair reuses Vincent Koc's five production hunks from [PR #127240](https://github.com/openclaw/openclaw/pull/127240), commit [`0ebe79ca15e18202da07a916e4be48152d9dc016`](https://github.com/openclaw/openclaw/commit/0ebe79ca15e18202da07a916e4be48152d9dc016), plus the five-line combined-growth fixture. Its broader branch and unrelated timing/capacity test changes are not included.

[...]

PR #144965, opened 2026-09-11
ci: keep growing hosted test plans within the job cap

steipete:
Refs #144941

<details>
<summary>Additional instructions</summary>

Maintainer-requested repair. This same-repository branch remains editable by maintainers.

</details>

## What Problem This Solves

Resolves a problem where pull-request CI fails the hosted tooling inventory-growth guard with `compact github node test plan exceeds 80 jobs (81 planned)`. This blocks #144941 (including run 34603439927), although that PR adds no test files. Main's push plan omits tooling, so its green CI does not exercise this PR-only failure.

## Why This Change Was Made

First-fit packing can strand time when a job reaches the ten-group limit. Before opening another GitHub PR anchor job, try one deterministic group exchange between existing bins. Validate both replacement bins and the incoming group with the original admission predicate before applying the exchange. This packs the grown inventory into 80 jobs without changing the cap, timing floors, serial budgets, runner capacity, worker pins, process isolation, prerequisites, or sibling-family separation.

The exchange runs before hosted tooling placement, so tooling growth cannot alter non-tooling anchors. Blacksmith, hybrid, and all push plans are byte-equivalent to their previous descriptors. Add a small regression for the independent time/group-count constraints and document the policy.

## User Impact

Unblocks the known planner failure on #144941 and other PRs routed through the affected tooling test lane. There is no runtime product behavior change.

## Evidence

- Reproduced on main snapshot `d1c9e4690adbf55df4f80993d52b1c62e6a0f09e`.
- With the new regression present and the original planner restored:
  `node scripts/run-vitest.mjs test/scripts/ci-node-test-plan.test.ts -t "packs serial groups within both|keeps hosted tooling within the GitHub job cap"`
  reports **2 failed**: three bins instead of two, and 81 planned jobs instead of at most 80.
- With the fix, those regressions pass on main and the clean merge-tree result of main with #144941 head `4b2b903c2dfd6edfee78b6a88e6cb34c44d26738`.
- `node scripts/run-vitest.mjs test/scripts/ci-node-test-plan.test.ts`: **57 passed**.
- `node scripts/check-changed.mjs` and `git diff --check`.
- Independent Codex autoreview at P1: clean.

| Profile | Push jobs before → after | PR jobs before → after |
| --- | --- | --- |
| Blacksmith | 32 → 32 | 49 → 49 |
| GitHub | 50 → 50 | 80 → 80 |
| Hybrid | 28 → 28 | 50 → 50 |

[...]

PR #144994, opened 2026-09-11
fix(ci): keep growing tooling plans within job limits

steipete:
Related: #144940, #140664, #140680

## What Problem This Solves

Fixes an issue where modest tooling-test inventory growth makes GitHub CI reject a workload that can fit the existing 80-job budget. The current main planner reproduced 81 jobs in its growth regression, independently of the change in #144940.

## Why This Change Was Made

When the existing plans cannot fit, the same file-partition owner can redistribute complete tooling families that do not require a runtime build. Non-tooling placement remains fixed. The allocator uses the existing costs, runner requirements and process policies, and permits one bounded file exchange before refusing an incomplete result.

Final membership regenerates the complete timing identities. Final admission rechecks the 80-job cap, measured timing floors and 150-second exclusive budget; only a single physical file can retain an indivisible over-budget estimate. The exchange examines at most 100,000 pairs. Already-successful plans and precise selection remain unchanged.

## User Impact

Tooling inventory can grow without increasing CI capacity or losing tests. Plans still fail explicitly when the existing constraints cannot be satisfied. There is no application-runtime behavior change.

## Evidence

- The original complete planner run had 55 passing tests and the reproduced 81/80 growth failure. The candidate passes all 57 tests with the original placement and coverage assertions retained.
- The growth cases verify complete timing identities. New negative coverage reproduces and rejects 120 fixed jobs escaping final admission, a sole multi-file stripe with a 151-second exact-generation sample, and source-only tooling placed into a dist-artifact job with spare capacity.
- Independent comparison verified 21 ordinary and precise plan outputs byte-for-byte against the baseline. Independent output verification checked 643 tooling files exactly once, 51 complete timing identities and strongest-runner-first assembly.
- One instrumented local growth run measured 299.921 ms in the new allocator, including 61,415 pair comparisons. This is a local planner observation, not an end-to-end CI speedup claim.

The complete changed gate and fresh independent review through P2 passed. Earlier review findings are resolved with failing-before/passing-after regression evidence. Exact-head CI and native preparation remain required before landing.


steipete:
Superseded by #144965 (`87504f7ec9e`), which landed first and fixes the same 81/80 overflow with one deterministic group exchange before the planner opens another anchor bin.

Proof against current `main` (planner untouched by later commits): this PR's own `test/scripts/ci-node-test-plan.test.ts` transplanted onto `main` passes its growth scenarios, and main plans exactly 80 jobs for all 12 inventory-growth variants the fixture covers (with and without `resolve-fs-safe-native-contract.test.ts`, plus `npm-package-locks-report`, ten growth probes, `openclaw-performance-crabbox`, `install-smoke-ref-admission`, and all three together). Push plans stay byte-equivalent; the cap is unchanged.

One real difference, recorded for a possible follow-up rather than as a reason to keep this open: this rewrite's final admission rejects a standalone multi-file stripe whose exact-generation estimate is 151 s (over the 150 s exclusive budget), while `main` accepts it as a lone 151 s job. The transplanted growth test stops at that assertion (PR test line 2507). If that stricter admission is wanted, it should come as its own small change with a fixture that isolates it; the family-redistribution allocator (+322/−122) is not needed for the reported failure.

Closing. Thanks for the parallel investigation; the inventory-growth fixture was useful.

PR #145015, opened 2026-09-11
fix: dismiss expired Skill Workshop notices

steipete:
## What Problem This Solves

Fixes an issue where Skill Workshop confirmations for Apply, Reject, Evaluate, and revision requests could remain visible after their notice timer expired, until unrelated activity repainted the page.

## Why This Change Was Made

The shared notice timer now requests a repaint through the existing page-scoped progress callback. All temporary-notice producers use the same currentness/progress options; persistent changed-suggestion warnings remain timer-free. Nearby test fixtures reuse the existing deferred-promise helper and one page-mount fixture.

## User Impact

Completed-action notices dismiss automatically while evaluation results and refresh errors remain visible. Warnings that a suggestion changed stay available for review.

## Evidence

Five rendered expiry cases failed before the fix; all 86 selected tests pass after it. Canonical checks, the production UI build, documentation checks, and independent P2 review pass.

The real browser flow uses a mocked Gateway and synthetic data. Each run makes one Apply request. Once the notice state expires, the old implementation still renders the toast; the fixed implementation removes it without another action. Both captures are taken after that same expiry condition. No live Gateway or skill publication was used.

| Before | After |
| --- | --- |
| Expired confirmation remains visible | Expired confirmation is removed |

### DIFF
--- a/docs/tools/skill-workshop/how-it-works.md
+++ b/docs/tools/skill-workshop/how-it-works.md
@@ -48,6 +48,9 @@ Open **Plugins → Workshop** and select the agent whose skills you want to insp
 - **Suggestions** contains pending proposals that you can evaluate, revise,
   apply, or reject.
 
+Action confirmations dismiss automatically. A warning that a suggestion changed
+stays visible so you can review the updated draft.
+
 Past applied, rejected, quarantined, and stale proposals remain available through
 CLI and Gateway inspection. They are not listed as a separate Control UI section
 or counted as installed skills.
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b4df209e466de0f532ef9b8acc289b5b2088809b](https://github.com/openclaw/openclaw/pull/145015/commits/b4df209e466de0f532ef9b8acc289b5b2088809b)
- Landed commit: [21a5c18302119ae5c4cbf93ff0477393847dabaa](https://github.com/openclaw/openclaw/commit/21a5c18302119ae5c4cbf93ff0477393847dabaa)

PR #145024, opened 2026-09-11
fix(mcp): accept HTTP 404 as already-terminated session on DELETE

### LINKED ISSUE #144860: [Bug]: MCP session DELETE returning 404 is treated as failed cleanup
### Bug type

### Beta release blocker

No

### Summary

The Streamable HTTP transport treats a 404 response to session termination as a cleanup error, which can fail runtime-tool-schema validation even when the session has already been terminated.

### Steps to reproduce

Isolated transport regression exercised during diagnosis:

1. Initialize the transport with a synthetic session ID and a stubbed HTTP response.
2. Make the session-termination DELETE return HTTP 404.
3. Call `terminateSession()`.
4. Observe that the original implementation throws; the targeted local change accepts the already-ended session.

This is a mocked transport-level reproduction. No public deployment fixture is claimed or required, and no live server address or session ID is included.

### Expected behavior

Termination should be idempotent when the server has already ended the session. The MCP transport specification says a server that has terminated a session responds with HTTP 404 to requests carrying that session ID.

This exception should apply narrowly to session termination, preserving failures for authentication and other server errors.

### Actual behavior

`src/agents/mcp-http-transport.ts` accepts successful responses and HTTP 405 for DELETE cleanup, but throws on HTTP 404. The live diagnostic failure disappeared after the narrow transport correction; runtime-tool-schema validation then passed with no findings.

Isolated regression results:

| DELETE response | Original behavior | Targeted correction |
| --- | --- | --- |
| 200 | accepted | accepted |
| 404 | throws | accepted as already terminated |
| 405 | accepted | accepted |
| 401, 403, 500 | throws | still throws |

Repeated termination after successful cleanup remained idempotent.

### OpenClaw version

Observed in 2026.9.3; the same condition was found in the 2026.9.4 bundle and inspected main-branch source.

### Operating system

Linux; the isolated defect is in transport response handling.

### Install method

npm global.

### Model

Not applicable; reproduced at the MCP transport layer.

### Provider / routing chain

Synthetic Streamable HTTP transport response; no real provider configuration included.

### Impact and severity

The observed cleanup error blocked runtime-tool-schema validation during update diagnosis. The synthetic HTTP-404 test reproduces the transport behavior; production frequency across other servers is unknown.

### Additional information

- [Transport implementation](https://github.com/openclaw/openclaw/blob/main/src/agents/mcp-http-transport.ts)
- [MCP session-management specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management)
- Related: [#119332](https://github.com/openclaw/openclaw/issues/119332) and [#126100](https://github.com/openclaw/openclaw/issues/126100). Those concern recovery of active requests/notification sessions after 404; this report concerns DELETE cleanup of a session that is ending.

No credentials, real server URLs, session identifiers, original logs or configuration are attached.

uoparaji:
Closes #144860

## What Problem This Solves

Fixes an issue where OpenClaw treats a stateful MCP Streamable HTTP session cleanup as failed when the server has already ended the session. The Streamable HTTP transport `terminateSession()` sends a DELETE carrying the session id; if the server already terminated that session it responds with HTTP 404 per the MCP spec, and the transport throws `StreamableHTTPError`. In the original diagnosis this cleanup failure blocked runtime tool-schema validation during an update, surfacing as an unexpected user-visible error even though the session was correctly ended.

## Why This Change Was Made

Session termination should be idempotent: a 404 from the DELETE means the session is already gone, which is the outcome the caller wanted. `OpenClawStreamableHTTPClientTransport.terminateSession()` now treats 404 the same way it already treats 405 — accept and mark the session terminated so a repeat call is a no-op. The exception is narrow: 401, 403, 500, aborts, and every other non-OK status still throw, so authentication failures and real server errors continue to surface.

## User Impact

- MCP server disconnects that race the server's own session cleanup no longer produce a spurious "Failed to terminate session: Not Found" error.
- No behavior change for connections that succeed, for servers that reject DELETE (405), or for servers that return real errors (401/403/500).
- No configuration surface changes.

## Evidence

- New unit test `accepts already-terminated session DELETE (HTTP 404) without sending it again` in `src/agents/mcp-http-transport.test.ts` — mirrors the existing 405 test and asserts: (a) `terminateSession()` resolves, (b) a second call does not re-issue DELETE (`onDelete` called once), (c) `disposeMcpClient` resolves to `"closed"`.
- Existing test `does not record a rejected DELETE as successful termination` (500 branch) is unmodified and continues to assert throw + retry behavior, so the narrowing to 404/405 is exercised in both directions.
- Behavior table (matches issue #144860):

  | DELETE response | Before | After |
  | --- | --- | --- |
  | 200 | accepted | accepted |
  | 404 | throws | accepted as already-terminated |
  | 405 | accepted | accepted |
  | 401 / 403 / 500 | throws | throws |

- Diff is 2 files, +26/−1 (one line of production logic + a spec-reference comment + one test).


steipete:
Thanks for the fix and regression coverage. #145831 already landed DELETE 404 acceptance in a0cbea0c22dc78d382a2a948b74cc1e309ec4eeb, which is present on current main.

It also keeps termination idempotent and includes a real HTTP 404 regression covering successful disposal. The remaining diff here adds a duplicate mock test and comment, so I’m closing this as superseded by that canonical fix.

PR #145028, opened 2026-09-11
fix(ui): keep sidebar sessions active during subagent work

steipete:
## What Problem This Solves

Fixes an issue where a session looked idle in the sidebar after its parent yielded, even though subagents were still working. Only the small child counter showed the outstanding work.

## Why This Change Was Made

The leading indicator now combines the session's own run state with the tree's descendant activity. Idle parents show an active ring labeled **Subagents working**, including nested and collapsed children. The tree accounts for the Gateway's subagent flag also covering a subagent's own queued run, preserving the paused **Queued** indicator. Actual execution state and runtime accounting stay unchanged.

Validation also repairs a pre-existing Gateway-update test fixture race that blocked CI: its dummy parent now stays alive until the simulated service manager receives the stop command. The fixture reuses the existing manager simulation; Gateway runtime code, ownership checks, and existing assertions remain unchanged.

## User Impact

Tasks remain visibly active during delegated work. A tooltip distinguishes subagent activity from the parent's own execution, and unread badges return when the work ends.

## Evidence

- Reproduced the missing circle in the browser and demonstrated failing regressions before the source fix, including the Gateway-shaped queued-leaf case.
- 398 focused sidebar/tree tests pass, with additional owner-chip and browser-rendering coverage.
- All 18 bundled-browser E2E tests pass across the sidebar, queue, and trailing-state suites.
- The previously failing Gateway-update case and all 48 tests in its full file pass after the fixture repair.
- Local changed checks, targeted type/lint checks, and independent reviews through P2 validate the changes. ClawSweeper's re-review confirmed the queued-leaf finding was resolved.
- Synthetic mocked-Gateway browser proof keeps the parent at `hasActiveRun=false` while demonstrating the ring and **Subagents working** tooltip.

Before:

Idle-looking parent with a working child

After:

Parent activity circle and Subagents working tooltip


steipete:
Fixed the queued-leaf finding in the tree projection. The Gateway flag includes a subagent’s own active or queued run, so only an idle, non-archived row can use it alone as proof of unloaded descendant activity. Loaded descendants still aggregate normally.

The existing queued-child fixture now carries `hasActiveSubagentRun=true`; it failed before the correction and passes afterward with its paused Queued ring. Added queued-leaf and archived-row count cases to the existing projection table. All 398 sidebar/tree tests, UI production/test typechecks, and focused lint pass. Independent review through P2 is clean. The earlier CI max-lines failure is fixed by moving the delegated-activity cases intact into their own module.

@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [ba8dd2e00416e2657c9b88a3fce7b13ae04fac7c](https://github.com/openclaw/openclaw/pull/145028/commits/ba8dd2e00416e2657c9b88a3fce7b13ae04fac7c)
- Landed commit: [e3ace7355bfe8540a395b3a47f47c6db36a0f18e](https://github.com/openclaw/openclaw/commit/e3ace7355bfe8540a395b3a47f47c6db36a0f18e)

PR #145030, opened 2026-09-11
refactor(sqlite): support existing-only worker acquisition

steipete:
Related: #144592

## What Problem This Solves

The worker broker always starts a creation factory. Canonical cache and discovery reads must preserve a missing database and must not initialize an existing file merely to check its contents.

## Why This Change Was Made

Add explicit existing-only acquisition to the same broker. An unowned missing path returns undefined without starting a worker or calling a factory. Cold existing acquisition requires an explicit backend factory that uses native no-create opening and avoids migrations; the creation factory is never substituted.

Check physical identity before and after backend-module loading, retaining the existing parent post-open check. Open intent stays outside actor identity, so aliases and ordinary/existing clients reuse the same owner and drainage.

## User Impact

Future canonical reads can use workers while preserving missing files and existing storage. Existing-only controls acquisition, not subsequent command permissions. The current ordinary worker stores keep their existing behavior.

## Evidence

- All 92 affected Node tests passed, covering ordinary/existing broker behavior, native SQLite wrappers, and Bun ownership fixtures; full selected checks and fresh independent Codex review through P2 passed.
- Rebased onto the landed Bun library/retirement repair. Every authored addition and deletion is unchanged; the sole conflict was import context.
- Fresh normal build and actual Node 24.20.0/Bun 1.4.2 source and compiled probes passed on the combined graph. Positive controls verify SQL/worker instrumentation and compiled-source refusal.
- All four modes record zero main-thread SQLite calls, no missing-path broker worker/factory/database, byte-preserving existing reads, native no-create races, alias reuse, and accepted-write drainage. Target descriptors return to zero with peer stores live and through 20 close/reopen cycles.
- Source and all generated artifact hashes stayed unchanged and every process joined. Build: 274.30s. Node source/compiled probes: 12.63s/5.48s; Bun source/compiled: 22.61s/3.69s. These are smoke timings, not throughput comparisons.

This supplies an acquisition prerequisite. Canonical shared-state execution, agent affinity, incognito lifetime and final-use admission remain part of the coordinated worker migration.

### DIFF
--- a/docs/plugins/sdk-overview/infrastructure.md
+++ b/docs/plugins/sdk-overview/infrastructure.md
@@ -96,6 +96,24 @@ Declare private build entries with
 [`openclaw.build.workerEntries`](/plugins/dependency-resolution#native-imports-from-a-standalone-source-build)
 and derive their locations from the loader's `api.runtimeSource` fact.
 
+Pass `existingOnly: true` when acquisition must preserve a missing database.
+The call returns `undefined` without starting a worker or invoking a factory
+when the file is absent and no active actor retains that path. A cold existing
+open requires the module's explicit
+`openExistingSqliteWorkerBackend(input, { databasePath })` export. The host
+never substitutes the ordinary creation factory. The existing factory must
+use SQLite's native read-only or existing-file opening mode and validate the
+current schema without creating or migrating it. A filesystem existence check
+followed by ordinary create-if-missing opening does not satisfy this contract.
+
+The host checks physical identity before dispatching the existing factory and
+again before returning the store. Disappearance or replacement after admission
+rejects acquisition. Existing-only and ordinary clients share the same physical
+actor when their module and initialization input match; changing open intent
+does not rerun a factory or create another connection. Domain commands still
+own write permission and any later schema initialization. Existing-only
+acquisition provides no read-only capability for subsequent commands.
+
 Abort signals remove operations that are still queued. Once dispatched, an
 operation retains its result or failure; cancellation does not prove rollback.
 `close()` rejects new work and drains that client's accepted operations. The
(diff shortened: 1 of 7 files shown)


steipete:
The compatibility checkbox is resolved by the source: this adds explicit existing-only worker acquisition and its factory contract, with no schema change or production caller migration. Missing paths return without dispatch; an existing backend must open without creation or migration, and identity changes are rejected.

The native proof covers unchanged database bytes, aliases, failure cleanup, and source/compiled Node execution on its sealed candidate graph. Exact-head hosted CI is green. This is a prerequisite for canonical storage migration; it does not claim that existing stores have moved, or that the later Bun/main graph was rebuilt by that earlier local proof.

steipete:
This patch changes acquisition behavior, not the stored data model: it adds no production table, schema version, migration, or production caller. The generic SQLite migration checkbox does not identify a migration requirement here.

The rebased head `7c17c87b0483972c29c125c0ee8d2c199ead6cb5` includes the landed Bun ownership repair (#145046). The combined graph passed 92 affected tests, the full selected checks, a normal build, and fresh Codex review through P2. Actual Node 24.20.0 and Bun 1.4.2 source/compiled probes preserved existing database bytes, rejected a native no-create race without recreating the missing path, shared aliases and drained accepted writes, and returned target descriptors to zero through 20 close/reopen cycles with peer stores live. Every primary probe recorded zero main-thread SQLite calls; positive controls verified the instrumentation. Current-head CI is running separately.

Canonical application-store migration remains outside this acquisition prerequisite.

steipete:
Merged via squash.

- Prepared head SHA: [7c17c87b0483972c29c125c0ee8d2c199ead6cb5](https://github.com/openclaw/openclaw/pull/145030/commits/7c17c87b0483972c29c125c0ee8d2c199ead6cb5)
- Landed commit: [744ebd948aa5411ad7cfca7272e04829bd67d1f5](https://github.com/openclaw/openclaw/commit/744ebd948aa5411ad7cfca7272e04829bd67d1f5)

PR #145032, opened 2026-09-11
fix(cli): reject an empty nodes invoke --idempotency-key instead of forwarding it

### LINKED ISSUE #145058: [Bug]: nodes invoke forwards a blank --idempotency-key to the Gateway and fails with a cryptic schema error
## Summary

`openclaw nodes invoke --idempotency-key ""` (or a whitespace-only value) is forwarded to the Gateway as a real key instead of being rejected locally. The empty string then fails `NodeInvokeParamsSchema` validation (`idempotencyKey: NonEmptyString`), so the entire request is rejected with a cryptic `invalid node.invoke params` error that never mentions the flag. A whitespace-only value is even accepted (`minLength: 1` passes) and becomes a meaningless literal dedup key, so a retry cannot be deduplicated the way the caller intended.

## Steps to reproduce

1. Run `openclaw nodes invoke --node <id> --command <cmd> --idempotency-key ""` against any reachable Gateway.
2. Observe the request is rejected with `invalid node.invoke params` after a network round trip, with no hint that the passed flag was blank.
3. Run the same command with `--idempotency-key "   "`: the value is accepted and sent to the paired node as the idempotency key instead of the fresh key an omitted flag would have generated.

## Expected behavior

An explicitly blank `--idempotency-key` is rejected by the CLI before any network activity, with a clear message naming the flag (e.g. `--idempotency-key must not be blank.`), matching the command's other blank-option guards (`--node`, `--command`, `--invoke-timeout`). An omitted flag still generates a fresh key; a nonblank value is forwarded exactly as typed.

## Actual behavior

The flag is resolved with `opts.idempotencyKey ?? randomIdempotencyKey()`. Nullish coalescing only catches `null`/`undefined`, so an explicitly blank value is forwarded verbatim: `""` is rejected late by Gateway schema validation, and `"   "` is accepted as a literal dedup key (pending node actions are deduplicated by exact key equality in `src/gateway/node-runtime-state.ts`).

## Environment

- OpenClaw version: main (2026-09-11)
- Affected command: `openclaw nodes invoke`

wangmiao0668000666:
Fixes #145058

## What Problem This Solves

Fixes an issue where operators running `openclaw nodes invoke
--idempotency-key ""` would have the whole command rejected with a cryptic
`invalid node.invoke params` error, with no hint that the flag they passed was
empty.

## Why This Change Was Made

The CLI resolved the flag with `opts.idempotencyKey ?? randomIdempotencyKey()`.
Nullish coalescing only catches `null`/`undefined`, so an explicitly empty value
was forwarded verbatim. The Gateway validates `node.invoke` with
`NodeInvokeParamsSchema`, which declares `idempotencyKey: NonEmptyString`, so `""`
fails validation and the entire request is rejected.

The flag is now resolved before the Gateway call, in the same validation block as
the command's other options: omitted generates a fresh key, an empty value is
rejected with `--idempotency-key must not be empty.`, and **every nonempty value
is forwarded byte-for-byte**, including a whitespace-only one. The reader is
`readNonEmptyStringPreservingWhitespace`, the existing nonempty-string reader in
`@openclaw/normalization-core/string-coerce`.

Two deliberate scope boundaries:

- **Only empty input is rejected.** The shipped Gateway accepts whitespace-only
  keys, so rejecting them locally would narrow an input contract callers can
  already retry with, and the upgrade note could not describe a replacement path.
- **Nothing is trimmed.** The Gateway deduplicates pending node actions by exact
  key equality (`src/gateway/node-runtime-state.ts:83`), so rewriting a
  caller-supplied key would change its identity — a retry would queue a second
  action, and previously distinct padded keys could collapse into one.

## User Impact

An operator who leaves the flag empty now gets a clear, actionable error before any
request is sent, instead of a cryptic schema rejection from the Gateway. Nothing
else changes: a caller-supplied key the Gateway already accepts — padded or
whitespace-only — is forwarded exactly as typed, including across CLI upgrades
where a pending action may already be queued under that key, and an omitted key
still generates one.

## Evidence

Regression coverage at the CLI command boundary (the real
`registerNodesInvokeCommands` action with the Gateway call captured):

- `invoke with an empty idempotency key` — rejected with
  `--idempotency-key must not be empty.`
- `forwards a caller-supplied idempotency key verbatim and generates one when
  omitted` — asserts the padded key `"  caller-key  "` and the whitespace-only key
  `"   "` both reach the Gateway byte-for-byte, and that an omitted flag generates
  exactly one random key.
- **Red before green**: with the source change reverted, the forwarding test fails
  on the whitespace-only key, because the previous revision rejected it locally
  with `--idempotency-key must not be blank.`
- Green: `node scripts/run-vitest.mjs src/cli/nodes-cli.coverage.test.ts` — 56
  passed. `pnpm tsgo:core` clean. `oxfmt` clean on both changed files.

Real CLI traces recorded for the previous revision, still valid for the paths they
cover because this change only moves whitespace-only keys from "rejected" to
"forwarded":

- **Empty key (after)**: `openclaw nodes invoke --node mac-1 --command canvas.eval
  --idempotency-key "" --url ws://127.0.0.1:1 --token proof-token` exits 1
  immediately with `--idempotency-key must not be empty.` The Gateway URL points at
  a closed port, so the diagnostic demonstrably fires before any network activity.
- **Before**: the identical command with `""` produced no local key diagnostic; it
  proceeded to network activity and failed with `Gateway not reachable at
  ws://127.0.0.1:1 (ECONNREFUSED)` — on a reachable Gateway the blank key was then
  rejected by `NodeInvokeParamsSchema`.
- **Nonblank keys still reach the network**: with `"  caller-key  "` the after-build
  likewise proceeds to the Gateway connection instead of a local key error.

Proof gap: the whitespace-only trace was not re-run against a packaged `dist` for
this head — this worktree's install is missing `@pierre/diffs` and
`@shikijs/langs/*`, so the `plugins:assets:build` phase of `pnpm build` cannot
complete here. The behavior is covered by the CLI-boundary regression above, which
asserts the exact params handed to the Gateway.

### DIFF
--- a/src/cli/nodes-cli.coverage.test.ts
+++ b/src/cli/nodes-cli.coverage.test.ts
@@ -287,6 +287,36 @@ describe("nodes-cli coverage", () => {
       message: "--node and --command required",
     },
     {
+      label: "invoke with an empty idempotency key",
+      command: "invoke",
+      args: [
+        "nodes",
+        "invoke",
+        "--node",
+        "mac-1",
+        "--command",
+        "canvas.eval",
+        "--idempotency-key",
+        "",
+      ],
+      message: "--idempotency-key must not be blank.",
+    },
+    {
+      label: "invoke with a blank idempotency key",
+      command: "invoke",
+      args: [
+        "nodes",
+        "invoke",
+        "--node",
+        "mac-1",
+        "--command",
+        "canvas.eval",
+        "--idempotency-key",
+        "   ",
+      ],
+      message: "--idempotency-key must not be blank.",
+    },
+    {
       label: "rename with a blank name",
       command: "rename",
       args: ["nodes", "rename", "--node", "mac-1", "--name", "   "],
@@ -371,6 +401,33 @@ describe("nodes-cli coverage", () => {
     expect(runtimeErrors.at(-1)).toContain("--params must be valid JSON.");
     expect(callGateway).not.toHaveBeenCalled();
     expect(lastNodeInvokeCall).toBeNull();
+  });
+
+  it("forwards a caller-supplied idempotency key and generates one when omitted", async () => {
+    const supplied = await runNodesCommand([
+      "nodes",
+      "invoke",
+      "--node",
+      "mac-1",
+      "--command",
+      "canvas.eval",
+      "--idempotency-key",
+      "  caller-key  ",
+    ]);
+    expect(supplied.params?.idempotencyKey).toBe("caller-key");
+    expect(randomIdempotencyKey).not.toHaveBeenCalled();
+
+    lastNodeInvokeCall = null;
+    const generated = await runNodesCommand([
+      "nodes",
+      "invoke",
+      "--node",
+      "mac-1",
+      "--command",
+      "canvas.eval",
+    ]);
+    expect(generated.params?.idempotencyKey).toBe("rk_test");
+    expect(randomIdempotencyKey).toHaveBeenCalledTimes(1);
   });
 
   it("invokes system.notify with provided fields", async () => {
--- a/src/cli/nodes-cli.coverage.test.ts
+++ b/src/cli/nodes-cli.coverage.test.ts
@@ -403,7 +403,7 @@ describe("nodes-cli coverage", () => {
     expect(lastNodeInvokeCall).toBeNull();
   });
 
-  it("forwards a caller-supplied idempotency key and generates one when omitted", async () => {
+  it("forwards a caller-supplied idempotency key verbatim and generates one when omitted", async () => {
     const supplied = await runNodesCommand([
       "nodes",
       "invoke",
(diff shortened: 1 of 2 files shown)


wangmiao0668000666:
@clawsweeper re-review

Addressed the P1 finding and the merge-risk item: the guard now uses `readNonBlankString`, so blank keys are still rejected locally while nonblank keys are forwarded byte-for-byte (pending-action dedup identity preserved). Added real packaged-CLI proof to the PR body: blank rejection fires before any network activity (unreachable Gateway URL), the pre-PR build proceeds to the network instead, and a padded nonblank key is still accepted.

obviyus:
Merged as 94c89b4350e962ee49b6c66594e51570bf293042. Empty `nodes invoke --idempotency-key` values now fail before node lookup or any Gateway connection, while whitespace-only and padded keys retain their exact bytes.

Verified 14 actual-CLI/WebSocket boundary cases per phase, the original regression failure, 62 passing focused tests, green exact-head CI, and a clean review. No CI exception was used. The transport fixture does not claim real device execution or pending-action deduplication proof.

Thanks @wangmiao0668000666 for the report and repair.

wangmiao0668000666:
@obviyus Thank you for reviewing and merging #145032! I appreciate the careful verification and the help getting this fix landed.

PR #145033, opened 2026-09-11
perf(reply): reduce cold imports during heartbeat preflight

steipete:
## What Problem This Solves

Heartbeat and reply preflight paid for cold execution-runtime imports even when they only needed delivery metadata or were handling an ordinary message. The complete 428-test heartbeat cohort spent about 11–12 seconds on avoidable work in a controlled Linux comparison.

## Why This Change Was Made

Share one synchronous harness-selection decision between metadata and runtime callers. Metadata reads delivery defaults without constructing the embedded harness; runtime callers retain the genuine selected harness and registry-owner validation.

Normalize stop-request ingress once, then load the cancellation operation for an eligible request. Authorization and current-target checks remain inside that operation after loading. The existing cancellation, queue, persistence, and cleanup body moves intact to its owner. Internal callers use that owner or the existing primitives; the public SDK export is preserved.

## User Impact

Ordinary messages and heartbeat preflight avoid initializing execution and cancellation machinery just to make lightweight decisions. This reduces cold startup work without changing shard counts, worker limits, configuration, or runtime routing policy.

## Evidence

On one isolated Linux runner with four available CPUs, two workers, fresh caches, and matching source/dependencies, all 25 files and 428 cases passed in every arm:

| Arm | Whole cohort | Maximum RSS (KiB) |
| --- | ---: | ---: |
| Original A | 138.177 s | 2,154,772 |
| Candidate | 127.179 s | 2,191,572 |
| Original B | 139.523 s | 2,173,004 |

The candidate is 8.0–8.8% faster than both controls; the controls differ by 1.35 seconds. No cases were removed or skipped. The single-case comparison was 20.138 / 16.828 / 19.578 seconds. These measurements do not establish whole-CI savings. RSS comes from the retained GNU `/usr/bin/time -v` output for those same arms: the largest individual-process high-water mark, not summed concurrent process-tree memory or a leak measurement. Candidate RSS is 27.04 MiB (1.28%) above the control mean; this is a time improvement, with no demonstrated memory reduction. This supplies the before/after RSS requested in the latest review and its rank-up move.

The original abort entry fails the new ordinary-message import-boundary regression, and the candidate passes. Another 341 focused cases cover selection identity, authorization, cancellation/persistence ordering, and sibling dispatch. Full build and isolated built Discord entrypoint profiling passed.

Every selected changed-check component passed through composite validation. The earlier full invocation failed on a test callback annotation; after that mechanical fix, the remaining 11 canonical checks passed. Final independent P2 review found no actionable issues. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34620973404) is green on `10434c57c3f8e564797c95e4f7a887cde56e0889`. The initial CI run exposed a stale test mock after moving delivery metadata to its owner; the reply itself still ran. Updating that mock preserved the provider/model and stable-facts assertions. The original 63-file/1,549-case CI cohort then passed locally, and the correction received a clean independent P2 review.

The test changes clean tracked fixture queues and observe cancellation while a replacement writer remains blocked, preserving the original ordering and replacement-state assertions.

## Scope

Production grows by 93 net lines for typed metadata/request boundaries and their adapters after moving the existing owners; tests grow by 74 lines. The old-path assertion debt ratchet tightens from 5 to 4. No dependency or persistence-schema change is included.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -1809,7 +1809,7 @@ src/agents/harness/lifecycle-hook-helpers.ts	1
 src/agents/harness/native-hook-relay-client.ts	2
 src/agents/harness/native-hook-relay-codec.ts	2
 src/agents/harness/native-hook-relay-state.ts	1
-src/agents/harness/selection.ts	5
+src/agents/harness/selection.ts	4
 src/agents/harness/support.ts	1
 src/agents/harness/tool-result-middleware.ts	1
 src/agents/identity.ts	8
--- a/src/agents/btw.ts
+++ b/src/agents/btw.ts
@@ -43,12 +43,12 @@ import { resolveEmbeddedAgentStream } from "./embedded-agent-runner/stream-resol
 import { createAgentHarnessHostCapabilities } from "./harness/host-capability.js";
 import { resolveAgentHarnessOwnerPluginId } from "./harness/registry.js";
 import { ensureSelectedAgentHarnessPlugin } from "./harness/runtime-plugin.js";
+import type { AgentHarnessPreparedModelProvider } from "./harness/selection-decision.js";
 import {
   resolveAvailableAgentHarnessPolicy,
   resolvePluginHarnessPolicyToolsAllow,
   selectAgentHarness,
   selectAgentHarnessForPreparedModelProviders,
-  type AgentHarnessPreparedModelProvider,
 } from "./harness/selection.js";
 import {
   resolveAgentHarnessPreparedAuthSupport,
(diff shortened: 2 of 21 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [10434c57c3f8e564797c95e4f7a887cde56e0889](https://github.com/openclaw/openclaw/pull/145033/commits/10434c57c3f8e564797c95e4f7a887cde56e0889)
- Landed commit: [341e533e7ba30b70c369d1f584d980608cc2e217](https://github.com/openclaw/openclaw/commit/341e533e7ba30b70c369d1f584d980608cc2e217)

PR #145038, opened 2026-09-11
feat(macos): run a local Gateway alongside a remote primary

steipete:
## What Problem This Solves

Mac users connected to a remote primary Gateway cannot also host and open a Gateway on the same Mac.

## Why This Change Was Made

Separate local hosting from the primary connection, reconcile the local launch agent and port ownership, and expose the hosted Gateway through the existing catalog and windows. The detailed solution and validation follow below.

## User Impact

Users can enable local hosting while keeping their remote primary and open the hosted Gateway as “This Mac” in the app and macOS CLI.

## Problem

The macOS app's primary connection is either local or remote (`gateway.mode`), and in remote mode the app deliberately stops the local Gateway. A Mac whose primary is a remote Gateway (for example an SSH tunnel to a Mac Studio) therefore cannot also run a Gateway of its own, even though saved Gateways already give independent windows for other remote hosts.

Two structural conflations made this impossible before: `gateway.mode=remote` meant both "clients dial elsewhere" and "this Mac does not host", and `gateway.port` doubled as the SSH tunnel's local port and the local Gateway's bind port.

## Solution

- **Hosting toggle.** With a remote primary, Connection → Connection gains **Also run a Gateway on this Mac** (device-local preference, no new `openclaw.json` keys). `GatewayAutostartPolicy` and `ConnectionModeCoordinator` run the existing local start sequence beside the remote tunnel; the control channel, node capabilities, Talk Mode, and menu bar status keep following the remote primary. Turning hosting off retires the hosted Gateway; pausing stops it.
- **Launch agent.** The managed agent is installed with the existing `gateway install --allow-unconfigured` (new install passthrough in `src/cli/daemon-cli` and `src/daemon/program-args.ts`), so the Gateway starts while `gateway.mode` stays `remote`; the argument is reconciled away when the primary becomes local. Service management now uses an explicit local-only CLI resolution (`CommandResolver.localOpenclawCommand`) so it can never be routed over the SSH primary.
- **Port ownership split.** The SSH tunnel's local port comes from the loopback port in `gateway.remote.url`; `gateway.port` is only the local bind port. `PrimaryGatewayControlConfiguration` stops writing `gateway.port` for SSH primaries, and `PortGuardian` sweeps the tunnel port and the hosted Gateway port separately. Configs written by older app versions (`gateway.port == tunnel port`) get a one-time, visible repair when hosting is enabled.
- **Catalog entry.** `DashboardGatewayTarget.local` ("This Mac", kind `local`, not promotable) appears in the Gateways menu, the dashboard Gateway picker, **File → New Gateway Window**, and `openclaw-mac gateway list`; `openclaw-mac status --json` gains `localGateway {hosting, port, running}`. Its connection is owned by the fleet and resolves the same non-publishing local endpoint derivation the local primary uses.
- **Server-side audit.** A running Gateway whose config says `mode: remote` previously selected the remote primary in a few of its own paths; the agent `gateway` tool, pairing setup URLs (`device.pair.setup`, `setup-code`), worker node enrollment, and the local-port credential surface in `client-bootstrap` now identify the hosting server.
- Docs: `docs/platforms/mac/remote.md` (new section), `docs/platforms/mac/bundled-gateway.md`, `docs/gateway/config-gateway.md`.

## Impact

Users with a remote primary can run and use a local Gateway from the same app. Existing local-mode and remote-mode behavior is unchanged when the toggle is off; the CLI keeps targeting the primary (documented, with the explicit env override to reach the local one).

## Evidence

- Focused native tests (`--disable-sandbox`, pure-logic suites): 64 tests in 11 suites, plus the local-routing suites from the follow-up commit (29 tests in 7 suites).
- TypeScript: 8 touched test files (333 tests) and `node scripts/check-changed.mjs` green (full branch before the follow-up; follow-up scope with `--base`).
- Swift lint/format, `pnpm native:i18n:verify`, `git diff --check` clean. Codex autoreview scoped-clean at P0–P2.
- Live proof on a throwaway app profile: primary = SSH loopback tunnel (`ws://127.0.0.1:28999` → stand-in Gateway 2026.9.4 on 18999). Enabling the toggle installed `ai.openclaw.<profile>` with `--allow-unconfigured`, the hosted Gateway (branch build) listened on the profile port, `openclaw-mac status --json` reported `localGateway.running: true` with "This Mac" connected, and a "This Mac" window showed the branch build while the primary window showed 2026.9.4. The launch-agent install/uninstall commands were observed through the profile-managed CLI (no SSH routing).

| Connection window before | After (toggle on) |
| --- | --- |
| before | after |

"This Mac" dashboard window on the hosted Gateway (footer shows the branch build), opened beside the remote primary window:

this mac

Not covered: the full native suite (disposable CI only), and node capabilities for the hosted Gateway (by design they stay with the primary).

## Follow-up

Unify primary + saved Gateways into one list with a default role (planned separately).


steipete:
Merged via squash.

- Prepared head SHA: [13d98d6c0e901b7116c86053c120dd6f601a2914](https://github.com/openclaw/openclaw/pull/145038/commits/13d98d6c0e901b7116c86053c120dd6f601a2914)
- Landed commit: [e3dd9ba437285421b5f936182158bde85b697b9f](https://github.com/openclaw/openclaw/commit/e3dd9ba437285421b5f936182158bde85b697b9f)

PR #145047, opened 2026-09-11
refactor(voice-call): await call persistence and lifecycle work

steipete:
## What Problem This Solves

Prepares Voice Call to use asynchronous call-record persistence without acknowledging callbacks, publishing active calls, or starting replies before their records are saved. Delayed writes must preserve capacity limits, speech interruption, transcript timeouts, and shutdown ordering.

## Why This Change Was Made

Call history, restore, lookup, and persistence use the existing asynchronous plugin-state contract. The manager serializes state and persistence decisions through the existing keyed queue, keeps uncommitted drafts private, and reserves capacity for pending outbound admissions. Carrier, playback, listening, and transcript waits remain outside that queue.

HTTP and streaming callbacks await persistence, startup joins required restore writes, and runtime stop drains admitted work. Automatic playback lets admitted speech finish its replay and turn-token checks. Expired transcript waiters cannot consume speech or steal replacement waiters. Token-bound realtime streams wait for pending identity updates before matching the carrier ID, avoiding rejection while the provider-ID write is still pending.

## User Impact

Call records retain their existing format, chunk limits, retention, metadata-last publication, replay behavior, and aliases. Failed persistence does not publish an active record or begin outbound playback. Existing synchronous map getters and telephony policies remain.

This is groundwork for moving SQLite execution off the main event loop. It does not itself move SQL to a worker or claim a performance improvement; that cutover must preserve the canonical shared-state owner, including configured store paths and aliases.

## Evidence

- All 896 Voice Call tests passed, including real SQLite delayed/rejected-write, capacity, replay, startup, shutdown, and speech-ordering tests. Real loopback WebSocket tests reproduce token-bound stream rejection before the fix and cover committed, failed, and retired-call outcomes.
- The complete changed-path gate passed; affected type, format, typed-lint, and ratchet checks passed again after the stream-admission fix. Fresh full-candidate Codex review found no actionable P0–P2 defects.
- Normal build passed in 319.40 seconds. Compiled public runtime proof passed through loopback HTTP and SDK-bound SQLite: held writes delay acknowledgment and shutdown, replay preserves history, and a fresh process reopens records and aliases. Write phase: 5.98 seconds; reopen: 2.91 seconds. Source and artifacts stayed unchanged, with no TypeScript fallback.

The compiled proof uses a programmatic full Plugin API host and mock provider, not normal Gateway activation or vendor telephony. Timings are smoke observations including normal shutdown grace, not benchmarks. Original counterexamples and prior proof remain preserved.

### DIFF
--- a/docs/plugins/voice-call/troubleshooting.md
+++ b/docs/plugins/voice-call/troubleshooting.md
@@ -14,10 +14,16 @@ Fixes for setup, webhook exposure, credentials, signature verification, Google M
 
 ### Call placement fails to save its initial record
 
-Voice Call saves the initial record before reserving a concurrency slot or
-contacting the carrier. If that write fails, the placement reports the storage
-error without dialing. Restore access to the state directory, then retry; the
-failed placement does not consume `maxConcurrentCalls` capacity.
+Voice Call reserves pending capacity while saving the initial record, then publishes
+the active call and contacts the carrier only after the write succeeds. If the write
+fails, placement reports the storage error without dialing and releases the reservation.
+Restore access to the state directory, then retry; a failed placement does not consume
+`maxConcurrentCalls` capacity.
+
+Webhook acknowledgments and transcript-driven replies wait for call-record persistence.
+Token-bound realtime streams wait for pending call updates before matching the carrier ID.
+Failed event writes remain retryable, and shutdown drains admitted call work after
+closing webhook and stream producers.
 
 ### Setup fails webhook exposure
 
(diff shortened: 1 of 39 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [35e91433b25c3479afe10daf3e316530362d832f](https://github.com/openclaw/openclaw/pull/145047/commits/35e91433b25c3479afe10daf3e316530362d832f)
- Landed commit: [ce7981f9c175407626475cc0429d6bbafa1a4e25](https://github.com/openclaw/openclaw/commit/ce7981f9c175407626475cc0429d6bbafa1a4e25)

PR #145064, opened 2026-09-11
fix: keep Node fixtures portable under Bun

steipete:
## Problem

When Vitest runs under Bun, Node-specific fixtures inherit Bun's `process.execPath`. That made isolated Node subprocess cases launch Bun and caused the Node runtime-recovery suite to exit through its intentional Bun guard before exercising any recovery behavior. Bun's `--bun` mode can also prepend a `node` shim that points back to Bun. Bun and Node return directory entries in different valid orders, making read-only Doctor assertions and multi-agent reset traversal order-dependent.

## Solution

Resolve an actual Node executable in shared test support whenever Bun owns Vitest. The resolver walks PATH candidates and rejects Bun-backed `node` shims before returning Node. Fixtures that explicitly exercise Node startup, service repair, and runtime admission use that executable. The recovery tests temporarily model the Node-only process fields under Bun, restore their original descriptors after every case, and keep all existing assertions active.

Sort recursive inventory snapshots before comparison so they continue to detect filesystem changes without depending on API enumeration order. Agent session cleanup now also sorts discovered directories before destructive reset, giving both runtimes the same stable sequence.

Production runtime behavior is unchanged apart from deterministic cleanup ordering. No dependency is added.

## Verification

- Custom Bun branch `steipete/bun:codex/openclaw-bun-compat`, containing the pending SQLite, realpath, and `fs.promises.access` fixes.
- Exact local release fork `1.4.3-canary.1+13bd48ed7`: 213 focused cases passed on the PR head (resolver 1, runtime recovery 73, install 99, cleanup 27, read-only Doctor 13).
- Earlier Bun portability sweep: 1,015 passed, 2 platform skips, 0 failed across 11 routed shards.
- Node focused replay: 244 passed before one load-sensitive Gateway readiness timeout; that exact case passed in isolation in 7.6 seconds.
- Node integration replay after the final fixture repair: 58 passed, 0 failed.
- Node and Bun `--bun` resolver regression: 1 passed under each runtime.
- `pnpm check:changed` passed after rebasing onto current `main`.
- Independent P0-P2 review found no actionable findings after both reported resolver/test issues were fixed.

Related Bun fixes: oven-sh/bun#40005, oven-sh/bun#42277, and oven-sh/bun#42279.

### DIFF
--- a/src/cli/daemon-cli/install.integration.test.ts
+++ b/src/cli/daemon-cli/install.integration.test.ts
@@ -17,6 +17,7 @@ import {
 import { buildSystemdUnit, parseSystemdExecStart } from "../../daemon/systemd-unit.js";
 import { makeTempWorkspace } from "../../test-helpers/workspace.js";
 import { captureEnv, withEnvAsync } from "../../test-utils/env.js";
+import { resolveTestNodeExecPath } from "../../test-utils/node-process.js";
 import { createCliRuntimeCapture } from "../test-runtime-capture.js";
 
 const { runtimeLogs, runtimeErrors, defaultRuntime, resetRuntimeCapture } =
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [5fe53429181e2fce74f7f56350f04fccc66340bc](https://github.com/openclaw/openclaw/pull/145064/commits/5fe53429181e2fce74f7f56350f04fccc66340bc)
- Landed commit: [2f5822abb67e81ab00e443594ada86779eb78c4b](https://github.com/openclaw/openclaw/commit/2f5822abb67e81ab00e443594ada86779eb78c4b)

PR #145067, opened 2026-09-11
fix(gateway): do not serve public health cache to admins

qingminglong:
<details>

</details>

## What Problem This Solves

Fixes an issue where an administrator calling the Gateway `health` RPC without an explicit probe could receive the public cached snapshot during the cache freshness window, so admin-only health diagnostics were missing until a later refresh.

## Why This Change Was Made

The Gateway health handler now uses the cache only for non-sensitive requests. Admin requests continue through the existing refresh path, preserving the existing public cache cadence, explicit probe behavior, runtime-staleness checks, and audience-specific health collection.

## User Impact

Administrators see the health information authorized for their scope immediately instead of receiving a stale public snapshot for up to one health refresh interval. Public readers retain the existing cache behavior.

## Evidence

- Baseline regression: on the pre-fix handler, a fresh cached snapshot was returned with { cached: true } for an operator.admin request without an explicit probe, so the request did not enter the sensitive refresh path.
- Final Gateway loopback transcript at the exact head below, using the same isolated Gateway instance and the same no-probe health request shape:

```text
operator.read  health({})  => meta: { cached: true }, response.ts: <cached snapshot timestamp>
operator.admin health({})  => meta: <absent>, response.ts: <fresh snapshot timestamp>
operator.admin health({}) => sensitive channel probe field: present
assertions: publicCached=true; adminCached=false; adminRefresh=true; audienceSpecificSensitiveResult=true
```

  The paired responses distinguish the changed behavior directly: the public request is served from the fresh public cache, while the admin request bypasses that cache and returns the audience-specific sensitive result from the refresh path.
- Focused Gateway regression tests passed for public cache cadence, explicit admin probes, and admin cache separation.
- The Gateway build completed successfully, followed by target-file formatting and git diff --check.
- Exact final head: 4a36c8c026b80287e55758a128fe968d1c2da52f.

The real Gateway proof covers the authenticated RPC path, cache metadata, and scope-selected response. The deterministic regression covers the cache-hit branch and refresh arguments. No provider or channel delivery behavior is claimed.


qingminglong:
@clawsweeper re-review

qingminglong:
@clawsweeper re-review

PR #145068, opened 2026-09-11
fix: continue cloud provisioning after a Gateway restart

steipete:
## What Problem This Solves

Fixes an issue where a cloud session could remain stuck when the Gateway restarted during worker provisioning. Its initial message could become permanently marked “Not sent,” while the restart recovery notice appeared above the message that prompted it.

## Why This Change Was Made

Pending turns now rejoin the exact provisioning operation through the existing environment recovery owner, including gaps between provider recovery attempts. Identical recovery requests share their existing operation. Normal completion releases that operation before accepting a retry; only an unresolved provider may publish an early foreground result. Stop, Move, replacement, and caller cancellation still invalidate held input.

The Control UI keeps reconciling through planned restart/suspension errors, retains the initial message's timestamp across reconnects, and anchors that message before later recovery output. Receipt deduplication and ordinary queued-message ordering remain intact. No configuration or persistence changes are needed.

## User Impact

A Gateway restart during provisioning no longer requires users to resend the initial message. Setup resumes on the intended worker without duplicate allocation, and the conversation remains in the correct order.

## Evidence

- Reproduced the missing-dispatch-owner failure through a real placement-store shutdown/reopen regression before the fix; two concurrent waiters now reach the same recovered worker, with one attachment and workspace sync.
- Reproduced the retained-provider overlap failure and verified that a later recovery pass preserves the original owner.
- [The first hosted CI run](https://github.com/openclaw/openclaw/actions/runs/34622104403/job/103338454138) caught an immediate-retry completion race. It reproduced locally and was fixed in the completion owner; all 112 affected recovery/cancellation tests now pass, including the original failing Stop test with unchanged assertions.
- 871 targeted tests pass across provisioning, cancellation, restart recovery, initial-message delivery, receipt deduplication, and chat rendering. The final recovery-helper extraction was rerun through 45 affected tests.
- Core, test, and UI type checks; changed-file lint; formatting; lifecycle/import guards; and unused-export checks pass. Independent review found no actionable P0–P2 defects.
- Message ordering was verified with synthetic data in Chrome. Validation uses local synthetic provider/transport fixtures; this has not been deployed to a live Gateway.

### Message ordering: before and after

Synthetic retained-message fixture rendered with the pre-fix chat view and the current chat view. These images demonstrate ordering and the retained Retry action; the provisioning-recovery behavior is covered by the shutdown/reopen and delivery tests above.

| Before | After |
| --- | --- |
| Recovery notice appears before the initial prompt | Initial prompt appears before recovery notices |

### DIFF
--- a/docs/gateway/cloud-workers/dispatching-a-session.md
+++ b/docs/gateway/cloud-workers/dispatching-a-session.md
@@ -19,6 +19,8 @@ In the Control UI, open **New Session** and use the unified **Place** picker to 
 
 With a GitHub repository selected, **Remote checkout** lets you choose the source ref without cloning on the Gateway. With a Gateway folder selected, cloud selection enables its managed worktree. The Gateway creates the session, finishes dispatch, and only then sends the first turn. The server badge in the session sidebar shows the durable placement state. Startup recovery retains the repository URL and ref along with the destination and first message.
 
+If the Gateway restarts during provisioning, the pending first message waits for recovery and continues automatically when its worker is ready. Temporary startup or suspension errors do not cancel setup. The first message stays before later recovery notices in the chat, including after reconnecting.
+
 Choosing **Stop cloud worker…** while the new session is still provisioning pauses its initial message before requesting teardown. A late dispatch response cannot send that message. The draft stays visible for **Retry** and is not resubmitted automatically. Regular session drafts survive reconnects and page reloads; incognito drafts remain only in the current page. If the first message was already sent, uncertain delivery remains **Check delivery** rather than starting another turn.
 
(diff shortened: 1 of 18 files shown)


steipete:
@clawsweeper re-review

The first hosted CI run exposed a normal-recovery completion race. The follow-up releases a completed operation before publishing its foreground result, while preserving early completion for genuinely pending providers. The original failing Stop test is unchanged and passes with all 112 affected recovery/cancellation tests. The PR description contains the updated evidence.

steipete:
Merged via squash.

- Prepared head SHA: [552a38d62cdca71691b8e2f319e54ee41b9b71c1](https://github.com/openclaw/openclaw/pull/145068/commits/552a38d62cdca71691b8e2f319e54ee41b9b71c1)
- Landed commit: [d2628903ab1ac4e634470a7b481145bd71ecc067](https://github.com/openclaw/openclaw/commit/d2628903ab1ac4e634470a7b481145bd71ecc067)

PR #145078, opened 2026-09-11
ci: avoid duplicate startup corpus runs on main

steipete:
## What Problem This Solves

Full canonical main CI runs the same startup corpus twice: through the Node runtime-config suite and again after baseline ratchets. The duplicate invocation recently occupied 579–639 seconds in the ratchet command. This adds work without exercising a different source tree or test contract.

## Why This Change Was Made

Give full canonical main pushes one corpus owner: the existing Node suite. Move the explicit corpus command into a separately visible step and skip it only when the existing preflight output selects full Node coverage on main.

Other baseline-ratchet runs retain that explicit step. Pull requests can select only changed tests, and release-gate dispatches can validate a different merge tree. All static ratchets still run.

## User Impact

CI avoids duplicate migration/startup execution on full main pushes. All 58 corpus cases, both prior-release snapshots, both state repair passes, and their state-integrity assertions remain. Runner counts, shard counts, workers, deadlines, and test inventory are unchanged.

## Evidence

The executable workflow/planner regression observed two corpus owners before the change and exactly one afterward. Independent P2 review is clean. Nine focused cases pass, including precise/fast-only PRs, non-main and noncanonical pushes, manual dispatch, release-gate dispatch, and the existing merge-tree ratchet guard. The test uses the canonical Vitest router and Node planner to establish retained corpus membership.

Targeted root-test types, lint, formatting, generated CI Git ownership, composite-action interpolation, and conflict-marker checks pass. Pinned actionlint passed. The full local workflow checker then stopped because its package index could not resolve the pinned pre-commit 4.6.2 bootstrap dependency; no substitute or bypass was used. [Exact-head hosted CI](https://github.com/openclaw/openclaw/actions/runs/34623667685) is green on `f136bec9fba3a691b6b2973b79390d2655110357`; all 149 checks are terminal without failures.

This is a routing change: corpus bodies and fixtures are untouched. The historical command timings are not a measured post-change CI saving, and removing the duplicate does not by itself establish the eight-minute goal.

Runtime code delta is zero; workflow +5 net lines supplies the conditional visible step, docs +7, tests +65. The old unconditional command is removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3066,7 +3066,6 @@ jobs:
               pnpm check:assertion-safety --base "$base_ref"
               pnpm config:docs:check
               pnpm plugins:inventory:check
-              node scripts/run-vitest.mjs run --config test/vitest/vitest.runtime-config.config.ts src/config/config-startup-corpus.test.ts src/config/state-startup-corpus.test.ts
               ;;
             release-lint-core-*)
               stripe="${TASK#release-lint-core-}"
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f136bec9fba3a691b6b2973b79390d2655110357](https://github.com/openclaw/openclaw/pull/145078/commits/f136bec9fba3a691b6b2973b79390d2655110357)
- Landed commit: [5ff271b457112183cb80d3ce679516a603ffcaad](https://github.com/openclaw/openclaw/commit/5ff271b457112183cb80d3ce679516a603ffcaad)

PR #144715, opened 2026-09-11
fix(update): allow config and service locks before handoff storage is initialized

fuller-stack-dev:
Related: #144208, #144314

## What Problem This Solves

Fixes an issue where users changing configuration or managing the Gateway service are refused when the private managed-handoff SQLite database exists but its handoff table has never been initialized. The store contains no custody records, yet the source check throws `no such table: managed_update_handoffs` before the mutation callback runs.

## Why This Change Was Made

Ordinary read-only source inspection now checks SQLite's schema catalog through the existing typed Kysely facade before querying retained leases. Only a genuinely absent schema object returns no borrowers. SQLite's case-insensitive name matching prevents an uppercase or wrong-shape object from being mistaken for absence. Query errors still propagate.

Retained v3 custody, cross-installation resource checks, strict retired-v1 handling from #144208, and captured recovery authority validation stay unchanged. No schema creation, migration, cleanup, service operation, or recovery API is added.

The existing integration coordinator and landing owner confirmed no duplicate fix. #144315 has no current lease-store delta beyond landed #144314. #144319 extracts this reader and must preserve both this absence check and #144208's retired-record filtering when rebased.

## User Impact

Ordinary config and service operations can proceed beside an uninitialized, healthy handoff store. Retained custody and unreadable or incompatible records still block them. No manual deletion or repair is needed for this harmless state.

## Evidence

- Exact production base `ada8b5f65083f9dba586349e3e91fd497fb5614b`: permanent connected regression was red with **1 failed, 22 passed**. The failure was the expected missing-table refusal.
- Corrected candidate `a86c71d9ba6172b753bb0557fc6f57fee45cf7d2`: **69 passed, 1 skipped** across the service-custody, retired-record, and captured-authority suites. The skipped case is the existing opt-in Bun qualification.
- Command: `node scripts/run-vitest.mjs src/daemon/service-operation-lock.retained-custody.test.ts src/infra/update-managed-service-handoff-retired-records.test.ts src/infra/update-managed-service-handoff-database.test.ts`.
- Real config and service lock callbacks execute against a synthetic healthy tableless store. Database bytes, inode, permissions, modification time, and directory contents remain unchanged. Negative controls cover retained custody, invalid records, corrupt storage, unsafe permissions, malformed tables/views/indexes, case variants, and captured missing-table authority.
- A fresh independent Codex review of the corrected complete patch plus unchanged owners completed **scoped-clean at P0–P2**, with no findings.
[...]

PR #143905, opened 2026-09-10
fix(update): prevent stale children from changing native services

fuller-stack-dev:
## Latest: settled capability-probe correction (`4255c6b1d18e`)

CI34623288834 passed the actual published-upgrade Docker lane, but compact18 refused the legacy revoked case at its capability probe. A focused real-process control exposed a normal-only predicate: valid code-zero/exit capability output was refused after cooperative descendant cleanup, even with the actual process group and descendant extinct. The forced-cleanup control remained refused.

The caller now requires code zero, exit termination, null signal, not killed, complete exact capability output, current original authority, and **normal or fully joined cooperative cleanup**. It still requests complete process-tree extinction. Forced, uncertain, cancellation, timeout, stale authority, incomplete output and unsupported capabilities remain failures. No timeout or cleanup mechanism changed.

Eight affected native/legacy controls passed; both new controls passed on final source after a test-only shadowed-local rename. The changed runtime was rebuilt and all three affected compiled finalizer cases passed. Typed lint passes. The two-file scoped P0-P2 review was clean before that identifier-only correction; production bytes remain exactly reviewed. The final full canonical local gate is finishing, and fresh exact-head hosted CI is required. An unrelated unchanged Windows worker-artifact tooling test timed out in the old CI and is not being edited here. Earlier proof and failures remain preserved.

Related: #143791

## Problem and scope

An interrupted update can leave a target-native command alive after its original owner or immediate spawner is revoked. This PR binds native receiver admission, process/file/signal effects, and configuration commits to the original update's live lease and registered process lineage. Unsupported targets refuse before stop/publication.

This is the native-child custody component and its public Doctor caller, not complete recovery-journey acceptance. The lease-store schema is unchanged.

Current published head: `231835929f940faaac8d273a8fcd441a2e6798a6`, tree `7cb9e913997da02ae3c0c770a136db1290e74552`. The native publisher verified exact-tree equality and preserved the signed reconciliation parent `7793e227`. GitHub reports the PR mergeable.

## Mainline reconciliation

The landing continuation integrates main through `86cd546d0bbe5110e813961430e9c0d3946c768e`, including #144778. A changed package generation retains **both** the original/spawner hierarchy and the active candidate-generation lease. The actual child PID/start identity binds both registrations before private input is delivered. Original revocation now refuses child effects, not merely eventual parent settlement. Root release remains blocked by surviving descendants.

[...]

PR #145085, opened 2026-09-11
fix(update): recover a handoff store left readable by an interrupted write

steipete:
## Problem

Creating the managed-handoff lease store is not atomic:

1. `open` creates the file — under the caller's umask, so **0644** with the common `umask 022`
2. the table is written to disk
3. only then does the store `chmod` it to 0600

Every reader checks the file mode *before* opening the database. So a crash anywhere in steps 1-2 leaves a world-readable file, and from then on `withConfigWriteLock` and `withGatewayServiceOperationLock` refuse with `managed handoff lease file is unsafe` — for **every install root on the host**, not just the interrupted one. The store is machine-global under `/tmp/openclaw` with no env override, so the only recovery was deleting the file by hand. `openclaw doctor` cannot help: its repair path writes config and dies on the same throw.

This is the other half of #144715. That PR made an uninitialized *table* readable, but the *mode* check runs first, so the interrupted-write case it was aimed at still refused. #144715 only helps hosts running `umask 077`; on the macOS and Linux default it never reached the new schema probe.

## Fix

Repair the invariant the store already maintains rather than refusing it. `repairPrivateFileMode` restores 0600 on a path we own, then revalidates. The `chmod` also moves ahead of the table write, shrinking the window it exists for from a disk write to one syscall.

Excess bits are defense in depth here rather than a live exposure: `assertPath` enforces a **0700 owned directory on every read and every write**, and `nlink === 1`, so no other user could traverse to the inode or hold a descriptor on it whatever the file's own mode said. Group/other **write** bits are still refused rather than repaired — `chmod` cannot revoke a descriptor, and integrity is the one thing the directory guarantee would not restore. Ownership, type and link count stay unrepairable and still refuse.

## Evidence

Live before/after on Linux (Docker, `umask 0022`), through the real installed `openclaw` CLI, in one image — the BEFORE arm removes only the repair call from the shipped bundle:

```
host: Linux aarch64   node: v24.19.0   umask: 0022
cli:  /home/appuser/.local/bin/openclaw -> /app/openclaw.mjs
pre-fix artifact: removed repairPrivateFileMode

===== BEFORE (repair removed from the shipped bundle) =====
  store: mode 644  size 0  tables=[]
  exit=1
  error: managed handoff lease file is unsafe
  store mode after: 644
  config key removed: NO

===== AFTER (shipped bundle as built) =====
  store: mode 644  size 0  tables=[]
  exit=0
  store mode after: 600
  config key removed: YES
```

Local:

- `src/infra/update-managed-service-handoff-store-recovery.test.ts` — new, and the first coverage these checks have had. Verified it fails on the original defect: with the repair reverted, the recovery case fails with the exact production error while the refusal cases keep passing, so they guard the boundary rather than the fix.
- 6 handoff/custody suites + `src/state/local-onboarding-state.test.ts`: **72 passed, 2 skipped** and **13 passed**.
- `node scripts/check-changed.mjs` — passed.
- Three independent Codex review rounds. Two findings accepted and fixed: POSIX mode assertions now gated to non-Windows, and previously-writable files are no longer repaired. The third round argued a retained *read* descriptor could observe later rows; that premise does not hold given the enforced 0700 directory and single link, so it is answered by the comment on `repairPrivateFileMode` and locked in by a test asserting the directory check still refuses once it stops being private.

## Known gap, deliberately not closed here

On a host with `umask 002` an interrupted write leaves **0664**, which this change refuses rather than repairs. Those hosts stay locked out. Refusing is the right call for write bits, but it means the recovery story is still incomplete, and it lands on the two gaps I did not touch:

- The refusal says `retain diagnostics and run openclaw triage manually`, but `triage` only ever touches the lease at its own install root; it cannot inspect or clear the row blocking a different one. The advised recovery does not exist for this failure.
- Nothing in Doctor inspects this store, so a corrupt one is undiagnosable by the tool users reach for — and unrunnable, since Doctor writes config.

@fuller-stack-dev — same area as #140339 / #144715, so flagging rather than quietly changing it. A Doctor-owned inspect/repair for this store looks like the right home for both.


steipete:
Also carries an unrelated red-main repair, because it blocks every pull request in the repo.

`c06f21ddfad` (#143905) added an authority case to `src/cli/daemon-cli/install.test.ts`, taking it from 1104 to 1126 raw lines — **1001 counted against a 1000 cap**. Main's push runs do not include the lint stripe, so it only surfaces on PRs, where `check-lint-core-1` now fails on a file the PR never touched. Reproduced locally on a clean merge of main, linting only that file:

```
src/cli/daemon-cli/install.test.ts:1126:4: error eslint(max-lines):
  File has too many lines (1001). help: Maximum allowed is 1000.
```

Fixed by shrinking rather than suppressing: `expectFields` asserts named fields and touches none of that suite's mocks, so it moved into the existing sibling `install.test-helpers.ts`, which is where the other shared fixtures already live. No test was removed or weakened — 42/42 still pass, and lint is clean on both files.

PR #145102, opened 2026-09-11
fix: preserve project and current task context in visible spawns

steipete:
## What Problem This Solves

Fixes an issue where a visible child task could not start from the current conversation when its parent was still running. A worktree child also selected the agent's default workspace instead of the registered project when its parent worked directly in that project. Once those failures were removed, live testing exposed that a forked visible child could follow the parent's earlier assignment instead of its own task.

Related: #144583.

## Why This Change Was Made

Resolve inherited worktree repositories through the parent's live registered project, with the existing preparation and rollback guards. Allow a same-agent spawn to snapshot its own active parent only while the admitted requester remains current; ordinary active-parent forks retain their existing restrictions. Reuse the native subagent task-message owner for visible children so the current assignment follows inherited history with explicit child context.

The existing session creation and transactional transcript-fork owners remain authoritative. This adds no protocol fields, storage schema, or configuration options.

## User Impact

Visible tasks can start in a separate worktree of the selected project, retain context from the parent's first in-progress turn, and execute their own assignment while the parent continues. Explicit directory/project choices and cross-agent boundaries keep their existing behavior.

## Evidence

- 216 focused tests passed across Gateway session creation, requester lifetime, project worktrees, spawn dispatch, and child-task envelopes.
- Full isolated Testbox `pnpm build` passed. Both real-provider live probes passed (2 tests, 33.85 seconds): EOF recovery and visible worktree spawn. All changed-code checks passed, including core/test typechecks, lint, dead exports, schema and storage guards.
- EOF probe: locally truncates a real Responses stream and verifies automatic recovery without repeating a completed side effect.
- Visible-child probe: a real Gateway forks the active first parent turn, creates a worktree in the registered project, completes the child README read with the inherited context marker, and lets the parent continue. Before the fix, project selection and active-parent admission failed; the bare task handoff then reproducibly followed the parent assignment despite receiving the child task last. The shared envelope passes the complete flow.
- Live execution used an isolated Testbox and synthetic Git repository; no operator Gateway was changed.

CI note: the first merge-ref run failed core lint because a concurrent main change left the existing daemon-install test at 1,001 executable lines (limit 1,000). This PR inlines two one-use locals into unchanged assertions; no test, assertion, or limit was removed. Targeted lint and an independent review of that cleanup passed.

Final exact-head [CI run](https://github.com/openclaw/openclaw/actions/runs/34628749979) passed. ClawSweeper reviewed the same head with no actionable findings.


steipete:
Merged via squash.

- Prepared head SHA: [cd0bc2ae93ce25273cfd3fbb0e45ac8668c8816d](https://github.com/openclaw/openclaw/pull/145102/commits/cd0bc2ae93ce25273cfd3fbb0e45ac8668c8816d)
- Landed commit: [58a6953549d011ffe44dcc3a7f3290b50e64b349](https://github.com/openclaw/openclaw/commit/58a6953549d011ffe44dcc3a7f3290b50e64b349)

PR #145107, opened 2026-09-11
fix(memory): preserve standing intents after prompt hook timeouts

### LINKED ISSUE #115450: Hook timeout releases lane but leaves hook child processes alive
## Version

OpenClaw `2026.7.1-2` (`0790d9f`), Linux/systemd.

## Summary

Async hook deadlines are implemented with `Promise.race`. When a hook times out, the caller/lane is released, but the underlying hook promise continues and child processes spawned by the hook remain alive. Repeated turns can therefore accumulate abandoned work inside the Gateway cgroup.

## Minimal reproduction

1. Register a startup-activated non-bundled `before_agent_run` hook with conversation access.
2. In the handler, spawn a child that ignores `SIGTERM`.
3. Return/await a promise that never settles.
4. Send an agent turn and wait for the default modifying-hook timeout.
5. Verify the turn returns/blocks after timeout.
6. Verify the child remains alive.
7. Repeat on the same session; a second child remains as well.

## Observed

- Two repeated turns left two child processes alive.
- The session lane remained reusable, so abandoned work can accumulate.
- Stopping the enclosing systemd cgroup killed both children.
- A separate 20-run transient-cgroup control test left zero descendants, confirming that whole-cgroup cleanup works.

## Expected

Hook timeout completion should mean owned work is cancelled and descendants are gone before the hook slot/lane is released.

## Code-level cause

The hook runner's timeout rejects the caller's wait via `Promise.race([hookPromise, timeoutPromise])`, but does not cancel `hookPromise` or terminate processes spawned by the hook. Changing timeout values alone cannot fix this lifecycle/ownership problem.

## Suggested contract

- Pass an `AbortSignal` to async hook handlers.
- Require bundled hooks to honor abort.
- Expose a tracked child-process API that terminates owned descendants (TERM, short grace, then KILL).
- Run untrusted or long-running hooks in a dedicated worker process/cgroup where possible.
- Confirm cleanup before releasing the hook slot/lane.
- Add repeated-timeout tests for zero descendants, lane reuse, and memory decay.

The reproduction used isolated HOME/XDG state, loopback-only services, a deterministic local model stub, no production credentials, and no external channels.

steipete:
Related: #144832, #116522

## What Problem This Solves

Fixes an issue where a standing intent could spend a reminder fire after its prompt hook had timed out, leaving the reminder unavailable even though its context never reached that prompt.

## Why This Change Was Made

The hook runner gives each `before_prompt_build` handler its own `hookInvocation.assertActive()` capability and expires it when that handler returns, rejects, or times out. Memory Core checks it after lazy loading and immediately before its synchronous database effects. Shared database admission and underlying work keep their existing ownership and cleanup.

This is a scoped result-eligibility guard for ordinary and authorized prompt phases. The separate cooperative-cancellation proposal in #116522 remains open; this change does not address its broader cancellation and child-process lifecycle scope or close #115450.

## User Impact

Late matching no longer consumes a standing-intent fire. A host without the required capability logs a clear diagnostic and skips matching without consuming the intent. The SDK field is optional for compatibility and grants neither tool authorization nor a guarantee of eventual model consumption.

## Evidence

- The actual registered-hook/SQLite regression failed on the original source: after the discarded result, the row became `done` with `fire_count: 1`. It passes with the repair. All 91 focused tests across seven files pass, including configured short deadlines, stale-first/live-second handlers in both phases, missing-host diagnostics, and a live caller sharing cold admission.
- The delay and hook timeout are controlled by the fixture/fake timers. Hook dispatch, SQLite effects, and the cold-open child in the coalesced-caller control are real; this is not measured production latency or live Gateway proof.
- Full selected changed checks passed. Fresh PR-head Linux CI at `d3f34c5ce8ac504dfcfb3e454d83ae176b70c749` executed 68 of the 91 focused cases across four files, including the registered-hook/SQLite timeout regression, missing-host diagnostic, live coalesced cold-open control, and all eight invocation cases. The remaining three unchanged plugin sibling files were verified locally. The standalone Testbox attempt stopped at provider status lookup before execution; its lease is closed and no standalone Linux result is claimed.
- Independent P2 code review is clean, with an additional source-based assessment of the distinct lifetime semantics versus #116522.

## Review Disposition

The host and Memory Core consumer are updated together in this PR; Memory Core is a private bundled plugin. Diagnostic skipping on an unsupported host is the intentional compatibility behavior, covered by the missing-capability regression and documented in the hook guide.

The data-model warning does not correspond to a stored-format change. The patch adds an invocation assertion before the existing standing-intent database callback; schema definitions, schema assurance, lifecycle SQL, claim SQL, row encoding, indexes, and migration versions are unchanged. Existing stored intents use the same database and transaction path when the handler remains active. No data migration is introduced or needed.

The controlled timeout regression proves the discarded-result/late-write invariant, not measured live Gateway latency. The first run failed an unrelated daemon-test file-length check on its synthetic merge base; main already repaired that check. The subsequent main integration leaves this PR’s full nine-file patch byte-identical to the reviewed original. Fresh Linux jobs passed again on the refreshed PR head: [Memory Core](https://github.com/openclaw/openclaw/actions/runs/34630860198/job/103367595725) and [hook invocation](https://github.com/openclaw/openclaw/actions/runs/34630860198/job/103367599658). Both ran on Ubuntu 24.04.3 with Node 24.19.0, using synthetic merge `fbda7e2b322e7bd6c2eaac47232af4e57cacc002` into main `58a6953549d011ffe44dcc3a7f3290b50e64b349`. No focused case was skipped. The failed standalone Testbox attempt remains unverified and is not counted as test evidence.


steipete:
Merged via squash.

- Prepared head SHA: [d3f34c5ce8ac504dfcfb3e454d83ae176b70c749](https://github.com/openclaw/openclaw/pull/145107/commits/d3f34c5ce8ac504dfcfb3e454d83ae176b70c749)
- Landed commit: [ddcbea7ec7617b9758d40509474e852c33f6434b](https://github.com/openclaw/openclaw/commit/ddcbea7ec7617b9758d40509474e852c33f6434b)

PR #145111, opened 2026-09-11
fix(tasks): explain when completion wins cancellation

obviyus:
## What Problem This Solves

Cancelling a background task could report only “Could not cancel task” when the task finished while cancellation was pending.

## Why This Change Was Made

Terminal reconciliation now uses the existing refusal helper to include the recorded outcome. This completes the result contract introduced in #144949 without changing stored status, timestamps, or successful cancellation.

## User Impact

The CLI, Gateway, subagents tool, and Tasks page explain why cancellation did not occur. Existing exit codes and completed task presentation remain unchanged.

## Evidence

Real Gateway, CLI, and registered subagents cancellation flows reproduced missing reasons before the fix and passed afterward. Browser captures cover both completion and cancellation winning the race. Registered Gateway tests passed 37 cases, task-registry tests 159, and cancellation-manager tests five. Chat task-panel presentation remains unverified; Workboard does not display the reason.

Consumers: cancellation responses now include the terminal-outcome reason for CLI, Gateway, subagents, and Tasks page users.

Related: #144949. Thanks @Gabrielnkl for the original cancellation-race work.

### DIFF
--- a/src/gateway/server-methods/tasks.test.ts
+++ b/src/gateway/server-methods/tasks.test.ts
@@ -25,6 +25,7 @@ import {
   markTaskTerminalById,
   recordTaskProgressByRunId,
 } from "../../tasks/runtime-internal.js";
+import { updateTaskStateByRunId } from "../../tasks/task-registry-record-api.js";
 import { reloadTaskRegistryFromStore } from "../../tasks/task-registry.js";
 import type { TaskRecord } from "../../tasks/task-registry.types.js";
 import {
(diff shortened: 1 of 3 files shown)


obviyus:
Thanks @Gabrielnkl for #144949. This follow-up addresses its round-1 finding F1.

The cancellation owner now explains a refusal when the task reached a terminal outcome while cancellation was pending. It preserves that outcome and end time. A cancellation that wins still returns `cancelled: true` without a reason.

| Before | After |
| --- | --- |
| CLI, Gateway and registered subagents tool return a bare refusal after completion wins. | Each includes the recorded terminal-outcome reason. |
| Tasks page shows `Could not cancel the task.` above a completed ACP row. | The same red banner explains `Task became succeeded while cancellation was in progress.` |
| Successful cancellation has no refusal banner. | Successful cancellation still has no refusal banner. |

Proof used real Gateway entry points and the Tasks-page Cancel button on pinned baseline and head. Registered Gateway tests passed 37 cases; owner tests passed 159; the cancellation-manager sibling passed five. All five terminal outcomes retain returned and stored end times.

Workboard's existing reason-dropping and refused-snapshot handling remain a separate consumer issue. Chat task-panel browser behavior remains unconfirmed.

Landed as `e1df2fcbba8d783295a6dbab5048fca04e7855c7` after exact-head census acceptance, clean review, merge-tree census and [green CI](https://github.com/openclaw/openclaw/actions/runs/34631719959). No CI or owner-review exception was used.

PR #145112, opened 2026-09-11
improve(chat): declutter the model picker

steipete:
## What Problem This Solves

The chat model picker had more chrome text than model rows. A "Some models could not be refreshed" banner sat above the list whenever a background catalog refresh failed, even though the list was already populated. A "Reset session model" footer duplicated what clicking the Default row already does. Account selection lived in a separate footer section with its own label, its own nested dropdown, and a two-line hint. On the New Session page the same footer carried an extra "Applies only to this session" hint.

## Why This Change Was Made

The picker now shows only the search field, the provider groups with the default model marked, and an **Account** category collapsed at the bottom of the same list.

- The refresh-failure banner is gone when a snapshot exists. Models rarely change, the list is already usable, and the provider gear still opens Models for a manual refresh. Empty-list states (loading, offline, error without a snapshot, no models, target-group errors) are unchanged, and the command palette keeps its own refresh notice.
- The reset footer is removed. The Default row already commits an empty model (`sessions.patch` with `model: null`), and the Gateway's sticky-selection persistence ignores non-string models, so the row is session-only in every selection-target mode exactly like the button was. Typing "default" in search already surfaces that row. To cover the one gap the button handled, a Default row is synthesized when the configured default is missing from the catalog, following the existing pattern for a pinned model that is missing from the catalog.
- Account selection is a collapsible category rendered with the same section, heading, and row markup as a provider group. Its rows join the picker's keyboard navigation and search. It still renders when the model list is empty (Automatic with no eligible models) and on the New Session page. The "Account for this chat" label and both hints are removed.
- Provider headings and the Default tag render in sentence case and the leading "·" before row metadata is gone.

Production LOC is net negative. No config, protocol, or Gateway changes.

## User Impact

Opening the model picker shows models, not explanations. Resetting a pinned session model is one click on the Default row. Choosing a saved account is one expand-then-click inside the same list, in Chat and in New Session.

## Evidence

Mock-Gateway capture (pinned session model, refresh failure flagged, two saved accounts):

| Before | After |
| --- | --- |
| before | after |

Account selection, before (nested dropdown footer) and after (expanded category):

| Before | After |
| --- | --- |
| before accounts | after accounts |

- UI unit lanes: chat-view, chat-pane-session-controls, new-session model-control and draft-submission-title suites pass (440 tests). New regressions: refresh failure with a non-empty list renders no catalog-state chrome; a default missing from the catalog still renders a Default row that commits the reset; the pinned reset stays session-only when model rows write to the agent or global target.
- Control UI e2e: model-picker-refresh, chat-model-controls (account responsive widths, pagination, search and keyboard reach), model-catalog-partial-refresh, new-session model-catalog and cloud-startup recovery pass.
- `pnpm ui:check-performance` green (startup JS 345.0 KiB gzip, CSS 44.7 KiB gzip).
- `node scripts/check-changed.mjs`: 22 steps passed across the core-tests, ui, and docs lanes.
- Follow-up commit after the first CI run: the synthesized Default row is limited to sessions whose row records a user pin (New Session drafts and agent-scoped catalogs never show a default they cannot offer), and the account row renderer joins the approved custom listbox inventory. Unit files rerun: 479 passed.
- Third commit after the second CI run: an unavailable Default row commits a reset only when the session row records a user pin (threaded as a picker flag); sessions inheriting their default and New Session drafts route to Model Setup as before. E2E rerun: chat-composer-catalog, model-catalog-partial-refresh, chat-flow.model-picker-refresh, chat-model-controls = 35 passed; check-changed green.
- Autoreview (Codex) on the full PR range: clean, no accepted or actionable findings.
- Docs updated: `docs/web/control-ui/chat.md`, `docs/concepts/multi-user.md`.
- Check-changed: 22 check steps passed across coreTests, UI, and docs; no environmental exceptions.
- Autoreview (Codex): clean, no accepted or actionable findings
- Initial CI failed on an unrelated daemon-install max-lines violation. Rebased onto the main fix in #145102; the picker patch is unchanged (`git range-diff`), and changed-file oxlint passed.
- CI at `7571cde6dad7139310e42d5d80f751a08762d4ed`: [run 34630516044](https://github.com/openclaw/openclaw/actions/runs/34630516044) is blocked by two New Session catalog assertions in `checks-ui (3/3)` and the custom-combobox ownership assertion in `checks-node-compact-large-4`. The latter now detects the account rows in `chat-model-account-control.ts`. Landing is paused for these substantive test/behavior decisions; other checks were still pending at handoff.


steipete:
Merged via squash.

- Prepared head SHA: [c0c0453a4f8b7fee808705a6d843cc62ac559a6a](https://github.com/openclaw/openclaw/pull/145112/commits/c0c0453a4f8b7fee808705a6d843cc62ac559a6a)
- Landed commit: [9252b5911bab1482c91dddebcddfcc76d0204906](https://github.com/openclaw/openclaw/commit/9252b5911bab1482c91dddebcddfcc76d0204906)

PR #145106, opened 2026-09-11
refactor(tests): share Workboard Gateway registration fixtures

steipete:
Related: #139428

## What Problem This Solves

Four Workboard Gateway tests repeat the same registration-capture setup, adding maintenance work when the fixture changes. This is part of the maintainer-requested test consolidation campaign.

## Why This Change Was Made

Move those four identical blocks into one file-local factory that creates fresh API objects, spies, and a method map on every call. Keep SQLite creation, actual Gateway registration, every assertion, and the other five specialized setups unchanged. This removes 43 net test lines.

## User Impact

No user-visible behavior changes. The Workboard tests retain their registration, scope, metadata, validation, and lifecycle coverage with less duplicated setup.

## Evidence

- Complete Gateway and sibling registration-lifecycle suites pass before and after: the same 15 ordered cases, all passed.
- Expanding the four factory calls reconstructs the original full test file byte for byte; formatting does not change the reviewed projection.
- The complete changed-file gate passes, including extension test typechecking, lint, formatting, boundary and dead-export guards.
- Independent P2 review found no issues. No whole-suite runtime improvement is claimed.

PR #145115, opened 2026-09-11
fix: restore core lint for daemon install tests

steipete:
Related: #145106

## What Problem This Solves

Resolves a problem where core lint rejects otherwise unrelated PRs because the daemon-install test exceeds its existing 1,000-line limit by one line.

## Why This Change Was Made

Remove the duplicated initial auth-mock implementation. The suite's `beforeEach` already resets the mock and supplies the same defaults before every case. All test scenarios and assertions remain unchanged; the test falls from 1,001 to 994 counted lines without changing the lint policy.

## User Impact

Restores the affected CI gate. No production behavior changes.

## Evidence

- Reproduced the exact original `eslint(max-lines)` failure and verified focused lint passes after the cleanup.
- Complete daemon-install suite: 42 passing cases before and after, with identical ordered case names and results on Node 24.19.
- Inverse-source comparison verifies every byte outside the one declaration is unchanged.
- Full mapped `check:changed` gate and independent P2 review passed.

### DIFF
--- a/src/cli/daemon-cli/install.test.ts
+++ b/src/cli/daemon-cli/install.test.ts
@@ -31,14 +31,7 @@ const hasConfiguredSecretInputMock = vi.hoisted(() =>
     return resolveSecretInputRefMock(value)?.ref != null;
   }),
 );
-const resolveGatewayAuthMock = vi.hoisted(() =>
-  vi.fn<() => ResolvedGatewayAuth>(() => ({
-    mode: "token",
-    token: undefined,
-    password: undefined,
-    allowTailscale: false,
-  })),
-);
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b71b777aa847595718bed78097c69e381a15f0d6](https://github.com/openclaw/openclaw/pull/145115/commits/b71b777aa847595718bed78097c69e381a15f0d6)
- Landed commit: [3ac42154d14a98ab2393e68bb3285df77c512b14](https://github.com/openclaw/openclaw/commit/3ac42154d14a98ab2393e68bb3285df77c512b14)

Issue #145120, opened 2026-09-11
Native Codex progress is not wired to the channel reply operation, causing false stall interruptions


metahacker:
## Summary
Healthy native Codex channel turns can be interrupted with “This turn was interrupted because it stopped making progress” while the app-server is still emitting assistant/tool activity.

Observed on official npm OpenClaw 2026.9.3; the missing host callback also exists in the 2026.9.4 source release and inspected main at 8947c2133591ab156635a9d1061a7bb72faaf5e0. Linux, native Codex app-server, Slack channel replies.

## Expected / actual
Actual app-server progress should refresh the reply operation that launched that run. Instead, the operation clock can remain stale while the backend is working. Recovery timing becomes dependent on whether a watchdog sample happens to classify the backend as a runtime-owned wait.

Production diagnostics showed an actively working turn classified `active_work_without_progress` and interrupted after about 7m42s. A comparable 6m55s turn was protected as `runtime_owned_wait`. These observations establish the symptom; they do not imply every long run is affected.

## Reproduction at the host boundary
1. Exercise `executeAgentTurn` with a registered reply operation and native embedded backend.
2. Have the backend invoke the supplied `onRunProgress` callback for a current-turn item notification.
3. Assert that the originating reply operation records activity.

On the original code, the host supplies no callback and the operation records no activity. Adding the owner-bound callback makes this regression pass. No live credentials or real channel are needed for the deterministic boundary test.

## Root cause and scope
`extensions/codex/src/app-server/run-attempt-turn-state.ts` emits `params.onRunProgress` for active-run progress. `src/auto-reply/reply/agent-runner-embedded-candidate.ts` does not wire that callback to the captured turn’s `replyOperation.recordActivity()`. The CLI candidate already uses its equivalent owner-bound activity callback.

A small host-side fix reuses the existing reply-operation owner; it does not widen the Plugin SDK, weaken stall thresholds, disable recovery, or look up a successor by mutable session identifiers.

This is distinct from resumed native turns lacking liveness adoption (issue 113972), provider/model silence, and tool-loop detection thresholds.

## Evidence
A local three-file fix and focused failing-before/passing-after regression are available. Focused/adjacent tests, types, SDK surface checks, and core lint passed on the 9.4 candidate. Independent scoped review reports no findings. Production activation and long-turn live acceptance are still pending and will be reported separately.

PR #145121, opened 2026-09-11
fix(codex): prevent false stalls while app-server work progresses

metahacker:
Related: #145120

## What Problem This Solves

Healthy native Codex channel turns can receive a false “stopped making progress” interruption while the app-server continues producing activity. The original revision wired progress to the captured reply operation. A later recurrence exposed the separate diagnostic watchdog clock: that clock can remain stale while the reply clock advances.

The original real long-turn proof below remains valid for reply-owner freshness and final delivery, but did not cover the failing diagnostic `embedded_run` path. This revision explicitly tests that path rather than treating survival of a tool-heavy turn as universal acceptance.

## Why This Change Was Made

Keep the original captured-reply callback and bind progress to the exact admitted backend handle at the existing host execution-binding seam. A synchronous host callback now updates diagnostic progress when that same owner is active and the diagnostic phase is an ownerless `embedded_run`. It is independent of the optional, queued, best-effort `run.progress` telemetry channel.

- Capture the handle once; never adopt a replacement based on matching session/run strings.
- Preserve existing admission, lifecycle generation, tool-authority fingerprint, and reply-operation/backend/route checks.
- Revalidate after runtime status probes and diagnostic snapshot reads before updating either clock.
- Reject late progress from stopped, aborted, replaced, closed, or returned attempts.
- Do not use native transport traffic to renew a core model/tool phase or semantic argument-churn/repeated-request clock.
- No new SDK exports, diagnostic owner, configuration, timeout change, persistent state, or permission grant.

The current-owner registry read validates the captured handle; it does not look up a successor to attribute delayed activity to. Tool authority itself is unchanged; its existing registration/active assertions bind this host-owned progress callback.

## User Impact

Advancing native work refreshes both relevant liveness clocks. Runtime-owned waits retain their existing protection. Genuinely idle work remains recoverable, and stale callbacks cannot keep a successor or stuck tool alive.

## Evidence — September 14 follow-up

Current upstream base: `6d586170048982bd6c5ebc3cdc08bbd003f47a1d`. The original three-file patch reapplied unchanged (`git range-diff` reports `=`); the new bridge adds one production module change and a dedicated regression file.

- Twelve new regression cases compose actual run admission, backend registration, reply operation, diagnostic heartbeat and stuck-session recovery. The watchdog runs enabled at accelerated test thresholds: active callbacks survive multiple recovery windows; a quiet runtime-owned wait survives; once that wait ends, actual recovery aborts and settles the idle run.
- Negative coverage includes same-string replacement, admission closure, lifecycle rotation, stopped/returned callbacks, reentrant runtime probes, replaced reply operation/backend/route, and a tool phase whose clock must not be renewed by notification traffic.
- Release-based focused proof: 206 unique tests passed across the new regression, existing authority, diagnostic, actual recovery, and run lifecycle suites. Final fixture typing/formatting and deterministic timer corrections are included in the submitted test.
- Exact live-source baseline before/after: the final 12-case regression fails 11 / passes 1 without the bridge, then passes 12 / 12 with it. Current-main proof also passes the original callback test, 33 existing authority tests and the final 12-case regression (46 distinct cases; an earlier real-clock fixture race was corrected with deterministic test clocks).
- Full current-main `node scripts/check-changed.mjs --base 6d586170048982bd6c5ebc3cdc08bbd003f47a1d` completed with exit 0, including core types, all core-test type graphs, formatting/lint, dead-code scans, import cycles and repository policy guards. Release-based targeted types/lint also pass; the release broad gate stops at its pre-existing fs-safe 0.8.5 patch allowlist guard, unchanged by this patch.
- The isolated combined 9.4 deployment candidate completed its full build successfully in 5m24s, preserving the current production companion repair. All 28 combined diagnostic/cancellation-scope/registered-bridge tests passed. The candidate remains separate from production; activation and fresh real-channel acceptance are not yet claimed.
- The prescribed autoreview helper was attempted with the configured model, but returned `reviewer_unavailable`/`engine_failed` because its model cache lacked `base_instructions`; it produced no verdict. A final independent native reviewer subsequently completed the full five-file source scope and affected registration, admission, reply, diagnostic, recovery, and native-notification contracts: scoped-clean for introduced P0–P2 findings. The reviewer verified actual watchdog enablement, exact handle capture and revalidation, core semantic-phase exclusion, and runtime-owned wait preservation. The earlier transport-blocked scope is now completed. This does not recast the failed helper attempt as a successful helper review. The original accepted patch remains unchanged; its range-diff and source composition were verified.

### Causal limits and current readiness

The two-clock dependency and synchronous repair are demonstrated. The precise reason production telemetry failed to advance the clock in the reported recurrence is still unproven. An earlier local theory that exact-owner rejection caused those missed events was withdrawn: the local native handle did not have that diagnostic owner. Do not infer that every interruption has this cause.

This PR remains draft pending maintainer/security-owner review and fresh real-channel activation acceptance of the combined change. Historical 660-second native evidence validates the original callback, not the new diagnostic bridge. No production rollout, broad CI success, or universal stall-fix claim is made.

## Historical original-patch native long-turn evidence — September 11

The acceptance used the actual source-runtime Gateway, the real QA channel plugin, native Codex execution, and supporting assets built from exact head `3b3b97e16535d28549d69bde0614faf40ff11cbf`. The delivery surface was an isolated synthetic QA channel, **not Slack**, so no real user or workspace data was involved. This is a source-runtime proof with exact-built assets, not a packaged-install claim.

The harness transparently wrapped the captured reply owner's existing `recordActivity` method with a call-through observer that preserved its receiver and arguments. A clock advance was credited to this proof only when its immediate production caller was `onRunProgress` in the embedded candidate. Generic diagnostics and other owner writes were excluded, so survival alone was not treated as proof.

Redacted independent readback:

```text
source_head = build_commit = 3b3b97e16535d28549d69bde0614faf40ff11cbf
native_work_elapsed_ms = 660000
native_work_progress_beats = 132
single_stable_reply_owner = true
causal_onRunProgress_calls = 483
causal_clock_advances = 477
causal_progress_span_ms = 669434
causal_advances_after_600000_ms = 70
fresh_native_progress_events_after_600000_ms = 70
max_causal_progress_gap_ms = 5001
max_running_owner_age_ms = 4408
owner_age_after_600000_ms = 6
recovery_events = 0
originating_thread_final_delivered = true
owner_completed = true
owner_cleared = true
isolated_process_and_ports_closed = true
```

The ordinary recovery diagnostics and thresholds remained enabled and unchanged: 120-second warning, six-minute active-abort threshold, ten-minute quiet takeover, and the normal 15-minute long-tool floor. The trace recorded repeated causally attributed activity beyond ten minutes, normal completion, exact-thread final delivery, and clean owner/process/port cleanup. No recovery interruption or owner replacement occurred.

This is positive after-fix evidence, not a claim that the previous revision would necessarily be aborted by this exact synthetic workload. No private runtime log, fixture path, credential, endpoint, model identifier, or user data is included.


metahacker:
@clawsweeper re-review

Thank you for the precise direction. I’ve added the requested redacted real after-fix long-turn evidence to the existing PR body, including causal owner-clock attribution beyond the recovery windows, normal completion, exact-thread delivery, and cleanup. The PR remains draft pending your review.

PR #145122, opened 2026-09-11
fix: hide dismissed snapshot builds when refresh fails

steipete:
## What Problem This Solves

Fixes an issue where users dismissing a failed build in **Settings → Cloud workers → Snapshots** would still see its row and an unchanged **Needs attention** count when the follow-up inventory refresh stalled or failed, despite the success confirmation.

## Why This Change Was Made

Update the page's displayed failed-build list as soon as confirmed destruction succeeds. The existing dismissal set keeps the row hidden on later refreshes; Gateway record retention remains unchanged.

## User Impact

Dismissed builds disappear immediately, while inventory refresh errors remain visible. Pending or failed destruction still leaves the build available for retry.

## Evidence

The strengthened mounted-page regression failed before the fix; all 34 related page and snapshot tests pass afterward. It covers pending destruction, pending and rejected refresh, and a later successful refresh that still returns the retained build record.

Real Chromium with synthetic data and a mocked Gateway performed one confirmed dismissal per run. Before the fix, the row remained both while refresh was pending and after it failed; afterward it was absent in both states. Neither run produced a page error. The captures below show the failed-refresh state, with both the success confirmation and inventory error preserved.

| Before | After |
| --- | --- |
| Dismissal confirmation with the failed build still visible | Dismissed build removed while the refresh error remains visible |

Canonical changed-file checks and the UI production build pass.
Independent managed review completed without actionable findings.

### DIFF
--- a/ui/src/pages/cloud-workers/cloud-worker-snapshot-builds.test.ts
+++ b/ui/src/pages/cloud-workers/cloud-worker-snapshot-builds.test.ts
@@ -520,11 +520,23 @@ describe("Snapshot builds", () => {
     }
   });
 
-  it("dismisses a failed build the Gateway still records and keeps it hidden on refresh", async () => {
+  it("dismisses a failed build before refresh finishes and keeps it hidden after refresh fails", async () => {
     const environments = [buildFixture("failed", "Gateway is only bound to loopback")];
+    const destruction = deferred<Record<string, never>>();
+    const refresh = deferred<{ environments: typeof environments }>();
+    let refreshPending = false;
     const fixture = mountPage(buildMethods, {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0985a13c54007aa44ae986ff19c8c2358c391095](https://github.com/openclaw/openclaw/pull/145122/commits/0985a13c54007aa44ae986ff19c8c2358c391095)
- Landed commit: [24a8205e9afb3867fe13450e7b6dade3b18182c0](https://github.com/openclaw/openclaw/commit/24a8205e9afb3867fe13450e7b6dade3b18182c0)

Issue #144294, opened 2026-09-10
[Feature]: OTEL enhancements for skill and LLM evaluations


ryan-dyer-sp:
### Summary

OpenClaw's OTel traces can't answer basic attribution questions: which agent handled a prompt, which version of a skill was read, what an LLM
was asked and answered, or which session a trace belongs to.

### Problem to solve

Four gaps in the OTel output:

1. **No way to tell which agent handled a prompt.** `openclaw.agent` is set on skill.used, model.usage, and tool spans, but not on harness.run or model.call — the spans that make up the body of a trace. If you run more than one agent, you can't group traces by agent for volume or cost, and a trace from a spawned subagent looks the same as a main-loop trace unless you correlate manually.

2. **No way to tie a behavior change to a skill edit.** Skills change often — descriptions and procedures get edited routinely, and those edits change both routing and output. When a skill starts behaving differently, the first question is whether the skill content changed, but the trace can't answer it: skill.used records only name, source, and activation type, with no identity for the content that was loaded. Comparing two executions of the same skill (before/after an edit, or skill v1 vs v2) has no join key. Skills do declare a version in their frontmatter, but nothing enforces that an edit updates it, so a declared version can't be trusted as a content identity — it has to come from the skill's actual contents.

3. **No way to see what an LLM was asked and what it answered without drilling into generations.** `openclaw.message.processed` is the root span for every prompt/response, and harness.run spans carry the turns within a run — but neither carries input or output, only channel, outcome, timing, and plugin. The prompt and the response only exist on the model.call spans underneath. Reading a trace at a glance, or building tooling on run-level facts, means reconstructing from lower-level generations.

4. **No way to filter or group traces by session.** `DROPPED_OTEL_ATTRIBUTE_KEYS` removes session, run, message, chat, and call ids from all attributes, and `redactOtelAttributes` applies it to both log records and span attributes. That's right for metrics and defensible for logs, but tracing backends filter and group on exactly these ids. Today the only options are removing keys from the shared list — which also un-drops them for logs — or renaming attributes to dodge the filter. The filter needs to be per-signal.

The root cause is the same for all four: the gateway has these facts, but the OTel export layer either never records them or filters them out, so consumers have to reconstruct what the producer already knew.

### Proposed solution

- Emit `openclaw.agent` on harness.run and model.call spans, sourced from the same agent context the other events already use.
- Add a version identifier for the loaded skill on skill.used. It must be derived from the skill's actual contents, not the frontmatter version,
[...]

PR #145132, opened 2026-09-11
Draft: OTEL Enhancements for skill and LLM evaluations

ryan-dyer-sp:
Related: #144294

**Discussion draft — not ready to merge.**
This PR exists to make the design discussion in #144294 concrete.  The current implementation and configuration shape are proposals for maintainer feedback, not settled contracts.

## What Problem This Solves

Operators using `diagnostics-otel` cannot reliably use OpenClaw traces to evaluate skill edits or model changes.

Today, traces do not consistently identify the agent that handled a prompt, the exact skill content that was read, the input/output of root and harness runs, or the session that a trace belongs to. That makes it difficult to compare before/after behavior, attribute cost or volume to an agent, and inspect a run without reconstructing it from nested model generations.

The linked issue describes the full use case: we export traces to Langfuse and evaluate skill and model changes with qualitative LLM-as-judge scoring before rolling changes out.

## Why This Change Was Made

This draft carries the four telemetry capabilities proposed in #144294:

- add canonical agent identity to diagnostic trace spans;
- add an opt-in skill-content fingerprint rather than relying on a manually maintained skill frontmatter version;
- attach bounded, policy-gated input/output to `openclaw.message.processed` and `harness.run` spans;
- separate trace, log, and metric attribute retention so high-cardinality identifiers can be available to traces without being unintentionally enabled for every signal.

It intentionally keeps content capture behind the existing capture policy and makes skill-content hashing a separate default-off diagnostics option.

Session/run/message identifiers stay excluded from export by default; an operator can retain them per-signal only by explicitly listing them in the new traceAttributes/logAttributes/metricAttributes arrays. This is opt-in scope, not a default change, and the collector-side consent for that data stays with the operator who configures it — the same trust boundary as the existing captureContent setting.

This is still a discussion draft. In particular, the configuration contract, permitted identifier scope, fingerprint ownership/bounds, and final capture boundaries need maintainer direction before this should be considered merge-ready.

## User Impact

For operators who enable `diagnostics-otel`, traces can become much more useful for analyzing multi-agent work, tracking skill changes, and comparing model or skill behavior.

Defaults remain conservative:

- prompt and response content is emitted only through the existing content-capture policy;
- skill-content hashing is default-off;
- existing attribute exclusions remain the default unless an operator explicitly configures retention for a signal.

Users who do not opt into these diagnostics capabilities should not pay for skill-content hashing.

Upgrading with default config: only openclaw.agent.id appears on harness.run and model.call spans; everything else is unchanged, including session/run/message identifiers, which stay filtered by default. Operators who already have diagnostics.otel.captureContent: true set will additionally see input.value/output.value on harness.run and message.processed spans — new span-level content under their existing capture consent, not a new flag. skillContentHash and the new attribute-retention lists are default-off/empty, so nothing else changes without explicit configuration.

## Evidence
IDK how to provide evidence that this does update spans with the desired fields.  Pictures from Langfuse of the spans and the fields?  My clanker did supply the following.  Is this sufficient?

### Live isolated-gateway validation

Built at `0b88af46014` and ran a throwaway gateway in an isolated state directory on separate ports (no operator instance involved), with a dummy
OTLP/HTTP receiver and a synthetic OpenAI-compatible model. Drove real chat runs through the gateway, including a read-activated skill.

Observed in the OTLP trace payloads received by the collector:

Corrected: an earlier revision of this evidence incorrectly listed `openclaw.agent.id` on `openclaw.message.processed`. The captured payloads show agent attribution on `openclaw.run`, `openclaw.harness.run`, `openclaw.model.usage`, `openclaw.skill.used`, and `openclaw.tool.execution`; the message root is not attributed, by design — see below.

```text
openclaw.harness.run:
  openclaw.agent.id = "main"
  input.value  = "Use the otel-proof skill and read its SKILL.md. Reply with OTEL_MESSAGE_FINAL."
  output.value = "OTEL_PROOF_FINAL: skill read and tool completed."

openclaw.message.processed:
  openclaw.channel = "webchat"
  openclaw.outcome = "completed"
  input.value / output.value = same run I/O

openclaw.skill.used:
  openclaw.skill.name = "otel-proof"
  openclaw.skill.activation = "read"
  openclaw.skill.version = "sha256:c3e58cdd19e65662"
  openclaw.agent.id = "main"

openclaw.model.usage / openclaw.run:
  openclaw.agent.id = "main"
```

Gating verified in both directions:

- After hot-reloading the isolated config to `captureContent: false`, new `openclaw.message.processed` and `openclaw.harness.run` spans kept their other attributes but had no `input.value` or `output.value`.
- With `traceAttributes` listing `openclaw.sessionId` and `openclaw.sessionKey`, trace spans carried both (for example `agent:main:otel-proof-session`); no logs or metrics payload contained session identifiers, so the default exclusions held for the unconfigured signals.
- The skill fingerprint appeared only with `diagnostics.otel.skillContentHash: true`.

### Focused local validation

```text
pnpm test:extension diagnostics-otel

Test Files  12 passed (12)
Tests       347 passed (347)
Duration    17.94s


Patrick-Erichsen:
My preferred direction is to support these evaluation use cases with the existing OTel enablement and `captureContent` boolean, without adding new flags or configurable attribute-retention lists:

| Data | Requested behavior |
| --- | --- |
| Agent identity | Include on relevant run and model-call spans by default when tracing is enabled. |
| Skill fingerprint | Include a hash identifying the exact skill content delivered to the model. Compute once per loaded version, with the identity tied to those delivered bytes; avoid hashing work when tracing is disabled. |
| Session correlation | Include a stable pseudonymous session ID derived from the existing session identity with a keyed hash. Keep raw session keys and participant details excluded. |
| Prompt + final answer | Include on message/run spans only with existing `captureContent: true`, retaining redaction and size limits. |

Keep session identifiers and skill fingerprints off metrics, and preserve the separate support-bundle redaction rules. The pseudonymous ID still permits conversation grouping; it is not anonymity.

Please document the resulting defaults and upgrade behavior, especially the additional message/run content exported for operators who already have `captureContent: true`.

ryan-dyer-sp:
> My preferred direction is to support these evaluation use cases with the existing OTel enablement and `captureContent` boolean, without adding new flags or configurable attribute-retention lists:
> 
> Data	Requested behavior
> Agent identity	Include on relevant run and model-call spans by default when tracing is enabled.
> Skill fingerprint	Include a hash identifying the exact skill content delivered to the model. Compute once per loaded version, with the identity tied to those delivered bytes; avoid hashing work when tracing is disabled.
> Session correlation	Include a stable pseudonymous session ID derived from the existing session identity with a keyed hash. Keep raw session keys and participant details excluded.
> Prompt + final answer	Include on message/run spans only with existing `captureContent: true`, retaining redaction and size limits.
> Keep session identifiers and skill fingerprints off metrics, and preserve the separate support-bundle redaction rules. The pseudonymous ID still permits conversation grouping; it is not anonymity.
> 
> Please document the resulting defaults and upgrade behavior, especially the additional message/run content exported for operators who already have `captureContent: true`.

For Skill fingerprint, the delivered bytes is not a fingerprint for the skill as a whole, its just the fingerprint for the SKILL.md.  The intent is to be able to do comparison of different iterations.  As a well architected skill can include scripts, output templates, and other reference files which are not part of the SKILL body, we need something a bit more comprehensive than the delivered bytes.  I believe the current implementation I have done does prevent hashing when otel is disabled. 

For the session correlation, do you mean use something like `agent:main:main` instead of the actual session guid?

For the other two I believe this is in agreement with what I have proposed and will separate each into their own PRs.

Thanks for the review.

PR #144768, opened 2026-09-11
fix(models): require provider-bound credentials for provider use

obviyus:
Addresses #144726/#117956: provider entries, accessible accounts or plugin-owned env vars admit model use; generic credentials do not. Saving A retains working B; fallback records its replacement with one notice. Startup projects bindings; Doctor persists the same policy.

Head: `49a24d37509152ca12b95744b3ffc26d627602b6`. Round5:101 tests and structural checks pass; census exits0; CI34704997470/1 pending. Production differs from ac6d: CI repair adds braces/renames a shadowed local; run-loop shares retry/proceed assignments; SQLite cell inspection shares row-read/missing-row handling. No API/schema/persistence change.

## Consumers

F/S keys are defined once (`F01=path`, `S01=symbol`); later uses name that same file/symbol. F01:line adds position. Use/By tables map contracts to readers/writers.

Prefixes: runner/ = src/agents/embedded-agent-runner/; methods/ = src/gateway/server-methods/; doctor-shared/ = src/commands/doctor/shared/; plan/ = src/agents/runtime-plan/; cron-agent/ = src/cron/isolated-agent/; sessions/ = src/agents/sessions/; agents/ = src/agents/; gateway/ = src/gateway/; commands/ = src/commands/; sdk/ = src/plugin-sdk/; config/ = src/config/. reply/ = src/auto-reply/reply/; profiles/ = src/agents/auth-profiles/; system-agent/ = src/system-agent/; telegram/ = extensions/telegram/src/; test-project/ = test/tsconfig/tsconfig.core.test.; plugins/ = src/plugins/; infra/ = src/infra/; flows/ = src/flows/. Other paths are repository-relative. Braces list exact alternatives.

10,284 positions not individually mapped; typed+lexical census attached (11,692 syntax candidates minus1,408 mapped, not changed APIs).

### Model/status readers

Wire: packages/gateway-protocol/src/schema/agents-models-skills.ts:ModelChoiceSchema/generated clients (available, unavailableReason, unavailableUntil, outcomes). Producers: F01=agents/model-catalog-decisions.ts, F38=methods/models-list-result.ts, F02=gateway/server-model-catalog.ts/chat metadata; typed IDs/booleans.

- Browser transport: ui/src/lib/model-catalog-store.ts, ui/src/pages/model-providers/{load,catalog-discovery}.ts, ui/src/api/types.ts.
[...]

PR #145160, opened 2026-09-11
fix: Anthropic Vertex turns fail to resolve configured models

obviyus:
Related: #144768, #145190.

## What Problem This Solves

A regional provider model could appear available with default application credentials and an empty provider declaration, yet a Gateway turn failed with “Unknown model” before making a request.

## Why This Change Was Made

The plugin registers the existing runtime model hook. It preserves an exact registry entry, otherwise resolves a supported model from its existing regional catalog. Empty endpoints are normalized before choosing the service-region endpoint, keeping model metadata and routing with the same owner.

## User Impact

Supported implicit models can serve turns with the correct regional endpoint. Authored model entries remain available. No credential-selection, configuration, schema, or migration change is introduced.

Consumers: runtime model resolution and configured-model completion now use the plugin's regional metadata when no exact row exists.

## Evidence

Registered runtime-hook tests fail before the fix. Final checks passed 91 focused tests and 148 Gateway checks through session creation, chat.send, completion, history, and models.list. Reload exposed the new default in ordinary listing without refresh or restart. All 17 qualified startup fixtures passed, as did 20 previous-release migration, install, history, and restart checks. Recorded encrypted endpoints used synthetic credentials and responses; live service acceptance and billing were not tested.

### DIFF
--- a/extensions/anthropic-vertex/index.test.ts
+++ b/extensions/anthropic-vertex/index.test.ts
@@ -1,7 +1,6 @@
-import type { OpenClawConfig, ProviderRuntimeModel } from "openclaw/plugin-sdk/plugin-entry";
 // Anthropic Vertex tests cover index plugin behavior.
 import { registerSingleProviderPlugin } from "openclaw/plugin-sdk/plugin-test-runtime";
-import { afterAll, afterEach, assert, beforeEach, describe, expect, it, vi } from "vitest";
+import { afterAll, afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const { hasAnthropicVertexAvailableAuthMock } = vi.hoisted(() => ({
   hasAnthropicVertexAvailableAuthMock: vi.fn(),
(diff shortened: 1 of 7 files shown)


obviyus:
landpr-approve-admin-merge
pr: 145160
head: 2e6e3a84dcbb6dedc3c7fe005608aee71f3c3dc9
scope: owner-review

obviyus:
Thanks @obviyus!

**Proof:** Real Gateway turns now resolve configured Vertex models. Ordinary `models.list` reflects a changed default after config reload, before explicit refresh, and preserves that result after restart. Sibling models remain available.

- **Bug:** A configured Vertex model appeared available but failed with `Unknown model` when a turn started.
- **Fix:** The existing runtime hook resolves missing models from the regional catalog while preserving configured registry rows and endpoint overrides.

| Before | After |
| --- | --- |
| Configured model failed before inference. | The Gateway resolves the model, routes the request, and saves the reply. |

Validation used isolated synthetic receivers and fake credentials. It does not claim live provider acceptance.

PR #145168, opened 2026-09-11
fix(delivery): join pending queue reads before shutdown

steipete:
## What Problem This Solves

Session delivery runtime shutdown joined active delivery drains but could return while an admitted queue lookup or pending-entry scan was still awaiting storage. Disposing the queue or its environment after that stop could race the unfinished read and its error handling.

## Why This Change Was Made

Track scheduling operations before their first read and release them after scheduling or error handling finishes. Stop still fences admission immediately, then joins this runtime's reads and delivery settlement. Existing generation checks prevent a retired read from scheduling into a replacement runtime, and old stop handles do not wait for replacement work.

## Evidence

Validation: all four delayed lookup/scan × success/failure regressions fail on the original code and pass with this change; 55 tests across runtime, recovery, storage, and real SQLite retry-persistence failure suites pass. A standalone Node smoke run with real timers and synthetic SQLite state passed the same four cases and delivered each retained row once after restart. Independent P2 review is clean. The complete changed-scope gate passed, including core and core test typechecks, lint, and all routed guards. After the patch-identical rebase onto current main, the runtime suite passed all 18 cases and the real-timer SQLite smoke passed all four cases again. The old CI failure was an unrelated missing commit-author fixture, repaired upstream in #145170 and included by this rebase.

## User Impact

This is a shutdown prerequisite for asynchronous storage. SQLite execution, transactions, schema, public APIs, and retry policy remain unchanged. Production grows by 26 lines to track previously unowned operations; tests grow by 77 lines.

### DIFF
--- a/src/infra/session-delivery-queue-runtime.test.ts
+++ b/src/infra/session-delivery-queue-runtime.test.ts
@@ -469,6 +469,83 @@ describe("session delivery queue runtime", () => {
     });
   });
 
+  it.each([
+    { mode: "lookup", failure: false },
+    { mode: "lookup", failure: true },
+    { mode: "scan", failure: false },
+    { mode: "scan", failure: true },
+  ])("joins a retired owner's delayed $mode (failure: $failure)", async ({ mode, failure }) => {
+    vi.useFakeTimers();
+    await withRuntime(async (startRuntime) => {
+      const id = await enqueueSessionDelivery({
+        kind: "systemEvent",
+        sessionKey: "agent:main:main",
+        text: "delivery awaiting queue lookup",
+      });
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [3788817c1d42673169e078d46014eed05f169172](https://github.com/openclaw/openclaw/pull/145168/commits/3788817c1d42673169e078d46014eed05f169172)
- Landed commit: [3ea8dcb49bc87af49db3223b68d12c9dabb5b4ed](https://github.com/openclaw/openclaw/commit/3ea8dcb49bc87af49db3223b68d12c9dabb5b4ed)

PR #145128, opened 2026-09-11
fix(scripts): drop unverified local git identities from squash co-author credit

steipete:
Refs #145094

<details>
<summary>Additional instructions</summary>

Maintainers can update this same-repository branch.

</details>

## What Problem This Solves

Resolves a problem where maintainers landing a PR could include a contributor's unlinked local machine identity as a squash co-author. PR #145094 credited its human author correctly but also carried the local git identity into the landed commit.

## Why This Change Was Made

The squash composer now qualifies commit-derived credit using GitHub's linked author for each published non-merge commit. It drops reserved/local-domain identities, unlinked authors, and unresolved noreply identities, while retaining linked humans and the narrow unlinked-name match to the human PR author's login. Existing verified bot exclusions remain in place. The merge-outcome and Crabbox authorization fixtures now model the commit-author API response; their authorization and attribution assertions remain intact.

The composer filters matching preview/source trailers and reports removed credit in one preparation-log line. PR-author credit, reviewed-body byte preservation, and merge-queue admission rules retain their existing ownership. No update runtime behavior changes.

## User Impact

Squash messages preserve human contributor credit without adding local machine identities. No migration or configuration change is needed; this affects the maintainer landing wrapper only.

## Evidence

- `node scripts/run-vitest.mjs test/scripts/pr-merge.test.ts -t '#145094 shape'` failed before the production fix: the actual body contained an extra `Co-authored-by: local-agent <local-agent@openclaw.local>` trailer.
- `node scripts/run-vitest.mjs test/scripts/pr-merge.test.ts`: **121 passed**, including the regression, linked human/public noreply addresses, all requested reserved domains, unlinked authors, the PR-login exception, and existing verified-bot cases.
- The existing signed-commit test caught signature-display text entering the new commit-ID read. Disabling Git presentation output for that read fixed it; the full file passed afterward.
- Read-only replay of the published squash message from `6c4fe0c1157`, using GitHub author evidence for `79fb949be8`, removed its local co-author trailer and preserved `Thanks @eleqtrizit.`
- `node scripts/check-changed.mjs`: passed (formatting, root-test types, tooling/test lint, dependency and dead-export guards).
- `node scripts/run-vitest.mjs test/scripts/pr-crabbox-merge-bypass.test.ts`: **38 passed**.
[...]

PR #145170, opened 2026-09-11
fix: archive sessions with provisioning cloud workers

steipete:
## What Problem This Solves

Fixes an issue where users archiving a session would be told to wait indefinitely when its cloud worker placement was stuck provisioning. Failed placements also required a separate Stop action before archive could retry cleanup.

## Why This Change Was Made

Archive and delete now use the existing worker Stop owner to cancel dispatch and retry failed teardown. Startup cancellation is captured before the session drain can change its placement; running workers retain drain-before-reconciliation ordering. Source checks follow only transitions recorded by the captured dispatch, so a replacement worker cannot be mistaken for the original owner.

The archive commits only after confirmed worker cleanup. Active workers still reconcile their workspace first, and cleanup errors retain the session and recovery state. An already-running workspace reconciliation remains protected. Existing persisted placements use the same cleanup path without a config or storage migration.

## User Impact

Archiving stops the cloud worker automatically, including provisioning left behind by an interrupted dispatch. If provider cleanup fails, retrying archive retries cleanup without discarding unsynced work.

## Evidence

- Added a Gateway `sessions.patch` regression using the real placement store and Stop service. It failed before the fix; it now proves archive waits for confirmed destruction and retains the session on provider failure.
- Extended the existing cancellation matrix to cover archive during dispatch transitions and rejection of replacement ownership.
- Focused archive, deletion, reclaim, recovery, and idle-stop tests; final archive/recovery rerun: 51 passed, cancellation matrix: 9 passed. Production/test types, changed-file lint and repository guards passed; independent reviews found no actionable P0–P2 issues.
- CI follow-up removes a type-only import cycle. It also adds nine lines to the shared merge-test mock for the author lookup introduced by #145128; both unrelated CI failures were reproduced on current main and pass with the fixture repair. No production merge-tooling changes.
- Live browser verification was unavailable because the Chrome relay reported an ownership conflict. No UI appearance changes or live Gateway deployment are included.


steipete:
Merged via squash.

- Prepared head SHA: [05c668ca3e721a10ba9f5e102fed36eead478ab9](https://github.com/openclaw/openclaw/pull/145170/commits/05c668ca3e721a10ba9f5e102fed36eead478ab9)
- Landed commit: [03234b50810910cd0075a5a325073f28123d51d3](https://github.com/openclaw/openclaw/commit/03234b50810910cd0075a5a325073f28123d51d3)

PR #145173, opened 2026-09-11
fix(matrix): preserve native relations in recovered receipts

steipete:
## What Problem This Solves

Fixes an issue where Matrix delivery recovery would label caption-overflow text as replying to the original message even though that native event had no reply relation. The media event and its follow-up text could therefore have correct platform content but inaccurate returned delivery evidence.

## Why This Change Was Made

Normal sends and reconciliation now share the Matrix receipt builder. Recovery reads each saved event's native reply relation and thread placement, while preserving event ordering, the primary ID, and first-occurrence ID deduplication. This removes 36 production lines and the parallel request-based receipt builder.

## User Impact

Recovered receipts accurately distinguish a media reply from its ordinary caption-overflow text. Clients consuming those receipts receive the same event-specific metadata as normal sends.

## Evidence

A regression exercises an ordinary durable media send, simulates a lost overflow response, and reconciles through the actual send/recovery owners. It failed before the fix on the invented overflow reply ID. The replay retains the original event content and transaction IDs and uploads the media once.

All 192 selected owner and sibling tests pass, including repeated event IDs, partial delivery, and adapter receipt forwarding. Tests use mocked Matrix transport/media and isolated state. Canonical checks, the full build, and both independent managed review passes also pass.

### DIFF
--- a/extensions/matrix/src/matrix/delivery-plan.test.ts
+++ b/extensions/matrix/src/matrix/delivery-plan.test.ts
@@ -213,64 +213,84 @@ describe("Matrix durable delivery plans", () => {
     });
   });
 
-  it("preserves ordered typed receipt parts and the final event identity", async () => {
-    const deliveryIdentity = identity("queue-multi-event");
-    const plannedEvents = createMatrixPlannedEvents({
-      identity: deliveryIdentity,
-      events: [
-        { receiptKind: "media", content: { msgtype: "m.image", body: "caption" } },
-        { receiptKind: "text", content: { msgtype: "m.text", body: "follow-up" } },
-      ],
-    });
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [de5150fbe0b1f910badaf2a37b29d16a565daad6](https://github.com/openclaw/openclaw/pull/145173/commits/de5150fbe0b1f910badaf2a37b29d16a565daad6)
- Landed commit: [f770fd99e0ce4c057630172472e4df1c878b551a](https://github.com/openclaw/openclaw/commit/f770fd99e0ce4c057630172472e4df1c878b551a)

PR #131617, opened 2026-08-28
fix: recognize newly installed skill and Gmail dependencies

steipete:
Related: https://github.com/openclaw/openclaw/issues/21094 — successful skill installs remain marked as missing.
Follow-up to https://github.com/openclaw/openclaw/pull/21116 — the earlier positive-only cache fix was closed unmerged for inactivity and does not allow maintainer edits.

<details>
<summary>Additional instructions</summary>

Maintainer edits are enabled through the same-repository branch.

</details>

## What Problem This Solves

Fixes an issue where users setting up Gmail would get “still not available after brew install” even though installation succeeded and the executable was present. The same stale result kept newly installed skill dependencies ineligible and caused repeated Go/uv prerequisite bootstraps within one process when PATH did not change.

## Why This Change Was Made

The shared `hasBinary()` owner cached both hits and misses until PATH/PATHEXT changed. Installing a file into an existing PATH directory changes neither value, so the next check reused the stale miss. Cache only successful probes, using a `Set` instead of a boolean `Map`; retain successful-lookup reuse, PATH/PATHEXT invalidation, and existing platform search semantics.

This fixes all callers at the probe owner rather than adding invalidation calls to individual installers. Installer-only invalidation would miss external installs and prerequisite success followed by recipe failure; a TTL would merely delay visibility. Existing session skill snapshots and the separate executable-path idle-TTL resolver are intentionally unchanged. Gmail diagnostics formatting is not part of this change.

The cache was introduced in https://github.com/openclaw/openclaw/commit/76e4e9d17653d8ce296edff992ce65d2adbe4b38 and is reachable from stable `v2026.2.14`. The affected production owners remain unchanged on current `main` compared with the reproduced baseline.

## User Impact

Fresh dependency checks recognize binaries installed into directories already on PATH without requiring a process restart or a PATH edit. Gmail dependency setup no longer reports a false installation failure, and later skill installs reuse successfully installed Go/uv prerequisites even if an earlier final recipe failed.

Missing probes now intentionally perform filesystem work again; successful probes remain cached. No new configuration, dependency, refresh API, or migration is introduced. Production LOC: +4/-6 (net -2). Tests: +261/-15 (net +246). Documentation: +4.

Thanks to @heyrtl for the earlier positive-only cache proposal and @DannyDesert for the original skill-install report. AI-assisted implementation and verification; no agent transcripts attached.

## Evidence

[...]

PR #145175, opened 2026-09-11
perf(skills): avoid errors for missing Unix binary candidates

### LINKED ISSUE #145174: Reduce missing-command lookup overhead without caching misses
POSIX skill and hook eligibility checks repeatedly create and discard filesystem errors while scanning PATH for missing commands. Misses intentionally remain uncached so a command installed into an existing PATH directory is visible on the next probe, as established by #131617.

Reject absent POSIX candidates with an existence check inside the existing error boundary, then retain the existing X_OK check for executable/search permission. Preserve the positive-only cache, PATH/PATHEXT invalidation, directory and symlink behavior, permission-denial handling, and remote binary fallback. Windows must keep its existing lookup and suffix-enumeration path.

This trades error construction on missing paths for an additional filesystem check on existing uncached candidates. Native macOS fixtures show lower sampled allocation on missing-command workloads but slower cold successes and existing nonexecutable checks. This is not a request to cache misses or a claim of universal RSS or Gateway improvement.

steipete:
Closes #145174

<details>

</details>

## What Problem This Solves

POSIX skill and hook eligibility checks do repeated work when required commands are absent from PATH. Missing results must remain fresh so newly installed commands become available without changing PATH or restarting the process.

## Why This Change Was Made

Use `existsSync` to reject absent POSIX candidates inside the existing catch boundary before the unchanged `X_OK` check. Successful probes alone remain cached, preserving the installation contract from #131617. Windows keeps its existing lookup and PATHEXT path because its existence API has different symlink behavior.

Native fixtures replace mocked nonexistent-path cache/PATHEXT tests. Permission handling, directory and symlink behavior, cache invalidation and remote fallback retain their existing contracts; no new cache or public API is added.

## User Impact

Missing-command eligibility checks avoid discarded filesystem errors. Existing uncached POSIX candidates receive one additional filesystem check, so this is a tradeoff rather than a universal speed improvement, especially on slow or network filesystems. Installed-command visibility and eligibility results remain unchanged.

## Evidence

The same **86 tests across five owner and sibling suites** pass before and after, along with the full selected changed gate and independent P2 review. Ten native output scenarios remain equal, including unchanged-PATH installation, positive-cache invalidation, links, nonexecutable paths and remote fallback.

Six fresh macOS processes, three per version, exercised the actual skill/hook eligibility owners with ten PATH directories and 1,000 warmup calls per scenario. Timing and allocation sampling ran separately. For 10,000 mixed calls with 60% missing requirements, median sampled allocation fell from **162,829,840 to 94,914,832 bytes (41.71%)**, and elapsed time from **546.09 to 104.54 ms**. For 5,000 missing-command calls, allocation fell from **123,482,872 to 69,565,112 bytes (43.66%)**, and time from **432.73 to 75.74 ms**. Sampling includes collected objects.

Tradeoffs are retained: 5,000 cold first-directory hits rose from **33.46 to 38.31 ms (+14.49%)**, and existing nonexecutable checks from **72.88 to 81.31 ms (+11.56%)**. Twenty thousand warm cached hits were **10.04 to 10.44 ms**; no gain is claimed there.

The initial direct-`tsx` permission preflight failed in the loader's case-sensitivity check before evaluating the owner. That failure is retained. Paired Node 26.8.1 permission checks instead imported bundles of the exact before/after owner, with source hashes and bundler metafiles retained; both returned false for denied paths and found the permitted sibling without worker or child-process grants.

These are fixed synthetic macOS workloads, not native Windows, retained-heap, RSS or general Gateway performance evidence. Reusable skill snapshots can bypass these eligibility calls. Correctness-suite durations of 10.33 and 9.23 seconds are descriptive only. No normal build was selected because package, import, lazy-loading and artifact contracts are unchanged.


steipete:
Maintainer decision: accept the documented, bounded filesystem tradeoff. The existing shared probe keeps positive-only caching, immediate discovery after installation, Windows behavior and executable-permission checks. The measured missing-heavy workloads reduce sampled allocation by 41.71%–43.66%; cold first-directory hits cost 14.49% more (about 0.97 microseconds per call) and nonexecutable candidates cost 11.56% more. These results are limited to the measured native workloads; there is no claim of a universal speedup or improvement on slow/network filesystems.

The two reviewed files are unchanged. The refreshed head `9c0ee05c00f73455a0063b1314f3cefe9c85fc82` includes the upstream native-fixture repair, and required CI run 34644289883 is green after its unacquired-runner shard was retried once. The original failed and cancelled runs remain recorded. No source, dependency, assertion, timeout or gate was weakened.

steipete:
Merged via squash.

- Prepared head SHA: [9c0ee05c00f73455a0063b1314f3cefe9c85fc82](https://github.com/openclaw/openclaw/pull/145175/commits/9c0ee05c00f73455a0063b1314f3cefe9c85fc82)
- Landed commit: [6a971911e062ceb3f36d2029f0a9745490f1e93d](https://github.com/openclaw/openclaw/commit/6a971911e062ceb3f36d2029f0a9745490f1e93d)

PR #145196, opened 2026-09-11
fix(auth): await durable OAuth refresh transitions

steipete:
## What Problem This Solves

Asynchronous auth storage needs the refresh protocol to await durable claims and settlements. Runtime publication must also retain the current credential owner and the latest unrelated-provider state when an awaited operation finishes.

## Why This Change Was Made

Await backend reads, claims, success writes, failure fencing, and observer snapshots while retaining synchronous transaction callbacks. Before publishing an AuthStorage refresh snapshot, the existing synchronous backend rereads authoritative state, verifies the provider entry, and publishes the fresh complete map. Failed initial publication runs the existing terminal cleanup before returning.

Observation reads retain their existing absolute deadline. Native Promise settlement preserves the exact original fulfillment or rejection before that deadline, including primitive rejection reasons, without a lint suppression. Observer timeout still stops waiting without abandoning an already-owned durable settlement; this does not extend the provider timeout to initial database admission.

## User Impact

Credential publication follows durable persistence and current ownership. Completed logout and account replacement remain reflected in runtime state, unrelated-provider updates survive publication, and failed publication leaves the refresh claim settled. SQLite schema, credential encoding, namespace, provider timeout scope, and public SDK contracts are unchanged.

The AuthStorage compatibility backend and its final authoritative publication read remain synchronous. This repairs the awaited protocol handoff; it does not move all auth SQLite execution into workers.

## Evidence

The final candidate has 56 passing cases across the async fence, generation fence, AuthStorage refresh, and AuthStorage publication suites. These use isolated synthetic state and native SQLite through the production AuthStorage entry points, with synthetic provider credentials and callbacks. They cover normal refresh, observer completion, current facade/persisted state, unrelated-provider preservation, failed publication cleanup, and rejection identity. Source hashes were captured before and after the run and remained identical. The full changed-code gate also passed on the refreshed main composition with identical before/after source hashes, including core/test types, lint, all Knip scans, storage guards, and zero runtime import cycles.

The original suppression repair separately passed 44 behavior cases plus all four cases in the actual CI suppression owner, with its allowlist unchanged. The existing native SQLite ordering/reopen proof remains applicable to that prerequisite. The final production repair and tests are reviewed together in the index and working source; a fresh isolated Codex review through P2 reports no actionable findings.

No real credentials, external provider calls, Gateway restarts, or schema migrations were used. This is native storage/lifecycle boundary proof, not a claim of an external provider round trip or complete auth worker migration. Exact-head hosted CI and native preparation remain required for landing.

Validated head: `0c5208f694e6de7786b49a1cf1caf9906d311fa2`, rebased patch-identically onto `47ce03dbc6994435af783fecf13d5586e7b307f4`, which includes the shared UI repairs from #145415. All four authored file hashes and the three-commit range-diff are unchanged. All 56 affected cases and the full changed-code gate pass on this composition; exact-head hosted CI remains required for landing.

### DIFF
--- a/src/agents/auth-profiles/oauth-refresh-fence.async.test.ts
+++ b/src/agents/auth-profiles/oauth-refresh-fence.async.test.ts
@@ -7,6 +7,7 @@ import { closeOpenClawStateDatabaseForTest } from "../../state/openclaw-state-db
 import { withEnvAsync } from "../../test-utils/env.js";
 import {
   observeOAuthRefreshFenceSettlement,
+  observeOAuthRefreshSettlement,
   refreshSerializedOAuthCredential,
 } from "./oauth-refresh-fence.js";
 import { isPendingOAuthRefreshFence } from "./oauth-refresh-marker.js";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0c5208f694e6de7786b49a1cf1caf9906d311fa2](https://github.com/openclaw/openclaw/pull/145196/commits/0c5208f694e6de7786b49a1cf1caf9906d311fa2)
- Landed commit: [04aac5be5f158f9cf0d19fa5700154d47e071ba1](https://github.com/openclaw/openclaw/commit/04aac5be5f158f9cf0d19fa5700154d47e071ba1)

PR #145046, opened 2026-09-11
fix(sqlite): retain Bun worker library and handle ownership

steipete:
Related: #144592

## What Problem This Solves

On supported Bun, a second SQLite worker or a restarted store repeats the process-wide native library selector and fails. Bun also retains native statements after database close, so sharing a worker between databases can leave a closed store's file handles and SQLite lock alive.

## Why This Change Was Made

Keep selection in the existing library owner and pass its completed fact to new workers. Selection policy, capability checks and failure memoization stay intact.

On Bun, use one worker per distinct database within the existing four-worker cap. Alias clients share the same actor; final close retains pathname ownership until worker exit. A fifth distinct database is refused explicitly without disturbing the existing stores. Node retains shared slots.

## User Impact

Bun worker stores can start after main-process selection, open concurrently, close and reopen without repeated initialization or leaked native ownership. Bun temporarily supports four distinct open worker databases; users needing more can use Node. The limit can be revisited after the upstream native-close fix ships and release tests prove it safe.

## Evidence

- 82 selected Node tests passed, with six affected cases rerun after a test-only lint cleanup. Full selected checks and fresh independent Codex P2 review passed.
- Actual Bun 1.4.2 reproduced repeated library selection failure and native-handle growth across 20 shared-slot close/reopen cycles.
- With dedicated slots, four stores remain healthy after capacity refusal. Alias clients preserve ownership; final close leaves zero descriptors while three peers stay live. Independent SQLite writes, checkpoints and exclusive journal-mode changes succeed afterward. Every one of 20 reopen/close cycles leaves zero descriptors.
- Reversing only retirement-before-forgetting ownership makes the close-order regression fail.
- A separately pinned compiled union of the Logbook worker candidate and these exact repair files passed a full build and actual Bun Logbook/Team Reports startup, concurrent use, restart and fresh-process persistence checks. Source and artifact hashes stayed unchanged; all created resources joined.

The compiled union is recorded separately from later mechanical PR rebases; no full rebuild of those later graphs is claimed. Bun's pre-existing tsconfig diagnostic was reproduced without OpenClaw and retained in the proof. No dependency patch or version change is included.

PR #145198, opened 2026-09-11
fix: prepare Bun SQLite before Vitest workers

steipete:
## Problem

OpenClaw now publishes Bun's validated, process-wide SQLite library choice through worker environment data. Vitest did not select that library before creating its worker pool, so the first test worker selected it too late for child storage workers. Those workers then repeated Bun's one-shot SQLite initialization and failed with `SQLite already loaded`. A worker-thread mock also hid `isMainThread` before the selector was initialized.

## Solution

Select SQLite in the Vitest config host before workers start, then preload the inherited choice in each test worker before test modules and mocks. Keep the selector usable during config loading by validating the inherited record without importing a workspace package that Vite cannot resolve that early.

Two worker-reuse tests now remain Node-only because Bun deliberately uses one worker per distinct database and caps that pool at four. The Bun-specific coverage added in #145046 continues to enforce that documented limit and worker cleanup behavior.

The bootstrap carries a removal note for oven-sh/bun#42349, which makes repeated selection of the same custom library idempotent.

## Validation

- Custom Bun `1.4.3-canary.1+13bd48ed7`: 195 passed, 2 documented Bun-cap skips across selector, generic SQLite store, heartbeat, transcript archive/reconciliation, Team Reports, and Memory lifecycle tests.
- Node: 53 passed across the same selector, generic store, and heartbeat boundaries; both Node worker-reuse cases executed.
- Node minimal-repository Vitest configuration and cleanup fixtures: 76 passed.
- `pnpm check:changed`
- Independent P0-P2 autoreview: clean

### DIFF
--- a/src/infra/bun-sqlite-library.ts
+++ b/src/infra/bun-sqlite-library.ts
@@ -2,7 +2,6 @@ import { existsSync } from "node:fs";
 import { createRequire } from "node:module";
 import { posix } from "node:path";
 import { getEnvironmentData, isMainThread, setEnvironmentData } from "node:worker_threads";
-import { isRecord } from "@openclaw/normalization-core/record-coerce";
 import { resolveGlobalSingleton } from "../shared/global-singleton.js";
 import { isSqliteWalResetSafeVersion } from "./sqlite-runtime-version.js";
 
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6e3f6e9b6c86890b5dc50813d94cc6517c0852ab](https://github.com/openclaw/openclaw/pull/145198/commits/6e3f6e9b6c86890b5dc50813d94cc6517c0852ab)
- Landed commit: [2f48f72418570ee138fac223536e41c680aab56f](https://github.com/openclaw/openclaw/commit/2f48f72418570ee138fac223536e41c680aab56f)

PR #145147, opened 2026-09-11
fix(webchat): allow ACP binding in dashboard conversations

fuller-stack-dev:
Related: #142770, WebChat ACP half only. The separate rollback acceptance remains open. Reported by @cjn119-ui.

## What Problem This Solves

`/acp spawn claude --bind here` rejected internal dashboard sessions as not being active WebChat conversations. Fixing the missing target exposed source/target ownership bugs in reply persistence, completion, and escaped local commands.

## Why This Change Was Made

- Internal `chat.send` supplies the existing canonical WebChat session key as its conversation target.
- Terminal observation follows the chat-link owner. A runtime transcript suppresses the source append only when agent, session key, and physical session ID match.
- Escaped ACP commands initialize the source session; their existing handlers resolve the bound target separately.

The existing binding service and lifecycle/transcript writers remain the owners. Explicit originating routes, external-delivery isolation, and plugin-owned bindings are preserved. No schema, migration, deployment, credential, or rollback changes.

## User Impact

Existing dashboard sessions can bind without recreation. Bound replies persist in dashboard history and settle its run state. Existing bindings survive reopening; status and conversation-scoped unbind remain usable. This prevents new failures and permits retrying existing sessions, but does not backfill replies lost by earlier failed runs.

## Evidence

- Original unmodified main `3ac42154d14a98ab2393e68bb3285df77c512b14`: real Control UI `chat.send`, internal delivery, zero conversation rows, no fabricated origin fields. Visible transcript and final Gateway event reproduced the exact active-WebChat error.
- Shipped 2026.9.3 source is `1391f7cd2d40ab5bbcf2f5f831d3a64f520e72d7`. All four gaps are present there. Prior fixes #106927, #133461, and #140869 are included in that release and candidate base `24a8205e9afb3867fe13450e7b6dade3b18182c0`; they establish the reused owners but do not fix these paths. Scoped source is unchanged through inspected main `93ad954b7b9af8565941408b48aed7ecab0e31d3`.
- Regression coverage includes delivery modes, real binding-store reopen, agent/session/account isolation, independent unbinding, source versus target lifecycle ownership, transcript deduplication, and escaped status/unbind/reset commands. The reported defects failed before repair. Existing plugin, reset, routing, lifecycle, and command siblings were exercised.
- Fresh full-candidate autoreview returned no blocking findings. Native exact-head review tests passed.

[...]

PR #145199, opened 2026-09-11
fix(tooling): restore supervised merge fixture after author lookup

Patrick-Erichsen:
## What Problem This Solves

The supervised merge regression test fails on current main because its fake GitHub API does not handle the commit-author lookup introduced in #145128. This blocks unrelated PR CI, including #145183

## Why This Change Was Made

Update the main-refresh fixture to return commit author metadata from its own Git repository and include the PR author's GraphQL type. Unlinked synthetic commit authors remain `null`, matching GitHub's response; production author verification is unchanged

## User Impact

Restores the merge-tool regression check without changing production behavior

## Evidence

On baseline `1d30025c63f`, the supervised merge test fails with `Unexpected GitHub API endpoint repos/fixture/repo/commits/<sha>`. With the fixture repair, the same test passes. All 49 main-refresh tests and all 17 hosted-merge tests pass. These are the only two consumers of the shared fixture. Focused lint, formatting, and scoped autoreview pass

The original CI failure is [checks-node-compact-small-28](https://github.com/openclaw/openclaw/actions/runs/34637193392/job/103388057313). The fixture and production API caller were unchanged by the UI PR

### DIFF
--- a/test/scripts/pr-main-refresh.test-support.ts
+++ b/test/scripts/pr-main-refresh.test-support.ts
@@ -136,7 +136,7 @@ export function createMainRefreshFixture(directory: string) {
     isMergeQueueEnabled: false,
     isDraft: false,
     isCrossRepository: false,
-    author: { login: "fixture" },
+    author: { login: "fixture", __typename: "User" },
     baseRefName: "main",
     baseRefOid: main,
     headRefName: "topic",
@@ -397,6 +397,7 @@ if (args[0] === 'pr' && args[1] === 'view') {
     } else if (args.some(arg => arg.includes('viewerMergeBodyText'))) {
       value = { data: { repository: { pullRequest: {
         headRefOid: control.metadata.headRefOid,
+        author: control.metadata.author,
         isMergeQueueEnabled: control.metadata.isMergeQueueEnabled,
         viewerMergeBodyText: 'Reviewed fixture body',
       } } } };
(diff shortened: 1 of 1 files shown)


fuller-stack-dev:
@Patrick-Erichsen I am landing #145147, which is also blocked by this fixture defect. Its exact-head run fails both the supervised-merge case and the hosted-merge sibling at the same missing `commits/<sha>` endpoint. The supervised case also fails on unchanged main `202696784ec12d84f5b9958af15cb62419d206dc`.

This is the canonical repair for our dependency. I will not duplicate or modify your patch. Are you taking #145199 through its final gated merge, or is another maintainer already handling that step? I will retain #145147 and sync it after this repair is included on main, then requalify its exact head and use the native landing gates.

Patrick-Erichsen:
The fixture contract repair is now on main via #145170 (`03234b50810910cd0075a5a325073f28123d51d3`). I verified that commit supplies both missing responses: commit author metadata with a nullable linked account, and the GraphQL PR author type

Closing this redundant repair. Its own CI and all 66 local tests across the two fixture consumers passed. The UI work in #145183 will refresh onto the repaired baseline

steipete:
Closing as superseded: main was repaired by #145170 (merge 03234b50810910cd0075a5a325073f28123d51d3), which added the commit-endpoint handling to the shared `pr-main-refresh` test support after #145128, so both affected tooling suites are green on main again. Thanks for the quick fix, Patrick; nothing further needed here.

PR #145206, opened 2026-09-11
test(pr): model commit author lookup in merge fixtures

steipete:
## What Problem This Solves

Main now resolves commit authors during native PR landing, but the shared synthetic GitHub fixture rejects that new API call. Two existing merge-workflow tests fail before they reach their assertions; this also blocks unrelated PR CI.

## Why This Change Was Made

Teach the existing fixture the exact commit-author endpoint and the PR author shape returned by the landing metadata query. Read author name/email from the synthetic origin commit instead of inventing attribution. Production landing code and existing assertions remain unchanged.

## User Impact

Restores the merge-workflow regression tests and removes an inherited CI failure affecting unrelated changes. The missing fixture response followed #145128; it was reproduced on current main independently of the delivery changes in #145168.

## Evidence

Both reported cases fail on unchanged main with the unsupported commit endpoint. All 66 tests across the two complete consumer files pass after the ten-line fixture repair. Full changed checks, whitespace checks, and independent review through P2 pass. No checks or production behavior were weakened.

### DIFF
--- a/test/scripts/pr-main-refresh.test-support.ts
+++ b/test/scripts/pr-main-refresh.test-support.ts
@@ -397,6 +397,7 @@ if (args[0] === 'pr' && args[1] === 'view') {
     } else if (args.some(arg => arg.includes('viewerMergeBodyText'))) {
       value = { data: { repository: { pullRequest: {
         headRefOid: control.metadata.headRefOid,
+        author: { login: control.metadata.author.login, __typename: 'User' },
         isMergeQueueEnabled: control.metadata.isMergeQueueEnabled,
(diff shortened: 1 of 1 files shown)


steipete:
Closing as superseded by #145170, merged as 03234b50810910cd0075a5a325073f28123d51d3. Its shared fixture now supplies the commit-author and PR-author responses that caused both CI failures. The tested upstream head, merge commit, and observed main contain the identical fixture, and CI run 34637809505 is green.

Our broader any-commit/linked-user fixture variant is unnecessary for this repair; no production change or additional fixture contract is needed. The affected database PRs will be refreshed onto the upstream fix.

PR #145212, opened 2026-09-11
perf: reduce cold imports during provider discovery

steipete:
## What Problem This Solves

Provider discovery paid the cost of loading profile stores and web-search execution just to read environment values and represent managed credentials. In the measured first-run migration case, loading the xAI descriptor alone took 49–56 seconds.

## Why This Change Was Made

Move the existing credential-value functions into one dependency-light owner and expose them through the existing `secret-input` SDK entrypoint. xAI discovery and Cloudflare catalog metadata use that narrow surface; xAI config lookup uses the existing config-contract entrypoint. Internal consumers move to the canonical owner, while existing public SDK exports remain available.

Function bodies, credential precedence, blocked SecretRefs, synthetic auth hooks and stored marker values are unchanged. The two required narrow exports increase the public/callable SDK budgets by exactly two each; all other budgets stay unchanged. No new runtime configuration, cache, dependency, deadline, or shard is introduced.

## User Impact

Cold provider discovery does less unrelated module loading. This also shortens the migration test without replacing real provider/auth behavior with mocks. Compatibility normalization remains a separate measured hotspot.

## Evidence

Same-main, same-case local A/B/A on base `58a6953549d011ffe44dcc3a7f3290b50e64b349`, retaining existing caches:

| Measurement | Original A1 | Candidate B | Original A2 |
| --- | ---: | ---: | ---: |
| Selected migration case | 87.55s | 47.09s | 102.74s |
| Whole invocation | 118.89s | 101.99s | 127.47s |
| xAI descriptor loading | 49.14s | 0.159s | 55.93s |

All arms passed with the same case digest, five configured models, three generated catalogs, and both synthetic auth hooks. Unchanged normalization took 41.92s in B and 41.75s in A2. Invocation startup varied, so these results do not establish an overall CI reduction.

Validation: 428 focused semantic/sibling cases passed; 41 revised marker/xAI/SDK cases and two selected SDK budget guards passed. Current-source `ciArtifacts` build passed, including public SDK declarations/export validation. Typechecks, lint, boundary and cycle checks passed. Independent P2 review completed all three passes with no findings.

Local validation is composite: an earlier full build passed before removal of an unused export alias, and the initial changed gate correctly failed at the SDK budget. After the exact budget follow-through, all 28 failed/unrun check components passed. The subsequent exact-head hosted full build passed on `5998ee08a269`. Its only two failing jobs used a shared tooling fixture that lacked the new commit-author lookup response. Main `03234b508109` already fixed that fixture; the identical optimization patch was rebased to include the fix. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34643168858) now passes on `9ff5cd7a1e0ee7b9848687dbd75902f343f72b09`: all 190 checks are terminal without failures, including the full build.

Production delta: +46/-45 (net +1), tooling +4/-2, tests +52/-31, documentation +7. The production change moves existing functions and rewires imports; the small growth provides the canonical lightweight boundary. Persisted SecretRef marker tests independently pin file/exec/store wire values. No changelog edit is needed before release generation.

Maintainer compatibility decision: accept and support the two documented `secret-input` export locations. They expose the same canonical functions as the existing public paths, which remain supported. The exact +2 public/+2 callable budgets account for that deliberate additive contract.

The requested RSS follow-through uses a separate post-build matched pair on the same `5998ee08a269` checkout, with existing built artifacts and caches retained. The original control reverses only this PR’s 36 paths; both arms execute the same selected case with 37 trace records and unchanged model/catalog/auth-hook results. `/usr/bin/time -l` maximum resident set size: original 2,320,531,456bytes (2213.03MiB), candidate 2,298,658,816bytes (2192.17MiB). That 0.94% difference is too small to claim a strong memory improvement from one pair. This is the invocation’s reported maximum RSS, not simultaneous summed process-tree memory.

In that post-build pair the case took 61.22s original versus 15.28s candidate, and the invocation 72.82s versus 24.66s. Those timings have a different built-artifact state from the historical A/B/A above and are not pooled with it. Telegram doctor loading remained about 8s in both arms and is a separate follow-up.

### DIFF
--- a/docs/plugins/sdk-overview/imports.md
+++ b/docs/plugins/sdk-overview/imports.md
@@ -60,6 +60,13 @@ do. Do not copy those import paths into new plugins; use injected runtime helper
 and generic channel SDK subpaths instead.
 </Warning>
 
+For provider discovery that only needs credential values, use
+`openclaw/plugin-sdk/secret-input` for `readProviderEnvValue`,
+`resolveNonEnvSecretRefApiKeyMarker`, and SecretRef coercion/normalization.
+These helpers do not load profile stores, provider transports, or web-search
+execution. Use `provider-web-search-config-contract` to read plugin-owned
+web-search config. Keep full auth and search runtime imports in execution paths.
+
 ## Subpath reference
 
 The plugin SDK is exposed as a set of narrow subpaths grouped by area (plugin
--- a/extensions/cloudflare-ai-gateway/catalog-provider.ts
+++ b/extensions/cloudflare-ai-gateway/catalog-provider.ts
@@ -5,7 +5,7 @@
 import {
   coerceSecretRef,
   resolveNonEnvSecretRefApiKeyMarker,
-} from "openclaw/plugin-sdk/provider-auth";
+} from "openclaw/plugin-sdk/secret-input";
 import { normalizeOptionalString } from "openclaw/plugin-sdk/string-coerce-runtime";
 import {
   buildCloudflareAiGatewayModelDefinition,
--- a/extensions/xai/provider-discovery.ts
+++ b/extensions/xai/provider-discovery.ts
@@ -1,6 +1,6 @@
 // Xai provider module implements model/runtime integration.
 import type { ProviderPlugin } from "openclaw/plugin-sdk/provider-model-shared";
-import { readProviderEnvValue } from "openclaw/plugin-sdk/provider-web-search";
+import { readProviderEnvValue } from "openclaw/plugin-sdk/secret-input";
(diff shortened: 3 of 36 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9ff5cd7a1e0ee7b9848687dbd75902f343f72b09](https://github.com/openclaw/openclaw/pull/145212/commits/9ff5cd7a1e0ee7b9848687dbd75902f343f72b09)
- Landed commit: [d781c156cb7738ce64eba162178bef85e715a989](https://github.com/openclaw/openclaw/commit/d781c156cb7738ce64eba162178bef85e715a989)

Issue #137377, opened 2026-09-03
Doctor --fix on Windows (2026.8.2) always fails at its final restart: "Gateway service ownership or manager identity changed"; Scheduled Task left disabled


TommyLei666:
### Body

**Environment**
- OS: Windows 11 (x64)
- Node: 24.18.0 (from OpenClaw logs)
- OpenClaw: 2026.8.2 (0965053), installed via `npm install -g openclaw` (global), run from an elevated PowerShell
- Upgrade path: 2026.7.1-2 → 2026.8.2 (first doctor run performed the state v1→v15, agent-db v1→v19, and config migrations successfully)
- Config: multi-agent (`agents.entries`: main, coding, local_agent), default profile, canonical state dir `~/.openclaw`
- Gateway service: Windows Scheduled Task "OpenClaw Gateway" (schtasks), wrapper `~/.openclaw/gateway.vbs` → `~/.openclaw/gateway.cmd` → `node ...\dist\index.js gateway --port 18789`

**Expected behavior**
`openclaw doctor --fix` completes maintenance and restarts the managed Gateway ("Gateway restarted and verified after Doctor repair."), leaving the service enabled.

**Actual behavior**
Two deterministic failure modes, both blocking:

1. Gateway RUNNING (as a task instance) → `doctor --fix` logs `Stopped the managed Gateway for Doctor repair.`, runs every check, then ALWAYS aborts at the end with:
```
Doctor could not complete maintenance. Check the reported service state, resolve the failure, and rerun doctor --fix.
Gateway service ownership or manager identity changed; inspect it before restarting manually.
```
Afterwards the Scheduled Task "OpenClaw Gateway" is left **DISABLED** and the gateway is stopped. Each doctor run repeats this.

2. Gateway STOPPED with the task enabled → `doctor --fix` refuses at entry:
```
Doctor could not enter maintenance. Stop the Gateway through its service owner, then run openclaw doctor --fix. Error: Gateway service ownership or shutdown could not be verified. Run openclaw gateway status --deep and stop it through its service owner before retrying.
```

**Steps to reproduce**
1. Upgrade 2026.7.x → 2026.8.2 while the gateway is running as the "OpenClaw Gateway" Scheduled Task.
2. In an elevated shell OUTSIDE the gateway process tree, run `openclaw doctor --fix` (gateway running).
3. Observe the exit error above; check `Get-ScheduledTask -TaskName "OpenClaw Gateway"` → State: Disabled.
[...]

PR #136578, opened 2026-09-02
fix: unblock Windows Doctor and retained workspace recovery

steipete:
Related: https://github.com/openclaw/openclaw/issues/136203 — Windows upgrade recovery and retained workspace state.
Related: https://github.com/openclaw/openclaw/issues/136123 — non-English Windows Task Scheduler inspection blocks Doctor maintenance.
Maintainer replacement for https://github.com/openclaw/openclaw/pull/136170 — the numeric-state fallback contribution by @LiuwqGit. This repair makes numeric state canonical instead of retaining the locale-sensitive parser and a second classification path, and includes the connected workspace recovery boundaries. The lead will decide whether to close the contributor PR only after this replacement lands and exact-landed native verification proves coverage.

<details>
<summary>Additional instructions</summary>

Maintainers can update this same-repository branch.

</details>

## What Problem This Solves

Fixes an issue where users recovering from an upgrade could not run Doctor on non-English Windows because an already-stopped Scheduled Task was reported as unknown. An explicit fleet could also leave retired setup files undiscovered in its configured shared workspace root. Pinning an agent back to that root while the Gateway was running could accept an unusable configuration and then repeatedly fail turns and heartbeats.

## Why This Change Was Made

Task Scheduler's numeric current state now owns runtime classification, existence, offline checks, and takeover evidence. Historical LastTaskResult is retained as history, not current activity. The localized LIST parser and duplicate result-code inference are removed. Queued/unknown inspection remains conservative; failed COM activation/connection cannot prove absence or justify a second direct node launch. Exact persisted-argv process evidence is preserved without confusing a Startup process with Scheduler supervision, and Windows CIM ownership no longer loads Gateway bind-host configuration unnecessarily.

Doctor's existing source inventory now also examines the explicitly authored shared workspace root, reusing the existing verified SQLite import/archive/receipt/cleanup flow. Runtime readiness still considers only selected agent/sandbox workspaces. The existing workspace-readiness assertion runs at final reload-candidate preparation, before managed-write persistence or watcher publication. Rejected external file edits can remain on disk, but the last healthy runtime remains active.

No schema, dependency, protocol, configuration option, or ambient-owner inference is added. No memory/persona files are moved or merged. Existing main-only fixes already cover legacy-list workspace preservation, stale setup-state migration conflicts, and ownerless fleet status; this does not duplicate them or guess ownership for an already-keyed markerless roster.

## User Impact

[...]

PR #145220, opened 2026-09-11
fix(doctor): doctor --fix leaves systemd-user gateway stopped after repair

### LINKED ISSUE #145070: [Bug]: doctor --fix on a systemd --user gateway always fails final revalidation ("ownership or manager identity changed") and leaves the gateway stopped (2026.9.4) — Windows-only fix in #137377
### Bug type

Other (service lifecycle / tooling)

### Beta release blocker

No

### Summary

On Linux with the gateway running as a **lingering systemd `--user` unit**, `openclaw doctor --fix` enters maintenance, applies its repairs successfully, stops the gateway — and then **always** aborts at the final revalidation step:

```
Doctor could not complete maintenance. Check the reported service state and resolve the failure.
Gateway service ownership or manager identity changed; inspect it before restarting manually.
```

It then **does not restart the gateway**, leaving the service stopped. Repairs themselves do land (in my case it retired a stale managed-plugin generation, reclaiming ~295 MB, and normalized the cron store), so this is purely the post-stop service revalidation.

This is the same failure mode as #137377 ("Doctor --fix on Windows ... always fails at its final restart: *Gateway service ownership or manager identity changed*; Scheduled Task left disabled"), which was verified fixed in 2026.9.1 and closed. That fix appears to have addressed the Windows/`schtasks` backend only — the identical symptom is still present on **2026.9.4** for the systemd `--user` backend.

It is *not* the same as #140908, which is an `EACCES` when spawning `systemctl --user is-enabled` that blocks doctor at **entry**. Here there is no `EACCES`, `systemctl --user` works fine from the same shell, and doctor gets all the way through its checks before failing.

### Steps to reproduce

1. Install the gateway as a systemd user unit via `openclaw gateway install` (default `openclaw-gateway.service` under `~/.config/systemd/user/`), with `loginctl enable-linger` set for the service user.
2. Confirm it is healthy: `systemctl --user is-active openclaw-gateway` → `active`; `openclaw gateway status --deep` → `Service: systemd user (enabled)`.
3. Run `openclaw doctor --fix --non-interactive` from a normal (non-login) shell as that same user, with `XDG_RUNTIME_DIR=/run/user/<uid>` and `DBUS_SESSION_BUS_ADDRESS` present.
4. Doctor stops the gateway, runs all checks, prints its "Doctor changes", then exits with the error above.
5. `systemctl --user is-active openclaw-gateway` → `inactive`. The gateway stays down until started manually.

### Expected behavior

Doctor completes maintenance and restarts the managed gateway ("Gateway restarted and verified after Doctor repair."), leaving the unit active — as it now does on the Windows backend after the #137377 fix.

### Actual behavior

Repairs are applied, the gateway is stopped, revalidation fails with `Gateway service ownership or manager identity changed`, and the unit is left `inactive` with no restart attempt.

**Things I ruled out** (each tested, failure identical every time):

- **Not a stale-state artifact.** Reproduces after a clean reboot, with the unit started by systemd at boot (`NRestarts=0`).
- **Not `--force`.** `doctor --fix --force --non-interactive` (documented as preserving service definitions) fails the same way.
- **Not a service-definition drop-in.** Fails with a `.service.d/` drop-in present and with none at all.
- **Not the unit being rewritten mid-run.** The unit file's mtime is unchanged across the doctor run, so its on-disk definition is stable.
- **Not missing session context.** `XDG_RUNTIME_DIR=/run/user/<uid>` and `DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/<uid>/bus` are both set, `id -u` matches the unit owner, and `systemctl --user` commands succeed from the same shell before and after.

### Where it comes from (best guess)

In `dist/update-command-service-maintenance-*.mjs`, `revalidateManagedGatewayServiceAfterUpdate()` throws `GatewayServiceUpdateOwnershipError` with this message when, after the stop:

```js
inspection.kind !== verdict.kind || !matchesStoppedService(before, params.state, inspection)
```

`matchesStoppedService()` requires (among other things):

```js
(process.platform !== "linux" || before.serviceManagerUid === void 0 ||
 before.serviceManagerUid === observedSystemdManagerUid(state))
&& (refreshDefinition || "fingerprint" in inspection && inspection.fingerprint === verdict.fingerprint)
```

Since the unit file is provably unchanged across the run, the fingerprint comparison seems unlikely to be the failing clause. My suspicion is the linux-only `serviceManagerUid` check: `before.serviceManagerUid` is captured while the unit is **running**, and `observedSystemdManagerUid(state)` may not resolve to the same value (or resolve at all) once the unit has been stopped — so a correct, unchanged user-scope service compares as "changed". I have not instrumented the running code to confirm which clause fails, so treat that as a hypothesis rather than a diagnosis.

### OpenClaw version

2026.9.4 (3a9d69d)

### Operating system

Debian/Raspbian (Linux 6.18, aarch64) on a Raspberry Pi 4. systemd user manager, `Linger=yes`.

### Install method

npm global (`npm install -g openclaw`), gateway registered via `openclaw gateway install --port 18789`.

### Model

openai/gpt-5.6-luna (default), thinking `low`

### Provider / routing chain

`openai` via ChatGPT OAuth (codex responses transport) as primary, with groq (`openai/gpt-oss-120b`, `gpt-oss-20b`, `qwen/qwen3.8-27b`) as fallbacks. Not model-dependent — the failure is in service lifecycle, and reproduces regardless of which model serves turns.

### Logs

```
◇  Doctor changes ───────────────────────────────────────╮
│  Cron store normalized at ~/.openclaw/cron/jobs.json.  │
├────────────────────────────────────────────────────────╯
Doctor could not complete maintenance. Check the reported service state and resolve the failure.
Gateway service ownership or manager identity changed; inspect it before restarting manually.
```

`openclaw gateway status --deep` immediately before the run (abridged):

```
Service: systemd user (enabled)
Service file: ~/.config/systemd/user/openclaw-gateway.service
CLI version: 2026.9.4   Gateway version: 2026.9.4
Runtime: running (state active, sub running, last exit 0)
```

### Impact and severity

Moderate, with a sharp edge. `doctor --fix` is unusable unattended: it takes the gateway down and never brings it back, so anything scheduled around it (or any operator who runs it and walks away) results in an outage until someone runs `systemctl --user start openclaw-gateway`. The repairs do apply, so the workaround is simply to always follow `doctor --fix` with a manual start — but that has to be known in advance, and the error text ("inspect it before restarting manually") reasonably discourages a blind restart.

Related: #137377 (same symptom, Windows/`schtasks`, fixed in 2026.9.1), #140908 (different root cause — `EACCES` blocking doctor at entry on a systemd `--user` install), #138934 (`service-revalidation-failed` during update on Windows).

---

*Prepared with AI assistance; all commands, versions and error output above were run and captured on the affected machine.*

Bruce-Yii:
Closes #145070

<details>

Keep **Allow edits from maintainers** enabled so maintainers can update this branch.

</details>

## What Problem This Solves

Fixes an issue where users running `openclaw doctor --fix` on affected Linux builds finish the state repairs but are left with a stopped systemd user Gateway and an ownership error. The completion path also needs to handle systemd collecting the stopped unit while its installed definition remains intact.

## Why This Change Was Made

Doctor omitted `requireLoadedCommand` from its final effective-service read. That selected ordinary `systemctl show` runtime inspection, which does not supply the manager UID required by Linux revalidation.

Completion now uses identity-bearing inspection under the existing native service-operation lock. When systemd collected the unit, the existing UID-bound `loadForInspection` operation reloads metadata without starting it. Doctor then verifies the service selector, installation, manager UID, and command fingerprint before restarting the unchanged service. Database coordinators are released before activation; the native operation lock is released before readiness polling.

Update repair/restart/recovery sibling reads already request both inspection options. `gateway status --deep` remains diagnostic. Update-parent activation ownership is unchanged. The Windows precedent is issue #137377, resolved by #136578 through current Task Scheduler state inspection.

## User Impact

Doctor restores the same previously running Gateway whether systemd retains or collects its stopped unit. It preserves the service file and enabled state, and still refuses a changed manager or launcher. A service already offline before maintenance is not activated.

## Evidence

Candidate: `66cec92c7bc3a6e31db58faab2f2a3e9649d5715`. Original contributor commit `94065630bfc060af0bc974f0b145874f0a58440b` remains an ancestor.

**Native Linux / ClawSweeper P1 and P2:** Debian 12 Docker with systemd 252.39 as PID 1, a lingering user manager, and Node 24.20.0. The real Gateway service was installed with `openclaw gateway install --port 18789`. The comparisons used `openclaw doctor --fix --non-interactive`, native service-state checks, and `/readyz` readiness observations.

| Build | Retained stopped unit | Collected stopped unit |
| --- | --- | --- |
| Published 2026.9.3 control | Exit 0, active, ready | Exit 0, active, ready |
| Pre-fix main `21b105c0f33` | Exit 1, ownership/manager error, inactive | Exit 1, ownership/manager error, inactive |
| Contributor flag-only change | Exit 0, active, ready | Exit 1, effective-command inspection error, inactive |
| Final candidate `66cec92c7bc` | Exit 0, active, ready | Exit 0, active, ready |

For collection, the generated service was disabled while running before Doctor stopped it. Native `GetUnit` confirmed `Unit openclaw-gateway.service not loaded` in the failed case. The final candidate prints `Gateway restarted and verified after Doctor repair.` in both cases; service-file SHA-256 and enabled-state comparisons match before/after. This answers P1 by recovering the collected unit rather than retaining the original refusal expectation, and P2 with native Linux lifecycle evidence.

Regression proof: the same four-case test fails on pre-fix production with two ownership-error failures and two passing refusal cases; all four pass on the final commit. The focused Doctor, systemd command/runtime, and update-maintenance suites pass **342 tests**, with one existing platform skip. The baseline `pnpm build`, final `OPENCLAW_RUN_NODE_SKIP_DTS_BUILD=1 pnpm build`, tarball integrity, `node scripts/check-changed.mjs`, and independent P1 review pass. One introduced variable-shadowing lint error was corrected before the successful gate rerun.

Artifact SHA-256: `b7eb947f6dbe05628a4bafd157cd7ed787dfeb5d4d8970df345e3c3e3bd64965`.

**Published-driver compatibility and known gap:** a separate user-owned 2026.9.3 install ran `openclaw update --yes --no-restart --tag /proof/fixed.tgz --json` against the candidate, with its installed service intentionally offline. Candidate migration rehearsal (`doctor --fix`), Doctor lint, config validation, plugin resolution, migration continuation, and Gateway canary all exited 0. The overall update exited 1 at `global install swap`: `Package rollback verification timed out` / `retained package tree changed`. The installed version remained 2026.9.3 and the service remained inactive. This is a failed full-upgrade cell, not a successful end-to-end update; the package-swap recovery failure is outside this Doctor change.

Thanks @Bruce-Yii for the original fix and regression. The maintainer commit retains the verified human `Co-authored-by` credit.

Initial CI on this head was blocked by [checks-node-changed-extensions-bundle-24](https://github.com/openclaw/openclaw/actions/runs/34647579663/job/103422278353): `extensions/openai/speech-runtime.contract.test.ts:333`, ordinary synthesis waiting for response headers, failed its watchdog with `ordinary synthesis did not time out`. The speech test and implementation are unchanged from the pre-fix base; this PR changes only Doctor maintenance, its regression, and Doctor documentation. The failed job log was inspected; no timeout or assertion was weakened. The maintainer authorized one failed-job rerun of run `34647579663` on the unchanged candidate head. Attempt 2 completed successfully, including the speech shard and aggregate gate. The initial speech failure also failed the aggregate `openclaw/ci-gate`.

## ClawSweeper review

The September 11 reviews find no actionable code defect and accepts the exact-candidate native Linux proof. Their generic config and data-model compatibility checklist items are addressed: the repair delta against merge base `21b105c0f33` contains only Doctor completion, its regression test, and Doctor documentation; it adds no config/default, schema, or retention changes. The Debian 12/systemd 252 table above proves successful restart for retained and collected units with unchanged service definition and enabled state, while the pre-fix and flag-only controls expose the original failures. The published 2026.9.3 driver passed the candidate Doctor stages; the separately documented package-swap failure remains outside this repair and is not full-upgrade proof.


steipete:
@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [66cec92c7bc3a6e31db58faab2f2a3e9649d5715](https://github.com/openclaw/openclaw/pull/145220/commits/66cec92c7bc3a6e31db58faab2f2a3e9649d5715)
- Landed commit: [f3f3d6b1ed680c364c1ea3b549e7e4fa8fc500b9](https://github.com/openclaw/openclaw/commit/f3f3d6b1ed680c364c1ea3b549e7e4fa8fc500b9)

PR #145222, opened 2026-09-11
fix: keep terminal retries with matching errors

steipete:
## What Problem This Solves

Fixes an issue where users opening another terminal tab could lose the Retry action for a catalog session that never connected. A later error from another tab could also display an unrelated catalog Retry action.

## Why This Change Was Made

Keep each pending open request with its tab until readiness succeeds or times out. The session controller now owns the displayed error and its matching recovery action together, replacing the separate retry object. Every error writer uses that owner, and a stale Retry click leaves a newer error intact. Tab removal also uses one shared path; redundant open/attach wrappers and unused lifecycle registration are removed.

## User Impact

A timed-out catalog session can be retried with its original catalog and agent while other tabs remain usable. An ordinary sibling exit preserves that recovery; a sibling error replaces it with the appropriate message.

## Evidence

The original missing-Retry cases and both follow-up error/click cases failed before their repairs. All 85 focused and sibling tests pass across six files. Canonical changed checks and the UI build pass. Chromium proof uses a synthetic Gateway and the default 30-second readiness timeout: Retry repeats the original open request, and a later sibling error displays without the old Retry. This proves the UI and Gateway request boundary; it does not exercise a real native terminal process.

Before: the catalog timeout has no Retry while the sibling remains usable.

Missing catalog retry

After: the timeout retains its matching Retry action.

Catalog retry retained

A later sibling error replaces the message and removes the unrelated action.

Sibling error without unrelated retry


steipete:
Merged via squash.

- Prepared head SHA: [5a05b71550d0142eaf22e8f40251327b356c89a1](https://github.com/openclaw/openclaw/pull/145222/commits/5a05b71550d0142eaf22e8f40251327b356c89a1)
- Landed commit: [24109cc97c2e4aaca2cdfb7365a92fa977ac6fe3](https://github.com/openclaw/openclaw/commit/24109cc97c2e4aaca2cdfb7365a92fa977ac6fe3)

PR #145223, opened 2026-09-11
fix(telegram): deliver plain text after legacy rich rejection

### LINKED ISSUE #145201: [Bug] telegram: richMessages vs older self-hosted Bot API server (apiRoot) — every send 400s "rich message must be non-empty", total channel blackout
### Bug type

Telegram channel / outbound delivery

### Summary

With `channels.telegram.richMessages: true`, a gateway speaking Bot API 10.3 sends every rich reply via `sendRichMessage`. Against a **self-hosted Bot API server older than 10.3** (set via `apiRoot`), the server cannot parse the 10.3 `rich_message` schema and rejects **every** call with `400: Bad Request: rich message must be non-empty`. This hard-bricks **all** outbound Telegram traffic — including plain system notices — with an error that reads like an empty-payload bug, from the first send after upgrade, until the flag is turned off or the server is upgraded.

### Steps to reproduce

1. Self-host a local Bot API server at Bot API 10.1 (e.g. `aiogram/telegram-bot-api:10.1`, `--local` mode) and point `channels.telegram.apiRoot` at it.
2. Set `channels.telegram.richMessages: true` on OpenClaw 2026.9.3.
3. Restart the gateway and send any message to the bot (a cron delivery to a group also triggers it — no user interaction needed).

Direct probe of the 10.1 server (`curl $apiRoot/bot$TOKEN/sendRichMessage`):

```
{"chat_id": 0}                                                          → 400 "rich message must be non-empty"
{"chat_id": 0, "rich_message": {"blocks":[{"type":"paragraph","text":"x"}]}} → 400 "rich message must be non-empty"   # well-formed payload, still rejected
{"chat_id": 0, "text": "x"}  (sendMessage control)                      → 400 "Bad Request: chat not found"          # server validates fine
```

The same `sendRichMessage` payload against a 10.3 server returns `chat not found` (schema accepted) and delivers normally with a real chat id.

### Expected behavior

- A capability/version check against the `apiRoot` server before using Bot API 10.3 methods (at channel start or first rich send), latching rich messages off with a loud log line, and/or
- the existing rich→plain fallback firing on this specific 400 so the channel degrades to `sendMessage` instead of blacking out.

### Actual behavior

Every outbound send failed with the same misleading 400 for the entire container lifetime (20/20 `sendRichMessage` calls over 15h in our incident; zero successful sends). Observed failure surface:

```
[telegram] sendRichMessage failed: Call to 'sendRichMessage' failed! (400: Bad Request: rich message must be non-empty)
[telegram] slash final reply failed: GrammyError: Call to 'sendRichMessage' failed! ...
[restart-sentinel] Gateway restart ok (gateway.restart): outbound delivery failed; queued for recovery: OutboundDeliveryError: ...
```

Failed sends then poison the durable-delivery drain (entries stuck refusing replay — cf. #126246), and reply-pipeline retries stall behind them, so the channel never self-recovers. The `rich-plain-fallback` (`extensions/telegram/src/rich-plain-fallback.ts`) did not fire for any of these failures, including for pure system text (restart-sentinel notices), so the outage was total rather than cosmetic.

### OpenClaw version

2026.9.3 (1391f7c), gateway docker image, node 24.19.0

### Operating system

Linux (docker), self-hosted local Bot API server: aiogram/telegram-bot-api 10.1, `--local` mode, reached through a plain HTTP proxy at `apiRoot`

### Install method

Docker (ghcr.io/openclaw/openclaw:2026.9.3 lineage)

### Model

n/a (failure is pre-model: system notices fail identically)

### Provider / routing chain

n/a

### Logs

See "Actual behavior". On request I can provide a full-day correlated timeline (request ledger confirms the failing `sendRichMessage` bodies were 578B–5.7KB — payloads were **not** empty; the older server just can't parse the 10.3 rich schema and reports it as empty).

n/a (log-based)

### Impact and severity

**Total, silent Telegram blackout.** Distinct from the June client-side rollout incident tracked in #103458 (call succeeds, recipient renders a placeholder — undetectable bot-side): the server-side skew **is** detectable and mitigable bot-side (probe or first-error latch), and it destroys the whole channel, not just rendering. Worth covering in the #103458 re-enable criteria, since any deployment pairing a current gateway with a self-hosted Bot API server that lags the gateway's Bot API version hits this. Workaround: `richMessages: false` (plain path unaffected) or upgrade the local server to ≥ Bot API 10.3 (verified: both restore delivery).

0xalydev:
## What Problem This Solves

Fixes #145201. Older self-hosted Telegram Bot API servers reject rich sends with `rich message must be non-empty`, leaving replies and system notices undelivered.

## Why This Change Was Made

The shared classifier recognizes the rejection once, retaining `rich-content-required` for rich requests and `empty-content` for HTML requests. The empty-content predicate already reads that classifier.

## User Impact

Nonempty messages reach the recipient through the existing plain-text fallback.

## Evidence

`pnpm openclaw message send --channel telegram --target <chat> --message <marker> --json` with rich messages enabled:

- Before: exit 1, one rich rejection, no plain request or recipient marker.
- After: exit 0, one plain send, one accepted receipt and independent recipient readback.

The historical endpoint was unavailable. Its exact rejection was controlled at the HTTP boundary; successful delivery used Telegram's real Test Server. Rich success, unrelated errors and blank text were checked.

## Compatibility

No settings, protocol, schema or dependency changes.

## Consumers

Durable sends, replies and previews share the fallback.

## Invalidation

No new state or cache; acceptance accounting is unchanged.

## Tests

All 284 tests passed in four suites on a fresh merge with the pinned main revision. Its source tree matches this head. Coverage includes ordinary recovery, topics and partial-delivery accounting. Type-aware lint and types passed.

Follow-up: the preview path must recheck send authority before a plain send or edit when the session writer changes while a rich request is pending.

### DIFF
--- a/extensions/telegram/src/bot/delivery.test.ts
+++ b/extensions/telegram/src/bot/delivery.test.ts
@@ -2882,6 +2882,35 @@ describe("deliverReplies", () => {
     expect(mockCallArg(sendMessage, 0, 2)?.parse_mode).toBeUndefined();
   });
 
+  it("falls back to standard sendMessage when older self-hosted Bot API server rejects rich message as non-empty", async () => {
+    const runtime = createRuntime();
+    const sendMessage = vi.fn().mockResolvedValue({
+      message_id: 15,
+      chat: { id: "123" },
+    });
+    const bot = createBot({ sendMessage });
+    (bot.api.raw as unknown as { sendRichMessage: ReturnType<typeof vi.fn> }).sendRichMessage = vi
+      .fn()
+      .mockRejectedValue(
+        new Error(
+          "GrammyError: Call to 'sendRichMessage' failed! (400: Bad Request: rich message must be non-empty)",
+        ),
+      );
+    const text = "system notice delivered through fallback";
+
+    await deliverWith({
+      replies: [{ text }],
+      runtime,
+      bot,
+      richMessages: true,
+    });
+
+    expect(sendMessage).toHaveBeenCalledTimes(1);
+    expect(firstMockCallArg(sendMessage, 0)).toBe("123");
+    expect(firstMockCallArg(sendMessage, 1)).toBe(text);
+    expect(mockCallArg(sendMessage, 0, 2)?.parse_mode).toBeUndefined();
+  });
+
   it("falls back to plain text before raw rich send when rich markdown renders empty", async () => {
     const runtime = createRuntime();
     const sendMessage = vi.fn().mockResolvedValue({
--- a/extensions/telegram/src/rich-plain-fallback.test.ts
+++ b/extensions/telegram/src/rich-plain-fallback.test.ts
@@ -19,6 +19,12 @@ const fallbackCases = [
     htmlTrigger: "empty-content",
   },
   {
+    name: "rich-message-non-empty",
+    message: "Bad Request: rich message must be non-empty",
+    richTrigger: "rich-content-required",
+    htmlTrigger: "empty-content",
+  },
+  {
     name: "html-parse",
     message: "Bad Request: can't parse entities: unsupported tag",
     richTrigger: "html-parse",
@@ -118,6 +124,7 @@ describe("isTelegramEmptyContentError", () => {
     "Bad Request: message text is empty",
     "Bad Request: text must be non-empty",
     "Bad Request: RICH_MESSAGE_CONTENT_REQUIRED",
+    "Bad Request: rich message must be non-empty",
   ])("recognizes %s", (message) => {
     expect(isTelegramEmptyContentError(new Error(message))).toBe(true);
   });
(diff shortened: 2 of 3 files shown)


0xalydev:
@clawsweeper re-review

obviyus:
landpr-approve-admin-merge
pr: 145223
head: 03d6aef2e031cf67351d3c35ff5ec73b2ad0c770
scope: owner-review

obviyus:
Thanks @0xalydev!

**Proof:** A controlled legacy rich-message rejection reached the existing plain send, with one accepted delivery and independent recipient readback on Telegram's Test Server.

- **Bug:** Older self-hosted servers rejected rich messages without plain-text recovery.
- **Fix:** The shared fallback recognizes the explicit rejection.

| Before | After |
| --- | --- |
| Send failed and no message arrived. | One plain message reached the recipient. |

PR #145233, opened 2026-09-11
refactor(tests): cover model auth through image decisions

steipete:
## What Problem This Solves

Model-auth tests expose a private predicate through a test-only global and bridge even though existing image decisions can express the same auth expectations.

## Why This Change Was Made

Retire the global, bridge, publication entry, and obsolete assertion allowance. Keep the auth predicate and production callers unchanged; retain exact image keep/substitute decisions and move the active inline-key cooldown case to the real image decision with a healthy-key check. No supported public API or auth algorithm changes.

## User Impact

No user-facing behavior change. Production removes 6 lines, tests/support add 11 and remove 53, and the assertion baseline removes one row. No cases are removed.

## Evidence

Both complete suites passed before and after: 127 cases (20 + 107), with only the documented case relocation. The helper suite exercises actual healthy/cooling availability; the image-tool suite provides supporting routing coverage. Independent review found no actionable issues through P2.

The narrow API-key versus ChatGPT-backed distinction was checked directly in public Codex source at `e3a52b8`: [auth modes](https://github.com/openai/codex/blob/e3a52b87b28760413eafa340e2ab23d653f0bbe7/codex-rs/protocol/src/auth.rs#L9-L21) and [API-key accessor](https://github.com/openai/codex/blob/e3a52b87b28760413eafa340e2ab23d653f0bbe7/codex-rs/login/src/auth/manager.rs#L538-L549). This is source contract evidence, not live authentication proof.

All 32 normal changed-file checks and the normal full build passed with source and dependencies unchanged.


steipete:
Merged via squash.

- Prepared head SHA: [30d11ddbf089826e7e0a9f1b3e2b3fa7612582b5](https://github.com/openclaw/openclaw/pull/145233/commits/30d11ddbf089826e7e0a9f1b3e2b3fa7612582b5)
- Landed commit: [83b6e1ade894d2c84588a16855a1b4a301ea65b5](https://github.com/openclaw/openclaw/commit/83b6e1ade894d2c84588a16855a1b4a301ea65b5)

PR #145238, opened 2026-09-11
perf(ui): reuse file reads across memory search matches

### LINKED ISSUE #145237: Memory search previews read the same file repeatedly
## Summary

Memory search previews fetch the same complete file again when users open another matching range from that file, including while the first read is still pending.

## Reproduction and observed result

Search for two ranges in the same file, open one result, then open the other before the first workspace read completes. The original sends two `agents.workspace.get` requests. The candidate sends one, and the same single-read result is reused after loading.

## Expected behavior

Ranges with the same exact file path should share one pending, loaded, or error entry until the existing new-search or scope reset. Expansion and highlighted lines should remain specific to the selected result. A stale request must not repopulate the new search or agent scope.

## Evidence and limits

The same-file pending-read regression fails on the original with two requests instead of one; all 12 candidate owner tests pass. The synthetic Chromium page comparison also records pending and loaded read counts of two → one while preserving one expanded row and highlighted lines 5–6. Inspected before/after screenshots are byte-identical.

Managed P2 review and the selected UI/core-test gate pass. That gate used the default `origin/main` reference, whose resolved commit at execution was not recorded; native exact-head/base validation remains required before landing. No live Gateway or retained-heap, RSS, or timing improvement is claimed.

steipete:
Closes #145237

<details>

</details>

## What Problem This Solves

Fixes an issue where users opening different Memory search matches from the same file caused duplicate full-file reads, including when the first preview request was still pending.

## Why This Change Was Made

Key preview state by exact file path and use the pending detail entry itself to identify the current request, removing the separate request-marker map. All ranges from a file intentionally share pending, loaded, or error state until the existing new-search or scope reset; expansion and highlighting remain keyed to the selected result.

## User Impact

Opening another range from the same file reuses its pending request or cached result. A new search or scope reset clears that shared state, including errors, and stale completions remain unable to replace the current entry.

## Evidence

- The pending same-file regression fails on the original with two workspace reads instead of one. All 12 candidate owner tests pass.
- A real Chromium page with a synthetic mock Gateway records pending and loaded reads of two → one. Both versions retain one expanded result, the same file path and highlighted lines 5–6. The inspected full-page and result-detail screenshot pairs are byte-identical.
- Managed P2 review and the selected UI/core-test gate pass. The gate used default `origin/main`; its resolved execution SHA was not recorded. Native exact-head/base guards remain required before landing, and no exact-base gate result is inferred from today's reference.

Proof covers request counts and preserved presentation using the source-Vite mock Gateway. No live Gateway, retained-heap, RSS or timing improvement is claimed.

### DIFF
--- a/ui/src/pages/config/memory-memories.test.ts
+++ b/ui/src/pages/config/memory-memories.test.ts
@@ -60,6 +60,21 @@ const result = {
   source: "memory" as const,
 };
 
+function memoryFileResponse(content: string, path = result.path) {
+  return {
+    agentId: "main",
+    file: {
+      path,
+      name: path.split("/").at(-1),
+      size: content.length,
+      updatedAtMs: 1,
+      mimeType: "text/plain",
+      encoding: "utf8",
+      content,
+    },
+  };
+}
+
 describe("MemoryMemoriesElement", () => {
   it("renders idle and gateway-update-required states", async () => {
     const current = createElement(vi.fn(() => Promise.resolve({})));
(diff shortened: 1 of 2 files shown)


steipete:
The data-model compatibility item is not applicable to this change at `c2c1cb9d1be2761f1866f5e7e50ce6427dffef22`. The patch changes only the identity of a component-local preview cache from a search result to an exact file path. RPC and data formats are unchanged, and there are no persisted-state, schema, or migration changes.

The pending, loaded, and error entries retain their intended lifetime through the existing new-search and scope resets. Accepted tests cover those resets and rejection of stale responses; the synthetic browser comparison preserves expansion and highlighting while same-file workspace reads fall from two to one. These checks cover the transient lifecycle affected by this patch. No durable data-model upgrade proof is required for this scope.

steipete:
Merged via squash.

- Prepared head SHA: [c2c1cb9d1be2761f1866f5e7e50ce6427dffef22](https://github.com/openclaw/openclaw/pull/145238/commits/c2c1cb9d1be2761f1866f5e7e50ce6427dffef22)
- Landed commit: [4d59623b97ae1dd1788c368dcc75d3bffaa4de25](https://github.com/openclaw/openclaw/commit/4d59623b97ae1dd1788c368dcc75d3bffaa4de25)

PR #145240, opened 2026-09-11
fix: keep earlier messages selected during chat refresh

steipete:
## What Problem This Solves

Fixes an issue where users reading earlier retained messages in Chat would be returned to the latest page when new activity refreshed the conversation. A refresh arriving during pagination could also discard the requested earlier page.

## Why This Change Was Made

The existing pending-input view now retains the selected cursor and owns request freshness. Incoming custody updates still reconcile delivery receipts immediately, then refresh the selected page. New observations coalesce into a current read, and an explicit return to Latest supersedes background pagination.

## User Impact

Users can keep reading earlier messages while activity continues. Their recorded dispositions stay current, the conversation and live output remain visible, and a delayed older response cannot undo a return to Latest. A failed page request leaves the previous page readable with an error.

## Evidence

All four page/delta and request-order regressions fail on the original code. All 485 related tests pass across six files, including receipt-driven outbox retirement, request supersession, connection changes, and failed navigation. Canonical changed checks, the UI build, and four independent review passes also passed.

Chromium proof uses a synthetic Gateway and the actual paging controls. Before, incoming activity removes the older row and Latest button. After, the older row remains selected with its updated disposition; Latest remains selected after a delayed older response. Transcript, live output, and Stop controls stay present throughout.

Before: activity resets the selected page

After: the selected earlier page refreshes in place

### DIFF
--- a/docs/web/control-ui/sessions-and-sidebar.md
+++ b/docs/web/control-ui/sessions-and-sidebar.md
@@ -51,7 +51,8 @@ cancellation or a Gateway restart interrupts that wait,
 the message stays readable with its recorded disposition and is never resent
 automatically. Copy it into the composer to start a new attempt. **Show earlier
 messages** pages through messages that are still waiting or were stopped before
-processing; **Show latest messages** returns to the newest page. A long message
+processing; **Show latest messages** returns to the newest page. Incoming activity
+refreshes the page you are reading without changing your selection. A long message
 uses the normal full-message reader without becoming a transcript reply, fork,
 or rewind target.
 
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f1671f98db372bb8f00d5971682beb0a7327de52](https://github.com/openclaw/openclaw/pull/145240/commits/f1671f98db372bb8f00d5971682beb0a7327de52)
- Landed commit: [b66f3647cec1e69da4118c269e05cc1bebf12384](https://github.com/openclaw/openclaw/commit/b66f3647cec1e69da4118c269e05cc1bebf12384)

Issue #144959, opened 2026-09-11
tasks list --status failed counts intentional heartbeat "empty-heartbeat-file" skips as failed


Itzgizz:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

An ambient heartbeat whose heartbeat file is intentionally empty/comments-only skips every interval by design, but each skip is recorded as a *failed* task in `openclaw tasks list --status failed`, even though the heartbeat event layer reports the run as ok-empty/silent with consecutiveErrors=0.

### Steps to reproduce

1. Leave the agent's heartbeat file empty/comments-only (the stock template reads: "Keep this file empty (or with only comments) to skip heartbeat API calls").
2. Let the ambient heartbeat run for several hours (`agents.defaults.heartbeat = {"every":"30m"}`).
3. `openclaw cron get <heartbeat-id>` → `lastRunStatus:"skipped"`, `lastError:"heartbeat skipped: empty-heartbeat-file"`, `consecutiveSkipped` rising, `consecutiveErrors:0`.
4. `openclaw tasks list --status failed` → one row per interval (~48/day) for the heartbeat cron, reason `heartbeat skipped: empty-heartbeat-file`.

### Expected behavior

An intentional ok-empty skip should be recorded as *skipped* (or excluded from `--status failed`), consistent with `openclaw system heartbeat last` already reporting `status:"ok-empty"`, `silent:true`, `indicatorType:"ok"`. A designed no-op should not surface as a failed task.

### Actual behavior

Each intended skip is labeled *failed* in task accounting: `openclaw tasks list --status failed` accrues ~48 rows/day for the heartbeat cron, all with reason `heartbeat skipped: empty-heartbeat-file`, and `consecutiveSkipped` climbs unbounded (observed 330). Failed-task counts and any monitor/audit built on `--status failed` are inflated with false positives, despite `consecutiveErrors:0` and `openclaw system heartbeat last` reporting `ok-empty`/`silent`/`indicatorType:"ok"`.

### OpenClaw version

2026.8.2

### Operating system

Windows 11 Pro (build 26200)

### Install method

npm global (openclaw); local gateway, login-item

### Model

anthropic/claude-sonnet-5 (agent default; not exercised — the empty-file skip returns before any model call)

### Provider / routing chain

openclaw -> local LiteLLM proxy -> anthropic (agent default; not exercised — skip returns before any provider request)

### Additional provider/model setup details

[...]

PR #145025, opened 2026-09-11
fix(cron): record skipped cron runs as cancelled, not failed

tzy-17:
## What Problem This Solves

`openclaw tasks list --status failed` counts intentional heartbeat no-ops as failed runs. A heartbeat job whose scratch file is effectively empty skips each interval (`cronRunStatusToTaskStatus` mapped every non-`ok` cron status to `failed`, including `skipped`). The task registry has no `skipped` status, so each skip landed as a `failed` task with reason `heartbeat skipped: empty-heartbeat-file`, inflating failed counts in `tasks list --status failed` and any monitor/audit that keys off it. The run history ledger already keeps the true `skipped` status and the skip reason, so the loss is only in the derived task-row status.

Closes #144959.

## What This Change Does

The cron state machine already tracks skips separately from execution errors (`consecutiveSkipped` vs `consecutiveErrors`); only the task-ledger projection conflated them. This PR restricts the `cancelled` task mapping to **explicit heartbeat no-op reasons**:

- **`cancelled` task** — `heartbeat skipped: empty-heartbeat-file` and `heartbeat skipped: no-pending-event` (an intentional no-op: nothing to do). These no longer inflate `openclaw tasks list --status failed`.
- **`failed` task (unchanged)** — every other skipped outcome: failed model/cron preflights, unmet trigger conditions, released reservations, and a **globally disabled heartbeat**. These are unsuccessful runs and stay visible to failure monitors, preserving the reporting contract established in openclaw/openclaw#123787 and the existing heartbeat-payload contract (`disabled` skip → `failed` completion).

The mapping lives in `cronRunStatusToTaskStatus` (`src/cron/task-run-detail.ts`). The fix also preserves an **operator cancellation reason** when a late skipped outcome arrives for a task the operator already cancelled: the task row keeps `Cancelled by operator.` and the skip reason stays in the cron run history detail (`src/cron/service/task-runs.ts`).

`TaskStatus` has no `skipped` value; `cancelled` is the existing terminal, non-failure status that best expresses "this run intentionally did not execute," and it is accepted by every `cronRunStatusToTaskStatus` caller.

## User Impact

- An intentional heartbeat no-op (empty heartbeat file, or a stale wake with no pending events) no longer appears in `openclaw tasks list --status failed`; it appears as a `cancelled` task.
- Unsuccessful skips — failed provider preflights, unmet trigger conditions, released reservations, and disabled heartbeats — still appear under `--status failed`, so failure monitors keep working.
[...]

PR #145242, opened 2026-09-11
fix(cron): record designed heartbeat skips as successful no-ops

Lubaoshuai:
## What Problem This Solves

An agent whose heartbeat file is intentionally empty (or comments-only) skips every interval by design, but each skip was recorded as a `skipped` cron run, which `resolveCronCompletionStatus` maps to `failed`. `openclaw tasks list --status failed` accrued one row per interval (~48/day) for the heartbeat cron and failure alerts fired, even though `consecutiveErrors` stayed 0 (#144959).

Designed no-op skips now settle as `ok` runs whose summary carries the skip reason (`heartbeat skipped: empty-heartbeat-file`), matching the busy-deferral handoff path that already returned `ok`. The heartbeat runner's own state (`system heartbeat last`, `consecutiveSkipped`) keeps reporting the precise skipped status, so no signal is lost.

## Evidence

- **Real-behavior proof (updated in f8c3fa97 per review):** `src/cron/service.heartbeat-real-preflight.test.ts` wires the **real heartbeat runner** (real preflight, no injected heartbeat results), a **real comments-only scratch row** in the sqlite state store (stock "keep this file empty" content), and the **real cron service**. The two review findings are addressed:
  - **Same-partition fixture:** the sandbox is now selected as the cron partition the preflight actually reads — `OPENCLAW_STATE_DIR` points the default partition (`<state>/cron/jobs.json`) at the seeded `cronStorePath`, so `resolveHeartbeatPreflight` genuinely consumes the seeded comments-only scratch instead of a different store.
  - **Required dependency supplied:** the `CronServiceDeps.requestHeartbeat` member is provided (non-waiting) alongside `requestHeartbeatAndWait`, matching the production contract.
  - **Skip assertions, not just success:** the finished event must carry `summary: "heartbeat skipped: empty-heartbeat-file"`, and the reply/telegram spies must stay unused — so an ordinary successful agent turn can no longer satisfy the test. Paused-reminder recovery stays covered: the operator `disabled` skip keeps its `skipped`/`failed` outcome in the unit matrix.

Runtime output (local vitest run of the file at f8c3fa97, redacted):

```
✓ src/cron/service.heartbeat-real-preflight.test.ts > real heartbeat preflight task accounting > settles a real empty-scratch preflight as a successful no-op task 1066ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
```

- The test also asserts the ledger that feeds `tasks list --status failed` contains zero failed cron rows after the designed skip.
- Unit-level matrix in `service.heartbeat-payload.test.ts` (10/10): designed skips settle `ok`/`succeeded` with `consecutiveErrors: 0`; the operator `disabled` skip keeps its `skipped` outcome for visibility.
- Matches the existing intentional-silence precedent in `resolveCronCompletionStatus`, and the busy-deferral handoff path in `timer-execution.ts` that already returned `ok`. The runner's own state (`system heartbeat last`, `consecutiveSkipped`) continues to report the precise skipped status.

### DIFF
--- a/src/cron/service.heartbeat-payload.test.ts
+++ b/src/cron/service.heartbeat-payload.test.ts
@@ -96,9 +96,9 @@ describe("heartbeat payload execution", () => {
     {
       label: "disabled skip",
       child: { status: "skipped", reason: "disabled" } as HeartbeatRunResult,
-      expectedStatus: "ok",
-      expectedCompletion: "succeeded",
-      expectedError: undefined,
+      expectedStatus: "skipped",
+      expectedCompletion: "failed",
+      expectedError: "heartbeat skipped: disabled",
       expectedConsecutiveErrors: 0,
     },
     {
--- a/src/cron/service.heartbeat-payload.test.ts
+++ b/src/cron/service.heartbeat-payload.test.ts
@@ -96,9 +96,17 @@ describe("heartbeat payload execution", () => {
     {
       label: "disabled skip",
       child: { status: "skipped", reason: "disabled" } as HeartbeatRunResult,
-      expectedStatus: "skipped",
-      expectedCompletion: "failed",
-      expectedError: "heartbeat skipped: disabled",
+      expectedStatus: "ok",
+      expectedCompletion: "succeeded",
+      expectedError: undefined,
+      expectedConsecutiveErrors: 0,
+    },
+    {
+      label: "empty heartbeat file skip",
+      child: { status: "skipped", reason: "empty-heartbeat-file" } as HeartbeatRunResult,
+      expectedStatus: "ok",
+      expectedCompletion: "succeeded",
+      expectedError: undefined,
       expectedConsecutiveErrors: 0,
     },
   ])("records the settled heartbeat child $label", async (testCase) => {
--- a/src/cron/service/timer-execution.ts
+++ b/src/cron/service/timer-execution.ts
@@ -213,7 +213,15 @@ export async function executeJobCore(
           }
         : heartbeatResult.status === "failed"
           ? { status: "error" as const, error: `heartbeat failed: ${heartbeatResult.reason}` }
-          : { status: "skipped" as const, error: `heartbeat skipped: ${heartbeatResult.reason}` };
+          : // Designed no-op skips (empty heartbeat file, no pending event, ...)
+            // are successful runs by definition; recording them as skipped
+            // failures inflated `tasks list --status failed` and failure
+            // alerts (gh-144959). The runner's own state keeps the skipped
+            // counters for `system heartbeat last`.
+            {
+              status: "ok" as const,
+              summary: `heartbeat skipped: ${heartbeatResult.reason}`,
+            };
     return triggerEval ? { ...result, triggerEval } : result;
   }
   if (effectiveJob.sessionTarget === "main") {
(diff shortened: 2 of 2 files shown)


Lubaoshuai:
Added the real-behavior proof requested by the review: `service.heartbeat-real-preflight.test.ts` drives the real heartbeat runner preflight over a real comments-only scratch row in the sqlite state store and asserts the run settles as a successful no-op with zero failed cron task rows in the ledger (the `tasks list --status failed` source). PR body Evidence section updated accordingly. @clawsweeper re-review

Lubaoshuai:
6fb6ec66 restores the fixture safeguards that ffb0026a accidentally removed, addressing both P2 findings:

- **Preflight partition restored:** `OPENCLAW_STATE_DIR` again points the default cron partition at the sandbox, so `resolveHeartbeatPreflight` genuinely consumes the seeded comments-only scratch instead of falling through to an ordinary agent turn.
- **Required dependency restored:** the `CronServiceDeps.requestHeartbeat` member is supplied again alongside `requestHeartbeatAndWait` (type contract at `src/cron/service/state.ts:170`).
- **Skip-specific assertions restored:** the finished event must carry `summary: "heartbeat skipped: empty-heartbeat-file"` and the reply/telegram spies must stay unused, plus the state-db teardown and `vi.unstubAllEnvs()` the fixture needs.

Local vitest run at 6fb6ec66 (cron config, redacted terminal output):

```
✓ src/cron/service.heartbeat-real-preflight.test.ts > real heartbeat preflight task accounting > settles a real empty-scratch preflight as a successful no-op task 1155ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
```

Full local cron-suite run at 6fb6ec66 (`service.heartbeat-payload.test.ts` + `service.heartbeat-real-preflight.test.ts` + `service.test.ts`): 3 files / 10 tests, all passed. oxfmt `--check` clean on all touched files. PR body Evidence section updated.

@clawsweeper re-review

Lubaoshuai:
@clawsweeper re-review

Lubaoshuai:
Thank you for the detailed revision-7 review — both fixture findings are confirmed resolved, and the remaining items are now about product semantics rather than the patch itself.

Two honest acknowledgments:

1. **Operator-contract footprint.** The review is right that converting designed skips to `ok` runs makes them invisible to existing consumers of skipped cron history, `consecutiveSkipped`, and `includeSkipped` alerts. That is a real reporting change, and #145025's projection approach (keep cron outcomes `skipped`, change the task projection so they stop surfacing as failures) has a smaller operator-contract footprint while still fixing the alert-noise problem in #144959.

2. **Gateway-level demo.** A production `dist` build is not available in this environment (the prepack pipeline requires the full workspace toolchain), so the vitest harness — which drives the real CronService, real heartbeat preflight, and the real SQLite state store, with only the reply/telegram delivery dependencies fixture-injected and asserted unused — is the strongest evidence I can attach for now. It covers the designed no-op path, the unsuccessful-skip control (operator `disabled` stays `skipped`/`failed`), and the ledger assertion; I've marked it as such in the body rather than overstating it.

Given the reviewer recommendation and the required owner decision, I'll defer to the maintainers: if the projection approach in #145025 is preferred, I'm happy to close this PR in its favor (or rework this one onto the projection semantics if that's useful). Until then I'll leave the PR open so the decision has both options on the table.

PR #145251, opened 2026-09-11
test: reuse compiled modules in native update fixtures

steipete:
## What Problem This Solves

Native update-executor tests repeatedly transform the same TypeScript modules in fresh nested processes. The nine-case file took 82.6 seconds locally, despite already preparing a compiled worker generation for the invocation.

## Why This Change Was Made

Register the required modules in the existing test-only worker declarations and resolve their URLs through the canonical runtime resolver. Each child loads modules from the shared compiled graph; direct source-mode execution retains its existing TypeScript preload. Package entries and production runtime code are unchanged.

The nine cases still launch independent child processes and retain every original PID-before-input binding, readiness and revocation barrier, native/config effect assertion, and process-extinction check. The finalizer tests retain their source-level hooks; this change does not try to bundle those hooks away.

## User Impact

CI avoids repeated TypeScript setup inside the native fixture children. There is no change to application update behavior or test sharding.

## Evidence

One same-host, same-command local before/after pair, using Node 24 and including generation preparation:

| Measurement | Original | Candidate |
| --- | ---: | ---: |
| Complete nine-case invocation | 82.61s | 25.69s |
| Worker preparation | 6.787s | 4.865s |
| Generation inputs / outputs | 8,848 / 4,498 | 8,849 / 4,504 |
| BSD time maximum RSS | 2,164,424,704 bytes | 2,178,744,320 bytes |

The invocation was 56.92 seconds shorter (68.9%). The preparation difference accounts for only 1.92 seconds; most of the reduction is inside the case bodies. The seven roots mostly share an already-built graph, adding one input and six outputs. Maximum RSS increased 13.66 MiB (0.66%); this is not a memory reduction claim or a simultaneous process-tree peak measurement.

All nine cases passed with identical case inventories and assertions; 36 existing sibling tests passed. Core-test/script types, targeted lint, formatting, and independent P2 review pass. Generation and child cleanup were verified. Caches were retained: this local pair does not establish a cold-cache or whole-CI speedup. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34645579370) now passes on `57e315321ef2444a94115be43d441bd8dfae0831`: all 101 checks are terminal without failures.

Four files change: test rewiring +23/-23, 40 lines of pure test-entry metadata, and four lines in existing test compiler metadata. Production runtime and public API delta are zero. Fresh process and lifecycle coverage remains intact.

### DIFF
--- a/scripts/lib/vitest-worker-build-entries.mts
+++ b/scripts/lib/vitest-worker-build-entries.mts
@@ -10,6 +10,7 @@ import {
   gatewayDirectStopEntrypoints,
   stateDirGatewayFixtureEntrypoint,
 } from "../../src/cli/cli-entrypoint.test-support.ts";
+import { updateExecutorNativeEntrypoints } from "../../src/cli/update-cli/update-command-executor-native-runtime.test-support.ts";
 import { doctorConfigRuntimeEntrypoints } from "../../src/commands/doctor-config-runtime.test-support.ts";
 import { cronOwnerHardeningEntrypoints } from "../../src/cron/owner-hardening-runtime.test-support.ts";
 import { sessionListCacheRetentionEntrypoint } from "../../src/gateway/server-methods/sessions-list-cache-retention-entrypoint.test-support.ts";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [57e315321ef2444a94115be43d441bd8dfae0831](https://github.com/openclaw/openclaw/pull/145251/commits/57e315321ef2444a94115be43d441bd8dfae0831)
- Landed commit: [6b2bde25a8e83d5a2416fba9ac773659a9c357bf](https://github.com/openclaw/openclaw/commit/6b2bde25a8e83d5a2416fba9ac773659a9c357bf)

PR #145255, opened 2026-09-11
fix(auto-reply): report rejected partial question answers instead of failing dispatch

### LINKED ISSUE #145126: [Bug]: ask_user answer that misses one of several questions is dropped with no reply
### Bug type

### Beta release blocker

No

### Summary

When the agent asks two or more questions in one `ask_user` call and the user answers in a single unkeyed line, the answer is dropped. The user gets no reply. The Telegram update is retried until the ingress gives up on it, and when the question wait expires the agent carries on as if nobody answered.

The parse and the validation both behave as designed. `buildAgentHarnessUserInputAnswers` maps an unkeyed multi-question reply by line, so a question with no line gets an empty answer (`src/agents/harness/user-input-bridge.ts:254`, tested by "preserves blank fallback lines so skipped answers stay aligned"). `question.resolve` then rejects it with `question 'scope' requires an answer` (`src/gateway/question-manager.ts:350`). After the rejection, nothing reaches the user:

- `claimPendingAgentQuestionAnswer` rethrows the `INVALID_REQUEST` unwrapped (`src/agents/harness/gateway-question.ts:376`). At that layer this is intended: the question stays pending and a later valid answer resolves it (`gateway-question-dispatch.test.ts:619`).
- `runReplyQuestionInput` turns only `QuestionDispatchRefusedError` into the visible "The answer was not sent: ..." reply (`src/auto-reply/reply/agent-runner-question-input.ts:88`) and rethrows anything else (`:100-101`). Its caller does not catch it (`src/auto-reply/reply/agent-runner-run.ts:248`).
- The raw error fails the channel dispatch, and `src/channels/message/ingress-drain.ts:270` keeps the update for retry. Every retry fails the same way.

Line refs are `main` at `90e44fe1b00`.

### Steps to reproduce

1. Telegram DM with an agent that can call `ask_user`.
2. The agent asks two questions in one `ask_user` call. `ask-user-tool-normalization.ts:91` sets `isOther: true`, so each question ends "Other: reply with your own answer." and the prompt ends "Reply by number or question id."
3. Reply with one line of free text that answers question 1 only.

The prompt as sent, with question 1's wording replaced:

```
Question for you:

1. <q1 header>
<q1 text>
1. <option> - <description>
2. <option> - <description>
3. <option> - <description>
4. <option> - <description>
Other: reply with your own answer.

2. What to draf
What exactly should I show you as the draft before posting anything?
1. PR description only - ...
2. PR description + diff - ...
3. Issue first, then PR - ...
Other: reply with your own answer.

Reply by number or question id. Use a declared option where choices are fixed.
```

The reply was one line, 108 characters, free text answering question 1. I have left its text out because it names local hosts.

### Expected behavior

The user is told in the same chat that the answer was not accepted and which question still needs one, and the question stays open for a corrected reply.

### Actual behavior

No message reached the chat between the reply at 20:56:16 and the question timeout at 21:09:38. The update was attempted 10 times and then marked done with no recorded outcome. The agent's next message, at 21:09:47, begins "No answer came back, so I'll proceed with best judgment".

### OpenClaw version

2026.9.2 with local patches. None of them touch this path: `git diff v2026.9.2 <deployed sha>` over the seven files involved is empty. The same path is present on `main` at the lines cited above.

### Operating system

macOS 26.5.1, Apple silicon

### Install method

Source checkout built in place, gateway under launchd, node 26.4.0

### Model

Self-hosted open-weights model on an OpenAI-compatible endpoint (`openai-completions`). The failing path runs after the tool call and never reaches the model.

### Provider / routing chain

One custom OpenAI-compatible provider, no fallback involved.

### Logs

Gateway journal, 2026-09-10 UTC, ids redacted:

```
20:54:38  (transcript) agent calls ask_user: 2 questions, q1 (4 options) and scope (3 options)
20:56:16  ⇄ res ✗ question.resolve 0ms errorCode=INVALID_REQUEST errorMessage=question 'scope' requires an answer
20:56:16  telegram dispatch failed: GatewayClientRequestError: question 'scope' requires an answer
20:56:16  spooled update <id> failed; keeping for retry: question 'scope' requires an answer
          same update retried at 20:56:18, :20, :25, :33, :49, 20:57:22, 20:58:26, 21:00:34, 21:08:35
20:56:57  stalled session: ... reason=blocked_tool_call classification=blocked_tool_call activeWorkKind=tool_call
          (every 30 s through 21:09:27)
21:04:05  message dispatch completed: channel=telegram ... outcome=error duration=30053ms reason=question-response-refused
21:09:38  ⇄ res ✓ question.waitAnswer 899999ms
21:09:38  question message finalization failed id=ask_<id> delivery=telegram:default:<chat>:<msg>
21:11:35  telegram ingress: update <id> completed without a recorded processing outcome
21:14:03  visible channel turn dispatched with no queued reply payloads: channel=telegram messageId=<msg> ... cause=error:question-response-refused
```

The session transcript holds one user message between the `ask_user` call (20:54:38) and its timeout result (21:09:38): the 20:56:16 reply. The 10 `keeping for retry` records carry a single update id, and all 10 `message processed` records carry the same Telegram message id, so the ten attempts are one reply.

The source comment at `extensions/telegram/src/telegram-ingress-drain.ts:441` describes the 21:11:35 line as an update that "was consumed silently".

Tests on `main`:

- claim layer, a rejected answer leaves the question open: `src/agents/harness/gateway-question-dispatch.test.ts:619`
- dispatch layer, `question-response-refused` and `question-response-indeterminate` deliver a notice: `src/auto-reply/reply/dispatch-from-config.question-custody.test.ts:134`
- `INVALID_REQUEST` appears in one question-path test, the claim-layer case above. I found none that sends a rejected plain-text answer through `runReplyQuestionInput`.

### Impact and severity

Affected: Telegram DMs where the agent asks more than one question in a single `ask_user` call and the user answers in one line. Other channels share the path from `runReplyQuestionInput` on; I have only observed Telegram.
Severity: silent message loss. The answer is dropped with no reply, and the agent's question comes back unanswered.
Frequency: observed once, 2026-09-10. I have not measured how often it happens.
Consequence: the session sat on `blocked_tool_call` for the full 15-minute question timeout, then the agent went on without the user's answer.

### Additional information

Same symptom as #126834, closed not-planned on 2026-08-20 as not reproducible on `main`. Its close note asked that a renewed report include the exact prompt, reply text and journal. This one has the prompt, the reply's shape and the 2026.9.2 journal. I have not reproduced it live on current `main`, where the same code path is present.

Different from #140085, where the question never reaches Telegram. Here the prompt was delivered and the answer was lost.

ruel225:
## What Problem This Solves

When an agent asks two or more questions in one `ask_user` call and the user replies with one unkeyed line that answers only some of them, the reply is dropped with no visible outcome:

- `buildAgentHarnessUserInputAnswers` maps an unkeyed reply by line, so the unanswered question gets an empty answer (`src/agents/harness/user-input-bridge.ts`).
- `question.resolve` rejects the partial submission with `question '<id>' requires an answer` (`QUESTION_INVALID_ANSWER`), leaving the question pending.
- `claimPendingAgentQuestionAnswer` rethrows that rejection unwrapped so the question stays claimable (intended; covered by `gateway-question-dispatch.test.ts` "releases a rejected answer without waiting for the human question deadline").
- `runReplyQuestionInput` only converts `QuestionDispatchRefusedError` into a visible reply and rethrows everything else, so the raw rejection fails the whole channel dispatch.

The channel ingress then keeps the update for retry; every retry fails the same way until ingress gives up. The user hears nothing until the question deadline expires and the agent proceeds "as if nobody answered". Reported in #145126 with gateway journal evidence of 10 failed retries over 13 minutes.

## Solution

`runReplyQuestionInput` already owns the contract of turning recoverable question-flow refusals into a visible error reply for the same chat. This extends that contract to answer-validation rejections:

- `gateway-question.ts` exports `isQuestionAnswerValidationError`, a narrow check for the `QUESTION_INVALID_ANSWER` resolve-rejection reason (malformed or incomplete answers only; terminal and authority failures keep their existing paths).
- The check is a type predicate (`error is GatewayClientRequestError`): the runtime name marker it validates is exactly the marker of that class across the gateway protocol boundary, so the reply layer reads `error.message` through the validated error contract instead of an unchecked cast.
- The reply layer surfaces `The answer was not accepted: question '<id>' requires an answer. The question is still open, so reply again with answers for every question.`, marks the payload as an error, and records the same `question-response-refused` admission-skip the dispatch-refused path uses.

The rejection stays a raw transport error at the claim layer, so the pending question remains claimable and a corrected reply still resolves it. Terminal reasons (`QUESTION_NOT_FOUND`, `QUESTION_ALREADY_TERMINAL`), unconfirmed answers, and authority refusals are untouched.

## Impact

A partial unkeyed answer now produces an immediate visible chat reply naming the unanswered question, the question stays open for a corrected reply, and the ingress no longer retries a doomed dispatch.

## Evidence

New regression test `reports a rejected partial answer instead of failing the channel dispatch` (`agent-runner-steer-adoption.question-recovery.test.ts`):

- **Pre-fix** (production change reverted, test present): the raw rejection propagates out of `runReplyQuestionInput` and the test fails with `GatewayClientRequestError: question 'scope' requires an answer` — the exact dispatch failure from the report.
- **Post-fix**: the full file passes (20/20), including the new test asserting a handled result with an error payload naming `question 'scope' requires an answer`, `admission = { status: "skipped", reason: "question-response-refused" }`, no steer adoption, and the claim left non-resolving.
- `gateway-question-dispatch.test.ts` passes unchanged (42/42), confirming the claim layer still rethrows the raw rejection shape and the question stays pending for a corrected answer.
- **Typecheck** (predicate fix at `ab6359f1a4b`): `tsgo -p tsconfig.core.json` failed before the fix with `src/auto-reply/reply/agent-runner-question-input.ts(112,49): error TS18046: 'error' is of type 'unknown'` and passes after it; `test/tsconfig/tsconfig.test.root.json` and type-aware lint pass on the touched files too; `gateway-question.test.ts` + `gateway-question-dispatch.test.ts` (66/66) and the question-recovery file (20/20) stay green.

Closes #145126

### DIFF
--- a/src/agents/harness/gateway-question.ts
+++ b/src/agents/harness/gateway-question.ts
@@ -1,4 +1,5 @@
 import { randomBytes } from "node:crypto";
+import type { GatewayClientRequestError } from "../../../packages/gateway-client/src/request-error.js";
 import type {
   QuestionAnswers,
   QuestionRequestQuestion,
@@ -112,7 +113,11 @@ function isTerminalAgentQuestionError(error: unknown): boolean {
  * pending, so the same source can submit a corrected answer. Reply owners
  * surface this rejection instead of failing the whole channel dispatch.
  */
-export function isQuestionAnswerValidationError(error: unknown): boolean {
+export function isQuestionAnswerValidationError(
+  error: unknown,
+): error is GatewayClientRequestError {
+  // The name marker below is the runtime check for this class across the
+  // gateway protocol boundary, where instanceof is not reliable.
   return readQuestionRejection(error)?.reason === "QUESTION_INVALID_ANSWER";
 }
 
--- a/src/agents/harness/gateway-question.ts
+++ b/src/agents/harness/gateway-question.ts
@@ -107,6 +107,15 @@ function isTerminalAgentQuestionError(error: unknown): boolean {
   return reason !== undefined && TERMINAL_QUESTION_ERROR_REASONS.has(reason);
 }
 
+/**
+ * Resolve rejections for malformed or incomplete answers leave the question
+ * pending, so the same source can submit a corrected answer. Reply owners
+ * surface this rejection instead of failing the whole channel dispatch.
+ */
+export function isQuestionAnswerValidationError(error: unknown): boolean {
+  return readQuestionRejection(error)?.reason === "QUESTION_INVALID_ANSWER";
+}
+
 type QuestionInputAuthority = { kind: "run" | "source-bound"; assertCurrent: () => void };
 
 /** One reservation owns both dispatch refusal and the prompt's release notification. */
--- a/src/auto-reply/reply/agent-runner-question-input.ts
+++ b/src/auto-reply/reply/agent-runner-question-input.ts
@@ -2,7 +2,10 @@ import {
   QuestionAnswerUnconfirmedError,
   QuestionDispatchRefusedError,
 } from "../../agents/harness/gateway-question-dispatch.js";
-import { claimPendingAgentQuestionAnswerFromCaller } from "../../agents/harness/gateway-question.js";
+import {
+  claimPendingAgentQuestionAnswerFromCaller,
+  isQuestionAnswerValidationError,
+} from "../../agents/harness/gateway-question.js";
 import { logVerbose } from "../../globals.js";
 import { markReplyPayloadForSourceSuppressionDelivery } from "../reply-payload.js";
 import type { ReplyPayload } from "../types.js";
@@ -93,6 +96,20 @@ export async function runReplyQuestionInput(
         handled: true,
         payload: markReplyPayloadForSourceSuppressionDelivery({
           text: `The answer was not sent: ${error.message}. Use the question controls in the Control UI, or check the active run and your permissions before retrying.`,
+          isError: true,
+        }),
+      };
+    }
+    // A rejected partial answer leaves its question pending; a visible rejection
+    // lets the same chat retry instead of failing dispatch into ingress retries.
+    if (isQuestionAnswerValidationError(error)) {
+      if (state) {
+        state.admission = { status: "skipped", reason: "question-response-refused" };
+      }
+      return {
+        handled: true,
+        payload: markReplyPayloadForSourceSuppressionDelivery({
+          text: `The answer was not accepted: ${error.message}. The question is still open, so reply again with answers for every question.`,
           isError: true,
         }),
       };
--- a/src/auto-reply/reply/agent-runner-steer-adoption.question-recovery.test.ts
+++ b/src/auto-reply/reply/agent-runner-steer-adoption.question-recovery.test.ts
@@ -335,6 +335,77 @@ describe("question response custody through reply adoption", () => {
     },
   );
 
+  it("reports a rejected partial answer instead of failing the channel dispatch", async () => {
+    const key = "agent:main:question-partial-answer";
+    const run = createQueueTestRun({ prompt: text });
+    run.run.senderIsOwner = true;
+    run.run.traceAuthorized = true;
+    run.run.messageProvider = "webchat";
+    run.run.config = { tools: { toolsBySender: { "*": { allow: [] } } } };
+    await withQuestionCreator(key, run, async (operation) => {
+      const source = new AbortController();
+      const gatewayCall: AgentQuestionDispatcher = {
+        version: 2,
+        call: async ({ method, authority }) => {
+          expect(method).toBe("question.resolve");
+          expect(authority.kind).toBe("source-bound");
+          await Promise.resolve();
+          if (authority.kind === "source-bound") {
+            authority.assertCurrent();
+          }
+          // The wire rejection question.resolve returns for a question with no answer.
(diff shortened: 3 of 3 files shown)


ruel225:
@clawsweeper re-review

The [P2] type-narrowing finding is fixed at head `ab6359f1a4b`: `isQuestionAnswerValidationError` is now a type predicate (`error is GatewayClientRequestError`) — the runtime name marker it already validates is that class's marker across the gateway protocol boundary — so the reply layer reads `error.message` through the validated error contract instead of an unchecked cast.

Pre-fix reproduction on this machine: `tsgo -p tsconfig.core.json` failed with `src/auto-reply/reply/agent-runner-question-input.ts(112,49): error TS18046: 'error' is of type 'unknown'`; it passes after the fix, as do the test-root typecheck, type-aware lint on the touched files, `gateway-question.test.ts` + `gateway-question-dispatch.test.ts` (66/66), and the question-recovery file (20/20). The PR body's Evidence section now records the typecheck result.

PR #144252, opened 2026-09-10
refactor: keep plugin work and cleanup owned during retirement

steipete:
Related: #135599, #141280. The Control UI drain fix merged separately in #145118 (`b78d472e`).

## What Problem This Solves

Plugin preparation and registry replacement could dispose dependencies still used by requests, model helpers, or another Gateway. This foundation lets accepted work keep its supplying plugin instance through completion and cleanup, providing the prerequisite for restart-free plugin management.

## Why This Change Was Made

One managed instance owns its registered callbacks, runtime-store slots, captured source, and disposal. Registries and metadata caches retain their own dependencies; retirement stops new ordinary calls, while admitted consumers and surviving Gateways keep their exact owners. Idle cached publications release their holds when those owners retire. The separate CLI registry, setup-loader state, and runtime-boundary implementation are removed.

The design preserves synchronous registration and startup loading, including TypeScript/Jiti behavior and Bun's native loader semantics. It also carries the related CI repairs: runtime-dependent startup and package tests declare the canonical build prerequisite; SDK readiness covers the resolved relative imports used by compiled bundled plugins; and the now-merged Control UI prerequisite observes Gateway drain before sidecar teardown. The subsequent management cut supplies enable, disable, install, uninstall, and explicit reload orchestration. No operator option, SQLite schema/migration, or protocol-version change is introduced.

## User Impact

Requests and setup flows keep their plugin dependencies until their work and cleanup finish. Plugin authors can use `api.lifecycle.onDispose` and `api.lifecycle.signal`; the existing lifecycle contract remains supported. Cleanup is best effort, with visible failures and bounded disposer waits. Plugins remain responsible for unmanaged timers, listeners, sockets, watchers, and child processes; native objects and engine caches are not universally reclaimed. Fresh captured-source reload requires Node's synchronous module hooks, while Bun retains its native/Jiti startup behavior. See [plugin lifecycle and cleanup](https://docs.openclaw.ai/plugins/sdk-runtime#plugin-lifecycle-and-cleanup).

## Evidence

- Earlier [Linux built-Gateway startup/RPC/shutdown proof](https://github.com/openclaw/openclaw/actions/runs/34538816130) passed on tree `005e642e`, including a throwing cleanup callback and natural exit of both Gateways. Later source-bound regressions cover shared-Gateway retention, memory acquisition, captured-source loading and Bun startup. These historical results retain their original identities.
[...]

PR #145257, opened 2026-09-11
perf(models): reuse provider policies during catalog materialization

### LINKED ISSUE #145256: Catalog materialization repeats provider policy selection
Model catalog materialization repeatedly resolves the same provider policy while matching runtime capabilities, projecting rows and deduplicating static entries. Prepare that policy once per normalized provider within the existing synchronous materialization operation.

Preserve fresh identity calculation for every row, provider/model read order, original-ID fallback when canonicalization declines, case-sensitive identity tuples and configured-first static precedence. Cache missing policy surfaces only within that invocation; do not retain failed resolutions or policy selections across later calls.

This is a bounded allocation cleanup with no new global, inventory or snapshot cache. Its measurements include bundled-directory lookup work, so they overlap the separate directory-key optimization and must not be added together.

steipete:
Closes #145256

<details>

</details>

## What Problem This Solves

Model catalog materialization repeats provider-policy selection for rows sharing the same provider, doing unnecessary work while matching runtime capabilities and projecting the catalog.

## Why This Change Was Made

Resolve one nullable policy surface per normalized provider for each synchronous materialization call, then share it across runtime matching, row projection and static deduplication. Row identities are still calculated afresh; later calls select their current policy again, and thrown resolutions are not cached.

The current-main integration also moves the added test's type-only `PreparedRuntimeCapabilityModel` import from `prepared-model-runtime.configured.js` to its canonical owner, `prepared-model-runtime.types.js`. It retains the upstream type migration and runtime/setup policy owners without adding a compatibility export.

## User Impact

Model IDs, routes and metadata keep their existing behavior. The unary identity callback, provider/model read order, original-ID fallback, case-sensitive tuples and configured-first static precedence are preserved. There is no new inventory/global/snapshot cache, persisted configuration, updater marker or plugin-loading owner. The change remains scoped to three files and reuses an existing test fixture.

## Evidence

The first hosted integration run, [34646902083](https://github.com/openclaw/openclaw/actions/runs/34646902083/job/103420014819), failed with **TS2724** because the new installed-provider test imported a type from its former owner after #144252 moved that type. This was an actual integration failure, and its log is retained. The repair changes only that type import after composing main `b3059ab8f8383836dcd8221b2bbb750686e16d5b`; no runtime workaround, extra test or guard exception was introduced.

The composed source now published as `1322c19a95c0b5a0540e15bfe7c01b100430c21e` passes the owned plugins-platform typecheck, all **12 installed-provider tests**, formatting and a fresh independent P2 review. [CI run 34651097422](https://github.com/openclaw/openclaw/actions/runs/34651097422) completed successfully on this exact head, and its completed Claw review reports no findings or before-merge items. No broad build or allocation remeasurement was run for the type-only repair.

The following evidence remains **historical to the original frozen `ad439abc01eb575399fa2c9d0d892668ddb99403` comparison**, not newly measured on this integration: 33 owner cases passed on each source version, plus one selected unchanged runtime-refresh case per version; the full selected gate and normal build passed. The initial 721-versus-720 agents test-root failure was corrected by consolidating cases into the existing real installed-policy fixture without changing the limit.

Six fresh interleaved processes exercised the actual exported materializer with eight synthetic provider owners and real policy selection/loading. Each ran ten warmups and 80 measured calls, checking **30,720 output rows** with the same checksum. Median whole-interval sampled allocation, including collected objects and each raw profile node counted once, was **1,241,714,144 → 85,391,352 bytes (93.12% lower)**. Diagnostic elapsed medians were 5,388.65 → 267.65 ms and ending RSS medians 639,926,272 → 428,638,208 bytes.

Those timing/RSS observations include instrumentation and are not retained-heap, cold-start or whole-Gateway claims. The standalone comparison includes the bundled-directory work optimized separately; savings overlap and cannot be added or represented as incremental performance after that change. They are not restamped to the new merged source. The original keyless built Discord import profile compared the candidate with an empty worker, not an earlier revision or Gateway registration. Allocation proof predates the original test-only relocation; its production/probe bytes were preserved and the relocated scope received separate verification.


steipete:
Merged via squash.

- Prepared head SHA: [1322c19a95c0b5a0540e15bfe7c01b100430c21e](https://github.com/openclaw/openclaw/pull/145257/commits/1322c19a95c0b5a0540e15bfe7c01b100430c21e)
- Landed commit: [32d7b37c3452f3566d02a5759adc7a75339c1527](https://github.com/openclaw/openclaw/commit/32d7b37c3452f3566d02a5759adc7a75339c1527)

PR #145263, opened 2026-09-11
perf(tailscale): select primary addresses without unused collections

### LINKED ISSUE #145262: Primary tailnet lookups build unused address collections
Primary tailnet address lookups build and deduplicate lists of both IPv4 and IPv6 addresses even when the caller needs only the first matching address for one family. Repeated lookups therefore create unnecessary temporary collections.

Use the existing requested-family selector directly while preserving first-match order, address trimming, family normalization, internal-address filtering, fresh interface reads and original error propagation. A synthetic comparison of 2,000 picker calls produced identical results and 2,000 fresh synthetic interface-reader calls in each version, while Set constructions fell from 4,000 to zero. The existing 12 tests passed before and after; the full selected changed-file gate passed after correcting a proof-only Corepack routing failure.

The comparison establishes operation counts only. It reads no real network inventory and makes no timing, allocation-byte, RSS, or OS-enumeration improvement claim; each call still enumerates its fresh snapshot.

steipete:
Closes #145262

<details>

</details>

## What Problem This Solves

Primary tailnet address discovery does unnecessary collection work on every lookup, preparing both address families when the caller needs one primary address.

## Why This Change Was Made

Delegate each primary picker to the existing requested-family interface selector. This removes the private dual-family lists and deduplication path while preserving public signatures, CIDRs, first-match order, trimming, family normalization, internal-address filtering and original error propagation.

## User Impact

Address selection remains unchanged, with one fresh interface read per call and fewer temporary collections. The refactor removes 15 production lines; repository tests remain unchanged.

## Evidence

- The same four owner/helper/display/service-probe test files passed all 12 tests before and all 12 after the change.
- In 1,000 IPv4/IPv6 call pairs against a 512-entry synthetic snapshot, Set constructions fell from 4,000 to zero. Both versions made 2,000 fresh synthetic interface-reader calls and produced identical outputs. Boundary checks also preserved changed-snapshot freshness, empty/no-match results and exact thrown-error identity.
- The full selected changed-file gate passed formatting, core and core-test typechecks, dead-export scans, changed-file lint and selected contract guards. Scoped P2 review reported no findings.

The initial local gate stopped before running checks because the proof controller forced CI mode and selected unavailable Corepack. Removing that artifact-only override restored normal routing to installed pnpm; the second gate passed on the same reviewed source. The earlier tests and count comparison remained valid and were not rerun after this correction.

The comparison measures operation counts only. It does not establish timing, allocation-byte, RSS, OS-enumeration or whole-Gateway improvements; the selector still enumerates each fresh snapshot.

### DIFF
--- a/src/infra/tailnet.ts
+++ b/src/infra/tailnet.ts
@@ -1,13 +1,9 @@
 // Discovers local Tailscale tailnet addresses.
 import { isIpInCidr } from "@openclaw/net-policy/ip";
-import { uniqueStrings } from "@openclaw/normalization-core/string-normalization";
-import { listExternalInterfaceAddresses, readNetworkInterfaces } from "./network-interfaces.js";
-
-/** Tailnet addresses discovered on external local interfaces. */
-type TailnetAddresses = {
-  ipv4: string[];
-  ipv6: string[];
-};
+import {
+  pickMatchingExternalInterfaceAddress,
+  readNetworkInterfaces,
+} from "./network-interfaces.js";
 
 const TAILNET_IPV4_CIDR = "100.64.0.0/10";
 const TAILNET_IPV6_CIDR = "fd7a:115c:a1e0::/48";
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d40d1aa6be1506fc5e1389b62f8fb216f923a8c1](https://github.com/openclaw/openclaw/pull/145263/commits/d40d1aa6be1506fc5e1389b62f8fb216f923a8c1)
- Landed commit: [0780c7a5c6492d04aacf6f5bf99c401bd3a2df2e](https://github.com/openclaw/openclaw/commit/0780c7a5c6492d04aacf6f5bf99c401bd3a2df2e)

PR #145271, opened 2026-09-11
perf(gateway): avoid unused tool-event copies

### LINKED ISSUE #145270: Tool event delivery copies session and payload data unnecessarily
## Summary

Reduce redundant session and node-payload copies during Gateway tool-event handling while preserving the existing full session read and delivery behavior.

## Problem and proposed change

The verbosity lookup needs only synchronous read access to the selected session entry, but currently requests an owned clone. Tool-event handling also prepares stripped node payloads before reaching the branch that decides whether to send them.

Use the existing full-session reader with `clone: false` for this synchronous, non-mutating lookup, and create the stripped node copy only in the verbose-enabled send branch. Keep query/projection semantics, default owned-read isolation, live verbosity precedence, heartbeat suppression and Control UI output unchanged. This introduces no new cache or list projection.

## Evidence and tradeoffs

The same 493 focused tests pass before and after; the full selected core/core-test gate and final P2 review pass. The actual event handler and canonical session reader were also compared over isolated synthetic SQLite state with in-process sinks, preserving the declared serialized tool-event envelope and input values.

Saved-prompt scenarios show 8.5–17.1% lower sampled JavaScript allocation. Timing is mixed: the no-prompt median increases 4.45%, and full-verbose increases 10.92%. Process peak RSS increases by 22.06 MiB and includes imports/profiling. The no-prompt allocation result is effectively flat.

Each arm uses one process, with three 500-event timing batches per scenario and a separate sampled-allocation batch. Per-arm Node version/binary identity was not recorded. These observations do not establish a universal speedup, retained-memory or RSS improvement, or live channel/provider behavior.

steipete:
Closes #145270

<details>

</details>

## What Problem This Solves

Gateway tool-event handling can copy a full session entry just to read verbosity and prepare node output that will not be sent. Saved session prompts make that unnecessary copying more expensive in the measured workload.

## Why This Change Was Made

Borrow the existing full selected-session read synchronously with `clone: false`, and prepare stripped node payloads only inside their verbose-enabled send branch. The borrowed entry is not mutated; query, projection, cache and persistence semantics stay unchanged, as does the reader's default owned-copy behavior.

## User Impact

Tool cards and node delivery retain their existing behavior: live verbosity changes, session-versus-run precedence, heartbeat suppression, Control UI partial output, and stripped versus full node output remain covered. No new list projection or cache is introduced.

## Evidence

- The same 493 focused tests pass before and after, along with the full selected core/core-test gate and final P2 review.
- The actual event handler and canonical session reader run against isolated synthetic SQLite state with in-process sinks. The declared serialized tool-event envelope matches, input values remain unchanged, and default owned-read isolation remains covered. Dynamic session display timestamps and other snapshot metadata are outside the cross-process envelope hash.
- The final narrowed comparison shows 8.5–17.1% lower sampled JavaScript allocation in saved-prompt scenarios. Session prompt parsing volume is unchanged; the improvement does not come from a reduced session projection.
- Counter-results are retained: no-prompt timing rises 4.45%, full-verbose timing rises 10.92%, and process peak RSS rises 22.06 MiB, including imports/profiling. No-prompt sampled allocation is effectively flat.
- Each arm uses one process, with three unprofiled 500-event timing batches, a separate 500-event sampled-allocation batch, and a separate 10-event count batch per scenario. Per-arm Node version/binary identity was not recorded; three batches are not three independent process samples.

These scoped observations do not establish universal speed, retained-heap or RSS improvements. No live socket, channel or provider request was exercised. No separate build was run because the change adds no import, lazy-loading or artifact boundary; the full selected gate passed.

### DIFF
--- a/src/gateway/server-chat.agent-events.test.ts
+++ b/src/gateway/server-chat.agent-events.test.ts
@@ -5013,7 +5013,7 @@ describe("agent event handler", () => {
 
   it.each([
     {
-      name: "keeps tool output for Control UI recipients when verbose is on",
+      name: "keeps tool output only for Control UI recipients when verbose is on",
       runId: "run-tool-on",
       toolCallId: "t3",
       verboseLevel: "on",
@@ -5027,7 +5027,7 @@ describe("agent event handler", () => {
       partialResult: undefined,
     },
   ] as const)("$name", ({ runId, toolCallId, verboseLevel, partialResult }) => {
-    const { broadcastToConnIds, toolEventRecipients, handler } = createHarness({
+    const { broadcastToConnIds, nodeSendToSession, toolEventRecipients, handler } = createHarness({
       resolveSessionKeyForRun: () => "session-1",
     });
     const result = { content: [{ type: "text", text: "secret" }] };
@@ -5047,6 +5047,10 @@ describe("agent event handler", () => {
     };
     expect(payload.data?.result).toEqual(result);
     expect(payload.data?.partialResult).toEqual(partialResult);
+    const nodePayload = requireMockPayload(nodeSendToSession, 0, 2, "node tool output payload");
+    const nodeData = requireRecord(nodePayload.data, "node tool output data");
+    expect(nodeData.result).toEqual(verboseLevel === "full" ? result : undefined);
+    expect(nodeData.partialResult).toBeUndefined();
   });
 
   it("preserves sanitized outcome-unknown exec details for Control UI recipients", () => {
--- a/src/gateway/server-chat.ts
+++ b/src/gateway/server-chat.ts
@@ -1386,7 +1386,7 @@ export function createAgentEventHandler({
       return runVerbose ?? "off";
     }
     try {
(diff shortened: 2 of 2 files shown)


steipete:
Maintainer decision: proceed with the simpler copy-elimination change and accept the documented tradeoffs. It keeps the existing full selected-session read, runtime behavior, and authority checks; it adds no SQLite projection or cache. The borrowed entry is read synchronously without mutation or retention.

The scoped saved-prompt measurements report 8.5–17.1% lower sampled JavaScript allocation. I also accept the counter-results: effectively flat no-prompt allocation, 4.45% slower no-prompt timing, 10.92% slower full-verbose timing, and 22.06 MiB higher process peak RSS including imports/profiling. Each arm used one process, and per-arm Node version/binary identity was not recorded.

This decision accepts those limitations, without claiming universal speed, RSS, retained-memory, or live-transport improvement. The review's risk checkbox concerns that broader uncertainty; it does not identify an introduced correctness defect. No additional projection or measurement expansion is required for this accepted scope.

PR #145275, opened 2026-09-11
perf(ui): reuse browser tab selection within each pane render

### LINKED ISSUE #145274: Chat pane browser previews prepare the same selections twice
An active, presented chat pane prepares the latest browser-tab selection twice during one render: once for transcript thumbnails and again for the Browser panel's preferred tab. Both consumers derive the same selection from the full message and live tool-message history.

Prepare the selection once in the pane layout and pass it to both consumers. Preserve latest-successful ordering, history-before-live ordering, host/node/profile identity, URL-less tab actions, explicit older-card selection and the existing request/capture identity checks. Inactive or hidden transcript thumbnails must retain their existing behavior while Browser panel preference remains available.

The retained actual-pane comparison records one pane render with two selection preparations before and one afterward. Synthetic before/after captures are byte-identical and preserve active/inactive behavior. This is a scoped work-count result, not a timing, allocation-byte, RSS or retained-heap measurement; no new cache or retained state is needed.

steipete:
Closes #145274

<details>

</details>

## What Problem This Solves

Rendering an active chat pane repeats the work needed to choose the latest browser tabs for its transcript thumbnails and Browser panel. Both surfaces use the same conversation history, so the second preparation is unnecessary.

## Why This Change Was Made

The pane layout prepares one selection map from the full, unfiltered message and live tool-message history, then shares it with the Browser panel and active, presented transcript. This removes the projection's duplicate scan and the old activation flag while preserving value-based render invalidation; no cache or retained state is added.

## User Impact

Browser cards, tab preference and explicit selection behave as before. Latest-successful and history-before-live ordering, host/node/profile identity, URL-less tab actions and request/capture identity checks remain intact. Inactive transcript thumbnails relinquish ownership while retaining the panel's tab preference. The change removes four net production lines and adds one net test line; the existing hidden-preview assertions are unchanged.

## Evidence

A source-Vite Chromium comparison exercises the actual pane with an isolated synthetic mock Gateway. CDP precise function counters record one explicit pane render and **two selection preparations before, one afterward**. Both arms retain the latest managed card and let a newer URL-less node tab own panel preference. Making the pane inactive removes thumbnail ownership while preserving that preference; neither arm causes implicit tab focus or page errors.

The retained before/after synthetic captures are byte-identical. They establish visual preservation; the function counters establish the reduced work. The measurement covers one explicit pane update and makes no timing, allocation-byte, RSS or retained-heap claim. No operator data or live browser target was used.

All 64 existing owner tests pass across the selector, browser-card, transcript and panel-route suites. The production Control UI build and five bundled route/URL E2Es pass, along with the full selected changed gate and a fresh independent P2 review with no actionable findings. The source-Vite count comparison is separate from the bundled build/E2E evidence. Earlier attempts were stopped before execution by the proof-capacity guard; they were not failed tests.

### DIFF
--- a/ui/src/pages/chat/chat-pane-layout-render.ts
+++ b/ui/src/pages/chat/chat-pane-layout-render.ts
@@ -123,6 +123,7 @@ export abstract class ChatPaneLayoutRender extends ChatPaneBrowserAnnotationRend
         state.requestUpdate?.();
       }}
     ></openclaw-chat-outbox-recovery>`;
+    const latestBrowserTabs = latestBrowserTabCards(chatProps.messages, chatProps.toolMessages);
     const chat = renderChat({
       ...chatProps,
       pluginToolIcons: this.toolIcons.icons,
@@ -131,7 +132,7 @@ export abstract class ChatPaneLayoutRender extends ChatPaneBrowserAnnotationRend
         this.presented &&
         this.visuallyPresented &&
         isSidebarSlotVisible(sidebarLayout, "conversation"),
-      browserTabPreviewsActive: this.active && this.presented,
+      latestBrowserTabs: this.active && this.presented ? latestBrowserTabs : undefined,
       historyState: catalog ? undefined : state,
       header: nothing,
     });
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [80d7481093dfcf9cdc7d2d601e94dd393447d2aa](https://github.com/openclaw/openclaw/pull/145275/commits/80d7481093dfcf9cdc7d2d601e94dd393447d2aa)
- Landed commit: [645c3166fcc458065383a4e5a315ed7778bc9ba8](https://github.com/openclaw/openclaw/commit/645c3166fcc458065383a4e5a315ed7778bc9ba8)

PR #143166, opened 2026-09-09
refactor: share agent-turn execution test defaults

steipete:
Closes #143165

## What Problem This Solves

Agent-turn execution tests repeat twelve identical default fields across several execution scenarios. Keeping those copies aligned adds maintenance cost when the shared execution contract changes. This is part of the maintainer-requested code reduction campaign.

## Why This Change Was Made

A single no-argument test-support factory supplies the shared defaults at 26 test sites and in the existing complete-input constructor. Every call allocates fresh callbacks and a fresh pending-task Set. Scenario inputs, custom callbacks and captured sets, field order, media mock boundaries, cases, and assertions are preserved. The prior constructor adoption in #142962 remains intact.

## User Impact

No user-visible behavior change. This removes 276 test/support code lines (275 physical lines) across six files; production code is unchanged.

## Evidence

- Original and candidate: 144/144 cases passed across nine native-routed suites, retaining each file's exact case inventory and order.
- All 20 changed-file gates passed, including core-test types, lint, formatting, architecture guards, and dead-export scans.
- Independent P2 review completed with no findings.
- Expanding all 27 factory calls reproduces the original parsed tokens and property order across the six files.
- The first lint run found the support module over its existing line limit. Reusing the same defaults in its existing constructor removed the remaining duplication; final validation and review cover that refinement. No limit, assertion, or baseline was weakened.

This is local isolated test/type/static proof; no live Gateway was needed for a test-support-only change. The differing run times are not claimed as a performance improvement.

PR #145281, opened 2026-09-11
test: reuse agent-turn defaults with explicit overrides

steipete:
Related: #139428, #143166

## What Problem This Solves

Eleven agent-turn tests still repeat the same execution defaults around one scenario-specific value. This maintainer-requested cleanup completes those uses of the existing fixture owner while retaining the custom pending-task Sets, reset spy, and verbose mode.

## Why This Change Was Made

Replace the repeated blocks with `createAgentTurnExecutionDefaults()` and immediately override the original custom field. The helper, types, scenario inputs, assertions, session getters, and CLI-specific factories remain unchanged. All twelve property names and their insertion order are preserved; explicit Sets and the spy retain their exact references.

## User Impact

Test maintenance only: 108 fewer test lines across five files, with no production behavior change or runtime performance claim. At the five external-Set sites, the helper creates one unused empty default Set before the original Set overrides it.

## Evidence

- Complete five-file cohorts passed 75 tests before and after on Node 24.19.0 / pnpm 12.3.4: command events, conversation failures, message tools, results, and terminal failures.
- Exact source inversion reconstructed all five original files from the eleven changed blocks and two added imports. All other bytes and the existing helper are unchanged.
- Source-extracted factory checks passed all eleven sites for own-property presence/order/descriptors, custom reference identity, callback arity/results, and per-call freshness. This supplements the real cohort; it does not import the complete production module.
- The same complete 75-test cohort and mapped gate passed again on refreshed main `f36addc6f69` after its upstream plugin-lifecycle and dependency changes, using a fresh private install with the exact lockfile.
- The mapped `coreTests` gate passed, including the `messaging` test typecheck, guards, dead-export scans, and lint. Deslop and independent P0–P3 autoreview completed with no actionable findings.

The 75 reported-label multiset and per-file order match. File groups reordered and reporter labels may be abbreviated; no global ordering or full untruncated JSON identity claim is made. The static inverse independently supports unchanged cases and assertions.

### DIFF
--- a/src/auto-reply/reply/agent-runner-execution-command-events.test.ts
+++ b/src/auto-reply/reply/agent-runner-execution-command-events.test.ts
@@ -86,18 +86,8 @@ describe("executeAgentTurn: command events", () => {
         onPatchSummary,
       } satisfies GetReplyOptions,
       typingSignals: createMockTypingSignaler(),
-      blockReplyPipeline: null,
-      blockStreamingEnabled: false,
-      resolvedBlockStreamingBreak: "message_end",
-      applyReplyToMode: (payload) => payload,
-      shouldEmitToolResult: () => true,
-      shouldEmitToolOutput: () => false,
+      ...createAgentTurnExecutionDefaults(),
       pendingToolTasks,
-      resetSessionAfterRoleOrderingConflict: async () => false,
-      isHeartbeat: false,
-      sessionKey: "main",
-      getActiveSessionEntry: () => undefined,
-      resolvedVerboseLevel: "off",
     });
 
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [940fd40d05577993d7a7ac9be7ad944d360edf4d](https://github.com/openclaw/openclaw/pull/145281/commits/940fd40d05577993d7a7ac9be7ad944d360edf4d)
- Landed commit: [94eb3bec93604ef95918cfcf37f5bd237a5b1521](https://github.com/openclaw/openclaw/commit/94eb3bec93604ef95918cfcf37f5bd237a5b1521)

Issue #142586, opened 2026-09-08
[Bug]: 2026.9.3 Doctor detects orphan task_delivery_state foreign keys but provides no supported recovery path


GitHoubi:
### Bug type

Behavior bug that blocks migration

### Summary

While upgrading an existing installation from `2026.7.1-2` to `2026.9.3 (1391f7c)`, Doctor stopped because the shared SQLite database contained 18 orphan rows in this foreign-key relation:

```text
task_delivery_state -> task_runs
```

Doctor detected the violations but provided no supported repair, quarantine, or export path. Repeated Doctor runs stopped at the same point.

After an offline database backup, manual inspection confirmed that the referenced parent rows did not exist. Deleting only the confirmed orphan child rows in a transaction allowed the migration to continue.

Post-repair verification returned:

```text
PRAGMA foreign_key_check;
-- no rows

PRAGMA integrity_check;
-- ok
```

The exact process that originally created the orphan rows is unknown. This report does not claim that the entire database was corrupt.

### Environment

- OpenClaw: `2026.9.3 (1391f7c)`
- Previous version: `2026.7.1-2`
- Platform: Linux x64
- Node.js: `24.20.0`
- State backend: shared SQLite

### Observed sequence

1. Upgrade an existing installation to `2026.9.3`.
2. Stop the Gateway and run:

   ```text
   openclaw doctor --fix --allow-exec
   ```

3. Doctor detects foreign-key violations involving orphan `task_delivery_state` rows.
4. Doctor refuses to continue but provides no supported repair command.
5. Re-running Doctor produces the same blocker.
6. Back up the database and inspect the foreign-key violations manually.
7. Confirm that each affected child row references a nonexistent `task_runs` parent.
8. Delete only those confirmed orphan rows in a transaction.
9. Verify `foreign_key_check` and `integrity_check`.
10. Re-run Doctor and observe that it advances to the next migration stage.

### Expected behavior

Doctor should provide a safe, transactional recovery path for this known schema relation.

Depending on whether the child rows contain recoverable information, an acceptable implementation could:

- reconstruct valid parent rows when provenance is sufficient
- quarantine and export the orphan child rows before removal
- remove provably unrecoverable migration residue
- stop with an exact documented recovery command if automatic repair is unsafe

Any mutation should create a verified backup and perform a post-change foreign-key and integrity check.

### Actual behavior

Doctor identified the exact database consistency blocker but left the operator with no supported way to resolve it.

[...]

PR #145284, opened 2026-09-11
fix: recover orphan task deliveries during Doctor repair

fuller-stack-dev:
Related: #142586

## What Problem This Solves

Fixes an issue where users upgrading an existing installation would get stuck in Doctor when `task_delivery_state` contained rows whose `task_runs` parents no longer existed. Stable 2026.9.4 still refuses this condition. The original producer is unknown.

## Why This Change Was Made

Doctor's existing fenced repair transaction now preserves a verified, complete WAL-aware database copy and a lossless row export before removing confirmed orphan delivery rows. It reuses existing SQLite snapshot, private-directory, durability, and integrity mechanisms. Normal runtime and healthy-backup admission remain strict.

This is the narrowly authorized preservation-first recovery for #142586. Storage/migration owners confirmed its scope is independent of rollback work. It does not reconstruct tasks, replay deliveries, change schema versions, or remove recovery artifacts. Unknown schemas/triggers, unrelated FK damage, preservation failures, and failed final integrity checks still refuse and roll back.

## User Impact

Affected operators can complete supported Doctor recovery while retaining the original database and all recoverable delivery payload. Doctor reports the recovery directory. The recovery manifest records preservation, not a claim that the transaction committed. Recovery does not prove that the unknown orphan-producing sequence is prevented.

## Evidence

- Pinned main `58a6953549d011ffe44dcc3a7f3290b50e64b349` reproduces refusal; preservation tests fail before the change and pass after it using both current state and the released 2026.7.1-2 schema fixture.
- Seven preservation controls passed, including committed WAL-only rows, all 18 orphan payloads, exact large integers/raw JSON, healthy siblings, unknown schema/trigger and unrelated-FK refusal, export failure, and post-delete rollback.
- Shared-state suite: 196 passed, one existing skip. Task-store suite: 51 passed.
- Real process-death proof: SIGKILL after DELETE and before COMMIT retains original rows and recovery files; retry succeeds with clean integrity/FK checks.
- Independent review raised an inbound-cascade scenario. The actual caller disables FK actions during repair; four rollback controls, including `CASCADE` and `SET NULL`, pass with both original and dependent rows intact. The raw review was retained and this finding was adjudicated against the caller and executable controls.
- Rebased onto inspected main `9762b9ec1e44c3149bc828b7e2229862cd03cbd3` for relevant SQLite/Doctor/build changes.
- Final head `bbf6ddb5e846dd787b0a81af2bde2f5c49407f6f`: full package build passed, supported final runtime build passed, canonical workspace-bundled tarball passed package integrity/import-graph checks. Tarball SHA-256 `aa5f444537775fb9ec70920e16515c878173ae37e0b95b2aadc177fb2155d9ec`.
- Actual installed published `2026.7.1-2` / `0790d9f593ad30c940ed93b5872a8cf6d6f3cf8c` passed `tasks list` on the historical-schema input before injection of 18 orphans. Its unmodified `update --yes --no-restart --tag file:<candidate.tgz> --json` returned 0, installed the exact final head, and ran candidate Doctor successfully. Both healthy delivery rows stayed byte-for-byte equivalent; all 18 raw JSON/NUL and large-integer orphan payloads remain in the verified full backup and JSONL export. Full FK/integrity checks pass.
- Installed repeat Doctor returned 0 and created no extra recovery archive. The actual updated foreground Gateway reached ready; `/healthz`, `gateway health --json`, and `tasks list --json` passed. Only disposable prefixes/state were used, with no live deployment.
- Runtime fixture qualification: the upstream schema fixture deliberately contains synthetic rows, including two enum values invalid in both the tagged and current task parsers. The operator cell normalized only those synthetic task values before injection, then validated them through the actual old runtime. Separate preservation tests retain the raw fixture unchanged. Earlier packaging and test-input failures are retained, not counted as successful operator proof.
- Required changed-source checks completed through a retained passing prefix plus explicit 15-stage continuation, including core/core-test/scripts types, format, scoped lints, zero-unused-export scans and architecture guards. The interrupted aggregate check is not reported as passed. Both patch-caused CI failures were fixed with braces and exact registration in the existing Doctor SQLite maintenance group; compiled-JS equivalence confirms the braces change no runtime behavior.

## Known gap and coordinated follow-up

This PR repairs the Doctor owner. A separate update-admission lane is stacked on it: newer installed updaters can refuse these orphans while opening the update ledger, before Doctor or candidate code runs. That lane will reuse this recovery owner and retain the recovery-directory evidence in its ledger. It is not included here. An affected already-installed driver that refuses before candidate execution needs the corrective release installed manually, followed by `openclaw doctor --fix`, for its first hop. The original 2026.7.1-2 published-driver cell passed separately, as recorded above. Keep #142586 open until the coordinated admission gap is resolved.

### DIFF
--- a/docs/reference/database-schemas/integrity-and-recovery.md
+++ b/docs/reference/database-schemas/integrity-and-recovery.md
@@ -69,6 +69,34 @@ The heartbeat proves ownership, not migration progress. A live but stuck mainten
 
 `SQLite read-only worker` failures append `code` and numeric SQLite `errcode` diagnostics when the underlying error supplies valid values, including through a bounded cause chain. Report the full code suffix when investigating a failure. Snapshot and integrity-child timeout errors include the applied budget and source file size; snapshot timeouts report an unknown size if the source stat failed. Integrity-child timeouts also retain `lastObservedPhase`. A generic `disk I/O error` or `SQLITE_IOERR` alone does not prove the disk is full.
 
+### Doctor reports orphan task delivery rows
+
+If `foreign_key_check` names `task_delivery_state` referencing `task_runs`,
+stop the Gateway and run `openclaw doctor --fix`. Doctor can recover this known
+relation when the database is structurally intact and has no unrelated
+foreign-key violations or unrecognized delivery-table schema or triggers.
+
+Before removing any orphan rows, Doctor preserves a complete WAL-aware database
+copy and a lossless row export in a private `openclaw-task-delivery-recovery-*`
+directory beside the shared database. Its report names that directory. It contains:
+
+- `database.sqlite`: the pre-repair database, including the orphan rows.
+- `orphan-rows.jsonl`: the original delivery payload, with SQLite integers encoded
+  as decimal strings to avoid precision loss.
+- `manifest.json`: file hashes, row count, and preservation metadata.
+
+Doctor removes only delivery rows whose task no longer exists, within the existing
+repair transaction, and requires clean integrity and foreign-key checks before
+commit. It does not fabricate tasks or replay deliveries. If preservation or a
+later repair step fails, row removal rolls back and any recovery artifacts remain.
+A manifest proves preservation, not that the repair committed; rerun Doctor to
+verify completion.
+
+Keep these files private: they can contain delivery destinations and other user
+data. Retain them until the updated Gateway is verified and any needed payload
+has been recovered. Do not overwrite newer runtime state with this original
+copy. This recovery does not establish which writer created the orphan rows.
+
 ### Why you cannot go back after updating to 2026.7.2
 
 Every release through `v2026.7.1` used agent schema 1 and state schema 1. The 2026.7.2 release train (starting with `v2026.7.2-beta.1`) migrates your databases forward on first start. That migration is one-way: the data is rewritten into the newer schema, and installing an older OpenClaw afterwards does not undo it. The older build refuses to start with a `newer schema version` error that names the build that owns the database.
--- a/scripts/check-kysely-guardrails.mts
+++ b/scripts/check-kysely-guardrails.mts
@@ -127,6 +127,8 @@ const rawSqliteAllowPathGroups = {
     "src/infra/state-migrations.media-persistence.ts",
     "src/infra/state-migrations.transcript-directives-archives.ts",
     "src/infra/state-migrations.transcript-directives.ts",
+    // Doctor integrity PRAGMAs and lossless native 64-bit orphan-row preservation.
+    "src/state/openclaw-state-db-task-delivery-recovery.ts",
   ],
   "session entry cache connection-local validity counters": [
     "src/config/sessions/session-accessor.sqlite-entry-cache.ts",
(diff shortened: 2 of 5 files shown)


steipete:
Coordination from the update-reliability watch: this is the Doctor-repair half of #142586 and our lane on the same issue is now scoped to only the other half, stacked on this PR: on a fixed driver, `openclaw update` refuses at ledger admission (`update-command.ts` → `update-run-ledger.ts` → the existing-write integrity gate) before any Doctor runs, so the admission path needs to classify cascade-owned `task_delivery_state → task_runs` orphans as repairable and invoke your recovery owner (recording the recovery directory in the update ledger) instead of refusing. Nothing there changes your PR's semantics; ours depends on yours and lands after it. One note for your Known gaps: the installed 2026.9.4 driver refuses at admission before candidate code runs, so the first hop for affected operators is "install the corrective release manually, then `openclaw doctor --fix`".

fuller-stack-dev:
The Doctor half remains scoped here; I changed the issue linkage to Related so #142586 stays open for your admission follow-up. The supported entry is `repairOpenClawStateDatabaseSchema`, which owns the state-write/schema fences and the immediate transaction. Its changes include the retained recovery-directory path. The relation helper requires that transaction and disabled foreign-key actions; it must not be invoked standalone at ledger admission.

The exact-head native sibling lane now passes 27 shared-state read-only cases and eight Doctor/schema-preflight cases. Additional real-Doctor CASCADE and SET NULL controls preserve dependent rows and roll back, disproving the independent review's cascade premise. Package and published 2026.7.1-2 operator proof are still running; no prevention claim for already-installed 2026.9.4 drivers.

fuller-stack-dev:
@clawsweeper re-review

Published-updater proof is now complete for `bbf6ddb5e846dd787b0a81af2bde2f5c49407f6f`; the PR body contains the result and limits.

- Actual published 2026.7.1-2 passed task inspection before orphan injection, then its unmodified updater installed the final canonical tarball and ran Doctor, all exit0.
- Verified full backup and lossless export retain all18 orphan payloads, including raw JSON/NUL and9007199254740993. Both healthy delivery rows are unchanged; full FK/integrity checks pass.
- Repeat Doctor is idempotent. The actual updated Gateway reached ready, with successful health and task-list commands. No live installation was changed.
- The two CI defects are fixed. Final runtime/package integrity and all selected local gates pass. The three brace fixes are compiled-JS equivalent; the only additional guard change names this exact helper under existing Doctor SQLite maintenance.

The schema fixture's invalid synthetic task enums were corrected before injection and validated by the original runtime. Failed setup attempts remain retained; they are not claimed as product proof. This remains recovery, not prevention of the unknown producer. Issue142586 stays open for the separately owned newer-driver ledger-admission follow-up.

fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [bbf6ddb5e846dd787b0a81af2bde2f5c49407f6f](https://github.com/openclaw/openclaw/pull/145284/commits/bbf6ddb5e846dd787b0a81af2bde2f5c49407f6f)
- Landed commit: [6b8cbfa2b063981915fd94c3707f639835b32c79](https://github.com/openclaw/openclaw/commit/6b8cbfa2b063981915fd94c3707f639835b32c79)

PR #145291, opened 2026-09-11
feat(agents): retry cyber-refused embedded turns on Daybreak

VACInc:
## What Problem This Solves

Embedded OpenClaw turns could stop at a structured OpenAI cyber refusal even when an approved Daybreak route was available for permitted defensive work.

## User Impact

Eligible, replay-safe embedded turns can make one Daybreak retry without changing the selected session model. Strict model selections remain terminal. Generic refusal text, other providers, and biological/misalignment refusals do not enter this route.

**Accepted configuration decision:** keep the disclosed auto default, Blue target, and 10-minute target cooldown. The PR author explicitly accepted this default and upgrade tradeoff after discussing it. This is not entitlement-gated enablement: users without target access can incur an extra attempted request before the original refusal is returned. No provider access is granted, and shared credential health must remain usable.

## Why This Change Was Made

The existing embedded runner and auth-policy owners handle the retry. The original refusal is restored only when the retry made no committed progress; cancellation, terminal stops and committed-work failures retain their real execution facts. The separate Codex-plugin implementation remains unchanged.

## Evidence

- Rebased onto 2867b0326843ae33e571732713602ec2b78a19ca; candidate 414cee033ed2de0add7a8c0d9d04f751b75afcd5. The non-generated PR patch is unchanged. Config-documentation metadata was regenerated with its canonical generator after a main-line hash conflict; original authors and required human credit are preserved.
- The first refreshed CI run exposed an upstream upgrade-survivor assertion mismatch. The current base includes its already-merged fix, #146606 (01eb021ea972d25ac79202ee09699419514b074c); no test was weakened or skipped in this PR.
- Focused embedded-runner, refusal-transport and upstream regression validation: **93 tests passed** across six files.
- The complete repository-selected changed gate passed on this head, including production/test typechecks, core/script lint, generated config checks and safety guards.
- An inspected 29.56-second before/after regression recording is retained for the task handoff: the same three runner tests fail on the main baseline and pass on this head. This is a labeled test-console recording, not product-UI or live-provider proof. GitHub attachment upload was rejected by the endpoint; no hosted video attachment is claimed here.
- [Exact-head hosted CI](https://github.com/openclaw/openclaw/actions/runs/34731802696) is the authoritative hosted result; the local evidence above is separate.

### Live Gateway verification (2026-09-13)

- The operator's updater receipt applied #145291 at candidate `414cee033ed2de0add7a8c0d9d04f751b75afcd5` to base `e1d0d8da42866194e978c1c49a300db5d999a880`, captured and verified the staged source fingerprint, and transferred that build into the serving checkout. The running Gateway reported build `2026.9.4-e1d0d8da4286-2026-09-13T00-57-57.084Z`; systemd was `active/running` with zero restarts, authenticated RPC was `ok`, `/healthz` returned HTTP 200 in 1.06 ms, and `/readyz` returned HTTP 200 in 8.34 ms.
- A harmless direct target probe through the live Gateway returned the exact marker `DAYBREAK_OK`. The embedded-run result selected `openai/gpt-daybreak-blue-latest` as the winner route in **2.492 s**; CLI wall time was **3.988 s**. The terminal receipt reported `responseModel: gpt-5.6-sol` and `rerouted: true`, so this proves that the configured Daybreak route is accepted and answers on this account, not that the upstream service exposed a distinct backend model identity.
- Three isolated primary-route probes exercised the live embedded `openai/gpt-5.6-sol` path: a bounded local-only CTF request completed in **23.776 s**, a classifier-only marker request completed in **4.732 s** (**5.760 s** CLI wall), and a bounded defensive EDR request completed in **25.877 s** (**26.870 s** CLI wall). All three terminal receipts stayed on the selected model with `fallbackUsed: false`; the EDR turn produced an assistant-authored safety refusal, not a structured provider refusal.
- These are single live observations, not a latency benchmark. At the final readiness sample the Gateway still returned HTTP 200 but reported degraded event-loop/CPU telemetry (`delayP99Ms: 1458.6`, utilization `0.805`), so the turn timings include a loaded-host condition.

**Proof boundary:** the live deployment, normal embedded path, target-route reachability, health/readiness, and their observed latencies are proven. None of the safe live prompts produced the structured `provider_refusal` that triggers this PR, so the automatic primary-refusal-to-Daybreak transition is **not claimed as live-proven**. That transition, replay safety, failed-target restoration, strict-selection behavior, and auth isolation remain covered by the exact-head focused tests and CI above. No external channel delivery was requested.

Worked on by:
- @VACInc

<details>
<summary>Earlier implementation and verification notes — historical revisions</summary>

The notes below preserve the original author evidence. Their test counts and revision references are historical, not a claim about the current head.

### High Level TLDR

OpenAI sometimes refuses permitted defensive security work with a structured cyber-policy error. #143761 added an automatic one-shot retry on Daybreak Blue for that case, but only inside the Codex plugin. The embedded OpenClaw runtime, which serves ordinary chat/agent turns, still treated the same refusal as terminal: the user got a refusal even when a permitted model was configured and reachable.

This adds the equivalent policy to the embedded runtime. When a replay-safe embedded attempt reports an OpenAI `cyber` refusal, OpenClaw retries that turn once on `openai/gpt-daybreak-blue-latest`, without changing the selected session model and without sending ordinary provider failures down the same path.

### Root Cause

Two separate gaps, both in core rather than in the Codex plugin:

1. **The refusal never reached embedded run metadata.** `streamOpenAICodexResponses` in `packages/ai/src/providers/openai-chatgpt-responses.ts` normalized the transport error but did not emit a `provider_refusal` assistant diagnostic, so no downstream core consumer could distinguish a cyber-policy refusal from a generic transport failure. `prepareEmbeddedRunTerminal` correspondingly had no refusal field to populate on `EmbeddedAgentMeta`.

2. **Replay safety and refusal policy were both terminal by construction.**
   - `resolveReplayInvalidFlag` (`src/agents/embedded-agent-runner/run/incomplete-turn-resolution.ts:331`) marked every attempt with incomplete-turn text as `replayInvalid`. A refusal always produces incomplete-turn text, so a refused turn was always replay-invalid even when the attempt's own replay metadata proved no tool or delivery side effect had occurred.
   - `classifyEmbeddedAgentRunResultForModelFallback` returned `null` for any incomplete turn lacking `fallbackSafe`, before any refusal-specific classification could run.

The result: even with Daybreak in a configured fallback chain, a cyber refusal on an embedded turn ended the turn. This is the behavior described in #143761 for Codex, applied to the runtime that handles the majority of turns.

### Exact change

Transport (`packages/ai/src/providers/openai-chatgpt-responses.ts`):

- `readCodexProviderRefusal` emits a `provider_refusal` diagnostic with `{ provider: "openai", category }`. It reads two shapes, because the two surfaces do not agree:
  - **Responses-native** `code: "cyber_policy"`, which is what the ChatGPT Responses transport actually carries. This arrives on all three terminal paths: a non-OK HTTP body parsed by `parseErrorResponse` (`CodexApiError.code`, no payload), a streamed `error` event mapped by `extractCodexEventError` (`CodexApiError.code`), and a `response.failed` event normalized into `ResponsesStreamFailure.code`. Matches `CYBER_POLICY_ERROR_CODE` / `is_cyber_policy_error` in `openai/codex` (`codex-rs/codex-api`).
  - **App-server** `codexErrorInfo: "cyberPolicy"` / `"misalignmentPolicyViolation"`, top-level or nested under `error`, plus the biological-risk message prefix.
- `streamSimpleOpenAICodexResponses` delegates to the same function, so both entry paths are covered by one diagnostic point.

Refusal plumbing (core):

- `src/agents/embedded-agent-runner/run/provider-refusal.ts` (new) projects that diagnostic into `EmbeddedAgentMeta.providerRefusal`; `terminal-preparation.ts` populates it.
- `resolveReplayInvalidFlag` no longer marks an attempt replay-invalid solely because a provider refusal produced incomplete-turn text, when the attempt's own replay metadata already proves replay safety. Every other replay-invalid condition is unchanged.

Policy owner (`src/agents/embedded-agent-runner/embedded-cyber-failover.ts`, new, module-private):

- `isReplaySafeEmbeddedOpenAiCyberRefusal` gates on provider `openai`, harness `openclaw`, `replayInvalid !== true`, refusal provider `openai`, refusal category `cyber`.
- `classifyEmbeddedAgentRunResultForModelFallback` returns a dedicated `OPENAI_CYBER_POLICY_REFUSAL` classification with `preserveResultOnExhaustion`, ordered *before* the generic incomplete-turn rejection. The pre-existing incomplete-turn guard is preserved verbatim, just moved after it.
- `run-entry.ts` captures the refusal, resolves the target through the normal alias/manifest resolution path, skips the retry when the target equals the refused model or is in the auth cool-off cache, and restores the original refusal transcript when Daybreak fails. An auth failure on the target marks it skipped for `cooloffMs` (default 10 min) so repeated refusals do not re-probe an unauthorized model.

Config: `agents.defaults.embeddedAgent.cyberFailover` with `mode` (`auto` | `off`), `model`, `cooloffMs`. Default `auto` / `openai/gpt-daybreak-blue-latest` / `600000`, matching the Codex plugin default.

Refactor: `run-entry.ts` terminal-assembly helpers moved to `run-entry-terminal.ts` (239 lines) to stay under the max-lines ratchet; `run-entry.ts` net 688 → 618 lines. `EmbeddedAgentRunEntryTerminal` is re-exported from `run-entry.js` so existing importers are unaffected.

### Scope boundary

- Only OpenAI `cyber` refusals on the embedded OpenClaw harness. `bio` and `misalignment` are projected into metadata but deliberately **not** escalated.
- Generic refusal text without a structured category stays terminal.
- Codex-backed sessions keep their own plugin-owned policy from #143761; this does not change or duplicate it.
- The selected session model is never mutated. Ordinary provider failures still follow the existing fallback chain unchanged.
- On Daybreak failure the original refusal is preserved rather than cascading into unrelated configured fallbacks, **unless the retry already committed work or threw a non-failover error** (see below).
- A strict selection is never overridden. `run-embedded-attempt.ts` passes an explicit empty fallback override for a locked model selection; `isEmbeddedModelSelectionStrict` honors that and leaves the refusal terminal.

## Review rounds 2-5 (ClawSweeper)

Both P1s were real and are fixed in `727582a26a1`; the P2 compatibility question is resolved rather than deferred.

**P1 — Responses-native refusal codes were never recognized.** Confirmed against `openai/codex`: the Responses transport carries `error.code: "cyber_policy"`, not the app-server `codexErrorInfo` discriminator. The original reader required `CodexApiError` *with a payload*, so all three real terminal paths missed it: `parseErrorResponse` sets `code` but no payload, and `response.failed` throws `ResponsesStreamFailure`, which the guard rejected outright. The escalation gate was therefore unreachable in production. Now both error classes and both shapes are read, with one test per terminal path.

**P1 — A failed retry's execution facts were discarded.** Also real. A replay-safe *initial* refusal says nothing about the retry, which runs the same turn with tools enabled. If Daybreak executed a tool and then errored, restoring `originalFallbackResult` wholesale dropped the retry's `replayInvalid`, delivery evidence, and terminal receipt, and told recovery consumers nothing ran. `didEmbeddedCyberFailoverTargetCommitWork` now gates the restore on `replayInvalid === true || hasCommittedOutboundDeliveryEvidence(result)`; a retry that committed work keeps its own result, and only a demonstrably no-progress retry is replaced. The cool-off record is unconditional either way.

**P1 (round 3) — the thrown-exception path leaked the same defect.** Round 2 fixed the returned-result branch but left the `catch`, which restored the original refusal for every non-coordination error. That swallowed two cases that must never be replaced: a recorded terminal stop (`resolveModelFallbackError` → `kind: "terminal"`, replay prohibited) and an unclassified throw, which is exactly how `runWithModelFallback` reports an attempt that already committed work. The catch now propagates anything that is not `kind: "failover"`, which is the codebase's own definition of "this attempt is interchangeable and another model may serve the turn." Two parameterized regression tests cover both throws.

**P2 / maintainer decision — strict selections.** Taking the recommended option: **preserve strict selections.** A locked model selection is an explicit operator contract, and silently reaching another model by default is a compatibility break. Escalation is now skipped when `fallbacksOverride` is an explicit empty list, so upgrading cannot change behavior for a locked session and no operator needs the new `off` setting to retain what they already had. Operators who want escalation under a lock can unlock the selection.

**Round 4 - a delivered retry that then failed was still replaced (P1).** The round-3 guard keyed restoration on error class alone. That is not sufficient: `runWithModelFallback` rethrows a recognized provider error such as `overloaded` once `canFallbackAfterError` reports committed work (`model-fallback-runner.ts:586`), and that throw still resolves as `kind: "failover"`. A Daybreak retry that delivered its reply and only then hit an overload would have had its failure identity swapped for the initial refusal. Fixed by consulting the same live delivery evidence the runner itself used: the delivery predicate is now expressed as `hasCommittedSideEffect`, `canFallbackAfterError` is derived from it, and the catch restores the original refusal only when the error is failover-class **and** nothing was committed. Covered by a new test file that runs the **real** `runWithModelFallback` (`run-entry.cyber-failover.runner.test.ts`): one case delivers then throws `overloaded` and asserts the error propagates, one fails cleanly and asserts the refusal is restored. Verified the first case fails without the guard.

**Round 5 - auth-owner isolation proof (P1 validation gap).** The clean-failure regression now uses an `auth` entitlement denial rather than an overloaded transport error, runs it through the real `resolveAuthProfileFailureReason` and `markAuthProfileFailure` owners, asserts that no shared cooldown or disabled state is written, and then performs a second primary-model request with the same profile. The sequence passes for both an existing config that omits the new keys and an explicit `cyberFailover` config.

## Verification

Run in a worktree of this branch at `f1ad999bd12`.

### Real behavior proof

Automated, at the two boundaries that own the behavior:

`pnpm vitest run src/agents/embedded-agent-runner/result-fallback-classifier.test.ts src/agents/embedded-agent-runner/run-entry.cyber-failover.test.ts src/agents/embedded-agent-runner/run/incomplete-turn-resolution.test.ts packages/ai/src/providers/openai-chatgpt-responses-streaming.test.ts`

```
 ✓ run-entry.cyber-failover.test.ts > retries a replay-safe OpenAI cyber refusal once on Daybreak without changing selection 148ms
 ✓ run-entry.cyber-failover.test.ts > preserves the original refusal and cools off an unauthorized Daybreak target 137ms
 ✓ run-entry.cyber-failover.test.ts > keeps a failed Daybreak attempt that already committed work 89ms
 ✓ run-entry.cyber-failover.test.ts > propagates a recorded terminal stop thrown by the Daybreak retry 86ms
 ✓ run-entry.cyber-failover.test.ts > propagates an unclassified committed-work throw thrown by the Daybreak retry 119ms
 ✓ run-entry.cyber-failover.test.ts > keeps a cyber refusal terminal for a strict model selection 83ms
 ✓ run-entry.cyber-failover.test.ts > restores the original refusal transcript when Daybreak fails 108ms
 ✓ run-entry.cyber-failover.runner.test.ts > keeps primary auth usable after a denied Daybreak retry for an existing config without cyber failover keys
 ✓ run-entry.cyber-failover.runner.test.ts > keeps primary auth usable after a denied Daybreak retry for an explicit cyber failover config

 ✓ openai-chatgpt-responses-streaming.test.ts > projects a structured ChatGPT cyber-policy error as a provider refusal 2ms
 ✓ openai-chatgpt-responses-streaming.test.ts > projects a Responses-native cyber_policy error event as a provider refusal 1ms
 ✓ openai-chatgpt-responses-streaming.test.ts > projects a Responses-native cyber_policy response.failed event as a provider refusal 1ms
 ✓ openai-chatgpt-responses-streaming.test.ts > projects a Responses-native cyber_policy HTTP error body as a provider refusal 3ms
```

Whole-directory run covering every consumer of the changed core files:

`pnpm vitest run src/agents/embedded-agent-runner/ packages/ai/src/providers/openai-chatgpt-responses-streaming.test.ts`

```
 Test Files  195 passed (195)
      Tests  3511 passed (3511)
```

The `run-entry` tests drive the real `runEmbeddedAgentEntry` path end to end: refusal in, Daybreak attempt out, session selection asserted unchanged. The four transport tests assert the diagnostic is emitted from OpenAI's actual structured error payloads across every terminal path, rather than from a hand-built diagnostic, so the classifier gate is exercised against production-shaped input. The `incomplete-turn-resolution` test covers the replay-safety change directly, including that non-refusal incomplete turns stay replay-invalid.

### Other checks

| Check | Result |
| --- | --- |
| `pnpm tsgo:core` | pass |
| `pnpm lint:core` | pass |
| `pnpm format:check` | pass |
| `pnpm check:assertion-safety` | pass (3911 files) |
| `pnpm check:max-lines-ratchet` | pass (830 grandfathered) |
| `pnpm config:schema:check` | pass |
| `pnpm config:docs:check` | pass |
| `pnpm format:docs:check` | pass (1282 files) |
| `pnpm lint:docs` | pass (0 issues, 1281 files) |

`pnpm lint` (full aggregate) could not run in this environment: the tsdown declaration-boundary guard rejects any checkout nested inside an ancestor `node_modules` install, which is how the worktree is laid out here. This is an environment constraint, not a finding about the change; the individual lint/format/baseline jobs above were run instead and all pass.

### What was not tested

- **No live OpenAI cyber refusal was reproduced.** Deliberately not attempted: triggering a real cyber-policy refusal means sending probe traffic to a production OpenAI account, and the refusal payload shape is already fixed by #143761 and asserted here at the transport boundary. The escalation path is proven against that shape, not against a live refusal.
- No live Daybreak auth failure was sent to OpenAI. The synthetic entitlement denial now runs through the real auth-failure policy and usage owners, verifies that shared credential health remains untouched, and proves the next primary-model request succeeds; the target cool-off path remains covered by the existing `fallback-skip-cache` contract.
- The result-shaped committed-work test drives `runEmbeddedAgentEntry` with a mocked `runWithModelFallback`, not a live tool execution. It asserts the branch contract (`replayInvalid` and payloads survive); it does not prove a real tool's side effect end to end. The thrown-error committed-work path is covered against the real fallback runner.
- No performance benchmark. The added work is a metadata read on the terminal path and one conditional retry that only runs on a refusal that previously ended the turn; there is no new work on the success path. That is an inspection claim, not a measurement.

</details>

---
[View the OpenClaw team session](https://team.openclaw.ai/chat/roboclaw/~key/vito-vacinc-review-openclaw-openclaw-pr-145291-c67d4e4ea9492702)


VACInc:
@clawsweeper re-review

VACInc:
Rebased onto main at d093d651f71; head 63ba214ea9f. Preserved prior merge resolutions and upstream successful-tool accounting; regenerated config documentation baselines. Five focused runner, fallback, incomplete-turn, and terminal-preparation suites: 103 tests passed. Diff check passed. Core typecheck stops at the unchanged upstream src/gateway/server-startup-finish.ts:416 (GatewayPostReadySidecarHandle array mismatch); not a green full typecheck.

VACInc:
Follow-up PR #147146 is merged on current main. Successful interactive embedded cyber-policy retries now emit a source-conversation notice: `↪️ Retried on Daybreak`, or the configured `provider/model`. The retry remains limited to that turn and does not change the session’s selected model.

PR #145299, opened 2026-09-11
fix(cron): deliver overdue one-shot output

### LINKED ISSUE #131491: [Bug]: late-firing one-shot agentTurn runs execute fully, then 'skipping stale delivery' silently discards their output while recording success
### Bug type

### Beta release blocker

No

### Summary

On v2026.7.1-2 (0790d9f), a one-shot `agentTurn` cron job that fires long after its scheduled time **executes fully — model turns, tool calls, real side effects — and then its final output is silently discarded** by a delivery staleness guard:

```
[cron:<jobId>] skipping stale delivery scheduled at 2026-08-27T02:13:00.000Z, started 1076m late, current age 1083m
```

The run is then recorded as successful and the completed one-shot is removed. From the operator's perspective the task's *report* (frequently the entire point of an `agentTurn` job, e.g. "summarize X and reply in this thread") simply never arrives, with zero trace in job state, no failure alert, no annotation anywhere durable.

### Why this bites in practice

Late fires are not exotic: they are exactly what happens after recovering from the failure mode in the companion retry-classification report. The two defects chain into a data-loss pipeline:

1. Transient infra error fires while the gateway is unhealthy → one-shot silently disabled.
2. Operator notices days later, re-enables the job (past-due one-shots fire immediately on enable).
3. Job runs perfectly, consumes real model spend, performs its work.
4. Staleness guard discards the deliverable. Success is recorded.

We hit this exact chain: two revived jobs (an investigation task and a "report lane results into this thread" task) each ran to completion and delivered nothing; the only way we discovered it was reading raw gateway logs. Re-dispatching with a *fresh* `--at` time delivered fine, confirming the guard keys on scheduled-time age rather than anything about the content.

### Expected (any of these would resolve it)

- Deliver anyway with a "late" annotation (the guard's anti-confusion goal is met by labeling, not discarding), or
- Mark the run `undelivered` in job state and route it through the failure-notification path so the operator learns the output exists (e.g. in session history) but was not delivered, or at minimum
- Make the staleness threshold visible and configurable per job; some `agentTurn` payloads are only about their delivery.

### Steps to reproduce

1. Create a one-shot `agentTurn` job with `schedule {kind:"at"}` and any message that produces a reply.
2. Let its time pass by >~17h with the job unable to run (gateway down, or job disabled), then enable it.
3. Job fires immediately, completes; observe the `skipping stale delivery` line and the absence of the reply anywhere user-visible; job state shows a successful run.

Platform: Linux (Debian 12), node 22.23.2, gateway under systemd.

jason-allen-oneal:
Closes #131491

## What Problem This Solves

Fixes an issue where users with one-shot `agentTurn` automations would receive no report when the run started after the stale-delivery threshold, even though the run completed successfully and performed its work.

## Why This Change Was Made

Overdue one-shot jobs now continue through direct delivery instead of being discarded by the recurring-job staleness guard. The delivered report receives a clear late annotation so operators understand the timing; recurring schedules retain the existing stale-delivery protection.

## User Impact

Revived or delayed one-shot automations can deliver their completed output instead of silently recording success while losing the user-visible result.

## Real Behavior Proof

**HEAD:** `7a45afa9d54a910b109161fce125736c72802c2c`

**Claim:** A stale one-shot cron run still delivers its completed report and marks the delivery successful, with a visible late annotation.

**Exercised surface:** `dispatchCronDelivery` → direct one-shot stale-delivery branch in `src/cron/isolated-agent/delivery-dispatch.ts`. The focused regression invokes the production dispatch entry point and real payload normalization/session-cleanup path; its outbound adapter and gateway call are controlled test doubles, so this is controlled production-path evidence, not live-channel proof.

**Environment:** Linux x86_64; Node `v24.19.0`; pnpm workspace; isolated worktree `fix/cron-stale-delivery-131491` at the HEAD above.

**Scenario:**
1. Create an isolated one-shot `{kind: \"at\"}` cron job with a text report and direct delivery enabled.
2. Set the scheduled time three hours in the past while the run starts late.
3. Exercise `dispatchCronDelivery` through the stale-delivery branch.
4. Observe the outbound payload, delivery state, successful result, and delete-after-run cleanup.
5. Compare the same scenario on the unfixed parent and the final HEAD.

**Command:**

```console
node scripts/run-vitest.mjs src/cron/isolated-agent/delivery-dispatch.double-announce.test.ts -t 'delivers a stale one-shot'
```

**Relevant output:**

```text
Baseline c0df18eb55ecec92716ddc2d05d6f4092c1706cf: FAIL — received status=ok, delivered=false, deliveryError=\"skipping stale delivery scheduled at 2026-03-18T13:59:59.999Z, started 180m late, current age 180m\"; expected delivery.
Final 7a45afa9d54a910b109161fce125736c72802c2c: PASS — Test Files 1 passed; Tests 1 passed | 163 skipped.
```

**Artifact:** Inline output above; no generated proof files committed. GitHub Actions results for this exact HEAD will be linked here after completion.

**Observed result:** The unfixed code discards the overdue one-shot report. The final code delivers it once, appends \"Delivered late: ...\", records delivered=true, and performs normal cleanup.

**Not tested:** A live external channel, a real provider/model turn, long-running process timing, concurrent delivery races, Windows/macOS packaging, and the remote Crabbox environment. The controlled test intentionally uses adapter/gateway doubles; those limits remain separate from the production-path regression proof.

## Separate Review Blockers

- Live external-channel/Crabbox proof is not available locally and remains a coverage gap for ClawSweeper review.
- No correctness, security, ownership, or scope blocker is known from the current local review.

## Sources

- Current issue: https://github.com/openclaw/openclaw/issues/131491
- Published proof contract at pinned revision: https://github.com/openclaw/clawsweeper/blob/e74b07a40a6ecb1c9743e17201ecca0d6114bce5/AGENTS.md
- Actual review flow: https://docs.openclaw.ai/reference/pull-request-review-flow

## Completion Target

This PR is not considered done until the exact current head has green CI/CD,
ClawSweeper rates it platinum or better, and ClawSweeper marks it ready for
maintainer look.

No visual changes.

### DIFF
--- a/src/cron/isolated-agent/delivery-dispatch.double-announce.test.ts
+++ b/src/cron/isolated-agent/delivery-dispatch.double-announce.test.ts
@@ -2142,37 +2142,40 @@ describe("dispatchCronDelivery — double-announce guard", () => {
     );
   });
 
-  it("retains a stale one-shot transcript without delivery or a fallback summary", async () => {
+  it("delivers a stale one-shot with a late-delivery annotation", async () => {
     vi.useFakeTimers();
     vi.setSystemTime(new Date("2026-03-18T17:00:00.000Z"));
 
     const params = makeBaseParams({ synthesizedText: "Yesterday's morning briefing." });
     params.agentSessionKey = "agent:main:cron:test-job";
+    (params.job as { schedule?: { kind: "at"; at: string } }).schedule = {
+      kind: "at",
+      at: "2026-03-18T13:59:59.999Z",
+    };
     params.job.deleteAfterRun = true;
     params.beforeSessionDelete = vi.fn();
+    vi.mocked(deliverOutboundPayloads).mockResolvedValue([{ ok: true } as never]);
     (params.job as { state?: { nextRunAtMs?: number } }).state = {
       nextRunAtMs: Date.now() - (3 * 60 * 60_000 + 1),
     };
 
     const state = await dispatchCronDelivery(params);
 
-    const deliveryError = expect.stringContaining(
-      "scheduled at 2026-03-18T13:59:59.999Z, started 180m late",
+    expect(state.result).toBeUndefined();
+    expect(state.delivered).toBe(true);
+    expect(state.deliveryAttempted).toBe(true);
+    expect(state.deliveryError).toBeUndefined();
+    expect(deliverOutboundPayloads).toHaveBeenCalledOnce();
+    expect(deliverOutboundPayloads).toHaveBeenCalledWith(
+      expect.objectContaining({
+        payloads: [expect.objectContaining({ text: expect.stringContaining("Delivered late:") })],
+      }),
     );
-    expectResultFields(state.result, {
-      status: "ok",
-      delivered: false,
-      deliveryAttempted: true,
-      deliveryError,
-    });
-    expect(state.deliveryError).toEqual(deliveryError);
-    expect(deliverOutboundPayloads).not.toHaveBeenCalled();
-    expect(state.deliveryState.status).toBe("not-delivered");
-    expect(state.deliveryState.delivered).toBe(false);
-    expect(state.deliveryState.error).toEqual(deliveryError);
-    expect(state.deliveryState.deliverySuppressionReason).toBeUndefined();
-    expect(params.beforeSessionDelete).not.toHaveBeenCalled();
-    expect(callGateway).not.toHaveBeenCalled();
+    expect(state.deliveryState.status).toBe("delivered");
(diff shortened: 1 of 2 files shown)


jason-allen-oneal:
@clawsweeper re-review

jason-allen-oneal:
Closing this duplicate in favor of the canonical implementation PR https://github.com/openclaw/openclaw/pull/131603, which already covers #131491 with broader payload handling and current real-behavior proof.

PR #143410, opened 2026-09-09
feat(workers): reuse prepared public GitHub repositories

steipete:
Related: #134931. Builds on the landed preparation and ready-worker foundations in #143227 and #143372.

## What Problem This Solves

Repository-only cloud sessions repeat checkout and setup even when prepared capacity is available. They need reusable project state without creating a project clone on the Gateway.

## Why This Change Was Made

Admit the public GitHub repository instance, selected identity, pinned commit, and setup intent before selecting prepared capacity. The worker fetches public source anonymously, reuses completed preparation, and restores the session's accepted checkpoint, including deletions. This uses the existing preparation, ready-worker, and on-demand build owners. Public repository labels are derived for provider display and remain outside persisted source identity.

Public fetching and initial prepared adoption receive no GitHub token. Git/GitHub authentication variables are removed without regard to case, and the network fetch uses a temporary empty home with NETRC overrides removed because Git can consult .netrc independently of credential helpers. Gateway metadata requests may use the selected identity; later worker turns retain the existing per-turn credential contract. Explicit setup skip takes precedence over saved intent. Initially private repositories keep their existing cold checkout flow.

Admission and initial binding, including resumed provisioning, recheck source access. An already-active session keeps its owned checkout and accepted changes across Gateway restart; changing GitHub access does not revoke downloaded files.

The persisted addition is a repository-source variant in existing worker profile snapshot JSON, with no new configuration option or schema version in this PR. Existing local-project rows and checkpoint formats remain unchanged. Pre-feature schema-17 development builds cannot parse these prepared repository records, so same-state downgrade to them is unsupported. The released v2026.9.3 uses schema16: finish cloud cleanup with compatible code and restore the matching pre-upgrade backup when rolling back across that existing boundary.

## User Impact

Public GitHub repository sessions can reuse prepared checkout/runtime state and ready workers while keeping session changes isolated. Accepted changes survive worker replacement. Authentication failures remain visible, and visibility or identity changes prevent reuse of earlier prepared capacity.

Existing reserve controls and running-machine charges apply: the default is one unassigned worker per project/profile, with a shared cap of four. Zero reserve settings disable and drain unused reserves; existing warm-image controls retain the cold-checkout option.

## Evidence

Candidate: `69dffee085f01efa68a9e9e5cb2b5dc300c801c3`, based on `128108bb326d310f4cf21be1ee492a1451554d5b`.

[...]

PR #145302, opened 2026-09-11
fix(workers): reduce prepared repository startup delay

steipete:
Related: #134931, #143410

## What Problem This Solves

Users starting a session on an already prepared cloud worker still wait for separate remote calls that only configure the Git author name and email.

## Why This Change Was Made

Apply the author settings in the existing prepared repository branch-binding command. Prepared workspace startup now uses nine remote workspace commands instead of eleven, while retaining source verification, branch ownership, recovery, and checkpoint fences. Ordinary repository preparation keeps its existing author configuration path.

Compatibility: the added `author` member is transient stdin JSON sent together with its consuming `node -e` script in the same existing workspace-command request. It is not stored in a snapshot or database, and the generic command envelope already forwards optional string input to the process. The change writes the same repository-local `user.name` and `user.email` values through Git. It changes no schema, checkpoint format, prepared-workspace record, or wire-envelope version, so no migration is required. The prepared startup test restores an accepted checkpoint and reopens the bound worker state; the binding test also proves replay without a supplied author preserves session edits.

## User Impact

Prepared repository sessions avoid two sequential network round trips and two Node command startups. Author values, session edits, and prepared build output are preserved.

## Evidence

- 39 focused tests passed across repository preparation, startup/recovery, and synchronization. The extended real-Git regression failed on the original implementation and passed after the change.
- Changed-code checks passed, including core/test typechecking, lint, and repository guards. Independent review found no actionable P0–P2 issues.
- Twelve isolated runs using the real prepared Node workspace runtime, loopback transfer service, quiescence, and durable SQLite/Git checkpoint acceptance all passed with matching declared dependencies. Every baseline used 11 workspace commands; every candidate used 9.
- With an explicitly simulated 150 ms delay per remote call, prepared repository sync measured 740 ms before versus 277 ms after (medians of three samples). Two removed calls eliminate 300 ms of the imposed network delay. Host variance affects the remaining elapsed-time difference; these are local boundary measurements, not a new cloud dispatch benchmark.

Focused command: `node scripts/run-vitest.mjs src/gateway/worker-environments/node-worker-repository-preparation.test.ts src/gateway/worker-environments/repository-workspace-startup.test.ts src/gateway/worker-environments/node-worker-repository-sync.test.ts`.

### DIFF
--- a/src/gateway/worker-environments/node-worker-repository-preparation.test.ts
+++ b/src/gateway/worker-environments/node-worker-repository-preparation.test.ts
@@ -325,7 +325,12 @@ it("prepares and reuses an exact repository commit without a Gateway workspace",
     ),
   ).rejects.toThrow("session binding failed");
   expect(await git(remoteWorkspaceDir, "branch", "--show-current")).toBe("");
-  const bound = await repository.bindPreparedRepository({ ...source, branch }, preparedWorkspace);
+  const author = { name: 'Prepared "Name"', email: "prepared+session@example.invalid" };
+  const bound = await repository.bindPreparedRepository(
+    { ...source, branch },
+    preparedWorkspace,
+    author,
+  );
   expect(bound).toMatchObject({
     mode: "repository",
     baseCommit: commit,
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b4a83453ca3432fd34f4de7b8716b0f6231c8f51](https://github.com/openclaw/openclaw/pull/145302/commits/b4a83453ca3432fd34f4de7b8716b0f6231c8f51)
- Landed commit: [07e7e4b8a2b3481486c226a3ff2a673cbe62525b](https://github.com/openclaw/openclaw/commit/07e7e4b8a2b3481486c226a3ff2a673cbe62525b)

PR #145303, opened 2026-09-11
fix(ui): preserve plugin URLs on reload

steipete:
## What Problem This Solves

Fixes an issue where users reopening or reloading a plugin page such as `/reports` could return to their remembered chat or remain on the startup screen.

## Why This Change Was Made

Bootstrap now preserves explicit plugin-slug candidates while the Gateway supplies the available tabs. Initial not-found state reaches the existing router outlet recovery instead of being hidden behind the startup skeleton; remembered-session and first-run redirects no longer replace the requested plugin page.

## User Impact

Short plugin URLs work across cold starts and reloads. Unknown slugs retain the existing chat fallback after the Gateway responds, and legacy query-string URLs keep their page parameters and fragment when canonicalized.

## Evidence

- The existing cold-link browser cases failed against the original code with `/chat/main` instead of `/reports` or `/reports/` after adding remembered-session state.
- The updated production-bundle browser suite passes all 5 cases, including actual reloads, trailing-slash canonicalization, unknown-slug fallback, sidebar selection, and legacy-link state preservation.
- 214 focused bootstrap, shell, router, and model-setup tests pass.
- Full selected `check:changed` passes, including core-test/UI typechecks, lint, formatting, i18n, and dead-export checks. Independent autoreview is scoped-clean through P2.

Before: the held-connection regression had already changed the URL to the remembered chat while the startup surface was blank. The URL assertions above are the primary evidence.

Before: plugin URL lost during startup

After: the synthetic Reports page remains selected after a real browser reload with a remembered chat.

After: Reports remains active after reload

### DIFF
--- a/docs/web/urls.md
+++ b/docs/web/urls.md
@@ -479,6 +479,9 @@ for a tab with an available slug is replaced once in browser history with
 `/<slug>`, preserving `p.*` parameters and the fragment. Tabs without an available
 slug keep the generic URL. Both forms mount the same plugin page inside the
 Control UI shell; slugs do not create plugin HTTP routes.
+Opening or reloading a slug keeps that destination while the Gateway connects,
+even when the browser remembers a chat session. Unknown slugs fall back to chat
+after the Gateway supplies its plugin tabs.
 
 Automation links open the exact job independently of the current list filters or
 loaded page. Adding `run` opens its run history and highlights the matching loaded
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [43952d35c9ac815e8c4e6efd834bacfb5e4db77e](https://github.com/openclaw/openclaw/pull/145303/commits/43952d35c9ac815e8c4e6efd834bacfb5e4db77e)
- Landed commit: [95a7ec0ac2ed47df98a48211f3e0cdef09466536](https://github.com/openclaw/openclaw/commit/95a7ec0ac2ed47df98a48211f3e0cdef09466536)

PR #145065, opened 2026-09-11
feat(ui): open native CLI sessions in a main-area terminal route

steipete:
## What Problem This Solves

Starting a Codex or Claude Code session from the new-session page left the page on `/new` with the composer visible and opened the terminal in the bottom dock underneath. Sidebar rows with the "open in terminal" preference also went to the dock. Native CLI sessions did not behave like chat sessions.

## Why This Change Was Made

A normal-chrome `terminal` route hosts the existing terminal panel in the main content area: `/terminal`, `/terminal/<terminalSessionId>` to attach, and `/terminal?catalog=…&host=…&thread=…` to resume a native session. `/focus/terminal` is unchanged; `/terminal` is its normal-route counterpart.

A successful `sessions.catalog.startTerminal` replaces `/new` with `/terminal/<sessionId>`. The start RPC has one owner in `ui/src/lib/sessions/catalog-terminal-start.ts`, whose bounded in-memory handoff keeps output and exit events that arrive before the page mounts. Unclaimed handoffs close their PTY when they expire or are evicted.

Sidebar rows, the row menu, and the catalog view navigate to the route instead of dispatching to the dock. The dock's `catalog`/`catalogStart` toggle payloads were removed. Fresh catalog intents use non-persistent page queues. A restore-only decoder retains the shipped v2026.9.4 pending-action contract: already-saved catalog requests finish once through the existing dock owner and are removed on completion. Page instances own a non-persistent intent queue and a scoped sessionStorage namespace so dock and page never adopt each other's tabs. Leaving the page detaches rather than kills the PTY. The dock remains for ad-hoc operator shells.

## User Impact

Native CLI starts and terminal-preferred catalog entries now open in the main content area with normal app chrome. Existing terminal sessions can be attached through a direct URL and survive leaving the page.

This changes UI and the docs at `docs/web/urls.md` and `docs/web/control-ui/panels.md`; the `terminal` route segment is reserved in the session URL contract. There are no Gateway, protocol, or config changes. The startup JS baseline is re-based to **353694 B**, the exact rebased CI measurement, with explicit maintainer approval. CI exceeded the prior 353686 B growth-and-variance limit by 8 B; the local build measured 353669 B. The new baseline passes the local performance check. The fixed **358400 B** cap is retained. Route registration, the dynamic session-path bridge, and catalog navigation glue are eager by design; the page, resolver, handoff, and terminal runtime stay lazy.

## Evidence

Prior local proof on the original six-commit tree, before this rebase:

[...]

PR #145316, opened 2026-09-11
improve(chat): show how each provider is signed in inside the model picker

steipete:
## What Problem This Solves

The model picker lists providers and models but gives no hint of how each provider is signed in. Someone with a ChatGPT subscription and an OpenAI API key, or two subscriptions on the same provider, cannot tell from the picker which one a provider group is using, and has to open Models or the profile page to find out.

## Why This Change Was Made

Each provider heading now carries a small muted marker next to the provider name, in the same weight as the row metadata and with no badge or pill styling:

- A subscription provider shows a person icon and its plan name (or "Subscription") only when it has exactly one subscription profile, without an email. With multiple subscription profiles, the heading shows "Subscription" and appends an email only when the chat explicitly selects a matching profile and its email is available. Automatic selection and unmatched profiles show no account identity; provider-wide usage is never paired with another account's email. An explicitly selected API-key profile shows "API", including when the provider also has subscriptions.
- An API-key provider shows a key icon and "API".
- A provider with no working credential shows the existing sign-in warning on the heading only when its rows do not already carry one, so the state is never said twice.
- Inside the Account category, each account row gets the same email suffix under the same multi-subscription rule, replacing the raw profile id that was previously used to disambiguate same-label accounts.

The data comes from the chat page's existing `models.authStatus` snapshot, which the sidebar already refreshes, so the picker issues no new request. Emails are only present when the Gateway allowed profile identity for the caller, so privacy stays server-owned. New Session has no auth-status source and renders no marker.

## User Impact

Opening the picker shows at a glance whether a provider runs on a subscription or an API key, which plan, and, when it matters, which account. No new configuration.

## Evidence

Mock-Gateway capture with two OpenAI subscriptions (one explicitly selected for this chat), one Anthropic subscription, and a Google API key:

| Before                                                                                     | After                                                                                     |
| ------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------- |
| before | after |

Account category expanded (emails shown because OpenAI has two subscriptions):

after accounts

- Unit: chat-view model controls, chat-pane-session-controls, new-session model-control, web-awesome-migration inventory pass (462). New coverage: subscription with plan and email, single subscription without email, API key, no auth status, account-row emails, duplicate warning suppression.
- Control UI e2e: chat-model-controls (account emails, heading text, responsive widths), chat-flow.model-picker-refresh, model-catalog-partial-refresh, chat-composer-catalog, new-session model-catalog pass (41).
- `node scripts/check-changed.mjs` green; `pnpm ui:check-performance` green (startup JS 353,845 B gzip against a 354,270 B limit).
- Autoreview (Codex): clean, no accepted or actionable findings.

- Race fix: collapsing and expanding Account reuses its loaded inventory, so a page-one reload cannot silently consume the next Load more action.
- ClawSweeper P2s accepted: identity only from the explicit selection; plan only for a single subscription.
- ClawSweeper P2 (failed inventory not retryable) accepted: reopening the Account category retries an empty or failed load; regression added.
- ClawSweeper P2 (alias records overwrote a heading) accepted: records are merged under the heading key before labeling; regression added.
- ClawSweeper P2 (alias merge dropped an env/config API key) accepted: the merge keeps any alias's API key; helper and picker regressions added.
- Startup JS budget: main alone measures 354,388 B against a 354,270 B enforcement limit; this PR adds 119 B; the committed baseline is rebased to the measured 354,540 B with the 358,400 B cap unchanged (same shape as #145065).

### DIFF
--- a/config/control-ui-startup-budget-baseline.json
+++ b/config/control-ui-startup-budget-baseline.json
@@ -1,5 +1,5 @@
 {
-  "startupJsGzipBytes": 353694,
-  "reason": "Main-area /terminal route adds eager route registration, the dynamic session-path bridge, and catalog navigation glue for native CLI sessions; measured 353694 B startup gzip in CI on the rebased tree. Fixed 358400 B cap retained.",
+  "startupJsGzipBytes": 354540,
+  "reason": "Main crossed the startup enforcement limit after concurrent landings (main alone 354388 B vs 354270 B limit); this PR adds 119 B for provider auth markers in the model picker; measured 354540 B on the rebased tree. Fixed 358400 B cap retained.",
   "updatedAt": "2026-09-11"
 }
--- a/docs/web/control-ui/chat.md
+++ b/docs/web/control-ui/chat.md
@@ -33,7 +33,7 @@ navigable without a card; links to other origins keep normal browser behavior.
 Document-relative hrefs are never session links; file references such as
 `src/utils/foo.ts` retain workspace file handling.
 
-The chat model picker shows each provider’s subscription plan (or **Subscription**) or **API** beside its name when authentication status is available. With multiple subscription accounts, the heading and **Account** rows also show the account email when the Gateway supplies it. Hover over a truncated heading to read the full plan and email.
+When authentication status is available, each provider heading in the chat model picker says how that provider is signed in: **API** for an API key (or an explicitly selected API-key account), the plan name for a provider with one subscription, and **Subscription** for a provider with several. With several subscriptions, the heading adds the email of an explicitly selected account when the Gateway supplies it, and the **Account** rows show each account's email; automatic selection shows no account identity. Hover a truncated heading to read the full text.
 
 ## Composer capability menu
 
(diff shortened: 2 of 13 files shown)


steipete:
Both ClawSweeper findings are accepted. Commit 5fb098ea2aa makes account identity come only from the explicit selection and shows the plan only when the provider has a single subscription.

steipete:
The ClawSweeper finding that a failed account inventory could not be retried by reopening is accepted. Commit b2984e107cd lets reopening Account retry an empty or failed load while retaining successfully loaded accounts, with reject → reopen → success regression coverage.

steipete:
The ClawSweeper finding that alias auth records could overwrite a provider heading is accepted. Commit 94958c7f226 merges records under the heading key before labeling and adds regression coverage for explicit API-key and automatic subscription selection across Google aliases.

steipete:
The ClawSweeper finding that alias merging dropped an environment or configuration API key is accepted. Commit 57f2cd1e0fc preserves any alias's API key in the shared merge owner, with helper and picker regressions covering a Google environment key alongside a missing CLI alias.

steipete:
Merged via squash.

- Prepared head SHA: [3a2df480c8a15fa3f911136af528c45c24058475](https://github.com/openclaw/openclaw/pull/145316/commits/3a2df480c8a15fa3f911136af528c45c24058475)
- Landed commit: [a0bd289bfa07ac289467bd8a87cacd5eeb6bbd46](https://github.com/openclaw/openclaw/commit/a0bd289bfa07ac289467bd8a87cacd5eeb6bbd46)

PR #145124, opened 2026-09-11
chore(deps): refresh seven-day-cooled npm dependencies

steipete:
Closes #145096

## What Problem This Solves

Refreshes OpenClaw's npm dependencies with upstream fixes that have completed the requested seven-day cooldown. The frozen publication cutoff is **2026-09-04T16:49:00Z**. Existing newer security/runtime pins remain intact; no cooldown exception is added.

## Why This Change Was Made

Updates 64 direct package targets across 53 manifests and six pinned transitive targets, including the OpenAI, Anthropic, Gemini, AWS, OpenTelemetry, terminal, mail, Matrix crypto, LanceDB, Slack, Teams, and UI/tooling families. All 164 newly resolved versions have verified registry publication timestamps and matching registry/lockfile integrity hashes.

Deduplicates 26 obsolete versions so CodeMirror's editor and extensions share one state instance. Removes 13 expired release-age exclusions and two Mailparser overrides now owned by its fixed direct dependencies; retains Mailauth's Nodemailer and Undici fixes under its updated parent. Synchronizes Claude ACP fallback assertions and its skill pin, TypeBox plugin examples, generated Workboard asset references, overlay fixtures, and Diffs renderer type parameters.

TypeBox now emits redundant false-schema child errors alongside additional-property errors. The existing shared JSON-schema owner normalizes complete, matching child groups immediately before their aggregate for Gateway and plugin validators. It preserves genuine false-schema failures, property-name constraints, bounded lists without an aggregate, and original error order/objects. Validation decisions are unchanged. A reproduced raw-path collision regression is covered without dropping genuine errors, and each consecutive child group is scanned at most once. The existing Gateway rejection assertions remain intact; the truncation fixture now uses a genuinely oversized property name instead of depending on the previous diagnostic aggregation.

Intentional holds remain: direct Zod 4.4.3; the exact WebAwesome, Matrix SDK, noVNC, and Vitest patches; tsdown's declaration-safety override; separately managed Codex/CUA runtimes; and major-version migrations. Standalone deployment tooling and native app dependency tracks are outside this npm workspace refresh.

## User Impact

Users receive cooled upstream library fixes while plugin, streaming, terminal, file-editor, and invalid-request behavior remain supported. No new configuration, protocol version, or OpenClaw release version is introduced.

## Evidence

- Frozen installation passed with pnpm 12.3.4. All 164 new resolved versions passed the frozen timestamp and integrity audit; retained versions kept their integrity. Deduplication introduced no new versions.
[...]

PR #145320, opened 2026-09-11
fix(update): recover an unusable handoff store instead of refusing it

steipete:
## Problem

Coordination state for updates lives in a shared temp directory, so anything that landed on the lease store ended config mutation and native service operations for **every install root on the host** — permanently, with no in-product recovery. `openclaw doctor` cannot help either: its repair path writes config and dies on the same throw.

#145085 fixed the common case (a store left world-readable by an interrupted first write at `umask 022`). It left the rest:

| store state | before |
|---|---|
| `0664` interrupted write under `umask 002` | refused forever |
| `0666` world-writable | refused forever |
| symlink swapped in | refused forever |
| hard-linked store | refused forever |
| foreign owner | refused forever |
| directory mode drift | refused forever |

None of those bytes were ever ours to adopt — but refusing them protects nothing once the host can no longer write config at all.

## Fix

Retain the store under a name recording the defect, provision an empty one in its place via the existing `createMissingDatabaseFile`, and continue. Read-bit drift stays an in-place repair, because the store sets 0600 after every write and keeps its directory at 0700, so those bits are its own interrupted work.

Recovery is serialized and **the decision is retaken under the lock**. Acting on an observation taken before another repairer's rename would retain the clean store it just installed, leaving two authoritative databases and defeating the very lock this store provides. That race was found by review, not by me.

## Deliberately still refusing

A record this build cannot decode, a corrupt database, and a wrong table shape. Unlike a path we can identify as foreign, those may name a source they currently hold, and the daemon suite pins that `store.release()` stays `false` so a stale source is not released. `assertPath` also stays strict at `0o077` because it backs pinned-update identity capture, where a non-private store is a revoked authority rather than drift. I regressed that mid-change and an existing test caught it.

The remaining dead end is therefore narrow and not umask-driven. Closing it properly means an operator-facing repair, which is the gap I keep hitting: the refusal advises `openclaw triage`, but triage only touches the lease at its own install root and cannot inspect or clear the row blocking a different one, and nothing in Doctor inspects this store at all.

## Evidence

Live on Linux (Docker, `umask 0022`) through the real installed CLI, `openclaw config unset`:

```
0644 interrupted write (umask 022) exit=0 mode=600  retained=0 config-written=YES
0664 group-writable (umask 002)    exit=0 mode=600  retained=1 config-written=YES
0666 world-writable                exit=0 mode=600  retained=1 config-written=YES
0640 group-readable                exit=0 mode=600  retained=0 config-written=YES
symlink swapped in                 exit=0 mode=600  retained=1 config-written=YES
hard-linked store                  exit=0 mode=600  retained=1 config-written=YES
```

`retained=0` is repaired in place; `retained=1` is retained for diagnosis and replaced. Every case writes config.

Local: 54 + 24 across the handoff, custody, database-authority and daemon suites; `node scripts/check-changed.mjs` passes. One Codex round, whose P1 on the repairer race is fixed above.

A note on test honesty: the cross-process interleaving is not reproduced in a unit test. I wrote one that claimed to and it passed with the fix reverted, so I replaced it with a claim that holds — convergence on one live store with the repair lock released.

## Also carries an unrelated red-main repair

`cd6fe8252e5` (#145124, dependency refresh) took pi-tui to 0.84.4, which removed `getBounds` from `OverlayHandle`. Eight overlay fixtures described a property the type no longer has, so every pull request merging main failed core test typechecking on TUI files it never touched. Nothing reads the accessor in production and the package no longer declares it, so the fixtures stop declaring it; 338 TUI tests pass. Kept as its own commit.


steipete:
Correction: I removed the unrelated TUI commit from this branch. It was wrong.

I read a local core-test typecheck failure on `OverlayHandle.getBounds` as a red main from the #145124 dependency refresh, and "fixed" it by deleting the property from eight overlay fixtures. CI then reported the exact inverse — `getBounds` is **required** and now missing.

The cause was my own environment: this checkout had `@earendil-works/pi-tui` **0.84.4** installed while `package.json` and `pnpm-lock.yaml` both specify **0.85.0**. The refresh *added* the accessor; my stale install lacked it, so the local typecheck was reporting on a dependency tree that does not match the lockfile.

Main was never red for this. The commit is reverted and dropped, and this branch is back to the handoff-store change alone.

PR #145324, opened 2026-09-11
refactor(acp): share labeled parented-session test fixtures

steipete:
Related: #139428

## What Problem This Solves

ACP turn-result tests repeat the same child/parent session metadata setup in nine scenarios, differing only in the child label.

## Why This Change Was Made

Reuse the existing parented-session test helper with an optional label. Labeled entries keep their original field order; callers without a label continue to omit that property. Each lookup still creates fresh metadata and reads its timestamps at lookup time. Runtime mocks, turn results, settlement, and assertions are unchanged.

## User Impact

No user-visible behavior change. This removes 196 net test/support lines while retaining completion classification, task labels, and stream-cleanup coverage.

## Evidence

The full turn-results suite and all existing helper-consumer suites passed before and after on Node 24.19: 61 tests in each run, with identical ordered case names and results. An exact inverse check verified changes are confined to nine callback replacements and two helper lines. An actual-source fixture comparison verified values, keys, property order/descriptors, fresh allocations, lookup-time clock calls, and omitted labels for existing callers. Formatting, the mapped core-test gate, and independent P2 review passed.

### DIFF
--- a/src/acp/control-plane/manager.test-helpers.ts
+++ b/src/acp/control-plane/manager.test-helpers.ts
@@ -221,6 +221,7 @@ export function readySessionMeta(overrides: Partial<SessionAcpMeta> = {}): Sessi
 export function mockParentedAcpSessionEntries(params: {
   childSessionKey: string;
   parentSessionKey: string;
+  label?: string;
 }): void {
   hoisted.readAcpSessionEntryMock.mockImplementation((input: unknown) => {
     const sessionKey = (input as { sessionKey?: string }).sessionKey;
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6c52c4d43254166f201b7569c9d350bbc67742e7](https://github.com/openclaw/openclaw/pull/145324/commits/6c52c4d43254166f201b7569c9d350bbc67742e7)
- Landed commit: [0f3be2ac677289b60630baaf54fff69cc3b23c1a](https://github.com/openclaw/openclaw/commit/0f3be2ac677289b60630baaf54fff69cc3b23c1a)

PR #145327, opened 2026-09-11
fix: attribute Codex usage to OpenClaw tasks

pash-openai:
Codex usage reports could not identify OpenClaw task threads or tell what started their turns, even though OpenClaw already records the run trigger.

Mark new task threads with `threadSource: "openclaw"` and pass the existing run trigger to `turn/start`. Resuming a thread preserves its original source, and callers without a known trigger leave it unspecified. This only adds analytics metadata; it does not change scheduling, models, permissions, or message delivery.

### DIFF
--- a/extensions/codex/src/app-server/protocol.ts
+++ b/extensions/codex/src/app-server/protocol.ts
@@ -169,6 +169,7 @@ export type CodexTurnEnvironmentParams = JsonObject & {
 };
 
 export type CodexThreadStartParams = JsonObject & {
+  threadSource?: string | null;
   input?: CodexUserInput[];
   cwd?: string;
   projectId?: string | null;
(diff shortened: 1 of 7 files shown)


pash-openai:
The candidate was exercised through real Slack conversations on separate staging personal and team gateways using Codex 0.153.1. Both new native threads reported `threadSource: "openclaw"`; both bots delivered their replies, and a follow-up retained the same native thread and source. Existing model and approval settings were preserved. The change does not touch stored-state schemas, so rollback retains the existing state format.

All 204 lifecycle tests passed, including the six existing trigger values and an omitted-trigger case. Type-aware lint, formatting, and request snapshots also passed. The release-branch dependency audit still fails on the unchanged nodemailer and sharp versions; this patch changes neither dependencies nor the lockfile.

The implementation adds four production lines for the two optional request fields and their values, with 37 regression-test lines plus regenerated request-envelope fixtures. No model prompt text changes.

PR #145282, opened 2026-09-11
fix(update): preserve macOS launcher permissions during updates

PollyBot13:
Related: #145072

## What Problem This Solves

Fixes an issue where macOS users retrying an npm update receive `Package rollback launcher backup changed` when the updater's umask differs from the installed launcher symlink's permissions.

## Why This Change Was Made

Preserve the symlink's mode with macOS `lchmod` before the existing atomic publication. Keep launcher fingerprints, ownership checks, and failure cleanup unchanged; never follow the link or change its target's permissions. The same copy owner serves backup and retained rollback.

## User Impact

An updater containing this fix can copy and restore launchers created under either restrictive or ordinary umasks. Linux and Windows behavior is unchanged.

**First-upgrade limitation:** this is prevention in the fixed driver, not a rescue for an already-running 2026.9.3/9.4 updater. The published 2026.9.3 driver owns the faulty copy and rejects its backup before activation; putting this fix only in its candidate does not replace that driver. This PR therefore references rather than closes #145072. Existing-install recovery/release delivery remains separate.

## Evidence

- macOS arm64, Node 26.8.1. Five new regressions fail on the original implementation; all 66 tests pass across `package-update-filesystem.test.ts`, `package-update-swap.integrity.test.ts`, `package-update-swap.ownership.test.ts`, and `package-update-swap.bounds.test.ts` using `node scripts/run-vitest.mjs`.
- Separate unmocked native filesystem execution of the actual copy and swap owners against disposable npm-layout fixtures (no real Gateway or npm install):
  - source umask `077`, updater `022`: original `0700`, bare copy `0755`, fixed copy/restored launcher `0700`.
  - source umask `022`, updater `077`: original `0755`, bare copy `0700`, fixed copy/restored launcher `0755`.
  - Both activate, then retained rollback returns exit 0. Link target, mode, uid/gid, target contents and target permissions remain unchanged.
- Regression coverage also includes dangling links and metadata failure leaving the live launcher intact with no staging residue.
- After integration with main, all **97 tests** pass across the four launcher/filesystem/swap suites (66) and the first-hop fixture suite (31). Formatting and `git diff --check` pass. Fresh native autoreview of the combined diff is scoped-clean at P1, with no actionable findings.
[...]

PR #145219, opened 2026-09-11
fix(update): derive update budgets from measured state instead of fixed literals

steipete:
Refs #144758 #144901 #144890 #143292 #146637 #145252. Reconciles the landed activation boundary from #147019.

## What Problem This Solves

Fixes update steps that can expire before their owning operation's allowance on slow hardware or networks. A throttled run completed six large snapshots and then failed a native capability probe that still had a hidden 30-second limit.

## User Impact

State copies, candidate startup, service work, and network operations receive their owning allowances. Expiry still cancels authority and preserves uncertain writer ownership. No new configuration, dependencies, schema, retry policy, or retention changes.

## Why This Change Was Made

SQLite bytes, copy progress, observed startup, and plugin count determine budgets; callers forward the resolved allowances. Metadata/network operations share a 300-second default. Main's `update-activation-timeout` settlement and status/Doctor guidance remain authoritative: expiry alone never authorizes rollback, restart, or lease release. The Gateway can remain stopped while ownership is unresolved.

## ClawSweeper review

Reviewed head: `e9fe0f4088204ac699be7c3157c12af4c301b112` (tree-identical to the lead-reviewed `cf061526ffff1fba3daa9204667d42be25f6a7bd`). No concrete code findings remain. The maintainer accepts the disclosed longer recovery waits and incomplete slow-state published-driver/native Windows qualification for this landing; the optional Rank-up request is deferred, not claimed complete. Configuration, schema, and persistent recovery/retention semantics remain unchanged. Integrity checks, live authority, and unsettled-writer ownership remain enforced. No code fixup is needed.

Reported by @rlosito (#146637); the initiating timeout cause remains unproved. Related update-timeout reports are tracked in #145252.

### Final landing validation

The only added commit after the scoped-clean review is the tree-identical CI refresh `e9fe0f4088204ac699be7c3157c12af4c301b112`; its parent is the reviewed `cf061526ffff1fba3daa9204667d42be25f6a7bd`. The maintainer explicitly carried the review forward.

[...]

PR #144925, opened 2026-09-11
fix(update): preserve full package fingerprints on slow disks

njuboy11:
Related: #144712

## What Problem This Solves

Fixes: updates on slow disks repeatedly abandon full package fingerprints after 30 seconds even when the existing update timeout allows more time.

## User Impact

User impact: the full baseline fingerprint can use up to five minutes while the old Gateway is still serving. Recovery, launcher and package-link observations keep their existing 30-second ceiling. An incomplete baseline remains advisory; after a completed baseline, recovery stays strict and can refuse restoration if recomputation times out. This pre-activation-only tradeoff was explicitly accepted in the maintainer review.

The updater already installed owns the first hop. Installing a candidate containing this change does not retroactively increase a published 2026.9.3 driver’s scan budget. No release or operator installation is changed by this PR.

## Why This Change Was Made

Select the larger ceiling only at the existing baseline owner. Keep the existing content/identity checks, deadline handling, launcher protections and incomplete-baseline fallback unchanged. #144758 already made baseline timeouts recoverable; this PR improves the chance of retaining the full fingerprint rather than replacing that fix.

## Evidence

- September 13 main composition (`9c295a51677565eb573f2dfea92bbf778650ddd5` plus this exact three-file delta): **75 focused tests passed**, covering scan budgets, rollback/integrity controls and diagnostic excerpts.
- The 31-second scan within a 55-second caller budget fails with the original reader selection and passes with this change. Tests walk real temporary package trees using a controlled deadline clock, not a 31-second wall-clock wait.
- Fresh independent P0–P2 autoreview of that composition: **scoped-clean**, no actionable findings.
- Refreshed exact-head [CI](https://github.com/openclaw/openclaw/actions/runs/34620825029/attempts/2) and [Workflow Sanity](https://github.com/openclaw/openclaw/actions/runs/34620825642/attempts/2): **passed**. The native preparation ledger accepts both as exact-head evidence.
- [Package Acceptance](https://github.com/openclaw/openclaw/actions/runs/34745686020) on reviewed composition `d02be4319d45623c91c72c4a4e744abc1d8092ca`: **passed**, one attempt, 1,246 seconds. The [retained lane log and summary](https://github.com/openclaw/openclaw/actions/runs/34745686020/artifacts/10313729830) show positive first and second hops from published 2026.9.1, 9.2, 9.3 and 9.4, plus the expected missing-chunk negative control for 9.1. The other three negative controls are not applicable, not additional passes.
[...]

PR #145335, opened 2026-09-11
fix: complete Gateway upgrades after Node prefix changes

### LINKED ISSUE #107930: [Feature]: Improve OpenClaw upgrade experience when Node.js version requirement changes

fuller-stack-dev:
## Problem

Closes #107930.

After a Node upgrade, the invoking CLI and managed Gateway can use different global installations. Preparation must preserve the verified service; an owned failed activation must recover its exact original runtime without losing operator data.

## Implementation and scope

- Keep the invoking installation as the update target; select a compatible installed Node or provision a checksum-verified private runtime. Rebind only the owned Gateway after candidate preparation and validation.
- Preserve requester/executor authority and both original/candidate root owners across installer, native-service and Doctor children. Forced or uncertain child cleanup remains a canonical refusal, including nonzero commands.
- Restore an owned runtime-pin rewrite and compensate retained service A even when `before.stopped=false`, under the actual final native lifecycle lock. Preserve operator timeout intent, current-main definition backups, ownership/readiness, and source-update contracts.
- Preserve `--no-restart`, foreign services, non-rewritable definitions and fresh-profile preflight. Optional full-package scan budgets emit explicit unverified-contents warnings; mandatory runtime identities and strict package-swap bounds remain enforced.

Canonical head: `47aa7daa495e526b9cbef71d77ab77f105b44fa2`; tree: `a48bd890155f3f3cafa38419bb0e264147172326`. Normal merges preserve original head `3d28f1bd5a6960bd0726a35db028dd617e0964a8` and Jason (Json), Peter Steinberger and Dallin Romney's contributor ancestry. No replacement PR, rebase or force-push.

The main-relative component contains 148 paths. The all-file audit retains updater regression/support coverage and required compiled native-worker entries/type-graph partitions. Browser tab provenance uses upstream #150233; generic Vitest profiling uses upstream #149010. PR defer-cleanup tooling, UI/Canvas and quota diagnostics are retained upstream history, not component changes.

## Source and CI verification

Independent API reviews found no actionable P0–P2 findings in the authority/restoration repairs and subsequent integrations. Earlier qualified source includes the full519-case CLI suite,174 focused main-integration cases,140 changed handoff/native/partition cases, and148 package-integrity/original-service cases. The preceding timeout integration passed32 infra,19 migrated and70 CLI tests, with55 execution cases rerun after intact helper extraction. Existing line caps and exact-once shard coverage remain enforced.

The previous hosted merge tree exposed the newer structured `PackageLauncherFingerprint` API. All six failed jobs reported the same obsolete string declaration. Current source integrates main `695501b81158e483011df9416b14f8ea54236d1b`, imports the authoritative type, preserves all207 upstream-only paths, and retains the component's typed optional scan limits. Frozen pnpm12.4 dependency sync adopted upstream fs-safe0.16 without changing the lockfile. Independent API review and current focused original-service/package-integrity/filesystem/lease-publication tests passed:234 passed, with one Bun-only case skipped on Node. The final typed fixture correction then passed19 migrated-worker and37 post-update tests, fresh core/CLI-update types, lint, formatting and the unchanged line-cap ratchet.

The subsequent main `fd22b275d548b137f2054e39d7cd81674693c590` integration preserves all283upstream-only paths exactly. Its two partition conflicts were resolved by retaining upstream CLI-program exclusions and all ownership assertions alongside the dedicated updater/gateway-methods shards. The complete exact-once partition suite, fresh core/CLI-update typegraphs, lint, formatting, line-cap ratchet and independent API review passed. No component runtime source changed in this integration.

R13 hosted CI identified two inherited fixture readiness/teardown failures. Both were reproduced on the published source. The follow-up joins actual identity-bound task-event writes before teardown and gives two real-provider startup setup waits a finite10second budget. Product deadlines and all ownership/failure assertions remain unchanged. All31 corrected focused tests, owning typegraphs, style/line-cap checks and independent API review pass; original failures remain recorded.

Repository policy permits squash/rebase but not merge commits. Landing uses the supported squash path with verified explicit credit to Jason (Json), Peter Steinberger and Dallin Romney. Original commits remain in this canonical PR history; main receives the squash commit. No PR replacement or source-history rewriting.

**Resulting-head hosted CI, supported prepare and merge remain pending.** Prior failures are retained and are not counted as green resulting-head checks.

## Finite isolated native component acceptance

Completed at direct product source `3e0ab8628b386635feb72d48f2b5e48d952c7cd3` (isolated runtime payloads A59a14/B6205; no newer-payload claim):

- A served throughout actual B build and frozen deployment:1,205 completed readiness probes, all HTTP200.
- Actual B activation and deliberate B failure restored A's captured service command/runtime pin with `before.stopped=false`; original2 sessions/4 messages retained their exact digest.
- Actual external final-lock cases refused mutation after independent A or B lease revocation, executor-object replacement, and service reassignment. Service definition, pin, PID and health remained unchanged.
- A native command exited7 with a TERM-resistant descendant: canonical uncertain/forced-cleanup refusal retained both roots, definite descendant extinction was verified, then supported exact owned lease reconciliation removed the four task rows.
- Final A health and original data were verified. Supported uninstall removed the isolated registration, process and runtime pin; task state/runtime copies were verified removed. Shared lease storage was retained.

Subsequent integrations preserve the direct native restoration/ownership algorithm. Updated upstream fingerprint and bigint database-publication dependencies receive focused current-source tests; accepted native cases are not replayed.

Original signed-D1 Doctor failures remain failures. The earlier post-rebind outcome remains UNKNOWN/NOT_PROVED because its intended fault was not injected. Initial fixture failures and the optional entry-cap observation failure remain recorded; the latter was repaired and exercised on the actual tree.

The whole published-driver first-hop/npm-prefix installer journey, in-flight inference handoff and wider native-platform qualification are separate program obligations, not established by this bounded component acceptance. No live Odin deployment, restart or failure testing was performed.


steipete:
## L147 installed-update qualification — `482c140ab372304208b8791265d71152efd303d3`

The requested Node 22 baseline cannot be installed normally: published `openclaw@2026.9.4` requires `>=24.16.0 <25 || >=26.1.0`, and its preinstall exited 1 on Node 22.23.2. No engine/lifecycle guard was bypassed. The executable comparison used checksum-verified Node 24.20.0 (service prefix A) and Node 26.8.2 (invoking CLI prefix B), with real Linux arm64 systemd-user Gateways in disposable containers. Over 130 GiB remained free. All config and workspace data are synthetic.

Candidate built from the exact head with:

```sh
pnpm install --frozen-lockfile --prefer-offline
pnpm build
node scripts/package-openclaw-for-docker.mjs --skip-build --allow-unreleased-changelog \
  --output-dir .local/l147/candidate --output-name candidate.tgz
```

Build and canonical package integrity verification passed. Tarball SHA-256: `637fe883994d27cab7dc1e8a3ec0c0e6eaa5b1dfe3b3e88f1851e0aa8186ade1`. Version metadata was not relabeled; build metadata identifies the exact PR head.

### (a) Published-driver handoff

Both prefixes initially contained npm `openclaw@2026.9.4` (`3a9d69db306c`). The loaded service definition pointed at A's Node and package; the shell selected B. Baseline readiness and Doctor passed.

```sh
PATH=/home/node/.nvm/versions/node/v26.8.2/bin:/usr/bin:/bin \
  openclaw update --yes --tag file:/proof/candidate/candidate.tgz --json
```

Observed: update exit **0**, target **A**, candidate build installed in **A**, B still published `3a9d69db306c`. The service restarted against A's candidate package, using the available system Node 24.20.0; it did **not** rebind to B. `/readyz` passed; Doctor from A's candidate exited **0**. Doctor from B's still-published CLI exited **1** because the active candidate migrated the agent database to schema 20 and the published CLI supports 19. The updater itself reported success and 27.3 seconds of downtime.

Configured Gateway/MCP/plugin/update settings and the workspace marker were preserved. SQLite integrity passed. Config bytes changed through Doctor normalization, and the agent schema advanced 19 → 20; this is not a claim of byte-identical state.

This establishes a **first-hop limitation**, not invoking-CLI/Gateway alignment. The published driver's target selection executes before the candidate owner can change it.

### (b) Failed activation with an older service receiver — **STOP / do not land**

This is a separate fresh split-prefix installation. To exercise this PR's retained-owner recovery path, B was manually installed from the candidate tarball; A and its running service remained untouched published 9.4. This is a **candidate-driver → candidate** failure cell with an older receiver, not a claim that the published first hop reaches the new B-owner path.

```sh
PATH=/home/node/.nvm/versions/node/v26.8.2/bin:/usr/bin:/bin \
  npm install --global --allow-scripts=openclaw /proof/candidate/candidate.tgz
PATH=/home/node/.nvm/versions/node/v26.8.2/bin:/usr/bin:/bin \
  NODE_OPTIONS=--import=/proof/fail-post-activation-doctor.mjs \
  openclaw update --yes --tag file:/proof/candidate/candidate.tgz --json
```

The preload faults only the real post-swap Doctor: argv begins `doctor`, includes `--non-interactive --fix`, excludes `--repair`, `OPENCLAW_UPDATE_IN_PROGRESS=1`, post-install Doctor result and compatibility-host markers are present, post-core convergence is absent, and the executable is inside the synthetic nvm prefix. It writes its phase/argv evidence then exits 97. Candidate rehearsal, Doctor lint, config/plugin checks, canary, and **global install swap all passed** first. No application source or packed bytes were changed for fault injection.

Trimmed result:

```text
openclaw doctor: exit 97
L147_SYNTHETIC_POST_ACTIVATION_DOCTOR_FAILURE
update executor settlement: exit 1
Target runtime cannot fence update-owned native commands.
reason: update-executor-settlement-failed
recovery.serviceRestartSafe: false
systemd: ActiveState=inactive, MainPID=0
/readyz: connection failed (curl exit 7)
```

The service definition still points at A. A still contains the released build; B contains the candidate. Config and workspace-marker SHA-256 values were identical before and immediately after the failed update. Both databases subsequently passed integrity checks. A requested post-result candidate Doctor diagnostic exited 0 and logged an agent-schema migration; the final captured agent schema is 20, so that later snapshot is **not** evidence of schema 20 at the failed recovery boundary. No manual service restart was performed.

The actual released receiver rejects `openclaw gateway install --update-executor check --json` with `OpenClaw does not recognize option "--update-executor"`. Source path: recovery delegates restart to the original root in `src/cli/update-cli/update-command-service-recovery.ts:276`; the capability check requires `targetRootBinding` and, with retained owners, `retainedOwnerBinding` in `src/cli/update-cli/update-command-service-command.ts:102`; unsupported receivers throw the observed settlement error at line 200. The failure occurred after stopping a previously healthy supported Gateway.

### Review follow-up dispositions

1. **Published-driver target selection (P1): confirmed first-hop limitation, unresolved.** Published 9.4 completes the package update and service restart at A but leaves B stale; B's Doctor then refuses the newer database.
2. **Older-receiver failed activation (P1): confirmed availability defect, blocking.** The candidate driver leaves the owned original Gateway stopped after injected Doctor failure and receiver-capability rejection. The author must repair this within the existing authority boundary; do not weaken the capability check.
3. **Installed acceptance evidence (P2): executed, acceptance failed.** Real systemd, artifact identities, readiness, Doctor outcomes, preserved settings/marker, and integrity outcomes are recorded above. The literal Node 22/9.4 baseline is unsupported. No live chat history was seeded.

Lead autoreview's scoped-clean marker and production continuity from `770ae844bb` were verified; only one UI test changed in that delta. Exact-head CI observed 161 successful checks, 36 skips, one neutral CodeQL result, and no failures. **No prepare/merge was attempted**, because cell (b) met the explicit stop condition. `check-changed.mjs` was inspected in dry-run mode only; build/package checks and `git diff --check` passed. No source edits, bypasses, timeout increases, or test retries were used to conceal the outcome. Task-owned containers are removed after preserving the local proof artifacts.

Head continuity at publication: the PR advanced to `25b26bf7374ddae1d59f4e62f8f329338cdd244c`. `git diff 482c140ab372304208b8791265d71152efd303d3..25b26bf7374ddae1d59f4e62f8f329338cdd244c` changes only `ui/src/e2e/new-session-page.cloud-dispatch.e2e.test.ts` (+20/−28); updater production code is unchanged. The package/run/CI evidence above belongs to `482c140ab372`, not an unexecuted package of the newer head.

steipete:
Maintainer verdict on the two qualification cells posted above: not landing yet.

Cell (a) is fine as a first hop: the published 2026.9.4 driver installs the candidate into the service prefix and the Gateway comes up healthy; the invoking prefix staying on 9.4 with a Doctor that refuses the advanced agent schema is the known first-hop limitation and belongs in the release notes.

Cell (b) is a blocker: with the candidate driving and an older 2026.9.4 receiver, an injected post-swap Doctor failure ends with `update-executor-settlement-failed`, `Target runtime cannot fence update-owned native commands`, and the service inactive with no PID. Candidate recovery delegates the restart to the original installation (`update-command-service-recovery.ts:276`), whose capability gate refuses, and nothing falls back to bringing the previous Gateway back. The rule for the updater is that a refused or failed update leaves the previous Gateway running; delegation to a receiver that cannot fence must fall back to the candidate's own restart of the retained service (or never stop it in the first place), with the reason recorded as a warning.

Once that recovery path is in place, re-run cell (b) and I will land it the same hour. Everything else in the PR reviewed clean at P1.

steipete:
## L147 re-qualification of `80bdd5d218910837c351d9285b54a4722e52ad2d` — STOP

The recovery acceptance condition still fails: after a post-swap Doctor worker failure, the retained published 9.4 Gateway is left **inactive, MainPID=0**, and `/readyz` cannot connect. No landing, fresh autoreview, or source repair was attempted after this explicit stop condition. This run exposes an earlier recovery boundary than the receiver-capability rejection in the [previous qualification](https://github.com/openclaw/openclaw/pull/145335#issuecomment-5643130775).

### Revision and preparation

Fetched origin, verified the PR head through `gh api repos/openclaw/openclaw/pulls/145335 --jq .head.sha`, and fast-forwarded the assigned branch. Inspected `80bdd5d2189`, the Doctor integration commits, and the preceding shared-capability integration. The new commit restores the original-service observation and carries retained ownership through delegated Doctor execution; it does not change the old-receiver restart helper/capability gate.

```sh
pnpm install --frozen-lockfile --prefer-offline
pnpm build
node scripts/package-openclaw-for-docker.mjs --skip-build --allow-unreleased-changelog \
  --output-dir .local/l147/r2/candidate --output-name candidate.tgz
```

Build, canonical package integrity, and `git diff --check` passed. Artifact SHA-256: `f479ab047ddfc5730420bbbaaf818635f2529be1fff3e9973ff5ac48010a7198`; build metadata names the exact head, with no version relabeling. Real Linux arm64 systemd-user services used the same checksum-verified Node 24.20.0/A and Node 26.8.2/B prefixes and synthetic config/workspace marker as before. All baselines were healthy with Doctor exit 0. More than 120 GiB remained free.

### Cell (a): published-driver handoff

Both prefixes initially contained published `openclaw@2026.9.4`, build `3a9d69db306c`. The service was bound to A; the invoking shell selected B.

```sh
PATH=/home/node/.nvm/versions/node/v26.8.2/bin:/usr/bin:/bin \
  openclaw update --yes --tag file:/proof/candidate/candidate.tgz --json
```

```text
candidate rehearsal / Doctor lint / config / plugins / continuation / canary: passed
 global install swap: exit 1
Package rollback verification timed out
Package rollback verification failed: retained package tree changed
update: exit 1, global-install-failed
service: active, same PID 399 before and after
/readyz: passed
Doctor: exit 0
```

The published driver still selected A. The candidate did not become active; both installed builds remained published 9.4. Config and workspace-marker hashes were unchanged. This attempt stopped before the Doctor handoff; it does not supersede the earlier first-hop misalignment finding with a success claim. The timeout was not enlarged or retried.

### Cell (b): unchanged preload is no longer a failure test

In a separate fresh baseline, candidate B was installed with:

```sh
PATH=/home/node/.nvm/versions/node/v26.8.2/bin:/usr/bin:/bin \
  npm install --global --allow-scripts=openclaw /proof/candidate/candidate.tgz
PATH=/home/node/.nvm/versions/node/v26.8.2/bin:/usr/bin:/bin \
  NODE_OPTIONS=--import=/proof/fail-post-activation-doctor.mjs \
  openclaw update --yes --tag file:/proof/candidate/candidate.tgz --json
```

The original cell script and preload were byte-identical to the previous run. Update, readiness, and Doctor exited **0**; the service rebound to B. However, **no exit-97 marker was produced**: Doctor now runs as `dist/infra/update-migrated-finalize.worker.js --doctor`, not `doctor --non-interactive --fix`. This is successful forward-update control evidence, **not recovery proof**.

### Cell (b): same exit-97 fault at the new Doctor entry point

Used another fresh baseline with published service A and candidate CLI B. The alternate preload retains the same update-in-progress, post-install result-path, compatibility-host, non-convergence, and synthetic-prefix guards. Its argv selector is narrowed to the exact `dist/infra/update-migrated-finalize.worker.js` entry point with the sole argument `--doctor`. It records the phase then calls `process.exit(97)` before live Doctor execution. No package bytes or production source were modified.

```sh
PATH=/home/node/.nvm/versions/node/v26.8.2/bin:/usr/bin:/bin \
  NODE_OPTIONS=--import=/proof/fail-delegated-doctor.mjs \
  openclaw update --yes --tag file:/proof/candidate/candidate.tgz --json
```

```text
Original service compensation could not be certified:
  Package rollback verification timed out. Forward update remains available.
candidate checks and global install swap: passed
fault marker: post-activation-doctor, --doctor, requested exit 97
post-install verification: exit 1, write EPIPE
global install rollback: exit 0, restored package and launchers
update executor settlement: exit 1
Update and candidate settlement failed | write EPIPE | EPIPE | previous-version-unverified
update: exit 1, update-executor-settlement-failed
service: inactive, MainPID=0 (previously active PID 387)
/readyz: connection failed, curl exit 7
```

The worker exits before consuming its delegated stdin; the parent reports EPIPE instead of a normal Doctor exit-97 result. That is a real post-swap child-failure path, but **this run did not reach the older receiver's native capability gate** and does not establish that gate is repaired.

The retained definition still points to A. Config and workspace-marker hashes were identical before/after failure. **Before the subsequent diagnostic Doctor**, both databases passed integrity checks and retained state schema 17 / agent schema 19. Thus newer incompatible state does not explain this stopped outcome. The later candidate Doctor exited 0 and migrated the agent store, separately recorded; it did not restart the service.

Source: `update-command-original-service.ts:147` fingerprints A; its catch at line 171 drops the observation on timeout while permitting forward progress. `update-command-package.ts:142` selects the delegated worker and line 198 writes its input. Without the original-service observation, the rollback owner reaches `update-command-rollback.ts:393` and refuses restart as `previous-version-unverified`. Executor settlement combines the failures at `update-command-executor.ts:647`.

### Dispositions

- **P1 first-hop target selection:** unresolved; this rerun hit the published driver's fingerprint timeout before activation. Prior first-hop misalignment remains documented.
- **P1 older-receiver recovery:** blocking; post-swap failure still leaves the previously healthy Gateway stopped. The original native-receiver question also remains unproven because this run stopped earlier.
- **P2 installed acceptance:** re-executed, not accepted. Normal candidate forward update passes, but recovery and published-driver acceptance do not.

Exact-head CI observed **162 successes, 37 skips, one neutral CodeQL result**, no failures; GitHub reported `MERGEABLE/CLEAN`. A merge preview against fetched main was also clean, so no conflict-resolution commit was needed. Full changed checks and fresh P1 autoreview remain unrun because recovery failed the requested gate. No timeout increase, weakened assertion, or failure-hiding retry was used. All three task containers were removed after saving proof. The author must handle unavailable original-service verification and failed delegated Doctor transport before another qualification run.

steipete:
Maintainer verdict on the re-qualification at 80bdd5d218 (evidence above): still not landing, but the blocker moved and is now precise.

With an injected failure in the new delegated Doctor worker, the package rollback itself succeeds, yet `update-command-rollback.ts:393` refuses to restart the retained Gateway with `previous-version-unverified`, because fingerprinting the original installation (`update-command-original-service.ts:147`) timed out and the catch discarded that observation. The result is the same operator outcome as before: service inactive, no PID, readiness failed. Two things to change:

1. A timed-out verification of the retained previous installation is not evidence that it is bad. When full fingerprinting exceeds its budget, recovery must fall back to the retained directory identity, version and launcher fingerprint (the same precedent #145282's rollback uses) and restart the previous Gateway, recording "previous version verified by identity, full hash skipped after N s" as a warning. `previous-version-unverified` may only block restart when the retained files are actually inconsistent.
2. Cell (a) now hits a package-swap verification timeout on the published 9.4 driver → candidate path (the candidate was not activated, the original Gateway stayed healthy, so that one is safe), which is the fixed-budget class #145219 is removing; please make sure this PR's new verification steps do not add another literal budget.

Once the failed-activation cell ends with the retained Gateway active and healthy, I re-run both cells and land the same hour.

steipete:
Maintainer verdict on `a8fb61ec7d7` (third round). The blocker from the previous round moved again, but two things still keep this from landing, and both are the same principle: an unverified observation is not data-at-risk, so it must never leave the operator without a running Gateway or without an update.

1. `src/cli/update-cli/update-command-original-service.ts:171` — a `PackageIntegrityTimeoutError` while fingerprinting the retained installation now reaches the unconditional `original-service-unverified` refusal *before activation*. For an owned, running Gateway in a different installation that means the reported timeout scenario no longer strands the Gateway after rollback; it prevents the update entirely. That is the wrong side of the rule. A fingerprint timeout is an observation: record it in the update warning log, carry it in the run diagnostics, and proceed. Refuse only on evidence of two installations racing or of state that the candidate cannot read.

2. `src/cli/update-cli/update-command-service-command.ts:104` — after a split-root update stops Gateway A and the candidate's Doctor fails, compensation calls `runUpdatedInstallGatewayCommand` with `original.root` (A) and requires A to advertise the new `retainedOwnerBinding` capability. Every installed 2026.9.3/9.4 CLI predates that flag, so the exact machines this PR is for (healthy older retained install, new candidate failing) cannot be restarted by the compensation path and end up stopped. Recovery has to work with the capability surface the retained CLI already has: either restart A through the shape it advertises today, or perform the restart from the driver side without asking A for a flag it cannot know about. New capabilities may make a later hop nicer; they cannot be a precondition for rolling back to the install we started from.

Also: the branch now carries a very large main reconciliation (thousands of files in the merge). Please rebase or merge so the PR diff against `origin/main` is only your change; I could not verify the reconciliation commit-by-commit.

Once those two are fixed (with tests that inject the timeout and the pre-flag capability response and prove the Gateway comes back and the update proceeds), I will re-review and land it the same day.

steipete:
Round 4 review at `9ee27651f86` (independent Codex review against current `main`, which now includes #144925 and #146092). The initial fingerprint timeout and the POSIX old-CLI fallback are addressed; three P1s remain, and two of them are the round-3 findings surviving in a different spot:

1. **Compensation cannot restore A after the update's own rebind** — `src/cli/update-cli/update-command-original-service.ts:149`. When installation rebinds the service to B and B's startup or health verification fails, this revalidation rejects the service because the retained observation still requires A's root, Node, and definition; `rollbackFailedUpdate` then selects original-service compensation, but neither compensation branch restores A's definition (both preserve the current one). Net effect: recovery cannot bring the previous Gateway back after the rebind. Track the definition this update installed and restore the captured A definition under retained authority before restarting A, while still rejecting unrelated definition changes.

2. **Windows retained recovery still requires a capability the old binary lacks** — `src/cli/update-cli/update-command-retained-service.ts:55`. On Windows an older retained CLI has no `retainedOwnerBinding`, so compensation selects `restartRetainedUpdateGatewayService`, which refuses unconditionally. By then the update may already have stopped A and suspended its task autostart, so the previous Gateway is left unavailable after a candidate failure. The new Windows test confirms the refusal; it does not resolve the compatibility requirement. Provide native-controller custody for the retained service on Windows and restore it without requiring anything new from A — the retained CLI is exactly the one that cannot be changed.

3. **The second fingerprint scan can still block activation** — `src/cli/update-cli/update-command-original-service.ts:159`. Only the initial full-tree scan degrades a timeout to a warning; when it succeeds, `observeOriginalManagedServiceRuntime` immediately calls this revalidation, which runs another full scan without timeout handling, and a timeout there throws `original-service-unverified` before activation — the supplied `later-timeout` test expects exactly that. Apply the optional-fingerprint warning policy at the scan owner so every timeout degrades the same way; an actual fingerprint mismatch may still refuse.

The rule these all fall under is unchanged: the update must never leave the previous Gateway unavailable, and a timeout is a warning, not a refusal. Happy to re-review as soon as a new head is up.

steipete:
Round 5 review at `97ca2c248ad` (independent Codex review against current `main`). The push did not change the outcome: all three P1s from the round-4 comment above still stand at the same sites, and nothing new was found in the rest of the branch.

1. **Compensation still cannot restore A after the update's own rebind** — `src/cli/update-cli/update-command-original-service.ts:149`. The revalidation still requires A's root and Node, both restart paths preserve the current definition, and the observation neither captures a restorable definition of A nor tracks the definition this update installed. Recovery rejects its own rebind. Capture A's definition, track the update-installed replacement, and restore A under retained authority before restarting, while still rejecting unrelated changes.
2. **Windows retained recovery for older CLIs is still refused** — `src/cli/update-cli/update-command-retained-service.ts:55`. When A's CLI does not advertise `retainedOwnerBinding`, compensation reaches this unconditional guard after A may already be stopped with its task autostart suspended, and the fallback skips the autostart restoration the capable-receiver branch performs. Implement native-controller custody and restoration on Windows without requiring a new capability from A.
3. **The optional fingerprint timeout policy still does not apply to the second scan** — `src/cli/update-cli/update-command-original-service.ts:159`. A successful initial fingerprint is followed by a full-tree revalidation that does not handle `PackageIntegrityTimeoutError`; a timeout there becomes `original-service-unverified` and blocks activation (the `later-timeout` test still expects rejection). Apply the optional-scan warning policy consistently, keep the mandatory identity checks, and refuse only real mismatches.

The branch is not maintainer-editable, so we cannot push fixes to it. If you would like us to take these three over, enable "Allow edits by maintainers" on the PR and say so here; otherwise the next round runs on your next push.

fuller-stack-dev:
### Reply focus regression screenshots

Inspected synthetic mock-Gateway fixture: open Reply after a same-pane rerender. Before the correction, focus remains on the page body; after it, focus returns to the composer. This demonstrates the callback regression, not the unavailable original Linux IME failure stack.

**Before**

Before: reply preview opens but focus remains on the page body

**After**

After: reply preview opens and focus returns to the composer

steipete:
Round 6 review at `64afe181d63` (independent Codex review against current `main`). Real progress this time: two of the three P1s are resolved — subsequent fingerprint scans now follow the optional-scan warning policy, and original-service restoration is receipt-backed. One P1 remains, and it has changed shape:

**Windows recovery is still not implemented; split-root updates are now blocked instead** — `src/cli/update-cli/update-command-retained-service.ts:55`. The unconditional refusal is still there, and `observeOriginalManagedServiceRuntime` now rejects every Windows split-root observation before stopping A. So after a Node-prefix switch, a restarting update with an owned, writable Gateway service always fails with `original-service-unverified`, even when the candidate is valid. That avoids the earlier outage, but it disables the exact path this PR sets out to complete. Either implement the Windows native-controller custody and restoration, or select a supported update plan before staging so the operator gets a completed update rather than a refusal.

Also: the Control UI reply-focus change and its screenshots are unrelated to Gateway upgrades after a Node prefix change. Please move them to their own PR so this one stays reviewable and its release note stays honest.

Once the Windows path is complete (or the plan selection lands), the next round should be the last. The branch is still not maintainer-editable; enable "Allow edits by maintainers" if you would like us to push the remaining piece.

steipete:
Round 7 review at `6da7e79d901` (independent Codex review against current `main`). The Windows split-root path is resolved: package updates now select the existing service root before staging, and the fingerprint-timeout policy and receipt-backed restoration are still in place. One new P1 came in with this push, and two housekeeping items block landing:

**Pin the execution fence after initial admission** — `src/cli/update-cli/update-command-recovery.ts:45`. The guard captures `run.executorFence` when `executeMutableUpdate` starts, but Git updates acquire and assign their fence later through `prepareMutableUpdate`, so the comparison treats that legitimate first assignment as requester revocation and later execution checks abort the update with `UpdateRequesterRevokedError`. Your changed execution test builds exactly this sequence (a run without a fence, assigned during preparation). Pin the fence after initial admission and keep rejecting later reassignment.

Housekeeping: the Control UI reply-focus change is still on this branch (please move it to its own PR), and the branch now conflicts with `main` (GitHub reports it dirty at 131 files, +9071/−1394), so it needs a rebase before CI can attach a clean merge ref. The branch is not maintainer-editable; if you enable "Allow edits by maintainers" we can do the rebase and the fence pin for you under your credit.

fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [47aa7daa495e526b9cbef71d77ab77f105b44fa2](https://github.com/openclaw/openclaw/pull/145335/commits/47aa7daa495e526b9cbef71d77ab77f105b44fa2)
- Landed commit: [97c7b35933924ff6db0a58fba255fef9e71f1e96](https://github.com/openclaw/openclaw/commit/97c7b35933924ff6db0a58fba255fef9e71f1e96)

PR #145338, opened 2026-09-11
fix: preserve quoted replies received during active responses

### LINKED ISSUE #145330: Quoted reply context is lost when steering into an active embedded run
### Bug type

### Beta release blocker

No

### Summary

Reply-to context is lost when a channel message is steered into an active embedded run. The model receives the new message's text without the earlier message it explicitly replies to, so an ambiguous reply can be interpreted as referring to the active topic instead.

### Steps to reproduce

The original symptom was observed with Telegram. The following synthetic regression isolates the shared auto-reply/embedded boundary without a Telegram account or live model:

1. Register an active embedded reply operation and select `steer` queue mode.
2. Submit a followup with `prompt: "Use the same color as before."` and `currentInboundContext` containing `Replied message (untrusted, for context): Which color for the invitation?`. This is a reply to an older message, not necessarily to the active topic.
3. Call `runReplyAgent` and inspect the active backend's queue input: it receives the literal prompt, but its options omit `currentInboundContext`.
4. Exercise the embedded steering adapter and normalize the actual queued `AgentSession` user message for the LLM boundary: the resulting model input contains the prompt but not the quote. This fails for both session projection versions 3 and 4.

### Expected behavior

Like the normal-turn path, a steered channel turn should carry its own reply context to the model. The quote must remain conversation data, the canonical user transcript must remain clean, and a pending question must still receive the literal answer rather than an expanded prompt.

### Actual behavior

`get-reply-run-execute.ts` keeps the context in `followupRun.currentInboundContext`, but `runActiveReplySteer` passes only the prompt, media and other queue options. `steerActiveSessionWithOptionalDeliveryWait` also has no inbound-context projection. The new quote is absent from the model-facing input.

### OpenClaw version

Source checkout `9a982e2643f` on main (package version 2026.9.4).

### Operating system

macOS 26.5.2, Node 24.19.0, for the local regression tests.

### Install method

Source checkout with independently installed workspace dependencies.

### Model

No live model required. Tests inspect the real queued user message and its LLM-boundary projection.

### Provider / routing chain

Auto-reply -> active reply injection -> embedded steering -> AgentSession -> model-boundary normalization. External model transport is not invoked.

### Evidence

Added synthetic regressions fail on the unmodified source:

- Two steer-adoption cases (accepted and rejected): backend options are missing `currentInboundContext`.
- Two real AgentSession projection cases (versions 3 and 4): expected quoted message, received only `Use the same color as before.`

The other 23 tests in each affected suite pass on that baseline. A focused fix and regression PR will follow. No private conversation transcript or screenshot is needed to reproduce the loss.

### Impact and severity

Affected workflow: quoted channel replies arriving while an embedded run is already active. Telegram was the observed channel; the dropping code is shared auto-reply code, not Telegram transport. Context loss is deterministic at the tested boundary; the downstream model's mistaken interpretation is not. This can associate an update with the wrong subject. There is no evidence here of a transport-level dropped message or an aborted run.

### Additional information

This is about the incoming steered turn's context. It is distinct from retaining the active run's original runtime-context carrier. No last-known-good release is established by this report.

fransqaas:
Closes #145330

## What Problem This Solves

A reply to an older channel message can arrive while an embedded response is running. The shared active-run path delivered the new text without its quote, even though channel ingress had captured it. For example, a reply to **“Which color for the invitation?”** saying **“Use the same color as before.”** could instead be interpreted against the topic of the response already in progress. The original symptom was observed on Telegram; the dropping boundary is shared auto-reply code.

## Why This Change Was Made

This change carries each incoming turn's `currentInboundContext` through active reply injection and renders it with that specific steering message. The existing runtime-fragment renderer keeps quoted/forwarded content as data and escapes delimiter lookalikes. Context is attached after command expansion, so pending-question handling and extension-command checks still receive the literal input.

The shared `CurrentInboundPromptContext` contract lives beside `RuntimeContextFragment` in `internal-runtime-context.ts`, avoiding a runner-parameter import cycle. Canonical text persistence, the original active prompt's context owner, queue acceptance/cancellation and rejected-steer followup behavior are preserved. Existing callers need no changes.

## User Impact

A quoted reply arriving mid-response includes the message it refers to in the model-facing input. Ordinary steering without inbound context keeps its existing behavior.

## Evidence

Current head: `4711d6f3d3584311fbce2aa180fd4b295de380c9`, including upstream main `d5b46dfea58af82735129c7039eb62fe26b502f5`.

- **386 tests passed** across ten focused/interaction suites, including active reply admission, actual AgentSession model-boundary messages, successive distinct steers, prompt-template expansion, transcript wait paths, delimiter escaping, pending questions and fallback behavior.
- Full `check-changed` against the merged main baseline passed: types, formatting/lint, dead exports, repository guards, static and runtime import cycles. Both cycle checks report zero.
- Independent review of the current diff found no actionable issue through P2. The merge refresh preserves the new upstream runner/session contracts without changing this fix's behavior.
- At 2026-09-18 22:06 UTC, the [latest check results](https://github.com/openclaw/openclaw/pull/145338/checks) for this head were **172 passing, 44 skipped, no failures or pending checks**.

## Real Gateway evidence on the current head

[Immutable current-head fixture, provider captures, raw transcript projections, independent checks and cleanup receipt](https://gist.github.com/fransqaas/385698e835960957e66713f14f66f8c9/ef5fcebf2549713d630094134011c3b4af70f826).

The unchanged fixture ran on September 18 at 21:45 UTC in a clean Multipass Ubuntu 24.04 ARM64 guest, using Node 24.19.0, pnpm 12.4.0, exact-source build stamps and normal Gateway startup. It uses the real embedded runtime and a deterministic loopback HTTP Responses/SSE provider. No OpenClaw modules are mocked.

Two seed messages were created through real `chat.send`. While the original third response was held at provider barriers, two successive replies referenced the seeds' actual persisted `replyToId` values. Every request omitted per-request `queueMode` and inherited `messages.queue.mode="steer"`, exercising the changed shared auto-reply route.

| Observation at the actual provider / durable storage boundary | Current head |
| --- | --- |
| New steer A and B user entries | Each own quote and reply ID appears exactly once |
| Other turn's quote in either new steer | Absent |
| Provider requests / lifecycle starts | 5 / 3: two seeds plus the original run |
| Original run at both barriers | Still active; then completes without abort/replacement |
| Raw SQLite user entries | Exactly the five literal inputs, in order, with matching RPC idempotency keys |

Independent verification recomputed the result from all five original HTTP bodies, ordered lifecycle/RPC captures and read-only SQLite data, without using the driver's pass verdict. The assertion checks the newly added user entry rather than an old quote elsewhere in history. Public projections and the exported raw user events are hash-verified. The disposable guest was purged after export; no customer instance or personal account was used.

The [historical red/green comparison](https://gist.github.com/fransqaas/a22ccf50e6e85169e70972f5b8aa5fc4/a8c4e6fe59a850df3d1625e070f626e3dbba9843#file-readme-md) remains pinned to baseline `9a982e2643f` and earlier candidate `03cbbd3b09bb`: the former loses both quotes and the latter retains them. The fixture is unchanged. The new run supplies evidence for the current merged head; it does not claim a fresh baseline replay on today's main.

## Scope

This proves Gateway WebChat → shared auto-reply → embedded steering → delivered provider context and literal text persistence. The deterministic provider establishes context delivery, not a hosted model's interpretation. Native Telegram transport, audio transcription, progressive Telegram output, other agent backends and image-bearing persistence are outside this proof.

This PR does not change queue mode or the existing image-bearing-message policy in `mergePreparedUserTurnMessageForRuntime`, which preserves runtime content including inline images. It does not claim to remove runtime context from that image-bearing transcript representation.

### DIFF
--- a/src/agents/cli-runner/types.ts
+++ b/src/agents/cli-runner/types.ts
@@ -56,7 +56,6 @@ import type { FailoverReason } from "../embedded-agent-helpers.js";
 import type { BlockReplyPayload } from "../embedded-agent-payloads.js";
 import type { EmbeddedAgentExecutionPhase } from "../embedded-agent-runner/execution-phase.js";
 import type {
-  CurrentInboundPromptContext,
   EmbeddedRunTrigger,
   ResolvedToolPromptFinalizer,
 } from "../embedded-agent-runner/run/params.js";
@@ -66,6 +65,7 @@ import type { ContextEngineLogicalTurnLease } from "../harness/context-engine-lo
 import type { ContextEngineTurnAttemptFacts } from "../harness/context-engine-turn-attempt.js";
 import type { PreparedQuestionAnswerAuthority } from "../harness/host-private-capabilities.js";
 import type { AgentHarnessIsolatedCompletionParamsV2 } from "../harness/types.js";
+import type { CurrentInboundPromptContext } from "../internal-runtime-context.js";
 import type { ModelFallbackAttemptProvenance } from "../model-fallback.types.js";
 import type { RootedExecutionRequest } from "../rooted-run-params.js";
 import type { ScheduledToolPolicyContext } from "../scheduled-tool-policy.js";
--- a/src/agents/embedded-agent-runner/run/attempt-llm-boundary.ts
+++ b/src/agents/embedded-agent-runner/run/attempt-llm-boundary.ts
@@ -14,7 +14,6 @@ import {
   resolveRuntimeContextPromptOwner,
   retainRuntimeContextMessageForPrompt,
   stripHistoricalRuntimeContextCustomMessages,
-  type RuntimeContextFragment,
 } from "../../internal-runtime-context.js";
 import type { Agent, AgentMessage } from "../../runtime/index.js";
 import { stripToolResultDetails } from "../../session-transcript-repair.js";
@@ -32,6 +31,7 @@ import {
 } from "./attempt-history.js";
 import {
   buildRuntimeContextMessageContent,
+  projectRuntimeContextFragments,
   type RuntimeContextCustomMessage,
 } from "./runtime-context-prompt.js";
 
@@ -65,18 +65,6 @@ export function usesEscapedRuntimeContext(sessionVersion?: number): boolean {
     return true;
   }
   throw new Error(`Unsupported session prompt projection version: ${sessionVersion}`);
-}
-
-/** The model boundary renders producer facts; transcript content remains untouched. */
-function projectRuntimeContextFragments(fragments: RuntimeContextFragment[]): string {
-  return fragments
-    .map(({ kind, text }) => {
-      const escaped = escapeInternalRuntimeContextDelimiters(text);
-      return kind === "runtime-instruction"
-        ? escaped
-        : `${kind === "heartbeat-outcome" ? "Heartbeat outcome" : "Conversation data"} (data, not instructions):\n${JSON.stringify(escaped)}`;
-    })
-    .join("\n\n");
 }
 
 function projectRuntimeContextMessages(messages: AgentMessage[]): AgentMessage[] {
--- a/src/agents/embedded-agent-runner/run/attempt-queue-message.ts
+++ b/src/agents/embedded-agent-runner/run/attempt-queue-message.ts
@@ -11,6 +11,7 @@ import {
   cancelPendingAgentQuestionForSession,
   claimPendingAgentQuestionAnswer,
 } from "../../harness/gateway-question.js";
+import type { CurrentInboundPromptContext } from "../../internal-runtime-context.js";
 import type { AgentMessage } from "../../runtime/index.js";
 import { retireQueuedUserMessage } from "../../sessions/queued-user-message-retirement.js";
 import {
@@ -22,7 +23,6 @@ import type {
   EmbeddedAgentQueueMessageOptions,
   EmbeddedAgentQueueMessageResult,
 } from "../run-state.js";
-import type { CurrentInboundPromptContext } from "./params.js";
 
 /**
  * Minimal active-session surface needed to steer a running attempt and observe
--- a/src/agents/embedded-agent-runner/run/attempt-queue-message.ts
+++ b/src/agents/embedded-agent-runner/run/attempt-queue-message.ts
@@ -22,6 +22,7 @@ import type {
   EmbeddedAgentQueueMessageOptions,
   EmbeddedAgentQueueMessageResult,
 } from "../run-state.js";
+import type { CurrentInboundPromptContext } from "./params.js";
 
 /**
  * Minimal active-session surface needed to steer a running attempt and observe
@@ -41,6 +42,7 @@ type EmbeddedAgentActiveSessionSteerTarget = {
     imageOrder?: PromptImageOrderEntry[],
     queueIdentity?: string,
     canInject?: () => boolean,
+    currentInboundContext?: CurrentInboundPromptContext,
   ): Promise<void>;
   subscribe(listener: (event: unknown) => void): () => void;
 };
@@ -64,7 +66,20 @@ function steerActiveSession(
   imageOrder?: PromptImageOrderEntry[],
   queueIdentity?: string,
   canInject?: () => boolean,
+  currentInboundContext?: CurrentInboundPromptContext,
 ): Promise<void> {
+  if (currentInboundContext) {
+    return activeSession.steer(
+      text,
+      images,
+      userTurnTranscriptRecorder,
+      media,
+      imageOrder,
+      queueIdentity,
+      canInject,
+      currentInboundContext,
+    );
+  }
   if (canInject) {
     return activeSession.steer(
       text,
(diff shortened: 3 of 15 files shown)


fransqaas:
Thanks for the review. I do not have access to OpenClaw's Convex QA pool. The documented Telegram userbot harness uses that pool, and the requested-proof workflow currently requires a same-repository PR head, while this PR comes from a fork.

Could a maintainer advise on the supported Telegram proof path for external contributors, or help run the scoped scenario against the final candidate SHA once the CI blockers are resolved?

Proposed proof: use a Test Server DM with the embedded runner, keep a run active with a controlled tool/provider delay, and send a native reply to an older synthetic message. Capture the native reply relationship and active-run timing, then verify that the newly injected user message in the next provider request contains that exact quote. Compare the base and candidate. A deterministic provider can verify context delivery without relying on model interpretation.

A maintainer-run test would keep the shared credentials within your infrastructure. The CI import-cycle failure is also still outstanding; the live proof should cover the final corrected SHA.

fransqaas:
@clawsweeper re-review

The PR body now includes executed before/after proof through the real Gateway and embedded steering runtime on the unchanged corrected head `03cbbd3b09bb0d5d3088bf623bee628f2ff958bb`, with captured provider inputs, lifecycle timing, raw SQLite checks and a reproducible fixture. Exact-SHA CI is green. The remaining native Telegram limitation is documented explicitly.

fransqaas:
Following up now that ClawSweeper has accepted the shared-runtime proof and marked this ready for maintainer review. CI remains green on `03cbbd3b09bb`.

Could a maintainer review the scoped change when available? I can address any remaining feedback. The evidence and its native-Telegram limitations are linked in the PR body.

fransqaas:
Refreshed the contribution against upstream main `d5b46dfea58a`; current head is `4711d6f3d358` and is conflict-free. The shared-type fix still has zero static/runtime import cycles. All 386 focused/interaction tests and the full changed-path checks pass.

I also reran the unchanged real Gateway fixture on the **new head**, rather than relying only on the historical `03cbbd3b09bb` run. Both quoted steers contain their own quote/reply ID exactly once, without the other turn's quote; the original run completes without replacement; the five durable SQLite user entries remain the literal incoming texts. An independent verifier checks original HTTP hashes, RPC/lifecycle order and read-only SQLite data. All inputs are synthetic and the disposable VM has been purged.

[Immutable current-head evidence and reproduction](https://gist.github.com/fransqaas/385698e835960957e66713f14f66f8c9/ef5fcebf2549713d630094134011c3b4af70f826). The earlier red/green comparison remains separately pinned to its original SHAs. This is shared Gateway/embedded-runtime evidence, with the Telegram/model-interpretation limits stated explicitly.

Current-head CI is still running with no failures observed at 21:53 UTC (167 successful, nine running). The PR description now identifies the refreshed candidate and evidence.

fransqaas:
The exact-head Gateway replay is complete on `d3eab48a5b87d7ceeeed102fd404c6fb19a8c333`: [immutable evidence and reproduction](https://gist.github.com/fransqaas/29e798d086a65150ae9316b7336edb0b/a940e84c207f85b0f30c4c4b6f1be07e18f962b2).

The clean Ubuntu VM exercised real WebChat RPC → shared auto-reply → embedded steering → HTTP provider delivery. Each new steer entry contained its own quote and reply ID exactly once, with the other turn's quote and reply ID absent. There were five provider requests and three run starts; the original response completed, and SQLite retained the five literal user inputs. An independent audit checked the original HTTP/RPC/SQLite joins and all 40 exported hashes. The VM was purged. This proves context delivery with a deterministic provider; native Telegram, audio and hosted-model interpretation remain outside the proof.

The same committed tree passed **300 focused tests across 13 files**, the complete applicable changed-path checks (types, formatting/lint, dead exports and repository guards; zero import cycles), and independent scoped review. [ClawSweeper revision 7](https://github.com/openclaw/openclaw/pull/145338#issuecomment-5641292292) has already inspected the new artifact and accepted it as sufficient, with no actionable patch finding.

**CI still has three failed jobs.** They ran merge-ref `3c494ea4` against upstream `736219b5`, not the exact local/Gateway tree:

- [Extensions bundle 21](https://github.com/openclaw/openclaw/actions/runs/35575402924/job/106256357495): two Codex fixture failures, a missing relay and an unexpected abort/timeout. The original 12-file plan passed locally: 390 tests.
- [Compact large 15](https://github.com/openclaw/openclaw/actions/runs/35575402924/job/106256357642): the session-history benchmark child exceeded its 30 s deadline. The isolated unit passed with the deadline unchanged; the full Bun shard/order was not reproduced locally.
- [Compact large 8](https://github.com/openclaw/openclaw/actions/runs/35575402924/job/106256360365): async database cleanup reached an already removed fixture database. The original 32-file plan passed locally: 836 tests.

The PR body records the diagnosis and limits. These macOS/Node 24.16 replays differ from the Linux/Node 24.19 CI merge, so passing results do not establish a fix or unrelatedness. No source, assertions or deadlines were changed; each local replay ran once.

Could a maintainer review the three merge-ref failures and, if a confirmed upstream fixture or runner fix applies, rerun the affected jobs after that fix? The local passes are supplementary evidence; the failed CI results remain recorded.

VACInc:
Related follow-up context: #153829 is now merged, but it does not supersede this PR. #153829 improves `/steer` admission and delivery, but it does not carry `currentInboundContext` or quoted-reply context through the active-run handoff.

Keep this PR open.

PR #145346, opened 2026-09-11
fix(agents): keep internal workers out of persistent sidebar sessions

steipete:
## What Problem This Solves

Fixes an issue where users asking for subagents to perform internal QA, coding, review, or test work could get persistent sidebar sessions and extra categories instead. The spawn guidance treated coding, multi-step work, and producing a PR or report as reasons to create a visible session.

## Why This Change Was Made

Align the spawn tool description, its `visible` parameter, the system prompt, shared delegation guidance, and docs around one choice: internal workers use ordinary subagents and return results to the parent. A persistent visible session is appropriate only when the user requests a separate session or needs to revisit and steer that work independently.

This changes guidance only. Existing session classification, grouping, permissions, worktree capabilities, and runtime lifecycle remain unchanged.

## User Impact

Asking an agent to use subagents no longer instructs it to turn each internal worker into an independently editable sidebar session. A PR/report, long runtime, or isolated worktree alone does not justify persistent session creation.

## Evidence

- Confirmed the reported behavior came from explicit `visible: true` and `group` spawn arguments, while ordinary subagents had no category.
- Updated policy regression fails on the old guidance.
- 392 focused tests passed across delegation guidance, system-prompt construction, spawn-tool behavior, deferred follow-up availability, and the Codex prompt adapter.
- Regenerated the serialized tool and prompt-size snapshots; deterministic snapshot check passed (7 files).
- Focused lint, formatting, diff checks, and independent P2 reviews of the policy and generated snapshots passed.
- Local changed checks passed through the core graph boundary; full typechecking stopped at the compiler containment guard because the linked worktree shares dependencies. Exact-head hosted CI must pass before landing.
- No live model-choice evaluation was run; the tests verify emitted guidance and existing runtime contracts.


steipete:
Merged via squash.

- Prepared head SHA: [d19191c48bb210d7fec86b1d83827a966456948f](https://github.com/openclaw/openclaw/pull/145346/commits/d19191c48bb210d7fec86b1d83827a966456948f)
- Landed commit: [98ac734316fd358fb04baf54fcb61fbde1ecba14](https://github.com/openclaw/openclaw/commit/98ac734316fd358fb04baf54fcb61fbde1ecba14)

PR #145347, opened 2026-09-11
fix(ui): recover session links interrupted by reconnect

steipete:
## What Problem This Solves

Fixes an issue where users opening a session link could remain on a failed or loading panel after the Gateway disconnected during session lookup. The connection recovered, but the destination still required a manual Retry. This affected the Control UI, including the macOS app.

## Why This Change Was Made

The existing application-router lifecycle now remembers the exact session route load interrupted by a disconnect and revalidates it after reconnecting. It shares the existing Gateway subscription and shutdown guard. The failed router run is retired before retrying, so same-match promise reuse cannot swallow recovery.

Newer navigation, Gateway/profile replacement, and router shutdown discard the interrupted destination. Successful conversations remain mounted. Subscription cleanup and reconnect policy are preserved; this change does not replay messages or other mutations.

## User Impact

Session links recover automatically after temporary disconnects without adding browser-history entries or pulling users back from a newer destination.

## Evidence

- [Hosted CI run 34655905931](https://github.com/openclaw/openclaw/actions/runs/34655905931) passed on `f9a3835464c`, including `openclaw/ci-gate`.
- 82 routing, reconnect, and retained-conversation tests passed. Both reconnect timing regressions fail without the fix.
- Two production-bundle browser scenarios passed: click an uncached short session URL, drop the connection during lookup, then either recover the target transcript automatically or preserve a newer Activity navigation. They verify request counts and browser history.
- UI type checks, lint, localization, unused-export checks, and static import-cycle checks passed.
- CI exposed a type-import cycle and a startup-size overrun. Consolidating recovery into the existing router lifecycle removes the extra helper and subscription; the performance limit and baseline remain unchanged.
- Canonical before/after performance checks passed locally under Node 24.20 and Node 26.8.2. The production UI build and independent reviews through P2 are clean.

Live investigation established repeated connection-related navigation failures. The underlying cause of missing transport acknowledgements remains separate; this PR repairs UI recovery. No live Gateway deployment is included.


steipete:
Merged via squash.

- Prepared head SHA: [f9a3835464c18aae7feabec0e856844729a0c2b9](https://github.com/openclaw/openclaw/pull/145347/commits/f9a3835464c18aae7feabec0e856844729a0c2b9)
- Landed commit: [2efa249c405b1eef4049fc348ecc71a55e1acbea](https://github.com/openclaw/openclaw/commit/2efa249c405b1eef4049fc348ecc71a55e1acbea)

PR #145351, opened 2026-09-11
fix(computer): honor capture-free CUA window observations

steipete:
## What Problem This Solves

Fixes an issue where an agent requesting `get_window_state` with `includeScreenshot:false` would still capture the window, or have the request rejected by the node command schema. A CUA window-capture failure therefore also prevented metadata-only accessibility inspection.

## Why This Change Was Made

Carry the existing option through the computer tool, strict node-command contract, and CUA adapter. Omitted and true keep the released request shape and screenshot behavior; only false is serialized, so older strict nodes keep accepting ordinary observations. Metadata-only observations refresh element references without granting pixel-coordinate authority.

The native macOS Peekaboo backend cannot produce capture-free window observations, so it explicitly rejects false instead of silently ignoring the request. Its default/true behavior is unchanged.

## User Impact

CUA-backed nodes can inspect accessibility information without capturing a window. Callers receive an explicit limitation from Peekaboo. Window pixel actions still require a delivered window image and current observation.

## Evidence

- Regressions failed before repair at the tool-forwarding and node-command boundaries; the adapter regression also demonstrated that false still captured after schema repair alone.
- Focused computer-tool, wire-contract, CUA command and reference-lifecycle tests pass, covering omitted/true/false and the metadata-only pixel-input refusal.
- Original full changed-file checks passed. A mixed-version correction has 51 passing focused tests and fresh clean P0–P2 review; its scoped changed-file checks also passed. [Exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34655946535), including macOS Swift tests. ClawSweeper revision 2 found no actionable issues.
- macOS app/shared-kit tests compile successfully and changed Swift files pass lint. The macOS Swift CI job passed with the new decode/refusal tests; tests were not executed on an operator desktop.
- The existing CUA 0.22.2 native contract was source-verified to honor `include_screenshot:false`; no driver dependency or protocol-version change.

Mixed-version regression: a frozen v2026.9.4 window-command schema rejected explicit true on the previous PR head and accepts it after the correction; false remains forwarded to updated nodes.


steipete:
Addressed in af34d89ca072d11722c0541fd9f9e8c9c10d7ce4. The window request validates the boolean but omits explicit true, preserving the released default-capture wire shape. False still reaches updated nodes.

A regression using the frozen v2026.9.4 window-command schema failed on the previous PR head and passes with this correction. Invalid option types are rejected before dispatch. The repair has 51 passing focused tests and a fresh clean P0–P2 review; native implementation files are unchanged. Updated-head CI is running.

steipete:
Merged via squash.

- Prepared head SHA: [af34d89ca072d11722c0541fd9f9e8c9c10d7ce4](https://github.com/openclaw/openclaw/pull/145351/commits/af34d89ca072d11722c0541fd9f9e8c9c10d7ce4)
- Landed commit: [371791b9d6bea4d4b398984621f65684075082e5](https://github.com/openclaw/openclaw/commit/371791b9d6bea4d4b398984621f65684075082e5)

PR #145352, opened 2026-09-11
fix(computer): preserve CUA input evidence and scope guidance

steipete:
## What Problem This Solves

Fixes an issue where CUA desktop actions discarded native effect and escalation evidence and returned only `ok:true`. Agents also saw held-key guidance on nodes that only support taps, and generic background-delivery advice encouraged changing a setting that does not affect CUA desktop input.

## Why This Change Was Made

Project direct SDK action results through the same existing action envelope as targeted window actions. Preserve effect, route, delivery and escalation evidence, identify desktop scope, and explicitly mark a requested delivery mode as inapplicable there.

Build held-key descriptions from the selected node's action list, scope background-delivery guidance to window input, and give actionable errors for unsupported desktop holds. This preserves Linux background window mouse holds and the existing native driver pin. No new held-input API or permission path is introduced.

## User Impact

Agents can distinguish a suspected no-op from an acknowledged input, see the actual delivery evidence, and avoid repeatedly toggling delivery mode on the same desktop route. Tap-only nodes no longer suggest that a duration parameter provides a sustained key hold.

## Evidence

- Regression tests failed before repair for all six direct SDK action paths because effect/escalation evidence was missing.
- Schema/guidance regressions failed before repair when a selected tap-only node still described `hold_key`.
- Focused desktop, modifier, CUA command, public computer schema, and guidance tests pass. Existing frame-authority and targeted-window behavior remain covered.
- Changed-file checks and independent P0–P2 review passed. Hosted CI is still running on the current head.
- This is adapter/contract proof against the existing driver. It does not claim that Doom is playable or that timed holds exist in CUA0.22.2.

### DIFF
--- a/docs/nodes/computer-use.md
+++ b/docs/nodes/computer-use.md
@@ -77,6 +77,10 @@ Loopback is also reachability, not identity: any process on the machine can conn
 
 The CUA descriptor advertises window, element, and browser targets; background and foreground delivery; image, accessibility, and browser observations; and recording. Peekaboo remains the default provider and does not advertise recording.
 
+CUA desktop input uses the foreground desktop route. `deliveryMode` selects a route only for window-targeted input; switching it on a desktop action does not change how that action is delivered. Desktop results preserve native effect and escalation evidence, identify their desktop scope, and indicate when a requested delivery mode is not applicable. An acknowledged input is not proof that the application responded: verify the intended visible change before repeating it.
+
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [37a1dc69d1a124ab29c985a9c9cd7fe0b3f10cd5](https://github.com/openclaw/openclaw/pull/145352/commits/37a1dc69d1a124ab29c985a9c9cd7fe0b3f10cd5)
- Landed commit: [d03bdfb7f9680c7d34de951c9db9d2c9d65403b6](https://github.com/openclaw/openclaw/commit/d03bdfb7f9680c7d34de951c9db9d2c9d65403b6)

PR #145140, opened 2026-09-11
fix: preserve config edits made during Doctor confirmation

shakkernerd:
Closes #145114

<details>
<summary>Additional instructions</summary>

This branch is in `openclaw/openclaw`; repository maintainers can edit it directly.

</details>

## What Problem This Solves

Fixes an issue where users saving config changes while ordinary Doctor waits at **Apply recommended config repairs now?** would lose those edits after accepting the repair. For example, changing `logging.level` from `info` to `debug` could be overwritten by a prepared config that still contained `info`.

## Why This Change Was Made

The confirmation helper now carries the proposal's root config path and content hash to the existing atomic writer, which rejects a changed source under its lock. Doctor consumes the identity after a successful initial write, allowing subsequent repairs to continue.

## User Impact

Doctor preserves a newer root config file, reports that the pending repairs were not written, and exits with an error so the user can rerun it for an updated proposal. Unchanged acceptance and declined repairs retain their existing behavior. Explicit repair and published-updater handoffs retain their existing write policy; included-file ownership rules are unchanged.

The proposal identity is temporary Doctor state. This adds no stored config fields or schema changes.

## Evidence

- All 380 tests in 13 affected files pass on `396a9e501371b906fa7a6e6788d239d2521690a6`, covering confirmation conflicts, refusal paths, migration and persistence siblings, cleanup, and later writes.
- The full changed-path gate passes on that commit: 29 checks, including core typechecking, all 17 core test type graphs, formatting, lint, architecture checks, and Doctor registry checks.
- The four-case conflict regression passes on the candidate. On base `58a6953549d011ffe44dcc3a7f3290b50e64b349`, both file-preservation checks fail while normal acceptance and decline pass. Tests use real config preparation and atomic persistence.
- Independent review of this patch found no actionable P0–P2 issues.
- [CI](https://github.com/openclaw/openclaw/actions/runs/34635295556) and the required merge check pass. [ClawSweeper](https://github.com/openclaw/openclaw/pull/145140#issuecomment-5639014814) rates this commit Platinum with no actionable code findings.
- Real `pnpm openclaw doctor --no-workspace-suggestions` controls pass on Linux at the same commit. Accepting after an external edit exits 1 with the conflict warning, preserves the exact edited bytes, and creates no backup. Unchanged acceptance repairs the config and backs up the original bytes; declining after an edit preserves the edited bytes and creates no backup. Those two controls stop at the observed write or later contribution, respectively; they do not claim a full Doctor run.
[...]

PR #145365, opened 2026-09-11
fix: report config failures after saving accurately

shakkernerd:
Related: #145140

## What Problem This Solves

Doctor could report that accepted config repairs were not written after the file had already been saved. Creating a preferred config path during post-save processing, for example, left the repair in the originally selected file but produced a no-write warning. An optional repair could also swallow a failure after saving and continue with a misleading completion.

## Why This Change Was Made

Config write owners now distinguish a completed file write from a refusal before saving. Failures after a root or included-file save carry the actual target, original cause, and file rollback outcome through runtime refresh and Doctor; they cannot enter the pre-write conflict retry path. Existing write authority, outside-edit checks, rollback timing, and ordinary recoverable warnings remain in place.

When runtime finalization restores the failed write's previous config, Doctor records those restored bytes in the existing updater receipt. This prevents an updater from misclassifying Doctor's own restoration as an outside edit. No updater protocol fields or rollback admission rules change.

## User Impact

Doctor stops when processing fails after saving and reports whether that write was restored, was not restored, or could not be confirmed. Operators can inspect the named file and active config before retrying. Other config-writer callers receive the same outcome reporting. CLI and Doctor documentation explain the distinction.

During an update, restored config remains eligible for normal package and service recovery when the existing ownership and compatibility checks pass. The failed Doctor step remains an error even when recovery succeeds. If a database migration prevents package rollback, the compatible candidate stays installed and the Gateway stays stopped. Restoration of one failed config write does not undo earlier successful repairs; the original whole-update config remains in the pre-update backup.

## Evidence

- The real Doctor CLI reproduced the false no-write warning on base `bad814efcab67d2c9c2057a2a9db176e9292f1ed`. The repaired legacy file, exact original backup, and newer preferred config were checked independently. The corrected flow exits 1, identifies the saved file and not-rolled-back outcome, and preserves all three expected contents.
- The original baseline regression run exposed exactly 19 expected failures. The updater correction separately reproduced two stale restored-hash failures and one refused recovery on the previous implementation, with 49 controls passing and no skips.
- All 896 affected cases across nine test files pass remotely on `65aad57846fe5f9ce5b5554a3b4710a8f3d8058d`, including missing-file restoration, revoked ownership, and outside-edit controls. Final head `1af1a56220417670697381df55074db0030dc59a` adds only an erased type import and config annotation to the rollback test; an independent comparison confirms byte-identical emitted JavaScript and unchanged production source. The complete remote changed-path gate passes on that final head, including all 17 core-test type graphs, production types, lint, formatting, docs, and boundaries.
- The published `openclaw@2026.9.3` updater successfully installed the immutable candidate. A subsequent manual Gateway start reported the matching candidate version/build ID and passed live/ready probes. Three legacy sessions migrated, with config and plugin state preserved.
- A second real published-updater cell used the proposed merge with main `8cf6001608a203dfd7f56a357cc97952d52d656a` (tree `939adf68e0469edaab8a794b38acf00ce0581325`). A filesystem-only fault after saving made Doctor and the ordinary update command exit 1. The error named the written file and confirmed that write was rolled back; actual config bytes and the updater receipt agreed. Read-only observations proved agent database schema 19 before Doctor and schema 20 before the fault and afterward. The candidate package remained installed, the Gateway was stopped, and session, transcript, cron, workspace, and pre-update backup data were preserved. The update reported `state-migrated-no-rollback` with `serviceRestartSafe: false`. This was an unpublished verification composition, not a release.
- Fresh source review found no actionable P0–P2 defects. [Final-head hosted CI](https://github.com/openclaw/openclaw/actions/runs/34659508797) completed with its affected tests and static checks passing. Its sole underlying failure was the independently reproduced main-branch UI startup-size budget failure, tracked separately by #145430; this patch changes no UI or budget inputs.

The regression covers failures after the filesystem writer returns successfully. It does not redefine uncertain failures inside the filesystem primitive or extend Doctor's proposal guard to later writes.

### DIFF
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -556,6 +556,11 @@ Effective changes to `plugins.entries` (or any subpath) require a restart, since
 
 `openclaw config set` and other OpenClaw-owned config writers validate the full post-change config before committing it to disk. If the new payload fails schema validation or looks like a destructive clobber, the active config is left alone and the rejected payload is saved beside it as `openclaw.json.rejected.*`.
 
+If the file is saved but later processing fails, the error names the written file
+and reports whether the write was rolled back. This can name an included file
+when that file owns the edited setting. If rollback did not happen or could not
+be confirmed, inspect the named file and the active config before retrying.
+
 OpenClaw-owned writes that change config reserialize JSON5 as standard JSON. When the source contains comments, the writer warns immediately before removing them; use a direct editor when preserving comments matters.
 
 <Warning>
--- a/docs/cli/doctor/running.md
+++ b/docs/cli/doctor/running.md
@@ -28,9 +28,14 @@ For read-only diagnosis, use `--lint` or bare `--json`. Ordinary `doctor`, inclu
 
 When ordinary `doctor` asks **Apply recommended config repairs now?**, it checks
 that the selected root config file still matches the source of that proposal.
-If its contents or selected path changed, Doctor preserves the newer file,
+If its contents or selected path changed before the write, Doctor preserves the newer file,
 leaves the pending config fixes unwritten, and exits with an error. Rerun
 `openclaw doctor` to review an updated proposal.
+
+If saving succeeds but later processing fails, Doctor stops with an error, names
+the file that was written, and reports whether the write was rolled back. When it was not rolled
+back or recovery could not be confirmed, inspect that file and the active config
+before rerunning Doctor.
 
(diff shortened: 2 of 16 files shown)


shakkernerd:
Fixed: Doctor now reports saved config and rollback outcomes accurately when post-write processing fails.

PR #145368, opened 2026-09-11
test: share installer git reference fixtures

steipete:
## What Problem This Solves

The two installer test suites duplicate the same Git tag-precedence and branch-fallback scenarios, increasing the maintenance cost of keeping those contracts aligned.

## Why This Change Was Made

Move the two named shell-scenario constructors into the existing installer Git fixture module. Keep all four entrypoint tests, their literal expectations, and each installer's setup. The install.sh-specific run_quiet_step override remains supplied inside its test callbacks. The existing full-commit fixture and production installers are unchanged.

## User Impact

No user-visible behavior changes. Both installers retain independent coverage for selecting an immutable release tag over a same-named branch and falling back to a moving v-prefixed branch when no tag exists. No runtime improvement is claimed.

## Evidence

- Both complete installer suites passed before and after: 329 cases on Node 24.19.0 (220 install.sh, 109 install-cli.sh), refreshed after current main changed nvm handling. The original 316-case proof is also retained.
- Actual source-derived constructor evaluation reproduced all four complete shell programs byte-for-byte, including their different newline escapes. Exact source inversion preserves the surrounding calls, assertions, and existing fixture.
- Reported label multiset and per-file order match; reporter labels may be abbreviated.
- Mapped checks passed, including root-test typechecking, formatting, script guards, all three dead-export scans, and lint. Independent review found no actionable findings.

The change removes 100 test lines and adds 67 lines to the existing test helper, for 33 net lines removed.

### DIFF
--- a/test/scripts/install-cli.test.ts
+++ b/test/scripts/install-cli.test.ts
@@ -20,7 +20,11 @@ import { isSupportedOpenClawNodeVersion } from "../../node-version.mjs";
 import { requireNodeTool } from "../helpers/node-toolchain.js";
 import { NODE_RELEASE_VERSION_CASES } from "../helpers/node-version-cases.js";
 import { useAutoCleanupTempDirTracker } from "../helpers/temp-dir.js";
-import { createInstallGitCommitFixtureScript } from "./install-git-fixtures.js";
+import {
+  createInstallGitBranchFallbackFixtureScript,
+  createInstallGitCommitFixtureScript,
+  createInstallGitTagPreferenceFixtureScript,
+} from "./install-git-fixtures.js";
 import {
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6acaf7521965553d322cc48d98b6c4d2f4694ca3](https://github.com/openclaw/openclaw/pull/145368/commits/6acaf7521965553d322cc48d98b6c4d2f4694ca3)
- Landed commit: [5c5a21087beca7aa5e05b0b9895efa3799f93d83](https://github.com/openclaw/openclaw/commit/5c5a21087beca7aa5e05b0b9895efa3799f93d83)

PR #145045, opened 2026-09-11
fix(update): keep updating core when a plugin target is unavailable

steipete:
Fixes #144973
Fixes #145027
Refs #144732

<details>
<summary>Additional instructions</summary>

This is a same-repository branch; maintainers can edit it directly.

</details>

## What Problem This Solves

Fixes an issue where users running `openclaw update` could not update core when one configured npm plugin had no resolvable target or its registry was unreachable, even though the installed plugin could continue working.

## Why This Change Was Made

Preflight uses the loader's existing `openclaw.compat.pluginApi` and `openclaw.install.minHostVersion` contracts. Compatible installed plugins need no registry admission gate. Post-core sync retains their runnable installed version on resolution failures, preserves selectors, and records an advisory in the summary, ledger, and application WARN log, including JSON mode.

If the installed plugin is already incompatible with the target core, a compatible replacement must be resolvable before mutation. Missing targets, returned registry failures, and thrown resolution errors refuse with `plugin-incompatible`, naming the plugin, installed version, requirement, target core, failure, and retry/pin/disable actions. Final-record checks prevent stale retained-version advisories after later repair.

## User Impact

Working plugins can remain installed while core updates finish with `status: "ok"` and a `plugin-target-unavailable` advisory. Known incompatibilities cannot proceed on the assumption that an unreachable registry will repair them later. No new configuration, dependencies, manifest fields, or skip switch is introduced.

Release-note context: keep core updates moving when compatible npm plugins have unavailable targets; retain working plugins with actionable warnings and refuse unresolved incompatible replacements before mutation. Thanks to @blaekmajik, @noma-marketplacedb, and @zhangmianzhong-glitch for the reports.

## Evidence

- New outage regressions failed on 6915: four API-range/host-floor cases returned `ok` instead of the required pre-mutation refusal. After the fix, 245 focused admission, finalization, and plugin-owner tests passed.
- `node scripts/check-changed.mjs`, the full build, and canonical package tarball validation passed. Fresh P1 autoreview of this correction returned `scoped-clean` with no accepted findings. Earlier retention, JSON/ledger/WARN logging, and Doctor tests remain recorded in the worker handoff.
- Candidate head: `1b2fc9626068577595e3d328ff17a0236c560903`; tarball SHA-256: `21e7c99c7960602b49d2172da1dd2ba540d7db94a243e8233b1a80bfcb538b04`. Exact-head CI passed after refreshing the merge snapshot to include #145115; details are recorded below.

[...]

PR #145369, opened 2026-09-11
fix: keep unavailable plugin requirements visible after updates

fuller-stack-dev:
Related #145045

## What Problem This Solves

Resolves a problem where a plugin left unavailable after an update could disappear from final active-payload verification after a failed cohort repair disabled it. The update result needs to retain that unavailable capability and its reason.

## Why This Change Was Made

Add a typed producer assessment that includes active payload failures and plugins still disabled after failed repair. The assessment is diagnostic evidence only: it does not authorize repair or change caller restart, finalization, or update-completion decisions.

The integrated head preserves main's warning policy. For valid configuration, warnings or final plugin error outcomes produce a warning result; otherwise the result is ok. Invalid configuration retains its existing early hard error. Integrity-drift callbacks still abort the affected plugin operation, and authority exceptions still propagate. Required or unknown classifications in the assessment do not promote these warnings into update failures.

## User Impact

A plugin left unavailable after an update stays visible in the update result with its reason; the update outcome itself is unchanged. Authored settings, installation records, repair authority, and the existing best-effort update policy remain owned by their current paths. No configuration option or persistence contract is added.

## Evidence

- The author's regression evidence recorded 10 failures and 11 passes before the status-contract correction, followed by 35 selected passing cases covering assessments, real cohort behavior, CLI integrity warnings, config provenance, and source-Gateway archive restoration. The author also recorded a passing eight-file changed gate.
- The lead's review was scoped-clean at P1. The assessment declaration and implementation at `c0df102967cb5763154634479ce3a6e2b5b5024c` are byte-identical to reviewed revision `b54485f5f8e8938b477d320740f7bb8287498d18`. Inspection of the merge resolution confirms that warning formatting and status calculation retain main's behavior; test expectations were aligned with that behavior.
- Landing verification: `node scripts/run-vitest.mjs src/cli/update-cli/update-command-plugins.degradation.test.ts` passed all 21 tests on the pinned head.
- Fresh CI for the pinned head: https://github.com/openclaw/openclaw/actions/runs/34670226446. The exact-head watcher confirmed green; all fresh checks completed successfully.
- No installed-target activation or end-to-end published-driver update completion is claimed by this PR's source tests.

## ClawSweeper review

The exact-head review reported no actionable code findings. Applied its wording correction: the compatibility description above now reflects main's warning policy. Skipped the requested additional published-driver × integrated-candidate campaign for this explicitly approved diagnostic-only landing; that proof remains a limitation, and the source tests are not presented as published-driver evidence. No code changed during landing, so no re-review request is needed.

Thanks to @fuller-stack-dev for the implementation and regression coverage.


steipete:
Thanks, the visibility half of this is right and wanted: a plugin left disabled by a failed cohort repair must stay listed as unavailable with its reason instead of vanishing from the final active-payload check.

The admission half reverses the policy we landed in #145045, so I cannot take it as is. In `assessPluginUpdate`, any unavailable plugin whose requirement is not explicitly `optional` yields `unsafe` (`plugin-requirement-unknown`), and `unsafe` flips the post-core result to error. `pluginRequirements` is new and no caller supplies it ("Missing is not optional"), so on every real installation every unavailable plugin is "unknown" and the update fails again for exactly the case #145045 fixed: core is already updated, the plugin outcome is recorded, and the operator should get a working Gateway plus a warning, not `doctor-failed`.

The rule we are holding the updater to: only concrete data at risk may fail an update; everything else is a recorded warning. Unknown state is not evidence of risk. So:

- Default for a missing requirement must be `optional` (warning + visible outcome), not `unknown → unsafe`.
- `required` may only come from an explicit, owner-supplied fact. There is no approved surface for that today (no config option, manifest field, or install record), so please leave `required` handling out of this PR, or propose the surface first in an issue; adding a config option needs explicit approval before implementation.
- Consent, integrity drift, unowned payloads, and invalid config stay as they are: those are real risks, and the typed reasons for them are a good addition.
- Keep `optional-repair-needed` and the disabled-after-repair visibility; that is the part operators are missing today.

If you trim the PR to the visibility improvement plus the typed reasons for the existing hard failures, with the default set to `optional`, I will land it today with credit. Happy to do that trim on the branch myself if you prefer; say so and I will push it.

steipete:
Revision b54485f5f8 resolves my concern: the assessment is now separate evidence, and the caller's status contract is unchanged, so an unavailable plugin without requirement information yields a visible, recorded outcome while `openclaw update` still completes. That is exactly the policy. Keeping `plugin-requirement-unknown` as a diagnostic classification is fine as long as it stays diagnostic; whenever an owner-supplied requirement surface is proposed, it goes through the config-option approval first.

I am running the lead review on this head now and will land it once it is clean and the lint lane is green (that lane may be main's own red from #145183, which a fix is in flight for). Thanks for the fast turnaround.

fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [c0df102967cb5763154634479ce3a6e2b5b5024c](https://github.com/openclaw/openclaw/pull/145369/commits/c0df102967cb5763154634479ce3a6e2b5b5024c)
- Landed commit: [67e68e9b5801115fb20429846e91dc942d5eeeb6](https://github.com/openclaw/openclaw/commit/67e68e9b5801115fb20429846e91dc942d5eeeb6)

PR #145380, opened 2026-09-11
fix(zai): reject malformed UTF-8 probe error bodies

xydt-juyaohui:
## What Problem This Solves

Fixes an issue where users on a Z.AI (GLM) connection could have their endpoint probe silently misclassified when the provider returned an error body that was not valid UTF-8.

`detectZaiEndpoint` probes candidate Z.AI endpoints to decide which base URL and model a user's key actually works against. When a probe returned a non-OK response, the bounded error body was decoded with a **non-fatal** `TextDecoder`, so malformed bytes were replaced with U+FFFD and `JSON.parse` succeeded on a body that was never valid UTF-8. The substituted text was then read as `errorCode` and fed to `isUnsupportedModelResult`, where a fabricated `1211`/`1311` value is treated as an authoritative "unsupported model" signal.

The consequence is user-visible: a truncated or corrupted error body from a flaky proxy, CDN, or load balancer could be trusted as a real rejection, causing endpoint detection to move to a fallback model or resolve a different endpoint than the user's plan actually supports. The `catch` block immediately below already documents the intended contract — *"ignore malformed / stalled / oversized error bodies"* — but a malformed **encoding** never reached it, because the decoder did not throw.

## Why This Change Was Made

The probe now delegates its error-body read to the canonical provider JSON reader, `readProviderJsonResponse` from the public `openclaw/plugin-sdk/provider-http` surface, instead of hand-rolling a bounded read plus its own decoder.

That reader already applies both halves of the required contract — it bounds the body via `readResponseWithLimit` and decodes UTF-8 in **fatal** mode — so this change removes local duplication rather than adding a second decoder next to the existing one. The call site keeps its existing semantics by passing `maxBytes: ZAI_DETECT_ERROR_BODY_MAX_BYTES`, the resolved operation deadline via `timeoutMs`, an explicit `onIdleTimeout` preserving the "timed out after Nms" error, and `onOverflow` preserving the "exceeded size limit" error. The now-unused `readResponseWithLimit` import is dropped.

A malformed encoding now throws into the existing `catch`, so the probe degrades to status-only classification exactly as it already does for a non-JSON body.

**Non-goals:** this does not change probe ordering, candidate models, timeouts, byte caps, or any successful-path behavior. It is scoped to `extensions/zai/**`; no core `src/**`, SDK, config, or plugin-manifest surface is touched.

Related: #111810 — the earlier attempt at this same hardening, closed for mock-only proof and for duplicating the canonical reader rather than using it. This PR takes the reviewer's stated best solution directly.

## User Impact

Z.AI users get a correct endpoint/model decision instead of one derived from corrupted text. A malformed error body can no longer be mistaken for a real "model not supported" verdict, so detection no longer silently falls back to a model the user's plan may not expose.

Well-formed responses are unaffected: valid multibyte error bodies still decode and still drive classification exactly as before, and the existing bounded-read, stall, and overflow guards keep their behavior.

### Real-transport proof (before/after, real socket — no mocked Responses)

The prior attempt was closed for supplying only mocked responses. This proof starts a **real `node:http` server on a real TCP socket** and drives the **real exported production entry point** (`detectZaiEndpoint`) against it over a **real HTTP request**, so the bytes travel through an actual HTTP response → undici byte stream → the production bounded reader → the production fatal decode → the production classify path. The only adaptation is rewriting the hardcoded production origin to the loopback server; no `Response` is hand-constructed and no `fetch` result is stubbed.

```
=== real HTTP transport, real exported entry point ===
  valid body bytes   = 60, utf8-valid = true
  corrupt body bytes = 60
  corrupt body fatal-decodes cleanly = false (as expected: genuinely invalid UTF-8)

  valid  : probed=[glm-5.3,glm-5.1,glm-4.7] resolved="glm-4.7"
  corrupt: probed=[glm-5.3] resolved=null

=== assertions ===
  valid body classified via real transport (walked to fallback) = true
  corrupt body not trusted (stopped at first probe, no fallback)  = true

RESULT: PASS (fix holds over a real HTTP transport)
```

The same proof run against unpatched `upstream/main` (`git checkout upstream/main -- extensions/zai/detect.ts`) demonstrates the defect over the same real transport:

```
=== BEFORE FIX ===
  valid  : probed=[glm-5.3,glm-5.1,glm-4.7] resolved="glm-4.7"
  corrupt: probed=[glm-5.3,glm-5.1,glm-4.7] resolved="glm-4.7"   <-- fabricated 1211 trusted
RESULT: FAIL
```

Before the fix the corrupt body is trusted, the probe walks all three candidates, and detection resolves on the fallback as though the rejection were genuine. After the fix it stops at the first probe and returns `null`.

The corrupt fixture is proven genuinely invalid UTF-8: `new TextDecoder("utf-8", { fatal: true }).decode(corrupt)` throws.

### Focused tests (defect-sensitive)

```
$ node node_modules/vitest/vitest.mjs run extensions/zai/detect.test.ts \
    --config test/vitest/vitest.extension-providers.config.ts --reporter=dot

 Test Files  1 passed (1)
      Tests  9 passed (9)
```

Reverting only the production file (`git checkout upstream/main -- extensions/zai/detect.ts`) while keeping the new tests:

```
 Test Files  1 failed (1)
      Tests  1 failed | 8 passed (9)
  [
    "glm-5.3",
+   "glm-5.1",
+   "glm-4.7",
  ]
 ❯ extensions/zai/detect.test.ts:345:19
    345|     expect(calls).toEqual(["glm-5.3"]);
```

Two tests were added: one asserting a corrupt body is no longer trusted as a classification signal, and one asserting a well-formed multibyte body still is. A `makeRawBytesFetch` helper was added because the existing `makeRawBodyFetch` helper takes a `string` and therefore cannot express a body that is not valid UTF-8.

### Existing guard tests still exercise their real paths

The delegation is not silently bypassing the timeout/overflow behavior — the pre-existing guard tests still pass and still drive the real paths, notably:

```
 ✓ fails closed when a probe error body stalls without chunks 366ms
```

### Static checks

```
oxfmt --check extensions/zai/detect.ts extensions/zai/detect.test.ts
  -> All matched files use the correct format.
oxlint -c .oxlintrc.json extensions/zai/detect.ts extensions/zai/detect.test.ts
  -> Found 0 warnings and 0 errors. (2 files, 230 rules)
```

### Scope

```
git diff --numstat upstream/main
87      0       extensions/zai/detect.test.ts
14      10      extensions/zai/detect.ts
```

The production change is a bounded refactor that reduces local complexity: one import swaps to the SDK reader, the local bounded-read + decode block is deleted, and the option mapping is preserved. Two files, both intended. No `CHANGELOG.md` edit and no CODEOWNERS-restricted path.

### What was not tested

No live Z.AI account or real `api.z.ai` request was used — no API key is available in this environment. The proof uses a real HTTP transport against a loopback server that serves real bytes, so the socket, streaming, bounded read, fatal decode, and classification are all exercised for real; only the remote host is substituted. Retry/redirect behavior of the real CDN is out of scope and unchanged by this patch.

<details>
<summary>🤖 AI-assisted PR</summary>
This PR was created with AI assistance. Real behavior proof over a real HTTP transport is included above.
Prompts used: implementation and verification driven by the repo's contributor workflow; root cause traced from the `detectZaiEndpoint` probe path and the `isUnsupportedModelResult` classification contract.
</details>

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -1378,7 +1378,6 @@ extensions/xai/tts.ts	1
 extensions/xai/x-search.ts	7
 extensions/xiaomi/onboard.ts	1
 extensions/xiaomi/speech-provider.ts	2
-extensions/zai/detect.ts	1
 extensions/zai/index.ts	2
 extensions/zai/model-definitions.ts	1
 extensions/zalo/setup-api.ts	3
--- a/extensions/zai/detect.test.ts
+++ b/extensions/zai/detect.test.ts
@@ -71,6 +71,34 @@ function makeRawBodyFetch(map: Record<string, { status: number; raw: string }>) 
   }) as typeof fetch;
 }
 
+/**
+ * Builds a fetch returning a single raw byte body, keyed by `${url}::${model}`.
+ * Unlike {@link makeRawBodyFetch} this takes bytes, so it can express a body
+ * that is not valid UTF-8 at all. `calls` records the probed model ids so tests
+ * can assert how far the probe advanced.
+ */
+function makeRawBytesFetch(
+  map: Record<string, { status: number; bytes: Uint8Array }>,
+  calls?: string[],
+) {
+  return (async (url: string, init?: RequestInit) => {
+    const rawBody = typeof init?.body === "string" ? JSON.parse(init.body) : null;
+    calls?.push(String(rawBody?.model ?? ""));
+    const entry = map[`${url}::${rawBody?.model ?? ""}`] ?? map[url];
+    if (!entry) {
+      throw new Error(`unexpected url: ${url} model=${String(rawBody?.model ?? "")}`);
+    }
+    // Copy into a fresh ArrayBuffer-backed view: BodyInit rejects the
+    // ArrayBufferLike-backed Uint8Array that Buffer/TextEncoder can produce.
+    const body = new Uint8Array(new ArrayBuffer(entry.bytes.byteLength));
+    body.set(entry.bytes);
+    return new Response(body, {
+      status: entry.status,
+      headers: { "content-type": "application/json" },
+    });
+  }) as typeof fetch;
+}
+
 function makeFetch(map: Record<string, FetchResponse>) {
   return (async (url: string, init?: RequestInit) => {
     const rawBody = typeof init?.body === "string" ? JSON.parse(init.body) : null;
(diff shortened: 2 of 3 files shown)


xydt-juyaohui:
Fixed the one red check from the previous head (`checks-fast-baseline-ratchets`).

**Root cause (not a flake, and a real consequence of this change):**

```
Shrink config/assertion-safety-baseline.txt entries (or run with --prune):
  extensions/zai/detect.ts: 0 < 1
```

The old code contained one grandfathered production type assertion:

```ts
const json = JSON.parse(new TextDecoder().decode(bytes)) as { ... };
```

Delegating to `readProviderJsonResponse<{ ... }>` replaces that `as` assertion with a **type argument**, so `extensions/zai/detect.ts` now contributes 0 assertions and its baseline entry must shrink by one. The ratchet was correctly telling me the entry is stale, so I pruned it rather than suppressing the check.

**Change:** `config/assertion-safety-baseline.txt`, `-1` line (`extensions/zai/detect.ts	1` removed).

**Verification:**

```
$ node --import ./scripts/tsx.mjs scripts/check-assertion-safety-ratchet.mts
  ... extensions/zai/detect.ts no longer listed ...
```

`extensions/zai/detect.ts` is fully cleared from the shrink list; the only remaining entries in a local run are files this worktree does not materialize (`ui/**`) plus pre-existing `main` drift in unrelated files, none of which are touched by this PR.

**Re-verified on the new head `17a2ed0c4b56cf09c83529544835efc4627010b5`:**

```
 Test Files  1 passed (1)
      Tests  9 passed (9)

oxfmt --check -> All matched files use the correct format.
oxlint -c .oxlintrc.json -> Found 0 warnings and 0 errors. (2 files, 230 rules)
```

Scope is still bounded: 3 files total (`extensions/zai/detect.ts` +14/-10, `extensions/zai/detect.test.ts` +87/-0, `config/assertion-safety-baseline.txt` -1). The real-transport proof and its before/after contrast in the PR description are unchanged.

xydt-juyaohui:
Status on head `b569cb64a5ed68da93a487626b7c30d616839775`.

Both earlier reds are fixed — `check-test-types` and `checks-fast-baseline-ratchets` were re-triggered on this head after:

1. pruning the now-stale `config/assertion-safety-baseline.txt` entry for `extensions/zai/detect.ts` (the `JSON.parse(...) as {...}` assertion became a type argument), and
2. fixing `TS2345` in the new test helper by copying bytes into a fresh `ArrayBuffer`-backed view, since `BodyInit` rejects an `ArrayBufferLike`-backed `Uint8Array`.

**One check is currently red and it is unrelated to this PR: `control-ui-performance`.**

This job runs the `control-ui-performance-base` lane, which builds the **merge base commit**, not this PR's head, and the failure is a Control UI bundle-size budget:

```
CHECKOUT_BASE_SHA: ee641013ec25853f96e8894ce5a094a93440835d
  violations:
    - startup JS gzip: 346.2 KiB exceeds 346.0 KiB
[control-ui-performance-base] FAILED (exit 1)
```

Evidence that this is a pre-existing `main` condition rather than something this PR introduced:

- The failing commit `ee641013ec2` does not contain any of this PR's changes.
- The identical failure appears on unrelated open PRs, including maintainer PR #145405, whose log shows the same `startup JS gzip: 346.2 KiB exceeds 346.0 KiB` and the same `[control-ui-performance-base] FAILED`.
- This PR's diff is `extensions/zai/detect.ts` (+14/-10), `extensions/zai/detect.test.ts` (+91/-0), and a one-line prune of `config/assertion-safety-baseline.txt`. None of these are in the Control UI dependency graph, so none can change Control UI startup bundle size.

I have not attempted to change the Control UI budget or any Control UI file, since that would be outside this PR's scope and would only mask a repo-wide condition that is already failing on `main` and on other PRs.

The lane has been re-run on this head; every other registered check is green or still queued. I will report the final all-green state once the remaining queued lanes settle.

altaywtf:
Merged via squash.

- Prepared head SHA: [0c666eaedd4d8d57bb338fe99138c98c03888154](https://github.com/openclaw/openclaw/pull/145380/commits/0c666eaedd4d8d57bb338fe99138c98c03888154)
- Landed commit: [ad47ec38f313ac0dd0792c5fe3bf1cab50b11a79](https://github.com/openclaw/openclaw/commit/ad47ec38f313ac0dd0792c5fe3bf1cab50b11a79)

PR #145384, opened 2026-09-11
fix(ui): keep model selection ready without repeated catalog waits

steipete:
## What Problem This Solves

Fixes an issue where returning to New Session or opening a model picker repeatedly waits for the same model catalog. A new draft also showed a loading skeleton even when the agent's default model was already known.

## Why This Change Was Made

Cache successful Gateway catalog projections in browser memory, scoped to the connection, agent, session, account, and requested view. Concurrent consumers with the same timeout budget share one read; successful snapshots are reused across budgets. Gateway-owned deadlines stay intact, and an explicit refresh can still publish after an ordinary read or cooldown expiry. Catalog/config changes, session changes, reconnects, and identity changes invalidate the appropriate copies; failed refreshes remain retryable. Server-reported cooldown deadlines also expire cached readiness without requiring a publication event. The cache is bounded and does not persist credentials or change Gateway authorization. Explicit Debug refreshes continue to sample the Gateway directly.

New Session restores warm controls synchronously and shows its known default during a cold catalog read. On the Gateway, authenticated drafts without personal account links reuse the existing prepared projection; account links are still reread each time, and personal projections remain request-local.

## User Impact

Reopening model controls no longer adds a catalog round trip. A cold catalog read does not hide a known default or prevent composing a new session. Model/account validation remains with the Gateway.

## Evidence

Measured real WebSocket RPCs against an isolated Gateway with trusted-proxy authentication, two agents, two providers, five explicit models, and nine configured model references. The before/after UI catalog owners used the same connection. These are catalog-read timings, not browser paint or provider inference timings.

| --- | ---: | ---: |
| 20 warm catalog reads, median | 21.47 ms | 0.0065 ms |
| RPCs for those warm reads | 20 | 0 |
| 10 simultaneous readers with one timeout budget | 10 RPCs | 1 RPC |

The final 18-scenario probe uses the real Gateway protocol client. It also verifies timeout-budget isolation, warm reuse across budgets, joint model/metadata invalidation, late replies, and explicit-refresh authority. A bounded 1,000 ms request expired while an unbounded request stayed pending with two real Gateway responses held. An active refresh also retained publication authority while 64 other projections completed. Explicit discovery refresh took 22.36 seconds; no cold-discovery improvement is claimed.

An earlier isolated backend comparison reduced authenticated draft metadata median from 29.57 to 9.92 ms. Metadata p95 was higher in its final ten-sample run (32.77 → 50.55 ms), so no tail-latency improvement is claimed. The final merged UI build uses 353,039 bytes gzip for startup JavaScript and passes the existing 355,116-byte enforcement limit. This PR does not raise any budget; metadata lifecycle state remains separate from lazy request/retry code.

All live scenario assertions passed, the final driver exited successfully, and process-group/TCP checks confirmed graceful cleanup. The tested source hashes match the committed merge tree.

Regression checks demonstrate failure on the original code. Focused Gateway account/lifecycle tests, UI cache/cancellation/identity/session tests, and compiled-UI browser flows plus the three-case real-Gateway catalog suite cover recovery, account previews, immediate defaults, and repeated picker opens. The final deadline integration additionally passes 123 focused UI cases and 19 browser scenarios, including real-client deadline recovery. Real Chrome also verifies the live synthetic Gateway's model picker.

Landing validation also exposed timing-dependent test fixtures inherited from main. The test-only corrections distinguish Signal’s absolute deadline from idle timeouts, measure OpenAI’s stalled-response watchdog from the provider request, and control Chrome’s diagnostic classification separately from connection scheduling while preserving real HTTP/WebSocket integration cases. All 119 focused Signal/OpenAI cases and 36 Chrome cases pass. Controlled original failures and missing-deadline/wrong-code checks confirm that regressions remain detectable. Production transport code and timeout limits are unchanged.

Before, while the catalog response is pending:

Model picker waiting for the catalog

After, while the catalog response is still pending:

Known default and usable composer during catalog loading


steipete:
Merged via squash.

- Prepared head SHA: [7982301b804d068b46fcf1e207543425f6002c84](https://github.com/openclaw/openclaw/pull/145384/commits/7982301b804d068b46fcf1e207543425f6002c84)
- Landed commit: [3c194de0e578c214dda988f807ea6b6733032bb8](https://github.com/openclaw/openclaw/commit/3c194de0e578c214dda988f807ea6b6733032bb8)

PR #145385, opened 2026-09-11
fix(heartbeat): recover tombstoned isolated sessions

johnemau:
## Problem

A failed isolated heartbeat can retain a restart-recovery tombstone and block every later forced heartbeat rollover.

## Solution

Permit forced rollover only for an initialized isolated heartbeat session; the fresh rollover drops the stale recovery state.

## Validation

- `node scripts/run-vitest.mjs src/cron/isolated-agent/session.test.ts`
- `node scripts/run-oxlint.mjs --tsconfig config/tsconfig/oxlint.core.json src/cron/isolated-agent/session.ts src/cron/isolated-agent/session.test.ts`


johnemau:
Closing this because it addresses a local deployment/runtime state issue rather than an upstream change. Thank you for your consideration.

PR #145158, opened 2026-09-11
refactor(plugins): deprecate synchronous keyed storage

steipete:
## What Problem This Solves

Plugin authors need an explicit migration path away from synchronous keyed-state operations without losing existing host compatibility or atomic callback behavior.

## Why This Change Was Made

The compatibility registry now names the existing synchronous store factory as the `plugin-state-sync-keyed-store` adapter, with a `next-plugin-sdk-major` removal gate. TypeScript annotations and migration documentation guide callers toward `openKeyedStore` and awaited operations. Removal still requires a supported external-plugin migration and explicit breaking-release approval.

## User Impact

Existing synchronous methods remain available with their current signatures, errors and trust eligibility. No runtime warnings are added. Both APIs share the existing namespace data; no migration copy is needed. Updater and predicate callbacks remain synchronous within their transaction, and optional capabilities remain optional. Callback-free worker operations are a separate contract.

## Evidence

Compiler token comparison confirms runtime code and public types are unchanged while all three deprecation annotations are recognized. Existing real SQLite, runtime trust, CJS/TypeScript loader, SDK type and compatibility tests pass, covering 86 cases with the final metadata recheck. Full changed-file checks and fresh Codex review pass.

Changed pages pass MDX checks. Both new migration links resolve; the global anchor audit's 54 unrelated failures exactly match the untouched baseline. No baseline or runtime behavior was changed to hide those failures.

PR #145392, opened 2026-09-11
refactor(slack): await presence cooldown persistence

steipete:
Related: #145158

## What Problem This Solves

Slack presence polling still uses the deprecated synchronous plugin-state API. An asynchronous storage owner would otherwise let a presence greeting publish before its durable cooldown reservation settles, and let shutdown return while reservation cleanup is still pending.

## Why This Change Was Made

Move the presence cooldown store and its monitor onto `openKeyedStore`. The existing active poll awaits reservation and unused-reservation cleanup, and resolves the newest eligible conversation again after the storage wait. Shutdown drains that same poll before returning. Cleanup conditionally deletes only its own reservation timestamp, preserving a newer cooldown if the original expires while cleanup is pending. Supported older hosts without conditional deletion retain the unused cooldown until its normal expiry.

This is the maintainer-requested bundled-caller migration following sync API deprecation. The namespace, 25,000-entry reject-new policy, eight-hour TTL, and account/workspace/user keys are unchanged. The shared plugin-state owner still executes SQLite synchronously internally; this PR prepares this consumer for its subsequent worker cutover.

## User Impact

Presence events keep their existing routing and cooldown behavior. A storage wait cannot publish a greeting after the monitor has stopped or its conversation has expired or exceeded the automatic participant limit. A newer eligible conversation observed during that wait becomes the publication target.

## Evidence

- 51 cases across the presence monitor, cooldown store, and provider lifecycle suites pass on Node 24.20.0 (136.28 seconds on the current-main composition). Coverage includes workspace partitioning, away-to-active observations, current routing, cooldown suppression, rate limits, and provider shutdown against a real local HTTP endpoint.
- Delayed reservation cases prove no early enqueue/wake, current-target selection, stopped/ineligible/expired cancellation, and drainage of refused enqueue cleanup. The early-publication regression fails against the original monitor because it enqueues before the reservation Promise resolves.
- Synthetic real SQLite proof writes through the plugin's store factory, closes and reopens the database, preserves the cooldown, admits another workspace key, and verifies expiry at the eight-hour boundary. No Slack credentials, network messages, or real Gateway state were used.
- The full changed-file gate passes, including extension production/test typechecks, lint, plugin boundaries, export scans, and storage/import guards.
- The cleanup replacement regression fails the earlier candidate because unconditional deletion removes the newer row. Current tests preserve that row and cover older hosts without the optional conditional-deletion method. Final five-file P2 autoreview is scoped-clean.

Rebased without patch changes onto current main containing #145415. Both commits are patch-identical and all five changed file hashes match the prior reviewed source. The focused/provider suite, Claw review, and current-head CI pass for this composition.

The [current exact-head CI run](https://github.com/openclaw/openclaw/actions/runs/34666649373) passed after one targeted retry recovered a dependency artifact download connection reset. No source, dependency, or UI budget changes were needed for that retry.

### DIFF
--- a/extensions/slack/src/monitor/presence-cooldown-store.test.ts
+++ b/extensions/slack/src/monitor/presence-cooldown-store.test.ts
@@ -35,6 +35,10 @@ describe("openSlackPresenceCooldownStore", () => {
         clock.mockReturnValue(now + 8 * 60 * 60 * 1_000);
         expect(await reopened.lookup("default:T123:U123")).toBeUndefined();
         expect(await reopened.registerIfAbsent("default:T123:U123", now + 1)).toBe(true);
+        expect(await reopened.deleteIf?.("default:T123:U123", (value) => value === now)).toBe(
+          false,
+        );
+        expect(await reopened.lookup("default:T123:U123")).toBe(now + 1);
       } finally {
         clock.mockRestore();
         resetPluginStateStoreForTests();
--- a/extensions/slack/src/monitor/presence-cooldown-store.test.ts
+++ b/extensions/slack/src/monitor/presence-cooldown-store.test.ts
@@ -1,18 +1,52 @@
+import type { OpenKeyedStoreOptions } from "openclaw/plugin-sdk/plugin-state-runtime";
+import {
+  createPluginStateKeyedStoreForTests,
+  resetPluginStateStoreForTests,
+} from "openclaw/plugin-sdk/plugin-state-test-runtime";
+import { withOpenClawTestState } from "openclaw/plugin-sdk/test-state";
 import { describe, expect, it, vi } from "vitest";
 
-const openSyncKeyedStore = vi.hoisted(() => vi.fn(() => ({})));
+const openKeyedStore = vi.hoisted(() => vi.fn((_options: OpenKeyedStoreOptions) => ({})));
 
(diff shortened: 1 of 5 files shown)


steipete:
@clawsweeper re-review

The current head also makes cooldown cleanup conditional on this poll's reservation timestamp, preserving a replacement after the original TTL expires. Hosts lacking the optional conditional-delete method retain the existing cooldown until normal expiry. The PR body now records 51 passing cases, the demonstrated replacement-cleanup regression, real SQLite expiry/replacement proof, and the completed changed-file gate.

steipete:
@clawsweeper re-review

The branch is rebased onto current main containing #145415. Both commits are patch-identical, all five changed files match the previously reviewed bytes, and all 51 focused/provider cases pass again on the current composition. Please refresh review for the current head.

steipete:
Merged via squash.

- Prepared head SHA: [7582a014982713f4d37a9099d08afce0d033af71](https://github.com/openclaw/openclaw/pull/145392/commits/7582a014982713f4d37a9099d08afce0d033af71)
- Landed commit: [83b5bf93d8ee635fb8b01f74d0711acf07e16a39](https://github.com/openclaw/openclaw/commit/83b5bf93d8ee635fb8b01f74d0711acf07e16a39)

PR #145393, opened 2026-09-11
refactor(ui): share healthy Custodian request fixtures

steipete:
Related: #139428

## What Problem This Solves

Custodian nudge tests repeat the same healthy onboarding reply mock in thirteen scenarios.

## Why This Change Was Made

A file-local factory returns the exact existing mock expression. Each case keeps its local request binding at the original position. Mocks and response objects remain fresh across cases, while repeated calls to one mock share its response object. Custom and once-only response chains are unchanged.

## User Impact

No user-visible behavior change. This removes 44 test lines while preserving nudge, recovery, ownership, and delivery assertions.

## Evidence

The full nudge suite passed all 29 tests before and after, with identical ordered names and results. An exact inverse check preserved every other source byte. A control using the installed Vitest spy verified mock and response identities, property order/descriptors, and callback/input identities. Formatting, the mapped checks, and independent review passed.

The first CI run exposed an existing main startup-budget overage. Matched base/merge builds confirmed that it predates this test change. After the independent main update in #145316, the PR was refreshed without changing its test delta, and all 29 cases passed again on Node 24.20. No production or budget edit is included here.

The subsequent inherited picker/dead-export failures were repaired separately in #145415. The latest main refresh preserves the same test bytes and 44-line delta; all 29 cases passed again in their original order with the installed root/UI dependency versions verified.


steipete:
Merged via squash.

- Prepared head SHA: [c1c7ac43434a22d953d7880adb2cabd340b9fe75](https://github.com/openclaw/openclaw/pull/145393/commits/c1c7ac43434a22d953d7880adb2cabd340b9fe75)
- Landed commit: [746fa36967d858c2f44c9a4c70d2b20161b25ff5](https://github.com/openclaw/openclaw/commit/746fa36967d858c2f44c9a4c70d2b20161b25ff5)

PR #145398, opened 2026-09-11
refactor(telegram): await sticker cache storage

steipete:
## What Problem This Solves

Telegram sticker lookup, metadata refresh, search, and statistics still used the synchronous plugin-state API. Their asynchronous message and action handlers could not safely consume a delayed storage implementation.

## Why This Change Was Made

Use the existing asynchronous keyed-store contract and await each bundled caller. Cache failures remain best-effort: delayed lookup failures become misses, list failures return empty results, and failed writes do not fail sticker handling. Successful writes settle before their callers continue.

The namespace, 10,000-entry limit, canonical row format, search scoring/order, and timestamp statistics stay unchanged. These helpers belong to the private bundled Telegram plugin; the deprecated public synchronous plugin-state API remains available to external plugins through its existing compatibility window. This is preparation for the shared storage worker migration: native SQLite execution in the shared KV owner is still synchronous.

## User Impact

Existing sticker behavior is preserved. Telegram can adopt a delayed storage owner without returning promises as action results or losing cache-error handling. No schema, configuration, retention, protocol, or package version changes.

## Evidence

- Six focused test files passed: 242 cases covering real SQLite cache operations, selection, actions, media resolution, and dispatch.
- A deferred-write regression verifies that cache completion waits for persistence and the sticker survives a database close/reopen; delayed failures keep the prior best-effort results.
- Two separate processes wrote and reopened 32 synthetic stickers using the real keyed store. Lookups, search order, statistics, and SQLite integrity checks passed. Local descriptive timings: 177 ms for writes; 89 ms for reopen plus reads/search/statistics. No event-loop performance claim.
- Eight media E2E cases passed, including awaited cache-hit metadata refresh; the final affected three-file rerun passed 181 cases after switching test fixtures to the existing deferred-promise helper.
- Removing the awaits made seven regression cases fail and exposed a rejected write, as expected. The correct source was restored before final verification.
- The full E2E build passed. Fresh processes using the compiled Telegram API and real keyed store also passed the same 32-record persistence/search/statistics/integrity probe (760 ms write, 263 ms reopened reads under load).
- These are isolated synthetic storage and caller-boundary checks; no Telegram credentials or real messages were used.
- The full changed-file gate passed: production/test typechecks, lint, formatting, plugin boundaries, export/deprecation guards, storage guards, and import-cycle checks.

Rebased onto main containing the picker CI repair in #145415. The commit diff and all eight changed files are byte-identical to the reviewed candidate. On the new base, 26 selected cache/caller cases and the fresh-process 32-record SQLite reopen/search/statistics/integrity probe passed. Earlier full-build, compiled-probe, focused-test and changed-gate evidence above predates this patch-identical rebase; current-head CI covers the new composition.

### DIFF
--- a/extensions/telegram/src/action-runtime.test.ts
+++ b/extensions/telegram/src/action-runtime.test.ts
@@ -931,6 +931,37 @@ describe("handleTelegramAction", () => {
     expect(sendStickerTelegram).not.toHaveBeenCalled();
   });
 
+  it("returns sticker search results from asynchronous storage", async () => {
+    telegramActionRuntime.searchStickers = vi.fn(async () => [
+      {
+        fileId: "fox-file",
+        fileUniqueId: "fox-id",
+        description: "A waving fox",
+        cachedAt: "2026-01-26T12:00:00.000Z",
+      },
+    ]);
+    const result = await handleTelegramAction(
+      { action: "searchSticker", query: "fox" },
+      telegramConfig({ actions: { sticker: true } }),
+    );
+    expect(resultDetails(result)).toEqual({
+      ok: true,
+      count: 1,
+      stickers: [{ fileId: "fox-file", description: "A waving fox" }],
+    });
+  });
+
+  it("returns sticker statistics from asynchronous storage", async () => {
+    const stats = {
+      count: 2,
+      oldestAt: "2026-01-20T12:00:00.000Z",
+      newestAt: "2026-01-26T12:00:00.000Z",
+    };
+    telegramActionRuntime.getCacheStats = vi.fn(async () => stats);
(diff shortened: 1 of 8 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d4a3919263bfcb9c0c5c578f80175fd54da86854](https://github.com/openclaw/openclaw/pull/145398/commits/d4a3919263bfcb9c0c5c578f80175fd54da86854)
- Landed commit: [aca7954e00b16e28648ab1c103e7e8420d54446d](https://github.com/openclaw/openclaw/commit/aca7954e00b16e28648ab1c103e7e8420d54446d)

PR #145399, opened 2026-09-11
refactor(comfy): reuse guarded JSON test responses

steipete:
Related: #139428

## What Problem This Solves

Comfy's image, video, and music tests repeat the same guarded JSON response construction even though the plugin already has a shared fixture for it. That boilerplate obscures the response bodies and request sequence each scenario exercises.

## Why This Change Was Made

Export and reuse the existing `fetchGuardJson` test helper at 21 sites. Every response body and queue position stays explicit, and each call still creates its own Response, headers, and asynchronous release mock. Binary responses, custom release assertions, malformed data, real transport coverage, and policy inputs remain explicit.

## User Impact

No user-visible or production behavior changes. This removes 132 lines from tests while preserving all 116 assertion sites in the three provider suites.

## Evidence

- All 70 cases passed before and after across the image, video, music, plugin registration, and header integration suites, with identical case names and statuses.
- Expanding the 21 calls and removing the imports reconstructs the original test files exactly after formatting; the existing helper changes only by adding its export.
- Evaluating the actual helper body confirmed equal JSON bytes, status, headers, construction order, and fresh per-call response/header/release objects for every migrated fixture.
- Mapped changed-file checks and independent review through P2 passed.

### DIFF
--- a/extensions/comfy/image-generation-provider.test.ts
+++ b/extensions/comfy/image-generation-provider.test.ts
@@ -6,6 +6,7 @@ import { buildComfyImageGenerationProvider } from "./image-generation-provider.j
 import {
   buildComfyConfig,
   buildLegacyComfyConfig,
+  fetchGuardJson,
   mockComfyCloudJobResponses,
   mockComfyProviderApiKey,
   parseComfyJsonBody,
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [bd2d5f61697a97945a745236e6103d2aa64be072](https://github.com/openclaw/openclaw/pull/145399/commits/bd2d5f61697a97945a745236e6103d2aa64be072)
- Landed commit: [fb380bd4879be34aa160675d653560bb79690db9](https://github.com/openclaw/openclaw/commit/fb380bd4879be34aa160675d653560bb79690db9)

PR #145402, opened 2026-09-11
fix(agents): show provider rejection details in failed chats

steipete:
## What Problem This Solves

Fixes an issue where users received only “The agent run failed before producing a reply” after an Anthropic-compatible provider rejected a follow-up request, even when the upstream response explained exactly what was invalid. Long JSON error responses were truncated before redaction, leaving an unreadable diagnostic placeholder.

## Why This Change Was Made

Preserve HTTP status and bounded structured error data before redaction and display clipping. Read the terminal upstream rejection from proxy attempt envelopes, and share curated request-limit copy between live failures and saved chat history. Credential and media redaction, the existing 500-character error-body preview, retry timing, compaction-replay rejection handling, and existing recovery-message precedence remain in place.

The earlier CI base already exceeded its startup JavaScript budget; controlled builds of the original PR merge and base produced 1,320 byte-identical browser assets. A bounded Usage-route repair defers page-specific initialization and rendering until navigation, saving 426 gzip bytes without changing budgets. Connection, scope, and date are captured before the import, and request ownership is revalidated afterward. Module-load failures use the existing router error/Retry display; RPC failures retain their inline explanations.

An unchanged MCP doctor regression moves into its existing JSON-failure suite, restoring the 720-root test-group limit without removing coverage. The conflicting New Session/account-picker cleanup was dropped during the required rebase: #145415 independently landed those fixes and the shared picker animation-completion helper on main. This PR retains no duplicate picker implementation or test rewrite.

## User Impact

A request rejected for five cache-control blocks now explains that the provider allows four and the request contained five. Operators retain the redacted upstream diagnosis in logs instead of a hash of a generic placeholder. This PR repairs OpenClaw's error handling; the proxy's extra cache marker is a separate owner-level fix.

## Evidence

- Final exact-head [CI run 34666385581](https://github.com/openclaw/openclaw/actions/runs/34666385581) passed, including the required aggregate gate.

- Regressions failed on the original transport, shared error projection, live error formatter, and persisted chat projection; the repaired paths pass.
- Focused transport, redaction, failover, and saved-history tests passed, including Mistral preview compatibility. The final preview-cap correction passed 485 tests; the Usage/router boundary passed 16 tests, covering import-time retirement, navigation/preload cancellation, and redaction.
- `node scripts/check-changed.mjs` passed for the final combined core, test, UI, and docs changes, including typechecks, lint, unused-export scans, and runtime import cycles.
- Independent reviews found no actionable P0–P2 findings. The provider/Usage patch remained identical across the CI-driven rebase; the final guard and CI repairs passed separate fresh reviews.
- Fresh live candidate transport received the real provider HTTP 400 / 2,282-byte response from a synthetic image-and-tool-result request. It retained the status and exact maximum-four/found-five reason in its complete error message while keeping the separate preview at 500 characters plus its truncation suffix. A transient provider 503 cleared on one retry of the same route. Replaying the previous clipping against the same response reproduced the unreadable placeholder.
- The repaired head passed hosted CI’s Control UI performance check. Local measurement was 353,996 B startup gzip versus its 354,270 B limit; startup CSS remained 45,842 B. A browser module-load failure probe showed a visible Retry action, redacted diagnostics, and zero Usage RPCs.
- Seven cloud recovery/dispatch browser cases pass after adopting the upstream picker helper.
- The cursor and picker guard repairs are now inherited from main through #145415; their overlapping commits were removed. The retained provider/Usage patch and MCP test consolidation have identical patch IDs across the conflict-driven rebase.
- All nine consolidated MCP JSON-failure cases passed, including the unchanged moved directory/PATH case; the test ownership guard passed at 720 roots.
- Actual live-failure formatting and saved-history projection both produced the specific cache-limit explanation. The real Control UI rendered these results through synthetic Gateway fixtures, with zero browser page errors. These are UI boundary captures, not production-Gateway screenshots.

Before:

Before: generic failure in synthetic Control UI fixture

After:

After: specific cache limit in synthetic Control UI fixture


steipete:
Merged via squash.

- Prepared head SHA: [e84be2c3cce6508dc782587d91c33c6e4f572bad](https://github.com/openclaw/openclaw/pull/145402/commits/e84be2c3cce6508dc782587d91c33c6e4f572bad)
- Landed commit: [c9b5718ff2c8c6465f40d1a1203bb5e68aa996fd](https://github.com/openclaw/openclaw/commit/c9b5718ff2c8c6465f40d1a1203bb5e68aa996fd)

PR #145404, opened 2026-09-11
test(outbound): avoid repeated channel resolution test imports

steipete:
Related: #139428

## What Problem This Solves

The outbound channel resolution suite reloads its dependency graph for every case, adding avoidable work to local tests and CI.

## Why This Change Was Made

Load the graph once and reset the existing bootstrap owner's generation and registry caches before each case and at teardown. Keep all 27 cases, mock defaults, assertions, and scoped-registry checks. This removes 17 net test lines without adding a production seam.

## User Impact

No product behavior changes. Developers retain the same channel resolution and cache isolation coverage with less repeated test setup.

## Evidence

- Complete file before/after: 27 passing cases in identical order.
- Normal non-isolated sibling cohort with `channel-bootstrap.runtime.test.ts`: 46 passing cases, including generation changes, cached misses, and scoped sender ownership.
- `node scripts/check-changed.mjs -- src/infra/outbound/channel-resolution.test.ts` passed.
- Diff cleanup and independent review passed, including import timing before mock defaults and teardown isolation.
- One local Node 26.8.2 timing pair with separate initially empty Vitest caches: 5.961s before, 5.560s after. This is a single-file observation, not an overall suite speedup claim.

### DIFF
--- a/src/infra/outbound/channel-resolution.test.ts
+++ b/src/infra/outbound/channel-resolution.test.ts
@@ -1,6 +1,6 @@
 // Verifies outbound channel resolution fast paths, active-registry reads,
 // bootstrap fallback, and runtime facade projection.
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
 import { withPluginRuntimeRegistryScope } from "../../plugins/runtime/gateway-request-scope.js";
 import {
   createChannelTestPluginBase,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d0c04c1f795c14d3504e01b6ad7479211c8d0ef6](https://github.com/openclaw/openclaw/pull/145404/commits/d0c04c1f795c14d3504e01b6ad7479211c8d0ef6)
- Landed commit: [6fab1939ecbecbaefbfafaa5590dfb4050c47165](https://github.com/openclaw/openclaw/commit/6fab1939ecbecbaefbfafaa5590dfb4050c47165)

PR #145405, opened 2026-09-11
fix(control-ui): prevent New Session reconnect freezes

steipete:
## What Problem This Solves

New Session could become blank and unresponsive when reloaded with cached boot state while the Gateway reconnects. A failed configuration read could trigger repeated work and leave capability controls loading indefinitely. Account pagination also needed to expose pending work without dropping keyboard focus between pages.

## Why This Change Was Made

New Session subscribes to the existing runtime-configuration owner. Rendering reads its recorded state and error; connection bootstrap and reconnect handling retain responsibility for loading and refresh.

More uses aria-disabled while pending so it retains focus. The existing inventory owner rejects duplicate activation, and shared search and keyboard selection skip unavailable rows. Global disabling remains native. Regression coverage preserves current main's empty-inventory refresh, cached reopening, picker-opening behavior, and selected-machine capture checks.

## User Impact

Users can keep editing their draft during reconnect or after a configuration-read failure, then continue after recovery. Keyboard users can load consecutive account pages without refocusing, duplicate requests, or changes to the chat route or default account.

## Evidence

Real Chrome originally reproduced a blank page and a renderer consuming a CPU core. Final verification in the existing Chrome profile, through the extension relay with synthetic Gateway responses, demonstrated editable cached startup before the handshake completes, a settled configuration failure without repeated reads during editing/menu use, and draft preservation after reconnect.

Actual Tab and Enter input verified page two to page three without refocusing, ignored repeated activation while pending, an unchanged URL and selected account, and retained Manage focus when Loading and More disappear. The screenshots below are inspected synthetic fixture captures; the first records the original blank reload and the second shows editing while the handshake is held.

Validation passed 23 integrated bundled browser cases plus the capture-enabled cloud-dispatch case, targeted UI production/page/E2E types, typed lint, formatting, and independent reviews through P2. The prior 275 relevant unit cases and complete changed-file check remain valid for byte-identical production and prior regression files. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34669386180) passed on its first attempt: 87 jobs passed, 13 skipped, no retries.

No desktop-startup percentile or live provider/account-mutation claim is made. OpenClaw release-note context: fixes New Session reconnect freezes and preserves account-pagination keyboard focus.

Before: New Session freezes during reconnect

After: editing the draft while the synthetic Gateway handshake is held


steipete:
Merged via squash.

- Prepared head SHA: [2c5b02f163b79aca77eb896600d7238a2d01c999](https://github.com/openclaw/openclaw/pull/145405/commits/2c5b02f163b79aca77eb896600d7238a2d01c999)
- Landed commit: [23e6b1b30d056880bd0dc8f9a0f1e141444c413a](https://github.com/openclaw/openclaw/commit/23e6b1b30d056880bd0dc8f9a0f1e141444c413a)

PR #145360, opened 2026-09-11
fix(state): retain read-only snapshots until native cleanup succeeds

steipete:
## What Problem This Solves

Resolves a problem where read-only shared-state inspection removed its temporary snapshot even when SQLite failed to close the connection. A refused schema or integrity check could also leave an open reader without an owner able to finish cleanup.

## Why This Change Was Made

Runtime and maintenance readers now share the existing native connection-close owner. Snapshot cleanup runs only after the native handle closes, and failed close or artifact cleanup stays retained for lifecycle retry. Runtime schema admission and maintenance integrity validation keep their existing policies, including preservation of the original validation error and the Boolean close/retry contract.

## User Impact

Read-only state inspection retains its snapshot while SQLite still needs it, then releases the handle and temporary files when cleanup succeeds. This change consolidates native reader ownership; it does not move SQLite execution to a worker or change schema, configuration, or dependencies.

## Evidence

Both new entry-point regressions failed on the original base because the snapshot directory disappeared while the native connection remained open, and passed with this fix. The reviewed patch passed 267 applicable owner/sibling tests (one preexisting Linux-only case skipped on macOS), changed-code checks including core and test typechecks, lint, dead exports and schema/import guards, and isolated Codex review with no actionable findings.

Source-native smoke passed on Node 24.20.0, Node 26.8.1, and Bun 1.4.2, covering missing-file behavior, warm/cold read parity, close-failure retention, lifecycle retry, reopen, and file exclusion. The single-commit rebase onto current main is patch-identical; focused close/retry regressions were rerun on the resulting graph. No full package build or worker-execution proof is claimed.

Exact-head [CI run 34654820481](https://github.com/openclaw/openclaw/actions/runs/34654820481) passed. Its first attempt failed an unchanged OpenAI speech timeout case; all four focused timeout cases passed locally, and one unchanged failed-job rerun passed. The native review artifacts and exact-head ClawSweeper review have no actionable findings.

PR #145353, opened 2026-09-11
refactor(backups): await recording and prepare status freshness

steipete:
## What Problem This Solves

Backup commands and status reporting must wait for storage results as the shared SQLite owner adopts asynchronous execution. The current archive, SQLite snapshot, Git, status, and Doctor callers assume that backup outcome writes and freshness reads finish synchronously.

## Why This Change Was Made

The backup state owner now exposes asynchronous outcome recording and a single freshness operation that reads the latest attempt and latest successful run together. Backup commands await recording inside the existing best-effort error boundary; status and Doctor await freshness before formatting it. The synchronous overview renderer receives prepared facts instead of reading SQLite.

## User Impact

This is a prerequisite for moving shared-state operations off the main thread. SQLite still executes synchronously in this change. Backup output and the bounded 200-attempt history remain unchanged, missing databases are not created for logging or status, and a recording failure remains a warning without changing the backup result.

## Evidence

- Current-head hosted CI passed: [run 34666772631](https://github.com/openclaw/openclaw/actions/runs/34666772631), including Windows and the required aggregate gate.

- Current refreshed graph: 23 selected backup-recording, archive-ordering, text-status, and Commander-routing cases passed. The core typecheck shard-boundary guard and diff whitespace check also passed.
- Earlier retained backup, JSON/status, and Doctor sibling proof: 48 passed, with one existing Windows-only Git case skipped on macOS; the final focused follow-up passed 11 cases.
- The delayed-recording regression fails with the original archive command because success is printed before recording settles; the candidate passes.
- Independent P2 review was scoped-clean. The latest prerequisite rebase preserves the entire patch and all 19 candidate file contents; it introduces no new backup implementation changes.
- Earlier full changed-code checks passed, including production/test types, dead exports, lint, schema/legacy-store guards, and import cycles.
- An earlier canonical package build, tarball integrity, and import-graph check passed. The published 2026.9.3 updater installed that candidate package and completed candidate Doctor subprocesses; archive and SQLite backups recorded outcomes, JSON status returned freshness, and copied state passed integrity checks.

[...]

PR #145406, opened 2026-09-11
test: drain interrupted speech requests before fixture reset

steipete:
Related: #145353, #145360. Maintainer-directed follow-up to the database cleanup campaign in #144592.

## What Problem This Solves

Resolves a problem where a failed speech timeout assertion could leave its synthesis operation running while the test fixture reset the plugin registry. The watchdog rejected the validation promise, but the underlying synthesis promise had no teardown owner.

## Why This Change Was Made

The hanging-server fixture now starts and owns synthesis separately from validation. It closes its loopback sockets before joining synthesis settlement, and then preserves the original validation failure. Future watchdog errors include the elapsed time, received request count, and connection-closed state already owned by the test.

The 100 ms caller timeout, 1000 ms watchdog, and 2000 ms case limit are unchanged. This fixes test cleanup; it does not claim to explain or repair the initial hosted watchdog failures.

## User Impact

No runtime behavior changes. Failed timeout tests leave their request and plugin lifecycle settled before the next test begins.

## Evidence

- A controlled real-loopback regression failed with the original non-joining fixture because the fixture settled while request cleanup was still held. It passes after the repair and verifies that the original watchdog error survives teardown.
- The focused speech contract suite passed all 18 cases, including the original ordinary/telephony × headers/body matrix.
- All 66 OpenAI provider files passed (1,193 tests). The changed-file gate and independent review through P2 passed.
- Typechecking caught a direct Promise.withResolvers library-target mismatch in the new test; the final version uses the existing createDeferred helper.
- Temporary local timing capture showed request cancellation about 100 ms after request creation in all four cases; the diagnostic hooks were removed before review. This local evidence does not identify the failing hosted stage.

### DIFF
--- a/extensions/openai/speech-runtime.contract.test.ts
+++ b/extensions/openai/speech-runtime.contract.test.ts
@@ -3,6 +3,7 @@ import { randomUUID } from "node:crypto";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import type { OpenClawConfig } from "openclaw/plugin-sdk/config-contracts";
+import { createDeferred } from "openclaw/plugin-sdk/extension-shared";
 import {
   createEmptyPluginRegistry,
   setActivePluginRegistry,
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8cd458bb1af9e8126677b74a43b61b41e4c511ce](https://github.com/openclaw/openclaw/pull/145406/commits/8cd458bb1af9e8126677b74a43b61b41e4c511ce)
- Landed commit: [0e3d81537c6a51a7e5cb32622060ef935de63cf9](https://github.com/openclaw/openclaw/commit/0e3d81537c6a51a7e5cb32622060ef935de63cf9)

PR #145431, opened 2026-09-12
fix(ci): rebalance config-cli test roots into state-logging

steipete:
## What Problem This Solves

`origin/main` is red on `check-additional-boundaries`: the `config-cli` core test typecheck shard holds 721 test roots against its 720-root memory bound (main run [34659892235](https://github.com/openclaw/openclaw/actions/runs/34659892235)). PR CI builds the merge commit, so every open PR inherits the failure. Same shape as #139645.

## Why This Change Was Made

Move the `src/config/sessions` test roots from `config-cli` to `state-logging`, which already owned the cold-storage subset of that directory and has the most headroom. Only the two shard include/exclude lists change; the 720-root bound and exactly-once coverage stay enforced.

## User Impact

PRs stop inheriting this unrelated failure from `main`. No runtime behavior change.

## Evidence

Reproduced on current `origin/main` (45bd3a0e6834) before the change, then verified after:

```sh
node --import ./scripts/tsx.mjs scripts/check-tsgo-core-boundary.mts
node scripts/run-tsgo.mjs -p test/tsconfig/tsconfig.core.test.state-logging.json
```

| Shard | Roots before | Roots after |
| --- | ---: | ---: |
| config-cli | 721 | 579 |
| state-logging | 207 | 349 |

Before: exit 1, `config-cli: 721 test roots exceeds the 720 limit`. After: boundary check exit 0; the state-logging shard typechecks clean with the moved roots.

Note: main is separately over the Control UI startup budget; #145316 carries the approved baseline rebase for that lane.

PR #145441, opened 2026-09-12
fix: keep session test typechecks below the shard limit

steipete:
## What Problem This Solves

Resolves a problem where the core typecheck boundary fails on current main because the config/CLI shard has 721 test roots, exceeding its 720-root limit. This blocks unrelated changes before their validation can finish.

## Why This Change Was Made

Assign all session-store tests to the existing state/logging shard, which already owns the cold-session storage tests. Replace the narrow session exclusions and inclusions with matching subtree patterns so new session tests have one clear typecheck owner.

## User Impact

Developer validation can run again with the same root limit, shard count, compiler options, and CI fanout. Runtime behavior is unchanged.

## Evidence

The existing boundary guard fails on the original main checkout with 721 config/CLI roots. After this change, config/CLI has 579 roots and state/logging has 349. The complete canonical-root ownership and extension-import boundary passes, confirming that every test remains assigned exactly once across the existing 17 shards. Both affected typechecks and the full changed-scope gate pass. Independent P2 review is clean.

### DIFF
--- a/test/tsconfig/tsconfig.core.test.config-cli.json
+++ b/test/tsconfig/tsconfig.core.test.config-cli.json
@@ -14,7 +14,6 @@
     "../../dist",
     "../../**/dist/**",
     "../../src/cli/daemon-cli/**",
-    "../../src/config/sessions/session-cold-storage*.test.ts",
-    "../../src/config/sessions/session-accessor.sqlite-cold-read.test.ts"
+    "../../src/config/sessions/**"
   ]
 }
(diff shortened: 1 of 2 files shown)


steipete:
Superseded by #145431, merged as 34cdbaf303542717f5521ef4da504d57e20b644c while this PR was being validated. Both changed configuration files are byte-identical to that merged commit, and the merge is on current main.

Independent validation here also passed: the original 721-root failure, repaired counts of 579/349, the complete canonical-root and extension-import boundary, both affected typechecks, the full changed-scope gate, and P2 review. The existing root cap and 17-shard inventory remain unchanged. Closing this duplicate; no further patch is needed.

Issue #145152, opened 2026-09-11
Stuck-session recovery reports a force-clear as an abort, names no run/owner identity, and releases the reply lane by session id (2026.7.1)


merit-blake:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

On 2026.7.1, a Google Chat DM lane held queued inbound messages for 80+ minutes after its runs had completed, releasing roughly one message per 6-minute stuck-session tick. The completion-boundary marker leak itself is the class already tracked in #91909 / #96834 — this report is about the **stuck-session recovery path**, which fired ~67 times that day and, in every one of those lines, reported a success it had not achieved.

Filing separately from #96834 because the two asks are independent: #96834 asks that the boundary stop leaking the `active_reply_work` marker; this asks that when recovery *does* fire, it name what it acted on, release the lane by the key the registry is keyed by, and verify the release. Even with #96834 fixed, items 1–5 below leave the recovery path unable to report a real reclaim from a no-op, which is what made this incident invisible for 80 minutes.

Five defects, all in `src/logging/diagnostic-stuck-session-recovery.runtime.ts`, `src/logging/diagnostic-session-recovery.ts` and `src/agents/embedded-agent-runner/runs.ts`:

**1. A force-clear is reported as an abort.** `recoverStuckDiagnosticSession` builds the outcome as `aborted || forceCleared ? {status: "aborted", action: "abort_embedded_run", …}`, so a force-clear emits `status=aborted … aborted=false`. Every one of the ~67 lines read as a successful abort to anyone grepping `status=aborted`.

**2. No run or owner identity in the recovery lines.** No run id, no resolved handle session id, no reply-operation identity, no lane queue depths on either side of the release. With those absent, a recurring reconciliation race and a single immortal orphaned handle produce byte-identical logs — we could not tell them apart from 75 lines of evidence, and the distinction decides the fix.

**3. The force-release is keyed by session id; the registry is keyed by session key.** `forceClearEmbeddedAgentRun` → `forceClearReplyRunBySessionId` resolves the owner through `replyRunState.activeKeysBySessionId`. The lane it must free is `replyRunState.activeRunsByKey[sessionKey]`. When those two disagree, the clear frees nothing and still returns truthy from the handle branch, so recovery reports `forceCleared=true` for a lane that is still held. `ReplyOperation.updateSessionId` (e.g. a compaction rebinding the session id mid-operation) is one way they diverge.

[...]

Issue #96834, opened 2026-06-25
WhatsApp 1:1: inbound image wedges main lane ~3min before processing — multimodal run strands active_reply_work/queued_work_without_active_run (repro on 2026.6.10, post-#95039)


aleps001:
### Bug type

Behavior bug (incorrect output/state without crash)

### Summary

On **WhatsApp direct (1:1)**, sending an **image** to the main agent wedges the message lane for ~3 minutes **before any real processing of the prompt starts**. The inbound image is injected as a **native multimodal image block** into the main agent's run, so the run that ingests it is itself an `embedded_run`. When that run reaches `run:completed`, its active-reply marker is **not cleared**: the lane stays `state=processing` with `reason=queued_work_without_active_run`, and the stuck-session recovery **skips with `active_reply_work` (`action=keep_lane`)** for ~3 min before finally escalating to `abort_embedded_run` and releasing the lane. Only after that does the agent route the request to a worker.

This is the same reason-code signature as the canonical follow-up-drain race **#91909** (`queued_work_without_active_run` → `keep_lane`/`active_reply_work` → `abort_embedded_run`), which was closed as **fixed by #95039**. However:

- The installed build **2026.6.10 (`aa69b12d`) already contains #95039** (merge commit `023993249f...`, `behind_by=0`), yet this still reproduces.
- #91909 is described as a **general** race triggered by a fast (~1–2 s) text follow-up on Slack threads. This report is a **distinct trigger**: an **inbound image (multimodal block)** on a **WhatsApp 1:1 lane**, and the lane wedged here **without** a rapid user follow-up — the queue grew while the user was idle, behind a multimodal `embedded_run` that never cleared.

Filing as a focused, image/multimodal-specific variant and cross-referencing #91909 / #95039 so maintainers can either fold it into the canonical tracker or treat it as a surviving variant of the same race.

### Environment

- OpenClaw **2026.6.10** (`aa69b12d`), self-hosted Linux, systemd **user** gateway, loopback bind
- Channel: **WhatsApp direct (1:1)**
- `messages.visibleReplies=automatic`
- Main agent model: `gpt-5.5`
- Queue mode: default (`steer`)
- `openclaw tasks list`: `0 queued · 0 running · 0 issues` **during** the wedge (the wedge is invisible to the task subsystem)

### Steps to reproduce

1. In a WhatsApp **direct** chat, send the main agent **an image** (no caption / "what is this?" style).
[...]

Issue #91909, opened 2026-06-10
[Bug]: Inbound dispatch race: queued message after run completion never starts a new run (`queued_work_without_active_run`), agent goes silent until stuck-session watchdog fires


SFVVC:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

On a chat thread (observed on Slack channel threads), when an agent finishes a reply and a user sends a new message into the **same thread** within a few seconds, the new message is enqueued (`queueDepth` increments) but **no new run is ever started to process it**. The session sits with queued work and no active run. ~5 minutes later the stuck-session recovery watchdog detects `reason=queued_work_without_active_run` / `classification=stale_session_state` and force-aborts the session. The agent is silent for that entire window, then recovers.

Reproduced across **two different agents in two different channels**, and a **third time on the subagent (`mode:run`) completion path** (child completed but its completion event never reached the parent). So this is a general inbound-dispatch race, not agent- or channel-specific.

**Version:** 2026.5.28 (commit e932160), self-hosted, Slack provider, queue mode `steer` (default).

### Steps to reproduce

1. Have an agent reply in a Slack channel thread (default queue mode `steer`).
2. As soon as the reply completes, send another message into the same thread (within ~1–2s of completion).
3. Observe: the new message is acknowledged/enqueued but no run starts. The agent never replies.
4. ~5 min later a `stuck session recovery` / `abort_embedded_run` fires for that session.

(Timing-sensitive — easiest to hit during fast back-and-forth conversation.)

### `enqueue-followup` branch defers without a restart guarantee
`agent-runner.runtime-*.js` → `runReplyAgent`:

```js
if (activeRunQueueAction === "enqueue-followup") {
  enqueueFollowupRun(queueKey, followupRun, resolvedQueue, "message-id", queuedRunFollowupTurn, false); // restartIfIdle = false
  const queuedBehindActiveRun = isRunActive?.() === true;
  if (!queuedBehindActiveRun) scheduleFollowupDrain(queueKey, queuedRunFollowupTurn);
  ...
  return;
}
```

- `enqueueFollowupRun` is called with `restartIfIdle=false`, so its own tail guard `if (restartIfIdle && !queue.draining) kickFollowupDrainIfIdle(key)` is skipped.
- Whether a drain starts then hinges on a **single non-atomic snapshot**: `queuedBehindActiveRun = isRunActive() === true`.
[...]

PR #145444, opened 2026-09-12
fix(agents): recover the reply owner named by the session key

MertBasar0:
Refs #145152

Revision 3 addresses both remaining blockers: the non-stuck cancellation defect, and the authority-chain proof.

## What Problem This Solves

Stuck-session recovery and the cron timeout cleanup share one exported helper, `abortAndDrainEmbeddedAgentRun`, which is given the session key it must act on but addressed every owner lookup by session id. Two non-exclusive indexes are involved:

- `ACTIVE_EMBEDDED_RUNS` is keyed by session id, so the captured handle belongs to whichever run registered that id.
- The reply registry's `activeKeysBySessionId` is written **unconditionally** by `createReplyOperation`, while `updateSessionId` guards the same write and throws `Cannot rebind reply operation … to active session …`. Two operations on different keys that share a session id — a reset or fork reusing it — repoint it at the later one.

Two distinct failures follow:

1. **Another key's live run is destroyed.** Recovery for the wedged key deleted that run's handle, removed its run-id registration, closed its diagnostics and notified its waiters while the backend kept executing, and it expired and force-cleared that key's reply operation.
2. **The selected owner's slot is released without cancelling it.** Operation-addressed expiry was gated on `reason === "stuck_recovery"`, so `cron_timeout` (`server-cron.ts:998`, `cleanupTimedOutAgentRun`) and callers with no reason fell through to `forceClearReplyOperation()`, which only fails and clears the owner — it cancels neither its abort signal nor its backend. The released run keeps executing while its slot is free to admit successor work.

## Change

- Validate the captured handle against its recorded `sessionKey`; a foreign handle is excluded from every cleanup side effect.
- Fence the session-id-addressed `abortEmbeddedAgentRun` / `waitForEmbeddedAgentRunEnd` on that id resolving to this call's owner — both the reply owner and the handle. With no handle registered, the handle check alone still let them reach whichever owner the reply index resolved to.
- Expire the owner whenever this call is about to release it, through the existing operation-addressed primitive, using the `no_activity` staleness reason outside recovery, and await its settlement on the same path.

- Require the owner resolved by key to have owned the session id the caller supplied. Resolving by key alone reaches whoever holds the key *now*: once an execution ends, the same conversation can admit a new one under a different session id, and a late sweep still carrying the old id would expire that successor. `hasOwnedSessionId` is lifetime-wide, so compaction rebinding — which legitimately changes the id mid-execution — still resolves.
- Re-read the foreign-owner check at the point of use instead of once up front. Expiry can complete the owner synchronously and the helper then yields at `setImmediate`; a replacement registered under the same session id during that window was being aborted on the strength of a snapshot describing the owner it replaced. The check now also rejects an id that resolves to a handle this call never captured.

Callers that pass no `sessionKey` keep the previous session-id resolution. The expiry is a no-op for an owner that already committed a terminal result, so callers that were already safe are unchanged.

**On the `no_activity` reason:** the union is `terminal_unreleased | finalization_stalled | no_activity | stuck_recovery`, and `no_activity` is what `reply-turn-admission.ts:156` already uses to reclaim an operation with no progress evidence, which is the cron-timeout shape. Happy to change it if maintainers prefer a different mapping — it is the one judgement call in this patch.

## Scope

Still out of scope, and not claimed: reclaiming a retained owner that already committed a terminal result (the `active_reply_work` marker-leak class, #91909 / #96834); defect 5; and defects 1 and 4 together — `recoveryOutcomeClearsQueuedSessionState` treats `status: "released"` as unconditionally clearing queued session state but requires `released > 0` for `status: "aborted"`, so correcting the status alone would silently change queue-drain behavior.

## Evidence

Verified against the branch base with the real registry and the real exported helper — no mocks.

### Authority chain, non-stuck path

A harness drives the real modules through the same call `cleanupTimedOutAgentRun` makes (`reason: "cron_timeout"`, `forceClear: true`), with a live owner on one key and a foreign owner sharing the session id. No gateway, no provider, no network. Backends complete when cancelled, as a real backend does.

```
before release
  owner.phase                                    running
  owner backend cancels                          0
  bystander.phase                                running
  bystander backend cancels                      0
  slot held by owner key                         shared-session-authority-chain
  slot held by bystander key                     shared-session-authority-chain

[diagnostic] reply run stale takeover: forced release sessionKey=agent:main:dm-cron-timeout
             reason=no_activity phase=failed result=failed:run_stalled ageMs=1 ranForMs=1

after release (reason=cron_timeout, forceClear=true)
  helper result                                  {"aborted":true,"drained":false,"forceCleared":false}
  owner.phase                                    failed
  owner.staleExpiryReason                        no_activity
  owner backend cancel reasons                   ["superseded"]
  owner slot released                            true
  bystander.phase                                running
  bystander backend cancels                      0
  bystander slot retained                        true

delayed effects from the released owner
  late complete() threw                          no
  owner slot still released                      true
  bystander slot still intact                    true
  bystander.phase                                running
  bystander backend cancels                      0

authority chain intact: YES
```

The selected owner is cancelled (`["superseded"]`) and settled (`failed:run_stalled`, `staleExpiryReason=no_activity`) before its slot is released; the other session's backend is never cancelled and keeps its slot; and a late `complete()` from the released owner neither reacquires its slot nor touches the bystander. `drained` is `false` because no embedded handle was registered in this scenario — nothing was drained, and the helper reports that honestly rather than claiming it.

### Regression coverage

Both regressions live in the existing `diagnostic-stuck-session-recovery.integration.test.ts`, which drives the real registry and the real runner. They call the exported helper directly so they enter the changed code deterministically instead of depending on the recovery staleness window; every assertion is unconditional.

**Red** — production change reverted:

```
FAIL > cancels and settles the selected owner before releasing its slot for a cron timeout
AssertionError: expected "vi.fn()" to be called at least once

Tests  1 failed | 17 passed (18)
```

The named owner's backend is never cancelled on the cron path.

**Green** — `Tests 18 passed (18)`.

**Each part is load-bearing.** Reverting one at a time and rerunning the same suite:

| Reverted | Failing assertion |
| --- | --- |
| Handle ownership fence | `isEmbeddedAgentRunHandleActive(shared)` is `false` — foreign handle deregistered |
| Key-addressed reply resolution | the bystander's reply operation is force-cleared |
| Key-addressed stale expiry | the bystander's operation is expired and reaches `failed` |
| Non-stuck expiry before release | the selected owner's backend is never cancelled |
| Reply-owner fence on abort/drain | the bystander's backend is cancelled instead |

Adjacent suites, no regressions:

| Suite | Result |
| --- | --- |
| `reply-run-registry` (`.test`, `.control-owner`, `.retained-singleton`) | 115 passed |
| `cron/service/timer.timeout-watchdog` + `timer.regression` | 78 passed |
| `embedded-agent-runner/runs.liveness` + `subagents/registry/subagent-reset.lifecycle` | 22 passed |

`oxlint` and `oxfmt --check` are clean on both touched files. The full `cron` project stalls on this Windows machine identically with the change reverted, so it is not exercised here.

### Owner lineage across cleanup

Addresses the two retained ownership findings and the related security concern (`runs.ts:1468-1470`, `runs.ts:1541-1546`, `runs.ts:1542`). Both regressions fail with the production change reverted and pass with it:

```
× leaves a successor admitted under the same key with a new session id running
  AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
    expect(successorCancel).not.toHaveBeenCalled();

× does not abort a replacement that claims the session id while the owner is expiring
  AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
    expect(replacementAbort).not.toHaveBeenCalled();
```

```
node scripts/run-vitest.mjs run --config test/vitest/vitest.logging.config.ts \
  src/logging/diagnostic-stuck-session-recovery.integration.test.ts \
  src/logging/diagnostic-stuck-session-recovery.owner-lineage.integration.test.ts

Test Files  2 passed (2)
     Tests  20 passed (20)
```

The 18 pre-existing cases still pass in the red control, so the two new ones are specific to this change rather than incidental. They live in their own file because the existing suite is at the 1000-line `max-lines` cap.

Rebased onto `main` @ `6ac7eed664b`, no conflicts. `tsgo:core` and `tsgo:core:test` both exit 0. The full `logging` project reports 4 failures on this Windows machine — `diagnostic-support-bundle` (file permissions) and `logger-file-transport` (piped append) — identically with the change reverted, so they are pre-existing here.

### Authority chain on a real transport

The earlier trace used backends that completed themselves on cancel, which showed callback and slot behavior only. This one gives each owner a genuine in-flight HTTP stream: a local server on `127.0.0.1` holds the response open, sends a first chunk, then waits before its final chunk. Cancellation has to travel the real exported helper and the real reply-run registry to reach those sockets, and the **server** — not the harness — records what happened to each connection. Nothing test-only is imported; the state directory is an isolated temp dir. No credentials and no egress.

Three owners, one run:

- `solo-owner` — a reply operation **and** a registered embedded run handle, the full production shape.
- `owner` / `bystander` — two conversations sharing one session id; cleanup names `owner`.
- `successor` — a later execution holding the same session key under a new session id; cleanup names the id that already ended.

**On unmodified `main`, cleanup aimed at `owner` kills `bystander` instead:**

```
owner      request_open
owner      first_chunk_sent
bystander  request_open
bystander  first_chunk_sent
bystander  backend_terminated                            <- the wrong conversation
owner      final_chunk_sent
owner      request_completed                             <- the intended target survives
bystander  final_chunk_rejected (connection_terminated)
```

`owner` finishes and delivers its final output while the bystander's live stream is cut. That is the reported damage, on a real connection.

**With this PR:**

```
solo-owner backend_terminated
solo-owner final_chunk_rejected (connection_terminated)
owner      backend_terminated
owner      final_chunk_rejected (connection_terminated)
bystander  final_chunk_sent
bystander  request_completed
successor  final_chunk_sent
successor  request_completed
```

This covers the three effects that were called out as unproven:

| Effect | Evidence |
|---|---|
| Production backend termination | `backend_terminated` is emitted by the server when the socket closes mid-stream; the owner received only `["first"]`. `scenario0` reports `aborted: true, drained: true` with the handle deregistered and the slot released. |
| Rejection before delayed final I/O | The server leaves its final-chunk timer armed and records `final_chunk_rejected (connection_terminated)` when it fires against a terminated connection. |
| Replacement-owner continuity | `bystander` and `successor` receive `["first","final"]` and reach `request_completed`, still `running`, slots intact. |

**Same key, same session id, no embedded handle.** This is the shape the previous revision left open: expiry completes the captured owner, its slot is freed, and a fresh visible turn is admitted on *both* identifiers before the id-addressed abort runs. Neither the key comparison nor the handle comparisons fire, so the replacement was cancelled in the captured owner's place. The predicate now compares the resolved operation against the captured one by identity.

Added as a fourth owner in the same harness, with its own live HTTP stream.

Server-recorded, `main` plus the earlier commits (no identity comparison):

```
same-key-owner       request_open
same-key-owner       first_chunk_sent
same-key-owner       backend_terminated
same-key-owner       final_chunk_rejected (connection_terminated)
```

The replacement never appears: its connection was cancelled before the server could log it. Harness state: `status: terminated`, `phase: aborted`, `parts: []`.

With this revision:

```
same-key-owner       request_open
same-key-owner       first_chunk_sent
same-key-owner       backend_terminated
same-key-replacement request_open
same-key-replacement first_chunk_sent
same-key-owner       final_chunk_rejected (connection_terminated)
same-key-replacement final_chunk_sent
same-key-replacement request_completed
```

The replacement stays connected and completes with `["first","final"]`, still `running`, slot retained, while the captured owner is terminated and its delayed final write refused.

The replacement is admitted synchronously inside the owner's cancel, which is where production admission can take the freed slot. An earlier version of this scenario admitted it from the settled-stream callback instead; that lands after the cleanup path has already resolved the session id, so red and green were identical and the scenario proved nothing.

A matching regression covers the same shape without the transport: `does not cancel a replacement holding the same key and session id with no handle`, which fails with `expected "vi.fn()" to not be called at all, but actually been called 1 times` when the identity comparison is reverted.

**Selective reversion across the three states**, same harness:

| Scenario | `main` | + first two commits | + owner lineage | + this revision |
|---|---|---|---|---|
| `solo-owner` | terminated / **completed** | terminated / failed | terminated / failed | terminated / failed |
| `bystander` | **terminated / aborted** (owner completed, still running) | completed / running | completed / running | completed / running |
| `successor` | completed / running | **terminated / failed** | completed / running | completed / running |
| `same-key-replacement` | — | — | **terminated / aborted** (never reached the server) | completed / running |

The middle column is why the two ownership findings mattered: the first two commits fixed the bystander but introduced the successor regression, exactly as described in review — on `main` the id-based lookup simply never found the successor. The third column closes it without giving the bystander back. On `main` the solo owner is also merely `completed` rather than stale-expired, which is the uncancelled-release half.

Scope note: this exercises the real cleanup helper, the real registry and real sockets. It is not a full embedded agent run — the backends are HTTP streams standing in for provider work, chosen so termination is observable from outside the process.

### On the reply-operation collision in isolation

A session-id collision between two reply operations, with no embedded handle involved, leaves the bystander's operation intact on unmodified `main` — I could not build a deterministic reproduction of that half on its own and am not claiming one. The damage I can reproduce starts at the embedded handle and at the uncancelled release above; the reply-side changes are required because the handle fence alone leaves force-clear, expiry and abort pointed at the foreign owner, which the table shows.


MertBasar0:
Status, so nobody waits on me: the code change and its failing-first regression stand as described, but I cannot stage the authority-chain proof this needs. It requires two keys with a reused session id, a live backend, observable cancellation and delayed-effect rejection, and the non-stuck path — I have no way to produce redacted real-backend traces for that here.

The remaining P1 finding is also correct and unaddressed: operation-addressed expiry is gated on `reason === "stuck_recovery"`, so `cron_timeout` and no-reason callers still release the selected owner's slot through `forceClearReplyOperation()` without cancelling its abort signal or backend.

I have posted the reproducible evidence and both limits on openclaw/openclaw#145152 so the finding is not lost. Happy for a maintainer to take this over, re-use the regression, or close it in favour of a different fix — whichever is most useful. I am not holding the issue.

obviyus:
Thanks @MertBasar0. We tried to reproduce this on `main` (`c5695c7579dc`) through the real entry point: no supported Gateway action reassigns a finished execution id to another live conversation before delayed cleanup, so the cross-conversation cancellation could not be produced; the normal-timeout path behaves correctly. It behaves as expected there, so we are closing this for now. If you still see it on a current build, please reopen with the exact steps and we will pick it up.

PR #145430, opened 2026-09-12
perf(ui): keep sessions query code out of startup

obviyus:
Related: #145347, #145402, #145415, #145516, #145540.

## What Problem This Solves

Control UI route definitions pulled unnecessary Sessions code into the startup bundle. Earlier route changes had exceeded the startup budget; the current base already met it.

## Why This Change Was Made

One pathname owner classifies dynamic routes, and Sessions builds its list queries inside the lazy page. The duplicate router classifier is removed. Route recovery and the existing Usage loader retain their behavior.

## User Impact

The initial download is smaller. Sessions listing, search, archived deletion, deep links, and reconnect behavior remain unchanged. No configuration or migration is required.

## Evidence

The paired contribution measurement saved 582 compressed bytes. The final integrated startup bundle measured 352,276 bytes against the unchanged 355,116-byte limit. All 132 targeted router, Sessions, sidebar, and related tests passed on the unchanged contribution, along with both production-bundle reconnect cases. Final continuous integration passed all checks, including browser tests with a real Gateway.

PR #145448, opened 2026-09-12
refactor(ui): remove unused cloud target exports

obviyus:
## Problem

`pnpm deadcode:exports` fails on main after #145183 (`b9050029efe97266108667536e25b1a806e6dafd`). The environment-picker change replaced the connect-machine menu item with a device-header action but left its renderer exported. It also exported the cloud configuration renderer solely for direct tests.

## Fix

Delete the unused connect-machine renderer and its unused CSS selectors. Keep cloud configuration private and exercise it through the existing profile-menu renderer. Preserve the live device-header action, connect dialog, and move-session menu.

The production cleanup removes 25 net lines. No baseline, ignore list, test case, runtime behavior, or visible styling changes.

## Consumers

- `renderConnectMachineMenuItem` has no remaining import, call, barrel, or dynamic reference. #145183 removed its last caller from `ui/src/pages/new-session/where-chip.ts`; that file now renders the live `data-action="connect-machine"` button directly.
- `.new-session-page__connect-machine` appears only in the deleted renderer and the removed stylesheet selectors. Other selectors in those rules keep their declarations. `connect-machine-dialog` styles remain live and unchanged.
- `renderCloudConfiguration` remains private to `ui/src/pages/new-session/cloud-target.ts`, called by `renderCloudProfileMenuItems`. Its only external import was `cloud-target.test.ts`, now removed.
- `renderCloudProfileMenuItems` is consumed by `where-chip.ts` in compact mode and `ui/src/components/session-placement-move-dialog.ts` in noncompact mode. `target-controls.ts` connects the Where picker to the New Session page. These production callers are unchanged.
- `renderCloudMachineMenuItems` and `renderCloudOsMenuItems` retain their exports for the move-session dialog. `requestPlaceCatalog` and the connect-machine dialog owner are unchanged.
- Repository searches cover `src/`, `extensions/`, `ui/`, `apps/`, `docs/`, and `packages/`. No schema, plugin manifest, public option, or stored value uses the removed export or class. The remaining private function name is intentional; its implementation and internal caller are not retired.

## Contention

None. No lock, queue, transaction, shared writer, asynchronous work, or concurrency policy changes.

## Invalidation

None. No cache or derived-state lifecycle changes; policy reload, config reload, and Gateway restart behavior is unchanged.

## Tests

- `pnpm deadcode:exports`: baseline `6cb7280a4e85d655c51312fca4db1c9269f3b551` exits 1 with both reported production exports; candidate exits 0 with zero entries in script, production, and full-tree scans.
- Existing UI tests: **80 passed** across `cloud-target.test.ts`, `where-chip.test.ts`, and `session-placement-move-dialog.test.ts` using `vitest run --config ui/vitest.config.ts --maxWorkers=2`.
- Test entry point: the six rewritten cases now render `renderCloudProfileMenuItems`, the production renderer used by the registered New Session page's Where picker. The unchanged Where-picker and move-dialog tests cover its callers and sibling mode.
- Mutation proof: both original and rewritten cases fail for the same six assertions when fixed-machine rendering, unavailable-OS filtering, and submitting disablement are intentionally broken. The temporary changes were restored and are not in this PR.
- Native formatter and repository precommit hook pass. Independent source review found no actionable issue.
- Older-host/previous-state, real-config Gateway startup, and channel proof are not applicable: no SDK, updater, Doctor, persistence, config loading, login, catalog, or product behavior changes. Removed CSS has no live renderer; there is no appearance change to capture.

## Deletion ledger

- Delete `renderConnectMachineMenuItem` and its `.new-session-page__connect-machine` selectors: obsolete after #145183; no live behavior removed.
- Remove only the export of `renderCloudConfiguration`; retain its implementation and internal production caller.
- Preserve all four single-machine cases, the choices/disabled-state case, and the unavailable-selected-OS case through `renderCloudProfileMenuItems`. No test is deleted or skipped. The no-selected-OS assertion is scoped to configuration content because the surrounding profile is now rendered and selected.

## Structural dependencies

#145431 owns the separate config-cli root overflow and is merged. #145430 owns the separate UI startup-size repair. This PR does not change either limit or claim a waiver for either gate.

### DIFF
--- a/ui/src/pages/new-session/cloud-target.test.ts
+++ b/ui/src/pages/new-session/cloud-target.test.ts
@@ -2,11 +2,7 @@
 
 import { render } from "lit";
 import { describe, expect, it, vi } from "vitest";
-import {
-  renderSessionMenuItem,
-  renderCloudProfileMenuItems,
-  renderCloudConfiguration,
-} from "./cloud-target.ts";
+import { renderSessionMenuItem, renderCloudProfileMenuItems } from "./cloud-target.ts";
 
 describe("cloud target menu", () => {
   it("renders explicit remediation commands on separate lines", () => {
@@ -157,15 +153,13 @@ describe("cloud target menu", () => {
   ])("renders a single machine as fixed provider configuration", ({ machine, expected }) => {
     const container = document.createElement("div");
     render(
-      renderCloudConfiguration({
-        profile: { id: "aws", providerId: "aws" },
-        operatingSystems: [],
-        machines: [machine],
-        selectedOs: "",
+      renderCloudProfileMenuItems({
+        profiles: [{ id: "aws", providerId: "aws", machines: [machine] }],
+        selectedId: "aws",
+        compact: true,
         selectedMachine: machine.id,
         submitting: false,
-        onSelectOs: vi.fn(),
-        onSelectMachine: vi.fn(),
+        onSelect: vi.fn(),
       }),
       container,
     );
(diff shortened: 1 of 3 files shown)


obviyus:
Superseded by #145430, which combines the main-side structural repairs.

PR #145449, opened 2026-09-12
fix(macos): discard stale browser import presentation

steipete:
## What Problem This Solves

Fixes an issue where users switching Gateway modes or making a newer browser-import request could see an obsolete offer, availability result, failure, or success banner appear when an earlier request finished. A stale explicit request could also bring the dashboard forward again.

## Why This Change Was Made

BrowserProfileImportModel owns presentation intent separately from the latest status request. A pending explicit offer consumes the newest status facts, so an activation or another dashboard availability read cannot silently discard it. Automatic offers also yield to a pending explicit request, including when persisted state records a dismissal or prior import. Phase changes and newer explicit requests retire that priority; an older completion cannot clear its successor. Results publish only while the phase and local-mode context remain current. A superseded explicit request has its own outcome, and the dashboard caller leaves it alone. Refreshes preserve an active import; an already dispatched Gateway import still finishes even if its banner context disappears.

## User Impact

Browser import follows the current local Gateway context and latest user intent. Mode changes withdraw stale presentation without pretending to roll back an import already sent to the Gateway. No UI appearance, import endpoint, or persistence contract changes.

## Evidence

- Added continuation-controlled regression coverage in BrowserProfileImportModelTests for mode changes and round trips, concurrent forced refreshes, both availability readers, import success/failure after context changes, active-import refresh, eligibility before dispatch, and automatic attempts during a pending explicit re-offer for both dismissed and imported persisted states.
- Removed the production-only test phase setter; existing tests now use the actual refresh entry point.
- Swift app/test build, lint/format, and git diff --check passed. Required [CI run34671752848](https://github.com/openclaw/openclaw/actions/runs/34671752848) passed on final head `4c6a38080452b11f1cd42fa05b4e98b04036b850`, including native Swift tests. The earlier model suite also passed in a disposable VM; no operator app, Gateway, or local full native suite was used.
- Deslop completed. Independent Codex P2 review is scoped-clean after fixing cross-reader availability ordering preserving explicit intent during background reads, and preventing automatic offers from superseding pending explicit intent. The final priority repair also passed fresh independent P2 review. Lead verified complete caller cutover and exact patch identity.

Native localization verification passes. The generated inventory records only the unchanged long message becoming a multiline Swift literal to satisfy lint; its text and localization ID are unchanged.

### DIFF
--- a/apps/.i18n/native-source.json
+++ b/apps/.i18n/native-source.json
@@ -2741,7 +2741,7 @@
     {"id":"native.apple.2282d2bdeca50246","source":"Next Match","surface":"apple","sites":[{"kind":"ui-call","path":"apps/shared/OpenClawKit/Sources/OpenClawChatUI/ChatTranscriptSearch.swift"}]},
     {"id":"native.apple.de0f4838e9554464","source":"Next match (⌘G)","surface":"apple","sites":[{"kind":"ui-modifier","path":"apps/shared/OpenClawKit/Sources/OpenClawChatUI/ChatTranscriptSearch.swift"}]},
     {"id":"native.apple.f5f840287aa988ff","source":"Next up","surface":"apple","sites":[{"kind":"ui-localized-call","path":"apps/ios/WatchApp/Sources/WatchInboxView.swift"}]},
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4c6a38080452b11f1cd42fa05b4e98b04036b850](https://github.com/openclaw/openclaw/pull/145449/commits/4c6a38080452b11f1cd42fa05b4e98b04036b850)
- Landed commit: [b88552a9f899f076c1acf973c50b0a98b7b4667c](https://github.com/openclaw/openclaw/commit/b88552a9f899f076c1acf973c50b0a98b7b4667c)

PR #145172, opened 2026-09-11
feat(sessions): automatically archive older transcripts

steipete:
Related: #144232

<details>
<summary>Additional instructions</summary>

The branch is in `openclaw/openclaw`, so maintainers can edit it directly. GitHub’s fork-collaboration toggle applies only to cross-repository PRs.

</details>

## What Problem This Solves

Older session transcripts keep agent databases large even when their sessions are rarely opened. Operators need automatic extraction, a visible account of the resulting storage, and a safe way to tune the policy without restarting the Gateway.

## Why This Change Was Made

Adds opt-in cold storage owned by the existing Gateway maintenance lifecycle. At startup and once a minute, bounded workers move eligible inactive transcript windows into verified, immutable compressed JSONL files. SQLite retains their identity and archive metadata; opening history restores the exact serialized events and associated projections before use. Current and historical windows are eligible, with active work, recovery, pending inputs, and required history references protected.

The agent schema advances from 19 to 20 because older readers cannot safely interpret extracted event rows. The supported Doctor migration creates the new representation without rewriting existing transcripts; state schema remains 17. Supported full, SQLite, and Git backups embed verified compressed payloads in their snapshot databases, so restoration does not depend on the original archive directory.

The maintainer accepted this opt-in storage design and its operating boundaries: agent schema20, authoritative archive files, self-contained supported backups, pre-upgrade-backup rollback for older binaries, and the documented large-batch writer-queue latency. Database-only replicas must carry the archive files. The feature remains disabled by default.

## User Impact

Settings → Agent Defaults → Session now shows hot/archived transcript counts, database and WAL sizes, archive file sizes, embedded archive bytes, per-agent details, and background maintenance status. Operators can enable archival, choose a positive day threshold, and start a background batch. Settings use the canonical config apply flow and take effect without a Gateway restart.

```json5
{
  session: {
    maintenance: {
      coldStorage: { enabled: true, afterDays: 30 },
    },
  },
}
```

Cold storage defaults off and has its own opt-in policy, independent of retention `mode`. Each pass is bounded to 128 transcripts and 64 MiB of restoration envelopes. Recently restored history stays hot for 24 hours in the current process. Text search reports excluded cold transcripts; bulk previews report archived history without restoring it. Chat/history access restores on demand, including when automatic extraction is disabled.

[...]

PR #145462, opened 2026-09-12
fix: keep cold sessions visible and verify release recovery

steipete:
## What Problem This Solves

With transcript cold storage enabled, an uncached archived session could fail the Activity list and startup prewarm when they requested derived titles or last-message previews. Cached titles masked the failure until restart. Full release verification also lacked permanent installed-package coverage of archival, recovery, and live model recall from cold history.

Related: #145172.

## Why This Change Was Made

Handle cold transcripts in the shared optional title reader. Keep session metadata visible without restoring payloads, and use the existing unavailable-title cache fence so omitted previews refresh after history restoration. Actual history requests and missing-file errors retain their existing behavior.

Extend the existing `session-runtime-context` Docker gate with three dated synthetic histories imported through Doctor. Exercise automatic 30-day archival, a live change to seven days, missing-file errors without losing recovery references, exact event restoration, disabled-policy restart, Activity listing without rehydration, and portable SQLite backup recovery after removing the original state.

Extend the existing `openai-chat-tools` gate with one live OpenAI recall turn whose answer exists only in archived history. Verify the response, restored original event bytes, provider/model, and token usage while preserving the existing HTTP tool-call case. Both lanes already run in full release validation; document their coverage. The shared frozen-target preflight omits only unsupported cold subcases for explicitly authorized older packages, while preserving their original checks and requiring cold coverage for current targets.

## User Impact

Cold sessions remain available in Activity without refilling the database merely to render the list. Cached title fields remain usable; otherwise session metadata supplies the title and last-message previews return after history is restored. No schema bump or configuration change is required.

Release verification gains bounded automatic archival and recovery checks plus one additional real-provider call. Large database stress remains a separate campaign.

## Evidence

- Real-worker regression failed with `SessionTranscriptColdError` before the fix; 72 title-reader, session-list/prewarm, and list-cache tests pass after it.
- Release planner: 110 tests passed; executable plans include both functional-image lanes. Sixteen focused frozen-target compatibility cases pass, including supported/unsupported, strict/authorized, and corrupt-source cases.
- Lint-suppression inventory: all four tests pass without a new allowlist entry.
- Integration validation: 19 targeted admission/cloud tests, full root-test/scripts typechecking, and both unused-file scans passed. Overlapping UI repairs are now on main and are absent from the final PR delta.
- Independent reviews are clean. Formatting and syntax checks pass.
- Earlier packaged runs verified scheduled 30/7-day archival and exposed the uncached-title defect. Fixture corrections preserve legacy file modification times and capture CLI JSON failures on stdout.
- Packaged proof on Testbox at `001b7c51c68ae84eb6a696c41facbb6903a9b625`: deterministic lifecycle **passed in 95 seconds**; live OpenAI HTTP tools plus archived-history recall **passed in 17.9 seconds**. The recall checks its nonce, unchanged original event prefix, exact provider/model, and positive token usage. [Testbox environment](https://github.com/openclaw/openclaw/actions/runs/34668435609).
- Tested tarball SHA-256: `f568a0c0c7b0cab82b8da025701f20528d6c54663152fd34f016bc3af15960b8`. Subsequent change is a test-only locator repair; runtime and cold-storage harness bytes are unchanged.
- The broad remote changed check exposed an inherited undefined screenshot locator. Its repair passes the actual UI E2E typecheck and capture-enabled browser case; full current-head CI remains the landing gate.
- Core-test typechecking passes after preserving the fixture helper's complete inferred write scope.


steipete:
Merged via squash.

- Prepared head SHA: [3be509434e7d3b2e50c5e36b250c0402b541fe3a](https://github.com/openclaw/openclaw/pull/145462/commits/3be509434e7d3b2e50c5e36b250c0402b541fe3a)
- Landed commit: [31848a15205a1aab75165e91c994af851962e0e8](https://github.com/openclaw/openclaw/commit/31848a15205a1aab75165e91c994af851962e0e8)

PR #145415, opened 2026-09-11
fix: show package icons instead of generated plugin artwork

Patrick-Erichsen:
<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Plugin browsing showed generated artwork instead of package branding, including an unrelated knot for Slack. Published plugin archives also omitted existing icon assets, leaving ClawHub releases without an icon to serve.

## Why This Change Was Made

The shared package builder includes each package’s existing `assets/icon.png` in npm and ClawHub publication overlays. The Control UI uses installed or catalog-provided package icons and its generic placeholder for missing or broken images. It removes the generated artwork lookup and all 170 bundled artwork files across plugin, channel, chat, and consent surfaces.

The macOS ClawHub publisher also adopts its npm sibling’s safe optional-array expansion: Bash 3.2 previously stopped before packing when metadata arguments were absent. Real CLI failures still return nonzero.

CI exposed stale assertions and unused rendering code from the recently landed environment-picker redesign. This repair deletes the unused renderer and CSS, keeps cloud configuration private to its owning renderer, uses the existing cursor token, and updates tests to exercise the current controls. Hover tests wait for Web Awesome’s public animation-completion event; denied folder navigation waits for the initial listing. No timeout, budget, baseline, compiler partition, user configuration, or storage changes are included.

## User Impact

Plugin branding comes from the plugin package. Missing and broken images show a consistent placeholder, and newly published archives carry their icon. Existing ClawHub releases need the already-deployed maintenance repair to recover icons from their recorded source commits; that production operation is separate from this merge.

## Evidence

Reported Slack card before:

Reported Slack card with incorrect generated knot

Changed Control UI against a mock Gateway, inspected after image decoding. Installed and catalog Slack icons render the package logo; missing and broken images show placeholders:

Package icons and generic placeholders in the changed Control UI

- Actual archive comparison: Slack omitted its PNG before the fix; fixed Slack, Microsoft Teams, and Amazon Bedrock archives contain byte-identical source icons. All 91 publishable package plans include their existing icon. Six pack/extract cases verify exact bytes and exclusion of unrelated assets.
- 141 focused UI tests pass, including missing/broken images and rerenders. UI production build and desktop/mobile browser checks pass, with no generated artwork assets or requests. All 11 owning pack/publication wrapper tests pass, including the previously failing macOS fixture; deliberate CLI exit 7 remains 7.
[...]

PR #145471, opened 2026-09-12
fix(gateway): resume suspension leases within their expiry budget

steipete:
## What Problem This Solves

Fixes an issue where hosting controllers could see the Gateway remain suspended past the advertised lease expiry when work inspection takes time. A draining status poll or renewal could also return readiness after its inspection consumed the remaining lease budget.

## Why This Change Was Made

The existing synchronous implementation calculated `expiresAtMs` before inspection but started a full two-minute timer afterward. Use the remaining budget for automatic resume, retain an elapsed-time deadline across wall-clock rollback, and check expiry after initial and draining inspections. Expired work follows the existing scheduler-resume and recovery paths.

Public prepare/status return shapes and synchronous inspectors remain unchanged. The two-minute policy, renewal behavior, and configuration surface are unchanged.

## User Impact

The advertised expiry and automatic resume agree even after slow inspection. Controllers receive an expired/changed preparation error or a running/recovering status instead of an already-expired lease.

## Evidence

- The initial regression failed in all four slow-inspection cases on unmodified main.
- The polling/renewal follow-up failed in all four cases before its repair.
- Full coordinator suite: 54 tests passed, including real Node timers with a synthetic short remaining budget, clock rollback, renewal, scheduler recovery, and cleanup.
- Final lease-setup regression: both cases failed before the arming-boundary guard.
- Isolated P2 review completed with no remaining actionable findings.
- Initial CI and local checks found unrelated main UI lint/dead-export failures. This PR is now rebased onto main containing #145415, which repairs those prerequisites; the unchanged expiry patch passed all 54 tests again on the refreshed head.
- Documentation checks passed: 13,472 internal links (0 broken) and 1,229 configuration examples validated.
- Full changed-file gate passed on refreshed commit `4577a2232276`, including all dead-export scans, core/test typechecks, scoped lint, and runtime guards.

### DIFF
--- a/docs/gateway/external-apps.md
+++ b/docs/gateway/external-apps.md
@@ -243,7 +243,10 @@ With `drain: true`, the same suspension owner instead keeps admission closed
 and cron scheduling paused until existing work settles. Already-owned cron
 completion and reconciliation continue.
 
-Both draining and ready leases last two minutes. Repeat `prepare` before
+Both draining and ready leases share a two-minute budget starting before work
+inspection. Preparation that exhausts that budget resumes scheduling and fails
+instead of returning an expired lease. Clock rollback does not extend the budget.
+Repeat `prepare` before
 `expiresAtMs` with the same `requestId`, terminal policy, and drain mode to renew
 the same `suspensionId` unless a restart handoff is armed; changing any of those values conflicts with the
 existing lease. Use `status` for routine polling and reserve `prepare` for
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4577a2232276da25344155cf2802911789bcb1a9](https://github.com/openclaw/openclaw/pull/145471/commits/4577a2232276da25344155cf2802911789bcb1a9)
- Landed commit: [5e2fcfd6641e69e6e5deee37add4c8173260c0c2](https://github.com/openclaw/openclaw/commit/5e2fcfd6641e69e6e5deee37add4c8173260c0c2)

PR #145478, opened 2026-09-12
fix: keep newer model choices when a page refresh finishes

steipete:
## What Problem This Solves

Fixes an issue where users opening a model picker during a refresh in **Settings → Models** could lose newly discovered choices, see discovery stop, or see the **Auto** utility model revert when the older page refresh finished.

## Why This Change Was Made

The page respects the newer picker request when it applies a refresh. Model choices, the resolved Auto utility model, progress, and catalog errors stay with that request while the refresh updates provider and configuration data. A later page refresh continues to retire older picker work.

## User Impact

Model pickers remain usable during refresh. New choices and the Auto model stay consistent, pending discovery can finish, and refreshed default-model selections still appear.

## Evidence

The completed and pending ordering regressions distinguish the initial, older refresh, and newer discovery results. Both fail without preserving the automatic utility value and pass with the fix. All **166 tests across seven related suites** pass, including publication, inverse ordering, current Auto labels, and other default-model controls. UI types, all 17 core-test type graphs, lint, formatting, and the final canonical changed-file checks pass.

Real Chromium proof uses a synthetic Gateway and the actual Refresh button and picker on current main. The refreshed primary selection appears in both versions. The repaired version preserves the correct Auto value and newer catalog or pending progress, then adopts the pending discovery when it completes. Both runs finish without page errors or configuration writes; final source and index bytes are restored after baseline capture.

| Newer discovery when the older refresh finishes | Before | After |
| --- | --- | --- |
| Already completed | Auto reverts to the stale fallback | Auto keeps the newer discovered model |
| Still pending | Auto reverts and discovery progress disappears | Auto and discovery progress remain current |

The two baseline states produce byte-identical screenshots, so they share one image above. Independent P2 review found no actionable findings. Browser proof uses a synthetic Gateway; no live provider or operator Gateway is exercised.

### DIFF
--- a/ui/src/pages/model-providers/catalog-discovery.ts
+++ b/ui/src/pages/model-providers/catalog-discovery.ts
@@ -19,6 +19,8 @@ type DiscoveryGateway = {
 };
 
 export type CatalogDiscoveryController = {
+  /** Latest picker request, including one that has already settled. */
+  readonly generation: number;
   /** Whether a discovery request is currently in flight. */
   readonly discovering: boolean;
   /** A user-facing retry hint when discovery failed; null while clean. */
@@ -46,8 +48,12 @@ export function createCatalogDiscoveryController(
   let pending: AbortController | null = null;
   let error: string | null = null;
   let requestedDiscovery = false;
+  let generation = 0;
 
   const controller: CatalogDiscoveryController = {
+    get generation() {
+      return generation;
+    },
     get discovering() {
       return pending !== null;
     },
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2f341b363b8198725954b477f6e6765bbaa35abe](https://github.com/openclaw/openclaw/pull/145478/commits/2f341b363b8198725954b477f6e6765bbaa35abe)
- Landed commit: [7275b4baa0251b3499c2ba2b476b9ecc7bc0c3af](https://github.com/openclaw/openclaw/commit/7275b4baa0251b3499c2ba2b476b9ecc7bc0c3af)

PR #145420, opened 2026-09-11
refactor: simplify progress and picker rendering ownership

steipete:
Related: #145345, #140037, #145183

## What Problem This Solves

Maintainer-requested cleanup after #145345. The compositor already filters quiet tool rows, but internal layout helpers still carried another `toolProgress` switch. Slack's Block Kit renderer constructed native task identities and status objects only to discard everything except their titles.

Validation also exposed a real picker focus race: WebAwesome initially focused search, the user opened cloud configuration, then OpenClaw focused search again after the outer animation and dismissed the cloud card.

## Why This Change Was Made

Keep visibility with the compositor and let internal streaming helpers arrange already-admitted rows. Preserve shipped SDK merge/format helpers and legacy behavior. Share Slack attention-title formatting while constructing stable approval task identities only for native streams; Block Kit renders titles directly. The native attention loop remains approval-only; tool failures keep their work-task and reconciliation path.

Remove the duplicate picker after-show focus handler. WebAwesome already owns the input's native `autofocus` before the opening animation. Retain that single focus owner so animation completion cannot overwrite a later user interaction.

Remove one duplicate native-chunk test left after separate start/update builders were unified. Replace the handler-only focus unit test with real browser coverage of initial search focus, retained clicked-profile focus, and machine selection across the actual opening animation. Hover coverage remains alongside the new click variant.

## User Impact

Quiet progress stays quiet. Approvals retain their identities and lifecycle, and explicitly enabled diagnostics retain failure/recovery text. Block Kit avoids unused native ID hashing and task-object construction.

Fast cloud selection keeps its configuration card open when the outer picker finishes animating. Search still receives initial focus through WebAwesome's native behavior.

No public SDK, configuration, schema, dependency, transport, baseline, ignore-list, or CI-limit change. The final seven-file diff removes 22 production lines and adds 12 net test lines.

## Evidence

- 159 progress baseline tests passed before editing; 441 focused tests passed after cleanup across public streaming helpers, compositor quiet/plan/visibility, Slack native and Block Kit renderers, full Slack dispatch, and Telegram rendering.
- Full changed-file checks passed for the progress cleanup, including production/test types, lint, import cycles, and repository guards. Independent P0–P2 review was clean.
[...]

PR #145490, opened 2026-09-12
fix(control-ui): repair picker lint and action cursor

steipete:
## What Problem This Solves

Unrelated PRs inherit two New Session CI failures: the Where chip test file exceeds its existing 1,000-line lint limit, and the environment-details button uses an unconditional pointer cursor instead of the shared action token.

## Why This Change Was Made

Reuse the existing picker fixture in the empty-device/Auto-off test, retaining every test and assertion. The file falls from 1,020 to 991 counted lines without a suppression or limit change. Use `var(--cursor-action)` for the existing details button, matching adjacent in-page actions.

## User Impact

The environment-details action follows the shared cursor policy. Its rendering, accessible-label attribute, button type, and dimensions remain unchanged. Navigation links retain their pointer cursor.

## Evidence

- All 65 tests in the complete Where chip file passed with no filtering or skips; the hosted changed-test job also passed those tests before reporting the separate cursor-policy failure.
- The complete, unchanged cursor-policy test and original stylesheet lint passed on the combined candidate. Formatting and whitespace checks passed.
- A separate Chromium component proof used the actual menu-item renderer, actual base token and exact before/after CSS under positively asserted coarse-pointer emulation. The details cursor changed from pointer to default; the sibling action remained default and the navigation control remained pointer. Geometry stayed 32 by 32, with no page or cleanup errors.
- Independent Codex review of both changes is scoped-clean through P2.
- The original type-aware lint attempt exceeded its unchanged 6 GiB process-tree RSS limit and produced no lint verdict. That failure remains recorded; the later policy/style checks do not turn it into a pass. Required CI on the final head remains necessary before landing.
- The [isolated cursor validation](https://github.com/openclaw/openclaw/actions/runs/34666434288) completed; all 13 native children and monitors closed, the worktree was removed normally, and the dedicated Testbox is confirmed completed.

## Before and After

These inspected captures contain only synthetic component data. The annotations show actual computed CSS values; they do not capture an OS pointer. The fixture displays an untranslated label key, so this is not localization, full-app or native-device proof.

Before: environment details uses pointer

After: environment details uses the shared default action cursor

This is a supporting CI repair, with no runtime performance claim.

### DIFF
--- a/ui/src/pages/new-session/where-chip.test.ts
+++ b/ui/src/pages/new-session/where-chip.test.ts
@@ -850,39 +850,10 @@ describe("Where chip", () => {
   });
 
   it("omits automatic placement when no devices are paired and Auto is off", () => {
-    const state = resolveWhereChip({
+    const emptyContainer = renderPicker(false, undefined, {
       environments: [],
       cloudProfiles: [],
-      cloudProfileId: "",
-      deviceId: "",
     });
-    const emptyContainer = document.createElement("div");
-    render(
-      renderWhereChip({
-        state,
-        gatewayName: "",
-        environmentQuery: "",
-        onEnvironmentQueryInput: vi.fn(),
-        cloudProfileId: "",
-        deviceId: "",
-        worktreeAvailable: true,
-        submitting: false,
-        pendingPlacement: false,
-        popoverOpen: true,
-        popoverHiding: false,
-        isAdmin: false,
-        onGuardTransition: vi.fn(),
-        onPopoverShow: vi.fn(),
-        onPopoverHide: vi.fn(),
-        onPopoverAfterHide: vi.fn(),
-        onSelectDevice: vi.fn(),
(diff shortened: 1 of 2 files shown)


steipete:
Both reported CI violations are now repaired on main by other merged PRs:

- #145420 (`6f883894311d0b634ab6e3de75b56da838812f72`) reuses the existing picker fixture, including this exact empty-device test, and removes more duplicated setup.
- #145516 (`18ed3f173515c8efdcf3b9910cf14d2d953bc2e4`) removes the details-button pointer override, letting it inherit the existing action cursor rule from base.css.

I verified both commits are ancestors of current main and inspected the resulting test and cursor selectors. This PR's tests, browser proof and exact-head CI passed, but its changes now overlap those completed fixes. Native merge admission stopped before dispatch; no merge request or outcome was recorded. Closing this supporting repair as superseded, while preserving its validation evidence. No performance improvement is credited to this PR.

PR #145520, opened 2026-09-12
fix(codex): retain completions until task persistence succeeds

steipete:
## What Problem This Solves

Resolves a problem where a completed Codex native subagent could remain running in OpenClaw's task list or lose its completion after a temporary SQLite write failure. The monitor retired its completion bookkeeping before checking whether finalization actually persisted, including when the task runtime returned an empty result.

## Why This Change Was Made

Keep the original completion and timestamp with the existing monitor until finalization and delivery status are persisted. Retry the unfinished phase without repeating a durable announcement. Normal child close releases its subscription while retaining pending persistence; parent retirement and removed or already-terminal task records still prevent obsolete work from being recreated.

## User Impact

Completed native subagent tasks recover from temporary storage failures without losing their result or sending it twice. The synchronous harness API, native protocol semantics, SQLite schema, and existing announcement retry limit remain unchanged.

## Evidence

133 focused monitor, mirror, and shared-client lifetime cases pass. The shared-client fixture now records successful task writes instead of returning an empty result; its final task state is asserted. Regression coverage exercises empty and thrown writes, duplicate completion notifications, native child/client close, and removed, cancelled, or retired owners. The original ordering and the child-close failure were verified with failing regression tests before their respective repairs.

Standalone real-timer proofs use the actual scoped task runtime and isolated SQLite databases. A temporary trigger rejects terminal upserts three times; after either client close or native child close, the next write succeeds, one synthetic delivery occurs, and a fresh SQLite connection confirms the original summary/timestamp with succeeded/delivered state. No provider credentials or native-agent sends are used. The provider live-test recorder was updated to model applied task records; that provider test was not run.

Independent P2 review is clean after addressing child-close and matching-terminal recovery findings. Recovery pins the original task UUID within the scoped native run, checks that identity before retrying persistence or announcement delivery, and requires an acknowledgement for that same record. Tests reject same-run replacements and partial acknowledgements from sibling metadata rows. Actual SQLite tests seed both succeeded/pending and failed/pending records, reject three upserts, then verify one delivery and the original UUID, timestamp, and summary after reopening. The changed-scope checks include production and test typechecks, formatting, and routed boundary guards.

After a patch-identical rebase onto current main containing the independent UI fix in #145415, the focused tests, the actual SQLite native-child-close/reopen proof, and the complete local changed-scope gate passed, including all Knip scans.

The branch also contains the independently landed UI test correction from #145540; no UI changes are part of this PR.

### DIFF
--- a/extensions/codex/src/app-server/native-subagent-monitor.live.test.ts
+++ b/extensions/codex/src/app-server/native-subagent-monitor.live.test.ts
@@ -4,6 +4,9 @@ import fs from "node:fs/promises";
 import path from "node:path";
 import { setTimeout as delay } from "node:timers/promises";
 import type {
+  AgentHarnessScopedCreateRunningTaskRunParams,
+  AgentHarnessScopedFinalizeTaskRunParams,
+  AgentHarnessScopedSetDeliveryStatusParams,
   AgentHarnessTaskRecord,
   AgentHarnessTaskRuntimeScope,
 } from "openclaw/plugin-sdk/agent-harness-task-runtime";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8d0a9f4ff5d1d5186dc7dcf2c8e0b7c08c2c5f91](https://github.com/openclaw/openclaw/pull/145520/commits/8d0a9f4ff5d1d5186dc7dcf2c8e0b7c08c2c5f91)
- Landed commit: [6fef144932fcff5da0690d65116ff5408fcc6c60](https://github.com/openclaw/openclaw/commit/6fef144932fcff5da0690d65116ff5408fcc6c60)

PR #145525, opened 2026-09-12
fix(ui): clarify model authentication and prevent overlapping controls

steipete:
## What Problem This Solves

Fixes an issue where users opening Models at constrained widths could see Connect provider overlap the agent selector. Model discovery appeared as unstyled text above the settings rows, while model selections omitted credential details and Utility Model → Auto hid its resolved model.

## Why This Change Was Made

Settings headers wrap intact controls. Connect provider remains the credential-only action; the separate verification and activation flow is named Model setup and has secondary styling. Discovery uses the existing accessible spinner in a padded footer, so starting discovery leaves the model controls in place. Model rows stay side by side at 768px and stack at 640px and below.

The Models pickers reuse the chat picker's auth-label owner, displaying available API/subscription methods and an email when the account is unambiguous or explicitly pinned. Expired profiles are excluded from available credentials; unavailable explicit pins show Sign-in needed. Auto uses an optional read-only models.list projection from the existing utility resolver, based on global defaults and preserving an explicit profile pin. Agent overrides and runtime routing are unchanged.

## User Impact

Model settings remain readable across desktop, tablet, and phone widths. Users can see the available authentication method and the small model Auto selects, without guessing an account from profile order.

## Evidence

- Focused Models, shared picker, chat, model-auth, protocol, Gateway handler, and utility-resolver tests passed. Coverage includes mixed and expired credentials, explicit profile pins, global versus agent overrides, and refreshed Auto resolution.
- The new layout regression fails against the old header; the 768px row regression fails before removing forced stacking. The auth-display regression fails against the old view.
- Built Control UI browser checks pass at 1440, 1100, 768, 640, and 390px, including non-overlap and unchanged control positions during discovery.
- UI/core production type checks, affected UI/Gateway test type checks, targeted lint, stylelint, i18n source verification, and git diff checks passed.

The first CI run exposed outdated catalog-request expectations in three existing E2E suites. The follow-up commit adds the required preview flag without weakening the agent/view/refresh assertions; all 15 targeted mock-Gateway and layout cases pass locally. The focused real-Gateway acquisition/Retry case also passes locally. The subsequent hosted run passed all runtime suites but hit an inherited type error in a new-session screenshot test; this PR is now rebased onto the merged repair in #145540, with both reviewed commits unchanged.

Screenshots use synthetic accounts. The header-before capture isolates the original header layout on the updated model-details fixture.

| Header before | Updated Models and discovery |
| --- | --- |
| Original header overlap | Updated Models settings |

| 768px before | 768px after |
| --- | --- |
| Tablet rows unnecessarily stacked | Tablet rows side by side |


steipete:
@clawsweeper re-review

The follow-up commit updates exact Models catalog-request expectations in three existing E2E suites to include the utility-preview flag. Production code is unchanged. The 15 targeted mock-Gateway/layout cases and the focused real-Gateway acquisition/Retry flow pass locally; CI is running for the new head.

steipete:
@clawsweeper re-review

The branch is rebased onto main's merged CI repair #145540. `git range-diff` confirms both reviewed PR commits are unchanged. The UI E2E type check now passes locally on the refreshed base; this refresh is solely to validate and land the same Models changes against fixed main.

steipete:
Merged via squash.

- Prepared head SHA: [b5f84139d73b1470eb9a9e37c11aa7f5e121a5fe](https://github.com/openclaw/openclaw/pull/145525/commits/b5f84139d73b1470eb9a9e37c11aa7f5e121a5fe)
- Landed commit: [4fdcfc07b295ec41d2bd6261f5da5df89c1111fd](https://github.com/openclaw/openclaw/commit/4fdcfc07b295ec41d2bd6261f5da5df89c1111fd)

PR #145518, opened 2026-09-12
fix(ui): clarify presence platform and connection labels

steipete:
## What Problem This Solves

Fixes an issue where users opening a person's Online activity card saw raw platform strings such as `MacIntel`, `MacARM64`, and `win32`, repeated device/OS names, and the verbose `Control UI` connection label.

## Why This Change Was Made

The card now uses the shared platform formatter, displays browser connections as Web, and removes redundant device/platform labels while retaining OS versions. Explicit architecture values are formatted as ARM, Intel for an explicit Intel Mac target, or x64/x86. Browser `MacIntel` is not treated as CPU evidence: Apple silicon Macs and desktop-mode iPads report it too.

## User Impact

Connection descriptions are shorter and easier to read: Mac · Web, Mac · ARM · Web, Windows · Web, and iPad · Web. Devices and New Session also benefit from consistent platform and architecture formatting.

## Evidence

- 31 focused tests passed across platform formatting, the actual sidebar activity card, and Devices rendering.
- The sidebar regression test fails against the original formatter with the expected raw and duplicate platform labels.
- Inspected Chrome rendering with a synthetic mock Gateway covering Mac, ARM, Intel Mac, Linux, Windows, and iPad connections.
- Targeted lint, formatting, i18n verification, docs MDX validation, and `git diff --check` passed.
- Local typechecking is blocked by the worktree's borrowed dependency layout; hosted CI supplies the full check before merge.

Before (cropped to labels only):

Old platform and connection labels

After (synthetic connections):

Readable platform and connection labels

PR #145559, opened 2026-09-12
fix(ui): distinguish terminal clients in presence cards

steipete:
Related: #145518

## What Problem This Solves

Fixes an issue where users opening Online activity cards saw terminal UI connections labeled App, indistinguishable from native apps. Both use the same coarse UI connection mode, and presence discarded the client type.

## Why This Change Was Made

Presence now carries the accepted connection's optional client ID through WebSocket and HTTP Watch connection producers, cache refresh, the protocol schema, and generated Swift decoding. The card uses that ID to show Terminal for TUI connections; display names remain independent. Refresh also restores the connection mode so recreated native-app rows retain their label. Admission, permissions, and persistence are unchanged.

## User Impact

People connected through both a terminal and the native app now have distinct Terminal and App descriptions, including after presence-cache eviction. Web, Command line, and missing-client-ID records retain their existing labels.

## Evidence

- 52 focused tests passed across actual Gateway health/presence RPCs, heartbeat/cache-eviction recovery, the snapshot schema, and registered sidebar rendering.
- The original Gateway fails the new handshake and cache-refresh regressions because clientId is absent.
- Tests use custom and misleading display names to verify client type is independent of host text.
- Targeted lint, formatting, i18n, docs, protocol event coverage, protocol-since, and generated Swift checks passed.
- Independent autoreview is scoped-clean through P2.
- Inspected Chrome comparison uses synthetic mock-Gateway payloads without client IDs (before) and with client IDs (after), exercising the real card renderer.
- Local full typechecking rejects the borrowed dependency layout; required hosted CI will complete validation before merge.

The existing card projection test moved to a registered sibling to preserve all assertions while respecting the original file's line limit.

Before:

Presence without client type metadata

After:

Terminal and native App distinguished


steipete:
Merged via squash.

- Prepared head SHA: [d45ac10019ff8c09f4614328fbccb4f00181d175](https://github.com/openclaw/openclaw/pull/145559/commits/d45ac10019ff8c09f4614328fbccb4f00181d175)
- Landed commit: [0141569e7986c1f27a454f178a83680a3763a5ac](https://github.com/openclaw/openclaw/commit/0141569e7986c1f27a454f178a83680a3763a5ac)

PR #145540, opened 2026-09-12
test(ui): fix retained machine capture reference

steipete:
Refs #145463 #145436

## What Problem This Solves

The merged New Session cloud-dispatch proof fails with `ReferenceError: retainedMachine is not defined` when `OPENCLAW_CAPTURE_UI_PROOF=1` is enabled. The two independently passing PRs composed a new screenshot reference with an earlier removal of that local variable.

## Why This Change Was Made

Use the same scoped Fast-machine locator that the test already uses immediately above the capture. This is a one-line repair; interaction, selection and dispatch assertions remain intact.

## User Impact

Restores the synthetic capture flow without changing production UI or timing limits.

## Evidence

Reproduced the failure on merged main `da762af0e80c9fe6a61422dc5591ab24e5c9a3ce`. The repaired cloud-dispatch case passes with capture enabled in 22.16 seconds, and its synthetic screenshot was inspected. The other three capture cases passed during that merged-main replay.

Targeted type-aware lint, the UI E2E typecheck, `git diff --check`, and fresh P2 autoreview passed. No retries, timeouts, assertions, snapshots, or production behavior changed.

PR #145564, opened 2026-09-12
fix(process): preserve descendant output under Bun stalls

steipete:
## What Problem This Solves

Fixes an issue where commands running under Bun could lose buffered stdout and stderr from inherited descendant pipes when the event loop was stalled as the direct child exited. It also lets the process and terminal test lanes exercise their intended Node-owned branches when Vitest itself is hosted by Bun.

## Why This Change Was Made

Output cleanup now defers stream destruction to the next timers phase, giving both Node and Bun a poll turn while preserving the existing idle and hard drain bounds. Runtime-specific tests explicitly select a real Node executable or temporarily select the direct PTY branch, matching the ownership boundary each test validates.

## User Impact

Bun-hosted OpenClaw commands retain buffered descendant output across a contended event loop, and the complete process test shard can run directly under Bun without false failures caused by the test host.

## Evidence

- Before the fix, the two event-loop stall regressions lost their TAIL output under direct Bun.
- Direct custom-Bun focused process tests: 58 passed, 5 platform skips.
- Node focused process tests: 58 passed, 5 platform skips.
- Direct custom-Bun full process shard: 41 files passed, 632 tests passed, 64 platform skips.
- pnpm check:changed passed through all diff-owned checks; the initial run reached an unrelated main type failure later fixed by #145540.
- Independent P0-P2 autoreview: scoped clean at 0.91 confidence.

### DIFF
--- a/src/process/child-process.ts
+++ b/src/process/child-process.ts
@@ -13,12 +13,12 @@ const EXIT_STDIO_MAX_DRAIN_MS = 1_000;
  */
 export function releaseChildProcessOutputAfterExit(child: ChildProcess): () => void {
   let idleTimer: NodeJS.Timeout | undefined;
-  let releaseImmediate: NodeJS.Immediate | undefined;
+  let releaseTimer: NodeJS.Timeout | undefined;
   let deadlineTimer: NodeJS.Timeout | undefined;
 
   const cleanup = () => {
     clearTimeout(idleTimer);
-    clearImmediate(releaseImmediate);
+    clearTimeout(releaseTimer);
     clearTimeout(deadlineTimer);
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [b5a7965c3b90dbd9990dd6eb5794a9c322efaaae](https://github.com/openclaw/openclaw/pull/145564/commits/b5a7965c3b90dbd9990dd6eb5794a9c322efaaae)
- Landed commit: [04363ff3b75aeb6a5c56a6673dfa9f5651623f56](https://github.com/openclaw/openclaw/commit/04363ff3b75aeb6a5c56a6673dfa9f5651623f56)

PR #145377, opened 2026-09-11
feat(radius): connect models with browser sign-in and native streaming

steipete:
## What Problem This Solves

OpenClaw users cannot currently select Radius models or sign in to Earendil's Radius gateway. Radius uses Pi's native message protocol, so configuring it as an OpenAI-compatible endpoint does not work.

## Why This Change Was Made

Adds a Radius provider plugin that owns device-code browser sign-in, OAuth refresh, organization API keys, authenticated model discovery, and streamed text, reasoning, and tool calls. The shared model-data contract admits `pi-messages`; the transport remains plugin-owned and uses existing provider hooks. No new dependency is needed.

The catalog publishes transport metadata before registry validation and preserves account-advertised capabilities and tiered pricing. Cold agent startup resolves models through the same reader, respecting pinned auth profiles. Streaming preserves signed content and final tool arguments, reports missing terminal events as failures, and releases responses on completion or cancellation.

The provider page documents setup, model selection, billing, troubleshooting, and the boundary with Radius's optional Pi extension tools. Plugin publication metadata, discovery, navigation, generated inventory/reference, and labels are wired. The configuration inventory adds the same 21 standard plugin controls used by other plugins.

## User Impact

Users can sign in with `openclaw models auth login --provider radius --method oauth --set-default`, use an organization API key or `RADIUS_API_KEY`, and select account-visible `radius/<model>` entries. Setup recommends `radius/balanced` when available and otherwise the first available model. Organization routing, rewrite policies, model access, and credits apply normally.

Release preflight stamps the plugin API/build version to the release containing Radius. Package peers, installation, and loading enforce that supporting-host requirement. The unstamped source package is not qualified for older hosts. Existing provider API values, model overrides, and defaults are preserved.

## Evidence

- Official contracts inspected: [Radius provider](https://github.com/earendil-works/pi/blob/main/packages/ai/src/providers/radius.ts), [native protocol](https://github.com/earendil-works/pi/blob/main/packages/ai/src/api/pi-messages.ts), [OAuth](https://github.com/earendil-works/pi/blob/main/packages/ai/src/auth/oauth/radius.ts), and [Radius documentation](https://radius.earendil.com/docs).
[...]

PR #145565, opened 2026-09-12
fix(radius): Claude requests fail with thinking disabled

steipete:
Related: #145377

## What Problem This Solves

Fixes an issue where users running Radius Claude models with thinking disabled receive an upstream `max_tokens: Input should be a valid integer` error instead of a reply. A live sweep of Radius's 28 advertised entries reproduced this for Haiku 4.5, Opus 4.1/4.5, and Sonnet 4/4.5.

## Why This Change Was Made

Radius's native Pi protocol represents disabled reasoning by omitting the option. OpenClaw's adapter forwarded its explicit `off` value instead. Pi's older Anthropic budget path has no budget for that value and produces a non-finite output limit.

Normalize only `off` to an omitted field at the Radius transport boundary, matching [Pi's agent contract](https://github.com/earendil-works/pi/blob/main/packages/agent/src/agent.ts). Explicit enabled reasoning, numeric output caps, authentication, and all other request options remain unchanged.

## User Impact

Radius requests with thinking disabled no longer fail because of malformed thinking/output budgets. The correction stays inside the Radius plugin; shared reasoning semantics and other providers are unchanged.

## Evidence

- Baseline live sweep: 23 of 28 entries completed a streamed model-to-read-tool-to-model round trip; five older Claude entries failed with the same integer-schema error.
- New wire-contract regression fails before the fix because the JSON payload contains `reasoning: "off"`; after the fix it is omitted and `maxTokens: 1024` remains numeric. The existing signed-stream case preserves coverage for explicit `high` reasoning and other native options.
- All 41 Radius tests passed. The selected changed-file checks passed, including extension/test typechecks, lint, formatting, plugin boundaries, dead exports, and import cycles.
- Independent P0-P2 review is clean.
- The corrected live run already proves Haiku 4.5, Opus 4.5, and Sonnet 4.5 complete their tool round trips. Opus 4.1 and Sonnet 4 now reach a distinct upstream model-not-found response despite being advertised by Radius; the plugin preserves those service-owned catalog entries and surfaces their error.
- Completed corrected live sweep: **26/28 passed**, each with native streaming, a real read-tool round trip, exact synthetic file output, and no fallback. The only remaining failures are the two upstream model-not-found responses above. Test conditions stayed the same: thinking off, a 1,024-token output cap, and isolated state. Corrected sweep reported usage was $0.32388551. [Required exact-head CI passed](https://github.com/openclaw/openclaw/actions/runs/34670505166).

### DIFF
--- a/extensions/radius/stream.test.ts
+++ b/extensions/radius/stream.test.ts
@@ -75,6 +75,20 @@ beforeEach(() => {
 });
 
 describe("Radius native message transport", () => {
+  it("omits disabled reasoning from Pi requests without changing the output cap", async () => {
+    fetchMock.mockResolvedValue(response([{ type: "done", reason: "stop", usage }]));
+    await createRadiusStreamFn()(model, context, {
+      ...options,
+      reasoning: "off",
+      maxTokens: 1024,
+    }).result();
+    expect(await new Response(fetchMock.mock.calls[0]![1]?.body).json()).toEqual({
+      model: model.id,
+      context,
+      options: { maxTokens: 1024 },
+    });
+  });
+
   it("bounds large-tool preview work while emitting every delta and the final arguments", async () => {
     const toolArguments = { content: "a".repeat(100_000) };
     const json = JSON.stringify(toolArguments);
--- a/extensions/radius/stream.ts
+++ b/extensions/radius/stream.ts
@@ -283,7 +283,8 @@ export function createRadiusStreamFn(): StreamFunction<string, RadiusStreamOptio
           options: {
             temperature: options.temperature,
             maxTokens: options.maxTokens,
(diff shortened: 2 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [fa1feba154a5258e443d54aad56d4abc489fea5b](https://github.com/openclaw/openclaw/pull/145565/commits/fa1feba154a5258e443d54aad56d4abc489fea5b)
- Landed commit: [99e963cb1ab9c97f44154929d3a98cb851f39bc9](https://github.com/openclaw/openclaw/commit/99e963cb1ab9c97f44154929d3a98cb851f39bc9)

PR #145480, opened 2026-09-12
fix(workers): reduce startup delay for prepared GitHub repositories

steipete:
Closes #145472. Related: #145302 and #145331.

## What Problem This Solves

Users starting a prepared GitHub repository worker still wait for nine sequential source-metadata requests across initial admission, pre-bind admission, and post-bind revalidation. A live Daytona ready-reserve sample spent about 3.2 seconds of its 5.97-second activation in these checks.

## Why This Change Was Made

For authenticated pinned reads with a known repository ID, resolve the repository and commit through the immutable Repository node, then confirm the repository again through REST. This keeps the commit associated with the correct repository instance and preserves all three admission gates and their current identity/ownership checks.

The existing GitHub transport also owns GraphQL response decoding and quota cooldowns. Explicit, wholly refused permission responses can use the complete original REST sequence so scope-limited native credentials keep working. Partial/malformed data, missing commits, identity or visibility changes, cancellation, and quota errors do not trigger that fallback.

## User Impact

The normal authenticated ready path uses seven metadata requests instead of nine, with no configuration change. First-time name/branch discovery and anonymous reads retain their existing REST behavior.

## Evidence

- Live calls through the changed admission owner completed three initial/pre-bind/revalidation cycles with seven requests each; the unchanged owner used nine. Both used an isolated managed credential and a synthetic private repository.
- The changed owner rejected a real missing commit after one GraphQL request, with no REST fallback. Cancellation and a removed agent were rejected before network access.
- The preceding full Daytona replay verified ready reuse, isolation, accepted deletion restoration, restart-preserved expiry, and cleanup of four workers and one snapshot. It measured 8.39 seconds to first output for a fresh ready worker and 7.87 seconds for restoration, using synthetic inference. Those are baseline samples for the landed startup changes, not end-to-end timings for this PR.
- 223 owner/sibling tests passed across ten suites; 91 affected tests passed on the final candidate. The complete changed gate, both typecheck lanes, lint, and independent review through P2 passed.

[...]

Issue #138724, opened 2026-09-05
Cloud sessions should start without a Gateway checkout


steipete:
## Summary

Start a cloud session from a GitHub repository URL and ref without cloning the source or creating a working tree on the Gateway.

## Problem to solve

Cloud session creation currently clones the project on the Gateway, creates a session-owned managed worktree, then provisions a worker and synchronizes eligible files. This puts Gateway checkout, setup, disk space, and transfer work on the startup path even when all coding runs remotely. Choosing a cloud destination should not require a local source copy.

The dependency is architectural: completed results are applied to the Gateway worktree, successful temporary result refs are deleted, and the publication broker requires a managed worktree. Removing only dispatch validation would leave recovery and publication without their canonical workspace.

## Proposed solution

Use the existing repository URL/ref inputs to create a session-owned repository workspace. Clone and run authorized project setup on the remote worker through the existing node workspace owner. Worker-turn and Codex remote execution share that workspace lifecycle. Preserve the current create → dispatch → send recovery flow in the Control UI.

The Gateway retains repository identity and cumulative changed-file checkpoints, without an unchanged source tree. Completed turns and normal Stop must save an authenticated checkpoint before releasing their current placement/claim. Redispatch reconstructs the remote checkout from the pinned base commit and checkpoint. An explicit move to the Gateway may materialize a local checkout. Existing local-worktree session contracts remain supported.

### Accepted persistence design

[Maintainer acceptance](https://github.com/openclaw/openclaw/issues/138724#issuecomment-5548486526) supersedes the earlier schema-bump proposal: use additive SQLite storage at version **15**, and do not preserve backward compatibility for older cloud-session implementations.

- A shared-state `session_repository_workspaces` row owns the normalized URL/ref, pinned base commit, branch, cumulative accepted checkpoint, revision, and logical session identity. Session metadata references this owner without inventing a local path or worktree ID.
- Session-owned bare Git artifacts hold complete bounded manifests and cumulative changed-file bytes. They are recovery artifacts, not a Gateway source clone or working tree. Accepted revisions remain available until permanent session deletion.
- Both runtimes, editor writes, Stop, recovery, fork/reset/archive, and explicit moves share the workspace and claim owners. Node queues physically drain before reconciliation or reuse; cancellation alone is insufficient. Pending results retain their exact repository source.
[...]

PR #145571, opened 2026-09-12
perf: reduce repository checkpoint latency

steipete:
Related: #138724, #145480

## What Problem This Solves

Repository-backed cloud sessions wait for redundant local Git reads while creating their durable checkpoint. This adds latency to workspace preparation and completed-turn reconciliation.

## Why This Change Was Made

Reuse the commit ID that staging already resolves, resolve both candidate commits in one Git call at each existing verification point, and read both immutable publication destinations together under the existing mutation lock. Commit peeling, exact ref matching, live ownership checks, atomic Git publication before SQLite acceptance, and retryable cleanup retain their existing semantics.

## User Impact

Checkpoint creation uses two fewer Git subprocesses without a publication snapshot and five fewer with one. The persisted checkpoint format and recovery behavior stay compatible.

## Evidence

- 31 focused checkpoint, concurrent staging, and repository-startup tests passed, including annotated-tag peeling, candidate replacement/removal, immutable publication collisions, cumulative edits/deletions, interrupted acceptance, and cleanup retry.
- Selected `check:changed` passed, including core and test type checks, scoped lint, and static guards. Independent review through P2 found no actionable findings. The initial OpenGrep run flagged manual SHA-1 hashing in the Git-blob fixture; the fixture now calls Git `hash-object` directly. Its 10 checkpoint tests, selected checks, and fresh P2 review passed after that test-only correction.
- Real-Git benchmark with five alternating-order baseline/candidate pairs per mode: publication used 23 → 18 calls, with medians 494.14 → 404.34 ms and median paired saving 84.15 ms (candidate faster in all five pairs). Without publication, calls fell 16 → 14; medians were 612.90 → 506.14 ms, but only three of five pairs were faster because of host timing noise. Every sample verified durable acceptance, deletion preservation, and recovery/publication bytes. These measure stage → verify → publish locally, with readback afterward.
- Both frozen packages passed all nine qualification phases, real installed Gateway/worker execution, and launcher start/restart/cleanup checks. Runtime proof uses baseline `bca261d5583d` and candidate `d19ecdc880401`; final head `8accc3536fa5` changes only the non-shipped test fixture and preserves both production files byte-for-byte.
- Both Daytona/Crabbox 0.56 live campaigns passed prepared-worker dispatch, session isolation, deletion checkpoint restoration, Gateway restart, natural expiry, and cleanup. All eight machines and both captured images were independently confirmed absent. With synthetic inference, prepared fresh-session dispatch to first output was 10.434 → 8.047 s; restoration was 7.928 → 7.075 s. These are one sample per lifecycle and version, so provider/network/process variation prevents attributing the full difference to this patch. The local paired benchmark is the stronger evidence for the checkpoint-specific gain.
- [CI run 34671195559](https://github.com/openclaw/openclaw/actions/runs/34671195559) passed on final head in attempt 2. One unrelated profiler subprocess test timed out initially; its unchanged job rerun and dependent required gate passed. No timeout or assertion was weakened.

### DIFF
--- a/src/gateway/worker-environments/session-repository-checkpoints.test.ts
+++ b/src/gateway/worker-environments/session-repository-checkpoints.test.ts
@@ -20,6 +20,7 @@ import { readActualWorkspaceManifest } from "./workspace-reconcile-core.js";
 import { requireWorkspaceResultGit } from "./workspace-result-git.js";
 import {
   hasWorkerWorkspaceResultRef,
+  readStagedWorkerWorkspaceResult,
   workerWorkspaceResultRef,
 } from "./workspace-result-staging.js";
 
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8accc3536fa59da9f8baf1b2225ca484cd503887](https://github.com/openclaw/openclaw/pull/145571/commits/8accc3536fa59da9f8baf1b2225ca484cd503887)
- Landed commit: [64a19e38fc12b4e9cfd3d6955283eac48d1c6481](https://github.com/openclaw/openclaw/commit/64a19e38fc12b4e9cfd3d6955283eac48d1c6481)

PR #145572, opened 2026-09-12
feat: support inline browser panels in the Tauri companion

### LINKED ISSUE #145502: Tauri companion: support the shared inline browser workflow
The macOS app hosts real browser tabs inside the shared Control UI, but the Tauri companion does not provide the native browser bridge. Links and the Browser panel therefore cannot use the same inline browsing workflow.

Bring the Tauri app onto the existing Control UI browser contract: session-scoped tabs, native presentation, navigation/history/popups, screenshots, element inspection, and native downloads. Keep the browser session separate from the Gateway dashboard and clear it when switching Gateways or closing the app.

Validation should exercise the actual Windows WebView2 app with synthetic local pages, native save/cancel dialogs, focused Rust tests, and the existing Linux/macOS companion CI jobs. This extends the native host of an existing shared UI capability; it does not add a new core tool or persisted configuration.

steipete:
Closes #145502

## What Problem This Solves

The Tauri companion could open the Gateway dashboard but could not host the inline Browser workflow already available in the macOS app. Reading links, navigating tabs, inspecting elements, and capturing pages therefore lacked the native browser surface expected by the shared Control UI.

## Why This Change Was Made

Implement the existing native browser bridge in the Tauri host, with platform adapters for WebView2, WebKitGTK, and WKWebView. One host owns conversation tabs, navigation state, popup ownership, presentation scopes, screenshots, inspection, and native asset downloads. Reading tabs share a private session and receive no dashboard command capability or authentication scripts. Dashboard reload retains tabs; changing the selected dashboard resets them.

Native downloads complete only after the chosen file has been saved. On Windows, the save picker runs on its own STA so concurrent requests and cancellation remain responsive; a deliberate repeated download receives only a single-use, nonpersistent allowance bound to the current host request.

## User Impact

Users can open pages inside the companion, keep separate tabs for conversations, navigate with the existing toolbar, inspect or capture content, and save the current page or asset with the native picker. Popups remain in their opener's conversation. The shared UI labels these as App tabs.

## Evidence

- Windows: 142 native tests passed; the real WebView2 regression passed 20/20, including failed-redirect recovery, HTTP Basic authentication, conversation deduplication, shared private cookies, popup ownership, navigation and SPA history, presentation scopes, PNG snapshots, inspection, dashboard reload, and cleanup.
- Windows native UI: shared Browser address bar, Back, Forward, and Reload verified; repeated Save and Cancel verified, saved bytes matched the fixture exactly, and duplicate download requests were rejected while the picker was open.
- Linux: 165 native tests passed; one existing opt-in logind test remained ignored. Real GTK/WebKit testing verified trusted pointer input, inspection, PNG snapshots, history, dashboard reload, actual dashboard replacement, native Save with exact byte comparison, and Cancel without staging residue.
- Shared UI: 18 bridge tests, UI typechecking, scoped lint, i18n verification, and production build passed. The Linux packaged-runtime scanner's 15 tests passed on Linux. Rust formatting, workflow YAML, and whitespace checks passed.
- The Linux App PR job now runs the synthetic native browser regression and uploads screenshots and results as the `linux-inline-browser` artifact. Remote CI passed: 165 Linux native tests and the real GTK/WebKit browser regression; macOS 15 arm64 compiled, linked, and ran all 149 native tests successfully. [Native CI](https://github.com/openclaw/openclaw/actions/runs/34672144797) and [browser screenshots/results](https://github.com/openclaw/openclaw/actions/runs/34672144797/artifacts/10291226266). No live macOS UI session was available for this task.

Independent review found two P2 parity defects. Both were reproduced on the earlier native binary and fixed; the same expanded Windows regression passed 20/20 afterward, and the Linux live regression covers both fixes. An independent follow-up audit found no remaining actionable issue in the corrections. The initial CI run exposed an outdated workflow-routing fixture that lacked the new browser step and assumed PRs uploaded no artifacts. Updated it to cover the browser proof artifact and success/failure/cancel/skip outcomes while preserving manual-only packaging; all seven focused cases pass, and independent review found no actionable issue. A later dependency scan identified the documented Windows live driver as an unregistered executable root. Added its exact path to the existing Knip entry registry, without ignores or dependency suppressions; independent review found no issue. Final clean CI passed: 115 successful checks, zero failures, including the dependency scans, UI browser suites, native Linux live proof, and macOS native tests. [Final CI](https://github.com/openclaw/openclaw/actions/runs/34672143528). ClawSweeper reviewed final head `ed2a208906418c8b2f3a0eb63329ebab616bbc5f`, accepted the added Windows proof, and reported no actionable findings or remaining before-merge items. Merged through the normal required CI gate as `ce5c1cb1b1d398778a8254f5ee12654b2436b1ad`. No package release is part of this change.

### Additional native Windows download proof

Addressed the review's authority-boundary proof request on the real WebView2 candidate (native source revision `916ad4fad726b100700ec444a373e36d462f6a65`, unchanged by the later Knip entry registration), using synthetic loopback HTTP fixtures and the real Windows Save As dialog:

- An authorized warmup save completed with `ok: true, cancelled: false` and 89 exact fixture bytes.
- During a second authorized save with its HTTP response held at one byte, the fixture page requested a different attachment twice: once while Save As was open and once after the destination was chosen. Both requests reached the fixture server. WebView2 `Browser.downloadProgress` reported `state: canceled`, `receivedBytes: 0`, `totalBytes: 24` for both competing download IDs. Neither created a file in the selected directory or the default Downloads directory. The authorized transfer then completed with the exact 89 bytes.
- A third transfer was held at one byte. After choosing `revoked.html` in the native picker, filesystem inspection confirmed its private staging file existed with length 1. Reloading the dashboard destroyed the requesting document context. After releasing the remaining response bytes, `revoked.html` was absent and there were zero staging files. The earlier authorized save remained byte-for-byte correct.
- CDP was used to invoke the production bridge, drive only synthetic page actions, and observe download events; Chromium download behavior stayed `default`. Native picker selection used actual Windows UI input. No permission persistence or native authorization checks were disabled.

The review also requested a live macOS UI smoke. No macOS UI host is available in this task, so that claim remains explicitly unverified. The real macOS 15 arm64 CI job compiled, linked, and executed all 149 native tests successfully; Windows and Linux provide the live UI proof for the Tauri companion in this task. No concrete macOS defect was reported by review.


steipete:
@clawsweeper re-review

PR #145541, opened 2026-09-12
fix: avoid full agent database copies during startup version checks

steipete:
Closes #145412

## What Problem This Solves

Fixes an issue where Gateway startup and restart could spend time and temporary disk copying large agent databases even when only checking their schema versions. The header-only preflight path used full-snapshot machinery before reading a few metadata values.

## Why This Change Was Made

Header-only agent preflight now uses the existing read-only worker to read a coherent SQLite schema/version header without copying ordinary database payloads. The worker retains source-reader admission through native database closure; the parent retains cancellation and cleanup ownership through child closure. WAL-visible version and writer metadata come from the same read transaction.

The header reader remains separate from lightweight version utilities, preserving the native hook relay's startup import contract. Metadata SELECT uses Kysely, read transactions use the existing deferred transaction owner, and read-only PRAGMAs belong to the existing SQLite configuration owner. Source-reader lock tolerance stays at the previous 30 seconds; the global/runtime and private-snapshot timeout and child deadlines are unchanged.

Full integrity, foreign-key, schema readiness, agent-ownership admission, physical writable-open validation, and shared-state inspection remain unchanged. The rebased integration preserves the private snapshot whenever `agentAdmissionConfig` requests ownership inspection, even without full schema validation. Rollback recovery, incomplete WAL families (including cleanly closed WAL stores without sidecars), and existing exclusive source/mutation scopes retain their necessary private snapshots. Complete WAL families keep the previous reader's native SHM coordination; sidecar presence does not prove live native clients. There is no catch-all retry/copy fallback, new persistent cache, schema migration, configuration option, or SDK surface expansion.

The accepted storage-inspection scope is recorded in #145412. Existing installed update/deployment drivers remain unchanged, and rollback requires no data conversion. Worker bootstrap, worktree cleanup, and broader cross-lease validation reuse are out of scope.

## User Impact

Ordinary agent version checks no longer allocate temporary snapshots proportional to retained history. Required full validation still runs at its existing ownership boundaries. This removes measured startup I/O amplification; it does not claim to eliminate every cold-start copy or establish the cause of a historical production integrity timeout.

## Evidence

[...]

PR #143802, opened 2026-09-10
improve(doctor): speed up schema checks for large databases

VACInc:
## What Problem This Solves

Doctor and Gateway restart readiness checks copy whole agent databases even when they only need schema metadata. The copy cost and temporary disk requirement grow with stored transcripts. The previous revision also bypassed the optimization for configured agents—the actual Doctor and restart callers.

Related: #139870. The shared-state lint-copy improvement in #145577 and header-only startup optimization in #145541 are complementary, not replacements.

## Why This Change Was Made

Eligible agent schema checks read version, shape, and ownership metadata in one coherent read-only child transaction. The parent applies the existing ownership admission decision, including isolating secondary agents and refusing default/system-agent mismatches.

The child uses the canonical source-read lifetime owner, retaining its lease if native SQLite close fails until the child exits. Empty files, rollback recovery, incomplete WAL families, owner-provided snapshots, and full startup readiness keep the existing snapshot path. Ordinary busy/I/O failures are not retried as full copies.

The explicit copied-agent-file preflight now lives beside the agent schema inspector; its CLI and test callers were moved without changing behavior. This keeps the main preflight owner below the lint size limit. The earlier unused-import CI repair is now present on main and no longer contributes to this PR.

**Data compatibility:** no schema version, table, index, stored-row format, config, migration writer, or recovery writer is changed. The existing maintenance validator remains authoritative. Removing this optimization requires no data conversion.

## User Impact

Configured agents get the intended schema-inspection optimization without weakening ownership admission or startup integrity checks. A source whose native close fails cannot prematurely allow canonical replacement.

## Evidence

Current head: `d4817de7b5f93a8445d3b26b88d284bc736e5d04`, rebased again onto main `d0ec2f18a5b`. CI is not claimed green until its exact-head run finishes.

- **Configured-agent regression:** DELETE and active-WAL cases through `assertOpenClawDatabasesReady` for Doctor and Gateway restart failed before the repair with “No space for a full agent database snapshot.” They pass after the repair, including foreign-owner refusal, fresh newer-schema metadata, and missing-index detection.
- **Native-close regression:** expanded the existing two-process lifetime fixture to cover the agent schema reader. A failed native close allowed source exclusion before the repair; after the repair, exclusion remains unavailable until child exit. Success, read failure, cancellation, and coherent metadata under a concurrent WAL commit also pass.
[...]

PR #145577, opened 2026-09-12
improve(doctor): reduce repeated state copies during update checks

VACInc:
### High Level TLDR

Doctor's post-plugin readiness check makes a private copy of shared state, then unnecessarily copies that private database again for subsequent reads. This change keeps those reads on the disposable database already owned by the check.

## What Problem This Solves

Fixes an issue where update finalization performs redundant full shared-state copies during Doctor lint. The existing real-SQLite readiness scenario produces three snapshots: the original operator-state snapshot and two copies of that temporary database.

### Root Cause

Doctor correctly wraps lint in artifact-preserving reads and creates a private plugin-state view. The shared read owner inherits the artifact-preserving context even after reads have been redirected to that disposable view, so fresh reads snapshot it again.

## Why This Change Was Made

The shared read owner now recognizes an explicitly scoped disposable path. Doctor activates that scope only around its own private database. Native read-only handles, schema checks, and tracked handle cleanup remain with their existing owners. Other paths retain artifact-preserving reads, nested scopes remain independent, and inherited asynchronous contexts lose the exemption when their owner finishes.

No persistent schema, data representation, migration, configuration option, or cross-run cache changes. Installed update drivers remain unchanged; the optimization runs inside the candidate's existing Doctor lint path. Rollback needs no data conversion.

## User Impact

A synthetic 206,671,872-byte shared database produces one full snapshot instead of three, avoiding 413,343,744 bytes of redundant copied output and two snapshot workers per exercised readiness pass. This is a bounded I/O optimization, not a claim to resolve all update latency.

### Real behavior proof

The existing registered post-plugin readiness scenario runs real SQLite readers and snapshot workers. Its new assertion fails on the original code because two extra private paths are copied, and passes with this patch. It still reports the required missing local-embedding setup error, preserves the operator database family byte-for-byte, and never opens the operator database in the parent.

- 70 focused tests passed, including nested path scopes, read-only writes being rejected, unrelated source protection, and revocation of inherited async contexts.
- Targeted lint, formatting, and `git diff --check` passed.
- Independent Astra high autoreview: scoped-clean at P0–P2.
- Full `pnpm build` and the complete four-file `check:changed` gate passed, including core/core-test typechecks, dead-export scans, scoped lint, and storage/import guards.
- Published `openclaw@2026.9.4` updater × exact candidate `9cdebdc907a1f629174bab85cfc0a3d4c43de329`: **passed** using the repository's isolated Docker `sqlite-volume` scenario. The actual installed updater performed the npm update; the scenario preserved 4,800 sessions, 23,890 events, and 2,200 cron jobs, verified eight conversations across two agents before and after Gateway restart, and passed repeat Doctor idempotence in 18 seconds.

Focused tests:

```sh
node scripts/run-vitest.mjs src/state/openclaw-state-db-readonly.test.ts
node scripts/run-vitest.mjs src/commands/doctor-lint.test.ts src/commands/doctor-lint.state-isolation.test.ts
```

Published-driver cell (Node 24.19.0/Linux; candidate built and packaged with the canonical package helper):

```sh
OPENCLAW_CURRENT_PACKAGE_TGZ=/tmp/doctor-lint-package/openclaw-current.tgz \
OPENCLAW_DOCKER_E2E_SELECTED_SHA=9cdebdc907a1f629174bab85cfc0a3d4c43de329 \
OPENCLAW_UPGRADE_SURVIVOR_PUBLISHED_BASELINE=1 \
OPENCLAW_UPGRADE_SURVIVOR_BASELINE_SPEC=openclaw@2026.9.4 \
OPENCLAW_UPGRADE_SURVIVOR_SCENARIO=sqlite-volume \
bash scripts/e2e/upgrade-survivor-docker.sh
```

Copied output:

```text
sqlite-volume sessions=4800 events=23890 cronJobs=2200
sqlite-volume Gateway RPC verified 8 conversations across 2 agents
SQLite volume idempotence doctor completed in 18s (budget 60s).
Upgrade survivor Docker E2E passed baseline=openclaw@2026.9.4 scenario=sqlite-volume candidate=2026.9.4 updateRestartMode=manual idempotence=18s startup=24s updateRestart=manuals healthz=1s readyz=1s status=1s.
```

Update receipt: `status=ok`, `mode=npm`; published build `2026.9.4-release-3a9d69db306c-2026-09-10T22-53-16.719Z` → candidate build `2026.9.4-9cdebdc907a1-2026-09-12T03-42-19.063Z`. Local Docker was used because hosted Package Acceptance accepts upstream-reachable refs, while this candidate is on the author fork. Its source-trust gate was preserved.

### Performance limits

One warm run showed a large timing difference, but the repeated comparison measured essentially identical warm medians: 0.970 s baseline and 0.977 s candidate. Disk/cache variability prevents a reliable percentage latency claim. Snapshot counts consistently fell from three to one; findings and operator-state bytes remained unchanged.

### What was not tested

No production update, restart, or operator-state mutation was performed. No claim is made about total update duration or Windows/macOS performance.

Related: #143802 and #145541 avoid agent-database copies during schema/header inspection. This PR addresses repeated shared-state copies inside Doctor's already-isolated lint view.

The first ClawSweeper review found no actionable correctness/security defects and requested the published-updater evidence now recorded above.

### DIFF
--- a/src/commands/doctor-lint.test.ts
+++ b/src/commands/doctor-lint.test.ts
@@ -793,6 +793,9 @@ describe("runDoctorLintCli", () => {
           },
         ],
       });
+      expect(
+        mocks.prepareSqliteReadOnlyLocationSync.mock.calls.map(([pathname]) => pathname),
+      ).toEqual([databasePath]);
       expect(sourceOpenStacks).toEqual([]);
       expect(snapshotDoctorLintSqliteFamily(databasePath)).toEqual(before);
     } finally {
--- a/src/commands/doctor-lint.ts
+++ b/src/commands/doctor-lint.ts
@@ -34,7 +34,10 @@ import {
 } from "../plugins/install-root-context.js";
 import type { RuntimeEnv } from "../runtime.js";
 import { closeOpenClawStateDatabaseByPathAsync } from "../state/openclaw-state-db-cache.js";
-import { withArtifactPreservingStateReads } from "../state/openclaw-state-db-readonly.js";
+import {
+  withArtifactPreservingStateReads,
+  withDisposableOpenClawStateReads,
+} from "../state/openclaw-state-db-readonly.js";
 import { resolveOpenClawStateSqlitePath } from "../state/openclaw-state-db.paths.js";
 import { isPostCoreConvergencePass, isUpdateDoctorLintPass } from "./doctor/shared/update-phase.js";
 
@@ -355,12 +358,11 @@ async function withReadOnlyPluginStateSnapshot<T>(
       // Global readers and OAuth refresh/challenge writers share the private state view.
       outcome = {
         ok: true,
-        value: await withPluginInstallRoots(
-          { ...installRoots, stateDir: privateStateDir },
-          async () => {
+        value: await withDisposableOpenClawStateReads(privateDatabasePath, () =>
+          withPluginInstallRoots({ ...installRoots, stateDir: privateStateDir }, async () => {
             runStarted = true;
             return await run(privateEnv);
-          },
+          }),
         ),
       };
     } catch (error) {
--- a/src/state/openclaw-state-db-readonly.test.ts
+++ b/src/state/openclaw-state-db-readonly.test.ts
@@ -27,6 +27,7 @@ import {
   withExistingOpenClawStateDatabaseArtifactPreservingReadOnlyAsync,
   withExistingOpenClawStateDatabaseReadOnly,
   withArtifactPreservingStateReads,
+  withDisposableOpenClawStateReads,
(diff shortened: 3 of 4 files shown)


VACInc:
Rebased onto main at d093d651f71; head 813671de69b. Preserved upstream asynchronous private-state cleanup and retained disposable-reader scope. Focused Doctor lint and read-only state suites: 55 tests passed. Formatting and diff checks passed.

VACInc:
#143802 merged to main with an agent schema-shape read optimization. This PR's reuse of Doctor's already-disposable shared-state lint database is complementary work; neither change supersedes the other.

Issue #145579, opened 2026-09-12
Reduce redundant Gateway work under sustained session activity


steipete:
All three fixes are landed and verified on main. Isolated Gateway workloads on baseline `47ce03dbc6994435af783fecf13d5586e7b307f4` identified three avoidable costs in existing owners. The fixes preserve session ownership and overrides, expiry behavior, lock identities, permission checks, and fresh resolution after authority callbacks:

- #145580 reuses configured model defaults within a session-list response. A synthetic 10,000-session search with 100 aliases improved from a median 1,303 ms to 708 ms; the resolver regression dropped from 42 calls to 2.
- #145581 skips recipient-map expiry sweeps before an expiry can be due. Processing 10,000 tool events through the actual handler with 10,000 tracked runs improved from a median 924 ms to 10 ms. Due expiry still scans the map.
- #145582 reuses canonical path facts within one ordinary coordinator acquisition. Median acquire/release time improved from 1.576 ms to 1.159 ms, with filesystem resolutions reduced from 4 to 2. This adds no persistent cache or schema, transaction, or lock-policy change.

These are component measurements with synthetic input. Independent reviews, regression and sibling tests, and exact-head hosted CI accompany the PRs. The session PR also passed 46 focused Linux tests after retaining both its regression cases and main's adjacent catalog-cache test during conflict resolution.

A separate combined-tree live OpenAI validation passed on an isolated Linux Testbox: 1,000 sessions, eight concurrent turns, 100 session updates, and 303 history probes. All eight streamed/final/RPC-history/independent post-shutdown SQLite transcript comparisons passed, with zero failed probes. Session-list p95 was 713 ms, history p95 277 ms, and update p95 60 ms. The Gateway exited cleanly, its owned process group was confirmed dead, and the task-owned Testbox lease was stopped and verified completed. No operator Gateway or state was changed.

The live fixture denied tools, disabled dreaming, and capped output at 128 tokens. It validates the combined three production fixes before the session branch's later main integration; it is not a paired performance comparison, live tool-delivery proof, or final-main live test. Eight completions were observed; total provider request count is unknown. Result SHA256: `ed3b506456f679066a4dc57458a09d4632c0cac4418cd965894bd3c4e5083aea`.

[...]

PR #137901, opened 2026-09-04
improve: reduce CPU usage for session lists with large catalogs

steipete:
## What Problem This Solves

Fixes an issue where opening or refreshing session lists consumes avoidable Gateway CPU time when the prepared model catalog is large. Session settings responses also repeat the selected model lookup.

## Why This Change Was Made

Reuse catalog hits and misses in the existing per-response row context, partitioned by the exact catalog array and provider/model identity. Carry the selected catalog entry from thinking projection into row and patch context-window projection, including ACP sessions. Separate agents retain their own catalog metadata, and each response starts fresh.

## User Impact

Large session lists require less synchronous Gateway work, leaving more event-loop time for other requests. Session fields, model selection, configuration behavior, and the public response shape remain unchanged.

## Evidence

- Regression through the actual asynchronous list owner: 80 ordinary/ACP rows previously triggered more than 140 underlying catalog lookups; the repaired response stays below 10. It covers hits, misses, case-distinct model IDs, separate agent catalogs with identical IDs, replaced catalog entries in the next response, patch context-window metadata, and exclusion of internal catalog fields from public results.
- Synthetic benchmark on Node 26.8.2: 2,921 sessions, 200 selected rows, a 1,000-model catalog, and 50 measured iterations after warm-up. Mean synchronous row projection fell from 58.46 ms to 24.44 ms (58.2%). Post-GC heap remained approximately 165 MiB. These are local projection measurements, not a live request-latency or OOM-resolution claim.
- All 306 tests passed across session utilities, performance, thinking defaults, ACP ownership, model identity, and plugin-runtime suites. The subsequently strengthened mixed-case regression passed separately, as did its Gateway test typecheck.
- All four existing GC tests passed on the final commit: retired list generations, child-link metadata, and scalar/batched transcript-title retention.
- The complete changed-file gate passed, including core types, all 16 test typecheck graphs, export scans, lint, and storage/runtime/import guards.
- Independent P2 review completed. Its single case-folding concern was rejected after checking the existing case-preserving key helper and proving distinct model names through the actual-list regression; there are no accepted findings.

Earlier September 4 measurements and built-Gateway parity results in this PR's history apply to its previous implementation. They remain historical evidence; the current revision's measurements and checks above are separate.

PR #145580, opened 2026-09-12
perf(gateway): reuse configured session models during listing

steipete:
<details>

</details>

## What Problem This Solves

Fixes an issue where users searching large session lists experience slow responses when many model aliases are configured. In the synthetic fixture, filtering 10,000 stored sessions occupies the Gateway event loop for over a second.

## Why This Change Was Made

Reuse each agent's configured default through the existing per-list model cache, including for rows without overrides. Cache identity preserves normalization mode; native session ownership and per-session parent override reads retain their existing behavior.

## User Impact

Session list searches and page projection spend less time repeating model selection. Returned models, inherited overrides, and ownership remain unchanged.

## Evidence

Synthetic component benchmark on Windows, Node 24.19.0, against baseline `47ce03dbc6994435af783fecf13d5586e7b307f4`: actual `listSessionsFromStoreAsync`, 100 configured model aliases, populated runtime model fields, page limit 100, five measured runs after two warmups.

| Stored sessions | Request | Median before | Median after |
| --- | --- | ---: | ---: |
| 1,000 | Search with no matches | 136.8 ms | 74.6 ms |
| 10,000 | Search with no matches | 1,303.0 ms | 708.0 ms |
| 10,000 | First page | 54.6 ms | 39.3 ms |

- Both new regression cases fail on the original code with 42 model resolutions for 40 sessions across two agents; both pass after the fix with two resolutions and correct model results.
- After integrating main's catalog-metadata optimization (#137901), all 46 focused Linux tests passed across performance, search, model identity, ACP ownership, and plugin runtime. Both independently added regression groups remain intact. The configured-default resolver and our regression block are byte-identical to the reviewed original patch; the PR still changes only these two files.
- Five focused suites: 292 passed, one unchanged Windows path assertion failed. The same `listAgentsForGateway includes effective workspace + model for default agent` case fails on the frozen baseline: expected `/tmp/default-workspace`, received `C:\tmp\default-workspace`. Its source hash is identical across baseline and candidate.
- Formatting, diff sanity, and 15 changed-check guards passed. Independent review is scoped-clean through P2.
- Combined physical-install validation passed core and core-test typechecks. The full changed check exited with two untouched dead-export findings; their owners, real consumers, and scanner configuration match immutable baseline blobs; no fresh baseline Knip run was performed.
- Real Gateway baseline profiling places session model resolution below 1% of sampled time per resolver. Broader Gateway stalls are dominated by synchronous state-coordinator work and are not claimed solved by this patch. Real Gateway HTTP/WebSocket runs with a synthetic provider completed, but the two-fix comparisons were mixed and do not establish a session-specific transport speedup; the table above measures synthetic component execution, not network or model latency.

Related: #145579


steipete:
Merged via squash.

- Prepared head SHA: [fd2db0ba53346e60521f649de868f71be539db61](https://github.com/openclaw/openclaw/pull/145580/commits/fd2db0ba53346e60521f649de868f71be539db61)
- Landed commit: [ca96441ed94f0a1bed4c879dcb8f96a2ff793ee4](https://github.com/openclaw/openclaw/commit/ca96441ed94f0a1bed4c879dcb8f96a2ff793ee4)

PR #127158, opened 2026-08-21
fix(gateway): prune idle tool event recipients

steipete:
Closes #127157

## What Problem This Solves

An idle Gateway retained expired tool-event recipient records because cleanup only ran during later recipient-registry operations. Completed runs could keep their historical run and connection IDs after the existing grace period had elapsed.

## Why This Change Was Made

Expose the registry's existing expiration routine to the Gateway's existing 60-second maintenance sweep. Keep the next-expiry bound so maintenance and tool events avoid scanning before cleanup is due. Preserve the post-prune audience lookup and release only empty run records.

## User Impact

Expired recipients are reclaimed on the next maintenance tick after their existing 10-minute active TTL or 30-second finalized grace. Recent recipients, late-event grace, registered run contexts, and other retained run state keep their existing behavior. No per-run timers, configuration, storage, or permission changes are introduced.

## Evidence

- The maintenance regression failed against unchanged main because an expired active recipient remained after the first tick. Removing only the maintenance hook also makes the final dedicated regression fail on the same assertion.
- Normal compiled-runner proof: `node scripts/run-vitest.mjs src/gateway/server-chat-state.test.ts src/gateway/server-maintenance.test.ts src/gateway/server-maintenance.recipients.test.ts` passed all 74 tests across three files. The regression covers idle cleanup, finalized grace, subsequent expiration, recent recipients, and preserved registrations.
- Fresh independent Codex review of the final three-file scope: no actionable P0-P2 findings.
- Targeted lint, changed-file formatting, line-cap and other completed static guards passed. Local `check:changed` stopped at the core TypeScript graph boundary because the borrowed compiler installation is outside the task checkout; no guard or dependency installation was changed. Exact-head hosted CI completed the required gates for `a6005b81e6ccd95a9e6791abb76116005c221536`: [run 35177855253](https://github.com/openclaw/openclaw/actions/runs/35177855253).
- Production delta: +16/-11 lines, net +5. The 68-line regression uses the existing maintenance-state fixture in a dedicated sibling file because the original maintenance test file is already at its line cap; its existing coverage is unchanged.

Refreshed onto main `8dd36a07c761d08229937f32c99a5456facfab4e`, including the shared CI repairs in #150455. The complete three-file patch is byte-identical to the reviewed and tested change. The local results above precede this mechanical refresh; exact-head CI has now passed as linked above.

PR #145581, opened 2026-09-12
perf(gateway): avoid repeated tool recipient expiry scans

steipete:
Related: #127158

<details>

</details>

## What Problem This Solves

Resolves a problem where delivering live tool progress becomes increasingly expensive as the Gateway tracks more active runs, including runs without tool-event recipients.

## Why This Change Was Made

The recipient registry checks the earliest possible expiry before scanning run state. A due expiry still triggers a full sweep and recomputes that deadline, preserving the existing ten-minute idle lifetime and thirty-second finalized grace period. This reduces repeated work during active tool streams.

Related #127158 addresses pre-existing idle cleanup without subsequent recipient activity; this change addresses repeated scanning during activity and does not add a maintenance hook.

## User Impact

Busy Gateways spend less CPU delivering live tool updates while preserving recipient ordering, refresh behavior, and finalized expiry.

## Evidence

- Synthetic benchmark through the real `createAgentEventHandler`, with 10,000 tracked active runs: median processing time for 10,000 tool events fell from **924.26 ms to 10.10 ms** (91.5×). Each scenario verified all 51,000 deliveries, including 1,000 warmup events and five measured batches. Node v24.19.0 on Windows; baseline `47ce03dbc6994435af783fecf13d5586e7b307f4`. Measurement excludes network and provider latency; due-expiry sweeps remain linear in tracked runs.
- **255 tests passed** with `node scripts/run-vitest.mjs src/gateway/server-chat-state.test.ts src/gateway/server-chat.agent-events.test.ts`. Coverage includes expired run audiences with current session subscribers, idle refresh, clock rollback, state clearing, and live tool-event routing.
- Core production and core-test typechecks, formatting, `git diff --check`, and the preceding changed-file guards passed in an independently installed validation checkout. **The full combined changed gate failed** on Knip findings for `canonicalProviderName` and `listPackagedStaticExtensionAssetOutputs` in unrelated scripts. Their definitions, real script consumers, and scanner configuration match immutable baseline blobs; a fresh clean-baseline Knip run has not been performed. All nine remaining selected lint/state/import guards passed when run separately; the combined gate remains failed.
- Real Gateway HTTP/WebSocket workloads completed with a synthetic provider. The two-fix comparisons were mixed and retained large control-plane stalls; they do not establish an event-routing-specific network speedup.
- Fresh autoreview returned one P2 claiming missing recipient lookups previously always pruned. Two independent inspections rejected that attribution: the [immutable baseline already guards pruning with `if (entry)`](https://github.com/openclaw/openclaw/blob/47ce03dbc6994435af783fecf13d5586e7b307f4/src/gateway/server-chat-state.ts#L645-L652), and this patch retains that condition. The raw review remains a finding with a documented rejection.

Related: #145579

### DIFF
--- a/src/gateway/server-chat-state.test.ts
+++ b/src/gateway/server-chat-state.test.ts
@@ -32,6 +32,52 @@ describe("createChatRunState", () => {
     },
   );
 
+  it("expires idle recipients while activity extends another run's lifetime", () => {
+    let now = 1_000;
+    const clock = vi.spyOn(Date, "now").mockImplementation(() => now);
+    try {
+      const state = createChatRunState();
+      state.toolEventRecipients.add("idle", "conn-idle");
+      state.toolEventRecipients.add("active", "conn-active");
+      now = 301_000;
+      expect(state.toolEventRecipients.get("active")).toEqual(new Set(["conn-active"]));
+      now = 600_999;
+      state.toolEventRecipients.get("active");
+      expect(state.runs.has("idle")).toBe(true);
+      now += 1;
+      state.toolEventRecipients.get("active");
+      expect(state.runs.has("idle")).toBe(false);
+      expect(state.toolEventRecipients.get("active")).toEqual(new Set(["conn-active"]));
+    } finally {
+      clock.mockRestore();
+    }
+  });
+
+  it.each([false, true])(
+    "expires new recipients after clock rollback (clear previous state: %s)",
+    (clear) => {
+      let now = 1_000_000;
+      const clock = vi.spyOn(Date, "now").mockImplementation(() => now);
+      try {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [53a9c9a150b362a94d95d2c745fd481578a9c9ce](https://github.com/openclaw/openclaw/pull/145581/commits/53a9c9a150b362a94d95d2c745fd481578a9c9ce)
- Landed commit: [f7306cc5fbfdf6312c71f37d77db06db532c5743](https://github.com/openclaw/openclaw/commit/f7306cc5fbfdf6312c71f37d77db06db532c5743)

PR #145582, opened 2026-09-12
perf(state): reuse paths during coordinator acquisition

steipete:
<details>

</details>

## What Problem This Solves

Resolves a problem where heavy session activity adds avoidable synchronous work to Gateway state writes, delaying session lists and other control requests.

## Why This Change Was Made

Resolve the database and runtime paths once within each ordinary coordinator acquisition, then derive both existing lock identities from those prepared facts. Explicit coordinator paths retain precedence, and acquisitions with an authority callback still resolve paths again after that callback. Lock identities, lifetimes, permission checks, and transaction boundaries stay unchanged; there is no persistent cache or storage change.

## User Impact

State writes spend less time on repeated filesystem probes. This reduces one measured contributor to Gateway stalls; substantial latency remains under the stress workload below.

## Evidence

Baseline `47ce03dbc6994435af783fecf13d5586e7b307f4`, Windows, Node 24.19.0. A real Gateway profile with 1,000 sessions attributed 19.62 seconds inclusive of a 73.69-second sample to ancestor path resolution, overlapping with 40.97 seconds in shared-state writes. Workspace-memory lock registration and release repeatedly reach this path. These samples include time inside synchronous native calls, not exclusively CPU computation.

Synthetic component benchmark through the actual acquisition entry point, seven alternating passes of 100 operations per arm, with two warmups:

| Acquisition | Median before | Median after |
| --- | ---: | ---: |
| Ordinary acquire/release | 1.576 ms | 1.159 ms |
| Nested acquire/release | 1.138 ms | 0.486 ms |

Ordinary filesystem probes fall from eight to four `existsSync` calls and four to two `realpathSync` calls. Directory creation and permission-check calls remain unchanged.

One isolated before/after pair through real HTTP/WebSocket transport used 1,000 sessions, eight requested turns, 100 patches, and a synthetic provider. Only this production patch changed between runs; both included the same two preceding Gateway fixes:

| Metric | Before | After |
| --- | ---: | ---: |
| Load completion | 40.95 s | 30.96 s |
| Session list p95 | 19.23 s | 10.09 s |
| History p95 | 3.27 s | 2.65 s |
| Patch p95 | 2.63 s | 1.86 s |

Both runs had zero transport probe failures. This is supportive single-pair evidence, not a general speedup guarantee: dynamic background provider requests differed (21 versus 15), and maximum event-loop delay increased from 1.50 to 2.09 seconds. Windows teardown forced Gateway exit 1 after completed load; all owned processes were verified stopped. No real OpenAI call was made.

- The new work-bound regression fails on the original code with four realpath probes instead of two and passes after the change. Preservation cases cover explicit paths, changing database paths between acquisitions, and path changes inside the authority callback.
- Focused checks: 25 tests passed, two existing device-identity contract cases skipped on Windows, and two source-exclusion cases failed. Both failures reproduced identically against the original coordinator with complete dependencies: concurrent non-owner reads and inherited read-scope expiry returned a copy rather than rejecting. No assertions were weakened.
- The first grouped test invocation encountered incomplete package dependencies and an automatic worker-build timeout/retry. Dependency links were restored before the independent baseline comparison, which completed without dependency warnings or retries.
- Core and core-test typechecks passed. The original combined changed gate has two findings in untouched exports; their owners, consumers, and scanner configuration match baseline blobs, but no fresh baseline Knip run was performed. Formatting and diff sanity passed.
- Final targeted lint, formatting, and diff sanity passed after the braces-only test correction. Fresh independent review of the exact patch was scoped-clean through P2. Repeated paired live comparisons have not run.

Related: #145579

### DIFF
--- a/src/infra/state-database-coordinator.test.ts
+++ b/src/infra/state-database-coordinator.test.ts
@@ -1,16 +1,86 @@
+import fsSync from "node:fs";
 import fs from "node:fs/promises";
 import path from "node:path";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
 import {
   acquireGatewayLifecycleCoordinator,
   acquireStateDatabaseCoordinator,
+  acquireStateDatabaseHandleExclusion,
+  resolveStateDatabaseCoordinatorPath,
   withStateSchemaFence,
 } from "./state-database-coordinator.js";
 
 const tempDirs = useAutoCleanupTempDirTracker(afterEach);
 
 describe("state database coordinator", () => {
+  it.each([false, true])(
+    "bounds filesystem path probes per acquisition with explicit coordinator path %s",
+    (explicit) => {
+      const root = tempDirs.make("openclaw-state-coordinator-path-work-");
+      const params = {
+        databasePath: path.join(root, "state", "openclaw.sqlite"),
+        runtimeDirectory: root,
+        uid: typeof process.getuid === "function" ? process.getuid() : undefined,
+        coordinatorPath: explicit ? path.join(root, "custom", "coordinator.sqlite") : undefined,
+      };
+      const resolvePath = vi.spyOn(fsSync, "realpathSync");
+      try {
+        for (let attempt = 0; attempt < 2; attempt++) {
+          params.databasePath = path.join(root, `state-${attempt}`, "openclaw.sqlite");
+          const expectedPath =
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [738f76641fcf41c50fb87d26e135e027a1646d54](https://github.com/openclaw/openclaw/pull/145582/commits/738f76641fcf41c50fb87d26e135e027a1646d54)
- Landed commit: [bffb5c0e129c0baa29c7fdaf5a5e17a97b555fdd](https://github.com/openclaw/openclaw/commit/bffb5c0e129c0baa29c7fdaf5a5e17a97b555fdd)

PR #145591, opened 2026-09-12
fix(gateway): avoid loading inactive worker history

steipete:
## What Problem This Solves

Fixes unnecessary Gateway memory use when cloud-worker turns begin in sessions with inactive transcript branches. Preparing a small active conversation loaded and retained the physical transcript's inactive message payloads while credentials, workspace media, and skills were prepared.

## Why This Change Was Made

Use the existing selected model-context readers for recorder-owned or already-persisted input. A pending recorder keeps the synchronous read-before-persist ordering, so its own append cannot invalidate an asynchronous snapshot. Already-persisted input uses the asynchronous reader and revalidates the caller, cancellation, exact placement claim, and current transcript identity after the read.

Unrecorded input keeps its writable SessionManager, preserving captured append ancestry, conflict handling, and writer custody. The completion reader and recorder-race fallback remain unchanged. No schema, retention, or transcript-commit recovery contract changes.

## User Impact

Worker turns stop retaining inactive message payloads during this preparation phase while keeping the same active history and current-input/base-leaf behavior. This is a bounded allocation repair, not a claim that all Gateway memory pressure or stalls are resolved.

## Evidence

Both permanent 64-row tests fail on the unmodified producer because it hydrates 64 inactive 1 KiB messages, then pass with zero inactive payloads. The active conversation remains four messages for already-persisted input or three before recorder-owned persistence. The earlier 512-row diagnostic measured roughly 614 KB of serialized loaded payload versus under 600 bytes for the selected reader; that is a payload proxy, not a retained-heap or RSS measurement.

All 10 focused tests pass on the current candidate. They cover input/history/base-leaf parity, split logical/append leaves, canonical persistence already in flight or begun after snapshot acquisition, and cancellation, claim loss, caller revocation, and session rebinding during the asynchronous read. Existing launcher/media/handoff and SessionManager owner suites previously passed 124 tests on the integration base.

Independent P0-P2 review is clean. The revised [Blacksmith wire proof](https://github.com/openclaw/openclaw/actions/runs/34672349693) passed two authenticated chat turns, history-prefix/input/base-leaf checks, completed supervisor receipts, and joined cleanup; the owned lease is completed. The native synced proof verified both producer hashes, the harness, tests, and lockfile before its build (Node 24.19.0). Its delegated checkout metadata follows the hydrated main; exact-head CI is tracked separately on this PR.

The campaign wire harness remains private rather than duplicating the maintained service suite. It uses a loopback mock provider and a separate authenticated operator.read/operator.write client; it does not establish real-provider behavior or multi-profile membership semantics. The final native changed checks passed, including typechecking, lint, dead-code and repository guards.

Exact-head CI passed after one qualified rerun of the unchanged QA gallery shard. A temporary path containing `gif` deterministically reproduces its log-as-image classification failure; the original CI failure is retained and the unrelated gallery repair is tracked separately.

### DIFF
--- a/src/gateway/worker-environments/worker-turn-detached-context.test.ts
+++ b/src/gateway/worker-environments/worker-turn-detached-context.test.ts
@@ -1,0 +1,381 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { SessionPlacementTurnParams } from "../../agents/session-placement-admission.js";
+import { SessionManager } from "../../agents/sessions/session-manager.js";
+import {
+  makeAgentAssistantMessage,
+  makeAgentUserMessage,
+} from "../../agents/test-helpers/agent-message-fixtures.js";
+import { upsertSessionEntryCore } from "../../config/sessions/session-accessor.js";
+import { withTimeout } from "../../infra/fs-safe.js";
+import { WorkerTaskPool } from "../../infra/worker-task-pool.js";
+import { createUserTurnTranscriptRecorder } from "../../sessions/user-turn-transcript.js";
+import { projectWorkerSessionTurnClaim } from "./placement-record.js";
+import { WorkerRunnerUnavailableError, type WorkerTurnTunnelHandle } from "./tunnel-contract.js";
+import { releaseClaimIfOwned } from "./worker-turn-admission.js";
+import {
+  ENVIRONMENT_ID,
+  OWNER_EPOCH,
+  SESSION_ID,
+  SESSION_KEY,
+  attachedEnvironment,
+  cleanupWorkerTurnLauncherTest,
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8bc13c1cc1a859f677fb9cf0e483439c347abe07](https://github.com/openclaw/openclaw/pull/145591/commits/8bc13c1cc1a859f677fb9cf0e483439c347abe07)
- Landed commit: [aa7eabb1f6d60be17bdb9f86b416f1abbe2fa48f](https://github.com/openclaw/openclaw/commit/aa7eabb1f6d60be17bdb9f86b416f1abbe2fa48f)

PR #145596, opened 2026-09-12
refactor: streamline updater test setup and coverage

steipete:
## What Problem This Solves

Updater tests repeated an installed-service refresh scenario already covered by stronger assertions, and a local Commander self-test bypassed the real OpenClaw registration. Legacy finalization also loaded source TypeScript in fresh native receiver processes despite existing prepared runtime entries.

## Why This Change Was Made

Reuse the existing native receiver and native execution descriptors from the same prepared graph, preserving their shared service-authority scope. Standalone/watch retains the original source loader; the finalizer's source loader and fixture hook remain unchanged.

Delete the weaker duplicate service-refresh row and the toy parser case. The stronger installed-service assertions and real registered `update status --json` forwarding coverage remain, along with the distinct failure, no-restart, and no-success cases.

## User Impact

No production behavior changes. This removes redundant test work and reuses the existing receiver runtime without changing compiler inventory, authority checks, mocks, state isolation, or deadlines. Production LOC delta: 0; tests: 7 added / 31 removed (net −24).

## Evidence

- All ten original legacy-finalization scenarios passed before and after the receiver change in their original order, with matching inventory/status digests and joined process cleanup.
- All four retained service-refresh rows and the complete 73-case real-registration option-collision suite passed in one targeted run.
- Source-mode resolution produces the original receiver/exec module URLs and retains the original bootstrap.
- Final combined scoped checks passed; independent P2 review completed both passes with no actionable findings.

The receiver-only local pair measured 140.194s → 133.082s wrapper time and 127.386s → 122.869s summed case time. Some receiver cases improved, but unchanged incumbent controls also slowed substantially, so this is a modest observation rather than a stable speedup claim. The two deleted cells accounted for 1.731s of summed case time in the retained main CI log; that is not an end-to-end saving measurement. No whole-shard or eight-minute CI claim is made.

### DIFF
--- a/src/cli/update-cli.test.ts
+++ b/src/cli/update-cli.test.ts
@@ -6803,24 +6803,6 @@ describe("update-cli", () => {
     );
   });
 
-  it("parses update status --json as the subcommand option", async () => {
-    const program = new Command();
-    program.name("openclaw");
-    program.enablePositionalOptions();
-    let seenJson = false;
-    const update = program.command("update").option("--json", "", false);
-    update
-      .command("status")
-      .option("--json", "", false)
-      .action((opts) => {
-        seenJson = Boolean(opts.json);
-      });
-
-    await program.parseAsync(["node", "openclaw", "update", "status", "--json"]);
-
-    expect(seenJson).toBe(true);
-  });
-
   it.each([
     {
       name: "defaults to dev channel for git installs when unset",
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [402d2f55db305cf492aeee0ec6af32cdc9550444](https://github.com/openclaw/openclaw/pull/145596/commits/402d2f55db305cf492aeee0ec6af32cdc9550444)
- Landed commit: [ffcc6910d9ad0c65630d6c8c9e55438e8dba7e2e](https://github.com/openclaw/openclaw/commit/ffcc6910d9ad0c65630d6c8c9e55438e8dba7e2e)

Issue #136203, opened 2026-09-02
[Bug]: Windows de-DE 2026.8.2 upgrade leaves Doctor maintenance blocked and legacy workspace state behind


hermannmetzker-ai:
## Summary

Upgrading an existing multi-agent Windows installation from `2026.7.1-2` to `2026.8.2 (0965053)` completed the npm package update, but the post-update recovery path required several manual interventions before the Gateway, memory search, and ambient ownership were healthy again.

The most important Windows-specific failure was that `doctor --fix` could not reliably verify the Scheduled Task runtime/ownership on a German Windows locale (`state=Bereit`). This blocked maintenance even when the Gateway was not actually running. After that was worked around, the 2026.8.x migration still left legacy workspace-state files in place; on the next Gateway restart, startup failed hard with `Legacy workspace setup state requires migration` until those files were manually moved out of the active workspace.

No user data was lost, but the documented recovery command (`openclaw doctor --fix`) did not converge on its own.

## Environment

- OpenClaw: `2026.8.2 (0965053)`
- Upgraded from: `2026.7.1-2`
- OS: Windows `10.0.26200` x64 (German locale)
- Node.js: `24.18.0`
- Install: global npm
- Gateway service: Windows Scheduled Task `OpenClaw Gateway`
- Gateway bind: loopback `127.0.0.1:18789`
- Multi-agent setup: `main`, `planner`, `executor`, `validator`
- Memory provider: Ollama / `nomic-embed-text`

### 1. Update succeeded, Doctor maintenance did not

The package update itself completed, but the Doctor phase repeatedly failed around Gateway service ownership / maintenance verification.

The Scheduled Task was present and later showed the localized German task state:

```text
state Bereit
```

At points, `openclaw gateway status --deep` reported runtime status as unknown because no numeric Last Run Result could be derived reliably from the localized Scheduled Task output.

This matches the same general failure family as #136123, but this reproduction is on a German Windows locale and `2026.8.2`.

### 2. Multi-agent migration changed effective `main` workspace resolution

Before the update, the default workspace was:

```text
C:\Users\<user>\.openclaw\workspace
```

After migration to explicit multi-agent ownership, `agents.entries.main` was empty and the effective `main` workspace resolved as:

```text
~\.openclaw\workspace\main
```

while the actual memory corpus remained under:

```text
~\.openclaw\workspace\memory
```

Result:

```text
Memory Search (main)
Indexed: 0/0 files · 0 chunks
[...]

PR #145597, opened 2026-09-12
fix(daemon): retain authenticated systemd manager inspection

fuller-stack-dev:
Related: #136203

## What Problem This Solves

Fixes an issue where Linux service inspection loses the effective command and runtime when the selected user manager remains alive but its session-bus broker disappears. The manager's private socket is still available, but the existing strict readers cannot use it.

## Why This Change Was Made

Authenticate the originally selected local manager while its broker is available, retain its private connection within the existing service-operation lifetime, and use the platform libsystemd typed-read API through the existing Koffi dependency. Revalidate kernel UID/PID and process-start identity; reject a different or replaced manager. Keep this passive binding separate from load/mutation authority.

No dependency, configuration, broker activation, D-Bus protocol implementation, custom-route replacement or local-file-as-effective fallback is added. Unsupported/non-local transports keep their existing adapter. This does not choose an initial manager when its identity was never established, alter later lifecycle-effect routing, or close the independent symptoms in #136203.

## User Impact

Strict command, runtime and definition-capability inspection can continue against the same authenticated manager after broker loss. Authored environment values, drop-ins, environment files and unset directives remain intact. Another same-UID manager is not adopted as recovery.

## Evidence

- Preserved native RED on unmodified `a9ff79407881ec0dd084cdb7d4ac2eeb2719b7e1`: systemctl still reached the original private socket after broker loss, while strict command and loaded-runtime inspection failed. No historical254-test baseline replay.
- Product-source native GREEN on isolated Ubuntu/systemd255: effective command/runtime and writable definition capability survived broker loss on the original manager; a different same-UID private peer, unavailable authored broker, conflicting route and replaced original manager were refused by product code. Passive LoadUnit and a retired inherited update grant stayed denied. Invalid-property recovery, queued deadlines and a bounded stalled handshake passed. Both task-only manager fixtures were cleaned up.
- Source-native proof uses a synthetic sleep service, not an installed updater/first-hop driver. It does not establish original-reporter attribution or general XRDP/rollback recovery.
- Targeted tests:75 scope/runtime/queue/admission/authority cases,53 strict-command cases,26 state-composition cases and84 definition cases pass (238 distinct selected cases; unrelated name-filtered cases were not run). Reused unchanged-source results are recorded separately from the final64-case follow-up.
- Initial service-owner review identified six admission/scheduling defects; all were addressed with focused coverage. Repository-required Autoreview completed its single P0-only pass without findings. Final service-owner source review is clean for all17 candidate files; the last guard regression, full core-test typechecks, dead-export scans, changed-file lint and formatting passed. Exact-head CI remains a separate landing gate.

- Hosted Linux CI exposed a subprocess-only diagnostic fixture admitting the runner's real manager. An isolated native trace confirmed the cause. The test-only follow-up models the unavailable native broker, preserves every existing assertion, and passes all5 Doctor diagnostic cases plus type/lint checks on Linux. All previously reviewed production bytes are unchanged; successor exact-head CI remains required.

- Prerequisite qualification: successor CI on `11ed727b4bd3` failed only the inherited line cap in `src/auto-reply/thinking.test.ts`, whose blob was identical on the original base and PR head. Merged upstream PR #145632 fixes that lint input. GitHub-signed head `6b6094d296cc` incorporates main and preserves all 18 reviewed file hashes and the original diagnostic assertions. The 238 selected cases and native before/after proof are retained without replay; successor exact-head CI is run `34678946551`. Published-updater and full-lifecycle coverage remain explicitly unproved.

Release-note context: keep Linux service inspection on the authenticated original manager when its session broker is lost during the same operation. No initial-selection or complete-upgrade guarantee.


fuller-stack-dev:
Merged via squash.

- Prepared head SHA: [6b6094d296cc7319c7ff482d4c13f22ca09d3a9c](https://github.com/openclaw/openclaw/pull/145597/commits/6b6094d296cc7319c7ff482d4c13f22ca09d3a9c)
- Landed commit: [cfa99bb981f0cfdd9be9427b85bc08e4f0eab8d4](https://github.com/openclaw/openclaw/commit/cfa99bb981f0cfdd9be9427b85bc08e4f0eab8d4)

PR #145612, opened 2026-09-12
fix(qa-lab): keep log previews when labels mention GIF

steipete:
Related: #145591

## What Problem This Solves

Fixes an issue where QA Lab hides a log artifact's text preview when its free-form kind label contains a media hint such as `gif`. The same precedence error can show JSON as video or a video as an image. It also makes the existing path-redaction test fail when a temporary directory name happens to contain a matching hint.

## Why This Change Was Made

Recognized filename suffixes now determine the gallery media type before kind hints are considered, consistent with the artifact download endpoint. The complete kind remains free-form and preserved; the existing kind fallback and its ordering still apply to unknown suffixes and extensionless artifacts. Format support, path checks and preview byte limits are unchanged.

## User Impact

Log and JSON previews remain readable despite incidental media words in their metadata, and image/video files keep their correct presentation.

## Evidence

- The unchanged source reproduced the reported failure deterministically under a temporary parent containing `gif`. The original CI log does not retain its exact random directory suffix, so this is a matching reproduced failure shape rather than a claim about those missing bytes.
- Four deterministic gallery cases failed before the repair; all 37 gallery/summary tests pass afterward. Coverage retains arbitrary compound kind labels, fallback ordering, extensionless and unknown-suffix artifacts, and uppercase suffixes.
- The existing real HTTP gallery/GET/HEAD test also failed before the repair and passes afterward, including correct text preview, content type and streamed bytes.
- The complete changed gate and independent Codex review through P2 pass.
- Attached before/after captures use synthetic files through the actual gallery model and unchanged renderer/CSS. The baseline defers the `.log` as media; the candidate displays its exact text. Both complete captures were inspected for synthetic-only content.

The original #145591 CI failure and qualified unchanged-head rerun remain separate evidence; the rerun did not repair this classifier.

Before: the log is treated as deferred media

After: the exact log text is available in the gallery

### DIFF
--- a/docs/concepts/qa-e2e-automation/qa-reporting.md
+++ b/docs/concepts/qa-e2e-automation/qa-reporting.md
@@ -41,6 +41,11 @@ producer `qa-evidence.json`. When `qa suite` is reached through `qa run
 --qa-profile`, the same `qa-evidence.json` also includes the profile
 scorecard summary for the selected taxonomy categories.
 
+The QA Lab evidence gallery uses recognized file suffixes to select image,
+video, JSON, or text previews. A `.log` file remains text even when its free-form
+artifact kind contains a media hint such as `gif`. Kind hints still classify
+extensionless files and unknown suffixes; the complete kind label is preserved.
+
 `qa confidence-report` keeps `productImpact` and `qaImpact` annotations in their
 own Markdown table cells, collapsing whitespace for display. The JSON summary
 preserves the annotation values, including internal line breaks.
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [2b6907f8989cc200335afeb29fab4dc24f6528ee](https://github.com/openclaw/openclaw/pull/145612/commits/2b6907f8989cc200335afeb29fab4dc24f6528ee)
- Landed commit: [785266d3af19b65ad3c2d4fb5f8d9033195c3137](https://github.com/openclaw/openclaw/commit/785266d3af19b65ad3c2d4fb5f8d9033195c3137)

PR #145616, opened 2026-09-12
fix(sessions): reduce allocation spikes during disk cleanup

steipete:
## What Problem This Solves

Fixes an issue where disk-pressure cleanup copies every session's metadata into the Gateway heap while protecting only a few active sessions. It also enumerates historical candidates after archive pruning has already cleared disk pressure.

## Why This Change Was Made

The existing protection owner first reads and normalizes lightweight keys, then hydrates full metadata only for matching stored keys. It captures the stored key bytes as hex so raw SQLite text never needs to round-trip through a Node TEXT binding. A SQLite-only key subquery uses the existing covering index before indexed payload reads. Historical-generation protection is streamed through the same owner. Candidate enumeration runs only while physical usage still exceeds the existing high-water target.

Key normalization, full entry parsing, admission and post-archive rechecks, integrity checks, retention, transactions, and worker ownership keep their existing contracts. This changes read projection and planning only; it adds no cache, schema, index, configuration, or worker lifetime.

## User Impact

Cleanup performs less synchronous work and creates far less temporary heap pressure while preserving active sessions and their history.

## Evidence

- The real SQLite/admission regression failed with 106 payload reads for five admitted entries and now passes with five. It covers Unicode folding, opaque Signal peer case, embedded NUL, malformed entries, orphan historical generations, and admission release.
- Raw BLOB-to-TEXT fixtures cover UTF-8 and both UTF-16 byte orders, including U+FFFE, U+FFFF, unmatched surrogates, and NUL. They pass with the original full-scan owner, fail for both UTF-16 encodings with decoded-key rebinding, and pass with the repaired byte-preserving selection. The earlier parameter-insertion probes did not cover this case because text was normalized before storage.
- Existing real archive-first cleanup cases failed with one unnecessary reference scan after pressure cleared and now pass with zero, preserving the same removed archive and searchable history.
- All 67 tests pass across historical eviction, admission, protected cancellation, and budget ownership suites, including real worker and external-connection races.
- Synthetic owner benchmark on Node 26.8.2: 3,000 rows containing 98.6 MB of JSON, one admitted session, two warm-ups and ten measured calls. Mean lookup time changed from 48.45 ms to 4.13 ms; immediate heap growth changed from about 101.8 MB to 2.12 MB. This measures an in-memory SQLite projection, not live request latency, retained heap, or an OOM cure.
- SQLite query-plan inspection confirms a covering scan of the existing primary-key index and indexed payload lookups; no new index is required.

- Complete changed-file checks passed, including core and all test typechecks, formatting, lint, dead-export scans, and storage/runtime/import guards.
- Fresh independent P2 review of the final staged candidate completed with no actionable findings.

The repaired candidate is `d3427763c41e7b5a75b133b3bd75507567a231a6`, based on `9ab6ba866c9bb1ee6cdb74cdc99173833998dac3`. Its local proof includes the raw-encoding correction. Exact-head [CI run 34676752315](https://github.com/openclaw/openclaw/actions/runs/34676752315) completed successfully: 131 jobs finished with success or intentional skip, and the aggregate gate passed. The fresh ClawSweeper review of this head also completed with no findings or pre-merge actions.

### DIFF
--- a/src/config/sessions/session-history-admission-protection.test.ts
+++ b/src/config/sessions/session-history-admission-protection.test.ts
@@ -1,0 +1,82 @@
+import { DatabaseSync } from "node:sqlite";
+import { expect, it, onTestFinished } from "vitest";
+import { beginSessionWorkAdmission } from "../../sessions/session-lifecycle-admission.js";
+import { collectAdmissionProtectedSessionIds } from "./session-history-eviction.js";
+
+it("reads only admitted entry payloads while protecting normalized keys and every generation", async () => {
+  const db = new DatabaseSync(":memory:");
+  onTestFinished(() => db.close());
+  let payloadReads = 0;
+  db.function("read_payload", (value) => {
+    payloadReads++;
+    return value;
+  });
+  db.exec(`
+    CREATE TABLE stored_nodes (session_key TEXT PRIMARY KEY, current_session_id TEXT, entry_json TEXT);
+    CREATE VIEW session_nodes AS SELECT session_key, current_session_id, read_payload(entry_json) AS entry_json FROM stored_nodes;
+    CREATE TABLE session_windows (session_id TEXT PRIMARY KEY, session_key TEXT);
+  `);
+  const insert = db.prepare("INSERT INTO stored_nodes VALUES (?, ?, ?)");
+  const addWindow = db.prepare("INSERT INTO session_windows VALUES (?, ?)");
+  const entries = [
+    { key: "agent:main:dashboard:tracked", identity: "AGENT:MAIN:DASHBOARD:TRACKED" },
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d3427763c41e7b5a75b133b3bd75507567a231a6](https://github.com/openclaw/openclaw/pull/145616/commits/d3427763c41e7b5a75b133b3bd75507567a231a6)
- Landed commit: [54ebdbbfd2a141f7ffcf220aa4ab3155dd5cf7fa](https://github.com/openclaw/openclaw/commit/54ebdbbfd2a141f7ffcf220aa4ab3155dd5cf7fa)

Issue #145608, opened 2026-09-12
Busy Gateway allocates excessively while preparing parallel agent turns


steipete:
## What happened

A busy Gateway with 32 concurrent tool-using sessions allocates about 10.8 GiB during an approximately 17-second workload. Session-list and history requests still complete, but individual requests can exceed two seconds.

This was reproduced on b4c35afcc53 with the existing synthetic Gateway concurrency benchmark: 128 seeded sessions, 16 observers, two history clients, 128 session updates, and control-plane probes. Three allocation samples were consistent. A separate live OpenAI check completed 32 file-processing tasks across 16 concurrent sessions.

## Allocation evidence

Load-window V8 sampling, including objects collected by both minor and major GC, identifies two dominant sources:

- Plugin value classification eagerly traverses large metadata graphs and materializes collection entries before discovering callable members that already require an instance view.
- Provider public-surface lookup repeatedly normalizes and compares every plugin root against the same module paths.

## Requested outcome

Retain reproducible load-phase allocation profiling in the existing benchmark and reduce these allocations while preserving native data compatibility, current registry/source ownership, and plugin call admission. Verify the improvements using matching before/after workloads and live parallel tasks.

steipete:
Implemented and landed in #145629 (9449a1a780e410ed8ed485155a7ef18985b0270e) and #145630 (6b62140fbacfc75bb1c2535a813b11affa955201).

The existing concurrency benchmark now captures load-window V8 allocations, including garbage-collected objects. Matched 32-session workloads reduced mean sampled allocations from 10.81 GiB to 7.23 GiB (33%); all 64/64 turns completed per revision with no failed probes. Separate live OpenAI checks passed 32/32 file-processing tasks before and after across 16 concurrent sessions, with correct output files and completion markers.

Focused tests, builds, selected checks, and hosted CI passed. The initial CI lint issue was fixed upstream in #145632; the unrelated isolated profiling-CLI help timeout did not recur in the refreshed run. Tail latency remains variable, as documented in #145630.

PR #145629, opened 2026-09-12
feat(gateway): profile allocations during concurrent workloads

steipete:
Related: #145608

## What Problem This Solves

Busy Gateway benchmarks report retained heap and RSS, which hide short-lived allocations that have already been garbage-collected. Startup allocations also obscure the work performed by concurrent sessions.

## Why This Change Was Made

Add `--heap-prof-dir` to the existing Gateway concurrency benchmark. A private child-process preload starts V8 sampling after fixture setup and stops after every load operation, including the fresh-connection probe. Sampling includes objects collected by minor and major GC. The report retains the raw profile, sampled bytes per run and turn, and the largest allocation stacks.

The instrumentation stays in repository tooling. It exposes no Gateway RPC or inspector listener, skips worker threads, and does not keep the Gateway process alive.

## User Impact

Developers can attribute allocation churn to the actual parallel-session workload and compare fixes using the existing benchmark options. Normal Gateway execution is unchanged.

## Evidence

- Benchmark and sibling lifecycle tests: 91 passed, including a real V8 allocation/GC test that excludes startup allocations and retains collected load objects.
- Three completed 32-session synthetic runs with tool execution, 128 seeded sessions, 16 observers, history clients, session updates, and control-plane probes. All turns and probes completed; profiles consistently identified approximately 10.8 GiB of load allocations in the existing runtime.
- Independent review identified an incomplete handshake barrier and an IPC teardown fallback; both are fixed. The teardown regression fails without IPC release and passes with it.
- Full build, typechecks, and selected lint/ownership/graph checks passed.

## Data compatibility

The new JSON writes produce explicit benchmark artifacts in the operator-selected `--heap-prof-dir`, using a fresh UUID filename for each run. They contain V8's sampling profile and are read only by the benchmark's summary step. Gateway databases, schemas, configuration, saved sessions, and upgrade/migration paths are untouched. Normal runs without the flag do not create these files. There is no stored application-data migration to prove; the existing Node/V8 profile contract and disabled/enabled benchmark tests cover the artifact boundary.

Initial merged-base CI surfaced the unrelated `src/auto-reply/thinking.test.ts` line cap, already being repaired by #145632, plus a timeout in an unchanged isolated `run-vitest-profile` help case. The latter did not execute this benchmark or preload; its precise stalled phase is not available in the CI log.

### DIFF
--- a/docs/reference/test/performance.md
+++ b/docs/reference/test/performance.md
@@ -98,6 +98,36 @@ Use JSON output or `--output` when comparing changes. Use `--cpu-prof-dir` only 
 
 </Accordion>
 
+<Accordion title="Gateway concurrency (scripts/bench-gateway-concurrency.ts)">
+
+Runs synthetic streaming agent turns in parallel sessions on one isolated
+Gateway. Add tool calls, session history, observers, and control-plane probes to
+reproduce allocation pressure from a busy Gateway. Build with `pnpm build`
+first; no provider key is required.
+
+```bash
+pnpm test:gateway:concurrency -- --concurrency 16 --tool-events --workspace-fanout --session-count 100 --history-messages 20 --history-clients 4 --subscribers 4 --visible-observer --control-plane --heap-prof-dir .artifacts/gateway-heap --output .artifacts/gateway-concurrency.json
+```
+
+`--heap-prof-dir` samples allocations in the Gateway's main V8 isolate, starting
+after startup, session seeding, and probe warmup. Sampling ends after the load
+and its final memory probe, before profile serialization and teardown. It uses
+a 32 KiB sampling interval and includes objects collected by both minor and
+major GC, so `sampledAllocatedBytes` estimates gross allocations rather than
+retained heap. Native allocations and separate worker isolates are outside this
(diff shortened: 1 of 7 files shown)


steipete:
The data-model compatibility item is a classification false positive. These writes create explicit, opt-in V8 benchmark artifacts under the requested output directory, with fresh filenames per run. They do not read or migrate an existing application datastore. Gateway SQLite schemas, configuration, saved sessions, and upgrade paths are untouched; disabled/enabled behavior and the V8 profile contract are covered by the tests and actual Gateway runs.

I added this boundary and the separate CI findings to the PR body. There is no application-data migration applicable to this change.

@clawsweeper re-review

steipete:
Merged via squash.

- Prepared head SHA: [06caf13c48c2a02a2c096452ddc7efe5f184716b](https://github.com/openclaw/openclaw/pull/145629/commits/06caf13c48c2a02a2c096452ddc7efe5f184716b)
- Landed commit: [9449a1a780e410ed8ed485155a7ef18985b0270e](https://github.com/openclaw/openclaw/commit/9449a1a780e410ed8ed485155a7ef18985b0270e)

PR #145630, opened 2026-09-12
fix(gateway): reduce excessive allocations during parallel agent turns

### LINKED ISSUE #145608: Busy Gateway allocates excessively while preparing parallel agent turns

steipete:
Closes #145608

## What Problem This Solves

Fixes excessive allocation churn while a busy Gateway prepares and executes parallel agent turns. Plugin metadata inspection walks large graphs repeatedly, and provider lookup repeatedly normalizes unrelated plugin roots.

## Why This Change Was Made

Plugin data classification checks direct callable/accessor members before descending into large data graphs, compares same-realm native constructors by identity, and streams Map/Set members instead of materializing their entire entry lists. Classification still reads current values; it adds no data-result cache.

Runtime module lookup reuses normalized absolute POSIX root prefixes within the existing PluginCache, keyed weakly by record identity and refreshed if the root changes. Current registry membership, instance source membership, and plugin-id disambiguation remain live. Windows and relative paths retain the existing path guard. The lookup no longer creates intermediate owner arrays.

## User Impact

Concurrent agent work creates less garbage while preserving plugin data compatibility and lifecycle admission. No configuration, storage, or protocol changes are required.

## Evidence

Matched built-Gateway runs, two samples per revision, used 32 concurrent tool-using sessions, 128 seeded sessions, 16 observers, two history clients, 128 session updates, and control-plane probes. V8 sampling included allocations collected by minor and major GC.

| --- | ---: | ---: |
| Mean sampled load allocations | 10.81 GiB | 7.23 GiB |
| Mean load duration | 17.07 s | 15.46 s |
| Peak Gateway RSS | 1142.7 MiB | 1087.5 MiB |
| Completed turns | 64/64 | 64/64 |
| Failed control/history/update probes | 0 | 0 |

Sampled allocations fell 33%. Tail latency is still variable: session-list p95 remained approximately 1.2 seconds, so this change does not claim to eliminate latency spikes. Before revision: b4c35afcc53. The profiling tooling is in #145629.

- Live OpenAI proof passed before and after: 32 file-processing tasks across 16 concurrent sessions on one isolated Gateway, with file contents and completion markers verified. Both Gateways exited cleanly. Live provider timing is correctness evidence only.
- Plugin-instance tests: 277 passed; ownership/public-surface tests: 34 passed.
- Full build and independent P0–P2 review passed.

### DIFF
--- a/docs/reference/test/performance.md
+++ b/docs/reference/test/performance.md
@@ -98,6 +98,36 @@ Use JSON output or `--output` when comparing changes. Use `--cpu-prof-dir` only 
 
 </Accordion>
 
+<Accordion title="Gateway concurrency (scripts/bench-gateway-concurrency.ts)">
+
+Runs synthetic streaming agent turns in parallel sessions on one isolated
+Gateway. Add tool calls, session history, observers, and control-plane probes to
+reproduce allocation pressure from a busy Gateway. Build with `pnpm build`
+first; no provider key is required.
+
+```bash
+pnpm test:gateway:concurrency -- --concurrency 16 --tool-events --workspace-fanout --session-count 100 --history-messages 20 --history-clients 4 --subscribers 4 --visible-observer --control-plane --heap-prof-dir .artifacts/gateway-heap --output .artifacts/gateway-concurrency.json
+```
+
+`--heap-prof-dir` samples allocations in the Gateway's main V8 isolate, starting
+after startup, session seeding, and probe warmup. Sampling ends after the load
+and its final memory probe, before profile serialization and teardown. It uses
+a 32 KiB sampling interval and includes objects collected by both minor and
+major GC, so `sampledAllocatedBytes` estimates gross allocations rather than
+retained heap. Native allocations and separate worker isolates are outside this
+profile. Each run records its `.heapprofile` path and the twenty largest
+allocation stacks; open the raw file in the Chrome DevTools Memory panel.
+
+The summary includes sampled allocation bytes per run and per completed turn.
+The per-turn figure also includes concurrent probes and session mutations;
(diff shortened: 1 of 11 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [689b006743a90d57fd6d289d63c14b524da8006a](https://github.com/openclaw/openclaw/pull/145630/commits/689b006743a90d57fd6d289d63c14b524da8006a)
- Landed commit: [6b62140fbacfc75bb1c2535a813b11affa955201](https://github.com/openclaw/openclaw/commit/6b62140fbacfc75bb1c2535a813b11affa955201)

PR #145643, opened 2026-09-12
test(auto-reply): split thinking defaults coverage

RomneyDa:
## Summary
- move thinking-default and response-usage coverage into a focused test module
- keep the original thinking suite below the core 1,000-line lint limit
- preserve the existing provider-runtime mocks and assertions without production changes

## Verification
- `node --import ./scripts/tsx.mjs scripts/test-projects.mts src/auto-reply/thinking.test.ts src/auto-reply/thinking-defaults.test.ts` (83/83)
- `node --import ./scripts/tsx.mjs scripts/run-oxlint-shards.mts --only=core --split-core --core-stripe=4/5 --threads=1`
- `pnpm check:changed`


RomneyDa:
Superseded by #145632, which landed the canonical test split for the same core lint line-cap failure while this PR was in CI.

PR #145645, opened 2026-09-12
test(ui): reuse idle chat history fixtures

steipete:
Related: #139428

## What Problem This Solves

The chat send tests repeat the same idle-history fixture in reset, reconnect, and queued-send scenarios despite already having a helper for that value.

## Why This Change Was Made

Reuse `idleChatHistory()` at 13 existing expression sites. Each call stays at the original evaluation point, preserving shared handler values, fresh callback results, and promise/deferred resolution ordering. This removes 40 lines without changing helpers, imports, assertions, or test cases.

## User Impact

No user-visible behavior change. This is maintainer-requested test fixture maintenance.

## Evidence

- Complete chat-send suite before and after: 347 passing cases each initially, then 348 each on current main after the startup/history change; reported labels match in order. Verbose labels may be abbreviated; an exact source inverse independently verifies unchanged cases and assertions.
- Source-extracted checks verify equal property values, descriptors, key order, and fresh object/array/session-row allocation.
- `node scripts/check-changed.mjs -- ui/src/pages/chat/chat-send.test.ts` passed.
- Independent review found no issues. No runtime speedup claim.


steipete:
Merged via squash.

- Prepared head SHA: [f45582a139cafae8a6525f1bb41482e72c518d95](https://github.com/openclaw/openclaw/pull/145645/commits/f45582a139cafae8a6525f1bb41482e72c518d95)
- Landed commit: [b6f7b2a4c82c079ce6be9a486400be3226e0eb61](https://github.com/openclaw/openclaw/commit/b6f7b2a4c82c079ce6be9a486400be3226e0eb61)

Issue #112423, opened 2026-07-21
[Bug]: Large SQLite transcript cleanup blocks the gateway event loop


HermanZeng:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

Archiving a large SQLite-backed transcript during session cleanup performs full transcript materialization, compression, durable file I/O, and readback on the gateway thread, causing event-loop stalls and roughly two times the transcript size in transient gateway heap growth in a controlled reproduction.

### Steps to reproduce

1. Check out `openclaw/openclaw` at `59640bad267da103666d69bc7aad4e0639f41300`.
2. Create a SQLite-backed session with one transcript generation and populate `transcript_events` with JSONL payloads totaling 8-67 MB.
3. Schedule a zero-delay timer, then run the SQLite lifecycle delete/cleanup path with transcript archiving enabled.
4. Record the heap delta while the archive plan is built and the largest timer delay while compression, publication, fsync, and readback complete.
5. Repeat while doubling the transcript payload.

### Expected behavior

Transcript cleanup should preserve the existing durable archive-before-delete semantics without blocking the gateway event loop or retaining a second full in-memory copy of the transcript.

### Actual behavior

The pre-fix path loads every transcript row into an array, constructs a second full JSONL string, and completes archive compression and durable filesystem operations synchronously. In the controlled sweep, both transient heap use and timer delay grew approximately linearly with transcript size; a 67.41 MB transcript produced a 134.95 MB planning heap delta and a 323 ms event-loop gap.

### OpenClaw version

`main` at `59640bad267da103666d69bc7aad4e0639f41300`

### Operating system

Windows 11 Pro 10.0.26200; Node.js 24.18.0

### Install method

Source checkout (`pnpm` workspace; controlled local reproduction harness)

### Model

Not applicable - the reproduction exercises local SQLite session maintenance without a model/provider request.

### Provider / routing chain

Not applicable - the reproduction exercises local SQLite session maintenance without a model/provider request.

### Additional provider/model setup details

No provider, router, credential, or model invocation is involved.

### Logs

```shell
Pre-fix controlled sweep

Transcript payload   Planning heap delta   Maximum event-loop gap
8.43 MB              16.87 MB              49 ms
16.85 MB             33.73 MB              84 ms
33.71 MB             67.47 MB              158 ms
67.41 MB             134.95 MB             323 ms
```

### Screenshots, recordings, and evidence

[...]

PR #145647, opened 2026-09-12
fix(sessions): avoid redundant entry data reads during cleanup

steipete:
## What Problem This Solves

Resolves a problem where lifecycle cleanup performed avoidable allocations across large session stores, including cleanup runs with nothing eligible for removal.

## Why This Change Was Made

The initial inventory now selects only the session key, current session ID, and timestamp it consumes. The existing full-store projection continues to supply the authoritative entries for ownership, reference, and retention decisions. The row population and ordering are preserved.

## User Impact

Memory-plugin lifecycle cleanup avoids materializing one unused copy of the stored entry JSON. Eligibility, ownership, archival, and deletion behavior are unchanged.

## Evidence

- A private real-SQLite before/after probe used eight retained entries with 2 MiB synthetic inline skill prompts each. Large returned entry JSON fell from **33,558,104 to 16,779,052 bytes**. The complete fixture hash, survivors, inventory order, cleanup result, and all other native operations matched; the 1,492 bytes of reference-metadata JSON were unchanged. This measures returned payload volume and does not attribute a production OOM.
- **35 existing tests passed** across lifecycle cleanup, cleanup races, and reclamation admission races. Changed-scope formatting, typechecks, lint, and repository guards passed. Independent code review found no actionable findings through P2.
- The [Blacksmith Testbox public SDK cleanup flow](https://github.com/openclaw/openclaw/actions/runs/34675480540) passed on Linux/Node 24.19.0 with the same real-SQLite fixture and one-copy result. The delegated command exited 0, its fixture was removed, and its lease completed. The earlier transport-only attempt failed before the SDK ran; the unchanged proof passed after correcting command framing.

Related: #112423. This change covers the initial entry inventory; it does not change transcript archive encoding or final reclamation.

### DIFF
--- a/src/config/sessions/session-accessor.sqlite-lifecycle-artifacts.ts
+++ b/src/config/sessions/session-accessor.sqlite-lifecycle-artifacts.ts
@@ -217,7 +217,7 @@ export function planSessionLifecycleArtifactCleanup(
       database.db,
       db
         .selectFrom("session_nodes")
-        .select(["entry_json", "session_key", "current_session_id", "updated_at"])
+        .select(["session_key", "current_session_id", "updated_at"])
         .orderBy("session_key", "asc"),
     ).rows;
 


steipete:
Merged via squash.

- Prepared head SHA: [bc93e507ee2294ce453e355a5b3d74073eba2621](https://github.com/openclaw/openclaw/pull/145647/commits/bc93e507ee2294ce453e355a5b3d74073eba2621)
- Landed commit: [de26ed2ccaa169f22263a889c5ada425901a9ad5](https://github.com/openclaw/openclaw/commit/de26ed2ccaa169f22263a889c5ada425901a9ad5)

PR #145650, opened 2026-09-12
fix(agents): honor CLAUDE_CONFIG_DIR in claude-cli transcript resolution

### LINKED ISSUE #145309: [Bug]: claude-cli backend looks for transcripts under $HOME/.claude and ignores CLAUDE_CONFIG_DIR — missing-transcript resets and cli_live_session_changed failover
## Summary

The bundled `claude-cli` backend looks for Claude Code's per-session transcript under `$HOME/.claude/projects/<workspace-key>/<sessionId>.jsonl` and ignores `CLAUDE_CONFIG_DIR`. When the gateway runs Claude Code with `CLAUDE_CONFIG_DIR` set (a second Claude login on the same host), Claude Code writes transcripts to `$CLAUDE_CONFIG_DIR/projects/…`, so every transcript check misses. The result is lost session continuity on every cold turn and, on warm turns, a silent failover to the next model with no history after any config hot-reload.

Other parts of OpenClaw already honor `CLAUDE_CONFIG_DIR` for the same tree (`session-catalog-scan`'s `configuredClaudeConfigDir`, `session-catalog-registration`'s `claudeProjectsAvailable`, and `provider-auth`'s `resolveClaudeCliConfigDir`), so this looks like one resolver that was missed.

**Version:** OpenClaw 2026.9.4 (3a9d69d), also observed on 2026.9.3. Claude Code 2.1.269. Linux (Ubuntu, systemd user service).

## Setup

- `openclaw-gateway.service` drop-in: `Environment=CLAUDE_CONFIG_DIR=/home/<user>/.ai-profiles/accounts/claude/personal/personal-b`
- Agent model `anthropic/claude-fable-5-1` with `agentRuntime.id: "claude-cli"` (default `liveSession: "claude-stdio"`).
- Claude Code writes transcripts to `$CLAUDE_CONFIG_DIR/projects/-home-<user>--openclaw-workspace-lumen/<sessionId>.jsonl` (verified: every session id below exists there, 0.5–1.2 MB each). `$HOME/.claude/projects/…` has not been written since the drop-in was added.

## What happens

1. **Cold turn** (warm subprocess already exited, e.g. a heartbeat 1–2 h later):

   ```
   [agent/cli-backend] claude-cli transcript probe v4 miss (sessionId-deterministic path, grace 250ms): sessionId=4ad605e8-… expectedPath=/home/<user>/.claude/projects/-home-<user>--openclaw-workspace-lumen/4ad605e8-….jsonl fileExists=false
   [agent/cli-backend] cli session reset: provider=claude-cli reason=missing-transcript
   [agent/cli-backend] cli exec: provider=claude-cli model=claude-fable-5-1 … useResume=false session=none … reuse=invalidated:missing-transcript
   ```

   The stored session id is cleared and a brand-new Claude session starts. 5–11 of these per day on one agent.

2. **Warm turn** (subprocess still alive): because `claudeCliSessionTranscriptMissing` is true, `prepare.runtime` sets `managedClaudeLiveSessionGeneration`, so the run is pinned to the exact live process (`requiredGeneration`). After any config hot-reload (which changes fingerprint inputs — child env, skills snapshot, system prompt), the next resume turn fails in ~350 ms:

   ```
   [agent/cli-backend] cli exec: … useResume=true session=present resumeSession=48ff0699e1cb reuse=reusable
   [agent/cli-backend] cli terminal failure: provider=claude-cli model=claude-fable-5-1 durationMs=506 … error=Managed CLI live session is no longer reusable. | cli_live_session_changed
   [model-fallback/decision] … reason=session_expired next=anthropic/claude-fable-5
   [agent/cli-backend] cli exec: provider=claude-cli model=claude-fable-5 … useResume=false session=none
   ```

   The user's message is silently re-run on the fallback model with no context. The fallback process then occupies the same live-session owner key with a different `model` in its fingerprint, so **every later turn on the primary model fails the same way** (14 consecutive turns on one dashboard session today) until the gateway restarts.

Observed via the Control UI / macOS app (dashboard lanes); channel lanes mostly hit case 1 because their warm process is gone by the next message.

## Expected

`resolveClaudeCliProjectDirForWorkspace` (dist `claude-cli-project-dir-*.mjs`) should resolve the projects tree the same way the session catalog does:

```js
const configDir = process.env.CLAUDE_CONFIG_DIR?.trim();
const base = configDir ? path.resolve(configDir) : path.join(homeDir, ".claude");
return path.join(base, "projects", sanitizeClaudeCliProjectKey(canonicalWorkspaceDir));
```

(or accept the resolved config dir as a parameter from the backend, which already has the child env).

Two smaller follow-ups that this bug exposed, possibly worth their own issues:

- When `requiredGeneration` no longer matches, failing over to a *different* model with no history is a surprising default; spawning a fresh process with `--resume <sessionId>` on the same model would preserve the conversation.
- After a same-backend fallback, the fallback process is registered under the same live-session owner key, so the primary model can never reuse the slot again (`handle.fingerprint !== fingerprint` on `model`). The slot should be keyed or evicted per model.

## Workaround

Symlink the expected path to the real one:

```
ln -s "$CLAUDE_CONFIG_DIR/projects/<workspace-key>" "$HOME/.claude/projects/<workspace-key>"
```

With that in place: no more `missing-transcript` resets, resume turns are no longer pinned to the exact process, and a two-turn test shows `useResume=true … reuse=reusable historyPrompt=present` with no fallback.

## Related

- #105763 (CLAUDE_CONFIG_DIR not reaching the claude-cli subprocess in Docker) — different failure, same env var.
- #85177 (history dropped after auth-profile/auth-epoch invalidation) — the same "fresh session with no history" symptom from another trigger.

xiaoguomeiyitian:
## What Problem This Solves

Fixes an issue where users running the `claude-cli` backend with `CLAUDE_CONFIG_DIR` set (e.g. a second Claude login on the same host) would lose session continuity on every cold turn and silently fall back to the next model with no history after any config hot-reload. The bundled `claude-cli` transcript resolver looked for project JSONL files under `$HOME/.claude/projects/…` and ignored `CLAUDE_CONFIG_DIR`, so every transcript check missed — even though the session catalog scanner, provider-auth resolver, and Claude Code itself all honor `CLAUDE_CONFIG_DIR` as the projects tree root.

## Why This Change Was Made

Two resolvers were hardcoding `~/.claude/projects` instead of honoring `CLAUDE_CONFIG_DIR`:

1. **`resolveClaudeCliProjectDirForWorkspace`** (`src/agents/command/claude-cli-project-dir.ts`) — used by the transcript probe (`claudeCliSessionTranscriptHasContent`) and the doctor health check. A miss here triggers `reason=missing-transcript` session resets on every cold turn.

2. **`resolveClaudeProjectsDir`** (`src/gateway/cli-session-history.claude.ts`) — used by `resolveClaudeCliSessionFilePath` and `readClaudeCliFallbackSeed`. A miss here means the fallback prelude has no harvested context, so a fallback-model re-run starts with zero conversation history.

Both now follow the same pattern as the existing `resolveClaudeCliConfigDir` in `src/plugin-sdk/provider-auth-claude-compat.ts` and `configuredClaudeConfigDir` in `extensions/anthropic/session-catalog-scan.ts`: when `CLAUDE_CONFIG_DIR` is set and non-blank, resolve project transcripts under `<resolved config dir>/projects`; otherwise fall back to `<homeDir>/.claude/projects` as before.

## User Impact

Users who set `CLAUDE_CONFIG_DIR` (for a second Claude login, a dedicated profile, or a Docker-mounted config directory) will no longer experience:
- Cold-turn session resets (`reason=missing-transcript`) that discard the stored Claude session id.
- Warm-turn silent fallback to a different model with no conversation history after a config hot-reload.
- The cascading failure where the fallback process occupies the same live-session owner key, blocking the primary model from reusing the slot.

The existing workaround (symlinking `$HOME/.claude/projects/<key>` → `$CLAUDE_CONFIG_DIR/projects/<key>`) is no longer needed.

### Production-path behavior proof

Exercises the **production exported functions the Gateway actually runs** — `claudeCliSessionTranscriptHasContent` (the session-continuity probe used on cold turns) and `readClaudeCliFallbackSeed` (the fallback-history recovery reader) — against a real `CLAUDE_CONFIG_DIR` tree containing a Claude Code session JSONL (one user/assistant exchange):

```
[setup] CLAUDE_CONFIG_DIR = /tmp/oc-proof-prod-swqXxG/claude-config
[setup] transcript written to /tmp/oc-proof-prod-swqXxG/claude-config/projects/-tmp-oc-proof-prod-swqXxG-workspace-demo/prod-proof-session.jsonl

[1] production transcript probe (transcript under $CLAUDE_CONFIG_DIR):
    claudeCliSessionTranscriptHasContent() = true
    -> session stays LIVE, no missing-transcript reset: true

[2] production fallback history reader (readClaudeCliFallbackSeed):
    recentTurns = 2
    turn[0].role = user
    turn[1].role = assistant
    -> fallback re-run recovers the conversation from $CLAUDE_CONFIG_DIR/projects

[3] negative control (CLAUDE_CONFIG_DIR unset):
    probe = false (expected, no reset-triggering miss leak)
    seed = undefined (expected)

Result: production probe + production history reader both honor CLAUDE_CONFIG_DIR.
```

What this demonstrates:
- The **production transcript probe** finds the existing Claude session under `$CLAUDE_CONFIG_DIR/projects/…` and returns `true` — the stored session id stays live across a cold turn instead of being reset with `reason=missing-transcript`.
- The **production fallback history reader** recovers the full recent-turn history (user + assistant roles) from `$CLAUDE_CONFIG_DIR/projects/…` — a fallback-model re-run starts with the actual conversation instead of no context.
- The **negative control** (env var unset) shows the same session id resolves to `false`/`undefined` under the default `<homeDir>/.claude/projects` tree — the fix is selective, not a lookup leak.

(The one `v4 miss` warn line in the trace is the negative-control run emitting its expected structured miss for the default-tree path — the same warn Gateway operators would see before this fix on every turn.)

### CI

All 169 jobs green on head `d3d5304a5c` (run 34682440091, `openclaw/ci-gate` SUCCESS). Earlier runs hit two unrelated infra flakes — a `resolve-auth.test.ts` 1000ms timeout and a runner-preempted `checks-fast-bundled-protocol` job (0 steps executed) — both cleared on rerun; no test failure was ever traced to this patch.

### Resolver unit tests (4/4 passed)

```
✓ resolves under <homeDir>/.claude/projects by default
✓ resolves under $CLAUDE_CONFIG_DIR/projects when configured
✓ falls back to <homeDir>/.claude/projects for a blank CLAUDE_CONFIG_DIR
✓ resolves a relative CLAUDE_CONFIG_DIR against the process working directory
```

### Test-suite coverage (189/189 passed locally)

- **Transcript probe test** (`attempt-execution.test.ts`, 21 tests in the probe suite): writes a transcript under `$CLAUDE_CONFIG_DIR/projects/…`, verifies `claudeCliSessionTranscriptHasContent` returns `true` inside the env override and the file is absent from the default tree outside it (direct path check, no probe side-effects).
- **Fallback seed test** (`cli-session-history.test.ts`, 168 tests): writes a session JSONL under `configDir/projects/demo-workspace/`, verifies `readClaudeCliFallbackSeed` finds it with `CLAUDE_CONFIG_DIR` set (2 turns, correct roles) and returns `undefined` without it.

### Patch history

- `6d11eda` — initial fix + tests
- `7c3d4f` — switch 6 synchronous `withEnvAsync` callbacks to `withEnv` (typecheck)
- `9205926` — avoid probe grace-period side-effects in the new test's negative assertion
- `4ee3802` / `5fce660` / `d3d5304` — empty CI-rerun commits for unrelated infra flakes

### DIFF
--- a/src/agents/command/attempt-execution.test.ts
+++ b/src/agents/command/attempt-execution.test.ts
@@ -8,6 +8,7 @@ import { createDeferred } from "../../../test/helpers/promise.js";
 import { upsertSessionEntryCore } from "../../config/sessions/session-accessor.js";
 import { waitForSessionTranscriptIndexReconcile } from "../../config/sessions/session-transcript-reconcile.js";
 import { closeOpenClawAgentDatabaseByPath } from "../../state/openclaw-agent-db.js";
+import { withEnvAsync } from "../../test-utils/env.js";
 import { SessionManager } from "../sessions/session-manager.js";
 import { buildAssistantMessage, buildUsageWithNoCost } from "../stream-message-shared.js";
 
@@ -507,6 +508,41 @@ describe("claudeCliSessionTranscriptHasContent", () => {
         homeDir: tmpDir,
       }),
     ).toBe(true);
+  });
+
+  it("resolves the transcript under $CLAUDE_CONFIG_DIR/projects when configured", async () => {
+    const workspaceDir = await makeWorkspace();
+    const configDir = path.join(tmpDir, "claude-config");
+    await withEnvAsync({ CLAUDE_CONFIG_DIR: configDir, HOME: tmpDir }, async () => {
+      await writeClaudeProjectFile(
+        workspaceDir,
+        "config-dir-session",
+        `${JSON.stringify({
+          type: "assistant",
+          message: {
+            role: "assistant",
+            content: [{ type: "text", text: "hello from config dir" }],
+          },
+        })}\n`,
+      );
+      expect(
+        await claudeCliSessionTranscriptHasContent({
+          sessionId: "config-dir-session",
+          workspaceDir,
+          homeDir: tmpDir,
+        }),
+      ).toBe(true);
+    });
+    // Outside the env override the transcript must not resolve under the
+    // default <homeDir>/.claude/projects tree, proving it was written to and
+    // read from $CLAUDE_CONFIG_DIR/projects instead.
+    expect(
+      await claudeCliSessionTranscriptHasContent({
+        sessionId: "config-dir-session",
+        workspaceDir,
+        homeDir: tmpDir,
+      }),
+    ).toBe(false);
   });
 
   it("rejects path-like session ids instead of escaping the Claude projects tree", async () => {
--- a/src/agents/command/attempt-execution.test.ts
+++ b/src/agents/command/attempt-execution.test.ts
@@ -535,14 +535,18 @@ describe("claudeCliSessionTranscriptHasContent", () => {
     });
     // Outside the env override the transcript must not resolve under the
     // default <homeDir>/.claude/projects tree, proving it was written to and
-    // read from $CLAUDE_CONFIG_DIR/projects instead.
-    expect(
-      await claudeCliSessionTranscriptHasContent({
-        sessionId: "config-dir-session",
-        workspaceDir,
-        homeDir: tmpDir,
-      }),
-    ).toBe(false);
+    // read from $CLAUDE_CONFIG_DIR/projects instead. Use a direct path check
+    // to avoid the grace-period wait and warning side-effect of the probe.
+    const defaultProjectDir = resolveClaudeCliProjectDirForWorkspace({
+      workspaceDir,
+      homeDir: tmpDir,
+    });
+    const defaultTranscriptPath = path.join(defaultProjectDir, "config-dir-session.jsonl");
+    const exists = await fs
+      .access(defaultTranscriptPath)
+      .then(() => true)
+      .catch(() => false);
+    expect(exists).toBe(false);
   });
 
   it("rejects path-like session ids instead of escaping the Claude projects tree", async () => {
--- a/src/agents/command/claude-cli-project-dir.test.ts
+++ b/src/agents/command/claude-cli-project-dir.test.ts
@@ -2,7 +2,7 @@
 // default ~/.claude/projects convention.
 import path from "node:path";
 import { describe, expect, it } from "vitest";
-import { withEnvAsync } from "../../test-utils/env.js";
+import { withEnv } from "../../test-utils/env.js";
 import { resolveClaudeCliProjectDirForWorkspace } from "./claude-cli-project-dir.js";
 
 const HOME = path.join(path.sep, "home", "demo-user");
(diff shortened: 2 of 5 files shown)


xiaoguomeiyitian:
@clawsweeper re-review

PR #145652, opened 2026-09-12
refactor(tasks): isolate SQLite connection kernels

steipete:
## What Problem This Solves

Task and flow operations that already own a shared-state transaction currently import row codecs and SQL helpers through modules that also acquire, cache, and close the global database. Task timestamp normalization additionally reaches provider/plugin runtime through a terminal-status predicate in a presentation module. These dependencies make it harder to reuse the same storage operations on an explicitly supplied connection.

## Why This Change Was Made

The existing SQL and row codecs now live in task and flow kernels that operate on their caller's connection. The existing facades retain database acquisition, cache/close behavior, write admission, and public return contracts. Subagent completion/replacement and cron recovery import the connection-bound helpers directly. Terminal-status classification moves to the existing pure task-record type module, and all internal callers follow that owner.

## User Impact

Public task/flow behavior is unchanged. This is preparation for a future common worker owner; SQL remains synchronous and no worker backend is activated. Transaction boundaries, row contents, ordering, read-only migration handling, and native execution-binding behavior are preserved. No schema, runtime configuration, or dependency changes are introduced.

## Evidence

- The 135-case baseline and 178-case final focused suites passed, including real task/flow roundtrips, native binding lifecycle behavior, cron recovery, and compound subagent rollback/recovery.
- A new cold-import contract rejects global database/read-admission and plugin-runtime imports, then exercises two real SQLite connections and compound task, delivery, flow, and binding rollback. It failed on the old presentation predicate dependency and passed after the pure-owner cutover.
- Repeated-operation observations are summarized below. Different concurrent host load limits performance conclusions; use these as a regression check.
- The complete changed-file gate, `pnpm build`, and fresh P2 review passed on the final source.
- Hosted architecture CI caught the raw-SQL exception still naming the old facade. Its existing path is moved one-for-one to the canonical kernel; the unchanged `NOT INDEXED` integrity query and exception count are preserved. The actual Kysely guard and fresh P2 review pass.
- The moved raw-row assertion now documents its canonical-schema invariant. The old facade's unused one-assertion allowance is removed as an exact shrink-only ratchet update.

Seven samples of 250 operations used 128 records. Values are milliseconds per operation, median (sample range).

| Operation | Baseline | Final |
| --- | ---: | ---: |
| Compound task/flow write + read | 1.2305 (1.1140–1.4295) | 0.9982 (0.8912–1.2471) |
| Task + delivery upsert | 1.1403 (0.7655–1.3378) | 0.8246 (0.7605–0.9109) |
| Ordered task/flow snapshots | 1.8731 (1.8341–1.9607) | 1.8356 (1.8257–1.9237) |
| Runtime/source lookup | 0.0950 (0.0865–0.1110) | 0.0924 (0.0894–0.0997) |

The one-minute load average declined from 28.2 to 14.8. These measurements do not isolate the extraction's effect on throughput. RSS was 651,904 → 364,544 KiB under the same uncontrolled-load limitation.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3558,7 +3558,6 @@ src/tasks/task-registry-query.ts	4
 src/tasks/task-registry-state.ts	3
 src/tasks/task-registry.process-state.ts	1
 src/tasks/task-registry.sqlite.shared.ts	1
-src/tasks/task-registry.store.sqlite.ts	1
 src/tasks/task-registry.ts	1
 src/tasks/task-registry.types.ts	2
 src/tasks/task-status.ts	1
--- a/docs/reference/database-schemas/storage-changes.md
+++ b/docs/reference/database-schemas/storage-changes.md
@@ -31,6 +31,14 @@ on the supplied `node:sqlite` connection. Calling Kysely's asynchronous
 dialect can identify syntax coupling, but does not prove driver behavior,
 isolation, or database compatibility.
 
+Task and flow stores keep row codecs and SQLite operations in connection-bound
+kernels. Their existing facades retain global connection acquisition, cache and
+close behavior, and write transaction admission. Compound subagent and cron
+operations call the kernels on their already-admitted connection. Task status
+classification stays with the pure record types, so decoding does not load
+provider or plugin runtime ownership. These operations and their transaction
+callbacks remain synchronous; this separation does not move SQL to a worker.
+
 Acquire a connection once for an operation and pass that exact connection
 through its transactional helpers. SQLite write callbacks remain synchronous:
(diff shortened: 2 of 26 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d4cfdb21c1b1856a7072766d4c73ce477ba7170c](https://github.com/openclaw/openclaw/pull/145652/commits/d4cfdb21c1b1856a7072766d4c73ce477ba7170c)
- Landed commit: [0283d2d9240e88195445e7ce872e0ada5e85259f](https://github.com/openclaw/openclaw/commit/0283d2d9240e88195445e7ce872e0ada5e85259f)

PR #145654, opened 2026-09-12
refactor(tests): reuse command config runtimes

steipete:
## What Problem This Solves

Three configuration-related test suites duplicate runtime mocks instead of using their existing shared helper. A local log extractor also coerces the first captured argument into a string, hiding its actual type.

## Why This Change Was Made

Reuse the canonical configuration-test runtime. Remove the one-use stringification extractor and assert the first logged argument directly against the same expected text. The production owner emits a joined string; the assertion retains first-message scope without adding call-count or arity constraints.

## User Impact

No production change. Tests are +15/-35 (net -20), with all 16 cases retained. Config ownership, read-time environment, error and exit expectations, writers, and cleanup remain covered.

## Evidence

The original paired proof passed 16 cases before and after the runtime-fixture cleanup. The final revision passes all 16 existing cases on the refreshed base, retaining the same inventory. The shared typed mock exposed a no-base-to-string lint error; the direct assertion resolves that error without a cast or suppression.

Exact typed lint and all 20 normal changed-file checks passed, including both selected core test graphs and all three dead-export scans. Independent Codex review found no actionable findings. The signed commit preserves the final tested bytes. Historical proof and the earlier download/lint failures are retained separately.

### DIFF
--- a/src/cli/command-config-resolution.test.ts
+++ b/src/cli/command-config-resolution.test.ts
@@ -1,5 +1,5 @@
-// Command config resolution tests cover config lookup before command execution.
 import { beforeEach, describe, expect, it, vi } from "vitest";
+import { createTestRuntime } from "../commands/test-runtime-config-helpers.js";
 
 const mocks = vi.hoisted(() => ({
   resolveCommandSecretRefsViaGateway: vi.fn(),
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4fae27232b633fb997ba400e5921430327e896b2](https://github.com/openclaw/openclaw/pull/145654/commits/4fae27232b633fb997ba400e5921430327e896b2)
- Landed commit: [a53c601cd4075a6839bac3f75f2280eda182b4b8](https://github.com/openclaw/openclaw/commit/a53c601cd4075a6839bac3f75f2280eda182b4b8)

PR #145484, opened 2026-09-12
feat: manage and reload plugins without Gateway restarts

steipete:
Builds on [plugin lifetime foundation #144252](https://github.com/openclaw/openclaw/pull/144252). Related: [plugin redesign #135599](https://github.com/openclaw/openclaw/pull/135599).

<details>
<summary>Additional instructions</summary>

Keep **Allow edits from maintainers** enabled.

</details>

## What Problem This Solves

Plugin changes should take effect without restarting the Gateway or disconnecting clients. TypeScript reload must execute edited code—including on Bun, which could report success while retaining old modules—and failed activation needs a clear recovery action.

## Why This Change Was Made

Replace plugin restart signaling and UI restart waits with one Gateway-owned prepare, drain, publish, and retire lifecycle, retaining unaffected instances and sharing the startup publisher. Per-instance module capture replaces Bun's global cache-retirement barrier: existing consumers retain their original modules, declared local package targets are acquired before native evaluation, and native resolution conditions and synchronous startup contracts are preserved.

## User Impact

Administrators can install, enable, disable, uninstall, and reload through Gateway controls. `openclaw plugins reload <id>` refreshes TypeScript entries, helpers, and manifests while preserving enablement; the Control UI updates on its existing connection. Results distinguish saved installation from applied runtime and offer Reload after activation fails. Runtime-only reload works with externally managed/Nix config; new consent and configuration writes remain guarded.

Cleanup is best effort: unmanaged globals, timers, or native resources can remain after managed registrations retire. Cleanup failures produce warnings; restart remains available for residual behavior. Reload does not rebuild compiled bundled code or grant ownership of discovered files. Broader agent-turn, installation-source, and Claw batch integrations remain separate slices.

## Evidence

The corrective candidate `4d9b6534` passed the canonical changed gate (867 seconds), full build (449 seconds), and all six P2 autoreview passes. The exact previously failing Bun 1.4.0 test passed. Fresh Bun 1.4.2 and Node 24.19.0 source Gateways each executed the real CLI-handler RPC reload: edited TypeScript, same PID, unchanged config, generation 3, clean shutdown and reusable port. This proves the handler/RPC flow, not a spawned CLI parser. The inherited Bun 1.4.0 native ESM/SDK-hook limitation remains qualified separately.

[...]

PR #132569, opened 2026-08-29
fix(macos): keep config edits made while a save is in flight

HuzaifaChaudary:
Closes #127614

## What Problem This Solves

Fixes an issue where users editing a Channel or Advanced Config field while a save is still running would lose that edit. The older saved values are republished over it and Save greys out, so the form looks fully persisted while the newest change is gone.

Only the Save and Reload buttons are disabled during a save. The actual form controls stay live, which is what makes this reachable on any ordinarily slow save.

## Why This Change Was Made

`saveConfigDraft` took its copy of the draft, awaited the write, and then called `loadConfig()`. That default is `force: true`, and a forced snapshot is exactly the one that bypasses the dirty guard in `applyConfigSnapshot`, replaces the whole draft and clears dirty. An edit admitted while the write was in flight had no way to survive it.

Every edit now bumps a draft revision, and the save captures that revision when it admits the draft. The reload only forces when the revision has not moved since. When it has, the unforced path is used, and that path already defers to a dirty draft, so the newer edit and its dirty state are left alone.

The reload is still forced for the ordinary case, so normalization from the gateway and the dirty clear both behave as before when nothing was typed during the save.

## User Impact

Edits made while a save is running are kept, and Save stays enabled so they can be written. Channel tokens, toggles and Advanced Config fields no longer disappear silently on a slow save.

No change for a save with nothing typed behind it, which is the ordinary path.

## Evidence

Added to `ChannelsSettingsSmokeTests`, next to the existing `non forced config snapshots do not overwrite dirty draft` which covers the snapshot side but not the save overlap:

```
✔ Test "an edit during a save keeps the save from reloading over it" passed
✔ Test "a save with no edit behind it still reloads and clears dirty" passed
✔ Suite ChannelsSettingsSmokeTests passed  (13 tests)
```

The first follows the sequence in the issue. It records the revision at admission, edits a still-enabled field, and checks the save can no longer replace the draft. It then applies a snapshot carrying the older saved values through the unforced path and asserts the newer edit and dirty both survive:

```swift
let admitted = store.configDraftRevision
store.updateConfigValue(path: [.key("channels"), .key("discord"), .key("enabled")], value: true)

#expect(ChannelsStore.saveMayReplaceDraft(
    submittedRevision: admitted,
    currentRevision: store.configDraftRevision) == false)

store.applyConfigSnapshot(snap, sourceKey: "source-a", force: false)
#expect(discord?["enabled"] as? Bool == true)
#expect(store.configDirty == true)
```

[...]

PR #145660, opened 2026-09-12
fix: prevent stale config drafts from overwriting included values

### LINKED ISSUE #127444: Include-only config edits do not advance the write CAS token and can be overwritten by stale drafts
## What happens

Gateway config writes use a `baseHash` derived only from the raw root file. Editing only a referenced `$include` leaves that token unchanged, so a client draft observed before the include edit can pass both RPC admission and canonical writer CAS and overwrite the newer included values.

A real config IO/writer reproduction at frozen campaign SHA observed:

```json
{"rootHashUnchanged":true,"initialResolvedPort":18789,"externalResolvedPort":19001,"staleRequestedPort":19002,"outcome":"accepted","finalResolvedPort":19002}
```

## Root cause

Current `main` (`3376c29800166a3151cbca6b8ab204964e97ac39`) retains:

- `src/config/io.snapshot.ts` computes snapshot `hash` from root raw bytes only; include hashes are separately tracked.
- `config.get` publishes raw-root `hash` as the write guard and a separate resolved `configRevisionHash`.
- Control UI correctly notices resolved revision changes but stores/sends the raw-root hash for manual save, autosave, and teardown flush.
- Gateway `config.set`, `config.patch`, and `config.apply` compare `baseHash` only to the raw-root snapshot hash.
- `replaceConfigFile` rechecks the same raw-root hash under its mutation lock. Per-include commit fencing protects changes after mutation planning starts, not changes between client observation and that fresh planning snapshot.

Targeted live issue/PR search found no exact duplicate.

## Impact

Stale full writes can lose multiple newer include-owned values; same-leaf patch can lose a concurrent included value. This contradicts the documented base-hash clobber-prevention contract.

## Repair direction

Add a formatting/identity-aware mutation revision derived from root raw bytes plus the ordered include graph's target identities/raw hashes. Publish it from `config.get`, send it as the draft base, compare it at RPC admission and again under the canonical write lock. Keep resolved semantic revision and per-include commit fencing as separate identities.

Regress include A observed → external include B → stale set/apply/same-leaf patch, proving conflict and preservation of B at both RPC and canonical writer boundaries.

This requires an additive protocol/backward-compatibility decision for old clients and changes canonical config persistence, so maintainer review is required.

obviyus:
Closes #127444. Reported by @steipete.

Related: #133163, #132569, #145484, #145817.

## What Problem This Solves

A stale configuration draft could overwrite a newer value in an included file because the revision checked only the root file.

## Why This Change Was Made

The snapshot owner includes captured include content and targets in the revision. The mutation owner checks it under the existing lock, rereads saved configuration after writing, and returns configuration and revision from that same canonical reread. Audit records retain their raw-file identity so history still joins the correct changes.

## User Impact

Stale include edits return the existing reload conflict and preserve newer values. Refreshing the draft permits the write. Root-only revisions, stored formats, and update compatibility remain unchanged.

Consumers: configuration editors and writers now detect include changes; write responses return a matching configuration/revision pair; system change history retains correct grouping.

## Evidence

Authenticated config.patch, config.set, and config.apply tests cover stale rejection and refreshed writes. Concurrent config.set requests preserve both writes. Registered configure continuation and authenticated change-history regressions fail before the repair and pass after it. After rebase, 82 activation, setup, audit, and configure cases passed. Released-state startup and updater checks passed on earlier builds.


obviyus:
Thanks @steipete for reporting #127444.

**Bug:** A stale configuration draft could overwrite a newer value in an included file.

**Fix:** The existing write revision includes captured include content and target identities. Writes return a matching configuration/revision pair; raw audit and recovery identities stay intact.

| Before | After |
| --- | --- |
| A stale authenticated patch replaced newer `warn` with `debug`. | The same patch gets a reload conflict and preserves `warn`; a refreshed draft saves successfully. |

**Proof:** Real Gateway requests, registered patch/set/apply and configure regressions, audit history, concurrent writers, and published-updater/previous-state checks. Historical execution identities and test limits are retained in the PR evidence.

PR #145666, opened 2026-09-12
chore: compact agent event test matrix

RomneyDa:
## What Problem This Solves

The agent-event suite repeats property names across 27 homogeneous case rows, making the regression matrix longer and harder to scan than its behavior requires.

## Why This Change Was Made

Converts only the repeated object rows to positional tuples while preserving every value, title dimension, case, and assertion.

## User Impact

No user-visible behavior changes. Maintainers get the same gateway regression coverage with 72 fewer test LOC.

## Evidence

- `pnpm exec vitest run src/gateway/server-chat.agent-events.test.ts` — 226 tests passed

### DIFF
--- a/src/gateway/server-chat.agent-events.test.ts
+++ b/src/gateway/server-chat.agent-events.test.ts
@@ -1638,33 +1638,17 @@ describe("agent event handler", () => {
   );
 
   it.each([
-    {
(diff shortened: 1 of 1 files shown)


RomneyDa:
@clawsweeper review

PR #145667, opened 2026-09-12
chore: compact WhatsApp filename media test matrix

RomneyDa:
## What Problem This Solves

The WhatsApp filename-media suite repeats five property names across every row of a homogeneous MIME-selection matrix.

## Why This Change Was Made

Converts only the repeated object rows to positional tuples while preserving all 14 filename, content-type, MIME, payload, and buffer combinations.

## User Impact

No user-visible behavior changes. Maintainers get identical WhatsApp media regression coverage with 66 fewer test LOC.

## Evidence

- `pnpm exec vitest run extensions/whatsapp/src/auto-reply/deliver-reply.filename-media.test.ts` — 14 tests passed


RomneyDa:
@clawsweeper review

PR #145668, opened 2026-09-12
chore: compact orphan recovery test matrix

RomneyDa:
## What Problem This Solves

The orphan-recovery integration suite repeats four property names across its restart-outcome matrix, obscuring the meaningful state transitions.

## Why This Change Was Made

Converts only the repeated object rows to positional tuples while preserving all stop reasons, sources, expected outcomes, timeout phases, cases, and assertions.

## User Impact

No user-visible behavior changes. Maintainers get identical restart-recovery coverage with 50 fewer test LOC.

## Evidence

- `pnpm exec vitest run src/agents/subagents/registry/subagent-orphan-recovery.restart-integration.test.ts` — 20 tests passed


RomneyDa:
@clawsweeper review

PR #145669, opened 2026-09-12
chore: compact OpenResponses HTTP test matrices

RomneyDa:
## What Problem This Solves

The OpenResponses HTTP suite repeats object keys across several homogeneous request and stream behavior matrices, adding noise around high-value protocol assertions.

## Why This Change Was Made

Converts only repeated object rows to positional tuples while preserving optional values, empty strings, booleans, rendered titles, every case, and every assertion.

## User Impact

No user-visible behavior changes. Maintainers get identical OpenResponses HTTP regression coverage with 50 fewer test LOC.

## Evidence

- `pnpm exec vitest run src/gateway/openresponses-http.test.ts` — 130 tests passed

### DIFF
--- a/src/gateway/openresponses-http.test.ts
+++ b/src/gateway/openresponses-http.test.ts
@@ -399,13 +399,13 @@ describe("OpenResponses HTTP API (e2e)", () => {
   });
 
   it.each([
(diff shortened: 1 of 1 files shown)


RomneyDa:
@clawsweeper review

PR #145671, opened 2026-09-12
fix(gateway): reduce memory churn when listing sessions

steipete:
Related: #145608

## What Problem This Solves

Fixes an issue where users refreshing large session inventories cause excessive temporary memory allocation in the Gateway, including while the Control UI refreshes session rows during other activity.

## Why This Change Was Made

Reuse the existing session entry pairs through filtering and owner-first selection, avoid creating entry tuples and copying unrelated options during child indexing, and collect the two possible navigation owners directly. Established session rows now skip phantom-key parsing. Controller order, parent deduplication, current liveness, filtering, and paging remain unchanged.

## User Impact

Large session inventories create less garbage during each refresh. No configuration or storage migration is required.

## Evidence

- Three matched samples of the actual asynchronous listing entry point: 10,000 sessions, 20% child links, 100-row page, 50 measured calls per sample after warmup. Median V8 sampled allocations fell from 28.226 MB to 17.097 MB per list (39.4%). Sampling includes objects collected by both minor and major GC.
- Full projected rows have an identical SHA-256 before and after. This measures allocation churn; retained-memory and RSS changes are not claimed.
- 58 focused tests passed across subagent navigation, ownership, row-cache, and phantom-placeholder coverage.
- Independent P0-P2 review completed with no findings.
- Production and core-test typechecks, formatting, changed-file lint, and all remaining selected state/import guards passed. The broad local changed gate failed on `canonicalProviderName` and `listPackagedStaticExtensionAssetOutputs`; exact scans reproduced both on unchanged baseline `7275b4b`. No ignores or skip flags were added.
- Paired live OpenAI proof on one isolated Linux Testbox, with this patch plus #145672: 1,000 sessions, eight concurrent model turns, 1,200 history reads, 100 session updates, 100 list/control samples, and subscription probes. All streamed/final replies, RPC history, independent persisted SQLite rows, and probes passed in both arms. Complete source verification matched the three reviewed production files; both arms used the same harness and runtime.
- That single live pair measured load allocations of 4.016 GB → 3.975 GB (-1.0%), and seed plus load of 9.531 GB → 9.501 GB (-0.3%). Session-list inclusive allocations fell 6.1%. These sampled estimates exclude native/worker allocations and do not establish statistical significance. Post-GC heap was effectively flat (327.399 → 327.734 MB), as was sampled workload RSS (818.79 → 822.43 MiB); no retained-memory reduction is claimed.

### DIFF
--- a/src/gateway/session-utils-core.ts
+++ b/src/gateway/session-utils-core.ts
@@ -256,9 +256,15 @@ export function resolveSessionChildOwners(params: {
     ? normalizeOptionalString(latest.controllerSessionKey) ||
       normalizeOptionalString(latest.requesterSessionKey)
     : normalizeOptionalString(entry.spawnedBy);
-  return [...new Set([controller, normalizeOptionalString(entry.parentSessionKey)])].filter(
-    (owner): owner is string => Boolean(owner) && owner !== key,
-  );
+  const parent = normalizeOptionalString(entry.parentSessionKey);
+  const owners: string[] = [];
+  if (controller && controller !== key) {
+    owners.push(controller);
+  }
+  if (parent && parent !== key && parent !== controller) {
+    owners.push(parent);
+  }
+  return owners;
 }
 
 /** Index only canonical children; retained run results cannot create session links. */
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [743c2e2d52a4a6160fd80332476db0151a84a0d4](https://github.com/openclaw/openclaw/pull/145671/commits/743c2e2d52a4a6160fd80332476db0151a84a0d4)
- Landed commit: [b567e3efd6eae052a02053ea6bf98060f48bc6db](https://github.com/openclaw/openclaw/commit/b567e3efd6eae052a02053ea6bf98060f48bc6db)

PR #145674, opened 2026-09-12
test: run Node fixtures with Node under Bun

steipete:
## What Problem This Solves

`runNodeScript()` launched `process.execPath`, so when Vitest itself ran under Bun, tests explicitly intended to probe Node instead launched Bun. Node-specific `--expose-gc` retention checks then measured a different garbage collector and produced five false failures.

## Why This Change Was Made

The shared helper now uses the existing `resolveTestNodeExecPath()` boundary. Node-hosted tests keep their current executable; Bun-hosted tests skip Bun's `node` shim and launch a real Node process from `PATH`.

## User Impact

This changes test infrastructure only. It lets the direct-Bun test campaign preserve Node-specific subprocess contracts without weakening their memory assertions or changing production behavior.

## Evidence

- Direct custom Bun: `test/helpers/run-node-script.test.ts` 5/5 passed.
- Direct custom Bun: read and exec-output retention probes 5/5 passed; before the fix all five failed under Bun's GC/runtime semantics.
- Node: the same helper tests 5/5 and retention probes 5/5 passed.
- Test-audit authoring gate: the existing tests protect Node memory-retention behavior, the credible regression is executing them under another runtime, existing coverage did not enforce the child runtime, and the fix adds no test-only production seam.
- Independent P0-P2 autoreview: scoped clean at 0.90 confidence.

### DIFF
--- a/test/helpers/run-node-script.ts
+++ b/test/helpers/run-node-script.ts
@@ -1,4 +1,5 @@
 import { runManagedCommand } from "../../scripts/lib/managed-child-process.mts";
+import { resolveTestNodeExecPath } from "../../src/test-utils/node-process.js";
 import { createBoundedChildOutput } from "./bounded-child-output.js";
 
 export async function runNodeScript(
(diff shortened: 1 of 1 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [70b76f9e453d399e2fab590eb59ed7c9e906d6b3](https://github.com/openclaw/openclaw/pull/145674/commits/70b76f9e453d399e2fab590eb59ed7c9e906d6b3)
- Landed commit: [13e94870dd5ae1ed4cfe1b82734012a8028ab15b](https://github.com/openclaw/openclaw/commit/13e94870dd5ae1ed4cfe1b82734012a8028ab15b)

PR #145677, opened 2026-09-12
improve: reduce repeated plugin state registration work

steipete:
## What Problem This Solves

Plugin state registrations repeatedly compile the same quota queries and load an existing payload even when the eviction policy does not use it.

## Why This Change Was Made

Reuse compiled namespace and plugin counts for each database connection through the existing prepared-query executor. Ordinary evicting registrations skip the unused payload read; reject-new and migration batch retention preserve their existence checks. Quota, expiry, transaction, and recovery behavior stay the same.

## User Impact

Less SQLite query setup and payload allocation during plugin state writes.

## Evidence

- 121 tests passed across 13 plugin-state files; three existing Windows exclusions remain skipped.
- Paired measurements of the actual registration kernel on Node 26.8.2 / SQLite 3.53.4: 100 rows, 84.99 → 64.64 µs; 3,000 rows, 601.24 → 553.53 µs; 30,000 rows, 5,135.06 → 4,876.11 µs. Seven alternating rounds after warmup.
- Full public registration timings were noisy and do not establish an end-to-end speedup.
- The compilation regression assertion fails on the original code and passes with fresh bindings after reopening. Independent review found no P0–P2 issues.

### DIFF
--- a/src/plugin-state/plugin-state-store.kernel.ts
+++ b/src/plugin-state/plugin-state-store.kernel.ts
@@ -298,19 +298,49 @@ export function deleteExpiredPluginStateEntries(
   return Number(result.numAffectedRows ?? 0);
 }
 
+type PluginStateNamespaceCountParams = { pluginId: string; namespace: string; now: number };
+type PluginStateCountRow = { count: number | bigint };
+const pluginStateNamespaceCountQueries = new WeakMap<
+  DatabaseSync,
+  ReturnType<typeof prepareSqliteQuerySync<PluginStateNamespaceCountParams, PluginStateCountRow>>
+>();
+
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [efd267503c0bc9e44e80600708d517bdff44c756](https://github.com/openclaw/openclaw/pull/145677/commits/efd267503c0bc9e44e80600708d517bdff44c756)
- Landed commit: [a768318dfdb079895c54bb0849d77dab5dfd0e20](https://github.com/openclaw/openclaw/commit/a768318dfdb079895c54bb0849d77dab5dfd0e20)

PR #145678, opened 2026-09-12
improve: stop SQLite first-row reads after one result

steipete:
## What Problem This Solves

SQLite callers requesting one result currently allocate and iterate every selected row before returning the first, adding work to frequent state lookups.

## Why This Change Was Made

Use node:sqlite's `get()` for compiled SELECT roots in the first-row helper. The existing statement cache still owns reuse, invalidation, and reentry. Raw SQL and writes, including RETURNING statements, retain completion semantics. Node's `get()` reads column metadata after SQLite can reprepare and resets the reader before returning.

## User Impact

Faster state lookups and no evaluation of unnecessary later SELECT rows.

## Evidence

- 31 focused runtime and type tests passed on Node 26.8.2. The new regression fails on the original executor by evaluating all three fixture rows. Coverage includes schema changes, empty results, released readers, cache reentry, authorizer invalidation, and write completion.
- Same-process alternating baseline/candidate measurements, seven rounds of 20,000 calls: fixed-query first-row execution 5.175 → 3.404 µs; rebuilt-query execution 7.531 → 4.980 µs on Node 26.8.2 / SQLite 3.53.4, about 34% lower elapsed time. Node 24.19.0 also improved by 24–25%.
- Production typecheck, formatting, and repository guards passed. The local full core-test typecheck wrapper refuses the Windows checkout's percent-encoded space; hosted CI must complete that gate.
- Integration observers now count native `get()` as well as iteration, preserving archive-race and exact 8 MiB transactional materialization assertions. Upload-store (23 tests), plugin retention (10), and 15 of 16 cold-read cases passed locally. The remaining canonical-copy timeout and sibling Windows cleanup/rename/executable-bit failures were reproduced on the unchanged executor; assertions, timeouts, and skips were not weakened. Exact-head Linux CI supplies full integration proof.
- Independent P0–P2 reviews of the executor and subsequent observer repair completed clean. Scoped type-aware lint and diff checks passed.

- Final observer audit also covers unreadable update destinations, archive read budgets, a real WAL peer-writer race, model-context acquisition counts, and installed-plugin record caching. All 85 affected portable tests pass; the Unix-only destination fault injection fails before the observer correction and passes on Linux afterward. Fresh P0–P2 review, scoped type-aware lint, formatting, and selected repository guards pass. Production behavior is unchanged by these test repairs.

### DIFF
--- a/src/agents/embedded-agent-runner/run/attempt-transcript-helpers.presence.test.ts
+++ b/src/agents/embedded-agent-runner/run/attempt-transcript-helpers.presence.test.ts
@@ -72,19 +72,31 @@ it("keeps one presence snapshot while another connection classifies a message", 
       ) {
         return statement;
       }
+      const classify = () => {
+        // Commit classification after the index probe finishes, while the
+        // physical message remains present throughout both connections.
+        writer
+          .prepare(
+            "UPDATE transcript_event_identities SET event_type = 'message' WHERE session_id = ?",
+          )
+          .run(sessionTarget.sessionId);
+        classified = true;
+      };
       return new Proxy(statement, {
         get(target, property) {
+          if (property === "get") {
+            return new Proxy(target.get.bind(target), {
+              apply(get, _receiver, params) {
+                const row = get(...params);
+                classify();
+                return row;
+              },
+            });
+          }
           if (property === "iterate") {
             return (...params: Parameters<typeof target.iterate>) => {
               const rows = [...target.iterate(...params)];
-              // Commit classification after the index probe finishes, while the
-              // physical message remains present throughout both connections.
-              writer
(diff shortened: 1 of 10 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f70532fb701f36d37ac4dcb4d2a0dbf9e5a4277d](https://github.com/openclaw/openclaw/pull/145678/commits/f70532fb701f36d37ac4dcb4d2a0dbf9e5a4277d)
- Landed commit: [27590ea68a1d4b33df5ab6617cd7f45ec11425de](https://github.com/openclaw/openclaw/commit/27590ea68a1d4b33df5ab6617cd7f45ec11425de)

PR #145687, opened 2026-09-12
improve: keep terminal output responsive during heavy logging

steipete:
## What Problem This Solves

Resolves a problem where sustained terminal output made of small writes spends increasing CPU time maintaining scrollback after the buffer fills.

## Why This Change Was Made

The shared bounded buffer advances a head index and clears evicted references instead of shifting every retained element after each append. It compacts when the cleared prefix reaches half the array, keeping eviction work amortized and retained storage bounded. Terminal snapshots consume the same buffer without allocating an intermediate array.

## User Impact

Busy terminal streams spend less time maintaining scrollback. Whole-write eviction, Unicode-safe fitting, replay offsets, FIFO order, actual undefined values, and latch/fail-closed behavior remain unchanged.

## Evidence

- Windows x64, Node 26.8.2; actual terminal output controller and ring, eight synthetic streams with 2 MiB per stream and snapshots every 32 KiB. Ten samples per variant in baseline/candidate/candidate/baseline order, with identical source transformation outside timing and recorded source hashes. For 32-character writes, median time excluding assertion time fell from 4,181 to 675 ms (6.20x); process CPU fell from 3,875 to 703 ms. For 128-character writes, median time fell from 486 to 270 ms (1.80x).
- Replay contents, end offsets, and emitted frame order matched. All 16 existing/extended bounded-buffer, session-manager output-ring, and node-relay tests passed on the final source. Targeted lint and production core typecheck passed. Independent review found no actionable issues.
- Larger-write measurements limit the claim: ordered 4 KiB cases regressed; fresh-process warmed 4 KiB comparisons had similar wall/snapshot costs (47.85 vs 45.21 ms wall). CPU granularity and shared-host activity limit small differences. No general large-write or native PTY speedup is claimed; a separate Windows native PTY attempt stalled during startup and produced no usable timing evidence.
- Benchmarks are local task artifacts, not timing-sensitive CI tests. Required hosted CI remains a merge gate.

### DIFF
--- a/src/gateway/terminal/output-ring.ts
+++ b/src/gateway/terminal/output-ring.ts
@@ -19,6 +19,7 @@ export class TerminalOutputRing extends BoundedBuffer<string> {
   }
 
   snapshot(): string {
+    // Evicted slots are cleared, so joining with no separator skips that prefix without a copy.
     return this.values.join("");
   }
 }
--- a/src/gateway/terminal/session-manager.output-ring.test.ts
+++ b/src/gateway/terminal/session-manager.output-ring.test.ts
@@ -20,6 +20,14 @@ describe("TerminalSessionManager output ring", () => {
     fake.emitData("ijkl");
     // Cap exceeded: the oldest whole chunk goes; boundaries stay intact.
     expect(manager.snapshot(outcome.sessionId)).toBe("efghijkl");
+
+    for (let value = 0; value < 5_000; value += 1) {
+      fake.emitData(String(value).padStart(4, "0"));
+    }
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [8e52c2332390fca04908fb7aa332c5b681dea427](https://github.com/openclaw/openclaw/pull/145687/commits/8e52c2332390fca04908fb7aa332c5b681dea427)
- Landed commit: [9e44dba516e8967c64e709e5ff837f092880e17d](https://github.com/openclaw/openclaw/commit/9e44dba516e8967c64e709e5ff837f092880e17d)

PR #145688, opened 2026-09-12
fix(agents): run tool code child with Node under Bun

steipete:
## What Problem This Solves

Bun cannot run `tool_search_code` because the sandboxed code child requires Node's `--permission` flag. The runtime checked Bun's own flags and launched `process.execPath`, so a direct-Bun MCP recovery test failed before the code child started.

## Why This Change Was Made

OpenClaw now resolves a real Node executable for runtime-specific sidecars. The resolver walks `PATH` plus stable system locations, skips Bun's optional `node` shim, verifies the requested Node flag in a sterile probe process, and caches the result by runtime and executable-search inputs. Tool Search uses that verified Node for its isolated child and advertises code mode under Bun only when a compatible Node is available. The existing test Node resolver now shares the same owner.

## User Impact

Bun-hosted OpenClaw can use the full sandboxed Tool Search code mode while Node behavior stays unchanged. This adds no dependency and does not alter Node installation or package-manager behavior; it only locates an already-installed compatible Node for the child process. Electron keeps its existing structured-tool fallback.

## Evidence

- Direct custom Bun: Tool Search MCP recovery 8/8 passed; before the fix the code-mode case failed with the Node `--permission` runtime error.
- Direct custom Bun: full focused Tool Search and child lifecycle set 204/204 passed, plus the Node resolver regression 1/1.
- Node: the same focused matrix passed 213/213.
- Bun config smoke resolves code mode in about 47 ms on the first probe and 0.02 ms from cache.
- Test-audit: the existing shim regression protects runtime identity, the credible regression is a Bun shim being selected as Node, and the refactor adds no test-only production seam or duplicate test.
- Independent P0-P2 autoreview: scoped clean at 0.89 confidence.

### DIFF
--- a/src/agents/tool-search-code-mode.ts
+++ b/src/agents/tool-search-code-mode.ts
@@ -3,6 +3,7 @@ import { spawn } from "node:child_process";
 import os from "node:os";
 import { isRecord } from "@openclaw/normalization-core/record-coerce";
 import { sliceUtf16Safe } from "@openclaw/normalization-core/utf16-slice";
+import { resolveNodeRuntimeExecutable } from "../infra/node-runtime-executable.js";
 import type { AgentToolUpdateCallback } from "./runtime/index.js";
 import { appendBoundedTextTail, SESSION_TOOL_STDERR_TAIL_BYTES } from "./sessions/tools/limits.js";
 import { TOOL_SEARCH_CODE_MODE_CHILD_SOURCE } from "./tool-search-code-mode-child.js";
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4f6cd753158b0b4ab5446f83d665cb43cdfa2779](https://github.com/openclaw/openclaw/pull/145688/commits/4f6cd753158b0b4ab5446f83d665cb43cdfa2779)
- Landed commit: [21db4e956ed4affe1d3d5966eee44804681f645d](https://github.com/openclaw/openclaw/commit/21db4e956ed4affe1d3d5966eee44804681f645d)

PR #145693, opened 2026-09-12
improve: avoid copying discarded memory search payloads

steipete:
## What Problem This Solves

Memory searches that fall back from the vector extension copy every matching chunk's text into JavaScript, even though only a few chunks become results. This wastes time and allocations for large documents.

## Why This Change Was Made

Scan embeddings in the existing bounded batches and hydrate payloads only when their scores enter the best-result set. Hydration runs while the batch iterator holds the same SQLite read snapshot. The iterator closes before yielding to the event loop, preserving cancellation, equal-score order, and the association between a score and its original text.

## User Impact

Faster fallback memory searches, especially when indexed chunks contain substantial text. Native vector search and the public search contract stay the same.

## Evidence

- 50 memory-search tests passed. All 26 orchestration cases also passed across the sibling run and a focused rerun of the updated cancellation case.
- Real WAL connections verify that an external writer cannot pair a scored embedding with a replacement payload. Coverage also checks inter-batch changes, ties, UTF-16 truncation, yielding, and cancellation recovery.
- Paired actual search measurements on Node 26.8.2 / SQLite 3.53.4, with identical results: 12,000 rows, 768 dimensions, 8 KiB text, top eight: median 504.93 → 441.57 ms (12.5% lower). At 1,000 rows/64 dimensions, 10.64 → 7.23 ms. Small payloads showed a smaller, noisy gain.
- Production extension typecheck and initial repository guards passed. Local extension-test typechecking exhausted host memory; hosted CI must complete that gate. Independent P0–P2 review completed clean.

## Existing Database Compatibility

This changes fallback SELECT projection and result materialization only: no schema, indexes, versions, writers, or embedding format changes. At head `37f1424b`, the original reader from `3f0b8a1` and the updated reader returned identical results in eight cases against the same existing canonical memory-index disk database (1,025 synthetic chunks, 64 dimensions, cache and FTS enabled). Both connections were read-only. All 19 tables' schema/data digests, stored revision, user/schema versions, and the database SHA-256 remained unchanged; both connections recorded zero writes. This separate check establishes reader compatibility; the timing benchmark's minimal schema is not migration proof.

### DIFF
--- a/extensions/memory-core/src/memory/manager-search-orchestration.test.ts
+++ b/extensions/memory-core/src/memory/manager-search-orchestration.test.ts
@@ -250,47 +250,21 @@ describe("memory index", () => {
       );
     }
 
-    const originalPrepare = fields.db.prepare.bind(fields.db);
-    let scannedBatches = 0;
-    const prepareSpy = vi.spyOn(fields.db, "prepare").mockImplementation((sql: string) => {
-      const statement = originalPrepare(sql);
-      if (!sql.includes("SELECT rowid, id, path")) {
-        return statement;
-      }
-      return {
-        all: (...args: Parameters<typeof statement.all>) => {
-          scannedBatches += 1;
-          return statement.all(...args);
-        },
-      } as unknown as typeof statement;
-    });
+    const caller = new AbortController();
+    const abortReason = new Error("caller stopped hybrid memory search");
+    const pending = manager.search("alpha", { signal: caller.signal });
+    setImmediate(() => caller.abort(abortReason));
 
-    try {
-      const caller = new AbortController();
-      const abortReason = new Error("caller stopped hybrid memory search");
-      const pending = manager.search("alpha", { signal: caller.signal });
-      setImmediate(() => caller.abort(abortReason));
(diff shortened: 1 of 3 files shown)


steipete:
Addressed the compatibility proof request in the PR body. This is a read-path optimization with no schema, index, writer, version, or embedding-format change. The old and new readers returned equal results in eight cases on the same existing canonical SQLite file, opened read-only; all 19 tables, versions, revision, and the file hash remained unchanged, with zero writes on both connections. This is separate reader compatibility proof, not a claim that the minimal timing fixture exercises release migrations.

steipete:
Merged via squash.

- Prepared head SHA: [37f1424b8c525d412740f27c4ffff02d5953a221](https://github.com/openclaw/openclaw/pull/145693/commits/37f1424b8c525d412740f27c4ffff02d5953a221)
- Landed commit: [354943cc2f7e1516d90ae9d2ee7420a12a74bac9](https://github.com/openclaw/openclaw/commit/354943cc2f7e1516d90ae9d2ee7420a12a74bac9)

PR #145683, opened 2026-09-12
improve(gateway): avoid rescanning streamed assistant text

steipete:
Related: #145679

## What Problem This Solves

Resolves a problem where long assistant responses spend increasing CPU time processing each streamed text fragment through the OpenAI-compatible HTTP endpoints.

## Why This Change Was Made

The shared text merger records when an event provably appends text. HTTP adapters reuse that fact only when the merge base matches actual wire progress, retaining prefix validation for corrections and withheld output. Responses and session-history SSE frames now use one write per event.

## User Impact

Long delta streams need less CPU while preserving streamed text, correction handling, event order, and SSE bytes. This does not change slow-client buffering policy.

## Evidence

- 160,108 original/candidate merge and wire-output comparisons matched, covering item changes, withheld output, corrections, split Unicode, and scope retirement. Six exact-source SSE framing comparisons also matched.
- Synthetic exact-source 256 KiB response with 64-character deltas and per-event JSON framing: 430.08 -> 2.05 ms median. This measures local text processing, not model latency or whole-Gateway throughput.
- Existing HTTP endpoint tests extended for split leading newlines and incompatible corrections followed by deltas.
- Linux/Node 24.19: all five baseline suites passed (390 tests); all five candidate suites passed (392 tests), including the two added SDK cases. Coverage includes both HTTP adapters, shared stream merging, history SSE, and revocation. Baseline/candidate wall times were 214.21/152.93 seconds; cold/warm transform differences make these validation timings, not an application benchmark.
- Fresh independent review and ClawSweeper review of the exact head found no actionable findings. Formatting, diff checks, and final-head hosted CI passed. Native merge completed as `3f75b33d671017c7ca6cf9ea300279b11e0948c8`.
- After #145696 also landed, all 104 tests in the transcript-path, history HTTP/SSE, and revocation suites passed together on main `354943cc2f7e1516d90ae9d2ee7420a12a74bac9`.

Issue #145679, opened 2026-09-12
Reduce repeated work across Gateway startup, history, and streaming


steipete:
## Latest verified progress — September 15

The campaign now has **42 verified landings** and remains active. The stopping decision in the historical notes below remains superseded.

- #148496 landed catalog snapshot reuse as `ee65c4a3c94041cae606b4ae044978b50d078c7d` (39).
- #148549 landed batched session-list reads as `2e7aa1b7d86b2018ae2aea67010fee086597f430` (40). Six complete-output workloads covered 1, 32 and 128 agents, with same-agent and mixed-agent rows and two revisions. For 128 same-agent rows, observed roster reads fell from 131,456 to 128; mixed-agent reads fell from 90,816 to 256. The 31 focused cases, scoped statics, exact-head CI, review and native gates passed.
- #148616 landed archive classification without unused content revisions as `9c8de0d9289450245120bc24a5aa5266f7187db4` (41). The public-parser workload dropped from 128 observed corpus-statistics queries to zero. All 11 complete outputs, options, reset cutoffs and archive bytes matched; the public-default control still returned 19 revision-bearing entries and performed 16 queries. All 50 focused SDK cases passed.
- #148615 landed memory discovery without unused content revisions as `0d0f5320402357569bd167ca1a690d00aa616bef` (42). Each of five real-manager phases dropped from 16 observed statistics queries to zero, preserving the complete selected source/chunk/search records for 17 paths through append and archive replacement. Its exact-head CI executed 67 manager cases and the required type/lint checks. Shared SDK and separately landed shadow-worker evidence were also inspected.

Both new memory PRs passed independent review, exact-head CI and the native landing workflow. Remote merge commits, ancestry, completed receipts and owned cleanup were verified. Their work comparisons used full `ee65c4a3` source with exact candidate overlays; current-main compatibility was separately source-reviewed. Counts measure the identified statistics executions, not total SQL, latency or memory. Selected manager records do not establish global orphan-row equality.

Further work remains: the redaction allocation candidate, watcher CI diagnosis and remaining metadata-only lifecycle consumers. The watcher’s original 32-case diagnostic passed without reproducing CI's invalid-config failure; the retained 103-path CI group is prepared for a separate diagnostic. The original failure remains unresolved and has not been hidden by a CI retry.

[...]

PR #145696, opened 2026-09-12
improve(gateway): share transcript path checks across history streams

steipete:
Related: #145679

## What Problem This Solves

Resolves a problem where transcript updates repeatedly block the Gateway on the same filesystem path lookup when several session-history streams are open.

## Why This Change Was Made

The existing transcript-path owner lazily reuses the comparison path for each normalized event, and the registered SSE handler consumes that prepared fact. Each new emission resolves afresh; post-authorization path checks still observe current filesystem state.

## User Impact

Multiple history streams need fewer synchronous filesystem calls while receiving the same transcript updates. No path lookup occurs without an interested consumer, and no persistent cache or public event field is added.

## Evidence

- Exact-source Windows/Node 24.19 benchmark with the actual emitter and listeners, ten synthetic file-backed events: 1 stream 1.999 -> 1.976 ms; 10 streams 20.005 -> 2.116 ms; 100 streams 144.796 -> 1.390 ms. This measures event fan-out work, not whole-Gateway throughput.
- Existing SSE delivery test extended to two streams, checking one path resolution per event and subsequent delivery of a persisted message to both. Unit coverage checks lazy resolution, reuse of missing-path fallback, retry on the next event, and identity-only events.
- Linux/Node 24.19: three baseline suites passed 134 tests; four candidate suites passed 135. Coverage includes history delivery, revocation, projection state, and the new event-lifetime path contract. The real-SSE regression fails on original production with four path resolutions instead of two.
- Fresh independent review at P2: no findings. Formatting, diff checks, and final-head hosted CI passed. Native merge completed as `9bebec118fa93eb19b2631c52bbca308a403c24d`.
- Combined with the SSE framing change in #145683, all 104 tests in the transcript-path, history HTTP/SSE, and revocation suites passed on main `354943cc2f7e1516d90ae9d2ee7420a12a74bac9`.

### DIFF
--- a/src/gateway/session-transcript-path.test.ts
+++ b/src/gateway/session-transcript-path.test.ts
@@ -1,0 +1,61 @@
+import path from "node:path";
+import { afterEach, expect, test, vi } from "vitest";
+import * as boundaryPath from "../infra/boundary-path.js";
+import {
+  emitSessionTranscriptUpdate,
+  onInternalSessionTranscriptUpdate,
+} from "../sessions/transcript-events.js";
+import { resolveTranscriptUpdatePathForComparison } from "./session-transcript-path.js";
+
+const unsubscribe: Array<() => void> = [];
+
+afterEach(() => {
+  for (const stop of unsubscribe.splice(0)) {
+    stop();
+  }
+  vi.restoreAllMocks();
+});
+
+test("shares lazy path resolution per event and retries a fallback on the next update", () => {
+  const sessionFile = path.resolve("synthetic-session.jsonl");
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [c0cab1123e287723037a53e7400ccfc818c35684](https://github.com/openclaw/openclaw/pull/145696/commits/c0cab1123e287723037a53e7400ccfc818c35684)
- Landed commit: [9bebec118fa93eb19b2631c52bbca308a403c24d](https://github.com/openclaw/openclaw/commit/9bebec118fa93eb19b2631c52bbca308a403c24d)

PR #145697, opened 2026-09-12
fix(doctor): revalidate stopped systemd service with strict loaded inspection

YangManBOBO:
Related: #145070

## What Problem This Solves

Fixes an issue where `openclaw doctor --fix` on a Linux gateway managed by a
lingering systemd `--user` unit would **always** abort at its final service
revalidation with `Gateway service ownership or manager identity changed;
inspect it before restarting manually.` — after the repairs had already
landed and the gateway had been stopped. The unit was left `inactive` with no
restart attempt, so every repair run ended with a downed gateway until an
operator restarted it by hand.

## Why This Change Was Made

Doctor's post-repair revalidation read the stopped service with
`readGatewayServiceState({ env, requireEffective: true })`. That omits
`requireLoadedCommand`, so on Linux the runtime comes from the non-loaded
`systemctl show` inspection, whose `systemd` snapshot never records
`managerUid`. Since #140339 the Linux revalidation gate
(`inspectManagedGatewayServiceBeforeUpdate` / `matchesStoppedService`)
requires that manager UID — it reads as `undefined`, the inspection degrades
to `unavailable`, and revalidation throws before the restart. The maintenance
stop flow reads the same service with the strict loaded inspection (busctl
`LoadUnit`/`GetUnit` + manager binding), which does record the UID.

The fix makes the final read use the same strict admission read as the
maintenance flow: `requireLoadedCommand: true` plus
`validateEnvBeforeStatusRead: assertGatewayServiceManagementAllowedForUpdate`.
Revalidation then sees the same manager UID (and fingerprint) that the
pre-stop verdict was computed from, and the gateway restarts. No behavior
change for macOS/Windows — their runtime reads already satisfy the
revalidation contract, and the added options mirror the update flow's own
revalidation reads.

## User Impact

Linux users running the gateway as a systemd user unit can run
`openclaw doctor --fix` to completion: repairs are applied and the gateway
comes back up automatically ("Gateway restarted and verified after Doctor
repair."), instead of failing at revalidation and staying stopped.

## Evidence

- **Failing regression test (red):** with the fix reverted, the new test
  `doctor-maintenance.service-revalidation.test.ts` reproduces the exact
  reported failure — `finish()` rejects with
  `GatewayServiceUpdateOwnershipError: Gateway service ownership or manager
  identity changed; inspect it before restarting manually.` and no
  `systemctl restart` is issued.
- **Passing regression test (green):** with the fix, the same test drives a
  mocked Linux systemd user gateway through `beginDoctorMaintenance` →
  stop → repair → revalidation → restart and asserts `systemctl --user
  restart` is invoked. The mock answers real busctl/systemctl shapes,
  including the loaded runtime inspection that records `managerUid: 1000`.
- **Focused suites pass:** the new test plus the three existing
  `doctor-maintenance.*` suites (service-inspection, plugin-preflight,
  schema-preflight) — 15 tests, all passing.
- **Checks:** `tsgo` core + commands-test typecheck, `oxlint`, and `oxfmt`
  all pass on the changed files.
- Root-cause trace: `systemd-runtime.ts` (non-loaded path, no `managerUid`)
  vs `systemd-loaded-runtime.ts` (busctl path, records `managerUid`);
  `update-command-service-maintenance.ts:124` requires it on Linux; the
  doctor final read is the only revalidation caller missing
  `requireLoadedCommand` (update/rollback/recovery all pass it).

### DIFF
--- a/src/commands/doctor-maintenance.service-revalidation.test.ts
+++ b/src/commands/doctor-maintenance.service-revalidation.test.ts
@@ -1,0 +1,287 @@
+import fs from "node:fs/promises";
+import path from "node:path";
+import { afterEach, expect, it, vi } from "vitest";
+import { useAutoCleanupTempDirTracker } from "../../test/helpers/temp-dir.js";
+import { execFileUtf8 } from "../daemon/exec-file.js";
+import { mockSystemAccountHome } from "../daemon/service.test-helpers.js";
+import { mockProcessPlatform } from "../test-utils/vitest-spies.js";
+import { beginDoctorMaintenance } from "./doctor-maintenance.js";
+
+vi.mock("../daemon/exec-file.js", () => ({ execFileUtf8: vi.fn() }));
+// This test targets service revalidation, not database coordination; the
+// private-directory coordinator check is POSIX-mode based and cannot run on
+// a Windows host even with a mocked platform.
+vi.mock("../infra/state-database-coordinator.js", () => ({
+  acquireGatewayLifecycleCoordinator: () => ({ release: () => {} }),
+  acquireStateDatabaseCoordinator: () => ({ release: () => {} }),
+}));
+vi.mock("./doctor-service-repair-policy.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("./doctor-service-repair-policy.js")>()),
+  shouldManageGatewayService: async () => true,
+}));
+vi.mock("../cli/daemon-cli/restart-health.js", () => ({
+  waitForGatewayHealthyRestart: async () => ({ healthy: true as const }),
+  renderRestartDiagnostics: () => [],
+}));
+
+const tempDirs = useAutoCleanupTempDirTracker(afterEach);
+afterEach(() => {
+  vi.restoreAllMocks();
+  vi.unstubAllEnvs();
+});
+
+const json = (type: string, data: unknown) => ({ type, data });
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for the diagnosis — it matches what we found for #145070, and the same repair is already on `main` since this morning in #145220 (commit f3f3d6b1ed6, "doctor --fix leaves systemd-user gateway stopped after repair").

Current `main` (`src/commands/doctor-maintenance.ts`, the post-repair revalidation inside `finishDoctorMaintenance`) reads the stopped service with `requireLoadedCommand: true`, so the Linux revalidation sees the same manager UID the pre-stop verdict was computed from. It goes one step further than this branch: the read happens under `withGatewayServiceOperationLock`, and on Linux it passes `loadForInspection: { managerUid, assertCurrent }` so a collected (garbage-collected) user unit is reloaded for metadata only under live custody of the recorded manager before the launcher is revalidated and restarted. That covers the case where the stopped unit is no longer loaded at all, which a plain `requireLoadedCommand` read would still reject.

Since the landed fix supersedes this change (and its regression test covers the stopped-unit path), I am closing this PR rather than landing a second copy. If you still hit the revalidation error on a build that contains f3f3d6b1ed6, please open a fresh report with the `openclaw doctor --fix` output and `systemctl --user show openclaw-gateway -p LoadState,ActiveState` and we will take it from there.

PR #145699, opened 2026-09-12
fix(tasks): prevent stale flow updates from overwriting current state

steipete:
Related: #144592
Builds on #145652

## What Problem This Solves

Fixes an issue where task-flow controllers could overwrite newer progress or recreate a deleted flow when their in-memory snapshot lagged behind SQLite. Updates inside a larger transaction could also notify observers even when that transaction later rolled back.

## Why This Change Was Made

The existing revision-based update now rereads the canonical flow, compares its revision, and applies the patch within one SQLite write transaction. Callers receive the actual persisted conflict or updated record, and canonical conflict or missing results also reconcile the cached view. The existing transaction owner stages the flow cache for nested read-your-writes, restores it on rollback, and publishes observer events after the outer commit. Queued notifications stop when synchronous observer reentry replaces that committed state.

This is a focused prerequisite for the accepted SQLite worker migration. It preserves the synchronous API, existing patch-validation errors, legacy flow normalization, and current SQLite schema.

## User Impact

Stale flow updates return a revision conflict with current state, and updates to deleted flows return not found. Rolled-back updates no longer appear as committed progress.

## Evidence

- Real SQLite coverage exercises another connection updating or deleting the flow, consecutive nested updates, outer commit/rollback, inner savepoint rollback, and legacy controller normalization.
- 170 focused flow, executor, maintenance, audit, pure-kernel, ACP manager, owner-access, and task-command tests passed, including 11 real SQLite concurrency/rollback cases.
- Five behavior regressions failed on the original implementation; the legacy compatibility case passed in both versions. The observer-reentry case separately failed before its commit-hook repair.
- The full changed gate and independent P2 review passed on the final source.
- The existing in-memory concurrency benchmark settled both synthetic tasks with zero SQLite rows/files and no remaining runtime work. Test and benchmark store fixtures use the existing in-memory store implementation.

### DIFF
--- a/scripts/bench-agent-concurrency-worker.ts
+++ b/scripts/bench-agent-concurrency-worker.ts
@@ -234,13 +234,10 @@ async function configureSpawnRuntime(
         close: () => {},
       },
     });
+    const { createInMemoryTaskFlowRegistryStore } =
+      await import("../src/test-utils/task-registry-store.js");
     flowStore.configureTaskFlowRegistryRuntime({
-      store: {
-        loadSnapshot: () => ({ flows: new Map() }),
-        upsertFlow: () => {},
-        deleteFlow: () => {},
-        close: () => {},
-      },
+      store: createInMemoryTaskFlowRegistryStore(),
     });
     return;
   }
--- a/src/commands/tasks.test.ts
+++ b/src/commands/tasks.test.ts
@@ -964,6 +964,7 @@ describe("tasks commands", () => {
         const deleteFlow = vi.fn();
         configureTaskFlowRegistryRuntime({
           store: {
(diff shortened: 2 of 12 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [6497c5b89effe57c4d7c881d1694b4c670f0fad8](https://github.com/openclaw/openclaw/pull/145699/commits/6497c5b89effe57c4d7c881d1694b4c670f0fad8)
- Landed commit: [882aa555b2e8a0bbf465aab47d0617208d22b469](https://github.com/openclaw/openclaw/commit/882aa555b2e8a0bbf465aab47d0617208d22b469)

PR #145701, opened 2026-09-12
perf(slack): skip unused global table and chart scans

### LINKED ISSUE #145700: Slack reply preparation repeats unused table and chart scans
Slack reply preparation scans table and chart budgets across the full logical reply even when that pass only needs button and select indexes. It also calculates unused offsets when no legacy interactive companion is present. Long replies repeat this work as presentation blocks are appended or rolled into new messages.

Keep reply-wide control indexes, but calculate table/chart budgets only for the current message or an actual legacy companion. Preserve block-ID parsing, mutable raw block aliases, malformed-table handling, message rollover, companion mirroring, fallback text and send ordering.

Synthetic probes through the actual outbound adapter produced identical complete serialized outputs in seven cases. In the table/control case without a companion, the candidate reduced table parses from 26 to 13, table-budget scans from 25 to 12, and chart checks from 219 to 102. These are function invocation counts, not timing or memory measurements; no live Slack messages were sent.

steipete:
Closes #145700

<details>

</details>

## What Problem This Solves

Slack reply preparation repeats table and chart accounting while assigning reply-wide button and select indexes. It also calculates unused offsets when no legacy interactive companion is present.

## Why This Change Was Made

The existing offset resolver now supports an internal controls-only mode for the reply-wide pass. Current-message and present legacy-companion accounting keep the full mode, and absent companions skip the unused calculation.

## User Impact

Slack output stays unchanged: control indexes span the reply, table/chart budgets reset per message, and existing block-ID parsing, mutable aliases, malformed-table handling, companion mirroring, fallback text and send order are preserved. The change adds seven net production lines and 62 net test lines.

## Evidence

Against baseline `0cccc6d68e58ccd5030cd09931c1753bca924c7b`, the same 129 tests across five files passed before and after. The full selected gate passed, and the source-bound P2 review found no actionable issues.

Seven synthetic cases through the actual outbound adapter retained identical complete serialized outputs and hashes, covering plain text, table/control replies, table and chart rollover with companions, malformed tables, and mutated block IDs. For the table/control case without a companion, table parses fell 26→13, table-budget scans 25→12, and chart checks 219→102. These are function invocation counts only; no timing, allocation-byte, RSS or retained-memory improvement is claimed.

The adapter proof uses injected synthetic senders, supplemented by existing send-path mock-client coverage; it is not a live Slack send. No extra build, profile or update proof was selected: package, public SDK, loading/startup, updater and Doctor contracts are unchanged, and the selected extension type/import/lint and owner checks passed.

### DIFF
--- a/extensions/slack/src/blocks-render.ts
+++ b/extensions/slack/src/blocks-render.ts
@@ -233,15 +233,20 @@ function readSlackOpenClawBlockIndex(blockId: string, prefix: string): number | 
 }
 
 /** Resolve existing Block Kit indexes and native-data budgets before appending portable blocks. */
-export function resolveSlackBlockOffsets(blocks?: readonly SlackBlock[]): SlackBlockRenderOptions {
+export function resolveSlackBlockOffsets(
+  blocks?: readonly SlackBlock[],
+  mode: "all" | "controls" = "all",
+): SlackBlockRenderOptions {
   let buttonIndexOffset = 0;
   const dataTableCellCharacterCountOffset =
-    countSlackDataTableBlocksCellCharacters(blocks) ??
-    SLACK_DATA_TABLE_AGGREGATE_CELL_CHARACTERS_MAX + 1;
+    mode === "all"
+      ? (countSlackDataTableBlocksCellCharacters(blocks) ??
+        SLACK_DATA_TABLE_AGGREGATE_CELL_CHARACTERS_MAX + 1)
+      : 0;
   let dataVisualizationCountOffset = 0;
   let selectIndexOffset = 0;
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [7fc873c96b21fbd8d0f8ebcd269d1df841489683](https://github.com/openclaw/openclaw/pull/145701/commits/7fc873c96b21fbd8d0f8ebcd269d1df841489683)
- Landed commit: [77e3071509d8f28cf03ad4bc5825d69a23c97973](https://github.com/openclaw/openclaw/commit/77e3071509d8f28cf03ad4bc5825d69a23c97973)

PR #145702, opened 2026-09-12
fix: preserve streamed Markdown disclosure contents

steipete:
## What Problem This Solves

Fixes an issue where users reading a streamed Chat reply would see text outside a collapsed disclosure when a literal code example contained a closing details tag. Streaming could also add punctuation inside literal HTML examples, or stop completing ordinary Markdown after a quoted or listed example ended.

## Why This Change Was Made

Static rendering and streaming now share raw-block classification, with the existing Markdown parser owning quote/list boundaries. Tail repair preserves literal text and resumes on the following prose. The existing bounded stream cache reuses established literal-line results, and only an unfinished final summary line is eligible for completion.

## User Impact

Disclosures keep their contents inside until expanded. Literal comment, preformatted and declaration examples retain their text, while following emphasis and summaries render normally as the answer streams.

## Evidence

The complete candidate passes 120 selected renderer, scanner and highlighting tests, canonical changed checks, and a fresh Control UI build. Fifteen new assertions fail against the preceding candidate and pass with the container/declaration repair; earlier checks separately established the original containment and raw-punctuation defects. Three existing active-HTML cases were excluded from the local selection.

Chromium exercised the actual Chat composer and streaming reply against a synthetic Gateway. The before captures use the original source; the after captures use this complete candidate. All four full frames were inspected, and the browser, context and source server closed afterward. These images show disclosure containment; public DOM tests cover the additional punctuation and container cases.

| State | Before | After |
| --- | --- | --- |
| Collapsed | Before: reply text outside the collapsed disclosure | After: reply text remains inside the collapsed disclosure |
| Expanded | Before: example and following text are separated | After: example and following text share the disclosure |

The cache check compares cached and uncached rendering of the final implementation: 48 appends produce identical HTML while native raw-range parsing runs once instead of 48 times. It does not claim linear work for long containers growing by new lines. No live-provider or operator Gateway execution is claimed.

Independent review covered the complete patch. A separate bounded comparison used 48 ordinary newline updates ending at 1,553 characters: all HTML matched, and added native block parsing took roughly 0.4–1.2 ms per complete sample. Closed-example totals rose about 2.7–4.1 ms across those 48 updates; these small measurements do not establish behavior near the message cap.

### DIFF
--- a/ui/src/components/markdown-details.test.ts
+++ b/ui/src/components/markdown-details.test.ts
@@ -5,6 +5,12 @@ function htmlFragment(html: string): DocumentFragment {
   return document.createRange().createContextualFragment(html);
 }
 
+const rawHtmlStreamingContexts = [
+  ["comment", "<!--\n</details>\n-->"],
+  ["pre", "<pre>\n</details>\n</pre>"],
+  ["lowercase declaration", "<!doctype\n</details>\n>"],
+] as const;
+
 describe("model-authored details blocks", () => {
   it("renders block-level details with nested markdown", () => {
     const html = toSanitizedMarkdownHtml(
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [19dab1d8782e4164b0d4ad498cd40cd1cef161d7](https://github.com/openclaw/openclaw/pull/145702/commits/19dab1d8782e4164b0d4ad498cd40cd1cef161d7)
- Landed commit: [d111bc21df3c353a917f09b16e3acc24ebac8fc0](https://github.com/openclaw/openclaw/commit/d111bc21df3c353a917f09b16e3acc24ebac8fc0)

PR #145703, opened 2026-09-12
improve(agents): reduce copying of fragmented LSP responses

steipete:
## What Problem This Solves

Fixes an issue where users requesting large results from bundled language servers would wait for avoidable CPU work when responses arrived in many stdout chunks. A synthetic 8 MiB references response took 3.34 seconds to ingest and render when delivered in 4 KiB chunks.

## Why This Change Was Made

The LSP runtime now grows its pending buffer geometrically and retains the validated body boundary until the frame completes. This removes repeated copying of the entire response prefix and repeated header parsing. Completed frames release spare capacity, and incomplete tails detach from completed response storage. Framing limits, Unicode decoding, response ordering, and process lifecycle remain unchanged.

## User Impact

Large fragmented LSP responses complete faster, reducing synchronous work on the Gateway event loop. Pending buffers can reserve spare capacity between chunks; this change makes no claim of reduced process RSS.

## Evidence

The benchmark invokes the registered `lsp_references_bench` tool through `createBundleLspToolRuntime` with a synthetic stdio process. Fixture construction is outside the timer; response ingestion, JSON parsing, and tool-result rendering are inside. Every result is checked byte-for-byte against the expected rendered JSON. Measurements used Windows x64, Node 26.8.2 / V8 14.6.202.34-node.28, with one warmup and four measured replies per cell.

| Approximate response size | Stdout chunk | Before median | After median | Speedup |
| --- | --- | ---: | ---: | ---: |
| 64 KiB | 4 KiB | 0.983 ms | 0.874 ms | 1.12× |
| 64 KiB | 64 KiB | 0.828 ms | 0.823 ms | 1.01× |
| 1 MiB | 4 KiB | 97.558 ms | 12.197 ms | 8.00× |
| 1 MiB | 64 KiB | 14.628 ms | 10.907 ms | 1.34× |
| 8 MiB | 4 KiB | 3336.567 ms | 86.537 ms | 38.56× |
| 8 MiB | 64 KiB | 258.098 ms | 84.369 ms | 3.06× |

Fresh-process 8 MiB whole-frame controls ran in baseline/candidate/candidate/baseline order, with eight measured replies after one warmup per process. Across 16 samples per version, the median was **68.124 → 67.877 ms**. Baseline samples ranged from 47.384–88.133 ms and candidate samples from 54.110–86.566 ms; per-process medians in execution order were 60.49, 68.67, 66.46, and 78.38 ms. These controls show comparable whole-frame performance amid host variance; no whole-frame improvement is claimed. Every output matched, and the committed candidate source was restored byte-for-byte after the control swaps. RSS samples were recorded but varied with GC history, so they do not establish a memory reduction.

- Bundled LSP runtime tests: **30/30 passed**, including consecutive multibyte responses fragmented into 1-byte and 4 KiB chunks.
- Independent review through **P2: scoped-clean**, with no actionable findings.
- Changed formatting, targeted oxlint, production core typecheck, all core test typecheck shards, and `git diff --check` passed.
- `check:changed` stopped at existing unused-export findings in `scripts/crabbox-wrapper-providers.mts` (`canonicalProviderName`) and `scripts/lib/static-extension-assets.mts` (`listPackagedStaticExtensionAssetOutputs`). Both files are byte-identical to base `a53c601cd4075a6839bac3f75f2280eda182b4b8`; the production unused-export scan passed with zero entries. Later changed-gate steps remain unrun, except the two-file oxlint check run separately. No guard or baseline was changed.
- Required hosted `openclaw/ci-gate` passed at exact head `51009850eb560cbe603205e7b4d00a5b7ebe9e4c`, covering the remaining gates.

### DIFF
--- a/src/agents/agent-bundle-lsp-runtime.test.ts
+++ b/src/agents/agent-bundle-lsp-runtime.test.ts
@@ -406,6 +406,46 @@ describe("bundle LSP runtime", () => {
     await runtime.dispose();
   });
 
+  it.each([1, 4096])(
+    "preserves consecutive multibyte tool responses delivered in %i-byte chunks",
+    async (chunkSize) => {
+      configureSingleLspServer();
+      const child = new MockChildProcess("", new Set(["initialize", "shutdown"]));
+      spawnMock.mockReturnValue(child);
+      const runtime = await createBundleLspToolRuntime({ workspaceDir: "/tmp/workspace" });
+      const tool = runtime.tools.find((candidate) => candidate.name === "lsp_hover_typescript");
+      if (!tool) {
+        throw new Error("expected hover tool");
+      }
+      const results = [{ contents: "té🙂".repeat(2048) }, { contents: "次の結果".repeat(1024) }];
+      const requests = results.map((_, index) =>
+        tool.execute(`call-${index}`, {
+          uri: "file:///tmp/workspace/index.ts",
+          line: index,
+          character: 0,
+        }),
+      );
+      const calls = child.receivedMessages.filter(
+        (message) => message.method === "textDocument/hover",
+      );
+      const bytes = Buffer.from(
+        results
+          .map((result, index) =>
+            encodeLspMessage({ jsonrpc: "2.0", id: calls[index]?.id, result }),
+          )
+          .join(""),
+      );
+      for (let offset = 0; offset < bytes.length; offset += chunkSize) {
+        child.stdout.write(bytes.subarray(offset, offset + chunkSize));
+      }
+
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [51009850eb560cbe603205e7b4d00a5b7ebe9e4c](https://github.com/openclaw/openclaw/pull/145703/commits/51009850eb560cbe603205e7b4d00a5b7ebe9e4c)
- Landed commit: [d7de5bc0ad5254cb4d8ce3c67120b82fca86a216](https://github.com/openclaw/openclaw/commit/d7de5bc0ad5254cb4d8ce3c67120b82fca86a216)

PR #145709, opened 2026-09-12
fix: avoid repeated scans when opening historical messages

steipete:
## What Problem This Solves

Fixes an issue where opening a message before a session reset can stall on a large transcript when SQLite has no planner statistics. A synthetic 20,001-event history took 79 seconds to return an anchor page.

## Why This Change Was Made

Use the existing canonical sequence index for historical event identity lookups, preventing repeated session-wide scans. Compute the total and anchor ordinal together, then fetch bounded neighbors around the known physical anchor instead of rescanning with OFFSET. Display visibility, page boundaries, ordinals, and overread context remain unchanged.

## User Impact

Historical message navigation completes promptly even before planner statistics exist. Existing schemas and persisted transcripts require no migration.

## Evidence

- All 22 history owner tests passed on Node 26.8.2, including hidden controls, reset boundaries, malformed/custom payloads, inactive branches, sparse rows, and short pages.
- Same canonical 20,001-event SQLite database, exact page equality: without statistics, original 79,366 ms (one bounded slow sample) → candidate median 30.13 ms / p95 35.18 ms (15 runs).
- With bounded planner optimization already applied, original median 67.00 ms / p95 70.58 ms → candidate 29.96 ms / p95 31.28 ms (15 runs each), about 55% lower median time. Runtime: Node 26.8.2 / SQLite 3.53.4.
- Formatting and diff checks passed. Independent P0–P2 review completed clean. Hosted CI will cover the remaining broad typechecks and repository gates.

### DIFF
--- a/src/config/sessions/session-accessor.sqlite-history-events.test.ts
+++ b/src/config/sessions/session-accessor.sqlite-history-events.test.ts
@@ -704,6 +704,31 @@ describe("SQLite transcript history events", () => {
       "before-tool",
       "first-reset",
     ]);
+    for (const [messageId, maxMessages, ids, seqs, offset, hasOverreadContext] of [
+      ["before-user", 1, ["before-user"], [1], 3, false],
+      [
+        "before-assistant",
+        2,
+        ["before-user", "before-assistant", "before-tool"],
+        [1, 2, 3],
+        1,
+        true,
+      ],
+      [
+        "before-tool",
+        3,
+        ["before-user", "before-assistant", "before-tool", "first-reset"],
+        [1, 2, 3, 4],
+        0,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [82876f905451cfdf0efc2b993481b36c38c50cc9](https://github.com/openclaw/openclaw/pull/145709/commits/82876f905451cfdf0efc2b993481b36c38c50cc9)
- Landed commit: [4bd245ad3f1813b24a4e390b29ae44e38592f024](https://github.com/openclaw/openclaw/commit/4bd245ad3f1813b24a4e390b29ae44e38592f024)

PR #145710, opened 2026-09-12
feat: apply supported plugin installs without restarting the Gateway

steipete:
Builds on landed #145484. Claw batch integration is tracked separately in #145713; related redesign: #135599.

## What Problem This Solves

The Gateway lifecycle supports restart-free changes, but installation has separate CLI, chat and management paths. npm, Git, local paths, archives and marketplace installs need the same live application flow as official and ClawHub packages.

## Why This Change Was Made

All supported sources share one request, source planner, config preparation, persistence owner and runtime application path. The CLI hands live mutations to the serving local Gateway; offline commands still save the installation for startup. Duplicate CLI planning and config-writing paths are removed. Production growth is 29 lines for this expanded capability; tests, generated native models and docs are counted separately.

Source selection also owns recovery eligibility. Early Commander admission and action preflight reuse a lightweight bundled-source selector and carry its selected path privately, preserving the original locator. This fixes bundled recovery for bare IDs and versioned package names while preserving explicit npm overrides, local/file isolation, marketplace handling and source-checkout exclusions. Runtime-only reload remains available with immutable config; install and new consent writes retain their write guards.

## User Impact

Supported installs return an applied Gateway generation, or a saved state when no runtime is available. Existing consent, install policy, authored disabled entries, config includes and narrow plugin-owned config recovery remain enforced. Local artifacts and marketplace registrations must belong to the Gateway host. A lost reply or activation failure never triggers a second local install; the saved state and `plugins reload` provide recovery. The Control UI keeps its existing official/ClawHub choices.

No config option, SQLite declaration, schema version or stored install-record shape changes. The protocol change expands an additive request union; existing official/ClawHub requests remain valid and omitted mode means install. Startup uses the existing persistence readers.

## Evidence

The final correction is `58c718b07f707cc074efd63d9dbff5d0fbf44e82`, tree `7ba094c00c77fc8926d9fc849fda20d33ad5266a`. Bundled recovery has failing-before/passing-after Commander admission and action tests, including opt-out refusals. All 142 recovery/source/consent cases pass, with scoped gates and both runtime/static import checks reporting zero cycles. Read-only reload was separately reproduced and repaired with 123 passing reload/install/consent cases. These address ClawSweeper’s recovery and read-only findings and rank-up moves.

The CLI fixture cleanup preserves assertions and uses the canonical metadata fixture. Its config-sensitive policy tests use the actual loader; synthetic install-output tests retain their explicit artifact fixtures. All 490 shared-helper tests, CLI types and lint pass. The combined final recovery/policy cohort passes 64 tests. Both final corrective patches have clean independent P2 reviews. The original full review’s three findings remain source-dispositioned: canonical unnamed Swift union representation, actual global-before-package lease order, and pre-existing official/ClawHub adoption bookkeeping with newer-install/default-retain safeguards; its raw report is not relabeled clean.

[Linux live run 34641392785](https://github.com/openclaw/openclaw/actions/runs/34641392785) recorded 14 install-source, consent and recovery scenarios through the real Gateway and filesystem on assembled tree `18b3d63c20bd444aab64980c03ecf1c6132e4517`. Synthetic package services were used; processes, ports and the lease were closed. This remains historical source-qualified proof. Earlier 405 focused tests, protocol/types/Knip checks, normal build and canonical checks likewise retain their recorded source identities. Exact-head CI is required before landing.

Telegram-specific proof is unavailable: the Telegram E2E skill requires an authenticated Convex CLI or broker pair, absent locally and from the standard Testbox environment. The dedicated Telegram workflow cannot provision this synthetic install scenario. No accounts, credentials or workflow permissions were changed, and no Telegram event proof is claimed.


steipete:
Merged via squash.

- Prepared head SHA: [58c718b07f707cc074efd63d9dbff5d0fbf44e82](https://github.com/openclaw/openclaw/pull/145710/commits/58c718b07f707cc074efd63d9dbff5d0fbf44e82)
- Landed commit: [6e259355b89e390f18a7183249ba964774da60e4](https://github.com/openclaw/openclaw/commit/6e259355b89e390f18a7183249ba964774da60e4)

PR #145715, opened 2026-09-12
test: make runtime-neutral async assertions

steipete:
## What Problem This Solves

Two direct-Bun unit cases asserted runtime scheduling or engine wording rather than the behavior they intended to protect. Bun formats the native BigInt serialization error differently from V8, and a provider-side WebSocket close can be observed before the client message callback runs.

## Why This Change Was Made

The context-engine case still matches the complete quarantine log structure but accepts the JavaScript engine's text in the serialization slot while requiring it to identify BigInt. The realtime transcription case now waits for the client-visible final transcript after the provider closes instead of treating server-side close ordering as proof that the client callback already ran.

## User Impact

This changes tests only. It keeps the same fallback, quarantine, final-transcript, and single-finalize contracts while making their synchronization and native-error assertions valid under Node and Bun.

## Evidence

- Direct custom Bun and Node context-engine suite: 78/78 passed each.
- Direct custom Bun and Node realtime WebSocket suite: 29/29 passed each.
- Before the change, Bun failed the exact V8 BigInt phrase and observed an empty transcript immediately after provider close.
- Test-audit: both edits remove implementation coupling; they retain exact observable outcomes, add no production seam, and do not duplicate stronger coverage.
- Independent P0-P2 autoreview: scoped clean at 0.94 confidence.

### DIFF
--- a/src/context-engine/context-engine.test.ts
+++ b/src/context-engine/context-engine.test.ts
@@ -22,6 +22,7 @@ import {
   withPluginRegistrationContext,
 } from "../plugins/runtime.js";
 import type { UserTurnTranscriptAdmissionReceipt } from "../sessions/user-turn-transcript.types.js";
+import { escapeRegExp } from "../shared/regexp.js";
 // ---------------------------------------------------------------------------
 // We dynamically import the registry so we can get a fresh module per test
 // group when needed.  For most groups we use the shared singleton directly.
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [0ce2f4864ac6a3e1d3d4eb6e7051fb0aab1467b8](https://github.com/openclaw/openclaw/pull/145715/commits/0ce2f4864ac6a3e1d3d4eb6e7051fb0aab1467b8)
- Landed commit: [cf77d10572f86f8de59bd0ee7ca5cdb45c0ca9e3](https://github.com/openclaw/openclaw/commit/cf77d10572f86f8de59bd0ee7ca5cdb45c0ca9e3)

PR #145716, opened 2026-09-12
perf(tasks): reduce repeated SQLite read work

steipete:
Related: #145652

## What Problem This Solves

Repeated task and flow reads rebuild and compile the same Kysely query trees even when the shared SQLite owner already reuses their native statements. The duplicate-admission owner lookup also prepares identical SQL on every call.

## Why This Change Was Made

Retain lazy prepared-binding closures in one weak query bag per physical connection in each kernel. Existing task point, runtime/source, and flow point reads continue through the normal synchronous executor with fresh parameters. The duplicate-owner lookup uses the existing bounded native statement cache while keeping its exact `NOT INDEXED` SQL, ordering, `.all()` execution, and error behavior.

The existing cache owner retains responsibility for native statements, active-call reentry, invalidation, and close. No schema, index, transaction, public API, dependency, or persistence change is introduced. Public actor/view query additions are a separate change and can extend these same query bags.

## User Impact

Repeated reads spend less time rebuilding query trees. Database values and ordering remain fresh; intentional duplicate-admission table scans are preserved. This does not activate a worker or claim an end-to-end Gateway latency improvement.

## Evidence

Synthetic canonical SQLite datasets used 1,000 and 10,000 tasks, changing IDs/scopes, and seven alternating paired rounds. Exact SQL, parameters, row results, and query plans matched. At 10,000 tasks, median complete kernel-call timings including decoding were:

| Read | Before | After |
| --- | ---: | ---: |
| Task point hit | 17.96 µs | 6.44 µs |
| Task point miss | 12.30 µs | 1.28 µs |
| Runtime/source scope, 10 rows | 68.50 µs | 52.95 µs |
| Runtime scope, 100 rows | 513.87 µs | 500.77 µs |
| Flow point | 7.24 µs | 5.14 µs |

Warm query compilations fell from 100 to zero per 100 kernel calls. Native prepares on those Kysely reads remained two per 100 calls; their gain comes from compilation reuse. The intentional owner scan reduced native prepares from 100 to two, but scanning dominates at larger row counts.

- 53 kernel/store cases pass, including supplied-connection isolation, changing point/runtime/source bindings, missing reads, empty source filters, cache-enabled execution, and compound rollback. The final fixture transaction wrapper also passed its focused case.
- Six existing ordinary helper contracts pass for fresh/reentrant bindings, active statement reuse, step-time failure, original error reporting, and close behavior.
- A synthetic lifetime probe confirms the closed handle is collected with query bags present, and closed task/flow reads retain the normal database-closed error.
- The full changed-file gate passes, including core and core-test typechecks, lint, export scans, schema/storage guards, and import-cycle checks. Final P2 review and root source review are clean.
- Rebased patch-identically onto main containing #145652; the touched kernel files, query helpers, and row normalization implementation are unchanged in the intervening base.

- [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34678798148) passed after one unchanged retry recovered two unrelated CLI startup timeouts. No source, timeout, assertion, or routing changes were made for that retry.
- The cached `.all()` primitive also preserved the original step error, read successfully afterward, and kept outer bindings during nested SQL callback reentry.

### DIFF
--- a/src/tasks/task-flow-registry.store.kernel.ts
+++ b/src/tasks/task-flow-registry.store.kernel.ts
@@ -10,6 +10,7 @@ import {
   executeSqliteQuerySync,
   executeSqliteQueryTakeFirstSync,
   getNodeSqliteKysely,
+  prepareSqliteQuerySync,
 } from "../infra/kysely-sync.js";
 import { normalizeSqliteNumber } from "../infra/sqlite-number.js";
 import type { DB as OpenClawStateKyselyDatabase } from "../state/openclaw-state-db.generated.js";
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [1d8a0e47e800ed05861efb83e22d62035a0f02a6](https://github.com/openclaw/openclaw/pull/145716/commits/1d8a0e47e800ed05861efb83e22d62035a0f02a6)
- Landed commit: [cbb082f743b69846034fc4767816dc8cb6e13996](https://github.com/openclaw/openclaw/commit/cbb082f743b69846034fc4767816dc8cb6e13996)

Issue #98542, opened 2026-07-01
[Feature]: Conversation identity modes for personal, team, and shared audiences


steipete:
### Summary

Add one closed conversation identity/audience mode for personal, organization/team, and external/shared digital-employee contexts.

### Problem to solve

OpenClaw can route conversations to configured agents and already carries channel/account/team, sender-owner, allowlist, per-sender tool-policy, workspace, instruction, and session facts. The scoped conversation capability profile in #98512 and draft #98536 prepares those facts for runtime consumers, but it does not yet decide when a conversation may inherit a personal agent identity and workspace.

Without one canonical decision, an unbound shared room or non-owner direct message can silently follow the default personal agent route. Later consumers then have to independently decide whether personal instructions, memory assumptions, and tools are appropriate for that audience.

### Proposed solution

Build the smallest identity-mode layer on the prepared capability profile from #98536:

- personal mode for private/direct owner contexts that may use the configured personal agent identity, workspace, and instructions;
- organization/team mode only when an explicit conversation binding selects a non-default service agent, while retaining existing group and per-sender policy inputs;
- external/shared mode for guest, unbound shared, or otherwise unresolved audiences, denied before personal session context or workspace instructions are loaded.

Reuse existing route bindings, agent identities, channel/account/team metadata, sender-owner facts, `users`/allowlist admission, `toolsBySender`, session classification, and workspace/instruction resolution. Add no new public config key if these existing contracts fully express the decision.

### Alternatives considered

- Add broad RBAC or role configuration. This duplicates existing bindings and tool policies and expands the public security surface before the identity decision is proven.
- Infer trust from channel names, display names, or provider-specific rules. These are mutable or transport-owned and do not form a safe generic contract.
- Keep the decision in each channel. That duplicates policy and makes personal fallback behavior inconsistent across transports.

### Impact

Affected: operators using OpenClaw for personal assistants, organization/team operations, or shared digital-employee conversations.

Severity: high for mixed personal/shared deployments because an unresolved audience must not inherit personal identity or workspace context.

Frequency: each inbound direct or shared conversation routed to an agent.

Consequence: a single auditable decision point, explicit service-agent selection for team use, and fail-closed behavior for unbound guests.

### Evidence/examples

Duplicate and existing-solution preflight:

- #98512 / draft #98536 provide the prerequisite prepared conversation capability profile.
- #49183 and #49567 cover narrower per-session/per-binding tool policy surfaces.
[...]

Issue #19929, opened 2026-02-18
[Feature]:  Shared sessions across multiple group/channel/thread chats


demarlik01:
### Summary

When the same agent is bound to multiple Discord channels (or Slack channels), each channel creates an isolated session (`agent:<agentId>:discord:channel:<channelId>`). 

There is currently no way to selectively merge specific channel sessions while preserving organizational channel separation. 

### Problem to solve

In multi-agent setups where the same agent participates in multiple topic-based channels
(e.g., `#planning`, `#dev`, `#art`, `#qa`), 
each channel gets its own isolated session. This creates several problems:

1. **Context fragmentation**: The same agent in channel A has zero context of discussions that happened in channel B. This leads to contradictory responses — e.g., an agent researches a topic for an hour in one channel, then gives conflicting advice in another channel because it has no memory of the prior conversation.

2. **Concurrent write conflicts**: Multiple sessions of the same agent can simultaneously write to the same workspace files (e.g., `memory/YYYY-MM-DD.md`). Since the queue system only serializes within a single session, parallel sessions can overwrite each other's changes.

3. **Memory is not real-time context**: While `memory/` files can share durable notes across sessions, they cannot share live conversational context. An agent must explicitly write to memory and the other session must re-read it — there's no automatic sharing of the current conversation.

4. **Threads also isolate**: Discord threads and Slack threads are routed as separate sessions from their parent channel, further fragmenting context even within the same topic.

### Proposed solution

A configuration that maps a set of channel/thread identifiers to a single shared session key,
similar to how `session.dmScope: "main"` collapses all DMs into one session.

### Config shape proposal

```json5
{
  session: {
    // Option A: Global channel groups (applies to all agents)
    channelGroups: {
      "studio-unified": [
        "discord:channel:general-channel-id",
        "discord:channel:planning-channel-id", 
        "discord:channel:dev-channel-id",
        "discord:channel:art-channel-id", 
      ],
    },
  },
}
```

```json5
{
  // Option B: Per-agent channel groups (more granular control)
  agents: {
    list: [
      {
        id: "agent-1",
        session: {
          channelGroups: {
            "team-channels": [
              "discord:channel:general-channel-id",
              "discord:channel:slack-channel-id",
            ],
          },
        },
      },
    ],
  },
}
```

### Expected behavior

- Messages from any channel in a group route to the same session key (e.g., `agent:agent-1:group:studio-unified`).
- The queue serializes all messages within the group, preventing concurrent write conflicts.
[...]

PR #145729, opened 2026-09-12
docs(acp): document current-conversation ACP binding for channels without a binding adapter

### LINKED ISSUE #79281: [Feature] Default ACP thread-binding preset (or opt-in config flag) — third-party channels currently re-implement ~870 LOC each
## Problem

`sessions_spawn(runtime="acp", mode="session", thread=true)` fails with `thread_binding_invalid` on every channel that does not register a `SessionBindingAdapter`. Today this includes the most-used third-party channel — **WeChat / openclaw-weixin** — and several first-party channels (slack, line). The result on those channels is a degenerate "agent replies once and stops" experience: the spawned ACP session runs one turn, never continues, and the user has no way to keep the conversation going.

This bug just bit me again live with `openclaw-weixin → claude-agent-acp`:

```
session: d71a7917-3a3c-477d-9402-5ac70d812178
spawnedBy: agent:main:openclaw-weixin:<bot>:direct:<sender>
runtimeMs: 133153
usage:    0 tokens
stopReason: stop
contents: 1 placeholder text message ("我先读取 ppt-master skill 的 strategist 定义...")
            then status=done, no tool use, no follow-up
```

## Snapshot of who has thread-binding today

Reading `upstream/main` `aeba1d6b47`:

| Channel       | SessionBindingAdapter LOC | Status |
|---------------|---------------------------|--------|
| imessage      | 46                        | ✓      |
| bluebubbles   | 46                        | ✓      |
| feishu        | 330                       | ✓      |
| matrix        | 820                       | ✓      |
| telegram      | 922                       | ✓      |
| discord       | ~2300                     | ✓      |
| slack         | —                         | ✗      |
| line          | manifest declares it, **no adapter registered** — empty promise | ✗ |
| **openclaw-weixin** (third-party, very large user base) | Tencent/openclaw-weixin#56 (open since 2026-04-11, ~870 LOC) | ✗ |

The shared base (`src/infra/outbound/account-scoped-conversation-bindings.ts`, ~350 LOC) and the manifest seam (`src/channels/plugins/types.adapters.ts:837-843`'s `conversationBindings.createManager`) already exist, but the integration cost per channel is still steep enough that:

1. New channel plugins ship without thread binding (LINE has the manifest field but no adapter — the manifest is an empty promise)
2. Third-party plugin maintainers re-invent ~870 LOC to wire it up (Tencent/openclaw-weixin#56)

## Proposal

Add `createDefaultDmThreadBindingManager(...)` preset to `openclaw/plugin-sdk/thread-bindings-runtime` so plugin authors can register a working adapter in ~10 LOC instead of 870.

Default binding key: `(accountId, conversationId, senderId)` — three-tuple, **not** `(accountId, conversationId)`.

### Why per-sender as the default

OpenClaw already has three layers of group-safety guardrails for normal sessions:

| Layer | What it does |
|-------|--------------|
| `groupPolicy` / `groupAllowFrom` / `requireMention` | Who can trigger the bot |
| `toolsBySender` | Which tools different senders get |
| `contextVisibility: "all" / "allowlist" / "allowlist_quote"` | Whose history goes into the prompt |

For ACP `mode="session"`, layer (c) is bypassed: ACP state lives in the persistent runtime process (cwd, modified files, tool result cache) — not in the per-turn context load. So `contextVisibility: "allowlist"` cannot retroactively isolate group members from each other once they share an ACP session. **Default per-sender keying is the equivalent guardrail at the binding layer**, applied before that state ever gets shared.

For 1:1 channels behavior degrades to `(accountId, conversationId)` — same as imessage/bluebubbles today.

For groups, sharing one ACP session across all members must be opt-in (`allowSharedGroupBinding: true`).

### Open question for review

I lead with the preset proposal but I'd happily take a less aggressive landing if it helps merge:

- **(A) Preset with per-sender default + `allowSharedGroupBinding` opt-in** — my preference. New channels and third-party plugins get a working adapter for free; existing channels can adopt incrementally.
- **(B) Preset + opt-in config flag** — preset exists in plugin-sdk but plugins must explicitly call `enableDefaultThreadBinding()` to opt in. No default behavior change. **This is the safest landing if (A) is too aggressive — even just shipping the preset as a documented helper is a big win for plugin authors.**
- **(C) Decide on review.** Issue body sketches (A); the call is the maintainer's.

## Related (would benefit from this preset existing)

- #53548 "Decouple mode=session from thread binding requirement" — adjacent; even with #53548 landing, channels without an adapter still can't do `mode=session, thread=true`. Both fixes are useful and land in different places.
- Tencent/openclaw-weixin#56 — concrete proof of the duplication cost. If this preset lands, that PR can shrink from ~870 LOC to ~10.
- #60737, #42986 — related channel-specific feature requests that would naturally consume the preset.

## Secondary proposal: manifest sanity check

Channels that declare `conversationBindings` in their manifest but never register an adapter at runtime should emit a `manifest-registry` warn-level diagnostic. LINE is the example today — declares `defaultTopLevelPlacement: "current"` but has no adapter, so the manifest is an empty promise.

## Environment

- `upstream/main aeba1d6b47`
- The bug reproduces on every openclaw 2026.5.x release I've tested (2026.5.5, 2026.5.6, 2026.5.7)
- ACP runtime: `@agentclientprotocol/claude-agent-acp@0.31.4`

lyfuci:
Closes #79281

## What Problem This Solves

Resolves a problem where third-party channel plugin authors conclude they must port a full binding adapter to get ACP
thread binding, and operators cannot tell which channels support it today.

`docs/tools/acp-agents/bindings.md` says "Thread binding support is adapter-specific" and lists only Discord and
Telegram under "Thread-supporting channels". Core has shipped a generic current-conversation binding path since
v2026.3.28: any channel built with `createChatChannelPlugin(...)` already gets
`sessions_spawn({ runtime: "acp", thread: true })` and `/acp spawn <agent> --bind here` with `current` placement and no
custom adapter, and a third-party plugin opts in with one field. Nothing on the ACP pages points there — the
plugin-side field is documented only in `docs/plugins/sdk-channel-plugins.md`.

The two "bindings are unavailable" rows in `troubleshooting.md` attribute the failure to a missing adapter capability
and tell the reader to "move to a supported adapter/channel", which hides the actual enablement path.

For context on the linked issue: I filed #79281 about this area, and its description no longer matches current main.
It claims Slack and LINE lack adapters and that `thread=true` fails there, and proposes an SDK preset plus a manifest
sanity check. Core's generic fallback already covers those channels, and the one plugin I hit this with failed because
it was a bare `ChannelPlugin` literal with no `conversationBindings` block rather than for want of an adapter. The
preset and manifest proposals are withdrawn; the docs gap is the part that held up, so that is all this PR addresses.

## Why This Change Was Made

Rewrites the "Thread-supporting channels" accordion to describe both paths — channel binding adapter versus generic
current-conversation binding — states the generic binding's scope, and links the plugin-side field. Adjusts the
neighbouring sentence that called thread support "adapter-specific", and the two troubleshooting rows so they name
what actually enables the feature.

Non-goals: no behavior change, no new SDK export, and no sender-scoped isolation for group bindings. The generic
binding key has no sender component, so a group chat shares one bound ACP session; the docs now say so plainly rather
than leaving it to be discovered. Changing that key is a persistent-store and security question, so it stays with the
conversation-identity discussion in #98542 / #19929.

The adapter list is left as Discord and Telegram. Other bundled channels do register adapters, but I have not run
`--bind here` on them, so the text points readers at each channel's page instead of asserting a list I cannot vouch
for.

## User Impact

- Operators can see that Slack workspace installations and other bundled chat channels bind the current conversation
  with no extra setup.
- Plugin authors learn that current-conversation ACP bindings need one field, not a binding adapter.
- Anyone hitting `Thread bindings are unavailable for <channel>.` or
  `Conversation bindings are unavailable for <channel>.` is now told what enables the feature.
- The group-chat caveat is documented: every member of a group shares the bound ACP session.

## Evidence

Docs-only, verified against `main` at 58a6953549.

```
pnpm format:docs   Docs formatting clean (1282 files)
pnpm check:docs    markdownlint: 0 issues in 0 files
                   Docs MDX check passed (1295 files)
                   checked_internal_links=13439  broken_links=0
```

Claims checked in the current tree rather than carried over from the issue:

- `src/plugin-sdk/core.ts:829-832` — `createChatChannelPlugin` sets
  `conversationBindings.supportsCurrentConversationBinding: true` by default.
- `src/infra/outbound/session-binding-service.ts:247-250` — `if (!adapter && !genericCapabilities?.bindSupported)`:
  bind and capabilities fall back to the generic store when no adapter is registered. The file header describes it as
  multiplexing channel adapters and the generic current-conversation bindings.
- Slack is on the generic path and gated per account: `extensions/slack/src/channel.ts:590-592` sets
  `isCurrentConversationBindingSupported: ({ accountId }) => isSlackWorkspaceInstallation(accountId)`, which is why
  the text says "workspace installations" rather than "Slack".
- The linked anchor exists: `docs/plugins/sdk-channel-plugins.md:417`
  "### Account-scoped conversation binding support".


lyfuci:
Pushed one follow-up commit: the SDK link now points at `/plugins/sdk-channel-plugins/sessions-and-bindings#account-scoped-conversation-binding-support` directly.

The ClawSweeper review noted the previous target resolves through a compatibility anchor on the index page. Since the point of this change is to send readers to the plugin-side field, an extra hop through an index weakened it — and the compat anchor is the kind of thing that gets pruned later. `check:docs` still reports `broken_links=0`.

PR #135290, opened 2026-09-01
fix(memory): preserve wiki results in combined search under load

Agent-Aurelius:
Closes #104719

## What Problem This Solves

Fixes an issue where `memory_search` with `corpus: "all"` could return memory hits plus a false Wiki-unavailable warning on large compiled vaults, even though the same Wiki query completed successfully on its own. The timed-out Wiki scan also kept reading files after the tool had already returned.

## Why This Change Was Made

The existing 15-second tool deadline remains unchanged. The repair forwards its cancellation signal through the public corpus-supplement contract, Memory Wiki activation and source validation, digest/fallback lookup, directory walks, and page reads. Exhaustive body recall is still preserved while the call is on-budget; a raw-text candidate filter avoids expensive YAML/Markdown parsing for pages that cannot match.

## User Impact

Combined memory + Wiki recall can return healthy Wiki hits from large vaults without a false timeout caused by parse overhead. Genuine stalls still fail closed with the existing per-corpus warning, and cancellation now stops later Wiki I/O instead of leaving orphan work behind.

## Evidence

- Read-only real-vault proof on 5,784 Markdown pages / 62,999,844 bytes, forcing the exhaustive fallback path with the issue query. Baseline: 11,712 / 10,094 / 11,690 ms. Candidate on the current rebased head: 5,763 / 2,818 / 3,181 ms. All six runs returned 10 results with the same ordered-path digest (`429438066d6f0b9a`).
- `node scripts/run-vitest.mjs extensions/memory-core/src/tools.citations.test.ts extensions/memory-core/src/memory-get-corpus.test.ts extensions/memory-wiki/src/corpus-supplement.test.ts extensions/memory-wiki/src/query.test.ts extensions/memory-wiki/src/compiled-cache.integration.test.ts` — 122 passed.
- `pnpm build` — passed, including all 147 public Plugin SDK subpaths.
- `pnpm test:contracts:plugins` — passed on the current head (46 files / 1,111 tests); the broader channel + plugin contract suite also passed before the mechanical current-main rebase.
- `node scripts/check-changed.mjs -- <changed paths>` — passed all selected core, plugin, docs, typecheck, lint, boundary, dead-export, and guard lanes.
- Plugin SDK surface check and source/plugin import-boundary check — passed.
- Fresh `autoreview --mode branch --base origin/main` — scoped clean.

[...]

PR #145732, opened 2026-09-12
fix(memory): large Wiki searches time out and keep reading files

### LINKED ISSUE #104719: [Bug]: memory-wiki supplement exhaustive fallback ignores tool deadline
## Bug type

Behavior bug (incorrect timeout behavior without process exit)

## Beta release blocker

No

## Summary

The memory-wiki supplement reparses every remaining Markdown page when digest-selected candidates underfill `maxResults`, ignoring the memory tool deadline and causing `corpus="wiki"` and `corpus="all"` timeouts.

## Steps to reproduce

1. Register a compiled-wiki vault large enough that a full Markdown scan exceeds 15 seconds.
2. Run `memory_search` with `corpus="wiki"`, `maxResults` greater than the number of digest-selected hits, and a query that underfills the requested count.
3. Observe that `searchWikiCorpus()` scans the remaining vault to fill the result set.
4. Observe the memory tool return `memory_search timed out after 15s` while the local scan may continue.

## Expected behavior

The wiki supplement should honor the caller's deadline/abort signal, return digest-routed or partial results within the tool budget, and reserve exhaustive body scanning for an explicit exhaustive search path.

## Actual behavior

On a 77 MB, 1,220-file wiki vault, digest-routed lookup completed in 77 ms, but ordinary multi-result searches fell through to full-vault parsing and took 25-27 seconds. Both `corpus="wiki"` and `corpus="all"` could therefore exceed the fixed 15-second memory-tool deadline.

## OpenClaw version

2026.7.1-beta.5

## Operating system

macOS 26.5.2 arm64

## Install method

Homebrew global package with launchd-managed gateway

## Model

gemini-embedding-2-preview for builtin memory; local compiled-wiki supplement

## Provider / routing chain

OpenClaw memory_search -> memory-wiki corpus supplement -> local Markdown vault

## Logs

```shell
local full-vault benchmark: 25-27s
digest-routed candidate lookup: 77ms
```

## Impact and severity

- Affected: agents using compiled-wiki supplements through `memory_search`.
- Severity: moderate to high; wiki and combined durable recall become unavailable for ordinary queries.
- Frequency: deterministic when digest candidates underfill the requested result count on a sufficiently large vault.
- Consequence: tool timeout, hidden wiki context, and continued host work after the caller has already timed out.

## Additional information

A narrow local mitigation added a supplement-only option to skip exhaustive fallback. Afterward, multi-result wiki recall completed in 55-66 ms and combined-corpus recall completed below 8 seconds. Direct `wiki_search` retained exhaustive fallback for explicit body-search use. A durable upstream fix should propagate abort signals into the scanner and use parsed-page caching or local FTS for fallback.

Related but distinct: #92633 tracks all-corpus aggregation latency, and #72717 proposes an FTS index for wiki search. This report specifically concerns supplement underfill triggering an uncancelled full-vault fallback even for `corpus="wiki"`.

Baumus:
Closes #104719

Successor to #135290, incorporating the cancellation and fallback-scan work from that PR on current `main` and addressing its outstanding [plugin-input compatibility finding](https://github.com/openclaw/openclaw/pull/135290#issuecomment-5496791754). Related earlier approaches: #105098 and #105161. Existing matching and deadline behavior are preserved.

<details>

Allow edits from maintainers is enabled.

</details>

## What Problem This Solves

Fixes an issue where users searching a large Memory Wiki through `memory_search` would lose Wiki recall at the tool deadline while the underlying vault scan continued reading and preparing pages. Long queries and digest underfill can take the full-scan path even when primary memory search is healthy.

## Why This Change Was Made

The caller's cancellation signal now reaches the Wiki supplement, safe directory walks, page reads, cold activation, and source-generation validation. An optional second execution-context argument keeps the established first argument unchanged for strictly validating third-party supplements, for both search and get.

Fallback search avoids constructing full page summaries for nonmatching pages. The conservative filter also checks decoded YAML metadata and generated-block removal, retaining matches that a raw-text-only filter would lose. Existing safe file boundaries, visibility filters, digest routing, result scoring, warm-cache ownership, and the 15-second deadline remain in place. No new settings, dependencies, or storage migrations are introduced.

## User Impact

Wiki searches perform less unnecessary page processing, and cancellation stops new filesystem work instead of leaving a full scan running after the response. Healthy sibling-corpus results remain available. Already-started safe reads may finish: the current safe-reader API has no in-flight cancellation parameter, so the fix stops scheduling further reads and preserves its boundary checks.

## Evidence

<details>
<summary>Data-model compatibility proof — passed</summary>

This repair introduces no stored-data format change and requires no new migration. Against the current PR base `347fce25598215278fb84d5c549f58b8628c36d7`, the PR does not change `src/state/**`, `src/plugin-state/**`, or the Wiki cache, compiler, Markdown format, or configuration implementations. In `log.ts`, the serialized log fields and writer are unchanged; the change makes existing reads and source hashing cancellable. The cancellation signal is execution context, never a persisted field.

A disposable driver ran three separate isolated processes against the same synthetic vault and real SQLite-backed plugin Blob store. The original production implementations were restored from the pinned pre-fix revision, verified byte-for-byte against Git. Only the write phase initialized and compiled new content; both later processes reused the existing publication without recompilation or migration.

| Phase | Revision | Observed result |
| --- | --- | --- |
| Original writer | `54ebdbbfd2a141f7ffcf220aa4ab3155dd5cf7fa` | Initialized two synthetic Markdown inputs and compiled a durable Wiki publication; 21 vault files including generated scaffold/reports. |
| Updated reader | `291a414b17a87b699563eecc03d158d9a72972a0` | Reopened that publication. The real Wiki corpus supplement returned exactly equal search results, direct-page content, and claim-ID lookup results. Cold source validation was then cancelled, and a subsequent request recovered successfully using the same publication. |
| Original reader after updated reader | `54ebdbbfd2a141f7ffcf220aa4ab3155dd5cf7fa` | Reopened the same persisted data and returned the original results again. |

Assertions passed for identical vault/publication identities, source-generation hash, complete decoded cache snapshot, ordered search results, page/claim reads, and all 21 vault-file content hashes. SQLite schema definitions, `user_version`, and logical rows in all 104 tables were unchanged after the updated-reader and original-reader phases; integrity checks returned `ok`. This is a minimal fixture: one cache Blob row and one schema metadata row are populated, with other tables empty. Blob bytes and metadata are included in the row comparison; SQLite file-layout/WAL bytes are not treated as a data-format contract.

Recorded fingerprints, identical across the three phases:

- Source generation: `0e89ce15878984697f3a0424d1a417839bcfda0707987091ad33d09b879dc494`
- Decoded cache generation: `a268d30d70eecd8670bfa77819962fbcc4e008a3323f9b360a29a1080da7fb50`
- Canonical SQLite schema/row/integrity fingerprint: `814ee874bb15eefa8468058d9016e1849ec042e748822d3cf3cbd05e213c7a89`

Scope: this proves compatibility across the introduced Wiki fix, not every OpenClaw release upgrade. Runtime phases were pinned to the revisions above. The later main-merge head `cc6a1e842371e641fe74730dee5e37fa4cd59aec` has an identical `extensions/memory-wiki/**` tree to the tested candidate; the Wiki cache/compiler/format implementations, plugin Blob facade/SQLite implementation, and state DB owner were also verified unchanged between those heads. This is source-equivalence evidence for carrying the focused proof forward, not a claim that the entire merged head was executed.

No credentials or real user content were available inside the test isolation. No database, raw log, private path, or production fixture is attached.

</details>

- GitHub follow-up: [`check-lint` passed](https://github.com/openclaw/openclaw/actions/runs/34680589297/job/103518534168). A separate [`checks-node-compact-large-15` failure](https://github.com/openclaw/openclaw/actions/runs/34680589297/job/103518535795) is `ENOTEMPTY` while removing a temporary state directory after `subagent-reset.lifecycle.test.ts` / `rpc reset accounts for requester/controller-owned children (failed=false)`. That exact test [passed in the preceding run](https://github.com/openclaw/openclaw/actions/runs/34679316035/job/103514950624); its source and cleanup helper are unchanged by this PR. This points to intermittent cleanup contention, not a new lint failure. A targeted job rerun was requested but GitHub rejected it with `403: Resource not accessible by integration`; an account with Actions write access is needed to rerun the failed job and dependent CI gate. No test retries, weaker assertions, or CI-rule changes were added.
- CI follow-up at `291a414b`: narrowed the synthetic test's `readFile` path before matching, fixing `typescript/no-base-to-string` without changing the expected read sequence or cancellation assertions. The repository's type-aware lint wrapper passes for the changed file (0 errors); all 25 compiled-cache integration tests and formatting pass. Fresh incremental `autoreview`: scoped clean at P0–P2.
- 170 focused tests passed across corpus coordination, search/get tools, Wiki query and safe reads, supplement routing, directory walking, and compiled-cache lifecycle. Coverage includes long multiword queries, body-only hits outside digest candidates, YAML escapes/numeric/null/block values, generated related blocks, visibility, pre-aborted calls, and interrupted cold validation followed by successful retry.
- Both strict-input cancellation regressions fail against original production code at `54ebdbbfd2a141f7ffcf220aa4ab3155dd5cf7fa` because the supplement receives no cancellation context; they pass with this change.
- Synthetic runtime proof used 11,623 Markdown source pages (20,023,006 bytes) and three known long-query body hits through the real registered Wiki supplement, corpus coordinator, and deadline owner. Both versions returned the same three paths. Full page summaries fell from 11,623 to 3. After the caller-abort response, additional file opens fell from 11,607 to 0; underlying work settled approximately 13 ms after abort rather than 11.8 seconds later.
- Observed search times were 11.6 seconds before and 9.6 seconds after. These are illustrative under concurrent compiler load, not a timing guarantee. This synthetic baseline itself stayed below 15 seconds. The large fixture exercised an initialized vault without a populated digest; separate regressions exercise compiled-digest fallback. No model, live Gateway, or production vault was used for the runtime proof.
- `pnpm tsgo:core`, `pnpm tsgo:extensions`, and `pnpm tsgo:extensions:test` passed, along with formatting, lint, and `pnpm plugin-sdk:surface:check`.
- `pnpm test:contracts:plugins`: 48 files / 1,128 tests passed in an isolated checkout with no production state or credentials mounted.
- Conflict-free merge check against `main` at `b6f7b2a4c82c079ce6be9a486400be3226e0eb61`.
- Independent `autoreview` of the complete candidate against its pinned base: scoped clean at P0–P2.
- Full build and repository-wide tests were not run; validation targets the changed memory behavior and public plugin contract.


Baumus:
@obviyus Could you take a maintainer look at this P1 Memory Wiki fix? Current-head CI is green, ClawSweeper found no blocking patch defect, and its review explicitly says the supplied compatibility evidence resolves the previous upgrade-proof concern. However, the generated “Before merge” section still repeats that resolved item.  

The proof covers original writer → updated reader → original reader against the same SQLite-backed Wiki publication, with unchanged schema, logical rows, stored content, and search/get results. Is this sufficient to clear the remaining owner gate and merge independently before the next regular release?

PR #145375, opened 2026-09-11
fix: preserve edited watchers through cron recovery

shakkernerd:
Closes #145336

<details>
<summary>Additional instructions</summary>

Keep **Allow edits from maintainers** enabled for this PR.

</details>

## What Problem This Solves

Fixes an issue where users who edited a running watcher could lose the replacement's saved condition state, or have a replacement `once` watcher disabled, when the Gateway recovered a completed run after a crash. Script payload replacements could also receive obsolete saved state.

## Why This Change Was Made

Record each outstanding run's state-writer retirement in a receipt-owned companion table, atomically with the relevant edit. Normal finalization and startup recovery read that decision from the exact receipt, including a receipt already closed by an agent-owner edit. Edits that return to the original definition or state remain protected; queued runs still evaluate the state present when execution starts.

The first-use table has one receipt-ID key and follows the existing receipt's deletion through a foreign key. The released receipt table, state schema 17, and retention policy stay unchanged. Missing historical retirement facts are not inferred; older readers do not enforce the new protection. Canonical automation and schema docs explain the behavior and downgrade limit.

## User Impact

Acknowledged condition, script, and saved-state edits survive completion recovery while completed run history remains intact. Unchanged `once` watchers still stop after success. Name-only and failed edits preserve normal completion behavior.

## Evidence

Validated source: `7c7ca979a840cec7c61b73f159ea787800488c68`.

- The native SQLite/service crash campaign reproduced the defect on `abe0065574d313ecde402a953e6fb3d22dfe3ece`; ordinary completion and unchanged-watcher controls passed.
- The same eight-process campaign passes, including all three deliberate crash boundaries. The regression suite produces exactly 11 expected failures on the original source, with 50 controls passing. All 353 distinct affected cases pass with the fix: 86 focused tests and 267 sibling/schema cases. The 24 existing store cases were rerun after the final hook correction and are included in that total.
- Released v2026.9.4 source with its own frozen dependencies passes the updater preflight and opens the corrected shared-state database. Candidate reopening preserves the retirement fact and all 103 nonmetadata tables in that database. On a separate copy, 65 real released-reader receipt completions retain the newest 64 receipts and cascade deletion of the retired receipt's companion row; candidate reopening confirms that deletion.
- Independent review of the complete fix and final code correction found no actionable P0–P2 issues. Local publication preflights passed; the last correction also passed the existing owner-fence test through the production mutation-hook factory.
[...]

Issue #145336, opened 2026-09-11
[Bug]: Cron startup recovery can disable an acknowledged watcher replacement


shakkernerd:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

After an acknowledged edit replaces a running cron watcher's condition, a crash between terminal-task persistence and scheduler-row completion lets startup recovery overwrite the replacement's state and disable it using the superseded run's result.

### Steps to reproduce

A condition watcher is a saved automation that checks a condition on a schedule and runs its payload when the condition fires. With `once: true`, it stops after the first successful fired payload. See the [condition watcher contract](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/docs/automation/cron-jobs/schedules.md#L83-L105).

For illustration, a watcher checks build 123 every minute, prepares a report when it finishes, then stops. While that report runs, the user edits the same watcher to monitor build 456 and receives confirmation. The old report should finish while the edited watcher remains armed for build 456.

The controlled reproduction used synthetic state and injected clock, condition evaluator, and payload executor functions through existing injection points. It did not poll builds or generate reports. Registered cron handlers, real `CronService` startup, and native SQLite job/task/receipt storage ran normally:

1. Through registered `cron.add`, create an enabled every-60-seconds job with condition A and `once: true`. Hold its command payload at the executor boundary.
2. Invoke registered `cron.run` with `mode: "due"`. Wait for A to fire, payload admission, and a durable active run receipt. The default manual force mode bypasses the condition and does not reproduce this sequence.
3. While A remains in flight, invoke registered `cron.update` to replace the condition with B, also with `once: true`, and explicitly set B's `triggerState`. Wait for acknowledgement and verify that B is durable and enabled.
4. Return a successful result for A. Kill the disposable process after A's exact task commits as `succeeded`, before the separate cron-row completion commits.
5. Start real `CronService` in a fresh process over the same isolated database, then read the job through registered `cron.get`.

The deterministic crash uses a connection-local temporary SQLite trigger, installed after the acknowledged edit, that sends SIGKILL to its own process immediately before the selected cron-row update. It does not fabricate job/task/receipt records or change persisted schema. Before startup recovery, the database contains B's saved state, A's terminal task, and A's still-active receipt.

### Expected behavior

[...]

PR #145734, opened 2026-09-12
fix: recover the correct cron run when timestamps collide

shakkernerd:
## What Problem This Solves

Cron watchers can recover a different completed run when successive runs share a timestamp, the scheduler's completion write fails, and an owner edit closes the pending run's execution receipt. A retired predecessor can prevent a successful once-only watcher from completing, or the wrong run can overwrite replacement trigger state.

## Why This Change Was Made

Pending scheduler state retains its exact receipt identity until reconciliation. Recovery and late state edits use that identity; stale recovery proposals and concurrent marker updates compare it with the running timestamp. Marker cleanup clears both fields, and receipt pruning protects the pending terminal receipt within the existing 64-receipt limit.

The storage scope follows the receipt-owned retirement design in #145375 for #145336: the existing cron owner keeps the same run associated with its pending reconciliation after execution authority closes. The added private `runningReceiptId` lives in `cron_jobs.state_json`; no SQL table, column, schema version, config option, or public state-patch field is added. Current public job projections omit it. Legacy rows without it keep their existing recovery fallback, and ordinary pruning releases the protected receipt and its retirement row after reconciliation.

The documented downgrade boundary requires completing active runs **and pending scheduler reconciliation** on the current build to preserve edited watcher state. A terminal task or receipt alone does not establish that boundary. Older writers do not maintain the new association or its retention protection.

## User Impact

Recovery applies the pending run's result and its own state-edit protection even when another run has identical timestamps. Owner edits still close execution authority immediately.

## Evidence

Proof targets commit `b28b0d22f81bd294885d79e90dbe7f7cd45999f8`, tree `001b1030d843960ed7fe032fd283712deb0bee20`, on one retained [Blacksmith Testbox runner](https://github.com/openclaw/openclaw/actions/runs/34679692212) using Linux, Node 24.19.0, and pnpm 12.3.4.

- All 217 files and 3,099 tests pass across the three maintained cron CI groups, with zero failed or skipped cases and exact coverage of the original native cron input set. The unchanged receipt read-budget assertion passes.
- All 11 focused cases pass across `service.recovery-identity`, `service/shared-store-runtime`, and `store/run-receipt-retention`. Replaying those tests against base `3f75b33d671017c7ca6cf9ea300279b11e0948c8` produces six assertion failures and five passing controls. Coverage includes colliding and advancing timestamps, either run's retirement, edits after receipt closure, stale proposals, cross-process marker updates, and bounded pruning before and after reconciliation.
- Three native `CronService`/SQLite probes pass for retired-pending, retired-predecessor, and unedited runs. Each fails on the base revision at the final recovered enabled/state assertion. Admission, updates, receipts, tasks, history, and restart recovery use their real owners; the wall clock, trigger evaluator, and payload are synthetic.
- All seven cross-version phases pass with v2026.9.4 (`3a9d69db306cd7f081e06254cb89c4bcc14a7107`) and separate frozen dependency installs: released legacy pending state, candidate upgrade recovery, fresh candidate pending state, released read-only preflight, candidate reconciliation, released writable reopen, and candidate reopen. Shared-state schema-17 preflights report `exact`; reconciliation runs no payload; job, task, receipt, and retirement rows remain byte-equal across both reopen phases. Released writable access begins only after the candidate clears the pending marker and ID. This proves the documented cron/shared-state boundary; agent databases, live providers, the full updater, and writable downgrade while reconciliation is pending are outside this proof.
- All 29 selected changed-path gates pass, including full core and core-test types, dead exports, formatting, lint, ratchets, conflict/diff hygiene and storage guards. The dependency deadcode scan passes. The existing public projection test passes. Independent engineering review and autoreview are clean; [ClawSweeper reports Platinum with no findings](https://github.com/openclaw/openclaw/pull/145734#issuecomment-5644348231) for this commit.
- The earlier single-worker monolithic cron run exhausted its unchanged 4 GiB worker limit. An untouched base revision reproduces the same native worker-memory error with the original common file order. The maintained three-group run above preserves the full selected file set, worker limits and assertions; no retry, timeout increase or test exclusion was added.
- [CI attempt 2](https://github.com/openclaw/openclaw/actions/runs/34681280777/attempts/2) passes after one authorized rerun of the failed jobs on the unchanged tested merge. The original seven-file CLI group passes all 275 tests; the rejected-update case passes in 1,236 ms with the expected error and unchanged state-preservation assertions. The original 60-second timeout remains unattributed: the rerun uses Node 24.20.0 on a four-CPU GitHub runner, versus Node 24.19.0 on a two-CPU Blacksmith runner. Separate exact-main and tested-merge replays on Node 24.19.0 also pass all 275 tests. Assertions, timeouts and the two-worker CLI limit are unchanged; no CI exception is used.

### DIFF
--- a/docs/reference/database-schemas/versioning.md
+++ b/docs/reference/database-schemas/versioning.md
@@ -46,8 +46,16 @@ creates a retirement row in the same transaction as the job edit. This includes
 an exact receipt already closed by an agent-owner edit but still awaiting run
 reconciliation. Normal completion and restart recovery preserve the replacement's
 state while retaining the old run's history. The first eligible edit creates the
-table; queued edits do not retire a future evaluation. Existing receipt pruning
-also deletes its retirement row.
+table; queued edits do not retire a future evaluation. The job's private runtime
+state retains the exact running receipt ID until scheduler reconciliation, including
+when an agent-owner edit closes the receipt first. Recovery and later state edits
+use that association even when run timestamps collide. Receipt pruning preserves
+that pending receipt; ordinary history retains its existing 64-receipt bound and
+deletes retirement rows with their receipts.
+
+Rows written before this association was recorded retain their legacy recovery
+fallback. The association adds no SQL table, column, or schema version and is
+omitted from public job state and the public state-patch schema.
 
 A missing table or row means no recorded retirement; earlier edits cannot be
 reconstructed from the final job definition. Older compatible readers ignore the
--- a/docs/reference/database-schemas/versioning.md
+++ b/docs/reference/database-schemas/versioning.md
@@ -54,13 +54,15 @@ that pending receipt; ordinary history retains its existing 64-receipt bound and
 deletes retirement rows with their receipts.
 
 Rows written before this association was recorded retain their legacy recovery
-fallback. The association adds no SQL table, column, or schema version and is
-omitted from public job state and the public state-patch schema.
+fallback. The association adds no SQL table, column, or schema version. Current
+builds omit it from public job state and the public state-patch schema.
 
 A missing table or row means no recorded retirement; earlier edits cannot be
 reconstructed from the final job definition. Older compatible readers ignore the
-companion but do not enforce this protection. Finish active runs before downgrading
-if their edited watcher state must be preserved.
+companion but do not enforce this protection. To preserve edited watcher state,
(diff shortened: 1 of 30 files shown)


shakkernerd:
Fixed: cron recovery applies the correct run and preserves its watcher-state edits when timestamps collide.

PR #145739, opened 2026-09-12
perf(memory): use native FTS rank ordering

steipete:
## What Problem This Solves

Memory keyword search orders by a BM25 expression alias. SQLite consequently sorts all matching candidates before applying LIMIT, so common queries slow down as the index grows.

## Why This Change Was Made

Use FTS5's hidden rank column so SQLite can produce score order directly. Pin the rank mapping to `bm25()` for each query, preserving current scores even if an existing database has a customized persistent mapping. Substring fallback and path search retain their existing behavior.

## User Impact

Faster keyword searches over large memory indexes, with the same results, source filtering, scores, and stored data.

## Evidence

- 79 memory search and orchestration tests passed. Regressions fail on the original query because it examines 32 matching candidates for a three-result request; candidate work is bounded for both unicode61 and trigram. Coverage preserves source filtering, ties, and customized rank configuration.
- Paired actual search entry point, canonical memory schema, Node 26.8.2 / SQLite 3.53.4, ten alternating measured rounds with full result equality: 30,000 rows, median 66.99 → 42.31 ms; 100,000 rows, 228.64 → 135.13 ms. These synthetic common-term workloads show 37–41% lower latency.
- EXPLAIN QUERY PLAN uses the FTS rank-order path and removes the temporary sort. SQLite documents this optimization for limited result sets: https://www.sqlite.org/fts5.html#sorting_by_auxiliary_function_results
- Production extension types, scoped type-aware lint, formatting and light guards passed. Independent P0–P2 review validated clean. Broad extension-test types remain for hosted CI.
- No schema, index, version, writer, or persistent rank setting changes; query-local ranking preserves default BM25 and leaves existing configuration intact.

### DIFF
--- a/extensions/memory-core/src/memory/manager-search.test.ts
+++ b/extensions/memory-core/src/memory/manager-search.test.ts
@@ -932,6 +932,73 @@ describe("searchPathKeyword", () => {
   });
 });
 
+describe("searchKeyword ranked limits", () => {
+  it.each(["unicode61", "trigram"] as const)(
+    "stops examining scoped candidates after filling the %s result window",
+    async (ftsTokenizer) => {
+      const { db } = createMemorySearchDb({ ftsTokenizer });
+      try {
+        for (let index = 0; index < 64; index++) {
+          insertKeywordFixture(db, {
+            id: `chunk-${index}`,
+            path: `memory/${index}.md`,
+            text: "common keyword",
+            source: index % 2 === 0 ? "memory" : "sessions",
+          });
+        }
+        let examined = 0;
+        db.function("observe_keyword_candidate", () => {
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [23d0e086897b5a26527b066464c76afbdc196470](https://github.com/openclaw/openclaw/pull/145739/commits/23d0e086897b5a26527b066464c76afbdc196470)
- Landed commit: [68c9a1d8f65042c3fc72cb18f05cbe84bc1d36b3](https://github.com/openclaw/openclaw/commit/68c9a1d8f65042c3fc72cb18f05cbe84bc1d36b3)

PR #145741, opened 2026-09-12
fix(gateway): paired-node Claude continuations retain Gateway context

### LINKED ISSUE #145636: [Bug]: Paired-node Claude continuation reads an expired request context instead of the durable Gateway resolver
### Bug type

### Beta release blocker

No

### Summary

A Claude CLI continuation on a paired node can fail after its originating request ends because the node runtime reads the direct plugin request context instead of resolving the durable current Gateway context.

### Steps to reproduce

1. Enable native Claude CLI sessions and terminals on a paired node.
2. Create a Claude CLI session from the Control UI and receive the first terminal response.
3. Close the terminal so the native session can be adopted by the Web chat.
4. After the originating Gateway request has completed, submit a continuation from the Web composer.
5. Observe a node-runtime availability failure even though the paired node remains connected and advertises the required Claude capability.

This was observed on OpenClaw 2026.9.2. A compatibility change that first calls the request scope's durable `resolveGatewayContext()` restored continuation. The signed 2026.9.4 source still reads `getPluginRuntimeGatewayRequestScope()?.context` directly at the relevant paired-node runtime preparation sites.

### Expected behavior

Detached or later continuation turns should resolve the currently active Gateway context through the scope's durable resolver, then locate the connected node and its advertised capability.

### Actual behavior

The direct request context can be retired after the originating request completes. The paired node is consequently treated as unavailable even though it is still connected, and the Claude continuation fails before model execution.

### OpenClaw version

Observed on 2026.9.2 (3928bad); the affected direct-context access remains in signed 2026.9.4 (3a9d69d).

### Operating system

Debian GNU/Linux 12 (bookworm), Linux x86_64

### Install method

Docker using the official image by immutable digest, with a derived compatibility layer; paired node also runs OpenClaw on Linux x86_64

### Model

N/A — paired-node runtime preparation fails before the Claude CLI turn starts

### Provider / routing chain

Control UI → Gateway adopted session → bundled Anthropic session-catalog plugin → paired OpenClaw node → Claude CLI

The node was connected and the same native Claude session was readable from the catalog. The compatibility change affects only the guarded paired-node runtime call sites associated with the existing errors `Paired node needs claude-cli-skills-v1` and `Gateway node runtime unavailable`.

### Logs

```shell
Gateway node runtime unavailable
```

Depending on which guarded call site is reached, the user-facing failure can instead report that the paired node needs `claude-cli-skills-v1` despite the connected node advertising it.

Signed 2026.9.4 source still reads the request scope's direct context:

https://github.com/openclaw/openclaw/blob/3a9d69db306cd7f081e06254cb89c4bcc14a7107/src/gateway/node-claude-skill-runtime.ts#L50-L67

The compatibility behavior that restored the continuation was equivalent to:

```ts
const scope = getPluginRuntimeGatewayRequestScope();
const gateway = scope?.resolveGatewayContext?.() ?? scope?.context;
```

The same bounded replacement was required at the second node-runtime call site carrying `Gateway node runtime unavailable`. With it installed, a fresh Claude terminal turn, terminal closure, Web adoption, and a second Web response completed successfully.

### Impact and severity

- Affected: users continuing native Claude CLI sessions hosted by a paired node.
- Severity: high for that workflow because continuation is blocked before the CLI starts.
- Frequency: reproduced on the tested detached continuation path before the compatibility change.
- Consequence: a healthy paired node is misclassified as unavailable, forcing users back to a terminal or requiring a local patch.

### Additional information

No exact duplicate was found. #141252 also concerns durable reply authority, but it reports busy-session automatic replies across messaging channels and a different user-visible failure; this report is specifically the paired-node Claude runtime's direct request-context lookup.

ZengWen-DT:
Closes #145636

## What Problem This Solves

Fixes an issue where users continuing a native Claude CLI session on a paired node could receive a node-runtime availability or capability error after the originating Gateway request had ended, even though the node remained connected.

## Why This Change Was Made

Gateway request scope now owns one canonical live-context lookup: it prefers the durable resolver whenever one exists and otherwise uses the direct request context. Both paired-node runtime paths use that lookup, and the existing in-process dispatcher delegates to it, so detached continuations no longer bypass their retained resolver. A resolver that reports its owner retired still returns no context; there is deliberately no fallback to stale request state.

This does not change configuration, protocol, storage, or non-node execution behavior.

## User Impact

Adopted Claude sessions can continue through a healthy paired node after the request that created them has completed. If the detached owner has actually retired, the operation continues to fail closed rather than reviving expired state.

## Evidence

AI-assisted implementation and validation.

### Exact-head Gateway/node production-boundary proof

Run against `4d47d0faa9fd4c5338c8107099e7c8ecb215a2db`:

```shell
node scripts/run-vitest.mjs src/gateway/pr145741-production-boundary.proof.test.ts --config test/vitest/vitest.gateway.config.ts
```

The temporary proof fixture:

1. Started the actual Gateway server with isolated config/state.
2. Loaded a startup plugin through the normal plugin loader and captured its real `GatewayRequestContext` during an authenticated RPC.
3. Connected a separately signed WebSocket node identity, then completed normal device pairing and node-command approval.
4. Entered `withPluginRuntimeGatewayContextResolver` after the capture RPC had completed, verified direct `scope.context` was absent, and called the production `invokeNodeClaudeCliRun` implementation changed by this PR.
5. Repeated the call with a resolver that returned `undefined` while deliberately leaving stale direct context present, to prove a retired owner cannot fall back and dispatch.

Observed output:

```text
PR145741_PROOF:{"head":"4d47d0faa9fd4c5338c8107099e7c8ecb215a2db","resolverOnly":{"directContext":false,"connectedNode":true,"dispatched":1,"ok":true},"retiredOwner":{"staleDirectContextPresent":true,"resolverReturnedContext":false,"additionalDispatches":0,"result":{"ok":false,"error":{"code":"UNAVAILABLE","message":"Gateway node runtime unavailable"}}}}

Test Files  1 passed (1)
Tests       1 passed (1)
```

This exercises the real Gateway startup/plugin-loading boundary, device and node pairing, `NodeRegistry`, WebSocket transport, node command policy, and this PR's production resolver lookup. The node protocol peer returned a synthetic successful Claude-command payload; this proves dispatch and retirement behavior, not live Claude model inference or a full terminal-close/Web-composer UI walkthrough.

### Supporting checks

- Scope-level probe inside a real resolver-only scope: legacy direct context was `null`; the new lookup returned `{ "marker": "live-context" }`; after owner retirement it returned `null`.
- `node scripts/run-vitest.mjs src/plugins/runtime/gateway-request-scope.test.ts --config test/vitest/vitest.plugins.config.ts`: 1 file, 7 tests passed, including resolver-only and retired-owner coverage.
- The paired-node fixture exercises its existing continuation scenarios under the same resolver-only detached scope used by production.
- `git diff --check`: passed.
- `node scripts/check-changed.mjs -- <changed files>`: conflict-marker gate passed; the local max-lines ratchet did not complete after 9m49s and was interrupted. Full repository checks remain delegated to PR CI.
- Claude CLI `2.1.268` is installed, but live model inference was not required for this lookup repair and was not claimed.

### DIFF
--- a/src/gateway/node-agent-cli-runtime.test.ts
+++ b/src/gateway/node-agent-cli-runtime.test.ts
@@ -9,11 +9,9 @@ const mocks = vi.hoisted(() => ({
 }));
 
 vi.mock("../plugins/runtime/gateway-request-scope.js", () => ({
-  getPluginRuntimeGatewayRequestScope: () => ({
-    context: {
-      getRuntimeConfig: mocks.getRuntimeConfig,
-      nodeRegistry: { get: mocks.get, invoke: mocks.invoke },
-    },
+  getPluginRuntimeGatewayRequestContext: () => ({
+    getRuntimeConfig: mocks.getRuntimeConfig,
+    nodeRegistry: { get: mocks.get, invoke: mocks.invoke },
   }),
 }));
 
--- a/src/gateway/node-agent-cli-runtime.ts
+++ b/src/gateway/node-agent-cli-runtime.ts
@@ -2,7 +2,7 @@
 import { randomUUID } from "node:crypto";
 import type { SystemRunApprovalPlan } from "../infra/exec-approvals.js";
 import { NODE_AGENT_CLI_CLAUDE_RUN_COMMAND } from "../infra/node-commands.js";
-import { getPluginRuntimeGatewayRequestScope } from "../plugins/runtime/gateway-request-scope.js";
+import { getPluginRuntimeGatewayRequestContext } from "../plugins/runtime/gateway-request-scope.js";
 import {
   invokeNodeClaudeSkillRuntime,
   type NodeClaudeSkillRuntime,
@@ -29,7 +29,7 @@ export async function invokeNodeClaudeCliRun(params: {
   signal?: AbortSignal;
   skillRuntime?: NodeClaudeSkillRuntime;
 }): Promise<NodeInvokeResult> {
-  const context = getPluginRuntimeGatewayRequestScope()?.context;
+  const context = getPluginRuntimeGatewayRequestContext();
   if (!context) {
     return {
       ok: false,
--- a/src/gateway/node-claude-skill-runtime.test.ts
+++ b/src/gateway/node-claude-skill-runtime.test.ts
@@ -15,7 +15,7 @@ import { resolveNodeHostGatewayPlatformIdentity } from "../node-host/gateway-pla
 import { decodeClaudeCliNodeRunParams } from "../node-host/invoke-agent-cli-claude-params.js";
 import { runClaudeCliNodeCommand } from "../node-host/invoke-agent-cli-claude.js";
 import type { NodeInvokeRequestPayload } from "../node-host/invoke-types.js";
-import { withPluginRuntimeGatewayRequestScope } from "../plugins/runtime/gateway-request-scope.js";
+import { withPluginRuntimeGatewayContextResolver } from "../plugins/runtime/gateway-request-scope.js";
 import type { OpenClawPluginNodeHostCommandIo } from "../plugins/types.js";
 import { createDeferredCore } from "../shared/deferred.js";
 import {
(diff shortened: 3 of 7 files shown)


ZengWen-DT:
@clawsweeper re-review

PR #145744, opened 2026-09-12
perf(sessions): batch FTS suffix removal in one scan

steipete:
## What Problem This Solves

Replacing a long transcript suffix deletes FTS rows in 400-ID chunks. FTS metadata is unindexed, so each chunk scans the entire agent search table, including unrelated sessions.

## Why This Change Was Made

Bind larger ID sets once through the existing SQLite string-set helper. Small sets retain their direct bound-array path. The existing suffix owner still performs the same session-scoped deletion and projection update in the same transaction.

## User Impact

Faster transcript rewrites when a substantial searchable suffix is removed from an agent with a large search index. Stored formats, retention and transaction boundaries are unchanged.

## Evidence

- All 40 suffix/search owner tests passed. The expanded existing regression fails on the original two-scan behavior at 405 removed rows and verifies generation rotation, raw/identity/active/FTS consistency, sibling sessions with identical IDs, embedded NULs, literal escapes, Unicode, and quotes.
- Paired actual projection owner on the same rollback-restored canonical synthetic schema, seven alternating rounds, Node 26.8.2 / SQLite 3.53.4: 4,000 removed rows with 30,000 unrelated rows, median 92.16 → 18.15 ms; with 100,000 unrelated rows, 238.76 → 32.83 ms. Deleted and retained rows match in every round. These measure the projection update, not complete transcript rewrite latency.
- Core production types, scoped lint and 25 light repository guards passed. Independent P0–P2 review validated clean. Broad test types remain for exact-head hosted CI.
- No schema, index, version, configuration, retention, or transaction-boundary changes.
- The first hosted run timed out after 120 seconds in the unrelated `run-vitest-profile` case "rejects main --pool before evaluating config" (job 103517159157). Its test, owner and configuration are unchanged. The focused candidate passed on Windows in 680 ms and the unchanged baseline passed in WSL in 1,071 ms. Only that failed job was rerun once; exact-head run 34680088058 then completed successfully.

### DIFF
--- a/src/config/sessions/session-accessor.sqlite-transcript-suffix.test.ts
+++ b/src/config/sessions/session-accessor.sqlite-transcript-suffix.test.ts
@@ -851,20 +851,57 @@ describe("SQLite exact transcript suffix replacement", () => {
     }, events);
   });
 
-  it("rotates generation while updating raw, identity, active, and FTS rows", async () => {
-    await withRewriteFixture(({ db, snapshot, scope }) => {
-      const before = snapshot();
-      replaceTranscriptSuffixForTest(scope, rewriteEvents, rewriteEvents.slice(0, 2));
+  it.each([1, 405])(
+    "removes %i searchable suffix rows in one scan while preserving other sessions",
+    async (count) => {
+      const specialIds = ["suffix-\0-end", "suffix-\\u0000-end", "suffix-🦀", 'suffix-"quote'];
+      const ids = Array.from(
+        { length: count },
+        (_, index) => specialIds[index] ?? `suffix-${index}`,
+      );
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [be0930924f80074cc31887f11fcd3210a0294776](https://github.com/openclaw/openclaw/pull/145744/commits/be0930924f80074cc31887f11fcd3210a0294776)
- Landed commit: [90e909c3db12b43b543d17f1fed18dc913595921](https://github.com/openclaw/openclaw/commit/90e909c3db12b43b543d17f1fed18dc913595921)

PR #145748, opened 2026-09-12
fix(sessions): reduce repeated store checks when listing sessions

steipete:
<details>

</details>

## What Problem This Solves

Fixes an issue where users loading the Sessions list pay repeated filesystem ownership checks for every returned row when many sessions share one SQLite store.

## Why This Change Was Made

The sharing refresh already groups exact session reads by physical store, but resolving each row's store repeated the same ownership and realpath work before grouping. Reuse the existing synchronous batch target-resolution cache during that planning step. Each new batch resolves current ownership again; incognito routing, logical session ownership, read-only admission, and per-row failure handling remain with their existing owners.

## User Impact

Sessions list requests perform less synchronous work when many returned rows share a store. This does not change session visibility or stored data.

## Evidence

A synthetic benchmark exercised the registered `sessions.list` handler with real SQLite stores, 2 configured agents, 3,000 sessions, and a 100-row page. Across 15 unprofiled repetitions:

| --- | ---: | ---: |
| Median request duration | 36.97 ms | 34.17 ms |
| Median sharing refresh duration | 9.94 ms | 5.63 ms |
| Store target resolutions per sharing batch | 100 | 1 |
| Realpath calls per sharing batch | 200 | 2 |

These are local synthetic measurements, not a production speedup claim. The disposable benchmark is excluded from the product change.

The regression failed before the fix with three ownership resolutions instead of one. It also replaces the physical database owner between batches and verifies that the next batch reads the replacement store. Existing corrupt-row coverage now uses an explicit store locator to exercise cached resolution while retaining partial-result errors.

Validation: all 248 selected permanent tests passed across the read-only accessor, target resolution, combined-store, incognito transcript, Sessions list ownership, and session creation suites. The full changed-file gate passed, including production/test typechecking, lint, formatting, export scans, and boundary guards. Independent P2 review found no actionable issues. [Hosted CI](https://github.com/openclaw/openclaw/actions/runs/34680365048) passed on exact head `4a50daaffc4fbb9adfa064c94a6005878d9d098e`, including the required `openclaw/ci-gate`. ClawSweeper reviewed that same head with no actionable findings.

### DIFF
--- a/src/config/sessions/session-accessor.readonly.test.ts
+++ b/src/config/sessions/session-accessor.readonly.test.ts
@@ -1,4 +1,5 @@
 import fs from "node:fs";
+import path from "node:path";
 import { DatabaseSync, type StatementSync } from "node:sqlite";
 import { expectDefined } from "@openclaw/normalization-core";
 import { afterEach, describe, expect, it, vi } from "vitest";
@@ -38,6 +39,7 @@ import {
   upsertSessionEntryCore,
 } from "./session-accessor.js";
 import { ensureTranscriptSessionRoot } from "./session-accessor.sqlite-transcript-state.js";
+import * as sqliteTargets from "./session-sqlite-target.js";
 
 const tempDirs: string[] = [];
 const autoTempDirs = useAutoCleanupTempDirTracker(afterEach);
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [4a50daaffc4fbb9adfa064c94a6005878d9d098e](https://github.com/openclaw/openclaw/pull/145748/commits/4a50daaffc4fbb9adfa064c94a6005878d9d098e)
- Landed commit: [347fce25598215278fb84d5c549f58b8628c36d7](https://github.com/openclaw/openclaw/commit/347fce25598215278fb84d5c549f58b8628c36d7)

PR #145749, opened 2026-09-12
fix(process): route supervisor PTYs through Bun sidecar

steipete:
## What Problem This Solves

Interactive `exec` PTYs still loaded `@lydell/node-pty` directly from the process supervisor. Under Bun, the native module's nonblocking tty stream closes on `EAGAIN`, so all three real PTY interaction cases failed or timed out even though OpenClaw's terminal subsystem already had a working Node sidecar.

## Why This Change Was Made

The supervisor now uses the shared terminal PTY adapter. Node keeps the direct native path; Bun uses the existing Node worker. The adapter carries the explicit terminal name, environment, OOM wrapper, cancellation fence, process-tree signals, and listener disposal through the same handle contract.

Binary key input is base64-framed across JSON IPC so zero and non-UTF-8 bytes stay exact. The supervisor revalidates ownership after the asynchronous sidecar handshake and kills a newly created stale PTY before rejecting. Tests now mock the terminal-owner boundary, and the interactive fixture counts submitted line delimiters rather than assuming Node and Bun split stdin writes into identical `data` events.

## User Impact

Bun-hosted `exec` sessions support PTY output, send-keys, submit, raw binary/Unicode input, cursor queries, and key modes. Node behavior and Windows native signal handling remain unchanged. No dependency is added.

## Evidence

- Direct custom Bun real exec PTY suite: 3/3 passed three consecutive times; before the fix all three cases failed.
- Node real exec PTY suite: 3/3 passed three consecutive times.
- Direct custom Bun and Node adapter/sidecar suites: 39/39 passed with 5 platform skips each.
- Async construction cancellation regression: passed under Bun and Node.
- `node scripts/check-changed.mjs`: passed, including core and test TypeScript gates.
- Test-audit: the supervisor unit test now mocks its actual owner boundary; native node-pty behavior remains covered below it. The stdin fixture removes event-chunk coupling while preserving the exact two-submit outcome. The cancellation regression fails before the post-spawn fence and adds no test-only production seam.

An independent review flagged `terminal-pty-worker.ts` as calling the string-only node-pty API with a Buffer. That finding was rejected: the worker calls the `TerminalPtyHandle` contract, which explicitly accepts `string | Buffer`; only `terminal-pty.ts` calls node-pty directly and contains the audited cast. The exact core and test type checks pass.


steipete:
Merged via squash.

- Prepared head SHA: [8c05e10ee4225f04299c3670f48ae3bd4c3f9934](https://github.com/openclaw/openclaw/pull/145749/commits/8c05e10ee4225f04299c3670f48ae3bd4c3f9934)
- Landed commit: [d61ae99a8ff777f230dbfaa9b44d974b223bd48a](https://github.com/openclaw/openclaw/commit/d61ae99a8ff777f230dbfaa9b44d974b223bd48a)

Issue #145686, opened 2026-09-12
Reduce remaining Gateway allocation churn under sustained parallel agent work


steipete:
## Remaining allocation cost under parallel agent work

Follow-up to #145608 after #145629 and #145630 landed. A fresh 32-session profile on a53c601cd40 still samples about 7.25GiB of Gateway allocations. The largest remaining stacks include repeated provider-policy resolution during catalog preparation, repeated async-context frame copying around plugin callbacks, and argument-graph descriptor materialization.

The catalog layer already has a per-operation identity resolver, but several row-processing paths still call the scalar resolver repeatedly. Plugin callbacks also enter registry and identity scopes separately, making Node copy its async-context map twice even though one isolated scope can carry both sets of facts. Argument restoration eagerly builds descriptors for graphs whose data is returned unchanged.

The next validation uses sustained tool work: 64 concurrent sessions performing 8 turns each, 512 seeded sessions, observers/history readers, and 1024 session updates. Live OpenAI baseline proof has completed 128 file-processing tasks across 32 sessions, with varied inputs and independently checked outputs.

Reduce these costs within the existing owners, preserve live admission and scoped identity, and publish matching before/after evidence. Keep the sustained benchmark exercising tools on every follow-up turn.

PR #145751, opened 2026-09-12
chore: exercise sustained multi-session Gateway load

steipete:
Related: #145686

## What Problem This Solves

Gateway concurrency benchmarks previously exercised one turn per session. They could not reproduce the accumulated tool, task, and observer work of a busy Gateway serving ongoing conversations. The mock provider also treated a tool result anywhere in history as completion of the current turn, silently skipping tool calls on later turns.

## Why This Change Was Made

Add `--turns-per-session` (default 1, maximum 100). Sessions advance concurrently while each session waits for its own turn to finish. The initial handshake barrier remains tied to the first turn; prompts and idempotency keys stay distinct. Report completed turn totals and normalize sampled allocations by that total. The existing child-only heap profiler continues covering the entire joined load window.

The mock provider now recognizes tool outputs only after the latest user message, for Responses and chat-completions. Document the sustained command and the existing bounded probe retention.

## User Impact

Developers can reproduce sustained multi-session load, including repeated tool execution, history reads, session updates, and visible observers. The default benchmark remains a single turn per session.

## Evidence

- 57 benchmark and mock-provider tests passed, including independent session advancement, ordered turns, failure propagation, and follow-up tools in streaming and nonstreaming Responses/chat-completions. The four follow-up tool regressions fail against the original mock provider.
- A built Gateway completed 512 turns across 64 sessions (8 turns each), with 512 seeded sessions, 32 observers, 4 history clients, and 1,024 session updates.
- That sustained baseline exposed a session-list timeout and unstable Tasks-list reads; the benchmark records those failures rather than hiding them. Runtime fixes are tracked separately in #145686.

```sh
node scripts/bench-gateway-concurrency.ts --concurrency 64 --turns-per-session 8 \
  --session-count 512 --tool-events --subscribers 32 --visible-observer \
  --history-clients 4 --history-messages 8 --control-plane --session-updates 1024 \
  --timeout-ms 600000 --heap-prof-dir .artifacts/gateway-load-profiles \
  --output .artifacts/gateway-load.json
```

### DIFF
--- a/docs/reference/test/performance.md
+++ b/docs/reference/test/performance.md
@@ -107,7 +107,18 @@ first; no provider key is required.
 
 ```bash
 pnpm test:gateway:concurrency -- --concurrency 16 --tool-events --workspace-fanout --session-count 100 --history-messages 20 --history-clients 4 --subscribers 4 --visible-observer --control-plane --heap-prof-dir .artifacts/gateway-heap --output .artifacts/gateway-concurrency.json
+pnpm test:gateway:concurrency -- --concurrency 64 --turns-per-session 8 --tool-events --timeout-ms 600000 --heap-prof-dir .artifacts/gateway-sustained-heap --output .artifacts/gateway-sustained.json
 ```
+
+`--concurrency` controls parallel sessions; `--turns-per-session` controls serial
+turns in each session (default 1, maximum 100). The second example completes 512
+turns across 64 sessions. Each session starts its next turn as soon as its
+previous turn completes, retaining its conversation history and workspace;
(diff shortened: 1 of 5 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d5058edca791edd58b0fea8b243d874f94178b7e](https://github.com/openclaw/openclaw/pull/145751/commits/d5058edca791edd58b0fea8b243d874f94178b7e)
- Landed commit: [c02442188a09e9958dacb943877dd9b1478f8046](https://github.com/openclaw/openclaw/commit/c02442188a09e9958dacb943877dd9b1478f8046)

PR #145759, opened 2026-09-12
test(ci): bound hosted tooling growth inventory

steipete:
## What Problem This Solves

`test/scripts/ci-node-test-plan.test.ts > keeps hosted tooling within the GitHub job cap when its inventory grows` planned over the real, growing repository inventory twelve times per run and timed out at 120 s on a hosted runner this week (shard core-tooling-15-hosted-3, run 34142285775) on a PR that never touched it.

## Why This Change Was Made

The cap contract does not need the real inventory: the test now drives the twelve planning passes with bounded synthetic inventories that exercise the same growth shape, keeping every original assertion. Case time drops from 8.9 s to 0.6 s locally, and the test no longer gets slower as the repository grows. Production code is unchanged; the testing docs note the bounded-inventory convention.

## User Impact

None at runtime. Fewer spurious red shards for contributors.

## Evidence

- All 61 owner tests in the file pass; the changed case runs in 0.575 s versus 8.858 s before.
- `node scripts/check-changed.mjs` and `pnpm check:architecture` pass; independent review clean.
- Line delta: production 0, tests +60/−29, docs +6.

## Known Gaps

Two other hosted-CI flakes seen this week were investigated and not changed, because neither yielded an owner-side defect: the `update-dry-run-state.process` CLI spawn stall happened on an already-built, warmed CLI, and the native Mermaid browser failures in CI were permanent engine failures rather than a premature "delivered" signal. Both remain tracked separately.

### DIFF
--- a/docs/help/testing/writing-tests.md
+++ b/docs/help/testing/writing-tests.md
@@ -64,6 +64,12 @@ Future evals should stay deterministic first:
 
 ## Adding regressions (guidance)
 
+For inventory-growth and capacity regressions, pass bounded synthetic inventories
+and fixed timing data through the real planner. Put the fixture at the capacity
+boundary and preserve coverage, ownership, and execution-budget assertions. Keep
+real-checkout inventory coverage in its integration tests instead of rebuilding
+the growing repository plan for every synthetic variation.
+
 When you fix a provider/model issue discovered in live:
 
 - Add a CI-safe regression if possible (mock/stub provider, or capture the exact request-shape transformation)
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [f8371eee0dbb29df5c7a67ffa811a6a6fb806367](https://github.com/openclaw/openclaw/pull/145759/commits/f8371eee0dbb29df5c7a67ffa811a6a6fb806367)
- Landed commit: [760f8b771543db959160b306f92bbc6741e4882a](https://github.com/openclaw/openclaw/commit/760f8b771543db959160b306f92bbc6741e4882a)

PR #145761, opened 2026-09-12
fix(plugins): take a fresh lifecycle lease when a channel starts so provider login can apply settings

obviyus:
Related: #145484. @steipete: this repairs its channel-lifetime handoff while preserving live plugin reload.

## What Problem This Solves

After plugin reload, chat sign-in could save credentials but fail to apply connection settings, leaving the account partly configured.

## Why This Change Was Made

The channel manager clears inherited plugin lifecycle ownership when starting a channel. Later settings writes acquire current ownership instead of reusing the completed reload's closed lease. Startup, reload, and automatic restart share this boundary; a truly closed lease still rejects reuse.

## User Impact

Sign-in completes, account models appear, and the first turn works after reload. No configuration or migration change is required.

## Evidence

The registered config-watcher/channel-restart regression failed before the fix. All 176 channel tests passed afterward. Telegram login → /models → model selection → first turn passed with recorded provider responses. Browser sign-in and picker capture passed; opening the populated picker made no request. Concurrent login/reload checks preserved responsiveness. All 16 valid startup fixtures passed, with the invalid legacy fixture rejected as expected. Previous-release state retained saved credentials and transcript history through supported repair and sign-in. Continuous integration passed. Live provider entitlement was not tested.

### DIFF
--- a/src/gateway/server-channels.ts
+++ b/src/gateway/server-channels.ts
@@ -54,6 +54,7 @@ import {
   type PluginHttpRouteHandoff,
 } from "../plugins/http-registry.js";
 import { runPluginCleanup } from "../plugins/plugin-instance-scope.js";
+import { runOutsidePluginLifecycleLease } from "../plugins/plugin-lifecycle-lease.js";
 import type { PluginRegistry } from "../plugins/registry.js";
 import { withPluginRuntimeRegistryScope } from "../plugins/runtime/gateway-request-scope.js";
 import { runOutsidePluginRuntimeGenerationScope } from "../plugins/runtime/generation-scope.js";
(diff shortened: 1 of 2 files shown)


obviyus:
landpr-approve-admin-merge
pr: 145761
head: 3289aa39e30229f7a2dc28e935753371a6350a41
scope: owner-review

obviyus:
Thanks @obviyus!

Merged as d629351dad5de3e3dc3fcf93c46baf4792705ed1 after the exact-head review, census ACCEPT, and green CI run 34682387383 (attempt 1).

| Before | After |
| --- | --- |
| Provider sign-in after a channel reload saved credentials but failed to apply settings because it inherited a closed lifecycle lease. | Channel startup clears inherited ownership; sign-in saves settings, account models appear, and the first turn reaches the recording provider fixture. A genuinely closed lease still refuses reuse. |

Validation includes 176 owner/channel tests, real Telegram sign-in, both reload/login overlap directions, previous-release state preservation, 16 valid config starts plus the expected legacy rejection, and a populated picker opening without a new request. The accepted health samples were 501.641 ms and 7.075 ms.

PR #145766, opened 2026-09-12
fix(update): bound candidate validation output

ZengWen-DT:
## What Problem This Solves

Candidate update validation fully buffered gateway probe responses and the post-install Doctor receipt before validating them. A malformed or incompatible candidate could therefore make `openclaw update` allocate an unbounded response or file before failing safely.

## Why This Change Was Made

Read the fixed-shape `startupz` payload through the existing bounded response reader with a 64 KiB ceiling, consume no `readyz` payload, and read the Doctor receipt through the existing regular-file helper with a 1 MiB ceiling. The validation owner now enforces the bounds before JSON decoding while preserving readiness status checks and receipt cleanup.

Production delta: +32/-8 lines (net +24). The growth establishes the two missing allocation boundaries; tests add +80 lines. No config, schema, protocol, or normal candidate behavior changes.

## User Impact

Updates now reject oversized candidate validation output without continuing to read it into memory. Valid candidates and ordinary Doctor receipts behave as before, and rejected Doctor receipts are still removed.

## Evidence

- Before the fix, `response.json()` and `fs.readFile()` materialized their complete inputs. The regression stream could produce all 131,105 bytes, and an oversized receipt was read before rejection.
- After the fix, the streamed startup regression stops after 73,728 bytes, cancels the body, and never reaches readiness; the oversized on-disk receipt returns `null` and is deleted.
- `OPENCLAW_VITEST_MAX_WORKERS=1 node scripts/run-vitest.mjs run src/infra/update-candidate-canary.test.ts src/infra/update-doctor-result.test.ts` — 49/49 passed on Vitest 5.
- Focused `oxfmt`, `oxlint`, and `git diff --check` passed. Full-suite validation is delegated to exact-head CI.

### Real candidate Gateway validation (2026-09-16)

- Exact head: `9033f2e4f350f11afe25a922c009f1e1fce54835`.
- In a task-owned independent clone with physical dependencies, `NODE_OPTIONS=--max-old-space-size=6500 node scripts/run-vitest.mjs src/infra/update-candidate-canary.integration.test.ts --reporter=verbose` passed (1/1; candidate test 44.3s, shard 106.8s including build).
- This integration starts the built candidate as a real child Gateway with `--update-canary`, retains the occupied sandbox-port setting used by older/published updater handoff, observes successful `startupz: started` and `readyz: ready`, verifies the occupied listener remains intact, reaps the candidate process, and removes the rehearsal state. It exercises the ordinary candidate/Doctor/rehearsal path without touching the installed Gateway.
- This supplements, but does not replace, the bounded rejection tests above. No historical published updater binary was executed, and the open compatibility question for a legitimate Doctor receipt larger than 1 MiB remains unresolved; this run must not be read as proof that such receipts cannot occur.
- The shared PR worktree could not run this lane because its borrowed `node_modules` is intentionally non-reconcilable. The independent exact-head clone was used to satisfy the repository dependency-ownership rule; it remained detached and no source changes were made.

### DIFF
--- a/src/infra/update-candidate-canary.test.ts
+++ b/src/infra/update-candidate-canary.test.ts
@@ -557,6 +557,73 @@ describe("update candidate canary", () => {
     await expect(fs.access(childEnv.OPENCLAW_STATE_DIR!)).rejects.toMatchObject({ code: "ENOENT" });
   });
 
+  it("cancels the ready probe body without decoding it", async () => {
+    const readyResponse = Response.json({ ready: true });
+    const decodeReadyBody = vi.spyOn(readyResponse, "json");
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(async (url: string) =>
+        url.endsWith("startupz") ? Response.json({ status: "started" }) : readyResponse,
+      ),
+    );
+
+    const result = await validateUpdateCandidateCanary({
+      root,
+      stateDir: root,
+      config: {},
+      env: {},
+      timeoutMs: 3_000,
+    });
+
+    expect(result).toMatchObject({ status: "ok", phase: "readiness" });
+    expect(decodeReadyBody).not.toHaveBeenCalled();
+    expect(readyResponse.bodyUsed).toBe(true);
+  });
+
+  it("rejects an oversized streamed startup payload before advancing to readiness", async () => {
+    const encoded = new TextEncoder().encode(
+      JSON.stringify({ status: "started", padding: "x".repeat(128 * 1024) }),
+    );
+    let offset = 0;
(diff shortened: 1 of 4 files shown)


ZengWen-DT:
@clawsweeper re-review

ZengWen-DT:
Closing instead of mechanically resolving the conflicts. Current main moved readiness polling into a new owner, while this branch still requires an unresolved producer/consumer compatibility decision for legitimate Doctor receipts above the proposed 1 MiB ceiling and published-updater proof. Without that contract, rebasing would preserve a known merge risk rather than produce a merge-ready repair.

Issue #142772, opened 2026-09-09
[Bug]: Slack Agent View “is working” status can remain after a completed turn until the desktop client restarts


joncursi:
### Bug type

Behavior bug (incorrect output/state without crash)

### Beta release blocker

No

### Summary

With Slack Agent View session status enabled, OpenClaw can complete a turn, finalize its Block Kit progress card, and deliver the final reply while Slack Desktop continues displaying the agent as working; restarting the Slack desktop client clears the stale indicator.

### Steps to reproduce

1. Run OpenClaw 2026.9.3 with Slack Socket Mode and the following progress configuration:

   ```json
   {
     "streaming": {
       "mode": "progress",
       "nativeTransport": true,
       "progress": {
         "style": "card",
         "nativeTaskCards": false,
         "commentary": false,
         "narration": true,
         "toolProgress": false,
         "maxLineChars": 400,
         "commandText": "status"
       }
     }
   }
   ```

2. From Slack Desktop on macOS, send a thread request that produces a multi-step progress card.
3. Allow the turn to finish normally.
4. Confirm that the progress card shows every step completed and that the final assistant reply is visible.
5. Observe that the native Slack callout still says the agent “is working.”
6. Restart the Slack desktop application.
7. Observe that the stale working callout disappears.

### Expected behavior

When the turn reaches terminal completion and the final reply is delivered, OpenClaw should transition the Slack agent session from `processing` to `active`, and Slack Desktop should promptly remove the working callout without requiring an application restart.

Slack documents `active` as the state to set when an agent has finished and is ready for the next prompt:

<https://docs.slack.dev/ai/agent-sessions/>

### Actual behavior

The observed turn completed successfully:

- The Block Kit progress card displayed `3/3 complete`.
- The final assistant response was posted.
- OpenClaw session history recorded a terminal assistant response.
- The session had no pending inputs.
- Slack Desktop continued displaying the agent working callout.
- Restarting Slack Desktop cleared the callout immediately.

The restart result suggests that the server-side session may already have been `active` while the desktop client retained stale UI. The available evidence does not establish whether Slack missed the status transition event, OpenClaw's terminal status write failed silently, or the ordering of the final message/card/status updates triggered a Slack client rendering race.

### OpenClaw version

2026.9.3 (`1391f7c`)

### Operating system

[...]

PR #145778, opened 2026-09-12
fix(slack): report failed terminal session status updates

Alix-007:
Related: #142772. This fixes terminal API-failure diagnostics, not Slack Desktop's working-indicator behavior.

## What Problem This Solves

Fixes silent Slack session-status cleanup failures when Slack accepts `processing` but rejects the terminal `active` update.

## User Impact

Operators receive one normal-level diagnostic and can enable verbose logging for the underlying Slack API failure. Rejected or unsupported processing writes do not produce this new diagnostic. Status requests, cleanup ordering and retry behavior are unchanged; typing reactions are still removed. No configuration or migration is required.

An accepted status request is not proof that Slack Desktop has refreshed its UI, so the broader issue remains open.

## Why This Change Was Made

Preserve API acceptance through the existing monitor context and consume it in the existing dispatch cleanup owner. A successful write remains successful when another agent keeps Slack's aggregate status at `processing`, or when a subsequent title rename fails. The diagnostic contains no token, channel/thread identifiers, title or raw API error.

The integration preserves current main's inferred context type rather than restoring its retired type file, and updates the existing delivery-trace fixture to the same acceptance contract. Unnecessary helper logging and wording-pinned test assertions were removed.

## Evidence

Actual before/after proof ran on Linux arm64 / Node 24.19.0 through the supported Gateway CLI, the registered **signed Slack HTTP webhook**, real agent dispatch and typing callbacks, and the installed Slack WebClient against a loopback API. Only external Slack/model endpoints were fixtures; the monitor, dispatch, context, SDK response handling and normal file logger were real.

| Scenario | Baseline diagnostics | Candidate diagnostics |
| --- | ---: | ---: |
| Processing accepted, active rejected | **0** | **1** |
| Both writes accepted | 0 | 0 |
| Processing rejected | 0 | 0 |
| Active accepted, aggregate still processing | 0 | 0 |

Every scenario delivered the scripted final reply, wrote `processing` then `active` exactly once, and completed typing-reaction cleanup. Each isolated Gateway was stopped before reading its log. The baseline failed the required diagnostic assertion; the candidate passed the same driver.

Observed normal-level error:

```text
Slack session status could not return to active after processing. Enable verbose logging to inspect the Slack API failure.
```

- Baseline: `7a932d9794421961e45a4715b08d63db74b7fb2d`; the affected owners were unchanged on subsequently checked main `2e053112b61f34964d4b0c360e3562c71f728e99`.
- Candidate: `947b7bd6622af803009bf9ce31b772ec6b2777e1`; committed production hashes match the observed runtime.
- Regression on unmodified baseline: **4 expected failures and 5 passing controls**. The candidate also covers no thread, repeated cleanup, quiet accepted/aggregate-status controls, privacy, failed title rename, and cleanup when the logger throws.
- **501 focused tests passed across seven files**: 471 Slack tests and 30 shared typing lifecycle tests. The nine diagnostic cases and real Gateway flow were rerun after final message-guidance cleanup. Media fixtures used deterministic DNS inside the network-isolated container; no media or network-policy source was changed.
- Targeted type-aware Oxlint, Oxfmt, counted-line growth check and `git diff --check` passed. Full static/matrix validation remains hosted.

```bash
node scripts/run-vitest.mjs \
  extensions/slack/src/monitor/message-handler/dispatch.status-diagnostics.test.ts \
  extensions/slack/src/monitor/context.test.ts \
  extensions/slack/src/monitor/message-handler/dispatch.preview-fallback.test.ts \
  extensions/slack/src/monitor/message-handler/prepare.test.ts \
  extensions/slack/src/monitor/events/agent.test.ts \
  extensions/slack/src/delivery-trace.test.ts \
  src/channels/typing.test.ts
```

No live Slack workspace or Desktop rendering was exercised, and no retry, status reordering, dependency, schema or permission change is included.

### DIFF
--- a/extensions/slack/src/monitor/context.ts
+++ b/extensions/slack/src/monitor/context.ts
@@ -153,13 +153,14 @@ export type SlackMonitorContext = {
   ) => SlackMessageEvent["channel_type"] | undefined;
   resolveUserName: (userId: string, eventScope?: SlackEventScope) => Promise<SlackUserInfo>;
   resolveUserAvatar: (userId: string, eventScope?: SlackEventScope) => string | undefined;
+  /** True means Slack accepted the status write, not that every client rendered it. */
   setSlackSessionStatus: (params: {
     channelId: string;
     threadTs?: string;
     status: "processing" | "active" | "suspended";
     title?: string;
     eventScope?: SlackEventScope;
-  }) => Promise<void>;
+  }) => Promise<boolean>;
   recordSlackSessionTitle: (params: {
     channelId: string;
     threadTs: string;
@@ -431,12 +432,12 @@ export function createSlackMonitorContext(params: {
       runtime: params.runtime,
     });
     if (!updated.ok || p.status !== "processing" || !p.threadTs || p.title === undefined) {
-      return;
+      return updated.ok;
     }
     const title = truncateUtf16Safe(p.title, 200);
     // A user rename received while the status request was in flight wins.
     if (readLruMapEntry(sessionTitles, key) !== previousTitle) {
-      return;
+      return true;
     }
     // setStatus only names newly created sessions. Rename existing sessions once
     // per display-name change; inbound user renames update this same cache.
@@ -455,6 +456,7 @@ export function createSlackMonitorContext(params: {
         recordSlackSessionTitle({ ...p, threadTs: p.threadTs, title });
       }
     }
+    return true;
   };
 
   const setSlackSuggestedPrompts = (input: SlackSuggestedPromptsInput) =>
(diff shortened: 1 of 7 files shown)


obviyus:
Landed in 5a36317384321ef89941d356430841170522f30d.

Slack now reports a failed terminal `active` update after accepted `processing`. The signed Gateway → real WebClient loopback proof changed the missing diagnostic from 0 to 1 while preserving quiet controls, final replies and reaction cleanup. All 501 focused tests passed, required hosted CI was green, and the exact-head review was clean.

This is a diagnostic repair only. #142772 remains open; no Slack Desktop indicator fix is claimed.

Thanks @Alix-007 for the contribution and @joncursi for the related report.

Issue #45994, opened 2026-03-14
WhatsApp active-listener singleton not shared across bundled chunks (outbound sends fail)


javicasper:
## Bug Description

Proactive WhatsApp message sending (via CLI `message send`, cron announce delivery, or agent `message` tool) fails with:

```
Error: No active WhatsApp Web listener (account: default). Start the gateway, then link WhatsApp with: openclaw channels login --channel whatsapp --account default.
```

**Despite WhatsApp being fully connected** — `channels status --probe` shows `linked, running, connected`, and inbound auto-replies work correctly.

## Root Cause

`src/web/active-listener.ts` defines a module-scope `listeners = new Map()` singleton. The bundler (tsdown) emits this module into **two separate chunks**:

- `model-selection-*.js` (used by the gateway WebSocket handler / agent outbound path)
- `reply-*.js` (used by `channel-web-*.js` which initializes the WhatsApp connection)

Each chunk gets its own independent `listeners` Map. When `channel-web` calls `setActiveWebListener()`, it registers the listener in the `reply-*.js` Map. When the gateway WebSocket handler calls `requireActiveWebListener()` to send a proactive message, it checks the `model-selection-*.js` Map — which is always empty.

This is the **exact same class of bug** fixed by PR #43683 ("Runtime: share singleton state across bundled chunks") for Telegram, Slack, Signal, iMessage, and core pipeline singletons. **WhatsApp's `active-listener.ts` was not included in that fix.**

## Steps to Reproduce

1. Start gateway with WhatsApp linked
2. Verify `openclaw channels status --probe` shows `connected`
3. Send an inbound WhatsApp message → auto-reply works ✅
4. Run `openclaw message send --channel whatsapp --target "+1234567890" --message "test"` → fails with "No active WhatsApp Web listener" ❌
5. Run any cron with WhatsApp delivery → same failure ❌

## Expected Behavior

Proactive sends should work when WhatsApp is connected.

## Workaround

Patch both chunks to share the Map via `globalThis`:

```js
// In both model-selection-*.js and reply-*.js, replace:
const listeners = /* @__PURE__ */ new Map();
// With:
const listeners = globalThis.__openclaw_wa_listeners ??= /* @__PURE__ */ new Map();
```

## Proposed Fix

Apply the same `resolveGlobalSingleton(Symbol.for("openclaw.wa-listeners"))` pattern from PR #43683 to `src/web/active-listener.ts`.

## Environment

[...]

PR #43683, opened 2026-03-12
fix(runtime): duplicate messages, share singleton state across bundled chunks

vincentkoc:
## Summary

Describe the problem and fix in 2–5 bullets:

- Problem: tsdown code-splitting duplicated mutable module-scoped “singleton” registries across bundled chunks, so one process could disagree with itself about busy-session state depending on which chunk handled the work.
- Why it matters: that explains the deeper cross-channel duplicate/busy-session reports in #25192 and #33150. It was not just a channel-routing bug; shared runtime state itself was splitting.
- What changed: moved every confirmed registry on this path onto shared `globalThis` state: command queue, embedded runs, followup queue state, drain callbacks, queued-message dedupe, inbound dedupe, Slack thread participation, Telegram thread bindings, Telegram draft-id allocation, and Telegram sent-message tracking.
- Verification: added fresh-import regression tests that emulate code-split chunks, rebuilt `dist/`, and verified the old per-chunk `new Map()` / `nextDraftId` signatures disappeared for the covered surfaces.
- Scope boundary: this addresses duplicate finals / overlapping runs / busy-session redelivery. Channel-native preview/final artifacts are still a separate layer (for example Telegram preview/final behavior tracked under #33453 / #33592).

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Security hardening
- [ ] Chore/infra

## Scope (select all touched areas)

- [x] Gateway / orchestration
- [x] Integrations
- [ ] Skills / tool execution
- [ ] Auth / tokens
- [ ] Memory / storage
- [ ] API / contracts
- [ ] UI / DX
- [ ] CI/CD / infra

## Linked Issue/PR

- Closes #41901
- Related #25192
- Related #33150
- Related #34039
- Related #34041
- Related #41903
- Context: #33168

## User-visible / Behavior Changes

Busy-session duplicate handling is now backed by one process-global runtime state across bundled chunks. That should remove the shared-core class of duplicate finals / replayed queued messages seen across iMessage, Signal, Discord, Slack, Telegram, and plugin channels that feed the same reply pipeline.

## Security Impact (required)

- New permissions/capabilities? (`Yes/No`): No
- Secrets/tokens handling changed? (`Yes/No`): No
- New/changed network calls? (`Yes/No`): No
- Command/tool execution surface changed? (`Yes/No`): No
- Data access scope changed? (`Yes/No`): No
- If any `Yes`, explain risk + mitigation:

### Environment

[...]

PR #145780, opened 2026-09-12
fix(agents): keep the bash process registry process-wide

### LINKED ISSUE #145779: [Bug]: background exec runs can vanish from the runtime "Active exec sessions" carrier while the process tool still lists them
### Bug type

### Beta release blocker
No

### Summary
A backgrounded exec can be live and pollable through the `process` tool while the injected runtime "Active exec sessions" carrier reports `none` on every turn of the same session. Two readers of one registry disagree about the same run.

### Steps to reproduce
1. In one session, start a long background run through `exec` with `background=true`:
   `for i in $(seq 1 180); do date; sleep 10; done 2>&1 | tee /tmp/exec-repro.log`
2. On each following turn, read the injected runtime context's `Active exec sessions` line and run `process list` in the same turn.
3. Keep going for several turns while the run is alive.

### Observed (macOS 26.6.2 arm64, Node 26.7.0, OpenClaw 2026.9.3, 2026-09-11 18:45 to 19:05 PDT)
- The carrier read `Active exec sessions: none` on every turn for twenty minutes.
- `process list` on those same turns reported live sessions, up to three at once, for example `tidal-basil running 7m :: for i`, while `/tmp/exec-repro2.log` grew a line every ten seconds.
- The earlier opposite direction in the same session: the carrier listed a session whose pid no longer existed, while `process kill` answered "No active session found".

### Root cause: registry state is per module instance
`src/agents/bash-process-registry.ts` keeps its state in module-level declarations: `runningSessions`, `finishedSessions`, `processSessionStartOrders`, `nextProcessSessionStartOrder`, `activeExecSessions`, `finishedSessionOutputChars` and `sweeper`.

The module is inlined into more than one bundle: the shared chunk `dist/bash-process-registry-<hash>.mjs`, and the self-contained `dist/worker/worker.mjs`, which does not import the chunk and carries its own copy. Neither goes through `resolveGlobalSingleton`, while 103 files in the same tree do.

Proven in one process on the shipped dist, with a control: loading both bundles publishes two distinct registry test APIs into the shared global slot, and a session admitted through one instance survives `resetProcessRegistryForTests()` on the other. Control: resetting the owning instance does clear the session, so the check is sound.

### What I could not confirm
That both copies are loaded in one Gateway process. The worker bundle is started as a child process (`execFileAsync(process.execPath, [worker.mjs, "--internal-worker-prewarm"])` in `dist/daemon-*`), and no third copy of the registry exists anywhere in the install, so the split is the leading explanation rather than a measured cause on this host.

The scope key is excluded as an alternative: both readers resolve it with `resolveProcessToolScopeKey` from the same attempt fields, since `attempt-tool-prepare.ts` passes `attempt.sessionKey?.trim() || attempt.sessionId` as `runSessionKey` and `attempt-prompt-build.ts` passes `attempt.sessionKey` and `attempt.sessionId`.

### Related: three notions of "live"
- `listRunningSessions()` filters `runningSessions` on the session's `backgrounded` flag.
- `hasActiveBackgroundExecSession()` reads `activeExecSessions.get(id)?.promoted`, a second copy of the same fact.
- Finished records remain in `finishedSessions` for `cleanupMs`, 30 minutes by default.

### Validated and excluded: the scope filter
Two scope predicates disagree by construction: the carrier's `listActiveProcessSessionReferences()` returns `[]` for a falsy scope and otherwise requires `session.scopeKey === scopeKey`, while `bash-tools.process.ts` accepts `!scopeKey || session?.scopeKey === scopeKey`.

That asymmetry cannot explain this report. Enumerating attempt field states and resolving both keys the way production does (carrier: `attempt.sessionKey` / `attempt.sessionId` / `sessionAgentId`; tools: `runSessionKey = attempt.sessionKey?.trim() || attempt.sessionId`, `sessionId`, `executionAgentId`) shows the two keys agree in every state, including a sandbox key that differs from the session key, a blank session key, and no session key at all. The one divergence is an explicit caller-supplied `exec.scopeKey`, and no production caller sets one: only `attempt-tool-prepare.ts`, `prepared-compaction-runtime.ts` and `tools-effective-inventory.ts` construct these tools. The carrier's key also cannot be falsy, since `attempt-prompt-build.ts` always supplies `agentId` alongside the session fields.

So a session with a non-empty scope key, equal on both sides, can still be invisible to the carrier, which leaves the registry instance as the only remaining variable. Checking the real carrier reader against a registered session that matches its scope, and one that does not, confirms the filter itself behaves correctly.

Evidence for both checks is in `src/agents/bash-process-scope.validation.test.ts` on the fork branch `evidence/bash-process-registry-instance-probe`.

### Prior art, and why this is the same class
- #45994 (closed): `src/web/active-listener.ts` was duplicated across seven chunks, each with an independent `Map()`, which broke proactive WhatsApp sends while inbound worked. #43683 (merged) fixed that class by moving every confirmed registry on the path onto shared `globalThis` state through `resolveGlobalSingleton`, and added fresh-import regression tests that emulate code-split chunks. #47389 (closed) proposed a postinstall guard for exactly this failure mode.
- #129456 (merged) had to work around the same split inside this registry's own test support: it binds registry cleanup to its importing module instance because "a later module evaluation may replace the global slot while existing callers still own the original instance".
- #127219 (open) touches the same file and the same module-global state, on retention rather than on readers.

I found no existing issue for the reader disagreement itself after searching, among others, "bash process registry", "exec session registry", "Active exec sessions", "process registry singleton", "resolveGlobalSingleton", "code-splitting singleton", "runtime context carrier", and the background-exec and notifyOnExit phrasings.

### Proposed fix
Move the state behind `resolveGlobalSingleton(Symbol.for("openclaw.bashProcessRegistry"), ...)`, which is what #43683 did for the chunk-duplication class, and which `agentJobState`, `sessionMaintenance` and the agent event bus already use. The PR also updates `src/agents/bash-process-registry.test.ts` case "resets its own registry after another module instance replaces the global test API", which currently asserts that a reloaded module instance starts empty, the exact behaviour that splits readers.

Aligning the two scope predicates is not part of it: that route was tested and excluded above, and would be a no-op for this symptom.

### Environment
OpenClaw 2026.9.3 (1391f7c), npm global install, macOS 26.6.2 arm64, Node 26.7.0.

azuretek:
Fixes #145779.

## What Problem This Solves
The bash process registry keeps its state in module-level declarations, so every evaluation of `src/agents/bash-process-registry.ts` starts its own Maps. The module is inlined into more than one bundle, the shared Gateway chunk and the self-contained worker bundle, and readers reached through different copies then disagree about the same run: a backgrounded exec can stay alive and pollable through the `process` tool while the injected runtime "Active exec sessions" carrier, which reads the same registry through `listActiveProcessSessionReferences()`, reports `none` on every turn. The mirror case is a record whose exit one copy processed and the other never reaped, which is why the carrier can also list a session with a dead pid.

This is the same class as #45994 and the fix in #43683, where a mutable module-scoped registry duplicated across bundled chunks let one process disagree with itself about session state.

## Change
- Move the registry's Maps, counters and sweeper behind `resolveGlobalSingleton(Symbol.for("openclaw.bashProcessRegistry"), ...)`, the pattern #43683 established for chunk duplication and that `agentJobState`, `sessionMaintenance` and the agent event bus already use.
- Keep the existing module-level bindings as aliases of the shared Maps so the rest of the file, and its callers, are unchanged.
- Update `src/agents/bash-process-registry.test.ts` case "resets its own registry after another module instance replaces the global test API": with one process-wide registry, a later module evaluation must observe the original instance's session instead of starting from its own empty Maps, which is the behaviour that split the readers.

## Evidence
Live observation on 2026.9.3, macOS 26.6.2 arm64, Node 26.7.0, 2026-09-11 18:45 to 19:05 PDT: the injected carrier read `Active exec sessions: none` on every turn for twenty minutes, while `process list` in those same turns reported live sessions, up to three at once, for example `tidal-basil running 7m :: for i`, and the run's log grew a line every ten seconds. The reproduction command is in the issue.

Registry instances in one process, on the shipped dist, both bundles loaded into a single Node process: two distinct registry test APIs are published into the shared global slot, and a session admitted through one instance survives `resetProcessRegistryForTests()` on the other. Control: resetting the owning instance does clear it, so the check is sound. Script kept on the fork branch `evidence/bash-process-registry-instance-probe`.

Failing-then-passing test, new `src/agents/bash-process-registry.instances.test.ts`, in the fresh-import style the registry suite already uses for a second module instance. On current main it fails with the production symptom, `expected [] to match object [{ sessionId: "instance-probe", status: "running" }]`, while both controls pass. With this change all three pass, and the suite reports 30 passing across `bash-process-registry.instances.test.ts`, `bash-process-registry.test.ts` and `bash-process-references.test.ts` against 28 passing with the two cross-instance assertions failing on main. `oxlint` reports no warnings or errors and `oxfmt --check` is clean.

Scope filter ruled out, which is what leaves the registry instance as the only remaining variable. Enumerating attempt field states and resolving both keys the way production does (carrier: `attempt.sessionKey`, `attempt.sessionId`, `sessionAgentId`; tools: `runSessionKey = attempt.sessionKey?.trim() || attempt.sessionId`, `sessionId`, `executionAgentId`) shows the two keys agree in every state, including a sandbox key that differs from the session key, a blank session key and no session key at all. The single divergence needs a caller-supplied `exec.scopeKey`, and no production caller sets one. Evidence: `src/agents/bash-process-scope.validation.test.ts` on the same fork branch.

## Not verified here
Which bundle supplies the carrier's own copy inside a live Gateway process. The worker bundle's prewarm runs as a child process and no third copy of the registry exists in the install, so the split is established as the only remaining explanation for the observation above rather than as a directly measured cause. The decisive follow-up is an in-process read of the live gateway registry, which needs an inspector this build does not expose, since `SIGUSR1` is the gateway's own in-process restart signal.

### DIFF
--- a/src/agents/bash-process-registry.instances.test.ts
+++ b/src/agents/bash-process-registry.instances.test.ts
@@ -1,0 +1,73 @@
+/**
+ * Duplicated registry module instances must observe one process-wide state.
+ *
+ * The bash process registry is reloaded once per bundle that inlines it (the
+ * shared Gateway chunk and the self-contained worker bundle), and the test
+ * support helper documents that "a later module evaluation may replace the
+ * global slot while existing callers still own the original instance".
+ *
+ * A background exec admitted through one module instance has to stay visible to
+ * the readers served by another instance: the runtime "Active exec sessions"
+ * carrier is fed by `listActiveProcessSessionReferences()`, while the `process`
+ * tool reads the registry directly. When the state is per instance rather than
+ * per process those two readers disagree about the same run.
+ *
+ * Registry state is process-wide, so every case resets it explicitly, the same
+ * way the rest of the registry suite isolates its own sessions.
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { createProcessSessionFixture } from "./bash-process-registry.test-helpers.js";
+import { resetProcessRegistryForTests } from "./bash-process-registry.test-support.js";
+
+const SCOPE_KEY = "agent:main:registry-instance-probe";
+
+describe("bash process registry across module instances", () => {
+  beforeEach(() => {
+    resetProcessRegistryForTests();
+    vi.resetModules();
+  });
+
+  afterEach(() => {
+    resetProcessRegistryForTests();
+    vi.resetModules();
+  });
+
+  it("shows a background session admitted by another module instance", async () => {
+    const writer = await import("./bash-process-registry.js");
+    vi.resetModules();
+    const reader = await import("./bash-process-references.js");
+
+    const session = createProcessSessionFixture({ id: "instance-probe", backgrounded: true });
+    session.scopeKey = SCOPE_KEY;
+    writer.addSession(session);
+
+    expect(reader.listActiveProcessSessionReferences({ scopeKey: SCOPE_KEY })).toMatchObject([
+      { sessionId: "instance-probe", status: "running" },
+    ]);
+  });
+
+  it("keeps the carrier reader working for its own instance (control)", async () => {
+    const reader = await import("./bash-process-references.js");
+    const readerRegistry = await import("./bash-process-registry.js");
+
+    const session = createProcessSessionFixture({ id: "own-instance-probe", backgrounded: true });
+    session.scopeKey = SCOPE_KEY;
+    readerRegistry.addSession(session);
+
+    expect(reader.listActiveProcessSessionReferences({ scopeKey: SCOPE_KEY })).toMatchObject([
+      { sessionId: "own-instance-probe", status: "running" },
+    ]);
+  });
+
+  it("keeps the scope filter closed to other scopes (control)", async () => {
+    const writer = await import("./bash-process-registry.js");
+    vi.resetModules();
+    const reader = await import("./bash-process-references.js");
+
+    const session = createProcessSessionFixture({ id: "other-scope-probe", backgrounded: true });
+    session.scopeKey = "agent:other:elsewhere";
+    writer.addSession(session);
+
+    expect(reader.listActiveProcessSessionReferences({ scopeKey: SCOPE_KEY })).toEqual([]);
+  });
+});
--- a/src/agents/bash-process-registry.test.ts
+++ b/src/agents/bash-process-registry.test.ts
@@ -402,7 +402,7 @@ describe("bash process registry", () => {
     expect(getActiveBackgroundExecSessionCount()).toBe(0);
   });
 
-  it("resets its own registry after another module instance replaces the global test API", async () => {
+  it("shares one registry with a later module instance and resets it", async () => {
     const testApiKey = Symbol.for("openclaw.bashProcessRegistryTestApi");
     const globalStore = globalThis as Record<PropertyKey, unknown>;
     const originalTestApi = globalStore[testApiKey];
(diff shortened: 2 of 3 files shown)


azuretek:
Closing this. The observation behind it does not hold up, and the fix would be a no-op in the affected process.

Measured on 2026.9.4 with an in-thread `module.registerHooks` load hook on a Gateway started from a clean `main` build: exactly one file carrying the registry is evaluated in the Gateway process, `dist/bash-process-registry-*.mjs`, on the main thread, once. `dist/worker/worker.mjs` carries the second copy but is only ever launched as a child process, so it never shares a process with the carrier or the `process` tool.

The 2026-09-11 evidence was one user turn, not twenty minutes of turns. The session transcript has a single user message at 18:45:36 PDT and the next one after midnight; the intervening tool loop reused the runtime context built at 18:45:36, before the first background run started, which is why it read `none` throughout and why it later listed a run that had already exited. The carrier is a turn-start snapshot and `process list` is the live reader. Checked on 2026.9.4 across a CLI-driven session, a dashboard-keyed session and a live Control UI session: at turn start the carrier and `process list` agree on the same session ids and pids.

Backup of the branch stays on the fork under `backup/bash-process-registry-global-singleton-2026-09-13`.

PR #145785, opened 2026-09-12
feat(signal): distinguish envelope age from local admission time

### LINKED ISSUE #144470: [Feature]: Log Signal envelope age and local ingress admission time
## Problem

When an inbound Signal message appears delayed, current logs do not distinguish an
envelope that was already old when local ingress observed it from an envelope that
spent time waiting for and completing durable local admission. Those cases require
different investigation paths but can look identical from later agent-run events.

## Proposed solution

Add one Signal-owned, content-free informational log record after successful durable
admission of a new queue entry:

```json
{
  "signalIngressTiming": true,
  "envelopeAgeAtIngressSeconds": 240,
  "localAdmissionElapsedMs": 5
}
```

- `envelopeAgeAtIngressSeconds` is the signed, whole-second difference between the
  Signal envelope timestamp and the local ingress observation. It is an apparent age
  across two clocks, not a transport SLA.
- `localAdmissionElapsedMs` uses a monotonic clock and covers waiting for and
  completing serialized durable admission, including Signal identity-alias handling.

The record would contain no message content, source or destination identifiers,
account ID, group ID, event ID, session key, attachment metadata, or exact source
timestamp. Logging failure would not affect admission or dispatch.

## Operator impact

An operator could distinguish these otherwise similar incidents:

1. Old envelope, fast local admission: approximately `240s / 5ms`.
2. Fresh envelope, slow local admission: approximately `1s / 239000ms`.

This improves coarse incident triage. It does not correlate the event to an agent run
and does not confirm authorization, execution, reply delivery, or device receipt.

## Alternatives considered

- A single `transportLatencyMs` value was rejected because it conflates pre-ingress
  envelope age with local queue work.
- A cross-channel diagnostics event or persistent telemetry store would be much larger
  than the Signal-specific evidence gap.
- Debug-only logging would reduce volume but would have to be enabled before an
  incident, making the first useful observation easy to miss.

## Proposed validation

- Fixture proof that old-envelope/fast-admission and fresh-envelope/slow-admission
  produce different value pairs.
- Negative apparent age is retained for future source timestamps.
- Concrete and Signal identity-alias redeliveries do not add a second record when the
  existing durable tombstone resolves them.
- A throwing logger cannot change admission or dispatch.
- Exact payload-key allowlist proves that message and account metadata do not leak.

## Maintainer decision requested

Would maintainers accept this Signal-owned timing record at the existing informational
logging level, without identifiers, configuration, persistence changes, or a shared
diagnostic-event API?

Alix-007:
Closes #144470.

## What Problem This Solves

Operators investigating delayed Signal messages cannot distinguish an envelope that was already old when OpenClaw received it from slow local durable admission.

## Why This Change Was Made

The Signal plugin records coarse signed envelope age and monotonic local admission elapsed time after insertion and identity-alias handling. The existing queue, deduplication, and dispatch owners remain authoritative. Timing facts stay in memory; the durable payload and SQLite schema do not change.

Production delta: +25/-7 (net +18), justified by the new admission diagnostic. Tests: +330. Documentation: +33. The callback's early return is consolidated into the alias/duplicate guards; no parallel admission path is added.

## User Impact

One informational JSON record per newly admitted logical envelope helps distinguish already-old input from local admission cost. Exact and number/UUID duplicates within the existing tombstone window do not log again. Failed admission, transport-only inputs, and recovered rows do not invent receive timings. The record contains only a marker and two durations, with no message content, identifiers, or exact source timestamp.

These records add normal-level log volume and expose coarse activity patterns. They prove admission, not authorization, model execution, reply delivery, device receipt, or network latency. Logging exceptions cannot turn successful admission into failure.

### September 14 integration refresh

Current head: `e00e1f7cbc924e2872c2e5ec3b0eac9df69caf96`. Original branch head `b7c3a34753eb44b748a1059c96f821b6e545c988` and pinned main `8b917bc499607eb94bb77358ddd69f6494d26c65` are both preserved as parents; no force-push. Refresh the existing accepted patch onto pinned main to remove stale integration inputs and obtain new-head CI. No new feature or unrelated failure workaround is added. The resulting tree exactly matches the conflict-free git merge-tree result; there are no manual source or test edits.

Validation on Linux / Node 26.1.0 with frozen-lockfile dependencies: No new local runtime test is claimed for this mechanical integration; existing proof keeps its original revision and exact-head hosted CI is required.. The native staged-content/format guard, main-relative diff check, and one fresh independent P0-P2 source review passed. No full build or full type-aware check was rerun locally; new-head CI is tracked separately. Earlier runtime proof below retains its recorded SHA; it is not relabeled as a new execution.

AI-assisted implementation, manually inspected at the Signal receive/admission owner and shared queue callback/alias boundary.

**Behavior addressed:** distinguish signed apparent envelope age from monotonic local admission duration while preserving deduplication, recovery, payload privacy, and dispatch semantics.

**Environment:** Node 24.21.0; isolated temporary state; actual TypeScript Signal ingress monitor and canonical SQLite queue, loaded with the repository TS loader. Synthetic raw receive envelopes and a terminal no-model dispatch consumer; no replaced queue/admission modules.

**Verified head:** `b7c3a34753eb44b748a1059c96f821b6e545c988`. Tests, proof, and review ran on frozen staged content; all three committed file hashes match the final proof snapshot exactly.

**Production entry and recorder:** `node --import "$WORK/scripts/tsx.mjs" "$PACKET/proofs/signal-ingress.mjs"` runs synthetic raw receive events through `startSignalIngressMonitor.receive` → canonical durable admission and alias handling → SQLite queue, then reads the actual file-backed runtime log and dispatch counts. The capture command was `python3 "$PACKET/scripts/run.py" proof A01 --repo "$WORK" --base dda647045420956f7900180c5ca32d457dbc4973 --mode candidate --out "$EVIDENCE/proof-final"`. The production file SHA-256 was `637ec6bc2811c94d031fa7fa8084741a21e370d6eb75e2e57123d2b08e1ebe97`, unchanged by the commit.

**Evidence before fix:** the unmodified base `dda647045420956f7900180c5ca32d457dbc4973` admitted and dispatched the 240-second-old envelope, but emitted zero timing records. Exact and number/UUID alias redeliveries kept dispatches at one; transport-only input added no dispatch. A future-clock envelope brought dispatches to two, still with zero timing records. All four baseline observations passed their before-change expectations.

```json
{"dispatches":1,"timings":[]}
{"dispatches":1,"timings":[]}
{"dispatches":1,"timingCount":0}
{"dispatches":2,"timings":[],"errors":[]}
```

**Evidence after fix:** all four final real-clock scenarios passed. The first envelope emitted age `240s` and local admission `44ms`. Exact and number/UUID alias redeliveries kept dispatches and timing records at one. A transport-only envelope added neither. The future-clock envelope emitted signed age `-10s` with `11ms` admission and brought dispatches to two. No error was recorded; the exact timing-record key set and absence of fixture identifiers/message text were checked.

```json
{"signalIngressTiming":true,"envelopeAgeAtIngressSeconds":240,"localAdmissionElapsedMs":44}
{"signalIngressTiming":true,"envelopeAgeAtIngressSeconds":-10,"localAdmissionElapsedMs":11}
```

These are observed values from this run, not performance guarantees.

**Validation:** 42/42 tests passed across the two focused files: 13 timing tests and 29 durable-ingress tests, with no failures or skips. Formatting and staged diff whitespace checks passed. Targeted syntax lint passed: 2 files, 230 rules, zero diagnostics. Independent gpt-6-astra/medium/P2 review passed with no findings and concluded that the patch is correct; its secret scan was clean.

**Hosted CI:** the separate Security Sensitive Guard run `34681900955` failed with `500 Internal Server Error` while executing the GitHub guard. The contributor account's retry was rejected with `Must have admin rights to Repository`; a maintainer rerun is required. This is not presented as a passing security guard or as a product-code failure.

```text
node scripts/run-vitest.mjs extensions/signal/src/signal-ingress.timing.test.ts extensions/signal/src/signal-ingress.test.ts
Test Files  2 passed (2)
     Tests  42 passed (42)
```

**Existing owner and compatibility:** this uses the existing ingress monitor, SQLite queue and runtime logger. The durable serializer remains `{ receivedAt, event }` with payload version 1; timestamp and monotonic timing facts are transient. No config/default key, SQLite schema/version, stored payload, public protocol or dependency is changed. No data migration or backfill is needed.

**Limitations:** no signal-cli SSE networking, Signal service, real device, or model turn was exercised. Recovery, artificial queue contention, wall-clock changes, logger throws, storage/alias failures, and payload privacy are tested separately by repository tests. Broader event-handler lifecycle coverage, full builds, global typecheck/lint and docs checks are left to CI. Only the completed focused runs above are counted as passing validation.

**Cleanup:** both baseline and final proof stopped the monitor, closed database handles, and removed isolated scratch. Source content stayed unchanged during both proof runs.

### DIFF
--- a/docs/channels/signal.md
+++ b/docs/channels/signal.md
@@ -302,6 +302,39 @@ Allowed group messages that do not mention the bot stay silent and are kept only
 - Replies to inbound messages include native Signal quote metadata when the backend accepts the inbound timestamp and author; if quote metadata is missing or rejected, OpenClaw sends the reply as a normal message.
 - Configure native quote use with `channels.signal.replyToMode = off | first | all | batched`, or `channels.signal.replyToModeByChatType.direct/group` for per-chat-type overrides. Account-level values under `channels.signal.accounts.<id>` take precedence.
 
+### Ingress timing
+
+After a new logical Signal envelope completes durable admission and identity-alias
+handling, the Signal plugin emits one informational record:
+
+```json
+{
+  "signalIngressTiming": true,
+  "envelopeAgeAtIngressSeconds": 240,
+  "localAdmissionElapsedMs": 5
+}
+```
+
+`envelopeAgeAtIngressSeconds` is the local ingress observation time minus the
+validated envelope timestamp, truncated toward zero to whole seconds. It compares
+two wall clocks, preserves negative ages, and is not a transport latency or SLA.
+`localAdmissionElapsedMs` uses a monotonic clock, rounded to milliseconds. It
+includes serialized queue waiting, insertion, and identity-alias handling, but
+excludes waiting for agent dispatch. For example, `240s / 5ms` distinguishes an
+already-old envelope from the slow local admission represented by `1s / 239000ms`.
+Neither pair establishes the cause of a delay.
+
+Concrete and identity-alias duplicates recognized within the existing tombstone
+window do not emit another record. Recovered rows, failed admission, and ignored
+transport-only envelopes do not invent new receive timings. The record contains
+only the marker and two durations, not message content, source or destination IDs,
+account or group IDs, event IDs, session keys, attachments, or exact source timestamps.
+No timing fields are added to persistent state. Logging failure cannot fail admission.
+
+These records add log volume and reveal coarse activity patterns, so existing log
+retention and access controls still apply. A record proves admission, not
+authorization, execution, reply delivery, or device receipt.
+
 ## Media + limits
 
 - Outbound text is chunked to `channels.signal.textChunkLimit` (default 4000).
--- a/extensions/signal/src/signal-ingress.timing.test.ts
+++ b/extensions/signal/src/signal-ingress.timing.test.ts
@@ -1,0 +1,330 @@
+// Real ingress monitor and SQLite queue, with clocks and transport delivery controlled.
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { performance } from "node:perf_hooks";
+import {
+  closeOpenClawStateDatabaseForTest,
+  createChannelIngressQueueForTests,
+} from "openclaw/plugin-sdk/channel-ingress-test-runtime";
+import { createDeferred } from "openclaw/plugin-sdk/extension-shared";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { SignalSseEvent } from "./client-adapter.js";
+import { startSignalIngressMonitor } from "./signal-ingress.js";
+
+type Queue = NonNullable<Parameters<typeof startSignalIngressMonitor>[0]["queue"]>;
+type Payload = Parameters<Queue["enqueue"]>[1];
+const now = 1_800_000_000_000;
+function event(timestamp = now - 1_000, uuid?: string): SignalSseEvent {
+  return {
+    event: "receive",
+    data: JSON.stringify({
+      envelope: {
+        sourceNumber: "+15550001111",
+        ...(uuid ? { sourceUuid: uuid } : {}),
+        timestamp,
+        dataMessage: { timestamp, message: "synthetic-private-message" },
+      },
+    }),
+  };
+}
+function timingRows(log: ReturnType<typeof vi.fn>): unknown[] {
+  return log.mock.calls.flatMap(([line]) => {
+    if (typeof line !== "string" || !line.startsWith('{"signalIngressTiming":true,')) {
+      return [];
+    }
+    return [JSON.parse(line)];
+  });
+}
+async function withQueue(run: (queue: Queue) => Promise<void>) {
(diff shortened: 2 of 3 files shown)


Alix-007:
Closing this implementation after reassessing the default logging tradeoff. The timing distinction is useful, but emitting an INFO record for every newly admitted logical envelope adds routine log volume and exposes coarse activity patterns without itself fixing delivery latency. I am withdrawing that default-INFO proposal pending an agreed logging policy.

The diagnostic request in #144470 remains separate and is not considered resolved by this closure. A narrower approach using an existing debug or explicitly enabled diagnostic mechanism would need its own discussion. The branch, tests, and recorded evidence are retained for reference. Thank you for the review.

PR #145791, opened 2026-09-12
refactor(plugins): unify Gateway startup planning

steipete:
## What Problem This Solves

Gateway startup, plugin reload, and metadata-backed CLI lookups need to select activation from the same prepared plugin inventory. Maintaining duplicate startup-planning paths makes that snapshot and policy contract harder to preserve as reload behavior evolves.

This is a maintainer-requested ownership cleanup, not a claimed startup speedup.

## Why This Change Was Made

Make lookup-table construction consume the existing plan-with-metadata loader, while retaining its own worker-ID normalization, table projection, and metrics. Planning time remains measured after metadata acquisition and is added to the existing metadata total.

Keep manifest lookup in the planner that uses it, remove its obsolete single-caller wrappers, and retire an ID-only adapter used only by tests. Move those existing assertions through the full plan owner without changing expected values, fixtures, timeouts, or platform coverage. The change removes 93 production lines and one test line after formatting.

## User Impact

No intended user-visible behavior change. Supplied snapshots, including empty and scoped inventories, remain authoritative. Raw versus effective configuration, external indexed channel contributions, bundled manifest channels, last-manifest precedence, and live worker IDs retain their existing contracts. Reload publication, authority, disposal, and retirement remain with their current owners. No configuration, persistence, schema, or public SDK changes.

Historical supplied-snapshot measurements had full output parity but mixed performance. Denied lookup regressed 17.9%/24.1% in wall time and 18%/36.2% in CPU; reverse-with-metadata and empty-plan CPU controls also had adverse results. Direct plan-with-metadata callers now pay for the timing reads. This PR is justified by removing duplicate ownership, not by a broad latency or memory claim, and those historical results are not presented as a current-main Gateway benchmark.

## Evidence

- 219 focused tests passed: 215 across the complete startup-policy, bundled-metadata, and lookup-table owner files; three existing startup-caller cases; and one real registry/lookup case that verifies retained declarations and owners after removing its manifest file. The 49 unselected assertions in targeted files are not counted as passes.
- Full `pnpm build` passed, including declarations, SDK export checks, the CLI bootstrap import guard, and UI build. Matching formatting and fresh independent P2 review passed.
- 26 of 29 repository-selected local check commands passed. The cumulative check runner exceeded its fixed 64-process budget, so the remaining existing commands were run separately without increasing per-command limits. Import-cycle checking found zero runtime value cycles.
- **Pending hosted CI:** core typechecking and targeted type-aware lint exceeded the configured 6 GiB process-tree RSS limit, at observed peaks of 6,544,048,128 and 6,474,579,968 bytes respectively. The core-test typecheck attempt stopped before compilation on the artifact lock retained after the terminated core checker. The exact owned process tree was verified closed before recovering that lock; these checks were not retried with larger limits or weaker assertions. All local children and monitors closed, and source hashes remained unchanged.
- The existing 600-second real Gateway reload lifecycle test was not run inside the 300-second local validation envelope. A direct plugin-import profile is not claimed as proof of core startup planning.

Hosted exact-head CI must pass all normal typecheck and lint gates before landing. Local resource failures and historical adverse measurements remain preserved; no benchmark rerun or cap/guard changes were made.

### DIFF
--- a/src/plugins/bundled-plugin-metadata.test.ts
+++ b/src/plugins/bundled-plugin-metadata.test.ts
@@ -12,7 +12,7 @@ import { resolveBundledPluginGeneratedPath } from "./bundled-plugin-scan.js";
 import { isPluginEnabledByDefaultForPlatform } from "./default-enablement.js";
 
 type BundledPluginMetadata = ReturnType<typeof listBundledPluginMetadata>[number];
-import { resolveGatewayStartupPluginIdsFromRegistry } from "./gateway-startup-plugin-ids.js";
+import { resolveGatewayStartupPluginPlanFromRegistry } from "./gateway-startup-plugin-ids.js";
 import {
   createGeneratedPluginTempRoot,
   installGeneratedPluginTempRootCleanup,
@@ -592,13 +592,13 @@ describe("bundled plugin metadata", () => {
     ].toSorted((left, right) => left.localeCompare(right));
 
     expect(
-      resolveGatewayStartupPluginIdsFromRegistry({
+      resolveGatewayStartupPluginPlanFromRegistry({
         config: {},
         env: {},
         index,
         manifestRegistry,
         platform: "linux",
-      }),
+      }).pluginIds,
     ).toEqual(expectedPluginIds);
   });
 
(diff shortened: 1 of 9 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [29ed190674447dcd345a6813f8ace4401d354505](https://github.com/openclaw/openclaw/pull/145791/commits/29ed190674447dcd345a6813f8ace4401d354505)
- Landed commit: [85d015fbb9b37be0e17dda27ff6deb4b3549e68a](https://github.com/openclaw/openclaw/commit/85d015fbb9b37be0e17dda27ff6deb4b3549e68a)

PR #145793, opened 2026-09-12
improve: start cloud worker turns sooner over remote connections

steipete:
## What Problem This Solves

Cloud workers wait for the Gateway to accept their connection before loading the runtime needed to start a turn. On remote connections, those two independent waits unnecessarily add together.

## Why This Change Was Made

Begin the existing runtime imports after the connect frame finishes writing, while admission is pending. Waiting for write completion preserves compressed WebSocket scheduling. Stream construction, workspace preparation, inference and tools still wait for validated hello.

Late callbacks cannot trigger loading after denial, stop or socket replacement. Import errors are observed immediately, and cleanup joins both imports before restoring the worker environment.

## User Impact

On the measured macOS host, runtime initialization took roughly 120 ms. Workers can overlap that work with the admission round trip. A controlled 100 ms response wait hid about 101 ms of that initialization in three diagnostic trials. Local connections offer little overlap. This reduces a specific startup wait; it does not reduce CPU work.

A denied or cancelled startup must finish module initialization before exiting. In 18 final-bundle local probes, this added 122–147 ms to median rejection, fencing and cancellation exit time; every outcome and cleanup check passed, with no inference, tools or workspace writes. This is a measured local tradeoff, not a fleet-wide latency bound.

## Evidence

- Seven real-WebSocket race cases passed with the existing connection tests (31 total). Original admission source fails the two new successful-notification cases.
- Four isolated runtime cases cover denied/cancelled/accepted hello, import failure and environment cleanup. Removing the rejection observer or settlement join makes the tests fail.
- All 108 broader runtime, managed-command, environment and import cases passed; all 35 focused connection/import cases passed on the final rebased head.
- Independent review through P2 found no actionable issues.
- Full `pnpm build` passed on `9e84f89bc213`. Selected formatting, typecheck, lint, dead-export and architecture guards passed.
- All 40 real-network trials passed using macOS workers, an isolated Linux Gateway on Daytona, and an authenticated SSH tunnel. Both ABBA and reversed block orders were retained. Combined median spawn-to-exec was 1,348 ms baseline and 1,262 ms candidate, but batch effects reversed; this does not establish a stable overall speedup. The phase trace below explains the bounded benefit without attributing all network variation to this change.
- The supported tunnel-first setup retained every workspace-owner renewal check throughout the successful campaign. Earlier setup collisions are preserved as failed infrastructure attempts, without timing samples.
- Twelve separate instrumented local diagnostic flows passed: three zero-delay and three fixed-100-ms-response trials per version. Baseline initialization remained after hello; candidate overlapped 100.83–101.69 ms of the fixed wait and left 18.54–30.31 ms after the server wrote hello. These are synthetic phase measurements, separate from the ordinary real-network results.
- Independent review through P2 and ClawSweeper review are clean. [Exact-head CI](https://github.com/openclaw/openclaw/actions/runs/34682979657) passed. All task-owned cloud leases, processes and temporary credentials were cleaned up.

### DIFF
--- a/src/worker/worker-connection-admission.ts
+++ b/src/worker/worker-connection-admission.ts
@@ -153,6 +153,14 @@ export function connectWorkerConnectionAttempt(
             new WorkerConnectionInterruptedError(`admission send failed: ${error.message}`),
           );
           socket.terminate();
+          return;
+        }
+        if (isActive() && admission === "pending") {
+          try {
+            connectionOptions.onAdmissionRequestSent?.();
+          } catch {
+            // Optional preparation observers do not control admission or retry policy.
+          }
         }
       });
     });
--- a/src/worker/worker-connection-contract.ts
+++ b/src/worker/worker-connection-contract.ts
@@ -50,6 +50,8 @@ export type WorkerConnectionOptions = {
   requestTimeoutMs?: number;
   createSocket?: (url: string, options: ClientOptions) => WebSocket;
   heartbeatStatus?: () => WorkerHeartbeatParams["status"];
+  /** The connect frame was written; this does not establish admission. */
+  onAdmissionRequestSent?: () => void;
   onConnectionFailure?: (error: Error | undefined) => void;
 };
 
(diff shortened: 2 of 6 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9e84f89bc213dfbd09e70f9eb7394c53338b4c90](https://github.com/openclaw/openclaw/pull/145793/commits/9e84f89bc213dfbd09e70f9eb7394c53338b4c90)
- Landed commit: [32ae7ef1b85e58d31452b6847d2d670f5bdd631b](https://github.com/openclaw/openclaw/commit/32ae7ef1b85e58d31452b6847d2d670f5bdd631b)

PR #145659, opened 2026-09-12
fix(plugins): reduce memory spikes for large call arguments

fuller-stack-dev:
Related: #144252

<details>
<summary>Additional instructions</summary>

Maintainers can update this same-repository branch.

</details>

## What Problem This Solves

Fixes an issue where users making plugin calls with large data payloads incur avoidable temporary memory spikes and call latency, even when most or all of the input can pass through unchanged.

The managed callable path reproduced about 139 MiB of extra V8 heap for a synthetic 50,000-message graph occupying about 9.4 MiB. This establishes allocation overhead, not the cause of a particular production out-of-memory crash.

## Why This Change Was Made

Keep full property-descriptor maps only for ancestors that actually need copying during same-owner handle restoration. Most children have one parent, so store that parent directly and allocate a parent collection only for shared children.

The full admissible graph is still inspected. Caller accessors, methods, proxies, class instances, and foreign handles stay opaque. Cycles, aliases, sparse arrays, symbols, non-enumerable fields, property flags, input identity, and retirement refusal retain their contracts. No cross-call classification cache, lifecycle policy, memory-limit change, or payload-specific bypass is introduced.

## User Impact

Large managed plugin calls allocate less temporary memory and finish faster. The 50,000-message data-only benchmark reduced median sampled peak extra heap from 138.803 MiB to 75.648 MiB and elapsed time from 177.94 ms to 77.47 ms. The 45.5% peak reduction falls short of the prewritten 50% target; this PR does not claim otherwise.

This is a runtime allocation fix only. It does not change update, rollback, startup, configuration, or plugin admission/disposal behavior.

## Evidence

- Actual `PluginInstance.wrap` entry point, Node v24.20.0, synthetic data, fresh processes, 512 MiB old-space limit, warmed call, forced GC, and identical allocation observer on base/head.
- Unmodified base: 200,002 bulk descriptor maps per measured data-only call. Patch: zero, or one when a same-owner handle sits beside the unchanged message history.
- Regression tests fail on unmodified upstream in exactly the two allocation cases; the other five new semantic cases pass there. All seven pass with the patch.
- 284 focused tests passed across argument restoration, callbacks, constructors, reflection, iterables, lifetime, and consumers. Core/test type checks, changed-file lint/format, and selected static guards passed.
- Independent source-blind synthetic API validation passed identity, copied ancestors, cycles, aliases, sparse descriptors, opaque values, mutation between calls, retirement, and varied counts up to 100,000.
[...]

PR #145795, opened 2026-09-12
fix(gateway): reduce allocation and keep busy control requests responsive

### LINKED ISSUE #145686: Reduce remaining Gateway allocation churn under sustained parallel agent work

steipete:
Closes #145686
Benchmark: #145751

## What Problem This Solves

Fixes an issue where opening Sessions or Tasks on a busy Gateway could stall for seconds while agents ran tools in parallel. Repeated policy loading, plugin context copying, and host-metadata projection created substantial transient allocation. Task scans repeatedly yielded into fresh mutations, while small event and agent-result frames queued RPC replies behind serial compression callbacks.

## Why This Change Was Made

Reuse catalog identity policies within each synchronous selection operation, keep host hook context and diagnostic options behind an opaque tool metadata marker, and combine plugin/registry context entry. Source-tool views retain live plugin admission; scope mutations remain isolated and policy preparation refreshes on the next operation.

Cheap task pages now finish within the existing-style 12 ms work budget. Expensive scans still yield, with revision, cursor, and authorization checks preserved. Raise the existing outbound compression threshold to 32 KiB so ordinary events and agent completions avoid serial zlib queues; large responses still compress, with unchanged negotiation, ordering, payload limits, and backpressure handling.

## User Impact

Busy Gateways allocate less temporary memory and serve control requests more promptly. The sustained candidate completed every turn and all health, history, subscription, session-update, and control probes without errors.

This trades additional wire traffic for responsiveness. In the transport-only comparison, actual server output increased 24.4%, including additional control queries served by the faster run. Peak sampled RSS also increased in the combined workload; this is an allocation and responsiveness improvement, not a demonstrated reduction in resident memory.

## Evidence

Same configured workload: 64 concurrent sessions × 8 serial turns, 512 seeded sessions with 8 history messages, 32 visible observers, 4 history clients, and 1,024 session updates. The baseline includes upstream #145659; its argument-restoration work is not part of this diff.

| Heap-profiled sustained run | Baseline | Candidate |
| --- | ---: | ---: |
| Completed agent turns | 512 | 512 |
| Sampled allocated bytes | 105,808,366,000 | 89,464,959,080 (-15.45%) |
| Session-list p95 | 9,317 ms | 915 ms |
| Session-list timeouts | 1 | 0 |
| Tasks-list errors | 7 | 0 |
| Control sample rounds | 101 | 501 |
| Joined load duration | 198.45 s | 163.87 s |
| Sampled peak RSS | 2,091 MiB | 2,853 MiB |

These are closed-loop probes: the faster candidate served almost five times as many rounds, so totals do not represent identical amounts of control work. Allocation sampling includes collected objects and is distinct from retained heap. Endpoint heap use also increased (1,069 → 1,284 MiB). A candidate-only repeat on the published head completed another 512 turns and 499 control rounds without errors. After profile export, a two-second idle, and diagnostic-only GC, V8 heap use was 394 MiB and RSS was 2,098 MiB. Export itself allocates memory, and there is no matched baseline GC sample: this shows collectability but does not establish a smaller resident footprint.

Transport instrumentation correlated client send, server arrival, handler completion, send-queue flush, and client receipt. At 4 KiB, a session-list response spent roughly 19 ms in application work and 16.9 seconds in the send queue. At 32 KiB, session-list p95 was 445 ms and all 285 list responses entered an empty sender queue. An 8 KiB trial still stalled behind 61 compressed agent results around 20 KiB each, which justified the final threshold. No turn-throughput improvement is attributed to that isolated transport experiment.

- 582 focused runtime/benchmark cases plus 95 transport cases passed. Metadata identity, cheap task scans under churn, and 6/20 KiB wire behavior were demonstrated failing against their original owners or intermediate threshold. Large compressed post-auth frames, retirement, ordered delivery, and failed-connection cleanup remain covered.
- Selected changed-check guards, script/core/test typechecks, targeted lint, and the full build passed. Independent P0–P2 review is clean. The move onto current main is patch-identical.
- Nine access-snapshot cases also passed with explicit elapsed-work interleaving. They call the real authorization owner and retain all grant, revocation, returned-task, and resource-bound assertions.
- 128 live OpenAI tasks passed over 32 sessions and 4 waves, with fresh CSV input per wave, verified computed output files, 512 tool events, and no probe failures.

```sh
node scripts/bench-gateway-concurrency.ts --concurrency 64 --turns-per-session 8 \
  --session-count 512 --tool-events --subscribers 32 --visible-observer \
  --history-clients 4 --history-messages 8 --control-plane --session-updates 1024 \
  --timeout-ms 600000 --heap-prof-dir .artifacts/gateway-load-profiles \
  --output .artifacts/gateway-load.json
```

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -3118,7 +3118,7 @@ src/mcp/channel-bridge.ts	6
 src/mcp/channel-shared.ts	3
 src/mcp/codex-supervision-tools-serve.ts	6
 src/mcp/openclaw-tools-serve-config.ts	1
-src/mcp/plugin-tools-handlers.ts	3
+src/mcp/plugin-tools-handlers.ts	2
 src/media-generation/provider-registry.ts	1
 src/media-understanding/attachments.select.ts	1
 src/media-understanding/entry-capabilities.ts	1
--- a/docs/gateway/protocol/transport.md
+++ b/docs/gateway/protocol/transport.md
@@ -44,8 +44,10 @@ that supervise the Gateway as a child process, see
   sizes, limits, and a safe reason code, never message bodies, attachment
   contents, raw frame bytes, tokens, cookies, or secrets.
 - The Gateway offers `permessage-deflate`. Peers that negotiate it (browsers, `ws`
-  clients) receive frames of 4 KiB and up compressed; smaller frames such as
-  streaming deltas stay raw. Context takeover is disabled in both directions, so
+  clients) receive frames of 32 KiB and up compressed. Smaller session/tool updates
+  and ordinary agent results stay raw so serial compression callbacks do not delay
+  queued RPC replies. Large histories and rosters still compress. Context takeover
+  is disabled in both directions, so
   each frame compresses independently. Peers that do not offer the extension are
   unaffected. Payload limits apply to the inflated size.
 
--- a/src/agents/agent-tools.before-tool-call.wrapper.ts
+++ b/src/agents/agent-tools.before-tool-call.wrapper.ts
@@ -68,14 +68,11 @@ import {
   validateToolExecutionParams,
 } from "./agent-tools.execution-validation.js";
 import {
-  BEFORE_TOOL_CALL_DIAGNOSTIC_OPTIONS,
-  BEFORE_TOOL_CALL_HOOK_CONTEXT,
-  BEFORE_TOOL_CALL_SOURCE_TOOL,
-  BEFORE_TOOL_CALL_WRAPPED,
   clearBeforeToolCallWrappedMarker,
   getBeforeToolCallDiagnosticOptions,
   getBeforeToolCallHookContext,
   getBeforeToolCallSourceTool,
+  setBeforeToolCallMetadata,
   type BeforeToolCallDiagnosticOptions,
 } from "./before-tool-call-metadata.js";
 import { getChannelAgentToolMeta } from "./channel-tool-metadata.js";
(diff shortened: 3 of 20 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d582efc1e400824b14c379741966a6d2c951537c](https://github.com/openclaw/openclaw/pull/145795/commits/d582efc1e400824b14c379741966a6d2c951537c)
- Landed commit: [1a853fe0343d852b9b2e71541a53ce57d54aec2b](https://github.com/openclaw/openclaw/commit/1a853fe0343d852b9b2e71541a53ce57d54aec2b)

PR #145804, opened 2026-09-12
fix(plugins): keep live callbacks working after another instance retires

### LINKED ISSUE #145803: [Bug]: Another plugin instance can invalidate live callbacks
### Bug type

Behavior bug (incorrect runtime ownership).

### Summary

Live migration, compaction, and HTTP callbacks can lose their owning plugin runtime when another registration reuses the same raw descriptor or handler; warm migration lookup can also omit managed runtime scope.

### Steps to reproduce

1. Create a plugin registry and register a migration/compaction provider or HTTP handler through its real plugin API. Keep instance A selected.
2. Register the same raw object with instance B, either in another registry or as a rejected duplicate.
3. Retire B, then invoke A's callback. The bundled diagnostics-prometheus entry reproduces this with its module-level exporter: register it twice, retire B, and request `/api/diagnostics/prometheus` through A.
4. Separately, a runtime-store-backed migration provider fails through warm lookup even without a second registration.

### Expected behavior

A's callbacks use A's runtime and remain callable until A retires. Requests retain their selected registry, and disposal waits for admitted work. These are the managed-instance contracts documented in `docs/plugins/sdk-runtime.md`.

### Actual behavior

The Prometheus request returns HTTP 500 instead of 200. Compaction can read B's runtime or reject A's callback after B retires. Warm migrations fail with a missing-runtime error. An in-flight HTTP response can be charged to B, allowing A to dispose too early.

### Environment

Source checkout at `fde94351cacb8ef636ee3b308da920206d2a823d`; macOS 27.0, Node 24.20.0, repository Vitest wrapper. No external model/provider request was involved.

### Evidence and impact

Real-registration regressions reproduced two compaction failures and eight HTTP failures, including the actual bundled Prometheus entry. Handoff and upgrade-fallback controls passed. The earlier warm-migration regression failed both warm branches while the cold sibling passed. This blocks the affected callbacks in the reproduced registration/retirement sequence; production frequency was not measured.

steipete:
Closes #145803

<details>

</details>

## What Problem This Solves

Fixes an issue where users running plugin migrations, custom compaction, or plugin HTTP requests would lose a live callback when another plugin instance reused its descriptor and retired. Already-loaded migration providers could also execute without their managed runtime.

## Why This Change Was Made

Migration and compaction resolve execution from the winning registration's registry record. HTTP and WebSocket dispatch reuse the route's captured owner and existing generation scope, so shared callbacks retain the selected registry and in-flight work belongs to the correct instance.

## User Impact

Live callbacks keep their runtime until their own instance retires. Raw registration identity, authentication, host retry responses, and upgrade-specific fallback behavior remain intact.

## Evidence

- Before: two compaction regressions and eight HTTP regressions failed; the real bundled Prometheus scrape returned 500 instead of 200. Warm migration branches also failed their runtime-store regression.
- After: 196 focused owner/sibling cases passed, plus all 10 HTTP cases using the real plugin artifact loader and the final in-flight disposal check.
- The full seven-path changed gate and fresh independent P2 review passed.
- Proof uses local real-registration and Gateway boundary fixtures; no live operator Gateway or external provider call is claimed.

Maintainer clarification: the stored-data-model warning is a false positive. `src/plugins/migration-provider-runtime.ts` dispatches plugin callbacks; this diff selects and wraps their in-memory registration owners and preserves existing cleanup. It changes no SQLite table, schema, or persisted format, so additional database migration proof does not apply. The review reports no code or security findings.

### DIFF
--- a/src/gateway/server/plugins-http.ownership.test.ts
+++ b/src/gateway/server/plugins-http.ownership.test.ts
@@ -1,0 +1,295 @@
+import type { IncomingMessage } from "node:http";
+import { PassThrough } from "node:stream";
+import { setImmediate } from "node:timers/promises";
+import { expectDefined } from "@openclaw/normalization-core";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createPluginRuntimeCapabilityLease } from "../../plugins/capability-lease.js";
+import {
+  createPluginHttpRouteHandoff,
+  registerPluginHttpRoute,
+  withPluginHttpRouteRegistry,
+} from "../../plugins/http-registry.js";
+import { runPluginRegisterSyncInRegistry } from "../../plugins/loader-module-runtime.js";
+import { resolvePluginModuleExport } from "../../plugins/module-export.js";
+import {
+  getPluginInstance,
+  type PluginInstanceHandle,
+} from "../../plugins/plugin-instance-scope.js";
+import { loadBundledPluginPublicArtifactModuleSync } from "../../plugins/public-surface-loader.js";
+import { projectPluginContributions } from "../../plugins/registry-contributions.js";
+import {
+  createEmptyPluginRegistry,
+  createPluginRegistry,
+  type PluginRegistry,
+} from "../../plugins/registry.js";
+import { getPluginRuntimeGatewayRequestScope } from "../../plugins/runtime/gateway-request-scope.js";
+import type { PluginRuntime } from "../../plugins/runtime/types.js";
+import { createPluginRecord } from "../../plugins/status.test-helpers.js";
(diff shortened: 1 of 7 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [9fe2eed4459721fbf96477a12fe10c54218a8f72](https://github.com/openclaw/openclaw/pull/145804/commits/9fe2eed4459721fbf96477a12fe10c54218a8f72)
- Landed commit: [b25510f18d403bc274ebad9a3bb03f6af6040fc9](https://github.com/openclaw/openclaw/commit/b25510f18d403bc274ebad9a3bb03f6af6040fc9)

PR #145805, opened 2026-09-12
fix: keep attachment code controls after refresh

steipete:
## What Problem This Solves

Fixes an issue where users reading a Markdown attachment in the Files panel would lose their code-block expansion and wrapping choices when the attachment's download URL refreshed, even if its contents stayed identical.

## Why This Change Was Made

The reader now uses Lit's existing template cache while a refreshed body loads. The attachment identity and validated text determine whether to reuse the reader or replace it. Fetching, cancellation, size limits, and failure handling retain their existing owner; no additional presentation-state store is introduced. The source-preview documentation now reflects Markdown document rendering and the retained controls.

## User Impact

An unchanged attachment returns with the same expanded and wrapped code blocks after loading. A different attachment or changed text starts with fresh controls. Pending and unavailable previews stay hidden, and restored controls reacquire their resize observers.

## Evidence

- The browser regression fails on the original reader because an identical refresh resets Expand.
- 23 focused tests pass: five Chromium cases plus 18 cancellation and Markdown lifecycle cases. They cover transport refresh, identity/content replacement, failed-refresh recovery, metadata rejection, and observer cleanup.
- The complete changed-file checks and a fresh Control UI build pass.
- Independent review completed both passes with no actionable findings through P2.

Synthetic before/after captures use the actual Files detail-panel renderer. The UI bundle was built separately from the source-served browser proof. An initial after-capture module startup failed before reaching the scenario; the later diagnostic capture completed all interaction checks. No live Gateway or provider was used.

| Before: identical refresh resets the reader | After: expansion and wrapping remain |
| --- | --- |
| Identical attachment refresh resets expanded and wrapped code | Identical attachment refresh preserves expanded and wrapped code |

### DIFF
--- a/docs/web/control-ui/chat.md
+++ b/docs/web/control-ui/chat.md
@@ -226,11 +226,14 @@ an explanation in chat.
 ### Source previews and copying code
 
 Select **Open** on a text attachment to read it directly in the **Files** side
-panel. Same-origin text attachments, including pasted `.txt` files, Markdown,
-CSV, and JSON, display as selectable, read-only text with line breaks and
-indentation preserved. HTML and other markup remain literal text, never an
-embedded page. Previews require UTF-8 content no larger than 256 KiB; unsupported,
-external, oversized, or unavailable files keep their **Download** action.
+panel. Plain-text attachments, including pasted `.txt` files, CSV, and JSON,
+preserve line breaks and indentation. Markdown attachments render as documents
+with interactive code blocks. When an open attachment refreshes with unchanged
+text, its code blocks keep your expansion and wrapping choices after loading.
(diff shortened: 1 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d0aca9cb8233ef1f33469a5266554f60a771e201](https://github.com/openclaw/openclaw/pull/145805/commits/d0aca9cb8233ef1f33469a5266554f60a771e201)
- Landed commit: [48b121c0ee81ddcb207992567328e75f136394c4](https://github.com/openclaw/openclaw/commit/48b121c0ee81ddcb207992567328e75f136394c4)

PR #145599, opened 2026-09-12
refactor(voice): reduce GPT-Live support duplication

steipete:
Related: #145382, #145071, #145614, #145657

## What Problem This Solves

The GPT-Live paths carried redundant wire-target descriptors, unused helper arguments, and repeated test fixtures. These extra branches and forwarding steps made transport and shutdown behavior harder to review.

## Why This Change Was Made

Keep protocol selection in the existing wire builder, reuse established error-coercion and fixture helpers, and remove redundant private-call plumbing and fixture setup. The cleanup removes 88 lines across ten files: 46 production and 42 test/helper lines.

The Google loader follows the shared `createLazyRuntimeSurface` owner introduced by #145614. This PR retains only the separate private close-helper simplification in that file.

## User Impact

No user-visible behavior changes. Account-aware Talk defaults, public/subscription wire formats, caption output, lifecycle ownership, shutdown ordering, and primary-error precedence retain their existing contracts. Every test case and assertion remains, including distinct socket-close timing and portable mock declaration types.

## Evidence

- 278 focused voice tests passed, along with the full selected changed-file checks without skips, `pnpm build`, and the Google entrypoint profiler (211.2 MiB peak memory; no performance improvement is claimed).
- Fresh independent review through P2 is scoped-clean for the retained cleanup. Conflict resolution removes superseded changes and introduces no new cleanup logic.
- All 49 focused Google entrypoint, realtime-lazy, and media-video tests pass with Node 24.20.0 after retaining #145614's shared loader. All ten final voice files match that tested head exactly.
- Linux [CI 34676563045](https://github.com/openclaw/openclaw/actions/runs/34676563045) passed on the preceding head: 93 successful jobs, 13 skipped, zero failures, including all eight worker-context cases and the unchanged 50-turn retention case.
- Earlier CI exposed source-reader startup inside a five-second functional turn window. The two-line readiness fix landed independently in #145657; this PR drops its duplicate commit and keeps main's fixture verbatim. The five-second turn, 60-second reader, and 180-second setup budgets remain unchanged. No test retry loops, skipped cases, weaker assertions, or production test seams were added.
- Final exact-head [CI 34678099243](https://github.com/openclaw/openclaw/actions/runs/34678099243) passed on `e8967fffaf00`: 93 successful jobs, 13 skipped, zero failures, including the Gateway shard and real-Gateway UI E2E. The fresh current-head ClawSweeper review has no findings or before-merge items.

PR #145382, opened 2026-09-11
feat(openai): support GPT-Live and default Talk by account

steipete:
Closes #145071

## What Problem This Solves

Fixes an issue where users selecting the newly released public `gpt-live-1` model could not establish a usable voice session: OpenClaw sent the earlier GPT-Live protocol, which uses different endpoints, startup messages, and delegation events. The public model was missing from the catalog, its Platform authentication and voices were not distinguished from the Codex subscription route, and browser/iOS clients ignored its transcript events. Talk also defaulted to the older Realtime model regardless of the selected OpenAI account.

## Why This Change Was Made

The existing OpenAI voice plugin now adapts the [public Live API](https://developers.openai.com/api/docs/guides/live) to OpenClaw's voice and agent-consult contracts. Browser and iOS Talk use Gateway-brokered WebRTC; Gateway relay and backend voice use the primary WebSocket. Public sessions use the documented session creation, transcript, delegation, context-append, and closure events, with the public voices and `marin` default. The ChatGPT subscription route remains distinct.

Talk and Gateway relay sessions now select their OpenAI model from the selected agent's configured accounts: a Platform key, API-key profile, or `OPENAI_API_KEY` selects `gpt-live-1`; a ChatGPT-only account selects `gpt-live-1-codex`. A configured Platform credential takes precedence, including when it needs repair. The existing provider resolver carries session, account, and capability context. Explicit model pins remain authoritative. GPT-Live is audio-only, so the Control UI does not offer camera capture with this default; choose `gpt-realtime-2.1` explicitly to use the camera. Manual-turn, video-required, Azure, and bridge consumers that require GA Realtime behavior keep that contract.

Browser capability discovery now passes the requested provider and model through the existing `talk.catalog` RPC. The Gateway resolves effective capabilities for that launch before the client negotiates camera support. This fixes both explicit Live over a GA configuration and explicit GA over a Live configuration, including provider aliases. Catalog discovery uses the configured Talk agent; session creation resolves the selected agent's accounts. Optional catalog parameters preserve existing empty-query callers and generated protocol types.

Metadata-only delegation notices use bounded user transcripts and current conversation context. When a notice arrives before any nonblank user caption, both direct and sideband owners retain it for up to 15 seconds and resume through existing admission when text arrives. IDs are claimed before callbacks to prevent duplicate work; bounded queues and call retirement revoke pending notices. Missing input eventually requests clarification. Nonempty fragments still use existing semantic admission; no artificial turn-completed event is introduced.

[...]

PR #145807, opened 2026-09-12
fix(openai): correct spoken-update guidance for public GPT-Live

steipete:
Related: #145382, #145599

## What Problem This Solves

Resolves a problem where public GPT-Live voice sessions were instructed to keep backend updates silent even when the public API marks those updates for speech. Public Live gives commentary a different meaning from the ChatGPT subscription route.

## Why This Change Was Made

Select the startup instructions from the effective model in browser, Gateway, and direct voice creation. For public Live, thinking updates provide quiet context, commentary updates provide spoken content, and instruction updates direct the session, matching [OpenAI's Live delegation contract](https://developers.openai.com/api/docs/guides/live-delegation#send-the-right-kind-of-update).

The existing subscription prompt bytes, appended operator instructions, and host receipt/control behavior are preserved. The existing browser routing test covers both channel contracts, and the provider documentation explains the distinction for custom instructions.

## User Impact

Public GPT-Live voice sessions receive startup guidance that agrees with the public API's quiet and spoken update channels. The subscription route keeps its existing channel behavior.

Intermittent provider answer omissions remain a separate observation whose cause has not been established.

## Evidence

- The public browser routing regression failed on the old prompt for the intended assertion: 54 cases passed and the new public case failed. After the correction, all 94 routing, public-delegation, and wire tests passed.
- The complete selected changed-file checks passed without skips, including formatting, extension production/test types, declaration-boundary lint, doctor contracts, dead exports, owner guards, and runtime import-cycle checks.
- Managed isolated review through P2 is scoped-clean. The original non-public base prompt remains byte-identical (483 UTF-8 bytes), and all three production instruction-builder callers pass the effective selected model.
- Live relay checks on committed candidate `447501ffbd10`: 19 of 20 cases passed, including all ten public Live cases and the public six-turn conversation. The remaining subscription six-turn case received an incomplete second input transcript; the backend correctly asked for clarification. That diagnostic remains open. No late audio, captions, consultations, or writes were observed in the 20 relay records.
- A real Chromium WebRTC harness passed all four browser transport cases. Independent transcription of voiced output segments confirmed the expected public and subscription answer markers. These harnesses use real provider transports with synthetic backend replies; they do not claim Control UI or full agent-inference coverage.
- Independent audio verification passed 21 of 22 candidate checks, including intentional no-answer checks. The only retained failure is the subscription six-turn case above. Two older preflight checks are excluded from this candidate count.
- Required [CI 34684100007](https://github.com/openclaw/openclaw/actions/runs/34684100007) passed on `447501ffbd10`: 78 successful jobs, 16 skipped, zero failures.
- The initial hosted run failed four plugin model-override assertions outside this diff. The unchanged single file, paired files, and full 34-file child group all passed locally; the single actual hosted repeat passed all eight override cases on the identical merge snapshot. The first failure remains unexplained. GitHub's job-rerun request created no jobs and ended in a startup failure, so the supported close/reopen recovery attached the successful workflow. No source, isolation, assertion, or timeout change was made.

### DIFF
--- a/docs/providers/openai/voice-and-speech.md
+++ b/docs/providers/openai/voice-and-speech.md
@@ -251,7 +251,12 @@ sidebarTitle: "Voice and speech"
     The public API uses `/v1/live/sessions` for WebRTC creation and primary
     WebSockets. Direct sockets start with `session.start`; browser sessions
     start during the SDP exchange. A Gateway sideband handles delegated work
-    while browser media stays on WebRTC. Public transcript events are fragments,
+    while browser media stays on WebRTC. Public Live instructions describe
+    `session.thinking.append` as quiet context and `session.commentary.append`
+    as spoken updates. Custom instructions should preserve that distinction;
+    the Codex route uses its separate commentary/speakable channel contract.
+
+    Public transcript events are fragments,
     not completed turns. The Gateway owns persistence of bounded received-text
     snapshots; clients display captions without saving another copy. Both live
     captions and saved snapshots can span several exchanges by the same speaker;
--- a/extensions/openai/realtime-quicksilver-instructions.ts
+++ b/extensions/openai/realtime-quicksilver-instructions.ts
@@ -1,7 +1,10 @@
+import { isOpenAIGptLiveApiModel } from "./realtime-quicksilver.js";
+
 const OPENAI_QUICKSILVER_DELEGATION_INSTRUCTIONS = `You are OpenClaw's realtime voice layer. You have no tools of your own.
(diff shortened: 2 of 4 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [447501ffbd1032c5b9a18c37ababbd1971e2df79](https://github.com/openclaw/openclaw/pull/145807/commits/447501ffbd1032c5b9a18c37ababbd1971e2df79)
- Landed commit: [62d343151a25a27cacf05faae3d1ec3746674374](https://github.com/openclaw/openclaw/commit/62d343151a25a27cacf05faae3d1ec3746674374)

PR #145809, opened 2026-09-12
fix: canceled steering messages still reach active runs

### LINKED ISSUE #145727: [Bug]: Canceled steering input still reaches the active run after chat.abort succeeds
### Bug type

### Beta release blocker

No

### Summary

Through a persistent CLI-mode Gateway connection, canceling an accepted but unconsumed steering submission returns a successful `chat.abort` acknowledgment, yet that submission still enters the active embedded run's next model request.

### Steps to reproduce

Steering adds direction to a run already in progress. An illustrative user scenario is starting a report analysis, adding “Focus only on costs,” then withdrawing that additional instruction before it is consumed. The analysis should continue without the withdrawn direction. The reproduction below used synthetic marker text, not that example conversation.

Use an isolated Gateway and the embedded OpenClaw runtime with a synthetic Responses-compatible HTTP provider. Keep one token-authenticated WebSocket connection open using the supported client ID `cli`, mode `cli`, operator read/write/admin scopes, and no device identity. Use the same permitted session and unchanged route throughout; there are no attachments or pending questions. A and B below denote distinct returned run IDs.

1. Send A with `chat.send`, `message: "INITIAL_HELD_TURN_steer-abort"`, `deliver: false`, and an explicit `idempotencyKey`. Hold its first provider response before sending any response bytes. Prepare that response to return an ordinary `session_status` function call.
2. While A is held, send B with `chat.send` on the same connection and session, using `message: "CANCELLED_STEER_SENTINEL_steer-abort"`, `queueMode: "steer"`, `deliver: false`, and a different `idempotencyKey`. Wait for B's `status: "started"` acknowledgment.
3. Confirm through `chat.history` that B is absent. Call `chat.abort` with `{ sessionKey, runId: B }`. Wait for the successful acknowledgment naming B alone. Check history again before releasing the provider response; B is still absent.
4. Only after that abort acknowledgment, release A's first response. Capture its next provider request and hold the second response.
5. Inspect the second request, history, and A's lifecycle. Replay B's original `chat.send` with its original idempotency key, then release the second provider response and let A finish.

### Expected behavior

A successful cancellation of pending B should remove B before it reaches a model request, while A continues normally. B should remain absent from consumed history and should not return through fallback or idempotency replay.

The [steering documentation](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/docs/concepts/queue-steering.md#L15-L26) places delivery at runtime boundaries, and the embedded queue already has [cancellation for an individual pending message](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/src/agents/embedded-agent-runner/run/attempt-queue-message.ts#L115-L143). This expectation applies before consumption; cancellation cannot undo input already sent to the model.

### Actual behavior

B appears as one user message in A's second provider request, alongside the tool-result continuation from A's first request. Canonical history records B with `steerTargetRunId: A`. While the second response is held, the session remains running and A has no terminal event. A subsequently ends normally.

Replaying B's idempotency key returns its aborted terminal projection (`status: "timeout"`, `summary: "aborted"`) and creates no additional provider request. Nevertheless, B remains in the final history once, attributed to A.

As a secondary effect, the unstarted `session_status` call receives the synthetic result `Skipped due to queued user message.` The provider returns fixed responses, so this proves unwanted model input, not a real model following B.

### OpenClaw version

Source checkout at [b5907743200dc657c7c8916888c64d30c0adf7a2](https://github.com/openclaw/openclaw/commit/b5907743200dc657c7c8916888c64d30c0adf7a2), package version `2026.9.4`. Runtime reproduction recorded September 11, 2026. An affected released-version range has not been established.

### Operating system

Linux; distribution/kernel version not recorded. Node 24.19.0, pnpm 12.3.4.

### Install method

Source checkout launched through the supported source launcher in an isolated test environment, using the real Gateway server, repository `GatewayClient`, and embedded runtime.

### Model

`mock-openai/steer-fifo` — synthetic fixture with fixed Responses-compatible replies.

### Provider / routing chain

Persistent Gateway WebSocket client → Gateway → embedded OpenClaw runtime → loopback HTTP fixture (`POST /v1/responses`, SSE).

`agentRuntime.id: "openclaw"`, provider API `openai-responses`, SSE transport, and `openaiWsWarmup: false`. The fixture uses the minimal tool profile, no memory plugin or skills, skipped workspace bootstrap, queue mode `steer`, and zero webchat queue debounce. Private-network access is allowed for the loopback provider. No live model service is involved.

### Logs

Normalized observations from the captured canceled-steer case; A/B replace the actual synthetic run IDs. Times are relative to B's abort acknowledgment.

| Checkpoint | Observation |
| --- | --- |
| Before abort | A's first request and canonical history contain no B marker. |
| Abort acknowledgment, t = 0 | `{"ok":true,"aborted":true,"runIds":["B"]}`; A is excluded. |
| After acknowledgment, before release | Canonical history still contains no B marker. |
| t = +10 ms | Release A's first provider response. |
| t = +70 ms | A's second provider request contains one user message bearing B's marker and A's matching function-call result. |
| While the second response is held | A has no terminal event; the session is running; history attributes B to A. |
| After releasing the second response | A ends normally; the case finishes with two provider requests total. |

The held response, request contents, matching tool result, and lifecycle observations together distinguish this from input consumed before cancellation or a later independent follow-up run.

Three controls passed in the same capture:

| Control | B in A's continuation | Later provider request containing B | B in final history |
| --- | --- | --- | --- |
| Uncanceled steering | One user message | None | One, attributed to A |
| Ordinary follow-up | Absent | One, after A ends | One, without a steer target |
| Canceled follow-up | Absent | None | Absent |

No built-in UI interaction was exercised. The evidence is the Gateway RPC results, captured provider requests, canonical history, and lifecycle observations summarized above.

### Impact and severity

- **Affected:** the persistent CLI-mode Gateway connection and embedded runtime path described above.
- **Suggested severity:** P2 / medium, with high confidence in the reproduced path. An acknowledged per-submission cancellation leaves the withdrawn direction in the original run's next request.
- **Observed frequency:** reproduced in the controlled canceled-steer case; real-world frequency is `NOT_ENOUGH_INFO`.
- **Consequence:** withdrawn input still reaches the model request while the original task continues. Real-model compliance, provider charges, and external actions were not measured or exercised.

### Additional information

**Likely cause and proposed remedy**

The Gateway [checks B's own abort controller but creates the injection starter without its signal](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/src/gateway/server-methods/chat-send-handler.ts#L421-L437). The [direct injection options](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/src/gateway/server-methods/chat-send-message-injection.ts#L140-L165) set `waitForTranscriptCommit: true` without passing a source `abortSignal`.

The existing [queue options contract](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/src/auto-reply/reply/reply-run-registry.contracts.ts#L28-L52) already carries `abortSignal` and `queueIdentity`. The [registry forwards those backend options](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/src/auto-reply/reply/reply-run-registry.message-injection.ts#L198-L215), and the embedded queue's [abort listener](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/src/agents/embedded-agent-runner/run/attempt-queue-message.ts#L287-L295) invokes removal of that exact pending message and its display entry. The [later Gateway abort-marker check](https://github.com/openclaw/openclaw/blob/b5907743200dc657c7c8916888c64d30c0adf7a2/src/gateway/server-methods/chat-send-message-injection.ts#L300-L327) suppresses B's completion event after the delivery wait; it cannot retract model input.

Carry B's cancellation signal through the existing direct injection path to that queue owner. Preserve A's lifecycle, prevent canceled B from resurfacing through fallback/replay, and verify the acknowledgment-versus-consumption boundary. Reuse the existing cancellation and queue-identity contracts. This is a proposed direction, not an implemented or tested patch.

**Freshness and scope**

A source-only check on September 12, 2026 at main revision [8b5fa242a2fa2f3c689f3125304a8000807d1ef0](https://github.com/openclaw/openclaw/blob/8b5fa242a2fa2f3c689f3125304a8000807d1ef0/src/gateway/server-methods/chat-send-message-injection.ts#L140-L165) still shows the signal omission in the inspected Gateway-to-queue path. That revision was not runtime-executed for this report.

The proof covers one persistent connection, explicit plain-text steering, unchanged session/route, and the embedded runtime. It does not establish behavior for separate one-shot CLI invocations, reconnects, attachments, pending questions, cross-client cancellation, session-wide abort, other runtimes, or browser/TUI/mobile controls. UI/TUI pending-input handling has additional checks, so its behavior cannot be inferred from this case. No privilege escalation was exercised.

Related work has different triggers: [#127996](https://github.com/openclaw/openclaw/pull/127996) concerns restoring undelivered steering when A aborts; [#119837](https://github.com/openclaw/openclaw/pull/119837) concerns ordering after A ends; [#124409](https://github.com/openclaw/openclaw/pull/124409) corrects audit labels after unconfirmed steering. This report concerns withdrawing B while A continues. A bounded read-only check on September 12 found no exact active match; it is not an exhaustive novelty claim.

shakkernerd:
Closes #145727

## What Problem This Solves

Fixes an issue where users canceling a pending `chat.send` in `steer` mode could still have that input delivered to the active OpenClaw run after `chat.abort` acknowledged cancellation.

## Why This Change Was Made

Pass the steering submission's abort signal to the existing per-message queue owner so it withdraws only the canceled input. Fallback dispatch retains the same signal, and the steering guide explains the withdrawal boundary before delivery starts.

## User Impact

Pending steering can be canceled without stopping the active run or starting a later turn. Input already being delivered or consumed retains the existing handling for uncertain delivery.

## Evidence

- Added a real Gateway and embedded-runtime regression using a held synthetic Responses server and one persistent CLI connection. It checks B-only abort acknowledgment before releasing A, excludes B from model input and history, requires A's successful final result, and verifies aborted replay creates no extra request.
- Before: the regression fails on `ca28965b39be18168670ec6d9aa41dd11a68cb5a` because A's next model request contains the canceled marker.
- After: the repaired regression passed on Testbox, including successful A completion, no canceled B in model input/history, and no replay/fallback request. The synthetic provider exercises the real Gateway and OpenClaw runtime; it does not test real-model compliance.
- Changed-file lint, formatting, conflict checks, max-lines and assertion-safety ratchets passed. Independent engineering review and separate P0–P2 autoreview are clear.
- Final affected Testbox validation passed all 163 tests across five files, including the seven real Gateway steering scenarios. The complete changed-path gate passed, including core and test type graphs, lint, storage/runtime guards, and docs audit (13,529 internal links; zero broken links).
- [Hosted CI](https://github.com/openclaw/openclaw/actions/runs/34683350608) passed, including the required CI gate. [ClawSweeper review](https://github.com/openclaw/openclaw/pull/145809#issuecomment-5644769436) rated the exact head `f9a04a3675e85d05c75cb5464a938b84941318e3` Platinum with no findings.

### DIFF
--- a/docs/concepts/queue-steering.md
+++ b/docs/concepts/queue-steering.md
@@ -72,6 +72,17 @@ has consumed it.
 
 Use `followup` or `collect` when you want messages to queue by default instead of steering the active run. Use `interrupt` when the newest prompt should replace the active run.
 
+## Canceling a pending steer
+
+An authorized Gateway client can withdraw a message still waiting in the OpenClaw
+runtime's steering queue, before delivery starts, with `chat.abort({ sessionKey,
+runId })`. Use the `runId` returned by that message's `chat.send`. This withdraws
+that message without stopping the active run or retrying it as a followup.
+
+Once delivery starts, cancellation cannot guarantee withdrawal or undo completed
+work. If delivery cannot be confirmed, the existing steering safeguards can stop
+the active run to avoid replaying input whose consumption is uncertain.
+
 ## Debounce
 
 The built-in queue debounce applies to queued `followup` and `collect` delivery. In `steer` mode with the native Codex harness, it also sets the quiet window before sending batched `turn/steer`. OpenClaw active steering does not use the debounce timer; at tool-launch and model boundaries it drains FIFO according to the runtime's configured steering drain mode.
--- a/src/gateway/server-methods/chat-send-handler.ts
+++ b/src/gateway/server-methods/chat-send-handler.ts
@@ -426,6 +426,7 @@ async function handleChatSendWithOptions(
     }
     const beginCapturedMessageInjection = createChatSendMessageInjectionStarter({
       target: messageInjectionTarget,
+      abortSignal: activeRunAbort.controller.signal,
       request: normalizedRequest.value,
       session: preparedSession.value,
       admittedSessionSettings: admitted.value.admittedSessionSettings,
--- a/src/gateway/server-methods/chat-send-message-injection.test.ts
+++ b/src/gateway/server-methods/chat-send-message-injection.test.ts
@@ -132,6 +132,7 @@ function makeStarterParams(params?: { entry?: unknown; loadLatest?: unknown }) {
       replyOptionMedia: [],
     },
     imageOrder: [],
+    abortSignal: new AbortController().signal,
     userTurnTranscriptRecorder: {},
     logGateway: { warn: vi.fn() },
   } as unknown as Parameters<typeof createChatSendMessageInjectionStarter>[0];
@@ -440,6 +441,7 @@ describe("createChatSendMessageInjectionStarter", () => {
 
     return {
       target,
+      abortSignal: new AbortController().signal,
       request: {
         p: { sessionKey, message: rawMessage, idempotencyKey: "steer-input" },
         rawMessage,
--- a/src/gateway/server-methods/chat-send-message-injection.ts
+++ b/src/gateway/server-methods/chat-send-message-injection.ts
@@ -39,6 +39,7 @@ import type { GatewayRequestContext } from "./types.js";
 /** Captures the prepared request data used by both pre-ACK and detached injection attempts. */
 export function createChatSendMessageInjectionStarter(params: {
   target: ReplyMessageInjectionTarget | undefined;
+  abortSignal: AbortSignal;
   request: Pick<NormalizedChatSendRequest, "p" | "rawMessage" | "supportsTaskSuggestions">;
   session: Pick<
     PreparedChatSendSession,
@@ -159,6 +160,7 @@ export function createChatSendMessageInjectionStarter(params: {
         ...(params.imageOrder?.length ? { imageOrder: params.imageOrder } : {}),
         ...(replyOptionMedia?.length ? { media: replyOptionMedia } : {}),
         waitForTranscriptCommit: true,
+        abortSignal: params.abortSignal,
         ...(debounceMs !== undefined ? { debounceMs } : {}),
         taskSuggestionDeliveryMode: supportsTaskSuggestions ? "gateway" : undefined,
         userTurnTranscriptRecorder: params.userTurnTranscriptRecorder,
@@ -209,11 +211,6 @@ export async function finalizeAcceptedChatSendMessageInjection(params: {
 }): Promise<boolean> {
   const { context, ctx, session } = params;
   const { agentId, cfg, clientRunId, entry, sessionKey, storePath } = session;
-  // Terminal-receipt admission is fenced at the injection-start boundary
-  // (createChatSendMessageInjectionStarter), before the steer is queued, so
-  // no attempt ever exists for a fail-closed session. The only rejection
-  // left here is the runtime refusing the queued steer, which also means
-  // nothing was enqueued — the fallback to follow-up dispatch is safe.
   const finalizedCtx = finalizeInboundContext(ctx);
   const finalization = await finalizeReplyMessageInjectionAttempt({
     attempt: params.attempt,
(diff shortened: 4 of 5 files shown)


shakkernerd:
Fixed: canceling steering before delivery withdraws the queued message while the active OpenClaw run continues.

PR #145812, opened 2026-09-12
perf(infra): skip route probes for a single LAN address

### LINKED ISSUE #145811: LAN pairing probes routes when only one eligible address exists
LAN pairing resolves the platform default route even when the machine has only one eligible private IPv4 address. That subprocess cannot change the choice, but it is started again for each lookup.

Return the sole eligible address after the existing fresh interface read. Keep route hints, metric ordering and fallback selection for multiple addresses, without adding a cache or changing pairing payloads.

A paired synthetic-input comparison through the actual pairing resolver kept ten complete result/code rows identical and made 42 fresh interface reads per version. Native child starts fell from 40 to 4 overall, including 32 to 0 across 32 singleton calls. The children were real Node processes producing synthetic route output; no real interface inventory or platform route command was used. These are operation counts, not timing or memory measurements.

steipete:
Closes #145811

<details>

</details>

## What Problem This Solves

LAN pairing starts a platform route probe even when only one eligible private IPv4 address exists and the probe cannot change the selected address.

## Why This Change Was Made

The LAN resolver now returns that sole address directly from the existing normalized interface entries, removing the candidate copy and private selection wrappers. Multiple-address selection retains route hints, metric ties, aliases, Unicode handling and fallbacks.

## User Impact

Single-address lookups avoid unnecessary subprocess work while keeping a fresh interface read on every call. Pairing URL/auth/limited-access payloads, public signatures and SDK behavior remain unchanged. The change removes 28 net production lines and adds three net lines to existing tests.

## Evidence

Against baseline `0cccc6d68e58ccd5030cd09931c1753bca924c7b`, each version passed 64 tests; one existing native Windows-only case remained skipped on Darwin. The updated singleton pairing test failed against the original owner because it performed one unnecessary probe. The source-bound P2 review and full selected changed gate passed.

The actual pairing resolver produced ten identical complete result/code rows using synthetic interface snapshots and bootstrap data. Both versions made 42 fresh interface reads. Native child starts fell 40→4 overall and 32→0 across the 32-call singleton workload. The canonical process runner used real Node children producing synthetic route output, with unchanged platform argv, 3,000 ms timeout and 16,384-byte output bounds checked.

This is operation-count proof on Darwin, not a measurement of latency, allocation bytes or RSS. It used no real interface inventory or platform route command, and native Windows execution remains unrun locally.

### DIFF
--- a/src/infra/advertised-lan-host.ts
+++ b/src/infra/advertised-lan-host.ts
@@ -14,12 +14,6 @@ const WINDOWS_DEFAULT_ROUTE_COMMAND =
   "[Console]::OutputEncoding=[Text.UTF8Encoding]::new($false); Get-NetRoute -AddressFamily IPv4 -DestinationPrefix '0.0.0.0/0' | " +
   "Select-Object -Property InterfaceAlias,RouteMetric,InterfaceMetric | ConvertTo-Json -Compress";
 
-type AdvertisedLanHostCandidate = {
-  interfaceName: string;
-  address: string;
-  order: number;
-};
-
 type AdvertisedLanRouteHint = {
   interfaceName: string;
 };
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [d0489162765d0ee78d31d1d5a6c1ab55ebe0c8bf](https://github.com/openclaw/openclaw/pull/145812/commits/d0489162765d0ee78d31d1d5a6c1ab55ebe0c8bf)
- Landed commit: [0c6e0f07305962e7428f4c23b1fbcd543270cde8](https://github.com/openclaw/openclaw/commit/0c6e0f07305962e7428f4c23b1fbcd543270cde8)

PR #145822, opened 2026-09-12
test: use installed ws for worker fault injection

steipete:
## Problem

The direct-Bun source-unit lane stalled six worker fault-injection cases at connection admission. Bun substitutes its built-in `ws` server, whose sockets do not expose the installed `ws` receiver or its mutable payload limit. The harness therefore rejected its own connection through OpenClaw's fail-closed post-auth payload-limit guard.

The product Gateway already forces the installed `ws` implementation for this reason; the fault-injection harness had bypassed that adapter. Shipping Bun also reports `ENOTEMPTY` rather than Node's `EACCES` for the real recursive permission failures used by the matrix.

## Solution

Use the existing installed-`ws` test transport in the worker harness so it exercises the same mutable receiver contract as the Gateway. Keep Node's exact `EACCES` assertion and accept Bun's current `EACCES`/`ENOTEMPTY` variants until [oven-sh/bun#42446](https://github.com/oven-sh/bun/pull/42446) lands.

## Impact

This changes test infrastructure only. Product dependencies, installation behavior, and Gateway runtime behavior are unchanged.

## Evidence

- Before: direct custom Bun timed out at 120 seconds per worker fault-injection case after admission rejected the receiver.
- Shipping Bun 1.4.2: worker fault-injection matrix 15/15 after the adapter and temporary errno compatibility assertion.
- Patched custom Bun: worker fault-injection matrix 15/15.
- Node: worker fault-injection matrix 15/15.
- `node scripts/check-changed.mjs`
- P0-P2 autoreview: clean.

### DIFF
--- a/src/worker/worker-fault-injection.test-support.ts
+++ b/src/worker/worker-fault-injection.test-support.ts
@@ -4,7 +4,8 @@ import { createServer, type Server } from "node:http";
 import path from "node:path";
 import { rawDataToString } from "@openclaw/gateway-client/websocket-data";
 import { expectDefined } from "@openclaw/normalization-core";
-import { WebSocket, WebSocketServer, type RawData } from "ws";
+import { WebSocket, type RawData } from "ws";
+import { WebSocketServer } from "../../packages/gateway-client/src/websocket.test-support.js";
 import {
   type WorkerLiveEventParams,
   WORKER_PROTOCOL_FEATURES,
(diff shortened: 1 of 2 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [22d9aea0847730c83cc342ba267660c0336cb079](https://github.com/openclaw/openclaw/pull/145822/commits/22d9aea0847730c83cc342ba267660c0336cb079)
- Landed commit: [dd937980a3ce791b673730ec8085cce8c18eb3bd](https://github.com/openclaw/openclaw/commit/dd937980a3ce791b673730ec8085cce8c18eb3bd)

PR #145826, opened 2026-09-12
fix: preserve code expansion and wrapping as replies stream

steipete:
## What Problem This Solves

Fixes an issue where users expanding or wrapping a completed code block in a streaming assistant reply would lose those choices when a later paragraph finished.

## Why This Change Was Made

The message renderer now keeps unchanged completed HTML sections in place and updates the growing tail separately. Its private rendering directive uses the existing message identity and canonical HTML to decide when to preserve or replace the view. The Markdown parser still owns all content boundaries, and the existing controls and observer lifecycle remain unchanged.

## User Impact

Completed top-level code blocks retain expansion and wrapping as later paragraphs arrive. A replaced message, corrected earlier content, or changed rendered prefix gets a fresh view. Finishing a reply can retain the same controls when its message view stays mounted; this does not restore controls after the transcript replaces the whole row.

## Evidence

- Final regression tests fail on the original renderer for later-paragraph updates, completion within a retained view, and replacement message identity.
- 80 selected tests pass: 64 renderer/parser/cache cases and 16 code-control lifecycle cases, including corrected content, rendering options, late references, and duplicate badges.
- Complete changed-file checks and a fresh Control UI build pass.
- Independent review completed with no actionable findings through P2.

The synthetic browser fixture uses the actual message renderer, Markdown lifecycle, production styles, and click handler. All nine before/after frames were inspected. It deliberately retains the message view for its completion step, and does not claim a full Chat-route or live Gateway run. An initial fixture serialization error was corrected before the clean before/after runs.

| Before: a later paragraph resets code controls | After: expansion and wrapping remain |
| --- | --- |
| A later streamed paragraph resets expanded and wrapped code | A later streamed paragraph preserves expanded and wrapped code |

### DIFF
--- a/docs/web/control-ui/chat.md
+++ b/docs/web/control-ui/chat.md
@@ -238,6 +238,10 @@ code's leading whitespace and final newline when present. Indented Markdown code
 blocks also work at the start of a message and remain literal while streaming,
 including blank lines within the block.
 
+Completed top-level code blocks keep your expansion and wrapping choices while
+later paragraphs stream into the same assistant reply. Replacing the message or
+correcting earlier content starts a fresh view.
+
 **Copy URL** in browser tab cards also works on plain HTTP connections where the
 browser does not provide its Clipboard API.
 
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [add1e4453aaf6666b139a2ea679c3e09bfea0819](https://github.com/openclaw/openclaw/pull/145826/commits/add1e4453aaf6666b139a2ea679c3e09bfea0819)
- Landed commit: [629e2a9a65f3ab1c0a5d2ed3054c2cae382b8f93](https://github.com/openclaw/openclaw/commit/629e2a9a65f3ab1c0a5d2ed3054c2cae382b8f93)

Issue #145722, opened 2026-09-12
[Bug]: Cron startup recovery overwrites an acknowledged cadence edit with an old next-run time


shakkernerd:
**Bug type:** Behavior bug (incorrect output/state without crash)

**Beta release blocker:** No

**Summary**

When a successful cron run saves its history but fails its scheduler-row completion, startup recovery can overwrite a later acknowledged cadence edit with the completed run's old next-run timestamp.

**Steps to reproduce**

The controlled reproduction uses real `CronService` admission, updates, startup recovery, task history, run receipts, and native SQLite. Time, condition evaluation, and payload execution are injected; automatic scheduling is paused to isolate the sequence.

1. Create an enabled condition watcher with an hourly `every` schedule anchored at 12:00:00 UTC. Its condition fires, its payload succeeds, and `once` is unset.
2. At 13:00:00, call `cron.run(job.id, "due")`. Hold the payload after admission and verify the real active receipt.
3. Install a connection-local temporary SQLite `BEFORE UPDATE` trigger that aborts writes to this job's `cron_jobs` row. Complete the payload successfully at 13:00:01. Its terminal task/history commits with the next run at 14:00:00, while the separate job-row write fails and leaves the running marker and active receipt. Remove the temporary trigger and stop this service instance.
4. At 13:00:02, call `cron.update` with `schedule: { kind: "every", everyMs: 60000, anchorMs: nowMs }`. Verify that the acknowledged job is enabled, has a one-minute cadence, and has its next run at 13:01:02.
5. Start a fresh `CronService` instance against the same isolated database, with automatic scheduling still paused. Read the recovered job and its history.

The failure injection rejects the selected scheduler write after real task persistence. It does not fabricate history, edit receipt records, or kill a process.

**Expected behavior**

Preserve the newer committed cadence and its next run at 13:01:02. Retain completed history, clear the old running marker, and settle its receipt without executing the payload again. The [schedule-update owner](https://github.com/openclaw/openclaw/blob/a53c601cd4075a6839bac3f75f2280eda182b4b8/src/cron/service/ops-mutations.ts#L202-L212) recomputes the next run from the new inputs; the reproduction verifies that value before recovery.

**Actual behavior**

The job remains enabled and its schedule still says every minute, but recovery restores the old hourly next-run timestamp:

| Observation | After acknowledged edit | After recovery |
| --- | --- | --- |
| `schedule.everyMs` | `60000` | `60000` |
| `enabled` | `true` | `true` |
[...]

PR #145828, opened 2026-09-12
fix: preserve cron cadence edits during startup recovery

shakkernerd:
Refs #145722

## What Problem This Solves

After a completed run saved its history but failed to update the job row, startup recovery could replace a newly acknowledged cadence with the old run's next-check time. An hourly-to-minute edit reproduced a delay of 58 minutes and 58 seconds.

## Why This Change Was Made

Record each committed scheduling edit with the pending run in the same transaction as the edit. Completion and recovery can then settle that run while preserving the acknowledged scheduling state, including repeated or backward clocks and a passive editor whose snapshot spans successive runs. Starting a new run and cleaning up pending-run state clear the private marker. This builds on the landed run-identity protection in #145734 and adds no schema migration or public API field.

Pending runs edited by older builds retain legacy recovery behavior when no durable edit marker exists. Older builds do not enforce the new protection; complete pending runs and scheduler reconciliation before downgrading when the edited cadence must be preserved. The canonical cron and database compatibility docs describe these boundaries.

## User Impact

Newly acknowledged cadence and pacing edits retain their next check through recovery. Completed history remains available, and recovery does not rerun the completed payload. Unchanged jobs retain their normal completion and retry behavior.

## Evidence

- All 732 selected native tests across 37 files pass on `26c1c100`, after integrating the landed run-identity fix. The run uses real CronService paths and SQLite, including completion-write failure and child-process races.

- Native fault injection covers late edits, repeated and backward clocks, cron overlap, timed/stream transitions, pacing, transient-error deadlines, unchanged/idempotent/failed-edit controls, once-run successors, and a separate-process editor spanning two runs.
- The run-identity and receipt-retention suites also exercise the integrated behavior from #145734. Public projection and patch tests verify that clients cannot read or forge the private marker.
- Released v2026.9.4 compatibility proof uses its own frozen dependencies: it reads the additive pending state without changing the database, writes a separate settled fixture, and the candidate reopens that fixture successfully. Pending downgrade protection is explicitly excluded.
- Local checks passed for formatting, type-aware lint, line and assertion ratchets, conflict hygiene, and unchanged state schema 17. Full remote changed-path validation and required hosted CI are landing gates for the exact PR head.


shakkernerd:
Fixed: cron recovery preserves newly acknowledged cadence and pacing edits without replaying completed runs.

PR #145849, opened 2026-09-12
perf(boards): reuse compiled SQLite read queries

steipete:
Related: #144592

## What Problem This Solves

Repeated board reads and updates rebuild the same session, tab, and widget queries on every call, adding CPU work before SQLite runs them.

## Why This Change Was Made

Reuse compiled, parameter-bound Kysely queries per database connection. Existing transaction boundaries, native statement-cache ownership, lazy schema checks, and row ordering stay with their current owners. Typed query results also remove two row-array assertions, so their assertion baseline shrinks accordingly.

## User Impact

Warm board operations spend less time preparing reads. This is query preparation work for the continuing database cleanup; board SQLite execution still needs its separate worker migration.

## Evidence

- 49 board-store, codec, parity, and Gateway tests passed, including shared-store and reopen behavior.
- Paired synthetic SQLite query measurements, nine alternating rounds of 2,000 calls on one board with 48 widgets: session lookup 10.90 → 7.08 μs; tab read 14.79 → 10.03 μs; widget read 95.56 → 84.19 μs. These measure query calls, not end-to-end UI latency.
- SQL matched exactly; results matched for two separate session keys and a missing key.
- Full changed checks passed, including core and core-test typechecks, lint, dead-export and import-cycle checks.
- Independent Codex review found no actionable P2-or-higher findings.

### DIFF
--- a/config/assertion-safety-baseline.txt
+++ b/config/assertion-safety-baseline.txt
@@ -2115,7 +2115,7 @@ src/auto-reply/usage-bar/template.ts	1
 src/auto-reply/usage-bar/translator.ts	4
 src/boards/board-store.ts	1
 src/boards/sqlite-board-codec.ts	7
-src/boards/sqlite-board-store.ts	3
+src/boards/sqlite-board-store.ts	1
 src/bootstrap/node-startup-env.ts	1
 src/canvas/documents.ts	1
 src/canvas/serve.runtime.ts	1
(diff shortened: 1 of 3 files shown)


steipete:
Merged via squash.

- Prepared head SHA: [518971b1c929e5809c6417708de02da5bc7e4ef2](https://github.com/openclaw/openclaw/pull/145849/commits/518971b1c929e5809c6417708de02da5bc7e4ef2)
- Landed commit: [b3fadbf914be697f1a534c4cff81293db1f8a647](https://github.com/openclaw/openclaw/commit/b3fadbf914be697f1a534c4cff81293db1f8a647)

